PR #968, opened 2015-11-01
Display Run Settings Report. Fixes #967.

### LINKED ISSUE #967: Run Settings Report is not being displayed.
The console runner has a run settings report, which is not being displayed for some reason. In addition, it was intended that this report would also be in the nunitlite runner but it is not.

CharliePoole:
Run settings was not working in nunit-console because of a change in the XML format. It had never been ported to nunitlite. Both are done in this PR.

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/ResultReporterTests.cs
@@ -40,7 +40,7 @@ namespace NUnit.ConsoleRunner.Tests
         private static readonly string[] REPORT_SEQUENCE = new string[] {
             "Tests Not Run",
             "Errors and Failures",
-            "Run Settings",
+            //"Run Settings",
             "Test Run Summary"
         };
 
--- a/src/NUnitConsole/nunit3-console/ResultReporter.cs
+++ b/src/NUnitConsole/nunit3-console/ResultReporter.cs
@@ -77,21 +77,25 @@ namespace NUnit.ConsoleRunner
 
         private void WriteRunSettingsReport()
         {
-            var settings = _result.SelectNodes("settings/setting");
-
-            if (settings.Count > 0)
+            var firstSuite = _result.SelectSingleNode("test-suite");
+            if (firstSuite != null)
             {
-                _writer.WriteLine(ColorStyle.SectionHeader, "Run Settings");
+                var settings = firstSuite.SelectNodes("settings/setting");
 
-                foreach (XmlNode node in settings)
+                if (settings.Count > 0)
                 {
-                    string name = node.GetAttribute("name");
-                    string val = node.GetAttribute("value");
-                    string label = string.Format("    {0}: ", name);
-                    _writer.WriteLabelLine(label, val);
-                }
+                    _writer.WriteLine(ColorStyle.SectionHeader, "Run Settings");
 
-                _writer.WriteLine();
+                    foreach (XmlNode node in settings)
+                    {
+                        string name = node.GetAttribute("name");
+                        string val = node.GetAttribute("value");
+                        string label = string.Format("    {0}: ", name);
+                        _writer.WriteLabelLine(label, val);
+                    }
+
+                    _writer.WriteLine();
+                }
             }
         }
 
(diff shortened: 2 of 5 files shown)


oznetmaster:
Should the DisplayRequestedOptions method and its test be removed as part of this fix?

Should a test for DisplayRunSettings be added in its place?

oznetmaster:
The parts that I have an understanding of look good.  I have never worked with the nunit3.console, so cannot really comment on that.

oznetmaster:
The problem I still have with this is that this information is now displayed on the console after the run, not before it as was the case previously.

If the run dies for some reason, there is no visible indication of the options used.  The runtime environment and test files are still displayed before the run, but not the settings.

rprouse:
Looks good to me too. We can always make minor changes between the RC and the final release if we have to and once we release and get on a faster release cadence, we can get fixes out more quickly. Also, I know this is important to @oznetmaster, but nunitlite is averaging less than 3% the number of downloads as the framework, so we shouldn't hold up the release for it.

oznetmaster:
Still leaves the issue that previously, the settings were displayed before the test was run, and now they are displayed after.  This is not an nunitlite only issue.  It is true in nunit3-console as well.

oznetmaster:
Still leaves this as well:

Should the DisplayRequestedOptions method in nunitlite.runner/TextUI.cs and its test be removed as part of this fix? (It is being removed nunitlite.runner/Portable/TextUI.cs)

Should a test for DisplayRunSettings be added?

CharliePoole:
@oznetmaster I'll look at those two items between the rc and the final.

oznetmaster:
My concern is that the first of those is a significant change between beta 5 and rc.  That is why I bring it up now.

CharliePoole:
Yes... it's possible we may roll it back.

I have the same notion as you, that significant changes should not be introduced in an RC. It's how I was taught. :-) But having worked on 3.0 since 2008, I'm getting impatient!

oznetmaster:
As I have expressed, not only is this significant, but I believe it removes valuable information form the console display that could help to determine/debug a problem in the tests if no results are generated for some reasons.

CharliePoole:
I take your point and I'm willing to go back if it makes sense after the
RC. That said, I'm not sure the old RunSettings section was displaying such
valuable info:

It had the trace level, work directory, process model and domain setting.
It also had the included and excluded categories, which is now replaced by
the where clause. The where clause is displayed in a different section at
the start of the run.

CharliePoole:
BTW, if the test fails with no results generated, the console leaves behind
an xml file with just the command line. I don't think we do that in
NUnitLite, but I'd have to check the code to be sure. In CF, we couldn't
use Environment.CommandLine as we do in other builds, but we could probably
come up with something.

oznetmaster:
It also included the number of worker threads.  I think the n umber of workers, work directory, and trace level, are all useful.

The command line appears to be added to the xml file AFTER the tests are run.  You are correct that it not done for CF.

oznetmaster:
Since we do get the command line, sort of, in the AutoRun.Execute method (the args are passed), the command line can be reconstituted in the CF build and made available.

CharliePoole:
Good idea.

CharliePoole:
Number of threads was initially only included if set on the command line.
People complained about that so I displayed it using the framework
algorithm. However, that's not really valid because the engine does not
even know if the framework in use supports threads! That's why I had the
framework include the number of threads in its own XML output. However,
that's not available till after the run.

The problem could be avoided by creating a query interface whereby a
framework could be asked about various capabilities. That's a major design
feature though.

In NUnitLite, we can do something different if we like, since nunitlite
only runs with one framework.

Regarding the small XML file left behind after a crash, see
https://github.com/nunit/nunit/blob/master/src/NUnitEngine/nunit.engine/Services/ResultWriters/NUnit3XmlResultWriter.cs#L45
The engine calls CheckWritability on a writer before trying to use it. The
writer can do anything it likes to verify writability. The NUnit3 driver
creates a mini result file.

oznetmaster:
I meant Number of Test Workers, which is an option setting.

CharliePoole:
Yes, that's what I meant as well. We used to display it if you entered it on the command-line, but not otherwise.

PR #975, opened 2015-11-02
Fix Run Settings display in CF. Fixes #970

### LINKED ISSUE #970: Define PARALLEL in CF build of nunitlite
Is the PARALLEL build being officially supported in nunitlite?

If so, setting it cause a build error in the CF build.

I had not noticed earlier because I always have supported it in my CF builds.

Specifically, in DisplayRunSettings: Environment.ProcessorCount is not supported in CF.

CharliePoole:
This takes care of the ProcessorCount issue in CF and adds PARALLEL to the build of NUnitLite for CF so that the number of worker threads is displayed.


oznetmaster:
Looks good to me.

PR #984, opened 2015-11-03
Add version option to console and nunitlite. Fixes #978

### LINKED ISSUE #978: It should be possible to determine version of NUnit using nunit console tool
Currently nunit console started without arguments outputs NUnit version. But it does not look like a reliable way to obtain version as this default behavior can easily change in the future. It would be great to have dedicated argument for the version.

CharliePoole:
First shot at fixing #978. I'd like @NikolayPianikov to review in addition to at least one of the committers, since he's the one who asked for it. :-)

Particular questions... 

1. Should I further change the format to move the (Debug) in front of the version so the first line reads like...
    NUnit Console Runner (Debug) 3.0.1234

2. I did the same thing with NUnitLite. NUnitLite also shows some stuff in parentheses after the version, even in release. If I move the parenthesized stuff for console, I'll do the same for NUnitLite.

3. In addition to --version | -V I can add some special, long option just for use by CI programs if it's really needed... something like --version-number perhaps.

### DIFF
--- a/src/NUnitConsole/nunit3-console/Program.cs
+++ b/src/NUnitConsole/nunit3-console/Program.cs
@@ -77,7 +77,7 @@ namespace NUnit.ConsoleRunner
             //log.Info("NUnit3-console.exe starting");
             try
             {
-                if (!Options.NoHeader)
+                if (Options.ShowVersion || !Options.NoHeader)
                     WriteHeader();
 
                 if (Options.ShowHelp || args.Length == 0)
--- a/src/NUnitConsole/nunit3-console/Program.cs
+++ b/src/NUnitConsole/nunit3-console/Program.cs
@@ -86,6 +86,10 @@ namespace NUnit.ConsoleRunner
                     return ConsoleRunner.OK;
                 }
 
+                // We already showed version as a part of the header
+                if (Options.ShowVersion)
+                    return ConsoleRunner.OK;
+
                 if (!Options.Validate())
                 {
                     using (new ColorConsole(ColorStyle.Error))
--- a/src/NUnitFramework/nunitlite.runner/AutoRun.cs
+++ b/src/NUnitFramework/nunitlite.runner/AutoRun.cs
@@ -86,6 +86,10 @@ namespace NUnitLite
                 return TextRunner.OK;
             }
 
+            // We already showed version as a part of the header
+            if (options.ShowVersion)
+                return TextRunner.OK;
+
             if (options.ErrorMessages.Count > 0)
             {
                 _textUI.DisplayErrors(options.ErrorMessages);
--- a/src/NUnitFramework/nunitlite.runner/AutoRun.cs
+++ b/src/NUnitFramework/nunitlite.runner/AutoRun.cs
@@ -77,7 +77,7 @@ namespace NUnitLite
 
             var _textUI = new TextUI(outWriter, options);
 
-            if (!options.NoHeader)
+            if (options.ShowVersion || !options.NoHeader)
                 _textUI.DisplayHeader();
 
             if (options.ShowHelp)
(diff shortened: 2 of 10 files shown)


oznetmaster:
I think that nunitlite.tests/commandlinetests.cs needs to be updated as well.

Same as nunit3-console.tests/CommandLineTests.cs.

CharliePoole:
Good call. I updated the tests.

oznetmaster:
You had said you wanted to follow the gnu guidelines.

They state that if the --version option is specified, ALL other options are ignored.  That does not appear to be how this has been implemented.

Both the header and the help option are processed before the version option is even checked . And the version option always terminates processing options even if the header has not been shown.

```
            if (!options.NoHeader)
                _textUI.DisplayHeader();

            if (options.ShowHelp)
            {
                _textUI.DisplayHelp();
                return TextRunner.OK;
            }

            // We already showed version as a part of the header
            if (options.ShowVersion)
                return TextRunner.OK;
```

Where is the version displayed if the header has not been shown?

src/NUnitConsole/nunit3-console/Program.cs
```diff
@@ -86,6 +86,10 @@ public static int Main(string[] args)
                     return ConsoleRunner.OK;
                 }
 
+                // We already showed version as a part of the header
+                if (Options.ShowVersion)
```
rprouse: I think the if statement for WriteHeader should be `if (Options.ShowVersion || !Options.Noheader)` here and in NUnitLite. I don't think --version --noheader should display nothing, I think version should take precedence. As for @oznetmaster's comment about other command line options being ignored, I don't think showing help is a terrible sin, but if we wanted, we could move this up above help.

src/NUnitFramework/nunitlite.runner/AutoRun.cs
```diff
@@ -86,6 +86,10 @@ public int Execute(string[] args)
                 return TextRunner.OK;
             }
 
+            // We already showed version as a part of the header
+            if (options.ShowVersion)
+                return TextRunner.OK;
+
```
rprouse: You missed the AutoRun in portable. After 3.0 is released, we should see if we can re-merge these files, probably with your branch to wrap the console?

CharliePoole: I've already done the merge, although not in this branch. That's probably why I forgot about it. :-)

In any case, for this branch to stand on it's own, I'll fix that as well.

src/NUnitConsole/nunit3-console/Program.cs
CharliePoole: You're right, it's more user friendly than a blank screen. :-) 

Even though they told us to display it and not to display it - I know, we could generate a random number to decide! Just kidding, of course. :-)

I figured that if you asked for both version and help you wanted both - or maybe you want help about --version. In either case, you get what you want.

CharliePoole:
@oznetmaster The guildelines actually say the same thing about --help, which makes them contradictory I suppose. However, I think they may be written for an environment where you process arguments in sequence and the first one found rules. We don't do that so we are only presented with the fact that the user requested help and requested version. In that case, I give them both, which seems like a reasonable thing to do.

I agree that we should not suppress the header if --version is requested, as @rprouse proposed, but I think we should keep the exit after the help display for the case where you asked for both.

BTW, I only use the GNU guidelines as an indication of what folks might normally expect to see on the comand-line. I suspect that we have more command-line users on linux than windows these days. There are whole areas of the guidelines, having to do with legalities and licensing that I pretty much ignore. :-)

CharliePoole:
@NikolayPianikov Still waiting to here from you on this PR. We have our code review on the details but still don't know whether you are satisfied with the solution for your purposes!

I'd prefer not to have the comments after we merge. :-(

CharliePoole:
@NikolayPianikov  Didn't actually comment on the PR, but he said in issue #978 that he was satisfied. I'm going ahead and merging.

NikolayPianikov:
@CharliePoole  It is good PR for me, thank you

Issue #990, opened 2015-11-05
V2 driver is passing invalid filter elements to NUnit V2


flaub:
Here's the output I get:

```
$ mono SOME_PATH/bin/nunit3-console.exe SOME_DLL --labels=All --where "class=='SOME_CLASS'"
NUnit Console Runner 3.0.5783
Copyright (C) 2015 Charlie Poole

Runtime Environment
   OS Version: MacOSX 14.5.0.0
  CLR Version: 4.0.30319.17020

Test Files
    SOME_DLL

Test Filters
    Where: class=='SOME_CLASS'


Errors and Failures

1) Error :
Invalid filter element: class
Parameter name: xmlNode
  at System.Runtime.Remoting.Proxies.RealProxy.PrivateInvoke (System.Runtime.Remoting.Proxies.RealProxy rp, IMessage msg, System.Exception& exc, System.Object[]& out_args) [0x00000] in <filename unknown>:0

Test Run Summary
    Overall result: Failed
   Tests run: 0, Passed: 0, Errors: 0, Failures: 0, Inconclusive: 0
     Not run: 0, Invalid: 0, Ignored: 0, Explicit: 0, Skipped: 0
  Start time: 2015-11-05 11:47:00Z
    End time: 2015-11-05 11:47:00Z
    Duration: 0.007 seconds

Results (nunit3) saved as TestResult.xml
```

It does work when filtering on the category (`cat == XXX`).

flaub:
BTW same goes for `test == XXX`

CharliePoole:
I have tried this with and without the double and single quotes and it works every time!

However, when I try it against a test that uses NUnit 2.6.4, I do get that error. Is that your situation? 

The v2 driver is passing filter elements that V2 does not understand. Using them should give an error, but it should be a much clearer one than what you are getting!

CharliePoole:
I changed the title of the issue to reflect what the problem is.

The driver is translating the filter to NUnit V2 filter elements. It detects the bad filter elements and throws an exception. This is definitely as it should be but the message is only clear if you know how the driver works. It should be an exception type for which we do not display a stack trace and should have an explanatory message like "The class filter cannot be used with NUnit V2 tests"

flaub:
Indeed this was against NUnit 2.6.4. Thanks!

flaub:
Is running v2 tests with the v3 console supported? How do I specify which tests to run against v2?

When I use --test, I get "Invalid filter element: test".

[...]

PR #993, opened 2015-11-05
Give error message when a regex filter is used with NUnit V2

CharliePoole:
This is something I just noticed when documenting the change made for #990.

When a regular expression is used, as in
   test =~ MyFixture
the driver was translating it to a V2 SimpleTestFixture, which only compares for equality.

So we now have another error message saying you can't use regular expressions to filter V2 tests.


rprouse:
Looks good to me. I was looking through the build output for the unrelated failed build and noticed that we have a lot of build warnings that have crept in with the filter changes, mainly around missing XML comments. Should we fix them or make the classes internal? Once we get 3.0 out, we should see if we can enable warnings as errors to prevent them creeping in, it happens too often.

CharliePoole:
I've been noticing the same. We can fix if it's convenient - either by adding the comments or making the class internal where appropriate. I don't think it's high priority, but it is annoying.

rprouse:
I don't think we need to fix for the release, I would just like to fix and put processes in place after the release to prevent it from happening.

PR #996, opened 2015-11-06
Return -100 if any test assemblies crash unexpectedly

### LINKED ISSUE #983: Inconsistent return codes depending on ProcessModel
Per @NikolayPianikov...

> When --process=InProces, nunit returns -100 for StackOverflow exception
but when --process=Multiple or Separate, nunit returns 0. Is it possible to return -100? TeamCity looks at the negative exit code to raise a build problem

and...

> When --process=InProces, nunit returns -100 for OutOfMemory exception
but when --process=Multiple or Separate, nunit returns 1 for one failed test.
It will be better if nunit console returns the same exit code

When a separate agent is used, it's the agent process that terminates with -100. We used to detect that the agent had died and terminate the console with -100 as well.

However, in NUnit 3.0, we wanted to continue in order to report tests from any other agents, so we trapped the exception that causes us to terminate. We report the problem with the terminated assembly in the XML and on the console output.

That's all well and good, but we need to somehow reflect the failure in the return code as well.

@NikolayPianikov Imagine two assemblies running in two agents. One of them runs to completion with two failures - the normal return code would be 2. The other gets an OutOfMemoryException - the normal return code would be -100. What should nunit-console return?

rprouse:
Fixes #983

If any test run crashes unexpectedly in an external runner, returns -100

### DIFF
--- a/src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
@@ -232,7 +232,7 @@ namespace NUnit.Engine.Runners
             XmlHelper.AddAttribute(suite, "fullname", TestPackage.FullName);
             XmlHelper.AddAttribute(suite, "runstate", "NotRunnable");
             XmlHelper.AddAttribute(suite, "testcasecount", "1");
-            XmlHelper.AddAttribute(suite, "result", "Failed");
+            XmlHelper.AddAttribute(suite, "result", "UnexpectedError");
             XmlHelper.AddAttribute(suite, "label", "Error");
             XmlHelper.AddAttribute(suite, "start-time", DateTime.UtcNow.ToString("u"));
             XmlHelper.AddAttribute(suite, "end-time", DateTime.UtcNow.ToString("u"));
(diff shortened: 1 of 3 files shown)


src/NUnitEngine/nunit.engine/Runners/ProcessRunner.cs
```diff
@@ -232,7 +232,7 @@ TestEngineResult CreateFailedResult(Exception e)
             XmlHelper.AddAttribute(suite, "fullname", TestPackage.FullName);
             XmlHelper.AddAttribute(suite, "runstate", "NotRunnable");
             XmlHelper.AddAttribute(suite, "testcasecount", "1");
-            XmlHelper.AddAttribute(suite, "result", "Failed");
+            XmlHelper.AddAttribute(suite, "result", "UnexpectedError");
```
CharliePoole: The three attributes result, label and site (not used here) map to the three fields in a ResultState object. The result attribute represents the value of TestStatus, which only has four values. We consider those to be the four fundamentally different outcomes of a test: Passed, Failed, Inconclusive and Skipped. It would be best not to add another one without a darned good reason, and it should really be fundamentally different if we do.

The label in a ResultState is just a string and is intended to be used to create new sub-results. You could do that.

That said, it's possible that you might just use Failed:Error rather than inventing anything new. It depends whether an assembly ever has a Failed:Error state for any other reason. For example, if I remember right, we get a Failed:Invalid if the assembly can not be found.

BTW, we use the ResultState class in the framework but we have not translated the attributes back to an object for the console because it didn't seem necessary so far. YAGNI. OTOH, in the Gui, I think it will be reconstituted.

rprouse:
Good point on the status. I have changed it to use the label as you suggested. I searched through the code to see if Failed:Error is ever used on an assembly, but cannot find any, but my search foo may be off :smile:

CharliePoole:
I don't think so, because Failed:Error really means an unexpected exception and you can't get one on the assembly except right in the code we are looking at.

Issue #988, opened 2015-11-05
Don't run portable tests from NUnit Console


rprouse:
Spawned from #948

We don't want tons of issues reported when users try to run tests compiled against the portable library and it fails to run. Better to have the driver look at the nunit.framework.dll and determine if it is portable before running it. If it is portable, give a message that portable tests must be run with platform specific runners or nunitlite.

rprouse:
@CharliePoole, I was going to add the following to the framework assemblies, then check in the drivers;

```C#
#if PORTABLE
[assembly: AssemblyMetadata("PCL", "True")]
#else
[assembly: AssemblyMetadata("PCL", "False")]
#endif
```

I am running into Mono.Cecil problems though. In `DriverService.GetDriver`, we get the assembly references, but they are lightweight `AssemblyNameReference` objects. From there, I can load the referenced assembly in Cecil, then get the assembly attributes as a collection of `Mono.Cecil.CustomAttribute`.

The problem is that the API assembly does not reference Mono.Cecil, so I cannot pass in the custom attributes to make a decision based on them. I want it to be as generic as possible, so I want to pass in as much info as possible.

Possible solutions,

1. Wrap the custom attributes in a new class in the API assembly.
2. Add a reference to Mono.Cecil in the API assembly.

Both seem a bit heavy handed for an RC change. What do you think? Is there an easier way we could identify PCL assemblies that I am not thinking of?

CharliePoole:
I'm a little shaky on why you need to have  the attributes defined in the API (which I agree seems a bit heavy-handed). Can't the driver service make the decision by itself? Or are you trying to think ahead for when we have a driver that will handle the PCL?

I see two (I think) simple ways to go:
1. If the service finds this attribute it simply returns the null driver (I forget it's name now) with a tailored message.
2. Go ahead and call the NUnit3 driver, which then finds the attribute using normal reflection and returns an error from it's Load entry point.

Am I missing something?

BTW, I'm not looking at the code right now - I can look a bit later.

rprouse:
The attribute is defined on the framework to tag the framework as portable or not. Ideally, I would like to pass as much information as possible to each driver factory to allow them to decide if they can execute tests.

CharliePoole:
Right. But the driver factory has access to the entire framework assembly. it can look at the attribute itself. Maybe I'm missing something though. Can you write a code snippet that shows what you're thinking of doing?

rprouse:
[...]

PR #1002, opened 2015-11-07
Preparing the RC2 release

rprouse:
**Do not merge**

Starting to make the updates for RC2. I dated it for tomorrow (Sunday), but if we get the last outstanding issue done today we can change the date.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,15 +6,16 @@ Engine
  
 Issues Resolved
 
- * 970 Define PARALLEL in CF build of nunitlite
- * 978 It should be possible to determine version of NUnit using nunit console tool
- * 983 Inconsistent return codes depending on ProcessModel
- * 986 Update docs for parallel execution
- * 988 Don't run portable tests from NUnit Console
- * 990 V2 driver is passing invalid filter elements to NUnit
- * 991 Mono.Options should not be exposed to public directly
- * 993 Give error message when a regex filter is used with NUnit V2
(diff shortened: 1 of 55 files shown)


CharliePoole:
Release notes are here: https://github.com/nunit/nunit/wiki/Release-Notes-for-RC2
When we are ready they can be cut and pasted to the top of the actual release notes and the temp page can be deleted.

I'd like to try something. Let's make changes on this branch, rather than master. If we fix any more issues, do the merge with this branch as a target. We can actually release from this branch, and then merge it into master afterward. That's how I'm hoping to do it in the future, so I'd like to give it a try. (@rprouse if you think there is a risk, we can experiment on a dummy repository later on.)

As a further test, if we merge the change deleting TeamCityEventHandler directly to master first, we can subsequently learn something: whether pulling the tagged release includes or excludes that change. Hopefully it will be excluded from the tag and we can use this approach for our further 3.0 releases.

rprouse:
Charlie, sounds good and I think it will work fine.

rprouse:
I think we are ready to release based on this branch. I merged in the NUnit console issue, but not the teamcity issue. I checked the history and it looks correct even though there are changes showing in this PR for issues that are already merged into master. I think it is because they got pulled in again by #988?

I am heading out soon. I will do the release in the morning.

CharliePoole:
I closed one more documentation issue and updated CHANGES and Release Notes. PR #1006 is still pending.

PR #1003, opened 2015-11-07
Delete TeamCityEventHandler as it is not used

linquize:
Delete `TeamCityEventHandler` as it is not used


CharliePoole:
Looks good to me. @rprouse Merge if you agree.

rprouse:
Looks good to me too. We are preparing the RC release so I will merge this to master, but we won't be releasing this right now.

PR #1005, opened 2015-11-07
Don't run portable tests from NUnit Console

### LINKED ISSUE #988: Don't run portable tests from NUnit Console

rprouse:
Fixes #988

I went with your suggestion @CharliePoole and catch the serialization error. If we add a driver that supports portable, we can remove it then. It is a bit hacky, but it does give a very clear error message that should prevent bugs being erroneously reported.

I also put in the AssemblyMetaData attribute on portable. It isn't used now, but my thinking is that it is how we will mark the assembly as portable so when/if we do add the ability to run portable to the engine or any of our other runners, they will be able to find this right back to the initial 3.0 release.

Feel free to merge this into master. I will merge into the release branch.

### DIFF
--- a/src/CommonAssemblyInfo.cs
+++ b/src/CommonAssemblyInfo.cs
@@ -33,8 +33,6 @@ using System.Reflection;
 
 #if PORTABLE
 [assembly: AssemblyMetadata("PCL", "True")]
-#else
-[assembly: AssemblyMetadata("PCL", "False")]
 #endif
 
 #if DEBUG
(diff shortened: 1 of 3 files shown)


CharliePoole:
I've been playing with this and I don't think there's any better solution. I also changed the interface but I'll merge yours and then push my additional changes separately.

PR #1006, opened 2015-11-07
Issue 988b

CharliePoole:
Some added changes I had done when looking at 988. If they look OK, merge them in - otherwise we can wait.

This doesn't really add any new capability, but it does give the driver factory access to a reference to the framework.

PR #1007, opened 2015-11-08
Modify IDriverFactory Interface

rprouse:
I have moved @CharliePoole's commit in #1006 into a new branch off the RC branch and created this PR from it. I will merge once I have reviewed a second time and all CI builds pass.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Drivers/NUnit2DriverFactory.cs
+++ b/src/NUnitEngine/nunit.engine/Drivers/NUnit2DriverFactory.cs
@@ -61,7 +61,7 @@ namespace NUnit.Engine.Drivers
         public IFrameworkDriver GetDriver(AppDomain domain, AssemblyName reference)
         {
             if (!IsSupportedTestFramework(reference))
-                throw new ArgumentException("Invalid framework name", "frameworkAssemblyName");
+                throw new ArgumentException("Invalid framework", "reference");
 
             return _driverNode.CreateExtensionObject(domain) as IFrameworkDriver;
         }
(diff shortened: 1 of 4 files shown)


src/NUnitEngine/nunit.engine/Drivers/NUnit2DriverFactory.cs
```diff
@@ -57,9 +58,9 @@ public bool IsSupportedTestFramework(string assemblyName, Version version)
         /// <param name="assemblyName">The name of the test framework reference</param>
         /// <param name="version">The version of the test framework reference</param>
         /// <returns></returns>
-        public IFrameworkDriver GetDriver(AppDomain domain, string assemblyName, Version version)
+        public IFrameworkDriver GetDriver(AppDomain domain, AssemblyName reference)
         {
-            if (!IsSupportedTestFramework(assemblyName, version))
+            if (!IsSupportedTestFramework(reference))
                 throw new ArgumentException("Invalid framework name", "frameworkAssemblyName");
```
rprouse: Minor point since it was wrong before your changes, but the ArgumentException is using the wrong parameter name. I have this new branch local, so I will fix it. I wish we could start using nameof() here. I think it would work everywhere except CF.

CharliePoole: It would force use of c# 6 though. Up to now we've been using 3.

rprouse: Probably a moot point, but if we have been forced into using VS2015 and the newer mono compilers for portable, then most of those C# 6 features work on all frameworks we target except for CF since they are just syntactic sugar. This one would be helpful because it gives us type safety. I know I have fixed a few misnamed arguments in our guard statements :smile:

PR #1019, opened 2015-11-12
Issue 1015

### LINKED ISSUE #1015: Specifying .nunit project and --framework on command line causes nunit3-console.exe to crash.
The command line that is run:
```ShellSession
C:\Program Files (x86)\NUnit.org\nunit-console\nunit3-console.exe" "I:\OpenSource\GitHub\nunit\EngineAndConsoleTests.nunit" --trace=verbose --framework=v3.0
```
The output that is received. 

```ShellSession
NUnit Console Runner 3.0.5790

Runtime Environment
  CLR Version: 4.0.30319.42000

Test Files
    I:\OpenSource\GitHub\nunit\EngineAndConsoleTests.nunit

Could not load file or assembly 'nunit.framework' or one of its dependencies. The system cannot find the file specified.
```
Running the command line with the test library directly and the framework specified gives. 
```ShellSession
C:\Program Files (x86)\NUnit.org\nunit-console\nunit3-console.exe" "I:\OpenSource\GitHub\nunit\bin\Debug\nunit.engine.tests.dll" --trace=verbose --framework=v3.0
```

```ShellSession
NUnit Console Runner 3.0.5790

Runtime Environment
  CLR Version: 4.0.30319.42000

Test Files
    I:\OpenSource\GitHub\nunit\bin\Debug\nunit.engine.tests.dll

=> NUnit.Engine.Tests.RuntimeFrameworkTests.CanListAvailableFrameworks
Available: Net 2.0.50727
Available: Net 3.0
Available: Net 3.5
Available: Net 4.0 - Full
Available: Net 4.5 - Full
Available: Net 4.0 - Client
Available: Net 4.5 - Client
Available: Net 4.0 - Client

Errors and Failures
...
```
I am not certain if this is by design, and NUnit is simply missing a guard condition to not allow .nunit projects to not be specified with the framework option.

Should the framework command be allowed to be specified with a nunit project, or is this an invalid test of the feature? I also have a sense that the error being reported is a path issue with the tests files themselves.

CharliePoole:
This fixes #1015. It's a bigger change than I really like to do between two RCs but it's an important issue.

I reworked the order of setting up the domain to make it easier to test and added some tests.

I also implemented a basic policy change as compared to NUnit V2: when running a project file named project.nunit (or any other project type) the config file project.nunit is only set up for the AppDomain if it exists, otherwise, we use the individual assembly configs.

This is not exactly a "breaking change" - it actually fixes something that was broken before - but it is a change in behavior.

Note that when multiple assemblies are made to run in the same AppDomain, there is no way for NUnit to give them separate config files.

### DIFF
--- a/src/NUnitEngine/nunit.engine.api/TestEngineActivator.cs
+++ b/src/NUnitEngine/nunit.engine.api/TestEngineActivator.cs
@@ -100,6 +100,22 @@ namespace NUnit.Engine
             // Check the Application BaseDirectory
             string path = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, DefaultAssemblyName);
             Assembly newestAssemblyFound = CheckPathForEngine(path, minVersion, ref newestVersionFound, null);
+
+            // Check Probing Path is not found in Base Directory. This
+            // allows the console or other runner to be executed with
+            // a different application base and still function. In
+            // particular, we do this in some tests of NUnit.
+            if (newestAssemblyFound == null)
+            {
+                foreach (string relpath in AppDomain.CurrentDomain.RelativeSearchPath.Split(new char[] { ';' }))
+                {
+                    path = Path.Combine(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, relpath), DefaultAssemblyName);
+                    newestAssemblyFound = CheckPathForEngine(path, minVersion, ref newestVersionFound, null);
+                    if (newestAssemblyFound != null)
+                        break;
+                }
+            }
+
             if (!privateCopy)
             {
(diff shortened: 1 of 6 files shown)


rprouse:
I've reviewed and everything looks good as far as I can tell with this sort of change. I am going to merge. I don't have time, but we should run through a few testing scenarios before RC 3 is released. I'm sure you are already thinking that though :smile:

CharliePoole:
Any particular scenarios in mind? Of course, I ran a few before pushing this, but more would be better.

OTOH, I'm confident that all the problems we were seeing had to do with the appbase and probing path not being set correctly. Based on the unit tests, it looks like we have that corrected now.

PR #1022, opened 2015-11-13
Support Comparing Unlike Collections using Is.EquivalentTo().Using(...)

crmckenzie:
I'm normally able to do this through the use of extension methods, but I ran into an insurmountable problem if one of the collection types is strings. 

The implementation of `EqualityAdapter.CanCompare` is opinionated in a way that is unhelpful in this specific context.

Here is the test I want to support:

```csharp
[Test]
public void EquivalentHonorsUsingWhenCollectionsAreOfDifferentTypes()
{
    ICollection set1 = new SimpleObjectCollection("1", "2", "3");
    ICollection set2 = new SimpleObjectCollection(1, 2, 3);

    Assert.That(set2, Is.EquivalentTo(set1).Using<int, string>((i, s) => i.ToString() == s));
}
```

The real use-case here is that a lot of test code I write is to test mappings between different representations of the same data (Think verifying AutoMapper behavior). Since the current `Is.EquivalentTo` implementation assumes unlike types, I created a new method that bypasses the current safety check and assumes the dev knows what s/he is doing.

I did not intend any changes to the `DotSettings` file - please ignore those.


CharliePoole:
Since we generally merge using the online GitHub facility, it would be simplest if you would use git reset to remove your last commit and then do a forced push to the same branch. Hopefully, github will update the PR accordingly. :-)

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
```diff
@@ -72,6 +72,20 @@ public CollectionItemsEqualConstraint Using(IComparer comparer)
             return this;
         }
 
+#if (NET_4_0 || NET_4_5) && !PORTABLE
```
CharliePoole: Why not portable?

src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintTests.cs
```diff
@@ -152,8 +152,19 @@ public void EquivalentHonorsUsing()
...
-#if ( NET_4_0 || NET_4_5 ) && !PORTABLE
+        [Test]
+        public void EquivalentHonorsUsingWhenCollectionsAreOfDifferentTypes()
+        {
+            ICollection set1 = new SimpleObjectCollection("1", "2", "3");
+            ICollection set2 = new SimpleObjectCollection(1, 2, 3);
+
+            Assert.That(set2, Is.EquivalentTo(set1).Using<int, string>((i, s) => i.ToString() == s));
+        }
+#endif
```
CharliePoole: You changed a base class and thereby affected CollectionEquivalentConstraint, CollectionContainsConstraint, SubsetConstraint, SupersetConstraint and UniqueItemsConstraint. We should have tests for all of those.

CharliePoole:
I like the idea of a new argument type for Using. It seems sensible to allow its use in EqualConstraint as well. We could also support it in the .NET 2.0 build if we simply defined a compatible delegate.

crmckenzie:
This is all good feedback and I'll work on it. It will likely be next week before I finish it.
Thanks!

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
crmckenzie: I wasn't sure if the portable framework had the Func delegate. In truth, I'm not sure how to test the code locally to find out. Do you have instructions for testing locally against the various framework versions? I haven't found this information in the documentation yet.

CharliePoole: Well, if you include it and the portable build compiles, then it works. (It does, btw.) In fact in VS2015, you only have to type it and you get a popup showing which targets support it.

For testing, just run the tests for that particular build, which are in the individual runtime folders. For .NET 2.0, 4.0 and 4.5 use the console runner. For Silverlight use the test app itself. For Portable, you'll find a console app that links to the portable tests.

CharliePoole: It's also supported for compact framework 3.5, so it's only **not** supported by .NET 2.0.

src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintTests.cs
crmckenzie: I don't see this logic will make sense for UniqueItemsContraint. I'm probably going to just implement it in the other three.

CharliePoole: Well, you have already implemented it in all those classes. We just don't have tests. :-)

I agree that the particular int to string comparison would not be so useful to UniqueItemsConstraint, but being able to override how equality comparison is done is useful and it already supports all the other overloads of Using. The thing is, I don't want users worrying about which overloads work for which constraints - they should all work.

crmckenzie: "I don't want users worrying about which overloads work for which constraints"
I agree.

What I'm proposing is to remove the implementation of Unique<TExpected, TActual>(lambda) out of the base class to three of subclasses. In this way it would only be available on the constraints where it makes sense.

`Is.Unique` is slightly different in what it's testing than the other constraint. I think it needs a generic method with the same _signature_, but the meaning of the arguments is different. Consider:

```csharp
        internal class Widget
        {
            public int Id { get; set; }
            public string Name { get; set; }
        }

        [Test]
        public void UniqueHonorsUsingByProperty()
        {

            ICollection set1 = new SimpleObjectCollection(
                new Widget()
            {
                Id = 1,
                Name = "Widget 1",
            }, new Widget()
            {
                Id = 2,
                Name= "Widget 2"
            }, new Widget()
            {
                Id = 3,
                Name = "Widget 3"
            });

            Assert.That(set1, Is.Unique(set1).Using<Widget, int>(w => w.Id));
        }
```

This would read "Assert that widgets are unique by their ids." 

There's an alternative way to implement this for Unique that only requires a single generic argument to Using:

```csharp
        [Test]
        public void UniqueHonorsUsingWithComparison()
        {

            ICollection set1 = new SimpleObjectCollection(
                new Widget()
                {
                    Id = 1,
                    Name = "Widget 1",
                }, new Widget()
                {
                    Id = 2,
                    Name = "Widget 2"
                }, new Widget()
                {
                    Id = 3,
                    Name = "Widget 3"
                });

            Assert.That(set1, Is.Unique(set1).Using<Widget>((left, right) => left.Id == right.Id));
        }
```

This latter signature could be supported in all 4 constraints with the same implementation and could be in the base class.

CharliePoole: OK, I'm convinced. :-)

crmckenzie:
I'm hoping that this is close. I'm expecting you to ask me to consolidate the 2 commits before you accept this PR. Can I get an approval status update?
Thanks!

CharliePoole:
Sorry that this has  been sitting for so long. It's a combination of other work and holidays!

To inform my detailed review, please explain a few oddities (at least to me) in **what** you are doing as opposed to how it's done...

1. You seem to allow this  comparison for subset, superset and equivalence of collections but not for straight equality. Is there a reason?

2. You only allow it for collections, not for simple values. Same question.  If we did allow for simple values, then all the rest could happen by itself.

Charlie

crmckenzie:
Hi Charlie,
I've been on vacation for the holidays too, so I'm just getting to this. My goal is to allow comparison of two collections of unlike types using a lambda expression `Func<T1, T2, bool>`. 

Since you say that if I implemented 2) that the rest could happen by itself, I would defer to you to indicate how I might go about implementing 2) in such a way. I'm happy to do the work, but I'm not clear on what you're suggesting. I'm excited about contributing to NUnit and about this functionality in particular so I'll do whatever is necessary to get this PR accepted.

Also, we agreed earlier that this functionality did not make sense for the Distinct operation as there are never two unlike types in a typed collection. We must be cognizant of this when deciding at what level to create the functionality.

Thanks!
Chris

rprouse:
@crmckenzie, what is the status of this PR? Are you waiting for a review?

crmckenzie:
Yes, I am waiting for review. We've had some back and forth on the implementation, but my current understanding is what I've done is the best implementation because my addition doesn't make sense for the Unique() method.

Please let me know if there is more I need to do. I'm excited to get this feature into NUnit!

rprouse:
Good to know, thanks. @CharliePoole if you get a chance to review today, let's get this done. I will try to give a second review tonight or tomorrow morning.

CharliePoole:
It looks ready to merge AFAIC. I'm afraid I lost track of it. :-(

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
```diff
@@ -116,12 +116,18 @@ public CollectionItemsEqualConstraint Using<T>(IEqualityComparer<T> comparer)
...
-        /// <summary>
-        /// Compares two collection members for equality
-        /// </summary>
-        protected bool ItemsEqual(object x, object y)
+#endregion
+
+    /// <summary>
+    /// Compares two collection members for equality
+    /// </summary>
+    protected bool ItemsEqual(object x, object y)
```
rprouse: Minor formatting issue, this lost an indent level and the #endregion should be indented.

src/NUnitFramework/framework/Constraints/EqualityAdapter.cs
```diff
@@ -24,11 +24,18 @@
...
-using System.Reflection;
 using NUnit.Framework.Compatibility;
+using System.Reflection;
 
 namespace NUnit.Framework.Constraints
 {
+
+#if NET_2_0
+
+    public delegate TResult Func<in T1, in T2, out TResult>(T1 arg1, T2 arg2);
```
rprouse: I was wondering how Func was working in .NET 2.0. This is a good idea, but I would like to do more of this. This should these be put in in the `NUnit.Compatibility` namespace so people know where it comes from and where to find it. Maybe add a `Net20Compatibility.cs` to `nunit/src/NUnitFramework/framework/Compatibility/` and add this there? You could also pull the `ExtentsionAttribute` out of `nunit/src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs` since it doesn't really belong there.

I would also be tempted to put `Func` in the same namespace as in the .NET Framework.

src/NUnitFramework/framework/Constraints/EqualityAdapter.cs
```diff
@@ -84,9 +91,9 @@ public override bool AreEqual(object x, object y)
             }
         }
 
-        #endregion
+#endregion
 
-        #region Nested IEqualityComparer Adapter
+#region Nested IEqualityComparer Adapter
```
rprouse: Formatting again, we indent regions. VS 2015 keeps doing this to me too when I add #if statements. I really need to figure out how to disable that :frowning:

src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintTests.cs
```diff
@@ -153,8 +153,17 @@ public void EquivalentHonorsUsing()
...
+        public void EquivalentHonorsUsingWhenCollectionsAreOfDifferentTypes()
+        {
+            ICollection strings = new SimpleObjectCollection("1", "2", "3");
+            ICollection ints = new SimpleObjectCollection(1, 2, 3);
+
+            Assert.That(ints, Is.EquivalentTo(strings).Using<int, string>((i, s) => i.ToString() == s));
+        }
+
+#if (NET_4_0 || NET_4_5) && !PORTABLE
+    [Test, Platform("Net-3.5,Mono-3.5,Net-4.0,Mono-4.0,Net-4.5,Mono-4.5")]
```
rprouse: Attribute should be indented.

rprouse:
Looks good except for some formatting issues and the requested refactor of `Func` out to the compatibility namespace.

src/NUnitFramework/framework/Constraints/EqualityAdapter.cs
crmckenzie: There are other unindented regions in this file. I'm assuming you'd like me to fix those as well.

crmckenzie:
Would you prefer a 2nd commit or a push --force?
My experience with push --force is that it's hard to recall the conversation as those commits no longer exist. On the other hand, there's less churn to the code.

crmckenzie:
Followup question: do you have a script to test NUnit across each framework it's built for?

rprouse:
- Indenting the other regions that aren't, yes
- Feel free to add additional commits. We only rollup commits if a PR get out of hand :smile: 

Build and test on the command line uses Cake Build. I use the following command in PowerShell from the root of the solution to build and test everything in release;

```
.\build.ps1 -Target TestAll
```

If you are using a DOS command prompt, you might need to switch to build.cmd depending on your setup.

crmckenzie:
It looks like I'll need to update my PR to the latest version first. I'm going to commit my changes, update my PR branch, then test.

crmckenzie:
@rprouse I think this is done! :)

rprouse:
Looks good. A bit of minor cleanup that I want to do to the compatibility namespace, but it is probably easier if I do it that going back and forth. Thanks for the contribution. I'm sorry we dropped the ball for awhile, but if you want to do more, we can always use the help.

crmckenzie:
I'd be happy to do more! If you've got something in mind let me know and I"ll take a look.

rprouse:
@crmckenzie, we can always use the help. If you still want to get your feet wet, you can pull issues from the [EasyFix label](https://github.com/nunit/nunit/issues?q=is%3Aopen+is%3Aissue+label%3Aeasyfix) and everything in the [Backlog milestone](https://github.com/nunit/nunit/issues?q=is%3Aopen+is%3Aissue+milestone%3ABacklog) is up for grabs if there is nobody assigned.

Issue #1011, opened 2015-11-09
Packaging the Engine


CharliePoole:
It would help us with our runners, as well as authors of third-party runners, to have the engine available as a NuGet package.

The adapter needed the bare engine, with no addins. Other folks might want that or they might want specific addins. Some design is needed.

For now, we may just want to create a bare engine package, consisting of nunit.engine, nunit.engine.api and Mono.Cecil.

rprouse:
Should we repurpose the Nunit Core Engine and package the engine, api and cecil in that?

https://www.nuget.org/packages/NUnit.Core.Engine/3.0.0-beta-4

We can then release a Nunit Full Engine package if there is demand.

CharliePoole:
Changed this to milestone 3.2 so it will show up on the release notes for that release. However, it is still being released out of band as version 3.0.

PR #1027, opened 2015-11-16
Add engine nuget package

CharliePoole:
This adds a new package for the engine. Since the two agents are relatively small, they are included as well. Package is being released using the 3.0 bits.


CharliePoole:
This fixes #1011

PR #1032, opened 2015-11-17
Create new nunit.runners package. Fixes #1030.

### LINKED ISSUE #1030: Should we update Nunit.Runners package to 3.0?
Several people have been complaining on twitter because they have updated their projects to use the NUnit 3 framework, but there is no corresponding update for NUnit.Runners. They do not realize that they need to switch to the new NUnit.Console NuGet package.

Even when you search NuGet for NUnit, the new NUnit.Console package is the 43'rd entry and on the third page. Most people won't find it. NUnit.Runners is the 2nd entry right after NUnit, so it is what everyone sees.

Even if we put the information out on as many channels as we control (mailing list, website, twitter, etc) it will only reach a small fraction of developers who use NUnit.

Should we create an update for the NUnit.Runners nuget package that contains the same binaries as the current NUnit.Console package? Once we have additional runners like the GUI, we can include them in that package and leave NUnit.Console as the lightweight, console only package.

I think we should deal with this soon. NUnit 3 has only been showing up as a non-prerelease update for a couple of days. I think many people are going to be upgrading soon and running into the same problem.

- https://www.nuget.org/packages/NUnit.Console/
- https://www.nuget.org/packages/NUnit.Runners/

CharliePoole:
I created the new package as a simple reference to nunit.console. I'll push that to nuget and add it to the 3.0 release on nuget.

I also made some updates to the text in nunit.console. No need to update the package on nuget for these, because I edited it manually when I pushed.

The new package is currently only produced when you use the target explicitly. I'll add it to the list of items produced in the new script I'm working on.

IOW, I tried to change as little as possible in this issue.

### DIFF
--- a/NUnit.proj
+++ b/NUnit.proj
@@ -467,6 +467,17 @@
 
   </Target>
 
+  <Target Name="PackageNuGetRunners" Label="Creates the NUnit.Runners NuGet package"
+      DependsOnTargets="_CreateImageIfNotPresent">
+    <Message Text="******************************************************************" />
+    <Message Text="* Creating the NUnit.Runners Nuget $(Configuration) package" />
+    <Message Text="******************************************************************" />
+
+    <Exec WorkingDirectory="$(ProjectBaseDir)"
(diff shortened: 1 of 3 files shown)


rprouse:
Looks good. If you have tested the NuGet package by building then installing it into a test solution and the binaries in your package directory look good, then go ahead and merge/ship :+1:

CharliePoole:
Yes, I did test. It creates an empty package directory for itself and also adds the nunit.console package. FYI, if you leave out the empty files element, it includes everything it can find!

PR #1050, opened 2015-11-20
Do not expose System.Runtime.CompilerServices.ExtensionAttribute to public

linquize:
Do not expose `System.Runtime.CompilerServices.ExtensionAttribute` to public.
It is only used internally for convenience of coding.
If the user already references `System.Core`, a compiler warning will be generated about duplicate attribute defined.

### DIFF
--- a/src/Common/nunit/XmlHelper.cs
+++ b/src/Common/nunit/XmlHelper.cs
@@ -30,7 +30,7 @@ using System.Xml;
 namespace System.Runtime.CompilerServices
 {
     [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method)]
-    public sealed class ExtensionAttribute : Attribute { }
+    sealed class ExtensionAttribute : Attribute { }
 }
 #endif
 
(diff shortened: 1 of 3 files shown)


rprouse:
Makes sense and looks good. We've been planning on making a number of classes internal now that we are using InternalsVisibleTo for the tests, but never got around to it.

rprouse:
CI is passing. I am going to merge without a second review, it is a very small change.

CharliePoole:
There are known problems with making the attribute internal. We ran into this with Mono.Cecil in the VS Adapter build and were only able to resolve it by removing the reference. That was OK for us, because the adapter doesn't use Mono.Cecil directly but it obviously can't work for tests that reference the framework.

What happens is that the compiler, in some situations, decides to use the internal attribute rather than the public one. Then it finds it can't access the constructor and issues a build error rather than the warning caused by the duplicate public attributes. From what I have found on SO, it doesn't seem possible to predict when this will happen and there are no easy user workarounds.

Anyway, that's why we haven't done this before.

The warning should only arise for users building tests for .NET 3.5.

This article has most of the information and several good links:
http://stackoverflow.com/questions/11025100/escape-catch-22-with-extension-attributes-in-net-2-0

PR #1053, opened 2015-11-20
Revert "Do not expose System.Runtime.CompilerServices.ExtensionAttribute to public"

rprouse:
Reverts nunit/nunit#1050

### DIFF
--- a/src/Common/nunit/XmlHelper.cs
+++ b/src/Common/nunit/XmlHelper.cs
@@ -30,7 +30,7 @@ using System.Xml;
 namespace System.Runtime.CompilerServices
 {
     [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method)]
-    sealed class ExtensionAttribute : Attribute { }
+    public sealed class ExtensionAttribute : Attribute { }
 }
 #endif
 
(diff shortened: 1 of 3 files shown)


rprouse:
@CharliePoole I am reverting the internalizing. I assume you think we should?

CharliePoole:
That was my first thought but I was waiting to see what you guys think. :-)

CharliePoole:
@rprouse Apparently this PR is still open - the revert needed to be merged but was not.

rprouse:
I'm not convinced we should revert this. I read the SO post linked in #1050 and it states that leaving it public "will cause a compiler error from a VB.NET project targeting .NET 3.5 or higher." They then go on to state that some users got a compiler error when it was internal.

I checked my standby library for seeing best practices for targeting multiple frameworks, Newtonsoft.Json and they make it internal,

https://github.com/JamesNK/Newtonsoft.Json/blob/2eed44a2f850e3ab5ab97a6d22ef7e5735b48b06/Src/Newtonsoft.Json/Utilities/LinqBridge.cs

BTW, I like their LinqBridge. We should consider pulling that in to enable Linq style queries. Total aside :smile: 

Reading that SO article gives me the impression that the problems arise when people target the 2.0 version of your library from 3.5+. Now that NUnit is compiled for multiple targets, that is only a problem for people targeting 3.5. Personally, the always fails for VB 3.5 if it is public compared to the "we can't figure it out" random failures of 3.5 when it is internal is a no-brainer. Not ideal, but ideal would have us re-instate the 3.5 version of the framework.

Am I missing anything in my assessment?

CharliePoole:
I agree let's drop the revert. I'm thinking we need to put this in a
separate assembly that users don't reference along with other compatibility
code used in our 2.0 build. I have Linq code from mono that I used to make
mono.addins run on 2.0 as well.

PR #932, opened 2015-10-26
add all AppVeyor builds to its NuGet feed

ctaggart:
As I mentioned at https://github.com/nunit/nunit/pull/915#issuecomment-150994444, here is a PR that makes all builds available via the AppVeyor NuGet feed. See 
http://blog.ctaggart.com/2015/07/open-source-net-with-appveyor-nuget.html

@CharliePoole just needs to go to 
https://ci.appveyor.com/project/CharliePoole/nunit/settings/nuget

and change it to something like:
https://ci.appveyor.com/nuget/nunit

The script support creating a build from a tag. If you pushed a `3.0.0-beta-6` tag, it will correctly create the packages in the AppVeyor feed. Let me know what questions you have.

PR #1054, opened 2015-11-20
Create nuget feeds for CI builds on Appveyor

CharliePoole:
I have taken @ctaggart 's PR #932, rebased it on top of the current master and published it to this PR so I can make a few changes.

Even though we want to use something like Cake for our build script, this powershell script can provide an interim solution. The important thing is that we establish a solid naming convention for our builds, which we can continue to use, and that we make sure we are only publishing the artifacts for those builds where we really want them.

### DIFF
--- a/NUnit.proj
+++ b/NUnit.proj
@@ -7,9 +7,9 @@
   <PropertyGroup Label="Common Properties">
     <ProjectName>$(MSBuildProjectName)</ProjectName>
     
-    <PackageVersion Condition="'$(PackageVersion)' == ''">3.0.0</PackageVersion>
+    <PackageVersion Condition="'$(PackageVersion)' == ''">3.1.0</PackageVersion>
     <PackageModifier Condition="'$(PackageModifier)' == ''"></PackageModifier>
-    <DisplayVersion Condition="'$(DisplayVersion)' == ''">3.0</DisplayVersion>
+    <DisplayVersion Condition="'$(DisplayVersion)' == ''">3.1</DisplayVersion>
     
     <PackageName>$(ProjectName)-$(PackageVersion)$(PackageModifier)</PackageName>
     <PackageNameCF>$(ProjectName)CF-$(PackageVersion)$(PackageModifier)</PackageNameCF>
--- a/appveyor.ps1
+++ b/appveyor.ps1
@@ -1,5 +1,5 @@
-# the version under development, update after a release
-$version = '3.0.0'
+# the default version under development, update after a release
+$version = '3.1.0'
 $modifier = '-ci'
 
 function isVersion($s){
@@ -9,7 +9,11 @@ function isVersion($s){
 
 # append the AppVeyor build number as the pre-release version
 if ($env:appveyor){
-    $modifier = $modifier + '-' + [int]::Parse($env:appveyor_build_number).ToString('000')
+	# force build number to four digits for correct ordering
+	$build_number = [int]::Parse($env:appveyor_build_number).ToString('0000');
+    $modifier = "-" + $build_number + $modifier;
+
+	# if there is a tag, it provides both version and modifier
     if ($env:appveyor_repo_tag -eq 'true'){
         $tag = $env:appveyor_repo_tag_name
         $i = $tag.IndexOf('-')
@@ -22,6 +26,7 @@ if ($env:appveyor){
             $modifier = ''
         }
     }
+
     if(-not(isVersion($version)))
     {
         Write-Error "error parsing version '$version' in tag '$tag'"
@@ -32,4 +37,5 @@ if ($env:appveyor){
 
 ./build.cmd NUnit.proj /t:BuildAll /p:Configuration=Release /p:PackageVersion="$version" /p:PackageModifier="$modifier" /v:m
 ./build.cmd NUnit.proj /t:TestAll /p:Configuration=Release /p:PackageVersion="$version" /p:PackageModifier="$modifier" /v:m
-./build.cmd NUnit.proj /t:Package /p:Configuration=Release /p:PackageVersion="$version" /p:PackageModifier="$modifier" /v:m
+./build.cmd NUnit.proj /t:PackageNuget /p:Configuration=Release /p:PackageVersion="$version" /p:PackageModifier="$modifier" /v:m
+./build.cmd NUnit.proj /t:PackageNugetNUnitSL /p:Configuration=Release /p:PackageVersion="$version" /p:PackageModifier="$modifier" /v:m
--- a/appveyor.ps1
+++ b/appveyor.ps1
@@ -26,17 +26,9 @@ if ($env:appveyor){
 		$build_number = [int]::Parse($env:appveyor_build_number).ToString('0000');
 		$modifier = $modifier + '-' + $build_number;
 
-		# add branch if not master
-		$branch = $env:appveyor_repo_branch;
-		if($branch -ne 'master')
+		if($env:appveyor_pull_request_number -ne '')
 		{
-			$modifier = $modifier + '-' + $branch;
-		}
-
-		$pr_number = $env:appveyor_pull_request_number;
-		if($pr_number -ne '')
-		{
-			$modifier = $modifier + '-pr' + $pr_number;
+			$modifier = $modifier + '-pr';
 		}
 	}
 
--- a/appveyor.ps1
+++ b/appveyor.ps1
@@ -26,7 +26,7 @@ if ($env:appveyor){
 		$build_number = [int]::Parse($env:appveyor_build_number).ToString('0000');
 		$modifier = $modifier + '-' + $build_number;
 
-		if($env:appveyor_pull_request_number -ne '')
+		if($env:appveyor_pull_request_number)
 		{
 			$modifier = $modifier + '-pr';
 		}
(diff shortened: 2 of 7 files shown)


CharliePoole:
@rprouse @nunit/core-team 

Points to finalize before we merge this.
1. Do we like the naming pattern (3.0.0-ci-xxxx)
2. Do we want to produce artifacts for all branches, only some branches, only merges, or want?
3. What artifacts do we want to produce? Nuget only? Msi? Zips? What about SL and CF?
4. What version should this be? It's now 3.0.0, should it be 3.1.0 to show it's post 3.0?

You can click on the Appveyor builds and look at the artifacts being produced right now.

rprouse:
Fairly busy this weekend, will try to get back to you tomorrow.

CharliePoole:
That will be good. Thinking about it, I feel like where we are in doubt we should publish less rather than more. We can always increase it later.

If we had some good ci build's out there with nuget packages, then you could have people use those in testing the runner.

rprouse:
For the naming pattern. Eventually, I would like the 3.0.0 part pulled from a single place and patched everywhere. I like how the Cake team updates their release notes right after release with a placeholder for the new version and that drives the version number everywhere.

I would also like to see it incremented to whatever our next release is right after the release. We should have it set now to 3.1.0. If we want to do a hotfix, we branch off the 3.0.0 release and set it to 3.0.1.

My only concern with the `ci-xxxx` moniker is that it prevents us from releasing betas to NuGet because 3.1.0-beta would come before for people that have the AppVeyor and NuGet.org feeds configured. Might (3.1.0.xxxx-ci) be better for this?

If we did it in that order, we could also use the `ci` moniker to identify the branch. `ci` would builds of master and `pr` would be for pull requests indicating to people that they are even less reliable. Having the build number as the third part would keep them in chronological order.

I would like to produce artifacts for all of our branches, pull requests and obviously master. 

Only core team can create branches, so they are safe and would allow us to easily test our changes in other projects without referencing assemblies in the bin directory which can be a pain moving between machines.

As you pointed out, pull requests could be dangerous, but nugets would allow us to test them without pulling down and building. If we have a clear naming scheme and they are only in the AppVeyor feed, then I think we mitigate the risk.

Lastly, artifacts from master could be used by users that need a fix for a bug that is not in a hotfix.

As for artifacts that I want to produce, primarily NuGet. Msi's are problematic because the third version digit needs to increment otherwise you need to uninstall first. We could produce zips, but my priority is to speed testing and nuget does that for me. As for SL and CF, much lower priority. I don't even think we can get AppVeyor to build CF. Let's get the primary stuff working and we can add to it if there is a need.

rprouse:
Another thing I saw on another project was setting it so that PR's could not be merged unless the CI builds pass. Notice the `Required`,

image

CharliePoole:
Some of this discussion is over-shadowed by my wanting to use the odd / even convention in the future, and it's not being entirely clear what I mean by that. So...

* Our releases would all have an even minor version number: 3.0, 3.2, etc.
* Consequently we would only have "pre-release" suffixes on packages with an even minor version number: 3.0.1-alpha, 3.2.0-beta-3
* I did not envision that the odd and even packages would ever appear on the same nuget feed, so I was not worried about conflicts. I'm seeing a few cracks in that assumption, however, so let's discuss.

With that in mind, I'll address the points you raised:

 * Single point of update for the semantic version - I agree. We had that before this fix @ctaggart 's change gives us two places, one for Appveyor and one for other uses of NUnit.proj. We could fix it in this PR but I didn't want to mess with powershell due to not liking it much better than XML. If you're OK with waiting, I'll fix it in the cake script.

 * Current semantic version for packages should be 3.1.0 in the repository and assembly versions should be set to 3.1.*.*.

 * As explained above, I don't envision we would ever have a 3.1.0-beta. But that only postpones the conflict to the next release - either 3.0.1 or 3.2.0. My assumption has been that these would never appear on then Appveyor feed but would be pushed somewhere else. I think our first question should be whether we can safely postpone this decision.

* I'm good with 3.1.0-xxxx-ci but not 3.1.0.xxxx-ci. I don't think the latter fits the semver spec.

* It seems like we have three kinds of builds to deal with rather than two. Do we need some way to distinguish these?

  1. Builds of any non-master branch, such as those we create ourselves.

  2. Builds of master as part of a merge request - either for one of our branches or from a fork.

  3. Builds of master after a merge.
  
* What if somebody wants to only get packages from master, excluding merges. Do they have to do that manually?

* I'm inclined to drop all the packages from the Appveyor build except for nugets. That's on the premise that it's easier to add something rather than remove it later. I would add silverlight, since we already build it. 

Next round of comments? :-)

CharliePoole:
So how do you tell github that a ci build is required to pass?

chris-smith-zocdoc:
@CharliePoole its in the repo settings https://github.com/blog/2051-protected-branches-and-required-status-checks

CharliePoole:
Thanks @chris-smith-zocdoc that's helpful.

rprouse:
@CharliePoole, I have made master a protected branch and require CI builds to be passing before a merge. We have been having problems with Travis, so we might want to turn it off for that. Let's see how it goes.

The only problem I can think of is if we want to merge a pull request offline. We'll have to see how it works out. It is simple enough to turn off if it causes problems, but I like the idea of not being able to push to master.

Thanks for the help @chris-smith-zocdoc

CharliePoole:
@rprouse I don't feel strongly about this but it feels like solving a problem we haven't actually seen.

rprouse:
@CharliePoole, I agree and don't feel strongly about it either. I do like that we can't push to master though. We don't, but now that 3.0 is out, there is a lot more interest in the project and I can see us adding more people to the team. This just keeps everyone honest. If it is a problem, we can remove it. It is just a quick setting that you or I can change.

rprouse:
Back to your comments on this PR. I am okay with everything you said. I don't feel like anything we are doing is backing us into a corner and that we can iterate and improve as we see how it works. We profess to be agile, so we don't need to do all the design up front :smile: 

I am fine with fixing some of the things we want to change once we get a better build system. I think this PR is just one step in the right direction.

> As explained above, I don't envision we would ever have a 3.1.0-beta. But that only postpones the conflict to the next release - either 3.0.1 or 3.2.0. My assumption has been that these would never appear on then Appveyor feed but would be pushed somewhere else. I think our first question should be whether we can safely postpone this decision.

I do think that we can postpone that decision, but I think that if AppVeyor is building our branches, it will end up producing artifacts for the 3.2 release as we are preparing it. I would actually hope it does. I envision a future where we create a release branch, AppVeyor builds it producing artifacts, we download and test the result and then push the official release if all is good. Pushing the build and packaging to a CI machine ensures that all steps are followed and god forbid, prevents us from doing something stupid like introducing a virus from a developers machine.

My suggestion for dealing with different types of builds is to distinguish them somehow with the postfix identifier. I'm just not sure how we do that. I think this is another thing that we can move towards and shouldn't hold up this PR.

CharliePoole:
I thought it only prevented push -f

We can always recover from a simple push.

CharliePoole:
Results of some experiments locally:
  3.1.0-ci-1234 is OK
  3.1.0-build-1234-ci is OK
  3.1.0-1234-ci fails with message that it's not a valid version for nuget
  3.1.0-ci-1234.56 fails the same way
  3.1.0.1234-ci builds but gives a warning that it's not a proper semantic version

From this, it appears that the suffix must start with hyphen, have an alpha as the char after the hyphen and not contain any periods.

CharliePoole:
This is now working with format 3.1.0-ci-xxxx

We should finalize the format and merge.

My thoughts:

* For a non-pr build of master, keep current format: 3.1.0-ci-1234
* For a build of some other branch, add the branch: 3.1.0-ci-1234-issue-999
* For a pr add -pr: 3.1.0-ci-1234-pr

Note that tags still have precedence over all this. That is, if you tag a commit as 3.0.0-beta1, that's what will be used.

Folks would have to pick and choose which packages they want, but I think that's how this feed would be used anyway.

@rprouse Let me know what you think and I'll make the "final" changes.

CharliePoole:
Additional possibility: Use the pr number as well - 3.1.0-ci-1234-pr999

CharliePoole:
Turns out that the size of what nuget calls the "special version part" can't exceed 20 characters and the branch name can make it too big. So my new plan would be to leave it as but just add -pr for pull requests. That is:
    3.1.0-ci-1234
    3.1.0-ci-1234-pr
For any other info, folks have to check the actual appveyor build.

CharliePoole:
This is now working and ready to merge provided we like the conventions it implements.

CharliePoole:
You can see the result of our experiments with this feature by setting http://ci.appveyor.com/nuget/nunit as a package source. Frankly, it's not that clear from the list which builds one might want to use. I think the ultimate solution might be to modify the nuspec file so that the release notes indicate what is being built. However, we should get what we have in place anyway.

rprouse:
All sounds good. I will merge after I get a chance to review it.

On 22 November 2015 at 16:00, CharliePoole <notifications@github.com> wrote:

> You can see the result of our experiments with this feature by setting
> http://ci.appveyor.com/nuget/nunit as a package source. Frankly, it's not
> that clear from the list which builds one might want to use. I think the
> ultimate solution might be to modify the nuspec file so that the release
> notes indicate what is being built. However, we should get what we have in
> place anyway.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/nunit/nunit/pull/1054#issuecomment-158799262>.
>



-- 

Rob Prouse



I welcome VSRE emails. Learn more at http://vsre.info/

rprouse:
Looks good to me so far. It think that this is at least a good move in the direction we want to go in. Like you, I would like to get rid of the PowerShell script, but we can cover that in the Cake script. Let's go with this for now and improve on it as we can.

PR #1059, opened 2015-11-23
Initial Cake script

### LINKED ISSUE #930: Switch from MSBuild based build system to Cake
This is a post 3.0 task and up for discussion, but I wanted to capture it. We have talked several times in the past about using a cross platform build system other than MSBuild. I am bringing this up again because NUnit.proj is getting really hard to read or maintain.

One option that I will put out there is [Cake](http://cakebuild.net/). It is cross platform, uses a C# based DSL which is very readable and it pulls down all its dependencies on run. Just looking at what they do in [the build script for their project](https://github.com/cake-build/cake/blob/develop/build.cake) is appealing. They clean, build, test, create zip files, package NuGet packages and deploy them to MyGet and even NuGet. I also like how they automatically version everything and auto-append build version numbers for CI builds. I also love how readable the script is.

CharliePoole:
This fixes #930. I'll update this description with progress as I proceed. Interim comments would be helpful.

Progress:
* Targets for cleaning, building, testing and some packages.
* Travis target does build and test.
* Appveyor target does build, test and package.
* Packaging is complete.


cakebuild.cmd
```diff
@@ -0,0 +1,2 @@
+@echo off
+powershell ./build.ps1 %*
```
rprouse: Is this needed? Isn't it functionally equivalent to just running build.ps1?

CharliePoole: Not on my machine. That's why I wrote it.

BTW, I should have clarified in the description of the PR that cakebuild.cmd and cake.cmd are placeholders. After we remove the old NUnit.proj and build scripts, I'll rename some of these. This one will be just build.cmd and cake.cmd may not be needed.

BTW#2: if you run this on a 32-bit windows machine, you should specify -experimental to cake or -Experimental to build.ps1. If you use -mono, there is a compiler error that shows up, which I will report to the cake guys.

build.cake
```diff
@@ -0,0 +1,213 @@
...
+	StartProcess(exePath, new ProcessSettings()
+	{
+		WorkingDirectory = binDir
+	});
+}
+
+void RunTestWithConsoleRunner(string testAssembly)
+{
+	StartProcess(NUNIT3_CONSOLE, new ProcessSettings()
+	{
```
CharliePoole: This is where an error arises using -mono. NUNIT3_CONSOLE and other global vars are not known to the compiler at this point when using the mono compiler.

rprouse:
I merged latest from master into this PR to keep it up to date.

CharliePoole:
Comparison of targets in NUnit.proj and cake.build

NUnit.proj | cake.build
-------------|--------------
DumpSettings | -
Clean |
CleanAll |
CleanFramework |
CleanAllFrameworks |
CleanEngine |
CleanConsole |
Build | Build
BuildAll |
BuildFramework |
BuildAllFrameworks |
BuildEngine |
BuildCppTestFiles | NYI
BuildConsole |
Test
TestAll
TestFramework
TestFrameworkUsingConsole
TestFrameworkIfSupported
TestAllFrameworks
TestNUnitLite
TestNUnitLiteIfSupported
TestEngine
TestAddins
TestV2Driver
TestConsole
Package
PackageSource
PackageBinaries
PackageNuGet
PackageNuGetNUnit
PackageNuGetNUnitLite
PackageNuGetConsole
PackageNuGetRunners
PackageNuGetEngine
PackageMsi
PackageFrameworkMsi
PackageConsoleMsi
PackageCF
PackageBinariesCF

CharliePoole:
@oznetmaster Last commit also fixes a compiler error in CF, which somehow got missed in the course of other changes.

CharliePoole:
This is now ready for review and merge. Overview...

* The cake script now does everything that the msbuild script did with the exception of creating the two msi files we are not distributing. They could be easily added.
* Granularity of commands is changed to some extent.
* Old script files are removed.
* Build.ps1 and build.sh are kept unchanged from the Cake distribution
* Build.cmd calls build.ps1 feeding in any environment settings from CAKE_ARGS. The main use of this is for 32-bit windows systems that require the -Experimental option. It also allows running powershell on systems that don't automatically know what to do with .ps1 files.
* The build bash script does the same as build.cmd on Linux, although I have not actually needed it.
* I used low-level commands to build one project at a time rather than using the solution file. This was needed on Linux and for Silverlight on Windows in order to specify different settings for each project.
* I used low-level commands for testing rather than the NUnit3 extension to Cake. The framework seems better served using the NUnitLite runners, especially if we later separate the project. My thought is that we should use the NUnit3 extension on our non-core projects.
* I intended to use GitVersion but it ended up not being necessary. The logic for naming artifacts is the same as was previously used in appveyor.ps1.

devlead:
@CharliePoole you've been really busy, nice progress and very structured Cake script :+1:

rprouse:
Looks good. I think we might be able to simplify and drop a few build targets that we don't end up using, but let's run with it for awhile.

PR #1064, opened 2015-11-24
Async support on Windows 10

### LINKED ISSUE #1037: Async tests not working on Windows 10 Universal
Copied from nunit/nunit.xamarin#35 reported by @tofutim. I have confirmed that this is an issue with the portable framework, so may be an issue with other platforms.

From the other issue

Is this a UWP, nunit, nunit.runners issue? I added this to the tests in this solution after seeing it in mine. This is a result from the UWP emulator (Windows 10 mobile) in VS2015 on Server 2012 R2.

        [Test]
        public async Task AsyncTest()
        {
            var result = await anInt();
            Assert.AreEqual(42, result);
        }

        private async Task<int> anInt()
        {
            return 42;
        }

image

rprouse:
Fixes #1037 

Do not attempt to get Result from VoidTaskResult which is not supported on all platforms

### DIFF
--- a/src/NUnitFramework/framework/Internal/AsyncInvocationRegion.cs
+++ b/src/NUnitFramework/framework/Internal/AsyncInvocationRegion.cs
@@ -115,6 +115,7 @@ at wrapping a non-async method invocation in an async region was done");
         {
             private const string TaskWaitMethod = "Wait";
             private const string TaskResultProperty = "Result";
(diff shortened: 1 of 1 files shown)


CharliePoole:
Let's make sure to incorporate this in other branches we work on for 3.0.1, so that we can be testing it in as many situations as possible.

PR #1069, opened 2015-11-26
Fixes #1014 - API assembly now has a file version

### LINKED ISSUE #1014: Ensure NUnit API assembly updates with MSI installs
We are currently versioning the assembly at 3.0.0.0  with the intention that the API will not change. The internals may change occasionally, but if the version doesn’t change, then the MSI won’t update it.
 
We’ll need to figure out how to ensure it gets updated. I can think of a few ways;

- Force uninstall on install in the MSI. 
- Continually increment the third digit of the file version and just leave the assembly version the same.
- Don't distribute the console/engine as an MSI
- Use a different installer like NSIS

rprouse:
Switched the assembly versions back to 3.0.* for the release. We will need to switch them back to 3.1 when we merge back into master.

Added an AssemblyFileVersion to match the release to the API assembly. I wanted to add this to `CommonAssemblyInfo.cs` so that all of our assemblies have a file version that matches the release version and it is only the assembly version that increments, but I could not do that because the MSI thinks 3.0.5797 is newer so it won't install. It sort of bothers me that we are releasing 3.0.1, but that isn't reflected in the binaries. I would like to switch when we go to 3.1.

### DIFF
--- a/src/NUnitEngine/nunit.engine.api/nunit.engine.api.csproj
+++ b/src/NUnitEngine/nunit.engine.api/nunit.engine.api.csproj
@@ -48,7 +48,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="..\..\CommonAssemblyInfo.cs">
-      <Link>CommonAssemblyInfo.cs</Link>
+      <Link>Properties\CommonAssemblyInfo.cs</Link>
     </Compile>
     <Compile Include="..\..\Common\nunit\Logging\ILogger.cs">
       <Link>ILogger.cs</Link>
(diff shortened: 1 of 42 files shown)


src/NUnitEngine/nunit.engine.api/nunit.engine.api.csproj
```diff
@@ -60,7 +60,7 @@
       <Link>TestSelectionParserException.cs</Link>
     </Compile>
     <Compile Include="..\EngineApiVersion.cs">
-      <Link>EngineApiVersion.cs</Link>
+      <Link>Properties\EngineApiVersion.cs</Link>
```
rprouse: I've been moving the various assembly info files into the Properties so that they are grouped together and where people would expect them. I took the opportunity here to finish doing it in all the projects.

CharliePoole: It's a good idea.

CharliePoole:
Agreed on making AssemblyFileVersion always match the release starting with 3.1.

When we finally merge the release branch to master, we'll then have to change master back to 3.1. That's one awkward thing about working this way. One thing to consider in the future is to only have odd numbers in the repo and do releases by setting a tag with the even number in it.

rprouse:
I don't think we will normally have to change the version when we merge back into master. It is only happening in this case because we did not branch off the 3.0 release, but a few commits out. In hindsight, we probably should have branched off the 3.0 release and cherry picked the commits we wanted in the 3.0.1 release leaving out the change to 3.1.

PR #1073, opened 2015-11-26
Added the pause option back in

### LINKED ISSUE #1024: Please add --pause flag to console runner
The console runner documentation states there is `--pause` flag. But, this option has magically disappeared in the latest release candidate. I would like to be able to attach a debugger before tests run.

rprouse:
Fixes #1024 

The behaviour is a bit different than the original behaviour,

- Only pauses in the console if your are running InProcess
- Increased the remoting timeout to 5 minutes if pausing or debugging. During testing, it timed out a couple of times leaving hung agent processes
- The console and the agent now output their name and PID from the process object to make it easier to attach
- I considered making the pause and debug options mutually exclusive in the agent, but didn't think it was necessary


oznetmaster:
How about nunitlite?

rprouse:
@oznetmaster Debug isn't in nunitlite and pause wasn't originally, so I didn't think it was important. To be honest, I don't think it is terribly important in the console either, but makes sense in the agents. In nunitlite, the tests are already self-executable and in your solution, so they are easy to debug.

What is your use case? Remote debugging? Even in CF, except for your devices and maybe a few others, there is no console, so there is no way for the user to continue execution once they have attached except maybe changing the execution pointer.

oznetmaster:
I implemented --pause:<pause time> in my CF builds to allow time for a remote debugger to attach.  If nothing happens, the tests continue after the wait time.  If a debugger is attached within the wait time, then a break into the debugger occurs.

---Sent from Boxer | http://getboxer.com

On November 26, 2015 at 11:07:50 AM CST, Rob Prouse <notifications@github.com> wrote:@oznetmaster Debug isn't in nunitlite and pause wasn't originally, so I didn't think it was important. To be honest, I don't think it is terribly important in the console either, but makes sense in the agents. In nunitlite, the tests are already self-executable and in your solution, so they are easy to debug.  What is your use case? Remote debugging? Even in CF, except for your devices and maybe a few others, there is no console, so there is no way for the user to continue execution once they have attached except maybe changing the execution pointer.  —Reply to this email directly or view it on GitHub.

CharliePoole:
It's true that the old --pause implementation was in the console and agent. However, I think that the --debug option puts the Debugger.Launch in a better place, in NUnitTestAssemblyRunner. That takes care of both in-process and out-of-process execution. Seems like the revived pause should be right after that code.

BTW, the debug-agent code is in agent because it's for us to use in debugging the agent. When I get around to refactoring Options.cs, I'd like this to be a hidden or developer option that doesn't normally show up in the help.

rprouse:
I can move it to where the new debug code is. My worry with that was determining if it should be console output depending on if it was the agent or console process. I could look for console in the name, or just always show a messagebox?

Aside, I wonder if there is a way we can use either of these options to break into the Visual Studio test runner process when we are debugging the Visual Studio Adapter? Just a thought :)

rprouse:
I have move the debug code to NUnitTestAssemblyRunner as suggested. While I was at it, I updated the icons for our three EXE's to the NUnit icon...

PR #1076, opened 2015-11-27
Added the filename to the error message when failed to load settings

### LINKED ISSUE #1061: Nunit30Settings.xml becomes corrupted
While trying to update a project with some failing tests (using NUnitForms), in some cases Nunit30Settings.xml becomes corrupted and is filled with a bunch of "nul"s, which, on the next run of nunit3-console results in the following exception:
```
     [exec] System.ApplicationException: Error loading settings file ---> System.Xml.XmlException: Root element is missing.
     [exec]    at System.Xml.XmlTextReaderImpl.ThrowWithoutLineInfo(String res)
     [exec]    at System.Xml.XmlTextReaderImpl.ParseDocumentContent()
     [exec]    at System.Xml.XmlLoader.Load(XmlDocument doc, XmlReader reader, Boolean reserveWhitespace)
     [exec]    at System.Xml.XmlDocument.Load(XmlReader reader)
     [exec]    at System.Xml.XmlDocument.Load(String filename)
     [exec]    at NUnit.Engine.Internal.SettingsStore.LoadSettings()
     [exec]    --- End of inner exception stack trace ---
     [exec]    at NUnit.Engine.Internal.SettingsStore.LoadSettings()
     [exec]    at NUnit.Engine.Services.SettingsService.StartService()
     [exec]    at NUnit.Engine.Services.ServiceManager.StartServices()
     [exec]    at NUnit.Engine.TestEngine.Initialize()
     [exec]    at NUnit.Engine.TestEngine.NUnit.Engine.ITestEngine.get_Services()
     [exec]    at NUnit.ConsoleRunner.ConsoleRunner..ctor(ITestEngine engine, ConsoleOptions options, ExtendedTextWriter writer)
     [exec]    at NUnit.ConsoleRunner.Program.Main(String[] args)
```

As a workaround, one can delete this file, but it is pretty annoying and most importantly it wasn't clear what settings file it was referring to in the first place.

rprouse:
Fixes #1061

### DIFF
--- a/src/NUnitEngine/nunit.engine/Internal/SettingsStore.cs
+++ b/src/NUnitEngine/nunit.engine/Internal/SettingsStore.cs
@@ -89,8 +89,8 @@ namespace NUnit.Engine.Internal
             }
             catch (Exception ex)
             {
-                string msg = string.Format("Error loading settings {0}.", _settingsFile);
-                throw new ApplicationException(msg, ex);
+                string msg = string.Format("Error loading settings {0}. {1}", _settingsFile, ex.Message);
+                throw new NUnitEngineException(msg);
             }
         }
 
--- a/src/NUnitEngine/nunit.engine/Internal/SettingsStore.cs
+++ b/src/NUnitEngine/nunit.engine/Internal/SettingsStore.cs
@@ -90,7 +90,7 @@ namespace NUnit.Engine.Internal
             catch (Exception ex)
             {
                 string msg = string.Format("Error loading settings {0}. {1}", _settingsFile, ex.Message);
-                throw new NUnitEngineException(msg);
+                throw new NUnitEngineException(msg, ex);
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/NUnitEngine/nunit.engine/Internal/SettingsStore.cs
```diff
@@ -89,7 +89,8 @@ public void LoadSettings()
             }
             catch (Exception ex)
             {
-                throw new ApplicationException("Error loading settings file", ex);
+                string msg = string.Format("Error loading settings {0}.", _settingsFile);
+                throw new ApplicationException(msg, ex);
```
CharliePoole: I think this should be an NUnitEngineException, which will get displayed without the stack trace.

CharliePoole:
See my note about the exception type. Otherwise, it's ready to merge.

src/NUnitEngine/nunit.engine/Internal/SettingsStore.cs
rprouse: Good idea, I will update. Do we show the inner exception in that case? If not, I will add the message from the inner exception. When I was testing, it told me exactly what was wrong with the file.

rprouse:
Changes based on your feedback have been pushed. The message from the inner exception is not displayed with the NUnitEngineException, so I added it to the message to tell people what is wrong with the settings file.

PR #1084, opened 2015-11-28
Issue 1049

rprouse:
Xml escaping is now done by NUnit so that users don't have to do it.

I also added new tests that verify the filters work by counting the number of cases found.

### DIFF
--- a/NUnit.proj
+++ b/NUnit.proj
@@ -7,9 +7,9 @@
   <PropertyGroup Label="Common Properties">
     <ProjectName>$(MSBuildProjectName)</ProjectName>
     
-    <PackageVersion Condition="'$(PackageVersion)' == ''">3.1.0</PackageVersion>
+    <PackageVersion Condition="'$(PackageVersion)' == ''">3.0.1</PackageVersion>
     <PackageModifier Condition="'$(PackageModifier)' == ''"></PackageModifier>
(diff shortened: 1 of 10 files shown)


rprouse:
@CharliePoole not sure where you branched from on this one, but it is pulling in the README changes that are on master. It isn't a problem, but I am assuming you want this in 3.0.1? I will review based on that assumption, please let me know if it is otherwise.

CharliePoole:
Not really surprising. Between Thanksgiving and being sick, I have been working on this one in small bits, coming back to it several times. I must have messed up.

I'd rather not have to do it over because it was quite a journey getting it all to work.

CharliePoole:
And yes, I did intend it for 3.0.1.

rprouse:
OK, looks good then. Hope you are feeling better.

CharliePoole:
Ah! If you look at the last commit, you can see how the master changes got into this PR. I let GitHub merge in the changes as it offered to do. It's clear that this approach to releasing is a little bit out of sync with how GitHub wants to work.

rprouse:
Yes, we need to be careful working this way. I think the original mistake was not setting the 3.0.1 target branch from the default of master. I keep doing the same thing. That is why it offered to merge latest.

PR #1086, opened 2015-11-29
Display output from OneTimeSetUp and OneTimeTearDown. Fixes #1062

### LINKED ISSUE #1062: Console.WriteLine statements int "OneTimeSetUp" and "OneTimeTearDown" annotated methods are not directed to the console when using nunit3-console.exe runner
Given this test fixture: 
```csharp
	[TestFixture]
	public class TestClass
	{
		[OneTimeSetUp]
		public void FixtureSetup ()
		{
			Console.WriteLine ("FixSetUp");
		}

		[SetUp]
		public void SetUp ()
		{
			Console.WriteLine ("SetUp");
		}

		[TearDown]
		public void TearDown ()
		{
			Console.WriteLine ("TearDown");
		}

		[OneTimeTearDown]
		public void FixtureTearDown ()
		{
			Console.WriteLine ("FixTearDown");
		}

		[Test]
		public void Test ()
		{
			Console.WriteLine ("Test");
		}
	}
```

It would be expected that `Console.WriteLine ("FixSetUp");` and `Console.WriteLine ("FixTearDown");` are correctly printed to the console. However the actual output is as follows:

```
NUnit Console Runner 3.0.5797 

Runtime Environment
   OS Version: MacOSX 15.0.0.0 
  CLR Version: 4.0.30319.17020

Test Files
    bin/Debug/XQA.XS.Tests.dll

Test Filters
    Test: XQA.XS.Tests.TestClass.Test

=> XQA.XS.Tests.TestClass.Test
SetUp
Test
TearDown

Run Settings
    WorkDirectory: <omitted>
    NumberOfTestWorkers: 4

Test Run Summary
    Overall result: Passed
   Tests run: 1, Passed: 1, Errors: 0, Failures: 0, Inconclusive: 0
  Start time: 2015-11-23 20:02:32Z
    End time: 2015-11-23 20:02:32Z
    Duration: 0.063 seconds

```
Notice there is no appearance of "FixSetUp" or "FixTearDown" - this is a major issue for build systems that rely on Console output to trigger events (such as the [mono build system] (https://github.com/mono/monkeywrench)).

Thanks!

CharliePoole:
This fix displays the output from OneTimeSetUp and OneTimeTearDown. The order of display is a bit messy, since it doesn't come until the suite is finished. We may want to change this in a future version, but it will require handling OneTimeSetUp and OneTimeTearDown as separate commands, which is a fairly significant change.

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/ConsoleOutputTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/ConsoleOutputTests.cs
@@ -1,0 +1,75 @@
+﻿// ***********************************************************************
+// Copyright (c) 2015 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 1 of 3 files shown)


rprouse:
I created #1096 so that we can fix this to move the output for the setup to before the tests. Other than that, looks good.

PR #1098, opened 2015-12-01
Fix warning message

CharliePoole:
Quick fix for a warning message about the class having no CLS-compliant constructors.


rprouse:
It really is a shame that we need to lose type safety in the name of CLS compliance. Oh well :frowning:

CharliePoole:
Fact is, we don't use those constructors, so we lose nothing. Seems like a
heavy fix for a warning. I guess the "right" way is to figure out what is
not CLS-compliant about the Types in the constructors. I went down two or
three levels and didn't see anything - decided to do the quick and dirty
fix. We can look more closely after the release.

PR #1120, opened 2015-12-08
NUnitProject should parse .nunit project files containing Xml Declarations

teyc:
I was handcrafting an .nunit project and had included the `<?xml version="1.0"?>` header in the file, and nunit3-console was unable to parse it properly, as `firstChild` refers to the xml declaration node. A one line fix to use the `documentElement` instead was sufficient to fix the bug.

I've included a test case illustrating the issue.

### DIFF
--- a/src/NUnitEngine/Addins/addin-tests/NUnitProjectLoaderTests.cs
+++ b/src/NUnitEngine/Addins/addin-tests/NUnitProjectLoaderTests.cs
@@ -86,10 +86,12 @@ namespace NUnit.Engine.Services.ProjectLoaders.Tests
             }
         }
 
+        [TestCase("NUnitProject.nunit")]
+        [TestCase("NUnitProject_XmlDecl.nunit")]
         [Test]
-        public void LoadNormalProject()
+        public void LoadNormalProject(string resourceName)
         {
-            using (TestResource file = new TestResource("NUnitProject.nunit"))
+            using (TestResource file = new TestResource(resourceName))
             {
                 IProject _project = _loader.LoadFrom(file.Path);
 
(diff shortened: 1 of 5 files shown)


rprouse:
Looks good to me. I am assigning it to a milestone so that it will get picked up in the release notes after it is merged.

src/NUnitEngine/Addins/addin-tests/NUnitProjectLoaderTests.cs
```diff
@@ -86,10 +86,12 @@ public void LoadEmptyConfigs()
             }
         }
 
+        [TestCase("NUnitProject.nunit")]
+        [TestCase("NUnitProject_XmlDecl.nunit")]
         [Test]
```
CharliePoole: TestAttribute is now redundant. As a matter of style, we don't normally use it wiith test cases.

src/NUnitEngine/Addins/nunit-project-loader/NUnitProject.cs
```diff
@@ -156,7 +156,7 @@ public string GetSetting(string name, string defaultValue)
         /// </summary>
         internal XmlNode RootNode
         {
-            get { return xmlDoc.FirstChild; }
+            get { return xmlDoc.DocumentElement ; }
```
CharliePoole: Good pickup. There are other places where we use FirstChild and should probably use DocumentElement.

CharliePoole:
Nice catch! And good on you for adding a test for it. :-)

@rprouse I think it's mergeable either as is or with fixes noted.

rprouse:
@teyc thanks for merging latest from master. I was just about to do that :smile: 

If you get a chance, can you remove the extra `[Test]` that @CharliePoole pointed out? If not, I will fix it.

PR #1123, opened 2015-12-09
NUnitProject.GetTestPackage(config) should fail on missing configuration

teyc:
I was caught by a rookie mistake. Running nunit3-console against a missing configuration results in BadImageFormatException. This is because the project loader tries to load the .nunit project as a .NET assembly when it couldn't find a configuration with the given name.

This patch gives the correct error message on the console, and also provides some helpful suggestions.


rprouse:
Looks good to me. Can we get a second review?

CharliePoole:
Looks good to me as well.

CharliePoole:
@teyc Please merge the latest changes from master into your branch and resolve any conflicts. We will then merge the result back to master.

Thanks for your work!

rprouse:
@teyc sorry we didn't get to this earlier, I didn't notice that you had merged in master. For future reference, let us know when you make additional commits to a PR. We get notification emails for comments, but not for commits.

I won't ask you to merge in from master again. I am going to merge manually. Thanks for your work.

Issue #533, opened 2015-03-12
Separate NUnitLite runner and autorunner


CharliePoole:
Changes in PR #526 included a new AutoRun class, for use by self-running tests. However, the class still includes some elements from the old runner, which are not appropriate for an autorunner. In particular, other input files are still accepted on the command line although they are ignored with a warning.

We should have a separate runner class for use in running one or more tests under nunitlite. We may want this to be in a separate assembly (an exe) as well.

Additionally, each runner should only accept the options that are relevant to it, which may involve some refactoring of the ConsoleOptions class.

oznetmaster:
Currently, I cannot adapt this at all for my  NETCF environment.

 

Are you writing a separate runner class for use in running one or more tests under nunitlite?  You have removed that capability in the autorun class.

oznetmaster:
I just found TextRunner.  That should work for me.

oznetmaster:
TextRunner appears to have lost a lot of options that TextUI used to have, and that AutoRun still has.

 

For instance, it does not initialize the log, and it does not set the outWriter and errWriter, or process the various other options.

 

Is this intentional, or is it just an oversight?

oznetmaster:
I adapted AutoRun into ConsoleRunner, and now have full capability equivalent to what was available before.

 

After I test it, I will generate a PR for it so that there will be a full console capability.

oznetmaster:
Since really the only current difference between AutoRun and ConsoleRunner is the support for specifying the test assemblies, perhaps it should be the same with just a flag?  No reason to support two basically identical runners.

CharliePoole:
@oznetmaster Please keep comments related to the issue on which you make them. 

Your series of comments here seem to be related to work already done on other issues and to the general direction of the nunitlite runnner. Github gives us what amounts to a separate discussion group for each issue and it seems to help our communications when we take advantage of that for individual issues. For general discussion, we have either the committers maillist or the nunit-developer group, depending on who you want to include as recipients.

This particular issue was created to support exactly the type of execution you are asking for. We'll work on it for the second beta.

CADbloke:
[...]

Issue #532, opened 2015-03-12
Batch runner for Silverlight tests


CharliePoole:
Prior to its 1.0 release, NUnitLite had a batch runner that allowed the CI build to run Silverlight tests. We should revive that feature and make it work for 3.0.

Issue #1018, opened 2015-11-12
ArgumentException when 2.x version of NUnit Framework is in the bin directory


alex-mazzariol:
When testing an assembly with a target framework version which has only 2 fields (say, major = 4 and minor = 5, which is what I am testing now, compiled with mono 4.0.4) the code in NUnit.Engine.Services.TestAgency.LaunchAgentProcess() fails while calling System.Version.ToString(), raising an ArgumentException with message "Argument must be between 0 and 2. Parameter name: fieldCount".

I believe the problem to be around lines 204 and 204 of file  nunit/src/NUnitEngine/nunit.engine/Services/TestAgency.cs as of commit 272eadd. There the version is supposed to have at least 3 components, when there may not be that many.

rprouse:
Which version of NUnit are you using and with what command line? We compile and on Linux with Mono, so we must be missing something. Can you provide more steps to reproduce?

alex-mazzariol:
Just updated everything to the last 3.0 RC; version of NUnit framework and console runner is 3.0.5790.16288, command line is just
`nunit3-console.exe --noc TestAssembly.dll`
Assembly compiled with MonoDevelop's xbuild engine version 12.0, with Mono Version 4.0.4.0 (from Xamarin's RPM repository, running on Fedora 21). All the assemblies in the solution target "Mono / .NET 4.5" (from Project Options).

Problem persists also with this last version; when starting the console runner, this is the output:
[console-runner-20151112.txt](https://github.com/nunit/nunit/files/33348/console-runner-20151112.txt)

CharliePoole:
The error would have to be somewhat upstream. We'll have to figure out how the ClrVersion field of the RuntimeFramework object ends up with only two digits. Do you know what this version of mono puts into the assembly as the image runtime version?

alex-mazzariol:
Thank you for your fast responses; I checked the "image runtime version" using ILDASM on Windows, looking for "Metadata Version" as per http://blogs.msdn.com/b/suzcook/archive/2003/06/20/57191.aspx, however she says that the ImageRuntimeVersion of an Assembly may not be the actual required one.

ILDASM says the Metadata Version is 4.0.30319.

[...]

Issue #971, opened 2015-11-01
Create a cmd file for running the silverlight tests


CharliePoole:
If possible, it should look in the registry to find where sllauncher.exe is installed.

PR #1128, opened 2015-12-10
Issue 922

### LINKED ISSUE #922: Wrap Console in NUnitLite
The portable version of NUnitLite passes in a TextWriter/Reader from Console.Out/In. It would be better to wrap the console in an interface and pass it into AutoRun so that we can have the same code for portable and other frameworks. We could abstract this detail out of frameworks that support Console by default using #defines, but it would allow us to have a color console for portable or to pass in different TextWriters/Readers for platforms that don't support a console.

CharliePoole:
Fixes #922, allowing portable tests to run using ColorConsole. 

The ExtendedTextWriter interface is used and an implementation must be supplied to the portable build of NUnitLite. This can be done through an executable (non-portable) test or a separate runner program.


rprouse:
@CharliePoole I figured out what is causing the test failure on Linux. Your changes to NUnit.proj are not building the target frameworks. Using your branch, I deleted my bin directories and ran the command that Travis uses,

```
xbuild NUnit.proj /p:Configuration=Release /p:Framework=net-4.5 /t:"Build;TestFramework"
```

It didn't build the .NET 4.5 framework, so when it went to test the framework, the assembly wasn't there. It looks like that issue #1018 is that we throw the ArgumentException if the assembly being tested does not exist. I tested on Windows and the correct File not found error is reported.

CharliePoole:
My guess is that the double indirection trick in this code is the problem:

```
  <Target Name="BuildFramework" Label="Build the framework for the current config and runtime">
    <MSBuild Targets="Build" Projects="@($(FrameworkProjects))" 
      Properties="Configuration=$(Configuration); Platform=AnyCPU" />
  </Target>
```

I'll started trying to simplify the build structure in MsBuild when it appeared that Cake wasn't going to work. I'll just revert it to it's earlier state when I get back to my Linux box - I'm on my laptop right now.

CharliePoole:
Confirmed that using the copy of mono on my Windows laptop. It 's looking for a project named @(FrameworkProjects_45)!

CharliePoole:
I guess the two flavors of Cake (mono versus Roslyn) are not any worse to work with than the two flavors of msbuild!

rprouse:
And at least the Cake scripts don't make me want to poke my eyes out like staring at Msbuild XML :smile:

CharliePoole:
This is ready to review and merge

rprouse:
Travis failures look like random mono failures so I have restarted. I will review later.

src/NUnitFramework/nunitlite.runner/AutoRun.cs
```diff
@@ -75,7 +82,48 @@ public int Execute(string[] args)
                 Console.SetError(errWriter);
             }
 
-            var _textUI = new TextUI(outWriter, options);
+            try
+            {
+                return Execute(testAssembly, new NUnitLiteOptions(args), new ColorConsoleWriter(), Console.In);
```
rprouse: Shouldn't you be passing in the options and the outWriter that you created above rather than creating new ones? This breaks the OutFile option.

CharliePoole: It sure looks like it! I'll double check and update.

rprouse:
You added ColorConsole, ColorConsoleWriter and ExtendedTextWrapper to the nunitlite.tests.exe-portable. How are users going to get those if they want to write portable nunitlite tests? Are we going to add them to the NuGet package like we do for Program.cs? If so, they should be added as part of this PR.

rprouse:
I think ExtendedTextWrapper can be moved to nunitlite-portable.

CharliePoole:
I agree we need to distribute something that lets users write both portable tests and .NET 4.5 tests that reference a portable component.

I'm not sure why ExtendedTextWrapper is not already in nunitlite-portable but I'll check.

Unfortunately, this will all have to wait till I get VS reinstalled. :-(

rprouse:
I think ExtendedTextWrapper is not already in portable because it wasn't used by the old custom autorunner or textui.

CharliePoole:
@rprouse Turns out that the portable version of TextWriter doesn't have a Close() method! So ExtendedTextWrapper would not compile.

However, it seems we don't actually call that method any longer but only use Dispose().

rprouse:
I switched all the usages of Close() to Dispose() as part of the .NET Core project since Close isn't supported on DNX. I guess I missed ExtendedTextWrapper because I wasn't using it.

CharliePoole:
Still a few commits away from ready. :-)

CharliePoole:
@rprouse While there's lots more that could be done on NUnitLite, I think this is good to go for now. We can look at it again when we are thinking about where to go with the runners.

CharliePoole:
Ah wait... still need to do the packaging stuff!

CharliePoole:
This PR is becoming bigger than I intended. The last series of commits add a new program to each framework build, which is able to run a single assembly specified on the command-line. It's a short stub that passes on the work to the TextRunner in NUnitLite. The portable and silverlight builds include their own copies of the color console writer.

Key changes:
* The new runner is called nunitlite-runner, but it could be renamed if someone has a better idea.
* Self-executing tests still use the simplified AutoRun class, which passes the work to TextRunner.
* The nunitlite-runner executable uses TextRunner directly.
* We now test the silverlight build using the new runner as a part of the Appveyor build.
* The nuget and zip packages include the new program. The msi installer doesn't have it yet.

More work can still be done on NUnitLite, but I think we should limit this PR to what's already there and any necessary fixes. We can come back to it.

With the new runner, this PR will now also fix #532 and remove the need for #971.

oznetmaster:
Sounds exactly like the "runner" I have been distributing with my CF builds since the beginning, but which you previously had indicated was only useful for the CF build for a device which had a console?

CharliePoole:
You could easily make one for cf as well, just as we do for silverlight,
which doesn't have a console. This runner is for ci purposes so what I'd
really like is a runner executing cf tests on the desktop.

rprouse:
@CharliePoole, is this ready for review and merge?

CharliePoole:
Yes it is.

CharliePoole:
Possible alternative names for the nunitlite runner:
 * nunitlite-runner (current name)
 * nunitlite-console 
 * nunitlite.ci.runner (or with dashes)

The last one is what I originally used. Neil's comment leads me to think about it again since this is not really intended to be a runner that runs on the platform itself, except in the case of .NET 2.0, 4.0 and 4.5. That is, the portable and silverlight runners execute as .NET 4.5 console applications. If I succeed in writing one for CF, it will be a desktop app that loads and runs CF tests. (But that won't be on this PR)

BTW, this PR also fixes #533

nuget/nunitlite.nuspec
```diff
@@ -37,10 +37,15 @@ How to use this package:
...
     <file src="bin/net-2.0/nunitlite.dll" target="lib/net20" />
+    <file src="bin/net-2.0/nunitlite-runner.exe" target="lib/net20" />
     <file src="bin/net-4.0/nunitlite.dll" target="lib/net40" />
+    <file src="bin/net-4.0/nunitlite-runner.exe" target="lib/net40" />
     <file src="bin/net-4.5/nunitlite.dll" target="lib/net45" />
+    <file src="bin/net-4.5/nunitlite-runner.exe" target="lib/net45" />
     <file src="bin/portable/nunitlite.dll" target="lib/dotnet" />
+    <file src="bin/portable/nunitlite-runner.exe" target="lib/dotnet" />
     <file src="bin/portable/nunitlite.dll" target="lib/portable-net45+win8+wp8+wpa81+Xamarin.Mac" />
+    <file src="bin/portable/nunitlite-runner.exe" target="lib/portable-net45+win8+wp8+wpa81+Xamarin.Mac" />
```
rprouse: I am pretty sure that this and the `lib/dotnet` one above are wrong. The runner is a .NET 4.5 console application and won't actually run on most of the platforms specified. It will only run on the desktop and is functionally equivalent to the net-4.5 runner, if not identical now that it doesn't include tests.

EXE's are not portable. If we want to go for this approach, we should be creating platform specific nunitlite-runners for each potential platform where it is possible and including them in platform specific lib directories. Of the targets listed there, .NET Core is probably the only one with a console that we can use. You can shell into Android to the console, but I am not sure how practical it would be, even in a CI scenario.

I don't think we need to solve this with this PR, it is getting too big already. I think you should remove those two nunitlite-runner.exe's from dotnet and portable and we can create another issue to create additional platform specific runners.

rprouse: If in the future we decide to merge nunitlite and the xamarin runners, they might fit this scenario.

rprouse:
I am fine with the name nunitlite-runner. nunitlite-console.exe also works, but we may decide in the future that some runners are not console runners, so the name you are using gives us more latitude in the future.

Other than my comment about the NuGet file, I think this is ready to merge. It is getting way too big to code review properly, so I would prefer to get it in early so we can give it a thorough testing in our day to day work.

nuget/nunitlite.nuspec
CharliePoole: You are right that this is not portable. However, it is the runner to be used to run portable tests from the desktop. In principle, I suppose it should go under tools, but since we have five different flavors of runner, I put it under lib. Do you have any ideas as to how we should distribute this?

BTW, the same question arises with the silverlight runner.

CharliePoole: Also, the question could arise with the xamarin runners. Hypothetically, you might have a runner that is used for CI and a different runner that executes on the platform itself.

rprouse: The runner will only run portable tests if the tests themselves are portable. That may be the case in some situations, but so far, almost all of the testing I am doing involves tests compiles for one of the platforms that portable supports. In that case, I am pretty sure the .NET 4.5 based runner will not run them.

If they do create portable tests, then the runner in the .net 4.5 directory will run them.

Thinking about it that way, I think all the runners should be in a tools directory with platform specific sub-directories.

Your comment on Xamarin is actually quite prescient. I wrote a blog post on [Testing Xamarin Projects using NUnit 3](http://www.alteridem.net/2015/12/21/testing-xamarin-projects-using-nunit-3/) today. In it, I am advocating that users create a shared project for there tests and reference it from a .NET 4.5 test project and from platform specific test projects. The .NET 4.5 project can then run the majority of tests in a CI environment, and the platform specific tests can add the ability to run the tests on the platform and additionally run a few platform specific tests. Currently working in Xamarin, I am seeing 90 to 95% code reuse across platforms and my tests also reflect that ratio.

CharliePoole: Exactly... that's why I put it under portable. :-)

I was planning on writing a blog post about testing portable apps, which includes the two different approaches:
 * Using portable tests
 * Using .NET 4.5 tests
The nunitlite-runner built against the portable library is only intended for the first case, which should be rare except for tests that involve gui elements.

I'll change the directory structure as you suggest. I'll use the standard NuGet names for the directories rather than our names, to avoid confusion.

I'm not giving up entirely on running xamarin tests from the desktop as well! People have said you can't run silverlight or CF tests from the desktop, but I've managed to do both. (CF was in the old NUnitLite project under NAnt. I want to figure out how to do it under Cake rather than wasting time with MsBuild.)

rprouse: Once we get this checked in, I will test running various types of tests using the runners and we can get a better feel for this. I agree that the majority of tests should run fine on the desktop. I am currently only running into problems in edge cases where a class expects to be initialized on a specific platform through dependency injection and I can't create a suitable mock for it on the desktop.

PR #1130, opened 2015-12-11
Issue 922 test

rprouse:
**Do not merge**

I want to see this fail on Travis because of #1018, then I am going to push a change to TestAgency to see if my suspicion about the cause are correct.


rprouse:
Switching to targetRuntime.ClrVersion.ToString() got past the problem and launched mono, but then mono reported the error,

```
WARNING: The runtime version supported by this application is unavailable.
Using default runtime: v4.0.30319
```

I am putting a temporary Console.WriteLine to see what we are parsing this as, then I am going to pull out my Linux box and update it to the latest version of Mono to try and figure out why we are not returning the correct ClrVersion anymore.

rprouse:
Well, that would do it. We are reporting the ClrVersion as `v0.0`.

CharliePoole:
ClrVersion 0.0 should mean "any version" It's valid on the target framework, if no version was specified. It should never come up on any real framework, like current framework or the framework for which an assembly was built.

CharliePoole:
If you are working this, I should block on issue 922.

rprouse:
I am closing this branch. I am just trying to figure out #1018, but have exhausted my ideas here so am moving to my linux box to repro.

PR #1136, opened 2015-12-13
Don't allow V2 framework to update in V2 driver tests

CharliePoole:
MonoDevelop was offering to update this package to 3.0.1, which would defeat the purpose of the v2 driver tests.


rprouse:
I have restarted the one failing build, it was a mono crash.

rprouse:
Looks good. I will merge when CI passes.

PR #1137, opened 2015-12-14
Do not merge yet. Initial attempt at a fix for issue 681.

rationull:
Notes:
- DirectTestRunner has to avoid installing the hook if it's run from a non-default AppDomain, in order to avoid installing a nonfunctional hook when running NUnit's own automated tests. The hook only makes sense to install in the default AppDomain given the issue at hand.
- Paths are removed from the resolver after running tests. This seems to work but I'm not 100% clear that it's impossible to re-run the tests without loading a new DirectTestRunner.
- The resolver hook is never actually uninstalled. This shouldn't be a functional problem but it would be cleaner to uninstall it.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Internal/ProvidedPathsAssemblyResolver.cs
+++ b/src/NUnitEngine/nunit.engine/Internal/ProvidedPathsAssemblyResolver.cs
@@ -1,0 +1,71 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.IO;
+using System.Diagnostics;
+
+namespace NUnit.Engine.Internal
+{
+    public class ProvidedPathsAssemblyResolver
+    {
+        public ProvidedPathsAssemblyResolver()
+        {
+            _resolutionPaths = new List<string>();
+        }
+
+        public void Install()
+        {
+            Debug.Assert(AppDomain.CurrentDomain.IsDefaultAppDomain());
+            AppDomain.CurrentDomain.AssemblyResolve += AssemblyResolve;
+        }
+
+        public void AddPath(string dirPath)
+        {
+            if (!_resolutionPaths.Contains(dirPath))
+            {
+                _resolutionPaths.Add(dirPath);
+            }
+        }
+
+        public void AddPathFromFile(string filePath)
+        {
+            string dirPath = Path.GetDirectoryName(filePath);
+            AddPath(dirPath);
+        }
+
+        public void RemovePath(string dirPath)
+        {
+            _resolutionPaths.Remove(dirPath);
+        }
+
+        public void RemovePathFromFile(string filePath)
+        {
+            string dirPath = Path.GetDirectoryName(filePath);
+            RemovePath(dirPath);
+        }
+
+        Assembly AssemblyResolve(object sender, ResolveEventArgs args)
+        {
+            foreach (string path in _resolutionPaths)
+            {
+                string filename = new AssemblyName(args.Name).Name + ".dll";
+                string fullPath = Path.Combine(path, filename);
+                try
+                {
+                    if (File.Exists(fullPath))
+                    {
+                        return Assembly.LoadFrom(fullPath);
+                    }
+                }
+                catch (Exception)
+                {
+                    // Resolution at this path failed. Do not interrupt the process; try the next path.
+                }
+            }
+
+            return null;
+        }
+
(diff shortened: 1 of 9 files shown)


rprouse:
@rationull, this PR has been open for over a month. Are you still working on it? Do you need help or advice?

CharliePoole:
@rationull We have been reducing the need for the engine to pre-load and check things in assemblies. Prior to deciding what runner to use for an assembly, we examine it using Mono.Cecil and determine the target framework and platform as well as picking up some assembly-level attributes. Any info that needs to be passed along is added to the package settings. I think we should do the same thing here rather than creating an appdomain and reflecting on the assembly an additional time.

However, I think we would probably like you to continue this - if you are willing - as a proof of concept. Once proven, either you or one of us could make it more conformable to other aspects of the engine.

rationull:
Charlie, the approach you describe sounds like a much cleaner design. I haven't looked at the codebase for a while so I'll have to check out what's been done with Mono.Cecil and adapt my changes to that approach.

This is in a tricky place for me as I would like to get the problem fixed but it is technically a rather low priority as it is an edge case and I've worked around it in the projects I have that are affected by it. I will revisit and continue but it may be a bit of time before I get a chance to.

In the mean time I will close this PR and create a new one when I've adapted the changes.

rationull:
@CharliePoole I've taken a look at the current code and use of Mono.Cecil, but I still question whether it's more appropriate to probe the test assembly and flag the new behavior in the engine services or in the agent/runner. I may also be missing some subtlety about the behavior when running with a separate agent process vs without.

In the engine services it looks like the probe would be triggered by TestAgency and implemented in RuntimeFrameworkService with the other checks (or perhaps more appropriately in a peer to RuntimeFrameworkService since the concern here isn't really about which framework to use) and then passed as an argument to the agent. However the agent/runner's default AppDomain must ultimately be responsible for implementing the new behavior so if the probing is also handled by the agent or direct runner then the feature touches fewer parts of the system. This was my logic for putting the check in DirectTestRunner in my initial commit.

Replacing the assembly load + reflection probing process with Mono.Cecil makes perfect sense but I'm less clear on this part. What's your opinion? Which approach fits better with the direction the architecture is moving?

CharliePoole:
@rationull I'm in the process of going through the engine logic and plotting out the chain of events when we load and run in-process versus through use of one or more agents. What I would like to accomplish is that everything is either checked at the beginning, in process, using Mono.Cecil or at the very end when we already are loading the test into a domain. I'm not sure that's possible, but I want to try for it. I'll try to have an answer for you in a day or two.

rationull:
@CharliePoole Sounds great. Certainly no rush for this on my account.

Given your description perhaps in-process in the beginning is a better spot for this particular check since the assembly resolver hook will need to be installed in the default AppDomain before the tests are actually being loaded.

CharliePoole:
Right... if we do it in the beginning, we can add the info to the TestPackage in some way and pass it along. That would enable us to install the hook in the default AppDomain of the original process, if running in-process, or of the agent if running in a separate process.

rationull:
@CharliePoole Is it the final word, then, that this should be done at the beginning, within the primary NUnit launcher process and then communicated to the runner (agent or otherwise) via TestPackage settings?

CharliePoole:
@rationull Yes, and PR #1263, inspired in part by this discussion, will make that possible once merged. Your comments on it are welcome, of course.

rationull:
I've modified my work given the changes in PR #1263 with the attribute check in RuntimeFrameworkService (not sure if it should actually be a new peer to RuntimeFrameworkService as this check isn't really about the runtime framework, but the x86 checks are also in there -- not sure what the intended scope is for that class) and the resolver hook install still in DirectTestRunner.

I can prepare another PR to replace this one when 1263 is merged.

CharliePoole:
In fact, RuntimeFrameworkService is already about more than RuntimeFrameworkService. It's about deciding how to go about executing the tests, which is a bit more general. It may get renamed and have new interfaces created when the dust settles.

rationull:
Perfect. I will monitor progress on that issue and proceed when those changes are buttoned up. Thanks!

CharliePoole:
@rationull #1263 is merged now

rationull:
PR #1272 created with the fix modified to use Mono.Cecil and analyze the test package in RuntimeFrameworkService.

PR #1142, opened 2015-12-17
[WIP] Fix #464: add support for Assert.ThrowsAsync

### LINKED ISSUE #464: Async delegate assertions
All assertions that take TestDelegate should have correlating methods that take an asynchronous equivalent (Task AsyncTestDelegate()). E.g., Assert.Throws(TestDelegate) should have a matching Assert.ThrowsAsync(AsyncTestDelegate) or Assert.Throws(AsyncTestDelegate). This would enable this kind of usage:

    await Assert.ThrowsAsync<Exception>(async () =>
    {
      await Task.Yield();
      throw new Exception();
    });

Also, some synchronous assertion code has outdated code; e.g., Assert.Throws will try to create an AsyncInvocationRegion that will never succeed, since AIR (correctly) rejects async void methods and TestDelegate returns void. Ideally, AIR should only be used when executing the [Test] methods themselves, not within any other code.

thomaslevesque:
Equivalent of `Assert.Throws` for non-void async methods.

### DIFF
--- a/src/NUnitFramework/framework/Assert.Exceptions.Async.cs
+++ b/src/NUnitFramework/framework/Assert.Exceptions.Async.cs
@@ -23,7 +23,6 @@
 
 #if NET_4_0 || NET_4_5 || PORTABLE
 using System;
-using System.Threading.Tasks;
 using NUnit.Framework.Constraints;
 using NUnit.Framework.Internal;
 
@@ -41,31 +40,25 @@ namespace NUnit.Framework
         /// <param name="code">A TestSnippet delegate</param>
         /// <param name="message">The message that will be displayed on failure</param>
         /// <param name="args">Arguments to be used in formatting the message</param>
-        public static Task<Exception> ThrowsAsync(IResolveConstraint expression, AsyncTestDelegate code, string message, params object[] args)
+        public static Exception ThrowsAsync(IResolveConstraint expression, AsyncTestDelegate code, string message, params object[] args)
         {
-            Exception caughtException;
-            try
+            Exception caughtException = null;
+            using (var region = AsyncInvocationRegion.Create(code))
             {
-                var task = code();
-                return task.ContinueWith(t =>
+                try
                 {
-                    // Case where code() either
-                    // - succeeded (synchronously or asynchronously)
-                    // - failed asynchronously
-                    var ex = t.IsFaulted ? t.Exception.InnerException : null;
-                    Assert.That(ex, expression, message, args);
-                    return ex;
-                }, TaskScheduler.Default);
-            }
-            // Case where code() failed synchronously
-            catch (Exception ex)
-            {
-                caughtException = ex;
+                    var task = code();
+                    region.WaitForPendingOperationsToComplete(task);
+                }
+                catch (Exception e)
+                {
+                    caughtException = e;
+                }
             }
 
             Assert.That(caughtException, expression, message, args);
 
-            return TaskFromResult(caughtException);
+            return caughtException;
         }
 
         /// <summary>
@@ -73,7 +66,7 @@ namespace NUnit.Framework
         /// </summary>
         /// <param name="expression">A constraint to be satisfied by the exception</param>
         /// <param name="code">A TestSnippet delegate</param>
-        public static Task<Exception> ThrowsAsync(IResolveConstraint expression, AsyncTestDelegate code)
+        public static Exception ThrowsAsync(IResolveConstraint expression, AsyncTestDelegate code)
         {
             return ThrowsAsync(expression, code, string.Empty, null);
         }
@@ -85,7 +78,7 @@ namespace NUnit.Framework
         /// <param name="code">A TestDelegate</param>
         /// <param name="message">The message that will be displayed on failure</param>
         /// <param name="args">Arguments to be used in formatting the message</param>
-        public static Task<Exception> ThrowsAsync(Type expectedExceptionType, AsyncTestDelegate code, string message, params object[] args)
+        public static Exception ThrowsAsync(Type expectedExceptionType, AsyncTestDelegate code, string message, params object[] args)
         {
             return ThrowsAsync(new ExceptionTypeConstraint(expectedExceptionType), code, message, args);
         }
@@ -95,7 +88,7 @@ namespace NUnit.Framework
         /// </summary>
         /// <param name="expectedExceptionType">The exception Type expected</param>
         /// <param name="code">A TestDelegate</param>
-        public static Task<Exception> ThrowsAsync(Type expectedExceptionType, AsyncTestDelegate code)
+        public static Exception ThrowsAsync(Type expectedExceptionType, AsyncTestDelegate code)
         {
             return ThrowsAsync(new ExceptionTypeConstraint(expectedExceptionType), code, string.Empty, null);
         }
@@ -111,9 +104,9 @@ namespace NUnit.Framework
         /// <param name="code">A TestDelegate</param>
         /// <param name="message">The message that will be displayed on failure</param>
         /// <param name="args">Arguments to be used in formatting the message</param>
-        public static Task<TActual> ThrowsAsync<TActual>(AsyncTestDelegate code, string message, params object[] args) where TActual : Exception
+        public static TActual ThrowsAsync<TActual>(AsyncTestDelegate code, string message, params object[] args) where TActual : Exception
         {
-            return Cast<TActual>(ThrowsAsync(typeof (TActual), code, message, args));
+            return (TActual)ThrowsAsync(typeof (TActual), code, message, args);
         }
 
         /// <summary>
@@ -121,7 +114,7 @@ namespace NUnit.Framework
         /// </summary>
         /// <typeparam name="TActual">Type of the expected exception</typeparam>
         /// <param name="code">A TestDelegate</param>
-        public static Task<TActual> ThrowsAsync<TActual>(AsyncTestDelegate code) where TActual : Exception
+        public static TActual ThrowsAsync<TActual>(AsyncTestDelegate code) where TActual : Exception
         {
             return ThrowsAsync<TActual>(code, string.Empty, null);
         }
@@ -137,7 +130,7 @@ namespace NUnit.Framework
         /// <param name="code">A TestDelegate</param>
         /// <param name="message">The message that will be displayed on failure</param>
         /// <param name="args">Arguments to be used in formatting the message</param>
-        public static Task<Exception> CatchAsync(AsyncTestDelegate code, string message, params object[] args)
+        public static Exception CatchAsync(AsyncTestDelegate code, string message, params object[] args)
         {
             return ThrowsAsync(new InstanceOfTypeConstraint(typeof(Exception)), code, message, args);
         }
@@ -147,7 +140,7 @@ namespace NUnit.Framework
         /// and returns it.
         /// </summary>
         /// <param name="code">A TestDelegate</param>
-        public static Task<Exception> CatchAsync(AsyncTestDelegate code)
+        public static Exception CatchAsync(AsyncTestDelegate code)
         {
             return ThrowsAsync(new InstanceOfTypeConstraint(typeof(Exception)), code);
         }
@@ -160,7 +153,7 @@ namespace NUnit.Framework
         /// <param name="code">A TestDelegate</param>
         /// <param name="message">The message that will be displayed on failure</param>
         /// <param name="args">Arguments to be used in formatting the message</param>
-        public static Task<Exception> CatchAsync(Type expectedExceptionType, AsyncTestDelegate code, string message, params object[] args)
+        public static Exception CatchAsync(Type expectedExceptionType, AsyncTestDelegate code, string message, params object[] args)
         {
             return ThrowsAsync(new InstanceOfTypeConstraint(expectedExceptionType), code, message, args);
         }
@@ -171,7 +164,7 @@ namespace NUnit.Framework
         /// </summary>
         /// <param name="expectedExceptionType">The expected Exception Type</param>
         /// <param name="code">A TestDelegate</param>
-        public static Task<Exception> CatchAsync(Type expectedExceptionType, AsyncTestDelegate code)
+        public static Exception CatchAsync(Type expectedExceptionType, AsyncTestDelegate code)
         {
             return ThrowsAsync(new InstanceOfTypeConstraint(expectedExceptionType), code);
         }
@@ -187,9 +180,9 @@ namespace NUnit.Framework
         /// <param name="code">A TestDelegate</param>
         /// <param name="message">The message that will be displayed on failure</param>
         /// <param name="args">Arguments to be used in formatting the message</param>
-        public static Task<TActual> CatchAsync<TActual>(AsyncTestDelegate code, string message, params object[] args) where TActual : System.Exception
+        public static TActual CatchAsync<TActual>(AsyncTestDelegate code, string message, params object[] args) where TActual : Exception
         {
-            return Cast<TActual>(ThrowsAsync(new InstanceOfTypeConstraint(typeof (TActual)), code, message, args));
+            return (TActual)ThrowsAsync(new InstanceOfTypeConstraint(typeof (TActual)), code, message, args);
         }
 
         /// <summary>
@@ -197,16 +190,14 @@ namespace NUnit.Framework
         /// or one derived from it when called and returns it.
         /// </summary>
         /// <param name="code">A TestDelegate</param>
-        public static Task<TActual> CatchAsync<TActual>(AsyncTestDelegate code) where TActual : System.Exception
+        public static TActual CatchAsync<TActual>(AsyncTestDelegate code) where TActual : Exception
         {
-            return Cast<TActual>(ThrowsAsync(new InstanceOfTypeConstraint(typeof (TActual)), code));
+            return (TActual)ThrowsAsync(new InstanceOfTypeConstraint(typeof (TActual)), code);
         }
 
         #endregion
 
         #region DoesNotThrowAsync
-
-        private static readonly Task CompletedTask = TaskFromResult(0);
 
         /// <summary>
         /// Verifies that an async delegate does not throw an exception
@@ -214,65 +205,17 @@ namespace NUnit.Framework
         /// <param name="code">A TestDelegate</param>
         /// <param name="message">The message that will be displayed on failure</param>
         /// <param name="args">Arguments to be used in formatting the message</param>
-        public static Task DoesNotThrowAsync(AsyncTestDelegate code, string message, params object[] args)
+        public static void DoesNotThrowAsync(AsyncTestDelegate code, string message, params object[] args)
         {
-            Exception caughtException;
-            try
-            {
-                var task = code();
-                return task.ContinueWith(t =>
-                {
-                    // Case where code() either
-                    // - succeeded (synchronously or asynchronously)
-                    // - failed asynchronously
-                    var ex = t.IsFaulted ? t.Exception.InnerException : null;
-                    Assert.That(ex, new ExceptionNotThrownConstraint(), message, args);
-                }, TaskScheduler.Default);
-            }
-            // Case where code() failed synchronously
-            catch (Exception ex)
-            {
-                caughtException = ex;
-            }
-
-            Assert.That(caughtException, new ExceptionNotThrownConstraint(), message, args);
-
-            return CompletedTask;
+            Assert.That(code, new ThrowsNothingConstraint(), message, args);
         }
         /// <summary>
         /// Verifies that an async delegate does not throw an exception.
         /// </summary>
         /// <param name="code">A TestDelegate</param>
-        public static Task DoesNotThrowAsync(AsyncTestDelegate code)
+        public static void DoesNotThrowAsync(AsyncTestDelegate code)
         {
-            return DoesNotThrowAsync(code, string.Empty, null);
-        }
-
-        #endregion
-
-        #region Helpers
-
-        private static Task<T> TaskFromResult<T>(T value)
-        {
-#if NET_4_5 || PORTABLE
-            return Task.FromResult(value);
-#else
-            var tcs = new TaskCompletionSource<T>();
-            tcs.SetResult(value);
-            return tcs.Task;
-
-#endif
-        }
-
-        private static Task<TActual> Cast<TActual>(Task<Exception> task) where TActual : Exception
-        {
-            return task
-                .ContinueWith(t =>
-                {
-                    if (t.IsFaulted)
-                        ExceptionHelper.Rethrow(t.Exception.InnerException);
-                    return (TActual)t.Result;
-                }, TaskScheduler.Default);
+            DoesNotThrowAsync(code, string.Empty, null);
         }
 
         #endregion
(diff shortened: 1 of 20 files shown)


src/NUnitFramework/framework/Assert.Exceptions.Async.cs
```diff
@@ -0,0 +1,280 @@
...
+        /// <param name="code">A TestSnippet delegate</param>
+        /// <param name="message">The message that will be displayed on failure</param>
+        /// <param name="args">Arguments to be used in formatting the message</param>
+        public static Task<Exception> ThrowsAsync(IResolveConstraint expression, AsyncTestDelegate code, string message, params object[] args)
+        {
+            Exception caughtException;
+            try
+            {
+                var task = code();
+                return task.ContinueWith(t =>
```
thomaslevesque: using `ContinueWith` instead of async/await so that it works in 4.0 too

src/NUnitFramework/framework/Assert.Exceptions.Async.cs
```diff
@@ -0,0 +1,281 @@
...
+                {
+                    // Case where code() either
+                    // - succeeded (synchronously or asynchronously)
+                    // - failed asynchronously
+                    var ex = t.IsFaulted ? t.Exception.InnerException : null;
+                    Assert.That(ex, new ThrowsNothingConstraint(), message, args);
```
thomaslevesque: Is there a better way? I had to modify `ThrowsNothingConstraint` to accept an exception as well as a delegate, which is probably not ideal.

src/NUnitFramework/framework/Assert.Exceptions.Async.cs
```diff
@@ -0,0 +1,281 @@
...
+            #endif
+        }
+
+        private static Task<TActual> Cast<TActual>(Task<Exception> task) where TActual : Exception
+        {
+            return task
+                .ContinueWith(t =>
+                {
+                    if (t.IsFaulted)
+                        ExceptionHelper.Rethrow(t.Exception.InnerException);
```
thomaslevesque: Need to rethrow the inner exception explicitly, otherwise it's the `AggregateException` that is rethrown

src/NUnitFramework/framework/Constraints/ThrowsNothingConstraint.cs
```diff
@@ -50,23 +50,9 @@ public override string Description
...
-
-            //try
-            //{
-            //    code();
-            //}
-            //catch (Exception ex)
-            //{
-            //    caughtException = ex;
-            //}
-
```
thomaslevesque: I took the liberty of deleting old commented code, let me know if I should put it back

src/NUnitFramework/framework/Constraints/ThrowsNothingConstraint.cs
```diff
@@ -50,23 +50,9 @@ public override string Description
...
-            caughtException = ThrowsConstraint.ExceptionInterceptor.Intercept(actual);
-
+            caughtException = actual as Exception;
+            if (caughtException == null && actual != null)
+                caughtException = ThrowsConstraint.ExceptionInterceptor.Intercept(actual);
```
thomaslevesque: Here's the change mentioned above. I'm not very satisfied with it, but I didn't see a better way to get the correct description ("No Exception to be thrown") without repeating it.

src/NUnitFramework/framework/Internal/AsyncInvocationRegion.cs
```diff
@@ -40,26 +40,6 @@ internal abstract class AsyncInvocationRegion : IDisposable
...
-                {
-                    PreserveStackTrace = (Action<Exception>)Delegate.CreateDelegate(typeof(Action<Exception>), method);
-                    return;
-                }
-                catch (InvalidOperationException) { }
-            }
-            PreserveStackTrace = _ => { };
-        }
-#endif
-
```
thomaslevesque: Moved to a `ExceptionHelper`

src/NUnitFramework/framework/Internal/AsyncInvocationRegion.cs
```diff
@@ -129,13 +109,7 @@ public override object WaitForPendingOperationsToComplete(object invocationResul
                 catch (TargetInvocationException e)
                 {
                     IList<Exception> innerExceptions = GetAllExceptions(e.InnerException);
-
-#if NET_4_5 || PORTABLE
-                    ExceptionDispatchInfo.Capture(innerExceptions[0]).Throw();
-#elif NET_4_0
-                    PreserveStackTrace(innerExceptions[0]);
-                    throw innerExceptions[0];
-#endif
```
thomaslevesque: Extracted to `ExceptionHelper.Rethrow`

src/NUnitFramework/framework/Internal/ExceptionHelper.cs
```diff
@@ -33,6 +34,41 @@ namespace NUnit.Framework.Internal
...
+            }
+            PreserveStackTrace = _ => { };
+        }
+#endif
+
+        /// <summary>
+        /// Rethrows an exception, preserving its stack trace
+        /// </summary>
+        /// <param name="exception">The exception to rethrow</param>
+        public static void Rethrow(Exception exception)
```
thomaslevesque: Should I make it internal? Actually I'm surprised that the whole class isn't internal, since it's in the `Internal` namespace... but it's probably too late to change that, since it would be a breaking change

src/NUnitFramework/tests/Assertions/AssertThrowsAsyncTests.cs
```diff
@@ -0,0 +1,214 @@
+﻿using System;
+using System.Threading.Tasks;
+using NUnit.TestUtilities;
+
+namespace NUnit.Framework.Tests.Assertions
+{
+    [TestFixture]
+    public class AssertThrowsAsyncTests
+    {
```
thomaslevesque: The tests in this class are equivalent to the ones in `AssertThrowsTests`, but with the async version. Some of them have two variants:
- `*Sync`: to test the case where the test delegate fails synchronously (i.e. throws before yielding control back to the caller)
- `*Async`: to test the case where the test delegate fails asynchronously (i.e. sucessfully returns a task that will fail later)

Since `ThrowsAsync` isn't using async/await, it has to handle those cases separately, so there are separate tests to cover both cases.

src/NUnitFramework/framework/Assert.Exceptions.Async.cs
StephenCleary: `ContinueWith` should always specify a `TaskScheduler` (e.g., `TaskScheduler.Default`).

src/NUnitFramework/framework/Assert.Exceptions.Async.cs
CharliePoole: I don't understand why you needed to do this or why you needed to assert on the exception. The semantics of the original method is to succeed if no exception is thrown and to fail if anything is thrown. The original method was void so Task is the proper choice here (Task is kindof the void of the async world) but you are actually returning a value when it seems you should be simply returning.

src/NUnitFramework/framework/Assert.Exceptions.Async.cs
CharliePoole: I'm not sure about using ContinueWith here. let's get some more eyes on it.

src/NUnitFramework/framework/Constraints/ThrowsNothingConstraint.cs
CharliePoole: Maybe we can do some sort of refactoring. One of us should probably pull this down and look more closely rather than just reviewing in GitHub.

src/NUnitFramework/framework/Internal/AsyncInvocationRegion.cs
CharliePoole: I need to download to see this in context.

src/NUnitFramework/framework/Internal/ExceptionHelper.cs
CharliePoole: For many years the policy of NUnit was to make lots of things public in order to be able to test them. This dates back to when we supported .NET versions where it wasn't possible to expose internals to your tests. the habit continues. Although the class is public, it's not part of any API, it's internal. People who use internal classes take their chances. Changing it would not be breaking but may be inconvenient for some of our tests.

src/NUnitFramework/tests/Assertions/AssertThrowsAsyncTests.cs
CharliePoole: These would break some of our builds and so needs to be protected by an #if statement. 

Note that our general practice is to include framework files in all the framework builds but #if them out where not wanted. That facilitates searching the projects for all the #if statements.

This is the first point where I noticed it but it applies to other files as well.

CharliePoole:
This looks good to me modulo a few questions I pointed out inline.

Our general policy is that two committers on the @nunit/core-team need to OK anything before merging. In this case, more would be better, with at least one of us actually downloading and stepping through the change because it's a significant one. 

@simoneb Since this has to do with async, which was your special creation originally, do you have time to step out of retirement and take a look?

CharliePoole:
@thomaslevesque I noticed your comment about C# 6. Since the framework has to be built for many different platforms, we keep the C# level down to 3.0. Sorry. :-(

In this case, going a little higher would be OK, because it's only built for .NET 4.0 and up. However we should avoid breaking things for folks who use VS2012 to build NUnit.

All of this should admittedly be better documented.

src/NUnitFramework/framework/Assert.Exceptions.Async.cs
thomaslevesque: @StephenCleary, good point, thanks. That's why I don't like `ContinueWith`, I never seem to get it right the first time ;)

@CharliePoole, well, there aren't that many alternatives, as far as I can tell. async/await don't work in .NET 4.0, unless you reference `Microsoft.Bcl.Async`, and we probably don't want to add external dependencies to NUnit...

src/NUnitFramework/framework/Assert.Exceptions.Async.cs
thomaslevesque: @CharliePoole, you're right, I don't need to return the exception, so I should drop the `return ex` part. But I still need an assertion that will produce the right message (i.e. "Expected: No Exception to be thrown/But was: ..."). That's why I used `ThrowsNothingConstraint`. Should I create a specific `Constraint`?

src/NUnitFramework/framework/Constraints/ThrowsNothingConstraint.cs
thomaslevesque: The most obvious refactoring I can think of is to create a new `Constraint` class that would check that the exception is null, and fail with the same message as `ThrowsNothingConstraint` otherwise, but I couldn't think of a good name; what bothers me is that the message wouldn't match what the constraint is actually checking.

src/NUnitFramework/tests/Assertions/AssertThrowsAsyncTests.cs
thomaslevesque: > These would break some of our builds and so needs to be protected by an #if statement.
> 
> Note that our general practice is to include framework files in all the framework builds but #if them out where not wanted. That facilitates searching the projects for all the #if statements.

OK, I thought it would be cleaner to include it only in the 4.5 test project, but I can change it.
BTW, I didn't include these tests in nunit.framework.tests-4.0, because the tests would be much harder to write without async/await support. Is this an issue? An option would be to reference `Microsoft.Bcl.Async` to support async/await in 4.0 (we can't do that in nunit.framework, but I don't think it would be a problem for a test project).

> This is the first point where I noticed it but it applies to other files as well.

I did it for Assert.Exceptions.Async.cs as well, which is only included in nunit.framework-4.0 and 4.5. I'll include it in the other projects as well.

thomaslevesque:
> I noticed your comment about C# 6. Since the framework has to be built for many different platforms, we keep the C# level down to 3.0. Sorry. :-(

You can use C# 6 features and still target .NET 2.0. None of the features of C# 6 require specific framework support (except the conversion of interpolated string to `FormattableString`). It makes no difference to the final user, who can still use the compiled library even if they're still using C# 2.

CharliePoole:
@thomaslevesque Because the PR does not refer to #464 there is no way to link to it from the issue... until now that is. :-) 

A fixes # link in the commit works for auto-closing the issue and inserts a link to the commit rather than the PR.

src/NUnitFramework/tests/Assertions/AssertThrowsAsyncTests.cs
thomaslevesque: > Note that our general practice is to include framework files in all the framework builds but #if them out where not wanted. That facilitates searching the projects for all the #if statements.

Regarding this, wouldn't it be simpler to use a shared project (.shproj)? Basically you have a dummy (not buildable) project that just acts as a container for source files, and you reference it from all the projects that need those files. It achieves the same result as including the same source files in several projects, but is easier to work with, because you only need to add new files to the shared project, instead of manually adding them to each project that needs them.

CharliePoole:
@thomaslevesque Yes, of course you can target .NET 2.0 using C# 6.0 if you use a compiler that supports it. That leaves two problems:

1. There are other platforms, such as CF 3.5 that you cannot target that way. The latest VS that will build for that runtime is VS2008, which only supports C# 3.0.

2. Aside from writing tests with NUnit, another thing that users do is build it from scratch. They may want to contribute to it, for example. We selected C# 3.0 because a good selection of compilers and ides support it. One special case of "users" building from source are the packagers who include NUnit in various Linux distributions. We definitely can't give them source that won't be built if we want NUnit to be part of the  distro.

The general rule is to stick with C# 3.0 in working on NUnit. For components that are not built for any of the legacy targets, try to remain compatible with VS2012 on the windows side. And the versions of Mono that are shipped with Debian on the Linux side.

src/NUnitFramework/framework/Assert.Exceptions.Async.cs
CharliePoole: In fact, our support for 4.0 is based on the assumption that the user has installed Microsoft.Bcl.Async. We don't use it ourselves in the framework but we reference it in our tests.

src/NUnitFramework/tests/Assertions/AssertThrowsAsyncTests.cs
CharliePoole: Yes, it would. But shared projects aren't supported in all the compilers and ides we support, most notably in MonoDevelop. I'm afraid there are a lot of things we do that would be much easy if we abandoned support for legacy runtimes and older IDEs. We could do that, but then we'd have to call ourselves Microsoft. :-)

thomaslevesque:
As discussed, I made the changes to make `ThrowsAsync` and friends synchronous. The code and the tests are now much more similar to those of `Throws`.

src/NUnitFramework/framework/Constraints/ThrowsConstraint.cs
```diff
@@ -197,22 +196,36 @@ private static IInvocationDescriptor GetInvocationDescriptor(object actual)
                 {
                     var testDelegate = actual as TestDelegate;
 
-                    if (testDelegate == null)
-                        throw new ArgumentException(
-                            String.Format("The actual value must be a TestDelegate or ActualValueDelegate but was {0}",
-                                actual.GetType().Name),
-                            "actual");
```
thomaslevesque: I had to restructure the flow in order to support both TestDelegate and AsyncTestDelegate. Unless I made a mistake, the overall logic should be the same as before.

src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
```diff
@@ -48,21 +49,43 @@ public override string Description
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
             TestDelegate code = actual as TestDelegate;
-            if (code == null)
-                throw new ArgumentException(
-                    string.Format("The actual value must be a TestDelegate but was {0}", actual.GetType().Name), "actual");
-
```
thomaslevesque: Same comment as in ThrowsConstraint.GetInvocationDescriptor

src/NUnitFramework/framework/Constraints/ThrowsConstraint.cs
```diff
@@ -161,10 +161,9 @@ internal static Exception Intercept(object invocation)
                 {
                     using (var region = AsyncInvocationRegion.Create(invocationDescriptor.Delegate))
                     {
-                        object result = invocationDescriptor.Invoke();
-
                         try
                         {
+                            object result = invocationDescriptor.Invoke();
```
thomaslevesque: Moved the `Invoke()` inside the try, because the operation can fail synchronously (i.e. throw before the first await, or before returning a task)

src/NUnitFramework/framework/Internal/ExceptionHelper.cs
rprouse: We have made framework internals visible to the tests, so you should be able to make these internal. If not, then we should look at why. I would like to get to a point where we are exposing fewer and fewer internal classes as public to prevent people from using them incorrectly or even just depending on classes that may change.

rprouse:
This looks good on my first pass. Like other people, using ThrowsAsync but not being awaitable makes me uncomfortable, but I am not sure what the alternatives are. As has been discussed, we are sync over async. Maybe that was the wrong decision for the throws calls, but it is what people are used to now.

Like @CharliePoole, I think I would like to pull this down and step through your unit tests to get a better feel for it. It is a large change and I can code review this better in the debugger :smile: 

The AppVeyor build looks like it just stopped part way through running the tests. Hopefully that doesn't mean the new tests blocked in some way. I have restarted them.

src/NUnitFramework/framework/Internal/ExceptionHelper.cs
thomaslevesque: Making `ExceptionHelper` internal works fine, but it will probably break someone's code. Admittedly, they shouldn't be using something that is in a namespace named "Internal", but:
- as soon as something is made public, you can't really blame people for using it
- it might not even be intentional. Tools like ReSharper can automatically import missing namespaces. If the user referenced a class or an extension method with the same name as one from the `Internal` namespace and used this feature, R# could have "accidentally" imported the `Internal` namespace. It happened to me a few times (not sure if it was with NUnit or something else).

thomaslevesque:
> The AppVeyor build looks like it just stopped part way through running the tests. Hopefully that doesn't mean the new tests blocked in some way. I have restarted them.

It seems to be blocked again... Not sure why; the tests ran fine on my computer

thomaslevesque:
> It seems to be blocked again... Not sure why; the tests ran fine on my computer

Actually, the 4.5 tests run fine, but the 4.0 tests never complete. I'm looking into it now.

thomaslevesque:
> The AppVeyor build looks like it just stopped part way through running the tests. Hopefully that doesn't mean the new tests blocked in some way. I have restarted them.

I was using a timer to simulate Task.Delay in 4.0, but it was GCed before it was elapsed, so the task was never marked completed and blocked forever... It should be fixed now.

thomaslevesque:
@rprouse AppVeyor won't build the last commit because the previous build is stuck... could you please abort it?

rprouse:
It is passing now. I will pull the code and review it when I get a chance. The holidays are starting though...

src/NUnitFramework/framework/Internal/ExceptionHelper.cs
CharliePoole: @thomaslevesque While I won't blame anybody for using it, I will warn them that it's subject to change. I've been doing it for so long that I tend to assume everyone knows it, but of course they don't. Shifting such things to internal is a good move. If it breaks something, at least it will break it in an obvious way, which is much better than what happens if we make a subtle change to the semantics of a method.

We have always stressed that there is a difference between public methods and public APIs. We'll continue to do that and try to give it better documentation where it's necessary.

rprouse:
Sorry, this fell through the cracks. @thomaslevesque I looked through your changes and they seem okay, but looking at it, I don't understand why Assert.That(TestDelegate, Throws...) worked before your changes? Can you comment on that?

CharliePoole:
I have no more comments beyond my original ones. It looks ready to merge to me.

thomaslevesque:
@rprouse, I think this is because `Assert.That` accepts a `Func<Task>` and waits for the task to complete (not on my dev PC right now, so I can't check, but I'll have a look later today).

@CharliePoole, great, but I'll have to rebase it first

CharliePoole:
GitHub says it can be merged.

CharliePoole:
When @rprouse is ready that is. 

@thomaslevesque We follow a rule of two committers for merging.

thomaslevesque:
> GitHub says it can be merged.

As you like, but I don't mind rebasing, and it makes the history cleaner ;)

CharliePoole:
Feel free... especially given how long it's been hanging around. :-)

thomaslevesque:
@rprouse, I just checked, it works because `Assert.That` has an overload that takes an `ActualValueDelegate<TValue>`. If you pass a lambda that returns a `Task`, it matches `ActualValueDelegate<Task>`. `Assert.That` then calls `Constraint.ApplyTo`, which explicitly waits for task completion if the method returns a `Task`.

CharliePoole:
@thomaslevesque Would it be possible to create an overload for Assert.Throws that works in a similar way, avoiding adding a separate name?

@everyone Would it be desirable?

thomaslevesque:
> @thomaslevesque Would it be possible to create an overload for Assert.Throws that works in a similar way, avoiding adding a separate name?

It would be possible, but it would break a (I think) pretty common scenario where you pass a method group to `Assert.Throws` (see [this comment](https://github.com/nunit/nunit/issues/464#issuecomment-165383715)). Basically, if you pass a `void` method, overload resolution would pick the overload that takes a `ActualValueDelegate<TActual>`, and then realize the method is not compatible....

CharliePoole:
Fair enough, let's leave it as you have it.

rprouse:
@thomaslevesque, thanks, that explains it. I was thinking the same as @CharliePoole that we could provide the same overload if there was one, but as you said, won't work. Don't worry about rebasing, I am going to merge. Thanks for the contribution. I believe this also fixes #1190

src/NUnitFramework/framework/Internal/ExceptionHelper.cs
CharliePoole: What I missed in  reviewing this is that it broke the compact framework build. Oddly (?) nobody noticed until I just came to make another change. No need to worry about it particularly or even to build for CF yourself. Just bare in mind that NETCF is one of our builds when you are working with #if statements. I ended up removing this entire method and the following from NETCF.

thomaslevesque: Sorry about that... I used the solution that didn't include the NETCF project, so I didn't notice...

CharliePoole: No problem, but thought I'd let you know. NETCF can only be built using VS2008, so it's a separate solution. You aren't the first to break it. We are talking about splitting it off in to a separate project and no longer sharing code with the other framework builds.

PR #1158, opened 2015-12-26
Clean up for Issue 922

CharliePoole:
A few minor items...

 * Added NUnitlite-runner to CF project
 * Fixed Linux solution
 * Removed an unused directory and it's contents


rprouse:
Merged

PR #1164, opened 2015-12-28
#1018 Throw a FileNotFoundException *before* the tests get run.

Oipo:
I haven't figured out an easy way to not display the output of the new console within a test yet. But at least this makes for a much clearer error message when executing something without the proper files.

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -96,6 +96,9 @@ namespace NUnit.ConsoleRunner
 
             DisplayTestFiles();
 
+			// Throw a FileNotFoundException if any input file is missing.
+			CheckExistenceOfInputFiles(_options);
+
             TestPackage package = MakeTestPackage(_options);
 
             // We display the filters at this point so  that any exception message
@@ -368,6 +371,17 @@ namespace NUnit.ConsoleRunner
 
             return builder.GetFilter();
         }
+
+		private void CheckExistenceOfInputFiles(ConsoleOptions options)
+		{
+			var workDirectory = options.WorkDirectory ?? Environment.CurrentDirectory;
+
+			foreach (string filepath in options.InputFiles)
+			{
+				if (!File.Exists(Path.Combine(workDirectory, filepath)))
+					throw new FileNotFoundException(String.Format("Input file \"{0}\" not found. Did you specify the correct filename?", filepath));
+			}
+		}
 
         #endregion
 
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -96,9 +96,6 @@ namespace NUnit.ConsoleRunner
 
             DisplayTestFiles();
 
-			// Throw a FileNotFoundException if any input file is missing.
-			CheckExistenceOfInputFiles(_options);
-
             TestPackage package = MakeTestPackage(_options);
 
             // We display the filters at this point so  that any exception message
@@ -371,17 +368,6 @@ namespace NUnit.ConsoleRunner
 
             return builder.GetFilter();
         }
-
-		private void CheckExistenceOfInputFiles(ConsoleOptions options)
-		{
-			var workDirectory = options.WorkDirectory ?? Environment.CurrentDirectory;
-
-			foreach (string filepath in options.InputFiles)
-			{
-				if (!File.Exists(Path.Combine(workDirectory, filepath)))
-					throw new FileNotFoundException(String.Format("Input file \"{0}\" not found. Did you specify the correct filename?", filepath));
-			}
-		}
 
         #endregion
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # NUnit 3.0 #
 
-[![Build Status](https://travis-ci.org/nunit/nunit.svg?branch=master)](https://travis-ci.org/nunit/nunit)
+[![AppVeyor Build status](https://ci.appveyor.com/api/projects/status/3xfkxtnkrts1x06q/branch/master?svg=true)](https://ci.appveyor.com/project/CharliePoole/nunit/branch/master) [![Travis Build Status](https://travis-ci.org/nunit/nunit.svg?branch=master)](https://travis-ci.org/nunit/nunit) [![NuGet Downloads](https://img.shields.io/nuget/dt/NUnit.svg)](https://www.nuget.org/packages/NUnit/) [![NuGet Version](https://img.shields.io/nuget/v/NUnit.svg)](https://www.nuget.org/packages/NUnit/) 
+
+[![Follow NUnit](https://img.shields.io/twitter/follow/nunit.svg?style=social)](https://twitter.com/nunit)
 
 NUnit is a unit-testing framework for all .Net languages. Initially ported from JUnit, the current production release, version 3.0, has been completely rewritten with many new features and support for a wide range of .NET platforms.
 
@@ -9,8 +11,6 @@ NUnit is a unit-testing framework for all .Net languages. Initially ported from 
 NUnit is Open Source software and NUnit 3.0 is released under the [MIT license](http://www.nunit.org/nuget/nunit3-license.txt). Earlier releases used the [NUnit license](http://www.nunit.org/nuget/license.html). Both of these licenses allow the use of NUnit in free and commercial applications and libraries without restrictions.
 
 ## Contributors ##
-
-NUnit 3.0 was created by Charlie Poole, Rob Prouse, Simone Busoli, Neil Colvin and numerous community contributors.
 
 NUnit 3.0 was created by [Charlie Poole](https://github.com/CharliePoole), [Rob Prouse](https://github.com/rprouse), [Simone Busoli](https://github.com/simoneb), [Neil Colvin](https://github.con/oznetmaster) and numerous community contributors. A complete list of contributors since NUnit migrated to GitHub can be [found on GitHub](https://github.com/nunit/nunit/graphs/contributors).
 
--- a/src/NUnitConsole/nunit3-console.tests/ConsoleRunnerTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/ConsoleRunnerTests.cs
@@ -1,0 +1,52 @@
+﻿// ***********************************************************************
+// Copyright (c) 2015 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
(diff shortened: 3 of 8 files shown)


CharliePoole:
Both CI builds had trouble communicating with NuGet. I restarted them.

You have picked a pretty tricky issue to work on, one that requires a fair amount of understanding of the internals of the engine. If it's an issue that's giving you pain and you still want to work on it, I'll be happy to give you some ideas. See my inline comments.

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -96,6 +96,9 @@ public int Execute()
 
             DisplayTestFiles();
 
+			// Throw a FileNotFoundException if any input file is missing.
+			CheckExistenceOfInputFiles(_options);
+
```
CharliePoole: As explained above, we don't want to do this. In fact, this is very much like the code we removed from NUnit 2 because we wanted to run any files we could run even if one turns out to be bad.

Oipo: Ah, right. Would a filter then be acceptable?

Oipo:
Sure, I'd like to keep working on it. What are your thoughts?

src/NUnitConsole/nunit3-console/ConsoleRunner.cs
CharliePoole: I'm not sure how a filter would fit here. The runner doesn't know which files are runnable or even which exist. Although it's not in the current version, it may be asking the engine to kick off tests on some other machine. All it has is a string that may represent an assembly, may be a project containing assemblies or may be invalid. It has to wait to be told.

CharliePoole:
I think the first thing is to figure out exactly when this happens. If I remember correctly from the discussion with Rob, a bad assembly name on the command-line is correctly handled but a bad assembly inside an NUnit or other project file is not. Does that match your experience? Are there other cases?

I think we have pretty much ruled out the business about having a 2.x assembly in the bin directory as being the real cause of this. Do you agree?

Oipo:
As far as I have observed, anything that triggers a wrong version inside TestAgency::LaunchAgentProcess() causes #1018.

The way that I personally triggered it, is because somehow, when one of the input files given does not exist, the targetRuntime.ClrVersion ends up being "0.0". Using the ToString(3) method then, obviously, causes an exception. I guess it'd be better to find out where the "0.0" comes from, but it's getting a bit late today, I'll continue searching for that tomorrow.

Oipo:
So I see two things at this moment:
1) Make sure that "0.0" does not come up with missing files
2) If it does *somehow* come up, throw a nice internal error or so?

CharliePoole:
Best thing would be to come up with a simple repro where we both see the same result.

Regarding internal error - I'd rather detect why this is happening and make it not happen. In principal, TestAgency should not even be called for a non-existent assembly, since the test-suite representing the assembly should be marked not-runnable in the first place.

That gives me an idea... does this happen when you use settings that tell NUnit to run several assemblies together in the same process and one of them does not exist?

Anyway, we need a repro. It doesn't matter what we start with, I think we'll end up finding the corner cases as we go.

CharliePoole:
OK, here's a repro...

nunit.engine.tests.dll is an existing test file
junk.dll does not exist.

nunit3-console nunit.engine.tests.dll 
    runs the tests successfully

nunit3-console junk.dll 
    runs to termination and displays the correct error

nunit3-console nunit.engine.tests.dll junk.dll
    runs to termination running the tests in nunit.engine.tests.dll and showing that junk.dll was not found

nunit3-console nunit.engine.tests.dll junk.dll --process:Single
    aborts the test run with an exception

Oipo:
What platform and stack are you running? I'm running Mono 4.2.1 on Linux.

nunit3-console nunit.engine.tests.dll
runs the tests successfully

nunit3-console junk.dll
Gives the error from #1018 (this apparently differs from your run)

nunit3-console nunit.engine.tests.dll junk.dll
Does not run ANY test and gives the error from #1018 (this also differs from your run)

nunit3-console nunit.engine.tests.dll junk.dll --process:Single
Aborts the tests with a missing file exception.

CharliePoole:
I tested on Windows without thinking about it. I'll retest on Linux.

Oipo:
After some testing, hacking nunit to accept mono-2.0 and mono-4.5 frameworks on windows, produces the correct error (could not load file) but continues to hang. Correct behaviour is observed using the net-2.0 and net-4.5 frameworks on windows.

Using the mono-2.0 framework on linux causes the hang. Using mono-4.5 on linux causes the correct behaviour.

CharliePoole:
@Oipo Good info, thanks. I'll be at a linux machine later today and check further.

I'd be curious to know what you had to do to make nunit run on windows under mono. I haven't actually tried that for quite a while.

Oipo:
Well, the current way nunit tries to detect a mono framework is by looking at the HKLM\Software\Novell\Mono\DefaultClr registry key. But that entire path is non-existent on my Windows 10 machine. So I just hardcoded the monoprefix variable to C:\Program Files (x86)\Mono. I am currently unaware of a fool-proof method of detecting where mono is installed on Windows.

Oipo:
Is there a way to chat with you in a less round-about way? I've further narrowed down where my issue is coming from, but I've now hit an obstacle with which I do not know what the intention is behind the code.

CharliePoole:
In a pinch, you can skype me (I'll send you my handle separately) but actually doing it right here has the advantage that the whole team sees it and can offer ideas and we keep a record of what we decided to do. It could be we need a group chat - something like Gitter - but we haven't set it up yet.

If I keep this window open, it should work pretty quickly, like IM.

CharliePoole:
Also, I'm pretty sure we can communicate better once I actually run the tests in the correct environment. :-)

Oipo:
Ah, in that case, I'll just stick to here:

RuntimeFrameworkService.SelectRuntimeFramework is where my problem lies. When not having a requested framework, the requestedFramework.FrameworkVersion has a value of 0.0. This should not be a problem, since in the explicit check for that version a little bit down there's another check for all the packages. No packages are found (junk.dll, remember?) which is also fine, we probably want the current CLR version. But here's the kicker: checkFramework.IsAvailable returns true with a version of 0.0! This is because on the default requestedFramework, any version will do. And thus, the version returned is 0.0 and thus the .ToString(3) returns a fieldCount exception.

Now the easy fix would be to check if the temporary variable targetVersion is still 0.0 after the packages check, and if so, skip the IsAvailable check altogether and just set the targetVersion to the currentFramework.FrameworkVersion. I'm just not sure if that's what you guys want.

CharliePoole:
I see the problem and I can see why you ask what to do since there are conflicting concerns here. Let me spend some time in the debugger with this and I'll post back.

Oipo:
Thanks for your patience though :)

CharliePoole:
OK, now I see the error at least. :-)

FYI, I'm switching between two different projects and two different machines and OS's. Initially, I forgot that this was a mono issue only. Now I've got it straight with the linux machine for this issue and the windows one for the nunit-gui work. :-)

CharliePoole:
Here's what I see happening...

When we enter the select statement in LaunchAgentProcess, targetRuntime is {net} under .NET and {mono} under Mono, so different case statements are executed. In the .NET case, we have these lines:

```C#
   if (targetRuntime.ClrVersion.Build < 0)
      targetRuntime = RuntimeFramework.GetBestAvailableFramework(targetRuntime);
```

It looks as if this code was inserted right before the user of Version.ToString(3) to avoid exactly the error we are seeing in Mono. An obvious fix is to replicate the code for mono or to simply make the substitution before the switch statement.

However, at a quick glance, I'm a bit suspicious of GetBestAvailableFramework. Could you check that out as well? I don't think it was really written anticipating that the version might be 0.0.

Oipo:
Since it's almost the end of this year, it'll have to wait until next year. Bit busy with the festive stuff :)

CharliePoole:
Not a problem. Happy new year!

Oipo:
Happy new year!

GetBestAvailableFramework does its job just fine, even with 0.0. It's just odd that it selects a runtime and then in the case of an error of that selection, there's another function to fix it. I think I'd solve it by changing RuntimeFramework.Supports(...) by checking for build < 0 and returning false if that is the case. This fixes the RuntimeFramework.IsAvailable property, making it do what the function name says it would do.

CharliePoole:
@Oipo Have you had a chance to get back to this?

Oipo:
Oh, sorry. I kinda was waiting for your input again and then kinda forgot to send you a notification. Is it ok if I update RuntimeFramework.Supports()? Or would you rather I simply duplicate the line of code for .NET into the MONO part as you mentioned before?

CharliePoole:
We definitely need those two lines. You mentioned that the RuntimeFrameworkSupports method  could do with some refactoring, but I'm not entirely sure what you would change. Seeing the code would be easier for me, but harder for you if you end up having to back it out. :-)

Oipo:
This is how I intended to solve it.

CharliePoole:
FYI, it's good to merge in master periodically when working on a problem for a period of time, but you don't have to do it every time you change your branch. The eventual merge process takes care of that.

Oipo:
@CharliePoole Thanks, but do you see where I'm going with my latest change and more importantly, what the ramifications of it would be?

CharliePoole:
Odd..your newest commit was not visible to me when I commented last, only the merge. Now it's visible. I'll review when I get back to my computer.

CharliePoole:
I see you are making a fairly deep change to RuntimeFramework, which is used extensively throughout NUnit. I think that the particular change would break some of the other usage. The key question in this particular usage is "Why is build < 0?" when it should not be? The answer seems to be in those two missing lines - which were apparently added to fix the exact problem for .NET.

The user has asked for any version of mono (mono-0.0). Or, when running mono, didn't ask for anything, which translates to asking for mono-0.0. This is the same as asking for mono-4.0 or net-4.5 in that the build is not specified. So we have to find the best available fit to the request, which is done by calling GetBestAvailableFramework.

Additionally, you have not yet made changes to reflect comments on the earlier commits, so we have a mix of stuff in the branch. Take a look at the third tab in GitHub and you'll see what I mean.

Oipo:
I have made those changes now. Honestly, I still think it's not a good solution, but I do not have the time nor desire to continue on this particular issue.

rprouse:
@CharliePoole, @Oipo has stated that he does not want to continue with this PR. Are you interested in taking this on? I am planning on closing it if not.

CharliePoole:
I'll take it and clean up. @Oipo is right - the fix I suggested is not very satisfying, but given the current state of the code, it's necessary. We need some deeper analysis than I really have time to do right now. The entire chain of events leading down to launching the new process has to be reviewed at some point.

@Oipo Thanks for all your work and for helping us figure out as much as we have!

@rprouse Leave the PR for now. I'll delete it when I get what I need from it.

Issue #1138, opened 2015-12-14
Extend the engine to run portable tests


rprouse:
This needs design, but could be platform specific agents. If so, it will likely be blocked by #362.

rprouse:
In #1168, @CharliePoole said,

>I'm not sure there is a problem here since this refers (in my mind anyway) to an engine running on the desktop. What we need to find out is whether an engine running under .NET 2.0/3.5 is able to analyze assemblies built for other targets. I don't see why not, since they are not actually loaded. Consider that we are already analyzing .NET 4.5 assemblies. Of course, this needs to be tested on a VM that has only .NET 2.0 or 3.5 installed.

Based on my spikes into this so far, I think this is the way we need to go. The engine can inspect the test assembly and determine it's target platform. We just can't run it directly.

This is also mixed up with #677 and #362. My idea so far is as follows,

- The current NUnit3 engine inspects the test assembly in addition to the framework. It looks for the `TargetFrameworkAttribute` which will tell you for example if it is targeting `.NETCore,Version=v5.0`.
- If the target framework is a non-desktop framework, disallow `--inprocess` since we can't run non-desktop targets. We might also want to disallow `--process=Single` or we make sure all test assemblies target the same platform.
- Modify `TestAgency` to allow it to launch platform specific agents and communicate with them in a manner other than .NET remoting. This is #362. If we are running those agents on the desktop machine, we could even capture Console.In/Out for communications.

The platform specific agents are where decisions still need to be made. It would be nice to have a portable mini-engine (#677) that is capable of loading and running the portable framework. Like the full framework, this would allow us to run test no matter which version of the NUnit 3 framework is used. It would likely be a stripped down version of the engine that can only load from the agent directory and is only intended to be used by agents. This might even boil down to a portable `IFrameworkDriver` and some supporting classes?

A simplified and not mutually exclusive approach might be that the agents are actually the nunitlite version of the tests. We could modify nunitlite to run directly, or be able to communicate with the engine.

CharliePoole:
I think we're pretty much on the same page at a high level. Some details...

* Can we count on TargetFrameworkAttribute being reliably set?

* We can't precisely disallow all those options, although we might be able to in some cases. Basically, we can only look at  explicit options, but if the defaults are used, we have to wait until we analyze the assembly and then reflect an error back to the runner.

[...]

Issue #677, opened 2015-06-07
Portable NUnit "Engine"


CharliePoole:
We need a portable engine for use with the portable framework. It's not clear if this component will really be an "engine" since it's functionality is so reduced, but that's the working title for now.

There are a number of big issues with a portable engine:

1. It isn't possible to open test assemblies as files, so we can only run in the NUnitLite style, where the component runs tests from the assembly that calls it.

2. There is no AppDomain creation, so tests must load and run wherever the engine is already running.

3. XmlNode is not available for return values or parameters, so a new interface is needed.

4. Engine Services as currently defined are not needed and many are not possible to implement.

CharliePoole:
After some review of the code, it's clear we need less than a full engine - much less- but also more than just an engine. The portable framework build doesn't include FrameworkController, so it's not possible to call into the framework without having a direct reference.

FrameworkController will need to be included in the portable build in some modified form.

@rprouse As the "Customer" for this task, let me know what you need in addition to what you are now using. As far as I can see, you are using two methods of the ITestAssemblyRunner interface.

1. You use the form of Load that takes an assembly rather than a file name. This doesn't currently have an equivalent in FrameworkController, so it will have to be added.

2. You use Run without any listener or filter.

Have I missed anything? Any ideas for further development?

Charlie

rprouse:
Charlie,

At the airport and about to board,  so I am not looking at the code.  Your
evaluation sounds pretty much right though.  On mobile,  I might not have
access to the file,  but I can always get the assembly,  so I need to load
using assembly.

I currently don't use filters,  but I may decide to in the future.
Eventually,  I might like to explore tests and allow the user to run
specific tests.

I don't think I need a listener.  I found it easier to take all of the test
results,  flatten them then create the control heirarchy from that.

That might be a problem for longer test runs though since I don't get the
results until the end.  Maybe switching to a listener would be better?

Boarding now,  I will read through the code on the flight if I get a
chance.

Rob

CharliePoole:
I think the key question is whether we will need to open assemblies with tests other than the original assembly that calls into the runner. In fact, I think that this is the main distinction between an nunitlite-style driver/runner and an engine/runner.

However, it turns out this is a can of worms, so I'm temporarily putting this issue on hold in order to work on getting the FrameworkController class to work in a portable library. It turns out that it all revolves on what facilities are available for parsing and manipulating XML. I'll describe it in detail in the new issue I'm creating.

rprouse:
[...]

Issue #362, opened 2014-11-19
Replace Remoting with a Portable Networking Protocol


rprouse:
NUnit currently uses .NET Remoting to communicate between the runner and agents. .NET Remoting is not supported by portable libraries or Silverlight, so it would be good to move to form of network communications that is supported by all platforms.

One minor complication is that NUnit makes use of client activated objects. These are effectively two-way communications between the client and the server. We will need to simulate this with some sort of polling by the agents. I see this working like,
1. Runner launches an agent with the URI to connect back to
2. Agent starts, connects back to the runner and registers
3. Agent polls the runner for work (Run, Explore, etc) and performs any work received. Polling continues until the Stop event is received.
4. Agent reports the results of all work back to the runner.

This is not radically different than the current communications. The only difference is that all communications are initiated by the client.

This just leaves the choice of which protocol to use for the communications. As far as I know, the options that will work with the portable framework are,

**TcpListener**

This will be the hardest option to implement. It is basically just a raw TCP socket. We will need to handle serialization and routing of all data.

**HttpListener**

This is a much nicer option as it allows us to create a very portable REST like interface. The HttpClient is portable and can be used on any platform. The problem is that unless you open the port using netsh as admin, the runner would require admin privileges to run. This is not acceptable.

There are 3rd Party HttpListener like options, but they would require taking a dependency on an external library. This might not be a serious issue since it is a dependency in the engine, not in the framework.

**WCF**

I am pretty sure that WCF is supported in portable libraries. This is a good option because it is closest to Remoting, but unfortunately it is only supported by .NET 3.0+. I don't think we are willing to drop support for .NET 2.0, so I don't think this is an option.

**Recommendation**

I think that using a 3rd Party HTTP library in the Engine and System.Net.HttpClient for the clients is our best option. It gives us a very portable and flexible networking protocol that is supported by all platforms.

If I have missed options or made a mistake in my assessment of the presented options, please add to the discussion. 

This could be a fairly big change, so I want to get some consensus on the approach before I start.

CharliePoole:
In spite of the earlier discussion, nobody ever assigned it to 3.0 before!

oznetmaster:
I have been working with a third party HTTP/WebSocket client/server package that is published under the MIT license.

 

I have it running on the Crestron system, and hence, CF.  It is designed for .NET 3.5 and up :(

 

It is one possible solution to the portable networking protocol.  And since it supports WebSocket, it can easily be interfaced using HTML5 as the GUI.

[...]

PR #1170, opened 2016-01-04
NUnit Driver for running portable tests

rprouse:
This is part of #677 and #1138. It creates a portable assembly which I call nunit.driver which is capable of loading and running tests in the NUnit Framework without binding to a specific version. Unfortunately, this required additions to FrameworkController (explained below), so it will not work with the already released 3.0.x framework. In that case, I throw an exception explaining that it is unsupported and that you should update your framework.

This assembly is currently **not signed, should it be?** If it is signed, should we use a static 3.0 version like the API so you can drop in updates?

The intention of this driver is to be a portable assembly that will make it very easy to write platform specific agents that can run unit tests for non-desktop framework platforms. It could also be used in the Xamarin runners to break their dependency on a specific framework version. If I do that, I might want to add classes that parse the XML results, create filters, etc. In other words, add a bit more engine functionality.

I had to switch the console and engine test projects to target .NET 4.5. I did this because those projects reference the framework and build into the same directory as my driver. Because my driver is portable, my tests needed to be portable or .NET 4.5, thus the need for the change. If this is a problem, I will need to build the driver into a driver subdirectory.

I also moved all of the compatibility classes from the framework into the Common directory so that we can use them in different projects.

I needed to make the changes to FrameworkController because it passes around ICallbackEventHandler's which are only available on the full framework. I tried passing in our compatibility version, but it crashes with a TypeLoadException since it conflicts with the real version in the non-portable versions of the nunit.framework.

To get around this, I added public Load, Run, Explore, etc methods and load/execute them through reflection. Where I needed a callback, I pass in an Action<string>. This isn't available for .NET 2.0, but this is intended for portable platforms.

If we want to keep the pull requests reasonable for this series of changes, then **this can be merged as is.** Following pull requests will do the following;

- Modify the engine to identify the target platform of tests for non-desktop tests,
- Add to TestAgency or create additional ones to launch platform specific agents, #1138
- Write platform specific agents
- Create additional communications protocols for communicating with the platform specific agents since remoting isn't available #362

If we take this approach, then this PR is likely all we will do for #677


CharliePoole:
I may not get to this today, but I would like to review it.

About the term "driver"... When I use it I am talking explicitly about  the drivers that are extensions to the engine and which know how to run tests for a particular framework. My assumption was that the driver would communicate with the framework, which would have code added to it to enable that. Alternatively, it could be a separate piece of software as you have done.

I've been talking about this for a while, but I guess I never clarified that I meant specifically the drivers that are part of the engine. The design of drivers does not require that there be only one driver per framework. They can map many to one, one to many or many to many, depending on how you decide what driver to use.

There's no need to respond to this comment directly, since I have not yet looked at your code.  But I thought I should clarify what I've been talking about as a "driver." I may change my mind when I see your code, of course. :-)

A further general comment: I was assuming that it would take a new release of both framework and engine to get this to work at all.

rprouse:
Driver may not be the best term, but it acts like the drivers in the engine, so that is what I called it. The difference is that it is a stand-alone driver that does not use any of the facilities of the engine and is intended to be used by agents that cannot load the full engine. I struggled with the name since it might cause confusion, but couldn't come up with anything that described it better. I am open to suggestions though.

As for needing new releases of the engine and framework, that is true. The framework changes have been made and I intend to modify the engine with subsequent pull requests. This seemed like the logical start because it will allow me to start writing agents that the modified engine can execute.

CharliePoole:
I'll review it as soon as I can rather than speculate. :-) I do think that
a consistent set of terms would help people in understanding how this all
works.

rprouse:
Not sure why the addins tests are failing on Linux. I am pretty sure nothing there got touched and they are passing on Windows.

```
TestAddins
Executing task: TestAddins
NUnit Console Runner 3.0.5847 
Runtime Environment
   OS Version: Linux 3.13.0.40 
  CLR Version: 4.0.30319.17020
Test Files
    addins/tests/addin-tests.dll
Missing method .ctor in assembly /home/travis/build/nunit/nunit/bin/Release/nunit.framework.dll, type System.Runtime.Versioning.TargetFrameworkAttribute
Can't find custom attr constructor image: /home/travis/build/nunit/nunit/bin/Release/nunit.framework.dll mtoken: 0x0a000356
```

rprouse:
I am getting quite confused by this Travis error :frowning: 

It only happens on Linux and only with Mono 3.2.8 and only if you build from a fresh source tree using the Cake build script.

My Linux boxes are all upgraded to Mono 4, so I setup a new Ubuntu VM with the default Mono 3.2.8 and cloned the repo. I build using `./build.sh -Target Travis` and ran the command `mono nunit3-console.exe addins/tests/addin-tests.dll` and was able to reproduce.

I then opened monodevelop and built debug, but I could not reproduce. I then switched to the release configuration and debugged from there. It does not happen --inprocess, so I modified the agent to show a messagebox so that I can attach (our debug.break doesn't seem to work on Linux as reported elsewhere), but the act of recompiling in MonoDevelop caused the problem to go away.

I then deleted the Release directory and compiled again using the Cake build script, but I still cannot reproduce. My best guess is that something in the obj directories is being used.

I am running out of ideas. Maybe revert the Cake script and see if that makes a difference? Anyone else have ideas?

rprouse:
- Reverting the Cake build script did not help
- Deleting all of the obj directories and the bin\Release directory then rebuilding from the command line did cause it to happen, so something is getting cached in the obj directories

rprouse:
The Travis problem running the addin tests on Mono 3.2.8 turned out to be because the copy of NUnit Framework in the nunit3-console directory was .NET 4.5 and the version in the test directory was 2.0. I tried, but I could not figure out why it was happening. Since it only happens on Mono 3.2.8, not on Windows or the newer Mono, and because it is not normal to have a copy of the framework in the test directory, and because it is not related to these changes, I stopped looking at it.

@CharliePoole this is ready for review. If you get a chance, I would like to move forward on this soon as it blocks the other engine work I want to do.

CharliePoole:
@rprouse I've been under the weather for the past day but I'll try to give it a review.

Can you explain what you mean by the "test directory?" It would appear to mean the directory containing the tests, except you say it's not normal to have a copy of the framework there.

rprouse:
@CharliePoole I am sorry to hear you are not well. I am just getting over a lengthy cold/flu myself, that is why I have been so quiet since Christmas.

The copy of the framework in the test directory is normal, it is the copy in the nunit3-console directory that is only normal for our tests. I found;

```
if ( running Mono 3.2.x &&
     nunit.framework.dll in console dir >= 4.0 &&
     nunit.framework.dll in test dir == 2.0 )
{
    crash
}
```

It does not happen if there is no nunit.framework.dll in the console dir, if it is 2.0 in the console dir, on any platform other than mono 3.2.x, etc. As far as I can tell, this issue has always been there, but nobody has run into it.

I don't think it is that serious and may be a mono bug since it works in 4.x, but I can enter an issue if you like.

rprouse:
I want to get moving on the engine changes, so I am going to base that work off this branch. I can update or change the name as necessary. @CharliePoole if you get a chance, can you at least take a quick scan of this PR to make sure you don't have any fundamental issues with the approach?

rprouse:
@CharliePoole any chance doing a review on this. In essence, it is the one driver file that needs reviewing.

CharliePoole:
I'm a bit worried about the approach...

1. Nomenclature - we overload lots of terms (like runner, for example) but "driver" is something that we have used very precisely up to now. A driver is an engine component that uses reflection in a defined way to interact with a particular framework. Yours works similarly to a driver but defines it's own API and is definitely not part of the engine. This isn't terrible, but it would be nice if we could avoid creating an overloaded meaning where one does not already exist.

2. The main issue is that you have changed the framework API. Currently, we define that API as a collection of well-known names of objects, which do their work as part of their constructor. You changed that to call a number of methods by reflection. Of course, we could have defined the API that way in the first place but didn't. Instead, I used a pattern that had  been working quite well for xunit.net. I don't quite get the reason for making this change, especially since it breaks compatibility.  Was there something that would not work using the existing API? Could it have been modified simply by adding to the existing API? Or are we saying that certain framework builds will require a differnet API in the future?

rprouse:
1 - I agree about the naming. It also made me a bit uncomfortable. I will try to come up with a better name. I am thinking maybe something like runner combined with agent since this project is intended to hold the code necessary for platform specific agents that run tests using the portable framework.

2 - Sorry, this has been out there awhile :) I needed to add to the FrameworkController because ICallbackEventHandler is only available on desktop. This allows me to pass in an Action instead for portable tests which is something I will know is available on both sides and will also always be the same type on both sides. As I state above, I was getting TypeLoadExceptions in some combinations where the desktop version was conflicting with the compatibility vesion.

I made the changes to the FrameworkController so that they are backwards compatible and should not break existing consumers.

You will also find that XUnit 2 changed their approach. They no longer use controller actions, but make calls to their engine to return runner interfaces. All of their projects (runner and framework) now reference an interfaces only assembly xunit.abstractions. Instead of ICallbackEventHandler, they use IMessageSink from their abstractions assembly which is referenced by their framework and their engine. Like our Engine.API, it has a static version.

Doing the same in NUnit would make things like this easier, but it would be a fundamental change and I don't want to do that.

I also decided to go with direct reflection and execution because it is easier. AppDomain is not available portable or on many of the platforms, so none of the CreateAndUnwrap methods are available. Without them, wrapping the reflection and execution of FrameworkContoller in the to-be-renamed FrameworkDriver class allows me to load and execute tests no matter which version of the framework we load. It won't work with pre-3.1 of course.

rprouse:
I dug a little deeper into the current XUnit code. Their `TestFrameworkProxy` class uses a wrapper for AppDomains. For their portable support, they have a no appdomain version that wraps Activator.Create like I am using.

CharliePoole:
Naming...

I went originally with the idea of drivers in the engine talking to controllers. Maybe my mistake was calling the current controller "FrameworkController" rather than "SomeKindOfFrameworkController."

How's this for a breakdown:

Engine = component that provides runners and other services to a client (which is also called a "runner" unfortunately). It includes the ability to run multiple frameworks and combine the results as well as making decisions as to where and on what platform something is run.

Runner = engine component that runs tests by communicating with a driver or an external agent. Currently, when running in process, we go through a driver but when going out of process we use an agent, which is provided by another engine instance. The latter approach will extend to running on other machines as well.

Driver = engine component that knows how to communicate with and run tests on a specific framework. How it does that will vary from framework to framework and possibly from platform to platform for a given framework. In the case of the current NUnit 3 driver, it uses reflection to communicate with the FrameworkController, which is a part of the framework.

Agent = a less precisely defined component at this point, but it is used to communicate between different executing processes. the current agent serves to provide a remote runner, which communicates with the original process runner. We could consider expanding this to include provision of other kinds of interfaces. For example, there might be a driver agent, which faciitated a local driver talking to a driver "somewhere else". Your new component looks a bit like an agent to me.


I noticed that xunit had that new assembly, but I didn't realize it was such a big change of design. Did they give up compatibility with xunit 1 in the process? Clearly, we can't give up 3.0 compatibility at this point, but we could test for the existence of the new interface if we want to add it generally. Note that we can change the interface of the framework rather simply, if we are willing to provide additional drivers. Each driver recognizes whether it can deal with the particular framework build or not.

Let's go ahead and merge this with some sort of name change. We can continue working on the concepts as we go. My tendency is to start with the engine and drill down, while you have been working "upward" from the portable framework. Hopefully, we'll meet somewhere in the middle. :-)

rprouse:
Based on your definitions above, the `NUnitDriver` class itself does fit into the Driver category except that it is not provided as a service. Maybe rename it to `NUnit3PortableDriver`?

As for the project itself, it is intended to be consumed by portable agents. Maybe `nunit.portable.agent`? Like our current agents, it is intended to provide the code required to run the agent while the platform specific agents only provide the main and possibly platform specific code.

As for working from the portable framework towards the engine, this PR does because I needed to figure out how I would be running the tests. That building block made the rest of the process of how the engine would discover and run portable and non-desktop tests clearer for me. The branch I am working on now is going from the other direction, 

1. determining what platform tests need to be run as
2. refactoring ProcessRunner and TestAgency to be able to launch and communicate with different agents
3. etc :smile: 

As for XUnit, they support XUnit 1 in a similar way that we do. They have a 'driver' that wraps XUnit 1 in an XUnit 2 interface. Their factory creates one or the other.

CharliePoole:
Engine drivers are not services, but they are provided by the DriverService. Looking ahead, I'm imagining that we are running (for example) xamarin tests from the desktop. The gui or console gets a runner from the engine. The engine talks to a driver, which is actually an engine extension. That driver then talks to something else - I guess to your component - possibly through a communications interface. It does not seem that the thing to which a driver talks should also be a driver. :-)

That's what I meant by one of us looking from the engine down and the other from the framework upward - I can see that it makes sense for what you created to be a driver if by driver you mean "the thing that drives a framework" it makes less sense if you mean "a thing that is used by a runner" However, maybe it is a driver and the engine driver will talk to an agent which provides a "local" driver. If you want to keep the names for now, I think it's fine since it's not an interface for users. We'll eventually either keep it or change it.

BTW, in my comments about agents I should have been clearer. I was not talking about nunit-agent.exe but about the internal agent class, which is a concept I started to develop but did not generalize beyond the notion of a remote agent that resides in a different process.

rprouse:
This is ready for another round of reviews and merging.

The projects have been renamed to `nunit.portable.agent` and are now built into an agents subdirectory. The intention is that platform specific agents will be in a subdirectory so that dependencies will not conflict with the engine.

I also renamed the one runner class in the project from `NUnit.Driver` to `NUnitPortableDriver`. It isn't a driver in the classic sense, but it is intended to be an interface to run unit tests like the other drivers.

I also merged in latest from master and resolved conflicts.

CharliePoole:
Looks fine to me once you resolve conflicts.

rprouse:
Damn, just resolved conflicts two days ago. I will do it again then merge.

PR #1174, opened 2016-01-07
TestCase with Explicit=true should not crash console runner

alfeg:
There is a simple test that reproduce problem and a fix.

If TestCase attribute is marked with Explicit true with no Reason message, then whole console runner output is broken due to bug.

### DIFF
--- a/src/NUnitFramework/framework/Interfaces/TNode.cs
+++ b/src/NUnitFramework/framework/Interfaces/TNode.cs
@@ -320,6 +320,8 @@ namespace NUnit.Framework.Interfaces
 
         private static string EscapeInvalidXmlCharacters(string str)
         {
+            if (str == null) return null;
+
             // Based on the XML spec http://www.w3.org/TR/xml/#charsets
             // For detailed explanation of the regex see http://mnaoumov.wordpress.com/2014/06/15/escaping-invalid-xml-unicode-characters/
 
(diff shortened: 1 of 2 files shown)


CharliePoole:
@alfeg Thanks for your contribution. This fixes a problem, for sure, but I am concerned about why the method is even being called with a null argument. I only looked at the code without running it, so I'm not sure if it happens when writing the a property or the reason element. In either case, if the value is null, no element should be created, not even an empty one.

@rprouse I can see going either way on this. I'm fine with merging it to avoid the error in this case and dealing with the underlying problem at a later point. Alternatively, we could try to ferret it out now. I suspect this crept in because the reason is now required for Ignore but not for Explicit.

alfeg:
@CharliePoole, Hi. I have think about this. In XML output there will be empty `reason` node. I don't know is this valid, cause I'm not familiar with nunit code base. Maybe some default message is better than empty node.

But anyway this null check is looks safe enough in this particular method. I've try to make fix as simple as possible.

From my perspective runner should not destroy whole output, just because there is no safeguards against nulls in upper level code. Destroyed output has zero useful information about reasons of fail.

I were looking through blame to find out why my tests are not working on CI after update to nunit3, that's why I leave comment in other PR. Then just because I'm already spent whole day on this issue, so decide to sent a simple fix. codebase.

CharliePoole:
@alfeg I understand. I think this is a good fix as it stands but we might want to do more at some point. In fact, an empty reason element may seem strange to code that uses the XML, but we have had that for a while and it was only the recent fix that caused the problem to surface.

I passed it on to @rprouse because our standard is that two committers have to approve any merge to master.

rprouse:
Looks like a good code change to have. I am going to merge.

Issue #1121, opened 2015-12-08
Usage of field set to null as value source leads to somewhat cryptic error


estrizhok:
Consider the following code snippet.

```csharp
  public class TestClassFour
  {
    private static string sourceFour = "Rumpelstilzchen";
    public static string sourceFive = null;

    [Test]
    public void Four([ValueSource("sourceFour")] char source)
    {
      Assert.AreNotSame("z", source);
    }

    [Test]
    public void Five([ValueSource("sourceFive")] string source)
    {
      Assert.AreEqual("Argentina", source);
    }
  }
```
If one attempts to run tests from such class, he will encounter the following issues:

1. An error being reported is not user friendly enough.

> An exception was thrown while loading the test.
> System.NullReferenceException: Object reference not set to an instance of an object.
>   at NUnit.Framework.Internal.Builders.ParameterDataProvider.GetDataFor(IParameterInfo parameter)
>   at NUnit.Framework.CombiningStrategyAttribute.BuildFrom(IMethodInfo method, Test suite)
>   at NUnit.Framework.Internal.Builders.DefaultTestCaseBuilder.BuildFrom(IMethodInfo method, Test parentSuite)
>   at NUnit.Framework.Internal.Builders.NUnitTestFixtureBuilder.BuildTestCase(IMethodInfo method, TestSuite suite)
>   at NUnit.Framework.Internal.Builders.NUnitTestFixtureBuilder.AddTestCasesToFixture(TestFixture fixture)
>   at NUnit.Framework.Internal.Builders.NUnitTestFixtureBuilder.BuildFrom(ITypeInfo typeInfo)
>   at NUnit.Framework.Internal.Builders.DefaultSuiteBuilder.BuildFrom(ITypeInfo typeInfo)

2. Test cases from perfectly valid value source are not discovered and cannot be run until the error is fixed.

rprouse:
Thanks for the report. It looks pretty serious, but should be fairly easy to fix.

rprouse:
Fixed by PR #1184

PR #1184, opened 2016-01-09
#1121 : ValueSourceAttribute.GetDataSource should throw instead of return null

alfeg:
My attempt to fix bug #1121.

### DIFF
--- a/src/NUnitFramework/tests/Attributes/ValueSourceTests.cs
+++ b/src/NUnitFramework/tests/Attributes/ValueSourceTests.cs
@@ -24,7 +24,11 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
+using NUnit.Framework.Compatibility;
 using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+using NUnit.Framework.Internal.Builders;
 using NUnit.TestUtilities;
 
 namespace NUnit.Framework.Attributes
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/tests/Attributes/ValueSourceTests.cs
```diff
@@ -168,6 +172,52 @@ public static IEnumerable<int> IntegerProvider()
...
+            var parameters = method.GetParameters();
+
+            foreach (var parameter in parameters)
+            {
+                var dataSource = parameter.GetCustomAttributes<IParameterDataSource>(false)[0];
+
+                Assert.Throws<InvalidDataSourceException>(() =>
+                {
+                    var data = dataSource.GetData(parameter);
+                    Assert.IsNotNull(data);
```
rprouse: I am assuming that the GetData is the line that throws, so is the Assert.IsNotNull necessary?

rprouse:
Very minor comment on my part. This looks good, it just needs a second team member to review.

@CharliePoole, this is easiest to review ignoring whitespace with `?w=1`

alfeg:
I'm not sure that those fix is enough. 

It's still easy to break test suite with non-enumerable values. Like:

``` csharp
public static int NonEnumerable = 1
```

```
1) Invalid : NUnit.Framework.Attributes.ValueSourceTests
System.InvalidCastException: Unable to cast object of type 'System.Int32' to type 'System.Collections.IEnumerable'.
   at NUnit.Framework.ValueSourceAttribute.GetDataSourceValue(MemberInfo[] members) in D:\tmp\nunit\src\NUnitFramework\framework\Attributes\ValueSourceAttribute.cs:line 131
   at NUnit.Framework.ValueSourceAttribute.GetDataSource(IParameterInfo parameter) in D:\tmp\nunit\src\NUnitFramework\framework\Attributes\ValueSourceAttribute.cs:line 111
   at NUnit.Framework.ValueSourceAttribute.GetData(IParameterInfo parameter) in D:\tmp\nunit\src\NUnitFramework\framework\Attributes\ValueSourceAttribute.cs:line 93
   at NUnit.Framework.Internal.Builders.ParameterDataSourceProvider.GetDataFor(IParameterInfo parameter) in D:\tmp\nunit\src\NUnitFramework\framework\Internal\Builders\ParameterDat
aSourceProvider.cs:line 67
   at NUnit.Framework.CombiningStrategyAttribute.BuildFrom(IMethodInfo method, Test suite) in D:\tmp\nunit\src\NUnitFramework\framework\Attributes\CombiningStrategyAttribute.cs:lin
e 142
```

Maybe there is also should be a check for object is IEnumerable?

rprouse:
I'm not sure we need to go so far as to check that the source is IEnumerable. If you do, you will just display an error. Personally, I think the error you show above is valid since it is an invalid test and therefore should stop the test run. In issue #1121, we want to carry on with the test run because a null or empty collection could be valid depending on where the data is coming from. The example you give above is just an invalid test that can never work.

rprouse:
This has been out there too long. I am going to merge this before merge conflicts start creeping in. It is a small and well tested change.

Thanks for the contribution.

Issue #1196, opened 2016-01-15
Custom value formatter for v3 via MsgUtils


et1975:
I'm trying to implement custom value formatting for fsprojects/FsUnit#82 and looking at MsgUtils that seems to be the place to get it working for all constraints.
I'd like to be able to inject a custom formatter that, if present, would be invoked from FormatValue(); something like:
```
    public static Func<object,string> CustomValueFormatter;
``` 

Edit: just noticed MsgUtils is internal... looking for alternatives then.

rprouse:
Just because MsgUtils is internal doesn't mean we couldn't provide a method for you to inject a formatter. Can you explain more about how you would want to reformat messages and why it is needed for FsUnit?

One thing that you need to be aware of is that NUnit needs to compile back to .NET 2.0, so you won't be able to use a Func, but will need to create a new delegate.

I am tagging this as an idea for discussion and design.

et1975:
F# has a few types (records, tuples, discriminated unions) that from .NET perspective are just objects. These types however have nice support when F#'s own [printf mudule functions](https://msdn.microsoft.com/en-us/library/ee370560.aspx) are used to format the values.

So, for example, asserting the equality of two records of type MyRecord (all of the above mentioned types are equattable in F#, btw) with default .NET formatting will tell me something like "[MyRecord] is not equal to [MyRecord]", while formatted with printf will be a lot more informative: "{field1:10} is not equal to {field1:20}".

In general though, you'd be able to address large group of "this format is not to my liking" scenarios, like #1004 with a pluggable formatter.

rprouse:
That sounds reasonable. I would accept a pull request for that, but first I would like to get an idea on how you would propose adding this. I am changing the milestone to backlog and the status to design.

Instead of a Func/delegate what about an interface for a custom formatter with a format method and a property or method that specifies the `Type` that the formatter handles? End users could implement the interface and pass the `Type` to NUnit which would instantiate it and call it where needed. This type could be passed in an assembly attribute or NUnit could also find all implementations of the interface when it explores the tests.

Another option could be an assembly attribute that takes a delegate and a .NET type that the delegate should format.

CharliePoole:
@rprouse Alternatively, we could leave the feature out of the .NET 2.0 build.

et1975:
I'd argue that interface+supported types would be great for introspection and dynamic lookup, but also an overkill for the use case.

[...]

PR #1198, opened 2016-01-15
#1196 pluggable value formatter

et1975:
+ Converted existing formatters

The order of registration maintains backwards compatibility with existing behavior: the last registered takes precedence.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/MsgUtils.cs
+++ b/src/NUnitFramework/framework/Constraints/MsgUtils.cs
@@ -29,9 +29,24 @@ using System.Globalization;
 namespace NUnit.Framework.Constraints
 {
     /// <summary>
+    /// Custom value formatter function
+    /// </summary>
+    /// <param name="val">The value</param>
+    /// <returns></returns>
+    public delegate string ValueFormatter(object val);
+
+    /// <summary>
+    /// Custom value formatter factory function
+    /// </summary>
+    /// <param name="next">The next formatter function</param>
+    /// <returns>ValueFormatter</returns>
+    /// <remarks>If the given formatter is unable to handle a certain format, it must call the next formatter in the chain</remarks>
+    public delegate ValueFormatter ValueFormatterFactory(ValueFormatter next);
+
+    /// <summary>
     /// Static methods used in creating messages
     /// </summary>
-    internal static class MsgUtils
+    public static class MsgUtils
     {
         /// <summary>
         /// Static string used when strings are clipped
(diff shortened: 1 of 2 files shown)


rprouse:
Nice, that was quick :+1: 

Looks good on a quick scan. I am just heading out skiing for the weekend. I will take a closer look on weekend if I can, or Monday at the latest. Thanks.

CharliePoole:
I ran all the tests, including Compact Framework, and it works fine.

I like the general idea of replacing our series of if statements with a chain of functions, with each factory responsible for knowing what it is capable of handling. I like the generality of a predicate as opposed to only testing for Type.

I also like the idea of allowing users to add their own formatters to the top of t he chain. However, I don't like that the head of the chain is a static. It's OK that our internal MsgUtils is static, but if users add something, then I'd prefer that new head to have some sort of scope. More comments to follow inline.

src/NUnitFramework/framework/Constraints/MsgUtils.cs
```diff
@@ -29,9 +29,24 @@
...
+    /// <param name="next">The next formatter function</param>
+    /// <returns>ValueFormatter</returns>
+    /// <remarks>If the given formatter is unable to handle a certain format, it must call the next formatter in the chain</remarks>
+    public delegate ValueFormatter ValueFormatterFactory(ValueFormatter next);
+
+    /// <summary>
     /// Static methods used in creating messages
     /// </summary>
-    internal static class MsgUtils
+    public static class MsgUtils
```
CharliePoole: It's not the end of the world to expose this entire class publicly, but I'm not in love with it.

src/NUnitFramework/tests/Constraints/MsgUtilTests.cs
```diff
@@ -32,6 +32,15 @@ namespace NUnit.Framework.Constraints
     public static class MsgUtilTests
     {
         #region FormatValue
+        class CustomFormattableType { }
+
+        [Test]
+        public static void FormatValue_CustomFormatterInvoked()
+        {
+            MsgUtils.AddFormatter(next => val => (val is CustomFormattableType) ? "custom_formatted" : next(val));
```
CharliePoole: When the test goes out of scope, the custom formatter remains in place.

I think I would prefer to see the head of the formatter list in the test execution context. That way, new formatters could be set up, passed to lower level test contexts, and removed when the setup context was removed.

For example:

In a OneTimeSetUp:

TestContext.AddFormatter (....

Use the custom format any place in the fixture

When the fixture exits, it's context is destroyed and the head reverts.

If it turns out to be more convenient, this could be done separately from the TestExecutionContext.

[Note: as described above, the static method in TestContexxt would access the current internal TestExecutionContext to update the current head of the list. This would mean that MsgUtils would need to access an existing TestContext. A side effect is that MsgUtils would be unusable outside of test execution. We would have to analyze to see if this poses a problem. If it does, then the stack of heads would need to be elsewhere.]

et1975:
@CharliePoole, I'd think that a formatter would be applicable for any test within a "domain" and as long as each domain gets its own test assembly it would work exactly as desired. 
What about combining the global setup and per-context setup?

CharliePoole:
Not sure what you mean by global setup unless it's the assembly-level context. All setup and teardown in nunit is organized around the context unless you defeat it by use of statically.

The problem I see with using a static in this case is that you have no control over where it is applied. The custom formatter in your test is active for that test and for any tests that happen to run after it. In addition, it's not thread-safe.

This isn't a fundamental problem with the idea but it is an important implementation detail.

et1975:
Yes, I meant global as in per-assembly. If I understand correctly, the proposed impl would be problematic to use safely because nunit has no established way to do this kind of initialization, unlike say [msunit's AssemblyInitialize](https://msdn.microsoft.com/en-us/library/microsoft.visualstudio.testtools.unittesting.assemblyinitializeattribute.aspx)?
Still, perhaps "I can do it under controlled conditions" is better than "I can't do it at all, unless I can use it per-context"?
As a side note, if per-context setup is the only solution you are going to consider, I don't think FsUnit could benefit from it, consequently it's a dead feature from F# testing perspective (since I don't see people introducing [OneTimeSetup] into every test module just to get the language-friendly formatting).

appel1:
Without having looked at the source code at all, couldn't this be implemented as a property similar to for example ParallelizableAttribute that can be used on an assembly, fixture and method level and that configures the test context using the IApplyToContext interface? Or perhaps that wouldn't be enough for FsUnit if test module is the same as a .NET assembly?

CharliePoole:
The equivalent of AssemblyInitialize in NUnit is a SetUpFixture in the
global namespace. It performs one-time initialization and cleanup for the
entire assembly.

To NUnit, the assembly is a test like any other. The topmost context is
created for that test and all others are spawned off of it. What you are
suggesting regarding a special property is entirely possible.

Unfortunately, wrt test module versus .NET assembly, you have the advantage
over me. I don't know what the term module means in F#!

I realize I made my suggestion in a very general way. We can get down to
specifics once @rprouse reviews this PR. That's because our policy is to
require two committers to agree before merging to master. In the end, we
should decide how we want it to work and then either you would implement it
or one of us could add something on top of what you have done. I think we
need you to continue with some level of involvement though, in order to
ensure that whatever we do works for F#. I have written tests for F#
myself, but they looked just like C# tests using the O-O paradigm. I know
that actual F# folks like to write things a bit differently.

et1975:
@appel1, thanks, wasn't aware of IApplyToContext. It might work, if I inherit from and hide under the same name the stock [Test] attribute. Guess I'll wait for "proper" implementation then :)
@CharliePoole F# favors immutability and tends focus on the smallest unit of functionality (i.e. functions), hence there is less need for setup/teardown because everything is either the input or the result. The tests often end up looking like [this ObservableSpecs.fs](
https://github.com/fsprojects/FSharp.Control.Reactive/blob/master/tests/ObservableSpecs.fs) where a single module (essentially a static class) contains tests for a large number of functions/scenarios under test.

CharliePoole:
@et1975 Yes, I am familiar with those concepts and in fact NUnit was modified early on to allow F# global test methods to be annotated. But I would never try to guess what any user wants, much less a user with a different background from mine. :-)

If you prefer an annotation, then it should be done as a separate attribute from Test, which is already overloaded in too many ways. I'm not really sure how you would expect such an attribute to work, however. Perhaps you could post some pseudo-code that indicates what you want to be able to do with such an attribute.

et1975:
I hope the talk of how I was intending to use the feature (outside of nUnit) did not derail the discussion about the nUnit changes themselves. 
Not sure if there's an outstanding objection to the pull request or we are just waiting for all the interested parties to comment...

CharliePoole:
We are waiting for two things...

1. You to respond to my comments about making the feature apply only to a limited contextual scope.
2. A second committer to review - so far it's just me. :-)

You can't do anything about item 2, but what do you think about item 1? Options would be for you to take it another step or for one of us to build on top of what you have already done. If you elect to do it, I'll be glad to feed you details of what is needed.

et1975:
> SetUpFixture in the global namespace
If the assembly-level initialization is already available and you simply don't want to make plugging of  formatters possible from there it sounds like it's the matter of principle and I'm not equipped to deal with that. 
I'll leave making it available per test context to someone who believes that the formatting is not a domain-scoped concern.

rprouse:
@et1975 I think the formatters will tend to be used in the global scope, but I think we should also make them available on more limited scopes. Our general approach to this sort of problem is to apply the functionality using attributes that can be specified at the whatever level is desired.

This makes it a deeper change. If you want to work with me, I can help you make it, otherwise I would be happy to pick it up. I think the feature you proposed would be a valuable addition to NUnit, so I would like to get it into the next release. We would just prefer it to be more flexible than this PR currently provides.

CharliePoole:
@et1975 Like @rprouse, I made a similar offer to work with you on this change, which I like and want to incorporate in NUnit. 

Closing the PR makes it seem that you don't want to work with us and are withdrawing the code contribution. However, I'm not sure if that's what you meant. 

Your last comment about NUnit not wanting to permit plugging in a formatter at a global scope is a misunderstanding. The problem I see with your change is not that it **permits** global use, but that it **requires** it. Users have to have the option to do it either way. The changes needed to make this possible are pretty simple, at least for someone who is familiar with NUnit internals. Because it is your code, I wanted you to have the opportunity to make further changes, if you prefer. If you do not want to do this, we can do it, but only if you are still willing to contribute your existing code.

Where would you like to go from here?

et1975:
To me global and contextual formatters are separate features. Apparently we all want the global but you are not willing to take it separately from the contextual and I respect your reasons, whatever they are. I'm just not interested in working on that, so it's all yours, gentlemen.

CharliePoole:
@rprouse The last comment is still a bit ambiguous, but I guess he is still offering the code, even though he cancelled the PR. I'll pull it down, make a few changes and submit a new PR.

@et1975 I hope we are simply having a communication problem here, which will be clarified when we use code to communicate. You can review the new PR as well, of course.

CharliePoole:
Further problem with this... the passed in delegate can't use any members of the value unless it tests for null first. We would need to either document that or force the null value formatter to be tested for first.

Issue #1187, opened 2016-01-11
ITestAssemblyRunner.StopRun as implemented by NUnitTestAssemblyRunner


oznetmaster:
NUnitTestAssemblyRunner.StopRun (true) seems to leave the system in a chaotic state.

It does correctly abort any threads that are started by ParallelWorkItemDispatcher and SimpleWorkItemDispatcher.

However, if the Workitem starts a thread of its own, it seems to never be aborted.

Also, it appears that the original NUnitTestAssemblyRunner.Run method never returns, and there is never any indication that NUnitTestAssemblyRunner.RunAsync has completed.

It seems like some housekeeping is getting lost here.

I realize I may be the first to ever exercise this code?

CharliePoole:
At one point, I had a test bed that exercised it and became aware of the problem. Since we had no code to exercise it, I pushed it off until we had a GUI, which will definitely need it. See issue #166, which is marked status:blocked.

WorkItem doesn't start any threads. Do you mean that the user code starts them?

Tests that start threads of their own have always been a problem for NUnit, so it's no surprise that they cause a problem here. I think this probably should be solved for the simpler cases first.

oznetmaster:
WorkItem starts a thread whenever there is a timeout on a Test.

Actually, looking deeper, much of what I said is incorrect.  The only externally visible problem is that TopLevelWorkItem.Completed is never called when StopRun is called.  This leaves everything hanging.

In NUnitTestAssemblyRunner:

```
			log.Info ("Running tests");
			if (LoadedTest == null)
				throw new InvalidOperationException ("The Run method was called but no test has been loaded");

			CreateTestExecutionContext (listener);

			TopLevelWorkItem = WorkItem.CreateWorkItem (LoadedTest, filter);
			TopLevelWorkItem.InitializeContext (Context);
			TopLevelWorkItem.Completed += OnRunCompleted;

			StartRun (listener);
```

oznetmaster:
One possible solution for tests that start their own thread is to go ahead and run the TearDown method for an aborted Test.  Currently, the TearDown method is not run if the test is aborted.

oznetmaster:
It is possibly related that the ITestListener.TestFinished event is only raised once if  StopRun is called - for the currently running test.  Not for parent tests, or for the fixture or assembly.

CharliePoole:
The theory is that the detection of a cancelled test should ripple upward
through all the containing tests. It's fragile though.

CharliePoole:
[...]

PR #1202, opened 2016-01-17
1187 stoprun fix

oznetmaster:
This PR implement/fixes the IAssemblyTestRunner.StopRun method to work in all cases.

It does this by always running a WorkItem on its own thread, where previously that was only done in cases of timeout, a separate thread or specific thread type required .
 
It then only aborts the actual thread the test is running on, and allows everything else to clean up itself normally.

This eliminates almost all special handling for both abort and stop operations.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2012 Charlie Poole
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
@@ -24,10 +24,10 @@
 using System;
 using System.Collections.Generic;
 using System.Threading;
+using System.Reflection;
 using NUnit.Framework.Compatibility;
 using NUnit.Framework.Internal.Commands;
 using NUnit.Framework.Interfaces;
-using System.Reflection;
 
 namespace NUnit.Framework.Internal.Execution
 {
@@ -38,7 +38,7 @@ namespace NUnit.Framework.Internal.Execution
     /// </summary>
     public class CompositeWorkItem : WorkItem
     {
-//        static Logger log = InternalTrace.GetLogger("CompositeWorkItem");
+        //        static Logger log = InternalTrace.GetLogger("CompositeWorkItem");
 
         private TestSuite _suite;
         private ITestFilter _childFilter;
@@ -55,7 +55,8 @@ namespace NUnit.Framework.Internal.Execution
         /// </summary>
         /// <param name="suite">The TestSuite to be executed</param>
         /// <param name="childFilter">A filter used to select child tests</param>
-        public CompositeWorkItem(TestSuite suite, ITestFilter childFilter) : base(suite)
+        public CompositeWorkItem(TestSuite suite, ITestFilter childFilter)
+            : base(suite)
         {
             _suite = suite;
             _childFilter = childFilter;
@@ -69,7 +70,7 @@ namespace NUnit.Framework.Internal.Execution
         protected override void PerformWork()
         {
             // Inititialize actions, setup and teardown
-            // We can't do this in the constructor because 
+            // We can't do this in the constructor because
             // the context is not available at that point.
             InitializeSetUpAndTearDownCommands();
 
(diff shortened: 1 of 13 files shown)


oznetmaster:
Strange errors about NaN and Infinity in these builds.  Nothing I ever touched :(

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -331,6 +330,31 @@ private static bool IsStaticClass(Type type)
...
+                {
+                    var ctx = child.Context;
+                    if (ctx != null)
+                        ctx.ExecutionStatus = force ? TestExecutionStatus.AbortRequested : TestExecutionStatus.StopRequested;
+
+                    if (child.State == WorkItemState.Running)
+                        child.Cancel(force);
+                }
+            }
+        }
```
CharliePoole: Am I missing something here? It looks like you do nothing for the current work item context but only set the child context execution status to Abort/RunRequested. Why do it for the current item context?

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
```diff
@@ -135,19 +137,28 @@ public void Start()
...
-            if (_workerThread != null && !_workerThread.Join(0))
-#else
-            if (_workerThread != null && _workerThread.IsAlive)
-#endif
-                ThreadUtility.Kill(_workerThread);
+            lock (cancelLock)
+                if (_workerThread != null && _currentWorkItem != null)
+                {
+                    {
+                        _currentWorkItem.Context.ExecutionStatus = force ? TestExecutionStatus.AbortRequested : TestExecutionStatus.StopRequested;
```
CharliePoole: OK, here is the answer to my earlier question. This is where the topmost work item has its context set.

I think this would be clearer if it all happened in the Cancel method. Also, it does not appear that you handle the top-level context in SimpleWorkItemDispatcher.

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
```diff
@@ -135,19 +137,28 @@ public void Start()
...
+
+                        _currentWorkItem.Cancel(force);
+                        if (force)
+                            _currentWorkItem = null;
+                    }
+                }
```
CharliePoole: Extraneous braces

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -230,87 +230,66 @@ public virtual void Execute()
...
             else
                 RunTest();
 #elif PORTABLE
             RunTest();
 #else
-            ApartmentState currentApartment = Thread.CurrentThread.GetApartmentState();
+            currentApartment = Thread.CurrentThread.GetApartmentState();
 
-            if (Test.RequiresThread || Test is TestMethod && timeout > 0 || currentApartment != TargetApartment && TargetApartment != ApartmentState.Unknown)
+            if (Test is TestMethod)
```
CharliePoole: Two questions:

1. Given that we have been on a long-term path toward removing these "extra" threads, is there any way you can make this work without running the test method on its own thread? I can explain what i had planned in more detail if necessary, but in sum it involves eliminating SimpleWorkItemDispatcher and creating worker threads dynamically, as needed. This change pushes us a step back on that path. That said, if it's necessary, then we will have to live with it.

2. This appears to break the functionality when Apartment(STA) is placed on a class while running under the SimpleWorkItemDispatcher. In that case, the test won't be a TestMethod.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -230,87 +230,66 @@ public virtual void Execute()
...
                 RunTestOnOwnThread(timeout, TargetApartment);
             else
                 RunTest();
 #endif
         }
 
-#if SILVERLIGHT || NETCF
+#if (SILVERLIGHT || NETCF) && !PORTABLE
```
CharliePoole: If it's Silverlight or NetCF then it's not portable by definition - unless somebody is misusing the defines, that is.

CharliePoole:
I still need to do some more review of RunTestOnOwnThread and RunTest - I got confused looking at the diffs with all the #if statements so I'll pull it down and look at it.

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
oznetmaster: Actually, this is done in WorkItem.Cancel, so it is actually redundant here.

That is why it is not in SimpleWorkItemDispatcher.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
oznetmaster: There will still need to be threads for tests that require their own threads, have a required apartment setting, or have timeouts.  I do not see any way around that.

Always running the test method on its own thread provides consistency so that terminating a thread is always done in the same place, and works the same, no matter what the reason (abort or timeout).  
It also allows everything to clean up after a test the same way, whether it was aborted, stopped, timeout, or just ended normally.

Not sure what you mean about STA being placed on a class under the SimpleWorkItemDispatcher?  I may have missed something, but I thought I had the handled :(

oznetmaster: I think I found the STA problem.  I think I have fixed it in the latest commit.

oznetmaster: I guess there needs to be a test for this?  Apparently there isn't one, since all tests were passing.

Since I do not understand this whole apartment model thing, perhaps someone else could recommend a suitable test?

CharliePoole: The overall plan for the new framework was documented very completely - or at least verbosely. :-) It seems to be lost now. We have this page: https://github.com/nunit/nunit/wiki/Framework-Design but the pages to which it points are no longer in the wiki. Too bad. If I get time I will try to recover them.

However, I can summarize...

In NUnit V2, the Test object executed itself. There was no overall control of execution.

In 3.0, we created the dispatcher to control test execution. The idea was as follows:

1. Identify special requirements for each test (timeout, STA, etc.)
2. Break each test into work items, each of which can be managed independently
3. Create queues and workers that can handle each special requirement
4. Allow dynamic expansion and contraction of the queues and workers
5. Eliminate the "old" style of running, as exemplified by SimpleWorkItemDispatcher

Like many things we do, this has taken a while with the result that some of it is forgotten, at least by some people. I thnk this is still the way we want to go.

What is done so far (before your fix):
1. We no longer needed a per-test thread for running in the STA, when using the parallel dispatcher.

2. We still needed a per-test thread for timeouts, because we can't yet restart the worker threads dynamically. That's a prerequisite if the worker thread is what gets cancelled by a timeout.

3. We still support RequiresThread, although we had talked about eliminating it.

4. We still support SimpleWorkItemDispatcher, which is used on certain platforms. We could eliminate it but it would require some work and would only have an internal effect.

CharliePoole: Running a fixture in the STA simply means...

```
[TestFixture, Apartment(ApartmentState.STA)]
public class MyTestFixture
{
    ...
}
```

The fixture itself and all it's test cases are run in the STA. The cases do not actually need a separate thread, since the fixture is already in the STA.

A valid test would be a class like the above with a test case examining the current apartmentstate. We probably have such a test but it is only exercised using the dispatcher that is in effect when running. If we continue to support the SimpleWorkItemDispatcher, we probably should have separate tests for it somewhere. I didn't write them because i thought we would be getting rid of it much faster than we have.

oznetmaster:
So, when I try to run the SL 5.0 tests with nunit3-console, I get the error:

```
C:\Projects\GitHub\nunit\bin\Debug\sl-5.0>..\nunit3-console nunit.framework.tests.dll
NUnit Console Runner 3.0.5860 (Debug)

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.10586.0
  CLR Version: 4.0.30319.42000

Test Files
    nunit.framework.tests.dll

The NUnit 3.0 driver does not support the portable version of NUnit. Use a platform specific runner.
```

Since I am running the Silverlight build, the error seems incorrect?

oznetmaster:
This now passes all CI tests.

Where do we go from here?

CharliePoole:
You were waiting for me to finish reviewing. Sorry to take so long but there have been a few family things going on.

CharliePoole:
I implemented the TestAssemblyRunnerTests for both Portable and Silverlight. They pass. :-)

I was going ot push these to your branch, but I ended up having to make changes to AssemblyHelper in addition to the tests, so I'll make this a separate branch after this merges.

I guess we should re-adress the whole issue of eliminating the thread-per-test-case implementation as a future change. This was all plotted out and documented in our original DevWiki, but apparently did not survive the transition to a GitHub wiki. The page that I pointed to is merely the summary page with pointers to five or six different design changes for the framework. Some of these are implemented but others, including this one, were still ongoing and I still think they are important. It's too bad the info got lost, since it would have been of help to @oznetmaster in making this change.

I'm OK with merging this and building on it. @rprouse What do you think?

rprouse:
I am also OK merging this and will do so. @CharliePoole, would you like to create a PR and/or Issue for the TestAssemblyRunnerTests?

rprouse:
@oznetmaster for future reference, could you put something like fixes #1187 into the description of the PR to automatically link and close the issue? You put the fix #1187 into one of the commits which will close the issue when merged, but the link ends up being to the commit, not the series of commits that made up the PR. Just a bit of housekeeping that makes navigation between issues, PR's and source easier.

CharliePoole:
@rprouse I'll merge and do a separate PR for the tests.

CharliePoole:
It looks like our Silverlight tests have not run correctly under Silverlight since this merge. It slipped past me, because they do work when run from the console using nunitlite-runner. However, when run directly as a silverlight app using nunit.framework.tests.xap, the run hangs and never terminates.

I have confirmed this by reseting to the commit prior to this merge and seeing the tests run and then to the merge commit itself and seeing them hang.

Since I came across this while working on adding further tests under Silverlight and Portable, I'll continue to look at it and try to include a fix in the same PR. 

@rprouse I was planning to do this as a standalone PR but if you would prefer an issue that's fine as well.

oznetmaster:
I am unclear how to run tests as a Silverlight app.  I am assuming this, like running the CF build tests, is not something that is done as part of either the CI process or by the -TestAll option?

CharliePoole:
That's correct. To run as a silverlight App, you make it the startup project in VS and then run with or without debugging. At one point, I tried creating a cmd file that would launch the xap file correctly but I abandoned the effort.

The -TestAll option runs the silverlight build under a .NET 4.5 console runner, which is handy to have but not the full deal.

FWIW, I think this is not a terribly important issue except that it may show up some threading issues under silverlight.

I'm currently debugging into the silverlight run and I'll let you know when I localize the problem.

CharliePoole:
It was the static constructor for ExceptionHelper,  where you use reflection to get the PreserveStackTrace delegate. It doesn't exist in Silverlight. Since running under the .NET 4.5 runner uses the 4.5 corlib, that run was able to succeed.

oznetmaster:
I never touched ExceptionHelper.  So it was always that way?

oznetmaster:
It appears that was done by @thomaslevesque back in December.

CharliePoole:
It's not super important who did it - it's not a "dumb" mistake, just something that happens when  there are so many alternative builds. I do know that the error appears at the point where Rob commited your changes. My guess is that Thomas planted the seed in the code and something in your changes caused the code in ExceptionHelper to be exercised under Silverlight, when it was not done previously. The rest of us compounded it by not running those particular tests.

I feel like we have searched for causes sufficiently. I'm just going ahead and fixing it. :-)

oznetmaster:
I guess what I was asking was if this Silverlight testing should be requisite for every PR prior to merge, in the same way we try to do for the CF testing of PR's.  Both are apparently manual procedures, so present similar obstacles.

CharliePoole:
It has been, but is sometimes forgotten, just like CF. I had hoped that creating the runner for executing under 4.5 would make the manual test unnecessary, but it clearly doesn't.

oznetmaster:
Also, the code that is there specifically handles the case of the method not existing, which it does not in both Silverlight and CF.  If reflection returns null, an empty method is stored as the delegate.

This works fine in CF, why not in Silverlight?

CharliePoole:
Why certain things don't work in Silverlight is a bit of a mystery to me. For example (not related to this problem) why does it appear to completely ignore BindingFlags.FlattenHierarchy?

Anyway... this is probably better continued when I put in a PR for the changes I made to fix it. I still have one other break that has crept into the code - not related to this issue.

oznetmaster:
@CharliePoole You commented a while back "In that case, the test won't be a TestMethod".

I sort of ignored that comment at the time, but now want to make sure I did not forget something.

When should a Test which is not a TestMethod be run on its own thread?  It seems like none of the other classes that are derived from Test ever need to be run on their own threads, but I may be missing something.

CharliePoole:
See my answer on the separate issue you created, #1244

PR #1206, opened 2016-01-20
add fix and test for NoArgProvided for TestCaseSource

### LINKED ISSUE #1103: Can't use TestCaseData from base class
Hi,
I've recently updated from NUnit.3.0.0-beta-2 to NUnit.3.0.0 release version and now i've got next problem.
I've got a child fixture class which uses BaseTestCaseSource defined in base class. 
```
public class BaseClass
{
        public static readonly object[] BaseTestCaseSource = 
        {
             new TestCaseData("testCase")
        };
}

[TestFixture]
public class ChildClass : BaseClass
{
    [TestCaseSource("BaseTestCaseSource")]
    public void ChildTest(string argument)
    {
       ...
    }
}
```
Everything worked perfectly, but now i've got an error "No arguments were provided" running this test. Defining BaseTestCaseSource in ChildClass solves the problem, as do alternative constructor calling
```
[TestCaseSource(typeof(BaseClass), "BaseTestCaseSource")]
public void ChildTest(string argument)
{
    ...
}
```
Is it a bug, supposed behaviour or just me doing something wrong?

jhamm:
fix #1103

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
@@ -216,7 +216,7 @@ namespace NUnit.Framework
                 return Reflect.Construct(sourceType) as IEnumerable;
 
             MemberInfo[] members = sourceType.GetMember(SourceName,
-                    BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance);
+                    BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance | BindingFlags.FlattenHierarchy);
 
             if (members.Length == 1)
             {
--- a/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
@@ -174,7 +174,7 @@ namespace NUnit.Framework
                 return Reflect.Construct(sourceType) as IEnumerable;
 
             MemberInfo[] members = sourceType.GetMember(SourceName,
-                    BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance);
+                    BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance | BindingFlags.FlattenHierarchy);
 
             if (members.Length == 1)
             {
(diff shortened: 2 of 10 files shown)


rprouse:
@jhamm the portable tests are failing. This is because portable uses extension methods for many of the reflection API's and implements its own BindingFlags. It looks like you have hit one of the binding flags that isn't supported or improperly supported. You will need to step into the extension method (in NUnit.Compatibility) and see what you can do.

When you fix it, you should also add a test to the compatibility tests for that error.

Let me know if you need help.

BTW, it passes on Travis because we don't build portable there.

jhamm:
Thanks for pointing out where to look in the code to fix it. I will take a look later tonight.

jhamm:
You are right, it is in the ReflectionExtensions.cs > TypeExtensions.GetAllMembers(Type) method. This method removes static members from the base types and the ValueSourceAttribute and TestCaseSourceAttribute require the 'sourceName' to be static.

I have some code that "works". I plan on spending some more time on it before committing it to make sure it has the correct behavior.

rprouse:
Do you need any help on this @jhamm? I wrote the extension methods, so I could add the fixes needed there if you get stuck.

jhamm:
I think I am ok for now. I should have some more time to work on this early this week.

jhamm:
I have corrected the failing tests for the portable target for AppVeyor.

It looks like the travis-ci build [mono:latest](https://travis-ci.org/nunit/nunit/jobs/104814173) timed out.
```
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
```

Is this normal? Travis CI status is reporting this [issue with travis-ci.org (public) container-based builds.](https://www.traviscistatus.com/incidents/ccjlfnfbhfdh) I am not sure if this has anything to do with this.

Do we need to change the DatapointProvider as well? I see it is using the same underlying methods as `GetMember`? If yes, should this be done in this commit or a new issue?

jhamm:
The Travis-CI build is now showing as passing.

CharliePoole:
I'm OK with merging this, but it takes two committers to approve it. @rprouse can you take another look?

Regarding DataPointProvider, it seems to me that a user would expect a static Datapoint to be found in a base class, so it probably should be changed. We would want a test and a fix. It's up to you whether to do it in this branch or to do a separate one. Looks like a quicky though.

jhamm:
I added the DatapointProvider with tests to this PR. I think this should cover it. :smile:

CharliePoole:
Looks good to me.

rprouse:
Looks good to me too. Merged.

rprouse:
@jhamm are you interested in continuing to work with the NUnit team? Email me at rob@prouse.org if you would be interested in joining us.

jhamm:
Thanks! Charlie Pool has already sent an invitation.

CharliePoole:
...which you accepted! :-) Unfortunately, I didn't realize you had till now.

As a contributor, we can give you credit for what you do by assigning the issue to you. I retroactively assigned #1103 to you. IIRC, GitHub won't let you assign yourself, but you can tell us if there is one you want to work on through a comment and one of us will do it.

PR #1222, opened 2016-01-28
Issue 1196

### LINKED ISSUE #1196: Custom value formatter for v3 via MsgUtils

CharliePoole:
This PR fixes #1196 and replaces PR #1198, which was withdrawn by its creator @et1975.

The initial commit includes the original code from that PR plus a fix to ensure that a check for a null value is done before any custom formatters are called. This means that it is not possible to add a special display format for `null` but in compensation avoids any issues with formatters that make use of the value without checking.

Further commits will make this available through the TestContext. I'm not yet sure whether we need to keep the static method for use by F#. From some of @et1975 's comments, that seems to be the case, but he never explained how the feature would be used from F#, only that he didn't want to use it from a OneTimeSetUp method.

F# folks: you can help us here. The OP has left the scene, so there's a good chance we will end up creating something that is not useful to you unless somebody else is willing to communicate. :-)

### DIFF
--- a/src/NUnitFramework/framework/Constraints/MsgUtils.cs
+++ b/src/NUnitFramework/framework/Constraints/MsgUtils.cs
@@ -29,9 +29,24 @@ using System.Globalization;
 namespace NUnit.Framework.Constraints
 {
     /// <summary>
+    /// Custom value formatter function
+    /// </summary>
+    /// <param name="val">The value</param>
+    /// <returns></returns>
+    public delegate string ValueFormatter(object val);
+
+    /// <summary>
+    /// Custom value formatter factory function
+    /// </summary>
+    /// <param name="next">The next formatter function</param>
+    /// <returns>ValueFormatter</returns>
+    /// <remarks>If the given formatter is unable to handle a certain format, it must call the next formatter in the chain</remarks>
+    public delegate ValueFormatter ValueFormatterFactory(ValueFormatter next);
+
+    /// <summary>
     /// Static methods used in creating messages
     /// </summary>
-    internal static class MsgUtils
+    public static class MsgUtils
     {
         /// <summary>
         /// Static string used when strings are clipped
@@ -52,48 +67,61 @@ namespace NUnit.Framework.Constraints
         private static readonly string Fmt_Default = "<{0}>";
 
         /// <summary>
+        /// Format function that will be invoked as last effort
+        /// for values of indeterminate type.
+        /// </summary>
+        static ValueFormatter ValueFormatter = val => string.Format(Fmt_Default, val);
+        static MsgUtils()
+        {
+            AddFormatter(next => val => val is ValueType ? string.Format(Fmt_ValueType, val) : next(val));
+
+            AddFormatter(next => val => val is DateTime ? FormatDateTime((DateTime)val) : next(val));
+
+            AddFormatter(next => val => val is decimal ? FormatDecimal((decimal)val) : next(val));
+
+            AddFormatter(next => val => val is float ? FormatFloat((float)val) : next(val));
+
+            AddFormatter(next => val => val is double ? FormatDouble((double)val) : next(val));
+
+            AddFormatter(next => val => val is char ? string.Format(Fmt_Char, val) : next(val));
+
+            AddFormatter(next => val => val is IEnumerable ? FormatCollection((IEnumerable)val, 0, 10) : next(val));
+
+            AddFormatter(next => val => val is string ? FormatString((string)val) : next(val));
+
+            AddFormatter(next => val => val.GetType().IsArray ? FormatArray((Array)val) : next(val));
+
+            AddFormatter(next => val => val == null ? Fmt_Null : next(val));
+
+#if NETCF
+            AddFormatter(next => val =>
+            {
+                var vi = val as System.Reflection.MethodInfo;
+                return (vi != null && vi.IsGenericMethodDefinition)
+                        ? string.Format(Fmt_Default, vi.Name + "<>") 
+                        : next(val);
+            });
+#endif
+        }
+
+        /// <summary>
+        /// Add a formatter to the chain of responsibility.
+        /// </summary>
+        /// <param name="formatterFactory"></param>
+        public static void AddFormatter(ValueFormatterFactory formatterFactory)
+        {
+            ValueFormatter = formatterFactory(ValueFormatter);
+        }
+
+
+        /// <summary>
         /// Formats text to represent a generalized value.
         /// </summary>
         /// <param name="val">The value</param>
         /// <returns>The formatted text</returns>
         public static string FormatValue(object val)
         {
-            if (val == null)
-                return Fmt_Null;
-
-            if (val.GetType().IsArray)
-                return FormatArray((Array)val);
-
-            if (val is string)
-                return FormatString((string)val);
-
-            if (val is IEnumerable)
-                return FormatCollection((IEnumerable)val, 0, 10);
-
-            if (val is char)
-                return string.Format(Fmt_Char, val);
-
-            if (val is double)
-                return FormatDouble((double)val);
-
-            if (val is float)
-                return FormatFloat((float)val);
-
-            if (val is decimal)
-                return FormatDecimal((decimal)val);
-
-            if (val is DateTime)
-                return FormatDateTime((DateTime)val);
-
-            if (val is ValueType)
-                return string.Format(Fmt_ValueType, val);
-
-#if NETCF
-            var vi = val as System.Reflection.MethodInfo;
-            if (vi != null && vi.IsGenericMethodDefinition)
-                return string.Format(Fmt_Default, vi.Name + "<>");
-#endif
-            return string.Format(Fmt_Default, val);
+            return ValueFormatter(val);
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/MsgUtils.cs
+++ b/src/NUnitFramework/framework/Constraints/MsgUtils.cs
@@ -44,9 +44,43 @@ namespace NUnit.Framework.Constraints
     public delegate ValueFormatter ValueFormatterFactory(ValueFormatter next);
 
     /// <summary>
+    /// Small class to allow adding ValueFormatter instances to MsgUtils
+    /// without exposing MsgUtils publicly.
+    /// </summary>
+    public static class ValueFormatters
+    {
+        /// <summary>
+        /// This method adds the a new ValueFormatterFactory to the
+        /// chain of responsibility.
+        /// </summary>
+        /// <param name="formatterFactory">The factory delegate</param>
+        public static void Add(ValueFormatterFactory formatterFactory)
+        {
+            if (Internal.TestExecutionContext.CurrentContext != null)
+                throw new InvalidOperationException("Cannot add global foratters once execution has started");
+
+            MsgUtils.AddFormatter(formatterFactory);
+        }
+
+        /// <summary>
+        /// This method provides a simplified way to add a ValueFormatter
+        /// delegate to the chain of responsibility, creating the factory
+        /// delegate internally. It is useful when the Type of the object
+        /// is the only criterion for selection of the formatter, since
+        /// it can be used without getting involved with a compould function.
+        /// </summary>
+        /// <typeparam name="TSUPPORTED">The type supported by this formatter</typeparam>
+        /// <param name="formatter">The ValueFormatter delegate</param>
+        public static void Add<TSUPPORTED>(ValueFormatter formatter)
+        {
+            Add(next => val => (val is TSUPPORTED) ? formatter(val) : next(val));
+        }
+    }
+
+    /// <summary>
     /// Static methods used in creating messages
     /// </summary>
-    public static class MsgUtils
+    internal static class MsgUtils
     {
         /// <summary>
         /// Static string used when strings are clipped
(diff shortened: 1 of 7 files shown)


sergey-tihon:
As I understand the usage from F# will look like [this](https://github.com/et1975/FsUnit/blob/c3035c89dd114d6e330127740c65b9cb953c8fbb/src/FsUnit.NUnit/FsUnit.fs#L12-L13)
```fsharp
MsgUtils.AddFormatter (ValueFormatterFactory(fun _ -> ValueFormatter(fun v -> sprintf "%A" v)))
```
what allow us to use [sprintf](https://msdn.microsoft.com/en-us/library/ee370455.aspx) from `FSharp.Code.dll` to pretty-print values.

CharliePoole:
That explains some of my confusion with the earlier discussion. You are using it in the initialization for FsUnit itself. In that context, it clearly needs to be a global and the issue of cleaning up after the context finishes does not arise.

Our prior discussion (or at least my half of it) was about how **users** of FsUnit might use a special formatter - specifically where they would put the call to AddFormatter in their test code. If they put it in a test, then the formatter would be available for that test and any tests that executed after it, which is basically a random effect.

Another problem that arises in the user context is that of parallelism. As implemented so far, adding a formatter is not thread-safe, as it will have to be if exposed to users.

sergey-tihon:
I also think that user should not call `AddFormatter` from test code. They should be able setup it once before execution of any fixture from the assembly.

CharliePoole:
@sergey-tihon Of course, why didn't I think of that? :-)

sergey-tihon:
What is the best practice to configure `NUnit` from the code?

Maybe I can implement some interface, you find this class, call the method and give me all internal configs, I modify them and return you back what I need?

Or maybe I can use assembly level attribute to specify class that should be called at the very beginning?

CharliePoole:
@sergey-tihon Normally, we configure NUnit by passing in settings with the test package. But that configuration is done by a runner and (if I understand correctly) you don't call into NUnit to run the tests, but rely on our runners or third parties to run them.

As I understand it, the only relation between the nunit framework and fsunit is that fsunit references the framework. If you were strong-named, we could simply make our internals visible to you and leave the AddFormatter method as an internal.

This is getting complex enough that I'm beginning to think that I should just make it work for you and worry about the consequences of misuse later, if it happens.

CharliePoole:
@sergey-tihon Here's an idea, together with a question.

We could simply throw an exception if anyone tries to add a formatter after the tests have started to run. Can we be sure that your static initialization will complete before any execution begins?

sergey-tihon:
@CharliePoole I guess we can be sure, but definitely would be better to test it.

So let me know when version on NuGet package will be available and I will try:
- F# tests with custom formatter in FsUnit
- F# tests with custom formatter registration in static code before tests
- C# tests with custom formatter registration

CharliePoole:
@sergey-tihon Appveyor build 3.1.0-CI-2076 should work for you, although I have not yet added the feature of restoring state after each test level finishes. That will be in the final merge.

CharliePoole:
@sergey-tihon Note that you have to call `ValueFormatters.Add` rather than `MsgUtils.AddFormatter`. I may have posted that already, but I can't see it anywhere.

sergey-tihon:
Thank you @CharliePoole. I will try to test it on the weekends and to share results with you

CharliePoole:
This is ready once the CI builds pass.

The implementation has changed yet again. Sorry @sergey-tihon :-)

Maintaining both a global entry and a context-based entry turned out to be more complicated than I imagined. Here is what I finally did:

* The global AddFormat methods are in our GlobalSettings class. That's a class I had thought we could phase out, doing everything within a TestExecutionContext. However, languages like F#, which allow global initialization to take place outside any test require some class to call into and GlobalSettings seems as good a place as any. Hopefully, the name will warn users away from calling into it to change settings while the tests are running. We can warn in the docs as well.

* Calls from within any test execution should use the AddFormat methods on TestContext. They access an implementation within the current TestExecutionContext, which restores the original head of chain when a context terminates. That's the normal expectation with other NUnit settings, so it makes sense here.

* The context chain and the global chain are tied together using a lambda in TestExecutionContext, so that global updates are actually accessible from within a test in a pinch.

* MsgUtils.FormatValue tests whether a TestExecutionContext exists and uses it if it does. Otherwise, it uses the global chain.

This deserves a good code review before merging.

rprouse:
I will review this in the morning. @sergey-tihon, we would also appreciate a look from you if you have time to make sure this meets your needs.

CharliePoole:
@sergey-tihon Please use 3.1.0-CI-2080 from our AppVeyor feed.

rprouse:
Looks good. This is definitely going to require some documentation and maybe even a blog post with examples :smile:

I am going to leave it open for a bit to give @sergey-tihon a chance to comment.

sergey-tihon:
Seems that is does not  work as expected either in VS and console runner.
Maybe it should not work in VS because test adapter does not support it yet, but definitely should in console runner.

image
image
image

You can find source code [here](https://github.com/sergey-tihon/nunit3.1-custom-formatter-tests).
Call `build.cmd` to reproduce.

sergey-tihon:
@CharliePoole could you please share appveyor project nuget feed url? something like 
>https://ci.appveyor.com/nuget/nunit-xxxxxxxxxxxx

CharliePoole:
It's https://ci.appveyor.com/nuget/nunit

You have to get version 3.1.0-CI-2080 only.

CharliePoole:
The code looks right to me, assuming that you can verify that the call to add the formatter actually runs and terminates before the test executes. That's the part that is unfamiliar to me. What exactly triggers the execution of that statement?

In fact, since this change is only to the framework, the VS adapter should work correctly. However, starting with the console gives me more assurance.

Can you verify that you are using the nunit package version 3.1.0-CI-2080? Based on the fact that the GlobalSettings call compiles, it does look correct to me.

You might try putting the call to GlobalSettings.AddFormatter or TestContext.AddFormatter in the test itself as a way of narrowing down the problem. That will match the only way I am able to test them in C#.

Meanwhile, I'll download your sample solution and try it myself.

BTW - the F# makes a lot more sense than the decompile. :-)

sergey-tihon:
I have changed [project dependencies](https://github.com/sergey-tihon/nunit3.1-custom-formatter-tests/blob/master/paket.dependencies), so now we can be sure that I use `3.1.0-CI-2080`.

I have tested `GlobalSettings.AddFormatter` and `TestContext.AddFormatter` used from constructor  of test fixture - it works from F# and [C#](https://github.com/sergey-tihon/nunit3.1-custom-formatter-tests/blob/master/src/CSharp.Tests/MyTests3.cs) ;=)

image

CharliePoole:
But it still doesn't appear to work when initialized in the module?

sergey-tihon:
No and i will not work, because of [this](http://stackoverflow.com/questions/18619093/how-to-execute-module-do-block) =(

So, we have no way to execute something one time before tests and probably do not need `GlobalSettings.AddFormatter` =(

The only way that I found is to add
```
[<SetUpFixture>]
type ``Set up custom formatter``() =
    do
      TestContext.AddFormatter(
        ValueFormatterFactory(fun _ -> ValueFormatter(sprintf "%A")))
```
in each assembly that has tests, but it cannot be implemented once in `FsUnit.dll`

sergey-tihon:
The solution with `SetUpFixture` may work with [FsUnit lightweight usage scenario](http://fsprojects.github.io/FsUnit/Paket.html) when users add FsUnit operators to their project, but will not with compiled `FsUnit.dll`

CharliePoole:
Is it possible to address this in another way? Can we identify the Types that you want to have formatted differently? If NUnit is just treating them as objects at this point, perhaps we can simply add some formatters to MsgUtils ourselves. If so, let's start a new issue.

@rprouse What do you think? I'm inclined to remove the GlobalSettings methods and clean up the code leaving only the TestContext methods available. I'll wait a few beats, however, incase @sergey-tihon or @et1975 come up with another solution.

CharliePoole:
@sergey-tihon I havn't used FsUnit, so maybe I'm missing something. Don't you use NUnit itself to run the tests? If that's so, it's hard to see how you could avoid such SetUpFixture code from being executed.

sergey-tihon:
@CharliePoole We use NUnit to execute test. `FsUnit` is very simple unit testing DSL for F# users.

Could NUnit execute `SetUpFixture` implemented in `FsUnit.dll` when I execure tests from `MyProject.Test.dll` that reference `FsUnit.dll` ?

rprouse:
> Could NUnit execute SetUpFixture implemented in FsUnit.dll when I execute tests from MyProject.Test.dll that reference FsUnit.dll ?

Anything is possible, but we don't normally walk out to referenced assemblies. We sometimes add special code for other consumers of NUnit, but we prefer not to if we can help it.

CharliePoole:
@sergey-tihon If you use NUnit (console, gui, adapter) to run the tests, then the framework actually does the discovery and execution in 3.0. So any SetUpFixtures created by the user should definitely get run.

However, we don't look at any other assemblies referenced by the test assembly, except for Types that serve as a base class for the test fixtures. I suspect that an empty Type, inheriting from a Type that you provide, and marked with SetUpFixtureAttribute would do the job, although it would seem a bit of an intrusion for the users.

Of course, we **could** inspect all the references, detect fsunit.dll and do something special, but that seems like a bad slope to start down. If we need to do something, I would prefer to think carefully about some sort of more general approach.

sergey-tihon:
OK, looks good for me. I do not see options to setup custom formatter besides `SetUpFixtures`.

So, you may remove `GlobalSettings.AddFormatter`, `TestContext.AddFormatter` works well.

@CharliePoole & @rprouse - thank you for help with this.

et1975:
@sergey-tihon I see the module executing the do() on first access to any let binding that is not a function. Since `should` is a function do() is not executed. I guess we could add a small cludge to `should`:
```fsharp
    let should (f : 'a -> #Constraint) x (y : obj) =
        Null |> ignore
        let c = f x
        let y =
            match y with
            | :? (unit -> unit) -> box (TestDelegate(y :?> unit -> unit))
            | _ -> y
        Assert.That(y, c)
```

sergey-tihon:
@et1975 As I understand it will be executed during test execution (not before). So we should call `TestContext.AddFormatte` instead of `GlobalSettings.AddFormatter` in `do` statement.

sergey-tihon:
@et1975 This hack works in general https://github.com/sergey-tihon/nunit3.1-custom-formatter-tests/commit/fb05fa5b99ec64cf0c9c0cc4d4a8920042af0ee6 and we still do not need `GlobalSettings.AddFormatter` for this, but I an not sure should we use it in `FsUnit` or not...

sergey-tihon:
@et1975 Such `hack` may be optimized by compiler during compilation in `Release` mode.

et1975:
> optimized by compiler during compilation in Release mode.

Shockingly enough it *is* optimized away! And not just the `do` body, but a lot more, like lazy binding enforcement, which might be a bug similar to Microsoft/visualfsharp#759.

Something like this however works across compilers and release modes:
```fsharp
module TopLevelOperators =
    let formatter = ref(TestContext.AddFormatter (ValueFormatterFactory(fun _ -> ValueFormatter(fun v -> sprintf "%A" v))))
```
forced by:
```fsharp
    let should (f : 'a -> #Constraint) x (y : obj) =
        ignore !formatter
        let c = f x
        let y =
            match y with
            | :? (unit -> unit) -> box (TestDelegate(y :?> unit -> unit))
            | _ -> y
        Assert.That(y, c)
```

Obviously I'm not going to press the issue, I just think that when crossing barriers like this there's no clean solution and every tool in the belt should be considered.

CharliePoole:
Removing GlobalSettings.AddFormatter, any user tests can (and should) use TextContext.AddFormatter if they want a special format.

You can use it as well but it will only be useful in the light configuration, where FsUnit.fs becomes part of the user's test assembly.

This brings me back to an earlier, unanswered question: are there specific types used in F# for which NUnit ought to have some native formatter in the first place? If we could do this, it would remove the need for any specialization on your part.

et1975:
FSharp.Core is not a small dependency to have, but it would be the only way to support F# types w/o some reflection magic. With reflection, one could look to see if FSharp.Core is loaded into current domain, find `sprintf` function in it and add a formatter based on it. Somewhat less deterministic, imho, than what's on the table.

CharliePoole:
@et1975 True, but the up side would be that it would support F# users, whether they made use of FsUnit or not.

In the original design for today's NUnit 3.0 (ca 2007) there were language adapters on top of the framework itself adding language-specific features. I'm wondering if it isn't time to think about implementing that.

Of course, that's a big deal and won't happen right away.

CharliePoole:
@rprouse et al... ready for re-review and merge

Too bad it didn't work out for fsunit, but we gained a minor feature for users anyway.

oznetmaster:
@CharliePoole you decided not to add DateTimeOffset to this PR after all?

oznetmaster:
Aside from my previous comment, this looks good to me (visibly) for merge.
@CharliePoole did you test this in the CF build yet, or should I?

CharliePoole:
@oznetmaster I didn't add DateTimeOffset.

I did test under CF before my final commit. It wouldn't hurt if you could test it in its present state - I'm away from my desk right now.

oznetmaster:
The final commit passes all the tests under CF.

CharliePoole:
Thanks. Merging.

PR #1230, opened 2016-01-31
Add ability for testcasedatasource to have parameters passed to methods

dicko2:
Comments on Issue #7 reflect changes in here. However @CharliePoole mentioned creating a new issue for this change as its actually unrelated to the original intention of issue number 7.

One thing i did consider changing is making the TestCaseDataSource "blow up", or not, if someone tries to use it with a property or field instead of a method. Personally I don't like things that silently fail, but would someone ever try to pass parameters to a field or property?

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
@@ -39,6 +39,7 @@ namespace NUnit.Framework
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = false)]
     public class TestCaseSourceAttribute : NUnitAttribute, ITestBuilder, IImplyFixture
     {
+
         private NUnitTestCaseBuilder _builder = new NUnitTestCaseBuilder();
 
         #region Constructors
@@ -52,6 +53,19 @@ namespace NUnit.Framework
             this.SourceName = sourceName;
         }
 
+        /// <summary>
+        /// Construct with a Type and name
+        /// </summary>
+        /// <param name="sourceType">The Type that will provide data</param>
+        /// <param name="sourceName">The name of a static method, property or field that will provide data.</param>
+        /// <param name="methodParams">A set of parameters passed to the method, works only if the Source Name is a method. 
+        ///                     If the source name is a field or property has no effect.</param>
+        public TestCaseSourceAttribute(Type sourceType, string sourceName, object[] methodParams)
+        {
+            this.MethodParams = methodParams;
+            this.SourceType = sourceType;
+            this.SourceName = sourceName;
+        }
         /// <summary>
         /// Construct with a Type and name
         /// </summary>
@@ -75,7 +89,11 @@ namespace NUnit.Framework
         #endregion
 
         #region Properties
-
+        /// <summary>
+        /// A set of parameters passed to the method, works only if the Source Name is a method. 
+        /// If the source name is a field or property has no effect.
+        /// </summary>
+        public object[] MethodParams { get; private set; }
         /// <summary>
         /// The name of a the method, property or fiend to be used as a source
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -284,6 +294,8 @@ public static IEnumerable HereIsTheData
                     //    .Throws(typeof(System.DivideByZeroException));
                     yield return new object[] { 100, 20, 5 };
                     yield return new object[] { 100, 4, 25 };
+
```
rprouse: You added a couple of unneeded lines. Sorry, pet peeve :smile:

rprouse:
Initial pass, looks good.

I guess you could have two similarly named members, one a method and one a property and you might misspell it to get the other? Unlikely. What do you think @CharliePoole, should we throw if invalid? If so, what should he throw?

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
dicko2: oops, sorry I'm usually pretty god with that :)

CharliePoole:
I like the feature and it's a good implementation. Nice work!

I would prefer more error checking...

1. Give an error if a non-null params value is provided and the named value is not a method.
2. Give an error if the number of args provided does not match the parameters of the found method.

I think we'll end up doing the error checks later if we don't do them now. I'd be OK with merging this and adding the checks in a separate PR. The checks will require some further tests as well.

It would also be nice if we had one test that is more realistic than what we have, as some folks will look at our tests for an example. The existing test just provides the arguments themselves, but of course that's not how the feature will really be used. This is just "nice to have" as far as I'm concerned.

dicko2:
Ok cool, I stared working on some of these, and I've got the checks and error messages done. but my question is how to write a unit test for failure of an attribute?

I've got this example of a test that generates the error about the wrong number of arguments.

```csharp
[Test, Category("Top"), TestCaseSource(typeof(DivideDataProvider), "HereIsTheDataWithParameters", new object[] { 100, 4 })]
        public void SourceInAnotherClassPassingSomeDataToConstructorWrongNumberParam(int n, int d, int q)
        {
// code here is never reached
        }
```
Normally i would be testing inside the test method itself, and be able to put a try-catch in, check the error thrown is correct and assert form there. This however is throwing an error from the attribute so I am a bit stumped at how to make the test pass, I am guessing that is why a few of them are erroring now yea? Or am I looking at this the wrong way maybe?

CharliePoole:
Tests of failures in NUnit generally have to use a second level of test code, which is loaded and executed by the running test. Here's an example: https://github.com/nunit/nunit/blob/master/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs#L217

In your case, you would add the "bad" test to TestCaseSourceAttributeFixture and your actual executed test would verify that the bad test is not runnable and has a proper message.

rprouse:
Travis build failures looked like a network error on their end. I have restarted both builds.

dicko2:
Spot on @CharliePoole thanks for that! I've got those checks/errors added now.

dicko2:
Oops, n00bed it! sorry checked in again, build should work this time.

src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
```diff
@@ -213,7 +231,7 @@ private IEnumerable GetTestCaseSource(IMethodInfo method)
 
             // Handle Type implementing IEnumerable separately
             if (SourceName == null)
-                return Reflect.Construct(sourceType) as IEnumerable;
+                return Reflect.Construct(sourceType, null) as IEnumerable;
```
CharliePoole: So, your intention is that the third form should not support arguments? That's OK for now at least (till folks complain) but I want to get it straight for the documentation.

CharliePoole:
Looks good to merge. Nice work!

CharliePoole:
@oznetmaster Do you see anything here that will break CF build? I didn't pull it down to test.

rprouse:
Looks good to me too. I've held off three days to give @oznetmaster a chance to review, but I don't want to leave this too long and end up with merge conflicts.

Thanks for the contribution.

rprouse:
@dicko2, you mention creating an issue for this at the top of the PR, but I can't find any. I am going to assume that there isn't one and tag this PR accordingly. If there is, can you point me at it so that we can keep the release notes straight? If there isn't an issue, don't bother creating one, we will use this PR as the issue.

dicko2:
Yep no issue create @rprouse, that's correct.

Issue #1232, opened 2016-01-31
Inconsistent compatibility methods break test in portable


CharliePoole:
I am temporarily #ifing out out the test ValueSourceTests.ValueSourceCanBeInheritedstaticProperty in the portable build.

It fails because of a call to Type.GetMember, which in the portable build is supplied by one of our reflection extensions. The extension method ignores the BindingFlags argument, returning all members instead. In this specific case, the FlattenHierarchy flag plays no part so that static public and protected members of the base class are not returned.

I'm not sure how this slipped through our review or why the problem is only arising in ValueSource and not TestCaseSource or FixtureSource. That's why I'm making it a separate issue rather than just fixing this one test.

In fact, I think we have to review the notion of "compatibility" methods that do not really work like the methods of the same name they replace. In this case, the extension method's xml comment includes the phrase "BindingFlags are IGNORED" but that doesn't show up when the line of code being typed is active for multiple builds. As a result, it took me quite a while to understand the source of the problem I was trying to fix.

`

CharliePoole:
I think we should still deal with this but it's no longer high priority since the original problem I had was solved in another way. Moving to Backlog.

CharliePoole:
Closing since we will be removing the portable build.

PR #1233, opened 2016-01-31
Add TestAssemblyRunner tests to both portable and silverlight builds

CharliePoole:
This PR started as a supplement to #1202, adding tests of the test assembly runner that were missing in the silverlight and portable builds. That's included here.

In addition, I found that Silverlight tests were hanging since #1202. I believe it is due to an earlier change creating a hidden bug that #1202 activated. The hang did not occur when running silverlight tests under .NET 4.5, only when run directly by our silverlight runner. Consequently, the CI build did not catch it.

I also turned up a bug in one of the ValueSourceTests, which I commented out for now. I filed #1232 to cover that one, which is a bit of a can of worms. In fact, it's possible that the underlying cause of #1232 will turn out to have something to do with the problem in #1202 leading to a modification of the fix I just made.

### DIFF
--- a/nunit.sln
+++ b/nunit.sln
@@ -161,6 +161,10 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "nunitlite-runner-4.0", "src
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "nunitlite-runner-sl-5.0", "src\NUnitFramework\nunitlite-runner\nunitlite-runner-sl-5.0.csproj", "{901DA110-F30A-4515-880D-E9D59B6ECB4E}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "slow-nunit-tests-portable", "src\NUnitFramework\slow-tests\slow-nunit-tests-portable.csproj", "{74CB429E-69D7-4525-BBDF-BBC47DBC7070}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "slow-nunit-tests-sl-5.0", "src\NUnitFramework\slow-tests\slow-nunit-tests-sl-5.0.csproj", "{A2B973C4-EA96-4B00-B2BE-04BFD9D5AA39}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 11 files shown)


src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
```diff
@@ -216,7 +216,7 @@ private IEnumerable GetTestCaseSource(IMethodInfo method)
                 return Reflect.Construct(sourceType) as IEnumerable;
 
             MemberInfo[] members = sourceType.GetMember(SourceName,
-                    BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance | BindingFlags.FlattenHierarchy);
+                    BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance);
```
rprouse: It looks like you are removing the two fixes recently submitted for TestCaseSource and TestFixutreSource not finding statics in base classes. Is this intentional? I don't see it mentioned in the PR.

src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
```diff
@@ -199,31 +199,35 @@ static IList<MemberInfo> GetAllMembers(this Type type)
...
+                // Skip statics on all base classes
                 var baseMembers = type.GetAllMembers();
-                members.AddRange(baseMembers.Where(NotPrivate));
+                members.AddRange(DiscardStatic(baseMembers));
             }
             return members;
         }
 
-        static bool NotPrivate(MemberInfo info)
+        static IEnumerable<MemberInfo> DiscardStatic(IEnumerable<MemberInfo> members)
```
rprouse: This also looks like a reversion. Was it not correct?

CharliePoole:
I think something got messed up due to my creating the branch from @oznetmaster 's original branch but not creating the PR until after some other changes were merged. I'll clean up.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -380,7 +380,7 @@ public virtual void Cancel(bool force)
 
             if (!tThread.Join(0))
             {
-                log.Debug("Killing thread {0} for cancel", tThread.ManagedThreadId);
+                log.Debug("Killing thread {0} for cancel", thread.ManagedThreadId);
```
oznetmaster: This is also a reversion of the code for #1202.  This will break WorkItem.

CharliePoole: Yes... a number of things are out of whack with this commit... see my reply to @rprouse comment.
I'm going through it now to straighten it out.

CharliePoole:
Not sure what happened with the earlier commit. I not only wiped out some fixes, but most of my own changes were not in it. I started over and did a forced update to the branch, so this PR should now reflect changes based on current master.

CharliePoole:
@rprouse This is ready for re-review now.

@oznetmaster It looks to me as if the problems I was having with Delegate creation in Silverlight must have been connected with the bad state of my branch. I don't see them any longer.

rprouse:
Much better, the previous code had me really confused. This makes sense. I was hoping it wasn't me :smile:

oznetmaster:
Actually, a little of this still does not make any sense :(

The change to ExceptionHelper.  The static constructor and the static PreserveStackTrace are only used in the NET_4_0,  build, but they are included in all builds except the NET_4_5, PORTABLE and SILVERLIGHT builds.

CharliePoole:
I went through a few places and tried to simplify the algebra of conditional statements - I guess I mised this one.  I had to prevent the static constructor from running in order for the tests to complete when run under silverlight. I checked that this is still a problem, and it is in spite of my earlier comment.

We should probably come back to this and clean up the code at some point.

oznetmaster:
Also, shouldn't WorkItemTests be in the Internal folder?

CharliePoole:
Really, yes. But of course it's merged now so I'll try to remember it when I'm in this code again.

oznetmaster:
In WorkItemTests, the methods StartExecution and ThreadProc are defined, but StartExecution is never called.  Is something missing?

CharliePoole:
It's a work in progress. I wanted to create true unit tests for WorkItem, independent of any of the higher-level classes, but I set it aside for now. Too much else.

PR #1237, opened 2016-02-02
Modifications to package version

### LINKED ISSUE #1236: Make Appveyor NuGet feed more useable
Because the nuget feed uses the AppVeyor build number, you have to look on AppVeyor to figure out the nature of each artifact. You can't just take the latest one.

We don't want to have multiple feeds, so we need to encode more information in the extended version number, after the existing 3.1.0-CI-nnnn

* For pull requests, append -PR- plus the PR # Normally, folks don't want these.

* For merges into master (builds of master that are not PRs), add an identifying suffix. This could be -MASTER or something shorter that won't be confused with anything else. Folks who want the "latest" build should take the most recent of these.

* For builds of other branches, we could append the branch name just as for master. The only problem is that we may run into problems if the version tag exceeds some maximum length. I've seen this happen in some experiments. In that case, we should truncate.

CharliePoole:
This fixes #1236 labelling artifacts more completely. See issue for details.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -83,9 +83,20 @@ Task("InitializeBuild")
 	if (BuildSystem.IsRunningOnAppVeyor)
 	{
 		var tag = AppVeyor.Environment.Repository.Tag;
-		var buildNumber = AppVeyor.Environment.Build.Number;
-		
-		packageVersion = tag.IsTag ? tag.Name : version + "-CI-" + buildNumber + dbgSuffix;
+
+		if (tag.IsTag)
+		{
+			packageVersion = tag.Name;
+		}
+		else
+		{
+			var buildNumber = AppVeyor.Environment.Build.Number;
+			packageVersion = version + "-CI-" + buildNumber + dbgSuffix;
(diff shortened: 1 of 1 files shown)


rprouse:
Looks like we may have introduced another timing related failure with the test stop run changes? I am rerunning the build because it is unrelated to this. 

Error in the build;

```
1) Failed : NUnit.Framework.Attributes.TimeoutTests.TestWithTimeoutRunsOnSeparateThread
Test exceeded Timeout value of 50ms
```

oznetmaster:
Not sure how.  The code for a test with a timeout has effectively not changed.  I think this may be something else entirely.

rprouse:
@oznetmaster, it could easily be something else, that was just the first thing that I thought of that has changed recently. It could also just be a weird glitch that we won't see again, but at least it is documented now.

CharliePoole:
Looks good. We got one build marked 3.1.0-CI-2090-issue-1236 and another 3.1.0-CI-2093-PR-1237. We should see 3.1.0-CI-xxxx-master when I merge.

PR #1240, opened 2016-02-03
EmptyConstraint - allow null string

### LINKED ISSUE #1171: A bug when using Assert.That() with Is.Not.Empty
I've noticed some unexpected behaviour when I use the following code:

```
// EXPECTED - throws NUnit.Framework.AssertionException
Assert.That(null, Is.Not.Null);

// UNEXPECTED - throws System.ArgumentException
Assert.That(null, Is.Not.Empty);

// EXPECTED - throws NUnit.Framework.AssertionException
Assert.That("", Is.Not.Empty)

// UNEXPECTED - throws System.ArgumentException
Assert.That(null, Is.Not.Null.Or.Empty)

// UNEXPECTED - doesn't throw exception at all
Assert.That("", Is.Not.Null.Or.Empty)
```

Note that in all cases I would expect `NUnit.Framework.AssertionException` to be thrown.

I guess this is a bug and I'm wondering if the code marked with `// UNEXPECTED` has been unit tested?

jhamm:
fix #1171 

The cases `Assert.That(null, Is.Not.Empty)` and `Is.Not.Null.Or.Empty` will continue to fail as they are written. The 'null' value cannot be checked to see if the EmptyConstraint types can be applied because this is lost.

If the test were rewritten as:
````
string actual = null;
Assert.That(actual, Is.Empty);

//or

Assert.That((string)null, Is.Empty);
````

I think that most tests will have the type information like the rewritten test cases as opposed to having just `null`.

These tests throw the AssertionException as expected. If the type cannot be applied, (not string, collection, or DirectoryInfo), the ArgumentException will be thrown.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/EmptyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/EmptyConstraint.cs
@@ -50,8 +50,10 @@ namespace NUnit.Framework.Constraints
         {
             var actualType = GetBaseType(actual);
 
-            if (actual is string || actualType == typeof(string))
+            if (actualType == typeof(string))
                 realConstraint = new EmptyStringConstraint();
+            else if (actual == null)
+                throw new System.ArgumentException("The actual value must be a non-null string, IEnumerable or DirectoryInfo", "actual");
 #if !SILVERLIGHT && !PORTABLE
             else if (actual is System.IO.DirectoryInfo || actualType == typeof(System.IO.DirectoryInfo))
                 realConstraint = new EmptyDirectoryConstraint();
--- a/src/NUnitFramework/framework/Constraints/EmptyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/EmptyConstraint.cs
@@ -48,19 +48,25 @@ namespace NUnit.Framework.Constraints
         /// <returns>True for success, false for failure</returns>
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
-            if (actual == null)
-                throw new System.ArgumentException("The actual value must be a non-null string, IEnumerable or DirectoryInfo", "actual");
+            var actualType = GetBaseType(actual);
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/NUnitFramework/framework/Constraints/EmptyConstraint.cs [hunk 1]
```diff
@@ -48,19 +48,25 @@ public override string Description
         /// <returns>True for success, false for failure</returns>
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
-            if (actual == null)
-                throw new System.ArgumentException("The actual value must be a non-null string, IEnumerable or DirectoryInfo", "actual");
+            var actualType = GetBaseType(actual);
 
-            if (actual is string)
+            if (actual is string || actualType == typeof(string))
```
CharliePoole: I don't understand the added term. Can you give an example where it is neede?

CharliePoole:
It seems like some new EmptyConstraint tests are needed. I realize that your Assert.That tests test the change indirectly, but we generally try to have direct tests for each constraint. See file EmptyConstraintTests.cs.

jhamm:
Thanks, I agree that direct tests are a good idea. I have added two tests to EmptyConstraintTests.

The previous check `actual is string` will not work if `actual` is null. The change to use `actualType == typeof(string)` will evaluate to true even if `actual` is null.

I have also moved the throw ArgumentException to immediately after the string check.

rprouse:
Looks good. @CharliePoole can you take a second look?

CharliePoole:
We started a discussion of how it should work on the issue, so maybe we need to hold this until that is completed and then check to see if it actually does what we decide the constraint should do.

jhamm:
Based on the discussion in #1171, this can be closed and not merged.

CharliePoole:
In that case, we should at least check to see if our current behavior matches what we defined as desirable in the discussion.

Issue #1212, opened 2016-01-25
Separate AssertionHelper project or assembly


CharliePoole:
Issue #1211 brings up the question. Maintaining a separate syntax using AssertionHelper is a bit of a burden and we have not kept it up to date. Perhaps it's time to drop it.

I'm creating this issue in case there is a silent fan base for AssertionHelper, so that folks may speak up in its favor. If we keep it, we should update it, but let's decide whether we want to keep it first,

A suboption would be to remove it from NUnit and publish it as a separate addon.

rprouse:
I didn't even know we had an AssertionHelper :fearful:

CharliePoole:
I take that as a vote for removal. :-)

oznetmaster:
I never knew we had one either :(

oznetmaster:
What does it do, or what is it supposed to do?

CharliePoole:
Your test class inherits from it, giving you a syntax for assertions that doesn't require a static class name. See, for example, https://github.com/nunit/nunit-csharp-samples/blob/master/syntax/AssertSyntaxTests.cs and scan for "Expect"

As I commented in #1211, it has not been kept up to date with the latest features of the Assert.That syntax. Ideally, if we keep it, I think it should be spun off into a separate project and maintained by people who actually use it.

rprouse:
> As I commented in #1211, it has not been kept up to date with the latest features of the Assert.That syntax. Ideally, if we keep it, I think it should be spun off into a separate project and maintained by people who actually use it.

I agree. How many ways to we want to support writing asserts? Keeping Assume in sync with Assert is hard enough :smile:

appel1:
In many cases you can use "using static nunit.framework.Assert" instead I think. So I don't mind if it was removed.

OsirisTerje:
Agree with the above,  didn't know it existed either.  
Remove or move it out just for safekeeping.

OsirisTerje:
Could we extract the helper in a nuget package - just create one manually for now since it is a kind of probe, and then see how many would download it ?  
If it gets  lets say less than 1000 over a 6 month period, then we drop it.

MSK61:
I use it frequently in my projects, but I'm strongly in favor of phasing it out of NUnit(by either complete removal or moving to subsidiary package). I can easily refactor my projects to do without it; it's just a couple of find and replace for Expect(and the compiler will take care of the rest flagging errors where appropriate).

CharliePoole:
[...]

PR #1250, opened 2016-02-06
Move AssertionHelper tests to a separate class

CharliePoole:
This is in support of #1212 but does not complete it. I'd like to get it merged since it supports whichever path we follow with AssertionHelper by creating a separate test class for it and removing all other references throughout our tests. That way we can either migrate it to a separate project or simply drop it in the future.


rprouse:
Merging

Issue #1231, opened 2016-01-31
Build may silently fail some tests


CharliePoole:
I just discovered this in running `build -t TestAll` locally. I haven't seen it in the CI build, but it would require scanning down all the results to find it.

When we used msbuild for the build, we were able to make use of a setting that allowed further tests to continue if one failed. But at the end, the build was reported as failed. We need to acheive the same effect with Cake.

ChrisMaddock:
@CharliePoole just been browsing build issues, believe this one is done? Apologies if I've missed something.

CharliePoole:
I agree.

PR #1259, opened 2016-02-11
Continue to run remaining Tests when a Test fails and still break the build

dicko2:
This one is related to issue #1231 I was keen to do some work on cake :) but not "blow away" by it so far

I think it will do what you need, I've left the pull request with the failing test so you can see the results of a broken build by a failing test that continues, and will remove if you are happy with the results. Hopefully I've understood the requirement correctly.

I've tested it on my local and in Travis CI, i tried to get it working in TeamBuild, but rolled back as there is some issues with VSTS and the nunit3 xml format that i found out the hard way.

You will see that there is an error midway but it still continues
errorandcontinue

Then at the end of the Cake script in the teardown event it blows up if there has been any error reported during the run.

buildblowsupatend

I tried to get more output from the exceptions in relation to what failed but still working cake out, ow it displays a count of the errors only at the end. I guess you have the test results XML files so what ever you are building it with should be able to consume these for reporting on specifics.


build.cake
```diff
@@ -5,8 +5,14 @@
 var target = Argument("target", "Default");
 var configuration = Argument("configuration", "Debug");
 var framework = Argument("framework", "net-4.5");
+var legacyOutputXml = Argument("LegacyOutputXml", "false");
```
CharliePoole: Leftover from TFS support I guess. We always want nunit3 output for our build anyway.

build.cake
```diff
@@ -160,7 +166,10 @@ Task("BuildCppTestFiles")
 //////////////////////////////////////////////////////////////////////
 Task("TestAllFrameworks")
   .IsDependentOn("Build")
-	.Does(() =>
+    .OnError(exception =>
+{
+    ErrorDetail.Add(exception.Message);
+}).Does(() =>
```
CharliePoole: Formatting. :-)

CharliePoole:
This is great! Normally __nobody__ volunteers to work on the build!

A few inline comments above and some general ones here:

1. Packaging should not run if the tests fail. Maybe it could depend on some sort of task that just checks for errors.

2. When a the framework tests fail for a given build (like .NET 4.5) the remaining builds should run if possible.

dicko2:
Ok, I've added the dependency to the package step. And fixed up the formatting as well as removing that TFS stuff :)

Yea i had a terrible time getting TFS to work, hence all those messy commits, unfortunately with builds its hard to test somethings end-to-end without committing.

I'm surprised they aren't supporting the nunit3 xml format in VSTS, they have been really reactive with updates I've noticed since the 2015 release. Do you have an ETA from MS on this out of interest? 

I "think" the only way to continue building the other frameworks would be to have one task executed per framework, I'm not sure how best to do this off the bat. I'll try work it out today before beer-o'clock ;)

dicko2:
Ok, I didn't realize we were the ones throwing the exceptions, so is was easier than i though, I'm just collecting the exception data into the array then blowing up the build when needed, I also added the framework into the error output so it'll be easier for someone to read the errors when they are all outputted at the end.

buildblowsupnice

Also stopped it from double exploding when the appveyor task is called, which i noticed was happening from my last commit.

If you are happy with this let me know and i'll remove that erroring test i put in there.

And it's not even beer-o'clock yet, I might look through the issues for something else to keep me busy :)

CharliePoole:
There are conflicts with master, so the CI builds can't run. Can you resolve them?

dicko2:
Ok @CharliePoole have merge in all the changes and resolved the conflict.

dicko2:
Also found a context error in the travis CI build, have fixed now and pushed up another commit

CharliePoole:
As you have already seen, the differences between Roslyn and Mono make it very tricky to get a Cake script to work cross-platform. Do you have a local Linux system on which to test? If not, I can pull it down and do some experimenting with it tomorrow.

It would also be nice if you could eliminate the double messaging at the end, but we can live without it.

dicko2:
Yea, cake/mono is a bit of a pain in the butt, you're right I should get a local Linux setup going to test.

I just need to rethink the way I've implemented it so i don't end up making a mess of it.

I'll have another crack at it tomorrow, beer is calling now :)

CharliePoole:
You may be able to get the same failure using the -Mono option to build in Windows.

dicko2:
She's not as clean as I would have liked but its working. I've just used a ref to pass in the Error collection for modifying inside the helper method.

It's still doing that double output thing at the end on appveyor though, going to look into it now, not sure if it will be an easy fix or not.

dicko2:
ok, was an easy fix, I n00bed it again, tried to run code after a throw :)

CharliePoole:
Looks good to me! @rprouse would you also take a look please? Once we both OK it, @dicko2 will have to remove the deliberate failures from the framework tests and then I'll merge.

build.cake
```diff
@@ -159,46 +164,69 @@ Task("BuildCppTestFiles")
 //////////////////////////////////////////////////////////////////////
 // TEST
 //////////////////////////////////////////////////////////////////////
+
+Task("CheckForError")
+    .Does(() => 
+    {
+    CheckForError(ref ErrorDetail);
+    });
```
rprouse: Very minor, but indenting is off here. Also I think you could simplify this and other short one liners down to `.Does(() => CheckForError(ref ErrorDetail));`

build.cake
```diff
@@ -206,28 +234,40 @@ Task("TestDriver")
...
-		RunTest(NUNIT3_CONSOLE, BIN_DIR, V2_PORTABLE_AGENT_TESTS);
+		RunTest(NUNIT3_CONSOLE, BIN_DIR, V2_PORTABLE_AGENT_TESTS,"TestV2Driver", ref ErrorDetail);
 	});
 
 Task("TestConsole")
   .IsDependentOn("Build")
+    .OnError(exception =>
+        {
+            ErrorDetail.Add(exception.Message);
+        })
```
rprouse: Here and other places, I would like to see consistent indenting of the methods and the brackets.

build.cake
```diff
@@ -578,7 +643,7 @@ void BuildProject(string projectPath, string configuration, MSBuildPlatform buil
     }
 }
  
-void RunTest(FilePath exePath, DirectoryPath workingDir)
+void RunTest(FilePath exePath, DirectoryPath workingDir, string framework, ref List<string> errorDetail)
```
rprouse: Why does the list need to be a `ref`, you are not assigning to it, just adding to it? Objects are already passed by reference.

rprouse:
Functionally looks good to me other than minor formatting errors. Thanks for tackling this.

We should probably add `build.cake` to `.editorconfig` to get consistent indent levels for those who use it.

build.cake
dicko2: This is to do with mono/linux compatibility, in .NET/windows from within the method you can access the variables,  but in mono you cannot. So i need to pass this by ref in to get modified. Objects default to getting passed by val (https://msdn.microsoft.com/en-us/library/14akc2c7.aspx), so using ref the modifications to the object will be carried out of the method in the event of an error. The list compiles a collection of all errors to be outputted at the end of the user.

build.cake
dicko2: Ok done, I've been using VSCode for this one sorry, I'm too used to VS/resharper taking care of these things for me lol

build.cake
dicko2: done

rprouse:
All good, I will take another look in the morning, on my phone now. I use VS Code for the cake files too. There is an EditorConfig extension for it now which helps with tabs and indents. Helpful when moving between projects ;)

rprouse:
@dicko2 can you remove the failing test? I think we can merge this once we get CI passing.

dicko2:
ok, failing test removed.

CharliePoole:
Thanks for doing this. Much better to get all the errors at one time!

PR #1262, opened 2016-02-13
#1261: Fix display name generation for generic types with nested classes

### LINKED ISSUE #1261: TypeHelper.GetDisplayName generates the wrong name for generic types with nested classes
Having classes:
```cs
public class Class1<T,S>
{
    public class Class2<K>
    {
        public void T()
        {}
    }
}
```
For variable:
```cs
     var s = new Class1<int, string>.Class2<long>();
```
GetDisplayName returns `"Class1<Int32,String,Int64>"` but it should return `"Class1<Int32,String>+Class2<Int64>"`

Green-Bug:
Fixes #1261.


CharliePoole:
Can you add a test for this situation?

Green-Bug:
Sure, I'll do that.

rprouse:
Looks good to me and the naming looks good.

PR #1263, opened 2016-02-13
Issue 1252

### LINKED ISSUE #1252: Exception thrown when any assembly is not found
This is a follow-up to #1018. The temp fix for that bug gives us a FileNotFoundException rather than an ArgumentException. That is arguably a great improvement, since it is clear what has gone wrong.

However, the intent of our design is that an exception should not ripple up to the runner, but be translated into a TestResult, even in the case when the erroneous test is an assembly. That is not happening here.

The root cause seems to be that we try to load assemblies in many different places in order to check various things. The latest culprit is the code in DomainManager that tries to read the TargetFrameworkAttribute. We need to examine the entire chain of calls that load the tests with several things in mind:

1. Try to minimize the number of times we examine the assembly.

2. Examine the assembly as early as possible.

3. Don't try to go forward with a missing or invalid assembly.

The last point is crucial. Right now, in the case of the simple command `nunit3-console junk.dll` we go so far as to create a separate process, which then discovers that the file it has  been told to load doesn't exist. Ideally, this should be discovered immediately and the correct result returned.

However, it's a bit problematic if the command includes some files that exist and some that don't, along with an option that causes them all to be run in the same process. That's a case that requires a bit of thought.

CharliePoole:
Fixes #1252. Restructures test loading and execution to eliminate the need for reflecting on the assembly in DomainManager.


CharliePoole:
I'll go ahead and merge it as is since it's blocking a few things.

@rprouse this is a fairly important refactoring, so an after-the-fact review wouldn't be amiss when you have the time.

Issue #170, opened 2014-08-06
Test Order Attribute


CharliePoole:
Any test method with an Order attribute will be run before those without one, in the order of the attribute argument.

For example:
  [Order(1)]
  [Order(2)]
  [Order(2)]
  [Order(3)]
before the remaining unordered tests.

Issue #51, opened 2013-11-03
Test Dependency Attribute


ashes999:
Hi,

I have a web app with extensive automated testing. I have some installation tests (delete the DB tables and reinstall from scratch), upgrade tests (from older to newer schema), and then normal web tests (get this page, click this, etc.)

I switched from NUnit to MbUnit because it allowed me to specify test orders via dependency (depend on a test method or test fixture). I switched back to NUnit, and would still like this feature.

The current work-around (since I only use the NUnit GUI) is to order test names alphabetically, and run them fixture by fixture, with the installation/first ones in their own assembly.

circa1741:
I have an idea that is more of a twist for dependency and ordering.

The discussion, so far, regarding dependency is "go ahead and run Test H (and possibly 8 other tests) only if Test 8 passes." In other words, there is no point of running Test H because if Test 8 fails then I know that Test H will also fail.

How about a dependency when a test fails?

Scenario:
I need a Smoke Test that covers a lot of ground. So, I am planning a Test Fixture that is an end-to-end test that has basic coverage of many of the SUT's features. The tests on said test fixture will use Test Ordering and are not "independent." The test order will be Test A then Test B then Test C, etc.

Now, because the tests are not "independent" I know that if Test C fails then all the following tests will also fail. Therefore, I need more tests to run in order to get a bigger picture of the Smoke Test.

I need to be able to configure to run Test 1 if Test A fails, Test 2 if Test B fails, Test 3 if Test C fails, etc.

My Test 3 is designed to be independent of Test B so if this fails then I have a better understanding of why Test C failed earlier. As it turns out my Tests 4 (for Test D), 5 (for E), 5 (for F), etc. all pass. Then I now understand that only the feature that was covered by Test C is the issue.

Why not run Tests 1, 2, 3, etc. instead? Well, because since those are isolated and independent tests then I am not doing Integration Tests. Again, I need a Smoke Test that covers a lot of ground.

Maybe something like:
- [DependsOnPassing("Test so and so")]
- [DependsOnFailing("Test blah blah blah")]

This will allow finer control in my automation test design.

circa1741:
How about something like these attributes instead:
- [DependsOnPassing("Test so and so")]
- [RunOnFail("Test blah blah blah")]

Please note to which test these are attached to. These attributes should be useable in different levels (and in any): assembly, test fixture, test.

[DependsOnPassing("Test E")]
Test F
- Test E will automatically be executed if its result is unknown.
[...]

Issue #663, opened 2015-05-26
Run tests in random order


AndriusJanusas:
Ideally each test should be able to run independently and stand alone, but when you have many of them and they always run in the same order it would be great to run them in random order to ensure, that they're actually independent.

Would it be hard to implement such feature?

CharliePoole:
We have talked about this in the past and it's a little harder than appears at first glance. That's because the "random" order has to be repeatable in order to debug any unexpected dependencies that are discovered. This makes it non-trivial but certainly not impossible. The random ordering would have to be tied in directly to NUnit's own Randomizer, which is used to generate random arguments among other things. In that way, the /seed command-line option or it's equivalent could be used when it came time to debug a particular problem.

[As an aside, we could definitely use some integration-level tests for the repeatability of random values and - if we implement this - test order.]

I'd be inclined to see this as a future enhancement to be added after we implement test ordering itself.

rprouse:
I might even combine this with the planned Order attribute #170. Maybe it could be a property on the Order, something like `[Order(Random=true)]`?

This will prevent the use of random along with an order, but we will need to decide how this will work with other tests that specify an order. Are they random after all ordered tests? Before? Interspersed? Probably after since the order attributes would indicate a requirement for a particular test state that might not be stable if we run random tests first.

CharliePoole:
I guess this depends on how the OrderAttribute works. I saw it as applying to an individual test, not a suite. Randomization OTOH seems like something you would specify on the class affecting all the tests. There are different ways you could handle conflicts but what feels natural to me is having random on the suite simply disable any other ordering at the method level. I can't think of a strong use case for wanting some tests to run randomly and others to run in a strict order.

There's another fundamental issue to resolve about random ordering. Does the random ordering happen when the tests are created or when they are run? If you are running in a future GUI or in the test adapter, should repeated runs re-randomize the tests? Should this be an option? Or is a given order "sticky" once it is generated?

All of this conspires to make me feel this is a bit farther in our future than simple ordering. However, #170 is already marked future, so this one can go the same way without our having to decide how far in the future it is. :-)

richseviora:
How about making random order as a suite/fixture level attribute? That way you'd be able to be explicit about which tests are subject to the randomness, and have that override (or throw an exception) any ordering attributes applied to child tests.

[...]

Issue #423, opened 2014-12-15
Integrate Code Coverage into CI


rprouse:
I just wanted to put this out there as an idea and possibly a future task. I have been experimenting with [OpenCover](https://github.com/OpenCover/opencover) to run code coverage on the project and used [ReportGenerator](http://www.palmmedia.de/OpenSource/ReportGenerator) to generate HTML reports.

So far I've just run one of the framework tests, but we are getting 86.7% coverage which is pretty good. It does highlight some areas that are poorly tested, so I think it is a worthy exercise.

I am using the command line versions of both tools, but they are also available as NuGet packages that add MSBuild tasks.

An alternative might also be to use JetBrains' DotCover. They have a free license for open source teams that we could take advantage of.

CharliePoole:
I have also noticed that the open source version of ncover has had some development activity by other folks since the original developer formed his company.

Since this is clearly something we should do at some point, I'm taking it out of the 'idea' category. I recognize it may not actually get done for 3.0.

jcansdale:
Interesting. Where have you seen this activity?

CharliePoole:
Well, I can't confirm because sourceforge is offline. The odd thing is that I saw activity as late as 2013 but there is no binary release after 2004. I'll check again when I can.

jcansdale:
Wow, Sourceforge has been down for quite a while. it looks like their
Subversion repository is down as well. They appear to be struggling!

agray:
@rprouse, I suggest your initial thoughts of using OpenCover and ReportGenerator are good.  I have used these tools on two previous projects.

We added OpenCover and ReportGenerator to the Solution via NuGet and included the packages in source control so the CI checked them out.

I have virtually generic batch files that can be executed as postbuild steps (we were using Jenkins)

The hardest part is to get the opencover exclusion syntax correct.  It can be a bit tricky and the OpenCover documentation is not the best.

rprouse:
From duplicate #883 

>We should be encouraging best practices, so it would probably be a good idea to determine the code coverage of our tests and automatically publish it in our README.

>Some ideas, use [OpenCover](https://github.com/sawilde/opencover) or [Mono Code Coverage (monocov)](http://www.mono-project.com/docs/debug+profile/profile/code-coverage/) for coverage and use [Coveralls.net](https://github.com/csmacnz/coveralls.net) to report coverage to [Coveralls.io](https://coveralls.io/).

CharliePoole:
Made this a build task.

[...]

PR #1264, opened 2016-02-13
Test order

dicko2:
This is based on the example code given here

https://github.com/nunit/nunit/issues/51#issuecomment-64908239

For issue #170

I originally did the same as issue 170, i.e. [TestOrder(1)], but revised this as later in the issues 170 it refers to 51 to get example code so i have implemented it like this [Test(Order = 1)]

I can move it back to the former if you like, i still have in my commit history.

I've implemented it as a double as well, I have change to an int but I think this is warranted by the comments in 170.

Also to note that I am unsure what to do when tests are mixed with and without orders? Currently its defaulting to 0, so it will technically run Tests without order specified first when mixed, i could change this to MaxInt or something so they would be run last, it's a bit subjective over which one to use so didn't know what to do.

If tests have the same order it will run in order of the test name.


src/NUnitFramework/tests/Attributes/TestOrderAttributeTests.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace NUnit.Framework.Tests.Attributes
+{
+    [TestFixture]
+    class TestOrderAttributeTests
+    {
+        [Test]
+        public void CheckOrderIsCorrect()
+        {
+            var testFixt = TestBuilder.MakeFixture(typeof (TestCaseOrderAttributeFixture));
+            testFixt.Sort();// need to manually call this method to mimic real world conditions
```
dicko2: Just noting on this, I tried with debugging and using the test order inside a fixture in the unit test project and it worked fine, there is the method in the assembly creation that doesn't get called when you use "MakeFixture", so not sure if we should add the sort call somewhere else. I had a though to call Sort every time when a test case is added maybe?

CharliePoole:
You're a brave one! This is an issue that has generated lots of opinions in the past. This will probably take a lot of back-and-forth to complete, more for reasons of what we want to support than for any technical difficulty.

So, here we go...
We plan to have both  ordering and dependency. I see them as separate and I see it as my problem to explain the difference. This is about ordering and will fix #170 when complete. Note that both #51 and #170 end up talking about ordering as well as dependency. It's a concept that's easily confused, so let's stay away from dependency for the moment.

I'm bunching my comments here, so we have an outline of a general plan in one place. Usually, I try to give such a plan before anybody grabs the issue to work on, but I missed the boat here.

1. Let's use separate attribute rather than a property. That's a general direction that NUnit has been taking, to move away from properties of an attribute where possible. For now, let's just call it Order, since it's all about tests anyway. It's possible we might rename it in future.

2. The ordering value should be an int, it's the most natural for most people. If they need to leave a gap, they can easily do that.

3. For now, let's have it apply only to test methods. The IApplyToTest code can do that, simply ignoring a test that is not a TestMethod. We might consider ordering fixtures in a future step.

4. Rather than add the order as a new ITest member, let's store it as a property of the test by adding it to the Properties dictionary. You can do that directly in IApplyToTest or you can derive from PropertyAttribute and add it to the attributes internal dictionary.

5. In the presence of parallel execution, Ordering is **not** a guarantee that a prior test will have finished. IMO, that's OK. Some users may want it different, but let's not worry about it for now.

I see two potential changes in the future that could make us backtrack on this. You'll want to keep them in mind so as not to block them but without adding anything in the code to "allow" for them. (IOW, possible YAGNI)

1. We may create a special fixture type that allows ordering, like OrderedFixture.

2. We may create a new interface like IApplyToWorkItem and order the work items rather than the tests. That would eliminate the sorting of tests. WorkItems are what we actually execute.

All of the above kind of sets you back a bit, but I hope not by too much.

@rprouse anything you want to add to this?

dicko2:
All good @CharliePoole, and yea i wasn't planning on touching the dependency in this scope, I agree it should be a different requirement .

Just one point though. With storing the order in the properties bag, thats easy enough, but doing the sort it will end up looking something like this snip from my linq pad

```csharp
myList.Sort((x, y) => int.Parse(x.PropertyBag["order"][0].ToString()).CompareTo(int.Parse(y.PropertyBag["order"][0].ToString())) != 0 
 ? int.Parse(x.PropertyBag["order"][0].ToString()).CompareTo(int.Parse(y.PropertyBag["order"][0].ToString())) : x.Name.CompareTo(y.Name));
 myList.Dump();
```
And she's looking a little messy already, and hits a key not found error in the event that the key is not set, as well i prefer using TryParse, which isn't possible inline with the sort. The problem here is there is no "easy" way to sort them without ending up with big inline ifs or creating a helper method to extract the order from the Test object as i can see it. So i think the cleanest way forward is to expose a property on the Test case that returns the Order out of the properties bag, which kinda puts us back where i started.

This way was well we can easily control the value that is returned when its not set, to make it either 0 or Max Int.

Just thinking out loud... I Guess i could write a custom sort delegate to pass in... hmmmm... I'll see what i can come up with.

dicko2:
Yep custom delegate did the trick. All of the above completed. Still two unanswered questions though...

1. If tests without order are mixed in with one with order where should they be placed?... currently those without an order come first.

2. Should we re-order the tests each time a new test is added?.. currently sorting is only occurring on assembly creation.

dicko2:
Also @CharliePoole i was going to look at these 2 next, if that's ok. I've got some spare time next weekend as well.

#663
#423

If you would like to update them with any information before i start this time.

src/NUnitFramework/framework/Attributes/OrderAttribute.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+
+namespace NUnit.Framework
+{
+    public class OrderAttribute : NUnitAttribute, IImplyFixture, IApplyToTest
```
CharliePoole: I don't think this should imply a fixture. That would mean that any methods with an Order attribute, even if it's not a test method, would make the class a fixture.

src/NUnitFramework/framework/Internal/Tests/Test.cs
```diff
@@ -246,6 +246,11 @@ public bool IsSuite
         /// </summary>
         public virtual object Fixture { get; set; }
 
+        /// <summary>
+        /// optional order in which to run the test.
+        /// </summary>
+        public double TestOrder { get; set; }
```
CharliePoole: Left over?

src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
```diff
@@ -100,7 +100,15 @@ public void Sort()
         {
             if (!MaintainTestOrder)
             {
-                this.tests.Sort();
+                this.tests.Sort(delegate (ITest x, ITest y)
+                {
+                    var xKey = 0;
+                    var yKey = 0;
+                    if (x.Properties.ContainsKey("order")) int.TryParse(x.Properties["order"][0].ToString(), out xKey);
+                    if (y.Properties.ContainsKey("order")) int.TryParse(y.Properties["order"][0].ToString(), out yKey);
```
CharliePoole: Should use PropertyNames.Order

src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
```diff
@@ -100,7 +100,15 @@ public void Sort()
...
+                    return xKey.CompareTo(yKey) != 0
+                    ? xKey.CompareTo(yKey) : x.Name.CompareTo(y.Name); ;
```
CharliePoole: I'm not sure about retaining the name sort, but see the general comment before making more changes.

src/NUnitFramework/tests/Attributes/TestOrderAttributeTests.cs
CharliePoole: I actually didn't understand what you meant here. What is the method in the assembly creation that doesn't get called? MakeFixture, of course, is only a test utility, so we can easily make it do whatever is needed.

CharliePoole:
I'm a bit unsure about continuing to sort tests to control the order. It's something I have always considered to be a hack to get rid of that the display order of the tests is the same as the execution order. This may be the time. Let me look at this a bit more... I'm still thinking about sorting work items rather than tests, but I need some time to look at it.

src/NUnitFramework/framework/Attributes/OrderAttribute.cs
dicko2: ok removed

src/NUnitFramework/framework/Internal/Tests/Test.cs
dicko2: doh! yes fixed

src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
dicko2: fixed

src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
dicko2: My 2 cents.. I recommend this should be retained, because in the previous issue comments someone has listed a work around as using A_ B_ etc for the test name, so you may well have some customers out there that are using this, so if you change the order it may break someones test when they upgrade. But it's up to you mate :)

CharliePoole: Regarding sort by name... we've been telling people that workaround is a
bad one and will not be supported in future for many years. However, let's
leave it for the moment. I still want to look at WorkItem ordering.

CharliePoole:
OK, I looked this over and here's my thoughts on what we should do...

We sort the tests by name because users asked to see them that way in the gui, many years ago. Eventually, I think  that sort should move to the gui and should apply to the tree nodes, not the tests themselves. For now, let's leave the sort unchanged.

Instead, let's sort workitems. I think the place to do it is in the CreateChildWorkItems method of CompositeWorkItem. This would use a delegate just as you now have with one more level of indirection to get the Test property of the work item.

This will accomplish the goal of separating the order of execution from the order of test display. I'll take it the next step - removing the test sort entirely - once the gui is doing it's own sorting.

Does this make sense?

dicko2:
Yep ok, I'll have another crack at it

CharliePoole:
I think it should be a pretty simple change.

dicko2:
So you have testbuilder to standing up tests for unit test, do you have anything similar for work times that i can't see? 

if not i'll just try creating something

dicko2:
ok, i think i found it nm

dicko2:
Ok so I am a bit stuck here. I can see that its ordering correctly, and executing them in the correct order.

However I've set the parallel to none to avoid problems, but i can see that the SimpleWorkItemDispatcher is launching each test in it's own thread, so they are finishing at different times, which makes the results come in at slightly different times to when the tests were executed. So the tests randomly fail/succeed, as I am validating that the test results are in the correct order as i can't access the work item order publicly.

image

I guess my question is which setting should I be using to execute them synchronously consistently? Or am i going about this the wrong way?

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -232,12 +232,29 @@ private void RunChildren()
...
                     _children.Add(WorkItem.CreateWorkItem(test, _childFilter));
+            Sort();
         }
 
+        /// <summary>
+        /// Sorts tests under this suite.
+        /// </summary>
+        private void Sort()
+        {
+                this._children.Sort(delegate (WorkItem x, WorkItem y)
```
CharliePoole: SortChildren? Coding standard: don't use this. The leading underscore is sufficient to make it clear.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -232,12 +232,29 @@ private void RunChildren()
...
+                {
+                    var xKey = 0;
+                    var yKey = 0;
+                    if (x.Test.Properties.ContainsKey(PropertyNames.Order)) int.TryParse(x.Test.Properties[PropertyNames.Order][0].ToString(), out xKey);
+                    if (y.Test.Properties.ContainsKey(PropertyNames.Order)) int.TryParse(y.Test.Properties[PropertyNames.Order][0].ToString(), out yKey);
+                    return xKey.CompareTo(yKey) != 0
+                    ? xKey.CompareTo(yKey) : x.Test.Name.CompareTo(y.Test.Name); ;
```
CharliePoole: Don't sort by test name here. It has nothing to do with work item order. Just run all ordered items first followed by ordered items. You may want to initialize the keys to MaxValue for that purpose.

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -242,10 +242,6 @@ private TestSuite BuildTestAssembly(Assembly assembly, string assemblyName, ILis
 #endif
             testAssembly.Properties.Set(PropertyNames.AppDomain, AppDomain.CurrentDomain.FriendlyName);
 #endif
-
-            // TODO: Make this an option? Add Option to sort assemblies as well?
-            testAssembly.Sort();
-
```
CharliePoole: This sort should remain as is, until we change it in a separate issue. It is not intended to provide execution order, but to control the display order of tests in the gui.

src/NUnitFramework/framework/Internal/PropertyNames.cs
```diff
@@ -133,6 +133,10 @@ public class PropertyNames
         /// </summary>
         public const string IgnoreUntilDate = "IgnoreUntilDate";
 
+        /// <summary>
+        /// The optional Order the test will run in
+        /// </summary>
+        public const string Order = "Order";
         #endregion
```
CharliePoole: No longer being used.

src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
```diff
@@ -94,42 +94,6 @@ public TestSuite(Type fixtureType)
...
-            foreach( Test test in Tests )
-            {
-                TestSuite suite = test as TestSuite;
-                if ( suite != null )
-                    suite.Sort(comparer);
-            }
-        }
-#endif
-
-        /// <summary>
```
CharliePoole: Should all be reinstated - you can just use git reset on the file.

dicko2:
Sorry @CharliePoole , miss read your point about you doing the removing of the test sort, will restore that.

CharliePoole:
See inline comments for some small stuff. More general comments here:

1. You have not updated the Silverlight or Portable builds. Are you working in Linux? If so, that's OK, we will do it as a separate step after you're done.

2. It would clean this up if you just reset the files that don't get "real" changes to their last commit before you started. Easier than typing stuff back in and creating false changes.

3. Generally, for an attribute, I create a test of the attribute alone. In this case, it only really needs to verify that the attribute creates a property entry for the Order.

4. Testing the sort is really a test of CompositeWorkItem.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
dicko2: done

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
dicko2: removed, and I've added in maxint. However i am still getting an issue with those test without order being in the wrong order. Not sure if its the issue i mentioned before or not. Still trying to work it out

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
dicko2: restored, sorry about that

src/NUnitFramework/framework/Internal/PropertyNames.cs
dicko2: Aren't we still storing the order data in the tests? Or should create a new property on the work item for it you reckon?

CharliePoole:
Yes, the removal of test sort has to be coordinated with adding its equivalent to the gui.

CharliePoole:
@dicko2 From your comments, it looks like you are following this by email. I did a substantial edit of the comment on how to test this and I'm not sure if you will get that update by mail. You can view it online of course.

I would be curious to know what you get in the mail when I edit a comment.

src/NUnitFramework/framework/Internal/PropertyNames.cs
CharliePoole: My mistake. My original thought was to add a property to work item and an entire new interface that allowed attributes to  modify work items, but that's obviously more complicated than we need.

dicko2:
Yep, no worries @CharliePoole a bit to take in, I'm getting started on it now, but might not finish till tomorrow because its almost beer-o'clock :)

I got the partially completed version only on email, but can see full edit online, didn't get a followup email with the edit.

dicko2:
Did you close the pull request by mistake though?

CharliePoole:
Yes... I keep thinking it means Close the window!

dicko2:
Ok, so I've got a chunk of the above done, with still a bit of tidy up left for tomorrow, however... removing the test name ordering at the workitem level means that the second level order becomes random for tests that don't have an order. Because the work item level all of the test CompareTo 0, so their order seems to vary randomly and not use the original order form the Sort on the TestSuite level, example below of a historical test case that is untouched and now failing that I've fixed in my latest commit.

image

Another example of one that I haven't fixed yet.

image

I thought it might just be a random thing with one of the instances of test data, so i fixed one by adding the new order attribute in, then noticed another was broken, fixed that, then found two more.

I can fix up all the test cases adding the new order in, but i think all these failing historical test cases might indicate that we have a problem that will make the new version not backwards compatible with some peoples projects, or cause unexpected results.

I would recommend adding the Test Case Name order back into the Work Item SortChildren method as a secondary sorting. Alternatively i can fix all the historical test cases up with the order attribute and maybe you can just put something in next versions release notes to warn people maybe. What do you think Boss? ( @CharliePoole )

Also to note you can't see "all" the failing tests in the CI build yet, but i think there is another pull request in to fix that ;)

I'll wait for you to confirm before I go any further.

CharliePoole:
Obviously - except I didn't think of it earlier - we should not be sorting WorkItems if none of the tests have the Order attribute. We can check that as we create them. That way, only folks who use Order will be affected.

Taking it a bit further, it seems like we should be able to sort in a way that doesn't break our tests. Isn't there a stable sort we can use instead?

For the issue of backward compatibility, my own view would be to stick to my guns: this has never been a feature but just an accident of implementation. We've been telling people that for years and that they should not rely on alphabetical ordering. I'll bet we could find 100 instances of this on various lists. OTOH, there are probably just as many SO posts where testing kiddies tell people to do this and even get up-voted for it!

@rprouse as you are taking this over, maybe you should put in your opinion here. :-)

@dicko2 Let's at least avoid sorting if there is no use of Order. Jow about this:
1. Add child items without Order to the end of the list
2. Insert child items with Order at the start, and count them
3. If you have more than one ordered item, do the sort using List.Sort(int, int, IComparer) so that the remaining items are not affected

oznetmaster:
Will there be a way to force alphabetical ordering if that is desired?

CharliePoole:
@oznetmaster I'd be against it myself.

oznetmaster:
Why is alphabetical ordering any worse then any other arbitrary (non-specified) ordering, or random ordering, for that matter?

CharliePoole:
History I supppose. We have never guaranteed it as the execution order. In fact we never supported any ordering of tests. Now we propose to support it in some limited way. Once this code is done I expect we'll have some serious discussions about how widely to support it at all. I think this whole business of accidental alphabetical ordering will simply make it harder to have the discussion. We can 're-address it subsequently as a deliberate choice much more fruitfully I think.

oznetmaster:
However, with a GUI, there is an ordering implied by the order in which the tests are displayed, whether it is historically guaranteed or not.

If the GUI (as most do) utilizes alphabetic ordering for it tree display, then that seemingly should be available as a (default) test ordering as well.

oznetmaster:
This may or may not be relevant, but I have been talking to users of my NUnit 3.0 GUI who have been using it for the last year, and 100% of them said they had an expectation that the tests would run in the same order as they are displayed in the GUI - alphabetical (which is currently the case).

I had never specifically claimed that would be the case, or even discussed execution order in my documentation.

iCodeIT:
By guaranteeing the order of execution you will also indirectly support that tests depend on each other, which they should not. It is nice to show the tests in alphabetical order in the gui, because it makes it (a lot) easier to locate a specific test. And yes - in the GUI you would expect the tests to be run in order, as has always been the case - but only because you can follow the execution and see how far you are.

oznetmaster:
Not really guaranteeing and order of execution such as alphabetical does not imply in any way that the tests have any interdependencey.

And the reason for the tests being run in display order in the GUI is exactly what you indicate, and should be the default.

appel1:
As far as I know the Visual Studio adapter, R# runner and console runner do not execute tests in alphabetical order. Does anyone really have any problem with that? At least the R# runner has a progress bar so you can track how far along you are despite tests not getting executed in the same order as is currently shown in the GUI.

CharliePoole:
In fact, it's not possible in the general case to execute tests as some ui displays them. The framework doesn't know what you is using it to run tests.

CharliePoole:
S/you/ui/

oznetmaster:
Actually, it is very simple.  If you have the ability to make tests run in the order specified in the -tests command line option, then the GUI can just list them in the desired order.  That provides the connection between GUI and test order.  Same is true if the list of tests to run is provided through any interface.

CharliePoole:
Indeed **that** would be possible. And even relatively simple once we implement a way of specifying test order.

That's not what this issue or discussion is about. It's about a proposed feature that will provide an experimental way of ordering tests in NUnit **for the first time**.

I was responding to your assertion that our default execution order should match the order used by the gui. That's not possible. There are many GUI runners. It's better design if we do what they tell us to do.

Our default remains what it has been: test execution order is indeterminate. Unless it is told otherwise, NUnit is allowed to run tests in any order it likes.

But again, none of that is what this PR is about. It's about adding one new feature, a new attribute, to NUnit. There are some things we have to decide about that feature, which we aren't getting to while having this discussion. Unfortunately, I just got sucked into it as this comment shows, but now I'd like to get back to the PR.

dicko2:
Ok cool, so after all that the remaining change is, if order is not used in any of the tests don't call sort? 

I think that's a good solution because it means the old tests won't break. I'll jump on it tomorrow and get it cleaned up.

CharliePoole:
Sounds right to me. :-)

dicko2:
All done, and unit tests are passing

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -45,7 +45,11 @@ public class CompositeWorkItem : WorkItem
         private TestCommand _setupCommand;
         private TestCommand _teardownCommand;
 
-        private List<WorkItem> _children;
+        public List<WorkItem> Children;
+        /// <summary>
+        /// A count of how many tests in the work item have a value for the Order Property
+        /// </summary>
+        public int CountOrder;
```
CharliePoole: Why is this public?

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -231,13 +236,32 @@ private void RunChildren()
...
+            Children = new List<WorkItem>();
 
             foreach (ITest test in _suite.Tests)
                 if (_childFilter.Pass(test))
-                    _children.Add(WorkItem.CreateWorkItem(test, _childFilter));
+                {
+                    Children.Add(WorkItem.CreateWorkItem(test, _childFilter));
+                    if (test.Properties.ContainsKey(PropertyNames.Order)) CountOrder++;
+                }
+
```
CharliePoole: Did you try my original suggestion of Inserting the ordered items at the start and Adding the non-ordered ones at the end? Once you do that, you only have to sort the first n items in the list, all of which have an Order property. This is supported by List.Sort and would leave the remaining items untouched. Additionally, ChildrenCount would then probably just be a local variable, which makes more sense since it's not used anywhere else.

oznetmaster:
I realize that this is late in the game, but I started trying to apply this to some of the tests in a few TestFixtures.

It appears that maintaining the test ordering rapidly becomes very difficult if there are a lot of tests in the TestFixture (> 100 for instance).  I see a preliminary mention about deciding to use an integer for ordering because it is "natural".  It seems to be to be more "natural" to specify that TestA must come either before or after TestB.  That make the ordering independent of other orderings specified, or new tests added.  It also removes the need to leave numeric gaps, that may or may not be sufficient. It is easy to develop a dependency tree to actually determine ordering (or ordering errors).

Also, it is unclear.  Is the ordering constrained to be only within a specific TestFixture, or is it global across all Tests and all TestFixtures in the assembly (or multiple assemblies)?  If it is currently global, should some form of constraint be possible?

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
dicko2: Yep, i was looking at that, I didn't realize until i re-read your post that Sort has the ability to do first N items, never used this before, I should RTFM more often :)

Had to create a class for backwards compatibility with .NET 2 as the Create method on Comparer is only available in 4.5, so its a little ugly but working.

dicko2:
@oznetmaster I believe dependencies are going to be done as a separate feature to ordering.

However I do think that a data type like double would be more flexible, in case you end up in a jam.

oznetmaster:
I am not talking about dependencies, ordering only.

A dependency would be TestB depends on TestA, and TestB would only be run of TestA was successful.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -231,13 +240,50 @@ private void RunChildren()
...
+                {
+                        Children.Add(WorkItem.CreateWorkItem(test, _childFilter));
+                        _countOrder++;
+                }
+
+            foreach (ITest test in _suite.Tests)
+                if (_childFilter.Pass(test) && !test.Properties.ContainsKey(PropertyNames.Order))
+                {
+                    Children.Add(WorkItem.CreateWorkItem(test, _childFilter));
+                }
```
CharliePoole: Why two separate loops? List.Insert will allow you to stick things at the front of the list as you go through the tests once.

src/NUnitFramework/tests/TestUtilities/TestBuilder.cs
```diff
@@ -135,6 +135,29 @@ public static ITestResult RunTestSuite(TestSuite suite, object testObject)
             return work.Result;
         }
 
+        public static CompositeWorkItem GenerateWorkItem(TestSuite suite, object testObject, ParallelScope? parallelScope)
+        {
+            TestExecutionContext context = new TestExecutionContext();
+            context.TestObject = testObject;
+            if(parallelScope != null) context.ParallelScope = (ParallelScope)parallelScope;
```
CharliePoole: ParallelScope does not seem to be relevant at all here. Tests run by testbuilder, as here, are executed by telling the workitem to execute them directly. ParallelScope is only relevant if the tests are to be run by a parallel dispatcher.

src/NUnitFramework/tests/TestUtilities/TestBuilder.cs
```diff
@@ -135,6 +135,29 @@ public static ITestResult RunTestSuite(TestSuite suite, object testObject)
...
+            {
+#if PORTABLE
+                System.Threading.Tasks.Task.Delay(1);
+#else
+                Thread.Sleep(1);
+#endif
+            }
+
+            return work;
+        }
```
CharliePoole: Did you find you actually had to execute the work items in order to check the order? I was guessing you would just call CreateChildWorkItems and then check without executing.

CharliePoole:
@dicko2 Further inline comments... you picked a pretty hard one to start working on. :-)

@oznetmaster We'll need to try it on CF eventually, but do you see anything problematic in the code?

CharliePoole:
@oznetmaster Putting this Issue in context: we had a number of requests pertaining to test order and dependency, which I persist in thinking are two related but nonetheless different things. We decided to do an initial release with only an order attribute. Dependency is scheduled for a further release.

If somebody is doing functional testing where the individual methods represents steps in a process, I agree that some sort of dependency tree is the right solution. If the dependency constrains order of execution tightly, then NUnit will discover and follow that order. If it's looser and allows NUnit to make choices, NUnit will do that too. An order attribute can simulate this at the cost of a great deal of effort on the part of the user, as you point out. But a simple order attribute is something we can do simply and quickly and will benefit at least some users - or so I hope.

As you point out, it's easy to derive ordering from a dependency tree. That's not why dependency is hard for us. The key difference is that dependent items will usually require that the preceding items completed, and most likely succeeded. Possibly we will implement some settings that control the nature of the dependency in each case but I think "test a depends on test b having run successfully" is likely to be the most common usage. That requires a whole layer of infrastructure in running tests that we don't yet have. The present PR, OTOH, is suitable for implementing right in the chain of events created by our current WorkItem infrastructure.

Does that make sense?

CharliePoole:
@oznetmaster This only has to do with ordering tests within a fixture.

src/NUnitFramework/tests/TestUtilities/TestBuilder.cs
dicko2: Yep. I think that was left over from some testing i was doing, have removed.

CharliePoole:
@oznetmaster You wrote...

"It seems to be to be more "natural" to specify that TestA must come either before or after TestB. That make the ordering independent of other orderings specified, or new tests added. It also removes the need to leave numeric gaps, that may or may not be sufficient. It is easy to develop a dependency tree to actually determine ordering (or ordering errors)."

That describes what this  group has always called "Test Dependency."

src/NUnitFramework/tests/TestUtilities/TestBuilder.cs
dicko2: Just tried it and if Execute tests isn't called i get a null reference from the child Test collection.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
dicko2: Yep, changed to 1 loop

dicko2:
@CharliePoole I didn't understand why you said i was "brave" doing this one at first, but now seeing all the comments on the PR i can see that while the concept of ordering is simple, it's implementation is a hot topic :)

src/NUnitFramework/tests/TestUtilities/TestBuilder.cs
CharliePoole: OK. That makes sense.

oznetmaster:
Test dependency involves determining if one test runs depending on the result of another test.

Implicitly, test dependency does specify test ordering, but the converse is not true.  It is possible to specify TestB afterTestA without specifying thatTestB will only run if TestA is successful.  That might be TestB after success TestA.  Similarly, there could be a dependency TestB after failure TestA.

It would seem that implementing ordering based on these semantics would greatly simplify adding the dependency characteristics at a later point in time, but the ordering is in no way reliant on the concept of dependency.  Ordering is static before any tests are run, dependency is dynamic as tests are run.

CharliePoole:
Code looks good to me now. We need one more committer to OK this. @rprouse ?

Finally, before merging, let's ask whether we want to release with OrderAttribute as a full-fledged, independent feature. Ordering has had much discussion in the past so I think we should be cautious. Once we release something, we're stuck with it.

Things we might do...

1. Only support it in some special fixture (OrderedFixture ?)

2. Make it experimental for now by putting it in NUnit.Framework.Experimental. That way we can feel comfortable changing it if we need to later without worrying about backward compatibility.

Thoughts @nunit/core-team ? @nunit/contributors ?

CharliePoole:
@oznetmaster Since we decided to handle OrderAttribute here and dependency under issue #51, I'm moving the discussion of dependency to #51.

oznetmaster:
I think this is totally an OrderAttribute discussion, and not a dependency one.

I think the OrderAttribute should not use a value, but should use a semantic ordering.

I think using a value will set a bad precedent, and we will be stuck with it.

If you want to put this in experimental, or somewhere else where no one will "depend" on it, that is ok for now.

CharliePoole:
@oznetmaster I think you are basically saying we should not have an OrderAttribute as defined in this discussion so far, but a DependsOnAttribute of sorts. Rather than give TestA Order(1) and TestB Order(2) we would annotate TestB with DependsOn("TestA"). Is that right, at least modulo the names?

IIRC, MbUnit always had both. Is that right @rprouse ?

Isn't it possible that some users prefer to express it one way and some another?

Sorry 'bout all this @dicko2 !

oznetmaster:
Sort of :(

DependsOn on implies some sort of execution dependency.

I would use DependsOn for just the, and execution dependency.

I would use ExecutesAfter (again, name for exploration purposes only) to indicate an execution order, with no dependency (does not care about what the results are from TestA, or even if it has an execution error with no results).

oznetmaster:
I am fine with supporting both.  I just have a real preference for the semantic ordering over the ordinal one.  I have dealt with ordinal ordering in many other applications, and it always eventually becomes a nightmare to maintain.

CharliePoole:
@oznetmaster I agree with you on the implications of DependsOn.

As I imagined using it myself the OrderAttribute would be a partial ordering. There would be a few tests that should run first, followed by those without the attribute. This construct comes from JUnit and has been implemented in a few other frameworks. It's really more like a priority than anything else since you could simply mark tests as 1, 2 or 3 and have duplicates.

@rprouse You input would be helpful here - so far it's just the three of us. :-)

I lean toward putting this into Experimental for 3.2 - as a warning that the implementation may change.

dicko2:
@CharliePoole how about using params int[] on the Attribute constructor?

So your order attribute on the test could look like this, or you could use just the single int.

```csharp
[Order(1,0)]
public void TestMethod1()
{
 //Test here
}

[Order(1,1)]
public void TestMethod2()
{
 //Test here
}

[Order(1,2,1)]
public void TestMethod3()
{
 //Test here
}
```
This would allow people to slot in tests in-between others, just thinking out loud.

oznetmaster:
Maybe it should be a PriorityAttribute?  That would deemphasize the ordinal ordering aspect.  Just a thought.

CharliePoole:
@dicko2 Seems too complicated...
@oznetmaster Maybe so...

We need some more voices in here @nunit/core-team @nunit/contributors !

iCodeIT:
I think that it would be nice if order-attribute and dependency-attribute could have the same syntax. Also I think that it is better to refer to other tests instead of using a number (for all the reasons stated earlier)

Order attribute should say: This test is after "that test" (but we don't care i "that test" fails. 

Dependency atrttibute says: This test depends on "that test" - i.e. "this test" only runs if "that test" succeeds. Further if "this test" is selected to run, then "that test" should (maybe?) implicitly also be selected to run.

CharliePoole:
Research on test order in other test frameworks...

JUnit 4.11 - you can specify an attribute on the class, which specifies the algorithm to be used to order the tests. One algorithm is to use an order attribute specified on the methods.

TestNG - uses dependsOnMethod and dependsOnGroups

MbUnit - supports ordering via an integer value as well as dependsOn

xUnit.net - doesn't support it but folks have extended to use a TestPriority property

CharliePoole:
Still hoping @rprouse will chime in, but if we were going to 3.2 tomorrow, I would opt for changing from OrderAttribute to PriorityAttribute and putting it into the Experimental namespace.

oznetmaster:
I agree with that.

rprouse:
There are so many potential ways to implement this, I don't think we can ever make everyone happy. This issue has been open here for several years and has been requested for much longer. Everyone has had plenty of time to comment on the issue or implement it themselves. I think arguing to this extent when someone has finally made the effort to implement this as we agreed in the issue is counter productive.

My initial interpretation of the OrderAttribute was that it would be the same as MbUnit's possibly extended to a double. Personally, I always incremented by 10 and never had a problem adding new tests, so was happy with an int. But then again, I started programming in Basic on a Commodore Pet :smile: If your tests are carefully ordered, how often are you going to add that many tests into the middle of the order? Even if you do, you just need to bump the next higher test up.

Someone else stated that you might have hundreds of ordered tests. Really? A class with hundreds of methods that must be specifically ordered? If you are dealing with that level of complexity, then I think you have bigger problems than worrying about the order attribute. Besides, if you are really worried, increment by 1000, That still allows over 2 million test methods. Seriously guys, are we just creating complexity for the sake of an elegant solution?

If we release and people can show us that it is limiting them, then we can consider extending the Order attribute with a string overload that implies the test that it depends on. Personally, I don't think it will be needed, but prove me wrong. If we keep going back and forth on this, then nothing will get done and you will be stuck ordering your tests with the old hack of alphabetically naming your methods.

I am going to review this, if it is good merge it and then we can put this behind us. If you disagree, put your money where your mouth is and submit a pull request :smile:

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -45,7 +45,15 @@ public class CompositeWorkItem : WorkItem
         private TestCommand _setupCommand;
         private TestCommand _teardownCommand;
 
-        private List<WorkItem> _children;
+        /// <summary>
+        /// List of Child WorkItems
+        /// </summary>
+        public List<WorkItem> Children;
```
rprouse: If we are going to promote this to public, it should be made a property with a private setter and change the type to `IList`. Can it be made internal?

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -231,13 +240,52 @@ private void RunChildren()
...
+            }
+        }
+
+        /// <summary>
+        /// Sorts tests under this suite.
+        /// </summary>
+        private void SortChildren()
+        {
+            Children.Sort(0, _countOrder, new WorkItemOrderComparer());
+        }
```
rprouse: Super minor, line break between methods.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -231,13 +240,52 @@ private void RunChildren()
 
         private void CreateChildWorkItems()
         {
-            _children = new List<WorkItem>();
+            Children = new List<WorkItem>();
 
             foreach (ITest test in _suite.Tests)
```
rprouse: Now that this foreach is getting longer, can you add brackets so that the scope is clear. We only use this form for short blocks like it was before.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
CharliePoole: It's public for testing - could easily be internal. I think I told @dicko2 to make it public, 'cause I tend to forget about using InternalsVisibleTo.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -231,13 +240,52 @@ private void RunChildren()
...
-                    _children.Add(WorkItem.CreateWorkItem(test, _childFilter));
+                {
+                    if (test.Properties.ContainsKey(PropertyNames.Order))
+                    {
+                        Children.Insert(0, WorkItem.CreateWorkItem(test, _childFilter));
+                        _countOrder++;
+                    }
+                    else
+                    {
+                        Children.Insert(Children.Count, WorkItem.CreateWorkItem(test, _childFilter));
```
rprouse: This inserting ordered items at the beginning of the list and non-ordered items to the end works, but feels like premature optimization to me. The [underlying implementation of List<T>](http://referencesource.microsoft.com/#mscorlib/system/collections/generic/list.cs,cf7f4095e4de7646) is an array, not a linked list, so repeatedly inserting at the beginning is likely more expensive than the eventual sort since it requires constant reallocations and array copies.

src/NUnitFramework/testdata/TestCaseOrderAttributeFixture.cs
```diff
@@ -0,0 +1,45 @@
...
+using NUnit.Framework;
+
+namespace NUnit.TestData
+{
+    [TestFixture]
+    public class TestCaseOrderAttributeFixture
+    {
+        [Test]
+        [Order(3)]
+        public void Z_ThirdTestWithSameOrderAsSecond()
```
rprouse: You say third test with same order as second, did you intend this to have `Order(2)`?

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -231,13 +240,52 @@ private void RunChildren()
...
+            /// A signed integer that indicates the relative values of <paramref name="x"/> and <paramref name="y"/>, as shown in the following table.Value Meaning Less than zero<paramref name="x"/> is less than <paramref name="y"/>.Zero<paramref name="x"/> equals <paramref name="y"/>.Greater than zero<paramref name="x"/> is greater than <paramref name="y"/>.
+            /// </returns>
+            /// <param name="x">The first object to compare.</param><param name="y">The second object to compare.</param>
+            public int Compare(WorkItem x, WorkItem y)
+            {
+                var xKey = int.MaxValue;
+                var yKey = int.MaxValue;
+                if (x.Test.Properties.ContainsKey(PropertyNames.Order)) int.TryParse(x.Test.Properties[PropertyNames.Order][0].ToString(), out xKey);
+                if (y.Test.Properties.ContainsKey(PropertyNames.Order)) int.TryParse(y.Test.Properties[PropertyNames.Order][0].ToString(), out yKey);
+                return xKey.CompareTo(yKey);
```
rprouse: If `xKey` and `yKey` are both `int.MaxValue`, should we switch to ordering by test name so that people who are using the old alphabetical test order hack will not be broken?

rprouse:
I have reviewed and like this implementation. I've made a few minor formatting points. The only real problem is the initial pre-sort which I think is less efficient than not doing it. Should we just use a `SortedList<T>` to start and skip the sort at the end? Less efficient if there is no `OrderAttribute`?

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
CharliePoole: I pushed for this, not because of optimization, but so we would not change the existing order of non-ordered items.

An alternative would be the two-loop approach that @dicko2 originally used: first add the ordered items, then add the unordered items. Then sort the ordered items only. Alternatively, do the sort before adding the unordered items! If List.Sort were stable (retained pre-existing order for equal items) then we wouldn't have to do this.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
CharliePoole: That creates a hidden binding between the original sort of  tests and this sort. IMO, it's better to leave those items unsorted. Of course, once they add OrderAttribute, they have to use the semantics of that attribute anyway.

I may be the only one, but I am totally unsympathetic with anyone who relied on the alphabetic sort hack, which we've been telling people not to use for years.

CharliePoole: BTW, he originally was using a secondary sort by name, but I made him take it out. :-)

I do feel that the WorkItems should not need to know how Tests were originally sorted as a matter of the SRP.

CharliePoole:
@rprouse I think a lot of these considerations depend on how we expect it to be used. My thought (hope?) was that this would be little used and that only some tests in a fixture would be ordered. If large numbers are involved we can optimize later.

Not sure what you mean by the pre-sort? The initial sort of tests? If so, I hope to remove it in a future issue, adding a sort by name to the Gui for display purposes. That's the only reason we sorted tests in V2 as well as here: so they would display that way in the Gui. But display order and execution order are not necessarily the same thing.

BTW, we discussed this on the PR, but the comments are attached to obsolete commits so you don't see them now.

Beyond this PR, I'd like to use something like the JUnit approach: a fixture-level attribute that tells us
to run in a particular order. Options might be: ByName, Random, UsingOrderAttribute.

oznetmaster:
@CharliePoole I like that idea.  Perhaps both fixture level and a command line option.  That would resolve many of my objections. Maybe sooner then later.

rprouse:
@CharliePoole I was wondering why this didn't replace the alpha-sort, your comments explain it. Based on that, I think there are just minor formatting issues remaining and we could merge with those since they are pretty minor.

src/NUnitFramework/testdata/TestCaseOrderAttributeFixture.cs
dicko2: Originally I had the sort by name as a secondary sort, so was testing that, but as Charlie mentioned before he requested it removed, I've updated the wording on the method now to just Third Test.

dicko2:
Commodore PET @rprouse ? that's real old school, I didn't start until I got a C64, peek and poke FTW :)

I've done most of the fix-ups, this new code lens in VS2015 has me missing my gaps between methods a lot, I should be checking my diffs before commit.

But there was the last one I didn't really get a consensus from you and charlie on. Did you want me to add back the order by name? or change the sorting of the list?

Originally a LOT of the unit tests started failing when I removed the order by name, but with the method Charlie got me to do the sorting with, they are now all passing so we can be confidant that while its not explicitly sorting by name its ending up in that order if you aren't using the order attribute.

CharliePoole:
My first was a KIM-1 board, with 6502 processor and 1 1/8 K (yes!) memory.

Sorry to confuse you @dicko2 - we'll leave the alpha order out of this PR, as discussed. 

@oznetmaster I'll create an issue for that - soon but not for next week's release I'm afraid.

rprouse:
@dicko2 It sounds like @CharliePoole has us both beat with the KIM-1 board. I don't even remember that one :laughing: I had to stay late at school to use the PET computers. The first one I bought myself was a TI-99/4A. I should have gone for the C64 though :+1: 

As @CharliePoole says, don't worry about the alpha sorting. Actually, with the fixes you've already pushed, I am good to merge. Do you have anything outstanding that you want to push before i merge?

dicko2:
Nope, nothing outstanding merge away

CharliePoole:
@oznetmaster has pointed out several issues about this PR, the most serious of which is that it won't compile under CF. I'll create a new PR to fix this.

@oznetmaster I'm wondering about your comments coming in so late. Is there something in how you are set up to be notified that causes you to frequently comment on PRs after the merge rather than before? You were part of the discussion on the PR itself and I asked you about CF compatibility in a comment a few  days ago. 

Additionally, as we have previously discussed, your comments frequently come in on the commits rather than on the PR, which makes it hard for others to see them. But now I'm wondering if this is not part of the same problem. Are you only dealing with PRs via email? If so, that would explain why you don't see the code changes...

I'm not criticising here, just trying to diagnose the situation. Your comments are potentially very useful if we get them while the code is being worked on rather than afterwards. FWIW, TryParse has been in the code we were reviewing for about a week. If it's a question of your time, I understand of course. But if it's a question of how you are getting notified, it seems worth fixing.

oznetmaster:
@CharliePoole 

First, I often never look at the actual code until I merge it into my CF build, which as I have pointed out before is a manual process.  Going the opposite direction is fully automated.  I usually do this about once a week, and only for code which has been merged into the master branch.  Hence the comments on code after it is merged.  When I am merging code, I am only dealing with the merged commits, so that is where my comments end up.

A large portion of the time I am responding to PRs and issues by email on my iPad, which is all I carry with me when I am out of my office.  I rarely take my laptop.  I find it difficult to try to review code on the iPad.

Not sure how to improve the process :(

dicko2:
@CharliePoole I'll get that item above fixed up shortly.

@oznetmaster I'd be keen to look at doing a CI build for CF, I like working on builds :) and we do some CF development at my workplace currently and have some CI builds running for it, so I have a little bit of experience in that area. 

But I'd like to understand what you have setup. You mentioned mentioning it before, is there some old issues i can look at? is everything in the repo that you use already?

Issue #1210, opened 2016-01-23
Available runtimes issues


CharliePoole:
The list of available runtimes may include `.Net 4.5 - Client`, which does not exist. This is due to our misinterpreting a registry entry intended for backword compatibility.

In addition, we return both `.Net 4.0 - Client` and `.Net 4.0 - Full` if registry entries exist for both. This is not exactly incorrect but it is misleading. If `Full` is installed, then it includes the assemblies and types for `Client`. However, we have no way to force the running of tests under `Client` only if `Full` is installed.

It may be best if we merely return a single entry for .NET 4.0, suffixing it with `Client` only if the full version is not installed. It would also be useful if the RuntimeFramework type provided a property indicating whether it represents the full or the client profile.

I marked this High priority and put it into the 3.2 milestone because it impacts the gui.

ChrisMaddock:
@CharliePoole Have started digging in to this if that's ok, it seems fairly self contained which I thought might make a good start.

One question, is the client/full profile information only required in AvailableFrameworks? It looks like CurrentFramework doesn't currently check - should it do, or is it not relevant at that point?

CharliePoole:
That's great. I assigned to you.

Try to keep this no more complicated than it deserves since client profile has now become a legacy thing. For now, I think it only impacts AvailableFrameworks. If we come up with a need for it, in CurrentFramework, we can add that later. Not much point making distinctions if we don't do anything about them.

PR #1267, opened 2016-02-13
Issue 1210

ChrisMaddock:
First attempt at #1210.

Charlie, as you suggested, have attempted to keep the changes as small as possible - this is only in Available frameworks for the engine, and the profile is only specified for .NET 4.0 versions.

The current tests for AvailableFrameworks rely on actual installed frameworks, so I wasn't sure how best to test this. I considered checking that AvailableFrameworks were unique by version number, so not both client and full exist for 4.0, but that seemed fairly vague and a bit over the top. Not sure if it's necessary to test explicitly - open to suggestion there!

### DIFF
--- a/src/NUnitEngine/nunit.engine/RuntimeFramework.cs
+++ b/src/NUnitEngine/nunit.engine/RuntimeFramework.cs
@@ -53,6 +53,19 @@ namespace NUnit.Engine
     }
 
     /// <summary>
+    /// Enumeration identifying profile of .NET Framework
+    /// </summary>
+    public enum Profile
+    {
+        /// <summary>Profile Unspecified</summary>
+        Unspecified,
+        /// <summary>Client Profile</summary>
+        Client,
+        /// <summary>Full Profile</summary>
+        Full
+    }
+
+    /// <summary>
     /// RuntimeFramework represents a particular version
     /// of a common language runtime implementation.
     /// </summary>
@@ -92,6 +105,7 @@ namespace NUnit.Engine
                 InitFromClrVersion(version);
 
             DisplayName = GetDefaultDisplayName(runtime, version);
+            Profile = Profile.Unspecified;
         }
 
         private void InitFromFrameworkVersion(Version version)
@@ -295,6 +309,11 @@ namespace NUnit.Engine
         /// The CLR version for this runtime framework
         /// </summary>
         public Version ClrVersion { get; private set; }
+
+        /// <summary>
+        /// The .NET Framework Profile for this framwork, where relevant
+        /// </summary>
+        public Profile Profile { get; private set; }
 
         /// <summary>
         /// Return true if any CLR version may be used in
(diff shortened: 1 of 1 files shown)


src/NUnitEngine/nunit.engine/RuntimeFramework.cs
```diff
@@ -53,6 +53,19 @@ public enum RuntimeType
...
+    /// </summary>
+    public enum Profile
+    {
+        /// <summary>Profile Unspecified</summary>
+        Unspecified,
+        /// <summary>Client Profile</summary>
+        Client,
+        /// <summary>Full Profile</summary>
+        Full
+    }
```
CharliePoole: Should this be an enum or a simple string? In the MS class FrameworkName, it's just a string. In theory, this allows us to add any new profiles that come up more easily. Of course, we have to be aware that client is a subset of full and that it only exists in certain versions.

Also, I wonder if Unspecified should actually mean Full?

@rprouse What do you think?

src/NUnitEngine/nunit.engine/RuntimeFramework.cs
```diff
@@ -608,24 +627,25 @@ private static void AppendOlderDotNetFrameworkVersion(List<RuntimeFramework> fra
...
                     var framework = new RuntimeFramework(RuntimeType.Net, new Version(4, 0));
-                    framework.DisplayName += " - " + profile;
+                    framework.DisplayName += " - " + profile.ToString();
+                    framework.Profile = profile;
                     frameworks.Add(framework);
 
                     var release = (int)profileKey.GetValue("Release", 0);
                     if (release > 0)
                     {
                         framework = new RuntimeFramework(RuntimeType.Net, new Version(4, 5));
```
CharliePoole: Looks as if you will potentially add 4.5 twice.

CharliePoole:
See line comments. Additionally, does anyone know what shows up for versions beyond 4.5? If it's not easy, could be a subsequent issue.

Regarding tests, I never figured out a better way to test this, so I just list them on the console for a manual check and also make sure that the current framework is included. That's actually multiple tests, since we run the tests for multiple builds of the framework.

src/NUnitEngine/nunit.engine/RuntimeFramework.cs
ChrisMaddock: My logic with enum/unspecified was that it seemed wrong to specify all the non-.NET frameworks as 'full profile' when there wasn't such a concept. Can change that if you think it's best kept simple however!

src/NUnitEngine/nunit.engine/RuntimeFramework.cs
ChrisMaddock: @CharliePoole The loop should be broken out of once frameworks have been added? i.e. if full frameworks are installed, it won't check for client. Have I perhaps missed something here - not sure I can see it?

I agree it's not the clearest method really, although again I went with minimising changes - do you think it would be better written out sequentially? For the three frameworks it's not so much duplication, looking at .NET4.6, looks like 4.0 is really the exception, and others follow a standardised patter.

ChrisMaddock:
Thanks @CharliePoole.

> Additionally, does anyone know what shows up for versions beyond 4.5? If it's not easy, could be a subsequent issue.

For 4.6 is just a change of the DWORD value under the same keys - meaning the code that detects 4.5 here could actually be detecting 4.6. Is that an issue?

src/NUnitEngine/nunit.engine/RuntimeFramework.cs
CharliePoole: OK, I missed that - you're right, it's not clear, but rewriting would add a lot of duplication. Maybe just add a comment in front of the return about why we can return at that point.

CharliePoole:
Looks good to me. We'll wait for @rprouse comments.

src/NUnitEngine/nunit.engine/RuntimeFramework.cs
ChrisMaddock: Great, have put that in.

ChrisMaddock:
Assuming the CI has just timed out again somewhere? Last commit only added a comment.

src/NUnitEngine/nunit.engine/RuntimeFramework.cs
ChrisMaddock: Have looked into this a little more. MS behaviour appears to be to not specify full profiles, only the restricted ones, which might lend something to @CharliePoole's comment that unspecified should mean full.
http://stackoverflow.com/a/3899324/1768779

The only issue I can think of then, is that in other methods such as CurrentFramework which don't check for Client Profile, we could potentially incorrectly specify .NETv4Client as Full, whereas at the moment Unspecified makes it clear the check doesn't take place. I don't know how this is used downstream however, maybe that will affect how it's best specified. :-)

CharliePoole:
@nunit/core-team I'd like to merge this. Can somebody give it a second review please?

CharliePoole:
Since I'm going to be away for a few days, I decided to go ahead and merge this so other work can proceed.

Issue #1181, opened 2016-01-09
BadImageFormatException when no extensions and unmanaged dll in directory


ChrisMaddock:
If there is no .addins file in the test directory, the ExtensionService looks though all dlls in that directory for the ExtensionAttribute. If there is an unmanaged dll in the directory, a BadImageException is thrown. (Running under the console runner)

Workaround is to include a blank .addins file.

````
System.BadImageFormatException: Format of the executable (.exe) or library (.dll) is invalid.
   at Mono.Cecil.PE.ImageReader.ReadOptionalHeaders(UInt16& subsystem, UInt16& dll_characteristics)
   at Mono.Cecil.PE.ImageReader.ReadImage()
   at Mono.Cecil.PE.ImageReader.ReadImageFrom(Stream stream)
   at Mono.Cecil.ModuleDefinition.ReadModule(Stream stream, ReaderParameters parameters)
   at Mono.Cecil.ModuleDefinition.ReadModule(String fileName, ReaderParameters parameters)
   at NUnit.Engine.Services.ExtensionService.FindExtensionsInAssembly(String assemblyName)
   at NUnit.Engine.Services.ExtensionService.FindExtensionsInDirectory(DirectoryInfo startDir)
   at NUnit.Engine.Services.ExtensionService.StartService()
   at NUnit.Engine.Services.ServiceManager.StartServices()
   at NUnit.Engine.TestEngine.NUnit.Engine.ITestEngine.get_Services()
   at NUnit.ConsoleRunner.ConsoleRunner..ctor(ITestEngine engine, ConsoleOptions options, ExtendedTextWriter writer)
   at NUnit.ConsoleRunner.Program.Main(String[] args)
````

rprouse:
Why is there a 32 bit assembly in the addins directory? I am inclined to close this because we cannot anticipate handling everything someone might drop into the addins directory. Can you explain your use case?

ChrisMaddock:
Hi,

There is no addins directory - the ExtensionService appears to search the directory of the executing assembly. Looking at the nuget package, it's only seems to be redirected to an addins directory by the content of the .addins file?

We are referencing the nunit3-console.dll and copying local, so it is in the same directory as all our other dlls. I do however seem to have been too quick to jump to the 32-bit conclusion, Mono seems to be able to handle those. (Which makes sense.) Will close this temporarily until I can pin down exactly what's unusual about the particular dll causing the crash.

ChrisMaddock:
[...]

Issue #1009, opened 2015-11-08
Cecil exceptions from engine


CharliePoole:
The engine can throw exceptions due to errors in Cecil Resolve(). I discovered this when working with the NUnit 3 VS Adapter. The errors are very hard to locate and occur when Cecil cannot resolve the item requested. The code assumes null will be returned rather than an exception being thrown.

We should find every place where we do a Resolve in the engine code and 
1. Catch it and take appropriate action
2. Try to set up the resolver correctly so it doesn't occur.

CharliePoole:
Issue #1144 points to some code where we simply don't check the possible null return at all!

CharliePoole:
Issue moved to [nunit/nunit-console #9](https://github.com/nunit/nunit-console/issues/9) via [**ZenHub**](https://www.zenhub.com/)

PR #1268, opened 2016-02-13
Issue 1181

ChrisMaddock:
This catches the crash found in #1181. Under certain circumstances, the ExtensionService will scan every dll in the executing directory to look for extensions. If an unmanaged dll is encountered at this point, NUnit crashes.

(#1181 was closed as a sub-issue of #1009, although this PR only catches the single case found in #1181 - I haven't done anything further to investigate other cases.)

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
+++ b/src/NUnitEngine/nunit.engine/Services/ExtensionService.cs
@@ -246,7 +246,7 @@ namespace NUnit.Engine.Services
                     ProcessAddinsFile(startDir, file.FullName);
             else
                 foreach (var file in startDir.GetFiles("*.dll"))
-                    FindExtensionsInAssembly(file.FullName);
+                    FindExtensionsInAssembly(file.FullName, false);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


CharliePoole:
This looks good to me, with the possible exception that maybe we should log any errors. @rprouse what do you think?

CharliePoole:
Something else I just thought of... we can get to the code in question in two ways: (1) scanning a directory using a wild-card pattern and (2) because the file was explicitly listed in a `.addins` file. In the first case, it makes sense to silently ignore the error, but in the second it seems wrong. What do you think @ChrisMaddock and @rprouse ?

ChrisMaddock:
@CharliePoole Thanks, the second use-case hadn't crossed my mind.

Sounds to me like `FindExtensionsInAssembly()` needs to know if it's scanning or looking at a specified assembly, and either skip, or throw an NUnitEngineException() specifying the invalid addin?

CharliePoole:
That's what it seems to me. If the file was actually listed by name, it's a configuration error.

ChrisMaddock:
Thanks @CharliePoole - have added that in.

CharliePoole:
Looks good to me. We need a second review. @rprouse ?

Issue #681, opened 2015-06-10
NUnit agent cannot resolve test dependency assemblies when mixed mode initialization runs in the default AppDomain


rationull:
I've run into an issue where NUnit fails to load a mixed mode test assembly due to code running during managed assembly initialization (in the "cctor" as covered here: https://msdn.microsoft.com/en-us/library/ms173266.aspx) which is dependent on another, collocated assembly. Part of the assembly initialization process (controlled by the mixed mode C runtime library) is forced to run in the default AppDomain, which in this case prevents dependent assembly resolution since the application base directory is nunit-agent's, not the test assembly's.

To reproduce, create a test assembly with a mixed mode static global variable whose initialization depends on code in a collocated assembly. I've created a minimal repro solution which should be simple to build and demonstrate the problem (may require tweaking the reference to nunit.framework.dll to build). The repro solution can be found here: https://github.com/rationull/NUnitReferenceFailure

To work around this problem, I've copied the NUnit binaries into the same directory as my test assembly, mingling the NUnit launcher's application base directory with my test assembly's.

I propose that to fix the issue, the NUnit launcher (nunit-agent in my case but I'm not clear on whether the agent is the only context in which tests are run at this point) should allow optional installation of an AppDomain.AssemblyResolve event handler which would probe the test assembly's directory. This could have unexpected side effects in the general case, but as an opt-in feature any complications could be avoided except when this functionality is required. I have hacked on my own build of nunit-agent as a proof of concept to verify that this does fix the problem, although I have not done any kind of exhaustive testing.

I have reproduced the problem, the workaround, and the suggested fix in NUnit 2.6.4 and the current 3.0 build.

If you consider my fix proposal valid, but it does not meet the priority bar for implementation by the core team, then any design suggestions which might help an external pull request be accepted are welcome.

Note: For additional context see the later parts of the discussion here: https://groups.google.com/forum/#!topic/nunit-discuss/7ZBQ02dcIBg

CharliePoole:
Thanks for filing this. It helps to nail down what has been a long-standing but poorly understood problem. We'd welcome a pull request to resolve this issue, some ideas on implementation follow...

[...]

PR #1272, opened 2016-02-16
For review: fix for issue 681 using a new assembly attribute and defa…

rationull:
This is a followup to PR #1137 . The custom attribute detection is rewritten to use Mono.Cecil and moved to RuntimeFrameworkService in line with the new design for assembly inspection and package settings.

Notes from the previous PR still apply:

 * DirectTestRunner has to avoid installing the hook if it's run from a non-default AppDomain, in order to avoid installing a nonfunctional hook when running NUnit's own automated tests. The hook only makes sense to install in the default AppDomain given the issue at hand.
 * Paths are removed from the resolver after running tests. This seems to work but I'm not 100% clear that it's impossible to re-run the tests without loading a new DirectTestRunner.
 * The resolver hook is never actually uninstalled. This shouldn't be a functional problem but it would be cleaner to uninstall it.

Also comments on issue 681 re the lack of automated testing for this fix still apply as well.


CharliePoole:
This looks right to me. I'm at a retreat right now, so I won't be able to actually exercise this as I would like until the weekend. The following comments are just from code-reading:

1. Would it make sense to do this at the process level rather than in DirectRunner? That is, we could require such assemblies to  run in a separate process. Then the code to create the hook would be in the agent and would always be run in the primary appdomain.

2. Is there any way to detect the specific situation with mixed mode C++ initialization even if the attribute is not present? We can look at a lot of things in Mono.Cecil, but I don't know just what to search for here.

3. Have you got a test project for this feature?

rationull:
Thanks for the review.


1. I don't see why that would be a problem. There's nothing intrinsic that requires a separate agent process to work around this issue although doing it that way would avoid polluting assembly resolution in the main NUnit process and resolve the issue with unhooking the resolver (it would be easier to unhook at process shutdown, I would think, but also wouldn't really matter as much). I assume this would be in RemoteTestAgent or something else called by it.

2. I can do some thinking on this but I'm not sure. The "nuclear option" would be to use the new behavior for all mixed-mode assemblies that refer to other managed/mixed assemblies that are not in the GAC. This would trigger the workaround in many cases that do not trigger the original issue which seems questionable to me, but it is an insidious issue and easy to trigger so maybe that solutions is not uncalled for. It would certainly avoid confusion on the part of people who would be debugging this.

As to better detection: The root cause is a dependency on another assembly in code that is run in the default AppDomain on assembly load. Theoretically it would be possible to detect by IL static analysis if we could map out all the code that is run in this phase of mixed mode assembly initialization. As far as I can tell this is an implementation detail of the VC++ mixed mode runtime so this could change between tools versions although I'm not sure how likely it is to change. In practice this also requires things like finding and analyzing static variable constructors which I think would quickly get very deep and complicated. I suppose it could even be dependent on config files or other side effects making it impossible to detect all cases.

In short it may be best to keep the manual workaround if we want to only surgically change behavior when absolutely required, or to cast a wider but simpler net if appropriate for automatic detection.

3. Yes, I have a test project here: https://github.com/rationull/NUnitReferenceFailure. TestAssembly.dll from that build triggers the issue, and adding the new attribute to the assembly in TestAssembly.cpp triggers the fix and no longer hits the crash.

Note that in this particular test assembly the problem is triggered by the C++ static variable "static MixedClass sMC" in TestAssembly.cpp, where MixedClass's constructor references ReferencedAssembly.dll. This is an example of a case that would require somewhat deep analysis to detect.

I would think this could easily be made into a test project as part of the automated test builds but I'm not sure of all the implications of adding a mixed mode C++/CLI project for Mono builds.

rationull:
The more I think about it, due to all the different ways this can happen (any referenced assembly load triggered at the right time, including loads due to deriving from a class in a neighboring assembly, for instance) the right solution without needing an attribute (or setting or whatever) might just be effectively to look for a failed load potentially caused by this problem and then step in to resolve it -- which basically boils down to just installing the assembly resolve hook all the time for mixed mode test assemblies.

Still seems somewhat invasive for a fringe issue. Perhaps we could reduce the likelihood it would cause undesired behavior by somehow only enabling the hook during the critical part of assembly loading, or only when loading an assembly referenced by the test assembly (based on a Cecil analysis) to prevent the hook from affecting anything down by the NUnit agent itself.

CharliePoole:
I think I agree with you about the risks of trying to predict whether this is needed. This looks good to me now, except for comment below that might delay merging it. In any case, we need one more committer to review it. @nunit/core-team ?

The "except" part...

I'm in the process of adding an AssemblyResolver for an entirely different purpose, related to assemblies that are referenced by extension assemblies. This may make merging a bit difficult. Whichever PR is merged first, the other one will need to make some changes. Ultimately, we should probably only have one central resolver in the primary domain, possibly a service. Let's see how each of our PR reviews go.

src/NUnitEngine/nunit.engine/Internal/ProvidedPathsAssemblyResolver.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System;
```
CharliePoole: Needs our standard copyright/license header

rationull:
@CharliePoole whatever reason you're adding an AssemblyResolver for, I would guess it's a more important (less fringe) issue than this one -- it seems fair that I should have to make the changes to adapt to your design whenever your other PR is finished.

src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs
```diff
@@ -86,10 +98,19 @@ protected override TestEngineResult LoadPackage()
...
+                {
+                    _assemblyResolver.AddPathFromFile(testFile);
+                }
+
                 IFrameworkDriver driver = driverService.GetDriver(TestDomain, testFile);
                 driver.ID = TestPackage.ID;
                 result.Add(driver.Load(testFile, subPackage.Settings));
                 _drivers.Add(driver);
+
+                _packageIndex.Add(driver, subPackage);
```
CharliePoole: I'm not quite following why we need _packageIndex. In the (rare, I imagine) case where you are loading multiple test assemblies into the same AppDomain, you set up the resolver for all the packages before running any. Why not wait till all of them run and then removing all those entries?

rationull:
@CharliePoole any comment on whether it would make sense to simply always enable this behavior for mixed mode assemblies, vs keeping the new attribute?

CharliePoole:
I just copied your resolver into my branch, so the fixup will probably be easy.

CharliePoole:
@rationull I'm waiting for some other comments on that issue. So far it's just the two of us talking here!

src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs
rationull: I like that simplification -- I will implement it when making other review changes.

rationull: Actually, does this contradict the idea of sharing a single, central resolver or does it mesh well with it? Once your central resolver is set up for dependencies of extension assemblies, will we be removing all resolution locations after running all the tests in that context as well, or will we need to separately keep track of which paths to remove in each case (or otherwise not share the resolver)?

CharliePoole: It makes it a little more complicated. Each user of the resolver would have to remove any entry that they had created at the appropriate time. However, for extensions, the appropriate time probably coincides with process exit. Leaving entries shouldn't break anything, unless there are duplicate assembly names, but it would be inefficient.

I just merged my PR with master. I think you should update your branch. I don't think there will be conflicts and the two resolvers should work OK together.

Once your work is done, we can look at cleaning up the architecture. Since we are pushing for 3.2 next week, that would be afterwards.

rationull: Sounds good. You'd also mentioned moving my logic from DirectTestRunner into something at the process level (presumably in RemoteTestAgent) but I have not looked at what would be required to always require these assemblies to run in a separate process. I will probably do a batch of changes first leaving that change out.

rationull: I should clarify on my previous question though, simply having each user track which entries they added and then remove those would be in error if it's possible for multiple users to add the same path. The shared resolver would need to count how many adds and removes have been issues for each path. Minor complication.

rationull: I should clarify on my previous question -- having each user track which entries it added and then remove those entries would be in error if multiple users can submit the same paths. Minor complication, the resolver would have to either track how many adds and removes were issued for each entry or it could allow duplicated entries.

rationull:
My first update hit a conflict with rprouse's commit, and I slightly botched the ensuing merge (mercurial guy still getting used to git -- that's my excuse!) but it looks to be resolved now.

Still pending items:

1) Lack of automated tests.
2) Possible move of the resolver logic to the process level rather than the test runner level.
3) Shared resolver as a service.
4) Still possibly open option of using the new behavior for all mixed mode assembles and dropping the new attribute.

CharliePoole:
I think those pending items are all other issues. That's literal: we'll create other issues and other PRs to do them later. Let's get this in to resolve issue #681.

CharliePoole:
I think this can be merged and we can go on from there, most likely post 3.2. The use of the attribute is OK for now. Even though it would be a breaking change to remove it, I think the usage of it is low-volume enough that we can risk that.

@rprouse Will you review as well?

rprouse:
Initial code review looks good, but I need a bit more time to read through the history of this PR and the issue. I will do that today and merge once I have done so and re-reviewed.

CharliePoole:
@rprouse At the moment, I feel that it is technically ready to merge but that we have to decide whether we want to trigger the use of assembly resolve by means of a special attribute. I think the alternative would be to require use of a command-line argument.

rationull:
A command line argument seemed cleaner to me initially but I'm not sure how well it'd fit into varied runner environments like Visual Studio plugins. If we guaranteed execution in a separate agent process then that would never be a problem but I don't know how those runners tend to work.

Further given that it really is the condition of the assembly that requires this behavior it seems like even abandoning the attribute and triggering for all mixed mode assemblies would be superior (keeping in mind the resolver only runs when the default resolution paths have already failed). JMO.

rprouse:
I don't like the command line argument. For things like this, I think it should behave the same no matter how or who runs it. Do you agree?

CharliePoole:
Yeah... I'd say merge as is. We'll do a bit of refactoring after the release.

rationull:
Is there anything I can further clarify or research wrt whether it's worth exploring automatically enabling for mixed mode assemblies? (Can apparently be detected via the ILOnly module attribute although I haven't verified yet). It doesn't take much for a mixed mode assembly to trigger this problem as long as it has dependencies on other assemblies, as described in my comment here: https://github.com/nunit/nunit/issues/681#issuecomment-168088149

Obviously could be done after the merge but would also be trivial to do before the merge assuming it works.

CharliePoole:
The thing is, once we merge we plan to release NUnit 3.2. So we want to be pretty sure of whatever we put in. It may make more sense to put it in our build right after the release. Then we will have time to discover problems - so long as somebody exercises the feature.

What do you think @rprouse ?

rprouse:
I also think we should wait until post 3.2 for any further changes. I am going to merge now. If there is anything else that you feel should be in 3.2, let me know and we can discuss doing it separately.

PR #1276, opened 2016-02-19
Package extensions using nuget

### LINKED ISSUE #1133: Provide a way to install extensions as nuget packages
Right now, there is no good way to separately install extensions. About the only thing you could do is to download a zip file, copy it to the proper location and update the addins file.

We need a mechanism to install our own extensions, either using the runner itself or a separate extension management program. The underlying mechanism could be nuget or we could provide our own - nuget would be more generally useful, so long as we can take control and have the extensions end up in a useful location.

Requirements...

* We should locate installed engines and allow the user to decide which ones get the extension.

* Either we should copy the extension to each addins directory or put it in some folder of its own and update the addins file for each engine. Alternatively, we could add a new addins file.

This is high priority because expanding the use of addins and distributing them separately depends on it.

CharliePoole:
This is a fixes #1133, dealing with nuget packages only. It's ready for review but we should discuss a few things before merging.

Included:
1. Improvements to wildcard processing to enable use of '*' in a directory name.
2. NUnit.Console package no longer includes extensions
3. NUnit.Console and NUnit.Engine package include an addins file that points to any extensions installed via nuget.
4. Each of our four existing extensions has a separate package.

I have tested installing the console runner package with and without extensions. There may still be some issues with the v2 driver extension - I'll work on setting up a better test bed for that one.

We need to decide if the separation into discreet packages as I have done it is the way we want to go. Obviously, that's what I think, but there may be other opinions about bundling extensions.

There are some file naming issues I want to address separately.

There may be some issues with error reporting when a needed extension is not present.

Since this is all about installing, it will be best if a few people actually install the nuget packages and exercise them in addition to doing code review.


rprouse:
I am a bit uncomfortable changing the behaviour of the current nunit.console package by removing all of the existing extensions. People who rely on any of those extensions and upgrade are unlikely to be aware of this change and need to waste valuable cycles figuring out what has changed.

Would it be better to maintain the current behaviour by including current extensions as dependencies in the existing nuget package and maybe release an extensionless package? The current package could even be switched to a meta-package that depends on the extensionless package and the extensions.

rprouse:
The code itself looks fine.

Further to my last, it looks like the nunit.console package won't even run unless you know that you need to include a result reporter package. I don't think that we should expect people new to NUnit to have to know stuff like this. It should just work out of the box, so even a minimal console package should contain what it needs to run.

A second concern is the engine package. It contains no extensions, but I don't think it allows you to add extensions. If I am writing a runner that uses the engine package, I add it and the engine/API is automatically referenced because they are in the lib folder. Extensions are added to a tools folder though so they are there for the console. If my runner wants to include some of those extensions, I will have to manually copy them out of the tools folders.

Is this supposed to work for that scenario where I am writing a runner and I want to include the engine and a selection of extensions?

CharliePoole:
@rprouse 
"Changing behavior..." Yes, that's what I meant about needing to decide about packaging. However, I think we should test this as is to ensure that it runs in the worst case - without extensions. I have and it would be good if you could do it as well. I like the idea of making the current package into a meta-package, especially if we are going to release soon.

"Console package won't run..." I've run it. The nunit3 writer is constructed like an extension but it's plugged in manually. Same is true of the nunit3 framework driver.

"Engine package..." It depends how it is used. I need to experiment more but here is what I did for the gui... I added an addins file to the gui itself that points where it needs to point to pick up any addins. This is only intended to work for the development of the gui. Once the gui is released, it would have a package that works just like the current (w/o extensions) engine package. IOW, working __on__ the gui (or any runner) is different from working __with__ it.

BTW, in the case of the gui this is made complicated because I am currently referencing the engine itself, not just the api. That's do to a separate issue I need to resolve with the unavailability of a list of available frameworks through the api. I'll do that next.

CharliePoole:
Seems like we have two issues here: where we want to be ultimately wrt packages and what we feel we can do for 3.2. I'd like to move to using 3.2 packages for the GUI, so that means I need the engine and the extensions. If we aren't sure about the console, it can stay as it is for now, although I would add to it an extra addins file so it can get additional extensions beyond the bundled ones via nuget.

CharliePoole:
Quick status:

1. I'm working on the packaging changes so we have backward compatibility. No problems there.

2. I have realized that there is a basic problem with the V2 driver when the extension is stored separately. At runtime, the nunit.core assembly is not actually available to it. That's because we put the addins directory on the probing path in the original installation. But extensions not in the addins directory that have helper assemblies have to use LoadFrom to load them - or we have to do it for them. It's probable it will come up with future extensions as well.

@rprouse If you get to where you want to move ahead with the release, let me know and I can leave that particular extension in the addins directory.

CharliePoole:
I tried to do some cleanup of the build script, extracting a method to package nuget files. This failed on Linux due to a rather common problem with Cake using the Mono compiler. Globally declared variables are not always accessible in helper methods. I did a `push -f` so those commits are no longer visible.

Bottom line, you need both linux and windows boxes handy to work on the Cake script!

nuget/nunit.console-runner.nuspec
```diff
@@ -10,10 +10,18 @@
     <projectUrl>http://nunit.org</projectUrl>
     <iconUrl>http://nunit.org/nuget/nunitv3_32x32.png</iconUrl>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
-    <summary>Console runner for version 3.0 of the NUnit unit-testing framework.</summary>
-    <description>This package includes the nunit3-console runner and test engine for version 3.0 of the NUnit unit-testing framework.
-    
-    The runner will run 2.6 tests using the included plugin.</description>
+    <summary>Console runner for version 3.0 of the NUnit unit-testing framework with common extensions.</summary>
```
rprouse: This looks like the text for NUnit.Console. This shouldn't say that includes the extra extensions.

CharliePoole: I'll fix.

Issue #1278, opened 2016-02-20
Fix optional parameters in TestCaseAttribute


ChrisMaddock:
e.g.

```
[TestCase(1, 2, 3)]
[TestCase(1, 2)]
public void Addition(int a, int b, int c=1)
...
```

Attempted this the other day, to find the reflection method throws a `TargetParameterCountException`. Is there any design-reason why this shouldn't be supported?

ChrisMaddock:
Actually, I wonder if this is a bug, rather than a feature request?

`CheckTestMethodSignature()` [seems to look for optional parameters](https://github.com/nunit/nunit/blob/66770788ea6ee1e970efd65aba81ccae23af8cd8/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs#L159). The issue is perhaps just that the reflection `Method.Invoke` should be passing in `Type.Missing` in place of optional parameters.

CharliePoole:
Yes, this feature was added for 3.0 alpha-5 by PR #457, resolving issue #257. Tests were added to TestCaseBuilderTests.cs in that PR. You should compare those tests to yours to see what broke. Since you have started a PR for this, I'll go ahead and assign to you.

CharliePoole:
@ChrisMaddock You have to precede the ref to the issue (#1278) wtih the word fix or fixes in order to make it close automagically when the PR is merged. Closing manually

PR #457, opened 2015-01-12
Issue #257 - support for optional test method parameters in TestCaseAttr...

hiredgunhouse:
...ibute.
Fix for issue #257. I finally got around to it, sorry for the delay. Any comments/suggestions are welcomed.

PR #1279, opened 2016-02-20
Allow TestCase attribute to work with optional parameters

ChrisMaddock:
Ref #1278.

Allows the `[TestCase]` attribute to work with optional parameters. Previously the `NUnitTestCaseBuilder` would flag the test as runnable, yet it would crash with a `TargetParameterCountException` when calling `MethodInfo.Invoke()`.

This doesn't fix the issue for `[TestCaseSource]` - however, that also doesn't appear to support the other special cases `[TestCase]` does (e.g. params). 

I had a quick look at doing both attributes - but it looks like there's additional complexity for that because of how TestCaseSource interprets it's argument types based on `GetParameters().Length`. I'd imagine the `GetParametersForTestCase(IMethodInfo method)` in `TestCaseAttribute` could be shared, but `TestCaseSource.GetTestCasesFor(IMethodInfo method)` will need a some modification. Given what's involved, I've left this for a separate issue.


rprouse:
@ChrisMaddock, minor point before I review. In the future, could you give the PR a more descriptive name? In cases like this where an issue is not fixed, I will assign the PR to the milestone and the title of the PR will be used in the changelog.

It also makes scanning the PR list easier. I'll never remember what issue #1278 was :smile: 

I have edited this issue, so you don't need to.

rprouse:
@ChrisMaddock, this looks fine to merge once you respond to my one comment.

This makes me think that we should add a future enhancement to `TestCase` to allow it to work with named optional parameters.

rprouse:
@ChrisMaddock in the issue, @CharliePoole points out that  #457 was the initial implementation for this issue. It has numerous unit tests. Can you identify why those tests are working but yours weren't? What is the difference?

CharliePoole:
@ChrisMaddock 
Regarding "bombing out"... I agree that the method of throwing an exception and immediately catching it to save in the ParameterInfo is not very pretty. :-) At some point we should try to improve this but I don't think this is the point.

"TestCaseSource" I agree that it's a separate issue. Once we work on that, we might end up pushing some of the special handling for Type.Missing down a level, maybe into the invoke itself.

I'm wondering why our tests passed before. Could you look at them as well? Maybe they are not really testing what they were supposed to test!

ChrisMaddock:
@rprouse @CharliePoole Good memory! I came across those tests on my way, the did confuse me a little. As I saw it, the tests in #457 correctly test the Runnable status set on a testcase, so knock out tests where the parameter combination is incorrect. However, when tests with valid parameter lists come to be ran, calling the test method throws an exception.

The part I'm less sure about is if/how it used to work. There doesn't look to be any end-to-end tests in place, which would fail, before this PR. Has the underlying method of invoking test methods changed at any point? As far as I can find out, `MethodBase.Invoke` always requires default parameters to be specified as `Type.Missing`.

Rob - thanks for the comments, and pointing out the PR name. I'll add the tests now.

CharliePoole:
Ah! I see that now. The purpose of NUnitTestCaseBuilder is to build the tests correctly, so that's all its tests need to check. It appears we never actually ran those tests anywhere until now.

CharliePoole:
A further comment about future work in TestCaseSource. It is indeed possible to share some code between the two attributes, and we can look at doing that. However, we want to bear in mind that attributes are intended to pretty much stand on their own. Users can provide their own attributes and so long as the right interfaces are implemented, it's supposed to work. Of course, for our part, we can use inheritance to share implementation, but it's something to keep in mind that our attribute code also serves as a model for users. Anyway, it is a future issue.

CharliePoole:
@rprouse I took this out of the milestone since it duplicates issue #1278

ChrisMaddock:
Have added the additional test cases Rob suggested. Also saw in #457 that apprantly `ParameterInfo.IsOptional` isn't available in the CF build, so have wrapped everything in `#if !NETCF` - is that the correct approach? I'm not sure how it would be possible to implement this without `ParameterInfo.IsOptional`.

@CharliePoole - our last posts crossed and I missed yours - although I think my ramblings about matched up to your points! Interested in your thoughts on how best to share the special-case functionality between the two attributes - I'd like to come back to `TestCaseSource` in the future and 'finish the job'. I was considering extracting `GetParametersForTestCase(IMethodInfo method)` to a separate static utility class, but it sounds like that's not the preferred way. Will create a separate issue for this shortly.

CharliePoole:
It's OK that `ParameterInfo.IsOptional` isn't available in CF 3.5. AFAIK optional parameters are not available in VS2008, which we use to build it. :-)

TestCaseSOurce: cross that bridge when we come to it. :-)

Issue #1004, opened 2015-11-07
Poor formatting of results for Assert.AreEqual(DateTimeOffset, DateTimeOffset)


JohanLarsson:
This test fails as expected:

    [Test]
    public void DateTimeOffsetsNotEqual()
    {
        var expected = DateTimeOffset.MinValue + TimeSpan.FromTicks(1);
        Assert.AreEqual(expected, DateTimeOffset.MinValue);
    }

The output is a little confusing:

    Expected: 01/01/0001 00:00:00 +00:00
    But was:  01/01/0001 00:00:00 +00:00

I don't have time to write a PR for this right now, hence this issue.

CharliePoole:
Thanks, we like issues. :-)

I'm not sure how we should display the values in this case, but it's worth thinking about.

JohanLarsson:
Simplest is perhaps:
```
time.ToString("O")
```
It formats according to [ISO8601](https://en.wikipedia.org/wiki/ISO_8601)
Sample:

    Expected: 0001-01-01T00:00:00.0000001+00:00
    But was:  0001-01-01T00:00:00.0000000+00:00

appel1:
Perhaps there should be some more generic object formatting system in place with default implementations for common types and that make it possible to override existing or provide new ones for types under test where ToString isn't enough.

oznetmaster:
Even with the addition of the new formatting capabilities, I still think that DateTimeOffset should have its own default formatter which is different then the DateTime formatter.

DateTimeOffset is a basic type in .NET since 3.5 (3.0?) and is actually recommended to be used instead of DateTime, so it would seem prudent to support it.

I have had this in my CF build for a long time (I fully support DateTimeOffset in CF), exactly for the reasons given in this issue.  I also use the "O" format as the default.

oznetmaster:
Interesting issue this brings up.  DateTime uses a custom format ("yyyy-MM-dd HH:mm:ss.fff") with the InvariantCulture rather then a standard one (like "O").

Which approach is better?  Does it matter?

CharliePoole:
@oznetmaster Agreed. This is an accepted bug.

We generally have used InvariantCulture for things like this.

CharliePoole:
It's an especially easy fix now that we are using a chain of formatters.

rprouse:
[...]

PR #1285, opened 2016-02-24
Nunit issue 1004

oznetmaster:
Add different format for DateTimeOffset in messages to show offset.


CharliePoole:
So is DTO not available in CF then?

CharliePoole:
Your branch build failed due to some mixup in the build script. However, the pr build succeeded and the code looks good, so I'm merging it.

CharliePoole:
PRs need a comment that  refers to the issue. The naming convention for branches works only for us humans. Using "Fixes" causes automatic closure: like so...

Fixes #1004

oznetmaster:
DTO is not available in normal CF.  I support it in my CF builds, however, and have a library with it in it.

oznetmaster:
I did not know about the Fixes thing.. sorry.

PR #1292, opened 2016-02-26
Add new interfaces. Expose IAvailableRuntimes as service.

### LINKED ISSUE #1283: Engine should expose available frameworks.
The GUI for one needs this in order to display in a menu. This will be an added interface in the API so it's an opportunity to test adding new interfaces in a backward-compatible manner.

It's only high priority because of the GUI.

CharliePoole:
This fixes #1283, providing an IAvailableRuntimes service as part of the API.

The Engine API has been changed by the addition of two new interfaces and an enum. Existing interfaces have not been changed.

### DIFF
--- a/src/NUnitEngine/nunit.engine.api/RuntimeType.cs
+++ b/src/NUnitEngine/nunit.engine.api/RuntimeType.cs
@@ -1,0 +1,24 @@
+﻿namespace NUnit.Engine
+{
+    /// <summary>
+    /// Enumeration identifying a common language
+    /// runtime implementation.
+    /// </summary>
+    public enum RuntimeType
+    {
+        /// <summary>Any supported runtime framework</summary>
+        Any,
+        /// <summary>Microsoft .NET Framework</summary>
+        Net,
+        /// <summary>Microsoft .NET Compact Framework</summary>
+        NetCF,
+        /// <summary>Microsoft Shared Source CLI</summary>
+        SSCLI,
+        /// <summary>Mono</summary>
+        Mono,
+        /// <summary>Silverlight</summary>
+        Silverlight,
+        /// <summary>MonoTouch</summary>
+        MonoTouch
+    }
+}
--- a/src/NUnitEngine/nunit.engine.api/IAvailableRuntimes.cs
+++ b/src/NUnitEngine/nunit.engine.api/IAvailableRuntimes.cs
@@ -1,0 +1,38 @@
+﻿// ***********************************************************************
+// Copyright (c) 2016 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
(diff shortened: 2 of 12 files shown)


CharliePoole:
Added unrelated cleanup of NUnitConfiguration class, removing all the stuff from V2 we no longer use and fixing a problem that comes up when the engine is stored in a directory with name 'lib', as happens when we install via NuGet. In that case, it expected to find nunit-agent in the parent directory of 'lib', which is not true any longer.

src/NUnitEngine/nunit.engine.api/RuntimeType.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// <summary>Microsoft .NET Compact Framework</summary>
+        NetCF,
+        /// <summary>Microsoft Shared Source CLI</summary>
+        SSCLI,
+        /// <summary>Mono</summary>
+        Mono,
+        /// <summary>Silverlight</summary>
+        Silverlight,
+        /// <summary>MonoTouch</summary>
+        MonoTouch
```
rprouse: Exposing this bothers me. I don't think we set it and we don't/can't detect it, so should we be exposing it? It also begs the question as to if we are exposing this, SL and CF, why are we not exposing all of the other platforms that we support like varios WinPhone versions, Win8.1, .NET Core, Android, Xamarin.iOS, etc? My assumption would be that this list is limited to runtimes that the engine supports, not runtimes that the framework supports.

CharliePoole: There are fields here that we don't use, of course, but they have been around  for a while. I think we should eventually merge the other platforms into this list, but the engine doesn't do anything with them right now.

In order to expose IAvailableRuntimes, I needed an interface for a runtime framework. I created IRuntimeFramework for that purpose. If memory serves, I asked you offline if you thought we should use the existing enumeration or a text field for this purpose and you suggested using the enum. Did we miscommunicate?

Basically, __something__ is needed in IRuntimeFramework to represent the particular type of runtime we are dealing with. It doesn't have to be the existing internal enum. It could be a new enum or it can be a string. Microsoft uses a string to represent the runtime in it's TargetFrameworkAttribute and we already use strings in lots of places in order to give us flexibility (at the cost of type-safety, of course).

I'm open to replacing this and even to changing the internal enum as appropriate, but we need something to distinguish the runtime in question beyond the simple display name. Eventually, this value will trigger the engine to do different things to run the tests depending on the runtime.

rprouse: I don't think we miscommunicated, I just forgot what a mess this enum was. I don't think that SSCLI or MonoTouch are ever set or used. If so, can we at least remove them? We can then start adding in new values as the engine supports them?

CharliePoole: Let me check to verify we don't use any of these - I don't believe we do.

CharliePoole: OK... The thing is that the RuntimeType enum is in both the engine and the framework, although it is not in our common code. All the values are actually used in the framework when processing PlatformAttribute. Of course, this means they are __not__ used in the portable build of the framework.

In the engine, we use Any, Net, Mono in a major way - those being the platforms on which the engine runs. NetCF and Silverlight are used by RuntimeFramework internally, but the engine never creates a RuntimeFramework with those values. SSCLI and MonoTouch are not used at all.

Since this has to be in both the engine and the framework  - and possibly in some of your runners - I think a magic string is the best choice, like "MONO" or "NET". That's how we handle the names of package settings already. If a new platform comes along, we can create a name for it and it can be added separately into each independent component that needs it.

For now, I would simply expose the string name in the interface and keep the internal enum. After the release, we may want to clean up internally as well.

What do you think?

rprouse: I would be ok with just deleting SSCLI and MonoTouch from the enum, or switching to a string. Personally, I like type safety and discoverability, so would lean towards the enum.

CharliePoole:
I did some thinking about what the RuntimeType would be used for in the client. The engine uses it, of course, because each runtime type requires different actions in order to load the tests. I couldn't really think of a use for it in the client so I dropped it from the interface.

I added Id to the interface. This is the thing you pass in the package in order to select a particular runtime, so it's important. Without it, we could use ToString() on the RuntimeFramework object, but that's an implementation detail the client doesn't need to know about.

I think this is ready to merge now, once the CI tests pass.

PR #1296, opened 2016-02-28
Make Arguments a separate array from OriginalArguments

### LINKED ISSUE #1282: TestCase using Params Behaves Oddly
I have a test method that takes an integer array. I use `params` to make writing test cases easier:
```csharp
[Test]
[TestCase()]
[TestCase(1)] // <-- this test case misbehaves
[TestCase(1, 2, 3)]
public void TestMethod(params int[] array) {
    Assert.That(array.Length, Is.Not.EqualTo(1)); // Fail arrays of length 1 (see below)
}
```
However, the middle test case above behaves oddly. To make the test case stand more out here, I've made the test fail only for that test case.

When running the tests inside VS using NUnit Adapter 3.0.8.0 the test shows up as follows:

As oppose to the others that show up as `TestMethod()`and `TestMethod(1, 2, 3)`, the middle test case is given as an array: `TestMethod(System.Int32[])`. I would expect it to show `TestMethod(1)`.

If I use ReSharper to run the tests (which I normally do), the test case shows up twice:

One time as a failed test `TestMethod(System.Int32[])`, and one time as a pending test `TestMethod(1)`. It seems to write the right thing for the pending test, but not the failed one.

CharliePoole:
This fixes #1282. It turns out that OriginalArguments and Arguments were both references to the same array, so that when the content of the Arguments array was modified for certain cases, the OriginalArguments were also modified, defeating the entire purpose of having two properties.

### DIFF
--- a/src/NUnitFramework/framework/Internal/TestParameters.cs
+++ b/src/NUnitFramework/framework/Internal/TestParameters.cs
@@ -50,8 +50,8 @@ namespace NUnit.Framework.Internal
         public TestParameters(object[] args)
         {
             RunState = RunState.Runnable;
+            InitializeAguments(args);
             Properties = new PropertyBag();
-            OriginalArguments = Arguments = args;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


CharliePoole:
@nunit/core-team This is a very simple fix and so a good opportunity for one of the new committers to participate in a review. We need it fast, however, since we're close to a release. :-)

rprouse:
Looks good to me. I will leave it open for a bit to give other @nunit/core-team members a chance to do a second review. If you need this soon @CharliePoole, feel free to merge.

rprouse:
Updated the branch. I will merge when CI passes.

PR #1304, opened 2016-03-01
Changes for CF build

dicko2:
I've got the build working on appveyor now at least, which is a good start as this will pick up the basic mistakes like the one i made with my last PR :)

But the tests I have left not running for now as I need some help with them, and I don't know how you guys package it either.

I am assuming to run them its a little different, as the build output from the tests project is a exe file. Currently the script errors with it cant find nunit.framework.test.dll when i remove the exception for "runtime != CF3.5".

Can you assist with this one maybe @CharliePoole or @oznetmaster ?


oznetmaster:
I cannot be any help here :(  I only build using VS.  Never used any of the other tools.

CharliePoole:
It's looking good to me but I'm about to go to bed! I'll review it tomorrow morning.

For running the tests, you should be able to just run the exe the same way we run the NUnit Lite tests. Of course it won't actually run under CF but it beats not testing.

More in the AM.

oznetmaster:
Is there any way that the CE emulator could be part of the CI process?  Then the build could actually be tested under CE.

dicko2:
I'm sure that possible @oznetmaster . How do you normally run the unit tests on CE, is there an agent that runs on the emulator that nunit talks to over the network? or do you load the files in via VS and execute on the emulator?

Which emulator do you use? is in the ones form the windows mobile SDK? is so which version? can you point me in the direction of the MS download site so I make sure I'm looking at the right ones?

Sorry for all the questions.. we never used nunit on our CF projects :(

dicko2:
ok, modified the cake script to run it on windows desktop but have failing unit tests now

dicko2:
Looking at one of the failures i can see this.

```csharp
#if NETCF
            Assert.That(result.Properties.Get(PropertyNames.SkipReason), Does.StartWith("File or assembly name").And.Contains(BAD_FILE));
#else
            Assert.That(result.Properties.Get(PropertyNames.SkipReason), Does.StartWith("Could not load").And.Contains(BAD_FILE));
#endif
```

I think we might have some issues running the tests on windows as the result from the failed test running the above code is

> 1) Failed : NUnit.Framework.Api.TestAssemblyRunnerTests.Load_BadFile_ReturnsNonRunnableSuite
>   Expected: String starting with "File or assembly name" and containing "mock-assembly.pdb"
>   But was:  "Could not load file or assembly 'mock-assembly.pdb' or one of its dependencies. The system cannot find the file specified."

I think i should be able to get the emulator running on the build agent. 

I'll have a crack

CharliePoole:
My initial reaction would not be to change the plan (i.e. use an emulator) but reduce scope. If we can't get the tests running on desktop .NET via unification, we can merge in the part that does the compile, giving us 90% of the value we are looking for. We often have committed changes in the past that won't build under CF and you have just solved that problem! 

That said, I think there is a good chance that we can run the tests as planned. As you see, they are running. We have some errors, which can be fixed one at a time. For example, many of the tests for the content of an exception message are brittle, because they depend on a particular platform. I recently relaxed some of them so they would pass on Linux. They may need further relaxation to work on CF, but it will take looking at the actual code.

@rprouse I see two ways to go here: (1) Keep working on this without merging until after 3.2 or (2) Clean up things and merge with CF compile only. What's your preference?

build.cake
```diff
@@ -735,8 +775,9 @@ Task("TestAll")
     .IsDependentOn("TestDriver")
 	.IsDependentOn("TestAddins")
 	.IsDependentOn("TestV2Driver")
-	.IsDependentOn("TestConsole");
-
+	.IsDependentOn("TestConsole")
+    .IsDependentOn("TestCompactFramework");
```
CharliePoole: Putting this in the TestAll target and not the Test target means that when we specify -t Test -Framework net-cf-3.5, the tests won't be run. However, the point is moot if you drop the TestCompactFramework task.

rprouse:
I think I would prefer to see this merged post 3.2. It wasn't really on our scope for 3.2 and I think getting an emulator working in AppVeyor will be non-trivial.

I really don't want to mess too deeply with the build process this close to release :smile:

CharliePoole:
Fair enough! I'm going to re-create this in a branch under the nunit org so multiple folks can more easily experiment with it. Frankly, I think an emulator is overkill given the level of community usage of the CF build - although it __is__ an interesting challenge! If we end up creating CF as a separate project, maintained by someone who has roots in CF, then it might make more sense to do this.

CharliePoole:
I created PR #1305, containing the same changes, to replace this one.

PR #1305, opened 2016-03-02
CF Build for Appveyor CI

### LINKED ISSUE #332: Add CF to the Appveyor CI build
As indicated by #331 we keep breaking the compact framework build without realizing it. We need to at least compile it in our CI and if possible run the tests. Without installing a bunch of stuff in Appveyor, we'll probably only be able to run it using assembly unification under desktop .NET for now.

CharliePoole:
**Do Not Merge Yet - for discussion**

I created this from PR #1304 incorporating @dicko2 's changes. It will be easier for us all to work with under the NUnit organization. See earlier comments on #1304.

This will fix #332 when merged. We won't merge it until after the 3.2 release, however.


dicko2:
The way I've used test frameworks on mobile platforms before (teleriks app platform) they use a remote agent installed on the device or emulator  that communicates over the network back to the test agent. Did I read that nunit can do something like this? It might be a better option as

1. Windows mobile 6.x sdk won't run on 2012 r2
2. Windows mobile 7.x sdk won't install side bide side with net 4
3. Later win ce sdks (7 and 2013) are 8 gb downloads 

The way we've done builds for cf in the past is a custom rolled agent running Windows 7 and vs 2008 with all the right sdks loaded. If appveyor had support for a custom agent image or on premise agent we could achieve this. But I think loading the sdks on the agent is out or our reach. 

How about the packaging though?

dicko2:
I thought what i would do is commit the working build but fix up the tests in another Issue/PR.

In the spirit of a "Potentially Shippable Increment" this seems the right way to approach it.

This means that at least it'll be easier to pick up the build breaking code now (as i notice an issue with the other day, not me this time :smile: ), then we can look at the CF framework unit tests on another issue/branch/PR, if we relax them to run on windows, or figure out how to run them on an emulator can be discussed in that Issue.

So I've done the fix ups requested by @CharliePoole in the other PR and just commented out the tests for CF and made a note that it will be done later in the comments.

dicko2:
Have raised issue #1317 for the compact framework unit tests.

build.cake
```diff
@@ -612,6 +620,16 @@ void BuildFramework(string configuration, string framework)
...
+        case "netcf-3.5":
+			BuildProjectCF("src/NUnitFramework/framework/nunit.framework-netcf-3.5.csproj", configuration);
+            BuildProjectCF("src/NUnitFramework/testdata/nunit.testdata-netcf-3.5.csproj", configuration);
+            BuildProjectCF("src/NUnitFramework/tests/nunit.framework.tests-netcf-3.5.csproj", configuration);
+            BuildProjectCF("src/NUnitFramework/mock-assembly/mock-assembly-netcf-3.5.csproj", configuration);
+            BuildProjectCF("src/NUnitFramework/slow-tests/slow-nunit-tests-netcf-3.5.csproj", configuration);
+            BuildProjectCF("src/NUnitFramework/nunitlite.tests/nunitlite.tests-netcf-3.5.csproj", configuration);
+            BuildProjectCF("src/NUnitFramework/nunitlite/nunitlite-netcf-3.5.csproj", configuration);
+            BuildProjectCF("src/NUnitFramework/nunitlite-runner/nunitlite-runner-netcf-3.5.csproj", configuration);
+            break;
```
CharliePoole: You changed the compilation order from the other framework builds. I don't see how this can work. For example, the framework tests depend upon mock-assembly. I suppose that it's possible that there is no dependency in CF due to use of some #if statements, but it's confusing anyway to see the order changed here.

appveyor-tool.ps1
```diff
@@ -0,0 +1,119 @@
...
+        New-ItemProperty -Path $registryPath -Name $name -Value $value `
+            -PropertyType String -Force | Out-Null
+        }
+        If(!((Get-ItemProperty -Path $registryPath -Name $Name).ProductDir -eq "C:\Program Files (x86)\Microsoft Visual Studio 9.0"))
+        {
+            throw "Registry path " + $registryPath + " not set to correct value, please check logs"
+        }
+        else
+        {
+            Progress("Regsitrty update ok to value " + (Get-ItemProperty -Path $registryPath -Name $Name).ProductDir)
```
CharliePoole: Spelling. :-)

oznetmaster:
CF tests do depend on mock-assembly, so order is important.

appveyor-tool.ps1
dicko2: yes!... i haven't been able to find a spell checker for VS Code that will check the spelling in my comments and strings, the only one available only appears to do markdown files

CharliePoole:
This looks ready to merge as far as I'm concerned. I'll update the branch from master and we can merge it after @rprouse does the 3.2 release.

PR #1306, opened 2016-03-02
Do not call ToString() for passing tests

### LINKED ISSUE #1301: Assert.AreNotSame evaluates ToString unnecessarily
We have an object where `ToString()` cannot return a meaningful result at runtime. We implemented this method to throw an exception in debug builds so our unit tests catch cases where the object is passed to methods like `string.Format` that implicitly call `ToString()`.

When switching from NUnit 2 to NUnit 3, many of our tests started to fail. This is caused by the fact that in NUnit 3 `Assert.AreNotSame` and `Assert.AreSame` call `ToString()` on the object *even when the assertion succeeds*. In addition to causing unnecessary performance overhead during testing, this is notably problematic for our code. Here is a stack trace showing the manner in which this occurs:

image

rprouse:
Fixes #1301 

I owed @sharwell one for his help on my GitHub extension. Unless there is a problem with the fix, this will be in the 3.2 release, hopefully within the week.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/ConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintResult.cs
@@ -55,6 +55,8 @@ namespace NUnit.Framework.Constraints
     /// </summary>
     public class ConstraintResult
     {
+        IConstraint _constraint;
+
         #region Constructors
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


sharwell:
:tada: :birthday:

PR #1307, opened 2016-03-02
workers=0 conflicts with Apartment attribute

### LINKED ISSUE #1066: ApartmentAttribute and TestCaseAttribute(s) do not work together
I have tried the following:

and I get:
```csharp
System.InvalidOperationException : The calling thread must be STA, because many UI components require this.
at System.Windows.Input.InputManager..ctor()
```

I've tried to add the Apartment attribute to TestIgnore and NumericTextBoxTest as well. No changes.
Using nunit3-console.exe

```csharp
    public class NumericTextBoxTest
    {
        [TestFixture]
        public class TestIgnore
        {
            [Apartment(ApartmentState.STA)]
            [TestCase(0, Key.OemMinus, true, true, "5", false)] // does not allow negative value if text is not empty
            [TestCase(0, Key.OemMinus, true, true, "", false)] //Allows minus value if enabled
            [TestCase(0, Key.OemMinus, true, false, "", false)] //Allows minus value if decimal values not allowed
            [TestCase(0, Key.NumPad5, true, true, "", false)] //Allows numpad 5 value
            [TestCase(0, Key.D6, true, true, "", false)]//Allows d6 value
            [TestCase(1, Key.OemComma, true, true, "8", false)] //Allows decimal value if enabled and if there is a valid value
            [TestCase(1, Key.OemComma, false, true, "0", false)] //Allows decimal value if negative not allowed
            [TestCase(4, Key.OemComma, true, true, "-555", false)] //Allows decimal value if valid negative value inserted
            [TestCase(0, Key.Tab, true, true, "555", false)] //Allows decimal value if valid negative value inserted
            [TestCase(0, Key.OemComma, true, true, "", true)] //Allows decimal value without zero
            [TestCase(0, Key.OemMinus, true, true, ",5", true)] //Allows negative decimal value without zero
            public void NumericTextBoxInputTestNotIgnored(int caretIndex, Key key, bool allowNegativeValues, bool allowDecimalValues, string text, bool allowNoZeroDecimals)
            {
                    var numericTextBox = new NumericTextBox { Text = text, CaretIndex = caretIndex };

                    var isIgnored = numericTextBox.Ignore(key, allowNegativeValues, allowDecimalValues, false, false, allowNoZeroDecimals);
                    // Assert
                    Assert.IsFalse(isIgnored);
            }
        }
    }
```

rprouse:
Fixes #1066 

The original issue was nearly fixed by PR #1202, this completes the fix for Apartment on TestFixtures.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
@@ -239,7 +239,7 @@ namespace NUnit.Framework.Internal.Execution
 #else
             currentApartment = Thread.CurrentThread.GetApartmentState();
 
-            if (Test is TestMethod)
+            if (Test is TestMethod || (currentApartment != TargetApartment && TargetApartment != ApartmentState.Unknown))
                 RunTestOnOwnThread(timeout, TargetApartment);
             else
                 RunTest();
@@ -333,7 +333,6 @@ namespace NUnit.Framework.Internal.Execution
                     WorkItemComplete();
                 }
             }
-
         }
 #endif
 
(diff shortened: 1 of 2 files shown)


CharliePoole:
I updated the branch from master. We should merge when all passes.

oznetmaster:
Looks good to me.

Issue #981, opened 2015-11-03
Define NUnit Versioning for post-3.0 Development


CharliePoole:
We can't push any code for post-3.0 changes without first deciding on future assembly versioning, at least if we think we may want to change away from use of 3.x.*.* That makes this an immediate priority, since some of us are working on 3.0 stuff already!

Our package versioning is only slightly less critical. We can push but we can't create any CI, nightly or weekly packages until we decide about it. So I'm dealing with both types here

ASSEMBLY VERSIONS
We currently define assembly versions as 3.0.*.* for most assemblies. The sole exception is nunit.engine.api, which uses 3.0.0.0 for all releases. We can change this only for 3.1 and subsequent versions, because changing it for 3.0 would mess up the ordering of versions.

QUESTION: Do we want to start controlling the third digit of assembly version with 3.1 as has been suggested by @oznetmaster ?

PACKAGE VERSIONS:
We have so far put out packages numbered 3.0.0, with various suffixes. We will eventually put out our final release packages as "3.0.0"

QUESTION: What do we want to call bug fix releases of 3.0?

QUESTION: What will we call packages that we generate in CI? We know they will start with 3.1, but what comes after that? 

QUESTION: Do we want any other development packages like nightly, weekly? What will we call them? It has to be consistent with the CI packages but also clearly distinguished, perhaps by a suffix.

QUESTION: What about the 3.2 release? We won't get to it for a few months but how do we want to version it. Is what we are doing for 3.0 OK or will it be something else? Obviously, this question is not critical immediately, but we may want to know where we are going in advance.

CharliePoole:
I'm posting the text of an email I sent to the developer list on the same subject. It contains my own thoughts on the various questions to be resolved.

rprouse:
I like what is implemented in #932, probably with some tweaks. I think that we should go with, `{major}.{minor}.{build}-{suffix}` where;

**Major** - Is currently 3 and will not change unless there are breaking changes to the API. Hopefully not for a long time :smile: 

**Minor** - Increments for feature releases. Hopefully we will be creating these releases often once 3.0 is released. I am fine if we use the even/odd semantic version to indicate development releases, but I think that a suffix conveys that more clearly.

**Build** - Increments constantly by AppVeyor. When released without a suffix and without incrementing Minor, is a bugfix only release (hotfix?). If we are doing regular feature releases, I would hope that these are rare and only to fix serious bugs.

[...]

PR #1311, opened 2016-03-03
Lock assembly version to "3.0.0.0" at least for nunit.framework.dll

halex2005:
There is best practice to lock major version number and leave remaining numbers as zeroes in AssemblyVersion, and use AssemblyFileVersion to communicate actual version to users.
Autogenerated AssemblyVersion breaks things like bindingRedirect on any update of nunit now.

Suppose you are building a framework assembly for your project which is used by lot of developers while building the application assemblies. If you release new version of assembly very frequently, say once every day, and if assemblies are strong named, Developers will have to change the reference every time you release new assembly. This can be quite cumbersome and may lead to wrong references also. A better option in such closed group and volatile scenarios would be to fix he 'Assembly Version' and change only the 'Assembly File Version'. Use the assembly file version number to communicate the latest release of assembly. In this case, developers will not have to change the references and they can simply overwrite the assembly in reference path. In central/final release builds it makes more sense to change the 'Assembly Version' and most keep the 'Assembly File Version' same as assembly version.

It's easy to autogenerate versions because nunit already uses cake for build. In this PR versions are generated like C# compiler does (3rd number is days count from 1 january 2000, 4th number is seconds count from midnight divided by two).

I don't know if NunitApiEngineVersion.cs needs version "3.0.1.0" specified, but I leaved this behavior in this PR.

What do you think about this PR?

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -551,13 +551,27 @@ Task("PackageCF")
 //////////////////////////////////////////////////////////////////////
 Setup(() =>
 {
-    // Executed BEFORE the first task.
-	});
+	// Executed BEFORE the first task.
+	var datetimeNow = DateTime.Now;
+	var daysPart = (datetimeNow - new DateTime(2000, 1, 1)).Days;
+	var secondsPart = (long)datetimeNow.TimeOfDay.TotalSeconds/2;
+	var assemblyInfo = new AssemblyInfoSettings
+	{
+		Version = "3.0.0.0",
+		FileVersion = string.Format("3.0.{0}.{1}", daysPart, secondsPart)
+	};
+	CreateAssemblyInfo("src/NUnitConsole/ConsoleVersion.cs", assemblyInfo);
+	CreateAssemblyInfo("src/NUnitEngine/EngineVersion.cs", assemblyInfo);
+	CreateAssemblyInfo("src/NUnitFramework/FrameworkVersion.cs", assemblyInfo);
+
+	assemblyInfo.FileVersion = "3.0.1";
+	CreateAssemblyInfo("src/NUnitEngine/EngineApiVersion.cs", assemblyInfo);
+});
 
 Teardown(() =>
 {
-    // Executed AFTER the last task.
-    CheckForError(ref ErrorDetail);
+	// Executed AFTER the last task.
+	CheckForError(ref ErrorDetail);
 });
 //////////////////////////////////////////////////////////////////////
 // HELPER METHODS


CharliePoole:
We're in the process of making changes to the assembly and file versions for the 3.2 release, which will happen this weekend. The approach we have taken is to eliminate the use of '*' in the versions. The 3.2 release will use assembly and file versions 3.2.0.0 (except for nunit.engine.api, which remains at 3.0.0.0 for the assembly version.

By using zeroes for the build and revision, we will be able to modify it after the release while still having the numbers increase correctly. We'll look at your proposal in more detail after the release as well.

For now, here are some comments...

1. What you describe is __a__ best practice but not the only one. I'd like to hear more about why it applies to NUnit. In particular, your comment about releasing daily definitely does not apply, since we are basically releasing on a quarterly schedule.

2. Our current plan (see issue #981) is to use the first three components in a meaningful way, so for example 3.2.1 would be a bug-fix release of 3.2. We plan to use that value in both the assembly and file versions. It's hard to see why that would cause a problem for folks installing the stable builds, but I suppose something else may be appropriate for the development (3.3) builds. Can you comment on that?

3. The File version of all our assemblies (including the api assembly) is  generally expected to track the package, so whatever version we are packaging that will go into the file version.

It may be that I'm not really understanding the problems you are running into, so feel free to explain further if necessary. :-)

halex2005:
1) In PR I just implement same version strategy as C# compiler does with "3.0.*" string. As you already use cake, so you could implement any other version increment algorithm without restrictions.

2) The problem with nunit I running into is configuration like (example for nunit v3.0.1):

```xml
<configuration>
  <runtime>
    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
      <dependentAssembly>
        <assemblyIdentity name="library-name" publicKeyToken="hex-key" culture="neutral"/>
        <bindingRedirect oldVersion="0.0.0.0-3.0.5813.39032" newVersion="3.0.5813.39032"/>
      </dependentAssembly>
    </assemblyBinding>
  </runtime>
</configuration>
```

When I'll update nunit to next version, I'll need to change all references to new version. Ok, Visual Studio and nuget can sometimes make it for me automatically, but there is other troubles:

- I need to do additional commits to my repository with new version number in *.csproj and *.config files
- I can't simply copy new version of nunit.framework.dll to folder with tests and run them without <bindingRedirect /> or recompiling.

3) File version could be any string. It does not participate in resolving assembly dependencies and therefore can be changed with any build. Assembly version should be static string (should not be autogenerated on daily basis).

I like versioning approach of [Json.NET](https://www.nuget.org/packages/Newtonsoft.Json/) - assembly version is static (like "8.0.0.0") and it changes only when major version changes. File version uses three digits for specific current version and 4th number is autogenerated (like "8.0.2.17603").

CharliePoole:
1) Yes, I saw that. It's a good idea and I think we will do something with it after the 3.2 release this weekend. It's simply too late to make such a big versioning change  right now.

2) I don't understand why you use a binding redirect. If you have compiled your tests against a particular version of the framework, why not continue to use that version? Why drop in a new framework, with a different version and  features unless you intend to recompile and use those features.

3) Yes, the File Version is there to give the users information. Immediately, it is the same as assembly version, but that's not the plan for the future. 

When you talk about changing daily, I think you mean the version "3.0.*.*" which we no longer use. There are no stars in our versions in the current master.

The versioning approach you describe (Json.Net) is exactly what we have decided to do in the issue I pointed to above. The date information is redundant for stable builds but will have some relevance for unstable builds. Which are you using?

halex2005:
Actually, I needed to add assembly bindings to nunit3-console.exe.config file :)
Last week I tried to update our nunit2-based tests to nunit3 (because of we need RetryAttribute for our selenium-based tests). There is no nunit2 assemblies at our project for now, but when I run nunit3-console, I got error "cannot load nunit.dll v2.6.4".
Can nunit2-driver be cause for this error?

I think, I'll wait for release v3.2.
Thanks

CharliePoole:
There is no assembly in NUnit v2 or v3 called `nunit.dll`. If that's actually the name that is being sought, it's caused by some other program. If it's `nunit.framework.dll` 2.6.4, that means that you have some test that's referencing the V2 framework.

halex2005:
Yes, you right, of course it was message for `nunit.framework.dll`.
Actually, we don't use nuget packages for now. We have `externals` folder, and there is only one `nunit.framework.dll` exist (V3). So, I don't know where reference to V2 framework can be.
I'll research this issue a bit later.

Anyway, in v3.2 you have fixed AssemblyVersion, thanks. 
I don't know if you would like to have auto-incremented file version and what algorithm you'll use. Here in PR is one but not the only option. Version policy is up to you :)

CharliePoole:
We left it as 3.2.0 for this release to allow flexibility in selecting an algorithm before the next one: 3.2.1, 3.2.1234, etc.

rprouse:
@halex2005 thanks for your input, but I am going to close this PR. We are open to adjusting the way we version assemblies and have made several changes and may make more going forward. If you would like to see further changes, please enter an issue and we can discuss/design it there.

PR #1312, opened 2016-03-04
Fixes to nuget packaging of the Engine

### LINKED ISSUE #1297: NUnit.Engine nuget package improvements
The nuget package for the engine currently puts the engine into the lib directory of the package. This causes all the engine assemblies to be referenced by the project to which it is added and all of the assemblies are copied to the output directory by default. This copying is the correct behavior for any application that uses the engine and wants to keep it in it's own directory, The references may or may not be needed:
 * nunit.engine.api - should always be referenced
 * nunit.engine - some apps may reference it directly if they intend to create it using new rather than through TestEngineActivator
 * nunit-agent and nunit-agent-x86 should not be referenced at all, but they must me in the same directory as the engine in order to be used.
 * Mono.Cecil assembly should be in the same directory as the engine but would not be referenced unless the application had a specific use for it.

Currently, it's necessary to manually remove the unwanted references. At a minimum we should automate this in an install.ps1 script.

However, other apps may not want to keep the engine local but may want to use a pre-existing engine installation. For those apps, we need an NUnit.Engine.Api package, which only installs the nunit.engine.api assembly and allows it to find the engine.

Finally, when developing runners such as the NUnit gui, it may be desirable to have a special version of the engine that is used (because new features are added in parallel) while developing but still rely on a separately installed engine at run time. This may lead to an engine package that keeps the engine in the tools directory, so that nothing is referenced or copied.

@rprouse I'm laying this out so we can decide how much to do for 3.2 and how much to save for the future. What are your thoughts?

CharliePoole:
This fixes #1297 

* TestEngineActivator now looks for an engine package under the solution packages directory if one is not found in the AppBase. This feature is considered experimental.

* Tracing is activated in the agent (needed while debugging this)

* Nuget package files are now grouped in subdirectories.

* There is a separate experimental engine package that installs the engine under tools rather than copying it to the appbase.

The experimental features are currently being used by the gui but have some problems.


rprouse:
Looks good. About time we got the logging in the agent working. I will take your word on the fact that the engine needs to be in the lib and tools folder. I see what you mean now from our discussion today.

PR #1313, opened 2016-03-04
Backout one change from PR #1117

### LINKED ISSUE #1122: Region may be disposed before test delegate is executed
See @oznetmaster's comments on https://github.com/nunit/nunit/commit/394243999c7a096f2cc99523b9b1435234ddd872

>This appears invalid, since region is could/will be disposed before the delegate is executed?

>This was not an issue when region was only referenced within the using statement.

And

>When compiled for the CF build, this results in the compile error:

>Error 2 The type arguments for method 'NUnit.Framework.Constraints.Constraint.ApplyTo(TActual)' cannot be inferred from the usage. Try specifying the type arguments explicitly.

rprouse:
Fixes #1122 

I have proven to myself that this is not a problem. The override of `GetTestObject` on the ThrowsConstraint that could cause the problem is never called in async operations, so the base version that just returns the passed in delegate is always used, so we always WaitForPendingOperationsToComplete. Because of that, I just removed the GetTestObject from that call since it is never needed in our codebase.

I added an async throws unit test to satisfy myself that this is the case.

We could just close #1122, but it is probably safer to merge. @oznetmaster can you take a look since you spotted the error in the first place?


oznetmaster:
Seems good to me.

PR #1315, opened 2016-03-04
Fix build warnings

rprouse:
Have I mentioned that I hate build warnings before? :smiling_imp: 

Mainly unused variables for test classes and missing XML Doc comments. I also added a couple of copyrights that I noticed missing and added AttributeUsage to the new OrderAttribute to ensure it is only added to test methods.


CharliePoole:
I keep meaning to do this! Looks good once the CI builds pass.

What's the deal with removing `private` in a few places - does that give a warning?

rprouse:
Removing private is not a build warning, but it is a recommendation/warning in the editor in VS 2015 now because it is redundant. Just my OCD forcing me to tidy wherever I am :smile: 

That said, most of our code uses private, so I should probably leave it alone or we can make a decision to change going forward.

CharliePoole:
It's a question of style __except__ in tests where we are actually trying to test the ability to access some kind of private data.

rprouse:
>It's a question of style except in tests where we are actually trying to test the ability to access some kind of private data.

Curious, why? Just because it makes it explicit that it is private? The lack of modifier on class members defaults to private, so it is functionally the same.

CharliePoole:
For clarity in the test - although maybe the definition of what is clear is a matter of style too. :-)

OsirisTerje:
I don't disagree with removing the private modifier, but just curious about which warning @rprouse refers to.  
VS2015 supports many different analyzers, and they need to be set to what the project feel is right.

oznetmaster:
I actually do disagree with removing the private modifier.  I am not sure what warning is being presented by what editor, but I do not get one.  Maybe I have it disabled.

I have always been told to explicitly put the private modifier on because it clarifies the scope of the variable.

To me, it is like only using var to define variables if the type of the variable is explicit on the RHS of the equals sign.  For instance, I would never use var i = 1;  It is a totally legal statement, and the in the statement int i = 1, the int is often marked as unnecessary by R#.

CharliePoole:
I prefer to see the modifier as well. The main trouble with most static code checkers is that they come bundled with a set of rules and teams have a tendency to just accept them.

OsirisTerje:
Since private is the default for members (fields and methods) in a class, I don't see it give any benefit, I normally prefer less clutter.  Just the same way I see C# 6 has been going, it is a lot about reducing unneeded clutter.
My question here is simply about the warning itself, I don't get it - and frankly - I WANT to have it :-)

About var,  I agree with primitive types should not be var's, but anything else I prefer var with RHS inferred types.  Again, less clutter, and also easier to refactor the code, changing from type to interfaces and so on.  

With VS 2015 I think all teams need to consider their rules.  If you just turn on e.g. the Sonarlint or CodeCracker analyzers on code and use the defaults, you can end up with 1000's of warnings - most being noise. In one case I had 20K warnings on a project, which is not usable.  So, all teams with VS 2015 need to really consider their rules. 

And, so should we too.

CharliePoole:
My last 10 years or so of work was as a coach. That work gave me a slightly different view of coding rules and standards than what I had before then. For me, coding standards are only partially about the code. They are about team cohesiveness. That means that many things can be left as a matter of style, so long as there is consistency within some given scope. So the primary "rule" I enforce is don't change things that are "just" style arbitrarily. Examples of arbitrary changes are going into a file and changing it when you have no other reason to be in that file or making global changes in a file where you had to fix a one-line bug.

That's for things that are "style preferences." Unfortunately, we don't make clear in our standards which things are rules and which things are preferences. I've been meaning to do that for a while and have never gotten around to it. I'm thinking of categories like Do, Consider, Avoid and Don't as used in some other kinds of standards.

I think this sort of liberal interpretation is important to us for a couple of reasons. On the "people" level, nobody has to work on this project. If somebody has a way they have learned to do something and if they prefer it, we might just put up with their style in order to get their contribution. Obviously, there are limits but it really isn't the same as working as an employee on a team. Similarly, different folks may work with different tools, so anything that is __only__ possible in VS2015 may not be so desirable. (Although I do agree that VS2015 changes things a lot, since most people working in Windows will end up using it.)

On the technical side, we have to remember that the NUnit code has to compile and run under Linux, using one of several available compilers. The framework is even more restricted, since the CF build has to compile under VS2008.

Back to the original point, I personally like to see `private` but that's because it's my habit. I wouldn't like a rule that says I can't use it but neither would I call for a rule that says not to use it. IMO it is one of those matters of style that should be consistent in  a file but no more. If the only reason for changing it is that some piece of software issues a warning, I'd simply turn that warning off.

For me, the question isn't "Should we always use private or not?" It's "Why do we need a rule about this particular item?"

appel1:
Because the default access modifier varies depending on context I prefer that it is declared explicitly since it makes it easier to read the code. Same reason for why I prefer to avoid var except for with long generic types, it makes the code harder to read since you have to read more of to find out what type a variable has.

PR #1325, opened 2016-03-07
Fixed the NUnit.Console NuGet package

### LINKED ISSUE #1323: NUnit.Console Package Changes
It appears that stuff in **NUnit.Console** has been moved from `tools` to `bin` (which breaks our build server). Furthermore it looks like the new **NUnit.ConsoleRunner** package might be a better replacement, as for one thing it still has got the "tools" in the `tools` folder and for another it's the smaller package (unfortunately we would still need to change our build server to use the other package).

Before we change our build server:
* Was the switch from `tools` to `bin` in **NUnit.Console** intentional or will that be changed back?
* Would it be recommended to use **NUnit.ConsoleRunner** or is that more or less only created for internal purposes? What constraints does it have?

Thanks.

rprouse:
Fixes #1323 

These changes are already packaged and pushed to NuGet.org. Because it needed to be versioned 3.2.0.1 but still depend on 3.2.0 packages, that was slightly jury-rigged.

### DIFF
--- a/nuget/runners/nunit.console-runner-with-extensions.nuspec
+++ b/nuget/runners/nunit.console-runner-with-extensions.nuspec
@@ -36,4 +36,5 @@
       </group>
     </dependencies>
   </metadata>
+  <files></files>
 </package>


CharliePoole:
Do we need to separately pull the 3.2 label and apply the change there or do we have to remember this for when we do 3.2.1?

rprouse:
Short story, already done.

Long story, I worked off the release branch, did the work there, tagged the update 3.2.0.1, then did a branch to bring those changes back into master with this PR. So far, it is looking like a good way of working with releases. Because I did the original merge back in with a branch that updated to 3.3, this PR only prompted the changes I made for 3.2.0.1.

image

CharliePoole:
Great. i was worried about how the slash in the release branch would impact
packaging, but I see you fixed it.

Issue #1329, opened 2016-03-09
Unable to build without Compact Framework


ChrisMaddock:
Have just updated my repository, and no longer appear to be able to run the build script without the compact framework installed. (As a result of #1305, I presume)

```
nunit\src\NUnitFramework\framework\nunit.framework-netcf-3.5.csproj(463,11): 
error MSB4019: The imported project "C:\Windows\Microsoft.NET\Framework\v3.5\Microsoft.CompactFramework.CSharp.targets" was not found. 
Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.
An error occured when executing task 'BuildAllFrameworks'.
```

I have no prior knowledge of CF, but it doesn't seem to be a simple matter of installing the relevant framework - is there, or should there be a built-in option to ignore the CF build? 

My only reason to build is to run the tests before submitting a PR, so being able to exclude CF allow me to check the majority, and then the CI would catch CF issues afterwards.

CharliePoole:
It could be #1305 or #1328, which was just merged.

Originally, @dicko2 had CF as a separate command, but I asked him to make it as part of the AllFrameworks target when on Windows. The build does work on AppVeyor, where we go ahead and install everything that's needed. We might not want to do that on user machines, however.

@rprouse What do you think is the right way for this to work?

CharliePoole:
I should have asked: How do you run the script, precisely?

CharliePoole:
OK... Here's the scoop... On AppVeyor, appveyor.yml runs a bootstrap program in appveyor-tool.ps1, which installs the necessary CF components.

I think this is a critical problem, since it impacts all our contributors, but we should pause a beat and figure out how we actually __want__ it to work before doing anything. Do we want to install CF components on the machine of every developer who runs the build script?

ChrisMaddock:
I normally run 'build -t TestAll'. (Although this fails on the build task i.e. just running 'build') I run on Windows.

I'd personally be in favour of all relevant components being installed by default - I imagine seeing different test results locally to those on a PR could be frustrating.

CharliePoole:
The alternative is detecting the presence of the CF components and only building if they are present. Come to think of it, we probably want to do that before installing anyway. At the moment, I'm not sure how to do that.

dicko2:
Could just check for the presence of the CF framework files in the cake script. Or only run the CF build if using the appveyor target might be better. just adding an if for now maybe.

[...]

PR #1332, opened 2016-03-10
detect if CF installed when not in appveyor

dicko2:
Have added code to check if CF installed.

Have also done some code sot hat if in appveyor and not CF installed it will crash to prevent this code from accidently stopping the CI build in the event that the CF install fails.


dicko2:
this is for issue #1329

CharliePoole:
@dicko2 Thanks for getting on this. Have you been able to test it on a system without CF installed? @ChrisMaddock You initiated the issue, can you pull down the new script and see if it works in your environment? (It definitely looks like it will)

If it works, I am inclined to put it in as at least a temporary fix. We were discussing on #1329 whether the right approach is to install CF or not run the tests. I'd prefer that @rprouse make the call.

dicko2:
Yep I uninstalled CF from my local and tested. But haven’t tried on a clean machine though.

build.cake
```diff
@@ -647,14 +648,23 @@ void BuildFramework(string configuration, string framework)
...
+                BuildProjectCF("src/NUnitFramework/slow-tests/slow-nunit-tests-netcf-3.5.csproj", configuration);
+                BuildProjectCF("src/NUnitFramework/nunitlite/nunitlite-netcf-3.5.csproj", configuration);
+                BuildProjectCF("src/NUnitFramework/nunitlite.tests/nunitlite.tests-netcf-3.5.csproj", configuration);
+                BuildProjectCF("src/NUnitFramework/nunitlite-runner/nunitlite-runner-netcf-3.5.csproj", configuration);
+            }
+            else
+            {
+                Information("Compact framework build skipped because files were not present.");
+                if(isAppveyor) 
+                    throw new Exception("Running Build on Appveyor, but CF not installed, please check that the appveyor-tools.ps1 script ran correctly.");
```
CharliePoole: This doesn't really seem to be needed. If the install fails we already get an error.

dicko2: Just being cautious as they are two separate scripts/processes, someone could remove or bugger up the first.

ChrisMaddock:
Great - thanks @dicko2 - will be able to pull and try it later on today.

ChrisMaddock:
This change sorts out `Build`, but doesn't exclude the CF tests, so still fails for `TestAll`. I'd probably expect them to also be excluded, if they project wasn't built? 

If failing the tests is seen as the way to go - it could use a custom error message. Below is what you currently see - so unless you catch the `TestCF` task go by, you've no idea on the source of the error.

```
----------------------------------------
Teardown
----------------------------------------
Executing custom teardown action...
An error occured in the custom teardown action.
Error: One or more unit test failed, breaking the build.
The directory name is invalid
```

Lastly, the "CF Build Skipped" message is (imo) lost in the sea of text - should it be `Warning()` rather than `Information()`? I guess this somewhat depends on whether we decide to auto-install the CF components or not - but if CI vs local are going to behave differently, I'd personally want to definitely know it's happening. =)

Thanks for the quick turnaround @dicko2!

CharliePoole:
@rprouse and I talked this over. We don't want to be installing CF on the machine of any contributor who runs the script. That would be too intrusive. So the way that @dicko2 is handling it is how we should go. So let's sort out the comments and we can merge this. :-)

CharliePoole:
@dicko2 Are you still working on this?

dicko2:
Been on holidays soz @CharliePoole , have some time now, at airport lounge, will action shortly :smile:

dicko2:
Have done the same for tests to skip if not installed, and also updated the level to warning. After thinking about the level of the message I think its important to highlight to people that aren't compiling and testing in CF that it will be at some point, so they are aware.

Having the warning level makes it stand out a bit more, but not too much.

image

CharliePoole:
No problem. Just wanted to be sure whose court the ball was in.

CharliePoole:
Let us know when it's ready to review again.

dicko2:
Good to check now @CharliePoole , i think i covered everything

CharliePoole:
Updated from master... waiting for results...

CharliePoole:
And again... :-)

CharliePoole:
Sorry... I see that the relevant change is in PR #1332 which has not yet been merged. It keeps falling behind master and requiring updates. I'll get it merged and then get back to you.

CharliePoole:
Third time merging in master! I'm going to merge this as soon as the tests pass to avoid a fourth.

PR #1339, opened 2016-03-11
Added missing TestAttribute

Degusto:
I found that one public method with assert is not marked as test so I added attribute.


rprouse:
Nice catch thanks. Very minor change and CI is passing, so I am going to merge without a second review.

Issue #1333, opened 2016-03-10
Single Thread per Worker


null511:
I have a framework in development based on _Selenium_ for automated web testing that successfully runs in parallel using NUnit 3.0.1 - However this seems to be broken since 3.2.0, which includes the useful ability to output during SetUp/TearDown events.

My current setup uses `Thread.CurrentThread` against a dictionary to limit one browser instance per Worker thread, as I was unable to find a way to determine which worker created the test/fixture that is running. This seemed to be working well in 3.0.1, but is since broken in 3.2.0.

I was wondering if you think this would be a useful feature for others as well, or if you had any thoughts on implementing an option to enable this. I will most likely pull down the current source and make the modifications for my own purposes, but I could submit a PR as well if there is interest.

oznetmaster:
In 3.2, there is now one thread per WorkItem, which in most cases corresponds to a Test method.  This is a different thread then the thread the TestFixture is running on, and only exists for the duration of the Test (including its SetUp and TearDown).

If you access your dictionary in SetUp, then it seems like it should work as desired.

CharliePoole:
I'm not clear what you mean by "limit one browser instance per worker thread." One at a time? One for the life of the thread? Why not just require your fixture to run on it's own thread using [RequiresThread] ? That thread will then be discarded and never used again.

null511:
Thank you for your prompt response.

@oznetmaster, I just tested this, and I'm seeing the same thread from SetUp/Test/TearDown, but a different thread per fixture, which breaks my usage. Tested by printing `Thread.CurrentThread.ManagedThreadId`

@CharliePoole, My goal is to match the _Browser_ (IWebDriver) count to the Worker count. Opening a browser instance requires a lot of overhead, so opening/closing per fixture is not an optimal solution. In 3.0.1, my tests were retrieving a browser instance from a dictionary by using `Thread.CurrentThread` as the key, creating an instance if it has not been attached.

```C#
ConcurrentDictionary<Thread, IWebDriver> instances;

[TestFixture]
void TestBase_SetUp() {
    this.Driver = instances.GetOrAdd(Thread.CurrentThread, () => new IWebDriver())
}
```

I think a more reliable solution would be to have a _WorkerContext_ available, or possibly an additional property on _TestExecutionContext_ to identify which worker the test is running under.

null511:
[...]

PR #1340, opened 2016-03-11
Added Test.WorkerId

null511:
- Implemented `WorkerId` property for identifying the worker that is
  running the current test. This is useful for projects which implement an
  instance-per-worker model.

### DIFF
--- a/src/NUnitFramework/framework/Interfaces/ITest.cs
+++ b/src/NUnitFramework/framework/Interfaces/ITest.cs
@@ -94,6 +94,12 @@ namespace NUnit.Framework.Interfaces
         ITest Parent { get; }
 
         /// <summary>
+        /// Gets the name of the TestWorker that
+        /// is executing this test.
+        /// </summary>
+        string WorkerId {get;}
+
+        /// <summary>
         /// Returns true if this is a test suite
         /// </summary>
         bool IsSuite { get; }
--- a/src/NUnitFramework/framework/Internal/Tests/Test.cs
+++ b/src/NUnitFramework/framework/Internal/Tests/Test.cs
@@ -51,6 +51,8 @@ namespace NUnit.Framework.Internal
         /// </summary>
         protected MethodInfo[] tearDownMethods;
 
+        private string _workerId;
+
         #endregion
 
         #region Construction
@@ -234,6 +236,20 @@ namespace NUnit.Framework.Internal
         /// Used by the core to set the parent.
         /// </summary>
         public ITest Parent { get; set; }
+
+        /// <summary>
+        /// Gets the name of the TestWorker that
+        /// is executing this test.
+        /// </summary>
+        public string WorkerId
+        {
+            get {
+                if (_workerId != null) return _workerId;
+                if (Parent != null) return Parent.WorkerId;
+                return null;
+            }
+            internal set { _workerId = value; }
+        }
 
         /// <summary>
         /// Gets this test's child tests
--- a/src/NUnitFramework/framework/TestContext.cs
+++ b/src/NUnitFramework/framework/TestContext.cs
@@ -326,6 +326,15 @@ namespace NUnit.Framework
             }
 
             /// <summary>
+            /// Gets the name of the TestWorker that
+            /// is executing this test.
+            /// </summary>
+            public string WorkerId
+            {
+                get { return _test.WorkerId; }
+            }
+
+            /// <summary>
             /// The properties of the test.
             /// </summary>
             public IPropertyBag Properties
--- a/src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
@@ -116,7 +116,10 @@ namespace NUnit.Framework.Internal.Execution
 
                     if (Busy != null)
                         Busy(this, EventArgs.Empty);
+
+                    _currentWorkItem.Test.WorkerId = Name;
                     _currentWorkItem.Execute();
+
                     if (Idle != null)
                         Idle(this, EventArgs.Empty);
 


CharliePoole:
We never resolved issue #1333 as to whether we actually wanted to accept, how we would do it or what it's priority might be. That said, sometimes it's easier to look at code, so thanks for providing this.

We can continue any required philosophical discussion here. Meanwhile I'll make line comments on the implementation itself. :-)

src/NUnitFramework/framework/Interfaces/ITest.cs
```diff
@@ -94,6 +94,12 @@ public interface ITest : IXmlNodeBuilder
         ITest Parent { get; }
 
         /// <summary>
+        /// Gets the name of the TestWorker that
+        /// is executing this test.
+        /// </summary>
+        string WorkerId {get;}
+
+        /// <summary>
```
CharliePoole: Starting at the top, this is the first thing I notice. I don't like adding info to a test about how it is being executed. That breaks the separation we tried to create so carefully by having a WorkItem, which knows about tests. Let's see where you go with this as I read further.

src/NUnitFramework/framework/Internal/Tests/Test.cs
```diff
@@ -236,6 +238,20 @@ public bool IsSuite
         public ITest Parent { get; set; }
 
         /// <summary>
+        /// Gets the name of the TestWorker that
+        /// is executing this test.
+        /// </summary>
+        public string WorkerId
+        {
+            get {
```
rprouse: We don't use C# 6.0 features. They do not build in all supported platforms, thus the Mono build failure.

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -326,6 +326,15 @@ public string ClassName
             }
 
             /// <summary>
+            /// Gets the name of the TestWorker that
+            /// is executing this test.
+            /// </summary>
+            public string WorkerId
+            {
+                get { return _test.WorkerId; }
```
CharliePoole: Obviously, you had to do the stuff with test, since you wanted to eventually write this line. But _test isn't really a test, it's an adapter class that holds only some items from the test on a read-only basis. If we want to expose details of how the test is being run to the user code somewhere else in the context.

null511:
Thank you for your time @CharliePoole, I appreciate you looking into this. I had to make a quick implementation to repair my project, and thought it might be easier to open a PR for discussion.

It is necessary for my project to know which "Worker" it is running under, so I can limit 1 "instance" per worker. I was previously using `Thread.CurrentThread` from Fixture-SetUp against a dictionary, but this has since broken with the new thread-per-test model in 3.2.

I believe the solution implemented here is a better approach though, as it is thread independent. However, it's probably very "dirty" since this is the first time I've looked into NUnit internals.

CharliePoole:
OK, I'm back. See my detailed comments.

Overall, there are two questions:

1. Do we want to do anything to let the user code know about how they are being run.
2. How to implement it.

On item 1, I was hesitant at first, but thinking about it I see no harm. All we would be giving the user is a string, representing some concept of "where am I running." I'd start getting nervous about tests knowing much more than that, but it's a reasonable compromise, especially in the use case you are working on. @rprouse Do you agree?

On item 2, as commented inline, I don't want NUnit test objects to know how and where they are running. We only just got rid of that in the code with NUnit 3.0 by creating WorkItems. Similarly, now that I think about it, it's better if WorkItems don't know where they are running, although I admit it would do little harm to add the field provided WorkItems didn't grow to use it in some way. But of course, they would, wouldn't they. That's how these things happen. :-)

It seems to me there is a much simpler way to handle this. The name of the agent could be a property of the TestExecutionContext, which is modelled for the user by the TestContext. That sounds much neater to me. If we later add some built-in exclusion principle, we let the WorkItem examine the TestExecutionContext to find out what agent is running it. For the moment, we wouldn't need that.

To do this, each worker would need to plug it's name into the TestExecutionContext at the point where it pulled a work item from the queue and before it executed it. The work item at that point already has its Context property set, so it would be an easy change. Please take a look at the code there and see what you think.

CharliePoole:
One further comment... this has to work under --workers=0, where there is no test worker. That probably requires a change to SimpleTestDispatccher but I haven't looked at it.

CharliePoole:
And by "work" I mean, not crash when somebody accesses the field in TestContext.

null511:
1. That is why I went with `TestWorker.Name`. It was an already available unique identifier, and providing access to the Queue itself seemed hazardous.

2. `TestExecutionContext` was my first thought, as it would be separate from the fixtures & tests, but I did not see a good route for implementation on first glance, so I took the latter. If this is something that could potentially be merged in at some point, I would be glad to spend more time implementing as suggested.

3. With `--workers=0`, I imagine the `WorkerId` would simply be _null_, since it's set by the _TestWorker_ , which would leave it to the user to check before using.

CharliePoole:
You might run into issues I haven't thought of, but I think a simple automatic instance property for the WorkerId would do the job. If we go with a default of null, then the constructors can ignore it. The corresponding property on TestContext would be readonly. The user would access it by `TestContext.CurrentContext.WorkerId`.

Note that the OneTimeSetUp method of a fixture might receive a different WorkerId from what the individual tests get. That's true with your imlementation as well. I assume that's not a problem for you.

null511:
> Note that the OneTimeSetUp method of a fixture might receive a different WorkerId from what the individual tests get. That's true with your imlementation as well. I assume that's not a problem for you.

Thank you for pointing that out, I did not realize this. I expected that with Parallelization set to Fixtures, the fixture would be run from one [/the same] worker. It should not be an issue though, my "driver" lookup is done from a property accessor on a per-test basis. I will just need to document for my users that _OneTimeSetUp/TearDown_ should not be used, which is unfortunate, but not critical.

CharliePoole:
In fact, for the time being, the OneTimeSetUp __will__ be on the same WorkerId, but that's only because we have not implemented parallelization of methods. Now that I think of it, if ParallelScope is set to the Fixture level, I can probably keep it working just as it now does, which would be more efficient anyway.

CharliePoole:
@null511 Are you still working on this?

null511:
Yes, though i haven't had a  chance yet to look into implementing it through ´TestExecutionContext´. I should have something together by tomorrow afternoon though.

null511:
Moved to #1351.

PR #1341, opened 2016-03-11
Update BUILDING.txt

### LINKED ISSUE #1338: BUILDING.txt is outdated
Following on from #1332, would someone with a more comprehensive knowledge mind updating BUILDING.txt?

From my point of view, I'm interested the definitive way to run all tests. I currently run `build -t TestAll`. From what I gather, if I run `appveyor-tool.ps1` - this will install the dependencies that allow me to also build/run CF tests? 

Would be good to also document other required dependencies - the first time I pulled, I manually installed the Silverlight SDK to keep VS happy - was this a necessary step, or does the build script take care of this?

CharliePoole:
This fixes #1338 

I have updated the file to reflect how the build script currently works, even though we are likely to have some changes pretty soon. We need to work to keep this up to date when the script is changed.

I removed the last section of the file regarding how to do a release. It seems to me the info on the wiki at https://github.com/nunit/dev/wiki/Packaging-NUnit-3.0 is far more complete and up to date. There's no reason to maintain duplicate info.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,4 +1,4 @@
-Building NUnit 3.0
+# Building NUnit 3.0
 
 NUnit 3.0 consists of three separate layers: the Framework, the Engine and the Console Runner. 
 The source code is kept in a single GitHub repository at http://github.com/nunit/nunit.git.
@@ -10,7 +10,7 @@ Developers should make sure not to introduce any other references.
 
 There are two ways to build NUnit: using the solution file in an IDE or through the build script.
 
-SOLUTION BUILD
+## Solution Build
 
 All three layers are built together using a single Visual Studio solution (nunit.sln on Windows 
 and nunit.linux.sln on Linux), which may be built with Visual Studio 2012+, SharpDevelop.
@@ -24,9 +24,9 @@ components are placed directly in the bin directory while framework components e
 subdirectories net-2.0, net4.0, sl-5.0, portable and netcf-3.5. Future platform
 builds will cause new subdirectories to be created.
 
-BUILD SCRIPT
+## Build Script
 
-We use Cake (http://cake-build.net) to build NUnit for distribution. The primary script that controls
+We use **Cake** (http://cake-build.net) to build NUnit for distribution. The primary script that controls
 building, running tests and packaging is build.cake. We modify build.cake when we need to add new 
 targets or change the way the build is done. Normally build.cake is not invoked directly but through
 a build.ps1 (on Windows) or build.sh (on Linux). These two scripts are provided by the Cake project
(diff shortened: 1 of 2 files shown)


rprouse:
The changes themselves look good, but do you think we should change this to a Markdown .md file? Most of the formatting is already Markdown'ish, we would just have to add in a bit for headings etc. It would make it much easier to read on GitHub while still being readable as text.

CharliePoole:
Good idea.

CharliePoole:
Updated to markdown.

BUILDING.md
```diff
@@ -0,0 +1,79 @@
...
+Key arguments to build.cmd / build:
+ * -Target, -t <task>                 The task to run - see below.
+ * -Configuration, -c [Release|Debug] The configuration to use (default is Release)
+ * -Experimental, -e                  Use the experimental build of Roslyn
+
+The build.cake script contains a large number of interdependent tasks. The most 
+important top-level tasks to use are listed here:
+ * Build               Builds everything. This is the default if no target is given.
+ * Rebuild             Cleans the output directory and builds everything
+ * Test                Runs all tests. Dependent on Build. (These tasks currently do the
```
ChrisMaddock: Is there an unfinished sentence here?Apologies if not - my retro phone is struggling with the diff!

ChrisMaddock:
Much appreciated - thank you!

BUILDING.md
```diff
@@ -0,0 +1,82 @@
...
+    you have to be very careful that the build is up to date before packaging.
+
+ 2. If the compact framework is not installed on a machine, building and testing tasks
+    for CF are skipped and a warning is issued.
+
+ 3. Currently, Silverlight must be installed or an error occurs when trying to build or
+    test it. We will be changing this soon to match the strategy for CF.
+
+ 4. The Test and TestAll targets currently due the same thing. Both are retained for backward
+    backward compatibility and the semantics of at least one of them may change in the future.
```
CharliePoole: Moved the unclear line to a note. Hope that's better.

ChrisMaddock: Have just seen this on a pc and it makes much more sense - the line breaks on mobile didn't match up to show the sentence split over two lines! Cheers!

BUILDING.md
```diff
@@ -0,0 +1,82 @@
...
+ * TestAllFrameworks   Runs all framework tests. Dependent on Build.
+ * Test45              Tests the 4.5 framework without building first.
+ * Test40              Tests the 4.0 framework without building first.
+ * Test20              Tests the 2.0 framework without building first.
+ * TestPortable        Tests the portable framework without building first.
+ * TestSL              Tests the Silverlight framework without building first.
+ * TestCF              Tests the compact framework without building first.
+ * TestEngine          Runs all engine tests. Dependent on Build.
+ * TestConsole         Runs the console tests. Dependent on Build.
+ * Package             Creates all packages without building first. See Note below.
```
rprouse: When viewed as an MD, the spacing here is lost. It doesn't look too bad, so three options,

- Leave it as it is for now
- Put the targets in bold so they stand out a bit better
- Turn this into a table which is a bit of a pain in markdown

If you want to leave it for now, go ahead and merge.

CharliePoole: I added a fence before and after so the spacing would be respected. That seems to be the least intrusive when viewed as a text file.

I think this exercise may demonstarate that markdown is not such a great idea for files that will almost always be used locally and viewed as text. However, I don't want to go back at the moment. We can revisit the format of such files at another time.

PR #1343, opened 2016-03-13
Apply Assembly attributes to context correctly

### LINKED ISSUE #1266: SetCultureAttribute does not work if set on assembly level
Hello,

I was trying recently to specify the invariant culture for all the tests run in my project.
I have found that in NUnit 3, the [SetCulture] attribute allows to control culture used in test(s) and it can be applied on method, class or assembly.

It looks however that if I specify it on assembly level, the culture is not applied to tests.

Below, there is a test reproducing this issue (I run it in VS2015 with R# 10.0.2 and NUnit 3.0.1):

```c#
using System.Globalization;
using NUnit.Framework;
[assembly: SetCulture("")]

namespace ClassLibrary1
{
    //[SetCulture("")]
    [TestFixture]
    public class SetCultureTests
    {
        [Test]
        public void It_should_set_proper_culture()
        {
            Assert.That(CultureInfo.CurrentCulture, Is.EqualTo(CultureInfo.InvariantCulture));
        }
    }
}
```

Thanks,
Wojtek

CharliePoole:
Fixes #1266 

The code in CommandBuilder that adds an ApplyChangesToContextCommand to the command chain was only looking at fixtures and methods, not assemblies.

I verified the fix by stepping through the code after temporarily adding various attributes to the assembly. We need tests for all our attributes at the assembly level, but that's a bit hard to do without proliferating test assemblies.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/CommandBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/CommandBuilder.cs
@@ -52,18 +52,26 @@ namespace NUnit.Framework.Internal.Execution
             TestCommand command = new OneTimeSetUpCommand(suite, setUpTearDown, actions);
 
             // Prefix with any IApplyToContext items from attributes
+            IList<IApplyToContext> changes = null;
+
             if (suite.TypeInfo != null)
+                changes = suite.TypeInfo.GetCustomAttributes<IApplyToContext>(true);
+            else if (suite.Method != null)
+                changes = suite.Method.GetCustomAttributes<IApplyToContext>(true);
+            else
             {
-                IApplyToContext[] changes = suite.TypeInfo.GetCustomAttributes<IApplyToContext>(true);
-                if (changes.Length > 0)
-                    command = new ApplyChangesToContextCommand(command, changes);
+                var testAssembly = suite as TestAssembly;
+                if (testAssembly != null)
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/framework/Internal/Execution/CommandBuilder.cs
```diff
@@ -52,19 +52,27 @@ public static TestCommand MakeOneTimeSetUpCommand(TestSuite suite, List<SetUpTea
...
-            if (suite.Method!=null)
-            {
-                IApplyToContext[] changes = suite.Method.GetCustomAttributes<IApplyToContext>(true);
-                if (changes.Length > 0)
-                    command = new ApplyChangesToContextCommand(command, changes);
+                var testAssembly = suite as TestAssembly;
+                if (testAssembly != null)
+#if PORTABLE
+                    // TODO: Fix this code! It's really quite ugly!
+                    changes = new List<IApplyToContext>(testAssembly.Assembly.GetAttributes<IApplyToContext>());
```
rprouse: Can't you just use the ToList() extension method to make it less ugly?

CharliePoole: This .NET 2.0. :-) Actually, the TODO was for my earlier code with a double cast.

PR #1344, opened 2016-03-13
Give error when trying to run using --x86 in process

### LINKED ISSUE #1161: NUnit3-Console should disallow the combination of --inprocess and --x86, giving an error message
I have ran into peculiar issue, that I feel is worth reporting.

Using NUnit.Console.3.0.1
Executing tests for example test library: Tests.DLL
With following arguments:
C:\Example\packages\NUnit.Console.3.0.1\tools\nunit3-console.exe C:\Example\Tests\bin\Release\QsehEiJfpnSP7yg5NCqm1zISM1sgEwqS.nunit --work=C:\Example --noresult --noheader --x86 --framework=net-4.0 --process=InProcess

QsehEiJfpnSP7yg5NCqm1zISM1sgEwqS.nunit is an autogenerated TeamCity .nunit project file, with contents:

```xml
<NUnitProject>
<Settings activeconfig="active" appbase="C:\Example"/>
<Config binpath="Tests\bin\Release" configfile="Tests\bin\Release\Tests.dll.config" name="active">
<assembly path="Tests\bin\Release\Tests.dll"/>
</Config>
</NUnitProject>
```

**Case 1:**

Tests.dll is compiled for AnyCPU and is a NET4.5 unit test library, 
Tests.dll references App.exe, which is compiled for x86 and is a NET4.5 WinForms application
App.exe references App.Core.dll, which is compiled for x86 and is a NET4.5 code library

Console runner exits with code -2 and error message:
*Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information*
There is no binding debug output.

App.Core.dll is now recompiled for AnyCPU, App.exe is still compiled for x86

Console runner works correctly and runs all tests.

**Case 2:**

Same as Case 1, except Tests.dll is compiled for x86

In this case, console runner fails with exception System.BadImageFormatException for Tests.dll, with binding error: Failed to complete setup of assembly (hr = 0x8007000b). Probing terminated.

Is this a bug or is console runner unable to load x86 compiled dlls?

CharliePoole:
This fixes #1161.

Two error cases are handled...

* Use of both --x86 and --inprocess is detected in processing the command-line and reported as an error.
* If the engine discovers that the test needs to be run as 32-bit and --inprocess was used it throws an exception.

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
@@ -77,6 +77,18 @@ namespace NUnit.Common
 
         #endregion
 
+        #region Overrides
+
+        protected override void CheckOptionCombinations()
+        {
+            base.CheckOptionCombinations();
+
+            if (RunAsX86 && ProcessModel == "InProcess")
+                ErrorMessages.Add("The --x86 and --inprocess options are incompatible.");
+        }
+
+        #endregion
+
         #region Configure Additional Options for Console
 
         protected override void ConfigureOptions()
--- a/src/Common/nunit/CommandLineOptions.cs
+++ b/src/Common/nunit/CommandLineOptions.cs
@@ -170,7 +170,7 @@ namespace NUnit.Common
         {
             if (!validated)
             {
-                // Additional Checks here
+                CheckOptionCombinations();
 
                 validated = true;
             }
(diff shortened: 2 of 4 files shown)


src/NUnitConsole/nunit3-console/ConsoleOptions.cs
```diff
@@ -77,6 +77,18 @@ public class ConsoleOptions : CommandLineOptions
...
         #endregion
 
+        #region Overrides
+
+        protected override void CheckOptionCombinations()
+        {
+            base.CheckOptionCombinations();
+
+            if (RunAsX86 && ProcessModel == "InProcess")
+                ErrorMessages.Add("The --x86 and --inprocess options are incompatible.");
```
rprouse: Are they incompatible if you are running on an x86 machine? I agree that there is no reason to specify the x86 flag in that case, but is it invalid?

CharliePoole: That's a point since we can only run at 32-bit. In that case, there is no point in specifying it, however, so I would tend not to worry about it.

PR #1351, opened 2016-03-14
TestExecutionContext.WorkerId

null511:
- Implemented read-only `.WorkerId` property on `TestExecutionContext` to
  enable applications to determine which worker spawned the test. Contains
  `TestWorker.Name` when parallel testing is enabled, otherwise `null`.


null511:
@CharliePoole Sorry for the additional PR, changed the branch name. Similar implementation as previous PR, but through _TestExecutionContext_. I've tested this against my framework implementing NUnit, and it seems to work as expected. When workers = 1 (or 0), the `.WorkerId` property is null; otherwise it contains the `TestWorker.Name`.

Since the property is only assigned immediately before execution, I had to have the get accessor bubble up through _priorContext_.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -360,6 +365,19 @@ internal IWorkItemDispatcher Dispatcher
...
+        /// For builds with out the parallel feature, it is null.
+        /// </summary>
+        public string WorkerId {
+            get {
+                if (_workerId != null) return _workerId;
+                if (_priorContext != null) return _priorContext.WorkerId;
+                return null;
+            }
+            internal set {_workerId = value;}
+        }
```
CharliePoole: This surprised me. I figured you would set the WorkerId in CompositeWorkItem.CreateChildWorkItems. This is good but to reduce confusion, please add a comment explaining that some items are executed directly by their parents under the same agent.

CharliePoole:
You haven't exposed this to the user in TextContext. While it's possible for users to reference the internal TestExecutionContext class, we really don't want them to do that! Otherwise, it looks good.

null511:
> You haven't exposed this to the user in TextContext. While it's possible for users to reference the internal TestExecutionContext class, we really don't want them to do that! Otherwise, it looks good.

Ah, sorry. I'm still familiarizing myself with the internals. That makes sense, I'll look into this.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
null511: I've got to admit, this took a lot more thought then expected to get it passed through at the correct time, but it appears to be working properly.

null511:
Unit test added. If your happy with the changes, everything seems to be working fine on my end.

CharliePoole:
Looks ready to merge. @rprouse Will you second that?

CharliePoole:
@null511 Can you pull the latest changes from master?

null511:
Merged. If you'd prefer a rebase, let me know and i can do so in the morning.

CharliePoole:
This is  fine. We just need a second committer to OK the merge.

rprouse:
Looks good. I will merge.

PR #1354, opened 2016-03-16
Fix handling of arrays of reference elements as method arguments

### LINKED ISSUE #1327: TestCaseSource in NUnit 3 converts an argument declared as String[] to String
A TestCaseSource worked fine with NUnit 2 defined as
`public static readonly string[][] WrongArgs =
        {
            null,
            new string[0],
            new[] {"1"},
        };`
in a test like this
`  [TestCaseSource (typeof (TestData), "WrongArgs")]
        public void Test (string[] args) {...}`

But with NUnit 3.0 and NUnit 3.2 (from NuGet, and executed via ReSharper 10.2), the 3rd case is passed to the test as a string (`"1"`) instead of using the string array (`new[] {"1"}`), causing the test not being executed because of an ArgumentException on invokation.
The previous cases are passing the test as expected.

Any ideas?

CharliePoole:
This fixes #1327 

In addition, I reorganized code in TestCaseSourceAttribute for clarity. I wrote it myself and I found that I was having trouble understanding it! So please review this part for clarity as well.

I fixed it so that it would keep working for all the kinds of data we have supported in the past:

1. TestCaseData
2. Object arrays
3. Arrays of other types, like int[], string[]
4. Single arguments to a method taking one parameter without an object[] wrapping the argument.

To clarify, we have always allowed the following options for a single test case row:

```C#
For a method taking three ints:
    new TestCaseData(1, 2, 3) // case 1
    new object[] { 1, 2, 3 } // case 2
    new int[] { 1, 2, 3 } // case 3
For a method taking 1 int:
    new TestCaseData(42) // case 1
    new object[] { 42 } // case 2
    new int[] { 42 } // case 3
    42 // case 4
```

That's what makes TestCaseSourceAttribute a bit tricky. In particular, a method taking a single array as it's only argument creates ambiguity and we have to do some work to resolve that ambiguity.

We could simplify all this if we are willing to break some user tests. I'd like the review to consider this as a possibility. We could either eliminate case 3 or case 4.

If we eliminate case 3, we would affect any tests taking multiple args of the same type IFF the user had taken advantage of the ability to write `new int[] { 1, 2, 3 }` rather than `new object[] { 1, 2, 3 }`. This seems reasonable to me, since arguments are always provided as an array of objects. However, we would want to check how it impacts the `new [] { 1, 2, 3 }` syntax, which is quite common.

If we eliminate case 4, a user with a method taking a single arg would always have to wrap it in an object array for each case. Alternatively, a ValueSource attribute could be placed on the parameter. My sense, however, is that this might be fairly inconvenient for users.

If we eliminate both cases, the code becomes extremely simple but we have the drawbacks of both cases as well.

Let's treat this as a design review (of the existing design) as well and decide if we want to change.


CharliePoole:
I'll merge in the changes to master once we decide on a direction.

CharliePoole:
@rprouse Any thoughts on this?

rprouse:
Personally, I would hate to drop support for case 3 or 4. Personally, I tend to use the simplified version `new [] { 1, 2, 3 }` if possible.

The code is complicated, but it is now well commented. Maybe you should add your scenarios/cases from this pull request as a comment to the code? I found the comments in the code easier to read and understand after seeing the examples in this PR.

I have merged in latest from master and CI is running. I think we can merge this. @CharliePoole what is your current thinking?

PR #1355, opened 2016-03-17
Refactor build script

CharliePoole:
Because the non-task helper methods have different access under the two compilers, it seems like a good idea to reduce their use. I moved code from multiple helper functions into the tasks themselves and created dependent tasks to perform each of our six framework builds, similarly to the way tests are now organized.


rprouse:
Looks good to me. The one AppVeyor build hasn't reported, but the other passed so I am going to merge.

PR #1357, opened 2016-03-17
Stop running tests on a separate thread when not needed

### LINKED ISSUE #1352: Tests losing data setup on thread
After upgrading to 3.2, tests that expected mocks setup on the thread (i.e. System.Web.HttpContext.Current = SomeMockContext) start failing. The OneTimeSetup is setting correctly, but it's gone in the test. I'm getting NullReferenceExceptions in tests that were expecting it. These tests run fine after downgrading back to 3.0.1

CharliePoole:
Fixes #1352.

I have modified WorkItem so that tests run on their own thread only if absolutely needed. See the issue for details of when that happens. I added a test to verify that the same thread is used for OneTimeSetUp, SetUp and the test method. It's redundantly included in several places so it will be run if the developer selects (for example) ApartmentTests and just runs that alone.

What's left to do is
1) Test this in @rolyv 's environment or with a more realistic test of some kind.
2) Decide if we need some special attributes to mark such fixtures and pick the names.
3) Add whatever attributes we decide, plus tests


src/NUnitFramework/tests/Attributes/RequiresThreadAttributeTests.cs
```diff
@@ -50,6 +42,13 @@ public void TestWithRequiresThreadRunsSetUpAndTestOnSameThread()
             Assert.That(Thread.CurrentThread, Is.EqualTo(SetupThread));
         }
 
+        [Test]
+        public void TestWithoutRequresTHreadRunsEverythingOnSameThread()
```
rprouse: Capitalization of THreadRuns

rprouse:
I think you need some tests that OneTimeSetup => SetUp => Test are on the same thread. Should we also test things like TestCase and/or TestCaseSource?

src/NUnitFramework/tests/Attributes/RequiresThreadAttributeTests.cs
CharliePoole: Removing the entire tests - it duplicates the one in the base class.

CharliePoole:
That's what the test you commented on was doing. I removed it because the equivalent test is in the base class. I was surprised that I only really added three tests, and those redundant anyway, but once we know that OneTimeSetUp, SetUp and the test run on the same thread, there isn't much else to do. We could test TearDown and OneTimeTeardown but we already test that they match the thread of their equivalent setups elsewhere.

TestCase and TestCaseSource don't have anything that influences the thread used by a test.

If we add some attributes, we will have to test them, but I can't think of anything else right now. I'm as surprised as you are. :-)

rprouse:
I was thinking TestCase and/or TestCaseSource just in case someone does something in the future, but I guess it is unlikely.

CharliePoole:
I think I see what you mean. We need the same testing we already have, but for a parameterized test case. AFAICS it shouldn't matter whether it's created by TestCase or TestCaseSource.

oznetmaster:
It appears that doing this will once again break Stop and Abort functionality, which was why I added the separate thread in the first place.

CharliePoole:
The tests all pass. Can you give me a test to add that will fail?

oznetmaster:
All the tests passed before I made the changes!

I have not found any unit tests that adequately test the Stop and Abort capabilities.  I will see what I can come up with.

CharliePoole:
That would be great! Note that I created the slow-tests assembly with lots of waiting for exercising this.

Failing a functional (probably not unit) test of some kind in our test suite, I can try to hook up something to our Gui that allows testing manually.

CharliePoole:
I created SingleThreadedAttribute as a proof of concept. Let's review and talk about it.

In the initial implementation, I am silently ignoring Timeout, RequiresThread and Apartment attributes, as shown in the tests. Probably, that's not what we really want to do. It would be easy to show these tests as either errors or as ignored. If we want to allow some sort of per-test override of the SingleThreadedAttribute, that's a bit more work and requires us to figure out how the user can specify that an override is intended.

CharliePoole:
This last commit is unrelated. I was working in the CF project noticed we had forgotten to add the OrderAttribute and its tests there.

CharliePoole:
Existing code only works because we have not yet implemented running test cases in parallel. I added a bit of code to ParallelWorkItemDispatcher to ensure that the same test worker is used for each test.

I also added a comment explaining what direction we are trying to go with this so that nobody will accidentally move us back to creating extra threads. (Should have had this comment in the first place!)

CharliePoole:
Discovered I hadn't added the tests to all our projects. They now pass in all builds. Also added a test that demonstrates you can run a fixtuer single-threaded in the STA and all tests are on the same STA thread.

@nunit/core-team Still waiting for comments on the new attribute as an approach.

CharliePoole:
@nunit/core-team Unless somebody really dislikes the new SingleThreaded attribute, I think this is ready to merge. Note that nobody has ever responded regarding what we should do when a user with a single threaded fixture puts RequiresThread, Timeout or Apartment on one of the test methods. So right now, they are just silently ignored. I'm willing to put it in that way and wait for another bug to be filed or to leave it as is.

rprouse:
@oznetmaster I would like to merge this. Do you have any further comments or suggestions. Personally, I think that this is more important for more users than the ability to cancel test runs.

oznetmaster:
Since being able to stop/abort test runs is more important to my user then this, I will not make this change in my CF release if it disables the ability to stop or abort tests, as it seemingly will.  Now that it is "stable", I will exercise it more fully and see if it does.  However, since it seems to revert to what was around before, I am assuming it will :(

It would be much better if the reasons that abort/stop did not work before I added the extra threads were understood.  @CharliePoole  had, at one point, indicated that he would investigate that further, but nothing ever came of it. It is still unclear why abort/stop never worked before I made it work by adding the dedicated threads.

The goals of the PR have always seemed reasonable, but the methodology seems patchwork, and the solution seems to have taken us back to where we were before I added the extra threads.  I had hoped a solution could have been found which did not seemingly re-break abort/stop functionality.

CharliePoole:
The problem is that NUnit's own users have no way to cancel a run. Therefore, the functionality never gets exercised, making this a low priority for fixing. The other issues that came up were real issues, which the original design of having one thread per worker were intended to deal with.

Further, the reasons for the push toward one thread per worker got lost over time and Neil was not aware of them. So it made sense for him to do what he did: go back to the separate thread model of NUnit V2 and even expand it to using separate threads most of the time. Frankly, I don't think I did a sufficient job of reviewing the change at the time, because the issues that have resurfaced are exactly those issues that caused us to start moving away from separate threads.

Cancellation becomes an issue for NUnit once we have a gui. I thought it might also affect some of the third-party runners, but we have heard no problems from them. Our gui currently only runs simple tests. It is just about ready for a usability review and then we will start adding functionality to it, including cancelling of tests. That will make successful cancellation a priority for us as well and I'm sure we will figure out how to do it.

I created issue #1363 to deal with the most serious challenge: making timeouts work without using a separate thread. I'm experimenting with the approach described there. SImple cancellation of a run does not appear to be a problem but the fact that timeouts initially look like a cancellation is. We do __not__ have a feature in NUnit of cancelling a single test and continuing, other than the timeout, and I don't see any demand for such a feature. The user has know way to know what test would be cancelled anyway and many tests may be running simultaneously.

CharliePoole:
BTW, this is another case where we need integration tests! It's practically impossible to test cancellation in a pure unit-testing context.

oznetmaster:
I have no interest in stopping or cancelling a single test, just the entire test run.

On a CF device, a test run can often last multiple hours.  Other have indicated that this is the case as well, even in a non-CF environment.

Hence the need to be able to stop/abort the run.

All of my CF environments allow the aborting and stopping of the current run (both the GUI and the console interface).  I was not even aware that the normal NUnit had no way to do so, since the stop/abort api has always been there.  I just made it work correctly.

I am sure there is some need for a single thread in some cases, although I am still not sure exactly why the other possible solutions that keep thread isolation are not workable, since I have never had the need for a single thread, nor have any of my users.  I have therefore never investigated what seem to be alternative approaches.

Keeping the stop/cancel functionality is of paramount importance to me.  I would have thought that it might have become part of this PR.  Implementing new functionality that knowingly breaks existing (and necessary) functionality seems the incorrect approach.  @CharliePoole  had previously indicated the need for working abort/stop in the GUI, and I also assumed that it would be added to the console interface as well after the comments be a number of normal desktop users about its desirability, and after it was actually working correctly.

CharliePoole:
@oznetmaster I didn't think you personally needed to cancel a single test - I was just enumerating the possible meanings of "cancel" and explaining which ones I think we need to support.

The need for a way to cancel a test run is clear. From the beginning, this has been accomplished for the console runner by killing the process. Since the console runner terminates at the end of a run anyway, anything fancier would not serve much purpose.

Our Gui has always had the ability to cancel the test run. So will our new gui. There is no reason to think that tests have to run on a separate thread in order to do this.

Tests have had to run on a separate thread in the past in order to make timeouts possible, since we abort the thread in order to create a timeout. Issue #1363 is about getting rid of that necessity.

This existing functionality you are referring to is something that you added a few months ago to solve a problem that said you were having. Nobody else has seen the problem and you are unable to write a test that demonstrates the problem. Further, it does not make sense that you need to run every test on a separate thread in order to cancel the run - as you have said yourself - and the fix is causing problems for other people. So, yes, in such a case, I think it's pretty normal to remove it.

I'd be glad to work on the cancellation issue immediately after this if you can demonstrate the problem. Are not the tests of cancellation we already have sufficient? Can you add something to those tests that will demonstrate the conditions under which it isn't working for you?

PR #1365, opened 2016-03-20
Adding a DeclaringClassName property to Test class. Required for nunit/nunit3-vs-adapter#133.

Green-Bug:
If a method is added as a test through inheritance the ClassName property of Test class will contain the name of the derived class. This makes it impossible in the VS adapter to figure out the source location for that method.
DeclaringClassName will be set to the class where the method was declared as that’s the class that can be used to find the source location when using DiaSession.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Tests/Test.cs
+++ b/src/NUnitFramework/framework/Internal/Tests/Test.cs
@@ -51,6 +51,16 @@ namespace NUnit.Framework.Internal
         /// </summary>
         protected MethodInfo[] tearDownMethods;
 
+        /// <summary>
+        /// Used to cache the declaring type for this MethodInfo
+        /// </summary>
+        protected ITypeInfo DeclaringTypeInfo;
+
+        /// <summary>
+        /// Method property backing field
+        /// </summary>
+        private IMethodInfo _method;
+
         #endregion
 
         #region Construction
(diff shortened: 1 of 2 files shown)


CharliePoole:
I can't think of any reason why we would want to have a second class name in the Test class. If the name is incorrect, it makes more sense to fix it rather than to leave the incorrect name and add a new property. Am I missing something?

Green-Bug:
It is a defensive change. I do not know who takes dependency on the ClassName property and Xml attribute. A method is included as a test because of the derived class but it may be defined in the base class. Having both gives someone more information to work with when parsing the Xml or when extending the adapter. I'm obviously not opposed to simply changing the behavior of ClassName property, it was in fact one of my attempts at the fix. I simply thought that having both gives more flexibility.

CharliePoole:
The change of adding className and methodName is a fairly recent one and was solely for the purpose of giving the adapter a way to locate the source code. We have never been happy about this intrusion of an implementation detail in the XML, and only did it as the lesser of evils. So I think it's better to avoid multiplying such attributes.

Green-Bug:
OK, I'll simply change the ClassName and rollback the other changes.

Thanks for taking a look.

CharliePoole:
Can you add a test that uses an inherited test and shows the correct name?

Green-Bug:
I've been looking for a place to do that, but as far as I can tell there is not test class for Test.cs. Do you have any suggestions? I'll keep looking and add a test for it.

CharliePoole:
Added an issue to correspond to this PR. Usual process is that somebody else confirms the issue first, although I suspect that will be pretty easy.

CharliePoole:
How come you closed this?

Green-Bug:
Git is still a big puzzle for me, sorry about it (I use something else in my day to day work).
I was trying to clean all my reverts so I did a rebase and forced push to have a cleaner history. This action seems to have closed the PR. As soon as I have my new change and test ready I'll reopen it with the new code. Again, sorry about all the noise.

CharliePoole:
A forced push to the same branch should normally keep the PR open in GitHub. if you are online to github, it's pretty easy to click the close button in error. I do it all the time because I actually want to close the window.

It is better to do a forced push to the same branch once a PR is open because that retains all the discussion in one place.

As I noted in #1366 classname has only one use so it's OK to change what is returned to be the class in which the method is actually declared. However, the TypeInfo on which it is based is used in various ways, so you will have to look for the references if you want to change that. I suspect you will not be able to do so. On platforms with DeclaredType you can use that of course, but that won't help where it isn't s upported.

Green-Bug:
It seems the PR has been closed automatically as I don't have any new commits in my fork. I also can't reopen it right now for the same reason.

That is the change I'm working one, to fix the ClassName property and tests.

As far as I can tell MemberInfo.DeclaringType is supported on all platforms, did I miss any of them?

CharliePoole:
Sorry... It's ReflectedType that's not supported everywhere.

rprouse:
@CharliePoole I've given this a quick review, but I am going to leave the decision on if/when to merge up to you since you have been working with this PR.

rprouse:
Thanks for your work on this. Sorry we left it so long to merge.

PR #1367, opened 2016-03-21
Improve memory usage

### LINKED ISSUE #1247: Potential memory issue
This is a follow-up to [this](https://groups.google.com/forum/#!topic/nunit-discuss/upGTPcmOCS4) issue from a few weeks ago. To summarize, on upgrading one of our test suites from NUnit 2.5.7 to NUnit 3, we found run-time tripled, due to vastly increased memory requirements and lots of paging. Initial signs pointed to a change in the persistence of TestFixture classes.

To test this out, I created a testsuite with 10 TestFixtures as below, solely allocating memory. Under NUnit 2.5.7, this runs with ~300MB of memory, in NUnit 3, requires ~2GB

    [TestFixture]
    public class Class1
    {
        private Byte[] _allocated;

        [Test]
        public void Allocation()
        {
            _allocated = new Byte[200000000];
            Assert.IsTrue(true);
        }
    }

Profiling this, there looked to be two retention paths to all TestFixtures which lived for the length of the program run - full tree is below. 
image

To confirm the paths, adding [this ugly-hack](https://github.com/ChrisMaddock/nunit/commit/f2e2f914c3df1181cab22a1916a8ba20ac7862cc) to remove the references in `WorkItemComplete()` brought NUnit 3's memory usage down to similar levels to 2.5.7. (Although I've no doubt that will break other functionality!)

I assume this isn't an intentional/necessary change in NUnit 3, based on @CharliePoole's earlier comments on the mailing list? I wasn't sure at what point TestExecutionContext should be cleared down (if at all), hence the proof-of-concept ugly hack rather than proper fix - apologies there.

ChrisMaddock:
First attempt to fix #1247.

This just clears out the two retention paths at the end of `WorkItem.WorkItemComplete()`.

I hadn't pushed this before as it's a little ugly, and I wasn't sure of the best approach architecturally, although it sounds like this might be as good as any. Open to ideas on how to do it more elegantly!

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/OneTimeTearDownCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/OneTimeTearDownCommand.cs
@@ -71,6 +71,10 @@ namespace NUnit.Framework.Internal.Commands
                 IDisposable disposable = context.TestObject as IDisposable;
                 if (disposable != null && Test is IDisposableFixture)
                     disposable.Dispose();
+
+                //Clear references to test objects to reduce memory usage
+                context.TestObject = null;
+                Test.Fixture = null;
             }
             catch (Exception ex)
             {
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -129,6 +129,7 @@ namespace NUnit.Framework.Internal
 #endif
 
             CurrentValueFormatter = (val) => MsgUtils.DefaultValueFormatter(val);
+            IsSingleThreaded = false;
         }
 
         /// <summary>
@@ -159,6 +160,7 @@ namespace NUnit.Framework.Internal
 
             Dispatcher = other.Dispatcher;
             ParallelScope = other.ParallelScope;
+            IsSingleThreaded = other.IsSingleThreaded;
         }
 
         #endregion
@@ -450,6 +452,11 @@ namespace NUnit.Framework.Internal
         /// The current head of the ValueFormatter chain, copied from MsgUtils.ValueFormatter
         /// </summary>
         public ValueFormatter CurrentValueFormatter { get; private set; }
+
+        /// <summary>
+        /// If true, all tests must run on the same thread. No new thread may be spawned.
+        /// </summary>
+        public bool IsSingleThreaded { get; set; }
 
         #endregion
 
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -516,14 +516,6 @@ namespace NUnit.Framework.Internal
             CurrentValueFormatter = formatterFactory(CurrentValueFormatter);
         }
 
-        /// <summary>
-        /// Clear reference to test object
-        /// </summary>
-        internal void ClearTestObject()
-        {
-            TestObject = null;
-        }
-
         #endregion
 
         #region InitializeLifetimeService
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -516,6 +516,14 @@ namespace NUnit.Framework.Internal
             CurrentValueFormatter = formatterFactory(CurrentValueFormatter);
         }
 
+        /// <summary>
+        /// Clear reference to test object
+        /// </summary>
+        internal void ClearTestObject()
+        {
+            TestObject = null;
+        }
+
         #endregion
 
         #region InitializeLifetimeService
--- a/src/NUnitFramework/framework/Internal/Tests/Test.cs
+++ b/src/NUnitFramework/framework/Internal/Tests/Test.cs
@@ -324,6 +324,14 @@ namespace NUnit.Framework.Internal
         }
 #endif
 
+        /// <summary>
+        /// Clear reference to TestFixture
+        /// </summary>
+        internal void ClearTestFixture()
+        {
+            Fixture = null;
+        }
+
         #endregion
 
         #region Protected Methods
--- a/src/NUnitFramework/framework/Internal/Tests/Test.cs
+++ b/src/NUnitFramework/framework/Internal/Tests/Test.cs
@@ -324,14 +324,6 @@ namespace NUnit.Framework.Internal
         }
 #endif
 
-        /// <summary>
-        /// Clear reference to TestFixture
-        /// </summary>
-        internal void ClearTestFixture()
-        {
-            Fixture = null;
-        }
-
         #endregion
 
         #region Protected Methods
--- a/build.cake
+++ b/build.cake
@@ -352,7 +352,7 @@ Task("TestDriver")
   .WithCriteria(IsRunningOnWindows)
   .Does(() =>
 	{
-		RunTest(NUNIT3_CONSOLE, BIN_DIR, PORTABLE_AGENT_TESTS, ref ErrorDetail);
+		RunTest(NUNIT3_CONSOLE, BIN_DIR, PORTABLE_AGENT_TESTS, "TestDriver", ref ErrorDetail);
 	});
 
 Task("TestAddins")
--- a/src/NUnitEngine/Addins/nunit.v2.driver/XmlExtensions.cs
+++ b/src/NUnitEngine/Addins/nunit.v2.driver/XmlExtensions.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Text;
 using System.Xml;
@@ -255,8 +256,31 @@ namespace NUnit.Engine.Drivers
         private static XmlNode AddElementWithCDataSection(this XmlNode node, string name, string data)
         {
             XmlNode childNode = node.AddElement(name);
-            childNode.AppendChild(node.OwnerDocument.CreateCDataSection(data));
+            foreach (var section in EscapeCDataString(data))
+                childNode.AppendChild(node.OwnerDocument.CreateCDataSection(section));
             return childNode;
+        }
+
+        private static List<string> EscapeCDataString(string data)
+        {
+            var sections = new List<string>();
+            int start = 0;
+            while (true)
+            {
+                int illegal = data.IndexOf("]]>", start, StringComparison.Ordinal);
+                if (illegal < 0)
+                    break;
+                sections.Add(data.Substring(start, illegal - start + 2));
+                start = illegal + 2;
+                if (start >= data.Length)
+                    return sections;
+            }
+
+            if (start > 0)
+                sections.Add(data.Substring(start));
+            else
+                sections.Add(data);
+            return sections;
         }
 
         #endregion
--- a/src/NUnitFramework/framework/Attributes/SingleThreadedAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/SingleThreadedAttribute.cs
@@ -1,0 +1,53 @@
+// ***********************************************************************
+// Copyright (c) 2016 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+
+namespace NUnit.Framework
+{
+    /// <summary>
+    /// SingleThreadedAttribute applies to a test fixture and indicates
+    /// that all the child tests must be run on the same thread as the
+    /// OneTimeSetUp and OneTimeTearDown. It sets a flag in the
+    /// TestExecutionContext and forces all tests to be run sequentially
+    /// on the current thread. Any ParallelScope setting is ignored.
+    /// </summary>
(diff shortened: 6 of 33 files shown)


rprouse:
This looks fine for the console runner. @CharliePoole do you think this will cause any problems for GUI runners?

rprouse:
@CharliePoole can you give this a review?

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -516,6 +516,14 @@ public void AddFormatter(ValueFormatterFactory formatterFactory)
...
         }
 
+        /// <summary>
+        /// Clear reference to test object
+        /// </summary>
+        internal void ClearTestObject()
+        {
+            TestObject = null;
+        }
+
```
CharliePoole: Why add a method, given that TestObject has a public setter, which we use elsewhere?

src/NUnitFramework/framework/Internal/Tests/Test.cs
```diff
@@ -324,6 +324,14 @@ public void ApplyAttributesToTest(ICustomAttributeProvider provider)
         }
 #endif
 
+        /// <summary>
+        /// Clear reference to TestFixture
+        /// </summary>
+        internal void ClearTestFixture()
+        {
+            Fixture = null;
+        }
```
CharliePoole: As with the TestObject, Fixture also has a public setter, used elsewhere. See general comment.

CharliePoole:
First off, I don't see any problems with this and I would say let's merge it if we needed something right away for an immediate release. We should improve it if we have time. I see three levels of improvement:

1. For the immediate fix, I don't understand the need for two new methods to clear the two objects. If we don't like either property having a public setter, we should remove it and use a method everywhere. But I see no need for that. If necessary, we can make the setter internal, but that's over-worrying the problem, at least in my mind.

2. (Second level) Really, it's best if things get cleared in a logical location that corresponds to where they get set. So, for example, we set the context TestObject in our SetUpCommand. If we want to clear it, we should do it in the corresponding TearDownCommand. Since there can be multiple pairs of setup and teardown commands, we might need to remember which one did the setup. Alternatively, we could move the whole creation/clearing operation to a different command in the command sructure. If @ChrisMaddock wants to go this way, I can help point to some better locations.

3. (Third level) Maybe Test doesn't need a Fixture property at all. We should analyze how we are using it. It may be that it's only there for use in our tests or it may even be an obsolete holdover from NUnit V2. If we don't have two entirely different chains of reference to the test fixture, the problem is an order of magnitude simpler.

So... I think it depends on how fast @rprouse wants to do a bug fix release and how much time @ChrisMaddock wants to spend on this. :-)

CharliePoole:
PS: No problem for the gui because the object is newly created for each test run. It's only our wrapper objects, derived from Test, that survive from run to run.

rprouse:
The only thing that might currently warrant a 3.2.1 release is the test per thread issue, but people seem to be working around that. The memory pressure has been an issue since 3.0 and only a few people have reported it. We see it at work, but it isn't pressing for us.

@ChrisMaddock, can you give us some before and after memory numbers on this? You reported the issue, so can you give us an indication how much this improves your performance? Is this a fix that would help with your move to 3.x?

I am happy to make a number of small improvements, or we could try to tackle this more broadly as @CharliePoole suggested.

CharliePoole:
On second thought, maybe my third point should be first. If we no longer need Test.Fixture, it will simplify the other approaches.

I'll take a look at this since I know the history. Be back in a bit.

ChrisMaddock:
Hi both,

@rprouse:
>Can you give us some before and after memory numbers on this? You reported the issue, so can you give us an indication how much this improves your performance? Is this a fix that would help with your move to 3.x.

Sure. We encountered this on first upgrading our test suites to nunit3. Memory numbers are difficult, as the effect was that it was maxing out the 4GB RAM on our CI machine and thrashing - leading to ~35 minute runtime. Post-fix, they run in ~6 mins, comfortably under 1.5GB of memory. This is a suite of ~2000 tests, we saw similar problems with another suite of 10k tests.

I make no argument that these suites are at all memory efficient, or a sensible way to test - but regardless, it meant v3 was essentially unusable for us. We're currently running off an internal fork with this patch in, if we hadn't been able to do that, we'd have had to revert to v2. I'm a little surprised it hasn't been raised more to be honest, surely there are other bad tests suites out there somewhere! =)

I agree it possibly doesn't warrant it's own release, but if there's a 3.2.1 coming out anyway, it would be great for us if this could go in!

@CharliePoole 
> [...] and how much time @ChrisMaddock wants to spend on this

Absolutely keen to structure this right - but wasn't sure how best to approach it myself. Submitting this as a PR seemed the best way to start the discussion. 

>I'll take a look at this since I know the history. Be back in a bit.

Will hold fire till then!

ChrisMaddock:
Additionally, @YannTrevin's [comment](https://github.com/nunit/nunit/issues/1247#issuecomment-198244519) that revived issue #1367 is probably worth consideration - are we aware of anything that would have worsened the situation between 3.0 and 3.2?

My profiling was done on v3.0.

CharliePoole:
@ChrisMaddock If you want to move ahead, I suggest next removing Fixture from ITest and all the derived classes as an experiment. It looks to me as if you will break a few of our tests of ActionAttributes and that's all. Of course, we also have to worry about an action attribute being able to get to the user test object, so we should then try to revise the tests to do it a different way. I think that an action attribute can always look at TestExecution.CurrentContext.TestObject if it needs that information.

CharliePoole:
@rprouse Of course, I am suggesting breaking some possible user functionality here, requiring them to reimplement it in a different way. However, the rules for using an action attribute have never really settled down since our changes in 3.0, so I'm hoping nobody will mind  very much. Not carrying a copy of the object around with the test keeps the test as an abstract representation, as intended, and leaves the entire job of dealing with instances of the user class, their creation, lifetime and disposal, to the TestExecutionContext. I think that's as it should be.

rprouse:
I am okay with the breaking changes in this case. @ChrisMaddock if you have time, go ahead and make the changes @CharliePoole suggests.

ChrisMaddock:
The new commit inlines the methods and moves then to OneTimeTearDown, to reflect them being set in OneTimeSetUp.

Having some problems removing TestSuite.Fixture, as the tests use the property to pass in an existing TestObject, but will keep looking at it.

ChrisMaddock:
@rprouse @CharliePoole 

Pushing this with broken tests for now, wonder if either of you may be able to give some advice as to why? The broken tests are those that build/run through DefaultTestAssemblyBuilder - although right now, I'm finding it hard to track down why. They seem to be failing in the TearDown which is odd, the breaking change is the one I made to OneTimeSetUpCommand (below.) Any ideas what could be going astray here?

Before: 
```
             if (_typeInfo != null)
              {
                  // Use pre-constructed fixture if available, otherwise construct it
                 if (!_typeInfo.IsStaticClass)
                 {
                     context.TestObject = _suite.Fixture ?? _typeInfo.Construct(_arguments);
                     if (_suite.Fixture == null)
                     {
                         _suite.Fixture = context.TestObject;
                     }
                     Test.Fixture = _suite.Fixture;
                 }
  
                  for (int i = _setUpTearDown.Count; i > 0; )
                      _setUpTearDown[--i].RunSetUp(context);
```
After:
```
             if (_typeInfo != null)
              {
                  // Use pre-constructed fixture if available, otherwise construct it
                 if (!_typeInfo.IsStaticClass && context.TestObject == null)
                     context.TestObject = _typeInfo.Construct(_arguments);
  
                  for (int i = _setUpTearDown.Count; i > 0; )
                      _setUpTearDown[--i].RunSetUp(context);
```

src/NUnitFramework/framework/Internal/Commands/OneTimeSetUpCommand.cs
```diff
@@ -65,15 +65,8 @@ public override TestResult Execute(TestExecutionContext context)
...
-                {
-                    context.TestObject = _suite.Fixture ?? _typeInfo.Construct(_arguments);
-                    if (_suite.Fixture == null)
-                    {
-                        _suite.Fixture = context.TestObject;
-                    }
-                    Test.Fixture = _suite.Fixture;
-                }
+                if (!_typeInfo.IsStaticClass && context.TestObject == null)
+                    context.TestObject = _typeInfo.Construct(_arguments);
```
CharliePoole: Your test for context.TestObject being null is not present in the original code. Did you do this because it's being set by some of the test code?

src/NUnitFramework/framework/Internal/Commands/OneTimeTearDownCommand.cs
```diff
@@ -71,6 +71,9 @@ public override TestResult Execute(TestExecutionContext context)
                 IDisposable disposable = context.TestObject as IDisposable;
                 if (disposable != null && Test is IDisposableFixture)
                     disposable.Dispose();
+
+                //Clear references to test objects to reduce memory usage
+                context.TestObject = null;
```
CharliePoole: I'm suspicious. If this is necessary, then that must mean that the context isn't being properly disposed.

src/NUnitFramework/nunitlite.tests/TestUtilities/TestBuilder.cs
```diff
@@ -48,9 +48,7 @@ public static TestSuite MakeFixture(Type type)
 
         public static TestSuite MakeFixture(object fixture)
         {
-            TestSuite suite = MakeFixture(fixture.GetType());
-            suite.Fixture = fixture;
-            return suite;
+            return MakeFixture(fixture.GetType());
```
CharliePoole: We'll need to look at every use of this method. When a test pre-creates a fixture, it's generally because we have "primed" the instance with certain values for the test. That fixture must be used for the test to succeed. Most likely, we will have to stuff it into the context, if we don't already. (Hart to tell from the narrow point of view of the diff)

This also applies to the following method.

src/NUnitFramework/tests/TestUtilities/TestBuilder.cs
```diff
@@ -48,9 +48,7 @@ public static TestSuite MakeFixture(Type type)
 
         public static TestSuite MakeFixture(object fixture)
         {
-            TestSuite suite = MakeFixture(fixture.GetType());
-            suite.Fixture = fixture;
-            return suite;
+            return MakeFixture(fixture.GetType());
```
CharliePoole: Same issue as for TestBuilder in nunitlite.

CharliePoole:
@ChrisMaddock I'll pull it down and step through. For future, you might consider that it's slightly easier for us to do that if you use an issue-xxx branch in the NUnit repo rather than your own fork. We all have local repos that are synched to the nunit repo so all we have to do is checkout the branch. Dont' worry for now - it's not that much harder to clone yours. :-)

src/NUnitFramework/nunitlite.tests/TestUtilities/TestBuilder.cs
ChrisMaddock: It *looks* like this is happening within `TestBuilder.RunTestSuite`, although I'm potentially misreading what's going on.

src/NUnitFramework/framework/Internal/Commands/OneTimeTearDownCommand.cs
ChrisMaddock: Aha! Do you mean the context is meant to go out of scope? When we first put this patch in, it was necessary to clear the reference on both the fixture and the context. (As mentioned above - this was back on 3.0 though, I'll should check things with the master - that this patch is still working, for one.)

src/NUnitFramework/framework/Internal/Commands/OneTimeSetUpCommand.cs
ChrisMaddock: Yes, I believe so. Sorry - could have used a comment! I wasn't much of a fan of doing that however, it seems like we could use a proper mechanism for 'seeding' testobjects? 

Just to clarify - where are testobjects intended to be created - I presume this should happen in the OneTimeSetUp for the testfixture, and then be passed context to context for each tests?

ChrisMaddock:
Sure, will do in future!

src/NUnitFramework/framework/Internal/Commands/OneTimeTearDownCommand.cs
CharliePoole: Yes, the context definitely goes out of scope once the test (case or suite) for which it was created completes. My operating theory was that the existence of a double chain of refs (context and test) one formed top down and the other bottom up was creating circularity that kept everything alive. So If we can do without the fixture in the test, I wonder if everything will start being released as it goes out of scope.

src/NUnitFramework/framework/Internal/Commands/OneTimeSetUpCommand.cs
CharliePoole: Normally, they are created in one time setup. If they implement IDisposable, they are disposed in OneTimeTearDown, so that's where they should go out of scope else we get an error for referencing a disposed object. They should be copied down to child contexts where appropriate. One issue I can imagine is that they should NOT be copied down from SetUpFixtures. Hmmm... have to check how we do that.

The test utilities have to create some tests and run them in a non-normal manner. They sometimes pre-create a fixture and store it in the Fixture property of the test. But they also store it in the context, which ought to be enough. And we can always change our test utilities if necessary.

src/NUnitFramework/nunitlite.tests/TestUtilities/TestBuilder.cs
CharliePoole: So the question is whether all the tests that need to pre-create a fixture eventually go through RunTestSuite.

src/NUnitFramework/framework/Internal/Commands/OneTimeSetUpCommand.cs
CharliePoole: Just started investigating this. If I remove `&& context.TestObject == null` the currently failing tests all pass. Unfortunately, over 50 tests that are passing start to fail. I'll keep delving.

CharliePoole:
@ChrisMaddock I've been playing with this and my conclusion is we should just go for the simple fix right now. You could do a hard reset to remove some of your later commits (I think the last three, but you should check me on that) and then a push -f to update the branch. That will leave the whole question of the "extra" reference to the user fixture in Test for another day.

Details... In OneTimeSetUpCommand, we should theoretically always need to create a new fixture instance, because that's how running a fixture is supposed to work. However, sometimes when we are testing NUnit itself, the fixture is pre-created. In the original logic, the presence of a non-null fixture in the test indicated that we should not create a fixture. It worked almost like a flag! with the fixture in the context being the one that was actually used.

In addition to that basic problem, several additional issues need to be considered:

1. The TestObject is passed down from one context to its children.  When we get to a TestFixture, it is normally null __unless__ a SetUpFixture was created. So a non-null TestObject may be a preset user test fixture or an instance of a setupfixture object, with no way to tell them apart.

2. ActionAttributes get an itest, which they may cast to a TestMethod or TestFixture. In that case, they may want to access the Fixture object of the test. Of course, they could also get it from the context, but we have to think carefully about how ActionAttributes work before getting rid of ITest.Fixture.

3. The logical place to put a preset fixture might be in the WorkItem itself. However, the decision to create a fixture or use a preset one is done in the OneTimeSetUpCommand. But Commands are not supposed to know anything about what WorkItems, so the preset fixture would have to be injected into the command itself. It's not impossible, but it's a can of worms, possibly requiring redesign of the way the TestBuilder utility runs tests.

All in all, I've learned a lot from working with your changes, so I don't think it's  time wasted. Nevertheless, I'm thinking we may want to cut this off and go for a simpler solution, coming back to do more some other time. If you agree, I'll cut one or more additional issues to cover what needs to be done.

ChrisMaddock:
@CharliePoole Sounds good to me - if the structural change isn't simple, it would be good to get the straight fix in. I've reset, and merged in master.

Issue #1318, opened 2016-03-05
NUnit console runner fails with error code -100


janno-p:
I'm using NUnit3 console runner in build server and lately build does not complete, because NUnit console runner process fails. I see two NUnit related exception stacks in my build log:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at NUnit.Engine.Internal.ResultHelper.Merge(IList`1 results)
   at NUnit.Engine.Runners.AggregatingTestRunner.RunTests(ITestEventListener listener, TestFilter filter)
   at NUnit.Engine.Runners.MasterTestRunner.RunTests(ITestEventListener listener, TestFilter filter)
   at NUnit.Engine.Runners.MasterTestRunner.NUnit.Engine.ITestRunner.Run(ITestEventListener listener, TestFilter filter)
   at NUnit.ConsoleRunner.ConsoleRunner.RunTests(TestPackage package, TestFilter filter)
   at NUnit.ConsoleRunner.Program.Main(String[] args)
Running build failed.
Error:
Fake.UnitTestCommon+FailedTestsException: NUnit test failed. Process finished with exit code UnexpectedError (-100).
```

```
Unhandled Exception: System.ArgumentException: Cannot have '] ]>' inside an XML CDATA block.
   at System.Xml.XmlTextWriter.WriteCData(String text)
   at System.Xml.XmlElement.WriteElementTo(XmlWriter writer, XmlElement e)
   at System.Xml.XmlNode.get_OuterXml()
   at NUnit.Engine.TestEngineResult..ctor(XmlNode xml)
   at NUnit.Engine.Runners.ProcessRunner.CreateFailedResult(Exception e)
   at NUnit.Engine.Runners.ProcessRunner.RunTests(ITestEventListener listener, TestFilter filter)
   at NUnit.Engine.Runners.TestExecutionTask.Execute()
   at NUnit.Engine.Runners.ParallelTaskWorkerPool.ProcessTasksProc()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()
```

rprouse:
[...]

PR #1377, opened 2016-03-27
Escape CDATA tags for NUnit2 driver

ChrisMaddock:
Fixes Issue #1318.

The crash was caused by the NUnit2 driver not escaping CDATA tags. (Irritatingly, XmlDocument.CreateCDataSection() doesn't handle this, and just throws instead...)

On the issue, I mentioned potentially combining this class with NUnit.Common.XmlHelper, however didn't - the classes were more different that I first thought.


ChrisMaddock:
Looks like I didn't run the tests properly. Looking in to it...

ChrisMaddock:
The tests that broke were the DirectoryFinder tests - I've just updated the expected results, as I presume it was just pointed to a random directory.

ChrisMaddock:
Not sure what I've done to break the Travis build, can someone point me in the right direction?

```
1) Error : 
object must be MarshalByRef
  at System.Runtime.Remoting.Proxies.RealProxy.PrivateInvoke (System.Runtime.Remoting.Proxies.RealProxy rp, IMessage msg, System.Exception& exc, System.Object[]& out_args) [0x00000] in <filename unknown>:0 
```

src/NUnitEngine/Addins/nunit.v2.driver.tests/XmlExtensionsTests.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        [Test]
+        public void CDataIsEscaped()
+        {
+            string innerXml = _result.ToXml(false).InnerXml;
+            string expected = "<message><![CDATA[]]]]><![CDATA[>]]></message>";
+            Assert.That(innerXml, Is.StringContaining(expected));
+        }
+    }
+}
```
CharliePoole: The existing v2 driver test is an integration test. Specifically, it's a sampling of some of NUnit V2's own tests, referencing the V2 framework and being run using the driver. It's isolated in it's own directory so that there is no possibility of it's referencing anything except the V2 framework.

This test, which you have added, is obviously a unit test. We should undoubtedly have such tests but at the time I was creating the driver, I decided to focus on the integration/system level of testing. The problem is that by adding this single test, you had to also add a reference from the tests to the v2 driver, which invalidates the overall integration test. If we want to create unit tests for this driver, they have to be in some other assembly, perhaps in the overall addins test assembly. Alternatively, we can wait until we actually separate the build of the driver from NUnit itself, which I suspect we will eventually do.

src/NUnitEngine/Addins/nunit.v2.driver.tests/nunit.v2.driver.tests.csproj
```diff
@@ -30,6 +30,9 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="nunit.core.interfaces">
+      <HintPath>..\nunit.v2.driver\lib\nunit.core.interfaces.dll</HintPath>
+    </Reference>
```
CharliePoole: Another outcome of adding the unit test here: NUnit V2 tests are not supposed to reference the nunit.core.interfaces - the only exception, of course, being NUnit's own tests. Like the reference to the driver, this reference also makes the overall test less realistic.

CharliePoole:
The fix looks good to me. Only problem is with the introduction of a test of the driver itself into the simulated NUnit V2 test suite - see line comments.

Since it's possible that that change to the tests caused the Travis issue, let's fix the tests and see what happens next. :-)

ChrisMaddock:
Thanks @CharliePoole - it sounds like, for the sake of a single unit test, it's not worth setting up an entirely new assembly! Considered moving it to addin-tests, but not sure how to force it to use the nunit2-driver other than references the nunit 2 framework, which would affect the other tests.

Have pushed it without.

CharliePoole:
I'm embarrassed to say those are all the same reasons I didn't do a set of unit tests in the first place. However, most of the code was pre-existing and tested where I got it from, so I decided to just do an overall test and "come back to it later."

Technical Debt at its finest!

CharliePoole:
Looks good. Looks like you need to resolve changes with master before merging. There are two ways to go: (1) pull latest master changes into your branch, resolve conflicts, push and click the button or (2)
follow the command line instructions.

ChrisMaddock:
Done! :+1:

PR #1378, opened 2016-03-27
Fix TestDriver target so the tests are run correctly

rprouse:
Simple fix, they were just broken in the re-org of the Cake script. Tests are now running and passing with the standard build targets.


CharliePoole:
Failure is due to long branch name. I'll create an issue for that.

Issue #1386, opened 2016-03-31
Bug when using Assert.Equals() with types that explicitly implement IEquatable<T>


daveaglick:
If a type involved in `Assert.Equals()` implements `IEquatable<T>`, the implementation of `IEquatable<T>` is used for the comparison (as expected). However, if a type *explicitly* implements `IEquatable<T>` then the corresponding `.Equals()` method isn't found and the comparison falls back to the default (typically reference equality).

I've isolated the problem to the `NUnit.Framework.Constraints.NUnitEqualityComparer.GetCorrectGenericEqualsMethod()` method. Here's a simple repro (from LINQPad) with the current head code for `GetCorrectGenericEqualsMethod()` that demonstrates the problem:

```csharp
void Main()
{
	MethodInfo a = GetCorrectGenericEqualsMethod(typeof(A), typeof(A));
	a.Dump(); // null
	MethodInfo b = GetCorrectGenericEqualsMethod(typeof(B), typeof(B));
	b.Dump(); // correct method found
}

MethodInfo GetCorrectGenericEqualsMethod(Type first, Type second)
{
	MethodInfo[] methods = first.GetMethods();
	foreach (var method in methods)
	{
		if (method.Name == "Equals")
		{
			ParameterInfo[] prms = method.GetParameters();
			if (prms.Length == 1 &&
				prms[0].ParameterType != typeof(Object) &&
				prms[0].ParameterType.IsAssignableFrom(second))
			{
				return method;
			}
		}
	}
	return null;
}

public class A : IEquatable<A>
{
	bool IEquatable<A>.Equals(A other)
	{
		return false;
	}
}

public class B : IEquatable<B>
{
	public bool Equals(B other)
	{
		return false;
	}
}
```

I plan to submit a PR for this soon - just opening an issue for tracking purposes.

CharliePoole:
Aside from a PR, what would help to convince us this is a problem is a failing test case that we all agree should have passed. In theory we should only be using an explicit implementation if we are called with an instance specifically typed as the interface. That said, the fact that we currently lose the typing of the arguments (they are treated as objects) in NUnitEqualityComparer may force us to either always use the explicit implementation or never use it.

A perfect solution probably has to wait until we have a generic implementation of the comparer and everything leading up to it - a tall order - so compromise may be necessary. I'll be interested to see how far you can take it in your PR.

CharliePoole:
[...]

PR #1387, opened 2016-03-31
(#1386) Refactored check for IEquatable interfaces to account for explicit implementations

daveaglick:
Ended up getting the method during the check for `IEquatable<T>` existence and passing it through so that the check and the invocation are consistent. Re #1386.

### DIFF
--- a/src/NUnitFramework/tests/Constraints/NUnitEqualityComparerTests.cs
+++ b/src/NUnitFramework/tests/Constraints/NUnitEqualityComparerTests.cs
@@ -150,6 +150,24 @@ namespace NUnit.Framework.Constraints
         }
 
         [Test]
+        public void IEquatableSuccess_WhenExplicitInterfaceImplementation()
+        {
+            IEquatableWithExplicitImplementation x = new IEquatableWithExplicitImplementation(1);
+            IEquatableWithExplicitImplementation y = new IEquatableWithExplicitImplementation(1);
+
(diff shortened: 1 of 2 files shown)


CharliePoole:
The code looks good. It's a question of whether we think explicitly implemented interfaces should be made to work. Since we can't currently (easily) implement the ideal result described in #1386, I'm ready to go for this as being the best choice, especially since we are a testing framework. @rprouse What do you think?

src/NUnitFramework/tests/Constraints/NUnitEqualityComparerTests.cs
```diff
@@ -150,6 +150,15 @@ public void IEquatableDifferentTypesSuccess_WhenExpectedImplementsIEquatable()
         }
 
         [Test]
+        public void IEquatableSuccess_WhenExplicitInterfaceImplementation()
+        {
+            IEquatableWithExplicitImplementation x = new IEquatableWithExplicitImplementation(1);
+            IEquatableWithExplicitImplementation y = new IEquatableWithExplicitImplementation(1);
```
rprouse: I might be overly cautious here and it may be unneeded, but I would like to see a second test for two `IEquatableWithExplicitImplementation` with different values to show the comparer.AreEqual returns false.

rprouse:
I'm good with this. If you can add the second test I asked for, I think we can merge.

daveaglick:
@rprouse The additional test has been added - thanks both for taking a look at this.

Issue #668, opened 2015-05-28
Running from console on visual studio solution fails due to non test assemblies


tburnett80:
I am running my tests via mono on ubuntu and the tests are being found and passed just like they are from windows when I run them in NCrunch or the test runner in visual studio, the problem is my tests are being marked as failed because of the non-test assemblies containing the code being tested. Is there a way to ignore these NotRunnable assemblies to keep it from failing the tests?

![capture](https://cloud.githubusercontent.com/assets/4324080/7869542/ff137ec6-0548-11e5-8ca3-79c44b96790a.PNG)

The xml from the "Invalid" test is :

``` xml
<test-suite type="Assembly" id="68-1" name="OutsideSources.DomainServices.dll" fullname="/home/toburn/Desktop/OutsideSources/OutsideSources.DomainServices/bin/Debug/OutsideSources.DomainServices.dll" testcasecount="0" runstate="NotRunnable" result="Failed" label="Invalid">
      <properties>
        <property name="_SKIPREASON" value="Unable to locate a driver for /home/toburn/Desktop/OutsideSources/OutsideSources.DomainServices/bin/Debug/OutsideSources.DomainServices.dll" />
      </properties>
      <reason>
        <message>Unable to locate a driver for /home/toburn/Desktop/OutsideSources/OutsideSources.DomainServices/bin/Debug/OutsideSources.DomainServices.dll</message>
      </reason>
</test-suite>
```

CharliePoole:
Is there a reason you prefer to execute the solution rather than the
assemblies directly? Understanding why folks use this approach may help us
decide how to fix it.

tburnett80:
I am seeking a generic way to run all the tests I have for a particular solution. I have automation that I have no choice on that generates jenkins jobs and not having to know what all the test assemblies names are up front helps me considerably. 

This is the model I used with prior versions of Nunit and am not sure why now it considers non-test assemblies as a reason to mark the test run as failed. 

Typically I build a solution with multiple projects, if the tests fail nothing gets released, if the tests pass everything gets released.

tburnett80:
I'm attempting to use a test config, but having the ability to specify a .sln file seems counter productive if it will fail when assemblies are present that contain no tests. 

Maybe Im the only one who keeps their unit tests separated from production code its testing?

rprouse:
You are definitely not the only one who has tests in the same solution as production code, I think it is quite common. Personally, I never use the solution file, so I don't have much experience in how it worked in the past. 

[...]

PR #1393, opened 2016-04-06
Support filtering project when nunit3-console is used with a solution…

pcalin:
…. See issue #668

### DIFF
--- a/src/NUnitEngine/Addins/addin-tests/VisualStudioProjectLoaderTests.cs
+++ b/src/NUnitEngine/Addins/addin-tests/VisualStudioProjectLoaderTests.cs
@@ -176,5 +176,19 @@ namespace NUnit.Engine.Services.ProjectLoaders.Tests
                 Assert.AreEqual(1, project.GetTestPackage("Debug").SubPackages.Count, "Debug should have 1 assembly");
             }
         }
+
+        [Test]
+        public void FromSolutionWithNonNunitTestProject()
+        {
+            using (new TestResource("csharp-sample.csproj", @"csharp-sample\csharp-sample.csproj"))
+            using (new TestResource("csharp-debug-only-no-nunit.csproj", @"csharp-debug-only-no-nunit\csharp-debug-only-no-nunit.csproj"))
+            using (TestResource file = new TestResource("solution-with-non-nunit-project.sln"))
+            {
+                IProject project = _loader.LoadFrom(file.Path);
+                Assert.AreEqual(2, project.ConfigNames.Count);
+                Assert.AreEqual(1, project.GetTestPackage("Release").SubPackages.Count, "Release should have 2 assemblies");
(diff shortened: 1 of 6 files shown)


CharliePoole:
I see a few top-level problems with this, so I'm not doing line by line comments right now.

1. IProject is an extensibility interface. Changing it affects all three NUnit bundled project loader extensions as well as any user extensions.

2. The filtering only makes sense for solution files, but you have applied it to NUnit projects as well. Of course, you had to do that once you changed the extensibility interface. In fact, user-defined projects would need to support a filter as well, whether they need one or not.

3. It seems as if this could be done in a more localized way, following some of the suggestions on issue #668.

I'd like to see us resolve this issue, but preferably in a way that doesn't impact every project loader we have or which may be written in the future. Ideally, we don't want to change the extensibility interfaces or the overall API if it can be avoided.

CharliePoole:
What if we supported `assembly==...` and `assembly~=...` in our test selection language used by the `--where` option?

The parsing would be simple enough. The trick would be to apply this particular filter in the engine, before loading the assembly, rather than passing it on to the framework for handling.

pcalin:
Awesome. I'll look into passing the where options.

pcalin:
A simpler option would be to implement the filter in the MasterTestRunner.ExpandProjects. Basically apply the filter after the expansion.

CharliePoole:
I agree it might be simpler to implement in MasterTestRunner, particularly if we did it in conjunction with analyzing the assemblies. We may end up doing this but it seems to me that the ideal solution would be that the loader only returns test projects.

The reason I say that is because IProject and IProjectLoader are not intended to be general ways of working with VS or any other kind of project. They are supposed to represent collections of test assemblies. So, if it's at all possible for the VSProjectLoader to determine which projects are really test projects, I think it should do so. If it's not possible - or even too messy - then we can fall back to filtering the list of projects after the fact.

pcalin:
I implemented checks in VSSolution to verify the project references nunit.framework. Should be enough for most needs and easy to document/understand.

CharliePoole:
This certainly seems to fix it for the nunit framework, but takes us down a path where the project loader extension has to know what frameworks we support. Of course it's the logical outcome of what I asked for - in order to only return test assemblies, the loader has to know all possible test framework names! @rprouse What do you think? Is this the right direction or do we need to filter test projects out external to the VS Project loader?

rprouse:
I think this is a decent solution for now since we only support nunit at the moment. 

@CharliePoole, can we use the `IDriverService` from a plugin? We could extend `IDriverService` with something like `IEnumerable<string> GetSupportedTestFrameworkAssemblies()` that queries each of the driver factories and yield returns the supported assembly names and use that list instead of the hard coded assembly name. We would also extend IDriverFactory with the same method, enumerable in case a future driver supports multiple frameworks.

CharliePoole:
Yes, that's the sort of thing I was thinking of. The VS project loader could make use of the service to find out the names of frameworks that indicate tests.

I agree that we can do this later, however. Merging.

PR #1399, opened 2016-04-07
Fixed NullReference issue introduced by the fix for #681

pcalin:


src/NUnitEngine/nunit.engine/Runners/DirectTestRunner.cs
```diff
@@ -158,7 +158,8 @@ protected override TestEngineResult RunTests(ITestEventListener listener, TestFi
                 packages.Add(TestPackage);
             foreach (var package in packages)
             {
-                _assemblyResolver.RemovePathFromFile(package.FullName);
+                if (_assemblyResolver != null)
+                    _assemblyResolver.RemovePathFromFile(package.FullName);
```
CharliePoole: It seems like the if statement can wrap all the code that runs through the packages, starting at line 156.

pcalin:
good point

CharliePoole:
Looks good. Waiting for CI to pass.

PR #1400, opened 2016-04-07
IsExplicitMatch correctly overriden for CompositeFilters. In a Compos…

### LINKED ISSUE #1141: Explicit Tests get run when using --where with some filters
Running an Assembly with a TestFixture like the following, the Explicit Tests will be run by the console-runner if the following --where argument is used.

nunit3-console.exe NUnitFailers.dll --where="cat!=test AND cat!=test2"

```
    [TestFixture]
    public class ExplicitTest
    {
        [Test]
        [Explicit]
        public void ExplicitTestCase()
        {
            Assert.Fail("Should never be run, even not if using exclude for categories.");
        }
    }
```

>Test Run Summary
>    Overall result: Failed
>   Tests run: 1, Passed: 0, Errors: 0, Failures: 1, Inconclusive: 0
>     Not run: 0, Invalid: 0, Ignored: 0, Explicit: 0, Skipped: 0
>  Start time: 2015-12-16 14:07:53Z
>    End time: 2015-12-16 14:07:53Z
>    Duration: 0.078 seconds

Similar bug was found in a NUnit3-beta #741

stffabi:
…iteFilter the IsExplicitMatch function must always be correctly combined by calling the sub-filters IsExplicitMatch function. Fixes #1141

### DIFF
--- a/src/NUnitFramework/framework/Internal/Filters/CompositeFilter.cs
+++ b/src/NUnitFramework/framework/Internal/Filters/CompositeFilter.cs
@@ -64,6 +64,24 @@ namespace NUnit.Framework.Internal.Filters
         public IList<ITestFilter> Filters { get; private set; }
 
         /// <summary>
+        /// Checks whether the CompositeFilter is matched by a test.
+        /// </summary>
+        /// <param name="test">The test to be matched</param>
+        public abstract override bool Pass(ITest test);
+
+        /// <summary>
+        /// Checks whether the CompositeFilter is matched by a test.
+        /// </summary>
+        /// <param name="test">The test to be matched</param>
+        public abstract override bool Match(ITest test);
+
+        /// <summary>
+        /// Checks whether the CompositeFilter is explicit matched by a test.
+        /// </summary>
+        /// <param name="test">The test to be matched</param>
+        public abstract override bool IsExplicitMatch(ITest test);
+
+        /// <summary>
         /// Adds an XML node
         /// </summary>
         /// <param name="parentNode">Parent node</param>
--- a/src/NUnitFramework/tests/Internal/Filters/AndFilterTests.cs
+++ b/src/NUnitFramework/tests/Internal/Filters/AndFilterTests.cs
@@ -71,6 +71,91 @@ namespace NUnit.Framework.Internal.Filters
         }
 
         [Test]
+        public void ExplicitMatchWithNotTest()
+        {
+            var filter = new AndFilter(
+                new NotFilter(
+                    new CategoryFilter("Dummy")),
+                new NotFilter(
+                    new IdFilter(_dummyFixture.Id)));
+
+            Assert.False(filter.IsExplicitMatch(_topLevelSuite));
+            Assert.False(filter.IsExplicitMatch(_dummyFixture));
+            Assert.False(filter.IsExplicitMatch(_dummyFixture.Tests[0]));
+
+            Assert.That(filter.Match(_anotherFixture), "4");
+        }
+
+        /// <summary>
+        /// Generic combination tests for the <see cref="AndFilter"/>. This test checks that the
+        /// <see cref="AndFilter"/> correctly combines the results from its sub-filters.
+        /// Furthermore it checks that the result from the correct match-function of the
+        /// sub-filters is used to calculate the AND combination.
+        /// 
+        /// The input is an array of booleans (<paramref name="inputBooleans"/>). For each boolean
+        /// value a <see cref="MockTestFilter"/> is added to the <see cref="AndFilter"/>
+        /// whose match-function (defined through the paramter <paramref name="matchFunction"/>)
+        /// evaluates to the boolean value indicated.
+        /// Afterwrards the requested match-function (<paramref name="matchFunction"/>) is called
+        /// on the <see cref="AndFilter"/> and is compared to the expected result
+        /// (<paramref name="expectedResult"/>).
+        /// This tests also throws an exception if the match-function call from the
+        /// <see cref="AndFilter"/> calls not the same match-function on the
+        /// <see cref="MockTestFilter"/>, thus checking that the <see cref="AndFilter"/>
+        /// combines the correct results from the sub-filters.
+        /// 
+        /// See also <see cref="MockTestFilter"/>.
+        /// </summary>
+        [TestCase(new bool[] { false, false}, false, MockTestFilter.MatchFunction.IsExplicitMatch)]
+        [TestCase(new bool[] { true, false }, false, MockTestFilter.MatchFunction.IsExplicitMatch)]
+        [TestCase(new bool[] { false, true }, false, MockTestFilter.MatchFunction.IsExplicitMatch)]
+        [TestCase(new bool[] { true, true }, true, MockTestFilter.MatchFunction.IsExplicitMatch)]
+        [TestCase(new bool[] { false, false }, false, MockTestFilter.MatchFunction.Match)]
+        [TestCase(new bool[] { true, false }, false, MockTestFilter.MatchFunction.Match)]
+        [TestCase(new bool[] { false, true }, false, MockTestFilter.MatchFunction.Match)]
+        [TestCase(new bool[] { true, true }, true, MockTestFilter.MatchFunction.Match)]
+        [TestCase(new bool[] { false, false }, false, MockTestFilter.MatchFunction.Pass)]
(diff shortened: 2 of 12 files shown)


src/NUnitFramework/framework/Internal/Filters/CompositeFilter.cs
```diff
@@ -64,6 +64,24 @@ public void Add(ITestFilter filter)
         public IList<ITestFilter> Filters { get; private set; }
 
         /// <summary>
+        /// Checks whether the CompositeFilter is matched by a test.
+        /// </summary>
+        /// <param name="test">The test to be matched</param>
+        public abstract override bool Pass(ITest test);
```
rprouse: I assume that you added these overrides to ensure they are implemented in all of the child classes. Correct?

stffabi: Yes, exactly.

rprouse:
The code looks good, but we need tests for this. If you look in the framework tests under `Internal\Filters`, there are `AndFilterTests` and `OrFilterTests`. These have `ExplicitMatchTests` which clearly are not working correctly. Can you determine what is wrong with these tests and fix them? If you can, I would actually like to see passing and failing tests in both for Match, Pass and ExplicitMatch. I think that is likely the problem, the positive tests are all correct, but we never test for failed matches.

I would recommend taking a TDD approach to the tests. Comment out your fixes, add/fix the tests to get them failing, then uncomment your code to prove that your tests and your fixes are correct. Whenever I work on any issue in NUnit, I try to always start with the tests because it is too easy to write tests after the fact that pass, but do not properly test the fix, like this one. :smile:

stffabi:
Thanks for pointing out where the tests are, I'll take a look into it.

stffabi:
The `ExplicitMatchTests` are correct, the problem is that for the specific test-case there's no difference in the result as long as the sub-filter doesn't override `IsExcplicitMatch`.

I've added a test-case with the `NotFilter`, which overrides the `IsExcplicitMatch` function. Furthermore I've added some generic `CombinationTest`, these first test the combination of the sub-filters in the AndFilter/OrFilter and second also checks that the expected "Match"-Function gets called and not a wrong one.

src/NUnitFramework/framework/Internal/Filters/CompositeFilter.cs
CharliePoole: I find the override a bit confusing in reading the code. I would definitely add it if this were a class intended to be extended by users, but it's purely internal.

src/NUnitFramework/tests/Internal/Filters/AndFilterTests.cs
```diff
@@ -71,6 +71,74 @@ public void ExplicitMatchTest()
...
+
+                filters.Add(strictFilter);
+            }
+
+            var filter = new AndFilter(filters.ToArray());
+            bool calculatedResult = Match(filter, expectedCall);
+            Assert.AreEqual(expectedResult, calculatedResult);
+        }
+
+        private bool Match(
```
CharliePoole: Having a test method with the same name as one of the underlying TestFilter methods is a bit confusing.

src/NUnitFramework/tests/Internal/Filters/AndFilterTests.cs
```diff
@@ -71,6 +71,74 @@ public void ExplicitMatchTest()
...
+            switch (expectedCall)
+            {
+                case StrictIdFilterForTests.EqualValueFunction.IsExplicitMatch:
+                    return filter.IsExplicitMatch(_dummyFixture);
+                case StrictIdFilterForTests.EqualValueFunction.Match:
+                    return filter.Match(_dummyFixture);
+                case StrictIdFilterForTests.EqualValueFunction.Pass:
+                    return filter.Pass(_dummyFixture);
+                default:
+                    throw new AssertionException(
```
CharliePoole: Don't throw the AssertionException, rather, use Assert.Fail as we tell users to do. :-) This will stop working as soon as we change how failures are reported. Alternatively, if you want to convey that this is an error in how the test was written rather than a true failure, throw something else, like an ArgumentException.

src/NUnitFramework/framework/Internal/Filters/CompositeFilter.cs
stffabi: I think that's a matter of preference. I personally prefer it even in purely internal code to force implementers to override methods that doesn't make sense how they are implemented in the base class.

But I can drop that from the commit, that's not a big deal :smile:.

src/NUnitFramework/tests/Internal/Filters/OrFilterTests.cs
```diff
@@ -76,6 +76,58 @@ public void ExplicitMatchTest()
...
+        [TestCase(new bool[] { false, true }, true, StrictIdFilterForTests.EqualValueFunction.Match)]
+        [TestCase(new bool[] { true, true }, true, StrictIdFilterForTests.EqualValueFunction.Match)]
+        [TestCase(new bool[] { false, false }, false, StrictIdFilterForTests.EqualValueFunction.Pass)]
+        [TestCase(new bool[] { true, false }, true, StrictIdFilterForTests.EqualValueFunction.Pass)]
+        [TestCase(new bool[] { false, true }, true, StrictIdFilterForTests.EqualValueFunction.Pass)]
+        [TestCase(new bool[] { true, true }, true, StrictIdFilterForTests.EqualValueFunction.Pass)]
+        public void CombineTest(IEnumerable<bool> inputBooleans, bool expectedResult,
+            StrictIdFilterForTests.EqualValueFunction expectedCall)
+        {
+            var filters = new List<StrictIdFilterForTests>();
```
CharliePoole: Would you add a comment in this and other tests like it - or maybe on the StrictIdFilterForTests class that explains what you are tryiing to test. It really takes a bit of mental exercise to figure it out and we would like folks to understand our tests, which are often used as examples.

CharliePoole:
Aside from some minor comments, this looks good to me.

src/NUnitFramework/framework/Internal/Filters/CompositeFilter.cs
CharliePoole: Except it's almost certain that all the derived classes we will ever have
are already implemented!

But it certainly does no harm.

CharliePoole: Except it's almost certain that all the classes we will ever have

In src/NUnitFramework/framework/Internal/Filters/CompositeFilter.cs
<https://github.com/nunit/nunit/pull/1400#discussion_r59176879>:

> @@ -64,6 +64,24 @@ public void Add(ITestFilter filter)
>          public IList<ITestFilter> Filters { get; private set; }
>
>          /// <summary>
> +        /// Checks whether the CompositeFilter is matched by a test.
> +        /// </summary>
> +        /// <param name="test">The test to be matched</param>
> +        public abstract override bool Pass(ITest test);

I think that's a matter of preference. I personally prefer it even in
purely internal code to force implementers to override methods that doesn't
make sense how they are implemented in the base class.

But I can drop that from the commit, that's not a big deal [image: :smile:].

—
You are receiving this because you commented.
Reply to this email directly or view it on GitHub
<https://github.com/nunit/nunit/pull/1400/files/9ca0604beef21f693409647b3389283c4333caf8#r59176879>

stffabi:
Yes, I agree with you about clarifying these kind of tests. I think the clearest thing would be to use some kind of mocking framework for tests like these, e.g. Moq. I have taken a look into the source, but couldn't find something like that.

Would you be interested to use a mocking framework for such tests? Would that be in the interest of the project, if yes I would be willing to include e.g. Moq.

CharliePoole:
We use NSubstitute where we use mocks, but don't have any now in the 3.0 code. It could be a bit much to introduce just for this fix.

stffabi:
Okey, then I will only clarify these tests with comments. The changes will be around tomorrow.

CharliePoole:
This looks good to merge for me. @rprouse?

rprouse:
Good by me too, thanks for adding the tests. 

FYI, in the future, add a comment after you push updates to a PR letting us know. We get notified by email for comments, but not for code commits, so we sometimes miss updates like this.

Thanks for the contribution.

oznetmaster:
Why the use of TestFilter in the foreach statements when previously we have always used ITestFilter?

Not only is it inconsistent, but it throws up warning.

Issue #1406, opened 2016-04-10
TextCapture is not threadsafe but is used to intercept calls that are expected to be threadsafe.


jtattermusch:
It looks like the TextWriter used in https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/Execution/TextCapture.cs is not thread safe, but the TextCapture is used to intercept  Console.WriteLine (which itself is thread safe).

A solution would be to use a synchronized TextWriter?

This is the stack trace I'm seeing in one of my tests
```
Unhandled Exception:
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: destinationIndex
  at System.Text.StringBuilder.ThreadSafeCopy (System.Char* sourcePtr, System.Char[] destination, Int32 destinationIndex, Int32 count) <0x1909ea0 + 0x000aa> in <filename unknown>:0 
  at System.Text.StringBuilder.Append (System.Char* value, Int32 valueCount) <0x1909970 + 0x0004c> in <filename unknown>:0 
  at System.Text.StringBuilder.Append (System.Char[] value, Int32 startIndex, Int32 charCount) <0x1906ec0 + 0x000a4> in <filename unknown>:0 
  at System.IO.StringWriter.Write (System.Char[] buffer, Int32 index, Int32 count) <0x185b9a0 + 0x0009f> in <filename unknown>:0 
  at System.IO.TextWriter.WriteLine (System.String value) <0x185e090 + 0x00174> in <filename unknown>:0 
  at NUnit.Framework.Internal.Execution.TextCapture.WriteLine (System.String value) <0x7d0900 + 0x0004a> in <filename unknown>:0 
  at System.Console.WriteLine (System.String value) <0x1b1a540 + 0x00045> in <filename unknown>:0 
```

jtattermusch:
I believe this is where the TextWrapper gets created.

https://github.com/nunit/nunit/blob/526534e93202f1da3f371d57d2b358341238c29f/src/NUnitFramework/nunitlite/TextRunner.cs#L123

jtattermusch:
Based on https://msdn.microsoft.com/en-us/library/system.io.streamwriter(v=vs.110).aspx System.IO.StreamWriter is not thread-safe. Perhaps ExtendedTextWrapper can add the synchronization while wrapping TextWriter methods?

CharliePoole:
Two clarifications please:
[...]

PR #1408, opened 2016-04-11
Allow disabling of output capturing

jtattermusch:
-- allow to fully disable output capturing for test run using NUnitLite's   TextRunner, that allows the user to work around various TextCapture thread safety issues (not synchronizing access to StringWriter, lazy initialization of some fields).
. Also, the choice if Console.Out should or should not be captures should be left upon user.
-- if Console.Out and Console.Err is captured to a file, make sure the StreamWriter is threadsafe.

Related issue: https://github.com/nunit/nunit/issues/1406

### DIFF
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -47,6 +47,7 @@ namespace NUnit.Framework.Api
 
         private ITestAssemblyBuilder _builder;
         private ManualResetEvent _runComplete = new ManualResetEvent(false);
+        private readonly bool _captureOutput;
 
 #if !SILVERLIGHT && !NETCF && !PORTABLE
         // Saved Console.Out and Console.Error
@@ -65,9 +66,19 @@ namespace NUnit.Framework.Api
         /// Initializes a new instance of the <see cref="NUnitTestAssemblyRunner"/> class.
         /// </summary>
         /// <param name="builder">The builder.</param>
-        public NUnitTestAssemblyRunner(ITestAssemblyBuilder builder)
+        public NUnitTestAssemblyRunner(ITestAssemblyBuilder builder) : this(builder, true)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="NUnitTestAssemblyRunner"/> class.
+        /// </summary>
+        /// <param name="builder">The builder.</param>
+        /// <param name="captureOutput">If true, capture test's stdout and stderr (if supported on given platform).</param>
+        public NUnitTestAssemblyRunner(ITestAssemblyBuilder builder, bool captureOutput)
         {
             _builder = builder;
+            _captureOutput = captureOutput;
         }
 
         #endregion
@@ -251,8 +262,11 @@ namespace NUnit.Framework.Api
             _savedOut = Console.Out;
             _savedErr = Console.Error;
 
-            Console.SetOut(new TextCapture(Console.Out));
-            Console.SetError(new TextCapture(Console.Error));
+            if (_captureOutput)
+            {
+                Console.SetOut(new TextCapture(Console.Out));
+                Console.SetError(new TextCapture(Console.Error));
+            }
 #endif
 
 #if PARALLEL
@@ -351,8 +365,11 @@ namespace NUnit.Framework.Api
 #endif
 
 #if !SILVERLIGHT && !NETCF && !PORTABLE
-            Console.SetOut(_savedOut);
-            Console.SetError(_savedErr);
+            if (_captureOutput)
+            {
+                Console.SetOut(_savedOut);
+                Console.SetError(_savedErr);
+            }
 #endif
 
             _runComplete.Set();
--- a/src/NUnitFramework/nunitlite/TextRunner.cs
+++ b/src/NUnitFramework/nunitlite/TextRunner.cs
@@ -72,6 +72,7 @@ namespace NUnitLite
         private Assembly _testAssembly;
         private readonly List<Assembly> _assemblies = new List<Assembly>();
         private ITestAssemblyRunner _runner;
+        private readonly bool _captureOutput;
 
         private NUnitLiteOptions _options;
         private ITestListener _teamCity = null;
@@ -87,7 +88,17 @@ namespace NUnitLite
         //// <summary>
         //// Initializes a new instance of the <see cref="TextRunner"/> class.
         //// </summary>
-        public TextRunner() { }
+        public TextRunner() : this(null, true)
+        {
+        }
+
+        //// <summary>
+        //// Initializes a new instance of the <see cref="TextRunner"/> class.
+        //// </summary>
+        //// <param name = "captureOutput" > If true, capture test's stdout and stderr (if supported on given platform).</param>
+        public TextRunner(bool captureOutput) : this(null, captureOutput)
+        {
+        }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="TextRunner"/> class
@@ -95,9 +106,20 @@ namespace NUnitLite
         /// </summary>
         /// <param name="testAssembly"></param>
         /// </summary>
-        public TextRunner(Assembly testAssembly)
+        public TextRunner(Assembly testAssembly) : this(testAssembly, true)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TextRunner"/> class
+        /// specifying a test assembly whose tests are to be run.
+        /// </summary>
+        /// <param name="testAssembly"></param>
+        /// <param name="captureOutput">If true, capture test's stdout and stderr (if supported on given platform).</param>
+        public TextRunner(Assembly testAssembly, bool captureOutput)
         {
             _testAssembly = testAssembly;
+            _captureOutput = captureOutput;
         }
 
         #endregion
@@ -120,7 +142,7 @@ namespace NUnitLite
             ExtendedTextWriter outWriter = null;
             if (options.OutFile != null)
             {
-                outWriter = new ExtendedTextWrapper(new StreamWriter(Path.Combine(options.WorkDirectory, options.OutFile)));
+                outWriter = new ExtendedTextWrapper(TextWriter.Synchronized(new StreamWriter(Path.Combine(options.WorkDirectory, options.OutFile))));
                 Console.SetOut(outWriter);
             }
             else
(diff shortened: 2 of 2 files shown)


jtattermusch:
@CharliePoole  FYI

CharliePoole:
Thanks for the PR. We're glad to have more contributors. In this case, I see a few problems, but I hope it won't discourage you.

1. Just so you know, it's slightly bad form to take over an issue that someone else is already working on, as indicated by the assignee field. No biggy, but you please check first if you see someone else has the task.

2. By allowing disabling the output, you are making a fairly big change, one we have not discussed publicly with users. I see that you preserved the old default behavior, but even so I'm hesitant to expose this new functionality without thinking through the implications.

3. Another problem I personally have with disabling is that you seem to be giving up on solving the original, fundamental problem that you and I discussed. If there is a problem with thread safety in the capture of the output, we need to solve it. Based on your report it's broken in a fairly fundamental way and I prefer fixing it to turning off the capture as a workaround.

4. Your approach of adding a new constructor parameter to the framework isn't the way we generally pass options in. We use the test package settings or (for nunitlite) the runSettings. The framework reflection API is one of our primary documented APIs, so I'd rather not change it without a very serious reason. The settings we pass in provide flexibility, since they are simply ignored if a particular version of the framework doesn't support them.

5. I realize you are right about the need to make the code thread-safe when we redirect to a file. In that case, we are not taking advantage of the built-in thread-safety of the console.

jtattermusch:
On Mon, Apr 11, 2016 at 11:41 AM, CharliePoole <notifications@github.com>
wrote:

> Thanks for the PR. We're glad to have more contributors. In this case, I
> see a few problems, but I hope it won't discourage you.
>
>    1.
>
>    Just so you know, it's slightly bad form to take over an issue that
>    someone else is already working on, as indicated by the assignee field. No
>    biggy, but you please check first if you see someone else has the task.
>
> Thanks for letting me know. To explain myself - I just thought that coming
up with an actual code change to demonstrate what changes I think could be
made is more effective than trying to describe that by words. I wanted to
rather use the PR as a means to communicate my thoughts than an attempt to
interfere with an ongoing work of yours - after all, you have all the
rights to just reject the PR.


>
>    1.
>
>    By allowing disabling the output, you are making a fairly big change,
>    one we have not discussed publicly with users. I see that you preserved the
>    old default behavior, but even so I'm hesitant to expose this new
>    functionality without thinking through the implications.
>
> Agreed.


>
>    1.
>
>    Another problem I personally have with disabling is that you seem to
>    be giving up on solving the original, fundamental problem that you and I
>    discussed. If there is a problem with thread safety in the capture of the
>    output, we need to solve it. Based on your report it's broken in a fairly
>    fundamental way and I prefer fixing it to turning off the capture as a
>    workaround.
>
> Makes sense. Btw, there are several places where the thing is broken and
the concept of setting a thread-local context to collect a per-test result
can never work properly with multithreaded tests. So I think at some point
we will need a way to not intercept Console.Out and Console.Err (btw, when
using a portable version of NUnit, Console.Out and Console.Err are not a
thing so having ability to disable intercepting Console.Out and Console.Err
for .NET 4.5 doesn't seem that odd).


>
>    1.
>
>    Your approach of adding a new constructor parameter to the framework
>    isn't the way we generally pass options in. We use the test package
>    settings or (for nunitlite) the runSettings. The framework reflection API
>    is one of our primary documented APIs, so I'd rather not change it without
>    a very serious reason. The settings we pass in provide flexibility, since
>    they are simply ignored if a particular version of the framework doesn't
>    support them.
>
>
I should have specified that properly, but my PR only was supposed to give
an idea of what kind of changes need to be done.


>
>    1.
>
>    I realize you are right about the need to make the code thread-safe
>    when we redirect to a file. In that case, we are not taking advantage of
>    the built-in thread-safety of the console.
>
>
I can create a separate PR to add the Text.Synchronized when we redirect to
a file. It's not a full fix, but I don't see that as a problem - we can
proceed incrementally.

CharliePoole:
@rprouse It would be good if you could weigh in on this.

@jtattermusch Thanks for the explanations. Yup, for us developers sometimes writing the code is the easiest way to express things.

Have you tried your fix when running with parallel tests? I'm thinking that the output may become difficult to interpret when running multiple tests together, without any indication of which test produced which output.

If it takes us a while to discuss this, then a separate PR for the case of output to a file would be a very good idea. We should check the console runner as well, although i think it's output is handled safely by being buffered at several stages along the line.

jtattermusch:
On Mon, Apr 11, 2016 at 2:53 PM, CharliePoole <notifications@github.com>
wrote:

> @rprouse <https://github.com/rprouse> It would be good if you could weigh
> in on this.
>
> @jtattermusch <https://github.com/jtattermusch> Thanks for the
> explanations. Yup, for us developers sometimes writing the code is the
> easiest way to express things.
>
> Have you tried your fix when running with parallel tests? I'm thinking
> that the output may become difficult to interpret when running multiple
> tests together, without any indication of which test produced which output.
>
I certainly agree that the effort to capture the outputs per tests is
legitimate and can work for many of the simple tests that are single
threaded in their nature. For projects that need to have more complicated
tests though (and need multiple threads), it is virtually impossible to
capture the per-test output if you allow more tests to be running at the
same time (which happens to be the case for NUnit3) - because there's just
no way to recognize to which test does which output belong. In those
situations, I think it's fair to allow users to just handle the problem of
not-well-readable outputs themselves and letting them configure NUnit to
not try to be clever when intercepting Console.Out and Console.Err.

My problem currently is that I'd love to upgrade to NUnit 3.2 (from NUnit
2.6.4), but my tests have become flaky by upgrading the version thanks to
the non-threadsafe nature of output capturing - which is a feature I don't
need and I have no way to disable it.

CharliePoole:
You may already understand this, but I want to make sure... NUnit 3 __does__ handle assigning the output to the proper test __except__ in the case where the test or SUT creates secondary threads.

What I would expect to see currently in the output for each test is:
1. Any output from secondary threads
2. The name of the test if you let labels default.
3. Any output from the primary thread.

Groups 2 and 3 would always be together and uninterupted but any output in group 1 might come from multiple threads spawned by multiple tests. You could make the order be 2, 1, 3 by using --labels=All.

OTOH, if we eliminate capture, then output in group 2 would not appear (unless you used --labels:All) and groups 1 and 2 could appear intermixed and in any order.

So it seems that there may be some legibility improvement in capturing, even if it doesn't work for secondary threads.

One way we have discussed fixing this problem in the past is to give you a way to spawn a thread that NUnit knows about and which will be associated with the same TestExecutionContext as the thread doing the creation. This would only be useful for threads created by the test itself rather than the SUT. Would it help in your situation? We could give TestContext a static method that created a thread, setting the CallContext to contain the current TestExecutionContext.

rprouse:
Time to weigh in :smile: 

I am not against the overall concept of a command line flag to not capture output, but I would prefer to leave the default as it is and have a flag to not capture output. We made the default what it is because of parallel tests and most people do not create additional threads in their tests, so I think the default we currently have is correct. Also, as @CharliePoole says, this will lead to problems interpreting output for parallel tests, so I see it as an advanced setting for users who opt in and understand that it may conflict with other command line options.

I don't like that this is only being added to NUnitLite and not to the console. I think that NUnitLite should be a subset of the console. That said, I haven't looked at the console code yet :wink: 

Doing this for both the console and NUnitLite would also give users with long running tests a workaround to seeing the progress of their tests as they run. We could even add the setting to the adapter.

>One way we have discussed fixing this problem in the past is to give you a way to spawn a thread that NUnit knows about and which will be associated with the same TestExecutionContext as the thread doing the creation. This would only be useful for threads created by the test itself rather than the SUT. Would it help in your situation?

I don't think this is a viable option with nearly every library using async these days. I would expect that spawning off threads within a test would be the minority. @jtattermusch what is your situation?

>I can create a separate PR to add the Text.Synchronized when we redirect to
a file. It's not a full fix, but I don't see that as a problem - we can
proceed incrementally.

I would like to see that as a separate PR. It can be the basis to discuss and move toward the proper fix or at least it will give us code to talk about.

jtattermusch:
On Mon, Apr 11, 2016 at 5:13 PM, Rob Prouse <notifications@github.com>
wrote:

> Time to weigh in [image: :smile:]
>
> I am not against the overall concept of a command line flag to not capture
> output, but I would prefer to leave the default as it is and have a flag to
> not capture output. We made the default what it is because of parallel
> tests and most people do not create additional threads in their tests, so I
> think the default we currently have is correct. Also, as @CharliePoole
> <https://github.com/CharliePoole> says, this will lead to problems
> interpreting output for parallel tests, so I see it as an advanced setting
> for users who opt in and understand that it may conflict with other command
> line options.
>

Agreed, a command line flag would be useful and the default should be to
capture (current behavior).


> I don't like that this is only being added to NUnitLite and not to the
> console. I think that NUnitLite should be a subset of the console. That
> said, I haven't looked at the console code yet [image: :wink:]
>
If possible, both the console and NUnitLite should support that. I believe
that NUnitTestAssembly runner is shared between both, so this should well
be doable.


> Doing this for both the console and NUnitLite would also give users with
> long running tests a workaround to seeing the progress of their tests as
> they run. We could even add the setting to the adapter.
>
> One way we have discussed fixing this problem in the past is to give you a
> way to spawn a thread that NUnit knows about and which will be associated
> with the same TestExecutionContext as the thread doing the creation. This
> would only be useful for threads created by the test itself rather than the
> SUT. Would it help in your situation?
>
> I don't think this is a viable option with nearly every library using
> async these days. I would expect that spawning off threads within a test
> would be the minority. @jtattermusch <https://github.com/jtattermusch>
> what is your situation?
>
"Marking the threads" is not a viable option. It might be doable for
threads spawned by the tests themselves, but there's no way you can do it
for threads spawned by the code you are trying to test (I have many in the
project I'm developing). Also, thing become more complicated when some of
your code runs in a threadpool  (Tasks not associated with an async
operation, which is a very common thing).
The async methods themselves might not be that much of an issue as the
CallContext is supposed to flow with the async execution flow (but I
haven't really dug deep to find out if that's true).

Basically, if you are trying to run integration tests that use lots of
threads using NUnit (and I don't see why you shouldn't do that), the best
option seems to be to use --workers=1 and   --nocaptureoutput (the new
cmdline argument I'm trying to advocate). Then the output you'll see will
be mostly correct in terms of what WriteLine happened sooner/later in time,
which can be really important for debugging complex intergration tests.


> I can create a separate PR to add the Text.Synchronized when we redirect to
> a file. It's not a full fix, but I don't see that as a problem - we can
> proceed incrementally.
>
> I would like to see that as a separate PR. It can be the basis to discuss
> and move toward the proper fix or at least it will give us code to talk
> about.
>

I will create the PR soon.

CharliePoole:
@jtattermusch Did you intend to close this request?

jtattermusch:
Yes, it seemed that more changes will be needed to address this properly.

PR #1412, opened 2016-04-12
Add NUnit Framework targeting .NET 3.5

### LINKED ISSUE #1349: Collision on System.Func from nunit.framework with System.Core in .Net 3.5 (CS0433)
**Problem**
When migrating the .Net 3.5 build of my test project from NUnit 2 to NUnit 3.2.0 I ran into the issue that `Func<T1,T2,TResult>` is defined twice, once in `nunit.framework` and once in `System.Core`. This lead to a compile time error whenever I'm trying to use this Func delegate directly within the test project:

> CS0433	The type 'Func<T1, T2, TResult>' exists in both 'System.Core, Version=3.5.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' and 'nunit.framework, Version=3.2.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb'

I cannot drop the reference to System.Core since it is apparently being added automatically by the build tools.

**Cause**
NUnit backports `Func<T1,T2,TResult>` in [Net20Compatibility.cs](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Compatibility/Net20Compatibility.cs#L35) because in .Net 2.0 it was not available yet. But it is part of System.Core since .Net 3.5.

NUnit used to have a build specifically for .Net 3.5 where this type was not back-ported, so everything worked fine. Unfortunately the .Net 3.5 build was removed in #512 #516, resulting in the .Net 2.0 build being used for .Net 3.5 and hence the collision.

**Workaround**
Might be possible to resolve the conflict with `extern alias` (currently verifying).

rprouse:
Fixes #1349

Adds the 3.5 Framework projects, updates Cake build, nuget packages and MSI. Also updated several copyright years that I found. @CharliePoole can you think of anything I missed? I tested by building, testing and packaging.


rprouse:
I didn't run the engine tests since I didn't think they would be affected, but obviously they are. I will post a fix soon. I am also going to revert the change to the License.RTF and hand edit the file.

Feel free to review in the current state and I will merge once that is done and I make those fixes.

rprouse:
License is fixed and the engine tests.

CharliePoole:
Looks good to me. The thing with the engine tests is rather dumb but I didn't want to set up a bunch of dummy files to test our ability to find them and mocking seemed like too much work at the time.

Issue #1348, opened 2016-03-14
TypeLoadException in nunit3-console 3.0.1


yaakov-h:
When using NUnit.Console 3.0.1 from the NuGet gallery, I'm getting an exception when starting the runner.

I'm only seeing this behaviour on some PCs.

```
NUnit Console Runner 3.0.5813
Copyright (C) 2015 Charlie Poole

System.TypeLoadException: Could not load type 'NUnit.Engine.IAvailableRuntimes' from assembly 'nunit.engine.api, Version=3.0.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb'.
   at NUnit.Engine.TestEngine.Initialize()
   at NUnit.Engine.TestEngine.NUnit.Engine.ITestEngine.get_Services()
   at NUnit.ConsoleRunner.ConsoleRunner..ctor(ITestEngine engine, ConsoleOptions options, ExtendedTextWriter writer)
   at NUnit.ConsoleRunner.Program.Main(String[] args)
```

This issue only occurs when trying to run NUnit.Console on a PC that has NUnit 3.2 installed globally by MSI.

rprouse:
@CharliePoole looks like an error with our engine resolution that we didn't think of. The console is finding the newer engine in the install directory and using that. Any ideas what we can do about it?

CharliePoole:
Ah ha! I didn't think of this. The new engine is getting loaded. It references the api assembly that is found in its own directory. But the old api assembly is already in memory.

@yaakov-h A workaround is to upgrade the console nuget package to 3.2.  You don't have to upgrade the framework if you don't want to.

A temporary fix we could make would be to only use the local engine.

A longer term fix that allows dropping in new engine releases will take some thinking.

yaakov-h:
Why is the old runner loading the new engine?

CharliePoole:
Because nunit.engine.api is designed to find the newest engine available and use it, within limits set by the client program - nunit3-console in this case. Clearly, we have a glitch, as described in my note above.

rprouse:
@CharliePoole, should we do a temporary fix for this for 3.2.1?

CharliePoole:
Good idea. It's just a matter of having the console runner specify `true` for privateCopy when it calls TestEngineActivator.CreateInstance. That's what Ive just done for the gui.

rprouse:
@CharliePoole, sounds good. I have put in the 3.2.1 milestone and assigned it to me for the temporary fix.

[...]

PR #1413, opened 2016-04-14
Temporary fix for #1348

rprouse:
Made loading the engine use local only. Also fixed the .NET 3.5 tests which were broken between two PR merges.


CharliePoole:
Looks OK to me once the CI passes.

rprouse:
One AppVeyor build is passing, the second is a bad link, so I am overriding and merging.

PR #1415, opened 2016-04-15
Use Synchronized writer for captured results

### LINKED ISSUE #1406: TextCapture is not threadsafe but is used to intercept calls that are expected to be threadsafe.

CharliePoole:
This fixes #1406 

In framework builds that support parallel execution, TestResult now uses a synchronized TextWriter to write to the buffer that is used for text capture. This allows multiple threads that are part of the same test to write text safely.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -24,7 +24,7 @@
 using System;
 using System.Globalization;
 using System.IO;
-using System.Text.RegularExpressions;
+using System.Text;
 using NUnit.Framework.Interfaces;
 
 namespace NUnit.Framework.Internal
@@ -58,7 +58,8 @@ namespace NUnit.Framework.Internal
         /// </summary>
         private System.Collections.Generic.List<ITestResult> _children;
 
-        private StringWriter _outWriter;
+        private StringBuilder _output = new StringBuilder();
+        private TextWriter _outWriter;
         private double _duration;
 
         #endregion
@@ -194,12 +195,16 @@ namespace NUnit.Framework.Internal
         /// <summary>
         /// Gets a TextWriter, which will write output to be included in the result.
         /// </summary>
-        public StringWriter OutWriter
+        public TextWriter OutWriter
         {
             get
             {
                 if (_outWriter == null)
-                    _outWriter = new StringWriter();
+#if PORTABLE || SILVERLIGHT
+                    _outWriter = new StringWriter(_output);
+#else
+                    _outWriter = TextWriter.Synchronized(new StringWriter(_output));
+#endif
 
                 return _outWriter;
             }
@@ -210,14 +215,12 @@ namespace NUnit.Framework.Internal
         /// </summary>
         public string Output
         {
-            get { return _outWriter == null
-                ? string.Empty
-                : _outWriter.ToString();  }
+            get { return _output.ToString(); }
         }
 
-        #endregion
+#endregion
 
-        #region IXmlNodeBuilder Members
+#region IXmlNodeBuilder Members
 
         /// <summary>
         /// Returns the Xml representation of the result.
@@ -286,7 +289,7 @@ namespace NUnit.Framework.Internal
             return thisNode;
         }
 
-        #endregion
+#endregion
 
         /// <summary>
         /// Adds a child result to this result, setting this result's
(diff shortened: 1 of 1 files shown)


src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -194,12 +195,16 @@ public System.Collections.Generic.IList<ITestResult> Children
...
         /// Gets a TextWriter, which will write output to be included in the result.
         /// </summary>
-        public StringWriter OutWriter
+        public TextWriter OutWriter
         {
             get
             {
                 if (_outWriter == null)
-                    _outWriter = new StringWriter();
+#if PORTABLE || SILVERLIGHT
```
rprouse: So, not fixed for Portable or SL. Should we create a secondary issue?

CharliePoole:
Well, they don't support parallel, so I think it's OK.

rprouse:
Good point. Reviewing too early in the morning :smile:

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -194,12 +195,16 @@ public System.Collections.Generic.IList<ITestResult> Children
         /// <summary>
         /// Gets a TextWriter, which will write output to be included in the result.
         /// </summary>
-        public StringWriter OutWriter
+        public TextWriter OutWriter
         {
             get
             {
                 if (_outWriter == null)
```
jtattermusch: Btw, the lazy initialization is not threadsafe. So if you are very unluckly, you can end up with two separate StringWriter instances writing to the same _output (each of them is synchronized on its own), which is still not threadsafe.

CharliePoole:
Ouch! Good catch. I'll put in a fix.

jtattermusch:
If I'm not mistaken, there are more places where things get lazy initialized across the stack. Ideally, all of them should be reviewed. Btw, the lazy initialization doesn't really buy much in cases like this one, so I think it would be safer to initialize things in constructor unless there is some costly resource getting created when during the lazy initialization.

rprouse:
@jtattermusch I agree with the comment about lazy initialization. I will create an issue to track this since it is on a closed PR. Thanks for the review.

CharliePoole:
I used a regular expression to find properties with lazy initialization. I'll write individual comments about each of these as needed.

1. ParallelWorkItemDispatcher.ParallelQUeue
2. ParallelWorkItemDispatcher.ParallelSTAQUeue
3. ParallelWorkItemDispatcher.NonParallelQUeue
4. ParallelWorkItemDispatcher.NonParallelSTAQUeue
5. OSPlatform.CurrentPlatform
6. Randomizer.SeedGenerator
7. TestResult.Children
8. RuntimeFramework.CurrentFramework
9. TestExecutionContext.CurrentContext
10. TestExecutionContext.Dispatcher
11. TestExecutionContext.RandomGenerator

PR #1417, opened 2016-04-17
Display list of projects detected by VSSolution

pcalin:
I found myself very unable to tell what test projects were ran so I think this may be useful for others as well.

### DIFF
--- a/src/NUnitEngine/Addins/vs-project-loader/VSSolution.cs
+++ b/src/NUnitEngine/Addins/vs-project-loader/VSSolution.cs
@@ -24,9 +24,8 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Text;
-using System.Text.RegularExpressions;
 using System.Xml;
+using NUnit.Common;
 using NUnit.Engine.Extensibility;
 
 namespace NUnit.Engine.Services.ProjectLoaders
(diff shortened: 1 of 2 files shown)


pcalin:
By the way, before merging can I use C# 6 language features? If yes, I would like to cleanup the code a bit.

CharliePoole:
Several comments, in general:

1. I don't think this is something we want to do. If you had filed an issue first, we could have discussed that and avoided your spending time working on the implementation. It's really not part of the job of a project loader to reveal inner details of how it is working, except perhaps in a debug level internal trace log. The project service already provides an IProject object with a list of the assemblies that are included in the project. The XML output contains the paths to all those projects.

2. In the design of NUnit, engine components do not write to the console. They may not even be running with a console available. The only components that are guaranteed to have a console are the console runner itself and some (but not all) platform builds of NUnitLite.

3. If we actually wanted to do this, we would have to provide an interface to return the information. But the current interface is implemented by __all__ project loaders, so they would all have to implement it as well. However, listing contained projects is not something that makes sense for any other type of project loader.

4. As discussed on an earlier commit, this one contains a number of unrelated stylistic changes. In order for a PR to be accepted, it must be reviewed and OKed by at least two committers. This takes time. If you make lots of unrelated changes, then it takes them more time and the PR may be rejected because of unrelated stuff even when it works well for it's main purpose. Keep PRs as small as is reasonable. If you are changing a file widely, then it __is__ OK to make changes to bring it into conformity with __our__ standards. We are familiar with those standards, so we can easily see what you are doing. But when you make changes to your personal preferences, it takes closer examination to be able to say "Oh this is doing the same thing as before."

6. In answer to the question: No, you should not use C# 6 features. Try to keep to C# 4 as that will allow people to build NUnit with VS2012 through 2015, SharpDevelop and MonoDevelop.

pcalin:
No worries.

PR #1418, opened 2016-04-17
Fixed IndexOutOfRangeException when --testlist file has blank line. Fixes #1390

### LINKED ISSUE #1390: --testlist with file with blank first line causes IndexOutOfRangeException
When I run a testlist file with a blank first line (and properly formatted test names on the remaining lines), NUnit crashes:

`Unhandled Exception: System.IndexOutOfRangeException: Index was outside the bounds of the array.                                                                                                                                                at NUnit.Common.CommandLineOptions.<ConfigureOptions>b__111_1(String v)                                                                                                                                                                      at NUnit.Options.Option.Invoke(OptionContext c)                                                                                                                                                                                              at NUnit.Options.OptionSet.ParseValue(String option, OptionContext c)                                                                                                                                                                        at NUnit.Options.OptionSet.Parse(String argument, OptionContext c)                                                                                                                                                                           at NUnit.Options.OptionSet.Parse(IEnumerable`1 arguments)                                                                                                                                                                                    at NUnit.ConsoleRunner.Program.Main(String[] args) `

Perhaps this case could be handled more gracefully :)

pcalin:
Fixes #1390

I'm also testing to see if I can use C# 6... :)

Cannot use C# 6 :(

### DIFF
--- a/src/Common/nunit/CommandLineOptions.cs
+++ b/src/Common/nunit/CommandLineOptions.cs
@@ -278,7 +278,7 @@ namespace NUnit.Common
                                 {
                                     var line = rdr.ReadLine().Trim();
 
-                                    if (line[0] != '#')
+                                    if (!string.IsNullOrEmpty(line) && line[0] != '#')
                                         ((List<string>)TestList).Add(line);
                                 }
                             }
--- a/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
@@ -21,6 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+using System.IO;
 using System.Reflection;
 using NUnit.Common;
 using NUnit.Options;
@@ -518,6 +519,19 @@ namespace NUnit.ConsoleRunner.Tests
 
             // Then
             Assert.AreEqual(actualTeamCity, expectedTeamCity);
+        }
+
+        #endregion
+
+        #region Testlist Option
+
+        [Test]
+        public void ShouldNotFailOnEmptyLine()
+        {
+            var testListPath = Path.Combine(TestContext.CurrentContext.TestDirectory, "TestListWithEmptyLine.tst");
+            var options = new ConsoleOptions("--testlist=" + testListPath);
+            Assert.That(options.errorMessages, Is.Empty);
+            Assert.That(options.TestList, Is.EqualTo(new[] {"AmazingTest"}));
         }
 
         #endregion
--- a/src/NUnitConsole/nunit3-console.tests/TestListWithEmptyLine.tst
+++ b/src/NUnitConsole/nunit3-console.tests/TestListWithEmptyLine.tst
@@ -1,0 +1,3 @@
+﻿
+# No funny business...
+AmazingTest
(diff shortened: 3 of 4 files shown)


CharliePoole:
Even if the tests had all passed, you can't use C# 6. Our practice is to try to keep NUnit buildable by contributors using a variety of compilers and IDEs. That has limited us to C# 4 features in the engine and console runner. The framework is generally limited to C# 3 because of the need to support certain platforms. In your case, you fell afoul of the CF build.

src/Common/nunit/CommandLineOptions.cs
```diff
@@ -274,11 +274,12 @@ protected virtual void ConfigureOptions()
...
-                                while (!rdr.EndOfStream)
+                                string line;
+                                while ( (line = rdr.ReadLine()) != null)
                                 {
-                                    var line = rdr.ReadLine().Trim();
+                                    line = line.Trim();
 
-                                    if (line[0] != '#')
+                                    if (!string.IsNullOrEmpty(line) && line[0] != '#')
                                         ((List<string>)TestList).Add(line);
```
CharliePoole: When fixing a bug, please don't make arbitrary changes to the code to reflect your preferred style. It makes it harder to review. In this case, I have to examine several changes in order to figure out that only one line really changes how the method works.

src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
```diff
@@ -522,6 +523,18 @@ public void ShouldSetTeamCityFlagAccordingToArgsAndDefauls(bool hasTeamcityInCmd
...
 
+        #region Testlist Option
+
+        [Test]
+        public void ShouldNotFailOnEmptyLine()
+        {
+            var testListPath = Path.Combine(TestContext.CurrentContext.TestDirectory, "TestListWithEmptyLine.tst");
+            var options = new ConsoleOptions("--testlist=" + testListPath);
+            Assert.That(options.errorMessages, Is.Empty);
+        }
```
CharliePoole: Good that you added a test. Why not go all the way and verify that the content of TestList?

CharliePoole:
When fixing an issue, the description or a comment should one of the two forms that github recognizes in order to automatically close the issue... fixes #1390 or fix #1390 Now it does :-)

CharliePoole:
Other than the inline comments, this looks good to merge. Needs a second committer to review it.

src/Common/nunit/CommandLineOptions.cs
```diff
@@ -274,11 +274,12 @@ protected virtual void ConfigureOptions()
                         {
                             using (var rdr = new StreamReader(fullTestListPath))
                             {
-                                while (!rdr.EndOfStream)
+                                string line;
+                                while ( (line = rdr.ReadLine()) != null)
```
ChrisMaddock: Sorry to be petty - but the spacing's inconsistent here =)

src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
ChrisMaddock: Agreed - if I was writing this, I'd want to ensure that empty strings were definitely being excluded here and not passed through to the test list, in case of future changes here.

ChrisMaddock:
Agreed with Charlie's comment on the test, and there's a stray space, but otherwise looks like a good fix to me. 👍

CharliePoole:
@pcalin This looks mergeable, but let's at least add a check to your test that the TestList is equal to `new [] { "AmazingTest" }`

rprouse:
Looks good. Merging.

PR #1433, opened 2016-04-21
Removal of some non-thread-safe code involving lazy initialization

### LINKED ISSUE #1425: Lazy initialization of OutWriter in TestResult is not thread safe
As pointed out by @jtattermusch in #1415, the lazy initialization is not thread safe and doesn't buy us much since it is always instantiated. We should change to early initialization and also review other instances of lazy initialization.

I am going to assign to 3.4 milestone to ensure it gets done for the next release.

CharliePoole:
This will fix #1425 as far as I think we should in a single issue.

This PR does the following:

1. Remove lazy initialization of TestResult.OutputWriter
2. Remove lazy initialization of Randomizer.SeedGenerator
3. Move implementation of TestResult.Children to TestSuiteResult and eliminate the lazy initialization.

@oznetmaster points out in the comments on #1425 a number of much bigger problems than were initially described. These will be documented and followed up in separate issues, for two reasons:

1. We prefer to do work in smaller increments where possible.
2. I'm about to go on vacation and this is all I have time for. :-)

### DIFF
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -154,7 +154,8 @@ namespace NUnit.Framework.Api
         {
             Settings = settings;
 
-            Randomizer.InitialSeed = GetInitialSeed(settings);
+            if (settings.Contains(PackageSettings.RandomSeed))
+                Randomizer.InitialSeed = (int)settings[PackageSettings.RandomSeed];
 
             return LoadedTest = _builder.Build(assemblyName, settings);
         }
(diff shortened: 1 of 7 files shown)


oznetmaster:
On first glance:

The Randomizers Dictionary is not thread safe,  Is that going to be an issue?

I am off to an appointment, but will be able to review all the changes later today (my time).

CharliePoole:
When you get a moment, create an issue for the Randomizers dictionary. It may not turn out to require thread-safety, but that would localize our discussion and analysis.

My first thought - without even looking at the code - is that it's probably OK now but may not be at some future point. Right now we load the tests on a single thread and I think that's the only time we use the dictionary. It should be checked though.

oznetmaster:
Looks good.  Merging.

Issue #1441, opened 2016-04-23
Linq


oznetmaster:
Currently, NUnit does not use Linq, or specifically Linq lambda expressions, because of the need to be compatible with the .NET 2.0 framework.

Has any consideration been given to including the mono or corefx Linq library in the NET 2.0 build of NUnit so that Linq lambda expressions could be used?  

It would greatly simplify some code, and increase its maintainability. And it would only be required in the NET 2.0 builds.

CharliePoole:
I added Linq support to allow a .NET 2.0 build of Mono.Addins when we were using that package. It was lots more than 2 files. Personally, I've never felt the lack of Linq, but then that's me. :-)

rprouse:
I would be okay with this. I do really miss using Linq as people have probably noticed reviewing my portable specific code. I think that Linq is one of those things that only becomes 'indispensible' after you have used it for awhile.

We seem to have a different opinion on how many files would be required though. @oznetmaster, can you confirm?

CharliePoole:
I'm OK if we can add code we don't need to maintain. As an aside, let's decide once and for all how we treat file headers for such classes.

CharliePoole:
Here is a link to the directory of files from mono I needed to add to get Mono.Addins to work under .NET 2.0 ...

https://github.com/nunit/nunit/tree/3e1c4f6ed889a75060d7d0d5fe003ed67580a719/src/NUnitEngine/mono-addins/Compatibility

It's possible there is more stuff here than what @oznetmaster is talking about adding.

oznetmaster:
That seems about right.  I will confirm.  However, many of those are very small files, one or two function, or an interface.  Enumerable is by far the major one.

oznetmaster:
I am curious.  How many people still use the .NET 2.0 build?  This is a similar question to how many still use the CF build (besides me :) )

oznetmaster:
Would the preference be to use the mono version, or the corefx version?  They both use the MIT license.  However, the corefx version is being actively maintained, and mono is rapidly converting to use corefx as its basis.

oznetmaster:
Probably the version from mono 3.12 might be the best, since  it is C# 3.0 compatible, where the corefx and mono 4.x use C# 6 in some places, and would have to be converted.

oznetmaster:
[...]

PR #1444, opened 2016-04-23
Add System.Linq (3.5) to .NET 2.0 Build

oznetmaster:
Adds a subset of Mono System.Linq to NET 2.0 build.  This is NET 3.5 compatible.

Does not add any IQueryable or related classes.

Also add HashSet, Action, and Func (all NET 3.5 compatible) to NET 2.0 build.

This is also CF compatible.


CharliePoole:
@oznetmaster We customarily use the is:idea label to mark things we are not ready to do actual work on, having not agreed to it. I added the label without commenting, so perhaps you didn't notice.

@rprouse You were favorable to having Linq available. If we want to go ahead with this, I'd be favorable to reviewing and using Linq going forward but not making change purely for the sake of change or because someone likes that style better.

CharliePoole:
If we merge this we could consider using it for the engine as well.

oznetmaster:
I think an important point is that Linq is already available in all build environments except for 2.0.  This includes CF.  I have limited this mono version to only those features of Linq that are in NET 3.5/CF so that the native Linq can be used in all other builds except 2.0.

oznetmaster:
I thought that the engine already used NET 3.5 as its basis?  I do not see any NET 2.0 specific engine builds.  Did I miss something?

oznetmaster:
@CharliePoole I did not see your is:idea label.  I assumed that since @rprouse indicated he liked the idea, it was worth going ahead and doing it and seeing what was involved.

oznetmaster:
Oh.  I now see the engine is built against the 2.0 framework all the time.

If it is desirable to use this in the engine, then perhaps going back to a separate System.Linq assembly for NET 2.0 builds makes more sense.  Same assembly could be sued for both the 2.0 framework, and the engine.

oznetmaster:
Or, the System.Linq folder could be moved to Common, and then used by both the 2.0 framework and the engine?

CharliePoole:
The engine has to use the lowest level of runtime we want to run on the desktop, because it is referenced by the agent. The agent may be launched using any runtime from 2.0 up.

This was an idea for future reference, so I don't think we should act on it right now. Once the engine is separated from the framework, it will inevitably contain duplicate source code but that's the price of making it independent. These files, which require no maintenance on our part, don't seem like a very big deal.

CharliePoole:
Again... using Common won't work once components are in separate repositories. Anyway, this is a relatively simple PR, which does one thing. I didn't intend to complicate it by mentioning something that might happen in the future.

oznetmaster:
I agree.  The PR is simple, and should involve no ongoing support.  Duplicating the small number of source files in another project is a minor thing.

I would think that if it is done here, doing it in the engine would provide a consistent programming environment for ongoing development.

It would be helpful if a decision could be reached on this so that if it is going to be merged, it is done soon.

This would allow all current development to start using these capabilities (which from my point of view is very desirable).

CharliePoole:
I'm OK with it but even though that makes two of us, I'd like to hear from Rob again before we merge.

oznetmaster:
@CharliePoole I included the ExtensionAttribute.cs source in this PR because it was in your list, and because it had been in the Net20Compatibility.cs which I removed.  Is this still necessary.

oznetmaster:
Whoops.  I also made it public.  It had not been public in Net20Compatibility.cs.

oznetmaster:
I guess it is necessary, since taking it out breaks things :(

I see it used to be public, and then was made not public.  Not sure why, since it is public in all other environments.

oznetmaster:
Considering @rprouse extensive use of Linq in the portable build, I doubt he is going to have a problem with this :)

CharliePoole:
I think making ExtensionAttribute non-public was one thing I attempted in
order to get around the notorious bug that it can cause for users who
reference our assembly from 3.5 tests. However, they should just use the
3.5 build!

oznetmaster:
Seems like a reasonable requirement :)

oznetmaster:
I had not thought about this, but should we add the appropriate mono System.Linq tests into the framework compatibility tests for the 2.0 build?

CharliePoole:
For me, that's only something we would need if we started maintaining the files ourselves. I think it's better if we can just freeze them and use as is.

oznetmaster:
Of course, the header issue is not resolved.  I just left the mono headers unchanged.

CharliePoole:
That's the way I  would resolve it too.

oznetmaster:
Once again, CI failed on a build that was unchanged in any way from the last time when it passed :(

How can I restart a CI build when this happens?

CharliePoole:
There's a button at the top of the display. If you don't see it, you may not be signed in to Travis. I restarted it, but it may fail again. It's an equality comparison within a tolerance.

oznetmaster:
I have no idea how to sign into Travis.  Not even sure I can sign into Travis?

oznetmaster:
Is this a problem with the unit tests for nunit framework?  Does some test have its tolerance set too low?

oznetmaster:
Well, it passed this time anyway :)

CharliePoole:
For Travis, go to travis-ci.org and click on Sign in with Github.

rprouse:
I am okay with moving forward to this. It will be one less gotcha for new developers who want to contribute and in my experience, if used properly can make much cleaner and readable code. Going forward, I want that to be on the team's minds as they are reviewing PRs. I have seen some pretty crazy linq queries :smile: 

For a coding standard, I don't want to favour either query syntax or fluent syntax. I think both have their place and are better for certain problems.

I agree with treating these as static files, not importing the tests and leaving the original copyright. I also don't think we should pull them into the engine yet, nor should they be in common. I would like to see the engine and the framework separated out to different repositories in 3.4 or 3.6 so they can start being released on different timelines.

I will finish reviewing and merge.

CharliePoole:
This fixed #1441

PR #1445, opened 2016-04-24
Issue 1423

### LINKED ISSUE #1423: SingleThreaded attribute should raise an error if a thread is required
The new `SingleThreaded` attribute would be more useful if it raised an error if a conflicting attribute that required a thread was used on a child test. Examples of such attributes are `Timeout` and the various thread related attributes, but it should be more generic and look at the requires thread property.

CharliePoole:
This PR changes the behavior of tests run under a fixture with the SingleThreaded attribute and fixes #1423. In the 3.2.1 release, any test attributes that would normally cause a separate thread to be used were silently ignored. After this fix, they result in an error.

In working this, I ran into the problem that three test utility classes were duplicated in both nunit.framework.tests and nunitlite.tests. I moved the classes to Common.


oznetmaster:
This all builds and works fine in my CF build as well.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -341,14 +341,15 @@ internal ITestListener Listener
...
         /// </summary>
-        internal IWorkItemDispatcher Dispatcher
+        public IWorkItemDispatcher Dispatcher
         {
             get
             {
-                if (_dispatcher == null)
-                    _dispatcher = new SimpleWorkItemDispatcher();
+                //if (_dispatcher == null)
+                //    _dispatcher = new SimpleWorkItemDispatcher();
```
rprouse: Why is this commented out now? If it isn't needed, should it be deleted?

CharliePoole:
Yes, it should be deleted. This BTW removes one instance of lazy initialization listed in #1425

Issue #1434, opened 2016-04-21
 TestResult class needs to be thread-safe


CharliePoole:
As any test finishes, on its own thread, it updates its own test result and adds that result to the parent Result. This is an obvious point where different threads can come into conflict, so the Children property of TestSuiteResult needs to be thread-safe. For the last child test of a parent, additional updates are done, which can ripple up the result hierarchy all the way to the top.

Additionally, the SetResult method should be examined since it is called on other occasions in addition to a call from within AddResult.

If it's reasonably easy, the whole class could be made thread-safe, but the above items seem to represent the minimum.

I'm calling this a bug, even though we have not had problems reported, since it's obvious that threading conflicts can occur. This should be fixed before we attempt to run test cases in parallel within a fixture, since that would increase the probability of a collision significantly.

CharliePoole:
@oznetmaster If it's OK with you, let's move on to this issue, which seems to be the most critical one, continuing our discussion from #1425 here.

My participation is getting more and more limited as my trip approaches, but I'll try to keep responding where I can. @rprouse will have to take over my role generally.

The volume of usage of AddResult, as explained above, is rather significant, since every test but the topmost Assembly suite has to add it's result to the parent. Adding a child result changes the collection of child results, the resultstate and possibly message and the various totals that are maintained (passed, failed, etc.). So this method definitely needs a lock of it's own and would need it even if the collection itself were thread-safe.

Can you take a look and tell me what other info I can  give you so as to drive a decision?

oznetmaster:
I am not sure the entire class needs a lock, if thread safe collections are used.

It appears that only AddResult and SetResult are an issue.

AddResult needs to be thread safe, for sure.  How that is accomplished depends on the decision about concurrent collections.

SetResult might need a lock, but since it only sets properties, and that setting is not dependent on anything except the parameters to SetResult, it may not be.  The only possible problem would be an inconsistent set of property values (ResultState, Message, StackTrace) if two different threads called SetResult at the same time.  With a lock, the values from the latest call would always be used.

Does AddToXml need to be thread safe?  I am not sure about its usage.

It does not seem that any other method or property has a thread safety issue.

oznetmaster:
TestSuiteResult has its own thread safety issues, independent of TestResult which it derives from.  Specifically, the various properties which are count accumulators.  But those can be easily handled by the Interlocked class.

Does TestSuiteResult require thread safety as well?  Are there other classes derived from TestResult that might also have related thread safety issues?

[...]

PR #1452, opened 2016-04-25
Fix issue #1434

oznetmaster:
This PR adds the mono ConcurrentQueue class to all builds for which it
is not available.  It uses the ConcurrentQueue and Interlocked classes
to make the TestResult class (and its derivatives) thread safe.

It also adds an Interlocked.Add method to the CF build, since it is not available in CF.


oznetmaster:
@rprouse I tried to use the ConcurrentQueue class that is supposed to be available in all portable .NET frameworks in the Portable and SL5 builds, but was unable to locate/reference it.  I therefore included the mono version in those builds as well.  If the version that is supposed to be in the .NET frameworks can be referenced, I will remove the mono version from those build.

src/NUnitFramework/framework/Compatibility/InterlockedExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+		///     atomic operation.
+		/// </summary>
+		/// <param name="target">
+		///     A variable containing the first value to be added. The sum of the two values is stored in
+		///     <paramref name="target" />.
+		/// </param>
+		/// <param name="value">The value to be added to the integer at <paramref name="target" />.</param>
+		/// <returns>The new value stored at <paramref name="target" />.</returns>
+		public static Int32 Add (ref Int32 target, Int32 value)
+			{
```
rprouse: Minor formatting issues. You are using tabs, not spaces and your brackets are indented an extra level. Please fix.

src/NUnitFramework/framework/Compatibility/System.Collections.Concurrent/ConcurrentQueue.cs
```diff
@@ -0,0 +1,413 @@
...
+//
+//
+
+using System;
+using System.Threading;
+using System.Collections;
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+
+namespace System.Collections.Concurrent
```
rprouse: I am worried about putting this in the same namespace as the original version. It is not available in the portable framework because it is not available on all platforms that the PCL targets, but it is available on many of them. I expect that if we leave this in the same namespace then we will get the same multiple definition warnings that we got with 2.0.

I think we need to move this to an NUnit.Compatibility namespace.

src/NUnitFramework/framework/Compatibility/System.Collections.Concurrent/IProducerConsumerCollection.cs
```diff
@@ -0,0 +1,90 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace System.Collections.Concurrent
```
rprouse: Same comment about the namespace.

src/NUnitFramework/framework/Constraints/EqualConstraint.cs
```diff
@@ -167,6 +167,7 @@ public EqualConstraint Within(object amount)
             return this;
         }
 
+#if !NETCF
```
rprouse: This seems unrelated to this PR. Why are you #ifdef'ing this out on CF?

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
```diff
@@ -52,17 +52,22 @@ public class TestWorker
...
         /// <summary>
         /// Construct a new TestWorker.
         /// </summary>
         /// <param name="queue">The queue from which to pull work items</param>
         /// <param name="name">The name of this worker</param>
         /// <param name="apartmentState">The apartment state to use for running tests</param>
-        public TestWorker(WorkItemQueue queue, string name
-#if !NETCF
-                          , ApartmentState apartmentState
+        public TestWorker(WorkItemQueue queue, string name, ApartmentState apartmentState)
```
rprouse: This is easier to read :+1:

src/NUnitFramework/framework/Internal/Results/TestCaseResult.cs
```diff
@@ -87,9 +88,9 @@ public override bool HasChildren
         /// <summary>
         /// Gets the collection of child results.
         /// </summary>
-        public override IList<ITestResult> Children
+        public override ConcurrentQueue<ITestResult> Children
         {
-            get { return new ITestResult[0]; }
+            get { return EmptyChildQueue; }
```
rprouse: Maybe overly cautious, but new that we create a single instance, there is nothing preventing callers of this code from adding children. I doubt we do, but should we clear the queue before returning? What do you think?

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -52,11 +54,23 @@ public abstract class TestResult : ITestResult
...
+
+        //        static Logger log = InternalTrace.GetLogger("TestResult");
 
         private StringBuilder _output = new StringBuilder();
         private double _duration;
 
+        /// <summary>
+        /// Aggregate assertion count
+        /// </summary>
+        protected int InternalAssertCount;
```
rprouse: This is named as a property and I don't see why you needed to add backing fields to this or the ResultState. I would prefer if we left it the way it is unless there is a reason.

rprouse: Sorry, I see the reason for ResultState below. I don't see the need for AssertCount. At a minimum, the backing field should be named correctly.

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
```diff
@@ -112,46 +117,53 @@ public override IList<ITestResult> Children
...
                         break;
                 }
+            } while (!stateSet);
 
-            AssertCount += result.AssertCount;
-            _passCount += result.PassCount;
-            _failCount += result.FailCount;
-            _skipCount += result.SkipCount;
-            _inconclusiveCount += result.InconclusiveCount;
+            Interlocked.Add (ref InternalAssertCount, result.AssertCount);
```
rprouse: This is why you named it like a property and added the backing field. This makes sense now. Sorry, ignore the previous comments. :+1:

rprouse:
For the most part this looks good except for

- The formatting which I assume came from the original file? Maybe we should leave it if it is that way in the original source. Odd that the second file is fine though.
- The namespace for the queue worries me
- Ignore my comments about the backing fields, sorry.
- Curious about the #if !NETCF's added

oznetmaster:
- The formatting which I assume came from the original file? Yes, it did.  Mono files are not always formatted consistently.

- If we change the namespace, then those builds which already have the System.Collections.Concurrent namespace will not compile, unless we start conditioning the includes.

- I was getting an error on the xml comment in the CF build, so I fixed it at the same time.

rprouse:
I am okay with everything except the namespace, but I think we need to condition the includes. There are other places I had to do this and hopefully it will be fairly limited?

oznetmaster:
Does the namespace have to be changed only for the portable build then?  For all other builds, it is seemingly not an issue.  That way it can be conditioned only with #if PORTABLE

src/NUnitFramework/framework/Compatibility/System.Collections.Concurrent/ConcurrentQueue.cs
CharliePoole: Since Portable doesn't support parallel, do we need this at all?

src/NUnitFramework/framework/Interfaces/ITestResult.cs
```diff
@@ -157,7 +157,7 @@ bool HasChildren
         /// <summary>
         /// Gets the the collection of child results.
         /// </summary>
-        System.Collections.Generic.IList<ITestResult> Children
+        System.Collections.Concurrent.ConcurrentQueue<ITestResult> Children
```
CharliePoole: I don't like allowing an implementation detail to leak out through the interface. We should examine how children is used. Since we can no longer use IList, we should look at how Children is used and pick another System interface if possible. If not, perhaps we should create an ITestResultChildren interface and implement it differentially according to the platform and use.

src/NUnitFramework/framework/Internal/Results/TestCaseResult.cs
CharliePoole: In the original code, I considered a single instance for efficiency but decided to avoid it because this member of a TestCaseResult should never be accessed. In fact, for a while, I left it null to see if any error arose. None did. It may be acceptable to throw a NotImplementedException here.

Alternatively, if we put ITestResult.Children back to being some simple interface, we could return a null implementation of it. In that case, it wouldn't matter whether we used a single instance or a new instance.

I did try working with two different interfaces but the code is really much better if every test returns an ITestResult.

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
```diff
@@ -112,46 +117,53 @@ public override IList<ITestResult> Children
...
+            Interlocked.Add (ref _passCount, result.PassCount);
+            Interlocked.Add (ref _failCount, result.FailCount);
+            Interlocked.Add (ref _skipCount, result.SkipCount);
+            Interlocked.Add (ref _inconclusiveCount, result.InconclusiveCount);
         }
```
CharliePoole: It seems to me that this method has become quite complex, while still leaving room for race conditions if two threads are in it at the same time. Why wouldn't we simply lock on the suite result while in this method?

src/NUnitFramework/framework/nunit.framework-portable.csproj
```diff
@@ -186,6 +186,9 @@
     <Compile Include="Attributes\ValuesAttribute.cs" />
     <Compile Include="Attributes\ValueSourceAttribute.cs" />
     <Compile Include="CollectionAssert.cs" />
+    <Compile Include="Compatibility\System.Collections.Concurrent\ConcurrentQueue.cs" />
+    <Compile Include="Compatibility\System.Collections.Concurrent\IProducerConsumerCollection.cs" />
+    <Compile Include="Compatibility\System.Collections\CollectionDebuggerView.cs" />
```
CharliePoole: Not sure why we do this in portable, which doesn't even support parallel execution.

src/NUnitFramework/framework/nunit.framework-sl-5.0.csproj
```diff
@@ -202,6 +202,9 @@
     <Compile Include="Attributes\ValuesAttribute.cs" />
     <Compile Include="Attributes\ValueSourceAttribute.cs" />
     <Compile Include="CollectionAssert.cs" />
+    <Compile Include="Compatibility\System.Collections.Concurrent\ConcurrentQueue.cs" />
+    <Compile Include="Compatibility\System.Collections.Concurrent\IProducerConsumerCollection.cs" />
+    <Compile Include="Compatibility\System.Collections\CollectionDebuggerView.cs" />
```
CharliePoole: Same as for portable: no parallel execution here.

src/NUnitFramework/framework/Compatibility/System.Collections.Concurrent/ConcurrentQueue.cs
rprouse: @CharliePoole I intended to ask the same question. I thought I had typed it, but I guess I forgot :smile: 

I agree with @oznetmaster that it would complicate our interfaces to return two different types, but, I believe that our code only ever adds to the Children and then enumerates over them and adding is only ever done through `AddResult`. Could we drop these two compatibility classes and have our `Children` just be `IEnumerable<ITestResult>` with a backing field of `ConcurrentQueue` on builds that support PARALLEL and the original `List<>` on PCL/SL?

This would allow us to drop these two compatibility classes that aren't really needed anyway since conflicts are not needed.

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
rprouse: If we just had a lock around this method, do we even need the concurrent queue? If Children were just IEnumerable, then consumers wouldn't be able to add children and we could enforce everything here greatly simplifying this entire PR.

CharliePoole: I believe that's correct. AddResult is not a member of ITestResult but exposing Children as an IList has always allowed for the possibility of adding directly to it. We do need to know if the enumerable is empty, for which we currently examine the Count, but we could get around that.

rprouse: Rather than have consumers of the Api look at the count, we can just use the HasChildren call.

src/NUnitFramework/framework/Compatibility/System.Collections.Concurrent/ConcurrentQueue.cs
CharliePoole: Nevertheless, I would argue that the interface should not use the implementation class, as it is now doing.

rprouse: @CharliePoole I agree.

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
oznetmaster: Currently, the AddResult method is lockless, which is highly desirable.  No chance of deadlocks and much higher performance.

I am curious where you see a possible race condition?  As far as I can tell, it is now totally thread safe.

Lockless code is always more complicated.  Most contemporary multi-threaded code strives to be lockless, hence the reason for the Concurrent classes and the Interlocked class.

Using these classes, I have written large multi-threaded libraries and programs without a single lock/Monitor in them.  No deadlocks or priority inversion issues.

It is always simpler to use lock/Monitor, if you do not mind the consequences of doing so.

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
```diff
@@ -112,48 +132,59 @@ public override IList<ITestResult> Children
         /// <param name="result">The result to be added</param>
         public virtual void AddResult(ITestResult result)
         {
-            Children.Add(result);
+#if PARALLEL
+            _children.Enqueue(result);
+#else
+            _children.Add(result);
+#endif
```
oznetmaster: Not sure what to do about this, since it breaks the virtualization of Children.  Can't use Children anymore, because it is an IEnumerable, which has no Add or equivalent methods.

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
CharliePoole: Looking it over, I can see that there is currently no race condition. AddResult performs the following seven operations (each) atomically:
1. Actually adds the child result to the collection.
2. Sets the overall result, based on the child result
3-7. Increments each of the counters separately.

All these things happen independently and the test in SelectIf ensures that the code will be retried if another thread performs step 1 before step 2 can be finished.

However, the state of the result is invalid between the steps. Between 1 and 2, the overall resultstate is incorrect. After 2 but before 7, the totals don't match. If anyone accesses the result during that time, it's state wil be invalid. Admittedly, this is exactly the case with the original code.

The key problems I see are

1. The code is (as you say) complex and may be hard for some contributors to follow. A simple lock would be easier. If we keep it as it is, I would want us to take great care in reviewing any changes.

2. The performance advantage seems questionable to me given how little AddResult actually does and the limitations of how and when it is called. I assume that the ConcurrentQueue is locking already.

3. When I talk about a lock, I don't necessarily mean the obvious lock/monitor implementation. This is all local to a process and AppDomain and could be optimized in a few ways.

4. The steps involved in AddResult are never supposed to happen independently, so there is no particular need for them to be separately thread-safe. The overall design is based on the idea that results should be added atomically.

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
CharliePoole: Sorry, I don't understand your comment. It seems natural to use the _children field internally while external clients use the Children property on the interface. External clients don't have or need access to adding/enqueuing results.

Is this commit an experiment?

oznetmaster: I fixed the issue.  However, since since Children can be overriden in a derived class (as is the case in TestCaseResult), Children for the current instance may be any IEnumerable<ITestResult>.

oznetmaster: I do not know what you mean about the commit being an experiment?

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
oznetmaster: ConcurrentQueue, like all the Concurrent classes, is lockless.  That is why it is used.

oznetmaster: Unless every property of ITestResult was locked individually, there is no way to provide consistent results to a reading thread.  Even if AddResult used a lock to make the entire method atomic, another thread could still access any of the counts, and they would not be consistent.

If that is the real issue, then each property should have a read lock, and AddResult should have a write lock.  That can be done using ReaderWriterLock or ReaderWriterLockSlim.

oznetmaster: Between 1 and 2, the overall ResultState is not invalid.  It is exactly as it would be if it was read prior to step 1, so it is not a problem in and of itself.  After 2 and before 7 was adressed above.

CharliePoole: What I mean is that we may have (for example) a passing fixture that contains a failing test for a brief time. That's an invalid state. We have these invalid transitional states in objects all the time and they (usually) cause no problem. The question is whether it causes a problem here, with multiple threads potentially accessing the parent result.

As a thought experiment, I'm imagining a fixture result that has so far succeeded. Now two tests run in parallel. One of them fails and the other is ignored. The parent result should end up as a failure, no matter how the two different test thread executions are interleaved. 

As a worst case, assume that the two threads are both reaching step 2 at approximately the same time. Both of them discover that the parent result is passing. They are competing to change that result - in one case to failure in the other to ignrored. As I read the code, if the ignored test gets there first, the result will be ignored. The failing test will discover that the result is changed and try again, finally making the result a failure. If the failing test gets there first, it will be the ignored test that tries again and it will back off not needing to change anything.

Is that how you see the particular scenario playing out as well?

CharliePoole: I wasn't aware that the concurrent collections are lockless - only that they are thread-safe.

CharliePoole: So we could make a threadsafe List, using a lock, but not (easily) a lockless threadsafe List.

CharliePoole: I would say that TestResult only needs to be made generally threadsafe if we allow users to access the context of the parent fixture from their own context. So far, I have resisted that and now I have another reason to do so.

oznetmaster:
Whoops.. these CI tests may deadlock :(

oznetmaster:
While debugging the deadlock, I examined the threads and their activity.  The tests are running with 8 worker threads.  These threads all seemed to be correctly waiting on locks, and taking locks, as required.

Given how often the various test threads were going through the locks and taking the locks, there really does seem to be some cross thread sensitivity here.

Even without a set of specific tests for these changes, having 8 worker threads while running the tests is seeming a very good test itself.  It found the deadlock (typing error) which resulted in ALL worker threads waiting for the lock.

oznetmaster:
Still waiting for a review of this.  It is a complete rewrite addressing all of the concerns raised.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -90,7 +117,26 @@ public TestResult(ITest test)
...
-        public ResultState ResultState { get; private set; }
+        public ResultState ResultState
+        {
+            get
+            {
+#if PARALLEL
+                RwLock.EnterReadLock();
+#endif
+                try
+                    {
```
rprouse: Formatting issues

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -194,7 +309,7 @@ public string Output
 
 #endregion
 
-        #region IXmlNodeBuilder Members
+#region IXmlNodeBuilder Members
```
rprouse: Regions got unindented. I find VS2015 annoyingly does this when I use #if's

src/NUnitFramework/framework/Compatibility/System.Collections.Concurrent/ConcurrentQueue.cs
```diff
@@ -0,0 +1,413 @@
...
+    /// <typeparam name="T">Specifies the type of elements in the queue.</typeparam>
+    /// <remarks>
+    /// All public  and protected members of <see cref="ConcurrentQueue{T}"/> are thread-safe and may be used
+    /// concurrently from multiple threads.
+    /// </remarks>
+#if !NETCF
+	[System.Diagnostics.DebuggerDisplay ("Count={Count}")]
+	[System.Diagnostics.DebuggerTypeProxy (typeof (CollectionDebuggerView<>))]
+#endif
+	public class ConcurrentQueue<T> : IProducerConsumerCollection<T>, IEnumerable<T>, ICollection,
```
rprouse: Can't we remove this since we are using List for non-parallel?

src/NUnitFramework/framework/nunit.framework-netcf-3.5.csproj
```diff
@@ -200,6 +200,9 @@
     <Compile Include="Attributes\ValuesAttribute.cs" />
     <Compile Include="Attributes\ValueSourceAttribute.cs" />
     <Compile Include="CollectionAssert.cs" />
+    <Compile Include="Compatibility\ReaderWriterLockSlim.cs" />
+    <Compile Include="Compatibility\System.Collections.Concurrent\ConcurrentQueue.cs" />
```
rprouse: I see, you are using ConcurrentQueue for CF.

src/NUnitFramework/framework/nunit.framework-sl-5.0.csproj
rprouse: We should remove it from CF though.

rprouse:
Thanks for pinging us. Looks good except for the formatting issues and regions. Plus we should remove the ConcurrentQueue from SL since it isn't used.

oznetmaster:
All changes requested by @rprouse  completed.

oznetmaster:
Travis CI keeps failing with:

The command "sudo apt-get install -qq mono-complete mono-vbnc fsharp" failed and exited with 100 during .

Did someone break the Travis CI environment?

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
```diff
@@ -112,46 +194,66 @@ public override IList<ITestResult> Children
...
+            if (childrenAsConcurrentQueue != null)
+                childrenAsConcurrentQueue.Enqueue(result);
+            else
+#endif
+            {
+                var childrenAsIList = Children as IList<ITestResult>;
+                if (childrenAsIList != null)
+                    childrenAsIList.Add(result);
+                else
+                    throw new NotSupportedException("cannot add results to Children");
```
CharliePoole: Is this logic left over from an earlier version? In this version _children is always a queue in the parallel build and always an IList in the non-parallel build. So I would expect something like...

```
#if PARALLEL
    _children.Enqueue(result);
#else
    _children.Add(result);
#endif

CharliePoole:
I have only the one inline comment, regarding what looks like unneeded logic around the use of a queue versus a list in AddResult. Looks good otherwise.

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
oznetmaster: The problem is that Children is a virtual property, and can be overriden by derived classes.  TestCaseResult is one example.  Therefore, _children cannot be used, but Children has to be.  Since Children can be anything that implements IEnumerable, it has to be done this way,

oznetmaster: I realize that TestCaseResult overrides TestResult, and not TestSuiteResult.  I was only using it as an example.

The alternative is to make TestSuiteResult a sealed class, which would prevent derivation, and allow the use of _children like you suggested.

CharliePoole: But we're in TestSuiteResult here and _children can't be anything. Its actual type is known to us. If PARALLEL is defined, the first cast will always succeed. If not, the second cast will always succeed.

The field _children doesn't even exist in the base or in TestCaseResult but is private here and the AddResult method only exists here. What you say would be true if we were talking about the Children of the result that is being added, but here we are dealing with `this` result and we know the implementation type.

CharliePoole: Bear in mind we are talking about an internal implementation class, albeit only enforced by convention. I don't envision there being further derivations, since the distinction is between a single result and a composite result.

As a matter of philosophy, we have not used `sealed` in the past even on classes not designed for inheritance. Some people advocate doing that. Our philosophy has been "If you mess with our internals, you take your chances." If you and @rprouse perfer to seal classes that present risks if incorrectly extended, I can go along.

oznetmaster:
I responded to @CharliePoole inline comment, with two choices, leave as is, or make TestSuiteResult a sealed class.

CharliePoole:
See my further inline comments... I'm afraid I may not be understanding your concerns. Perhaps if you could give an example of how a derived class could run into a problem I'd catch on.

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
oznetmaster: If there are never going to be any further derivation, the seal the class.  If it is not sealed, and someone were to derive from it in the future, then the use of _children could fail.

I think that sealing any class for which derivation is neither a bad thing, or not supported, makes sense.

It does eliminate the need for future debugging when someone does derive from it, and it fails or messes things up.

oznetmaster: If there was no intent to ever derive from TestSuiteResult, then why was AddResult made virtual?

That woudl only be done if there were plans to derive from the class.

oznetmaster:
See my further inline comment.

If a derived class, for instance,  used a List<ITestResult> for Children, then your code would fail.

rprouse:
I think the extra check or sealed would be fine. You have the check, so let's go for it. As @CharliePoole says, it isn't a problem, but I do also agree with @oznetmaster that we may as well be safe. Besides, I don't think it is worth all the comments :smile:

CharliePoole:
OK, I understand now. And as I said I'm OK with sealing the class. Bear in mind that we have a separate issue that aims to make many of these classes in the Internal namespace actually internal.

oznetmaster:
If it is ok the way it is, then should we merge it?

rprouse:
Yes, merging. I was just waiting for CI to pass. I made some minor formatting changes for a few misplaced braces. It was easier to fix than go back and forth :smile: 

Thanks.

PR #1453, opened 2016-04-25
Add xml comments to all public classes, properties and methods

oznetmaster:
Although this adds all the needed xml comments, it only actually fills
in a few of them.  There are over 1000 to do :(  More will be filled in
over time.

This does eliminate the large number of warnings about no xml
documentation in the 2.0 build, however.


CharliePoole:
I agree that the warnings are annoying, but this starts us modifying all those mono files we said we wouldn't modify. If we need to do something, I would rather see a simple one-line pragma at the start of each file to suppress the warning. Even better, why not suppress it for the entire 2.0 project? @rprouse what's your preference?

oznetmaster:
If the comments are not added, then anyone using these classes under the 2.0 build will have no intellisense for these classes.

I thought this was a good thing to do.

oznetmaster:
I also added them to the ConcurrentQueue class, for the same reason.

rprouse:
You did a lot of work here, but I am also uncomfortable bringing in that many changes to these files. I wanted to keep them static and the same as the originals so that we can just drop in the new files if we ever want to take changes. Doing this will require us to manually merge changes or redo the comments. I would much rather see a global ignore on the 2.0 projects.

If anyone needs the intellisense when working on this, they just need to change the dropdown at the top of the editor for VS to one of the newer projects. We seldom, if ever, work on the 2.0 project alone.

image

Sorry, but I am going to close this. I don't want to take on the maintenance going forward. We should fix the warnings though. Can you add the ignore to the projects?

PR #1454, opened 2016-04-25
Different default SL install loc for 32-bit Windows

### LINKED ISSUE #1450: Silverlight detection doesn't work when building on 32-bit OS
The fix made to build.cake so that Silverlight would be skipped if not installed is not working correctly on my 32-bit Windows system, causing the build and test to always be skipped. There is no ProgramFiles(x86) environment variable on this system, only ProgramFiles.

ChrisMaddock:
Fixes #1450 

@CharliePoole - I don't have a 32-bit OS to hand to test this on, but I can't see any reason why this shouldn't work. If you get a chance to try it out before you're away, that would be great!


CharliePoole:
Works for me. Nifty use of ??

PR #1456, opened 2016-04-25
Broken badge fixed in README.md

elv1s42:
NuGet package downloads count badge from [Shields.io](http://shields.io/) is not working now (if I am not mistaking, because of [this](https://github.com/badges/shields/issues/655) issue).

Another way to display Nuget Version and Downloads count is to use [CI-BuildStats](https://github.com/dustinmoris/CI-BuildStats) badge.


elv1s42:
I think it is a very small addition, so there is no need to create an issue for this.

rprouse:
Thanks, I didn't even notice that is was broken. I have reviewed and this looks good.

rprouse:
Yes, no need for an issue. This won't be in the release notes, so we don't need to track it further.

PR #1468, opened 2016-04-29
Makes ParallelWorkItemDispatcher and WorkItemQueue lockless threadsafe

### LINKED ISSUE #1435: Parallel queue creation needs to be thread-safe
Duplicating comment made on #1425 ...

Starting with documenting our use of queue creation in ParallelWorkItemDispatcher.

There can be up to four queues but they may not all be needed for a given test run. For example, if there are no tests requiring the STA, then ParalellelSTAQueue and NonParallelSTAQueue never need to be created.

The dispatcher starts with no queues. When the first item is dispatched, it decdes what queue to add it to based on it's characteristics. As a result of referencing that Queue through the property, it gets created. All other queues are created as the existing CompositeWorkItems create and dispatch child work items. For example, if the child item is not parallelizable, then the NonParallelQueue will get created as part of dispatching it.

There is a risk of two different threads trying to dispatch an item and create a new queue at the same time. Let's say we are running the first shift (parallel) with four worker threads. Two of them discover a child item that requires the STA queue. Both will be trying to create it.

We probably want to retain the lazy initialization here. In fact, we may want to expand it's use in the future by allowing the number of workers to change dynamically. I think this means that ParallelWorkItemDispatcher must be at least partially thread-safe - that is, at least insofar as it's creation of queues.

The queues themselves are already threadsafe, using a home-brew approach. We could decide to clean this up and use a concurrent queue from the mono code as @oznetmaster has suggested. This should be discussed before we commit to it.

I don't see other methods that need to be made thread-safe. The Enqueue method would be an obvious choice, but it hands the job to the queue itself, so I'm thinking we are OK there.

@oznetmaster Would you review this and tell me what you think of it in terms of
1. Is this the info that you feel is needed to work on it?
2. Do my conclusions make sense?

oznetmaster:
Fixes #1435 

Makes both ParallelWorkItemDispatcher and WorkItemQueue lockless threadsafe.

It adds the mono SpinWait, Lazy and LazyThreadSafetyMode to the 2.0, 3.5, and CF builds.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
@@ -67,17 +67,17 @@ namespace NUnit.Framework.Internal.Execution
         private readonly ConcurrentQueue<WorkItem> _innerQueue = new ConcurrentQueue<WorkItem>();
 
         /* This event is used solely for the purpose of having an optimized sleep cycle when
-		 * we have to wait on an external event (Add or Remove for instance)
-		 */
+	     * we have to wait on an external event (Add or Remove for instance)
+	     */
         private readonly ManualResetEventSlim _mreAdd = new ManualResetEventSlim(false);
 
         /* The whole idea is to use these two values in a transactional
-		 * way to track and manage the actual data inside the underlying lock-free collection
-		 * instead of directly working with it or using external locking.
-		 *
-		 * They are manipulated with CAS and are guaranteed to increase over time and use
-		 * of the instance thus preventing ABA problems.
-		 */
+         * way to track and manage the actual data inside the underlying lock-free collection
+         * instead of directly working with it or using external locking.
+         *
+         * They are manipulated with CAS and are guaranteed to increase over time and use
+         * of the instance thus preventing ABA problems.
+         */
         private int _addId = int.MinValue;
         private int _removeId = int.MinValue;
 
--- a/src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
@@ -23,8 +23,12 @@
 
 #if PARALLEL
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading;
+#if NET_2_0 || NET_3_5 || NETCF
+using ManualResetEventSlim = System.Threading.ManualResetEvent;
+#endif
 
 namespace NUnit.Framework.Internal.Execution
 {
@@ -56,14 +60,15 @@ namespace NUnit.Framework.Internal.Execution
     /// </summary>
     public class WorkItemQueue
     {
+        private const int spinCount = 5;
+
         private Logger log = InternalTrace.GetLogger("WorkItemQueue");
 
-        private Queue<WorkItem> _innerQueue = new Queue<WorkItem>();
-        private object _syncRoot = new object();
-#if NETCF
-        private ManualResetEvent _syncEvent = new ManualResetEvent(false);
-        private int _waitCount = 0;
-#endif
+        private readonly ConcurrentQueue<WorkItem> _innerQueue = new ConcurrentQueue<WorkItem>();
+        private readonly ManualResetEventSlim _mreAdd = new ManualResetEventSlim(false);
+
+        private int _addId = int.MinValue;
+        private int _removeId = int.MinValue;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="WorkItemQueue"/> class.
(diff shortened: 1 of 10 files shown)


rprouse:
I've had a quick look at this and it looks good so far. Code reviewing on my phone sucks though. I will take a closer look this weekend.

src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
```diff
@@ -117,22 +138,29 @@ public bool IsEmpty
         /// <param name="work">The WorkItem to process</param>
         public void Enqueue(WorkItem work)
         {
-            lock (_syncRoot)
+            do
```
rprouse: Can you explain your reasoning behind the changes here and in Dequeue? What is wrong with protecting the queue/dequeue with a lock and why is your spinning approach better?

oznetmaster: There is a large amount of published information on this issue, including MSDN.  Lockless thread safety is the current design goal for most contemporary programming.  It is higher performance, deadlock free, and much easier to debug.  The entire reason for the concurrent collections is to provide for lock free thread safe collections.  All of the regular collections can be make thread safe by using locks, but that can be fraught with problems and performance issues.  Using locks on enqueu and dequeu then requires that locks be used everywhere that any of the properties and methods of the class be used.  By using spinning and CAS ([https://en.wikipedia.org/wiki/Compare-and-swap]), there is no need for that.

oznetmaster: Spinning is only used in Dequeue, and only for a maximum of five  tries before the thread waits on the event.  This avoids any transitions to the underlying OS kernel for any transitional or short lived usage of the class by another thread.  This means that if another thread is simple using Dequeue to get an entry, in most cases no transition to the OS kernel will occur.  Only if the queue stays empty, or the state stays paused, will the event be waited on.  Using a lock means that a transition to the kernel will occur every time the method is called/exited, even if no other thread is using the class/method.

src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
```diff
@@ -56,14 +60,15 @@ public enum WorkItemQueueState
...
-        private object _syncRoot = new object();
-#if NETCF
-        private ManualResetEvent _syncEvent = new ManualResetEvent(false);
-        private int _waitCount = 0;
-#endif
+        private readonly ConcurrentQueue<WorkItem> _innerQueue = new ConcurrentQueue<WorkItem>();
+        private readonly ManualResetEventSlim _mreAdd = new ManualResetEventSlim(false);
+
+        private int _addId = int.MinValue;
+        private int _removeId = int.MinValue;
```
CharliePoole: Would you add a comment somewhere explaining the purpose of these ids in uniquely identifying each operation?

src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
CharliePoole: I agree with the general statement and I think it fits well with these low overhead operations. @oznetmaster are you using any sort of backing delay before retry? It's hard to fully examine the code on my phone.

oznetmaster: On the 4.0 and later builds, this code uses ManualResetEventSlim for the event.  That class also uses CAS and spin waiting internally to avoid kernel calls wherever possible.  I originally added that mono class to the pre 4.0 builds, but it requires a bunch of support classes as well, and it did not seem to worth doing so for those legacy builds.  I still could, however.  For instance, my CF build does use that class, since my CF environment has it already.

CharliePoole:
I'd like to see us redo the pump queue in the same way. Another issue however.

src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
oznetmaster: There is no need for a backing delay, since CAS automatically waits for the "simultaneous" operations to complete (the atomicity of the operation).

src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
oznetmaster: These ids are only a mechanism for avoiding accessing the ConcurrentQueue itself until it is appropriate to do so.  Accessing the ConcurrentQueue is relatively expensive, while a simple CAS operation is very cheap.  They have no other use then that, and could be eliminated at a performance cost.

For instance, calling TryDequeue and having it fail is significantly more expensive then simple doing the CAS on the ids.

The ids simply indicate that it is a good time to access the ConcurrentQueue.  The ConcurrentQueue could still end up doing its own CAS retries internally, or even be unable to do the operation, but it is much less likely.  Since there are no locks, another thread could still "slip through the cracks".  This does no harm, because the ConcurrentQueue itself is thread safe.

oznetmaster:
The pump queue is on my list :)

oznetmaster:
I am just waiting for this to get merged, so that the support classes are part of the master branch (SpinWait and Lazy) before I tackle other thread safety issues, both already posted as issues (like CurrentFramework and CurrentPlatform), and also those mentioned in our discussions but not posted yet (like the pump queue)

CharliePoole:
Yes, I understood that. I meant a code comment, since it might confuse some future contributors.

As I'm "absent" merging is up to @rprouse, but it's looking like a good direction to me.

rprouse:
It looks good to me to, but like @CharliePoole, I would like a few code comments. Coming at this PR code, it was not immediately obvious what those ids were for and why you took the approach you did. I also don't think this pattern is common enough yet to be quickly recognized.

Can you add short comments to the beginning of the enqueue and dequeue explaining the approach. It doesn't need to be as detailed as your explanation above, just 2 or 3 lines. Then add a short comment for the id's as @CharliePoole suggested. Just enough to point developers in the right direction if the code confuses them. Maybe even link to this PR?

Thanks for doing this, it is a great improvement.

oznetmaster:
I already added the comments that @CharliePoole wanted - about 14 hours ago. :(

rprouse:
@oznetmaster, sorry I didn't look a the diff again because the checkins weren't showing up in the timeline. Looks like there was a glitch in GitHub, the changes are now showing as "New changes since you last viewed". Weird since they are older than my comment.

I will give it a quick last review and merge.

oznetmaster:
Well, I added a lot more - now the most commented class in NUnit :)

PR #1471, opened 2016-05-01
Add thread safety and clean up classes

### LINKED ISSUE #1436: CurrentFramework and Current Platform need to be more thread-safe
These are static methods of RuntimeFramework and OsPlatform respectively. Both are used by PlatformHelper and only by PlatformHelper, which means they are not needed unless the user tests include a Platform attribute or property on the tests. They can end up doing a bunch of work checking the registry etc., so it would be good to maintain the lazy initialization but make them thread-safe.

Probably a good use for `Lazy<T>`

oznetmaster:
Fixes #1436

### DIFF
--- a/src/NUnitFramework/framework/Internal/PlatformHelper.cs
+++ b/src/NUnitFramework/framework/Internal/PlatformHelper.cs
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2007-2014 Charlie Poole
+// Copyright (c) 2007-2016 Charlie Poole
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
@@ -22,6 +22,7 @@
 // ***********************************************************************
 #if !PORTABLE
 using System;
+using System.Linq;
 
 namespace NUnit.Framework.Internal
 {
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/framework/Internal/PlatformHelper.cs
```diff
@@ -87,13 +88,9 @@ public PlatformHelper( OSPlatform os, RuntimeFramework rt )
...
         public bool IsPlatformSupported( string[] platforms )
-        {
-            foreach( string platform in platforms )
-                if ( IsPlatformSupported( platform ) )
-                    return true;
-
-            return false;
-        }
+            {
+            return platforms.Any (IsPlatformSupported);
```
rprouse: Linq for the win! :wink: 

You can't tell me this isn't more readable and an improvement...

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
```diff
@@ -154,7 +234,7 @@ private void InitFromFrameworkVersion(Version version)
                                 switch (version.Minor)
```
rprouse: You didn't make the change and I don't want to do the change within this PR, but next time we are in this code we should clean up these nested switch statements. I can see we want to support future versions of the CF framework, but double nested If's with one case each?

rprouse:
Looks good, but you need to watch your formatting. I fixed a number of cases where the brackets were indented an extra level and the space was before the ( in method calls, not after. Minor, but we should be consistent.

For example,

```C#
private static bool IsRuntimeTypeName(string name)
    {
    return TypeHelper.GetEnumNames (typeof(RuntimeType)).Any (item => item.ToLower () == name.ToLower ());
    }
```

rprouse:
CI passed, looks good, merging. Thanks. :+1:

PR #1472, opened 2016-05-01
Make EventQueue and EventPump thread-safe and lock-less

### LINKED ISSUE #1470: Make EventQueue and associated classes lock-less and thread safe
This is another phase of the thread safety modifications.

This would apply the same techniques used in WorkItemQueue and ParallelItemDispatcher to the EventQueue and its associated classes

oznetmaster:
Fixes #1470

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2006 Charlie Poole
+// Copyright (c) 2006-2016 Charlie Poole
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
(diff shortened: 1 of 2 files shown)


rprouse:
Looks good

CharliePoole:
I realize that this change merely converts what was already there but we should come back to this and examine the use or non-use of "synchronous" events. The whole idea of isolating test events through this queue is that they should be handled asynchronously and independently. I think the code for synchronous events may be left over from earlier implementations and may no longer be used.

oznetmaster:
I looked at that, but had no context as to why it was there, so just left it as it was.

I agree that revisiting it would be a good thing.  Perhaps by someone like @CharliePoole  who has an appropriate context to do so.

CharliePoole:
I can't figure out how to add an issue on my phone. Would you add one and assign to me?

FWIW I've never been sure about these events. I think that code was added to avoid conflicts in text output events, which occurred when the thread handling the event did a console write, which was then redirected to the queue itself. This is no longer possible.

Issue #1432, opened 2016-04-21
NUnit hangs when reporting to TeamCity


martinskuta:
Recently we have updated our build to be run by Cake (C# make) script which runs our tests through dotCover from JetBrains on TeamCity agent. We have noticed that NUnit automagically realizes it is run on team city agent and reports to the server, which is cool, but it seems that it can make the nunit-console-runner to hang. I think it is related to to that, because after inspecting the process dump there are two worker threads doing Socket.Recieve and other two doing TeamCityServiceMessagePublisher.RegisterMessage. 

Some more facts:

- The hang is very random, 90% of the time our build finishes fine. 
- When it is stuck it is stuck untill it is killed (24h)
- NUnit 3.2.1, ran via dotCover on WindowsServer 2012R2 x64, CLR 4.6.1055
- we have four dlls containing tests, these are run in parallel

![process explorer](https://dl.dropboxusercontent.com/u/6537800/nunit/process_explorer_scr_when_stuck.png "process explorer screenshot")

[full process dump](https://dl.dropboxusercontent.com/u/6537800/nunit/nunit3-console.dmp)

Let me know if you need more info.

rprouse:
Are you using any of the parallel features in NUnit other than parallel assemblies?

martinskuta:
> Are you using any of the parallel features in NUnit other than parallel assemblies?
No, just parallel assemblies. Maybe what might be relevant is that we have some async tests `public async Task` and some that requires STA, but they run sequentially.

martinskuta:
Today we experienced another build stuck while running tests. 75% of CPU (3 of 4 cpu cores) have been full at work! Looking at the process dump it is same issue again.. 
![image](https://cloud.githubusercontent.com/assets/14070352/14743021/9390e84a-089f-11e6-8448-56667829a2b4.png)

Yesterday it was 50% CPU and two threads doing Socket.Recieve and one doing TeamCityServiceMessagePublisher.RegisterMessage and one doing TryFindPartnerId. You can see on the screenshot today it were three threads doing Socket.Recieve, two doing TeamCityServiceMessagePublisher.RegisterMessage and one doing TryPartnerId ... so RegisterMessage and TryFindPartnerId there seems to lie the issue. Seems like the RegisterMessage and TryFindPartnerId are deadlocking ?

It is kind of annoying since we cannot find the root cause. I will start looking at NUnit code when won't be in release week. Someone some idea?

martinskuta:
[...]

Issue #1476, opened 2016-05-04
Examine need for "synchronous" events in event queue


oznetmaster:
Transferred from #1472.

Repeating @CharliePoole:

We should come back to this and examine the use or non-use of "synchronous" events. The whole idea of isolating test events through this queue is that they should be handled asynchronously and independently. I think the code for synchronous events may be left over from earlier implementations and may no longer be used.

CharliePoole:
Here is the result of my review. @oznetmaster especially, please comment.

In NUnit V2 the three start events (RunStarted, SuiteStarted, TestStarted) are synchronous. That means it's not even possible to enqueue (for example) a TestStarted event unitl the SuiteStarted event has been pulled from the queue by the pumping thread.

The need for this arose when some users created TestListener addins that contained Console.Write statements. Such a statement would create a TestOutput event, which would then be enqueued. In order to avoid deadlock, it was necessary that the processing thread, which had dequeued the event, continue to hold a lock. That allowed it to enqueue further events.

Currently, we don't have output events, so this would not be a problem. I could remove the code now and I'm confident it won't cause a problem. However, I'm about to put the TestOutput event back in the system in order to deal with issue #1139, continuous test output.

@oznetmaster To my understanding, the changes you made to the queue should prevent such a deadlock from occuring. Consequently, I think I can remove the manual reset event and the whole notion of synchronous events from the code. To be safe, we should also add some tests where the dequeuing thread tries to enqueue a new event. Would you agree?

@NikolayPianikov You proposed replacing the entire queue structure with your ThreadScheduler code. How do you think it would perform in the situation described? If we do this, it will be at a later stage - I'm a big believer in taking small steps. :-)

oznetmaster:
I believe that is the case.  I never understood why they were there in the first place, but I left them untouched.

CharliePoole:
In that case, I'm removing it. High pri because it impacts other stuff being worked on.

PR #1486, opened 2016-05-10
#1432 NUnit hangs when reporting to TeamCity - DO NOT MERGE !

NikolayPianikov:
Lock object has been added to ensure that all events "OnTestEvent" in "ITestEventListener" will be handled synchronously

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Runners/TestEventDispatcherTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Runners/TestEventDispatcherTests.cs
@@ -1,0 +1,103 @@
+﻿// ***********************************************************************
+// Copyright (c) 2015 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using NUnit.Engine.Runners;
+using NUnit.Framework;
+
+namespace NUnit.Engine.Tests.Runners
+{
+    using System.Collections.Generic;
+
+    [TestFixture]
+    public class TestEventDispatcherTests
+    {
+        private TestScheduler _testScheduler;
+        private TestEventListener _testEventListener1;
+        private TestEventListener _testEventListener2;
+
+        [SetUp]
+        public void SetUp()
+        {
+            _testScheduler = new TestScheduler();
+            _testEventListener1 = new TestEventListener();
+            _testEventListener2 = new TestEventListener();
+        }
+
+        [Test]
+        public void ShouldNotSendEventsDirectrlyToListeners()
+        {
+            // Given
+            var dispatcher = CreateInstance();
+            dispatcher.Listeners.Add(_testEventListener1);
+            dispatcher.Listeners.Add(_testEventListener2);
+
+            // When       
+            dispatcher.OnTestEvent("abc");
+
+            // Then
+            Assert.AreEqual(1, _testScheduler.QueueSize);
+            Assert.AreEqual(0, _testEventListener1.Repots.Count);
+            Assert.AreEqual(0, _testEventListener2.Repots.Count);
+        }
+
+        [Test]
+        public void ShouldSendEventsViaScheduler()
+        {
+            // Given
+            var dispatcher = CreateInstance();
+            dispatcher.Listeners.Add(_testEventListener1);
+            dispatcher.Listeners.Add(_testEventListener2);
+
+            // When       
+            dispatcher.OnTestEvent("abc");
+            _testScheduler.Run();
+
+            // Then
+            Assert.AreEqual(0, _testScheduler.QueueSize);
+            Assert.AreEqual(1, _testEventListener1.Repots.Count);
+            Assert.AreEqual("abc", _testEventListener1.Repots[0]);
+            Assert.AreEqual(1, _testEventListener2.Repots.Count);
+            Assert.AreEqual("abc", _testEventListener2.Repots[0]);
+        }
+
+        private TestEventDispatcher CreateInstance()
+        {
+            return new TestEventDispatcher(_testScheduler);
+        }
+
+        private class TestEventListener : ITestEventListener
+        {
+            public IList<string> Repots { get; set; }
+
+            public TestEventListener()
+            {
+                Repots = new List<string>();
+            }
+
+            public void OnTestEvent(string report)
+            {
+                Repots.Add(report);
+            }
+        }
+    }
+}
--- a/src/NUnitEngine/nunit.engine.tests/Runners/TestScheduler.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Runners/TestScheduler.cs
@@ -1,5 +1,5 @@
 ﻿// ***********************************************************************
-// Copyright (c) 2016 Charlie Poole
+// Copyright (c) 2015 Charlie Poole
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
@@ -21,39 +21,40 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-using System;
 using System.Collections.Generic;
+using NUnit.Engine.Runners;
 
-namespace NUnit.Engine.Runners
-{
-    /// <summary>
-    /// TestEventDispatcher is used to send test events to a number of listeners
-    /// </summary>
-    public class TestEventDispatcher : MarshalByRefObject, ITestEventListener
+namespace NUnit.Engine.Tests.Runners
+{    
+
+    internal class TestScheduler: IScheduler
     {
-        private readonly IScheduler _scheduler;
+        private readonly Queue<Action> _actions = new Queue<Action>();
 
-        public TestEventDispatcher(IScheduler scheduler)
+        public int QueueSize
         {
-            if (scheduler == null) throw new ArgumentNullException("scheduler");
-
-            _scheduler = scheduler;
-            Listeners = new List<ITestEventListener>();
+            get
+            {
+                return _actions.Count;
+            }
         }
 
-        public IList<ITestEventListener>Listeners { get; private set; }
 
-        public void OnTestEvent(string report)
+        public void Schedule(Action action)
         {
-            _scheduler.Schedule(() =>
-                {
-                    foreach (var listener in Listeners) listener.OnTestEvent(report);
-                });
+            _actions.Enqueue(action);
         }
 
-        public override object InitializeLifetimeService()
+        public void Dispose()
+        {            
+        }
+
+        public void Run()
         {
-            return null;
+            while (_actions.Count > 0)
+            {
+                _actions.Dequeue()();
+            }          
         }
     }
 }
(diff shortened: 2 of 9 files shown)


rprouse:
@oznetmaster can you give this a review based on the multi-threading work you have been doing. I expect you will be concerned with the lock :wink:

CharliePoole:
Let's also look at whether this is the right place for a fix. Console runner is just one client. Should this be an engine issue?

NikolayPianikov:
I can implement another way if it is needed. I could use "SynchronizationAttribute" instead. The lock works quite simple and fast. Could you add your recommendations? 

You have to take in account that some calls are from current domain:
*`nunit3-console.exe!NUnit.ConsoleRunner.TestEventHandler.OnTestEvent`
*`nunit.engine.dll!NUnit.Engine.Runners.MasterTestRunner.RunTests`
*`nunit.engine.dll!NUnit.Engine.Runners.MasterTestRunner.NUnit.Engine.ITestRunner.Run`	
*`nunit3-console.exe!NUnit.ConsoleRunner.ConsoleRunner.RunTests`
*`nunit3-console.exe!NUnit.ConsoleRunner.ConsoleRunner.Execute`
*`nunit3-console.exe!NUnit.ConsoleRunner.Program.Main`

and another are via Remoting:
*`nunit3-console.exe!NUnit.ConsoleRunner.TestEventHandler.OnTestEvent`
*`mscorlib.dll!System.Runtime.Remoting.Messaging.StackBuilderSink.SyncProcessMessage`

CharliePoole:
My thinking is that it shouldn't be possible for a client runner to cause the test to block. Would you agree with that statement?

NikolayPianikov:
How we can achieve this goal when we are having more then one agent? Should we use a queue and some dispatching thread?

NikolayPianikov:
All output operations are synchronous. From my point of view we should not put off
intention to write something to a console to a queue. It is not scalable and reliable. And on the other hand the Lock statement in the .net is actually lock-free and very lightweight (without OS critical section) at the first microseconds.

CharliePoole:
Currently, neither the framework nor engine performs output operations except for the internal trace, which is only used for debugging nunit. Rather, they generate events. It's clients that may interpret some events in a way to produce output.

All I'm saying is that we need to prevent clients from messing up the engine and framework by errors and delays. I think we are pretty well protected WRT one process running at a time but not so well with parallel processes.

src/NUnitConsole/nunit3-console/nunit3-console.csproj
```diff
@@ -25,6 +25,7 @@
     <WarningLevel>4</WarningLevel>
     <Externalconsole>true</Externalconsole>
     <Commandlineparameters>net-4.5/nunit.framework.tests.dll --wait --inprocess</Commandlineparameters>
+    <LangVersion>3</LangVersion>
```
NikolayPianikov: Could we add this limitation on all our projects to guarantee a correct compilation?

CharliePoole:
@NikolayPianikov This is still changing. Could you add a Do Not Merge notice when you create a PR early before you are actually done? For this one please post a comment when it's ready for review, so we're not spinning our wheels.

My general comment remains the same as before... I feel you're putting too much in the console runner with the implication that every runner needs similar code.

CharliePoole:
@oznetmaster I'm thinking you missed @rprouse 's initial request for a review.

I'd like to add to that a further request... can you suggest a way to avoid the need for console runner to deal with this problem. That is, can't we guarantee that all messages arrive sequentially from the engine in the first place? 

In the framework, we use a queue to ensure that events arrive in sequence and that they don't execute on the test thread itself. For the engine, we need to forward those framework events .and add a few more - not many but they can come from different processes. It seems to me that we could put somethng in the engine itself that would provide the necessary guarantees so that we don't have to implement synchronization in every client.

rprouse:
@oznetmaster if you get a chance, can you give this a review? If not, let me know and we will move forward. Thanks.

NikolayPianikov:
@CharliePoole I've done PR as a concept. I add my tests and some changes related to code style if the suggested concept is fine.

CharliePoole:
@NikolayPianikov Have you seen my earlier comments? My thought is that the logic needed to serialize events doesn't belong in the console runner.

There are many runners that use the engine, so it seems to me that the serialization of events should take place in the engine. Then, each runner can simply use the events without any overhead.

What is your opinion of that? Also @oznetmaster what do you think?

NikolayPianikov:
From my point of view we could add some events' dispatcher (_EventDispatcher_). 
Its main responsibility is to send events to the handlers (implementing _ITestEventListener_) consequentially in one thread. 
Another one ability is to provide stability against exceptions and to exclude long delays in events' handling. _EventDispatcher_ could be a part of a shared assembly, for example _nunit.engine.api_. 
I can suggest something if you are interested.

CharliePoole:
@NikolayPianikof I don't think the actual implementation should be in the
api assembly, since we try to keep that as minimal as possible. But it will
be accessed through that assembly. Let me outline what I think is needed,
starting with the origin of (most) events in the framework.

How it currently works...

1. The framework serializes events by use of an EventQueue and a dedicated
thread that reads that queue and sends the events back to the event
listener. @Oznetmaster recently converted the queue to a non-locking
mechanism. Each instance of the framework driver is therefore guaranteed to
receive events on the same thread, in the proper order.

Note: Use of a dedicated thread also prevents the CallContext, which may
contain serialized objects the engine does not know about, from leaking
outward. In fact, this was the original reason for creating the queue, back
before we ever had parallel test execution. The context issue is therefore
not present at higher levels.

2. In general, the driver and the chain of runners that call it do not
intercept the framework messages. Hence, the calling runner may receive
these messages in parallel, on multiple threads. In general, there is one
thread in play for each executing assembly.

3. MasterTestRunner creates a TestEventDispatcher to receive all messages
and re-issue them to all listeners. Otherwise, only the console runner
would receive events. It's this mechanism that allows extensions that
handle events to receive them - the teamcity message processor, for
example. TestEventDispatcher currently sends out the events using the same
thread on which it receives them.

What I propose...

TestEventDispatcher should use a non-locking mechanism to ensure that only
one message is being processed at a time. This will remove the need for the
console runner, the teamcity extension or any future runners or extensions
to deal with messages that are out of sync.

The risk, of course, is that a runner or extension will throw an exception,
hang or simply take too long in processing the message. Since this is not
user test code that we are calling, I think we can reasonably allow the
execution to fail on an exception. WRT time delays, I suggest we wait until
they happen. Any such delay is essentially an error in the runner or
extension, which should be queuing any events that require longer
processing or starting a background thread to handle them. If the problem
occurs, there are several ways we can deal with it and we will have more
information to use in selecting the best way to do it if we wait.

Charlie

NikolayPianikov:
@CharliePoole  As I see TestEventDispatcher is a part of agent. MasterTestRunner creates TestEventDispatcher before running tests on the agent side. It also registers proxy of TestEventDispatcher as NUnit.ConsoleRunner.TestEventHandler and uses it via .net remoting from the agent's process.
And we can have several process which send events to NUnit.ConsoleRunner.TestEventHandler and there are no ways to achive our goals if TestEventDispatcher is on the agent's side only.
Do you suggest to use TestEventDispatcher on the console side? Should we move it to some place before?

CharliePoole:
@NikolayPianikov TestEventDispatcher is used by MasterTestRunner, which is part of the engine assembly. The engine assembly is loaded by both the console runner and the agent. I was thinking of the copy in the runner process, but forgot it was also loaded by the agent.

Probably, if we implement this, it should be done on the client runner side only. The agent could tell the engine to exclude the event handler when it initializes it. We do this already for other engine features that don't make sense in the agent, like starting new processes. It would require some changes to the code to do, of course.

If we don't do this, then the serialization would take place redundantly in both the agent and the client process.

NikolayPianikov:
I like #1522. What about the non-locking mechanism. Should I close my PR?
I looked at EventQueue/EventPump and ParallelTaskWorkerPool.

The approach using EventQueue/EventPump  - looks a bit strange (see line ```_mreAdd.Wait(500)``` in EventQueue) and over-complicated.
We could achieve this goal by ```new ThreadScheduler(1);```

ParallelTaskWorkerPool - is not suitable at all because we should invoke ```WaitAll();``` and ```Start();```, thus actually we should block the thread.

Also both approaches are hard to test via unit testing.

NikolayPianikov:
It is just my vision:

- I would like to have a simple way to run something on the sinlge/multiple thread(s) from the privat/shared pool.

- I would not like to have a business logic and a code related to threads in a one class.

- I would like to have a simple way to put something to a queue for processing in a dedicated thread.

- I would like to have ability to test any class by the unit tests in a simple way - in the one testing thread.

Let's start with IServiceLocator:
IServiceLocator - is quite good but not enough. Each class could get it and an create instance of any class across it. As for me - it's not so good. We look to the constructor and we don't see any dependencies except IServiceLocator. But actually this class could have a bunch of dependencies. That is why IoC can help us here. Also IServiceLocator does not allow to resolve instances by the name (for example when we want to resolve several implementations of the one interface, or when instances can be shared or private).

Next is multi-threading:
We could have one approach to run some activity on the thread(s). We could inject this ability to the any classes by the common way. Also it could solves another issues like non-locking queues.

A simple example:

IScheduler - main responsibility is to run action on some thread (current/dedicated/from the pool/UI thread ...):
```
    interface IScheduler: IDisposable
    {        
        IDisposable Schedule(Action action);
    }
```

ThreadScheduler - implementation for the cases dedicated/from the pool

IoC container:
```
        interface IRegistry
	{
		IRegistry Register<T>(Func<T> factory, string name = "");

		IRegistry Register<TArg, T>(Func<TArg, T> factory, string name = "");
	}
```

```
        interface IResolver
	{
		T Resolve<T>(string name = "");

		T Resolve<TArg, T>(TArg arg, string name = "");
	}
```

```
    interface IContainer : IResolver, IRegistry
    {
    }
```

```
        interface IConfiguration
	{
		IContainer Apply(IContainer container);
	}
```

Configuration of schedulers  for IoC container:

```
public class ReactiveContainerConfiguration: IConfiguration
    {
        public IContainer Apply(IContainer container)
        {
            if (container == null) throw new ArgumentNullException(nameof(container));

            container = container.Resolve<IContainer>(typeof(ReactiveContainerConfiguration).Name);

            var taskFactory = new TaskFactory();
            var sharedSingleThreadScheduler = new Lazy<IScheduler>(() => CreateSingleThreadScheduler(taskFactory));
            var sharedMultiThreadScheduler = new Lazy<IScheduler>(() => CreateMultiThreadScheduler(taskFactory));

            container
                .Register(() => sharedSingleThreadScheduler.Value, WellknownSchedulers.SharedSingleThread)
                .Register(() => CreateSingleThreadScheduler(taskFactory), WellknownSchedulers.PrivateSingleThread)
                .Register(() => sharedMultiThreadScheduler.Value, WellknownSchedulers.SharedMultiThread)
                .Register(() => CreateMultiThreadScheduler(taskFactory), WellknownSchedulers.PrivateMultiThread);

            return container;
        }

        private static Scheduler CreateMultiThreadScheduler(TaskFactory taskFactory)
        {
            if (taskFactory == null) throw new ArgumentNullException(nameof(taskFactory));

            return new Scheduler(taskFactory, Environment.ProcessorCount);
        }

        private static IScheduler CreateSingleThreadScheduler(TaskFactory taskFactory)
        {
            if (taskFactory == null) throw new ArgumentNullException(nameof(taskFactory));

            return new Scheduler(taskFactory, 1);
        }
    }
```

The names of schedulers for different needs:

```
public static class WellknownSchedulers
    {
        public static readonly string SharedSingleThread = "SharedSingleThread";
        public static readonly string PrivateSingleThread = "PrivateSingleThread";
        public static readonly string SharedMultiThread = "SharedMultiThread";
        public static readonly string PrivateMultiThread = "PrivateMultiThread";
    }
```

The usage:
```
internal class ExplorerContainerConfiguration: IConfiguration
    {
        public IContainer Apply(IContainer container)
        {
            if (container == null) throw new ArgumentNullException(nameof(container));            

            container
                .Register<ITool>(() => new Tool(
                    container.Resolve<IScheduler>(WellknownSchedulers.PrivateSingleThread),
                    container.Resolve<IEventAggregator>()));

            return container;
        }
    }
```

CharliePoole:
@NikolayPianikov WRT non-locking mechanisms, I decided that the C# lock statement was the simplest possible thing. If the wait is short, it won't lock at all.

One key thing - none of the listeners called are considered true "user code" Of course, a user may write and install a listener extension, but that extension becomes part of the system. So we don't have to be quite so robust as when calling a user test. If an extension is bad - or even a third-party runner - it seems OK to me that the application should crash, so long as the stack trace gives good info to pinpoint what caused it. The bad extension should be de-installed until it's fixed.

That's not the same as when we call into code that is part of the test or the sut. Then, we have to isolate NUnit itself from any errors so it can give an adequate report.

In the same way, if a user test listener causes NUnit to slow down because it holds the lock for too long, it's the fault of the listener, which should be fixed.

Anyway, those are all the reasons why I went with lock instead of some sort of spinwait.

Yes, I think your PR can be closed if this solves the problem. Have you tested it yet?

I need to think a bit more about your longer comment...

CharliePoole:
@NikolayPianikov WRT EventQueue being overly complex...

Yes, now that I look more closely, the recent changes have made it more complex. I think it's because @oznetmaster tried to make it as general as possible, with multiple writers and readers. However, the background of the class is that it's intended as a way to get multiple events handled by a single thread, without blocking the writers. That's why the queue exists. We can clean this up in another pass after I have dealt with issue #1476.

I like the idea of using ThreadScheduler rather than a queue in the framework provided we could make it work with all builds, including Silverlight, Portable and Compact. @oznetmaster What do you think of this?

CharliePoole:
@NikolayPianikov WRT ParallelTaskWorkerPool...

It always seemed a bit too complicated to me as well. However, there seems to be little problem with the thread blocking, since it merely waits for a separate process to finish.

Still, an improvement would be welcome. Would you use ThreadScheduler here as well? Obviously, since this is built using .NET 2.0, you would need to define the Action delegate.

rprouse:
@CharliePoole can we close this now that the extension is in an extension? Was this fixed there?

CharliePoole:
I thought we might be able to repurpose some of @NikolayPianikov 's code in the engine - particularly the ThreadScheduler class that uses delegates. It seems like a great lightweight way to ensure that events are only issued serially. Since it's from a fork, I think everything just disappears if we close it.

I'll take a look later today and grab the code I want to save before closing it.

CharliePoole:
@oznetmaster You never got back to me about this. Any thoughts?

Issue #1483, opened 2016-05-08
Remoting exceptions during test execution


vgriph:
In the process of migrating from NUnit 2.6 to 3.2.1 I repeatedly run into errors with tests that fail to report the result during the execution, and the internal trace files of the test run fills up with remoting exceptions similar to this on:

> System.Runtime.Remoting.RemotingException: Object '/4f1134d2_f841_416d_b1a9_a0a78447fc4a/upwuhw+18ooxwlbxdcaourel_1.rem' has been disconnected or does not exist at the server.
> 
> Server stack trace: 
>    at System.Runtime.Remoting.Channels.ChannelServices.CheckDisconnectedOrCreateWellKnownObject(IMessage msg)
>    at System.Runtime.Remoting.Channels.ChannelServices.DispatchMessage(IServerChannelSinkStack sinkStack, IMessage msg, IMessage& replyMsg)
> 
> Exception rethrown at [0]: 
>    at System.Runtime.Remoting.Proxies.RealProxy.HandleReturnMessage(IMessage reqMsg, IMessage retMsg)
>    at System.Runtime.Remoting.Proxies.RealProxy.PrivateInvoke(MessageData& msgData, Int32 type)
>    at NUnit.Engine.ITestEventListener.OnTestEvent(String report)
>    at NUnit.Engine.RunTestsCallbackHandler.ReportProgress(String state)
>    at NUnit.Engine.CallbackHandler.RaiseCallbackEvent(String eventArgument)
>    at System.Runtime.Remoting.Messaging.StackBuilderSink._PrivateProcessMessage(IntPtr md, Object[] args, Object server, Object[]& outArgs)
>    at System.Runtime.Remoting.Messaging.StackBuilderSink.SyncProcessMessage(IMessage msg)
> 
> Exception rethrown at [1]: 
>    at System.Runtime.Remoting.Proxies.RealProxy.HandleReturnMessage(IMessage reqMsg, IMessage retMsg)
>    at System.Runtime.Remoting.Proxies.RealProxy.PrivateInvoke(MessageData& msgData, Int32 type)
>    at System.Web.UI.ICallbackEventHandler.RaiseCallbackEvent(String eventArgument)
>    at NUnit.Framework.Internal.TestProgressReporter.TestFinished(ITestResult result)

[...]

PR #1489, opened 2016-05-10
 #1483 Remoting exceptions during test execution

vgriph:
Give the TestEventDispatcher infinite life time to prevent it from being
disconnected during a long running test run. Somewhere between 2 and 3 minutes
test running time was enough for all test results after that test to fail to report back.

I've not added a unit test for this change, since the time required to reproduce the error was 3 minutes. I've verified that this test case now works.

    [TestFixture]
    public class LongRunningTest`
    {
        public static IEnumerable<TimeSpan> SleepTimes
        {
            get
            {
                yield return TimeSpan.FromSeconds(120); // Succeeded before
                yield return TimeSpan.FromSeconds(180); // Failed before this change
            }
        }

        [Test]
        [TestCaseSource(nameof(SleepTimes))]
        public void ShallBeAbleToExecuteForSeconds(TimeSpan executionTime)
        {
            Thread.Sleep(executionTime);
            TestContext.WriteLine($"Slept for {executionTime}");
        }
    }

### DIFF
--- a/src/NUnitEngine/nunit.engine/Runners/TestEventDispatcher.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/TestEventDispatcher.cs
@@ -43,5 +43,10 @@ namespace NUnit.Engine.Runners
             foreach (var listener in Listeners)
                 listener.OnTestEvent(report);
         }
+
+        public override object InitializeLifetimeService()
+        {
+            return null;
+        }
     }
 }


CharliePoole:
Looks good to me. In some of our projects we use a derived base class for this butch this is the only instance in the console runner, your approach us simpler.

rprouse:
Looks good to me too. Merging.

rprouse:
Thanks for the help. :+1:

ChrisMaddock:
Removed from milestone after adding the original issue instead.

ivanball:
Good morning. Where can I download an nunit console with this issue fixed?

bjharo:
@ivanball It looks like this was included in version 3.4 per the [release notes](https://github.com/nunit/docs/wiki/Release-Notes).

CharliePoole:
That's correct. @ivanball We don't generally show the release milestone in a PR. However if you click the link to the issue this fixes (#1483) it does indicate the milestone.

PR #1494, opened 2016-05-12
NUnit 2 test results report ParameterizedMethod but should be ParameterizedTest

imakowski:


src/NUnitEngine/Addins/nunit-v2-result-writer/NUnit2XmlResultWriter.cs
```diff
@@ -216,7 +216,7 @@ private void StartTestElement(XmlNode result)
             {
                 var suiteType = result.GetAttribute("type");
                 xmlWriter.WriteStartElement("test-suite");
-                xmlWriter.WriteAttributeString("type", suiteType);
+                xmlWriter.WriteAttributeString("type", suiteType == "ParameterizedMethod" ? "ParameterizedTest" : suiteType);
                 string nameAttr = suiteType == "Assembly" || suiteType == "Project" ? "fullname" : "name";
```
CharliePoole: if we're taking the trouble to fix this, maybe someone should look at other values we generate for the attribute. However, it doesn't seem like a high priority so I'd be OK with merging as is too.

rprouse:
Looks good to me. I would prefer not to go back and forth on pull requests, so I have merged. If anything else is reported, we can fix it then. Seems like fairly low priority if people aren't running into issues.

PR #1497, opened 2016-05-15
Apply same fix to nunitlite as to nunit3-console

CharliePoole:
Made the equivalent one-line change to nunitlite as was made to nunit3-console.

NOTE: the two codebases are separate because the console builds it's output from the nunit3 XML while nunitlite uses the internal object references in the framework. For a small loss in efficiency, we could make the nunitlite version work the same as the console version. This might be advisable.


rprouse:
Thanks @CharliePoole. I didn't realize that code was duplicated.

PR #1508, opened 2016-05-20
NullReferenceException with null arguments in TestFixtureAttribute

### LINKED ISSUE #1507: NullReferenceException when null arguments are used in TestFixtureAttribute
My team is in the process of trying to convert our unit tests from 2.6.4 to 3, and we ran into several errors while running the console runner (just using `nunit3-console.exe tests.dll`), similar to the following:

```
13) Invalid : Tests.Controls.Query.when_execute_is_clicked

 at NUnit.Framework.Internal.Builders.NUnitTestFixtureBuilder.CheckTestFixtureIsValid(TestFixture fixture)
 at NUnit.Framework.Internal.Builders.NUnitTestFixtureBuilder.BuildFrom(ITypeInfo typeInfo, ITestFixtureData testFixtureData)
 at NUnit.Framework.TestFixtureAttribute.<BuildFrom>d__48.MoveNext()
```

This corresponds to the following test fixture and its constructor:
```
[TestFixture(null, "foo", "foo")]
[TestFixture("", "foo", "foo")]
[TestFixture("bar", "foo", "bar")]
public class when_execute_is_clicked : RefreshTestBase
{
    private string _selectedText;
    private string _query;
    private string _issuedQuery;

    public when_execute_is_clicked(string selectedText, string query, string issuedQuery)
    {
        _selectedText = selectedText;
        _query = query;
        _issuedQuery = issuedQuery;
    }
```

Judging by the code I see in master, it would appear the NRE is coming from [NUnitTestFixtureBuilder, line 229](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/Builders/NUnitTestFixtureBuilder.cs#L229).

rprouse:
Fixes #1507

Turned out to not be the easy fix I expected, but not too bad. I will explain the changes inline with the code.

### DIFF
--- a/src/Common/nunit/Compatibility/ReflectionExtensions.cs
+++ b/src/Common/nunit/Compatibility/ReflectionExtensions.cs
@@ -24,10 +24,7 @@
 using System;
 using System.Collections.Generic;
 using System.Reflection;
-
-#if PORTABLE
 using System.Linq;
-#endif
 
 #if NUNIT_ENGINE
 namespace NUnit.Engine.Compatibility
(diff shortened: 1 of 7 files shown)


src/Common/nunit/Compatibility/ReflectionExtensions.cs
```diff
@@ -521,4 +474,72 @@ public static IList<Type> GetTypes(this Assembly asm)
...
+    /// </summary>
+    public static class AdditionalTypeExtensions
+    {
+        /// <summary>
+        /// Determines if the given <see cref="Type"/> array is castable/matches the <see cref="ParameterInfo"/> array.
+        /// </summary>
+        /// <param name="pinfos"></param>
+        /// <param name="ptypes"></param>
+        /// <returns></returns>
+        public static bool ParametersMatch(this ParameterInfo[] pinfos, Type[] ptypes)
```
rprouse: Most of this is just moved from within the PORTABLE block.

src/Common/nunit/Compatibility/ReflectionExtensions.cs
```diff
@@ -521,4 +474,72 @@ public static IList<Type> GetTypes(this Assembly asm)
...
+        /// <param name="from"></param>
+        /// <returns></returns>
+        public static bool IsCastableFrom(this Type to, Type from)
+        {
+            if (to.IsAssignableFrom(from))
+                return true;
+            
+            // Look for the marker that indicates from was null
+            if (from == typeof(NUnitNullType) && to.GetTypeInfo().IsClass)
+                return true;
```
rprouse: This if block is the only change, checking if we found null in the parameter list by looking for the marker class since there is no typeof(null)

src/NUnitFramework/framework/Internal/Builders/NUnitTestFixtureBuilder.cs
```diff
@@ -215,18 +215,7 @@ private static void CheckTestFixtureIsValid(TestFixture fixture)
...
-
-                // Note: This could be done more simply using
-                // Type.EmptyTypes and Type.GetTypeArray() but
-                // they don't exist in all runtimes we support.
-                argTypes = new Type[args.Length];
-
-                int index = 0;
-                foreach (object arg in args)
-                    argTypes[index++] = arg.GetType();
+                Type[] argTypes = Reflect.GetTypeArray(fixture.Arguments);
```
rprouse: Removed duplicate code

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -181,12 +183,15 @@ public static object Construct(Type type, object[] arguments)
...
-        private static Type[] GetTypeArray(object[] objects)
+        internal static Type[] GetTypeArray(object[] objects)
         {
             Type[] types = new Type[objects.Length];
             int index = 0;
             foreach (object o in objects)
-                types[index++] = o.GetType();
+            {
+                // NUnitNullType is a marker to indicate null since we can't do typeof(null) or null.GetType()
+                types[index++] = o == null ? typeof(NUnitNullType) : o.GetType();
```
rprouse: Hack? Probably, but open to suggestions :smile:

CharliePoole: What happens if we get a typed null like `(string)null`? Do we recognize that separately? Do we need to?

rprouse: Even if it is a typed null, in this case it still come in as a null parameter. Just to be sure, I added a typed null test and it works fine.

danlyons-softek:
It may be worth adding some tests for `Nullable<T>` arguments, since `Nullable<T>` is itself a struct.

rprouse:
I added tests where the constructor parameters are Nullable and it worked fine. Changes pushed to this PR.

CharliePoole:
This doesn't affect the merging but did you figure out why we have this problem with fixtures but not methods?

rprouse:
I did look at the difference between TestCase and TestFixture. TestCase works because there is only one method, so we just make sure the parameters can be converted to the target parameters. With TestFixture, we were using GetConstructor with types. It was different logic although it is nearly the same now since we get all constructors and check the types of each.

rprouse:
Build failures are for the Nullable types. Not sure why it worked on my machine, I might have made a mistake. Looking at it.

CharliePoole:
@rprouse That makes sense now that you explain it!

rprouse:
Not sure what I did wrong before, but Nullable<T> was not working. I have pushed a fix.

rprouse:
@CharliePoole, anything else that you would like to see in this PR, or can we merge?

PR #1510, opened 2016-05-21
Enable use of TestContext.TestDirectory when tests are being loaded.

### LINKED ISSUE #1241: Make TestDirectory accessible when TestCaseSource attributes are evaluated
TestDirectory is currently a property of the TestExecutionContext, which means it is unavailable at the point TestCaseSource's are evaluated. This prevents test setup's such as the below:

    [Test, TestCaseSource(nameof(TestCases))]
    public void ImportFileDoesNotThrow(string fileName)
    {
       Assert.DoesNotThrow(() => importer.Import(fileName));
    }

    private static IEnumerable<string> TestCases
    {
        get { yield return TestContext.CurrentContext.TestDirectory + "filepath"; }
    }

This came out of a [support query](https://groups.google.com/forum/#!topic/nunit-discuss/YvFwSfvF7aU) on the Google Group - to quote @CharliePoole's considerations:

> Things we could consider include: 
>  * Making TestDirectory a static property of TestContext and basing it on the executing assembly rather than the executing test. 
>  * Moving TestDirectory to some other static class and doing the same thing. 
>  * Leaving it where it is but having TestContext create a "fake" CurrentContext when no test is running. 
> Most likely, if I had thought of this in the first place, I would have put TestDirectory somewhere else entirely.

CharliePoole:
Fixes #1241.

I ended up checking for a null CurrentTest and using the calling assembly in that case. This allows TestContext.TestDirectory to work from within a TestCaseSource method, for example.

At the same time, I did some refactoring of TestExecutionContext to make it easier to understand the three different approaches used to hold current context depending on the build platform.

### DIFF
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -158,6 +158,7 @@ namespace NUnit.Framework.Api
                 Randomizer.InitialSeed = (int)settings[PackageSettings.RandomSeed];
 
             return LoadedTest = _builder.Build(assemblyName, settings);
+
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


rprouse:
Looks good, merging.

PR #1512, opened 2016-05-22
Make teamcity event handler an engine extension

### LINKED ISSUE #1225: The --teamcity option should really be an extension
It's a bit odd that teamcity support is native to NUnit and no other CI server is supported. It would be best if this could be made into an extension.

Doing that requires two things:

1. Actually creating the extension point and extension in the engine, as we have done for other purposes. This one would be some kind of EventListener extension point, and so could be used for a lot of different purposes.

2. Figuring out how to support the console option. Currently, the console is not extensible.

CharliePoole:
Fixes #1225 

I made the code that was previously part of the console runner into an extension for the engine, making minimal changes so it would work. All teamcity-related code is now gone from the console runner except for:
 * Recognizing the --teamcity option
 * Recognizing the environment variable set by TeamCity
 * Enabling / Disabling the extension as needed.

I'd like  both @rprouse and @NikolayPianikov to review this before we merge.

### DIFF
--- a/nunit.sln
+++ b/nunit.sln
@@ -212,6 +212,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "nunit.testdata-3.5", "src\N
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "nunit.framework.tests-3.5", "src\NUnitFramework\tests\nunit.framework.tests-3.5.csproj", "{B93FC354-D59D-4650-84E4-7FCA7D51689A}"
 EndProject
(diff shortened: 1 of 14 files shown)


CharliePoole:
@NikolayPianikov Please see the two questions I asked you in #1225

NikolayPianikov:
@CharliePoole  Looks good. Will this extension a part of zip/msi/nuget package of NUnit?

CharliePoole:
@NikolayPianikov Yes, at least for now. In future, we may want to make it either a separate download or an optional selection in the install. The same decision will need to be made about some of our other extensions like the V2 framework driver.

rprouse:
Code looks good to me, but we should take care of the packaging in the MSI and maybe add a NuGet, or at least add issues committed to 3.4 so we don't forget.

CharliePoole:
I added an issue to the 3.4 milestone. I'm committing this so that it gets out of the way of the issue that @NikolayPianikov is working on.

PR #1517, opened 2016-05-24
Create a NuGet package from the portable agent

rprouse:
**Ready to Merge**

This PR creates a NuGet package that is consumed by the `dotnet-test-nunit` .NET Core runner/command as per #1371 

I want to put the `dotnet-test-nunit` runner in a separate repository, so anything from this repo that I want to consume must be a NuGet package.

I will continue to push to this PR as I ensure that the API provided by `nunit.portable.agent` is complete and provides everything needed. Once that is done, this PR can be merged. Meanwhile, it will provide a convenient feed of NuGet packages from the AppVeyor artifacts.

A first experiment in how we will split the engine and console out of this repo?


rprouse:
@CharliePoole can you review this? I think it is ready to merge. I have the dotnet-test-nunit runner running and exploring tests. It still needs some cleaning up and it not working in Visual Studio yet, but I am looking into those issues.

rprouse:
This PR is blocking additional work. Can an @nunit/core-team member review?

src/NUnitEngine/Portable/nunit.portable.agent/ResultSummary.cs
```diff
@@ -0,0 +1,342 @@
...
+                                ExplicitCount++;
+                            else
+                                SkipCount++;
+                            break;
+                        default:
+                            SkipCount++;
+                            break;
+                    }
+                    break;
+
```
ChrisMaddock: Does this need to be updated with the fix in #1530 - for test fixtures which crash during setup?

src/NUnitEngine/Portable/nunit.portable.agent/ResultSummary.cs
```diff
@@ -0,0 +1,342 @@
...
+using System.Xml.Linq;
+
+namespace NUnit.Engine
+{
+    /// <summary>
+    /// Summary description for ResultSummary.
+    /// </summary>
+    public class ResultSummary
+    {
+        const string TEST_RUN_ELEMENT = "test-run";
```
ChrisMaddock: Unused?

src/NUnitEngine/Portable/nunit.portable.agent/ResultSummary.cs
```diff
@@ -0,0 +1,342 @@
...
+        /// <summary>
+        /// Gets the count of invalid assemblies
+        /// </summary>
+        public int InvalidAssemblies { get; private set; }
+
+        /// <summary>
+        /// An Unexpected error occurred
+        /// </summary>
+        public bool UnexpectedError { get; private set; }
+
```
ChrisMaddock: Should there be something to handle this written to the xml? Would an unexpected error currently be lost and return a failure for no clear reason? I'm entirely clear on what's upstream from this.

ChrisMaddock:
Rob - had a quick look through after the bit similar to #1530 peaked my interest. I'm not 100% on top of  what is required upstream, but what's here looks good to me.

ChrisMaddock:
Does `bin/agents` want to be included in the Cake `PackageZip` task at this point?

src/NUnitEngine/Portable/nunit.portable.agent/ResultSummary.cs
rprouse: Good catch. I also need to apply those changes to the `dotnet-test-nunit` project. That highlights a good reason why we need this class for those other projects like Xamarin and .NET Core, keep it all in one place. I wonder if we could somehow merge the three similar but slightly different `ResultSummary` classes?

src/NUnitEngine/Portable/nunit.portable.agent/ResultSummary.cs
rprouse: Thanks, removed

rprouse:
> Does bin/agents want to be included in the Cake PackageZip task at this point?

This is only intended to ever be used as a NuGet package and doesn't currently work with anything else in this repo, so I left it out. I would also like to see a different release cadence for this package. There is a chance that in the future it will be used by the engine in platform specific agents. If that happens then we can include it.

rprouse:
> I'm not 100% on top of what is required upstream

Currently it is being used by the .NET Core console runner/Visual Studio adapter, so it is used very much like the other `ResultSummary` classes except it can't use the older XmlNode style XML classes, it has to use the newer Linq to XML classes.

To those that question another Visual Studio Adapter, it is just the way that Microsoft did it for .NET Core and UWP. They even released the MSTest version as an alpha last week. Luckily it is just a console runner with some small changes.

Once this PR is merged and I get the `dotnet-test-nunit` runner out as alpha, I want to use this NuGet package in the Xamarin project to allow it to no longer be dependent on a specific version of the NUnit Framework and to remove some code duplication. There is a very similar `ResultSummary` there.

src/NUnitEngine/Portable/nunit.portable.agent/ResultSummary.cs
rprouse: Another good catch, thank you. The errors will be captured in the `test-suite` elements in the final `TestResults`, but I didn't notice that I forgot to include those results when I build the final XML because I am not using/saving the `TestResults` upstream yet.

I have fixed and added another unit test.

rprouse:
@chris-smith-zocdoc thanks for the help, I really appreciate it. You made several good catches that will save me a lot of time. It really shows the value of code reviews :+1: 

If you are interested in helping out more in the .NET Core support, I would appreciate the company :smile: 

https://github.com/nunit/dotnet-test-nunit

rprouse:
To my last comment, I meant @ChrisMaddock, not @chris-smith-zocdoc, sorry. I mistakenly selected the first auto-complete :frowning:

ChrisMaddock:
No problem. :+1: Not currently looking at .NET Core anywhere else, but am intrigued, may try digging in to something if I get a more substantial block of time!

Issue #1518, opened 2016-05-25
NUnit does not send the "testStarted" TeamCity service message when exception was thrown from SetUp/OneTimeSetUp


NikolayPianikov:
Actual:
```
##teamcity[testFailed name='dotNet40NUnitTests.OneTimeSetupTests.TestDo1' message='OneTimeSetUp: System.Exception : OneTimeSetUp exception' details='' flowId='0-1001']

##teamcity[testFinished name='dotNet40NUnitTests.OneTimeSetupTests.TestDo1' duration='0' flowId='0-1001']

##teamcity[flowFinished flowId='0-1001']
```

Expected:
```
##teamcity[flowStarted flowId='0-1001' parent='0-1014']

##teamcity[testStarted name='dotNet40NUnitTests.OneTimeSetupTests.TestDo1' captureStandardOutput='false' flowId='0-1001']

##teamcity[testFailed name='dotNet40NUnitTests.OneTimeSetupTests.TestDo1' message='OneTimeSetUp: System.Exception : OneTimeSetUp exception' details='' flowId='0-1001']

##teamcity[testFinished name='dotNet40NUnitTests.OneTimeSetupTests.TestDo1' duration='0' flowId='0-1001']

##teamcity[flowFinished flowId='0-1001']
```

CharliePoole:
Fixed by PR #1519

PR #1519, opened 2016-05-25
#1518 NUnit does not send the "testStarted" TeamCity service message when exception was thrown from SetUp/OneTimeSetUp

NikolayPianikov:


CharliePoole:
@NikolayPianikov This looks good to me, assuming it's doing what TeamCity wants it to do - which you know better than I do. :-)

I do have one follow-up question. I see that you simulate events from NUnit in order to create unit tests for the TC handler. That makes sense. But do we have tests to ensure that your assumptions about what events NUnit sends remain correct?

@rprouse I'm going ahead and merging this. I suggest that in future one committer review should be enough when @NikolayPianikov is working within the TC extension. For contributions to (e.g.) the engine, we would go back to two reviews. OK?

PR #1520, opened 2016-05-25
Detect Portable, Silverlight and Compact and give error message

CharliePoole:
This is a small step toward resolving #1138. It correctly detects Portable and Silverlight builds by use of the TargetFrameworkAttribute and CF builds by the public key token that is used. For each of these, the driver service refuses to work and returns an error. The error applies only to the specific assembly, in case multiple assemblies are being run, and the execution is not terminated.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Drivers/NUnit3FrameworkDriver.cs
+++ b/src/NUnitEngine/nunit.engine/Drivers/NUnit3FrameworkDriver.cs
@@ -80,7 +80,7 @@ namespace NUnit.Engine.Drivers
             }
             catch (SerializationException ex)
             {
-                throw new NUnitEngineException("The NUnit 3.0 driver does not support the portable version of NUnit. Use a platform specific runner.", ex);
+                throw new NUnitEngineException("The NUnit 3.0 driver cannot support this test assembly. Use a platform specific runner.", ex);
             }
 
             CallbackHandler handler = new CallbackHandler();
--- a/src/NUnitEngine/nunit.engine/IDriverService.cs
+++ b/src/NUnitEngine/nunit.engine/IDriverService.cs
@@ -37,7 +37,8 @@ namespace NUnit.Engine
         /// </summary>
         /// <param name="domain">The AppDomain in which to run the tests</param>
         /// <param name="assemblyPath">The path to the test assembly</param>
+        /// <param name="targetFramework">The value of any TargetFrameworkAttribute on the assembly, or null</param>
         /// <returns></returns>
-        IFrameworkDriver GetDriver(AppDomain domain, string assemblyPath);
+        IFrameworkDriver GetDriver(AppDomain domain, string assemblyPath, string targetFramework);
     }
 }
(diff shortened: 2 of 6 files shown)


src/NUnitEngine/nunit.engine/Drivers/NUnit3FrameworkDriver.cs
```diff
@@ -80,7 +80,7 @@ public string Load(string testAssemblyPath, IDictionary<string, object> settings
             }
             catch (SerializationException ex)
             {
-                throw new NUnitEngineException("The NUnit 3.0 driver does not support the portable version of NUnit. Use a platform specific runner.", ex);
+                throw new NUnitEngineException("The NUnit 3.0 driver cannot support this test assembly. Use a platform specific runner.", ex);
```
CharliePoole: The old error message was actually being triggered by Silverlight and CF builds as well as portable, so it has been reworded. It is now an exception of last resort so it's still appropriate to terminate execution here. Once we are this far in the code, it's hard to do anything else.

src/NUnitEngine/nunit.engine/IDriverService.cs
```diff
@@ -37,7 +37,8 @@ public interface IDriverService
         /// </summary>
         /// <param name="domain">The AppDomain in which to run the tests</param>
         /// <param name="assemblyPath">The path to the test assembly</param>
+        /// <param name="targetFramework">The value of any TargetFrameworkAttribute on the assembly, or null</param>
         /// <returns></returns>
-        IFrameworkDriver GetDriver(AppDomain domain, string assemblyPath);
+        IFrameworkDriver GetDriver(AppDomain domain, string assemblyPath, string targetFramework);
```
CharliePoole: This interface is internal to the engine, so changing it has no compatibility impact.

src/NUnitEngine/nunit.engine/Services/RuntimeFrameworkService.cs
```diff
@@ -211,6 +211,20 @@ private static void ApplyImageData(TestPackage package)
...
+                        if (reference.Name == "mscorlib" && BitConverter.ToUInt64(reference.PublicKeyToken, 0) == 0xac22333d05b89d96)
+                        {
+                            // We assume 3.5, since that's all we are supporting
+                            // Could be extended to other versions if necessary
+                            // Format for FrameworkName is invented - it is not
+                            // known if any compilers supporting CF use the attribute
+                            frameworkName = ".NETCompactFramework,Version=3.5";
+                            break;
+                        }
+                }
```
CharliePoole: If any future compilers use TargetFrameworkAttribute for CF, then this code won't run as frameworkName will be non null. Use of the PublicKeyToken is documented for all known versions of CF. Ideally, we would distinguish various versions of CF here, but I didn't do so. The Version part of the string is not used for anything.

rprouse:
I already have similar changes in the branch https://github.com/nunit/nunit/tree/issue-1138c which detects the target platform of the given test assembly, not of the framework itself. I don't think these changes will conflict so I am okay to merge this.

If you are interested in testing various target platforms, my branch contains a mocks directory in solution root with our mock assembly compiled to target most common platforms.

rprouse:
Typical mono crash. I have restarted the build.

CharliePoole:
Yes, it's the target platform of the test assembly that I'm detecting as well. To be accurate, we were already detecting Silverlight and Portable, but not using the info. The CF detection was trickier because there isn't generally a TargetFrameworkAttribute on our CF assemblies.

PR #1526, opened 2016-05-27
Remove synchronous event processing

### LINKED ISSUE #1476: Examine need for "synchronous" events in event queue

CharliePoole:
Fixes #1476. See comments in the issue.


oznetmaster:
Looks good, and works fine on regular CF build and my special CF build.

PR #1530, opened 2016-05-28
Return non-0 code on test fixture setup failure

### LINKED ISSUE #1379: Console returns 0 for invalid fixtures
Separated out from #1346, if the console encounters invalid test suites, it still returns zero. Invalid assemblies and invalid tests contribute to the return code, but not invalid test suites.

Should we add an additional return code for this? -3 FIXTURE_NOT_FOUND currently appears to be unused - what was the intention of that, for invalid filtering?

```
public static readonly int OK = 0;
public static readonly int INVALID_ARG = -1;
public static readonly int INVALID_ASSEMBLY = -2;
public static readonly int FIXTURE_NOT_FOUND = -3;
public static readonly int UNEXPECTED_ERROR = -100;
```

ChrisMaddock:
Fixes #1379 - if a test fixture setup fails, then an error code of zero is returned.

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleRunner.cs
@@ -44,7 +44,8 @@ namespace NUnit.ConsoleRunner
         public static readonly int OK = 0;
         public static readonly int INVALID_ARG = -1;
         public static readonly int INVALID_ASSEMBLY = -2;
-        public static readonly int FIXTURE_NOT_FOUND = -3;
+        //public static readonly int FIXTURE_NOT_FOUND = -3;    //No longer in use
+        public static readonly int INVALID_TEST_FIXTURE = -4;
         public static readonly int UNEXPECTED_ERROR = -100;
 
         #endregion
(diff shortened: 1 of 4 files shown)


src/NUnitConsole/nunit3-console/ConsoleRunner.cs
```diff
@@ -44,7 +44,8 @@ public class ConsoleRunner
         public static readonly int OK = 0;
         public static readonly int INVALID_ARG = -1;
         public static readonly int INVALID_ASSEMBLY = -2;
-        public static readonly int FIXTURE_NOT_FOUND = -3;
+        //public static readonly int FIXTURE_NOT_FOUND = -3;    //No longer in use
+        public static readonly int INVALID_TEST_FIXTURE = -4;
```
ChrisMaddock: Commented out the FIXTURE_NOT_FOUND code here and in NUnit lite as it's no longer in use. Left it in rather than removing it entirely so it's not accidentally repurposed.

ChrisMaddock:
Tried to extract the two return code lists to a single NUnit.Common file while I was doing this - it seems to two runners should be consistent here.

Didn't do it in this PR as there's one inconsistency - NUnit Lite uses `FILE_NOT_FOUND = -2` whereas Console has `INVALID_ASSEMBLIES = -2`. Given these are similar in purpose (I guess file_not_found is a type of invalidity!) - would it be ok to create a shared list, and have NUnitLite use the `INVALID_ASSEMBLIES` const?

CharliePoole:
This looks good to me.

Regarding the constant names. We could reasonably use INVALID_ASSEMBLY in NUnitLite, since it can't have more than one assembly. INVALID_ASSEMBLIES is used for the console runner because it's possible to have one valid and one invalid assembly in the same run.

However, I would not want to create a single shared list in a common file. We are on a path to break the console runner and the framework into separate repositories and that would just give us one more common file to resolve.

ChrisMaddock:
Fair enough, thanks @CharliePoole

PR #1531, opened 2016-05-28
Fixes 1529

### LINKED ISSUE #1529: Settings
Settings are always created as a `Dictionary<string, object>`, but are often passed as an` IDictionary`.  Sometime they are passed correctly as an `IDictionary<string, object>`.

I would think that all references to the Settings should be `IDictionary<string, object>`.

This will improve compile time checking, and will also improve the quality of the code generated when the Settings are accessed.

oznetmaster:
Fixes #1529 

Implements the change suggested in issue #1529.

Uses `IDictionary<string, object>` instead of `IDictionary` for all settings/options references.

### DIFF
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -67,7 +67,7 @@ namespace NUnit.Framework.Api
         /// <param name="assemblyNameOrPath">The AssemblyName or path to the test assembly</param>
         /// <param name="idPrefix">A prefix used for all test ids created under this controller.</param>
         /// <param name="settings">A Dictionary of settings to use in loading and running the tests</param>
-        public FrameworkController(string assemblyNameOrPath, string idPrefix, IDictionary settings)
+        public FrameworkController(string assemblyNameOrPath, string idPrefix, IDictionary<string, object> settings)
         {
             this.Builder = new DefaultTestAssemblyBuilder();
             this.Runner = new NUnitTestAssemblyRunner(this.Builder);
@@ -82,7 +82,7 @@ namespace NUnit.Framework.Api
         /// <param name="assembly">The test assembly</param>
         /// <param name="idPrefix">A prefix used for all test ids created under this controller.</param>
         /// <param name="settings">A Dictionary of settings to use in loading and running the tests</param>
-        public FrameworkController(Assembly assembly, string idPrefix, IDictionary settings)
+        public FrameworkController(Assembly assembly, string idPrefix, IDictionary<string, object> settings)
             : this(assembly.FullName, idPrefix, settings)
         {
             _testAssembly = assembly;
@@ -98,7 +98,7 @@ namespace NUnit.Framework.Api
         /// <param name="settings">A Dictionary of settings to use in loading and running the tests</param>
         /// <param name="runnerType">The Type of the test runner</param>
         /// <param name="builderType">The Type of the test builder</param>
-        public FrameworkController(string assemblyNameOrPath, string idPrefix, IDictionary settings, string runnerType, string builderType)
+        public FrameworkController(string assemblyNameOrPath, string idPrefix, IDictionary<string, object> settings, string runnerType, string builderType)
         {
             Builder = (ITestAssemblyBuilder)Reflect.Construct(Type.GetType(builderType));
             Runner = (ITestAssemblyRunner)Reflect.Construct(Type.GetType(runnerType), new object[] { Builder });
@@ -117,27 +117,27 @@ namespace NUnit.Framework.Api
         /// <param name="settings">A Dictionary of settings to use in loading and running the tests</param>
         /// <param name="runnerType">The Type of the test runner</param>
         /// <param name="builderType">The Type of the test builder</param>
-        public FrameworkController(Assembly assembly, string idPrefix, IDictionary settings, string runnerType, string builderType)
+        public FrameworkController(Assembly assembly, string idPrefix, IDictionary<string, object> settings, string runnerType, string builderType)
             : this(assembly.FullName, idPrefix, settings, runnerType, builderType)
         {
             _testAssembly = assembly;
         }
 
-        private void Initialize(string assemblyPath, IDictionary settings)
+        private void Initialize(string assemblyPath, IDictionary<string, object> settings)
         {
             AssemblyNameOrPath = assemblyPath;
             Settings = settings;
 
-            if (settings.Contains(PackageSettings.InternalTraceLevel))
+            if (settings.ContainsKey(PackageSettings.InternalTraceLevel))
             {
                 var traceLevel = (InternalTraceLevel)Enum.Parse(typeof(InternalTraceLevel), (string)settings[PackageSettings.InternalTraceLevel], true);
 
-                if (settings.Contains(PackageSettings.InternalTraceWriter))
+                if (settings.ContainsKey(PackageSettings.InternalTraceWriter))
                     InternalTrace.Initialize((TextWriter)settings[PackageSettings.InternalTraceWriter], traceLevel);
 #if !PORTABLE && !SILVERLIGHT
                 else
                 {
-                    var workDirectory = settings.Contains(PackageSettings.WorkDirectory) ? (string)settings[PackageSettings.WorkDirectory] : Env.DefaultWorkDirectory;
+                    var workDirectory = settings.ContainsKey(PackageSettings.WorkDirectory) ? (string)settings[PackageSettings.WorkDirectory] : Env.DefaultWorkDirectory;
                     var logName = string.Format(LOG_FILE_FORMAT, Process.GetCurrentProcess().Id, Path.GetFileName(assemblyPath));
                     InternalTrace.Initialize(Path.Combine(workDirectory, logName), traceLevel);
                 }
@@ -174,7 +174,7 @@ namespace NUnit.Framework.Api
         /// <summary>
         /// Gets a dictionary of settings for the FrameworkController
         /// </summary>
-        public IDictionary Settings { get; private set; }
+        public IDictionary<string, object> Settings { get; private set; }
 
         #endregion
 
@@ -420,7 +420,7 @@ namespace NUnit.Framework.Api
         /// <param name="targetNode">Target node</param>
         /// <param name="settings">Settings dictionary</param>
         /// <returns>The new node</returns>
-        public static TNode InsertSettingsElement(TNode targetNode, IDictionary settings)
+        public static TNode InsertSettingsElement(TNode targetNode, IDictionary<string, object> settings)
         {
             TNode settingsNode = new TNode("settings");
             targetNode.ChildNodes.Insert(0, settingsNode);
@@ -430,7 +430,7 @@ namespace NUnit.Framework.Api
 
 #if PARALLEL
             // Add default values for display
-            if (!settings.Contains(PackageSettings.NumberOfTestWorkers))
+            if (!settings.ContainsKey(PackageSettings.NumberOfTestWorkers))
                 AddSetting(settingsNode, PackageSettings.NumberOfTestWorkers, NUnitTestAssemblyRunner.DefaultLevelOfParallelism);
 #endif
 
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -129,18 +129,18 @@ namespace NUnit.Framework.Api
             AssemblyNameOrPath = assemblyPath;
 
             var newSettings = settings as IDictionary<string, object>;
-            Settings = newSettings ?? settings.Cast<DictionaryEntry>().ToDictionary(de => (string)de.Key, de => de.Value);
+            InternalSettings = newSettings ?? settings.Cast<DictionaryEntry>().ToDictionary(de => (string)de.Key, de => de.Value);
 
-            if (Settings.ContainsKey(PackageSettings.InternalTraceLevel))
+            if (InternalSettings.ContainsKey(PackageSettings.InternalTraceLevel))
             {
-                var traceLevel = (InternalTraceLevel)Enum.Parse(typeof(InternalTraceLevel), (string)Settings[PackageSettings.InternalTraceLevel], true);
+                var traceLevel = (InternalTraceLevel)Enum.Parse(typeof(InternalTraceLevel), (string)InternalSettings[PackageSettings.InternalTraceLevel], true);
 
-                if (Settings.ContainsKey(PackageSettings.InternalTraceWriter))
-                    InternalTrace.Initialize((TextWriter)Settings[PackageSettings.InternalTraceWriter], traceLevel);
+                if (InternalSettings.ContainsKey(PackageSettings.InternalTraceWriter))
+                    InternalTrace.Initialize((TextWriter)InternalSettings[PackageSettings.InternalTraceWriter], traceLevel);
 #if !PORTABLE && !SILVERLIGHT
                 else
                 {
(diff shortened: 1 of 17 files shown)


rprouse:
Nice cleanup. Looks like a lot of this is leftover from the .NET 1.x days.

rprouse:
Not sure what this error is about, did you use the cake build script with a TestAll?

```
TestDriver
Executing task: TestDriver
NUnit Console Runner 3.3.0 
 
Runtime Environment
  CLR Version: 4.0.30319.42000
 
Test Files
    agents/nunit.portable.agent.tests.dll
 
 
Errors and Failures
 
1) Invalid : C:\projects\nunit\bin\Release\agents\nunit.portable.agent.tests.dll
The given key was not present in the dictionary.
 
Run Settings
    WorkDirectory: C:\projects\nunit\bin\Release
    ImageRequiresX86: False
    NumberOfTestWorkers: 2
 
Test Run Summary
  Overall result: Failed
  Test Count: 0, Passed: 0, Failed: 0, Inconclusive: 0, Skipped: 0
  Start time: 2016-05-28 23:53:16Z
    End time: 2016-05-28 23:53:16Z
    Duration: 0.123 seconds
 
Finished executing task: TestDriver
```

oznetmaster:
I used the cake Build script, but mistyped the TestAll option, so it built everything, but ran no tests :(

oznetmaster:
The fact that this bug occurred is exactly why making these changes is so valuable :)

CharliePoole:
See my extended comment on #1529 

There are two questions I'd like to answer before merging this...

1. Are we changing any of our published APIs?

2. If we are, are we sure we want to do that?

oznetmaster:
I cannot answer about published APIs :(

The issue of what is published and what is not, and what should not be if it is, seems to often be a mystery :(

oznetmaster:
If there is an API issue, I can add a "compatibility" API that simple maps the 'IDictionary' to an 'IDictionary<string, object>' and mark it obsolete so that people will change, eventually.

Please let me know if any APIs need to be fixed in this way.

oznetmaster:
The only APIs that I can find already use `IDictionary<string, object>`!

I think this is a non-issue.

oznetmaster:
Since the published API already uses the generic version, it appears that there already was a decision to always require that the framework support generic collections, so this is seemingly a non-issue as well.

oznetmaster:
Ok, I see that the FrameworkDriver actually previosuly "dumbed down" the `IDictionary<string, object>` to an `IDictionary` before passing it to the FrameworkController, apparently in anticipation of a framework which had no generic collections.  But again, the last framework that did not have generic collections was .NET 1.0

CharliePoole:
See further comments on the issue. I'll add that this question is __not__ about what frameworks __we__ will support. It's about use of our APIs by other people over whom we have no control.

Frankly, this is becoming a tempest in a teapot. I feel I asked a very simple and relatively small question.

oznetmaster:
As to the questions @CharliePoole asked, they are seemingly not so small and simple.

It seems that the definition of the "published API" is itself unclear.  What I have distilled from this discussion is that whatever is currently implemented defines the "published API".  Is that correct?  Not any documentation?

oznetmaster:
This PR is now fully compliant with the "published API" for FrameworkController.

As far as I can tell, this is the only "published API" compliance issue.

CharliePoole:
What the documents and notes tell you is that certain APIs are guaranteed. For example, the names of the Types and the expected constructor parameters for the Framework API are a published API. 

Obviously, if we actually changed things when we built the application, the docs should be brought up to date. But that doesn't make the interface any less of a published API.

This seems like a really late date to be arguing about what the term "published API" means. We have been using the term in all our work toward NUnit 3.0 since I started in 2007. We have told people that the existence of published APIs is one of the key differences between 3.0 and 2.x. We have been saying that they could guarantee that their third-party runners would continue to work if they only stuck to those APIs and didn't just pick up any odd public class they found within NUnit.

If it's not clear, then please tell me how we can make it clear because it's pretty darned important.

oznetmaster:
I always assumed that "published API" was what was published in some document that defined the API, and that the documented API would be guaranteed to exist, and be maintained.

That was apparently a misunderstanding on my part.  I now understand that the documentation is a hindsight product, and will be written to reflect whatever is actually implemented, but will not drive or define the implementation.

This is a good thing to know.

Anyway, this PR now is, as far as I can tell, 100% compliant with the "published API".

oznetmaster:
I have looked, and cannot find a statement equivalent to what @CharliePoole  wrote above.

There is so much written material I may have missed it. If so, ignore this.

I think that a clear list of "usable" APIs should exist in a single place (page?), and that it should be stated that their actual definition is not in the documentation, but in the implementation.  That the documentation is only for explanatory purposes, not for definitional ones, and that it may not correspond to the actual implementation.  

I also notice that the documentation of these classes (like FrameworkController) makes no references to the properties of the classes, just their methods.  Are the properties not part of the "published API"?  Should they be?

oznetmaster:
I remember reading somewhere that there should be a cleaning up of each of the "published API" classes to make any properties or methods that were not to be part of the published API either internal or private.

It would seem that a simple list of the classes, methods and properties that we wanted to be part of the "published API" would simplify this process.  Everything else should be made either internal or private.

This could precede the documentation process for these published API's.

CharliePoole:
@oznetmaster Your interpretation is correct except for your use of the future tense. The API was designed on paper, discussed among us, implemented and modified. We worked with third-party vendors to get it as they needed and we published it (as code) when we released NUnit 3.0. This was quite some time ago. It would have been good if we had found the time to convert the specs into as-built documentation, but we were working on other things.

Until we bring everything up to date, perhaps we can just use some common sense.

CharliePoole:
@oznetmaster I see you're still making changes. Please post a comment when it's OK to do a code review.

oznetmaster:
No, I was done when I said above: Anyway, this PR now is, as far as I can tell, 100% compliant with the "published API".

oznetmaster:
Still waiting for a code review. @CharliePoole had asked for a post when it was ready for review.

src/NUnitFramework/framework/Api/FrameworkController.cs
```diff
@@ -174,7 +177,15 @@ private void Initialize(string assemblyPath, IDictionary settings)
...
-        public IDictionary Settings { get; private set; }
+        public IDictionary Settings
+        {
+            get { return (IDictionary)InternalSettings; }
+        }
+
+        /// <summary>
+        /// Gets a dictionary of settings for the FrameworkController
+        /// </summary>
+        internal IDictionary<string, object> InternalSettings { get; private set; }
```
rprouse: I dislike the need for the InternalSettings, but i can see why you had to do it in this context. It is too bad that we didn't catch this before release. I do prefer the generic dictionary.

rprouse:
I do prefer the generic dictionary, but on the other hand, this change does make me a bit uncomfortable, but I'm not sure why. @CharliePoole can you see any potential side effects or consequences that we are not thinking about? If not, then I think the code looks good. I don't think there are external API changes and I don't think the internal API changes will cause problems although I do think it is that second part that is making me nervous.

I've code reviewed and the code itself looks good.

src/NUnitFramework/framework/Api/FrameworkController.cs
oznetmaster: I am not sure this is even necessary.  All the other classes that use Settings keep their copy internal.

It is only because this is released that this seems necessary.  If we ask, and no one is using the public Settings property, it can be removed without any problem, and the InternalSettings can then be renamed to Settings.

CharliePoole:
@rprouse I'm equally uncomfortable. The change was proposed as purely internal. In my view, it's a coding improvement. However, I don't see us getting enough benefit to justify changing our APIs.

I'll start my own code review now and come back with my own conclusion shortly.

oznetmaster:
I am confused.  What API's have been changed?  There are now no published API's that have been changed, that I am aware of.

src/NUnitFramework/framework/Api/FrameworkController.cs
CharliePoole: I don't see the need and frankly would have prefered to review without that change, as I originally suggested. We seem to be fixing a compatibility problem that @oznetmaster sees with the framework API, but I don't see this property as a part of that API. The API description indicates that the names and accepted types of the constructors for the FrameworkController and it's subtask classes must remain constant. That's all.

Since users of the API never reference the controller object as anything other than object, the Settings property is not available to them. We should check to see if it's referenced by anything other than tests and, if not, change its visibility.

I would prefer to see the whole business with InternalSettings dropped as an unnecessary complication.

CharliePoole:
@oznetmaster As you are quite aware, I have not yet reviewed the code. Therefore, I don't know if any APIs are broken. When I have reviewed the code, I will have an opinion.

rprouse:
I only see internal APIs changed. I just want to make sure that doesn't have some unintended consequences with different engine versions.

src/NUnitFramework/framework/Api/FrameworkController.cs
oznetmaster: @CharliePoole made it very clear that the actual implementation defined the "published API", not any documentation.

I asked the question multiple times if this includes the properties of the public classes, as well as all of the public methods.  The implication of all replies to those was that if it is implemented and public, then it is part of the "published API" for those classes which are defined as being part of the API.

Now, something different is being said.  Someone needs to put a stake in the ground on this issue.

I did not want to have to do this InternalSettings things, but I felt compelled to by the statements that @CharliePoole  made on the "published API" issue.

oznetmaster: @CharliePoole  specifically stated that there were third part developers who were, in fact, using the FrameworkController directly, not as an opaque object.

Now he is saying that is not the case. I am understandably confused by all of this.

CharliePoole: @oznetmaster I told you that third party developers were using the Framework API but that I would have to ask them how they were using it. I couldn't ask them until I had some questions answered. I asked you not to make changes based on speculation, but to first determine for me if the API was broken. That's still where we a

It sounds like "using the framework API directly" translated for you into having a direct reference to the FrameworkController. If anyone is doing that, then they are not using the Framework API, which is specifically defined as a Reflection API. I could have easily clarified that for you if you had asked rather than focusing on your criticisms of the level of documentation.

oznetmaster:
As usual, the Travis CI build failed for some spurious reason.  Restarting.

CharliePoole:
Unfortunately, and through my own fault, I just lost my entire general summary of the review I did, while responding to @oznetmaster 's latest commentary. I'm starting over.

@oznetmaster We will come to a conclusion much sooner if you can simply wait for the review.

oznetmaster:
Both @rprouse  and @CharliePoole indicated that InternalSettings should be removed, so I did.

src/NUnitFramework/framework/Api/FrameworkController.cs
```diff
@@ -126,18 +127,20 @@ public FrameworkController(Assembly assembly, string idPrefix, IDictionary setti
         private void Initialize(string assemblyPath, IDictionary settings)
         {
             AssemblyNameOrPath = assemblyPath;
-            Settings = settings;
 
-            if (settings.Contains(PackageSettings.InternalTraceLevel))
+            var newSettings = settings as IDictionary<string, object>;
+            Settings = newSettings ?? settings.Cast<DictionaryEntry>().ToDictionary(de => (string)de.Key, de => de.Value);
```
CharliePoole: I'm unsure about the need for this copying. I'll enlarge in the general comment on the PR.

CharliePoole:
Code itself looks good, with the possible exception of making a copy of the dictionary.

Regarding API compatibility...

The Engine and Driver APIs are not affected at all. Only the Framework API is potentially impacted.

The Framework API consists of a number of well-known class names as well as the Types of their constructor arguments. Nothing else in the  framework is intended to be referenced by clients (with the exception of our own NUnitLite code, which is able to use internal Types).

As it currently stands, there is no change to the Framework API. In fact, @oznetmaster has jumped through a bit of a hoop to avoid changing the Type of the collection passed to the constructor of the FrameworkController. I'm not sure this is necessary.

If we simply changed the constructor, it would be a break in the API. However, it may be one that third parties could live with. NUnit itself is already passing in a generic dictionary of the proper type and it's possible that the third party vendor (I believe there is only one: NCrunch) is doing the same thing. We could certainly ask.

OTOH, since the work is done, we could continue to copy the dictionary, which might avoid some problems since it is not a readonly dictionary. OR... we could check the actual type and copy if needed. I'm OK with any of the three options, but I'll have to contact the vendor if we select the first.

oznetmaster:
It should be noted that `IDictionary` does not have to be a Dictionary<string, object>.  It could be a Hashtable, a SortedList, or a HybridDictionary, for instance.  That is why I chose to copy it.

oznetmaster:
I am currently checking the actual type, and only copying if needed.  I am not sure what @CharliePoole is suggesting.

CharliePoole:
I listed three options. They are mutually exclusive.

1. Break the API and require an `IDictionary<string, object>`
2. Always copy the provided `IDictionary` to a `Dictionary<string, object>`
3. Only copy when the type provided is not `IDictionary<string, object>`

We should pick one (together) before making any changes.

oznetmaster:
I have to say that this entire process has been both painful and overly time consuming.

I made these changes, as I always do, in my CF build first, and then, after receiving encouragement to do so, back ported them to the main branch.  If I had realized how much of an issue this was going to be, I would have never bothered.

It was important in my CF build because boxing and unboxing are very expensive on CF, and therefore using non-generic collections is to be avoided at all costs.

If this is not something that is really desirable in the main NUnit build, then lets just forget it.  I will keep it in my CF build no matter what.  As always, in my CF build, it is currently conditioned.

Frankly, I do not have any more time to spend on what seemed (obviously incorrectly) a simple performance enhancement.  I am not sure it is worth anyone else spending any more time on it either.  I doubt it will provide any noticeable performance change on non-CF builds.  The only real benefit will be the compile time error checking. The need to involve third parties in the process really makes it seem a waste of everyone's time.

It is up to @CharliePoole  and @rprouse  what happens with this going forward.

oznetmaster:
@CharliePoole 

2 is actually bad programming practice in most cases.  Why would it make  sense to copy something for no reason, since it is already in the desired format, and it is read only within FrameworkController?

oznetmaster:
Just so I am absolutely clear in the future, is it being stated that FrameworkController should have no public properties?  Properties should only be public if they are available for use by a third party.

If the answer is no, then perhaps they should all be made internal as part of the PR, or another one.

CharliePoole:
@oznetmaster Answering your last three comments in one shot...

It has indeed been painful. Hopefully, you have some awareness of your own responsibility in making it so. If not and if you want to discuss it, we should take it out of the public sphere. Feel free.

I didn't advocate option 2. I just listed it.

The issue of public properties and methods is old news. We have talked about it extensively. Once again: NUnit has always made classes public in order to test them. This has roots in the capabilities of early .NET releases as well as the background of the earliest developers of xUnit-style frameworks in languages that do not have private or protected visibility. We have discussed changing this and are in the process of doing it. It will be a slow and careful process. Meanwhile, the fact that something is public does not mean that it is supported for external use.

oznetmaster:
"Meanwhile, the fact that something is public does not mean that it is supported for external use."

Then how can any developer know what is and is not supported for external use?

Obviously, I was unable to, even though I examined all of what I thought was relevant documentation.

We have already discussed the fact that the current documentation is of limited use in this area..

CharliePoole:
@oznetmaster They can't know. Which is why we are changing it. Why do you keep beating this dead horse?

oznetmaster:
Because you seemingly expected me to know.  And it is the ONLY cause of all of this pain.

rprouse:
I think this looks good now. If anyone has problems because they have used an `Internal` API, then they can update. Thanks for the changes. Sorry it took so much back and forth.

PR #1536, opened 2016-05-30
Direct output from tests

### LINKED ISSUE #1139: Nunit3 console doesn't show test output continously
The Nunit 2 console had a great feature, it reported the test output on the fly and not only after the test finished. It was very useful because you could track the test case and check the progress, but now I have to wait the end of the test case to see the output.
Do you plan to reimplement this feature?

### LINKED ISSUE #1528: Use of Sleep(0) in NUnit
There seems to be a lot of use of Sleep(0) throughout NUnit.

Microsoft seems to really indicate that this is a bad practice:

[http://joeduffyblog.com/2006/08/22/priorityinduced-starvation-why-sleep1-is-better-than-sleep0-and-the-windows-balance-set-manager/](url)

Should we examine the usage and adjust accordingly?

CharliePoole:
Fixes #1139 

Console.Error output is now sent to the console using an event, as in NUnit V2. Additionally, TestContext.Error and TestContext.Progress may also be used. Console.Error and TestContext.Error output display in red on the color console, while TestContext.Progress uses the standard output color.

Also fixes #1528

### DIFF
--- a/src/NUnitConsole/nunit3-console/TestEventHandler.cs
+++ b/src/NUnitConsole/nunit3-console/TestEventHandler.cs
@@ -64,7 +64,7 @@ namespace NUnit.ConsoleRunner
                     break;
 
                 case "test-output":
-                    WriteTestOutput(testEvent);
+                    TestOutput(testEvent);
                     break;
             }
         }
@@ -79,14 +79,14 @@ namespace NUnit.ConsoleRunner
             var outputNode = testResult.SelectSingleNode("output");
 
             if (_displayLabels == "ALL")
-                WriteTestLabel(testName);
+                WriteLabelLine(testName);
 
             if (outputNode != null)
             {
                 if (_displayLabels == "ON")
-                    WriteTestLabel(testName);
+                    WriteLabelLine(testName);
 
-                WriteTestOutput(outputNode);
+                WriteOutputLine(outputNode.InnerText);
             }
         }
 
@@ -98,25 +98,43 @@ namespace NUnit.ConsoleRunner
             if (outputNode != null)
             {
                 if (_displayLabels == "ON" || _displayLabels == "ALL")
-                    WriteTestLabel(suiteName);
+                    WriteLabelLine(suiteName);
 
-                WriteTestOutput(outputNode);
+                WriteOutputLine(outputNode.InnerText);
             }
         }
 
-        private void WriteTestLabel(string name)
+        private void TestOutput(XmlNode outputNode)
         {
-            using (new ColorConsole(ColorStyle.SectionHeader))
-                _outWriter.WriteLine("=> {0}", name);
+            var testName = outputNode.GetAttribute("testname");
+
+            if (_displayLabels == "ON" && testName != null)
+                WriteLabelLine(testName);
+
+            WriteOutputLine(outputNode.InnerText);
         }
 
-        private void WriteTestOutput(XmlNode outputNode)
+        private string _currentLabel;
+
+        private void WriteLabelLine(string label)
+        {
+            if (label != _currentLabel)
+            {
+                using (new ColorConsole(ColorStyle.SectionHeader))
+                    _outWriter.WriteLine("=> {0}", label);
+
+                _currentLabel = label;
+            }
+        }
+
+        private void WriteOutputLine(string text)
         {
             using (new ColorConsole(ColorStyle.Output))
             {
-                _outWriter.Write(outputNode.InnerText);
+                _outWriter.Write(text);
+
                 // Some labels were being shown on the same line as the previous output
-                if (!outputNode.InnerText.EndsWith("\n"))
+                if (!text.EndsWith("\n"))
                 {
                     _outWriter.WriteLine();
                 }
--- a/src/NUnitConsole/nunit3-console/TestEventHandler.cs
+++ b/src/NUnitConsole/nunit3-console/TestEventHandler.cs
@@ -147,11 +147,6 @@ namespace NUnit.ConsoleRunner
             }
         }
 
-        private void WriteErrorLine(string text)
-        {
-
-        }
-
         #endregion
 
         #region InitializeLifetimeService
--- a/src/NUnitConsole/nunit3-console/TestEventHandler.cs
+++ b/src/NUnitConsole/nunit3-console/TestEventHandler.cs
@@ -41,7 +41,7 @@ namespace NUnit.ConsoleRunner
 
         public TestEventHandler(TextWriter outWriter, string displayLabels)
         {
-            _displayLabels = displayLabels;
+            _displayLabels = displayLabels.ToUpperInvariant();
             _outWriter = outWriter;
         }
 
--- a/src/NUnitConsole/nunit3-console/TestEventHandler.cs
+++ b/src/NUnitConsole/nunit3-console/TestEventHandler.cs
@@ -62,6 +62,10 @@ namespace NUnit.ConsoleRunner
                 case "test-suite":
                     SuiteFinished(testEvent);
                     break;
+
+                case "test-output":
+                    WriteTestOutput(testEvent);
+                    break;
             }
         }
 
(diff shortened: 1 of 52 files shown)


src/NUnitConsole/nunit3-console/TestEventHandler.cs
```diff
@@ -94,31 +98,60 @@ private void SuiteFinished(XmlNode testResult)
...
                 // Some labels were being shown on the same line as the previous output
-                if (!outputNode.InnerText.EndsWith("\n"))
+                if (!text.EndsWith("\n"))
                 {
                     _outWriter.WriteLine();
                 }
             }
         }
 
+        private void WriteErrorLine(string text)
```
rprouse: Should this empty unused method be removed, or is there more coming?

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -245,17 +245,37 @@ public bool WaitForCompletion(int timeout)
...
+
+        #endregion
+
+        #region Helper Methods
+
+        /// <summary>
         /// Initiate the test run.
         /// </summary>
-        public void StartRun(ITestListener listener)
+        private void StartRun(ITestListener listener)
```
rprouse: Minor organization, but I think that StartRun should come before StopRun and be in the same region now that they are both public. Does this change our internal API at all making this public?

src/NUnitFramework/framework/Interfaces/TestOutput.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// Construct with text, ouput destination type and
+        /// the name of the test that produced the output.
+        /// </summary>
+        /// <param name="text">Text to be output</param>
+        /// <param name="stream">Destination of output</param>
+        /// <param name="testName">Name of teset that produced the output</param>
+		public TestOutput(string text, string stream, string testName)
+        {
+            Text = text;
+            Stream = stream;
```
rprouse: Is Stream a good name for this? It confused me until I looked at the type and the XMLDocs.

src/NUnitConsole/nunit3-console/TestEventHandler.cs
CharliePoole: Either it's something I didn't finish or the file did not get saved before I committed. I'll check.

src/NUnitFramework/framework/Interfaces/TestOutput.cs
rprouse: Or is it the XMLDoc that is confusing, reading the code, it looks like the name of the output stream.

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
CharliePoole: Actually, they are not both public. I'm making StartRun, which is a helper method, private. StopRun was always public and still is. The diff is making it look like StartRun is new, but if you look further down, you'll see its location merely changed.

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -75,6 +76,18 @@ public static TextWriter Out
...
+#if !NETCF && !SILVERLIGHT && !PORTABLE
+        /// <summary>
+        /// Gets a TextWriter that will send output directly to Console.Error
+        /// </summary>
+        public static TextWriter Error = new EventListenerTextWriter("Error", Console.Error);
+
+        /// <summary>
+        /// Gets a TextWriter for use in displaying immediate progress messages
+        /// </summary>
+        public static readonly TextWriter Progress = new EventListenerTextWriter("Progress", Console.Error);
```
rprouse: Double check my understanding, Progress goes to Console.Error because we are capturing Console.Out, but in this case it will not get output in red, correct? I will probably answer my question as I get further down in the code review :smile:

src/NUnitFramework/framework/Interfaces/TestOutput.cs
CharliePoole: It's the name of the "stream" of output - which is in no way a .NET Stream. In the code for TestOutput, I call it StreamName, but Stream seemed better here. However, if you think it's confusing, I'm glad to change it.

CharliePoole: BTW, in V2, this is called Type, which could be even more confusing. :-)

rprouse: Yah, I saw the usage further down and I think it will be clear in context in the XML. Should we update the XMLDoc accordingly?

src/NUnitConsole/nunit3-console/TestEventHandler.cs
CharliePoole: It represents a road not taken. I pushed a new commit.

src/NUnitFramework/framework/TestContext.cs
CharliePoole: Exactly... sending it to Console.Error is an implementation detail - a convenience at this point. All immediate output goes to Console.Error and becomes a TestOutput event. The Stream name of the event is actually what indicates where the output is going and the runner client can do whatever it wants. For example, a GUI might put progress noticess into some special panel. All the console runner does is color the Error stream red.

CharliePoole: Even if I say it myself, I think it's rather cool how the console is handling this output. There is a test called TextOutputTests.MultipleWrites that displays various kinds of output on the console. If you run it you can see how the output combines. All the labels get combined so long as no parallel-running test produces output in the middle. Each line is colored as needed.

oznetmaster:
I still have a problem with the name Progress, since it is not really reflective of any measure of Progress, but is a TextWriter.

Why call it Progress?

rprouse:
It isn't perfect, but Progress because they are intended for people who want to track the progress of their tests? Do you have a better alternative?

CharliePoole:
@oznetmaster Well, the top reason is that it's what was discussed with users on the issue. :-)

However, I think it is a general progress message - or can be used as one. I think of this as the Progress output stream as distinct from a Progress event, which might carry a percentage or other enumeration in a formatted way. That's a separate thing, IMO.

A percentage would be a very specific kind of progress. I've often seen both combined in Linux, where there is a progress bar driven by some sort of progress events, coupled with a series of detailed messages that are displayed optionally below the bar.

In any case, I don't think having a Progress TextWriter should be confusing at all, since if we actually wanted to report a progress percent, it would probably be a method call and not a TextWriter.

CharliePoole:
Reviewing the discussion on the issue, nobody was that concerned about the name one  way or another. We discussed using Progress, Immediate and ImmediateOut. I picked Progress and put it into a micro-spec comment but that was only last week.

I like Progress because it doesn't say what NUnit should do with the message. Rather, the name says what the thing is - a progress-related text message. NUnit is empowered to do whatever is appropriate with that stream. In the case of a console runner, the logical thing is to display it immediately, but it might be different under some other runner. In general, tests do not know what runner is running them.

There doesn't seem to be any confusion with anything we currently have in NUnit. However, I think @oznetmaster is concerned about the name overlapping a future feature that allows tests to report a percent complete. I'm inclined to deal with that if and when we come to it.

espenalb:
For what it is worth: Thank you for implementing this. Feature was sorely missed. Looking forward to the next release. :smile:

src/NUnitFramework/framework/Interfaces/TestOutput.cs
CharliePoole: Do you mean the page in the docs wiki? I'd say yes. :-)

rprouse: Actually, I was just thinking the XMLDoc. I think the purpose is pretty clear when viewed in the final XML output.

CharliePoole: Of course! I always forget what those /// comments are called. :-)

Changing it inline...

CharliePoole: I guess not... "You must be on a branch to edit a file"

oznetmaster:
I would not implement the progress percentage as a method call (in fact I did not in my CF build), but as a property on the TextContext.  Setting it to a new value causes the event to be raised.

Like setting the value on a progress bar.  It is a property, not a method.

PR #1544, opened 2016-06-03
Fixes #1484

### LINKED ISSUE #1484: Comparing Equality using IEquatable<T> Should Use Most Specific Method
Consider this example:
```C#
public class EquatableObject : IEquatable<EquatableObject>
{
    public int SomeProperty { get; set; }
    public bool Equals(EquatableObject other)
    {
        if (other == null)
            return false;

        return SomeProperty == other.SomeProperty;
    }
}

public class InheritingEquatableObject : EquatableObject, IEquatable<InheritingEquatableObject>
{
    public int OtherProperty { get; set; }
    public bool Equals(InheritingEquatableObject other)
    {
        if (other == null)
            return false;

        return OtherProperty == other.OtherProperty && Equals((EquatableObject) other);
    }
}
```
The following test currently fails:
```C#
[Test]
public void InheritingAndOverridingIEquatable()
{
    var obj1 = new InheritingEquatableObject { SomeProperty = 1, OtherProperty = 2 };
    var obj2 = new InheritingEquatableObject { SomeProperty = 1, OtherProperty = 2 };
    var obj3 = new InheritingEquatableObject { SomeProperty = 1, OtherProperty = 3 };
    var obj4 = new InheritingEquatableObject { SomeProperty = 4, OtherProperty = 2 };

    Assert.That(obj1, Is.EqualTo(obj2));
    Assert.That(obj1, Is.Not.EqualTo(obj3)); // <-- Fails
    Assert.That(obj1, Is.Not.EqualTo(obj4));
}
```
The second test fails, because the `EquatableObject`'s `Equals()` is used. It should be `InheritingEquatableObject`'s `Equals()` instead.

This is because [`NUnitEqualityComparer.FirstImplementsIEquatableOfSecond()`](https://github.com/nunit/nunit/blob/6f5c24f0a46d152a1fbffd907941e7e44c0058d7/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs#L250-257)  just returns the first usable method from the list generated by [`NUnitEqualityComparer.GetEquatableGenericArguments()`](https://github.com/nunit/nunit/blob/6f5c24f0a46d152a1fbffd907941e7e44c0058d7/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs#L259-275):
```C#
ivate static MethodInfo FirstImplementsIEquatableOfSecond(Type first, Type second)
{
    var equatableGenericArguments = GetEquatableGenericArguments(first);
    foreach (var xEquatableArgument in equatableGenericArguments)
        if (xEquatableArgument.Key.IsAssignableFrom(second))
            return xEquatableArgument.Value;

    return null;
}
```
The result from `GetEquatableGenericArguments()` should somehow be sorted, or `FirstImplementsIEquatableOfSecond()` should return the best match, not just the first. The second approach seems to make most sense.

lundmikkel:
This fixes #1484 by not simply using the first possible method, but take a later one, if it is more specific than a former match.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
@@ -249,11 +249,14 @@ namespace NUnit.Framework.Constraints
 
         private static MethodInfo FirstImplementsIEquatableOfSecond(Type first, Type second)
         {
+            var pair = new KeyValuePair<Type, MethodInfo>();
+
             foreach (var xEquatableArgument in GetEquatableGenericArguments(first))
                 if (xEquatableArgument.Key.IsAssignableFrom(second))
-                    return xEquatableArgument.Value;
+                    if (pair.Key == null || pair.Key.IsAssignableFrom(xEquatableArgument.Key))
+                        pair = xEquatableArgument;
 
-            return null;
+            return pair.Value;
         }
 
         private static IList<KeyValuePair<Type, MethodInfo>> GetEquatableGenericArguments(Type type)
--- a/src/NUnitFramework/tests/Constraints/NUnitEqualityComparerTests.cs
+++ b/src/NUnitFramework/tests/Constraints/NUnitEqualityComparerTests.cs
@@ -203,6 +203,28 @@ namespace NUnit.Framework.Constraints
         }
 
         [Test]
+        public void InheritingAndOverridingIEquatable()
+        {
+            var obj1 = new InheritingEquatableObject { SomeProperty = 1, OtherProperty = 2 };
+            var obj2 = new InheritingEquatableObject { SomeProperty = 1, OtherProperty = 2 };
+            var obj3 = new InheritingEquatableObject { SomeProperty = 1, OtherProperty = 3 };
+            var obj4 = new InheritingEquatableObject { SomeProperty = 4, OtherProperty = 2 };
+
+            Assert.That(obj1, Is.EqualTo(obj2));
+            Assert.That(obj1, Is.Not.EqualTo(obj3));
+            Assert.That(obj1, Is.Not.EqualTo(obj4));
+
+            var n = new NUnitEqualityComparer();
+            var tolerance = Tolerance.Exact;
+            Assert.That(n.AreEqual(obj1, obj2, ref tolerance), Is.True);
+            Assert.That(n.AreEqual(obj2, obj1, ref tolerance), Is.True);
+            Assert.That(n.AreEqual(obj1, obj3, ref tolerance), Is.False);
+            Assert.That(n.AreEqual(obj3, obj1, ref tolerance), Is.False);
+            Assert.That(n.AreEqual(obj1, obj4, ref tolerance), Is.False);
+            Assert.That(n.AreEqual(obj4, obj1, ref tolerance), Is.False);
+        }
+
+        [Test]
         public void ImplementingIEquatableDirectlyOnTheClass()
         {
             var obj1 = new EquatableObject { SomeProperty = 1 };
@@ -427,6 +449,18 @@ namespace NUnit.Framework.Constraints
                 return false;
 
             return SomeProperty == other.SomeProperty;
+        }
+    }
+
+    public class InheritingEquatableObject : EquatableObject, IEquatable<InheritingEquatableObject>
+    {
+        public int OtherProperty { get; set; }
+        public bool Equals(InheritingEquatableObject other)
+        {
+            if (other == null)
+                return false;
+
+            return OtherProperty == other.OtherProperty && Equals((EquatableObject) other);
         }
     }
 


src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -249,11 +249,14 @@ public bool AreEqual(object x, object y, ref Tolerance tolerance)
 
         private static MethodInfo FirstImplementsIEquatableOfSecond(Type first, Type second)
         {
+            var pair = new KeyValuePair<Type, MethodInfo>();
+
             foreach (var xEquatableArgument in GetEquatableGenericArguments(first))
                 if (xEquatableArgument.Key.IsAssignableFrom(second))
-                    return xEquatableArgument.Value;
+                    if (pair.Key == null || pair.Key.IsAssignableFrom(xEquatableArgument.Key))
```
rprouse: I assume that you checked that GetEquatableGenericArguments always returns in the order base -> derived? Even if so, there is no guarantee that the Generic argument is more specific than that of the base. I agree that it is likely the case, so this is probably good enough.

rprouse:
@CharliePoole I raised one concern, but this looks good to me. You've been working with the original issue, mind taking a look when you get a chance?

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
CharliePoole: The order of the return values is the order in which they are found by `Type.GetInterfaces`.

The Microsoft docs say

> The GetInterfaces method does not return interfaces in a particular order, such as alphabetical or declaration order. Your code must not depend on the order in which interfaces are returned, because that order varies.

That's pretty much par for the course. NUnit can never depend on the order of finding things by reflection. Even if it's constant for a give runtime, it will vary when we use others. So we would have to descend the inheritance hierarchy in order to provide a guaranteed order. We do this in other places, for example, when we are finding SetUp methods. That said, this is an improvement based on the tests that are passing.

@lundmikkel Do you want to take this further before we merge or should we go ahead with what you have done and do more work on it at another time? I'm OK either way.

CharliePoole:
@rprouse Ideally, I'd like to see us explicitly order the interfaces if possible, but if nobody has time to take that next step, we could merge for now. We should leave some issue behind if we do that.

You brought up a much more subtle issue of also ordering the argument types by how specific they are. I think that's a further and more difficult step.

rprouse:
I think we are okay to merge if we create issues for the potential deficiencies. This is better than the current code, so is a step in the right direction. I've been working all night and need to get to bed. I will create issues and merge in the morning.

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
lundmikkel: @rprouse @CharliePoole Sorry, I didn't get back to you soon; I have been a bit busy.

The idea is as follows:
If there exists any interface with the type of `second`, we take that. If any other interface also matches, we replace the previous interface, if the other is inheriting from the current chosen one; otherwise we ignore it.

This ensures that we always find one, if there is one. It does, however, not ensure that we get the best interface. Consider:

```
interface IFoo{}
interface IBar : IFoo;
interface IBaz : IFoo;
interface IQux : IBaz;
```
We will always end up with either `IEquatable<IBar>` or `IEquatable<IQux>` depending on which interface we see first:
- if we first see `IEquatable<IFoo>`, it depends on the order of the next interface (see two options below)
- if we first see `IEquatable<IBar>`, we end up with that.
- if we first see `IEquatable<IBaz>` or `IEquatable<IQux>`, we end up with `IEquatable<IQux>`.

I don't think there is a best choice here, so I'm not sure if there is a best way to handle this. If the first object implements `IEquatable` for different unrelated types of `second` then there might be a problem that we cannot easily solve.

CharliePoole: There are two inheritance issues here:

1. The inheritance hierarchy of the class on which the interface is implemented

```C#
interface IFoo{}
class Foo : IFoo{}
class Foo2 : Foo, IFoo{}
```

We want to use the IFoo implementation in Foo2, when dealing with a Foo2 object. That's what @rprouse and I were refering to regarding the lack of guarantees in the return order from GetInterfaces. Since Microsoft doesn't guarantee the order, you would need to do this yourself.

2. The inheritance hierarchy of the Type parameters for the  interface. I think that's a subtle, difficult and highly postponable issue. :-)

The first issue is purely mechanical, although complicated, and I think it should be dealt with in the PR.

rprouse: I've merged this PR and created an issue to ensure the order is what we expect. Even if it isn't, I don't think this implementation is any worse than what we already have and on many/most platforms it is better.

lundmikkel: Ah, I see. Yeah, misunderstood what you meant 😄

PR #1545, opened 2016-06-05
Add .Net Standard 1.3 support

jasonwilliams200OK:
Discussion: nunit/dotnet-test-nunit#9.

### DIFF
--- a/nunit.netstandard.sln
+++ b/nunit.netstandard.sln
@@ -1,0 +1,64 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 14
+VisualStudioVersion = 14.0.25123.0
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".SolutionItems", ".SolutionItems", "{76D1A814-CDD2-4A17-A983-49B54926BB0E}"
+	ProjectSection(SolutionItems) = preProject
+		.gitignore = .gitignore
+		global.json = global.json
+		README.md = README.md
+	EndProjectSection
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "mock-assembly", "src\NetStandard\mock-assembly\mock-assembly.xproj", "{8C53C716-3E54-4DFE-8CD9-F4099042FCFA}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "nunitlite", "src\NetStandard\nunitlite\nunitlite.xproj", "{D339BFC2-AF3F-46FA-899A-14BAD4BCA35B}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "slow-nunitlite-tests", "src\NetStandard\slow-tests\slow-nunitlite-tests.xproj", "{74CB429E-69D7-4525-BBDF-BBC47DBC7070}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "nunit.testdata", "src\NetStandard\testdata\nunit.testdata.xproj", "{9ED200E4-2026-4955-87E0-D8CEA7BD9E7F}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "nunit.agent", "src\NetStandard\nunit.agent\nunit.agent.xproj", "{D1567C63-4A0D-4E18-A14E-79699B9BFFFF}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "nunit.framework", "src\NetStandard\nunit.framework\nunit.framework.xproj", "{EE2163B8-13C6-4E3A-9616-1DA5CB08AFEA}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "nunit.framework.tests", "src\NetStandard\nunit.framework.tests\nunit.framework.tests.xproj", "{3AFA44B9-9DF0-494E-B633-6E6B9F2BB808}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{8C53C716-3E54-4DFE-8CD9-F4099042FCFA}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{8C53C716-3E54-4DFE-8CD9-F4099042FCFA}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{8C53C716-3E54-4DFE-8CD9-F4099042FCFA}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{8C53C716-3E54-4DFE-8CD9-F4099042FCFA}.Release|Any CPU.Build.0 = Release|Any CPU
+		{D339BFC2-AF3F-46FA-899A-14BAD4BCA35B}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{D339BFC2-AF3F-46FA-899A-14BAD4BCA35B}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{D339BFC2-AF3F-46FA-899A-14BAD4BCA35B}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{D339BFC2-AF3F-46FA-899A-14BAD4BCA35B}.Release|Any CPU.Build.0 = Release|Any CPU
+		{74CB429E-69D7-4525-BBDF-BBC47DBC7070}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{74CB429E-69D7-4525-BBDF-BBC47DBC7070}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{74CB429E-69D7-4525-BBDF-BBC47DBC7070}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{74CB429E-69D7-4525-BBDF-BBC47DBC7070}.Release|Any CPU.Build.0 = Release|Any CPU
+		{9ED200E4-2026-4955-87E0-D8CEA7BD9E7F}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{9ED200E4-2026-4955-87E0-D8CEA7BD9E7F}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{9ED200E4-2026-4955-87E0-D8CEA7BD9E7F}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{9ED200E4-2026-4955-87E0-D8CEA7BD9E7F}.Release|Any CPU.Build.0 = Release|Any CPU
+		{D1567C63-4A0D-4E18-A14E-79699B9BFFFF}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{D1567C63-4A0D-4E18-A14E-79699B9BFFFF}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{D1567C63-4A0D-4E18-A14E-79699B9BFFFF}.Release|Any CPU.ActiveCfg = Release|Any CPU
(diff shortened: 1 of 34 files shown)


nunit.netstandard.sln
```diff
@@ -0,0 +1,43 @@
...
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".SolutionItems", ".SolutionItems", "{76D1A814-CDD2-4A17-A983-49B54926BB0E}"
+	ProjectSection(SolutionItems) = preProject
+		.gitignore = .gitignore
+		global.json = global.json
+		README.md = README.md
+	EndProjectSection
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "NunitEngine", "NunitEngine", "{73808E53-CBA8-43AF-83DB-9D8E31AD2228}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "nunit.agent", "src\NetStandard\NunitEngine\nunit.agent.xproj", "{D1567C63-4A0D-4E18-A14E-79699B9BFFFF}"
```
rprouse: Our standard is to keep different targets for the same projects in the same directory. Since you put them under NetStandard, I assume you ran into the same problem I did with the `xproj` and `project.json` files being in the same directory as other `csproj` files?

rprouse:
If we are going to create a NetStandard version of the framework, we are going to want to pull in the unit test for it too. We can probably forget about NUnitLite for now.

If we have a NetStandard version of the framework, should we remove .NET Core from the PCL version of the framework?

We are also going to need to add these new projects to the `build.cake` file, to the MSI install and to the Framework NuGet file.

nunit.netstandard.sln
jasonwilliams200OK: Yes that is exactly what I ran into. At some point, we can converge back into csproj targeting NetStandard+net45 (I found that it is possible) and then specify all the dependencies in project.json for each framework. Do you think that work can be done at a later point?

rprouse:
I am curious why you selected netstandard1.3? The PCL version of our framework targets everything under netstandard1.1. Could we not use that target to replace our PCL entirely?

I am still unclear as to what the intended migration path is from PCL to .NET Platform Standard and how the resolution will work in NuGet packages that contain other assembly versions that target compatible frameworks. For example, we have a `net45` target, if there is also a `netstandard1.1` target, will the more specific `net45` be used? Since this is a NuGet thing, will it work with older versions of Visual Studio? How about MonoDevelop 3.x? Should we be dropping our `portable-...` and `dotnet` targets?

jasonwilliams200OK:
The NetStandard version of framework project constitutes the same files which Portable version has. Can you tell me which unit test project covers it?

Yes the NetCore version can be removed from csproj. When we converge back to portable, we will have all the frameworks supported by portable in project.json.

I am not quite familiar with cake build process. I can take a look into it next. Should this be covered in this PR?

rprouse:
For @nunit/core-team members who are less familiar with the .NET Platform Standard, see https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md

jasonwilliams200OK:
> I am curious why you selected netstandard1.3? 

The new projects are only targeting netstandard now. Three dependencies:

* System.IO.FileSystem
* System.Reflection.TypeExtensions
* System.Threading.Thread

do not support NetStandard1.1. We would need to add extra set of dependencies and probably shim some code for it.

rprouse:
That makes sense for `netstandard1.3`. It would allow us to re-enable more of the full functionality of the NUnit Framework. Based on that, I guess we will need to continue to create the PCL if we want to continue to support WinPhone and Windows 8.1. I don't think dropping those would be a terrible loss though :smile:

jasonwilliams200OK:
If we can hold off to it for a little longer, I will try to upgrade the existing Portable csproj with project.json in a separate branch targeting all of the supported matrix. If that works out fine, I will open a new PR (where hopefully we wouldn't need to add a new target in the cake script). :)

jasonwilliams200OK:
@rprouse, I have added 7 projects in total now. I fiddled with csproj way, seems like there are ways to make one csproj ready for multiple targets ([example](http://www.cyotek.com/blog/targeting-multiple-versions-of-the-net-framework-from-the-same-project)), but that does not build for all frameworks at once. I looked at Newtosoft.JSON repo and found that they are also using xproj to target multiple frameworks. First I successfully made it run with project.json, xproj and csproj in same folder, then I moved back to have then in a separate folder due to the complexity it was incurring to the entire build process.

I have added you as a collaborator to my fork. Please let me know how would you like to go about the cake script.

rprouse:
Reviewing, meanwhile, we are now getting a failure in the .NET 2.0 version of the framework running under Mono 3.2.8,

```
1) Failed : NUnit.Framework.Syntax.EqualityTests.EqualityTestsUsingDefaultFloatingPointTolerance
  Expected: 5.0f
  But was:  4.98999977f
at NUnit.Framework.Syntax.EqualityTests.EqualityTestsUsingDefaultFloatingPointTolerance () [0x00000] in <filename unknown>:0
```

rprouse:
@jasonwilliams200OK this is looking like a good start. I will add your fork as a remote tomorrow and pull it down. I've been coding for nearly 12 hours straight now, I need a break and a beer :wink:

jasonwilliams200OK:
That seems like an intermittent failure. I just forced re-pushed the branch as is and Mono 3.2.8 build passed. I tried to identify what is causing this flakiness in float precision, but I am unsure. Might be something to do with Mono's .NET 2 implementation of float has some edge case which we just encountered?

> I've been coding for nearly 12 hours straight now

Wow! You definitely deserve a b&b :wink:

rprouse:
@jasonwilliams200OK one of your stated reasons behind this PR was that the current implementation required Mono on Linux and that was my major reason for considering this PR.

> Otherwise all consumers targeting nunit would need to import dnxcore50 and on Unix, we would have dependency on Mono PCL reference assemblies which is something nobody is going to like IMO.

@jskeet used the current implementation on Linux for Noda.Time and stated;

> I haven't tried running it on Mono - only .NET Core on Linux. That worked fine with the exact project.json above.

Can you point me to some documentation explaining why this PR is needed? Sorry, I'm still playing catch-up with .NET Core. This PR is a big change to our solution and I was planning on waiting until post-3.4 before I tackled a .NET Standard build of the framework, so I need to understand the priority.

jskeet:
Just to be clear, I'd like to *also* be able to run the tests on Mono. After all, I may well have some users who want to use Mono on Linux, and I want to be confident that Noda Time will work for them. I just haven't tried it yet using the runner.

CharliePoole:
@rprouse There hasn't been a really good place for me to jump in with this comment, so I'll just do it here... It seems we are suffering a little by having this work in the main repository. It's pretty hard to mix leading-edge stuff in with support for "legacy" (albeit very recent legacy) without something suffering. Would we be better off to have **all** the work on this being in a separte repository where issues of NUnit release versions and dates don't apply?

rprouse:
@CharliePoole I agree with you, but a version of the .NET Framework that supports a new target for the .NET framework is hard to pull into another repository without a lot of code duplication. I would also want to ship it in the main nunit NuGet package. I could see moving the `nunit.portable.agent` out to it's own repository even though it shares a fair amount of code with the engine now.

I am thinking I might want to merge this onto a separate dev branch for now. That would allow us to do separate smaller PRs without touching master until it is ready to ship. My thinking is that I want to get alphas and then betas out of the `nunit.portable.agent` and `dotnet-test-nunit` out before we ship 3.4 and before July 27th when .NET Core goes RTM. I would then like to ship the RTM version of those packages using the RTM version of .NET Core along with the 3.4 release at the end of the month.

I was originally planning on working on a .NET Core/netstandard version of the nunit framework with 3.6 this summer, but @jasonwilliams200OK felt that it was important to get it out for the `dotnet-test-nunit` runner, thus this PR. I still need to be convinced that it is a priority though. Over the next few days, I want to set up a Linux VM without Mono installed and test the current implementation before I commit to this PR.

jasonwilliams200OK:
@rprouse, sorry I couldn't find link to official documentation explicitly static this fact that repercussions of having `imports` vs no-imports but this is my personal experimentation which led me this conclusion. Let me give you a better evidence:

On fresh Ubuntu box (we can BTW try it on TravisCI by setting lang:c++ where there is no mono installed instead of lang:c#)

```sh
git clone https://github.com/nunit/dotnet-test-nunit
cd dotnet-test-nunit

dotnet restore
# all good so far

dotnet build **/project.json
```

Errors:

```sh
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(46,16): error DOTNET1012: The reference assemblies directory was not specified. You can set the location using the DOTNET_REFERENCE_ASSEMBLIES_PATH environment variable.
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(48,37): error NU1001: The dependency System.ObjectModel could not be resolved.
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(49,41): error NU1001: The dependency System.Threading.Tasks could not be resolved.
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(50,39): error NU1001: The dependency System.Xml.XDocument could not be resolved.
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(25,73): error NU1001: The dependency System could not be resolved.
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(25,73): error NU1001: The dependency System could not be resolved.
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(25,73): error NU1001: The dependency System.Core could not be resolved.
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(25,73): error NU1001: The dependency System.Core could not be resolved.
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(25,73): error NU1001: The dependency Microsoft.CSharp could not be resolved.
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(25,73): error NU1001: The dependency Microsoft.CSharp could not be resolved.
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(25,73): error NU1001: The dependency System.IO.Compression could not be resolved.
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(25,73): error NU1001: The dependency System.Xml could not be resolved.
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(25,73): error NU1001: The dependency System.Xml.Linq could not be resolved.
/home/peter/nunit/dotnet-test-nunit/src/dotnet-test-nunit/project.json(28,45): error NU1001: The dependency System.Net.Http could not be resolved.
```

The first error about `DOTNET_REFERENCE_ASSEMBLIES_PATH` means it is looking for PCL reference assembly.

When/if we publish a demo package with nunit assemblies produced by this change, and use that version in dotnet-test-unit, we can get rid of `imports` in nunit project.json and the said error will go away.

Then here is the comment by @ericstj (the person leading the packaging for CoreFX) supporting this claim: https://github.com/dotnet/corefx/issues/6801#issuecomment-211451042.

jasonwilliams200OK:
> Just to be clear, I'd like to also be able to run the tests on Mono. After all, I may well have some users who want to use Mono on Linux, and I want to be confident that Noda Time will work for them. I just haven't tried it yet using the runner.

@jskeet, this PR is not changing any of the existing RIDs supported by nunit, but adding one more. This is nuget feature that it selects the correct assembly from the package based on the target RID. After this PR we just need to change the nuget pack script to create another folder for NETSTANDARD1.3 and copy the netstandard assemblies to that at packaging time. So when someone would ask for nunit using net standard or net core, it won't fallback to "closest matching", but get the first-hand exact match for netstandard.

rprouse:
@jasonwilliams200OK I wanted to update you on our current thinking for this PR. I am planning on moving the nunit.portable.agent to a separate repository and making it netstandard1.3. I will likely base that off this PR.

Then after the 3.4 release, I am going to once again use this PR to create a netstandard version of the NUnit framework.

So, I won't be merging directly, but I will add your fork as a remote and pull in your changesets to create new PRs. You will get full credit for the work you have done in both repos. I appreciate this work and it will be used. Please keep this PR open for reference and keep your fork available.

I just wanted to let you know that I have not forgotten about this PR. We are two weeks away from the 3.4 release, so we are juggling quite a few balls at the moment.

jasonwilliams200OK:
@rprouse, no problems at all. Thanks for the update.  :)

rprouse:
@jasonwilliams200OK I am going to close this PR so it doesn't clutter up the radar. I've already merged the portable agent half and will do the framework side after the 3.4 release, so please don't delete your fork. Thanks for the help.

PR #1550, opened 2016-06-07
Avoid error when already running in a 32-bit process

### LINKED ISSUE #1548: Invalid Exception when engine is in a 32-bit process
This is related to nunit/nunit3-vs-adapter#181 but may arise with other runners as well.

When a test requires a 32-bit process on a 64-bit machine and is also requested to be run in process, the engine throws an exception with the message "Cannot run tests in process - a 32 bit process is required." This happens even if the engine is already running in a 32-bit process, having been loaded there by the adapter or other runner.

The error originates because the engine, when run by nunit3-console is never loaded in a 32-bit process on a 64-bit machine. Consequently, the assumption got "baked in." 

What is needed is for the engine to determine what kind of a process it is already running in before making the check at https://github.com/nunit/nunit/blob/fe3ef10843dcef8f789f65d8b484cb975a3b1410/src/NUnitEngine/nunit.engine/Runners/MasterTestRunner.cs#L90

CharliePoole:
This fixes #1548 by eliminating the check for running as 32 bits in process unless we are in a 64 bit process.

The change in MasterTestRunner should resolve the current adapter issue nunit/nunit3-vs-adapter#181

The change in the console runner is not currently causing a problem but could do so if someone managed to get the console started in a 32-bit process.

### DIFF
--- a/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/CommandLineTests.cs
@@ -237,8 +237,16 @@ namespace NUnit.ConsoleRunner.Tests
             Assert.AreEqual(0, options.ErrorMessages.Count, "command line should be valid");
         }
 
-        [Test]
-        public void X86AndInProcessAreNotCompatible()
+        [Test, Platform("32-Bit")]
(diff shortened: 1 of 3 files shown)


CharliePoole:
The Travis error seems to be unrelated to this change.

Issue #189, opened 2014-08-19
Add String Operators to 'Does' prefix


CharliePoole:
As suggested by Rob: 
"I would like to add a new verb Does for this. Does.Exist and Does.Not.Exist are much nicer. If I am doing that, it seems a shame to leave it with just exists. I could also add equivalents for Is.StringContaining, Is.StringStarting, etc. Does.Contain, Does.StartWith read better IMHO."

I like this. I documented the necessary changes at https://github.com/nunit/dev/wiki/Breaking-Changes

Note that StringContaining also appears in ConstraintExpression and ConstraintFactory. ConstraintFactory also uses DoesNotContain, DoesNotStartWith, DoesNotEndWith and DoesNotMatch.
We'll have to figure out how to handle each of these as well, probably deprecating them.

PR #1556, opened 2016-06-09
Change default test case name pattern and add console option

### LINKED ISSUE #1549: Changing default behavior for generating test case names
Hello,

Last week I have created issue: https://github.com/nunit/nunit/issues/1537#issuecomment-222730218
It is bigger issue for us, since we have more than ten thousand test cases spread across different visual studio projects. It would be easier for us to actually change the code of NUnit to simply not trim the test case name.
Could this feature be considered in next release?
OR
Can you please point out, which file we will need to change inside the NUnit source code, so that we would "hard code" the maximum value for test case arguments ourselves?

CharliePoole:
Fixes #1549

### DIFF
--- a/nunit.sln
+++ b/nunit.sln
@@ -160,6 +160,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "extensions", "extensions", 
 		nuget\extensions\nunit-v2-result-writer.nuspec = nuget\extensions\nunit-v2-result-writer.nuspec
 		nuget\extensions\nunit.v2.driver.addins = nuget\extensions\nunit.v2.driver.addins
 		nuget\extensions\nunit.v2.driver.nuspec = nuget\extensions\nunit.v2.driver.nuspec
+		nuget\extensions\teamcity-event-listener.nuspec = nuget\extensions\teamcity-event-listener.nuspec
 		nuget\extensions\vs-project-loader.nuspec = nuget\extensions\vs-project-loader.nuspec
 	EndProjectSection
 EndProject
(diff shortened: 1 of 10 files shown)


CharliePoole:
Hey @nunit/core-team ! This needs a review to merge it. :-)

rprouse:
Looks good, but do we need an issue in the adapter to set the pattern? Do you need to do the same in the GUI?

CharliePoole:
There's already PR #189 out for review on the V3 adapter.

I don't think the gui will need to do anything but use the default. IMO, any GUI should take  what we give it and reformat it for display. :-)

oznetmaster:
This PR does not update the CommandLineTests in the nunitlitetests project.

CharliePoole:
It wasn't intended to affect nunitlite, so shouldn't have been in command-line options but in console options. Would you create an issue?

oznetmaster:
Why wouldn't it affect nunitlite?  The option is just as useful there as anywhere else.

oznetmaster:
I am already using it in fact.

CharliePoole:
You're right I was thinking of something else. If you feel like putting in a PR with the tests we can merge either before or after the release.

oznetmaster:
The last time this happened, it had been noted that the proper solution might be to put commandlinetests into the common tests.  I thought that was a PR already?

CharliePoole:
Since we're planning on separating the framework and engine+console into different repo, we will need to either eliminate common code or find another approach to it.

oznetmaster:
The problem with that is the fact that nunitlite and console share a lot of common code.  Having to maintain two versions of the same source file is a real pain :(

oznetmaster:
They currently share 14 source files, plus the tests for those classes

CharliePoole:
Yes, we will have to deal with that. See #1589.

rprouse:
@oznetmaster, I agree that it is a pain. There are actually 3 sets of many of these files now with the new .NET Core runner, but I think we need to initially accept the duplication as we separate the runners and the framework into their own repositories. We really want to decouple them and get them on their own release cycles. Once we do that, we can start figuring out ways to reduce the duplication, possibly through a new shared dll?

PR #1557, opened 2016-06-10
Lazy load or hard code constraint DisplayNames

### LINKED ISSUE #1555: EndsWith calls in Constraint constructor can cause major perf issues
This is really due to a CoreCLR issue (which I've [filed](https://github.com/dotnet/coreclr/issues/5612)) but the constructor for [`Constraint`](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Constraints/Constraint.cs) performs three `EndsWith` checks:

```
DisplayName = this.GetType().Name;
if (DisplayName.EndsWith("`1") || DisplayName.EndsWith("`2"))
    DisplayName = DisplayName.Substring(0, DisplayName.Length - 2);
if (DisplayName.EndsWith("Constraint"))
    DisplayName = DisplayName.Substring(0, DisplayName.Length - 10);
```

That's nice and quick when the current culture is en-US and the invariant culture, but slow as molasses under other cultures. When pretty much every assertion (as far as I can see) creates a new constraint, that's far from ideal.

Firstly, given that this is going to be a fixed value per constraint type anyway, I wonder whether it would make sense to cache this or let the caller pass it in - but even if the code is broadly kept the same, just changing this to use `EndsWith(..., StringComparison.Ordinal)` would make it even faster than the en-US codepath.

rprouse:
Fixes #1555 

First, I applied the quick fix recommended by @jskeet and use `StringComparison.Ordinal` when constructing the `DisplayName`. I walked out from the `DisplayName` and as far as I can tell, it is only used in our tests, but I am reluctant to change a published API, so I changed it to be lazy loaded on the base `Constraint` class and hard coded on constraints that override it.

This results in a breaking change for any custom constraints that expect the base `Constraint.DisplayName` to be settable. They will have to override `DisplayName` instead. `IConstraint.DisplayName` only has a getter though, so it is not breaking the API.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/AllItemsConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/AllItemsConstraint.cs
@@ -39,10 +39,16 @@ namespace NUnit.Framework.Constraints
         public AllItemsConstraint(IConstraint itemConstraint)
             : base(itemConstraint)
         {
-            this.DisplayName = "All";
-            this.descriptionPrefix = "all items";
+            DescriptionPrefix = "all items";
         }
 
+        /// <summary> 
+        /// The display name of this Constraint for use by ToString().
+        /// The default value is the name of the constraint with
+        /// trailing "Constraint" removed. Derived classes may set
+        /// this to another name in their constructors.
+        /// </summary>
+        public override string DisplayName { get { return "All"; } }
 
         /// <summary>
         /// Apply the item constraint to each item in the collection,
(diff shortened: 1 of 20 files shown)


CharliePoole:
Looks good to merge to me, once the CI passes.

I think we __could__ take it further, but there is no pressing need. My reasoning is that IConstraint is an Extensibility interface. That is, it's an interface users creating constraints need to implement. It's not an interface we provide for users to use in any other way. However, after your change, it's not really worth our while to spend time on this.

InPermutation:
You could also avoid using strings by replacing 

    displayName.EndsWith("`1", ...)

with 

    this.GetType().IsGenericType

asbjornu:
@InPermutation Implemented in #1568. Thanks for the suggestion! :smile:

PR #1561, opened 2016-06-10
Don't write Settings unnecessarily

### LINKED ISSUE #1560: Engine writes setting file unnecessarily
Issue nunit/nunit3-vs-adapter#190 reports a case where multiple processes are trying to write the settings file. The problem can be at least reduced if the settings file is only written when it needs to be. The engine is writing the file whether there has been a change or not, even when there are no settings and the file is empty.

Currently, the only time the engine changes the settings is when recent files are saved. This is only done by the gui. So the vs adapter and console runner should never trigger a write. And an empty file should never be created, since it serves no purpose.

CharliePoole:
Fixes #1560

### DIFF
--- a/src/NUnitEngine/nunit.engine/Internal/SettingsStore.cs
+++ b/src/NUnitEngine/nunit.engine/Internal/SettingsStore.cs
@@ -96,38 +96,46 @@ namespace NUnit.Engine.Internal
 
         public void SaveSettings()
         {
-            if (_writeable)
+            if (_writeable && _settings.Keys.Count > 0)
             {
-                string dirPath = Path.GetDirectoryName(_settingsFile);
-                if (!Directory.Exists(dirPath))
-                    Directory.CreateDirectory(dirPath);
-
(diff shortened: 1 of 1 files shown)


src/NUnitEngine/nunit.engine/Internal/SettingsStore.cs
```diff
@@ -96,38 +96,46 @@ public void LoadSettings()
 
         public void SaveSettings()
         {
-            if (_writeable)
+            if (_writeable && _settings.Keys.Count > 0)
```
rprouse: Is there ever a case where the settings file contains settings, but this run we want to write a new version with no settings? If not, then this looks good.

CharliePoole: No current case. In future we might have a settings editor, but that's not a sure thing. I think we may eventually convert to json and let people edit with their favorite editor.

PR #1563, opened 2016-06-10
Clean common

CharliePoole:
This is a substantial but incomplete refactoring of our common files.

1. Files not actually shared were moved to the framework.
2. Compatibility files using system namespaces were placed in the namespace subfolders.
3. I changed the standard practice of including all files in all framework builds for those files in the Compatibility folder. It's confusing to see an alternate implementation included when it's not needed.

Much more needs to be done on this but it can't go much further until the portable driver, which shares about 20 files, is moved elsewhere.


CharliePoole:
Everything builds now... @rprouse see if you like the structure as I've set it up. We can move it around if we need to.

rprouse:
This looks good. I will need to update `dotnet-test-nunit` to use the new namespaces, but that will be easy enough. For the portable agent, if we change it to `netstandard` from PCL, it can become more of a proper engine because of the added classes. That said, we likely can't use a lot of the code because the .NET 2.x `XmlNode` classes are not available anymore in favor of the newer Linq to XML classes. It really is too bad Linq to XML isn't available on 2.0, they are much nicer to work with.

Because of that, I don't think there will be as much code sharing between the engine and the portable.agent as I had hoped, so it can probably be moved. It does use a lot of compatibility classes though. To move it, are you thinking we should duplicate the code across repositories? I was actually hoping to go the other way and merge the three `ResultSet` classes that are currently so similar.

Still not sure how we should approach it though. What were your ideas?

CharliePoole:
For the framework, I wrote the TNode class to deal with the lack of availability of XmlNode on all platforms. Would that be of any help to you?

I'd say create the duplication and then we can figure out how to remove it. Trying to __avoid__ duplication is at least an order of magnitude harder than removing it!

Let's at least consider the option of git submodules, which we have avoided up to now.

Regarding the portable agent, I think we should do a serious "whiteboard" session by phone one of these days - perhaps after you get out the netcore runner. I don't think the agent really should try to become an engine. Rather, I think it should become one half of what I am calling a "split" driver, where one piece of the driver lives in the actual engine and another piece is injected into wherever (an AppDomain, Process, machine, device) the actual test is running. What we currently do in the engine is actually a degenerate instance of this, where FrameworkController forms the bottom half of the driver and is bundled with the framework itself. This PR is probably not the best place to conduct the discussion, however. :-)

CharliePoole:
Did you mean `ResultSummary` classes? I agree that something should be done. As code, they are trivial and so could easily live on duplicated. But they incorporate logic that sets policies about what it means to pass or fail, how we figure out return codes, etc., which should really be central.

This PR is really about low-hanging fruit. Most of what I removed from common was not actually used across multiple modules, but was the result of over-enthusiastic generalization in advance of need.

Are you ready to merge this and deal with the fallout later? :-)

rprouse:
Yes, merging, we can discuss the overall architecture elsewhere.

Issue #1553, opened 2016-06-08
Allow ordering tests to be done in multiple properties


labilbe:
On SQL queries, I tend to order by multiple columns and not only one.
I would like NUnit handles this case and allows unit testing not only on one property but on multiple properties.

The general idea is to be able to do the following:

`var list = new List<Tuple<string, string>>`
`{`
`    new Tuple<string, string>("aaaa", "bbbb"),`
`    new Tuple<string, string>("bbbb", "aaaa"),`
`    new Tuple<string, string>("aaaa", "cccc"),`
`    new Tuple<string, string>("cccc", "bbbb")`
`};`
`var ordered = list.OrderBy(p => p.Item1).ThenBy(p => p.Item2);`

`Assert.That(ordered, Is.Ordered.By("Item1").By("Item2"));`
or
`Assert.That(ordered, Is.Ordered.By("Item1").And.By("Item2"));`
or
`Assert.That(ordered, Is.Ordered.By("Item1").ThenBy("Item2"));`

I was looking for this feature on [SO](http://stackoverflow.com/questions/37676739/how-to-check-list-orders-further-properties-with-nunit/) and Charlie Poole sent me here.

Thank you.

CharliePoole:
Thanks for filing this. 

The And.By option would require `And` to have different meanings based on context. We try to avoid that although there are a few places where it happens.

Just repeating `By` would work but I like `ThenBy` because it makes clear the order of application.

What do others think?

ChrisMaddock:
I would agree with `ThenBy` - it matches up to LINQ behaviour.

We'd get some use out of this functionality too, we do a few similar tests. I've always got round it the lazy way - by ordering my own list, then asserting that `SequenceEqual()`.

CharliePoole:
Looks pretty straightforward. All code is in the constraint class. It needs a new modifier and should save a list of property names rather than just one name. We could avoid checking the sequence of By, ThenBy... by letting them both do the same thing, or we could check... By would expect an empty list, ThenBy a non-empty one. The hardest part would be to construct a comparator using the list of properties, but not so hard as all that really. :-)

jnsn:
I created a pull request for this. The pull requests adds 3 additional methods to accommodate the requested feature:

- `ByDescending`
- `ThenBy`
- `ThenByDescending`

The descending methods were added to match the LINQ behavior and to allow for the existing `Descending` constraint to keep existing and not depend on whether or not one or more properties are added to the sort requirement.

[...]

PR #1564, opened 2016-06-11
Implemented multiple properties sort for #1553.

jnsn:
This pull requests adds 3 additional methods to accommodate the requested feature:

- `ByDescending`
- `ThenBy`
- `ThenByDescending`

The descending methods were added to match the LINQ behavior and to allow for the existing `Descending` constraint to keep working and not depend on whether or not one or more properties are added to the sort requirement.

As this is my first pull request, any feedback is appreciated.


src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
```diff
@@ -63,6 +65,10 @@ public CollectionOrderedConstraint Descending
             get
             {
                 descending = true;
+
+                if (this.properties != null && this.properties.Count != 0)
+                    this.properties[this.properties.Keys.Max()] = true;
+
```
CharliePoole: I suspect you have changed behavior of `Is.Ordered.Descending.By((SomeProp")` here. If so, we're all at fault because there's no test of it. Can you add one?

src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
```diff
@@ -103,7 +109,55 @@ public CollectionOrderedConstraint Using<T>(Comparison<T> comparer)
         /// </summary>
         public CollectionOrderedConstraint By(string propertyName)
         {
-            this.propertyName = propertyName;
+            if (this.properties != null && this.properties.Count != 0)
+                throw new InvalidOperationException("Cannot use By on existing collection. Use ThenBy to add additional items.");
```
CharliePoole: Why not? Seems like we can figure out what is intended.

Even if we want to forbid it, the message needs to be more user oriented. The user knows nothing about the internals of our implementation, so it would have to say something like "Cannot use By more than once" 

However, I see no reason to forbid it, since it's clear what the user wanted to do. Seems as if By and ThenBy can actually work just the same, with the two spellings simply serving to make things more readable.

src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
```diff
@@ -103,7 +109,55 @@ public CollectionOrderedConstraint Using<T>(Comparison<T> comparer)
...
+        }
+
+        /// <summary>
+        /// Modifies the constraint to test descending ordering by the value of
+        /// a specified property and returns self.
+        /// </summary>
+        public CollectionOrderedConstraint ByDescending(string propertyName)
+        {
+            if (this.properties != null && this.properties.Count != 0)
+                throw new InvalidOperationException("Cannot use ByDescending on existing collection. Use ThenByDescending to add additional items.");
```
CharliePoole: As above..

src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
```diff
@@ -103,7 +109,55 @@ public CollectionOrderedConstraint Using<T>(Comparison<T> comparer)
...
+        }
+
+        /// <summary>
+        /// Modifies the constraint to test additional ordering by
+        /// the value of a specified property and returns self.
+        /// </summary>
+        public CollectionOrderedConstraint ThenBy(string propertyName)
+        {
+            if (this.properties == null || this.properties.Count < 1)
+                throw new InvalidOperationException("Cannot use ThenBy on an empty collection.");
```
CharliePoole: Again, as above.

src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
```diff
@@ -103,7 +109,55 @@ public CollectionOrderedConstraint Using<T>(Comparison<T> comparer)
...
+        }
+
+        /// <summary>
+        /// Modifies the constraint to test additional descending ordering by
+        /// the value of a specified property and returns self.
+        /// </summary>
+        public CollectionOrderedConstraint ThenByDescending(string propertyName)
+        {
+            if (this.properties == null || this.properties.Count < 1)
+                throw new InvalidOperationException("Cannot use ThenByDescending on an empty collection.");
```
CharliePoole: And finally, here again.

src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
```diff
@@ -37,7 +39,7 @@ public class CollectionOrderedConstraint : CollectionConstraint
     {
         private ComparisonAdapter comparer = ComparisonAdapter.Default;
         private string comparerName;
-        private string propertyName;
+        private Dictionary<string, bool> properties;
         private bool descending;
```
CharliePoole: This code has not been worked on for a while so it doesn't meet our current standards. All memver fields should get an initial underscore and you should drop `this.` in front of them. However, I suggest saving this for a final pass to make it easier to review the "real" changes.

src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
```diff
@@ -114,13 +168,33 @@ public CollectionOrderedConstraint By(string propertyName)
...
+                }
+                else
+                {
+                    var first = this.properties.ElementAt(0);
+                    desc = "collection ordered by " + MsgUtils.FormatValue(first.Key);
+
+                    if (first.Value)
+                        desc += ", descending";
+
+                    for (int i = 1; i < this.properties.Count; i++)
```
CharliePoole: Are we guaranteed that all dictionary entries will be returned in the same order we added them? Does this guarantee hold across all platforms? If not, the dictionary should probably be replaced by a list of pairs or a pair of lists.

CharliePoole:
Nice work! I generally like it with a few comments on the details.

Two general comments we should resolve before merging:

1. How rigid versus forgiving should we be regarding the syntax? Your implementation wants things to come in exactly the right order or it throws an exception. As commented inline, I think we can be a bit looser than that in what we accept, although I'm willing to listent ot arguments.

2. What is the interation between the old Descending and the new [Then]ByDescending? A few thoughts:
 * By plus Descending and Descending plus By have always done the same thing. They should keep doing it.
 * Descending followed by multiple By entries should probably change them all to descending in order to remain consistent. Does that make sense to you?
 * Descending in the midst of a group of Bys should apply only to the preceding one, just as you have implemented it.
 * How we handle a mix of Descending and [Then]ByDescending is worthy of discussion. It could lead to confusion. Should we forbid it?

CharliePoole:
@rprouse If we can resolve all of the questions in time, it would be nice to put this in 3.4. I think it's important to resolve them, however.

CharliePoole:
Fixes #1553 

@jnsn Your PRs should always have something like the above in the description. It causes the merge to automatically close the issue.

src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
jnsn: Yes, 

I initially removed the `Descending` back-end field and did it all with the dictionary. But that broke the test where no sort property was given and it was just testing if a list was sorted descending. 

So I decoupled the functionality and (incorrectly) assumed that if the `Descending` is called first, no sort properties are given later on. 
It should work if it's called in the -`Is.Ordered.By("SomeProp").Descending.ThenBy("AnotherProp").Descending` order. 

I will add another test to handle the current behavior.

src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
jnsn: It can be allowed, but we must be careful to only take the status of the `Descending` property into account if there's no item in the collection yet.

If someone would call `Is.Ordered.Descending.By("SomeProp").By("AnotherProp")` the descending status should only be set to the first property and discarded for all subsequent calls.

jnsn:
I've added some inline remarks and will address the issues shortly.

CharliePoole:
I'm coming around to the view that `ByDescending` and `ThenByDescending` are too much.

1. The user can get the same behavior by using `By().Descending` or `ThenBy().Descending`
2. Without them, we don't have to worry about things like `ByDescending().Descending` or `Descending.ByDescending`.

CharliePoole:
There remains the question of how this behaves...

```C#
Is.Ordered.Descending.By("Prop1").ThenBy("Prop2")
```

@jnsn argues that the `Descending` should only apply to the first item. I'm inclined to apply it to all of them and give a warning if any further instances of `Descending` appear. What do you think @nunit/core-team ?

src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
CharliePoole: @jnsn My research indicates that the order is not guaranteed to be anything in particular, only that the order of keys and values will match one another. This probably needs to be re-implemented independent of whatever we decide for the user interface.

CharliePoole:
@jnsn Let's move ahead with this. Use last two comments on the issue as a guide.... in summary...

1. Modifiers: By, Then, Ascending, Descending.
2. Any order is allowed so long as we don't get By+Descending+By or By+Ascending+By

I'm happy to wait for you to do it and review or to give you suggestions about how to do it, whichever you prefer. I'd like #2 to be detected at compile time, which will involve some extra work. If you aren't sure how to do that, leave it for last.

Let me know how you  want to proceed. If we can get it done by the weekend it can probably go into the 3 .4 release.

jnsn:
@CharliePoole I will pick it up when I get home from work later tonight and see how far I get.

jnsn:
@CharliePoole I'm not sure how to implement this correctly.

I can achieve something along the lines of what's proposed, but it would only notify the user of issues at run-time, not at compile-time. I did a quick first draft in commit 8e603f9 to get a feel on how it would work and look.

To achieve this at compile-time, I've been looking at the `With` modifier (and by extend how the entire constraint building system works) hoping it would clarify the intended implementation.

I was thinking of implementing three additional constraints, inheriting from the `BinaryConstraint` in a similar way as the `AndConstraint` works, and then use the `Matches` method of the `CollectionOrderedConstraint` to get the list of operations (by, then, descending, querying tall the builders of the right-side constraint) and build a result with that.

But I'm not sure whether that's the way to go, so if you can guide me on the correct path, that would be greatly appreciated.

CharliePoole:
A general comment in advance: 

Disallowing the invalid combo dynamically at run time and disallowing it at compile time require entirely different approaches. So the first step towards doing it would involve __removing__ the run time checks you have put in.

For the first pass, I'll just comment on details of what you coded.

src/NUnitFramework/tests/Constraints/CollectionOrderedConstraintTests.cs
```diff
@@ -227,6 +227,46 @@ public void IsOrderedBy_Handles_heterogeneous_classes_as_long_as_the_property_is
...
+        [Test]
+        public void IsOrderedByDescendingThenByDescending()
+        {
+            var al = new List<OrderedByThenByTestClass>();
+            al.Add(new OrderedByThenByTestClass("DEF", 2));
+            al.Add(new OrderedByThenByTestClass("ABC", 1));
+
+            Assert.That(al, Is.Ordered.By("A").Descending.Then.By("B").Descending);
+        }
+
```
CharliePoole: Need cases for...
 * Ordered.Descending.By
 * Ordered.By.Descending
 * Ordered.Descending.By.Then.Descending.By
   etc.

src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
```diff
@@ -63,6 +65,25 @@ public CollectionOrderedConstraint Descending
...
                 descending = true;
+
+                if (!this.allowBy && this.properties != null && this.properties.Count != 0)
+                {
+                    this.properties[this.properties.Count - 1].SortDescending = true;
+                    descending = false;
+                }
+
+                return this;
+            }
```
CharliePoole: Also need Ascending

src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
```diff
@@ -103,7 +124,12 @@ public CollectionOrderedConstraint Using<T>(Comparison<T> comparer)
         /// </summary>
         public CollectionOrderedConstraint By(string propertyName)
         {
-            this.propertyName = propertyName;
+            if (!this.allowBy) 
+                throw new InvalidOperationException("The use of 'By' is not allowed.");
+
+            this.properties.Add(new OrderedPropertyInfo(propertyName, this.descending));
+            this.allowBy = false;
```
CharliePoole: I think this is incorrect. At least I don't see anything wrong with Ordered.By.By. However Ordered.By.Descending.By should be wrong. IMO, you should entirely remove the allowBy thing and get this working first without that error check. Once we have the right overall shape, then we can deal with the edge case.

CharliePoole:
@jnsn  I'm going to start fresh on a branch, pulling in your stuff and adding some of my own. Let's see what I come up with and how it compares.

jnsn:
@CharliePoole OK.

I did some other work, implementing the `Ascending`, which just resets the internal `descending` boolean to `false` so that when adding new items it sorts as ascending. I haven't pushed it yet because it didn't work as I wanted. Unless I use quite some internal booleans to check the state and throw exceptions, I couldn't figure out how to block the unwanted behavior.

CharliePoole:
So I tried to _start_ with different classes inheriting from one another in typical fluent style. That got very complicated because we have three different variables to deal with:

 * Ascending versus descending
 * By
 * Using

Each of these should really only occur once, either in total or per column. That means you need a total of seven classes to reflect what should be allowed at any point and many of those classes could end up with duplicate code.

So, I'm going back to the runtime check approach, with either a single class or maybe two  for Ordered versus OrderedBy. It may be possible to build very simple adapters on top of them that do compile time checking.

For ascending/descending I switched from a boolean to an enum: Unspecified, Ascending, Descending. That allows me to check more easily.

I'll try to have something to look at today.

CharliePoole:
OK, I created PR #1604 based on your work here. I think it fixes a few things and it has better tests, but there is more work to do on it.

Feel free to take off from my branch or combine with yours. Let us know what you plan to do. I won't be coding on it any more for a while in any case as other things have taken priority.

CharliePoole:
With 93 files changed, this is getting hard to review. It's what happens in GitHub when the master is changing underneath you. I think you need to wait for the 3.4 release to merge back into master and then rebase on master and do a forced push. @rprouse do you think that's the right approach?

jnsn:
Ah, OK, I can do that.

Basically, right now, I just merged your pull request into my fork and discarded all my previous changed. 

If wanted, I could also close this PR and open a new one when I've got something worth reviewing.

CharliePoole:
My PR was based on a much later master than yours, so I suppose all those things showed up as changes. Eventually, if we merged it, Git would sort it all out. It's just hard to look at the PR online.

If it's easy (might not be) to rebase locally, I would do that. Alternatively, I can give you authority to work in my branch if that's actually the code you are using.

CharliePoole:
@jnsn Have you decided how you want to move this forward?

CharliePoole:
This issue needs to be closed or merged before we can split the repositories per #1598.

It's a little confusing actually. The issue seems like it should only affect the framework but there are changes in the engine and console as well.

jnsn:
@CharliePoole appoligies, but I simply did not have the time yet to look at this any further.

CharliePoole:
@jnsn Let's figure out how to revive this! With work done by people on their own time, it's not unusual for something to start out and then run into time issues that stall it. If you are not able to carry the ball further, someone else should be able to take it to the next stage.

That will require some review of what's been done - it hasn't been on my mind either - and what remains, so let's start with that.

As for this particular PR. I think it can probably be closed, but I'll look at it a bit move closely today to be sure. I think the appearance that so much has changed is due to merging in other changes from master over time. It may be that the branch and PR I created encapsulates most of the essentials from this change. More later today...

jnsn:
@CharliePoole you are correct. Basically what I did was revert all my changes, pull your PR and pull in the latest changes from master. Therefore anyone should be able to start from your PR and have the essential bits.

CharliePoole:
In that case, I'll close this and take another look at the issue on PR #1604. I hope you'll follow along and maybe even get back into it.

PR #1565, opened 2016-06-12
Use NUnit.System.Linq for .NET 2.0 build

### LINKED ISSUE #1547: Create separate assembly for System.Linq compatibility classes
Our .NET 2.0 build includes mono classes from System.Linq namespace. It's not the full set AFAIK but it helps us write cleaner, more consistent code across the various builds. 

These classes are not referenced anywhere else right now, but they would be useful in the engine, which has to be built for 2.0 and possibly in the adapter, which we changed from 2.0 to 3.5 in order to use System.Linq.

We should investigate putting these in a separate assembly and repository and creating a package. It would essentially be a package for our own use, although others could download it. Once it has everything we want in it, I would  expect it to remain unchanged for long periods, however, we would want to version the packages just in case.

@nunit/core-team What do you think?

CharliePoole:
Fixes #1547


CharliePoole:
I added deployment to MyGet here so as to avoid a merge conflict later. The next build should __not__ anything to our MyGet feed, but once this is merged, it should rebuild master and the packages should be deployed there.

ChrisMaddock:
@CharliePoole - Minor point: adding the myget feed as a required package source does now mean the solution can't be built in VS straight off.

I know the cake file is intended to be used for building, but would it be worth adding a solution level nuget.config with the myget package source, so it can still be build directly in VS? 
Such as: http://stackoverflow.com/q/33831689/1768779

rprouse:
@ChrisMaddock that is my fault for not catching it in the merge. I know better :frowning: 

Here is NuGet.config that will work. I will create a PR and add.

```
<configuration>
  <packageSources>
    <clear/>
    <add key="CI Builds (AppVeyor)" value="https://ci.appveyor.com/nuget/nunit" />
    <add key="Internal Builds (MyGet)" value="https://www.myget.org/F/nunit/api/v2" />
    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
  </packageSources>
</configuration>
```

CharliePoole:
So guys... educate me here please... I as able to build in VS - I just selected the MyGet source in the IDE. What am I missing?

rprouse:
@CharliePoole did you add the MyGet source in the Visual Studio options? It shouldn't be in the dropdown unless you added it or it is in a higher level NuGet.config.

I am testing now...

rprouse:
Confirmed, doesn't work for me. PR to follow shortly.

CharliePoole:
I did add it in the VS options. Doesn't it persist that for the solution?

rprouse:
No, those are just personal, they are saved in the Visual Studio settings.

ChrisMaddock:
Cheers @rprouse. :-)

CharliePoole:
So if we add the config, should we remove the explicit reference to the two repositories in build.cake?

rprouse:
@CharliePoole I think we can remove those explicit references, but I haven't.

PR #1568, opened 2016-06-13
Use Type.IsGenericType instead of String.EndsWith

asbjornu:
This PR implements @InPermutation's suggestion in #1557 to use `Type.IsGenericType` instead of:

```c#
displayName.EndsWith("`1", StringComparison.Ordinal) || displayName.EndsWith("`2", StringComparison.Ordinal)
```

I don't know how well `Constraint.DisplayName` is tested, but I hope that since everything's green after running `./build -t TestAll`, I haven't broken anything. If I have, please let me know.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Constraint.cs
+++ b/src/NUnitFramework/framework/Constraints/Constraint.cs
@@ -54,8 +54,9 @@ namespace NUnit.Framework.Constraints
 
             _displayName = new Lazy<string>(() =>
             {
-                var displayName = this.GetType().Name;
-                if (displayName.EndsWith("`1", StringComparison.Ordinal) || displayName.EndsWith("`2", StringComparison.Ordinal))
+                var type = this.GetType();
+                var displayName = type.Name;
+                if (type.IsGenericType)
                     displayName = displayName.Substring(0, displayName.Length - 2);
                 if (displayName.EndsWith("Constraint", StringComparison.Ordinal))
                     displayName = displayName.Substring(0, displayName.Length - 10);
--- a/src/NUnitFramework/framework/Constraints/Constraint.cs
+++ b/src/NUnitFramework/framework/Constraints/Constraint.cs
@@ -22,8 +22,10 @@
 // ***********************************************************************
 
 using NUnit.Framework.Internal;
+using NUnit.Compatibility;
 using System.Collections;
 using System;
+using System.Reflection;
 
 namespace NUnit.Framework.Constraints
 {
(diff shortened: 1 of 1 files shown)


rprouse:
The portable version of the build is failing because 

```
=======================================
Executing task: BuildPortable
Microsoft (R) Build Engine version 14.0.25123.0
Copyright (C) Microsoft Corporation. All rights reserved.
 
Constraints\Constraint.cs(59,26): error CS1061: 'Type' does not contain a definition for 'IsGenericType' and no extension method 'IsGenericType' accepting a first argument of type 'Type' could be found (are you missing a using directive or an assembly reference?) [C:\projects\nunit\src\NUnitFramework\framework\nunit.framework-portable.csproj]
An error occured when executing task 'BuildPortable'.
```

You likely need to switch to  `.GetTypeInfo().IsGenericType`  to get it working. On platforms that don't support it, we have just defined `GetTypeInfo()` as an extension method that returns the `Type`, so you can use it everywhere. You might just need to pull in the NUnit.Compatibility namespace.

src/NUnitFramework/framework/Constraints/Constraint.cs
```diff
@@ -54,8 +54,9 @@ protected Constraint(params object[] args)
 
             _displayName = new Lazy<string>(() =>
             {
-                var displayName = this.GetType().Name;
-                if (displayName.EndsWith("`1", StringComparison.Ordinal) || displayName.EndsWith("`2", StringComparison.Ordinal))
+                var type = this.GetType();
+                var displayName = type.Name;
+                if (type.IsGenericType)
```
rprouse: Yah, you will need to switch this to `type.GetTypeInfo().IsGenericType`.

CharliePoole: But was the portable build passing on your Mac? I so, then there may be more here than meets the eye.

rprouse: In the other PR, he mentions that he was building using Cake from the command line. On a Mac, IsRunningOnWindows is obviously false which means it will build as if it is Linux which does not include the portable build.

CharliePoole: Makes sense.

asbjornu:
@rprouse Thanks for the feedback! I've added the `GetTypeInfo()` extension method and it seems like I've stumbled upon [the problem @CharliePoole mentioned on Gitter](https://gitter.im/cake-build/cake?at=575f45f097e1b2d245e293fe):

```
WARNING: Error getting response stream (Write: The authentication or decryption has failed.): SendFailure
Unable to find version '0.6.4' of package 'Cake'.
Could not find Cake.exe.
```

Everything's green locally, though.

rprouse:
Looks good. I am going to override and merge now that AppVeyor is passing and the Travis latest is.

asbjornu:
@rprouse Thanks! :+1:

PR #1570, opened 2016-06-14
Allow passing arguments into the test run

### LINKED ISSUE #720: Need a way to get test-specific command-line arguments at runtime
We are trying to implement NUnit in our project and hit a road block because of 2 things:

1. There is no way in which we can pass any parameter from command line to the test called.
2. We tried exploring "Category" feature if that can help in achieving the above point but again hit a roadblock as there is no way in which I can get the list of Categories passed using console runner.

Considering the below scenario, if I can get the list of all Categories which are passed during execution then I can write my own logic in code. This can also work as a minimal way of achieving the command line custom parameter passing (as explained in point#1).

[Test]
[Category("PDF")]
[Category("DOC")]
Public void SomeTest()
{
       string[] myCategories=TestContext.CurrentContext.Test.SelectedCategories;

            if (myCategories.Contains("PDF"))
            {
                ParsePDF();
            }
            else
            {
                if(myCategories.Contains("DOC")))
                {
                    ParseDoc();
                }
                else
                {
                    ParseHTML();
                }
            }
}

Proposed Syntax:
TestContext.CurrentContext.Test.SelectedCategories
- Should return list of string having each selected categories.

CharliePoole:
Fixes #720

I implemented it in the simplest way possible. Everything is a string until you decide to fetch it as an int, at which time it gets converted.

The Mono 3.2.8 failure in travis seems to be an infrastructure problem. I confirmed that by rerunning a successful build of master from about five hours ago and getting the same failure.

### DIFF
--- a/src/Common/nunit/CommandLineOptions.cs
+++ b/src/Common/nunit/CommandLineOptions.cs
@@ -86,6 +86,8 @@ namespace NUnit.Common
         private List<string> testList = new List<string>();
         public IList<string> TestList { get { return testList; } }
 
+        public string TestParameters { get; private set; }
+
         public string WhereClause { get; private set; }
         public bool WhereClauseSpecified { get { return WhereClause != null; } }
 
@@ -295,6 +297,23 @@ namespace NUnit.Common
 #endif
             this.Add("where=", "Test selection {EXPRESSION} indicating what tests will be run. See description below.",
                 v => WhereClause = RequiredValue(v, "--where"));
+
+            this.Add("params|D=", "Define a test parameter.",
+                v =>
+                {
+                    string parameters = RequiredValue( v, "--params");
+
+                    foreach (string param in parameters.Split(new[] { ';' }))
+                    {
+                        if (!param.Contains("="))
+                            ErrorMessages.Add("Invalid format for test parameter. Use NAME=VALUE.");
+                    }
+
+                    if (TestParameters == null)
+                        TestParameters = parameters;
+                    else
+                        TestParameters += ";" + parameters;
+                });
 
             this.Add("timeout=", "Set timeout for each test case in {MILLISECONDS}.",
                 v => defaultTimeout = RequiredInt(v, "--timeout"));
--- a/src/Common/nunit/CommandLineOptions.cs
+++ b/src/Common/nunit/CommandLineOptions.cs
@@ -298,7 +298,7 @@ namespace NUnit.Common
             this.Add("where=", "Test selection {EXPRESSION} indicating what tests will be run. See description below.",
                 v => WhereClause = RequiredValue(v, "--where"));
 
-            this.Add("params|D=", "Define a test parameter.",
+            this.Add("params|p=", "Define a test parameter.",
                 v =>
                 {
                     string parameters = RequiredValue( v, "--params");
--- a/src/NUnitFramework/framework/TestParameters.cs
+++ b/src/NUnitFramework/framework/TestParameters.cs
@@ -44,17 +44,23 @@ namespace NUnit.Framework
         /// <returns>Value of the paramter or null if not present</returns>
         public string Get(string name)
         {
-            return _parameters.ContainsKey(name) ? _parameters[name] : null;
+            return this[name];
         }
 
         public string Get(string name, string defaultValue)
         {
-            return _parameters.ContainsKey(name) ? _parameters[name] : defaultValue;
+            return this[name] ?? defaultValue;
         }
-        
-        public int Get(string name, int defaultValue)
+
+        public int GetInt(string name)
         {
-            return _parameters.ContainsKey(name) ? int.Parse(_parameters[name]) : defaultValue;
+            return GetInt(name, 0);
+        }
+
+        public int GetInt(string name, int defaultValue)
+        {
+            string val = this[name];
+            return val != null ? int.Parse(val) : defaultValue;
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/TestParameters.cs
+++ b/src/NUnitFramework/framework/TestParameters.cs
@@ -28,7 +28,7 @@ namespace NUnit.Framework
         }
 
         /// <summary>
-        /// Indexer is the only public access to the internal dictionary
+        /// Indexer provides access to the internal dictionary
         /// </summary>
         /// <param name="name">Name of the parameter</param>
         /// <returns>Value of the parameter or null if not present</returns>
@@ -41,22 +41,39 @@ namespace NUnit.Framework
         /// Get method is a simple alternative to the indexer
         /// </summary>
         /// <param name="name">Name of the paramter</param>
-        /// <returns>Value of the paramter or null if not present</returns>
+        /// <returns>Value of the parameter or null if not present</returns>
         public string Get(string name)
         {
             return this[name];
         }
 
+        /// <summary>
+        /// Get the value of a parameter or a default string
+        /// </summary>
+        /// <param name="name">Name of the parameter</param>
+        /// <param name="defaultValue">Default value of the parameter</param>
+        /// <returns>Value of the parameter or default value if not present</returns>
         public string Get(string name, string defaultValue)
         {
             return this[name] ?? defaultValue;
         }
 
+        /// <summary>
+        /// Get the int value of a parameter or zero
+        /// </summary>
+        /// <param name="name">Name of the parameter</param>
+        /// <returns>Value of the parameter or default value if not present</returns>
         public int GetInt(string name)
         {
             return GetInt(name, 0);
         }
 
+        /// <summary>
+        /// Get the int value of a parameter or a default int
+        /// </summary>
+        /// <param name="name">Name of the parameter</param>
+        /// <param name="defaultValue">Default value of the parameter</param>
+        /// <returns>Value of the parameter or default value if not present</returns>
         public int GetInt(string name, int defaultValue)
         {
             string val = this[name];
--- a/src/NUnitFramework/framework/TestParameters.cs
+++ b/src/NUnitFramework/framework/TestParameters.cs
@@ -1,0 +1,70 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace NUnit.Framework
+{
+    /// <summary>
+    /// TestParameters class holds any named parameters supplied to the test run
+    /// </summary>
+    public class TestParameters
+    {
+        private readonly Dictionary<string, string> _parameters = new Dictionary<string, string>();
+
+        /// <summary>
+        /// Gets the number of test parameters
+        /// </summary>
+        public int Count
+        {
+            get { return _parameters.Count; }
+        }
+
+        /// <summary>
+        /// Gets a collection of the test parameter names
+        /// </summary>
+        public ICollection<string> Names
+        {
+            get { return _parameters.Keys; }
+        }
+
+        /// <summary>
+        /// Indexer is the only public access to the internal dictionary
+        /// </summary>
+        /// <param name="name">Name of the parameter</param>
+        /// <returns>Value of the parameter or null if not present</returns>
+        public string this[string name]
+        {
+            get { return _parameters.ContainsKey(name) ? _parameters[name] : null; }
+        }
+
+        /// <summary>
+        /// Get method is a simple alternative to the indexer
+        /// </summary>
+        /// <param name="name">Name of the paramter</param>
+        /// <returns>Value of the paramter or null if not present</returns>
+        public string Get(string name)
+        {
+            return _parameters.ContainsKey(name) ? _parameters[name] : null;
+        }
+
+        public string Get(string name, string defaultValue)
+        {
+            return _parameters.ContainsKey(name) ? _parameters[name] : defaultValue;
+        }
+        
+        public int Get(string name, int defaultValue)
+        {
+            return _parameters.ContainsKey(name) ? int.Parse(_parameters[name]) : defaultValue;
+        }
+
+        /// <summary>
+        /// Adds a parameter to the list
+        /// </summary>
+        /// <param name="name">Name of the parameter</param>
+        /// <param name="value">Value of the parameter</param>
+        internal void Add(string name, string value)
+        {
+            _parameters[name] = value;
+        }
+    }
+}
--- a/src/NUnitFramework/framework/TestParameters.cs
+++ b/src/NUnitFramework/framework/TestParameters.cs
@@ -28,13 +28,23 @@ namespace NUnit.Framework
         }
 
         /// <summary>
+        /// Gets a flag indicating whether a parameter with the specified name exists.N
+        /// </summary>
+        /// <param name="name">Name of the parameter</param>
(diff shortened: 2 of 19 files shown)


rprouse:
Looks good, but it is late, so I am going to take another look in the morning. I will also want to create an issue to add this to `dotnet-test-nunit`.

src/Common/nunit/CommandLineOptions.cs
```diff
@@ -296,6 +298,23 @@ protected virtual void ConfigureOptions()
             this.Add("where=", "Test selection {EXPRESSION} indicating what tests will be run. See description below.",
                 v => WhereClause = RequiredValue(v, "--where"));
 
+            this.Add("params|D=", "Define a test parameter.",
+                v =>
+                {
+                    string parameters = RequiredValue( v, "--params");
+
+                    foreach (string param in parameters.Split(new[] { ';' }))
```
rprouse: Based on #1571, should we be using semi-colons, or is there an easy workaround to that issue?

src/Common/nunit/CommandLineOptions.cs
```diff
@@ -296,6 +298,23 @@ protected virtual void ConfigureOptions()
             this.Add("where=", "Test selection {EXPRESSION} indicating what tests will be run. See description below.",
                 v => WhereClause = RequiredValue(v, "--where"));
 
+            this.Add("params|D=", "Define a test parameter.",
```
rprouse: `--params` and `-D` seem a bit confusing especially since we aren't using `-p` or `-P`.

CharliePoole: @oznetmaster had asked for -D but I guess you're right. In addition to being the wrong letter, it's the first one where we would be using a capital letter. No reason to do that. I suggest we either use `params|p=` or `define|d=`

@oznetmaster do you want to weigh in on this?

src/Common/nunit/CommandLineOptions.cs
CharliePoole: Let me try it out when I get home to my Linux machine. We have had  the result option for a long time and nobody has brought up the semicolon before.

src/Common/nunit/CommandLineOptions.cs
CharliePoole: @nunit/core-team @nunit/contributors Defining a new command-line option is kind of a big deal, so please weigh in with your opinions!

oznetmaster: I prefer --define and --d, but it is only a preference.

rprouse: I think we should expand the help text here a bit. Maybe specify the format and that the parameters will be available in the `TestContext.Parameters`?

src/NUnitFramework/framework/TestParameters.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            return _parameters.ContainsKey(name) ? _parameters[name] : null;
+        }
+
+        public string Get(string name, string defaultValue)
+        {
+            return _parameters.ContainsKey(name) ? _parameters[name] : defaultValue;
+        }
+        
+        public int Get(string name, int defaultValue)
```
rprouse: Do you think we should be more explicit with the names, `GetString` and `GetInt`? Most other C# API's follow that pattern.

CharliePoole: The underlying form returned by the indexer is string, so I figured GetString would be redundant. I thought about GetInt. I'm open to ideas - it would be great to get feedback from more folks here, since this forms part of our most important API - the one used by tests.

One option, of course, is to just toss out the Get methods and require the user to do any conversion.

src/Common/nunit/CommandLineOptions.cs
CharliePoole: I don't see a problem. I commented on #1571

src/Common/nunit/CommandLineOptions.cs
CharliePoole: Agreed.

CharliePoole:
I'll make final changes once we settle on the syntax.

src/Common/nunit/CommandLineOptions.cs
rprouse: I am okay with both forms, but I probably lean towards `--define|-d`. To me `--params` made me think that they would be passed in as parameters to test methods or fixtures. If we switch it to `--define`, should we rename the `TestParameters` class and the property on `TestContext` so they all match to prevent confusion? Maybe `TestDefines` and `TestContext.Defines`? Doesn't read as well, but I like consistency.

src/NUnitFramework/framework/TestParameters.cs
rprouse: I like the conversions, especially for a common type like `int`, so I don't want to see them dropped. I would prefer to see us write them once rather than consumers rewrite them many times.

An alternative might be to provide a number of string extension methods as a separate enhancement so they could write `.Get("name").AsInt(0)`. An `AsDateTime()` would also be useful for converting string DateTime values to real DateTimes since you can use DateTime in `TestCase`. Just a thought :smile:

CharliePoole: I started doing as a generic...

```C#
T Get<T>(string name, T defaultValue);
```

That's how we do it for `TestPackage.GetSetting` However, this is harder to implement because everything is stored as a string. The package settings are stored as objects. PackageSettings work because we are in control of what goes into the package and what we try to parse out of it.

If we implemented generic Get, and the user tried to do...

```C#
var x = TestContext.Parameters.Get("some_name", default_that_cant_be_parsed)
```

Then a runtime error would occur. Do you think that would be acceptable?

src/Common/nunit/CommandLineOptions.cs
CharliePoole: After our further discussion, I'm going with `--params|p` and sticking with `TestContext.Parameters`. The main motivating factor is that you call these "TestParameters" in the VS .runsettings file.

src/NUnitFramework/framework/TestParameters.cs
CharliePoole: Going with the following as a compromise...

```C#
string Get(string name);
string Get(string name, string defaultValue);
int GetInt(string name);
int GetInt(string name, int defaultValue);
```

Note that we can't use C# default argument values because we have a .NET 2.0 build.

The semantics of GetInt will be as follows...
1. If there is a value that can be parsed as an int, that will be returned.
2. If there is a value that fails parsing an exception will be thrown, causing the test to fail.
3. If there is no value, the default value will be returned. If there is no default provided, we have a choice of throwing an exception or returning a value like 0 or -1. Any thoughts on this?

CharliePoole:
Final changes are in as described on #720. Ready for final review and merge.

CharliePoole:
Travis Mono 3.2.8 build failing as it has for the past week. I sent in a support request.

src/NUnitFramework/framework/TestParameters.cs
```diff
@@ -44,17 +44,23 @@ public ICollection<string> Names
...
 
         public string Get(string name, string defaultValue)
         {
-            return _parameters.ContainsKey(name) ? _parameters[name] : defaultValue;
+            return this[name] ?? defaultValue;
         }
-        
-        public int Get(string name, int defaultValue)
+
+        public int GetInt(string name)
```
rprouse: I hate to be a pain, but these are all public API's and should have XMLDOC comments. Aren't they creating build warnings?

rprouse:
Add the documentation comments and I think we are good.

src/NUnitFramework/framework/TestParameters.cs
oznetmaster: How about GetDouble as well?

src/NUnitFramework/framework/TestParameters.cs
CharliePoole: What a pain! :-) Yeah, I worked too fast.

src/NUnitFramework/framework/TestParameters.cs
CharliePoole: If you have an actual need for it in your work, I could add it now. But I'd prefer not to add it speculatively because we can hardly ever remove anything once we add it. Same goes for DateTime, which @rprouse  had asked about.

CharliePoole: BTW, since the base value is passed as a string, adding new methods to the TestParameters class should be pretty trivial.

oznetmaster: It was the need for a double that started this whole thing for me (it is a tolerence kind of value), so yes, I think it would be useful.  I like @rprouse DateTime suggestion as well.

oznetmaster: Perhaps adding a generic <TValue> Get<Tvalue> (string name, TValue default value) would solve the problem?

CharliePoole: I'm OK with double. What about you @rprouse?

I suggested a generic Get earlier. The simplest approach is to use Convert.ChangeType. It will work for any T that is supported by IConvertible interface, which includes most common numeric types and DateTime. I'm happy to revert to this approach if you guys think it would be better for the future. @rprouse?

oznetmaster: I like having the generic Get<TValue> plus the plain non-generic Get that returns a string.

Perhaps there should be an Exists(string name) (or Contains) method as well.

Or maybe simply derive from Dictionary<string, string> but as a readonly collection?

That would eliminate a number of methods that are already in Dictionary.

CharliePoole:
Code I have locally for generic Get is pretty simple...

```C#
public T Get<T>(string name, T defaultValue)
{
    string val = this[name];
    return val != null ? (T)Convert.ChangeType(val, typeof(T), null) : defaultValue;
}
```

This would handle bool, Byte, char, most numeric types and DateTime. It would throw on any Types it could not handle. I don't think that's a problem, since we only need to convert strings you can enter on the command line.

I haven't pushed the code as I would want to remove GetInt and modify the tests to cover more types.

Which way shall we go guys - specific methods per Type or generic?

src/NUnitFramework/framework/TestParameters.cs
CharliePoole: I considered Exists(string name) but thought that for all practical purposes, getting the null back was sufficient.

I prefer aggregating the dictionary as now to inheritance. That way we only expose what we want people to see  and forwarding methods are easy to write.

oznetmaster: It is sometimes easier and cleaner to have an if (..Exists ("abc")) then an if (...Get("abc") = null).

Also, it generates less code.

CharliePoole: Fair enough.

oznetmaster: Actually, having the indexer for simple string value access, and the generic Get<T> for anything else may be cleaner.  Having a non-generic Get and a Generic get may be confusing.

rprouse: I am fine with the generic. Should it catch and rethrow an NUnit exception on conversion error, or just error the test? I am fine with either.

rprouse: As for `Exists`, I thought null was easier, `var param = TestContext.Parameters.Get("abc") ?? "default"`

CharliePoole:
Based on the comments, I'm going with generic.

I'll add Exists as well.

Let's get this finished!

src/NUnitFramework/framework/TestParameters.cs
oznetmaster: Why would you do that, when Get takes a default argument: Get ("abc", "default").

CharliePoole:
Ready for re-review, hopefully final. :-)

src/NUnitFramework/framework/TestParameters.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// <summary>
+        /// Get method is a simple alternative to the indexer
+        /// </summary>
+        /// <param name="name">Name of the paramter</param>
+        /// <returns>Value of the parameter or null if not present</returns>
+        public string Get(string name)
+        {
+            return Exists(name) ? _parameters[name] : null;
+        }
+
```
oznetmaster: This is very inefficient.  Should be:

string value;
 _parameters.TryGetValue (name, out value)

return value;

CharliePoole: I used Exists for clarity, but it's basically as if I wrote...

```C#
if (_parameters.ContainsKey(name) return _parameters[name];
```

This is a pattern we use all over NUnit. If memory serves, it's because TryGetValue doesn't (or maybe didn't) exist on some platforms. As far as I have always understood, it's a little bit inefficient, but not enough to make me want to prematurely optimize it, especially at the expense of clarity. The various TryXxx functions always seem very hard to read. 

Do you have an estimate of how inefficient it is in a typical use? For that matter, since it's new, do we have any idea what typical use will be? I'm thinking of some small number of users having maybe three or four parameters so maybe that biases my approach.

oznetmaster: Using exists and then getting the value requires two completely independent lookups in the dictionary.  Trygetvalue only does a single lookup.  I never use the pattern you used.  As far as I know every platform has trygetvalue for the generic dictionary class.

---Sent from Boxer | http://getboxer.com

On June 17, 2016 at 9:08:04 AM GMT+8, CharliePoole <notifications@github.com> wrote:In src/NUnitFramework/framework/TestParameters.cs:  > + /// <returns>Value of the parameter or null if not present</returns> > + public string this[string name] > + { > + get { return Get(name); } > + } > + > + /// <summary> > + /// Get method is a simple alternative to the indexer > + /// </summary> > + /// <param name="name">Name of the paramter</param> > + /// <returns>Value of the parameter or null if not present</returns> > + public string Get(string name) > + { > + return Exists(name) ? _parameters[name] : null; > + } > +   I used Exists for clarity, but it's basically as if I wrote...  if (_parameters.ContainsKey(name) return _parameters[name];   This is a pattern we use all over NUnit. If memory serves, it's because TryGetValue doesn't (or maybe didn't) exist on some platforms. As far as I have always understood, it's a little bit inefficient, but not enough to make me want to prematurely optimize it, especially at the expense of clarity. The various TryXxx functions always seem very hard to read.   Do you have an estimate of how inefficient it is in a typical use? For that matter, since it's new, do we have any idea what typical use will be? I'm thinking of some small number of users having maybe three or four parameters so maybe that biases my approach.  —You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or mute the thread.

rprouse:
I think users are likely to use a few dozen of these at most in their tests and I also think we have more serious memory and performance issues that we can look at. It is likely a bit less efficient this way, but I don't think it is enough to worry about. Right now, my priority is getting the 3.4 release ready.

oznetmaster:
How are quotations marks included in a string parameter?

oznetmaster:
This seems to be missing T Get (string name) method which would return default(T) if the parameter is not there.

PR #1572, opened 2016-06-14
Report unloading errors

### LINKED ISSUE #329: (CLI) Runner does not report AppDomain unloading timeout
Currently `DomainManager.DomainUnloader.Unload` is silently* timing `AppDomain.Unload(domain) ` out after 30 seconds.

* an entry like `09:35:36.420 Error [ 1] DomainManager: Unable to unload AppDomain test-domain-FOOBAR.dll, Unload thread timed out` is logged to the NUnit trace log (%AppData%\..\Local\NUnit\logs.

[msdn](http://msdn.microsoft.com/en-us/library/system.appdomain.unload%28v=vs.110%29.aspx) states that

>  If a thread does not abort, for example because it is executing unmanaged code, or because it is executing a finally block, then after a period of time a CannotUnloadAppDomainException is thrown in the thread that originally called Unload. If the thread that could not be aborted eventually ends, the target domain is not unloaded. Thus, in the .NET Framework version 2.0 domain is not guaranteed to unload, because it might not be possible to terminate executing threads. 

To sum it up:
* the `AppDomain` is not necessarily unloaded when `DomainUnloader.Unload` finishes
 * could this lead to issues when running other test assemblies subsequently?
 * it is still unloaded, however, when the process is exited (which is still possible)
* the issue is not made visible to the user.
* There is no statement made by microsoft how long `.Unload(domain)` takes before it times out with an `CannotUnloadAppDomainException`
* `CannotUnloadAppDomainException` may indicate serious issues in either productive or test code
 * i don't know about the part "for example because it is executing unmanaged code"... i think when a unit test ends, so should all processing of the unit under test.

Also see https://groups.google.com/forum/#!topic/nunit-discuss/Y_qy00H44ZQ

CharliePoole:
This will fix #329 but should not be merged until we agree on what it is doing!!!

As discussed in #329, I'm doing two things here:

1. I fixed it so that we get an error in the console runner when the AppDomain unload throws one. I think we need to discuss if we really want this.

2. I eliminated all use of a private shadow cache directory. We no longer create it or try to clean it up. The user or the OS will have to clean out any leftover files periodically. I think this part is good to go.

With the unload errors showing up, our console run now fails if we have any problem unloading. Even though the tests have run to completion, we don't get any output from them, just a message saying that we failed to unload the AppDomain. For me, this kind of reinforces the original V2 decision to allow these errors to go unnoticed.

What we could do next...

1. Leave it as it is... for the reasons already stated, I don't think this is a good idea.
2. Go back to ignoring the exceptions
3. Catch the exception but report it as some kind of warning - not 100% sure how to do this, we might need a new exception or a Severity property on NUnitEngineException.
4. We could possibly handle actual errors differently from timeouts.
5. Whatever we do in the engine, we need a way to avoid the error showing up in TFS runs using the VS adapter - for that reason, we would need to distinguish it from other exceptions.
6. We could pass a package setting to the engine to indicate whether we want such exceptions depending on which runner is doing the running.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/DomainManager.cs
+++ b/src/NUnitEngine/nunit.engine/Services/DomainManager.cs
@@ -145,62 +145,61 @@ namespace NUnit.Engine.Services
         #region Nested DomainUnloader Class
         class DomainUnloader
         {
-            private Thread thread;
-            private AppDomain domain;
+            private Thread _unloadThread;
+            private AppDomain _domain;
+            private Exception _unloadException;
 
             public DomainUnloader(AppDomain domain)
             {
-                this.domain = domain;
+                _domain = domain;
             }
 
             public void Unload()
             {
-                string domainName;
-                try
+                _unloadThread = new Thread(new ThreadStart(UnloadOnThread));
+
+                _unloadThread.Start();
+
+                if (!_unloadThread.Join(30000))
                 {
-                    domainName = "UNKNOWN";//domain.FriendlyName;
-                }
-                catch (AppDomainUnloadedException)
-                {
-                    return;
+                    string msg = "Unable to unload AppDomain, Unload thread timed out";
+
+                    log.Error(msg);
+                    _unloadThread.Abort();
+
+                    throw new NUnitEngineException(msg);
                 }
 
-                log.Info("Unloading AppDomain " + domainName);
-
-                thread = new Thread(new ThreadStart(UnloadOnThread));
-                thread.Start();
-                if (!thread.Join(30000))
-                {
-                    log.Error("Unable to unload AppDomain {0}, Unload thread timed out", domainName);
-                    thread.Abort();
-                }
+                if (_unloadException != null)
+                    throw new NUnitEngineException("Exception encountered unloading AppDomain", _unloadException);
             }
 
             private void UnloadOnThread()
             {
                 bool shadowCopy = false;
-                string cachePath = null;
                 string domainName = "UNKNOWN";               
 
                 try
                 {
-                    shadowCopy = domain.ShadowCopyFiles;
-                    cachePath = domain.SetupInformation.CachePath;
-                    domainName = domain.FriendlyName;
+                    shadowCopy = _domain.ShadowCopyFiles;
+                    domainName = _domain.FriendlyName;
 
-                    AppDomain.Unload(domain);
+                    // Uncomment to simulate an error in unloading
+                    //throw new Exception("Testing");
+
+                    // Uncomment to simulate a timeout while unloading
+                    //while (true) ;
+
+                    AppDomain.Unload(_domain);
                 }
                 catch (Exception ex)
                 {
+                    _unloadException = ex;
+
                     // We assume that the tests did something bad and just leave
                     // the orphaned AppDomain "out there". 
                     // TODO: Something useful.
                     log.Error("Unable to unload AppDomain " + domainName, ex);
-                }
-                finally
-                {
-                    if (shadowCopy && cachePath != null)
-                        DeleteCacheDir(new DirectoryInfo(cachePath));
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


CharliePoole:
BTW, there is a test failure because we actually were testing that these errors are suppressed. I'll fix the tests after we decide what to do.

CharliePoole:
@rprouse @nunit/core-team The key decision here is whether we want this error to be highlighted and fail the build (as originally requested), continue to be silently ignored or reported in some other way. Deciding what's best is probably the hardest part of it.

rprouse:
My thinking is that if the error is ours (like the shadow cache), then the error should be suppressed and not fail the build, but if the error is the user's, then it should show up. We can't always distinguish that, but we could make an educated guess. Do you think that the shadow cache was the main problem that caused the timeout and remaining errors will be user errors? If so, then we should surface it. If timeouts/errors are likely just random or our failures, we should suppress. 

Non-committal on my part, but you know the code and that is how I would make the call.

CharliePoole:
OK, I think this is ready to merge. Here's what it does...

1. NUnitEngineException is thrown if there is an error unloading or if the unloading thread times out.
2. ProcessRunner now rethrows the exception from unload, rather than merely logging it.
3. ConsoleRunner catches NUnitEngineException, which was previously caught in Main()
4. If a test result was returned, ConsoleRunner displays the results before doing anything with the exception. In this case, the exception is treated as a warning.
5. If no test result was returned, the exception is treated as an error and UNEXPECTED_ERROR is returned.

I think this is the best we can do at the moment. Further work would involve some sort of tracking of user-created threads, which I don't think we want to get into this week. :-)

BrunoJuchli:
two ideas: 

- runner creates a memory dump of itself - so the user can actually debug the issue. The warning by itself is not worth alot, it doesn't give any information at all. I don't know whether that's possible at all (without failing the process)
- alternatively `Debugger.Launch` or something similar, to enable a user to debug the issue "live". Problem here is with automated runs like on CI. That probably won't work there. So maybe a CLI-switch is necessary to manipulate this behavior.

src/NUnitEngine/nunit.engine/Services/DomainManager.cs
```diff
@@ -145,63 +145,62 @@ public void Unload(AppDomain domain)
...
+                    string msg = "Unable to unload AppDomain, Unload thread timed out";
+
+                    log.Error(msg);
+                    _unloadThread.Abort();
+
+                    throw new NUnitEngineException(msg);
                 }
+
+                if (_unloadException != null)
+                    throw new NUnitEngineException("Exception encountered unloading AppDomain", _unloadException);
```
espenalb: This is a good change - our test suite started to fail with this in place. Unfortunately, it is really hard to get the _unloadException. I propose to change the exception to:
`throw new NUnitEngineException("Exception encountered unloading AppDomain: " + _unloadException.Message, _unloadException);` - or even add more of the stack trace.

I tried to get more details out of the test runner by using command line option `--trace=Info`, but the only trace of the log statements were found in `InternalTrace.15632.log`
```

17:48:04.787 Info  [ 1] ProcessRunner: Unloading remote runner
17:48:17.747 Warning [ 1] ProcessRunner: Failed to unload the remote runner. Exception encountered unloading AppDomain
17:48:17.748 Info  [ 1] ProcessRunner: Unloading remote runner
17:48:17.811 Info  [ 1] ProcessRunner: Unloading remote runner
17:48:17.845 Info  [ 1] ProcessRunner: Unloading remote runner
17:48:17.902 Info  [ 1] ProcessRunner: Unloading remote runner
17:48:17.950 Info  [ 1] ProcessRunner: Unloading remote runner
17:48:18.051 Info  [ 1] ProcessRunner: Unloading remote runner
17:48:18.075 Info  [ 1] ProcessRunner: Unloading remote runner
17:48:18.087 Info  [ 1] ProcessRunner: Unloading remote runner
17:48:18.118 Info  [ 1] ProcessRunner: Unloading remote runner
17:48:18.143 Info  [ 1] ProcessRunner: Unloading remote runner
17:48:18.155 Info  [ 1] ProcessRunner: Unloading remote runner
17:48:18.220 Info  [ 1] ProcessRunner: Unloading remote runner
17:48:18.226 Info  [ 1] ProcessRunner: Unloading remote runner
```
I execute tests in multiple test assemblies, so as you can see, I had multiple test runners. I could not find anything to determine which of my test assemblies caused the problem - here also a little more info would be nice...

PR #1581, opened 2016-06-16
Clear retained reference to test objects

### LINKED ISSUE #1430: TestObjects are retained for lifetime of test run, causing high memory usage
This is the same symptoms as #1247 - with 3.2.1, we're finding the same high memory usage, and TestObjects not being disposed again. I'm not sure yet if this is a regression, or if #1367 was never disposing correctly.

Background: we run a patched version of 3.0.1, which disposes of the test objects, and test fixture reference in `WorkItem.WorkItemComplete()`. This manages the memory 'correctly'. In #1367, we moved this to `OneTimeTearDown()` to be more semantically correct, which seemed like it would be addressing the same problem. I don't know if freeing the references in OneTimeTearDown rather than WorkItemComplete caused a problem, or if this is a separate regression. 

Have a busy couple of weeks at work, but can look at this in the future - unless anyone else is able to beat me to it!

ChrisMaddock:
Fixes #1430.

Quick history, this was initially submitted as #1367. In that PR, we moved the reference clear from `WorkItem.WorkItemComplete` to `OneTimeTearDown`, to match up to the fields being initialised in `OneTimeSetUp`. When v3.2.1 was released this didn't work, I presume I missed checking the references were still released after making the change.

I've verified that moving the change back to `WorkItemComplete` solves the problem - what I haven't had time to do yet is dig in to why having it in `OneTimeTearDown` did not work - wonder if anyone has any ideas on this?

Submitting this now as it would be great for us if it can squeezed in to 3.4. :-) I'm fairly tight for time up till the end of this week, but will do what I can if any changes are required, although have done it on a branch so anyone else can modify as necessary!

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/OneTimeTearDownCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/OneTimeTearDownCommand.cs
@@ -71,10 +71,6 @@ namespace NUnit.Framework.Internal.Commands
                 IDisposable disposable = context.TestObject as IDisposable;
                 if (disposable != null && Test is IDisposableFixture)
                     disposable.Dispose();
-
-                //Clear references to test objects to reduce memory usage
-                context.TestObject = null;
-                Test.Fixture = null;
             }
             catch (Exception ex)
             {
--- a/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
@@ -450,6 +450,10 @@ namespace NUnit.Framework.Internal.Execution
 
             if (Completed != null)
                 Completed(this, EventArgs.Empty);
+
+            //Clear references to test objects to reduce memory usage
+            Context.TestObject = null;
+            Test.Fixture = null;
         }
 
 #endregion


rprouse:
Does anyone remember why it got moved to the `OneTimeTeardown`? Was it causing another error?

rprouse:
Partially unrelated to this PR, but a good place to inform @nunit/core-team members. I have disabled Travis as a required check for merging to master until it gets fixed. Once it is, I will re-enable.

rprouse:
Travis Mono latest is also failing in this PR as in mine. It looks like something has changed with the latest mono update?

Travis is reporting

```
XBuild Engine Version 14.0
Mono, Version 4.4.0.0
```

Last working version reported

```
XBuild Engine Version 12.0
Mono, Version 4.2.4.0
```

Looks like we have breaking changes between Mono 4.2.4.0 and 4.4.0.0.

I am entering an issue. I think we can ignore for now.

ChrisMaddock:
> Does anyone remember why it got moved to the OneTimeTeardown? Was it causing another error?

There wasn't any error, but `WorkItemComplete` seemed a bit of a random place to be clearing the references. It was suggested to move it to the equivalent location to where the references were set. Looking back through the PR, it looks like I deduced that OneTimeTearDown was the appropriate place, which may well have been totally wrong!

There's a lot of noise in the PR, as we tried to change the structure of the chain of references, but there are some tests dependent on it being initialised how it is. (From memory!) We eventually reverted that attempt, and just commited the two lines. :-)

CharliePoole:
I think the problem is that OneTimeTearDown isn't run for test cases. Should have seen that!

Let's merge this and try to find a better fix when we have more leisure. My goal would be to remove the Fixture member entirely from the Test object. It really doesn't belong there any longer and it's a hold-over from the 2.x architecture. However, a few things depend on it, so it will take some work.

Issue #1582, opened 2016-06-16
Mono 4.4.0 Causes Test Failures


rprouse:
Travis Mono latest is also failing in PRs.

Travis is reporting

```
XBuild Engine Version 14.0
Mono, Version 4.4.0.0
```

Last working version reported

```
XBuild Engine Version 12.0
Mono, Version 4.2.4.0
```

Looks like we have breaking changes between Mono 4.2.4.0 and 4.4.0.0.

Error reported are in the engine test suites;

```
1) Failed : NUnit.Engine.Services.Tests.RuntimeFrameworkServiceTests.SelectRuntimeFramework("mock-assembly.exe","2.0.50727",False)
  String lengths are both 9. Strings differ at index 0.
  Expected: "2.0.50727"
  But was:  "4.0.30319"
  -----------^
at NUnit.Engine.Services.Tests.RuntimeFrameworkServiceTests.SelectRuntimeFramework (System.String assemblyName, System.String expectedVersion, Boolean runAsX86) <0x4099d310 + 0x001ef> in <filename unknown>:0
2) Failed : NUnit.Engine.Services.Tests.RuntimeFrameworkServiceTests.SelectRuntimeFramework("net-2.0/mock-assembly.exe","2.0.50727",False)
  String lengths are both 9. Strings differ at index 0.
  Expected: "2.0.50727"
  But was:  "4.0.30319"
  -----------^
at NUnit.Engine.Services.Tests.RuntimeFrameworkServiceTests.SelectRuntimeFramework (System.String assemblyName, System.String expectedVersion, Boolean runAsX86) <0x4099d310 + 0x001ef> in <filename unknown>:0
3) Failed : NUnit.Engine.Services.Tests.RuntimeFrameworkServiceTests.SelectRuntimeFramework("nunit-agent.exe","2.0.50727",False)
  String lengths are both 9. Strings differ at index 0.
  Expected: "2.0.50727"
  But was:  "4.0.30319"
  -----------^
at NUnit.Engine.Services.Tests.RuntimeFrameworkServiceTests.SelectRuntimeFramework (System.String assemblyName, System.String expectedVersion, Boolean runAsX86) <0x4099d310 + 0x001ef> in <filename unknown>:0
4) Failed : NUnit.Engine.Services.Tests.RuntimeFrameworkServiceTests.SelectRuntimeFramework("nunit-agent-x86.exe","2.0.50727",True)
  String lengths are both 9. Strings differ at index 0.
[...]

PR #1583, opened 2016-06-16
Update mono paths for latest mono

ChrisMaddock:
Mono moved some dlls in 4.4.

http://www.mono-project.com/docs/about-mono/releases/4.4.0/

This change adds the new paths in addition to the old ones, for older versions of mono. No change to 1.0, as it's no longer supported. I don't have much of an overview of how these paths are used, and if anything else could be affected, but I couldn't find the path specified anywhere else.

### DIFF
--- a/src/NUnitEngine/nunit.engine/RuntimeFramework.cs
+++ b/src/NUnitEngine/nunit.engine/RuntimeFramework.cs
@@ -593,14 +593,16 @@ namespace NUnit.Engine
                     frameworks.Add(framework);
                 }
 
-                if (File.Exists(Path.Combine(monoPrefix, "lib/mono/2.0/mscorlib.dll")))
+                if (File.Exists(Path.Combine(monoPrefix, "lib/mono/2.0/mscorlib.dll")) ||
+                    File.Exists(Path.Combine(monoPrefix, "lib/mono/2.0-api/mscorlib.dll")))
                 {
                     RuntimeFramework framework = new RuntimeFramework(RuntimeType.Mono, new Version(2, 0, 50727));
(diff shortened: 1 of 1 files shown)


ChrisMaddock:
Fixes #1582

rprouse:
That was quick @ChrisMaddock :+1: The one Travis build is passing, just waiting on AppVeyor.

ChrisMaddock:
Google did all the work on this one. ;-)

Any guesses why I'm seeing `Build not found or access denied.` on the PR Appveyor, Rob? Anything changed? I can see the branch build fine.

Link is: https://ci.appveyor.com/project/CharliePoole/nunit/build/1.0.2832

ChrisMaddock:
Hmm...I can view it now it's completed. Strange, pretty sure I can normally see them when they're pending, but never mind.

CharliePoole:
Looks good to merge. I think this should be in the release notes, so we should either have an issue or label this PR like an issue. Either way is good.

CharliePoole:
Never mind... it looks like this fixes #1582

CharliePoole:
I'm working on nunit/nunit-console#34 and corresponding about detection of mono versions on the mono list. It looks like this fix was a mistake on our part.

The mono-?.?-api directories are not the same as the mono-?.? directories. They only contain type definitions to allow compiling  for a 2.0, etc. target. There is no implementation.

Since we use our enumeration of availble runtimes to determine under what profiles we can actually _run_ the tests, this doesn't serve us. It did have the side-effect of fixing the issue that mono was not being recognized because of the absence of profile directories. That's because newer versions of mono don't support these profiles and so we have to detect them differently.

This comment is for info only- I'll fix it in nunit/nunit-console#34.

Issue #1579, opened 2016-06-16
NUnit v3 dangerously overrides COMPLUS_Version environment variable


ww898:
Why? All child processes will inherit it and TRY running under CLR v4.0.30319, but in fact some of them can't. Please remove it from your code or tell me the big reason why you did it. Now we should remove COMPLUS_Version  before run our tests...

hypersw:
We are testing IIS Express which is a native entry point and it's instructed to run with CLR2. When managed code starts running within it with CLR2, it tries to favor the env variable and load CLR4 assemblies, and it causes the server to fail. That's now it seems now.

Also this env variable would affect all of the child processes run from within the test process, which is hardly a right thing to do because their CLR version selection is not in sync with the NUnit test CLR selection.

The agent should probably clear this env variable when it starts — I assume it's only set there to run the agent in the right CLR. Or maybe we could have a set of agents for different CLRs as before, because this hack with env vars turns out to have side-effects.

CharliePoole:
A little more information please...

We set COMPLUS_Version for the agent process because that's the accepted way to run a program using a specific framework. We do this whenever we are the tests running in a separate process either based on the user-specified `--framework` option or the target framework of the test assembly. The agent itself will never start a subprocess, so that can only happen if your own code does it. Most likely, you are allowing environment settings from the current process when you create a subprocess.

What I'm missing is this: If you need to test something under .NET 2.0, why are your tests targetting .NET 4? What you are calling the "NUnit test CLR selection" is actually selected by the user.

But it's quite possible I'm missing something here. Please explain if so.

jcansdale:
I can see this being an issue if you want to run tools as part of your tests. I've run into this kind of issue before (building test assemblies with an older version of csc.exe). Could the agent nuke the `COMPLUS_Version` variable when it starts? That's the way TestDriven.Net handles it.

hypersw:
There're lots of accepted ways for CLR selection, like `ICLRMetaHost` and `.config`. and so on, all of them non-viral for child processes. The problem with this one is that it affects any child processes run from tests in a totally unexpected way. But this can be mitigated if the agent process restores the state of the env vars after they have played their role.

[...]

PR #1585, opened 2016-06-17
TestAgency modifies the COMPLUS_Version env var for its child agent p…

hypersw:
…rocess to run in the specific CLR, and the agent process now restores it to the original value for test code to run in the original env. (gh-1579)

### DIFF
--- a/src/NUnitEngine/nunit-agent/Program.cs
+++ b/src/NUnitEngine/nunit-agent/Program.cs
@@ -89,6 +89,22 @@ namespace NUnit.Agent
                 Environment.Version,
                 RuntimeFramework.CurrentFramework.DisplayName);
 
+            // Restore the COMPLUS_Version env variable if it's been overridden by TestAgency::LaunchAgentProcess
+            try
+            {
+              string cpvOriginal = Environment.GetEnvironmentVariable("TestAgency_COMPLUS_Version_Original");
+              if(!string.IsNullOrEmpty(cpvOriginal))
+              {
+                log.Debug("Agent process has the COMPLUS_Version environment variable value \"{0}\" overridden with \"{1}\", restoring the original value.", cpvOriginal, Environment.GetEnvironmentVariable("COMPLUS_Version"));
+                Environment.SetEnvironmentVariable("TestAgency_COMPLUS_Version_Original", null, EnvironmentVariableTarget.Process); // Erase marker
+                Environment.SetEnvironmentVariable("COMPLUS_Version", (cpvOriginal == "NULL" ? null : cpvOriginal), EnvironmentVariableTarget.Process); // Restore original (which might be n/a)
+              }
+            }
+            catch(Exception ex)
+            {
+              log.Warning("Failed to restore the COMPLUS_Version variable. " + ex.Message); // Proceed with running tests anyway
+            }
+
             // Create TestEngine - this program is
             // conceptually part of  the engine and
             // can access it's internals as needed.
--- a/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
+++ b/src/NUnitEngine/nunit.engine/Services/TestAgency.cs
@@ -212,8 +212,12 @@ namespace NUnit.Engine.Services
 
                 case RuntimeType.Net:
                     p.StartInfo.FileName = agentExePath;
+                    // Override the COMPLUS_Version env variable, this would cause CLR meta host to run a CLR of the specific version
                     string envVar = "v" + targetRuntime.ClrVersion.ToString(3);
                     p.StartInfo.EnvironmentVariables["COMPLUS_Version"] = envVar;
+                    // Leave a marker that we have changed this variable, so that the agent could restore it for any code or child processes running within the agent
+                    string cpvOriginal = Environment.GetEnvironmentVariable("COMPLUS_Version");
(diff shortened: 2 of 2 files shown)


hypersw:
Travis on linux failed with `943Unable to find version '0.6.4' of package 'Cake'.`, does not look like a PR problem.

src/NUnitEngine/nunit-agent/Program.cs
```diff
@@ -89,6 +89,22 @@ public static int Main(string[] args)
                 Environment.Version,
                 RuntimeFramework.CurrentFramework.DisplayName);
 
+            // Restore the COMPLUS_Version env variable if it's been overridden by TestAgency::LaunchAgentProcess
+            try
+            {
+              string cpvOriginal = Environment.GetEnvironmentVariable("TestAgency_COMPLUS_Version_Original");
+              if(!string.IsNullOrEmpty(cpvOriginal))
+              {
+                log.Debug("Agent process has the COMPLUS_Version environment variable value \"{0}\" overridden with \"{1}\", restoring the original value.", cpvOriginal, Environment.GetEnvironmentVariable("COMPLUS_Version"));
```
jcansdale: I wonder if we should warn if the `TestAgency_COMPLUS_Version_Original` variable hasn't been set rather than when it has? Restoring it should be the normal case.

hypersw: I'm not sure about the ways an agent might be run.
For instance, on mono this won't be set, as it's gotten different ways for enforcing the CLR version (or I've heard Mono has dropped anything but the recent one).

src/NUnitEngine/nunit.engine/Services/TestAgency.cs
```diff
@@ -212,8 +212,12 @@ private Guid LaunchAgentProcess(TestPackage package)
 
                 case RuntimeType.Net:
                     p.StartInfo.FileName = agentExePath;
+                    // Override the COMPLUS_Version env variable, this would cause CLR meta host to run a CLR of the specific version
                     string envVar = "v" + targetRuntime.ClrVersion.ToString(3);
                     p.StartInfo.EnvironmentVariables["COMPLUS_Version"] = envVar;
+                    // Leave a marker that we have changed this variable, so that the agent could restore it for any code or child processes running within the agent
+                    string cpvOriginal = Environment.GetEnvironmentVariable("COMPLUS_Version");
+                    p.StartInfo.EnvironmentVariables["TestAgency_COMPLUS_Version_Original"] = string.IsNullOrEmpty(cpvOriginal) ? "NULL" : cpvOriginal;
                     p.StartInfo.Arguments = arglist;
```
CharliePoole: Needing to restore the old value is an edge case I hadn't thought of.

CharliePoole:
@hypersw The travis error is something I reported to them as probably having to do with their infrastructure. If we don't hear back soon we will most likely drop the Mono 3.2.8 build.

We are manually checking that mono latest passes as well as the Appveyor builds before merging.

CharliePoole:
I made a general comment about merging this from my phone yesterday but I don't see it here.  Here goes again...

If time is short for the release that @rprouse is doing or for @hypersw to modify, we can merge.

But if you can get to it I'd really prefer to get rid of setting TestAgency_COMPLUS_Version_Original to "NULL" when there was no original setting.

The case where there is no setting originally is the most common case, since users don't generally use the COMPLUS_Version environment variable. In that case, there's no need to save an "original" value. Agent can simply clear the environment value if no original was saved.

As far as logging goes, I would only log in the unusual case where agent actually finds something to restore.

Hopefully, this comment will post. :-)

rprouse:
I am hoping to start the 3.4 release process next weekend, so there is still time.

CharliePoole:
@hypersw Is there any reason to keep this setting to "NULL" that I missed seeing?

CharliePoole:
@hypersw This is awaiting your reply in order to make it into the 3.4 release.

hypersw:
Hmm, missed the orignial comment, sorry.

The idea behind the PR is to keep the original environment of the calling process in the agent process (with the exception of our own technical variable `TestAgency_COMPLUS_Version_Original`, which I consider we can exempt from these guarantees).

The possible cases are (not in probability order):

№ | User has  `COMPLUS_Version`? | We set `COMPLUS_Version`? | How it works
------------ | ------------- | ------------- | -------------
1 | Yes | Yes | We are restoring user's value from `_Original`.
2 | Yes | No | `_Original` not set, we don't touch anything.
3 | No | No | `_Original` not set, we don't touch anything.
4 | No | Yes | `NULL` in `_Original`, so we make the `COMPLUS_Version` prop empty.

`NULL` tells apart cases `2` and `4`, essentially. Without it, we would erase user's value in `2`.
Currently, `2` might happen under Mono, where `TestAgency` does not set the env var, and the agent does not know of this choice (apart from the `NULL` token). Whether this is a useful case is arguable, but things might change with CoreCLR and combinations of running tests with Mono and testing with CoreCLR and so on. So I thought I'd just make it work correctly whatever happens around — restore user's env as it were, whatever its cause, that's all.

CharliePoole:
Ah! Well thought out. Merging.

rprouse:
Great job @hypersw, thanks for the help :+1:

PR #1590, opened 2016-06-19
Avoid duplicate extensions

### LINKED ISSUE #1427: Engine extensions load old packages
As reported by @asbjornu in #1323 

I just upgraded `NUnit.ConsoleRunner` from version `3.2.0` to version `3.2.1` and got the following output from `nunit3-console.exe` that I haven't seen before:

```
NUnit Console Runner 3.2.1

ProjectLoader extension .nunit is already handled by another extension.
```

The workaround was to delete NuGet packages in the packages directory and restore. The engine plugin resolution was probing the old 3.2.0 packages that were left on the system after an upgrade.

I think the solution is that as we are probing for engine extensions we need to look at the directory names, parse out the package name/version and discard directories where there are newer versions.

CharliePoole:
Fixes #1427 

The initial commit solves the problem of duplicate extension assemlies in the full installation but not when using NuGet packages.

NOTE: We will need to decide on a versioning scheme for extensions. In the current build, extensions all remain as 1.0.0.0.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -707,7 +707,8 @@ Task("PackageNuGet")
         });
         NuGetPack("nuget/extensions/teamcity-event-listener.nuspec", new NuGetPackSettings()
         {
-            Version = packageVersion,
+		    // The teamcity-event-listener extension uses its own versioning
+            Version = "1.0.0" + dbgSuffix,
             BasePath = currentImageDir,
             OutputDirectory = PACKAGE_DIR,
             NoPackageAnalysis = true
--- a/src/NUnitEngine/nunit.engine/Services/ExtensionAssembly.cs
+++ b/src/NUnitEngine/nunit.engine/Services/ExtensionAssembly.cs
@@ -99,27 +99,8 @@ namespace NUnit.Engine.Services
             if (version < otherVersion)
                 return false;
 
-            // Versions are equal, check for a nuget package version.
-            
-            // Override only if this one was specified exactly while the other wasn't
+            // Versions are equal, override only if this one was specified exactly while the other wasn't
             return !FromWildCard && other.FromWildCard;
         }
-
-        //private bool TryGetNugetPackageVersion(out version)
-        //{
-        //    var dir = new DirectoryInfo(Path.GetDirectoryName(FilePath));
-        //    if (dir.Name != "tools")
-        //        return false;
-
-        //    dir = dir.Parent;
-        //    if (dir == null)
-        //        return false;
-
-        //    var packages = dir.GetFiles("*.nupkg");
-        //    if (packages.Length != 1)
-        //        return false;
-
-        //    return packages[0].FullName;
-        //}
     }
 }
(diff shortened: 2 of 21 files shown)


src/NUnitEngine/nunit.engine/Services/ExtensionAssembly.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        public string FilePath { get; private set; }
+        public bool FromWildCard { get; private set; }
+
+        private AssemblyDefinition _assemblyDefinition;
+        public AssemblyDefinition Assembly
+        {
+            get
+            {
+                if (_assemblyDefinition == null)
```
rprouse: We may not need it here, but should we get in the habit of using Lazy<T> for all lazy initialization? It may not be needed now, but it is easy to implement and future proofs the code.

CharliePoole: That would mean adding the NUnit.System.Linq package with the engine. We planned on that but haven't done it yet. If you think we should, I'll do a separate PR to make sure it doesn't break anything and then merge it .

rprouse: I'm not sure, it was just an observation, do you think it is worth the effort? I would like to see the Linq package added to the engine eventually, but wasn't planning on doing it so soon. Do you think it is worth it?

CharliePoole: I'd rather get the duplicate extension problem solved for the release. Use of Lazy etc. only impacts our dev work.

CharliePoole:
This isn't really helping us now, but it will in the future. I added a named property `EngineVersion` to `ExtensionAttribute` in order to indicate the minimum engine version required for an extension. Teamcity extension is set to 3.4, while the others don't use it. I tested by increasing the value and watching in debug to see that the extension was ignored.

CharliePoole:
This is ready to merge. In addition to the CI tests passing, I have  tested the ConsoleRunner NuGet package with two different versions of the NUnitProjectLoader package. Only the latest one gets loaded.

In addition to the base fix, I added the EngineVersion named property as described above and updated the AssemblyVersion of our existing extensions to 3.4, matching the nuget packages. The new teamcity listener is versioned as 1.0.0.0 for the assembly and 1.0.0 for the package.

build.cake
```diff
@@ -707,7 +707,8 @@ Task("PackageNuGet")
         });
         NuGetPack("nuget/extensions/teamcity-event-listener.nuspec", new NuGetPackSettings()
         {
-            Version = packageVersion,
+		    // The teamcity-event-listener extension uses its own versioning
+            Version = "1.0.0" + dbgSuffix,
```
rprouse: If we are going to be versioning each plugin on it's own update cycle, we should switch each of them to use their own version/AssemblyInfo and extract the version to use here from that. I just did that in a private project with lots of plugins and it worked well. It meant that I only have to update the version in one place. We might even want to do the same for the framework and the engine, pull the common version, rather than update it there and here. Just thinking out loud :smile:

PR #1604, opened 2016-06-22
Issue 1553

### LINKED ISSUE #1553: Allow ordering tests to be done in multiple properties

CharliePoole:
Fixes #1553 

This is my shot at #1553, based on #1564 by @jnsn.

This version...

 * Uses an enum for direction

 * Generalizes the algorithm into a series of steps, which will always be one step if there is no "By"

 * Adds test cases and reorganizes the tests

Still needed...

 * The Comparer should logically be part of each step, allowing such things as

   ```C#
   Is.Ordered.By("A").Using(myComparer).Then.By("B").Using(myOtherComparer)
   ```

 * This implementation still gives errors at runtime rather than compile time. I realized while working on it that this particular constraint is much more difficult than many others in this regard. I think it's possible, but would probably be better as a separate follow-on issue.

@jnsn Feel free to continue from this or combine it with your version. I don't plan to work further on it for a while.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
@@ -37,10 +37,7 @@ namespace NUnit.Framework.Constraints
     {
         private List<OrderingStep> _steps;
         // The step we are currently building
-        private OrderingStep _active;
-
-        private ComparisonAdapter _comparer = ComparisonAdapter.Default;
-        private string _comparerName;
+        private OrderingStep _activeStep;
 
         enum OrderDirection
         {
@@ -75,9 +72,9 @@ namespace NUnit.Framework.Constraints
         {
             get
             {
-                if (_active.Direction != OrderDirection.Unspecified)
+                if (_activeStep.Direction != OrderDirection.Unspecified)
                     throw new InvalidOperationException("Only one directional modifier may be used");
-                _active.Direction = OrderDirection.Ascending;
+                _activeStep.Direction = OrderDirection.Ascending;
                 return this;
             }
         }
@@ -89,9 +86,9 @@ namespace NUnit.Framework.Constraints
         {
             get
             {
-                if (_active.Direction != OrderDirection.Unspecified)
+                if (_activeStep.Direction != OrderDirection.Unspecified)
                     throw new InvalidOperationException("Only one directional modifier may be used");
-                _active.Direction = OrderDirection.Descending;
+                _activeStep.Direction = OrderDirection.Descending;
                 return this;
             }
         }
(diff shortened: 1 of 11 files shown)


rprouse:
@CharliePoole are you planning on getting this in for the release on the weekend?

CharliePoole:
No... it's primarily being worked on by

jnsn:
I'll attempt to pick this up this weekend if I can carve out some time.

CharliePoole:
That's great... when you have time, please add any thoughts you may have. BTW, if you plug in my tests to your implementation, you'll see why I added the `break` statement in that inner loop.

CharliePoole:
It would be desirable to merge this before splitting the repositories per #1598, but not absolutely necessary since it only deals with the framework.

CharliePoole:
Rebased on top of current master.

CharliePoole:
I'm having to repeatedly restart the Apple builds when they time out. It's getting annoying! We may be hanging and causing it, but I don't see how.

CharliePoole:
@rprouse @jnsn I think this is ready to merge now. Error handling is at run time rather than compile time because that's a major bit of work I don't have time for right now.

Basically, the syntax is that you can have one or more groups with `By` or just one group without `By`. The groups are separated explicitly by `Then` or implicitly by a new `By` coming along. Each group can have one `By`, one directional and one `Using`. See the tests for examples. I'll need to do a revision of the documentation page to cover all this.

rprouse:
I will look at this again in the morning when I am not as tired. Meanwhile, @jnsn can you give this a second look?

CharliePoole:
@rprouse I've also been mentioning @jnsn but he seems to have disappeared from our world. :disappointed:

PR #1612, opened 2016-06-23
Issue 1605

### LINKED ISSUE #1605: TeamCity package has no pre-release suffix
As pointed out on #1593, there are ~~two~~ three problems with the Appveyor and MyGet feeds for the new teamcity extension:

1. The package was previously deployed as development version 3.3, but is being  released as version 1.0. Consequently, older packages in the 3.3 series must be removed.

2. The package is not getting a -CI-nnnn suffix added as should be done, so it appears as 1.0.0 in both feeds. This needs to be fixed.

3. The NUnit.Console nuget package is still depending on the same version because the dependency is being generated automatically.

CharliePoole:
Fixes #1605 

This PR essentially does two things...

1. Makes sure that the Appveyor builds of the teamcity event listener get the proper CI suffix.
2. Makes sure that the two runner packages that depend on the listener use the correct version.

Old event listener packages in the 3.3 range have been removed from MyGet. I can't figure out how to remove them from AppVeyor. From references, it may be necessary to remove the NUnit project from AppVeyor and re-add it, which will eliminate all builds. Not sure we want to do that.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -54,6 +54,7 @@ var PROJECT_DIR = Context.Environment.WorkingDirectory.FullPath + "/";
 var PACKAGE_DIR = PROJECT_DIR + "package/";
 var BIN_DIR = PROJECT_DIR + "bin/" + configuration + "/";
 var IMAGE_DIR = PROJECT_DIR + "images/";
+var TOOLS_DIR = PROJECT_DIR + "tools/";
 
 var SOLUTION_FILE = IsRunningOnWindows()
 	? "./nunit.sln"
@@ -69,6 +70,9 @@ var PACKAGE_SOURCE = new string[]
 // Test Runners
 var NUNIT3_CONSOLE = BIN_DIR + "nunit3-console.exe";
 var NUNITLITE_RUNNER = "nunitlite-runner.exe";
+
+// NuGet.exe
+var NUGET_EXE = TOOLS_DIR + "nuget.exe";
 
 // Test Assemblies
 var FRAMEWORK_TESTS = "nunit.framework.tests.dll";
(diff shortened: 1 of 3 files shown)


rprouse:
My research on removing old AppVeyor artifacts from the feed led me to believe the same thing @CharliePoole. I think that is a bit dangerous to do, especially this close to a release :smile:

build.cake
```diff
@@ -618,20 +629,25 @@ Task("PackageNuGet")
...
-            BasePath = currentImageDir,
-            OutputDirectory = PACKAGE_DIR,
-            NoPackageAnalysis = true
-        });
+
+		PackageRunnerWithExtensions("nuget/runners/nunit.console-runner-with-extensions.nuspec", packageVersion, teamcityVersion, currentImageDir, PACKAGE_DIR);
+
+		PackageRunnerWithExtensions("nuget/runners/nunit.runners.nuspec", packageVersion, teamcityVersion, currentImageDir, PACKAGE_DIR);
+
+        //NuGetPack("nuget/runners/nunit.console-runner-with-extensions.nuspec", new NuGetPackSettings()
```
rprouse: Do we need to maintain the commented out code?

CharliePoole: Not at all... I'll delete.

CharliePoole:
@rprouse Can we use the account feed instead?

rprouse:
@CharliePoole we could likely use the account feed if we rename it and push the packages to it. I don't think it picks up artifacts automatically. We could also rename this feed to something like `nunit-framework` and then rename the account feed to `nunit`?

Do you want to do that as part of this PR? It will require changes to build.cake or maybe to the AppVeyor config to publish the packages.

CharliePoole:
If we do that, it should be separate I think - after the release. Once there is a new release, the other feeds will become momentarily less important so that might be the time to do something.

rprouse:
Sounds good. Once we start splitting repos we are going to have to do a major re-org anyway. Plus if we have more repositories, it might make sense to consolidate on the organization feed as opposed to the project feeds.

PR #1613, opened 2016-06-24
Fix extensions path for nuget 3 and handle path errors

CharliePoole:
This fixes #1607 and fixes #1551 

In the case of #1607, this is a quick fix, simply adding another entry in the addins file. I'm working on a better fix and it's possible I'll have it before Saturday but I think this little one should be merged anyway.

Issue #1607, opened 2016-06-23
nunit.nuget.addins discovery pattern is wrong then restored through project.json


maximpashuk:
I have project.json file

``` json

{
  "version": "0.0.0",

  "frameworks": {
    "net461": { }
  },

  "dependencies": {
    "NUnit.ConsoleRunner": "[3.2.1]",
    "NUnit.Extension.NUnitV2ResultWriter": "[3.2.1]",
    "NUnit.Extension.NUnitV2Driver": "[3.2.1]"
  }

}

```

Then I restore it through NuGet3

```
nuget restore project.json
```

I receive folder structure

```
%UserProfile%
    .nuget
        packages
            NUnit.ConsoleRunner
                3.2.1
                    tools
                        nunit.nuget.addins
            NUnit.Extension.NUnitV2Driver
                3.2.1
                    tools
            NUnit.Extension.NUnitV2ResultWriter
                3.2.1
                    tools
```

```nunit.nuget.addins``` file have filter ```../../NUnit.Extension.*/tools/``` which is fine for NuGet2&packages.config case.
In case of NuGet3&project.json filter is wrong, no extensions is loaded automatically
This is bug.

I think in this case pattern should be like ```%UserProfile%/.nuget/packages/NUnit.Extension.*/**/tools/```.

So to work correctly in NUnit3&project.json case, please add one more search pattern in ```nunit.nuget.addins``` file

CharliePoole:
Thanks for letting us know about that. The problem with the solution you are suggesting is that %UserProfile% won't be recognized on all environments. I think the second pattern would have to be `../../../NUnit.Extension.*/**/tools` in order to work.

Meanwhile, you should of course simply edit the `.addins` file, which is intended to be user-modified. In a future release, we may remove these entries in favor of a programmed search for the packages directory.

PR #1616, opened 2016-06-24
Revise approach to locating extensions

CharliePoole:
PR #1613 will take care of issue #1607 as written. This PR aims to make improvements to that fix but it does so by making a number of changes to how NUnit finds extensions. If we can verify it today, it would be a good addition to the 3.4 release but we shouldn't put it in without testing the packages thoroughly.

Changes included:

* Use of `/**/` in a `.addins` file now means 0 or more subdirectories rather than 1 or more. This allows the use of `**/tools` for both nuget v2 and v3.

* The initial directory used for starting the search for addins does not have all its assemblies examined. This eliminates the need for an empty `.addins` file in the directory.

* The search for extensions in the packages directory is now done programmatically, without the need for wildcard entries in the addins file to locate them. Consequently, nunit.nuget.addins has been removed.

I'll go ahead and install this in a project using the NUnit.Console nuget package. I should be able to copy over all the addin tests and run them in that project.

### DIFF
--- a/nuget/engine/nunit.engine.nuspec
+++ b/nuget/engine/nunit.engine.nuspec
@@ -28,6 +28,5 @@
     <file src="bin\nunit-agent.exe.config" target="lib" />
     <file src="bin\nunit-agent-x86.exe" target="lib" />
     <file src="bin\nunit-agent-x86.exe.config" target="lib" />
-    <file src="..\..\nuget\engine\nunit.nuget.addins" target="lib"/>
   </files>
 </package>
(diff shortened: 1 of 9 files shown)


CharliePoole:
I should add one more feature of this PR that may not be obvious... The older approach using a `.addins` file only works when the engine and the extensions are both installed as nuget packages. In the current approach, the engine can be xcopied or whatever and the nuget extensions will still be found.

CharliePoole:
I created a test app and installed the nunit.console package, which includes all the extensions. I ran tests that exercised all the extensions and everything passed except for my V2 test. I suspect that this has never worked when installed as an extension because the v2 driver is not able to locate the v2 nunit assemblies in it's own directory, when that directory is outside of the appbase.

We could decide to merge this as doing some things better or wait till after the release to spend more time on it. If it's the latter, we should merge the issue-1607 branch and leave this one unmerged.

CharliePoole:
My mistake... all extensions including the V2 driver work quite well now when NUnit.ConsoleRunner and the extensions are installed as nuget packages. I just hadn't gotten the setup of my test project right earlier.

That being the case, I'd like to get this revised fix merged into the 3.4 release. I'll check for any review items first thing in the morning.

rprouse:
This looks good to me, let's go with this one.

PR #1617, opened 2016-06-25
Load user profile on test runners

JohanO:
## Problem
The product I am working on uses Microsoft SQL LocalDb for data storage. I have a number of tests verifying that data is stored and retreived correctly to/from the database. These tests works fine when run inside Visual Studio with NUnit 3 Test Adapter, but fails with 'SQL connection error' on the build server using nunit3-console.

## Cause
LocalDb requires the user profile (some registry keys under HKEY_CURRENT_USER) to be loaded on the process attempting to connect. The NUnit 3 Test Adapter runs in process of Visual Studio and the user profile is already loaded and my tests work. nunit3-console runs the tests in a separate test runner processes. These processes do not load the user profile and thus my tests fails.

## Solution
Added a '--loaduserprofile' switch to unit3-console. Setting this will load the user profile when starting the test runners.  As loading the user profile takes time this switch is default off.
[See [https://msdn.microsoft.com/en-us/library/system.diagnostics.processstartinfo.loaduserprofile(v=vs.110).aspx](url)]

### DIFF
--- a/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
@@ -131,7 +131,7 @@ namespace NUnit.Common
             this.Add("shadowcopy", "Shadow copy test files",
                 v => ShadowCopyFiles = v != null);
 
-            this.Add("loaduserprofile", "Load user profile on test runner processes",
+            this.Add("loaduserprofile", "Load user profile in test runner processes",
                 v => LoadUserProfile = v != null);
 
             this.Add("agents=", "Specify the maximum {NUMBER} of test assembly agents to run at one time. If not specified, there is no limit.",
--- a/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
+++ b/src/NUnitConsole/nunit3-console/ConsoleOptions.cs
@@ -65,6 +65,8 @@ namespace NUnit.Common
 
         public bool ShadowCopyFiles { get; private set; }
 
+        public bool LoadUserProfile { get; private set; }
+
         private int maxAgents = -1;
         public int MaxAgents { get { return maxAgents; } }
         public bool MaxAgentsSpecified { get { return maxAgents >= 0; } }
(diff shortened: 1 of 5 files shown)


src/NUnitConsole/nunit3-console/ConsoleOptions.cs
```diff
@@ -129,6 +131,9 @@ protected override void ConfigureOptions()
             this.Add("shadowcopy", "Shadow copy test files",
                 v => ShadowCopyFiles = v != null);
 
+            this.Add("loaduserprofile", "Load user profile on test runner processes",
```
rprouse: Would this read better as "**in** test runner processes"?

JohanO: Ah, yes. English is not my first language so you are probably right. I will update it

rprouse:
@JohanO talk about trying to get stuff in under the wire, I am hoping to do the release today or tomorrow. :smile: 

This looks fairly straight forward and I am okay with including it after a second team member reviews. I will get it in 3.4 if we can do that in time. @nunit/core-team can someone take a look?

I am pretty sure this is only needed launching the agent where you did it because the main console process is already running under the user profile. Is that correct?

My only other question is if false should be the default. You say it is expensive, how expensive is it? I think that running tests in-process or in the agent should have the same behaviour by default. Your issue indicates that it is not the case.

rprouse:
I just checked the docs which state,

>If the value is true, the user's profile in the HKEY_USERS registry key is loaded. Loading the profile can be time-consuming. Therefore, it is best to use this value only if you must access the information in the HKEY_CURRENT_USER registry key.

Based on that, the default of false is probably best.

JohanO:
@rprouse You are correct. The main console is started by the build server with the user profile loaded. It is the runner agents that needed to load the user profile.

rprouse:
@JohanO can you add your new command line to the tests in `CommandLineTests.cs`?

Meanwhile, I am going to look at the Travis CI build failure.

rprouse:
Travis CI build failure is yet another Mono or Travis failure. It failed to restore Cake. Hopefully after you update the tests, the next run will pass.

rprouse:
@CharliePoole if you are available, could you look at this? I am willing to add it to 3.4 if we can get a second review soon.

rprouse:
I am going to merge this without the second review. I would really like to get this is 3.4. The code is very straight-forward and off by default.

Issue #995, opened 2015-11-06
Enable Warning as Error


rprouse:
We constantly have new warnings creeping into the code, often because of missing XML documents, or unused variables on one platform or another. We should fix these warnings, then enable warning as error on all projects to prevent new ones from being introduced.

ChrisMaddock:
@rprouse - Sorry, I always miss checking to see if old issues are assigned to anyone before I start. Mind if I take this on? The remaining changes are minimal.

rprouse:
I don't mind, but you need to make sure it builds without warnings on
Windows and Linux including Mono 3.2.8

rprouse:
Always feel free to clean up warnings though

CharliePoole:
@rprouse I set up the NUnit.System.Linq project to only be built on AppVeyor, since it's not built as part of NUnit but pulled in as a package.

ChrisMaddock:
Thanks @rprouse. It looks like most of the work is already done to be honest, but hopefully enabling this should mean no regressions.

ChrisMaddock:
Ah - I'd naively assumed warnings present would be similar for xbuild and msbuild. Looks like there's more to do that I thought, will chip away further and see how far I get.

rprouse:
Now you see why an apparently easy issue isn't done yet :smile_cat:

PR #1619, opened 2016-06-25
Set Warnings as errors

ChrisMaddock:
This is work towards #995.

This should cover all existing build warnings in MSBuild and latest mono. Without the mono 3.2.8 Travis build working, I don't have a set up to test it, so haven't enabled warning-as-error in the csproj's as I can't verify it will also build in 3.2.8.

If anyone else is set up to check manually, it may be possible to make the switch - else, I'll come back to this after the 3.2.8 Travis is up and running again. :-)

N.B. Nuget upgrade is for nunit/NUnit.System.Linq#1

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -4,7 +4,6 @@
 
 var target = Argument("target", "Default");
 var configuration = Argument("configuration", "Debug");
-var framework = Argument("framework", "net-4.5");
 
 //////////////////////////////////////////////////////////////////////
 // SET ERROR LEVELS
--- a/src/NUnitEngine/Addins/teamcity-event-listener/teamcity-event-listener.csproj
+++ b/src/NUnitEngine/Addins/teamcity-event-listener/teamcity-event-listener.csproj
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <PropertyGroup>
     <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
--- a/src/NUnitEngine/Addins/teamcity-event-listener/teamcity-event-listener.csproj
+++ b/src/NUnitEngine/Addins/teamcity-event-listener/teamcity-event-listener.csproj
@@ -20,6 +20,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
@@ -28,6 +29,7 @@
     <DefineConstants>TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="System" />
--- a/src/NUnitEngine/nunit.engine/Internal/NUnitConfiguration.cs
+++ b/src/NUnitEngine/nunit.engine/Internal/NUnitConfiguration.cs
@@ -59,34 +59,27 @@ namespace NUnit.Engine.Internal
         {
             get
             {
-                if (_monoExePath == null)
+                if (_monoExePath == null && IsWindows())
                 {
-                    string[] searchNames = IsWindows()
-                        ? new string[] { "mono.bat", "mono.cmd", "mono.exe" }
-                        : new string[] { "mono", "mono.exe" };
-                    
-                    if (_monoExePath == null && IsWindows())
+                    RegistryKey key = Registry.LocalMachine.OpenSubKey(@"Software\Novell\Mono");
+                    if (key != null)
                     {
-                        RegistryKey key = Registry.LocalMachine.OpenSubKey(@"Software\Novell\Mono");
-                        if (key != null)
+                        string version = key.GetValue("DefaultCLR") as string;
+                        if (version != null)
                         {
-                            string version = key.GetValue("DefaultCLR") as string;
-                            if (version != null)
+                            key = key.OpenSubKey(version);
+                            if (key != null)
                             {
-                                key = key.OpenSubKey(version);
-                                if (key != null)
-                                {
-                                    string installDir = key.GetValue("SdkInstallRoot") as string;
-                                    if (installDir != null)
-                                        _monoExePath = Path.Combine(installDir, @"bin\mono.exe");
-                                }
+                                string installDir = key.GetValue("SdkInstallRoot") as string;
+                                if (installDir != null)
+                                    _monoExePath = Path.Combine(installDir, @"bin\mono.exe");
                             }
                         }
                     }
-
-                    if (_monoExePath == null)
-                        _monoExePath = "mono";
                 }
+
+                if (_monoExePath == null)
+                    _monoExePath = "mono";
 
                 return _monoExePath;
             }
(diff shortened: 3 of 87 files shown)


src/NUnitEngine/nunit.engine/Internal/NUnitConfiguration.cs
```diff
@@ -59,35 +59,28 @@ public static string MonoExePath
         {
             get
```
ChrisMaddock: In this method, searchNames was unused. It seems to relate to old functionality to look for mono on the path, which was previously removed, but this var left. Also reduced the nesting of if's a little here - shoulnd't be any difference in functionality.

src/NUnitEngine/nunit.engine/Services/ExtensionAssembly.cs
```diff
@@ -50,7 +48,7 @@ public AssemblyDefinition Assembly
                     resolver.AddSearchDirectory(System.IO.Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location));
                     var parameters = new ReaderParameters() { AssemblyResolver = resolver };
```
ChrisMaddock: Here, `ReaderParameters` was constructed but not passed in to `ReadAssembly()`. This still compiled because of an overload. I presumed passing the object in was the intention.

CharliePoole: Correct.

src/NUnitFramework/framework/nunit.framework-2.0.csproj
```diff
@@ -436,7 +436,7 @@
   </ItemGroup>
   <ItemGroup>
     <Reference Include="NUnit.System.Linq, Version=0.2.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL">
-      <HintPath>..\..\..\packages\NUnit.System.Linq.0.2.0-CI-12-master\lib\net20\NUnit.System.Linq.dll</HintPath>
+      <HintPath>..\..\..\packages\NUnit.System.Linq.0.2.0-CI-17-master\lib\net20\NUnit.System.Linq.dll</HintPath>
```
rprouse: I did the same change in the release 3.4 branch, so you will get a merge conflict. I would prefer if we merge my 3.4 PR before this one so that this one will get labeled 3.5 once we merge.

CharliePoole: I don't think we should ever merge references to a CI package into master.

rprouse: @CharliePoole did you create a new release version of Linq? If so, @ChrisMaddock should probably update to use that.

CharliePoole: Several. :-) Waiting for (I hope) the final one to build now.

CharliePoole:
There's an 0.6.0 build of NUnit.System.Linq on MyGet that you can use.

ChrisMaddock:
Thanks @CharliePoole - rebased.

CharliePoole:
Ready to merge now or do you have more you want to do?

ChrisMaddock:
This is good to merge, thanks @CharliePoole.

Now we have the 3.2.8 build back, we can theoretically finish off #995 - but I'd like to dig into all the ignored warnings a bit first, and not sure when I'll next have a block of time. So if this part can be merged to prevent needing further rebasing, that would be great. :-)

CharliePoole:
Let's do it after I finish 3.4.1

ChrisMaddock:
@CharliePoole - I retract the above for now, I'm doing a little more meddling now, and it would be easier to do it on top of this. Will shout up when it's ready again!

src/NUnitEngine/Addins/teamcity-event-listener/teamcity-event-listener.csproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
```
ChrisMaddock: mono 3.2.8 build uses xbuild 12.0, so it warns 14.0 is unrecognised. Other extensions all use ToolsVersion=4.0

build.cake
```diff
@@ -4,7 +4,6 @@
 
 var target = Argument("target", "Default");
 var configuration = Argument("configuration", "Debug");
-var framework = Argument("framework", "net-4.5");
```
ChrisMaddock: (was unused)

CharliePoole:
@ChrisMaddock OK but let's merge the next time it's at a stable point. The amount charged us by GitHub for creating a new PR is pretty low! When you have a long-running PR like this, the reviewers end up having to go back and look over the same changes again and again - not to mention that things are changing underneath you as you go.

ChrisMaddock:
@CharliePoole - Sure, sorry about that. I just needed the changes in this branch to enable warning as errors, and I'm not sure my git-fu is strong enough to start branching off my branches!

This should be good to go again now, just waiting for Appveyor to build the frameworks I don't build locally. Will shout up when it's done.

ChrisMaddock:
@CharliePoole, @rprouse

Ok, this is ready to merge again. :-) It now fixes #995

Warning as Error should now be enabled on all projects. I only haven't touched the MockCpp ones, as I don't have VS2012/don't know what they're there for!

I've also been through and removed almost all the warnings that were ignored per-assembly. The only remaining case is "Unused field" being ignored for the framework tests, as this would have required a load of pragma's everywhere, because of all the reflection. Maybe a thought for the (long-term!) future, when we go c#6 and can use `nameof`.

CharliePoole:
@ChrisMaddock The trick to branching off your branches is that you can push the new branch in order to see it build but you shouldn't do a PR until you have merged in or rebased over master, after the original PR merges. PRs adjust nicely to changes in your branch but not in the underlying branch. It's not a problem with git, which can deal with all that easily enough, but with the visual representation of the changes online.

Will re-review. :-)

src/NUnitEngine/nunit.engine/Internal/NUnitConfiguration.cs
CharliePoole: I'm slightly worried about the "shouldn't make a difference" part. Have you verified that it still works correctly? I can see that the code "should" do the same thing, but if I made these changes myself, I would not trust it without seeing it work. Since this is a level of functionality beyond unit testing, you would have to run it on a Windows machine with mono installed and either step through or look at the test output. We do have a test that lists the available runtimes to the console, so that might do it for you.

This is a good example of where even a small change would benefit from being done in a separate PR, but let's try to keep this in if we can. If you don't have a machine set up to test it, I can do it.

ChrisMaddock: That's a fair point. I do have mono installed, but have neither the registry key nor mono on my path, so without artificially creating the registry key (which doesn't seem sensible!) - I'm not able to check this. Is it only certain versions of mono which install the key, do you know? If your installation does have the relevant key, would be great if you could take a look.

The tidying is non-essential to the PR, just habit! We could just remove the unused var instead, if we can't find a set up to verify this.

ChrisMaddock:
Great, thanks. Will try that next time!

ChrisMaddock:
@CharliePoole - apologies, I didn't refresh and see you later post. Have made a separate comment on the mono issue. I also submitted #1642 earlier to re-enable the 3.2.8 build, if that would be easier, otherwise go ahead and close it. :-)

CharliePoole:
I'll pull it down and run it when I get home - in an hour or so.

We've all gradually learned certain efficiencies when working as a part-time, distributed team, which are not always the same efficiencies you adopt when working either alone or as part of a dedicated team sitting in the same room. One of them is to do small focused bits of work per issue/PR when possible. That gets them reviewed and into the codebase quickly and avoids merge problems. Of course, it isn't always possible... 

Global changes like this one are sometimes better if you partition them by the code they affect. At first, the business of doing separate PRs seems like a bit of an obstacle, but once you've done 100 or so, it's second nature. :-)

CharliePoole:
Had a little trouble syncing up to this issue. My windows machine treats "Issue-995a" the same as "issue-995a" but GitHub doesn't. Our standard naming is lower case.

CharliePoole:
I'm not seeing mono detected with either your branch or with master, so this is already broken. I'm going to merge your code and continue to look at the problem.

ChrisMaddock:
Whoops - will try and remember that from now on! And also break down the issues further...think this is probably they first I've done which has been particularly widespread.

Thanks for looking into the last bit!

Issue #1596, opened 2016-06-21
Eliminate code sharing across projects to be split


CharliePoole:
We could do this as part of each issue in the Epic, but it will be simpler if we eliminate or reduce sharing first. Steps to take:

1. Identify which projects actually use source files in Common.
2. Move any files only used by one project to that project.
3. Copy files as needed to the projects that use them.

Note... projects not being split can obviously continue to use common, e.g.: framework and nunitlite.

CharliePoole:
This is an inventory of the files in common, organized in logical groups that can be dealt with in the same way. Where a test file exists in common the file is marked with **(test)**.

I'll use these groupings in further comments to design a plan for eliminating common.

**Internal Trace**
These files are used to create internal trace logs. All of them are in the logging subdirectory. Shared by **framework** and **engine**.
 * ILogger.cs
 * InternalTrace.cs
 * InternalTraceLevel.cs
 * InternalTraceWriter.cs
 * Logger.cs

**Color Console Writer**
These files implement color output to the console and also provide interfaces to be used in creating the same output in non-console environments. Shared by **console** and **nunitlite**.
 * ColorConsole.cs
 * ColorConsoleWriter.cs
 * ColorStyle.cs
 * ExtendedTextWrapper.cs **(test)**
 * ExtendedTextWriter.cs

**Command Options**
These files implement command-line processing. Shared by **console** and **nunitlite**.
 * CommandLineOptions.cs
 * DefaultOptionsProvider.cs **(test)**
 * IDefaultOptionsProvider.cs
 * Options.cs
 * OutputSpecification.cs **(test)**
 * TestNameParser.cs **(test)**
 * TestSelectionParser.cs **(test)**
 * TestSelectionParserException.cs
 * Tokenizer.cs **(test)**

**Miscellaneous**
These files need to be dealt with individually.
 * AssemblyHelper.cs **(test)**<br/>
   Allows finding the path or directory of an assembly. Shared by **engine** and **framework**.
 * Env.cs<br/>
   Defines a small number of cross-platform values. Shared by **console** and **framework**.
 * Guard.cs<br/>
   Implements guard clauses for methods. Shared by **engine** and **framework**.
 * PackageSettings.cs<br/>
   Defines names of settings used in test packages. Shared by **console**, **engine**, **framework** and **nunitlite**.
 * XmlHelper.cs<br/>
[...]

Issue #1621, opened 2016-06-25
Create command-line client helper assembly


CharliePoole:
As part of issue #1596, I'd like to create an assembly that incorporates the command-line options code and the color console code that is shared between the console runner and nunit-lite. Other things could be added to it if we find we need them, but the focus would be on sharing between those programs as well as with other command-line programs. In particular, we could try to make it useful for the dotnet runner as well.

This would __not__ address sharing of code that affects the framework and engine but it would eliminate 14 out of 24 common classes that we now have. 

Initially, this will be a separate assembly under the nunit solution. Eventually, it would be a separate project and would produce private packages that we would reference.

CharliePoole:
@nunit/core-team I'd like to get some agreement that this is the right way to go before I do it. My reasoning is that these are both runners with references to other assemblies already and another one won't hurt. That's a bit different from nunit-framework, which has traditionally been standalone. We may decide to do something else for its shared code.

CharliePoole:
I'm dropping the approach of a separate assembly, which would have to be built in four or five variants merely to eliminate a small number of classes. The code will be duplicated.

PR #1622, opened 2016-06-25
Remove files from common

CharliePoole:
This is a partial fix toward #1596. It removes Env.cs and XmlHelper.cs from common.


CharliePoole:
This is ready to merge once the builds pass.

CharliePoole:
@nunit/core-team I'm building another PR on top of this one, so it will be helpful is someone can review and merge.

src/Common/nunit/CommandLineOptions.cs
```diff
@@ -66,9 +75,9 @@ public CommandLineOptions(params string[] args)
                 Parse(args);
         }
         
-        #endregion
+#endregion
         
-        #region Properties
+#region Properties
```
rprouse: All the regions got unindented. This happens with VS2015 whenever you add new `#if` statements. Should we just accept the changes until VS gets its act together?

rprouse:
Looks good except for my one question on the regions. Make a call, I am good either way. Fix it, or we accept it.

src/Common/nunit/CommandLineOptions.cs
CharliePoole: Let's not worry about it for now. This is the first of a series of changes that will hit this file anyway. I'll try to clean them all up in #1621

Issue #1623, opened 2016-06-27
NUnit 3.4 is not integrated with TeamCity


NikolayPianikov:
TeamCityEventListener is not activated by the call `            _extensionService.EnableExtension("NUnit.Engine.Listeners.TeamCityEventListener", _options.TeamCity);`
TeamCityEventListener has this attribute `[Extension(Enabled = false, EngineVersion = "3.6")]`
But `ENGINE_VERSION = 3.5.0.0`

CharliePoole:
This is a bad one! I verified that both the master and the 3.4.0 release use 3.4. That means the team-listener extension will never actually be used.

The value should be 3.4. I changed it back and forth several times in order to test it manually. I don't know how 3.6 got into the release undetected, because we have tests that verify the proper extensions are detected.

Seems like a critical bug, since nobody will be able to use 3.4 under TeamCity till it's fixed. This would probably call for an immediate 3.4.1 release if the error were in any other component. However, since we already planned to release this extension separately, maybe we should should accelerate releasing the process. @rprouse What do you want to do about this?

rprouse:
I assume that this only made it into the Team City extension? We can release an update to this extension, but I think we need to also release an update to at least the NUnit.Runners meta package. It has specified all of the dependencies as `=` not `>=` so it won't pick up the update.

![image](https://cloud.githubusercontent.com/assets/493828/16385171/c2125120-3c59-11e6-86d5-0b6b6f5974d7.png)

rprouse:
https://github.com/nunit/nunit/blob/master/nuget/runners/nunit.console-runner-with-extensions.nuspec should just have `version="$version$"` not `version="[$version$]"` to take the default behaviour. 

https://docs.nuget.org/create/versioning#Specifying-Version-Ranges-in-.nuspec-Files

rprouse:
Should we be doing a 3.4.1 of the runners and extensions?

rprouse:
I can probably knock out a quick 3.4.1 release tonight. We should branch off the `release\3.4.0` branch and update the `nuspec` files to use `>=` and pull in the updated extensions that need it. Unfortunately, I think we would need to update all the extensions because we use the single `$version$` in the nuspec files. Or if the other extensions don't need updates, hack it and hard code `3.4.0` for those extensions in this branch only.

@CharliePoole if you have time to get some of the groundwork done, it would help.

CharliePoole:
[...]

PR #1624, opened 2016-06-27
#1623 Change EngineVersion to "3.4.0.0" for attribute Extension for TeamCityEventListener

NikolayPianikov:


rprouse:
I thought @charliepoole said this wouldn't kick in until 3.6,thus the higher number. It could be it isn't loaded for another reason. Can you comment @charliepoole. 

@NikolayPianikov, would it make sense for you to set up a TeamCity CI server against NUnit to ensure we don't break your integrations. If you want to do that, probably best to wait a few weeks until we split up the repository.

NikolayPianikov:
@rprouse  I've tested the integration - extension works fine

CharliePoole:
@rprouse What I meant was that the check would not be useful till 3.6, since 3.4 is the first one implementing it. I put 3.6 into the attribute to do tests. It should not have got into a PR but apparently did.

@NikolayPianikov I think Rob is saying we need to run integration tests before releasing!

CharliePoole:
I merged this, but we need to do some more work to fix the problem.

NikolayPianikov:
Yes we have integration tests, but for the last relesed version only. I will add them for the master branch.

rprouse:
Thanks @NikolayPianikov. Are those tests internal only, or can we get some visibility into them?

NikolayPianikov:
Another one question is what happens when user restores packages via a nuget command? Will all plugins have to work without any triks?

NikolayPianikov:
I will try to move them to the external build server.

CharliePoole:
The big question for me is was why our CI passes. The answer is simple...

1. We have no unit tests of the teamcity extension.
2. We have one functional test that ensures all known extensions are recognized, but the teamcity extension was never added to it.
3. And, of course, we have no automated integration tests.

Embarrassing as it is, maybe one theme for the 3.6 release is to improve our own testing!

@NikolayPianikov I'm going into a meeting now. I'll come back to you in a few hours and maybe we can work together to get this done better, first for 3.4.1 and then for later on.

CharliePoole:
@NikolayPianikov Keep watch on the 3.4.1 milestone, which we will finish very quickly. It has only two issues in it. let me know if there are other things that seem critical and have a quick solution!

NikolayPianikov:
@rprouse @CharliePoole see integration tests [here](http://win10nik.cloudapp.net/viewType.html?buildTypeId=NUnit_NUnit3IntegrationTests).

PR #1630, opened 2016-06-28
Fix required minimum engine version for teamcity extension

### LINKED ISSUE #1623: NUnit 3.4 is not integrated with TeamCity

CharliePoole:
This fixes #1623 

Contents...

* Change required engine version for extension to 3.4
* Advance extension version to 1.0.1
* Update runner packages to use 1.0.1 
* Add minimal tests to check attributes on each extension

The last point is only slightly helpful but at least it means we have to make the same mistake in two places for the tests to pass.

We'll need to do more around the issue of integration testing after the 3.4.1 release.

### DIFF
--- a/src/NUnitEngine/Addins/addin-tests/VisualStudioProjectLoaderTests.cs
+++ b/src/NUnitEngine/Addins/addin-tests/VisualStudioProjectLoaderTests.cs
@@ -56,6 +56,45 @@ namespace NUnit.Engine.Services.ProjectLoaders.Tests
         }
 
         [Test]
+        public void CheckExtensionAttribute()
+        {
+            Assert.That(typeof(VisualStudioProjectLoader),
+                Has.Attribute<ExtensionAttribute>());
+        }
+
+        // Note for review:
+        // The following test doesn't pass because AttributeConstraint always uses the
+        // first attribute it finds. Should we fix the syntax or just document it?
+        //[TestCase(".sln")]
+        //[TestCase(".csproj")]
+        //[TestCase(".vbproj")]
+        //[TestCase(".vjsproj")]
+        //[TestCase(".vcproj")]
+        //[TestCase(".fsproj")]
+        //public void CheckExtensionPropertyAttributes(string ext)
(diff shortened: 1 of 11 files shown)


src/NUnitEngine/Addins/addin-tests/VisualStudioProjectLoaderTests.cs
```diff
@@ -56,6 +56,45 @@ public void ValidExtensions(string project, bool isGood)
...
         [Test]
+        public void CheckExtensionAttribute()
+        {
+            Assert.That(typeof(VisualStudioProjectLoader),
+                Has.Attribute<ExtensionAttribute>());
+        }
+
+        // Note for review:
+        // The following test doesn't pass because AttributeConstraint always uses the
+        // first attribute it finds. Should we fix the syntax or just document it?
```
rprouse: This sounds like one we should fix, but not now. Does `AttributeConstraint` just work on the first attribute, or the first attribute of an attribute type? Either way, I think we should fix it. Can you enter an issue?

rprouse:
Looks good. Can you enter an issue about `AttributeConstraint`?

Waiting for CI to pass...

src/NUnitEngine/Addins/addin-tests/VisualStudioProjectLoaderTests.cs
CharliePoole: I will. I'm not 100% sure we can fix it - I know it's not an easy one. The evaluation proceeds from left to right and the attribute we use is selected by name when we hit HasAttribute. What we want to do instead is select all the attributes and test each of them in (logical) parallel to see if one passes.

PR #1632, opened 2016-06-28
Fix hanging agent process at end of run

### LINKED ISSUE #1628: Agent's process stays in memory when it was failed to unload appdomain

CharliePoole:
Fixes #1628 

This first commit adds code to do a better job of killing the unloading thread if it hangs. This is only one hypothesis about the cause of the problem, so it would be helpful to have it tested before going further. @NikolayPianikov you seem to be able to reproduce this better than the rest of us, could you test?

Even if the first commit works, I think I'll go on to also kill the process if necessary, but it would be good to know the effect of the first change first.

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Internal/DirectoryFinderTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Internal/DirectoryFinderTests.cs
@@ -62,7 +62,7 @@ namespace NUnit.Engine.Internal.Tests
         [TestCase("*/v2-tests/*.dll", 2)]
         [TestCase("add*/v?-*/*.dll", 2)]
         [TestCase("**/v2-tests/*.dll", 2)]
-        [TestCase("addins/**/*.dll", 18)]
+        [TestCase("addins/**/*.dll", 17)]
         [TestCase("addins/../net-*/nunit.framework.dll", 4)]
         public void GetFiles(string pattern, int count)
         {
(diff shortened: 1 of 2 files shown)


rprouse:
Looks good, but I assume by your last statement that you are also thinking of killing the process?

CharliePoole:
I was hoping to hear from @NikolayPianikov before adding a new commit with that in it. I looked at the link he sent to integration tests, but they don't seem to relate to this problem at all. 

@NikolayPianikov ? We can release without your feedback, but it's not likely we will do a 3.4.2 release just for this issue so it's best if you can test.

NikolayPianikov:
I can't reproduce this issue in the integration tests. I am waiting for the feedback from one guy. Unfortunately he is unavailable yet.

CharliePoole:
@rprouse How do you want to do this?

rprouse:
@NikolayPianikov, by "I can't reproduce", does that mean that this version is working for you but you want to check with someone else?

Waiting for morning may not work for me. It is a long weekend here and I am going away, so it will be difficult for me to do the release after tonight EST.

rprouse:
@CharliePoole do you want to merge this or should we wait?

CharliePoole:
I would merge it except this is another one where I forgot to do it against the release branch. I can selectively merge those things into the release branch locally and push to that branch though.

rprouse:
I am merging as is.

PR #1636, opened 2016-06-30
Revise approach to locating extensions

### LINKED ISSUE #1626: NUnit 3.4: NUnit.ConsoleRunner is not picking up NUnit.Extension.NUnitV2ResultWriter
When running `NUnit.ConsoleRunner` and `NUnit.Extension.NUnitV2ResultWriter` `nuget` packages in the following layout

`.\tools\NUnit.ConsoleRunner\tools\nunit3-console.exe`
`.\tools\NUnit.Extension.NUnitV2ResultWriter\tools\nunit-v2-result-writer.dll`

the call `"C:\some\dir\tools\NUnit.ConsoleRunner\tools\nunit3-console.exe" "C:\some\dir\src\Tests\Test.Assembly.dll" "--work=C:\some\dir\tests" --full "--result=C:\some\dir\tests\TestResult.xml;format=nunit2" --labels=All --noheader` succeeds when both packages have version `3.2.1` but fails if both packages have version `3.4.0`. The error message is `Unknown result format: nunit2`

### LINKED ISSUE #1635: Console option to list loaded extensions.
I have this in a debug build and it helps to solve a lot of problems. I think we want users to be able to use it as well to make it easier for them (and us) to diagnose bugs.

Current format is a list of all extension points. Under each one I list each extension installed, whether it is disabled and if it has any properties set. I don't list the EngineVersion because that would only be helpful in cases where the extension isn't installed. Debugging exactly why an extension is not installed will always take use of the log and / or debugger but it's helpful to know __whether__ they are installed to start out with.

CharliePoole:
Fixes #1626 Fixes #1635 
Changes:

* Returned to use of nunit.nuget.addins file in our nuget packages. The file now has a line for nuget V2 format installs and another for V3, which incorporates an extra directory level. Will find extensions for the NUnit.ConsoleRunner package provided the common parent of all packages is either two or three directories up.

* Eliminated programmatic scan for `packages` dir, since the name is not guaranteed.

* Re-defined `**` in the addins file to mean zero or more directory components rather than one or more.

* Tracks assemblies that have already been seen so they are not processed twice.

* Adds an option to the console runner to allow displaying all extensions that are in use.

I originally planned to add the option of another addins file in the TestDirectory. Since addins are initialized before any tests are loaded, this would require too much work to do right now.

### DIFF
--- a/nuget/engine/nunit.engine.nuspec
+++ b/nuget/engine/nunit.engine.nuspec
@@ -28,5 +28,6 @@
     <file src="bin\nunit-agent.exe.config" target="lib" />
     <file src="bin\nunit-agent-x86.exe" target="lib" />
     <file src="bin\nunit-agent-x86.exe.config" target="lib" />
+    <file src="..\..\nuget\engine\nunit.nuget.addins" target="lib"/>
   </files>
 </package>
--- a/nuget/engine/nunit.engine.tool.nuspec
+++ b/nuget/engine/nunit.engine.tool.nuspec
@@ -29,5 +29,6 @@
     <file src="bin\nunit-agent.exe.config" target="tools" />
     <file src="bin\nunit-agent-x86.exe" target="tools" />
     <file src="bin\nunit-agent-x86.exe.config" target="tools" />
+    <file src="..\..\nuget\engine\nunit.nuget.addins" target="tools"/>  
   </files>
 </package>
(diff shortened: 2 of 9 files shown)


CharliePoole:
Travis doesn't seem to be processing builds at the moment.

rprouse:
Travis had some down time yesterday, so I think their queue is long.

CharliePoole:
Yeah. I guess I submitted it during the down time. Done now though.

rprouse:
Looks good.

CharliePoole:
Do I need to push this into the release branch as well now?

rprouse:
No, I have done that. See my PR.

Issue #345, opened 2014-11-08
Order of Fixture Execution


CharliePoole:
In additon to ordering test cases as envisioned by #170, we will probably want to provide for ordering of test fixtures. See discussion at https://groups.google.com/forum/#!topic/nunit-discuss/u2wKj0gTSko

Sebazzz:
I'd vote for this. This is very useful for integration tests and actually the only reason I use MSTest for that.

NecroFilja:
Are there any workarounds?
Generally what is default sort order of testfixtures/suites and any way to inject inside of it?

CharliePoole:
In the past they were sorted alphabetically. You can experiment to see if that's still true since its really just a side effect. You have to take the full namespace into account.

CharliePoole:
If course, if you're a developer you can always volunteer to implement this, under guidance if needed.

Sebazzz:
@NecroFilja You could add set the `Order` property on an `ITest` by creating a attribute implementing `IBeforeTest`. However, the ordering is only applied within namespace. You'd need to implement a custom `ITestAssemblyBuilder`, however, the default implementation [can't currently bey overwritten](https://github.com/Sebazzz/nunit/tree/custom-testassemblybuilder). I wrote a [short piece](http://damsteen.nl/blog/2016/06/13/ordered-tests-with-nunit-mstest-xunit-pt5-nunit-implementation-revised) about implementing test fixture ordering.

CharliePoole:
@Sebazzz Wow! All that work and no contrib to NUnit... too bad. :-(

Seriously, ordering of tests is always within the parent test in NUnit and the namespace represents a parent test. It really has to be that way, since you can associate OneTimeSetUp methods with namespaces using SetUpFixtures.

yauhen:
> In the past they were sorted alphabetically. 

@CharliePoole as far as I could see on current moment there is only GetTypes order of which is **generally** unpredictable (of course we now that in reality it return alpha sorted)

> CompositeWorkItem.cs#L248

@CharliePoole Thanks for pointing out. Just found myself. For me it looks like it is just enough to removed limitation for Order attribute (only allow to be used for methods). And due of this code it will works as it should (as temporary solution). Or are there any conceptual pitfalls (generally why this limit was set).

@Sebazzz I will try to read more carefully, initially looks a little bit over-complicated.

CharliePoole:
[...]

PR #1648, opened 2016-07-04
Allow custom assembly builder by attribute

Sebazzz:
As shortly discussed in #345. 

In summary, the following changes have been made:
- Add attribute that is to be added to target test assembly, allowing a custom `ITestAssemblyBuilder` to be selected.
- The RuntimeAssemblyBuilder instantiates the correct `ITestAssembyBuilder` or uses the default implementation.
- No tests have been developed, not sure to proceed here or if it is useful here since there isn't much logic here.

If you open a wiki page I then I can write the documentation if you like.

### DIFF
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -70,7 +70,11 @@ namespace NUnit.Framework.Api
         /// <param name="settings">A Dictionary of settings to use in loading and running the tests</param>
         public FrameworkController(string assemblyNameOrPath, string idPrefix, IDictionary settings)
         {
-            this.Builder = new DefaultTestAssemblyBuilder();
+#if !NETCF
+            this.Builder = new RuntimeAssemblyBuilder();
+#else
+            this.Builder = new DefaultTestAssemblyBuilder(); // TODO: Implement RuntimeAssemblyBuilder for .NET CF
+#endif
             this.Runner = new NUnitTestAssemblyRunner(this.Builder);
 
             Test.IdPrefix = idPrefix;
@@ -453,7 +457,7 @@ namespace NUnit.Framework.Api
 
         #region Nested Action Classes
 
-        #region TestContollerAction
+        #region TestControllerAction
 
         /// <summary>
         /// FrameworkControllerAction is the base class for all actions
--- a/src/NUnitFramework/framework/Api/RuntimeAssemblyBuilder.cs
+++ b/src/NUnitFramework/framework/Api/RuntimeAssemblyBuilder.cs
@@ -1,0 +1,153 @@
+﻿// ***********************************************************************
+// Copyright (c) 2016 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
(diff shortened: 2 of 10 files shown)


CharliePoole:
Let's review and discuss before doing much more... I'll be able to look at the code this evening (PDT).

The failure on AppVeyor was because the compact framework didn't build. Easiest thing to do right now is to use #if !NETCF around the new stuff and let CF use the default builder. If we accept this, then somebody who is already set up for CF builds can implement it in that solution.

The travis build gets farther but fails on FrameworkControllerTests - are you running the tests locally? The test is pretty dumb and just expects the builder to be of the right type. You should fix that so we can see it run all the way through.

I'll come back to this later when I have a bit more time.

CharliePoole:
Portable tests are failing on AppVeyor. FYI, you can run all the tests locally as well...

```
build -t Test [-c Debug]
```

Sebazzz:
Yes, got it. Also fixed the code formatting.

CharliePoole:
@rprouse I thought I had already tagged you on this one, but apparently not. We need to decide whether we actually want to give users this degree of control before asking @Sebazzz to spend any more time on it.

This PR allows them to replace the builder that is used to recognize tests. It's definitely a brute force approach. Users are empowered to take over test discovery completely, although they can only easily use components already present in the framework to do it. Still, it's possible to break NUnit, creating a test tree that does not look like what we expect to receive, particularly in the NUnit gui. That's what @Sebazz originally did in attempting to solve #345. This fix no longer includes that...

By supplying this, we give users a very strong workaround tool for any problems we have not yet fixed but it's a tool that can really mess them up if used incorrectly.

Thoughts? @nunit/core-team ?

Sebazzz:
It would open up possibilities for a wide range of scenarios. For example, an extension to NUnit could implement discovery of tests by convention like XUnit allows. I agree that implementing a custom `ITestAssemblyRunner` also comes with a lot of responsibility.

CharliePoole:
@Sebazzz AFAIK, xUnit doesn't natively support discovery by convention - assuming you mean the same thing by that as the Fixie folks do. NUnit currently only supports user-created conventions based on attributes. We deliberately decided not to allow non-attribute approaches.

That said, if we wanted an approach where users could define the conventions, I think we would want to write the code that interprets the conventions ourselves, not have it written by users. That is, the users would create the conventions, but we would create the code that allows them to do that. That is, of course, how Fixie works.

So, for me, this is an example of something that ought not be done as a user-injected feature because it runs the risk that the user will inject things that the rest of NUnit can't deal with. 

Of course, our existing extensibility interfaces introduce risks as well. We just have to balance the risks against the benefits.

rprouse:
I admit, this feature worries me. As @CharliePoole says, it could very easily change the tests produced in ways that we never anticipated. This could lead to very hard to diagnose bugs and also potentially put an expectation on the NUnit team to allow for greater flexibility. I would worry that a different test structure may not run as expected, may end up ignoring some of our attributes or produce results that third party tools cannot parse.

@CharliePoole do you think this would give users the ability to create a test hierarchy that produces XML results and events that is different from what we produce? If so, isn't that effectively breaking the contract of the engine API?

We can say that with great power comes great responsibility, but we will still end up being the first line of support and triage of issues. The burden of support is already high, so I am reluctant to possibly add to it especially for a feature that will only be used for a very small subset of our users.

CharliePoole:
In fact, the example that @Sebazz used produced xml that broke the API. 

So I'm inclined not to OK this, in spite of the fact that @Sebazzz has put some time into it and in spite of being quite impressed by the cleverness of it!

Maybe the next question to ask is... "Is there anyplace slightly below this level where we can introduce some greater flexibility in test discovery, but which doesn't make it possible to break our contracts?"

BTW, the API does allow for alternate builders but only if they are present in the framework assembly itself. Up to now, we have only included a single builder but we could include alternates or experimental versions in the future.

CharliePoole:
This PR applies to the framework only. Ideally, it should be closed or merged before we split per #1598, but it's not absolutely required.

rprouse:
@Sebazz I am going to close this PR. I am sorry that you put so much work into it, but I feel the possible abuses and the fact that we will need to support this feature outweighs its benefits. Thanks for working with us on this.

PR #1649, opened 2016-07-05
Allow specifying partial list of optional params

### LINKED ISSUE #1640: When submitting only part of optional parameters, all are overriden by defaults
When there is more than one optional parameter in test method, submitting only some of them via TestCase attribute leads into overriding them all by defaults.

observed in: 3.2.1, 3.4.0

STR:
        [TestCase(1,1)]
        public void Test(int expected, int p1 = 0, int p2 = 0)
        {
            if (p1 + p2 != expected) Assert.Fail();
        }

ER: p1 will use submitted value, p2 will use default => test pass

AR: p1 and p2 uses default values => test fail

ChrisMaddock:
Fixes #1640.

Two problems - the first that my implementation would overwrite any specified optional parameters with `Type.Missing`. 

The second issue being that by passing in the default value in the `TestCaseAttribute` tests - you counldn't tell if you're getting the passed in value, or the default. Whoops!


CharliePoole:
Looks good - merging.

PR #1652, opened 2016-07-07
Remove PackageSettings.cs from Common

ChrisMaddock:
Another step towards #1596.

This duplicates Package Settings for the console, framework and engine. In each case, everything that was unused was removed.

### DIFF
--- a/src/NUnitConsole/nunit3-console/EnginePackageSettings.cs
+++ b/src/NUnitConsole/nunit3-console/EnginePackageSettings.cs
@@ -21,7 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-namespace NUnit.Engine
+namespace NUnit
 {
     /// <summary>
     /// EngineSettings contains constant values that
--- a/src/NUnitConsole/nunit3-console/EnginePackageSettings.cs
+++ b/src/NUnitConsole/nunit3-console/EnginePackageSettings.cs
@@ -1,5 +1,5 @@
-// ***********************************************************************
-// Copyright (c) 2014 Charlie Poole
+﻿// ***********************************************************************
+// Copyright (c) 2016 Charlie Poole
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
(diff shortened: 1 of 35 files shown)


src/NUnitEngine/nunit.engine/Services/DomainManager.cs
```diff
@@ -120,7 +118,7 @@ AppDomainSetup CreateAppDomainSetup(TestPackage package)
                 }
             }
 
-            if (package.GetSetting("ShadowCopyFiles", false))
+            if (package.GetSetting(PackageSettings.ShadowCopyFiles, false))
             {
```
ChrisMaddock: ShadowCopyFiles came up as only being set in console which seemed odd - this should be using the setting rather than a string.

CharliePoole:
I'm not sure about changing the content of this file in each usage, so it only has the settings that are currently in use by the particular module. Except in the framework, all the settings are available to every using program, even if not currently used. It seems more helpful to have a single file that documents all the settings we use and how we use them, copying in a new version when the old one changes.

However, I'd rather have a second opinion on this.... @rprouse what do you think?

ChrisMaddock:
Hmm ok. I went that way, as on the issue you suggested that future settings could only be added where they're needed. i.e. if a new setting is only relevant to the framework/console, it wouldn't need to be added to the engine. If we're going to go that way (where we'll let them get out of sync) - then starting off with everything in sync and therefore various unused settings feels like it might add further confusion. 

I'd be in favour of **either** having everything everywhere, in sync, and working to keep it that way, **or** having only what's required in each component. The former sounds like it will become significantly more work when the files are spread across multiple repos.

ChrisMaddock:
(Branch name failed the Appveyor)

CharliePoole:
@ChrisMaddock I definitely see your point, which is why I'd like some more opinions. :-)

I can see the framework perhaps only having that small set of stuff it uses. I can even see the engine with that set it uses itself. But it seems like somebody writing a runner would want to be able to grab a file that contains everything both the engine and the framework are capable of understanding.

If we wanted to split this up into multiple files, there would be
 * Settings used by the framework (set by the runner or engine)
 * Settings used by the engine (set by the runner)
 * Internal engine settings (set by one part of the engine, used by another part)

The framework would need the first file. The engine would need all three. The runners would need the first and second files.

That's another way to go... let's see what @nunit/core-team thinks.

ChrisMaddock:
That sounds like a sensible middle ground to me! :-)

CharliePoole:
@ChrisMaddock BTW, in the larger discussion I mentioned this... files we replicate files may become shared again in the future, using some other means of sharing across projects.

ChrisMaddock:
@rprouse - Any thoughts? I prefer @CharliePoole's idea of three separate files at this point - but will wait to hear before moving on.

rprouse:
As @charliepoole says, I am fine with separate files for now and removing unnecessary items in each. I expect console will have most since it sets them from command line options, then engine sets a couple more. Actually, I believe we had two copies early on when the console was in another repo.

CharliePoole:
@ChrisMaddock Please go ahead as we discussed, splitting into separate files for framework and engine, based on who uses the settings. If you find anything that doesn't fit into that mold, please ask.

FWIW, since you are a committer, it only takes one other person to work with you and agree on how things are going to be done. So you won't usually here from both me and @rprouse on a particular issue unless there is some sort of ongoing discussion about the direction to take. If you want added comments on an issue, you can ask for them by name right away... you'll notice I do that sometimes.

ChrisMaddock:
Thanks both. Will have a look at some point this week

> FWIW, since you are a committer, it only takes one other person to work with you and agree on how things are going to be done. So you won't usually here from both me and @rprouse on a particular issue unless there is some sort of ongoing discussion about the direction to take.

Sure - I presumed this fell in to the latter category however. The last we talked in this thread you pinged Rob for his opinion, so it seemed sensible to wait for that before continuing. :-)

ChrisMaddock:
Second version ready to take a look at.

This breaks it up in to three classes, 

- `InternalEnginePackageSettings` - in engine only
- `EnginePackageSettings` - in engine and runners only
- `FrameworkPackageSettings` - in framework and runners only

Settings have been grouped solely based on the region they were in in the old PackageSettings.cs. Accordingly, there are some settings in EnginePackageSettings which are currently only used internally in the engine - I wasn't sure if they should remain exposed to potentially future runners...perhaps better for someone with a better overview to look at.

Namespaces: either NUnit.Engine or NUnit.Framework as expected - does mean a file in the NUnit.Framework namespace is now in the console runner, do we want that, or should it be moved to NUnit.ConsoleRunner?

Accessibility: `InternalEnginePackageSettings` is now internal, the others seemed like they should stay public?

CharliePoole:
Broken build is due to length of branch name.

src/NUnitConsole/nunit3-console/EnginePackageSettings.cs
```diff
@@ -21,43 +21,16 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-namespace NUnit.Common
+namespace NUnit.Engine
```
CharliePoole: If we don't want to maintain the NUnit.Common namespace, then I would add these to the NUnit namespace.

CharliePoole:
Looks good to me. I'd be inclined to put the two public classes in the NUnit namespace and leave the internal class in NUnit.Engine. I think we did that with Guard and Env, didn't we?

@rprouse You may want to look at the FrameworkPackageSettings class for use in the core runner.

ChrisMaddock:
Change made. :+1:

CharliePoole:
Something broke. :-(

ChrisMaddock:
Yeah, that last commit was dumb. Attempt number two...

ChrisMaddock:
Appveyor failure was the branch name, and Travis had timed out, so just set it off again.

CharliePoole:
Still that nasty problem with special version length. It's a nuget problem, but I think I'll fix our script to work around it for now. They don't have any intention of fixing it soon.

Anyway, this can be merged.

Issue #1575, opened 2016-06-15
Escaping control chars in custom message


Chisamba:
Assert.That<TActual>(TActual actual, IResolveConstraint expression, string message, params object[] args)

In this method Nunit escapes control chars in message.
Can you explain why it's necessary? And can I turn off it?

CharliePoole:
NUnit doesn't escape control characters in the user message by design, so if it's happening it's a bug. Can you provide a repro test case?

CharliePoole:
BTW, as a part of the repro, please indicate where the erroneous output appears. We've already had a separate report of this but were never able to find out that information.

CharliePoole:
No need to repro, I see the problem. Somebody fixed issue #41 "Check for zeroes in assert messages" by leveraging a method designed for a different purpose, which escapes __all__ control characters. Specifically, the intent of that method was to be used for the display of expected versus actual values, in which case it's necessary to see all control characters in order to make a visual comparison.

Clearly, we have to escape nulls, even in the user-provided message, in order to avoid errors in the display. Any thoughts on what else - if anything - we should escape?  The method currently being used is here: https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Constraints/MsgUtils.cs#L314

Chisamba:
No idea what else :)

CharliePoole:
OK, we'll figure out something. I suggest you review the PR when it's available.

lpszBuffer:
Hey guys, let me know if you think PR #1659 looks suitable.

13xforever:
Would it be possible to push this fix as a 3.4.2 hotfix on nuget?

rprouse:
@13xforever we will not release this as a hotfix, but if you need this fix, you can use our CI NuGet feed until the next release. To do so, add a NuGet.config file to the root of your solution and add our CI NuGet feed to it as an [additional package source](https://docs.nuget.org/consume/nuget-config-file#package-sources). The address for our feed is, https://ci.appveyor.com/nuget/nunit

Once you do that, if you select pre-release in the package explorer, you can upgrade NUnit to version 3.5.0-CI-3020-master which contains this fix.

jnm2:
Wow, awesome! I could really use this fix.

PR #1659, opened 2016-07-11
Fix: Escaping control chars in custom message #1575

lpszBuffer:
Adds EscapeNullCharacters to MsgUtils to specifically escape only null characters from a string.
Alters TextMessageWriter's WriteMessageLine to utilize new EscapeNullCharacters method instead of EscapeControlCharacters.

### DIFF
--- a/nunit.sln
+++ b/nunit.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.25123.0
+VisualStudioVersion = 14.0.25420.1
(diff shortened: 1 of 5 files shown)


CharliePoole:
Looks good to me.

rprouse:
This looks good and very straight forward. Still, we are a test framework, so I would like to see unit tests in `MsgUtilTests` and at least one in `TextMessageWriterTests` to show that this does fix the issue. :smile:

lpszBuffer:
I added the unit tests, but it seems my attempt to squash my commits did not end well. :(

rprouse:
No need to squash. I'll take a look.

rprouse:
Thanks for the contribution.

PR #1660, opened 2016-07-12
Merge 3.4.1 Changes back into master

rprouse:
We need to get this back into master before we forget what we did for the hotfix release :smile: 

@CharliePoole I've branched off the 3.4.1 branch, merged latest from master and resolved conflicts. Could you go through this and make sure nothing is getting reverted by mistake and what you see here is what you would expect to come in from the hotfix?


src/NUnitEngine/Addins/teamcity-event-listener/TeamCityEventListener.cs
```diff
@@ -31,10 +31,10 @@
 namespace NUnit.Engine.Listeners
 {
     // Note: Setting mimimum engine version in this case is
-    // purely documentary since engines prior to 3.6 do not
-    // check the EngineVersion property and will try to 
+    // purely documentary since engines prior to 3.4 do not
+    // check the EngineVersion property and will try to
     // load this extension anyway.
-    [Extension(Enabled = false, EngineVersion = "3.4.0.0")]
+    [Extension(Enabled = false, EngineVersion = "3.4")]
```
rprouse: This is one of the changes that I wasn't sure about. The changes on master and on the release were different. I am assuming 3.4 is correct because that matches the unit tests.

CharliePoole: I think what you have is correct.

CharliePoole:
Looks good and tests are passing.

Issue #1661, opened 2016-07-12
Create .NET Standard Framework Build


rprouse:
.NET Standard is intended to replace portable libraries and gives us access to more APIs which will allow us to re-introduce features that were dropped in the PCL version of NUnit like Timeouts, Platform detection, File and Directory asserts, etc.

I am marking this as design as the planning needs more work, but I hope to complete this for the next release, so I am moving to the 3.5 milestone.

We will need to decide if this replaces the portable version or if we deprecate PCL for one release. We also need to figure out which build tools support building .NET Standard libraries.

rprouse:
In order to build a .NET Standard version of the library, developers will need to have at least Visual Studio 2015 Update 3 installed if we put the projects in the main solution. In order to build from the command line, they will need `dotnet` installed. Because of this, we should discuss approaches before I start working.

For installing `dotnet`, 

1 - the cake script could check if it is installed and install it if it is not. It is a small download. 
2 - we could do like we do for CF and check if it is installed and skip compile/test if it is not
3 - we could do option 2 and add a cake task to install dotnet

I prefer option 1 so that every developer is always building `netstandard`, but would like opinions from @nunit/core-team, @nunit/contributors and @nunit/net-core-runner-team 

For adding the projects to the solutions, as I said, it will require Visual Studio Update 3 or I think the latest versions of MonoDevelop, options are,

1 - Make running the latest versions of Visual Studio or Monodevelop a requirement for NUnit developers
2 - Create a separate solution for `netstandard`

I prefer option 1 because it will ensure all of our code is compiled and checked, but I can understand this being an issue for some developers, please speak up.

If we go with option 1, one thing I would like to explore is the new multi-targeting. For example, we could have only one copy of the NUnit Framework project that targets .NET 2.0, 3.5, 4.0, 4.5 and NetStandard. It would make our solution a lot smaller and more manageable. For an example of this, you can look at the https://github.com/nunit/dotnet-test-nunit project that targets `netstandard1.3` and `.NET 4.5.1`. 

The single project file also has the advantage in that it includes the NuGet packaging info so we could also drop our NuSpec files.

As I see it, we have three decisions to make,

1 - Should we automatically install `dotnet`
2 - Should we require Visual Studio 2015 Update 3+ or latest MonoDevelop for developers
3 - Should we merge each of our targets into one project

I plan to start working on this issue as soon as we do the repository split, so I thought we should get the discussion going. Let the jousting begin :smile:

[...]

Issue #1426, opened 2016-04-20
Make internal classes internal


rprouse:
NUnit has many public classes in various internal namespaces that are not intended for public consumption nor are they documented for use, but people keep using them and posting questions about them here and on StackOverflow.

We've begun the job of making those classes truly internal using InternalsVisibleTo, but we should complete the job.

CharliePoole:
It occurs to me that some of these are for use by custom attributes. We'll have to be careful.

rprouse:
Should we move classes that are intended for consumption by custom attributes out to a different namespace then?

CharliePoole:
Maybe so. I think this means a lot of our restrictions will have to be at the method level.

CharliePoole:
I took easyfix off since it isn't really something newcomers can undertake.

Should we figure out some way to split this up? One namespace at a time, perhaps?

CharliePoole:
Review of visibility by namespace. NUnitLite uses some of these so we would have to use the `InternalsVisibleToAttribute` to allow it to continue to do so if we make them internal.

#### Builders
Most classes are potentially usable by a hypothetial custom attribute although we didn't design them with that intention. I think these have to be gone through class by class, with each one examined for its present and future use. A hard one.

#### Commands

This is easier. Some of the classes are intended as base classes for custom commands, others not. They still have to be reviewed one at a time, of course, but the majority can probably be internal.

#### Execution
The vast majority of these are truly internal. However, a few have been exposed so we will have to look at those exposures and either make changes or leave them public.

#### Filters 
These classes can probably all be made internal. It would be good to look at the use of `[Serializable]` at the same time. I don't see why filters would need to be serializable any longer. In the past, we used passed them from the runner via remoting but we no longer do that.

#### Logging
`Logger`, `InternalTraceLevel` and `InternalTraceWriter` can probably be made internal. If we want user-defined attributes to log, we would have to leave InternalTrace.GetLogger as public but otherwise I think the entire class can be made internal.

#### Results
We need to go through all usage of each result and determine whether use of an interface is better in each case. If we once replace the majority of the non-internal usage with the interface, then we can consider making the visibility internal. Alternatively, we may find that these classes are needed by extensions, in which case they have to be given appropriate visibility - could be just leaving as is - or an extensibility interface (IWritableTestResult?) should be provided.

#### Tests
Same approach as Results, just more of it. 😸 

#### Internal Itself
[...]

PR #1666, opened 2016-07-14
Remove AssemblyHelper from Common

ChrisMaddock:
More on #1596.

This one was a bit more gnarly, so I'll detail what's changed.

The engine change was simple - it maintains the original namespace, and lots of unused methods were removed.

The framework Assembly helper just tidies, and removes a the `GetAssemblyPath(Type type)` method, which was only used in tests - and even then, inconsistently. (Many places already used `AssemblyHelper.GetAssemblyPath(typeof(MyType).Assembly))` - I just updated the remaining two to do the same, and removed the overload.)

The framework tests were more messy:

1. `GetNameForAssembly()` was missing a `[Test]` attribute so wasn't being run. Once I'd added this, I then removed the line testing for FullName, as a) it required version number/public key etc and b) seemed to be more a test of `System.Reflection.AssemblyName` than any NUnit code.

2. The tests were excluded for both SL and Portable. As the methods were used in both builds, I adapted it to run as much as possible for SL. I left stuff excluded for portable based on #1661

Finally - I'd have ideally made these classes internal, but they currently need to be public for the tests. Should we use `[InternalsVisisbleTo]` instead? I didn't do that here as it's really a separate discussion.

### DIFF
--- a/src/NUnitEngine/nunit.engine.tests/Internal/AssemblyHelperTests.cs
+++ b/src/NUnitEngine/nunit.engine.tests/Internal/AssemblyHelperTests.cs
@@ -1,0 +1,75 @@
+// ***********************************************************************
+// Copyright (c) 2016 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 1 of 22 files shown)


ChrisMaddock:
(Also, this was originally branched off my previous pull request, I attempted to rebase it, but it looks like I've merged in master instead. Is that an issue?)

rprouse:
> (Also, this was originally branched off my previous pull request, I attempted to rebase it, but it looks like I've merged in master instead. Is that an issue?)

The diff in the PR looks correct so far. That is what I'm concerned about :smile:

rprouse:
Looks good to me. Waiting for CI.

CharliePoole:
Looks good. Waiting for CI to finish.

As far as using InternalsVisibleTo for the tests, we already do that for the framework so I see no problem doing it for the engine. I think I'd check first to see if there are a lot of public classes that would benefit from it - doesn't seem useful to do it just for this one class.

CharliePoole:
@rprouse Looks like you're already on this!

ChrisMaddock:
Didn't realise it was already in the framework, could have made it internal there in that case. I was thinking of #1426 when doing it - was that issue referring more the the framework then?

Never mind - I'll know for next time!

CharliePoole:
#1426 is there because it takes a bit of effort to figure out what can be made internal in the framework. Testing doesn't actually enter into it.

PR #1669, opened 2016-07-14
Remove Logging classes from Common

### LINKED ISSUE #1596: Eliminate code sharing across projects to be split

ChrisMaddock:
Excluding what's been moved into #1621, I believe this fixes #1596. Phew! 

There's nothing of note I came across with this one.


rprouse:
Looks good

PR #1675, opened 2016-07-15
Limit length of pre-release suffix

CharliePoole:
I got tired of seeing failures in AppVeyor!

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -130,6 +130,10 @@ Task("InitializeBuild")
 				else
 					suffix += "-" + AppVeyor.Environment.Repository.Branch;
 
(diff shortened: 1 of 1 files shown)


rprouse:
Simple enough fix. About time someone did something :+1:

CharliePoole:
I'll merge after CI passes.

I've noticed that the IoS build takes a very long time. The mono 3.28 build is also slow, although not as bad. Should we limit these to certain branches or in some other way?

rprouse:
I think that running all builds on PRs and master is important. We could pair it down for branches.

I see that I still have Travis turned off as a required CI Build. I will turn it back on now that it is working again.

PR #1680, opened 2016-07-16
Remove console and command-line options from Common

### LINKED ISSUE #1621: Create command-line client helper assembly

### LINKED ISSUE #1677: Console help should indicate extension needed to use --teamcity
Right now this is the only option that is conditionally available. If we end up with more, we should probably group them in the help output.

CharliePoole:
Fixes #1621 Fixes #1677

This removes the last files from Common, a step toward separating console/engine from framework project. Also adds notation to console help indicating that an extension is required to use the `--teamcity` option.

NOTE: CommandLineOptions class continues to exist in both framework and console. It may eventually be merged into ConsoleOptions and NUnitLiteOptions respectively but I'm keeping it as a separate file for the present.


ChrisMaddock:
@CharliePoole - all looks good to me. I presume we're keeping the NUnit.Common namespace alive for these duplicated files?

Only nitpick - looks like there's now some redundant preproc directives in Options.cs in Console/NUnitLite. Will leave you to decide whether they're worth fixing, or to merge without. :+1:

PR #1682, opened 2016-07-16
Add TC listener option to portable nunitlite

### LINKED ISSUE #1681: NUnitLite under .net core doesn't support TeamCity output
Hi,
When compiling a project with NUnitLite under a full .NET Framework, the `--teamcity` option exists.

When trying to run the same project compiled for .NET core, the `--teamcity` option is gone.

Digging into the code, @ [CommandLineOptions.cs:379](https://github.com/nunit/nunit/blob/8927ec6f6c4977531b241749adc08170b3af8f91/src/Common/nunit/CommandLineOptions.cs#L379) and the surrounding area, it's pretty clear that under PORTABLE compilation, the `--temcity` command line option, and perhaps more code around it is left out of the compilation of the NUnit Engine...

Is it possible to bring back more options when compiling with NUnitLite?

Perhaps, it for general Portable build adding the teamcity addin is impossible, it might make sense to provide explicit `netstandard` builds of NUnit that add back more functionality lost with completely `PORTABLE` builds?

CharliePoole:
This fixes #1681.

The `--teamcity` option may now be used with the portable build of nunitlite.

### DIFF
--- a/src/Common/nunit/CommandLineOptions.cs
+++ b/src/Common/nunit/CommandLineOptions.cs
@@ -59,9 +59,8 @@ namespace NUnit.Common
         {
             // Apply default options
             if (defaultOptionsProvider == null) throw new ArgumentNullException("defaultOptionsProvider");
-#if !PORTABLE
+
             TeamCity = defaultOptionsProvider.TeamCity;
-#endif
             
             ConfigureOptions();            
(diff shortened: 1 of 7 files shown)


rprouse:
Looks good.

PR #1691, opened 2016-07-21
Issue 1683

FantasticFiasco:
Mark `SerializableAttribute` as `internal` instead of `public` to prevent build error when consumers of NUnit them self has declared a `SerializableAttribute` shim in their project.

### DIFF
--- a/src/NUnitFramework/framework/Compatibility/System/SerializableAttribute.cs
+++ b/src/NUnitFramework/framework/Compatibility/System/SerializableAttribute.cs
@@ -28,7 +28,7 @@ namespace System
     /// A shim of the .NET attribute for platforms that do not support it.
     /// </summary>
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Enum | AttributeTargets.Delegate, Inherited = false)]
-    public sealed class SerializableAttribute : Attribute
+    internal sealed class SerializableAttribute : Attribute
     {
     }
 }
--- a/src/NUnitFramework/nunitlite/Options.cs
+++ b/src/NUnitFramework/nunitlite/Options.cs
@@ -491,7 +491,7 @@ namespace Mono.Options
         }
     }
 
-#if !PORTABLE
+#if !SILVERLIGHT && !PORTABLE
     [Serializable]
 #endif
     public class OptionException : Exception
--- a/src/NUnitFramework/nunitlite/TestSelectionParserException.cs
+++ b/src/NUnitFramework/nunitlite/TestSelectionParserException.cs
@@ -30,7 +30,7 @@ namespace NUnit.Common
     /// TestSelectionParserException is thrown when an error 
     /// is found while parsing the selection expression.
     /// </summary>
-#if !PORTABLE
+#if !SILVERLIGHT && !PORTABLE
     [Serializable]
 #endif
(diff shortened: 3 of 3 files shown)


CharliePoole:
Silverlight build is failing. Did you build the entire solution locally?.. Or are you perhaps not running on Windows?

FantasticFiasco:
I was building using `build.cmd -t Build`, and it seemed to work, but I'll look into the issue and get it green. I'll give you a note if I need more help.

Btw, thanks for the quick feedback.

CharliePoole:
@FantasticFiasco Try `build -t Test` :-)

FantasticFiasco:
Sorry for the failed build, I was missing Silverlight support in Visual Studio and the Cake scripts didn't highlight the information about skipping that target type in red, my bad.

CharliePoole:
Do you need somebody with Silverlight support to pull it down and work on it?

FantasticFiasco:
No, I've updated my Visual Studio setup and enabled Silverlight, thus the projects can now be loaded. AppVeyor and Travis seems happy with the changes.

jcansdale:
Is Silverlight still a thing? Where is it being used?

CharliePoole:
Up to now, we have never use InternalsVisibleTo for anything but testing. I think we should pause a moment before extending it to allow nunitlite to see the framework internals.

On the pro side, nunitlite exists in a one to one relationship with a particular framework build. They are almost one assembly and at one point in the past, they actually were one assembly. That could be a reason for taking this step, provided we don't start taking further steps.

On the con side, we made the two separate for a reason. Perhaps we should honor that separation. Additionally, we have talked about making nunitlite independent of a particular build of the framework, although we have never followed through on it.

I notice we already have some code that brackets `[Serialization]` in an #if. We could obviously do that for NUnitLite, which only has this one usage of it.

@rprouse What's your thought on this?

CharliePoole:
@jcansdale Not abandoning users who are stuck with out-of-date software is how we distinguish ourselves from Microsoft. :-)

CharliePoole:
OTOH, to @jcansdale 's point, I'd rather not expose internals just for Silverlight.

jcansdale:
> Not abandoning users who are stuck with out-of-date software is how we distinguish ourselves from Microsoft. :-)

I hear you. Up until today I still supported Visual Studio 2008. Then I find out this:
https://twitter.com/Johnmont/status/756166750379450368

Do many browsers still support Silverlight? I know Chrome doesn't.

FantasticFiasco:
I understand that exposing the internals can be bad, especially if the projects have been designed to be independent. How about copying or linking the attribute instead?

I think the important aspect is to exclude all shims from the public surface of NUnit. It is not the responsibility of NUnit to provide those.

CharliePoole:
The `SerializableAttribute` is used 24 times in the frameworkassembly, only twice in nunitlite. Both uses in nunitlite are #if'ed out for PORTABLE but not for SILVERLIGHT. I think the simplest is to drop `InternalsVisibleTo` and change those two cases to

```
#if !SILVERLIGHT && !PORTABLE
```

FantasticFiasco:
I've updated the code according to your feedback, and the CI servers are happy. Thank you for your participation in this PR, I am pleased to see that an essential dependency of mine is in good hands.

CharliePoole:
Thanks for your work!

Issue #1595, opened 2016-06-21
Separate installer project


CharliePoole:
Assuming we are going to continue with at least some msi installers, they should be in a separate project repo. There is already a separate solution in the nunit repo, so the main change will be in designating where the installer finds the source files to include in the installers.

There seems to be no reason to have separate projects for _each_ installer, however. They can remain together in one project.

ChrisMaddock:
@CharliePoole - what so you see this looking like? 

The simple method would be to just have an empty dir in the new repository, where the user would manually copy the files that they want packaging up. 

Alternatively, could we have nuget do the heavy lifting? So we'd have a packages.config which always pulls the latest packages. (Or allows a specific version to be specified if required.) Then the release process would just involve releasing the nuget packages before building a new installer.

CharliePoole:
@ChrisMaddock I think that would be a strong possibility. Alternatively, we could have "feeds" that use other formats, like simply zipping all the files. That would mean less knowledge needed in the installer about how each item is structured within the nuget package.

This issue is on hold at the moment while @rprouse and I have been discussing it offline. There are a few issues...

1. If we are not going to release any more msis (my favorite option) then removing this from the nunit repo is all we have to do.

2. If we are not going to release a separate msi for the framework, then we can cut this down in size.

3. Depending on who is going to maintain it, we may want to structure it in different ways. That may just be my own issue - if I'm going to do it, I don't want a VS solution because I find that VS makes WiX harder to use, not easier. But somebody else may feel differently and should get to do it as they prefer.

For the moment, we have so much stuff going on that this has  basically been just sitting there.

rprouse:
What if we drop the framework installer and just leave the engine/console installer in that repo? I think relying on installed libraries is bad development practice, so we should follow xUnit's lead and only distribute via NuGet (or the Zip file for NuGet haters). I am fine with installing the console runner since it is a tool not a library, but I wouldn't mind dropping it either.

So, I vote for 2, but would be fine with 1. Personally, I think 3 is too much of a hassle for the few people who use it.

rprouse:
To add some stats, looking back at 3.0.1 since it was out the longest, 

- the MSI had 42,865 downloads (12%)
- the ZIP had 13,810 (4%)
- the NuGet package for the framework had 290,142 (84%)

[...]

PR #1695, opened 2016-07-24
Refactor dependencies in build.cake

### LINKED ISSUE #1610: Refactor dependencies in build.cake
The biggest problem here is that some actions depend on Build while others don't. I'd like to use this issue to discuss and decide how we want things to work for devs who build, test and package from the command-line and then implement any changes.

Key issues:

 * Packaging does _not_ depend on Build. I did it that way initially because I was working on packaging a lot and didn't want to rebuild each time in order to test packages. We could have a BuildAndPackage command, of course. Or we could make Package depend on Build and have a JustPackage or PackageOnly or TestPackage command.

 * Individual framework tests like Test45 do not depend on building. All other test commands depend on the full build, whether they need it or not. We should decide how we want this to work in a standard way. For me, if I am modifying and testing the console runner, I don't want to wait for every framework to rebuild.

@nunit/core-team Please add your comments, including any other issues you see.

CharliePoole:
Fixes #1610 

As discussed in #1610, I made the following changes:

* Tasks are now split between framework, engine and adapter where possible in order to make it easier for us to split the repository.

* Tests of each framework depend on building the framework.

* Test of all the frameworks depends on building all the frameworks. This extra dependency is actually not needed, but it keeps the builds all together in the output.

* A source package is no longer created.

* Added build of the teamcity extension, which was missing.

What isn't done in this PR...

* Packaging does **not** depend on build so as to facilitate repackaging. We can address this after the split.

* The zip and msi packages can't be split by component since they combine multiple components.

* Packaging of the CF build is still separate. I think we may want to change this. We build and run tests on CF if it's installed, so it makes sense to also package it in that case. We may eventually split out CF, but so long as it's part of the nunit project we should treat it the same as everything else, if only to make the script easier to understand.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -278,7 +278,7 @@ Task("BuildEngine")
         BuildProject("./src/NUnitEngine/Addins/vs-project-loader/vs-project-loader.csproj", configuration);
         BuildProject("./src/NUnitEngine/Addins/nunit-v2-result-writer/nunit-v2-result-writer.csproj", configuration);
         BuildProject("./src/NUnitEngine/Addins/nunit.v2.driver/nunit.v2.driver.csproj", configuration);
-        BuildProject("./src/NUnitEngine/addins/teamcity-event-listener/teamcity-event-listener.csproj", configuration);
+        BuildProject("./src/NUnitEngine/Addins/teamcity-event-listener/teamcity-event-listener.csproj", configuration);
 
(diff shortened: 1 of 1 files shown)


ChrisMaddock:
All looks good to me - going to merge this so I modify the cake script for #1595 without collision. :-)

PR #1699, opened 2016-07-25
Separate Runner/Framework installers

### LINKED ISSUE #1595: Separate installer project

ChrisMaddock:
Fixes #1595. This separates out the framework and runner installers. We never reached a consensus over whether to move this all to a separate repository - so I've made the changes in place for now, with the view that we could then move it out with history intact. This way around - it's also not blocking #1598.

### Upgrade Paths
As discussed, I changed what we currently use as the installer to install engine/console only, and pulled out a separate framework installer. There was previously an existing engine/console installer which is now gone - I changed the main nunit one instead so that upgrades will be in place and remove the old framework dlls.

### Appveyor
Msi's were commented out from the Appveyor artefacts, was there a reason for that? I re-enabled them.

### TeamCity
As the teamcity plugin won't be in the new repo, I removed it from the installer. Does that make sense, I've assumed TC setup's would use nuget packages rather than installations? If it sounds contentious, perhaps we should pull the nuget package to use in the msi instead.

To go with this, it was also necessary to remove the TC dll from `nunit.engine.addins` - which as far as I can see is only used for the MSI - is that correct, or could this break something?

Finally, TC.dll will no longer be installed in a fresh installation - but won't be removed with an upgrade (like the framework will be). I presume this is because the `.wxi` was missing a GUID for the TC dll - I'm not sure there's anything we can do about that retrospectively.


appveyor.yml
```diff
@@ -15,7 +15,7 @@ test: off
 
 artifacts:
 - path: package\*.nupkg
-#  - path: package\*.msi
+- path: package\*.msi
```
CharliePoole: I commented this out in the face of uncertainty over whether we would continue to have msi packages. Not a serious issue here, but would be if in our more public feeds.

build.cake
```diff
@@ -766,12 +766,31 @@ Task("PackageZip")
...
+            .WithProperty("DisplayVersion", version)
+            .WithProperty("OutDir", PACKAGE_DIR)
+            .WithProperty("InstallImage", IMAGE_DIR + "NUnit-" + packageVersion)
+            .SetMSBuildPlatform(MSBuildPlatform.x86)
+            .SetNodeReuse(false)
+        );
+    });
+
+
+Task("PackageFrameworkMsi")
```
CharliePoole: I thought we agreed to eliminate an msi for the framework.

build.cake
```diff
@@ -766,12 +766,31 @@ Task("PackageZip")
         Zip(currentImageDir, File(ZIP_PACKAGE_SL), zipFiles);
     });
 
-Task("PackageMsi")
+
+Task("PackageRunnersMsi")
```
CharliePoole: I don't like the name, as we are only packaging one runner, the Console runner. We have already deprecated the NUnit.Runners nuget package in favor of NUnit.Console, so I don't think  we want to create another package we may subsequently deprecate.

runners-msi/common/common-directories.wxi
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
CharliePoole: For these wxi files, it's almost impossible to see what has changed because the paths have changed as well. Could you put the paths at least temporarily back as they were and only change if necessary at the end of the review?

If that's too much work, just say so and I'll download and use some local tools to compare. However, it's always good to play to the lowest common denominator in viewing diffs, which is the github online view.

build.cake
ChrisMaddock: @rprouse's suggestion was to not distribute the framework msi, but keep the files around for one release in case we needed to put it in. Because of that I added the CI build, so we know if we broke it - but I'd be happy removing it entirely and resurrecting it if need be.

Let me know which you'd prefer. :-)

build.cake
ChrisMaddock: Sure - `Runners` was the name for the existing engine/console msi. If we do remove the framework msi (see above) - we could just revert to PackageMsi?

Runners-Msi is also used in the file structure for now to distinguish from framework-msi for now - these can perhaps be renamed again once the repos are split?

runners-msi/common/common-directories.wxi
ChrisMaddock: It would be tricky, as the one original directory was split into two new ones. Is listing them a help instead? I went for two distinct subdirs, to make splitting easier.

The only changes to wxi's specifically are removing the TeamCity addin from `runner-features.wxi` and `runner-files.wxi`.

The only real Wix changes in general are the above, and the merge of `runners.wxs` and `nunit.wxs`.

In general, there were previously three installers. NUnit (everything), Runners (console/engine) and Framework. Framework is untouched, NUnit is removed, and Runners replaces NUnit, with the framework and TC extension removed.

build.cake
CharliePoole: The thing is, with the change you made  to appveyor.yml, it will be automatically distributed on myget. So I'd say just comment out the msi deployment for now.

ChrisMaddock: Makes sense, we can eventually have it as an artefact for the console/engine repo and not the framework? Will sort it tomorrow - it's getting on a bit UK time!

runners-msi/common/common-directories.wxi
CharliePoole: I'm OK with the changes then except for one thing...

We stopped distributing Runners a while back so if it has any files that are __not__ also in master, they could be out of date. But if it's all the same files, we should be OK.

CharliePoole:
Final question... what guids should be be using for the console msi? Same as master? Same as old Runners? How will this impact users wanting to upgrade? @rprouse What do you think?

runners-msi/common/common-directories.wxi
ChrisMaddock: That's checked, they're all good. :+1: 

What I actually did was deleted the runners project, and adjusted NUnit-Everything to essentially be what runners used to be. But the above made a clearer explanation!

ChrisMaddock:
My thoughts (and what I've done) is continued the GUID from the master. As far as I know, that's the main msi in release, and the one where we want to 'replace' installations, and remove any old framework files kicking around.

Do we know when Framework/Runners were last distributed? I'd never seen them before.

CharliePoole:
That makes sense to me. 

The last time we distributed the framework/runners msis was with NUnit 3.0 Beta 5! That's why I thought we were pretty agreed on dropping them completely. In fact, removing framework from "master" as we are essentially doing is a bigger change for users than dropping the msis, which we have never distributed in a 3.x production release.

CharliePoole:
This PR applies to both the framework and the engine. It must be merged or closed before we split per #1598. 

@rprouse Waiting for comments from you on the direction.

build.cake
rprouse: I agree with @CharliePoole, drop the packaging from here and just keep the WiX files around for now in case we need them. I would also rename `PackageRunnersMsi` to just `PackageMsi`.

nunit-install.sln
```diff
@@ -1,41 +0,0 @@
-﻿
-Microsoft Visual Studio Solution File, Format Version 12.00
```
rprouse: Can't we just stick with `nunit-install.sln` instead of renaming it to runners? We are dropping the framework project from the solution, so there is no need for multiple solutions.

rprouse:
If we want the installer to upgrade the old version and not conflict, then I think we should go with the same GUID.

I had thought that we would just remove the Framework project and WiX files from the install solution and the master project rather that reorganize everything and split into two projects. Then we could just drop the framework files from one repo and the remaining from the other repo.

ChrisMaddock:
Sure - will work stuff around so that the old /install/ is as before, and the framework stuff is just orphaned. Think I've misunderstood what you intended by 'keep the files around'! Will look at it tomorrow.

ChrisMaddock:
Ok, all the changes are done, this is ready for review again. I'm afraid the diff is still a little ugly as git is mixing up the renames/deletions (is there any manual control over that?) - because of this I'll explain the changes.

- The framework files are now orphaned under `/framework-install/`. I've moved them to a) make the split cleaner and b) because `/common/` will need to be duplicated - that's done in this PR instead of needing to be worked out at the next stage.

- Git has mixed up the renames/deletions with runners/master. They are as before, I've just renamed the `master` subdir as the name didn't make much sense when there's only the one installer!

- Solution is back to being called nunit-install, but in the `/install/` subdir now to keep the top level tidier. Git also hasn't picked that up as being moved, and shows it as deletion/addition...

- TC has still been removed at the moment

- Appveyor artifacts are back on, as the framework msi isn't being built

ChrisMaddock:
Was thinking after our earlier conversation - if we're moving all the extensions into separate repos, the installer will almost certainly end up needing it's own repo too. I'd be happy to look at that - but perhaps we can merge this as is for now, to get it out the way of the big split. :-)

CharliePoole:
@ChrisMaddock I'll have time to review this in a few hours. Yes, maybe doing this in two stages is a good idea.

rprouse:
Still need to review the code, but your comments sound good. I also agree with the two stage approach to this

rprouse:
Looks good to me, but it should probably have a second set of eyes @CharliePoole.

@ChrisMaddock you also need to merge latest from master and resolve the conflicts.

CharliePoole:
Looks good to me too.

ChrisMaddock:
Thanks both. Have rebased, and will merge when CI has passed.

I'd like to have a look at separating this out into a separate repository next, now we've decided the engine extensions are moving out. I think setting it up to use NuGet lastest seems sensible for now - we can worry about GUIs and such later. Is that ok?

CharliePoole:
@ChrisMaddock Separating this into a different repository was the original intent of this issue when I wrote it. At Rob's request, we kept it together, at least for now. I don't think we want to create a new issue that blocks the "big split" especially now that I've started work on it!

For the future, I don't think we will want a new repo just for msi installs. We may want a new repo for creating mixed distributions with various nunit and nunit-related components. That might include msi installs, or it might not. So, in retrospect, I think Rob had the right idea for us to do the absolute minimum right now in order to enable the split. If we do any more for msis right now, we may be investing time in a technology that we end up abandoning. So applying JIT decision-making seems like the right thing to do.

I may come up with a proposal for an nunit distribution or bundling project in a few months, after we have released the 3.4.1 adapter and successfully split the framework from the engine.

ChrisMaddock:
Sure - I just thought the previous decision was based on the addins staying the the same repo - which it now seems they won't be? Meaning to package an installer in it's current state, you'll now need to copy and paste all the dlls into the image dir separately.

I was hoping this wouldn't block the 'big split' at all - that was part of the motivation for doing it now, that the work could be in a separate repo while this one was being divided up. Not a problem to leave it in it's current state for now, however.

CharliePoole:
Removing the addins isn't new, but removing __all__ of them definitely is. They will have to be copied somewhere to produce a new msi that includes them. But making a new repo means copying a __lot__ more files.

My assumption is that you would be trying to solve all the copying issues - not exactly a trivial task - before I'd be able to split the repos. Remember... there cannot be any outstanding PRs that cover both parts of our existing repo at the moment when we split. I've already started (locally) doing the split. If you have a plan that would __not__ block that split, I'd be all for it. Can you lay it out either here or in a new issue?

ChrisMaddock:
@CharliePoole Sure, have done it as #1716.

Issue #1693, opened 2016-07-21
2 unit tests fail due to localization


FantasticFiasco:
The following unit tests fail on my computer due to localization:

`NUnit.Framework.Tests.TestParametersTests.GetDecimal_ValueIsCorrect`
`NUnit.Framework.Tests.TestParametersTests.GetDouble_ValueIsCorrect`

Here in Sweden we write a numerical value with decimals as _1,234_ instead of _1.234_.

I would happily produce a PR that will make the tests pass indifferent of the localization.

rprouse:
Sure, we always appreciate the help.

CharliePoole:
Your help on this would be appreciated. We should first decide how it should work.

When NUnit produces messages, it generally uses the InvariantCulture. Should we expect parameters to use invariant cultulre or local culture? Our tests are expecting InvariantCulture to work, but that's not what we are using. Should we change the or should we change the implementation?

ChrisMaddock:
I'd personally be a fan of enforcing invariant culture on the command line - I'd want the same command line to work in the same way on different machines. 

That said, en-GB is pretty similar to Invariant, so not too painful for me, maybe @FantasticFiasco would think differently?

rprouse:
I think that all formatting in our messages should be invariant too because it is in the XML results and I don't want it to change based on the CI server my tests are running on. 

That said, invariant is also easy for me because it is similar to en-CA, so I would like to to get the opinion of non-English speakers.

FantasticFiasco:
I also think that invariant culture is the way to go. Current culture will always burn you on the CI server since it almost never run the same culture as the developers might have on their machines.

Regarding the culture setup for non-English developers, I think we have a complicated picture there. I don't think there is one configuration that everyone use, it comes down to personal preference. Me for instance, I run with _CurrentCulture=sv-SE_ and _CurrentUICulture=en-US_ to get English UI but numbers, currency and time displayed in the format I am used to (what's up with AM and PM anyway ;-))

CharliePoole:
In that case, nothing to fix here?

FantasticFiasco:
If we want to enforce invariant culture, shouldn't we specify that in method `TestParameters.Get<T>(string, T)`? The method is currently written like this:

```
public T Get<T>(string name, T defaultValue)
{
    string val = Get(name);
    return val != null ? (T)Convert.ChangeType(val, typeof(T), null) : defaultValue;
}
```

[...]

PR #1703, opened 2016-07-25
Issue 1693

FantasticFiasco:
Change method `TestParameters.Get<T>(string , T)` to use invariant culture instead of current culture.

### DIFF
--- a/src/NUnitFramework/framework/TestParameters.cs
+++ b/src/NUnitFramework/framework/TestParameters.cs
@@ -1,6 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Text;
+using System.Globalization;
 
 namespace NUnit.Framework
 {
@@ -9,6 +9,8 @@ namespace NUnit.Framework
     /// </summary>
     public class TestParameters
     {
+        private static readonly IFormatProvider MODIFIED_INVARIANT_CULTURE = CreateModifiedInvariantCulture();
+
         private readonly Dictionary<string, string> _parameters = new Dictionary<string, string>();
 
         /// <summary>
@@ -78,7 +80,7 @@ namespace NUnit.Framework
         public T Get<T>(string name, T defaultValue)
         {
             string val = Get(name);
-            return val != null ? (T)Convert.ChangeType(val, typeof(T), null) : defaultValue;
+            return val != null ? (T)Convert.ChangeType(val, typeof(T), MODIFIED_INVARIANT_CULTURE) : defaultValue;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


CharliePoole:
Looks OK to me although I'd like to see a test that passes in the wrong culture.

@rprouse?

FantasticFiasco:
Do you mean making sure that a non-English culture is set and the test should pass anyway?

CharliePoole:
No... sorry for the ambiguity: "passes in" == "supplies" :-) In such a case an exception should be thrown.

CharliePoole:
This PR applies to the framework only. See #1598.

FantasticFiasco:
I haven't made any new changes yet because I've thought about the problem, and especially after your last comment about verifying that an exception is thrown if a specific culture format is used.

As always, the devil is in the details, and I apologize upfront for the amount of baby-sitting you are forced to provide me with in this issue.

In an ideal world an exception should be thrown when an specific culture format is used, but in many cases the supplied format will be valid, but the _intent_ of the code is misinterpreted. Lets start with a test that in an ideal world should throw an exception due to a specific format:

```
[Test]
public void GetDouble_BadFormatThrowsException()
{
    _parameters.Add("SomeParm", "1,234");  // 1,234 is a number in e.g. sv-SE
    Assert.That(() => _parameters.Get("SomeParm", 99.9), Throws.TypeOf<FormatException>());
}
```

This test will fail. Why? Because `1,234` is in invariant culture read as `1234`. The comma is in invariant format no longer a _number decimal separator_ but a _number group separator_ ([source](https://msdn.microsoft.com/en-us/library/system.globalization.numberformatinfo.invariantinfo.aspx)). This is why I made a statement that the intent of the code is misinterpreted.

Back to your remark about writing a new test that validates that an exception is thrown when a non-invariant culture is used. I can surely make up a number format that is valid in some culture but throws an exception when trying to convert it using invariant culture, but the _value_ of that test is reduced due to the fact that many culture specific formats are valid in an invariant context.

What do you think?

Btw, I am happy to write the test, but am also a strong believer that all decisions should be backed with as much details about the consequences as possible.

CharliePoole:
I'm glad you brought this up! It seems to me that this could lead to a bit of confusion in many cases. Is there a format we can use that won't accept group separators?

CharliePoole:
BTW... this proves the benefit of tests... even if you just think about them! :-)

FantasticFiasco:
I don't know but I'll look into it.

There is no chance of moving away from having pure text as input? Having the parameters being backed by JSON would remove the confusion all together since the JSON number type is clearly defined.

CharliePoole:
Well, this particular feature is about typing a parameter at the command line or alternatively adding the value into a .runsettings file. So I don't think there's any chance of structured data in this case.

OTOH, there could be some other feature to be designed that uses JSON files for data to be used in the test.

FantasticFiasco:
I've updated the PR to remove comma (,) as group separator since it _might_ confuse developers that use comma as decimal separator. I would also like to make the following statement: Culture is probably more for end users than developers.

Microsoft and the compiler team made a great decision when they decided that source code should compile to the same binary indifferent on the culture used on the computer. I cannot specify a `double` using the _sv-SE_ culture, nor can I use comma as group separator in source code. If we can assume that developers are using the class `TestParameters`, then we probably can assume that most of them specify numbers as they would in source code.

CharliePoole:
This looks like a great solution to me! @rprouse if you agree let's merge this.

rprouse:
Looks good, thanks for the help. :+1: 

Fixes #1693

PR #1705, opened 2016-07-26
Eliminate error write in netcf

### LINKED ISSUE #1615: Appveyor error in TestCF
The TestCF section of the CI build is currently logging an `Immediate Error Message`. It seems to have been doing this for a while, although didn't throw the error when the `TestCF` task was originally added.

https://ci.appveyor.com/project/CharliePoole/nunit/build/3.3.0-CI-2887-issue-1607#L1302

I don't have the CF tools installed, so can't currently verify if this happens locally or not. It also appears to still be running the actual tests, so potentially isn't an important error, but we should probably investigate. (And look in to why it isn't error-ing the build, if it should be.)

CharliePoole:
Fixes #1615.

Error write in the netcf tests was not causing the appearance of a cake error in the CI build.

### DIFF
--- a/src/NUnitFramework/mock-assembly/MockAssembly.cs
+++ b/src/NUnitFramework/mock-assembly/MockAssembly.cs
@@ -136,7 +136,7 @@ namespace NUnit.Tests
             [Test]
             public void FailingTest()
             {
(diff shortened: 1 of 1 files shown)


ChrisMaddock:
Looks good!

PR #1709, opened 2016-07-27
Feature Request: Allow TextMessageWriter.DEFAULT_LINE_LENGTH to be updated

mwaddell:
Currently, TextMessageWriter.DEFAULT_LINE_LENGTH is a constant.  In this patch, I have changed it to  a class property.  This allows test classes (in the [SetUp] and [TearDown] methods) to change the default length length that will be used during the execution of a particular set of tests without the need to extend both the Assume and Assert classes.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
@@ -37,7 +37,7 @@ namespace NUnit.Framework.Internal
     public class TextMessageWriter : MessageWriter
     {
         #region Message Formats and Constants
-        private static readonly int DEFAULT_LINE_LENGTH = 78;
+        private static readonly int Default_Line_Length = 78;
 
         // Prefixes used in all failure messages. All must be the same
         // length, which is held in the PrefixLength field. Should not
@@ -57,7 +57,7 @@ namespace NUnit.Framework.Internal
         
         #endregion
 
-        private int maxLineLength = DEFAULT_LINE_LENGTH;
+        private int maxLineLength = Default_Line_Length;
 
         #region Constructors
         /// <summary>
@@ -79,6 +79,15 @@ namespace NUnit.Framework.Internal
         #endregion
 
         #region Properties
+        /// <summary>
+        /// Gets or sets the class-level default maximum line length for this writer
+        /// </summary>
+        public static int DefaultLineLength
+        {
+            get { return Default_Line_Length; }
+            set { Default_Line_length = value; }
+        }
+
         /// <summary>
         /// Gets or sets the maximum line length for this writer
         /// </summary>
--- a/src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
@@ -208,7 +208,7 @@ namespace NUnit.Framework.Internal
         /// <param name="clipping">If true, clip the strings to fit the max line length</param>
         public override void WriteValue(object val, bool clipping)
         {
-            Write(MsgUtils.FormatValue(val, clip));
+            Write(MsgUtils.FormatValue(val, clipping));
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
@@ -37,7 +37,7 @@ namespace NUnit.Framework.Internal
     public class TextMessageWriter : MessageWriter
     {
         #region Message Formats and Constants
-        private static int Default_Line_Length = 78;
+        private static int _defaultLineLength = 78;
 
         // Prefixes used in all failure messages. All must be the same
         // length, which is held in the PrefixLength field. Should not
@@ -57,7 +57,7 @@ namespace NUnit.Framework.Internal
         
         #endregion
 
-        private int maxLineLength = Default_Line_Length;
+        private int maxLineLength = _defaultLineLength;
 
         #region Constructors
         /// <summary>
@@ -84,8 +84,8 @@ namespace NUnit.Framework.Internal
         /// </summary>
         public static int DefaultLineLength
         {
-            get { return Default_Line_Length; }
-            set { Default_Line_Length = value; }
+            get { return _defaultLineLength; }
+            set { _defaultLineLength = value; }
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
@@ -120,7 +120,7 @@ namespace NUnit.Framework.Internal
         public override void DisplayDifferences(ConstraintResult result)
         {
             WriteExpectedLine(result);
-            WriteActualLine(result);
+            WriteActualLine(result, result.ClipStrings);
         }
 
         /// <summary>
@@ -131,10 +131,11 @@ namespace NUnit.Framework.Internal
         /// </summary>
         /// <param name="expected">The expected value</param>
         /// <param name="actual">The actual value causing the failure</param>
-        public override void DisplayDifferences(object expected, object actual)
+        /// <param name="clipping">If true, clip the strings to fit the max line length</param>
+        public override void DisplayDifferences(object expected, object actual, bool clipping)
         {
             WriteExpectedLine(expected);
-            WriteActualLine(actual);
+            WriteActualLine(actual, clipping);
         }
 
         /// <summary>
@@ -147,7 +148,7 @@ namespace NUnit.Framework.Internal
         public override void DisplayDifferences(object expected, object actual, Tolerance tolerance)
         {
             WriteExpectedLine(expected, tolerance);
-            WriteActualLine(actual);
+            WriteActualLine(actual, false);
         }
 
         /// <summary>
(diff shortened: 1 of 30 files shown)


src/NUnitFramework/tests/Internal/TextMessageWriterTests.cs
```diff
@@ -41,6 +41,23 @@ public void SetUp()
         }
 
         [Test]
+        public void TestMaLineLength()
```
rprouse: This method has a spelling mistake and you are testing many different (but related) things in this one test. I would prefer if it was broken up into discrete tests. I would also prefer if you saved off the current `DefaultLineLength` in `SetUp` and restore it in `TearDown` so that it has no chance of polluting other tests.

mwaddell:
Thank you for the feedback, Rob - I have addressed your concerns and will try to be less sloppy in the future.  -Michael

rprouse:
@CharliePoole can you take a quick look at this and give your opinion. It works, but I am wondering if there is a cleaner way to do this.

src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
```diff
@@ -37,7 +37,7 @@ namespace NUnit.Framework.Internal
     public class TextMessageWriter : MessageWriter
     {
         #region Message Formats and Constants
-        private static readonly int DEFAULT_LINE_LENGTH = 78;
+        private static int Default_Line_Length = 78;
```
rprouse: Now that this is no longer constant, please rename to _defaultLineLength.

mwaddell: Done.  FYI - my intention in this patch is to achieve something similar to the maxDiff property of python's unittest framework (https://docs.python.org/2.7/library/unittest.html).  I admit that there may be a cleaner way to achieve this.

rprouse:
My main issue with this PR is that it changes the setting globally and does not change it back. If I set a length in a `SetUp` method, would I expect it to be scoped to the duration of the test? Switching it mid-test run also seems wrong.

I think I would be more comfortable with the idea if it was scoped by the entire test run, possibly with an Assembly level attribute?

Sorry to be going back and forth like this, I should have stated my concerns before doing the code review, but the code review coloured my concerns. FYI, we prefer if you enter an issue before submitting a PR so that we can discuss the idea and the approach before you put too much work into it. We can deal with it as is, but I would ask you to work with us on this. It is important that we make changes in a way that does not confuse people, is intuitive to use and is maintainable over the life of the project.

CharliePoole:
I think there _may_ be a need for this feature as an adjunct to the exsiting `.NoClip` modifier, which entirely suppresses truncation of lines. But we should really discuss the precise need and how to expose a solution to users before settling on an approach. In particular, requiring users to make use of an internal class to change this doesn't make sense to me and a static member poses additional problems.

@mwaddell As a programmer, I fully understand what it means to see an immediate problem and just wanting to resolve it in code in a way that appears obvious to you. We on the nunit team use issues in part as a way to stifle our own urges to jump right in to the code. I've found that a little discussion of what we are trying to accomplish goes a long way in making sure we don't end up backing out or redoing changes. It's particularly important when we are exposing a new API to users.

In this case, we can have the same discussion here...

Where have you found this necessary? That is is it general or does it really only apply to certain assertions types? In the case of `.NoClip` it seemed to be a problem with EqualConstraint only, because truncated values made it hard to see where the error arose. So `.NoClip` was added to that particular constraint and not to others.

Would you want to apply this on a test case by test case basis? To all the tests in a fixture? Globally? What are the use cases for each of these - that is, when will other users want to do it?

Should this be a part of TestContext, so that it gets established at a certain level in the test hierarchy and then backed out? Should it be exposed as an attribute? Should it be in some class (as you have it) but possibly in a class that's already exposed to users? IOW what's the best and most sustainable interface?

CharliePoole:
@rprouse Looks like we have each arrived at the same questions by different routes. :-)

mwaddell:
@CharliePoole, Thank you for the feedback.  I understand the value of discussing issues thoroughly before accepting a code change into the trunk.  My main motivation in creating a PR was to have some context to discuss as opposed to discussing the issues in abstraction.  I just created an issue for this as well as a better place for this discussion.

The instances where I've found .NoClip to be insufficient are primarily when doing assertions on collections since these are the main instances where the truncation of additional data makes debugging more difficult.  If I have an "AllItemsAreNotNull" assertion and there are multiple nulls throughout a large array, then seeing them all at once versus just seeing the first couple and iteratively fixing my code to address each few.  If .NoClip were to work on all assertions (instead of just IsEqual), then I agree that there would be no need for this PR.  It may be nice to have something like .LargeClip() which would prevent a huge collection from spitting out MBs of content into your test output, but that's less of a need than a nice-to-have.

CharliePoole:
Not a problem. The main reason for discussion is to save you time. In this case it was a small change, but we should decide if we like the general "shape" of the change before trying to fine-tune the code.

I think the first thing to decide is what the user (ii.e. you) might want to do in the test code. With `.NoClip`, I felt that the would vary from assert to assert, so I made it a modifier to the constraint that is passed to each assert. Do we all think that's the case for an extension that limits clipping rather than eliminating it completely?

If that's so, I'd favor another modifier - maybe `.Clip(nnn)` or `.LineLength(nnn)` that could be added to the constraint. The modifier could be added to the base constraint class and used by individual constraints where appropriate. We already have some base-class modifiers that are used by constraints where they make sense and ignored in other cases.

We could also move `.NoClip` to the base class but we would need to make some code changes before any other constraints would pay attention to it.

OTOH, if we think this is something that should apply more globally, then we need to seek other approaches, such as an attribute or a TestContext setting. That's not how I'm leaning, however.

BTW, a constraint modifier has the advantage that it's discoverable through Intellisense.

@mwaddell @nunit/core-team What do you think?

mwaddell:
I'm in favor of being able to add .Clip(nnn) to any constraint.  It would make sense then to move .NoClip to the base class as well in order to maintain consistency between them.  

It is also worth noting that the existing .NoClip extension to EqualConstraint only applies to string differences, it is ignored in the DisplayCollectionDifferences, DisplayStreamDifferences, etc. -- so currently the NoClip in "Assert.That(list1, Is.EqualTo(list2).NoClip)" has no effect.

mwaddell:
I think that having NoClip moved to the base Constraint class and functional globally would solve the majority of use cases.

CharliePoole:
That makes sense to me. @rprouse What do you think?

As I indicated, moving it is trivial, making it actually do something is harder. But we can add more functionality incrementally, by implementing it for constraints where it makes sense.

mwaddell:
Since my own issues have been with collections, I looked into how collections are displayed and formatted a bit more.  Unfortunately, string clipping doesn't have anything to do with collections.  Displaying Actual/Expected collections uses MsgUtils.FormatCollection which is statically defined within DefaultValueFormatter with a max number of items of 10 (though in EqualConstraintResult.DisplayCollectionDifferences -- it's called with a max number of items of 3 instead).  In order to have a NoClip that prevents collections from being truncated would require changing how the MsgUtils class identifies a formatter to use for collections based on the constraint being tested.  In this case, perhaps an assembly-level attribute might be more appropriate since it would need to add an additional ValueFormatter to the head of the chain in the current TestExecutionContext.

Is there a better approach?

CharliePoole:
Our collection error display is notoriously bad, particularly for constraints other than equality.

Could you post some examples of bad displays and what you wish they looked like? If you're interested, I think we could make some real improvements here.

Presumably, however, the longer clip limit does help you with some other error displays. Right?

rprouse:
I would be okay with moving NoClip to the base constraint. Is the MsgUtils called from the collection constraints? Could we pass in the NoClip value to it?

rprouse:
@mwaddell how would you like to proceed with this? Would you like to continue to work on this PR, or enter an issue and start fresh with a new PR?

mwaddell:
@rprouse Sorry for the delay.  After considering a few different options, it seemed to me like changing the definition for the ValueFormatter delegate was the cleanest way to pass the NoClip information from the base Constraint class down the line to where it needs to go.  

There are currently no unit tests covering the formatting of collections/arrays/etc which I would like to add.  Also, the formatting of arrays does not have a "max" property like that of collections - is that intentional?

rprouse:
@mwaddell is this ready for another review?

rprouse:
@mwaddell this PR has been outstanding for a long time. Is it ready for review? If I don't hear back from you, I will be closing.

rprouse:
We haven't heard back on this, so I am closing.

FrankRay78:
Is there a better solution yet compared to explicitly putting NoClip on individual asserts?

I have an existing code base with many many asserts such as the following:

            Assert.AreEqual("CLC1,Client Limited Company 1,,5000,London,Mr Principal 1,MP1 Consulting Services Ltd,5000.00,0.00,0.00,5000.00,3500.00,0.00,500.00,0.00,Client Limited Company 1 Oct20,5882,MP1LTD1-1,MP1LTD1-2,", paymentLine.Export());

To allow full output, I can change it to the following:

            Assert.That(paymentLine.Export(), Is.EqualTo("CLC1,Client Limited Company 1,,5000,London,Mr Principal 1,MP1 Consulting Services Ltd,5000.00,0.00,0.00,5000.00,3500.00,0.00,500.00,0.00,Client Limited Company 1 Oct20,5882,MP1LTD1-1,MP1LTD1-2,").NoClip);

It seems wrong somehow to refactor 728 other assert statements with the NoClip attribute.

nb. Previously I was using MS Test (where this wasn't a problem) but recently switched to over to NUnit3.

[using NUnit3 in VS2019 with Nuget runner, resharper on Windows 10]

PS - I may consider having a go at mwaddell's last attempt at a PR if this is still warranted.

Issue #1707, opened 2016-07-27
TestContext.CurrentContext.Test.MethodName get Unnamed job in sauce labs


giang-hai-nguyen:
I got this issue when i want to get method name in sauce labs.
`caps.SetCapability("name", TestContext.CurrentContext.Test.MethodName);`
Pls help me if anyone has solution on this? thanks

Also, (maybe this is sauce labs issue instead) when i run a class and this class have 2 TCs (1 Pass, 1 Fail). However, This class marked Pass

CharliePoole:
@giang-hai-nguyen We would need more informatio to help you with this.

What's the context? Where does this line of code live? Is it in a test? In a setup method? somewhere else? How does sauce labs work with NUnit? Bear in mind that we know NUnit, but we (at least I) don't know much about what sauce labs is or how it works.

giang-hai-nguyen:
Hi Charlie,
I am new with sauce labs too. Here is the example for sauce labs (You need to register an account to fill info SAUCE_USERNAME & SAUCE_ACCESS_KEY to run this example). In the setup method, i called the `TestContext.CurrentContext.Test.MethodName` (get Method Name) of Nunit but it doesn't work.
Noticed that in this case, `TestContext.CurrentContext.Test.Name` method (get Class Name) of Nunit works fine. Thanks
```
using NUnit.Framework;
using System;
using Selenium;
using System.Web;
using System.Text;
using System.Net;
using OpenQA.Selenium;
using OpenQA.Selenium.Remote;
using OpenQA.Selenium.Support.UI;
 
namespace Saucey_Selenium {
    [TestFixture("chrome", "45", "Windows 7", "", "")]
    public class SauceNUnit_Test
    {
        private IWebDriver driver;
        private String browser;
        private String version;
        private String os;
        private String deviceName;
        private String deviceOrientation;
 
        public SauceNUnit_Test(String browser, String version, String os, String deviceName, String deviceOrientation)
        {
            this.browser = browser;
            this.version = version;
            this.os = os;
            this.deviceName = deviceName;
            this.deviceOrientation = deviceOrientation;
        }
 
        [SetUp]
        public void Init()
        {
            DesiredCapabilities caps = new DesiredCapabilities();
            caps.SetCapability(CapabilityType.BrowserName, browser);
            caps.SetCapability(CapabilityType.Version, version);
            caps.SetCapability(CapabilityType.Platform, os);
            caps.SetCapability("deviceName", deviceName);
[...]

PR #1714, opened 2016-07-28
Additional tests of TestContext and TestExecutionContext

CharliePoole:
Issue #1707 was closed as invalid but in investigating it I found we were missing certain test cases in TestContext and TestExecutionContext. Here they are.


rprouse:
Looks good

PR #1717, opened 2016-07-31
Change suffix for master builds

CharliePoole:
For builds of master (normally after a merge) change the suffix for both the AppVeyor and MyGet feeds from `-CI-nnnn-master` to `-DEV-nnnn` (without -master).

This is intended to make the builds stand out better in both feeds and to make all master build versions higher than any other builds in the AppVeyor feed.

Considerations for review:
* Should we use five digits rather than four?
* Should we use lower-case rather than upper?

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -113,22 +113,31 @@ Task("InitializeBuild")
 			}
 			else
 			{
-				var buildNumber = AppVeyor.Environment.Build.Number;
+				var buildNumber = AppVeyor.Environment.Build.Number.ToString("00000");
+				var branch = AppVeyor.Environment.Repository.Branch;
+				var isPullRequest = AppVeyor.Environment.PullRequest.IsPullRequest;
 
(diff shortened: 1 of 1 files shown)


rprouse:
Looks good to me, I restarted the AppVeyor build that failed for unrelated reasons.

I prefer lowercase, but I don't really care that much :smile: 

As for number of digits, will AppVeyor just bump it up to 5 digits once we roll over?

CharliePoole:
I'll change to lower case.

Appveyor will give us as many digits as it needs, but `3.5-dev-10000` sorts before `3.5-dev-9999`

CharliePoole:
I see there is a problem here in that the PR is coming out as a master build becuase I'm now checking the branch name before I check to see if it's a PR. I don't think we want that since PR builds are really only hypothetical master builds.

CharliePoole:
Unfortunately, merging is the only way to  see how it works when the branch is master. :-(

PR #1718, opened 2016-08-03
Minimize use of agents in loading tests for execution

### LINKED ISSUE #1712: ConsoleRunner spawns too many agents
@Trass3r commented on [Thu Jul 28 2016](https://github.com/nunit/docs/issues/93)

```sh
packages/NUnit.ConsoleRunner.3.2.0/tools/nunit3-console.exe" [lots of *UnitTest.dlls]  --work=... --out=TestOutput.log --result=TestResult.xml --labels=On --framework=net-4.5 --dispose-runners --agents=8
```

* spawns 1 nunit-agent.exe per dll even though we specified --agents=8
* those agents are created serially which takes a really long time and tests aren't run until this creation phase finishes

CharliePoole:
Fixes #1712 

This change causes the loading of tests to be delayed until the assembly is actually about to be executed. As a result, loading takes place in parallel for up to the number of agents specified.

Note that loading is still sequential if the client (e.g. a gui) actually asks for tests to be loaded or explored. Further changes could be considered to allow those loads to proceed in parallel. Since this would require significantly more code to be changed,  this pr only aims to improve performance for batch execution of tests.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -130,7 +130,7 @@ Task("InitializeBuild")
 					else if (AppVeyor.Environment.Repository.Branch.StartsWith("release", StringComparison.OrdinalIgnoreCase))
 						suffix += "-pre-" + buildNumber;
 					else
-						suffix += "-" + AppVeyor.Environment.Repository.Branch;
+						suffix += "-" + branch;
 
 					// Nuget limits "special version part" to 20 chars. Add one for the hyphen.
 					if (suffix.Length > 21)
--- a/src/NUnitConsole/nunit3-console.tests/MakeTestPackageTests.cs
+++ b/src/NUnitConsole/nunit3-console.tests/MakeTestPackageTests.cs
@@ -134,13 +134,12 @@ namespace NUnit.ConsoleRunner.Tests
         //}
 
         [Test]
-        public void WhenNoOptionsAreSpecified_PackageContainsWorkDirectorySettingOnly()
+        public void WhenNoOptionsAreSpecified_PackageContainsOnlyTwoSettings()
         {
             var options = new ConsoleOptions("test.dll");
             var package = ConsoleRunner.MakeTestPackage(options);
 
-            Assert.AreEqual(1, package.Settings.Count);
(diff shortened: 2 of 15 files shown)


rprouse:
@CharliePoole this looks good to me, but it is a big change. Did you do manual integration testing with more projects than the level of parallelism and observe the behaviour in process monitor? I assume a max number of agents running at one time?

@nunit/core-team is anyone else available to give this a second review? We normally don't do second reviews for team members, but I would like a second set of eyes on this one if someone can help out.

CharliePoole:
@rprouse I did but my tests were limited: three projects with various numbers of agents. It could use more. In the process of working on it I also modified the logging so I could observe exactly when the tests were loaded and unloaded and it confirmed that they work as expected.

Here's a further explanatory comment on the changes to the runner structure, which I had intended to include in the description...

MasterTestRunner is actually significantly different form all other runners. It implements an interface that no other runners implement. However, because of how things were originally developped, it was still sharing a base class and interface with the other (internal) runners. This was getting a bit complicated, so I separated the code of MasterTestRunner completely from the internal runners. It no longer inherits from AbstractTestRunner and no longer implements ITestEngineRunner, the internal interface. So now MasterTestRunner implements ITestRunner and holds an internal runner to to its work. Only the internal runners implement ITestEngineRunner. Hope that's clear enough. :-)

@nunit/core-team I'd welcome any exercising you can give this. I'd also like users to exercise it, which is why I'd like to get it merged and into the myget feed. We have a number of users who are now following that feed and they may discover problems we should fix before the 3.5 release.

rprouse:
@CharliePoole thanks for the more detailed explanation, the changes are a bit easier to grok with that info :smile: 

If nobody on the @nunit/core-team has time to review today, let's merge. As you say, it needs as extensive testing as possible, plus it will hold up the console/engine repo split which I would like to get behind us as soon as possible.

CharliePoole:
I'll merge as soon as this becomes blocking for me, which will probably be in a few hours.

rprouse:
I'm good with that :+1:

PR #1720, opened 2016-08-04
Include teamcity extension in our runner packages

### LINKED ISSUE #1719: Restore teamcity extension to NUnit.Console and NUnit.Runners packages
Even though the extension is in a separate repo, we still want to include the latest version of it as a dependency for our runner packages, as requested by JetBrains.

CharliePoole:
Fixes #1719

### DIFF
--- a/nuget/runners/nunit.console-runner-with-extensions.nuspec
+++ b/nuget/runners/nunit.console-runner-with-extensions.nuspec
@@ -19,10 +19,11 @@
       * VSProjectLoader        - loads tests from Visual Studio projects
       * NUnitV2ResultWriter    - saves results in NUnit V2 format.
       * NUnitV2FrameworkDriver - runs NUnit V2 tests.
+      * TeamCityEventListener - supports special progress messages used by teamcity.
 
       Other extensions, if needed, must be installed separately.
     </description>
-    <releaseNotes>This release removes the TeamCityEventListener extension, which must now be installed separately.</releaseNotes>
+    <releaseNotes>This release uses the latest version of the TeamCityEventListener extension.</releaseNotes>
     <language>en-US</language>
     <tags>nunit test testing tdd runner</tags>
     <copyright>Copyright (c) 2016 Charlie Poole</copyright>
@@ -33,6 +34,7 @@
         <dependency id="NUnit.Extension.VSProjectLoader" version="$version$" />
         <dependency id="NUnit.Extension.NUnitV2ResultWriter" version="$version$" />
         <dependency id="NUnit.Extension.NUnitV2Driver" version="$version$" />
+        <dependency id="NUnit.Extension.TeamCityEventListener" version="1.0" />
       </group>
     </dependencies>
   </metadata>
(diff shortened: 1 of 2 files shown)


CharliePoole:
I need to merge this so it will deploy to MyGet for final testing.

ChrisMaddock:
Should the TC version be 1.0.1 instead? I don't know what the difference was between 1.0 and 1.0.1, but I think this will currently use 1.0.

Looks good, otherwise :smile:

CharliePoole:
Latest version is 1.0.2, so it probably should be that one. That said, I'm going to merge in order to test further because we can't really test it as it is actually distributed until it gets to MyGet and it won't reach MyGet unless I merge.

In fact, this is a great simulation of what will happen with separate teams - or maybe it __is__ separated teams! Nobody is working on the extension except Nikolay. Without looking in the repo, I don't know what the current version is. I can only specify the minimum version that makes NUnit happy. Nikolay, OTOH, could test with the latest console runner each time he makes a change. He could let us know what he wants in the combined package. I think living with this for a while will teach us something and I'm pretty confident that anyone who is using the myget feed should be able to live with any glitches that occur.

ChrisMaddock:
@CharliePoole: Hmm...ok. Although this still means Nikolay has to notice that this is the case, and then come to this repository, and submit a PR to update. (Assuming 1.0.1 contains tested changes.) I'm afraid I'm still not convinced - although I think we'll have to agree to disagree here!

A potential issue from just checking this out, the latest NuGet (non-prerelease) version is 1.0.1, where as on MyGet, it's 1.0.0. MyGet seems to not have the 1.0.1 release package, which could mean different behaviour between the two feeds. So if you *had* specified 1.0.1 here, then the package would have worked on NuGet, but not MyGet. Possibly something we need to sort? I'm not sure how packages are build for release.

CharliePoole:
Right. Everything you say would be an argument for not separating the teamcity extension. But we decided to separate it for non-technical reasons. We don't actually know how teamcity works or what it needs. It's better for people who understand what teamcity needs to work on it.

I originally pictured a total separation - I would donate the code to JetBrains and they would maintain it. We would not distribute it with NUnit at all. That's why it was removed. But the JetBrains guys thought that would be confusing to users, so I agreed we would try to accomodate them - at least in the short term.

The compromise is that they will maintain the code under the nunit organization. They will release it. We will incorporate their releases in at least some of our own packages. Details are obviously not yet worked out and I don't know how we will finally do this. It has to be worked out between both organizations as a "business" decision.

How they do their side is going to be up to them in the end. But if it were me, I would have a continuous integration going that notices when we update master and runs some tests. They probably need to run tests that combine at least two versions of the runner and engine (latest release and latest build) with the corresponding two versions of the extension.

When our own extensions are separated, I think things will be a lot simpler because we get to make the decisions on both sides. :-)

CharliePoole:
@ChrisMaddock Regarding 1.0.1 versus 1.0.2... All our releases are manual. That is, Rob or I builds locally, does some final testing and pushes the binaries. That's why they don't match our automated CI builds. Eventually, I hope it will all be automated but we don't even have CI working fully yet.

CharliePoole:
It seems to need the latest build in order to work. I think I'll need to fix that and also push a non-prerelease 1.0.2 just to get us started.

It looks to me like the Console and Runners packages are only really usable when they are fully released to NuGet unless we specify a greater than range. I'll fool around with this in a separate PR.

PR #1725, opened 2016-08-07
Improve appdomain unload error message

espenalb:
Related to my comment on nunit/nunit#1572. By adding a little bit more info in the runner, it is immediately clear which test assembly is the culprit.

The test output becomes:
`Exception encountered unloading AppDomain test-domain-common.tests.dll: Error while unloading appdomain. (Exception from HRESULT: 0x80131015)`

### DIFF
--- a/src/NUnitEngine/nunit.engine/Services/DomainManager.cs
+++ b/src/NUnitEngine/nunit.engine/Services/DomainManager.cs
@@ -166,7 +166,9 @@ namespace NUnit.Engine.Services
                 }
 
                 if (_unloadException != null)
-                    throw new NUnitEngineException("Exception encountered unloading AppDomain", _unloadException);
+                {
+                    throw new NUnitEngineException($"Exception encountered unloading AppDomain {_domain.FriendlyName}: {_unloadException.Message}", _unloadException);
+                }
             }
 
             private void UnloadOnThread()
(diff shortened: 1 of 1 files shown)


src/NUnitEngine/nunit.engine/Services/DomainManager.cs
```diff
@@ -166,7 +166,10 @@ public void Unload()
                 }
 
                 if (_unloadException != null)
-                    throw new NUnitEngineException("Exception encountered unloading AppDomain", _unloadException);
+                {
+                    var msg = string.Format("Exception encountered unloading AppDomain '{0}': {1}", _domain.FriendlyName, _unloadException.Message);
+                    throw new NUnitEngineException(msg, _unloadException);
```
CharliePoole: The problem I see here is that _domain.FriendlyName may throw an exception as well. This will happen, for example, if the original exception is thrown for an appdomain that no longer exists. In fact, if memory serves, that's why the message doesn't include the domain name.

You could probably work around this by changing the code inside the try to fetch the FriendlyName and saving it in a member. Your message would need to be different, depending on whether that member contained a name or not. Or you could initialize it to "UNKNOWN".

CharliePoole:
@espenalb Did you plan to follow up on the comment?

espenalb:
Yes - I will initialize to unknown - like it is allready done in another part of the code. Will update and push a change probably tomorrow.

rprouse:
@espenalb this PR has been idle for a few months now. If you are going to continue work on it, please let us know, otherwise I will close.

CharliePoole:
This is obsolete anyway, since the code is no longer in this repo. I think there is a new issue that covers the problem.

rprouse:
Thanks @CharliePoole I didn't even notice that this was an engine issue, I was just doing a quick pass on everything to clean up 😄 

Closing...

PR #1728, opened 2016-08-08
Include method and class name in XML for V2 tests when it is available

### LINKED ISSUE #1727: V2 Framework Driver does not include class or method name in XML
NUnit 2.6.4 added ClassName and MethodName properties that are used by the visual studio adapter in order to locate the source code for a test. The NUnit V2 framework driver does not include those fields in the NUnit 3 formatted XML it generates. This is needed in order for the V3 adapter to run V2 tests and provide source code locations.

CharliePoole:
**Do Not Merge Yet** Being tested with the adapter.

This fixes #1727. As explained there, the fix is needed in order for the NUnit 3 VS adapter to be able to find the source info for NUnit 2 tests. My current implementation doesn't find it at all since these fields are never available.


CharliePoole:
Failed build is the "usual" timeout on Apple. I restarted it.

rprouse:
Looks good

CharliePoole:
Verified with my adapter build - merging.

PR #1736, opened 2016-08-12
Ensure test gets loaded for all runners by moving code to the base class

### LINKED ISSUE #1732: NullReferenceException when using either --inprocess or --domain=Multiple options
Due to a problem in project set-up (it appears so) I receive NullReferenceException when running unit test with following stack trace:
```
Test Run Summary
  Overall result: System.NullReferenceException: Object reference not set to an instance of an object.
   at NUnit.Common.ColorConsoleWriter.WriteLabel(String label, Object option, ColorStyle valueStyle)
  at NUnit.Common.ColorConsoleWriter.WriteLabelLine(String label, Object option, ColorStyle valueStyle)
 at NUnit.ConsoleRunner.ResultReporter.WriteSummaryReport()
 at NUnit.ConsoleRunner.ConsoleRunner.RunTests(TestPackage package, TestFilter filter)
  at NUnit.ConsoleRunner.Program.Main(String[] args)
```
No other evidence of the problem are there in report or console.
When attached with VS to the nunit process I've managed to discover that the report is the following:
```
<test-run id="2" testcasecount="0" engine-version="3.4.1.0" clr-version="4.0.30319.42000" start-time="2016-08-10 12:48:48Z" end-time="2016-08-10 12:48:48Z" duration="0.049014">
  <command-line><![CDATA["F:\Projects\[censored]\./packages/NUnit.ConsoleRunner.3.4.1/tools/nunit3-console.exe" F:\Projects\[censored]\UnitTests.nunit]]></command-line>
 <test-suite type="Project" id="2" name="UnitTests.nunit" fullname="F:\Projects\[censored]\UnitTests.nunit" testcasecount="0" />
</test-run>
```
It appears NullReferenceException happens because ResultReporter relies on the fact that "result" property exists in test run which is not true:
```
public ResultReporter(XmlNode result, ExtendedTextWriter writer, ConsoleOptions options)
{
    _result = result;
    _writer = writer;

    _overallResult = result.GetAttribute("result");
    if (_overallResult == "Skipped")
        _overallResult = "Warning";

    _options = options;

    Summary = new ResultSummary(result);
}
```

Could we improve logging in this case?

CharliePoole:
Fixes #1732 

I've run a bunch of manual tests using various combinations of settings this time, so I believe this fixes any latent problems. We should talk about whether we want a 3.4.2 release or to simply accelerate 3.5. Alternatively, people would have to be told to run all tests using the default process and domain settings.

### DIFF
--- a/src/NUnitEngine/nunit.engine/Runners/AbstractTestRunner.cs
+++ b/src/NUnitEngine/nunit.engine/Runners/AbstractTestRunner.cs
@@ -167,6 +167,7 @@ namespace NUnit.Engine.Runners
         /// <returns>A TestEngineResult.</returns>
         public TestEngineResult Explore(TestFilter filter)
         {
+            EnsurePackageIsLoaded();
             return ExploreTests(filter);
         }
 
@@ -209,6 +210,7 @@ namespace NUnit.Engine.Runners
         /// <returns>The count of test cases.</returns>
         public int CountTestCases(TestFilter filter)
         {
+            EnsurePackageIsLoaded();
             return CountTests(filter);
         }
 
@@ -221,6 +223,7 @@ namespace NUnit.Engine.Runners
         /// <returns>A TestEngineResult giving the result of the test execution</returns>
         public TestEngineResult Run(ITestEventListener listener, TestFilter filter)
         {
+            EnsurePackageIsLoaded();
             return RunTests(listener, filter);
         }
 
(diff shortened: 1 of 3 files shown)


rprouse:
This would be really easy to cherry pick into the 3.4.1 branch to create a 3.4.2, so I would prefer to create a new 3.4.1 release. It only needs to be for the console/engine plus the MSI which could contain the 3.4.1 framework. A good reason to carry on with splitting our repos and dropping the framework from the MSI.

Code looks good, I restarted the hung Travis build.

CharliePoole:
I'm OK either way... If we do a 3.4.2 release, then we may want to not merge anything else until we do the split. We can consider whether to do a 3.5 release with just the split then.

CharliePoole:
So... I guess we want to go ahead and merge this into master before doing the cherry-picking. Right?

rprouse:
Yes, CI has passed. I will try to do a 3.4.2 this weekend.

PR #1740, opened 2016-08-14
Create separate mock-assembly for engine tests

### LINKED ISSUE #1739: Create separate copies of MockAssembly for framework, engine and extensions
The tests of the framework, the engine and several extensions use MockAssembly, which needs to be copied for use in each of the new repositories we are creating.

CharliePoole:
Fixes #1739


CharliePoole:
The build script wasn't building the new assembly. Fixed now and  ready to merge after review.

PR #1746, opened 2016-08-15
Attributes with description constructors

asbjornu:
It has annoyed me to no end that the `Description` on `TestFixtureAttribute` and `TestAttribute` don't have a corresponding constructor argument, meaning I always have to prefix descriptions for these two attributes with `Description = `. Seeing how the `DescriptionAttribute` and in some ways `CategoryAttribute` have the convenient constructors, I would love it if `TestAttribute` and `TestFixtureAttribute` could have them added as well.

It's not all sunshine and roses, though. Adding a `.ctor(string description)` to `TestFixtureAttribute` makes it collide with the `.ctor(params object[] arguments)` in cases where the latter is used with only one string parameter as such: `[TestFixture("First param argument")]`. I would still argue that having the constructor taking `description` is so useful that we should do something about it, and this PR provides what I hope are the grounds to discuss exactly **what** can be done.

@nunit/core-team Thoughts?

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestAttribute.cs
@@ -57,6 +57,26 @@ namespace NUnit.Framework
         private object _expectedResult;
         private readonly NUnitTestCaseBuilder _builder = new NUnitTestCaseBuilder();
 
+        #region Constructors
+
+        /// <summary>
+        /// Construct a test <see cref="Attribute"/>
+        /// </summary>
+        public TestAttribute()
+        {
+        }
+
+        /// <summary>
+        /// Construct a test <see cref="Attribute"/>
+        /// </summary>
+        /// <param name="description">Descriptive text for this test</param>
+        public TestAttribute(string description)
+        {
+            Description = description;
+        }
+
+        #endregion
+
         /// <summary>
         /// Descriptive text for this test
         /// </summary>
--- a/src/NUnitFramework/framework/Attributes/TestFixtureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestFixtureAttribute.cs
@@ -40,10 +40,19 @@ namespace NUnit.Framework
         #region Constructors
 
         /// <summary>
+        /// Constructs a test fixture <see cref="Attribute"/>
+        /// </summary>
+        /// <param name="description">Descriptive text for this fixture</param>
+        public TestFixtureAttribute(string description) : this()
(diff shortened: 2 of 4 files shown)


CharliePoole:
Use of a Description property on Test and TestFixture is a legacy of very old versions of NUnit. We have retained it so as not to arbitrarily break code. The general design approach for many years has been to use separate Attributes for separate purposes, hence the DescriptionAttribute.

Additionally, use as an argument to TestFixture is not feasible because it conflicts with the function of TestFixture's arguments for parameterized tests. If you use

```C#
[TestFixture("My description")
public class SomeClass
{
    ....
}
```

then we expect SomeClass to have a constructor that takes a string argument.

It is possible to make this change for TestAttribute, but I don't think we want to.

As a minor comment on your code, which may be useful in the future, you should not change MockAssembly unnecessarily. It is used by our tests that require loading an entire assembly and each change has ramifications all over the codebase.

asbjornu:
@CharliePoole Thanks for the feedback. Perhaps a good alteration of this PR would be to add `[Obsolete("Use [Description(\"...\")] instead"]` to the `Description` property? That would make the migration path a bit more obvious. I, for one, wasn't aware that `[Description]` was interchangeable with `TestAttribtue.Description`. Is this documented somewhere?

> As a minor comment on your code, which may be useful in the future, you should not change MockAssembly unnecessarily. It is used by our tests that require loading an entire assembly and each change has ramifications all over the codebase.

Noted. :+1:

CharliePoole:
Yes, I'd go for deprecating the property. @rprouse What do you think?

As far as I can see, the docs don't give any preference to one approach over the other. Perhaps they should. Also, they don't explicitly say that one approach is equivalent to the other. If you edit the docs in the wiki (I believe you have authority to do that) let me know. Otherwise, I'll take a closer look myself when I get a chance.

rprouse:
I think we should edit the docs to state that the two are equivalent to one another, but I am reticent to deprecate what is likely a widely used property without allowing the community to weigh in. Deprecating will introduce build warnings and that could create a large headache for shops that use one method over another. I also don't see it as any different than continuing to support classic asserts.

CharliePoole:
I updated the docs. Let's leave the code alone.

@asbjornu Thanks for your effort. Can I help you find something else to do? :-)

asbjornu:
@CharliePoole:
> I updated the docs.

Thanks!

> Can I help you find something else to do? :-)

Sure. How are we doing on GitVersion, btw?

PR #1750, opened 2016-08-15
Implement the "Big Split"

### LINKED ISSUE #1598: Split framework and console/engine into separate projects
This is the biggest split of the bunch! We will initially create separate repos for the framework and the engine plus console. For the moment at least the engine and console will stay together.

We will have to decide on the mechanics. Alternatives...
1. Copy both repos and delete what is not being kept, so both have a full history.
2. Export files to create one of the repos and delete files in the other. Only one will have the full history in that case.
3. Copy both repos as in (1) and then delete the entire history of the removed folders. This will give a full history for both repos, but only for the portion that is being retained. Each repo will appear as if it was always separate. This requires advanced techniques and the result has to be reviewed carefully to ensure nothing is lost but what we want to lose.

CharliePoole:
Fixes #1598 by removing Engine and Console subdirectories from the nunit repository. All past history is retained. Some further items may need to be removed.


rprouse:
There are a bunch of other things to remove, but we can clean them up as we go. Let's get this merged and put a stake in the sand.

elbaloo:
Don't forget to update this repo's description so it doesn't include console and engine anymore :P

CharliePoole:
@elbaloo Done! Thank you.

PR #1755, opened 2016-08-19
Removing IsAsynchronous

chris-smith-zocdoc:
IsAsynchronous was never set, so this expression `if (!Test.IsAsynchronous || timeout > 0)` in WorkItem.RunThread is always true. 

I'm not exactly sure of the intention behind IsAsynchronous was, but I believe this can safely be removed.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
@@ -311,46 +311,43 @@ namespace NUnit.Framework.Internal.Execution
 #endif
 
             thread.Start();
+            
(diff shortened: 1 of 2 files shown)


CharliePoole:
I'm nervous that we don't know why it was ever there, but it looks like we don't need it. I searched the history to find any indication it was ever set and didn't find one. I can only guess it was a field we thought we needed, never used and never removed. I'm OK to merge this.

chris-smith-zocdoc:
Looks like the appveyor build failed because of my branch name `3.5.0-ci-03147-cs_deadcode' is not a valid version string` Should I merge this anyway?

CharliePoole:
Done!

PR #1759, opened 2016-08-22
Fix typos (".Net" => ".NET")

ulrichb:
Reference: https://www.microsoft.com/net

Note that there were already other mentions of ".NET" (e.g. in `README.md`) which are now in sync.


rprouse:
Looks good, thanks. :+1:

Issue #1334, opened 2016-03-10
Add Assert.Zero and Assert.NotZero


oznetmaster:
I have been asked by a number of my users if these two forms of Assert could be added to compliment the addition of the Is.Zero and Is.Not.Zero constraints (#1246 and #1248).

It seems reasonable given that we have Assert.Null and Assert.NotNull.

Easy to do, but does this violate some philosophical issue?

CharliePoole:
Well, only that we have not been updating the collection of what we call legacy asserts (aka classical) for a long time. We actually discussed removing them entirely, but decided that we should keep them for compatibility reasons. I'm don't feel strongly, however. @rprouse,  what do you think?

oznetmaster:
Probably should be a separate issue, but there appears to be no documentation for the Zero constraint?

ChrisMaddock:
It's documented under the Equal Constraint. Perhaps sensible, unless you are explicitly looking for it?

https://github.com/nunit/docs/wiki/EqualConstraint

oznetmaster:
I think it should be in the alphabetical list of constraints.

rprouse:
We haven't been adding to the classic constraints and I don't use them anymore, but I have been looking at quite a few projects in the wild lately and my feeling is the majority of users are still writing classic asserts. I can sort of see why, they are shorter to write and you get intellisense earlier.

So, I am ok with adding limited and clear new methods, as long as it doesn't become a slippery slope where we support every combination of the constraint syntax :smile:

CharliePoole:
We didn't create a new constraint, so we can't document one. :-)

Maybe there should be a list of syntactic elements with links to the
constraints they create.

CharliePoole:
I think it depends on the size of the list. I would start with a single page, since it elminates one click.I created an issue in docs for this.

I'm OK with Assert.Zero and Assert.NotZero as well. Shall we add it to the backlog?

rprouse:
Labeled and added to backlog

oznetmaster:
I agree with @rprouse observation that "the majority of users are still writing classic asserts".  And for exactly the reasons he indicated, shorter, and intellisense.  I am also told that most find them more readable.

Has any consideration ever been given to making the classic asserts first class citizens again, and adding to them whenever it seems appropriate (like in the case of zero)?

CharliePoole:
I don't think we want to go back to adding new implementation code in the Assert class as we used to. There should always be a Constraint that actually implements the predicate in question.

[...]

PR #1767, opened 2016-08-24
Add tests to CF build

CharliePoole:
Adding new files to CF that were not included in the fix to #1334. I found one other test file that was not included and added it as well.


rprouse:
Looks good, just waiting on Travis. Feel free to merge when it is done.

rprouse:
Travis failure was a hung build. I am not going to restart since Travis doesn't run CF anyway. Merging.

PR #1775, opened 2016-08-30
Support NUnit assertions in partial trust code.

ggeurts:
NUnit 2 supported running test code with NUnit assertions in partial-trust. The same test fail however on NUnit 3. I have added a test fixture to the test projects for the .NET 2, 3.5, 4.0 and 4.5 platforms and security assertions to the NUnit.Framework assemblies to get the tests to run.

### DIFF
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -29,6 +29,7 @@ using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Reflection;
+using System.Security;
 using System.Web.UI;
 using NUnit.Compatibility;
 using NUnit.Framework.Interfaces;
@@ -123,6 +124,7 @@ namespace NUnit.Framework.Api
             _testAssembly = assembly;
         }
 
+        [SecuritySafeCritical]
         private void Initialize(string assemblyPath, IDictionary settings)
         {
             AssemblyNameOrPath = assemblyPath;
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -125,6 +125,10 @@ namespace NUnit.Framework.Api
         }
 
 #if !SILVERLIGHT && !NETCF && !PORTABLE
+        // This method invokes members on the 'System.Diagnostics.Process' class and must satisfy the link demand of 
+        // the full-trust 'PermissionSetAttribute' on this class. Callers of this method have no influence on how the 
+        // Process class is used, so we can safely satisfy the link demand with a 'SecuritySafeCriticalAttribute' rather
+        // than a 'SecurityCriticalAttribute' and allow use by security transparent callers.
         [SecuritySafeCritical]
 #endif   
         private void Initialize(string assemblyPath, IDictionary settings)
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -124,7 +124,9 @@ namespace NUnit.Framework.Api
             _testAssembly = assembly;
         }
 
+#if !SILVERLIGHT && !NETCF && !PORTABLE
         [SecuritySafeCritical]
+#endif   
         private void Initialize(string assemblyPath, IDictionary settings)
         {
             AssemblyNameOrPath = assemblyPath;
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -31,6 +31,7 @@ using System.Collections.Generic;
 using System.IO;
 #if !SILVERLIGHT && !NETCF && !PORTABLE
 using System.Diagnostics;
+using System.Security;
 using System.Windows.Forms;
 #endif
 
@@ -384,6 +385,7 @@ namespace NUnit.Framework.Api
 #endif
 
 #if !SILVERLIGHT && !NETCF && !PORTABLE
+        [SecuritySafeCritical]
         private static void PauseBeforeRun()
         {
             var process = Process.GetCurrentProcess();
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -385,6 +385,10 @@ namespace NUnit.Framework.Api
 #endif
 
 #if !SILVERLIGHT && !NETCF && !PORTABLE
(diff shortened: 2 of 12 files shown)


rprouse:
Your tests are failing on Linux. You will probably want to use the platform attribute to exclude those tests on anything but Windows.

```
1) Failed : NUnit.Framework.Assertions.LowTrustFixture.AssertThrowsInLowTrustSandBox
  Expected: <System.Security.SecurityException>
  But was:  null
at NUnit.Framework.Assertions.TestSandBox.Run (System.Reflection.MethodInfo method, System.Object[] parameters) <0x41597540 + 0x0023f> in <filename unknown>:0
at NUnit.Framework.Assertions.TestSandBox.Run (System.Action action) <0x415974d0 + 0x00056> in <filename unknown>:0
at NUnit.Framework.Assertions.LowTrustFixture.AssertThrowsInLowTrustSandBox () <0x415c7ff0 + 0x000a8> in <filename unknown>:0
```

rprouse:
Other than the test failure, this looks good to me, but I would like to get a second review from another @nunit/core-team member.

Since there is no associated issue, I have assigned a milestone and label for tracking. In the future, please create and reference an issue before submitting a PR. It helps us keep track for our release notes and allows discussion before you put too much work into an issue.

Thanks for your contribution...

CharliePoole:
@ggeurts In NUnit V2, we merely applied AllowPartiallyTrustedCallersAttribute to the framework and didn't use the SecuritySafeCriticalAttribute. What does the latter do for us in addition, in terms of user functionality?

The MS documentation states "Security-safe-critical code must validate the permissions of callers to determine whether they have authority to access protected resources used by the code." We don't particularly do any checking anywhere in our code. I'd like to know if using the attribute opens up any doors that we don't want to open.

I think it would be useful to us and to future contributors to have a comment along with each use of the attribute that indicates why that particular method needs the attribute.

CharliePoole:
In addition to the test failures in Linux, there are warnings indicating that your copy of the nunit-linux solution still contains references to console and engine projects. I assume this will be cleared up when we merge, however.

ggeurts:
@CharliePoole Just using the AllowPartiallyTrustedCallersAttribute works on the .NET 2 runtime, but is not sufficient on the .NET 4 runtime due to changes in the security model. The security attributes that have been added are the ones that are flagged up as missing when performing a Code Analysis (FxCop) run on the .NET 4.0 NUnit.Framework assembly. I will add comments why they were required.

src/NUnitFramework/framework/Api/FrameworkController.cs
```diff
@@ -123,6 +124,7 @@ public FrameworkController(Assembly assembly, string idPrefix, IDictionary setti
             _testAssembly = assembly;
         }
 
+        [SecuritySafeCritical]
```
ggeurts: Initialize method uses security critical Process,GetCurrentProcess() method.

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -384,6 +385,7 @@ private int GetLevelOfParallelism()
 #endif
 
 #if !SILVERLIGHT && !NETCF && !PORTABLE
+        [SecuritySafeCritical]
```
ggeurts: Method uses security critical Process,GetCurrentProcess() method.

src/NUnitFramework/framework/Compatibility/LongLivedMarshalByRefObject.cs
```diff
@@ -38,6 +39,7 @@ public class LongLivedMarshalByRefObject : MarshalByRefObject
         /// <summary>
         /// Obtains a lifetime service object to control the lifetime policy for this instance.
         /// </summary>
+        [SecurityCritical]
```
ggeurts: Method overrides security critical method in base class.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -241,6 +242,7 @@ public static TestExecutionContext GetTestExecutionContext()
         /// </summary>
         public static TestExecutionContext CurrentContext
         {
+            [SecuritySafeCritical]
```
ggeurts: Method calls security critical CallContext methods.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -252,6 +254,7 @@ public static TestExecutionContext CurrentContext
 
                 return context;
             }
+            [SecuritySafeCritical]
```
ggeurts: Method calls security critical CallContext methods.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -264,6 +267,7 @@ private set
         /// <summary>
         /// Get the current context or return null if none is found.
         /// </summary>
+        [SecuritySafeCritical]
```
ggeurts: Method calls security critical CallContext methods.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -550,6 +554,7 @@ public void AddFormatter(ValueFormatterFactory formatterFactory)
         /// Obtain lifetime service object
         /// </summary>
         /// <returns></returns>
+        [SecurityCritical]
```
ggeurts: Method overrides security critical method in base class.

src/NUnitFramework/framework/Internal/OSPlatform.cs
```diff
@@ -25,12 +25,14 @@
...
 using System;
 using System.Runtime.InteropServices;
+using System.Security;
 
 namespace NUnit.Framework.Internal
 {
     /// <summary>
     /// OSPlatform represents a particular operating system platform
     /// </summary>
+    [SecuritySafeCritical]
```
ggeurts: Method depends on security critical PInvoke operations

CharliePoole:
Actually, I was thinking of code comments but maybe the github comments are good enough. 

@rprouse What do you think?

ggeurts:
@CharliePoole I am happy to add them to the code instead.

CharliePoole:
Seems like a good idea to me... that's where future devs will find them.

Your build is still failing on Linux under Mono 3.2.8 because of a `nameof` in the test fixture. We limit ourselves to C# 3.0 features in order to be able to build on all platforms.

ggeurts:
@charliePoole Inline comments have been added and the build issue should be fixed now as well. That was an omission on my side. Thanks for the feedback!

rprouse:
Build is passing now, merging. Thanks for the contribution.

selfie-0
[_GitHub Selfies_](https://github.com/thieman/github-selfies/)

PR #1780, opened 2016-09-01
Made ExactCount standalone and added Items no-op

### LINKED ISSUE #1668: Need implementation-independent way to test number of items in a collection
Here's the issue:

``` c#
IReadOnlyCollection<object> collection = someApi.GetItems(...);
Assert.That(collection, Has.Count.EqualTo(n));
```

At the API's discretion, it may return `object[]` or `List<object>` or some completely internal type. The contract is that I am allowed to call `GetEnumerator()` and `Count`.

Currently, `Has.Count` uses reflection and thus is coupled to the internal type itself rather than the interface. This means that I have to know about the internal choice the API makes and choose `Has.Count` and `Has.Length` accordingly. This leaves all my tests explicitly reflecting an internal detail I shouldn't have to care about, and all my tests could break if the API has a non-breaking change to its internals.

`Has.Count` is also limited in that it doesn't examine the collection as a series of items. It examines it as an object itself, such as `new TimesIClicked { Count = 3 }`. This means that it in addition to being useless with arrays, it **can't be used with enumerables**, something I touched on in #1667.

Here are two possible alternatives I have.

**Case 1.** If `Has.Count` is semantically interested in not how many items a thing contains but what the property on an object reads, to the degree that they diverge:
- The `Count` property on arrays and other collections should be discovered and used even though it's only available via casting to `IReadOnlyCollection<>`, because it is certainly the property that is relevant to the user and `Count` is probably the property that is available at compile type which the user is looking at.
- This avoids all the irritations in the introduction, caused by coupling tests to knowledge (Length vs Count vs other/not available) that isn't part of the API contract for APIs returning collection interfaces.
- It will never work for enumerables, so an additional solution should be provided separately for them

**Case 2.** If `Has.Count` is semantically interested in how many items a thing contains rather than what a certain property on the object reads, to the degree that they diverge:
- It should only access the Count property via a collection interface, and fall back to enumeration if not available.
- This avoids all the irritations in the introduction, caused by coupling tests to knowledge (Length vs Count vs other/not available) that isn't part of the API contract for APIs returning collection interfaces.
- It will work with enumerables

Unless there is a very clear and compelling alternative API for collection counting assertions, for instance suggestions in #1667, users will try to put their Case 2 style burdens on `Has.Count` because it is named Count, just like collection.Count and LINQ's Count(). It makes sense. If Case 2 is not the way to go with `Has.Count`, a compelling API should be able to handle the collection-focused and enumerable-focused workloads.

jnm2:
Fixes #1668.

I already solved the issue with self resolving before I read your post, so these are all supported:

``` c#
Assert.That(collection, Has.Exactly(3).EqualTo(val)); // Current syntax
Assert.That(collection, Has.Exactly(3).Items.EqualTo(val);
Assert.That(collection, Has.Exactly(3));
Assert.That(collection, Has.Exactly(3).Items);
```

I took into consideration that you aren't trying to preserve compatibility.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
@@ -31,9 +31,19 @@ namespace NUnit.Framework.Constraints
     /// item in a collection, succeeding only if a specified
     /// number of items succeed.
     /// </summary>
-    public class ExactCountConstraint : PrefixConstraint
+    public class ExactCountConstraint : Constraint
     {
-        private int expectedCount;
+        private readonly int _expectedCount;
+        private readonly IConstraint _itemConstraint;
+
+        /// <summary>
+        /// Construct a standalone ExactCountConstraint
+        /// </summary>
+        /// <param name="expectedCount"></param>
+        public ExactCountConstraint(int expectedCount)
+        {
+            _expectedCount = expectedCount;
+        }
 
         /// <summary>
         /// Construct an ExactCountConstraint on top of an existing constraint
@@ -43,12 +53,10 @@ namespace NUnit.Framework.Constraints
         public ExactCountConstraint(int expectedCount, IConstraint itemConstraint)
             : base(itemConstraint)
         {
-            this.expectedCount = expectedCount;
-            this.DescriptionPrefix = expectedCount == 0
-                ? "no item"
-                : expectedCount == 1
-                    ? "exactly one item"
-                    : string.Format("exactly {0} items", expectedCount);
+            Guard.ArgumentNotNull(itemConstraint, "itemConstraint");
+
+            _itemConstraint = itemConstraint.Resolve();
+            _expectedCount = expectedCount;
         }
 
         /// <summary>
@@ -63,11 +71,36 @@ namespace NUnit.Framework.Constraints
                 throw new ArgumentException("The actual value must be an IEnumerable", "actual");
 
             int count = 0;
-            foreach (object item in (IEnumerable)actual)
-                if (BaseConstraint.ApplyTo(item).IsSuccess)
+            if (_itemConstraint == null)
+            {
+                foreach (object item in (IEnumerable)actual)
                     count++;
+            }
+            else
+            {
+                foreach (object item in (IEnumerable)actual)
+                    if (_itemConstraint.ApplyTo(item).IsSuccess)
+                        count++;
+            }
 
-            return new ConstraintResult(this, actual, count == expectedCount);
+            return new ConstraintResult(this, actual, count == _expectedCount);
+        }
+
+        /// <summary>
+        /// The Description of what this constraint tests, for
+        /// use in messages and in the ConstraintResult.
+        /// </summary>
+        public override string Description
+        {
+            get
+            {
+                var descriptionPrefix =
+                    _expectedCount == 0 ? "no item" :
+                    _expectedCount == 1 ? "exactly one item" :
+                    string.Format("exactly {0} items", _expectedCount);
+
+                return _itemConstraint != null ? PrefixConstraint.FormatDescription(descriptionPrefix, _itemConstraint) : descriptionPrefix;
+            }
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/ItemsConstraintExpression.cs
+++ b/src/NUnitFramework/framework/Constraints/ItemsConstraintExpression.cs
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2007 Charlie Poole
+// Copyright (c) 2016 Charlie Poole
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
@@ -24,50 +24,29 @@
 namespace NUnit.Framework.Constraints
 {
     /// <summary>
-    /// Abstract base class used for prefixes
+    /// An extension of ResolvableConstraintExpression that adds a no-op Items property for readability.
     /// </summary>
-    public abstract class PrefixConstraint : Constraint
+    public sealed class ItemsConstraintExpression : ResolvableConstraintExpression
     {
         /// <summary>
-        /// The base constraint
+        /// Create a new instance of ItemsConstraintExpression
         /// </summary>
-        protected IConstraint BaseConstraint { get; set; }
+        public ItemsConstraintExpression() { }
 
         /// <summary>
-        /// Prefix used in forming the constraint description
+        /// Create a new instance of ResolvableConstraintExpression,
+        /// passing in a pre-populated ConstraintBuilder.
         /// </summary>
-        protected string DescriptionPrefix { get; set; }
+        /// <param name="builder"></param>
+        public ItemsConstraintExpression(ConstraintBuilder builder)
+            : base(builder) { }
 
         /// <summary>
-        /// Construct given a base constraint
+        /// No-op property for readability.
         /// </summary>
-        /// <param name="baseConstraint"></param>
-        protected PrefixConstraint(IResolveConstraint baseConstraint)
-            : base(baseConstraint)
+        public ResolvableConstraintExpression Items
         {
-            Guard.ArgumentNotNull(baseConstraint, "baseConstraint");
-
-            BaseConstraint = baseConstraint.Resolve();
-        }
-
-        /// <summary>
-        /// The Description of what this constraint tests, for
-        /// use in messages and in the ConstraintResult.
-        /// </summary>
-        public override string Description
-        {
-            get { return FormatDescription(DescriptionPrefix, BaseConstraint); }
-        }
-
-        /// <summary>
-        /// Formats a prefix constraint's description.
-        /// </summary>
-        internal static string FormatDescription(string descriptionPrefix, IConstraint baseConstraint)
-        {
-            return string.Format(
-                baseConstraint is EqualConstraint ? "{0} equal to {1}" : "{0} {1}",
-                descriptionPrefix,
-                baseConstraint.Description);
+            get { return this; }
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/Operators/ExactCountOperator.cs
+++ b/src/NUnitFramework/framework/Constraints/Operators/ExactCountOperator.cs
@@ -29,7 +29,7 @@ namespace NUnit.Framework.Constraints
     /// Represents a constraint that succeeds if the specified 
     /// count of members of a collection match a base constraint.
     /// </summary>
-    public class ExactCountOperator : CollectionOperator
+    public class ExactCountOperator : SelfResolvingOperator
     {
         private int expectedCount;
 
@@ -39,17 +39,24 @@ namespace NUnit.Framework.Constraints
         /// <param name="expectedCount">The expected count</param>
         public ExactCountOperator(int expectedCount)
         {
+            // Collection Operators stack on everything
+            // and allow all other ops to stack on them
+            this.left_precedence = 1;
+            this.right_precedence = 10;
+
             this.expectedCount = expectedCount;
         }
 
+
         /// <summary>
-        /// Returns a constraint that will apply the argument
-        /// to the members of a collection, succeeding if
-        /// none of them succeed.
+        /// Reduce produces a constraint from the operator and 
+        /// any arguments. It takes the arguments from the constraint 
+        /// stack and pushes the resulting constraint on it.
         /// </summary>
-        public override IConstraint ApplyPrefix(IConstraint constraint)
+        /// <param name="stack"></param>
+        public override void Reduce(ConstraintBuilder.ConstraintStack stack)
         {
-            return new ExactCountConstraint(expectedCount, constraint);
+            stack.Push(stack.Empty ? new ExactCountConstraint(expectedCount) : new ExactCountConstraint(expectedCount, stack.Pop()));
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
@@ -198,9 +198,10 @@ namespace NUnit.Framework.Constraints
         /// the following constraint to all members of a collection,
         /// succeeding only if a specified number of them succeed.
         /// </summary>
-        public ConstraintExpression Exactly(int expectedCount)
+        public ItemsConstraintExpression Exactly(int expectedCount)
         {
-            return this.Append(new ExactCountOperator(expectedCount));
+            builder.Append(new ExactCountOperator(expectedCount));
+            return new ItemsConstraintExpression(builder);
         }
         
         #endregion
--- a/src/NUnitFramework/framework/Has.cs
+++ b/src/NUnitFramework/framework/Has.cs
@@ -95,7 +95,7 @@ namespace NUnit.Framework
         /// the following constraint to all members of a collection,
         /// succeeding only if a specified number of them succeed.
         /// </summary>
-        public static ConstraintExpression Exactly(int expectedCount)
+        public static ItemsConstraintExpression Exactly(int expectedCount)
         {
             return new ConstraintExpression().Exactly(expectedCount);
         }
--- a/src/NUnitFramework/framework/nunit.framework-2.0.csproj
+++ b/src/NUnitFramework/framework/nunit.framework-2.0.csproj
@@ -91,6 +91,7 @@
     <Compile Include="Constraints\FileExistsConstraint.cs" />
     <Compile Include="Constraints\FileOrDirectoryExistsConstraint.cs" />
     <Compile Include="Constraints\IConstraint.cs" />
+    <Compile Include="Constraints\ItemsConstraintExpression.cs" />
     <Compile Include="Constraints\Interval.cs" />
     <Compile Include="Constraints\Operators\AllOperator.cs" />
     <Compile Include="Constraints\Operators\NoneOperator.cs" />
--- a/src/NUnitFramework/framework/nunit.framework-3.5.csproj
+++ b/src/NUnitFramework/framework/nunit.framework-3.5.csproj
@@ -90,6 +90,7 @@
     <Compile Include="Constraints\FileExistsConstraint.cs" />
     <Compile Include="Constraints\FileOrDirectoryExistsConstraint.cs" />
     <Compile Include="Constraints\IConstraint.cs" />
+    <Compile Include="Constraints\ItemsConstraintExpression.cs" />
     <Compile Include="Constraints\Interval.cs" />
     <Compile Include="Constraints\Operators\AllOperator.cs" />
     <Compile Include="Constraints\Operators\NoneOperator.cs" />
--- a/src/NUnitFramework/framework/nunit.framework-4.0.csproj
+++ b/src/NUnitFramework/framework/nunit.framework-4.0.csproj
@@ -188,6 +188,7 @@
     <Compile Include="Constraints\InstanceOfTypeConstraint.cs" />
     <Compile Include="Constraints\Interval.cs" />
     <Compile Include="Constraints\IResolveConstraint.cs" />
+    <Compile Include="Constraints\ItemsConstraintExpression.cs" />
     <Compile Include="Constraints\LessThanConstraint.cs" />
     <Compile Include="Constraints\LessThanOrEqualConstraint.cs" />
     <Compile Include="Constraints\MessageWriter.cs" />
--- a/src/NUnitFramework/framework/nunit.framework-4.5.csproj
+++ b/src/NUnitFramework/framework/nunit.framework-4.5.csproj
@@ -193,6 +193,7 @@
     <Compile Include="Constraints\InstanceOfTypeConstraint.cs" />
     <Compile Include="Constraints\Interval.cs" />
     <Compile Include="Constraints\IResolveConstraint.cs" />
+    <Compile Include="Constraints\ItemsConstraintExpression.cs" />
     <Compile Include="Constraints\LessThanConstraint.cs" />
     <Compile Include="Constraints\LessThanOrEqualConstraint.cs" />
     <Compile Include="Constraints\MessageWriter.cs" />
--- a/src/NUnitFramework/framework/nunit.framework-netstandard.csproj
+++ b/src/NUnitFramework/framework/nunit.framework-netstandard.csproj
@@ -196,6 +196,7 @@
     <Compile Include="Constraints\InstanceOfTypeConstraint.cs" />
     <Compile Include="Constraints\Interval.cs" />
     <Compile Include="Constraints\IResolveConstraint.cs" />
+    <Compile Include="Constraints\ItemsConstraintExpression.cs" />
     <Compile Include="Constraints\LessThanConstraint.cs" />
     <Compile Include="Constraints\LessThanOrEqualConstraint.cs" />
     <Compile Include="Constraints\MessageWriter.cs" />
--- a/src/NUnitFramework/framework/nunit.framework-portable.csproj
+++ b/src/NUnitFramework/framework/nunit.framework-portable.csproj
@@ -192,6 +192,7 @@
     <Compile Include="Constraints\IConstraint.cs" />
     <Compile Include="Constraints\InstanceOfTypeConstraint.cs" />
     <Compile Include="Constraints\IResolveConstraint.cs" />
+    <Compile Include="Constraints\ItemsConstraintExpression.cs" />
     <Compile Include="Constraints\LessThanConstraint.cs" />
     <Compile Include="Constraints\LessThanOrEqualConstraint.cs" />
     <Compile Include="Constraints\MessageWriter.cs" />
--- a/src/NUnitFramework/tests/Constraints/ExactCountConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/ExactCountConstraintTests.cs
@@ -82,5 +82,34 @@ namespace NUnit.Framework.Constraints
             var ex = Assert.Throws<AssertionException>(() => Assert.That(names, new ExactCountConstraint(2, Is.EqualTo("Fred"))));
             Assert.That(ex.Message, Is.EqualTo(expectedMessage));
         }
+
+        [Test]
+        public void ExactlyFourItemsNoPredicate()
+        {
+            Assert.That(names, new ExactCountConstraint(4));
+            Assert.That(names, Has.Exactly(4));
+        }
+
+        [Test]
+        public void ExactlyFourItemsNoopNoPredicate()
+        {
+            Assert.That(names, Has.Exactly(4).Items);
+        }
+
+        [Test]
+        public void ExactlyOneItemNoPredicateFails()
+        {
+            var expectedMessage =
+                TextMessageWriter.Pfx_Expected + "exactly one item" + Environment.NewLine +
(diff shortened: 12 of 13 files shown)


src/NUnitFramework/framework/Constraints/Operators/ExactCountOperator.cs
```diff
@@ -39,17 +39,24 @@ public class ExactCountOperator : CollectionOperator
         /// <param name="expectedCount">The expected count</param>
         public ExactCountOperator(int expectedCount)
         {
+            // Collection Operators stack on everything
```
jnm2: This is copied from the former base constructor in CollectionOperator. I don't particularly like this.

src/NUnitFramework/framework/Constraints/PrefixConstraint.cs
```diff
@@ -54,15 +54,17 @@ protected PrefixConstraint(IResolveConstraint baseConstraint)
         /// The Description of what this constraint tests, for
         /// use in messages and in the ConstraintResult.
         /// </summary>
-        public override string Description
+        public override string Description => FormatDescription(DescriptionPrefix, BaseConstraint);
+
+        /// <summary>
+        /// Formats a prefix constraint's description.
+        /// </summary>
+        internal static string FormatDescription(string descriptionPrefix, IConstraint baseConstraint)
```
jnm2: This is another consequence of changing the base class.

jnm2:
Looks like I messed up the Compact Framework build.

src/NUnitFramework/framework/Constraints/Operators/ExactCountOperator.cs
```diff
@@ -39,17 +39,24 @@ public class ExactCountOperator : CollectionOperator
...
+            // and allow all other ops to stack on them
+            this.left_precedence = 1;
+            this.right_precedence = 10;
+
```
jnm2: This is copied from the former base constructor in CollectionOperator. I don't particularly like this.

src/NUnitFramework/framework/Constraints/PrefixConstraint.cs
```diff
@@ -54,15 +54,17 @@ protected PrefixConstraint(IResolveConstraint baseConstraint)
...
-            {
-                return string.Format(
-                    BaseConstraint is EqualConstraint ? "{0} equal to {1}" : "{0} {1}", 
-                    DescriptionPrefix, 
-                    BaseConstraint.Description);
-            }
+            return string.Format(
+                baseConstraint is EqualConstraint ? "{0} equal to {1}" : "{0} {1}",
+                descriptionPrefix,
+                baseConstraint.Description);
```
jnm2: This is another consequence of changing the base class.

jnm2:
Looks like I messed up the Compact Framework build by using string templates.

jnm2:
I see 9 commits crept up on me between forking and actually writing code. I'll rebase on master and squash the Compact Framework commit if you like.

CharliePoole:
I have a meeting tonight so I will have to get back to you in the morning.

jnm2:
Awesome, I'm not in a hurry if you're not.

CharliePoole:
As a separate comment (before I get into line-by-line) I think we should discuss what syntax we want to support. I'm particularly thinking about `Assert.That(collection, Has.Exactly(3));` which leaves me asking "Three what?" :-)

@rprouse What do you think of that syntax?

CharliePoole:
Regarding compatibility... we do want to preserve it in terms of the user syntax. It just isn't important internally... anyway, I'll see what you meant by that in the code. :-)

CharliePoole:
Re contributing.md... it seems we are starting to ask it to do a lot more for us. If we add so much too it, we end up duplicating documentation... should we link instead?

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
```diff
@@ -31,9 +31,19 @@ namespace NUnit.Framework.Constraints
     /// item in a collection, succeeding only if a specified
     /// number of items succeed.
     /// </summary>
-    public class ExactCountConstraint : PrefixConstraint
+    public class ExactCountConstraint : Constraint
```
CharliePoole: Maybe we really want two different constraints, analogous to PropertyConstraint and PropertyExistsConstraint. Of course, you have it working already, but it seems a little awkward to have one constraint that does two things.

In the case of `Has.Property` the prop operator can end up creating either a of the two constraints, depending on whether or not it comes at the end of the expression. Perhaps Items could do the same thing.

src/NUnitFramework/framework/Constraints/ItemsConstraintExpression.cs
```diff
@@ -0,0 +1,52 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// An extension of ResolvableConstraintExpression that adds a no-op Items property for readability.
+    /// </summary>
+    public sealed class ItemsConstraintExpression : ResolvableConstraintExpression
```
CharliePoole: It's an odd name. That is, Exactly creates an Items constraint, which has an Items property that produces a Resolvable constraint expression. I can't think of a good name, however. :-)

src/NUnitFramework/framework/Constraints/ItemsConstraintExpression.cs
```diff
@@ -0,0 +1,52 @@
...
+        /// passing in a pre-populated ConstraintBuilder.
+        /// </summary>
+        /// <param name="builder"></param>
+        public ItemsConstraintExpression(ConstraintBuilder builder)
+            : base(builder) { }
+
+        /// <summary>
+        /// No-op property for readability.
+        /// </summary>
+        public ResolvableConstraintExpression Items
```
CharliePoole: I presume you can't compile `Has.Exactly(3).Items.Items.Items...`?

src/NUnitFramework/framework/Constraints/Operators/ExactCountOperator.cs
CharliePoole: Are you saying you don't like the code or the copying? Any reason to change precedence of Exactly?

jnm2:
> I'm particularly thinking about `Assert.That(collection, Has.Exactly(3));` which leaves me asking "Three what?" :-)

It's consistent with `Assert.That(collection, Has.Exactly(3).EqualTo(x));`- neither more nor less confusing. It's left up to the user who can easily type `.Items` if they dislike the briefer version.

> Regarding compatibility, we do want to preserve it in terms of the user syntax. It just isn't important internally...

Yes, binary compatibility. User syntax is unaffected.

> Re contributing.md... it seems we are starting to ask it to do a lot more for us. If we add so much too it, we end up duplicating documentation... should we link instead?

Linking is a great idea. I'm not a fan of duplication.

src/NUnitFramework/framework/Constraints/ItemsConstraintExpression.cs
jnm2: That's correct. Intentionally, `ItemsConstraintExpression.Items` returns a `ResolvableConstraintExpression` rather than an `ItemsConstraintExpression`.

src/NUnitFramework/framework/Constraints/ItemsConstraintExpression.cs
jnm2: At first I called it `ExactCountContstraintExpression`, but then I realized it was more generic than that and could easily be picked up by some other expression.

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
jnm2: There is a sense in which it already has been doing two things: apply a predicate constraint, and count the items. I modified it so that it applies a predicate constraint if one is given, and counts the items.

The benefit of a single constraint is it keeps all enumerable count-related logic together, rather than having behavior coupling across two classes which don't reference one another, and it's less confusing if you ever construct the constraints manually. The benefit of separating into two constraints isn't clear to me yet.

src/NUnitFramework/framework/Constraints/Operators/ExactCountOperator.cs
jnm2: I don't like the copying. Maybe the code is simple enough that' it's obvious and the copying isn't harmful, but I'm not familiar enough with precedence and the magic numbers there.

I'm actually adding this code in order *not* to change the precedence of Exactly, since `ExactCountOperator` used to inherit from `CollectionOperator` and doesn't anymore. This way the precedence of Exactly does not change pre- to post-merge.

Or rather than from `CollectionOperator`, are you asking if there's any reason to change it from the defaults of `SelfResolvingOperator`? I am not familiar enough with precedence to answer that. I assumed it should stay the same as it was when it was a `CollectionOperator` but I was hoping for your guidance on precedence.

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
CharliePoole: Well, originally the property constraint was the same way: it either did one thing or two things, depending on whether a constraint followed it. It was refactored into two constraints. One of them does one thing (checks that the property exists) while the other does two (checks that it exists and applies a constraint to it).

The disadvantage of the approach is that the existence check has to be in two different places. The advantage is a somewhat cleaner code line in the constraint. The test for whether there is a constraint would have to be done in at least two places in the code of a single class: once for applying the constraint and again for the message.

The big advantage though is a potential one in the design of operators. Reading your code, I got to thinking that we could generalize operators like prop and exactly. They would produce one of two constraints depending on context. A derived operator would only have to specify the Types of the two constraints. This is just an idea fluttering in my head, however. I haven't worked it out in code. But it opens the door to user-created operators as well as constraints, so that's interesting.

src/NUnitFramework/framework/Constraints/Operators/ExactCountOperator.cs
CharliePoole: I'm uncertain... so maybe I want a test. :-)

ExactCountOperator was IIRC the last CollectionOperator added... long after the others. So it probably didn't get exhaustive syntactic tests. We should see what happens when it's followed by things like `And` and `Or`. That will give us a better idea of the priority to use. Take a look at the tests we have (I hope they are better) for some of the original CollectionOperators, like Some, No and All.

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
jnm2: That is interesting. I wonder if we could manage it so that the count or existence check would be in one constraint class and the additional constraint modification would be in another class, and they could be composed so that the class that applies an additional constraint does not duplicate any of the count or exists logic.

src/NUnitFramework/framework/Constraints/ItemsConstraintExpression.cs
jnm2: Think of it this way: It's a constraint on an object that has items. It is solely concerned with items.

CharliePoole:
@jnm2 Just so you know... I'm waiting for another committer to join in. We require two committers to merge. @nunit/core-team ?

jnm2:
@CharliePoole No problem. I thought you were waiting for me to write some tests for precedence. You don't have any further concerns at the moment?

Before you merge, can I rebase on top of your master and squash my Compact Framework fix? I had not realized the csproj was out of sync with the rest.

CharliePoole:
Yes, I'd like to see those tests. The second reviewer may ask for other stuff. :-)

Rebasing and squashing those changes is a good idea. Because it messes up the online diff on GitHub, it should wait till after all the review.

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
```diff
@@ -31,9 +31,19 @@ namespace NUnit.Framework.Constraints
...
     {
-        private int expectedCount;
+        private readonly int expectedCount;
+        private readonly IConstraint itemConstraint;
```
rprouse: Our new coding standard is for member variables to start with an underscore. We don't require that submitters update every file they touch, but if you are adding new members or making large changes to a file, refactoring to `_expectedCount` and adding `_itemContraint` would be nice.

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
rprouse: The dual logic also bothers me. We have the if/else in `ApplyTo` and the final expression in the `Description`. I like @CharliePoole's idea of producing one of the two constraints based on context, but like him, I am not in the code at the moment :smile:

src/NUnitFramework/framework/Constraints/ItemsConstraintExpression.cs
rprouse: The name `ItemsConstraintExpression` doesn't read well, but I think it is descriptive enough of it's purpose and can't think of anything better :smile:

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
jnm2: Will do. On a side note, is that something you can put in the dotSettings files I see in the repo so that ReSharper warns people like me about the new convention out of the box?

rprouse:
No problems with the code and comments inline. I would like to see more unit tests around the precedence to ensure that this isn't introducing any regressions and also to make sure we had it right in the first place.

As I stated inline, I am also concerned with the `ExactCountConstraint` having two behaviours. This approach bothers me and I would prefer to see another, cleaner solution.

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
rprouse: If that is a Resharper option, feel free to add it as a part of this PR or as a separate PR if you prefer. I stopped using Resharper so stopped keeping the dotsettings up to date.

CharliePoole: If it's made a resharper setting, does that mean that new contributors who
use resharper will see tons of warnings? Do we want that to happen?

2016-09-02 11:49 GMT-07:00 Rob Prouse <notifications@github.com>:

rprouse: Good point since we haven't converted a lot of code yet, although I think they are just visual (not compile time) yellow lines in the code and I expect there are already many. @jnm2 can you confirm and maybe post a screenshot to refresh my memory?

CharliePoole:
@rprouse What are your thoughts on 

```C#
Assert.That(collection, Has.Exactly(n));
```

Should we support that syntax or should it be (as now) illegal?

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
CharliePoole: Of course, I always say you can't remove duplication until you produce it. :-)

Maybe we should merge it this way and look at simplifying later, particularly if the simplification spans more constraints and operators than the ones in play here.

CharliePoole:
I'm ready to merge with some further tests. At a minimum, I think we need to see Exactly used...

* before `.And`
* after `.And`
* before `.Or`
* after `.Or`

In each case, it should work with a constraint as well as without a constraint.

Take a look at how we do syntax tests without actually executing the constraints. It's sufficient if the syntactic construct produces the proper constraint expression representation string.

rprouse:
> Should we support that syntax or should it be (as now) illegal?

I think we should continue to support it. I am wary about potentially breaking people's tests with an update.

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
jnm2: The dotSettings is already showing naming convention inspections as warnings. Somehow my personal settings were overriding the project settings. I'm still not clear what  I messed up but I removed my preference for camelCase and added my preference back and it's working. As far as the experience goes, it's perfect. You see blue squiggles in the code but no warnings in the error list.

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
jnm2: I also want to point out that from my perspective it was doing two things that should have been separate before I got here, so perhaps the fix for that should also be isolated from this PR. The change I made was to allow itemConstraint to be null rather than create a dummy `true` constraint. The only reason there are two loops is to save a null check per iteration but I could change it back to a single loop with the logical equivalent of `itemConstraint?.ApplyTo(item).IsSuccess != false` if it makes a big difference right now.

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
jnm2: Is the underscore naming convention switch something I can squash before merging to keep the noise level down?

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
jnm2: To put it another way, the IEnumerable loop and the count++ is one concern and the itemConstraint filter is another concern. If we have a way to solve it so that those two concerns are separated into two classes that can compose, perfect! I hope that's the eventual solution. However, a short term solution of separating into two classes is more harmful than anything if both classes still retain the IEnumerable loop counting concern. In that case not only does the class with the filter still deal with both concerns, but now you have duplicated code which results in invisible behavioral coupling between two classes instead of coupling between the branches of the itemConstraint null check.

CharliePoole:
@rprouse I think you misunderstood.

We do not now support that syntax. I'm asking if it should be added because it seems a bit iffy to me and once we add it we have to keep supporting it pretty much forever.

Since the comments are spread all over, I'll summarize again... the PR supports...

```C#
Assert.That(collection, Has.Exactly(3).EqualTo("Charlie")); // 1
Assert.That(collection, Has.Exactly(3).Items.EqualTo("Charlie")); // 2
Assert.That(collection, Has.Exactly(3)); // 3
Assert.That(collection, Has.Exactly(3).Items); // 4
```
Of those, only item 1 is currently supported. I like Items 2 and 4. I'm questioning item 3.

CharliePoole:
So... summarizing where I am on this...

1. We should resolve the supported syntax. If @rprouse likes it I can be convinced. :-)
2. I'm happy to postpone major internal reorganization to another time - but if we decide to change the syntax, we will have to revise anyway so we could look at it now.
3. I'd like to see those tests

With those things resolves I think we can merge.

jnm2:
I like the symmetry with `Has.Exactly(3).EqualTo`. But I could go either way.

I'll do the naming convention change and tests tomorrow. Spent this evening in an Ultimate playoff game.

src/NUnitFramework/framework/Constraints/Operators/ExactCountOperator.cs
```diff
@@ -29,7 +29,7 @@ namespace NUnit.Framework.Constraints
     /// Represents a constraint that succeeds if the specified 
     /// count of members of a collection match a base constraint.
     /// </summary>
-    public class ExactCountOperator : CollectionOperator
+    public class ExactCountOperator : SelfResolvingOperator
     {
         private int expectedCount;
```
jnm2: @rprouse I was going to refactor this as well since I made significant changes to the file, but then I realized it would be inconsistent with the base members `left_precedence` and `right_precedence` which are referenced in the added code. Shall I just leave this file as-is?

jnm2:
@CharliePoole Ran into another issue. `CollectionOperator` is used in two type tests inside `BinaryOperator` which overrides [LeftPrecedence](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Constraints/Operators/BinaryOperator.cs#L51) and [RightPrecedence](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Constraints/Operators/BinaryOperator.cs#L564). `ExactCountOperator` is no longer a `CollectionOperator` so the type test should check `is CollectionOperator || is ExactCountOperator`, and now the type test is beginning to show itself as a code smell.

Here are the concerns I have:
 1. Why is the BinaryOperator precedence overridden and dynamically calculated depending on the operators it's used on? I thought the point of numbers for precedence rules was in order to avoid having to do conditional logic like that.
 2. Why does the `LeftPrecedence` override check the type of `RightContext` rather than `LeftContext`?

I'm afraid this is deeper than I was prepared for and will take a lot more time unless someone involved in the original rationale can share their insight.


The symptom I was chasing was that
`Has.Exactly(3).Items.LessThan(0).And.Exactly(3).Items.GreaterThan(10)` gives
`<exactcount <and <lessthan 0> <exactcount <greaterthan 10>>>>`.

Contrary to that, `Has.Some.LessThan(0).And.Some.GreaterThan(10)` gives the more desirable
`<and <some <lessthan 0>> <some <greaterthan 10>>>`.

rprouse:
> @rprouse I think you misunderstood.

@CharliePoole I did misunderstand. My fault for parachuting in and not keeping up on the full discussion. Based on that, I am also against option 3, I don't think it reads well and it is likely confusing without looking at the docs.

jnm2:
@rprouse @CharliePoole Sounds good. I'll need to introduce a new type in order to prevent `Has.Exactly(3)` from working.

I'll still have the precedence issue. It looks like the scope of this PR expanded so I'll try to figure out the best solution for precedence that does not involve a magic type check.

rprouse:
@jnm2 thanks for the update. Let us know if you would like some help.

jnm2:
I would appreciate receiving some background if you happen have knowledge of the rationale behind the precedence system, otherwise I'm very new to the scene.

CharliePoole:
I'll write something on the Wiki and add a pointer. That way it will be available to others in the future.

jnm2:
Okay... got this far:

| Operator           | Left | Right |
|--------------------|------|-------|
| CollectionOperator | 1    | 10    |
| AndOperator        | 2    | 2     |
| AttributeOperator  | 1    | 1     |
| ThrowsOperator     | 1    | 100   |
| NotOperator        | 1    | 1     |
| OrOperator         | 3    | 3     |
| PropOperator       | 1    | 1     |
| WithOperator       | 1    | 4     |

jnm2:
@CharliePoole @rprouse Because the precedence of binary operators is actually intended to change, precedence will not be able to be constant per operator type:

```
Some.X.And.Y		<some <and x y>>
Some.X.And.Some.Y	<and <some x> <some y>>
```

The `Some.Y` actually changes the relationship between `Some.X` and `And`.
So now the issue is decoupling `And` and `Or`'s collection-sensitivity from the `CollectionOperator` class since `And` needs to be treated differently around `ExactCountItemsOperator` (which can no longer be `CollectionOperator`) as well as `CollectionOperator`:

```
Exactly(3).Items.X.And.Y					<exactcount <and x y>>
Exactly(3).Items.X.And.Exactly(3).Items.Y	<and <exactcount x> <exactcount y>>
```

I don't like the type test but one will be necessary since `RightContext` is `object`. The simplest way forward is introducing `ICollectionOperator` and testing for that type.
Another possibility is having a more fundamental property on the base class indicating whether the operator applies to multiple items (or multiple properties or attributes, really): `ConstraintOperator.IsMultiple`.

Also, I don't like that `CollectionOperator` has right precedence of `10` which is a magic number which also shows up at the `BinaryOperator` type test. Better for `BinaryOperator` to return `LeftContext.LeftPrecedence + 1` or `LeftContext.LeftPrecedence + this.left_precedence`. Since they're coupled anyway, lets make it a refactor-safe coupling.

jnm2:
Also, I'm thinking this may be less a collection-version of `And` and more a general principle. The same logic applied grammatically to `And.Some` is just as legitimately applied to `And.Property`:

```
Has.Property("A").GreaterThan(1).And.LessThan(10)
    <property <and <greaterthan 1> <lessthan 10>>>

Has.Property("A").GreaterThan(1).And.Property("B").LessThan(10)
    <and <property A <greaterthan  1>> <property B <lessthan 10>>>
```

It seems that any justification you have for `And` changing precedence when followed by `Some` would apply equally to `Property`. If you want with that idea (and it would be in v4 as a breaking change), it's less a collection thing and more a grammatical rule of *we've repeated the word `Some` or the word `Property` twice, so `And` applies to the two repeated words.*

Another thing to consider is `Has.Property("Count").GreaterThan(2).And.Some.LessThan(10)`.

What are your thoughts?

CharliePoole:
I'll have to wait till this release cycle is done to dig into this.

rprouse:
@CharliePoole, you and I need to look at this again.

jnm2:
@CharliePoole I'm pretty sure that without changes, `And` followed by `ExactCount` will change behavior which could be breaking because `ExactCount` no longer inherits from `CollectionOperator`.
The very least change we could make would breaking is `ICollectionOperator` for `And` to check for so that `ExactCount` is able to inherit it.

CharliePoole:
@jnm2 Can you create a test that demonstrates the problem? That is, one that passes under the unmodified master code but fails with your changes? You could even check it in, although it will break your build.

jnm2:
@CharliePoole I have the test written, are you fine waiting till this evening for me to do that?

CharliePoole:
Sure thing. Actually, I thought you were ready to merge this and didn't want to hold you up. But if you think there are still problems, we should take the time to get it right.

jnm2:
In a nutshell:
 * precedence is currently fragile because `BinaryOperator` is coupled to `CollectionOperator` via type check
 * with the current `SelfResolvingOperator` and `CollectionOperator`, it's not possible for an operator (`Exactly(x).Items`) to both be a collection operator (to preserve precedence rules) and to be self-resolving.

Solutions:
 1. Refactor usage of `CollectionOperator` as a marker class into a marker interface `ICollectionOperator` (smallest possible change)
 2. Revisit the way precedence works to see if we can get rid of the `BinaryOperator`/`CollectionOperator` coupling (would feel cleaner, especially with info I laid out)

CharliePoole:
It seems to me that issues of fragility and possible refactoring can be dealt with either here or in a separate PR. But if there is a known failing test we should deal with that right here, so let's get that test into the mix.

It looks to me as if the necessary refactoring might involve the entire hierarchy of operators, in which case I would lean toward doing it separately, but soon.

jnm2:
@CharliePoole pushed the commit to this branch.

jnm2:
@CharliePoole I rebased on master and pushed the precedence test commit.

jnm2:
@CharliePoole I rebased on master and pushed the precedence test commit.

What would you like to do, internal `ICollectionOperator` for this PR to make the tests pass and then a followup PR to deal with the reset of the precedence story?

CharliePoole:
@jnm2 GitHub can't merge your code. Did you forget to pull master before doing the rebase?

jnm2:
@CharliePoole yes, I did. Then found some csproj merge conflicts. 🙄 All good now.

jnm2:
Two tests that I added are failing until we deal with `CollectionOperator` vs `SelfResolvingOperator`.
And I still want to squash before merge, especially now that the CF project file is no more lol.

jnm2:
Actually hang on and I'll squash them now.

CharliePoole:
Those tests are pointing me to the right place - I see what you are talking about now.

rprouse:
@CharliePoole, merging your PR which was based on this one automatically merged this one too. No harm, just interesting how GitHub works...

CharliePoole:
@rprouse That is interesting.

PR #1783, opened 2016-09-02
Add 

pflugs30:
Related to Issue #144.  Adds several overloads to `Assert.That` and `Assume.That` to support a test inserting a `Func<string>` or `Func<ConstraintResult, string>` to build a message to pass to the particular exception but only if the Assertion fails.  That is, it is lazily executed.

PR #1784, opened 2016-09-03
Add assertion overloads that delay evaluation of the message until needed

### LINKED ISSUE #144: Pass a Func<string> to lazily evaluate an exception message
I find several cases in my tests where I only want to build a message when the Assertion fails.  Consider the following snippet:

// Assert
var comparer = new MyEqualityComparer();
Assert.That(actual, Is.EqualTo(expected).Using(comparer));

I have several cases where I use an IEqualityComparer<T> to simplify multiple Assertion statements into a reusable EqualityComparer.  It's great, until the the verification fails, and I have to manually step through the test comparing properties manually.

To help automate this task, I built a reusable helper class to not only compare two classes but enumerate the differences between their properties.  Then, the differences are rendered to a string, and I can use this string as the failure message in the Assertion.

Ideally, this message would not need to be created unless the verification fails.  However, none of the Assert methods allow for passing a function as a parameter which would lazily evaluate the failure message.  Here's the syntax I'd like to use:

// Assert
var comparer = new MyEqualityComparer();
Func<string> getFailureMessage = () => comparer.WriteDifferencesToString(expected, actual);
Assert.That(actual, Is.EqualTo(expected).Using(comparer), getFailureMessage);

Using this syntax, client test methods could lazily evaluate the failure messages using any valid Function.  I think this would really help users clean up their tests and extend the capabilities of the NUnit framework.

I'm open to discussion on this idea as well as direction on how to help make it a possibility.  Thank you for your consideration.

CharliePoole:
This PR builds on PR #1783 contributed by @pflugs30, which got caught in the big split of framework, engine and console. See some discussion on the original PR as well as here.

Fixes #144

### DIFF
--- a/src/NUnitFramework/framework/Assert.That.cs
+++ b/src/NUnitFramework/framework/Assert.That.cs
@@ -35,31 +35,7 @@ namespace NUnit.Framework
     {
         #region Assert.That
 
-#region Boolean
-
-#if !NET_2_0
-        /// <summary>
-        /// Asserts that a condition is true. If the condition is false the method throws
-        /// an <see cref="AssertionException"/>.
-        /// </summary> 
-        /// <param name="condition">The evaluated condition</param>
-        /// <param name="getExceptionMessage">A function to build the message included with the Exception</param>
-        public static void That(bool condition, Func<ConstraintResult, string> getExceptionMessage)
-        {
-            Assert.That(condition, Is.True, getExceptionMessage);
-        }
-
-        /// <summary>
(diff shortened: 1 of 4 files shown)


CharliePoole:
@pflugs30 @rprouse Please review

pflugs30:
Looks good to me. Reviewed on phone and not compiled locally, but assuming that's already been done. Thanks!

rprouse:
Looks good to me too.

pflugs30:
Thanks, @rprouse!

PR #1785, opened 2016-09-06
Don't enforce timeout when debugger is attached

### LINKED ISSUE #1479: Don't enforce [Timeout] when debugger is attached
When debugging a unit test, the [Timeout] attribute is a great pain as tests fail just because the programmer spends some time in a breakpoint. It probably make sense to disable [Timeout] (and probably also [Maxtime]) when a debugger is attached.

chris-smith-zocdoc:
Don't enforce timeout when debugger is attached.

We perform this check after the initial timeout has passed to give the user additional time to attach a debugger after the test has started execution. 

Fixes #1479

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
@@ -317,6 +317,16 @@ namespace NUnit.Framework.Internal.Execution
 
             if (!thread.Join(timeout))
             {
+                // Don't enforce timeout when debugger is attached.
+                // We perform this check after the initial timeout has passed to 
+                // give the user additional time to attach a debugger 
+                // after the test has started execution
+                if (Debugger.IsAttached)
+                {
+                    thread.Join();
+                    return;
+                }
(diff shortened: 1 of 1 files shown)


src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -317,6 +317,16 @@ private void RunThread(int timeout)
 
             if (!thread.Join(timeout))
             {
+                // Don't enforce timeout when debugger is attached.
+                // We perform this check after the initial timeout has passed to 
+                // give the user additional time to attach a debugger 
+                // after the test has started execution
+                if (Debugger.IsAttached)
```
CharliePoole: Interesting. I wouldn't have thought to do it quite this way but I guess it allows for a very last minute attaching of the process.

CharliePoole:
Ironically, the Apple build on travis timed out. I restarted it, but it looks OK to me for merging.

rprouse:
Looks good, thanks.

Issue #54, opened 2013-11-07
Store message and stack trace in `TestContext` (for use in `[TearDown]`)


killerswan:
Right now, in, e.g., NUnit 2.5.10, inside of a `[TearDown]` method I can use `TestContext.CurrentContext.Result.Status` to see whether the test has failed/passed/skipped.

It would be really cool if I could also get all the other information which is logged in the console output, or in result XML, too... particularly the message and stack traces!

Issue #1787, opened 2016-09-07
ANSI characters in Console Output


null511:
I'm trying to push ANSI color characters through the test output and into the console for highlighting key-items in my test output. I'm able to do this successfully outside of NUnit (writing directly to `Console.Out`), but it looks like the colors are being overwritten by the appender. I was wondering if you guys had any insight as to why this might not be working? I've been digging through the code, but haven't identified anything yet.
- Worth noting, the ansi codes I am pushing are being intercepted (not printed to console).

Before you try to stop me:
- I know ANSI has limited terminal support, but we are running in bash and displaying final results in custom reporting tool that is capable of parsing these characters.
- We are using NUnit for integration tests (using Selenium) which have lengthy output describing each step taken, so separating output into smaller unit-tests is not an option.

null511:
**more info:** It appears that the characters are being intercepted, and converted back to a unicode representation. For example, if I input `\x1b[1;34mHello World`, the console will print `\u001b[1;34mHello World`, instead of a *blue* `Hello World`.

CharliePoole:
What version of NUnit are you using?

null511:
3.4.1

null511:
Sorry for the jumbled addition of information.

- Writing to `TestContext.Out` will print the escaped unicode characters. `\x1b[1;34m` -> `\u001b[1;34m`.
- Writing to `TestContext.Error` & `TestContext.Progress` does **not** shown the escaped characters, but it appears the colors are overridden by the Appended.
- Writing to `Console` directly **without NUnit** displays the message colorized properly.

CharliePoole:
At first I thought this might be a duplicate of #1575, which is fixed in master. However, that has to do with escaping the user message provided as a part of an Assert call.

If you use `Console.Write` **in an NUnit test** I would expect it to perform the same as `TestContext.Out`. `And Console.Erorr.Write` should work like `TestContext.Error`.

One minor thing to be aware of is that nunit3-console colors the error output itself, by setting the console foreground color. The result will probably depend on who gets there last.

We'll have to look a bit deeper to see where your escaping is removed. For now we will keep this issue in the nunit repository, because I suspect the framework is the culprit.

CharliePoole:
This is clearly happening. I marked 'confirm' in the sense of confirming it happens in the framework.

null511:
[...]

PR #1789, opened 2016-09-10
escape backslashes

null511:
References nunit-console [PR #54](https://github.com/nunit/nunit-console/pull/54).

This branch adds backslash escaping for proper unicode unescaping.

### DIFF
--- a/src/NUnitFramework/framework/Interfaces/TNode.cs
+++ b/src/NUnitFramework/framework/Interfaces/TNode.cs
@@ -321,11 +321,12 @@ namespace NUnit.Framework.Interfaces
         private static readonly Regex InvalidXmlCharactersRegex = new Regex("[^\u0009\u000a\u000d\u0020-\ufffd]|([\ud800-\udbff](?![\udc00-\udfff]))|((?<![\ud800-\udbff])[\udc00-\udfff])");
         private static string EscapeInvalidXmlCharacters(string str)
         {
-            if (str == null) return null;
+            if (string.IsNullOrEmpty(str)) return str;
 
             // Based on the XML spec http://www.w3.org/TR/xml/#charsets
             // For detailed explanation of the regex see http://mnaoumov.wordpress.com/2014/06/15/escaping-invalid-xml-unicode-characters/
 
+            str = str.Replace("\\", "\\\\");
             return InvalidXmlCharactersRegex.Replace(str, match => CharToUnicodeSequence(match.Value[0]));
         }
 


rprouse:
This looks okay to me, but I would like to see unit tests for it and we need a second review.

src/NUnitFramework/framework/Interfaces/TNode.cs
```diff
@@ -321,11 +321,12 @@ private static NodeList ApplySelection(NodeList nodeList, string xpath)
         private static readonly Regex InvalidXmlCharactersRegex = new Regex("[^\u0009\u000a\u000d\u0020-\ufffd]|([\ud800-\udbff](?![\udc00-\udfff]))|((?<![\ud800-\udbff])[\udc00-\udfff])");
         private static string EscapeInvalidXmlCharacters(string str)
         {
-            if (str == null) return null;
+            if (string.IsNullOrEmpty(str)) return str;
 
             // Based on the XML spec http://www.w3.org/TR/xml/#charsets
             // For detailed explanation of the regex see http://mnaoumov.wordpress.com/2014/06/15/escaping-invalid-xml-unicode-characters/
 
+            str = str.Replace("\\", "\\\\");
```
CharliePoole: Should this be in the `EscapeInvalidXmlCharacters` method? AFAIK backslash is not an invalid XML character.

null511: It's not because backslashes are invalid, otherwise I would be adding more invalid characters...

The backslashes are being escaped so that the escaped unicode characters can be properly unescaped ([See last comment in PR #1787](https://github.com/nunit/nunit/issues/1787)). This allows users to enter sequences such as `\\x1b001` and still print `\x1b001` on the screen (without it being converted to the character).

CharliePoole:
So this depends on PR #1787 to function correctly? What happens if we independently merge it?

rprouse:
@null511 I am not sure where this PR stands now with #1787. Is it still needed? Can you update us, otherwise I am going to close.

null511:
I will update this branch today. The issue now is that to properly encode Unicode characters in test output, back slashes need to be double escaped. This would cause changes in the XML output though, which may be unexpected and undesirable for many users.

Console output would be unaffected, since part of the change is to decode the Unicode characters, and unescape back slashes before printing to console window.

This could be handled by adding another option, but I was trying to avoid that route.

null511:
The underlying issue here, is that while XML supports most unicode characters, it does not support the range of control characters which can be used to colorize console output. These characters have to be escaped with back slashes `π` > `\u03a0`. To properly decode this output, existing back-slashes would have to be double-escaped '\' > '\\'.

Since this is actually more specific to control characters than unicode characters, perhaps there could be an additional option `enableControlCharacters` which activates double escaping of back slashes. This option would have to be accessible from both the framework and the console.

rprouse:
@null511 thanks for the update. That would be a feature with a fairly limited audience, so I am reluctant to add it. Before you do too much more work, I think we should have the discussion on whether or not this is something we will accept.

I understand your use case is to colourize test output to make visually scanning it easier. Correct?

@nunit/core-team and @nunit/contributors what do you think of additional command line options or special XML escaping being added for this? Do you think that it is a broad enough use case to warrant the changes?

CharliePoole:
Hmmm... Here's something I only just thought about. If this is about XML encoding, then maybe this belongs in the code that creates XML, not in the general console output code. If you use NUnitLite, for example, there's no XML involved. It only comes into play for runners that use the engine.

CharliePoole:
OK... never mind... I see you actually are doing this in TNode, where the XML is created.

I'll give this a new review to refresh my memory. I'm still not completely understanding the issue with backslashes, however. Could you post an example of what happens with and without the double escaping?

null511:
The same concept is used in C#. All strings double-escape back slashes because that is the character used for escaping special characters. This extra back slash is used when unescaping the data, to differentiate a back slash from an escaped character.

Without double-escaping slashes, any slashes that were input will be improperly decoded.
```
in         xml        out
---------------------------
π      >  \u03a0   >  π
\π     >  \\u03a0  >  ?π
\0123  >  \0123    >  ?
```

When double-escaped, `\\` will be converted to `\`, and `\*` will be treated as an escaped character.
```
in         xml        out
---------------------------
π      >  \u03a0    >  π
\π     >  \\\u03a0  >  \π
\0123  >  \\0123    >  \0123
```

CharliePoole:
Just to be clear, the C# convention you describe applies to source code, not strings read by a C# program. When you read in a string, you get exactly what is there. No special processing of backslash happens.

Are you saying that your first table shows what NUnit is currently doing? I'm a bit surprised at the second entry. I would expect to see \π in that case.

Because the XML encoding and decoding are taking place in separate projects, we will have to set up special tests that both encode and decode the strings, using the same algorithms used in the framework and the console.

null511:
Quickly, I want to note that `π` is valid xml, and would not need to be escaped. I'm simply using that character to represent 'control characters', since they are non-printed characters.

> Just to be clear, the C# convention you describe applies to source code, not strings read by a C# program.

Correct. I was going to say "... in Visual Studio", but it's more specifically C# since VB does not support escaped characters.

> Are you saying that your first table shows what NUnit is currently doing? I'm a bit surprised at the second entry. I would expect to see \π in that case.

Partially. Currently no output is unescaped, so the output would be the same as the XML.
```
in         xml        out
---------------------------
\π     >  \\u03a0  >  \\u03a0
```

When the escaped string is decoded, it cannot differentiate between a back-slash and an escaped character, because they are the same character `\`. For example, if you wanted to print the text `\uniform\`, it would be improperly unescaped to `\unifo`+`rm`+`\`=`?rm?`.

(Original Issue #1787)[https://github.com/nunit/nunit/issues/1787#issuecomment-246152366]

rprouse:
@null511 thanks for the detailed explanations. reading the most recent comments, I would be okay with this change, but only if it includes unit tests. Do you have time to pick this up again? Sorry it has been out there so long.

rprouse:
I am closing since we have not heard back from you. If you come back to it, we will reopen.

PR #1798, opened 2016-09-20
[Do Not Merge]Remove wix files migrated to nunit/nunit-distribution

### LINKED ISSUE #1716: Move installer to new repository
With moving all the engine extensions to separate repositories, the msi installer will no longer have direct access to most of the assemblies it packages. As such, I suggest we go ahead with moving it to a separate repository, and retrieve the required assemblies via NuGet restore.

If we did this, a component release would be two stage:
1. Build the release-assembly itself, and package, and publish the individual NuGet package.
2. Run a script in the new repository, which retrieves the new package, and all other dependencies from NuGet, and creates a new msi installer.

In future, it might make sense to even wrap a script round this whole process, so a "ReleaseEngine" task fetches the relevant sources from GitHub, increments versions, builds, packages and publishes all individual and "group" NuGet packages and installers. But that's perhaps a way down the line!

This new repository would also seems like a sensible place for other combined packages e.g. NUnit.Runners to live - but I think that should be handled as a separate issue. 

In terms of how this fits in with the 'big split' - as far as I can see, it can be done completely independently. I'd expect to `filter-branch` out the msi files from a clone of nunit-master, and then transfer that back as a new repository. At some point the exisiting msi files should be removed from the new engine repository, but that can be done at any point, before or after the split. 

Thoughts? :-)

ChrisMaddock:
Fixes #1716 

Looks like the msi build was already removed from Cake, so as far as I can see, it's just the wix files themselves to be removed, and a reference in the appveyor.yml


ChrisMaddock:
Sorry about this one, the history rewrite seems to have confused my local repo! New PR at #1799 👍

PR #1810, opened 2016-09-29
Improve error message for invalid TestCase construction

### LINKED ISSUE #1805: Error message "arguments provided for method not taking any" seems incomplete / doesn't make much sense
I got an error message while using NUnit with `TestCaseSource` attribute.
The message is:

> Arguments provided for method not taking any

And is found in the line below:

https://github.com/nunit/nunit/blob/5f6b47144c2d7b69caa52b3ae24dbb740444f347/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs#L217

I couldn't figure out what this means, since the sentence seems to be missing some part, it is grammatically weird. Taking any _what_?

My code, for context:

    [TestFixture]
    public class TransformaçãoLinearTestes
    {
        private TransformaçãoLinear target;

        [Test, TestCaseSource("Valores")]
        public void TransformNãoModificaValorPorDefault(double entrada)
        {
            target = new TransformaçãoLinear();
            Assert.AreEqual(entrada, target.Transform(entrada));
        }

        public static double[] Valores = new double[]
        {
            0, 1, -1, 0.5, -0.5
        };

    }

ChrisMaddock:
Fixes #1805 

Looking at how the error is displayed in console, including the method name as I suggested on the issue would be daft, so instead have just reworded the error to be more explicit.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
@@ -214,7 +214,7 @@ namespace NUnit.Framework.Internal.Builders
                 return MarkAsNotRunnable(testMethod, "Method has non-void return value, but no result is expected");
 
             if (argsProvided > 0 && maxArgsNeeded == 0)
-                return MarkAsNotRunnable(testMethod, "Arguments provided for method not taking any");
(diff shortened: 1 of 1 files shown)


CharliePoole:
Looks good to merge when it passes CI.

rprouse:
Looks good, thanks.

PR #1812, opened 2016-09-29
Issue 1792 test case source too many arguments provided

debugosaurus:
Refactored logic to determine TestCase parameters from TestCaseeSourceAttributes. Logic now should support transposing arrays (vertical) into method parameters (horizontal) but still respect IEnumerable / Array parameters. The refactored logic treats strings as a special case as while these are technically IEnumerable existing functionality treats them as if they were not (which makes sense).

Also added a number of additional IEnumerable tests these each cover the current issue along with known regression risks.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
@@ -139,6 +139,8 @@ namespace NUnit.Framework
         /// <returns></returns>
         private IEnumerable<ITestCaseData> GetTestCasesFor(IMethodInfo method)
         {
+            var testName = method.TypeInfo.FullName + "." + method.Name;
+
             List<ITestCaseData> data = new List<ITestCaseData>();
 
             try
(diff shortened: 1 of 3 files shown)


rprouse:
Thanks for the contribution. Before I review it, builds are failing because you can't use the `nameof` operator or any other C# 6 features because the Compact Framework version uses the same source files.

debugosaurus:
Sorry about that, hopefully this is all fixed now

debugosaurus:
Requires reworking to support CompactFramework

CharliePoole:
Your build timed out on Apple, which is common. I kicked it off again.

CharliePoole:
I didn't notice you had closed this. You'll usually get better results if you keep the PR open so we can see your problems, follow your changes and give advice as needed. I'm re-opening so I can see the problem better.

CharliePoole:
OK, I see that there are bunches of failures. I'll re-close this. Basically, your options are to make CF work or exclude CF from the change. For something like this, which is fairly fundamental, I would at least try to make it work.

I notice you actually made some code changes to code inside an #if NETCF. If you aren't actually using and testing with CF, that's probably a bad idea. However, you can still be messed up by the code that isn't specific to CF, especially if you are dealing with generics.

A third choice would be to put this on hold until we remove CF to a separate legacy project, which will happen fairly soon.

PR #1815, opened 2016-09-30
"&" IS NOT "&&"

kloun:
Fix eric
(from http://www.viva64.com/ru/b/0420/ )

### DIFF
--- a/src/NUnitFramework/framework/Constraints/SubPathConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/SubPathConstraint.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2008 Charlie Poole
 //
 // Permission is hereby granted, free of charge, to any person obtaining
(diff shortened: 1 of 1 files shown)


CharliePoole:
This looks good to me - it's an obvious error. Since there is no issue, I labeled this as if it were a bug report.

kloun:
please merge this patch as soon as possible

CharliePoole:
It will get merged as soon as a second committer looks at it and finds it OK.

ChrisMaddock:
Looks good to me - thanks @kloun!

I believe this is functionally equivalent, although more efficient, as the second condition won't necessarily be evaluated using a logical AND instead of bitwise. Good spot!

ChrisMaddock:
Added this to the milestone, and re-titled to something that will make more sense in a changelog.

CharliePoole:
@ChrisMaddock It's not an efficiency improvement it's to stop a NullReferenceException from being thrown in the case that actual is null.

ChrisMaddock:
Didn't think that through - did I! 😆 Thanks Charlie - retitled again.

Issue #1609, opened 2016-06-23
Upgrade Cake build to latest version


CharliePoole:
Currently seems to be 0.13

ChrisMaddock:
When this get's done, it might be nice to also make use of cake-build/cake#908. :-)

CharliePoole:
@ChrisMaddock I agree!

ChrisMaddock:
#1615 should probably be investigated before this. :-)

ChrisMaddock:
It's probably worth now waiting for Cake v15, which I believe is coming soon. It will have required fixes for  cake-build/cake#908, and also full error messages from the NUnit3 tool, which might make it a preferable alternative. 

This might even be better done with a bit of a refresh, post-split.

CharliePoole:
Seems fair. BTW, I'm using Cake 0.14 for the teamcity extension and having some issues with mono 3.2.8. So we'll need to see if that gets fixed.

ChrisMaddock:
Good to know.

Reopening this issue as I presume closing it was accidental?

CharliePoole:
@ChrisMaddock Some of us can't use Cake 16+ due to a bug. I've been using 15.2 in all the latest updates to other projects. NUnit-Console and our own four extensions have now been updated to use the latest bootstrap scripts from Cake as well as having tools/packages.config updated to 15.2. You may be able to simply copy .gitignore, build.sh and build.ps1 from those projects.

ChrisMaddock:
Thanks Charlie - I'd been using 16.1. Will go back a version!

ChrisMaddock:
@CharliePoole - looks like Cake 16.2 has just been released today. 

Was this the bug you were referring to: https://github.com/cake-build/cake/issues/1247? I'll go ahead using 16.2 if it's fixed, else I'll stick to 15.2.

CharliePoole:
No. It's an issue I discussed in chat. I should file an issue report on it.

In any case, I think it's best to get everything on the same level, then
continue to upgrade things in sync. I'm working on something to facilitate
that.

PR #1823, opened 2016-10-04
Remove engine nuspecs and old global.json

ChrisMaddock:
Starting on #1609 - just separating this out for easier review.

Removes all the nuspec's used by the engine and console which were moved. Also removes `global.json` - which I believe was for old projects which are no longer here? @rprouse?


rprouse:
The `global.json` file was for when we had a .NET Core test runner in the repo. It will get re-added soon with the .NET Core version of the framework, but the content will be different, so no harm in removing it.

Issue #1765, opened 2016-08-23
TestCaseSourceAttribute constructor for method with parameters


oznetmaster:
TestCaseSourceAttribute has a constructor that takes a string with the name of the method/field/property that provides it data, and a constructor that takes type of the source, a string with its name, and parameters if it is a method,

It would be convenient if a constructor that took just the string name of the method and parameters were available as well,

``TestCaseSourceAttribute (string sourceName, object[] methodParams)``

CharliePoole:
Makes sense to me. I made it low priority, but if it's higher for you, feel free to do a PR.

PR #1831, opened 2016-10-11
Added new constructor for #1765 with test case

JustinRChou:
Added new constructor for #1765 with test case.

### DIFF
--- a/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
@@ -186,7 +186,7 @@ namespace NUnit.Framework.Attributes
         }
 
         [Test, Category("Top"), TestCaseSource("HereIsTheDataWithParametersFor1765", new object[] { 8000, 8, 1000 })]
-        public void SourceInAnotherClassPassingSomeDataToConstructor1765(int n, int d, int q)
+        public void SourceCanBeStaticMethodPassingSomeDataToConstructor(int n, int d, int q)
         {
             Assert.AreEqual(q, n / d);
         }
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -185,6 +185,12 @@ public void SourceInAnotherClassPassingSomeDataToConstructor(int n, int d, int q
             Assert.AreEqual(q, n / d);
         }
 
+        [Test, Category("Top"), TestCaseSource("HereIsTheDataWithParametersFor1765", new object[] { 8000, 8, 1000 })]
+        public void SourceInAnotherClassPassingSomeDataToConstructor1765(int n, int d, int q)
```
CharliePoole: The whole point of the issue is to allow use of parameters with data in the same class, specifying only the method name. You've implemented that correctly, but the name of your test says you are testing data in another class. Also, use of the issue number in the test name won't help anyone reviewing the code after we merge it. If you just call it SourceInSameClassPassingSomeDataToConstructor, that tells someone reading the code next year what the test does.

CharliePoole:
That is, ready to merge with the comment fixed.

I find it really annoying this new review facility really annoying. It doesn't match the way we work, with two committers needed to merge and it won't let me edit a comment after I make it. I wonder if it can be turned off?

rprouse:
Looks good to me. Any @nunit/core-team member may merge after the change requested by @CharliePoole is finished.

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
JustinRChou: Updating the test name and source name.

rprouse:
Looks good, waiting for CI to pass.

JustinRChou:
I have a question about running all the current tests in Visual Studio 2015.
I am using the NUnit test adapters and it doesn't seem to finish.

rprouse:
@JustinRChou The Visual Studio adapter doesn't work well because of the many projects targeting the same frameworks which contain the same tests and test assemblies that aren't intended to be run directly, but are used in the unit tests themselves.

You should build and run the unit tests from the command line using a Visual Studio command prompt. You can do so from the solution root with the command `build.cmd -Target=Test`

CharliePoole:
Unfortunately, timeouts in the OSx build are all too common. I restarted it and it may complete but I'm going ahead with the merge anyway.

PR #1832, opened 2016-10-11
Update Build script and artefacts

### LINKED ISSUE #1609: Upgrade Cake build to latest version

### LINKED ISSUE #1811: Build.ps1 fails if spaces in path
If the nunit repo is in a location with spaces in the path, the build.ps1 bootstrapping script fails.

The [latest version from Cake](https://github.com/cake-build/cake/blob/develop/build.ps1) (26/09/2017) appears to fix this issue. It does however:

- Specify a specific Cake version. I haven't yet looked at how this interacts with our current method of specifying in packages.config, but we should look in to that.
- Install the .NET Core cli. Are we ok with this?

Duplicating this in nunit-console, as it looks to have the same build.ps1

ChrisMaddock:
Updates to Cake 15.2, makes use of some new features, and removes unused files. Fixes #1609.

**Changes made**
- Switch to `DotNetBuild`, and separate out CF/SL builds for future extraction
- Update powershell bootstrap script to cake latest (fixes #1811)
- Update mono builds to 4.2.4 and latest, as done in nunit-console
- Remove/Update format of SetUps/Teardowns - old style had deprecation warnings
- Upload artefacts in cake, rather than appveyor.yml
- Remove unused files
   - Licence.rtf used only in msi
   - nunit.ico used in msi and console, not in this repo anymore
   - Also removed files downloaded by cake bootstrapper, and old msbuild community tasks
- Added PackageCF to appveyor build - assumed it wasn't left out on purpose?

Ironically I didn't end up doing the one thing I started this for, which was uploading the test results to the appveyor interface. Decided as not all the tests ran in this repo produce an xml file for upload, we were probably better having no tests integrated, then just _some_ tests displayed. May add that to the console repo instead, in time.

For review: all the actual changes are in the first commit, the second is purely tidying up whitespace.


build.cake
```diff
@@ -547,6 +550,12 @@ void RunGitCommand(string arguments)
     });
 }
 
+void UploadArtifacts(string packageDir, string searchPattern)
+{
+    foreach(var zip in System.IO.Directory.GetFiles(packageDir, searchPattern))
```
CharliePoole: zip isn't such a good name here.

CharliePoole:
@ChrisMaddock Actually, I did have one question for you. Why do you feel it's better to do the package uploads in build.cake rather than appveyor.yml? Which should be our standard approach?

ChrisMaddock:
> Why do you feel it's better to do the package uploads in build.cake rather than appveyor.yml? Which should be our standard approach?

Two reasons:
- I wanted to move everything possible in to cake, and keep appveyor.yml and travis.yml as simple wrappers which call in to cake - and so keep everything important in one place. I'd like to carry this on with the rest of the content in appveyor.yml, as far as possible - but this PR was getting a little messy for review.
- We talked before about sharing cake tasks between repos - uploading `*.nupkg` and `*.zip` as artefacts is working towards something we could share. 😄

Issue #1808, opened 2016-09-27
Disambiguate error messages from EqualConstraint


CharliePoole:
Although we try to avoid it, EqualConstraint occasionally fails with an error message like

```
Expected: some_value
 But was: some_value <== Same value in both messages
```

Issue #1778 illustrates such a case, producing the message

```
Expected: 0
 But was: 0
```

It is proposed that we append the type in such a case.,.

```
Expected: 0 (System.Int32)
 But was: 0 (System.IntPtr)
```

Note this has to be done somewhere in the code where **both** the expected and the actual values are available and after the string representations have already been determined. Finding the right spot may be slightly tricky.

Notes:
1. We should only do this if the types are actually different, otherwise, do not bother.
2. If a tolerance or other constraint modifier is in effect, we should investigate whether to enable this feature. The best course may be different for each modifier.

oznetmaster:
If this can be done, it would be great!

jplebre:
What would happen if namespaces are incredibly long?
(Mysolution.myproject.tests.helpers.factory.stub.consumeraddress)
Or something silly like that.

Could it be worth to just have the type?

CharliePoole:
One could just use the name in most cases, I think.

oznetmaster:
Just use as much of the name as needed to make it unique (right to left).

JustinRChou:
Tentatively I wrote the following to get the difference type names. Still looking into how to get it into the Write function naturally.
```
        private void ResolveTypeNameDifference(object expected, object actual, out string expectedType, out string actualType) {
            string[] expectedOriginalType = expected.GetType().ToString().Split('.'); //NUnit.Framework.Constraints.Outer.Middle.Inner.Class1
            string[] actualOriginalType = actual.GetType().ToString().Split('.'); //NUnit.Framework.Constraints.Outer.Hey.Inner.Class1
            int actualStart = 0, expectStart = 0;
            for (int expectLen = expectedOriginalType.Length - 1,actualLen = actualOriginalType.Length - 1;
                expectLen >= 0 && actualLen >= 0;
                expectLen--, actualLen--) {
                if (expectedOriginalType[expectLen] != actualOriginalType[actualLen]) {
                    actualStart = actualLen;
                    expectStart = expectLen;
                    break;
                }
            }
            expectedType = String.Join(".", expectedOriginalType, expectStart, expectedOriginalType.Length- expectStart); //Middle.Inner.Class1
[...]

PR #1835, opened 2016-10-12
#1758 Added methods for Does.ContainKey and Does.ContainValue, added …

### LINKED ISSUE #1758: No direct inverse for Contains.Key
The `DictionaryContainsKeyConstraint` allows an assertion to be written against the presence of a key in a dictionary:
```
Assert.That(dictionary, Contains.Key("test"));
```
However it's not possible to _directly_ write the inverse - that a dictionary does not contain a key:
```
Assert.That(dictionary, Does.Not.ContainKey("test"));
```
The closest is probably to inspect the dictionary's `Keys` property and assert on that:
```
Assert.That(dictionary.Keys, Does.Not.Contains("test"));
```
Aside from the lack of symmetry, the grammar of this approach is slightly jarring and a dedicated `ContainKey` method would be preferable.

As an aside it's possible to solve this locally in the test code using an extension method:
```
public static class Extensions
{
	public static DictionaryContainsKeyConstraint ContainKey(this ConstraintExpression constraint, string value)
	{
		var keyConstraint = new DictionaryContainsKeyConstraint(value);

		constraint.Append(keyConstraint);

		return keyConstraint;
	}
}
```

JustinRChou:
…units tests.

Added methods for Does.ContainKey and Does.ContainValue.

Added unit tests for new methods. Not sure if they are sufficient.

Added unit test for dictionaries under
CollectionContainsConstraintTests.

Fixes #1758


src/NUnitFramework/tests/Constraints/CollectionContainsConstraintTests.cs
```diff
@@ -51,6 +51,32 @@ public void CanTestContentsOfObjectList()
             Assert.That(list, new CollectionContainsConstraint(item));
         }
 
+        [Test]
+        public void CanTestContentsOfDictionary()
```
rprouse: This is a pretty massive test that does many things. I would prefer if it was broken down into smaller tests. It looks like your kvExists123 are good candidates for a `TestCaseSource` and your kvDoesNotExist for a separate test?

CharliePoole:
I agree - good to merge if the test is fixed.

src/NUnitFramework/tests/Constraints/CollectionContainsConstraintTests.cs
JustinRChou: Do i keep the CollectionContainsConstraint test in CollectionContainsConstraintTests and move the Contain asserts to TestCaseSource or move everything to TestCaseSource?

rprouse:
You can leave the tests here. I just suggested TestCaseSource as a way of writing the tests to reduce duplication.

JustinRChou:
I see now. Rewriting it using the testsource attributes then.

JustinRChou:
Is it an issue when Does.Contain refers to the CollectionContainsConstraint but Does.Not.Contain refers to ContainsConstraint?

It was causing me some confusion.

CharliePoole:
@JustinRChou You appear to have added a commit that's intended for issue #1808. You should use different local branches and create different PRs for resolving each issue, unless two issues have some close relationship to one another. That makes it faster to finish reviewing each thing and get it merged. Adding new stuff makes it a moving target for us.

As far as I can tell, the last commit that relates to issue #1758 is the one with the description "Reorganize the unit test" Is that correct? If you need help resetting the branch to that commit while preserving your new work on a new local branch, let me know. Meanwhile, this branch and PR should wait for @rprouse to review the change he requested.

CharliePoole:
Looks good now. Just waiting for @rprouse to re-review.

CharliePoole:
@rprouse This can't be merged until you re-review, according to the new review system. That's one of the things I dislike about how it works. [I realize I can override and merge it, but that's not playing by the rules.]

rprouse:
Interesting point on the review system. I didn't realize another team member can't just switch to Approved. We'll need to think on that...

CharliePoole:
Most likely, you or I as owners can do anything. But we shouldn't rely on
that.

PR #1845, opened 2016-10-14
Allow delayed assert to be set in minutes

### LINKED ISSUE #1837: [Request] AfterContraint to support more readable usage
As a developer
I want to be able to use the AfterConstraint in a more readable and DSL like fasion
so that anyone else that reads my code, quickly understands what the numbers means.

**Background**
Today when using the After constraint you have to define the delay in milliseconds. If I then want to wait for 3 minutes I can do one of the two:
Specify the delay in milliseconds: 
`Assert.That(checkStatusFunc, Is.True.After(180000));`
Our use a `TimeSpan`: 
`Assert.That(checkStatusFunc, Is.True.After((int) TimeSpan.FromMinutes(3).TotalMilliseconds));`

None of them is easy to read.
Suggestion, that looks the same as you can do with the WithinConstraint.
`Assert.That(checkStatusFunc, Is.True.After(3).Minutes);`

If you want to go further you could add the polling interval in the same manner.
`Assert.That(checkStatusFunc, Is.True.After(3).Minutes.ByPollingEvery(1).Seconds);`

rvignesh89:
Fixes #1837 

This was a simple change indeed. If you think this makes sense I could continue doing something similar for polling interval. I'll also cleanup the magic numbers in the tests along the way.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
@@ -37,7 +37,7 @@ namespace NUnit.Framework.Constraints
     {
         // TODO: Needs error message tests
 
-        private readonly int delayInMilliseconds;
+        private int delayInMilliseconds;
         private readonly int pollingInterval;
 
         ///<summary>
@@ -72,6 +72,18 @@ namespace NUnit.Framework.Constraints
         public override string Description
         {
             get { return string.Format("{0} after {1} millisecond delay", BaseConstraint.Description, delayInMilliseconds); }
+        }
+
+        /// <summary>
+        /// Converts delayInMilliseconds input from minutes to milliseconds
+        /// </summary>
+        public DelayedConstraint Minutes
+        {
+            get
+            {
+                delayInMilliseconds = (int)TimeSpan.FromMinutes(delayInMilliseconds).TotalMilliseconds;
+                return this;
+            }
         }
 
         /// <summary>
--- a/src/NUnitFramework/tests/Constraints/DelayedConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/DelayedConstraintTests.cs
@@ -103,6 +103,13 @@ namespace NUnit.Framework.Constraints
         }
 
         [Test]
+        public void SimpleMinutesTest()
+        {
+            SetValuesAfterDelay(60000);
+            Assert.That(DelegateReturningValue, new DelayedConstraint(new EqualConstraint(true), 1).Minutes);
+        }
+
+        [Test]
         public void SimpleTestUsingBoolean()
         {
             SetValuesAfterDelay(DELAY);


src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -75,6 +75,18 @@ public override string Description
         }
 
         /// <summary>
+        /// Converts delayInMilliseconds input from minutes to milliseconds
+        /// </summary>
+        public DelayedConstraint Minutes
+        {
+            get
+            {
+                delayInMilliseconds = (int)TimeSpan.FromMinutes(delayInMilliseconds).TotalMilliseconds;
```
CharliePoole: Suppose I use `After(60).Minutes.Minutes`?

Take a look at how EqualConstraint handles Minutes as well as the other time units. There is a tolerance member with a numeric value and a unit. The unit gets set by the modifier property. If more than one is used, we use the last one seen. I could have generated a runtime error, but that seems unnecessary and a compile-time error is much more complicated.

Additionally, if we are having minutes, let's have Milliseconds and Seconds as well. You can initialize the units to milliseconds. Longer than minutes, seems silly in this context.

rprouse:
@rvignesh89 do you need some help with the changes that @CharliePoole requested?

CharliePoole:
Wish we could set a timer on an issue... to remember to look back at it in X number of days!

rvignesh89:
@rprouse @charliepoole I've been traveling over the last couple of days so couldn't follow up on this. I started to use the Tolerance class and could make it work. But noticed the description of DelayedConstraint has to include the variable factor as well. Working on that now.

CharliePoole:
@rvignesh89 IIRC the Tolerance class is quite small, with just two members and its units include things that don't make sense in this contaxt. I suggest copying it as an internal class, removing what is not needed and giving it an appropriate name, since we are not using it as a tolerance.

src/NUnitFramework/framework/Constraints/Tolerance.cs
```diff
@@ -217,5 +217,16 @@ public bool IsUnsetOrDefault
...
+        /// <summary>
+        /// Returns a string that represents the current object.
+        /// </summary>
+        /// <returns>
+        /// A string that represents the current object.
+        /// </returns>
+        public override string ToString()
+        {
+            return string.Format("{0} {1}", ((TimeSpan)amount).Milliseconds, "millisecond");
+        }
```
CharliePoole: Why are we changing Tolerance.cs? Should reset.

CharliePoole: Additionally, the code seems wrong - amount is an object with may or may not be a timespan. It might be a double, for example, used in comparing two doubles for equality. Anyway, please reset to HEAD.

CharliePoole:
Unfortunately, the new class breaks the compact framework build. Easy enough to fix if you have VS2008 but if you don't, just make it a nested class inside the constraint and get rid of the new file. We'll have CF out of this build in a few weeks.

rvignesh89:
Oh, I didn't see your message just now. Let me do what you suggested

rvignesh89:
@CharliePoole I manually added the new class into the csproj for the cf framework project to trigger a compile. I think this should also fix the build error.

rvignesh89:
@CharliePoole I think the build should be fine now. 

As for the change itself I've followed your suggestion to create an internal class called Interval to track the delay. There is one thing I haven't done is to test the Description property in the DelayedConstraint class for the different messages generated for the different delays. 

I noticed the ContrainstTestBase is not suitable for me to make test a description which is dynamic in nature. So I'll have to add these special Description Tests into DelayedConstrainTest itself. What do you think about that?

rvignesh89:
Oops. This code shouldn't have been committed. I'll remove my changes in Tolerance.cs

On 20 Oct 2016, at 18:58, CharliePoole <notifications@github.com<mailto:notifications@github.com>> wrote:


@CharliePoole commented on this pull request.

________________________________

In src/NUnitFramework/framework/Constraints/Tolerance.cs<https://github.com/nunit/nunit/pull/1845>:

> @@ -217,5 +217,16 @@ public bool IsUnsetOrDefault
         {
             get { return mode == ToleranceMode.Unset; }
         }
+
+        /// <summary>
+        /// Returns a string that represents the current object.
+        /// </summary>
+        /// <returns>
+        /// A string that represents the current object.
+        /// </returns>
+        {
+            return string.Format("{0} {1}", ((TimeSpan)amount).Milliseconds, "millisecond");
+        }


Additionally, the code seems wrong - amount is an object with may or may not be a timespan. It might be a double, for example, used in comparing two doubles for equality. Anyway, please reset to HEAD.

-

src/NUnitFramework/framework/Constraints/Tolerance.cs
rvignesh89: I've reverted Tolerance.cs to the original version

CharliePoole:
Looks good. Waiting for second review.

src/NUnitFramework/tests/Assertions/AssertionHelperTests.cs
```diff
@@ -319,23 +319,23 @@ public void After()
...
+            Expect(constraint.ToString(), EqualTo("<after 1000 millisecond <property X <equal 10>>>"));
         }
 
         [Test]
         public void After_And()
         {
             var constraint = GreaterThan(0).And.LessThan(10).After(1000);
             Expect(constraint, TypeOf<DelayedConstraint>());
-            Expect(constraint.ToString(), EqualTo("<after 1000 <and <greaterthan 0> <lessthan 10>>>"));
+            Expect(constraint.ToString(), EqualTo("<after 1000 millisecond <and <greaterthan 0> <lessthan 10>>>"));
```
rprouse: I would like to see additional tests here using the various intervals

src/NUnitFramework/tests/Syntax/AfterTests.cs
```diff
@@ -55,7 +55,7 @@ public class AfterTest_AndOperator : SyntaxTest
         [SetUp]
         public void SetUp()
         {
-            parseTree = "<after 1000 <and <greaterthan 0> <lessthan 10>>>";
+            parseTree = "<after 1000 millisecond <and <greaterthan 0> <lessthan 10>>>";
```
rprouse: I would also like to see additional tests here using the various intervals

src/NUnitFramework/tests/Assertions/AssertionHelperTests.cs
rvignesh89: I'm curious if additional intervals needs to be tested here? Since this looks like a test for the AssertionHelper and not related to After.

src/NUnitFramework/tests/Syntax/AfterTests.cs
rvignesh89: It makes sense to add more tests to covert different intervals here. Will do that soon.

src/NUnitFramework/tests/Assertions/AssertionHelperTests.cs
rprouse: I would like to see at least one of the intervals used to ensure that it changes the message correctly. You have some other tests, but this will double check it in combination.

Issue #890, opened 2015-10-13
Allow file references anywhere in the command line.


oznetmaster:
Currently, the --testlist option allows a file containing a list of tests to be specified in the command line.

It may make sense to provide a more general mechanism for allowing files of arbitrary text to be specified on the command line.

This would replace the --testlist option.

One possibility would be to allow something like `[<filepath>]` to be used anywhere on the command line, and would result in the entire contents of the file being inserted in the command line at that point.  All newlines in the file would be converted to spaces.

In fact, the only thing on the command line could be a file reference, and the entire command line would then be retrieved from the file.

As an extension, it might be useful to allow an extension to this syntax `[<filepath>,<delimiter>]`, where `<delimiter>` would specify the string that newlines would be replaced with in the file (for instance a comma).

File references should be nestable up to some limit (three, perhaps).

The processing of the file references would occur at the top level, before any actual parsing of the command line occurred.

CharliePoole:
I like the idea of allowing options - and possibly even the names of assemblies -  to be placed in some sort of response file. I don't want to over-engineer it however, especially since we would be originating this feature ourselves, rather than in response to user request.

Some thoughts...

What about the user of @filename, which is used by some other programs? The contents of filename would simply be inserted in the command-line, just as you are suggesting.

We may not want to completely replace --testlist, since it is used for simple lists of tests, one per line, usually generated by external programs. Although... it would only require such programs to put --test at the start of each line.

Nesting seems unnecessary as does the specification of delimiters.

CharliePoole:
We will, of course, have to decide what command-line we are talking about: console, nunitlite or both.

oznetmaster:
Well, I am a user, and I need it badly, since CF has such a short command line limit.

Almost all my tests already use the --testlist option if there are more then two or three tests in the list (using full names), and that has hit the command line limit on a couple of occasions.

oznetmaster:
@filename is fine for me, as long as it will not be confused with any other commandline object element.

Reason for the delimiter is that it allows things like --testlist to be emulated, where the tests can be listed one per line, and they are processed correctly.  Without it, the tests listed in the file would have to all have commas after them, except for the last one.

oznetmaster:
Nesting is required if we allow entire command lines in the file.  The command line may well need to reference a list of test generated elsewhere, for instance.

oznetmaster:
I would think we would want to be consistent across both console and nunitlite?

CharliePoole:
[...]

PR #1846, opened 2016-10-15
nunitlite implementation of the @filename option, as described in issue #890

oznetmaster:
This is the nunitlite implementation of the @filename option, as described in issue #890.

It allows nesting (max 3) and an alternative delimiter for multi-line files.

Tests have been added for the various formats.

### DIFF
--- a/src/NUnitFramework/nunitlite/CommandLineOptions.cs
+++ b/src/NUnitFramework/nunitlite/CommandLineOptions.cs
@@ -124,17 +124,26 @@ namespace NUnit.Common
                     continue;
                 }
 
-#if NETCF
                 string contents;
-                using (var sr = new StreamReader (filename))
+                try
                     {
-                    contents = sr.ReadToEnd ();
-                    }
+#if NETCF
+                    using (var sr = new StreamReader (filename))
+                        {
+                        contents = sr.ReadToEnd ();
+                        }
 #else
-                var contents = File.ReadAllText(filename);
+                    contents = File.ReadAllText (filename);
 #endif
+                    }
+                catch (IOException ex)
+                    {
+                    ErrorMessages.Add("Error reading \"" + filename + "\": " + ex.Message);
+                    continue;
+                    }
+
                 var newArgs = GetArgs(contents.Replace("\r", "").Replace('\n', delim));
-                listArgs.AddRange(PreParse(newArgs));
+                    listArgs.AddRange(PreParse(newArgs));
             }
 
             --_nesting;
--- a/src/NUnitFramework/nunitlite/CommandLineOptions.cs
+++ b/src/NUnitFramework/nunitlite/CommandLineOptions.cs
@@ -24,6 +24,8 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
 using NUnit.Options;
 
 namespace NUnit.Common
@@ -53,7 +55,7 @@ namespace NUnit.Common
         private bool noresult;
 #endif
 
-#region Constructor
+        #region Constructor
 
         internal CommandLineOptions(IDefaultOptionsProvider defaultOptionsProvider, params string[] args)
         {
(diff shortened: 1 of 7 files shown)


oznetmaster:
Once again, the Travis CI builds just never run :(

CharliePoole:
They seem to be running and passing.

oznetmaster:
Huh?

CharliePoole:
This page https://github.com/nunit/nunit/pull/1846 says all checks have passed. Here are the two travis runs you wrote about:

https://travis-ci.org/nunit/nunit/builds/168455865
https://travis-ci.org/nunit/nunit/builds/168455913

oznetmaster:
Since this is locking my work on another issue, which requires modifying the same source files, it would be good to know if there are any problems with it.

src/NUnitFramework/nunitlite/CommandLineOptions.cs
```diff
@@ -64,18 +66,102 @@ internal CommandLineOptions(IDefaultOptionsProvider defaultOptionsProvider, para
...
+                    ErrorMessages.Add("The file \"" + filename + "\" was not found");
+                    continue;
+                }
+
+                string contents;
+                try
+                    {
+#if NETCF
+                    using (var sr = new StreamReader (filename))
+                        {
```
rprouse: Minor, but I would prefer if the indenting was consistent here with the rest of the project.

src/NUnitFramework/nunitlite/CommandLineOptions.cs
```diff
@@ -64,18 +66,102 @@ internal CommandLineOptions(IDefaultOptionsProvider defaultOptionsProvider, para
...
+#endif
+        }
+
+#if !PORTABLE
+        internal static IEnumerable<string> GetArgs(string commandLine)
+        {
+            const string re = @"\G(""((""""|[^""])+)""|(\S+)) *";
+            var ms = Regex.Matches(commandLine, re);
+            return ms.Cast<Match>().Select(m => Regex.Replace(m.Groups[2].Success ? m.Groups[2].Value : m.Groups[4].Value, @"""""", @""""));
+        }
```
rprouse: Your code seems cleaner, but it bothers me that we have two entirely different implementations of this method, here and in https://github.com/nunit/nunit-console/pull/98/files. I do prefer using the named groups in the other rather than positional.

BTW, I hate code reviewing RegEx 😝

src/NUnitFramework/nunitlite/CommandLineOptions.cs
```diff
@@ -64,18 +66,102 @@ internal CommandLineOptions(IDefaultOptionsProvider defaultOptionsProvider, para
...
-                Parse(args);
+                Parse(PreParse(args));
         }
-        
+
+#if !PORTABLE
+        private int _nesting = 0;
+#endif
+
+        internal IEnumerable<string> PreParse(IEnumerable<string> args)
```
rprouse: This is another method that is very similar to the Expand function in nunit/nunit-console#98. @oznetmaster do you think you and @NikolayPianikov could work together to bring your two implementations closer together? If we find bugs in the future, it would be easier to fix it in multiple places if we were dealing with nearly identical code.

oznetmaster:
All I did was backport the code I wrote over a year ago for my CF build.

I had not looked at the other code until after I had done so.

I am concerned that the RegExin the nunit/nunit-console#98 does not seems to handle nested quotation marks correctly.  However, like you, I have trouble reviewing RegEx  The RegEx I used was obtained from Microsoftas the recommended regex for parsing command line arguments.  It has not failed on any test I have ever used with it, and I use that RegEx in many different programs.

If @NikolayPianikov would rather just port his code to nunitlite, that is fine.  However, his does not support alternate delimiters, or nesting, which mine does.

rprouse:
@oznetmaster I would actually prefer if some of this code moved into @NikolayPianikov's although I like that he separated out your `GetArgs` class to a separate helper class that can be tested on its own.

At a minimum, can we make the names of the similar methods the same?

oznetmaster:
I did not see the need, since that was Microsoft provided code.  It could be, I guess.

I am currently working on backporting the parameter discovery/validation code, so perhaps @NikolayPianikov could rename similar methods?

rprouse:
@NikolayPianikov made the changes at his end, so I think we can merge this. @CharliePoole can you give this a second review? You were more involved in the feature in the console, so I would like your input. If it is good, go ahead and merge.

oznetmaster:
Do you want to port the additional features in the nunitlite implementation to the nunit-console one (recursion and alternative delimters)?

CharliePoole:
Unfortunately I didn't realize we had included these extra features in NUnitLite when I merged it.

@rprouse We need to decide what features we want and what we don't and bring the two together.

oznetmaster:
My users have found the recursion (I allow only three levels, but that is arbitrary) really useful.

I have only seen two levels used (one level being the command line which nests another files containing standard options, but no file or test or category references, which are on the top level command line).

rprouse:
I also didn't twig on the fact that this included the extra features even though you did mention it. With Hacktoberfest, there has been way too many pull requests coming in 😄 

What is the alternative delimiter and why do we need alternatives?

I can see the use of two levels of nesting. The top level file points to a bunch of other files and those contain command lines. You could then create top level files for different run configurations. That said, I always thought people would just use one at the command line to simplify things. I hate adding complexity if it isn't widely used.

For now, I think I would like to leave these extra features out. If you want, you can create an issue in the console repo for these enhancements and we can let the community decide.

oznetmaster:
I only back-ported my existing code that my user base has been using for over a year.

They use both features (nesting and alternative delimiter).

Nest I discussed further above.  I believe at least one level of nesting is a requirement, and I see it used all the time.

Alternative delimiter allows the included file to be multiline, and to have those lines separated by some other character then a space (the default option per line format) when merged into the command line.  For instance, it can be a comma (comma separated list, one per line, similar to the format of a testlist file).  Sample is in the CommandLineTests.  Since any desired delimiter can always be explicitly included on each line, this is definitely not required, and is only syntactic sugaring.

oznetmaster:
I guess I am recommending we leave at least one level of nesting (currently my code has two - it is only a constant), and remove the alternative delimiter.  Then modify the console code to support the same syntax.

CharliePoole:
See issue #1877

oznetmaster:
Should my comments be moved to this new issue?

CharliePoole:
My pointer to #1877 creates a back pointer there, to this issue, so it's all available.

PR #1847, opened 2016-10-15
Support nullable booleans and enums in Theories

### LINKED ISSUE #1836: Support nullable enums in Theories
I would like nUnit to support nullable enums in [Theory] tests, it should iterate all values plus provide null case.

[Theory]
public void AppShouldWork(MyEnum? nullableValue) {...}

bartw:
Fixes #1836

### DIFF
--- a/src/NUnitFramework/framework/Internal/Builders/DatapointProvider.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/DatapointProvider.cs
@@ -132,6 +132,13 @@ namespace NUnit.Framework.Internal.Builders
 
             if (datapoints.Count == 0)
             {
+                var underlyingParameterType = Nullable.GetUnderlyingType(parameterType);
+                if (underlyingParameterType != null)
+                {
+                    datapoints.Add(null);
+                    parameterType = underlyingParameterType;
+                }
+
                 if (parameterType == typeof(bool))
                 {
                     datapoints.Add(true);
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/framework/Internal/Builders/DatapointProvider.cs
```diff
@@ -132,6 +132,13 @@ public System.Collections.IEnumerable GetDataFor(IParameterInfo parameter)
 
             if (datapoints.Count == 0)
             {
+                var underlyingParameterType = Nullable.GetUnderlyingType(parameterType);
+                if (underlyingParameterType != null)
+                {
+                    datapoints.Add(null);
+                    parameterType = underlyingParameterType;
+                }
+
```
CharliePoole: This will add a null for any underlying type that is not supplied, not just bool and enum.

CharliePoole:
This looks good to me if you can avoid the null value for double? etc. when no data is provided.

bartw:
I added a test for the problem with the unsupported types.
I made the fix so that when new types are supported their nullable counterparts will work out of the box.

CharliePoole:
Looks good to merge once CI passes. Will need one more committer to review.

rprouse:
Looks good to me too. Thanks for the help.

seraphim207:
When do you plan to release this update?

CharliePoole:
We just released NUnit 3.5. The 3.6 release is scheduled at year end.

rprouse:
@seraphim207 if you need this now, you can always use our MyGet feed and the CI builds of master.

PR #1848, opened 2016-10-16
Resolve ambiguous messages for EqualConstraint

JustinRChou:
Added type name for equal constaint.

Added unit tests.

Updated failed unit tests due to the addition of the type name in
exception message.


CharliePoole:
@JustinRChou I changed your title to reflect what this PR is trying to fix and moved the issue number into the description using a comment in the correct form (Fixes or Fix plus number) so that GitHub will notice it. This will show a cross-reference in the original issue and close the issue when we merge this.

src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
```diff
@@ -54,10 +54,14 @@ public class TextMessageWriter : MessageWriter
...
         /// </summary>
         public static readonly int PrefixLength = Pfx_Expected.Length;
-        
+
         #endregion
 
+        #region Instance Fields
         private int maxLineLength = DEFAULT_LINE_LENGTH;
+        private bool diffTypes = false;
+        private string _expectedType, _actualType;
```
CharliePoole: If the corresponding properties are eliminated, these instance fields don't seem to be needed either since they are only used locally in various methods as far as I can see.

src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
```diff
@@ -252,7 +305,9 @@ private void WriteExpectedLine(object expected, Tolerance tolerance)
         {
             Write(Pfx_Expected);
             Write(MsgUtils.FormatValue(expected));
-
+            if (diffTypes) {
```
CharliePoole: This writes the type whenever the two types differ, but we only wanted to do it when the types differ and the display strings for the two values are identical.

JustinRChou:
@rprouse Made the changes based on your suggestion.

Left the MsgUtils.FormatValue issue alone.

rprouse:
I am okay with this change. @CharliePoole do you want to take a second look?

rprouse:
@CharliePoole just trying a new twist on the reviews for this PR. You had denied requesting changes, but didn't update after the requested changes were made. I was waiting for you to change it, but you have been busy. I noticed that I can dismiss your review with comments, so I did that.

I think that is a possible way to get around your concerns with the limitations of the new review process. It doesn't require us to use admin privileges.

@JustinRChou I am merging, thanks for your contribution.

CharliePoole:
@rprouse That's OK with me. I still don't like the new reviews that much and I'm a big fan of using our judgement!

Issue #1281, opened 2016-02-20
TestCaseSourceAttribute to support optional arguments


ChrisMaddock:
From #1279. `TestCaseAttribute` supports a number of 'special-cases' that `TestCaseSourceAttribute` does not - principally optional arguments and the `params` keyword.

This is managed within `TestCaseAttribute.GetParametersForTestCase(IMethodInfo method)`. This functionality could probably be shared by both attributes, however, there's some additional complexity in how `TestCaseSourceAttribute.GetTestCasesFor(IMethodInfo method)` derives the type of it's parameters which will need modification.

CharliePoole:
Params arguments are deliberately absent from TestCaseSource because it never seemed useful to me. If you are returning args from a method, you simply return the final set of values as an array. The syntactic icing of `params` isn't really needed.

Optional arguments are missing because they were never implemented.

It may be useful to know that the code for these two attributes (and some others) was once combined in the class that created the tests. The attributes were just "dumb" argument-holders at that time. When we went to NUnit 3.0, I made a tradeoff of having independent attributes at the price of some duplication.

If you want to work on this, I'll be glad to add your name but you should probably come up with a plan for review first, including a few examples of user code that does not currently work but that you want to work. I don't want you to write code that we might turn down for non-code reasons.

CharliePoole:
Another take on this: I consider TestCase as being a starting point. Peole using TestCaseSource should be more knowledgable and have less need of syntactic niceties.

ChrisMaddock:
@CharliePoole That's a valid point about params. I'd mainly considered

ChrisMaddock:
@CharliePoole That's a valid point about params not being so necessary. My main thought behind this is that I was expecting that the two attributes to interpret arguments in the same way. (While implementing optional arguments in TestCase, I expected to just put similar in to TestCaseSource - but didn't realise the special-cases implemented separately, per attribute.)

Can you clarify what you mean by plan? As in, how the code would be structured? I'd like to look at it - although it may be a few days before I can look into it properly.

CharliePoole:
Strictly speaking, the attributes don't have the same "arguments"  Those of `TestCase` are the actual arguments of the test method. Those of `TestCaseSource` are info about where to find those arguments. Additionally, `TestCase` has special considerations because some test arguments can't be specified as constructor argumetns on an attribute. For example, you can't use a `decimal` so we have special conversion in `TestCase`.

[...]

PR #1850, opened 2016-10-18
Support for param array argument in TestCaseSourceAttribute

cPetru:
Hi, 

I noticed that TestCaseSourceAttribute does not implement support for params array argument as the TestCaseAttribute does. So I extracted the method that does that in the TestCaseAttribute and invoked it in the TestCaseSourceAttribute. 

I also included in the extracted method the code that handles the optional arguments, just to align the 2 attributes.

Br, 
Petru

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
@@ -270,70 +270,11 @@ namespace NUnit.Framework
 
                 IParameterInfo[] parameters = method.GetParameters();
                 int argsNeeded = parameters.Length;
-                int argsProvided = Arguments.Length;
 
                 parms = new TestCaseParameters(this);
 
-                // Special handling for params arguments
-                if (argsNeeded > 0 && argsProvided >= argsNeeded - 1)
-                {
-                    IParameterInfo lastParameter = parameters[argsNeeded - 1];
-                    Type lastParameterType = lastParameter.ParameterType;
-                    Type elementType = lastParameterType.GetElementType();
-
-                    if (lastParameterType.IsArray && lastParameter.IsDefined<ParamArrayAttribute>(false))
-                    {
-                        if (argsProvided == argsNeeded)
-                        {
-                            Type lastArgumentType = parms.Arguments[argsProvided - 1].GetType();
-                            if (!lastParameterType.GetTypeInfo().IsAssignableFrom(lastArgumentType.GetTypeInfo()))
-                            {
-                                Array array = Array.CreateInstance(elementType, 1);
-                                array.SetValue(parms.Arguments[argsProvided - 1], 0);
-                                parms.Arguments[argsProvided - 1] = array;
-                            }
-                        }
-                        else
-                        {
-                            object[] newArglist = new object[argsNeeded];
-                            for (int i = 0; i < argsNeeded && i < argsProvided; i++)
-                                newArglist[i] = parms.Arguments[i];
-
-                            int length = argsProvided - argsNeeded + 1;
-                            Array array = Array.CreateInstance(elementType, length);
-                            for (int i = 0; i < length; i++)
-                                array.SetValue(parms.Arguments[argsNeeded + i - 1], i);
-
-                            newArglist[argsNeeded - 1] = array;
-                            parms.Arguments = newArglist;
-                            argsProvided = argsNeeded;
-                        }
-                    }
-                }
-
-#if !NETCF
-                //Special handling for optional parameters
-                if (parms.Arguments.Length < argsNeeded)
-                {
-                    object[] newArgList = new object[parameters.Length];
-                    Array.Copy(parms.Arguments, newArgList, parms.Arguments.Length);
-
-                    //Fill with Type.Missing for remaining required parameters where optional
-                    for (var i = parms.Arguments.Length; i < parameters.Length; i++)
-                    {
-                        if (parameters[i].IsOptional)
-                            newArgList[i] = Type.Missing;
-                        else
-                        {
-                            if (i < parms.Arguments.Length)
-                                newArgList[i] = parms.Arguments[i];
-                            else
-                                throw new TargetParameterCountException("Incorrect number of parameters specified for TestCase");
-                        }
-                    }
-                    parms.Arguments = newArgList;
-                }
-#endif
+                SpecialArgumentsHandling(parms, parameters);
+                int argsProvided = parms.Arguments.Length;
 
                 //if (method.GetParameters().Length == 1 && method.GetParameters()[0].ParameterType == typeof(object[]))
                 //    parms.Arguments = new object[]{parms.Arguments};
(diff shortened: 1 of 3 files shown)


rprouse:
This is causing 3 test failures.

```
1) Failed : NUnit.Framework.Attributes.TestCaseSourceTests.SourceInAnotherClassPassingParamsToField
  Expected string length 217 but was 85. Strings differ at index 0.
  Expected: "You have specified a data source field but also given a set o..."
  But was:  "System.NullReferenceException : Object reference not set to a..."
  -----------^
at NUnit.Framework.Attributes.TestCaseSourceTests.SourceInAnotherClassPassingParamsToField()
2) Failed : NUnit.Framework.Attributes.TestCaseSourceTests.SourceInAnotherClassPassingParamsToProperty
  Expected string length 225 but was 85. Strings differ at index 0.
  Expected: "You have specified a data source property but also given a se..."
  But was:  "System.NullReferenceException : Object reference not set to a..."
  -----------^
at NUnit.Framework.Attributes.TestCaseSourceTests.SourceInAnotherClassPassingParamsToProperty()
3) Failed : NUnit.Framework.Attributes.TestCaseSourceTests.SourceInAnotherClassPassingSomeDataToConstructorWrongNumberParam
  Expected string length 291 but was 85. Strings differ at index 0.
  Expected: "You have given the wrong number of arguments to the method in..."
  But was:  "System.NullReferenceException : Object reference not set to a..."
  -----------^
at NUnit.Framework.Attributes.TestCaseSourceTests.SourceInAnotherClassPassingSomeDataToConstructorWrongNumberParam()
```

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
```diff
@@ -360,6 +301,79 @@ private TestCaseParameters GetParametersForTestCase(IMethodInfo method)
...
+                        else
+                            throw new TargetParameterCountException("Incorrect number of parameters specified for TestCase");
+                    }
+                }
+                parms.Arguments = newArgList;
+            }
+#endif
+        }
+
+        /// <summary>
```
CharliePoole: It's not a good idea to refactor an unrelated class when fixing something different. Remember that the reviewers have to look at every change you make, line by line, to see what it does. There's nothing in GitHub to tell us "He only extracted the method but didn't make changes." OK for this time, but please separate refactoring from features the next time.

src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
```diff
@@ -241,6 +246,8 @@ private IEnumerable<ITestCaseData> GetTestCasesFor(IMethodInfo method)
                             foreach (string cat in this.Category.Split(new char[] { ',' }))
                                 parms.Properties.Add(PropertyNames.Category, cat);
 
+                        TestCaseAttribute.SpecialArgumentsHandling(parms, method.GetParameters());
```
CharliePoole: Ah! Now I see why you refactored. Please don't make one attribute depend on another like this. Aside from having to work correctly, our attributes are supposed to show people how to write attributes and they can't depend on this sort of thing. In this case, you could create a helper class that does the job easily enough. However, I suggest you don't make any code changes till we decide whether we want this feature... see overall comments.

CharliePoole:
@cPetru It's usually a better strategy to start with an issue submittal, to find out if we will include a feature in NUnit before you write the code. Absent that, we can discuss it here on the PR but, of course, you have already taken the time to implement it.

There are two features here: (1) recognition of parameter arrays and (2) optional arguments. Both features were left out of TestCaseSourceAttribute by design, because of how the class is typically instantiated and used.

TestCaseSourceAttribute is generally expected to be used with TestCaseData although in very specific circumstances you can pass an object array of parameters or (in even more specific cases) a typed array of some kind. IME these features add more exceptions to the exception case and make the rules for what you can use as data for the attribute extremely difficult to explain for users.

I'm not absolute on this. To be convinced, I need to see some use cases like: I wanted to write _this_ but I had to write _this_. Can you do that?

ChrisMaddock:
I think this would be a sensible feature, purely because I _expect_ `TestCaseSourceAttribute` to have the same functionality as `TestCaseAttribute`. I don't see a solid reason why it can't - call it [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). 😄 

I raised this previously as issue #1281 - but never got back to it. Thanks for putting the PR together @cPetru!

CharliePoole:
@ChrisMaddock This seems to be a general problem with those things we categorize as "ideas" and put into the "Discussion" pipeline. Generally, we don't get much participation in pre-coding discussions and the ideas stay ideas. @rprouse and I have been talking about doing a big clean-out of those ideas that have never gone anywhere - one of which I think is #1281.

We need to figure out some better way of handling these things. Maybe doing it in issues is not the right way because we don't get much user input - except insofar as we are ourselves users. Anyway, we should talk about this offline.

IMO, the need for these two features on TestCaseAttribute comes from the fact that you input the actual argument list as a constructor to the attribute itself. This seems totally different from TestCaseSourceAttribure, where you simply give a pointer to where arguments may be found. To me, TestCaseAttribute "feels" like a call to a method and so should have the same features as a method call, whereas, the source data for TestCaseSourceAttribute is either a data declaration or a return value from a program. I have never felt a need to use shortcuts in either the data declaration or in the value my code returns... it's my code, so I should just spell it out.

However, let's say we end up agreeing these are desirable features... Are we saying they are desirable in TestCaseAttribute and TestCaseSourceAttribute only? Or should every attribute that creates test data enjoy the same features? If we go with the former approach, then this implementation is correct. OTOH, if we select the latter, then the processing actually belongs somewhere entirely different: in the NUnit code that processes the test case data returned from data attributes.

So, I'd say our priority is to get some feedback as to whether this is needed for TestCaseSourceAttribute, then to examine whether we want it to be implied in every data attribute and finally to review the implementation. For the first thing, can we have examples? 

Also, can we get more actual users to tell us what they think? By that I mean users who don't work on NUnit, 'cause those are the people to whom we try to defer in designing features - if only they will talk to us!

cPetru:
Hi Gents, 

I'm overwhelmed with the amount of feedback. I appreciate it. Thank you.

I faced this problem while trying to add more complicated test cases, so I wanted to move the cases from the decoration of the method to an other part of the file, for readability. 

So I notice that moving the cases into a property like 
  `public static readonly TestCaseData[] Cases = { new TestCaseData( new object[] { ... } };`
makes the case not runnable. The documentation does not suggest it. So I started looking into the framework as homework. 

Yes, I was also not happy with the dependency between the 2 attributes, but there was a reason for it: 
On my home machine (linux host, VirtualBox, Microsoft Edge provided virtual machine, VS2015 comunity, Git, clean NUnit sources) the build.cmd fails at some step. So I tested with NUnitGui loading the 4.5 test.dll, and it was not all green. So I decided to go for the minimal amount of changes and also not create new files (for which I should also touch the .csproj files). After adding my changes, the same tests were failing.  So I concluded that my changes might bring some positive value without breaking more, hence the PullRequest.

I also noticed the Fixture part, and I would gladly try to implement all 4 of them. 

All in all, I kind of need this feature, at least for cases, not fixtures. 

Regards, 
Petru

CharliePoole:
OK, I'm convinced it's a good feature. Let's address the second point. @rprouse @ChrisMaddock Should this apply to every attribute that generates test cases? Or should the attribute do it itself?

If it should apply generally, then our internal code that creates test cases should get this fix. If each attribute decides, then we need to expose in a utility class available to both internal and custom attributes.

If we are unsure, we can accept this as it has been implemented with the risk we won't remember to go back to the question.

CharliePoole:
@cPetru By the fixture part, do you mean TestFixture and TestFixtureSource?

ChrisMaddock:
Great!

> Should this apply to every attribute that generates test cases?

I definitely think so, on the principle of least surprise again. This is verging on implementation rather than feature - it concerns how NUnit resolves parameters. In my opinion, that should be consistent across all our attributes.

CharliePoole:
@ChrisMaddock That makes sense. As this decision has a lot of ramifications, let's wait till @rprouse gets up. :-) I'm going back to bed for a while.

rprouse:
OK, I've looked through the code and understand this issue now, another reason to start with an issue and some example code 😄 

I am also convinced that this is a good feature. I haven't run into this particular issue before, but I have refactored TestCases to TestCaseSources on several occasions and would have expected them to behave the same.

Assuming we accept this, which it looks like we will, I would also like to see the same done for `TestFixtureSource`. I would prefer to see that as a separate issue and PR though.

After reviewing the code, I don't have any comments that @CharliePoole has not made.

Since there is not issue, I am going to label this PR and assign it to the next milestone so that it gets included in the release notes.

@cPetru for building on Linux, `build.sh` should work. What errors are you seeing? Which Linux distro and what version of Mono?

cPetru:
@rprouse I'm using a virtual machine with Win 7, downloaded from modern.ie, on which I installed latest VS comunity. I just tried again with the latest nunit version.

builderror

CharliePoole:
Are you running on a 32 bit machine? If so use the -Experimental (abbr: -e) option on that system. You can also `set CAKE_ARGS=-Experimental` on that machine and our build.cmd will take care of it.

rprouse:
@cPetru any update on this? Do you need any help?

ChrisMaddock:
This will fix #1883. :+1:

rprouse:
@cPetru are you going to continue work on this? Your build errors may be because you are not using the Visual Studio command prompt.

CharliePoole:
@rprouse If we close this, let's create an issue for it.

rprouse:
@CharliePoole one of us could also finish this up. I think it is won't take much.

CharliePoole:
I'm working on another one right now, but I can take a look afterwards if you don't get to it.

rprouse:
I'm concentrating on the release right now. This one can wait.

ChrisMaddock:
I'm removing this PR from 3.7 milestone.

I'm hoping the work @Yarmonov is looking at in https://github.com/nunit/nunit/pull/2104 will incorporate this.

rprouse:
@nunit/framework-team does anyone want to pick up and finish this PR? If so, please claim it.

StanEgo:
@rprouse support for params, optional args and many other related stuff is already implemented in #2104. Just some tests updates required caused by a big amount of changes.

PR #1855, opened 2016-10-19
#1674 Using InRange with objects and no IComparable

### LINKED ISSUE #1674: InRange-Constraint must work with object
The following code does work even if myObj is from a class no special interface is implemented:

``` C#
Assert.That(myObj, Is.EqualTo(expcted).Using(myComparer));
Assert.That(myObj, Is.InRange(expcted1, expcted2).Using(myComparer));
```

There must be a inheritance of IComparable, or it does not compile.  (NUnit 3.4.1)
This is an error, because it must also work with objects.
- the type of myObj and expected1 I cannot change, because it is a external type from a library
- a comparer is supported, so it should work
- the constructor of InRangeConstraint calls the Compare-Methode of IComparable of expected1 not regarding the comparer in the Using to check the relation between from and to. This is a big bug in my eyes, using the wrong comparer.

JustinRChou:
Added constructor for objects in rangeconstraint

Updated Is and Not for new constructor.

Added unit test for case when class did not implement IComparable but
provides a IComparer through Using.

Fixes #1674


jplebre:
The tests only include certain platforms, any immediate reason for this?
There's an issue where this change has been discussed?

otherwise looks okay, trying to understand context

src/NUnitFramework/tests/Constraints/RangeConstraintTests.cs
```diff
@@ -87,5 +87,77 @@ public void ShouldThrowExceptionIfFromIsLessThanTo()
         {
             Assert.Throws<ArgumentException>(() => new RangeConstraint( 42, 5 ));
         }
+
+        #if !NETCF && !SILVERLIGHT  && !PORTABLE
```
rprouse: Why is this define needed? RangeConstraint works on all platforms, so it should be tested on all platforms.

rprouse:
@JustinRChou we usually need more context in order to review this issue. I assume it is for #1674. Is it a full fix, or partial?

As an initial pass, the code looks good, but I would like to look at it more closely after reading the issue.

src/NUnitFramework/tests/Constraints/RangeConstraintTests.cs
JustinRChou: There was a build issue with Silverlight, CF and portable not having Systems.Collections as the issue was dealing with classes that did not implement IComparable or ICompare. So it was affecting my wrapper class for the unit test.

I know that there are plans to completely remove these tags which those are being spun off into different projects?

JustinRChou:
From the looks of the issue, I would say full.

I think my comment just muddled things.

src/NUnitFramework/tests/Constraints/RangeConstraintTests.cs
rprouse: Can you remove the NETCF and SILVERLIGHT conditionals, they have been removed from the main solution now.

rprouse:
@JustinRChou I was just checking back through the PRs. You pushed a bunch of changes. Is this ready for review again? If so, in the future, add a comment asking for a second review. We have many pull requests across several repositories and tend to come back when we are notified via comments.

JustinRChou:
Should be good for review.

rprouse:
@nunit/core-team can we get a second review on this? I think it is ready to go.

@JustinRChou you don't need to @rprouse to get my attention, but it helps. The core team gets email notifications for all changes even when not mentioned.

ChrisMaddock:
Hi both, sorry I haven't got to this earlier.

Am I missing a subtly as to why we need to keep the old `IComparable` constructors as well as the `object` constructors in `ConstraintExpression`, `RangeExpression` and `Is`? Can they not just be replaced with the more general constructor?

Otherwise, this looks good. 😄

JustinRChou:
@ChrisMaddock 

1. That is a good point since all the methods are working with objects.
Tried your suggestion and 2 tests were failing.
ShouldThrowExceptionIfFromIsLessThanTo() which is to be expected.
I don't know why my new test NotInRangeTest() now throws ArgumentException("from must be less than to").

Going to check on the reason later.

JustinRChou:
Everything seems fine to remove
`public RangeConstraint(IComparable from, IComparable to)`
And move the from to check into `Usings `and `ApplyTo`

CharliePoole:
What happened here? Are we just closing this? Are we closing the issue as well?

JustinRChou:
I am cleaning it up a bit.
There was a NETSTANDARD issue with my test comparer.

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -56,6 +56,18 @@ public RangeConstraint(IComparable from, IComparable to) : base( from, to )
         }
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="RangeConstraint"/> class for objects not implementing IComparable. A class implementing IComparable must be passed through Using for a successful test.
+        /// </summary>
+        /// <remarks>A class implementing IComparable must be passed through Using for a successful test. from must be less than or equal to true</remarks> 
+        /// <param name="from">Inclusive beginning of the range. Must be less than or equal to to.</param>
+        /// <param name="to">Inclusive end of the range. Must be greater than or equal to from.</param>
+        public RangeConstraint(object from, object to) : base(from, to)
```
CharliePoole: The remarks don't seem to make sense - do you mean "A class NOT implementing IComparable...?"

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -56,6 +56,18 @@ public RangeConstraint(IComparable from, IComparable to) : base( from, to )
...
+        {
+            this.from = from;
+            this.to = to;
+        }
+
```
CharliePoole: It seems that this constructor doesn't actually create a usable RangeConstraint, since a Using must be provided. Two options seem possible: 1) Add the comparer to the constructor or 2) construct some other type, which implements Using and returns a range constraint when Using is provided.

JustinRChou: I would be fine with 1 which would also apply to the `Not.InRange` issue. 

I don't quite get 2, that it mean that it would be a new type of constraint? That seems unnecessary unless there's other constraints that have the same problem with `IComparable`?

CharliePoole:
Option 2 is a technique we use a lot in fluent interfaces. So...

`Is.InRange(object, object)` would return some non-constraint class, e.g.: `RangeProtoConstraint`

Since `RangeProtoConstraint` is not a constraint. an assertion the following would not complile.

```C#
Assert.That(val, Is.InRange(obj1, obj2));
```

Give `RangeProtoConstraint` a property Using, exactly like what is defined for `RangeConstraint`. Those properties would return a `RangeConstraint` so the following would compile...

```C#
Assert.That(val, Is.InRange(obj1, obj2).Using(mycomparer));
```

Depending on whether there are other modifying properties on RangeConstraint, some inheritance could be used here. I didn't get down to that level of detail in looking at it. It's easiest, however, if the intermediate class is not a constraint at all.

However, as cool as that is, I'm not sure it's worth the effort. Defining a constructor and syntax helpers that take `(object, object, IComparer)` may be sufficient.

JustinRChou:
I might be overthinking it but should `ApplyTo` perform another range check if a custom comparer is set through `Using`?

ChrisMaddock:
I think the checks should _only_ be in `ApplyTo`. What if I did `Is.InRange(big, small).Using(InverseComparer)`. That would currently throw in the InRange constructor, although the statement is valid. 

Point being, the comparer can be changed after the RangeConstraint() constructor is wrong, so this seems the wrong place to check anything. (This is very much edge case, but something like the above might be a good test case!)

ChrisMaddock:
I think the last change needed here is to move the range check out of the `RangeConstraint()` constructor, and into ApplyTo<T> - for the reasons above. That will affect some test cases, but other than that, I think this is good to merge. 

@nunit/framework-team - would someone give this a second look over, given there's been a bit of back and forth on it already, and I don't want to have missed anything?

@JustinRChou - are you in a position to finish this off? If not, someone else should, or we can merge as is and create a separate issue for the range check. It would be good to get it merged in, either way.

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -949,11 +949,27 @@ public SamePathOrUnderConstraint SamePathOrUnder(string expected)
         /// Returns a constraint that tests whether the actual value falls 
         /// within a specified range.
         /// </summary>
+        /// <remarks>from must be less than or equal to true</remarks> 
+        /// <param name="from">Inclusive beginning of the range. Must be less than or equal to to.</param>
```
jnm2: How about `Must be less than or equal to <paramref name="to"/>.`?

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -949,11 +949,27 @@ public SamePathOrUnderConstraint SamePathOrUnder(string expected)
...
+        /// <param name="to">Inclusive end of the range. Must be greater than or equal to from.</param>
```
jnm2: How about `Must be greater than or equal to <paramref name="from"/>.`?

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -949,11 +949,27 @@ public SamePathOrUnderConstraint SamePathOrUnder(string expected)
...
         {
             return (RangeConstraint)this.Append(new RangeConstraint(from, to));
         }
 
+        /// <summary>
+        /// Returns a constraint that tests whether the actual value falls 
+        /// within a specified range.
+        /// </summary>
+        /// <remarks>from must be less than or equal to true</remarks> 
+        /// <param name="from">Inclusive beginning of the range. Must be less than or equal to to.</param>
```
jnm2: Same [How about Must be less than or equal to <paramref name="to"/>.?]

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -949,11 +949,27 @@ public SamePathOrUnderConstraint SamePathOrUnder(string expected)
...
+        /// <param name="to">Inclusive end of the range. Must be greater than or equal to from.</param>
```
jnm2: Same [How about `Must be greater than or equal to <paramref name="from"/>.`?]

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -56,6 +56,24 @@ public RangeConstraint(IComparable from, IComparable to) : base( from, to )
         }
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="RangeConstraint"/> class for objects that did not implement <see cref="IComparable"/>.
+        /// </summary>
+        /// <remarks>from must be less than or equal to true</remarks> 
+        /// <param name="from">Inclusive beginning of the range. Must be less than or equal to to.</param>
```
jnm2: Same [How about Must be less than or equal to <paramref name="to"/>.?]

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -56,6 +56,24 @@ public RangeConstraint(IComparable from, IComparable to) : base( from, to )
...
+        /// <param name="to">Inclusive end of the range. Must be greater than or equal to from.</param>
```
jnm2: Same [How about Must be less than or equal to <paramref name="from"/>.?]

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -56,6 +56,24 @@ public RangeConstraint(IComparable from, IComparable to) : base( from, to )
...
+        /// <param name="comparer">Class that implements <seealso cref="IComparer"/> used to compare the objects.</param>
+        public RangeConstraint(object from, object to, IComparer comparer) : base(from, to)
+        {
+            // from must be less than or equal to to
```
jnm2: Not sure why this comment is needed, but it contradicts the exception message just below.

src/NUnitFramework/framework/Is.cs
```diff
@@ -543,6 +543,20 @@ public static RangeConstraint InRange(IComparable from, IComparable to)
             return new RangeConstraint(from, to);
         }
 
+        /// <summary>
+        /// Returns a constraint that tests whether the actual value falls
+        /// inclusively within a specified range.
+        /// </summary>
+        /// <remarks>from must be less than or equal to true</remarks> 
+        /// <param name="from">Inclusive beginning of the range. Must be less than or equal to to.</param>
+        /// <param name="to">Inclusive end of the range. Must be greater than or equal to from.</param>
```
jnm2: Same [How about Must be less than or equal to <paramref name="from"/>.?]

src/NUnitFramework/framework/Is.cs
```diff
@@ -543,6 +543,20 @@ public static RangeConstraint InRange(IComparable from, IComparable to)
             return new RangeConstraint(from, to);
         }
 
+        /// <summary>
+        /// Returns a constraint that tests whether the actual value falls
+        /// inclusively within a specified range.
+        /// </summary>
+        /// <remarks>from must be less than or equal to true</remarks> 
+        /// <param name="from">Inclusive beginning of the range. Must be less than or equal to to.</param>
```
jnm2: Same [How about Must be less than or equal to <paramref name="to"/>.?]

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -949,11 +949,27 @@ public SamePathOrUnderConstraint SamePathOrUnder(string expected)
         /// Returns a constraint that tests whether the actual value falls 
         /// within a specified range.
         /// </summary>
+        /// <remarks>from must be less than or equal to true</remarks> 
+        /// <param name="from">Inclusive beginning of the range. Must be less than or equal to to.</param>
+        /// <param name="to">Inclusive end of the range. Must be greater than or equal to from.</param>
         public RangeConstraint InRange(IComparable from, IComparable to)
```
jnm2: Why can't these two methods be `(object from, object to, IComparer comparer = null)` with `comparer` defaulting to `Comparer<object>.Default`? That's more what I expect from a .NET library.

src/NUnitFramework/tests/TestUtilities/Comparers/ObjectToStringComparer.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Collections;
+using System.Collections.Generic;
+
+namespace NUnit.TestUtilities.Comparers
+{
+    /// <summary>
+    /// ObjectToStringComparer is used in testing the <see cref="NUnit.Framework.Constraints.RangeConstraint"/> when the object does not implement the <see cref="IComparer"/>  interface.
+    /// Compares them as numbers when both arguments are <see cref="int"/>, else it uses <seealso cref="string.CompareTo(string)"/>.
+    /// </summary>
+    public class ObjectToStringComparer : System.Collections.IComparer
```
jnm2: I did not see discussion on this, but what was wrong with using `Comparer<object>.Default` like every other .NET library?
My expectation is that people should be comparing objects of the same type via a custom comparer or `Comparer<>.Default`. If they don't, they'll run into the same trouble everywhere. That's not unique to NUnit.

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -949,11 +949,27 @@ public SamePathOrUnderConstraint SamePathOrUnder(string expected)
         /// Returns a constraint that tests whether the actual value falls 
         /// within a specified range.
         /// </summary>
+        /// <remarks>from must be less than or equal to true</remarks>
```
ChrisMaddock: Should this be `**F**rom must be less than or equal to **to**`

jnm2: Good catch. I would say either `<paramref name="from"/>` or something else entirely because it's awkward English.

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
ChrisMaddock: Comparer already defaults to `ComparisonAdapter.Default` down in RangeConstraint. But I think that's the NUnit-y equivalent of `Comparer<object>.Default`.

The optional parameter is a good point - although I have a vague memory that this signature was added to prevent users using the constraint to compare non-IComparables by default comparer? I don't see that that's strictly necessary though...maybe @rprouse or @CharliePoole who did the earlier reviews can remember the context better?

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
jnm2: What about "The <paramref name="from"/> value must be less than or equal to the <paramref name="to"/> value.`?

jnm2: What about `The <paramref name="from"/> value must be less than or equal to the <paramref name="to"/> value.`?

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
CharliePoole: The optional parameter would be OK if every .NET language had optional parameters, which may not be true. More to the point, injecting the comparer via a modifier property is how we do it in order to support the fluent syntax.

src/NUnitFramework/tests/TestUtilities/Comparers/ObjectToStringComparer.cs
CharliePoole: This is a test utility, not a library for users. The point is to guarantee that the test object implements __only__ the desired interface. So, for example, if we are testing non-generic IComparer, we want to be sure that no generic is implemented. At least that's what the other comparers are for and I assume this one is the same.

Remember, when we wrote this in the first place, generics didn't even exist.

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
jnm2: If optional params are not an option :D, this signature should still be changed to `(object, object)` to allow the fluent syntax.

src/NUnitFramework/tests/TestUtilities/Comparers/ObjectToStringComparer.cs
jnm2: I am not seeing this guarantee in the code. I'm sorry I'm not following yet! Why would you want to be sure that no generic is implemented?

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
rprouse: I second the `<paramref />`

src/NUnitFramework/framework/Is.cs
jnm2: [Following up](https://github.com/nunit/nunit/pull/1855#discussion_r115353617)
What about `Must be greater than or equal to the <paramref name="from"/> value.`?

src/NUnitFramework/framework/Is.cs
jnm2: [Following up](https://github.com/nunit/nunit/pull/1855#discussion_r115353617)
What about `Must be less than or equal to the <paramref name="to"/> value.`?

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
jnm2: [Following up](https://github.com/nunit/nunit/pull/1855#discussion_r115353617)
What about `Must be less than or equal to the <paramref name="to"/> value.`?

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
jnm2: [Following up](https://github.com/nunit/nunit/pull/1855#discussion_r115353617)
What about `Must be greater than or equal to the <paramref name="from"/> value.`?

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
jnm2: [Following up](https://github.com/nunit/nunit/pull/1855#discussion_r115353617)
What about `Must be greater than or equal to the <paramref name="from"/> value.`?

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
jnm2: [Following up](https://github.com/nunit/nunit/pull/1855#discussion_r115353617)
What about `Must be less than or equal to the <paramref name="to"/> value.`?

JustinRChou:
From what I am reading, 
1. Move range check to apply to.
2. Update the remark to something different?

JustinRChou:
Kept the range check in the constructor for one test.

src/NUnitFramework/framework/Is.cs
```diff
@@ -534,15 +534,29 @@ public static SamePathOrUnderConstraint SamePathOrUnder(string expected)
...
         /// inclusively within a specified range.
         /// </summary>
-        /// <remarks>from must be less than or equal to true</remarks> 
-        /// <param name="from">Inclusive beginning of the range. Must be less than or equal to to.</param>
-        /// <param name="to">Inclusive end of the range. Must be greater than or equal to from.</param>
+        /// <remarks>The <paramref name="from"/> value must be less than or equal to the <paramref name="to"/> value.</remarks> 
+        /// <param name="from">Must be less than or equal to the <paramref name="to"/> value.</param>
+        /// <param name="to">Must be greater than or equal to the <paramref name="from"/> value.</param>
         /// <returns></returns>
         public static RangeConstraint InRange(IComparable from, IComparable to)
```
jnm2: Did I misunderstand, or should this signature be `(object, object)` so that you can use the fluent syntax to specify a comparer? A test for `.InRange(notIComparable, notIComparable).Using(customComparer)` would be nice. /cc @CharliePoole

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -33,26 +33,41 @@ namespace NUnit.Framework.Constraints
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RangeConstraint"/> class for objects that did not implement <see cref="IComparable"/>.
+        /// </summary>
+        /// <remarks>The <paramref name="from"/> value must be less than or equal to the <paramref name="to"/> value.</remarks> 
+        /// <param name="from">Must be less than or equal to the <paramref name="to"/> value.</param>
+        /// <param name="to">Must be greater than or equal to the <paramref name="from"/> value.</param>
+        /// <param name="comparer">Class that implements <seealso cref="IComparer"/> used to compare the objects.</param>
+        public RangeConstraint(object from, object to, IComparer comparer) : base(from, to)
+        {
+            // from must be less than or equal to to
```
jnm2: Nit: redundant? Repeated 3 times a couple lines above.

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -33,26 +33,41 @@ namespace NUnit.Framework.Constraints
...
-        /// <remarks>from must be less than or equal to true</remarks> 
-        /// <param name="from">Inclusive beginning of the range. Must be less than or equal to to.</param>
-        /// <param name="to">Inclusive end of the range. Must be greater than or equal to from.</param>
+        /// <remarks>The <paramref name="from"/> value must be less than or equal to the <paramref name="to"/> value.</remarks> 
+        /// <param name="from">Must be less than or equal to the <paramref name="to"/> value.</param>
+        /// <param name="to">Must be greater than or equal to the <paramref name="from"/> value.</param>
         public RangeConstraint(IComparable from, IComparable to) : base( from, to )
         {
             // Issue #21 - https://github.com/nunit/nunit-framework/issues/21
             // from must be less than or equal to to
```
jnm2: Nit: redundant? Repeated 3 times a couple lines above. And do we need the issue link in the source code since there won't be anything pending from that issue once this PR is merged?

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -72,7 +87,7 @@ public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
             if ( from == null || to == null || actual == null)
                 throw new ArgumentException( "Cannot compare using a null reference", "actual" );
-
+            CompareFromAndTo();
```
jnm2: I'd be happier if this was checked when `comparer` is changed in the `Using` methods because that catches possible bugs immediately rather than deferring until this part of the constraint is evaluated- which may be never.
That also guarantees that this check can be removed.

src/NUnitFramework/tests/TestUtilities/Comparers/ObjectToStringComparer.cs
jnm2: Oh wow, it took me way too long to realize that this is a test utility as in, \*not used by the framework\*. I'm sorry!

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -72,7 +87,7 @@ public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
             if ( from == null || to == null || actual == null)
                 throw new ArgumentException( "Cannot compare using a null reference", "actual" );
```
jnm2: The `IComparer` interface is designed to work with null references, so I strongly expect it to work here. I'm positive I will hit this in real world code.

jnm2: Ideally we'd have tests verifying that `Is.InRange(null, null).ApplyTo(null)` does not throw an exception.

jnm2: (See http://stackoverflow.com/a/23787253)

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
JustinRChou: Not sure how xml docs work, so it's able to pull the remarks and param docs from the another constructor?

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
ChrisMaddock: This is a fair comment, although potentially a separate pull request, depending how much further work @JustinRChou wants to do on this. 🙂 I'm just very aware this PR has already grown significantly from the original #1674, and there's been a lot of back and forth - it might be good to get this one merged and look at the above separately!

I'll leave that decision with @JustinRChou 😄

JustinRChou: I don't mind if it is related to the issue.
But wouldn't this change also mean that any range assert with a null and a non-null would be valid now?

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
jnm2: No I'm sorry, I meant the code comments (lines 49 and 50).

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
jnm2: I totally overlooked that it was preexisting code. @JustinRChou unless you really feel like doing it now, why don't we do a separate PR after this merges?

jnm2: Yes, any range assert with a null and a non-null would be valid.

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -949,11 +949,27 @@ public SamePathOrUnderConstraint SamePathOrUnder(string expected)
...
+        /// <summary>
+        /// Returns a constraint that tests whether the actual value falls 
+        /// within a specified range.
+        /// </summary>
+        /// <remarks>The <paramref name="from"/> value must be less than or equal to the <paramref name="to"/> value.</remarks> 
+        /// <param name="from">Must be less than or equal to the <paramref name="to"/> value.</param>
+        /// <param name="to">Must be greater than or equal to the <paramref name="from"/> value.</param>
+        /// <param name="comparer">Class that implements <seealso cref="IComparer"/> used to compare the objects.</param>
+        public RangeConstraint InRange(object from, object to, IComparer comparer)
+        {
```
CharliePoole: I don't think we want this syntax. It doesn't match our usual.

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -949,11 +949,27 @@ public SamePathOrUnderConstraint SamePathOrUnder(string expected)
         /// Returns a constraint that tests whether the actual value falls 
         /// within a specified range.
         /// </summary>
+        /// <remarks>The <paramref name="from"/> value must be less than or equal to the <paramref name="to"/> value.</remarks> 
+        /// <param name="from">Must be less than or equal to the <paramref name="to"/> value.</param>
+        /// <param name="to">Must be greater than or equal to the <paramref name="from"/> value.</param>
```
CharliePoole: Is there a good use for Remarks when it says basically the same as what the params say? Where do they show up?

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
CharliePoole: That's not something we do unless it requires reading the issue to understand the code. Which it shouldn't. And doesn't here IMO.

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -33,26 +33,41 @@ namespace NUnit.Framework.Constraints
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RangeConstraint"/> class for objects that did not implement <see cref="IComparable"/>.
+        /// </summary>
+        /// <remarks>The <paramref name="from"/> value must be less than or equal to the <paramref name="to"/> value.</remarks> 
+        /// <param name="from">Must be less than or equal to the <paramref name="to"/> value.</param>
+        /// <param name="to">Must be greater than or equal to the <paramref name="from"/> value.</param>
+        /// <param name="comparer">Class that implements <seealso cref="IComparer"/> used to compare the objects.</param>
+        public RangeConstraint(object from, object to, IComparer comparer) : base(from, to)
+        {
```
CharliePoole: I'd rather not see this constructor. The same can be done by `new RangeConstraint(from, to).Using(comparer)`

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
CharliePoole: Yes... keep separate. I think it's very debatable whether a range like [null, 5] makes any sense at all.

src/NUnitFramework/framework/Is.cs
CharliePoole: Use of IComparable was an old old experiment at making constraints more type-safe. Obviously, this issue shows that the approach doesn't work when we allow arbitrary comparers to be used. So I'd go back to two-object constructor and syntax elements in all cases.

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
JustinRChou: Trying to remember why I added that constructor.
I think it was because of the range check that occurs in the constructor which would have crashed.

CharliePoole: Actually, it would make a lot of sense if the normal way to use it were through the constructor. But for us the constructors are there but not primary. Construction is generally through the fluent syntax.

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
jnm2: @CharliePoole it's the same as Is.LessThanOrEqualTo(5) but it's the more general form. Consider for instance having pairs of values supplied by `TestCase` attributes- no sense having to rewrite your code to use InRange or LessThanOrEqualTo, etc. The endpoints could also be calculated from another source.

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
JustinRChou: It appears to only show up in the xml documentation and not through the intellisense.

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
CharliePoole: @jnm2 I very strongly disagree. I understand what you mean and that it can be __thought__ of in that way, but a range is a concept, separate from any mathematical embodiment, which doesn't seem to me to have a place in it for null values. Anyway, it's good that we have pushed off the question to another discussion.

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
JustinRChou: In my test, if I remove the check in ApplyTo it has the following failure
```
            Assert.Throws<ArgumentException>(() => test.Using(rComparer));
            Assert.Throws<ArgumentException>(() => test.Using(rComparer).ApplyTo(7));
            Assert.Throws<ArgumentException>(() => test.ApplyTo(7)); //fails now
```
But I suppose in a normal use case, the failed line would never have a chance to occur?

jnm2: I would expect exactly that, because the comparer is never set. IOW you can never put the constraint into an invalid state, consistent with the constructor check.
```c#
Assert.DoesNotThrow(()=> test.ApplyTo(7));
Assert.Throws<ArgumentException>(() => test.Using(rComparer));
Assert.DoesNotThrow(()=> test.ApplyTo(7));
```

CharliePoole:
@JustinRChou in general, do not push a commit to revert a change in a pr.  Such a commit is needed when something has already been published to master, but when using a feature branch, we allow changing history prior to merge. Just do a local reset, make your corrections and push -f. A comment is polite whenever you do that.

With the extra commits, this now has to be squashed.

JustinRChou:
Should I continue to reset my commits to a single one?

CharliePoole:
No need with a normal sequence of commits, but if you go back it's best to just go back.

Naturally, we don't do that with master because we consider master as "published".

ChrisMaddock:
@CharliePoole - could you take a second look, if @JustinRChou's latest commit covers your concerns? I'm struggling to follow the timeline a little on this one.

I'd like to get this merged in for 3.7 (potentially this weekend), if possible.

src/NUnitFramework/framework/Is.cs
```diff
@@ -534,15 +534,29 @@ public static SamePathOrUnderConstraint SamePathOrUnder(string expected)
...
+        /// <summary>
+        /// Returns a constraint that tests whether the actual value falls
+        /// inclusively within a specified range.
+        /// </summary>
+        /// <remarks>The <paramref name="from"/> value must be less than or equal to the <paramref name="to"/> value.</remarks> 
+        /// <param name="from">Must be less than or equal to the <paramref name="to"/> value.</param>
+        /// <param name="to">Must be greater than or equal to the <paramref name="from"/> value.</param>
+        /// <param name="comparer">Class that implements <seealso cref="IComparer"/> used to compare the objects.</param>
+        /// <returns></returns>
+        public static RangeConstraint InRange(object from, object to, IComparer comparer)
```
jnm2: I'm assuming removing this goes along with removing the `(object, object, IComparer)` constructor. /cc @CharliePoole @rprouse

src/NUnitFramework/framework/Is.cs
jnm2: I'd still like to see a test calling `.InRange(notIComparable, notIComparable).Using(customComparer)` just to lock that API in, but skip it if it blocks merging. /cc @rprouse

src/NUnitFramework/framework/Is.cs
```diff
@@ -534,15 +534,27 @@ public static SamePathOrUnderConstraint SamePathOrUnder(string expected)
         /// Returns a constraint that tests whether the actual value falls
         /// inclusively within a specified range.
         /// </summary>
-        /// <remarks>from must be less than or equal to true</remarks> 
-        /// <param name="from">Inclusive beginning of the range. Must be less than or equal to to.</param>
-        /// <param name="to">Inclusive end of the range. Must be greater than or equal to from.</param>
+        /// <param name="from">Must be less than or equal to the <paramref name="to"/> value.</param>
+        /// <param name="to">Must be greater than or equal to the <paramref name="from"/> value.</param>
         /// <returns></returns>
         public static RangeConstraint InRange(IComparable from, IComparable to)
```
jnm2: This must be (object, object). An `Is.InRange(notIComparable, notIComparable).Using(customComparer)` test would force this to be done. /cc @rprouse

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -33,26 +33,35 @@ namespace NUnit.Framework.Constraints
...
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RangeConstraint"/> class.
         /// </summary>
-        /// <remarks>from must be less than or equal to true</remarks> 
-        /// <param name="from">Inclusive beginning of the range. Must be less than or equal to to.</param>
-        /// <param name="to">Inclusive end of the range. Must be greater than or equal to from.</param>
+        /// <param name="from">Must be less than or equal to the <paramref name="to"/> value.</param>
+        /// <param name="to">Must be greater than or equal to the <paramref name="from"/> value.</param>
         public RangeConstraint(IComparable from, IComparable to) : base( from, to )
```
jnm2: This must be (object, object). An `Is.InRange(notIComparable, notIComparable).Using(customComparer)` test would force this to be done. /cc @rprouse

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -949,11 +949,27 @@ public SamePathOrUnderConstraint SamePathOrUnder(string expected)
...
         public RangeConstraint InRange(IComparable from, IComparable to)
```
jnm2: This must be (object, object). A non-`Is` `.InRange(notIComparable, notIComparable).Using(customComparer)` test would force this to be done. /cc @rprouse

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -33,26 +33,35 @@ namespace NUnit.Framework.Constraints
...
         {
-            // Issue #21 - https://github.com/nunit/nunit-framework/issues/21
-            // from must be less than or equal to to
-            if (comparer.Compare(from, to) > 0)
-                throw new ArgumentException( "from must be less than to" );
+            this.from = from;
+            this.to = to;
+            CompareFromAndTo();
```
jnm2: You would not believe how much I hate to bring this up.
If we change this constructor from (IComparable, IComparable) to (object, object), we will have to remove this `CompareFromAndTo()` and add it to `Apply` instead. The reason is that if the objects are not comparable, this constructor will throw: [NUnitComparer.cs#L81](https://github.com/JustinRChou/nunit/blob/1f97672f5c0c4ad4093ba48bad68af0523dab828/src/NUnitFramework/framework/Constraints/NUnitComparer.cs#L81)

We can't `CompareFromAndTo()` until we have a comparer, or else .InRange(notComparable, notComparable).Using(customComparer)` will throw at the `.InRange(notComparable, notComparable)` part.

@CharliePoole should this change our minds about providing only (IComparable, IComparable) and (object, object, IComparer)?


That test I wanted awhile back would have poked us in the eye with this.

/cc @CharliePoole @rprouse

rprouse:
@nunit/core-team this pull request is the last open item in the 3.7 release. We have not heard back from @JustinRChou, so unless someone takes this over and cleans it up, I will drop it from the release.

For release planning, I am hoping to do it either tomorrow evening or Sunday, so this is up to the wire.

jnm2:
@nunit/core-team I think we still need a [design discussion](https://github.com/nunit/nunit/pull/1855#discussion_r118140973) :-(

JustinRChou:
So dropping all the other constructors and just go with 
`RangeConstraint(object, object)` ?

CharliePoole:
I think we should hold for after the release. My own comments do not seem to be addressed yet.

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
JustinRChou: @jnm2 
Would there be anything wrong to check the type first?
```
        public RangeConstraint(object from, object to) : base(from, to)
        {
            this.from = from;
            this.to = to;
            if (this.from is IComparable && this.from is IComparable) {
                CompareFromAndTo();
            }
        }
```

JustinRChou:
@CharliePoole 
Could you repost your comments that need addressing?

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
jnm2: @JustinRChou That solves one case where one or the other is not IComparable, but there's a further case which that does not solve: two object that *do* implement IComparable, but where you have a custom comparer with completely different logic such that `from.CompareTo(to) > 0` but `customComparer.Compare(from, to) <= 0`. (That should be a test case too.)

It's not very far-fetched. I have run into and written types that implement IComparable just for the purpose of delegating to comparing one of many properties. I have also used and exposed custom comparers in business logic that do similar things.

jnm2: What happens in that scenario is that you are blocked from ever creating a range with those objects because they only way you possibly could is to 1) not check that `from < to` for *any* types or 2) pass the custom comparer in up front.

JustinRChou: I see what you mean about the new test now.

rprouse:
@JustinRChou you can review open comments by going to the files view at https://github.com/nunit/nunit/pull/1855/files

Are you still interested in cleaning this up, or should we close it?

jnm2:
I can finish it if @JustinRChou is not able.

JustinRChou:
So it's just removing the 
`public RangeConstraint(object from, object to, IComparer comparer)` 
and 
`public static RangeConstraint InRange(object from, object to, IComparer comparer)`
?

jnm2:
Oy. @nunit/framework-team It would be cool if we could respond in fewer than 26 days. I'm the biggest one to blame, I will improve. 

Back in May I asked if we could discuss a design problem that is preventing @JustinRChou from moving forward. Please respond: https://github.com/nunit/nunit/pull/1855#discussion_r118140973

jnm2:
My recommendation is to only have constructors `(object, object)` and `(object, object, IComparer)`. The latter is needed because the `(object, object)` constructor should throw if `left > right` before the `.Using(reverseComparer)` even gets a chance to execute. It's cleaner than deferring argument error checking to the last possible moment.

CharliePoole:
Sorry, I don't see the design problem. @JustinRChou Can you explain what feedback you are waiting for?

JustinRChou:
@jnm2 I am going to admit that this PR has so many comments that I was very confused about what are the changes that needed to be done.
@CharliePoole I think I was asking about removing
`public RangeConstraint(IComparable from, IComparable to)`
and just deal with 
`public RangeConstraint(object from, object to)`
since the `from` and `to` are objects now.
```
        private readonly object from;
        private readonly object to;
```

jnm2:
@CharliePoole You said you didn't want `(object, object, IComparer)` because we usually do `(object, object)` and `.Using(IComparer)` instead. I think we should reconsider and keep `(object, object, IComparer)` because otherwise we will have to remove argument error checking from the constructor.

CharliePoole:
@jnm2 What syntax would the user use that would cause the three arg constructor to be used? I think that's what we have to start from.

CharliePoole:
@JustinRChou Yes, that makes sense to me. You would have to check the types in order to compare them.

CharliePoole:
@JustinRChou Obviously, all existing calls with IComparables have to continue to work with the object constructor.

jnm2:
@CharliePoole Yes, good point. We need something like `Is.InRange(from, to, comparer)` in case the default comparer would consider `from` to be greater than `to`.

CharliePoole:
@jnm2 Except that syntax is ugly! 😸 Not to mention inconsistent with most of our other syntax. 

A "range" has two things the start and end values, it doesn't have a comparer in normal parlance.

I'd rather see `Is.InRange(from, to).Using(comparer)` keep working. If that means removing the test from the constructor and doing it elsewhere, that's just a detail for @JustinRChou to figure out.

jnm2:
@CharliePoole We could still have both syntaxes. But we're accepting objects that aren't inherently comparable to define a comparison-based range, so I'd say it would be expected that we would have an option to take a comparer.

jnm2:
The comparer could be comparing the objects on arbitrary criteria like a certain property value and the comparison could ignore case or not. The comparer could even sort in descending order.

CharliePoole:
We do essentially the same thing in comparison constraints, but any compared is specified in a modifier. I think consistent syntax is essential for a fluent interface. Making a comparison test in the constructor is dispensable.

ChrisMaddock:
I can't wait for this to be merged... sorry it's been such a long one, @JustinRChou! My opinions below:

I would prefer one of two things:

1. Two ctors. `RangeConstraint(IComparable, IComparable)` and `RangeConstraint(object, object, IComparable)` (This is what Justin currently has.)

or:

2. Just `RangeConstraint(object, object)`

I don't like the `(object, object)` and `(object, object, IComparer)` combination - the point of having multiple ctors (as I saw it!) was to add a little type-safety, and ensure the user can't compare to objects with no Comparer. If we're not having an (IComparable, IComparable) ctor, I think we should _just_ have the `(object, object).Using()` syntax instead of the two ctors.

In both cases, the `FromIsLessThatTo()` check should be done in the ApplyTo method, and NOT in the constructor. This is necessary in the case the user uses a `Using()` which inverts the comparer - i.e. the from/to args could fail the check under the defaultComparer, but pass under the user-specified comparer - which isn't known about in the constructor. For this reason, this check can't be done until the constraint is fully built.

I agree consistent syntax is the issue here - it's difficult, as @CharliePoole's suggestion of (2) is more consistent with NUnit, but @jnm2's suggestion (1 - pretty much) is more consistent with .NET as a whole. Can I suggest we go with NUnit-consistency for now (option 2) - and if we wanted to add `(object, object, IComparable)` methods at some point, this should be done as a whole, e.g. including EqualConstraint and other constraints?

In the interests of moving this on, here's a concrete proposition. @CharliePoole @jnm2 @JustinRChou - would be good to hear everyone's thoughts before Justin writes any more code!

1. RangeConstraint to have a single constructor - which takes (object, object). Everything upwards in ConstraintExpression.cs/Is.cs should be updated to reflect this.
2. `FromIsLessThanTo()` check to be moved out of the constructor, and happen only once the constraint has been fully built. (Which I assume means it should be done in the ApplyTo method.) Some of tests of NUnit will need to be changed to allow this.

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
ChrisMaddock: @jnm2 - if you're still interested in this - would you mind adding a separate issue? Don't want this to be lost in amongst this PR - I think it's worth saving, but let's get this one merged!

ChrisMaddock:
I've added this to 3.8 - let's aim to get it resolved and merged for then! 👍

CharliePoole:
@ChrisMaddock I agree entirely with your suggestion. As far as I understand things, it's also what @JustinRChou was in the process of doing before this became an item for discussion.

As far as consistency with other .NET usage, I'd say that's a non-issue in the case of constructors that are not really used directly by users, or at least not that frequently. If you're using constraints, then you are most likely doing it in order to use the fluent syntax.

jnm2:
I'm not opposed to that suggestion. :+1:

rprouse:
Great writeup of what needs to be done @ChrisMaddock. Let's stick with that and get this one out of the way. @JustinRChou do you think we can wrap this up this week?

JustinRChou:
Ok. I will see if I could update this for tomorrow and we could discuss changes from there.

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -961,12 +961,14 @@ public SamePathOrUnderConstraint SamePathOrUnder(string expected)
...
         #region InRange
-
         /// <summary>
         /// Returns a constraint that tests whether the actual value falls
         /// within a specified range.
         /// </summary>
-        public RangeConstraint InRange(IComparable from, IComparable to)
+        /// <remarks>The <paramref name="from"/> value must be less than or equal to the <paramref name="to"/> value.</remarks> 
+        /// <param name="from">Must be less than or equal to the <paramref name="to"/> value.</param>
+        /// <param name="to">Must be greater than or equal to the <paramref name="from"/> value.</param>
```
CharliePoole: These restrictions no longer apply, correct?

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -33,26 +33,24 @@ namespace NUnit.Framework.Constraints
...
 
         /// <summary>
         /// Initializes a new instance of the <see cref="RangeConstraint"/> class.
         /// </summary>
-        /// <remarks>from must be less than or equal to true</remarks> 
-        /// <param name="from">Inclusive beginning of the range. Must be less than or equal to to.</param>
-        /// <param name="to">Inclusive end of the range. Must be greater than or equal to from.</param>
-        public RangeConstraint(IComparable from, IComparable to) : base( from, to )
+        /// <param name="from">Must be less than or equal to the <paramref name="to"/> value.</param>
+        /// <param name="to">Must be greater than or equal to the <paramref name="from"/> value.</param>
```
CharliePoole: Same comment as above. Shouldn't these restrictions be moved elsewhere?

src/NUnitFramework/framework/Constraints/RangeConstraint.cs
```diff
@@ -33,26 +33,24 @@ namespace NUnit.Framework.Constraints
...
-            // Issue #21 - https://github.com/nunit/nunit-framework/issues/21
-            // from must be less than or equal to to
-            if (comparer.Compare(from, to) > 0)
-                throw new ArgumentException( "from must be less than to" );
-
             this.from = from;
             this.to = to;
+            if (from is IComparable && to is IComparable)
+            {
+                CompareFromAndTo();
```
CharliePoole: So, in the case where the two args are `IComparable`, you do the comparison early. What benefit does that give us over simply waiting until execution? Is it even accurate? What if a comparer is specified that reverses the test? I think I would feel more comfortable moving __all__ the testing to ApplyTo, when any special comparer has already been set.

ChrisMaddock: I agree, with this and your other two comments. This fits with what we agreed on above. 🙂

jnm2: Yes to all.

JustinRChou:
Not sure why CI is failing.
Builds fine on my machine.

CharliePoole:
It's a timing issue not related to your changes. I restarted the build on AppVeyor.

ChrisMaddock:
@JustinRChou - there's just one more of Charlie's comments you've missed in CoinstraintExpression.cs

https://github.com/nunit/nunit/pull/1855/files#r134530838

After that - I think this is ready to merge. 😄

ChrisMaddock:
Everything has now been addressed and CI has now passed - so merged!

Thanks for sticking with this one, everyone - especially @JustinRChou. I'm pretty happy we managed to pin down the right solution in the end! 😄

jnm2:
Yes, honestly thanks everyone for putting up with my nitpicking.

Issue #1771, opened 2016-08-27
Runtime parameter validation


oznetmaster:
Now that there is the ability to specify runtime parameters on the command line, it would be very useful to be able to validate them.

I am thinking of something like a UseParameterAttribute that could be applied to the assembly, namespace, TestFixture, or Test.  The attribute would have the parameter names, acceptable value type, and perhaps even a description.

When the test suite is initiated, all of the valid parameter names (and perhaps types) would be obtained from the test assemblies and accumulated in a dictionary.  Then, that dictionary would be used to validate the parameters entered on the command line prior to actually running any tests.

If a parameter is entered which is does not validate, either a warning could be given, or the test run could be inhibited.

Validation could be both for parameter name and for type.

Also, the explore command line option would not only list the tests, but that acceptable parameters.  This would allow a GUI to provide a list of parameters (and descriptions), and to only accept valid values for them.

oznetmaster:
This could even be carried further, and allow the parameters specified in the attribute to be passed to the test as method arguments.

```
[UseParameter ("mystring", typeof(string)]
public void Test (string mystring)
{
}
```

oznetmaster:
No comments at all? :(

CharliePoole:
Well, I've been keeping quiet to see if somebody came along to agree with it. I don't like it as a feature.

When we implemented run parameters, we decided to do it in a way consistent with how other frameworks, notably MsTest implement them. That is, it's up to the test code to access the parameters. They are all strings, which is what they start out as when entered either on the command-line or in a .runparameters file (for the adapter). It's up to the test code to verify that the string is actually an int.

I would support enhancements to the way we do it, perhaps making it easier to verify that the string passed can be converted, or even having some conversion embedded in the TestContext for the user to call. I can't support backing away from the earlier decision in favor of an attribute-based approach.

oznetmaster:
After I added support for the parameters to my GUI, I discovered my users were using them a lot.

Being able to have all available parameters displayed in the GUI, along with validation of values entered for them, was a significant enhancement to just have a place to enter them.

The only way I could see to easily do that was with attribute based definition of the parameters used in each test, if any.

I implemented it that way, and it works very well.

CharliePoole:
I'll continue to wait for others to chime in. :-)

oznetmaster:
How do you propose to handle parameters in the NUnit GUI?

CharliePoole:
I haven't thought much about it since it's a brand new feature to support.
[...]

PR #1863, opened 2016-10-25
Implement Runtime Parameter Discovery and Validation #1771

oznetmaster:
This commit is the result of backporting my existing Crestron/CF code implmenting #1771  to
the full nunit release.

Had to adjust a lot of things, since the original code was CF specific.

This needs a number of additional tests to validate non-CF specific compatibilities.

Note that no value validation is currently done in this code. In my implmentation, that is done in the GUI using the information returned during test exploration.

It might be worth adding that validation to the framework as well.


rprouse:
For anyone reviewing this, add `?w=1` to the end of the URL to ignore whitespace, https://github.com/nunit/nunit/pull/1863/files?w=1

rprouse:
@oznetmaster this is a lot more than I expected. I thought that you wanted to add an attribute that could be reported on in the explore phase, but it looks like you have also rewritten and added to much of the existing test parameter code. I am going to have to review this.

oznetmaster:
This is everything I discussed in the issue.  It includes the test method parameter reference option that I described there, which is much of the code, as well as the enhanced typing of the parameters.

It also includes a set of tests for all of the implemented capabilities.

The PR was for review, since you wanted to "see some code" :)

The CF version has been in use for some time, and is not going away in my build, so this is really just for the general nunit review.

I really think it has a general applicability, but that is for everyone else to decide.

I will not expand the tests or do anything else with this version until I get some feedback.

oznetmaster:
The failure is seemingly a bug in the cake script.

oznetmaster:
Actually, I have not changed the functionality or implementation of the existing test parameter in any way.

This is all built on top of that,

CharliePoole:
@rprouse This is more than I bargained for as well, based on the discussion of #1771. As I stated, I don't like the proposed feature but was willing to go for what sounded like a simple exposure of the parameters used by the assembly.

This goes much farther and exposes two new features:

1. Inventory of run parameters at the assembly level is made possible by the TestParameter attribute. A runner could optionally use this attribute for validation if desired.

2. Linkage of run parameters to method parameters by placing two attributes (TestCase and TestParameter) on a method. This seems like an overly complicated way to do it even if we want the feature. Additionally, it involves lots of changes to existing code in a way that does not reflect how NUnit now uses attributes.

Of the two features, I think we might want the second one more than the first. However, I'd rather see an implementation that is consistent with the rest of NUnit - i.e. creating an attribute that actually provides data rather than modifying TestCaseAttribute to work in conjunction with a second attribute. That said, it's a fairly low priority right now.

oznetmaster:
I am confused, since this is EXACTLY what I described in #1771.

I only back-ported it because @rprouse asked to see actual code.  I assumed that was because he liked what he read in #1771.

Since I use all of these capabilities in by CF build and its GUI, I am happy to leave it at that.

I do find it difficult to imagine a fully functional GUI without the first.  The second is syntactic sugaring that simplifies and documents the use of runtime parameters.

There is NO requirement that the TestCase attribute be used with TestParameter attribute.  Not sure why you are implying that is the case.  I think only one of my test cases actually does that.

In fact, I believe that I have never seen the two used together by my users, but allowing it is for consistency.

oznetmaster:
It should be noted that exactly the same capability can be accomplished with a lot more work by using TestCaseSource and having the method return objects based upon the TestContext.Parameters values.

It is just a more complex way of doing it.

CharliePoole:
@oznetmaster it is what you initially described but I understood you and rob to have negotiated a reduced set of changes. Frankly, I was busy with other things and didn't read closely.

In looking at your code, the first change I came across was the TestCaseAttribute, which examined it's own method for an instance of TestParameter attribute. I assumed that was the main way of using it, which I now see not to be the case.

It's an interesting observation that one could use TestCodeSource to do the same thing. The code is a bit complicated if it has to be done by the user each time but could be encapsulated easily in an attribute that actually served data, which would be reusable.

oznetmaster:
The TestParameter attribute is in fact just that attribute, since it can serve data.  That is the whole idea of the TestMethodParameterName property.  It maps the test parameter to the test method parameter, in much the same way as TestCase and TestCaseSource do.

I could see the usefulness of a TestCaseParameter attribute which was more like TestCase, but listed the parameters to use for each test case, and could be applied multiple times to a given test method.  It would be positional (like TestCase) rather then name driven.  It would be pretty trivial to implement.

It also, by its very nature, would both define and type test parameters, but would not provide any validation values.

oznetmaster:
I have also been considering adding the ability for a test parameter to be defined on the command line as an array of values (using a {} notation).  This could directly result in the TestCaseParameter attribute building multiple tests, one for each value in the array.  If multiple test parameters were involved, then all of the possible combinations would be generated.

rprouse:
@oznetmaster this is much more than I expected, so I am going to close this. Thanks for showing us the code, it is more work, but a much more precise way of discussing a feature.

PR #1868, opened 2016-10-28
Add Assert.AreEqual(DateTime expected, DateTime actual?, TImeSpan delta)

Corniel:
Add Assert.AreEqual(DateTime expected, DateTime actual?, TImeSpan delta) and fine tuned the message in case of a mismatch on ticks level.


rprouse:
I am on my phone so will review later, but is there an associated issue for this PR?

Corniel:
Not yet. Do I have to create that myself?

CharliePoole:
You do need to tell us somewhere why we may want this feature. You could create an issue or just add some explanation to the description of this PR if you prefer. Normally, it's better to do this before coding, since we may not actually want a proposed feature.

This particular featire has a bit of a hill to climb. We have generally stopped enhancing the "classic" asserts from several years back. New functionality goes into the constraint-based model. We have made a few exceptions for simple things like Assert.IsZero, but this is a bit more. 

The (semi) frozen status of classic asserts is a compromise we made back when we were thinking of dropping them entirely. We kept the AreEqual overload with delta for doubles for backward compatibility. If we add this new overload for DateTime, it could be argued that we should add it for every data type supported by AreEqual. I don't think that would be useful and it would clutter the user interface with options that are not very useful to most people.

If I have missed something that would justify adding this overload, please explain it. I don't mean to go negative on you, but just asking "why do you want this?" without giving the history wouldn't be very transparent on my part.

Corniel:
@CharliePoole I see your point. I have to admit I wasn't aware about the new syntax before (I adapted it in my implementation) . What I also changed in my pull request, is the way DateTime are rendered. Quite often, the difference is in the ticks, but that difference is normally not displayed, by using 'F' instead of 'f' for displaying milliseconds/ticks.

It was also a good exercise for me to get more insight in the internals of NUnit.

CharliePoole:
Yes the new syntax allows you to write something like

```C#
Assert.That(actualDT, Is.EqualTo(expectedDT).Within(3).Ticks);
```

which makes for a much more readable test.

If we wanted this overload, I'd approve your change as you wrote it. The code is fine. But I really don't want to support Assert.AreEqual with tolerances except for backward compatibility. 

I just created issue #1870 to cover the matter of the ticks display. I made it a low-priority bug and added it to our backlog. That's the normal process. Adding it to the backlog tells folks it's something to work on. Sometimes people submit a PR without an issue and that's OK, but they are taking their chances if it's not a simple bug fix. Enough said. 😄 

If you want to work on this smaller bug, just for the practice, I'll wait for a new PR. Otherwise, it's a quick fix and I'll just copy your change to a branch and push it myself. Feel free to submit other issues and PRs - we do appreciate the help!

@nunit/core-team Do you agree with this disposition? Since the work is done, I won't close this until one of you concurs.

rprouse:
I agree and was going to close myself. I had something typed up, but somehow it never got posted.

We would accept a PR for displaying differences in ticks. We just merged a similar PR for different object types that had the same ToString() output, #1848. If you want to work on the ticks issue, you may want to look at that PR first.

CharliePoole:
@rprouse I checked out #1848 before I replied here. It provides a fallback to the basic formatting of each Type if two types end up producing the same string. The new issue deals with the basic formatting for DateTime and DateTimeOffset and will keep the duplication from happening in the first place. It's merely a matter of not showing sufficient resolution in the strings produced.

@Corniel please post a comment on #1870 if you want to work on it.

rprouse:
@CharliePoole I only intended #1848 as a pointer to a place where @Corniel might want to add the code for the ticks resolution and do it in a similar way.

CharliePoole:
Ah, sorry, I thought you were saying it might no longer be needed. In his PR, @Corniel already made the change in MsgUtils, which is where the formatting string is currently located.

PR #1869, opened 2016-10-28
Fix #1866 More readable way to set polling interval in After constraint

### LINKED ISSUE #1866: [Request] More readable way to set polling interval in After constraint
This issue is a follow up of #1837 (originally requested by @mbendtsen ) to work on improving the readability of setting a polling interval in the After constraint.

rvignesh89:
Changes for #1866 

@CharliePoole I followed you suggestion for nested classes. This structure follows your idea as below,
- After(int,int) would give the base class, which would allow no further mods.
- After(int) would give nested class 1, allowing entry a dimension (minutes, etc.) or PollEvery
- Entering Minutes would give nested class 2, allowing only PollEvery
- PollEvery would give nested class 3, allowing only a polling dimension (minute, etc.)
- Minute would give the base class. Allowing no further modifiers.

I had to change the following to make this work,
1. Store delayInMilliseconds(int) apart from the delayInterval so that I could create nested class objects later when I needed.
2. Fix AssertionHelper tests for After(int) since it returns a nested type now.

There are many things missing (like tests and intervals for PollEvery) which I will add on later. Wanted to get the feedback if this is what you were looking for first.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Constraint.cs
+++ b/src/NUnitFramework/framework/Constraints/Constraint.cs
@@ -294,13 +294,13 @@ namespace NUnit.Framework.Constraints
         /// <summary>
         /// Returns a DelayedConstraint.WithRawDelayInterval with the specified delay time.
         /// </summary>
-        /// <param name="delayInMilliseconds">The delay in milliseconds.</param>
+        /// <param name="delay">The delay, which defaults to milliseconds.</param>
         /// <returns></returns>
-        public DelayedConstraint.WithRawDelayInterval After(int delayInMilliseconds)
+        public DelayedConstraint.WithRawDelayInterval After(int delay)
         {
             return new DelayedConstraint.WithRawDelayInterval(new DelayedConstraint(
                 Builder == null ? this : Builder.Resolve(),
-                delayInMilliseconds));
+                delay));
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/Constraint.cs
+++ b/src/NUnitFramework/framework/Constraints/Constraint.cs
@@ -296,9 +296,9 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         /// <param name="delayInMilliseconds">The delay in milliseconds.</param>
         /// <returns></returns>
-        public DelayedConstraint After(int delayInMilliseconds)
+        public DelayedConstraint.DelayedConstraint1 After(int delayInMilliseconds)
         {
-            return new DelayedConstraint(
+            return new DelayedConstraint.DelayedConstraint1(
                 Builder == null ? this : Builder.Resolve(),
                 delayInMilliseconds);
         }
--- a/src/NUnitFramework/framework/Constraints/Constraint.cs
+++ b/src/NUnitFramework/framework/Constraints/Constraint.cs
@@ -292,15 +292,15 @@ namespace NUnit.Framework.Constraints
 
 #if !PORTABLE
         /// <summary>
-        /// Returns a DelayedConstraint with the specified delay time.
+        /// Returns a DelayedConstraint.WithRawDelayInterval with the specified delay time.
         /// </summary>
         /// <param name="delayInMilliseconds">The delay in milliseconds.</param>
         /// <returns></returns>
-        public DelayedConstraint.DelayedConstraint1 After(int delayInMilliseconds)
+        public DelayedConstraint.WithRawDelayInterval After(int delayInMilliseconds)
         {
-            return new DelayedConstraint.DelayedConstraint1(
+            return new DelayedConstraint.WithRawDelayInterval(new DelayedConstraint(
                 Builder == null ? this : Builder.Resolve(),
-                delayInMilliseconds);
+                delayInMilliseconds));
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
@@ -25,7 +25,6 @@
 using System;
 using System.Diagnostics;
 using System.Threading;
-using NUnit.Compatibility;
 using NUnit.Framework.Internal;
 
 namespace NUnit.Framework.Constraints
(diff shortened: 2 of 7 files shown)


src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -35,9 +35,129 @@ namespace NUnit.Framework.Constraints
...
+            public DelayedConstraint3(DelayedConstraint1 parent)
+                : base(parent.BaseConstraint, parent._delayInMilliseconds, parent._pollingInterval)
+            {
+                _parent = parent; ;
+            }
+
+            /// <summary>
+            /// 
+            /// </summary>
+            public DelayedConstraint Minute
```
CharliePoole: I guess Minutes reads better. :-) I assume you'll add the other dimensions as well as PollEvery.

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -35,9 +35,129 @@ namespace NUnit.Framework.Constraints
...
+            {
+                _parent._pollingInterval = milliSeconds;
+                return new DelayedConstraint3(_parent);
+            }
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public class DelayedConstraint1 : DelayedConstraint
```
CharliePoole: We should think up some better names. Maybe...

Existing | Better
--------|-------
DelayedConstraint.DelayedConstraint1 | DelayedConstraint.RawDelay
DelayedConstraint.DelayedConstraint2 | DelayedConstraint.DimensionedDelay
DelayedConstraint.DelayedConstraint3 | DelayedConstraint.RawPollingInterval

I'm not super happy with these names, so feel free to invent. They can be based on what has been specified or what has not yet been specified but it should probably be the same approach for each class.

CharliePoole: Also, we __really__ need a comment on each of these classes so future devs will understand.

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -35,9 +35,129 @@ namespace NUnit.Framework.Constraints
...
+            public DelayedConstraint1(IConstraint baseConstraint, int delayInMilliseconds) : base(baseConstraint, delayInMilliseconds) { }
+
+            ///<summary>
+            /// Creates a new DelayedConstraint
+            ///</summary>
+            ///<param name="baseConstraint">The inner constraint to decorate</param>
+            ///<param name="delayInMilliseconds">The time interval after which the match is performed, in milliseconds</param>
+            ///<param name="pollingInterval">The time interval used for polling, in milliseconds</param>
+            ///<exception cref="InvalidOperationException">If the value of <paramref name="delayInMilliseconds"/> is less than 0</exception>
+            public DelayedConstraint1(IConstraint baseConstraint, int delayInMilliseconds, int pollingInterval) : base(baseConstraint, delayInMilliseconds, pollingInterval) { }
```
CharliePoole: What's the need for this constructor?

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -35,9 +35,129 @@ namespace NUnit.Framework.Constraints
...
+        /// </summary>
+        public class DelayedConstraint1 : DelayedConstraint
+        {
+            ///<summary>
+            /// Creates a new DelayedConstraint
+            ///</summary>
+            ///<param name="baseConstraint">The inner constraint to decorate</param>
+            ///<param name="delayInMilliseconds">The time interval after which the match is performed</param>
+            ///<exception cref="InvalidOperationException">If the value of <paramref name="delayInMilliseconds"/> is less than 0</exception>
+            public DelayedConstraint1(IConstraint baseConstraint, int delayInMilliseconds) : base(baseConstraint, delayInMilliseconds) { }
```
CharliePoole: What if this one took the parent instance, just like the others? Then the same parent could be passed around all the time.

CharliePoole:
This looks pretty good to me as an approach. I made a number of detail comments. In particular...
* Figuring out some better names
* Consider using outer class instance as an argument for all inner classes

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
rvignesh89: For this constructor I have 2 options. If I kept it in the new nested class then I wouldn't have to change tests like the one here to call the new class.

```
[TestFixture]
	public class AsyncDelayedConstraintTests
	{
		[Test]
		public void ConstraintSuccess()
		{
			Assert.IsTrue(new DelayedConstraint(new EqualConstraint(1), 100)
				.ApplyTo(async () => await One()).IsSuccess);
		}

```

Also have to change the After method in Constraint.cs to this,

```
/// <summary>
        /// Returns a DelayedConstraint with the specified delay time.
        /// </summary>
        /// <param name="delayInMilliseconds">The delay in milliseconds.</param>
        /// <returns></returns>
        public DelayedConstraint.DelayedConstraint1 After(int delayInMilliseconds)
        {
            return new DelayedConstraint.DelayedConstraint1(new DelayedConstraint( 
                Builder == null ? this : Builder.Resolve(),
                delayInMilliseconds));
        }
```

But thinking about it I realize if I give the nested class a better name it would still make sense to make this change.  This would also make the nested class depend on the parent which would be consistent. I'll change this and see how it looks first.

CharliePoole:
All sounds good!

rprouse:
@rvignesh89 looks good so far. Ping us when you push the changes and I will give this a proper second review.

rvignesh89:
@rprouse @CharliePoole I've made the changes requested. Some things to note,

1. For the nested classes names I prefixed them with, "With" because I felt it reads nicely when combined with the parent class. Like DelayedConstraint.WithRawDelayInterval. 
2. I've followed the existing pattern for writing tests for PollEvery. However this makes the entire Fixture slower to run. Now `nunit-framework-tests.dll` takes around 200s to run while previously it takes ~95s. I think this is more related to running integration style tests which depend on Thread.Sleep inside this suite. Have you considering moving those slower tests to a different dll?
3. I changed polling interval from int to also use my new Interval class.

rvignesh89:
I missed to add tests for WithDimensionedDelayInterval.PollyEvery(), will add that now as well.

CharliePoole:
Sorry, I didn't notice you had updated... will review now.

src/NUnitFramework/framework/Constraints/Constraint.cs
```diff
@@ -292,15 +292,15 @@ public ConstraintExpression Or
 
 #if !PORTABLE
         /// <summary>
-        /// Returns a DelayedConstraint with the specified delay time.
+        /// Returns a DelayedConstraint.WithRawDelayInterval with the specified delay time.
         /// </summary>
         /// <param name="delayInMilliseconds">The delay in milliseconds.</param>
         /// <returns></returns>
-        public DelayedConstraint After(int delayInMilliseconds)
+        public DelayedConstraint.WithRawDelayInterval After(int delayInMilliseconds)
```
CharliePoole: This is a bit of a nit pick, but is it possibly confusing to keep the name "delayInMilliseconds" when it may end up being changed? This shows up in intellisense with the explanation "The delay in milliseconds."

Maybe it should be simply "delay" with a description like " The delay, which defaults to milliseconds"

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -35,9 +34,159 @@ namespace NUnit.Framework.Constraints
     ///</summary>
     public class DelayedConstraint : PrefixConstraint
     {
+        /// <summary>
+        /// Allows only changing the time dimension of the polling interval of a DelayedConstraint
+        /// </summary>
+        public class WithRawPollingInterval : DelayedConstraint
```
CharliePoole: Another nit pick: I got confused because the nested classes are listed in the reverse order in which the syntax appears normally.

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -35,9 +34,159 @@ namespace NUnit.Framework.Constraints
...
+            {
+                _parent._pollingInterval = new Interval(milliSeconds).InMilliseconds;
+                return new WithRawPollingInterval(_parent);
+            }
+        }
+
         // TODO: Needs error message tests
         private Interval _delayInterval;
-        private readonly int _pollingInterval;
+        private Interval _pollingInterval;
```
CharliePoole: These private Intervals are of course accessible in the nested classes but it may be clearer to future devs if we make this protected. Could also be a protected automatic Property.

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -35,9 +34,159 @@ namespace NUnit.Framework.Constraints
...
+
+            /// <summary>
+            /// Changes delay interval dimension to seconds
+            /// </summary>
+            public WithDimensionedDelayInterval Seconds
+            {
+                get
+                {
+                    _delayInterval = _delayInterval.InSeconds;
+                    return new WithDimensionedDelayInterval(this);
```
CharliePoole: Shouldn't this be _parent?

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -35,9 +34,159 @@ namespace NUnit.Framework.Constraints
...
+
+            /// <summary>
+            /// Changes delay interval dimension to milliseconds
+            /// </summary>
+            public WithDimensionedDelayInterval MilliSeconds
+            {
+                get
+                {
+                    _delayInterval = _delayInterval.InMilliseconds;
+                    return new WithDimensionedDelayInterval(this);
```
CharliePoole: _parent again?

src/NUnitFramework/framework/Constraints/Constraint.cs
rvignesh89: Agree. Makes sense to mention that it is defaulted to milliseconds.

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
rvignesh89: Haha, I didn't expect this. But if you find it confusing someone else might as well. Will modify it

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
rvignesh89: Yes, will fix it.

CharliePoole:
@rprouse This looks ready to merge for me. Will you re-review?

rprouse:
Looks good to me too, thanks for the contribution. 👍

PR #1871, opened 2016-10-28
Fix for #1870 Fine tune for the DateTime and DateTimeOffset format

### LINKED ISSUE #1870: EqualConstraint result failure message for DateTime doesn't show sufficient resolution
As reported on PR #1868 the failure message does not show ticks, which may cause the two values to be displayed the same even though they are not equal. We should use 'F' rather than 'f' in the format string.

Corniel:
Fine tuned the DateTime and DateTimeOffset format, so that ticks are also shown in a assertion message if required. Added unit test too.

Fixes #1870


CharliePoole:
Thanks for doing this. Looks good to me. Waiting for second review from one of the @nunit/core-team members.

rprouse:
Looks good, nice and simple 👍 

Thanks for the contribution.

Issue #391, opened 2014-11-27
Multiple Assertions


circa1741:
May I ask if there is a plan to have something similar to MbUnit's Multiple Assertions capability?

This might provide some insight:
https://code.google.com/p/mb-unit/source/browse/trunk/v3/src/MbUnit/MbUnit/Framework/Assert.cs?r=1570#282

Evaluates a block of code that contains multiple related assertions.

While the action delegate runs, the behavior of assertions is change such that
failures are captured but do not cause a <see cref="AssertionFailureException" />
to be throw.  When the delegate returns, the previous assertion failure behavior
is restored and any captured assertion failures are reported.  The net effect
of this change is that the test can continue to run even after an assertion failure
occurs which can help to provide more information about the problem.

A multiple assertion block is useful for verifying the state of a single
component with many parts that require several assertions to check.
This feature can accelerate debugging because more diagnostic information
become available at once.

Also:
https://code.google.com/p/mb-unit/source/browse/trunk/v3/src/MbUnit/MbUnit/Framework/MultipleAssertsAttribute.cs?r=1570

Runs the test as if it were surrounded by <see cref="Assert.Multiple(Action)" /> so that
multiple assertion failures within the test are tolerated.

When an assertion failure occurs, it is reported but the test is allowed to proceed
until it completes or throws an unhandled exception.  When the test finishes, it will
still be marked as failed, as usual.
## 

Partial specs as entered below by @CharliePoole 

There are a number of things needed here and I guess it's time to make a list! I may end up breaking some of these out into separate issues, but for the moment let's just stick with bullet points.

- We will have to redesign the xml result format to support multiple results per test case. Currently, each test can only have a single result and tests are nested inside other tests (suites). However, if each assertion can generate a separate result, we need to hold them somewhere. This might be, for example, a `<results>` element containing one or more `<result>` elements and superseding some existing elements.
- The internal framework TestResult Type will need to hold multiple results and generate xml as designed in the first bullet.
- Clients and the engine will need a way to recognize which format is being used.
- Clients and the engine may need a way to control which format gets used.
- We'll have to firm up the syntax we intend to support. This is the part we have mostly discussed here, so I think it's pretty ready.
- We need to introduce a mechanism for reporting results during a test without generating an exception, which would, of course, stop the test.
- Finally, we have to implement the syntax for multiple assertions.
 

The preliminary part of this, supporting multiple results per test, will actually be useful for three different features we would like to have:
1. Multiple Assertions (this issue)
[...]

PR #1875, opened 2016-10-31
Add multiple assertion results to XML output

### LINKED ISSUE #1743: Provide multiple results for a test case in the XML output
This feature is needed in order to support
- Multiple Asserts - so that multiple failures can be reported
- Warning-level Asserts - so that both warnings and any eventual failure can be reported
- General assertion reporting - so that all asserts can be reported

Since MultipleAsserts is queued up first, I'm making it a part of that Epic.

CharliePoole:
Fixes #1743 

The initial commit adds elements to the XML that allow specifying information about assertions.

Additionally, the TestResult tests have been refactored to eliminate duplication and each test fixture has been moved to it's own file. There are new tests for AssertionResults.

The next commit will deal with actually using the new capability by storing info about the failing assert that ends a test in the result, duplicating what is in the existing `<failure>` element. 

Actually using this capability for the MultipleAsserts feature will be done under a separate PR.


CharliePoole:
Ready for review and merging. Further work will follow on issue #391.

CharliePoole:
Failure is an unrelated timing issue that has been newly introduced with changes in DelayConstraint tests. I restarted.

CharliePoole:
@rprouse You should take a look at this as it's a major change to the XML output.

@remcomulder Love to have you check it out as well.

Note that `AssertionStatus` attribute is currently only "Passed" or "Failed" but could have other values added in future.

remcomulder:
@CharliePoole Do these commits also add anything to allow us to access the new xml through the runner? I had a go at trying to surface the new results for a good look.  Is the new code still disconnected?

CharliePoole:
@remcomulder This PR and issue is about the form of the result file. Right now the only way that anything is included is if an assert or assume fails. In the case of an assertion failure, the info in `<assertions>` duplicates what is in `<failure>`.

There is a method that can be used to add assertion info, but it's only intended for use by the framework itself. When I move to issue #391, every failure under a multiple assertion will create an entry.

It's unusual for us to implement an underlying capability separately from exposing it to the users, but in this case I wanted to make it behave consistently from the start, even without the presence of new user features to make use of it.

A key thing to review, either now or on the future PR for #391, is whether we want any more info about the assertion. Of course, this can develop as we work with it but I'd like it to be fairly complete for the 4.0 release.

src/NUnitFramework/framework/Interfaces/AssertionResult.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <summary>The message produced by the assertion, or null</summary>
+        public string Message { get; private set; }
+
+        /// <summary>The stacktrace associated with the assertion, or null</summary>
+        public string StackTrace { get; private set; }
+
+        /// <summary>
+        /// ToString Override
+        /// </summary>
+        public override string ToString()
```
rprouse: I don't see this used anywhere except the GetHashCode. Is it just here for debugging and to feed the hash code? If so, then it is fine, otherwise, if it will be used externally we may not want the newline if the stacktrace is null.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -574,6 +615,23 @@ private TNode AddOutputElement(TNode targetNode)
...
+            {
+                TNode assertionNode = assertionsNode.AddElement("assertion");
+                assertionNode.AddAttribute("result", assertion.Status.ToString());
+                if (assertion.Message != null)
+                    assertionNode.AddElementWithCDATA("message", assertion.Message);
+                if (assertion.StackTrace != null)
+                    assertionNode.AddElementWithCDATA("stack-trace", assertion.StackTrace);
+            }
+
+            return assertionsNode;
```
rprouse: This is the meat of all this, the XML format that will be output. The format looks fine to me. As far as I can tell, your code matches the description in the original issue, so no problems.

rprouse:
@CharliePoole you mention the 4.0 release, are you thinking the adapter?

This looks good. I am going to let you merge in case you want anyone else to review.

src/NUnitFramework/framework/Interfaces/AssertionResult.cs
CharliePoole: Purely meant for debugging. I generally don't add stuff to the SUT for test perposes, but I make an exception for ToString(), Equals and GetHashcode because they are so commonly needed and relatively easy to implement in most cases.

CharliePoole:
@rprouse I meant 3.6 - updated the comment. I'll merge and we can add any other stuff to it with the PR for issue #391 - or after.

PR #1882, opened 2016-11-04
.NET Standard Version of the Framework

### LINKED ISSUE #1661: Create .NET Standard Framework Build

rprouse:
**~~Do not merge~~**

This is nearly done, but still needs a bit more work,

- [x] Get Travis Building
- [x] Determine if .NET Core tools are installed and don't build those targets if not
- [x] Update the NuGet package
- [x] Ensure it is building on Linux
- [x] Test in MonoDevelop

This adds the .NET Standard into the main `NUnit.sln` and **requires Visual Studio 2015 Update 3**. I think it also requires the [.NET Core SDK](https://www.microsoft.com/net/core#windows) to be installed. I would prefer to keep everything in the one solution and let people with older versions of Visual Studio use the `NUnit.Linux.sln`, but I will leave that open for discussion.

This version of the .NET Standard Framework is functionally equivalent to the Portable version. After this is merged, we can start re-enabling features like timeout, platform detection, directory/file contraints, etc as separate PRs.

Fixes #1661


rprouse:
I don't expect Travis to pass yet. AppVeyor is failing on NuGet restore. I will look into it tomorrow.

src/NUnitFramework/nunitlite.tests/CommandLineTests.cs
```diff
@@ -172,7 +172,7 @@ public void CanRecognizeLowerCaseOptionValues(string propertyName, string option
 
             foreach (string canonicalValue in canonicalValues)
             {
-                string lowercaseValue = canonicalValue.ToLower(CultureInfo.InvariantCulture);
+                string lowercaseValue = canonicalValue.ToLower();
```
rprouse: The form of `ToLower()` that takes a `CultureInfo` is not available on .NET Standard so I dropped it rather than use another `#if`

oznetmaster: Shouldn't you use ToLowerInvariant(), which I believe is in .Net Standard (at least the documentation says it is.

rprouse: Yes, probably, thanks. It was late and this was just a test, so I didn't spend much time on it 😄 

I will push the change.

ChrisMaddock:
> I would prefer to keep everything in the one solution and let people with older versions of Visual Studio use the NUnit.Linux.sln, but I will leave that open for discussion.

If Visual Studio is missing .net core, I believe it just doesn't load the project in question. Does MonoDevelop work the same way? If so, could we just go down to one solution, and just let the netstandard project be ignored if it's not supported?

I was hoping we could get rid of the Linux sln, now CF and SL are gone, and everything in this repo should be running cross-platform. 😄

src/NUnitFramework/Fakes.cs
```diff
@@ -26,7 +26,7 @@
 using NUnit.Framework.Internal;
 using NUnit.Framework.Internal.Execution;
 
-#if PORTABLE && !NUNITLITE
+#if PORTABLE && !NUNITLITE && !NETSTANDARD1_6
```
CharliePoole: You should add the new defines to our list on the wiki.

rprouse:
@CharliePoole I was also surprise how few code changes were needed. Last time I attempted this with the XPROJ format I didn't define PORTABLE, so most of the code changes were adding NETSTANDARD conditionals. When you think about it though, it isn't too surprising since NETSTANDARD is the new Portable library. Portable was very restrictive, but the .NET Standard version I selected has more APIs, so no changes were needed and hopefully we will be able to re-enable more features with the expanded API coverage.

My plan going forward is to stop defining PORTABLE in the .NET Standard projects and add the conditionals where necessary. This will start to highlight what we can re-enable. Defining PORTABLE however is a quick way to get this building and get the .NET Standard version of the framework into the project.

@CharliePoole, you asked if netcore is temporary. It is not. Think of .NET Standard as the replacement for Portable libraries and .NET Core being a CLR that .NET Standard can run on in addition to the full framework (4.5, 4.6.1, etc.), Xamarin, etc. Notice that the two executables, the NUnitLite tests and runner are compiled to target .NET Core and must be run with the `dotnet` command.

As for choosing .NET Standard 1.6, I selected that because it was the recommendation when using the CSPROJ format, is the .NET Core 1.0 standard, and it gives fairly broad API support. 2.0 is also appealing as it is what vNext is targeting, but we might also want to look at backing off to 1.4 to give us original UWP support. Take a look at https://blogs.msdn.microsoft.com/dotnet/2016/09/26/introducing-net-standard/, specifically the version chart.

src/NUnitFramework/Fakes.cs
CharliePoole: I understand they are standard, and the wiki should mention that, but I think we need a list of those defines we actually make use of. I've always found it annoying that we have to define the target runtime rather than having it done for us for by VS.

rprouse:
@jplebre do you have any idea what is causing the AppVeyor Nuget restore issues for the NuGet restore? First error is,

`System.Text.RegularExpressions 4.1.0 provides a compile-time reference assembly for System.Text.RegularExpressions on .NETCoreApp,Version=v1.0, but there is no run-time assembly compatible with osx.10.10-x64.`

rprouse:
I've been spinning my wheels getting this to build on Travis/Linux so I am thinking of deferring that to a separate issue/PR if people feel it is important. For now, I don't build .NET Standard if not running on Windows similar to the portable project.

build.cake
```diff
@@ -87,10 +106,22 @@ Task("InitializeBuild")
     .Description("Initializes the build")
     .Does(() =>
     {
-        NuGetRestore(SOLUTION_FILE, new NuGetRestoreSettings()
+        foreach(var package in packages)
         {
-            Source = PACKAGE_SOURCE
-        });
+            NuGetRestore(package, new NuGetRestoreSettings
```
rprouse: NuGet restore fails when it hits the .NET Standard projects that need to be restored using the `dotnet restore` below.

build.cake
```diff
@@ -483,6 +570,27 @@ void RunTest(FilePath exePath, DirectoryPath workingDir, string arguments, strin
...
 }
 
+void RunDotnetCoreTests(FilePath exePath, DirectoryPath workingDir, string framework, ref List<string> errorDetail)
+{
+    RunDotnetCoreTests(exePath, workingDir, null, framework, ref errorDetail);
+}
+
+void RunDotnetCoreTests(FilePath exePath, DirectoryPath workingDir, string arguments, string framework, ref List<string> errorDetail)
+{
+    int rc = StartProcess(
```
rprouse: The .NET Core tests need to be run using the `dotnet` command.

src/NUnitFramework/tests/Constraints/DelayedConstraintTests.cs
```diff
@@ -103,6 +103,7 @@ public void SimpleTest()
         }
 
         [Test]
+        [Explicit("This test takes over a minute to run")]
```
rprouse: This test slipped in recently in a pull requests. This one test causes over a minute to run. Multiplied by the six platforms made running the tests painful. We have two other tests that test canceling and stopping test runs that add 20 seconds per platform. I would like to make those explicit also, but wanted to get @nunit/core-team's opinion first. I think that our tests taking a couple of minutes to run is too long. Do you agree?

nuget/nunitlite/nunitlite.nuspec
```diff
@@ -40,12 +40,13 @@ How to use this package:
...
-    <file src="bin/portable/nunitlite.dll" target="lib/dotnet" />
+    <file src="bin/netstandard/nunitlite.dll" target="lib/netstandard1.6" />
     <file src="bin/portable/nunitlite.dll" target="lib/portable-net45+win8+wp8+wpa81+Xamarin.Mac" />
     <file src="bin/net-2.0/nunitlite-runner.exe" target="tools/net20" />
     <file src="bin/net-3.5/nunitlite-runner.exe" target="tools/net35" />
     <file src="bin/net-4.0/nunitlite-runner.exe" target="tools/net40" />
     <file src="bin/net-4.5/nunitlite-runner.exe" target="tools/net45" />
+    <file src="bin/netstandard/nunitlite-runner.exe" target="tools/netcore" />
     <file src="bin/portable/nunitlite-runner.exe" target="tools/portable" />
     <file src="..\..\nuget\nunitlite\Program.cs" target="content" />
```
ChrisMaddock: I presume the netstandard nunitlite runner should be replacing the portable, here, like the other nuspecs?

build.cake
```diff
@@ -27,7 +27,7 @@ var packageVersion = version + modifier + dbgSuffix;
 //////////////////////////////////////////////////////////////////////
 
 var WindowsFrameworks = new string[] {
-    "net-4.5", "net-4.0", "net-3.5", "net-2.0", "portable" };
+    "net-4.5", "net-4.0", "net-3.5", "net-2.0", "netstandard", "portable" };
```
ChrisMaddock: The `PackageZip` task will also needs updating here. I think it can probably refer to this frameworks list, rather than have a duplicate!

nuget/nunitlite/nunitlite.nuspec
rprouse: @ChrisMaddock this is the .NET Core version of the runner that uses the .NET Standard version of the framework, that is why it is being copied. Confused yet?

build.cake
rprouse: @ChrisMaddock just so people are not confused by your comment, the packaging of the ZIP is done, but I had to add `netstandard` there too. We should refactor it to use this list. 👍

rprouse:
>If we're looking to remove portable, does choosing 1.6 mean we're effectively dropping support for the current UWP version via nunit.xamarin runners?

Targeting 1.6 would cause us to drop support for many of the older platforms if we dropped the PCL. I think if we drop the PCL, then I will add a `NETSTANDARD1_3` version to replace the support there. If I am going to do that, I will want multi-targeting from this one .NET Standard CSPROJ file though. I am still undecided, but I think we may end up with a .NET Standard 1.3, 1.6 and 2.0 version of the framework before we drop the PCL. I want to create a matrix of NUnit features and attributes that are not available in the PCL version along with which can be re-enabled with each version of .NET Standard.

rprouse:
@CharliePoole or @nunit/contributors, does anyone have a machine with Visual Studio 2015 Update 3 but not the .NET Core SDK installed? If so, can you pull this branch and tell me what your development experience is? Do the .NET Standard projects just fail to load?

Also, can you try opening the solution in an older version of Visual Studio? Can you do it and build everything except the .NET Standard projects?

I would appreciate the help. I am travelling this week and only have VS 2015 on this laptop.

rvignesh89:
@rprouse yes, the projects fail to load. The first time I opened the project VS highlighted that I need to make configuration changes on the app

image

On VS 2013 its the same. However I can build the other projects.

image

rprouse:
@rvignesh89 thanks for the detailed report, I really appreciate it. One issue I see is that the two EXE projects remain loaded and therefore will fail to build because the projects they reference are not loaded.

rprouse:
@nunit/core-team this is reviewed. Any objections to merging? Should we merge to a development branch or to master?

ChrisMaddock:
All looks great to me! Don't see why not to merge to master - the changes are pretty self-contained, right?

CharliePoole:
I think it's ready to merge. I'd like to see us have a development branch, but setting it up in the build correctly would take a little work that may best be done separately, so I'd put this to master.

drauch:
Is there any update on when we might expect to have a NuGet (prerelease) package available which supports .NET Standard 1,6 projects?

rprouse:
@drauch the next full release is due at the end of the year, most likely beginning January with the holidays coming. We are no longer doing pre-releases on NuGet for most releases, so if you want to test it early, you can download from our MyGet feed, the feed URI is https://www.myget.org/F/nunit/api/v2 and the package you want is https://www.myget.org/feed/nunit/package/nuget/NUnit

Snipx:
Hi @rprouse ,

I get problems using Timeout attribute with the NuGet package you posted the link to, although for other attributes like Test it is ok. I see from the first post that this is a known issue. Will is be solved before the next release?

rprouse:
Timeout won't be available in 3.6. Hopefully it will be in 3.7.

PR #1884, opened 2016-11-04
Drop installation of CF from AppVeyor build

CharliePoole:
We left this when we removed the CF build


ChrisMaddock:
Looks good to me, but Appveyor has had a sporadic failure, and I don't seem to have the ability to merge without it, or re-run! Perhaps you can, @CharliePoole. :-)

rprouse:
Looks good. Just to be sure, I have restarted the build.

@ChrisMaddock I have sent you an invite as an AppVeyor user for the account to your gmail address. I think that should allow you to restart builds. If there are permission problems, let me know. As an aside, do you prefer your live or gmail address?

ChrisMaddock:
Thanks @rprouse! 

I primarily use live, but receive both. My gmail only came into the picture because of the Google groups 🙂

ChrisMaddock:
Looks like the new account just has read-only access - I still don't have the "New build" button. Also, is it possible to add it under my live account? I log in to appveyor through Github normally - turns out that's linked to my @live.co.uk address. Thanks!

PR #1888, opened 2016-11-06
Improve error message if too few args are provided on TestCaseAttribute

### LINKED ISSUE #1872: Parameterized method being called with no parameter
```
[TestCase, TestCaseSource(typeof(BooleanTestData))]
public void Test(bool authorized)
{
  ...
}

public class BooleanTestData : IEnumerable
{
  public IEnumerator GetEnumerator()
  {
    yield return true;
    yield return false;
  }
}
```

-OR-

```
[TestCase, TestCaseSource(typeof(BooleanTestData), "Value")]
public void Test(bool authorized)
{
  ...
}

public class BooleanTestData
{
  public static IEnumerable Value
  {
    get
    {
      yield return new TestCaseData(true);
      yield return new TestCaseData(false);
    }
  }
}

 -or-

public class BooleanTestData
{
  public static IEnumerable Value
  {
    get
    {
      yield return true;
      yield return false;
    }
  }
}

 -or-

public class BooleanTestData
{
  public static bool[] Value
  {
    get
    {
      return new bool[] { true, false };
    }
  }
}
```

results in this:

```
<test-case id="0-1001" name="Test(True)" fullname="MyDll.TestAuthorizer.Test(True)" methodname="Test" ...
  <reason>
    <message><![CDATA[OneTimeSetUp: true]]></message>
  </reason>
</test-case>
<test-case id="0-1002" name="Test(False)" fullname="MyDll.TestAuthorizer.Test(False)" methodname="Test" ...
  <reason>
    <message><![CDATA[OneTimeSetUp: true]]></message>
  </reason>
</test-case>
<test-case id="0-1003" name="Test" fullname="MyDll.TestAuthorizer.Test" methodname="Test" ...
  <properties>
    <property name="_SKIPREASON" value="System.Reflection.TargetParameterCountException : Incorrect number of parameters specified for TestCase" />
    <property name="_PROVIDERSTACKTRACE" value="   at NUnit.Framework.TestCaseAttribute.GetParametersForTestCase(IMethodInfo method)" />
  </properties>
  <reason>
    <message><![CDATA[OneTimeSetUp: true]]></message>
  </reason>
</test-case>
```

It appears that an extra call to the parameterized method is being generated that uses no parameters.

CharliePoole:
Fixes #1872 

I'm hoping that a message like

```
Method requires 2 arguments but TestCaseAttribute only supplied 0
```

will make it clearer.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
@@ -320,7 +320,10 @@ namespace NUnit.Framework
                             if (i < parms.Arguments.Length)
                                 newArgList[i] = parms.Arguments[i];
                             else
-                                throw new TargetParameterCountException("Incorrect number of parameters specified for TestCase");
+                                throw new TargetParameterCountException(string.Format(
+                                    "Method requires {0} arguments but TestCaseAttribute only supplied {1}",
+                                    argsNeeded, 
+                                    argsProvided));
                         }
                     }
                     parms.Arguments = newArgList;


rprouse:
Looks good

PR #1890, opened 2016-11-06
Make NUnitLite use AssertionResult info in it's reporting

### LINKED ISSUE #1889: Modify nunitlite to display multiple assert information
Currently, the nunitlite runner uses the single failure message created for compatibility in order to display the result of a test using multiple asserts. It should use the actual result of each assert to display both the message and stack frame.

CharliePoole:
Fixes #1889 

The first commit is an initial pass at using AssertionResult for our reporting. Before merging, let's use this as an opportunity to do a final review on the formatting of our reports.

In order to make this change I did some refactoring of StackFilter so it could be reused for different kinds of filtering.

### DIFF
--- a/src/NUnitFramework/nunitlite/TextUI.cs
+++ b/src/NUnitFramework/nunitlite/TextUI.cs
@@ -482,19 +482,33 @@ namespace NUnitLite
 
         private void DisplayTestResult(ITestResult result)
         {
-            string status = result.ResultState.Label;
+            ResultState resultState = result.ResultState;
+            string fullName = result.FullName;
+            string message = result.Message;
+            string stackTrace = result.StackTrace;
+
+            if (result.AssertionResults.Count > 0)
+                foreach (var assertion in result.AssertionResults)
+                    DisplayTestResult(resultState, fullName, assertion.Message, assertion.StackTrace);
+            else
+                DisplayTestResult(resultState, fullName, message, stackTrace);
+        }
+
+        private void DisplayTestResult(ResultState resultState, string fullName, string message, string stackTrace)
+        {
+            string status = resultState.Label;
             if (string.IsNullOrEmpty(status))
-                status = result.ResultState.Status.ToString();
+                status = resultState.Status.ToString();
 
             if (status == "Failed" || status == "Error")
             {
-                var site = result.ResultState.Site.ToString();
+                var site = resultState.Site.ToString();
                 if (site == "SetUp" || site == "TearDown")
                     status = site + " " + status;
             }
 
             ColorStyle style = ColorStyle.Output;
-            switch (result.ResultState.Status)
+            switch (resultState.Status)
             {
                 case TestStatus.Failed:
                     style = ColorStyle.Failure;
@@ -509,13 +523,13 @@ namespace NUnitLite
 
             Writer.WriteLine();
             Writer.WriteLine(
-                style, string.Format("{0}) {1} : {2}", ++_reportIndex, status, result.FullName));
+                style, string.Format("{0}) {1} : {2}", ++_reportIndex, status, fullName));
 
-            if (!string.IsNullOrEmpty(result.Message))
-                Writer.WriteLine(style, result.Message.TrimEnd(TRIM_CHARS));
+            if (!string.IsNullOrEmpty(message))
+                Writer.WriteLine(style, message.TrimEnd(TRIM_CHARS));
 
-            if (!string.IsNullOrEmpty(result.StackTrace))
-                Writer.WriteLine(style, result.StackTrace.TrimEnd(TRIM_CHARS));
+            if (!string.IsNullOrEmpty(stackTrace))
+                Writer.WriteLine(style, stackTrace.TrimEnd(TRIM_CHARS));
         }
 
 #if FULL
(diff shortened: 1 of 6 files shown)


CharliePoole:
Here's an example of output when we were using the "compatible" `<failure>` element:

```
Errors and Failures

1) Failed : NUnit.Framework.Assertions.Tests.AssertMultipleTests.MultipleAssertFailureDemo
  Multiple Assert block had 2 failure(s).

  1)   RealPart
  Expected: 5.0d
  But was:  5.2000000000000002d


  2)   ImaginaryPart
  Expected: 4.2000000000000002d
  But was:  3.8999999999999999d



at NUnit.Framework.Assertions.Tests.AssertMultipleTests.MultipleAssertFailureDemo()
```

Here's the same test using the new `<assertion>` elements:

```
Errors and Failures

1) Failed : NUnit.Framework.Assertions.Tests.AssertMultipleTests.MultipleAssertFailureDemo
  RealPart
  Expected: 5.0d
  But was:  5.2000000000000002d
at NUnit.Framework.Assertions.Tests.AssertMultipleTests.<>c.<MultipleAssertFailureDemo>b__3_0()
at NUnit.Framework.Assert.Multiple(TestDelegate testDelegate) in D:\Dev\NUnit\nunit-3.0\src\NUnitFramework\framework\Assert.cs:line 290
at NUnit.Framework.Assertions.Tests.AssertMultipleTests.MultipleAssertFailureDemo()

2) Failed : NUnit.Framework.Assertions.Tests.AssertMultipleTests.MultipleAssertFailureDemo
  ImaginaryPart
  Expected: 4.2000000000000002d
  But was:  3.8999999999999999d
at NUnit.Framework.Assertions.Tests.AssertMultipleTests.<>c.<MultipleAssertFailureDemo>b__3_0()
at NUnit.Framework.Assert.Multiple(TestDelegate testDelegate) in D:\Dev\NUnit\nunit-3.0\src\NUnitFramework\framework\Assert.cs:line 290
at NUnit.Framework.Assertions.Tests.AssertMultipleTests.MultipleAssertFailureDemo()
```

Essentially, the new way, you get two entries for the same test. The count of failures is based on tests, not asserts, so the summary would indicate 1 failure in both cases. Does anyone think that's a problem?

Let me know if you think the layout should be in any way different - right now it is exactly the same as for tests that do not use `Assert.Multiple`.

rprouse:
>  The count of failures is based on tests, not asserts, so the summary would indicate 1 failure in both cases. Does anyone think that's a problem?

I am okay with that. One idea to make it clearer might be for multiple results to use a two part number. It would require refactoring the display results. It could look like this,

```
Errors and Failures

1-1) Failed : NUnit.Framework.Assertions.Tests.AssertMultipleTests.MultipleAssertFailureDemo
  RealPart
  Expected: 5.0d
  But was:  5.2000000000000002d
at NUnit.Framework.Assertions.Tests.AssertMultipleTests.<>c.<MultipleAssertFailureDemo>b__3_0()
at NUnit.Framework.Assert.Multiple(TestDelegate testDelegate) in D:\Dev\NUnit\nunit-3.0\src\NUnitFramework\framework\Assert.cs:line 290
at NUnit.Framework.Assertions.Tests.AssertMultipleTests.MultipleAssertFailureDemo()

1-2) Failed : NUnit.Framework.Assertions.Tests.AssertMultipleTests.MultipleAssertFailureDemo
  ImaginaryPart
  Expected: 4.2000000000000002d
  But was:  3.8999999999999999d
at NUnit.Framework.Assertions.Tests.AssertMultipleTests.<>c.<MultipleAssertFailureDemo>b__3_0()
at NUnit.Framework.Assert.Multiple(TestDelegate testDelegate) in D:\Dev\NUnit\nunit-3.0\src\NUnitFramework\framework\Assert.cs:line 290
at NUnit.Framework.Assertions.Tests.AssertMultipleTests.MultipleAssertFailureDemo()
```

It would also be nice to clean up the extra newlines in the the "compatible" `<failure>` element so that older runners look okay. It looks like we have two or three newlines between blocks.

Other than that, looks good. Feel free to merge when ready.

CharliePoole:
@rprouse I like your idea. I thought about a further refinement which would eliminate the replication of the first line, but I think that's for the future, when we may have mixed assertion results listed under a test and some of the statuses don't match the overall status of the test.

src/NUnitFramework/nunitlite/TextUI.cs
```diff
@@ -482,19 +482,40 @@ private void DisplayNotRunResults(ITestResult result)
...
         private void DisplayTestResult(ITestResult result)
         {
-            string status = result.ResultState.Label;
+            ResultState resultState = result.ResultState;
+            string fullName = result.FullName;
+            string message = result.Message;
+            string stackTrace = result.StackTrace;
+            string reportID = (++_reportIndex).ToString();
+
+            //if (result.AssertionResults.Count > 0)
```
rprouse: I assume you didn't intend to keep this commented out. Is there another push coming?

CharliePoole: You're right and there is! I'm just experimenting with the "compatible" display there.

PR #1894, opened 2016-11-08
Fixups for DelayedConstraint tests

CharliePoole:
One of the new tests added with the recent syntax changes (Minutes, Seconds, PollEvery, etc.) keeps failing. It was requiring the test to succeed in too short a time frame. I intended to just adjust the delays, but discovered some other problems with the tests.

1. DelayedConstraintTests.cs had some rather long-running tests added in order to exercise the Seconds and Minutes options. Such tests don't really belong in that class. Normally, we test syntax separately (AfterTests in this case) from the actual execution tests of the constraint. I removed the long-running tests, including the one that was repeatedly failing.

2. AfterTests is set up to verify that the proper delay is set by the syntax. It does this by examining the string representation of the constraint. That string representation is not intended to be user-friendly, but is used for testing only. The string representation had been changed to include the time dimension (minutes, seconds, etc.) I changed it back to raw milliseconds. That way we test that `After(1).Minutes` causes a delay of 60000 milliseconds. This seems more certain than merely checking how the dimensioned interval displays itself.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
@@ -362,7 +362,7 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         protected override string GetStringRepresentation()
         {
-            return string.Format("<after {0} {1}>", DelayInterval, BaseConstraint);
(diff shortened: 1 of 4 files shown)


rprouse:
Looks good.

PR #1901, opened 2016-11-14
Make nunitlite-runner Prefer32Bit option consistent across Debug/Release

oneam:
I have interfaces with native code that are designed to only run 64bit mode. The current nunitlite-runner build options cause my tests to run in Debug but not in Release.

This change makes the two environments consistent.

### DIFF
--- a/src/NUnitFramework/nunitlite-runner/nunitlite-runner-2.0.csproj
+++ b/src/NUnitFramework/nunitlite-runner/nunitlite-runner-2.0.csproj
@@ -35,6 +35,7 @@
     <DefineConstants>TRACE;NET_2_0</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <Prefer32Bit>false</Prefer32Bit>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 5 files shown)


CharliePoole:
The code looks perfectly safe but I'd prefer to understand why it's needed. Nunitlite is pure IL and should run in 64 bits by default, shouldn't it?

oneam:
Pure IL code does run without any problems. The running assembly, however, is in either 32bit or 64bit mode. This only becomes apparent when you attempt to interface with managed DLLs (via PInvoke for example). With Prefer32Bit false you can only interface with a 64bit managed DLL (and vice versa).

In my particular case, I have a 64bit DLL but not 32bit. Prefer32bit true causes the tests to fail and false causes them to succeed. Having two different values for different versions of nunitlite-runner.exe was unexpected.

oneam:
Also, 64bit is not the default anymore (“Any CPU 32-bit preferred” is):

http://blogs.microsoft.co.il/sasha/2012/04/04/what-anycpu-really-means-as-of-net-45-and-visual-studio-11/

rprouse:
Looks good to me except you missed the portable project. This is a miss on our part, thanks for pointing it out.

oneam:
Oops, missed that one. I'll add it as well.
Would you prefer a swashed commit?

rprouse:
No need to squash. Looks good now. I will merge once CI is complete.

PR #1911, opened 2016-11-25
Changing misleading comment that implies that every ICollection<T> is a list

m-wilczynski:
Misleading because of `List<T>` from .NET Framework.
**Collection** is a broader term than **list** for anything that implements `ICollection<T>` (including for ex. `HashSet<T>`, `SortedSet<T>` etc.).


rprouse:
I'm okay with this. Normally we require a second reviewer, but no code was changed, so I am labeling and merging.

PR #1915, opened 2016-11-30
Enable Path, File and Directory Assert/Constraints in the .NET Standard Build

### LINKED ISSUE #1904: Add .NET Standard 1.6 Dependencies to the Nuspec Files
I just noticed that I didn't add a dependency on the .NET Standard libraries to the Nuspec files for the .NET Standard version of NUnit and NUnitLite.

Probably something like,

```xml
    <dependencies>
      </group>
    </dependencies>
```

rprouse:
This started off as just dropping the PORTABLE define from the .NET Standard builds so that we can start to re-enable some of the features that are not available in the PORTABLE version of the framework. As a part of that, I have documented what is still not enabled and will create issues to document or re-enable those that are remaining.

As a part of this, I enabled all of the file, path and directory based asserts that weren't available in PORTABLE. I don't have an issue for that, so I am marking this PR with the 3.6 milestone to track for the release.

The steps I took for this were,

1. Remove the PORTABLE define in all of the .NET Standard projects
2. Wherever possible, rewrite code that wasn't compiling to work in .NET Standard
3. As a last resort, add the needed `#if NETSTANDARD1_6` blocks and document the not supported functionality

This is a large PR because so many files were touched, but most of the changes are fairly rote.

I also updated the nuspec files which fixes #1904

### DIFF
--- a/src/NUnitFramework/nunitlite-runner/nunitlite-runner-netstandard.csproj
+++ b/src/NUnitFramework/nunitlite-runner/nunitlite-runner-netstandard.csproj
@@ -24,7 +24,7 @@
     <Optimize>false</Optimize>
     <OutputPath>..\..\..\bin\Debug\netstandard16\</OutputPath>
     <StartArguments>..\..\..\bin\Debug\netstandard16\nunitlite-runner.exe nunit.framework.tests.dll</StartArguments>
-    <DefineConstants>TRACE;DEBUG;PORTABLE;NETCOREAPP1_0</DefineConstants>
+    <DefineConstants>TRACE;DEBUG;NETCOREAPP1_0</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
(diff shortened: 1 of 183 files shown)


src/NUnitFramework/nunitlite-runner/nunitlite-runner-netstandard.csproj
```diff
@@ -49,12 +49,6 @@
     <Compile Include="..\FrameworkVersion.cs">
       <Link>Properties\FrameworkVersion.cs</Link>
     </Compile>
-    <Compile Include="..\nunitlite\ColorConsole.cs">
```
rprouse: These are now included in NUnitLite like in the other projects.

src/NUnitFramework/nunitlite/Options.cs
```diff
@@ -539,23 +539,6 @@ public override void GetObjectData (SerializationInfo info, StreamingContext con
 
     public class OptionSet : KeyedCollection<string, Option>
     {
-#if !PORTABLE
```
rprouse: We never use the localizer, so easier to easier to just use the PORTABLE null-op version for everything.

rprouse:
I have switched all of the async code to use `#if ASYNC` which is more explicit, like we do with `#if PARALLEL`

CharliePoole:
It looks good to me but I think you may want to merge in the latest from master manually. I believe I made some changes to defines that you don't have. I'm OK either way though.

rprouse:
I will pull in latest from master and make sure everything builds and tests pass before merging.

rprouse:
I've merge latest from master and ran all tests which are passing locally. This can be merged once CI passes.

CharliePoole:
I cancelled the Travis build so the netstandard branch could run. Everything was passing but it couldn't seem to find any Apple agents to run on.

CharliePoole:
Travis doesn't seem to be able to find Apple agents tonight, but the Linux jobs all passed so I'm going to go ahead and merge.

Issue #71, opened 2014-01-17
NUnit swallows console output from AppDomains created within tests


rationull:
I have a system test suite that uses NUnit, and the system under test creates an AppDomain internally during the test. Log messages within the system are routed to standard output (System.Console.Out) so that they'll get picked up in the build output in our continuous integration system (TeamCity).

NUnit (both the console and GUI runners) seems to swallow console output from the AppDomain created in the test. From a cursory look, it seems this has to do with the fact that Console.Out is redirected in the test AppDomain (presumably somewhere such that the output will end up in the process's console) but that redirection clearly doesn't happen in arbitrary AppDomains created by tests.

Not sure whether there's a clean solution for this. Assuming NUnit really does need to redirect Console.Out then arbitrary new AppDomains may just be out of luck.

I was able to work around this in our system by thunking console logging into the "parent" AppDomain via a proxy object, but that's kind of a heavyweight solution.

I have a repro Visual Studio solution that I can post if needed but there doesn't seem to be any way to attach files here.

Issue #42, opened 2013-10-28
NUnit does not show custom exceptions when running tests in a separate AppDomain


CharliePoole:
If a test throws one of our custom exception types (e.g., InterfaceBrokerException) and you're running tests via the NUnit GUI, then NUnit shows the following instead of the exception that was actually thrown:

An unhandled System.Runtime.Serialization.SerializationException was thrown while executing this test : Unable to find assembly 'Profitstar.Library, Version=2008.2.338.27793, Culture=neutral, PublicKeyToken=null'.

This is because the tests get run in one AppDomain, and then the results are marshaled to the main AppDomain via .NET serialization; but the Profitstar.Library assembly isn't loaded into the main AppDomain (nor should it be, because then it couldn't be unloaded), so it can't deserialize the exception.

So we end up with no stack trace, no original exception message, and no idea of even what exception type got thrown.

Is there any way to circumvent the use of a separate AppDomain when using the NUnit GUI? If not, can something be added to configure that feature?

See discussion at https://bugs.launchpad.net/nunit-3.0/+bug/494119

CharliePoole:
I don't believe this issue is relevant to NUnit 3 any longer. We still create a test AppDomain but exceptions thrown by the tests are no longer allowed to escape. The test execution code is contained in the framework itself and the exception is transformed into an NUnit TestResult in XML format before it is returned to the original AppDomain.

I'm closing the issue but if anyone thinks I have missed something, you can still add a comment here, which the developers will review.

PR #1916, opened 2016-12-01
Eliminate spurious AssertionResults from Assert.Throws

### LINKED ISSUE #1914: Extra AssertionResult entries in TestResults
There are a few situations where spurious AssertionResult entries are created in the result:

1. Assert.Throws or ThrowsConstraint when the code executed by the delegate argument includes asserts. The second-level asserts may leave failure and warning entries in the result that is being used for the top level test.

2. Executing code that causes an AssertionException and then catches it. This is done in some of our NUnit tests and it's a "trick" used by some users. The failure represented by the exception is added to the result before we can catch it.

3. Any Assert that takes a delegate could theoretically do the same thing if there were Asserts in the code executed. However, this is something of a pathological situation, since there isn't much reason to use Asserts in such delegates.

To resolve this, all such "second level" code should be executed in an isolated context and should apply to a separate result, which is ignored by the top-level code.

CharliePoole:
Fixes #1914 

This fixes the problem of second-level asserts, run under Assert.Throws or ThrowsConstraint, leaving AssertionResults for their failures in the current result. I've verified manually that the output of our test run doesn't include any `<assertion>` elements.

### DIFF
--- a/src/NUnitFramework/tests/Assertions/LowTrustFixture.cs
+++ b/src/NUnitFramework/tests/Assertions/LowTrustFixture.cs
@@ -61,7 +61,12 @@ namespace NUnit.Framework.Assertions
             });
         }
 
-        [Test, Platform(Exclude="Mono,MonoTouch", Reason= "Mono does not implement Code Access Security")]
+        //[Test, Platform(Exclude="Mono,MonoTouch", Reason= "Mono does not implement Code Access Security")]
+        // TODO: This test fails to run correctly when Assert.Throws
+        // attempts to create an isolated TestExecutionContext.
+        // The TestExecutionContext class would have to be modified
+        // in order to allow chaining of contexts across a 
+        // remoting boundary.
         public void AssertThrowsInLowTrustSandBox()
         {
             _sandBox.Run(() =>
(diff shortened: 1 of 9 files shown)


src/NUnitFramework/tests/Assertions/LowTrustFixture.cs
```diff
@@ -61,7 +61,12 @@ public void AssertEqualityInLowTrustSandBox()
             });
         }
 
-        [Test, Platform(Exclude="Mono,MonoTouch", Reason= "Mono does not implement Code Access Security")]
+        //[Test, Platform(Exclude="Mono,MonoTouch", Reason= "Mono does not implement Code Access Security")]
+        // TODO: This test fails to run correctly when Assert.Throws
+        // attempts to create an isolated TestExecutionContext.
+        // The TestExecutionContext class would have to be modified
+        // in order to allow chaining of contexts across a 
+        // remoting boundary.
```
rprouse: @CharliePoole do you think this is something that we should continue to support? Should an issue be created?

CharliePoole: I thought of a way to fix it so I will create an issue. We used to have a request for attributing tests to run in a separate domain, but I think we got rid of it. We also have a few issues around user-created AppDomains ( #42 and #71 ) which may have a similar cause to this. I'll put details in the issue.

PR #1919, opened 2016-12-01
Fix remoting issue with isolated context

### LINKED ISSUE #1917: Use of IsolatedContext breaks tests in user-created AppDomain
In fixing issue #1914, I had to comment out a test in LowTrustFixture.cs, which ran code in a separate AppDomain created by the user.

The problem is that the TestExecutionContext is marshalled by reference, while some of it's members are neither MBR nor serializable. Accessing any such member causes a remoting error. This is a consequence of the fact that we don't explicitly support running tests in a separate AppDomain but nothing stops a user from creating an AppDomain in their tests.

The particular test commented out uses Assert.Throws, which now creates an isolated context for the delegate it executes. Creating the context in a separate domain involves copying members across the remoting boundary. I think we can avoid this by using a public method of TestExecutionContext to create the isolated domain in the original context.

We have a few other issues that relate to problems in user-created AppDomains: #42 and #71. They may be due to the same phenomenon, so I'll look at them at the same time.

CharliePoole:
Fixes #1917 

This problem arose when the isolated context was being created in a separate, user-created AppDomain. The IsolatedContext constructor would try to create a new TestExecutionContext in the user AppDomain, initializing it with values from the original context in the original AppDomain. This caused an exception, since some fields that need to be copied are not serializable.

The fix puts everything back in the original AppDomain. The IsolatedContext calls a method on the original TestExecutionContext (which is MBR) so that __it__ will create the new context. Both the old and the new contexts are in the original AppDomain.

This solves the immediate problem but some things will still not work. If any user code tries to access a non-serializable member of the TestExecutionContext, an exception will be thrown. However, TestExecutionContext is an internal class, so I think this is acceptable.

Before merging, I want to look at how TestContext accesses TestExecutionContext, since that is more likely to be used by users.

### DIFF
--- a/src/NUnitFramework/tests/Assertions/LowTrustFixture.cs
+++ b/src/NUnitFramework/tests/Assertions/LowTrustFixture.cs
@@ -61,7 +61,7 @@ namespace NUnit.Framework.Assertions
             });
         }
 
-        //[Test, Platform(Exclude="Mono,MonoTouch", Reason= "Mono does not implement Code Access Security")]
+        [Test, Platform(Exclude="Mono,MonoTouch", Reason= "Mono does not implement Code Access Security")]
         // TODO: This test fails to run correctly when Assert.Throws
         // attempts to create an isolated TestExecutionContext.
         // The TestExecutionContext class would have to be modified
--- a/src/NUnitFramework/tests/Assertions/LowTrustFixture.cs
+++ b/src/NUnitFramework/tests/Assertions/LowTrustFixture.cs
@@ -62,11 +62,6 @@ namespace NUnit.Framework.Assertions
         }
 
         [Test, Platform(Exclude="Mono,MonoTouch", Reason= "Mono does not implement Code Access Security")]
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/tests/Assertions/LowTrustFixture.cs
```diff
@@ -61,7 +61,7 @@ public void AssertEqualityInLowTrustSandBox()
             });
         }
 
-        //[Test, Platform(Exclude="Mono,MonoTouch", Reason= "Mono does not implement Code Access Security")]
+        [Test, Platform(Exclude="Mono,MonoTouch", Reason= "Mono does not implement Code Access Security")]
         // TODO: This test fails to run correctly when Assert.Throws
         // attempts to create an isolated TestExecutionContext.
         // The TestExecutionContext class would have to be modified
```
rprouse: These comments are no longer relevant and should be removed as to not confused...

rprouse:
Looks good except for the comments that need to be removed.

Issue #1913, opened 2016-11-26
Report Warnings in NUnitLite


CharliePoole:
Warnings should be listed along with errors and failures.

Issue #406, opened 2014-12-05
Warning-level Assertions


CharliePoole:
Sometimes, you just want a warning! Please review the spec at https://github.com/nunit/docs/wiki/Warning-Level-Assert-Spec and comment here.

CharliePoole:
I'd like to come back to discussion of this issue. I just updated the spec at  https://github.com/nunit/docs/wiki/Warning-Level-Assert-Spec

Let's try to resolve whether we want to do this (I favor it) and if so pick a syntax and put it onto our backlog again.

ig-sinicyn:
@CharliePoole 
I'd expect warning assertions to be added into classic model API some day, so I vote for 'Expect.*' option.
This way `Expect.AreEqual()`, `Expect.Throws()` or `Expect.Contains()` will look fine.

In case we need more choices: `Assume.AreEqual()`, `Suspect.NotNull()` and `Guess.Greater()` to the rescue:)

CharliePoole:
@ig-sinicyn I particularly liked Expect myself, so I'm sorry to say it's already a synonym for Assert in the AssertionHelper class. In case you didn't know, Assume already exists with its own special meaning.

How about `Hope.That` or `Pray.That`. 😆

rprouse:
I like the idea of the feature, but I worry about how it will surface in the Visual Studio adapter or various CI servers. We have control over the CI adapter, so would we treat them something like ignored tests?

For the syntax, how about `IThink.That`? 😋 

`Warn.That` does not read well to me at all, I would use the alternative `Warn.If`. I would be okay with `Warn.Unless` too. I prefer the `Warn` keyword since its behaviour is explicit in its name. The other could be interpreted as synonyms for `Assert`.

CharliePoole:
@rprouse I think I agree with you that `Warn` is the right keyword. In fact, nothing prevents us from going all perl-like and supporting both `Warn.If` and `Warn.Unless`!

CharliePoole:
I updated the spec and I think we are ready to move take this out of design and add it to the backlog. Please check it over again @rprouse.

rprouse:
Looks good to me. On the runner side, I am assuming we will count warnings and display them in the summary block at the end of the run and that they will be displayed like ignores. I also assume they will be summarized at each level of the test result XML. Something like `passed="58" failed="0" warnings="12"`.

CharliePoole:
Good points. I'll add them to the spec.

CharliePoole:
[...]

PR #1920, opened 2016-12-06
Implement Warning result and assertions

### LINKED ISSUE #1912: Add new warning status and result state
This issue is preliminary to all other work on the Warning-level Assertions Epic.

To do, while making sure all tests continue to pass:
* Create a warning TestStatus
* Create a warning ResultState
* Ensure that all code that switches on either TestStatus or ResultState  takes the new values into account.
* Redefine the Ignored result state as a warning and change logic around ignored tests
* Create and Test Assert.Warn
* Create and Test Warn.If and Warn.Unless (same overloads as Assume.That)
* Ensure that warnings are all saved in the `<assertions>` element, with status Warning.

CharliePoole:
Fixes #1912 

See the issue itself for a list of what is included. 

Specifically excluded is the display of warnings by any of the runners. Other issues exist to deal with this after this PR is merged:

 * #1913 for NUnitlite
 * nunit/nunit-console#138 for the console runner
 * nunit/nunit3-vs-adapter#273 for the VS adapter


rprouse:
The code itself looks fine except for a few bits of commented out code that should be removed. I am a bit worried about changing `Skipped:Ignored` to `Warning:Ignored` though. We need to update all of our runners (including Xamarin and dotnet-test-nunit), but all third party runners will need to do the same won't they? There are also tools like VSTest, ReportUnit and other CI tools that parse our XML will need to support it.

Also, if any tool (nunit-console included) runs pre-3.6 versions of NUnit, they will still be reporting `Skipped:Ignored` which we won't be parsing or handling.

Also, if we run 3.6+ tests in an older runner (I just tested with 3.5), the ignored tests are no longer reported and Ignored reports 0.

Isn't changing our XML this way in-effect breaking our API?

CharliePoole:
@rprouse I was concerned about making that change too, which is why I paused work and waited for reactions to https://github.com/nunit/nunit/issues/406#issuecomment-263101955

It was a lot of work to implement this, and I would have much preferred that you had responded to that comment or said something when you reviewed the spec in the first place.

In my view, NUnit always treated Ignore as a warning, but in an adhoc way. We have been incorrectly presenting it as a simple Skip (i.e. simply not run without predjudice) for historical reasons. I see that as an error, which I wanted to fix.

I'm aware of the tradeoff of breaking code for somebody who is relying on the way we encode results. That's why I made a fairly big deal of spelling it all out out in the proposal before doing the work. The proposal was commented on by several third-party runner authors as well as by you and others.
 
If the general concensus turns out  to be that this is too much of a change, I'm willing to drop the PR, but can we try to do a better job of reviewing issues __before__ the effort of writing the code is done?

CharliePoole:
I'll hunt down the commented out code and remove.

CharliePoole:
NUnit-console still has to be updated, but by making the minimum of changes, it will count Skipped:Ignored as skipped and Warning:Ignored as a warning. The individual displays will list "Ignored" for both cases, because of the way the console status labels are determined. I was expecting to put Warnings into the Errors, Failures and Warnings (renamed) report, while Skipped:Ignored would continue to display in the Not Run report.

CharliePoole:
@nunit/core-team This seems like a good test case for the role of the Core Team. We have an issue that is not resolved within the nunit project in the usual way of concensus. I'm asking the Core Team members to take a look at it and decide what to do.

The spec for Warnings included a plan to re-categorize ignored tests as warnings. That is, `Result.Ignored` would be implemented as "Warning:Ignored" rather than "Skipped:Ignored" and would be recorded as a warning in the output. This was always a bit of an iffy change and I debated whether to include it in the spec, which I finally did. We all reviewed the spec and it's now implemented.

In his comment above, @rprouse gives his reasons for thinking we shouldn't make this change. I can see both sides of the question, but I'm a bit biased for having already written the code. To take it out will be a bit of effort, but that particular change is mostly found in one commit so it's not out of the question. Currently, this work is blocked for the lack of a decision.

There's a more general question behind the immediate one: what does it mean to have a "breaking change" in our XML. My own view has always been that changing the shape of the XML in a non-additive way is breaking. Changing the content of the fields is, in my view, acceptable. It's possible to take a more rigorous view, although I would argue it's difficult to achieve in practice when the format is as new as ours.

Since Rob and I have already weighed in I'm asking that the other Core Team members do the same:

1. Should we redefine Ignored as Warning:Ignored rather than Skipped:Ignored?

2. What should be considered breaking wrt the NUnit 3 result format.

rprouse:
@CharliePoole, I would also like to apologize for not groking the changes to ignore until the pull request. I read the spec including the changes to ignore and agreed with them, but I did not realize the consequences until I saw the code. 

I agree with the reasoning behind the change, but I think that we are going to get a flood of issue reports and disgruntled users when they find that ignored tests effectively disappear in older runners. We are only now, after a couple of years, getting runners like VSTest and Bamboo to support the NUnit 3 results format. I worry that they won't be making the changes again soon and we will bear the brunt of the support requests.

CharliePoole:
@rprouse As I said, I was always ready to go either way on this and even brought the question up twice. I got a bit tiffed when it only came up after I wrote the code, but I can see how that would happen.

This is not a change I'm wedded to. I think it's "right" in a certain sense but I think it will surprise many users if we don't list the ignored tests on the line of tests not run. However, I figured they would get used to it, since the warnings will still show up - just in a different place.

WRT third party folks, I'm not aware of anyone actually supporting the NUnit 3 format... AFAIK they are only beginning to work on it. During the lifetime of NUnit V2, we changed the values in various attribute fields many times. Our commitment then was to maintain the documented format - the overall shape of the file documented in its schema - rather than to guarantee that any values would always be the same.

A proper reading of the results by a runner would make allowances for any string value in the `label` field, since it is defined as open for redefinition by us or users. I suspect that most runners don't do that and even ours has special handling of "Ignored".

I don't think the overall Warning Assert thing rises and falls on this issue, but we should resolve it one way or another before anyone writes any more code on this.

BTW - any change in the fields can be mitigated in the reports if we want to do that. Of course, other runners would have to do the same thing.

oznetmaster:
All my tools use the NUnit 3 format. Never even looked at the NUnit 2 format.

ChrisMaddock:
Am I understanding correctly, that if we were to make the change, and runner which supports any NUnit 3.x framework would now have to look for both "Warning:Ignored" and "Skipped:Ignored"? If so, then yes, I'd agree that's a small piece of technical debt that's best being kept in the framework, rather than distributed out to all the runners. Sorry Charlie - you did highlight this before, I didn't entirely understand the consequences.

> A proper reading of the results by a runner would make allowances for any string value in the label field, since it is defined as open for redefinition by us or users.

I would hope they do make allowances for unknown test types. :-) My concern here would be that we're 'renaming' a 'known' test type. I think 'breaking' existing functionality for a 3rd party runner is a different matter to a runner not supporting new functionality.

> 2. What should be considered breaking wrt the NUnit 3 result format?

I think this would be much better handled looking at case-by-case, dependent of the cost/gain from making the change. Here, I think we can wrap up the slight oddity in the framework. Other cases that may be blockers on new functionality should be handled differently. There's of course cases which are definitely ok/definitely not, but a slight change to existing behaviour I'm not sure we can lay down a one-size-fits-all rule for.

CharliePoole:
@ChrisMaddock Let me try to spell out the change as it is and as it might be a little bit more clearly than I've done heretofore.

CURRENTLY: Each `<test-case>` (and `<test-suite>` result contains one to three fields that define the outcome of the test:
 * **result** is always present and may have one of four values: Passed, Failed, Inconclusive and Skipped.
 * **label** is optional and may have any string value
 * **site** is optional and may have one of the values: Test, SetUp, TearDown, Parent or Child
In talking about results, we sometimes use the ToString format to explain what we mean. If all three parts are present, the format is Result:Label(Site). Site is left out if it's Test, the default.

THIS PR:
 * **result** has been updated to allow a fifth value, Warning.

This is the biggest change and we don't appear to be debating about it. Runners that only look for failures simply won't see Warnings. I think they are forced to do some updating for this, since Warning propagates and the overall test run result can be Warning! At a minimum, they will have to decide to treat Warning as Success or Failure.

THIS PR (Additionally)
 * Ignored tests have been encoded in the past as `Skipped:Ignored` If we merge this, then they will be encoded as `Warning:Ignored`.

In a sense, this is minor, since it doesn't introduce anything new beyond the use of Warning. Labels have always been defined as an arbitrary string. However, most runners treat that string as "magic" and will probably be affected.

Our runner is not yet updated, but my intent is to extend the way it already works. Labels take precedence over results so Warning:Ignored and Skipped:Ignored will be listed in the same way... with a status of "Ignored". One hopes that other runners are equally aware of the openness of the Label field.

Where we have some "magic" in our runner is that we notice "Skipped:Ignored" and color it yellow. We'll continue to do that as well as coloring all Warnings the same way. Additionally, we list Skipped items in the Tests Not Run report. I anticipate that our own runner will work with pre-3.6 results just as it does now but will not list 3.6+ Ignored results in the Not Run report. Instead they would be in an Errors, Failures and Warnings report, that keeps all the red and yellow items together.

Obviously, we could use whatever magic we want to list the Ignored items where we think they belong, but the above makes the most sense to me, especially if mixed with a potential future "Treat warnings as errors options."

Third party runners have the same choices but how hard or easy it is for them will depend on how they are written in the first place.

This comment is intended to explain some things not to advocate for a particular solution. I do think that classifying Ignored tests as warnings makes a lot of sense, but it's not absolutely required for this feature.

rprouse:
> Ignored tests have been encoded in the past as Skipped:Ignored If we merge this, then they will be encoded as Warning:Ignored.

> In a sense, this is minor, since it doesn't introduce anything new beyond the use of Warning. Labels have always been defined as an arbitrary string. However, most runners treat that string as "magic" and will probably be affected.

This is the only part of the PR that worries me. I built a small test suite using this branch and ignored tests disappeared in all of the runners I tried it in, including the existing releases of the console. They didn't show up in the output and they didn't get reported as skipped.

It worries me because it took a couple of years before some CI systems like VSTO and Bamboo started supporting the NUnit 3 format and it will take awhile for 3rd party runners to update and support it. In the meantime, I expect that we will be fielding constant issues and questions around "Where did my ignored tests go?"

As a compromise, can we accept this PR minus switching ignore to warning and leave it as skipped?

CharliePoole:
@rprouse It makes sense that they disappeared in our runners, since we have issues outstanding to make it work in each of them. If third party runners or CI systems change to recognize Warnings, then the ignored tests will reappear for them as well.

Nevertheless, I take your point. It's a bigger loss to have ignored tests disappear than to have a new feature like warnings not show up at all. I'll change Ignored back to Skipped.

rprouse:
>@rprouse It makes sense that they disappeared in our runners, since we have issues outstanding to make it work in each of them.

I know that it would have been restored in the next release of our runners, but anyone running 3.6 tests with already release nunit console runners (<= 3.5) would have the ignored tests disappear.

CharliePoole:
@rprouse I agree that it's a good idea to switch back, but there's an interesting implication here.

We have never before guaranteed __forward__ compatibility of runners with framework versions, only __backward__.

It sounds as if we want to guarantee some degree - possibly limited - of forward compatibility as well. If so, maybe we should spell it out somewhere.

CharliePoole:
OK. I think this is ready to review and merge again. As a reminder, neither nunitlite nor the console will actually list warnings yet.

CharliePoole:
@nunit/framework-team @nunit/core-team Can somebody review this please?

rprouse:
Looks good, sorry it took so long to re-review...

PR #1925, opened 2016-12-11
Test the full framework using NUnit console

### LINKED ISSUE #1924: Run tests using the NUnit Console Runner
When we split the repositories, we switched the test runs from using the NUnit Console to using the NUnitLite runner. This works, but using the released version of the NUnit Console will provide a small measure of integration testing between the projects.

rprouse:
Fixes #1924 

This switches testing of the full framework tests for 2.0 through 4.5 over to using the console. PCL and .NET Standard are still tested using the NUnitLite Runner and NUnitLite is still self-testing.

My thinking on this is,

- This gives us a measure of integration testing with the released version of the console and engine
- It also allows us to test the Cake NUnit task
- The console project using the released framework allows a measure of integration testing in the other direction
- The NUnitLite runner is still tested using the PCL and .NET Standard targets
- NUnitLite continues to test itself

I also tested that we get useful error messages at the end of the run by adding a failing test. The messages still contain the number of failing tests and is very similar to the previous output.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -558,7 +558,10 @@ void RunNUnitTests(DirectoryPath workingDir, string testAssembly, string framewo
     try
     {
         var path = workingDir.CombineWithFilePath(new FilePath(testAssembly));
-        NUnit3(path.ToString());
+        var settings = new NUnit3Settings();
+        if(!IsRunningOnWindows())
+            settings.Process = NUnit3ProcessOption.InProcess;
+        NUnit3(path.ToString(), settings);
     }
     catch(CakeException ce)
     {
--- a/build.cake
+++ b/build.cake
@@ -1,6 +1,7 @@
 //////////////////////////////////////////////////////////////////////
 // ARGUMENTS
 //////////////////////////////////////////////////////////////////////
+#tool nuget:?package=NUnit.ConsoleRunner&version=3.5.0
 
 var target = Argument("target", "Default");
 var configuration = Argument("configuration", "Debug");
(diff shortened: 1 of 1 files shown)


CharliePoole:
Approach makes sense to me if the CI can be made to pass. I've had problems with using the NUnit3 cake extension under travis myself.

build.cake
```diff
@@ -558,7 +558,10 @@ void RunNUnitTests(DirectoryPath workingDir, string testAssembly, string framewo
     try
     {
         var path = workingDir.CombineWithFilePath(new FilePath(testAssembly));
-        NUnit3(path.ToString());
+        var settings = new NUnit3Settings();
+        if(!IsRunningOnWindows())
```
CharliePoole: This bothers me a little. Is it purely a limitation of the NUnit3 extension? Or does it matter what version of the runner we use with the tool?

rprouse: I added this in to see what would happen if we ran in-process on Linux. I looked at the failed Travis runs and it looked like there was an issue launching the agent, so I switched to in-process to see what would happen. I also figured that it is no different than running in-process in the nunitlite runner.

It would be interesting to try to debug this on a Linux box. Cake just spawns a process, so their could be a setting in `ProcessSettings` they are missing, or it could be a Mono problem?

I am not wed to this, I could also just spawn the console like we used to.

rprouse:
@CharliePoole would you like me to look into the problems with Cake on Linux more, switch it to a straight nunit3-console call, or leave it as is?

CharliePoole:
I can go either way. If you leave it as is, I think we should add a new issue to figure it out later.

rprouse:
I created #1926 to investigate why we can't run inprocess on Linux using Cake.

PR #1929, opened 2016-12-14
Rename zip and remove source zip

ChrisMaddock:
Relates to https://github.com/nunit/nunit-distribution/pull/11. This one's optional, but general consensus seemed to be that it was a good idea. 

This renames the framework zip to `NUnit.Framework.{Version}`, to avoid ambiguity. See general notes on https://github.com/nunit/nunit-distribution/pull/11.

I also removed the source zip at the same time. We stopped distributing this previously in favour of GitHub's automated version, and the task wasn't hooked up to everything. The equivalent has already been removed in the console project.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -63,8 +63,7 @@ var FRAMEWORK_TESTS = "nunit.framework.tests.dll";
 var EXECUTABLE_NUNITLITE_TESTS = "nunitlite.tests.exe";
 
 // Packages
-var SRC_PACKAGE = PACKAGE_DIR + "NUnit-" + version + modifier + "-src.zip";
-var ZIP_PACKAGE = PACKAGE_DIR + "NUnit-" + packageVersion + ".zip";
+var ZIP_PACKAGE = PACKAGE_DIR + "NUnit.Framework." + packageVersion + ".zip";
 
 bool isDotNetCoreInstalled = false;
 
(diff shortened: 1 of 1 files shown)


CharliePoole:
I'd prefer to keep the hyphen between the  package name and version.

Also, while we are thinking about zips... we used to go to a lot of trouble to ensure that each zip had a single root folder named after the package itself, e.g. nunit.framework-1.2.3. Aren't we doing that any longer?

And while I'm at it :smile: are we abandoning the standard that distributed packages are named in lower case? I realize we have made a few exceptions over the years, so that too has become inconsistent. I just hate to see us kind of creep from one way of doing things to another, rather than just making a group decision.

ChrisMaddock:
Happy with hyphens, see https://github.com/nunit/nunit-distribution/pull/11. Regarding the other points, we could do both - although neither seem to have been in place since back before 3.0. From a quick check, it looks like both would also need to be implemented across all repo's - adapter and console both seem to also use SentenceCase and no root folder in the zip.

I'd be tempted to say let's change the convention, and save ourselves some work in changing all the build scripts!

CharliePoole:
I guess when you keep making exceptions, it becomes unclear what the standard is or whether there is a standard. Let's see what @rprouse thinks.

ChrisMaddock:
Ok, change made as discussed on https://github.com/nunit/nunit-distribution/pull/11 😄

PR #1935, opened 2016-12-21
Pass zero-case tests

### LINKED ISSUE #1933: Tests should pass if test case source provides 0 test cases
I'm in a situation where I need to run a set of integration tests sets against multiple databases. It's been _extremely_ helpful to use TestCaseSource to give separate test cases for each object in certain database tables. In fact, this is what motivated the move to NUnit in the first place.

The only issue is that it's perfectly valid for tables to be empty.  In that scenario, I would expect that no test cases would exist or be run. Surprisingly, and problematically, even though I'm returning an empty `IEnumerable<TestCaseData>` from the test case source, NUnit still attempts to run the test anyway as a single case with no arguments. And of course the test fails, since the arguments are necessary.

Is this helpful behavior or a bug? If this is helpful behavior to some people, can I globally configure NUnit to always consider 0 test cases as 0 test cases which matches my intuition and needs?

jnm2:
Closes #1933.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TheoryAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TheoryAttribute.cs
@@ -21,6 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 using System;
+using NUnit.Framework.Internal;
 
 namespace NUnit.Framework
 {
@@ -49,7 +50,7 @@ namespace NUnit.Framework
     /// </example>
     /// 
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited=true)]
-    public class TheoryAttribute : CombiningStrategyAttribute, ITestBuilder, IImplyFixture
+    public class TheoryAttribute : CombiningStrategyAttribute, ITestPostProcessor, IImplyFixture
     {
         /// <summary>
         /// Construct the attribute, specifying a combining strategy and source of parameter data.
@@ -58,6 +59,20 @@ namespace NUnit.Framework
             new CombinatorialStrategy(),
             new ParameterDataProvider(new DatapointProvider(), new ParameterDataSourceProvider()))
         {
+        }
+
+        /// <summary>
+        /// Validate or otherwise process a test or suite which has just been built, possibly by multiple builders.
+        /// </summary>
+        /// <param name="test">The test or suite which has just been built.</param>
+        public void AfterBuild(Test test)
+        {
+            var suite = test as TestSuite;
+            if (suite != null && suite.TestCaseCount == 0)
+            {
+                suite.RunState = RunState.NotRunnable;
+                suite.Properties.Add(PropertyNames.SkipReason, "Theory has no test cases");
+            }
         }
     }
 }
--- a/src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
@@ -119,35 +119,46 @@ namespace NUnit.Framework.Internal.Builders
             List<ITestBuilder> builders = new List<ITestBuilder>(
                 method.GetCustomAttributes<ITestBuilder>(false));
 
-            // See if we need a CombinatorialAttribute added
-            bool needCombinatorial = true;
+            var isSuite = false;
+
             foreach (var attr in builders)
             {
-                if (attr is CombiningStrategyAttribute)
-                    needCombinatorial = false;
+                var buildAction = attr.BuildFrom(method, parentSuite);
+                if (buildAction.IsSuite)
+                {
+                    isSuite = true;
+                    foreach (var test in buildAction.SuiteTests)
+                        tests.Add(test);
+                }
             }
 
-            // This check must be done before CombinatorialAttribute gets added to the builders collection
-            var hasBuildersSpecified = builders.Count > 0;
-
+            
             // We could check to see if here are any data attributes specified
             // on the parameters but that's what CombinatorialAttribute does
             // and it simply won't return any cases if it finds nothing.
             // TODO: We need to add some other ITestBuilder than a combinatorial attribute
             // because we want the attribute to generate an error if it's present on
             // a generic method.
-            if (needCombinatorial)
-                builders.Add(new CombinatorialAttribute());
-
-            foreach (var attr in builders)
+            if (!builders.OfType<CombiningStrategyAttribute>().Any())
             {
-                foreach (var test in attr.BuildFrom(method, parentSuite))
-                    tests.Add(test);
+                var buildAction = new CombinatorialAttribute().BuildFrom(method, parentSuite);
+                if (buildAction.SuiteTests.Count != 0) // Since this attribute was not specified by the user, only turn into a suite if there is at least one test case
+                {
+                    isSuite = true;
+                    foreach (var test in buildAction.SuiteTests)
+                        tests.Add(test);
+                }
             }
 
-            return hasBuildersSpecified || tests.Count > 0
+
+            var builtTest = isSuite
                 ? BuildParameterizedMethodSuite(method, tests)
                 : BuildSingleTestMethod(method, parentSuite);
+
+            foreach (var postProcessor in method.GetCustomAttributes<ITestPostProcessor>(false))
+                postProcessor.AfterBuild(builtTest);
+
+            return builtTest;
         }
 
         #endregion
--- a/src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
@@ -127,6 +127,9 @@ namespace NUnit.Framework.Internal.Builders
                     needCombinatorial = false;
             }
 
+            // This check must be done before CombinatorialAttribute gets added to the builders collection
+            var hasBuildersSpecified = builders.Count > 0;
+
             // We could check to see if here are any data attributes specified
             // on the parameters but that's what CombinatorialAttribute does
             // and it simply won't return any cases if it finds nothing.
@@ -142,7 +145,7 @@ namespace NUnit.Framework.Internal.Builders
                     tests.Add(test);
             }
 
-            return tests.Count > 0 || builders.OfType<TestCaseSourceAttribute>().Any()
+            return hasBuildersSpecified || tests.Count > 0
                 ? BuildParameterizedMethodSuite(method, tests)
                 : BuildSingleTestMethod(method, parentSuite);
         }
--- a/src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal.Commands;
 
@@ -141,7 +142,7 @@ namespace NUnit.Framework.Internal.Builders
                     tests.Add(test);
             }
 
-            return tests.Count > 0
+            return tests.Count > 0 || builders.OfType<TestCaseSourceAttribute>().Any()
                 ? BuildParameterizedMethodSuite(method, tests)
                 : BuildSingleTestMethod(method, parentSuite);
         }
--- a/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
@@ -290,6 +290,21 @@ namespace NUnit.Framework.Attributes
             Assert.AreEqual(lhs, rhs);
         }
 
+
+        private static IEnumerable<TestCaseData> ZeroTestCasesSource() => Enumerable.Empty<TestCaseData>();
+
+        [TestCaseSource("ZeroTestCasesSource")]
+        public void TestWithZeroTestSourceCasesShouldNotBeRun()
+        {
+            Assert.Fail();
+        }
+
+        [TestCaseSource("ZeroTestCasesSource")]
+        public void TestWithZeroTestSourceCasesShouldPassWithoutRequiringArguments(int requiredParameter)
+        {
+        }
+
+
         static object[] testCases =
         {
             new TestCaseData(
--- a/src/NUnitFramework/tests/Attributes/TheoryTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TheoryTests.cs
@@ -44,13 +44,15 @@ namespace NUnit.Framework.Attributes
         [Test]
         public void TheoryWithNoDatapointsIsNotRunnable()
         {
-            TestAssert.IsNotRunnable(fixtureType, "TheoryWithArgumentsButNoDatapoints");
+            Test test = TestBuilder.MakeParameterizedMethodSuite(fixtureType, "TheoryWithArgumentsButNoDatapoints");
+            TestAssert.IsNotRunnable(test);
         }
 
         [Test]
         public void UnsupportedNullableTypeArgumentWithNoDatapointsAreNotRunnable()
         {
-            TestAssert.IsNotRunnable(fixtureType, "TestWithUnsupportedNullableTypeArgumentWithNoDataPoints");
+            Test test = TestBuilder.MakeParameterizedMethodSuite(fixtureType, "TestWithUnsupportedNullableTypeArgumentWithNoDataPoints");
+            TestAssert.IsNotRunnable(test);
         }
 
         [Test]
--- a/src/NUnitFramework/tests/Attributes/TheoryTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TheoryTests.cs
@@ -56,7 +56,7 @@ namespace NUnit.Framework.Attributes
         }
 
         [Test]
-        public void UnsupportedNullableTypeArgumentWithNoDatapointsAreNotRunnable()
+        public void UnsupportedNullableTypeArgumentWithNoDatapointsIsNotRunnable()
         {
             Test test = TestBuilder.MakeParameterizedMethodSuite(fixtureType, "TestWithUnsupportedNullableTypeArgumentWithNoDataPoints");
             TestAssert.IsNotRunnable(test);
--- a/src/NUnitFramework/tests/Attributes/TheoryTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TheoryTests.cs
@@ -49,6 +49,13 @@ namespace NUnit.Framework.Attributes
         }
 
         [Test]
+        public void TheoryWithTestCasesButNoDatapointsIsRunnable()
+        {
+            Test test = TestBuilder.MakeParameterizedMethodSuite(fixtureType, "TheoryWithTestCasesButNoDatapoints");
+            TestAssert.IsRunnable(test);
+        }
+
+        [Test]
         public void UnsupportedNullableTypeArgumentWithNoDatapointsAreNotRunnable()
         {
             Test test = TestBuilder.MakeParameterizedMethodSuite(fixtureType, "TestWithUnsupportedNullableTypeArgumentWithNoDataPoints");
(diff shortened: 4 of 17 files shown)


src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -290,6 +290,21 @@ public void HandlesExceptionInTestCaseSource_GuiDisplay(string lhs, string rhs)
...
+        public void Test_with_zero_test_cases_should_not_be_run()
+        {
+            Assert.Fail("Test should not be run.");
+        }
+
+        [TestCaseSource(nameof(ZeroTestCasesSource))]
+        public void Test_with_zero_test_cases_should_pass_without_requiring_arguments(int requiredParameter)
+        {
+        }
+
```
jnm2: Again, I'm assuming this is the correct goal.

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -290,6 +290,21 @@ public void HandlesExceptionInTestCaseSource_GuiDisplay(string lhs, string rhs)
...
+
```
jnm2: Again, I'm assuming this is the correct goal.

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
```diff
@@ -141,7 +142,7 @@ public Test BuildFrom(IMethodInfo method, Test parentSuite)
                     tests.Add(test);
             }
 
-            return tests.Count > 0
+            return tests.Count > 0 || builders.OfType<TestCaseSourceAttribute>().Any()
```
CharliePoole: Much cleaner. Ideally, this code should not know the names of any concrete ITestBuilder types, although it does.

An alternative may be to just check if builders.Count > 0.

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
CharliePoole: Well, #1933 is an unreviewed new issue, so we don't actually have a goal yet. :smile:

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
jnm2: With `builders.Count != 0` or `builders.OfType<ITestBuilder>`, would that affect theories?

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -290,6 +290,21 @@ public void HandlesExceptionInTestCaseSource_GuiDisplay(string lhs, string rhs)
             Assert.AreEqual(lhs, rhs);
         }
 
+
+        private static IEnumerable<TestCaseData> ZeroTestCasesSource() => Enumerable.Empty<TestCaseData>();
+
+        [TestCaseSource(nameof(ZeroTestCasesSource))]
+        public void Test_with_zero_test_cases_should_not_be_run()
```
immeraufdemhund: I might be out of line here, but the rest of this test class does not use underscores between words, so I think if/when this gets accepted you might want to consider renaming the methods to match the convention in the rest of the test class.

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -290,6 +290,21 @@ public void HandlesExceptionInTestCaseSource_GuiDisplay(string lhs, string rhs)
...
+        {
+            Assert.Fail("Test should not be run.");
+        }
+
+        [TestCaseSource(nameof(ZeroTestCasesSource))]
+        public void Test_with_zero_test_cases_should_pass_without_requiring_arguments(int requiredParameter)
```
immeraufdemhund: Same comment as above

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -290,6 +290,21 @@ public void HandlesExceptionInTestCaseSource_GuiDisplay(string lhs, string rhs)
...
+
+        private static IEnumerable<TestCaseData> ZeroTestCasesSource() => Enumerable.Empty<TestCaseData>();
+
+        [TestCaseSource(nameof(ZeroTestCasesSource))]
+        public void Test_with_zero_test_cases_should_not_be_run()
+        {
+            Assert.Fail("Test should not be run.");
+        }
+
+        [TestCaseSource(nameof(ZeroTestCasesSource))]
```
immeraufdemhund: same as above comment about name of

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -290,6 +290,21 @@ public void HandlesExceptionInTestCaseSource_GuiDisplay(string lhs, string rhs)
             Assert.AreEqual(lhs, rhs);
         }
 
+
+        private static IEnumerable<TestCaseData> ZeroTestCasesSource() => Enumerable.Empty<TestCaseData>();
+
+        [TestCaseSource(nameof(ZeroTestCasesSource))]
```
immeraufdemhund: I know in other projects we said it was OK to use C# 6.0 features, but I don't remember if it was acceptable for the framework itself. Namely you might need to change from nameof to a hardcoded string.

jnm2: All tests pass and all builds succeed, and that's really all that matters, right?

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
jnm2: No, good comment and I did it intentionally for the POC. @CharliePoole will probably agree with you?

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
jnm2: With `builders.Count != 0`, wouldn't that affect theories?
These are all the things that `builders` may contain:
 * **`TheoryAttribute`**
 * `SequentialAttribute`
 * `PairwiseAttribute`
 * `CombinatorialAttribute`
 * `TestCaseSourceAttribute`
 * `TestCaseAttribute`
 * Future implementations of `ITestBuilder`

jnm2: Alternatively, we could introduce `ITestBuilder.FailIfNoCases` and check `builders.Any(_ => _.FailIfNoCases)`.

jnm2: Alright. Both `builders.Count > 0` and `(builders.Count > 0 && !builders.OfType<TheoryAttribute>().Any())` make 134 tests fail with:

>   Expected: <NUnit.Framework.Internal.TestMethod>
>   But was:  <NUnit.Framework.Internal.ParameterizedMethodSuite>

jnm2: @CharliePoole Lots of failures at [`TestBuilder.MakeTestCase`'s  `Assert.That(test, Is.TypeOf<TestMethod>());`](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/TestBuilder.cs#L82) because it's actually getting a `ParameterizedMethodSuite`.

jnm2: Because all these tests have `CombinatorialAttribute`.
Oh! Lol! We're getting nailed by https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs#L136.

`builders.Count` will just have to be checked before `CombinatorialAttribute` is added to the builders collection.

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
```diff
@@ -141,7 +145,7 @@ public Test BuildFrom(IMethodInfo method, Test parentSuite)
                     tests.Add(test);
             }
 
-            return tests.Count > 0
+            return shouldParameterizeForFewerThanTwoTests || tests.Count > 0
```
jnm2: @CharliePoole continuing our conversation, this commit will cause `TheoryWithNoDatapointsIsNotRunnable` and `UnsupportedNullableTypeArgumentWithNoDatapointsAreNotRunnable` to fail with:

>   Expected: <NUnit.Framework.Internal.TestMethod>
>   But was:  <NUnit.Framework.Internal.ParameterizedMethodSuite>

I assume this isn't just overzealous testing in these two tests?

These are all the things that `builders` may contain:
 * **`TheoryAttribute`**
 * `SequentialAttribute`
 * `PairwiseAttribute`
 * `CombinatorialAttribute`
 * `TestCaseSourceAttribute`
 * `TestCaseAttribute`
 * Future implementations of `ITestBuilder`

Alternatively, we could introduce `ITestBuilder.FailIfNoCases` and check `builders.Any(_ => _.FailIfNoCases)`. `TheoryAttribute` would be the only current test builder to return true for `FailIfNoCases`.

jnm2: Decided it was not harming the tests to make them expect parameterization.

src/NUnitFramework/tests/Attributes/TheoryTests.cs
```diff
@@ -44,13 +44,15 @@ public void TheoryWithNoArgumentsIsTreatedAsTest()
...
+            Test test = TestBuilder.MakeParameterizedMethodSuite(fixtureType, "TheoryWithArgumentsButNoDatapoints");
+            TestAssert.IsRunnable(test);
         }
 
         [Test]
         public void UnsupportedNullableTypeArgumentWithNoDatapointsAreNotRunnable()
         {
-            TestAssert.IsNotRunnable(fixtureType, "TestWithUnsupportedNullableTypeArgumentWithNoDataPoints");
+            Test test = TestBuilder.MakeParameterizedMethodSuite(fixtureType, "TestWithUnsupportedNullableTypeArgumentWithNoDataPoints");
+            TestAssert.IsRunnable(test);
```
jnm2: Decided it was not harming the tests to make them expect parameterization.

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
CharliePoole: @jnm2 Using C# 6 features means somebody can't build the framework using VS2012 (or 2013?) It's likely we will soon enough change our coding standard to allow require VS2015+ or equivalent, but we haven't done that yet. Rather than starting the discussion here, let's just go back to the use of constants, as we do everywhere.

[FYI @immeraufdemhund is referring to the fact that we did already make that change for the NUnit Gui, but not for any other software.]

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
CharliePoole: Yes, although I missed it, it's best to follow our standards.

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
CharliePoole: This seems slightly complicated to me, since I'm still thinking that checking

```C#
if (tests.Count > 0 && builders.Count > 0)
...
```
won't do the job.

CharliePoole: __will__ do the job!

jnm2: @CharliePoole we cannot do `if (tests.Count > 0 && builders.Count > 0)` due to this line: https://github.com/jnm2/nunit/blob/56283ed961bdb9f685864408208bd121c3132d2b/src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs#L140

For reasons I hope are obvious. 134 tests fail.

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
jnm2: You know it is possible to set a max C# language level in each .csproj and that way all our tooling won't bug us to use good C# 6 style.

CharliePoole: Well, if I'd only known that! Of course, we set this standard a long time back and it might not have been available. In any case, I agree with your logic. But since we'r making a real effort to create and empower a committee that will run NUnit in the future, I'd really like to wait for them to change the policy. Sorry for the trouble - it is in the coding standard, however.

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
CharliePoole: How about changing the variable to something like `originalBuildersCount`?

The name is confusing to me on several levels, especially the use of "Two" rather than "One".

jnm2: Okay. I figured I'd make it describe the decision being made, but I'm okay with that.

jnm2: What about `shouldForceParameterization`?

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
```diff
@@ -135,15 +139,33 @@ public Test BuildFrom(IMethodInfo method, Test parentSuite)
...
-                    tests.Add(test);
+                try
+                {
+                    foreach (var test in attr.BuildFrom(method, parentSuite))
+                        tests.Add(test);
+                }
+                catch (Exception ex)
+                {
+                    exceptions.Add(ex);
+                }
```
jnm2: Handling exceptions for this purpose has precedent in NUnit. The other option is to return more info from BuildFrom, or somehow give BuildFrom an instance of the test suite so it can be modified.

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
```diff
@@ -126,6 +127,9 @@ public Test BuildFrom(IMethodInfo method, Test parentSuite)
                     needCombinatorial = false;
             }
 
+            // This check must be done before CombinatorialAttribute gets added to the builders collection
+            var shouldForceParameterization = builders.Count > 0;
```
jnm2: This entire method would get simpler if we weren't running the CombinatorialAttribute on single tests in case they have parameters. We could decide up front whether the test would be a suite. Perhaps based on whether it has parameters and whether it has builders.

src/NUnitFramework/framework/Attributes/TheoryAttribute.cs
```diff
@@ -59,5 +62,22 @@ public class TheoryAttribute : CombiningStrategyAttribute, ITestBuilder, IImplyF
...
+        /// </summary>
+        /// <param name="method">The MethodInfo for which tests are to be constructed.</param>
+        /// <param name="suite">The suite to which the tests will be added.</param>
+        /// <returns>One or more TestMethods</returns>
+        public override IEnumerable<TestMethod> BuildFrom(IMethodInfo method, Test suite)
+        {
+            var cases = base.BuildFrom(method, suite).ToList();
+
+            if (cases.Count == 0 && IsTreatedAsSuite(method))
+                throw new InvalidOperationException("There are no test cases");
```
jnm2: No problem. I wanted you to see this.
The next option I think is the simplest is `bool ITestFixture.RequireTestCases`. It's also much cleaner, no exception handling. I'll demonstrate that next.

jnm2: Or even more powerful, `void ITestBuilder.PostProcess(Test suite)`.

jnm2:
@CharliePoole 

> Even though it's all passing, the change to TheoryAttribute won't work. My suggestion is to drop the last two commits, go back to commit eb4c6f7 and push -f. That will get us back to where everything worked except for Theories. Then we can take another look at fixing them.

Well then, let me add a test so that it isn't passing. :-)

CharliePoole:
Well there were two failures back at that commit.

jnm2:
> Well there were two failures back at that commit.

Yes. I'll add a test that would have prevented this commit from passing so that we have clearer goals when we start from eb4c6f7. Since it's a requirement that wasn't codified, I'll codify it. =)

CharliePoole:
So are you not backing out the last two commits then?

CharliePoole:
In case you're worrying about "changing history" we generally have not considered branches as "published" until they are merged. It's a slightly less pure approach to Git usage, but it seems to work better for us and keeps PRs in a state where they are easier to review.

jnm2:
> So are you not backing out the last two commits then?

I am, but I don't have the time at the moment. I expect to do this Monday. I also like changing history. =) Probably too much.

> What's the uncodified requirement? Clearly the two failing Theory tests are codified.

This one:

> Sorry... that changes the semantics of the TheoryAttribute. If it finds no cases, that means there are no DataPoint values in the class. But that isn't the only way that data can be provided - it's just the only one that TheoryAttribute knows about. The only point where it all comes together is in the code that loops through all test builders, where you started working initially.
> 
> We don't have a test for this situation, because it has never come up. However, a Theory that finds no Datapoints but has cases provided through TestCaseAttribute, TestCaseSourceAttribute, RandomAttribute, etc. is perfectly valid.

CharliePoole:
Good point! I was thinking of it as an error that couldn't come up in a normal way... but I guess you trying to fix a bug is pretty normal. 😄 It's not a test that came up in TDDing TheoryAttribute, because "Generate an error if there is no data" just wasn't a part of the Theory stories.

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
jnm2: I could see valid future scenarios for builders without parameters. And for parameters without builders, we simply create the single test and let the usual error occur. I do not fully understand the comments written with the TODO so I'll leave this alone for now.

jnm2: My instinct would be to check wither the test method has parameters. If not, return a single, If so, return a suite. But you might know of a scenario where that won't work.

jnm2:
I think this latest solution is pretty clean.
There's one outstanding test failing that I need to look more into to see if the test needs to be corrected or if it's pointing to an implementation oversight. That is `.TestMethodSignatureTests.RunningTestsThroughFixtureGivesCorrectResults`, where the summary is off because `TestMethodSignatureFixture.TestMethodWithArgumentsNotProvidedAndExtraAttributes` is getting skipped and it didn't used to.

jnm2:
I think this latest solution is pretty clean. No failing tests.

jnm2:
Since we've decided to change theories with no cases from not being runnable to being runnable but fail, https://github.com/nunit/nunit/pull/1940 supersedes this PR.

CharliePoole:
If you prefer to close it that's your right. At the moment I can still see and make use of the changes, which was the reason I asked to keep it open. I'm not sure how github handles this with forks in case you delete your branch however.

jnm2:
Did you want it open or just in existence? I don't tend to like long-running PRs in my list anymore than repo owners 😆 Ideally this PR and branch would stay as-is in case of future reference, but it's unlikely that anyone will edit or merge it so I guess it doesn't matter.

CharliePoole:
As is is fine. I'd just like to be able to address some of your changes in the new PR and I don't know what GitHub will do if you delete the branch in your fork. As far as long-running, it has only been a matter of a day so far. 😄

jnm2:
I just protected the branch so that I don't accidentally hit the big Delete Branch button. (Like I do with the close and reopen comment buttons -_-)

Not that it matters, it's been eight days from my perspective. I'm the kind of person that gets a really good feeling every time I throw something away. 😆

CharliePoole:
I meant since I posted the new PR, but I take your point.

PR #1937, opened 2016-12-23
Extended TestContext.TestAdapter with IMethodInfo Method property to obtain access to test metadata

Suremaker:
Hello,

I have created a pull request extending the `TestContext.TestAdapter` class with `IMethodInfo Method` property.

The reason for that change is that I would like to get access to the test method metadata, such as:
* name,
* parameter types information
* attributes applied to method or method parameters.

While currently it would be possible to use `TextContext.Test.FullName` and then use reflection to findout the MethodInfo, such implementation would be problematic for overloaded test methods as well as more complicated to implement.

Thanks

### DIFF
--- a/src/NUnitFramework/framework/TestContext.cs
+++ b/src/NUnitFramework/framework/TestContext.cs
@@ -353,6 +353,11 @@ namespace NUnit.Framework
             }
 
             /// <summary>
+            /// The method representing the test, which may be null if test is not implemented as a method.
+            /// </summary>
+            public IMethodInfo Method { get { return _test.Method; } }
+
+            /// <summary>
             /// The FullName of the test
             /// </summary>
             public string FullName
--- a/src/NUnitFramework/tests/TestContextTests.cs
+++ b/src/NUnitFramework/tests/TestContextTests.cs
@@ -138,7 +138,22 @@ namespace NUnit.Framework.Tests
         {
             Assert.That(TestContext.CurrentContext.Test.MethodName, Is.EqualTo("TestCanAccessItsOwnMethodName"));
         }
-        
+
+        [Test]
+        public void TestCanAccessItsOwnMethod()
+        {
+            Assert.That(TestContext.CurrentContext.Test.Method, Is.Not.Null);
+            Assert.That(TestContext.CurrentContext.Test.Method.Name, Is.EqualTo("TestCanAccessItsOwnMethod"));
+        }
+
+        [Test]
+        [Description("")]
+        public void TestCanAccessItsOwnMethodAttributes()
+        {
+            Assert.That(TestContext.CurrentContext.Test.Method, Is.Not.Null);
+            Assert.That(TestContext.CurrentContext.Test.Method.GetCustomAttributes<DescriptionAttribute>(false), Is.Not.Empty);
+        }
+
         [TestCase(5)]
         public void TestCaseCanAccessItsOwnMethodName(int x)
         {


CharliePoole:
I'm dubious about this as a desirable feature. TestContext provides the TestAdapter class rather than the "real" test, as a way to make sure that user tests don't get involved with internals. I can't think of a good reason why a test would need to have this info, unless it's a test of NUnit itself.

Can you give a use case for needing this information __in your test__? I bolded that because I understand you might need it in a test runner or a test logger etc. But IMO those functions don't really belong in a test and there are other ways to perform them outside of the test code.

Suremaker:
Sure, I developed [LightBDD](https://github.com/Suremaker/LightBDD) framework allowing to write BDD tests in C#, on top of NUnit and other frameworks.

The test scenarios are written in form of test method like below:
```c#
[Test]
[Label("Ticket-1")]
public void Successful_login() //scenario name
{
    Runner.RunScenario(
        Given_user_is_about_to_login, //steps
        Given_user_entered_valid_login,
        Given_user_entered_valid_password,
        When_user_clicked_login_button,
        Then_login_is_successful,
        Then_welcome_message_is_returned_containing_user_name);
}
```

Full example can be found [here]( https://github.com/Suremaker/LightBDD/blob/master/README.md#example).

When test is executed, the runner uses test method metadata (name, attributes and possibly parameters) to render scenario on console during test execution as well as to build the test report after tests (the example test report is [here](http://htmlpreview.github.io/?https://github.com/Suremaker/LightBDD/blob/master/ExampleSummaryFiles/FeaturesSummary.html)).

In current version of LightBDD for NUnit 2x and 3x integration I am using stack trace to find the test method in order to interpret scenario details. The implementation of Runner.RunScenario () is calling following code to get test method:  `new StackTrace().GetFrame(2).GetMethod();`

The problem with current implementation is that:
* this functionality of StackTrace is not supported in netstandard, so I cannot write multiplatform version of LightBDD.NUnit,
* it is rather a workaround than a proper implementation because this code tries to infer the test method, while it is well known to NUnit engine.

To solve the issue I thought I will extend the TestContext as it is used to access the currently executed test details (it already offers access to method name, class name and even passed parameters), so I added the property allowing to retrieve MethodInfo of the test method.
It eliminates any ambiguity that would raise when only class and method name is used.

I hope that context shows better why I want to make the change. If there is a better way to do it I would be more than happy to try it out but I am aiming for the solution that will be seamless in usage from user perspective.

Best regards

CharliePoole:
Looking at the history, it's easy to see the "feature creep" here. I added ClassName myself, I believe because we already had MethodName. MethodName had previously been added, but I can't find an issue that asked for it. 😢 

From what you are saying, it sounds as if you need it for LightBDD itself, but that your users really don't have a need to access the field. That being the case, as an extension on top of NUnit, I think it's perfectly reasonable for you to access the TestExecutionContext directly. It's public in the NUnit.Framework.Internal namespace for exactly that reason.

Note that this is just my personal opinion. You may want to pause a beat to see if other @nunit/core-team members have a different view.

Suremaker:
Thank you for quick reply and suggestion. 

That is totally true that regular tests should not need to know such details, and it is only needed by frameworks written on top of nunit.
Currently, I do not have access to computer, but will try this option as soon as I come back home in January.

Best regards

CharliePoole:
I think you'll find that `TestExecutionContext.CurrentContext.CurrentTest` gives you what you need.

Be aware that either CurrentContext or CurrentTest could turn out to be null if you make the call when no execution context has been established.

Suremaker:
Hello,
I have checked the recommended approach and it worked perfectly for me, so I think this pull request can be abandoned.

Thank you very much for suggestion.
Happy new year :)

PR #1940, opened 2016-12-28
Issue 1933

### LINKED ISSUE #1933: Tests should pass if test case source provides 0 test cases

CharliePoole:
**Do Not Merge**
Fixes #1933

### DIFF
--- a/src/NUnitFramework/tests/Attributes/TheoryTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TheoryTests.cs
@@ -42,15 +42,15 @@ namespace NUnit.Framework.Attributes
         }
 
         [Test]
-        public void TheoryWithNoDatapointsIsNotRunnable()
+        public void TheoryWithNoDatapointsIsRunnable()
         {
-            TestAssert.IsNotRunnable(fixtureType, "TheoryWithArgumentsButNoDatapoints");
+            TestAssert.IsRunnable(fixtureType, "TheoryWithArgumentsButNoDatapoints");
         }
 
         [Test]
-        public void UnsupportedNullableTypeArgumentWithNoDatapointsAreNotRunnable()
+        public void UnsupportedNullableTypeArgumentWithNoDatapointsAreRunnable()
         {
-            TestAssert.IsNotRunnable(fixtureType, "TestWithUnsupportedNullableTypeArgumentWithNoDataPoints");
+            TestAssert.IsRunnable(fixtureType, "TestWithUnsupportedNullableTypeArgumentWithNoDataPoints");
         }
 
         [Test]
--- a/src/NUnitFramework/tests/TestUtilities/TestAssert.cs
+++ b/src/NUnitFramework/tests/TestUtilities/TestAssert.cs
@@ -81,7 +81,7 @@ namespace NUnit.TestUtilities
 
         public static void IsNotRunnable(Type type, string name)
         {
-            IsNotRunnable(TestBuilder.MakeTestCase(type, name));
+            IsNotRunnable(TestBuilder.MakeTestFromMethod(type, name));
         }
 
         public static void ChildNotRunnable(Type type, string name)
--- a/src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
@@ -127,6 +127,9 @@ namespace NUnit.Framework.Internal.Builders
                     needCombinatorial = false;
             }
 
+            // This check must be done before CombinatorialAttribute gets added to the builders collection
+            var hasBuildersSpecified = builders.Count > 0;
+
             // We could check to see if here are any data attributes specified
             // on the parameters but that's what CombinatorialAttribute does
             // and it simply won't return any cases if it finds nothing.
@@ -142,7 +145,7 @@ namespace NUnit.Framework.Internal.Builders
                     tests.Add(test);
             }
 
-            return tests.Count > 0 || builders.OfType<TestCaseSourceAttribute>().Any()
+            return hasBuildersSpecified || tests.Count > 0
                 ? BuildParameterizedMethodSuite(method, tests)
                 : BuildSingleTestMethod(method, parentSuite);
         }
--- a/src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal.Commands;
 
@@ -141,7 +142,7 @@ namespace NUnit.Framework.Internal.Builders
                     tests.Add(test);
             }
 
-            return tests.Count > 0
+            return tests.Count > 0 || builders.OfType<TestCaseSourceAttribute>().Any()
                 ? BuildParameterizedMethodSuite(method, tests)
                 : BuildSingleTestMethod(method, parentSuite);
         }
--- a/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
@@ -290,6 +290,21 @@ namespace NUnit.Framework.Attributes
             Assert.AreEqual(lhs, rhs);
         }
 
+
+        private static IEnumerable<TestCaseData> ZeroTestCasesSource() => Enumerable.Empty<TestCaseData>();
+
+        [TestCaseSource("ZeroTestCasesSource")]
+        public void TestWithZeroTestSourceCasesShouldNotBeRun()
+        {
+            Assert.Fail();
+        }
+
+        [TestCaseSource("ZeroTestCasesSource")]
+        public void TestWithZeroTestSourceCasesShouldPassWithoutRequiringArguments(int requiredParameter)
+        {
+        }
+
+
         static object[] testCases =
         {
             new TestCaseData(


CharliePoole:
@jnm2 I really needed to see your code back at that third commit, without anything else added, so I created this branch and PR to do it. Basically, this is your first three commits plus one more of mine, which you can look at.

As you can see, the tests needed to be changed because the purpose of the issue is that no tests will throw an error any longer because of not containing data. That includes Theories as well. 

However, Theories have an extra added thing. Right now the theories with no data are passing, but they are supposed to fail (not Error). So my next test additions will check that they fail.

Sorry to have jumped in on this, but I felt you were getting too many changes in too many places for what we actually want to do.

src/NUnitFramework/tests/Attributes/TheoryTests.cs
```diff
@@ -42,15 +42,15 @@ public void TheoryWithNoArgumentsIsTreatedAsTest()
...
-        public void TheoryWithNoDatapointsIsNotRunnable()
+        public void TheoryWithNoDatapointsIsRunnable()
         {
-            TestAssert.IsNotRunnable(fixtureType, "TheoryWithArgumentsButNoDatapoints");
+            TestAssert.IsRunnable(fixtureType, "TheoryWithArgumentsButNoDatapoints");
         }
 
         [Test]
-        public void UnsupportedNullableTypeArgumentWithNoDatapointsAreNotRunnable()
+        public void UnsupportedNullableTypeArgumentWithNoDatapointsAreRunnable()
```
jnm2: Nit: should be `UnsupportedNullableTypeArgumentWithNoDatapointsIsRunnable` because it's really talking about running the test, not running the datapoints which is the object of the preposition 'with' which is modifying 'runnable.'

jnm2:
> As you can see, the tests needed to be changed because the purpose of the issue is that no tests will throw an error any longer because of not containing data. That includes Theories as well.

That was not my understanding, but I'm just as happy with that goal as any. I was working within the limitations of not being allowed to change the meaning of `TheoryWithNoDatapointsIsNotRunnable` and `UnsupportedNullableTypeArgumentWithNoDatapointsAreNotRunnable` or things would surely have turned out differently.

So if I'm understanding you right, this PR still lacks the ability for theories to fail because they have no test cases. That is something I've already done with `ITestPostProcessor` in my PR. What's your opinion on that?

src/NUnitFramework/tests/TestUtilities/TestAssert.cs
```diff
@@ -81,7 +81,7 @@ public static void IsNotRunnable(Type type)
 
         public static void IsNotRunnable(Type type, string name)
         {
-            IsNotRunnable(TestBuilder.MakeTestCase(type, name));
+            IsNotRunnable(TestBuilder.MakeTestFromMethod(type, name));
```
jnm2: Nice to know about this.

CharliePoole:
Yes, I know that was not your understanding. I couldn't manage to explain it in words, but that's why I wanted you to reset to the third commit and pause so I could show you. But you are very energetic and the time zone difference also makes it hard to keep up with one another. I decided it was easier to write the code to show what I meant.

Of course, when fixing a bug, we have to change the behavior that's broken - including Theories. I just didn't want you to change __other__ behavior or to make wider changes than (I'm guessing) are really needed. We'll see. :smile: 

I didn't go on to fix the problem because I wanted to show you my steps. Now that you have seen we do want Theories to be runnable without data (like tests) I'll submit another change to show the true failure and then we can look at your fix to see if it still applies.

jnm2:
> I couldn't manage to explain it in words, but that's why I wanted you to reset to the third commit and pause so I could show you. But you are very energetic and the time zone difference also makes it hard to keep up with one another. I decided it was easier to write the code to show what I meant.

I'm sorry! I didn't realize I was doing something that was inconveniencing you. My motivation was to make my work life easier and I was hoping I'd get this moved all the faster.

I'm assuming that my PR is going to be closed, correct?

src/NUnitFramework/tests/TestUtilities/TestAssert.cs
CharliePoole: The test utilities are a bit adhoc. Could do with some reorganizing. Basically, they started out before we had parameterized test cases and were only ever changed to the extent needed.

jnm2:
In hindsight, I would have liked to have been able to respond to the two insights that this PR represents, about changing theories without cases to be runnable but fail, and the definition of `IsNotRunnable`. But ultimately I'm confident there's no one more qualified than yourself to do the work, so I'm not complaining. 👍

src/NUnitFramework/tests/Attributes/TheoryTests.cs
CharliePoole: For sure!

CharliePoole:
@jnm2 We'll hang on to your PR in case we need the code.

You've been doing fine, but we needed to get straight what behavior we wanted and I wasn't manage to convey it without getting into the code. I find reviewing sometimes harder than coding.

jnm2:
I know we missed this release, but I'm really looking forward to getting this out because it's a conspicuous blocker at work for which I have no workaround. Is there anything I can do to help with this?

jnm2:
Looks like the TheryResultCommand is not executed because the teardown work is not performed because the Theory test suite work item's Completed event is not subscribed to.

jnm2:
And that is because RunChildren is not run for the parameterized test suite, only for the test fixture and the fixtures parent suites.

jnm2:
And that is because the fixture's `CompositeWorkItem.CreateChildWorkItems` is calling `CreateWorkItem` with a `TestMethod` for the theory instead of a parameterized test suite.

jnm2:
Following the breadcrumb trail:

 * The `TheoryResultCommand` is not executed because the teardown work is not performed because the theory test suite work item's `Completed` event is not subscribed to.
 * And that is because `CompositeWorkItem.RunChildren` is not run for the parameterized test suite, only for the test fixture and the fixtures parent suites.
 * And that is because the theory parameterized test suite is being given a SimpleWorkItem instead of a CompositeWorkItem.
 * And that is because the fixture's `CompositeWorkItem.CreateChildWorkItems` is calling `CreateWorkItem` with a `TestMethod` for the theory instead of a parameterized test suite.
 * And that is because `DefaultTestCaseBuilder` has a **bad condition** for whether to create a test suite- it's erroneously allowing a test with the `Theory` attribute to remain a `TestMethod` instead of creating a suite.

@CharliePoole And so we see yet again that we need test builders to be able to either have no effect, or to return zero test cases while making the decision to convert the test to a suite.

jnm2:
Following the breadcrumb trail:

 * The `TheoryResultCommand` is not executed because the teardown work is not performed because the theory test suite work item's `Completed` event is not subscribed to.
 * And that is because `CompositeWorkItem.RunChildren` is not run for the parameterized test suite, only for the test fixture and the fixtures parent suites.
 * And that is because the theory parameterized test suite is being given a SimpleWorkItem instead of a CompositeWorkItem.
 * And that is because the fixture's `CompositeWorkItem.CreateChildWorkItems` is calling `CreateWorkItem` with a `TestMethod` for the theory instead of a parameterized test suite.
 * And that is because `DefaultTestCaseBuilder` has a **bad condition** for whether to create a test suite- it's erroneously allowing a test with the `Theory` attribute to remain a `TestMethod` instead of creating a suite.

@CharliePoole And so we see yet again that we need test builders to be able to either have no effect, or to return zero test cases while making the decision to convert the test to a suite.

This is the exact scenario my original PR provides for. I still think TestBuilderAction is the best API. It obviously provides needed decision making semantics which the builders need and which they currently don't have. It's clear and easy to extend later on if need be.

jnm2:
I have merged all the knowledge of our two PRs, plus my discoveries [above](https://github.com/nunit/nunit/pull/1940#issuecomment-270383243), into an example PR so that you can see what I'm talking about: #1951 

What I've done is:
 * Replace my [Fixed theory tests to expect tests with no datapoints to be parameterized](https://github.com/nunit/nunit/pull/1935/commits/5fb916b8f95c954811d6230c94c60b5024da337c) commit with your two commits (https://github.com/nunit/nunit/pull/1940/commits/3ca78a42a99396f33fc4305108338b1e4f0b8985 and https://github.com/nunit/nunit/pull/1940/commits/22cd3317d687c1c09b06884a911171107cbff1aa)
 * Retain the important [Added test for passing theory with test cases instead of datapoints](https://github.com/nunit/nunit/pull/1935/commits/3fc2191773dd82f528bc2d6c160c288d81299ba7) commit.
 * Retain the "ITestBuilders decide whether to create a suite" portion of https://github.com/nunit/nunit/pull/1935/commits/0e7fddd5a38ce25d82f38f50ffdf8b9df4ac173a
 * Add a new commit demonstrating the next step

----
And for the next step:

There is now one remaining design conflict which predates this whole issue, but also blocks this issue. These two things have been in conflict:

 * When a test suite has no child tests, it is not allowed to run one-time setup or **one-time teardown**
 * Theories with no child tests should fail, but the TheoryResultCommand which implements the failing logic can only be run as part of **one-time teardown**

When this conflict is resolved, everything will work. (Note that we will still need the `TestBuilderAction` as well for separate reasons, pointed out [above](https://github.com/nunit/nunit/pull/1940#issuecomment-270383243).)

jnm2:
Ideas to resolve the conflict:

 * Maybe provide a facility other than `DelegatingTestCommand` for zero-case theories to use to fail
 * Maybe stop conflating the execution of user teardown code with the concept of `TestCommand`

jnm2:
This non-`TheoryAttribute`-controlled condition strikes me as a code smell: https://github.com/nunit/nunit/blob/bf7e527b0475165992f0d01be1acdd6090f84b4d/src/NUnitFramework/framework/Internal/Execution/CommandBuilder.cs#L89

jnm2:
I'll be able to get at this some more around 4 pm EST.

jnm2:
* `TheoryResultCommand` [checks `TestResult.HasChildren`](https://github.com/nunit/nunit/blob/bf7e527b0475165992f0d01be1acdd6090f84b4d/src/NUnitFramework/framework/Internal/Commands/TheoryResultCommand.cs#L52) instead of `TestResult.Test.HasChildren`- may not be important, but want to double check this because I just ran across a case where `TestResult.Test.HasChildren` but not `TestResult.HasChildren`. I presume the result was in the process of being built.

jnm2:
@CharliePoole [New commit](https://github.com/nunit/nunit/pull/1951/commits/9a4798cc462370521decf9a707896f0cd63c62af), all tests passing. What do you think of this proposal? Seems the most minimalist route we can go; the cleanup in https://github.com/nunit/nunit/pull/1940#issuecomment-270411026 is more involved and can happen in an isolated PR.

CharliePoole:
It turns out upon analysis that the problem of Theories without data being reported as passed requires some fundamental changes, perhaps even to the framework interfaces. It's not something we want to deal with in the days immediately preceding a release. OTOH, we would like to get the basic issue reported here fixed for that release.

I'm rolling back to the state where we had resolved the initial problem and uncovered the fact that Theories without data were being passed. I think this PR can then be merged for the release once CI passes.

I'll next create an issue for the problem with Theories, which we can resolve more at leisure. It's an issue that has existed for at least a year without anyone stumbling across it so I think it can wait another cycle to be fixed.

CharliePoole:
Intermittent failure in travis-ci/push build is unrelated and is fixed in the merge.

Issue #1941, opened 2016-12-28
Use dictionary-based property for test run parameters


jnm2:
This issue tracks the orchestration of the work assigned to me by @CharliePoole in https://github.com/nunit/nunit/issues/1885#issuecomment-268872044.

 * ~~Superseding~~ Supplementing the string-based test parameters property with a dictionary-based one. Decouples parsing test parameters between the various runners, from each other and from the framework which test assemblies happen to be built against.
 * Providing backward compatibility for both old runners with new framework and old framework with new runners.

- [x] Framework: https://github.com/nunit/nunit/pull/1942, commit https://github.com/nunit/nunit/pull/1942/commits/cdf060459ccd7412d1a3b6270de63d53548e167a
- [x] NUnitLite: https://github.com/nunit/nunit/pull/1942, commit https://github.com/nunit/nunit/pull/1942/commits/00fb62de50b0f8648eefd5b46ebb4c60ef3816f8
- [x] nunit3-vs-adapter: https://github.com/nunit/nunit3-vs-adapter/pull/289
- [x] nunit-console: https://github.com/nunit/nunit-console/pull/155
- [x] dotnet-test-nunit: https://github.com/nunit/dotnet-test-nunit/pull/96
     - [x] nunit.portable.agent: https://github.com/nunit/nunit.portable.agent/pull/20
- [ ] <s>nunit-gui</s> Does not seem to have a concept of test parameters?
- [ ] Any others?

Due to the two-way backwards compatibility guarantee, each of these top-level tasks is completely  independent of the others.

CharliePoole:
This seems to be the same as #1885, which is already assigned to you. Is there a reason you prefer to use another issue or are you trying to create an Epic to cover all the work?

CharliePoole:
OK.. I see you are calling this the umbrella issue. Good idea, just not implemented as we do it. I'll make this a Zenhub Epic when I get home. Can't do that on my phone.

jnm2:
@CharliePoole however you manage it is fine. I saw this issue as a grouping of tasks, but itself a subtask under #1885.

jnm2:
Also did it for pragmatic reasons, because I guessed we'd have some discussion on the more specific issue (moving the parameter parsing out of the framework) which I didn't want arbitrarily spread across PRs, and also because the parent issue (fixing test parameters with semicolons) is already fairly long.

oznetmaster:
[...]

PR #1942, opened 2016-12-28
Test parameter parsing out of framework

jnm2:
Umbrella: https://github.com/nunit/nunit/issues/1941

### DIFF
--- a/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
+++ b/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
@@ -79,7 +79,7 @@ namespace NUnit.Framework.Api
 #else
             log.Debug("Loading {0} in AppDomain {1}", assembly.FullName, AppDomain.CurrentDomain.FriendlyName);
 #endif
-            
+
 #if PORTABLE
             string assemblyPath = AssemblyHelper.GetAssemblyName(assembly).FullName;
 #else
@@ -131,26 +131,41 @@ namespace NUnit.Framework.Api
                 if (options.ContainsKey(FrameworkPackageSettings.DefaultTestNamePattern))
                     TestNameGenerator.DefaultTestNamePattern = options[FrameworkPackageSettings.DefaultTestNamePattern] as string;
 
-                if (options.ContainsKey(FrameworkPackageSettings.TestParameters))
+                if (options.ContainsKey(FrameworkPackageSettings.TestParametersDictionary))
                 {
-                    string parameters = options[FrameworkPackageSettings.TestParameters] as string;
-                    if (!string.IsNullOrEmpty(parameters))
-                        foreach (string param in parameters.Split(new[] { ';' }))
-                        {
-                            int eq = param.IndexOf("=");
+                    var testParametersDictionary = options[FrameworkPackageSettings.TestParametersDictionary] as IDictionary<string, string>;
+                    if (testParametersDictionary != null)
+                    {
+                        foreach (var parameter in testParametersDictionary)
+                            TestContext.Parameters.Add(parameter.Key, parameter.Value);
+                    }
+                }
+                else
+                {
+                    // This cannot be changed without breaking backwards compatibility with old runners.
+                    // Deserializes the way old runners understand.
 
-                            if (eq > 0 && eq < param.Length - 1)
+                    if (options.ContainsKey(FrameworkPackageSettings.TestParameters))
+                    {
+                        string parameters = options[FrameworkPackageSettings.TestParameters] as string;
+                        if (!string.IsNullOrEmpty(parameters))
+                            foreach (string param in parameters.Split(new[] { ';' }))
                             {
-                                var name = param.Substring(0, eq);
-                                var val = param.Substring(eq + 1);
+                                int eq = param.IndexOf("=");
 
-                                TestContext.Parameters.Add(name, val);
+                                if (eq > 0 && eq < param.Length - 1)
+                                {
+                                    var name = param.Substring(0, eq);
+                                    var val = param.Substring(eq + 1);
+
+                                    TestContext.Parameters.Add(name, val);
+                                }
                             }
-                        }
+                    }
                 }
 
                 IList fixtureNames = null;
-                if (options.ContainsKey (FrameworkPackageSettings.LOAD))
+                if (options.ContainsKey(FrameworkPackageSettings.LOAD))
                     fixtureNames = options[FrameworkPackageSettings.LOAD] as IList;
                 var fixtures = GetFixtures(assembly, fixtureNames);
 
@@ -243,7 +258,7 @@ namespace NUnit.Framework.Api
         // the full-trust 'PermissionSetAttribute' on this class. Callers of this method have no influence on how the 
         // Process class is used, so we can safely satisfy the link demand with a 'SecuritySafeCriticalAttribute' rather
         // than a 'SecurityCriticalAttribute' and allow use by security transparent callers.
-        [SecuritySafeCritical]  
+        [SecuritySafeCritical]
 #endif
         private TestSuite BuildTestAssembly(Assembly assembly, string assemblyName, IList<Test> fixtures)
         {
--- a/src/NUnitFramework/framework/Api/FrameworkPackageSettings.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkPackageSettings.cs
@@ -106,9 +106,14 @@ namespace NUnit
         public const string DefaultTestNamePattern = "DefaultTestNamePattern";
 
         /// <summary>
-        /// Parameters to be passed on to the test
+        /// Parameters to be passed on to the tests, serialized to a single string which needs parsing. Obsoleted by <see cref="TestParametersDictionary"/>; kept for backward compatibility.
         /// </summary>
         public const string TestParameters = "TestParameters";
+
+        /// <summary>
+        /// Parameters to be passed on to the tests, already parsed into an IDictionary&lt;string, string>. Replaces <see cref="TestParameters"/>.
+        /// </summary>
+        public const string TestParametersDictionary = "TestParametersDictionary";
 
         #endregion
     }
--- a/src/NUnitFramework/nunitlite/TextRunner.cs
+++ b/src/NUnitFramework/nunitlite/TextRunner.cs
@@ -25,7 +25,9 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Reflection;
+using System.Text;
 using NUnit.Common;
 using NUnit;
 using NUnit.Framework.Api;
@@ -360,8 +362,8 @@ namespace NUnitLite
             if (options.DefaultTestNamePattern != null)
                 runSettings[FrameworkPackageSettings.DefaultTestNamePattern] = options.DefaultTestNamePattern;
 
-            if (options.TestParameters != null)
-                runSettings[FrameworkPackageSettings.TestParameters] = options.TestParameters;
+            if (options.TestParameters.Count != 0)
+                runSettings[FrameworkPackageSettings.TestParametersDictionary] = options.TestParameters;
 
             return runSettings;
         }
--- a/src/NUnitFramework/nunitlite.tests/CommandLineTests.cs
+++ b/src/NUnitFramework/nunitlite.tests/CommandLineTests.cs
@@ -550,7 +550,7 @@ namespace NUnitLite.Tests
         {
             var options = new NUnitLiteOptions("--params=X=5");
             Assert.That(options.errorMessages, Is.Empty);
-            Assert.That(options.TestParameters, Is.EqualTo("X=5"));
+            Assert.That(options.TestParameters, Is.EqualTo(new Dictionary<string, string> { {"X", "5" } }));
         }
 
         [Test]
@@ -558,7 +558,7 @@ namespace NUnitLite.Tests
         {
             var options = new NUnitLiteOptions("--params:X=5;Y=7");
             Assert.That(options.errorMessages, Is.Empty);
-            Assert.That(options.TestParameters, Is.EqualTo("X=5;Y=7"));
+            Assert.That(options.TestParameters, Is.EqualTo(new Dictionary<string, string> { { "X", "5" }, { "Y", "7" } }));
         }
 
         [Test]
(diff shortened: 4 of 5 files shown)


src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -124,20 +124,25 @@ public ITest Build(string assemblyName, IDictionary<string, object> options)
...
         private TestSuite Build(Assembly assembly, string assemblyPath, IDictionary<string, object> options)
         {
-            TestSuite testAssembly = null;
-
             try
             {
-                if (options.ContainsKey(FrameworkPackageSettings.DefaultTestNamePattern))
-                    TestNameGenerator.DefaultTestNamePattern = options[FrameworkPackageSettings.DefaultTestNamePattern] as string;
+                var testNamePattern = GetValueOrDefault(options, FrameworkPackageSettings.DefaultTestNamePattern) as string;
+                if (testNamePattern != null) TestNameGenerator.DefaultTestNamePattern = testNamePattern;
```
CharliePoole: Please don't arbitrarily change code that isn't relevant to the thing you are working on based on personal stylistic preferences.

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -124,20 +124,25 @@ public ITest Build(string assemblyName, IDictionary<string, object> options)
...
-                    string parameters = options[FrameworkPackageSettings.TestParameters] as string;
-                    if (!string.IsNullOrEmpty(parameters))
-                        foreach (string param in parameters.Split(new[] { ';' }))
+                    // Kept for backwards compatibility with old runners
+                    var testParametersString = GetValueOrDefault(options, FrameworkPackageSettings.TestParameters) as string;
+                    if (!string.IsNullOrEmpty(testParametersString))
+                        foreach (var param in testParametersString.Split(';'))
                         {
-                            int eq = param.IndexOf("=");
+                            var eq = param.IndexOf("=");
```
CharliePoole: We're generally conservative about var and use int in a case like this. See the coding standards. Are you using some 3rd party software that makes recommendations about things like this?

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -148,22 +153,24 @@ private TestSuite Build(Assembly assembly, string assemblyPath, IDictionary<stri
                             }
                         }
                 }
+                
+                var fixtures = GetFixtures(assembly, GetValueOrDefault(options, FrameworkPackageSettings.LOAD) as IList);
```
CharliePoole: Again... what does this have to do with the issue?

src/NUnitFramework/nunitlite/TextRunner.cs
```diff
@@ -361,12 +363,52 @@ private int ExploreTests()
                 runSettings[FrameworkPackageSettings.DefaultTestNamePattern] = options.DefaultTestNamePattern;
 
             if (options.TestParameters != null)
-                runSettings[FrameworkPackageSettings.TestParameters] = options.TestParameters;
+            {
+                var dictionary = new Dictionary<string, string>();
```
CharliePoole: As with the console, I'd prefer to see the dictionary creation centralized in the Options hierarchy so that the framework assembly simply receives a dictionary. Otherwise, there is no point in making it a framework run setting. The framework isn't intended to change the settings that are passed to it, just to use them.

As an aside, I suppose that working on so many versions of the same thing is giving both you and the reviewers extra work. 😢

CharliePoole:
An added "learning" here, if you don't mind. Let's say you do a PR that contains code that functionally does three different things. It will only be merged if two committers agree with what you have done for __all three__ things. So as a matter of self-interest, it's best to keep issues focused.

In fact, that was part of your logic (which convinced me!)  in separating what seemed like one issue at first into two. :smile:

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
jnm2: Yes, ReSharper. The original code mixed var and non-var styles so I figured I'd pick one. Easy enough to undo.

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
jnm2: :+1:

jnm2:
This is, I suppose, changing two things: the framework, and nunit. Should I split into two PRs then?

jnm2:
This is, I suppose, changing two things: the framework, and nunit. They are split into two commits. Should I split into two PRs then? I saw all these PRs as a single unit of work, only split because repos. But they are intrinsically separable so perhaps two PRs is better.

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
jnm2: Consistency, instead of having multiple ways to do the same thing. No problem to change back.

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
oznetmaster: It is easy to set ReSharper to only suggest/change to var when the assignment meets the criteria of NUnit (the RHS explicitely defines the type of the variable).

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
jnm2: Consistency, instead of doing the same thing multiple ways. No problem to change back.

Perhaps also I should have copied the original `ContainsKey`...indexer pattern. Writing that code is harder for me because it's not only less efficient, it's less DRY, but I can see how maybe you'd want to discuss that first.

src/NUnitFramework/nunitlite/TextRunner.cs
jnm2: I'll simply wait for the conclusion of [these comments](https://github.com/nunit/nunit-console/pull/155#issuecomment-269808071) and pick up here. =)

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
jnm2: @CharliePoole my bad on missing https://github.com/nunit/docs/wiki/Coding-Standards#use-of-the-var-keyword. Will be more attentive.

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -79,7 +79,7 @@ public ITest Build(Assembly assembly, IDictionary<string, object> options)
 #else
             log.Debug("Loading {0} in AppDomain {1}", assembly.FullName, AppDomain.CurrentDomain.FriendlyName);
 #endif
-            
+
```
jnm2: @CharliePoole I am not seeing these whitespace changes as they happen. Do you care about these? I can work on it.

src/NUnitFramework/nunitlite/TextRunner.cs
```diff
@@ -361,12 +363,52 @@ private int ExploreTests()
...
+            runSettings[FrameworkPackageSettings.TestParametersDictionary] = testParameters;
+
+            if (testParameters.Count != 0)
+            {
+                // Kept for backwards compatibility with old frameworks.
+                // Reserializes the way old frameworks understand, even if the parsing above is changed.
+                // This reserialization cannot be changed without breaking compatibility with old frameworks.
+
+                var oldFrameworkSerializedParameters = new StringBuilder();
+                foreach (var parameter in testParameters)
```
jnm2: @CharliePoole I'm willing to follow https://github.com/nunit/docs/wiki/Coding-Standards#use-of-the-var-keyword and use `foreach (KeyValuePair<string, string> parameter in testParameters)` here too. Would if be good for me to update this?

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
CharliePoole: No, it happens to everyone and if there are too many we just use look at the diff ignoring whitespace.

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
CharliePoole: Be sure to pay attention to the most important part:
  https://github.com/nunit/docs/wiki/Coding-Standards#general---please-read-this
especially paragraph 2. 😄

CharliePoole:
This is, I suppose, changing two things: the framework, and nunit. They are split into two commits. Should I split into two PRs then? I saw all these PRs as a single unit of work, only split because repos. But they are intrinsically separable so perhaps two PRs is better

No, the "thing" that's being changed is "How nunitlite passes the run parameters to the framework" It requires a change in two places, but it's one issue in one repository.

OTOH, changing the same thing for the console crosses an arbitrary administrative boundary we have created... between the team of people who work on the framework and the team that works on the engine and console. So it needs to be two issues. If too many things cross that boundary and cause us problems, it's in the realm of the possible that we could re-examine the decision. But for now, it sticks.

For a newcomer, this is probably hard to see because we created the boundary but for the moment the same people are working on both sides of it. We hope that won't last and that more new people will come to each project, allowing them to move ahead independently. IOW it's a bet we are making.

src/NUnitFramework/nunitlite/TextRunner.cs
CharliePoole: The standard is deliberately vague. IOW, we could have written "only use var if the rhs contains the name of the type" but we didn't. The idea (my idea) is that developers are supposed to use judgement - a notion that is too often forgotten. Whether it's sufficiently clear is a team judgement, since all code is clear to the guy who wrote it (at least for a while). My judgement is that it makes sense in context. Somebody else may say different.

src/NUnitFramework/nunitlite/TextRunner.cs
CharliePoole: Fair enough.

jnm2:
> Since it's a common option, it should go into the base class, CommandLineOptions in both cases. If you do it in the framework project first, we may be able to get it into the release this week. The console release will be next week.

Framework done. @CharliePoole The CommandLineOptions route does feel cleaner.

jnm2:
Not quite done, forgot to check in the separate commit for the parameters test fixes.

jnm2:
Decided it was simpler to put the updated tests in the same NUnitLite commit.

rprouse:
Let me take another look at this in the morning and we can figure out what to do for the release. I plan on branching for the release tomorrow, but I can always rebase last minute to pull this in if it is ready.

jnm2:
@CharliePoole not a problem. I'd like to clean that up before merge.

jnm2:
@CharliePoole @rprouse Should be ready.

CharliePoole:
This looks good to me but it's sufficiently important, affecting console, gui, adapter, etc., that I'd like to get another set of eyes on it as well. @ChrisMaddock can you take a look? If we can get this into the release, then the corresponding runner changes will also be able to go in.

src/NUnitFramework/framework/Api/FrameworkPackageSettings.cs
```diff
@@ -106,10 +106,15 @@ public static class FrameworkPackageSettings
         public const string DefaultTestNamePattern = "DefaultTestNamePattern";
 
         /// <summary>
-        /// Parameters to be passed on to the test
+        /// Parameters to be passed on to the tests, serialized to a single string which needs parsing. Obsoleted by <see cref="TestParametersDictionary"/>; kept for backward compatibility.
         /// </summary>
         public const string TestParameters = "TestParameters";
 
+        /// <summary>
+        /// Parameters to be passed on to the tests, already parsed into an IDictionary&lt;string, string>. Replaces <see cref="TestParameters"/>.
```
ChrisMaddock: Just a little typo here. Not important enough to hold up the release, but perhaps can be fixed if you get back to it before it's merged, @jnm2 . 😄

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -131,26 +131,41 @@ private TestSuite Build(Assembly assembly, string assemblyPath, IDictionary<stri
...
-                            int eq = param.IndexOf("=");
+                    var testParametersDictionary = options[FrameworkPackageSettings.TestParametersDictionary] as IDictionary<string, string>;
+                    if (testParametersDictionary != null)
+                    {
+                        foreach (var parameter in testParametersDictionary)
+                            TestContext.Parameters.Add(parameter.Key, parameter.Value);
+                    }
+                }
+                else
+                {
```
ChrisMaddock: Just a suggestion - would this be clearer as an `if...else if` now the noise has been removed? I think that's me going well into the grounds of stylistic preference by now though - so feel free to ignore that. 😄

CharliePoole:
I see that the release branch has already been created so I'm merging this to master. It will be up to @rprouse to decide whether to update th branch with all our merged fixes or to cherry pick them.

src/NUnitFramework/framework/Api/FrameworkPackageSettings.cs
jnm2: Ugh! Where? I'm not seeing it!

CharliePoole: It's merged in any case.

jnm2: I know but it'll drive me crazy. How am I not seeing the typo?

ChrisMaddock: Sorry - it's not! I presumed `&lt;string, string>` should be either all escaped or not escaped, but I just tried my 'correction' - and it doesn't compile, as I presume you already knew. Sorry, false alarm!

jnm2: Oh no problem! I would have thought `&gt;` would compile. Weird.

ChrisMaddock: Probably - I only tried `<string, string>` which didn't. :smile:

jnm2: I would have done the whole `<see cref="IDictionary{X,Y}"/>` but there is no facility for specifying closed generics, so users would have seen `IDictionary<TKey, TValue>` no matter what I wrote for `X` and `Y`. :(

PR #1951, opened 2017-01-04
[Example] Zero-case test passing

jnm2:
Serves as an example for discussion here:
https://github.com/nunit/nunit/pull/1940#issuecomment-270383243

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/OneTimeSetUpCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/OneTimeSetUpCommand.cs
@@ -62,6 +62,8 @@ namespace NUnit.Framework.Internal.Commands
         /// <returns>A TestResult</returns>
         public override TestResult Execute(TestExecutionContext context)
         {
+            TestResult suiteResult = context.CurrentResult;
+
             if (_typeInfo != null)
             {
                 // Use pre-constructed fixture if available, otherwise construct it
@@ -75,15 +77,18 @@ namespace NUnit.Framework.Internal.Commands
                     Test.Fixture = _suite.Fixture;
                 }
 
-                for (int i = _setUpTearDown.Count; i > 0; )
-                    _setUpTearDown[--i].RunSetUp(context);
+                if (!(suiteResult.Test is TestFixture && !suiteResult.Test.HasChildren))
+                {
+                    for (int i = _setUpTearDown.Count; i > 0;)
+                        _setUpTearDown[--i].RunSetUp(context);
+                }
             }
-
+            
 
             for (int i = 0; i < _actions.Count; i++)
                 _actions[i].BeforeTest(Test);
 
-            return context.CurrentResult;
+            return suiteResult;
         }
     }
 }
--- a/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
@@ -106,10 +106,14 @@ namespace NUnit.Framework.Internal.Execution
 
                             CreateChildWorkItems();
 
+                            // This is important to do even when there are no child tests.
+                            // For example, the teardown command is how theories get an
+                            // opportunity to fail when there are no test cases. To execute
+                            // teardown commands, we must execute setup commands.
+                            PerformOneTimeSetUp();
+
                             if (_children.Count > 0)
                             {
-                                PerformOneTimeSetUp();
-
                                 if (!CheckForCancellation())
                                 {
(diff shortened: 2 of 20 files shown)


jnm2:
Question [moved to gitter](https://gitter.im/nunit/nunit?at=586e9ce07a3f79ef5dde42af).

src/NUnitFramework/framework/Internal/Commands/OneTimeSetUpCommand.cs
```diff
@@ -75,15 +77,18 @@ public override TestResult Execute(TestExecutionContext context)
                     Test.Fixture = _suite.Fixture;
                 }
 
-                for (int i = _setUpTearDown.Count; i > 0; )
-                    _setUpTearDown[--i].RunSetUp(context);
+                if (!(suiteResult.Test is TestFixture && !suiteResult.Test.HasChildren))
```
jnm2: @CharliePoole Maybe this check could be encapsulated into an `ITest` extension method or some other place?

jnm2: Actually `if (suiteResult.Test.HasChildren)` passes all tests too. The rule (that user setup and teardown are not run for 0 child tests) only makes sense in the context of fixtures, right?

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -128,13 +128,16 @@ protected override void PerformWork()
...
-                                // were skipped, failed or set to inconclusive in one time setup
-                                // unless we are aborting.
-                                if (Context.ExecutionStatus != TestExecutionStatus.AbortRequested)
-                                    PerformOneTimeTearDown();
                             }
+
+                            // Directly execute the OneTimeFixtureTearDown for tests that
+                            // were skipped, failed or set to inconclusive in one time setup
+                            // unless we are aborting.
+                            // This is important to do even when there are no child tests,
```
jnm2: Alright! Sounds like we may have a solution then?
How would you word it? A "for example, theories" part could maybe still be useful.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -106,10 +106,14 @@ protected override void PerformWork()
 
                             CreateChildWorkItems();
 
+                            // This is important to do even when there are no child tests.
+                            // For example, the teardown command is how theories get an
+                            // opportunity to fail when there are no test cases. To execute
+                            // teardown commands, we must execute setup commands.
+                            PerformOneTimeSetUp();
```
jnm2: @CharliePoole how is this?

src/NUnitFramework/framework/Internal/Commands/OneTimeSetUpCommand.cs
CharliePoole: That does the job, but it does it by having one component (OneTimeSetUpCommand) be aware of the necessary behavior for different types of Suites. See my comments about how this all came about in #1962. I think we want a solution that makes individual suite types responsible for how their result is calculated based on the results of contained tests.

CharliePoole:
I've gone ahead and finished a limited fix in PR #1940 and created a new issue #1962 to deal with the additional structural problem that we discovered by working on this. Some of your code is in the fix and your ideas about alternative ways of doing this are the reason I'm delaying it.

As you work with us, you will notice that the week or two right before a release is not a great time to try to solve fundamental problems. :wink:

src/NUnitFramework/framework/Internal/Commands/OneTimeSetUpCommand.cs
jnm2: Excellent. I assume #1962 covers the concerns I had in https://github.com/nunit/nunit/pull/1940#issuecomment-270411026?

jnm2:
> As you work with us, you will notice that the week or two right before a release is not a great time to try to solve fundamental problems

Haha! I did not expect this to get pushed through for this release but I am very happy about the #1933  getting fixed. Thank you!

> your ideas about alternative ways of doing this are the reason I'm delaying it.

Hopefully we can discuss that over at #1962 after the release? I'm not married to any particular solution, but I think there is a good reason to make the test builders responsible to deciding whether a test should become a suite, and not couple that decision to the number of cases collected from all the builders. I'll have more to say if that's okay with you.

CharliePoole:
Yes, let's discuss. I have to admit I never understood your comment abut making test builders responsible... Currently they are responsible. Perhaps you can clarify on #1962.

Issue #1896, opened 2016-11-09
Test has passed however Reason with an empty message is printed in the xml


psandhu79:
```
       <test-case id="0-2262" name="TestPickedPoint" fullname="A.B.C.TestPickedPoint" methodname="TestPickedPoint" classname="A.B.C" runstate="Runnable" seed="978314103" result="Passed" start-time="2016-11-08 15:28:16Z" end-time="2016-11-08 15:28:16Z" duration="0.010144" asserts="1">
                    <properties>
                      <property name="Category" value="SmokeTest" />
                    </properties>
                    <reason>
                      <message><![CDATA[]]></message>
                    </reason>
                  </test-case>
```

This when imported into TeamCity reports a failure

rprouse:
The fact that TeamCity reports a passed test as a failure is a problem that should be fixed there. @NikolayPianikov any comments on that? What about `Assert.Pass("Message")`?

@psandhu79 can you give us a code example of a test that displays an empty message?

psandhu79:
@rprouse Thank you looks like a bug in TeamCity found this 
https://youtrack.jetbrains.com/issue/TW-15717

However the tests pass in Nunit 2 for teamcity and fail in Nunit 3 i am guessing its a regression in the Nunit importer for the xml

CharliePoole:
Even if it is not the cause of the TC errror, NUnit should not be creating an empty reason element under any circumstance. We need to confirm whether this happens, and when, and fix it.

Based on the TC issue, it appears to happen when AssertPass is used. We should check out Assert.Pass both with and without a message.

psandhu79:
TC reports a failed test either with Assert.Pass() or Asset.Pass("No exception should occur until this point with current setup")

With a message the reason element contains the message

CharliePoole:
There should only be a reason element if there is a reason. Of course, TC should deal with it anyway, but we should fix our output as well.

NikolayPianikov:
@psandhu79 it looks like a bug in the TeamCity xml parser for NUnit. I will check it

psandhu79:
@NikolayPianikov could you please update this bug 
https://youtrack.jetbrains.com/issue/TW-15717

If possible attached a new xml plugin and i can test

NikolayPianikov:
@psandhu79 I can't reproduce this issue. Could you attach nunit xml report file?

NikolayPianikov:
@psandhu79 Looks like this issue was fixed in TeamCity 10.0.3
[...]

PR #1955, opened 2017-01-04
Suppress reason in xml if null or whitespace

cristobalito:
This should address issue #1896.

First commit to the project (and a novice to git(hub) to boot), so please feel free to point out any issues - I'll do my best to address promptly.

Issue only mentions a problem with the Success case (e.g. using Assert.Pass), but it appears that Inconclusive and Ignore can also produce reason elements in the xml and therefore would also have this issue. It's possible other states could be a problem, but I was basing the work of existing tests that check for reason.

There is a fair bit of duplication across the (existing) tests - I've not really tried to address that, aside from some low hanging fruit.

Changes should be relatively self explanatory, but please let me know if you'd like any more details.

Chris

### DIFF
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -386,8 +386,7 @@ namespace NUnit.Framework.Internal
                 case TestStatus.Passed:
                 case TestStatus.Inconclusive:
                 case TestStatus.Warning:
-                    if(Message != null)
-                    //if (Message != null && Message.Trim().Length > 0)
+                    if (Message != null && Message.Trim().Length > 0)
                         AddReasonElement(thisNode);
                     break;
             }
@@ -397,7 +396,6 @@ namespace NUnit.Framework.Internal
 
             if (AssertionResults.Count > 0)
                 AddAssertionsElement(thisNode);
-
 
             if (recursive && HasChildren)
                 foreach (TestResult child in Children)
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -386,7 +386,8 @@ namespace NUnit.Framework.Internal
                 case TestStatus.Passed:
                 case TestStatus.Inconclusive:
                 case TestStatus.Warning:
-                    if (Message != null)
+                    //if(Message != null)
+                    if (Message != null && Message.Trim().Length > 0)
                         AddReasonElement(thisNode);
                     break;
             }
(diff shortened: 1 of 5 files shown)


CharliePoole:
@cristobalito We're in the midst of putting out the next nunit release, so this will probably need to wait for review until afterwards. We'll be back to you!

cristobalito:
@CharliePoole No worries - good luck with the release.

CharliePoole:
@ChrisMaddoc That makes sense to me.

ChrisMaddock:
@cristobalito - would you mind expanding this PR to cover the `Assert.Fail()` case also? If not - no worries, we can merge this and make that a separate issue. Thanks! 😄

cristobalito:
Sure - happy to fix that as part of this PR.

rprouse:
Please hold off on merging this one until after the release and whomever merges, please do a squash from the merge dropdown. I don't mind a few commits, but over 20 is a bit much 😄

ChrisMaddock:
I think we should merge this in before it gets too stale and needs rebasing. @nunit/core-team - someone mind doing a second review?

We can then make the `Assert.Fail()` case a separate issue (it wasn't originally part of #1896 anyway). @cristobalito - if you're still interested, we can assign the new issue to you, and it can be done as a separate pull request?

cristobalito:
Hi Chris, still happy to do the work. Apologies I've been sidetracked a
little in the last few weeks.

Congrats on getting the recent release out btw.

Chris

ChrisMaddock:
Great, @cristobalito! No rush at all, I just thought it would be best if we get this in before we have merge problems. Although actually, maybe it's a good job this was delayed, I'm not entirely sure if our thinking was right here...

The [xml spec](https://github.com/nunit/docs/wiki/Test-Result-XML-Format#failure) states that `<reason>` is optional, and can be omitted like has just been done in this PR.  It seems to suggest however that `<message>` is a required child-element for `<failure>`.

The current code omits failure-messages when the message is null, but not when it is an empty string. I think we should change one way or the other. If we decide the element is required, then if message=null, an empty element should be put in. On the other hand we could make failure messages match other cases and be optional - I don't know how much is depending on its existence, and how strict this spec is.

@CharliePoole - any thoughts?

CharliePoole:
@ChrisMaddock I think it's slightly easier for those who want to read and process the XML if there are no empty elements. 

The document is not a specification to be followed. It's something I wrote after the fact describing what NUnit3 implemented. So, it's not in the least strict. 😄

ChrisMaddock:
Great - I updated the spec and created https://github.com/nunit/nunit/issues/2005. 😄

Issue #1954, opened 2017-01-04
Change Error Message for Assert.Equals


rprouse:
Based on the number of votes on this StackOverflow question, http://stackoverflow.com/questions/11584429/nunits-assert-equals-throws-exception-assert-equals-should-not-be-used-for-ass, many users find the message in our `Assert.Equals` override to be confusing. It currently says "Assert.Equals should not be used for Assertions", we should change it to be "Assert.Equals should not be used for Assertions, use Assert.AreEqual(...) instead".

JustinRChou:
Should the summary and ReferenceEquals also be updated?

CharliePoole:
@JustinRChou That makes sense
@rprouse Since we are unable to use EditorBrowsable in this context, I suggest that the intellisense should start with something like "DO NOT USE"!

CharliePoole:
Whoops! 😄

JustinRChou:
Also looks like following classes needs updating too.
```
Assume
CollectionAssert
DirectoryAssert
FileAssert
StringAssert
Warn
```
And subsequent test cases.

CharliePoole:
@rprouse Can of worms? :smile: I think most of this can wait till after the release though.

JustinRChou:
When is the next release? 
And I have updated as such before I start applying it to the rest.
```
        /// <summary>
        /// DO NOT USE! Use Assert.AreEqual(...) instead.
        /// The Equals method throws an InvalidOperationException. This is done 
        /// to make sure there is no mistake by calling this function.
        /// </summary>
        /// <param name="a"></param>
        /// <param name="b"></param>
        [EditorBrowsable(EditorBrowsableState.Never)]
        public static new bool Equals(object a, object b)
        {
            throw new InvalidOperationException("Assert.Equals should not be used for Assertions, use Assert.AreEqual(...) instead.");
        }
```

And should the EditorBrowsable attribute be applied to ReferenceEquals? I am not too familiar on how the attribute works.

CharliePoole:
@JustinRChou In a few days.
I like what you have done as a comment. I would usually drop the param elements if not using them. Are you planning to do a PR?

After much experimentation, I believe we have found that EditorBrowsable is no help at all for these messages. Feel free to experiment further. 😄

JustinRChou:
Sure.

Might not have to update the test cases since those use StartsWith.

JustinRChou:
@nunit/core-team 
Created the PR.
[...]

PR #1956, opened 2017-01-05
#1954 assert equals error message

JustinRChou:
Updated Summary and Exception Messages of the classes with the overridden Equals and ReferenceEquals with "DO NOT USE!" and suggested actual methods to use if they exist.

I assumed that AreEquals should also be used for ReferenceEquals.

Also I didn't really run any tests to check if the suggested methods are correct, I just looked at the code.

I couldn't find the appropriate methods for Warn and Assume.

Lastly, I think that the ReferenceEquals should probably be changed to something closer in test to Equals or vice versa?


JustinRChou:
Rebased everything to fix the commits.

Closing this and reopening a new PR.

PR #1957, opened 2017-01-05
#1954 assert equals error message fixed

### LINKED ISSUE #1954: Change Error Message for Assert.Equals

JustinRChou:
Updated Summary and Exception Messages of the classes with the overridden Equals and ReferenceEquals with "DO NOT USE!" and suggested actual methods to use if they exist.

I assumed that AreEquals should also be used for ReferenceEquals.

Also I didn't really run any tests to check if the suggested methods are correct, I just looked at the code.

I couldn't find the appropriate methods for Warn and Assume.

Lastly, I think that the ReferenceEquals should probably be changed to something closer in test to Equals or vice versa?

Fixes #1954

### DIFF
--- a/src/NUnitFramework/framework/Assert.cs
+++ b/src/NUnitFramework/framework/Assert.cs
@@ -66,6 +66,7 @@ namespace NUnit.Framework
         #region Equals and ReferenceEquals
 
         /// <summary>
+        /// DO NOT USE! Use Assert.AreEqual(...) instead.
         /// The Equals method throws an InvalidOperationException. This is done 
         /// to make sure there is no mistake by calling this function.
         /// </summary>
@@ -74,19 +75,21 @@ namespace NUnit.Framework
         [EditorBrowsable(EditorBrowsableState.Never)]
         public static new bool Equals(object a, object b)
         {
-            throw new InvalidOperationException("Assert.Equals should not be used for Assertions");
+            throw new InvalidOperationException("Assert.Equals should not be used for Assertions, use Assert.AreEqual(...) instead.");
         }
 
         /// <summary>
+        /// DO NOT USE! Use Assert.AreEqual(...) instead.
         /// override the default ReferenceEquals to throw an InvalidOperationException. This 
         /// implementation makes sure there is no mistake in calling this function 
         /// as part of Assert. 
         /// </summary>
         /// <param name="a"></param>
         /// <param name="b"></param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public static new void ReferenceEquals(object a, object b)
         {
-            throw new InvalidOperationException("Assert.ReferenceEquals should not be used for Assertions");
+            throw new InvalidOperationException("Assert.ReferenceEquals should not be used for Assertions, use Assert.AreEqual(...) instead.");
         }
 
         #endregion
--- a/src/NUnitFramework/framework/Assert.cs
+++ b/src/NUnitFramework/framework/Assert.cs
@@ -79,17 +79,16 @@ namespace NUnit.Framework
         }
 
         /// <summary>
-        /// DO NOT USE! Use Assert.AreEqual(...) instead.
-        /// override the default ReferenceEquals to throw an InvalidOperationException. This 
-        /// implementation makes sure there is no mistake in calling this function 
-        /// as part of Assert. 
+        /// DO NOT USE!
+        /// The ReferenceEquals method throws an InvalidOperationException. This is done 
+        /// to make sure there is no mistake by calling this function.
         /// </summary>
         /// <param name="a"></param>
         /// <param name="b"></param>
         [EditorBrowsable(EditorBrowsableState.Never)]
         public static new void ReferenceEquals(object a, object b)
         {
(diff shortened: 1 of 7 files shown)


src/NUnitFramework/framework/Assert.cs
```diff
@@ -74,19 +75,21 @@ public partial class Assert
...
         /// implementation makes sure there is no mistake in calling this function 
         /// as part of Assert. 
         /// </summary>
         /// <param name="a"></param>
         /// <param name="b"></param>
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public static new void ReferenceEquals(object a, object b)
         {
-            throw new InvalidOperationException("Assert.ReferenceEquals should not be used for Assertions");
+            throw new InvalidOperationException("Assert.ReferenceEquals should not be used for Assertions, use Assert.AreEqual(...) instead.");
```
CharliePoole: Should use Assert.AreSame instead - or just don't offer an alternative since this is probably not used much.

CharliePoole:
@JustinRChou I didn't understand this:

"Lastly, I think that the ReferenceEquals should probably be changed to something closer in test to Equals or vice versa?"

src/NUnitFramework/framework/Assert.cs
rprouse: Recommending `AreEqual` isn't right. I'd be okay with `AreSame` or nothing.

BTW, it probably doesn't make a different, but shouldn't this method return a bool to be consistent?

rprouse:
One last thing @JustinRChou please add a Fixes #1954 next time you submit a pull request. That will auto-close the related issue and cross-links them.

src/NUnitFramework/framework/Assert.cs
JustinRChou: So for the ReferenceEquals, suggest AreSame if the method exists?

From the looks of it only Assert has AreSame()?

CharliePoole: That's right. For the others, there's no recommendation we can make that makes any sense at all.

JustinRChou:
@CharliePoole That comment is with regards to how the summary is written for the methods. I just found it odd that there were different in the style of writing.

@rprouse I don't quite get how that whole Fixes thing works.

CharliePoole:
@rprouse "I think that since the methods throw an exception, it can't hurt to flag them as unusable"

Right... but we are the ones who decided to throw an exception to warn the users. It probably made sense for Assert, but I wonder if it makes any sense at all for the other classes that copied it.

Additionally, these classes were created before there were static classes. Would the problem go away if we made Assert, etc. static? A future discussion I guess.

CharliePoole:
@JustinRChou Perhaps due to many different people over a period of 16 years writing them. 😄 

As a bit of history, many of us didn't use IDEs at all, just an editor, so we never saw the intellisense.

JustinRChou:
@jnm2 Neat, thanks.

@CharliePoole I am going to also update the summary for the ReferenceEquals to match Equals.

CharliePoole:
@JustinRChou Be sure to post a comment here when it's ready for re-review. Otherwise, it can get lost.

jnm2:
What about `[Obsolete(message, true)]` or at least `[Obsolete(message)]`?

JustinRChou:
Using 
`[Obsolete("Do not use!", true)]`
looks much better.

But if I update it using the attribute, should I remove the exceptions since it becomes a compile error.

jnm2:
I'd leave the exceptions but copy the exception message to the `Obsolete` message. You never know, sometimes methods are executed via reflection.

jnm2:
In fact I'd keep the signature the same so that the *only* error the user sees is the obsoletion message. It's more 1:1 with your intent than mangling the signature. Then, you'll have to throw because it's no longer `void`.

jnm2:
Wait a minute- that was the original signature. Never mind.

JustinRChou:
Also would have to decide if forcing the compile error is the best option as that would also require that the assertion tests for these would have to be removed.

Unless the warning variation is used and the methods would be peppered with
```
#pragma warning disable 618
#pragma warning restore 618
```

jnm2:
That's something I was hoping @CharliePoole and @rprouse would have an opinion on.

CharliePoole:
I don't think we want to say that a System method is obsolete. :-)

jnm2:
Technically it's not the same method as object.ReferenceEquals, which you should never reference via `Assert`, so does that matter either way?

jnm2:
I'm always a proponent of failing fast and obviously. If it's going to be misleading at runtime, better to make it throw. If it's always going to throw at runtime, better to make it fail to compile.

CharliePoole:
Yes, let's get it out. We've spent far too much time on something that's not so important. I don't think I've ever seen anybody actually be confused by ReferenceEquals in the entire time I've worked on NUnit although a few people have fallen into the trap of Assert.Equals.

PR #1959, opened 2017-01-05
Add AttributeUsage to two attributes

### LINKED ISSUE #1880: AttributeUsage for various Attributes
I have noticed that many the attributes do not have the AttributeUsageAttribute applies to them.

For instance, the RandomAttribute, which only applies to a parameter, is not constrained by an AttributeUsageAttribute with a AttributeTargets.Parameter flag set.

It seems like it should be.

I have not examined all of the attributes, but there may be others that are not constrained that should be.

CharliePoole:
I forgot AttributeUsage on the new NonTestAssemblyAttribute. I figured I would deal with the existing issue #1880 at the same time so this fixes #1880 as well


CharliePoole:
@rprouse I don't see any urgency to getting this in the release if you don't want to bother with it right away.

PR #1967, opened 2017-01-09
Remove dependency on Microsoft.NETCore.Portable.Compatibility

### LINKED ISSUE #1966: Xamarin Runner cannot reference NUnit NuGet Package
As initially reported by @ChrisMaddock , since the .NET Standard target was added to the NuGet package, the Xamarin projects cannot reference the package. I have confirmed this and tracked it down to the reference to `Microsoft.NETCore.Portable.Compatibility` in the package. The package lists this as a dependency because the framework references it, but it only references it because of some unused using statements.

rprouse:
Fixes #1966 

I have tested this change by building locally and upgrading the Xamarin runner to use the fixed package.

### DIFF
--- a/nuget/framework/nunit.nuspec
+++ b/nuget/framework/nunit.nuspec
@@ -30,7 +30,6 @@ Supported platforms:
     <dependencies>
       <group targetFramework="netstandard1.6">
         <dependency id="NETStandard.Library" version="1.6.0" />
-        <dependency id="Microsoft.NETCore.Portable.Compatibility" version="1.0.1" />
         <dependency id="System.Runtime.Loader" version="4.0.0" />
       </group>
     </dependencies>
@@ -50,7 +49,11 @@ Supported platforms:
     <file src="bin/net-4.5/nunit.framework.xml" target="lib/net45" />
     <file src="bin/netstandard16/nunit.framework.dll" target="lib/netstandard1.6" />
     <file src="bin/netstandard16/nunit.framework.xml" target="lib/netstandard1.6" />
-    <file src="bin/portable/nunit.framework.dll" target="lib/portable-net45+win8+wp8+wpa81+Xamarin.Mac+MonoAndroid10+MonoTouch10+Xamarin.iOS10" />
-    <file src="bin/portable/nunit.framework.xml" target="lib/portable-net45+win8+wp8+wpa81+Xamarin.Mac+MonoAndroid10+MonoTouch10+Xamarin.iOS10" />
+    <file src="bin/portable/nunit.framework.dll" target="lib/portable-net45+win8+wp8+wpa81" />
+    <file src="bin/portable/nunit.framework.xml" target="lib/portable-net45+win8+wp8+wpa81" />
+    <file src="bin/portable/nunit.framework.dll" target="lib/MonoAndroid" />
+    <file src="bin/portable/nunit.framework.xml" target="lib/MonoAndroid" />
+    <file src="bin/portable/nunit.framework.dll" target="lib/Xamarin.iOS10" />
+    <file src="bin/portable/nunit.framework.xml" target="lib/Xamarin.iOS10" />
   </files>
 </package>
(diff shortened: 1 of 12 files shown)


nuget/framework/nunit.nuspec
```diff
@@ -50,7 +49,11 @@ Supported platforms:
...
     <file src="bin/netstandard16/nunit.framework.dll" target="lib/netstandard1.6" />
     <file src="bin/netstandard16/nunit.framework.xml" target="lib/netstandard1.6" />
-    <file src="bin/portable/nunit.framework.dll" target="lib/portable-net45+win8+wp8+wpa81+Xamarin.Mac+MonoAndroid10+MonoTouch10+Xamarin.iOS10" />
-    <file src="bin/portable/nunit.framework.xml" target="lib/portable-net45+win8+wp8+wpa81+Xamarin.Mac+MonoAndroid10+MonoTouch10+Xamarin.iOS10" />
+    <file src="bin/portable/nunit.framework.dll" target="lib/portable-net45+win8+wp8+wpa81" />
+    <file src="bin/portable/nunit.framework.xml" target="lib/portable-net45+win8+wp8+wpa81" />
+    <file src="bin/portable/nunit.framework.dll" target="lib/MonoAndroid" />
+    <file src="bin/portable/nunit.framework.xml" target="lib/MonoAndroid" />
+    <file src="bin/portable/nunit.framework.dll" target="lib/Xamarin.iOS10" />
+    <file src="bin/portable/nunit.framework.xml" target="lib/Xamarin.iOS10" />
```
rprouse: I couldn't find any common NuGet packages still including the Android and iOS as targets within the PCL, they are all broken out this way, so I did the same.

nuget/framework/nunit.nuspec
```diff
@@ -30,7 +30,6 @@ Supported platforms:
     <dependencies>
       <group targetFramework="netstandard1.6">
         <dependency id="NETStandard.Library" version="1.6.0" />
-        <dependency id="Microsoft.NETCore.Portable.Compatibility" version="1.0.1" />
```
rprouse: I confirmed that the Android and iOS are loading their versions of the framework, so I don't know why they are pulling in the .NET Standard dependencies, but they are. I even tested by removing the .NET Standard target and leaving these dependencies and they still get pulled in. It is this dependency that is causing the problems and it was trivial to remove it.

src/NUnitFramework/framework/Internal/InvalidDataSourceException.cs
```diff
@@ -24,7 +24,9 @@
 namespace NUnit.Framework.Internal
 {
     using System;
+#if !NETSTANDARD1_6
     using System.Runtime.Serialization;
```
rprouse: Turns out this is all we are using the `Microsoft.NETCore.Portable.Compatibility` dependency for.

CharliePoole:
Makes sense to me and it's building without the using statements.

nuget/framework/nunit.nuspec
ChrisMaddock: This dependency is also in the nunitlite nuspec - should that be cleaned up at the same time? The NUnitLite portable targets a slightly different profile to the nunit package, perhaps they could be matched up too.

nuget/framework/nunit.nuspec
ChrisMaddock: I presume we're assuming here that nobody actually uses the Xamarin.Mac/MonoTouch targets? 😄 I wonder how far away we are from netstandard 1.4, which (I think?) would just be consumable by all Xamarin targets...

ChrisMaddock:
Thanks for looking into this, Rob! 😄

nuget/framework/nunit.nuspec
rprouse: @ChrisMaddock I was targeting just what was required for Xamarin, but you are right, so I removed the reference from the remaining projects and from the NUnitLite nuspec. The reference is still needed in two of our test projects.

nuget/framework/nunit.nuspec
rprouse: We don't have any runners for Xamarin.Mac and MonoTouch is now Xamarin.iOS10 AFAIK.

As for NetStandard 1.4, our PCL is currently at least 1.3 if not 1.0 compliant, so I think we should drop PCL for the next release and switch it to .NET Standard 1.0 or 1.3. Checking other common packages, they have already done so.

ChrisMaddock: Sounds like a plan! 🙂

PR #1968, opened 2017-01-09
Implement Has.Exactly(n).Items

### LINKED ISSUE #1668: Need implementation-independent way to test number of items in a collection

CharliePoole:
Fixes #1668 

I took @jnm2's branch and imported it, then added some cleanup and further tests. This is a useful feature and has been hanging out there for a while so I'd like to get it reviewed and into the release if possible.

@jnm2 Basically, you were treading in some fragile areas of the code, which will need some further work in the near future. However, I think what I've done will hold us for the release.

@nunit/core-team Please review asap.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
@@ -31,9 +31,19 @@ namespace NUnit.Framework.Constraints
     /// item in a collection, succeeding only if a specified
     /// number of items succeed.
     /// </summary>
-    public class ExactCountConstraint : PrefixConstraint
+    public class ExactCountConstraint : Constraint
     {
-        private int expectedCount;
+        private readonly int _expectedCount;
+        private readonly IConstraint _itemConstraint;
+
+        /// <summary>
+        /// Construct a standalone ExactCountConstraint
+        /// </summary>
+        /// <param name="expectedCount"></param>
+        public ExactCountConstraint(int expectedCount)
+        {
+            _expectedCount = expectedCount;
+        }
 
         /// <summary>
         /// Construct an ExactCountConstraint on top of an existing constraint
@@ -43,12 +53,10 @@ namespace NUnit.Framework.Constraints
         public ExactCountConstraint(int expectedCount, IConstraint itemConstraint)
             : base(itemConstraint)
         {
-            this.expectedCount = expectedCount;
-            this.DescriptionPrefix = expectedCount == 0
-                ? "no item"
-                : expectedCount == 1
-                    ? "exactly one item"
-                    : string.Format("exactly {0} items", expectedCount);
+            Guard.ArgumentNotNull(itemConstraint, "itemConstraint");
+
+            _itemConstraint = itemConstraint.Resolve();
+            _expectedCount = expectedCount;
         }
 
         /// <summary>
@@ -63,11 +71,36 @@ namespace NUnit.Framework.Constraints
                 throw new ArgumentException("The actual value must be an IEnumerable", "actual");
 
             int count = 0;
-            foreach (object item in (IEnumerable)actual)
-                if (BaseConstraint.ApplyTo(item).IsSuccess)
+            if (_itemConstraint == null)
+            {
+                foreach (object item in (IEnumerable)actual)
                     count++;
+            }
+            else
+            {
+                foreach (object item in (IEnumerable)actual)
+                    if (_itemConstraint.ApplyTo(item).IsSuccess)
+                        count++;
+            }
 
-            return new ConstraintResult(this, actual, count == expectedCount);
+            return new ConstraintResult(this, actual, count == _expectedCount);
+        }
+
+        /// <summary>
+        /// The Description of what this constraint tests, for
+        /// use in messages and in the ConstraintResult.
+        /// </summary>
+        public override string Description
+        {
+            get
+            {
+                var descriptionPrefix =
+                    _expectedCount == 0 ? "no item" :
+                    _expectedCount == 1 ? "exactly one item" :
+                    string.Format("exactly {0} items", _expectedCount);
+
+                return _itemConstraint != null ? PrefixConstraint.FormatDescription(descriptionPrefix, _itemConstraint) : descriptionPrefix;
+            }
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/Operators/BinaryOperator.cs
+++ b/src/NUnitFramework/framework/Constraints/Operators/BinaryOperator.cs
@@ -48,7 +48,7 @@ namespace NUnit.Framework.Constraints
         {
             get
             {
-                return RightContext is CollectionOperator
+                return RightContext is CollectionOperator || RightContext is ExactCountOperator
                     ? base.LeftPrecedence + 10
                     : base.LeftPrecedence;
             }
--- a/src/NUnitFramework/framework/Constraints/Operators/ExactCountOperator.cs
+++ b/src/NUnitFramework/framework/Constraints/Operators/ExactCountOperator.cs
@@ -29,7 +29,7 @@ namespace NUnit.Framework.Constraints
     /// Represents a constraint that succeeds if the specified 
     /// count of members of a collection match a base constraint.
     /// </summary>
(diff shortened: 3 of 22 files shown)


src/NUnitFramework/framework/Constraints/Operators/ExactCountOperator.cs
```diff
@@ -29,7 +29,7 @@ namespace NUnit.Framework.Constraints
     /// Represents a constraint that succeeds if the specified 
     /// count of members of a collection match a base constraint.
     /// </summary>
-    public class ExactCountOperator : CollectionOperator
+    public class ExactCountOperator : SelfResolvingOperator
```
CharliePoole: Ideally, this should be a CollectionOperator, which would eliminate some of the hacks I had to add. However, it also needs to be a SelfResolvingOperator but it can't be both. In the future, we should probably eliminate SelfResolvingOperator since it's only used as a marker. Operators that are self-resolving could indicate it by use of a  property.

src/NUnitFramework/framework/Constraints/Operators/ExactCountOperator.cs
```diff
@@ -39,17 +39,26 @@ public class ExactCountOperator : CollectionOperator
...
         /// </summary>
-        public override IConstraint ApplyPrefix(IConstraint constraint)
+        /// <param name="stack"></param>
+        public override void Reduce(ConstraintBuilder.ConstraintStack stack)
         {
-            return new ExactCountConstraint(expectedCount, constraint);
+            if (RightContext == null || RightContext is BinaryOperator)
+                stack.Push(new ExactCountConstraint(expectedCount));
+            else
+                stack.Push(new ExactCountConstraint(expectedCount, stack.Pop()));
```
CharliePoole: @jnm2 Checking for an empty stack works in almost all cases but breaks the case where the expression ends with `.And.Exactly(n).Items`. The test using RightContext is how it's done elsewhere and works for all cases I've found.

src/NUnitFramework/framework/Constraints/Operators/BinaryOperator.cs
```diff
@@ -48,7 +48,7 @@ public override int LeftPrecedence
         {
             get
             {
-                return RightContext is CollectionOperator
+                return RightContext is CollectionOperator || RightContext is ExactCountOperator
```
CharliePoole: Could this be any uglier?!? It wouldn't be needed if ExactCountOperator were a CollectionOperator, of course. However, the original test is almost as bad. We increase the precedence of binary operators `And` and `Or` when followed by a collection operator. This is what allows us to write

```C#
    Exactly(3).EqualTo("Charlie").And.Exactly(1).EqualTo("Fred")
```
which means __almost but not quite__ the same thing as

```C#
    Exactly(3).EqualTo("Charlie").Or.EqualTo("Fred")
```

It would be better if we could figure out a way to avoid this context sensitivity in the grammar.

CharliePoole:
@rprouse Can we get this into the release?

jnm2:
Cool. Now should I start a new issue for https://github.com/nunit/nunit/pull/1780#issuecomment-252793328?

rprouse:
We can get this into the release. I've just made the additional changes to the .NET Standard PR and will merge those. I'll review this now and also do a quick pass through the remaining PRs. Once that is done and everything merged, I will continue with the release, probably starting this evening EST.

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
```diff
@@ -43,12 +53,10 @@ public class ExactCountConstraint : PrefixConstraint
         public ExactCountConstraint(int expectedCount, IConstraint itemConstraint)
             : base(itemConstraint)
         {
-            this.expectedCount = expectedCount;
-            this.DescriptionPrefix = expectedCount == 0
-                ? "no item"
-                : expectedCount == 1
-                    ? "exactly one item"
-                    : string.Format("exactly {0} items", expectedCount);
+            Guard.ArgumentNotNull(itemConstraint, "itemConstraint");
```
rprouse: Not to be fixed for this PR, but did we tentatively decide in another project to start using C# 6 for all projects or just that project? If so, we should start using `nameof()` wherever possible.

CharliePoole: @rprouse  I proposed and made the decision for the gui because it's "my" project. I didn't deign to propose it for the framework. :smile:

Seriously, I was __hoping__ that such an exercise could be something to help get the incipient core team activated. I suppose somebody has to create a proposal and shepherd it through them if that's going to happen. I think the standard should allow for differences in individual projects that need to preserve more backward compatibility - for example if we started working again on CF - but that individual projects should need to get permission to use a higher level than the standard

That said, as far as I can see, the framework is the project that would have the most chance of needing backward compatibility.

CharliePoole:
Everything I wanted to get in is in. 😄

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
ChrisMaddock: > I suppose somebody has to create a proposal and shepherd it through them if that's going to happen.

Ok - I'll bite. ;-)

I've posted a thread on NUnit Developers, as I think it's time for a full cross-project decision, and I think this is something that would be advantageous to discuss in a forum accessible to all team members.

https://groups.google.com/forum/#!topic/nunit-developer/m00Oeo9bSqA

CharliePoole: Well, I was looking for a core team member to bite. :-)

I didn't expect it to go to the public list but it's actually a good idea to get feedback before deciding. Note that the nunit-developer list is for anyone who is __interested in__ NUnit development and has quite a few members.

At the moment you can only contact all the core team members by typing their names. I suggested setting up a list of some kind but somebody thought we had too many already. However, since there are only five of us, you only have to type four names so it's not so bad.

Now that we are between releases for a day or two I'll try to get out my draft of the governance document for review.

Issue #1965, opened 2017-01-08
TestContext does not flow in async method


Suremaker:
Example tests recreating issue:
```C#
[TestFixture]
public class TestContext_async_issue
{
    [Test]
    public async Task TestContext_Out_should_flow_with_async_execution()
    {
        var expected = TestContext.Out;
        await Task.Yield();
        Assert.AreSame(expected, TestContext.Out);
    }

    [Test]
    public async Task TestContext_CurrentContext_should_flow_with_async_execution()
    {
        var expected = TestContext.CurrentContext;
        await Task.Yield();
        Assert.AreSame(expected, TestContext.CurrentContext);
    }
}
```

The side effect of that behaviour is that it is not possible to use `TestContext.WriteLine()` after await as it causes `NullReferenceException`.

Tested with `NUnit 3.5.0`, `net46` and `netcoreapp1.0`

Suremaker:
The problem is that `TestExecutionContext` is holding current context with `[ThreadStatic]` attribute ([source code link](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/TestExecutionContext.cs#L182)).

While it works fine in scenarios where multiple tests are executed in parallel (but each in given thread at a time), in async scenarios it will not work as each method continuation may execute in other thread.
Replacement of `[ThreadStatic]` with [AsyncLocal<T>](https://msdn.microsoft.com/en-us/library/dn906268(v=vs.110).aspx) would be solution here, but unfortunately this class is available since .net 4.6 which is a pain for frameworks supporting older net versions :(

rprouse:
Thanks for the analysis. It doesn't sound like it is going to be an easy fix.

Suremaker:
For pre 4.6 there are some workarounds like:
* [CallContext](http://blog.stephencleary.com/2013/04/implicit-async-context-asynclocal.html) however it would not work below 4.5 (according to article) as well as should be replaced with AsyncLocal in >4.6
* custom implementation of [SynchronizationContext](https://msdn.microsoft.com/en-us/library/system.threading.synchronizationcontext(v=vs.110).aspx) however implementing it right in pre 4.5 will be difficult, as way how tasks were supporting SynchronizationContext flow was buggy - it is available since .net 2.0 however.

jhickson:
Could the issue I reported a few days ago in #1952 be explained by this? Though there .net46 is fine - it's just .netcoreapp1.1 that is the issue.

[...]

Issue #1952, opened 2017-01-04
TestContext.Out null when used in task with .NET Core


jhickson:
Hi,

I have a test project targeting both .NET Framework (4.6.2) and .NET Core (1.1) and am finding that for the latter `TextContext.Out` is null when used from a `Task`. The code below reproduces the issue:

```C#
[Test]
public void SomeTest()
{
    if (TestContext.Out == null)
    {
        Console.WriteLine("Out null");
    }
    else
    {
        Console.WriteLine("All ok");
    }

    Task.Factory.StartNew(() =>
                          {
                              if (TestContext.Out == null)
                              {
                                  Console.WriteLine("Out null");
                              }
                              else
                              {
                                  Console.WriteLine("All ok");
                              }
                           }).Wait();
}
```

For .NET Core 1.1 I get
```
All ok
Out null
```
printed at the console.

I'm using NUnitLite 3.5.0 with dotnet-test-nunit 3.4.0-beta-3. The failure occurs when using `dotnet test` from the command-line as well as from Visual Studio.

CharliePoole:
The problem is a bit more general than that. The same thing will happen if you create your own thread in any way at all. To avoid crashes, a new TestExecutionContext and TestContext are created as soon as you reference them. However, they are not initialized because NUnit has no opportunity to intervene until it's too late - after the coded is already run.

It has been a fundamental problem with NUnit for some time but shows up more frequently as more use is made of asychronicity. We need to look into this.

CharliePoole:
@jhickson I think this is now fixed in master. Can you try one of our myget builds >= 3.7.0-dev-03707 ?

jhickson:
@CharliePoole That's looking good now. Many thanks.

PR #1973, opened 2017-01-10
Implemented TestExecutionContext to use AsyncLocal<> for NETSTANDARD1_6

Suremaker:
This is a partial fix for #1965.
I tested the change later with project using `net46` as well as `netcoreapp1.0`.
The pull request contains also test covering #1952

### DIFF
--- a/src/NUnitFramework/tests/TestContextTests.cs
+++ b/src/NUnitFramework/tests/TestContextTests.cs
@@ -21,9 +21,10 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+using System;
 using System.IO;
 using System.Collections.Generic;
-#if ASYNC && !NET_4_0
+#if ASYNC
 using System.Threading.Tasks;
 #endif
 using NUnit.Framework.Interfaces;
@@ -141,7 +142,7 @@ namespace NUnit.Framework.Tests
         {
             Assert.That(TestContext.CurrentContext.Test.MethodName, Is.EqualTo("TestCanAccessItsOwnMethodName"));
         }
-        
+
         [TestCase(5)]
         public void TestCaseCanAccessItsOwnMethodName(int x)
         {
@@ -236,7 +237,7 @@ namespace NUnit.Framework.Tests
             Assert.That(fixture.Message, Is.EqualTo(TestResult.CHILD_ERRORS_MESSAGE));
             Assert.That(fixture.StackTrace, Is.Null);
         }
-#if ASYNC && !NET_4_0
+#if ASYNC
 
 #if PORTABLE
         [Ignore("Not implemented yet as TestExecutionContext uses TLS in portable implementation")]
@@ -245,7 +246,7 @@ namespace NUnit.Framework.Tests
         public async Task TestContextOut_ShouldFlowWithAsyncExecution()
         {
             var expected = TestContext.Out;
-            await Task.Yield();
+            await YieldAsync();
             Assert.AreSame(expected, TestContext.Out);
         }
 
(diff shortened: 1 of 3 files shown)


CharliePoole:
Would this be a partial fix to #1965?

Suremaker:
What I understand the `NetStandard` change would cover `netcoreapp1.0` and other frameworks which are not generic `.NETFramework`.

I checked in added tests that implementation which is already present for everything else than `PORTABLE` and `NETSTANDARD1_6` was actually working fine.

So I believe that the only problematic scenario now would be `PORTABLE` implementation.

Suremaker:
I have managed to make my tests passing for both `net46` and `netcoreapp1.0`.

jnm2:
According to [apisof.net](https://apisof.net/catalog/System.Threading.AsyncLocal%3CT%3E), `AsyncLocal` is available to `netcore10`, `netstandard13`, `net46`, `win10`, and newer.

CharliePoole:
@Suremaker So I guess that's a yes. :smile: 

We need to figure out an additional solution that covers .NET 2.0 through 4.5. But I'll review this as a fix for at least certain platforms.

oznetmaster:
Would using Thread.GetData/SetData work on all versions of .NET correctly for async?

Suremaker:
@CharliePoole I think that `CallContext.SetData(CONTEXT_KEY, context); / CallContext.GetData(CONTEXT_KEY);` solution is enough for .net 4.5 - the tests I have added covers net4.5 and they are passing.
I did not add any tests for .net 2.0 - 4.0 as they do not have async/await constructs, but it should be possible to write tests using tasks to verify if starting from net4.0.
The potentially biggest issue would be to implement it properly in Portable version - currently I have no idea how to do it better for it.
Also, it would be probably worth to add NET46 version and ensure that it uses AsyncLocal<> ?

@oznetmaster As far I am aware `Thread.GetData/SetData` uses thread local storage so it will not work with async/await in my opinion.

oznetmaster:
According to this article, that will not work (CallContext.SetData/GetData): http://stackoverflow.com/questions/13010563/using-threadstatic-variables-with-async-await

jnm2:
> Also, it would be probably worth to add NET46 version and ensure that it uses `AsyncLocal<>` ?

I think so, either that or build for .NET Standard 1.3 and run tests on the .NET Framework runtime 4.6.
This whole multi-targeting thing is about to get ridiculously easier with MSBuild 15.

jnm2:
@oznetmaster is right, but it should be possible to write a test proving this.

The weird part is we have async language support polyfillsbback to .NET Framework 3.5 but don't they come with an AsyncLocal?

Suremaker:
@oznetmaster @jnm2 It really look like CallContext.SetData/GetData should not work but so far all the tests I created passed so far. There is an other article (linked also on stackoverflow) describing that CallContext.LogicalSetData/GetData could make the work for .net 45: http://blog.stephencleary.com/2013/04/implicit-async-context-asynclocal.html
So I imagine it would be good to write some tests recreating this issue with CallContext.SetData/GetData and try to fix it with other approach.

rprouse:
>I did not add any tests for .net 2.0 - 4.0 as they do not have async/await constructs, but it should be possible to write tests using tasks to verify if starting from net4.0.

Our 4.0 framework and its tests reference the BCL library and have async/await support, so you should be able to enable your tests for the 4.0 framework.

Suremaker:
I have updated tests to cover .net40 and added two more complicated scenarios where async method branches and executes it's bits in parallel.
All the tests came green on my box - also because of this reason I have not changed `CallContext.SetData/GetData` as I cannot prove them not working.

The last scenario that comes to my mind is to run those tests on a machine with only .net40 installed - maybe it will reveal issues with CallContext that I cannot reproduce as my machine has .net46 which is in-place replacement for .net45 / .net40 (and it may changed the behaviour of those classes). I do not have however any box with .net40 only to try it out.

Suremaker:
Is there anything else I should do here in order to accept the pull request?

rprouse:
@Suremaker we are in the middle of a release cycle, so it will take us a bit longer than normal to review pull requests and move things along. There is a lot going on in the various projects right now. Please bare with us 😄

src/NUnitFramework/tests/TestContextTests.cs
```diff
@@ -233,6 +237,123 @@ public void TestContextStoresFailureInfoForOneTimeTearDown()
             Assert.That(fixture.Message, Is.EqualTo(TestResult.CHILD_ERRORS_MESSAGE));
             Assert.That(fixture.StackTrace, Is.Null);
         }
+#if ASYNC
+
+#if PORTABLE
+        [Ignore("Not implemented yet as TestExecutionContext uses TLS in portable implementation")]
```
CharliePoole: My first thought was that you should use a PlatformAttribute here, but portable isn't really a platform but a combination of platforms. Not sure how to deal with that any better than you have done. 😞

rprouse:
@CharliePoole looks like we were reviewing at the same time. This looks good to me. I am pretty sure we will drop PCL in favor of NET Standard 1.0 or 1.1, so that omission doesn't worry me. Does the @nunit/core-team agree with that?

Issue #1651, opened 2016-07-07
Too little information passed on ReflectionTypeLoadException in DefaultTestAssemblyBuilder.GetCandidateFixtureTypes


chm-tm:
When a required assembly isn't found, an exception like this is thrown by the Console Runner and framework 3.4.1:
```
    1) Invalid : [...]\UnitTestProject1.dll
    Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.
```
Since I didn't initially guess the root cause, I used the NUnit source code to track down the `ReflectionTypeLoadException`, which is thrown in `DefaultTestAssemblyBuilder.GetCandidateFixtureTypes` upon calling `GetTypes` on `assembly`.
It would be helpful to propagate information from the `LoaderExceptions` into the test result.
Repro sample project: [UnitTestProject1.zip](https://github.com/nunit/nunit/files/351691/UnitTestProject1.zip)

ChrisMaddock:
This has been brought up before as #1238. Will close this issue as a duplicate, the sample project should be useful for whoever eventually works on it however - good shout!

PR #1976, opened 2017-01-11
Print loader exception messages on ReflectionTypeLoadException

### LINKED ISSUE #1238: Feature request: Print LoaderExceptions when fixture loading fails
Hi.

I have a problem in my test fixture - when trying to run it with the console runner, I get 

```
Agent process 13984 starting
Running under version 4.0.30319.42000, Net 4.5
Agent process 13984 exiting

Errors and Failures

1) Invalid : c:\PathToMy\Tests.dll
Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.
```

Obviously this error message is not very helpful - the real value would be to print the contents of the LoaderExceptions.

So I cloned the nunit code and fired up the the console runner - only to realize that the problem of course happens inside the nunit agent. Setting the --debug-agent flag did not help - the "c:\program files..\nunit3-agent" was loaded. I then realized that I had to explicitly compile the agent (so that it exists in the same folder as my debug console). This was of course all natural and intentional, but not obvious on first go. Just realized that the command to get all the debug dll's compiled is `build.cmd build.cake Default Debug` :smile: 

Further investigation showed that if I _explicitly_ set the framework: --framework=net-4.5 - I get a nice BadImageFormatException (Because that was the error).

However - if I do not explicitly specify the framework, I get the generic, not very helpfull error message.
Ideally NUnit should print the content of the loader exception directly...

ChrisMaddock:
Fixes #1238.

This one kept biting me. I think this has mostly been encountered due to issues with the console/engine (e.g. the framework flag problem) - however, can also occur under 'user error' - such as #1651. This prints out the loader exceptions when a ReflectionTypeLoadException is encountered, in the same way as we would handle an innerexceptions. It changes the (console) output from:

```
1) Invalid : C:\Users\ChristopherMaddock\Downloads\UnitTestProject1\UnitTestProject1\bin\Debug\UnitTestProject1.dll
Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.
```

to:
```
1) Invalid : C:\Users\ChristopherMaddock\Downloads\UnitTestProject1\UnitTestProject1\bin\Debug\UnitTestProject1.dll
Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.
  ----> Could not load file or assembly 'ClassLibrary1, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified.
```

I changed two occurrences in the DefaultTestAssemblyBuilder - I'm not sure if this covers everything, but I did a quick check for other places where exception messages are recorded without using ExceptionHelper, and they mostly seem to be specific (unrelated) exception types.

### DIFF
--- a/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
+++ b/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
@@ -116,7 +116,7 @@ namespace NUnit.Framework.Api
             {
                 testAssembly = new TestAssembly(assemblyName);
                 testAssembly.RunState = RunState.NotRunnable;
-                testAssembly.Properties.Set(PropertyNames.SkipReason, ex.Message);
+                testAssembly.Properties.Set(PropertyNames.SkipReason, ExceptionHelper.BuildFriendlyMessage(ex));
             }
 
             return testAssembly;
@@ -175,7 +175,7 @@ namespace NUnit.Framework.Api
             {
                 testAssembly = new TestAssembly(assemblyPath);
                 testAssembly.RunState = RunState.NotRunnable;
-                testAssembly.Properties.Set(PropertyNames.SkipReason, ex.Message);
+                testAssembly.Properties.Set(PropertyNames.SkipReason, ExceptionHelper.BuildFriendlyMessage(ex));
             }
 
             return testAssembly;
--- a/src/NUnitFramework/framework/Internal/ExceptionHelper.cs
+++ b/src/NUnitFramework/framework/Internal/ExceptionHelper.cs
@@ -71,22 +71,26 @@ namespace NUnit.Framework.Internal
         // TODO: Move to a utility class
         /// <summary>
         /// Builds up a message, using the Message field of the specified exception
-        /// as well as any InnerExceptions.
+        /// as well as any InnerExceptions. 
         /// </summary>
         /// <param name="exception">The exception.</param>
         /// <returns>A combined message string.</returns>
         public static string BuildMessage(Exception exception)
         {
-            StringBuilder sb = new StringBuilder();
-            sb.AppendFormat(CultureInfo.CurrentCulture, "{0} : {1}", exception.GetType().ToString(), exception.Message);
+            const bool isFriendlyMessage = false;
+            return BuildMessage(exception, isFriendlyMessage);
+        }
 
-            foreach (Exception inner in FlattenExceptionHierarchy(exception))
-            {
-                sb.Append(Environment.NewLine);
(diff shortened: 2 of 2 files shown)


CharliePoole:
This looks good to me with one reservation. Isn't really that friendly to leave out the name if the exception? It works in your example because the message is clear. Is it equally clear with BadImageFormat? What if (I don't have an example) there's no message?

ChrisMaddock:
The exception names were excluded to match existing functionality - I initially used the existing method, but that broke a suite of tests. I was going to update the tests, but then wasn't sure if the 'friendly' messages were preferred for user-facing output, e.g. console summary/gui.

In the simpler case of a missing file, we would currently show this:
```
1) Invalid : C:\Users\ChristopherMaddock\tests.dll
Could not load file or assembly 'tests.dll, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified.
```

We could instead keep the exception name in, and show this:
```
1) Invalid : C:\Users\ChristopherMaddock\tests.dll
System.Exception.FileNotFoundException : Could not load file or assembly 'tests.dll, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified.
```

That was my first solution but it breaks tests, so I added the additional method to replicate current behaviour. I'd personally prefer the latter, but thought it might be unpopular.

CharliePoole:
Are you testing with NUnitLite or NUnit3-Console? It just occurred to me that the file not found situation shouldn't actually happen with the console, since it's handled before the framework is ever called.

ChrisMaddock:
Console, but this isn't what occurs when the primary assembly is missing, the above was just an example. 

This is a case where a TestFixture class references another assembly which is missing - so the FileNotFound isn't encountered until the assembly is loaded. I don't think we can do anything to check that before runtime, can we?

@chm-tm provided this case as a repo for this issue on https://github.com/nunit/nunit/issues/1651. It's definitely user-error, so I think the right solution here is just to present the full error message for the user to deal with.

CharliePoole:
I agree. While the engine could theoretically probe all references in the test assembly I think that would be too much work and also unreliable, since the engine doesn't use reflection for it's probing.

I'm OK with your fix if you are ready to merge.

ChrisMaddock:
Thanks. 😄

ChrisMaddock:
Maybe we weren't quite finished. Re:

> This looks good to me with one reservation. Isn't really that friendly to leave out the name if the exception? 

This PR doesn't change current behaviour on that issue. I'm fairly apathetic personally, but it could be a new issue if you thought it was worthwhile, @CharliePoole? Will leave that one with you.

CharliePoole:
No... I'm OK as is.

PR #1977, opened 2017-01-12
#1000 Multiple Author Attribute

### LINKED ISSUE #1000: Support multiple Author attributes per test
Support multiple Author attributes per test

https://github.com/nunit/nunit/wiki/Author-Attribute

Note: you can currently only have one Author attribute per fixture or test.

JustinRChou:
resolves #1000 

Update to Author Attribute to allow for multiple use.

### DIFF
--- a/src/NUnitFramework/tests/Attributes/AuthorTests.cs
+++ b/src/NUnitFramework/tests/Attributes/AuthorTests.cs
@@ -95,5 +95,32 @@ namespace NUnit.Framework.Tests.Attributes
             var testCase = (Test)parameterizedMethodSuite.Tests[0];
             Assert.AreEqual("Charlie Poole", testCase.Properties.Get(PropertyNames.Author));
         }
+
+        #region Multiple Authors
+        [Test]
+        public void TestMethodMultipleAuthors()
+        {
+            Test test = TestBuilder.MakeTestFromMethod(FixtureType, "TestMethodMultipleAuthors");
+            var authors = test.Properties[PropertyNames.Author];
+            Assert.That(authors, Has.Exactly(3).Items);
+            Assert.AreEqual(authors[0], "Rob Prouse <rob@prouse.org>");
+            Assert.AreEqual(authors[1], "Charlie Poole <charlie@poole.org>");
+            Assert.AreEqual(authors[2], "NUnit");
+        }
+
+        [Test]
+        public void TestFixtureMultipleAuthors()
+        {
+            var suite = new TestSuite("suite");
+            suite.Add(TestBuilder.MakeFixture(FixtureType));
+            var mockFixtureSuite = (TestSuite)suite.Tests[0];
+            var authors = mockFixtureSuite.Properties[PropertyNames.Author];
+            Assert.That(authors, Has.Exactly(3).Items);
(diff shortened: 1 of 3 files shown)


JustinRChou:
I wasn't sure how should I go about testing the XML output of the authors.

And should TestCaseSource have the author property?

src/NUnitFramework/tests/Attributes/AuthorTests.cs
```diff
@@ -95,5 +95,33 @@ public void AuthorOnTestCase()
...
+        [Test(Author = "Rob Prouse <rob@prouse.org>"),Author("Charlie Poole", "charlie@poole.org")]
+        [Author("NUnit")]
+        public void TestMethodMultipleAuthors()
+        {
+            Test test = TestBuilder.MakeTestFromMethod(FixtureType, "TestMethodMultipleAuthors");
+            var authors = test.Properties[PropertyNames.Author];
+            Assert.That(authors, Has.Exactly(3).Items);
+            Assert.AreEqual(authors[0], "Rob Prouse <rob@prouse.org>");
+            Assert.AreEqual(authors[1], "Charlie Poole <charlie@poole.org>");
+            Assert.AreEqual(authors[2], "NUnit");
```
CharliePoole: This is OK but I'd normally just do...

```C#
Assert.That(Test.Properties[PropertyNames.Author], Is.EqualTo(
    new [] { "Rob Prouse <rob@prouse.org", "Charlie Poole <charlie@poole.org", "NUnit" }));
```

JustinRChou: I am probably going to drop 
`Assert.That(authors, Has.Exactly(3).Items);`
as well.

And regarding TestCaseSource Author property, thoughts?

CharliePoole:
Sorry, I missed the questions...

The XML can be tested by converting the test that is built to XML and looking at the elements. However, we test that properties are reported correctly in the XML elsewhere and you have tested here that the properties are correctly set, so you __should__ be OK. As a final manual test, look at the TestResul.xml for your run of the framework tests. You should see the multiple property entries for the fixture and the method that have multiple authors.

I'm not a fan of Author on TestCaseSource. I don't think there's a need to identify authorship at every level. Frankly, I think we already took it too far with TestCase.

src/NUnitFramework/tests/Attributes/AuthorTests.cs
CharliePoole: Right. My change would drop the five lines. Or four if you use authors as a temporary.

JustinRChou:
Made the changes.
I just found it odd that TestCase had more properties than TestCaseSource.

src/NUnitFramework/tests/Attributes/AuthorTests.cs
CharliePoole: We have code that tests the count and then each item in some of our tests, but it's old code that pre-dates use of EqualConstraint with arrays.

CharliePoole:
TestCase looks like it's similar to TestCaseSource but it represents __one__ test case, while TestCaseSource represents a bunch of them.

CharliePoole:
Feel free to merge once CI passes.

PR #1978, opened 2017-01-12
Update release dates in changes.txt

ChrisMaddock:
@dbenzhuser pointed out that the dates were wrong on the release notes in https://github.com/nunit/docs/issues/156. I've made the change here - do these also get copied anywhere else that I may have forgotten?


rprouse:
LOL, we missed that for two releases!

PR #1979, opened 2017-01-13
Remove SpecialValue class

### LINKED ISSUE #1963: Investigate removing SpecialValue
`SpecialValue.cs` has the comment:

    /// Null represents a null value, which cannot be 
    /// used as an argument to an attribute under .NET 1.x

We no longer support .NET 1.x, so we should check if there are any uncommented dependencies, and whether it can be removed. 

There's also a very similar `NUnitNullType` - perhaps these can be combined, if SpecialValue can't be removed.

CharliePoole:
Fixes #1963


ChrisMaddock:
Looks good to me. 👍

PR #1983, opened 2017-01-15
Add missing ConstraintExpression.Contain overload

polsys:
This works:
```cs
string str = "a";
Assert.That(str, Does.Contain("a"));
Assert.That(str, Does.Not.Contain("b"));
```

But this fails to compile (cannot convert from `int` to `string`):
```cs
int[] ints = { 2 };
Assert.That(ints, Does.Contain(2));
Assert.That(ints, Does.Not.Contain(1));
```

This PR adds the `Contain(object)` overload for consistency. The same can be achieved with at least `Has.No.Member()` and `Does.Not.Contains()`, so I don't object if this addition is considered unnecessary. This is somewhat related to #1758.

There are no test changes, since the `Contain(string)` overload is not tested either, as it duplicated work in `Contains()`, and now just calls it directly.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
@@ -737,6 +737,15 @@ namespace NUnit.Framework.Constraints
         }
 
         /// <summary>
+        /// Returns a new CollectionContainsConstraint checking for the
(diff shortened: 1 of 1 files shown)


CharliePoole:
I'm good with this change. It makes things more consistent. Thanks for doing it.

We need one more committer to approve it before merging.

Issue #1971, opened 2017-01-10
Switch CHANGES.txt to Markdown


rprouse:
Our `CHANGES.txt` is currently nearly in Markdown format except for the headings. When the changes are copied into releases or the release notes in the docs, they need to be modified to Markdown format. I propose that we just switch `CHANGES.txt` to `CHANGES.md` and format it accordingly.

I am putting this out there as an idea. If the @nunit/core-team likes the idea, we can create issues for the other repos that have CHANGES files.

CharliePoole:
The changes for release notes are pretty minimal. I've always thought of the file as primarily for the purpose of distribution to users in our packages, where text seems to make more sense.

ChrisMaddock:
I think markdown is readable enough as plain text - and anything which saves work is a bonus. I like the idea!

smwentum:
can i try?

ChrisMaddock:
Let's make a decision here, so we can handle @smwentum's pull request 😄 

@nunit/core-team @nunit/framework-team - any further opinions? @CharliePoole - how averse to the change are you?

As above, my personal thoughts are that the benefit of doing this would outweigh the negative.

CharliePoole:
I really don't see this as a useful improvement for users. It's not a file that GitHub treats specially and we already have the release notes for those who view it online.

So the reason to do it is because we want to make our own lives easier. Fair enough. But why focus on making the __manual__ process we hope to replace easier?

Still, I'd be OK with it if we can remember to keep it legible as text by using only very simple markdown.

rprouse:
I only propose that we add the # characters for the titles and change the extension to MD. I think the hash characters don't detract much from TXT readability. Users can also click on the `CHANGES.md` file in GitHub to see a formatted version.

Our target audience is also developers, many of which will have Markdown support enabled in Visual Studio.

ChrisMaddock:
I think ideally we'd want to keep the current format - save changing everything/having mixed formats for older releases. That would look like this, raw:

```
###NUnit 3.6 - January 9, 2017

This release of the framework no longer includes builds for Compact Framework or for SilverLight, but adds a .NET Standard 1.6 build. If anyone still using Compact Framework or SilverLight and would like to continue development on those versions of the framework, please contact the NUnit team.

####Framework

 * .NET Standard 1.6 is now supported

####Issues Resolved

 * 406 Warning-level Assertions
...
```

CharliePoole:
Fair enough... 

[...]

PR #1984, opened 2017-01-16
issue #1971

smwentum:
# Title

fixed some typos in the change.txt file and made change.md file

## Summary

fixed some typos in the change.txt file and made change.md file 

## Details
this is related to issue #1971  there were typos in the change .txt file and i made a change.md file

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-﻿NUnit 3.6 - January 9, 2017
+NUnit 3.6 - January 9, 2017
 
 This release of the framework no longer includes builds for Compact Framework or
 for SilverLight, but adds a .NET Standard 1.6 build. If anyone still using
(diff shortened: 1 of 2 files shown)


CharliePoole:
Issue #1971 is marked as an idea and is under discussion. When we decide which way to go, we will either close it as won't fix or change it to something else - probably documentation. You shouldn't put time in on ideas because it could be wasted. However, this is looking like it will go ahead and if so we'll review the PR.

FYI, there are typos in the CHANGES file because that's how users created their issues. I for one have always kept them as (mis)spelled by the users, so long as they are understandable. But there's something to be said for correct spelling too. 😄 

Again... we'll review this if we decide to go ahead with the change.

BTW, you can see more info about where a particular issue is in the pipleline if you use ZenHub.

rprouse:
If we go forward with this, I think we should be renaming changes.txt to .md, not maintaining two copies going forward. The MD file also needs the headers specified to match what we use in [the release notes](https://github.com/nunit/docs/wiki/Release-Notes).

Hold off on making the changes until we come to a decision on the future of #1971 though 😄

ChrisMaddock:
@smwentum - we've finished up the discussion on issue #1971 - sorry it wasn't clear that was still in progress when you submitted this! I've added a proper spec of what we think needs to be done in the comment below - are you still interested in making the changes? You could just add them as an extra commit on this branch. 

https://github.com/nunit/nunit/issues/1971#issuecomment-274031803

ChrisMaddock:
This is replaced by #2002

PR #1987, opened 2017-01-17
Pass NumberOfTestWorkers from nunitlite to the framework

### LINKED ISSUE #1986: NUnitLite ignores --workers option
I came across this working on #164. Changing the number of test workers with the command-line `--workers` option didn't have any effect on the duration of the test. On investigation, it's clear that although the option is accepted there is no code to transmit it to the framework as a setting.

CharliePoole:
Fixes #1986


ChrisMaddock:
Looks good to me. 👍

Issue #1990, opened 2017-01-18
Add namespace filter


CharliePoole:
As a pre-requisite to nunit/nunit-console#162, implement a namespace filter in the framework.

Modify the TSL implementation in nunitlite to handle namespace as a keyword.

CharliePoole:
Closed via #1995 

@tom-dudley The exact comment phrase "Fixes #1990" would have been needed in the PR in order for ths to be closed automatically by GitHub.

jnm2:
`Fixes` and `Closes` both work. I tend to say `fixes` for bugs and `closes` for features.

PR #1995, opened 2017-01-19
Add namespace filter

tom-dudley:
Implementation of: https://github.com/nunit/nunit/issues/1990

I've added what I believe are the required changes on the framework side to add support for the namespace keyword. I'm not sure what the deal with this getting added to the different versions of .NET, but since I believe everything added it compatible, I went ahead with adding to all of them.

nunit-console change dependent on this PR is here: https://github.com/nunit/nunit-console/pull/165

### DIFF
--- a/src/NUnitFramework/framework/Internal/Filters/NamespaceFilter.cs
+++ b/src/NUnitFramework/framework/Internal/Filters/NamespaceFilter.cs
@@ -63,10 +63,5 @@ namespace NUnit.Framework.Internal.Filters
         {
             get { return "namespace"; }
         }
-
-        private string GetNamespaceFromTest(Test test)
-        {
-            return test.TypeInfo.Namespace;
-        }
     }
 }
--- a/src/NUnitFramework/framework/Internal/Filters/NamespaceFilter.cs
+++ b/src/NUnitFramework/framework/Internal/Filters/NamespaceFilter.cs
@@ -1,0 +1,72 @@
+﻿// ***********************************************************************
+// Copyright (c) 2015 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using NUnit.Framework.Interfaces;
+
+namespace NUnit.Framework.Internal.Filters
+{
+    /// <summary>
+    /// ClassName filter selects tests based on the class FullName
+    /// </summary>
+#if !PORTABLE && !NETSTANDARD1_6
+    [Serializable]
+#endif
+    public class NamespaceFilter : ValueMatchFilter
+    {
+        /// <summary>
+        /// Construct a NamespaceFilter for a single namespace
+        /// </summary>
+        /// <param name="expectedValue">The namespace the filter will recognize.</param>
+        public NamespaceFilter(string expectedValue) : base(expectedValue) { }
+
+        /// <summary>
+        /// Match a test against a single value.
+        /// </summary>
+        public override bool Match(ITest test)
+        {
+            string containingNamespace = null;
+
+            if (test.TypeInfo != null)
+            {
+                containingNamespace = test.TypeInfo.Namespace;
+            }
+
+            return Match(containingNamespace);
+        }
+
+        /// <summary>
+        /// Gets the element name
+        /// </summary>
+        /// <value>Element name</value>
+        protected override string ElementName
+        {
+            get { return "namespace"; }
+        }
+
+        private string GetNamespaceFromTest(Test test)
+        {
+            return test.TypeInfo.Namespace;
+        }
+    }
+}
--- a/src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
+++ b/src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
@@ -54,6 +54,7 @@ namespace NUnit.Framework.Api
 #if PORTABLE
         private static readonly string MOCK_ASSEMBLY_NAME = typeof(MockAssembly).GetTypeInfo().Assembly.FullName;
 #endif
+        private const string INVALID_FILTER_ELEMENT_MESSAGE = "Invalid filter element: {0}";
 
         private static readonly IDictionary<string, object> EMPTY_SETTINGS = new Dictionary<string, object>();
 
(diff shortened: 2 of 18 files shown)


tom-dudley:
Ah yes, good point! 

Is there going to be any issue with me doing an interactive rebase and force pushing back to here? I see I messed up some csproj files in that second commit.. 

Speaking of, what's the recommended way to do development on the framework and console at the same time? I ended up just creating a solution and pulling in the separate projects but I'm certain there's a better way!

tom-dudley:
I'm a bit confused about the various ExplicitMatchTests by the way. Looks like the tests are disabled in a few of the filters and enabling them makes them fail. Ideally I'd add a test case for the expected behaviour, but I can't tell what that should be!

CharliePoole:
No problem with that. We don't consider any branches as "published" except for master.

Generally, we try to work in the two projects separately, although you can experiment locally as you are doing. We do occasionally have issues of dependency but not here. Your console changes will work differently with a newer framework but it has to handle the older versions as well.

Once the framework changes are merged, the nuget dev package will be available on myget. That's why the console issue is labeled as blocked.

ChrisMaddock:
> I'm a bit confused about the various ExplicitMatchTests by the way. Looks like the tests are disabled in a few of the filters and enabling them makes them fail. Ideally I'd add a test case for the expected behaviour, but I can't tell what that should be!

Not totally sure what you're looking at here @tom-dudley - did this get resolved? If not, can you be more specific? I can't find the disabled tests?

tom-dudley:
Aha, I was about to say I couldn't find them, but here we go:

Note the absence of the [Test] attribute [here](https://github.com/nunit/nunit/blob/757feb7a721ee8ef21b20e76ada61d00834c3bc7/src/NUnitFramework/tests/Internal/Filters/FullNameFilterTests.cs#L63-L64)

Adding the attribute makes the test fail*, hence my confusion.

*It might actually pass on this one, but all the filter tests seem to have the attribute missing for this test. Some of them do fail though.

src/NUnitFramework/framework/Internal/Filters/NamespaceFilter.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// </summary>
+        public override bool Match(ITest test)
+        {
+            string containingNamespace = null;
+
+            if (test.TypeInfo != null)
+            {
+                containingNamespace = test.TypeInfo.Namespace;
+            }
+            else if (test is TestSuite || test is TestFixture)
```
jnm2: Could be simplified to `(test is TestSuite)`

src/NUnitFramework/framework/Internal/Filters/NamespaceFilter.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                if (test.HasChildren)
+                {
+                    containingNamespace = test.Tests[0].TypeInfo.Namespace;
+                }
+                else
+                {
+                    var outerMostClass = test.ClassName.Substring(0, test.ClassName.IndexOf("+"));
+                    containingNamespace = outerMostClass.Substring(0, outerMostClass.LastIndexOf("."));
```
jnm2: This will crash if `test.ClassName` is null, does not contain `+`, or does not contain `.`.

I don't know why `ClassName` would ever be null, but I just saw it happen. Then against I don't understand why `TypeInfo` would ever be null lol.

Could you add a test for a non-nested class in the global namespace, where the entirety of `ClassName` is something like `Foo`? An empty string should be returned for the namespace.

src/NUnitFramework/framework/Internal/Filters/NamespaceFilter.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            if (test.TypeInfo != null)
+            {
+                containingNamespace = test.TypeInfo.Namespace;
+            }
+            else if (test is TestSuite || test is TestFixture)
+            {
+                if (test.HasChildren)
+                {
+                    containingNamespace = test.Tests[0].TypeInfo.Namespace;
```
jnm2: I'm curious, do we have a good understanding of why `test.TypeInfo` is null and why `test.Tests[0].TypeInfo` would not be null?

src/NUnitFramework/framework/Internal/Filters/NamespaceFilter.cs
tom-dudley: @jnm2 I think you're right in that test.TypeInfo should be being used where possible. The reason it is null in some cases is that the construction of the `MySuite` test suite doesn't pass in any type information like is done in creating the test fixtures. Now I'm not sure if we can say for certain that this TypeInfo will always be present, but if it were the logic could simply be:

>             string containingNamespace = null;
> 
>             if (test.TypeInfo != null)
>             {
>                 containingNamespace = test.TypeInfo.Namespace;
>             }
> 
>             return Match(containingNamespace);

But this means in cases where a TestSuite is constructed without type information then a suite may be skipped when it shouldn't be.

jnm2: Given the [implementation of `ClassName`](https://github.com/nunit/nunit/blob/3.6/src/NUnitFramework/framework/Internal/Tests/Test.cs#L163-L178), what if you did:

```c#
if (test.TypeInfo != null)
    return Match(test.TypeInfo.Namespace);
else if (test.Method != null)
    return Match(test.Method.MethodInfo.DeclaringType.Namespace)
else
    return false;
```

src/NUnitFramework/framework/Internal/Filters/NamespaceFilter.cs
CharliePoole: @jnm2 A test suite is represented in the code in many different ways. That representation might be a class, in which case there will be a non-null `TypeInfo` member. But it could represent an assembly or a namespace or a group of methods. In those cases, there is not type to deal with.

NUnit tries hard to separate the concepts of Test, TestSuite etc. from their representation. But it's in the classes derived from Test that the two collide and have to be reconciled.

Some future test implementation might represent a test suite by an XML file or a test case by a class!

src/NUnitFramework/framework/Internal/Filters/NamespaceFilter.cs
CharliePoole: This should be fixed if we keep it, since a `TestFixture` __is__ a `TestSuite`.

However, the more fundamental question is why we have this branch at all. If there is no class type, then there is no namespace to match, so the test should not be matched.

src/NUnitFramework/framework/Internal/Filters/NamespaceFilter.cs
```diff
@@ -0,0 +1,84 @@
...
+                }
+                else
+                {
+                    var outerMostClass = test.ClassName.Substring(0, test.ClassName.IndexOf("+"));
+                    containingNamespace = outerMostClass.Substring(0, outerMostClass.LastIndexOf("."));
+                }
+            }
+
+            return Match(containingNamespace);
+        }
```
CharliePoole: It looks as if you are trying to make the `Match` method do the entire job of the filter. But it's only supposed to match those tests that have the namespace you ask for. It's not supposed to find either the antecedents or descendants of those tests. That's up to the parent classes, `ValueMatchFilter` and `TestFilter`. 

Take a look at how `TestNameFilter` does it as an example. It merely matches the name of a test. But when the filter is used, we automatically run all the tests above it in the hierarchy, starting with the assembly, and all the methods below it, if the name for example is a suite.

src/NUnitFramework/framework/Internal/Filters/NamespaceFilter.cs
jnm2: @CharliePoole Is it possible to have a `Method` without a `TypeInfo`? [`Test.ClassName`](https://github.com/nunit/nunit/blob/3.6/src/NUnitFramework/framework/Internal/Tests/Test.cs#L163-L178)

src/NUnitFramework/framework/Internal/Filters/NamespaceFilter.cs
tom-dudley: I think if `ITest` had a `Namespace` property then we could simply just call `Match(test.Namespace)`. Most of the code is really just determining what the namespace of the test actually is, something like:

> var namespace = GetNamespaceOfTest(test);
> Match(namespace);

The reason I was looking at the decedent of the test (in this case a TestSuite) was in order to get the namespace. Perhaps I'm way off the mark with the way I'm approaching this, but I don't see an obvious alternative for knowing what the namespace of a test is.

CharliePoole: Exactly right. But there's no need to get a namespace unless you are dealing with a test that has a TypeInfo. Anything that doesn't have a TypeInfo should not match at all.

A test without a TypeInfo that is the parent of a matched test will pass the filter and be executed. A test without a TypeInfo that is a child of a matched test will pass the filter and be executed. All this is taken care of by the Pass method. The match method only needs to find tests that match, which means, basically, that  TypeInfo != null and TypeInfo.Namespace == the specified namespace.

CharliePoole:
@tom-dudley Have you made any changes since the last reviews? We don't necessarily get notices unless you post a comment.

rprouse:
There are no commits since Jan 18th. @tom-dudley are you still working on this?

tom-dudley:
Sorry for the delay! I've simplified the `Match` method as @CharliePoole suggested. I hadn't realised about the `return Match(test) || MatchParent(test) || MatchDescendant(test);` in `TestFilter.cs`

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
```diff
@@ -223,6 +224,21 @@ public void Run_FileNotFound_ReturnsNonRunnableSuite()
...
 
+        [Test]
+        public void RunTestsAction_WithInvalidFilterElement_ThrowsArgumentException()
+        {
+            LoadMockAssembly();
+
+            var ex = Assert.Catch(() =>
+                {
+                    var invalidFilter = TestFilter.FromXml("<filter><invalidElement>foo</invalidElement></filter>");
+                    _runner.Run(this, invalidFilter);
```
CharliePoole: No need to run tests. Exception is thrown by TestFilter.FromXml.

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
```diff
@@ -223,6 +224,21 @@ public void Run_FileNotFound_ReturnsNonRunnableSuite()
...
+                });
+
+            Assert.That(ex, Is.TypeOf<ArgumentException>());
```
CharliePoole: You can use `Throws<ArgumentException>` rather than `Catch` and eliminate this line.

PR #1998, opened 2017-01-20
Re-enable use of --trace in nunitlite

### LINKED ISSUE #1994: NUnitLite runner crashing when --trace is specified
Using the nunitlite.runner with arguments `nunit.framework.tests.dll --trace:Info` causes null reference exception in `TextRunner.GetLogFileName()`

CharliePoole:
Fixes #1994 

I need trace to be able to work on parallel threads and found it wasn't working. I also found and fixed a few errors at the same time and refactored the various Execute overloads so I could better understand what was going on. NUnitLite seems to have accrued a lot of technical debt, but this is all I have time for at the moment.

I'd like to get this in as soon as it looks good so I can rebase my other work on top of it.

### DIFF
--- a/src/NUnitFramework/nunitlite-runner/Program.cs
+++ b/src/NUnitFramework/nunitlite-runner/Program.cs
@@ -35,14 +35,15 @@ namespace NUnitLite
     /// be used for test assemblies built against that framework version.
     /// 
     /// In the special case of the portable version of nunitlite-runner,
-    /// the program is a .NET 4.5 console application.
+    /// the program is a .NET 4.5 console application. In that case, we
+    /// create a ColorConsoleWriter, since the portable build can't do it.
     /// </summary>
     class Program
     {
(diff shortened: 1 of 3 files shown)


rprouse:
Looks good. We have so much code duplicated between NUnitLite, the console and even the other runners like .NET Core and Xamarin. I wonder if we should think about creating an additional project to contain this duplication. We've talked about it, but never done it. Someone would have to analyze the overlap.

To reduce the number of assemblies, we could always ILMerge the results in.

Just thinking out loud based on your technical debt comment. 😄

CharliePoole:
Cleaning up NUnitLite is attractive from a housekeeping perspective but seems fairly low priority from a user perspective. The nuget package for nunitlite has had a total of 131K downloads compared to 7.5M for the framework! I think we should avoid doing too much until we get rid of portable.

PR #2000, opened 2017-01-21
#1999 Fix for Multiple Author Ordering

### LINKED ISSUE #1999: Author Tests assume ICustomAttributeProvider.GetCustomAttributes return order is defined
After some research, it appears that the order of the attributes returned by ICustomAttributeProvider.GetCustomAttributes is undefined (random).

However, it seems like the assumption is made in Nunit that custom attributes are always returned in the order they appear in the source code.

For example, the AuthorAttribute allows multiple occurrences of the attribute.  The tests for multiple AuthorAttribute assumes a specific order for the values in the resulting Author element in the test's Properties.  That test can fail because the order is actually not defined.

JustinRChou:
Fixes #1999

Change Is.EqualTo to Is.EquivalentTo due to order issues.


CharliePoole:
Looks good. Thanks.

PR #2002, opened 2017-01-22
this is the change to make CHANGES.txt to CHANGES.md see issue #1971 for more details

smwentum:
See issue #1971 for more details


CharliePoole:
It looks like what we agreed to. I'm not happy that `<` has to be changed to `&lt` but it's obviously necessary. Maybe we can try to avoid use of angle brackets descriptions in the future.

CHANGES.md
```diff
@@ -51,7 +51,7 @@ Issues Resolved
  * 1901 Make nunitlite-runner Prefer32Bit option consistent across Debug/Release
  * 1904 Add .NET Standard 1.6 Dependencies to the Nuspec Files
  * 1907 Handle early termination of multiple assert block
- * 1911 Changing misleading comment that implies that every ICollection<T> is a list
+ * 1911 Changing misleading comment that implies that every ICollection&lt;T> is a list
```
rprouse: Odd that we never noticed that the `<` were getting eaten in the past. Good catch.

Might I suggest that since every use in this file is for a generic type, we convert 
```md
ICollection<T> to `ICollection<T>` instead of ICollection&lt;T>
```

Trying here, `ICollection<T>`.

This will make it much more readable as plain text.

CharliePoole: Good call!

smwentum: oh ok

CHANGES.md
```diff
@@ -51,7 +51,7 @@ versions of the framework, please contact the NUnit team.
  * 1901 Make nunitlite-runner Prefer32Bit option consistent across Debug/Release
  * 1904 Add .NET Standard 1.6 Dependencies to the Nuspec Files
  * 1907 Handle early termination of multiple assert block
- * 1911 Changing misleading comment that implies that every ICollection&lt;T> is a list
+ * 1911 Changing misleading comment that implies that every ICollection`<T>` is a list
```
rprouse: Minor change, but these should be `ICollection<T>` not ICollection`<T>`

ChrisMaddock:
@smwentum - are you able to make the correction @rprouse pointed out?

If not, perhaps one of us can cherry-pick and merge this in.

smwentum:
someone else can take it if they want

rprouse:
@ChrisMaddock do you have time to make this change?

ChrisMaddock:
I'm sure I can. 😄

ChrisMaddock:
Actually, @rprouse - if you've already got #2049 in place, I suggest this waits until afterwards. I can rebase this and update for 3.6.1, and then it won't be blocking.

ChrisMaddock:
I've finished this off as #2114.

Thanks for your help, @smwentum! 😄

jnm2:
> I'm not happy that < has to be changed to &lt but it's obviously necessary. Maybe we can try to avoid use of angle brackets descriptions in the future.

I'm pretty sure that `\<` is the recommended way to escape everything in markdown.
\<test>

jnm2:
`\<test>` is definitely more readable as plain text than `&lt;test>`, and they both yield `<test>`.

ChrisMaddock:
@jnm2 - see #2114 - @smwentum came up with the idea of escaping with backticks, which I think works even better. 😄

jnm2:
Yes, that does look better in that context! In cases where you don't want to use backticks, backslash should be the go-to escaping mechanism for everything in markdown. I don't know of any reason to resort to HTML entities or elements.

jnm2:
Heh, I'm wrong. there is one instance I know of where you need the HTML element `<code>`- when you want to insert links inside inline code:

```markdown
<code>[IReadOnlyDictionary](https://msdn.microsoft.com/en-us/library/hh136548.aspx)<[string](https://msdn.microsoft.com/en-us/library/system.string.aspx), [MyType](#just-an-example)></code>
```

Gives:
<code>[IReadOnlyDictionary](https://msdn.microsoft.com/en-us/library/hh136548.aspx)<[string](https://msdn.microsoft.com/en-us/library/system.string.aspx), [MyType](#just-an-example)></code>

CharliePoole:
Backticks seem to be the standard for GitHub markdown, rather than backslash.

@jnm2 your example is correct but isn't exactly markdown related: you're embedding a link in the markdown, so it has to follow the rules for links. The part between parentheses isn't interpreted as markdown. Of course, lots of folks could be confused by that nesting - even though we aren't. :smile:

jnm2:
> Backticks seem to be the standard for GitHub markdown, rather than backslash.

Backticks are for styling as code, and treat the contents as code:
`<test>`
`# Test`

Backslashes are for escaping:
\<test>
\# Test

jnm2:
Is this easier to read?

```markdown
<code>[IReadOnlyDictionary]<[string], [MyType]></code>

 [IReadOnlyDictionary]: https://msdn.microsoft.com/en-us/library/hh136548.aspx
 [string]: https://msdn.microsoft.com/en-us/library/system.string.aspx
 [MyType]: #just-an-example
```

<code>[IReadOnlyDictionary]<[string], [MyType]></code>

 [IReadOnlyDictionary]: https://msdn.microsoft.com/en-us/library/hh136548.aspx
 [string]: https://msdn.microsoft.com/en-us/library/system.string.aspx
 [MyType]: #just-an-example

Another option: 

```markdown
<code>[IReadOnlyDictionary][1]<[string][2], [MyType][3]></code>

 [1]: https://msdn.microsoft.com/en-us/library/hh136548.aspx
 [2]: https://msdn.microsoft.com/en-us/library/system.string.aspx
 [3]: #just-an-example
```

jnm2:
Technically [markdown includes HTML](https://daringfireball.net/projects/markdown/syntax), so HTML is always valid in any .md file. I haven't "left" the markdown syntax at any point, any more than I've stopped using Razor when I write HTML or C# in an ASP.NET razor view file.

CharliePoole:
@jmm2 What's in parentheses isn't markdown or html. It's a URL.

jnm2:
Okay, but that does not mean that this example is not markdown-related: `<code>[Test](#url)</code>`

The example is still markdown-related even though there is a URL between the parentheses because the `[]()` syntax is uniquely markdown's. I must be misunderstanding your point.

CharliePoole:
I believe (can't try it on my phone) they the parentheses are a special escape mechanism in that whatever is inside is not treated as markdown.

However, I don't think I disagreed with anything you said, just tried to put it in a different light. So, when the content is a URL, you obviously have to use escaping that works in URLs.

jnm2:
Ah, I see. Yet I don't agree that the \<code>-escaping has anything to do with URLs. I used it to put a URL in code, but I could also use the exact same tactic to put any markdown style in code, not just link markdown:

`<code>var *x* = 1;</code>`
<code>var *x* = 1;</code>

jnm2:
The point is, the \<code> HTML element is necessary. There's no other way in markdown to make part of the code bold, or italic, or a link, or a picture, etc.

CharliePoole:
No doubt of that. I wasn't actually responding to that part of the discussion.

jnm2:
Okay. :-)
To follow the subject change, this is interesting. When it comes to URLs, you can use markdown escaping rather than URL-escaping:

`[Test](#silly-url-with-\)-in-it)` -> [Test](#silly-url-with-\)-in-it)

Issue #2005, opened 2017-01-23
Exclude empty failure messages from results xml


ChrisMaddock:
From the discussion on #1955 - if the message property for a failure is an empty string, then the message element should not be written to the xml.

To reproduce:
    
    Assert.Fail("")

will currently produce:

    <message><![CDATA[]]></message>

@cristobalito is going to take a look at this one.

cristobalito:
So an interesting observation. If you specify a stacktrace, but no failure message, the `<message>` element is suppressed!

CharliePoole:
@cristobalito What do you mean by specifying a stacktrace?

cristobalito:
@CharliePoole Just going through the existing Failure tests (TestResultFailureTests.cs), trying out a few different combinations of values for message and stacktrace - the test sets these properties via SetResult. I appreciate non-test code will likely always have a reasonable value for stacktrace (or null).

My comment was just that I found it interesting that only supplying a stacktrace and a null message suppressed the `message`. Looking now at the code, that makes sense.

CharliePoole:
I see. I thought you meant it from a user perspective, rather than internally.

PR #2009, opened 2017-01-26
Suppress null or whitespace failure messages

cristobalito:
Also covers null or whitespace failure stack traces.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -645,10 +645,10 @@ namespace NUnit.Framework.Internal
         {
             TNode failureNode = targetNode.AddElement("failure");
 
-            if (Message != null)
+            if (Message != null && Message.Trim().Length > 0)
                 failureNode.AddElementWithCDATA("message", Message);
 
-            if (StackTrace != null)
+            if (StackTrace != null && StackTrace.Trim().Length > 0)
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -645,10 +645,10 @@ private TNode AddFailureElement(TNode targetNode)
...
 
-            if (Message != null)
+            if (Message != null && Message.Trim().Length > 0)
                 failureNode.AddElementWithCDATA("message", Message);
 
-            if (StackTrace != null)
+            if (StackTrace != null && StackTrace.Trim().Length > 0)
                 failureNode.AddElementWithCDATA("stack-trace", StackTrace);
 
             return failureNode;
```
CharliePoole: In theory, we could end up with an empty failure element, although it's not likely. We have to keep the element in any case because people often collect failed tests by looking for that element.

CharliePoole:
This looks good to me. Waiting for CI to pass and a second committer to review.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
cristobalito: Yeah - one way to get it would be to simply call Assert.Fail(), Assert.Fail(null), Assert.Fail(string.Empty), etc. Shouldn't this still result a test failure element, even if empty?

ChrisMaddock:
This fixes #2005

Issue #1638, opened 2016-06-30
TimeoutAttribute not available when targeting netcoreapp framework


rdghickman:
It seems that TimeoutAttribute is not available in the portable version of NUnit 3 when targeting .NET core with a netcoreapp test. This is extremely useful for any sort of test involving threading or integration test, so it would be really useful if this was made available.

Using: NUnit 3.4.0, dotnet-test-nunit 3.4.0-beta1.

rprouse:
We hope to reimplement timeout and other features that were dropped in the PCL when we add a netstandard version of the framework. They were dropped because timeout requires threads which aren't available in PCL.

jhickson:
Is there any timeline for enabling TimeAttribute and DelayedConstraint for the netstandard version? We're currently porting our libraries to target both .NET Framework and .NET Core but our test suites make quite heavy use of DelayedConstraint in particular such that porting them to support both platforms would be a pain.

rprouse:
The first step of porting to .NET Standard is done. Hopefully we can read the thread bases attributes Q1

ChrisMaddock:
@rprouse 

I was contemplating having a look at reimplementing these things in a Task-based, manner, but then found `System.Threading.Thread` is actually available for .NET Standard 1.3. 

https://www.nuget.org/packages/System.Threading.Thread/

Actual cross-platform support is apparantley [a little sketchy](https://github.com/dotnet/corefx/issues/2576#issuecomment-187184341), but not in a way that I think would affect us - our UWP runner will be using the PCL build (or subsequent netstandard 1.0) for the forseeable.

Would you be happy with us pulling in this package, or was this something you specifically didn't do? I can PR, at some point.

rprouse:
I was planning on pulling in the threading package as we start to tenable functionality. What about Thread utility isn't convertible?

ChrisMaddock:
Thread.Abort() and such isn't available in .NET Standard. This is used to kill the thread used for timeouts. 

DelayedConstraint is works with System.Threading.Thread - I could push that as a separate PR, and then revisit something Task based for Timeouts? 

The reason I didn't want to convert everything to be task based, is that they're not available pre-.NET4, so it would require two separate implementations for pre and post NET4. Hence why I think pulling in the package is good for the thread usages in DelayedConstraint, instead of converting that to use Task.Delay()'s instead of Thread.Sleep(). Thoughts?

abbotware:
[...]

PR #2010, opened 2017-01-27
Add DelayedConstraint in NetStandard 1.6 build

ChrisMaddock:
@rprouse - this is related to our discussion on #1638. I'd already done this as feasibility check, so thought it made sense to push it, and continue the discussion as to if we actually want to use it or not here - I'm not worried if we close instead. 😄 

This re-enables DelayedConstraint in the netstandard build. It includes the `System.Threading.Thread` package - so I also tidied up a few places where `Thread.Sleep()` had been excluded, as that's now available. 

This doesn't tackle timeout's, as I don't think Thread.Abort() is available in netstandard - notes on that are in #1638.

### DIFF
--- a/src/NUnitFramework/TestBuilder.cs
+++ b/src/NUnitFramework/TestBuilder.cs
@@ -179,7 +179,7 @@ namespace NUnit.TestUtilities
             // TODO: Replace with an event - but not while method is static
             while (work.State != WorkItemState.Complete)
             {
-#if PORTABLE || NETSTANDARD1_6
+#if PORTABLE
                 System.Threading.Tasks.Task.Delay(1);
 #else
                 Thread.Sleep(1);
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -232,11 +232,7 @@ namespace NUnit.Framework.Api
         /// <returns>True if the run completed, otherwise false</returns>
         public bool WaitForCompletion(int timeout)
         {
-#if !PORTABLE && !NETSTANDARD1_6
-            return _runComplete.WaitOne(timeout, false);
-#else
             return _runComplete.WaitOne(timeout);
-#endif
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/Constraint.cs
+++ b/src/NUnitFramework/framework/Constraints/Constraint.cs
@@ -290,7 +290,7 @@ namespace NUnit.Framework.Constraints
 
         #region After Modifier
 
-#if !PORTABLE && !NETSTANDARD1_6
+#if !PORTABLE
         /// <summary>
         /// Returns a DelayedConstraint.WithRawDelayInterval with the specified delay time.
         /// </summary>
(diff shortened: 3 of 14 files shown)


src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -232,11 +232,7 @@ public void RunAsync(ITestListener listener, ITestFilter filter)
         /// <returns>True if the run completed, otherwise false</returns>
         public bool WaitForCompletion(int timeout)
         {
-#if !PORTABLE && !NETSTANDARD1_6
-            return _runComplete.WaitOne(timeout, false);
-#else
             return _runComplete.WaitOne(timeout);
-#endif
         }
```
ChrisMaddock: These two methods do the same thing on all platforms I believe, just Portable and Netstandard don't have the .WaitOne(int, bool) overload. I don't think the conditional is necessary here.

src/NUnitFramework/TestBuilder.cs
```diff
@@ -179,7 +179,7 @@ public static ITestResult ExecuteWorkItem(WorkItem work)
             // TODO: Replace with an event - but not while method is static
             while (work.State != WorkItemState.Complete)
             {
-#if PORTABLE || NETSTANDARD1_6
+#if PORTABLE
                 System.Threading.Tasks.Task.Delay(1);
 #else
                 Thread.Sleep(1);
```
CharliePoole: In general, do we think it's better to use `Thread.Sleep` or `Task.Delay` where both are available? I'm wondering if we should change this generally in a future PR.

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
CharliePoole: I think the `.WaitOne(int, bool)` overload was missing from some platforms we no longer support.

src/NUnitFramework/TestBuilder.cs
ChrisMaddock: I think it might be worth making a decision on the future of PCL, before we decide that. We previously talked about it targeting netstandard 1.0, but I'm not sure if we have much need to target that over netstandard 1.3 - which would allow us to use `System.Threading` - and then `Thread.Sleep()` across all platforms. 

If we did that - I'd vote for reducing differences between platforms over any minor performance benefit `Task.Delay()` may give us. 😄

CharliePoole: That's fair enough for our framework code itself. For tests, it may be different. I replaced `Thread.Sleep` with `Task.Delay` in some of our tests to eliminate spurious failures. My general impression is that Sleep has more of a tendency to go extra long in some circumstances.

CharliePoole:
@ChrisMaddock Shall we go ahead and merge this?

src/NUnitFramework/TestBuilder.cs
jnm2: That impression matches my understanding which is that `Delay` means "schedule using a precise timer" and `Sleep` means "I have nothing particularly important to do, deprioritize me."

ChrisMaddock:
I think it can be - thanks!

Issue #164, opened 2014-08-05
Run test methods within a fixture in parallel


CharliePoole:
Currently, we only have implemented parallel execution at the fixture level. It should be possible to run individual methods of a fixture in parallel as well as individual test cases for a parameterized method.

We should be able to specify the exact level of parallelism on the fixture or method.

This issue was previously part of #66

agray:
I would have thought this was relatively easy to implement given the capability exists in the old MbUnit.

Parallel execution of Test cases is an excellent feature of MbUnit and JUnit. It is a major gaping hole in my opinion that it is not also part of NUnit.

CharliePoole:
Well it's in the spec and scheduled for post-3.0 implementation. If it were so easy, we probably would have done it. Not sure what connection you see with mbunit. The fact that they had it doesn't help us.

agray:
My understanding of the connection with MbUnit is two fold.

1. @rprouse was a key contributor to MbUnit in a past life.
2. MbUnit version 3 has parallel test case execution just like a JUnit 4 does... and NUnit 3 should have.

If Compaq could find a way to build a team in the 1980's to reverse engineer an IBM machine, surely there is a way to bring this functionality into NUnit.

CharliePoole:
This is getting a bit tiresome. A few points, already stated but apparently missed...

1. There is a plan to implement what you are asking for.

2. We decided as a team to schedule it at a certain point.

3. When it is scheduled has primarily to do with our assessment of the value of the feature as compared with other features.

4. There are several people on the NUnit team who could implement the feature.

5. They would be able to implement it, depending on priorities, out of their heads and would not need to copy it from anywhere.

I find your talk of "reverse engineering" very disturbing. Open source is only possible in a context where copyright is respected. So if you are suggesting that we might ignore the mbunit licensing terms, you are way off base.

rprouse:
While I did contribute many patches to MbUnit and used it for years, I was
never a key contributor. I wouldn't want my status to be misrepresented :)

As for reverse engineering, there isn't really any use here. NUnit works
entirely different than MbUnit did. We will get to this issue in due time,
but are approaching it cautiously because other similar issues might change
the way we need to implement this or even conflict.

CharliePoole:
A much more tactful comment than mine!

Our current priorities in the world of parallel execution are:
1. Parallel process execution
2. Exclusion groups
3. Parallel method (in one fixture) execution.

This seems to represent the order of greatest usefulness to users.

[...]

PR #2011, opened 2017-01-28
First steps toward parallel test cases

### LINKED ISSUE #2003: Better user info about ParallelizableAttribute and ParallelScope
This comes up as I work on issue #164 and I'll implement it in conjunction with that is.sue. However, I think it needs to have separate review, which is why I'm creating this issue.

`ParallelizableAttribute` takes an optional argument of Type `ParallelScope`. The values of `ParallelScope` are a mix of things. Some apply to the item on which the attribute appears and others to descendants. In addition, `ParallelScope.None` is used both when specified by the user and as a default. This turns out to lose info for us, since there are three, rather than two, actions the dispatcher may take in running a test:
 * Run directly on the same thread as parent
 * Enqueue on a parallel queue
 * Enqueue on a non-parallel queue

I'm proposing a number of changes to make parallel test cases work and also to clarify things for users:

1. I have already defined `ParallelScope.Default` which indicates that no attribute was used. This turned out to be necessary to effectively implement parallel test cases.

2. For convenience internally I have re-defined the various scopes in two ranges, separating those flags that affect the item on which they appear from those that affect descendants. I use two masks to pick out the flags that are needed in a particular instance.

[Remaining items have to do with how this appears to users]

3. I'd like to deprecate `ParallelScope.None` in favor of `ParallelScope.NonParallel`. The former name is confusing enough that it was confusing me in working #164. Its implementation has always been "run test on non-parallel queue" but it sounds as if it means "not specified" or "use default". I'm open to other names. I considered "Never" as an option.

4. For consistency, I'd like to define `ParallelScope.Parallel` as a synonym for `ParallelScope.Self`. I wouldn't deprecate the latter, however, because combos like `ParallelScope.Self+ParallelScope.Children` read quite well.

5. To make it even easier for users. It may be useful to define `NonParallelizableAttribute` with no arguments. Just as `[Parallelizable]` defaults to `[Parallelizable(ParallelScope.Self)`, this new attribute would mean the same as `[Parallelizable(ParallelScope.NonParallel)`

6. Some values of `ParallelScope` make no sense in certain places and are currently ignored silently. For example, `Parallelizable(ParallelScope.Fixtures)` has no effect if placed on a test case. Rather than silently ignoring the attribute, I propose that we mark the test as invalid so that the user knows it won't do whatever is expected and removes it. Implementation will require looking at each scope value and deciding whether to allow it on an Assembly, a class or a method. It seems to me that getting an early error is better for the user than having their intent (which we don't know of course) ignored silently.

7. I propose adding `ParallelScope.Cases` to the set of scopes that impact descendants. Used at the assembly level without combining with other scopes, it would cause test fixtures to run sequentially but all the cases within them to run in parallel. I don't think this will get a lot of use but it's consistent with the presense of `ParallelScope.Fixtures`. I would redefine `Children` as `Fixtures+Cases` if we do this.

Please comment! For implementation details, you'll be able to review the PR of course.

CharliePoole:
This started as initial work toward #164 - parallel test cases - but ended up also incorporating changes in how users express parallelism of tests. It now fixes #2003.

The changes toward parallel test cases are primarily around recognizing that we have three - rather than two - ways of running tests: from the parallel queue, from the non-parallel queue and directly on the thread of the parent. Creating `ParallelScope.Default` allows recognizing the situation where no attribute has been specified, which leads to the third way of running.

In the current commit, test cases are still not run in parallel. If the `#define` that keeps them from running in parallel is removed, other things have to be fixed. So that will be for a further PR. However, I think that the new code structure in place for __deciding__ how to run the tests is the way it will need to be in the future.

As far as the user-facing stuff goes, I added `NonParallelizableAttribute` as an equivalent for `Parallelizable(ParallelScope.None)`. Rather than deprecating anything, I hid it from Intellisense so that old code will still compile but users will not be prompted to use it in new code. Some other new fields, used internally only, are also hidden in this way.

This approach is one of several I could have used. It's the one that requires the least change by users but complicates our own code a bit more to accomodate. Two alternatives to consider, which could be implemented pretty easily:

1. Define some new internal enumeration to replace ParallelScope and keep ParallelScope pretty much as is for users. In that case, I would still need to either deprecate or hide ParallelScope.None.

2. Make the constructor that uses ParallelScope obsolete and create some new, simpler enumeration for public use. Use ParallelScope internally only. This would simplify our code a bit but would be the biggest change for users.

I would like to merge this with or without changes before moving on to do more work on parallelizing test cases.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/NonParallelizableAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/NonParallelizableAttribute.cs
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2014 Charlie Poole
+// Copyright (c) 2017 Charlie Poole
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
@@ -32,24 +32,14 @@ namespace NUnit.Framework
     /// ParallelizableAttribute is used to mark tests that may be run in parallel.
     /// </summary>
     [AttributeUsage( AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method, AllowMultiple=false, Inherited=false )]
-    public sealed class ParallelizableAttribute : PropertyAttribute, IApplyToContext
+    public sealed class NonParallelizableAttribute : PropertyAttribute
     {
-        private ParallelScope _scope;
-
         /// <summary>
-        /// Construct a ParallelizableAttribute using default ParallelScope.Self.
+        /// Construct a NonParallelizableAttribute.
         /// </summary>
-        public ParallelizableAttribute() : this(ParallelScope.Self) { }
-
-        /// <summary>
-        /// Construct a ParallelizableAttribute with a specified scope.
-        /// </summary>
-        /// <param name="scope">The ParallelScope associated with this attribute.</param>
-        public ParallelizableAttribute(ParallelScope scope) : base()
+        public NonParallelizableAttribute()
         {
-            _scope = scope;
-
-            Properties.Add(PropertyNames.ParallelScope, scope);
+            Properties.Add(PropertyNames.ParallelScope, ParallelScope.None);
         }
 
         /// <summary>
@@ -60,14 +50,8 @@ namespace NUnit.Framework
         {
             base.ApplyToTest(test);
 
-            if (test.RunState == RunState.NotRunnable)
-                return;
-
-            if (_scope.HasFlag(ParallelScope.Self) && _scope.HasFlag(ParallelScope.None))
-                MarkTestInvalid(test, "Test may not be both Parallel and NonParallel");
-
-            if (test is TestMethod && _scope.HasFlag(ParallelScope.ContextMask))
-                MarkTestInvalid(test, "ParallelScope of a test method may not specify Children or Fixtures");
+            if (test.RunState != RunState.NotRunnable && test is TestAssembly)
+                MarkTestInvalid(test, "Assemblies may not be marked as NonParallel");
         }
 
         private void MarkTestInvalid(Test test, string message)
@@ -75,20 +59,5 @@ namespace NUnit.Framework
             test.RunState = RunState.NotRunnable;
             test.Properties.Add(PropertyNames.SkipReason, message);
         }
-
-        #region IApplyToContext Interface
-
-        /// <summary>
-        /// Modify the context to be used for child tests
-        /// </summary>
-        /// <param name="context">The current TestExecutionContext</param>
-        public void ApplyToContext(TestExecutionContext context)
-        {
-            // Don't reflect Self in the context, since it will be
-            // used for descendant tests.
-            context.ParallelScope = _scope & ParallelScope.ContextMask;
-        }
-
-        #endregion
     }
 }
--- a/src/NUnitFramework/framework/Attributes/NonParallelizableAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/NonParallelizableAttribute.cs
@@ -41,23 +41,5 @@ namespace NUnit.Framework
         {
             Properties.Add(PropertyNames.ParallelScope, ParallelScope.None);
         }
-
-        /// <summary>
-        /// Overridden to check for invalid combinations of settings
-        /// </summary>
-        /// <param name="test"></param>
-        public override void ApplyToTest(Test test)
-        {
-            base.ApplyToTest(test);
-
-            if (test.RunState != RunState.NotRunnable && test is TestAssembly)
-                MarkTestInvalid(test, "Assemblies may not be marked as NonParallel");
-        }
-
-        private void MarkTestInvalid(Test test, string message)
-        {
-            test.RunState = RunState.NotRunnable;
-            test.Properties.Add(PropertyNames.SkipReason, message);
-        }
     }
 }
--- a/src/NUnitFramework/framework/Attributes/ParallelizableAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/ParallelizableAttribute.cs
@@ -64,16 +64,10 @@ namespace NUnit.Framework
                 return;
 
             if (_scope.HasFlag(ParallelScope.Self) && _scope.HasFlag(ParallelScope.None))
-                MarkTestInvalid(test, "Test may not be both Parallel and NonParallel");
+                test.MakeInvalid("Test may not be both parallel and non-parallel");
 
             if (test is TestMethod && _scope.HasFlag(ParallelScope.ContextMask))
-                MarkTestInvalid(test, "ParallelScope of a test method may not specify Children or Fixtures");
-        }
-
-        private void MarkTestInvalid(Test test, string message)
-        {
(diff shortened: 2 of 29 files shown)


CharliePoole:
@nunit/core-team @nunit/framework-team Can somebody please review this?

src/NUnitFramework/framework/Attributes/NonParallelizableAttribute.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// <summary>
+        /// Overridden to check for invalid combinations of settings
+        /// </summary>
+        /// <param name="test"></param>
+        public override void ApplyToTest(Test test)
+        {
+            base.ApplyToTest(test);
+
+            if (test.RunState != RunState.NotRunnable && test is TestAssembly)
+                MarkTestInvalid(test, "Assemblies may not be marked as NonParallel");
```
CharliePoole: Also, the question arises, what should we do if no attribute is present on the assembly? Is it better to start off on the non-parallel queue or the parallel queue? I guess it depends on what we are going to find at lower levels, which we can't know at the point of kicking things off.

src/NUnitFramework/framework/Attributes/ParallelizableAttribute.cs
```diff
@@ -51,6 +52,30 @@ public ParallelizableAttribute(ParallelScope scope) : base()
...
+        /// <param name="test"></param>
+        public override void ApplyToTest(Test test)
+        {
+            base.ApplyToTest(test);
+
+            if (test.RunState == RunState.NotRunnable)
+                return;
+
+            if (_scope.HasFlag(ParallelScope.Self) && _scope.HasFlag(ParallelScope.None))
+                MarkTestInvalid(test, "Test may not be both Parallel and NonParallel");
```
CharliePoole: Good catch. They were in an earlier version of the enum. I think I prefer "Test may not be both parallel and non-parallel".

src/NUnitFramework/framework/Attributes/ParallelizableAttribute.cs
```diff
@@ -51,6 +52,30 @@ public ParallelizableAttribute(ParallelScope scope) : base()
...
+
+            if (test is TestMethod && _scope.HasFlag(ParallelScope.ContextMask))
+                MarkTestInvalid(test, "ParallelScope of a test method may not specify Children or Fixtures");
+        }
+
+        private void MarkTestInvalid(Test test, string message)
```
CharliePoole: It could easily be a local extension method, used by this attribute only. Do you think that would be clear, or would it necessitate going to find the definition in order to understand it?

It couldn't be a general extension method within NUnit without violating slightly the independence of attributes. I've tried to keep most attribute definitions such that they could have been implemented as extensions because they then serve as a model for extensions and also because it forces us to exercise the extensibility mechanism all the time. The violation would only be slight, however, and the extension methods might be helpfull for those extending NUnit.

CharliePoole:
@jnm2 For some reason, my responses to your inline comments are showing up as "pending" review items, even though I'm not actually reviewing my own code. Let me know if you can't see them.

jnm2:
@CharliePoole I cannot see them and my guess is that Pending means it's waiting for you to click a publish button. Maybe you clicked Add review comment instead of Add single comment?

CharliePoole:
@jnm2 No, I just clicked reply on your comments. Then after typing an answer, my only options are cancel and add review comment. Very annoying! I'll finish my "review".

jnm2:
@CharliePoole 

> No, I just clicked reply on your comments. Then after typing an answer, my only options are cancel and add review comment. Very annoying! I'll finish my "review".

Let me know if there's something better I can be doing when I initially leave comments then.

src/NUnitFramework/framework/Attributes/ParallelizableAttribute.cs
jnm2: I like that better too. Hopefully people will be able to quickly see what they did wrong (in terms of ParallelScope.Self and ParallelScope.None).

src/NUnitFramework/framework/Attributes/ParallelizableAttribute.cs
jnm2: It's already used by two attributes at least and could be used by more. I'd prefer a `Test.MakeNotRunnable(string reason)` instance method. I'm not quite following where that violates the independence of attributes. Anyone who can modify RunState and Properties could call this method and at least it'd encourage consistency.

src/NUnitFramework/framework/Attributes/NonParallelizableAttribute.cs
jnm2: > So, my inclination is to go ahead and allow the attribute at the assembly level, removing this check.

Sounds less arbitrary to me, I like it.

> Also, the question arises, what should we do if no attribute is present on the assembly? Is it better to start off on the non-parallel queue or the parallel queue

I think everything should go in the parallel queue until it's discovered that it's in the wrong place. I'd prefer us to optimize for tests to be parallelized. It'll be rare that they can't be- particularly at the fixture level and above.

CharliePoole:
@jnm2 I don't think it's anything you're doing... just some GitHub weirdness.

src/NUnitFramework/framework/Attributes/ParallelizableAttribute.cs
CharliePoole: Makes sense...

CharliePoole: Rather than an extension method, I took the more direct approach of adding a MakeInvalid method to Test. It's only used in two places in this PR but could be used in over a dozen places elsewhere. I don't want to mess up this PR with that unrelated change, however, so it can be done later.

CharliePoole:
No reviews from any @nunit/core-team members, which I would have liked because of the change to the user interface. However, I'll go ahead and merge once this passes and if there are further comments I'll deal with them before we get to the next release.

rprouse:
I only had time to do a quick review, but I think that the changes to the interface make this more intuitive to use. The code I've reviewed looks good.

CharliePoole:
Cool! The interface is what's important the deeper internals are transitional and changeable.

src/NUnitFramework/framework/Internal/Tests/Test.cs
```diff
@@ -349,6 +349,16 @@ public void ApplyAttributesToTest(ICustomAttributeProvider provider)
...
 #endif
 
+        /// <summary>
+        /// Mark the test as Invalid (not runnable) specifying a reason
+        /// </summary>
+        /// <param name="reason">The reason the test is not runnable</param>
+        public void MakeInvalid(string reason)
+        {
+            RunState = RunState.NotRunnable;
+            Properties.Add(PropertyNames.SkipReason, reason);
```
jnm2: I like this. Just one thing, maybe we should skip the`Properties.Add` if reason is null or empty, or else throw if it should be required?

CharliePoole: Good idea. I'll do that in the following PR.

PR #2018, opened 2017-01-31
Fix tests failing on systems with comma decimal mark settings

### LINKED ISSUE #2017: Two NUnit project's tests fail on systems with comma decimal mark settings
The following tests fail on systems with comma decimal mark settings:
- NUnit.Framework.Internal.Results.TestResultGeneralTests.SuiteResultXml_Duration
- NUnit.Framework.Internal.Results.TestResultGeneralTests.TestResultXml_Duration

Steps to reproduce:
1. Change system region, or language settings to one with comma  decimal mark ([Countries using Arabic numerals with decimal comma](https://en.wikipedia.org/wiki/Decimal_mark#Countries_using_Arabic_numerals_with_decimal_comma))
2. Run the build script with `-Target Test` option
3. See aforementioned tests fail for every framework

GCichosz:
Two tests fail because Assert.AreEqual expected double is transformed to a string with "0.000000" format. Such method call on systems with comma decimal mark returns a string with comma as decimal mark. Setting ToString provider to InvariantCulture ensures a string with point as decimal mark is returned.

Fixes #2017

### DIFF
--- a/src/NUnitFramework/tests/Internal/Results/TestResultGeneralTests.cs
+++ b/src/NUnitFramework/tests/Internal/Results/TestResultGeneralTests.cs
@@ -22,6 +22,7 @@
 // ***********************************************************************
 
 using System;
+using System.Globalization;
 using System.Linq;
 using NUnit.Framework.Interfaces;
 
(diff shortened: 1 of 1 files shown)


rprouse:
Looks good to me too. Thanks for your help. 👍

Issue #1996, opened 2017-01-20
Timeout does not work if native code is running at the time


jnm2:
If you have native code running, Timeout has no effect. I assume this is because it relies on Thread.Abort.

Is there anything that could be done to improve this? A native way to nuke it if Thread.Abort does nothing?

ChrisMaddock:
https://github.com/nunit/nunit/issues/1363 is probably relevant here. Just springs to mind as I came across it only a couple of days ago with regards to making Timeout available in the netstandard build, where System.Threading isn't available. I was previously considering a separate implementation using Tasks (which wouldn't be available in <net-4ish) - but sounds like a non-thread-based implementation of TimeoutAttribute could solve all three problems.

@CharliePoole - what thoughts did you have on the above issue, as its assigned to you?

CharliePoole:
The intent of #1363 is to cancel the entire worker thread when there is a timeout, thereby eliminating the need for a "side thread" to run those tests with timeout. To do that, we need a way to add and remove workers smoothly while the tests are running. I think that's a general improvement to the dispatcher that we could use and I'd like to work on it.

Ideally, it would be great if we could deal with timeout in a different way. In that case, I'd re-title #1363 and just do the enabling bit, not the timeout.

Even if your implementation using tasks were only available in .NET 4+ and NetStandard, I think it would be a great improvement, so I say go for it!

jnm2:
Timeouts also have no effect if the execution is inside a call to `Application.Run()`. My original question was kind of specialized because I was calling MsgWaitForMultipleObjectsEx, but this is much more general of a problem.

CharliePoole:
Are we just doing a simple Thread.Abort? We used to have code that did a bit more and took care of some of these cases.

jnm2:
I would have expected Thread.Abort to work against the WinForms message pump, since I would have expected all waits and joins in the entire framework to be hardened for thread abort.

Apparently it just calls [either MsgWaitForMultipleObjectsEx](https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/Application.cs,2182) or [WaitMessage](https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/Application.cs,2195) without waiting on any kind of 'thread abort' handle.

The solution seems to be kind of easy perhaps: just send a thread message and the native to managed transition will abort the thread.

jnm2:
[...]

PR #2023, opened 2017-02-01
Ability to abort threads running a message pump

jnm2:
Closes everything in #1996 that is actionable at this time.
For reference to make it easy to see what is involved in aborting threads when testing GUI.

### DIFF
--- a/src/NUnitFramework/framework/Internal/ThreadUtility.cs
+++ b/src/NUnitFramework/framework/Internal/ThreadUtility.cs
@@ -173,13 +173,24 @@ namespace NUnit.Framework.Internal
 
 
 
+        private static bool isNotOnWindows;
+
         /// <summary>
         /// Captures the current thread's native id. If provided to <see cref="Kill(Thread,int)"/> later, allows the thread to be killed if it's in a message pump native blocking wait.
         /// </summary>
         [SecuritySafeCritical]
         public static int GetCurrentThreadNativeId()
         {
-            return GetCurrentThreadId();
+            if (isNotOnWindows) return 0;
+            try
+            {
+                return GetCurrentThreadId();
+            }
+            catch (EntryPointNotFoundException)
+            {
+                isNotOnWindows = true;
+                return 0;
+            }
         }
 
 
--- a/src/NUnitFramework/framework/Internal/ThreadUtility.cs
+++ b/src/NUnitFramework/framework/Internal/ThreadUtility.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
@@ -23,6 +23,9 @@
 
 #if !PORTABLE && !NETSTANDARD1_6
 using System;
+using System.ComponentModel;
+using System.Runtime.InteropServices;
+using System.Security;
 using System.Threading;
 
 namespace NUnit.Framework.Internal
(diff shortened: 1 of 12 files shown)


jnm2:
Darn, I forgot to write the most important test exercising the WorkItem's new capability. Will do tomorrow.

jnm2:
The master branch build fails of my machine, so in order to let the build succeed for this PR I'll try to find the most recent point in master where it succeeded and rebase there.

It looks like a typo, `Internal.` shouldn't be there: https://github.com/nunit/nunit/blob/fc4f1b4310f9f99b6a77c26c7aa3ec3afbd4e6e8/src/NUnitFramework/tests/TextOutputTests.cs#L152

jnm2:
Did another clean checkout of master, this time didn't open VS. Now it's not erroring on build45 but rather buildnetstandard. Some reference is not getting restored.

jnm2:
The `build45` and `buildnetstandard` errors seem to be related to NuGet packages issues? With a clean clone of master, just running `build.ps1` doesn't work. You have to open VS and build, and that restores NuGet packages for a while.

CharliePoole:
I didn't think we had decided what to do about #1996 yet.

I'll kick off a build of master to check that it's OK - my local builds are all working.

I haven't reviewed this in detail, but I notice you are using a Windows dll, which isn't possible for Linux.

CharliePoole:
I see that the merge I did 18 minutes ago built, so master is OK.

jnm2:
> I didn't think we had decided what to do about #1996 yet.

This PR served two purposes: mainly because #1996 could be hard to grasp without seeing it and the implementation was instructive, and also on the off-chance that it'd be implemented. This does not represent a decision or an expectation of commitment. :-D

> I haven't reviewed this in detail, but I notice you are using a Windows dll, which isn't possible for Linux.

This is the kind of thing I was hoping we could discuss. `ThreadUtility` is ifdef'd for .NET Framework only, but since you support Mono I'd have to implement a one-time DllNotFoundException check to figure out which platform we're on.

> I see that the merge I did 18 minutes ago built, so master is OK.

Fresh clone gave the same error locally as the PR did on [Travis](https://travis-ci.org/nunit/nunit/builds/197395659) and [Appveyor](https://ci.appveyor.com/project/CharliePoole/nunit/build/3.7.0-ci-03641-pr-2023), so there's some contradiction. I'll try a fifth time, with the latest master.

jnm2:
1. `git clone https://github.com/nunit/nunit.git .` in new folder
2. `build.ps1 -t test`, all builds succeed except netstandard build fails with over a thousand "type not found" errors
3. Open `nunit.sln` in VS2017, builld project `nunit.framework.tests-netstandard`, success
4. `build.ps1 -t test`, all builds succeed except the .NET Framework ones which fail with the same error as [Travis](https://travis-ci.org/nunit/nunit/builds/197395659) and [Appveyor](https://ci.appveyor.com/project/CharliePoole/nunit/build/3.7.0-ci-03641-pr-2023)

jnm2:
Okay, it's gotta be my PR that caused the error, but that doesn't explain why the error persists when I delete the folder contents and reclone master.

jnm2:
@CharliePoole Oh, it's me. I introduced a new namespace when I created file. Ouch. That's embarrassing because I was surer than I should have been of my delete-reclone-rebuild evidence. :-/

jnm2:
@CharliePoole Inside `\src\NUnitFramework\tests\Internal`, namespaces used are:

 * `NUnit.Framework.Internal` (mostly)
 * `NUnit.Framework.Internal.Tests` (few)
 * `NUnit.Framework.Internal.Execution` (few)
 * `NUnit.Framework.Tests` (one)

And Visual Studio, going by the path, thinks it should be `NUnit.Framework.Tests.Internal` and uses that in new files. ReSharper complains if it's different.

I'll go with majority choice here.

CharliePoole:
Our test namespaces are definitely inconsistent. Can be confusing.

Going back to an earlier comment about ThreadUtility being #ifdefed for .NET only: that's not true. We don't distinguish between .NET desktop and mono. There's only one build for both platforms.... So basically the net-4.5 build is for both MS and Mono. It would be a major major change to start producing different builds for mono and one we wouldn't want to make. BTW, the netstandard builds are expected to work on mono eventually as well.

jnm2:
@CharliePoole Good to know the details. Since this is a proof-of-concept I'd like to try running against Mono. I expect from experience on versions of Windows that it's easy to handle DllNotFoundException so that it happens at most once and doesn't slow anything down.

jnm2:
I can hardly wait for when we can use the new Roslyn Project system and no longer list .cs files inside .csprojs and multitarget to build all platforms from a single .csproj. Those will be the days! =)

jnm2:
This is ready to go. Tested on Linux and also against real-world Windows Forms unit tests.

src/NUnitFramework/framework/Internal/ThreadUtility.cs
```diff
@@ -33,22 +36,82 @@ namespace NUnit.Framework.Internal
...
+            private readonly Timer _timer;
+            private readonly WaitCallback _threadPoolWork;
+            private readonly object _state;
+
+            public DelayClosure(int milliseconds, WaitCallback threadPoolWork, object state)
+            {
+                _threadPoolWork = threadPoolWork;
+                _state = state;
+                _timer = new Timer(Callback, null, Timeout.Infinite, Timeout.Infinite);
+                // Make sure _timer is set before starting the timer so that it is disposed.
```
rprouse: So the timer won't get disposed if you start it immediately? I didn't know that. Any references I can read? This won't block merging, just curious.

src/NUnitFramework/framework/Internal/ThreadUtility.cs
```diff
@@ -68,6 +131,96 @@ public static void Kill(Thread thread, object stateInfo)
...
+
+            public CheckOnAbortingThreadState(Thread thread, int nativeId)
+            {
+                Thread = thread;
+                NativeId = nativeId;
+            }
+        }
+
+
+
```
rprouse: For future reference, excess blank lines triggers my OCD like trailing whitespace does for you 😄

jnm2: Hehe! We'll have to decide who is triggered more, cause I get claustrophobic if I can't group with whitespace (kidding).
Two lines okay? Better way to organize the file in general, regions?

src/NUnitFramework/framework/Internal/ThreadUtility.cs
jnm2: @rprouse Only in this case. As I was writing the code I visualized this sequence of events:

 1. The expression `new Timer(Callback, null, milliseconds, Timeout.Infinite)` is evaluated
 2. `milliseconds` is so small that the callback is called before step 3. You get a null reference exception in `Callback` because the `_timer field is not yet set.
 3. The `_timer` field is set to the value returned from evaluating the` new Timer` expression.

jnm2: I should leave a note in the source mentioning the possible race condition if the field is not set before allowing the timer to call back.

src/NUnitFramework/framework/Internal/ThreadUtility.cs
rprouse: I just dislike more than one blank line. It just means that I have to scroll more to grok the code. No biggy 😄

src/NUnitFramework/framework/Internal/ThreadUtility.cs
rprouse: @jnm2 I think the comment you have is good enough to indicate that you did it for a reason. The assignment call being a separate op didn't even click for me, good work 👍

rprouse: Based on your comment, I was thinking, that the constructor might not return before the timer fired if you used a very short timer interval, but I wouldn't have thought about it without your comment.

jnm2:
I was supposed to document what I did here so that the others feel more confident approving this. I prioritized `Fixing SocketException caused by --explore` and `ThrowsAsync reports OperationCanceledException as TaskCanceledException` instead, but I can still get to this.

jnm2:
Might be tomorrow.

jnm2:
> **jnm2** was assigned by **rprouse** an hour ago

@rprouse I've been wondering whether to assign myself to my PRs automatically.

rprouse:
@jnm2 normally we don't assign people to PRs, but this one is listed as part of the release milestone so it helped me to see who is working on what and ensure that everything is covered while looking at the milestone as we lead up to the release. I did it only as a convenience for my release process.

PR #2024, opened 2017-02-01
Remove unused --verbose and --full command line options

### LINKED ISSUE #1997: Remove unused --verbose and --full command line options
@thomaslevesque commented on [Fri Oct 14 2016](https://github.com/nunit/nunit-console/issues/99)

What is the `--verbose` flag supposed to do? I can't see any difference between running nunit3-console.exe with or without this flag.

(btw, is there a way to print to the console which test is currently executing? that's what I was trying to do with `--verbose`)

---

@thomaslevesque commented on [Fri Oct 14 2016](https://github.com/nunit/nunit-console/issues/99#issuecomment-253811189)

I just looked at the code: apparently the `CommandLineOptions.Verbose` property is never used (except to define it in `ConfigureOptions`). The same is true of the `Full` property.

---

@ChrisMaddock commented on [Fri Oct 14 2016](https://github.com/nunit/nunit-console/issues/99#issuecomment-253821915)

I think you're looking for `--labels`? I'll leave someone else to comment on`--verbose`

---

@thomaslevesque commented on [Fri Oct 14 2016](https://github.com/nunit/nunit-console/issues/99#issuecomment-253828074)

Thanks @ChrisMaddock, this is what I was looking for!

---

@CharliePoole commented on [Fri Oct 14 2016](https://github.com/nunit/nunit-console/issues/99#issuecomment-253849981)

Once upon a time, `--verbose` did something. When I removed that code, I thought I would be coming back to add some different verbose output. We'll either kill the option or use it.

The `--full` option was an experimental report of all tests run. We'll either remove or use this one too.

---

@ChrisMaddock commented on [Fri Jan 20 2017](https://github.com/nunit/nunit-console/issues/99#issuecomment-274046250)

This is potentially easyfix once a decision is made. 

Can I suggest we remove both unused options, with the view that they could always be re-added, if something is implemented at a later date? @nunit/engine-team - thoughts?

---

@rprouse commented on [Fri Jan 20 2017](https://github.com/nunit/nunit-console/issues/99#issuecomment-274058775)

I say remove for now

---

@CharliePoole commented on [Fri Jan 20 2017](https://github.com/nunit/nunit-console/issues/99#issuecomment-274114222)

I agree

GCichosz:
Fixes #1997

### DIFF
--- a/src/NUnitFramework/nunitlite.tests/CommandLineTests.cs
+++ b/src/NUnitFramework/nunitlite.tests/CommandLineTests.cs
@@ -97,7 +97,6 @@ namespace NUnitLite.Tests
         [TestCase("WaitBeforeExit", "wait")]
 #if !PORTABLE
         [TestCase("NoHeader", "noheader|noh")]
-        [TestCase("Full", "full")]
 #endif
         [TestCase("TeamCity", "teamcity")]
         public void CanRecognizeBooleanOptions(string propertyName, string pattern)
--- a/src/NUnitFramework/nunitlite/CommandLineOptions.cs
+++ b/src/NUnitFramework/nunitlite/CommandLineOptions.cs
@@ -232,9 +232,6 @@ namespace NUnit.Common
         public string InternalTraceLevel { get; private set; }
         public bool InternalTraceLevelSpecified { get { return InternalTraceLevel != null; } }
 
-        /// <summary>Indicates whether a full report should be displayed.</summary>
-        public bool Full { get; private set; }
-
 #if !PORTABLE
         private List<OutputSpecification> resultOutputSpecifications = new List<OutputSpecification>();
         public IList<OutputSpecification> ResultOutputSpecifications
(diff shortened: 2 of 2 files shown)


CharliePoole:
Looks good to me. Needs to pass CI and have one more committer approval.

JustinRChou:
Should this also be removed from the console runner?

ChrisMaddock:
Looks good to me. CI was cancelled for some reason? I've restarted - I don't have the ability to merge without - but I think this is good to go! 😄 

This removes the options from NUnitLite - there's a separate issue for the console here: https://github.com/nunit/nunit-console/issues/99 I think this also needs removing from the docs. @GCichosz - if you're interested, would be great to get the other issue and docs tidied up too! 😁

JustinRChou:
I was always wondering if the verbose command listed in 
https://github.com/nunit/docs/wiki/Console-Command-Line
was the one that is going to be removed. Or it is a different one?

ChrisMaddock:
@JustinRChou - that's the one! Although (since NUnit3) it was never plugged in, so it's been a no-op. 

https://github.com/nunit/nunit-console/issues/99 was the original issue, if you want to see the back-story. 😄

PR #2028, opened 2017-02-03
Change Assert.Positive()'s description to positive

tobarja:
This looks like a copy/paste error. The descriptions all currently read `Asserts that an <type> is negative.`


CharliePoole:
Looks good. I'm merging right away since it's comments only.

PR #2032, opened 2017-02-04
Set all projects to C# language level 6

### LINKED ISSUE #2031: Limit Language level to C#6
The framework and console currently use a maximum language level of C#6 - we want to maintain that to ensure the projects can still be built by the C#6 generation of IDEs.

To enforce this, it would be great if we could add the `<LangVersion>` attribute to all the csproj's. This would prevent the projects compiling with C#7 features, help new contributors discover the required language level, and ensure no C#7 code slips through code review.

This can be set in VS by going to:
    
    Project Properties => Build tab => Advanced button => Language Version

Making that change will add a new property to the csproj file - it may be easier to edit the rest of the csproj's by hand, once the required XML has been found!

jnm2:
Closes #2031

There is a [silverlight project](https://github.com/nunit/nunit/compare/language-level-6?expand=1#diff-66d131603aed424ec7487dbdab8c8b8d) in here, is that supposed to be there?

### DIFF
--- a/src/NUnitFramework/framework/nunit.framework-2.0.csproj
+++ b/src/NUnitFramework/framework/nunit.framework-2.0.csproj
@@ -17,6 +17,7 @@
     <OldToolsVersion>3.5</OldToolsVersion>
     <UpgradeBackupLocation />
     <IntermediateOutputPath>obj\$(Configuration)\net-2.0\</IntermediateOutputPath>
+    <LangVersion>6</LangVersion>
   </PropertyGroup>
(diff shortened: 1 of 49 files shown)


CharliePoole:
I did a PR for the silverlight project. No need for an issue, I think.

ChrisMaddock:
> I did a PR for the silverlight project. No need for an issue, I think.

Did you forget to push this one or something @CharliePoole? I can't see anything. 😄

CharliePoole:
It was waiting for me to click the button to create the PR.

PR #2034, opened 2017-02-05
Remove silverlight project - no longer used

CharliePoole:


ChrisMaddock:
We'll be finally be rid of Silverlight one day. 😆

PR #2037, opened 2017-02-09
Make Theories fail if no data at all is provided

### LINKED ISSUE #1962: A Theory with no data passes
As a result of changes made for issue #1933, an internal problem was made visible. A Theory, for which no data is supplied, will pass.

If a Theory has data but all of it is unsuitable (assumptions fail), then the Theory itself fails. It's somewhat inconsistent for a Theory without data to pass - it should fail as well.

The source of the problem is in how NUnit runs test suites with no test cases. In NUnit V2, such suites were not handled in a special way. Any one-time setup (TestFixtureSetUp) and teardown (TestFixtureTearDown) was run, whether there were tests or not. In NUnit 3, we decided to change this behavior and only run one-time setup and teardown if there are actually child tests to be run. This saves wasted cycles when the one-time setup performs a lot of work, which is more common these days with folks doing database and web testing using NUnit.

Unfortunately, we did not change how Theories examine the result of their individual cases and decide whether they should pass or faill. This was (and is still) done by a special command added to the one-time teardown for the parameterized test suite representing the theory. Since one-time teardown is not run when there are no tests (i.e. no data) the command is never run.

To fix this requires re-thinking how we want to handle the overall result of a suite. In the general case, this is done each time a test case result is added to the suite result. In the case of Theories, a special action is taken to override the default behavior of suites. We will probably want to handle this by creating a special result for Theories with it's own aggregation behavior.

Since this is part of the fundamental architecture of the framework, I'm assigning it to myself for now.

CharliePoole:
Fixes #1962 

With this fix, a Theory that gets no data shows the same failed result as a theory that gets data that fails the assumptions, as we would expect.

At the same time, I fixed an issue where test methods without any argument were being silently ignored if they had a TestCaseSourceAttribute. They are now marked as invalid.

This fixes the immediate issue but leaves even more knowledge about kinds of tests in the work item, where it really doesn't belong. Some sort of refactoring is needed eventually.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
@@ -100,7 +100,7 @@ namespace NUnit.Framework.Internal.Execution
                         case RunState.Runnable:
                         case RunState.Explicit:
                             // Assume success, since the result will otherwise
-                            // edault to inconclusive.
+                            // default to inconclusive.
                             Result.SetResult(ResultState.Success);
 
                             CreateChildWorkItems();
--- a/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
@@ -99,9 +99,8 @@ namespace NUnit.Framework.Internal.Execution
                         default:
                         case RunState.Runnable:
                         case RunState.Explicit:
-                            // Assume success, since the result will be inconclusive
-                            // if there is no setup method to run or if the
-                            // context initialization fails.
+                            // Assume success, since the result will otherwise
+                            // edault to inconclusive.
                             Result.SetResult(ResultState.Success);
 
                             CreateChildWorkItems();
@@ -135,6 +134,8 @@ namespace NUnit.Framework.Internal.Execution
                                 if (Context.ExecutionStatus != TestExecutionStatus.AbortRequested)
                                     PerformOneTimeTearDown();
                             }
+                            else if (Test.TestType == "Theory")
+                                Result.SetResult(ResultState.Failure, "No test cases were provided");
                             break;
 
                         case RunState.Skipped:
@@ -153,7 +154,6 @@ namespace NUnit.Framework.Internal.Execution
             // Fall through in case nothing was run.
             // Otherwise, this is done in the completion event.
             WorkItemComplete();
-
         }
 
         #region Helper Methods
@@ -223,6 +223,9 @@ namespace NUnit.Framework.Internal.Execution
 
         private void RunChildren()
         {
+            if (Test.TestType == "Theory")
+                Result.SetResult(ResultState.Inconclusive);
+
             int childCount = _children.Count;
             if (childCount == 0)
                 throw new InvalidOperationException("RunChildren called but item has no children");
@@ -384,6 +387,9 @@ namespace NUnit.Framework.Internal.Execution
             _childTestCountdown.Signal();
             if (_childTestCountdown.CurrentCount == 0)
             {
+                if (Test.TestType == "Theory" && Result.ResultState == ResultState.Success && Result.PassCount == 0)
+                    Result.SetResult(ResultState.Failure, "No test cases were provided");
+
                 if (Context.ExecutionStatus != TestExecutionStatus.AbortRequested)
                     PerformOneTimeTearDown();
 
--- a/src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
@@ -220,74 +220,60 @@ namespace NUnit.Framework.Internal
         public virtual void AddResult(ITestResult result)
         {
 #if PARALLEL
-            var childrenAsConcurrentQueue = Children as ConcurrentQueue<ITestResult>;
-            if (childrenAsConcurrentQueue != null)
-                childrenAsConcurrentQueue.Enqueue(result);
-            else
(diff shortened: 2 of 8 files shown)


CharliePoole:
@nunit/framework-team If you are part of the team, please monitor and review PRs. This one has been ready for review and merge for a while.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -99,9 +99,8 @@ protected override void PerformWork()
                         default:
                         case RunState.Runnable:
                         case RunState.Explicit:
-                            // Assume success, since the result will be inconclusive
-                            // if there is no setup method to run or if the
-                            // context initialization fails.
+                            // Assume success, since the result will otherwise
+                            // edault to inconclusive.
```
jnm2: Nit: typo

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -384,6 +387,9 @@ private void CountDownChildTest()
             _childTestCountdown.Signal();
             if (_childTestCountdown.CurrentCount == 0)
             {
+                if (Test.TestType == "Theory" && Result.ResultState == ResultState.Success && Result.PassCount == 0)
```
jnm2: I wish we could stop doing the `TestType == "Theory"` check (tell-don't-ask) sometime in the future. This is the third one being added.

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -294,17 +294,10 @@ public void HandlesExceptionInTestCaseSource_GuiDisplay(string lhs, string rhs)
         private static IEnumerable<TestCaseData> ZeroTestCasesSource() => Enumerable.Empty<TestCaseData>();
 
         [TestCaseSource("ZeroTestCasesSource")]
-        public void TestWithZeroTestSourceCasesShouldNotBeRun()
-        {
-            Assert.Fail();
-        }
```
jnm2: I think there is value to both tests. I know this is redundant now, but the knowledge that it is redundant is a detail which is subject to change. Each test does something unique and perpendicular: one ensures the test is not executed, the other ensures that NUnit doesn't fail the test prior to running it due to missing a parameter.

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
```diff
@@ -220,74 +220,60 @@ public override IEnumerable<ITestResult> Children
         public virtual void AddResult(ITestResult result)
         {
 #if PARALLEL
-            var childrenAsConcurrentQueue = Children as ConcurrentQueue<ITestResult>;
-            if (childrenAsConcurrentQueue != null)
-                childrenAsConcurrentQueue.Enqueue(result);
```
jnm2: I don't understand this refactoring is being done. Is this PR fixing more than one thing at a time?

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
CharliePoole: It bothered me writing it as well... but I think the best solution is to introduce a type of ITest that is a theory and I didn't want to mix that in with this issue.

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
CharliePoole: Actually, no... This test, if left in, would now cause the run to fail since we are treating TestCaseSourceAttribute on a method without args as an error. Since it's an error, it can only be tested at second level, so we have TestMethodBuilderTests.TestCaseSourceAttribute_NoArgs_NoData_NotRunnable.

We could stand a great deal of refactoring of these tests for clarity, I'll admit.

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
CharliePoole: I had to modify this method because the way results were merged was not correct in the case of Theories. I refactored mostly so as to make sure I understood it. But the above lines are just wrong: if PARALLEL is defined, then Children is already a ConcurrentQueue - no need to convert it. If it's not defined, then it's a List. Somebody got mixed up at some time in the past, which is easy to do with conditional compilation.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
CharliePoole: I wouldn't normally go back for something as small as this, but I notice a lot of commented out code as well, which should be removed, so I'll do another push.

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
jnm2: Ahh... sorry. It all makes sense now.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
rprouse: Agreed with the need to refactor, but also agree that it doesn't belong here.

src/NUnitFramework/framework/Internal/Results/TestSuiteResult.cs
rprouse: This is a lot more readable now. 👍

rprouse:
I did a squash merge on this so we can cherry pick the 1 commit easily into the hotfix release.

PR #2041, opened 2017-02-16
Handle null arguments in TestFixtureAttribute 

### LINKED ISSUE #1934: NullReferenceException when null arguments are used in TestFixtureAttribute
This seems to be related to #1507. We started getting several errors after moving from 3.4 to 3.5. The errors can be seen in the Visual Studio test runner and the nunit3-console runner. The errors are all similar to the following:
```
1) Invalid : Tests.Controls.iBus.When_the_passed_in_domain_is_null
   at NUnit.Framework.Internal.Builders.DefaultSuiteBuilder.HasArguments(IFixtureBuilder attr)
   at NUnit.Framework.Internal.Builders.DefaultSuiteBuilder.GetFixtureBuilderAttributes(ITypeInfo typeInfo)
```

This is one example of the test fixtures/constructors that are failing again:
```
[TestFixture(null)]
[TestFixture("")]
[TestFixture(" ")]
[TestFixture("\t")]
public class When_the_passed_in_domain_is_null : BasePresenterTest
{
    private readonly string _domain;

    public When_the_passed_in_domain_is_null(string domain)
    {
        _domain = domain;
    }
}
```

GCichosz:
The null reference checks I added solve the issue of `NullReferenceException`, and a fixture with null argument is marked as invalid with "No suitable constructor was found" message instead. I believe that is the intention, because of the way .NET handles params method parameter.

I suppose this is by design:
```
// This code doesn't result in an invalid test fixture
[TestFixture((object)null)]
public class SomeTestFixture
{
    public SomeTestFixture(object someObject)
    {
    }
}

// This code does result in an invalid test fixture
[TestFixture(null)]
public class SomeTestFixture
{
    public SomeTestFixture(object someObject)
    {
    }
}
```

Unfortunately the bug is in `DefaultSuiteBuilder`, and there are no unit tests for builders, so I don't know whether I should write tests for that bug, or not.

Fixes #1934

### DIFF
--- a/src/NUnitFramework/framework/Internal/Builders/DefaultSuiteBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/DefaultSuiteBuilder.cs
@@ -176,7 +176,7 @@ namespace NUnit.Framework.Internal.Builders
             // Only TestFixtureAttribute can be used without arguments
             var temp = attr as TestFixtureAttribute;
 
-            return temp == null || temp.Arguments.Length > 0 || temp.TypeArgs.Length > 0;
+            return temp == null || (temp.Arguments != null && temp.Arguments.Length > 0) || (temp.TypeArgs != null && temp.TypeArgs.Length > 0);
         }
 
         #endregion
--- a/src/NUnitFramework/framework/Internal/Builders/DefaultSuiteBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/DefaultSuiteBuilder.cs
@@ -176,7 +176,7 @@ namespace NUnit.Framework.Internal.Builders
             // Only TestFixtureAttribute can be used without arguments
             var temp = attr as TestFixtureAttribute;
 
-            return temp == null || (temp.Arguments != null && temp.Arguments.Length > 0) || (temp.TypeArgs != null && temp.TypeArgs.Length > 0);
+            return temp == null || temp.Arguments.Length > 0 || temp.TypeArgs.Length > 0;
         }
 
         #endregion
--- a/src/NUnitFramework/framework/Attributes/TestFixtureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestFixtureAttribute.cs
@@ -53,7 +53,7 @@ namespace NUnit.Framework
         public TestFixtureAttribute(params object[] arguments)
         {
             RunState = RunState.Runnable;
-            Arguments = arguments;
+            Arguments = arguments ?? new object[] { null };
             TypeArgs = new Type[0];
             Properties = new PropertyBag();
         }
--- a/src/NUnitFramework/tests/Attributes/TestFixtureAttributeTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TestFixtureAttributeTests.cs
@@ -80,5 +80,14 @@ namespace NUnit.Framework.Attributes
             Assert.That(attr.Arguments, Is.EqualTo(combinedArgs));
             Assert.That(attr.TypeArgs.Length, Is.EqualTo(0));
         }
+
+        [Test]
+        public void ConstructWithWeakTypedNullArgument()
+        {
+            TestFixtureAttribute attr = new TestFixtureAttribute(null);
+            Assert.That(attr.Arguments, Is.Not.Null);
+            Assert.That(attr.Arguments[0], Is.Null);
+            Assert.That(attr.TypeArgs, Is.Not.Null);
+        }
     }
 }


src/NUnitFramework/framework/Internal/Builders/DefaultSuiteBuilder.cs
```diff
@@ -176,7 +176,7 @@ private bool HasArguments(IFixtureBuilder attr)
             // Only TestFixtureAttribute can be used without arguments
             var temp = attr as TestFixtureAttribute;
 
-            return temp == null || temp.Arguments.Length > 0 || temp.TypeArgs.Length > 0;
+            return temp == null || (temp.Arguments != null && temp.Arguments.Length > 0) || (temp.TypeArgs != null && temp.TypeArgs.Length > 0);
         }
```
CharliePoole: While this fixes the immediate issue, it doesn't seem to get to the heart of the problem. It shouldn't be possible for the Arguments or TypeArgs properties of the attribute to be null. They should be set to an empty array. Can you figure out what changed to make them be null?

jnm2: Should the attribute be throwing if Arguments is set to null or uninitialized?

CharliePoole: @jnm2 It would have to be a postcondition in every constructor in that case. Maybe simpler to add a check to the tests of the attribute itself.

jnm2: I'm a huge fan of failing early. As a consumer of NUnit, not running NUnit's own tests, if I write `[TestFixture(null)]` and that isn't legal, I'd expect an error to be thrown in the constructor. 

Inserting a single `ArgumentNullException` precondition on `arguments` [here](https://github.com/nunit/nunit/blob/release/3.6/src/NUnitFramework/framework/Attributes/TestFixtureAttribute.cs#L55) would be all that's needed, right? Either that or `?? new object[0]`.

jnm2: Actually if I write `[TestFixture(null)]` I'm probably intending `null` to be the first argument.
Another overload might be good, so that I don't have to write `[TestFixture(new object[] { null })]` which would be weird next to non-null usages of the attribute.

CharliePoole: The intent of the code is that `TestFixture(null)` should be legal and equivalent to `TestFixture(new object[] { null })]`. I'm pretty sure that once worked, but perhaps it isn't protected by a test and something changed.

Probably, we need to review all the alternative constructors of `TestFixtureAttribute` and simplify.

jnm2: Gotcha, that would definitely be a job for unit tests like you said.

GCichosz:
I couldn't track what changed to allow `TestFixtureAttribute` Arguments to be null, but I noticed that a similar check is done in `TestCaseAttribute` constructor, so that solution seems reasonable.

GCichosz:
Is a regular contributor, like me, able to repeat a build on Travis? The one for this pull request failed in a test which seem to be completely unrelated, and it "works on my machine" ;-)

CharliePoole:
The Contributors Team is something we made up so as to be able to assign issues to people. GitHub has it set up so that you can only assign an issue to someone who is a part of your Organization, so we created Contributors with __read__ access to most of our repositories, write access to the wiki and not much else. Once you've had a few things merged, we generally add you to a team with write access, like the framework team. I'm actually not sure if that would allow you to restart Travis builds, but anyway, I just restarted the failing job.

The travis failure is something I'm seeing more frequently in the last few weeks. These are timing issues that appear to indicate that travis is running slower than it used to.

Your change looks good, but I want to dig a little deeper to make sure it's sufficient. `TestFixtureAttribute` is unfortunately complicated by the fact that it does for fixtures what two separate attributes do for methods. With an argument, it's similar to `TestCaseAttribute` and without it's like `TestAttribute`. Probably a poor decision, but made a long time ago!

ChrisMaddock:
> I'm actually not sure if that would allow you to restart Travis builds...

For the record, it does. You need push access to a repo to restart the build...it's a shame there isn't more fine-grained control. 👎

GCichosz:
Should I investigate those tests failed in Travis build? Because even the tests failed differ between builds.

ChrisMaddock:
I don't think so - the failure I saw should be fixed by #2057 👍

Issue #2035, opened 2017-02-06
NullReferenceException inside failing Assert.That call


yaakov-h:
I have some integration tests written in NUnit that test the code I have that calls the NUnit Engine APIs . This is particularly to ensure that I am treating the NUnit XML correctly and handling certain edge cases

i.e.: Test Runner Unit Tests ---> NUnit Engine ---> Dummy Unit Tests

I am in the process of upgrading this from NUnit 3.4 to 3.6.

When using the 3.6 framework and binaries for all of the above listed components, a failing Test Runner unit test throws a NullReferenceException:

```
Result StackTrace:	
at NUnit.Framework.Assert.ReportFailure(String message)
   at NUnit.Framework.Assert.ReportFailure(ConstraintResult result, String message, Object[] args)
   at NUnit.Framework.Assert.That[TActual](TActual actual, IResolveConstraint expression, String message, Object[] args)
   at NUnit.Framework.Assert.That[TActual](TActual actual, IResolveConstraint expression)
   at My.Test.Case.Here() in C:\Path\To\File.cs:line 108
Result Message:	System.NullReferenceException : Object reference not set to an instance of an object.
```

It appears that in the process of creating, calling into and disposing of engine, `TestExecutionContext.CurrentContext.CurrentResult` becomes null.

yaakov-h:
The error looks very much the same as #2025, #1965 and #1952, except there is no asynchronicity going on here. My tests are synchronous.

CharliePoole:
Can you post a sample test? What framework are you targeting?

yaakov-h:
https://github.com/yaakov-h/NUnit2035/blob/master/NUnitEngineTesting/NUnitEngineRunnerTest.cs#L10

```
Test Name:	NUnitIssue2035
Test FullName:	NUnitEngineTesting.NUnitEngineRunnerTest.NUnitIssue2035
Test Source:	C:\Temp\NUnitEngineTesting\NUnitEngineTesting\NUnitEngineRunnerTest.cs : line 11
Test Outcome:	Failed
Test Duration:	0:00:00.948

Result StackTrace:	
at NUnit.Framework.Assert.ReportFailure(String message)
   at NUnit.Framework.Assert.ReportFailure(ConstraintResult result, String message, Object[] args)
   at NUnit.Framework.Assert.That[TActual](TActual actual, IResolveConstraint expression, String message, Object[] args)
   at NUnit.Framework.Assert.That(Boolean condition)
[...]

PR #2046, opened 2017-02-18
Fix compatibility issue with AssertionException

### LINKED ISSUE #2043: Regression in 3.6.0 when catching AssertionException
@pdpurcell commented on [Thu Feb 16 2017](https://github.com/nunit/docs/issues/168)

In version 3.5.0, catching a failed assertion would allow the test to continue and the test would then pass.  Version 3.6.0 continues running the code, but the final state will be set to failed.  Is this a regression?  I see no mention of this sort of behaviour change in the [release notes](https://github.com/nunit/docs/wiki/Release-Notes#nunit-36---january-9-2017).  
I have [attached an example solution](https://github.com/nunit/docs/files/780587/WithProblem36.zip) that contains two projects that have identical tests; one using 3.5.0 and one using 3.6.0.  
eg:
```
try
{
    Assert.AreEqual(0, 1);
}
catch(AssertionException)
{
    // The failed assertion is caught but in 3.6.0 the test still fails
}
```

---

@CharliePoole commented on [Thu Feb 16 2017](https://github.com/nunit/docs/issues/168#issuecomment-280498245)

The bad news is that this has __never actually been supported__ even though it's a technique promulgated on many blogs and forums. That is...

*NUnit makes no guarantee that failures cause an AssertionException that catching the exception will prevent the failure from happening. This is certainly not a guarantee in NUnit 3 and - to my memory - never was in NUnit V2 either.*

When this technique has come up, I have frequently tried to warn people off using it. There are lots of reasons to avoid it:

* It implies that your test knows something about the framework that is executing it, which is a serious violation of encapsulation - that is, unless you are actually testing the internals of framework itself.

* There are usually supported ways to accomplish whatever you are trying to do, without the  try / catch.

* Where there is an unsupported need, it makes more sense to request a feature, which will be supported now and in the future.

* It's an implementation detail of NUnit, which could change at any time.

Of course, the last point has come true in the latest release.

In NUnit 3.6, the details of the implementation have changed. We still use an exception to terminate execution of your test. It's the only way we have to unwind any nested calls you have made. However, we no longer use the exception to report that the test failed. That info is recorded before the exception is thrown. This allows us to use a consistent approach for single assertion failures, multiple assertion failures and warnings.

As you say, there is no mention of this in the release notes. That's because we felt that there was no need to document an internal change that should never be used by users. For the same reason, we didn't list this as a breaking change. From some of the feedback, it's clear that there are users who did __not__ consider this a detail of implementation and relied on it. There are already several issues.

Most of the issues - like yours - illustrate the "problem" without motivating the need to be able to intercept a failure. In each case, I have asked the submitter of the issue to give a motivating example that would help us meet the need. IOW we need to know __why__ you want to be able to prevent a failure by catching the AssertionException. In the long term, we may want to add features that make it unnecessary for you to do such a thing!

Since you're the fourth or fifth person to raise the issue, it is clear that we need to do something, in spite of my continued belief that you should not really be relying on how NUnit is implemented. I'll close this comment and start a fresh one with a proposal to partially role back the change.

---

@CharliePoole commented on [Thu Feb 16 2017](https://github.com/nunit/docs/issues/168#issuecomment-280501306)

@nunit/framework-team @nunit/core-team 
As stated in my first comment here and on several other issues, I continue to believe that folks who rely on AssertionException and particularly on catching it are making a mistake. "NUnit implements failures via an assertion exception and you can stop the failure by catching it" is __not a sentence in our documentation__ and not part of the API we provide to programmers.

Several people have pointed out reasons why it should work, however:
* Most other frameworks work that way
* It has worked in the past
* Exceptions thrown is part of the API of many methods

I could argue against all of those points, but why bother? It seems that a number of users are affected and I think there's a change that can be made. Here's what I propose...

1. Stop creating the failure message in the result before throwing an AssertionException outside of a Multiple Assert block.

2. Continue to create the failure message without throwing when inside a Multiple Assert block.

3. Upon exit from a MultipleAssertBlock in which failures have occured, throw a (new) MultipleAssertionException rather than an AssertionException, so that the framework knows that one or more failures have already been recorded and need to be reported.

4. Continue to handle warnings as we do, recording them without throwing an exception.

This will ensure that __where we do throw an AssertionException__ it works exactly as before. Those places where we don't throw one are new features so no compatibility issue arises.

I think this is critical enough to warrant a hot fix release. Please let me know if you agree with that as well as with the fix itself. I can do the fix myself pretty quickly once we agree on it.

CharliePoole:
Fixes #2043 

This fix allows users to catch the AssertionException thereby preventing a test from failing. This used to work until NUnit 3.6 and some users have complained about the change so I refactored the implementation to allow the old code to keep working, so long as the failure didn't occur inside a multiple assert block.

This doesn't mean we are now supporting the practice as a feature of NUnit. Just that we are avoiding breaking things we don't have to break. I'd like to add a page on NUnit's use of exceptions in the docs and tell users that it's inadvisable to try to catch them.

### DIFF
--- a/src/NUnitFramework/framework/Assert.cs
+++ b/src/NUnitFramework/framework/Assert.cs
@@ -360,19 +360,18 @@ namespace NUnit.Framework
 
         private static void ReportFailure(string message)
         {
-            // Failure is recorded in <assertion> element in all cases
-            TestExecutionContext.CurrentContext.CurrentResult.RecordAssertion(
-                AssertionStatus.Failed, message, GetStackTrace());
-
             // If we are outside any multiple assert block, then throw
             if (TestExecutionContext.CurrentContext.MultipleAssertLevel == 0)
                 throw new AssertionException(message);
+
+            // Otherwise, just record the failure in an <assertion> element
+            var result = TestExecutionContext.CurrentContext.CurrentResult;
+            result.RecordAssertion(AssertionStatus.Failed, message, GetStackTrace());
         }
 
         private static void IssueWarning(string message)
         {
             var result = TestExecutionContext.CurrentContext.CurrentResult;
-
             result.RecordAssertion(AssertionStatus.Warning, message, GetStackTrace());
         }
 
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -482,6 +482,11 @@ namespace NUnit.Framework.Internal
             if (ex is NUnitException)
                 ex = ex.InnerException;
 
+            if (ex is AssertionException)
+            {
+                AssertionResults.Add(new AssertionResult(AssertionStatus.Failed, ex.Message, StackFilter.DefaultFilter.Filter(ex.StackTrace)));
+            }
+
             if (ex is ResultStateException)
             {
                 string message = ex is MultipleAssertException
--- a/src/NUnitFramework/tests/Assertions/AssertFailTests.cs
+++ b/src/NUnitFramework/tests/Assertions/AssertFailTests.cs
@@ -22,6 +22,7 @@
 // ***********************************************************************
 
 using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
 using NUnit.TestData.AssertFailFixture;
 using NUnit.TestUtilities;
 
@@ -71,6 +72,10 @@ namespace NUnit.Framework.Assertions
                 "CallAssertFail");
 
             Assert.AreEqual(ResultState.Failure, result.ResultState);
+
+            Assert.AreEqual(result.AssertionResults.Count, 1);
+            var assertion = result.AssertionResults[0];
+            Assert.That(assertion.Status, Is.EqualTo(AssertionStatus.Failed));
         }
 
         [Test]
@@ -82,6 +87,11 @@ namespace NUnit.Framework.Assertions
 
             Assert.AreEqual(ResultState.Failure, result.ResultState);
             Assert.AreEqual("MESSAGE", result.Message);
+
+            Assert.AreEqual(result.AssertionResults.Count, 1);
+            var assertion = result.AssertionResults[0];
+            Assert.That(assertion.Status, Is.EqualTo(AssertionStatus.Failed));
+            Assert.That(assertion.Message, Is.EqualTo("MESSAGE"));
         }
 
         [Test]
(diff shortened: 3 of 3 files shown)


CharliePoole:
Travis failure appears to be do to an error in the container itself. Our code never gets executed.

CharliePoole:
Looks like the travis problem fixed itself.

jnm2:
LGTM, do you want me to hit the approve button on things like this or is that more for the core team?

CharliePoole:
Feel free to do a review. In this case, because it's so crucial, I'll click on merge myself when I feel like enough people have seen it. Just tell me if I missed anything. 😄

src/NUnitFramework/framework/Assert.cs
```diff
@@ -360,19 +360,18 @@ private static void ReportFailure(ConstraintResult result, string message, param
...
-            // Failure is recorded in <assertion> element in all cases
-            TestExecutionContext.CurrentContext.CurrentResult.RecordAssertion(
-                AssertionStatus.Failed, message, GetStackTrace());
-
             // If we are outside any multiple assert block, then throw
             if (TestExecutionContext.CurrentContext.MultipleAssertLevel == 0)
                 throw new AssertionException(message);
+
+            // Otherwise, just record the failure in an <assertion> element
+            var result = TestExecutionContext.CurrentContext.CurrentResult;
```
rprouse: This looks good in the context of the issue, but we have seen a lot of NullReferenceExceptions in this method reported since the 3.6 release. Some have been fixed, but I was wondering if we can think of to make the experience better? I hate hiding the error, but I think it might have always been hidden. Maybe,

```C#
if (TestExecutionContext.CurrentContext == null) 
    throw new AssertionException(message);
```

CharliePoole: I've also been looking at this issue. If I can't come up with something better, I'll add your suggested fix here before merging.

rprouse:
@CharliePoole this is good to merge, but I was wondering if you think we should do anything about the null reference exception we have been seeing reported so often recently? My worry with my 'fix' above is that it will just move the error out to where the exception is caught.

I am going to merge this so the release branch is ready. We can do another quick pull request on the branch if we need to.

CharliePoole:
We crossed. I was about to put in your fix. :-)

However, since the error can now only occur in a multiple assert block, we may not see it so frequently any longer.

As I commented on #2035, there is a fairly pervasive problem with how we handle the current execution context. I'll create an issue to refactor this so we no longer blindly create a context when we discover we don't have one.

CharliePoole:
Created issue #2050 to deal with this.

PR #2055, opened 2017-02-24
Refactor TestExecutionContext

### LINKED ISSUE #2050: Creation of TestExecutionContext should be explicit
Originally, in V2,  it seemed to make sense to set `CurrentContext` lazily to a new context in the getter, whenever it was found to be null.

Now, with parallel dispatching, things are more complicated. The context is  changed as different tests begin to execute and the current work item both controls and depends upon it's content. When the context is discovered to be null, it's an error that we should not be masking.

The basic refactoring here is
* Make the getter function as a simple getter, returning null if that's what is found.
* Remove the `GetTestExecutionContext()` method, replacing all calls with use of the property.
* Discover all points in the code where the context is found to be null, and fix them.

Obviously, the last point is going to be difficult.

CharliePoole:
Fixes #2050 

Changes:
* As described in issue #2050, removed the "lazy" creation of TestExecution Context
* This caused five tests to fail, which I fixed.
* I eliminated the clearing of the current context before exiting the framework controller. This was needed at some point but I can't seem to make anything fail right now. Any ideas? We used to have tests for this in NUnit V2 so maybe I'll take a look at them and see if there's a way to make it fail. I don't want us to have code that used to be needed if it isn't needed any longer.

The travis build is not passing. Right now, it's not even succeeding in installing Mono. That may be a drawback of mono latest.

### DIFF
--- a/src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
+++ b/src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
@@ -22,13 +22,18 @@
 // ***********************************************************************
 
 using System;
+using System.Diagnostics;
 using System.Reflection;
 using System.Threading;
 using System.Globalization;
-using NUnit.Framework;
 using NUnit.Framework.Constraints;
 using NUnit.Framework.Internal.Execution;
 using System.Security.Principal;
+using NUnit.Framework.Interfaces;
+
+#if ASYNC
+using System.Threading.Tasks;
+#endif
 
 namespace NUnit.Framework.Internal
 {
@@ -46,6 +51,9 @@ namespace NUnit.Framework.Internal
         IPrincipal originalPrincipal;
 #endif
 
+        DateTime _fixtureStartTime = DateTime.UtcNow;
+        long _fixtureStartTicks = Stopwatch.GetTimestamp();
+        
         [OneTimeSetUp]
         public void OneTimeSetUp()
         {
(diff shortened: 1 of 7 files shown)


CharliePoole:
By "smoke test" do you mean merge and keep using or not merge?

I'll go in and fix the comment.

ChrisMaddock:
Yes, I think this should be merged. I just agree with what you said on the issue - I think there's a wide range of issues this could potentially expose. I don't think there's much more that can be added in the way of unit tests - but I think it would be good to throw some stuff at this before release.

@jnm2 - I get the impression you do much more low-level thread stuff that I do. Would you have a couple of libraries you could run against this, at somepoint before we release it as 3.7? 🙂

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
```diff
@@ -137,7 +137,25 @@ public void Cleanup()
         }
 #endif
 
-        #endregion
+#if !PORTABLE && !NETSTANDARD1_6
```
ChrisMaddock: This will also be a valid test for NETSTANDARD1_6 after #2010 is merged, I think. We should remember to enable it. 👍

CharliePoole: If I understand correctly, #2010 is bringing System.Threading into the .NET Standard build as a "side effect." That will lead to a vast number of other changes, including this one. 😄

jnm2:
@ChrisMaddock I will try to do that this week.

ChrisMaddock:
Great - thanks! No rush. :-)

jnm2:
I ran https://ci.appveyor.com/api/buildjobs/eo486llinl78ie4f/artifacts/NUnit.3.7.0-ci-03706-pr-2055.nupkg multiple times and did not discover any issues, for what it's worth.

CharliePoole:
Thanks for checking! This seems to have fixed several issues.

PR #2058, opened 2017-02-26
Prevent NUnitlite runner from crushing when run with no file provided

### LINKED ISSUE #2045: NUnitlite-runner crashes if no file is provided
If you run nunitlite-runner without a file argument, it crashes. It should actually check and give an error message, just like nunit3-console.

This is slightly tricky because nunitlite-runner passes the arguments on to nunitlite. NUnitlite is often called without a file argument because the tests are in the failing exe. We need to fix this issue without breaking the behavior when nunitlite is called form a test. Most likely, we need a separate entry point.

GCichosz:
This is the solution I was thinking of. When NUnit lite is sure there is absolutely no files to test it returns.

Fixes #2045

### DIFF
--- a/src/NUnitFramework/nunitlite/TextRunner.cs
+++ b/src/NUnitFramework/nunitlite/TextRunner.cs
@@ -230,6 +230,11 @@ namespace NUnitLite
                     if (_testAssembly == null)
                         _testAssembly = AssemblyHelper.Load(testFile);
                 }
+                else
+                {
+                    _textUI.DisplayHelp();
+                    return TextRunner.OK;
+                }
 
 
                 if (_options.WaitBeforeExit && _options.OutFile != null)


CharliePoole:
Travis failure is unrelated - I restarted it.

I'm OK with how you have fixed it even though it doesn't give error messages for each case. What do the other @nunit/framework-team members think?

jnm2:
Lgtm

GCichosz:
@CharliePoole which cases do you mean?

CharliePoole:
@GCichosz The two error cases I described in the issue:

1. Calling Autorun from an assembly without test cases.
2. Using nunitlite-runner without specifying an assembly on the command line.

I think what you have done is an improvement to crashing. 😄  We can do the other stuff separately.

PR #2060, opened 2017-02-27
Make individual tests changing default tolerance non-parallelizable

CharliePoole:
Temporary fix to prevent failures due to NonParallelizable not working correctly on fixture level.


ChrisMaddock:
Looks good. 👍 

I still have the replacement attribute for `DefaultFloatingPointTolerance` sat on a branch somewhere, I really need to get that pushed at some point...

PR #2068, opened 2017-03-03
3.6.1 Merge

rprouse:
This merges the 3.6.1 changes back into master. ~~It isn't ready to merge yet, I still need to resolve conflicts.~~


rprouse:
I've merged latest from master and resolved the conflicts. The merge diff now looks more like I would expect. Can another @nunit/framework-team member review and merge when ready?

CharliePoole:
@rprouse Was failing on master due to a timeout error. I restarted and now it's passing.

However, examining the merged code, I see that the Timeout changes I merged six days ago are not in your merge branch. Master has a timeout of 500ms on the first two TimeoutTests, while your branch still has 50ms. 

TimeoutTests.cs is not listed as a file you changed in the PR so I'm thinking that all changes to master between the time we created the branch and now need to be checked. I can look more closely tomorrow.

CharliePoole:
Just had a thought before going to bed. What I'm seeing may simply be an artifact of how GitHub shows the diff. One of us should merge locally and see how it turns out. The timeou

rprouse:
@CharliePoole I merged latest from master to resolve merge conflicts, so the only changes we should see in this PR are those to CHANGES.txt and the three files you changes in #2046. That is what I am seeing plus a few minor whitespace changes. Those were the only changes made on the release branch only, so the only changes that get merged back.

CharliePoole:
I'm merging and will check the results. I may be confused by a difference in how Travis shows things versus AppVeyor. The job I re-ran, which is now passing,  was the pr build. I would expect that to be using a merge of your branch into master, as AppVeyor does. However, it was failing two timeout tests because they took more than 50 ms where the master has 500ms for the same tests. I'll merge and check the results quickly to reassure myself.

CharliePoole:
Well, it looks good now. What I found odd is that the two builds (pull and pr) both say they are building the 361-merge branch. If you scroll back in the history, you can see that there is generally one build of the submitted branch (pull) followed by a build of master with the branch merged (pr). For some reason that didn't happen here.

PR #2072, opened 2017-03-08
Make WorkDirectory available to TestCaseSource

CharliePoole:
Fixes @2061

Rather than fix `TestExecutionContext.WorkDirectory`, which was actually only used to supply `TestContext.WorkDirectory`, I removed it entirely. `TestContext.WorkDirectory` now gets the directory from a static internal field, which is set when the assembly is loaded.

Ideally, `TestContext.WorkDirectory` would be a static field itself, but that would break existing test code so it remains an instance property but gets its value from the static property.


rprouse:
Looks good to me.

PR #2085, opened 2017-03-19
Changes to Action Attribute Processing

### LINKED ISSUE #652: Add ability to execute test actions before SetUp or OneTimeSetUp
There is quite an old discussion about the issue here: https://groups.google.com/forum/#!topicsearchin/nunit-discuss/testaction/nunit-discuss/IW7qXnYhIJ8

When test action attribute is applied on a class or base class level it should take precedence over methods marked with SetUp and OneTimeSetUp attributes.

Running a test in attached code block I get the following on the test level:

```
BaseSetUp
SetUp
BeforeTest (BaseFixture)
BeforeTest (Fixture)
BeforeTest (Test)
Test
AfterTest (Test)
AfterTest (Fixture)
AfterTest (BaseFixture)
TearDown
BaseTearDown
```

And the following on the fixture level:

```
BaseOneTimeSetUp
OneTimeSetUp
BeforeTest (BaseFixture - OneTime)
BeforeTest (Fixture - OneTime)
AfterTest (Fixture - OneTime)
AfterTest (BaseFixture - OneTime)
OneTimeTearDown
BaseOneTimeTearDown
```

This behaviour causes difficulties when you try to decompose a monolithic setup code in the base class and split it between reusable attributes and setup in the derived class. In this case there is no ability to execute test action code before SetUp. This should probably be changed to the following (or we at least need ability to override existing NUnit behaviour):

```
BeforeTest (BaseFixture)
BaseSetUp
BeforeTest (Fixture)
SetUp
BeforeTest (Test)
Test
AfterTest (Test)
TearDown
AfterTest (Fixture)
BaseTearDown
AfterTest (BaseFixture)
```

And:

```
BeforeTest (BaseFixture - OneTime)
BaseOneTimeSetUp
BeforeTest (Fixture - OneTime)
OneTimeSetUp
OneTimeTearDown
AfterTest (Fixture - OneTime)
BaseOneTimeTearDown
AfterTest (BaseFixture - OneTime)
```

Here is the test code:

``` c#
    public class Action : Attribute, ITestAction
    {
        private readonly string _className;

        public Action(string className)
        {
            _className = className;
        }

        public void BeforeTest(ITest test)
        {
            Trace.Mark(PrepareMarker(test));
        }

        public void AfterTest(ITest test)
        {
            Trace.Mark(PrepareMarker(test));
        }

        private string PrepareMarker(ITest test)
        {
            string s = _className;
            if (test.IsSuite)
            {
                s += " - OneTime";
            }
            return s;
        }

        public ActionTargets Targets { get { return ActionTargets.Suite | ActionTargets.Test; } }
    }

    [Action("BaseFixture")]
    public class BaseFixture
    {
        [OneTimeSetUp]
        public void BaseOneTimeSetUp()
        {
            Trace.Mark();
        }

        [SetUp]
        public void BaseSetUp()
        {
            Trace.Mark();
        }

        [TearDown]
        public void BaseTearDown()
        {
            Trace.Mark();
        }

        [OneTimeTearDown]
        public void BaseOneTimeTearDown()
        {
            Trace.Mark();
        }
    }

    [Action("Fixture")]
    public class Fixture : BaseFixture
    {
        [OneTimeSetUp]
        public void OneTimeSetUp()
        {
            Trace.Mark();
        }

        [SetUp]
        public void SetUp()
        {
            Trace.Mark();
        }

        [TearDown]
        public void TearDown()
        {
            Trace.Mark();
        }

        [OneTimeTearDown]
        public void OneTimeTearDown()
        {
            Trace.Mark();
        }

        [Test]
        [Action("Test")]
        public void Test()
        {
            Trace.Mark();
        }
    }

    public static class Trace
    {
        static Trace()
        {
            Debug.Listeners.Add(new ConsoleTraceListener());
        }

        public static void Mark(string marker = "")
        {
            string s = new StackTrace().GetFrame(1).GetMethod().Name;
            if (!string.IsNullOrEmpty(marker))
            {
                s += " (" + marker + ")";
            }
            Debug.WriteLine(s);
        }
    }
```

CharliePoole:
Fixes #652 

~~We haven't yet resolved the user-facing side of issue #652.~~ Please comment there so I can either close this or carry it on further. What is currently included are changes that enable re-ordering actions as we decide.


CharliePoole:
This PR includes a lot of change, so I thought I should summarize.

Normally, when implementing a new feature or fixing a bug requires structural changes, I break those changes out into a separate PR. In this case, the interplay between the refactoring changes and the target feature led me to make rather big changes incrementally, so it's all together, at least for now.

COMMANDS

I have made individual TestCommands simpler. Most now do only one thing and others will probably be refactored as this work progresses. This originally led to a lot of duplication, so I restructured the Command hierarchy into three main groups: those that do something before the test, those that do something after the test and those that do both. Derived classes simply set the BeforeTest and AfterTest delegates in their constructors, using lambdas. I think this makes things a lot simpler.

I'm also on the way to eliminating the CommandBuilder static class. It no longer builds the command chain for test cases and I expect to eliminate it for suites as well. Currently, the building of commands for test cases is the responsibility of SimpleWorkItem. I'm not sure if this responsibility will stay in the WorkItem hierarchy or move back to the Test hierarchy, where it once was. Currently, there is a lot of checking for just what type of test is being dealt with, so putting it into the Test itself makes some sense.

If we decide it's preferable, all the Command stuff can actually be redone in a separate PR.

ATTRIBUTE ACCESS

It's a small change, but tests now have a method that retrieves all of the attributes of a given type on that test. This encapsulates the difference between methods, classes and assemblies.

ACTIONS

Changes to the command structure include a TestActionCommand (used for methods) and BeforeTestAction and AfterTestAction commands (used for suites). Each command instance handles a single action specified at a single level, which enables actions to be executed in any order we end up specifying.

Currently, the only change in action ordering is that actions specified on test cases execute after the setup. I did this as a proof of concept. Before re-ordering things further, I'm waiting for more responses from the team on issue #652.

****************************************************************************
I'll update this comment to reflect current status as work continues.

CharliePoole:
@nunit/core-team @nunit/framework-team Ready to review and merge!

Sebazzz:
> We haven't yet resolved the user-facing side of issue #652.

What do you mean with that? Isn't that solved by 0b2c401?

CharliePoole:
@Sebazzz It does only if we agree that that small change is the sum of the changes we want to make in ordering action attributes. It's significantly less than was requested.

However, that comment was coupled with a "Do not merge" notation, which I removed. Lacking further comment over a fairly long period, I decided to go with what I have. It's now waiting for review by another committer to be merged. I'll delete the comment as well.

rprouse:
I like the structure and how it simplifies things. It is a very large review, I don't see anything obvious, but stringing everything together in a code diff is difficult, so I need to trust the unit tests. This is so large, we could use a second review. Can anyone else from @nunit/core-team or @nunit/framework-team take a look?

src/NUnitFramework/framework/Internal/Tests/Test.cs
```diff
@@ -289,11 +283,40 @@ public bool IsSuite
         /// <value></value>
         public int Seed { get; set; }
 
-        #endregion
+        /// <summary>
+        /// The SetUp methods.
```
jnm2: Since the other line is `The teardown methods`, I'd prefer to see `The setup methods` here.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -418,6 +416,79 @@ protected void WorkItemComplete()
...
+        // We need to execute setup and teardown methods one level
+        // at a time. However, we can't discover them by reflection
+        // one level at a time, because that would cause overridden
+        // methods to be called twice, once on the base class and
+        // once on the derived class.
+        // 
+        // For that reason, we start with a list of all setup and
+        // teardown methods, found using a single reflection call,
+        // and then descend through the inheritance hierarchy,
+        // adding each method to the appropriate level as we go.
```
jnm2: I've solved this in the past using `BindingFlags.DeclaredOnly`, but this works just as well.
(`BindingFlags.DeclaredOnly` is the only solution if you're also looking at non-public methods.)

src/NUnitFramework/framework/Internal/Tests/Test.cs
jnm2: (Since neither is a sentence, I wouldn't expect to see a period at the end of either, but in any case I'd expect to see the capitalization, use of spaces, and punctuation to be consistent.)

CharliePoole: If we have other, more substantive, changes, I'll fix these comments as well.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
CharliePoole: That was my first thought as well. I can't precisely remember why it didn't work. Possibly because `BindingFlags.DeclaredOnly` was not supported on one of our platforms at the time.

rprouse: I implemented our own version of BindingFlags and many of the reflection APIs for portable, but only included BindingFlags that were in use in our code. That could be why DeclaredOnly isn't used, or it might go back to CF.

Issue #1151, opened 2015-12-22
Include differences in output for Is.EquivalentTo


appel1:
It'd be useful if the output for Is.EquivalentTo included at least one of the differences. Similar to how Is.EqualTo outputs the first index with a difference. 

As it is right now you have to debug the test to figure out why the test failed. 

```
        [Test]
        public void DummyTest()
        {
            var expected = Enumerable.Range('a', 12).Select(n => (char)n);
            var actual = expected.Take(11);

            Assert.That(actual, Is.EquivalentTo(expected));
        }
```

```
Result StackTrace:  at Dummy.DummyTest() in Dummy.cs:line 20
Result Message: 
Expected: equivalent to < 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j'... >
  But was:  < 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j'... >
```

CharliePoole:
Since Equivalent collections may not be in the same order, it's not that simple to come up with an index where the assertion failed. It is possible to list some of the non-matching items, either as "extras", "missing" or both.

JustinRChou:
I wonder if using Linq would be good enough to use for this?
```
var same = expected.Intersect(actual);
// a,b,c,d,e,f,g,h,i,j,k
var difference = actual.Except(expected).Union(expected.Except(actual));
// l
```

CharliePoole:
It's a way to implement the constraint and to get differences... but I think this is more about what we want the error message to look like. It would be interesting to re-implement the constraint using Linq though.

jnm2:
If you use a hash set and a loop, you can perform both set operations in one pass whereas LINQ will necessarily duplicate some of of the work.

jnm2:
I really like the breakdown of Actual, Expected, Missing, and Unexpected.

CharliePoole:
So, in this example, something like...

```
Expected: equivalent to < 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j'... >
  But was:  < 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j'... >
 Missing: <'l'>
```

The EqualConstraint already does this, using "Extra" and "Missing". It's a fairly easy implementation although not giving an "index" as requested. Nevertheless, could be worth doing as an improvement.

jnm2:
[...]

PR #2086, opened 2017-03-20
CollectionIntersectConstraint added

slawwan:
Hi guys!
I've found solution on stackoverflow, but I think it is better to have such constraint as built in.

http://stackoverflow.com/questions/4501598/nunit-assertion-for-an-empty-intersection-between-collection

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionIntersectConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionIntersectConstraint.cs
@@ -1,0 +1,92 @@
+// ***********************************************************************
+// Copyright (c) 2007 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Collections;
+
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// CollectionIntersectConstraint is used to determine whether
+    /// one collection intersects with another
+    /// </summary>
+    public class CollectionIntersectConstraint : CollectionItemsEqualConstraint
+    {
+        private IEnumerable _expected;
+
+        /// <summary>
+        /// Construct a CollectionIntersectConstraint
+        /// </summary>
+        /// <param name="expected">The collection that the actual value is expected to be intersects with</param>
+        public CollectionIntersectConstraint(IEnumerable expected)
+            : base(expected)
+        {
+            _expected = expected;
+        }
+
+        /// <summary> 
+        /// The display name of this Constraint for use by ToString().
+        /// The default value is the name of the constraint with
+        /// trailing "Constraint" removed. Derived classes may set
+        /// this to another name in their constructors.
+        /// </summary>
+        public override string DisplayName { get { return "IntersectsWith"; } }
+
+        /// <summary>
+        /// The Description of what this constraint tests, for
+        /// use in messages and in the ConstraintResult.
+        /// </summary>
+        public override string Description
+        {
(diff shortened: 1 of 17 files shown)


CharliePoole:
I marked this as an idea for discussion, since we really need to decide if we want this feature before doing the actual code review.

My vew... this allows writing

```C#
Assert.That(expected, Is.IntersectsWith(actual));
```

rather than

```C#
Assert.That(expected.Intersect(actual), Is.Not.Empty);
```

I'm not sure that it's justified as a built-in item, especially since it's easy to create as a user extension.

@nunit/core-team @nunit/framework-team What do you think?

jnm2:
I wish there was a community extensions library, because I understand your desire not to bloat the framework and yet there's so much richness that could be expressed by things like this that you currently have to reinvent time and again, or just live without.

rprouse:
@jnm2 you could always start a community extensions library 😄 

Personally, I agree with @CharliePoole that this might be more than we want in the framework. Whenever we add stuff like this, we need to think about all of the possible combinations of uses that might be possible and it ends up bloating the code and potentially confusing users.

slawwan:
There is a difference not only in syntax. Console output for failed tests is more detailed with CollectionIntersectConstraint

Some examples of code and console output

`Assert.That(new[] {1, 2, 3}, Is.Not.IntersectsWith(new[] {3, 4, 5}));`
>   Expected: not intersects with < < 3, 4, 5 > >
  But was:  < 1, 2, 3 >

`Assert.That(new[] {1, 2, 3}.Intersect(new[] {3, 4, 5}), Is.Empty);`
>   Expected: < empty >
>   But was:  < 3 >

`Assert.That(new[] {1, 2, 3}.Intersect(new[] {4, 5, 6}), Is.Not.Empty);`
>   Expected: not < empty >
>   But was:  < empty >

Similar situation with built in SupersetOf and SubsetOf constraints
I can use it

`Assert.That(new[] {1, 2, 3}, Is.SubsetOf(new[] {3, 4}));`
>   Expected: superset of < 3, 4 >
>   But was:  < 1, 2, 3 >

Or use Except method from LINQ like this
`Assert.That(new[] {3, 4}.Except(new[] {1, 2, 3}), Is.Empty);`
>   Expected: < empty >
>   But was:  < 4 >

Of course, at first we added CollectionIntersectConstraint in our library and we faced two unpleasant moments:
1. Syntax
We can write extensions for ConstraintExpression and use constraint this way:
`Is.Not.IntersectsWith(...)`
But we can not write
`Is.IntersectsWith(...)`
because `Is` - static class
2. Collection formating for console output
There is `MsgUtils.FormatValue(...)` method to format values for console output
But `MsgUtils` is internal class and it is invisible for external libraries
The simplest way to format values like NUnit - call this method with reflection

So, before creating "community extensions library", I think, some сontroversial issue should be solved

Some IL to call `MsgUtils.FormatValue(...)` from external library
```
public static class NUnitFormatter
{
    private static readonly Func<object, string> formatValue;

    static NUnitFormatter()
    {
        var msgUtilsType = typeof(Constraint).Assembly.GetType("NUnit.Framework.Constraints.MsgUtils");
        var formatValueMethodInfo = msgUtilsType.GetMethod("FormatValue", BindingFlags.Public | BindingFlags.Static, null, new[] {typeof(object)}, null);

        var dynamicMethod = new DynamicMethod("formatValue", typeof(string), new[] {typeof(object)}, true);
        var ilGenerator = dynamicMethod.GetILGenerator();
        ilGenerator.Emit(OpCodes.Ldarg_0);
        ilGenerator.Emit(OpCodes.Call, formatValueMethodInfo);
        ilGenerator.Emit(OpCodes.Ret);

        formatValue = (Func<object, string>) dynamicMethod.CreateDelegate(typeof(Func<object, string>));
    }

    public static string FormatValue(object value)
    {
        return formatValue(new[] {value});
    }
}
```

jnm2:
@slawwan Hey, did you know that code can be replaced by:

```c#
public static class NUnitFormatter
{
    private static readonly Func<object, string> formatValue;

    static NUnitFormatter()
    {
        var msgUtilsType = typeof(Constraint).Assembly.GetType("NUnit.Framework.Constraints.MsgUtils");
        var formatValueMethodInfo = msgUtilsType.GetMethod("FormatValue", BindingFlags.Public | BindingFlags.Static, null, new[] { typeof(object) }, null);

        formatValue = (Func<object, string>)formatValueMethodInfo.CreateDelegate(typeof(Func<object, string>));
    }

    public static string FormatValue(object value)
    {
        return formatValue(new[] { value });
    }
}
```

JustinRChou:
I remember there was a issue for something like this. #1151 
So wouldn't the intersect solve that issue?

CharliePoole:
@JustinRChou That issue relates to the display format of errors from the existing CollectionsEquivalentConstraint and doesn't introduce any new constraint.

@slawwan I'll try to get back to this today with some further ideas about how we could make it easier to create an extension. However, one thing you might consider is creating your own static "Is" class, deriving from NUnit.Framework.Is. In some cases you may need to have a using alias, but most of the time I have been successful working this way by putting the "Is" class in the same namespace as my tests.

ChrisMaddock:
@nunit/framework-team - I think we should make a decision on this one, before it goes any more stale.

I personally think it might be worthwhile. I'd normally not be a fan of adding 'niche' API methods, but given we already have CollectionSubsetConstraint, CollectionSupersetConstraint, and CollectionContainsConstraint, I wonder if it fits that set quite nicely.

Initial opinion here seems to disagree though. What do other people think?

jnm2:
It seems to fit well with the other `Collection*Constraint`s.

CharliePoole:
As commented way back, neither @rprouse nor I were crazy about this one as an addition to our API. As a meta-point, we don't do well in resolving discussion items - they just seem to stay open. I'm still pretty lukewarm about it, but could go along if others really want it badly. If so, we need to get it updated and do a real code review.

ChrisMaddock:
Courtesy of our new governance docs, I think this is a decision that falls to @rprouse to make, as 'Project Leader'. 😄

rprouse:
I am also not very enthusiastic about this feature. Adding niche features like this becomes a very slippery slope and eventually the API becomes so large that it is nearly impossible for developers to find what they are looking for. @slawwan we appreciate your effort, but I am going to close this one without merging.

PR #2093, opened 2017-03-24
Fix error message for invalid attributes in a single-threaded fixture

### LINKED ISSUE #2087: Enhance error message: Test is not runnable in single-threaded context. Timeout
When using the SingleThreadedAttribute and a TimeOutAttribute on a class, you get the error message:
"Test is not runnable in single-threaded context. Timeout".
It took quite a while to figure out what went wrong (I initially thought that my TestFixtureSetup code took too long).

Suggested error message:
"The SingleThreaded attribute and the Timeout attribute cannot be combined".

CharliePoole:
Fixes #2087

I implemented separate messages for Timeout, RequiredThread and different apartment. I simplified things by not trying to display a message for all errors, but only for the first one detected.

There is one remaining question, not addressed in this fix. It's possible to specify a timeout at a  higher level, rather than on the test, or even at the command line. Currently, this gives the same message in all cases. Options we could consider:

1. Silently ignore a timeout specified at a higher level or at the command-line when we are in a single threaded fixture. We could still give the error message if Timeout is actually specified on the test itself.

2. Treat the entire fixture as an error if a non-zero timeout attribute is specified on the fixture.

3. Treat the entire fixture as an error if a non-zero timeout is specified at a higher level or at the command line.

To my thinking, options 1 and 2 are reasonable, as is leaving it as it is now. Option 3 seems to me to be too intrusive since it prevents using a global timeout for other tests. 

As stated earlier, __allowing__ timeout on a single-threaded fixture is a bigger and separate issue.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
@@ -164,15 +164,6 @@ namespace NUnit.Framework.Internal.Execution
         private ApartmentState CurrentApartment { get; set; }
 #endif
 
-        [Flags]
-        private enum OwnThreadReason
-        {
-            NotNeeded = 0,
-            RequiresThread = 1,
-            Timeout = 2,
-            DifferentApartment = 4
-        }
-
         #endregion
 
         #region Public Methods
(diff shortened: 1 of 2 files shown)


rprouse:
Code looks good. As for options, I think leaving it as is or option 1 would be fine.

Issue #1143, opened 2015-12-17
NUnitLite - Explore flag does not apply where filter to output


chriscore:
Hi there

Running nunit3-console.exe with arguments similar to the following does not apply the filter specified in /where to the results which are seen on the output.

"C:\MyRepo\MyAssembly.dll" /explore /noheader /where="cat==Awesome"

All tests are returned by /explore, including those which do not satisfy cat==Awesome

This was found in NUnit Console Runner v3.0.5813

CharliePoole:
I was uncertain whether it made sense to use a filter when listing tests, but I suppose it does if you tried it!

chriscore:
Hey Charlie

Thanks, before a test run I am currently trying to call this to see how many tests my filter would return - it would certainly be very useful for me at this point :)

Great work with nunit 3 so far

Chris

CharliePoole:
@chriscore If you have a runner, just call CountTestCases on it with your filter!

If you are simply scripting nunit3-console, that's not an option, of course.

chriscore:
@CharliePoole I don't currently have a runner instance as I haven't been able to figure out how to get that going, like you say I have just scripted nunit3-console, but having that control in process would be ideal.

I did find it difficult to find how to host a runner, do you have a C# snippet anywhere which might give some clues how to start? Being able to load and run tests in process would be ideal for what I'm trying to do here.

CharliePoole:
Add a reference to nunit.engine.api.dll. Then (simplified)...

```
ITestEngine engine = TestEngineActivator.CreateInstance();
TestPackage package = new TestPackage(ASSEMBLY_PATH);
ITestRunner runner = engine.GetRunner(package);
```

This leaves out setting options, using a filter, etc. For the full story, start here:
    https://github.com/nunit/nunit/blob/master/src/NUnitConsole/nunit3-console/Program.cs#L111

chriscore:
Thanks so much for that, I didn't know about the TestEngineActivator!

rprouse:
When this is fixed, ensure it also works with NUnitLite. See #1213

NCarter3:
I'd also love to see filters working with --explore!

ChrisMaddock:
Copying this issue to the console repo, and keeping this one for NUnitLite

CharliePoole:
[...]

PR #2094, opened 2017-03-24
Apply filter to output of Explore

Poimen:
This is still a work in progress. This PR as a result of the discussion on the group chat regarding this functionality (apologies for the delay as I have been unwell)

Updates to nunit framework for nunitlite to display explored tests category limits.

Command line used:
nunit.framework.tests.dll /explore /where="cat==FixtureCategory" /noheader

The XML processing is the difficult one. At the moment I'm experimenting with the AddToXML in the Test class.

### DIFF
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -46,8 +46,6 @@ namespace NUnit.Framework.Api
 
         private ITestAssemblyBuilder _builder;
         private ManualResetEvent _runComplete = new ManualResetEvent(false);
-        private Assembly _assembly;
-        private string _assemblyName = string.Empty;
 
 #if !PORTABLE
         // Saved Console.Out and Console.Error
@@ -150,7 +148,6 @@ namespace NUnit.Framework.Api
         public ITest Load(string assemblyName, IDictionary<string, object> settings)
         {
             Settings = settings;
-            _assemblyName = assemblyName;
 
             if (settings.ContainsKey(FrameworkPackageSettings.RandomSeed))
                 Randomizer.InitialSeed = (int)settings[FrameworkPackageSettings.RandomSeed];
@@ -168,7 +165,6 @@ namespace NUnit.Framework.Api
         public ITest Load(Assembly assembly, IDictionary<string, object> settings)
         {
             Settings = settings;
-            _assembly = assembly;
 
             if (settings.ContainsKey(FrameworkPackageSettings.RandomSeed))
                 Randomizer.InitialSeed = (int)settings[FrameworkPackageSettings.RandomSeed];
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -193,7 +193,10 @@ namespace NUnit.Framework.Api
         public ITest ExploreTests(ITestFilter filter)
         {
             if(LoadedTest == null)
-                throw new InvalidOperationException("The CountTestCases method was called but no test has been loaded");
+                throw new InvalidOperationException("The ExploreTests method was called but no test has been loaded");
+
+            if(filter == TestFilter.Empty)
+                return LoadedTest;
 
             return new TestSuite(LoadedTest as TestSuite, filter);
         }
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -46,6 +46,8 @@ namespace NUnit.Framework.Api
 
         private ITestAssemblyBuilder _builder;
         private ManualResetEvent _runComplete = new ManualResetEvent(false);
+        private Assembly _assembly;
+        private string _assemblyName = string.Empty;
 
 #if !PORTABLE
         // Saved Console.Out and Console.Error
@@ -148,6 +150,7 @@ namespace NUnit.Framework.Api
         public ITest Load(string assemblyName, IDictionary<string, object> settings)
         {
             Settings = settings;
+            _assemblyName = assemblyName;
 
             if (settings.ContainsKey(FrameworkPackageSettings.RandomSeed))
                 Randomizer.InitialSeed = (int)settings[FrameworkPackageSettings.RandomSeed];
@@ -165,6 +168,7 @@ namespace NUnit.Framework.Api
         public ITest Load(Assembly assembly, IDictionary<string, object> settings)
         {
             Settings = settings;
+            _assembly = assembly;
 
             if (settings.ContainsKey(FrameworkPackageSettings.RandomSeed))
                 Randomizer.InitialSeed = (int)settings[FrameworkPackageSettings.RandomSeed];
@@ -183,6 +187,26 @@ namespace NUnit.Framework.Api
                 throw new InvalidOperationException("The CountTestCases method was called but no test has been loaded");
 
             return CountTestCases(LoadedTest, filter);
+        }
+
+        /// <summary>
+        /// Explore the test cases using a filter
+        /// </summary>
+        /// <param name="filter">The filter to apply</param>
+        /// <returns>Test Suite that contains </returns>
+        public ITest ExploreTests(ITestFilter filter)
+        {
+            if(LoadedTest == null)
+                throw new InvalidOperationException("The CountTestCases method was called but no test has been loaded");
+
+            var filterSettings = new Dictionary<string, object>(Settings);
+            filterSettings.Add(FrameworkPackageSettings.InternalTestFilter, filter);
+
+            if(!string.IsNullOrWhiteSpace(_assemblyName))
+            {
+                return _builder.Build(_assemblyName, filterSettings);
+            }
+            return _builder.Build(_assembly, filterSettings);
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -202,7 +202,7 @@ namespace NUnit.Framework.Api
             var filterSettings = new Dictionary<string, object>(Settings);
             filterSettings.Add(FrameworkPackageSettings.InternalTestFilter, filter);
 
-            if(!string.IsNullOrWhiteSpace(_assemblyName))
+            if(_assemblyName != string.Empty)
             {
                 return _builder.Build(_assemblyName, filterSettings);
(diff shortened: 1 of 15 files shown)


CharliePoole:
Fixes #1143 

@Poimen This is how we tie PRs to Issues, rather than using the title. The "Fixes" comment is understood by github, whereas putting the number in the title is only useful to humans. 😄

CharliePoole:
@Poimen In our discussion on the developer list, I defined three possible approaches to this problem.
1. pass a filter in the load call, so that anything not matching the filter is ignored.

2. pass a filter as an argument to explore and use it to modify what's returned (all tests are loaded but only some are returned by explore)

3. put the logic to filter the XML in the runner or the engine rather than the framework.

Reviewing the discussion, we appear to have decided to go with option 2. 

Option 1, as you pointed out would cause problems in a gui situation, where the tests were run multiple times, with and without filters. It also suffers from a much more serious problem: our filter implementations in the framework are designed to work with __tests__ and we don't have tests to filter until after we load them. That would mean completely re-impementing the filter hierarchy to work on classes and methods rather than tests. Predicting whether a test would be included would mean performing pretty much the same logic we already do in loading tests, since we would need to know the name of the test, it's categories, etc.

Option 3 would require two completely different implementations, one in NUnitLite that worked with test objects and another in the engine to work with XML. The second would again call for re-implementing the filter hierarchy based on XML.

So I think that we did right in saying that Load should continue to load all tests while Explore would use it's filter option to only return those tests that match the filter. An additional benefit of this would be that only the framework would change and the engine, if it passed in a filter, would only get back those tests it wanted, without any further work.

That said, looking at your changes so far, I seem to see a mix of option 2 and 3. That is, you are making changes to how the explore output is generated but also changes to parts of NUnitLite. I'd like to understand why you had to do the latter to make this work and see if we can avoid it.

I'll add in-line comments to the changes that I'm wondering about.

src/NUnitFramework/nunitlite/OutputManager.cs
```diff
@@ -120,7 +120,7 @@ public void WriteTestFile(ITest test, OutputSpecification spec)
                         "spec");
             }
 
-            outputWriter.WriteTestFile(test, outputPath);
+            outputWriter.WriteTestFile(test, filter, outputPath);
             Console.WriteLine("Tests ({0}) saved as {1}", spec.Format, outputPath);
         }
     }
```
CharliePoole: An example of what I'm confused by. Can't we just pass a pre-filtered test hierarchy to the write methods? If it becomes the job of the output writers to do the filtering, then every output writer will have to support it, including those that are implemented as extensions to the engine. And in the case of the engine, we will be changing a public API.

src/NUnitFramework/nunitlite/TextRunner.cs
```diff
@@ -318,21 +318,21 @@ public void ReportResults(ITestResult result)
             _textUI.DisplaySummaryReport(Summary);
         }
 
-        private int ExploreTests()
+        private int ExploreTests(ITestFilter filter)
         {
 #if !PORTABLE
             ITest testNode = _runner.LoadedTest;
```
CharliePoole: I think this line is the key to the problem I pointed out. In our online discussion, we talked about having NUnitLite actually call an Explore method on the runner, with a filter applied. If testNode were already filtered, then none of the other stuff would be needed. Does this seem feasible to you?

src/NUnitFramework/nunitlite/OutputManager.cs
Poimen: My intention was to filter the tests at the lowest possible point to allow any other layer to fiddle with the filter if required.

On further consideration, I can see that it complicates the implementation.

I'll see about the pre-filtered selection of tests.

src/NUnitFramework/nunitlite/TextRunner.cs
Poimen: Ok - I think that I misunderstood here...

The runner in nunitlite comes from:
`private ITestAssemblyRunner _runner;`

And there is no `Explore` method on that interface. Hence I updated the current code...

If I look at the `FrameworkController` code for the other `ExploreTests` implementation I can see why putting it into the interface is preferable. I had missed this last time I looked at it.

I'll look at implementing the `ExploreTests` onto that interface.

On that - is there any objection/reason the code should not re-read the assembly to filter? i.e. if I re-use the builders to re-build the tests based on a filter set - or is it preferable to just modify the Test list?

src/NUnitFramework/nunitlite/OutputManager.cs
Poimen: @CharliePoole  I have now removed the filtering that happens in the output writer

src/NUnitFramework/nunitlite/TextRunner.cs
Poimen: @CharliePoole  I have updated the runner to now have an Explore(filter) function. I have gone with changing the builder to rebuild a new test list based on a filter. I tried a number of other solutions to the problem but this seemed to be the least intrusive.

The `LoadedTest` suite `Test` list being readonly was the hardest challenge. Attempting to work around it by creating new lists was getting very cumbersome. I even attempted to remove tests based on the filter. The secondary issue I was having was attempting to clone the `LoadedTest` structure. If I didn't them the changes kept changing the `LoadedTest` internal list of tests, which was undesirable. 

Hence I went for building a new test suite based on the filter.

Let me know what you think.

CharliePoole:
This looks good. It will work for the console runner or nunitlite because they only perform one operation and then exit. I think there remains an issue for any clients that load a test assembly and then perform __multiple__ operations on it.

Imagine a gui that has some sort of commands - menu entries, perhaps - for running a set of tests under a filter and exploring a set of tests under a filter. A sequence of actions might be...

1. Load the tests (prerequisite to any filtering)
2. Run tests in a category A
3. Run all tests
4. Run tests in category B
5 Save tests to a file that match Category C

This indicates that __no__ filtering should occur when the tests are loaded. By modifying DefaultTestAssemblyBuilder, you are changing step 1. This implies that the tests must now be reloaded each time they are either run or explored. I believe you have taken care to do that, which makes the code correct, but loading of tests is one of the most resource-intensive things we do.

I think this can be resolved by keeping the filtering in the Runner rather than the builder. The Explore method can apply the filter to the tests that have already been loaded and only create XML output for them. The loaded tests should not change as a side effect of running or exploring under a filter.

I __think__ this is an easy change but let me know if you see any problems with it. I'll follow up with more detailed line comments.

Poimen:
I was concerned about the resource intensity of the loading so I have tried not to change the builder. As mentioned above the Test list is read-only and the only current way of building a new test list is through the builder.

Due to the linked-list nature of the test lists, filtering and copying the tests to a new list is not quite as trivial as one would think :( 

I have a secondary fix that creates a copy-constructor like operation to the test suite and test classes. In this it creates a copy of the test cases that pass the filter. I'll have a re-look at that code to see what can be done.

Thanks for the input so far :)

CharliePoole:
My suggestion would be to leave loaded tests unchanged and do a recursive-descent copy under the filter in the runner's explore method. That would bypass the actual loading of assembly and tests.

Poimen:
100% - My concern was the implementing the copy code through the test suite and test classes as this would be extra maintenance for the project. I was also getting memory aliasing between the test suite and the new filtered listing.

If you are happy with doing copies, I'll revive that branch.

Poimen:
@CharliePoole Do you want me to keep the changes in the builder at all?

Poimen:
@CharliePoole I have reverted the builder changes and implemented the filter in a copy style constructor of the Test Suite.

src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
```diff
@@ -151,6 +166,22 @@ public void LoadTestsAction_BadFile_ReturnsNonRunnableSuite()
...
+            new FrameworkController.LoadTestsAction(_controller, _handler);
+            new FrameworkController.ExploreTestsAction(_controller, FIXTURE_CAT_FILTER, _handler);
+            var result = TNode.FromXml(_handler.GetCallbackResult());
+
+            Assert.That(result.Name.ToString(), Is.EqualTo("test-suite"));
+            Assert.That(result.Attributes["type"], Is.EqualTo("Assembly"));
+            Assert.That(result.Attributes["id"], Is.Not.Null.And.StartWith("ID"));
+            Assert.That(result.Attributes["name"], Is.EqualTo(EXPECTED_NAME));
+            Assert.That(result.Attributes["runstate"], Is.EqualTo("Runnable"));
+            Assert.That(result.Attributes["testcasecount"], Is.EqualTo(MockAssembly.Tests.ToString()));
```
CharliePoole: This assertion proves that the filter didn't work! The count is equal to all the mock tests, with no filtering. That's because you haven't actually changed the code in FrameworkController to use the filter.

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -186,6 +186,19 @@ public int CountTestCases(ITestFilter filter)
...
 
         /// <summary>
+        /// Explore the test cases using a filter
+        /// </summary>
+        /// <param name="filter">The filter to apply</param>
+        /// <returns>Test Suite that contains </returns>
+        public ITest ExploreTests(ITestFilter filter)
+        {
+            if(LoadedTest == null)
+                throw new InvalidOperationException("The CountTestCases method was called but no test has been loaded");
```
jnm2: `nameof` 😉

CharliePoole: Except if you copy `nameof(CountTestCases)` it remains "CountTestCases" 😄

jnm2: Oh dear, good point 🙄

Poimen: Copy-n-paste error :blush:

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -186,6 +186,19 @@ public int CountTestCases(ITestFilter filter)
...
+
+            return new TestSuite(LoadedTest as TestSuite, filter);
```
Poimen: Agreed - I'll add that

src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
Poimen: Yes - I have been focusing on the nunitlite runner rather that the other plumbing for the framework. Once I have the basic structure done (which I think it pretty much there with minor update), I'll add in the extra items into the mix :smiley:

rprouse:
Looks good to me too. Thanks for the contribution.

PR #2096, opened 2017-03-25
#2079 Exposed object[] ITest.Arguments {get; } property

### LINKED ISSUE #2079: Make TestMethod.Arguments public or otherwise accessible (e.g. TestContext)
NUnit 3.6.1 (from NuGet)

I need a property in test's base class that returns a different value depending on test method's parameters (details in this SO question: http://stackoverflow.com/questions/42763574/how-to-get-current-parameter-value-from-testcontext - I don't feel they are strictly relevant here though).

I've created a `ITestAction` attribute and in the `BeforeTest` call I have access to `TestMethod` instance describing currently executed test. 

If I could read the `Arguments` property I could react and accordingly set some properties in the fixture's base. Unfortunately, it's internal. This doesn't seem justified - a `TestMethod` description should naturally present method call arguments to the clients.

As a side note, I've tried also to use properties for this scenario, but even though I added them to my `TestCaseSource` results, they are not accessible from test method's level (probably related to issues #1358, #548 or #796).

Suremaker:
Fixes #2079


src/NUnitFramework/framework/Internal/Tests/Test.cs
```diff
@@ -188,6 +188,14 @@ public virtual string MethodName
         }
 
         /// <summary>
+        /// The arguments to use in creating the test or null if none required.
+        /// </summary>
+        public virtual object[] Arguments
+        {
+            get { return null; }
```
CharliePoole: Why not object[0]? Alternatively, why not just make this abstract since you end up overriding it anyway?

src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
```diff
@@ -174,7 +176,7 @@ public override int TestCaseCount
         /// <summary>
         /// The arguments to use in creating the fixture
         /// </summary>
-        public object[] Arguments { get; internal set; }
+        public override object[] Arguments { get { return arguments; } }
```
CharliePoole: I don't understand the change away from the automatic property, which works fine and is the way we generally do things nowadays.

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
```diff
@@ -256,6 +256,13 @@ public void TestCanAccessItsOwnProperties()
             Assert.That(TestExecutionContext.CurrentContext.CurrentTest.Properties.Get("Answer"), Is.EqualTo(42));
         }
 
+        [Test]
```
CharliePoole: Although permitted, we don't use the extra `[Test]` in our own tests - except where we test that it works, of course.

src/NUnitFramework/framework/Internal/Tests/Test.cs
Suremaker: Sure, I can make it abstract.

src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
Suremaker: Unfortunately with the change above, it is not possible to add setter to overridden property if abstract/virtual version of it does not have setter.

As the `TestMethod` implementation does not use setters, i.e. it is defined as follows:
```c#
public override object[] Arguments
{
    get { return parms != null ? parms.Arguments : null; }
}
```
I did not wanted to add setter to `Test` class as it will have no sense from `TestMethod` perspective.
With those constraints in mind I implemented `TestSuite` with separate `SetArguments()` method.

I'll take a second look at TestSuite to check if I can alter it so it will not expose setter for Arguments but instead accept them in constructor.

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
Suremaker: Here I was not really sure where to add tests for the `Test/TestMethod/TestSuite` classes. I see there is a lot of various test files checking the behaviour of functionalities depending on such classes but I could not locate the tests for those classes themselves.

Because of that I implemented the test in place that may use those exposed Arguments property, and written the test in the same way as all other tests are written in that file.

Could you please point me to the place that would be best suited for testing this change?

Suremaker:
Hello, I have replied to the comments and will take a look to apply all the suggestions.
With the comment above, you could please explain what do you have in mind? I checked locally that all the existing and my new tests have passed. I see that on CIs where build is run for PullRequests they pass as well, so I am not sure what I missed.

Suremaker:
Ok, I have modified `Test` class to have abstract Arguments property, reworked the `TestSuite` to not have `SetArguments()` method and added more tests for `TestSuite` as well as `TestMethod` classes.

Please review.

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
CharliePoole: https://github.com/nunit/nunit/blob/master/src/NUnitFramework/tests/TestContextTests.cs

src/NUnitFramework/tests/Internal/TestMethodTests.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            Assert.That(test.HasChildren, Is.True);
+            Assert.That(test.Arguments, Is.EqualTo(new object[0]));
+            Assert.That(test.Tests[0], Is.TypeOf<TestMethod>());
+            Assert.That(test.Tests[1], Is.TypeOf<TestMethod>());
+            Assert.That(test.Tests[0].Arguments, Is.EqualTo(new object[] { 42, "abc" }));
+            Assert.That(test.Tests[1].Arguments, Is.EqualTo(new object[] { 24, "cba" }));
+        }
+    }
+}
```
CharliePoole: These are good tests. I think I might have added them to https://github.com/nunit/nunit/blob/master/src/NUnitFramework/tests/Internal/NUnitTestCaseBuilderTests.cs but I see no problem in keeping them here.

Suremaker:
Oh, I missed the change about TestContext.CurrentContext.Test - thank you for clarification.
Both changes are done now in the newest commit.

src/NUnitFramework/testdata/TestFixtureData.cs
```diff
@@ -121,6 +121,7 @@ public void SomeTest(int x)
     [TestFixture]
     public class FixtureWithParameterizedTestAndArgsSupplied
     {
+        [Test]
```
CharliePoole: Perhaps I didn't state it clearly enough... I was asking you to remove the redundant `[Test]` attributes.... not add more!

CharliePoole:
I approved it as is. It would still be nice to get rid of those unneeded `[Test]` attributes, but not essential, We will have to wait for a second committer to approve the merge, as usual.

Thanks for your work!

src/NUnitFramework/framework/Internal/Tests/TestMethod.cs
```diff
@@ -76,17 +78,19 @@ internal object ExpectedResult
...
         #endregion
 
         #region Test Overrides
 
         /// <summary>
+        /// The arguments to use in executing the test method.
+        /// </summary>
+        public override object[] Arguments
+        {
+            get { return parms != null ? parms.Arguments : NoArguments; }
```
ChrisMaddock: Is this change from returning `null` to `object[0]` definitely ok? This property is eventually used to invoke the method via `MethodBase.Invoke()` - which [states null should be used when there are no parameters](https://msdn.microsoft.com/en-us/library/a89hcwhh(v=vs.110).aspx). 

If all the tests pass, I assume it works, but I'm a little wary that I might be missing something!

ChrisMaddock: (Explanation to the above comment: this property previously returned null if `parms == null`, rather than `NoArguments`. The diff doesn't show that very well!)

CharliePoole:
@ChrisMaddock Regarding implementing `ITest`... It's conceivable that somebody could implement it in creating an extension that defined a new kind of test, but it's more likely that they would derive from abstract class `Test`. However, there is currently no way to create an extension of this type without a dependency on a particular framework version, so I don't anticipate any problems.

I have always used `object[0]` rather than null in making explicit calls to reflection and in fact I thought it was required. I suspect that I'm thinking back to .NET 1 or CF. FYI, I was the one who asked @Suremaker to change from null to the empty array. Usually, it's not such a great idea to return null from a property that implements a collection. I'll check into this further before either merging or asking for another 😞 change.

Suremaker:
Sorry, it looks like I misunderstood the change request regarding [Test] attributes. I always thought it is required to have [Test] attribute while [TestCase] just provides values for it. After few years of using nunit I learned something new ;)
Currently I am away, but I'll correct the redundant attributes when I come back in next couple of hours.

Regarding `null` to `object[0]` I can also update the method summary, as long as you are happy with the change.

ChrisMaddock:
Ah, sorry. I didn't read enough to see that had already been commented on!

Suremaker:
I sent the last PR with post review corrections.
I see that one of the build failed due to operation timeout.

ChrisMaddock:
Is this ready to merge, @CharliePoole? It has conflicts to resolve, but I think any one of us could do that.

CharliePoole:
Restarted the failing travis job.

ChrisMaddock:
Merging, thanks @Suremaker!

bartoszkp:
Thank you all, this is really great!

src/NUnitFramework/tests/Internal/TestMethodTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        [Test]
+        public void CaptureArgumentsForParameterizedTestMethodWithMultipleArguments()
+        {
+            var test = _builder.BuildFrom(new MethodWrapper(typeof(FixtureWithParameterizedTestAndMultipleArgsSupplied), nameof(FixtureWithParameterizedTestAndMultipleArgsSupplied.SomeTest)));
+
+            Assert.That(test.HasChildren, Is.True);
+            Assert.That(test.Arguments, Is.EqualTo(new object[0]));
+            Assert.That(test.Tests[0], Is.TypeOf<TestMethod>());
+            Assert.That(test.Tests[1], Is.TypeOf<TestMethod>());
+            Assert.That(test.Tests[0].Arguments, Is.EqualTo(new object[] { 42, "abc" }));
```
oznetmaster: Order of the TestCase arguments is not defined.  This test will fail some of the time.

src/NUnitFramework/tests/Internal/TestFixtureTests.cs
```diff
@@ -112,6 +112,29 @@ public void CanRunConstructorWithArgsSupplied()
...
+            Assert.That(fixture.Arguments, Is.EqualTo(new object[0]));
+        }
+
+        [Test]
+        public void CapturesArgumentsForConstructorWithMultipleArgsSupplied()
+        {
+            var fixture = TestBuilder.MakeFixture(typeof(FixtureWithMultipleArgsSupplied));
+            Assert.True(fixture.HasChildren);
+            Assert.That(fixture.Tests[0].Arguments, Is.EqualTo(new[] { 7, 3 }));
+            Assert.That(fixture.Tests[1].Arguments, Is.EqualTo(new[] { 8, 4 }));
```
oznetmaster: Order of the TestFixture Tests  is not defined.  This test will fail some of the time.

oznetmaster:
As I noted in the code, and many time elsewhere, the order of the TestCase and Tests within TestFixture is not defined, and will change from environment to environment.

Tests should not be written assuming a given order for these.

This PR does just that.

Issue #2097, opened 2017-03-26
CollectionAssert.AreEqual should always compare elements, ignoring IEquatable if implemented


JeffAshton:
I came across this issue when comparing collections of type System.Collections.Immutable.ImmutableArray<T>.  This works correctly in NUnit 2, but now fails in NUnit 3.  The reason is that ImmutableArray implements IEquatable<T> under the covers and checks that the underlying arrays are the same reference.

https://github.com/dotnet/corefx/blob/master/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray_1.Minimal.cs#L317

CollectionAssert.AreEqual claims to check the items, but with the reuse of NUnitEqualityComparer.  This likely worked before this change that put comparing IEquatable<T> before IEnumerable<T>, but then comparing objects which also happen to implement IEnumerable<T> would have been broken.

https://github.com/nunit/nunit/commit/7c7487036accf1d530f4338fbd74d62dc9175993#diff-ebfb835776d7283aa42e5db9b6403200

When I'm comparing and Enumerable<T>, I shouldn't know about Equals(object) or IEquatable<T>, or IDictionary<TKey,TValue>, etc.  I'm giving an stream of items, and should only treat it as such.  That fact that CollectionAssert takes IEnumerable, but then does casting under the covers is a huge smell.

When I'm comparing Object's, I shouldn't know about enumerables.  For example, consider an implementation for [System.Collection.IGrouping<TKey,TElement>](https://msdn.microsoft.com/en-us/library/bb344977(v=vs.110).aspx).  To be equal, both the key and items have to match, but the implementation for NUnitEqualityComparer would disregard the Key altogether because they're an Enumerable<T>.

JeffAshton:
In reviewing the implementation for [NUnitEqualityComparer.AreEqual](https://github.com/nunit/nunit/blob/757feb7a721ee8ef21b20e76ada61d00834c3bc7/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs#L132), I can also say that anything which implements IDictionary is problematic to the same as IEnumerable<T>.

It is only safe to do this sort of optimization for well known:

* structs
* sealed classes
* specific instance of a type, but those which inherit ( for example if type is Dictionary<TKey,TValue>, but it isn't safe for:

```
public class ExtendedDictionary<TKey,TValue> : Dictionary<TKey, TValue > {
    public int ExtendedProperty { get; }
}
```

[...]

PR #2098, opened 2017-03-26
Tests to prove issue 2097

JeffAshton:
Providing you with tests to demonstrate how CollectionAssert.Equal and Assert.Equal are broken with respsent to issue https://github.com/nunit/nunit/issues/2097

Just to be clear, these tests obviously fail at the moment as expected:

https://travis-ci.org/nunit/nunit/jobs/215316103#L1553

### DIFF
--- a/src/NUnitFramework/tests/Assertions/CollectionAssertTest.cs
+++ b/src/NUnitFramework/tests/Assertions/CollectionAssertTest.cs
@@ -47,14 +47,14 @@ namespace NUnit.Framework.Assertions
         [Test()]
         public void ItemsOfType()
         {
-            var collection = new SimpleObjectCollection("x", "y", "z");
+            var collection = new SimpleEnumerable("x", "y", "z");
             CollectionAssert.AllItemsAreInstancesOfType(collection,typeof(string));
         }
 
         [Test]
         public void ItemsOfTypeFailure()
         {
-            var collection = new SimpleObjectCollection("x", "y", new object());
+            var collection = new SimpleEnumerable("x", "y", new object());
 
             var expectedMessage =
                 "  Expected: all items instance of <System.String>" + Environment.NewLine +
@@ -69,14 +69,14 @@ namespace NUnit.Framework.Assertions
         [Test()]
         public void ItemsNotNull()
         {
-            var collection = new SimpleObjectCollection("x", "y", "z");
+            var collection = new SimpleEnumerable("x", "y", "z");
             CollectionAssert.AllItemsAreNotNull(collection);
         }
 
         [Test]
         public void ItemsNotNullFailure()
         {
-            var collection = new SimpleObjectCollection("x", null, "z");
+            var collection = new SimpleEnumerable("x", null, "z");
 
             var expectedMessage =
                 "  Expected: all items not null" + Environment.NewLine +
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/tests/Assertions/CollectionAssertTest.cs
```diff
@@ -47,14 +47,14 @@ public class CollectionAssertTest
         [Test()]
         public void ItemsOfType()
         {
-            var collection = new SimpleObjectCollection("x", "y", "z");
+            var collection = new SimpleEnumerable("x", "y", "z");
```
JeffAshton: CollectionAssert takes IEnumerable, so switch to use SimpleEnumerable to match.  Down below, I added IEquatable<T> to SimpleEnumerable<T> to match the correct behavior of Object.

src/NUnitFramework/tests/Assertions/CollectionAssertTest.cs
```diff
@@ -133,8 +133,8 @@ public void UniqueFailure_WithTwoNulls()
         [Test]
         public void AreEqual()
         {
-            var set1 = new SimpleObjectCollection("x", "y", "z");
-            var set2 = new SimpleObjectCollection("x", "y", "z");
+            var set1 = new SimpleEnumerable("x", "y", "z");
+            var set2 = new SimpleEnumerable("x", "y", "z");
 
             CollectionAssert.AreEqual(set1,set2);
```
JeffAshton: This test fails now, since Object.ReferenceEquals( set1, set2 ) == false, but CollectionAssert.AreEqual should be comparing the items, which are the same

src/NUnitFramework/tests/Assertions/CollectionAssertTest.cs
```diff
@@ -322,8 +322,8 @@ public void AreNotEqual()
         [Test]
         public void AreNotEqual_Fails()
```
JeffAshton: This counter part test also fails now, since Object.ReferenceEquals( set1, set2 ) == false, but the items are the same.

src/NUnitFramework/tests/Constraints/EqualConstraintTests.cs
```diff
@@ -520,6 +521,90 @@ public void ErrorIfUlpsIsUsedOnDecimal()
 
         #endregion
 
+        #region ObjectsThatAreAlsoIEnumerableEquality
+
+        public sealed class ObjectsThatAreAlsoIEnumerableEquality
```
JeffAshton: These tests demonstrate the flip side, where clearly the object isn't equal, but because it so happens to implement IEnumerable<T>, it matches incorrectly

src/NUnitFramework/tests/Assertions/CollectionAssertTest.cs
CharliePoole: Not sure what you're doing here. This isn't a test of equality at all.

JeffAshton: I was just trying to match the IEnumerable interface.   But I wouldn't expect this to be merged as is.   Was trying to help communicate the conflicting nature with tests. Fixing one breaks the other when just reordering.

ChrisMaddock:
Given #2097 is closed, and #2118 is merged - are we good to close this? @JeffAshton, @CharliePoole? 🙂

CharliePoole:
Should have been closed earlier.

PR #2101, opened 2017-03-28
Removing some unused legacy code

JeffAshton:


rprouse:
I am good with this. Anyone see any reason that we might want to reintroduce that code?

Issue #2020, opened 2017-01-31
Better message when timeout fails


jnm2:
Currently when a timeout fails NUnit reports, "Test cancelled by user." While in a certain sense it's not wrong, it's confusing. It led me to believe that I was clicking the test abort icon in ReSharper. (I've been having tooling issues unrelated to NUnit.) In an integration test, it makes it sound as though the build is being canceled.

It's also missing the most relevant information about the test failure: how long it was expected to take.

Ideally it would say "Timeout of 3000 ms exceeded" or something similar. This way, you instantly know what caused the cancellation (the timeout attribute) without guessing, and you know how long it took. This is in the same spirit as every assertion message.

jnm2:
Also, the test result should not be `ResultState.Cancelled` but `Failed`.

Places where test messages are set to `Cancelled by user`:
https://github.com/nunit/nunit/blob/3.6/src/NUnitFramework/framework/Internal/Results/TestResult.cs#L496-L499
https://github.com/nunit/nunit/blob/3.6/src/NUnitFramework/framework/Internal/Results/TestResult.cs#L533-L536
https://github.com/nunit/nunit/blob/3.6/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs#L400
https://github.com/nunit/nunit/blob/3.6/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs#L165
https://github.com/nunit/nunit/blob/3.6/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs#L393

CharliePoole:
The framework does exactly what you say it should do and has tests to prove it! How are you running? Can you provide a repo?

jnm2:
> The framework does exactly what you say it should do and has tests to prove it! How are you running? Can you provide a repo?

Running via VSTest, both in the VS2017 IDE and via the Cake build script in a powershell window. Maybe it's an adapter issue then? I thought I found sufficient evidence in the framework repo to the contrary.

Here's a repro: https://github.com/jnm2/misc-codesamples/blob/master/Bug%20reports/NUnit/Test%20NUnit%20timeout%20message.zip

Just open a powershell window and run `.\build.ps1 -t test`.

CharliePoole:
Closing as I believe it was fixed by PR #2106. Feel free to re-open if you see evidence that it was not fixed.

PR #2106, opened 2017-04-03
Handle test timeout without running tests on their own thread

### LINKED ISSUE #1363: Make Timeouts work without running test on its own thread
PR #1357 reduces the number of supplementary threads created by TestWorkers. One place where we still create a thread is if there is a timeout. For some users who set a default timeout, this may mean every test case.

In order to cancel a running test that is using the actual worker thread, we need to force that worker to stop. Then we either need a way to restart it on a different thread or dispose of it and create a new worker. One general way to do this could be to allow TestWorkers to be added and removed as the tests are running, without affecting the tests themselves except possibly for the one test being cancelled. That's the approach I will try initially.

CharliePoole:
FIxes #1363

This fixes a long-standing issue around the handling of timeouts in NUnit.

Old approach:
* Test methods with a timeout were run on a separate thread. 
* A join with a time limit specified was used to synchronize the completion of the test.
* If the test timed out, the test thread was killed and a timeout failure reported.

Issues:
* Old code was very complex
* The WorkItem was involved in determining the result of the test, which is not the normal division of labor in NUnit
* It was impossible to have a Timeout on any method in a SingleThreaded fixture.

New Approach:
* Timeouts are now handled in the Command hierarchy rather than the WorkItem.
* A TimeoutCommand is added whenever a timeout is specified for the test
* A TImer is used to cancel the test if it exceeds the time limit.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
@@ -31,8 +31,9 @@ namespace NUnit.Framework.Internal.Commands
     using Interfaces;
 
     /// <summary>
-    /// SetUpTearDownCommand runs SetUp methods for a suite,
-    /// runs the test and then runs TearDown methods.
+    /// TimeoutCommand creates a timer in order to cancel
+    /// a test if it exceeds a specified time and adjusts
+    /// the test result if it did time out.
     /// </summary>
     public class TimeoutCommand : BeforeAndAfterTestCommand
     {
@@ -40,14 +41,14 @@ namespace NUnit.Framework.Internal.Commands
         private bool _commandTimedOut = false;
         
         /// <summary>
-        /// Initializes a new instance of the <see cref="SetUpTearDownCommand"/> class.
+        /// Initializes a new instance of the <see cref="TimeoutCommand"/> class.
         /// </summary>
         /// <param name="innerCommand">The inner command</param>
         /// <param name="timeout">Timeout value</param>
         public TimeoutCommand(TestCommand innerCommand, int timeout)
             : base(innerCommand)
         {
-            Guard.ArgumentValid(innerCommand.Test is TestMethod, "SetUpTearDownCommand may only apply to a TestMethod", "innerCommand");
+            Guard.ArgumentValid(innerCommand.Test is TestMethod, "TimeoutCommand may only apply to a TestMethod", "innerCommand");
             Guard.ArgumentValid(timeout > 0, "Timeout value must be greater than zero", "timeout");
 
             BeforeTest = (context) =>
--- a/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
@@ -1,0 +1,83 @@
+﻿// ***********************************************************************
+// Copyright (c) 2017 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace NUnit.Framework.Internal.Commands
+{
+    using Execution;
+    using Interfaces;
+
+    /// <summary>
+    /// SetUpTearDownCommand runs SetUp methods for a suite,
+    /// runs the test and then runs TearDown methods.
+    /// </summary>
+    public class TimeoutCommand : BeforeAndAfterTestCommand
+    {
+        Timer _commandTimer;
+        private bool _commandTimedOut = false;
+        
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SetUpTearDownCommand"/> class.
+        /// </summary>
+        /// <param name="innerCommand">The inner command</param>
+        /// <param name="timeout">Timeout value</param>
+        public TimeoutCommand(TestCommand innerCommand, int timeout)
+            : base(innerCommand)
+        {
+            Guard.ArgumentValid(innerCommand.Test is TestMethod, "SetUpTearDownCommand may only apply to a TestMethod", "innerCommand");
+            Guard.ArgumentValid(timeout > 0, "Timeout value must be greater than zero", "timeout");
+
+            BeforeTest = (context) =>
+            {
+                var testThread = Thread.CurrentThread;
+
+                // Create a timer to cancel the current thread
+                _commandTimer = new Timer(
+                    (o) =>
+                    {
+                        _commandTimedOut = true;
+                        testThread.Abort();
+                        // No join here, since the thread doesn't really terminate
+                    },
+                    null,
+                    timeout,
+                    Timeout.Infinite);
+            };
+
+            AfterTest = (context) =>
+            {
+                _commandTimer.Dispose();
+
+                // If the timer cancelled the current thread, change the result
+                if (_commandTimedOut)
+                {
+                    context.CurrentResult.SetResult(ResultState.Failure,
+                        string.Format("Test exceeded Timeout value of {0}ms", timeout));
+                }
+            };
+        }
+    }
+}
--- a/src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
@@ -35,7 +35,6 @@ namespace NUnit.Framework.Internal.Execution
     {
         private static Logger log = InternalTrace.GetLogger("TestWorker");
 
-        private WorkItemQueue _readyQueue;
         private Thread _workerThread;
 
         private int _workItemCount = 0;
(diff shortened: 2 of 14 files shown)


src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -205,149 +205,49 @@ public virtual string Name
...
-
-                log.Debug("Changing result from {0} to Timeout Failure", Result.ResultState);
-
-                Result.SetResult(ResultState.Failure,
-                    string.Format("Test exceeded Timeout value of {0}ms", timeout));
-
-                WorkItemComplete();
-            }
-        }
+#if DEBUG
```
ChrisMaddock: Is there a reason this logging is `#if DEBUG`? That doesn't seem to be something we normally do for `log.Debug`?

src/NUnitFramework/tests/Attributes/TimeoutTests.cs
```diff
@@ -37,9 +37,9 @@ namespace NUnit.Framework.Attributes
     public class TimeoutTests : ThreadingTests
     {
         [Test, Timeout(500)]
-        public void TestWithTimeoutRunsOnSeparateThread()
+        public void TestWithTimeoutRunsOnSameThread()//SeparateThread()
```
ChrisMaddock: I guess this comment should be gone?

src/NUnitFramework/tests/Attributes/SingleThreadedFixtureTests.cs
```diff
@@ -40,10 +40,10 @@ public void TestRunsOnSameThreadAsParent()
             Assert.That(Thread.CurrentThread, Is.EqualTo(ParentThread));
         }
 
-        [Test]
-        public void TestWithTimeoutIsInvalid()
+        [Test, Timeout(100)]
+        public void TestWithTimeoutIsValid()
         {
-            CheckTestIsInvalid<SingleThreadedFixture_TestWithTimeout>("TimeoutAttribute may not be specified");
+            Assert.That(Thread.CurrentThread, Is.EqualTo(ParentThread));
```
ChrisMaddock: This is now the same as `TestWithTimeoutRunsOnSameThread()` below - do we need both?

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
```diff
@@ -581,9 +581,12 @@ public void CanAccessParallelScope()
...
         {
-            Assert.That(_fixtureContext.TestWorker, Is.Not.Null);
-            Assert.That(_setupContext.TestWorker, Is.SameAs(_fixtureContext.TestWorker));
-            Assert.That(TestExecutionContext.CurrentContext.TestWorker, Is.SameAs(_setupContext.TestWorker));
+            if (TestExecutionContext.CurrentContext.Dispatcher is ParallelWorkItemDispatcher)
+            {
+                Assert.That(_fixtureContext.TestWorker, Is.Not.Null);
+                Assert.That(_setupContext.TestWorker, Is.SameAs(_fixtureContext.TestWorker));
+                Assert.That(TestExecutionContext.CurrentContext.TestWorker, Is.SameAs(_setupContext.TestWorker));
+            }
```
ChrisMaddock: Can you clarify this change? Am I right that this will mean this test will now be empty, if we ever run the tests in serial? I'm a little confused as to what's changed that requires this, sorry?

src/NUnitFramework/tests/Attributes/TimeoutTests.cs
```diff
@@ -120,19 +121,41 @@ public void TestTimeOutElapsed()
...
+            [Test, Timeout(50), RequiresThread]
+            public void TestTimesOutUsingRequiresThread()
+            {
+                while (true) ;
+            }
+
+            [Test, Timeout(50), Apartment(ApartmentState.STA)]
+            public void TestTimesOutInSTA()
+            {
+                while (true) ;
```
ChrisMaddock: Suggestion: could these instead throw `Assert.Fail` after 100ms or something? Slightly concerned that the result of this breaking is the test suite hangs - which would prevent a merge, but wouldn't be so easy to debug!

ChrisMaddock: Oh, just seen that these are all Explicit! Should they be? I assume that's because of the timing concern?

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            AfterTest = (context) =>
+            {
+                _commandTimer.Dispose();
+
+                // If the timer cancelled the current thread, change the result
+                if (_commandTimedOut)
+                {
+                    context.CurrentResult.SetResult(ResultState.Failure,
+                        string.Format("Test exceeded Timeout value of {0}ms", timeout));
```
ChrisMaddock: Suggestion: Could we add a test for this failure message, in case it ever gets caught up with "Cancelled by user" or something similar again?

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
```diff
@@ -66,24 +65,29 @@ public class TestWorker
...
         public TestWorker(WorkItemQueue queue, string name, ApartmentState apartmentState)
         {
-            _readyQueue = queue;
-
-            _workerThread = new Thread(new ThreadStart(TestWorkerThreadProc));
-            _workerThread.Name = name;
-            _workerThread.SetApartmentState(apartmentState);
+            WorkQueue = queue;
+            Name = name;
+            Apartment = apartmentState;
```
ChrisMaddock: Can you run through the reasoning behind the changes to this file, for the ignorant? This is the only one I didn't follow. 😄

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                var testThread = Thread.CurrentThread;
+
+                // Create a timer to cancel the current thread
+                _commandTimer = new Timer(
+                    (o) =>
+                    {
+                        _commandTimedOut = true;
+                        testThread.Abort();
+                        // No join here, since the thread doesn't really terminate
```
ChrisMaddock: Oh wait - probably not - I'd not yet got to the bit where this was caught later on!

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace NUnit.Framework.Internal.Commands
+{
+    using Execution;
+    using Interfaces;
+
+    /// <summary>
+    /// SetUpTearDownCommand runs SetUp methods for a suite,
+    /// runs the test and then runs TearDown methods.
+    /// </summary>
```
CharliePoole: Right!

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -0,0 +1,83 @@
...
+    public class TimeoutCommand : BeforeAndAfterTestCommand
+    {
+        Timer _commandTimer;
+        private bool _commandTimedOut = false;
+        
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SetUpTearDownCommand"/> class.
```
CharliePoole: Yes.

src/NUnitFramework/tests/Attributes/TimeoutTests.cs
CharliePoole: Yes

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
CharliePoole: I think you're probably right... If we needed it before we still need it now. It's the test we are trying to cancel and we have no control over what the user may have in that code. I'll investigate further.

I'll address the later bit in a general comment.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
CharliePoole: The message isn't new. Issue #2020 must be some kind of race situation. We'll see if the new code alleviates it, but I wasn't trying to fix that.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
CharliePoole: No particular reason... I may have intended this as temporary... I'll check.

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
CharliePoole: I originally had substantial changes in this file but later found a simpler way to do things. The remaining changes are just cleanup like use of properties... which I left in place. It could be backed out entirely, but it seems to read better this way.

src/NUnitFramework/tests/Attributes/SingleThreadedFixtureTests.cs
CharliePoole: Right - I'll drop one.

CharliePoole:
I'll explain some things about timeout and cancel here. This is not something that has changed at all, so it's not visible in the diffs. However, I see from some comments that this background is needed to understand the code.

1. We can cancel a test for two reasons: user cancelled the run or test timed out.

2. In both cases, the currently running test needs to be aborted.

3. Once the test is cancelled, our own code does Thread.ResetAbort to stop the abort process.

4. If we timed out, our code knows that because it initiated it and can handle the change in result correctly.

5. If the user requested to cancel the run, we know that too because the info is saved in the test context. We then stop executing.

src/NUnitFramework/tests/Attributes/SingleThreadedFixtureTests.cs
CharliePoole: Correction - they are not the same because one is in a SingleThreaded fixture and tests that it is now legal to have a timeout in such a fixture. It wasn't before.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
CharliePoole: FYI there are already tests for the message.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
CharliePoole: ThreadUtility.Kill introduced some instability - intermittent hangs. I went back to `Thread.Abort()`.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
ChrisMaddock: Ah ok. Sounds like #2020 still needs to be investigated separately then!

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
ChrisMaddock: Ah ok! I was looking for some significance in why the thread creation needed to be moved. If it's just a minor optimisation, that makes sense. 😄

src/NUnitFramework/tests/Attributes/SingleThreadedFixtureTests.cs
ChrisMaddock: Fair enough!

ChrisMaddock:
Thanks for explaining - all looks good to me.

PR #2112, opened 2017-04-05
Allow failure on osx

CharliePoole:
We haven't had a successful run for two days. Even before that, I found myself restarting the osx build several times before it somehow, magically worked. 

I'm for moving osx to the allowed failures, at least for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ os:
 matrix:
   allow_failures:
     - mono: latest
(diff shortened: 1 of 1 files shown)


rprouse:
I'm good with this, the majority of OSX failures have been unrelated and causes me to doubt the validity of our Travis builds. Should we even be building on OSX if we are ignoring the results?

PR #2125, opened 2017-04-13
Show correct actual value for chained property constraints

### LINKED ISSUE #2121: Chained PropertyConstraint constraints report incorrect ActualValue
When a constraint includes multiple `.Property()` calls, the actual value portion of any error message indicates the value of the first property rather than the value being compared.

In the top portion of the following example, the `ActualValue` property is set to `3`. However, the `ActualValue` property in the bottom portion is set to the anonymous type in the property `Foo`, resulting in less useful messages in the various test runners.

```CSharp
var inputObject = new { Foo = new { Bar = "Baz" } };

// Single property
var result = ((IResolveConstraint)Has
    .Length.EqualTo(3))
    .Resolve()
    .ApplyTo(inputObject.Foo.Bar);

Console.WriteLine("Expected: " + result.Description);
Console.WriteLine("Actual: " + result.ActualValue);

Console.WriteLine();

// Nested property
result = ((IResolveConstraint)Has
    .Property("Foo").Property("Bar").Length.EqualTo(3))
    .Resolve()
    .ApplyTo(inputObject);

Console.WriteLine("Expected: " + result.Description);
Console.WriteLine("Actual: " + result.ActualValue);
```

This issue was introduced November 17, 2012 as part of a [refactoring effort](https://github.com/nunit/nunit/commit/29bac2be2e8e44206fce00da75ab2ad6ceb593f0). The previous functionality could be restored by returning the `ActualValue` supplied by the `BaseConstraint` instead of the value of the property.

CharliePoole:
Fixes #2121 

Let's discuss before merging. See my own inline comments to the code.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
@@ -70,7 +70,8 @@ namespace NUnit.Framework.Constraints
                 throw new ArgumentException(string.Format("Property {0} was not found", name), "name");
 
             propValue = property.GetValue(actual, null);
-            return new ConstraintResult(this, propValue, BaseConstraint.ApplyTo(propValue).IsSuccess);
+            var baseResult = BaseConstraint.ApplyTo(propValue);
+            return new ConstraintResult(this, baseResult.ActualValue, baseResult.IsSuccess);
         }
 
         /// <summary>
--- a/src/NUnitFramework/tests/Constraints/PropertyTests.cs
+++ b/src/NUnitFramework/tests/Constraints/PropertyTests.cs
@@ -52,7 +52,7 @@ namespace NUnit.Framework.Constraints
         }
     }
 
-    public class PropertyTest : ConstraintTestBase
+    public class PropertyTests : ConstraintTestBase
     {
         [SetUp]
         public void SetUp()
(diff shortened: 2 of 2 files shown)


src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
```diff
@@ -70,7 +70,8 @@ public override ConstraintResult ApplyTo<TActual>(TActual actual)
                 throw new ArgumentException(string.Format("Property {0} was not found", name), "name");
 
             propValue = property.GetValue(actual, null);
-            return new ConstraintResult(this, propValue, BaseConstraint.ApplyTo(propValue).IsSuccess);
+            var baseResult = BaseConstraint.ApplyTo(propValue);
+            return new ConstraintResult(this, baseResult.ActualValue, baseResult.IsSuccess);
```
CharliePoole: This is a minimal fix. A maximal fix, which we could consider as a separate issue, would be for `PropertyConstraint` to use a specially derived type of `ConstraintResult` in order to better tailor its own messages.

src/NUnitFramework/tests/Constraints/PropertyTests.cs
```diff
@@ -89,5 +89,54 @@ public void PropertyEqualToValueWithTolerance()
...
+            Assert.That(inputObject, Has.Property("Foo").Property("Bar").EqualTo("Baz"));
+            Assert.That(inputObject, Has.Property("Foo").Property("Bar").Property("Length").EqualTo(3));
+            Assert.That(inputObject, Has.Property("Foo").With.Property("Bar").EqualTo("Baz"));
+            Assert.That(inputObject, Has.Property("Foo").With.Property("Bar").With.Property("Length").EqualTo(3));
+
+            // Failure message
+            var c = ((IResolveConstraint)Has.Property("Foo").Property("Bar").Length.EqualTo(5)).Resolve();
+            var r = c.ApplyTo(inputObject);
+            Assert.That(r.Status, Is.EqualTo(ConstraintStatus.Failure));
+            Assert.That(r.Description, Is.EqualTo("property Foo property Bar property Length equal to 5"));
```
CharliePoole: I don't really like this description very much. Suggestions?

src/NUnitFramework/tests/Constraints/PropertyTests.cs
```diff
@@ -89,5 +89,54 @@ public void PropertyEqualToValueWithTolerance()
...
+            // Combine the tests
+            Assert.That(inputObject, Has.Property("Foo").And.Property("Bar").EqualTo("Baz"));
+            Assert.That(inputObject, Has.Property("Foo").And.Property("Bar").With.Length.EqualTo(3));
+
+            // Failure message
+            var c = ((IResolveConstraint)Has.Property("Foo").And.Property("Bar").With.Length.EqualTo(5)).Resolve();
+            var r = c.ApplyTo(inputObject);
+            Assert.That(r.Status, Is.EqualTo(ConstraintStatus.Failure));
+            Assert.That(r.Description, Is.EqualTo("property Foo and property Bar property Length equal to 5"));
+            Assert.That(r.ActualValue, Is.EqualTo(inputObject));
```
CharliePoole: Ideally, I'd like the ActualValue to be 3 here, as it is in the chained example. However, that will involve changing how the AndConstraint does its messages and I thought that was a bit out of scope. 

In general, where we can have compound constraints of any kind, it's possible for the failure to occur at any point. Ideally, the message should tell us exactly where it happened - e.g. on which branch of an And.

rprouse: I agree that this is a bit out of scope here. This could turn into a pretty deep rabbit hole if we let it 😄

src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
rprouse: I'm in favor of the minimal fix here. When I first read the chain of constraints in the issue, it took me awhile to parse it. To me, even though it is a valid construct, it isn't easy to understand, so that would lower my priority on fixing it. If you create overly complex asserts, you should expect error messages of the same complexity 😄 I tend to break things down into multiple asserts for that reason.

PR #2126, opened 2017-04-16
Implement tolerance for comparison constraints

### LINKED ISSUE #2095: Extend use of tolerance to ComparisonConstraints
The WithIn Option is available for Is.EqualTo. That's fine, but it is missing for the other ones:
`double x = ...;
Assert.That(x, Is.GreaterThan(1.0).WithIn(1.0e-12));`

The interpretation is:
  `if (x <= 1.0 - 1.0e-12) Assert.Fail(...);`
It is expected to have this for:
GreaterThan
GreaterThanOrEqualTo
LessThan
LessThanOrEqualTo
AtLeast
AtMost

CharliePoole:
Fixes #2095 

This required a completely separate implementation from the use of tolerance in equality comparisons. I limited the feature set to what appeared to me to be useful in this context.

* Linear and Percent tolerances are supported
* Only numeric values can use tolerances

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Tolerance.cs
+++ b/src/NUnitFramework/framework/Constraints/Tolerance.cs
@@ -34,9 +34,6 @@ namespace NUnit.Framework.Constraints
     /// </summary>
     public class Tolerance
     {
-        private readonly ToleranceMode mode;
-        private readonly object amount;
-
         private const string ModeMustFollowTolerance = "Tolerance amount must be specified before setting mode";
         private const string MultipleToleranceModes = "Tried to use multiple tolerance modes at the same time";
         private const string NumericToleranceRequired = "A numeric tolerance is required";
@@ -72,17 +69,14 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         private Tolerance(object amount, ToleranceMode mode)
         {
-            this.amount = amount;
-            this.mode = mode;
+            Amount = amount;
+            Mode = mode;
         }
 
         /// <summary>
         /// Gets the <see cref="ToleranceMode"/> for the current Tolerance
         /// </summary>
-        public ToleranceMode Mode
-        {
-            get { return this.mode; }
-        }
+        public ToleranceMode Mode { get; private set; }
         
 
         /// <summary>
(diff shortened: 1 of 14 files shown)


src/NUnitFramework/framework/Constraints/Tolerance.cs
```diff
@@ -72,17 +69,14 @@ public static Tolerance Exact
...
         }
 
         /// <summary>
         /// Gets the <see cref="ToleranceMode"/> for the current Tolerance
         /// </summary>
-        public ToleranceMode Mode
-        {
-            get { return this.mode; }
-        }
+        public ToleranceMode Mode { get; private set; }
```
jnm2: Could be `public ToleranceMode Mode { get; }` now that we support C# 6.

src/NUnitFramework/framework/Constraints/Tolerance.cs
```diff
@@ -91,22 +85,19 @@ public ToleranceMode Mode
...
 
         /// <summary>
-        /// Gets the value of the current Tolerance instance.
+        /// Gets the magnitude of the current Tolerance instance.
         /// </summary>
-        public object Value
-        {
-            get { return amount; }
-        }
+        public object Amount { get; private set; }
```
jnm2: Could be `public object Amount { get; }` now that we support C# 6.

src/NUnitFramework/tests/Constraints/GreaterThanOrEqualConstraintTests.cs
```diff
@@ -4,6 +4,7 @@
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
 // "Software"), to deal in the Software without restriction, including
+
```
jnm2: Typo

src/NUnitFramework/tests/Constraints/GreaterThanOrEqualConstraintTests.cs
```diff
@@ -16,7 +17,7 @@
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
 // NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
-// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTIONg
```
jnm2: Typo

src/NUnitFramework/framework/Constraints/Tolerance.cs
CharliePoole: You're right. I'm not used to being able to do this yet.

CharliePoole:
Yes... I like this approach better than the original. At some point I might refactor to use Range in the equality test as well.

PR #2127, opened 2017-04-16
Add DefaultFloatingPointToleranceAttribute

### LINKED ISSUE #1918: Changing DefaultFloatingPointTolerance breaks tests running in parallel
GlobalSettings.DefaultFloatingPointTolerance is a static. When it is changed for a single test, other tests may be affected. Even if the test resets it to the original value in TearDown, any tests that run in parallel with the test making the change are affected.

We should deprecate the GlobalSettings class (this is its sole member) and provide some other way to change the default tolerance. As it stands, the default tolerance should only be set at the command-line or in a global SetUpFixture.

CharliePoole:
Fixes #1918 

While it would be possible to add more functionality to the new attribute, I limited it to doing no more than what the GlobalSetting was already doing. I'm not a big fan of default settings. :smile:

I made GlobalSetting.DefaultFloatingPointTolerance obsolete but didn't cause the compile to fail. I'm not sure if that's the best choice though - might be better to force the user to change the code.


ChrisMaddock:
This looks good to me.

Regarding obsolete - I think Obsolete is the right choice here, rather than a breaking change. The previous Global Setting still works in certain non-parallel scenarios, so marking depracated seems suitable.

We should follow up with removing deprecated things in a few releases time, however.

rprouse:
I actually disagree with making the global setting obsolete in this case. I think it is the right case when we plan to remove things in the future, but you have effectively removed it now by making it non-functional. People will upgrade and existing tests will break unless they notice the warning and in my experience, many developers don't notice warnings. I expect that will lead to additional issues entered that we will need to answer.

If we want to make it obsolete, then I think we need to make it continue to work for now. Personally, I think we should just remove it, make sure it is called out in the release notes as a breaking change and force users to fix their code which they would need to do anyway with the current detached implementation.

ChrisMaddock:
Sorry - I agree. Charlie's approached this differently to how I did it, and I hadn't spotted the difference.

I'd kept the property plugged in, but allowed the attribute to override it. And then obsoleted the property.

CharliePoole:
So what are we agreeing to? Remove GlobalSettings entirely?

ChrisMaddock:
I don't mind either. Plugging it back in and obsoleting, or removing entirely.

rprouse:
I am okay with either approach too, just not the half way approach. If I had to pick, remove and get the pain over with.

CharliePoole:
OK, I'll remove. Justification is that it's a built-in bug if you use any parallelism at all.

Issue #1830, opened 2016-10-10
Add --labels switch changes to nunilite and nunitlite tests


oznetmaster:
As per @CharliePoole, I am opening this issue to make sure that the --label changes (issue24 nunit/nunit-console#78) made to the console are also made to nunitlite and its tests.

CharliePoole:
That's what issue #1829, on which you just commented, is about. :-)

oznetmaster:
No, it is not.  One is about --params, the other is about --labels. No relationship at all!

CharliePoole:
Sorry... my mistake.

CharliePoole:
I corrected the reference in your description, which was not pointing to the --labels issue but to PR #78 in this repository. Use `nunit/REPONAME#nn` to refer to issues and PR that are in another repo.

oznetmaster:
I will implement this as soon as the nunit-console PR is merged, and my pending PR for @filename is merged, since it also modifies the same files as need to be modified for this.

CharliePoole:
@oznetmaster We're still waiting for comments on that console PR about --labels, regarding a possible change to what we show for each option. It would be great to hear from you and @rprouse, as well as others!

oznetmaster:
I am actually indifferent to the decision.  As long as the desired information is available by selecting the correct set of options, the rest is syntactic sugaring.

oznetmaster:
Any progress on getting the console version of this merged?

CharliePoole:
This is issue nunit/nunit-console#24 and PR nunit/nunit-console#124, which are still being worked on.

PR #2129, opened 2017-04-20
Fixed issue with sending extra new lines when writing to listeners

### LINKED ISSUE #1819: TestContext.Progress.Write writes new line
Hi,

The TestContext.Progress.Write method now works exactly the same as TestContext.Progress.WriteLine because it writes a newline character at the end.

MoFtZ:
Fixes #1819

The issue with having a string being emitted as a single character per line is because of the base implementation of TextWriter.Write(char[]).

Refer to the .NET Reference Source Code:
http://referencesource.microsoft.com/#mscorlib/system/io/textwriter.cs

Each character of the array is emitted one at a time.

The proposed solution in this PR is to override all the Write and WriteLine methods so that we do not rely on the base implementation.

### DIFF
--- a/src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
+++ b/src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
@@ -1,0 +1,336 @@
+﻿// ***********************************************************************
+// Copyright (c) 2007 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using NUnit.Framework.Interfaces;
+
+namespace NUnit.Framework.Internal
+{
+    [TestFixture]
+    public class EventListenerTextWriterTests
+    {
+        private static readonly string NL = Environment.NewLine;
+
+        TestListenerHelper TestResult;
+        TextWriter TestWriter;
+
+        [SetUp]
+        public void SetUp()
+        {
+            TestResult = new TestListenerHelper();
+            TestExecutionContext.CurrentContext.Listener = TestResult;
+
+            TestWriter = TextWriter.Synchronized(TestContext.Error);
+        }
+
+        [Test]
+        public void TestWriteStringArgArray()
+        {
+            var format = "{0} {1} {2} {3}";
+            var arg = new object[] { "Hello", 4, 2, "World" };
+
+            TestWriter.Write(format, arg);
+            TestWriter.WriteLine(format, arg);
+
+            var expected = "Hello 4 2 World";
+            Assert.That(TestResult.Outputs.Count, Is.EqualTo(2));
+            Assert.That(TestResult.Outputs[0], Is.EqualTo(expected));
+            Assert.That(TestResult.Outputs[1], Is.EqualTo(expected + NL));
+        }
+
+        [Test]
+        public void TestWriteStringArg1()
+        {
+            var format = "{0:dd MMM yyyy}";
+            var arg0 = new DateTime(2017, 4, 20);
+
+            TestWriter.Write(format, arg0);
+            TestWriter.WriteLine(format, arg0);
+
+            var expected = "20 Apr 2017";
+            Assert.That(TestResult.Outputs.Count, Is.EqualTo(2));
+            Assert.That(TestResult.Outputs[0], Is.EqualTo(expected));
+            Assert.That(TestResult.Outputs[1], Is.EqualTo(expected + NL));
+        }
+
+        [Test]
+        public void TestWriteStringArg2()
+        {
+            var format = "{0:00.00} {1}";
+            var arg0 = 5;
+            var arg1 = "@";
+
+            TestWriter.Write(format, arg0, arg1);
+            TestWriter.WriteLine(format, arg0, arg1);
+
+            var expected = "05.00 @";
+            Assert.That(TestResult.Outputs.Count, Is.EqualTo(2));
+            Assert.That(TestResult.Outputs[0], Is.EqualTo(expected));
+            Assert.That(TestResult.Outputs[1], Is.EqualTo(expected + NL));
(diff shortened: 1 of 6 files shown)


CharliePoole:
Thanks for this! It looks like a good approach to me but I'd feel better if there were a few tests that failed before the change and pass now.

MoFtZ:
hi Charlie, wasnt sure how you wanted to handle this, so I've just pushed both commits here.

Let me know if you want me to split them into two separate pull requests.

rprouse:
@MoFtZ your tests show that your code works correctly and your code looks fine to me, but I think that what @CharliePoole is getting at is that there are no tests that demonstrate the original issue was fixed. One of your tests may cover it, but the original issue is so sparse that we can't tell which tests passed and failed. You also pushed both commits at the same time, so the first one did not build, so we cannot see which of your tests passed or failed before your fix.

That isn't really an issue with your PR, we just need confidence that the issue is actually fixed, so one of us will need to pull down your initial commit and test it manually, or, if you have time, do the same for us and show us the failing tests.

MoFtZ:
Raised a new PR just for the unit tests, as #2135 

This is the same commit as in this PR, but will allow the continuous integration to report on the test results.

src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
```diff
@@ -0,0 +1,336 @@
...
+using NUnit.Framework.Interfaces;
+
+namespace NUnit.Framework.Internal
+{
+    [TestFixture]
+    public class EventListenerTextWriterTests
+    {
+        private static readonly string NL = Environment.NewLine;
+
+        TestListenerHelper TestResult;
```
CharliePoole: I found this naming confusing. `TestResult`, of course, is an internal `Type` in NUnit. To add to the confusion, it even has a property `Output`. I suggest you call it some sort of "listener" since that's what it is.

src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
```diff
@@ -0,0 +1,336 @@
...
+        TextWriter TestWriter;
+
+        [SetUp]
+        public void SetUp()
+        {
+            TestResult = new TestListenerHelper();
+            TestExecutionContext.CurrentContext.Listener = TestResult;
```
CharliePoole: I like the use of the listener class to capture output. We do that a lot. I'm not in love with plugging it into the execution context, when that is not necessary. Why not just call your listener class directly?

If you are trying to verify that the writes hook up properly to what is in the context, that's the job of other tests, not this one. :smile:

There are a few places in our existing tests where we deliberately munge the context in some way, but only if we have to and can't think of a way to avoid it. The execution context is part of the superstructure that runs tests, so tests shouldn't mess with it lightly - even tests of NUnit.

MoFtZ: The scenario is that I'm trying to test the EventListenerTextWriter class - that is where the newline issue occurs.

The constructor for this class is EventListenerTextWriter(string streamName, TextWriter defaultWriter). It will then communicate with a listener by calling TestExecutionContext.CurrentContext.Listener.OnTestEvent().

So,  I cannot call the listener directly because it is hidden inside the class. How would you suggest I proceed?

src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
MoFtZ: Cool, easy to do.

src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
CharliePoole: Ah! I see the problem. It's not with your tests but with the fact that EventListenerTextWriter gets the current context internally. I guess that makes it one of those cases where you are forced to mess with the context.

It's still a problem however, because all  events go through the Listener of the currentcontext. Your helper class ignores all the events except for the one you testing, which could lead to problems.

I would give the helper class constructor an ITestListener argument and pass it the current listener when you construct it. Internally, Each event can forward to the original listener.

src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
```diff
@@ -270,64 +280,71 @@ public void TestWriteCharBufferSubstring()
...
+                    DefaultListener.TestFinished(result);
             }
 
             void ITestListener.TestOutput(TestOutput output)
             {
                 Assert.IsNotNull(output);
                 Outputs.Add(output.Text);
+
+                if (DefaultListener != null)
+                    DefaultListener.TestOutput(output);
```
MoFtZ: Due to the original behaviour of WriteLine(char[]) sending one char at a time, sending Environment.NewLine will actually send \r and \n separately.

This causes an issue in the NUnit Visual Studio Test Adapter (see 
https://github.com/nunit/nunit3-vs-adapter/blob/749594c656b01df1495918fb07d87d3fca2ed981/src/NUnitTestAdapter/NUnitEventListener.cs#L187)

As noted in the source code, an empty string will cause SendMessage to throw an exception. Unfortunately, so does sending just \r. 

Tested under NUnit 3 Test Adapter v3.7.0.0 with VS2017.

Is this a known issue?

CharliePoole: No, that would require a new adapter issue to resolve. We definitely don't want to worry about it here because it relates to how the output is used by the runner not with what we produce in the  framework. 

As a thought experiment, imagine that some listener passes messages on to a system that blows up whenever it sees an exclamation point! That would, of course, be a bug in the system in question, just as VS throwing on an empty message seems like a bug to me. In the case where that downstream system can't be fixed the adapter/runner would have to adjust the messages to remove any exclamation points, but other runners would probably still want to see and display them.

You refer to the "original behavior." Do you think this will remain as a problem when this PR is merged?

src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
```diff
@@ -25,24 +25,34 @@
...
 
-            TestWriter = TextWriter.Synchronized(TestContext.Error);
+            ListenerWriter = TextWriter.Synchronized(new EventListenerTextWriter(STREAM_NAME, TextWriter.Null));
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
+            // Restore the original listener
+            TestExecutionContext.CurrentContext.Listener = ListenerResult.DefaultListener;
```
CharliePoole: Of course, the context is about to be destroyed anyway, but I think this is good practice.

CharliePoole:
For some reason, the AppVeyor run is not showing up here as completed: https://ci.appveyor.com/project/CharliePoole/nunit/build/3.7.0-ci-03863-pr-2129

The portable and netstandard builds are passing but you have 12 errors in each of the others. In addition, for some reason, the text output is showing up twice.

You'll note that the output still shows one character per line in some cases. That's because we run the tests through NUnitLite. We'll have to decide whether to fix this as a part of this PR or separately.

src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
MoFtZ: Once this PR is merged, it will still be possible, but less likely. For example, if the user calls Write(char) explicitly with just \r.

MoFtZ:
The AppVeyor issue is my fault. I actually pushed to the branch twice. Can you manually trigger it again? They should pass on all the builds.

NUnitLite also passes the unit tests, but the outfitted text (using TextUI) adds it's own new lines, just like Console Runner.

CharliePoole:
I kicked off the build again.

The NUnitLite issues are expected, since you didn't change it. Normally we would fix NUnitLite in the same PR as the framework, since they are part of the same solution. However, if the output isn't too bad, it may be more efficient to fix #1830 first, bringing nunitlite in line with the console runner.

Note that what you see in the ci build is NUnitLite output.

CharliePoole:
The output looks good and the tests pass. The only issue is that the output from your new test fixture is appearing twice. Are you sending it twice?

MoFtZ:
Yes, the output is being sent twice. Each test makes a call to Write and then WriteLine. It should appear on one line, but NUnitLite (currently) will always append a newline after each Write.

CharliePoole:
I'm inclined to merge. @rprouse what do you think?

CharliePoole:
I'm going ahead and merging without a second review. We seem to have too few @nunit/framework-team and @nunit/core-team members doing reviews.

PR #2130, opened 2017-04-23
Resolve differences with console runner wrt argument files

### LINKED ISSUE #1877: Resolve differences between NUnit Console and NUnitLite implementations of @filename
We (well actually I) mistakenly merged the nunitlite changes thinking that the differences between the console runner and nunitlite implementations had already been resolved between @NikolayPianikov and @oznetmaster 

We need to decide first what features to include / exclude then change one or both of them so that users do not get confused and features we have not decided to support do not become supported by default.

Eventually, we should also resolve code differences for the common features but that can await the point where we create a shared codebase for conmmand-line option processing.

I'm putting this in the discussion pipeline so that the NUnit team can work out what we want to support. Everyone else is welcome to participate in the discussion as well.

CharliePoole:
Fixes #1877 

**Discuss before merging**

This resolves differences in the implementation of argument files between nunitlite and the console runner. The following changes have been made.

1. The "alternate separator" feature has been removed

2. Comments are supported in argument files.

However...

@rprouse has suggested we remove support for argument files from NUnitLite, so we should probably resolve that before we merge this commit.

### DIFF
--- a/src/NUnitFramework/nunitlite/CommandLineOptions.cs
+++ b/src/NUnitFramework/nunitlite/CommandLineOptions.cs
@@ -27,6 +27,7 @@ using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 using NUnit.Options;
+using System.Text;
 
 namespace NUnit.Common
 {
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/nunitlite/CommandLineOptions.cs
```diff
@@ -103,64 +104,63 @@ internal IEnumerable<string> PreParse(IEnumerable<string> args)
...
+        }
+
+        // Get args from an included file
+        private static IEnumerable<string> GetArgsFromFile(string filename)
+        {
+            var sb = new StringBuilder();
+
+            foreach (var line in File.ReadAllLines(filename))
+            {
+                if (!string.IsNullOrEmpty(line) && line[0] != '#')
```
rprouse: Should we do a `line.Trim()` before these two checks so that lines with just whitespace are ignored along with indented comments?

CharliePoole: I'll do that and then merge. I think the same thing occurs in the nunit-console PR, so I'll fix it there as well.

PR #2131, opened 2017-04-24
Remove "Version 3" suffix from NUnitLite NuGet Package

ChrisMaddock:
Just spotted this today. The NUnit NuGet package was renamed "NUnit Version 3" -> "NUnit" with 3.0, the NUnitLite package should be consistent. Otherwise it creates weird display strings like:

`NUnitLite Version 3 3.6.1`

This is just the display name, so shouldn't have any side effects.


rprouse:
I'm good with this. Does anyone else have issues with the rename?

CharliePoole:
IIRC the "NUnitlite" package is the old 1.0 release. I'll be on my computer in an hour where I can check further.

CharliePoole:
This is therefore good to go.

PR #2135, opened 2017-04-25
Added unit tests for EventListenerTextWriter

MoFtZ:
Created a separate PR for the unit tests from #2129

This is the same commit as previously raised.

Each of the tests will track the number of times that TestOutput is called. They do this by calling Write and then WriteLine, so the expectation is that TestOutput will be called twice.

In the failure cases, eg. Write(char[]), TestOutput is called many times.


CharliePoole:
@MoFtZ  FYI, while we always want to have tests for a fix, it's not necessary to make a separate PR for them. Reviewers will download your changes and run locally if they can't tell what's happening by inspection. You can delete this one when you are done with it - I'll review the tests as part of the main PR, where they are also included.

PR #2136, opened 2017-04-25
Add .gitattributes entry so that .sh files checked out in Windows have Unix line endings

jnm2:
See https://github.com/nunit/docs/issues/163#issuecomment-297059889

This makes the build.sh file usable when you check out and develop on Windows but mount the repo in Linux to run tests.

This fixes the problem for new clones of the repo. Existing clones will keep their problematic CRLFs unless you delete `build.sh` and run `git checkout build.sh` to replace it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,3 +11,7 @@
 # Monodevelop on Linux uses CRLF for these files
 *.sln	text	eol=crlf
 packages/repositories.config	text	eol=crlf
+
+# Allows checking out and developing in Windows
+# while mounting and running tests in Linux
+*.sh text eol=lf


CharliePoole:
Not sure I understand the use case. Are you talking about using the working directory that you checked out under windows by sharing it with a Linux system? Normally, we do everything in one OS, so if we are building under Linux, we check out under Linux.

OTOH, I see no harm to the fix other than not being able to edit the shell file in notepad on windows any longer. I'm probably the only person who ever does that. :smile:

jnm2:
> Not sure I understand the use case. 

[Answeredd](https://github.com/nunit/docs/issues/163#issuecomment-297096451) under the other instance of your question. 😄 

> I see no harm to the fix other than not being able to edit the shell file in notepad on windows any longer.

Notepad++ is amazing. It doesn't get in your way, but it is so powerful you'll wonder why you stuck with Notepad for so long.

CharliePoole:
I'm OK with this. Comment about build file is optional.

rprouse:
Looks good to me too. We could also do a `git update-index --add --chmod=+x build.sh` to make the file executable when checked out on Linux, but that is a separate issue.

I've started playing with docker as per your instructions, so this will make life easier.

rprouse:
@jnm2 do you want to make the same change to the nunit-console project?

jnm2:
@rprouse Sure.

CharliePoole:
I see that my comment about the missing file was pending, so didn't get read. I completed the review so it would post. (I really don't like the "improvements" GitHub has made to the review process.)

I'm not sure if the `build` script gets used in Linux. It's intended as a counterpart to `build.cmd` so that we can force certain options as needed in particular local environments, without changing the ps1 or sh scripts that are provided by Cake.

jnm2:
@CharliePoole Sounds good, will do another PR off this branch.

PR #2139, opened 2017-04-27
Require filename with nunitlite-runner and disallow it with AutoRun

### LINKED ISSUE #344: NUnitLite could process multiple assemblies
Currently in the built-in runner in nunitlite, TextUI correctly captures multiple assemblies on the command line, but then only processes the first of them.

```
                // TODO: For now, ignore all but first assembly
                Assembly assembly = _assemblies[0];
```

Is there a reason that all assemblies provided could not be processed? It would be seem to be beneficial to be able to.

CharliePoole:
Fixes #344 

Hopefully, this resolves the question of what we want NUnitLite to support.

### DIFF
--- a/src/NUnitFramework/nunitlite-runner/Program.cs
+++ b/src/NUnitFramework/nunitlite-runner/Program.cs
@@ -43,7 +43,7 @@ namespace NUnitLite
         static int Main(string[] args)
         {
 #if PORTABLE
-            return new TextRunner().Execute(new ColorConsoleWriter(), Console.In, new NUnitLiteOptions(args));
+            return new TextRunner().Execute(new ColorConsoleWriter(), Console.In, args);
 #else
             return new TextRunner().Execute(args);
 #endif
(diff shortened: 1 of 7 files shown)


CharliePoole:
@rprouse @nunit/framework-team I need somebody to review this as there are several other issues that depend on it and can't go to PRs until this is merged.

Ideally, several folks should take a look, since this implements #344, where we discussed the overall future direction for nunitlite as distinct from the console runner.

CharliePoole:
This is blocking other work by both me and @MoFtZ so I'm going to merge. There are some other PRs pending related to NUnitLite, where I can correct any problems that come up from this one. 

I should know better than to base one PR off of another PR. 😞

oznetmaster:
I still find this unnecessary.

My CF NUnitLite build runner has always supported multiple test assemblies on the command line, and it has not had any negative affect that I can see.

In fact, it is a widely used capability.

I often start a command line with three or four test assemblies in it.

I understand not allowing any test assembly name in the AutoRun mode, but I am totally unclear as to why it is necessary to remove multiple assemblies from the nunitlite runner :(

CharliePoole:
As you can see here and on the issue, we did discuss the options. This change matches the concensus of what we wanted to support in NUnitLite, versus the full featured console runner. Bear in mind that we currently don't even distribute nunitlite runner - it's merely used by our own tests for platforms that don't have a console.

oznetmaster:
Hopefully, someday, there will be an engine which allows true remote execution support over TCP/IP without remoting.

Then, this will be a mute issue :(

PR #2140, opened 2017-04-28
Added build-mono-docker.ps1

jnm2:
Via https://github.com/nunit/docs/issues/163#issuecomment-297122590

### DIFF
--- a/build-mono-docker.ps1
+++ b/build-mono-docker.ps1
@@ -4,4 +4,4 @@ trap [Management.Automation.CommandNotFoundException] {
   continue;
 }
 
-docker run --rm -it -v ${pwd}:/nunit -w=/nunit mono:latest bash build.sh $args
+docker run --rm -it -v ${PSScriptRoot}:/nunit -w=/nunit mono:latest bash build.sh $args
--- a/build-mono-docker.ps1
+++ b/build-mono-docker.ps1
@@ -1,0 +1,7 @@
+trap [Management.Automation.CommandNotFoundException] {
+  Write-Error 'Docker cannot be found. Make sure it is installed and added to the path.'
+  Start-Process -FilePath 'https://docs.docker.com/docker-for-windows/install/'
+  continue;
+}
+
+docker run --rm -it -v ${pwd}:/nunit -w=/nunit mono:latest bash build.sh $args
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -5,6 +5,7 @@ The source code is kept in two GitHub repositories at http://github.com/nunit/nu
 http://github.com/nunit/nunit-console.
 
 There are two ways to build NUnit: using the solution file in an IDE or through the build script.
+See also [Building and testing for Linux on a Windows machine](#building-and-testing-for-linux-on-a-windows-machine).
 
 ## Solution Build
 
@@ -67,3 +68,19 @@ For a full list of tasks, run `build.cmd -ShowDescription`.
     you have to be very careful that the build is up to date before packaging.
 
  2. For additional targets, refer to the build.cake script itself.
+
+### Building and testing for Linux on a Windows machine
+
+Most of the time, it's not necessary to build or run tests on platforms other than your primary
+platform. The continuous integration which runs on every PR is enough to catch any problems.
+
+Once in a while you may find it desirable to be primarily developing the repository on a Windows
+machine but to run Linux tests on the same set of files while you edit them in Windows.
+One convenient way to do this is to pass the same arguments to [build-mono-docker.ps1](.\build-mono-docker.ps1) that you would pass to build.ps1. It requires
+[Docker](https://docs.docker.com/docker-for-windows/install/) to be installed.
+
+For example, to build and test everything: `.\build-mono-docker.ps1 -t test`
+
+This will run a temporary container using the latest
+[Mono image](https://hub.docker.com/r/library/mono/), mounting the repo inside the container
+and executing the [build.sh](build.sh) Cake bootstrapper with the arguments you specify.


ChrisMaddock:
Would you maybe mind adding a note to BUILDING.md on what this files there for? It's fairly self-explanatory in isolation - but we have a lot of build scripts!

build-mono-docker.ps1
```diff
@@ -0,0 +1,7 @@
+trap [Management.Automation.CommandNotFoundException] {
+  Write-Error 'Docker cannot be found. Make sure it is installed and added to the path.'
+  Start-Process -FilePath 'https://docs.docker.com/docker-for-windows/install/'
+  continue;
+}
+
+docker run --rm -it -v ${pwd}:/nunit -w=/nunit mono bash build.sh $args
```
rprouse: Not sure about this, so putting it out for discussion. Should we switch `mono` to `mono:latest` so it is explicit?

rprouse:
It was mentioned elsewhere, but I had to delete `build.sh` from my source directory and check it out again for the line endings to switch to `LF`. Other than that, this worked well, thanks. Every time I play with Docker, I like it more 😄

build-mono-docker.ps1
jnm2: Yes, sounds best.

jnm2:
How is that? /cc @CharliePoole

CharliePoole:
I'm OK with merging this script as a start... I see some issues...

1. For anyone who has not used Docker before (like me) it's a fairly steep learning curve just running the script dealing with each error message that comes up. Things I had to do to run it...
  * Install Docker for Windows
  * Read the docs and figure out how to set it up and use it
  * Enable VtX in my bios
I don't think your instructions should explain all this stuff, but it should point people to the info - or just tell them not to try to use it without first gaining some mastery of Docker.

2. I'd rather see the instructions in a separate file from Building.md. As we all know, I would prefer a text file, but you guys are all in love with markdown. 😄 

3. Personally, I would prefer to simply run bash in a linux+mono environment and then issue issue the appropriate build commands myself rather than going through the build-mono-docker command each time. I can type Exit when I'm done.

4. The script name seems too long to me.

5. We should emulate docker itself and have a .cmd script as well as a .ps1 script.

jnm2:
1. Docker was extremely plug and play for me. Besides double-clicking the installer there was only one step which I left out, that you have to go to Settings > Shared Drives and tick the box next to whichever drive the repo is on. I left it out of BUILDING.md because 1) I figured it would just be noise to most people that use Docker, unrelated to NUnit and 2) the error message if you leave this step out is so simple and descriptive you can't go wrong.
VT-x makes sense. I already had it on so I assume it gave a similarly helpful error message? Anyone who has run Android emulators or VirtualBox or Hyper-V has already set this bit.

    So if I mention that and I mention the VT-x bit, that covers all known setup contingencies so far? At least that way people will know this could involve a reboot to the BIOS.

 2. Yeah, why are we so in love with markdown? Weird. 😁 I have no problem with this, but what would I name it? (Also, I dislike adding files starting with `b` because I already have to type <kbd>b</kbd> <kbd>Tab</kbd> <kbd>Tab</kbd> <kbd>Tab</kbd> <kbd>Tab</kbd> to autocomplete `build.cmd` in the terminal. See point 4.)

 3. Not a bad idea. @rprouse What do you think of this?

 4. Suggestions for the name? `build` is because it matches the other build scripts, `mono` is so that you know why you'd want to use the file, and  `docker` is because there are other ways to build for Mono on Linux from Windows than Docker which we may add, like WSL. Right now all it takes is typing `build-m` <kbd>Tab</kbd>- does that alleviate the problem? Or see point 3- maybe we will end up not having a build script that mirrors the arguments and behavior of `build.ps1`. (@rprouse?)

 5. So just copy `build.cmd`? `@powershell ./build-mono-docker.ps1 %CAKE_ARGS% %*`

CharliePoole:
@jnm2 I like to think about the potential audiences for a script or doc. In this case...

For hard-core windows-only folks, separate the doc, don't care about the command.  

For newbies to docker and maybe Linux as well, make a simple command that does everything. I think that's what you have, if the doc gets beefed up.

For those who are already familiar with the tech, or for newbies who decide to learn more, just tell them what containers are available. This could be at the end of the doc, for example. Some folks will want to vary the mono build and possibly the distro.

1. Between us, I think we listed most of the points to add. One more: I spent a silly amount of time figuring out where the docker settings are!

2. DOCKER-BUILD ?

3. I think that would be a different script for a different audience or they would just type a docker command.

4. Not shorter but docker-mono-build is a more natural word order for me and works better with intellisense.

5. Or just use the identical docker command and don't display the site.

rprouse:
I approved changes based on my original requests. I think we have a few of @CharliePoole's requests to deal with now, then we can merge.

**Info on setting up Docker** - I don't think we should add too much here, maybe just point to the documentation because it is fairly new for most developers. We don't include information on setting up Visual Studio, Mono or .NET Core.

**Location of build info** - I am fine with `BUILDING.md` or in a new file. I would lean towards `BUILDING.md` because we already have so many files in the root, but I am fine with `DOCKER-BUILD.md` too.

**Using Markdown** - I just like it because it displays nicely in GitHub. I often explore files like `BUILDING.md` on the web in GitHub rather than locally, but maybe that is just me 😄 

**Name of the build file** - `docker-mono-build` 👍 

**Interactive session** - I like both ideas of just dropping into an interactive bash session, or passing the CAKE arguments.  For larger changes I work in stages running individual build tasks to save time.

CharliePoole:
The CAKE_ARGS were intended for local use. Running in a container is "sort of" local, but also a bit different. Also, the args I used on my old windows system dealt with a problem of Cake on 32-bit Windows systems. I don't know what may be needed under Linux, but it's likely to be different. Until we know more, I wouldn't pass them.

CharliePoole:
I think this is going to develop over time so I'm OK with it as is or with any of the changes we discussed. I don't think changing our build scripts in the future is an issue.

@jnm2 I think you should consider who your primary audience is. Either satisfy those who know a lot about docker or those who know nothing and just want to run a quick test. Make what changes make sense to you and then merge.

Here's one point of view from an old-timer. 😄 When I'm working at the command line, I expect to see text. I don't expect windows or web pages to pop up. I don't like surprises. Ideally, our build command would have a help option displaying text, but I never added it. That said, I'm pretty sure I'm not your typical audience.

CharliePoole:
It may be good to mention in the doc that we are using mono:latest under Debian 7 (Wheezy) for those who care.

jnm2:
Oops- I did not get around to responding to everyone's feedback here. I guess anyone can feel free to improve the setup in their own PR. =)

rprouse:
No worries, it is good enough and it works.

PR #2142, opened 2017-04-30
Make mono:latest the standard (and only) mono version for ci

CharliePoole:
We have been using 4.2.4, which is way behind latest (4.8). We run latest but allow it to fail. In fact, the "latest" tag for mono is the latest stable version build, not the latest development build as I once thought. I think we should use mono:latest for all our builds, just as @jnm2 did for the docker build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,11 @@ language: csharp
 sudo: false
 mono:
   - latest
-  - 4.2.4
 os:
   - linux
   - osx
 matrix:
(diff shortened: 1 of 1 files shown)


rprouse:
Looks good, waiting for CI.

rprouse:
We should do the same for the `nunit-console` project, but before we do, we need to figure out why `mono:latest` is failing for that project. Unfortunately, looks like a mono bug to me.

Issue #2119, opened 2017-04-08
Add Linux .NET Standard build to CI


ChrisMaddock:
I have this up and running on a fork, there's just some tidying up to do, and a couple of failing tests to investigate.

ChrisMaddock:
I'm getting very stuck with this, so am taking my name off - if someone else fancies a go.

Installing dotnet core on a separate Travis machine seemed the best method - seems like it's necessary to have separate mono and dotnet-core machines at the moment on Travis.

I got restore and build up and running - I was then attempting to run the tests with the NUnitLite runner. I was running:

`dotnet nunitlite-runner.exe nunit.framework.tests.dll`

Although this and the nunitlite.tests executable both seem to be returning 2, and providing no output, which I'm stuggling to debug with no local Unix machine. 🤔 

The above seems to do what I'd expect on Windows - I'm not sure why there's no output in Travis - and for that matter, I'm not even entirely sure exactly what's being run! (I find it hard to believe that 2 tests failed, in both test suites I was running...)

Here's my fork if anyone's interested: https://github.com/ChrisMaddock/nunit/commits/travis_dotnet

Happy for someone else to take this over, try a different approach, or just suggest some further ideas - I'm not sure where to go next, right now!

jp2masa:
I think that the problem is that the projects aren't being built, as you can see [here](https://travis-ci.org/ChrisMaddock/nunit/builds/229524812#L620).
I pulled your branch to my fork and tried to make it work, but no success. I also noticed that the script is trying to build csproj, but they are .NET Standard PCLs, right? So I think that the project files are the *.project.json ones.
I also found [this bug](https://github.com/dotnet/core/blob/master/cli/known-issues.md#dotnet-commands-in-the-root-of-the-file-system-fails), which may be related to the build failure.

rprouse:
@jp2masa we use PCL CSPROJ files that target .NET Standard, so building those is actually correct. It is an undocumented hack that was suggested so that we could continue to build in 2015 once 2017 was installed. At some point we will want to move to requiring VS2017 and use the new CSPROJ format, but we want to be as inclusive as possible for potential developers.

ChrisMaddock:
@jp2masa - I'm not sure about that bug, the files aren't in the root directory. I never figured out what to make of all the null reference exceptions however - but if the builds failing, it's off that cake doesn't pick that up. Maybe worth adding a File.Exists() check just to see what's actually created?

[...]

PR #2150, opened 2017-05-06
Add portable-slow-tests to Cake file

ChrisMaddock:
These are missing, spotted while working on #2119. Committing separately, as I'm at least temporarily abandoning #2119!


rprouse:
Looks good. I caught this while porting from Portable to .NET Standard, but that work isn't ready ready. 👍

CharliePoole:
How were our tests passing before?

ChrisMaddock:
MSBuild takes care of building project dependencies - so this was being built when the framework tests dll was being built. I don't know if that's the same for XBuild - portable isn't built there of course.

Given we build everything else separately though - it seemed to make sense to correct this one. I imagine it's just a copy and paste error.

PR #2151, opened 2017-05-07
Ensure that parallel cases under a non-parallel fixture only run with one another

### LINKED ISSUE #2143: Non-parallel fixture with parallel children runs in parallel with other fixtures
I'm breaking out this issue from the much larger complex of issues represented by #1474 because (1) it will be easier to solve and (2) it will require creating part of the solution for the larger issue.

Quoting from the above issue...

> Let's say you have a [NonParallelizable] test fixture. It's executing (by itself, of course) from the non-parallelizable queue. Now let's say that some of the test methods in that fixture are marked [Parallelizable]. As they are encountered, they will be added to the Parallelizable queue.

> Eventually, the dispatcher will exhaust the non-parallelizable queue and switech execution to the parallelizable queue and those test cases will get run. However, there may be other fixtures and their test cases on that queue and they will be run even though the original non-parallelizable fixture is still active.

[In fact, the problem can arise when __any__ type of non-parallel test has parallelizable children - for example, a non-parallelizable SetUpFixture with parallelizable TestFixtures under it. In this issue, however, let's deal with the simplest case of a non-parallelizable TestFixture containing parallelizable tests.]

What we would like to see happen here is this:
1. The fixture runs it's one-time setup in isolation, with no other fixtures running.
2. All the test cases run in isolation from other fixtures, even if they run in parallel.
3. The fixture runs it's one-time teardown, still in isolation.

In reality, only (1) and (3) happen correctly. That's because our notion of parallelism deals with what tests are __running__ and a fixture is only considered to be running when it's own code (not that of it's children) is executing.

In fact, we need to define a notion of a fixture being __active__ even if not running. It should be considered as active from the start of its one-time setup to the end of its one-time teardown. Our definition of parallelism has to switch from which tests are __running__ in parallel to which tests are __active__ in parallel. This is a big change, which is why I'm trying to start with a very limited test case.

As a first cut at the problem, I'm thinking of modifying how the parallel dispatcher is used.

Current Approach:
* The top level (composite) workitem, representing the assembly, is used to initialize the dispatcher.
* As each composite workitem is executed, it creates it's own children and dispatches them.
* The one-time teardown for a composite item is scheduled when all it's children have finished.
* Execution is complete when the top-level item finishes.

New Approach:
* ~~Prior to creating and dispatching work items, all fixtures will be located and identified as either parallelizable or non-parallelizable.~~
* ~~Non-parallel fixtures will be processed separately, one at a time, by adding the single fixture as the top-level item and re-starting the dispatcher shifts for each one of them.~~
* ~~Execution is complete when all queues, both fixture and work item, are empty.~~

>**Update:** The above approach complicated the code too much. Trying a different approach. See following comment.

Note that the above description __only__ solves the specific problem with test fixtures and test cases. For non-parallel setup fixtures, as described in #1474, multiple levels of recursion would be needed since SetUpFixtures may have other SetUpFixtures as well as TestFixtures as children. If the recursion looks simple, I'll try to fix that here as well. Otherwise a separate PR is needed.

CharliePoole:
Fixes #2143

### DIFF
--- a/src/NUnitFramework/tests/Attributes/TestOrderAttributeTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TestOrderAttributeTests.cs
@@ -1,11 +1,9 @@
-﻿using System;
-using NUnit.Framework.Internal;
+﻿using NUnit.Framework.Internal;
 using NUnit.Framework.Internal.Execution;
 using NUnit.TestData;
-using NUnit.TestData.TestCaseSourceAttributeFixture;
 using NUnit.TestUtilities;
 
-namespace NUnit.Framework.Tests.Attributes
+namespace NUnit.Framework.Attributes
 {
     [TestFixture]
     public class TestOrderAttributeTests
--- a/src/NUnitFramework/tests/Internal/Execution/ParallelWorkItemDispatcherTests.cs
+++ b/src/NUnitFramework/tests/Internal/Execution/ParallelWorkItemDispatcherTests.cs
@@ -1,0 +1,79 @@
+﻿// ***********************************************************************
+// Copyright (c) 2017 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+#if PARALLEL
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NUnit.Framework.Internal.Execution
+{
+    public class ParallelWorkItemDispatcherTests
+    {
+        private const int LEVEL_OF_PARALLELISM = 4;
+        private ParallelWorkItemDispatcher _dispatcher;
+
+        [SetUp]
+        public void CreateDispatcher()
+        {
+            _dispatcher = new ParallelWorkItemDispatcher(LEVEL_OF_PARALLELISM);
+        }
+
+        [Test]
+        public void ConstructorCreatesQueues()
+        {
+            var queues = new List<WorkItemQueue>(_dispatcher.Queues);
+
+            Assert.That(List.Map(queues).Property("Name"), 
+                Is.EquivalentTo(new[] { "ParallelQueue", "NonParallelQueue", "ParallelSTAQueue", "NonParallelSTAQueue" }));
+        }
+
+        [Test]
+        public void ConstructorCreatesShifts()
+        {
+            var shifts = new List<WorkShift>(_dispatcher.Shifts);
+
+            Assert.That(shifts, Is.Unique);
+            Assert.That(shifts.Count(), Is.EqualTo(3));
+
+            // Parallel Shift
+            Assert.That(shifts[0].Name, Is.EqualTo("Parallel"));
+            Assert.That(shifts[0].Queues.Count, Is.EqualTo(2));
+            Assert.That(shifts[0].Queues[0].Name, Is.EqualTo("ParallelQueue"));
+            Assert.That(shifts[0].Queues[1].Name, Is.EqualTo("ParallelSTAQueue"));
+            Assert.That(shifts[0].Workers.Count, Is.EqualTo(LEVEL_OF_PARALLELISM + 1));
+
+            // NonParallel Shift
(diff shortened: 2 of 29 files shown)


CharliePoole:
Not sure what happened here. The commits up to this point were merged through a different PR and GitHub says I closed it. It won't even let me reopen it, since there are no new commits. I'll add some more work and see if it re-opens.

CharliePoole:
OK, I pushed a new commit and was able to get this to re-open.

In the new code, whenever any non-parallel fixture is started, a completely new set of queues is reserved for the children of that fixture. This prevents anything else executing while the fixture is active. Only tests that are children of that fixture will end up running. After that test is done, we go back to the original set of queues and continue working.

I'm not totally happy with the organization, but I don't see any better options...

1. The dispatcher has to control allocating and removing the queues, since it has to be done for all queues at once.
2. The change actually happens inside the WorkItemQueue, so that code is called by the dispatcher.
3. The TestWorker triggers the creation of the new queues, calling the dispatcher, because it can only happen after the work item is dequeued and before it is executed.
4. The WorkShift triggers  the removal of the queues when no longer needed, because that's where the end of a shift is detected.

I also wish I had a better way to test this than watching it in the debugger and examining the trace. Suggestions would be very welcome.

CharliePoole:
I'm much happier with the code after the second commit. All control is now in ParallelWorkItemDispatcher, which uses private methods to save and restore the queues. WorkShift, WorkItemQueue and TestWorker are passive participants, which is as I think it should be.

@nunit/framework-team This is ready for review. If I can figure out how to do it, I'll add some tests. 🤢

CharliePoole:
It's really annoying that an initialized but unused field doesn't give a warning in VS but does so on Travis.

jnm2:
@CharliePoole
> It's really annoying that an initialized but unused field doesn't give a warning in VS but does so on Travis.

The only explanation for this that comes to mind is conditional compilation or debug- vs release-specific settings in the csproj?

CharliePoole:
Nope. I do both locally and the same difference exists between Appveyor and Travis.

tom-dudley:
Looks like the XBuild version of CS0413 doesn't match the 'spec' used by msbuild: [https://docs.microsoft.com/en-us/dotnet/articles/csharp/misc/cs0414](https://docs.microsoft.com/en-us/dotnet/articles/csharp/misc/cs0414) which only applies to constant values.

The XBuild version seems a bit saner to me..

CharliePoole:
@tom-dudley Well, it's good to know anyway. 😄

CharliePoole:
While waiting, I did some refactoring, which will help with the next bit of work. I'm adding it to this PR since nobody has reviewed it yet.

The new commit takes the creation of child work items completely away from the dispatcher and composite work item. All work items are now created before the dispatcher is invoked. This simplifies CompositeWorkItem by a small amount but, more important, it gives the dispatcher access to the full tree of work items in order to later implement dependencies between them.

Even so, I hope somebody will review this soon @nunit/framework-team !!!

CharliePoole:
BTW, we are unable to build on travis using Mono 5.0, so I turned the clock back to 4.8.1 and put latest in the allowed failures for now.

CharliePoole:
Turns out that "latest" is still 4.8.1 on OSX.

CharliePoole:
@nunit/framework-team @nunit/core-team 
Now the weekend is over, I hope one of you will review this. :smile:

src/NUnitFramework/tests/Attributes/TestOrderAttributeTests.cs
```diff
@@ -1,11 +1,9 @@
-﻿using System;
-using NUnit.Framework.Internal;
+﻿using NUnit.Framework.Internal;
 using NUnit.Framework.Internal.Execution;
 using NUnit.TestData;
-using NUnit.TestData.TestCaseSourceAttributeFixture;
 using NUnit.TestUtilities;
 
-namespace NUnit.Framework.Tests.Attributes
+namespace NUnit.Framework.Attributes
```
rprouse: Curious, why the change in namespace?

src/NUnitFramework/tests/Internal/Execution/ParallelWorkItemDispatcherTests.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            _dispatcher = new ParallelWorkItemDispatcher(LEVEL_OF_PARALLELISM);
+        }
+
+        [Test]
+        public void ConstructorCreatesQueues()
+        {
+            var queues = new List<WorkItemQueue>(_dispatcher.Queues);
+
+            Assert.That(List.Map(queues).Property("Name"),
```
rprouse: Minor, but assuming we have LINQ, these two lines could be simplified down to `_dispatcher.Queues.Select(q => q.Name)` and then it would be refactor safe.

src/NUnitFramework/tests/Internal/Execution/ParallelWorkItemDispatcherTests.cs
```diff
@@ -0,0 +1,79 @@
...
+                Is.EquivalentTo(new[] { "ParallelQueue", "NonParallelQueue", "ParallelSTAQueue", "NonParallelSTAQueue" }));
+        }
+
+        [Test]
+        public void ConstructorCreatesShifts()
+        {
+            var shifts = new List<WorkShift>(_dispatcher.Shifts);
+
+            Assert.That(shifts, Is.Unique);
+            Assert.That(shifts.Count(), Is.EqualTo(3));
```
rprouse: shifts is a List, why use Count() over Count? `Assert.That(shifts, Has.Count.EqualTo(3));`?

src/NUnitFramework/tests/Attributes/TestOrderAttributeTests.cs
CharliePoole: Most attribute tests are in NUnit.Framework.Attributes, with a few in NUnit.Framework.Tests.Attributes. I considered a mass change but didn't want to complicate this PR any further.

src/NUnitFramework/tests/Internal/Execution/ParallelWorkItemDispatcherTests.cs
CharliePoole: Good point... after years of not having it, I tend to forget about it.

src/NUnitFramework/tests/Internal/Execution/ParallelWorkItemDispatcherTests.cs
CharliePoole: Copy / paste :smile:

rprouse:
I've gone ahead and merged since nobody else has had a chance to code review.

Issue #1670, opened 2016-07-15
Attaching files to the test result


giang-hai-nguyen:
Hi guys,
I am new with NUnit and i am wondering that how can i add attachment file to output in test explorer.
In unit test, there is a method in TestContext class call AddResultFile that can do this.
is there anybody can help me out.

thanks,
giang

CharliePoole:
The method you mention is part of the TestContext class in Microsoft's test framework. NUnit also has a class of the same name, but it has no relation to the MsTest class and doesn't have such a method. There is no facility in NUnit to attach files to the output.

giang-hai-nguyen:
Thank you Charlie.
But will AddResultFile method add to NUnit in the future?
This method really helpful.

CharliePoole:
There is no plan for it at this time. If you would like to change this issue into a request for such a feature, we can do that and re-open it. 

If so, you should explain in more detail how it would work. How are files created, where are they stored, etc. Are you asking that NUnit actually create the files or simply retain a reference to them?

Of couse, we would have to accept the idea, prioritize it and schedule it for some release.

giang-hai-nguyen:
Thank you for your response Charlie.
It would be great if that method is added to NUnit.
it would work just like AddResultFile method of TestContext class in Microsoft's test framework.
something likes link the test result (file html test result or screenshot) to the Test Output in Test Explorer in VS. You can see the screenshot as below.

![screenshot](https://cloud.githubusercontent.com/assets/15973559/16868652/12d13336-4aa3-11e6-811a-542847eba03d.jpg)

CharliePoole:
Thanks for the idea. I'll reopen this.

To take it further, however, we would have to figure out how it works for NUnit itself, not merely the adapter. I'm marking this as an Idea, which is what we call something that is not well enough defined to consider as a future feature. Maybe other folks will chime in with suggestions.

Things we need to consider:
 * How would this look in the XML result
 * Can we package all this in some way that the result and the attachments are kept together?
 * What would the NUnit console do with it, if anything.
 * Would the NUnit gui display it in some way

fredgate:
This feature is a must have.
We are running our automated behavior tests with NUnit and Selenium. If a test fail, we love to attach a screenshot of the browser in the test result : this is a great help to fix the test.
To do this, we had to use Microsoft's test framework, but it is not as powerfull than NUnit (attribute at assembly or class level, generic test class...). So we look forward to this feature in NUnit.

[...]

PR #2152, opened 2017-05-07
Allow attaching files to TestResults

ChrisMaddock:
This is the framework functionality required for #1670. (Adding this PR to the milestone separately, as I want to track runner implementation with the issue.)

Specific design points:

- API is `TestContext.AddTestAttachment(string filepath, string description = null)` Are we happy with that?

- I forced all files to be absolute paths - based on the logic that it's the runner that decides where the test result XML will end up, I believe? This is the same approach MSTest takes.

What do people think? I've also added some line comments on things to check.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 #if PARALLEL
 using System.Collections.Concurrent;
 #endif
@@ -81,12 +82,7 @@ namespace NUnit.Framework.Internal
         private string _stackTrace;
 
         private readonly List<AssertionResult> _assertionResults = new List<AssertionResult>();
-
-        /// <summary>
-        /// TestAttachments attached to result.
-        /// Only exposed for testing.
-        /// </summary>
-        internal ICollection<TestAttachment> TestAttachments { get; } = new List<TestAttachment>();
+        private readonly List<TestAttachment> _testAttachments = new List<TestAttachment>();
 
 #if PARALLEL
         /// <summary>
@@ -194,8 +190,13 @@ namespace NUnit.Framework.Internal
         /// <param name="attachment">The TestAttachment object to attach</param>
         internal void AddTestAttachment(TestAttachment attachment)
         {
-            TestAttachments.Add(attachment);
+            _testAttachments.Add(attachment);
         }
+
+        /// <summary>
+        /// Gets the collection of files attached to the test
+        /// </summary>
+        public ICollection<TestAttachment> TestAttachments => new ReadOnlyCollection<TestAttachment>(_testAttachments);
 
         /// <summary>
         /// Gets the message associated with a test
@@ -283,7 +284,7 @@ namespace NUnit.Framework.Internal
                 InternalAssertCount = value;
             }
         }
-        
+
         /// <summary>
         /// Gets the number of test cases that failed
         /// when running the test and all its children.
@@ -413,8 +414,8 @@ namespace NUnit.Framework.Internal
             if (AssertionResults.Count > 0)
                 AddAssertionsElement(thisNode);
 
-            if (TestAttachments.Count > 0)
-                AddAttatchmentsElement(thisNode);
+            if (_testAttachments.Count > 0)
+                AddAttachmentsElement(thisNode);
 
             if (recursive && HasChildren)
                 foreach (TestResult child in Children)
@@ -719,20 +720,22 @@ namespace NUnit.Framework.Internal
         }
 
         /// <summary>
-        /// Adds a attachments element to a node and returns it.
+        /// Adds an attachments element to a node and returns it.
         /// </summary>
         /// <param name="targetNode">The target node.</param>
         /// <returns>The new attachments element.</returns>
-        private TNode AddAttatchmentsElement(TNode targetNode)
+        private TNode AddAttachmentsElement(TNode targetNode)
         {
             TNode attachmentsNode = targetNode.AddElement("attachments");
 
-            foreach (var attachment in TestAttachments)
+            foreach (var attachment in _testAttachments)
             {
-                var attachmentNode = attachmentsNode.AddElement("attachment", attachment.Filepath);
+                var attachmentNode = attachmentsNode.AddElement("attachment");
+
+                attachmentNode.AddElement("filePath", attachment.FilePath);
 
                 if (attachment.Description != null)
-                    attachmentNode.AddAttribute("description", attachment.Description);
+                    attachmentNode.AddElementWithCDATA("description", attachment.Description);
             }
 
             return attachmentsNode;
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -82,6 +82,12 @@ namespace NUnit.Framework.Internal
 
         private readonly List<AssertionResult> _assertionResults = new List<AssertionResult>();
 
+        /// <summary>
+        /// TestAttachments attached to result.
+        /// Only exposed for testing.
+        /// </summary>
+        internal ICollection<TestAttachment> TestAttachments { get; } = new List<TestAttachment>();
+
 #if PARALLEL
         /// <summary>
         /// ReaderWriterLock
@@ -183,9 +189,13 @@ namespace NUnit.Framework.Internal
         public DateTime EndTime { get; set; }
 
         /// <summary>
-        /// Gets or sets the list of test attachments
+        /// Adds a test attachment to the test result
         /// </summary>
-        internal ICollection<TestAttachment> TestAttachments { private get; set; }
+        /// <param name="attachment">The TestAttachment object to attach</param>
+        internal void AddTestAttachment(TestAttachment attachment)
+        {
+            TestAttachments.Add(attachment);
+        }
 
         /// <summary>
         /// Gets the message associated with a test
@@ -403,7 +413,7 @@ namespace NUnit.Framework.Internal
             if (AssertionResults.Count > 0)
                 AddAssertionsElement(thisNode);
 
-            if (TestAttachments?.Count > 0)
+            if (TestAttachments.Count > 0)
                 AddAttatchmentsElement(thisNode);
 
             if (recursive && HasChildren)
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -80,7 +80,7 @@ namespace NUnit.Framework.Internal
         private string _message;
         private string _stackTrace;
 
-        private List<AssertionResult> _assertionResults = new List<AssertionResult>();
+        private readonly List<AssertionResult> _assertionResults = new List<AssertionResult>();
 
 #if PARALLEL
         /// <summary>
@@ -183,6 +183,11 @@ namespace NUnit.Framework.Internal
         public DateTime EndTime { get; set; }
 
         /// <summary>
+        /// Gets or sets the list of test attachments
+        /// </summary>
+        internal ICollection<TestAttachment> TestAttachments { private get; set; }
+
+        /// <summary>
         /// Gets the message associated with a test
         /// failure or with not running the test
         /// </summary>
@@ -268,7 +273,7 @@ namespace NUnit.Framework.Internal
                 InternalAssertCount = value;
             }
         }
-
+        
         /// <summary>
         /// Gets the number of test cases that failed
         /// when running the test and all its children.
@@ -397,6 +402,9 @@ namespace NUnit.Framework.Internal
 
             if (AssertionResults.Count > 0)
                 AddAssertionsElement(thisNode);
+
+            if (TestAttachments?.Count > 0)
+                AddAttatchmentsElement(thisNode);
 
             if (recursive && HasChildren)
                 foreach (TestResult child in Children)
@@ -698,6 +706,26 @@ namespace NUnit.Framework.Internal
                 case AssertionStatus.Error:
                     return ResultState.Error;
             }
+        }
+
+        /// <summary>
+        /// Adds a attachments element to a node and returns it.
+        /// </summary>
+        /// <param name="targetNode">The target node.</param>
+        /// <returns>The new attachments element.</returns>
+        private TNode AddAttatchmentsElement(TNode targetNode)
+        {
+            TNode attachmentsNode = targetNode.AddElement("attachments");
+
+            foreach (var attachment in TestAttachments)
+            {
+                var attachmentNode = attachmentsNode.AddElement("attachment", attachment.Filepath);
+
+                if (attachment.Description != null)
+                    attachmentNode.AddAttribute("description", attachment.Description);
+            }
+
+            return attachmentsNode;
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Internal/TestAttachment.cs
+++ b/src/NUnitFramework/framework/Internal/TestAttachment.cs
@@ -1,0 +1,53 @@
+﻿// ***********************************************************************
+// Copyright (c) 2017 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+namespace NUnit.Framework.Internal
+{
+    /// <summary>
+    /// The TestAttachment class represents a file attached to a TestResult,
+    /// with an optional description.
+    /// </summary>
+    internal class TestAttachment
+    {
+        /// <summary>
+        /// Absolute filepath to attachment file
+        /// </summary>
+        public string Filepath { get; }
+
+        /// <summary>
+        /// Optional user specifed description of attachment
+        /// </summary>
+        public string Description { get; }
+
+        /// <summary>
+        /// Creates a TestAttachment class to represent a file attached to a test result.
+        /// </summary>
+        /// <param name="filepath">Absolute filepath to attachment file</param>
+        /// <param name="description">Optional user specifed description of attachment</param>
+        public TestAttachment(string filepath, string description)
+        {
+            Filepath = filepath;
+            Description = description;
+        }
+    }
+}
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -97,11 +97,6 @@ namespace NUnit.Framework.Internal
         /// </summary>
         private TestResult _currentResult;
 
-        /// <summary>
-        /// Files attached to the current test context
-        /// </summary>
-        private readonly IList<TestAttachment> _testAttachments = new List<TestAttachment>();
-
 #if !PORTABLE && !NETSTANDARD1_6
         /// <summary>
         /// The current Principal.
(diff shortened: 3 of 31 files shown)


src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -493,6 +502,28 @@ public void AddFormatter(ValueFormatterFactory formatterFactory)
...
+        /// <param name="description">Optional description of attachment</param>
+        public void AddTestAttachment(string filepath, string description = null)
+        {
+            Guard.ArgumentNotNull(filepath, nameof(filepath));
+            Guard.ArgumentValid(filepath.IndexOfAny(Path.GetInvalidPathChars()) == -1, 
+                $"Test attachment filepath contains invalid path characters. {filepath}", nameof(filepath));
+
+            if (!Path.IsPathRooted(filepath))
+                filepath = Path.Combine(TestContext.CurrentContext.WorkDirectory, filepath);
+
```
ChrisMaddock: Is this the correct way to build an absolute file path? I wanted to use `Path.GetFullPath()` - but it's not available in PCL.

src/NUnitFramework/tests/nunit.framework.tests-3.5.csproj
```diff
@@ -122,6 +123,7 @@
     <Compile Include="Internal\Results\TestResultOutputTests.cs" />
     <Compile Include="Internal\Results\TestResultSuccessTests.cs" />
     <Compile Include="Internal\Results\TestResultTests.cs" />
+    <Compile Include="Internal\Results\TestResultWarningTests.cs" />
```
ChrisMaddock: This is unrelated - just noticed this file was missing from all except the 2.0 build.

ChrisMaddock:
Regarding file location - MSTest also copies the file - but has a fully fledged directory structure to support that.

This currently just references whatever the user passes in - meaning the link will be broken if the results are transferred between machines. Do we want to go any further, and (e.g.) copy the files to a directory next to the result xml? I wasn't entirely sure of the advantage, given a) potential copy permission errors and b) still requiring the user to copy both files and results, if that was the case.

I'm tempted to think the simple option will be ok, until such a point as someone has a compelling argument to change it?

CharliePoole:
@ChrisMaddock What do you mean by "I want to track runner implementation with the issue?"

ChrisMaddock:
@CharliePoole - There's a number of people following issue #1670. Rather than close it, and leave them to track down the related runner issues, which need to be completed before they can use the functionality, I've converted it to an Epic, and will add issues for the different runners, once we've decided what to do for each.

The side effect of that, is that there's no dedicated issue for just the framework component of #1670 - so I added the PR to the Epic.

CharliePoole:
I'm glad you took this on as it's a needed feature. I do need a bit of general explanation before doing a detailed review.

You are adding an entry to TestExecutionContext for attachments and then copying them to the result. Is there any reason you don't want to simply add them to the result directly? This would be analogous to how we add text output to the result.

In the TestExecutionContext constructor, you are copying the attachments from each higher level to the lower level. Is there a specific reason for that? Won't that mean the attachments are duplicated all the way down the tree of results?

ChrisMaddock:
> You are adding an entry to TestExecutionContext for attachments and then copying them to the result. Is there any reason you don't want to simply add them to the result directly? This would be analogous to how we add text output to the result.

I hadn't realised the TestResult was created at the start of the test - I followed where StartTime was set, which was in WorkItem.WorkItemComplete(). This sounds like a much better idea.

> In the TestExecutionContext constructor, you are copying the attachments from each higher level to the lower level. Is there a specific reason for that? Won't that mean the attachments are duplicated all the way down the tree of results?

Nope - that's just me not looking into things properly. Sorry.

It's getting on a bit UK-time, will make the changes tomorrow. Thanks for those!

CharliePoole:
@ChrisMaddock Great! I'll wait till you update to do a detailed review.

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
```diff
@@ -907,6 +918,65 @@ public void SetAndRestoreValueFormatter()
...
+            Assert.That(() => context.AddTestAttachment("NotAFile.txt"), Throws.InstanceOf<FileNotFoundException>());
+        }
+
+        public TestExecutionContext CreateFakeContext()
+        {
+            var fakeResult = TestUtilities.Fakes.GetTestMethod(this, "FakeMethod").MakeTestResult();
+            var context = new TestExecutionContext { CurrentResult = fakeResult };
+            return context;
+        }
+
```
ChrisMaddock: `CreateFakeContext()` is pretty ugly. Am I missing a better way to create temporary context to test with? I didn't want to attach anything to the actual TestExecutionContext.

ChrisMaddock:
Initial changes made, and ready for review. Thanks Charlie.

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
CharliePoole: It's easy to create a context, harder to populate it with meaningful values. In this case, we needed a context with a valid result. The alternative is to use NSubstitute but I think you'll be doing a lot of setup if you do that.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -268,7 +283,7 @@ internal set
                 InternalAssertCount = value;
             }
         }
-
+
```
jnm2: I really hate to nitpick whitespace. I like these tools:
VS: https://marketplace.visualstudio.com/items?itemName=MadsKristensen.TrailingWhitespaceVisualizer
VS Code: https://marketplace.visualstudio.com/items?itemName=shardulm94.trailing-spaces with `"trailing-spaces.trimOnSave": true`

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -701,6 +719,26 @@ private ResultState AssertionStatusToResultState(AssertionStatus status)
         }
 
         /// <summary>
+        /// Adds a attachments element to a node and returns it.
+        /// </summary>
+        /// <param name="targetNode">The target node.</param>
+        /// <returns>The new attachments element.</returns>
+        private TNode AddAttatchmentsElement(TNode targetNode)
```
jnm2: Nit: should be `AddAttachmentsElement`

src/NUnitFramework/framework/Internal/TestAttachment.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// <summary>
+    /// The TestAttachment class represents a file attached to a TestResult,
+    /// with an optional description.
+    /// </summary>
+    internal class TestAttachment
+    {
+        /// <summary>
+        /// Absolute filepath to attachment file
+        /// </summary>
+        public string Filepath { get; }
```
jnm2: I've never seen this as one word. Can it be `FilePath`?

src/NUnitFramework/framework/Internal/TestAttachment.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// Optional user specifed description of attachment
+        /// </summary>
+        public string Description { get; }
+
+        /// <summary>
+        /// Creates a TestAttachment class to represent a file attached to a test result.
+        /// </summary>
+        /// <param name="filepath">Absolute filepath to attachment file</param>
+        /// <param name="description">Optional user specifed description of attachment</param>
+        public TestAttachment(string filepath, string description)
```
jnm2: Same; I've never seen this as one word. Can it be `filePath`?

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -493,6 +494,28 @@ public void AddFormatter(ValueFormatterFactory formatterFactory)
             CurrentValueFormatter = formatterFactory(CurrentValueFormatter);
         }
 
+        /// <summary>
+        /// Attach a file to the current test result
+        /// </summary>
+        /// <param name="filepath">Filepath to attachment. Can be absolute, or relative to working directory.</param>
+        /// <param name="description">Optional description of attachment</param>
+        public void AddTestAttachment(string filepath, string description = null)
```
jnm2: Same; I've never seen this as one word. Can it be filePath?

src/NUnitFramework/tests/Internal/Results/TestResultAttachmentTests.cs
```diff
@@ -0,0 +1,106 @@
...
+namespace NUnit.Framework.Internal.Results
+{
+    public class TestResultAttachmentTests
+    {
+        private TestResult _result;
+
+        [SetUp]
+        public void SetUp()
+        {
+            _result = TestUtilities.Fakes.GetTestMethod(this, "FakeMethod").MakeTestResult();
```
jnm2: `nameof(FakeMethod)` 😀

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
```diff
@@ -43,28 +44,38 @@ namespace NUnit.Framework.Internal
...
+        private string _tempFilePath;
+        private DateTime _fixtureCreateTime = DateTime.UtcNow;
+        private long _fixtureCreateTicks = Stopwatch.GetTimestamp();
         
         [OneTimeSetUp]
         public void OneTimeSetUp()
         {
             _fixtureContext = TestExecutionContext.CurrentContext;
             _fixtureResult = _fixtureContext.CurrentResult.ResultState;
+            _tempFilePath = Path.Combine(TestContext.CurrentContext.WorkDirectory, "NUnitTests.tmp");
```
jnm2: Replace `"NUnitTests.tmp"` with `TempFileName`?

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
```diff
@@ -907,6 +918,65 @@ public void SetAndRestoreValueFormatter()
...
+        public void NoneExistandFileThrowsFileNotFoundException()
+        {
+            var context = CreateFakeContext();
+            Assert.That(() => context.AddTestAttachment("NotAFile.txt"), Throws.InstanceOf<FileNotFoundException>());
+        }
+#endif
+
+        public TestExecutionContext CreateFakeContext()
+        {
+            var fakeResult = TestUtilities.Fakes.GetTestMethod(this, "FakeMethod").MakeTestResult();
```
jnm2: `nameof(FakeMethod)` 😀

src/NUnitFramework/framework/Internal/Results/TestResult.cs
ChrisMaddock: I never could spell that word! Thanks!

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
ChrisMaddock: Thanks. I had to do a bit of shuffling here to get things playing nice with PCL - must have missed that. 🙂

src/NUnitFramework/framework/Internal/TestAttachment.cs
ChrisMaddock: I Googled, and you're right. I'll change throughout - thanks!

ChrisMaddock:
Thanks @jnm2. I'll wait for comments from others, then fix everything in one go. 😄

src/NUnitFramework/framework/Internal/Results/TestResult.cs
rprouse: @jnm2, LOL, nice to see other developers as OCD as me 😄

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -701,6 +719,26 @@ private ResultState AssertionStatusToResultState(AssertionStatus status)
...
+        {
+            TNode attachmentsNode = targetNode.AddElement("attachments");
+
+            foreach (var attachment in TestAttachments)
+            {
+                var attachmentNode = attachmentsNode.AddElement("attachment", attachment.Filepath);
+
+                if (attachment.Description != null)
+                    attachmentNode.AddAttribute("description", attachment.Description);
```
rprouse: Description is a user supplied string that may be long or require CDATA encoding. Based on that, we should probably make it a child element rather than an attribute and CDATA encode it.

```C#
var attachmentNode = attachmentsNode.AddElement("attachment");
attachmentNode.AddElement("filename", attachment.Filepath);

if (attachment.Description != null)
    attachmentNode.AddElementWithCDATA("description", attachment.Description);
```

Tests will need to be changed accordingly.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
rprouse: Looks good to me, although I think `Path.GetFullPath()` will be available once I get the .NET Standard version working.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
ChrisMaddock: Sure. 

I did consider this - the current AddAttribute goes through an ExcapeXmlCharacters() - which should sufficiently encode any user string. If length is a concern for readability however, I can change it over. Will do that with the rest. 🙂

ChrisMaddock:
Thanks both. @CharliePoole - you said earlier you were planning to come back to this - so I'll hold on for your feedback, then deal with everything in one go.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
CharliePoole: OTOH, if VS doesn't really care...

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -80,7 +80,13 @@ public abstract class TestResult : ITestResult
...
         private string _stackTrace;
 
-        private List<AssertionResult> _assertionResults = new List<AssertionResult>();
+        private readonly List<AssertionResult> _assertionResults = new List<AssertionResult>();
+
+        /// <summary>
+        /// TestAttachments attached to result.
+        /// Only exposed for testing.
+        /// </summary>
+        internal ICollection<TestAttachment> TestAttachments { get; } = new List<TestAttachment>();
```
CharliePoole: Working my way down... I can't see why this would not be `private readonly List`. Maybe it will become clear as I go. 😄

src/NUnitFramework/framework/Internal/Results/TestResult.cs
CharliePoole: Spelling is not a nitpick. Just whitespace. 😄

src/NUnitFramework/framework/Internal/Results/TestResult.cs
CharliePoole: Amazing how things grow! I didn't imagine we would have a description at all at the start but if we have one I think @rprouse is right.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
CharliePoole: Actually `Path.GetFullPath()` would be incorrect. We promise to store all output files relative to our WorkDirectory, which is __not__ the current directory. The choice of name for the WorkDirectory, may have been a bad one, since it sounds a bit like the working directory, i.e. cwd, but I believe we are stuck with it.

A more basic question... why do we want to make this absolute at all? Why not just use what the user gives us?

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -493,6 +494,28 @@ public void AddFormatter(ValueFormatterFactory formatterFactory)
...
+            Guard.ArgumentNotNull(filepath, nameof(filepath));
+            Guard.ArgumentValid(filepath.IndexOfAny(Path.GetInvalidPathChars()) == -1, 
+                $"Test attachment filepath contains invalid path characters. {filepath}", nameof(filepath));
+
+            if (!Path.IsPathRooted(filepath))
+                filepath = Path.Combine(TestContext.CurrentContext.WorkDirectory, filepath);
+
+#if !PORTABLE
+            if (!File.Exists(filepath))
+                throw new FileNotFoundException("Test attachment filepath could not be found.", filepath);
```
CharliePoole: Why do we care that the file doesn't exist? Maybe the user asked to add the attachment before saving it. Maybe some third party software is expected to add it later. If we are not doing anything with the file, why not just record what the user tells us to record?

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -295,6 +295,16 @@ public static void AddFormatter(ValueFormatterFactory formatterFactory)
         }
 
         /// <summary>
+        /// Attach a file to the current test result
+        /// </summary>
+        /// <param name="filepath">Relative or absolute filepath to attachment</param>
+        /// <param name="description">Optional description of attachment</param>
+        public static void AddTestAttachment(string filepath, string description = null)
+        {
+            TestExecutionContext.CurrentContext.AddTestAttachment(filepath, description);
```
CharliePoole: Why the extra level of indirection? In other code, TestContext just accesses `TestExecutionContext.CurrentContext.CurrentResult`  directly.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
jnm2: Comment says "Only exposed for testing." but `readonly` sounds good.

CharliePoole: Ah! I didn't notice the comment. So this field is not the private backing field for a public property then? That seems a bit odd, since it leaves a part of the TestResult inaccessible to clients.

ChrisMaddock: You're right. I had been keeping this as an 'add-only' collection, but it would be more in-keeping with the rest of the TestResult class if this was exposed to the user in a ReadOnlyCollection. I'll expose this list and the TestAttachment type.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
ChrisMaddock: This was a point I went back and forth on. 🙂 My thinking was that adding a non-existant file is far more likely to occur through user-error, rather than intention, and you'd want to know about that ASAP. I imagine a common use case may be where people only refer to attachments when the test has failed - so broken paths may only be discovered 6 months after writing, when the test first fails, and it's too late.

The opposite arguments e.g. 'adding an attachment before saving it', mostly seem like things that could be managed by the user. (e.g. Don't add an attachment before you've created it!) That seems like fairly bad practice anyway, what if the creation of the attachment later went wrong, and the file couldn't be written? We could always add a 'don't check file existence' overload in the future if such cases arose.

Not that it matters, but MSTest checks if the File Exists. @rprouse @jnm2 - what do you guys think?

src/NUnitFramework/framework/TestContext.cs
ChrisMaddock: I can change that if this ways preferred. I get a little mixed up about logic should live between TestContext and TestExecutionContext - I'd followed how `AddFormatter()` was written.

This is also more possible now I'm no longer storing an intermediate list of attachments on TestExecutionContext. 😄

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
ChrisMaddock: Thanks for the explanation! 

This was something else I went back and forth on. I wanted to make it absolute, as I had previously thought:
- The attachments should always be relative the the results xml - as the file paths written there should be correct to the files current location.
- The runner controls where the results xml is written, meaning that the runner could theoretically move where the results ended up.

What I didn't realise is that WorkDirectory wasn't the working directory. 😄 So will any results xml file always end up in TestContext.WorkDirectory? Or is that a runner setting, could TestResults be written anywhere, like I'd first thought?

I did think full paths would be easier to resolve should the user move the result xml file around for any CI etc. Is there a disadvantage to writing full paths, if we already define that they're relative to WorkDirectory? I guess if they remain relative, and you were adding all your attachments to the same subdir, you could move the directory as a whole?

ChrisMaddock:
Thanks @CharliePoole! I've replied to the inline comments, the remaining ones...

> I would have tested this using a new class in the testdata-assembly. That would eliminate all worries about test context, execution context, etc. We would simply be looking at the result and it's fields.

I'll try it. 🙂 

> There is no test of TestContext
No - I'll move the tests currently in TestExecutionContext to use TestContext, once I move everything out of there.

> There are no new tests of TestResult, particularly to verify that it creates XML correctly.
Is this not covered by TestResultAttachmentTests?

Will hopefully make the changes this weekend - thanks all!

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
CharliePoole: I guess checking for existence makes sense in general.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
CharliePoole: The runner passes Work directory to the framework to be used in exactly this circumstance. The XML result file is saved there. 
In the case of the console runner, it can be set by a command-line argument.

ChrisMaddock: Ok, cool. I'll remove the stuff that meddles with paths in that case - if we have the concrete definition that paths will always be relative to WorkDirectory/results location. Thanks.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
jnm2: Typically, if you're going to do anything else with the file other than verify that it exists, the existence check is meaningless because it's race condition prone. To have meaningful handling you'd have to catch FileNotFoundException because it's the only atomic API.

> I imagine a common use case may be where people only refer to attachments when the test has failed - so broken paths may only be discovered 6 months after writing, when the test first fails, and it's too late.

But this is a great point and would be very helpful.

ChrisMaddock:
Review changes made - this is ready for a second look. 🙂 

Only thing to note, is that this contains `if #PORTABLE` symbols, so either this or #2174 will need adjusting, depending which is merged first.

oznetmaster:
I commented on the commit.  Probably the wrong place, again :(

PR #2153, opened 2017-05-08
NonParallelizable has no effect on TestFixtures

CharliePoole:
I discovered this working on #2143 and it seems important enough to create a PR right away. I'm labeling the PR to include in the the release.

Because of the order of checking in `ParallelWorkItemDispatcher.GetExecutionStrategy()`, a fixture with `[NonParallelizable]` is run in parallel if `ParallelScope.Fixtures` was used at the assembly level.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -23,8 +23,6 @@
 
 #if PARALLEL
 
-//#define NO_PARALLEL_CASES
-
 using System;
 using System.Collections.Generic;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


rprouse:
I assume there is no way to unit test this so it is checked in the future?

CharliePoole:
We really should have some unit tests of the innards of our dispatcher. This bug was a big surprise to me so I wanted to fix it right away but I'll add some tests as a part of #2143 so we are protected.

rprouse:
Thanks. If you are taking care of it in #2143, I will merge this.

CharliePoole:
Somehow, the change is not here. The only commit is for some dead code I removed. Clearly I did something wrong. What's in is OK, but it isn't the fix for the problem.

rprouse:
I was wondering how that code had an effect, but since there were no tests I took your word for it. Are you going to push to the branch?

CharliePoole:
I'll rename my local branch and do a new PR. I found the code in my index - I committed it but didn't push! As a side benefit, there are tests now.

PR #2154, opened 2017-05-09
Fix execution of non-parallel test fixtures

CharliePoole:
This is the code that was missing in #2153 plus tests.

### DIFF
--- a/src/NUnitFramework/tests/Internal/Execution/ParallelExecutionStrategyTests.cs
+++ b/src/NUnitFramework/tests/Internal/Execution/ParallelExecutionStrategyTests.cs
@@ -1,0 +1,95 @@
+﻿// ***********************************************************************
+// Copyright (c) 2017 Charlie Poole
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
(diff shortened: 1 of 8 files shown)


src/NUnitFramework/tests/Internal/Execution/ParallelExecutionStrategyTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        public void ParallelExecutionStrategy_TestCase(ParallelScope testScope, ParallelScope contextScope, string expectedStrategy)
+        {
+            _testMethod.Properties.Set(PropertyNames.ParallelScope, testScope);
+            _context.ParallelScope = contextScope;
+
+            WorkItem work = WorkItem.CreateWorkItem(_testMethod, TestFilter.Empty);
+            work.InitializeContext(_context);
+
+            // We use a string for expected because the ExecutionStrategy enum is internal and can't be an arg to a public method
+            Assert.That(ParallelWorkItemDispatcher.GetExecutionStrategy(work).ToString(), Is.EqualTo(expectedStrategy));
```
rprouse: I'd written a comment asking why the `ToString` before I saw this comment, so an appropriate comment 😄

PR #2159, opened 2017-05-09
Prevent crash in .NET standard with log file path

ChrisMaddock:
`DateTime.Now.ToString("o")` was producing a time stamp with colons, which was an invalid file path. Swapped it out for a fixed format.


rprouse:
Well, that was silly of me wasn't it 😄

PR #2160, opened 2017-05-09
Suppress --timeout command line option in Portable/NetStandard

### LINKED ISSUE #2147: Test Assembly using NUnitLite & Nunit 3.6.1 hangs under .NET Core when `--timeout` is supplied on command line
Given a program using NuGet packages NUnitLite and NUnit:

```csharp
using System;
using System.Reflection;
using NUnitLite;
using NUnit.Framework;

namespace NetCoreNUnitLiteTest
{
    public class Program
    {
        public static int Main(string[] args)
        {
            return new AutoRun(Assembly.GetEntryAssembly()).Execute(args);
        }

        [Test]
        public static void T1()
        {
            Console.WriteLine("hello");
        }
    }
}
```

When a `--timeout=1` is supplied on the command line, it hangs after printing out:

```
NUnitLite 3.6.1 (NetStandard 1.6)

Runtime Environment
   OS Version: Microsoft Windows 10.0.15063
  CLR Version: .NET Core 4.6.25009.03

Test Files
    c:/users/xxxxx/documents/visual studio 2017/Projects/NetCoreNUnitLiteTest/NetCoreNUnitLiteTest/bin/Debug/netcoreapp1.1/NetCoreNUnitLiteTest.dll
```

I did some experiment to debug it and realized that it hangs **before** running the test, not after.

The program runs perfectly under .Net Framework, so I guess this is a bug specially for the netstandard build.

Later I realized that timeout is not supported by a netstandard build, but I wasn't able to find any reason accounting for the hang.

BTW, is there going to be `timeout` support for .NET Core in upcoming releases of NUnit? :-)

ChrisMaddock:
Fixes #2147.

The reported hang appears to have been fixed in between 3.6.1 and the current master - so the last thing to do was remove the --timeout command line option for portable/netstandard - as has been done for the other unsupported options.

### DIFF
--- a/src/NUnitFramework/nunitlite.tests/CommandLineTests.cs
+++ b/src/NUnitFramework/nunitlite.tests/CommandLineTests.cs
@@ -273,8 +273,9 @@ namespace NUnitLite.Tests
                 Assert.AreEqual(canonicalValue, (string)property.GetValue(options, null), "Didn't recognize " + optionPlusValue);
             }
         }
-
+#if !PORTABLE && !NETSTANDARD1_6
         [TestCase("DefaultTimeout", "timeout")]
+#endif
         [TestCase("RandomSeed", "seed")]
 #if PARALLEL
         [TestCase("NumberOfTestWorkers", "workers")]
@@ -316,12 +317,14 @@ namespace NUnitLite.Tests
         // }
 
         [TestCase("--where")]
-        [TestCase("--timeout")]
 #if !PORTABLE
         [TestCase("--output")]
         [TestCase("--err")]
         [TestCase("--work")]
         [TestCase("--trace")]
+#if !NETSTANDARD1_6
+        [TestCase("--timeout")]
+#endif
 #endif
         public void MissingValuesAreReported(string option)
         {
@@ -396,7 +399,7 @@ namespace NUnitLite.Tests
             Assert.True(options.Validate());
             Assert.AreEqual(-1, options.DefaultTimeout);
         }
-
+#if !PORTABLE && !NETSTANDARD1_6
         [Test]
         public void TimeoutThrowsExceptionIfOptionHasNoValue()
         {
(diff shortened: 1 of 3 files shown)


ChrisMaddock:
Whoops - missed commenting out some tests. Will come back to this tomorrow.

ChrisMaddock:
Ok, I removed the tests.

This made me question this a little more however. The current state of CommandLineOptions.cs is that lots of unsupported options are #ifdef'd out for Portable, but this is the first one for .NET Standard.

This made me think:
1. Do we want to be not recognising unsupported options? We can do it for NUnitLite, but of course, can't do the equivalent for the console, as the console doesn't know what framework it is loading, and if the option will be available.
2. If we DO want to merge this - @rprouse, could you have a quick skim through CommandLineOptions.cs, and check if any other command line options are not available in the .NET Standard 1.6 build?

Thanks.

oznetmaster:
I have always left all commands in place, and returned an informative error message that indicated the command was not valid in that context, and what context it was valid in.

ChrisMaddock:
That would be nice for NUnit Lite, but wouldn't work for the console, where the runner doesn't know what the framework supports.

One solution could be the framework warning if it encounters any settings it doesn't recognise - but am I right in thinking that the current engine passes all EngineSettings/FrameworkSettings to the framework, and doesn't distinguish between the two? That would make that solution not backwards compatible with old runners/new frameworks.

Otherwise, I guess everywhere the framework reads a setting which isn't interpreted in all platforms, it could warn on those platforms unsupported. That sounds messy, though.

@rprouse - leaving this open for you to have a skim through re: point 2 above. No rush. :smil

CharliePoole:
@ChrisMaddock I think the question of special handling for options that are only known to one framework build is peculiar to NUnitLite, which is after all a different program from the console. It could reasonably go either way but I don't think it belongs as part of this issue. 

Giving an error message is definitely preferable to silently ignoring the option. Giving a specially tailored message for some options in some conditions might be an enhancement to consider or it might just confuse people. IMO, it's not worth it, but it's definitely not part of this issue.

BTW, I once had the notion that NUnitLite should give a special warning message for options that the console runner uses but which it doesn't. We decided that wasn't worth it either. 😄

PR #2174, opened 2017-05-18
Convert Portable to .NET Standard 1.3

### LINKED ISSUE #2092: Convert Portable library to .NET Standard 1.3
The portable library is .NET Standard 1.3 (or possibly 1.0 or 1.1), but it cannot easily be installed in a .NET Standard 1.3 project. Attempting to do so results in the following error,

```
Restoring packages for C:\src\spikes\NetstandardLibrary\NetstandardLibrary\NetstandardLibrary.csproj...
Package NUnit 3.6.1 is not compatible with netstandard1.3 (.NETStandard,Version=v1.3). Package NUnit 3.6.1 supports:
  - monoandroid (MonoAndroid,Version=v0.0)
  - net20 (.NETFramework,Version=v2.0)
  - net35 (.NETFramework,Version=v3.5)
  - net40 (.NETFramework,Version=v4.0)
  - net45 (.NETFramework,Version=v4.5)
  - netstandard1.6 (.NETStandard,Version=v1.6)
  - portable-net45+win8+wp8+wpa81 (.NETPortable,Version=v0.0,Profile=Profile259)
  - xamarinios10 (Xamarin.iOS,Version=v1.0)
One or more packages are incompatible with .NETStandard,Version=v1.3.
Package restore failed. Rolling back package changes for 'NetstandardLibrary'.
```

If we convert the project to the lowest .NET Standard version it supports instead of PCL, it will still work in all the PCL targets, plus it will work in .NET Standard.

Before we do this, can anyone see any issues with this approach?

### LINKED ISSUE #2165: Trying to install NUnit 3.6.1 on .NET Framework asks for download of 20 more packages
I can install NUnit on .Net Framework 4.6.1 without a problem but when I try to install it via Nuget on the newly released 4.7 it asks for these packages:
capture

And this is only half of the list

rprouse:
Fixes #2092
Fixes #2165 

We had talked about targeting .NET Standard 1.0, but I decided to go with 1.3 because we don't have and runners that support < 1.3 and we never had runners that targeted any of the platforms that aren't supported by 1.3. Also, targeting 1.3 allowed me to re-enable many of the features like path and directory constraints that weren't available in Portable or 1.0.

I will add additional comments inline as I do my own code review.

I used the trailing whitespace remover extension recommended by @jnm2 so you may want to append `?w=1` to the file diff to reduce the noise.

If you have an older copy of `nuget.exe` in the `tools` directory, you will need to delete it so it refetches an updated version that understands .NET Standard.


build.cake
```diff
@@ -81,6 +81,37 @@ var packages = new string[]{
 
 Setup(context =>
 {
+    if (isAppveyor)
```
rprouse: Moved this to make it consistent with other projects and because setting variables belongs in a setup method, not a build task that may not be called.

global.json
```diff
@@ -1,6 +1,6 @@
 {
     "projects": [ "src" ],
     "sdk": {
-        "version": "1.0.0-preview2-003131"
+        "version": "1.0.3"
```
rprouse: Time to updated to a release version of .NET Core. Developers may need to download the updated package if you don't have it.

nuget/framework/nunit.nuspec
```diff
@@ -28,32 +28,33 @@ Supported platforms:
...
-    <file src="bin/net-4.5/nunit.framework.xml" target="lib/net45" />
-    <file src="bin/netstandard16/nunit.framework.dll" target="lib/netstandard1.6" />
-    <file src="bin/netstandard16/nunit.framework.xml" target="lib/netstandard1.6" />
-    <file src="bin/portable/nunit.framework.dll" target="lib/portable-net45+win8+wp8+wpa81" />
-    <file src="bin/portable/nunit.framework.xml" target="lib/portable-net45+win8+wp8+wpa81" />
-    <file src="bin/portable/nunit.framework.dll" target="lib/MonoAndroid" />
-    <file src="bin/portable/nunit.framework.xml" target="lib/MonoAndroid" />
-    <file src="bin/portable/nunit.framework.dll" target="lib/Xamarin.iOS10" />
-    <file src="bin/portable/nunit.framework.xml" target="lib/Xamarin.iOS10" />
+    <file src="bin/net-2.0/nunit.framework.dll" target="lib\net20" />
```
rprouse: The backslashes were causing warnings with the latest version of nuget.exe.

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
```
rprouse: Quite a bit of trimmed whitespace since I started using the extension that @jnm2 recommended. Not sure I like that much pollution in the PRs ☹️

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -348,7 +347,7 @@ public override ConstraintResult ApplyTo<TActual>(ref TActual actual)
 
         private static object InvokeDelegate<T>(ActualValueDelegate<T> del)
         {
-#if NET_4_0 || NET_4_5
+#if NET_4_0 || NET_4_5 || NETSTANDARD1_3 || NETSTANDARD1_6
```
rprouse: This was broken in .NET Standard 1.6 and nobody noticed.

src/NUnitFramework/TestBuilder.cs
```diff
@@ -34,7 +34,7 @@
 using NUnit.Framework.Internal.Commands;
 using NUnit.Framework.Internal.Execution;
 
-#if PORTABLE && !NETSTANDARD1_6 && !NETCOREAPP1_0
+#if NETSTANDARD1_3 && !NETSTANDARD1_6 && !NETCOREAPP1_0
```
ChrisMaddock: More a question - why do we have `NETCOREAPP1_0` precompilation symbols as well as NETSTANDARD?

src/NUnitFramework/framework/Assert.cs
```diff
@@ -378,28 +378,8 @@ private static void IssueWarning(string message)
...
-                stackTrace = ex.StackTrace;
-            }
-#else
-            stackTrace = SystemEnvironmentFilter.Filter(Environment.StackTrace);
-#endif
-
-            return StackFilter.DefaultFilter.Filter(stackTrace);
-        }
+        private static string GetStackTrace() =>
+            StackFilter.DefaultFilter.Filter(SystemEnvironmentFilter.Filter(Environment.StackTrace));
```
ChrisMaddock: Nice to see all those conditionals going away here! 😄

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
ChrisMaddock: I agree to be honest, it makes reviewing harder.

I'd be in favour of doing a single PR to 'fix the whitespace' - to clear up anything that might get autocorrected in PRs like this. 🙂

src/NUnitFramework/framework/Properties/AssemblyInfo.cs
```diff
@@ -40,7 +40,7 @@
 [assembly: AssemblyTitle("NUnit Framework .NET 4.0")]
 #elif NET_2_0
 [assembly: AssemblyTitle("NUnit Framework .NET 2.0")]
-#elif PORTABLE
+#elif NETSTANDARD1_3
 [assembly: AssemblyTitle("NUnit Framework Portable")]
```
ChrisMaddock: Stray "Portable" here. Although looks like this file is generally out of date, it's missing the .NET 3.5 and .NET Standard 1.6 builds too.

src/NUnitFramework/mock-assembly/mock-assembly-netstandard13.csproj
```diff
@@ -63,14 +64,17 @@
     <None Include="..\..\nunit.snk">
       <Link>nunit.snk</Link>
     </None>
-    <None Include="mock-assembly-portable.project.json" />
+    <None Include="mock-assembly-netstandard13.project.json" />
   </ItemGroup>
   <Import Project="$(MSBuildExtensionsPath32)\Microsoft\Portable\$(TargetFrameworkVersion)\Microsoft.Portable.CSharp.targets" />
```
ChrisMaddock: Probably don't want this line anymore? Was this csproj auto-upgraded?

ChrisMaddock: This is in a few of the other csproj's too - maybe worth a grep. 🙂

src/NUnitFramework/nunitlite/Options.cs
```diff
@@ -123,7 +123,7 @@
 //      p.Parse (new string[]{"-a-"});  // sets v == null
 //
 // The NUnit version of this file introduces conditional compilation for
-// building under a portable class library  (PORTABLE).
+// building under a portable class library  (NETSTANDARD1_3).
 //
```
ChrisMaddock: Find and replace error? If not, the comment probably wants updating too. 🙂

src/NUnitFramework/framework/Properties/AssemblyInfo.cs
ChrisMaddock: This was also true for other AssemblyInfo files - again, maybe worth a grep.

nuget/framework/nunit.nuspec
ChrisMaddock: That's very odd. I would have thought forward slashes would be preferred for the new, x-plat NuGet!

nuget/nunitlite/nunitlite.nuspec
```diff
@@ -30,29 +30,34 @@ How to use this package:
...
+    <file src="bin/net-3.5/nunitlite.dll" target="lib\net35" />
+    <file src="bin/net-4.0/nunitlite.dll" target="lib\net40" />
+    <file src="bin/net-4.5/nunitlite.dll" target="lib\net45" />
+    <file src="bin/netstandard13/nunitlite.dll" target="lib\netstandard1.3" />
+    <file src="bin/netstandard16/nunitlite.dll" target="lib\netstandard1.6" />
+    <file src="bin/net-2.0/nunitlite-runner.exe" target="tools\net20" />
+    <file src="bin/net-3.5/nunitlite-runner.exe" target="tools\net35" />
+    <file src="bin/net-4.0/nunitlite-runner.exe" target="tools\net40" />
+    <file src="bin/net-4.5/nunitlite-runner.exe" target="tools\net45" />
+    <file src="bin/netstandard16/nunitlite-runner.exe" target="tools\netcore1.0" />
```
ChrisMaddock: I presume this should have netstandard13 nunitlite-runner.exe added too?

src/CommonAssemblyInfo.cs
```diff
@@ -31,7 +31,7 @@
 [assembly: AssemblyCopyright("Copyright (C) 2017 Charlie Poole")]
 [assembly: AssemblyTrademark("NUnit is a trademark of NUnit Software")]
 
-#if PORTABLE
+#if NETSTANDARD1_3
 [assembly: AssemblyMetadata("PCL", "True")]
```
ChrisMaddock: This probably should be removed?

src/NUnitFramework/TestBuilder.cs
rprouse: The NUnitLite tests pull in this file and it is compiled .NET Core 1.0.

src/NUnitFramework/framework/Properties/AssemblyInfo.cs
rprouse: I've done a quick search and fixed a few. Thanks.

src/NUnitFramework/mock-assembly/mock-assembly-netstandard13.csproj
rprouse: No, we do want these. We are using the old `CSPROJ` portable format, not the new .NET Core one, so the `project.json` files are still needed. This allows us to build in VS2015. At some point I would like to upgrade all of our projects to 2017 and the new project so we can use multi-targeting, but that is a ways out.

nuget/nunitlite/nunitlite.nuspec
rprouse: No, the two runners are both .NET Core 1.0, so we only need one of them. I build two because it is easier conceptually to keep them in sync.

rprouse:
Some good finds @ChrisMaddock. I think I've fixed all instance of left over portables in the comments and docs. 👍

src/NUnitFramework/mock-assembly/mock-assembly-netstandard13.csproj
ChrisMaddock: Sorry, it was the line 69, the `Microsoft.Portable.CSharp.targets` I was referring to - which I assume is PCL specific?

nuget/nunitlite/nunitlite.nuspec
ChrisMaddock: D'oh, sorry. I was reading the source and target backwards!

src/NUnitFramework/mock-assembly/mock-assembly-netstandard13.csproj
rprouse: I tried removing them just to be sure, but they are needed. Technically, these are portable projects, that is why you see the (Portable) after them in solution explorer.

ChrisMaddock: Ah ok - I assumed that was Portable as in PCL, rather than Portable as in X-Plat. Sorry!

ChrisMaddock:
I just tried to update the nunit.xamarin solution, and the UWP project is unhappy with the System.Threading.Thread dependency. I'm sure you'll have a better idea how to resolve this than I would, Rob. 😞 

    System.Threading.Thread 4.3.0 provides a compile-time reference assembly for System.Threading.Thread on UAP,Version=v10.0, but there is no run-time assembly compatible with win10-x86-aot.

rprouse:
Well, that really sucks. I'll take a look.

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
jnm2: Ack I'm sorry! :-( You can turn off "trim on save" in both extensions so that you can still see whitespace but it won't touch it.

jnm2: The weird part is that Visual Studio won't show differing whitespace. Only GitHub and the git command line. Wish this could be set for GitHub via .gitattributes or something.

jnm2: Hey guys, look at this version of the page: [?w=1](?w=1) (https://github.com/blog/967-github-secrets)

rprouse:
I added the fix for #2165 because the fix for that relies on the `nuspec` changes here. See 16d60b3
for just those fixes.

rprouse:
@ChrisMaddock we might need to update the Xamarin UWP project. I just created a UWP library and a UWP app and managed to add the package without any issues,

```
Restoring packages for C:\src\Spikes\TargetingTests\AppUwp\AppUwp.csproj...
Committing restore...
Writing lock file to disk. Path: C:\src\Spikes\TargetingTests\AppUwp\obj\project.assets.json
Restore completed in 932.19 ms for C:\src\Spikes\TargetingTests\AppUwp\AppUwp.csproj.
Successfully installed 'NETStandard.Library 1.6.0' to AppUwp
Successfully installed 'NUnit 3.7.0' to AppUwp
Successfully installed 'System.Console 4.0.0' to AppUwp
Successfully installed 'System.Runtime.InteropServices.RuntimeInformation 4.0.0' to AppUwp
Successfully installed 'System.Threading.Thread 4.3.0' to AppUwp
Successfully installed 'runtime.win.System.Console 4.0.0' to AppUwp
Executing nuget actions took 1.41 sec
Time Elapsed: 00:00:02.4456822
========== Finished ==========
```

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
rprouse: @jnm2 mentioned that above, but too much noise building up on this PR 😸 

> I used the trailing whitespace remover extension recommended by @jnm2 so you may want to append `?w=1` to the file diff to reduce the noise.

Sorry, I probably wasn't very clear in my instructions. It is an old trick that @CharliePoole and I use, so thought it was obvious 😄

jnm2: No, I didn't even see that. 👍

ChrisMaddock:
Ah ok. I updated to the latest UWP version - but I guess there must be other settings. Cool - can revisit that one later than, if it's an nunit.xamarin issue, rather than framework.

global.json
ChrisMaddock: The bootstrapper build scripts will need updating to match this - they specify a version of core to install. (The cake version number in the same place is also out of date, if you wouldn't mind, while you're there! 😄 )

src/NUnitFramework/framework/nunit.framework-netstandard13.project.json
```diff
@@ -0,0 +1,10 @@
+﻿{
+  "supports": {},
+  "dependencies": {
+    "NETStandard.Library": "1.6.0",
```
ChrisMaddock: I get a raft of warnings from this (and the .NET Standard 1.6) project referencing version 1.6.0 here, which go away when I upgrade this to 1.6.1. I _think_ that's the right solution to the problem?

> Warning		Detected package downgrade: System.Globalization from 4.3.0 to 4.0.11 
 nunitlite-runner-netstandard16 (>= 1.0.0) -> nunitlite-netstandard16 (>= 1.0.0) -> NETStandard.Library (>= 1.6.0) -> System.Net.Primitives (>= 4.0.11) -> runtime.unix.System.Net.Primitives (>= 4.3.0) -> System.Globalization (>= 4.3.0) 
 nunitlite-runner-netstandard16 (>= 1.0.0) -> nunitlite-netstandard16 (>= 1.0.0) -> NETStandard.Library (>= 1.6.0) -> System.Globalization (>= 4.0.11)

ChrisMaddock:
The following test files are missing from the new .NET Standard 1.3 project, which I don't think should be? (Some of these are also missing from .NET Standard 1.6 and other builds, which could be related.) For reference, I used the 'Show Missing Files' extension to find these. 🙂 

- IntervalTests.cs
- EventListenerTextWriterTests.cs
- PropertyFilterTests.cs
- TestResultWarningTests.cs

ChrisMaddock:
They're the only issues I could find, other than those, everything looks great! 😄

rprouse:
Great review @ChrisMaddock thanks. If nobody else has comments or suggestions, I would like to get this merged after I make the suggested changes.

I'm away for a long weekend with a full house, but I will do my best to carve out some time to get these changes done today.

CharliePoole:
@rprouse Did I misunderstand you? I thought you were going to do the release first.

rprouse:
I wanted to get it in the release because it will make it easier to support more platforms in the adapter and it reenables many features that weren't available in portable. Is there a reason to hold off?

CharliePoole:
Just that it's a big deal and has had no exercising by users in the form of a dev release. Those folks who are using portable will suddenly not have it available. 

I've always been accustomed to freezing an app before release because new features at the last minute have a tendency to break things.

But mainly, because that's what I though you had said in an earlier comment.

global.json
rprouse: @ChrisMaddock the .NET version in the build script is the version that Cake requires, not what we require. This is the version that Cake ships with 1.19.5, I'm not sure if we should be changing it. What do you think? https://github.com/cake-build/cake/blob/v0.19.5/build.ps1

ChrisMaddock: Oh, sorry - I thought we'd set it. Just coincidence that they're the same - I saw the output from the bootstrapped installing the old version. 

It would be good to have a bootstrapped to install the right .NET core SDK, but that's probably a separate issue. 🙂

rprouse:
@CharliePoole , when I talked about the release, I said, that I wanted to get this in but I was having trouble getting it working at that time. My issues were build problems ensuring that it would build with older versions of Visual Studio, not issues with code. The code actually changed very little except that this enables features for users. Those features are also already tested in the .NET Standard 1.6 target, so the risk is minimal.

Current users of portable will also not see any change when they update and won't have to do anything.

I also don't think many people test our dev feeds before we do a release no matter how much time we give them to test. If you look at our [myget feed](https://www.myget.org/feed/nunit/package/nuget/NUnit) most builds have zero downloads and the max is 4. I think most of those are us testing internally.

I worry that the longer we leave this the more merge conflicts we will have and the greater chance of portable code creeping back in.

rprouse:
@ChrisMaddock I have added those missing files to both of the .NET Standard projects that were missing them. Thanks.

src/NUnitFramework/framework/nunit.framework-netstandard13.project.json
rprouse: That is really weird, I don't get any warnings on initialize, build or run, but I do if I attempt to upgrade to 1.6.1. I was reticent to upgrade to 1.6.1 because we haven't tested that with the Visual Studio adapter, but i don't know if that is a problem. I'm also not seeing the warnings in the AppVeyor build, https://ci.appveyor.com/project/CharliePoole/nunit/build/3.7.0-ci-03975-pr-2174

rprouse: I think I see the problem. I am building on the command line and see no warnings, but I see them building in Visual Studio. I just did a `which dotnet` on the command line and it looks like Cake overrides the PATH and puts the version of dotnet it downloads in, so I end up using the older version at the command line, not the version specified in `global.json`.

rprouse: @ChrisMaddock are you building in Visual Studio 2017? I am and get a whole raft of warnings compiling master too. I think it is because 2017 uses the new MSBUILD. I could not update to 1.6.1 I think because it requires the newer 2017 tooling, but we are using the old PCL project files so we can support building in 2015. At some point, we are going to have to make the call on moving to Visual Studio 2017 because not requiring it is making it really hard to support newer platforms.

For now, I downgraded the .NET Core SDK back to what it was originally and left the projects targeting 1.6.0.

CharliePoole:
@rprouse If that's the only problem, I'm OK with it. We should talk more about release policies in the core team.

rprouse:
@CharliePoole I am assuming that is an okay to merge? I am going to resolve conflicts again and would like to get this out of the way so I can concentrate on getting ready for the release. Does anyone have objections?

PR #2176, opened 2017-05-19
Typo fix

jnm2:
Same as https://github.com/nunit/nunit-console/pull/226


rprouse:
You don't speak French? That is supposed to be the attaché that communicates between NUnit and the debugger! 😄

rprouse:
Build failures are caused by the branch name. I am overriding and merging.

jnm2:
Sadly, I do not. Is verbification of nouns prevalent in French too? 😀

Issue #1200, opened 2016-01-16
async test + Apartment(ApartmentState.STA) => await not returning on STA thread


yar-shukan:
There is the old post on SO: 
http://stackoverflow.com/questions/18838116/nunit-async-test-requiressta-await-not-returning-on-sta-thread
It seems like this issue has been there long time ago, but we faced it only after upgrading to NUnit 3.0.1.

We have tests that use async/await with Apartment(ApartmentState.STA) and before the upgrade to NUnit 3.0.1 they used to work, but after the upgrade they fail because WPF components don't run on STA thread after await statement.

rprouse:
Are you saying that when the test is marked STA, the thread is no longer STA after an await in NUnit 3, or are you saying that the workarounds in the SO post no longer work?

yar-shukan:
Yes, it is no longer STA after await in NUnit 3.2.1 (the same issue in 3.0.1 as well).
Marking test method with [Apartment(ApartmentState.STA)] as well as obsolete [RequiresSTA] doesn't work. The managed thread ID is different after await.

keithallenjackson:
My current work project has ran into this very issue. I'm curious if any work has been done to address this. I'm interested in at least looking into it, but wanted to touch base and get a status update to not duplicate work effort.

Rincoln:
I've had to do the same thing to work with proprietary ODBC drivers.  My solution is a little convoluted, but works.  

public Task<T> RunSTAThread<T>(Func<T> action)
        {
            var tcs = new TaskCompletionSource<T>();
            System.Threading.Thread runThread = new System.Threading.Thread(new System.Threading.ThreadStart(() =>
            {
                try
                {
                    var result = action();
                    tcs.TrySetResult(result);
                }
                catch (Exception ex)
                {
                    tcs.TrySetException(ex);
                }
            }));
            runThread.SetApartmentState(System.Threading.ApartmentState.STA);
            runThread.Start();
            return tcs.Task;
        }

CharliePoole:
This was still marked as `confirm` and nothing has been done. I guess we should at least consider it confirmed.

jnm2:
There's an easy fix for this, I just did this the other day. We need to set a synchronization context and message pump on the STA thread. Otherwise, `await` schedules continuations on a random thread pool thread.

jnm2:
Stephen Toub gives an excellent example:
[...]

Issue #34, opened 2013-10-20
Async testing with F#


haf:
I have been trying to find on how to test asynchronous code from F# without having to manually unwrap the async value, but rather having the test framework do that for you.

Since it seems 3.0 will bring with it improvements to async as well as the ability to parallelise tests, it would be great if there was some guidance on how to use the project with F#.

My question is simpler than this however; how do I create tests with F# that are async without wrapping them in TPL?

nosami:
I just discovered a pattern that works for F#

```fs
open System.Threading.Tasks
open System.Runtime.CompilerServices

let toTask computation : Task = Async.StartAsTask computation :> _

[<Test>]
[<AsyncStateMachine(typeof<Task>)>]
member x.``Test``() = toTask <| async {
    let! x = doAsyncStuff()
}
```

nosami:
Would be much nicer if NUnit supported F# async natively though

rprouse:
@nosami, yes it would. Contributions from the F# community are welcome, or even just some help for us functional neophytes :smile:

yreynhout:
I just discovered another pattern that works for F#

```fsharp
open System.Threading.Tasks
open NUnit.Framework

let toAsyncTestDelegate computation = new AsyncTestDelegate(fun () -> Async.StartAsTask computation :> Task)

[<Test>]
member x.``Test``() = 
    Assert.ThrowsAsync<InvalidOperationException>(asyncStuff 123 |> toAsyncTestDelegate) |> ignore
```

nosami:
[This](https://github.com/xunit/xunit/commit/a2aa665f1b9633e8ec00f549c5df73a0c784a742) is how @bradwilson implemented it for xunit

jnm2:
@rprouse Should I add an .fsproj to the solution (`nunit.testdata.fsharp`) for the IDE experience, or should I compile F# test fixtures on the fly using by referencing the FSharp.Compiler.Service nupkg from `nunit.framework.tests`?

The .fsproj would require all NUnit contributors to have the desktop F# SDK and the .NET Core F# SDK installed which may or may not be default. I could check.

mikkelbu:
I prefer the second option, but I'm also a little biased as I already have F# installed 😄.

rprouse:
I would prefer an F# project because it is easiest and I have it installed too, but it isn't a default workload in VS. How hard is it to use the FSharp.Compiler.Service?

bradwilson:
[...]

PR #2181, opened 2017-05-21
[WIP] Using C# awaitable pattern for test methods and delegates

### LINKED ISSUE #2168: ThrowsAsync reports OperationCanceledException as TaskCanceledException
The `ThrowsAsync` method reports exceptions of type `OperationCanceledException` as `TaskCanceledException` instead.

Repro:

```
Assert.ThrowsAsync<OperationCanceledException>(async () =>
{
    await Task.Yield();
    throw new OperationCanceledException();
});
```

Result:

```
Failed : NUnitRepro.Tests.Repro
  Expected: <System.OperationCanceledException>
  But was:  <System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at NUnit.Framework.Assert.ThrowsAsync(IResolveConstraint expression, AsyncTestDelegate code, String message, Object[] args)>
```

Other exception types are reported correctly. The following test passes.

```
Assert.ThrowsAsync<TimeoutException>(async () =>
{
    await Task.Yield();
    throw new TimeoutException();
});
```

I believe it's possible to workaround this issue using the "constraint model". The following test passes.

```
Assert.That(async () =>
{
    await Task.Yield();
    throw new OperationCanceledException();
}, Throws.InstanceOf<OperationCanceledException>());
```

I observed the issue in version 3.5.0. I updated to 3.6.1 and confirmed that the issue is still present in the latest version.

I also installed the console runner and tested from the command line as per the contributing guidelines. Test output below.

```
C:\...\NUnitRepro\NUnitRepro\bin\Debug>..\..\..\packages\NUnit.ConsoleRunner.3.6.1\tools\nunit3-console.exe NUnitRepro.exe
NUnit Console Runner 3.6.1

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.14393.0
  CLR Version: 4.0.30319.42000

Test Files
    NUnitRepro.exe


Errors, Failures and Warnings

1) Failed : NUnitRepro.Tests.Repro
  Expected: <System.OperationCanceledException>
  But was:  <System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at NUnit.Framework.Assert.ThrowsAsync(IResolveConstraint expression, AsyncTestDelegate code, String message, Object[] args)>
at NUnitRepro.Tests.Repro() in c:\...\NUnitRepro\NUnitRepro\Tests.cs:line 16

Run Settings
    DisposeRunners: True
    WorkDirectory: C:\...\NUnitRepro\NUnitRepro\bin\Debug
    ImageTargetFrameworkName: .NETFramework,Version=v4.6
    ImageRequiresX86: True
    RunAsX86: True
    NumberOfTestWorkers: 4

Test Run Summary
  Overall result: Failed
  Test Count: 3, Passed: 2, Failed: 1, Warnings: 0, Inconclusive: 0, Skipped: 0
  Start time: 2017-05-15 15:43:50Z
    End time: 2017-05-15 15:43:50Z
    Duration: 0.564 seconds

```

jnm2:
Please do not merge until I add a couple more tests.

Fixes #2168.

I've been wanting to write this code for so long! Reviews requested because this is another of those deep dives.

Task.Wait() is no longer used. It's from the pre-async era and it has a number of problems, one of which showed up in #2168. Instead, the C# awaitable spec is followed, calling GetAwaiter, IsCompleted, OnCompleted and GetResult as appropriate. A ManualResetEventSlim is used to block wait for completion if the awaitable is not already completed. `Task` also uses a MRES if you block wait via .Wait() or .GetAwaiter().GetResult().

https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#awaitable-expressions
https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#runtime-evaluation-of-await-expressions


rprouse:
I will review later, but I would prefer to get the.net standard changes merged before we attempt more deep dives to prevent merge conflicts

CharliePoole:
This is a big change. Let's hold it till after the release that's underway.

rprouse:
We dropped the ball on this. What is the status @jnm2?

_Sent from my Google Pixel using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

jnm2:
I'm interested in feedback on the approach, I need to write some more tests, and one think I was looking at doing was consistency tests where instead of testing each feature in its own class, I test cross-cutting aspects that we'd like to have the same everywhere like "async void is invalid."

Also it connects to https://github.com/nunit/nunit/issues/1200; I'd like to finish this work first so that we don't end up implementing that issue twice.

jnm2:
Looks like I have some merge conflicts, nothing too intense.

jnm2:
I should also see what it entails to implement native F# async support (https://github.com/nunit/nunit/issues/34).

jnm2:
Note to self: need to propose Assert methods that are themselves async for use with async delegates after this PR is through.

Problem:
```cs
[Test]
public static void Foo()
{
    WindowsFormsUtils.RunWithMessagePump(async () =>
    {
        Assert.That(async () =>
        {
            // ...

            // For the duration of this delay, the message pump cannot run
            // because it is being blocked by Assert.That which is waiting for this async delegate.
            // So, the test deadlocks.
            await Task.Delay(1000);

            // ...
        }, Throws.InstanceOf<SomeException>());
    });
}
```

Simplest solution for Assert.Throws (Assert.That is more complex):

```cs
public static class AssertAsync
{
    public static async Task<TActual> Throws<TActual>(AsyncTestDelegate code) where TActual : Exception
    {
        if (code == null) throw new ArgumentNullException(nameof(code));

        try
        {
            await code.Invoke();
        }
        catch (TActual ex)
        {
            return ex;
        }
        catch (Exception ex)
        {
            throw new AssertionException($"Expected {typeof(TActual).FullName} to be thrown, but {ex.GetType()} was thrown.");
        }

        throw new AssertionException($"Expected {typeof(TActual).FullName} to be thrown, but no exception was thrown.");
    }
}
```

Usage:

```cs
[Test]
public static void Foo()
{
    WindowsFormsUtils.RunWithMessagePump(async () =>
    {
        await AssertAsync.Throws<AssertionException>(async () =>
        {
            // ...

            // Yay, AssertAsync.Throws doesn't block while waiting for this async delegate.
            // So the message pump is running for the duration of this delay! No deadlock.
            await Task.Delay(100);

            // ...
        });
    });
}
```

jnm2:
Idea: I have a 129-line proof-of-concept AsyncVoidVerificationScope class which uses Event Tracing for Windows to detect exceptions thrown by async void methods and async void tasks which have not finished running. If there is any parallelism, it might be difficult or impossible to correlate it with the correct test. Needs investigation. Even without correlation, it's still valuable to know that *some* test threw an exception so we might do this on a test-run basis rather than a per-test basis.

jnm2:
I think everything in this PR has been covered by the merged https://github.com/nunit/nunit/pull/2776, the soon-to-be-merged https://github.com/nunit/nunit/pull/2818, and the custom awaitable PR which will be merged once I have time to transfer it from my head to the keyboard.

These recent PRs are a much more mature and targeted version of this PR, so, closing. 😊

jnm2:
> Note to self: need to propose Assert methods that are themselves async for use with async delegates after this PR is through.

While this API might be nice, the tests will no longer deadlock since https://github.com/nunit/nunit/pull/2776 causes `Assert.That` to run the message pump.

PR #2183, opened 2017-05-22
Fixed typos

jnm2:


jnm2:
Oh! How hard would it be to replace unsupported chars with `-` when creating the package version string?

ChrisMaddock:
@jnm2 - we already make modifications for length, we should just do the same for special chars. I was just about to log an issue - shall I leave it with you instead? 🙂 

https://github.com/nunit/nunit/blob/master/build.cake#L157

jnm2:
>  I was just about to log an issue - shall I leave it with you instead?

Makes no difference to me.

jnm2:
@ChrisMaddock https://github.com/nunit/nunit/issues/2186

ChrisMaddock:
Thanks!

PR #2184, opened 2017-05-22
[Discussion] Adding TestParameters.TryGetValue

jnm2:
Please do not merge until consensus is reached.

This is a primitive that comes in very handy.

```c#
var foo = TestContext.Parameters.TryGetValue("name", out var passedValue)
    ? ParseFoo(passedValue)
    : defaultFooValue;
```

### DIFF
--- a/src/NUnitFramework/framework/TestParameters.cs
+++ b/src/NUnitFramework/framework/TestParameters.cs
@@ -60,6 +60,18 @@ namespace NUnit.Framework
         }
 
         /// <summary>
+        /// Gets the value of the parameter with the specified name.
+        /// </summary>
+        /// <param name="name">The name of the parameter. If <see langword="null"/>, throws <see cref="ArgumentNullException"/>.</param>
+        /// <param name="value">The value of the parameter, if the name is found.</param>
+        /// <returns>True if a parameter with the given name exists, otherwise false.</returns>
+        /// <exception cref="ArgumentNullException">Thrown when <paramref name="name"/> is <see langword="null"/>.</exception>
+        public bool TryGetValue(string name, out string value)
+        {
+            return _parameters.TryGetValue(name, out value);
+        }
+
+        /// <summary>
         /// Get the value of a parameter or a default string
         /// </summary>
         /// <param name="name">Name of the parameter</param>
--- a/src/NUnitFramework/tests/TestParametersTests.cs
+++ b/src/NUnitFramework/tests/TestParametersTests.cs
@@ -107,6 +107,34 @@ namespace NUnit.Framework.Tests
 
         #endregion
 
+        #region TryGetValue
+
+        [Test]
+        public void TryGetValue_Existing()
+        {
+            _parameters.Add("SomeParm", "5");
+            string value;
+            Assert.That(_parameters.TryGetValue("SomeParm", out value), Is.True);
+            Assert.That(value, Is.EqualTo("5"));
+        }
+
+        [Test]
+        public void TryGetValue_Missing()
+        {
+            _parameters.Add("SomeParm", "5");
+            string _;
+            Assert.That(_parameters.TryGetValue("SomeOtherParm", out _), Is.False);
+        }
+
+        [Test]
+        public void TryGetValue_NullKeyThrowsException()
+        {
+            string _;
+            Assert.That(() => _parameters.TryGetValue(null, out _), Throws.ArgumentNullException);
+        }
+
+        #endregion
+
         #region Get String
 
         [Test]


CharliePoole:
@jnm2 It's really much better to write an issue if you want to add (even a small) feature. That way we can discuss it without the pressure that is created when you have already spent a little time writing code. It's much harder to resist feature creep when the coding has already been done, even though that's what I'm about to do. 😸 

Anyway, let's discuss it here. `TestParameters` already provides `Get`, which returns null when no parameter has been provided. There is no other way for a parameter to be null. What is this API adding? Will this make something that's hard to write with the existing API easier?

jnm2:
> Will this make something that's hard to write with the existing API easier?

Yes, that is my motivation. This is a single expression:

```c#
TestContext.Parameters.TryGetValue("name", out var passedValue)
    ? ParseFoo(passedValue)
    : defaultFooValue
```

There's no nice way to replicate this with `Get`. You are fixed to declare a local variable in a separate statement:

```c#
// Statement:
var passedValue = TestContext.Parameters.Get(name);
// Expression:
passedValue != null ? ParseFoo(passedValue) : defaultFooValue
```

It's also more verbose, and it couples my code to the fact that there is not *currently* a way to add a parameter with a value of `null`. That's not obvious to the readers of my code.

CharliePoole:
@ChrisMaddock Yes, I think you are right to not merge. If we don't get agreement first that something is a feature we want to support, but rather just do it as code review, then any two devs can add any feature any time. That's not what we want as a general practice.

In this context, when I say agreement, I mean either consensus or at least a decision by the project lead. If this were an internal class, I'd see less problem in skipping the issue step, but it's a public API used in tests, so making any changes, even additions, has to be reviewed for pros and cons.

As far as pros and cons for this particular case, I feel the confusion of providing two ways to do the same thing outweighs the possible convenience, but I could be convinced if there is a good use case.

jnm2:
Next time I will start a discussion issue and add a Do Not Merge message if I do illustrate with a PR.

CharliePoole:
@jnm2 To be completely honest, I didn't know you could write a statement that way in C# 6.0, so my alternative solution would have been:

```C#
string passedValue;
TestContext.Parameters.TryGetValue("name", out passedValue)
    ? ParseFoo(passedValue)
    : defaultFooValue
```

which is clearly not as good. But even so, it seems to me that the best solution of all is

```C#
ParseFoo(TestContext.Parameters.Get("name", "default_string_value"));
```

jnm2:
@CharliePoole 
> To be completely honest, I didn't know you could write a statement that way in C# 6.0

This is a C# 7.0 feature. I see no reason for NUnit's internal development language level to influence the expectations of the limits NUnit's consumers have.

> ```C#
> ParseFoo(TestContext.Parameters.Get("name", "default_string_value"));
> ```

The other motivation for TryGetValue was so that I did not have to serialize the default value into a string. On a theoretical level it may be impossible to serialize the default value to a string. In my case, it's just an annoyance.

CharliePoole:
@jnm2 I see the point. 

That said, the design of the TestParameters facility __is__ that parameters are supposed to be strings and that null means no parameters were provided. We discussed this design and decided to do it that way even though some folks thought it should be different. So, worrying that null may eventually mean something different seems like extreme YAGNI to me - kind of like worrying that NUnit will stop recognizing some particular attribute.

However, I'm not irreconcilably opposed if others like it. I'm frankly not a fan of writing tests as concisely as possible. Extra lines often help both in clarity and in giving me a place to put a break and look at a particular value. To the extent this is style, we can always leave it to the users.

As far as the more general question of issues... all that would have been needed to make me happy was something like "I'd like to add a method like this... what do you guys think." In a co-located team ("normal" team as I call it) you would just shout that out to the room and, if there were a need to discuss, go over to the white board. In our situation, we are forced to use the issue as an intermediary.

Outside contributors will, of course, provide PRs without issues. They don't know any better. :smile:

This is my own view, of course. Since I'm kind of a lame duck here, it's possible that things may be done differently at some point. But I think it will be helpful if we all start from the same point, whatever comes.

oznetmaster:
Alternatively, one can use the Get<T>(string name, T default value) method.

If T implements IConvertible, or has a TypeConverter from string (that I assume would use ParseFoo) then this is the most concise way to write this.

oznetmaster:
Actually, I need to check.

My CF build has supported the use of TypeConverter in processing parameters for years.  I know I proposed, in a PR, adding that capability to the regular build, but I am not sure that it was ever accepted :(

jnm2:
@oznetmaster IConvertible or TypeConverter both would require me to change the definition of `Foo` which may affect the entire shipping product over something that is a development-only and testing-only concern: how to parse test parameters.

jnm2:
This needs to work with a simple ad-hoc conversion function.

oznetmaster:
Why would TypeConverter affect the the entire shipping product?  It is a simple attribute on the class.

It has no affect unless GetTypeConverter is called, which can be done even if no TypeConverter is declared for the class.

Does your product use GetTypeConverter elsewhere?

jnm2:
@oznetmaster
> Why would TypeConverter affect the the entire shipping product? 

Because I'm now maintaining test code that belongs in an ad-hoc function in a test assembly in the main product, which does use GetTypeConverter.

I could write a custom wrapper class, but in any case it's just a ton of coding. I can't see myself using that over just doing this:

```c#
// Statement:
var passedValue = TestContext.Parameters.Get(name);
// Expression:
passedValue != null ? ParseFoo(passedValue) : defaultFooValue
```

oznetmaster:
Then use:

```
var value = ParseFoo (TestContext.Parameters.Get(name, defaultValue.ToString()));
```

jnm2:
oznetmaster

> Then use:
> 
> ```
> var value = ParseFoo (TestContext.Parameters.Get(name, defaultValue.ToString()));
> ```

See https://github.com/nunit/nunit/pull/2184#issuecomment-303222654

oznetmaster:
If you write ParseFoo so that it returns null if given a null argument, then:

```
var foo = ParseFoo (TestContext.Parameters.Get (name)) ?? defaultFooValue;
```

would seem to be the simplest solution.

jnm2:
@oznetmaster Nice! I wonder why I overlooked that? You are not wrong.

jnm2:
So is the consensus that this API is not beneficial?

CharliePoole:
Mainly, I just wanted to see a discussion before a decision.

I'm slightly against it. Let's say a -1 on a scale of -2 to +2. The method is not strictly needed but could be convenient for some people some of the time. It would add one more thing to the Intellisense, which is a negative if you aren't using it.

As far as consensus, I am not sure anybody else has expressed a definite opinion.

oznetmaster:
I see no reason to add it. As @jnm2 acknowledged above, my one statement solution using existing methods was totally satisfactory.

rprouse:
I'm not a huge fan. I think @oznetmaster's alternative is even cleaner than the original code that this would enable. I vote to close 👎

jnm2:
Alright, so three downvotes and I'm happy with my alternative (for the time being).

PR #2185, opened 2017-05-22
Fix tests failing on systems with CultureInfo other than en

### LINKED ISSUE #2179: Some NUnit project's tests fail on systems with CultureInfo other than en
The following tests fail on systems with comma decimal mark settings:
- NUnit.Framework.Internal.EventListenerTextWriterTests.TestWriteDecimal
- NUnit.Framework.Internal.EventListenerTextWriterTests.TestWriteDouble
- NUnit.Framework.Internal.EventListenerTextWriterTests.TestWriteStringArg2
- NUnit.Framework.Internal.EventListenerTextWriterTests.TestWriteStringArg3
- NUnit.Framework.Constraints.LessThanConstraintTests.SimpleTolerance_Failure

The following test fails on systems with CultureInfo other than en:
- NUnit.Framework.Internal.EventListenerTextWriterTests.TestWriteStringArg1

Steps to reproduce:
1. Change system region, or language settings to one with comma decimal mark ([Countries using Arabic numerals with decimal comma](https://en.wikipedia.org/wiki/Decimal_mark#Countries_using_Arabic_numerals_with_decimal_comma)), and with language other than English
2. Run the build script with `-Target Test` option
3. See aforementioned tests fail

gcichosz:
Some tests contain comparisons to strings with hardcoded dot decimal mark in double numbers, which fails on systems using comma decimal mark. One test contains comparison to string with hardcoded month abbreviation, which fails on systems using language other than English. Since it's known what is expected in those tests, changing those values to take culture info into consideration fixes the issue.

Fixes #2179

### DIFF
--- a/src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
+++ b/src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal.Execution;
@@ -34,6 +35,7 @@ namespace NUnit.Framework.Internal
     {
         private static readonly string STREAM_NAME = "EventListenerTextWriterTestsStream";
         private static readonly string NL = Environment.NewLine;
+        private readonly string decimalSeparator = CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator;
 
         TestListenerIntercepter ListenerResult;
         TextWriter ListenerWriter;
@@ -95,7 +97,7 @@ namespace NUnit.Framework.Internal
             ListenerWriter.Write(format, arg0, arg1);
             ListenerWriter.WriteLine(format, arg0, arg1);
 
-            var expected = "05.00 @";
+            var expected = $"05{decimalSeparator}00 @";
             Assert.That(ListenerResult.Outputs.Count, Is.EqualTo(2));
             Assert.That(ListenerResult.Outputs[0], Is.EqualTo(expected));
             Assert.That(ListenerResult.Outputs[1], Is.EqualTo(expected + NL));
@@ -112,7 +114,7 @@ namespace NUnit.Framework.Internal
             ListenerWriter.Write(format, arg0, arg1, arg2);
             ListenerWriter.WriteLine(format, arg0, arg1, arg2);
 
-            var expected = "Quick 9.00 Fox";
+            var expected = $"Quick 9{decimalSeparator}00 Fox";
             Assert.That(ListenerResult.Outputs.Count, Is.EqualTo(2));
             Assert.That(ListenerResult.Outputs[0], Is.EqualTo(expected));
             Assert.That(ListenerResult.Outputs[1], Is.EqualTo(expected + NL));
@@ -155,7 +157,7 @@ namespace NUnit.Framework.Internal
             ListenerWriter.Write(value);
             ListenerWriter.WriteLine(value);
 
-            var expected = "2.731";
+            var expected = $"2{decimalSeparator}731";
             Assert.That(ListenerResult.Outputs.Count, Is.EqualTo(2));
             Assert.That(ListenerResult.Outputs[0], Is.EqualTo(expected));
             Assert.That(ListenerResult.Outputs[1], Is.EqualTo(expected + NL));
@@ -169,7 +171,7 @@ namespace NUnit.Framework.Internal
             ListenerWriter.Write(value);
             ListenerWriter.WriteLine(value);
 
-            var expected = "-1.5";
+            var expected = $"-1{decimalSeparator}5";
             Assert.That(ListenerResult.Outputs.Count, Is.EqualTo(2));
             Assert.That(ListenerResult.Outputs[0], Is.EqualTo(expected));
             Assert.That(ListenerResult.Outputs[1], Is.EqualTo(expected + NL));
--- a/src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
+++ b/src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
@@ -81,7 +81,7 @@ namespace NUnit.Framework.Internal
             ListenerWriter.Write(format, arg0);
             ListenerWriter.WriteLine(format, arg0);
 
-            var expected = "20 Apr 2017";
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
```diff
@@ -79,7 +81,7 @@ public void TestWriteStringArg1()
             ListenerWriter.Write(format, arg0);
             ListenerWriter.WriteLine(format, arg0);
 
-            var expected = "20 Apr 2017";
+            var expected = $"20 {CultureInfo.CurrentCulture.DateTimeFormat.GetAbbreviatedMonthName(4)} 2017";
```
jnm2: Why `CultureInfo.CurrentCulture` here and `CultureInfo.InvariantCulture` in LessThanConstraintTests?

src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
```diff
@@ -95,7 +97,7 @@ public void TestWriteStringArg2()
             ListenerWriter.Write(format, arg0, arg1);
             ListenerWriter.WriteLine(format, arg0, arg1);
 
-            var expected = "05.00 @";
+            var expected = $"05{decimalSeparator}00 @";
```
jnm2: For these can we do `$"{5:00.00} @"` instead?

jnm2: If this is supposed to be CurrentCulture and not InvariantCulture, an we do `$"{5:00.00} @"` for these instead?

src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
jnm2: If this is supposed to be CurrentCulture and not InvariantCulture, can we do `arg0.ToString("dd MMM yyyy")` or `$"{arg0:dd MMM yyyy}"` instead?

GCichosz: `CultureInfo.InvariantCulture` here would still return `"20 Apr 2017"`, and that's not what a `Date` formatted with `"{0:dd MMM yyyy}"` returns on non-English system. In LessThanConstraintTests the `Assert` exception message is formatted using dot decimal mark, so using `CultureInfo.InvariantCulture` ensures that the expected message string contains dot decimal mark formatted double.

src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
GCichosz: Good point, that's a lot simpler and nicer :smile:

src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
jnm2: Okay cool. So in the same spirit as the changes below what about `$"{arg0:dd MMM yyyy}"`?

CharliePoole:
I think we need to back up a moment and decide how we want this to work. When I read the original issue, I assumed that it was the writer that was broken, not the tests. So... if a user is writing output from a test, what CultureInfo should be used? My assumption is that it should be the culture of the current thread, which defaults to the system culture unless the user changes it with an attribute.

But, whatever we think it should be, let's define it and make sure the tests reflect that.

BTW, this is different from NUnit internal writes, which always use the InvariantCulture.

src/NUnitFramework/tests/Internal/EventListenerTextWriterTests.cs
GCichosz: Also changed ☺️

GCichosz:
@jnm2  Turns out there was no need to use `InvariantCulture` at all. The bug was that `tolerance.ToString()` was used instead of `MsgUtils.FormatValue(tolerance)` like in the rest of that fixture.

@CharliePoole Checking that would require a little digging inside the writer, but it seems like it uses the `InvariantCulture` unless it's supplied with another format. Anyway that looks to me, like a separate issue to confirm.

ChrisMaddock:
> So... if a user is writing output from a test, what CultureInfo should be used? My assumption is that it should be the culture of the current thread, which defaults to the system culture unless the user changes it with an attribute.

That is what is happening here, isn't it? And @GCichosz's changes are just fixing a bug in this behaviour? Have I missed something, @CharliePoole?

@rprouse - can we squeeze this into 3.7, if we get it resolved in time? Not to worry if you've already set up the changenotes and stuff.

GCichosz:
@ChrisMaddock My changes aren't fixing any application behaviour, they change unit tests only, so they pass also on systems with culture different than en-*.

The writer does function properly. The exception might be exceptions, because their messages are formatted with `InvariantCulture`, but that would require a confirmation. In `LessThanConstraintTests` we check exception messages with help of `MsgUtils` and it uses `InvariantCulture` all the way, so we expect `InvariantCulture` in those tests. Question is, do we want exception messages to be formatted using `InvariantCulture`?

Lette:
I have a question regarding the proposed fix. Isn't it a discrepancy that the `MsgUtils` methods use `CultureInfo.InvariantCulture` and thus makes the assertion messages culture-invariant, while the `EventListenerTextWriter` produces culture-dependent output? Are those outputs used in different ways, so that this difference in behavior is what we'd expect?

CharliePoole:
Is this waiting for me for some reason? I thought that @jnm2 and @ChrisMaddock had approved.

@Lette `MsgUtils` is for messages created by NUnit, while `EventListenerTextWriter` handles user output. It's not impossible, however, that they have been used at cross purposes over time, which would be a bug. If so, it's a separate issue, however, since this issue is only about our own tests passing in different environments.

Lette:
@CharliePoole Thanks. So we've established that the production code is correct, and that the tests were wrong. I was just afraid that it could've been the other way, but I can see the rationale behind keeping the `TextWriter` culture-dependent. Good info.

This PR fixes the same seven tests that I had originally failing when looking at #2213.

CharliePoole:
@jnm2 @ChrisMaddock Didn't you guys approve this? Normally, the second guy merges. I'm merging now.

ChrisMaddock:
> Is this waiting for me for some reason?

> @jnm2 @ChrisMaddock Didn't you guys approve this? Normally, the second guy merges. I'm merging now.

I was waiting for your response - it looked good to me, but you seemed to suggest in https://github.com/nunit/nunit/pull/2185#issuecomment-303233395 that we needed to back up and decide how things should work before merging. Never mind, it's merged now. 😄

PR #2187, opened 2017-05-22
Replace special characters as part of converting branch names to package versions

### LINKED ISSUE #2186: Replace special characters as part of converting branch names to package versions
Branches with underscores in their names currently fail the build.

http://semver.org/#spec-item-9:
> Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]

https://github.com/nunit/nunit/blob/master/build.cake#L157

jnm2:
Fixes #2186


ChrisMaddock:
Looks good to me. 👍

PR #2197, opened 2017-05-28
Added the Distinct option on the Random attribute.

### LINKED ISSUE #2144: Allow option on RandomAttribute to produce distinct values
RandomAttribute produces not distinct values. So NUnit Adapter 3.7.0.0 tells errors that test with the same name does already exists.

`public void MyTest([Random(0, 9, 7)] int x)`

The range is small in this case, so it is nearly sure to get the same value twice.
I expect the values should be distinct. If the range is too small the number of random numbers to produce shopuld be reduced.

jp2masa:
Fixes #2144

### DIFF
--- a/src/NUnitFramework/framework/Attributes/RandomAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RandomAttribute.cs
@@ -25,6 +25,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using System.Reflection;
 using NUnit.Compatibility;
 using NUnit.Framework.Interfaces;
@@ -41,6 +42,25 @@ namespace NUnit.Framework
     {
         private RandomDataSource _source;
         private int _count;
+        private bool _distinct;
+
+        /// <summary>
+        /// If true, the random values will be all different
+        /// </summary>
+        public bool Distinct
+        {
+            get
+            {
+                return _distinct;
+            }
+            set
+            {
+                _distinct = value;
+
+                if (_source != null)
+                    _source.Distinct = _distinct;
+            }
+        }
 
         #region Constructors
 
@@ -189,6 +209,8 @@ namespace NUnit.Framework
                 _source = new RandomDataConverter(_source);
             }
 
+            _source.Distinct = _distinct;
+
             return _source.GetData(parameter);
 
             //// Copy the random _values into the data array
@@ -221,6 +243,7 @@ namespace NUnit.Framework
         abstract class RandomDataSource : IParameterDataSource
         {
             public Type DataType { get; protected set; }
+            public bool Distinct { get; set; }
 
             public abstract IEnumerable GetData(IParameterInfo parameter);
         }
@@ -231,6 +254,8 @@ namespace NUnit.Framework
             private T _max;
             private int _count;
             private bool _inRange;
+
+            private List<T> previousValues = new List<T>();
 
             protected Randomizer _randomizer;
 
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/framework/Attributes/RandomAttribute.cs
```diff
@@ -258,14 +283,36 @@ public override IEnumerable GetData(IParameterInfo parameter)
 
                 _randomizer = Randomizer.GetRandomizer(parameter.ParameterInfo);
 
+                if (Distinct && _inRange && !CanBeDistinct(_min, _max, _count))
+                    throw new NUnitException($"Random values can't be distinct. Method: \"{parameter.Method}\"; Parameter: \"{parameter.ParameterInfo.Name}\"");
```
CharliePoole: What if we didn't throw but just ignored distinct in this case. I don't like ignoring user errors in general, but this one seems somewhat harmless. What do other @nunit/framework-team folks think?

If we do want a message, then 
1. It should be more user friendly, like "Cannot generate 5 distinct values for X because the range only contains 4 values" or "Cannot generate 5 distinct values from 1 to 3 for X."

2. We generally use standard exceptions in situations like this. See `Guard.ArgumentValid`. 

3. `NUnitException` is only used as the base class of our own exceptions and to wrap exceptions in certain cases.

src/NUnitFramework/framework/Attributes/RandomAttribute.cs
```diff
@@ -41,6 +42,25 @@ public class RandomAttribute : DataAttribute, IParameterDataSource
     {
         private RandomDataSource _source;
         private int _count;
+        private bool _distinct;
+
+        /// <summary>
+        /// If true, the random values will be all different
```
jnm2: Just an idea; how about `If true, no value will be repeated.`?

src/NUnitFramework/framework/Attributes/RandomAttribute.cs
```diff
@@ -41,6 +42,25 @@ public class RandomAttribute : DataAttribute, IParameterDataSource
...
+            get
+            {
+                return _distinct;
+            }
+            set
+            {
+                _distinct = value;
+
+                if (_source != null)
+                    _source.Distinct = _distinct;
```
jnm2: Is there any reason to set this here since it's always being set just in time in `GetData`? I'd rather take Occam's razor to this if possible.

src/NUnitFramework/framework/Attributes/RandomAttribute.cs
jnm2: I'm always in favor of failures as soon and loud as possible when the end user asks for something inherently contradictory.

src/NUnitFramework/framework/Attributes/RandomAttribute.cs
```diff
@@ -221,6 +243,7 @@ private bool WeConvert(Type sourceType, Type targetType)
         abstract class RandomDataSource : IParameterDataSource
         {
             public Type DataType { get; protected set; }
+            public bool Distinct { get; set; }
```
jnm2: I would have preferred making this `public bool Distinct { get; }` and only settable through the constructor, so that the same instance can't be used for both distinct and non-distinct generation. It's not entirely clear what the behavior should be if you turn this off, generate some numbers, and turn it back on again and generate more. I don't know how big of an issue this is though.

src/NUnitFramework/framework/Attributes/RandomAttribute.cs
jp2masa: There's a constructor which only takes one param, `int count`, and source only gets created on GetData, and there are other constructors, with `int min` and `int max`, where the source is created, before the property Distinct gets set.

src/NUnitFramework/framework/Attributes/RandomAttribute.cs
jp2masa: If no exception is thrown, then there will be an infinite loop when generating random values. the infinite loop can be avoided, checking if this situation occurs.

ChrisMaddock:
Argh! Sorry - I fat-fingered the merge button while looking on an app - that's dangerous! 

I'll revert this now, @jp2masa - can you open a new PR with the changes in? Sorry!

ChrisMaddock:
https://github.com/nunit/nunit/pull/2198 reverts this accidental merge, I'll merge that as soon as CI has passed. Sorry everyone... 👎

src/NUnitFramework/framework/Attributes/RandomAttribute.cs
jnm2: Yes, but what I'm saying is [this](https://github.com/jp2masa/nunit/blob/63169cfea2a27a1dad58bc6b97ece92df1c26bfe/src/NUnitFramework/framework/Attributes/RandomAttribute.cs#L214) is the only place I can see the source is used and [this](https://github.com/jp2masa/nunit/blob/63169cfea2a27a1dad58bc6b97ece92df1c26bfe/src/NUnitFramework/framework/Attributes/RandomAttribute.cs#L212) is guaranteed to happen. If you remove lines 60 and 61, I still can't think of any way you could get any data out of the source without causing line 212 to run. What am I missing?

jp2masa: You're right, I hadn't noticed that.

PR #2199, opened 2017-05-29
Added the Distinct option on the Random attribute

### LINKED ISSUE #2144: Allow option on RandomAttribute to produce distinct values

jp2masa:
Fixes #2144

### DIFF
--- a/src/NUnitFramework/framework/Attributes/RandomAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RandomAttribute.cs
@@ -25,6 +25,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using System.Reflection;
 using NUnit.Compatibility;
 using NUnit.Framework.Interfaces;
@@ -41,6 +42,11 @@ namespace NUnit.Framework
     {
         private RandomDataSource _source;
         private int _count;
+
+        /// <summary>
+        /// If true, no value will be repeated.
+        /// </summary>
+        public bool Distinct { get; set; }
 
         #region Constructors
 
@@ -186,8 +192,11 @@ namespace NUnit.Framework
             }
             else if (_source.DataType != parmType && WeConvert(_source.DataType, parmType))
             {
+                _source.Distinct = Distinct;
                 _source = new RandomDataConverter(_source);
             }
+
+            _source.Distinct = Distinct;
 
             return _source.GetData(parameter);
 
@@ -208,7 +217,7 @@ namespace NUnit.Framework
 
             if (targetType == typeof(decimal))
                 return sourceType == typeof(int) || sourceType == typeof(double);
-            
+
             return false;
         }
 
@@ -221,6 +230,7 @@ namespace NUnit.Framework
         abstract class RandomDataSource : IParameterDataSource
         {
             public Type DataType { get; protected set; }
+            public bool Distinct { get; set; }
 
             public abstract IEnumerable GetData(IParameterInfo parameter);
         }
@@ -231,6 +241,8 @@ namespace NUnit.Framework
             private T _max;
             private int _count;
             private bool _inRange;
+
+            private List<T> previousValues = new List<T>();
 
             protected Randomizer _randomizer;
 
@@ -258,14 +270,36 @@ namespace NUnit.Framework
 
                 _randomizer = Randomizer.GetRandomizer(parameter.ParameterInfo);
 
+                Guard.OperationValid(!(Distinct && _inRange && !CanBeDistinct(_min, _max, _count)), $"The range of values is [{_min}, {_max}[ and the random value count is {_count} so the values cannot be distinct.");
+                
+
                 for (int i = 0; i < _count; i++)
-                    yield return _inRange
-                        ? GetNext(_min, _max)
-                        : GetNext();
+                {
+                    if (Distinct)
+                    {
+                        T next;
+
+                        do
+                        {
+                            next = _inRange
+                                ? GetNext(_min, _max)
+                                : GetNext();
+                        } while (previousValues.Contains(next));
+
+                        previousValues.Add(next);
+                        
+                        yield return next;
+                    }
+                    else
+                        yield return _inRange
+                            ? GetNext(_min, _max)
+                            : GetNext();
+                }
             }
 
             protected abstract T GetNext();
             protected abstract T GetNext(T min, T max);
+            protected abstract bool CanBeDistinct(T min, T max, int count);
         }
 
         #endregion
@@ -330,6 +364,11 @@ namespace NUnit.Framework
             {
                 return _randomizer.Next(min, max);
             }
+
+            protected override bool CanBeDistinct(int min, int max, int count)
+            {
+                return count <= max - min;
+            }
         }
 
         #endregion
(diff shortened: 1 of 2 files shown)


CharliePoole:
All your tests are in test-assembly rather than nunit.framework.tests, so they are not getting executed.

I don't know what is happening with the build. There seems to be an error in the cake script after everything else is done. Did someone just change the script?

jp2masa:
I made some requested changes between the other pull request and this one, and I probably missed something, so I'll continue the work on this tomorrow.

src/NUnitFramework/framework/Attributes/RandomAttribute.cs
```diff
@@ -42,6 +43,11 @@ public class RandomAttribute : DataAttribute, IParameterDataSource
         private RandomDataSource _source;
         private int _count;
 
+        /// <summary>
+        /// If true, no value will be repeated
```
jnm2: Nit: the general style is to end both sentences (this is one) and fragments with a period. I don't know if this is in the coding standard but NUnit mostly follows it and the .NET BCL always follows it.

src/NUnitFramework/framework/Attributes/RandomAttribute.cs
```diff
@@ -258,14 +269,38 @@ public override IEnumerable GetData(IParameterInfo parameter)
 
                 _randomizer = Randomizer.GetRandomizer(parameter.ParameterInfo);
 
+                Guard.OperationValid(Distinct && _inRange && !CanBeDistinct(_min, _max, _count), "");
+
+                if (Distinct && _inRange && !CanBeDistinct(_min, _max, _count))
+                    throw new NUnitException($"Random values can't be distinct. Method: \"{parameter.Method}\"; Parameter: \"{parameter.ParameterInfo.Name}\"");
```
jnm2: @CharliePoole had asked for a better message?

src/NUnitFramework/framework/Attributes/RandomAttribute.cs
```diff
@@ -542,8 +629,27 @@ public override IEnumerable GetData(IParameterInfo parameter)
                 Randomizer randomizer = Randomizer.GetRandomizer(parameter.ParameterInfo);
                 DataType = parameter.ParameterType;
 
+                if (Distinct && _count > Enum.GetValues(DataType).Cast<int>().Distinct().Count())
+                    throw new NUnitException($"Random values can't be distinct. Method: \"{parameter.Method}\"; Parameter: \"{parameter.ParameterInfo.Name}\"");
```
jnm2: @CharliePoole had asked for a better message?

src/NUnitFramework/framework/Attributes/RandomAttribute.cs
CharliePoole: Also, not to use an NUnitException in this way. Use one of the Guard methods.

src/NUnitFramework/framework/Attributes/RandomAttribute.cs
jp2masa: I had looked at other source files and they didn't have a period in the end of the sentences, so I opted not to write it, but I'll write it.

ChrisMaddock: I've wondered this recently too! Good to have a standard to go by. 😄

jnm2:
Looks good other than what @CharliePoole already said about putting the tests in the `nunit.framework.tests` projects.

jp2masa:
I didn't understand what is nunit.testdata. I saw the RandomAttribute tests implemented there, so I implemented the new ones there too.

jnm2:
Sorry, it's pretty meta 😆 `nunit.framework.tests` is where all the tests go that you want to directly run to test the NUnit framework, and `nunit.testdata` is where all the tests go that need to be run by tests in `nunit.framework.tests` that are verifying _how_ tests are run.

Okay. I can't tell if I helped or made it worse. 😆

CharliePoole:
Let me try too!

Some tests need tests as data. Those "data tests" can't go into nunit.framework.tests because many of them are expected to fail. So we put them in a separate assembly that is referenced by our tests. Those tests are never run, except under control of our own tests. 

It doesn't look as if you need to have any second-level tests, so yours should go into nunit.framework.tests, which also contains the __real__ RandomAttribute tests.

BTW, if you happen to be using the VS adapter, then those second-level tests will be discovered and run directly. That's why we don't use it to develop NUnit itself.

src/NUnitFramework/framework/Attributes/RandomAttribute.cs
```diff
@@ -542,8 +628,28 @@ public override IEnumerable GetData(IParameterInfo parameter)
                 Randomizer randomizer = Randomizer.GetRandomizer(parameter.ParameterInfo);
                 DataType = parameter.ParameterType;
 
-                for (int i = 0; i < _count; i++ )
-                    yield return randomizer.NextEnum(parameter.ParameterType);
+                int valueCount = Enum.GetValues(DataType).Cast<int>().Distinct().Count();
+
+                Guard.OperationValid(!(Distinct && _count > valueCount), $"The enum \"{DataType.Name}\" has {valueCount} values and the random value count is {_count} so the values cannot be distinct.");
```
CharliePoole: Nice message!

jp2masa:
I'm not understanding one thing, there is only one test in the [RandomAttribute tests](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/tests/Attributes/RandomAttributeTests.cs#L41), which has a TestCaseSource that consists of all the methods in RandomAttributeFixture, which is in nunit.testdata.
Why are the RandomAttribute tests implemented in nunit.testdata?

CharliePoole:
The single test method in `RandomAttributeTests` creates a bunch of test cases, one for each method in `RandomAttributeFixture`.  All the methods in `RandomAttributeFixture` use Random to generate COUNT test cases comprising a single `ParameterizedMethodFixture`. We run each of those fixtures and verify that it generates COUNT test cases and passes. Passing means that the values were all of the desired type and in the expected range.

The reason the tests were implemented this way is that we want to examine the result of running the entire `ParameterizedMethodFixture`, looking at the number of cases generated. No single test can do that, hence the second level. 

I had not noticed that the first level test uses reflection. So I see now that your new tests do actually get run. By saving up the values of all the arguments you are able to include the test for distinct values in your tests. I'm OK with this now.

CharliePoole:
@rprouse I think it's up to you whether you want to merge this for the release or wait till after.

oznetmaster:
I left a number of comments on the actual commit for this.

jnm2:
@oznetmaster I've heard people talk about GitHub commit comments. They don't notify and are hard to find because the in between space is not collapsed. Just curious, why not comment on the PR?

oznetmaster:
Because, as far as I know, there is no way to do code line comments on the PR.

jnm2:
@oznetmaster In a PR under the Files Changed tab, you can add a code line comment anywhere near a change. It doesn't let you add a comment more than three lines awey from a changed line though.

PR #2200, opened 2017-05-29
Correct platform name

ChrisMaddock:
Just spotted in a CI log that I'd used an invalid platform name here - I guess I missed the 'skipped' message before in amongst all the 'Explict' tests.

### DIFF
--- a/src/NUnitFramework/tests/TestContextTests.cs
+++ b/src/NUnitFramework/tests/TestContextTests.cs
@@ -367,7 +367,7 @@ namespace NUnit.Framework.Tests
 
         [TestCase(null)]
 #if !NETSTANDARD1_3 && !NETSTANDARD1_6
-        [TestCase("bad<>path.png", IncludePlatform = "Windows")]
+        [TestCase("bad<>path.png", IncludePlatform = "Win")]
 #endif
         public void InvalidFilePathsThrowsArgumentException(string filePath)
         {


src/NUnitFramework/tests/TestContextTests.cs
```diff
@@ -367,7 +367,7 @@ public void FilePathAndDescriptionDoesNotThrow()
 
         [TestCase(null)]
 #if !NETSTANDARD1_3 && !NETSTANDARD1_6
-        [TestCase("bad<>path.png", IncludePlatform = "Windows")]
+        [TestCase("bad<>path.png", IncludePlatform = "Win")]
```
rprouse: Going forward, I would love to see these as an enum of platforms that we support, but for backwards compatibility we could create constants for all the strings for platforms we support and start using those. Obviously not for this pull request, but I would prefer to see `Platform.Windows`.

rprouse:
Looks good, just waiting for CI to pass.

src/NUnitFramework/tests/TestContextTests.cs
ChrisMaddock: I filed https://github.com/nunit/nunit/issues/2201 about the fact this wasn't an error, but yes, PlatformAttribute definitely needs some work! An enum would be great.

ChrisMaddock:
Thanks Rob!

CharliePoole:
I'd rather see the work go toward Include and Exclude attributes, eventually replacing Platform.

PR #2214, opened 2017-06-02
Handle null or empty filter when exploring tests

### LINKED ISSUE #2205: Ncrunch: System.Xml.XmlException: Root element is missing. when adding nUnit 3.7.0
Hi,

When creating a new solution (wep api) with a unit test project I get the following error when I add nUnit 3.7.0 to the unit test project.

`An error occurred while analysing this project after it was built: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.Xml.XmlException: Root element is missing.
   at System.Xml.XmlTextReaderImpl.Throw(Exception e)
   at System.Xml.XmlTextReaderImpl.ParseDocumentContent()
   at System.Xml.XmlTextReaderImpl.Read()
   at System.Xml.XmlLoader.Load(XmlDocument doc, XmlReader reader, Boolean preserveWhitespace)
   at System.Xml.XmlDocument.Load(XmlReader reader)
   at System.Xml.XmlDocument.LoadXml(String xml)
   at NUnit.Framework.Interfaces.TNode.FromXml(String xmlText)
   at NUnit.Framework.Internal.TestFilter.FromXml(String xmlText)
   at NUnit.Framework.Api.FrameworkController.ExploreTests(ICallbackEventHandler handler, String filter)
   at NUnit.Framework.Api.FrameworkController.ExploreTestsAction..ctor(FrameworkController controller, String filter, Object handler)
`

To clarify The project will run properly in Visual Studio. The error is shown in the NCrunch test runner and therefor I cannot start TDDing properly. The problem is resolved when reverting to nUnit version 3.6.1. In this new solution I also installed specFlow and I have resharper installed in Visual Studio.

I do not know if this is an NCrunch, nUnit, VS issue. Because the problem is fixed when going back to an older version of nUnit i suspect it to be here hence this post.

rprouse:
Fixes #2205

Branched off the 3.7 release branch and merging back into that branch in preparation for the hotfix. Once the 3.7.1 release is out, I will merge those changes back into master in another PR.

### DIFF
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -336,8 +336,6 @@ namespace NUnit.Framework.Api
 
         private void ExploreTests(ICallbackEventHandler handler, string filter)
         {
-            Guard.ArgumentNotNull(filter, "filter");
-
             if (Runner.LoadedTest == null)
                 throw new InvalidOperationException("The Explore method was called but no test has been loaded");
 
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -216,8 +216,6 @@ namespace NUnit.Framework.Api
         /// <returns>The XML result of exploring the tests</returns>
         public string ExploreTests(string filter)
         {
-            Guard.ArgumentNotNull(filter, "filter");
-
             if (Runner.LoadedTest == null)
                 throw new InvalidOperationException("The Explore method was called but no test has been loaded");
 
@@ -231,8 +229,6 @@ namespace NUnit.Framework.Api
         /// <returns>The XML result of the test run</returns>
         public string RunTests(string filter)
         {
-            Guard.ArgumentNotNull(filter, "filter");
-
             TNode result = Runner.Run(new TestProgressReporter(null), TestFilter.FromXml(filter)).ToXml(true);
 
             // Insert elements as first child in reverse order
@@ -275,8 +271,6 @@ namespace NUnit.Framework.Api
         /// <returns>The XML result of the test run</returns>
         public string RunTests(Action<string> callback, string filter)
         {
-            Guard.ArgumentNotNull(filter, "filter");
-
             var handler = new ActionCallback(callback);
 
             TNode result = Runner.Run(new TestProgressReporter(handler), TestFilter.FromXml(filter)).ToXml(true);
@@ -296,8 +290,6 @@ namespace NUnit.Framework.Api
         /// <param name="filter">A string containing the XML representation of the filter to use</param>
         private void RunAsync(Action<string> callback, string filter)
         {
-            Guard.ArgumentNotNull(filter, "filter");
-
             var handler = new ActionCallback(callback);
 
             Runner.RunAsync(new TestProgressReporter(handler), TestFilter.FromXml(filter));
@@ -320,8 +312,6 @@ namespace NUnit.Framework.Api
         /// <returns>The number of tests</returns>
         public int CountTests(string filter)
         {
-            Guard.ArgumentNotNull(filter, "filter");
-
             return Runner.CountTestCases(TestFilter.FromXml(filter));
         }
 
@@ -344,8 +334,6 @@ namespace NUnit.Framework.Api
 
         private void RunTests(ICallbackEventHandler handler, string filter)
         {
-            Guard.ArgumentNotNull(filter, "filter");
-
             TNode result = Runner.Run(new TestProgressReporter(handler), TestFilter.FromXml(filter)).ToXml(true);
 
             // Insert elements as first child in reverse order
@@ -358,8 +346,6 @@ namespace NUnit.Framework.Api
 
         private void RunAsync(ICallbackEventHandler handler, string filter)
         {
-            Guard.ArgumentNotNull(filter, "filter");
-
             Runner.RunAsync(new TestProgressReporter(handler), TestFilter.FromXml(filter));
         }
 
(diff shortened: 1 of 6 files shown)


src/NUnitFramework/framework/Api/FrameworkController.cs
```diff
@@ -336,8 +336,6 @@ private void LoadTests(ICallbackEventHandler handler)
 
         private void ExploreTests(ICallbackEventHandler handler, string filter)
         {
-            Guard.ArgumentNotNull(filter, "filter");
-
```
CharliePoole: Removing this is not necessary for @Remcomulder 's case.

rprouse: I know that, but if we are handling an empty string, why not handle null? Why force people to pass in an empty string?

CharliePoole:
I thought we were fixing an unfortunate error, which is that we didn't guard against empty as well as null. You're doing a more global fix than I imagined. I'm OK with that if it's what you prefer. Don't we still guard the other entry points though? Will you change the API documentation?

src/NUnitFramework/framework/Internal/TestFilter.cs
```diff
@@ -125,6 +125,9 @@ protected virtual bool MatchDescendant(ITest test)
         /// </summary>
         public static TestFilter FromXml(string xmlText)
         {
+            if (string.IsNullOrEmpty(xmlText))
```
jnm2: What is your opinion on `IsNullOrWhiteSpace`, even?

src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
```diff
@@ -177,11 +177,13 @@ public void ExploreTestsAction_FilterCategory_ReturnsTests()
             Assert.That(result.SelectNodes("test-suite").Count, Is.GreaterThan(0), "Explore result should have child tests");
         }
 
-        [Test]
-        public void ExploreTestsAction_AfterLoad_ReturnsRunnableSuite()
+        [TestCase(null)]
+        [TestCase("")]
+        [TestCase(EMPTY_FILTER)]
```
jnm2: `[TestCase("\r\n")]` too, for `IsNullOrWhiteSpace`?

src/NUnitFramework/tests/Internal/Filters/EmptyFilterTests.cs
```diff
@@ -35,10 +35,12 @@ public void IsEmpty()
             Assert.That(TestFilter.Empty.IsEmpty);
         }
 
-        [Test]
-        public void BuildFromXml()
+        [TestCase(null)]
+        [TestCase("")]
+        [TestCase("<filter/>")]
```
jnm2: Same [`[TestCase("\r\n")]` too, for `IsNullOrWhiteSpace`?]

rprouse:
I would prefer to do a wider fix. I think we should be able to pass in null to all the methods. I will do that assuming nobody objects.

I am happy to update the documentation.

src/NUnitFramework/framework/Internal/TestFilter.cs
rprouse: I tend to prefer to use `IsNullOrWhitespace`, but it isn't available in .NET 2.0. I could write an extension in compatibility. What do people think? Worth it for a hotfix?

jnm2: If you think it's an improvement, which I'm leaning towards, you're already in here anyway...

CharliePoole:
I think white space is definitely too far for an sang that's supposed to be valid XML. For that matter, empty should be unacceptable except for the one place that we're trying to help out @Remcomulder. I like to see errors given as soon as possible.

jnm2:
But inconsistent errors? `""` is fine but we'll blow up on `"\r\n"`? :-/

CharliePoole:
I'm OK with a wider fix WRT null. As I wrote early on, we used to do this in some older V2 versions as well. My only comment is that changing the API is not what we set out to do for the hotfix. It's not a bug fix at all but an enhancement. However, I can pass on that if you'd like to get it in.

OTOH, I think that accepting a __non-null__ string that isn't valid XML is a bad idea if you do it generally. I would prefer to see us substitute `<filter/>` for the empty string in the one location where @remcomulder needs it. All that said, I'm not reviewing here, just commenting.

CharliePoole:
@jnm2 I don't think `""` is fine at all. @remcomulder used an invalid argument that we neglected to give an error on. We like him so we want to make a very small fix to support him, without telling anybody about it in the docs. At least that's what I'd prefer.

This is a case where, if we stuck to principles, we wouldn't fix it at all. But sometimes we have to be a bit flexible when people have come to rely on  a bug.

Null, as I said, is a different matter, beyond the scope of what we set out to do. If @rprouse wants to do it that's OK with me.

jnm2:
I see your point. I'm just afraid person Y will come along and pass in the contents of an empty file and figure it works and forget about it, and down the road `\r\n` gets added to the file. But I guess that's far-fetched. So I won't maintain that argument.

CharliePoole:
@jnm2 Well thanks for that! 😄 

Bear in mind that this is an API used by drivers and and the odd third-party developer. It's not for users. If a runner accepts filter info from a user, then it has to create a properly formed XML string to pass us. For example, the console runner will accept a `--where` clause with `\r\n` but it doesn't send that to the driver, it creates XML to do it. Heck, we even provide a service in the engine to create the XML for them.

I suppose some runner might let users actually create their own XML, but it seems far-fetched. However, if we wanted to support that, we would have to be able to parse XML containing newlines, which we don't
currently do.

For this sort of interface, I think you get fewer bugs if you are tough in validating the input args.

rprouse:
I have pushed the wider changes to allow null or empty filters on all `FrameworkController` actions. Please give it a quick review. I did not allow whitespace. I tend to use `IsNullOrWhitespace` whereever possible just to be safe, but in this case it is unneeded as @CharliePoole says.

remcomulder:
Looks good to me. Thanks again for the fix.

CharliePoole:
Looks good to me too.

PR #2215, opened 2017-06-03
Ensure all work finishes before shutdown

### LINKED ISSUE #2143: Non-parallel fixture with parallel children runs in parallel with other fixtures

### LINKED ISSUE #2209: NUnit occasionally hangs when parallelizable TestFixture has OneTimeSetUp and OneTimeTearDown
This issue happens sporadically and appears to be a regression in NUnit Framework 3.7.  I can reproduce it within ~10 tries with NUnit 3.7, but I cannot reproduce it with NUnit 3.6.1.  I have reproduced it with the console runner and the Visual Studio runner.  It is most easily reproduced with a batch file that calls the console runner in a loop.

I have a TestFixture with a `OneTimeSetUp` method, a `OneTimeTearDown` method, and a single `Test` method.  I have the TestFixture marked with `[Parallelizable(ParallelScope.Self | ParallelScope.Children)]` (which looks silly in this simple example, but it makes more sense in my larger project.)  Occasionally NUnit will hang after executing the Test, but before executing the OneTimeTearDown.  I was able to reproduce it in a project that contains this as the only class:
```csharp
namespace Tests
{
    using System.Diagnostics;
    using NUnit.Framework;

    [TestFixture]
    [Parallelizable(ParallelScope.Self | ParallelScope.Children)]
    public class TestFixture
    {
        [OneTimeSetUp]
        public void OneTimeSetUp()
        {
            Debug.WriteLine("OneTimeSetUp");
        }

        [OneTimeTearDown]
        public void OneTimeTearDown()
        {
            Debug.WriteLine("OneTimeTearDown");
        }

        [Test]
        public void Test()
        {
            Debug.WriteLine("Test");
        }
    }
}
```

CharliePoole:
Fixes #2209 (not 2143 as the branch name suggests - sorry)

To deal with the window during which all queues are empty and all workers are idle, but the teardowns for some fixtures may not yet have been scheduled, I added code to simply busy-wait on completion of the top level item. Rather than shutting down when there is no work to do, we don't shut down until there is no work AND the top level item is complete. Meanwhile, we keep trying to start a new shift to deal with the any one-time teardowns or other tasks that may be scheduled.

At some point, we may need to structure this better with more events to wait on but as a hotfix I think the busy-waiting is a better solution.

As far as testing... I set up a script that was reliably duplicating the hang reported in #2209. After the fix, I was able to rapidly run the script commands hundreds of times in a loop without a hang. No guarantee that other problems won't arise, of course.

Because this has to do with timing and race conditions, I think at least two reviews would be a good idea.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -37,6 +37,8 @@ namespace NUnit.Framework.Internal.Execution
     {
         private static readonly Logger log = InternalTrace.GetLogger("Dispatcher");
 
+        private WorkItem _topLevelWorkItem;
+
         #region Constructor
 
         /// <summary>
@@ -137,6 +139,8 @@ namespace NUnit.Framework.Internal.Execution
         /// </summary>
         public void Start(WorkItem topLevelWorkItem)
         {
+            _topLevelWorkItem = topLevelWorkItem;
+
             var strategy = topLevelWorkItem.ParallelScope.HasFlag(ParallelScope.None)
                 ? ExecutionStrategy.NonParallel
                 : ExecutionStrategy.Parallel;
@@ -238,11 +242,19 @@ namespace NUnit.Framework.Internal.Execution
             if (_savedQueueLevel > 0)
                 RestoreQueueState();
 
-            if (!StartNextShift())
+            // Shift has ended but all work may not yet be done
+            while (_topLevelWorkItem.State != WorkItemState.Complete)
             {
-                foreach (var shift in Shifts)
-                    shift.ShutDown();
+                // This will fail if there is no work - all queues empty.
+                // In that case, we just continue the loop until either
+                // a shift is started or all the work is complete.
+                if (StartNextShift())
+                    return;
             }
+
+            // All work is complete, so shutdown.
+            foreach (var shift in Shifts)
+                shift.ShutDown();
         }
 
         private bool StartNextShift()
--- a/src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
@@ -192,7 +192,7 @@ namespace NUnit.Framework.Internal.Execution
                 q.Pause();
 
             // Signal the dispatcher that shift ended
-            EndOfShift(this, EventArgs.Empty);
+            EndOfShift?.Invoke(this, EventArgs.Empty);
         }
 
         /// <summary>


jnm2:
Forgot- do you have the test available to add too?

CharliePoole:
My test involves running nunit many times in succession till it hangs.

CharliePoole:
I feel like the busy wait is a minor problem since we only do it after determining there's no work to do.

jnm2:
I could add an integration test feature similar to what we introduced here: https://github.com/nunit/nunit-console/pull/223/commits/e0bf6691be913cb7e5ca921e57f6b79c6ab19c06

That would make me feel a tiny bit safer.

CharliePoole:
I'd favor that but it may be too much for the hot fix. I didn't realize you were setting up integration tests or I would have commented. That commit is already merged, isn't it?

CharliePoole:
@rprouse I asked for two reviews on this and only have one so far, so I'm not merging. But it's ready for the merging and using in the hotfix if you agree.

jnm2:
Yeah, let's not hold up the hotfix. But can we track a todo item to create the integration test for this fix?

CharliePoole:
Yup. I'll wait for @rprouse before merging, however.

rprouse:
Sorry for the delay, I was away this weekend. This looks good. I will merge and prepare another PR cherry picking this change into the release branch for the hotfix.

PR #2218, opened 2017-06-04
Ensure all work finishes before shutdown

rprouse:
This cherry picks the fix for #2209 into the release branch for the hotfix. Changes here should match PR #2215. If they do, I will just merge myself.


rprouse:
I have checked the changes here against #2215 and they are identical. I will merge once CI passes.

PR #2221, opened 2017-06-06
Merge 3.7.1 changes back into master

rprouse:
The hotfix consisted of an update to changes, the fix for empty filters and the parallizable hang issue.

Everything is appearing here in the diff even though some of it is already in master. Not sure why that is.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,90 @@
+### NUnit 3.7 - May 29, 2017
+
+This release of NUnit expands on parallel test execution to allow test methods to
+be run in parallel. Please see the [Parallelizable Attribute](https://github.com/nunit/docs/wiki/Parallelizable-Attribute)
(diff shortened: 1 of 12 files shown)


jcansdale:
> Everything is appearing here in the diff even though some of it is already in master. Not sure why that is.

It's showing everything since `release-371-merge` originally diverged from master. Would rebasing `release-371-merge` onto master be an option?

rprouse:
> It's showing everything since release-371-merge originally diverged from master. Would rebasing release-371-merge onto master be an option?

We could try it. In this PR it isn't too bad and it is a good overview of what was included in the hotfix, so I'm not sure it is worth it.

Speaking of, @nunit/core-team or @nunit/engine-team  can we get a review?

ChrisMaddock:
Looks good to me. 👍

PR #2229, opened 2017-06-08
Correct order of OneTimeTearDown commands

### LINKED ISSUE #2225: OneTimeTearDown and Dispose Ordering
Using v3.7.0/v3.7.1 from nuget.
Run with NUnit3TestAdapter, NUnit-Gui-0.3, and nunit3-console.

I have some TestFixtures that implement IDisposable. Prior to the v3.7.0 release these test fixtures always executed the OneTimeTearDown (and TestFixtureTearDown before that) method before the TestFixture class' Dispose method. After performing a NuGet update to NUnit v3.7.0 and NUnit v3.7.1 these tests started throwing exceptions in the OneTimeTearDown method as some members had been disposed already.

I was able to reproduce this behavior with the cs-money sample project by having MoneyTest implement IDisposable, add a OneTimeTearDown method, and set breakpoints in each method. NuGet NUnit version before 3.7 execute the OneTimeTearDown method first, 3.7 and 3.7.1 execute Dispose first.

Should I be assuming an ordering of these methods?

CharliePoole:
Fixes #2225

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
@@ -220,22 +220,22 @@ namespace NUnit.Framework.Internal.Execution
         {
             TestCommand command = new EmptyTestCommand(Test);
 
-            // Dispose of fixture if necessary
-            if (Test is IDisposableFixture && typeof(IDisposable).IsAssignableFrom(Test.TypeInfo.Type))
-                command = new DisposeFixtureCommand(command);
-
-            // Create the OneTimeTearDown commands
-            foreach (SetUpTearDownItem item in setUpTearDownItems)
-                command = new OneTimeTearDownCommand(command, item);
+            // For Theories, follow with TheoryResultCommand to adjust result as needed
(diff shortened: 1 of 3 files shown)


oznetmaster:
Ok, now I am concerned.

That code looked fine.

However, the code seems inconsistent with that in MakeOneTimeSetUpCommand.

That creates commands in the order:

BeforeTestActionCommand
OneTimeSetUpCommand
ConstructFixtureCommand
ApplyChangesToContextCommand

MakeOneTimeTearDownCommand creates commands in the order:

TheoryResultCommand
AfterTestActionCommand
OneTimeTearDownCommand
DisposeFixtureCommand

I am concerned that these are not the inverse of each other.  It would seem that eithger BeforeTestActionCommand should happen after OneTimeSetUpCommand, or AfterTestActionCommand should occur after OneTimeTearDownCommand.

I am confused again :(

CharliePoole:
The order of actions depends on both the order of the list and the internal structure of each command. A command can be written to perform its action either before or after calling the nested command.

This I'd very confusing and could do with more refactoring but I wanted to do the smallest possible fix in case a hotfix is to be released.

PR #2234, opened 2017-06-08
Upgrade cake bootstrappers

halex2005:
Here is PR with lastest bootstrappers for cake:

- http://cakebuild.net/download/bootstrapper/linux
- http://cakebuild.net/download/bootstrapper/windows

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,129 +1,228 @@
+##########################################################################
+# This is the Cake bootstrapper script for PowerShell.
+# This file was downloaded from https://github.com/cake-build/resources
+# Feel free to change this file to fit your needs.
+##########################################################################
+
 <#
+
 .SYNOPSIS
 This is a Powershell script to bootstrap a Cake build.
+
 .DESCRIPTION
 This Powershell script will download NuGet if missing, restore NuGet tools (including Cake)
 and execute your Cake build script with the parameters you provide.
+
+.PARAMETER Script
+The build script to execute.
 .PARAMETER Target
 The build script target to run.
 .PARAMETER Configuration
 The build configuration to use.
 .PARAMETER Verbosity
 Specifies the amount of information to be displayed.
+.PARAMETER Experimental
+Tells Cake to use the latest Roslyn release.
 .PARAMETER WhatIf
 Performs a dry run of the build script.
 No tasks will be executed.
+.PARAMETER Mono
+Tells Cake to use the Mono scripting engine.
+.PARAMETER SkipToolPackageRestore
+Skips restoring of packages.
 .PARAMETER ScriptArgs
 Remaining arguments are added here.
+
 .LINK
 http://cakebuild.net
+
 #>
 
 [CmdletBinding()]
 Param(
+    [string]$Script = "build.cake",
     [string]$Target = "Default",
     [ValidateSet("Release", "Debug")]
     [string]$Configuration = "Release",
     [ValidateSet("Quiet", "Minimal", "Normal", "Verbose", "Diagnostic")]
     [string]$Verbosity = "Verbose",
+    [switch]$Experimental,
+    [Alias("DryRun","Noop")]
     [switch]$WhatIf,
+    [switch]$Mono,
+    [switch]$SkipToolPackageRestore,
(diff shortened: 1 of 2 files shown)


jnm2:
I thinking keeping these up to date is good. Ironically I'm in the middle of a PR at https://github.com/cake-build/resources/pull/30 to make the bootstrappers more useful 😆 but don't let that affect this.

Out of curiosity, what does updating the bootstrappers enable you do to?

halex2005:
@jnm2 Actually, I just tried to run build.sh on my ubuntu and I got error ))
After upgrading script works well

CharliePoole:
The bootstrappers are the biggest problem with Cake. At various times, different bootstrappers were available at different URLs without any indication as to which to use. It's also not clear whether the bootstrappers require a particular Cake version or (the opposite) whether a Cake version requires particular bootstrapers. Some statement about this in a readme would be very helpful!

CharliePoole:
Since it's passing our CI, I'm going to say LGTM!

jnm2:
@CharliePoole Couldn't agree more. The times I've tried to get that across to the Cake team, they agreed to disagree. Hopefully https://github.com/cake-build/resources/pull/30 will end up minimizing the bootstrapper's role and make it less likely to need to be updated.

CharliePoole:
That would be great. But at a minimum, there should be info that lets us decide whether we can safely use the latest bootstrapper. Do they guarantee the latest bootstrapper to work with all versions of Cake earlier than some version? Kind of like our guaranteeing the the NUnit V2 console to work with framework versions 2.0 through 2.6.4 and the NUnit 3 console with 3.0 through current - or 2.0 with an extension.

halex2005:
@CharliePoole 

I found these bootstrappers in cake's documentation here: [Setting Up A New Project](http://cakebuild.net/docs/tutorials/setting-up-a-new-project). [This cake-build/resources repo](https://github.com/cake-build/resources) contains lastest version of bootstrappers where they can be found and be contributed.

> It's also not clear whether the bootstrappers require a particular Cake version or (the opposite) whether a Cake version requires particular bootstrapers. Some statement about this in a readme would be very helpful!

Is that so important? As I know, bootstrappers always installs lastest cake version from nuget, you should not specify cake version. If something went wrong, it's likely already fixed or could be fixed with `cake-build/resources` repo.

CharliePoole:
Actually, the bootstrapper installs the version you specify in the tools/packages.config directory. We always specify the latest version that has worked for us because an unexpected upgrade can cause a problem. 

It's also important to keep in mind that the Cake install in your working directory does not get updated automatically when packages.config is updated. You have to delete it for that to happen.

halex2005:
@CharliePoole, despite master branch of nunit contains 0.19.5 cake version in `tools/packages`, the bootstrapper installs 0.18.0 version. I think it is because this version was hardcoded into `build.ps1`: 

```
$CakeVersion = "0.18.0"
```

Is that was intentional? New bootstrapper will honor version from `tools/packages.config` and will install v0.19.5 of cake (and it works, btw).

CharliePoole:
Are you saying someone in NUnit project did that? I can't check right now in repo.

Our practice is not to change these files from cake build.

halex2005:
Here is line: https://github.com/nunit/nunit/blob/master/build.ps1#L34
And it's appeared in [Updated Cake to 0.19.5 commit](https://github.com/nunit/nunit/commit/033dbc0da7915b25ecc27eaff41836acf896fe1e).

jnm2:
@rprouse We don't need this line, do we?

CharliePoole:
@rprouse I'm assuming you didn't actually edit the file but just replaced it with a new one from Cake, right?

rprouse:
Yes, I just dropped in the latest version from Cake. I'm not sure about this change, but the only thing I've ever changed is the version of NuGet. I was also confused the the cake version in the file was not equal to the released version of cake, but it seemed to be respecting what was in my `tools\packages.config` so I just left it.

halex2005:
Let's merge this PR?

jnm2:
Yay the simplified bootstrapper is now merged: https://github.com/cake-build/resources/pull/30

jnm2:
Not sure when it ends up in master- they don't have releases per se.

PR #2235, opened 2017-06-08
Remove trim of lines in stacktrace

### LINKED ISSUE #2208: StackFIlter trims leading spaces from each line
I noticed in the gui that stack traces from a test failure were showing up without leading spaces before the "at" on each line, whereas error failures had it. I confirmed same behavior on the console. While not terribly important, this does make the reports of combined failures and errors look odd.

I think it's due to a `Trim()` call in StackFilter.cs, which should probably simply be eliminated.

mikkelbu:
Fixes #2208


CharliePoole:
Looks good to me!

PR #2236, opened 2017-06-09
Simplified event queue for discussion relating to #2217

### LINKED ISSUE #2217: Console runner performance varies wildly depending on environmental characteristics

appel1:
I believe this fixes the regression described in #2217. 

As I wrote in the issue I don't think this breaks anything but the comments and logs doesn't make it clear to me why the Dequeue implementation was so complex so it is very possible I've missed testing such situations. 

Fixes #2217

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
@@ -140,7 +140,6 @@ namespace NUnit.Framework.Internal.Execution
                 _pumpThread = new Thread (PumpThreadProc)
                     {
                     Name = "EventPumpThread" + Name,
-                    Priority = ThreadPriority.Highest
                     };
 
                 _pumpThread.Start();
--- a/src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
@@ -210,10 +210,6 @@ namespace NUnit.Framework.Internal.Execution
                     return null;
 
                 _mreAdd.Wait(-1);
-
-                if (Interlocked.CompareExchange(ref _stopped, 0, 0) != 0)
-                    return null;
-
             } while (true);
         }
 
--- a/src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
@@ -147,7 +147,7 @@ namespace NUnit.Framework.Internal.Execution
         /* This event is used solely for the purpose of having an optimized sleep cycle when
          * we have to wait on an external event (Add or Remove for instance)
          */
-        private readonly ManualResetEventSlim _mreAdd = new ManualResetEventSlim(false);
+        private readonly AutoResetEvent _mreAdd = new AutoResetEvent(false);
 
         private int _stopped;
 
@@ -209,7 +209,7 @@ namespace NUnit.Framework.Internal.Execution
                 if (!blockWhenEmpty || Interlocked.CompareExchange(ref _stopped, 0, 0) != 0)
                     return null;
 
-                _mreAdd.Wait(-1);
+                _mreAdd.WaitOne();
             } while (true);
         }
 
--- a/src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
@@ -140,8 +140,6 @@ namespace NUnit.Framework.Internal.Execution
     /// </summary>
     public class EventQueue
     {
-        private const int spinCount = 5;
-
 //        static readonly Logger log = InternalTrace.GetLogger("EventQueue");
 
         private readonly ConcurrentQueue<Event> _queue = new ConcurrentQueue<Event>();
@@ -150,16 +148,6 @@ namespace NUnit.Framework.Internal.Execution
          * we have to wait on an external event (Add or Remove for instance)
          */
         private readonly ManualResetEventSlim _mreAdd = new ManualResetEventSlim(false);
-
-        /* The whole idea is to use these two values in a transactional
-         * way to track and manage the actual data inside the underlying lock-free collection
-         * instead of directly working with it or using external locking.
-         *
-         * They are manipulated with CAS and are guaranteed to increase over time and use
-         * of the instance thus preventing ABA problems.
-         */
-        private int _addId = int.MinValue;
-        private int _removeId = int.MinValue;
 
         private int _stopped;
 
@@ -180,26 +168,11 @@ namespace NUnit.Framework.Internal.Execution
         /// <param name="e">The event to enqueue.</param>
         public void Enqueue(Event e)
         {
-            do
-            {
-                int cachedAddId = _addId;
+            // Add to the collection
+            _queue.Enqueue(e);
 
-                // Validate that we have are the current enqueuer
-                if (Interlocked.CompareExchange(ref _addId, cachedAddId + 1, cachedAddId) != cachedAddId)
-                    continue;
-
-                // Add to the collection
-                _queue.Enqueue(e);
-
-                // Wake up threads that may have been sleeping
-                _mreAdd.Set();
-
-                break;
-            } while (true);
-
-            // Setting this to anything other than 0 causes NUnit to be sensitive
-            // to the windows timer resolution - see issue #2217 
-            Thread.Sleep(0);  // give EventPump thread a chance to process the event
+            // Wake up threads that may have been sleeping
+            _mreAdd.Set();
         }
 
         /// <summary>
@@ -224,58 +197,23 @@ namespace NUnit.Framework.Internal.Execution
         /// </returns>
         public Event Dequeue(bool blockWhenEmpty)
         {
-            SpinWait sw = new SpinWait();
-
             do
             {
-                int cachedRemoveId = _removeId;
-                int cachedAddId = _addId;
-
-                // Empty case
-                if (cachedRemoveId == cachedAddId)
-                {
-                    if (!blockWhenEmpty || _stopped != 0)
-                        return null;
-
-                    // Spin a few times to see if something changes
-                    if (sw.Count <= spinCount)
-                    {
-                        sw.SpinOnce();
-                    }
-                    else
-                    {
-                        // Reset to wait for an enqueue
-                        _mreAdd.Reset();
-
-                        // Recheck for an enqueue to avoid a Wait
-                        if (cachedRemoveId != _removeId || cachedAddId != _addId)
-                        {
-                            // Queue is not empty, set the event
-                            _mreAdd.Set();
-                            continue;
-                        }
-
-                        // Wait for something to happen
-                        _mreAdd.Wait(500);
-                    }
-
-                    continue;
-                }
-
-                // Validate that we are the current dequeuer
-                if (Interlocked.CompareExchange(ref _removeId, cachedRemoveId + 1, cachedRemoveId) != cachedRemoveId)
-                    continue;
-
-
                 // Dequeue our work item
                 Event e;
-                while (!_queue.TryDequeue (out e))
+                if (_queue.TryDequeue (out e))
                 {
-                    if (!blockWhenEmpty || _stopped != 0)
-                        return null;
+                    return e;
                 }
 
-                return e;
+                if (!blockWhenEmpty || Interlocked.CompareExchange(ref _stopped, 0, 0) != 0)
+                    return null;
+
+                _mreAdd.Wait(-1);
+
+                if (Interlocked.CompareExchange(ref _stopped, 0, 0) != 0)
+                    return null;
+
             } while (true);
         }
 


Singh400:
> I believe this fixes the regression described in #2217.

It does. I just checked it against my test suite.

CharliePoole:
My earlier comments seem to be lost.... trying again...

The extensive use of spinwait was added about a year ago as a way to avoid locking and improve throughput. Before going any further, I'd like to see @oznetmaster 's review of this change. He understands the code and why it's there better than I do.

oznetmaster:
@CharliePoole Will review when I am back at my computer, probably Monday my time (Australia).

CharliePoole:
Folks, I don't see a rush here since the release is out. Let's wait for @oznetmaster.

appel1:
Unfortunately I can't figure out why the attachment tests fail. They pass on my Windows computer.

Singh400:
> Folks, I don't see a rush here since the release is out. 

I'd prefer see a fix issued sooner rather than later. Even with a small test suite using v3.4.0 the duration of test execution is around 7.7 seconds. Whereas the same test suite takes 0.6 seconds with v3.2.1. This particular regression could add minutes to a larger test suite.

Singh400:
Is there an update on a potential solution for this regression?
> This particular regression could add minutes to a larger test suite.

I was curious so I tested this on a larger test suite I have access to;-

- NUnit v3.6.1: 350 seconds
- NUnit v3.2.1: 85 seconds

CharliePoole:
@oznetmaster Did you have a chance to review this?

rprouse:
I think this could warrant another hotfix once it is reviewed. @nunit/core-team what do you think?

@jnm2 could you also give this a quick review in case @oznetmaster doesn't get a chance to review?

CharliePoole:
@rprouse I would have merged this long ago if it were limited to fixing the issue with timing due to our use of Sleep(1). However, @appel1 did a bunch of other stuff, basically putting us back to the queue implementation we had before @oznetmaster improved it by using a non-locking approach. 

Those changes are not needed to solve the problem and I don't know that they are an improvement. I don't know why @oznetmaster hasn't been able to review this yet, but in his absence I'd prefer to eliminate those changes that are aimed at simplifying the queue implementation rather than fixing the exact problem that the issue addresses. 

@appel1 @Singh400 and @mintsoft did a great job of figuring out the cause of the problem in #2217. As the "triage guy" on this issue, I summarized what I thought we should fix right now. But in doing it @appel1 succumbed to the temptation of doing too much, IMO. 

I am not an expert in lockless concurrency management, but I do know it tends to improve performance in many cases. That's why we thought it was a good idea a year ago. I don't see what is changing our mind right now. OTOH, if the changes we identified don't solve the problem, we can always come back.

appel1:
Yes, for this issue I did. ☺️

But I also don't think the raised thread priority, sleep and code duplication of what is already in ConcurrentQueue is needed. Of course, it is very possible I'm wrong. Which is why I created this pull request to bring it up for discussion.

oznetmaster:
Sorry, I have been in transition from Australia back to the US.

I have tried to review on my iPhone, but that is almost impossible.

Should be up and running by Monday.

Much of that code was directly from Microsoft.  I need to review and see why they said it was necessary.

CharliePoole:
@oznetmaster I guessed as much, knowing your annual migratory pattern. 😄 

@appel1 The class has (or maybe used to have) concurrency issues outside the wrapped ConcurrentQueue. I suggested we just drop the sleep and the priority stuff, both of which date back to NUnit V2 but - perhaps in an excess of caution - I was less confident about removing the lockfree stuff we only just added. Note also that we are using different implementations of ConcurrentQueue in our different builds. All worrisome stuff.

CharliePoole:
My vote is to wait for @oznetmaster or, if we are in a hurry on this, only do the smaller change.

appel1:
Didn't see any mentions of working around issues in ConcurrentQueue implementations when I looked at comments or blame so I didn't consider that. Can't even imagine what the CompareExchange checks could workaround since it looks like to me that you could still get multiple threads calling both Enqueue and TryDequeue at the same time. 

But I'm fine with leaving things as is. From my very simple benchmarks on our codebase it doesn't seem to have any noticeable impact on total runtime for our tests.

mintsoft:
> My vote is to wait for @oznetmaster or, if we are in a hurry on this, only do the smaller change.

By the smaller change do you mean: https://github.com/nunit/nunit/pull/2233/files ?

CharliePoole:
On the issue, I had proposed we only drop Sleep and the setting of thread priority. That's the smaller change I meant.

CharliePoole:
Travis failure is an unrelated intermittent issue, for which I just filed a bug report.

rprouse:
If my memory serves me, the lock-less solution didn't replace a locking solution, but an unprotected queue. I may be mistaking it for one of the other multi-threaded fixes though. I believe the lock-less solution is more efficient, but I've always felt that it adds more complexity than the benefits warrant. That is why I was okay with stripping it out especially after seeing how much cleaner the code was.

Now that the issue is fixed by the other PR however, I'm no longer sure this PR is necessary. At this point, it seems to be change for change's sake.

oznetmaster:
Should be up and running by tomorrow

oznetmaster:
The code that I implemented included a suggested solution for the ABA problem.  I am searching for the original code on the Microsoft site, but have not found it yet.

For those unaware of the ABA problem: [https://en.wikipedia.org/wiki/ABA_problem](url)

oznetmaster:
Also, the reason for Sleep(1) instead of Sleep(0) was documented in many places.

Here is one: [http://joeduffyblog.com/2006/08/22/priorityinduced-starvation-why-sleep1-is-better-than-sleep0-and-the-windows-balance-set-manager/](url)

oznetmaster:
I am pretty sure the higher priority for the event pump thread was there before I added the lock-less concurrent queue implementation.

I am pretty sure that no changes were made in the EventPump code at that point in time, but I could be wrong.

oznetmaster:
A far better way (and the correct way) to handle this would be to get rid of the sleep entirely, and to use a ManualResetEventSlim which is set when an even is queued, and is reset when the event pump removes any event from the queue (does not have to be the same event).

oznetmaster:
As I look a the code, I realize that the Sleep(x) was there prior to the implementation of the lock-less queue.

I am actually not sure why it is needed at all?

Maybe @CharliePoole might remember from the prior locking implementation?

Was the concern that too many Enqueue calls might occur without the event pump thread being able to dequeue anything?

appel1:
Which is essentially what this PR is. It is still lock-less. Since we don't keep any state outside of the ConcurrentQueue there's no ABA problem as I understand it. 

Since there's only one reader thread I don't think using a ManualResetEvent makes any difference.

CharliePoole:
@oznetmaster This thread is getting pretty long I'm afraid. Anyway, we already discussed both Sleep and the Priority setting as being historical code that simply didn't get removed at the time you wor ked on it. I think the minimum change in this case should be removing them both.

I see a comment in the class "Validate that we are the current dequeuer." Does this indicate you were trying to set up the class for multiple dequeuers? This isn't a general purpose queue. It's main purpose is to take events from multiple sources and create a single stream of events, so there would never be more than one reader of the queue. That reader might potentially fan the events out to multiple or different destinations, but that is another matter. So any simplification that can be made on the dequeuing side seems desirable.

@rprouse The original implementation from V2 was an unprotected queue, since there was only ever one reader and one writer. In V2, the entire purpose of the queue was to avoid sending events on the thread running the tests, both for performance reasons and to isolate the CallContext, which could include user types not known to NUnit.

I created an initial implementation for NUnit 3 using locks, since ConcurrentQueue wasn't available to us. @oznetmaster used the ConcurrentQueue from mono for the .NET 2.0 build and added the lockless implementation.

I wouldn't want us to take a step backward here, but if there is extra complexity not needed because of the assumption of multiple dequeuers, then I'd rather see it out. Mainly, I'd like to see us get rid of the Sleep and the setting of a higher Priority for the event pump.

@appel1 For those of us reading this online, your comment starting "Which is exactly..." doesn't make a lot of sense... i.e. I can't understand which prior comment you are  referring to. An unfortunate communication gap between those who use email versus github online. 😦 

Your comment about not having (possibly no longer having) any state outside of the concurrent queue prompts this question: Do we still need this class? In what way does `ConcurrentQueue`, which we didn't have in the past, fail to meet our needs. It seems to me that the answer to that question would tell us what extra code is needed in the EventQueue. If it's nothing, them maybe we don't need the class.

appel1:
I was responding to @oznetmaster:
> A far better way (and the correct way) to handle this would be to get rid of the sleep entirely, and to use a ManualResetEventSlim which is set when an even is queued, and is reset when the event pump removes any event from the queue (does not have to be the same event).
I believe that would remove the need for the event pump thread to run at a higher priority as well.

> Your comment about not having (possibly no longer having) any state outside of the concurrent queue prompts this question: Do we still need this class? In what way does ConcurrentQueue, which we didn't have in the past, fail to meet our needs. It seems to me that the answer to that question would tell us what extra code is needed in the EventQueue. If it's nothing, them maybe we don't need the class.

We need the reset event to make the dequeueing thread block when the queue is empty.

oznetmaster:
If we were to port the Mono BlockingCollection<T> to NUnit, then all of this would be unnecessary.  Just use a BlockingCollection instead of a ConcurrentQueue.  It blocks when the queue is empty.

If that is not done, then the reset event is necessary.

jnm2:
Why doesn't Mono have BlockingCollection yet? Let's do that. :D

jnm2:
It's hard to see the actual changes because of all the merges. If each merge had been a rebase, we would have perfect clarity. 😈

appel1:
It seems I fail at git. I was trying to rebase. :(

jnm2:
Lol no worries, it happens to everyone! I usually just do `git pull -r upstream master` while I'm on the feature branch and double check that the history looks okay and then `git push -f` once I'm sure I'm not losing anything.

Git being what it is, it probably wont be hard at all to cherry pick your implementation commits onto master.

CharliePoole:
@appel1 @oznetmaster The event queue has only one client on the dequeue side: EventPump. Is it worth considering simply using ConcurrentQueue and letting EventPump block itself?

@appel1 The .NET 2.0 build of the framework provides the implementation of ConcurrentQueue itself, taken from Mono. That's the approach @oznetmaster is referring to when he talks about porting BlockingCollection.

CharliePoole:
I think we should decide what we want to do before any more coding. I just hate to see good code-effort go to waste. 😃

jnm2:
@appel1 

To see what was going on I ran `
git log issue2217 --author="johan.appelgren@gmail.com" --pretty=format:'%h %ad | %s%d [%an]' --graph --date=short`:

```
*   1732c602 2017-06-26 | Merge branch 'issue2217' of https://github.com/appel1/nunit into issue2217 (origin/issue2217, issue2217) [Johan Appelgren]
|\
| * b7ae4204 2017-06-10 | Use AutoResetEvent so that Dequeue blocks if the queue is empty. [Johan Appelgren]
| * fa3bd762 2017-06-09 | Don't return null when stopping queue since that prematurely stops the event pump thread before all events have been dispatched. [Johan Appelgren]
| * 03e90f5f 2017-06-09 | Removed thread priority change and thread sleeps and spinwaits [Johan Appelgren]
* b3fe8a6a 2017-06-10 | Use AutoResetEvent so that Dequeue blocks if the queue is empty. [Johan Appelgren]
* c9993d5a 2017-06-09 | Don't return null when stopping queue since that prematurely stops the event pump thread before all events have been dispatched. [Johan Appelgren]
* 398326e8 2017-06-09 | Removed thread priority change and thread sleeps and spinwaits [Johan Appelgren]
```

It appears that we just want to move the 3 commits 03e90f5f, fa3bd762 and b7ae4204 onto master.

So I did `git fetch --all` to make sure everything was up to date, then `git checkout upstream master -b issue2217_fixed` to start a new branch tracking off master.
Then `git cherry-pick 03e90f5f^..b7ae4204` to cherry pick the range 03e90f5f inclusive to b7ae4204 inclusive.
Then there is a merge conflict which can be resolved in VS. Seems like you'd want to take your incoming change. Then after the file is saved and staged, `git cherry-pick --continue`.
Then another merge conflict, took your incoming change again, saved and staged, then `git cherry-pick --continue`.

Then whew... we're finished! Does this set of commits look right to you? https://github.com/nunit/nunit/compare/issue2217_fixed

appel1:
@jnm2 yes

jnm2:
@appel1 If you want you can either duplicate what I did, or even just checkout `upstream/issue2217_fixed`. When you're on the fixed branch, you could do something like:
`git branch --force issue2217` (make `issue2217` the same as `issue2217_fixed`, the current head)
`git checkout issue2217`
`git push --force` (to replace all the commits in this PR with the three new cherry-picked commits)
`git branch -D issue2217_fixed` to delete your clone of the branch since `issue2217` is now pointing to the same thing

–if you want. Hopefully that's helpful.

rprouse:
I agree that we should agree on what should be done before we go any further on this. The issue was fixed by #2233. It seems that a more correct solution is to just remove the thread priority and possibly the Sleep call.

At this point, if the lockless code is working, I think we should leave it and just concentrate on the root problem.

Issue #2008, opened 2017-01-25
TestCase and Values do not correctly convert DateTime string in current culture


morgen2009:
Converting `string` into `DateTime` the `TestCase`, `Values` attributes use invariant culture `System.Globalization.CultureInfo.InvariantCulture`. I expect that when I change the `TestExecutionContext.CurrentCulture` using `SetCultureAttribute` for test / test fixture, the converter should use the new culture info. What do you think about?

The following tests fail

```c#
[TestCase("12.10.1942")]
[SetCulture("de-DE")]
public void CanConvertStringToDateTimeInCurrentCulture(DateTime dt)
{
     Assert.AreEqual(new DateTime(1942, 10, 12), dt);
}

[Test]
[SetCulture("de-DE")]
public void CanConvertStringToDateTimeInCurrentCulture([Values("12.10.1942")] DateTime dt)
{
     Assert.AreEqual(new DateTime(1942, 10, 12), dt);
}
```

pgrawehr:
It seems to dependent on the user settings, especially if the user has overridden some locale settings. 
Adding 
CultureInfo.CurrentCulture = new CultureInfo("de-DE", **false**)
at the beginning of a test has solved many similar issues for me. 
-> The fix is probably to have NUnit use this constructor for CultureInfo when applying a SetCulture attribute.

CharliePoole:
The problem here is that NUnit uses various attributes at different points in the life cycle of the test.

In your case, the TestCase, Test and Values attributes are used to create the test. At a later stage - in the case of a gui, it could be much later - the test is executed. At that point, the SetCulture attribute sets the current culture of the execution thread.

By using a string argument but making the test parameter a DateTime, you're forcing the conversion to take place at the earliest stage, when the test is created. If you used a string parameter and converted it in your test, it would work as you expect.

Alternatively, you could use a TestCaseSource attribute on a method that yields DateTimes constructed in any way you like.

morgen2009:
Thank everyone for quick response. `TestCaseSource` "just" for converting seems for me a bit overhead, I would prefer to have test parameter values closer to the test method header. As well repeating the same code in each constructor. Now my code looks like ...

```c#
[TestCase("01.02.2013")]
public void TestFeature(string dateString) {
  var date = DateTime.Parse(dateString);
}
```

[...]

PR #2240, opened 2017-06-11
Modified comparisons to pass the Turkey test.

### LINKED ISSUE #2228: Several tests for nunit fails when run in a development environment with certain regional settings
My development PC has its regional settings set to Swedish. When running `.\build -t test` in the nunit repository I get 7 test failures in each project. Worse yet, when switching to Turkish regional settings I get 48 test failures in the 4.5/4.0 projects, 46 test failures in the 3.5/2.0 projects and another 16 in the netstandard projects.

For now, this is only a problem for contributors that do not have regional settings that are English (or any invariant-like setting), and even then it's an easy fix - just change your regional settings to "English (U.S.)"  and re-run the tests.

I assume that the test suite should not have a dependency on the regional settings of the computer it's running on.

Is it safe to say that all output from nunit should be English/invariant? Some of the test failures are likely the result of the code not being explicit about culture in calls to `.ToString()` and its culture-dependent brethren. I'd guess that that culture in most (all?) cases should be `CultureInfo.InvariantCulture`.

I'm happy to help, if you so wish.

Lette:
This PR fixes #2228.

Please review it carefully. All tests are passing, as they should. I'm just a little bit worried that we might be missing a test that would expose something that would break some code, somewhere in another culture...

### DIFF
--- a/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
@@ -330,16 +330,16 @@ namespace NUnit.Framework.Constraints
 
         private bool StringsEqual(string x, string y)
         {
-            string s1 = caseInsensitive ? x.ToLower() : x;
-            string s2 = caseInsensitive ? y.ToLower() : y;
+            string s1 = caseInsensitive ? x.ToUpperInvariant() : x;
+            string s2 = caseInsensitive ? y.ToUpperInvariant() : y;
 
             return s1.Equals(s2);
         }
 
         private bool CharsEqual(char x, char y)
         {
-            char c1 = caseInsensitive ? Char.ToLower(x) : x;
-            char c2 = caseInsensitive ? Char.ToLower(y) : y;
+            char c1 = caseInsensitive ? Char.ToUpperInvariant(x) : x;
+            char c2 = caseInsensitive ? Char.ToUpperInvariant(y) : y;
 
             return c1 == c2;
         }
--- a/src/NUnitFramework/framework/Internal/StringUtil.cs
+++ b/src/NUnitFramework/framework/Internal/StringUtil.cs
@@ -41,8 +41,10 @@ namespace NUnit.Framework.Internal
         /// strB is sorted first</returns>
         public static int Compare(string strA, string strB, bool ignoreCase)
         {
-            var comparison = ignoreCase ? StringComparison.CurrentCultureIgnoreCase : StringComparison.CurrentCulture;
-            return string.Compare(strA, strB, comparison);
+            string s1 = ignoreCase ? strA.ToUpperInvariant() : strA;
+            string s2 = ignoreCase ? strB.ToUpperInvariant() : strB;
+
+            return String.CompareOrdinal(s1, s2);
         }
 
         /// <summary>
--- a/src/NUnitFramework/tests/Internal/PlatformDetectionTests.cs
+++ b/src/NUnitFramework/tests/Internal/PlatformDetectionTests.cs
@@ -72,7 +72,7 @@ namespace NUnit.Framework.Internal
                 bool shouldPass = false;
 
                 foreach( string platform in expected )
-                    if ( shouldPass = platform.ToLower() == testPlatform.ToLower() )
+                    if (shouldPass = platform.ToUpperInvariant() == testPlatform.ToUpperInvariant())
                         break;
 
                 bool didPass = helper.IsPlatformSupported( testPlatform );
--- a/src/NUnitFramework/framework/Constraints/Constraint.cs
+++ b/src/NUnitFramework/framework/Constraints/Constraint.cs
@@ -174,7 +174,7 @@ namespace NUnit.Framework.Constraints
             System.Text.StringBuilder sb = new System.Text.StringBuilder();
 
             sb.Append("<");
-            sb.Append(DisplayName.ToLower());
+            sb.Append(DisplayName.ToLowerInvariant());
 
             foreach (object arg in Arguments)
             {
--- a/src/NUnitFramework/framework/Internal/PlatformHelper.cs
+++ b/src/NUnitFramework/framework/Internal/PlatformHelper.cs
@@ -164,7 +164,7 @@ namespace NUnit.Framework.Internal
 //				versionSpecification = parts[1];
 //			}
 
-            switch( platformName.ToUpper() )
+            switch (platformName.ToUpperInvariant())
             {
                 case "WIN":
                 case "WIN32":
--- a/src/NUnitFramework/tests/Constraints/EqualConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/EqualConstraintTests.cs
@@ -25,6 +25,9 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
+#if !NETSTANDARD1_3 && !NETSTANDARD1_6
+using NUnit.Framework;
+#endif
 using NUnit.Framework.Internal;
 using NUnit.TestUtilities.Comparers;
 
@@ -672,6 +675,29 @@ namespace NUnit.Framework.Constraints
             var ex = Assert.Throws<AssertionException>(() => Assert.AreEqual(expected, actual));
             Assert.That(ex.Message, Does.Match(@"\s*Expected\s*:\s*.*\s*\(.+\)\r?\n\s*But\s*was\s*:\s*.*\s*\(.+\)"));
         }
+
+        #region StringsIgnoringCaseEquality
+
+#if !NETSTANDARD1_3 && !NETSTANDARD1_6
+        public class StringsIgnoringCaseEquality
+        {
+            [Test]
+            [SetCulture("tr-TR")]
+            public void TurkishI()
+            {
+                Assert.That("I", Is.Not.EqualTo("i").IgnoreCase);
+            }
+
+            [Test]
+            [SetCulture("en-US")]
+            public void EnglishI()
+            {
+                Assert.That("I", Is.EqualTo("i").IgnoreCase);
+            }
+        }
+#endif
+
+#endregion
     }
     namespace ExampleTest.Outer.Middle.Inner.Outer.Middle.Inner.Outer.Middle.Outer.Middle.Inner.Outer.Middle.Inner.Outer.Middle.Inner.Outer.Middle.Inner.Clip {
         class ReallyLongClassNameShouldBeHere {


src/NUnitFramework/tests/Internal/PlatformDetectionTests.cs
```diff
@@ -72,7 +72,7 @@ public class PlatformDetectionTests
                 bool shouldPass = false;
 
                 foreach( string platform in expected )
-                    if ( shouldPass = platform.ToLower() == testPlatform.ToLower() )
+                    if (shouldPass = platform.ToUpperInvariant() == testPlatform.ToUpperInvariant())
```
jnm2: I think `testPlatform.Equals(platform, StringComparison.InvariantCultureIgnoreCase)` is superior because it states the actual goal and also doesn't allocate and transform temporary strings. Is there a reason we can't do this?

src/NUnitFramework/framework/Internal/StringUtil.cs
```diff
@@ -41,8 +41,10 @@ public class StringUtil
         /// strB is sorted first</returns>
         public static int Compare(string strA, string strB, bool ignoreCase)
         {
-            var comparison = ignoreCase ? StringComparison.CurrentCultureIgnoreCase : StringComparison.CurrentCulture;
-            return string.Compare(strA, strB, comparison);
+            string s1 = ignoreCase ? strA.ToUpperInvariant() : strA;
+            string s2 = ignoreCase ? strB.ToUpperInvariant() : strB;
+
+            return String.CompareOrdinal(s1, s2);
```
jnm2: What breaks if we implement this as `string.Equals(strA, strB, ignoreCase ? StringComparison.InvariantCultureIgnoreCase : StringComparison.InvariantCulture)`?

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -330,16 +330,16 @@ private bool DictionaryEntriesEqual(DictionaryEntry x, DictionaryEntry y, ref To
 
         private bool StringsEqual(string x, string y)
         {
-            string s1 = caseInsensitive ? x.ToLower() : x;
-            string s2 = caseInsensitive ? y.ToLower() : y;
+            string s1 = caseInsensitive ? x.ToUpperInvariant() : x;
+            string s2 = caseInsensitive ? y.ToUpperInvariant() : y;
 
             return s1.Equals(s2);
```
jnm2: Same [What breaks if we implement this as `string.Equals(s1, s2, caseInsensitive ? StringComparison.InvariantCultureIgnoreCase : StringComparison.InvariantCulture)`?]

jnm2:
@nunit/framework-team Please, everyone take a look at this.  I don't know the rationale behind when comparisons should be invariant culture vs current culture vs ordinal.

| | Current (/ ignore case) |
|-|-------------------------|
| `NUnitEqualityComparer.StringsEqual`, `NUnitEqualityComparer.CharsEqual` | `Ordinal`/`CurrentCultureIgnoreCase` |
| `StringUtil.Compare` | `CurrentCulture`/`CurrentCultureIgnoreCase` |
| `PlatformHelper.IsPlatformSupported` | `CurrentCulture` |

| | Proposed (/ ignore case) |
|-|--------------------------|
| `NUnitEqualityComparer.StringsEqual`, `NUnitEqualityComparer.CharsEqual` | `Ordinal`/`InvariantCultureIgnoreCase` |
| `StringUtil.Compare` | `Ordinal`/`InvariantCultureIgnoreCase` |
| `PlatformHelper.IsPlatformSupported` | `InvariantCultureIgnoreCase` |

src/NUnitFramework/tests/Internal/PlatformDetectionTests.cs
Lette: Your example works well, and that overload of `string.Equals` is in `netstandard13` so we're good. Also, all tests still pass - even while on Turkish regional settings.

My attempt only changed as little code as possible, but yours id probably better.

src/NUnitFramework/framework/Internal/StringUtil.cs
Lette: Well, `String.CompareOrdinal` returns an `int` while `String.Equals` returns a `bool`, so it'll break compilation... :-)

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
Lette: Unfortunately, `netstandard` doesn't support those `StringComparison` values.

Lette:
> Also, I'd feel better if this PR came with new tests that fail before your changes and succeed after your changes.

That's the thing about this issue. There actually were failing tests (at most 48, in the 4.0 and 3.5 projects), but only when run on a computer with non-English/non-invariant settings.

So, indeed, I had failing tests before I made the changes and no failing tests afterwards.

jnm2:
@Lette this is difficult though because we need something to fail in CI and even better, on ever contributor's computer without them changing culture settings. I guess we'd need to set up an integration test for that to work.

I'm starting to doubt that invariant-culture comparisons are really what we want. Perhaps the real fix for this needs to be `[SetCulture("en")]` on all the tests that were failing, keeping current-culture comparisons everywhere.

ChrisMaddock:
I haven't dug into this too deeply, but I think there's probably a mix here? At a guess, PlatformHelper should probablty be using Invariant culture, but NUnitEqualityComparer should be using the test's current culture?

From what I gather, comparisons within a test should run based on the Test's current culture (which may have been set to something different from the machine's default, via the SetCulture attribute.). Actual test framework code (e.g. the PlatformAttribute, which the changes to PlatformHelper here are affecting) uses Invariant Culture (Based on https://github.com/nunit/nunit/issues/2008 - although I'm not sure I personally agree with the case specified there!)

I agree this should have tests, if nothing else, there should be something to stop regression - as I doubt many of our developers will double check Turkish Culture before pushing! 😄 It would be good to have duplicate Turkish culture tests using `[SetCulture("turkey")]` where the culture genuinely broke NUnit code. I imagine in other instances - the bug will actually be in our tests, rather than the NUnit code.

I think it might be worth waiting for an authoritative response from @rprouse on this one - I imagine we may end up with a lot of back and forth otherwise, as it's not totally clear what we would want in each culture!

Lette:
A couple of the tests that were failing, failed because they compared `"Unix"` to `"UNIX"` and `"Universe"` to `"UNIVERSE"` by ignoring case through `.ToLower()` (culture-dependent, that is) and expected them to be equal. That is *not* the case in Turkish, since `"i"` is *not* the lower case of `"I"`.

I found [this page on MSDN](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#choosing-a-stringcomparison-member-for-your-method-call) that describes when to use what `StringComparison`member. Alas, all of them are not available in `netstandard`.

jnm2:
@Lette But if you wrote a test with `[SetCulture("tr-TR")]`, I'm thinking you would want `"I"` and `"i"` to be _inequal_ with a case-insensitive assertion. If you force invariant-culture comparisons, that can't happen. We should have a test on this.

I'm thinking the test with the work `UNIVERSE` needs `[SetCulture("en")]`, that's all.

Lette:
One of the places in the production code where I do believe we have a faulty behavior is with the `PlatformAttribute` . If I wrote a test on my Turkish computer that had the `[Platform("Unix")]` attribute, it wouldn't get picked up, since the `PlatformHelper.IsPlatformSupported` method does a `.ToUpper()` on the argument and then compares it to "UNIX", which will not be equal.

But I agree that there might be places where we only should be explicit with the culture of the test itself.

jnm2:
I agree about platform names.

Can you add this? I'm pretty sure we can all agree on this, but it requires current-culture comparisons to be used instead of invariant:

```c#
[Test]
[SetCulture("tr-TR")]
public void TurkishI()
{
    Assert.That("I", Is.Not.EqualTo("i").IgnoreCase);
}

[Test]
[SetCulture("en")]
public void EnglishI()
{
    Assert.That("I", Is.EqualTo("i").IgnoreCase);
}
```

Lette:
I'd love to. But where?

jnm2:
@Lette Probably https://github.com/nunit/nunit/blob/master/src/NUnitFramework/tests/Constraints/EqualConstraintTests.cs. Where is the UNIVERSE test?

But first I want to make sure that this makes sense to @nunit/framework-team.

I think it's weird that `Is.EqualTo(str)` is Ordinal and `Is.EqualTo(str).IgnoreCase` is CurrentCultureIgnoreCase. Seems it should be Ordinal and OrdinalIgnoreCase, or CurrentCulture and CurrentCultureIgnoreCase.

`Ordinal` is honestly probably the better default, contrary to the tests I posted above. We could add APIs for `Is.EqualTo(str).CurrentCulture` and `Is.EqualTo(str).CurrentCultureIgnoreCase`.

I'll wait for others to comment.

Lette:
The `UNIVERSE` test is here: https://github.com/nunit/nunit/blob/master/src/NUnitFramework/tests/Constraints/DictionaryContainsValueConstraintTests.cs#L87

ChrisMaddock:
> I think it's weird that Is.EqualTo(str) is Ordinal and Is.EqualTo(str).IgnoreCase is CurrentCultureIgnoreCase. Seems it should be Ordinal and OrdinalIgnoreCase, or CurrentCulture and CurrentCultureIgnoreCase.

Hmm. Given we allow the user to set a Culture, I feel like the comparison should respect that, and use a CultureSensitive comparison. To change that now however feels pretty fundamental, and a bit of a subtle breaking change.

I agree they should be consistent. In an ideal world, I think we'd make them both culture sensitive, but I imagine changing the behaviour of the `IgnoreCase` version would affect significantly less users.

So in conclusion - I have no conclusion. 😄

Lette:
Added the tests as requested. And as you suspected, the `tr-TR` version is now failing.

jnm2:
@ChrisMaddock I guess it makes sense to keep it even if it's a bit arbitrary. Ordinal comparisons are the 99.9% case, but once you're ignoring case, I can see why culture becomes important. XML doc comments should clearly reference `<see cref="StringComparison.Ordinal"/>` and `<see cref="StringComparison.CurrentCultureIgnoreCase"/>` then.

I'm in favor of just these changes then:
 * Platform should compare using `OrdinalIgnoreCase` (or `InvariantCultureIgnoreCase`? but why?)
    - New test that fails before we make this change no matter what your system settings are.
      If @Lette would rather not, one of us can follow up in a later PR?
 * The UNIVERSE test should have `[SetCulture("iv")]` for invariant. (Or possibly `[SetCulture("en")]` because 'universe' is an English word, but I don't think that was the point of the test.)
 * The added tests, plus this to lock in the current ordinal behavior:
   ```
   [Test, SetCulture("iv")]
   public void Ordinal_comparer_used_by_default()
   {
       Assert.That("ss", Is.Not.EqualTo("ß"));
   }
   ```

src/NUnitFramework/framework/Internal/StringUtil.cs
jnm2: @Lette My bad! `string.Compare` then...

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
jnm2: @Lette Looks like we want to keep the original behavior, so let's not make any changes at all. `string.Equals(,,)` would be an improvement but orthogonal to the purposes of this PR.

src/NUnitFramework/framework/Internal/StringUtil.cs
jnm2: Same [@Lette Looks like we want to keep the original behavior, so let's not make any changes at all. `string.Compare(,,)` would be an improvement but orthogonal to the purposes of this PR.]

Lette:
The [docs](https://msdn.microsoft.com/en-us/library/system.globalization.cultureinfo.invariantculture(v=vs.110).aspx) state that:

> You specify the invariant culture by name by using an empty string ("") [...]

Where is `"iv"` documented?

Lette:
Btw, I'm very happy to help. I'll clean up the PR according to the conclusions in this discussion.

jnm2:
@Lette https://msdn.microsoft.com/en-us/library/system.globalization.cultureinfo.aspx#Invariant:

> An invariant culture is culture-insensitive. Your application specifies the invariant culture by name using an empty string ("") or by its identifier. 

I took `TwoLetterISOLanguageName` to be its identifier, but I'm not sure. `""` should be fine.

jnm2:
@Lette 
> Btw, I'm very happy to help. I'll clean up the PR according to the conclusions in this discussion.

Great, thanks! Feel free to rebase if you like.

It'll be a little complex to add a test on the bad platform behavior that fails no matter what your system settings are. I think you'd have to add a dummy fixture in a new `PlatformFixture.cs` in `https://github.com/nunit/nunit/tree/master/src/NUnitFramework/testdata` which contains the original test that was failing, and then a new test in a new `PlatformAttributeTests.cs` in `https://github.com/nunit/nunit/tree/master/src/NUnitFramework/tests/Attributes` which uses `[SetCulture("tr")]` and checks that the `testdata` test runs properly using code similar to [this](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/tests/Attributes/TestFixtureSourceTests.cs#L50). Let us know if you get stuck.

CharliePoole:
Very general comment and question - haven't done a real review...

1. It seems as if this is fixing a bunch of things that would probably be separate PRs if not for the fact that they all came up using the Turkish culture. Should we perhaps work on one area at a time?

2. Can the errors  be duplicated using SetCulture on a test or does it actually require changing the system culture? If SetCulture will work then we can write unit tests that fail.

jnm2:
@CharliePoole 

1. Perhaps, but this is where where are; is it worth making @Lette split his PR into two?

2. My theory is that all we need is a SetCulture on a test which does a `TestBuilder.RunTest` on a test in `testdata` that has a `Platform` attribute. The tricky part will be that we will need to test both platform include `UNIX` and platform exclude `UNIX`, to observe the behavior whether or not you're on UNIX.
@Lette seriously, if you want me to take on this particular test, it's no problem.

CharliePoole:
@jnm2 

1. The reason to do extra work is if we are not confident that we can adequately review what amounts to multiple logical changes in one PR. If we think we may break something, then going slower (i.e. smaller pieces) is worth it. The extra work for someone (not necessarily @Lette) pays for the assurance that we are not breaking anything for users. However, my remark was only based on a cursory examination. I'm happy to rely on the judgement of those doing a closer review.

2. Exactly.

Lette:
@jnm2 , @ChrisMaddock , @CharliePoole 

Maybe you should count me out. I'll be travelling for the next week and I'll have limited coding time. If you can wait until next weekend with this, I can do it. Otherwise, you might be better off assigning someone else.

And, btw, fwiw,, I believe @jnm2 's theory regarding @CharliePoole 's second question is correct.

ChrisMaddock:
@Lette - I don't think there's any rush, so long as you're still willing to look at it. 🙂

Lette:
So, I'm restarting on this one.

Thought I'd start with the easy parts and promptly got stuck.

**Question:** Is it safe to say that `Wın` and `Unıx` (with Turkish small dotless i) and `WİN` and `UNİX (with Turkish large dotted I) are invalid platform names?

Right now, tests executed on Turkish Windows with `[Platform("Win")]` wont be run since `PlatformHelper.IsPlatformSupported` will do `.ToUpper()` on the name and get `WİN` which is not a match to a known platform. However, and for the same reason, tests with `[Platform("Wın")]` *will* be run.

If we "fix" this we might end up with having tests with `[Platform("Wın")]` *no longer run*... Could that be a problem?

ChrisMaddock:
> If we "fix" this we might end up with having tests with [Platform("Wın")] no longer run... Could that be a problem?

I think that's a niche enough breaking change we can go with it! 🙂 My opinion would be to go for `Win` - the invariant string.

jnm2:
I agree. I think InvariantIgnoreCase is exactly what we want. Platform names are trademarks and feel like things that should be spelled invariantly, in the same way that `while` and `async` are spelled invariantly in C#. I'd have a hard time believing this would break something.

IgnoreCase is important to me though because I'm almost sure we will break things if we require the correct capitalization. So more like VB.NET than C# in this regard. =)

CharliePoole:
So let's fix PlatformHelper. The only reason it has ToUpper is that it was written to work on .NET 1.1!

CharliePoole:
"If we only had the time" we could go through every class once every two or three years and remove the compromises due to previously supported platforms.

ChrisMaddock:
Hey @Lette - are you still interested in coming back to this? 

I think it would be good to put in some fixes to these issues, but someone would need to go through each case, and make sure we're making the correct fix.

ChrisMaddock:
Closing as this has gone stale. Happy to look at this again if anyone else wanted to pick it up.

PR #2249, opened 2017-06-13
Add KeyValuePair<K, V> to default formatters

### LINKED ISSUE #2244: Add KeyValuePair<TKey, TValue> to the default formatters
Many times the ToString representation of KeyValuePair isn't very helpful so I think it'd make sense if there was a default formatter that formatted the key and value separately using the registered formatters.

mikkelbu:
Also replaced some tabs with spaces and removed trailing spaces.

Fixes #2244

### DIFF
--- a/src/NUnitFramework/framework/Constraints/MsgUtils.cs
+++ b/src/NUnitFramework/framework/Constraints/MsgUtils.cs
@@ -24,6 +24,7 @@
 using System;
 using System.Text;
 using System.Collections;
+using System.Collections.Generic;
 using System.Globalization;
 using NUnit.Framework.Internal;
 
@@ -95,7 +96,9 @@ namespace NUnit.Framework.Constraints
 
             AddFormatter(next => val => val is string ? FormatString((string)val) : next(val));
 
-            AddFormatter(next => val => val.GetType().IsArray ? FormatArray((Array)val) : next(val));            
+            AddFormatter(next => val => val.GetType().IsArray ? FormatArray((Array)val) : next(val));
+
+            AddFormatter(next => val => TryFormatKeyValuePair(val) ?? next(val));
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/framework/Constraints/MsgUtils.cs
```diff
@@ -200,6 +205,42 @@ private static string FormatArray(Array array)
...
+        }
+
+        private static string FormatKeyValuePair(object key, object value)
+        {
+            StringBuilder s = new StringBuilder();
+            s.Append('[');
+            if (key != null)
+            {
+                s.Append(FormatValue(key));
+            }
```
jnm2: I don't think we want the `if`. `FormatValue(null)` returns `MsgUtils.Fmt_Null` which is `"null"`.

src/NUnitFramework/framework/Constraints/MsgUtils.cs
```diff
@@ -200,6 +205,42 @@ private static string FormatArray(Array array)
...
+            s.Append(", ");
+            if (value != null)
+            {
+                s.Append(FormatValue(value));
+            }
```
jnm2: Same [I don't think we want the `if`. `FormatValue(null)` returns `MsgUtils.Fmt_Null` which is `"null"`.]

src/NUnitFramework/framework/Constraints/MsgUtils.cs
```diff
@@ -200,6 +205,42 @@ private static string FormatArray(Array array)
...
+
+            object k = valueType.GetProperty("Key").GetValue(value, null);
+            object v = valueType.GetProperty("Value").GetValue(value, null);
+
+            return FormatKeyValuePair(k, v);
+        }
+
+        private static string FormatKeyValuePair(object key, object value)
+        {
+            StringBuilder s = new StringBuilder();
```
CharliePoole: sb?

src/NUnitFramework/framework/Constraints/MsgUtils.cs
CharliePoole: Agree.

src/NUnitFramework/framework/Constraints/MsgUtils.cs
CharliePoole: +1

CharliePoole: Once the ifs are gone, you can replace multiple lines with use of ApendFormat.

mikkelbu:
I've simplified the formatting as requested by @jnm2 and combined the two test methods into one.

PR #2256, opened 2017-06-15
Add testcases for formatting of char

mikkelbu:
Prior formatting of regular chars were only indirectly tested in `AllItemsAreInstancesOfTypeFailureMessage`.


CharliePoole:
It looks good to me if you can fix the conflict.

Issue #2265, opened 2017-06-18
Tests to run when opening solution file not clear


OmicronPersei:
The unit tests expected to be ran when opening the solution file is not clear.  I suggest updating the `BUILDING.md` file to make which projects' tests to run more clear.  We should probably also update the wiki to mention reading the `BUILDING.md` file for direction on which unit tests to run.

jnm2:
This has crossed my mind a number of times. It would be nice for new contributors.

This may apply to other repos too.

CharliePoole:
So, I have always thought of this as "obvious" since the actual tests are all in the tests directory. If anyone is having a problem, then it's clear I'm wrong, but since the two of you understand one another so well, could you explain where the confusion lies! 😺

jnm2:
I think this issue is now fixed, at least for this repo. https://github.com/nunit/nunit/pull/2266#issuecomment-309523146

@CharliePoole The `tests` directory structure is not visible anywhere in Visual Studio, including Solution Explorer, the test window, and ReSharper's test session tabs.

Visual Studio populates its test window and ReSharper populates its test window using all loaded projects that are test projects, including `nunit.testdata` projects. "Run all tests in solution" does just that.

A fix might be to use solution folders to group the projects into `src` and `test` project directories.

jnm2:
@CharliePoole So we have this documented, but it's still an irritation for me because my IDE makes the valid assumption that when I say "run all tests," it should directly test every test project in the solution.

Besides hitting cancel every time and using ReSharper to delete those project nodes, or changing the VSTest window view to group by projects and selecting all nodes and unselecting the testdata nodes for every run– the only way to deal with this irritation would be some attribute like [`assembly: NUnitNonTestAssembly]` together with some way to ignore that attribute when testing with `TestRunner.RunTest`.

Why not just use `build.ps1` whenever I need to run all tests?
1) context switch away from the IDE and code
2) can't double click on tests to jump to code
3) can't browse and filter nearly as nicely

jnm2:
We could have a new `NUnitIndirectTestAssembly` attribute in the `Internal` namespace. It would make no sense for end uses to use it but it would be nice for NUnit development itself.

CharliePoole:
I tried to comment on this several times but was not able to get my comments to post for some (probably network-related) reason. I really never understood what it was about until made a reference on a separate doc issue to the testdata tests. I guess you and the OP were on the same wavelength, but ideally he should have been made to spell out the problem.

[...]

PR #2266, opened 2017-06-18
Issue 2265 (better documentation for unit tests in BUILDING.md)

OmicronPersei:


BUILDING.md
```diff
@@ -15,6 +15,15 @@ or MonoDevelop.
 
 The solutions all place their output in a common bin directory under the solution root.
 
+## Unit Testing
```
jnm2: I'm thinking "Running Tests" instead because the tests aren't all unit tests.

BUILDING.md
```diff
@@ -15,6 +15,15 @@ or MonoDevelop.
...
+
+The unit tests that should be run in the solution are grouped by project name:
+ * `nunit.framework.tests-*`
+ * `nunitlite.tests-*`
+ * `slow-nunit-tests-*`
+
+Other test projects contain tests designed to fail purposefully for integration purposes.
```
jnm2: How about "fail purposely for integration tests"?

OmicronPersei: As in "Other test projects contain tests designed to fail purposely for integration tests." ?

BUILDING.md
OmicronPersei: Sounds good.

BUILDING.md
jnm2: Right.

jnm2:
Feel free to squash those commits if you want to.

rprouse:
> Feel free to squash those commits if you want to.

@jnm2 we can always choose to do that ourselves when we merge with the dropdown in GitHub.

Looks good to me.

jnm2:
@OmicronPersei I forgot, your PR should have had either `Fixes #2265` or `Closes #2265` in the body.
When a PR is merged with text like that, it automatically closes the issue with a link to the PR.

BUILDING.md
```diff
@@ -15,6 +15,15 @@ or MonoDevelop.
 
 The solutions all place their output in a common bin directory under the solution root.
 
+## Running Tests
+
+The tests that should be run in the solution are grouped by project name:
+ * `nunit.framework.tests-*`
+ * `nunitlite.tests-*`
+ * `slow-nunit-tests-*`
```
CharliePoole: Sorry to have missed this PR - I was still back on the issue, asking questions, because I didn't understanding what problem we were trying to solve

Anyway, this is wrong. The `slow-nunit-tests` assemblies are not intended to be run. They are merely data for our actual tests to use.

BUILDING.md
```diff
@@ -15,6 +15,15 @@ or MonoDevelop.
...
+
+Other test projects contain tests designed to fail purposely for integration tests.
```
CharliePoole: Not precisely true, unless you consider some of our unit tests to be integration tests. Up to now, we have merely considered them as bad unit tests. 😄 We need separate integration tests for sure.

IMO, this is the problem you will run into when documenting a moving target. 😢

jnm2: @CharliePoole what less controversial wording do you suggest? 😄

BUILDING.md
jnm2: @CharliePoole see, even I wasn't sure of that.

CharliePoole: IMO, you should know as a dev but not have to know as a user. The script knows what to test!

PR #2270, opened 2017-06-20
Fix Assembly marked with ParallelScope.None hanging

### LINKED ISSUE #2261: When an assembly is marked with ParallelScope.None and there are Parallelizable tests Nunit hangs
Using the following

```
<packages>
  <package id="NUnit" version="3.7.1" targetFramework="net452" />
  <package id="NUnit.Console" version="3.6.1" targetFramework="net452" />
  <package id="NUnit.ConsoleRunner" version="3.6.1" targetFramework="net452" />
  <package id="NUnit.Extension.NUnitProjectLoader" version="3.5.0" targetFramework="net452" />
  <package id="NUnit.Extension.NUnitV2Driver" version="3.6.0" targetFramework="net452" />
  <package id="NUnit.Extension.NUnitV2ResultWriter" version="3.5.0" targetFramework="net452" />
  <package id="NUnit.Extension.TeamCityEventListener" version="1.0.2" targetFramework="net452" />
  <package id="NUnit.Extension.VSProjectLoader" version="3.5.0" targetFramework="net452" />
</packages>
```

in a class library targetting .NET Framework 4.5.2

If I include 
`[assembly: Parallelizable(ParallelScope.None)]`

in my AssemblyInfo.cs file and then have tests like the following

```
using NUnit.Framework;

namespace ClassLibrary1
{

    namespace ClassLibrary1.Tests
    {
        [TestFixture]
        [Parallelizable(ParallelScope.Children)]
        public class Tests
        {
            [Test]
            public void Test1()
            {
                Assert.Fail();
            }

            [Test]
            public void Test2()
            {
                Assert.Fail();
            }
        }
    }
}
```

The test runner hangs and will not return, as soon as I remove the assembly attribute the tests run fine.

Repro solution here https://ufile.io/1zbgk

CharliePoole:
Fixes #2261

Note: The branch is named real-2261 because I mistakenly used issue-2261 for a previous fix.

This fixes the problems we have seen and one that I anticipated. See my inline comments.

I have tested this extensively but manually, using simple assemblies with various mixes of fixtures. I'm working on an automated way to do this as a part of a set of integration tests for the framework. For now, this fix will help with two problems, so let's get it in.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
@@ -86,7 +86,7 @@ namespace NUnit.Framework.Internal.Execution
                             Result.SetResult(ResultState.Success);
 
                             if (Children.Count > 0)
-                            {
+                            {   
                                 InitializeSetUpAndTearDownCommands();
 
                                 PerformOneTimeSetUp();
--- a/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
@@ -86,7 +86,7 @@ namespace NUnit.Framework.Internal.Execution
                             Result.SetResult(ResultState.Success);
 
                             if (Children.Count > 0)
-                            {   
+                            {
                                 InitializeSetUpAndTearDownCommands();
 
                                 PerformOneTimeSetUp();
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -82,6 +82,8 @@ namespace NUnit.Framework.Internal.Execution
 
         private void OnStartNonParallelWorkItem(TestWorker worker, WorkItem work)
         {
+            // This captures the startup of TestFixtures and SetUpFixtures,
+            // but not their teardown items, which are not composite items
             if (work is CompositeWorkItem && work.Test.TypeInfo != null)
                 IsolateQueues(work);
         }
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -304,8 +304,12 @@ namespace NUnit.Framework.Internal.Execution
                 work.Test is TestFixture && work.Context.ParallelScope.HasFlag(ParallelScope.Fixtures))
                     return ExecutionStrategy.Parallel;
 
-            // If all else fails, run on same thread
-            return ExecutionStrategy.Direct;
+            // There is no scope specified either on the item itself or in the context.
+            // In that case, simple work items are test cases and just run on the same
+            // thread, while composite work items and teardowns are non-parallel.
+            return work is SimpleWorkItem
+                ? ExecutionStrategy.Direct
+                : ExecutionStrategy.NonParallel;
         }
 
 #endregion
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -82,8 +82,8 @@ namespace NUnit.Framework.Internal.Execution
 
         private void OnStartNonParallelWorkItem(TestWorker worker, WorkItem work)
         {
-            if (work is CompositeWorkItem)
-                SaveQueueState(work);
+            if (work is CompositeWorkItem && work.Test.TypeInfo != null)
+                IsolateQueues(work);
         }
 
         #endregion
@@ -197,12 +197,12 @@ namespace NUnit.Framework.Internal.Execution
         }
 
         private object _queueLock = new object();
-        private int _savedQueueLevel = 0;
+        private int _isolationLevel = 0;
 
         /// <summary>
-        /// Save the state of the queues
+        /// Save the state of the queues and create a new isolated set
         /// </summary>
-        internal void SaveQueueState(WorkItem work)
+        internal void IsolateQueues(WorkItem work)
         {
             log.Info("Saving Queue State for {0}", work.Name);
             lock (_queueLock)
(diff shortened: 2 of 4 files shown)


src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -86,7 +86,7 @@ protected override void PerformWork()
                             Result.SetResult(ResultState.Success);
 
                             if (Children.Count > 0)
-                            {
+                            {
```
CharliePoole: Whoops! Didn't see that. 😊 I'll wait to see if there are any more serious issues.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -90,6 +90,9 @@ public WorkItem(WorkItem wrappedItem)
             Result = wrappedItem.Result;
             Context = wrappedItem.Context;
             ParallelScope = wrappedItem.ParallelScope;
+#if PARALLEL
+            TestWorker = wrappedItem.TestWorker;
+#endif
```
CharliePoole: An earlier version of this was failing due to null TestWorker property. It's no longer relevant to the change, but I thought it best to leave this in.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
```diff
@@ -82,8 +82,8 @@ public ParallelWorkItemDispatcher(int levelOfParallelism)
 
         private void OnStartNonParallelWorkItem(TestWorker worker, WorkItem work)
         {
-            if (work is CompositeWorkItem)
-                SaveQueueState(work);
+            if (work is CompositeWorkItem && work.Test.TypeInfo != null)
+                IsolateQueues(work);
```
CharliePoole: This is the actual bug fix. It was not caused by what I thought was the problem and described in the issue, although that was also a potential bug.

Specifically, this makes sure we don't create a new set of queues for anything that's not a fixture of some kind. We were doing that for the assembly, when it had `ParallelScope.None` specified.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
```diff
@@ -306,8 +304,12 @@ internal static ExecutionStrategy GetExecutionStrategy(WorkItem work)
...
                     return ExecutionStrategy.Parallel;
 
-            // If all else fails, run on same thread
-            return ExecutionStrategy.Direct;
+            // There is no scope specified either on the item itself or in the context.
+            // In that case, simple work items are test cases and just run on the same
+            // thread, while composite work items and teardowns are non-parallel.
+            return work is SimpleWorkItem
+                ? ExecutionStrategy.Direct
+                : ExecutionStrategy.NonParallel;
```
CharliePoole: This is the problem I described in the issue, although not the cause of the user-reported error.

We want to make sure that any work items representing fixtures get a set of isolated queues to run on. I kept the `ExecutionStrategy.Direct` optimization for test cases, since they don't spawn any children.

CharliePoole:
@rprouse Why does this say that a review is not required to merge? I thought we were set up to require a review on all merges to master.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
jnm2: I wish we could get rid of the branch-on-type-test. Could we add `work.GetExecutionStrategy()` someday?

CharliePoole: The dispatcher is responsible for deciding how to run a work item. Different dispatchers do it differently and in principle the work item doesn't know what those strategy choices are.

it seems more reasonable to me for the. dispatcher to know about work items since it operates on them, but I agree that a test on type is ugly.

in thus case the check is on whether the item might have children. I could use work.Test.IsSuite or give the item itself an IsComposite member.

jnm2: Since the work is not responsible to know how different dispatchers work, what information *can* the dispatcher push down to the work item level to avoid the type test? Can we give that piece of info a name and make it explicit?

CharliePoole: The only info the dispatcher could push to it that it doesn't have access to already would necessarily be information about the dispatcher.

As in the Test hierarchy, we use the **Composite** pattern here. Normally, when referring to a Test / WorkItem, you treat them all the same. But in the case of Test, it turned out we needed a member `IsSuite` on some occasions. Maybe it's the same here... we might want either `IsSimpleItem` or `IsCompositeItem`, which would have to be implemented by all WorkItems.

CharliePoole: An alternative, which I mentioned in a different comment, is to use `work.Test.IsSuite` to make the distinction. I don't particularly like this pushing through to the test, but I've done it in a few places in this method already. My thought is that it will eventually become clear what members are needed in the WorkItem in order to avoid that.

rprouse:
@CharliePoole I thought we were setup that way too. I will check and fix. I think we were having trouble with reviews when they were first released, so we might have turned it off then and never turned it back on. I will check the console project too.

I will get back to reviewing this afterwards.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
rprouse: I know that `TypeInfo` is only not null when it is a fixture, but that isn't obvious reading this code. At a minimum we need a comment, but might it be better to add a property to `Workitem` or `ITest` like `IsFixture` that wraps this test to make it more explicit? I am just thinking that I shouldn't need to read the code and understand `WorkItem` and it's derived classes to understand this if statement.

jnm2: Which could get rid of the type test too...

CharliePoole:
Refactoring turned out to be less simple than one would think, due to the fact that we have three basic kinds of items, rather than only simple and composite. In the end, I just added a comment where Rob suggested. Some refactoring should be done when somebody has a bit more time.

CharliePoole:
I guess any change cancels reviews. I added a comment! I'm merging anyway.

PR #2279, opened 2017-06-24
Allow running tests directly

### LINKED ISSUE #2242: Running tests directly should never surface a NullReferenceException
We're gathering a few bug reports around this issue. People try to manually run or debug an NUnit test and get a NullReferenceException. This sends the message to the user that this is a bug we do not know about. It would be better if we surfaced an exception with a helpful message getting the user on the right track.

Even better than that, I wish we could fall back gracefully and continue. That is going to be the user's expectation and it allows conveniences during development such as ad-hoc debugging into a test on F5 like what's manually done here:

```c#
public static class Program
{
    public static void Main() => new TheFixture().TheTest();
}

[TestFixture]
public sealed class TheFixture
{
    [Test]
    public void TheTest() => Assert.That(...);
}
```

CharliePoole:
Fixes #2242

This PR goes back to the pre-3.7 approach of creating a new TestExecutionContext when the current context is null. However, it uses a special adhoc context, that sets up non-null values in several members to avoid null reference exceptions.

This is not intended as a general fix to allow running tests directly. The executing program has no way to find out about warnings or multiple assertion failures and there is no intention to support those new features. This merely keeps things that worked before working for the user.

### DIFF
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -24,9 +24,10 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
-using System.Diagnostics;
 using System.IO;
+using System.Reflection;
 using System.Threading;
+using NUnit.Compatibility;
 using NUnit.Framework.Constraints;
 using NUnit.Framework.Interfaces;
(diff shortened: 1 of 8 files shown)


CharliePoole:
Unrelated timing error in travis-ci/pr run. Restarted.

PR #2287, opened 2017-06-29
Create Randomizer.NextGuid()

### LINKED ISSUE #2251: Randomizer.NextGuid()
Creating a random GUID is not as straightforward as setting all 128 bits randomly. Ref: https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29

https://github.com/nunit/nunit3-vs-adapter/issues/331 is an example of where someone needs random GUIDs using the repeatable randomness from `TestContext.CurrentContext.Random`.

MikeFH:
Fixes #2251 

Do you think a test checking that the version of the generated Guids is correctly set to 4 would be useful ?
Also, is there a way for me to update the doc regarding Randomizer https://github.com/nunit/docs/wiki/Randomizer-Methods ?

### DIFF
--- a/src/NUnitFramework/framework/Internal/Randomizer.cs
+++ b/src/NUnitFramework/framework/Internal/Randomizer.cs
@@ -616,7 +616,7 @@ namespace NUnit.Framework.Internal
         #region Guid
 
         /// <summary>
-        /// Returns a random <see cref="Guid"/>.
+        /// Generates a valid version 4 <see cref="Guid"/>.
         /// </summary>
         public Guid NextGuid()
         {
--- a/src/NUnitFramework/framework/Internal/Randomizer.cs
+++ b/src/NUnitFramework/framework/Internal/Randomizer.cs
@@ -611,6 +611,25 @@ namespace NUnit.Framework.Internal
             return unchecked(raw % range + min);
         }
 
+        #endregion 
+
+        #region Guid
+
+        /// <summary>
+        /// Returns a random <see cref="Guid"/>.
+        /// </summary>
+        public Guid NextGuid()
+        {
+            //We use the algorithm described in https://tools.ietf.org/html/rfc4122#section-4.4
+            var b = new byte[16];
+            NextBytes(b);
+            //set the version to 4
+            b[7] = (byte)((b[7] & 0x0f) | 0x40);
+            //set the 2-bits indicating the variant to 1 and 0
+            b[8] = (byte)((b[8] & 0x3f) | 0x80);
+            return new Guid(b);
+        }
+
         #endregion
 
         #region Helper Methods
--- a/src/NUnitFramework/tests/Internal/RandomizerTests.cs
+++ b/src/NUnitFramework/tests/Internal/RandomizerTests.cs
@@ -687,6 +687,17 @@ namespace NUnit.Framework.Internal
 
         #endregion
 
+        #region Guids
+
+        [Test]
+        [Description("Test that all generated Guids are unique")]
+        public void RandomGuidsAreUnique()
+        {
+            UniqueValues.Check(() => _randomizer.NextGuid(), 10, 10);
+        }
+
+        #endregion
+
         #region Repeatability
 
         public static class Repeatability
--- a/src/NUnitFramework/tests/Internal/RandomizerTests.cs
+++ b/src/NUnitFramework/tests/Internal/RandomizerTests.cs
@@ -696,6 +696,18 @@ namespace NUnit.Framework.Internal
             UniqueValues.Check(() => _randomizer.NextGuid(), 10, 10);
         }
 
+        [Test]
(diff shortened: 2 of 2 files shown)


jnm2:
> Do you think a test checking that the version of the generated Guids is correctly set to 4 would be useful ?

Yes please, and a comment with that test noting that we want to stay consistent with `Guid.NewGuid()`.

> Also, is there a way for me to update the doc regarding Randomizer https://github.com/nunit/docs/wiki/Randomizer-Methods ?

You've been invited to join which will give you access.
@rprouse Do we update the docs before the change is RTM?

src/NUnitFramework/framework/Internal/Randomizer.cs
```diff
@@ -611,6 +611,25 @@ public decimal NextDecimal(decimal min, decimal max)
...
+        /// </summary>
+        public Guid NextGuid()
+        {
+            //We use the algorithm described in https://tools.ietf.org/html/rfc4122#section-4.4
+            var b = new byte[16];
+            NextBytes(b);
+            //set the version to 4
+            b[7] = (byte)((b[7] & 0x0f) | 0x40);
+            //set the 2-bits indicating the variant to 1 and 0
+            b[8] = (byte)((b[8] & 0x3f) | 0x80);
```
jnm2: Thanks for spotting my error!  🎉

src/NUnitFramework/framework/Internal/Randomizer.cs
```diff
@@ -611,6 +611,25 @@ public decimal NextDecimal(decimal min, decimal max)
             return unchecked(raw % range + min);
         }
 
+        #endregion 
+
+        #region Guid
+
+        /// <summary>
+        /// Returns a random <see cref="Guid"/>.
```
jnm2: Can we add here that it's a type 4 guid just like `Guid.NextGuid()`?

CharliePoole:
I restarted AppVeyor. The failure is very suspicious. The 3.5 test run was "cancelled by user" during our test of `ThreadUtility.Kill`!

This doesn't appear to be due to this PR, but might be a recently merged change that's biting us.

jnm2:
@CharliePoole can you open an issue? Depending on which test failed I might assign myself.

CharliePoole:
Sure. FYI here is the failing build https://ci.appveyor.com/project/CharliePoole/nunit/build/3.8.0-ci-04129-pr-2287

src/NUnitFramework/framework/Internal/Randomizer.cs
MikeFH: > just like Guid.NextGuid()
I'm sorry but I don't understand what you mean by that. Do you think it is necessary to talk about `Guid.NewGuid()` here ? It seems that the guids created by `Guid.NewGuid()` always are v4 guids but I don't see that specified anywhere in the documentation, even in the documentation of `CoCreateGuid`.

jnm2: I am expecting people to want to know that it's some form of valid guid and which format. Absent information, it would seem entirely likely that someone had just done the naive thing and generated 16 random bytes and put them in a `Guid` and left it at that.

It seems like there is some value in acknowledging somehow that `Guid.NewGuid()` also generates type 4 guids because people will probably often be replacing that with this. But I suppose they could do their own research if they wanted to know.

jnm2: Apparently since Windows 2000, `CoCreateGuid` has only ever been observed generating type 4 guids. But we shouldn't make it seem like that's documented by Microsoft.

jnm2: @nunit/framework-team Thoughts on this?

CharliePoole: generates a valid type 4 guid? 

Put the bigger story in the documentation if necessary.

rprouse:
We can update the docs early, but we need to specify that it's available in 3.8. On my phone at the moment, so I can't point to another page with the format, but there should be several, just look for new features like parallelism.

_Sent from my Google Pixel using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

MikeFH:
Regarding the documentation, I didn't find the format to specify the version @rprouse mentioned. 
Does this looks OK to you ?

### Guid (System.Guid)

##### `NextGuid()` (available in version 3.8)
Generates a version 4 Guid conforming the [RFC 4122](https://tools.ietf.org/html/rfc4122#section-4.4). Version 4 Guids are made of random data.

jnm2:
@MikeFH That looks perfect for the documentation.

MikeFH:
Thanks !  
The docs are updated : https://github.com/nunit/docs/wiki/Randomizer-Methods.  
There seems to be a error in the travis build but it doesn't look like it is related to the modifications.

PR #2290, opened 2017-06-29
Add support for ValueTuples

### LINKED ISSUE #2254: EqualTo on ValueTuple with Nullable unexpected
I recently observed unexpected test failure when comparing ValueTuples with nullable types.

This works:
`(string name, int val) tuple = ("Hello", 3);`   
`Assert.That(tuple, Is.EqualTo(("Hello", 3)));`

Fails with msg: "Message:   Expected: (Hello, 3) (Int32])  But was:  (Hello, 3) (Int32]])"
`(string name, int? val) tuple = ("Hello", 3)`
`Assert.That(tuple, Is.EqualTo(("Hello", 3)));`

With correct type spec it works again:
`(string name, int? val) tuple = ("Hello", 3);`
`Assert.That(tuple, Is.EqualTo(("Hello", (int?)3)));`

I would expect one of this:

1. check succeeds

1. message shows type difference with Int32? or Nullable<Int32>

mikkelbu:
Support have been added for `ValueTuples`.

As part of this `NUnitEqualityComparer` has been refactored such that the individual comparison functions have been extracted into separate classes and the method `AreEqual` just use these comparison classes.

Fixes #2254


jnm2:
Oh boy, I forgot that this meant we're adding the `ValueTuple` NuGet package... oh boy...
Until we can use the newer package with type forwards to net47 when appropriate, we will run into this if anyone uses NUnit with net47: https://github.com/dotnet/roslyn/issues/20135#issuecomment-310662478

To prevent that from being an issue, we will have to have a separate dependency group in NUnit framework NuGet package for net47 which prevents System.ValueTuple from being a dependency on net47. /cc @ChrisMaddock @rprouse

src/NUnitFramework/framework/Constraints/Comparers/ValueTupleComparer.cs
```diff
@@ -0,0 +1,77 @@
...
+                    return false;
+            }
+
+            return true;
+        }
+
+        private string GetTypeNameWithoutGenerics(string fullTypeName)
+        {
+            int index = fullTypeName.IndexOf('`');
+            return index == -1 ? fullTypeName : fullTypeName.Substring(0, index);
```
jnm2: Can we do `type.Namespace + '.' + type.Name` instead? Or even `type.Namespace == "System" && type.Name == "ValueTuple"`?

jnm2:
Oh wait, false alarm! It's only needed in the test project and we don't build targeting net47 so we're fine.

src/NUnitFramework/framework/Constraints/Comparers/IEquatablesComparer.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Collections.Generic;
+using System.Reflection;
+using NUnit.Compatibility;
+
+namespace NUnit.Framework.Constraints.Comparers
+{
+    /// <summary>
+    /// Comparator for two types related by <see cref="IEquatable{T}"/>.
+    /// </summary>
+    internal class IEquatablesComparer
```
jnm2: Nit: I've gone back and forth on this, but I came to the conclusion that it's standard to remove the `I` from an interface name when it's the first word in a non-interface type name. I believe it's less confusing and just as understandable.

jnm2:
@mikkelbu The `ValueTuple` implementation is top-notch. :+1:

The reorg is a lot to go through, not done that yet.

src/NUnitFramework/framework/Constraints/Comparers/IComparer.cs
```diff
@@ -0,0 +1,45 @@
...
+// ***********************************************************************
+
+using System;
+
+namespace NUnit.Framework.Constraints.Comparers
+{
+    /// <summary>
+    /// Interface for comparing two <see cref="Object"/>s.
+    /// </summary>
+    internal interface IComparer
```
jnm2: I think this approach is very cool, but I'm not a fan of this name. Can we do `IChainOfResponsibilityComparer` since that's exactly what it is, or `IChainComparer` for short?

rprouse: I am okay with the interface name, although `IChainComparer` would be fine too. I wouldn't hold back the merge on this.

src/NUnitFramework/framework/Constraints/Comparers/ArraysComparer.cs
```diff
@@ -0,0 +1,62 @@
+﻿// ***********************************************************************
+// Copyright (c) 2009 Charlie Poole, Rob Prouse
```
rprouse: Don't bother to change, but for future reference, we tend to update the copyright to this year for new files or major changes.

src/NUnitFramework/framework/Constraints/Comparers/IComparer.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// Method for comparing two objects with a tolerance.
+        /// </summary>
+        /// <param name="x">The first object to compare.</param>
+        /// <param name="y">The second object to compare.</param>
+        /// <param name="tolerance">The tolerance to use when comparing the objects.</param>
+        /// <returns>
+        ///     <c>null</c> if the objects cannot be compared using the method.
+        ///     Otherwise the result of the comparison is returned.
+        /// </returns>
+        bool? Equal(object x, object y, ref Tolerance tolerance);
```
rprouse: Interesting using `bool?` here. When I was thinking of doing this, I was going to add a `CanCompare()` method. This saves the extra cast. 👍

src/NUnitFramework/framework/Constraints/Comparers/IComparer.cs
jnm2: It will become a pain to actually use `System.Collections.IComparer` in all the files inside or which import the `NUnit.Framework.Constraints.Comparers` namespace. The refactor is a matter of renaming the file and saying OK when VS2015 or VS2017 asks if you want t orename the code symbol. @mikkelbu what is your opinion?

CharliePoole: I don't like using the name of a System.Collections interface. it's confusing.

src/NUnitFramework/framework/Constraints/Comparers/IEquatablesComparer.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Collections.Generic;
+using System.Reflection;
+using NUnit.Compatibility;
+
+namespace NUnit.Framework.Constraints.Comparers
+{
+    /// <summary>
+    /// Comparator for two types related by <see cref="IEquatable{T}"/>.
+    /// </summary>
+    internal class IEquatablesComparer : IComparer
```
CharliePoole: I'd prefer EquatablesComparer, since it's not an interface.

mikkelbu:
Thanks for all the comments. I'll be busy most of the weekend, but I'll take a look at them Sunday or beginning of next week.

rprouse:
@mikkelbu contacting you through this PR since I don't have your email.

You have been very active in both the framework and console/engine repositories lately and we really appreciate your help, so I have added you to the two teams @nunit/framework-team and @nunit/engine-team. This is to recognize your help and make it easier to work, but is not an obligation on your part.

You can now push branches to both repositories which means that you don't have to work in a fork. Master is locked, so no worries about messing up. As a member, your PRs will only require one other team member to review going forward and you are welcome to review and merge other PRs. Our policy is for two team members to review non-team PRs and one team member to review other team members PRs. Of course, use your discretion 😉 

If you have any questions now or at any time, please email me at rob@prouse.org.

Welcome 😄

src/NUnitFramework/framework/Constraints/Comparers/ValueTupleComparer.cs
mikkelbu: We will still have "problems" with '`'. 
```
            var myTuple = new ValueTuple<int>(123);
            var tupleType = myTuple.GetType();
            Console.WriteLine(tupleType.Name);
```
Prints `ValueTuple\`1`. So I've kept the code as it is. But input is most welcome :)

jnm2: Oh, I see. I doubt there can be an improvement over what you've done, or [Jon Skeet would likely have posted it](https://stackoverflow.com/a/6386234). 😆

mikkelbu:
I've pushed 3 commits:
* 079399e. Added formatting for ValueTuples (I forgot to push this commit when making the PR)
* d81bfe5. Made the requested renamings: `IComparer` => `IChainComparer` and `IEquatablesComparer` => `EquatablesComparer`
* f3c2bdd. Due to @jnm2 request for tests with 8 (or more) items I discovered that the initial code did not handle this. So I've extended the code to also handle this situation (both in the equality and in formatting)

mikkelbu:
@CharliePoole Do you also approve the renamings?

jnm2:
I'd really like to know what's up with Travis. The console output makes no sense to me.

CharliePoole:
Me either. I restarted.

jnm2:
@mikkelbu Your contribution is appreciated!

jnm2:
I upgraded a project at work to use 3.9 (because 3.7.1's Has.Property was pulling an entirely wrong value and I wondered if my property shadowing changes fixed it). The code has a custom constraint which uses `new NUnitEqualityComparer()`.
As of this PR (released in 3.8), that constructor became internal, so now the code doesn't compile. I wonder what the right fix would be in this custom constraint implementation?

jnm2:
Context: this is `HasEqualPropertyValues(expectedObj, "Property1", "Property2", ...)` constraint, so using `NUnitEqualityComparer` seems like a legit thing to do. But maybe I'm missing the correct entry point.

jnm2:
Oh, ignore me. Finally spotted `NUnitEqualityComparer.Default`. 😆

PR #2294, opened 2017-07-01
Add Using(StringComparison) method

MhdTlb:
Add Using(StringComparison) method that will flag the constraint to use the supplied StringComparison enum.


jnm2:
Great, I'll take a look next! The CI failure is because your code will not build under netstandard16. Before you create each commit, it's recommended to run `build -t test` which builds and tests against all targets. That needs to be fixed.

jnm2:
My bad, netstandard13.

jnm2:
@MhdTlb This is my bad, I mistakenly said the `Using` method belongs in `StringConstraint` but it actually belongs in `SubstringConstraint`. The reason we don't want it in the `StringConstraint` class is that it makes no sense on regex or path constraints which derive from `StringConstraint`.

I'm sorry!

The good news is it's a lot less work because you only have to add behavior to `SubstringConstraint.Matches`, not *every* subclass of `StringConstraint`.

src/NUnitFramework/framework/Constraints/StringConstraint.cs
```diff
@@ -98,12 +98,12 @@ public override ConstraintResult ApplyTo<TActual>(TActual actual)
 
             return new ConstraintResult(this, actual, Matches(actualAsString));
         }
-
+
```
jnm2: It would be cool if this file didn't have to show up as changed in this PR. By the way, feel free to squash and rebase however suits you.

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
```diff
@@ -30,6 +32,11 @@ namespace NUnit.Framework.Constraints
     public class SubstringConstraint : StringConstraint
     {
         /// <summary>
+        /// Indicates whether tests should be case-insensitive
```
jnm2: I think this message is just a little inaccurate because there's more going on than just cas-sensitivity, but you don't have to have XML docs on private members anyway.

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
```diff
@@ -45,10 +52,18 @@ public SubstringConstraint(string expected) : base(expected)
         /// <returns>True for success, false for failure</returns>
         protected override bool Matches(string actual)
         {
-            if (this.caseInsensitive)
-                return actual != null && actual.ToLower().IndexOf(expected.ToLower()) >= 0;
-            else
-                return actual != null && actual.IndexOf(expected) >= 0;
+            return actual != null && actual.IndexOf(expected, comparisonType) >= 0;
```
CharliePoole: User may have specified CaseInsensitive as a modifier. This should chagne the comparisonType.

CharliePoole: We also have to deal with both Using and CaseInsensitive being specified - probably as an error.

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
```diff
@@ -45,10 +52,18 @@ public SubstringConstraint(string expected) : base(expected)
...
+        }
+
+        /// <summary>
+        /// Flag the constraint to use the supplied StringComparison type.
+        /// </summary>
+        /// <param name="comparisonType ">The StringComparison type to be used.</param>
+        /// <returns>Self.</returns>
+        public SubstringConstraint Using(StringComparison comparisonType)
```
CharliePoole: This is a different use of Using from any others we have, but I think it should be clear to users.

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
jnm2: To keep the same behavior as `IgnoreCase` and `Using` on `EqualityComparer`, we'll need the `comparisonType` to be nullable in case no one calls `Using`. Then in here, if `comparisonType` is not null, it overrides everything else and returns exactly what you have here. If it is null, we need to run the original `caseInsensitive` code. (/cc @CharliePoole)

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
```diff
@@ -45,10 +52,18 @@ public SubstringConstraint(string expected) : base(expected)
...
-                return actual != null && actual.ToLower().IndexOf(expected.ToLower()) >= 0;
-            else
-                return actual != null && actual.IndexOf(expected) >= 0;
+            return actual != null && actual.IndexOf(expected, comparisonType) >= 0;
+        }
+
+        /// <summary>
+        /// Flag the constraint to use the supplied StringComparison type.
+        /// </summary>
+        /// <param name="comparisonType ">The StringComparison type to be used.</param>
```
jnm2: Nit: space in `"comparisontype "`.

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
```diff
@@ -45,10 +52,18 @@ public SubstringConstraint(string expected) : base(expected)
...
         {
-            if (this.caseInsensitive)
-                return actual != null && actual.ToLower().IndexOf(expected.ToLower()) >= 0;
-            else
-                return actual != null && actual.IndexOf(expected) >= 0;
+            return actual != null && actual.IndexOf(expected, comparisonType) >= 0;
+        }
+
+        /// <summary>
+        /// Flag the constraint to use the supplied StringComparison type.
```
jnm2: Since I've verified that this is the behavior that matches `EqualConstraint`, can the summary be `Overrides the <see cref="IgnoreCase"/> setting and instead uses the specified comparison.`?

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
jnm2: Plus, I recommend not even including the `<param>` tag because there isn't any information to convey that isn't already in the type and name and summary.

jnm2: While I'm talking about this, feel free to remove the `<returns>` tag if you want. I recommend that because it almost certainly won't be seen by library consumers.

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
CharliePoole: I think we should get rid of the original code with ToLower however, and just pick a could case-insensitive comparisonType. IOW IgnoreCase in this class would just be an alias for a particular Using.

jnm2: Sorry, race condition in our reviews. At least we came to the same conclusion. =)

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
jnm2: @CharliePoole Yes. The BCL has a precedent of taking `StringComparison` being equivalent to taking an `IEqualityComparer<string>`.

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
jnm2: So yes, in `EqualityComparer`, `Using` just silently overrides `IgnoreCase`.

jnm2: `.ToLower()` is specifically the equivalent of `StringComparison.CurrentCultureIgnoreCase`. So the logic, without breaking changes, is:

`comparisonType ?? (caseInsensitive ? StringComparison.CurrentCultureIgnoreCase : StringComparison.Ordinal)`

jnm2:
You don't have to do it this way but I wanted to let you know this option is available:

```c#
[TestCase(" ss ", "ß", StringComparison.CurrentCulture, true)]
[TestCase(" SS ", "ß", StringComparison.CurrentCulture, false)]
[TestCase(" ss ", "s", StringComparison.Ordinal, true)]
[TestCase(" SS ", "s", StringComparison.Ordinal, false)]
[TestCase(" ss ", "ß", StringComparison.CurrentCultureIgnoreCase, true)]
[TestCase(" SS ", "ß", StringComparison.CurrentCultureIgnoreCase, false)]
[TestCase(" ss ", "s", StringComparison.OrdinalIgnoreCase, true)]
[TestCase(" SS ", "s", StringComparison.OrdinalIgnoreCase, false)]
public void Using(string actual, string expected, StringComparison comparison, bool succeeds)
{
    var constraint = (Constraint)Contains.Substring(expected).Using(comparison);
    if (succeeds) constraint = new NotConstraint(constraint);
    Assert.That(succeeds);
}
```

jnm2:
We should also have `[TestCase(" ss ", "ß", StringComparison.Ordinal, false)]`.

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
MhdTlb: Sure. But how to resolve `XML comment has cref attribute 'IgnoreCase' that could not be resolved`	error?

jnm2: That's weird, I thought it would find `StringConstraint.IgnoreCase` because you're inside the `SubstringConstraint` class. If not, try `cref="StringConstraint.IgnoreCase"`?

jnm2: Wow, I learned something tonight! You do need the base class name.

MhdTlb: That worked. Thanks

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
```diff
@@ -56,13 +53,12 @@ protected override bool Matches(string actual)
...
+        /// Overrides the <see cref="StringConstraint.IgnoreCase"/> setting and instead uses the specified comparison.
         /// </summary>
-        /// <param name="comparisonType ">The StringComparison type to be used.</param>
-        /// <returns>Self.</returns>
-        public SubstringConstraint Using(StringComparison comparisonType)
+        public SubstringConstraint Using(StringComparison? comparisonType)
         {
-            this.comparisonType = comparisonType;
+            this.comparisonType = comparisonType ?? (caseInsensitive ? 
+                StringComparison.CurrentCultureIgnoreCase : StringComparison.Ordinal);
```
CharliePoole: The two options seem contradictory. Shouldn't it be either CurrentCulture[IgnoreCase] or Ordinal[IgnoreCase]?

jnm2: @CharliePoole This is necessary for back-compat. If we want to change this, we need a design discussion.

Currently, `Assert.That("ss", Contains.Substring("ß"))` fails and `Assert.That("ss", Contains.Substring("ß").IgnoreCase)` succeeds. If we made both `CurrentCulture*`, they would both succeed. If we made both `Ordinal*`, the would both fail.

jnm2: Wait, I'm wrong. They both succeed. One second.

jnm2: I did not know this till just now. `string.IndexOf` operates in `CurrentCulture` unless otherwise specified. I thought it was ordinal like `string.Equals` is. So @CharliePoole I was wrong. We should in fact have: `caseInsensitive ? StringComparison.CurrentCultureIgnoreCase : StringComparison.CurrentCulture`.

/cc @MhdTlb

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
```diff
@@ -45,10 +49,17 @@ public SubstringConstraint(string expected) : base(expected)
...
+            return actual != null && actual.IndexOf(expected, comparisonType) >= 0;
+        }
+
+        /// <summary>
+        /// Overrides the <see cref="StringConstraint.IgnoreCase"/> setting and instead uses the specified comparison.
+        /// </summary>
+        public SubstringConstraint Using(StringComparison? comparisonType)
+        {
+            this.comparisonType = comparisonType ?? (caseInsensitive ? 
+                StringComparison.CurrentCultureIgnoreCase : StringComparison.Ordinal);
```
jnm2: @MhdTlb I apologize! In the comment I cc'd you on, I realized that we need `StringComparison.CurrentCultureIgnoreCase : StringComparison.CurrentCulture` to match the current behavior because in the old code, the `IndexOf` was actually doing a `CurrentCulture` comparison.

src/NUnitFramework/tests/Constraints/SubstringConstraintTests.cs
```diff
@@ -67,6 +67,88 @@ public void SetUp()
...
+
+        [TestCase(" ss ", "�", true, StringComparison.CurrentCulture, true)]
+        [TestCase(" SS ", "�", true, StringComparison.CurrentCulture, false)]
+        [TestCase(" ss ", "s", true, StringComparison.CurrentCulture, true)]
+        [TestCase(" SS ", "s", true, StringComparison.CurrentCulture, false)]
+        [TestCase(" ss ", "�", true, StringComparison.CurrentCultureIgnoreCase, true)]
+        [TestCase(" SS ", "�", true, StringComparison.CurrentCultureIgnoreCase, true)]
+        [TestCase(" ss ", "s", true, StringComparison.CurrentCultureIgnoreCase, true)]
+        [TestCase(" SS ", "s", true, StringComparison.CurrentCultureIgnoreCase, true)]
+#if !NETSTANDARD1_3
```
jnm2: Looks like we need `!(NETSTANDARD!_3 || NETSTANDARD1_6)`. (You can test this on your own too by running `build -t test` at the command prompt from your local repo root.)

src/NUnitFramework/tests/Constraints/SubstringConstraintTests.cs
```diff
@@ -67,6 +67,88 @@ public void SetUp()
...
+        [TestCase(" ss ", "�", false, StringComparison.Ordinal, false)]
+        [TestCase(" SS ", "�", false, StringComparison.Ordinal, false)]
+        [TestCase(" ss ", "s", false, StringComparison.Ordinal, true)]
+        [TestCase(" SS ", "s", false, StringComparison.Ordinal, false)]
+        [TestCase(" ss ", "�", false, StringComparison.OrdinalIgnoreCase, false)]
+        [TestCase(" SS ", "�", false, StringComparison.OrdinalIgnoreCase, false)]
+        [TestCase(" ss ", "s", false, StringComparison.OrdinalIgnoreCase, true)]
+        [TestCase(" SS ", "s", false, StringComparison.OrdinalIgnoreCase, true)]
+
+        [TestCase(" ss ", "�", false, null, false)]
```
jnm2: This will have to change too because of the thing I was wrong about.

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
MhdTlb: Just a suggestion: Why don't we let `indexOf` decides which `StringComparison` to use. I mean we keep `SubstringConstraint.comparisonType` nullable also and change `Matches` method instead to be
 `protected override bool Matches(string actual)
  {
        if (actual == null)
              return false;

        if (comparisonType == null)
             return actual.IndexOf(expected) >= 0;

       return actual.IndexOf(expected, comparisonType) >= 0;
  }`

jnm2: @MhdTlb Because if `comparisonType == null` we still need to make a decision based on whether or not they called `.IgnoreCase`. If you try the above code, tests will fail demonstrating this.

jnm2:
@MhdTlb I see test failures: https://ci.appveyor.com/project/CharliePoole/nunit/build/3.8.0-ci-04151-pr-2294#L812 Those test failures will happen when you run tests in your IDE or using `build.cmd -t test`.

MhdTlb:
@jnm2  I'm trying to debug test failures by running `build.cmd -t test` command, I got the same results in the command prompt but they are not in the `TestResult.xml` file. Am I missing some set-ups?
Also, all failed tests are related to `SubstringConstraint` old methods not `Using` method that I added.
I can't see how this could happen?!

jnm2:
I don't know about TestResults.xml. (@nunit/framework-team help?) So far I have only looked at the console output when running build.ps1.

As to how old tests are failing, let me pull the code and get a better look at it.

jnm2:
The reason for the 3 test failures inside the original `SubstringConstraintTestsIgnoringCase` class is that `Contains.Substring(str).IgnoreCase` no longer ignores case. It no longer ignores case because `SubstringConstraint.Matches` does not check the `caseInsensitive` field. `Matches` still needs to act as it did before when no one calls the `Using` method. That's why `Matches`, not `Using`, is supposed to contain this code:

```c#
comparisonType ?? (caseInsensitive ? 
    StringComparison.CurrentCultureIgnoreCase : StringComparison.CurrentCulture)
```

jnm2:
The way you have it now, the code `caseInsensitive ? StringComparison.CurrentCultureIgnoreCase : StringComparison.CurrentCulture` only runs if someone calls the `Using` method. But we want it to run every time `Matches` runs, even if `Using` was never called. Like in `Contains.Substring(str).IgnoreCase`.

The `Using` method should take a `StringComparison`  parameter rather than a `StringComparison?` parameter. The `comparisonType` class field is what should be typed `StringComparison?`: null means no one has ever called `Using`, and a value means someone called `Using`.

Then `Matches` runs code similar to this:
```c#
if (actual == null) return false;

var actualComparison = comparisonType ?? (caseInsensitive ? 
    StringComparison.CurrentCultureIgnoreCase : StringComparison.CurrentCulture);

return actual.IndexOf(expected, actualComparison) >= 0;
```

Does that all make sense? Sorry I didn't see it sooner in the GitHub review. I'll be more careful next time.

jnm2:
Failing tests means there is a puzzle to be solved, and tracking down that root cause can be fun (depending on the bug). Hopefully I didn't take that away from you too much. 😈

jnm2:
You'll still have to figure out how to write the `UsingTests.Using` method now that `SubstringConstraint.Using` only takes a `StringComparison` parameter.

As for the three remaining errors in the `UsingTests` methods inherited from `StringConstraintTests`-  `StringConstraintTests` expects you to be following the SetUp/SuccessData/FailureData pattern like `SubstringConstraintTests`, so unless you want to do that for your tests (which you don't, there's no reason to), you shouldn't inherit from `StringConstraintTests`.

Now you could stick your single `Using` test method inside the existing `SubstringConstraintTests` class and get rid of the `UsingTests` class. No harm in that. I sort of prefer that because `UsingTests` is not specific enough of a name. `Using` appears on other constraints that are not `SubstringConstraint`.

Happy to help if you get stuck. =)

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
```diff
@@ -43,22 +44,34 @@ public SubstringConstraint(string expected) : base(expected)
         }
 
         /// <summary>
+        /// Modify the constraint to ignore case in matching.
+        /// </summary>
+        public override StringConstraint IgnoreCase
+        {
+            get { caseInsensitive = true; comparisonType = null; return this; }
```
jnm2: @CharliePoole Is this what we want? As we discussed earlier and I double checked via testing, currently `Is.EqualTo(x).Using(StringComparer.Ordinal).IgnoreCase` does *not* reset the `Using` setting.. all `IgnoreCase` calls are ignored if any `Using` is present no matter the order.

I'd be more in favor of throwing an exception, if anything.

MhdTlb: Any decision about this?

jnm2: @nunit/framework-team Chime in please? Should we stay with what EqualConstraint does, or allow this behavior change and probably also change EqualConstraint?

CharliePoole:
@jnm2 I think we've all already chimed in. My view, again, is that using both `Using` and `IgnoreCase` is an error. We don't report it as an error in EqualConstraint, but rule in favor of `Using`. That follows the rule "If you tell me to do contradictory things, I can do whatever I like" but I think an explicit error would be better.

I think we should treat `IgnoreCase` as equivalent to a particular `Using` and give an error if more than one is entered.

If, in the future, we were to add the same `Using` overload to `EqualConstraint`, I think we should do it the same way. That might also lead us to reconsider how we handle the already existing `Using` overloads, but that's another issue.

jnm2:
Okay, so we need to get rid of the `caseInsensitive` field and make the IgnoreCase property delegate to `Using(StringComparison.CurrentCultureIgnoreCase)`. The IgnoreCase XML summary should say that it is shorthand for `Using(StringComparison.CurrentCultureIgnoreCase)`.

`comparisonType` should stay nullable for error checking. If it's null when `Using` is called, `Using` just sets it. If it's not null, `Using` does nothing if the new value matches. If the new value does not match, `Using` throws `InvalidOperationException`.

The XML documentation summary for both `Using` and `IgnoreCase` should warn about the InvalidOperationException if you call with two differing values.

We really appreciate your patience with this one, MhdTlb!

src/NUnitFramework/tests/Constraints/SubstringConstraintTests.cs
```diff
@@ -38,117 +38,98 @@ public void SetUp()
...
+            Assert.DoesNotThrow(() =>
+            {
+                var newConstriant = subStringConstriant.Using(StringComparison.CurrentCultureIgnoreCase).IgnoreCase;
+            });
+
+            var stringConstriant = theConstraint as StringConstraint;
+            Assert.DoesNotThrow(() =>
+            {
+                var newConstriant = stringConstriant.IgnoreCase as SubstringConstraint;
+                newConstriant = newConstriant.Using(StringComparison.CurrentCultureIgnoreCase);
```
jnm2: Nit: spelling error, should be "constraint" (19 times)

jnm2: Otherwise, nice tests!

MhdTlb: My bad. Will fix it

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
```diff
@@ -39,16 +44,38 @@ public SubstringConstraint(string expected) : base(expected)
         }
 
         /// <summary>
+        /// Modify the constraint to ignore case in matching.
+        /// This will call Using(StringComparison.CurrentCultureIgnoreCase) 
+        /// </summary>
+        public override StringConstraint IgnoreCase
+        {
+            get { caseInsensitive = true; Using(StringComparison.CurrentCultureIgnoreCase); return this; }
```
jnm2: Since this class doesn't "own" the `caseInsensitive` field, perhaps we should be calling `var _ = base.IgnoreCase` instead of coupling to `caseInsensitive`. Probably not important either way.

MhdTlb: I agree with you. It would be better to invoke `base.ignoreCase`

MhdTlb:
@jnm2  Is there a particular reason for `The Travis CI build failed`? 
Also, I tried to run `build -t test` command before pushing and I did not get the same errors as `The Travis CI build`

jnm2:
@MhdTlb Looks like a file encoding issue perhaps?

> 
> 1) Failed : NUnit.Framework.Constraints.SubstringConstraintTests.SpecifyComparisonType(" SS ","�",CurrentCulture,False)
> 
>   Expected: not String containing "�"
> 
>   But was:  " SS "
> 
>   at NUnit.Framework.Constraints.SubstringConstraintTests.SpecifyComparisonType (System.String actual, System.String expected, System.StringComparison comparison, System.Boolean succeeds) [0x0001b] in <e9d9c89fbd59443fa5ef5f87d3361ec2>:0 
> 
> 2) Failed : NUnit.Framework.Constraints.SubstringConstraintTests.SpecifyComparisonType(" SS ","�",InvariantCulture,False)
> 
>   Expected: not String containing "�"
> 
>   But was:  " SS "
> 
>   at NUnit.Framework.Constraints.SubstringConstraintTests.SpecifyComparisonType (System.String actual, System.String expected, System.StringComparison comparison, System.Boolean succeeds) [0x0001b] in <e9d9c89fbd59443fa5ef5f87d3361ec2>:0 

Can you make sure you do Save with Encoding and choose UTF8 without byte order mark?

jnm2:
I just pulled your source and verified that the encoding was indeed ANSI rather than UTF8. Without exception, any text file in the repo should be UTF8.

We should have a .editorconfig to guide people's editors onto the right track so that you don't have to think about things like encodings.

MhdTlb:
Any update regarding last push?

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
```diff
@@ -39,16 +44,38 @@ public SubstringConstraint(string expected) : base(expected)
...
-                return actual != null && actual.IndexOf(expected) >= 0;
+            if (actual == null) return false;
+
+            var actualComparison = comparisonType ?? StringComparison.CurrentCulture;
+            return actual.IndexOf(expected, actualComparison) >= 0;
+        }
+
+        /// <summary>
+        /// Modify the constraint to the specified comparison.
+        /// </summary>
```
jnm2: My comment must have gotten lost on an earlier commit, but any time we purposely throw an exception throw an exception we need to put a message in the summary saying what the exception is an under what circumstances it will happen. Here's an example: https://github.com/nunit/docs/wiki/Best-practices-for-XML-documentation#exception

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
```diff
@@ -39,16 +44,38 @@ public SubstringConstraint(string expected) : base(expected)
         }
 
         /// <summary>
+        /// Modify the constraint to ignore case in matching.
+        /// This will call Using(StringComparison.CurrentCultureIgnoreCase) 
+        /// </summary>
```
jnm2: I think that this should have the same `InvalidOperationException` message as `Using` just so people don't have to guess how the API works.

jnm2:
@MhdTlb Since we went back and forth a little, how do you feel about you or me squashing some of these commits and rebasing to get rid of the merge commits?

The way I like to keep up to date with the tip of master, rather than merging, is to rebase using `git pull -r upstream master`. It makes it look like all the coding was done at an atomic point in the history. I think it makes it easier to follow for people reviewing the code. Especially in cases where there are no merge conflicts, I don't see a reason to have merge commits within a feature branch.

(When there are merge conflicts, I still prefer to rebase because it takes me commit by commit and lets me decide how to react to the underlying changes while still keeping the original intention of my commit message. But if you have way too many conflicts and way too much code, then merging makes sense to be because you deal with it all at once. The all-at-once conflict resolution is still sometimes harder to understand later. This is all personal preference.)

What do you think? Maybe I can help open up an easier workflow, but I'm not trying to change your mind.
If you're interested, I can tell you the git commands I'd use to squash this PR into fewer commits.

MhdTlb:
I totally agree. It's a good idea. What commands should I use?

jnm2:
I was going to do it via interactive rebase, but that would throw out all your merge commits and you'd have to redo all the conflict resolution.

This is what `git log --pretty=format:'%h %ad | %s%d [%an]' --graph --date=short -n19` shows (I have it aliased so I can type `git hist -n19`):
```
* 24066808 2017-07-13 | Make SubstringConstraint.IgnoreCase invoke base IgnoreCase property (HEAD -> patch-1, origin/patch-1) [MhdTlb]
* bc3c946f 2017-07-13 | Switch to encoding UTF-8 without BOM [MhdTlb]
* 6a52ddda 2017-07-13 | Fixes spelling error [MhdTlb]
* 8ceecefe 2017-07-13 | Throws exception when Using and IgnoreCase are used in contradictory way [MhdTlb]
* eb1ac321 2017-07-09 | Use Nullable comparison type [atalbi]
* c24c6a45 2017-07-03 | Use CurrentCulture[Ignore] as String comparisonType [atalbi]
* 3e3c71fb 2017-07-02 | Use Nullable StringComparison, Add tests for all StringComparison enums [atalbi]
*   83b70fd3 2017-07-01 | merge [atalbi]
|\
| * a2d79288 2017-07-01 | Remove Using(StringComparison) method [atalbi]
| *   7109078a 2017-07-01 | Merge branch 'issue-2195' into patch-1 [atalbi]
| |\
| | * 218a7d0a 2017-07-01 | merge (origin/issue-2195) [atalbi]
| | *   2fcf226f 2017-07-01 | Merge branch 'issue-2195' of https://github.com/MhdTlb/nunit into issue-2195 [atalbi]-
| | |\
| | | * 3bec1890 2017-07-01 | Add Using(StringComparison) method [MhdTlb]
| | * | ef5ac78b 2017-07-01 | Add Using(StringComparison) method [atalbi]
| | |/
| * | 64b37df2 2017-07-01 | Add Using(StringComparison) method [MhdTlb]
| |/
* | 788cc88f 2017-07-01 | Add Using method and tests [atalbi]
|/
*   8f49cbc9 2017-06-30 | Merge pull request #2293 from nunit/ordered_by (origin/master, origin/issue-2294, origin/issue-2194, origin/HEAD, master) [Rob Prouse]
```

Let's definitely squash `*   83b70fd3 2017-07-01 | merge [atalbi]` and older so we don't have merge conflicts when rebasing. Based on the back and forth in this PR, I'm thinking I would keep the top 4 commits and squash everything also into a single commit. So we're squashing `* eb1ac321 2017-07-09 | Use Nullable comparison type [atalbi]` and older.



First, let's make sure we have the latest master.
```
git fetch --all
```

Now create a temporary branch for scratch work.
We want the parent commit of the squash commit to be the latest upstream/master commit so that we're already up-to-date with upstream/master when we're done.
```
git checkout upstream/master -b patch-1_partial_squash
```

Squash all the history up to and including `* eb1ac321 2017-07-09 | Use Nullable comparison type [atalbi]`:
```
git merge eb1ac321 --squash
```

Create a commit from the squashed changes which are now staged. Because we aren't using `-C`, this acts like a brand new commit and resets the metadata like author and timestamp which is a nice side effect because it fixes the problem that you had a bunch of commits with the username `atalbi` which GitHub does not attribute to you. (You might want to fix your .gitconfig on that machine.)
```
git commit -m "Add Using(StringComparison) method and tests
```

Now we want to move the last four commits onto the squashed commit.
```
git checkout patch-1
```

Move the four commits which are currently sitting on top of eb1ac321 to instead sit on top of patch-1_partial_squash:
```
git rebase eb1ac321 --onto patch-1_partial_squash
```

Temporary branch no longer needed:
```
git branch -D patch-1_partial_squash
```

Now let's eyeball this and make sure that we didn't mess up, so that when we push, we don't overwrite origin/patch-1 with garbage:


```
git log --pretty=format:'%h %ad | %s%d [%an]' --graph --date=short -n6
```

```
* 629eefdd 2017-07-13 | Make SubstringConstraint.IgnoreCase invoke base IgnoreCase property (HEAD -> patch-1) [MhdTlb]
* 24ef3c75 2017-07-13 | Switch to encoding UTF-8 without BOM [MhdTlb]
* 0b72525f 2017-07-13 | Fixes spelling error [MhdTlb]
* 7502f75c 2017-07-13 | Throws exception when Using and IgnoreCase are used in contradictory way [MhdTlb]
* 8abd9858 2017-07-14 | Add Using(StringComparison) method and tests [jnm2]
*   1a3c728a 2017-07-12 | Merge pull request #2310 from StanEgo/fixes/issues-1851 (upstream/master) [Rob Prouse]
```

And we're done! We can update your PR branch. Force is necessary because we rewrote history, which is normal for PR branches.
```
git push -f
```


---



There were a couple times I messed up and had to delete my clone of patch-1 and pull it fresh from your fork:
```
git checkout origin/patch-1
git branch -f patch-1
git checkout patch-1
```

jnm2:
Okay, I hadn't seen your last two commits yet. That last merge brought in a crazy amount of commits to the PR. Let me revise my comment.

jnm2:
We definitely need to undo that merge. We want to keep the previous commit, `* 24066808 2017-07-13 | Make SubstringConstraint.IgnoreCase invoke base IgnoreCase property [MhdTlb]`:

```
git checkout 24066808 
git branch -f patch-1
git checkout patch-1
```

Then follow https://github.com/nunit/nunit/pull/2294#issuecomment-315507376 and ignore where it says "top four commits" because it will be five. The instructions won't change.

jnm2:
My rule of thumb is: never do `git pull`. If what I'm trying to do is update my clone of the `upstream/` or `origin/` branches, I use `git fetch` . If what I'm trying to do is bring my feature branch up to date with master, I use `git pull -r` so git does fetch-rebase instead of fetch-merge.

jnm2:
`git pull` is also a terrible idea for synchronizing work in the same branch between computers, because you don't want merge commit s from your branch to another version fo the same branch. `git pull -r` is much better, and best of all is doing `git push` every time you stop work at a computer and `git pull -r` every time you start work at a computer so that you only ever do fast-forwards and never have to merge or rebase commits.

jnm2:
@MhdTlb so if all that looks like too much work, it's okay to leave it as is and one of us will rebase and squash before merging to master. That's totally fine.

jnm2:
Here's what it looks like after I performed the steps above: https://github.com/nunit/nunit/compare/using_stringcomparison

MhdTlb:
@jnm2  Thanks for your patience. I really appreciate your help. This is my first PR in Github :)

MhdTlb:
@jnm2  What would be the next step? Should I close the PR?

jnm2:
The next step is for a second @nunit/framework-team to review and merge. (psst)

Issue #2288, opened 2017-06-29
Killing thread cancels test run


CharliePoole:
I've only seen this once, but it's pretty scary. In the following CI build, an entire test run is cancelled while we are testing ThreadUtility.Kill. I've never see this before.

https://ci.appveyor.com/project/CharliePoole/nunit/build/3.8.0-ci-04129-pr-2287

rprouse:
Interesting, we had a timeout in those same tests and also in .NET 3.5 only in #2283 a few days ago.

jnm2:
This is a test I wrote, so I'll take this one.

jnm2:
I'll work on it tonight EST.

jnm2:
Well, I have extra work so I'll have to put this off a day.

jnm2:
Yikes. Worked 14 hours, slept 4.5, worked 11.5. Brain is cooked for today. But work deadline is over so I'll try to do this tomorrow at some point.

jnm2:
The worst thing that's happening is a call to thread.Abort() on a thread I created, not the test thread, so I'm not sure how that got propagated back up to NUnit.

jnm2:
I'm running stress tests via `build -t test35` and not getting a repro on my computer.

CharliePoole:
It shouldn't... for exactly this reason. There is code that NUnit uses to cancel an Abort unless a particular flag is set in the TestExecutionContext that tells us "Yes, we really want to abort the entire run."

jnm2:
Do we have a second example of this failure on AppVeyor? I haven't found one so far.

jnm2:
AppVeyor needs a log search option.

CharliePoole:
I saw it once before. Same stack trace as far as I remember. I wrote the bug on the second appearance of the problem.

jnm2:
I can't think of any lines of attack. Can't think of any ways that a ThreadAbortException could be raised on a test thread via this test.

CharliePoole:
I took a look at it. You have a TimeoutAttribute set, which will kill the test thread if it takes more than 1000ms. Maybe that interacts with the kill in your test in some unforseen way. You could try setting that timeout very short to see what happens.

jnm2:
Now I feel silly. :D

jnm2:
The problem with timeouts is that a busy server might not have the resources to complete the test in the desired time. We could try bumping this up to ten seconds. It's not expected to ever run that long; it's just a way to keep the test from running forever worst-case.

[...]

Issue #2134, opened 2017-04-25
TimeoutAttribute and TimeoutCommand


oznetmaster:
When a timeout elapses, the TimeoutCommand correctly aborts the thread and records the result of "Test exceeded Timeout value of {0}ms".

However, when the SimpleWorkItem subsequently catches the ThreadAbortException, it then appears to overwrite that result with a "Test Cancelled by User" result.

This is on a platform which does not support Thread.ResetAbort ().  However, from the code:

```
		protected override void PerformWork ()
			{
			try
				{
				Result = MakeTestCommand ().Execute (Context);
				}
			catch (Exception ex)
				{
				// Currently, if there are no command wrappers, test 
				// actions, setup or teardown, we have to catch any
				// exception from the test here. In addition, since
				// users may create their own command wrappers, etc.
				// we have to protect against unhandled exceptions.

#if !PORTABLE && !NETSTANDARD1_6 && !SSHARP
				if (ex is ThreadAbortException)
					Thread.ResetAbort ();
#endif

				Context.CurrentResult.RecordException (ex);
				}
			finally
				{
				WorkItemComplete ();
				}
			}
```

that seemingly does not make any difference, since the ThreadAbortException is always recorded, overwriting the prior result.

It may be that the platform makes a difference in the order in which things happen.  If the Thread.Abort in the TimeoutCommand were to cause the exception before the TimeoutCommand set the new result, then it would overwrite the "Test Cancelled by User" result.  However, that cannot be assumed, which is what seems to be the case here.

oznetmaster:
To clarify further, I am seeing the TimeoutCommand.AfterTest occur before the ThreadAbortException occurs.

oznetmaster:
This comment in the code:
					// No join here, since the thread doesn't really terminate
after the Thread.Abort puzzles me, since it seems that, in fact, there should be a Join there, at least on platforms which do not support Thread.ResetAbort ()?

oznetmaster:
I added the Join, and it returned seemingly immediately after the abort.

However, AfterTest is still being called before the ThreadAbortException is thrown.

oznetmaster:
I finally realized that the Before/After are running on a different thread then the one which was aborted.

The current order makes perfect sense.  Therefore, it is unclear to me how this could ever work as expected :(

I must be missing something.

oznetmaster:
[...]

PR #2295, opened 2017-07-01
Stop using [Timeout] so that there should be no thread abort on the t…

jnm2:
Possibly fixes https://github.com/nunit/nunit/issues/2288, at least closes it for now.

This attempts to simplify the situation so that a thread abort exception will never happen on the test execution thread.

I would have originally picked `Assert.That(thread.Join(1000), ...` over `[Timeout(1000)]` if it had crossed my mind when I first wrote it because it's less wasteful.

### DIFF
--- a/src/NUnitFramework/tests/Internal/ThreadUtilityTests.cs
+++ b/src/NUnitFramework/tests/Internal/ThreadUtilityTests.cs
@@ -9,7 +9,6 @@ namespace NUnit.Framework.Internal
     public class ThreadUtilityTests
     {
         [Platform("Win")]
-        [Timeout(1000)]
         [TestCase(false, TestName = "Abort")]
         [TestCase(true, TestName = "Kill")]
         public void AbortOrKillThreadWithMessagePump(bool kill)
@@ -34,9 +33,8 @@ namespace NUnit.Framework.Internal
                 else
                     ThreadUtility.Abort(thread, nativeId);
 
-                thread.Join();
+                Assert.That(thread.Join(1000), "Native message pump was not able to be interrupted to enable a managed thread abort.");
             }
-
         }
 
         [DllImport("user32.dll")]


ChrisMaddock:
@jnm2 - have you noticed these tests failed on a CI run? 🙂 

https://ci.appveyor.com/project/CharliePoole/nunit/build/3.8.0-ci-04140-threading-t#L1213

jnm2:
@ChrisMaddock Did not see them, but it looks like a fluke. The first test is not related an also had a timeout issue. It means the AppVeyor server was busy and that we should be increasing our timeout lengths. If we see this more than once we can go to two seconds. The amount is not the important thing, it's just a convenience.

> 1) Failed : NUnit.Framework.Internal.Execution.EventQueueTests+DequeueBlocking_StopTest.DequeueBlocking_Stop
> Test exceeded Timeout value of 1000ms
> 2) Failed : NUnit.Framework.Internal.ThreadUtilityTests.Abort
>   Native message pump was not able to be interrupted to enable a managed thread abort.
>   Expected: True
>   But was:  False
>    at NUnit.Framework.Internal.ThreadUtilityTests.AbortOrKillThreadWithMessagePump(Boolean kill)
> 3) Failed : NUnit.Framework.Internal.ThreadUtilityTests.Kill
>   Native message pump was not able to be interrupted to enable a managed thread abort.
>   Expected: True
>   But was:  False
>    at NUnit.Framework.Internal.ThreadUtilityTests.AbortOrKillThreadWithMessagePump(Boolean kill)
>

ChrisMaddock:
So long as you're happy - I've reran CI. 🙂

jnm2:
Yeah. But does that make sense though? AppVeyor inserts arbitrary delays sometimes and I don't think there's anything else we can do about it, except perhaps getting the enterprise product after joining the .NET Foundation.

ChrisMaddock:
Yes, sorry. Maybe that reply didn't read how I meant it. 😄 

Sure, Appveyor can sometimes lag a little - it just seemed coincidental that the two tests you'd just added were the ones to fail. If you're happy that's all it is however, we can merge, and see how it goes.

Rather than upping the Timeout - I think we could use RetryAttribute. Unfortunately, I think https://github.com/nunit/nunit/issues/2134 might be a bug that would prevent that from working currently. 😞 

CI has passed now, merging. 👍

jnm2:
Sorry, no worries! Retry is an interesting idea, but is that really any different than doubling the timeout?

ChrisMaddock:
My thinking was, it allows for occasional machine lag, but ensures the action still can be completed within the time specified. So adding Retry would ensure the action can be completed within 1000ms, most-of-the-time. Doubling the timeout would only be ensuring the action can be completed within 2000ms - so 1500ms would be fine.

Perhaps less relevant for these ones, but I'm not sure if we use Timeout for performance purposes anywhere?

jnm2:
Ah, that makes sense for tests where we care about the amount of time it takes.

jnm2:
I'm not sure if we have any of those. I've had mixed luck with them myself at work on our build server.

PR #2296, opened 2017-07-01
Fix rename refactor gone rogue

jnm2:


ChrisMaddock:
Whoops! When did this happen?

jnm2:
2013 :-)

ChrisMaddock:
Wow, good find!

PR #2297, opened 2017-07-02
[Proposal] Hide indirect-only test assemblies from autodiscovery

jnm2:
[Proposal: please do not merge]

Autodiscovery of tests that aren't intended to be directly run has been a more or less constant irritation for me going back to my first PR. It is apparently also a source of confusion for new contributors. At the very least it's an unnecessary distraction every time you run tests in a tool such as an IDE which autodetects test assemblies.

This will benefit not just the NUnit project itself but any project which extends or integrates with NUnit.
Now it will even make sense for me to unit test custom constraints in personal projects and at work which require an indirect test assembly.

It seemed worthwhile to see how straightforward the implementation would be so that we have a concrete idea to discuss. I put it in the Internal namespace to give us more freedom of expression in the future.

/cc @OmicronPersei

### DIFF
--- a/src/NUnitFramework/framework/Internal/IndirectTestAssemblyAttribute.cs
+++ b/src/NUnitFramework/framework/Internal/IndirectTestAssemblyAttribute.cs
@@ -1,0 +1,13 @@
+﻿using System;
+
+namespace NUnit.Framework.Internal
+{
+    /// <summary>
+    /// Causes the assembly to be skipped when autodiscovering tests.
+    /// Useful for tests-on-tests scenarios so that tools that run all tests in the solution do not execute the indirect tests directly.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly)]
+    public sealed class IndirectTestAssemblyAttribute : Attribute
+    {
+    }
+}
--- a/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
+++ b/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
@@ -40,14 +40,31 @@ namespace NUnit.Framework.Api
     /// </summary>
     public class DefaultTestAssemblyBuilder : ITestAssemblyBuilder
     {
-        static Logger log = InternalTrace.GetLogger(typeof(DefaultTestAssemblyBuilder));
+        /// <summary>
+        /// Shared instance with default options.
+        /// </summary>
+        public static DefaultTestAssemblyBuilder Default { get; } = new DefaultTestAssemblyBuilder();
+
+        /// <summary>
+        /// Shared instance for running only tests intended to be run indirectly.
+        /// See <see cref="IndirectTestAssemblyAttribute"/>.
+        /// </summary>
+        public static DefaultTestAssemblyBuilder IndirectMode { get; } = new DefaultTestAssemblyBuilder(indirectMode: true);
+
+        private static readonly Logger log = InternalTrace.GetLogger(typeof(DefaultTestAssemblyBuilder));
 
         #region Instance Fields
 
         /// <summary>
         /// The default suite builder used by the test assembly builder.
         /// </summary>
-        ISuiteBuilder _defaultSuiteBuilder;
+        private readonly ISuiteBuilder _defaultSuiteBuilder;
+
+        /// <summary>.
+        /// Whether to run only tests intended to be run indirectly, or tests intended to be run directly.
+        /// See <see cref="IndirectTestAssemblyAttribute"/>.
+        /// </summary>
+        private readonly bool _indirectMode;
 
         #endregion
 
@@ -59,6 +76,12 @@ namespace NUnit.Framework.Api
         public DefaultTestAssemblyBuilder()
         {
             _defaultSuiteBuilder = new DefaultSuiteBuilder();
+        }
+
+        private DefaultTestAssemblyBuilder(bool indirectMode)
+            : this()
+        {
+            _indirectMode = indirectMode;
         }
 
         #endregion
@@ -119,6 +142,12 @@ namespace NUnit.Framework.Api
 
         private TestSuite Build(Assembly assembly, string assemblyPath, IDictionary<string, object> options)
         {
+            var isIndirectTestAssembly = assembly.GetCustomAttribute<IndirectTestAssemblyAttribute>() != null;
+            if (isIndirectTestAssembly != _indirectMode)
+            {
+                return new TestAssembly(assemblyPath);
+            }
+
             TestSuite testAssembly = null;
 
             try
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -69,10 +69,25 @@ namespace NUnit.Framework.Api
         /// <param name="idPrefix">A prefix used for all test ids created under this controller.</param>
         /// <param name="settings">A Dictionary of settings to use in loading and running the tests</param>
         public FrameworkController(string assemblyNameOrPath, string idPrefix, IDictionary settings)
+            : this(assemblyNameOrPath, idPrefix, settings, indirectMode: false)
+        {
+        }
+
+        /// <summary>
+        /// Construct a FrameworkController using the default builder and runner.
+        /// </summary>
+        /// <param name="assemblyNameOrPath">The AssemblyName or path to the test assembly</param>
+        /// <param name="idPrefix">A prefix used for all test ids created under this controller.</param>
+        /// <param name="settings">A Dictionary of settings to use in loading and running the tests</param>
+        /// <param name="indirectMode">
+        /// Whether to run only tests intended to be run indirectly, or tests intended to be run directly.
+        /// See <see cref="IndirectTestAssemblyAttribute"/>.
+        /// </param>
+        public FrameworkController(string assemblyNameOrPath, string idPrefix, IDictionary settings, bool indirectMode)
         {
             Initialize(assemblyNameOrPath, settings);
 
-            this.Builder = new DefaultTestAssemblyBuilder();
+            this.Builder = indirectMode ? DefaultTestAssemblyBuilder.IndirectMode : DefaultTestAssemblyBuilder.Default;
             this.Runner = new NUnitTestAssemblyRunner(this.Builder);
 
             Test.IdPrefix = idPrefix;
@@ -86,6 +101,22 @@ namespace NUnit.Framework.Api
         /// <param name="settings">A Dictionary of settings to use in loading and running the tests</param>
         public FrameworkController(Assembly assembly, string idPrefix, IDictionary settings)
             : this(assembly.FullName, idPrefix, settings)
+        {
+            _testAssembly = assembly;
+        }
+
+        /// <summary>
+        /// Construct a FrameworkController using the default builder and runner.
+        /// </summary>
+        /// <param name="assembly">The test assembly</param>
+        /// <param name="idPrefix">A prefix used for all test ids created under this controller.</param>
+        /// <param name="settings">A Dictionary of settings to use in loading and running the tests</param>
+        /// <param name="indirectMode">
+        /// Whether to run only tests intended to be run indirectly, or tests intended to be run directly.
+        /// See <see cref="IndirectTestAssemblyAttribute"/>.
+        /// </param>
+        public FrameworkController(Assembly assembly, string idPrefix, IDictionary settings, bool indirectMode)
+            : this(assembly.FullName, idPrefix, settings, indirectMode)
         {
             _testAssembly = assembly;
         }
--- a/src/NUnitFramework/framework/nunit.framework-2.0.csproj
+++ b/src/NUnitFramework/framework/nunit.framework-2.0.csproj
@@ -152,6 +152,7 @@
     <Compile Include="Internal\Filters\ClassNameFilter.cs" />
     <Compile Include="Internal\Filters\MethodNameFilter.cs" />
     <Compile Include="Internal\Filters\NamespaceFilter.cs" />
+    <Compile Include="Internal\IndirectTestAssemblyAttribute.cs" />
     <Compile Include="Internal\Logging\ILogger.cs" />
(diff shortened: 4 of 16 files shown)


jnm2:
Here is what VS's test window looks like:

Test explorer window showing only projects indented to be directly tested

CharliePoole:
@jnm2 Sorry, I don't get what problem you are trying to solve. An issue is good in this situation, because it kind of forces you to tell people... well... the issue. :smile: 

Since you are only changing the framework here, I'm particularly puzzled. The framework is given an assembly and runs tests from classes in that assembly. It will, of course, run test methods in base classes that are in other assemblies, but it's always supposed to do that - not just sometimes.

I'm sure the problem is obvious to you. Hope you'll make it so for me. :smile:

src/NUnitFramework/framework/Internal/IndirectTestAssemblyAttribute.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+
+namespace NUnit.Framework.Internal
+{
+    /// <summary>
+    /// Causes the assembly to be skipped when autodiscovering tests.
+    /// Useful for tests-on-tests scenarios so that tools that run all tests in the solution do not execute the indirect tests directly.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly)]
```
CharliePoole: If this is an attribute for tools to look for, why isn't it sufficient to just have the attribute? Why all the changes in the framework code itself?

jnm2:
@CharliePoole It fixes the situation we discussed in https://github.com/nunit/nunit/issues/2265 where you are in an IDE (or a build server) which autodiscovers test DLLs to execute.

We're trying to avoid this kind of thing both in VS (and on build servers where I may be developing an NUnit integration):

image

src/NUnitFramework/framework/Internal/IndirectTestAssemblyAttribute.cs
jnm2: @CharliePoole Because it's a framework API like any other which can be solved in one place instead of having to solve it in every tool.

jnm2:
It's a logical and inviting development experience too. =)

CharliePoole:
I understand the problem now. I just didn't recognize it because you are solving it in the framework, where it appears to me to be an engine issue. I agree it should not be solved in every runner, but test execution goes through the engine, specifically through a driver that maps one-to-one to each framework. If a particular assembly is known not to be a test assembly, then just don't run it. But if you tell the framework to run an assemblies tests, it should run them. The solution where the framework simply decides you __didn't__ really want to run the tests you asked to run doesn't make sense to me.

Perhaps part of my attitude here comes from the fact that I think it's incorrect for any runner to just run all the tests in an assembly. The only one I know of that does that is the VS test window. I think it's OK to compensate for it in the VS adapter, since that's built to accomodate VS but it feels wrong to change the framework because some runners tell it to run the wrong assemblies.

jnm2:
There is one more reason I implemented it at the framework level. Instead of `[IndirectTestAssembly]`, I'd actually like to be able to put `[IndirectTest]` on certain fixtures or methods (including assemblies) so that dummy assemblies aren't even necessary in the common case.

jnm2:
It's a pain not to be able to keep related things together.

jnm2:
Everyone's mental model is that all test assemblies in a solution are meant to be directly executed whenever you run tests. That's how everyone is introduced to tests and I see no problem with that whatsoever.

The only exception is when writing an NUnit integration or testing tool, or NUnit itself, which requires non-succeeding tests.

jnm2:
Hmm, look which thread just became active: https://github.com/nunit/nunit/issues/668#issuecomment-106589306

@CharliePoole 
> Is there a reason you prefer to execute the solution rather than the
assemblies directly? Understanding why folks use this approach may help us
decide how to fix it.

@tburnett80
> I am seeking a generic way to run all the tests I have for a particular solution. I have automation that I have no choice on that generates jenkins jobs and not having to know what all the test assemblies names are up front helps me considerably. [...]

I can confirm that in large projects at work, having test assemblies autodiscovered is indispensable. We've had things not get tested due to having to maintain redundant coupled information about exactly where the test assemblies are.

jnm2:
Further, if we do add a dummy assembly to test our NUnit extensions against, that's just headache. A perfect solution from my POV would be that `[IndirectTest]` attribute on methods or fixtures and once in a while on an assembly- but that's only needed for integration tests (like nunit-console's).

CharliePoole:
@jnm2 This seems to be a feature or set of features you would like to have. I think we need to discuss somewhere why it's wanted. It definitely isn't wanted for NUnit testing, because we have to test using the __real__ attributes that users use. If you introduce another type of attribute then all our tests of different kinds of test case and fixture generators then have to be doubled - once with  the indirect attribute and once without. So this doesn't help us.

It's also not something that users have asked for: what does "Indirect Test" mean to a user? That's not to say it wouldn't be something users would like. Sometimes we do things out of our own vision and present them to see if there is adoption by users. But I'm not seeing anything that explains the need.

I think we __do__ need an assembly level attribute that tells the engine "Don't treat this as a test assembly, no matter what it looks like to you" and another that says "This is an NUnit 3 test assembly, even though it doesn't look like one." 

Each of those attributes solve actual problems, albeit in an inelegant way. I think of this as being similar to the `--X86` option on the command-line. It's not generally needed because we have code to figure out whether a test needs to run under a 32 bit process. But when we are unable to figure it out, as occasionally happens, then that option is there for the user to make use of. We actually put in the option first and only later added the code to detect bitness.

I'd do the same with this set of problems that users are giving us. Provide a simplistic immediate fix and then work on improvements that end up making fix unnecessary in many cases.

In any case, we are having a discussion of a proposed feature on a PR here. I assume you did it this way thinking we would understand better if we saw the code. That's true to an extent, but the code never explains __why__ you are doing something.

ChrisMaddock:
Does `NonTestAssemblyAttribute` not cover this case, or am I missing something?

jnm2:
@ChrisMaddock 
> Does `NonTestAssemblyAttribute` not cover this case, or am I missing something?

I assumed it was an absolute, but if we can already put `[NonTestAssemblyAttribute]` in our testdata assemblies in such a way that our tests don't fail, then I'm all for it!

> R#

ReSharper does not involve any NUnit code in NUnit test discovery. It searches for attributes by namepsace and name and makes its own decisions, which is both 1) insanely fast because it's possible to show proper UI in the source code even before it's compiled, and 2) completely incorrect all the time and a major annoyance for me. For example, they think that my test's actual name is `{{m}}(text)` and constantly show it as never run.

jnm2:
ReSharper also doesn't respect `NonTestAssembly` but I was already planning to contact them as soon as we agree on a direction here.

(Aside: I can see ReSharper's side of things though. Worth a different discussion with them on how we could refactor to be able to serve their needs while still remaining in control of the logic, to make people's experiences better.)

ChrisMaddock:
> I assumed it was an absolute

Is this an Americanism again? 😛 I don't know what you mean!

If I'm understanding your objectives right, imo, this is too close to what NonTestAssemblyAttribute is, that we should make the changes to make that work in this case, instead of implementing a second attribute. As far as I can think, that just means implementing the `--skipNonTestAssemblies` console flag for the adapter - but I may be wrong there. 😄

>  It searches for attributes by namespace and name and makes its own decisions, which is both 1) insanely fast because it's possible to show proper UI in the source code even before it's compiled

That's interesting! Especially as I wouldn't call my R# test runner anything like 'insanely fast'. 😄 I've often thought about putting a little work into parallelising Explore - to see if we couldn't get tests to be discovered a little faster.

ChrisMaddock:
Alternatively, it could be suggested that the adapter should use EnginePackageSettings.SkipNonTestAssemblies by default - as the user has no control over which assemblies are passed to it.

That was the argument behind the `--skipNonTestAssemblies` flag for the console runner.

jnm2:
@ChrisMaddock 

Lol, I have no idea if calling something an “absolute” is an Americanism! The Oxford English Dictionary defines the noun as, “A value or principle which is regarded as universally valid or which may be viewed without relation to other things,” so probably not. 😛 

Anyway, I assumed that `NonTestAssemblyAttribute` was absolutely 😛 enforced, so that putting it on the testdata assemblies would make them invisible even for internal testing purposes. Either way it sounds semantically invalid because they are not in fact non-test assemblies, but I'm willing to adulterate the meaning in order to hide the testdata assemblies from tools if you don't think the idea of `[IndirectTestAssembly]` or the idea of `[IndirectTest]` on classes and methods (and assemblies) is worth it.

jnm2:
> Especially as I wouldn't call my R# test runner anything like 'insanely fast'.

Lol! The running is not fast. The discovery is fast, inserting the correct icons in the code as you type.

jnm2:
> Alternatively, it could be suggested that the adapter should use EnginePackageSettings.SkipNonTestAssemblies by default - as the user has no control over which assemblies are passed to it.

The fact that this isn't already true today makes me very sad. Why would the engine not have that on by default unless overridden by a runner, and why wouldn't every runner leave that as the default unless overridden by an explicit command from the user?

CharliePoole:
So the default should be that the engine refuses to run tests in an assembly when I call Run?

jnm2:
> So the default should be that the engine refuses to run tests in an assembly when I call Run?

If I didn't want that exact outcome, why would I have written `[assembly: NonTestAssembly]`?

jnm2:
In my mind, `[NonTestAssembly]` should cause tests to not be discovered in exactly the same way that `[Test]` normally causes them to be discovered.

CharliePoole:
Your point makes partial sense to me. 

However, the idea of `NonTestAssembly` (not yet implemented AFAIK) is that you would put it on an assembly to indicate that the assembly is not a test assembly and should not be examined for tests even though it has a reference the nunit framework.

IF you now put the attribute on an assembly that __does__ contain tests, you are creating a contradiction. I don't think anyone should do that and expect that they can predict what would happen. Maybe NUnit will obey the attribute and not look. That seems like the easiest thing to implement. However, that's different from __documenting__ and __guaranteeing__ the behavior. I would simply tell users not to do that.

If __we__ want to use it for our own tests, that's another matter. We are allowed to rely on accidents of implementation of NUnit for our own tests because we control the implementation.

If the implementation is in the framework - which I would not expect - then we could not use the attribute on an assembly like our test data assembly. If it's in the engine, we would be able to use it in our tests __except__ when we are loading a full assembly. That's not a big loss, in my opinion.

jnm2:
> If we want to use it for our own tests, that's another matter. We are allowed to rely on accidents of implementation of NUnit for our own tests because we control the implementation.

The other use case for the `[IndirectTest]` attribute is for people like me at work, or Remco with NCrunch, or anyone building an integration or custom extension of NUnit where you must have indirect tests because testing requires indirect test failures. Then it's no longer about us and whether we can rely on accidents of implementation for our own tests.

> If it's in the engine, we would be able to use it in our tests except when we are loading a full assembly. That's not a big loss, in my opinion.

I hope to do exactly this with nunit-console's mock-assembly already.

ChrisMaddock:
> However, the idea of NonTestAssembly (not yet implemented AFAIK)

Didn't you implement this, @CharliePoole? 😄 

https://github.com/nunit/nunit/blob/7b87bdc7b48ee1b13023dd282efea449c2d66b81/src/NUnitFramework/framework/Attributes/NonTestAssemblyAttribute.cs

IIRC, we had a long discussion about this when it was implemented. Peoples concern's were that one developer might add the attribute, and then a future developer might not be able to work out why the tests weren't running - hence why people wanted both an attribute and an enginesetting - belt and braces sorta thing.

jnm2:
> a future developer might not be able to work out why the tests weren't running

Do we have logging levels? Info-level log message should be perfect for this, right?

ChrisMaddock:
I'd agree a "skipped because NonTestAssemblyAttribute" would be enough. 

The original issue is here: https://github.com/nunit/nunit-console/pull/150 I believe the relevant bit is the bit on the PR.

CharliePoole:
I knew I had written the code, but I thought we had not merged it because there was so much opposition. It's hard to keep track of some of these issues!

CharliePoole:
@ChrisMaddock Regarding "a future developer" = my intention was that the attribute would be used by somebody with a third-party framework that referenced the nunit framework. In that case, I don't see that the possibility of confusion is very great.

But if used on an assembly with second-level tests, as @jnm2 suggests, that possibility exists. For example, we have to remember not to put tests we expect to have executed into the test-assembly. If we do, they won't run. This seems a slightly higher but still pretty low probability of occurence.

jnm2:
> For example, we have to remember not to put tests we expect to have executed into the test-assembly. If we do, they won't run. This seems a slightly higher but still pretty low probability of occurence.

So let me strike this proposal in favor of a `[IndirectOnly]` attribute in the internal namespace, applicable to methods, classes, and assemblies. I would find locality of code very convenient. We could literally have the second-level test method right next to the direct test using it. No more complicated assembly-level runner stuff.

CharliePoole:
@jnm2 Essentially, we use second-level tests for two reasons:

1. To test that our attributes properly set up tests.

2. To test Failure conditions

We can't use IndirectOnly in the first case, becaise we would not be testing the thing we need to test. Rather, we would be testing that Test with IndirectOnly works.

It could give us some help in the second case.

Note that we don't always put second level tests in a separate assembly. For some kinds of tests we just create a TestMethod for a method without any attribute and use that.

Maybe it's just that I am used to it. I've been using test-assembly to hold second level tests without a glitch for a long time. It has the advantage that I can sometimes directly run a particular test in order to look at the error display.

ChrisMaddock:
Where did this one get to? 🙂 

It would be good to solve the issue. My opinion is as yet unconvinced that this is sufficiently different to `[NonTestAssembly]` that we need to introduce a new attribute for it - I think allowing `NonTestAssembly` to be used via the adapter (or just always on by default) would be a better solution.

CharliePoole:
I think this should be closed.

rprouse:
Agreed, this isn't the way we want to solve this issue.

jnm2:
:+1:

Issue #2299, opened 2017-07-04
NUnitLite NuGet package no longer installs NUnit NuGet package


ChrisMaddock:
The NUnit Lite package no longer installs the NUnit Framework package as a dependency.

We added empty dependency groups to the nuspec, to fix https://github.com/nunit/nunit/issues/2165 - but missed that the NUnitLite package _should_ have a dependency on all platforms of the NUnit package. 😞 

![image](https://user-images.githubusercontent.com/4837132/27834023-27ac41ec-60cd-11e7-8505-76314537810b.png)

CharliePoole:
When did the dependency get dropped? Can we fix it at the package level without a new release? I'd call this critical.

ChrisMaddock:
This would have been broken since 3.7.0, which has been released over a month now. I guess there's not that many people adding NUnitLite to new projects!

I debated 'critical' but went for 'pri:high' as a) it's not been reported for a month and b) there's a simple workaround. I don't think we can fix it without a release, no.

CharliePoole:
Too bad NuGet doesn't support a way to do package releases containing the same software. Apt does that with a special suffix.

rprouse:
Odd, NUnit is still listed as an overall dependency and we still have the blank 

```xml 
      <group>
          <dependency id="NUnit" version="[$version$]" />
      </group>
```

I guess that only applies to not specified targets and we need to duplicate it for each?

We could grab the binaries out of the existing NuGet package and release an updated version with the dependencies fixed using the same binaries. I might even be able to edit the existing one using NuGet package manager for a hotfix.

rprouse:
@ChrisMaddock I added myself as a second assignee on this. Let's coordinate. Maybe you can do the proper fix for the next release and I can see if I can get a hotfix out?

CharliePoole:
I wondered about the use of $version$ in other places in the file besides the primary version entry for the package. Are you sure that works as a general substitution?

rprouse:
@CharliePoole that dependency hasn't changed and it looks right in the image @ChrisMaddock posted above. Notice the first dependency is correct, it just doesn't seem to get picked up because of the target specific dependencies.

CharliePoole:
Seems like a bug in NuGet, which we will have to work around.

ChrisMaddock:
@rprouse Sounds good. On my phone, but there's a PR from yesterday for the in-repo fix.

rprouse:
[...]

PR #2300, opened 2017-07-04
Add explicit NUnit package dependencies for all target frameworks

ChrisMaddock:
Fixes https://github.com/nunit/nunit/issues/2299

I've tested this under VS 2015/NuGet 3.5/packages.config. I haven't tried any other versions of anything, but it seems logical given our current understanding.


CharliePoole:
LGTM

PR #2304, opened 2017-07-06
Revert accidental doc removal

ChrisMaddock:
Another @oznetmaster special -https://github.com/nunit/nunit/commit/acdc5477eae0aa489331199ec07053c9e43b69f2#diff-773ff02e2f8e671f005e32bcf2e9d03eL44

#2296 seems to have caught something it couldn't.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/ContainsConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/ContainsConstraint.cs
@@ -41,6 +41,7 @@ namespace NUnit.Framework.Constraints
         /// <summary>
         /// Initializes a new instance of the <see cref="ContainsConstraint"/> class.
         /// </summary>
+        /// <param name="expected">The expected value contained within the string/collection.</param>
         public ContainsConstraint(object expected)
         {
             this._expected = expected;


ChrisMaddock:
Thanks @mikkelbu 🙂

mikkelbu:
@ChrisMaddock Should it not be an issue that is assigned to a milestone? (this is the only PR with a milestone AFAICT)

ChrisMaddock:
If there's both Issue and PR, assign just the issue. If there's only a PR, assigning the PR is fine. 

The tool that generates the release notes doesn't mind which one is there, but of course we don't want any duplicates. 🙂

mikkelbu:
Understood 👍

CharliePoole:
Having an issue is only important for the humans, especially if there is a need to decide what to do. Also OOF you see a bug and don't plan to fix it in the next few minutes.

jnm2:
Heh, it wasn't accidental. But :+1:

ChrisMaddock:
Ah, ok! I thought it was an odd regex malfunction. ;-) Either way - it's documented !

jnm2:
> I thought it was an odd regex malfunction.

People say that about me a lot. 😁 ;-)

PR #2310, opened 2017-07-09
TestCaseSource unable to pass one element byte array

### LINKED ISSUE #1851: TestCaseSource unable to pass one element byte array
**NUnit version:**  3.5.0
**Runner**: NUnit3TestAdapter 3.4.1
**Environment**: Visual Studio 2015 Pro Update 1

I'm having troubles using one element byte array passed to NUnit test with `TestCaseSourceAttribute`.

How to reproduce:

```
public class Test
    {
        private static List<byte[]> _byteArrays = new List<byte[]>
        {
            new byte[] { 80 }, // - works fine with commented out
            new byte[] { 80, 75 },
            new byte[] { 80, 75, 03 }
        };

        [Test]
        [TestCaseSource("_byteArrays")]
        public void TestStuff(byte[] byteArray)
        {
            Assert.IsFalse(false);
        }
    }
```

When you run `TestStuff` it will attempt to test things in seperate tests like this:

nunit

It clearly seems that first element is treated as an `System.Byte` instead of `System.Byte[]`.

Error produced by NUnit tells it clearly:

> Result Message:System.ArgumentException : Object of type 'System.Byte' cannot be converted to type 'System.Byte[]'

Any ideas on fixing this?
Thanks!

StanEgo:
While working on #2104 I have also optimized TestCaseSourceAttribute and additionally check it against this issue. Looks like it is fixed.

Fixes #1851

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
@@ -189,48 +189,29 @@ namespace NUnit.Framework
 
                         if (parms == null)
                         {
+                            object[] args = null;
+
                             // 3. An array was passed, it may be an object[]
                             //    or possibly some other kind of array, which
                             //    TestCaseSource can accept.
-                            var args = item as object[];
-                            if (args == null && item is Array)
+                            var array = item as Array;
+                            if (array != null)
                             {
-                                Array array = item as Array;
-                                int numParameters = method.GetParameters().Length;
-                                if (array != null && array.Rank == 1 && array.Length == numParameters)
+                                // If array has the same number of elements as parameters
+                                // and it does not fit exactly into single existing parameter
+                                // we believe that this array contains arguments, not is a bare
+                                // argument itself.
+                                var parameters = method.GetParameters();
+                                var argsNeeded = parameters.Length;
+                                if (argsNeeded > 0 && argsNeeded == array.Length && parameters[0].ParameterType != array.GetType())
                                 {
-                                    // Array is something like int[] - convert it to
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
```diff
@@ -189,48 +189,29 @@ private IEnumerable<ITestCaseData> GetTestCasesFor(IMethodInfo method)
...
+                                var parameters = method.GetParameters();
+                                var argsNeeded = parameters.Length;
+                                if (argsNeeded > 0 && argsNeeded == array.Length && parameters[0].ParameterType != array.GetType())
                                 {
-                                    // Array is something like int[] - convert it to
-                                    // an object[] for use as the argument array.
                                     args = new object[array.Length];
-                                    for (int i = 0; i < array.Length; i++)
+                                    for (var i = 0; i < array.Length; i++)
                                         args[i] = array.GetValue(i);
```
jnm2: `array.CopyTo(args, 0);` is also an option.

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -142,6 +142,12 @@ public void SourceMayReturnArgumentsAsIntArray(int n, int d, int q)
             Assert.AreEqual(q, n / d);
         }
 
+        [Test, TestCaseSource("MyArrayData")]
+        public void SourceMayReturnArrayForArray(int[] array)
+        {
+            Assert.That(true);
```
jnm2: `Assert.Pass();` or an empty method body are also options.

PR #2312, opened 2017-07-10
Prevent crash on invalid --result parsing in NUnitLite

ChrisMaddock:
This prevents a crash in the NUnitLite runner when an invalid `--result` command is specified. It essentially ports https://github.com/nunit/nunit-console/pull/257 to the NUnitLite runner.

The major difference is that NUnitLite would previously read the transform specifier, but made no use of it. In line with our vision of not having NUnitLite and the Console have exact matching functionality, I stripped the parsing logic for this out, so it a user specifies a transform, it will be seen as invalid.

### DIFF
--- a/src/NUnitFramework/nunitlite.tests/CommandLineTests.cs
+++ b/src/NUnitFramework/nunitlite.tests/CommandLineTests.cs
@@ -413,9 +413,9 @@ namespace NUnitLite.Tests
         }
 #endif
 
-#endregion
+        #endregion
 
-#region EngineResult Option
+        #region EngineResult Option
 
         [Test]
         public void FileNameWithoutResultOptionLooksLikeParameter()
@@ -509,9 +509,9 @@ namespace NUnitLite.Tests
             Assert.That(options.ErrorMessages, Has.Exactly(1).Contains("invalid output spec").IgnoreCase);
         }
 
-    #endregion
+        #endregion
 
-    #region Explore Option
+        #region Explore Option
 
         [Test]
         public void ExploreOptionWithoutPath()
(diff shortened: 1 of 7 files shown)


src/NUnitFramework/nunitlite.tests/CommandLineTests.cs
```diff
@@ -522,9 +500,18 @@ public void NoResultSuppressesAllResultSpecifications()
...
+        [Test]
+        public void InvalidResultSpecRecordsError()
+        {
+            var options = new NUnitLiteOptions("test.dll", "-result:userspecifed.xml;format=nunit2;format=nunit3");
+            Assert.That(options.ResultOutputSpecifications, Has.Exactly(1).Items
+                    .And.Exactly(1).Property(nameof(OutputSpecification.OutputPath)).EqualTo("TestResult.xml"));
+            Assert.That(options.ErrorMessages, Has.Exactly(1).Contains("invalid output spec").IgnoreCase);
+        }
+
+    #endregion
```
mikkelbu: The coding standards is currently empty regarding regions (see e.g. https://github.com/nunit/docs/wiki/Coding-Standards#use-of-regions) - and personally I don't like them :) - but I think that region pairs should be indented to the same level.

src/NUnitFramework/nunitlite.tests/CommandLineTests.cs
```diff
@@ -522,9 +500,18 @@ public void NoResultSuppressesAllResultSpecifications()
...
 
-        #region Explore Option
+    #region Explore Option
```
mikkelbu: The coding standards is currently empty regarding regions (see e.g. https://github.com/nunit/docs/wiki/Coding-Standards#use-of-regions) - and personally I don't like them :) - but I think that region pairs should be indented to the same level.

ChrisMaddock:
Thanks @mikkelbu 🙂 

I fixed the #region alignment to be consistent in the file at least. Will merge once CI has passed again.

rprouse:
I prefer indented regions too, but unfortunately adding various `#if` blocks causes `#region` to become unindented in VS 2015 and 2017, so I have nearly given up trying to fix that. If anyone can show me the setting that stops that I will be eternally grateful 😄 

Looks good...

rprouse:
Failed AppVeyor build is the unrelated .NET 3.5 failure that we already have an issue for, so I am merging. Very odd that it is only 3.5 failing since we have so few `#if` blocks for 3.5 (if any?).

Issue #2201, opened 2017-05-29
Invalid platform name passed to PlatformAttribute should mark test NotRunnable


ChrisMaddock:
From https://github.com/nunit/nunit/pull/2200 - if an invalid platform name is passed into PlatformAttribute, the test is currently skipped - the same as if a platform the test is not currently running on was passed in.

An invalid platform name would always be a mistake, so in my opinion should mark the test NotRunnable, and fail the test run.

Thoughts?

CharliePoole:
Agree on included platforms, what about excluded?

ChrisMaddock:
I think this should cover all lists. Excluding an invalid platform would currently skip the test on all platforms, which is potentially a hard-to-spot, wrong-side failure.

CharliePoole:
I thought excluding an invalid platform caused it to run. I'm on my phone so I can't check.

ChrisMaddock:
I haven't tried, it, but it looks like anything invalid throws an ArgumentException, which is caught by `IsPlatformSupported()`, which then returns false.

https://github.com/nunit/nunit/blob/077aa758275edaa069aa16e4ae36a45421adc022/src/NUnitFramework/framework/Internal/PlatformHelper.cs#L135

Either way, I still would prefer an invalid test, over the test running on a platform I didn't intent it to. 😄

jnm2:
My opinion on correctness issues is always that it should fail as soon and as clearly as possible. :smile:

ChrisMaddock:
Relabelling this as an easyfix bug.

PlatformHelper currently throws and catches an Exception, for invalid framework names. In my opinion, that exception should instead be caught within PlatformAttribute, which can then mark the test as NotRunnable, with an appropriate reason.

david-proctor:
Is this still up for grabs? I'd like to take it if so.

ChrisMaddock:
Great - it's yours! Thanks!

rprouse:
@david-proctor I've sent you an invite to the contributors team so that we can assign this issue to you. Thanks for the help.

david-proctor:
@rprouse Thanks! I've put a commit together but I don't think I have push access to the repo. If someone could grant that I'll start a pull request.

rprouse:
@david-proctor normally we start new contributors out working in forks of the repo to get started, but since you are Canadian (joking) I will temporarily add you to the framework team. We would normally add you after a couple of PRs and activity in the issues, so if you are planning on sticking around I will leave you on the framework team.

[...]

PR #2314, opened 2017-07-11
Tests with invalid platform names are marked Not Runnable

david-proctor:
Resolves issue #2201: Invalid platform name passed to PlatformAttribute should mark test NotRunnable.

Based on Chris Maddock's suggestion in the comments on that issue. Previously, passing an invalid platform name to a PlatformAttribute would cause the test to be skipped. This change alters where the exception generated in this case is handled, allowing the test to be marked NotRunnable instead.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
@@ -56,12 +56,20 @@ namespace NUnit.Framework
         /// <param name="test">The test to modify</param>
         public void ApplyToTest(Test test)
         {
-            if (test.RunState != RunState.NotRunnable && 
-                test.RunState != RunState.Ignored && 
-                !platformHelper.IsPlatformSupported(this))
+            try
             {
-                test.RunState = RunState.Skipped;
-                test.Properties.Add(PropertyNames.SkipReason, platformHelper.Reason);
+                if (test.RunState != RunState.NotRunnable &&
+                    test.RunState != RunState.Ignored &&
+                    !platformHelper.IsPlatformSupported(this))
+                {
+                    test.RunState = RunState.Skipped;
+                    test.Properties.Add(PropertyNames.SkipReason, platformHelper.Reason);
+                }
+            }
+            catch (Exception ex)
+            {
+                test.RunState = RunState.NotRunnable;
+                test.Properties.Add(PropertyNames.SkipReason, ex.Message);
             }
         }
 
--- a/src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
@@ -56,27 +56,23 @@ namespace NUnit.Framework
         /// <param name="test">The test to modify</param>
         public void ApplyToTest(Test test)
         {
-            if (!IsPlatformSupportedAndValid(test) &&
-                test.RunState != RunState.NotRunnable &&
+            if (test.RunState != RunState.NotRunnable &&
                 test.RunState != RunState.Ignored)
             {
-                test.RunState = RunState.Skipped;
-                test.Properties.Add(PropertyNames.SkipReason, platformHelper.Reason);
+                try
+                {
+                    if (!platformHelper.IsPlatformSupported(this))
+                    {
+                        test.RunState = RunState.Skipped;
+                        test.Properties.Add(PropertyNames.SkipReason, platformHelper.Reason);
+                    }
+                }
+                catch (InvalidPlatformException ex)
+                {
+                    test.RunState = RunState.NotRunnable;
+                    test.Properties.Add(PropertyNames.SkipReason, ex.Message);
+                }
             }
-        }
-
-        private bool IsPlatformSupportedAndValid(Test test)
-        {
-            try
-            {
-                return platformHelper.IsPlatformSupported(this);
-            }
-            catch (InvalidPlatformException ex)
-            {
-                test.RunState = RunState.NotRunnable;
-                test.Properties.Add(PropertyNames.SkipReason, ex.Message);
-            }
-            return false;
         }
 
         #endregion
--- a/src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
@@ -56,21 +56,27 @@ namespace NUnit.Framework
         /// <param name="test">The test to modify</param>
         public void ApplyToTest(Test test)
         {
+            if (!IsPlatformSupportedAndValid(test) &&
+                test.RunState != RunState.NotRunnable &&
+                test.RunState != RunState.Ignored)
+            {
+                test.RunState = RunState.Skipped;
+                test.Properties.Add(PropertyNames.SkipReason, platformHelper.Reason);
+            }
+        }
+
+        private bool IsPlatformSupportedAndValid(Test test)
+        {
             try
             {
-                if (test.RunState != RunState.NotRunnable &&
-                    test.RunState != RunState.Ignored &&
-                    !platformHelper.IsPlatformSupported(this))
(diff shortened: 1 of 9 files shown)


src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
```diff
@@ -56,12 +56,20 @@ public class PlatformAttribute : IncludeExcludeAttribute, IApplyToTest
...
+                {
+                    test.RunState = RunState.Skipped;
+                    test.Properties.Add(PropertyNames.SkipReason, platformHelper.Reason);
+                }
+            }
+            catch (Exception ex)
+            {
+                test.RunState = RunState.NotRunnable;
+                test.Properties.Add(PropertyNames.SkipReason, ex.Message);
             }
```
CharliePoole: what if an exception is thrown for some other reason?

rprouse: Aren't those other reasons also valid reasons to mark the test as not runnable? If we expect them to bubble up and set the test to error, then maybe we shouldn't even have the catch here and just let an invalid platform error the test?

CharliePoole: Here we are assuming that every exception gives a simple, user-focused message without a need to display the stack-trace or show inner exceptions. So we need to catch some exception type here, but not others. Maybe we need a special exception.

rprouse: Weird, @CharliePoole I saw your response in email, but it isn't here. I agree with the idea that some exceptions might need more info such as stack traces and that a good solution would be a new exception. How about adding an `InvalidPlatformException`?

david-proctor: An `InvalidPlatformException` sounds good to me.

Earlier, we caught all exceptions in `PlatformHelper`, so in addition to an `InvalidPlatformException` I expect we will still want to have a general catch for all types of exceptions here.

I'll update this pull request with these changes as soon as I'm able.

CharliePoole: @david-proctor There's already a catch higher up in the chain that takes care of general exceptions that occur when trying to discover tests. Here we should deal with the exception we know about.

jnm2: Even better, what if we don't use `InvalidPlatformException` for flow control at all and instead return an enum value of `Supported`, `NotSupported`, `InvalidName`?

jnm2: If we stick with the exception route, I'd want `isPlatformSupported = platformHelper.IsPlatformSupported(this)` to be the only thing inside the try block.

CharliePoole: @jnm2 IMO this is a class that doesn't have much useful life left. So, while I think your idea of avoiding an exception is valid, I'd go with the simpler approach of transforming the exception we already generate into it's own special type and do a better design when we obsolete PlatformAttribute.

src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
```diff
@@ -56,21 +56,27 @@ public class PlatformAttribute : IncludeExcludeAttribute, IApplyToTest
         /// <param name="test">The test to modify</param>
         public void ApplyToTest(Test test)
         {
+            if (!IsPlatformSupportedAndValid(test) &&
+                test.RunState != RunState.NotRunnable &&
+                test.RunState != RunState.Ignored)
```
CharliePoole: The test of RunState has to come first. The Runstate may have been set to NotRunnable or Ignored by other attribute that executed before this one. In that case, we don't want to check the platform at all.

I see why you did it that way, since this attribute may itself set the test to non-runnable. But  that's the case with almost every attribute. I would keep the try / catch in this method and eliminate the new method so that everything is visible in one place - something like this...

```c
if (test.RunState != RunState.NotRunnable &&
    test.RunState != Runstate.Ignored)
    {
        try
        {
            if (!platformHelper.IsPlatformSupported(test))
            {
                test.Runstate = Runstate.Skipped;
                test.Properties.Add(PropertyNames.SkipReason, platformHelper.Reason);
            }
        }
        catch(InvalidPlatformException ex)
        {
            test.RunState = Runstate.NotRunnable;
            test.Properties.Add(PropertyNames.SkipReason, ex.Message);
        }
    }
```

This example (forum code, not tested) just rearranges what you have already so that there is both a pre-check and a post-check for non-runnability.

david-proctor:
I see! That makes perfect sense - I'll re-arrange it.

src/NUnitFramework/tests/Internal/PlatformDetectionTests.cs
```diff
@@ -429,9 +429,9 @@ public void PlatformAttribute_IncludeAndExclude()
         public void PlatformAttribute_InvalidPlatform()
         {
             PlatformAttribute attr = new PlatformAttribute( "Net-1.0,Net11,Mono" );
-            Assert.IsFalse( winXPHelper.IsPlatformSupported( attr ) );
-            Assert.That( winXPHelper.Reason, Does.StartWith("Invalid platform name"));
-            Assert.That( winXPHelper.Reason, Does.Contain("Net11"));
```
ChrisMaddock: We seem to have lost this specific assertion on line 434: that the message contains the invalid platform name. Can we add that back in somewhere? It could be here - but actually, I agree with the changes you've made - at this level, we only care that the expected type of exception is being thrown. 

I think I nicer solution might be to add this line to your new 'InvalidPlatformAttributeIsNotRunnable' test - and check that the skipreason contains the invalid platform name. 🙂

src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
```diff
@@ -56,12 +56,22 @@ public class PlatformAttribute : IncludeExcludeAttribute, IApplyToTest
...
-                test.Properties.Add(PropertyNames.SkipReason, platformHelper.Reason);
+                try
+                {
+                    if (!platformHelper.IsPlatformSupported(this))
+                    {
+                        test.RunState = RunState.Skipped;
+                        test.Properties.Add(PropertyNames.SkipReason, platformHelper.Reason);
+                    }
+                }
+                catch (InvalidPlatformException ex)
```
jnm2: Since we went this route, it would make the code cleaner if `isPlatformSupported = platformHelper.IsPlatformSupported(this)` was the only thing inside the `try` block so that people understand the API usage by looking at the call site.

src/NUnitFramework/framework/Internal/InvalidPlatformException.cs
```diff
@@ -0,0 +1,67 @@
+﻿// ***********************************************************************
+// Copyright (c) 2017 Charlie Poole
```
jnm2: Good catch!

david-proctor:
I've made the changes listed in the recent reviews, but I also noticed an AppVeyor failure on the previous commit:

1) Failed : NUnit.Framework.Internal.ThreadUtilityTests.Kill
Test exceeded Timeout value of 1000ms

I can't reproduce this on my dev machine. Is there anything we want to do to address this failure?

CharliePoole:
LGT*M

david-proctor:
All right - looks like the build passed, so I have merged. Thanks for all the assistance and feedback, everyone!

jnm2:
> I can't reproduce this on my dev machine. Is there anything we want to do to address this failure?

Thanks for the heads up. I double checked; the AppVeyor error happened on your commit based of e9e391d51d15bb89ee07566c74591e4723702a93 in master, and https://github.com/nunit/nunit/pull/2295 was merged shortly after that commit, so as far as we know this won't happen again.

Issue #2313, opened 2017-07-10
Incorrect xmldoc on RetryAttribute


ChrisMaddock:
The xmldoc on RetryAttribute has been copy/pasted from RepeatAttribute. This should be corrected.

https://github.com/nunit/nunit/blob/7b87bdc7b48ee1b13023dd282efea449c2d66b81/src/NUnitFramework/framework/Attributes/RetryAttribute.cs

We already have some doc for RetryAttribute - so this could be a good issue for a first time contributor. 🙂 

https://github.com/nunit/docs/wiki/Retry-Attribute

MhdTlb:
Can I work on this issue?

mikkelbu:
Sure. I've assigned the issue to you.

jnm2:
@mikkelbu Fyi, we remove the `up-for-grabs` when we assign people and try to add `up-for-grabs` (have to document first) whenever it's approved on the backlog and not assigned.

PR #2317, opened 2017-07-11
Add proper xmldoc for RetryAttribute

MhdTlb:
Correct xmldoc to be similar to the documentation

### DIFF
--- a/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
@@ -35,15 +35,15 @@ namespace NUnit.Framework
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
     public class RetryAttribute : PropertyAttribute, IWrapSetUpTearDown
     {
-        private int _count;
+        private int _tryCount;
 
         /// <summary>
         /// Construct a <see cref="RetryAttribute" />
         /// </summary>
-        /// <param name="count">The maximum number of times the test should be run if it fails</param>
-        public RetryAttribute(int count) : base(count)
+        /// <param name="tryCount">The maximum number of times the test should be run if it fails</param>
+        public RetryAttribute(int tryCount) : base(tryCount)
         {
-            _count = count;
+            _tryCount = tryCount;
         }
 
         #region IWrapSetUpTearDown Members
@@ -55,7 +55,7 @@ namespace NUnit.Framework
         /// <returns>The wrapped command</returns>
         public TestCommand Wrap(TestCommand command)
         {
-            return new RetryCommand(command, _count);
+            return new RetryCommand(command, _tryCount);
         }
 
         #endregion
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/framework/Attributes/RetryAttribute.cs
```diff
@@ -22,7 +22,7 @@
 // ***********************************************************************
 
 // TODO: Rework this
-// RepeatAttribute should either
+// RetryAttribute should either
 //  1) Apply at load time to create the exact number of tests, or
 //  2) Apply at run time, generating tests or results dynamically
```
mikkelbu: I think that we should delete all the comments regarding the TODO. As far as I can tell the comment is about `RepeatAttribute` and not `RetryAttribute`, e.g. we cannot at load time calculate the exact number of tests.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
```diff
@@ -36,16 +36,16 @@
...
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
     public class RetryAttribute : PropertyAttribute, IWrapSetUpTearDown
     {
         private int _count;
 
         /// <summary>
-        /// Construct a RepeatAttribute
+        /// Construct a RetryAttribute
         /// </summary>
         /// <param name="count">The number of times to run the test</param>
```
mikkelbu: Perhaps we should change the description of `count` to something like "The maximum number of times the test should be rerun if it fails".

mikkelbu: We could also remove the same TODO in nunit/src/NUnitFramework/tests/Attributes/RetryAttributeTests.cs

Ps. You should remember to reference the issue number in the commitmessage or PR, like fixes #2313 such that the issue is automatically closed.

mikkelbu:
@nunit/framework-team can I also kick the buildserver (aka Travis) or does that require special rights?

CharliePoole:
It does. Kicked!

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
jnm2: Also we should definitely start using `<see cref="RetryAttribute" />` here.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
```diff
@@ -36,16 +36,16 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// RepeatAttribute may be applied to test case in order
-    /// to run it multiple times.
+    /// RetryAttribute is used on a test method to specify that it should
```
jnm2: We should definitely start using <see cref="RetryAttribute" /> here.

ChrisMaddock:
@mikkelbu

I thought being a member of the framework team gave you access to restart builds? You'll need to log in to travis with your GitHub credentials. Once you've done that, there should be a 'restart build' button on the build page? This is what it looks like on mobile.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
```diff
@@ -36,18 +29,18 @@
...
     public class RetryAttribute : PropertyAttribute, IWrapSetUpTearDown
     {
         private int _count;
 
         /// <summary>
-        /// Construct a RepeatAttribute
+        /// Construct a <see cref="RetryAttribute" />
         /// </summary>
-        /// <param name="count">The number of times to run the test</param>
+        /// <param name="count">The maximum number of times the test should be rerun if it fails</param>
```
jnm2: Oh. This should say <code>The maximum number of times the test should be **run**</code>, not `rerun`.

jnm2:
Well, we have an important preexisting problem to fix. I don't know if you want to do it in this PR or if we'll need a PR right after:

https://github.com/MhdTlb/nunit/blob/1446e65dc054dbc0711422b82baf70cace3f6f17/src/NUnitFramework/framework/Attributes/RetryAttribute.cs#L70-L80

Parameter `retryCount` and field `_retryCount` need to be renamed `tryCount` and `_tryCount`, and the param XML for the parameter needs to match the param XML for `RetryAttribute`'s parameter.

Also, it probably behooves us to rename `RetryAttribute`'s `count` parameter to `tryCount` just for clarity.

MhdTlb:
@jnm2  I will try to fix it within this PR

PR #2320, opened 2017-07-14
Editorconfig charset and coding standards

### LINKED ISSUE #2319: Add .editorconfig to set file encodings so that people don't have to think about it
These days we want every file to be UTF-8 with no byte order mark. It's universally understood and compatible, unlike the ANSI encoding Visual Studio will use by default. This is a typical problem any time you copy and paste special characters from the clipboard into VS. It results in tests running fine locally and blowing up in Travis (see https://github.com/nunit/nunit/pull/2294#issuecomment-315223398).

Obviously, this is not something we want people to have to even think about if we can possibly avoid it. So many things have native `.editorconfig` support that it is worthwhile to start one just to override the charset of every file on save.

(EditorConfig helps developers define and maintain consistent coding styles between different editors and IDEs. http://editorconfig.org/)

jnm2:
Fixes #2319

Removed `end_of_line = crlf` because git already normalizes everything to `lf` internally and some people are on different platforms where `crlf` doesn't make sense. If certain files like the .sh files need to be `lf`-only, we do that in `.gitattributes`.

Added `insert_final_newline = true` because otherwise GitHub and git complain every time you view files and I figured it was a nice default, but if you see a problem with it we can remove it.

#### Coding standard

Most of these rules are already defaults in VS and are just useful in case anyone has customized VS to do things differently for them.
I made everything that goes against the coding standard show as an error. It doesn't prevent any activities like building, but as we discussed, things in the coding standard are things we're prepared to request changes on at review time. If we don't like certain rules annotating code this way, we should start by revising the coding standard and afterwards update the .editorconfig.

I did not see any reason to add non-coding-standard warnings, suggestions, and silent preferences (which affect autocomplete and code generation). If we ever see a reason to add things like that, it should be kept in a separate section from the coding standard.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,8 +11,7 @@ root = true
 indent_style = space
 indent_size = 4
 insert_final_newline = true
-# VS 15.0–15.2 misinterpret as utf-8-bom https://developercommunity.visualstudio.com/content/problem/22922
-# charset = utf-8
+charset = utf-8
 
 [*.{proj,csproj,vcxproj,xproj,json,config,nuspec,xml}]
 indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,28 +10,77 @@ root = true
 [*]
 indent_style = space
 indent_size = 4
-end_of_line = crlf
+insert_final_newline = true
+# VS 15.0–15.2 misinterpret as utf-8-bom https://developercommunity.visualstudio.com/content/problem/22922
+# charset = utf-8
 
-[*.proj]
+[*.{proj,csproj,vcxproj,xproj,json,config,nuspec,xml}]
 indent_size = 2
 
-[*.csproj]
-indent_size = 2
 
-[*.vcxproj]
-indent_size = 2
 
-[*.xproj]
-indent_size = 2
+[*]
+# https://github.com/nunit/docs/wiki/Coding-Standards#namespace-class-structure-interface-enumeration-and-method-definitions
+csharp_indent_braces = false
+csharp_new_line_before_catch = true
+csharp_new_line_before_else = true
+csharp_new_line_before_finally = true
 
-[*.json]
-indent_size = 2
+# https://github.com/nunit/docs/wiki/Coding-Standards#spaces
+csharp_space_after_cast = false
+csharp_space_after_comma = true
+csharp_space_after_keywords_in_control_flow_statements = true
+csharp_space_around_binary_operators = before_and_after
+csharp_space_around_declaration_statements = do_not_ignore
+csharp_space_before_open_square_brackets = false
+csharp_space_before_semicolon_in_for_statement = false
+csharp_space_between_empty_square_brackets = false
+csharp_space_between_method_call_empty_parameter_list_parentheses = false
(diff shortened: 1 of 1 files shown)


jnm2:
Please do not merge yet. I have some improvements yet to push, and I'm waiting for some decisions on our naming standards.

rprouse:
Should we merge this minus the UTF-8 to get the code style improvements and then do a quick UTF-8 update when it is unblocked?

jnm2:
Yes, once I update this PR to match our code style decisions.

ChrisMaddock:
I remember discussion on these points becoming a little long-winded... 😉 

This would be good to get merged. @jnm2 - would it be possible to merge a 'current version' - and add the items still under discussion, later on?

jnm2:
@ChrisMaddock 
> long-winded

It feels like I have a knack for starting that kind of issue. 😄 😢 😭 

Anyhow, 😃 I will try get everything non-controversial through now.

jnm2:
I think everything in here is uncontroversial. Ready to be reviewed!

.editorconfig
```diff
@@ -10,28 +10,72 @@ root = true
...
+dotnet_naming_rule.namespaces_types_and_non_field_members.style = pascal_case
+
+dotnet_naming_symbols.public_fields.applicable_kinds = field
+dotnet_naming_symbols.public_fields.applicable_accessibilities = public
+dotnet_naming_rule.public_fields.severity = error
+dotnet_naming_rule.public_fields.symbols = public_fields
+dotnet_naming_rule.public_fields.style = pascal_case
+
+dotnet_naming_style.camel_case.capitalization = camel_case
+dotnet_naming_symbols.parameters_and_locals.applicable_kinds = parameter, local
```
mikkelbu: Can we add an empty line between `dotnet_naming_style.camel_case.capitalization...` and `dotnet_naming_symbols.parameters_and_locals` to make it more consistent with `pascal_case`. At first I did not notice the `dotnet_naming_style` here.

ChrisMaddock:
>It feels like I have a knack for starting that kind of issue. 😄 😢 😭

That's a good quality to have! :-D

This looks good to me. 🙂

jnm2:
@mikkelbu 

> I've not checked whether there is something missing in the rules that we state in the coding standards. But we can always extend the configuration later on.

I skipped over a bunch of rules where the .editorconfig setting is not granular enough and would go too far.
Hang on, I'm going to add back `csharp_style_var_when_type_is_apparent` in line with https://github.com/nunit/docs/wiki/Coding-Standards#use-of-the-var-keyword. I skipped it because `true:error` would result in a lot of errors in opening existing code, but `true:none` (affects code generation only) or `true:suggestion` would be a good fit now that I think about it.

.editorconfig
jnm2: Thanks for catching! I forgot to go back and make that consistent.

jnm2:
@mikkelbu Kasey Uhlenhuth also fleshed out the official docs to a great degree. It documents almost every setting that I found via the Roslyn source: https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-code-style-settings-reference

jnm2:
Updated, so new reviews necessary.

PR #2337, opened 2017-07-28
Include differences in output for Is.EquivalentTo

OmicronPersei:
Fixes issue #1151 .

Allows the returned `ConstraintResult` from `CollectionEquivalentConstraint` description message to show collection differences (up to 10) of either extra or missing items when attempting to compare the actual collection to the expected.

For example,
```
List<string> expectedCollection = new List<string>() { "one", "two" }; 
List<string> actualCollection = new List<string>() { "three", "one" }; 
 
ConstraintResult cr = new CollectionEquivalentConstraint(expectedCollection)
    .ApplyTo(actualCollection); 
```

would yield a test failure message of:

```
  Expected: equivalent to < "one", "two" >
  But was:  < "three", "one" >
  Missing (1): < "two" >
  Extra (1): < "three" >
```

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraint.cs
@@ -75,8 +75,9 @@ namespace NUnit.Framework.Constraints
         {
             //Store the tally so the collection is only iterated over once.
             _lastPerformedTally = Tally(_expected);
+            _lastPerformedTally.TryRemove(actual);
 
-            return _lastPerformedTally.TryRemove(actual) && _lastPerformedTally.MissingItems.Count == 0;
+            return ((_lastPerformedTally.Result.ExtraItems.Count == 0) && (_lastPerformedTally.Result.MissingItems.Count == 0));
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraint.cs
@@ -103,7 +103,7 @@ namespace NUnit.Framework.Constraints
             bool matchesResult = Matches(enumerable);
 
             return new CollectionEquivalentConstraintResult(
-                this, _lastPerformedTally, actual, matchesResult);
+                this, _lastPerformedTally.Result, actual, matchesResult);
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraint.cs
@@ -33,16 +33,12 @@ namespace NUnit.Framework.Constraints
     {
         private readonly IEnumerable _expected;
 
-        /// <summary>
-        /// The <see cref="CollectionTally"/> of the last performed collection equivalence
-        /// comparison.
-        /// </summary>
-        private CollectionTally _lastPerformedTally;
+        /// <summary>The result of the <see cref="CollectionTally"/> from the collections
+        /// under comparison.</summary>
+        private CollectionTally.CollectionTallyResult _tallyResult;
 
-        /// <summary>
-        /// Construct a CollectionEquivalentConstraint
-        /// </summary>
-        /// <param name="expected"></param>
+        /// <summary>Construct a CollectionEquivalentConstraint</summary>
+        /// <param name="expected">Expected collection.</param>
         public CollectionEquivalentConstraint(IEnumerable expected)
             : base(expected)
         {
@@ -73,11 +69,14 @@ namespace NUnit.Framework.Constraints
         /// <returns></returns>
         protected override bool Matches(IEnumerable actual)
         {
-            //Store the tally so the collection is only iterated over once.
-            _lastPerformedTally = Tally(_expected);
-            _lastPerformedTally.TryRemove(actual);
+            CollectionTally ct = Tally(_expected);
+            ct.TryRemove(actual);
 
-            return ((_lastPerformedTally.Result.ExtraItems.Count == 0) && (_lastPerformedTally.Result.MissingItems.Count == 0));
+            //Store the CollectionTallyResult so the comparison between the two collections
+            //is only performed once.
+            _tallyResult = ct.Result;
+
+            return ((_tallyResult.ExtraItems.Count == 0) && (_tallyResult.MissingItems.Count == 0));
         }
 
         /// <summary>
@@ -103,7 +102,7 @@ namespace NUnit.Framework.Constraints
             bool matchesResult = Matches(enumerable);
 
             return new CollectionEquivalentConstraintResult(
-                this, _lastPerformedTally.Result, actual, matchesResult);
+                this, _tallyResult, actual, matchesResult);
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraint.cs
@@ -34,6 +34,12 @@ namespace NUnit.Framework.Constraints
         private readonly IEnumerable _expected;
 
         /// <summary>
+        /// The <see cref="CollectionTally"/> of the last performed collection equivalence
+        /// comparison.
+        /// </summary>
+        private CollectionTally _lastPerformedTally;
+
+        /// <summary>
         /// Construct a CollectionEquivalentConstraint
         /// </summary>
         /// <param name="expected"></param>
@@ -67,13 +73,36 @@ namespace NUnit.Framework.Constraints
         /// <returns></returns>
         protected override bool Matches(IEnumerable actual)
         {
-            // This is just an optimization
-            if (_expected is ICollection && actual is ICollection)
-                if (((ICollection)actual).Count != ((ICollection)_expected).Count)
-                    return false;
+            //Store the tally so the collection is only iterated over once.
+            _lastPerformedTally = Tally(_expected);
 
-            CollectionTally tally = Tally(_expected);
-            return tally.TryRemove(actual) && tally.Count == 0;
+            return _lastPerformedTally.TryRemove(actual) && _lastPerformedTally.MissingItems.Count == 0;
+        }
+
+        /// <summary>
+        /// Test whether the collection is equivalent to the expected.
+        /// </summary>
+        /// <typeparam name="TActual">
+        /// Actual collection type.
+        /// </typeparam>
+        /// <param name="actual">
+        /// Actual collection to compare.
+        /// </param>
+        /// <returns>
+        /// A <see cref="CollectionEquivalentConstraintResult"/> indicating whether or not
+        /// the two collections are equivalent.
+        /// </returns>
+        public override ConstraintResult ApplyTo<TActual>(TActual actual)
+        {
+            // TODO: Use an error result if actual is not IEnumerable
+            IEnumerable enumerable = actual as IEnumerable;
+            if (enumerable == null)
+                throw new ArgumentException("The actual value must be an IEnumerable", "actual");
+
+            bool matchesResult = Matches(enumerable);
+
+            return new CollectionEquivalentConstraintResult(
+                this, _lastPerformedTally, actual, matchesResult);
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
@@ -89,22 +89,22 @@ namespace NUnit.Framework.Constraints
             //Write the expected/actual message first.
             base.WriteMessageTo(writer);
             
-            if (_tally.MissingItems.Count > 0)
+            if (_tally.Result.MissingItems.Count > 0)
             {
-                int missingItemsCount = _tally.MissingItems.Count;
+                int missingItemsCount = _tally.Result.MissingItems.Count;
 
                 string missingStr = $"Missing ({missingItemsCount}): ";
-                missingStr += MsgUtils.FormatCollection(_tally.MissingItems, 0, MaxDifferingElemsToWrite);
+                missingStr += MsgUtils.FormatCollection(_tally.Result.MissingItems, 0, MaxDifferingElemsToWrite);
 
                 writer.WriteMessageLine(missingStr);
             }
 
-            if (_tally.ExtraItems.Count > 0)
+            if (_tally.Result.ExtraItems.Count > 0)
             {
-                int extraItemsCount = _tally.ExtraItems.Count;
+                int extraItemsCount = _tally.Result.ExtraItems.Count;
 
                 string extraStr = $"Extra ({extraItemsCount}): ";
-                extraStr += MsgUtils.FormatCollection(_tally.ExtraItems, 0, MaxDifferingElemsToWrite);
+                extraStr += MsgUtils.FormatCollection(_tally.Result.ExtraItems, 0, MaxDifferingElemsToWrite);
 
                 writer.WriteMessageLine(extraStr);
             }
--- a/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
@@ -24,87 +24,59 @@
 
 namespace NUnit.Framework.Constraints
 {
-    /// <summary>
-    /// Provides a <see cref="ConstraintResult"/> for the <see cref="CollectionEquivalentConstraint"/>.
-    /// </summary>
-    /// <remarks>
-    /// Assumes that a <see cref="CollectionTally"/> has been created and used to compare the expected
-    /// equivalent collections.
-    /// </remarks>
+    /// <summary>Provides a <see cref="ConstraintResult"/> for the <see cref="CollectionEquivalentConstraint"/>.</summary>
     public class CollectionEquivalentConstraintResult : ConstraintResult
     {
-        /// <summary>
-        /// Preprocessed <see cref="CollectionTally"/> of the collections to compare for equivalence.
-        /// </summary>
-        private CollectionTally _tally;
+        /// <summary>Result of a <see cref="CollectionTally"/> of the collections to compare for equivalence.</summary>
+        private CollectionTally.CollectionTallyResult _tallyResult;
 
-        /// <summary>
-        /// Maximum amount of elements to write to the <see cref="MessageWriter"/> if there are
-        /// extra/missing elements from the collection.
-        /// </summary>
+        /// <summary>Maximum amount of elements to write to the <see cref="MessageWriter"/> if there are
+        /// extra/missing elements from the collection.</summary>
         private const int MaxDifferingElemsToWrite = 10;
         
-        /// <summary>
-        /// Construct a <see cref="CollectionEquivalentConstraintResult"/> with a <see cref="CollectionTally"/>
-        /// already applied to the actual collection.
-        /// </summary>
-        /// <param name="Constraint">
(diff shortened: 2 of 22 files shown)


CharliePoole:
I started reviewing this but got called to family activities for the day - I'll come back to it tonight (GMT). I do think we need some changes.

OmicronPersei:
Any ideas on why CI is failing?  The looks look less than obvious.

src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintResultTests.cs
```diff
@@ -0,0 +1,125 @@
...
+            List<object> actualList = new List<object>() { "three", "one", };
+
+            ConstraintResult cr = _constraint.ApplyTo(actualList);
+            cr.WriteMessageTo(_writer);
+
+            string expectedMsg = 
+                "  Expected: equivalent to < \"one\", \"two\" >" + Environment.NewLine +
+                "  But was:  < \"three\", \"one\" >" + Environment.NewLine +
+                "  Missing (1): < \"two\" >" + Environment.NewLine +
+                "  Extra (1): < \"three\" >" + Environment.NewLine;
```
CharliePoole: Do we want such a complicated message when the expected and actual are this short? One option would be to only display missing elements if expected is truncated and only display extras if actual is truncated.

OmicronPersei:
> It appears that CollectionTally is asking to be refactored. Too much of its internal workings are exposed to the caller. It really only needs to indicate if there are either extra or missing items in a comparison.

That is all it does.  Looking at its public methods / properties, it exposes missing items, extra items, and whether or not individual items or collections can be tallied.  What in particular is extra or is showing too much to the caller?

src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintResultTests.cs
OmicronPersei: I think we do.  The alternative is to only show extra/missing elements when either the expected or actual collections themselves are too large to display, but even waiting until then can make it hard for the end user to see the differences in the collection.  For example:

```
List<string> a = new List<string>() { "0", "1", "1", "1", "0", "1", "0", "1", "0", "1" };
List<string> b = new List<string>() { "1", "0", "1", "1", "1", "0", "1", "1", "1", "0" };
```

Although the `List`s are still at 10 in length, it's not immediately obvious what's different between the collections (you can spend the time analyzing it, but that would defeat the entire purpose of the PR, I think).

OmicronPersei: Although perhaps there's a less verbose way of displaying both "Extra" and "Missing" separately?

ChrisMaddock: Personally, I like the exception message. 👍

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraint.cs
```diff
@@ -67,13 +73,36 @@ public override string Description
...
+            //Store the tally so the collection is only iterated over once.
+            _lastPerformedTally = Tally(_expected);
+
+            return _lastPerformedTally.TryRemove(actual) && _lastPerformedTally.MissingItems.Count == 0;
+        }
+
+        /// <summary>
+        /// Test whether the collection is equivalent to the expected.
+        /// </summary>
+        /// <typeparam name="TActual">
```
ChrisMaddock: Picky - can we put all these `<param>`/`<returns>` etc. docs into single lines - to match the rest of the code-base?

src/NUnitFramework/tests/Internal/CollectionTallyTests.cs
```diff
@@ -0,0 +1,81 @@
+﻿using NUnit.Framework.Internal;
```
ChrisMaddock: Please can you add the copyright notice to here and CollectionEquivalentConstraintResultTests? 🙂

src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -32,28 +32,47 @@ namespace NUnit.Framework.Constraints
...
-        }
-
-        /// <summary>
-        /// The number of objects remaining in the tally
-        /// </summary>
-        public int Count
-        {
-            get { return list.Count; }
+                _items.Add(o);
         }
```
ChrisMaddock: Good choice - removing `Count`. 😄

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
```diff
@@ -0,0 +1,113 @@
...
+// ***********************************************************************
+
+
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// Provides a <see cref="ConstraintResult"/> for the <see cref="CollectionEquivalentConstraint"/>.
+    /// </summary>
+    /// <remarks>
+    /// Assumes that a <see cref="CollectionTally"/> has been created and used to compare the expected
```
ChrisMaddock: On Charlie's comment about CollectionTally needing refactoring - comments like this make me agree. (The word 'assumes' set of warning sirens! 😆)

Do you think it would be overkill to introduce a `CollectionTallyResult` class? (possibly named something better!) That would give type-safety here, and avoid the potential issue of 'unfinished' collection tallys being passed about.

Open to debate on this one! 😄

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
```diff
@@ -0,0 +1,113 @@
...
+            CollectionEquivalentConstraint Constraint,
+            CollectionTally Tally,
+            object actual,
+            bool isSuccess)
+            : base(Constraint, actual, isSuccess)
+        {
+            if (Tally == null)
+            {
+                throw new System.ArgumentNullException("Tally was null.");
+            }
```
ChrisMaddock: We have a `Guard.ArgumentNotNull()` method we use for cases like this - just a little tidier, and gives consistent exception messages. 😄

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// <param name="isSuccess">
+        /// Whether or not the <see cref="CollectionEquivalentConstraint"/> succeeded.
+        /// </param>
+        /// <remarks>
+        /// The already processed <see cref="CollectionTally"/> is provided to reduce
+        /// unnecessary iterations over the collections to determine equivalence.
+        /// </remarks>
+        public CollectionEquivalentConstraintResult(
+            CollectionEquivalentConstraint Constraint,
+            CollectionTally Tally,
```
ChrisMaddock: lowercase parameter names is NUnit convention, if that's ok!

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
```diff
@@ -0,0 +1,113 @@
...
+            {
+                int missingItemsCount = _tally.MissingItems.Count;
+
+                string missingStr = $"Missing ({missingItemsCount}): ";
+                missingStr += MsgUtils.FormatCollection(_tally.MissingItems, 0, MaxDifferingElemsToWrite);
+
+                writer.WriteMessageLine(missingStr);
+            }
+
+            if (_tally.ExtraItems.Count > 0)
```
ChrisMaddock: Great that you've got this condition in! Can I suggest adding a couple of tests where there are only missing/only extra items, to enforce the empty label never being output? I don't think there currently are any - sorry if I missed them!

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+
+            _tally = Tally;
+        }
+
+        /// <summary>
+        /// Write the custom failure message for this object's 
+        /// <see cref="CollectionEquivalentConstraint"/>.
+        /// </summary>
+        /// <param name="writer"></param>
```
ChrisMaddock: Empty doc tag - can you remove or fill-in, please.

ChrisMaddock: (The code base is currently full of them, yes. But we're working towards improving. :-p )

src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintResultTests.cs
OmicronPersei: Okay.  One yes, one no.  Anyone else wanna way in?  Maybe @jnm2 ?

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
OmicronPersei: No, I agree actually.  I think my edits (which are coming) will address this.

OmicronPersei: I'm okay with `CollectionTallyResult` unless.. it's too boring sounding, lol.  Seems the most intuitive name in this situation.

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
OmicronPersei: So if the actual collection is empty, we don't want to output 

> But was:  <empty>

?  I believe one of my tests ensures that line is in the output message, so that would be a change.

If we do this, that would mean that the end user isn't immediately aware that the expected/actual provided collection was empty.  Would that make sense?

The output would look like
> Expected: < empty >
> Extra(1): < "one" >

or

> Expected: < "one" >
> Missing(1): < "one" > 

I think these two cases would make sense.  Thoughts?  Just making sure we're on the same page.

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
```diff
@@ -46,16 +46,13 @@ public class CollectionEquivalentConstraintResult : ConstraintResult
...
-                throw new System.ArgumentNullException("Tally was null.");
-            }
+            Guard.ArgumentNotNull(tallyResult, "tallyResult");
 
             _tallyResult = tallyResult;
         }
 
         /// <summary>Write the custom failure message for this object's <see cref="CollectionEquivalentConstraint"/>.</summary>
-        /// <param name="writer"></param>
+        /// <param name="writer">The object to write the failure message to.</param>
```
CharliePoole: "MessageWriter instance"

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraint.cs
```diff
@@ -34,6 +34,12 @@ public class CollectionEquivalentConstraint : CollectionItemsEqualConstraint
         private readonly IEnumerable _expected;
 
         /// <summary>
+        /// The <see cref="CollectionTally"/> of the last performed collection equivalence
+        /// comparison.
+        /// </summary>
+        private CollectionTally _lastPerformedTally;
```
CharliePoole: This hasn't changed. I still don't like it. Are we performing more than one tally?

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
```diff
@@ -0,0 +1,82 @@
...
+// ***********************************************************************
+
+
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>Provides a <see cref="ConstraintResult"/> for the <see cref="CollectionEquivalentConstraint"/>.</summary>
+    public class CollectionEquivalentConstraintResult : ConstraintResult
+    {
+        /// <summary>Result of a <see cref="CollectionTally"/> of the collections to compare for equivalence.</summary>
+        private CollectionTally.CollectionTallyResult _tallyResult;
```
CharliePoole: Ideally, I'd prefer that the CollectionTally class take on all the responsibilities of CollectionTallyResult, rather than creating yet another class here. I realize it is used in other classes that do not need the extra responsibilities, but they would simply not use them. Eventually, I think we will want to change how all the constraints that use CollectionTally work.

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
ChrisMaddock: @OmicronPersei - good job checking. We're not on the same page! 😄 

I agree what you currently have is good. The lines I **wouldn't** want to see are:

> Missing(0): < empty >

or

> Extra(0): < empty >

From looking at the code, this is what you've already done? I agree `Expected: <empty>` is the right thing to show.

My suggestion was that, I don't believe there are any tests that `Extra(0): < empty >` is _not_ shown when there are no extra items? I think all your tests have both Extra AND missing items? I may have just missed the test however, in which case, apologies!

OmicronPersei: Had to go back and check, but yes I do have those cases under test: `TestOneMissingValue` and `TestOneExtraValue` within the CollectionEquivalentConstraintResultTests.cs file.

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
OmicronPersei: Not sure why I wrote object instead of the actual type I'm using to write the msg to... sorry about that!

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
OmicronPersei: You're right, the code base does not make use of `CollectionTally` to check multiple collections, however the original source code / documentation did make reference to it potentially happening, so I guess I just kept it.  Specifically, the original < summary > tag of `CollectionTally` was:
>CollectionTally counts (tallies) the number of occurrences of each object in one or more enumerations.

Since that does not appear to be the case, I would agree it would make sense to no longer mention the usage of `CollectionTally` for comparing multiple enumerations.

OmicronPersei: >Ideally, I'd prefer that the CollectionTally class take on all the responsibilities of CollectionTallyResult, rather than creating yet another class here.

Can you expound upon this, not sure I understand?

OmicronPersei: Additionally, the reason I kept a reference to `_lastPerformedTally` was because of how the overrides for `CollectionItemsEqualConstraint` works.  The `Matches` method is called within `ApplyTo` method, and I wanted to avoid performing the tally between the two collections for efficiency reasons.

I've since refactored out a `CollectionTally.CollectionTallyResult` object to hold onto the result of a `CollectionTally`'s processing, so the variable type would need to change accordingly.

CharliePoole:
The reference to multiple collections had to do with using a single tally to create combined info on several collections.  So there would still only be one tally object. I don't know if we ever actually used it that way but we definitely don't now.

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
CharliePoole: What I meant was to just give CollectionTally itself Missing and Extra members and use it directly. But that may require reviewing all the different uses of CollectionTally in other constraints, so we probably don't want to do it in this issue.

CharliePoole:
Travis failures unrelated to these changes. I restarted the jobs.

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
ChrisMaddock: Great, you're right! Sorry I missed them!

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
OmicronPersei: >What I meant was to just give CollectionTally itself Missing and Extra members and use it directly.

That's actually exactly what I originally did.

OmicronPersei: I'm unsure what to do.  My original situation was viewed as unacceptable, but now it's being suggested when I offer an alternative.  Other opinions would be helpful, @jnm2 , @ChrisMaddock , @rprouse , I'm kind of at a loss for what to do at this point.  ☹️

ChrisMaddock: Personally, I'm happy with the solution in place here, and prefer it to the original version - as it gives extra type safety in terms a distinct 'result' object. @CharliePoole - would you mind clarifying the next steps you think @OmicronPersei should take?

CharliePoole:
@ChrisMaddock Not sure what there is to clarify. 

@OmnicronPersei It's pretty normal that we ask for changes.

Recapping my comments...
1. I asked you to change the name of one field.
2. I suggested that we come back to this in a separate issue / PR to got through all the constraints that use CollectionTally. That's in no way keeping this from being approved.

OmicronPersei:
Can't merge, it says there's still a change requested from @CharliePoole  ?

CharliePoole:
See my last comment - I was only waiting on _lastPerformedTally. Is that changed? You sometimes have to post a comment saying you made the changes to get people to re-review your work. We don't all necessarily see every time you push.

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraint.cs
OmicronPersei: Ok, it should now be:
```C#
/// <summary>The result of the <see cref="CollectionTally"/> from the collections
/// under comparison.</summary>
private CollectionTally.CollectionTallyResult _tallyResult;
```
Surprised it's not shown here.

CharliePoole:
@OmicronPersei Didn't realize you were a team member. Congratulations. The rule is that we need two team members (committers) to merge. In your case, you count as one of them, so it takes one other of us to approve the merge. Not sure if this got explained to you when you joined.

Anyway, it looks like you made the requested change so I'm approving. I don't know if @ChrisMaddock is waiting for changes, so I didn't merge it yet.

ChrisMaddock:
Great - thanks for taking this one on @OmicronPersei. 😄

src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintResultTests.cs
jnm2: Sorry I didn't weigh in. I think this is good. We can always make adjustments in the future too.

jnm2:
[Edited to fix @ mention]

src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -62,35 +83,26 @@ private bool ItemsEqual(object expected, object actual)
...
 
-        /// <summary>
-        /// Try to remove a set of objects from the tally
-        /// </summary>
-        /// <param name="c">The objects to remove</param>
-        /// <returns>True if successful, false if any object was not found</returns>
-        public bool TryRemove(IEnumerable c)
+        /// <summary>Try to remove a set of objects from the tally.</summary>
+        /// <param name="c">The objects to remove.</param>
+        public void TryRemove(IEnumerable c)
```
jnm2: Note for future PR: we gotta be careful depending where this class is used because `TryRemove(stringExpression)` will match `TryRemove(IEnumerable)`, not `TryRemove(object)`.

Issue #2219, opened 2017-06-04
Remove Obsolete Attributes


rprouse:
We obsoleted several attributes like `FixtureSetup` with the 3.0 release. It is probably about time we removed them.

Other candidates for removal are some of the fluent syntax for constraints where we came up with better naming.

stevenaw:
I can take this on.
Can I please be assigned to it?

CharliePoole:
Hi @stevenaw  Sure thing. I sent you an invitation as a contributor. After you accept I'll be able to make assignments to you.

stevenaw:
Thanks @CharliePoole , I've accepted the invite 🙂

Lette:
Just a thought:

If we're following SemVer, we have to consider that removing these attributes might be a breaking change, which will require a bump of the major version. So, this is probably something that shouldn't be done before the 4.0 release.

CharliePoole:
That's a point. Not sure how much into semver everyone is at this point though. @rprouse ?

NN---:
Please do it in 4.0.
It will break my code and many other projects too.

CharliePoole:
@rprouse How do you want to handle this? Should we have a 4.0 Milestone?

rprouse:
If we follow SemVer to the letter, then we probably should hold off on this to a 4.0 release. I think you could argue though that since most of the attributes were obsoleted in 3.0 and there have been many releases to fix the warnings it isn't a breaking change.

AFAIK, the only obsolete feature that isn't a simple search and replace update is the `AssertHelper`. `AssertHelper` has only been obsolete for one release, so removing  it may be pre-mature. Updating everything else should only take a few minutes, even for large solutions. @NN or @Lette is there anything other than `AssertHelper` that you see as a major or difficult change?

Personally, I am surprised that people haven't already updated their code, but maybe I just hate warnings more than other people 😄 I am also surprised that people continue to use `AssertHelper` since it is so limited compared to the other asserts.

stevenaw:
I have some of these changes sitting in a branch. Should I go back to it,
wrap it up, and open a PR?
Or should I wait until we know when we'll be releasing it?

NN---:
I have some obsolete attributes. Not really much work to find and replace.

The main difficulty was ExpectedException attribute removal in 3.0 which I can say was a very good idea.
Too bad it existed from the beginning..

rprouse:
[...]

Issue #2332, opened 2017-07-28
Update build script to use NUnitConsoleRunner v3.7.0


OmicronPersei:
As the title says.

jnm2:
Sounds good- I'm getting consistent SocketExceptions and 3.7 fixes it for me and I saw it did the same for you.

Are you interested in submitting the PR for this?

OmicronPersei:
Yes

jnm2:
Thanks!

ChrisMaddock:
Fixed by https://github.com/nunit/nunit/pull/2340#pullrequestreview-53435939

PR #2338, opened 2017-07-29
Sample PR - Issue2219 remove obsolete attributes

stevenaw:
A sample PR for Issue #2219 to visualize the impact of removing obsolete attributes.


ChrisMaddock:
I'm in favour of merging this into 3.8. I appreciate it doesn't follow SemVer - but we don't follow SemVer to strictly anyway. We should just come to peace with that, and call what we do NUnitVer instead. 😄 

They've been obsolete since 3.0, and people have had plenty of time to refactor. What I _would_ like to see is everything that was obsoleted in 3.0 removed in one go, to minimise user disruption. As far as I can see, the only other thing is the `Is.StringContaining` variants - but it would be good to double check that.

src/NUnitFramework/tests/Attributes/PairwiseTests.cs
```diff
@@ -38,13 +38,13 @@ public class LiveTest
             private PairCounter pairsTested = new PairCounter();
 
             [OneTimeSetUp]
-            public void TestFixtureSetUp()
+            public void OneTimeSetUp()
```
CharliePoole: Unneeded but OK. We often refer to the OneTimeSetUp of a fixture this way in the docs.

src/NUnitFramework/tests/Attributes/SetUpFixtureAttributeTests.cs
```diff
@@ -31,38 +31,22 @@ namespace NUnit.Framework.Attributes
 {
     public class SetUpFixtureAttributeTests
     {
-        [TestCase(typeof(Class1))]
-        [TestCase(typeof(Class2))]
-        [TestCase(typeof(Class3))]
-        [TestCase(typeof(Class4))]
+        [TestCase(typeof(TestSetupClass))]
+        [TestCase(typeof(TestSearDownClass))]
```
CharliePoole: Spelling.

CharliePoole:
Looks good to me as well if you fix the misspelled class name.

rprouse:
I am good with this too. As @CharliePoole says, the spelling needs fixing, but after that, I think we should merge. I like that this only changes things that are easy to update with a global search and replace. I would prefer to deal with changes that will require hand updating as a SemVer breaking change.

rprouse:
@stevenaw if you want to resolve conflicts and fix that one spelling error, I think we can merge this and take the rest of the obsolete code in steps.

src/NUnitFramework/tests/Attributes/SetUpFixtureAttributeTests.cs
stevenaw: Thanks, good catch

stevenaw:
Thanks @rprouse @CharliePoole @ChrisMaddock .
Spelling corrected and conflicts resolved.

Build completes locally, but running the tests locally has 2 failures due to `SocketException`s. I'm assuming it's just related to this https://github.com/nunit/nunit/issues/2332#issuecomment-318535845

ChrisMaddock:
Looks good - re-running CI. 🙂

rprouse:
CI passed, merging

PR #2351, opened 2017-08-06
Support for async lambda with Assert.Multiple

### LINKED ISSUE #2348: Can't use Assert.Multiple with async code
Passing an async lambda into Assert.Multiple compiles fine with no warnings but always passes instantaneously, even if the code inside the lambda should fail.

Sample test code:

```
namespace app
{
    [TestFixture]
    public class Tests
    {
        [Test]
        public void TestMultipleSync()
        {
            Assert.Multiple(() =>
            {
                Thread.Sleep(TimeSpan.FromSeconds(5));
                Assert.Fail();
            });
        }

        [Test]
        public void TestMultipleAsync()
        {
            Assert.Multiple(async () =>
            {
                await Task.Delay(TimeSpan.FromSeconds(5));
                Assert.Fail();
            });
        }
    }
}
```

I would intuitively expect both these tests to take 5 seconds to run before then failing. However, only TestMultipleSync fails; TestMultipleAsync passes successfully in less than a second. Output from console runner:

```
PS C:\dev\scratch\test> .\packages\NUnit.ConsoleRunner.3.7.0\tools\nunit3-console.exe .\app\bin\Debug\app.exe
NUnit Console Runner 3.7.0
Copyright (c) 2017 Charlie Poole, Rob Prouse

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.15063.0
  CLR Version: 4.0.30319.42000

Test Files
    .\app\bin\Debug\app.exe


Errors, Failures and Warnings

1) Failed : app.Tests.TestMultipleSync
at app.Tests.<>c.<TestMultipleSync>b__0_0() in C:\dev\scratch\test\app\Tests.cs:line 17
at NUnit.Framework.Assert.Multiple(TestDelegate testDelegate)
at app.Tests.TestMultipleSync() in C:\dev\scratch\test\app\Tests.cs:line 14

Run Settings
    DisposeRunners: True
    WorkDirectory: C:\dev\scratch\test
    ImageRequiresX86: True
    RunAsX86: True
    NumberOfTestWorkers: 8

Test Run Summary
  Overall result: Failed
  Test Count: 2, Passed: 1, Failed: 1, Warnings: 0, Inconclusive: 0, Skipped: 0
  Start time: 2017-08-04 10:23:02Z
    End time: 2017-08-04 10:23:08Z
    Duration: 5.504 seconds
```

I guess the lambda is being interpreted as an "async void" rather than an "async Task" and so its results end up ignored? There doesn't seem to be any obvious other way of doing an async multiple assert though - most of the other Assert methods that take a TestDelegate have an equivalent AsyncTestDelegate method (Catch/CatchAsync, DoesNotThrow/DoesNotThrowAsync), but there's no MultipleAsync method.

There's also nothing on the documentation page for Multiple Asserts (https://github.com/nunit/docs/wiki/Multiple-Asserts) which calls out that you can't use it with async code - on the contrary, it says the assert block "may contain any arbitrary code".

Using NUnit 3.7.1, console runner 3.7.0.

CharliePoole:
Fixes #2348

### DIFF
--- a/src/NUnitFramework/testdata/AssertMultipleData.cs
+++ b/src/NUnitFramework/testdata/AssertMultipleData.cs
@@ -1,7 +1,12 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.Text;
 using NUnit.Framework;
+
+#if ASYNC
+using System.Threading.Tasks;
+#if NET_4_0
+using Task = System.Threading.Tasks.TaskEx;
+#endif
+#endif
 
 namespace NUnit.TestData.AssertMultipleData
 {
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/testdata/AssertMultipleData.cs
```diff
@@ -260,6 +265,88 @@ public void AssumptionInBlock()
...
             });
         }
+
+#if ASYNC
+        [Test]
+        public void ThreeAssertsSucceed_Async()
+        {
+            Assert.Multiple(async () =>
+            {
+                await Task.Delay(1000);
```
rprouse: Does this delay (and the 500 below) have to be so long? Our tests are already getting fairly long to run...

Would something like `await Task.FromResult(false)` suffice?

CharliePoole: Could be shorter. I added delays in most cases as a way to use await, so that the lambda actually runs asynchronously. In a few cases, some delay was needed or the test would never have failed in the first place. I'll go through them all and see what I can do.

CharliePoole: I removed the async Assert.Multiple overload and reduced the times of all the delays to 100ms. All five async tests still were failing. Then I put back the async overload and they passed. I'll push and merge when CI passes.

Issue #2062, opened 2017-02-28
TestCaseSource attribute causes test to pass when source is not defined


estrizhok:
Steps to reproduce:
1. Create a new project,
2. Reference nunit.framework.dll version 3.6.1
3. Paste the following snippet as a class into the project:
4. Compile project and run the tests using nunit3-console.exe

```csharp
public class TestClass 
{
   [TestCaseSource("UndefinedSource")]
   public void Test(int param) { }
}
```

Observed result:
1. Test Count: 0, Passed: 0, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 0
2. TestResult.xml contains node <test-suite name="Test" result="Passed" /> with no children

Expected result:
1. NUnit should report absence of class member referenced in by the attribute as an error

IngoVals:
This also happens if source exists but is empty. I think at least the test should be *inconclusive*.

CharliePoole:
@IngoVals These are of course two different situations.

If the member referenced can't be found it's an error and the test is invalid.

If it exists but is empty, no data is provided so the test can't be run. It's a question of design whether this should pass or be inconclusive. It definitely can't be an error, since there are situations where people deliberately create sources that sometimes do not have data.

The design decision in NUnit up to now has been that __Tests__ with no data provided pass, while __Theories__ with no data are inconclusive. IOW, a Test is innocent till proven guilty but a Theory is a failure unless it can be proven a success. 😄

jnm2:
Re: pass or inconclusive, think of this situation. I have integration tests going against a database. It might be a database full of data, or it might be a newly created, empty database. I have tests such as `Report_does_not_have_invalid_filter(string reportName)` and a test case source generating a case for every report in the database.

For a brand new database, there are no reports set up yet, so no test cases. Does that mean that it is inconclusive whether there are any reports with an invalid filter? No; the answer is entirely conclusive that there are no reports with an invalid filter and that I should be seeing a green light.

If I wasn't using TestCaseSource, I'd be using something like `Assert.That(reports.All(HaveValidFilter));`. If there are no reports, it succeeds. You'd intuitively expect the refactoring to a `TestCaseSource` to preserve this semantic.

CharliePoole:
@jnm2 That's a convincing use case. Of course, folks who want it to work differently can probably come up with counter-cases. 😄 But basically, I'd have to say this is one of those things decided long ago, which can't really be changed without breaking compatibility significantly.

jnm2:
[...]

PR #2352, opened 2017-08-08
Fixes TestCaseSource bug where no error is reported upon non existing source

OmicronPersei:
Fixes issue #2062 .

### DIFF
--- a/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
@@ -304,6 +304,15 @@ namespace NUnit.Framework.Attributes
         {
         }
 
+        [Test]
+        public void TestMethodIsNotRunnableWhenSourceDoesNotExist()
+        {
+            TestSuite suiteToTest = TestBuilder.MakeParameterizedMethodSuite(typeof(TestCaseSourceAttributeFixture), "MethodWithNonExistingSource");
+            
+            Assert.That(suiteToTest.Tests.Count == 1);
+            Assert.AreEqual(RunState.NotRunnable, suiteToTest.Tests[0].RunState);
+        }
+
         static object[] testCases =
         {
             new TestCaseData(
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -304,6 +304,15 @@ public void TestWithZeroTestSourceCasesShouldPassWithoutRequiringArguments(int r
         {
         }
 
+        [Test]
+        public void TestMethodIsNotRunnableWhenSourceDoesNotExist()
+        {
+            TestSuite suiteToTest = TestBuilder.MakeParameterizedMethodSuite(typeof(TestCaseSourceAttributeFixture), "MethodWithNonExistingSource");
+            
+            Assert.That(suiteToTest.Tests.Count == 1);
+            Assert.AreEqual(RunState.NotRunnable, suiteToTest.Tests[0].RunState);
```
rprouse: Interesting, I just noticed that all of our other `Assert.AreEqual` in this file have the expected and actual parameters backwards but you got it right. 👍 

I've always felt that we do it backwards and that it should be `Assert.AreEqual(actual, expected)` so that it matches the order of `Assert.That(actual, Is.EqualTo(expected))` but too late to change that 😄

rprouse:
@OmicronPersei, you have been very active lately and we really appreciate your help, so I have added you to the @nunit/framework-team. This is to recognize your help and make it easier to work, but is not an obligation on your part.

You can now push branches to the repository which means that you don't have to work in a fork. Master is locked, so no worries about messing up. As a member, your PRs will only require one other team member to review going forward and you are welcome to review and merge other PRs. Our policy is for two team members to review non-team PRs and one team member to review other team members PRs. Of course, use your discretion 😉 

If you have any questions now or at any time, please email me at rob@prouse.org.

Welcome 😄

CharliePoole:
I think it was too late when NUnit was invented. I think it came from sUnit via jUnit.

OmicronPersei:
@rprouse Thank you!  Glad to be working on such a high visibility, well coded project!

PR #2356, opened 2017-08-09
Update mono versions to build

CharliePoole:
**Do Not Merge**

This is not yet working but I created the PR as a place to discuss what we do.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ matrix:
     - os: osx
       mono: latest
   allow_failures:
-    - os: osx
     - mono: latest
   fast_finish: true
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,18 +2,17 @@ language: csharp
 sudo: false
 mono:
   - 4.6.2 # in most common distros
-  - 4.2.1 # in Ubuntu LTS
-  - 5.0.1 # latest stable
+  - latest
 os:
   - linux
   - osx
 matrix:
   exclude:
     - os: osx
-      mono: 5.0.1
+      mono: latest
   allow_failures:
     - os: osx
-    - mono: 5.0.1
+    - mono: latest
   fast_finish: true
 script:
   - ./build.sh --target "Travis"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,19 @@
 language: csharp
 sudo: false
 mono:
-  - 4.8.1
-  - latest
+  - 4.6.2 # in most common distros
+  - 4.2.1 # in Ubuntu LTS
+  - 5.0.1 # latest stable
 os:
   - linux
   - osx
 matrix:
   exclude:
     - os: osx
-      mono: latest
+      mono: 5.0.1
   allow_failures:
     - os: osx
-    - mono: latest
+    - mono: 5.0.1
   fast_finish: true
 script:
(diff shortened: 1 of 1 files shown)


CharliePoole:
Result of first commit...

* Using Mono 4.2.1 brings in xbuild version 12. We fail to build. We could mess with the installed software, but I think 4.2.1 is simply not worth the effort.

* 4.6.2 seems to work well

* 5.0 is not a valid mono, so I used 5.0.1. This got us a failure with no log. I guess we need to stick with "latest"

I'm going for a second shot with 4.6.2 and latest

CharliePoole:
OK, the build seems to be plagued with unrelated intermittent failures due to timing issues. However, the 4.6.2 build works well. The latest build, which we have been running for a long time with allowed failures, is failing in every build. I'll check that next but at least it seems like the choice of mono versions is now working correctly.

CharliePoole:
I went back two months. The travis build using mono latest has been failing in the same way every time at least that far back. This convinces me that we shouldn't allow failures for it.

What do you think @rprouse @ChrisMaddock ?

ChrisMaddock:
Do you suggest we just drop it totally? I'd be ok with that.

Should we take OSX out of allowed failures, while we're here? They went through a rough patch last year or so, when there seemed to be a shortage of agents, but have been much more stable recently.

CharliePoole:
No, I'm thinking it should be required. If it had been, we would have fixed it months ago.

CharliePoole:
Depending on the importance of the platform to us, we might require OSX to pass too.

rprouse:
I am good with these versions. I also think we should require OSX to pass as it has become a very common development platform. Should we merge this as-is?

CharliePoole:
I should change the OSX build first. What about mono latest?

CharliePoole:
My inclination is to leave the mono latest build as an Allowed Failure until the problems with it are solved. I'll create an issue to do that.

CharliePoole:
I figured out the problem with latest, so I think I'll just fix it on this PR if I can find a quick enough fix.

A while back, @ChrisMaddock created a method `ExceptionHelper.BuildFriendlyMessage`, which built a message that leaves out the name of the exception. It has the unstated assumption - or at least it's used with the unstated assumption - that an Exception always contains a non-blank, non-null Message. Turns out not to be the case WRT Mono 5.0. A `FileNotFoundException` and apparently others, may not have a message. Instead, it just has a field that gives the name of the file that wasn't found.

We use the blank message to set the test assembly as non-runnable using `Test.MakeInvalid` but that method has a Guard clause that checs for a null or blank message!

It's an interesting bug because it points out something that has changed with how we work. It used to be that we didn't trust the platforms we were running on. Silverlight or NetCF might turn around and do just about anything, not to mention Mono. Now we have fewer platforms and I guess we tend to trust them more. In this case, it bit us.

CharliePoole:
Well, it looks like more than a few lines of code, so I'll create a new issue and merge this.

ChrisMaddock:
Great find Charlie!

That's interesting - as I create BuildFriendlyException message specifically to keep to existing functionality - vs what would have been the way I'd have written it green field. I wonder what I missed?!

CharliePoole:
Well, I guess you didn't miss anything. It's just that things changed.

PR #2359, opened 2017-08-12
Fix Mono 5.0 build

### LINKED ISSUE #2358: Get framework to build under Mono 5.0
Tests have been failing for some months now under Mono 5.0. We need to get this working correctly and then remove mono latest from the Allowed Failures.

CharliePoole:
Fixes #2358 

I compensated for the blank message field we are seeing in `FileNotFoundException` by creating a message. This is probably a bug in 5.0 and 5.2 beta, but we need to deal with it.

Once fixed, I made the linux mono latest build required, so we will check into problems that arise. I also rewrote .travis.yml to use matrix include rather than exclude. It seems a lot clearer that way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,6 @@ matrix:
     - os: linux
       mono: latest
     - os: osx
-      mono: 4.6.2
+      mono: latest
 script:
   - ./build.sh --target "Travis"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,12 @@
 language: csharp
 sudo: false
-mono:
-  - 4.6.2 # in most common distros
-  - latest
-os:
-  - linux
-  - osx
 matrix:
-  exclude:
-    - os: osx
+  include:
+    - os: linux
+      mono: 4.6.2
+    - os: linux
       mono: latest
-  allow_failures:
-    - mono: latest
-  fast_finish: true
+    - os: osx
+      mono: 4.6.2
 script:
   - ./build.sh --target "Travis"
--- a/packages/repositories.config
+++ b/packages/repositories.config
@@ -1,5 +1,10 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <repositories>
+  <repository path="../src/NUnitFramework/framework/packages.config" />
+  <repository path="../src/NUnitFramework/nunitlite.tests/packages.config" />
+  <repository path="../src/NUnitFramework/nunitlite/packages.config" />
+  <repository path="../src/NUnitFramework/testdata/packages.config" />
+  <repository path="../src/NUnitFramework/tests/packages.config" />
   <repository path="..\src\NUnitFramework\testdata\packages.config" />
   <repository path="..\src\NUnitFramework\tests\packages.config" />
 </repositories>
--- a/src/NUnitFramework/nunitlite-runner/nunitlite-runner-4.5.csproj
+++ b/src/NUnitFramework/nunitlite-runner/nunitlite-runner-4.5.csproj
@@ -25,7 +25,8 @@
     <WarningLevel>4</WarningLevel>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <Prefer32Bit>false</Prefer32Bit>
-    <Externalconsole>true</Externalconsole>
+    <Commandlineparameters>nunit.framework.tests.dll</Commandlineparameters>
+    <ConsolePause>false</ConsolePause>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <PlatformTarget>AnyCPU</PlatformTarget>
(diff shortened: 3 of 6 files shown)


.travis.yml
```diff
@@ -1,17 +1,12 @@
...
+  include:
+    - os: linux
+      mono: 4.6.2
+    - os: linux
       mono: latest
-  allow_failures:
-    - mono: latest
-  fast_finish: true
+    - os: osx
+      mono: 4.6.2
```
ChrisMaddock: I know this isn't a change in this PR - but do you know why we test osx on an old version? I thought OSX mono isn't bundled, so it makes more sense to test against latest stable?

.travis.yml
```diff
@@ -1,17 +1,12 @@
...
-  exclude:
-    - os: osx
+  include:
+    - os: linux
+      mono: 4.6.2
+    - os: linux
       mono: latest
-  allow_failures:
-    - mono: latest
-  fast_finish: true
```
ChrisMaddock: Was removing this intentional?

packages/repositories.config
```diff
@@ -1,5 +1,10 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <repositories>
+  <repository path="../src/NUnitFramework/framework/packages.config" />
+  <repository path="../src/NUnitFramework/nunitlite.tests/packages.config" />
+  <repository path="../src/NUnitFramework/nunitlite/packages.config" />
+  <repository path="../src/NUnitFramework/testdata/packages.config" />
+  <repository path="../src/NUnitFramework/tests/packages.config" />
```
ChrisMaddock: Was this required for a build? I thought `repositories.config` was a historical file, which was no longer used. We removed it in the console

src/NUnitFramework/nunitlite-runner/nunitlite-runner-4.5.csproj
```diff
@@ -25,7 +25,8 @@
     <WarningLevel>4</WarningLevel>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <Prefer32Bit>false</Prefer32Bit>
-    <Externalconsole>true</Externalconsole>
+    <Commandlineparameters>nunit.framework.tests.dll</Commandlineparameters>
+    <ConsolePause>false</ConsolePause>
```
ChrisMaddock: Intentional changes? To just the 4.5 csproj?

.travis.yml
CharliePoole: Yes, per our discussion of builds, I thought we want to require every ci build to pass.

CharliePoole: I couldn't do this on the earlier PR because latest was failing.

.travis.yml
CharliePoole: My assumption is that some people are early adopters and others are not. No particular reason to pick 4.6.2 for OSX, however. I just used it because it's common with Linux. But I'm actually in the  dark in terms of what OSX users may actually want. ☹️

packages/repositories.config
CharliePoole: I was working on Linux and MonoDevelop keeps putting it back when I remove it.

src/NUnitFramework/nunitlite-runner/nunitlite-runner-4.5.csproj
CharliePoole: MonoDevelop stores the Startup program in the project file, while VS doesn't. It does no harm.

.travis.yml
ChrisMaddock: Sorry - I meant the `fast_finish` line - not the `allow_failures`

.travis.yml
ChrisMaddock: I have no OSX access either. 😞 I had a Google though - and `brew install mono` currently gets 5.0.1.1 - it looks like it's kept fairly up to date. While we're here - can I suggest just test on OSX latest, rathere than just 4.6.2?

packages/repositories.config
ChrisMaddock: Ahh, so it comes from MonoDevelop! Does MonoDevelop auto-regenerate it? Could we .gitignore it instead?

.travis.yml
ChrisMaddock: (Although actually, given all our build is wrapped up in Cake - I guess fast_finish is a no-op anyway.

CharliePoole:
Fast_finish only applies if you allow failures.

CharliePoole:
All our other builds use both latest and 4.6.2 for osx. I guess I didn't add it here simply because it wasn't what the issue is about.

Are you saying we should drop 4.6.2?  I can do it all in this issue to save time.

ChrisMaddock:
We only test unix 4.6.2 because that's what's bundled. On OSX, I don't believe anything is bundled - and anyone using the package manager would default get 5.x. So yes, I think we can drop it. 🙂

CharliePoole:
I'll switch it here and we can see if it builds. I'll drop it from the others as well, both pending and already merged.

PR #2378, opened 2017-08-22
Fix package for supported iOS/Android/UWP platforms

### LINKED ISSUE #2237: System.Runtime.Loader not available for iOS/Android
I'm not 100% clear on the latest NuGet behaviour - so would appreciate a second opinion on the below. This is what I think is going on in https://github.com/nunit/nunit.xamarin/issues/87.

Out .NET Standard 1.6 build relies on System.Runtime.Loader 4.3.0. If you unzip that NuGet package, the structure looks like this:

```
lib -> netstandard1.5    -> System.Runtime.Loader.dll
          net462         -> _._
          MonoAndroid10  -> _._
          MonoTouch10    -> _._
```

When this package is installed into an Android project (and I assume .NET Framework 4.6.2, or iOS also) - NuGet resolves to look in the `MonoAndroid10` directory, and finds the `_._` (and fails). As I understand (although I can't find anything official on the matter) - this is used as a placeholder, when an API isn't supported on a certain platform yet. See below:

https://github.com/dotnet/corefx/issues/8965#issuecomment-287138692
https://github.com/dotnet/corefx/issues/11023#issuecomment-244131549

If I'm understanding the above correctly, then our .NET Standard 1.6 NuGet package won't actually work for iOS, Android, or .NET 4.6.2. I'd assume we're only just discovering this as .NET Core is unaffected, which I imagine is the main use of the .NET Standard builds.

As a potential solution, perhaps we could package the NET Standard 1.3 build of NUnit (which _doesn't_ rely on System.Runtime.Loader) under a `monoandroid` directory in our NuGet package. (And of course additionally for the remaining affected frameworks.) Project.json unfortunately seems to ignore such subdirectories (unlike packages.config, which recognises them...) - but I wonder if that's the same issue we saw in https://github.com/nunit/nunit/issues/2165, which would be fixed by adding an empty dependency group for MonoAndroid etc. Not sure.

It's only the latest Android framework which pulls in .NET Standard I believe, which I imagine is why we didn't see this issue when Rob first created the .NET Standard 1.6 build. Unfortunately, unless someone knows a way to force project.json to look in a particular subdirectory for it's dependency, I think this could potentially be critical for both NUnit and nunit.xamarin.

ChrisMaddock:
Fixes #2237 

Unfortunately, I think we've been a bit liberal regarding adding new functionality to the .NET Standard builds, and assuming it will be available on all platforms. (Until I started looking in to this issue, I thought that was the POINT of .NET Standard - turns out it's not quite there yet.)

This PR includes a various changes to allow the package to be used on latest Android/iOS/UWP. I've upgraded and tested both the nunit.xamarin Android and UWP projects as a test. iOS I'm not set up to test, however NUnit 3.7.1 wasn't restoring in the same manner as the Android platform, so I assume the same fix applies. (NuGet now restores the package successfully, which it didn't previously.)

The first part of the issue, is that System.Runtime.Loader is available for .NET Standard 1.6, but not available on Android/iOS. To combat this, I altered the NuGet package to use the .NET Standard 1.3 build for these platforms, which doesn't have the System.Runtime.Loader dependency.

The next issue was that UWP doesn't support the System.Threading.Thread API, which was a dependency of NUnit's .NET Standard 1.3 build. Rather than create an entirely new build solely for UWP - I just removed the System.Threading.Thread dependency from .NET Standard 1.3. This is technically a breaking change - however given the build already wasn't working on the main platform we expect people to use the .NET Standard 1.3 build on - I don't think that's too much of a concern.

It would be good if we could test this stuff in CI. I had a quick experiment with the project.json, but couldn't get the build to fail - I don't have much more time to look in to that before the 3.8 release, so wanted to push this as is.

### DIFF
--- a/nuget/framework/nunit.nuspec
+++ b/nuget/framework/nunit.nuspec
@@ -28,9 +28,17 @@ Supported platforms:
       <group targetFramework="net35" />
       <group targetFramework="net40" />
       <group targetFramework="net45" />
+      <group targetFramework="Xamarin.iOS">
+        <dependency id="NETStandard.Library" version="1.6.0" />
+      </group>
+      <group targetFramework="MonoAndroid">
+        <dependency id="NETStandard.Library" version="1.6.0" />
+      </group>
+      <group targetFramework="uap">
+        <dependency id="NETStandard.Library" version="1.6.0" />
+      </group>
       <group targetFramework="netstandard1.3">
         <dependency id="NETStandard.Library" version="1.6.0" />
-        <dependency id="System.Threading.Thread" version="4.3.0" />
       </group>
       <group targetFramework="netstandard1.6">
         <dependency id="NETStandard.Library" version="1.6.0" />
@@ -52,6 +60,12 @@ Supported platforms:
     <file src="bin/net-4.0/nunit.framework.xml" target="lib\net40" />
     <file src="bin/net-4.5/nunit.framework.dll" target="lib\net45" />
     <file src="bin/net-4.5/nunit.framework.xml" target="lib\net45" />
+    <file src="bin/netstandard13/nunit.framework.dll" target="lib\xamarin.ios" />
+    <file src="bin/netstandard13/nunit.framework.xml" target="lib\xamarin.ios" />
+    <file src="bin/netstandard13/nunit.framework.dll" target="lib\monoandroid" />
+    <file src="bin/netstandard13/nunit.framework.xml" target="lib\monoandroid" />
+    <file src="bin/netstandard13/nunit.framework.dll" target="lib\uap" />
+    <file src="bin/netstandard13/nunit.framework.xml" target="lib\uap" />
     <file src="bin/netstandard13/nunit.framework.dll" target="lib\netstandard1.3" />
     <file src="bin/netstandard13/nunit.framework.xml" target="lib\netstandard1.3" />
     <file src="bin/netstandard16/nunit.framework.dll" target="lib\netstandard1.6" />
--- a/src/NUnitFramework/TestBuilder.cs
+++ b/src/NUnitFramework/TestBuilder.cs
@@ -202,7 +202,11 @@ namespace NUnit.TestUtilities
             // TODO: Replace with an event - but not while method is static
             while (work.State != WorkItemState.Complete)
             {
+#if NETSTANDARD1_3
+                System.Threading.Tasks.Task.Delay(1);
+#else
                 Thread.Sleep(1);
+#endif
             }
 
             return work.Result;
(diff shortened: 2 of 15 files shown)


nuget/framework/nunit.nuspec
```diff
@@ -52,6 +60,12 @@ Supported platforms:
     <file src="bin/net-4.0/nunit.framework.xml" target="lib\net40" />
     <file src="bin/net-4.5/nunit.framework.dll" target="lib\net45" />
     <file src="bin/net-4.5/nunit.framework.xml" target="lib\net45" />
+    <file src="bin/netstandard13/nunit.framework.dll" target="lib\xamarin.ios" />
+    <file src="bin/netstandard13/nunit.framework.xml" target="lib\xamarin.ios" />
+    <file src="bin/netstandard13/nunit.framework.dll" target="lib\monoandroid" />
+    <file src="bin/netstandard13/nunit.framework.xml" target="lib\monoandroid" />
+    <file src="bin/netstandard13/nunit.framework.dll" target="lib\uap" />
+    <file src="bin/netstandard13/nunit.framework.xml" target="lib\uap" />
```
ChrisMaddock: A separate UWP dir here isn't strictly necessary, as it would pull in the netstandard1.3 build anyway - but given the problems, I wanted to be explicit about the platforms we wanted to support.

rprouse:
This looks good to me. Thread related functionality was always intended to be part of the .NET Standard 1.6 build, not the 1.3 target and it was never available in the PCL which 1.3 replaces, so I think we are okay dropping it, especially since it isn't supported on the platforms that would likely pull in 1.3 anyway.

AppVeyor error is the typical .NET 3.5 message pump error. @jnm2 any chance you could look at that?

```
1) Failed : NUnit.Framework.Internal.ThreadUtilityTests.Kill
  Native message pump was not able to be interrupted to enable a managed thread abort.
  Expected: True
  But was:  False
```

The Travis build failure was another .NET 3.5 threading issue,

```
1) Failed : NUnit.Framework.Internal.Execution.TestWorkerTests.BusyExecuteIdleEventsCalledInSequence
  Expected: "BusyExecIdle" after 200 milliseconds delay
  But was:  <string.Empty>
  at NUnit.Framework.Internal.Execution.TestWorkerTests.BusyExecuteIdleEventsCalledInSequence () [0x0009f] in <a9dce29302ac4f1ba5dd3a44cd37ddc6>:0 
```

What is up with .NET 3.5? Both errors look unrelated to me.

rprouse:
New builds kicked off ☹️

ChrisMaddock:
Merging, based on Robs comments.

jnm2:
Wow Chris, cool! 🎉

PR #2381, opened 2017-08-24
Added a public, read-only Scope property to ParallelizableAttribute

### LINKED ISSUE #2196: Expose ParallelizableAttribute (and other attribute) constructor arguments as properties
NUnit 3.6.1
I have some code that uses Reflection to look for ParallelizableAttribute on my tests and checks the Scope value passed to the constructor, but currently there is no easy way to get this value. I have to manually look in the Properties bag for a string (that I found by inspecting the source, which could easily change) and cast the result. Please provide type-safe access to all constructor arguments for all attributes via public properties. That seems like best practice for such tests anyway, as opposed to relying on the global environment.

jadarnel27:
Fixes #2196 

I chose to replace the private member variable `_scope`, as it seemed redundant in the presence of the auto-property.  Hopefully that is alright.

One thing I was unsure about is the content of the documentation comment.  I took inspiration from the documentation of the ParallelScope enum itself, which seems appropriate.  But if the comment is misleading in some way, I'm definitely open to guidance on rewording.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/ParallelizableAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/ParallelizableAttribute.cs
@@ -34,8 +34,6 @@ namespace NUnit.Framework
     [AttributeUsage( AttributeTargets.Assembly | AttributeTargets.Class | AttributeTargets.Method, AllowMultiple=false, Inherited=true )]
     public class ParallelizableAttribute : PropertyAttribute, IApplyToContext
     {
-        private ParallelScope _scope;
-
         /// <summary>
         /// Construct a ParallelizableAttribute using default ParallelScope.Self.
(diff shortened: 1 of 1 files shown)


jadarnel27:
Thanks!  I was not sure of the process for updating the docs, so definitely let me know if I made a misstep by directly editing the wiki like that.  I'm pretty new to using GitHub, and didn't see an operation comparable to a "pull request" on the wiki page.

CharliePoole:
@jadarnel27 No problem... direct editing is how we do simple changes. Just use judgement and for anything that needs a review, just make the change and ask somebody to review it after the fact. For a completely new page, just create it without linking it in anywhere. For a major rewrite, treat like a new page with a temporary name.

PR #2382, opened 2017-08-24
Write TestParametersDictionary to xml result file in readable format (#2298)

### LINKED ISSUE #2298: Write TestParametersDictionary to xml result file in readable format
I just went to look up TestParameters in the results xml.

```
<setting name="TestParameters" value="1=d;2=e" />
<setting name="TestParametersDictionary" value="System.Collections.Generic.Dictionary`2[System.String,System.String]" />
```

It would be good if we can write that out TestParametersDictionary in a more readable fashion in the TestResult xml - the current output isn't much use.

grimmi:
Fixes #2298

This writes dictionary parameters as extra nodes below the setting element.

### DIFF
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -437,7 +437,25 @@ namespace NUnit.Framework.Api
         {
             TNode setting = new TNode("setting");
             setting.AddAttribute("name", name);
-            setting.AddAttribute("value", value.ToString());
+
+            if (value is IDictionary)
+            {
+                var values = value as IDictionary;
+
+                var valueEntries = new List<string>();
+                foreach (var key in values.Keys)
+                {
+                    var settingValue = values[key];
+                    valueEntries.Add($"[{key}, {values[key]}]");
+                }
+
+                var valueString = string.Join(", ", valueEntries.ToArray());
+                setting.AddAttribute("value", valueString);
+            }
+            else
+            {
+                setting.AddAttribute("value", value.ToString());
+            }
 
             settingsNode.ChildNodes.Add(setting);
         }
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -440,17 +440,7 @@ namespace NUnit.Framework.Api
 
             if (value is IDictionary)
             {
-                var values = value as IDictionary;
-
-                var valueEntries = new List<string>();
-                foreach (var key in values.Keys)
-                {
-                    var settingValue = values[key];
-                    valueEntries.Add($"[{key}, {values[key]}]");
-                }
-
-                var valueString = string.Join(", ", valueEntries.ToArray());
-                setting.AddAttribute("value", valueString);
+                AddDictionaryEntries(setting, value as IDictionary);
             }
             else
             {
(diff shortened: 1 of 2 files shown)


jnm2:
@grimmi Awesome, thanks for doing this! I added “Fixes #2298” to your PR body because that tracks and closes that issue.

rprouse:
This looks good to me, but I agree with @jnm2 that we should use the same names in the XML as IDictionary exposes. Please rename `entry` to `item` and `name` to `value`. Once that is done, I think we can merge. I am going to give my approval on the assumption that this change will be made.

@grimmi I am hoping to get the 3.8 release out in the next day or two. Do you think you could knock this off by then? I am going to add the associated issue to the milestone so I can track it.

grimmi:
Yep, going to do it this evening. I'm out at the moment, but will do it some time tonight.

jnm2:
@rprouse Just checking, looks like you meant changing `name` to `key`, and a a veto on moving the `value` attribute contents to the element contents?

CharliePoole:
I'm dubious about the use of `<item>` because it seems much too general. As it is, any code looking at a `<setting>` will now have to distinguish between settings with a `value` attribute and settings with inner XML as it's value. If, in the future, we add other collection types, we will have to distinguish `<item>` entries that contain `key` and `value` attributes from those that only contain values. This is pretty complicated to do, particularly in a transform. Of course, our code can handle it, but I'm thinking about people who post-process the XML.

There are two ways to deal with possible future change. One is to wait for the future, which we all know is my preferred approach. Change is inevitable but any __particular__ change is not. Preparing for the future often means preparing for the wrong future.

The other approach is to act in a way that leaves everything open.

> _In fact, the first approach is actually a degenerate case of this one - doing nothing leaves everything open. But I'm talking about doing __something__ here. 😄_

The approach being picked seems to fall somewhere between those two choices. If we __really__ want to be general, then the suggestion of using inner text or xml as the value is the most general. Simply picking a particular attribute or element name only works if __some particular future__ arrives.

As I stated on the issue, I'm backing off this one. Frankly, it looks to me as if it is being decided in haste in order to get it into the 3.8 release. @ChrisMaddock originally designated this as an `easyfix`, probably thinking we should do the minimum in order to solve the immediate issue. But if folks want to generalize for possible (?) future expansion, I think that the "easiness" drops out, even if the code is easy to write.

CharliePoole:
Point of procedure: When @rprouse approves, is that intended to cancel the "Changes Requested" by @jnm2? GitHub, of course, thinks not but that's how I perceive it. @rprouse ?

mikkelbu:
Regarding procedure, I agree with @CharliePoole. @rprouse should have the final say (but of cause we should try to get consensus about a given issue and its solution).
Ps. Sorry for the short comments, but I on my phone.

jnm2:
Rob can also feel free to dismiss my reviews for expediency. In this case was no longer relevant anyway. =)

grimmi:
I hope this is what we ended up with in the discussion. You wrote to change `name` to `value`, but that was a typo, right?

rprouse:
> Point of procedure: When @rprouse approves, is that intended to cancel the "Changes Requested" by @jnm2? GitHub, of course, thinks not but that's how I perceive it. @rprouse ?

In most cases, I don't think that my approval overrides others. It is just that if someone else on the team has requested changes, I feel that if I also request the same changes it slows the process down because now two team members need to come back and re-review. So my approval is pending the other team members requests being satisfied.

@grimmi there seems to be a fair amount of disagreement on this issue, sorry. Thanks for pushing the changes, I will re-review, but I am also going to go back and re-read the issue. With this much disagreement I am clearly missing something. 😄 

I will get back to you soon...

PR #2383, opened 2017-08-25
Limit throws to exception types. For both TypeOf and InstanceOf.

WdeBruin:
See the discussion in the issue topic for detailed info.

### DIFF
--- a/src/NUnitFramework/framework/Throws.cs
+++ b/src/NUnitFramework/framework/Throws.cs
@@ -130,7 +130,7 @@ namespace NUnit.Framework
         /// <summary>
         /// Creates a constraint specifying the exact type of exception expected
         /// </summary>
-        public static ExactTypeConstraint TypeOf<TExpected>()
+        public static ExactTypeConstraint TypeOf<TExpected>() where TExpected: Exception
         {
             return TypeOf(typeof(TExpected));
         }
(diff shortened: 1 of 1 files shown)


CharliePoole:
What you have looks good but I think you also need to look at `Throws.TypeOf`` and `Throws.InstanceOf`.

CharliePoole:
@jnm2 `Assert.Throws` already uses a constraint.

jnm2:
> liiks liek 

😆 Guess I'm abusing the single-line "dismiss review" textbox– hard to proofread typos!

rprouse:
This looks good to me. @CharliePoole are you saying he should make changes to the non-generic `TypeOf` and `InstanceOf` too? If so, are you proposing a guard clause to check the type passed in?

CharliePoole:
@rprouse I misspoke (typed)! Comes from trying to review on the phone. 😕 

I got it backwards. He has changed the generic Throws methods but not the constructor, which I think should also be changed since it can be used independently.

WdeBruin:
Thanks for the feedback. But what exactly do you mean with the ThrowsExceptionConstraint constructor?@CharliePoole 
That class doesn't have a constructor override as far as I can see. The ThrowsExceptionConstraintResult is already taking an Exception as parameter.

CharliePoole:
@WdeBruin You're right. I was looking at local experimental code.

PR #2384, opened 2017-08-26
Mikkelbu remove collection contains constraint

### LINKED ISSUE #1324: Remove CollectionContainsConstraint

rprouse:
This is @mikkelbu's PR with merge conflicts resolved, see #2239. I cannot push to his fork, so I pushed to this branch. I tested locally on my machine, but want to see CI pass before I do the final merge. I will merge once that is done since the review was completed in #2239.

This should auto-close #2239 when it this is merged. Also fixes #1324.


rprouse:
While waiting for CI, I have also gone through the diff for this PR and compared it to the original PR as a sanity check. Both are the same.

rprouse:
AppVeyor failure is the .NET 3.5 message pump issue we have been seeing regularly. Travis failure on latest is another Mono crash.

PR #2391, opened 2017-08-28
Write test parameters dictionary out to the legacy XML value attribute

### LINKED ISSUE #2390: ColorConsoleWriter.WriteLabel causes NullReferenceException
@jnm2 commented on [Mon Aug 28 2017](https://github.com/nunit/nunit-console/issues/285)

I've gotten this 7 times out of the last 9 builds on the build server at work. <s>The only thing that should have changed from last night is that there's a lot more parallelization going on.</s> Oh, also I updated to NUnit Framework 3.8.0, duh...

```c#
Run Settings
    DisposeRunners: True
    WorkDirectory: [redacted]
    TestParametersDictionary: System.NullReferenceException: Object reference not set to an instance of an object.
   at NUnit.ConsoleRunner.ColorConsoleWriter.WriteLabel(String label, Object option, ColorStyle valueStyle)
   at NUnit.ConsoleRunner.ColorConsoleWriter.WriteLabelLine(String label, Object option, ColorStyle valueStyle)
   at NUnit.ConsoleRunner.ResultReporter.WriteRunSettingsReport()
   at NUnit.ConsoleRunner.ResultReporter.ReportResults()
   at NUnit.ConsoleRunner.ConsoleRunner.RunTests(TestPackage package, TestFilter filter)
   at NUnit.ConsoleRunner.Program.Main(String[] args)
```

Fails with error -100.

---

@jnm2 commented on [Mon Aug 28 2017](https://github.com/nunit/nunit-console/issues/285#issuecomment-325459795)

This looks like the only line that could possibly be causing the NRE: https://github.com/nunit/nunit-console/blob/master/src/NUnitConsole/nunit3-console/ColorConsoleWriter.cs#L111. We should either write nothing or add an ArgumentNullException check to start with.

---

@jnm2 commented on [Mon Aug 28 2017](https://github.com/nunit/nunit-console/issues/285#issuecomment-325461740)

And here is the root problem: https://github.com/nunit/nunit-console/blob/master/src/NUnitConsole/nunit3-console/ResultReporter.cs#L95 needs to be updated in concert with https://github.com/nunit/nunit/pull/2382. We have a compatibility issue now that none of us thought of. Due to NUnit Console's NRE, NUnit Framework 3.8.0 is incompatible with NUnit Console 3.7.0 and earlier. Even without the NRE, there's no fallback so it would look like there are no test parameters and that may be a compat concern in itself.

Can we tag this issue `integration test wishlist`? ... 😟

/cc @ChrisMaddock @rprouse @CharliePoole 

---

@rprouse commented on [Mon Aug 28 2017](https://github.com/nunit/nunit-console/issues/285#issuecomment-325468394)

Well, damn. @CharliePoole was worried that removing the value attribute and moving it into the items might cause problems for transforms. None of us thought to check our own code and our own tests don't cause this to happen.

I think this warrants an emergency hotfix which I can do tonight. I think we have two options,

1. Back out #2382
2. Rework #2382 so that it is backwards compatible by including the dictionary settings as a single string and also include the individual elements for updated runners.
3. Rework #2382 removing the child elements and just add the value.

My preference is for the second option. I can make the changes starting in about an hour.

@jnm2 will you be available this evening for code reviews?

---

@ChrisMaddock commented on [Mon Aug 28 2017](https://github.com/nunit/nunit-console/issues/285#issuecomment-325470960)

2 sounds good to me too.

Is it worth investigation https://github.com/nunit/nunit/issues/2386 before doing a hotfix? That also sounds potentially critical.

---

@rprouse commented on [Mon Aug 28 2017](https://github.com/nunit/nunit-console/issues/285#issuecomment-325471747)

@ChrisMaddock I agree, nunit/nunit#2386 also looks serious, we should track both. I will comment there.

---

@rprouse commented on [Mon Aug 28 2017](https://github.com/nunit/nunit-console/issues/285#issuecomment-325472312)

I am also going to copy this issue to the framework repo for the backwards compatible fix. We should also address this in the next release of the framework to display the child elements if available, so I will leave this open in this repo.

rprouse:
This writes the dictionary settings out to a value attribute in addition to the child elements. This allows legacy runners (like the console) to continue to work if they expect a value attribute.

The XML now looks like this,

```xml
<setting name="TestParametersDictionary" value="[1, d], [2, e]">
  <item key="1" value="d" />
  <item key="2" value="e" />
</setting>
<setting name="TestParameters" value="1=d;2=e" />
```

And the console output looks like,

```
Run Settings
    DisposeRunners: True
    WorkDirectory: C:\src\NUnit\nunit
    TestParametersDictionary: [1, d], [2, e]
    TestParameters: 1=d;2=e
    ImageRequiresX86: False
    NumberOfTestWorkers: 8
```

This merge targets the release branch in preparation for a hotfix. I will create a PR to merge the hotfix into master once the hotfix is released.

Fixes #2390

### DIFF
--- a/src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
+++ b/src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
@@ -92,30 +92,30 @@ namespace NUnit.Framework.Api
         }
 
         #region SettingsElement Tests
-        
-        [Test]
-        public void InsertSettingsElement_MixedSettings_CreatesCorrectSubNodes()
+
+        [TestCaseSource(nameof(SettingsData))]
+        public void InsertSettingsElement_MixedSettings_CreatesCorrectSubNodes(string value)
         {
             var outerNode = new TNode("test");
             var testSettings = new Dictionary<string, object>
             {
                 ["key1"] = "value1",
-                ["key2"] = new Dictionary<string, object> { ["innerkey"] = "innervalue" }
+                ["key2"] = new Dictionary<string, object> { ["innerkey"] = value }
             };
 
             var inserted = FrameworkController.InsertSettingsElement(outerNode, testSettings);
 #if PARALLEL
             // in parallel, an additional node is added with number of test workers
-            Assert.That(3, Is.EqualTo(inserted.ChildNodes.Count));
+            Assert.That(inserted.ChildNodes.Count, Is.EqualTo(3));
 #else
-            Assert.That(2, Is.EqualTo(inserted.ChildNodes.Count));
+            Assert.That(inserted.ChildNodes.Count, Is.EqualTo(2));
 #endif
-            Assert.That("key1", Is.EqualTo(inserted.ChildNodes[0].Attributes["name"]));
-            Assert.That("value1", Is.EqualTo(inserted.ChildNodes[0].Attributes["value"]));
+            Assert.That(inserted.ChildNodes[0].Attributes["name"], Is.EqualTo("key1"));
+            Assert.That(inserted.ChildNodes[0].Attributes["value"], Is.EqualTo("value1"));
 
             var innerNode = inserted.ChildNodes[1].FirstChild;
-            Assert.That("innerkey", Is.EqualTo(innerNode.Attributes["key"]));
-            Assert.That("innervalue", Is.EqualTo(innerNode.Attributes["value"]));
+            Assert.That(innerNode.Attributes["key"], Is.EqualTo("innerkey"));
+            Assert.That(innerNode.Attributes["value"], Is.EqualTo(value));
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
```diff
@@ -92,30 +92,30 @@ public void CreateController()
...
                 ["key1"] = "value1",
-                ["key2"] = new Dictionary<string, object> { ["innerkey"] = "innervalue" }
+                ["key2"] = new Dictionary<string, object> { ["innerkey"] = value }
             };
 
             var inserted = FrameworkController.InsertSettingsElement(outerNode, testSettings);
 #if PARALLEL
             // in parallel, an additional node is added with number of test workers
-            Assert.That(3, Is.EqualTo(inserted.ChildNodes.Count));
+            Assert.That(inserted.ChildNodes.Count, Is.EqualTo(3));
```
rprouse: All of these asserts had the expected/actual backwards.

src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
```diff
@@ -132,63 +132,89 @@ public void InsertSettingsElement_SettingIsValue_CreatesASettingElementPerKey()
...
+            var testSettings = new Dictionary<string, object>
+            {
+                ["outerkey"] = new Dictionary<string, object> { { "key1", "value1" }, { "key2", value } }
+            };
+
+            var inserted = FrameworkController.InsertSettingsElement(outerNode, testSettings);
+            var settingNode = inserted.FirstChild;
+
+            Assert.That(settingNode.ChildNodes.Count, Is.EqualTo(2));
+            Assert.That(settingNode.Attributes["value"], Is.EqualTo($"[key1, value1], [key2, {value}]"));
```
rprouse: This is the main test for the change.

src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
```diff
@@ -132,63 +132,89 @@ public void InsertSettingsElement_SettingIsValue_CreatesASettingElementPerKey()
...
+        }
+
+        public static IEnumerable SettingsData()
+        {
+            yield return new TestCaseData("value");
+            yield return new TestCaseData("");
+            yield return new TestCaseData("<value>");
+            yield return new TestCaseData("\"value\"");
+            yield return new TestCaseData("'value'");
+            yield return new TestCaseData("value1;value2");
```
rprouse: I wanted to test values that might cause problems in an attribute. There was nothing to worry about, but it gave me a bit of confidence in the change 😄

src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
jnm2: Do we need the `ChildNodes.Count` test? It seems like that assert matches the test name `CreatesEntriesWithKeysAndValuesFromDictionary` better.

jnm2: Oh wait, it's also in `CreatesEntriesForDictionaryElements` already.

jnm2: @rprouse No biggie =)

PR #2393, opened 2017-08-28
Fixes Contains.Item().And.Contains()

### LINKED ISSUE #2386: Contains.Item() fails for collections in NUnit 3.8
Hi,

just updated from NUnit 3.7.1 to 3.8.0.
Now, a test starts to fail consistently which succeeded in the past.

Repro:

```

        [Test]
        public void Collection_regression_in_380() => Assert.That(new[] { "test", "test2", "test3" }, Contains.Item("test").And.Contains("test2").And.Contains("test3"));
```

Best regards,
Ralf Koban

rprouse:
Fixes #2386 

This PR is being merged into release in preparation for a 3.8.1 hotfix release. I will create a separate PR to merge all changes from the hotfix back into master.

### DIFF
--- a/src/NUnitFramework/framework/Contains.cs
+++ b/src/NUnitFramework/framework/Contains.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
@@ -37,10 +37,8 @@ namespace NUnit.Framework
         /// Returns a new <see cref="EqualConstraint"/> checking for the
         /// presence of a particular object in the collection.
         /// </summary>
-        public static EqualConstraint Item(object expected)
-        {
-            return Has.Some.EqualTo(expected);
-        }
+        public static ContainsConstraint Item(object expected) =>
+            new ContainsConstraint(expected);
 
         #endregion
 
--- a/src/NUnitFramework/framework/Contains.cs
+++ b/src/NUnitFramework/framework/Contains.cs
@@ -34,11 +34,11 @@ namespace NUnit.Framework
         #region Item
 
         /// <summary>
-        /// Returns a new <see cref="EqualConstraint"/> checking for the
+        /// Returns a new <see cref="SomeItemsConstraint"/> checking for the
         /// presence of a particular object in the collection.
         /// </summary>
-        public static ContainsConstraint Item(object expected) =>
-            new ContainsConstraint(expected);
+        public static SomeItemsConstraint Item(object expected) =>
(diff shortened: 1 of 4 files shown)


src/NUnitFramework/framework/Contains.cs
```diff
@@ -37,10 +37,8 @@ public class Contains
         /// Returns a new <see cref="EqualConstraint"/> checking for the
```
jnm2: XML doc needs to be updated

src/NUnitFramework/framework/Does.cs
```diff
@@ -63,20 +63,7 @@ public static FileOrDirectoryExistsConstraint Exist
         /// Returns a new <see cref="EqualConstraint"/> checking for the
```
jnm2: XML doc also out of date

src/NUnitFramework/framework/Does.cs
```diff
@@ -63,20 +63,7 @@ public static FileOrDirectoryExistsConstraint Exist
...
-        /// <summary>
-        /// Returns a new ContainsConstraint. This constraint
-        /// will, in turn, make use of the appropriate second-level
-        /// constraint, depending on the type of the actual argument. 
-        /// This overload is only used if the item sought is a string,
-        /// since any other type implies that we are looking for a 
-        /// collection member.
-        /// </summary>
-        public static ContainsConstraint Contain(string expected)
+        public static ContainsConstraint Contain(object expected)
```
jnm2: I think we should keep the string overload for the sake of the XML docs, don't you?

src/NUnitFramework/tests/Constraints/CollectionEqualsTests.cs
```diff
@@ -72,5 +75,32 @@ public void HonorsIgnoreCase( IEnumerable expected, IEnumerable actual )
...
+            var collection = new[] { 1, 2, 3 };
+            Assert.That(collection, Does.Contain(1).And.Contains(2).And.Contains(3));
+        }
+
+        [Test]
+        public void ContainsMultipleItemsInt()
+        {
+            var collection = new[] { 1, 2, 3 };
+            Assert.That(collection, Contains.Item(1).And.Contains(2).And.Contains(3));
+        }
```
jnm2: Tests look good, but seems weird to put them in `CollectionEqualsTests`. Do we have other tests on (collection) Contains already?

src/NUnitFramework/framework/Does.cs
```diff
@@ -63,23 +63,8 @@ public static FileOrDirectoryExistsConstraint Exist
...
-        /// will, in turn, make use of the appropriate second-level
-        /// constraint, depending on the type of the actual argument. 
-        /// This overload is only used if the item sought is a string,
-        /// since any other type implies that we are looking for a 
-        /// collection member.
-        /// </summary>
-        public static ContainsConstraint Contain(string expected)
-        {
-            return new ContainsConstraint(expected);
-        }
```
jnm2: I think we should keep the string overload for the sake of intellisense.

src/NUnitFramework/framework/Contains.cs
rprouse: Good catch, thanks. I shouldn't rush 😄

src/NUnitFramework/tests/Constraints/CollectionEqualsTests.cs
rprouse: Agreed, I'd looked for a better place, but didn't find one. I went back and the best I can find is a fixture called `ListContentTests` so I am moving there.

rprouse:
@jnm2 I have pushed changes based on your review. When you get a chance, can you re-review? Thanks 👍

PR #2401, opened 2017-08-31
Fix spurious kill failure

### LINKED ISSUE #2398: NUnit CI spurious failures, NUnit.Framework.Internal.ThreadUtilityTests.Kill
Too frequently I see NUnit's CI fail due to the below failure:
>1) Failed : NUnit.Framework.Internal.ThreadUtilityTests.Kill
  Native message pump was not able to be interrupted to enable a managed thread abort.
  Expected: True
  But was:  False
   at NUnit.Framework.Internal.ThreadUtilityTests.AbortOrKillThreadWithMessagePump(Boolean kill)

as seen [here](https://ci.appveyor.com/project/CharliePoole/nunit/build/3.9.0-ci-04352-pr-2379#L1233).

jnm2:
Fixes #2398 

So what I'm measuring is that when the CPU is under load, timers get around ten times slower to execute callbacks. Can't think of any mitigations for CPU load except a sufficiently long timeout. When the CPU is under load, what else is there?
We don't ever expect failures from this test, so how long can you tolerate waiting for results from CI? I say we make this a 15 second timeout and reopen #2398 if it comes back. Sound good?

### DIFF
--- a/src/NUnitFramework/tests/Internal/ThreadUtilityTests.cs
+++ b/src/NUnitFramework/tests/Internal/ThreadUtilityTests.cs
@@ -34,7 +34,15 @@ namespace NUnit.Framework.Internal
                 else
                     ThreadUtility.Abort(thread, nativeId);
 
-                Assert.That(thread.Join(1000), "Native message pump was not able to be interrupted to enable a managed thread abort.");
+                // When not under CPU load, we expect the thread to take between 100 and 200 ms to abort.
+                // However we get spurious failures in CI if we only wait 1000 ms.
+                // Using AbortOrKillThreadWithMessagePump_StressTest (times: 1000, maxParallelism: 20),
+                // I measured timer callbacks to be firing around ten times later than the 100 ms requested.
+                // All this number does is manage how long we can tolerate waiting for a build if a bug
+                // is introduced and the thread never aborts.
+                const int waitTime = 15000;
+
+                Assert.That(thread.Join(waitTime), "Native message pump was not able to be interrupted to enable a managed thread abort.");
             }
         }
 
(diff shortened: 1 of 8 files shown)


jnm2:
@rprouse Wheeler says he's seeing this in many CI failures so this may stabilize things.

src/NUnitFramework/tests/Internal/ThreadUtilityTests.cs
```diff
@@ -33,10 +34,28 @@ public void AbortOrKillThreadWithMessagePump(bool kill)
...
+                // All this number does is manage how long we can tolerate waiting for a build if a bug
+                // is introduced and the thread never aborts.
+                const int waitTime = 15000;
+
+                Assert.That(thread.Join(waitTime), "Native message pump was not able to be interrupted to enable a managed thread abort.");
             }
         }
 
+        [Platform("Win")]
+        [Test, Explicit("For diagnostic purposes; slow")]
```
CharliePoole: I'm OK with this being here so long as it's explicit. I'd actually prefer it to move to our integration tests, whenever we set some up in a separate repo. I may end up creating something.

src/NUnitFramework/tests/TestUtilities/StressUtility.cs
```diff
@@ -0,0 +1,102 @@
+﻿using System;
```
mikkelbu: We have forgot a copyright notice here

jnm2: Oh, thank you!

jnm2:
Updated with headers.

PR #2406, opened 2017-09-01
Summary descriptions replaced by more detailed ones

RalfKoban:
Hi,

I know there was no issue for that but it would be really nice if the summary descriptions could be updated so that IntelliSense could show more detailed descriptions than "Summary description for ...".

Best regards,
Ralf Koban

### DIFF
--- a/src/NUnitFramework/framework/Attributes/SetCultureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/SetCultureAttribute.cs
@@ -30,13 +30,10 @@ namespace NUnit.Framework
 {
     /// <summary>
     /// Sets the current Culture for the duration of a test.
-    /// </summary>
-    /// <remarks>
+    /// <para />
     /// It may be specified at the level of a test or a fixture.
     /// The culture remains set until the test or fixture completes and is then reset to its original value.
-    /// <para />
-    /// If you wish to use the current culture setting to decide whether to run a test, use the <see cref="CultureAttribute"/> instead of this one.
-    /// </remarks>
+    /// </summary>
     /// <seealso cref="SetUICultureAttribute"/>
     [AttributeUsage(AttributeTargets.Class|AttributeTargets.Method|AttributeTargets.Assembly, AllowMultiple=false, Inherited=true)]
     public class SetCultureAttribute : PropertyAttribute, IApplyToContext
--- a/src/NUnitFramework/framework/Attributes/SetCultureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/SetCultureAttribute.cs
@@ -29,8 +29,15 @@ using NUnit.Framework.Internal;
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Summary description for SetCultureAttribute.
+    /// Sets the current Culture for the duration of a test.
     /// </summary>
+    /// <remarks>
+    /// It may be specified at the level of a test or a fixture.
+    /// The culture remains set until the test or fixture completes and is then reset to its original value.
+    /// <para />
+    /// If you wish to use the current culture setting to decide whether to run a test, use the <see cref="CultureAttribute"/> instead of this one.
+    /// </remarks>
+    /// <seealso cref="SetUICultureAttribute"/>
     [AttributeUsage(AttributeTargets.Class|AttributeTargets.Method|AttributeTargets.Assembly, AllowMultiple=false, Inherited=true)]
     public class SetCultureAttribute : PropertyAttribute, IApplyToContext
     {
--- a/src/NUnitFramework/framework/Attributes/SetCultureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/SetCultureAttribute.cs
@@ -30,9 +30,10 @@ namespace NUnit.Framework
 {
     /// <summary>
     /// Sets the current Culture for the duration of a test.
-    /// <para />
+    /// <para>
     /// It may be specified at the level of a test or a fixture.
     /// The culture remains set until the test or fixture completes and is then reset to its original value.
+    /// </para>
     /// </summary>
     /// <seealso cref="SetUICultureAttribute"/>
     [AttributeUsage(AttributeTargets.Class|AttributeTargets.Method|AttributeTargets.Assembly, AllowMultiple=false, Inherited=true)]
--- a/src/NUnitFramework/framework/Attributes/SetUICultureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/SetUICultureAttribute.cs
@@ -29,8 +29,15 @@ using NUnit.Framework.Internal;
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Summary description for SetUICultureAttribute.
+    /// Sets the current UI Culture for the duration of a test.
     /// </summary>
+    /// <remarks>
+    /// <see cref="SetUICultureAttribute"/> may be specified at the level of a test or a fixture.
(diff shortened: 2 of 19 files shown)


src/NUnitFramework/framework/Attributes/SetCultureAttribute.cs
```diff
@@ -29,8 +29,15 @@
...
     /// <summary>
-    /// Summary description for SetCultureAttribute.
+    /// Sets the current Culture for the duration of a test.
     /// </summary>
+    /// <remarks>
+    /// It may be specified at the level of a test or a fixture.
+    /// The culture remains set until the test or fixture completes and is then reset to its original value.
+    /// <para />
+    /// If you wish to use the current culture setting to decide whether to run a test, use the <see cref="CultureAttribute"/> instead of this one.
+    /// </remarks>
```
CharliePoole: Two uncertainties here...

1. Where do remarks show up? If they don't come up when I right-click, I don't see any advantage to having them - better to expand the summary. We've discussed this before but I can't remember the resolution - I _think_ we decided not to use Remarks. @nunit/framework-team ?

2. I don't know about putting usage info like "If you wish to use..." in the Intellisense. Seems like it gets in the way, giving the reader something extra to look at each time and figure out if they are in the right place **Set**Culture seems pretty clear to me.

src/NUnitFramework/framework/Attributes/SetUICultureAttribute.cs
```diff
@@ -29,8 +29,15 @@
...
-    /// Summary description for SetUICultureAttribute.
+    /// Sets the current UI Culture for the duration of a test.
     /// </summary>
+    /// <remarks>
+    /// <see cref="SetUICultureAttribute"/> may be specified at the level of a test or a fixture.
+    /// The UI culture remains set until the test or fixture completes and is then reset to its original value.
+    /// <para />
+    /// If you wish to use the current culture setting to decide whether to run a test, use the <see cref="CultureAttribute"/> instead of this one.
+    /// </remarks>
+    /// <seealso cref="SetCultureAttribute"/>
```
CharliePoole: See SetCulture comments

src/NUnitFramework/framework/Attributes/SetCultureAttribute.cs
jnm2: 1. `<remarks>` are invisible to IntelliSense unless you are missing the `<summary>`, in which case it replaces it. Since we require `<summary>`, the Object Browser is the only place in the IDE that you can see `<remarks>`, and hypothetically if we generated website content it would show up there.
    
    If the information is interesting to the general user, I'd say it should be in the summary because putting it in `<remarks>` gives the false impression that people are going to see it and benefit from it. If we agree on this I can add this rationale to our best practices page. I want to know what others think.

 2. I'm actually amenable to this kind of documentation. Not that the user doesn't know what <code>**Set**Culture</code> means, but just a way to notify you that the other thing exists if that interests you. I could go either way.

CharliePoole: 1. If we were to generate website content, then I'd be in favor of putting the first `para` in the summary and the second in remarks.  In fact, that's what I suggest we do here. 

As an experienced user, I'm frequently annoyed at Microsoft for telling me "obvious" things, without the ability to turn off the stream of advice. I don't want us to do that.

If the Intellisense has a sentence suggesting I may not be using the correct attribute, I'm going to pause for a few seconds, read it and think about it. I don't want to do that in this case. I think that comparisons with other attributes belong in the docs. Warnings about use of an attribute only belong here in the case where that use is __usually wrong__.  And, of course, if an attribute is usually used incorrectly, we shold probably change it!

jnm2: > I think that comparisons with other attributes belong in the docs.

Which in your temporary hypothesis earlier you assumed was generated from XML docs...

jnm2: > I think that comparisons with other attributes belong in the docs.

If we went the direction of generating website content from XML docs, which would it be? Comparison in the docs = website = XML docs, or not?

CharliePoole: Yes. I think that would be a good direction.

But please see my comment about what I think should be done with the current PR in the first paragraph of my last comment. Let's resolve this PR rather than getting into general discussions about the website here... that's a separate matter and doesn't actually involve me.

jnm2:
@RalfKoban Thanks for taking it on yourself to improve this!

src/NUnitFramework/framework/Attributes/SetCultureAttribute.cs
jnm2: Ah, I mixed up which sections you meant. I don't think that's how `<para>` is meant to be used, but I could be wrong.  I've only ever seen it used to contain each paragraph as contents.

Anyway, I'm good with that. @RalfKoban Does that make sense?

RalfKoban: @jnm2  Yes, I'm OK with that.

jnm2:
@RalfKoban Just fyi, rebases are preferable on feature and PR branches rather than merges. The diff gets confusing to review with a normal merge. What I recommend using to update to the latest upstream master is `git pull -r upstream master`– this fetches upstream/master and then rebases (replays your changes on top of the new tip), giving you a chance to resolve conflicts for each of your commits (if any).

If that's too confusing, don't worry about it and we'll squash your changes into a single commit when we merge the PR. Git can be confusing and it's totally up to you how much time you want to sink into it. =) If you have questions I'm happy to help.

OmicronPersei:
Also, @RalfKoban , I recommend creating a temporary backup branch of your feature branch before you rebase.  It helps in the event that you mess up manually merging conflicts... I can speak from experience that it helps 😄 .

RalfKoban:
@jnm2 , @OmicronPersei : Sorry, the merge should not have happened. Did not want to merge it into nunit master branch. Instead I wanted to merge the latest changes into my own nunit branch / fork.

jnm2:
@RalfKoban Well no, you did what you intended to: you merged the latest changes from upstream/master into your own branch in your own fork. That's just not the best for GitHub PRs though which is why I was recommending that you *not* merge the latest changes from upstream/master into your own branch but rather rebase your own branch on top of upstream/master. Does that sorta make sense?

jnm2:
@RalfKoban Don't worry, you can't mess up master. It takes two framework team members to merge into master.

RalfKoban:
As I'm little bit new to Git/GitHub, I'll close this pull request and will create another one which will contain the wanted changes on the remarks tag.

jnm2:
@RalfKoban Actually please don't do that! We'd prefer if you reopen and then push your changes to your `master` branch since that's what's driving the PR. They will show up here automagically. =)

jnm2:
Also if you're wanting to do a force push, force pushing to a pr branch is expected and welcome.

RalfKoban:
@jnm2 OK, I'll refactor that as well. Just as info - both ways are working; <para /> simply adds a new line.

jnm2:
@RalfKoban I expect it's similar to how `<p/>` works in browsers even though it's not idiomatic HTML. (In fact, HTML is exactly what happens to the `<para/>` were we to drive website or CHM documentation with it.) We should follow the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle).

Issue #2402, opened 2017-08-31
--labels=All doesn't show anything in console output executing NUnitLite Console Runner


netonjm:
Hello!

Seems I am getting an issue with the --labels=All. Seems this argument is not showing any output while executing the nunitlite console runner (from master 3256c91) over a my test library. 

Doing some debugging over the console runner I realized every result here:
https://github.com/nunit/nunit/blob/master/src/NUnitFramework/nunitlite/TextUI.cs#L251

It gets an empty Output value (this is correct?), and because this we never reach _displayBeforeOutput check to show Test.Fullname.

Also @CharliePoole would be fantastic include additional information here like the Duration in the same line, to catch timeout issues or if our runner is killed for some reason and we don't have the result file generated.

Thanks for your hard work guys it's awesome!

CharliePoole:
Could you clarify exactly what you are running? What is your command-line?

netonjm:
Ops! Sorry. Working from Mac.

mono build/UnitRunner/nunitlite-runner.exe XamarinProfiler.Tests.dll --labels=All

CharliePoole:
@ChrisMaddock I already fixed this in my ongoing PR for parallel execution, since I needed it in order to debug. It's a one-line change. Should I do a separate PR for it?

ChrisMaddock:
@netonjm was going to revisit this - so long as he's aware you've got it sorted @CharliePoole, sounds good to me!

Will leave it up to you as to whether to separate it out or not, I don't have any particular objection to it not being.

netonjm:
Sounds good to me 👍 !
Thanks @CharliePoole !

mikkelbu:
Hi @netonjm. The issue has automatically closed due to #2482. Regarding the wish for additional information in the labels, I think that this is best addressed in a new issue, as this is not a bug, but an improvement.

PR #2407, opened 2017-09-01
Fix show output labels when labels=All in runner

netonjm:
Fixes Bug #2402
In console runner the result.Output has an empty values
in Passed statements. We want show _displayBeforeOutput
in every case.

### DIFF
--- a/src/NUnitFramework/nunitlite/TextUI.cs
+++ b/src/NUnitFramework/nunitlite/TextUI.cs
@@ -248,11 +248,11 @@ namespace NUnitLite
 
         public void TestFinished(ITestResult result)
         {
+            if (_displayBeforeOutput)
+                WriteLabelLine(result.Test.FullName);
+                
             if (result.Output.Length > 0)
             {
-                if (_displayBeforeOutput)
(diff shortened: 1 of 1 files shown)


ChrisMaddock:
It would also be good to add a new unit test for this as well. We must be missing something, if it's been broken. 🙂

CharliePoole:
I suspect some or all of those 16 cases are covering this change. We seem to be expecting no before label to appear!

ChrisMaddock:
@netonjm - Are you interested in coming back and fixing the tests here?

This also appears to have inadvertently turned on labels permenantly for the NUnitLite runner - unless there's something going on in the CI build I'm not aware of.

netonjm:
Of course @ChrisMaddock! I have had a busy few days, but for this week I will try investigate about the test failures... 👍

ChrisMaddock:
Great, thanks! 😄 No rush, just been checking in on some of our older PRs!

ChrisMaddock:
This PR has been superseded by https://github.com/nunit/nunit/pull/2482. Thanks for your efforts regardless, @netonjm! 👍

Issue #2289, opened 2017-06-29
ResolveTypeNameDifference does not handle generic types well


mikkelbu:
Given two different classes that gives the same result via ToString(), e.g.
```
class Dummy
{
    internal readonly int value;

    public Dummy(int value)
    {
        this.value = value;
    }

    public override string ToString()
    {
        return "Dummy " + value;
    }
}

class Dummy1
{
    internal readonly int value;

    public Dummy1(int value)
    {
        this.value = value;
    }

    public override string ToString()
    {
        return "Dummy " + value;
    }
}
```
Then the following tests fails correctly with a message from `ResolveTypeNameDifference` that describes the type difference.
```
[Test]
public void SomeTest()
{
    var d1 = new Dummy(12);
    var d2 = new Dummy1(12);
    var tuple1 = new KeyValuePair<Dummy>("", d1);
    var tuple2 = new KeyValuePair<Dummy1>("", d2);

    Assert.That(tuple1, Is.EqualTo(tuple2));
}
```
The error message, however, does not handle the generic class very well
```
  Expected: (Dummy 12) (ValueTupleEqualityTests+Dummy1])
  But was:  (Dummy 12) (ValueTupleEqualityTests+Dummy])
```
I would have expected something like
```
  Expected: (Dummy 12) (KeyValuePair`2[String,ValueTupleEqualityTests+Dummy1])
  But was:  (Dummy 12) (KeyValuePair`2[String,ValueTupleEqualityTests+Dummy])
```

At the moment the code just splits on "." and there is no special handling of generic types.

OmicronPersei:
(Somewhat off topic...): What `KeyValuePair` type are you using that accepts only one templated parameter?

Also, I was able to confirm the same using the below example test (based on your `Dummy` and `Dummy1` types you specified):
```C#
[Test]
public void SomeTest()
{
	var d1 = new Dummy(12);
	var d2 = new Dummy1(12);
	var tuple1 = new Tuple<Dummy>(d1);
	var tuple2 = new Tuple<Dummy1>(d2);

	Assert.That(tuple1, Is.EqualTo(tuple2));
}
```
with a failure message of:
>Message:   Expected: <(Dummy 12)> (TextMessageWriterTests+Dummy1])
  But was:  <(Dummy 12)> (TextMessageWriterTests+Dummy])

OmicronPersei:
Also I'm willing to take up this one, unless you wanted to @mikkelbu ?

[...]

PR #2420, opened 2017-09-06
Added type name difference resolver for displaying differences in type

OmicronPersei:
Closes issue #2289 .

The feature adds the ability to display a shortened text output of the differences between two `Type`s when an equality of two objects is performed.  It also better handles comparing generic types.


src/NUnitFramework/tests/Constraints/EqualConstraintTests.cs
```diff
@@ -666,6 +666,66 @@ public void SameValueDifferentTypeExactMessageMatch()
...
+                this.value = value;
+            }
+
+            public override string ToString()
+            {
+                return "Dummy " + value;
+            }
+        }
+
+        class DummyTemplatedClass<T>
```
jnm2: Complete nit: the IL/C# lingo is "generic"– C++ has templates, IL and C# do not.

src/NUnitFramework/framework/Internal/TypeNameDifferenceResolverUtils.cs
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        /// <summary>
+        /// Get the generic parameters of a given <see cref="Type"/>.
+        /// </summary>
+        public List<Type> GetGenericParams(Type type)
+        {
+            Guard.ArgumentNotNull(type, nameof(type));
+
+            return new List<Type>(type.GetGenericArguments());
```
jnm2: I'm curious why we don't use `type.GetGenericArguments()` instead of maintaining the helper method? Also arguments and parameters are not the same thing, though the reflection API is already abusing the distinction...

src/NUnitFramework/framework/Internal/TypeNameDifferenceResolverUtils.cs
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        /// <summary>
+        /// Shorten the given <see cref="Type"/> names by only including the relevant differing namespaces/types, if they differ.
+        /// </summary>
+        /// <param name="expectedType">The expected <see cref="Type"/>.</param>
+        /// <param name="actuallType">The actual <see cref="Type"/>.</param>
+        /// <param name="expectedTypeShortened">The shortened expected <see cref="Type"/> name.</param>
+        /// <param name="actualTypeShortened">The shortened actual <see cref="Type"/> name.</param>
+        public void ShortenTypeNames(Type expectedType, Type actuallType, out string expectedTypeShortened, out string actualTypeShortened)
```
jnm2: `actuallType` should be `actualType`

src/NUnitFramework/tests/Constraints/EqualConstraintTests.cs
OmicronPersei: Whoops.  Will rename then.

src/NUnitFramework/framework/Internal/TypeNameDifferenceResolverUtils.cs
OmicronPersei: I think I intended to refactor that away after I refactored my functionality to use `Type`s instead of relying on the string representation of a `Type`. I think you're right, we don't need a new method for that, so as long as I put `using NUnit.Compatability;` at the top of the file!

OmicronPersei:
> The utils class is bugging me. I can't tell what responsibility is being isolated into the utils class. Every helper method could live happily in the main class in my opinion. Am I overlooking something we'd lose by moving them in?

Well, I was worried it would violate SRP.  The utilities provided aren't necessarily to provide `TypeNameDifferenceResolver`, but are solely used by the class.  Truthfully, I'm new to getting my code to conform to things like SRP, maybe I was too eager to move it to its own class.

jnm2:
> Well, I was worried it would violate SRP.

I don't think it matters in this case. In my mind to whatever extent you were violating SRP by having `IsGenericType` next to the methods in `TypeNameDifferenceResolver`, that's the same extent that you're still violating SRP by having it next to `TypeNameDifferenceResolverUtils.GetShortenedGenericTypes`.
If you were to worry about that, you'd add an `IsGenericType` extension method to `ReflectionExtensions`. **But** again I don't think it's worth worrying about. 

What I don't like is that `Foo` and `FooUtils` are acting as one class. By its name `FooUtils` is just an extension of `Foo`, each totally dependent on the other and pretty much useless without the other.

The type of scenario SRP should kick in for is if you had a bunch of related methods for which it would be reasonable to have more than one implementation, say for example parsing and formatting type names using a C#- or VB.NET-specific style rather than the generic CLR style names. Then instead of `Utils` you'd call the class `FormatProvider` or something like that. Formatting to a specific language in that scenario is a "reason to change" (because you might need to change to a new language) and so should be abstracted into a separate class through an interface or base class.
It's also usually unfruitful to guess ahead at potential future abstraction points so I'd wait until we have a concrete scenario. We aren't designing an API and we can quickly refactor at any time.

tl;dr; Since from what I see we don't have more than one "reason to change" or responsibility, let's stick with one class.

jnm2:
Restarted the failing Travis job.

mikkelbu:
Restarting CI once more

jnm2:
> Restarting CI once more

Oh boy.

You can restart individual jobs rather than the whole build, but either way, the original build log disappears! I actually wanted to file and issue on the error I saw before I clicked Restart Job and now it's too late.

mikkelbu:
> You can restart individual jobs rather than the whole build, but either way, the original build log disappears! 
I first discovered while restarting this time.

> I actually wanted to file and issue on the error I saw before I clicked Restart Job and now it's too late.
Sorry. We can hope for (yet) another error

jnm2:
I'm sorry, the "oh boy" was directed at Travis, not you.

mikkelbu:
> I'm sorry, the "oh boy" was directed at Travis, not you.

No need to be sorry 😄  (and I think that the build is going to fail again)

PR #2423, opened 2017-09-08
Fix Has.Property causing AmbiguousMatchException for shadowing properties

### LINKED ISSUE #2422: Has.Property causes AmbiguousMatchException for shadowing properties
```cs
ThirdPartyFoo actual = new ThirdPartyFooEx();
ThirdPartyFoo expected = new ThirdPartyFooEx();

Assert.That(actual, Has.Property(nameof(expected.Options)).EqualTo(expected.Options));
```

Causes:

> ```
> System.Reflection.AmbiguousMatchException : Ambiguous match found.
>    at System.RuntimeType.GetPropertyImpl(String name, BindingFlags bindingAttr, Binder binder, Type returnType, Type[] types, ParameterModifier[] modifiers)
>    at System.Type.GetProperty(String name, BindingFlags bindingAttr)
>    at NUnit.Framework.Constraints.PropertyConstraint.ApplyTo[TActual](TActual actual)
>    at NUnit.Framework.Assert.That[TActual](TActual actual, IResolveConstraint expression, String message, Object[] args)
>    at NUnit.Framework.Assert.That[TActual](TActual actual, IResolveConstraint expression)
>    at NUnit_repro.SomeFixture.SomeTest()
> ```

Given:

```cs
public class ThirdPartyFooOptions
{
}

public class ThirdPartyFoo
{
    public ThirdPartyFooOptions Options { get; }
}

public class ThirdPartyFooExOptions : ThirdPartyFooOptions
{
}

public class ThirdPartyFooEx : ThirdPartyFoo
{
    public new ThirdPartyFooExOptions Options => (ThirdPartyFooExOptions)base.Options;
}
```

We should be settling on the ultimate shadowing property, right? This is how `dynamic` would handle the situation and it makes sense to me since we likewise don't know the compile-time type of the expression being passed into `Assert.That` because that's not how we roll. In other words, `2` is printed for:

```cs
dynamic expected = new ThirdPartyFooEx();
Debug.WriteLine((int)expected.Options);

public class ThirdPartyFoo
{
    public int Options => 1;
}

public class ThirdPartyFooEx : ThirdPartyFoo
{
    public new int Options => 2;
}
```

So I think we should match that. I expect all .NET languages with dynamic features to treat IL shadowing the same.

jnm2:
Fixes #2422

### DIFF
--- a/src/NUnitFramework/framework/Internal/Reflect.cs
+++ b/src/NUnitFramework/framework/Internal/Reflect.cs
@@ -226,5 +226,89 @@ namespace NUnit.Framework.Internal
         }
 
         #endregion
+
+#if NETSTANDARD1_3
+        /// <summary>
+        /// <para>
+        /// Selects the ultimate shadowing property just like <see langword="dynamic"/> would,
+        /// rather than throwing <see cref="AmbiguousMatchException"/>
+        /// for properties that shadow properties of a different property type.
+        /// </para>
+        /// <para>
+        /// If you request both public and nonpublic properties, every public property is preferred
+        /// over every nonpublic property. It would violate the principle of least surprise for a
+        /// derived class’s implementation detail to be chosen over the public API for a type.
+        /// </para>
+        /// </summary>
+#elif NETSTANDARD1_6
+        /// <summary>
+        /// <para>
+        /// Selects the ultimate shadowing property just like <see langword="dynamic"/> would,
+        /// rather than throwing <see cref="AmbiguousMatchException"/>
+        /// for properties that shadow properties of a different property type
+        /// which is what <see cref="TypeInfo.GetProperty(string, BindingFlags)"/> does.
+        /// </para>
+        /// <para>
+        /// If you request both public and nonpublic properties, every public property is preferred
+        /// over every nonpublic property. It would violate the principle of least surprise for a
+        /// derived class’s implementation detail to be chosen over the public API for a type.
+        /// </para>
+        /// </summary>
+        /// <param name="type">See <see cref="TypeInfo.GetProperty(string, BindingFlags)"/>.</param>
+        /// <param name="name">See <see cref="TypeInfo.GetProperty(string, BindingFlags)"/>.</param>
+        /// <param name="bindingFlags">See <see cref="TypeInfo.GetProperty(string, BindingFlags)"/>.</param>
+#else
+        /// <summary>
+        /// <para>
+        /// Selects the ultimate shadowing property just like <see langword="dynamic"/> would,
+        /// rather than throwing <see cref="AmbiguousMatchException"/>
+        /// for properties that shadow properties of a different property type
+        /// which is what <see cref="Type.GetProperty(string, BindingFlags)"/> does.
+        /// </para>
+        /// <para>
+        /// If you request both public and nonpublic properties, every public property is preferred
+        /// over every nonpublic property. It would violate the principle of least surprise for a
+        /// derived class’s implementation detail to be chosen over the public API for a type.
+        /// </para>
+        /// </summary>
+        /// <param name="type">See <see cref="Type.GetProperty(string, BindingFlags)"/>.</param>
+        /// <param name="name">See <see cref="Type.GetProperty(string, BindingFlags)"/>.</param>
+        /// <param name="bindingFlags">See <see cref="Type.GetProperty(string, BindingFlags)"/>.</param>
+#endif
+        public static PropertyInfo GetUltimateShadowingProperty(Type type, string name, BindingFlags bindingFlags)
+        {
+            Guard.ArgumentNotNull(type, nameof(type));
+            Guard.ArgumentNotNull(name, nameof(name));
+
+            if ((bindingFlags & BindingFlags.DeclaredOnly) != 0)
+            {
+                // If you're asking us to search a hierarchy but only want properties declared in the given type,
+                // you're in the wrong place but okay:
+                return type.GetProperty(name, bindingFlags);
+            }
+
+            if ((bindingFlags & (BindingFlags.Public | BindingFlags.NonPublic)) == (BindingFlags.Public | BindingFlags.NonPublic))
+            {
+                // If we're searching for both public and nonpublic properties, search for only public first
+                // because chances are if there is a public property, it would be very surprising to detect the private shadowing property.
+
+                for (var publicSearchType = type; publicSearchType != null; publicSearchType = publicSearchType.GetTypeInfo().BaseType)
+                {
+                    var property = publicSearchType.GetProperty(name, (bindingFlags | BindingFlags.DeclaredOnly) & ~BindingFlags.NonPublic);
+                    if (property != null) return property;
+                }
+
+                // There is no public property, so may as well not ask to include them during the second search.
+                bindingFlags &= ~BindingFlags.Public;
+            }
+
+            for (var searchType = type; searchType != null; searchType = searchType.GetTypeInfo().BaseType)
+            {
+                var property = searchType.GetProperty(name, bindingFlags | BindingFlags.DeclaredOnly);
+                if (property != null) return property;
+            }
+
+            return null;
+        }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
@@ -52,7 +52,6 @@ namespace NUnit.Framework.Constraints
         /// Test whether the constraint is satisfied by a given value
         /// </summary>
         /// <param name="actual">The value to be tested</param>
-        /// <returns>True for success, false for failure</returns>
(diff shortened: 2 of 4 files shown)


src/NUnitFramework/framework/Internal/ReflectionUtils.cs
```diff
@@ -0,0 +1,99 @@
...
+#if !NET_4_5
+using NUnit.Compatibility;
+#endif
+
+namespace NUnit.Framework.Internal
+{
+    /// <summary>
+    /// Common reflection utility methods.
+    /// </summary>
+    internal static class ReflectionUtils
```
CharliePoole: Any reason not to add this to existing Reflect utility?

jnm2: Nope, missed it. Something felt wrong. Thanks!

jnm2: Fixed.

jnm2: Umm... or not, hang on.

jnm2: Much better.

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -226,5 +226,90 @@ public static object InvokeMethod( MethodInfo method, object fixture, params obj
...
+        /// <summary>
+        /// <para>
+        /// Selects the ultimate shadowing property just like <see langword="dynamic"/> would,
+        /// rather than throwing <see cref="AmbiguousMatchException"/>
+        /// for properties that shadow properties of a different property type.
+        /// </para>
+        /// <para>
+        /// If you request both public and nonpublic properties, every public property is preferred
+        /// over every nonpublic property. It would violate the principle of least surprise for a
+        /// derived class�s implementation detail to be chosen over the public API for a type.
```
ChrisMaddock: There's a corrupted character in this line - repeated for the other platforms as well. 🙂

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -226,5 +226,90 @@ public static object InvokeMethod( MethodInfo method, object fixture, params obj
...
+        /// </summary>
+        /// <param name="type">See <see cref="Type.GetProperty(string, BindingFlags)"/>.</param>
+        /// <param name="name">See <see cref="Type.GetProperty(string, BindingFlags)"/>.</param>
+        /// <param name="bindingFlags">See <see cref="Type.GetProperty(string, BindingFlags)"/>.</param>
+#endif
+
+        public static PropertyInfo GetUltimateShadowingProperty(Type type, string name, BindingFlags bindingFlags)
+        {
+            if (type == null) throw new ArgumentNullException(nameof(type));
+            if (name == null) throw new ArgumentNullException(nameof(name));
```
ChrisMaddock: How about `Guard.IfNull()`? For consistent handling. 🙂

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -226,5 +226,90 @@ public static object InvokeMethod( MethodInfo method, object fixture, params obj
...
+                {
+                    var property = publicSearchType.GetProperty(name, (bindingFlags | BindingFlags.DeclaredOnly) & ~BindingFlags.NonPublic);
+                    if (property != null) return property;
+                }
+
+                // There is no public property, so may as well not ask to include them during the second search.
+                bindingFlags &= ~BindingFlags.Public;
+            }
+
+            for (; type != null; type = type.GetTypeInfo().BaseType)
```
ChrisMaddock: More out of interest - what's going on here? 🙂 i.e. why half a for loop, and not something more explicit. I've no doubt you have a good reason!

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: Good idea! Forgot about them.

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: Oh yeah, I forgot to save as UTF8 when I moved this function into this file.
That's a terrific eye you have, thank you!

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: I guess I had better write it like the other loop then. It's doing the exact same iteration, only it's using `type` as the iteration variable rather than introducing a new `searchType` variable.
I wrote this loop first but I should come back and make it consistent with the one in the `if` block.

ChrisMaddock: Ah - it's working up the tree? Clever!

jnm2: Yeah, it's the linked-list style of `for` loop rather than the array style.

jnm2:
This actually has me stuck at work and I can't think of a workaround besides implementing my own property constraint temporarily:

```cs
private static void AssertSameAppearance<T>(T actual, ConstraintExpression root, AppearanceObject expected)
{
    if (root == null) root = new ConstraintExpression();

    Assert.That(actual, root.Property(nameof(expected.Options)).With.Property(nameof(expected.Options.UseBackColor)).EqualTo(expected.Options.UseBackColor));

    // ...

    Assert.That(actual, root.Property(nameof(expected.Options)).With.Property(nameof(expected.Options.UseBorderColor)).EqualTo(expected.Options.UseBorderColor));

    // ...
```

jnm2:
I've restarted the OS X jobs four or five times now and they fail with no build log so we may just have to ignore it.

jnm2:
Hey!! OS X finished successfully! :tada:

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -226,5 +226,89 @@ public static object InvokeMethod( MethodInfo method, object fixture, params obj
...
+        /// Selects the ultimate shadowing property just like <see langword="dynamic"/> would,
+        /// rather than throwing <see cref="AmbiguousMatchException"/>
+        /// for properties that shadow properties of a different property type.
+        /// </para>
+        /// <para>
+        /// If you request both public and nonpublic properties, every public property is preferred
+        /// over every nonpublic property. It would violate the principle of least surprise for a
+        /// derived class’s implementation detail to be chosen over the public API for a type.
+        /// </para>
+        /// </summary>
```
mikkelbu: Should we describe the three parameters here? (or just leave it be, since this is an internal implementation detail)

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -226,5 +226,89 @@ public static object InvokeMethod( MethodInfo method, object fixture, params obj
...
+
+                // There is no public property, so may as well not ask to include them during the second search.
+                bindingFlags &= ~BindingFlags.Public;
+            }
+
+            for (var searchType = type; searchType != null; searchType = searchType.GetTypeInfo().BaseType)
+            {
+                var property = searchType.GetProperty(name, bindingFlags | BindingFlags.DeclaredOnly);
+                if (property != null) return property;
+            }
```
mikkelbu: AFAICT we can extract the for-loops into a helper method that will accept a Type and BindingFlags and return a property or null, but I'm unsure whether it is worth the work.

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: The reason I didn't is because there is no `TypeInfo.GetProperty(string, BindingFlags)`– indeed, no BindingFlags even– on netstandard1.3. So there is no way to link the overload I want to link. If you have a suggestion of what else you'd expect to see I can put that in.

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: I guess we could. I'd be more happy about doing that if we could use C# 7 local functions because otherwise e'd have to decide things like whether to apply `| BindingFlags.DeclaredOnly` inside the method (since it's coupled to the particular loop operation) and whether to use guard conditions.

mikkelbu: Let us just keep it as it is, then we can always improve it later when we get to C# 7 (or newer).

src/NUnitFramework/framework/Internal/Reflect.cs
mikkelbu: I was thinking of something like
```
<param name="type">The type to search for the property with <paramref name="name"/>.</param>
<param name="name">The name of the property to get.</param>
<param name="bindingFlags">The specified requirements of the property.</param>
```

jnm2:
Thanks everyone and especially @mikkelbu. I'm going to go ahead and get this in then.

PR #2424, opened 2017-09-08
Fix undefined preprocessor

jnm2:
Small stuff.


CharliePoole:
@mikkelbu That's been the norm for a while. If we just leave the files out, then future devs can't see so easily that they were left out deliberately. Using an #if makes it much clearer.

jnm2:
> InvalidPlatformExpresion.cs

Yikes 😳 too late! I'll start reading at the top of the conversation next time. 😆

PR #2431, opened 2017-09-11
Fix stack trace filtering of async methods

### LINKED ISSUE #2429: Stack trace shown for Assert.Warn
@jnm2 commented on [Fri Sep 08 2017](https://github.com/nunit/nunit-console/issues/287)

I'm having a hard time reading my build logs because every Assert.Warn puts 40-50 lines of stack trace under the message. I only see this when looking at NUnit console output.

Is this by design? Is there a way to disable this?

---

@CharliePoole commented on [Fri Sep 08 2017](https://github.com/nunit/nunit-console/issues/287#issuecomment-328222728)

Yes. Make this issue a bug and somebody will fix it. 😄 

Our StackTrace filter removes certain class entries from the top of the stack. Most likely, being new, `Warn` isn't one of them. Alternatively, it may only do it for failures. Somebody should fix it.

---

@CharliePoole commented on [Sat Sep 09 2017](https://github.com/nunit/nunit-console/issues/287#issuecomment-328309089)

Actually, if you decide this should be fixed, it's a framework issue.

---

@jnm2 commented on [Sat Sep 09 2017](https://github.com/nunit/nunit-console/issues/287#issuecomment-328310341)

Thanks Charlie, will move.

CharliePoole:
Fixes #2429 

@jnm2 This fixes it for me. You might find you need to add more in your environment. As I'm leaving in the morning for a two-week family trip, feel free add yourself to the issue to take the branch and PR over if you think more should be done.

### DIFF
--- a/src/NUnitFramework/framework/Internal/StackFilter.cs
+++ b/src/NUnitFramework/framework/Internal/StackFilter.cs
@@ -35,7 +35,7 @@ namespace NUnit.Framework.Internal
     public class StackFilter
     {
         private const string DEFAULT_TOP_OF_STACK_PATTERN = @" NUnit\.Framework\.(Assert|Assume|Warn|CollectionAssert|StringAssert|FileAssert|DirectoryAssert)\.";
-        private const string DEFAULT_BOTTOM_OF_STACK_PATTERN = @" System\.R(eflection|untimeMethodHandle)\.";
+        private const string DEFAULT_BOTTOM_OF_STACK_PATTERN = @" System\.(Reflection|RuntimeMethodHandle|Threading\.ExecutionContext)\.";
 
         /// <summary>
         /// Single instance of our default filter
--- a/src/NUnitFramework/testdata/WarningFixture.cs
+++ b/src/NUnitFramework/testdata/WarningFixture.cs
@@ -22,6 +22,7 @@
 // ***********************************************************************
 
 using System;
+using System.Threading;
 using NUnit.Framework;
 using NUnit.Framework.Constraints;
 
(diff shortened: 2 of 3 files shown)


jnm2:
Thanks Charlie. I want to pull this code when I get back home later tonight because I'm on my phone and missing context. I'll probably be able to merge but if not I'll just pick up the PR.

jnm2:
There's no test that fails prior to the StackFilter class change, so I want to add that.

jnm2:
I'm leery of trimming the stack trace from another thread unless it is marked [Resume Async Method].
The reason is that you might want to know if the bottom of the stack was `System.Threading.TimerQueue.FireNextTimers()` or `System.Threading.ThreadPoolWorkQueue.Dispatch()` or any number of other triggers. If we cut it off at `System.Threading.ExecutionContext` it obfuscates that.

However, I can also see being irritated at the same boilerplate ten methods appearing after every warning. Since I'm conflicted, I'll go with the direction you started in which is hiding anything below and including `System.Threading.ExecutionContext`.

jnm2:
I discovered something that bothers me, which is that Assert.Warn on a background thread logs a warning but Assert.Fail does nothing. However, we have more pressing issues and if it matters it'll come back. =)

CharliePoole:
@jnm2 Glad you're picking this up. I can give advice via email, but it will probably be only once a day.

A simple test is just an `Assert.Warn` after a task delay. I used a temp test to TDD the fix but I didn't take the trouble to put it in the `WarningFixture` and have a test that executed it since I was in a hurry.

I see the problem with trimming that way, but it seemed like the easiest way to go.

In my original implementation of multiple assert, failures worked like warnings, with the side effect you noticed that it worked on a background thread. However, that broke people who wanted to directly execute tests outside of NUnit and the compromise was to go back to throwing an exception for failure while keeping the current behavior for warnings, which can't throw an exception anyway.

CharliePoole:
Nice tests!

jnm2:
Oh, shoot. Two of the stack traces are a frame larger in CI for net20 amd ney35 than they are on my Win10 machine. I'll start a quick PR.

jnm2:
See https://github.com/nunit/nunit/pull/2433

CharliePoole:
Looks like it got worse. 😞 

One small thing: you index `AssertionResults[0]` without verifying that there are any assertion results. That could be the cause of the one Error I see in the output. It would occur if the corresponding method in WarningFixture either threw an exception or failed. I haven't looked further.

jnm2:
Yep, fixing as we speak.

I only index AssertionResults after checking that FailureCount == 0. SHould I just replace the FailCount check with an AssertionResults.Count check?

jnm2:
Decided to add

```cs
if (result.AssertionResults.Count != 1 || result.AssertionResults[0].Status != AssertionStatus.Warning)
{
    Assert.Fail("Expected a single warning assertion. Message: " + result.Message);
}
```

CharliePoole:
That looks like a good solution. Testing `FailureCount == 0` leaves out the real possibility that an exception was thrown. Anyway, it's always best to test the exact thing you are using.

PR #2433, opened 2017-09-11
CI fixes for new warning tests

jnm2:
Quick fix. Spotted all four CI builds failing after my tests were pushed in https://github.com/nunit/nunit/pull/2431.

### DIFF
--- a/src/NUnitFramework/testdata/WarningFixture.cs
+++ b/src/NUnitFramework/testdata/WarningFixture.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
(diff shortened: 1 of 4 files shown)


jnm2:
Looks like I need to add the System.Theading.Thread package to the netstandard1.3 build. No idea why it works on my machine and not CI.

jnm2:
Oops, I canceled the wrong AppVeyor build.

jnm2:
Looks like BeginInvoke silently fails to run in CI if you provide a null callback.
Looks like it started working now. :+1:

jnm2:
Blegh, still not firing on Travis. I think I know what I did wrong; the wait handle should fire directly after the warn and not in the callback.

I'll try that and if that fails I'll just drop the BeginInvoke test.

jnm2:
At this point I observe that warning inside a BeginInvoke does not seem to do anything under Mono. @ChrisMaddock I don't know if the combination of Mono and threading means anything to you, but I'm going to assume this doesn't point to a deeper bug that we care about and just exclude the platform Mono for that test.

jnm2:
Yay for incrementally revealed failures.

jnm2:
Travis/Linux is happy! I am going to wait for Travis/OS X though...

jnm2:
@nunit/framework-team Every CI job completed successfully! 🎉 Please merge to fix our master builds.

ChrisMaddock:
I just had to remove system.threading.thread to get things to work for andoird/iOS. The package doesn't exist for those platforms (it pretends to support it.)

ChrisMaddock:
(Sorry - that post wasn't complete. Using an app on holiday, and it's not very good!). The netstandard1.3 build is pretty much only around to support the mobile platforms - and system.thread.threading doesn't work for all of them. (Maybe it was uwp which doesn't support it - I can't remember!) For 1.3, can we remove the relevant code, rather than adding the dependency? Sorry if that's not possible - can't really look at code right now!

ChrisMaddock:
One last note - system.threading.thread states it supports nestandard 1.3, but if you unzip the package, you'll see otherwise. I did a PR for 3.8 which gives a lot more detail on this. :-)

jnm2:
@ChrisMaddock Thanks for the info! I did not add System.Threading.Thread or System.Threading.ThreadPool to the framework project, only the `testdata` project. Do I still need to remove them?

jnm2:
Looking at your PR and reading the above, it makes sense to me that `nunit.framework` projects should not reference those packages but `nunit.framework-testdata` projects can. But if that's not correct, it's no problem to remove them.

ChrisMaddock:
Ah ok. Sorry - I hadn't actually loaded the code! Yes, I think that makes sense - the test you've written I don't think would be possible to implement for uwp, but that's not to say someone couldn't be doing such a thing on a netstandard1.3 platform where it IS supported. :-)

jnm2:
Okay, got it. If we did run tests on UWP, we'd exclude that platform for those tests cases then.

Issue #2019, opened 2017-01-31
TestFixtureSourceAttribute doesn't work with items implementing ITestFixtureData


dajuris:
Consider the following test:

```
namespace MyTestProject
{
    [TestFixtureSource(typeof(SourceProvider), "FixtureArgs")]
    public class MyTestClass
    {
        public MyTestClass(string word, int num)
        {
        }

        [Test]
        public void Test()
        {
        }
    }

    public class SourceProvider
    {
        static object[] FixtureArgs =
            {
                new MyTestFixtureData("Question", 1),
                new MyTestFixtureData("Answer", 42),
            };
    }

    public class MyTestFixtureData : ITestFixtureData
    {
        ...
    }
}
```

Console runner fails with `System.InvalidCastException`:

```
NUnit Console Runner 3.6.0
Copyright (C) 2017 Charlie Poole

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.10586.0
  CLR Version: 4.0.30319.42000

Test Files
    MyTestProject.dll


Errors, Failures and Warnings

1) Invalid : MyTestProject.MyTestClass
An exception was thrown while loading the test.
System.InvalidCastException: Unable to cast object of type 'MyTestProject.MyTestFixtureData' to type 'NUnit.Framework.Internal.TestFixtureParameters'.
   at NUnit.Framework.TestFixtureSourceAttribute.<BuildFrom>d__17.MoveNext()
   at NUnit.Framework.Internal.Builders.DefaultSuiteBuilder.BuildFrom(ITypeInfo typeInfo)
```

According to [documentation](https://github.com/nunit/docs/wiki/TestFixtureSource-Attribute) for `TestFixtureSourceAttribute` this should be supported scenario.

I have used NUnit and NUnit.Console 3.6.0.

rprouse:
Yes, that is what the documentation says. It looks like we let a cast to the concrete type rather than the interface slip through a code review.

In `TestFixuteSourceAttribute.BuildFrom()`

```C#
            foreach (TestFixtureParameters parms in GetParametersFor(sourceType))
                yield return _builder.BuildFrom(typeInfo, parms);
```

CharliePoole:
I marked this for discussion temporarily, as I think it could be a documentation error.

Over the past few releases - I haven't searched to  see exactly when - I modified TestCaseSourceAttribute to require the use of TestCaseParameters type, rather than just an interface. When I added TestFixtureSourceAttribute, I set it up the same way.

[...]

PR #2439, opened 2017-09-14
Some fixes for TestFixtureSource applied to generic classes

### LINKED ISSUE #1346: NullReferenceException when [TestFixtureSource] refers to data in a generic class.
NUnit output:

```
Executing: C:\Tools\NUnit3\bin\nunit3-console.exe
NUnit Console Runner 3.0.5813 

Runtime Environment
  CLR Version: 4.0.30319.42000

Test Files
    ValveKeyValue/ValveKeyValue.Test/bin/Release/ValveKeyValue.Test.dll


Errors and Failures

1) Invalid : ValveKeyValue.Test.ArrayTestCase<TEnumerable>
   at NUnit.Framework.Internal.Builders.NUnitTestFixtureBuilder.BuildFrom(ITypeInfo typeInfo, ITestFixtureData testFixtureData)
   at NUnit.Framework.TestFixtureSourceAttribute.<BuildFrom>d__17.MoveNext()

2) Invalid : ValveKeyValue.Test.TopLevelEnumerableTestCase<TEnumerable>
   at NUnit.Framework.Internal.Builders.NUnitTestFixtureBuilder.BuildFrom(ITypeInfo typeInfo, ITestFixtureData testFixtureData)
   at NUnit.Framework.TestFixtureSourceAttribute.<BuildFrom>d__17.MoveNext()

Run Settings
    WorkDirectory: C:\projects\valvekeyvalue
    NumberOfTestWorkers: 2

Test Run Summary
    Overall result: Failed
   Tests run: 85, Passed: 85, Errors: 0, Failures: 0, Inconclusive: 0
  Start time: 2016-03-13 09:36:40Z
    End time: 2016-03-13 09:36:41Z
    Duration: 1.241 seconds
```

Full build process output is available at https://ci.appveyor.com/project/xPaw/valvekeyvalue/build/1.0.0.20-AppVeyor#L60

Seemingly, nunit3-console returns a status code of 0. I'd expect in a failure case such as this, it would return non-zero.

Code that causes this issue:

`[TestFixtureSource]` here: https://github.com/SteamDatabase/ValveKeyValue/blob/dc20dcde8d4efddfcc0ed7730fd693563401300a/ValveKeyValue/ValveKeyValue.Test/Text/ArrayTestCase.cs#L7
Points to this collection here: https://github.com/SteamDatabase/ValveKeyValue/blob/dc20dcde8d4efddfcc0ed7730fd693563401300a/ValveKeyValue/ValveKeyValue.Test/TestFixtureSources.cs#L9

StanEgo:
This is a simple solution when we pass generic type arguments directly to the TestFixture. Works fine.

```
public interface IValidEntity { }
public class EntityOne : IValidEntity { }
public class EntityTwo { }
public class EntityThree : IValidEntity { }

    [TestFixture(typeof(EntityOne))]
    [TestFixture(typeof(EntityTwo))]
    [TestFixture(typeof(EntityThree))]
    public class TestUsingSimpleCases<TEntity>
    {
        [Test]
        public void TestMethod()
        {
            Assert.That(typeof(IValidEntity).IsAssignableFrom(typeof(TEntity)));
        }
    }
```

The same, but we fetch the types array from the "Source" field with array of types.

```
[TestFixtureSource(typeof(RegistryType), nameof(RegistryType.Source))]
public class TestUsingTypeSource<TEntity>
{
	[Test]
	public void TestMethod()
	{
		Assert.That(typeof(IValidEntity).IsAssignableFrom(typeof(TEntity)));
	}
}

public class RegistryType
{
	public static readonly Type[] Source = new[]
	{
		typeof(EntityOne),
		typeof(EntityTwo),
		typeof(EntityThree)
	};
}
```

Firstly we get error in NUnitTestFixtureBuilder.cs:100, because TypeArgs is null. For TestFixtureAttribute it's set as an empty array by default, so things are ok. But when generated from source ITextFixtureData.TypeArgs contains null.

We can also create an empty array, but then we should do it in any ITestFixtureData implementation including some created by users. So I think it is better to apply a small fix to builder itself so it will have the same behaviour not only when TypeArgs.Length == 0, but also when it is null.

But when I tried to solve this problem without NUnit fixing, have found another issue. For example I've replaced types array with ITestFixtureData array that was looking as valid solution and I was able to initialize TypeArgs with an empty array as TestFixtureAttribute does (it is also ITestFixtureData).

```
[TestFixtureSource(typeof(RegistryData), nameof(RegistryData.Source))]
public class TestUsingDataSource<TEntity>
{
	[Test]
	public void TestMethod()
	{
		Assert.That(typeof(IValidEntity).IsAssignableFrom(typeof(TEntity)));
	}
}

public class CustomData : ITestFixtureData
{
	public Type[] TypeArgs { get; }
	public string TestName { get; }
	public RunState RunState { get; }
	public object[] Arguments { get; }
	public IPropertyBag Properties { get; }

	public CustomData(Type type)
	{
		Arguments = new object[] { type };
		TypeArgs = new Type[0];
		Properties = new PropertyBag();
		RunState = RunState.Runnable;
	}
}

public class RegistryData
{
	public static readonly ITestFixtureData[] Source = new ITestFixtureData[]
	{
		new CustomData(typeof(EntityOne)),
		new CustomData(typeof(EntityTwo)),
		new CustomData(typeof(EntityThree)),
	};
}
```

But now exception happens in in TestFixtureSourceAttribute.cs:100. It takes iterator through ITestFixtureData items and send each item from it to NUnitTestFixtureBuilder.BuildFrom that also accepts ITestFixtureData. But parameter in foreach defined as TestFixtureParameters, so we get two implicit castings and the first one fails for any other implementation of ITestFixtureData rather than TestFixtureParameters. I've applied a small fix here as well.

I believe this PR also closes #1346.

PS. This PR failed to build, but for the reason inherited from master branch. I can't build it as well with the same error.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Builders/NUnitTestFixtureBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/NUnitTestFixtureBuilder.cs
@@ -97,7 +97,7 @@ namespace NUnit.Framework.Internal.Builders
             if (typeInfo.ContainsGenericParameters)
             {
                 Type[] typeArgs = testFixtureData.TypeArgs;
-                if (typeArgs.Length == 0)
+                if (typeArgs == null || typeArgs.Length == 0)
                 {
                     int cnt = 0;
                     foreach (object o in arguments)
--- a/src/NUnitFramework/testdata/TestFixtureSourceData.cs
+++ b/src/NUnitFramework/testdata/TestFixtureSourceData.cs
@@ -24,6 +24,7 @@
 using System.Collections;
 using NUnit.Framework;
 using NUnit.Framework.Internal;
+using System;
 
 namespace NUnit.TestData.TestFixtureSourceData
 {
@@ -304,6 +305,26 @@ namespace NUnit.TestData.TestFixtureSourceData
             yield return "Internet Explorer";
         }
     }
+
+    public class GenericFixtureSource
+    {
+        public static readonly Type[] Source = new Type[]
+        {
+            typeof(short),
+            typeof(int),
+            typeof(long)
+        };
+    }
+
+    [TestFixtureSource(typeof(GenericFixtureSource), "Source")]
+    public class GenericFixtureSourceWithProperArgsProvided<T>
+    {
+        [Test]
+        public void SomeTest()
+        {
+        }
+    }
+
 
     #region Source Data Classes
 
--- a/src/NUnitFramework/tests/Attributes/TestFixtureSourceTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TestFixtureSourceTests.cs
@@ -117,5 +117,14 @@ namespace NUnit.Framework.Attributes
             Assert.That(suite.Tests.Count, Is.EqualTo(1));
             Assert.That(suite.Tests[0], Is.TypeOf<TestMethod>());
         }
+
+        [Test]
+        public void CanRunGenericFixtureSourceWithProperArgsProvided()
+        {
+            TestSuite suite = TestBuilder.MakeFixture(typeof(GenericFixtureSourceWithProperArgsProvided<>));
+            Assert.That(suite.RunState, Is.EqualTo(RunState.Runnable));
+            Assert.That(suite is ParameterizedFixtureSuite);
+            Assert.That(suite.Tests.Count, Is.EqualTo(GenericFixtureSource.Source.Length));
+        }
     }
 }


rprouse:
@StanEgo our master branch is currently building and tests passing, so I believe that your build failures are related to your changes. I haven't reviewed your code yet, so I am unsure why.

ChrisMaddock:
Master was broken a few days ago with the Warning Tests. I've kicked things off again - I think you may need to rebase, if possible @StanEgo 🙂

StanEgo:
@ChrisMaddock sure, done.

src/NUnitFramework/framework/Internal/Builders/NUnitTestFixtureBuilder.cs
```diff
@@ -97,7 +97,7 @@ public TestSuite BuildFrom(ITypeInfo typeInfo, ITestFixtureData testFixtureData)
             if (typeInfo.ContainsGenericParameters)
             {
                 Type[] typeArgs = testFixtureData.TypeArgs;
-                if (typeArgs.Length == 0)
+                if (typeArgs == null || typeArgs.Length == 0)
```
ChrisMaddock: This part of the PR looks good to me. 👍

src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
```diff
@@ -111,7 +111,7 @@ public IEnumerable<TestSuite> BuildFrom(ITypeInfo typeInfo)
         {
             Type sourceType = SourceType ?? typeInfo.Type;
 
-            foreach (TestFixtureParameters parms in GetParametersFor(sourceType))
+            foreach (ITestFixtureData parms in GetParametersFor(sourceType))
```
ChrisMaddock: This change needs more discussion - see https://github.com/nunit/nunit/issues/2019, where we decided that users were required to use the TestFixtureData class, rather than the interface. Re-reading that issue, I think I would still be in favour of keeping TestFixtureData inline with TestCaseData - if the latter is harder to change. @StanEgo - did you have a particular reason for wanting this functionality, or did you just assume it should work? I imagine the code/comments could be mis-leading here.

rprouse: Good catch on this. It looks like the docs still haven't been updated for either [TestCaseData](https://github.com/nunit/docs/wiki/TestCaseData) or [TestFixtureData](https://github.com/nunit/docs/wiki/TestFixtureData) which state 

> Although any object implementing IFixtureData may be used to provide specific instance information for a parameterized fixture, NUnit provides the TestFixtureData class for this purpose.

The docs should be updated appropriately. I agree that this should match `TestCaseData`. I will update the documentation now.

rprouse: The documentation has been updated. Please revert this change and then I think we are good to go.

CharliePoole: @rprouse Strictly speaking, we decided to require `TestFixtureParameters` and `TestCaseParameters`. The corresponding `TestXxxxData` classes are the same with the addition of a fluent interface.

StanEgo: I see this as a bug in any case. If you want to allow TestFixtureParameters only, then this should  be controlled in builder, throwing proper exception and having relevant tests. Right now pretty strange class is responsible with very strange checking algorithm. Between GetParametersFor method that returns IEnumerable&lt;ITestFixtureData&gt; and NUnitTestFixtureBuilder.BuildFrom method that accepts ITestFixtureData as well, we insert redundant double casting that throws absolutely unclear InvalidCaseException for the user. Also no tests if this is an expectable behavior.

On the other side I believe that current ITestFixtureData-based design is absolutely SOLID. We have an SRP-interface to access test arguments and implementations those can fetch these arguments from any type of input. I don’t think that having only one implementation like TestFixtureParameters that can accepts object[] or another ITestFixtureData is a scalable solution. (btw, let's then also remove the constructor with ITestFixtureData, because it doesn't make sense to create one TestFixtureParameters from another).

In my initial message there is a sample, how this scalability was able to solve problem in existing code. This point of extension can be also used for helpers (when some complex object converted into array of args), args reusing (e.g. combining existing args and other set-of-args operations), conditions (depends on runtime environment), dispatching, proxying, multiplexing, there are millions of different cases. E.g. right now I have in one project solution like:

```
[Feature(typeof(IDeleteCommand<>))]
public class DeleteNonExistingSpecification<TEntity> {
	public override void Then(Exception exception) => Assert.That(
		exception,
		Is.InstanceOf<EntityNotFoundException>()
	);
}
```

That applies one specification to all IDeleteCommand<> implementations that registered in IoC. And FeatureAttribute works like a FixtureSource fetching types from IoC, but I had to write it from scratch, because currenly it is not extensible.

But I don't think that this is a question of trying to forecast all possible usages. The question is, why having a good design we don’t use it?

Explanation in #2019 sounds for me like: "If there is no easy fix visible, let's suppose that there is no problem" :smiley:

But probably this fix should be extrapolated to TestCaseParameters too, that looks not so easy, but possible.

CharliePoole: @StanEgo I think what's missing in the discussion among the developers and therefore invisible to you, is that we previously decided to make `TestFixtureParameters` part of the interface. As you point out, the decision was not implemented properly.

There's nothing wrong with an interface that specifies a particular data class that must be returned and that was the intention. Obviously, the existing `GetParametersFor` doesn't carry out the decision correctly.

I think the team has to make the whole thing consistent in some way... either going back to accepting the interface without error or moving ahead to using the data class in the interface.

ChrisMaddock:
@StanEgo - It would be great to get the generic fixtures bug sorted. Would you be willing to extract out the ITestFixtureData, so we can just merge the un-controversial bit, for now?

Regarding ITestFixtureData, my thought process is...
1. We should keep TestFixtureData and TestCaseData with the same behaviour. This PR changes that for one but not the other - I don't think it would be a good idea to make the two differ.
2. I absolutely agree the current code around this should be made more consistent, one way or the other.
3. My impression of #2019 was that nobody had yet given a solid use case for expanding this to take an interface, rather than specific class. You seem to have one, which sounds sensible to me! 😄 On that basis, I think it would be good to update _both_ testfixturedata and testcasedata to work with the interfaces - but they should be done in sync, and that's probably big enough for a separate PR. (Interested? :-p) This should also really be posted as an issue first, to give other @nunit/framework-team member's the chance to disagree with me!

jnm2:
@ChrisMaddock I agree with each point.

StanEgo:
@ChrisMaddock, ageed that these are independendent fixes, so removed ITestFixtureData from this PR and rebased it on the actual master. For ITestFixtureData there will a more interesting solution 😃

ChrisMaddock:
Great, thanks @StanEgo! 

This looks good to me, but I'd like it to have a proper second review, given we've had a few API problems in this area.

src/NUnitFramework/framework/Internal/Builders/NUnitTestFixtureBuilder.cs
jnm2: Everything else in the PR is tests, which also look good to me so we're good! 😃

ChrisMaddock:
Thanks @StanEgo, merging!

ChrisMaddock:
We should make an issue for the 'other bit'! @StanEgo, would you mind doing that, as I feel you're in a good position to state the case?

StanEgo:
@ChrisMaddock sure, I'll do it. Just working on the new formula of this issue, because it might be more interesting that was initially. E.g. the reason we can't use ITestFixtureData was trying to access OriginalArguments, but this arguments may be the part of adjustment interface in #2104, so there is a change the design would be more SOLID. But this requires some additional investigation.

ChrisMaddock:
Sounds promising! I'm not too sure how that will work in terms of backwards compatibility, but will leave it with you until there's a more definite plan! Thanks! 😄

Issue #2428, opened 2017-09-08
Fix NullReferenceExceptions caused by WorkItemQueue not being thread-safe


ChrisMaddock:
NUnit 3.8 started throwing various null reference exceptions e.g.
```
 Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
    at NUnit.Framework.Internal.Execution.WorkItemQueue.Dequeue()
    at NUnit.Framework.Internal.Execution.TestWorker.TestWorkerThreadProc()
    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
    at **System.Threading.ThreadHelper.ThreadStart()**
```
```
at NUnit.Framework.Internal.Execution.WorkItemQueue.get_IsEmpty()
at NUnit.Framework.Internal.Execution.WorkShift.get_HasWork()
at NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.StartNextShift()
at NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.OnEndOfShift(Object sender, EventArgs ea)
at NUnit.Framework.Internal.Execution.WorkShift.<StartWorkers>b__26_1(TestWorker s, WorkItem ea)
at NUnit.Framework.Internal.Execution.TestWorker.TestWorkerThreadProc()
at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
at System.Threading.ThreadHelper.ThreadStart()
```

@garuma diagnosed the below

> This seems to come from the queue array having null in it. In turn, I believe this is due to InitializeQueues being not concurrent-safe, creating an array and initializing it directly on shared state.

[...]

Issue #2417, opened 2017-09-05
WorkItemQueue is not thread safe


garuma:
General note: there are various things in this class that makes me itchy, it would really be better if this used something like `BlockingCollection` instead of rolling its own CAS logic (importing it in the source tree if necessary for earlier .NET). I pushed a possible re-implementation at: https://github.com/garuma/nunit/commit/da63106d8119ab4b32e48fbe0ed63ad88d3ed9a3

On the core of the issue though, with NUnit 3.8 running our test suite started throwing NREs:

```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
    at NUnit.Framework.Internal.Execution.WorkItemQueue.Dequeue()
    at NUnit.Framework.Internal.Execution.TestWorker.TestWorkerThreadProc()
    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
    at **System.Threading.ThreadHelper.ThreadStart()**
```

This seems to come from the queue array having null in it. In turn, I believe this is due to `InitializeQueues` being not concurrency safe, creating an array and initializing it directly on shared state.

Looking at the commit I referenced above, using my slightly different `CreateQueues` way makes it concurrent-safe.

CharliePoole:
Yes, I believe you're right. `InitializeQueues` was not written to be thread-safe - and originally didn't need to be. But when we introduced the `Save` and `Restore` operations, it began to be called on worker threads and should have been made thread-safe.

I would hesitate to introduce a complete re-implementation - at least not without a great deal of review. Could you give a brief summary of why you find`BlockingCollection` more suitable than `ConcurrentQueue`?

garuma:
Sure, the approach you have currently (monotonic dual integers manipulated via CAS) actually used to be how `BlockingCollection` worked on Mono (now that we use referencesource it's slightly different).

[...]

PR #2442, opened 2017-09-16
Min-fix to make WorkItemQueue.Save() threadsafe

ChrisMaddock:
Fixes https://github.com/nunit/nunit/issues/2428

This is intended as the minimal necessary fix to make this method thread-safe, and prevent the exceptions which are currently being thrown. @garuma has proposed a more wide-scale re-write in https://github.com/nunit/nunit/issues/2417. This commit is intended for a hotfix, meaning we can fix the issue at hand with minimal changes, and look at a better solution for the next full release.

I'm believe this is all that's strictly necessary - @garuma, @CharliePoole - would you both mind checking you agree?


ChrisMaddock:
I'm seeing the exceptions again, this hasn't fixed the issue.

PR #2451, opened 2017-09-21
Compile RegEx to improve performance

rprouse:
@OsirisTerje and I have been working with the Visual Studio test team looking at performance. Our XML serialization in TNode shows up as a bottleneck for performance which isn't really a surprise. One area in there that showed up fairly high was our use of `RegEx` to escape invalid XML characters. We create this `RegEx` as a static and use it many times. Adding the compiled option to the `RegEx` produced a small performance gain, so I did the same for our two other long lasting `RegEx`'s.

This was tested using a test project from the VSTest team with 10,000 empty test methods. Running those tests using the console, in-process, compiled release, took an average of 10.65 sec over 4 runs. With this fix, the average for 4 runs drops to 9.65 sec.

Even though this is nearly a 10% performance gain, it is only for trivial tests and the gain would be smaller if the tests were more complex. It is a minor fix though, so probably worth it.

### DIFF
--- a/src/NUnitFramework/framework/Interfaces/TNode.cs
+++ b/src/NUnitFramework/framework/Interfaces/TNode.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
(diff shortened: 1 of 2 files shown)


ChrisMaddock:
Looks good!

Out of interest, did you compare this to using something like `XmlConvert.IsXmlChar()` over our own regex? It's only post-.NET-4.0, but could be worth using on the platforms where it's available.

Issue #2418, opened 2017-09-05
Support equality comparison delegate


jnm2:
This is already possible today:

```cs
Is.EqualTo(...).Using<Foo>((x, y) => x.CompareTo(y)))
```

The delegate returns `int` which allows you to pick one: equal, greater than, or less than. It's a sorting comparison rather than an equality comparison.

I have a similar scenario with objects in a third party library that provide `IsEqual` but do not override `Equals` (:thinking:...). This should work:

```cs
Is.EqualTo(...).Using<Foo>((x, y) => x.IsEqual(y)))
```

But because it returns `bool` rather than `int`, it does not work. I am forced to arbitrarily convert equal/inequal to either 0/1 (equal/greater than) or 0/-1 (equal/less than). I don't like this because an ordering comparison does not make sense in this context, only a pure equality comparison.

```cs
Is.EqualTo(...).Using<Foo>((x, y) => x.IsEqual(y) ? 0 : 1))
```

In my opinion, I should be able to return a `bool` directly from any arbitrary delegate and NUnit should use it like it uses `bool IEqualityComparer<>.Equals` similar to the effect of this final workaround:

```cs
private sealed class FooEqualityComparer : IEqualityComparer<Foo>
{
    public static FooEqualityComparer Instance { get; } = new FooEqualityComparer();
    private FooEqualityComparer() { }

    public bool Equals(Foo x, Foo y) => x.IsEqual(y);

    // I don't even want to think about implementing the hash code on this property bag.
    public int GetHashCode(Foo obj) => 0;
}
```

jnm2:
To implement, start by creating a new test helper called `GenericEqualityComparison`, imitating [`GenericComparison.cs`](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/tests/TestUtilities/Comparers/GenericComparison.cs). We need a delegate type with the signature `(T, T) => bool`. We should probably use `Func<T, T, bool>` since the BCL does not have an `EqualityComparison<T>` delegate type. Then make the `GenericEqualityComparison` version of this test:

https://github.com/nunit/nunit/blob/e7c7826da8b064cfcc518b1035e1c270588abc79/src/NUnitFramework/tests/Constraints/EqualConstraintTests.cs#L564-L570

[...]

PR #2460, opened 2017-09-25
EqualConstraint.Using<T>() can be passed a delegate that returns bool

david-proctor:
Addresses issue #2418.

The only difference from the direction that @jnm2 kindly provided on the issue page is that there is already a EqualityAdapter.For(Func<T, T, bool>) overload that returns a PredicateEqualityAdapter. What I have here is basically the filler needed to connect that function to a new EqualConstraint.Using<T>(Func<T, T, bool>) function.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/EqualConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/EqualConstraint.cs
@@ -323,9 +323,9 @@ namespace NUnit.Framework.Constraints
         }
 
         /// <summary>
-        /// Flag the constraint to use the supplied (T, T) => bool object.
+        /// Flag the constraint to use the supplied boolean-returning delegate.
         /// </summary>
-        /// <param name="comparer">The (T, T) => bool object to use.</param>
+        /// <param name="comparer">The boolean-returning delegate to use.</param>
         /// <returns>Self.</returns>
         public EqualConstraint Using<T>(Func<T, T, bool> comparer)
         {
--- a/src/NUnitFramework/framework/Constraints/EqualConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/EqualConstraint.cs
@@ -323,6 +323,17 @@ namespace NUnit.Framework.Constraints
         }
 
         /// <summary>
+        /// Flag the constraint to use the supplied (T, T) => bool object.
+        /// </summary>
+        /// <param name="comparer">The (T, T) => bool object to use.</param>
+        /// <returns>Self.</returns>
+        public EqualConstraint Using<T>(Func<T, T, bool> comparer)
+        {
+            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            return this;
+        }
+
+        /// <summary>
         /// Flag the constraint to use the supplied Comparison object.
         /// </summary>
         /// <param name="comparer">The IComparer object to use.</param>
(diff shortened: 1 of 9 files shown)


src/NUnitFramework/framework/Constraints/EqualConstraint.cs
```diff
@@ -323,6 +323,17 @@ public EqualConstraint Using<T>(IComparer<T> comparer)
         }
 
         /// <summary>
+        /// Flag the constraint to use the supplied (T, T) => bool object.
+        /// </summary>
+        /// <param name="comparer">The (T, T) => bool object to use.</param>
```
jnm2: `(T, T) => bool` was my pseudo-C# shorthand, perhaps influenced by [Tyler's proposal](https://github.com/dotnet/csharplang/issues/269), but I wonder if it wouldn't be more in line with the other overloads' XML documentation to be more language-agnostic and say `boolean-returning delegate` instead of `(T, T) => bool object`?

(2 places)

src/NUnitFramework/tests/TestUtilities/Comparers/GenericEqualityComparison.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace NUnit.TestUtilities.Comparers
+{
+    internal class GenericEqualityComparison<T>
+    {
+        public bool WasCalled = false;
+
+        public Func<T, T, bool> Delegate
+        {
+            get { return new Func<T, T, bool>(Compare); }
+        }
```
jnm2: Totally up to your personal style, but to point out an optional improvement in case you weren't aware:
 `new Func<T, T, bool>` can be implicit here, and since we now allow C# 6, this generates the exact same output:
`public Func<T, T, bool> Delegate => Compare;`. No need to change though.

CharliePoole: I agree with the last sentence. 😄 I like the explicitness of it when I'm reading the code.

src/NUnitFramework/tests/Constraints/EqualConstraintTests.cs
```diff
@@ -570,6 +570,14 @@ public void UsesProvidedGenericComparison()
             }
 
             [Test]
+            public void UsesProvidedGenericEqualityComparison()
+            {
+                var comparer = new GenericEqualityComparison<int>();
+                Assert.That(2 + 2, Is.EqualTo(4).Using<int>(comparer.Delegate));
+                Assert.That(comparer.WasCalled, "Comparer was not called");
+            }
```
CharliePoole: Can we add a test that uses a lambda expression as envisioned by the original issue?

src/NUnitFramework/tests/TestUtilities/Comparers/GenericEqualityComparison.cs
david-proctor: I actually hadn't thought of the C# 6 expression-bodied member, thanks for the note. I think I'll leave it as is, though, since I also find it nice and explicit this way.

jnm2:
@david-proctor Your contribution is appreciated!

david-proctor:
No problem! Should I close issue #2418 myself or allow someone else to do so?

CharliePoole:
@david-proctor I closed the issue. You should normally put a reference to the issue in your PR description, using a format that GitHub will recognize. For example "Fixes #2418". The reference you used created a link but didn't cause it to auto-close. I should have checked for that myself when I merged your code.

Issue #2395, opened 2017-08-29
NUnit 3.8+ does not finish running tests


DavidZidar:
There seems to be some kind of deadlock going on since NUnit 3.8.0 (I have tried 3.8.1 as well). My tests run fine but when all tests are done NUnit never completes the test run and one CPU remains maxed out until the process is killed.

I am experiencing the same problem with both the console runner and the runner in CodeRush for Roslyn.

I attached a debugger to nunit-agent.exe and this is how my threads look like while this is happening, thread 6488 in my screenshot is the one using all the CPU.

![image](https://user-images.githubusercontent.com/381720/29811559-456af5e0-8ca4-11e7-9ee5-ebfbec7a34f7.png)

ChrisMaddock:
@DavidZidar Can you add details of how you're parallelising your tests? By fixtures/test-cases? Do you have a mix of parallel and non-parallel items?

This sounds like it could be related to https://github.com/nunit/nunit/issues/2261, which was in 3.8.

jnm2:
Ah... I think I saw this yesterday running tests via ReSharper.

DavidZidar:
I'm using `[assembly: Parallelizable(ParallelScope.Fixtures)]` in combination with some tests that can not be run in parallell that use the new `[NonParallelizable]` attribute.

I tried running the console runner with --workers=1 but it made no difference.

ChrisMaddock:
@DavidZidar 

The mix of parallel fixtures and nonparallel items sounds key. Would be really helpful if you could experiment removing a few - and see if you can find a minimum reproducible example.

DavidZidar:
I found one offending nonparallizable test fixture that for some reason had 2x `[SetUp]` methods and one `[TearDown]`, it also had a `[OneTimeSetUp]` and a `[OneTimeTearDown]`. When I deleted this class the CodeRush test runner seems to finish every time but the console runner still keeps freezing most of the time although now it has run to completion at least twice.

Unfortunately I am unable to reproduce this problem in a new project and this project is fairly large and I can't share any of the code.

Also, now that I am running my tests over and over again some other tests marked with `[NonParallelizable]` fail randomly every once in a while like they are being run in parallell even though they are not supposed to be.

DavidZidar:
[...]

Issue #2388, opened 2017-08-28
Parallelization causes test cases to stop respecting fixture's apartment state


jnm2:
Affected versions: 3.7.0–3.8.0

Fails:

```c#
[assembly: Parallelizable(ParallelScope.Children)]

[Apartment(ApartmentState.STA)]
public class StaFixture
{
    [TestCase(1)]
    [TestCase(2)]
    public void Test_cases_should_inherit_apartment_from_fixture(int n)
    {
        Assert.That(Thread.CurrentThread.GetApartmentState(), Is.EqualTo(ApartmentState.STA));
    }
}
```

Also fails if `[Parallelizable(ParallelScope.Children)]` is specified on the fixture or the method.

Does not fail if nothing is parallelizable.

Does not fail if the apartment is specified on the method:
```c#
[assembly: Parallelizable(ParallelScope.Children)]

public class StaFixture
{
    [TestCase(1)]
    [TestCase(2)]
    [Apartment(ApartmentState.STA)]
    public void Test_cases_should_inherit_apartment_from_fixture(int n)
    {
        Assert.That(Thread.CurrentThread.GetApartmentState(), Is.EqualTo(ApartmentState.STA));
    }
}
```

Does not fail for non-parameterized test methods.

jnm2:
Affected versions: 3.7.0–3.8.0

jnm2:
@CharliePoole, you're the most knowledgeable here. Are you in a position to take this on?

CharliePoole:
It's pretty obvious where it is happening. Essentially, apartment state is similar to parallelism in that it is determined by both the test itself and the containing tests from which it "inherits"* the behavior.

I'm tied up for the moment but I put my name on it to get to in a week or so.

========

*I put "inheritance" in quotes because it has nothing to do with either test object or attribute inheritance. One of those overloaded terms! 😸

CharliePoole:
I'm going away on a family trip for a couple of weeks so I'm taking my name off this to free it up for somebody else to work on.

CharliePoole:
I'll add this to the PR I'm working on.

CharliePoole:
Right, that's why I threw it back in the pond. It's not exactly a parallelization error, or not in the same way as the other errors are due to parallelization. It's simply that the ParallelWorkItemDispatcher, when dispatching the suite that wraps the individual cases looses the information that everything should be run in parallel. I think there's another issue that relates to this, which someone is working on. Can't find it at the moment though.

Issue #2426, opened 2017-09-08
Regression in 3.8.1: ApartmentAttribute no longer works when applied to an assembly


thecontrarycat:
Create a test project containing the following test:

```
[assembly:Apartment(ApartmentState.STA)]
namespace TestProject
{
    [TestFixture]
    public class TestFixture
    {
        [Test]
        public void Test()
        {
            if (Thread.CurrentThread.GetApartmentState() != ApartmentState.STA)
            {
                throw new ThreadStateException("The current threads apartment state is not STA");
            }
        }
    }
}
```

Compile it against NUnit 3.7.1 - Test Passes
Compile it against NUnit 3.8.1 - Test Fails

Compile it against NUnit 3.8.1 but with the `[Apartment]` attribute on the TestFixture - Test Passes
Compile it against NUnit 3.8.1 but with the `[Apartment]` attribute on the Test - Test Passes

Note that this bug is present in 3.8.0 as well as 3.8.1

CharliePoole:
I'm suspicious that this could be a side effect of the fix to #2361

CharliePoole:
This is different from but probably related to #2388

CharliePoole:
I'm going away on a family trip for a couple of weeks so I'm taking my name off this to free it up for somebody else to work on.

rinubaby:
We've already upgraded to 3.8.1 (was on 2.6.2 earlier), since we have a large number of tests, we are being blocked by this issue, any timeline as to when the next release (fix) will be available? Also, seeing occasionally that test runs are getting hung. Is this only a 3.8.1 problem, wondering if going back to 3.7.1 would fix the hanging issue as well. Thanks.

ChrisMaddock:
Both these issues were regressions in 3.8.x - so 3.7.1 should work for you until the next release.

I don't believe anyone's currently working on this issue - so as to a fix it really depends when someone picks it up and submits a pull request.

rinubaby:
I have a fix for this issue, can I do a pull request so that you can look at it? I am getting access denied.

CharliePoole:
Anyone can do a pull request. Did you fork to your own account first?

rinubaby:
Thanks Charlie, have opened a pull request.

rprouse:
I tested with 3.9.0-dev-04639 and this has not been fixed with our recent parallel fixes.

This is possibly a duplicate of #2388

PR #2461, opened 2017-09-27
Ensure queues are not started when they are supposed to be paused or stopped

### LINKED ISSUE #2454: SetUpFixture not respecting NonParallelizable tag on TestFixtures.
It seems since version 3.8.0, the inclusion of a SetUpFixture causes issues when trying to run multiple fixtures marked as parallelizable but some fixtures marked as non-parallelizable.

I have parallelism set at the assembly level with the following in my AssemblyInfo.cs:
`[assembly: Parallelizable(ParallelScope.Fixtures)]`

Here's my test setup:
```
namespace Tests
{
	[SetUpFixture]
	public class TestsSetUpFixture
	{

	}

	[TestFixture]
	public class TestFixture1
	{
		[Test]
		public void TestFixture1_Test()
		{
			Thread.Sleep(1000);
		}
	}

	[TestFixture]
	[NonParallelizable]
	public class TestFixture2
	{
		[Test]
		public void TestFixture2_Test()
		{
			Thread.Sleep(1000);
		}
	}

	[TestFixture]
	public class TestFixture3
	{
		[Test]
		public void TestFixture3_Test()
		{
			Thread.Sleep(1000);
		}
	}
}
```

All three test fixtures get run in parallel, I would expect 1 and 3 to run in parallel whilst 2 is run separately. This is the behaviour when running under version 3.7.1, however upgrading to 3.8 causes them all to run in parallel (the same is also true under 3.8.1).

We get the same problem if you remove the parallelism setting from AssemblyInfo and instead mark TestFixture1 and TestFixture3 as [Parallelizable(ParallelScope.Self)].

As work around, for now,  you can change the ParallelScope in AssemblyInfo to Children instead of Fixtures and the behaviour returns to what is expected where all TestFixture 2 waits for the other 2 fixtures to finish executing. If you have more than 1 test in a fixture  you do end up with tests within fixtures running in parallel; this can be further fixed by adding the ParallelScope.Self above the fixtures which need to run in parallel.

CharliePoole:
Fixes #2454 and perhaps some others. We should check after merging.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -242,6 +242,11 @@ namespace NUnit.Framework.Internal.Execution
 
         private void OnEndOfShift(object sender, EventArgs ea)
         {
+            // This shift ended, so see if there is work in any other
+            // TODO: Temp fix - revisit later when object model is redesigned.
+            if (StartNextShift())
+                return;
+
             if (_isolationLevel > 0)
                 RestoreQueues();
 
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -25,6 +25,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 
 namespace NUnit.Framework.Internal.Execution
@@ -222,7 +223,8 @@ namespace NUnit.Framework.Internal.Execution
         /// </summary>
         private void RestoreQueues()
         {
-            Guard.OperationValid(_isolationLevel > 0, $"Internal Error: Called {nameof(RestoreQueues)} with no saved queues!");
+            Guard.OperationValid(_isolationLevel > 0, $"Internal Error: Called {nameof(RestoreQueues)} with no saved queues.");
+            Guard.OperationValid(Queues.All(q => q.IsEmpty), $"Internal Error: Called {nameof(RestoreQueues)} with non-empty queues.");
 
             // Keep lock until we can remove for both methods
             lock (_queueLock)
--- a/src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
@@ -322,13 +322,16 @@ namespace NUnit.Framework.Internal.Execution
         /// </summary>
         internal void Save()
         {
-            Pause();
+            bool isRunning = State == WorkItemQueueState.Running;
+            if (isRunning)
+                Pause();
 
             _savedState.Push(new SavedState(this));
 
             InitializeQueues();
 
-            Start();
+            if (isRunning)
+                Start();
         }
 
         /// <summary>


src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
```diff
@@ -242,6 +242,11 @@ private void RestoreQueues()
 
         private void OnEndOfShift(object sender, EventArgs ea)
         {
+            // This shift ended, so see if there is work in any other
+            // TODO: Temp fix - revisit later when object model is redesigned.
+            if (StartNextShift())
+                return;
+
```
CharliePoole: This change is an "extra" not related to the issue I'm fixing. But it's an obvious error. Without it, we would restore all the queues, possibly losing work items not yet started, when the shift starts. For example, if a NP fixture has parallel tests, they could be waiting in the parallel queue to be run when the non-p shift ends!

More work is needed here - I just fixed the obvious.

src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
```diff
@@ -322,13 +322,16 @@ public void Pause()
...
+            if (isRunning)
+                Pause();
 
             _savedState.Push(new SavedState(this));
 
             InitializeQueues();
 
-            Start();
+            if (isRunning)
+                Start();
```
CharliePoole: This fixes the cause of the problem. When saving the state of queues, we started all the new queues running, not just those that were already running under the control of a `WorkShift`.

rprouse:
Looks good so far. I am just finishing for the night, so will pull in the morning and run through some tests before final approval 👍

ChrisMaddock:
Could I suggest adding the guard added in #2445 to `RestoreQueues()` - to get additional protection while we're in this intermediate state before refactoring? I think the fix in general is good - but if we've missed any other edge cases, this may make diagnosis easier.

https://github.com/nunit/nunit/pull/2445/files#diff-5c2e53665a6b5117adeb8d5d162e4726R227

rprouse:
I created a test project based on the tests that I was using for #2445 and they all run to completion now. I am going to add in the tests from #2454 and check those, but I think this is good to go although I agree the guard may be helpful. I will leave that to you to add and merge when ready. Once this is merged, I will continue to add to my test project to see what other related issues are fixed.

If anyone wants to play with it, the project is at https://github.com/rprouse/NUnit.Parallelism.Integration.Tests

rprouse:
I have verified this fixes #2454 and #2438 👍 

I think it will also fix #2395

rprouse:
This **does not** fix #2388.

Testing #2426, I now get a `NullReferenceException` on 

```
nunit.framework.dll!NUnit.Framework.Internal.Execution.WorkItemQueue.IsEmpty.get()
nunit.framework.dll!NUnit.Framework.Internal.Execution.WorkShift.HasWork.get()
nunit.framework.dll!NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.StartNextShift()
nunit.framework.dll!NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.OnEndOfShift(object sender, System.EventArgs ea)
nunit.framework.dll!NUnit.Framework.Internal.Execution.WorkShift.StartWorkers.AnonymousMethod__26_1(NUnit.Framework.Internal.Execution.TestWorker s, NUnit.Framework.Internal.Execution.WorkItem ea)
nunit.framework.dll!NUnit.Framework.Internal.Execution.TestWorker.TestWorkerThreadProc()
```

PR #2462, opened 2017-09-27
Test utilities

CharliePoole:
These are changes I made to logging and test utilities in order to debug issue #2454. There are no functional changes as far as users are concerned.

I think it's all an improvement and some of it helped me figure out a solution to the above issue. I'm making it a separate PR so that the main one for #2454 can get merged quickly. However, it would be nice to get this one in as I'm building on it locally. 😄

### DIFF
--- a/src/NUnitFramework/FakeTestMethod.cs
+++ b/src/NUnitFramework/FakeTestMethod.cs
@@ -1,0 +1,49 @@
+﻿// ***********************************************************************
+// Copyright (c) 2014 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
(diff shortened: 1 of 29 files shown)


src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
```diff
@@ -242,6 +242,11 @@ private void RestoreQueues()
 
         private void OnEndOfShift(object sender, EventArgs ea)
         {
+            if (StartNextShift())
```
ChrisMaddock: This is a non-trivial change, should it be in this PR?

src/NUnitFramework/FakeTestMethod.cs
```diff
@@ -0,0 +1,49 @@
+﻿// ***********************************************************************
+// Copyright (c) 2014 Charlie Poole, Rob Prouse
```
ChrisMaddock: Trivial - but wrong date here. 🙂

CharliePoole:
@ChrisMaddock Your comment is understandable as I separated the changes from the things that motivated it, which were:

1. Issue #2454 - these are all changes I created in working on that issue in order to debug it. The fix is in a separate PR, of course.

2. Unit testing of parallelism - part of but going beyond #2454, I wanted to be able to run tests that examine how parallelism works without creating a separate assembly. I have not included any of those tests yet as it's a work in progress.

CharliePoole:
On the placement of the classes: TestBuilder and the Fake methods are in the TestUtilities namespace, of course, but they are linked from the common parent directory NUnitFramework because they are used in both the framework and the nunitlite tests. Our past practice/convention has been that anything in that directory is shared between projects - for example the CommonAssemlblyInfo file. What you suggest is, of course, an alternative convention that we haven't been using lately but did use in the past.

src/NUnitFramework/FakeTestMethod.cs
CharliePoole: That's the year the code was created in the old `Fakes.cs` file. Splitting it into a separate file without change didn't seem like a reason to update the year.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
CharliePoole: I didn't intend it to be here - messed up separating my changes.

ChrisMaddock:
Apologies - I was aware of the issues this PR related to, what I wanted to understand was the reason behind this refactor. e.g. `Fakes.GetTestMethod()` vs `new FakeTestMethod()`. I'm sure there is a good reason - I just wasn't sure what it was from just looking at the code! 😄 

> Our past practice/convention has been that anything in that directory is shared between projects - for example the CommonAssemlblyInfo file. What you suggest is, of course, an alternative convention that we haven't been using lately but did use in the past.

That would make more sense to me - but I don't have significantly strong opinion to push on it! 🙂

CharliePoole:
It's definitely a change we could make (in file locations). Doing it as we do has the advantage that the path from the project directory is simply `..`

A possibility is to put a `TestUtilities` directory under `NUnitFramework`.

Regarding the motivation for changing, I think I'll add one sample test that may illustrate why I liked the use of `new` better than static constructors.

CharliePoole:
I'm closing this without merging. The remaining changes will go into actual fixes of parallelism, rather than in a separate PR for the utilities.

Issue #2102, opened 2017-03-29
NUnitLite incorrectly reports Win 10 OS name


ChrisMaddock:
NUnitLite incorrectly reports `Environment.OSVersion` as Windows 8, when running on Windows 10. The property has been [deprecated](https://msdn.microsoft.com/en-us/library/system.environment.osversion(v=vs.110).aspx). This is shown in the main console output, xml, and a couple of other places.

The simple workaround is to add an [app.manifest](https://github.com/nunit/nunit-console/blob/8cb7226c0ad683fe5eb7a4eee8989aff08dc4ccb/src/NUnitConsole/nunit3-console/app.manifest), as the console runner has. (The Console runner reports correctly.)

Alternatively, looks like the existing `OSPlatform` class might already work around this situation.

CharliePoole:
In the console output, do you mean under RunTime Environment? Unless you are going to take this and do a PR, it would help to be more specific for all the places.

I'm not sure we can add an app manifest, since the user writes the main exe when running NUnitLite. If it looks like OSPlatform would help, that might be best.

ChrisMaddock:
You're right about the app.manifest...I didn't really think that through, did I...

Regarding where it's used, I just did a grep for `Environment.OSVersion` 😄 But the place I noticed it first was the main console output (TextUi.cs) - although it also seems to be used in various other places.

Looks like these should possibly use OSPlatform then.

rprouse:
I wonder if the behavior isn't correct since it is reporting that you are running in Win 8 compatibility mode because you don't have a proper manifest?

ChrisMaddock:
That may well be the reason behind it, yes. 😄 I don't have much of an idea how these things work.

The note from MSDN is:

> Starting with Windows 8, the OSVersion property returns the same major and minor version numbers for all Windows platforms. Therefore, we do not recommend that you retrieve the value of this property to determine the operating system version.

This property doesn't seem like the best way to be getting this info anymore.

Kintar:
@ChrisMaddock : Can you give steps to reproduce this?  Everything that I'm seeing looks like it reports the correct OS version on Windows 10.

ChrisMaddock:
@Kintar - Are you definitely looking at the NUnitLite runner, as opposed to the nunit3-console?

I suggest you download the .zip release from here. https://github.com/nunit/nunit/releases. Then navigate to `/bin/net-4.5` and run `nunitlite.tests.exe`. You should see this - which highlights the problem. 🙂 

[...]

PR #2468, opened 2017-10-02
Bugfix/issue2102

Kintar:
Corrects issue #2102 by implementing a ToString() in OSPlatform and substituting that for the existing Environment.OSVersion.ToString() used in TextUI.

Note that there are still two uses of Environment.OSVersion that could report the compatibility mode version instead of the true OS version.  One is in the FrameworkController at line 394, and one is in NUnit2XmlOutputWriter at line 143.  I left them as they're technically out of scope for this issue.

### DIFF
--- a/src/NUnitFramework/framework/Internal/OSPlatform.cs
+++ b/src/NUnitFramework/framework/Internal/OSPlatform.cs
@@ -229,9 +229,9 @@ namespace NUnit.Framework.Internal
         }
 
         /// <summary>
-        /// Return a 
+        /// Implemented to use in place of Environment.OSVersion.ToString()
         /// </summary>
-        /// <returns></returns>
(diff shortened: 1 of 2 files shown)


CharliePoole:
Looks good to me.

src/NUnitFramework/framework/Internal/OSPlatform.cs
```diff
@@ -228,6 +229,37 @@ public PlatformID Platform
         }
 
         /// <summary>
+        /// Return a 
+        /// </summary>
+        /// <returns></returns>
+        public override string ToString()
```
mikkelbu: The summary-element seems a bit incomplete. Also I would leave out the returns-element if it is empty.

Kintar: Yeah...not sure how that happened.  Thanks!

PR #2470, opened 2017-10-03
Return also protected and internal properties

### LINKED ISSUE #2465: Possible wrong properties are returned by reflection in ReflectionExtensions.cs
When debugging #2448 I noticed that the filtering of properties in `ApplyBindingFlags` has the same problem as filtering of methods in #2448. The documentation says
https://msdn.microsoft.com/en-us/library/kyaxdd3x(v=vs.110).aspx#Remarks 
> Specify BindingFlags.NonPublic to include non-public properties (that is, private, internal, and protected properties) in the search. Only protected and internal properties on base classes are returned; private properties on base classes are not returned.

But the code only returns private properties for `BindingFlags.NonPublic`
https://github.com/nunit/nunit/blob/6980e958df0a94a4969258f8ed35aaccedb9ff31/src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs#L346-L349

The solution to this can probably be done similar to #2448 (replacing `p.GetMethod.IsPrivate` with `!p.GetMethod.IsPublic`), so the main issue is to write some tests that capture the intended behaviour.

mikkelbu:
As the documentation states "Specify BindingFlags.NonPublic
to include non-public properties (that is, private, internal,
and protected properties) in the search.

Fixes #2465

### DIFF
--- a/src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
+++ b/src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
@@ -346,7 +346,7 @@ namespace NUnit.Compatibility
             if (pub && !priv)
                 infos = infos.Where(p => (p.GetMethod != null && p.GetMethod.IsPublic) || (p.SetMethod != null && p.SetMethod.IsPublic));
             if (priv && !pub)
-                infos = infos.Where(p => (p.GetMethod == null || p.GetMethod.IsPrivate) && (p.SetMethod == null || p.SetMethod.IsPrivate));
+                infos = infos.Where(p => (p.GetMethod == null || !(p.GetMethod.IsPublic)) && (p.SetMethod == null || !(p.SetMethod.IsPublic)));
 
             bool stat = flags.HasFlag(BindingFlags.Static);
             bool inst = flags.HasFlag(BindingFlags.Instance);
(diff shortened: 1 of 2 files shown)


ChrisMaddock:
@mikkelbu - Have you spotted this failed the appveyor CI build? 🙂

mikkelbu:
@ChrisMaddock Nope. I made the PR and then went to bed :). The test that fails is `PropertyValue_ShadowingPropertyOfDifferentType` which was merge into master just before I made the PR, so probably I have run the tests locally before fetching the latest commits from master, and forgot to run the tests again. I'll take a look at it tonight.

src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
```diff
@@ -250,7 +250,9 @@ public static PropertyInfo GetProperty(this Type type, string name)
         /// <returns></returns>
         public static PropertyInfo GetProperty(this Type type, string name, BindingFlags flags)
         {
+            bool declaredOnly = flags.HasFlag(BindingFlags.DeclaredOnly);
             return type.GetRuntimeProperties()
+                .Where(prop => declaredOnly ? prop.DeclaringType.Equals(type) : true)
```
mikkelbu: I was considering writing this more succinctly as 
`.Where(prop => !declaredOnly || prop.DeclaringType.Equals(type))`
but chose the more verbose solution for readability (also I guess/hope that the compiler will generate the same IL).

ChrisMaddock: Good choice - I agree!

ChrisMaddock: Good choice - I agree!

PR #2473, opened 2017-10-06
Separate `RangeAttribute` from 'ValuesAttribute`

### LINKED ISSUE #2469: Allow RangeAttribute to be specified multiple times for the same argument
The RangeAttribute (e.g. [Range(0, 20)]) is a way to define the value range to test for in a theory, but it's restricted to occur only once per parameter of the test method.

Sometimes the same test code should be tested for different intervals.
Therefore it would be good to allow multiple ranges to be specified for the same parameter:

```[Theory]
public void testcode([Range(-10, 1), Range(20, 25)] value)
{
  // do something
}```

### PR DESCRIPTION
**Pull request by pknouff**
Previously, `RangeAttribute` derived from `ValuesAttribute`. However, there are certain use cases which apply to the `RangeAttribute`, but not to the `ValuesAttribute`. For example, we may want specify multiple, distinct ranges for a parameter by adding the `RangeAttribute` multiple times, but there isn't really a corresponding use case for the values attribute. We could just specify all the desired values in on attribute.

This change updates the `RangeAttribute` to no longer derive from the `ValuesAttribute`. Instead, it now derives from `DataAttribute` and implements `IParameterDataSource`. The  `IParameterDataSource.GetData` method has been implemented in `RangeAttribute` to preserve the behavior of that method in the `ValuesAttribute`. However, certain cases that could not be hit in the `RangeAttribute` (e.g. when the internal data was non-numeric) were removed.

Fixes #2469

### CODE REVIEW

File: src/NUnitFramework/framework/Attributes/RangeAttribute.cs [hunk 1]
```diff
@@ -23,15 +23,28 @@
...
+    public class RangeAttribute : DataAttribute, IParameterDataSource
     {
+        /// <summary>
+        /// The collection of data to be returned.
+        /// We use an object[] so that the individual
+        /// elements may have their type changed in GetData
+        /// if necessary
+        /// </summary>
+        // TODO: This causes a lot of boxing so we should eliminate it
+        protected object[] data;
```
CharliePoole: This can now be private.

pknouff: Alright, cool. I wasn't sure if I could do that or if I needed to leave it protected in case another attribute needed to derive from `RangeAttribute` in the future. 

I went ahead and renamed this to `_data` while I was at it to be more consistent with the coding standards. It's the only private field in this class, so it seemed like a safe time to make that change, but if it's not I can back it out.

CharliePoole:
Looks like you did a forced push. Sometimes that's actually necessary - as when you are completely redoing what you previously did - but it's not good as a normal thing because it destroys a lot of information. Better to just do another commit and let us squash if we want to when we merge.

In this case, I can't figure out what you changed in order to make the tests pass. I'm a bit curious about that. 😄 

Since you are going with the standard underscore, you should also drop the use of `this` with it.

pknouff:
Sorry about that. I did do a force push. I'll push new commits in the future. :+1:

The `CheckValues` and `CheckValuesWithinTolerance` in the `RangeAttributeTests` were trying to get a `ValuesAttribute` off of the parameters for the various test methods. That worked when `RangeAttribute` derived from it, but not after my change. This basically broke all the range attribute tests for all the frameworks, which is why there were so many failures. To fix them, all I had to do was change those methods to look for a `RangeAttribute`.

CharliePoole:
Looks good to me. Our standard practice is to have two team members review, so we'll wait for somebody else to agree. 😄 

Thanks for your work.

pknouff:
:+1: Sounds good. I'm happy to help. Thanks for your responsiveness.

src/NUnitFramework/framework/Attributes/RangeAttribute.cs
```diff
@@ -23,15 +23,25 @@
...
+using NUnit.Framework.Interfaces;
 
 namespace NUnit.Framework
 {
     /// <summary>
     /// RangeAttribute is used to supply a range of values to an
     /// individual parameter of a parameterized test.
     /// </summary>
-    public class RangeAttribute : ValuesAttribute
+    [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = true)]
```
ChrisMaddock: Most similar attributes are marked `Inherited = false`. (It's true by default.) Should this be consistent - or have I missed a design decision here? 🙂

src/NUnitFramework/framework/Attributes/RangeAttribute.cs
```diff
@@ -192,12 +202,76 @@ public RangeAttribute(float from, float to, float step)
...
 
         #endregion
+
+        /// <summary>
+        /// Get the range of values to be used as arguments
+        /// </summary>
+        public IEnumerable GetData(IParameterInfo parameter)
+        {
+            Type tartgetType = parameter.ParameterType;
+
```
ChrisMaddock: Whoops - typo here! 🙂

src/NUnitFramework/framework/Attributes/RangeAttribute.cs
```diff
@@ -192,12 +202,76 @@ public RangeAttribute(float from, float to, float step)
...
+                if (targetType == typeof(short) || targetType == typeof(byte) || targetType == typeof(sbyte))
+                {
+                    convert = arg is int;
+                }
+                else if (targetType == typeof(decimal))
+                {
+                    convert = arg is double || arg is int;
+                }
+
```
ChrisMaddock: Can you clarify why you removed TimeSpan and DateTime here? It's possible to use range attribute with both of those values...isn't it?

src/NUnitFramework/framework/Attributes/RangeAttribute.cs
CharliePoole: Good catch! AFAIK all our data attributes use `Inherited = false`

src/NUnitFramework/framework/Attributes/RangeAttribute.cs
CharliePoole: `RangeAttribute` has a fixed set of constructors for specific numeric types. Supporting TimeSpan and DateTime would be an enhancement requiring a bit of new logic. Nobody has requested it and it would be better to do separately anyway, I think.

ChrisMaddock: Ah - never mind, you're right - turns out you can't use RangeAttribute with TimeSpan or DateTime!

ChrisMaddock: We cross-posted @CharliePoole, but yes - sorry, I'd mistakenly remembered this already being a feature!

src/NUnitFramework/framework/Attributes/RangeAttribute.cs
```diff
@@ -192,12 +202,76 @@ public RangeAttribute(float from, float to, float step)
...
+        /// <summary>
+        /// Get the range of values to be used as arguments
+        /// </summary>
+        public IEnumerable GetData(IParameterInfo parameter)
+        {
+            Type tartgetType = parameter.ParameterType;
+
+            if (tartgetType.GetTypeInfo().IsEnum && _data.Length == 0)
+            {
```
ChrisMaddock: The enum and bool evaluation isn't required for RangeAttribute, is it?

CharliePoole: Correct. I approved too soon. 😄

src/NUnitFramework/framework/Attributes/RangeAttribute.cs
```diff
@@ -192,12 +202,76 @@ public RangeAttribute(float from, float to, float step)
...
+                }
+
+#if !NETSTANDARD1_3 && !NETSTANDARD1_6
+                if (arg is DBNull)
+                {
+                    _data[i] = null;
+                    continue;
+                }
+#endif
+
```
ChrisMaddock: This DBNull isn't needed for RangeAttribute either

ChrisMaddock:
@pknouff - Sorry, I've had a bit of a brain splurge on this one, and my comments haven't ended up in a very logical order!

> I'm not totally happy with the code duplication for getting the parameter values - there's quite a lot of logic here, and we have issues with out other TestCaseBuilding attributes constantly being out of sync regarding things like optional and nullable parameters.

With hindsight - I'm withdrawing this comment - I think once the unneeded logic is removed, there will probably be too little left to worry. 🙂

CharliePoole:
@ChrisMaddock I agree about adding a test. We already have tests with multiple attributes, I believe, but not two range attribugtes, since that wasn't supported.

Note that this is not a test case building attribute as you called it, but a single parameter provider. One of the test case providers has to post-process the values to produce test cases and some conversions are done there. Default is Combinatorial.

src/NUnitFramework/framework/Attributes/RangeAttribute.cs
pknouff: I'm not entirely clear about what `Inherited = false` means in combination with `AttributeTargets.Parameter`. Can you help me understand what that does and why it's used on the other attributes?

pknouff:
@ChrisMaddock Thanks for looking over this. I'll remove the unnecessary enum and bool evaluation and the `DBNull` check, and fix the typo. As I look at it more, the `arg == null` check on 239 also seems unnecessary since `RangeAttribute` can only be constructed with non-nullable types, so I'll probably take that out as well. The resulting `GetData` is a fair bit simpler, but still has some logic:

```cs
/// <summary>
/// Get the range of values to be used as arguments
/// </summary>
public IEnumerable GetData(IParameterInfo parameter)
{
    Type targetType = parameter.ParameterType;

    for (int i = 0; i < _data.Length; i++)
    {
        object arg = _data[i];

        if (targetType.GetTypeInfo().IsAssignableFrom(arg.GetType().GetTypeInfo()))
        {
            continue;
        }

        var convert = false;

        if (targetType == typeof(short) || targetType == typeof(byte) || targetType == typeof(sbyte))
        {
            convert = arg is int;
        }
        else if (targetType == typeof(decimal))
        {
            convert = arg is double || arg is int;
        }

        if (convert)
        {
            _data[i] = Convert.ChangeType(arg, targetType, System.Globalization.CultureInfo.InvariantCulture);
        }
    }

    return _data;
}
```

I'm open to DRYing this up with `ValuesAttribute` if it's still too much code duplication. Let me know if you'd like me to explore that still or not.

pknouff:
Also, I'll add tests for multiple range attributes (forgot to mention that in the last comment)

ChrisMaddock:
> I'm open to DRYing this up with ValuesAttribute if it's still too much code duplication. Let me know if you'd like me to explore that still or not.

Actually, thinking this through - if we had a static helper class to convert the types in the data array, it could just be the functionality that's in ValuesAttribute, and it wouldn't matter than not all of it's relevant for RangeAttribute, would it? That may be the best solution, I think it would be best not to duplicate this logic if possible...

Sorry for to-ing and fro-ing, these things don't always seem so clear on the first reading!

pknouff:
I extracted a helper class for the param conversion logic. I put it in the `NUnint.framework.Internal` namespace. I'm not sure if that is the best place or not, but can move it if desired. Extracting this gave me the opportunity to add some tests around that logic specifically as well. 

I'm working on the tests for the multiple range attributes right now.

pknouff:
Alright, I added tests for multiple instances of the `RangeAttribute` for all support tpyes of ranges. Let me know if I've missed any feedback, or if there are any additional things I should address.

src/NUnitFramework/framework/Attributes/RangeAttribute.cs
```diff
@@ -23,15 +23,26 @@
...
+using NUnit.Framework.Internal;
 
 namespace NUnit.Framework
 {
     /// <summary>
     /// RangeAttribute is used to supply a range of values to an
     /// individual parameter of a parameterized test.
     /// </summary>
-    public class RangeAttribute : ValuesAttribute
+    [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = true, Inherited = false)]
```
ChrisMaddock: Sorry - I missed your comment earlier. I think you're right, I'm not sure what `Inherited` does on a Parameter attribute either - I hadn't considered that this attribute was parameter only! I wouldn't worry about reverting it though, at least it's consistent now 🙂

src/NUnitFramework/framework/Internal/ParamNumberConversions.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Reflection;
+using NUnit.Compatibility;
+using NUnit.Framework.Interfaces;
+
+namespace NUnit.Framework.Internal
+{
+    /// <summary>
+    /// Helper methods for converting parameters to numeric values
+    /// </summary>
+    internal static class ParamNumberConversions
```
ChrisMaddock: Namespace is good, glad to see this is internal aswell!

src/NUnitFramework/tests/Attributes/RangeAttributeTests.cs
```diff
@@ -414,21 +463,58 @@ private void CheckValues(string methodName, params object[] expected)
...
-            var attr = param.GetCustomAttributes(typeof(ValuesAttribute), false).First() as ValuesAttribute;
+            var attr = param.GetCustomAttributes(typeof(RangeAttribute), false).First() as RangeAttribute;
 #else
-            var attr = param.GetCustomAttributes(typeof(ValuesAttribute), false)[0] as ValuesAttribute;
+            var attr = param.GetCustomAttributes(typeof(RangeAttribute), false)[0] as RangeAttribute;
 #endif
             Assert.That(attr.GetData(new ParameterWrapper(new MethodWrapper(GetType(), method), param)), Is.EqualTo(expected));
         }
 
+        private void CheckValuesWithMultipleAttributes(string methodName, params object[] expected)
```
ChrisMaddock: I'm not sure this is totally the right test - I think what you're testing here would be the same as testing two range attributes in two separate tests? Unless I'm missing something extra?

The think I was thinking more is that the test building logic could resolve two attributes on the same parameter. Take a look at `NUnit.Framework.Attributes.TheoryTests.AllValuesMayBeSuppliedByAttributes`  - and how that uses `TestBuilder.MakeParameterizedMethodSuite()` and then checks the number of tests generated. I think we would need a test like that, if we wanted to make sure there was no regression here. 🙂

src/NUnitFramework/tests/Internal/ParamNumberConversionsTests.cs
```diff
@@ -0,0 +1,73 @@
+using System;
```
ChrisMaddock: This is a great test class - thanks for adding it! Definitely shows the benefits of the change!

src/NUnitFramework/tests/Internal/ParamNumberConversionsTests.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        [Test]
+        public void TestStringToDecimal()
+        {
+            var data = new object[] { "0.1", "1.0", "-1.0" };
+            var expected = new decimal[] { 0.1M, 1.0M, -1.0M };
+
+            IEnumerable result = ParamNumberConversions.ConvertDataToNumeric(data, typeof(decimal));
+
+            Assert.That(result, Is.EquivalentTo(expected));
```
ChrisMaddock: Do these tests definitely fail in the correct way? I'm suspicious that `Assert.That(result, Is.EquivalentTo(data));` may also pass - due to NUnits non-strict type comparison in `NUnitEqualityComparer`. If so, it may be better to do something like `Assert.That(result, Is.EquivalentTo(expected).And.All.TypeOf<decimal>());`. Same applies to the DateTime test below.

ChrisMaddock: (I may be wrong here - I haven't tested this!)

src/NUnitFramework/framework/Internal/ParamNumberConversions.cs
```diff
@@ -0,0 +1,78 @@
+using System;
```
ChrisMaddock: Could you add the NUnit copyright notice to this new file, and your new test file please. 😄

src/NUnitFramework/framework/Internal/ParamNumberConversions.cs
ChrisMaddock: One concern, this class doesn't just comvert numbers now we've also included the functionality from ValuesAttribute! Perhaps `ParamAttributeTypeConversions` would be a better name, or something similar? A few of the comments in this class also specify 'numeric', when the methods convert other types too.

Sorry - I think this was the name I suggested - at the time, we were just talking about it having the functionality from RangeAttribute!

src/NUnitFramework/framework/Attributes/AttributeHierarchy.txt
```diff
@@ -7,7 +7,7 @@ NUnitAttribute
 	DataAttribute
 		ValuesAttribute               : IParameterDataSource
 			RandomAttribute           : IParameterDataSource
-			RangeAttribute
+		RangeAttribute
```
ChrisMaddock: RangeAttrivute implements `: IParameterDataSource` now - not that this file is vastly important!

CharliePoole: Funny thing... this is a file I created for my own use a long time ago and accidentally committed. I would have removed it, but by the time I discovered it other people had started to maintain it.

src/NUnitFramework/tests/Attributes/RangeAttributeTests.cs
pknouff: Ah, got it. I updated the tests to use `TestBuilder.MakeParameterizedMethodSuite`.

src/NUnitFramework/framework/Attributes/RangeAttribute.cs
pknouff: 👍 FWIW, I pulled my branch down on my work computer where I have Resharper. It told me that setting the `Inherited` property was redundant here, so it sounds like it has no effect on parameter only attributes. Like you said, it's consistent now, so I left it in. It should be pretty trivial to clean all of them up in the future if that is desired.

ChrisMaddock: Ah cool. Sorry about that!

Issue #1905, opened 2016-11-18
SetupFixture without namespace will make assembly-level Parallelable attribute useless


kfstorm:
I've tried to combine parallel testing and SetupFixture. But no luck.

I added `[assembly:Parallelizable(ParallelScope.Fixtures)]` in code. And if I put SetupFixture into root namespace, then test fixtures in this assembly will run sequentially. However, if I put SetupFixture into the same namespace as other text fixtures, test fixtures can run in parallel.

I've tried both NUnit 3.4.1 and 3.5.0. The results are the same.

Below is a simple file to repro the issue.
```csharp
using System.Threading;
using NUnit.Framework;

[assembly:Parallelizable(ParallelScope.Fixtures)]

[SetUpFixture]
public class GlobalSetup
{
    [OneTimeSetUp]
    public void Setup()
    {
        Thread.Sleep(5000);
    }
}

namespace NUnitTest
{
    [TestFixture]
    public class TestFixture1
    {
        [Test]
        public void Test1()
        {
            Thread.Sleep(5000);
        }
    }

    [TestFixture]
    public class TestFixture2
    {
        [Test]
        public void Test2()
        {
            Thread.Sleep(5000);
        }
    }
}
```

Here is the test result file.
```xml
<?xml version="1.0" encoding="utf-8"?>
<test-run id="2" name="NUnitTest.dll" fullname="GlobalSetup" testcasecount="2" result="Passed" start-time="2016-11-18 14:25:20Z" end-time="2016-11-18 14:25:35Z" duration="15.037411" total="2" passed="2" failed="0" inconclusive="0" skipped="0" asserts="0" random-seed="1862713937">
  <command-line><![CDATA[XXX\NUnit-3.5.0\bin\net-4.5\nunitlite-runner.exe  NUnitTest.dll]]></command-line>
  <filter />
  <test-suite type="SetUpFixture" id="1000" name="NUnitTest.dll" fullname="GlobalSetup" classname="GlobalSetup" runstate="Runnable" testcasecount="2" result="Passed" start-time="2016-11-18 14:25:20Z" end-time="2016-11-18 14:25:35Z" duration="15.037411" total="2" passed="2" failed="0" inconclusive="0" skipped="0" asserts="0">
[...]

PR #2476, opened 2017-10-07
Initial tests of parallel execution, shift changing, etc.

### LINKED ISSUE #1239: NUnit3 sometimes hangs if SetUpFixtures are run in parallel
If I have multiple `SetUpFixture`s in the same namespace, and I enable parallelization at the Assembly level, NUnit 3.0.1 _sometimes_ hangs after running the `[OneTimeSetUp]` methods, before running the tests.

It happens with both the console runner and the Visual Studio test adapter.  In my case it remains hung for 24 hours (at which point our automated build tool kills the process.)

While debugging this issue, I added some trace statements and confirmed that both `[OneTimeSetUp]` methods ran and completed, but no `[Test]` methods were called.

The simplified code below reproduces it pretty reliably (but not every time!)  If you comment out the `Thread.Sleep()`, the problem goes away.  The problem also goes away if you turn off parallelization, or if parallelization is enabled with only a single `[SetUpFixture]`.

``` C#
using System;
using System.Threading;
using NUnit.Framework;

[assembly: Timeout(60000)]
[assembly: Parallelizable(ParallelScope.Children)]

namespace ParallelSetUpFixtures
{
    [SetUpFixture]
    public class FooSetUp
    {
        [OneTimeSetUp]
        public void FooSetUpMethod()
        {
            Console.WriteLine("FooSetUp");
        }
    }

    // [Parallelizable(ParallelScope.None)]  // if this line is uncommented, then NUnit will not hang
    [SetUpFixture]
    public class BarSetUp
    {
        [OneTimeSetUp]
        public void BarSetUpMethod()
        {
            Thread.Sleep(500); // if this line is commented out, then NUnit will not hang
            Console.WriteLine("BarSetUp");
        }
    }

    [TestFixture]
    public class MyTests
    {
        [Test]
        public void MyTest()
        {
            Console.WriteLine("MyTest");
        }
    }
}
```

Here is the command line I'm using:

```
nunit3-console.exe parallel_set_up_fixtures-w64r-16-3.dll --framework=4.5
```

While it is hung, I see three threads with NUnit code in the call stacks.  The other threads are just worker threads waiting for work.  Here are partial stack traces for the three NUnit threads:

```
    mscorlib.dll!System.Threading.WaitHandle.InternalWaitOne    Unknown
    mscorlib.dll!System.Threading.WaitHandle.WaitOne    Unknown
>   nunit.engine.dll!NUnit.Engine.Agents.RemoteTestAgent.WaitForStop Line 109   C#
    nunit-agent.exe!NUnit.Agent.NUnitTestAgent.Main Line 133    C#

    mscorlib.dll!System.Threading.WaitHandle.InternalWaitOne    Unknown
    mscorlib.dll!System.Threading.WaitHandle.WaitOne    Unknown
>   nunit.framework.dll!NUnit.Framework.Api.NUnitTestAssemblyRunner.WaitForCompletion Line 236  C#
    nunit.framework.dll!NUnit.Framework.Api.NUnitTestAssemblyRunner.Run Line 200    C#
    nunit.framework.dll!NUnit.Framework.Api.FrameworkController.RunTests Line 208   C#
    nunit.framework.dll!NUnit.Framework.Api.FrameworkController.RunTestsAction.RunTestsAction Line 404  C#

    // the size of queue is 0
>   nunit.framework.dll!NUnit.Framework.Internal.Execution.EventQueue.Dequeue Line 274  C#
    nunit.framework.dll!NUnit.Framework.Internal.Execution.EventPump.PumpThreadProc Line 208    C#
    mscorlib.dll!System.Threading.ExecutionContext.RunInternal  Unknown

```

### LINKED ISSUE #2364: Parallelizable attribute not invalidating invalid parallel scope combinations
I was looking through the ParallelizableAttribute.cs, and the changes made with https://github.com/nunit/nunit/pull/2264. I noticed that there is the functionality to invalidate tests with invalid parallel combinations in the ApplyToTest() method, but it does not appear to be actually invalidating them. Is this intentional? 

Using 3.7.1, 
I tested and was able to reproduce the hanging found in https://github.com/nunit/nunit/issues/2261 by placing [NonParallelizable] on the test fixture, and [Parallelizable(ParallelScope.Self)] on the test method. 

I placed [Parallelizable(ParallelScope.Fixture)] on a test method with no parallelization on the fixture, and the tests ran and passed normally without any parallelization.

Instead of invalidating the setting, it could potentially be useful. What are your thoughts on allowing ParallelScope.Fixtures for methods? It would be useful in the case where you would only like to run one particular test one at a time at the fixture level, but run the rest of the tests in the fixture at ParallelScope.Self. It allows for the specific test in the fixture to be isolated without affecting other fixtures in the way ParallelScope.None does.

### LINKED ISSUE #2395: NUnit 3.8+ does not finish running tests

### LINKED ISSUE #2467: Test execution hangs when using [SetUpFixture] with NUnit 3.8.x
It hangs with console runner (3.7) and also with  R# (2017.2).

Here are the log files when using `--labels:before`:
[nunit_3_8_1.txt](https://github.com/nunit/nunit/files/1348973/nunit_3_8_1.txt)
[nunit_3_7.txt](https://github.com/nunit/nunit/files/1348974/nunit_3_7.txt)

When running in R# we see that tests from other namespace never start:
image

It seems related to `public class AssemblySetup` which has the attribute `[SetUpFixture]`.

### LINKED ISSUE #2471: Parametrized testcases not running in parallel
## Issue

I am trying to run several cases of a parametrized test in parallel. Here is the code I'm using:

```c#
class Dummy
{
    static TestCaseData Case(int i)
        => new TestCaseData(TimeSpan.FromSeconds(2)).SetName($"Case {i}");

    public static IEnumerable<TestCaseData> Cases()
        => Enumerable.Range(1, 5).Select(Case);

    [TestCaseSource(nameof(Cases)), Parallelizable]
    public void ItShouldSleep(TimeSpan t)
        => Thread.Sleep(t);
}
```

When I run this using the Visual Studio test runner, I can see the individual cases being executed one at a time, and the overall test run takes ~10 seconds. I would expect all the cases to be executed simultaneously on separate threads, and for the overall run to take ~2 seconds.

## Setup

I am using the following .NET Standard based csproj file alongside the code above:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Library</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
    <ApplicationIcon />
    <StartupObject />
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
    <PackageReference Include="NUnit" Version="3.8.1" />
    <PackageReference Include="NUnit3TestAdapter" Version="3.8.0" />
  </ItemGroup>

</Project>
```

**NuGet packages:**
 - Microsoft.NET.Test.Sdk: v15.3.0
 - Microsoft.NETCore.App: v2.0.0
 - NUnit: v3.8.1
 - NUnit3TestAdapter: v3.8.0

**Visual Studio:**
 - NUnit 3 Test Adapter: 3.8.0.0
 - Visual Studio: 15.3.5

### LINKED ISSUE #2478: Attributes on SetUpFixture are not applied
Investigating parallel execution, I discovered that `ParallelizableAttribute` on a `SetUpFixture` has no effect on the test at all. In fact, aside from `SetUpFixtureAttribute` itself, attributes on a SetUpFixture have no effect at all!

Among other things, this probably explains some of our parallelization issues.

CharliePoole:
This PR is intended to collect fixes for a number of parallelism issues, particularly around use of SetUpFixtures but also with the handling of parallel and non-parallel TestFixtures and the smooth changing of shifts within the dispatcher.

The initial commit sets up some infrastructure and adds a few unit tests of the parallel dispatcher. Those tests are known to fail intermittently, just as user runs are failing intermittently in some cases.

The issues I hope to address in this PR include #1239, ~~#1905~~, #2364, #2388, #2467 and #2471. I'll add "Fixes" notations when I think any of them is fixed.

I'd like it if @nunit/framework-team members followed along and commented.

Fixes #1239 (needs review)
Fixes #2364
Fixes #2395
Fixes #2467
Fixes #2471
Fixes #2478

### DIFF
--- a/src/NUnitFramework/TestBuilder.cs
+++ b/src/NUnitFramework/TestBuilder.cs
@@ -215,6 +215,8 @@ namespace NUnit.TestUtilities
         /// </summary>
         class SuperSimpleDispatcher : IWorkItemDispatcher
         {
+            public int LevelOfParallelism { get { return 0; } }
+
             public void Start(WorkItem topLevelWorkItem)
             {
                 topLevelWorkItem.Execute();
--- a/src/NUnitFramework/TestBuilder.cs
+++ b/src/NUnitFramework/TestBuilder.cs
@@ -46,7 +46,8 @@ namespace NUnit.TestUtilities
     /// </summary>
     public static class TestBuilder
     {
-#region Build Tests
+        #region Build Tests
+
         public static TestSuite MakeSuite(string name)
         {
             return new TestSuite(name);
@@ -90,9 +91,48 @@ namespace NUnit.TestUtilities
             return new DefaultTestCaseBuilder().BuildFrom(new MethodWrapper(type, method));
         }
 
-#endregion
+        #endregion
 
-#region Run Tests
+        #region Create WorkItems
+
+        public static WorkItem CreateWorkItem(Type type)
+        {
+            return CreateWorkItem(MakeFixture(type));
+        }
+
+        public static WorkItem CreateWorkItem(Type type, string methodName)
+        {
+            return CreateWorkItem(MakeTestFromMethod(type, methodName));
+        }
+
+        public static WorkItem CreateWorkItem(Test test)
+        {
+            var context = new TestExecutionContext();
+            context.Dispatcher = new SuperSimpleDispatcher();
+
+            return CreateWorkItem(test, context);
+        }
+
+        public static WorkItem CreateWorkItem(Test test, object testObject)
+        {
+            var context = new TestExecutionContext();
+            context.TestObject = testObject;
+            context.Dispatcher = new SuperSimpleDispatcher();
+
+            return CreateWorkItem(test, context);
+        }
+
+        public static WorkItem CreateWorkItem(Test test, TestExecutionContext context)
+        {
+            var work = WorkItemBuilder.CreateWorkItem(test, TestFilter.Empty, true);
+            work.InitializeContext(context);
+
+            return work;
+        }
+
+        #endregion
+
+        #region Run Tests
 
         public static ITestResult RunTestFixture(Type type)
         {
@@ -141,41 +181,6 @@ namespace NUnit.TestUtilities
         public static ITestResult RunTest(Test test, object testObject)
         {
             return ExecuteWorkItem(CreateWorkItem(test, testObject));
-        }
-
-        public static CompositeWorkItem CreateWorkItem(Type type)
-        {
-            return CreateWorkItem(MakeFixture(type)) as CompositeWorkItem;
-        }
-
-        public static WorkItem CreateWorkItem(Type type, string methodName)
-        {
-            return CreateWorkItem(MakeTestFromMethod(type, methodName));
-        }
-
-        public static WorkItem CreateWorkItem(Test test)
-        {
-            var context = new TestExecutionContext();
-            context.Dispatcher = new SuperSimpleDispatcher();
-
-            return CreateWorkItem(test, context);
-        }
-
-        public static WorkItem CreateWorkItem(Test test, object testObject)
-        {
-            var context = new TestExecutionContext();
-            context.TestObject = testObject;
-            context.Dispatcher = new SuperSimpleDispatcher();
-
-            return CreateWorkItem(test, context);
-        }
-
-        public static WorkItem CreateWorkItem(Test test, TestExecutionContext context)
-        {
-            var work = WorkItemBuilder.CreateWorkItem(test, TestFilter.Empty, true);
-            work.InitializeContext(context);
-
-            return work;
         }
 
         public static ITestResult ExecuteWorkItem(WorkItem work)
--- a/src/NUnitFramework/TestBuilder.cs
+++ b/src/NUnitFramework/TestBuilder.cs
@@ -47,6 +47,10 @@ namespace NUnit.TestUtilities
     public static class TestBuilder
     {
 #region Build Tests
+        public static TestSuite MakeSuite(string name)
+        {
+            return new TestSuite(name);
+        }
 
         public static TestSuite MakeFixture(Type type)
         {
@@ -60,28 +64,6 @@ namespace NUnit.TestUtilities
             return suite;
         }
 
-        public static TestSuite MakeFixture(List<Type> types)
-        {
-            // following the pattern found in DefaultTestAssemblyBuilder
-            var fixtures = new List<Test>();
-            var defaultSuiteBuilder = new DefaultSuiteBuilder();
-            foreach (var type in types)
-            {
-                var typeInfo = new TypeWrapper(type);
-                var test = defaultSuiteBuilder.BuildFrom(typeInfo);
-                fixtures.Add(test);
-            }
-
-            var assembly = AssemblyHelper.Load("nunit.testdata");
-            var assemblyPath = AssemblyHelper.GetAssemblyPath(assembly);
-            TestSuite testSuite = new TestAssembly(assembly, assemblyPath);
-
-            var treeBuilder = new NamespaceTreeBuilder(testSuite);
-            treeBuilder.Add(fixtures);
-
-            return treeBuilder.RootSuite;
-        }
-
         public static TestSuite MakeParameterizedMethodSuite(Type type, string methodName)
         {
             var suite = MakeTestFromMethod(type, methodName) as TestSuite;
@@ -89,22 +71,12 @@ namespace NUnit.TestUtilities
             return suite;
         }
 
-        public static TestSuite MakeParameterizedMethodSuite(object fixture, string methodName)
-        {
-            var test = MakeTestFromMethod(fixture.GetType(), methodName) as ParameterizedMethodSuite;
-            Assert.That(test, Is.TypeOf<ParameterizedMethodSuite>());
-
-            TestSuite suite = test as TestSuite;
-            suite.Fixture = fixture;
-            return suite;
-        }
-
         public static TestMethod MakeTestCase(Type type, string methodName)
         {
-            var test = MakeTestFromMethod(type, methodName);
-            Assert.That(test, Is.TypeOf<TestMethod>());
+            var test = MakeTestFromMethod(type, methodName) as TestMethod;
+            Assert.NotNull(test, "Unable to create TestMethod from {0}", methodName);
 
-            return (TestMethod)test;
+            return test;
         }
 
         // Will return either a ParameterizedMethodSuite or an NUnitTestMethod
@@ -161,15 +133,6 @@ namespace NUnit.TestUtilities
             return RunTest(testMethod, fixture);
         }
 
-#if !NETSTANDARD1_3 && !NETSTANDARD1_6
-        public static ITestResult RunAsTestCase(Action action)
-        {
-            var method = action.Method;
-            var testMethod = MakeTestCase(method.DeclaringType, method.Name);
-            return RunTest(testMethod);
-        }
-#endif
-
         public static ITestResult RunTest(Test test)
         {
             return RunTest(test, null);
@@ -177,18 +140,38 @@ namespace NUnit.TestUtilities
 
         public static ITestResult RunTest(Test test, object testObject)
         {
-            return ExecuteWorkItem(PrepareWorkItem(test, testObject));
+            return ExecuteWorkItem(CreateWorkItem(test, testObject));
         }
 
-        // NOTE: The following two methods are separate in order to support
-        // tests that need to access the WorkItem before or after execution.
+        public static CompositeWorkItem CreateWorkItem(Type type)
+        {
+            return CreateWorkItem(MakeFixture(type)) as CompositeWorkItem;
+        }
 
-        public static WorkItem PrepareWorkItem(Test test, object testObject)
+        public static WorkItem CreateWorkItem(Type type, string methodName)
+        {
+            return CreateWorkItem(MakeTestFromMethod(type, methodName));
+        }
+
+        public static WorkItem CreateWorkItem(Test test)
+        {
+            var context = new TestExecutionContext();
+            context.Dispatcher = new SuperSimpleDispatcher();
+
+            return CreateWorkItem(test, context);
+        }
+
+        public static WorkItem CreateWorkItem(Test test, object testObject)
         {
             var context = new TestExecutionContext();
             context.TestObject = testObject;
             context.Dispatcher = new SuperSimpleDispatcher();
 
+            return CreateWorkItem(test, context);
+        }
+
+        public static WorkItem CreateWorkItem(Test test, TestExecutionContext context)
+        {
             var work = WorkItemBuilder.CreateWorkItem(test, TestFilter.Empty, true);
             work.InitializeContext(context);
 
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -284,41 +284,12 @@ namespace NUnit.Framework.Internal.Execution
 
             WorkShift nextShift = null;
 
-#if false
-            // Shift has ended but all work may not yet be done
-            while (_topLevelWorkItem.State != WorkItemState.Complete)
-            {
-                // This will return if all queues are empty.
-                nextShift = SelectNextShift();
-                if (nextShift != null)
-                    break;
-
-                // If the shift has ended for an isolated queue and there
-                // is no more work, we restore the queues and keep trying.
-                if (_isolationLevel > 0)
-                    RestoreQueues();
-
-                // We are at level zero - just continue to wait
-            }
-
-            // If we have a shift to start, do it
-            if (nextShift != null)
-            {
-                ShiftStarting?.Invoke(nextShift);
-                nextShift.Start();
-            }
-            else // otherwise, shutdown.
-            {
-                foreach (var shift in Shifts)
-                    shift.ShutDown();
-            }
-#else
             while (true)
             {
                 // Shift has ended but all work may not yet be done
                 while (_topLevelWorkItem.State != WorkItemState.Complete)
                 {
-                    // This will return if all queues are empty.
+                    // This will return null if all queues are empty.
                     nextShift = SelectNextShift();
                     if (nextShift != null)
                     {
@@ -339,7 +310,6 @@ namespace NUnit.Framework.Internal.Execution
             // All done - shutdown all shifts
             foreach (var shift in Shifts)
                 shift.ShutDown();
-#endif
         }
 
         private WorkShift SelectNextShift()
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -40,6 +40,26 @@ namespace NUnit.Framework.Internal.Execution
 
         private WorkItem _topLevelWorkItem;
 
+        #region Events
+        
+        /// <summary>
+        /// Handler for ShiftChange events.
+        /// </summary>
+        /// <param name="shift">The shift that is starting or ending.</param>
+        public delegate void ShiftChangeEventHandler(WorkShift shift);
+
+        /// <summary>
+        /// Event raised whenever a shift is starting.
+        /// </summary>
+        public event ShiftChangeEventHandler ShiftStarting;
+
+        /// <summary>
+        /// Event raised whenever a shift has ended.
+        /// </summary>
+        public event ShiftChangeEventHandler ShiftFinished;
+
+        #endregion
+
         #region Constructor
 
         /// <summary>
@@ -66,17 +86,17 @@ namespace NUnit.Framework.Internal.Execution
             // TODO: Avoid creating all the workers till needed
             for (int i = 1; i <= levelOfParallelism; i++)
             {
-                string name = string.Format("Worker#" + i.ToString());
+                string name = string.Format("ParallelWorker#" + i.ToString());
                 ParallelShift.Assign(new TestWorker(ParallelQueue, name));
             }
 
-            ParallelShift.Assign(new TestWorker(ParallelSTAQueue, "Worker#STA"));
+            ParallelShift.Assign(new TestWorker(ParallelSTAQueue, "ParallelSTAWorker"));
 
-            var worker = new TestWorker(NonParallelQueue, "Worker#STA_NP");
+            var worker = new TestWorker(NonParallelQueue, "NonParallelWorker");
             worker.Busy += OnStartNonParallelWorkItem;
             NonParallelShift.Assign(worker);
 
-            worker = new TestWorker(NonParallelSTAQueue, "Worker#NP_STA");
+            worker = new TestWorker(NonParallelSTAQueue, "NonParallelSTAWorker");
             worker.Busy += OnStartNonParallelWorkItem;
             NonParallelSTAShift.Assign(worker);
         }
@@ -149,8 +169,11 @@ namespace NUnit.Framework.Internal.Execution
                 : ParallelExecutionStrategy.Parallel;
 
             Dispatch(topLevelWorkItem, strategy);
-          
-            StartNextShift();
+
+            var shift = SelectNextShift();
+
+            ShiftStarting?.Invoke(shift);
+            shift.Start();
         }
 
         /// <summary>
@@ -244,41 +267,45 @@ namespace NUnit.Framework.Internal.Execution
 
         private void OnEndOfShift(object sender, EventArgs ea)
         {
-            // This shift ended, so see if there is work in any other
-            // TODO: Temp fix - revisit later when object model is redesigned.
-            if (StartNextShift())
-                return;
-
-            if (_isolationLevel > 0)
-                RestoreQueues();
+            ShiftFinished?.Invoke(sender as WorkShift);
+            WorkShift nextShift = null;
 
             // Shift has ended but all work may not yet be done
             while (_topLevelWorkItem.State != WorkItemState.Complete)
             {
-                // This will fail if there is no work - all queues empty.
-                // In that case, we just continue the loop until either
-                // a shift is started or all the work is complete.
-                if (StartNextShift())
-                    return;
+                // This will return if all queues are empty.
+                nextShift = SelectNextShift();
+                if (nextShift != null)
+                    break;
+
+                // If the shift has ended for an isolated queue and there
+                // is no more work, we restore the queues and keep trying.
+                if (_isolationLevel > 0)
+                    RestoreQueues();
+
+                // We are at level zero - just continue to wait
             }
 
-            // All work is complete, so shutdown.
-            foreach (var shift in Shifts)
-                shift.ShutDown();
+            // If we have a shift to start, do it
+            if (nextShift != null)
+            {
+                ShiftStarting?.Invoke(nextShift);
+                nextShift.Start();
+            }
+            else // otherwise, shutdown.
+            {
+                foreach (var shift in Shifts)
+                    shift.ShutDown();
+            }
         }
 
-        private bool StartNextShift()
+        private WorkShift SelectNextShift()
         {
             foreach (var shift in Shifts)
-            {
                 if (shift.HasWork)
-                {
-                    shift.Start();
-                    return true;
-                }
-            }
+                    return shift;
 
-            return false;
+            return null;
         }
 
 #endregion
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -167,16 +167,22 @@ namespace NUnit.Framework.Internal.Execution
         {
             _topLevelWorkItem = topLevelWorkItem;
 
-            var strategy = topLevelWorkItem.ParallelScope.HasFlag(ParallelScope.None)
-                ? ParallelExecutionStrategy.NonParallel
-                : ParallelExecutionStrategy.Parallel;
-
-            Dispatch(topLevelWorkItem, strategy);
+            Dispatch(topLevelWorkItem, InitialExecutionStrategy(topLevelWorkItem));
 
             var shift = SelectNextShift();
 
             ShiftStarting?.Invoke(shift);
             shift.Start();
+        }
+
+        // Initial strategy for the top level item is solely determined
+        // by the ParallelScope of that item. While other approaches are
+        // possible, this one gives the user a predictable result.
+        private static ParallelExecutionStrategy InitialExecutionStrategy(WorkItem workItem)
+        {
+            return workItem.ParallelScope == ParallelScope.Default || workItem.ParallelScope == ParallelScope.None
+                ? ParallelExecutionStrategy.NonParallel
+                : ParallelExecutionStrategy.Parallel;
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -39,9 +39,10 @@ namespace NUnit.Framework.Internal.Execution
         private static readonly Logger log = InternalTrace.GetLogger("Dispatcher");
 
         private WorkItem _topLevelWorkItem;
+        private Stack<WorkItem> _savedWorkItems = new Stack<WorkItem>();
 
         #region Events
-        
+
         /// <summary>
         /// Event raised whenever a shift is starting.
         /// </summary>
@@ -246,6 +247,9 @@ namespace NUnit.Framework.Internal.Execution
                 foreach (WorkItemQueue queue in Queues)
                     queue.Save();
 
+                _savedWorkItems.Push(_topLevelWorkItem);
+                _topLevelWorkItem = work;
+
                 _isolationLevel++;
             }
         }
@@ -255,9 +259,8 @@ namespace NUnit.Framework.Internal.Execution
         /// </summary>
         private void RestoreQueues()
         {
-            Guard.OperationValid(_isolationLevel > 0, $"Internal Error: Called {nameof(RestoreQueues)} with no saved queues.");
-            Guard.OperationValid(Queues.All(q => q.IsEmpty), $"Internal Error: Called {nameof(RestoreQueues)} with non-empty queues.");
-
+            Guard.OperationValid(_isolationLevel > 0, $"Called {nameof(RestoreQueues)} with no saved queues.");
+            
             // Keep lock until we can remove for both methods
             lock (_queueLock)
             {
@@ -265,6 +268,7 @@ namespace NUnit.Framework.Internal.Execution
 
                 foreach (WorkItemQueue queue in Queues)
                     queue.Restore();
+                _topLevelWorkItem = _savedWorkItems.Pop();
 
                 _isolationLevel--;
             }
@@ -280,6 +284,7 @@ namespace NUnit.Framework.Internal.Execution
 
             WorkShift nextShift = null;
 
+#if false
             // Shift has ended but all work may not yet be done
             while (_topLevelWorkItem.State != WorkItemState.Complete)
             {
@@ -307,6 +312,34 @@ namespace NUnit.Framework.Internal.Execution
                 foreach (var shift in Shifts)
                     shift.ShutDown();
             }
+#else
+            while (true)
+            {
+                // Shift has ended but all work may not yet be done
+                while (_topLevelWorkItem.State != WorkItemState.Complete)
+                {
+                    // This will return if all queues are empty.
+                    nextShift = SelectNextShift();
+                    if (nextShift != null)
+                    {
+                        ShiftStarting?.Invoke(nextShift);
+                        nextShift.Start();
+                        return;
+                    }
+                }
+
+                // If the shift has ended for an isolated queue, restore
+                // the queues and keep trying. Otherwise, we are done.
+                if (_isolationLevel > 0)
+                    RestoreQueues();
+                else
+                    break;
+            }
+
+            // All done - shutdown all shifts
+            foreach (var shift in Shifts)
+                shift.ShutDown();
+#endif
         }
 
         private WorkShift SelectNextShift()
@@ -321,7 +354,7 @@ namespace NUnit.Framework.Internal.Execution
 #endregion
     }
 
-    #region ParallelScopeHelper Class
+#region ParallelScopeHelper Class
 
 #if NET_2_0 || NET_3_5
     static class ParallelScopeHelper
@@ -333,6 +366,6 @@ namespace NUnit.Framework.Internal.Execution
     }
 #endif
 
-    #endregion
+#endregion
 }
 #endif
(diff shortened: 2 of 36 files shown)


CharliePoole:
The second commit removes saving and restoring of queues entirely from the dispatcher. This feature was added so that non-parallel fixtures containing parallel test cases could run in absolute isolation from other fixtures. The non-parallel fixture gets a complete set of queues all to itself.

I'm removing it so that it can be re-implemented in a cleaner way. It's possible to do this temporarily because our own tests don't present the situation for which the feature was designed - i.e. having a non-parallel fixture with parallel child tests.

CharliePoole:
Fixes #2478 - a new issue found while debugging parallel SetUpFixtures

jnm2:
Spotted a typo in the first commit message.

src/NUnitFramework/TestBuilder.cs
```diff
@@ -60,51 +64,19 @@ public static TestSuite MakeFixture(object fixture)
...
-        }
-
         public static TestSuite MakeParameterizedMethodSuite(Type type, string methodName)
         {
             var suite = MakeTestFromMethod(type, methodName) as TestSuite;
             Assert.NotNull(suite, "Unable to create parameterized suite - most likely there is no data provided");
             return suite;
         }
 
-        public static TestSuite MakeParameterizedMethodSuite(object fixture, string methodName)
```
jnm2: The hard part with reviewing commits this big is that it seems like this refactor will be a very cross-cutting change, but I'm all for streamlining this API. 😃

src/NUnitFramework/TestBuilder.cs
```diff
@@ -161,34 +133,45 @@ public static ITestResult RunTestCase(object fixture, string methodName)
...
         {
-            return ExecuteWorkItem(PrepareWorkItem(test, testObject));
+            return ExecuteWorkItem(CreateWorkItem(test, testObject));
         }
 
-        // NOTE: The following two methods are separate in order to support
-        // tests that need to access the WorkItem before or after execution.
+        public static CompositeWorkItem CreateWorkItem(Type type)
+        {
+            return CreateWorkItem(MakeFixture(type)) as CompositeWorkItem;
```
jnm2: Are we expecting consuming code to be tolerant of null returns? I prefer direct casts when possible because diagnosis at the call site is easier.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
```diff
@@ -66,17 +86,17 @@ public ParallelWorkItemDispatcher(int levelOfParallelism)
...
-                string name = string.Format("Worker#" + i.ToString());
+                string name = string.Format("ParallelWorker#" + i.ToString());
                 ParallelShift.Assign(new TestWorker(ParallelQueue, name));
             }
 
-            ParallelShift.Assign(new TestWorker(ParallelSTAQueue, "Worker#STA"));
+            ParallelShift.Assign(new TestWorker(ParallelSTAQueue, "ParallelSTAWorker"));
 
-            var worker = new TestWorker(NonParallelQueue, "Worker#STA_NP");
+            var worker = new TestWorker(NonParallelQueue, "NonParallelWorker");
```
jnm2: This always confused me! I like the changes.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
```diff
@@ -244,41 +267,45 @@ private void RestoreQueues()
...
         private void OnEndOfShift(object sender, EventArgs ea)
         {
-            // This shift ended, so see if there is work in any other
-            // TODO: Temp fix - revisit later when object model is redesigned.
-            if (StartNextShift())
-                return;
-
-            if (_isolationLevel > 0)
-                RestoreQueues();
+            ShiftFinished?.Invoke(sender as WorkShift);
```
jnm2: Same thing. Can we direct-cast sender to `WorkShift` or at least throw an `ArgumentException`? Would seem to indicate abuse and safe casts create a secondary breach of expectations (raising ShiftFinished with a null WorkShift) while hiding the root cause.

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
```diff
@@ -122,10 +122,19 @@ private void TestWorkerThreadProc()
...
                     _currentWorkItem.TestWorker = this;
 
+                    // During this Busy call, the queue state may be saved.
+                    // This gives us a new set of queues, which are initially 
+                    // empty. The intention is that only children of the current
+                    // executing item should make use of the new set of queues.
                     Busy(this, _currentWorkItem);
 
+                    // Because we execute the current item AFTER the queue state
+                    // is saved, it's children end up in the new queue set.
```
jnm2: `it's` should be `its`

src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
```diff
@@ -161,11 +161,11 @@ private void StartWorkers()
                 worker.Idle += (s, ea) =>
                 {
                     // Quick check first using Interlocked.Decrement
-                    if (Interlocked.Decrement(ref _busyCount) == 0)
+                    if (Interlocked.Decrement(ref _busyCount) == 0 && !HasWork)
```
jnm2: Two concerns with `HasWork` outside a `lock`: 
 - In the worst case scenario, the current thread's view of `HasWork` is stale and never updates to become `false`, so the lock is never taken. We need a memory barrier like `Volatile.Read` or `Interlocked` which brings me to the next point:
 - Reading the `IList` can result in corrupted results (dirty read, tearing- I've observed this in practice) if the `IList` is being changed by another thread. Either we need an absolute guarantee that the `IList` is thread safe (which solves the first point as well), or we need to lock read as well as write access to it (which again solves the first point).

jnm2: Two concerns with `HasWork` outside a `lock`: 
 - In the worst case scenario, the current thread's view of the list memory behind `HasWork` is stale and never updates to become `false`, so the `if` branch is never executed. We need a memory barrier which brings me to the next point:
 - Reading the `IList` can result in corrupted results (dirty read, tearing- I've observed the weird internal exceptions caused by this in practice) if the `IList` is being changed by another thread. Either we need an absolute guarantee that the `IList` itself is a thread safe collection (which solves the first point as well), or we need to lock read access as well as write access to it (which again solves the first point).

jnm2: And I see it's backed by a `List<>` which means we definitely can't read it without locking access.

src/NUnitFramework/tests/Attributes/TestOrderAttributeTests.cs
```diff
@@ -27,18 +26,15 @@ public void CheckOrderIsCorrect()
...
         [Test]
         [TestCaseSource(nameof(Cases))]
-        public void CheckClassOrderIsCorrect(List<Type> candidateTypes)
+        public void CheckClassOrderIsCorrect(Type[] candidateTypes)
         {
-            var testSuite = TestBuilder.MakeFixture(candidateTypes);
+            var testSuite = new TestSuite("dummy").Containing(candidateTypes);
 
-            var work = TestBuilder.PrepareWorkItem(testSuite, null) as CompositeWorkItem;
+            var work = TestBuilder.CreateWorkItem(testSuite) as CompositeWorkItem;
```
jnm2: Correct me if I'm wrong but my memory is that the return type of `TestBuilder.CreateWorkItem` is already `CompositeWorkItem` s othe `as` is redundant?

src/NUnitFramework/tests/Internal/Execution/ParallelExecutionTests.cs
```diff
@@ -0,0 +1,194 @@
...
+
+            dispatcher.ShiftFinished += (shift) =>
+            {
+                AddEvent("ShiftFinished " + shift.Name);
+            };
+
+            var workItem = TestBuilder.CreateWorkItem(_testSuite, context);
+
+            var run_complete = new ManualResetEvent(false);
+            workItem.Completed += (s, e) => { run_complete.Set(); };
```
jnm2: (`(s, e) => run_complete.Set();` is an option too)

jnm2: Also `run_complete` should really be in a `using` statement because it holds unmanaged handles.

src/NUnitFramework/tests/Internal/Execution/ParallelExecutionTests.cs
```diff
@@ -0,0 +1,194 @@
...
+        public void AllTestsPassed()
+        {
+            if (_result.ResultState != ResultState.Success)
+                Assert.Fail(DumpEvents("Not all tests passed"));
+        }
+
+        [Test]
+        public void AllTestsRan()
+        {
+            if(_result.PassCount != _numCases)
```
jnm2: Inconsistent spacing after the `if` keyword

src/NUnitFramework/tests/Internal/Execution/ParallelExecutionTests.cs
```diff
@@ -0,0 +1,194 @@
...
+                Assert.Fail(DumpEvents("Incorrect number of test cases"));
+        }
+
+        [Test]
+        public void OnlyOneShiftMayBeActive()
+        {
+            int count = 0;
+            foreach (var @event in _events)
+            {
+                if (@event.StartsWith("ShiftStarted") && ++count > 1)
```
jnm2: I guess this is really not important for tests, but if this was shipping code, I'd yell about the missing `StringComparison.Ordinal` parameter because thread culture can wreck string comparisons.

src/NUnitFramework/tests/Internal/Execution/ParallelExecutionTests.cs
```diff
@@ -0,0 +1,194 @@
...
+                3);
+
+            yield return new TestFixtureData(
+                TestBuilder.MakeSuite("NUnit")
+                    .Containing(TestBuilder.MakeSuite("TestData")
+                        .Containing(TestBuilder.MakeSuite("ParallelExecutionData")
+                            .Containing(TestBuilder.MakeFixture(typeof(TestSetUpFixture))
+                                .Containing(typeof(TestFixture1), typeof(TestFixture2), typeof(TestFixture3))))),
+                3);
+        }
```
jnm2: Neat!

jnm2:
Sorry, gotta go but I'll be back.

CharliePoole:
I initially worked on this with the idea of completely changing the structure of parallel execution. Test fixtures that needed to run in isolation would use an entirely new dispatcher, the old dispatcher would quiesce until the new one finished and the new dispatcher would get new work shifts, queues and workers.

This turned out to be very complex, requiring intercommunication among the dispatchers. After spending about 20 hours on it, I'm going back to the original design and trying to make it work. That is, `WorkItemQueue` maintains a stack of real queues under the covers. This is used to isolate non-parallel fixtures so that it's impossible for anything else to run at the same time. 

The current / continuing design is a bit ugly in my opinion, but it has been working and I think it's a better use of time to fix the bugs right now.

src/NUnitFramework/tests/Internal/Execution/ParallelExecutionTests.cs
CharliePoole: I agree with both points. 😈

src/NUnitFramework/tests/Attributes/TestOrderAttributeTests.cs
CharliePoole: It returns `WorkItem` in this overload. But I'll look at the overloads again.

src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
CharliePoole: This could be a source of some intermittent issues I'm seeing. I'll check.

BTW, this is far from done!

CharliePoole: @jnm2 Thinking about this, I realized that I'm relying on the fact that the underlying list is never changed once set up. A lot of the parallel dispatcher code is like that...

* `ParallelWorkItemDispatcher` has fixed lists of shifts, queues and workers.
* `WorkShift` has a specific set of queues and workers.
* `WorkItemQueue` has a specific set of workers.

For that reason, I've played a little loose with locking. Maybe it would be better to use a collection that can't be updated at all, just in case.

CharliePoole:
Merges are pulling in code that uses the old `TestBuilder` interfaces. I'll rebase again.

src/NUnitFramework/TestBuilder.cs
CharliePoole: Different philosophy on my part for tests than for production code, I guess. In general, the return is used immediately and we get a null reference exception. I think it's probably six of one etc. unless the return is being held on to for a long time and then used later.

src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
jnm2: Ah, and I'm so used to looking for `IReadOnlyList<>` to make an appearance for that guarantee... It would simplify my understanding of the code if there was a builder type. How much code would break if `WorkShift` stopped exposing the `IList<>` and made a defensive array copy of the list passed into the constructor, stored as a private readonly `WorkItemQueue[]` and an XML comment on it that it should be immutable? (Or we could even add a lightweight `ImmutableArray` to our BCL-like toolbox?)

So long as no thread mutates the `IList<>` after the first time another thread reads it, we don't need to lock access to it.

src/NUnitFramework/tests/Attributes/TestOrderAttributeTests.cs
jnm2: Taking another look, this does make sense. Though again I think the safe cast clouds intent, whichever way you look at it. But nbd.

jnm2:
GitHub thinks a bunch of my comments are no longer relevant but they are. Should I delete and recomment?

CharliePoole:
@jnm2 GitHub does that after a forced push and I just had to do two in a row. Wait till I fix the things you already pointed out - I can see them with a few extra clicks.

src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
CharliePoole: Queues is only public for test purposes. Could easily be internal. I'll take a look at the other collections of queues, shifts and workers as well. This may be an easy fix.

src/NUnitFramework/TestBuilder.cs
CharliePoole: I added a comment clarifying that the cast is guaranteed - WorkItem created from a TestSuite is always composite.

jnm2: In my mind, safe cast + "cast is guaranteed" comment can be refactored to direct cast and communicate precisely the same thing. But no big deal.

CharliePoole: Hmm... I don't think it communicates that. IT gives rise to the comment "How do you know this cast won't throw?" There's no satisfying everyone, I guess. 😄

CharliePoole: I changed the return type. Problem solved!

src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
CharliePoole: For now, I'm only changing this to internal. In fact, determining whether a shift is really over is the key to a lot of our issues so I expect to be back in this bit of code in subsequent commits as I add tests for the various issues.

src/NUnitFramework/TestBuilder.cs
jnm2: In either case you'd also have to explain _how_ you know it won't throw. But no big deal; I'm satisfied either way. Just exchanging viewpoints!

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
```diff
@@ -244,41 +276,46 @@ private void RestoreQueues()
 
         private void OnEndOfShift(object sender, EventArgs ea)
         {
-            // This shift ended, so see if there is work in any other
-            // TODO: Temp fix - revisit later when object model is redesigned.
-            if (StartNextShift())
-                return;
+            ShiftFinished?.Invoke(sender as WorkShift);
```
jnm2: This one is production code. Can we direct-cast sender to `WorkShift` or at least throw an `ArgumentException`? Would seem to indicate abuse and safe casts create a secondary breach of expectations (raising ShiftFinished with a null WorkShift) while hiding the root cause.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -248,6 +253,16 @@ public virtual void Execute()
 #endif
         }
 
+        private ManualResetEvent _completionEvent = new ManualResetEvent(false);
```
jnm2: `ManualResetEvent` is `IDisposable` and therefore the owning class should implement `IDisposable` and dispose it to free the kernel handles.

jnm2: The field should be `readonly` too.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -177,6 +177,11 @@ public virtual ParallelExecutionStrategy ExecutionStrategy
                 return _executionStrategy.Value;
             }
         }
+
+        /// <summary>
+        /// Indicates whether this work item should use a separate dispatcher.
+        /// </summary>
+        public virtual bool IsolateChildTests { get; } = false;
```
jnm2: We can save an autogenerated field per instance by doing `IsolateChildTests => false;` since we aren't setting `IsolateChildTests` in a constructor.

src/NUnitFramework/tests/Internal/Execution/ParallelExecutionTests.cs
```diff
@@ -0,0 +1,196 @@
...
+        public void AllTestsPassed()
+        {
+            if (_result.ResultState != ResultState.Success)
+                Assert.Fail(DumpEvents("Not all tests passed"));
+        }
+
+        [Test]
+        public void AllTestsRan()
+        {
+            if(_result.PassCount != _numCases)
```
jnm2: Inconsistent spacing after the `if` keyword

src/NUnitFramework/tests/Internal/Execution/ParallelExecutionTests.cs
```diff
@@ -0,0 +1,196 @@
...
+                _events.Add(@event);
+        }
+
+        private string DumpEvents(string message)
+        {
+            var sb = new StringBuilder(message + NL);
+            foreach (string @event in _events)
+            {
+                sb.Append(@event);
+                sb.Append(NL);
```
jnm2: `sb.AppendLine()`?

src/NUnitFramework/tests/Internal/Execution/ParallelExecutionTests.cs
```diff
@@ -0,0 +1,196 @@
...
+
+        private void AddEvent(string @event)
+        {
+            lock (_events)
+                _events.Add(@event);
+        }
+
+        private string DumpEvents(string message)
+        {
+            var sb = new StringBuilder(message + NL);
```
jnm2: `new StringBuilder(message).AppendLine();`?

src/NUnitFramework/tests/Internal/Execution/ParallelExecutionTests.cs
jnm2: `new StringBuilder(message).AppendLine();`, and `sb.AppendLine(@event)` here? This would allow you to remove the `NL` declaration too.

src/NUnitFramework/tests/Internal/Execution/ParallelExecutionTests.cs
```diff
@@ -0,0 +1,196 @@
...
+using NUnit.TestData.ParallelExecutionData;
+using NUnit.TestUtilities;
+
+namespace NUnit.Framework.Internal.Execution
+{
+    [TestFixtureSource(nameof(GetParallelSuites))]
+    public class ParallelExecutionTests : ITestListener
+    {
+        private TestSuite _testSuite;
+        private int _numCases;
```
jnm2: Both can be `readonly`.

src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
jnm2: I got sidetracked talking about the list, but we still have a staleness issue checking `WorkItemQueue.IsEmpty`. We need locked access to that mutable state.

jnm2: [Just checked on it.](https://github.com/nunit/nunit/blob/0162e7c1d8640769be2714ea81a3002903f36617/src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs#L169-L179) No staleness issue because you're looking through an array (which I assume is also mot mutated once multiple threads see it) and each array item is a ConcurrentQueue.

Though on a new topic now, having more than one concurrency primitive in action makes me feel like everything is susceptible to race conditions. What happens if one of the queues gets an item added on a background thread right after `IsEmpty` returns false?

jnm2:
Pulling the code.

jnm2:
I'm trying to put hte whole picture together but I am feeling uneasy about things like this: Between checking `.HasWork` and returning the shift, another thread may cause `HasWork` to become false.

https://github.com/nunit/nunit/blob/0162e7c1d8640769be2714ea81a3002903f36617/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs#L312-L318

The next thing that happens is that we call `shift.Start()` on a shift that has potentially just ended:

https://github.com/nunit/nunit/blob/0162e7c1d8640769be2714ea81a3002903f36617/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs#L299-L304

Which starts a new thread and loops forever until it finds work which may never happen because the work ended between `HasWork` checking it and returning it:

https://github.com/nunit/nunit/blob/0162e7c1d8640769be2714ea81a3002903f36617/src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs#L273-L278

Am I missing something or could this cause the thread to busy-wait forever?

jnm2:
All the busy waits I'm seeing could also be contributors to the CPU spikes.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
CharliePoole: Changing delegate to be type-safe.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
CharliePoole: Good one!

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
CharliePoole: I don't love this syntax, especially when the property is virtual. IMO, the existing syntax is clearer to read and avoids confusion.

CharliePoole:
@jnm2 Regarding HasWork, here's the logic that __should__ apply, unless there is a bug.

* Only one shift ends at a time, so when a shift has ended, no shift is running.
* We only call HasWork when no shift is running
* Since no shift is running, no thread can create more work.

Admittedly, there could be a problem here. However, I'm more suspicious of the correct detection of the end of the shift than I am of the checking logic. If a shift ends while some thread is still running, then all of the assumptions above can be wrong.

What I've done so far is merely creating an infrastructure for us to test all this stuff for real. As I add more test cases now, I'm hoping they will tell us more.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
jnm2: This is theoretical but C# 7.0 allows `IsolateChildTests { get => false; }` as well, if you still want to use the `get` keyword without allocating a field.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -300,9 +315,35 @@ public virtual void Cancel(bool force)
...
+        #region IDisposable Implementation
+
+        /// <summary>
+        /// Standard Dispose
+        /// </summary>
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
```
jnm2: I'm sorry; it crossed my mind if I should mention this originally because MSDN is out of date and misleads people. Real quick: since .NET 4.0 introduced the SafeHandle class, it's now an anti-pattern to use the `Dispose(bool)` pattern or to involve the finalizer. (The SafeHandle contains the finalizer logic and gets other special treatments. Any unmanaged resources (bare IntPtr handles) should go inside a SafeHandle class instead, and the SafeHandle, being itself a managed resource just gets disposed like any other managed object.)

Now I know we run on net35 and net20, but the same logic applies: the `WorkItem` class itself, and any subclass, will not and should not ever directly own bare unmanaged (IntPtr) handles. So as a safeguard we can remove everything except `public void Dispose() { _completionEvent.Dispose(); }`. (And we don't need the null check because the variable is readonly.)

jnm2: ManualResetEvent on net35 and net20 already contains the finalizer logic too, so it would be redundant for the `WorkItem` class to have finalizer logic. It's already guaranteed to be disposed without `WorkItem` doing anything special.

jnm2: I know that this is counter to a lot of common practice, so it's understandable that you might want confirmation that I'm not crazy on this. 😀 This comes up a lot at dotnet/csharplang and I can loop in sharwell or someone.

CharliePoole: @jnm2 I won't be changing this right away - it's also an anti-pattern to keep changing minor bits in code that is not yet even doing what it's supposed to do!!! I'll hold it in mind as I work on the next commit with actual functionality.

Based on what you are now saying, it sounds as if the initial code was correct and that WorkItem does not need to implement dispose. I understood that MRE basically took care of itself in different ways on each platform, so we didn't have to. I'm not in favor of implementing an empty Dispose "just in case". Or are you saying we need to implement and invoke DIspose solely for the purpose of getting the release done early?

Please give it some thought and answer at your leisure.

jnm2: > it's also an anti-pattern to keep changing minor bits in code that is not yet even doing what it's supposed to do!!!

Yes! :D

> Or are you saying we need to implement and invoke DIspose solely for the purpose of getting the release done early?

Yes, exactly. Deterministic release is always preferable to waiting for the garbage collector to finalize an object. One reason is that the garbage collector reacts to memory pressure but not to leaking handles so you can go for an arbitrarily large amount of time without a GC and yet still leaking kernel handles.

The finalization handling is a backstop, not a great one but a necessary one, and since it's built into ManualResetEvent we don't need to build it ourselves. But we'd prefer not to use the backstop.

CharliePoole: Do we believe this to be true on all platforms, including Mono?

jnm2: I checked Mono and it's using a SafeHandle. (https://github.com/mono/mono/blob/master/mcs/class/referencesource/mscorlib/system/threading/eventwaithandle.cs)

If it's not true in a platform's BCL, it's a bug in the platform just as much as it would be a bug in NUnit if we used a bare IntPtr handle without encapsulating it in a SafeHandle class.

.NET Framework and .NET Core are identical. Mono has been incrementally replacing its BCL with .NET Core's BCL implementation. Xamarin and Unity use Mono's BCL.

CharliePoole: Sounds like the right way to go then. Ironic to be doing what was called bad practice for so long.

jnm2: This is over 3 years old, but may be relevant: http://www.mono-project.com/docs/advanced/safehandles/

CharliePoole:
This now fixes #2395 

Restoring the queues was starting all queues, causing __everything__ to run in parallel.

CharliePoole:
I feel like this is ready for a final review and merge. I imagine there is more to be done around parallel execution, but since this is getting pretty big I think we're better off to put it out there and fix further problems incrementally.

The parallel tests I created can be extended as needed to simulate parallel execution in other environments.

rprouse:
I am going through these and writing repros for each of the issues to see if this PR fixes them. So far,

#1239 - I cannot reproduce on my machine, could be number of cores?
#2364 - Unit tests are enough
#2388 - Using the tests in the issue, this is **not fixed**, the given tests still fail with this code
#2395 - This looks to be a duplicate of #2388, so **likely won't be fixed**
#2467 - Confirmed to be a framework problem in 3.8.1 and **fixed** with this PR
#2471 - Confirmed **fixed**
#2478 - I think I've confirmed this by applying `Parallizable` to namespace `OneTimeSetup`s, mixing them with `TestContext.Progress.WriteLine` and `Thread.Sleep` and observing the behavior. Will also look in the code.

So, at this point, assuming the code review finds the same, all issues marked fixed except for #2388 and #2395 are fixed by this PR. We can treat those issues in a separate PR unless you feel they should be included in this PR @CharliePoole.

Code review to follow. I've done a quick pass, now doing a more detailed pass after reading and testing all the issues.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
rprouse: @CharliePoole I don't think it is an issue since this is the way we have been doing it, but were you planning on updating this based on the comments?

src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
```diff
@@ -339,14 +339,60 @@ internal void Save()
         /// </summary>
         internal void Restore()
         {
-            Pause();
+            //Guard.OperationValid(State != WorkItemQueueState.Running, $"Attempted to restore state of {Name} while queue was running.");
```
rprouse: Is this a valid guard? Should it be uncommented or deleted?

src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
```diff
@@ -161,16 +170,18 @@ private void StartWorkers()
...
+                    {
                         lock (_syncRoot)
                         {
-                            // Check busy count again under the lock. If there is no work
-                            // try to restore any saved queues and end the shift.
+                            // Check again under the lock. If there is no work
```
rprouse: @jnm2 said above,

> having more than one concurrency primitive in action makes me feel like everything is susceptible to race conditions. What happens if one of the queues gets an item added on a background thread right after IsEmpty returns false?

@jnm2 isn't this okay because we check `_busyCount` and `HasWork` again after getting the lock? The comment is pretty obvious, so what you are seeing is more subtle than that. Can you expand on your concern?

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
```diff
@@ -242,49 +278,83 @@ private void RestoreQueues()
...
-            // TODO: Temp fix - revisit later when object model is redesigned.
-            if (StartNextShift())
-                return;
+            ShiftFinished?.Invoke(endingShift);
 
-            if (_isolationLevel > 0)
-                RestoreQueues();
+            WorkShift nextShift = null;
 
+#if false
```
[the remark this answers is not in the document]
CharliePoole: Slipped through!

src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
CharliePoole: It's logically valid (in my mind) but breaks things if I use it. I think I'll keep it commented but add an explanation in front of it. At some point, someone will discover __why__ some queues end up running at a point where we are supposedly between shifts. My guess is that some other thread is getting in their and restarting things.

This is where I'd continue if I had time to do it by the way. We need one thread to detect end of shift and take care of starting next shift, restoring, etc. before any other shift tries to do the same. 

However, I think that overall this fixes a lot of things and is not any worse in the areas it doesn't fix.

src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
CharliePoole: I agree with both of you. I'm worried like @jnm2 and not seeing how it happens like @rprouse. 😄 

My observation is that things get in - it's rare but happens.

CharliePoole:
I'll make the changes in the morning. This is far from perfect, but I think it's better than what we have and should be merged so that others can carry on from this point. In particular, the tests can be expanded and possibly the test mechanism improved.

BTW, if #2388 is not fixed, that would be because the simulation (that's what it is) of executing an assembly doesn't fully match the reality. I was hoping to avoid adding true assemblies that have to be loaded when we are working on unit tests. We need a set of tests that run out of actual assemblies to really test what happens with assembly-level attributes but I don't think they should be part of the framework tests.

That said, it may be better to improve the simulation so that we run more of the production code and less `TestBuilder` code. We may be able to inject a context right at the start with a particular (simulated) assembly-level attribute. If nobody else gets too it I may try to take it further in December.

rprouse:
I agree that this is an improvement and something we should get in sooner than later. This is also getting huge, so it might be better to split out the remaining issues. I also agree that we need true test assemblies to test some of these things because our unit tests are not catching them. That is why I was suggesting some sort of integration tests that test some of these scenarios. Some of these issues are difficult to even test that way though. For some, I had to add in sleeps and watch the output before and after the fixes.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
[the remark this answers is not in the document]
CharliePoole: Fixed

CharliePoole:
I removed the fixes comment for #2388 based on your tests. However, I'm pretty sure that #2395 is fixed. It doesn't seem to be a duplicate of the other at all. According to the extract from the internal trace, multiple threads were running at the same time, when they should not have been. That is, the parallel and non-parallel workers were active simultaneously, which of course defeats their purpose.

I found this was due to an error in how the dispatcher restored queues when we switch from an isolated non-parallel situation back to the lower level parallel one. It was starting all the queues! I fixed this and I think it needs to be checked to see if there is more to fix. Best bet would be to merge and then encourage the reporter of the issue to try the dev build. Since we don't want anyone else to work on it until that is done, it seems best to close it as done.

CharliePoole:
@rprouse I made the minor changes you requested. The only open question is whether this should really close #2395. I left that "fixes" in the description, but we can re-open it if it turns out I'm wrong.

PR #2482, opened 2017-10-10
Honor --label option settings in nunitlite

### LINKED ISSUE #2402: --labels=All doesn't show anything in console output executing NUnitLite Console Runner

CharliePoole:
Fixes #2402 

This is a small fix I included in my parallelism-issues branch, extracted here as a separate PR. It's important for anyone that wants to see labels on nunitlite output for debugging.

### DIFF
--- a/src/NUnitFramework/nunitlite/TextRunner.cs
+++ b/src/NUnitFramework/nunitlite/TextRunner.cs
@@ -448,6 +448,8 @@ namespace NUnitLite
         {
             if (_teamCity != null)
                 _teamCity.TestStarted(test);
+
+            _textUI.TestStarted(test);
         }
 
         /// <summary>


mikkelbu:
I've restarted the Travis build.

Issue #2091, opened 2017-03-22
When a native exception of corrupted state is thrown, nunit test thread crashes and the nunit-console process hangs


endrih:
When using NUnit in C++/CLI and the state is somehow corrupted (unrecoverable errors), the nunit thread that runs the test crashes. The nunit-console thread that is waiting on the operation to complete, never exits the wait and therefore hangs the whole process.

Maybe using NUnit for C++/CLI is not supported, but we have a lot of them.

I have made a quick fix like this in NUnit.Framework.Internal.Reflect in Reflect.cs:
```
[HandleProcessCorruptedStateExceptions] //put here to handle native corrupt state exceptions. 
        public static object InvokeMethod( MethodInfo method, object fixture, params object[] args )
        {
            if(method != null)
            {
                try
                {
                    return method.Invoke(fixture, args);
                }
#if !PORTABLE && !NETSTANDARD1_6
                catch (System.Threading.ThreadAbortException)
                {
                    // No need to wrap or rethrow ThreadAbortException
                    return null;
                }
#endif
                catch (TargetInvocationException e)
                {
                    throw new NUnitException("Rethrown", e.InnerException);
                }
                catch (Exception e)
                {
                    throw new NUnitException("Rethrown", e);
                }
#pragma warning disable 1058
                catch
#pragma warning restore 1058
                {
                    throw new NUnitException("Rethrown", new Exception("Unknown Native exception thrown."));
                }
            }

            return null;
        }
```

jnm2:
A note on `catch`: In C#, any non-`Exception` objects get wrapped and `catch (Exception e)` already catches everything.

See `RuntimeWrappedException` on [MSDN](https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.runtimewrappedexception.aspx#Anchor_4).

jnm2:
Fyi: this was never fixed. The PR has never had any effect because https://github.com/nunit/nunit/pull/1775 added `[assembly: AllowPartiallyTrustedCallers]` and the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.handleprocesscorruptedstateexceptionsattribute?view=netframework-4.8) say this means that `[HandleProcessCorruptedStateExceptions]` is ignored:

> The HandleProcessCorruptedStateExceptionsAttribute attribute is ignored when it is encountered in partially trusted or transparent code, because a trusted host should not allow an untrusted add-in to catch and ignore these serious exceptions.

Issue #1275, opened 2016-02-19
parameters to configure test runner so that it runs in ui thread - calling thread


jmecosta:
hi,

we have adapted the visual studio adapter, to run nunit tests inside our product. we have a requirement now that we need to test some ui code, and for this to happen we need,
 . run tests in same appdomain as our product
 . run tests in the same thread, UI with STA set.

For the appdomain, we fixed it in the test adapter. For the thread it is a bit harder and i am ensure about the parameters that we can use to set this in the Nunit engine. we look at the doc about the RequiresThread Attribute but from reading looks like the engine always creates a thread, and this we would like to avoid.

is this possible?

thanks in advance.

br,
jorge costa

rprouse:
The tests run by default in a different app domain as nunit itself, but all of your application and test code is in one AppDomain.

As for UI threads, first, mark your tests as requiring STA, then you will need to marshal your test code back to the UI thread. How you do that depends on the UI framework, but is usually something like `control.BeginInvoke(() => // run tests)`.

I am closing this issue as a question. Feel free to continue posting. I am also sorry that we did not get back to you sooner, we will try to do better in the future.

CharliePoole:
We should also note that 3.2 is running each test on a separate thread. This is fixed in the source. The 3.0.1 release is friendlier in this respect. Also in the latest source, take look at the SingleThreaded attribute.

jmecosta:
@rprouse  yep, i was afraid we would not be able to disable all domain, thread from our side. and creating code to marshal unmarshal is not something we want to maintain. So we decided to create are own simple runner that just read test attribute and run those without changing any domains. fortunately all nunit assert are working also so we are not blocked. 

@CharliePoole we havent move to 3.2 yet, so good to know once we move the other tests we have using nunit runner thanks.

CharliePoole:
Once the engine has run your tests (called by the console runner or some other runner) everything is under control of nunit.framework. There is no code in the framework to create new domains or processes - that's higher level in the NUnit architecture. There is code to run on threads and to create new threads at your request. But tests all run on the thread used by their fixture unless you apply an attribute (e.g. RequiresThread) to the test method to force use of a new thread.

The new SingleThreadedAttribute is applied to a fixture and is intended to guarantee that everything in that fixture runs on the same thread. It will be in the next release.

I repeat from above that NUnit 3.2 is broken for your purposes, since it creates a new thread for every test case. Use 3.0.1 or wait for 3.2.1 or higher.

[...]

PR #2483, opened 2017-10-10
Nunit3.6 same thread changes

endrih:
We wanted to execute NUnit tests within our product (similar to issue  #1275), and in order to achieve the requirement to run tests on the same thread and app domain as our product, we had to do the following changes. 
The change in Reflect.cs is related to issue #2091 that I have posted earlier. 

We would like to achieve the same thing in 3.8.1 and possibly avoid to have to do these changes in the future when we want to upgrade our NUnit framework. Do you think this approach is the right way to achieve what we want?

In our product, in order to run the tests, later we have to give the following package to the engine in the GetRunner(package) method, 

```
TestPackage package = new TestPackage(assemblyName);
package.AddSetting(EnginePackageSettings.ProcessModel, "InProcess");
package.AddSetting(EnginePackageSettings.DomainUsage, "None");
package.AddSetting(EnginePackageSettings.DisposeRunners, true);
package.AddSetting(EnginePackageSettings.ShadowCopyFiles, false);
package.AddSetting("NumberOfTestWorkers", 0);
 // Always add work directory, in case current directory is changed
var workDirectory = Environment.CurrentDirectory;
package.AddSetting("WorkDirectory", workDirectory);

ITestRunner runner = engine.GetRunner(package);
result = runner.Run(eventHandler, filter);
```

We would appreciate your comments or suggestions on this.


ChrisMaddock:
This feels like it should be in the domain of `SingleThreadedAttribute`, rather than based on NumberOfTestWorkers = 0. The latter already has special meaning, I believe.

Out of interest, if you modified SingleThreadedAttribute to allow it to be applied at assembly level, and used that, does that work for your situation?

ChrisMaddock:
(Sorry - I removed my previous comment as it was incorrect. Will revisit this later.)

ChrisMaddock:
My concern with this, is that `NumberOfTestWorkers=0` already creates specific test-running scenario, which people are relying on. If we were to add this functionality, I think it would need to be via different means.

The general idea - I have no objection to, but probably worth waiting to hear from others in the @nunit/framework-team.

CharliePoole:
I agree that this can't be simply piggybacked on top of `--workers=0`, which does it's own thing.

I think the fix of #2091 belongs in it's own PR.

I'd want to see an example of how this is intended to be used. How does the user access this feature?

When this code executes, an AppDomain has already been created. The framework doesn't create AppDomains, the runner does that.

We have something similar in the VS adapter, where a setting forces us to run in the same AppDomain.

Note that we already have a setting that makes tests run on the same thread as the parent test. SO all that is necessary here is to get the top-level test to run on the correct thread.

As implemented, this setting applies to all tests in an assembly. Is that the intent? Isn't there any need to distinguish __which__ tests have to run on the main thread? 

We may want to look at the changes that Xamarin created for the same purpose.

endrih:
Sorry, I didn't actually know that `--workers=0` was doing anything special, also I could not find something related to it in the documentation. What is that option supposed to do? I saw in the code only conditions referring to it being >=0, so I thought it makes sense to ask to run on the same thread when you say number of worker threads to be  zero. 

We basically want to include the NUnit engine within our product, in order to run integration test from the product itself, on the same AppDomain as the running application. Also, the tests have to run on the same thread as the running application; Other threads are not allowed to access the data structures inside.

So, I understand there is a setting for running tests within the same AppDomain relative to each other and the fixture, however we want them to run on the same AppDomain as the running application, on the same thread. 

I don't know if the following parallelism would make our scenario clearer: If nunit3-console.exe was our application, we would need all the tests in an assembly to run on the same thread as the nunit3-console itself and on the same AppDomain.

mikkelbu:
Hi @endrih There is a little about `workers=0` in https://github.com/nunit/docs/wiki/Framework-Parallel-Test-Execution 

> If the command line specifies zero workers, all use of the dispatcher and its queues is bypassed and tests are run sequentially on a single thread.

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItemDispatcher.cs
```diff
@@ -59,17 +59,17 @@ public void Dispatch(WorkItem work)
             if (work != null)
                 work.Execute();
 #else
-            if (_topLevelWorkItem != null)
+            if (_topLevelWorkItem != null || work.RunsOnMainThread)
```
CharliePoole: The `if` statement is intended to distinguish the first time any `WorkItem` is run, so that we  set its value in `_topLevelWorkItem`.  I think you need a different approach here. I'll try to outline in a top-level comment so it won't be lost as changes are made.

CharliePoole:
@endrih Adding to @mikkelbu 's comment, the setting `--workers=0` causes NUnit to essentially run tests in the same way they were run in NUnit V2: one test at a time, no queues, no workers, no parallelism. With this setting we use `SimpleWorkItemDispatcher` rather than `ParallelWorkItemDispatcher`. You definitely want to use this setting in your package.

However, you want something extra. You want to run the tests on the same thread that initiates the run, which, in your case, happens to be the main application thread. Since you want something extra, you need an extra setting that will be recognized by the framework. Add the new setting to `PackageSettings.cs` and use it when creating your package. When you see it, don't create a new thread.

Although the new setting needs to be separate from `NumberOfTestWorkers`, it probably should force `NumberOfTestWorkers` to be zero, so there is no chance that the parallel dispatcher will be used in error.

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItemDispatcher.cs
ChrisMaddock: I hadn't commented on implementation, as I think we need to discuss the concept itself first - but I'd prefer to see a separate IWorkItemDispatcher for this, chosen by a frameworkpackagesetting higher up. Both the netstandard builds also execute directly on the main thread - which is currently in SimpleWorkItemDispatcher with #ifdef's - separating this out into a separate implementation would be cleaner imo.

ChrisMaddock: (Charlie - apologies, your other comment hadn't displayed when I typed this! Hope we're in agreement.)

CharliePoole: @ChrisMaddock I think we are in fundamental agreement. Whether we have a separate dispatcher or use a setting that effects how the dispatcher works is a detail that could be decided based on how many #if statements apply.

CharliePoole: I looked at the code. Seems like it could be done either way. However, I think we should delay till my parallel work, which may change the `IWorkItemDispatcher` interface is done. Specifically, I'm trying to get rid of the Start method entirely, which is where one of the biggest differences is found.

endrih: Thank you for your comments. Do you think I should wait until your changes @CharliePoole are done in 3.8 and then try to reapply my changes (with your suggestions) on top of that, in the `Start` method of `SimpleWorkItemDispatcher`?

CharliePoole:
@endrh My work is merged. It turns out there's only one small addition to the interface.

Issue #2486, opened 2017-10-12
Message when asserting null with Is.EquivalentTo could be more helpful


appel1:
If `result.Things` is null in the following example the failure message is a little bit confusing. 

```csharp
Assert.That(result.Things, Is.EquivalentTo(new[] { "abc" }));
```

Message:
`System.ArgumentException : The actual value must be an IEnumerable`

Perhaps the null case could have a more specific message?

ChrisMaddock:
I don't see any harm in someone implementing this. I think it may be slightly non-trivial, but I haven't looked in detail.

As I'm sure you're aware, if `result.Things` is possibly null, for now you could do:

    Assert.That(result.Things, Is.Not.Null.And.EquivalentTo(new[] { "abc" }));

ChrisMaddock:
I'll label this up-for-grabs soon, but would like to give it 24 hours or so, and give anyone else the chance to disagree!

mikkelbu:
I also think that this is a reasonable wish.

Note, that "the change" probably needs to be done in approx 10 places to be consistent, as we have similar checks for `IEnumerable` in most collection constraints.

ChrisMaddock:
I was overthinking this.

It would be trivial to include a different _error_ message when actual is null. The current error message is meant for when a non-ienumerable is passed in, not when a 'null ienumerable' is passed in. It would be more involved to instead produce a custom failure message.

jnm2:
I'm not opposed to changing it to a failure message like "Expected: IEnumerable But was: null`.

But if we keep it as an exception message, I've always thought it was the perfect message. It's a well-known C# fact that `null` is not `IEnumerable` (see `null is IEnumerable`).

josteink:
I've submitted a PR for this issue. Consumer-wise and code-wise the result should be the same as before, but with another exception-message.

PR #2489, opened 2017-10-13
Check for null before checking for type IEnumerable.

josteink:
This closes https://github.com/nunit/nunit/issues/2486.

### DIFF
--- a/src/NUnitFramework/framework/Internal/ConstraintUtil.cs
+++ b/src/NUnitFramework/framework/Internal/ConstraintUtil.cs
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2015 Charlie Poole, Rob Prouse
+// Copyright (c) 2017 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
@@ -34,11 +34,11 @@ namespace NUnit.Framework.Internal
         /// <summary>
         /// Require that the provided object is actually of the type required,
         /// and that it is not null.
-        /// <param name="actual">The object to verify</param>
-        /// <param name="paramName">Name of the parameter as passed into the checking method.</param>
-        /// <typeparam name="T">The type to require</typeparam>  
-        /// <returns>A properly typed object, or throws. Never null.</returns>
         /// </summary>
+        /// <param name="actual">The object to verify.</param>
+        /// <param name="paramName">Name of the parameter as passed into the checking method.</param>
+        /// <typeparam name="T">The type to require.</typeparam>  
+        /// <returns>A properly typed object, or throws. Never null.</returns>
         public static T RequireActual<T>(object actual, string paramName)
         {
             if (actual is T) return (T)actual;
--- a/src/NUnitFramework/framework/Internal/ConstraintUtil.cs
+++ b/src/NUnitFramework/framework/Internal/ConstraintUtil.cs
@@ -1,0 +1,49 @@
+// ***********************************************************************
+// Copyright (c) 2015 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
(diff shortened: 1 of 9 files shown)


CharliePoole:
Looks good to me.

josteink:
I like that. I’ll make it so.

jnm2:
:+1: `ConstraintUtils` was just off the top of my head, so if you find an existing helper class that seems related, we should probably use that.

josteink:
I've updated the code in the places I've found which needed changing, and I've corrected the issues identified by the CI-build.

If you think there are more changes which needs to be made, please let me know. Otherwise I'll assume this is ready for a new review :)

src/NUnitFramework/framework/Internal/ConstraintUtil.cs
```diff
@@ -0,0 +1,50 @@
...
+    internal class ConstraintUtil
+    {
+        /// <summary>
+        /// Require that the provided object is actually of the type required,
+        /// and that it is not null.
+        /// <param name="actual">The object to verify</param>
+        /// <param name="paramName">Name of the parameter as passed into the checking method.</param>
+        /// <typeparam name="T">The type to require</typeparam>  
+        /// <returns>A properly typed object, or throws. Never null.</returns>
+        /// </summary>
```
mikkelbu: I usually put `param`, `typeparam`, and `returns` outside the `summary` element (I don't know if this makes a difference or not). Also it is a bit inconsisten whether the the sentence ends with a period.

src/NUnitFramework/framework/Internal/ConstraintUtil.cs
```diff
@@ -0,0 +1,50 @@
+// ***********************************************************************
+// Copyright (c) 2015 Charlie Poole, Rob Prouse
```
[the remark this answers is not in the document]
josteink: Fair enough. Will fix.

src/NUnitFramework/framework/Internal/ConstraintUtil.cs
josteink: I agree that makes sense. Will fix.

josteink:
> Also, I really like the new method. Is there any reason why you did not use it in

Yes. I realized such a reusability concern could possibly arise in the future, but decided that this could be a good "base-commit" to land properly before making any changes other aspects of the framework.

There were also the case of poor tooling. I was fixing up the PR on my home machine running whatever version of mono ships with Fedora, and I couldn't get the whole build-chain running (I multiple causes of build-failure with the Linux-solution).

I therefore decided to be defensive about it, and not touch any more components than I felt I could guarantee the correctness of.

josteink:
I've pushed a final commit which should address most of the concerns @mikkelbu had.

ChrisMaddock:
image

@mikkelbu - do you not see "Dismiss review" at the bottom here? We should have the same access rights, I believe.

Merging - thanks @josteink! We'd be interested to learn about/fix your mono problems - if you'd fancy debugging and opening an issue? You can see the mono versions we currently test in CI in the .travis.yml file.

ChrisMaddock:
(Also, @mikkelbu - can I leave it with you to open an issue re: the additional changes? Thanks!)

jnm2:
Ooh, race condition! I was going to request a change but we can do it later, it's internal and minor.

jnm2:
(`static` on the new class)

ChrisMaddock:
Ah, sorry! I was happy it was internal, so didn't read further down the line! 🙈

jnm2:
@ChrisMaddock That's not actually a bad strategy! :D

mikkelbu:
@ChrisMaddock  I don't think I've the same rights. I went through all the open PRs and all I got was the following
image
But accordingly to the guide (https://help.github.com/articles/dismissing-a-pull-request-review/) "repository administrators or **people with write access** can dismiss a review." I should have enough?

Ps. I'll create the new issue

ChrisMaddock:
@rprouse - could you investigate? Or @CharliePoole - if you're still an admin on the repo?

CharliePoole:
Too late to investigate this one, as it is merged! In general, I can see "Dismiss review" on both my own reviews and those of others. However, Rob and I are owners of the organization so that doesn't prove much.

ChrisMaddock:
Sorry - I meant could you investigate for any differences between mine and @mikkelbu's admin rights. AFAIK, they should be the same!

ChrisMaddock:
To be clear, currently: I can dismiss reviews and @mikkelbu can't. I believe we should have the same permissions for this!

CharliePoole:
I can't see any difference in GitHub. Two thoughts:

* @mikkelbu are you sure you were logged in?
* Could it be a browser issue? I use Firefox.

mikkelbu:
@CharliePoole I'm quite sure that I'm logged in. Also the box about "Checks and Change requests" in the bottom is not visible if one is not logged in (as far as I can tell from testing now). I've tried Chrome, IE, and Edge on all the open PRs in this project I see the same "See review", but no "Dismiss review".

CharliePoole:
As far as I can see, the only difference is that @ChrisMaddock is on the Core Team. But that doesn't grant any special rights as an admin or maintainer. @rprouse can you see anything?

PR #2490, opened 2017-10-13
Allow null values in CollectionOrderedConstraint

### LINKED ISSUE #1473: Allow Is.Ordered to Compare Null Values
If you try to assert that an enumerable containing a `null` value is ordered:

``` C#
Assert.That(new[] { null, "Hello", "World" }, Is.Ordered);
```

An exception is thrown:

```
System.ArgumentNullException: Null value at index 0
```

This is because the `CollectionOrderedConstraint` [explicitly checks for `null` values](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs#L134):

``` C#
if (obj == null)
    throw new ArgumentNullException("actual", "Null value at index " + index.ToString());
```

This seems like an unnecessary strict assertion. In [the MSDN documentation](https://msdn.microsoft.com/en-us/library/system.icomparable.compareto.aspx#Anchor_2), it states that:

> By definition, any object compares greater than (or follows) **null**, and two null references compare equal to each other.

This explicitly says that `null` is before anything else, and implicitly that `null` values are comparable and allowed.

I understand that `null` values are often not desired, but they are a part of C#'s type system and it should be allowed to check if `null` references are ordered – at least using a setting if the default behavior is to disallow them:

``` C#
Assert.That(new[] { null, "Hello", "World" }, Is.Ordered.AllowingNull); // Should not thrown an exception
```

(this is merely the best name I could come up with on the fly :wink:)

pknouff:
Allows null values for simple ordering in `CollectionOrderedConstraint`. Null sorts less than everything else. An `ArgumentNullException` will still be thrown if ordering by a property.

This is a potentially breaking change. If the previous behavior of throwing an exception if a collection contains null items is desired, it can be achieved using `Has.None.Null.And.Ordered`.

Fixes #1473

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
@@ -192,13 +192,13 @@ namespace NUnit.Framework.Constraints
             int index = 0;
             foreach (object current in actual)
             {
-                if (current == null)
-                    throw new ArgumentNullException(nameof(actual), "Null value at index " + index.ToString());
-
                 if (previous != null)
                 {
                     if (_steps[0].PropertyName != null)
                     {
+                        if (current == null)
+                            throw new ArgumentNullException(nameof(actual), "Null value at index " + index.ToString());
+
                         foreach (var step in _steps)
                         {
                             string propertyName = step.PropertyName;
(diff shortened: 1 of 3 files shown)


jnm2:
@pknouff I've restarted the job that timed out. It wasn't you!

jnm2:
Good tests; thanks for doing this!

Issue #1373, opened 2016-03-24
Setting with a null value


oznetmaster:
I happened to add a setting to runSettings that had a valid value of null.

Everything worked correctly until the results were written.  

FrameworkController.AddSetting throws an exception if the value of a setting is null.

This may be unique to my use of runSettings, but I still think that AddSetting should not crash on a null setting value?

rprouse:
Can you provide a repro for this? I am going to mark it as a bug and put it in the backlog, but an easy repro would help newer developers fix this.

OmicronPersei:
Closed by merge #2493 .

PR #2493, opened 2017-10-15
Fix for Issue #1373 - Setting with a null value

matt40k:
Fix for Issue #1373

### DIFF
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -438,15 +438,22 @@ namespace NUnit.Framework.Api
             TNode setting = new TNode("setting");
             setting.AddAttribute("name", name);
 
-            var dict = value as IDictionary;
-            if (dict != null)
+            if (value != null)
             {
-                AddDictionaryEntries(setting, dict);
-                AddBackwardsCompatibleDictionaryEntries(setting, dict);
+                var dict = value as IDictionary;
+                if (dict != null)
+                {
+                    AddDictionaryEntries(setting, dict);
+                    AddBackwardsCompatibleDictionaryEntries(setting, dict);
+                }
+                else
+                {
+                    setting.AddAttribute("value", value.ToString());
+                }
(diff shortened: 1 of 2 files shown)


CharliePoole:
Your fix looks reasonable but nobody has ever provided a repo of the problem. Can you add a test that fails without your fix and passes after it?

My biggest concern would be that we not mask a problem in the calling code. How do we get a null here?

ChrisMaddock:
I agree re: this wanting a test. What does this actually look like, when you try and write null with `XmlWriter.WriteAttributeValue`?

> My biggest concern would be that we not mask a problem in the calling code. How do we get a null here?

I couldn't see a way we would require this currently - however, it's a situation @oznetmaster came across in production - and it seems sensible to me that this should work as expected, when it's only currently crashing on the result-writing stage.

matt40k:
Thanks for the feedback - so just to check. We're happy to handle nulls in this way - ie it should handle it and not throw an exception.

And I need to add a unit test(s) to src/NUnitFramework/tests/Api/FrameworkControllerTests.cs

Is this correct?

CharliePoole:
Yes to the first question.

As for tests, we need to test both ways that the framework can be invoked: through `FrameworkController` and directly through `NUnitTestAssemblyRunner` as NUnitLite does it. I think the second test may lead you to either moving or duplicating your fix.

mikkelbu:
Hi @matt40k the new test is failing at the build server
https://ci.appveyor.com/project/CharliePoole/nunit/build/3.9.0-ci-04583-pr-2493#L1777.

matt40k:
Mmm, not sure why its returning empty rather then null

jnm2:
Probably because it's writing the XML `key="key1" value=""` when you give it `null`.
The usual way to preserve the null would be to prevent the `value ` attribute from being shown, but I'm not sure `null` vs empty is important in run settings?

rprouse:
@matt40k just update your tests to look for the empty string. I am pretty sure @jnm2 is correct.

PR #2495, opened 2017-10-16
Make ConstraintUtils static

ChrisMaddock:
Adds the missing static keyword @jnm2 spotted.


jnm2:
Ooh can we make it `Utils` too? I know we have a `Util` but it's outnumbered by plural forms in our various projects.

ChrisMaddock:
We can, but I'll need to pull it down for that - I did this on the web UI. Will get back to it. 😄

jnm2:
Can I do it? Easy where I'm at.

jnm2:
Wondered why you were so fast! :D

jnm2:
I went ahead and did it so see how well VS Code refactors. Faster than opening full VS. You can drop the commit if you don't want to rename, but I figured I may as well not waste it.

ChrisMaddock:
That's a great help - thanks! 😄

Issue #2494, opened 2017-10-16
CollectionAssert.AllItemsAreUnique() very slow


rvhuang:
Consider following simple test case:

```csharp
// using NUnit.Framework;
        [TestCase(10000)]
        public void TestMethod1(int range)
        {
            CollectionAssert.AllItemsAreUnique(Enumerable.Range(0, range).ToArray());
        }
```
it takes about 14 seconds to complete.

However, applying MS Test Framework to same test case, it takes about 12 ms only to finish.

```csharp
// using Microsoft.VisualStudio.TestTools.UnitTesting;
        [TestMethod]
        public void TestMethod1()
        {
            CollectionAssert.AllItemsAreUnique(Enumerable.Range(0, 10000).ToArray());
        }
```

The gap between two should not be that huge.

ChrisMaddock:
NUnit uses a different way to test for uniqueness, than MSTest.

MSTest uses simple object equality. NUnit equality tests go through [NUnitEqualityComparer ](https://github.com/nunit/nunit/blob/e7c7826da8b064cfcc518b1035e1c270588abc79/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs)- which has extra complexity.

Looking at that class however, there may be low hanging fruit. Before comparing for object equivalence, we create an empty List of FailurePoint objects, which doesn't seem to be used. We then do two null checks prior to object.ReferenceEquals(), which I think are unnecessary. 

I'll dig in to this one more later.

rvhuang:
@ChrisMaddock Thanks for looking into the issue. We have a test case which requires to check uniqueness of 10k items and it cannot finish within one minute. The performance improvement shall help us a lot.

CharliePoole:
@ChrisMaddock I'm not looking at the code but I believe we accept an IEnumerable rather than an ICollection. You. Might be able to create a special case for collections. Of course we are also doing a lot of boxing in the case of an int value.

ChrisMaddock:
@CharliePoole - we do, but I'm not sure what sort of special case you're thinking of? Would love to hear, if you get a chance to look later!

CharliePoole:
A simple thing to try would be (in `Matches`)

```C#
IList list = actual as IList ?? new List<object>(actual);
```

That way, if an IList were passed in, we would not create a new one.

Following that (if tests showed it necessary) I'd use a generic helper to eliminate boxing. I'm busy on parallelization stuff or I'd give it a try myself.

[...]

PR #2496, opened 2017-10-16
Added overload for 'CollectionAssert' with IEqualityComparer.

dafenix:
Useful for comparisons of collections with slightly different objects. I hope you find it useful.

with kind regards
Daniel

### DIFF
--- a/src/NUnitFramework/framework/CollectionAssert.cs
+++ b/src/NUnitFramework/framework/CollectionAssert.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.ComponentModel;
 using NUnit.Framework.Constraints;
 
@@ -212,6 +213,17 @@ namespace NUnit.Framework
         {
             Assert.That(actual, Is.EquivalentTo(expected), message, args);
         }
+
+		/// <summary>
+		/// Asserts that expected and actual are equivalent, containing matching objects but the match may be in any order.
+		/// </summary>
+		/// <param name="expected">The first IEnumerable of objects to be considered</param>
+		/// <param name="actual">The second IEnumerable of objects to be considered</param>
+		/// <param name="equalityComparer">The EqualityComparer which will be used for matching</param>
+		public static void AreEquivalent<T>(IEnumerable<T> expected, IEnumerable<T> actual, IEqualityComparer<T> equalityComparer)
+		{
+			Assert.That((object)actual, new CollectionEquivalentConstraint(expected).Using(equalityComparer));
+		}
         #endregion
 
         #region AreNotEqual


CharliePoole:
I'd prefer not to add this. We have always treated `CollectionAssert` as a stable, legacy API, created to match the Microsoft class of the same name for those who like the syntax. Everything one can do with `CollectionAssert` and more can be done using `Assert.That`.

That said, I don't ,now how the Microsoft `CollectionAssert` has evolved since I created our class. I'd be open to a review of the classes with some overall decisions about where it should go before making a final call on this PR. In particular, if we are adding such an overload to `AreEquivalent` users will expect it on other members as well.

Marking this as an idea till we have a concensus on overall direction.

CharliePoole:
I made this a discussion item as well and assigned myself to facilitate the discussion. I'll remove myself the moment we make it a backlog item. All this is a little hard to follow on a PR, of course, which is why we generally like to have issues before we have code. However, let's give it a try.

jnm2:
@CharliePoole This gives us an immediate fix for https://github.com/nunit/nunit/issues/2494 though. We just use a HashSet and rely on the hashes generated by the IEqualityComparer that is passed in. Then we tell people with a performance problem to provide their own IEqualityComparer instance and voila, we're as fast as the BCL itself.

CharliePoole:
@jnm2 This has nothing to do with #2494, although a similar change could be made for that issue. I'd raise the same objection, however.

There is already a workaround for people who want to do this, which is to use `Assert.That`. I think we need to decide what our policy toward `CollectionAssert` really is before embellishing it.

ChrisMaddock:
We need to be careful with that - we made changes a couple of releases ago so that CollectionAssert already behaves _differently_ to Assert-Is-Equal.

https://github.com/nunit/nunit/issues/2097#issuecomment-292729733

CharliePoole:
@ChrisMaddock But issue #2494, brought up here by @jnm2, doesn't have to do with equality. It relates to the `UniqueItemsConstraint`, which works the same way whether you invoke it from `Assert.That` or from `CollectionAssert`.

CharliePoole:
@nunit/framework-team Are you happy discussing the question "Should we add any enhancements to CollectionAssert?" here or would you prefer a separate issue?

jnm2:
> This has nothing to do with #2494,

> doesn't have to do with equality

#2494 has to do with IEqualityComparer because without a hash code generator we have an inherent time complexity performance problem. This issue proposes to give us a hash code generator which in turn gives an obvious fix for #2494. Plus, uniqueness is concerned with equality.

jnm2:
We don't have to affect `UniqueItemsConstraint` to use the API in this issue to fix #2494.

CharliePoole:
@jnm2 Sorry, I'm not seeing what you mean. How can adding an overload to `CollectionConstraint.AreEquvalent` fix performance in a test that uses `CollectionConstraint.AllItemsAreUnique`? If you are saying a similar change could fix the latter, I agree and already said so - although I don't want to do it. I would rather fix the initial problem rather than giving users a tool to create a workaround, especially since it's trivial to do so.

CharliePoole:
@jnm2 This is a seriously odd way to discuss #2694!

However... If we want `CollectionConstraint.AllItemsAreUnique` to work differently, we have to either change it or change the underlying UniqueItemsConstraint. 

Please lets stop discussing #2694 here. I recognize that I answered your reference to it and contributed to everyone else's confusion so I'm going to stop.

jnm2:
@CharliePoole Oh! I erroneously thought he was talking about adding an overload to `AllItemsAreUnique`. You are quite right and I'm only causing confusion. And, I agree with you that we should direct people away from CollectionAssert.

jnm2:
> This is a seriously odd way to discuss #2694!

Lol! It certainly is! I'm so bad at this.

rprouse:
It has been our position for quite awhile now that we are not going to be enhancing the legacy asserts. The legacy asserts suffer from the fact that we need to add overloads or new methods for different usages like this whereas the constraint syntax can be combined. This code is functionally equivalent to `Is.EquivalentTo().Using()`, so it doesn't add new behavior, just another way to do the same thing.

@CharliePoole has raised that we might want to continue to evolve to match the MSTest API. This does make it easier to migrate, but I don't believe that NUnit should just be MSTest plus 😄 

Thanks for your contribution, but I am going to close this.

PR #2501, opened 2017-10-17
Initial fix for #2494

CharliePoole:
This goes towards fixing issue #2494. I wanted to try some things, but I'm happy for somebody else to continue. Here's what I did...

1. Created a test using `Enumerable.Range(0, 10000)`. Initially it took a105 seconds to complete.
2. Modified the code to make comparisons in place. This reduced the time to 97 seconds, which is not satisfactory. That modified code is still in the repository as `LegacyMatches`. It's the best time I acheived without changing any semantics.
3. Created a new `Matches` method using a `HashSet`. It completed my test in 24 __milliseconds__! Unfortunately, there were semantic changes. 
4. Fixed the most obvious problem, that of ignoring case. The fix is a hack for sure, but it works. Other more subtle semantic changes remain. This is the version that's currently in the code.

The key semantic difference is that my code no longer uses `NUnitEqualityComparer` in order to detect uniqueness, where the original code did. Thinking about it, I'm not sure that's what we want in a test of unique items. We might prefer object identity if we were doing this from scratch. We'll have to decide how to carry this forward.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
@@ -34,7 +34,10 @@ namespace NUnit.Framework.Constraints
     /// </summary>
     public abstract class CollectionItemsEqualConstraint : CollectionConstraint
     {
-        private readonly NUnitEqualityComparer comparer = NUnitEqualityComparer.Default;
+        /// <summary>
+        /// The NUnitEqualityComparer in use for this constraint
+        /// </summary>
+        private NUnitEqualityComparer _comparer = NUnitEqualityComparer.Default;
 
         /// <summary>
         /// Construct an empty CollectionConstraint
@@ -47,6 +50,26 @@ namespace NUnit.Framework.Constraints
         /// <param name="arg"></param>
         protected CollectionItemsEqualConstraint(object arg) : base(arg) { }
 
+        #region Protected Properties
+
+        /// <summary>
+        /// Get a flag indicating whether the user requested us to ignore case.
+        /// </summary>
+        protected bool IgnoringCase
+        {
+            get { return _comparer.IgnoreCase; }
+        }
+        
+        /// <summary>
+        /// Get a flag indicating whether any external comparers are in use.
+        /// </summary>
+        protected bool UsingExternalComparer
+        {
+            get { return _comparer.ExternalComparers.Count > 0; }
+        }
+
+        #endregion
+
         #region Modifiers
 
         /// <summary>
@@ -56,7 +79,7 @@ namespace NUnit.Framework.Constraints
         {
             get
             {
-                comparer.IgnoreCase = true;
+                _comparer.IgnoreCase = true;
                 return this;
             }
         }
@@ -68,7 +91,7 @@ namespace NUnit.Framework.Constraints
         /// <returns>Self.</returns>
         public CollectionItemsEqualConstraint Using(IComparer comparer)
         {
-            this.comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
@@ -79,7 +102,7 @@ namespace NUnit.Framework.Constraints
         /// <returns>Self.</returns>
         public CollectionItemsEqualConstraint Using<T>(IComparer<T> comparer)
         {
-            this.comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
@@ -90,7 +113,7 @@ namespace NUnit.Framework.Constraints
         /// <returns>Self.</returns>
         public CollectionItemsEqualConstraint Using<T>(Comparison<T> comparer)
         {
-            this.comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
@@ -101,7 +124,7 @@ namespace NUnit.Framework.Constraints
         /// <returns>Self.</returns>
         public CollectionItemsEqualConstraint Using(IEqualityComparer comparer)
         {
-            this.comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
@@ -112,13 +135,13 @@ namespace NUnit.Framework.Constraints
         /// <returns>Self.</returns>
         public CollectionItemsEqualConstraint Using<T>(IEqualityComparer<T> comparer)
         {
-            this.comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
         internal CollectionItemsEqualConstraint Using(EqualityAdapter adapter)
         {
-            comparer.ExternalComparers.Add(adapter);
+            _comparer.ExternalComparers.Add(adapter);
             return this;
         }
 
@@ -130,7 +153,7 @@ namespace NUnit.Framework.Constraints
         protected bool ItemsEqual(object x, object y)
         {
             Tolerance tolerance = Tolerance.Default;
-            return comparer.AreEqual(x, y, ref tolerance);
+            return _comparer.AreEqual(x, y, ref tolerance);
         }
 
         /// <summary>
@@ -139,7 +162,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="c">The collection to be included in the tally</param>
         protected CollectionTally Tally(IEnumerable c)
         {
-            return new CollectionTally(comparer, c);
+            return new CollectionTally(_comparer, c);
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
@@ -37,7 +37,7 @@ namespace NUnit.Framework.Constraints
         /// <summary>
         /// The NUnitEqualityComparer in use for this constraint
         /// </summary>
-        private NUnitEqualityComparer _comparer = NUnitEqualityComparer.Default;
+        private readonly NUnitEqualityComparer _comparer = NUnitEqualityComparer.Default;
 
         /// <summary>
         /// Construct an empty CollectionConstraint
--- a/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
@@ -147,8 +147,8 @@ namespace NUnit.Framework.Constraints
 
             return type.IsArray
                 || typeof(IEnumerable).IsAssignableFrom(type) // Covers lists, collections, dictionaries as well
-                || type == typeof(System.IO.Stream)
-                || type == typeof(System.IO.DirectoryInfo)
+                || typeof(System.IO.Stream).IsAssignableFrom(type) // Covers all streams
+                || typeof(System.IO.DirectoryInfo).IsAssignableFrom(type) // Unlikely to be derived, but just in case
                 || type.FullName == "System.Tuple"
                 || type.FullName == "System.ValueTuple";
         }
--- a/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
@@ -79,7 +79,7 @@ namespace NUnit.Framework.Constraints
             // If IEnumerable<T> is not implemented exit,
             // Otherwise return value is the Type of T
             Type memberType = GetGenericTypeArgument(actual);
-            if (memberType == null)
+            if (memberType == null || IsHandledSpeciallyByNUnit(memberType))
                 return null;
 
             string methodName = nameof(ItemsUnique);
@@ -99,9 +99,9 @@ namespace NUnit.Framework.Constraints
                     methodName = nameof(CharsUniqueIgnoringCase);
                     makeGeneric = false;
                 }
-        }
+            }
 
-        MethodInfo method = GetType().GetMethod(methodName, BindingFlags.Static | BindingFlags.NonPublic);
+            MethodInfo method = GetType().GetMethod(methodName, BindingFlags.Static | BindingFlags.NonPublic);
             if (method == null)
                 throw new InvalidOperationException("Internal error - Method not found: " + methodName);
             
@@ -156,6 +156,19 @@ namespace NUnit.Framework.Constraints
             }
 
             return true;
+        }
+
+        // Return true if NUnitEqualityHandler has special logic for Type
+        private static bool IsHandledSpeciallyByNUnit(Type type)
+        {
+            if (type == typeof(string)) return false; // even though it's IEnumerable
+
+            return type.IsArray
+                || type == typeof(IEnumerable) // Covers lists, collections, dictionaries as well
+                || type == typeof(System.IO.Stream)
+                || type == typeof(System.IO.DirectoryInfo)
+                || type.FullName == "System.Tuple"
+                || type.FullName == "System.ValueTuple";
         }
 
         private Type GetGenericTypeArgument(IEnumerable actual)
--- a/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
@@ -82,34 +82,22 @@ namespace NUnit.Framework.Constraints
             if (memberType == null || IsHandledSpeciallyByNUnit(memberType))
                 return null;
 
-            string methodName = nameof(ItemsUnique);
-            bool makeGeneric = true;
-
             // Special handling for ignore case with strings and chars
             if (IgnoringCase)
             {
                 if (memberType == typeof(string))
-                {
-                    methodName = nameof(StringsUniqueIgnoringCase);
-                    makeGeneric = false;
-                }
-                else
-                if (memberType == typeof(char))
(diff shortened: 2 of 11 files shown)


ChrisMaddock:
As expressed on the issue - I personally consider this too much of a breaking change to justify the big perfomance benefits - but others may disagree!

CharliePoole:
@ChrisMaddock As noted in the description, this is not proposed for merging. I was assuming you - as the original assignee - would take it on from this point. 😈 

That said, your comments on the issue gave me an idea, so I'll take it through another commit. BTW, note that the "LegacyMatch" method, when I used it gave about a 6% improvement in speed over the original code.

CharliePoole:
I managed to get a dictionary to use `NUnitEqualityComparer` but it didn't give me any performance improvement. Two reasons:

1. The comparer's use of objects, as @ChrisMaddock already noted
2. The fact that there is no override of GetHashCode that works correctly with the equality comparer. I was forced to return the same hash for every object so that NUnitEqualityComparer would be called.

I'm noting this as a dead end unless we do a major rewrite of NUnitEqualityComparer.

CharliePoole:
I did a `push -f` after starting over. The current code runs the test of 20000 ints many orders of magnitude faster (26 ms). It only uses the faster algorithm when the argument is `IEnumerable<int>`. This works, because `NUnitEqualityComparer` treats ints normally except when a tolerance is specified, and it's not possible to specify a tolerance with `UniqueItemsConstraint`.

We could expand this approach to `IEnumerable<T>` where T is any of the fundamental types plus string.

@ChrisMaddock What do you think about this approach?

ChrisMaddock:
I think this is a good stop gap. Two comments:

1. The principles good, but the ints were just an example, @rvhuang's actual issue was with strings. Could we add the second special case?
2. Is it correct that there's no special handling for two ints, or two strings in NUnitEqualityComparer? I think this is good to merge if there's no change in behaviour at all - if there is anything, I'd like to understand it.

CharliePoole:
@ChrisMaddock 
1. Yes, that's easy enough to do. We can add as many exceptions as we want, although beyond two or three the code gets cluttered and we might want to extract it all to a method.

2. It's not so much a question of what `NUnitEqualityComparer` does but of which features are accessible to the user with this particular constraint. That is, what can the user type after `Is.Unique` that effects how equality is determined? The answer is two things: `IgnoreCase`, which only affects strings, and `Using`, which can affect any type at all.

So my special case for `IEnumerable<int>` is actually invalidated if the user specified `Using`. A special case for `string` would also need to handle `IgnoreCase`.

I'll take another pass with int and string and try to deal with `Using` by falling back to the original code.

CharliePoole:
Latest refactor gets this into the shape I think works best. I still want to add more tests and make sure to eliminate `IEnumerable<T>` where T is handled specially by `NUnitEqualityComparer` from using the fast algorithm, but the general structure seems right.

CharliePoole:
I think this is now ready for review and merge. Fixes #2494.

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
```diff
@@ -47,6 +50,26 @@ public abstract class CollectionItemsEqualConstraint : CollectionConstraint
...
+            get { return _comparer.IgnoreCase; }
+        }
+        
+        /// <summary>
+        /// Get a flag indicating whether any external comparers are in use.
+        /// </summary>
+        protected bool UsingExternalComparer
+        {
+            get { return _comparer.ExternalComparers.Count > 0; }
+        }
```
jnm2: How about `protected bool UsingExternalComparer => _comparer.ExternalComparers.Count > 0;`? Same for IgnoreCase?

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
```diff
@@ -34,7 +34,10 @@ namespace NUnit.Framework.Constraints
     /// </summary>
     public abstract class CollectionItemsEqualConstraint : CollectionConstraint
     {
-        private readonly NUnitEqualityComparer comparer = NUnitEqualityComparer.Default;
+        /// <summary>
+        /// The NUnitEqualityComparer in use for this constraint
+        /// </summary>
+        private NUnitEqualityComparer _comparer = NUnitEqualityComparer.Default;
```
jnm2: Why are we removing the `readonly` modifier?

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -60,5 +69,123 @@ protected override bool Matches(IEnumerable actual)
...
+                {
+                    methodName = nameof(StringsUniqueIgnoringCase);
+                    makeGeneric = false;
+                }
+                else
+                if (memberType == typeof(char))
+                {
+                    methodName = nameof(CharsUniqueIgnoringCase);
+                    makeGeneric = false;
+                }
```
jnm2: I'd be much happier (and the code is faster) not using reflection for these two cases. Also there's a newline between `else` and `if`.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -60,5 +69,123 @@ protected override bool Matches(IEnumerable actual)
...
+            return true;
+        }
+
+        private static bool StringsUniqueIgnoringCase(IEnumerable<string> actual)
+        {
+            var hash = new HashSet<string>();
+
+            foreach (string item in actual)
+            {
+                string s = item.ToLower();
```
jnm2: We should not be using `.ToLower()` anywhere. It's expensive because it allocates a new string. The entire method can be made more general than hardcoding CurrentCultureIgnoreCase if we instead pass an `IEqualityComparer<>` to the HashSet constructor. In fact, this whole method can be removed if `ItemsUnique` starts accepting an `IEqualityComparer<T>` parameter.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -60,5 +69,123 @@ protected override bool Matches(IEnumerable actual)
...
+        {
+            if (type == typeof(string)) return false; // even though it's IEnumerable
+
+            return type.IsArray
+                || type == typeof(IEnumerable) // Covers lists, collections, dictionaries as well
+                || type == typeof(System.IO.Stream)
+                || type == typeof(System.IO.DirectoryInfo)
+                || type.FullName == "System.Tuple"
+                || type.FullName == "System.ValueTuple";
+        }
```
jnm2: How do we ensure that this list doesn't get stale?
Also `type == typeof(IEnumerable)` doesn't cover collections etc, but `typeof(IEnumerable).IsAssignableFrom(type)` does.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -60,5 +69,123 @@ protected override bool Matches(IEnumerable actual)
 
             return true;
         }
+
+        private bool? TryFastAlgorithm(IEnumerable actual)
+        {
+            // If the user specified any external comparer with Using, exit
+            if (UsingExternalComparer)
+                return null;
```
jnm2: Just a note: if the external comparer is an `IEqualityComparer<>`, we can use it with the `ItemsUnique` method below and be super fast.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -60,5 +69,123 @@ protected override bool Matches(IEnumerable actual)
...
+        {
+            var hash = new HashSet<T>();
+
+            foreach (T item in actual)
+            {
+                if (hash.Contains(item))
+                    return false;
+
+                hash.Add(item);
+            }
```
jnm2: You can get rid of one of the hash lookups by replacing the loop body with `if (!hash.Add(item)) return false`.

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
oznetmaster: Just because C# 7 allows the syntax does not necessarily make it better.

Many, including myself, find it obscure and difficult to maintain.

jnm2: @oznetmaster C# 7 is not allowed in the codebase yet, but luckily I'd never make that insupportable argument about C# 6 either. ;-)

oznetmaster: I guess it is a matter of policy/preference.  None of my programmers are allowed to use that compact syntax because of clarity and maintainability.

CharliePoole: I agree Neil in not liking that syntax. Let's just leave it as a question of style.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
CharliePoole: Correct. That could be another step in improving this, but it would involve analyzing all the external comparers to see if any of them applied and that effort would be wasted if we end up rewriting the NUnitEqualityComparer anyway. My goal in this PR was to do as much as I could while remaining inside the UniqueItemsConstraint. 

Since the issue is about `CollectionAssert`, we can never have either an external comparer or a case-insensitive comparison. You have to use `Is.Unique` to specify those.

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
jnm2: Yup! My reaction to the C# 1 version is pretty much the same as yours to the C# 6 one (minus the policy). 😊 Sorry Charlie, I think I remember now you saying you didn't like it, so no need to have brought it up.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
CharliePoole: Didn't know that!

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
CharliePoole: Understood, but it's what the existing `NUnitEqualityComparer` does.

However, I do like your idea, subject to my time limitations as noted.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
CharliePoole: I don't think staleness is a big problem, since any change in the definition of NUnit equality would be a pretty big breaking change.

Regarding `type == typeof`` you are right. Not sure if I can use `IsAssignableFrom` in all our builds, but I'll give it a shot.

jnm2: We added ValueTuple and later Tuple and if this list had existed back then and Mikkel hadn't run across it, I'm imagining it being missed in review and indefinitely.

I think importing the reflection compatibility namespace will handle `IsAssignableFrom`.

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
CharliePoole: It changed to a property and then back again - forgot to re-add it.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
CharliePoole: Let me take a look again. An earlier version required it but I guess I got rid of that need and didn't notice. That said, there's a limit on how much time I want to spend on this one and  the slower code is about 100 times faster than what we had before.

rprouse:
@jnm2 you have requested changes. Is there anything outstanding or is it resolved? I don't want to dismiss your review prematurely.

jnm2:
I should have seen this along with `IEnumerable`- we need `typeof(System.IO.Stream).IsAssignableFrom(type)` too.

CharliePoole:
Hopefully, this is only a temporary thing and we will eventually be able to use NUnitEqualityComparer without sacrificing efficiency. In any case, this is all the time I want to devote to the issue of uniqueness, which is probably not our most used feature. 😸

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
oznetmaster: I agree that IEqualityComparer<> should be used for string and char to handle the case sensitivity.  It is significantly faster.  StringComparer already exists, and it is easy to write a CharComparer that does the same.  

It does not seem that it would be a breaking change.

CharliePoole: As I said, I like it too. I'm out of time for this issue, however.

oznetmaster: I may have missed this, but is using the CurrentCulture, which is the default for string.ToLower(), the correct thing to do, or should string.ToLowerInvariant() or string.ToLower (CultureInfo.InvariantCulture) be used?

jnm2: NUnitEqualityComparer currently uses CurrentCultureIgnoreCase if you set `.IgnoreCase` and (iirc) Ordinal if you don't.

oznetmaster: Then it appears that this is doing the same for the ignore case situation.  Sorry for the misunderstanding.

CharliePoole: NUnitEqualityComparer uses simple ToLower just as I did in this code. That's why I did it that way. Main objective is to produce the same results as NUnitEqualityComparer but more quickly.

oznetmaster:
Well, this is really strange.

Running the two new PerformanceTests on my CF build, they all pass except for the List<int> tests!

This is 100% consistent.

IEnumerable<int>, and all the rest, pass at less then the 100 msec specified, but List<int> takes 200 msec to 400 msec to run.

I have checked, and the tests are all using the FastAlgorithm.  The only apparent difference between the two int tests is that one is iterating over an IEnumerable<int>, and the other over a List<int>.

However, the non-int tests also iterate over List<T>, and they pass.

This makes no sense to me.

I wonder if this difference is actually happening on the faster processors as well, but in those cases, they all still complete within the 100 msec?  My test are running on a 800 Mhz ARM processor.

Any thoughts?

oznetmaster:
Looking more closely, the elapsed time for the failing test is about 50x the elapsed time for all the passing tests!

CharliePoole:
I saw similar differences I couldn't explain, and so set the max time about three times higher than my worst case. The original algorithm was using around 100,000 Ms!

It's really not possible to do anything but extremely rough performance testing using MaxTime.

jnm2:
@oznetmaster So `List<int>` is 50x slower than `List<double>`, `List<string>`, and `RangeOfInt`?

jnm2:
Yep, look at that!  I would have expected RangeOfInt to have the only overhead, but my i7-7700k shows the same thing! In fact I can't even get the other `List<>` tests to register, while `List<int>` gets really slow if I multiply the range by 1000.

Now I'm getting OutOfMemoryExceptions, so something is allocating.

@oznetmaster What's the native word size on the ARM chip?

jnm2:
`List<long>` acts the same. The difference seems to be whether `List<>` is constructed from a bare `RangeOfInt` or whether you chain a `.Select` after it.

oznetmaster:
I am not happy we have this weird bug, but I am glad that everyone can see it regardless of their environment.

Now, the question is, how do we find it?

oznetmaster:
The ARM chip I am working with is 32-bits, I believe.

oznetmaster:
So, I tried:
```
		static readonly IEnumerable[] PerformanceData =
        {
            RANGE,
            new List<int>(RANGE),
            new List<int>(RANGE.Select(v => v)),
            new List<short>(RANGE.Select(v => (short)v)),
            new List<double>(RANGE.Select(v => (double)v)),
            new List<string>(RANGE.Select(v => v.ToString()))
        };
```

And got:
115 msec
485 msec
4 msec
13 msec
12 msec
8 msec

This really makes no sense :(

oznetmaster:
Reran it and got:
6
474
4
5
5
6

CharliePoole:
To avoid interference from other tests, we should probably mark the fixture as NonParallelizable.

jnm2:
I'm back. Oh, heh. If I redefine RANGE as `Enumerable.Range` instead of `RangeOfInt`, they all get slow. There has to be caching going on.

jnm2:
Aha! Replace the test with this, and the only one test passes; the others fail with 0 items:

```cs
        [TestCaseSource(nameof(PerformanceData))]
        [NonParallelizable]
        public void PerformanceTests(IEnumerable values, bool ignoreCase)
        {
            Assert.That(values, Has.Exactly(10000).Items);
        }
```

jnm2:
Oh duh, I missed the obvious. All tests are sharing the same IEnumerable instance and the current count is being stored in the enumerable instead of the enumerator.

This is what it should be:
public class RangeOfInt : IEnumerable<int>
{
    private readonly int _start;
    private readonly int _count;

    public RangeOfInt(int start, int count)
    {
        _start = start;
        _count = count;
    }

    public IEnumerator<int> GetEnumerator()
    {
        for (var i = 0; i < _count; i++)
        {
            yield return _start + i;
        }
    }

    IEnumerator IEnumerable.GetEnumerator()
    {
        return GetEnumerator();
    }
}

jnm2:
The entire implementation could have also been:

```cs
    public static class Range
    {
        public static IEnumerable<int> OfInt(int start, int count)
        {
            for (var i = 0; i < count; i++)
                yield return start + i;
        }
    }
```

jnm2:
Who wants to do the PR? I will if everyone is busy.

CharliePoole:
I'm back on parallelism and would rather stay there. Looking at a trip coming up, so I want to wrap up.

jnm2:
No problem. I'll take it.

Issue #2331, opened 2017-07-27
Repo does not build in VS without running `build -t build` first


jnm2:
For example, nunit.frameworktest-4.5 has 35 errors, all centered on ValueTuple not being found.

This afflicts every fresh clone and anyone who had a clone before https://github.com/nunit/nunit/pull/2290.

It looks like some type of NuGet restore problem.

@rprouse We don't need packages.config for VS2015 support, so why do the test projects have both `packages.config` and `project.json` files?

jnm2:
Here is something we should consider including in our build.ps1, even though it's not worth running on every test run:

```c#
Task("FreshCloneBuildTest")
    .Description("Ensures that we don't accidentally introduce a change that prevents new clones from building without build.cake.")
    .Does(() =>
    {
        var tempDir = Directory(System.IO.Path.Combine(System.IO.Path.GetTempPath(), System.IO.Path.GetRandomFileName()));
        CreateDirectory(tempDir);
        try
        {
            StartProcess("git", new ProcessSettings().WithArguments(a => a.Append("clone . ").AppendQuoted(tempDir)));
            MSBuild(tempDir + File("nunit.sln"));
        }
        finally
        {
            DeleteDirectory(tempDir, true);
        }
    });
```

jnm2:
This currently gives a ton of NuGet restore warnings and these errors:
```
C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\MSBuild\15.0\Bin\amd64\Microsoft.Common.CurrentVersion.targets(1964,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "NUnit.System.Linq, Version=0.6.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb, processorArchitecture=MSIL". Check to make sure the assembly exists on disk. If this reference is required by your code, you may get compilation errors. [C:\Users\Joseph\AppData\Local\Temp\dpyw0pqs.bhr\src\NUnitFramework\framework\nunit.framework-2.0.csproj]
Api\FrameworkController.cs(25,14): error CS0234: The type or namespace name 'Linq' does not exist in the namespace 'System' (are you missing an assembly reference?) [C:\Users\Joseph\AppData\Local\Temp\dpyw0pqs.bhr\src\NUnitFramework\framework\nunit.framework-2.0.csproj]
[...]

PR #2508, opened 2017-10-19
Correct divergence from shadowed Is / Has members.

BobSammers:
`InRange()` and `Exactly()` in `AssertionHelper` mimic (in fact facade in the case of `Exactly()`) members of the `Is` and `Has` classes, respectively. These members have recently had their signatures altered in `Is` and `Has` but the corresponding signatures in `AssertionHelper` were not amended. This change fixes that. `AH` is being deprecated, but I'm assuming this will count as maintenance.

Impetus: this change is referenced in #1212. As `AssertionHelper` is obsolete and is likely to be removed, `NUnit.StaticExpect` has been produced to help users migrate away from it with minimal changes (remove inheritance from `AssertionHelper`, add `using static NUnit.StaticExpect`). Tests in the `StaticExpect` project attempt to guarantee complete coverage of the original `AssertionHelper` classes by reflecting on `AH`. These tests have highlighted this discrepancy and fixing it should provide correctness within NUnit and will also allow the `StaticExpect` tests to pass without special-casing.

I should point out that I haven't compiled or tested these changes because I wasn't able to easily compile NUnit (#2331). That said, the proposed change merely replicates the underlying behaviour (`AssertionHelper.Exactly()` calls `Has.Exactly()` which already returns an `ItemsConstraintExpression`; `AssertionHelper.InRange()` returns a new `RangeConstraint` whose ctor already takes a pair of `object`s rather than `IComparable`s).

### DIFF
--- a/src/NUnitFramework/framework/AssertionHelper.cs
+++ b/src/NUnitFramework/framework/AssertionHelper.cs
@@ -217,7 +217,7 @@ namespace NUnit.Framework
         /// the following constraint to all members of a collection,
         /// succeeding only if a specified number of them succeed.
         /// </summary>
-        public static ConstraintExpression Exactly(int expectedCount)
+        public static ItemsConstraintExpression Exactly(int expectedCount)
         {
             return Has.Exactly(expectedCount);
         }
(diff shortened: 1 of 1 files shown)


ChrisMaddock:
Also, thanks for the contribution, Bob! It's normal practice that PRs to the framework repo are reviewed by two committers before merge - so this will also need to be a second person to take a look.

CharliePoole:
This looks good to me as well. We can merge when CI passes.

Thanks for doing this!

ChrisMaddock:
Merged - thanks @BobSammers!

BobSammers:
No problem - thank you very much for accepting and getting onto it so quickly!

Issue #2245, opened 2017-06-13
Make the delegate for TextContext.AddFormatter<TSUPPORTED> generic


appel1:
Since AddFormatter checks the type it'd make the formatter callback simpler.

CharliePoole:
Seems like a good idea, but do we need this issue plus #2246 ?

appel1:
I think it would be helpful.

CharliePoole:
I agree. I'm asking why this is not a duplicate of #2246 ?

appel1:
If the delegate was generic you could use it like this instead:
```
TestContext.AddFormatter<SomeType>(value => value.SpecialFormattingMethod());
```

Instead of:
```
TestContext.AddFormatter<SomeType>( value => ((SomeType)value).SpecialFormattingMethod());
```

But not a big deal, just a little thing I thought worth mentioning that could simplify usage.

appel1:
Perhaps it would also make it a little bit more obvious that the delegate will only be called with an instance of the specified type as argument.

CharliePoole:
We are missing one another. 😄 I understand both issues. Aren't they both part of one issue, with one solution? This issue appears to be at least a partial solution for the problem in #2246.

I see why #2244 is separate, although I'd probably fix it in the same PR as the other two, but #2245 and #2246 seem to be looking at the same problem from two different perspectives.

CharliePoole:
@appel1 "Perhaps it would..." What is "it" in that sentence? If it's this (#2245) issue, there's no doubt that it's a very good idea. My doubts were about #2246.

appel1:
This one is for the signature of the delegate to make using AddFormatter easier. #2246 is about making it possible to implement the callback to recursively take advantage of the existing formatters, in my current specific case (KeyValuePair) that would be handled by #2244 but we have other classes not part of the framework that could take advantage of that. 

Perhaps I'm just too tired to be able express myself coherently. :)

CharliePoole:
@appel1 No, I think it's me not expressing my question clearly.

First... thanks for the explanation... I did understand the details of what you were asking for. I was more on the higher level of whether we should manage these things as separate issues or just unify them and make one fix. Let's just keep them separate now and do a single PR for both of them if that seems appropriate to whomever takes them on.

rprouse:
Closing as won't fix because the change will break the existing API. See the discussion in #2509

PR #2509, opened 2017-10-21
Make TestContext.AddFormatter more generic

Breemua:
This is a proposal to fix issue #2245

Tell me if we should make ValueFormatter delegate generic, and get rid of the associated "object" parameters. It would be more invasive while looking the same to callers.

### DIFF
--- a/src/NUnitFramework/framework/TestContext.cs
+++ b/src/NUnitFramework/framework/TestContext.cs
@@ -329,9 +329,9 @@ namespace NUnit.Framework
         /// </summary>
         /// <typeparam name="TSUPPORTED">The type supported by this formatter</typeparam>
         /// <param name="formatter">The ValueFormatter delegate</param>
-        public static void AddFormatter<TSUPPORTED>(ValueFormatter formatter)
+        public static void AddFormatter<TSUPPORTED>(Func<TSUPPORTED, string> formatter)
         {
-            AddFormatter(next => val => (val is TSUPPORTED) ? formatter(val) : next(val));
+            AddFormatter(next => val => (val is TSUPPORTED) ? formatter((TSUPPORTED) val) : next(val));
         }
 
 #endregion
--- a/src/NUnitFramework/tests/Constraints/MsgUtilTests.cs
+++ b/src/NUnitFramework/tests/Constraints/MsgUtilTests.cs
@@ -33,7 +33,10 @@ namespace NUnit.Framework.Constraints
     public static class MsgUtilTests
     {
         #region FormatValue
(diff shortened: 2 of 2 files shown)


CharliePoole:
Looks good to me

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -329,9 +329,9 @@ public static void AddTestAttachment(string filePath, string description = null)
         /// </summary>
         /// <typeparam name="TSUPPORTED">The type supported by this formatter</typeparam>
         /// <param name="formatter">The ValueFormatter delegate</param>
```
mikkelbu: This should be updated, as the argument is not a `ValueFormatter delegate` anymore

mikkelbu:
Wrt. making the "ValueFormatter delegate generic" then I think we should do that in another issue/PR, since we also need to decide how to handle `CurrentValueFormatter`.

CharliePoole:
I'm afraid I didn't look too deeply - just saw the comment that it would look the same to callers. If it doesn't, but instead breaks them, I question the need to change.

jnm2:
On the other hand, if we add this as an overload, it will break any code that passes a lambda or anonymous method and relies on implicit delegate allocation. Example of existing code that would break if we add an overload:

```cs
TestContext.AddFormatter<Foo>(obj => ((Foo)obj).Bar);
```

jnm2:
From my perspective, breaking people who pass a `ValueFormatter`-typed argument as the first argument is okay if we think that number is small enough and if we're willing to clearly communicate the break as such in the NuGet package manager UI and the release notes. It's certainly a cleanup that we should track for some future version if not the next one.

The simplest non-breaking option I can think of is to leave the original method alone and write a new method named `AddGenericFormatter`.

jnm2:
@nunit/framework-team We need to make a decision here. Are we willing to break code similar to my first example, break code similar to my second example, or come up with a brand new name from the method?

ChrisMaddock:
Good catch there Joseph. 😄 

I wouldn't want to add a newly named method - from the issue it looks like this was only ever a nice-to-have anyway. It sucks that the overload doesn't work - in that case, I think my preference would be to close as wont-fix. This seems to only offer a minor improvement, which doesn't justify breaking changes in my eyes.

rprouse:
Good eye @jnm2. I agree with @ChrisMaddock in that this is a minor improvement, so let's close as won't fix. @Breemua thanks for the work you did. I'm sorry we aren't going to take it, but I hope you come back and continue to contribute.

PR #2510, opened 2017-10-22
Issue1678 allow special chars in category

### LINKED ISSUE #1678: Result Message:	OneTimeSetUp: Category name must not contain ',', '!', '+' or '-'
@Andrey-Bushman commented on [Fri Jul 15 2016](https://github.com/nunit/nunit3-vs-adapter/issues/206)

I want to group my tests on the base of tested method signature (one test checks return values, and also exist the individual tests for each `out` parameter checking). For getting this result I use the `Category` attribute:

`[Category("LocalizedProduct.TryParse(String, out ILocalizedProduct)")]`

But tests which was marked by such attribute value are failed with the message:

> Result Message:   OneTimeSetUp: Category name must not contain ',', '!', '+' or '-'

Why this restriction exists?

I could be use this variant:

`[Category("LocalizedProduct.TryParse(String | out ILocalizedProduct)")]`

But the comma using would be _more legible_ for me at this case.

---

@CharliePoole commented on [Fri Jul 15 2016](https://github.com/nunit/nunit3-vs-adapter/issues/206#issuecomment-232997494)

We don't allow characters that have a special meaning in NUnit TSL (Test Selection Language). See https://github.com/nunit/docs/wiki/Test-Selection-Language

It may be that the exclusion of the comma is obsolete - I'll have to check further.

---

@CharliePoole commented on [Fri Jul 15 2016](https://github.com/nunit/nunit3-vs-adapter/issues/206#issuecomment-233078139)

An earlier version supported

```
--where "cat == A,B,C"
```

in place of the longer 

```
--where "cat==A||cat==B||cat==C"
```

We have an outstanding issue asking us to restore the shorter syntax. We would have to decide about that issue before it could be possible to change this.

I'm going to go ahead and move this issue to the nunit project where it belongs.

stevenaw:
Fixes #1678 

This is a bit of an intermediate point in my proposed fix for #1678 . It removes the check for special characters `","` and `"-"` in category names, along with adding some tests in various places that we use Category names. Based on what I can see in the console runner (and per [this comment](https://github.com/nunit/nunit/issues/1678#issuecomment-334960196) ) the other two special characters of `"!"` and `"+"` should also be OK. I'll be investigating that next.

This is my first real feature change to the NUnit internals, so I'm trying to take it slow and careful. I'd appreciate a review to see if I'm on the right track.

### DIFF
--- a/src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
+++ b/src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2014 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
--- a/src/NUnitFramework/framework/Attributes/CategoryAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/CategoryAttribute.cs
@@ -82,9 +82,6 @@ namespace NUnit.Framework
         public void ApplyToTest(Test test)
         {
             test.Properties.Add(PropertyNames.Category, this.Name);
-
-            if (this.Name.IndexOfAny(new char[] { '!', '+' }) >= 0)
-                test.MakeInvalid("Category name must not contain '!' or '+'");
         }
 
         #endregion
--- a/src/NUnitFramework/framework/Attributes/CategoryAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/CategoryAttribute.cs
@@ -83,8 +83,8 @@ namespace NUnit.Framework
         {
             test.Properties.Add(PropertyNames.Category, this.Name);
 
-            if (this.Name.IndexOfAny(new char[] { ',', '!', '+', '-' }) >= 0)
-                test.MakeInvalid("Category name must not contain ',', '!', '+' or '-'");
+            if (this.Name.IndexOfAny(new char[] { '!', '+' }) >= 0)
+                test.MakeInvalid("Category name must not contain '!' or '+'");
         }
 
         #endregion
--- a/src/NUnitFramework/nunitlite.tests/TestSelectionParserTests.cs
+++ b/src/NUnitFramework/nunitlite.tests/TestSelectionParserTests.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2015 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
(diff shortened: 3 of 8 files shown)


src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
```diff
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
```
stevenaw: EOL change I think?
Sorry, there's a few in this PR from files I've opened. Please let me know if I should revert.

jnm2: This is fine. It likely is removing the byte order marks. This is because people have saved all the files with random character encodings over the years and we recently added `charset = utf-8` to our .editorconfig to bring us toward consistency.

ChrisMaddock:
Looks 'on the right track' to me - thanks! 🙂

src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
stevenaw: Ah that makes sense. Ok cool, thanks for clarifying.

CharliePoole:
This looks OK by itself. However, the limitation on special characters arises from the needs of runners that have to parse TSL from the command-line. We need to ensure that categories with these special characters are usable from both the NUnitLite and Console command-lines before we can merge the change.

stevenaw:
That makes sense @CharliePoole thanks!
I've added a few more XML-oriented tests for the core framework. I'll start playing around with the NUnitLite  runner next.

If changes need to be made in tandem to the console runner and to the core framework, how does that typically work? Does one need to be updated before the other, or can it happen relatively independently?

CharliePoole:
It depends on the effect of having one updated without the other. In this case, the risk is either that the user creates categories that can't be specified in the DSL or that the DSL allows you to ask for categories, which can never be present. The second seems less risky to me.

I __believe__ that NUnitLite's parsing is currently the same as that for the console runner, so that's a good place to start. Since they are separate projects, it's possible for them to get out of sync, of course.

sirk1989:
Any idea whats happening with this? Are there any appropriate reviewers that can approve this? This is something i am pretty desperately going to need quite soon. Thanks.

jnm2:
I've been passing over this because the top says Do Not Merge. Is it ready for a final round of reviews?

sirk1989:
@jnm2 I missed that sorry.

@stevenaw anyidea why this is flagged as do not merge and when we can expect to see this completed?

jnm2:
@stevenaw If you don't have the time, I am happy to pick this up. You may have merge conflicts since our csproj migration and my head is still in that game, so I can handle it if you want.

jnm2:
Oh, except one typo. `TestWithVvalidCategoryNameIsNotRunnable`. @stevenaw Wanna fix that or should I close this PR and copy your branch to a PR I can push to?

CharliePoole:
I agree with merging this when ready.

stevenaw:
I've fixed this typo: https://github.com/nunit/nunit/pull/2510#issuecomment-350907377

Thanks @CharliePoole @ChrisMaddock and especially @jnm2  for helping out at the end

jnm2:
@stevenaw You did all the work! We really appreciate it.

jnm2:
@sirk1989 If you need an immediate workaround in a project, you can merge this with your project's nuget.config and upgrade NUnit to version 3.10.0-dev-05043 or later:

```xml
<configuration>
  <packageSources>
    <add key="NUnit MyGet" value="https://www.myget.org/F/nunit/api/v3/index.json" />
  </packageSources>
</configuration>
```

sirk1989:
@stevenaw and @jnm2 This is great thanks for the swift work guys.

PR #2512, opened 2017-10-23
Handling of native exceptions, to attempt to catch C++ errors …

### LINKED ISSUE #2091: When a native exception of corrupted state is thrown, nunit test thread crashes and the nunit-console process hangs

endrih:
Fixes #2091 . It is useful for cases when people are using NUnit in C++/CLI projects and a native exception of corrupt state occurs that crashes the testing thread. When the testing thread crashes, NUnit hangs as a result.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Reflect.cs
+++ b/src/NUnitFramework/framework/Internal/Reflect.cs
@@ -26,7 +26,9 @@ using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Reflection;
+#if !NET_2_0 && !NET_3_5 && !NETSTANDARD1_3 && !NETSTANDARD1_6
 using System.Runtime.ExceptionServices;
+#endif
 using NUnit.Compatibility;
 using NUnit.Framework.Interfaces;
 
@@ -199,7 +201,9 @@ namespace NUnit.Framework.Internal
         /// <param name="fixture">The object on which to invoke the method</param>
         /// <param name="args">The argument list for the method</param>
         /// <returns>The return value from the invoked method</returns>
+#if !NET_2_0 && !NET_3_5 && !NETSTANDARD1_3 && !NETSTANDARD1_6
         [HandleProcessCorruptedStateExceptions] //put here to handle C++ exceptions. 
+#endif
         public static object InvokeMethod( MethodInfo method, object fixture, params object[] args )
         {
             if(method != null)
--- a/src/NUnitFramework/framework/Internal/Reflect.cs
+++ b/src/NUnitFramework/framework/Internal/Reflect.cs
@@ -223,10 +223,6 @@ namespace NUnit.Framework.Internal
                 {
                     throw new NUnitException("Rethrown", e);
                 }
-                catch
-                {
-                    throw new NUnitException("Rethrown", new Exception("Unknown Native exception thrown."));
-                }
             }
 
             return null;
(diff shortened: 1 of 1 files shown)


jnm2:
Tip: you can edit the initial PR comment to say "Fixes #2091" or "Closes #2091" ([doc](https://help.github.com/articles/closing-issues-using-keywords/)) so that GitHub closes the issue when this PR is merged.

endrih:
Do I need to do anything else for this pull request so that it is merged into main? (it is the first time I do it here, so I don't know how the full process goes)

CharliePoole:
The fixes phrase should be in a comment, as I edited it, not the title.

@jnm2 approved the PR. It's waiting for a second team approval to be merged.

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -24,7 +24,11 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.ComponentModel;
```
mikkelbu: Major nit pick :). Is this necessary?

mikkelbu:
Merged. Thanks for the contribution @endrih 👍 .

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: Nice spot! I just had a thought– maybe we should leave some time for the contributor to respond if there are questions before merging? :D

mikkelbu: Actually, I did not find this important enough to hold back the PR, and perhaps I should just have ignored it :)

jnm2:
This PR has never had any effect because https://github.com/nunit/nunit/pull/1775 added `[assembly: AllowPartiallyTrustedCallers]` and the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.handleprocesscorruptedstateexceptionsattribute?view=netframework-4.8) say this means that `[HandleProcessCorruptedStateExceptions]` is ignored:

> The HandleProcessCorruptedStateExceptionsAttribute attribute is ignored when it is encountered in partially trusted or transparent code, because a trusted host should not allow an untrusted add-in to catch and ignore these serious exceptions.

The NUnit hang is due to NUnit Framework not being able to catch such exceptions. See https://github.com/nunit/nunit/issues/3295#issuecomment-507503500.

Issue #2021, opened 2017-01-31
Alternate Timeout API for dynamic timeout intervals


jnm2:
There is only one way to specify a timeout as far as I can tell: [`TimeoutAttribute`](https://github.com/nunit/docs/wiki/Timeout-Attribute). The fact that it's an attribute means that the timeout time must be a compile-time constant.

However, I am searching for a way to override that timeout if a given environment variable is present. If `TF_BUILD` is present, I want to use `7000`. Otherwise, I want to use `3000`.

If you want to stick with attributes, a conditional timeout attribute could be added such as `TimeoutIfEnv("TF_BUILD", 7000)`. However a more flexible API might be an assertion, taking a page from `Throws`:

```c#
var timeoutMs = Environment.GetEnvironmentVariable("TF_BUILD") != null ? 7000 : 3000;
Assert.That(() =>
{
    // ...
}, Does.Not.Timeout(timeoutMs));
```

The non-aborting version could be `Has.MaxTime`.
There could be a `TimeSpan` overload as well.

What are your opinions?

JustinRChou:
I am trying to figure out what other cases would you need a dynamic timeout. 
Though it would be interesting for performance testing.

jnm2:
> I am trying to figure out what other cases would you need a dynamic timeout.
Though it would be interesting for performance testing.

Parameterized tests might calculate the timeout based on parameter values.

jnm2:
In fact, that allows the timeout to be one of the parameters in a list of `TestCase` attributes.

JustinRChou:
So would it equivalent to doing something like
```
        [TestCase("a", "b", 1000)]
        public void TimeElasped(object a, object b, int timeout)
        {
            System.Diagnostics.Stopwatch watch = new System.Diagnostics.Stopwatch();
            watch.Start();
            RunMethod(a, b);
            watch.Stop();
            Assert.That(watch.ElapsedMilliseconds, Is.InRange(timeout - 50, timeout + 50));
        }
```

jnm2:
@JustinRChou Yes. That's the `Has.MaxTime` variant. The `Does.Not.Timeout` variant would have to also abort the thread to keep par with TimeoutAttribute.

bmolnar2:
I'd be happy to check out this one. Ready to dive into open-source contribution. :)

jnm2:
@bmolnar2 Awesome! It's yours. I'm assigning myself to the issue to keep track of it. Let us know if you need anything.

[...]

PR #2513, opened 2017-10-23
Add MaxTime constrain property to has.

bmolnar2:
This change adds the feature of Has.MaxTime constraint for assertions: #2021

The implementation is based on https://gist.github.com/jnm2/f5c4e2224114d23e566467dbf1467128, however I rather organized out InvocationDescriptor from ThrowConstraint an resued that.

Please comment on the timeout values use in that test, whether they should be less or just fine with this timeout.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/MaxTimeConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/MaxTimeConstraint.cs
@@ -1,0 +1,102 @@
+// *****************************************************************************
+//Copyright(c) 2017 Charlie Poole, Rob Prouse
+
+//Permission is hereby granted, free of charge, to any person obtaining a copy
+//of this software and associated documentation files (the "Software"), to deal
+//in the Software without restriction, including without limitation the rights
+//to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//copies of the Software, and to permit persons to whom the Software is
+//furnished to do so, subject to the following conditions:
+
+//The above copyright notice and this permission notice shall be included in
+//all copies or substantial portions of the Software.
+
+//THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//THE SOFTWARE.
+// ****************************************************************************
+
+using System;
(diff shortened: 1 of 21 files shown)


src/NUnitFramework/framework/Constraints/MaxTimeConstraint.cs
```diff
@@ -0,0 +1,102 @@
...
+                    stopwatch.Start();
+                    object result = invocationDescriptor.Invoke();
+                    region.WaitForPendingOperationsToComplete(result);
+                    stopwatch.Stop();
+                }
+            }
+            else
+#endif
+            {
+                using (new TestExecutionContext.IsolatedContext())
```
CharliePoole: What's the purpose of using an IsolatedContext here? If the delegate should Assert, don't you want it to be reported?

src/NUnitFramework/framework/Constraints/Operators/MaxTimeOperator.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System;
+
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// Operator that tests if executed code terminated within
+    /// specified maximum time.
+    /// </summary>
+    public class MaxTimeOperator : SelfResolvingOperator
```
CharliePoole: There appears to be know reason for this to be a SelfResolvingOperator. If there is a reason, please explain. In fact, I'm not clear why any operator at all is needed. This could be because I'm missing something in how this is supposed to be used. You don't actually have any tests that show it in use, only formal tests of the constraint.

src/NUnitFramework/framework/Constraints/MaxTimeConstraint.cs
bmolnar2: Copy-paste error. It's not needed here, I'll remove that using.

CharliePoole:
@rprouse I'm not convinced we want this feature at all and I think it's a shame that @bmolnar2 has done so much work on it. We should somehow continue and wrap up the discussion on the issue, which even now contains no solid examples of how this is supposed to be used.

@bmolnar2 I've only done the barest code review and asked some questions inline. IMO, it's a waste of time for you to actually make any changes until we resolve what we want to see. More on that on the issue.

bmolnar2:
Ok, I'll stand down. Let me know if you need to revoke this one.

jnm2:
Chronologically:

> I agree that we need something like that. If we all agree - as I suspect we do - then that moves us from the idea to the design stage

> I'm not convinced we want this feature at all

> On the substance of the feature: I think we all have agreed that there's a need for some more granular way to test the maximum time that things take within a test.

The middle statement seems harsh to me.

CharliePoole:
@jnm2 Ref our separate discussion about chatter, you're generating notices again, in this case a rather cryptic one. Now I'm generating one to say "what's your point other than to criticise my language?"

IMO "I'm not convinced we want this feature" isn't harsh, it's a fairly dispassionate way of saying I didn't like "this feature" - whatever that was in the context.

Let's keep our discussions here substantive rather than ad hominem.


To that point, I'll re-summarize that I think we have agreed we need __some__ more granular feature but not on the syntax and precise semantics of it. We are collectively at the API design stage, not the coding stage, even though you may feel the design is settled.

jnm2:
I apologize that my communication was cryptic and ad hominem. Not my intent at all, but I recognize that's the effect of it. My desire for visible communication and my desire to not go into a long meta came into conflict in that comment. I'll follow up with the concern I have via email.

rprouse:
Closing because nobody has touched this PR in so long.

PR #2517, opened 2017-10-29
Replaced RangeOfInt with Enumerable.Range

jnm2:
Discovery at https://github.com/nunit/nunit/pull/2501#issuecomment-340278428 necessitated a fix to RangeOfInt. Finding Enumerable.Range already in NUnit.System.Linq, I opted to remove `RangeOfInt`.

/cc @CharliePoole @oznetmaster


CharliePoole:
This looks good to me. Apparently, I tried `Enumerable.Range` before I started using `NUnit.System.Linq` for some other purposes. I'll hold on merging to give @oznetmaster another shot at reviewing.

oznetmaster:
I am still confused by my new results :(

777
935
909
3843
2124
206

String is the fastest??  Short is the slowest by a factor of 3?  Still very puzzled.

(Realize this is a very slow processor, comparatively)

jnm2:
@oznetmaster I'll try some stuff. How about with NonParallelizable?

Wild guess but `short` is going to be unaligned and everything? I know `int` is measurabley slower than `long` in 64-bit systems (which is why the proposed `nint` will be particularly interesting to me).

oznetmaster:
This is a single core / single thread processor.

oznetmaster:
Xamarin has nint (and nfloat) but I was not aware it had been proposed for .NET Core?

jnm2:
Multiplying by 100, I'm getting:

image

jnm2:
@oznetmaster You can probably find a lot of the discussions and meeting notes from https://github.com/dotnet/csharplang/issues/435.

jnm2:
Modified the range so I could do `short`:

image

The double speed is what confuses me. Then again, we are talking about doing hash lookups and expanding hash tables. Pretty sensitive to things like how hashes are calculated.

oznetmaster:
That really makes no sense.  The second fastest is the string with ignore case?  That is the one which does the ToLower in the constraint.  The string which does not ignore case is is slower?  And the double is the next fastest?  I wonder what a long would do?  And a float?

jnm2:
Yeah, good questions. Have to leave it for a while again, but my next step would be to see if I can reproduce without NUnit.

jnm2:
I changed the order of the test cases and got dramatically inconsistent results:

image

I checked using a stopwatch; `MaxTimeAttribute` is working properly.

jnm2:
Perhaps doubles of consecutive integers just hash with a better distribution than ints until you get up to a certain amount.

Anyway, okay to merge?

CharliePoole:
If users reported irregularity in using NUnit to measure performance, I'd tell them that NUnit isn't designed to do that. The original issue was merely to correct an extraordinarily bad level of performance and I think we should stick with that goal rather than trying to make a good performance tool... unless you want to actually build a performance tool. 😈

jnm2:
Yes. I think BenchmarkDotNet is the gold standard for performance testing, and it already has NUnit bindings. :D

PR #2528, opened 2017-11-02
Fixes CollectionAssert not using IEquatable of items in collections

### LINKED ISSUE #2271: When CollectionAssert.AreEqual do compare each element, it will ignore the IEquatable of the element too
With the code changes #2118 for [Issue2097](https://github.com/nunit/nunit/issues/2097), the test below still will fail, since when the `NUnitEqualityComparer` does each item comparison, the property `compareAsCollection` is still `true`,  so it will not get in the  [condition](https://github.com/nunit/nunit/blob/release/3.7/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs#L229), which ignoring the `IEquatable` of the element and will fail the assertion. 
```
namespace Nunit3TestsSamples {
	[TestFixture]
	class TestSample1 {

		[Test]
		public void test() {
			var set1 = new List<TestObject>() {
				new TestObject { Value = "test1" },
				new TestObject { Value = "test2" },
				new TestObject { Value = "test3" }
			};
			var set2 = new List<TestObject>() {
				new TestObject { Value = "test1" },
				new TestObject { Value = "test2" },
				new TestObject { Value = "test3" }
			};

			CollectionAssert.AreEqual( set1, set2 );
		}

		class TestObject : IEquatable<TestObject> {
			public string Value { get; set; }

			public bool Equals( TestObject other ) {
				return Value.Equals( other.Value );
			}
		}
	}
}
```

OmicronPersei:
Closes #2271 .

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/DictionariesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/DictionariesComparer.cs
@@ -54,7 +54,7 @@ namespace NUnit.Framework.Constraints.Comparers
                 return false;
 
             foreach (object key in xDictionary.Keys)
-                if (!_equalityComparer.AreEqual(xDictionary[key], yDictionary[key], ref tolerance, false))
+                if (!_equalityComparer.AreEqual(xDictionary[key], yDictionary[key], ref tolerance, topLevelComparison: false))
                     return false;
 
             return true;
--- a/src/NUnitFramework/framework/Constraints/Comparers/DictionariesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/DictionariesComparer.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2009 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -37,7 +37,7 @@ namespace NUnit.Framework.Constraints.Comparers
             _equalityComparer = equalityComparer;
         }
 
-        public bool? Equal(object x, object y, ref Tolerance tolerance)
+        public bool? Equal(object x, object y, ref Tolerance tolerance, bool topLevelComparison = true)
         {
             if (!(x is IDictionary) || !(y is IDictionary))
                 return null;
@@ -54,7 +54,7 @@ namespace NUnit.Framework.Constraints.Comparers
                 return false;
 
             foreach (object key in xDictionary.Keys)
-                if (!_equalityComparer.AreEqual(xDictionary[key], yDictionary[key], ref tolerance))
+                if (!_equalityComparer.AreEqual(xDictionary[key], yDictionary[key], ref tolerance, false))
                     return false;
 
             return true;
--- a/src/NUnitFramework/framework/Constraints/Comparers/IChainComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/IChainComparer.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2009 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -36,10 +36,11 @@ namespace NUnit.Framework.Constraints.Comparers
         /// <param name="x">The first object to compare.</param>
         /// <param name="y">The second object to compare.</param>
         /// <param name="tolerance">The tolerance to use when comparing the objects.</param>
+        /// <param name="topLevelComparison">Flag indicating whether or not this is the top level comparison.</param>
         /// <returns>
         ///     <c>null</c> if the objects cannot be compared using the method.
         ///     Otherwise the result of the comparison is returned.
         /// </returns>
-        bool? Equal(object x, object y, ref Tolerance tolerance);
+        bool? Equal(object x, object y, ref Tolerance tolerance, bool topLevelComparison = true);
     }
 }
--- a/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2009 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -60,8 +60,6 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         private readonly List<IChainComparer> _comparers;
 
-        private bool _topLevelComparison = true;
-
         #endregion
 
         internal NUnitEqualityComparer()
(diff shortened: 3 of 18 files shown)


CharliePoole:
Saving and restoring the `TopLevelComparison` flag seems fragile to me. What if you simply made it a parameter to `ObjectsEqual` with a default value of true. All the nested calls could then set it to false. Cleanup wouldn't be needed since it would just be on the call stack.

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -85,6 +85,11 @@ internal NUnitEqualityComparer()
             };
         }
 
+        internal NUnitEqualityComparer(NUnitEqualityComparer copy)
```
CharliePoole: I can't figure out where this is used.

OmicronPersei:
Tests still pass even if I don't change all of the calls to the equals methods but I thought it would be better to be consistent, especially if a method is refactored to make use of the new topLevelComparison flag

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
OmicronPersei: It's not, looks like I forgot to remove it.

src/NUnitFramework/framework/Constraints/Comparers/DictionariesComparer.cs
```diff
@@ -54,10 +54,10 @@ internal DictionariesComparer(NUnitEqualityComparer equalityComparer)
                 return false;
 
             foreach (object key in xDictionary.Keys)
-                if (!_equalityComparer.AreEqual(xDictionary[key], yDictionary[key], ref tolerance))
+                if (!_equalityComparer.AreEqual(xDictionary[key], yDictionary[key], ref tolerance, false))
```
jnm2: Very minor, makes the code more readable IMO to use `topLevelComparison: false`.

src/NUnitFramework/framework/Constraints/Comparers/IChainComparer.cs
```diff
@@ -36,10 +36,11 @@ internal interface IChainComparer
         /// <param name="x">The first object to compare.</param>
         /// <param name="y">The second object to compare.</param>
         /// <param name="tolerance">The tolerance to use when comparing the objects.</param>
+        /// <param name="topLevelComparison">Flag indicating whether or not this is the top level comparison.</param>
         /// <returns>
         ///     <c>null</c> if the objects cannot be compared using the method.
         ///     Otherwise the result of the comparison is returned.
         /// </returns>
-        bool? Equal(object x, object y, ref Tolerance tolerance);
+        bool? Equal(object x, object y, ref Tolerance tolerance, bool topLevelComparison = true);
```
jnm2: The only reason to use default values is for the convenience of the caller. Is `IChainCaller` directly called a lot? I have a small preference for making things explicit at the call site. We don't want to make it easy for future developers to miss the significance of the decision going into the parameter.

src/NUnitFramework/tests/Assertions/CollectionAssertTest.cs
```diff
@@ -220,6 +220,62 @@ public void AreEqual_UsingIterator()
...
+            IEnumerable set1 = new SimpleEnumerableWithIEquatable("x", "y", "z");
+            IEnumerable set2 = new SimpleEnumerableWithIEquatable("x", "z", "z");
+
+            Assert.AreNotEqual(set1, set2);
+
+            Assert.Throws<AssertionException>(() => CollectionAssert.AreEqual(set1, set2));
+        }
+
+        [Test]
+        public void IEnumerablesAreEqualWithCollectionsObjectsImplemetingIEquatable()
```
jnm2: `Implemeting` is typo

src/NUnitFramework/tests/Assertions/CollectionAssertTest.cs
```diff
@@ -220,6 +220,62 @@ public void AreEqual_UsingIterator()
...
+        //}
+
+        //[Test]
+        //public void IEnumerablesAreNotEqualWithCollectionObjectsImplementingOverridenEquals()
+        //{
+        //    IEnumerable set1 = new SimpleEnumerable(new SimpleOverridenEqualsObj());
+        //    IEnumerable set2 = new SimpleEnumerable(new SimpleOverridenEqualsObj(), new SimpleOverridenEqualsObj());
+
+        //    CollectionAssert.AreNotEqual(set1, set2);
+        //}
```
jnm2: Why leave the commented-out code?

src/NUnitFramework/tests/TestUtilities/Collections/SimpleEnumerable.cs
```diff
@@ -54,4 +54,69 @@ public IEnumerator GetEnumerator()
 
         #endregion
     }
+
+    class SimpleEnumerableWithIEquatable : IEnumerable, IEquatable<IEnumerable<object>>
```
jnm2: I'm confused why this isn't either `IEquatable<SimpleEnumerableWithIEquatable>` or `IEquatable<IEnumerable>`?
The class doesn't implement `IEnumerable<object>` so right now it is not equatable to itself.

src/NUnitFramework/framework/Constraints/Comparers/DictionariesComparer.cs
OmicronPersei: That works!  I didn't even know that was valid C# syntax... how long has that been a language feature?

jnm2: 1.0 I think :D

CharliePoole: 4.0 actually.

src/NUnitFramework/framework/Constraints/Comparers/IChainComparer.cs
CharliePoole: I geneally feel the same way but I have to disagree in this case. The external caller, which actually makes the top level call, should not have to say "This is a top level call" and should not be able to say "This is not a top level call."

 Admittedly, exposing the argument this way is a bit of a compromise. I suggested it because it allows keeping the information about whether it's a top level call on the stack.

OmicronPersei:
All good, then?

src/NUnitFramework/tests/TestUtilities/Collections/SimpleEnumerable.cs
```diff
@@ -54,4 +54,59 @@ public IEnumerator GetEnumerator()
 
         #endregion
     }
+
+    class SimpleEnumerableWithIEquatable : IEnumerable<object>, IEquatable<SimpleEnumerableWithIEquatable>
+    {
+        public List<object> Contents { get; private set; } = new List<object>();
```
oznetmaster: Why the initializer?

Both constructors initialize the value of Contents, so this default initialization is redundant.

jnm2: Good catch; also the private set is unnecessary.

OmicronPersei: Just saw this, should I write a PR to do this small refactor?  Don't mind, just not sure if it's "PR worthy" or something like that since it is pretty minor.

ChrisMaddock: If you're willing, a small PR would be great. 😄

jnm2: I agree. If something just makes sense to do, I don't care how small it is. =)

PR #2530, opened 2017-11-03
Running tests on main thread. Revisiting #2483

endrih:
Adding setting option RunOnMainThread = true for when the test is required to run on the main thread. This is only needed when you are embedding the engine in your application and the tests need to run on the same thread as the application


ChrisMaddock:
I agree with Charlie. Tieing in with that, I think it may be a cleaner implementation if direct dispatch is broken out into a separate IWorkItemDispatcher. Besides just being a cleaner implementation, this would also mean:

a) The dispatcher can just be set based on the FrameworkPackageSettings in NUnitTestAssemblyRunner - no need to set anything on any WorkItem.
b) We can get rid of all the pre-compilation symbols in SimpleWorkItemDispatcher - it looks like the two NetStandard builds should just use the new IWorkItemDispatcher by default instead.
c) Methods like IWorkItemDispatcher.Cancel(true) which aren't implemented in a way that works for running on the main thread I believe, could correctly throw a NotImplementedException.

jnm2:
I think the standard is to use `NotSupportedException` if it's by design and `NotImplementedException` if you're planning to implement something later.

endrih:
@ChrisMaddock @CharliePoole I see your points and I agree. I will try to implement @ChrisMaddock 's suggestions soon. Thanks

endrih:
@ChrisMaddock @jnm2 Is there a way to exclude SimpleWorkItemDispatcher from NETSTANDARD1_6 build altogether? I just put it between `#if !NETSTANDARD1_6 #endif`, I'm not sure if this is the right way.

src/NUnitFramework/framework/Api/FrameworkController.cs
```diff
@@ -138,6 +138,12 @@ private void Initialize(string assemblyPath, IDictionary settings)
             var newSettings = settings as IDictionary<string, object>;
             Settings = newSettings ?? settings.Cast<DictionaryEntry>().ToDictionary(de => (string)de.Key, de => de.Value);
 
+            if (Settings.ContainsKey(FrameworkPackageSettings.RunOnMainThread) &&
+                (bool)Settings[FrameworkPackageSettings.RunOnMainThread])
+            {
+                Settings[FrameworkPackageSettings.NumberOfTestWorkers] = 0;
+            }
+
```
ChrisMaddock: Is this still needed now?

src/NUnitFramework/framework/Api/FrameworkPackageSettings.cs
```diff
@@ -115,6 +115,11 @@ public static class FrameworkPackageSettings
         /// </summary>
         public const string TestParametersDictionary = "TestParametersDictionary";
 
+        /// <summary>
+        /// Parameters to be passed on to the tests, already parsed into an IDictionary&lt;string, string>. Replaces <see cref="TestParameters"/>.
+        /// </summary>
```
ChrisMaddock: Comment needs fixing. 🙂

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -343,16 +343,22 @@ private void CreateTestExecutionContext(ITestListener listener)
...
+            Context.Dispatcher = new MainThreadWorkItemDispatcher();
+#else
 #if PARALLEL
             int levelOfParallelism = GetLevelOfParallelism();
 
             if (levelOfParallelism > 0)
                 Context.Dispatcher = new ParallelWorkItemDispatcher(levelOfParallelism);
+            else if (Settings.ContainsKey(FrameworkPackageSettings.RunOnMainThread) &&
+                     (bool)Settings[FrameworkPackageSettings.RunOnMainThread])
+                Context.Dispatcher = new MainThreadWorkItemDispatcher();
```
ChrisMaddock: Ah - now I see why setting TestWorkers = 0 is required!

Given you could never run in parallel on the main thread, can I suggest that the `FrameworkPackageSettings.RunOnMainThread` check is done first, and that overrides any parallelism settings? That would hopefully simplify the logic - unless there's another reason to need `NumberOfTestWorkers` to be set to zero?

src/NUnitFramework/framework/Internal/Execution/MainThreadWorkItemDispatcher.cs
```diff
@@ -0,0 +1,80 @@
...
+
+namespace NUnit.Framework.Internal.Execution
+{
+    /// <summary>
+    /// SimpleWorkItemDispatcher handles execution of WorkItems by
+    /// directly executing them. It is provided so that a dispatcher
+    /// is always available in the context, thereby simplifying the
+    /// code needed to run child tests.
+    /// </summary>
+    public class MainThreadWorkItemDispatcher : IWorkItemDispatcher
```
ChrisMaddock: I much prefer this class - thanks for writing it up! A couple of nitpicks:
- The class comment, and `Start()` comment need updating. `CancelRun()` would ideally say it is not supported in the doccomment too?
- There's a stray blank line on 54.
- Can you update the year in the copyright notice to be 2017?

Thanks!

ChrisMaddock:
> @ChrisMaddock @jnm2 Is there a way to exclude SimpleWorkItemDispatcher from NETSTANDARD1_6 build altogether? I just put it between #if !NETSTANDARD1_6 #endif, I'm not sure if this is the right way.

That's what I would have done, but I'm so familiar with the new csproj system. @jnm2 - are we missing anything? 😄

jnm2:
@endrih @ChrisMaddock Sorry I missed the question! That is indeed how you exclude something from the .NET Standard 1.6 build. If you're asking whether we *should* exclude it from the .NET Standard build, I don't know what's going on yet. 😁

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
endrih: I guess you are right. If the condition for `RunOnMainThread` comes first, I didn't see any reason for the `NumberOfTestWorkers` to be forced to 0, even though conceptually it implies that `NumberOfTestWorkers` is 0. So I am removing that part from FrameworkController.cs for now.

src/NUnitFramework/framework/Api/FrameworkPackageSettings.cs
endrih: :sweat_smile:

endrih:
@ChrisMaddock Thanks for the review. I believe I addressed all of your comments, but let me know if you want anything else changed. I am keeping it as WIP because I am running our local test suite to make sure that everything works as it previously did. I'll remove it when my testing is done.

ChrisMaddock:
Fixes look good, thanks @endrih! You've just missed the main class doccomment on `MainThreadWorkItemDispatcher` - that's still the one from `SimpleItemWorkItemDispatcher`. 😄 

After that, I'm happy for this to be merged (and looks like @jnm2 is too!). Just give us a ping once you're done testing with your system. 👍

src/NUnitFramework/framework/Internal/Execution/MainThreadWorkItemDispatcher.cs
```diff
@@ -0,0 +1,80 @@
...
+        public void Dispatch(WorkItem work)
+        {
+            if (work != null)
+                work.Execute();
+        }
+
+        /// <summary>
+        /// This method is not supported for 
+        /// this dispatcher. Using it will throw a
+        /// NotSupportedException
```
jnm2: Can you add an end mark to the second sentence?

ChrisMaddock:
@endrih - You've just missed one of my comments above - otherwise, good to merge!

https://github.com/nunit/nunit/pull/2530#issuecomment-350263501
> You've just missed the main class doccomment on MainThreadWorkItemDispatcher - that's still the one from SimpleItemWorkItemDispatcher. 😄

rprouse:
This looks okay. We will need a second review.

jnm2:
@endrih Thanks for the contribution! 🎉

PR #2531, opened 2017-11-06
Fix #2516: Add testId to StrErr output

### LINKED ISSUE #2516: When test writes something to the stdErr there is no guaranteed way to link a test-output event to a target test using ITestEventListener
@NikolayPianikov commented on [Thu Oct 26 2017](https://github.com/nunit/nunit-console/issues/305)

For example:

- Unit test `SuccessfulStdErrorTest` has code: `System.Console.Error.WriteLine("errorout");`

- ITestEventListener receives the following events:
```
<start-test id="0-1001" parentId="0-1000" name="SuccessfulStdErrorTest" fullname="Foo.Tests.UnitTests1.SuccessfulStdErrorTest" type="TestMethod" />

<test-output stream="Error" testname="Foo.Tests.UnitTests1.SuccessfulStdErrorTest"><![CDATA[errorout
]]></test-output>
```

`test-output` has no information id/parentId about a target test and there is no way to attach this error output. It is important to note that these events could be received from different tests simultaneously.

See this [sample](https://github.com/nunit/nunit-console/files/1418559/sample.zip)

MhdTlb:


src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
```diff
@@ -68,10 +68,14 @@ private bool TrySendToListener(string text)
...
-                : null;
-            context.Listener.TestOutput(new TestOutput(text, _streamName, testName));
+            string testName = null;
+            string testId = null;
+            if (context.CurrentTest != null)
+            {
+                testName = context.CurrentTest.FullName;
+                testId = context.CurrentTest.Id;
+            }
+            context.Listener.TestOutput(new TestOutput(text, _streamName, testId, testName));
```
CharliePoole: Could simplify this using ?. and ?? but it's not essential for merging as it's a stylistic decision.

```C#
var test = context.CurrentTest ?? null;
context.Listener.TestOutput(new TestOutput(text, _streamName, test?.Id, test?.FullName);
```

src/NUnitFramework/nunitlite.tests/TextUITests.cs
```diff
@@ -539,6 +539,28 @@ public void TwoTests_NestedExecution(string labels, string expected)
...
+            }
+
+            if (testName != null)
+            {
+                expected.AppendFormat(" testname=\"{0}\"", testName);
+            }
+
+            expected.AppendFormat("><![CDATA[{0}]]></test-output>", text);
+            Assert.That(testOutput.ToXml(), Is.EqualTo(expected.ToString()));
+        }
```
CharliePoole: It's a good test, but it's a low-level test of `TestOutput` not of `TextUI`. In addition, `TextUI` is a bad place because it's part of the `nunitlite` assembly, not of `nunit.framework`.

Best choice is to create a new class for `TestOutputTests` with this single test in it. Second, acceptable to me, choice is to piggyback it on `TextOutputTests` because that's at least in the correct assembly. 😄

src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
MhdTlb: Good suggestion

src/NUnitFramework/tests/nunit.framework.tests-4.5.csproj
```diff
@@ -207,6 +207,7 @@
     <Compile Include="Constraints\ToStringTests.cs" />
     <Compile Include="Constraints\UniqueItemsConstraintTests.cs" />
     <Compile Include="Constraints\XmlSerializableTest.cs" />
+    <Compile Include="Interfaces\TestOutputTests.cs" />
```
CharliePoole: I like it! But now you have to add it to the other five projects. :-) 

That's 2.0, 3.5, 4.0 and netstandard 1.3 and 1.6.

MhdTlb: Is there a specific way to add them? Or simply copy/paste? I noticed that files in different projects are linked somehow: changes in file propagate to others

CharliePoole: You just add them. The separate csproj files are all located in the same directory so no link is needed. It's tedious but easy. 😄

MhdTlb: I see now. Thanks @CharliePoole

CharliePoole:
@MhdTlb We require two committer reviews before merging. @jnm2 @ChrisMaddock Can one of you review this?

src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
```diff
@@ -68,10 +68,8 @@ private bool TrySendToListener(string text)
             if (context == null || context.Listener == null)
                 return false;
 
-            string testName = context.CurrentTest != null
-                ? context.CurrentTest.FullName
-                : null;
-            context.Listener.TestOutput(new TestOutput(text, _streamName, testName));
+            var test = context.CurrentTest ?? null;
```
jnm2: Can you remove the `?? null` since it has no effect? It might even make sense to inline the `test` variable.

src/NUnitFramework/nunitlite.tests/TextUITests.cs
```diff
@@ -527,18 +527,17 @@ public void TwoTests_NestedExecution(string labels, string expected)
...
+            textUI.TestOutput(new TestOutput("Another immediate output from first test\n", "", null, test1.FullName));
+            textUI.TestOutput(new TestOutput("Immediate output from second test\n", "", null, test2.FullName));
             result2.OutWriter.WriteLine("OUTPUT2");
             textUI.TestFinished(result2);
             result1.OutWriter.WriteLine("OUTPUT1");
             textUI.TestFinished(result1);
 
             Assert.That(Report, Is.EqualTo(expected));
         }
-
```
jnm2: This isn't a big deal, but if you see accidental changes like this can you undo them them?

src/NUnitFramework/tests/nunit.framework.tests-4.5.csproj
jnm2: This is about to get much easier really soon when we move to the new csproj SDK which doesn't require listing `<Compile>` items in the csproj at all.

ChrisMaddock:
Looks good to me - with the one point @jnm2 picked up on. Thanks @MhdTlb! 😄

src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
MhdTlb: Nice catch ;)

src/NUnitFramework/nunitlite.tests/TextUITests.cs
MhdTlb: @jnm2  I'm having hard time with line feed. Giving that the change was made in previous commit, How can I undo it in later commit?

jnm2: I see you added a new line to reverse what you did which is great. What you're seeing now is that instead of adding an empty line, you added a line with eight spaces in it. If you want to get rid of this change, just delete the eight spaces and there will no longer be a net change in this PR. But feel free to leave this if you've already put enough time into it; it's not a big deal.

MhdTlb: Let me give it another try

jnm2:
Tests hung on AppVeyor again. I restarted the build. Travis' OSX queue is just backed up as per normal.

jnm2:
Merged. Thank you @MhdTlb for another great contribution!

MhdTlb:
Glad I could contribute to such interesting project  😃

PR #2532, opened 2017-11-06
Filter visibility

CharliePoole:
This is a partial contribution toward issue #1426.

All classes in the `NUnit.Framework.Internal.Filters` are changed to have `internal` visibility.

In addition, I removed `[Serializable]` from all of these classes because I can't see any reason they would need to be serializable, since NUnit no longer passes filters over remoting as it did in V2. If it turns out they need to be serialized because of some other, external usage we can squash out the second commit. However, ideally, external classes should be passing XML strings to represent filters.

The `TestFilter` class itself has some external usage and its visibility remains unchanged for now.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Filters/AndFilter.cs
+++ b/src/NUnitFramework/framework/Internal/Filters/AndFilter.cs
@@ -31,9 +31,6 @@ namespace NUnit.Framework.Internal.Filters
     /// Combines multiple filters so that a test must pass all 
     /// of them in order to pass this filter.
     /// </summary>
-#if !NETSTANDARD1_3 && !NETSTANDARD1_6
-    [Serializable]
-#endif
     internal class AndFilter : CompositeFilter
     {
         /// <summary>
--- a/src/NUnitFramework/framework/Internal/Filters/AndFilter.cs
+++ b/src/NUnitFramework/framework/Internal/Filters/AndFilter.cs
@@ -34,7 +34,7 @@ namespace NUnit.Framework.Internal.Filters
 #if !NETSTANDARD1_3 && !NETSTANDARD1_6
     [Serializable]
 #endif
-    public class AndFilter : CompositeFilter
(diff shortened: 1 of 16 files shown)


jairbubbles:
FYI in my application I'm running tests like that:

```csharp
NUnitTestAssemblyRunner runner = new NUnitTestAssemblyRunner(new DefaultTestAssemblyBuilder());
runner.Load(assembly, new Dictionary<string, object>());
var results = runner.Run(null, new NotFilter(new CategoryFilter("Functional")));
```
Updating to a recent version (we were using 3.5) will break compilation.

We'll use `TestFilter.FromXml` now but I prefer the old syntax.

jnm2:
Thanks for letting us know, @jairbubbles!

@nunit/framework-team If we productized a C# filter construction API, we'd want to move it out of the `Internal` namespace. I have a preference for the style `TestFilter.Not(TestFilter.Category("Functional"))` which would fit well alongside `TestFilter.Empty`. Is this something we'd be interested in starting an issue for?

ChrisMaddock:
I'd be keen for us to provide an API for creating filters.

jnm2:
Great, me too! I started a simple proposal: https://github.com/nunit/nunit/issues/2931

CharliePoole:
@ChrisMaddock We do provide such an API in the engine's `ITestFilterBuilder` service. I'd be opposed to duplicating the same logic in the framework. The OP here is writing code that directly calls methods and types we never intended to be called that way. The fact that we made them public has been discussed a lot and we owe him an apology for misleading him. However, I think we should keep working to restrict that sort of usage, not extend it.

Note that the same logic is already duplcated in NUnitLite, so this would make three copies.

jnm2:
> Note that the same logic is already duplicated in NUnitLite, so this would make three copies.

You're referring to `--where`/test-list parsing, right? As far as I know, no other API exists to _programmatically_ build filter XML.

PR #2535, opened 2017-11-07
Reorganize nunit.framework.tests namespaces

### LINKED ISSUE #2311: Resolve test projects' namespace situation
We are fighting uphill against convention and against most contributors' tooling by not namespacing according to folder structure. We're inconsistent which leads to confusion and a bunch of `using` directives with the name segments in a different order when adding new files.

I propose that we (I'm willing) do a sweep and make everything match the folder structure. Most people are used to working this way. The root namespace in every file will be `NUnit.Framework.Tests`.

`NUnit.Framework.Internal.Tests` would become `NUnit.Framework.Tests.Internal`.
`NUnit.Framework.Helpers.Tests` would become `NUnit.Framework.Tests.Helpers` (as an example; we'll actually see this when we get to NUnit.Console).

It's bothered me continually since my first PR here and it bothers new contributors as well. I've had it in my mind for months to open this issue but I put it off since it will obviously create merge conflicts in all open PRs. We need to just pick a time and get 'er done.

CharliePoole:
Fixes #2311 

We've been talking about doing this for a while! All tests are now in the same namespaces as the thing they test. Default namespace is changed and resources are fixed (that part was a real pain)!


ChrisMaddock:
Brilliant - thanks for sorting this one Charlie!

PR #2537, opened 2017-11-07
Respect ApartmentState when running tests in Parallel

### LINKED ISSUE #2288: Killing thread cancels test run

### LINKED ISSUE #2426: Regression in 3.8.1: ApartmentAttribute no longer works when applied to an assembly

rprouse:
Fixes #2426 Confirmed by adding the tests from the issue to an integration test project.
Fixes #2288 Confirmed by adding the tests from this issue to the framework tests.

Walking through the code, I think this used to work before because test methods were not run in parallel. Because of this, if their parent was running STA, the tests would just run in that thread. Now the tests are possibly run in parallel too, they cannot rely on their parent. It still worked for simple tests where the test was the direct child of whatever was marked STA, but not more than one level of difference.

This fix has the tests walk up their parent hierarchy until they find the first level that specifies an `ApartmentState` other than `Unknown`. The tests then assume they should be running in the `ApartmentState` that their parent specifies.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
@@ -67,9 +67,7 @@ namespace NUnit.Framework.Internal.Execution
                 : ParallelScope.Default;
 
 #if !NETSTANDARD1_3 && !NETSTANDARD1_6
-            TargetApartment = Test.Properties.ContainsKey(PropertyNames.ApartmentState)
-                ? (ApartmentState)Test.Properties.Get(PropertyNames.ApartmentState)
-                : ApartmentState.Unknown;
+            TargetApartment = GetTargetApartment(Test);
 #endif
 
             State = WorkItemState.Ready;
@@ -496,8 +494,8 @@ namespace NUnit.Framework.Internal.Execution
 #if PARALLEL
         private ParallelExecutionStrategy GetExecutionStrategy()
         {
-            // If there is no fixture and so nothing to do but dispatch 
-            // grandchildren we run directly. This saves time that would 
+            // If there is no fixture and so nothing to do but dispatch
+            // grandchildren we run directly. This saves time that would
             // otherwise be spent enqueuing and dequeing items.
             if (Test.TypeInfo == null)
                 return ParallelExecutionStrategy.Direct;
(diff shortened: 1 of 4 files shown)


src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -527,7 +525,24 @@ private ParallelExecutionStrategy GetExecutionStrategy()
...
+        /// <see cref="ApartmentState"/> has been set on any of the parent tests.
+        /// </summary>
+        static ApartmentState GetTargetApartment(ITest test)
+        {
+            var apartment = test.Properties.ContainsKey(PropertyNames.ApartmentState)
+                ? (ApartmentState)test.Properties.Get(PropertyNames.ApartmentState)
+                : ApartmentState.Unknown;
 
+            if (apartment == ApartmentState.Unknown && test.Parent != null)
+                return GetTargetApartment(test.Parent);
```
rprouse: This is the meat of the fix, the rest is pretty much just tests.

rprouse:
@CharliePoole just to be sure, I added tests that request a different apartment than their parent. I also created versions of the tests that are non-parallel. I then stepped through the `ParallelWorkItemDispatcher`. Each of these was executed with a `ParallelExecutionStrategy` of `Parallel` or `NonParallel` as appropriate, never `Direct`. Based on this, I think we are good. Just waiting for the CI to pass on the newly added tests.

jnm2:
Sorry @rprouse, was just about to review since I pushed the PR I was head-down on. Looks excellent.

Issue #2529, opened 2017-11-02
TestFixtureSource attribute inhibits discovery when exploring, but not when running tests


oznetmaster:
It appears, at least in my CF build, that the use of the TestFixtureSource attribute (which is currently only used by the new ParallelExecutionTests) inhibits the test fixture from being discovered during an explore, but still allow the fixture to be correctly discovered and executed by a run.

Not sure how to go about pinning this one down.

oznetmaster:
Perhaps it is because TestFixtureAttribute implements ITestFixtureData, and TestFixtureSourceAttribute does not?

oznetmaster:
Maybe not, since TestAttribute and TestCaseSourceAttribute follow the have the same pattern in interfaces they implement.

CharliePoole:
What do you mean by not being able to discover them in an explore? Is there a command that will show me the problem?

oznetmaster:
Just do an explore of the framework tests assembly, and the resulting list will not include the ParallelExecutionTests

oznetmaster:
It will be the only missing test fixture in the list.

CharliePoole:
All of the fixtures use `.SetName` so the class name `ParallelExecutionTests` never appears. But the list does actually include all of them...

```
NUnit.Framework.Internal.Execution.Issue-2464.AllTestsPassed
NUnit.Framework.Internal.Execution.Issue-2464.CorrectInitialShift
NUnit.Framework.Internal.Execution.Issue-2464.OnlyOneShiftIsActiveAtSameTime
NUnit.Framework.Internal.Execution.Issue-2464.TestsRunOnExpectedWorkers
NUnit.Framework.Internal.Execution.Issue-2467.AllTestsPassed
NUnit.Framework.Internal.Execution.Issue-2467.CorrectInitialShift
NUnit.Framework.Internal.Execution.Issue-2467.OnlyOneShiftIsActiveAtSameTime
NUnit.Framework.Internal.Execution.Issue-2467.TestsRunOnExpectedWorkers
NUnit.Framework.Internal.Execution.SingleFixture_AssemblyAndFixtureParallelizable.AllTestsPassed
NUnit.Framework.Internal.Execution.SingleFixture_AssemblyAndFixtureParallelizable.CorrectInitialShift
NUnit.Framework.Internal.Execution.SingleFixture_AssemblyAndFixtureParallelizable.OnlyOneShiftIsActiveAtSameTime
[...]

PR #2538, opened 2017-11-07
[WIP] SetArgDisplayNames for TestCase/FixtureData and templating for TestFixtureData.SetName

### LINKED ISSUE #2536: SetArgDisplayNames for TestCaseData and TestFixtureData
##### Edit: `TestFixtureData.SetName` is being made public again after we pulled it from 3.9 now that the behavior is consistent with `TestCaseData`

`TestCaseData.SetName` leaves you in an impossible situation explored by https://github.com/nunit/nunit/issues/1943 if you are forced to customize parameterized test names and the parameter names contain curlies. To summarize:

> The thing that makes me uncomfortable here is that NUnit forces us to use SetName because otherwise the autogenerated test names collide, which in turn forces us to replace curlies with some other character. Feature holes are fine unless the feature is forced on you.

/cc @stewart-r as promised, who said:

> Maybe I am guilty of seeing the world through the prism of my own recent experience but I have to say am really surprised that it doesn't seem to affect more people. People test edge cases, any strings with curly braces are edge cases for string manipulation code.... where is the tsunami!?? :-p

Also to say nothing of the escaping problems we ran into with curlies, `SetName` is just plain verbose. The style just about everyone wants to use, and I agree with them, is `"{m}(" + string.Join(", ", customParameterNames) + ")"`.

We explored different alternatives and this is the only one that addresses all my concerns, @stewart-r's concerns, and @oznetmaster's concerns with the `SetName` being added to `TestFixtureSource` (https://github.com/nunit/nunit/issues/2529):

```diff
namespace NUnit.Framework
{
    public class TestCaseData
    {
+       public TestCaseData SetArgDisplay(params string[] names);
    }
    public class TestFixtureData
    {
+       public TestFixtureData SetName(string name);
+       public TestFixtureData SetName(params string[] names);
    }
}
```

Usage example for TestFixtureData: `.SetArgDisplay("Issue-2464")`
```
NUnit.Framework.Internal.Execution.ParallelExecutionTests(Issue-2464).AllTestsPassed
```

TestCaseData usage example:

```cs
// x = new Person { Name = "Stewart {C} Robertson" }, y = 42
new TestCaseData(x, y, z).SetArgDisplay(x.Name, y)
```

Resulting test name:
```
Namespace.FixtureName.MethodName(Stewart {C} Robertson, 42)
```

`SetArgDisplay` fills a need that no other solution fills. It's urgent for my real world use cases. If I implement it in time for 3.9, it also allows us to remove `TestFixtureData.SetName` until we have a proper discussion about expansions and escaping (and whether we'd even want it). Thus I'm marking this design discussion pri:high.

jnm2:
Closes #2536 and also brings `TestFixtureData.SetName` into consistency with `TestCaseSource.SetName`.

See the TestNameGeneratorTests for a quick intro to the approach I took. Walking through the commits one by one could help because they are separable.

I also added it to TestFixtureData which made it super easy to make `TestFixtureData.SetName` consistent with `TestCaseData` with name generation. Expansions will work. This is something we should discuss and agree on before `TestFixtureData.SetName` makes its debut in 3.9. /cc @rprouse @CharliePoole 

API changes:

```diff
namespace NUnit.Framework
{
    public class TestCaseData : NUnit.Framework.Internal.TestCaseParameters
    {
+       public TestCaseData SetArgDisplay(params string[] displayNames);
    }

    public class TestFixtureData : NUnit.Framework.Internal.TestFixtureParameters
    {
+       public TestFixtureData SetArgDisplay(params string[] displayNames);
    }
}
namespace NUnit.Framework.Internal
{
-   public abstract class TestParameters : NUnit.Framework.Interfaces.ITestData, NUnit.Framework.Interfaces.IApplyToTest
+   public abstract class TestParameters : NUnit.Framework.Interfaces.ITestData
    {
-       public void ApplyToTest(Test test)
+       public void ApplyToTest(Test test, TestNameGenerator defaultTestNameGenerator)
    }
}
```

/fyi @stewart-r

### DIFF
--- a/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
@@ -75,19 +75,7 @@ namespace NUnit.Framework.Internal.Builders
 
             if (parms != null)
             {
-                parms.ApplyToTest(testMethod);
-
-                if (parms.TestName != null)
-                {
-                    // The test is simply for efficiency
-                    testMethod.Name = parms.TestName.Contains("{")
-                        ? new TestNameGenerator(parms.TestName).GetDisplayName(testMethod, parms.OriginalArguments)
-                        : parms.TestName;
-                }
-                else
-                {
-                    testMethod.Name = _nameGenerator.GetDisplayName(testMethod, parms.OriginalArguments);
-                }
+                parms.ApplyToTest(testMethod, _nameGenerator);
             }
             else
             {
@@ -134,7 +122,7 @@ namespace NUnit.Framework.Internal.Builders
             int minArgsNeeded = 0;
             foreach (var parameter in parameters)
             {
-                // IsOptional is supported since .NET 1.1 
+                // IsOptional is supported since .NET 1.1
                 if (!parameter.IsOptional)
                     minArgsNeeded++;
             }
--- a/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
@@ -79,7 +79,7 @@ namespace NUnit.Framework.Internal.Builders
             }
             else
             {
-                testMethod.Name = _nameGenerator.GetDisplayName(testMethod, null);
+                testMethod.Name = _nameGenerator.GetDisplayName(testMethod, null, null);
             }
 
             testMethod.FullName = prefix + "." + testMethod.Name;
--- a/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
@@ -44,13 +44,12 @@ namespace NUnit.Framework.Internal.Builders
         }
 
         /// <summary>
-        /// Builds a single NUnitTestMethod, either as a child of the fixture
-        /// or as one of a set of test cases under a ParameterizedTestMethodSuite.
+        /// Builds a single <see cref="TestMethod"/>, either as a child of the fixture
+        /// or as one of a set of test cases under a <see cref="ParameterizedMethodSuite"/>.
         /// </summary>
-        /// <param name="method">The MethodInfo from which to construct the TestMethod</param>
+        /// <param name="method">The <see cref="IMethodInfo"/> from which to construct the <see cref="TestMethod"/></param>
         /// <param name="parentSuite">The suite or fixture to which the new test will be added</param>
-        /// <param name="parms">The ParameterSet to be used, or null</param>
-        /// <returns></returns>
+        /// <param name="parms">The <see cref="TestCaseParameters"/> to be used, or null</param>
         public TestMethod BuildTestMethod(IMethodInfo method, Test parentSuite, TestCaseParameters parms)
         {
             var testMethod = new TestMethod(method, parentSuite)
--- a/src/NUnitFramework/framework/Internal/Builders/NUnitTestFixtureBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/NUnitTestFixtureBuilder.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
@@ -22,7 +22,6 @@
 // ***********************************************************************
 
 using System;
-using System.Collections;
 using System.Reflection;
 using NUnit.Compatibility;
 using NUnit.Framework.Interfaces;
@@ -51,6 +50,7 @@ namespace NUnit.Framework.Internal.Builders
         #region Instance Fields
 
         private ITestCaseBuilder _testBuilder = new DefaultTestCaseBuilder();
+        private readonly TestNameGenerator _nameGenerator = new TestNameGenerator(TestNameGenerator.DefaultTestFixturePattern);
 
         #endregion
 
@@ -82,7 +82,7 @@ namespace NUnit.Framework.Internal.Builders
 
         /// <summary>
         /// Overload of BuildFrom called by tests that have arguments.
-        /// Builds a fixture using the provided type and information 
+        /// Builds a fixture using the provided type and information
         /// in the ITestFixtureData object.
         /// </summary>
         /// <param name="typeInfo">The TypeInfo for which to construct a fixture.</param>
@@ -128,10 +128,26 @@ namespace NUnit.Framework.Internal.Builders
             var fixture = new TestFixture(typeInfo, arguments);
 
             string name = fixture.Name;
-            if (testFixtureData.TestName != null)
-                fixture.Name = testFixtureData.TestName;
-            else if (arguments != null && arguments.Length > 0)
-                fixture.Name = typeInfo.GetDisplayName(arguments);
+
+            var parms = testFixtureData as TestParameters;
+            if (parms != null)
+            {
+                parms.ApplyToTest(fixture, _nameGenerator);
+            }
+            else
+            {
+                if (testFixtureData.TestName != null)
+                    fixture.Name = testFixtureData.TestName;
+                else if (arguments != null && arguments.Length > 0)
+                    fixture.Name = typeInfo.GetDisplayName(arguments);
+
+                if (fixture.RunState != RunState.NotRunnable)
+                    fixture.RunState = testFixtureData.RunState;
+
+                foreach (string key in testFixtureData.Properties.Keys)
+                    foreach (object val in testFixtureData.Properties[key])
+                        fixture.Properties.Add(key, val);
+            }
 
             if (fixture.Name != name) // name was changed
             {
@@ -140,13 +156,6 @@ namespace NUnit.Framework.Internal.Builders
                     ? nspace + "." + fixture.Name
                     : fixture.Name;
             }
-
-            if (fixture.RunState != RunState.NotRunnable)
-                fixture.RunState = testFixtureData.RunState;
-
-            foreach (string key in testFixtureData.Properties.Keys)
-                foreach (object val in testFixtureData.Properties[key])
-                    fixture.Properties.Add(key, val);
 
             if (fixture.RunState != RunState.NotRunnable)
                 CheckTestFixtureIsValid(fixture);
@@ -196,7 +205,7 @@ namespace NUnit.Framework.Internal.Builders
         /// any global TestCaseBuilder addin wants to build the
         /// test case. If not, it uses the internal builder
         /// collection maintained by this fixture builder.
-        /// 
+        ///
         /// The default implementation has no test case builders.
         /// Derived classes should add builders to the collection
         /// in their constructor.
--- a/src/NUnitFramework/framework/Internal/TestCaseParameters.cs
+++ b/src/NUnitFramework/framework/Internal/TestCaseParameters.cs
@@ -31,7 +31,7 @@ namespace NUnit.Framework.Internal
     /// other selected parameters needed for constructing
     /// a parameterized test case.
     /// </summary>
-    public class TestCaseParameters : TestParameters, ITestCaseData, IApplyToTest
+    public class TestCaseParameters : TestParameters, ITestCaseData
     {
         #region Instance Fields
 
--- a/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
+++ b/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
@@ -39,7 +39,11 @@ namespace NUnit.Framework.Internal
         // pattern all the way down the test builder hierarchy
 
         /// <summary>
-        /// Default pattern used to generate names
+        /// Default pattern used to generate test fixture names
+        /// </summary>
+        public static string DefaultTestFixturePattern = "{c}{a}";
+        /// <summary>
+        /// Default pattern used to generate test method names
         /// </summary>
         public static string DefaultTestNamePattern = "{m}{a}";
 
@@ -77,9 +81,9 @@ namespace NUnit.Framework.Internal
         }
 
         /// <summary>
-        /// Get the display name for a TestMethod and its arguments
+        /// Get the display name for a test and its arguments
         /// </summary>
-        /// <param name="testMethod">A TestMethod</param>
+        /// <param name="test">A test</param>
         /// <param name="argDisplayNames">
         /// Overrides argument list display.
         /// The strings in this list are already display names and will be rendered as is, without quotes.
@@ -88,7 +92,7 @@ namespace NUnit.Framework.Internal
         /// Arguments to be used if <paramref name="argDisplayNames"/> is <see langword="null"/>.
         /// A string in this list will be rendered with quotes.
         /// </param>
-        public string GetDisplayName(TestMethod testMethod, string[] argDisplayNames, object[] args)
+        public string GetDisplayName(Test test, string[] argDisplayNames, object[] args)
         {
             if (_fragments == null)
                 _fragments = BuildFragmentList(_pattern);
@@ -96,7 +100,7 @@ namespace NUnit.Framework.Internal
             var result = new StringBuilder();
 
             foreach (var fragment in _fragments)
-                result.Append(fragment.GetText(testMethod, argDisplayNames ?? args, argDisplayNames != null));
+                result.Append(fragment.GetText(test, argDisplayNames ?? args, argDisplayNames != null));
 
             return result.ToString();
         }
@@ -210,14 +214,9 @@ namespace NUnit.Framework.Internal
 
         private abstract class NameFragment
         {
-            private const string THREE_DOTS = "...";
+            private const string ELLIPSIS = "...";
 
-            public virtual string GetText(TestMethod testMethod, object[] args, bool argsAreDisplayNames)
-            {
-                return GetText(testMethod.Method.MethodInfo, args, argsAreDisplayNames);
-            }
-
-            public abstract string GetText(MethodInfo method, object[] args, bool argsAreDisplayNames);
+            public abstract string GetText(Test test, object[] args, bool argsAreDisplayNames);
 
             protected static void AppendGenericTypeNames(StringBuilder sb, MethodInfo method)
             {
@@ -229,6 +228,35 @@ namespace NUnit.Framework.Internal
                     sb.Append(t.Name);
                 }
                 sb.Append(">");
+            }
+
+            protected static string LimitString(string value, int maxLength)
+            {
+                if (value == null) return null;
+                if (value.Length <= maxLength) return value;
+
+                var valueLengthToAppend = maxLength - ELLIPSIS.Length;
+                if (valueLengthToAppend <= 0) return ELLIPSIS;
+
+                var buffer = new char[maxLength];
+                value.CopyTo(0, buffer, 0, valueLengthToAppend);
+                ELLIPSIS.CopyTo(0, buffer, valueLengthToAppend, ELLIPSIS.Length);
+                return new string(buffer);
+            }
+
+            protected static void AppendLimited(StringBuilder sb, string value, int maxLength)
+            {
+                if (value == null) return;
+                if (value.Length <= maxLength)
+                {
+                    sb.Append(value);
+                }
+                else
+                {
+                    var valueLengthToAppend = maxLength - ELLIPSIS.Length;
+                    if (0 < valueLengthToAppend) sb.Append(value, 0, valueLengthToAppend);
+                    sb.Append(ELLIPSIS);
+                }
             }
 
             protected static string GetDisplayString(object arg, int stringMax)
@@ -312,7 +340,7 @@ namespace NUnit.Framework.Internal
                 {
                     var str = (string)arg;
                     bool tooLong = stringMax > 0 && str.Length > stringMax;
-                    int limit = tooLong ? stringMax - THREE_DOTS.Length : 0;
+                    int limit = tooLong ? stringMax - ELLIPSIS.Length : 0;
 
                     StringBuilder sb = new StringBuilder();
                     sb.Append("\"");
@@ -321,7 +349,7 @@ namespace NUnit.Framework.Internal
                         sb.Append(EscapeCharInString(c));
                         if (tooLong && sb.Length > limit)
                         {
-                            sb.Append(THREE_DOTS);
+                            sb.Append(ELLIPSIS);
                             break;
                         }
                     }
@@ -430,14 +458,9 @@ namespace NUnit.Framework.Internal
 
         private class TestIDFragment : NameFragment
         {
-            public override string GetText(MethodInfo method, object[] args, bool argsAreDisplayNames)
+            public override string GetText(Test test, object[] args, bool argsAreDisplayNames)
             {
-                return "{i}"; // No id available using MethodInfo
-            }
-
-            public override string GetText(TestMethod testMethod, object[] args, bool argsAreDisplayNames)
-            {
-                return testMethod.Id;
+                return test.Id;
             }
         }
 
@@ -450,7 +473,7 @@ namespace NUnit.Framework.Internal
                 _text = text;
             }
 
-            public override string GetText(MethodInfo method, object[] args, bool argsAreDisplayNames)
+            public override string GetText(Test test, object[] args, bool argsAreDisplayNames)
             {
                 return _text;
             }
@@ -458,14 +481,17 @@ namespace NUnit.Framework.Internal
 
         private class MethodNameFragment : NameFragment
         {
-            public override string GetText(MethodInfo method, object[] args, bool argsAreDisplayNames)
+            public override string GetText(Test test, object[] args, bool argsAreDisplayNames)
             {
+                var method = test.Method;
+                if (method == null) return null;
+
                 var sb = new StringBuilder();
 
                 sb.Append(method.Name);
 
                 if (method.IsGenericMethod)
-                    AppendGenericTypeNames(sb, method);
+                    AppendGenericTypeNames(sb, method.MethodInfo);
 
                 return sb.ToString();
             }
@@ -473,24 +499,27 @@ namespace NUnit.Framework.Internal
 
         private class NamespaceFragment : NameFragment
         {
-            public override string GetText(MethodInfo method, object[] args, bool argsAreDisplayNames)
+            public override string GetText(Test test, object[] args, bool argsAreDisplayNames)
             {
-                return method.DeclaringType.Namespace;
+                return (test.Method?.TypeInfo ?? test.TypeInfo)?.Namespace;
             }
         }
 
         private class MethodFullNameFragment : NameFragment
         {
-            public override string GetText(MethodInfo method, object[] args, bool argsAreDisplayNames)
+            public override string GetText(Test test, object[] args, bool argsAreDisplayNames)
             {
+                var method = test.Method;
+                if (method == null) return null;
+
                 var sb = new StringBuilder();
 
-                sb.Append(method.DeclaringType.FullName);
+                sb.Append(method.TypeInfo.FullName);
                 sb.Append('.');
                 sb.Append(method.Name);
 
                 if (method.IsGenericMethod)
-                    AppendGenericTypeNames(sb, method);
+                    AppendGenericTypeNames(sb, method.MethodInfo);
 
                 return sb.ToString();
             }
@@ -498,17 +527,17 @@ namespace NUnit.Framework.Internal
 
         private class ClassNameFragment : NameFragment
         {
-            public override string GetText(MethodInfo method, object[] args, bool argsAreDisplayNames)
+            public override string GetText(Test test, object[] args, bool argsAreDisplayNames)
             {
-                return method.DeclaringType.Name;
+                return (test.Method?.TypeInfo ?? test.TypeInfo)?.Name;
             }
         }
 
         private class ClassFullNameFragment : NameFragment
         {
-            public override string GetText(MethodInfo method, object[] args, bool argsAreDisplayNames)
+            public override string GetText(Test test, object[] args, bool argsAreDisplayNames)
             {
-                return method.DeclaringType.FullName;
+                return (test.Method?.TypeInfo ?? test.TypeInfo)?.FullName;
             }
         }
 
@@ -521,7 +550,7 @@ namespace NUnit.Framework.Internal
                 _maxStringLength = maxStringLength;
             }
 
-            public override string GetText(MethodInfo method, object[] arglist, bool argsAreDisplayNames)
+            public override string GetText(Test test, object[] arglist, bool argsAreDisplayNames)
             {
                 var sb = new StringBuilder();
 
@@ -532,7 +561,19 @@ namespace NUnit.Framework.Internal
                     for (int i = 0; i < arglist.Length; i++)
                     {
                         if (i > 0) sb.Append(",");
-                        sb.Append(argsAreDisplayNames ? (string)arglist[i] : GetDisplayString(arglist[i], _maxStringLength));
+
+                        if (argsAreDisplayNames)
+                        {
+                            var displayName = (string)arglist[i];
+                            if (_maxStringLength != 0)
+                                AppendLimited(sb, displayName, _maxStringLength);
+                            else
+                                sb.Append(displayName);
+                        }
+                        else
+                        {
+                            sb.Append(GetDisplayString(arglist[i], _maxStringLength));
+                        }
                     }
 
                     sb.Append(')');
@@ -553,11 +594,17 @@ namespace NUnit.Framework.Internal
                 _maxStringLength = maxStringLength;
             }
 
-            public override string GetText(MethodInfo method, object[] args, bool argsAreDisplayNames)
+            public override string GetText(Test test, object[] args, bool argsAreDisplayNames)
             {
-                return args.Length <= _index ? string.Empty :
-                    argsAreDisplayNames ? (string)args[_index] :
-                    GetDisplayString(args[_index], _maxStringLength);
+                if (args.Length <= _index) return string.Empty;
+
+                if (argsAreDisplayNames)
+                {
+                    var displayName = (string)args[_index];
+                    return _maxStringLength != 0 ? LimitString(displayName, _maxStringLength) : displayName;
+                }
+
+                return GetDisplayString(args[_index], _maxStringLength);
             }
         }
 
--- a/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
+++ b/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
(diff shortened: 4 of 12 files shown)


jnm2:
Hmm, perhaps I could avoid the `IApplyToTest` changes by making the `TestCaseParameters` and `TestFixtureParameters` be responsible for the default name generator instance rather than the fixture builder and test case builder?

CharliePoole:
What does "tell-dont-ask" to apply test names mean? 

Also, why drop the IApplyToTest interface, which is used in other places to modify a test once it has been constructed?

jnm2:
It means to let the test parameters decide how to apply the Name rather than querying TestName and now ArgNames and possibly future properties, then making decisions based on those properties before applying the result to test.Name.

It allowed me to make the ArgNames properry protected instead of public, and to not duplicate the whole chunk of logic when applying the test fixture name.

I dropped the interface because it was in the way and I can't see any reason for TestParameters to use it. There is no code which needs an IApplyToTest to which we pass a TestParameters.

It might make sense to let TestParameters be in control of the default test name generators though, in which case I could put the interface back.

jnm2:
I just checked to make sure I wasn't neglecting a cast to `object` and then to `IApplyToTest`. This is the only client code of `IApplyToTest` in the entire framework, and it's all about attributes: https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/Tests/Test.cs#L336-L378

CharliePoole:
I found a few places to comment on details, but I'm finding a more general problem with this PR, so I'll just make it as a comment.

In addition to what the title of the PR says, you are making several other changes, which appear to me to be orthogonal to what the issue was about. IF you want to entirely change how we go about naming tests, that's valid. But I think you need to create an issue that explains the changes, have it discussed and approved and then implement it. Just putting the code out there forces me to read the code merely to deduce your intention.

Here's what I see in this issue

* Implements `SetArgNames` for both fixtures and cases (as the title says)

* Implements template-based naming of fixtures, a much bigger change and one we haven't decided we even want. Personally, I thought we might want it one day, which is why `{c}` etc. exist, but over time I came to feel it was overkill. The entire template thing initially got pushed because of problems with name duplication in VS. Fixtures don't enter into that at all.

* Restructures the underlying implementation rather more than is needed for the first point, but probably needed if we wanted to do templates for fixtures.

As a smaller issue, it looks like you are trying to make both `SetName` and `SetArgNames` work when they are both used. My suggestion would be to just tell users they can't do that and either use the last one we find or give an error.

As a matter of simple change control, I wouldn't want to do all this a few days before a release anyway, even if I agreed with the changes. If SetArgNames is something you really need for your own work, can't you implement it simply and in a way that will allow later doing more, if we decide we want more?

CharliePoole:
BTW, this is the sort of thing that makes me wish for a `develop` branch!

CharliePoole:
Currently, `NUnitTestCaseBuilder` creates the single default `TestNameGenerator` but doesn't maintain the default value. It might make sense to move that default value there and eliminate the default constructor of `TestNameGenerator`.

I'm not sure I understand what you mean by having `TestParameters` maintain the default instances. That wouldn't make them available to non-parameterized tests, would it?

jnm2:
None of the changes are orthogonal from the perspective of illustrating how painless and unifying it is to have `TestFixtureData.SetName` template just like `TestCaseData`, which I needed to reference in your conversation on https://github.com/nunit/nunit/issues/2529. But like I said, depending on that design discussion, it's no problem to peel off that part and leave the templating for SetArgNames.

> If SetArgNames is something you really need for your own work, can't you implement it simply and in a way that will allow later doing more, if we decide we want more?

So yes, I very much can. Further, I'm even asking for `TestFixtureData.SetName` to be pulled if a longer discussion makes you more comfortable because of the issue I have with the status quo. Of course Rob has the final say; if he or any third @nunit/framework-team member think I should let it go, I will.

> The entire template thing initially got pushed because of problems with name duplication in VS. Fixtures don't enter into that at all.

I don't understand this. They control a segment in the test name. If the full test name is duplicated, both VS and NCrunch have problems.

jnm2:
> I'm not sure I understand what you mean by having TestParameters maintain the default instances. That wouldn't make them available to non-parameterized tests, would it?

True. I believe there is a TestParameters that gets created for non-parameterized tests but I'd have to look. Creating a default TestParameters if none exists would tie up a couple loose ends in Test*Parameters. Alternately, I think making static DefaultTestNameGenerator instances is a perfectly suitable way to handle this too.

CharliePoole:
@jnm2 Answering your last comment first, It merely seemed to me that since
`NUnitTestCaseBuilder` is in charge of how tests are built, it should (and
does) create a default instance of the `TestNameGenerator`. If we generate
names for fixtures, then I'd say that the Fixture builder should be in
charge of that one.

jnm2:
And I'm fine leaving it that way, as it is in the PR> :+1:

CharliePoole:
Thanks for clarifying your 👍 I was just about to ask what you were plus-one-ing. 😄

CharliePoole:
Issue for this?

src/NUnitFramework/framework/Internal/TestCaseParameters.cs
```diff
@@ -31,7 +31,7 @@ namespace NUnit.Framework.Internal
     /// other selected parameters needed for constructing
     /// a parameterized test case.
     /// </summary>
-    public class TestCaseParameters : TestParameters, ITestCaseData, IApplyToTest
+    public class TestCaseParameters : TestParameters, ITestCaseData
```
CharliePoole: If the `ApplyToTest` method does the same thing that Attributes implementing the `IApplyToTest` interface do, then I like the idea that we signal this to the developers, even though an interface is not strictly needed.

src/NUnitFramework/framework/Internal/TestParameters.cs
```diff
@@ -123,23 +123,29 @@ private void InitializeArguments(object[] args)
 
         #endregion
 
-        #region IApplyToTest Members
-
         /// <summary>
-        /// Applies ParameterSet values to the test itself.
+        /// Applies the encapsulated parameters to the test method.
```
CharliePoole: `TestMethod`

src/NUnitFramework/framework/Internal/TestParameters.cs
```diff
@@ -123,23 +123,29 @@ private void InitializeArguments(object[] args)
...
 
             foreach (string key in Properties.Keys)
                 foreach (object value in Properties[key])
                     test.Properties.Add(key, value);
-        }
 
-        #endregion
+            if (TestName == null)
+            {
+                test.Name = defaultTestNameGenerator.GetDisplayName(test, OriginalArguments);
```
CharliePoole: I'll have to look elsewhere to understand this but in the original implementation, the test already has a name when it is created. The name is changed by `ApplyToTest` after the fact. In that way, the method doesn't need to have the default name generator passed to it.

If we do end up deciding to keep this, however, then let's change the name of the method so it doesn't cause confusion with the IApplyToTest method. Objects that implement IApplyToTest are supposed to only work with a small, local context and not have extra information available like whether the user specified a different default on the command line.

src/NUnitFramework/framework/Internal/TestParameters.cs
```diff
@@ -141,7 +141,7 @@ public void ApplyToTest(TestMethod test, TestNameGenerator defaultTestNameGenera
             }
             else
             {
-                test.Name = TestName.Contains("{")
+                test.Name = TestName.IndexOf('{') != -1
```
CharliePoole: I'm not in love with using a less clear code structure for hypothetical efficiency. If there were a `String.Contains(char)` method, I'd use that. It's quite possible that the internal implementation of `Contains` deals with the special case of a one-character string, but even if it doesn't, we don't really know that we need efficiency here.

CharliePoole:
Do you have a particular use case in mind for wanting SetArgNames to interoperate with SetName, rather than making them mutually exclusive?

src/NUnitFramework/framework/Internal/TestParameters.cs
jnm2: Fair. I actually agree with you rather strongly so there's that. 😆

src/NUnitFramework/framework/Internal/TestCaseParameters.cs
jnm2: This is not possible unless we stop passing the TestNameGenerator to the method which requires either the Test*Parameters classes to know the default test name generator, or for us to go back to the builder asking for all name-related arguments rather than telling the parameters to apply the name (plus duplicating the name application logic between the two builders).

src/NUnitFramework/framework/Internal/TestParameters.cs
jnm2: :+1: This change gets reversed in a later commit.

src/NUnitFramework/framework/Internal/TestParameters.cs
jnm2: We still need the default name generator because we pass the argument display names to it (seen in a later commit). It's either this, or make `ArgNames` public and keep the logic external- hopefully in a single place and not in each builder.

I'm happy with changing the method name. I almost did so.

jnm2:
> Do you have a particular use case in mind for wanting SetArgNames to interoperate with SetName, rather than making them mutually exclusive?

I deliberated over this and first tried the mutually exclusive approach, but found it more arbitrary and more complex not to allow the symbiosis. I already knew I wanted to use TestNameGenerator for this to allow for simple implementation (like character limits; the default pattern used to have a limit) and to keep similar things in close proximity to each other for consistency. Once there, the interoperation with `SetName` was already present. I'm of course open to other ideas. Each decision I made seemed the simplest at the time.

jnm2:
I removed the Indexof commit. Did my replies give you all the information you needed? Is this direction acceptable to you or would you like me to try a different approach first?

src/NUnitFramework/framework/Internal/TestCaseParameters.cs
CharliePoole: I agree that the way the code is written currently is wrong. It's not how any other attribute or object works. The `TestCaseParameters` object should decide what to do in it's ApplyToTest method __if anything__ and the method should always be called. But ApplyToTest should not be responsible for the __default__ name. That should just be there before it is ever called.

It's true that doing it that way causes the name to be set twice, but that doesn't seem to have caused any problems up to now.

jnm2: Let me clarify: I'm not passing the default name generator in order to generate the default name. I'm passing the default name generator in order to use the default name pattern with *non-default* overridden argument display names.

jnm2:
I'm actually interested in making `TestParameters.ArgNames` internal so that it doesn't affect our public API at all. I'd make it `private protected` if this was C# 7.2 because that's exactly what it should be.

src/NUnitFramework/framework/Internal/TestCaseParameters.cs
CharliePoole: Ah! That makes a bit more sense. My assumption was that `SetArgNames` would be just a wrapper around `SetName("{m}(whatever,you,tell,it)")`.

jnm2: Can't do that in case an argument contained curlies... there's no escape method =D

jnm2: To be honest I do think it's cleaner this way, plus `{m}{a:40}` works.

jnm2: I do actually believe I did the simplest thing possible. =D Could be wrong, but so far I think so.

ChrisMaddock:
Sorry - I'm struggling for time this week and have had to skim this and the associated threads - but I agree Joseph's concerns about the consistency between setting fixture names and test case names, and agree we should temporarily pull the `Fixture.SetName` API for 3.9, if we can't get this resolved in time.

jnm2:
I am going to add the remaining tests and make a couple small changes and ask for reviews again soon.

CharliePoole:
@ChrisMaddock After all this (confusing) discussion, I'm in agreement on removing the feature for 3.9. It's only used in my `ParallelExecutionTests` where I needed it when working on them so I could debug one fixture instance at a time.

jnm2:
@CharliePoole Would you like `SetArgNames` to wait as well as `SetName` and use some other naming tactic, or were you meaning we should use `SetArgNames` instead of `SetName` as I suggested earlier?

jnm2:
Tests are almost done btw.

CharliePoole:
I think whether to wait on this is a release management decision but it might help if @ChrisMaddock chimed in. My own view is that we have a history of putting in fairly big changes at the last minute and regretting it. Not primarily for technical reasons, but because we didn't think about some particular use case.

If we do put it in, I'm still a bit leary of the `SetArgNames` name, which I'm afraid will lead some users to imagine they are changing the arguments to the function.

BTW this is where I'm a big fan of the RM function and the use of various levels of "freeze" as a release approaches. As a rough starting point, for our team and environment, I'd do a "feature freeze" two weeks before and a full freeze one week before the release - exception made for critical errors that pop up, of course.

jnm2:
Yes, Chris (and Rob) please do comment. I'm happy so long as we either pull SetName or merge this PR before release.

jnm2:
@CharliePoole `SetArgDisplay(params string[] displayNames)`?

CharliePoole:
`SetArgDisplay` works for me!

jnm2:
Renamed to SetArgDisplay, added final tests and made final corrections! Updated API diff at the top.

jnm2:
@rprouse: you said to ping for 3.9 stuff. I feel strongly that we should either make `TestFixtureData.SetName` internal for 3.9 or else merge this PR for 3.9. [Chris agrees](https://github.com/nunit/nunit/pull/2538#issuecomment-342818787) and Charlie [agrees](https://github.com/nunit/nunit/pull/2538#issuecomment-342869594) but [warns](https://github.com/nunit/nunit/pull/2538#issuecomment-342877649) that we have a history of big last-minute changes.

I still don't think of this as a big change, and @ChrisMaddock hasn't chimed in on that, but if anyone thinks this deserves more time I'm all for safe rather than sorry. Please let me know which way you want to go; if it helps, I could PR to remove `TestFixtureData.SetName` or (easier) make it internal.

oznetmaster:
Since this still seems to be under discussion, it would seem that making TestFixtureData.SetName internal for the 3.9 release would be the best thing to do.

I do not feel strongly about this.  Just read all the discussion.

oznetmaster:
There is a set of tests written by @CharliePoole that are in the current build that use TestFixtureData.SetName, so we cannot just get rid of it.

jnm2:
@oznetmaster Since we can make it internal I don't see any *reason* to remove it, but I believe it could be removed if we wanted. Charlie says above:

> I'm in agreement on removing the feature for 3.9. It's only used in my ParallelExecutionTests where I needed it when working on them so I could debug one fixture instance at a time.

That said, I'll start up a PR to make it internal. Thanks for your feedback.

jnm2:
@rprouse: Here's the PR to make it internal: https://github.com/nunit/nunit/pull/2554

jnm2:
@nunit/framework-team What's remaining to discuss on this?

jnm2:
I summarized the entire PR as it stands at https://github.com/nunit/nunit/issues/2536#issuecomment-350503635.

@CharliePoole, you're still requesting changes. Would you please review and make me aware of any outstanding issues with the PR?

src/NUnitFramework/framework/Internal/TestParameters.cs
```diff
@@ -116,30 +116,42 @@ private void InitializeArguments(object[] args)
...
+        public void ApplyToTest(Test test, TestNameGenerator defaultTestNameGenerator)
         {
-            if (this.RunState != RunState.Runnable)
-                test.RunState = this.RunState;
+            if (RunState != RunState.Runnable)
+                test.RunState = RunState;
 
             foreach (string key in Properties.Keys)
                 foreach (object value in Properties[key])
                     test.Properties.Add(key, value);
```
ChrisMaddock: I find these two things being done in both TestParameters and TestFixtureBuilder-but-only-under-certain-circumstances kinda confusing. Is there a reason they need to be done at this point? Setting the RunState doesn't feel like anything to do with TestParameters? (I know this isn't your code, but while we're making big changes in this area! 😄)

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
```diff
@@ -430,14 +458,9 @@ private static string EscapeControlChar(char c)
 
         private class TestIDFragment : NameFragment
         {
-            public override string GetText(MethodInfo method, object[] args, bool argsAreDisplayNames)
-            {
-                return "{i}"; // No id available using MethodInfo
-            }
-
```
ChrisMaddock: What's this change doing?

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
```diff
@@ -450,65 +473,71 @@ public FixedTextFragment(string text)
...
         }
 
         private class MethodNameFragment : NameFragment
         {
-            public override string GetText(MethodInfo method, object[] args, bool argsAreDisplayNames)
+            public override string GetText(Test test, object[] args, bool argsAreDisplayNames)
             {
+                var method = test.Method;
+                if (method == null) return null;
+
```
ChrisMaddock: Is there a case where test.Method should be null? Can we comment it, and if not, maybe `Guard.NotNull()` would be more explicit?

src/NUnitFramework/tests/Internal/TestNameGeneratorTests.cs
```diff
@@ -84,6 +84,19 @@ public string ParameterizedTests(string pattern, object[] args)
...
 
+        [TestCase("{a}", new[] { "1", "2" }, ExpectedResult = "(1,2)")]
+        [TestCase("{a}", new[] { ",", " " }, ExpectedResult = "(,, )")]
+        [TestCase("{a}", new string[0], ExpectedResult = "()")]
+        [TestCase("{0}", new[] { "1", "2" }, ExpectedResult = "1")]
+        [TestCase("{0}{1}", new[] { "1", "2" }, ExpectedResult = "12")]
+        [TestCase("{0},{1},{2}", new[] { "1", "2" }, ExpectedResult = "1,2,")]
+        [TestCase("{0:20}", new[] { "Now is the time for all good men to come to the aid of their country." }, ExpectedResult = "Now is the time f...")]
+        [TestCase("{a:20}", new[] { "42", "Now is the time for all good men to come to the aid of their country." }, ExpectedResult = "(42,Now is the time f...)")]
+        public string ArgumentDisplayNames(string pattern, string[] argDisplayNames)
```
ChrisMaddock: Glad to see these tests - there's a lot going on now!

src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
```diff
@@ -75,23 +74,11 @@ public TestMethod BuildTestMethod(IMethodInfo method, Test parentSuite, TestCase
...
-                else
-                {
-                    testMethod.Name = _nameGenerator.GetDisplayName(testMethod, parms.OriginalArguments);
-                }
+                parms.ApplyToTest(testMethod, _nameGenerator);
             }
             else
             {
-                testMethod.Name = _nameGenerator.GetDisplayName(testMethod, null);
+                testMethod.Name = _nameGenerator.GetDisplayName(testMethod, null, null);
```
ChrisMaddock: I'm not a massive fan of how this code has ended up reading - namely how it looks like the TestName is generated entirely differently depending on whether there are params or not. 

I'd prefer this to be `_nameGenerator.GetDisplayName()` in all cases, and for TestNameGenerator to look after all the resolution of testnames, and deal with all the if/else cases which are currently scattered through TestCaseBuilder/TestFixtureBuilder/TestNameGenerator/TestParameters.

Would that be technically possible, or am I missing why it's currently done this way? Is it just that doing that would be a bigger refactor than this PR has set out to do? It feels like we're changing the structure fundamentally enough at this point for that to be worthwhile.

src/NUnitFramework/framework/Internal/Builders/NUnitTestFixtureBuilder.cs
```diff
@@ -128,10 +128,26 @@ public TestSuite BuildFrom(ITypeInfo typeInfo, ITestFixtureData testFixtureData)
...
+                else if (arguments != null && arguments.Length > 0)
+                    fixture.Name = typeInfo.GetDisplayName(arguments);
+
+                if (fixture.RunState != RunState.NotRunnable)
+                    fixture.RunState = testFixtureData.RunState;
+
+                foreach (string key in testFixtureData.Properties.Keys)
+                    foreach (object val in testFixtureData.Properties[key])
+                        fixture.Properties.Add(key, val);
+            }
```
ChrisMaddock: Is there a reason setting properties/runstate was moved up to where testnames are set? Digging in to the code, I realise this is also done in `parms.ApplyToTest` - but it now looks these steps aren't done if a test has parameters? I'd rather this logic was moved out of test name setting - and duplicated for test cases if need be. (Or better yet - helper class! 😄)

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
```diff
@@ -424,14 +458,9 @@ private static string EscapeControlChar(char c)
 
         private class TestIDFragment : NameFragment
         {
-            public override string GetText(MethodInfo method, object[] args)
-            {
-                return "{i}"; // No id available using MethodInfo
-            }
-
```
ChrisMaddock: Can you clarify what this change is?

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
```diff
@@ -444,65 +473,71 @@ public FixedTextFragment(string text)
...
+                var method = test.Method;
+                if (method == null) return null;
+
                 var sb = new StringBuilder();
 
                 sb.Append(method.Name);
 
                 if (method.IsGenericMethod)
-                    AppendGenericTypeNames(sb, method);
+                    AppendGenericTypeNames(sb, method.MethodInfo);
```
ChrisMaddock: Why does this need to do `method.MethodInfo` when none of the other uses of `method` have changed?

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
```diff
@@ -444,65 +473,71 @@ public FixedTextFragment(string text)
...
         }
 
         private class MethodFullNameFragment : NameFragment
         {
-            public override string GetText(MethodInfo method, object[] args)
+            public override string GetText(Test test, object[] args, bool argsAreDisplayNames)
             {
+                var method = test.Method;
+                if (method == null) return null;
+
```
ChrisMaddock: As above

ChrisMaddock:
And thanks for the summary - I'd have been lost without it!! 😄

src/NUnitFramework/framework/Internal/TestParameters.cs
```diff
@@ -116,30 +116,42 @@ private void InitializeArguments(object[] args)
...
+            {
+                test.Name = defaultTestNameGenerator.GetDisplayName(test, ArgDisplayNames, ArgDisplayNames != null ? null : OriginalArguments);
+            }
+            else
+            {
+                test.Name = TestName.Contains("{")
+                    ? new TestNameGenerator(TestName).GetDisplayName(test, ArgDisplayNames, ArgDisplayNames != null ? null : OriginalArguments)
+                    : TestName;
+            }
+        }
```
ChrisMaddock: I think this is the key bit I find unexpected. Why do TestParameters have responsibility for setting a test name? It feels more natural to me that the Case/FixtureBuilders calll TestNameGenerator, and pass in TestParams.

ChrisMaddock:
Final disclaimer: I'm aware you and Charlie have been discussing this implementation for a while - if any of my comments are the direct opposite of what you've agreed on - apologies!

src/NUnitFramework/framework/Internal/TestParameters.cs
CharliePoole: `TestParameters` has a `RunState` because it can include properties like `Ignore` and can end up as `RunState.Invalid` if the parameters don't match the method to which the parameter applies.


Basically there are a number of components, including `TestParameters` and `TestFixtureBuilder`, which can detect problems calling for setting the `RunState`. Various attributes do the same thing.

ChrisMaddock: Ah, I hadn't realised TestParameters also kept a RunState.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
ChrisMaddock: Ah, presumably where Test is a TestFixture, so there's no method.

jnm2:
@ChrisMaddock Sorry I'm only getting to this now! I love the feedback. Thank you for taking the time!


> If the user sets less/more argnames than actual args, we currently have a silent failure. I'd assume that's more likely to be user error than it would be a useful feature - can we throw?

A major use case I have is to provide fewer display names than arguments. Can you clarify what you mean by silent failure?

> I actually prefer SetArgNames to SetArgDisplay - I don't really know what the latter is. Could we compromise on SetArgDisplayNames? Longer I know... 😞

I'm a sucker for clarity. `SetArgDisplayNames` it is. :+1:

src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
jnm2: I made the builder stop reaching into the parameters and pulling out pieces of data like TestName and ArgNames and possibly future properties, then making decisions based on those properties before applying the result to test.Name.

Practically speaking, it allowed me to make the ArgNames property protected instead of public and also to not duplicate the whole chunk of logic when applying the test fixture name.

Pattern-wise, it's consistent with the direction we're moving with everything that was already in the ApplyTo method. Test parameters decide how to apply the Name rather than querying TestName and now ArgNames and possibly future properties.

src/NUnitFramework/framework/Internal/Builders/NUnitTestFixtureBuilder.cs
jnm2: Yes. Going along with https://github.com/nunit/nunit/pull/2538#discussion_r156253485, the idea is to make the parameters responsible for looking at their private data and uniformly applying themselves whether it's a test case or a test fixture.

When a fixture doesn't have a TestParameters instance, it has to fall back to doing it itself. This is ugly. The fix I'd expect is for TestParameters to always be created for it, just like the test case builder does.

`TestParameters` *is* supposed to be the helper class. 😃 If I made sure there is always a TestParameters instance, would that address your concerns? Or is there another pattern that might be better for us than tell-don't-ask?

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
```diff
@@ -67,30 +71,36 @@ public TestNameGenerator(string pattern)
...
+        /// <param name="test">A test</param>
+        /// <param name="argDisplayNames">
+        /// Overrides argument list display.
+        /// The strings in this list are already display names and will be rendered as is, without quotes.
+        /// </param>
+        /// <param name="args">
+        /// Arguments to be used if <paramref name="argDisplayNames"/> is <see langword="null"/>.
+        /// A string in this list will be rendered with quotes.
+        /// </param>
+        public string GetDisplayName(Test test, string[] argDisplayNames, object[] args)
```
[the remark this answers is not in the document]
jnm2: I could go either way. I believe my thinking was that overloads were too confusing to read but positional distinction would remind you that there were two things going on. If we used two different method names that concern would be addressed, but what would the two names be?

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
```diff
@@ -67,30 +71,36 @@ public TestNameGenerator(string pattern)
...
         {
             if (_fragments == null)
                 _fragments = BuildFragmentList(_pattern);
 
             var result = new StringBuilder();
 
             foreach (var fragment in _fragments)
-                result.Append(fragment.GetText(testMethod, args));
+                result.Append(fragment.GetText(test, argDisplayNames ?? args, argDisplayNames != null));
```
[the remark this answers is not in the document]
jnm2: The reason is that I followed the existing pattern. I think it's a good suggestion, but we lose the ability to reuse the parsed pattern fragment list for more than one test case. I'd consider that a refactor outside the scope of this PR.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
jnm2: I removed the `MethodInfo` overload when I widened the test name generator from `TestMethod` to any `Test`, including fixtures. It was only a helper overload anyway. I feel like everything is cleaner.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
jnm2: This code used to be inside the removed overload with parameter `MethodInfo method`. Now it's inside a `Test` overload and I did `var method = test.Method` rather than `var method = test.Method.MethodInfo`. Now most of the code is operating on our IMethodInfo rather than SR.MethodInfo which I think is what we're meant to do with `IMethodInfo` when possible.

src/NUnitFramework/framework/Internal/TestParameters.cs
jnm2: > TestParameters and TestFixtureBuilder-but-only-under-certain-circumstances

I am sympathetic to this and would have made TestFixtureBuilder consistent with TestCaseBuilder already (always having a TestParameters instance no matter what) but I was afraid that didn't belong in this PR. See https://github.com/nunit/nunit/pull/2538#discussion_r156254117.

src/NUnitFramework/framework/Internal/TestParameters.cs
jnm2: Recap from https://github.com/nunit/nunit/pull/2538#discussion_r156253485: it allows TestParameters to keep its private state private and decide how to apply it. Do we really want to make both `SetArgDisplayNames` and `ArgDisplayNames` public? If another name-generation-state field is added to TestParameters, do we really want to add them in a pair? ... Maybe. It wasn't my first choice. I like the ApplyTo pattern.

jnm2:
@ChrisMaddock 

Continuing from https://github.com/nunit/nunit/pull/2538#discussion_r155955581:

> I'd prefer this to be _nameGenerator.GetDisplayName() in all cases, and for TestNameGenerator to look after all the resolution of testnames, and deal with all the if/else cases which are currently scattered through TestCaseBuilder/TestFixtureBuilder/TestNameGenerator/TestParameters.
> 
> Would that be technically possible, or am I missing why it's currently done this way? Is it just that doing that would be a bigger refactor than this PR has set out to do? It feels like we're changing the structure fundamentally enough at this point for that to be worthwhile.

I don't like exposing `ArgDisplayNames`, but even so this part is intriguing. After we make sure there is always a TestParameters instance for TestFixtureBuilder, what other if/elses are left?

> `_nameGenerator.GetDisplayName()`

Right now TestNameGenerator parses the pattern once and can be used for many test cases with varying arguments and even varying methods. This change would make us reparse the pattern for each test case. Might not be a bad thing. Just wanted to draw attention.

src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
ChrisMaddock: Ah ok. I see the ApplyTo architecture was pre-existing. I'm not sure I'm totally a fan in this instance, where the test name is sometimes set by TestParams, and sometimes fetched from the NameGenerator, but I see when this has been done like it is. 🙂

src/NUnitFramework/framework/Internal/Builders/NUnitTestFixtureBuilder.cs
ChrisMaddock: I'm not sure TestParameters would be the right name for such a class, but that might be a nicer idea in general! I think such a refactor is outside of the scope of this PR now however, I didn't realises some of these points you're mentioning while reviewing. 😄

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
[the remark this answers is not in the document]
ChrisMaddock: Good point - `GetNameFromArgs` and `GetNameFromNamedArgs`? This one isn't work blocking on. 🙂

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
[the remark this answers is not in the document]
ChrisMaddock: I didn't realise fragment lists were reused - thanks. 👍

ChrisMaddock:
> A major use case I have is to provide fewer display names than arguments. I'm often passing supplementary instances or data in which is just visual noise and is already properly represented by the display names I've chosen for the initial arguments. Can you clarify what you mean by silent failure?

Silent failure in that I intend to name all my args, but only actually named three of them. If that's a feature for you, then lets leave it as is. 😄 

> I don't like exposing ArgDisplayNames, but even so this part is intriguing. After we make sure there is always a TestParameters instance for TestFixtureBuilder, what other if/elses are left?

Exposing internally, I'd mean.

It looks like TestCaseBuilder would need the same fix - the testname can currently either come from the params or the generator. Both TestCaseBuilder and TestFixtureBuilder have multiple paths. TestParams then has two different paths depending if a TestName has been set or not, before it goes into TestNameGenerator.

I no longer think that this PR is the right place to do that refactor! 😄 My concern is that test naming logic seems to branch into a number of paths in a number of different classes. It would be nice to have e.g. `TestNameApplier` with e.g. `GimmeTestName(ITest)` which wraps all that logic in one place. That brings us even closer the the `ApplyToTest` architecture you were talking about before.

With your responses however - I think we should get this in as it, and worry about this as a 'future plan'!

> Right now TestNameGenerator parses the pattern once and can be used for many test cases with varying arguments and even varying methods. This change would make us reparse the pattern for each test case. Might not be a bad thing. Just wanted to draw attention.

Not sure what you're replying to here!? 😆

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
jnm2: Yes.

jnm2:
Ah, gotcha! I would be interested in looking at a refactor like that.

(I was replying to the idea that we might refactor to the API `_nameGenerator.GetDisplayName()` which would mean that the three parameters would have to become three fields and force you to create a new test generator instance for each test case.)

So it sounds like you're happy with merging as-is?

ChrisMaddock:
Oh. No, I didn’t actually mean a paramterless method, I was just too lazy to write them out! 😄

Yes, fine by me, sorry, no GitHub review on my phone. 🙁 Maybe worth someone else looking this over given the size of it - but will leave that decision with you.

jnm2:
Alright, ready now! @nunit/framework-team One more please?

jnm2:
🕸 This PR got a bit stale and had to be rebased on master.

The walkthrough is still accurate though some API names have slight changes: https://github.com/nunit/nunit/issues/2536#issuecomment-350503635

Anyone willing to do that second review @ChrisMaddock asked for? 🤓

jnm2:
Ran into another use for this today. 😃

jnm2:
Hey, quick question for everyone just crossed my mind! Would you expect this to mean "set the argument display names back to the defaults" or "set the argument display names to new string[] { null }"?

```cs
testCaseData.SetArgDisplayNames(null);
```

I lean towards thinking that you'd do `testCaseData.ArgDisplayNames = null;` in the rare event that you actually wanted to reverse a possible earlier customization. You'd more likely expect `testCaseData.SetArgDisplayNames(null);` to behave consistently with `testCaseData.SetArgDisplayNames(null, null);`.

This requires a new commit if you agree.

I'm starting to wonder if it would make sense to have a static analysis test that makes sure we spec the `null` behavior for every `params` array we accept in the entirety of the framework. This stuff is too easy to miss and risk a NRE (or worse, a misinterpretation).

jnm2:
@rprouse I can't think of one, other than an over-engineered multi-level process inside a TestCaseSource.

jnm2:
I went ahead and made sure we interpret `.SetArgDisplayNames(null)` as string-typed and not as resetting `ArgDisplayNames`. This is in line with our decisions on other params APIs. Please review!

jnm2:
@nunit/framework-team, is there anything I can do to make it easier to review this PR?

It's worth it to me to break it into smaller PRs if that means it'll make it in 3.10. What do you think?

CharliePoole:
I can't review it because i disagree with the sub-objective of making naming of fixtures use a template approach like we do for test cases. This highlights one of the problems of mixing multiple objectives in a single PR, even if the implementations are related. If you do three things in one PR, I have to agree with all three of them to approve it.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
rprouse: Should we return an empty string instead of null? The only usage I see currently is a `StringBuilder.Append()` which is safe, but it might change in the future.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
rprouse: As above 😄

jnm2:
> You've bitten off a bit too much in this PR and is outside of the scope of the original issue. Could you back up a bit, tackle just what was proposed in #2536 in a new PR?

😄 With pleasure!

I've looked at it a bit. What I do here will look quite different depending on the decision at https://github.com/nunit/nunit/issues/2675, so I plan to wait. Either way this PR will be cut way down to scope.

jnm2:
@Brar If I remember right, we hesitated to implement escaping because that would change the meaning of folks' existing code. In most cases, SetArgDisplayNames would provide a more semantically useful workaround anyway. I can certainly work on this next if that unblocks you.

Brar:
> SetArgDisplayNames would provide a more semantically useful workaround anyway.

I'd be happy with that.

> I can certainly work on this next if that unblocks you.

Yes I think it would but what I'm doing is not a high priority project, so there's no need to hurry.

jnm2:
This PR has a lot of comments and the branch name is outdated. Going to reboot as a new PR.

jnm2:
PR is up: #2919

Issue #1215, opened 2016-01-26
Generic test cases


DzonnyDZ:
Currently NUnit support test methods being generic, but it can only infer which types to supply to generic parameters if parameters of method are of the typeof generic parameter, so the type of generic parameter can be inferred from type of values in `TestCaseAttribute`.
I would like to be able to supply my own types for generic parameters, independently on values passed to the test case. This would allow me to test generic instances instantiated with generic parameters that are currently impossible to pass via parameters (e.g. custom classes), or for which we currently have to use fake parameters for.
The solution seems pretty easy, and can be even implemented outside of NUnit itself, but I would like to see it as part of NUnit, either as extension of current `TestCaseAttribute` or as a new attribute (possibly derived from `TestCaseAttribute`). BTW: It would be quite helpful for extending the framework to make some methods and properties of `TestCaseAttribute` virtual.
Below is an implementation I managed to create in just few minutes. It definitely has some flaws - the test cases defined using this attribute are not nicely shown in VS Test Explorer, nor in ReSharper, however console runner sees them and runs them, which leads me to conclusion that I'm on a right way.
I would like to implement it as a pull request for NUnit myself, but I would need a little bit guidance
- Shall I extend existing NUnit `TestCaseAttribute` or create a new derived one
- In case of writing derived attribute, am I allowed to make appropriate methods and properties of `TestCaseAttribute` virtual?
- How to make the test case name to be nicely generated, like it is currently generated for normal method parameters when `TestName` is not supplied - e.g. `"MethodName`2[System.String, MyNs.MyClass]("param1", 6)"`?

``` c#
[AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
public class GenericTestCaseAttribute : TestCaseAttribute, ITestBuilder, ITestData
{
    public GenericTestCaseAttribute(params object[] arguments) : base(arguments) { }
    public Type[] Types { get; set; }
    IEnumerable<TestMethod> ITestBuilder.BuildFrom(IMethodInfo method, Test suite)
    {
        if (method.IsGenericMethodDefinition && Types != null && Types.Length == method.GetGenericArguments().Length)
        {
            var gm = method.MakeGenericMethod(Types);
            return base.BuildFrom(gm, suite);
        }
        return base.BuildFrom(method, suite);
    }
}
```

and usage

``` c#
[GenericTestCase("/aaa/bbb", Types = new[] { typeof(string) }, TestName = "GetMetadata_null<string>")]
[...]

PR #2546, opened 2017-11-08
Fix for Matches<T>(Predicate<T>) not accepting null for nullable types

### LINKED ISSUE #2533: Matches<T>(Predicate<T>) throws ArgumentException or Fails when actual is null
**Class:** `NUnit.Framework.Constraints.ConstraintExpression`
**Method:** `Constraint Matches<TActual>(Predicate<TActual> predicate)`

Hello,
I'm trying to use the Matches constraint that takes a predicate but I'm getting an ArgumentException:

> System.ArgumentException : The actual value is not of type String
> Parameter name: actual

or the test fails:

> Expected: <System.String> and value matching lambda expression
> But was:  null

I would expect the test to succeed/pass without an exception.

The same can be achieved with the Null.Or.Empty constraint, but I need to provide the predicate as parameter to the test method and the parameter cannot depend to the NUnit framework.

I'm assuming that the problem is at the [line 70 of the PredicateConstraint](https://github.com/nunit/nunit/blob/7b87bdc7b48ee1b13023dd282efea449c2d66b81/src/NUnitFramework/framework/Constraints/PredicateConstraint.cs#L70), but I can be wrong.
It checks if the actual passed to the ApplyTo method is of the type defined in the predicate. Since the actual passed is null, the check fails.

Thank you for your time.

## Test code:
```csharp
[TestFixture]
public class MatchesPredicateTests {

    [Test]
    public void HasAll_Matches_predicate_fails_When_actual_is_null_string() {
        var actuals = new[] {"", null};
        Predicate<String> predicate = o => (o == null) || (o == String.Empty);

        Assert.That(actuals, Has.All.Null.Or.Empty); // <-- Success
        Assert.That(actuals, Has.All.Matches(predicate)); // <-- Failure
    }

    [Test]
    public void Matches_predicate_fails_When_actual_is_null_string() {
        var actuals = new[] {"", null};
        Predicate<String> predicate = o => (o == null) || (o == String.Empty);

        foreach (String actual in actuals) {
            Assert.That(actual, Is.Null.Or.Empty); // <-- Success
            Assert.That(actual, Is.TypeOf<String>().And.Matches(predicate)); // <-- Failure
        }
    }

    [Test]
    public void HasAll_Matches_predicate_fails_When_actual_is_null_int() {
        var actuals = new Int32?[] {0, null};
        Predicate<Int32?> predicate = o => (o == null) || (o == 0);

        Assert.That(actuals, Has.All.Null.Or.Zero); // <-- Success
        Assert.That(actuals, Has.All.Matches(predicate)); // <-- Failure
    }

    [Test]
    public void Matches_predicate_fails_When_actual_is_null_int() {
        var actuals = new Int32?[] {0, null};
        Predicate<Int32?> predicate = o => (o == null) || (o == 0);

        foreach (Int32? actual in actuals) {
            Assert.That(actual, Is.Null.Or.Zero); // <-- Success
            Assert.That(actual, Is.TypeOf<Int32?>().And.Matches(predicate)); // <-- Failure
        }
    }

}
```

jnm2:
Fixes #2533.

### DIFF
--- a/src/NUnitFramework/framework/Internal/ConstraintUtils.cs
+++ b/src/NUnitFramework/framework/Internal/ConstraintUtils.cs
@@ -22,7 +22,6 @@
 // ***********************************************************************
 
 using System;
-using System.Globalization;
 
 namespace NUnit.Framework.Internal
 {
@@ -32,16 +31,18 @@ namespace NUnit.Framework.Internal
     internal static class ConstraintUtils
     {
         /// <summary>
-        /// Require that the provided object is actually of the type required,
-        /// and that it is not null.
+        /// Requires that the provided object is actually of the type required.
         /// </summary>
         /// <param name="actual">The object to verify.</param>
         /// <param name="paramName">Name of the parameter as passed into the checking method.</param>
-        /// <typeparam name="T">The type to require.</typeparam>  
-        /// <returns>A properly typed object, or throws. Never null.</returns>
-        public static T RequireActual<T>(object actual, string paramName)
+        /// <param name="allowNull">
+        /// If <see langword="true"/> and <typeparamref name="T"/> can be null, returns null rather than throwing when <paramref name="actual"/> is null.
+        /// If <typeparamref name="T"/> cannot be null, this parameter is ignored.</param>
+        /// <typeparam name="T">The type to require.</typeparam>
+        public static T RequireActual<T>(object actual, string paramName, bool allowNull = false)
         {
             if (actual is T) return (T)actual;
+            if (allowNull && actual == null && default(T) == null) return default(T);
 
             var actualDisplay = actual == null ? "null" : actual.GetType().Name;
             throw new ArgumentException($"Expected: {typeof(T).Name} But was: {actualDisplay}", paramName);
--- a/src/NUnitFramework/tests/Constraints/PredicateConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/PredicateConstraintTests.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2009 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -22,6 +22,9 @@
 // ***********************************************************************
 
 using System;
+using System.Linq;
+using System.Reflection;
+using NUnit.Compatibility;
 
 namespace NUnit.Framework.Constraints
 {
@@ -31,12 +34,12 @@ namespace NUnit.Framework.Constraints
         [SetUp]
         public void SetUp()
         {
-            theConstraint = new PredicateConstraint<int>((x) => x < 5 );
+            theConstraint = new PredicateConstraint<int>((x) => x < 5);
             expectedDescription = @"value matching lambda expression";
             stringRepresentation = "<predicate>";
         }
 
-        static object[] SuccessData = new object[] 
+        static object[] SuccessData = new object[]
         {
             0,
             -5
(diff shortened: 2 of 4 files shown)


src/NUnitFramework/framework/Internal/ConstraintUtils.cs
```diff
@@ -32,16 +31,18 @@ namespace NUnit.Framework.Internal
...
-        /// and that it is not null.
+        /// Requires that the provided object is actually of the type required.
         /// </summary>
         /// <param name="actual">The object to verify.</param>
         /// <param name="paramName">Name of the parameter as passed into the checking method.</param>
-        /// <typeparam name="T">The type to require.</typeparam>  
-        /// <returns>A properly typed object, or throws. Never null.</returns>
-        public static T RequireActual<T>(object actual, string paramName)
+        /// <param name="allowNull">
+        /// If <typeparamref name="T"/> can be null, returns null rather than throwing when <paramref name="actual"/> is null.
```
mikkelbu: Perhaps we should state that this only is the case when `allowNull` is true, but perhaps this is implicit?

jnm2: I'm always interested in making things clearer. `If <see langword="true"/> and`?

jnm2:
Oops, thanks for fixing my typo! :D

jnm2:
The tests are difficult to read; I was just complaining about that over at https://github.com/nunit/nunit/issues/1215#issuecomment-342920782 Want to help? \</nerdsnipe> 😁

src/NUnitFramework/framework/Internal/ConstraintUtils.cs
mikkelbu: Sounds good

mikkelbu:
Perhaps next week. Currently, all I've time for are small reviews :(

jnm2:
That's completely fine.

jnm2:
My push dismissed your review, sorry.

jnm2:
Heard. 👍

oznetmaster:
I added some line comments to the PredicateRestrainTests.cs code, forgetting, as usual, that no notification of those comments is sent :(

jnm2:
I can't find any commit comments, and I didn't even force push or anything.

oznetmaster:
[https://github.com/nunit/nunit/commit/34026d878bbc8234b3c3efd179b0cbb9c72b4e3e](url)

src/NUnitFramework/tests/Constraints/PredicateConstraintTests.cs
```diff
@@ -52,5 +55,46 @@ public void CanUseConstraintExpressionSyntax()
...
+        [TestCase(typeof(object))]
+        [TestCase(typeof(string))]
+        [TestCase(typeof(int?))]
+        [TestCase(typeof(AttributeTargets?))]
+        public static void ActualMayBeNullForNullableTypes(Type type)
+        {
+            // https://github.com/nunit/nunit/issues/1215
+            var methodInfo = typeof(PredicateConstraintTests)
+                .GetMethods(BindingFlags.NonPublic | BindingFlags.Static)
+                .Single(method => method.Name == nameof(ActualMayBeNullForNullableTypes) && method.GetParameters().Length == 0)
```
oznetmaster: Isn't the test for the number of parameters redundant since there is only one matching method?

src/NUnitFramework/tests/Constraints/PredicateConstraintTests.cs
```diff
@@ -52,5 +55,46 @@ public void CanUseConstraintExpressionSyntax()
...
+        }
+
+        [TestCase(typeof(int))]
+        [TestCase(typeof(AttributeTargets))]
+        public static void ActualMustNotBeNullForNonNullableTypes(Type type)
+        {
+            // https://github.com/nunit/nunit/issues/1215
+            var methodInfo = typeof(PredicateConstraintTests)
+                .GetMethods(BindingFlags.NonPublic | BindingFlags.Static)
+                .Single(method => method.Name == nameof(ActualMustNotBeNullForNonNullableTypes) && method.GetParameters().Length == 0)
```
oznetmaster: Isn't the test for the number of parameters redundant since there is only one matching method?

oznetmaster:
Well, I added them again...

jnm2:
Wait duh... *this* is the PR where I force-pushed. I got mixed up.

src/NUnitFramework/tests/Constraints/PredicateConstraintTests.cs
jnm2: Yes. I see I overlooked that while juggling different approaches. The compatibility extension methods are far from complete; sometime I am going to flesh them out.

oznetmaster: I only noticed because GetParameters cannot be used on an open generic method in CF.  Since it was redundant anyway, I just removed it in my build.

jnm2: Oh, very interesting!

mikkelbu:
@oznetmaster I got notifications about your comments (both before and now), so I think that it is working.

Ps. I edited your link to make it work.

jnm2:
Huh, that's so weird!

PR #2547, opened 2017-11-09
 Fixes Throws.Exception failing to catch async exceptions

### LINKED ISSUE #1899: Constraint Throws.Exception does not catch exception with async lambdas
After converting tests from NUnit 2.6.4 to 3.5.0, a few of them starting failing for no reason. After a quick diagnostic, all of them use the constraint `Throws.Exception` with an async lambda (don't care about the actual exception type, just checking that it fails on invalid input).

The behavior is that the test case fails by throwing the actual Exception, instead of catching it.

I have other places that use `Throws.Exception` with regular non-async lambda, and all of them work as intended.

Finally, I noticed that if I replace `Throws.Exception` with `Throws.Exception.TypeOf(...)` then the test pass properly with async lambdas.

Simple repro here:
```CSharp
namespace Tests
{
	using System;
	using System.Threading;
	using System.Threading.Tasks;
	using NUnit.Framework;

	[TestFixture]
	public class Test_Throw_Exception
	{

		[Test]
		public void Test_Blocking()
		{
			Assert.That(() => { Thread.Sleep(1000); throw new Exception("KAPOW"); }, Throws.Exception);
			// => PASS
		}

		[Test]
		public async Task Test_Async_1()
		{
			Assert.That(async () => { await Task.Delay(1000); throw new Exception("KAPOW"); }, Throws.Exception);
			// => will throw
		}

		[Test]
		public async Task Test_Async_2()
		{
			Assert.That(async () => { await Task.Delay(1000); throw new Exception("KAPOW"); }, Throws.Exception.TypeOf(typeof(Exception)));
			// => PASS
		}

	}
}
```

With NUnit 2.6.4, all three test cases pass.

With NUnit 3.5.0, tests `Test_Blocking` and `Test_Async_2` pass, but test `Test_Async_1` fails with the following output:
```
Test 'Tests.Test_Throw_Exception.Test_Async_1' failed: System.Exception : KAPOW
	Project\Test.cs(310,0): at Tests.Test_Throw_Exception.<>c.<<Test_Async_1>b__1_0>d.MoveNext()
	--- End of stack trace from previous location where exception was thrown ---
	at NUnit.Framework.Internal.ExceptionHelper.Rethrow(Exception exception)
	at NUnit.Framework.Constraints.Constraint.ApplyTo[TActual](ActualValueDelegate`1 del)
	at NUnit.Framework.Assert.That[TActual](ActualValueDelegate`1 del, IResolveConstraint expr, String message, Object[] args)
	at NUnit.Framework.Assert.That[TActual](ActualValueDelegate`1 del, IResolveConstraint expr)
	Project\Test.cs(310,0): at Tests.Test_Throw_Exception.<Test_Async_1>d__1.MoveNext()
	--- End of stack trace from previous location where exception was thrown ---
	at NUnit.Framework.Internal.ExceptionHelper.Rethrow(Exception exception)
	at NUnit.Framework.Internal.Commands.TestMethodCommand.RunAsyncTestMethod(TestExecutionContext context)

0 passed, 1 failed, 0 skipped, took 1,95 seconds (NUnit 3.5.0).
```

Tested with Resharper, TestDriven.net and NUnit.ConsoleRunner.3.5.0 package, under VS2015

jnm2:
Fixes #1899.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2012 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -22,7 +22,11 @@
 // ***********************************************************************
 
 using System;
+
+#if ASYNC
+using System.Threading.Tasks;
 using NUnit.Framework.Internal;
+#endif
 
 namespace NUnit.Framework.Constraints
 {
@@ -90,13 +94,17 @@ namespace NUnit.Framework.Constraints
         }
 
         /// <summary>
-        /// Returns the ActualValueDelegate itself as the value to be tested.
+        /// Applies the constraint to an ActualValueDelegate that returns
+        /// the value to be tested. The default implementation simply evaluates
+        /// the delegate but derived classes may override it to provide for
+        /// delayed processing.
         /// </summary>
-        /// <param name="del">A delegate representing the code to be tested</param>
-        /// <returns>The delegate itself</returns>
-        protected override object GetTestObject<TActual>(ActualValueDelegate<TActual> del)
+        public override ConstraintResult ApplyTo<TActual>(ActualValueDelegate<TActual> del)
         {
-            return new TestDelegate(() => del());
+#if ASYNC
+            if (typeof(TActual) == typeof(Task)) return ApplyTo(new AsyncTestDelegate(() => (Task)(object)del.Invoke()));
+#endif
+            return ApplyTo(new TestDelegate(() => del.Invoke()));
         }
 
         #region Nested Result Class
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
```diff
@@ -22,7 +22,11 @@
 // ***********************************************************************
 
 using System;
+
+#if ASYNC
+using System.Threading.Tasks;
 using NUnit.Framework.Internal;
```
oznetmaster: Should this be inside the #if/#endif?  It was not previously.

jnm2: Yes. It was only needed by the original conditional ASYNC section.

CharliePoole:
Not  important but FWIW our practice has varied on wrapping using statements in conditionals. Originally, I only did it if the using statement failed to compile. As many contributors started to use tools that gave warnings about such things, we began to wrap them to turn off the warning. I don't use such tools myself, so I'm happy with the original practice, but it's not an issue I care about much one way or the other.

jnm2:
@CharliePoole I think we're in the same mental space. I only do it if it fails to compile (as in this case) but then I also move the related using directives into the same group for clarity.

Issue #2325, opened 2017-07-19
Retry attribute doesn't retry the test.


marcelopontipedra:
I've already used assert.true(), but it doesn't retry the test.

```c#
[Test, Order(2)]
[Retry(3)]
public void Login()
{
    step.BasicActions.LoginAndOpenPart15038(_driver);
    Assert.True(false == _driver.Url.Contains("login.html"));
}
```

jnm2:
I can repro using:
```c#
private static int times = 0;

[Test, Retry(3)]
public void Login()
{
    times++;
    Assert.Fail(times + " times");
}
```

And looking at the message, or debugging.

jnm2:
`Repeat` acts the same. Something weird is going on. I wonder if it's related to running via the VSTest adapter.

jnm2:
I would have expected https://github.com/nunit/nunit/blob/master/src/NUnitFramework/tests/Attributes/RetryAttributeTests.cs to cover it.

jnm2:
Nope, fails in NUnit Console run too.

jnm2:
@CharliePoole Insight would be appreciated whenever you are available. I'm not missing anything obvious, am I?

jnm2:
Retry is working under `TestBuilder.RunTestFixture` but not under direct run. I'm going to add a direct run test since this is the case.

jnm2:
Doing a manual bisect. This problem didn't exist before the multiple asserts feature was introduced.

jnm2:
Looks like 295fd187 is the commit that introduced this.

jnm2:
https://github.com/nunit/nunit/pull/2085 is the culprit.

jnm2:
Understanding why will be a challenge.

jnm2:
Correction: Repeat is fine, I did the test wrong. It's just Retry.

jnm2:
Added failing test: https://github.com/nunit/nunit/compare/fix_retry

jnm2:
While the smell test is good, what I'm really hoping to see come out of this is a test that enforces an invariant behavior between direct runs and ``TestBuilder.RunTestFixture`, because it would be nice if ``TestBuilder.RunTestFixture` reflected reality.

jnm2:
[...]

Issue #2328, opened 2017-07-24
Test execution hangs without any progress.


vharavy:
Recently we started encountering strange issues with our NUnit tests. In the middle of the execution, progress stops. If we attach to the application we see that no test is actually executed - NUnit just waits for the previous test to finish and apparently there is no one to signal that. After some investigation, we believe this is due to the following exception that we keep encountering:

````
System.Threading.ThreadAbortException occurred
  HResult=-2146233040
  Message=Thread was being aborted.
  Source=nunit.framework
  StackTrace:
       at NUnit.Framework.Internal.TestExecutionContext.get_ExecutionStatus() in C:\Sources\NUnit\src\NUnitFramework\framework\Internal\TestExecutionContext.cs:line 275
  InnerException: 
````

The stack trace for this exception is as follows:

````
>	nunit.framework.dll!NUnit.Framework.Internal.TestExecutionContext.ExecutionStatus.get() Line 277	C#
 	nunit.framework.dll!NUnit.Framework.Internal.TestExecutionContext.ExecutionStatus.get() Line 275	C#
 	nunit.framework.dll!NUnit.Framework.Internal.TestExecutionContext.ExecutionStatus.get() Line 275	C#
 	nunit.framework.dll!NUnit.Framework.Internal.TestExecutionContext.ExecutionStatus.get() Line 275	C#
 	nunit.framework.dll!NUnit.Framework.Internal.TestExecutionContext.ExecutionStatus.get() Line 275	C#
 	nunit.framework.dll!NUnit.Framework.Internal.Execution.CompositeWorkItem.CheckForCancellation() Line 145	C#
 	nunit.framework.dll!NUnit.Framework.Internal.Execution.CompositeWorkItem.RunChildren() Line 274	C#
 	nunit.framework.dll!NUnit.Framework.Internal.Execution.CompositeWorkItem.PerformWork() Line 100	C#
 	nunit.framework.dll!NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread() Line 442	C#
 	nunit.framework.dll!NUnit.Framework.Internal.Execution.WorkItem.Execute() Line 226	C#
 	nunit.framework.dll!NUnit.Framework.Internal.Execution.SimpleWorkItemDispatcher.Dispatch(NUnit.Framework.Internal.Execution.WorkItem work) Line 75	C#
 	nunit.framework.dll!NUnit.Framework.Internal.Execution.CompositeWorkItem.RunChildren() Line 285	C#
[...]

PR #2565, opened 2017-11-16
Fix for failure to retry when tests have no setup and teardown

jnm2:
Fixes https://github.com/nunit/nunit/issues/2325

### DIFF
--- a/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
@@ -91,16 +91,10 @@ namespace NUnit.Framework
 
                 while (count-- > 0)
                 {
-                    try
-                    {
-                        context.CurrentResult = innerCommand.Execute(context);
+                    innerCommand.ExecuteAndSetCurrentResult(context);
 
-                        if (context.CurrentResult.ResultState != ResultState.Failure)
-                            break;
-                    }
-                    catch when (count > 0)
-                    {
-                    }
+                    if (context.CurrentResult.ResultState.Status != TestStatus.Failed) break;
+                    if (context.CurrentResult.ResultState.Label == ResultState.Error.Label) break;
 
                     // Clear result for retry
                     if (count > 0)
--- a/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
@@ -91,10 +91,16 @@ namespace NUnit.Framework
 
                 while (count-- > 0)
                 {
-                    context.CurrentResult = innerCommand.Execute(context);
+                    try
+                    {
+                        context.CurrentResult = innerCommand.Execute(context);
 
-                    if (context.CurrentResult.ResultState != ResultState.Failure)
-                        break;
+                        if (context.CurrentResult.ResultState != ResultState.Failure)
+                            break;
+                    }
+                    catch when (count > 0)
+                    {
+                    }
 
                     // Clear result for retry
                     if (count > 0)
--- a/src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2014 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -54,27 +54,21 @@ namespace NUnit.Framework.Internal.Commands
 
             try
             {
-                BeforeTest(context);
-
-                context.CurrentResult = innerCommand.Execute(context);
-            }
-            catch (Exception ex)
-            {
-#if !NETSTANDARD1_6
-                if (ex is ThreadAbortException)
-                    Thread.ResetAbort();
-#endif
-                context.CurrentResult.RecordException(ex);
+                context.ExecuteWithExceptionHandling(c =>
+                {
+                    BeforeTest?.Invoke(c);
+                    c.CurrentResult = innerCommand.Execute(c);
+                });
             }
             finally
             {
                 if (context.ExecutionStatus != TestExecutionStatus.AbortRequested)
-                    AfterTest(context);
+                    context.ExecuteWithExceptionHandling(AfterTest);
             }
 
             return context.CurrentResult;
         }
-        
+
         /// <summary>
         /// Perform the before test action
         /// </summary>
--- a/src/NUnitFramework/framework/Internal/Commands/TestCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TestCommand.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
@@ -31,7 +31,7 @@ namespace NUnit.Framework.Internal.Commands
     /// in the framework. A TestCommand represents a single stage in
     /// the execution of a test, e.g.: SetUp/TearDown, checking for
     /// Timeout, verifying the returned result from a method, etc.
-    /// 
+    ///
     /// TestCommands may decorate other test commands so that the
     /// execution of a lower-level command is nested within that
     /// of a higher level command. All nested commands are executed
@@ -68,7 +68,7 @@ namespace NUnit.Framework.Internal.Commands
         /// </summary>
         /// <param name="context">The <see cref="TestExecutionContext"/> to be used for running the test.</param>
         /// <exception cref="ArgumentNullException">Thrown when <paramref name="context"/> is <see langword="null"/></exception>
-        public void ExecuteAndSetCurrentResult(TestExecutionContext context)
+        internal void ExecuteAndSetCurrentResult(TestExecutionContext context)
         {
             Guard.ArgumentNotNull(context, nameof(context));
 
--- a/src/NUnitFramework/framework/Internal/Commands/TestCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TestCommand.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2010 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -20,6 +20,9 @@
 // OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
+
+using System;
+using System.Threading;
 
 namespace NUnit.Framework.Internal.Commands
 {
@@ -55,11 +58,22 @@ namespace NUnit.Framework.Internal.Commands
         public Test Test { get; private set; }
 
         /// <summary>
-        /// Runs the test in a specified context, returning a TestResult.
+        /// Runs the test in a specified context, returning a <see cref="TestResult"/>.
         /// </summary>
-        /// <param name="context">The TestExecutionContext to be used for running the test.</param>
-        /// <returns>A TestResult</returns>
+        /// <param name="context">The <see cref="TestExecutionContext"/> to be used for running the test.</param>
         public abstract TestResult Execute(TestExecutionContext context);
+
+        /// <summary>
+        /// Runs the test in a specified context, either setting <see cref="TestExecutionContext.CurrentResult"/> to the result or handling the exception.
+        /// </summary>
+        /// <param name="context">The <see cref="TestExecutionContext"/> to be used for running the test.</param>
+        /// <exception cref="ArgumentNullException">Thrown when <paramref name="context"/> is <see langword="null"/></exception>
+        public void ExecuteAndSetCurrentResult(TestExecutionContext context)
+        {
+            Guard.ArgumentNotNull(context, nameof(context));
+
+            context.ExecuteWithExceptionHandling(c => c.CurrentResult = Execute(c));
+        }
 
         #endregion
     }
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -508,7 +508,7 @@ namespace NUnit.Framework.Internal
         /// Invokes <paramref name="action"/> in this context instance, handling all exceptions.
         /// </summary>
         /// <param name="action">May be <see langword="null"/>.</param>
-        public void ExecuteWithExceptionHandling(Action<TestExecutionContext> action)
+        internal void ExecuteWithExceptionHandling(Action<TestExecutionContext> action)
         {
             if (action == null) return;
 
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -504,6 +504,29 @@ namespace NUnit.Framework.Internal
             return context;
         }
 
+        /// <summary>
+        /// Invokes <paramref name="action"/> in this context instance, handling all exceptions.
+        /// </summary>
+        /// <param name="action">May be <see langword="null"/>.</param>
+        public void ExecuteWithExceptionHandling(Action<TestExecutionContext> action)
+        {
+            if (action == null) return;
+
+            try
+            {
+                action.Invoke(this);
+            }
+            catch (Exception ex)
+            {
+#if THREAD_ABORT
+                if (ex is ThreadAbortException)
+                    Thread.ResetAbort();
+#endif
+                if (CurrentResult == null) CurrentResult = CurrentTest.MakeTestResult();
+                CurrentResult.RecordException(ex);
+            }
+        }
+
 #endregion
 
 #region InitializeLifetimeService
--- a/src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
@@ -56,22 +56,7 @@ namespace NUnit.Framework.Internal.Execution
         {
             try
             {
-                Result = MakeTestCommand().Execute(Context);
-            }
-            catch (Exception ex)
-            {
-                // Currently, if there are no command wrappers, test 
-                // actions, setup or teardown, we have to catch any
-                // exception from the test here. In addition, since
-                // users may create their own command wrappers, etc.
-                // we have to protect against unhandled exceptions.
-
-#if !NETSTANDARD1_6
-                if (ex is ThreadAbortException)
-                    Thread.ResetAbort();
-#endif
-
-                Context.CurrentResult.RecordException(ex);
+                MakeTestCommand().ExecuteAndSetCurrentResult(Context);
                 Result = Context.CurrentResult;
             }
             finally
--- a/src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
@@ -72,6 +72,7 @@ namespace NUnit.Framework.Internal.Execution
 #endif
 
                 Context.CurrentResult.RecordException(ex);
+                Result = Context.CurrentResult;
             }
             finally
             {
--- a/src/NUnitFramework/testdata/RetryFixture.cs
+++ b/src/NUnitFramework/testdata/RetryFixture.cs
@@ -42,7 +42,7 @@ namespace NUnit.TestData.RepeatingTests
         public void FailsEveryTime()
         {
             count++;
-            Assert.IsFalse (true);
+            Assert.IsFalse(true);
         }
     }
 
@@ -187,6 +187,13 @@ namespace NUnit.TestData.RepeatingTests
 
             if (Count < 3)
                 Assert.Fail();
+        }
+
+        [Test, Retry(3)]
+        public void FailsEveryTime()
+        {
+            Count++;
+            Assert.Fail();
         }
     }
 }
--- a/src/NUnitFramework/testdata/RetryFixture.cs
+++ b/src/NUnitFramework/testdata/RetryFixture.cs
@@ -175,4 +175,18 @@ namespace NUnit.TestData.RepeatingTests
             Assert.IsTrue(false);
         }
     }
+
+    public sealed class RetryWithoutSetUpOrTearDownFixture
+    {
+        public int Count { get; private set; }
+
+        [Test, Retry(3)]
+        public void SucceedsOnThirdTry()
+        {
+            Count++;
+
+            if (Count < 3)
+                Assert.Fail();
+        }
+    }
 }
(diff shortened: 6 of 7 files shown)


jnm2:
Hang on, I got some tests failing.

jnm2:
I removed something that's the tip of an iceberg and needs to be addressed in a subsequent PR.

Ready for review!

jnm2:
Ready for reviews again!
 
Notes:
(@CharliePoole, you may be interested.)

 1. The fourth commit is a tiny fix but I am beginning to suspect that this fix may fix other issues which might seem unrelated at face value.

 2. In the last commit, I was forced to deal with one of the 21 occurrences of https://github.com/nunit/nunit/issues/2568 in the last commit in RetryAttribute. The original misleading code accidentally causes existing tests to pass but does *not* cause my copies of those existing tests to pass. The only difference is that my copies use fixtures without setup and teardown. 
   
   The expectations are explicit now, and all is well:
   
   ```cs
   if (context.CurrentResult.ResultState.Status != TestStatus.Failed) break;
   if (context.CurrentResult.ResultState.Label == ResultState.Error.Label) break;
   ```

 3. In the last commit, I was forced to duplicate in `RetryAttribute` the same [exception handling in `BeforeAndAfterTestCommand`](https://github.com/nunit/nunit/blob/d596ec027a02ffa85c073d70a6550f3568ee6f8e/src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs#L55-L76), which was already [duplicated in `SimpleWorkItem`](https://github.com/nunit/nunit/blob/d596ec027a02ffa85c073d70a6550f3568ee6f8e/src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs#L57-L75). Rather than every test command having to duplicate the responsibility of catching exceptions (including ThreadAbortException) as well as handling failure states, it's now nicely factored into TestExecutionContext.

   Again, like point 1, I suspect this may fix other issues. For example, I've had reason to believe this issue is connected to https://github.com/nunit/nunit/issues/2328.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
```diff
@@ -91,10 +91,10 @@ public override TestResult Execute(TestExecutionContext context)
 
                 while (count-- > 0)
                 {
-                    context.CurrentResult = innerCommand.Execute(context);
+                    innerCommand.ExecuteAndSetCurrentResult(context);
 
-                    if (context.CurrentResult.ResultState != ResultState.Failure)
-                        break;
+                    if (context.CurrentResult.ResultState.Status != TestStatus.Failed) break;
+                    if (context.CurrentResult.ResultState.Label == ResultState.Error.Label) break;
```
CharliePoole: This is odd when you read it. I guess you're avoiding use of the string but we use them elsewhere. Should we have manifest constants for the known strings? 

Still I'm OK with this if you plan to come back to it right away on the general issue. But if itd likely to sit for a while can we make this a but clearer?

src/NUnitFramework/framework/Internal/Commands/TestCommand.cs
```diff
@@ -55,12 +58,23 @@ public TestCommand(Test test)
...
+        /// Runs the test in a specified context, either setting <see cref="TestExecutionContext.CurrentResult"/> to the result or handling the exception.
+        /// </summary>
+        /// <param name="context">The <see cref="TestExecutionContext"/> to be used for running the test.</param>
+        /// <exception cref="ArgumentNullException">Thrown when <paramref name="context"/> is <see langword="null"/></exception>
+        public void ExecuteAndSetCurrentResult(TestExecutionContext context)
+        {
+            Guard.ArgumentNotNull(context, nameof(context));
+
+            context.ExecuteWithExceptionHandling(c => c.CurrentResult = Execute(c));
+        }
```
CharliePoole: To my own eye, this shortens the code but obscures it. Others should weigh in.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
jnm2: Maybe. Would an `ResultState.ErrorLabel` or `ResultStateLabels.Error` constant be any clearer or nicer?
I can do `"Error"` too if that's you preference and it's more consistent with the codebase that way.

src/NUnitFramework/framework/Internal/Commands/TestCommand.cs
jnm2: Shortening is not a goal I have for this. I would like to make it clearer. What would you suggest?

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -501,6 +501,29 @@ private TestExecutionContext CreateIsolatedContext()
...
+            {
+#if PARALLEL
+                if (ex is ThreadAbortException)
+                    Thread.ResetAbort();
+#endif
+                if (CurrentResult == null) CurrentResult = CurrentTest.MakeTestResult();
+                CurrentResult.RecordException(ex);
+            }
+        }
+
```
CharliePoole: I like the idea of such a method but this seems a tad too general for clarity. What if it took a work test command rather than an action.

jnm2: I'm open to ideas, but that would break the usage in BeforeAndAfterTestCommand:
https://github.com/nunit/nunit/pull/2565/files#diff-d7f76a7573ec9da3e9654efe7393eaddR57

jnm2:
@nunit/framework-team Charlie and I are both asking for other reviewers, fyi.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
ChrisMaddock: This is a semantic change, right? Retry will now Retry on both failure and error? Or has this functionality just moved?

I have no problem with that if it is, just want to understand what the change is. 🙂

src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
```diff
@@ -54,27 +54,21 @@ public override TestResult Execute(TestExecutionContext context)
...
-            catch (Exception ex)
-            {
-#if !NETSTANDARD1_3 && !NETSTANDARD1_6
-                if (ex is ThreadAbortException)
-                    Thread.ResetAbort();
-#endif
-                context.CurrentResult.RecordException(ex);
+                context.ExecuteWithExceptionHandling(c =>
+                {
+                    BeforeTest?.Invoke(c);
```
ChrisMaddock: I haven't pulled the code to look at this myself, but why do you check BeforeTest for being null, but not AfterTest?

ChrisMaddock: Oh, never mind. `ExecuteWithExceptionHandling` handles nulls.

src/NUnitFramework/framework/Internal/Commands/TestCommand.cs
```diff
@@ -55,12 +58,23 @@ public TestCommand(Test test)
         public Test Test { get; private set; }
 
         /// <summary>
-        /// Runs the test in a specified context, returning a TestResult.
+        /// Runs the test in a specified context, returning a <see cref="TestResult"/>.
         /// </summary>
-        /// <param name="context">The TestExecutionContext to be used for running the test.</param>
-        /// <returns>A TestResult</returns>
+        /// <param name="context">The <see cref="TestExecutionContext"/> to be used for running the test.</param>
         public abstract TestResult Execute(TestExecutionContext context);
```
ChrisMaddock: Should `Execute` be called anywhere except in `ExecuteAndSetCurrentResult` now?

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
jnm2: No, this code avoids the semantic change. The existing tests enforcing no retry on error are untouched and the tests I added follow the same semantics.
This line makes sure retry still does not happen if there is an error:
```cs
if (context.CurrentResult.ResultState.Label == ResultState.Error.Label) break;
```

src/NUnitFramework/framework/Internal/Commands/TestCommand.cs
ChrisMaddock: Only certain test command's handle results and exceptions though - don't they? Coming to this now, my question is why do some places do `Execute`, and some do `ExecuteAndSetCurrentResult`? I'm not sure how better to achieve that.

Did you intend to make `Execute()` return void here?

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
jnm2: The reason why it's confusing to look at the diff is that the original line does not do what it looks like it does.

ChrisMaddock: I think I'm a bit confused as to where the original bug was! 😆 Can you clarify why both checks are needed?

ChrisMaddock: Oh, does `Error` come under `TestStatus.Failed`?

src/NUnitFramework/framework/Internal/Commands/TestCommand.cs
jnm2: Execute passes the responsibility of handling exceptions up to the test command above you (or the SimpleWorkItem if you're the top). Replacing `Execute` with `ExecuteAndSetCurrentResult` introduces semantic (error vs failure) changes so it should be a case-by-case look.

`Execute` can't return void without forcing us to rewrite all uses of it, but maybe we should?
`ExecuteAndSetCurrentResult` returns void because the entire system requires you to track the result as a side effect on the context. Returning the result would mislead you into thinking that the system is primarily functional rather than side-effecting.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
jnm2: Each of the three "fixes" commits fixes a separate bug. :D Looking commit-by-commit will probably be enlightening because I tried to keep each one simple.
Yes, `Error` is `TestStatus.Failed` and `Label = "Error"`.

ChrisMaddock: Well I'm glad you've opened an issue about making this simpler! 😄

src/NUnitFramework/framework/Internal/Commands/TestCommand.cs
ChrisMaddock: > Execute can't return void without forcing us to rewrite all uses of it, but maybe we should?

Are there any more? I had a quick look - I think that may be a good idea to make the API clearer. 🙂 

Thanks for the explanation.

ChrisMaddock: Looking through this again - I think this is probably more confusing trying to work out the diff, then it would be to someone coming fresh to the codebase after the change is in. I like that. 🙂

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
jnm2: @Chris Now that you mention it, it would probably be a good idea to make `Execute` protected now rather than later and make everyone use it. Right now there are three kinds of failure every delegating test command must handle:

- innerCommand.Execute returning Failed without "Error" label
- innerCommand.Execute returning Failed with "Error" label
- innerCommand.Execute throwing or letting an exception bubble through

Having two ways to signify the same thing causes a chill to go down my spine :D that's where bugs spontaneously generate.

src/NUnitFramework/framework/Internal/Commands/TestCommand.cs
jnm2: I'll make it protected and see what it would look like to add a commit making everything uniform.

jnm2: This does not play well with `ExpectedExceptionCommand` and I'd rather tackle this in its own PR so long as this PR isn't creating a problem.

ChrisMaddock:
Good catch making them internal - I'd missed that.

Based on discussions above, and two follow-up PRs for `TestResult` things and `TestCommand.Execute`, this looks good to me.

jnm2:
I realized that I needed to make the new APIs internal for now until I open an issue for a related discussion on fixing  a problem with the test command delegation API. 

(In short, we have potential problems lurking such as ExpectedExceptionCommand only catching exceptions and not checking to to see if the inner command has caught and recorded an exception. We have two ways to do the same thing and it means double the work for every test command, work we're currently **missing**. https://github.com/nunit/nunit/pull/2565#discussion_r151843617)

The APIs need to be internal because the classes are used by anyone using our wrapping API and I foresee the need to change them depending on how this future discussion goes.

jnm2:
I appreciate what you brought up, Chris. Thanks for the review.
Due to the nature of the conversation above I would feel better if at least one other @nunit/framework-team member signed off on this.

jnm2:
I rebased on master. I guess I shouldn't have waited for two reviews. @ChrisMaddock would you mind re-reviewing? I didn't change anything of substance.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
```diff
@@ -91,10 +91,10 @@ public override TestResult Execute(TestExecutionContext context)
 
                 while (count-- > 0)
                 {
-                    context.CurrentResult = innerCommand.Execute(context);
+                    innerCommand.ExecuteAndSetCurrentResult(context);
```
CharliePoole: The separate method to do two things seems wrong to me and makes this harder to read IME. What's wrong in principle with the idea that the __caller__ should decide what to do with the returned result? It's a pretty fundamental assumption in the `Command` structure and I worry about how invalidating that assumption will affect other things.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
CharliePoole: This code that separately tests status and label makes no sense to me, especially now that you fixed the problem whereby we were using object equality. It peeks into the implementation of a test outcome (`ResultState`) and deigns to parse the internals, which can easily change. It also changes behavior when the `Label` is neither empty nor "Error".

I think we might want to further encapsulate test outcome rather than making use of the internals.

CharliePoole: The three things you list are not three different ways to signify the same thing. They are three different things. Not only that, but you have underestimated how many different things there are... it's not just with and without "Error" label, it's an unlimited number of possible `Label` values. We currently use empty (a simple failure), "Error", "Cancelled" and "Invalid". We have discussed adding a few other internal values and users are expected to define their own (future) labels.

src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
```diff
@@ -54,27 +54,21 @@ public override TestResult Execute(TestExecutionContext context)
...
-                context.CurrentResult = innerCommand.Execute(context);
-            }
-            catch (Exception ex)
-            {
-#if !NETSTANDARD1_6
-                if (ex is ThreadAbortException)
-                    Thread.ResetAbort();
-#endif
-                context.CurrentResult.RecordException(ex);
+                context.ExecuteWithExceptionHandling(c =>
```
CharliePoole: This make the `TestExecutionContext` know something about how Commands work and how to handle exceptions. I've been slowly moving in the opposite direction: Commands know about Context and not the other way around. I have to admit it's hard to make gradual changes when the project now has lots of participants and might be a bad idea. Something for my next project. 😄

CharliePoole:
I've looked and added comments. I don't like the design changes as stated before. I'm avoiding doing an actual review at this point, since you guys are the ones who need to maintain it in future.

CharliePoole:
That is an actual "GitHub" review...

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
jnm2: To answer multiple comments at once: 

Everything I'm reading shows that the Command structure relies heavily on the most current result always being set to the execution context. I was confronted with this during my three "fix" commits in this PR. Every command which need to be aware of result status is responsible to **both** use try...catch (in case the inner command doesn't catch its own exceptions) **and also** check the inner command's result if it returns without exception to see if it had caught and recorded an exception. So whatever we do, we need a helper method to make sure these two tasks stay in sync.

Since many commands are tied so intimately to the execution context's CurrentResult property, I figured this was the shortest path to a safe place. My preference would definitely be to remove all side-effects from the Command structure and make the returned TestResult be the single source of truth but even then, delegating commands need a helper method to catch exceptions and merge them into the inner command's result. Does that make sense?

src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
jnm2: This is that helper method which I can't see any way around. It doesn't have to be on `TestExecutionContext` for sure, but it needs to be somewhere. See https://github.com/nunit/nunit/pull/2565#discussion_r155956395.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
jnm2: No, but two of the three are. You're right, there are many more. The current system makes it difficult to property implement test commands which need to react differently to different kinds of failure, as RetryAttribute is currently specced to do.

src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
jnm2: Also: we definitely needed to deduplicate the code which ended up in this helper method. Leaving things the way they were is not an option.

jnm2:
@CharliePoole At the end of the day, I just need to find the best fix for RetryAttribute. I went with what I see as the most future-proof change to make subtle bugs less likely, but it's certainly fair to not like the design changes and I'm okay with reversing them (if I can find an alternative that works). In my mind they are making a scary system more understandable but it's possible both of us would prefer a third design.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
jnm2: I think making use of the internal label is justified because that is exactly what the test is specced to do. The code transparently shows that. It's *not* specced to react differently if the site fails to match, so I could not consider comparing the `ResultState` directly.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
CharliePoole: It certainly makes for simpler code if you do it that way. However...

Nobody should be handling exceptions if they don't know what those exceptions mean. In the case of a user-provided BeforeAfterCommand, we don't actually know what an exception means if it once escapes from the Command. This tells me that each command should be responsible for handling its own exceptions. At a minimum, it should translate the exception into a proper result and return it. From your comments, I think we agree up to that point at least.

That, BTW, was how the command structure worked before some changes messed it up. I dislike having to have a try / catch in every command, but that's what I would have done initially - in keeping with my mantra of creating duplication before removing it. 😄 I have trouble with an approach based on first __imagining__ the duplication and then mentally removing it because I've watched so many teams mess up by doing exactly that. That's off-topic here, however.

Where I think we disagree is with the notion that the returned `TestResult` should also be set into the `TestExecutionContext`. The Command structure doesn't - or shouldn't - depend on this result being set correctly. Rather, the code in `WorkItem` depends on it. It needs to be set after an entire chain of commands executes. We can achieve that by having each command __exclusively__ make use of the result that is returned to it and not the current result. Doing that has the side effect that you can test Commands without requiring more than a dummy context.

That's my view for your consideration.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
CharliePoole: There is only one __Failure__ value for ResultState and that's the one that `RetryAttribute` is supposed to repeat on as currently designed. Making a negative test is what opens the door to other values. As written now, it will repeat even if the user cancels the test!

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
jnm2: Well, I'm not talking about imagined duplication. [SimpleWorkItem](https://github.com/nunit/nunit/blob/3c028034e717c79bf00661c5852b358f9097d2e1/src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs#L57-L75) [BeforeAndAfterTestCommand](https://github.com/nunit/nunit/blob/3c028034e717c79bf00661c5852b358f9097d2e1/src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs#L55-L68)
And notice what we don't have in BeforeTestCommand and AfterTestCommand...

Great! We don't actually disagree there! The problem is that we currently do everything in terms of the execution context's CurrentResult, so I say we should either double down on what we're already doing and make it safe and consistent (which is what you see here) or we should rewrite much of the Command structure. I considered my choice of sticking with the current pattern to be a reasonable compromise but I'm not opposed to an overhaul. I just don't think we should leave it inconsistent. To me, that's asking for misunderstandings as future features and fixes are implemented.

jnm2: Please don't misunderstand! I'm not disagreeing with your vision of making the command structure purer. That would be my ideal for sure! I'm used to being told that I need to be more pragmatic and not to make such big changes. 😜 But it's urgent to me to move to *some* safe place where there is a pit of success for us and even for NUnit consumers when implementing test commands. Perhaps not this PR then.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
CharliePoole: But the `Site`, at this point, is guaranteed to be `FailureSite.Test`. I think you are trying to make this more general than it really is - we only do retry on individual test cases.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
CharliePoole: Right... I could hypothetically review this PR in two modes: (1) a small fix that got too big (2) a restructure that didn't go far enough. Maybe we'll opt for (3) a fix that initiates a future restructure.

While I abhor the phrase "pit of success" I agree that we should make it relatively easy for a user to write a Command, that being one of our supported extensions. It seems easier to me if the user merely has to return a result and not handle errors. In fact, even if the user is supposed to handle errors, we should probably allow for user error and handle them ourselves.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
jnm2: That's not obvious to me. How could I have figured out by myself that site is guaranteed to be Test?

jnm2:
To address this feedback, I will redo the fix without making any design changes and probably without comparing individual components of ResultState.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
CharliePoole: I suppose you only could have by a very wide review of the code, much wider than you would want to do for the purpose of this issue. So it's understandable that you wouldn't know that.

OTOH, I don't think it's a really great pattern for every member of a team to have to figure out things on their own. We need to have ways to tell each other the stuff we know and to rely on one another for information. In this case, I thought it was something you already knew.

Which is not to say that the internal design of ResultState couldn't be better. Clearly it could. But I think that the improvement should be in applying the Open Closed Principle so it works better without revealing more of it's internals.

jnm2: Oh! I'm finally seeing this comment. That put it into focus for me. Thanks!

rprouse:
> Which is not to say that the internal design of ResultState couldn't be better. Clearly it could. But I think that the improvement should be in applying the Open Closed Principle so it works better without revealing more of it's internals.

I agree that this could be better. If I haven't looked at it in awhile, I always need to go back and refresh my memory. 😄 

@jnm2 should we close this? I agree that it is a bit too much of an architectural change for this issue.

jnm2:
@rprouse I was about to suggest that, to make reviewing the final PR easier. I've `git reset head~~ --hard` (that's Greek for dropping the last two commits) and as we speak am doing a minimal fix. What do you think?

jnm2:
I'm ready to push now and have it reviewed. 😄

jnm2:
Actually, if you don't mind, I'm really curious what GitHub will do when I push the branch with the PR closed.

jnm2:
Huh. Cool. It keeps my old commits. 😄 

New PR: https://github.com/nunit/nunit/pull/2615

PR #2575, opened 2017-11-24
Add -- and != operators to ResultState

### LINKED ISSUE #2568: ResultState API is confusing
I was editing here and ReSharper helpfully warned that that the `!=` comparison was actually doing `!ReferenceEquals(context.CurrentResult.ResultState, ResultState.Failure)`.

https://github.com/nunit/nunit/blob/1dbea8eb4e9554fa182c2358da51bade4d4dfc82/src/NUnitFramework/framework/Attributes/RetryAttribute.cs#L92-L97

That's a good point and well worth fixing; it means that these comparisons **return true:**

`new ResultState(TestStatus.Failed) != ResultState.Failure`
`ResultState.Failure.WithSite(x) != ResultState.Failure.WithSite(x)`

So, why not fix it? Surely the intent in RetryAttribute was to check for *any* failure, not just that the returned instance is the same instance stored in the `ResultState.Failure` field?

```cs
if (context.CurrentResult.ResultState.Status != TestStatus.Failed)
    break;
```

Suddenly, tests start to fail. **A non-obvious side effect of the original misleading code is that our tests pass.** This worries me.

There are 17 such occurrences. In each, it looks like we're comparing ResultStates as though they were enums– checking for simple success, failure, etc. But the reality is we're merely comparing the references to see if the ResultStates are the same instance.

CharliePoole:
Fixes #2568 if we can all agree this is the right approach.

### DIFF
--- a/src/NUnitFramework/framework/Interfaces/ResultState.cs
+++ b/src/NUnitFramework/framework/Interfaces/ResultState.cs
@@ -208,7 +208,7 @@ namespace NUnit.Framework.Interfaces
         public override bool Equals(object obj)
         {
             var other = obj as ResultState;
-            if (other == null) return false;
+            if (object.ReferenceEquals(other, null)) return false;
 
             return Status.Equals(other.Status) && Label.Equals(other.Label) && Site.Equals(other.Site);
         }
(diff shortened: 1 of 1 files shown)


rprouse:
Looks fine to me, but it is @jnm2's issue, so please take a look.

src/NUnitFramework/framework/Interfaces/ResultState.cs
```diff
@@ -226,6 +226,29 @@ public override int GetHashCode()
...
+        public static bool operator ==(ResultState left, ResultState right)
+        {
+            if (object.ReferenceEquals(left, null))
+                return object.ReferenceEquals(right, null);
+
+            return left.Equals(right);
+        }
+
+        /// <summary>
+        /// Overload !+ operator for ResultStates
```
jnm2: Typo `!+`

jnm2:
I disagree that it fixes #2568 though– I'm positive that each of those enum-looking comparisons is going to end up being changed to an actual enum comparison or else a `.Matches` test.

CharliePoole:
I'll fix the typo and merge.

I'll leave the issue to you but I don't think you're right. I couldn't find any problem with the existing uses of == or !=. That's not surprising to me because the ones that needed to change would likely have been changed long ago. Those remaining expect Site to have it's default value of Test unless otherwise specified.

PR #2578, opened 2017-11-26
Correct handling of warnings in TearDown

### LINKED ISSUE #2577: Warning in TearDown is inconsistent with Assertion failure
When an Assertion is used in `TearDown` and fails, an error is generated. That is, the result is `Failed:Error` rather than simply `Failed`. Similarly, failure of an Assumption generates an error.

NUnit has worked this way historically because it is generally necessary for the `TearDown` method to complete, in order to clean things up.

As we implemented Warnings, a Warning failure in `TearDown` works differently from Assertion failure. The assertion is counted and a Warning level `AssertionResult` is created, but it has no effect on the overall result of the test. If there are no other warnings or failures, the test passes.

Alternatives:

1. Document that Warnings in TearDown have no effect on the result of the test although they may be reported by some runners. This seems confusing to me. If we are going to report it, then the test should have an overall warning result.

2. Handle Warnings in TearDown by changing the result to Warning if the test passed. This sounds like it could be useful but is still inconsistent with failing Assertions.

3. Generate an error for failed warnings in TearDown just as we do for assertion failures.

A Warning in `TearDown` generally gives seems to be ignored entirely

CharliePoole:
Fixes #2577 

First commit merely adds tests. The three ignored tests are the heart of the issue. Two of the tests succeed as written, but I'm not sure they really should. One of them fails. They should all work the same way, obviously.

To proceed, we need to select from the options in #2577 as to how we want a warning in TearDown to work.


jnm2:
@CharliePoole [Five team members](https://github.com/nunit/nunit/issues/2577#issuecomment-347471294) support option 2. Is there anything else to do before this PR is ready to review?

CharliePoole:
@jnm2 Yes, implement it. 😄 

I appear to have done a PR on this issue with only tests added to show what is currently happening but never assigned the issue to myself. I can't remember if that was deliberate or not. In any case, I track my issues (not PRs) using a GitHub query and naturally this one doesn't show up!

I'll look to see what's involved.

CharliePoole:
I rebased and added the implementation. Warnings now function as warnings even in teardown.

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
```diff
@@ -119,7 +119,7 @@ private TestCommand MakeTestCommand()
                 foreach (var item in setUpTearDownList)
                     command = new SetUpTearDownCommand(command, item);
 
-                // In the current implementation, upstream actions only apply to tests. If that should change in the future,
+                // In the current implementation, upstream actions only apply to tests. If that should cha nge in the future,
```
jnm2: Accidental keypress.

src/NUnitFramework/tests/Assertions/WarningTests.cs
```diff
@@ -172,6 +172,27 @@ public void WarningFails(string methodName, string expectedMessage)
...
+        [TestCase(typeof(WarningInTearDownFails), nameof(WarningInTearDownFails.WarningPassesInTest), 2, 1)]
+        [TestCase(typeof(WarningInTearDownFails), nameof(WarningInTearDownFails.WarningFailsInTest), 2, 2)]
+        [TestCase(typeof(WarningInTearDownFails), nameof(WarningInTearDownFails.ThreeWarningsFailInTest), 4, 4)]
+        public void WarningUsedInSetUpOrTearDown(Type fixtureType, string methodName, int expectedAsserts, int expectedWarnings)
+        {
+            var result = TestBuilder.RunTestCase(fixtureType, methodName);
+
+            Assert.That(result.ResultState, Is.EqualTo(expectedWarnings == 0 ? ResultState.Success : ResultState.Warning));
+            Assert.That(result.AssertCount, Is.EqualTo(expectedAsserts), "Incorrect AssertCount");
+            Assert.That(result.AssertionResults.Count, Is.EqualTo(expectedWarnings), $"There should be {expectedWarnings} AssertionResults");
```
jnm2: We could consider testing separately that `result.AssertCount` is consistent with `result.AssertionResults.Count` foir a wider range of inputs so that tests like this don't have to bother.

CharliePoole: Well, these are the only tests of the warning assertion. So it makes sense to test that (1) each warning statement counts as an assert and (2) each failed warning adds an assertion result. True, we could organize differently, testing all counting in one place but I think it's better to organize tests by user-visible functionality where possible.

PR #2586, opened 2017-11-29
Reduction of platform-conditional preprocessor

jnm2:
Finishes what I'm assigned to with https://github.com/orgs/nunit/teams/framework-and-engine-team/discussions/3.

<s>The first four commits should be looked at first and each on its own. Each of those was called out in its own commit because each intentionally affects behavior.

In order to do some of this, I had to add package references to System.Diagnostics.Process and System.Xml.XmlDocument (see the nuspec update commit).
System.Xml.XPath.XmlDocument was also useful to reference from our tests but it does not affect the product.

There's a lot to cover so please ask for explanations if something stands out.</s>

See https://github.com/nunit/nunit/pull/2586#issuecomment-349806336

### DIFF
--- a/src/NUnitFramework/framework/Internal/ExceptionHelper.cs
+++ b/src/NUnitFramework/framework/Internal/ExceptionHelper.cs
@@ -34,7 +34,7 @@ namespace NUnit.Framework.Internal
     /// </summary>
     public class ExceptionHelper
     {
-#if !NET45 && !NETSTANDARD1_6
+#if NET20 || NET35 || NET40
         private static readonly Action<Exception> PreserveStackTrace;
 
         static ExceptionHelper()
@@ -60,11 +60,11 @@ namespace NUnit.Framework.Internal
         /// <param name="exception">The exception to rethrow</param>
         public static void Rethrow(Exception exception)
         {
-#if NET45 || NETSTANDARD1_6
-            System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(exception).Throw();
-#else
+#if NET20 || NET35 || NET40
             PreserveStackTrace(exception);
             throw exception;
+#else
+            System.Runtime.ExceptionServices.ExceptionDispatchInfo.Capture(exception).Throw();
 #endif
         }
 
--- a/src/NUnitFramework/tests/Attributes/PairwiseTests.cs
+++ b/src/NUnitFramework/tests/Attributes/PairwiseTests.cs
@@ -68,18 +68,6 @@ namespace NUnit.Framework.Attributes
         // http://www.freequality.org/sites/www_freequality_org/documents/tools/Tagarray_files/tamatrix.htm
         static internal object[] cases = new object[]
         {
-            /* Original:
-            new TestCaseData( new int[] { 2, 4 }, 8, 8 ).SetName("Test 2x4"),
-            new TestCaseData( new int[] { 2, 2, 2 }, 5, 4 ).SetName("Test 2x2x2"),
-            new TestCaseData( new int[] { 3, 2, 2 }, 6, 6 ).SetName("Test 3x2x2"),
-            new TestCaseData( new int[] { 3, 2, 2, 2 }, 7, 6 ).SetName("Test 3x2x2x2"),
-            new TestCaseData( new int[] { 3, 2, 2, 2, 2 }, 8, 6 ).SetName("Test 3x2x2x2x2"),
-            new TestCaseData( new int[] { 3, 2, 2, 2, 2, 2 }, 9, 8 ).SetName("Test 3x2x2x2x2x2"),
-            new TestCaseData( new int[] { 3, 3, 3 }, 12, 9 ).SetName("Test 3x3x3"),
-            new TestCaseData( new int[] { 4, 4, 4 }, 22, 16 ).SetName("Test 4x4x4"),
-            new TestCaseData( new int[] { 5, 5, 5 }, 34, 25 ).SetName("Test 5x5x5")
-            */
-
             new TestCaseData( new int[] { 2, 4 }, 8, 8 ).SetName("Test 2x4"),
             new TestCaseData( new int[] { 2, 2, 2 }, 4, 4 ).SetName("Test 2x2x2"),
             new TestCaseData( new int[] { 3, 2, 2 }, 6, 6 ).SetName("Test 3x2x2"),
--- a/src/NUnitFramework/tests/Attributes/PairwiseTests.cs
+++ b/src/NUnitFramework/tests/Attributes/PairwiseTests.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2009 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -68,7 +68,7 @@ namespace NUnit.Framework.Attributes
         // http://www.freequality.org/sites/www_freequality_org/documents/tools/Tagarray_files/tamatrix.htm
         static internal object[] cases = new object[]
         {
-#if ORIGINAL
+            /* Original:
             new TestCaseData( new int[] { 2, 4 }, 8, 8 ).SetName("Test 2x4"),
             new TestCaseData( new int[] { 2, 2, 2 }, 5, 4 ).SetName("Test 2x2x2"),
             new TestCaseData( new int[] { 3, 2, 2 }, 6, 6 ).SetName("Test 3x2x2"),
@@ -78,7 +78,8 @@ namespace NUnit.Framework.Attributes
             new TestCaseData( new int[] { 3, 3, 3 }, 12, 9 ).SetName("Test 3x3x3"),
             new TestCaseData( new int[] { 4, 4, 4 }, 22, 16 ).SetName("Test 4x4x4"),
             new TestCaseData( new int[] { 5, 5, 5 }, 34, 25 ).SetName("Test 5x5x5")
-#else
+            */
+
             new TestCaseData( new int[] { 2, 4 }, 8, 8 ).SetName("Test 2x4"),
             new TestCaseData( new int[] { 2, 2, 2 }, 4, 4 ).SetName("Test 2x2x2"),
             new TestCaseData( new int[] { 3, 2, 2 }, 6, 6 ).SetName("Test 3x2x2"),
@@ -88,7 +89,6 @@ namespace NUnit.Framework.Attributes
             new TestCaseData( new int[] { 3, 3, 3 }, 9, 9 ).SetName("Test 3x3x3"),
             new TestCaseData( new int[] { 4, 4, 4 }, 17, 16 ).SetName("Test 4x4x4"),
             new TestCaseData( new int[] { 5, 5, 5 }, 25, 25 ).SetName("Test 5x5x5")
-#endif
         };
 
         [Test, TestCaseSource("cases")]
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -364,37 +364,26 @@ namespace NUnit.Framework.Api
         /// </summary>
         /// <param name="targetNode">Target node</param>
         /// <returns>The new node</returns>
-#if NETSTANDARD1_6
         public static TNode InsertEnvironmentElement(TNode targetNode)
         {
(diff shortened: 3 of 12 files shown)


jnm2:
Also @nunit/framework-team this is in the way of me adding netstandard2.0 to our build. I know these PRs are demanding and I appreciate your patience.

src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
```diff
@@ -415,72 +417,36 @@ public static MethodInfo GetGetMethod(this PropertyInfo pinfo, bool nonPublic)
...
-        /// parameter since it doesn't make sense on Assemblies. This version just ignores the
-        /// inherit parameter.
-        /// </summary>
-        /// <param name="asm">The assembly</param>
-        /// <param name="attributeType">The type of attribute you are looking for</param>
-        /// <param name="inherit">Ignored</param>
-        /// <returns></returns>
-        public static object[] GetCustomAttributes(this Assembly asm, Type attributeType, bool inherit)
-        {
-            return asm.GetCustomAttributes(attributeType).ToArray();
```
jnm2: This had also been incorrectly returning an `Attribute[]` instance, whereas the BCL method it's polyfilling returns an `object[]` instance.

mikkelbu:
@jnm2 I think I'll have time to look at it tonight (aka in 12 hours).

nuget/framework/nunit.nuspec
```diff
@@ -30,6 +30,8 @@ Supported platforms:
       <group targetFramework="net45" />
       <group targetFramework="netstandard1.6">
         <dependency id="NETStandard.Library" version="1.6.0" />
+        <dependency id="System.Diagnostics.Process" version="4.3.0" />
+        <dependency id="System.Xml.XmlDocument" version="4.3.0" />
```
ChrisMaddock: Oops, these are in two PRs! To copy my comments across from the other:

Both of these packages look like they're not actually supported on iOS/Android. 😞 I haven't tried it, but unzipping the packages, they have those _._ files...

We should figure this out one way or the other before merging. Adding dependencies to netstandard is what broke everything last time. 😞

src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
```diff
@@ -66,13 +66,15 @@ private static IEnumerable GetData(object[] data, Type targetType)
                     continue;
                 }
 
-#if !NETSTANDARD1_6
+#if NETSTANDARD1_6
+                if (arg.GetType().FullName == "System.DBNull")
```
ChrisMaddock: Doesn't this version work for all platforms? I'd choose that (slightly less efficient?) over the additional conditionals, but that's preference. 😄

src/NUnitFramework/nunitlite.tests/NUnit2XmlOutputWriterTests.cs
```diff
@@ -47,14 +46,12 @@ public void RunMockAssemblyTests()
             Assert.NotNull(result);
 
             StringBuilder sb = new StringBuilder();
-            StringWriter writer = new StringWriter(sb);
-            new NUnit2XmlOutputWriter().WriteResultFile(result, writer, null, null);
-            writer.Close();
```
ChrisMaddock: Good find! :-)

ChrisMaddock:
I started reviewing everything, but realised my package comments were probably a bit more fundamental, so thought we should resolve that first before looking through the rest. 😞

nuget/framework/nunit.nuspec
ChrisMaddock: Point number 2: this is making me think more and more we should have Android and UWP CI. 😞

jnm2: I'll check into that. I'm not liking the trend, but that part is simple enough to undo.

jnm2:
@ChrisMaddock About the `_._` files– they are used both when a package is lying about which platforms it supports and also used when no DLL is needed because the types are already in the target framework. For example, if UWP already has System.Diagnostics.Process as part of the platform without referencing any packages, then you'd expect the System.Diagnostics.Process package's \lib\netstandard1.6 folder to contain `_._`.

Still need to verify.

jnm2:
Edited. Gotta admit, this is quite confusing. I'm going to chat at https://gitter.im/dotnet/corefx.

ChrisMaddock:
Oh yes - I think I encountered that with one of the Android dependencies before! It's a pain there isn't a better system. 😞 

I also really struggled to find any doc on this sorta thing before - I had to piece things together mainly through GitHub comments.

jnm2:
I also found out today that UWP picks up `netcore45` DLLs.

jnm2:
@nunit/framework-and-engine-team I brought the question to corefx chat and Karel was kind enough to answer my confused questions: https://gitter.im/dotnet/corefx?at=5a21990471ad3f873654320c

The overall answer is that newer tooling might have fixed some things and that we should do our best to assume that 99% of netstandard1.6 platform extension packages are benevolent until proven guilty because there's no reasonable way to validate all combinations. Once we or one of our users hits a real-world issue, we can back out of the dependency and file and issue at dotnet/corefx.

### Proposal

Let's add a UWP integration test to this repo as a separate build.cake step which validates the actual NUnit nupkg. We will need to wrangle CI, no doubt. The test should be skipped for NUnit contributors who haven't installed the (hefty) Visual Studio workload.

Once that is done, we should feel better about adding .NET Standard \*.\* libraries going forward. If an NUnit user still manages to hit a bug, we just replicate it in a similar integration test. I'm tempted to ask that we even give System.Threading.Thread another shot too.

What do you think?

src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
jnm2: Sold. Efficiency < maintainability unless profiling proves otherwise.

ChrisMaddock:
As a simpler version, I was wondering about adding a integration step which just pulls nunit.xamarin and try’s to nugget upgrade to the new package version. More smoke test-y perhaps, but would cover more. In the long run, id love to be able to run the fullframework tests on device simulators, at least in CI. But that’s a long way down the priority list!

If we’re doing it, I’d like to have it for all three of the platforms we’re supporting - iOS, Android and UWP. (Unless we decide we want to change swhat we support here!)

CharliePoole:
Going somewhat meta here...

In this and other issues we are dealing with platform questions that we don't entirely understand and ate even sometimes in flux as we try to work with them.

Would we be better off using a dev branch for such problems, reserving master for the stuff we have already mastered?

jnm2:
@ChrisMaddock Seems like a good place to start. That'll couple our tests to the other repo which may not be a bad thing because it'd be less comfortable to force upgrades.

@CharliePoole I don't really have an opinion unless I have an idea of how it would be used. Is the benefit that it would allow a PR to go though but leave the branch in a state that we definitely wouldn't ship without a followup PR?

CharliePoole:
@jnm2 We could define branches in various ways. My main thought here was based on platforms. If we are unsure about either a new platform or even an existing one with changes at the source, then just make a branch that includes it and keep that uncertainty out of master. I'm not talking about a single development branch like GitFlow just about keeping master stable and eliminating the need for risky global changes to master, which is always supposed to be releasable.

jnm2:
@CharliePoole If we diverge, we'll have to do a lot of work to converge again. Let's say I did five PRs and merged them to an intermediate branch before making sure there was no risk. At some point, we're going to have to get rid of that risk before merging the risky branch to master. Does that process really buy us anything, compared to making sure each PR is not risky before merging it?

CharliePoole:
IME, yes. This is something I've seen where a team gradually starts doing bigger and bigger changes. At a point it becomes more effective to suffer the inefficiency of a branch in order to get back to the ideal of making small incremental changes.

I see no reason why it should be harder to merge a given set of changes made incrementally to master than it is now to merge those changes done in one big bang. It's the big bang that brings risk, so I'd favor anything that discourages it.

jnm2:
@CharliePoole I see, so what you are trying to avoid is putting too many changes in a single PR (such as these integration tests)? I did not specify, but I was planning to freeze this PR and start a second PR to add integration tests on master before applying this PR to that and seeing what breaks. Is that not a good compromise, rather than merging to a risk branch and validating later, and possibly having to undo work in this PR?

CharliePoole:
@jnm2 Bear in mind I'm out of the loop and have been for a few weeks. I only have my phone so I'm functioning like an outside observer and I'm not reviewing code because it's too painful on the phone. My only excuse is that observing and suggesting is pretty much what people paid me to do with pretty good success in the last years of my career.

So what I'm observing is...
 * big changes, some global
 * comments among the team questioning how some underlying platforms actually work.
 * ripple effects where changes get added to changes because it's the only way to make it work.

Where it was my job to help projects, I would generally not suggest adding a level of complexity like a new branch to fix it. Rather, I'd just get them to stop doing changes that couldn't be finished and merged in a few hours.

In this case, that doesn't seem like advice that would fly, so my fallback is a separate branch where you all might feel able to work in small steps.

jnm2:
And thanks! I'm totally not against a separate branch. I'm trying to understand how to use it first because otherwise I will make a mess and miss the point. So what's an example of what I might do right now?

 1. Create a working branch called `jnm/risk` for lack of imagination until I know what I'm doing
 2. Base a PR on it adding UWP and Xamarin integration tests
 3. Once merged, put this PR on top of that
 4. Once merged, put https://github.com/nunit/nunit/pull/2588 on top of that
 5. Once merged, put a netstandard2.0 PR on top of that
 6. PR merging all four sets of changes back to master and deal with any merge conflicts from the 4 PRs all at once
 7. Delete `jnm2/risk`

Versus just doing 2–5 directly on master in between other people's PRs. Does that look like what you're talking about?

jnm2:
Or in that picture, would it be better for me to group the six commits in this PR into five PRs, which I almost did?

CharliePoole:
@jnm2 Quick answer waiting for boarding call, after which I'll be traveling the rest of the day.

If you alone are working on all this stuff, then accumulating changes within commiting to master would seem like an anti-pattern. I was observing situations where you and others were collaborating.

I think your plan of separate PRs is best and if they can be smaller that's better. Larger PRs tend to get less review and I've had managers who called that efficient. Usually it's just less attention in the review process.

This is getting out of scope for your issue. I'll start a team discussion when I get home.

jnm2:
I played around with integration tests but they will cause us to postpone some important updates we need to make, so I moved the changes that need integration tests to their own branches for now:
https://github.com/nunit/nunit/compare/jnm2/netstandard16_process
https://github.com/nunit/nunit/compare/jnm2/netstandard16_xml

I also split out the first commits as their own PRs because they really should stand alone:
https://github.com/nunit/nunit/pull/2601
https://github.com/nunit/nunit/pull/2602
https://github.com/nunit/nunit/pull/2603

@nunit/framework-team What's left in this PR is lighter cleanup which would still be nice to get through.

src/NUnitFramework/framework/Internal/ExceptionHelper.cs
```diff
@@ -34,7 +34,7 @@ namespace NUnit.Framework.Internal
     /// </summary>
     public class ExceptionHelper
     {
-#if !NET45 && !NETSTANDARD1_6
+#if NET20 || NET35 || NET40
```
rprouse: Probably better to be explicit as to what they support rather than what they don't. Good idea 👍

src/NUnitFramework/tests/Attributes/PairwiseTests.cs
```diff
@@ -68,7 +68,7 @@ public void OneTimeTearDown()
         // http://www.freequality.org/sites/www_freequality_org/documents/tools/Tagarray_files/tamatrix.htm
         static internal object[] cases = new object[]
         {
-#if ORIGINAL
+            /* Original:
```
rprouse: These old TestCases have been unused in the code for 8 years. Just remove them.

src/NUnitFramework/framework/Internal/ExceptionHelper.cs
jnm2: Yep. It's nice to call out "which frameworks are making us do this" :D

jnm2:
@rprouse Done. You'll have to reapprove because I added a new commit that overlaps a comment of your review.

rprouse:
Approved, waiting on CI

PR #2588, opened 2017-11-30
Use IsInstanceOfType(obj) instead of IsAssignableFrom(obj.GetType())

jnm2:
## Motivation

They are semantically identical (http://referencesource.microsoft.com/#mscorlib/system/type.cs,fe506f282c6fb9c5) except for the null check (for which this PR does include some bug fixes):

```cs
public virtual bool IsInstanceOfType(Object o) 
{
    if (o == null)
        return false;

    // No need for transparent proxy casting check here
    // because it never returns true for a non-rutnime type.
    
    return IsAssignableFrom(o.GetType());
}
```

However, there is a real benefit to using IsInstanceOfType instead. It results in terser code.

```diff
-.Any(a => typeof(T).IsAssignableFrom(a.GetType())
+.Any(typeof(T).IsInstanceOfType)
```

And the runtime is optimized to run same logic much faster because there is no need to allocate an expensive `Type` object and compare it. Rather, it compares the type handle straight off the object reference (http://referencesource.microsoft.com/#mscorlib/system/rttype.cs,11e2cc8ec5607523) and the JIT optimizes it:

```cs
public override bool IsInstanceOfType(Object o)
{
    return RuntimeTypeHandle.IsInstanceOfType(this, o);
}
```

## Outcomes

I discovered that the type checks in our generic equality and comparer adapters threw NRE rather than following proper runtime semantics and delegating the nulls to the adaptee.

In fixing this, I unified the identical logic with ConstraintUtils.RequireActual. The resulting API makes use of out parameters which is okay, following a dignified tradition in .NET. But once we move to C# 7, the try/out pattern will seem much more appealing too.

### DIFF
--- a/src/NUnitFramework/framework/Internal/TypeHelper.cs
+++ b/src/NUnitFramework/framework/Internal/TypeHelper.cs
@@ -250,25 +250,22 @@ namespace NUnit.Framework.Internal
             {
                 object arg = arglist[i];
 
-                if (arg != null && arg is IConvertible)
+                if (arg is IConvertible)
                 {
                     Type argType = arg.GetType();
                     Type targetType = parameters[i].ParameterType;
                     bool convert = false;
 
-                    if (argType != targetType && !argType.IsAssignableFrom(targetType))
+                    if (argType != targetType && IsNumeric(argType) && IsNumeric(targetType))
                     {
-                        if (IsNumeric(argType) && IsNumeric(targetType))
-                        {
-                            if (targetType == typeof(double) || targetType == typeof(float))
-                                convert = arg is int || arg is long || arg is short || arg is byte || arg is sbyte;
+                        if (targetType == typeof(double) || targetType == typeof(float))
+                            convert = arg is int || arg is long || arg is short || arg is byte || arg is sbyte;
+                        else
+                            if (targetType == typeof(long))
+                                convert = arg is int || arg is short || arg is byte || arg is sbyte;
                             else
-                                if (targetType == typeof(long))
-                                    convert = arg is int || arg is short || arg is byte || arg is sbyte;
-                                else
-                                    if (targetType == typeof(short))
-                                        convert = arg is byte || arg is sbyte;
-                        }
+                                if (targetType == typeof(short))
+                                    convert = arg is byte || arg is sbyte;
                     }
 
                     if (convert)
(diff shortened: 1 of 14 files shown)


nuget/framework/nunit.nuspec
```diff
@@ -30,6 +30,8 @@ Supported platforms:
       <group targetFramework="net45" />
       <group targetFramework="netstandard1.6">
         <dependency id="NETStandard.Library" version="1.6.0" />
+        <dependency id="System.Diagnostics.Process" version="4.3.0" />
+        <dependency id="System.Xml.XmlDocument" version="4.3.0" />
```
ChrisMaddock: Early comment: Both of these packages look like they're not _actually_ supported on iOS/Android. 😞 I haven't tried it, but unzipping the packages, they have those `_._` files...

jnm2: Conversation continued at https://github.com/nunit/nunit/pull/2586#issuecomment-348131519

jnm2:
This PR is ready for review.

src/NUnitFramework/framework/Internal/TypeHelper.cs
```diff
@@ -250,25 +250,22 @@ public static void ConvertArgumentList(object[] arglist, IParameterInfo[] parame
...
                 object arg = arglist[i];
 
-                if (arg != null && arg is IConvertible)
+                if (arg is IConvertible)
                 {
                     Type argType = arg.GetType();
                     Type targetType = parameters[i].ParameterType;
                     bool convert = false;
 
-                    if (argType != targetType && !argType.IsAssignableFrom(targetType))
```
ChrisMaddock: Can you clarify why !IsAssignableFrom() can be safely dropped here? Sorry - there's a lot going on in this method!

ChrisMaddock:
Derp, wrong button - that was meant to be the 'neutral review option' rather than declined - sorry!

src/NUnitFramework/framework/Internal/TypeHelper.cs
jnm2: Oh! Sorry. This would have been the benefit of making this change in its own PR to start with– there should have been a whole commit making just this change with a commit message explaining why.

The reason is that the only way `convert` can become true is if both types are considered numeric by `IsNumeric`. There are no such types that are assignable to one another except in the trivial case of `argType == targetType` which is already guarded against, so `argType.IsAssignableFrom(targetType)` is guaranteed to always be false when evaluated after the two `&& IsNumeric` checks.

jnm2: Oh, and I did a quick brute-force LINQPad test to make sure that my instinct was right about this. because that's an important thing to verify. 😄

ChrisMaddock: Great - thanks! 😄 Approving - just waiting on CI

PR #2594, opened 2017-12-02
Added .props with ProjectCapability to suppress test project service GUID item

### LINKED ISSUE #2592: Add .props with ProjectCapability to suppress test project service GUID item
The decision was made that it's the responsibility of the test framework package to tell VS to stop autoinserting this boilerplate in test csprojs:

```xml
  <ItemGroup>
    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
  </ItemGroup>
```

See:
https://github.com/Microsoft/vstest/issues/394#issuecomment-276254646
https://github.com/Microsoft/vstest/pull/440/files#r111104404

Any project that references NUnit and does not reference `Microsoft.NET.Test.Sdk` will benefit from this.

This is a super easy fix. We add the following `\build\NUnit.props` to the nupkg and we're done:

```xml
<Project>

  <ItemGroup>
    <ProjectCapability Include="TestContainer" />
  </ItemGroup>

</Project>
```

From my perspective it's much nicer to do this in the framework package than the adapter package. The only projects where I use the adapter package are where I need to use `dotnet test`, and one day, I'll be using our future .NET Core console runner. (`dotnet nunit` anyone? =D)

Let me know if you agree and I'll add the file for 3.10.

jnm2:
The first commit fixes #2592.
@rprouse I hesitated to edit the CreateImage task to copy to `build\NUnit.props`, so the nuspec uses the path `..\..\nuget\framework\NUnit.props`. I'm not sure what you would want here.
(I'm already planning to propose moving to MSBuild-based Pack, so if that happens everything gets rearranged anyway.)

The second commit is cosmetic. If we wanted, we could also replace the item group with `<Import Project="..\..\..\nuget\framework\NUnit.props" />`. What do you think?

### DIFF
--- a/nuget/framework/NUnit.props
+++ b/nuget/framework/NUnit.props
@@ -1,0 +1,7 @@
+<Project>
+
+  <ItemGroup>
+    <ProjectCapability Include="TestContainer" />
+  </ItemGroup>
+
+</Project>
--- a/nuget/framework/nunit.nuspec
+++ b/nuget/framework/nunit.nuspec
@@ -48,5 +48,6 @@ Supported platforms:
     <file src="bin/net45/nunit.framework.xml" target="lib\net45" />
     <file src="bin/netstandard1.6/nunit.framework.dll" target="lib\netstandard1.6" />
     <file src="bin/netstandard1.6/nunit.framework.xml" target="lib\netstandard1.6" />
+    <file src="..\..\nuget\framework\NUnit.props" target="build" />
   </files>
 </package>
--- a/src/NUnitFramework/mock-assembly/mock-assembly.csproj
+++ b/src/NUnitFramework/mock-assembly/mock-assembly.csproj
@@ -12,7 +12,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
+    <ProjectCapability Include="TestContainer" />
   </ItemGroup>
 
 </Project>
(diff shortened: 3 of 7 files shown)


jnm2:
I did not. Here's the link: https://github.com/Microsoft/vstest/issues/472#issuecomment-348700341

The information we're declaring seems benevolent enough though that I'm not worried about carrying it forward forever. I'm more trying to figure out whether it should be classified as a workaround and tracked.

rprouse:
The .NET 3.5 test run timed out at 60 minutes again. I am going to merge this since it is unrelated.

jnm2:
Got an answer now: https://github.com/Microsoft/vstest/issues/472#issuecomment-348870120

They'd like us to consider adding an `IsTestProject` property too. I'll open an issue.

jnm2:
Actually I'll wait for a response to my [reply](https://github.com/Microsoft/vstest/issues/472#issuecomment-348870120) now, since I read Brad's pushback which is where I was with it too.

CharliePoole:
I see you're waiting but I'll chime in anyway... AFAIC this would be a mistake. An IsTestProjectAttribute would mean "This is a test project for Visual Studio Test Explorer purposes." We could never change the semantics but Microsoft could change it at any time. That's not useful. Any attribute that Test Explorer needs in order to adequately support uses should be defined by Microsoft and used by those users who need it.

We have other reasons for defining what is or isn't a test project for our own purposes. We, of course, are specifically talking about an __NUnit__ test project. We need to be able to create attributes and interfaces for those purposes and control what they mean.

Even __If__ VS Test Explorer should become the main way (numerically) that NUnit gets used in the future, I think we have to be careful to distinguish what is NUnit's and what is theirs. And of course, personally speaking, I hope that the Explorer never becomes the main way that NUnit is used. 😄

PR #2595, opened 2017-12-04
Adding minClientVersion to both nuspecs.

### LINKED ISSUE #2564: Add minClientVersion to .nuspec files
We recently discovered (https://github.com/nunit/nunit/issues/2369) that our NuGet packages require a NuGet client version of at least 2.12. We should add the `minClientVersion` element to our nuspecs, so users get a better experience for this.

Whoever looks into this should check that 2.12 is actually the correct version, and we're not using any features only added in later versions of the NuGet client.

phillipsj:
This fixes #2564 by adding the tag to the nuspec files for both nunit and nunitlite.

### DIFF
--- a/nuget/framework/nunit.nuspec
+++ b/nuget/framework/nunit.nuspec
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
-  <metadata>
+  <metadata minClientVersion="2.12">
     <id>NUnit</id>
     <title>NUnit</title>
     <version>$version$</version>
--- a/nuget/nunitlite/nunitlite.nuspec
+++ b/nuget/nunitlite/nunitlite.nuspec
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
-  <metadata>
+  <metadata minClientVersion="2.12">
     <id>NUnitLite</id>
     <title>NUnitLite</title>
     <version>$version$</version>


ChrisMaddock:
Hey @phillipsj - thanks for the PR.

You'll see the packaging has broken on the Appveyor CI. 😞 [Digging around a bit](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Packaging/compiler/resources/nuspec.xsd) it looks like the nuspec docs are wrong, and `minClientVersion` should actually be an attribute on the `metadata` element, rather than an element of it's own? 😕 I haven't tried it - maybe we could submit a correction PR to the NuGet docs as well, if this is the case.

Would you mind trying it out? If you want to test this locally, you can do `build -t Build` then `build -t Package` with the Cake script.

jnm2:
Huh, I've only every seen it as an attribute and yet the docs put it in a table under the column header "Single elements": https://docs.microsoft.com/en-us/nuget/schema/nuspec#single-elements

phillipsj:
From all my research and other examples I have found on the web, it is in
the correct location.

phillipsj:
I will dig into it again today and make the change.

ChrisMaddock:
Great, thanks @phillipsj. 👍 

I've never used it before - just spotted it in the docs while doing something else and thought it was exactly what we needed! The docs definitely read like it should be an element. :-|

jnm2:
Out of 957 random .nuspecs in my `%userprofile%/.nuget/packages` cache, there are no occurrences of `<minClientVersion` and there are 473 occurrences of `minClientVersion=`.

mikkelbu:
Also the release notes states the following https://docs.microsoft.com/en-us/nuget/release-notes/nuget-2.5##Add_a_%27Minimum_NuGet_Version%27_property_to_packages

> A new metadata attribute called 'minClientVersion' can now indicate the minimum NuGet client version required to consume a package.+

jnm2:
@phillipsj Thanks so much for pushing this through! 🎉

PR #2597, opened 2017-12-06
Use array argument contents in name of parameterized tests

### LINKED ISSUE #1382: Use array argument contents in name of parameterized tests rather than just array type.
TestCase and TestCaseSource make it too easy to create tests with duplicate names.

Case 1: Arrays

Arrays are not formatted further, only their type name becomes part of the test name by default. For example, given the following test fixture:

```
using NUnit.Framework;

namespace NUnitDuplicateTests
{
    public class DuplicateTestFixture
    {
        [TestCase((object)new[] { "John", "Smith" })]
        [TestCase((object)new[] { "Emperor", "Palpatine" })]
        [TestCase((object)new[] { "Groot" })]
        public void EverybodyHasAtLeastTwoNames(string[] nameParts)
        {
            Assert.That(nameParts, Has.Length.GreaterThanOrEqualTo(2));
        }
    }
}
```

Only one test is discovered: `NUnitDuplicateTests.DuplicateTestFixture.EverybodyHasAtLeastTwoNames(System.String[])`.

Only one test shows up in the Visual Studio test explorer.

Only one unique test shows up with `nunit3-console --explore`, but is printed 3 times.

This creates issues for applications that use the Explore API, if they wish to track each test individually for various purposes.

I would expect the test name to be something along the lines of `NUnitDuplicateTests.DuplicateTestFixture.EverybodyHasAtLeastTwoNames(["Emperor", "Palpatine"])`.

Case 2: Test Names

I would logically expect a test name from TestCaseAttribute or TestCaseData to be local to the method that declares the test. It is often convenient to use the same test data for multiple test methods, e.g.

```
using System.Collections;
using NUnit.Framework;

namespace NUnitDuplicateTests
{
    public class DuplicateTestFixture
    {
        [TestCaseSource(nameof(Names))]
        public void EverybodyHasAtLeastTwoNames(string[] nameParts)
        {
            Assert.That(nameParts, Has.Length.GreaterThanOrEqualTo(2));
        }

        [TestCaseSource(nameof(Names))]
        public void EverybodyIsOnTheImperialPayroll(string[] nameParts)
        {
            // TODO: Talk to the Payroll division on Coruscant.
        }

        public static IEnumerable Names
        {
            get
            {
                yield return new TestCaseData(new[] { "Emperor", "Palpatine" }).SetName("Our grand Emperor");
                yield return new TestCaseData(new[] { "Darth", "Vader" }).SetName("Luke's Dad");
            }
        }
    }
}
```

Although this fixture declares four tests, there are only two unique test names - `NUnitDuplicateTests.DuplicateTestFixture.Our grand Emperor` and `NUnitDuplicateTests.DuplicateTestFixture.Luke's Dad`.

I would expect there to be four, somewhat resembling the following:
1. `NUnitDuplicateTests.DuplicateTestFixture.EverybodyHasAtLeastTwoNames."Our grand Emperor"`
2. `NUnitDuplicateTests.DuplicateTestFixture.EverybodyHasAtLeastTwoNames."Luke's Dad"`
3. `NUnitDuplicateTests.DuplicateTestFixture.EverybodyIsOnTheImperialPayroll."Our grand Emperor"`
4. `NUnitDuplicateTests.DuplicateTestFixture.EverybodyIsOnTheImperialPayroll."Luke's Dad"`

Case 3: String truncation

String parameters are truncated to 40 characters, which makes it difficult to use long strings (such as URLs) in test cases. e.g.:

```
using NUnit.Framework;

namespace NUnitDuplicateTests
{
    public class DuplicateTestFixture
    {
        [TestCase("The Imperial Death Star is a collossal waste of money. The Empire could better use those funds on supplies and soldiers.")]
        [TestCase("The Imperial Death Star is a collossal monument to the power of the Empire.")]
        public void AllOpinionsAreValid(string opinion)
        {
            Assert.That(opinion, Is.Not.Null, "Not having an opinion is not in itself an opinion.");
        }
    }
}
```

This fixture only has one unique test name, `NUnitDuplicateTests.DuplicateTestFixture.AllOpinionsAreValid("The Imperial Death Star is a collossa...")`.

From this, I see three resolutions:
1. Can test names be made smarter by introspecting arrays?
2. Can explicit test names be made local to their method, not local to their fixture?
3. Can strings be not truncated, or the truncation limit be increased well beyond 40 characters?

yaakov-h:
... rather than just array type.

Closes #1382.

I hope these tests follow the nunit convention... I've tried to more-or-less mirror what the existing tests were doing.


yaakov-h:
Note that this applies the parameter length limit to each item in the array, rather than the array overall.

If we want it to apply to the array overall, I'm really not sure on the best way to achieve that.

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -353,6 +369,13 @@ public void Issue1337SingleMemberArrayAsArgument(string[] args)
             new int[] { 12, 6, 2 }
         };
 
+        static object[] ComplexArrayBasedTestInput = new[]
```
mikkelbu: I don't think that this is used anywhere.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
```diff
@@ -231,7 +231,30 @@ protected static string GetDisplayString(object arg, int stringMax)
...
+                    {
+                        StringBuilder builder = new StringBuilder();
+                        builder.Append("[ ");
+
+                        for (int i = 0; i < argArray.Length; i++)
+                        {
+                            if (i > 0)
+                                builder.Append(", ");
+
+                            string valueDisplayString = GetDisplayString(argArray.GetValue(i), stringMax);
```
mikkelbu: As you stated in the discussion. Right now we only use `stringMax` to ensure that strings are not too long. I think we should also apply this to restrict the the length of the result of `builder.ToString` (I've actually not an idea of the default value of `stringMax`, so it could be that it is not sufficient?).

yaakov-h: I think the default is 40.

yaakov-h:
Can we use `nameof`? I noticed the existing code wasn't using it, and it also still targets .NET 2.0... is that using Roslyn, or the old SDK/Compiler?

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
yaakov-h: Fixed.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
yaakov-h: It looks like the default is actually 0, so it would only truncate if the consumer specifies a limiter when overriding the test name.

CharliePoole: Actual default is defined by NUnit's default naming template, and is currently 40. Since it's documented to apply only to strings, we either need another setting or should change the docs.

CharliePoole: Actual default is defined by NUnit's default naming template, and is currently 40. Since it's documented to apply only to strings, we either need another setting or should change the docs.

CharliePoole:
What if my argument is a multidimensional array?

yaakov-h:
Hmm, I keep forgetting that they exist.

I’d think we should ignore them, or perhaps just list the dimensions (`System.String[2, 3]`). I don’t see any sensible way to generate a simple display name for a multidimensional array.

CharliePoole:
Let's first agree on the representation. See #1382

CharliePoole:
If you just check the Rank of the array as equal to 1 and ignore others I think we're OK. Otherwise, I think you may blow up on a multidimensional array.

ChrisMaddock:
I haven't had a chance to look through this in detail - but I love the idea. 😄 

One thought - what if my array has 1000 items? I've suddenly got a mighty long test name! Can we limit this to arrays of under 5 elements, or something?

CharliePoole:
I like the idea of a limit. I'll update the little spec in #1382.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
jnm2: @CharliePoole When I was in there last I thought I noticed that the framework itself does not have a limit by default. Found the PR that removed the limit: https://github.com/nunit/nunit/pull/1556

CharliePoole: There is no default maximum length any longer. Instead, there is a default naming template, which happens to specify a maximum length of 40 for string arrays. From the user point of view, that's what counts. __Internally__ there is no limit on length.

jnm2: Ah, so the limit is defined by the adapter but not the framework itself.
https://github.com/nunit/nunit3-vs-adapter/pull/189
If any limit comes into play when using the framework with the console or NUnitLite, I'm overlooking it.

CharliePoole: I think we are getting messed up by two meanings of Default in this context.

1. The default for the framework is "{m}{a}" with no limits. Think of this as the "default default".
2. The default for each runner depends on the runner. If it isn't set, we get the framework default.

The delegation of the actual (effective) default to the runner was decided on about two years ago in the discussions around implementing the template naming feature. The logic was that the framework doesn't care about the default length whereas some runners do.

At one point, I modified the NUnit 3 adapter to use a default length of 40. I think this may now have been overridden and we are using no truncation if the user doesn't specify a template in the `.runsettings` file. This is from a cursory examination, since I don't work on the adapter any longer.

I believe the console now only overrides the "default default" if the user specifies a template on the command-line.

As far as I can see, nobody ever modified the nunitlite runner.

jnm2: Okay, that makes more sense. That means that @yaakov-h is right and the default (defined by the default naming template) is currently 0?

CharliePoole: Yes, but you have to allow for what the runner sets as the effective default if you are talking to users. My comments about what each runner does right now shouldn't be viewed as definitive, since lots of people have made changes and I'm not following most of them.

jnm2:
I'm not liking the extra padding spaces `[ x, y ]` as much as `[x, y]`. Did I miss discussion on this?

yaakov-h:
No, that’s just how it was in my custom attributes. I’d be fine to remove the padding.

jnm2:
@nunit/framework-team Which way do you prefer?

CharliePoole:
I like it better without the extra padding.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
```diff
@@ -220,7 +220,47 @@ protected static string GetDisplayString(object arg, int stringMax)
...
                     : Convert.ToString(arg, System.Globalization.CultureInfo.InvariantCulture);
 
-                if (arg is double)
+                var argArray = arg as Array;
+                if (argArray != null)
+                {
+                    if (argArray.Length == 0)
+                        display = "[]";
+                    else
+                    {
```
jnm2: Lol, braces on the `else` but not the `if` always made me feel unbalanced. But I don't see anything wrong with it.

jnm2:
@mikkelbu When you get a chance, can you update your review?

jnm2:
Thanks Mikkel!
And thank you for the great work, @yaakov-h. I'm stoked that this is in!

Brar:
@CharliePoole wrote:

> What if my argument is a multidimensional array?

I just stumbled over this while implementing some advanced testing for [Npgsql](https://github.com/npgsql/npgsql/).

As I already had implemented it for my tests I created this [Gist](https://gist.github.com/Brar/260a670863121bb0da616976d45681c4) that shows a possible implementation of multidimensional array formatting for NUnit.

If you think that something like this would be accepted into NUnit I am willing to work on a pull request.

Issue #2598, opened 2017-12-06
Is.Not.EquivalentTo is extremely slow


Jcomper:
Just updated my tests from NUnit 2.4.8 to 3.9.0 and noticed that tests have started running too long.
And here is a synthetic test with allows me to reproduce the problem:

> [Test]
> public void TestEquivalent()
> {
> 	byte[] data = new byte[0x80000 + 7];
> 	byte[] encrypted = new byte[0x80000 + 7];
> 	encrypted[0] = 2;
> 	encrypted[1] = 3;
> 
> 	var watch = Stopwatch.StartNew();
> 	Assert.That(encrypted, Is.Not.EquivalentTo(data));
> 	watch.Stop();
> 	Debug.WriteLine($"Takes: {watch.ElapsedMilliseconds} milliseconds");
> }
> 

It used to work fast (with version 2.4.8), but now it takes 123116 milliseconds to run on my computer. In real test when two arrays contain completely different values, it takes about 22 minutes to run. Any idea why is it so?

mikkelbu:
I think it is because of #2337.

It is easy to discover that the arrays are different, but due to the tallying we need to "calculate" the differences, hence for each `0` in one of the arrays we need to remove a matching `0` from the other array, and since we do this from the beginning of the arrays a lot of reshuffling will have to be done (each deletion will have to move all the following elements). Using a better datastructure we could probably improve the performance quite a bit. Or we could run this algorithm from the end of the arrays.

Ps. Do you really mean `Is.Not.EquivalentTo` (unordered comparison) when you use words like `encrypted`?

CharliePoole:
@mikkelbu 's last question is key. Ate you sure you want equivalence and not equality?

I implemented this s long time ago using a simple, unoptimized approach. I didn't figure it would ever be used for large arrays and that has seemed to be true so far. Once in a while someone complains and then discovers that our definitions of equality and equivalence are not what they thought!

Jcomper:
@mikkelbu @CharliePoole Thanks! I'm not the owner of that code, so I can only guess.
 
 Original test generate array of bytes, encrypt it and then use in the following assert:

` Assert.That(encrypted, new NotConstraint(new CollectionEquivalentConstraint(data)), "Encrypted == Data for {0}", encryptor.GetType().Name);
` 

So I guess there should have been used equality definition and not equivalent as @CharliePoole pointed out.
 
[...]

PR #2600, opened 2017-12-06
Reverse iteration in TryRemove

mikkelbu:
Improves the performance of large collections with
many equivalent elements by minimising the shifting
of elements within _missingItems.

Relates to #2598


CharliePoole:
As far as I can see, it's not possible to make the comparisons done in `NUnitEqualityComparer` hashable so long as we are supporting equality within a tolerance. Conceivably, we could split the class into a simple comparer and a derived class that supports tolerance. Then the simple class could be hashable.

jnm2:
Thanks Mikkel, this kind of change is great!

PR #2602, opened 2017-12-06
Same DBNull behavior for netstandard1.6

jnm2:
This makes our various DLLs behave more consistently and reduces platform-conditional preprocessor.
By contract, it's possible for our netstandard1.6 DLL to run on several platforms.

(Split out of #2586 for quicker merge and because it can stand on its own.)


jnm2:
@mikkelbu Quite! Thanks for spotting.

Issue #2542, opened 2017-11-08
NUnit does not support parallelism on .NET Core 2.0


PYasonau:
It is known, but very critical issue for those, who use nunit as testframework for .net core 2.0 targer. In some cases it is not possible to use target framework different from .net core.  Parallelism is critical feature and test society need to know should they build they test system for .net core based on nunit or migrate to other test frameworks

rprouse:
We hope to add a .NET Standard 2.0 version of the framework soon that will fully support Parallelism and re-enable many of the NUnit features that were disabled because of missing APIs in earlier versions of .NET Standard.

SergiusZurawski:
Hello,  I have noticed that you have scheduled your next release 3.10 on February. Is there any chance that this issue will be solved by 3.10 release?

jnm2:
@rprouse Once issue-2331 is merged I would be happy to get this done. What so you think?

rprouse:
@jnm2 I agree, I would really love to go through what is missing from the .NET Standard 1.6 version, create issues and re-enable whatever we can.

jnm2:
@rprouse Is this something you'd like to do or shall I take it? My plan would probably be to start from net45 rather than netstandard1.6 since I expect netstandard2.0 to have all the APIs we need.

rprouse:
@jnm2 I am fine with you taking it. Are you planning on adding a .NET Standard 2.0 target and basing that on the .NET 4.5 target?

jnm2:
@rprouse That's my plan, after finishing my work underway at https://github.com/nunit/nunit/tree/jnm2/reduce_platform_defines which will resolve my last two items at https://github.com/orgs/nunit/teams/framework-and-engine-team/discussions/3.

jnm2:
Just waiting for https://github.com/nunit/nunit/pull/2586 to be merged, then I will start on this right away.

potocnika:
Eagerly awaiting this. We switched from xunit to nunit some time ago because there were issues with xunit one time setup equivalent (fixtures) and parallelism and we never got parallelism to work on .net core 1.0 or 2.0. Really hoping this feature is available in nunit soon.

jnm2:
[...]

PR #2607, opened 2017-12-08
Added .NET Standard 2.0 build! 🎉

jnm2:
Fixes https://github.com/nunit/nunit/issues/2542.

In the first commit I got the thing building and in each other commit I fixed tests.

I introduced two new feature defines: PLATFORM for platform detection and THREAD_ABORT for things needing Thread.Abort like timeout and cancellation since CoreCLR throws PlatformNotSupportedException for Thread.Abort.

Conditional preprocessor is the simple workaround. A more sophisticated approach would be to leave the Thread.Abort call there and to do platform capability detection at runtime, but that is complexity we can implement at any point in the future if desired.

### DIFF
--- a/nuget/framework/nunit.nuspec
+++ b/nuget/framework/nunit.nuspec
@@ -17,7 +17,7 @@ This package includes the NUnit 3 framework assembly, which is referenced by you
 
 Supported platforms:
 - .NET Framework 2.0+
-- .NET Standard 1.6
+- .NET Standard 1.6+
 - .NET Core</description>
     <releaseNotes>This package includes the NUnit 3 framework assembly, which is referenced by your tests. You will need to install version 3 of the nunit3-console program or a third-party runner that supports NUnit 3 in order to execute tests. Runners intended for use with NUnit 2.x will not run NUnit 3 tests correctly.</releaseNotes>
     <language>en-US</language>
@@ -29,7 +29,10 @@ Supported platforms:
       <group targetFramework="net40" />
       <group targetFramework="net45" />
       <group targetFramework="netstandard1.6">
-        <dependency id="NETStandard.Library" version="1.6.0" />
+        <dependency id="NETStandard.Library" version="1.6.1" />
+      </group>
+      <group targetFramework="netstandard2.0">
+        <dependency id="NETStandard.Library" version="2.0.0" />
       </group>
     </dependencies>
   </metadata>
@@ -48,6 +51,8 @@ Supported platforms:
     <file src="bin/net45/nunit.framework.xml" target="lib\net45" />
     <file src="bin/netstandard1.6/nunit.framework.dll" target="lib\netstandard1.6" />
     <file src="bin/netstandard1.6/nunit.framework.xml" target="lib\netstandard1.6" />
+    <file src="bin/netstandard2.0/nunit.framework.dll" target="lib\netstandard2.0" />
+    <file src="bin/netstandard2.0/nunit.framework.xml" target="lib\netstandard2.0" />
     <file src="..\..\nuget\framework\NUnit.props" target="build" />
   </files>
 </package>
--- a/src/CommonAssemblyInfo.cs
+++ b/src/CommonAssemblyInfo.cs
@@ -33,30 +33,42 @@ using System.Reflection;
 
 #if DEBUG
 #if NET45
-[assembly: AssemblyConfiguration(".NET 4.5 Debug")]
+[assembly: AssemblyConfiguration(".NET Framework 4.5 Debug")]
 #elif NET40
-[assembly: AssemblyConfiguration(".NET 4.0 Debug")]
+[assembly: AssemblyConfiguration(".NET Framework 4.0 Debug")]
 #elif NET35
-[assembly: AssemblyConfiguration(".NET 3.5 Debug")]
+[assembly: AssemblyConfiguration(".NET Framework 3.5 Debug")]
 #elif NET20
-[assembly: AssemblyConfiguration(".NET 2.0 Debug")]
+[assembly: AssemblyConfiguration(".NET Framework 2.0 Debug")]
+#elif NETSTANDARD1_6
+[assembly: AssemblyConfiguration(".NET Standard 1.6 Debug")]
+#elif NETSTANDARD2_0
+[assembly: AssemblyConfiguration(".NET Standard 2.0 Debug")]
 #elif NETCOREAPP1_1
 [assembly: AssemblyConfiguration(".NET Core 1.1 Debug")]
+#elif NETCOREAPP2_0
+[assembly: AssemblyConfiguration(".NET Core 2.0 Debug")]
 #else
-[assembly: AssemblyConfiguration("Debug")]
+#error Missing AssemblyConfiguration attribute for this target.
 #endif
 #else
 #if NET45
-[assembly: AssemblyConfiguration(".NET 4.5")]
+[assembly: AssemblyConfiguration(".NET Framework 4.5")]
 #elif NET40
-[assembly: AssemblyConfiguration(".NET 4.0")]
+[assembly: AssemblyConfiguration(".NET Framework 4.0")]
 #elif NET35
-[assembly: AssemblyConfiguration(".NET 3.5")]
+[assembly: AssemblyConfiguration(".NET Framework 3.5")]
 #elif NET20
-[assembly: AssemblyConfiguration(".NET 2.0")]
+[assembly: AssemblyConfiguration(".NET Framework 2.0")]
+#elif NETSTANDARD1_6
+[assembly: AssemblyConfiguration(".NET Standard 1.6")]
+#elif NETSTANDARD2_0
+[assembly: AssemblyConfiguration(".NET Standard 2.0")]
 #elif NETCOREAPP1_1
 [assembly: AssemblyConfiguration(".NET Core 1.1")]
+#elif NETCOREAPP2_0
+[assembly: AssemblyConfiguration(".NET Core 2.0")]
 #else
-[assembly: AssemblyConfiguration("")]
+#error Missing AssemblyConfiguration attribute for this target.
 #endif
 #endif
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -29,9 +29,13 @@ using NUnit.Framework.Internal;
 using NUnit.Framework.Internal.Execution;
 using System.Collections.Generic;
 using System.IO;
+
 #if !NETSTANDARD1_6
 using System.Diagnostics;
 using System.Security;
+#endif
+
+#if NET20 || NET35 || NET40 || NET45
 using System.Windows.Forms;
 #endif
 
@@ -56,7 +60,7 @@ namespace NUnit.Framework.Api
         private EventPump _pump;
 #endif
 
-        #region Constructors
+#region Constructors
 
         /// <summary>
         /// Initializes a new instance of the <see cref="NUnitTestAssemblyRunner"/> class.
@@ -67,9 +71,9 @@ namespace NUnit.Framework.Api
             _builder = builder;
         }
 
-        #endregion
+#endregion
 
-        #region Properties
+#region Properties
 
 #if PARALLEL
         /// <summary>
@@ -133,9 +137,9 @@ namespace NUnit.Framework.Api
         /// </summary>
         private TestExecutionContext Context { get; set; }
 
-        #endregion
+#endregion
 
-        #region Public Methods
+#region Public Methods
 
         /// <summary>
         /// Loads the tests found in an Assembly
@@ -265,9 +269,9 @@ namespace NUnit.Framework.Api
             }
         }
 
-        #endregion
+#endregion
 
-        #region Helper Methods
+#region Helper Methods
 
         /// <summary>
         /// Initiate the test run.
@@ -316,7 +320,7 @@ namespace NUnit.Framework.Api
                 }
             }
 
-#if !NETSTANDARD1_6
+#if NET20 || NET35 || NET40 || NET45
             if (Settings.ContainsKey(FrameworkPackageSettings.PauseBeforeRun) &&
                 (bool)Settings[FrameworkPackageSettings.PauseBeforeRun])
                 PauseBeforeRun();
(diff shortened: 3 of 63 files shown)


src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -29,9 +29,13 @@
 using NUnit.Framework.Internal.Execution;
 using System.Collections.Generic;
 using System.IO;
+
 #if !NETSTANDARD1_6
 using System.Diagnostics;
 using System.Security;
+#endif
+
+#if NET20 || NET35 || NET40 || NET45
```
rprouse: I probably wrote that WinForms code, but really? 😄

jnm2: We could make this cross-platform... :D

jnm2:
Building.md sounds good. Should I do that in this PR?

rprouse:
Yes, please do, then we can merge. Ping me when it is done and I will bypass CI since it is just a text change.

jnm2:
Hang on, I forgot to update AssemblyInfos too. 😳

jnm2:
- Added to BUILDING.md (is it too long?)
 - Fixed  assembly info

nuget/framework/nunit.nuspec
```diff
@@ -29,7 +29,10 @@ Supported platforms:
       <group targetFramework="net40" />
       <group targetFramework="net45" />
       <group targetFramework="netstandard1.6">
-        <dependency id="NETStandard.Library" version="1.6.0" />
+        <dependency id="NETStandard.Library" version="1.6.1" />
```
mikkelbu: What is effect of going from version 1.6.0 to 1.6.1 ?

src/CommonAssemblyInfo.cs
```diff
@@ -33,30 +33,42 @@
...
-[assembly: AssemblyConfiguration(".NET 2.0 Debug")]
+[assembly: AssemblyConfiguration(".NET Framework 2.0 Debug")]
+#elif NETSTANDARD1_6
+[assembly: AssemblyConfiguration(".NET Standard 1.6 Debug")]
+#elif NETSTANDARD2_0
+[assembly: AssemblyConfiguration(".NET Standard 2.0 Debug")]
 #elif NETCOREAPP1_1
 [assembly: AssemblyConfiguration(".NET Core 1.1 Debug")]
+#elif NETCOREAPP2_0
+#error Missing AssemblyConfiguration attribute for this target.
```
mikkelbu: Is this on purpose?

jnm2: No; it's me after not being able to sleep till 2 am and taking NyQuil. 😷  Thanks.

src/NUnitFramework/framework/Properties/AssemblyInfo.cs
```diff
@@ -51,17 +51,19 @@
                               "723bc8d775a66b594adeb967537729fe2a446b548cd57a6")]
 
 #if NET45
-[assembly: AssemblyTitle("NUnit Framework .NET 4.5")]
+[assembly: AssemblyTitle("NUnit Framework (.NET Framework 4.5)")]
 #elif NET40
-[assembly: AssemblyTitle("NUnit Framework .NET 4.0")]
+[assembly: AssemblyTitle("NUnit Framework (.NET Framework 4.0)")]
 #elif NET35
-[assembly: AssemblyTitle("NUnit Framework .NET 4.0")]
```
mikkelbu: God spot 👍

nuget/framework/nunit.nuspec
jnm2: I don't know, but I do know that we are currently compiling against 1.6.1 so I made the nuspec match the csproj. Perhaps I should change the csproj.

jnm2: Actually, I doubt that these are supposed to show up in the .nuspec at all– https://andrewlock.net/what-is-the-netstandard-library-metapackage/

src/NUnitFramework/nunitlite.tests/TextUIReportTests.cs
```diff
@@ -120,6 +120,8 @@ public void SummaryReportTest()
                 "    Duration: 0.123 seconds\n\n"));
         }
 
+        // Requires mock-assembly to be compiled with /optimize-
+        // or else the test methods do tail calls and the stack frame disappears
```
mikkelbu: Are you sure about the tail calls? I would expect it to be method inlining and not tail calls (not that it matters much 😄 )

src/NUnitFramework/testdata/TestCaseAttributeFixture.cs
```diff
@@ -96,7 +96,7 @@ public void MethodWithIncludePlatform(int num)
         [TestCase(2, ExcludePlatform = "Linux")]
         [TestCase(3, ExcludePlatform = "MacOSX")]
         [TestCase(4, ExcludePlatform = "XBox")]
-        public void MethodWitExcludePlatform(int num)
+        public void MethodWithExcludePlatform(int num)
```
mikkelbu: Note that this is also corrected in https://github.com/nunit/nunit/pull/2597/files#diff-e41bbd1635751322b84ea837982ac8bdR99

src/NUnitFramework/tests/Assertions/AdhocTestExecutionTests.cs
```diff
@@ -3,7 +3,7 @@
 using System.Reflection;
 using NUnit.Framework.Internal;
 
-#if !NETCOREAPP1_1
+#if NET20 || NET35 || NET40 || NET45
```
mikkelbu: Why not `PLATFORM` as in `AssertThatTests.cs`

src/NUnitFramework/nunitlite.tests/TextUIReportTests.cs
jnm2: Definitely not inlining. The test method is invoked directly via reflection so inlining is impossible, and it's the test method that is the missing stack frame. If Assert.Warn inlines into the test method that's totally fine because we still see the test method stack frame, but if the test method tail-calls Assert.Warn then the test method stack frame is lost before the call.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -455,25 +458,26 @@ protected void ChangeResult(ResultState resultState, string message)
...
 #if !NETSTANDARD1_6
         private Thread thread;
-        private int nativeThreadId;
 
         private void RunOnSeparateThread(ApartmentState apartment)
         {
             thread = new Thread(() =>
             {
+                thread.CurrentCulture = Context.CurrentCulture;
+                thread.CurrentUICulture = Context.CurrentUICulture;
```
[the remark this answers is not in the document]
jnm2: It's in the commit message:

> Fixed "this operation must be performed on the same thread as that represented by the Thread instance" on .NET Core

src/NUnitFramework/tests/Internal/CallContextTests.cs
```diff
@@ -21,7 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-#if PARALLEL
+#if NET20 || NET35 || NET40 || NET45
```
mikkelbu: `#if PLATFORM`?

src/NUnitFramework/tests/Internal/RuntimeFrameworkTests.cs
```diff
@@ -21,7 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-#if !NETCOREAPP1_1
+#if !(NETCOREAPP1_1 || NETCOREAPP2_0)
```
mikkelbu: `#if PLATFORM`?

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
```diff
@@ -962,7 +962,7 @@ public void ExecutionStatusIsPromulgatedAcrossBranches()
 
 #region Cross-domain Tests
 
-#if !NETCOREAPP1_1
+#if NET20 || NET35 || NET40 || NET45
```
mikkelbu: `#if PLATFORM`?

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
```diff
@@ -1017,7 +1017,7 @@ private Task<T[]> WhenAllAsync<T>(params Task<T>[] tasks)
 #endregion
     }
 
-#if !NETCOREAPP1_1
+#if NET20 || NET35 || NET40 || NET45
```
mikkelbu: `#if PLATFORM`?

src/NUnitFramework/framework/Internal/AssemblyHelper.cs
```diff
@@ -92,7 +92,7 @@ public static AssemblyName GetAssemblyName(Assembly assembly)
 
         #region Load
 
-#if NETSTANDARD1_6
+#if NETSTANDARD1_6 || NETSTANDARD2_0
```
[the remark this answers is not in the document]
jnm2: I like to try to call out the relevant platforms. This only works on .NET Core but the closest I can get is `NETSTANDARD1_6 || NETSTANDARD2_0`. .NET Framework is not relevant here in the same way that if we had a UWP target it would also not be relevant here. At least that's the angle I saw. Do you see a different angle?

src/NUnitFramework/tests/Syntax/InvalidCodeTests.cs
```diff
@@ -21,7 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-#if !NETCOREAPP1_1
+#if !(NETCOREAPP1_1 || NETCOREAPP2_0)
```
mikkelbu: `#if PLATFORM`?

src/NUnitFramework/tests/Syntax/TestCompiler.cs
```diff
@@ -21,7 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-#if !NETCOREAPP1_1
+#if !(NETCOREAPP1_1 || NETCOREAPP2_0)
```
mikkelbu: `#if PLATFORM`?

src/NUnitFramework/testdata/TestCaseAttributeFixture.cs
jnm2: Let's see who wins 😁

src/NUnitFramework/tests/Assertions/AdhocTestExecutionTests.cs
jnm2: I'm not seeing anything specific to the NUnit platform-detection feature in this file. In `AssertThatTests` we're using a PlatformAttribute which must be #if'd out in order to compile, so O used `#if PLATFORM` there because the definition of PlatformAttribute itself exists iff `PLATFORM` is defined.

jnm2:
@mikkelbu Thanks for the sharp eye. I've raised an issue at dotnet/standard to learn more about NETStandard.Library in a nuspec.

src/NUnitFramework/framework/Internal/AssemblyHelper.cs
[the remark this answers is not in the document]
mikkelbu: No. I'm just a bit confused 😄 .

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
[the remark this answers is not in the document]
mikkelbu: Sorry, overlooked that in all the code.

src/NUnitFramework/framework/Internal/AssemblyHelper.cs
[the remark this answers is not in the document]
jnm2: Okay. If for example NUnit had both .NET Standard and .NET Core targets, I'd have used `#if NETCOREAPP1_1 || NETCOREAPP2_0` instead because that section only works on .NET Core.

src/NUnitFramework/tests/Assertions/AdhocTestExecutionTests.cs
mikkelbu: Ok. I think I understand it now (and misunderstod the `PLATFORM` define). I'll delete most of my comments below 😄 .

src/NUnitFramework/framework/Internal/AssemblyHelper.cs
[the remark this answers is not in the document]
jnm2: And then I also wouldn't need to use reflection 😆

jnm2:
@mikkelbu These are all helpful questions.

mikkelbu:
Sorry for a huge number of perhaps stupid questions, but I'm trying to get my head around when we use `#if NET20 || NET35 || NET40 || NET45` and when we use `#if PLATFORM`.

Also why we sometimes translate `!NETSTANDARD1_6` into `#if NET20 || NET35 || NET40 || NET45` and some into `#if PLATFORM` (and similar for the negation, i.e. `#if NETSTANDARD1_6`)

Edit: I've deleted the questions and you answered my questions above.

jnm2:
@mikkelbu These are all helpful questions. Maybe I could explain better in BUILDING.md.
`PLATFORM` means we're enabling the APIs and behaviors of platform detection, mainly deciding whether or not a test case should run based on the OS or runtime. Specifically `PlatformAttribute`, `TestCaseAttribute.IncludePlatform`, etc.

I should probably have called it `PLATFORM_DETECTION`...

jnm2:
@mikkelbu Just catching up with you. So what do you think, is `PLATFORM_DETECTION` an improvement?

mikkelbu:
@jnm2 I think so.

CharliePoole:
@rprouse We previously documented our use of conditional compilation here: https://github.com/nunit/docs/wiki/Standard-Defines. It looks like it has not been maintained. IMO, it should be deleted if you prefer to keep the documentation elsewhere, otherwise it's misleading.

CharliePoole:
How about PLATFORM_DETECTION rather than just PLATFORM?

jnm2:
Sounds good.

Thanks Charlie! Tracking at https://github.com/nunit/docs/issues/245.

jnm2:
CI succeeded! Just need a review on the final commit before we can merge.

ChrisMaddock:
🎉 Awesome - great to see this in @jnm2. 😄

PR #2615, opened 2017-12-12
Fix for failure to retry when tests have no setup and teardown

### LINKED ISSUE #2325: Retry attribute doesn't retry the test.

jnm2:
Fixes #2325

(The fourth commit is an example of the TestCommand API being a little unsettling for me. I wonder if there's a way to make it less magical in the future.)

### DIFF
--- a/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
@@ -94,13 +94,15 @@ namespace NUnit.Framework
                     try
                     {
                         context.CurrentResult = innerCommand.Execute(context);
-
-                        if (context.CurrentResult.ResultState != ResultState.Failure)
-                            break;
                     }
-                    catch when (count > 0)
+                    catch (Exception ex)
                     {
+                        if (context.CurrentResult == null) context.CurrentResult = context.CurrentTest.MakeTestResult();
+                        context.CurrentResult.RecordException(ex);
                     }
+
+                    if (context.CurrentResult.ResultState != ResultState.Failure)
+                        break;
 
                     // Clear result for retry
                     if (count > 0)
--- a/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
@@ -91,10 +91,16 @@ namespace NUnit.Framework
 
                 while (count-- > 0)
                 {
-                    context.CurrentResult = innerCommand.Execute(context);
+                    try
+                    {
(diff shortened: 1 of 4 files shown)


rprouse:
☹️ 

1) Failed : NUnit.Framework.Assertions.CollectionAssertTest.PerformanceTests(System.Collections.Generic.List`1[System.Double])
Elapsed time of 153.3003ms exceeds maximum of 100ms

jnm2:
@rprouse 

> ☹️

This is happening on basically every other run this evening. At least we're about to turn it into a warning.

jnm2:
@ChrisMaddock, @CharliePoole: Would you like to review this drastically simpler fix?

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
```diff
@@ -91,7 +91,15 @@ public override TestResult Execute(TestExecutionContext context)
 
                 while (count-- > 0)
                 {
-                    context.CurrentResult = innerCommand.Execute(context);
+                    try
+                    {
+                        context.CurrentResult = innerCommand.Execute(context);
+                    }
+                    catch (Exception ex)
```
CharliePoole: Nice... I think I'd add a comment here. Something like "All inner commands, including those created by users, are supposed to handle exceptions themselves, but in case they don't we catch it here."

Of course, some of our own commands are not handling exceptions correctly as we know. But that's for a redesign, IMO.

CharliePoole: I approved, so it's a suggestion. :smile:

Further suggestion would be to do it if more work is likely to wait a while rather than happening soon. I'm being cagey as I know I won't be doing it. 😈

jnm2: I'm just not looking forward to waiting for CI all over again, but that's not a good reason not to add it! 😆

jnm2:
Ready for review of comment commit.

jnm2:
I could always review it myself just to merge it but that could enter grey area. =D

CharliePoole:
I personally prefer the grey area. 😄

rprouse:
@jnm2 for these "Approved with suggestions" and minor fixes, feel free to re-approve and merge yourself. I've been considering turning off the dismiss reviews because of this, but I would like to keep it on for non-team members 🙍‍♂️

jnm2:
👂

PR #2622, opened 2017-12-16
Fix flakey test

ChrisMaddock:
Just hit a test failure because two tests were trying to access `Test1.txt`.

To fix this, I added an overload to the `TestFile` utility which uses `Path.GetTempFileName()` when the file name doesn't matter. (It then turned out the actual file name didn't matter in any case where `TestFile` was used.


ChrisMaddock:
@CharliePoole - Yes, `FileOrDirectoryExistsConstraintTests` and `FileAssertTests`. I don't think it's anything new, maybe I'm just unlucky today. ¯\_(ツ)_/¯

Issue #2033, opened 2017-02-04
Nameof refactor


jnm2:
In the [C# version conversation](https://groups.google.com/forum/#!topic/nunit-developer/m00Oeo9bSqA), I asked if we'd be interested in doing a total `nameof` refactor now that the keyword is available. The core reason is well stated by @ChrisMaddock:

>  I'd personally be a fan of doing it globally for nameof(), if nothing else. nameof adds compile time checking where there currently is none, which would be great of all the reflection code within the framework. It would be some nice extra security for future refactors.
>
> When we first moved to VS 2015, nameof was the only thing we 'bulk converted', and it caught a couple of bugs in rotten-code. I'm sure NUnit is better maintained, but I feel it could be justified as a 'code enhancement' rather than just a syntactic nicety.

Besides the security, I've also noticed it makes navigation to the related member *so much easier*. Whether it's a TestCaseSource or it's an NUnit test which references a method in a helper test fixture, a single F12 press jumps you right where to want to be. I guess CodeLens would also pick up on it and provide navigation in the opposite direction.

As noted, the refactor itself is not difficult, especially with ReSharper. It would involve a single click to take care of attributes such as `TestCaseSource`. However, we'd need to manually check all call sites of `TestAssert.IsRunnable` and Properties keys and similar methods. I volunteer because this is something I really want to see happen. Of course extra eyes on the same hunt couldn't hurt. Maybe I can twist the arm of @ChrisMaddock's inner perfectionist. :smile:

jnm2:
However, we'd need to manually check all call sites of `TestAssert.IsRunnable` and similar methods. I volunteer because this is something I really want to see happen, and of course extra eyes on the same hunt couldn't hurt.

CharliePoole:
Reposting steps I suggested in the discussion on the mail list...

0. Decide if we even need such a standard or if we will view it as a
matter of individual style.
1. If yes, try to devise a standard that tells people when to use nameof.
2. Get it approved by either a project team or the core team. This has
to wait a little till we are organized, but I hope not long.
3. Examine the standard to see if it's even capable of being automated.
4. Decide project by project whether to do a mass (possibly automated)
approach or to leave it to the brains of the developers.

I marked this for discussion. Please, no PRs till we make it a backlog item.

If we were to merely allow Resharper to decide what to change, I believe we would be adopting by default a standard that says "Developers are required to use nameof whenever it is possible." That isn't a standard statement I would agree with and I suspect most of us wouldn't either.

[...]

PR #2623, opened 2017-12-16
Initial nameof conversion

ChrisMaddock:
Fixes https://github.com/nunit/nunit/issues/2033. 

@yaakov-h's [comment](https://github.com/nunit/nunit/pull/2597#issuecomment-349589183) made me think we should make some movement on this one:

> Can we use nameof? I noticed the existing code wasn't using it

I've no doubt this isn't everything, but hopefully it's a good start, and we can patch up other places as we go. Things covered include:

- ArgExceptions 
- TestCaseSources
- Guard class
- TestAsserts
- TestBuilders

I've reviewed the changes by hand, and reverted changes to the mono classes that we don't generally edit. Plenty of places where `nameof` has gone in would be nicer with string interpolation or Guards, but let's work on one improvement at a time! :smile:


CharliePoole:
This looks safe to merge, although I personally would not have advocated doing it as a single large fix.

PR #2624, opened 2017-12-16
Prevent emails for successful builds on Travis

ChrisMaddock:
Does anyone else get irritated by getting emails from Travis on every build? This will prevent emails on success - and mean the build author only gets an email on failure.

What do people think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: csharp
 sudo: false
 
+notifications:
+  email:
+    on_success: never
+
 # Need 2.*, should keep up to date with security releases
 dotnet: 2.0.3
 


mikkelbu:
@ChrisMaddock The change looks fine, but I don't get emails from Travis on every build? It only happens once in a while (but perhaps most of the emails from Travis end in the spam folder?)

CharliePoole:
Actually, I think it defaults to sending the email only on failure or when a previous failure changes to a success. The latter are rather handy, in fact.

jnm2:
That's a hard one for me to answer. Sometimes they are annoying and sometimes not. Success emails are nice when waiting for hours for OS X before merging. Seems like it should be a per-user decision.

ChrisMaddock:
If some people like it, we can leave as is. I can just filter them on my side. 😄 

Closing...

Issue #2626, opened 2017-12-17
SetUp/TearDown methods are invoked multiple times before/after test in .NET Standard targeted projects


jalbrzym:
When I run test that is in test fixture that **inherits** from class containing **virtual method** in base class marked as [SetUp]/[TearDown] AND other derived classes overrides this method, then method **[SetUp] is invoked multiple times** before test execution AND method marked as **[TearDown] is invoked multiple times** after test execution - as many times as method is overriden throughout chain of inheritance. 

This issue occurs **only in .NET Standard** targeted projects.

Platform:
- NUnit 3.9.0
- NUnit 3 TestAdapter 3.9.0
- Visual Studio 2017 (v. 15.5.2)
- .NET Core SDK 2.1.2
- .NET Standard 1.6 targeted project

Sample:

```csharp
[TestFixture]
public class SampleIntegrationTest : SqlServerDatabaseIntegrationTest
{
    private int _setupCounter = 0;

    [Test]
    public void SetUpShouldInvokeOnce()
    {
        Assert.AreEqual(1, _setupCounter);
    }

    // SetUp method runs 3 times before test
    public override void SetUp()
    {
        ++_setupCounter;

        base.SetUp();
        // Additional setup logic
    }
}

public abstract class SqlServerDatabaseIntegrationTest : DatabaseIntegrationTest
{
    public override void SetUp()
    {
        base.SetUp();
        // Additional setup logic
    }
}

public abstract class DatabaseIntegrationTest : BaseIntegrationTest
{
    public override void SetUp()
    {
        base.SetUp();
        // Additional setup logic
    }
}

public abstract class BaseIntegrationTest
{
    [SetUp]
    public virtual void SetUp()
    {
    }
}
```

jalbrzym:
After quick investigation I found that there is a problem with GetMethods extension method: 
[https://github.com/nunit/nunit/blob/04c6b7b4984a01831be94baeab0744b1cc686980/src/NUnitFramework/framework/Internal/Reflect.cs#L78](url)

.NET Standard targeted project uses custom extension method:
https://github.com/nunit/nunit/blob/3c028034e717c79bf00661c5852b358f9097d2e1/src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs#L332

As opposed to GetMethods standard implementation in Type class, it returns all methods throughout inheritance chain. This causes a polymorphic multiple call to the SetUp/TearDown method.

ChrisMaddock:
[...]

PR #2633, opened 2017-12-19
Fixed GetAllMethods for multi-level inheritance scenario - .NET Standard ReflectionExtensions (#2626)

jalbrzym:
This is my proposal to fix issue #2626. 

The previous implementation of **GetAllMethods** (ReflectionExtensions for .NET Standard) worked properly only for simple two-level inheritance scenario. My change adds multi-level inheritance support for **GetAllMethods**. When virtual method is declared in base class and is overriden in derived classes then **GetMethods** returns a methods top most in inheritance hierarchy, so virtual **SetUp/TearDown** methods are correctly invoked only once before/after each test.

### DIFF
--- a/src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
+++ b/src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
@@ -348,7 +348,7 @@ namespace NUnit.Compatibility
             if (type != null)
             {
                 var baseMethods = type.GetAllMethods(includeBaseStatic)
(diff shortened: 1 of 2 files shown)


jalbrzym:
Sorry, I accidentally closed PR for a while :)

jnm2:
@ jalbrzym No worries. I was asleep 😁 That close button subconsciously looks way too much like a cancel button, IMO.

jnm2:
We appreciate you taking time to improve NUnit. Thank you! 🎉

PR #2640, opened 2017-12-22
Correctly update add and remove counts when saving and restoring queues

### LINKED ISSUE #2632: Parallel tests are loading 100% CPU when nested SetUpFixture exists
Using 3.9.0.0 NUnit framework and adapter,  I have similar to https://github.com/nunit/nunit/issues/2507 issue.

We have [SetUpFixture] class in root namespace and [SetUpFixture] class in nested namespace.

Please refer code in https://github.com/undron/NUnitHangsAtNestedSetUpFixture
When tests from namespace NUnitParallels.Some is running it is 100% CPU utilization:
image

CharliePoole:
Fixes #2632 

This fix has the advantage that it's quick and seems to work. It's disadvantage is that it makes a complicated class a bit more complicated. Nevertheless, I think we should get it in place and review how the queue class works with a bit more care before possibly making some changes.

@oznetmaster You're the expert on this class, so I'd like your review.

It's possible this may fix some other issues around CPU utilization and hanging in the presence of non-parallel setup fixtures.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
@@ -77,10 +77,14 @@ namespace NUnit.Framework.Internal.Execution
         private class SavedState
         {
             public ConcurrentQueue<WorkItem>[] InnerQueues;
+            public int AddId;
+            public int RemoveId;
 
             public SavedState(WorkItemQueue queue)
             {
                 InnerQueues = queue._innerQueues;
+                AddId = queue._addId;
+                RemoveId = queue._removeId;
             }
         }
 
@@ -135,6 +139,8 @@ namespace NUnit.Framework.Internal.Execution
 
             for (int i = 0; i < PRIORITY_LEVELS; i++)
                 _innerQueues[i] = new ConcurrentQueue<WorkItem>();
+
+            _addId = _removeId = 0;
         }
 
 #region Properties
@@ -359,17 +365,19 @@ namespace NUnit.Framework.Internal.Execution
             // else can take a look at it later on.
             //Guard.OperationValid(State != WorkItemQueueState.Running, $"Attempted to restore state of {Name} while queue was running.");
 
-            var savedQueues = _savedState.Pop().InnerQueues;
+            var state = _savedState.Pop();
 
             // If there are any queued items, copy to the next lower level
             for (int i = 0; i < PRIORITY_LEVELS; i++)
             {
                 WorkItem work;
                 while (_innerQueues[i].TryDequeue(out work))
-                    savedQueues[i].Enqueue(work);
+                    state.InnerQueues[i].Enqueue(work);
             }
 
-            _innerQueues = savedQueues;
+            _innerQueues = state.InnerQueues;
+            _addId += state.AddId;
+            _removeId += state.RemoveId;
         }
 
 #endregion


src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
```diff
@@ -359,17 +365,19 @@ internal void Restore()
...
                 WorkItem work;
                 while (_innerQueues[i].TryDequeue(out work))
-                    savedQueues[i].Enqueue(work);
+                    state.InnerQueues[i].Enqueue(work);
             }
 
-            _innerQueues = savedQueues;
+            _innerQueues = state.InnerQueues;
+            _addId += state.AddId;
+            _removeId += state.RemoveId;
```
rprouse: These two lines bother me. As you say, @oznetmaster is the expert in this area, but everywhere else in the code we use `Interlocked.CompareExchange` to change these values, plus the `_addId` and `_removeId` start at `int.MinValue` and increment, so won't this subtract from the ids and then eventually start adding to them? That doesn't seem like it is restoring the values. How am I reading this wrong?

CharliePoole: If it's possible for anyone else to be changing the queues while we are saving or restoring the state, then we have lots bigger problems than the counts! We rely on the fact that the methods are only called when nothing else can change the queue state.

This reliance on being called correctly has bothered me as well. Can we place a Guard statement somewhere such that it will throw if the precondition is not satisfied?

As to the specific issue, I wanted the newly created levels to start at zero so I could add them into the previous level when restoring the state. An alternative approach would be to leave the values as they are - truly restore them - but that will only work if the queue we are restoring had completely balanced adds and removes. That is not always the case, which is why I copy any "left over" queue items. If you think it's better, I could simply count the number of items I copy and add those values to the `_addId`.

Bear in mind that the only purpose of these integers is to provide two monotonic increasing values as a unique signature for the state of the queue as well as an indication of whether the queue contains anything.

rprouse: Sounds good, thanks for the explanation. As I said before, I find this non-locking code to be overly complex and I wonder if it is truly necessary. One day, I would like to do a spike, switch it to locking code, see how much cleaner it is and also run some timings on large test suites.

The `_addId` and `_removeId` usages might be clearer if they were `uint` so they started at zero, but  `Interlocked.CompareExchange` doesn't have an overload for that 😦

CharliePoole:
@rprouse Ping: you approved but wanted to leave this open 10 days ago so @oznetmaster could review.

oznetmaster:
I will be able to look at it later today (Perth time).  At breakfast now.

src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
oznetmaster: @rprouse Yes, non-locking code is more complex then locking code.  Look at one of the concurrent collection classes if you want to blow your mind.  However, the speed difference can really be significant. 

@CharliePoole I do not think that the ids ever need to be saved and restored.  They are really only used to provide the non-locking control over the queue, to make sure that some other thread has not gotten there before you.  As long as they are unique, and continue to increase, they should just work, no matter what level of queue stack they are used on.

oznetmaster: @CharliePoole I guess I am confused. "but that will only work if the queue we are restoring had completely balanced adds and removes.".  I am not sure why that would be the case?

However, if new items are "copied" to the restored queue as part of restoring it, then the _addId should already reflect those items having been previously added, and since they have not been removed, the _removeId should reflect that as well.

oznetmaster: This would also imply that the initialization of the two ids should be left at int.MinValue.

CharliePoole: @oznetmaster The reason for my confusing comment is that there is code in `WorkItemQueue` that uses the difference between adds and removes to determine if the queue has data. I don't know if it's your code or if somebody else added it, but that was the source of the error that this fixes. So I tried to make sure that adds and removes increase in a monotonic fashion __and__ that the difference between them always represents the number of items in the queue. Proof of the pudding is that it fixed the problem it was meant to fix.

Your comment starting "However..." is correct. That was my intent in adding in the adds and removes.

I use an initialization of zero on any isolated queues that are created so that the value can be meaningfully added to the original queue when we go back to it. That is, the original queue ends up with the values it would have had if we had continued to use it rather than creating an isolated queue. As you say, it's more complicated than a simple lock. 😄

oznetmaster: If you do not save and restore the two ids, and do the following, everything should work correctly.

You should replace 
```
cachedRemoveId == cachedAddId
```
with
```
IsEmpty
```

and then the relative values of the two ids will no longer matter, since they will not be used to determine whether the queue is empty.

I think doing it this way should be more reliable.  I have not actually tried it, however :(

CharliePoole: Yes, I didn't change that because I wasn't sure at the time how the ids worked and why you were making that comparison. Instead, since my introduction of save / restore had broken the add and remove ids, I decided to fix them so they would end up with the value they would have had without the save and restore.

If you see a problem with what is done, i.e. if it will cause problems, then it can either wait till I'm able to go back to it or somebody else can take over and redo it. Originally, the problem this solves appeared to be quite critical, but since nobody has been complaining, it may not have been. I wish I could have had your eyes on this when I was doing it! Right now it would take me a bit of effort to bring my mind back to this PR. Do you want to take it on?

oznetmaster: Since the environment I currently support NUnit in is with CF on a single core processor, I really have no way to adequately test any changes I might make.

I think someone else will need to take this on.

CharliePoole: Use of IsEmpty in the middle of the dequeue loop causes an exception. See my further comment below.

CharliePoole:
OK... I cleared off some other stuff and got back to this. Since the discussion has rambled a little, I'll summarize...

We are trying to eliminate CPU pegging at 100% when dealing with nested SetUpFixtures. I'm testing using the sample code provided in the issue while monitoring performance. The goal is to eliminate the 100% usage while not breaking any existing tests.

With the code in the PR, that's what happens. CPU utilization on my laptop remains at about 18% throughout the test run. The frameowork tests all continue to pass.

I tried two other approaches, based on @oznetmaster 's input:

1. Keeping my changes but using `IsEmpty` rather than `cachedRemoveId == cachedAddId`.
2. Removing my changes and using `IsEmpty` rather than `cachedRemoveId == cachedAddId`.

Both alternative approaches result in an exception. I didn't debug further. My assumption is that the comparison of the two ids was used in the code precisely because IsEmpty is not usable at that point in the code. The non-locking code seems to be pretty carefully thought out and I think I understand what's going on now that I've worked for it a while. (Wish it had an explanatory comment though)

There are two weaknesses I see in the PR code right now:

1. Each new level of saving the queues starts with the same value. I don't think this is a problem because all workers are stopped at the point where the queues are saved or restored.

2. The starting value for secondary queues is zero rather than min int. That gives only half the room for growth. Because that counting space is only used for a single fixture before the queue is restored, I don't think it's a real issue and it makes the code simpler when the queue is restored and the values are added in to the lower level. The lowest level queue (original queue) ends up with the same values it would have had if no save / restore had occured.

This was written as a quick fix and I think we are spending too much time worrying about it. We have a PR that completely changes how the queue works anyway. If we intend to pursue that, we can either merge this for interim use or just drop it.

@oznetmaster As I understand your comments, you are not pointing out anything problematic in the changes, but are questioning their necessity. Do I have that right? However, I wasn't able to figure out any other way to eliminate the CPU pegging that is taking place.

oznetmaster:
@CharliePoole Correct.

PR #2644, opened 2017-12-29
Updated build.sh so that the correct version of Cake is used

jnm2:
Build.sh was so old that it was not respecting the pinned Cake version at `tools/packages.config`. This made a long Travis debugging session longer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,4 +21,4 @@ matrix:
       osx_image: xcode9.1
 
 script:
-  - ./build.sh --target "Travis" --configuration "Release"
+  - ./build.sh --target=Travis --configuration=Release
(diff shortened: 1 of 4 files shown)


rprouse:
Same question, should the `build.ps1` also be updated?

jnm2:
@mikkelbu @rprouse Sure, consistency would be good. I can do that.

rprouse:
Cake 0.24 was just released. Do we want to update at the same time?

jnm2:
I'm happy to do that.

jnm2:
If CI passes, this should be good to go.

jnm2:
CI is happy on the first try. =) Just need a review then.

PR #2645, opened 2017-12-30
Stop dotnet publish on single project from restoring and building projects a second time

jnm2:
I keep meaning to do this. The lack of redundant restores should cut the time down ten seconds or so.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -141,26 +141,29 @@ Task("Build")
     .IsDependentOn("NuGetRestore")
     .Does(() =>
     {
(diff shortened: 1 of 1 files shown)


jnm2:
Made a mistake. Time to break out [msbuildlog.com](http://msbuildlog.com).

jnm2:
Ah, it was `PublishDir`, not `OutputDirectory`, that I wanted.

rprouse:
LGTM

Issue #2647, opened 2018-01-05
Add Current Attempt indicator in TestContext for use with RetryAttribute


bicarbon8:
the RetryAttribute https://github.com/nunit/nunit/blob/638b4011b102e506a42f65baa515abb010710a2f/src/NUnitFramework/framework/Attributes/RetryAttribute.cs currently uses an internal _count_ variable within its _Execute_ method to track the remaining number of Attempts that can / should be made before giving up (in the case of a constantly failing test).

I would like to request that this instead be a _currentAttempt_ variable that is added to the TestContext so I can read the value from within my test. This would allow me to indicate in my test logs which attempt I am on thereby making debugging via logs an easier process. Additionally, I could use the field to conditionally try additional steps if the number of failures exceeds a certain threshold thereby allowing for additional mechanisms for ensuring robustness of tests

CharliePoole:
I see you've done a PR. It's usually a good idea to wait for the issue to be accepted first, in order to avoid wasting your time. It's possible that the team may not want to make this change or may want to make it in a different way from what you propose.

Personally, I like this idea, although I can see some dangers in it. If tests have the retry count available, they may end up written to do something entirely different on each attempt. While that may be considered a feature, I'd call it an anti-pattern.  Nevertheless, the benefit of being able to log the count probably outweighs the dangers.

I made this a low-priority feature. If any @nunit/framework-team  member disagrees we should discuss.

That said, I note that your PR does __not__ implement this as proposed here. Rather than make the count a property of the context, you made it a property of the test. I'll comment on the implementation in #2648.

jnm2:
Rather than exposing the value programmatically, what about automatically logging when retries (and repeats) happen?

CharliePoole:
@jnm2 Could you expand on that idea? We don't really log anything else automatically, so it's a bit unclear to me.

jnm2:
We could start automaticaly adding lines to the log such as "Try 1 of 3 [RetryAttribute]" or we could prefix every line "[retry 2]" though that doesn't scale as well. That solves the logging problem without making the antipattern easy.

CharliePoole:
You mean the internal trace log?

jnm2:
I was thinking `TestContext.WriteLine`. What if `RetryAttribute` defaulted to doing `TestContext.WriteLine("Try 2 of 3 [RetryAttribute]")` before each *re*try?

bicarbon8:
[...]

PR #2648, opened 2018-01-08
Enable access to the current Retry attempt count from within a Test

bicarbon8:
### Changes:
* modification to RetryAttribute such that it exposes the current attempt in the Test properties as per issue #2647
* addition of unit test that uses the newly exposed value to set a _Count_ field which is checked at the end of execution and compared to the expected number of executions performed. If these match then it means the value exposed via the _TestContext.CurrentContext.Test.Properties_ is working as expected

### DIFF
--- a/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
@@ -87,12 +87,15 @@ namespace NUnit.Framework
             /// <returns>A TestResult</returns>
             public override TestResult Execute(TestExecutionContext context)
             {
-                int count = _tryCount;
+                int currentAttempt = 0;
 
-                while (count-- > 0)
+                while (currentAttempt++ < _tryCount)
                 {
                     try
                     {
+                        // update CurrentAttempt property in test
+                        context.CurrentTest.Properties.Set(PropertyNames.CurrentAttempt, currentAttempt);
+
                         context.CurrentResult = innerCommand.Execute(context);
                     }
                     // Commands are supposed to catch exceptions, but some don't
@@ -107,7 +110,7 @@ namespace NUnit.Framework
                         break;
 
                     // Clear result for retry
-                    if (count > 0)
+                    if (currentAttempt < _tryCount)
                         context.CurrentResult = context.CurrentTest.MakeTestResult();
                 }
 
--- a/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
@@ -87,15 +87,12 @@ namespace NUnit.Framework
             /// <returns>A TestResult</returns>
             public override TestResult Execute(TestExecutionContext context)
             {
-                int currentAttempt = 0;
+                int count = _tryCount;
 
-                while (currentAttempt++ < _tryCount)
+                while (count-- > 0)
                 {
                     try
                     {
-                        // update CurrentAttempt property in test
-                        context.CurrentTest.Properties.Set(PropertyNames.CurrentAttempt, currentAttempt);
-
                         context.CurrentResult = innerCommand.Execute(context);
                     }
                     // Commands are supposed to catch exceptions, but some don't
@@ -110,8 +107,11 @@ namespace NUnit.Framework
                         break;
 
                     // Clear result for retry
-                    if (currentAttempt < _tryCount)
+                    if (count > 0)
+                    {
                         context.CurrentResult = context.CurrentTest.MakeTestResult();
+                        context.CurrentRepeatCount++; // increment Retry count for next iteration. will only happen if we are guaranteed another iteration
+                    }
                 }
 
                 return context.CurrentResult;
--- a/src/NUnitFramework/framework/Internal/PropertyNames.cs
+++ b/src/NUnitFramework/framework/Internal/PropertyNames.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2010 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -137,6 +137,11 @@ namespace NUnit.Framework.Internal
         /// The optional Order the test will run in
         /// </summary>
         public const string Order = "Order";
+
+        /// <summary>
+        /// The attempt currently being executed in a test using the <see cref="RetryAttribute"/>
+        /// </summary>
+        public const string CurrentAttempt = "CurrentAttempt";
         #endregion
     }
 }
--- a/src/NUnitFramework/framework/Internal/PropertyNames.cs
+++ b/src/NUnitFramework/framework/Internal/PropertyNames.cs
@@ -137,11 +137,6 @@ namespace NUnit.Framework.Internal
         /// The optional Order the test will run in
         /// </summary>
         public const string Order = "Order";
-
-        /// <summary>
-        /// The attempt currently being executed in a test using the <see cref="RetryAttribute"/>
-        /// </summary>
-        public const string CurrentAttempt = "CurrentAttempt";
         #endregion
     }
 }
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -428,6 +428,12 @@ namespace NUnit.Framework.Internal
         /// </summary>
         public bool IsSingleThreaded { get; set; }
 
+        /// <summary>
+        /// The number of times the current test has been scheduled for execution.
+        /// Currently only being executed in a test using the <see cref="RetryAttribute"/>
+        /// </summary>
+        public int CurrentRepeatCount { get; set; }
+
 #endregion
 
 #region Instance Methods
--- a/src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
+++ b/src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
@@ -984,6 +984,16 @@ namespace NUnit.Framework.Internal
         }
 #endif
 
+        #endregion
+
+#region CurrentRepeatCount Tests
+        [Test]
+        public void CanAccessCurrentRepeatCount()
+        {
+            Assert.That(_fixtureContext.CurrentRepeatCount, Is.EqualTo(0), "expected value to default to zero");
+            _fixtureContext.CurrentRepeatCount++;
+            Assert.That(_fixtureContext.CurrentRepeatCount, Is.EqualTo(1), "expected value to be able to be incremented from the TestExecutionContext");
+        }
 #endregion
 
 #region Helper Methods
--- a/src/NUnitFramework/framework/TestContext.cs
+++ b/src/NUnitFramework/framework/TestContext.cs
@@ -181,6 +181,16 @@ namespace NUnit.Framework
             get { return _testExecutionContext.AssertCount; }
         }
 
+        /// <summary>
+        /// Get the number of times the current Test has been repeated. This is currently only 
+        /// set when using the <see cref="RetryAttribute"/>.
+        /// TODO: add this to the RepeatAttribute as well
+        /// </summary>
+        public int CurrentRepeatCount
+        {
+            get { return _testExecutionContext.CurrentRepeatCount; }
+        }
+
         #endregion
 
         #region Static Methods
--- a/src/NUnitFramework/testdata/RetryFixture.cs
+++ b/src/NUnitFramework/testdata/RetryFixture.cs
@@ -23,6 +23,7 @@
 
 using System;
 using NUnit.Framework;
+using NUnit.Framework.Internal;
 
 namespace NUnit.TestData.RepeatingTests
 {
@@ -201,6 +202,17 @@ namespace NUnit.TestData.RepeatingTests
         {
             Count++;
             throw new Exception("Deliberate exception");
+        }
+    }
+
+    public class RetryTestVerifyAttempt : RepeatingTestsFixtureBase
+    {
+        [Test, Retry(3)]
+        public void NeverPasses()
+        {
+            string currentAttemptStr = TestContext.CurrentContext.Test.Properties.Get(PropertyNames.CurrentAttempt).ToString();
+            count = int.Parse(currentAttemptStr);
+            Assert.Fail("forcing a failure so we retry maximum times");
         }
     }
 }
--- a/src/NUnitFramework/testdata/RetryFixture.cs
+++ b/src/NUnitFramework/testdata/RetryFixture.cs
@@ -210,9 +210,19 @@ namespace NUnit.TestData.RepeatingTests
         [Test, Retry(3)]
         public void NeverPasses()
         {
-            string currentAttemptStr = TestContext.CurrentContext.Test.Properties.Get(PropertyNames.CurrentAttempt).ToString();
-            count = int.Parse(currentAttemptStr);
+            count = TestContext.CurrentContext.CurrentRepeatCount;
             Assert.Fail("forcing a failure so we retry maximum times");
+        }
+
+        [Test, Retry(3)]
+        public void PassesOnLastRetry()
+        {
+            Assert.That(count, Is.EqualTo(TestContext.CurrentContext.CurrentRepeatCount), "expected CurrentRepeatCount to be incremented only after first attempt");
+            if (count < 2) // second Repeat is 3rd Retry (i.e. end of attempts)
+            {
+                count++;
+                Assert.Fail("forced failure so we will use maximum number of Retries for PassesOnLastRetry");
+            }
         }
     }
 }
--- a/src/NUnitFramework/tests/Attributes/RetryAttributeTests.cs
+++ b/src/NUnitFramework/tests/Attributes/RetryAttributeTests.cs
@@ -101,5 +101,15 @@ namespace NUnit.Framework.Attributes
             Assert.AreEqual(1, categories.Count);
             Assert.AreEqual("SAMPLE", categories[0]);
         }
+
+        [Test]
+        public void RetryUpdatesCurrentAttemptProperty()
+        {
+            RepeatingTestsFixtureBase fixture = (RepeatingTestsFixtureBase)Reflect.Construct(typeof(RetryTestVerifyAttempt));
+            ITestResult result = TestBuilder.RunTestCase(fixture, "NeverPasses");
+
+            Assert.AreEqual(fixture.TearDownResults.Count, fixture.Count, "expected the CurrentAttempt property to equal the number of executions");
+            Assert.AreEqual(result.Test.Properties.Get(PropertyNames.CurrentAttempt), fixture.Count, "expected the CurrentAttempt property in the result to be set correctly after execution");
+        }
     }
 }
--- a/src/NUnitFramework/tests/Attributes/RetryAttributeTests.cs
+++ b/src/NUnitFramework/tests/Attributes/RetryAttributeTests.cs
@@ -103,13 +103,23 @@ namespace NUnit.Framework.Attributes
         }
 
         [Test]
-        public void RetryUpdatesCurrentAttemptProperty()
+        public void RetryUpdatesCurrentRepeatCountPropertyOnAlwaysFailingTest()
         {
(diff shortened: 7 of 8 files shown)


CharliePoole:
Please use Fixes #2647 in initial description or a comment (as I just did) rather than putting the issue number in the title of the PR. That will cause the issue to be automatically closed when the the PR is merged.

src/NUnitFramework/framework/Internal/PropertyNames.cs
```diff
@@ -137,6 +137,11 @@ public class PropertyNames
         /// The optional Order the test will run in
         /// </summary>
         public const string Order = "Order";
+
+        /// <summary>
+        /// The attempt currently being executed in a test using the <see cref="RetryAttribute"/>
+        /// </summary>
+        public const string CurrentAttempt = "CurrentAttempt";
```
CharliePoole: When used in the code of `RetryAttribute` the name `PropertyNames.CurrentAttempt` makes sense. When accessed outside of the attribute, however, it's a bit ambiguous. Current attempt to do what? I suggest "CurrentRetryCount".

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
```diff
@@ -87,12 +87,15 @@ public RetryCommand(TestCommand innerCommand, int tryCount)
...
-                int count = _tryCount;
+                int currentAttempt = 0;
 
-                while (count-- > 0)
+                while (currentAttempt++ < _tryCount)
                 {
                     try
                     {
+                        // update CurrentAttempt property in test
+                        context.CurrentTest.Properties.Set(PropertyNames.CurrentAttempt, currentAttempt);
```
CharliePoole: I'm dubious about making the current retry count a property of the test. Test properties are persistent and pertain to the test itself. If results had properties (a change we have considered in the past) then it might make sense to include it there, since the retry count pertains to the __execution__ of the test rather than the test _per se_. Unfortunately, results don't have properties currently, although we have considered adding them in the past.

At the moment, based on the options available  to us, I think this is more a property of the execution context than of the test or the result.

bicarbon8: Thanks @CharliePoole, your comments and direction here are appreciated. Would you then suggest this be a public property of the _TestExecutionContext_ or that a _Properties_ dictionary be added to the _TestExecutionContext_ (I'm guessing the former as it is a smaller change, but the latter adds more flexibility for later additions).

src/NUnitFramework/framework/Internal/PropertyNames.cs
bicarbon8: _CurrentRetryCount_ would be 0 on the first execution of the test as I would understand it though right (since it hasn't yet been retried)? Would _CurrentTestExecutionIteration_ work for you or do you want to keep the association of _RetryAttribute_ with _CurrentRetryCount_?

CharliePoole: Good point. How does `CurrentRepeatCount` seem to you? It could be set by both the Retry and the Repeat attributes. A value of zero would indicate that no repetition was in progress - i.e. neither attribute has been used. I can imagine `Repeat` and `Retry` being merged at some time in the future since `Retry` is just `Repeat` with a particular rule for when to stop.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
CharliePoole: We have considered adding a dictionary to the `TestExecutionContext` - accessible through `TestContext` - for use by users. However, this count is really something internal and I think it's more in keeping with how we define other properties to make it a public property.

BTW, changes to `TestExecutionContext` are internal implementation changes in concept, even though the class is public for purpose of use by extensions. We will feel relatively free to change this in future where we would not feel quite so free to change what is exposed to general users via `TestContext`.

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
```diff
@@ -984,6 +984,16 @@ private class TestClass
...
 
+        #endregion
+
+#region CurrentRepeatCount Tests
+        [Test]
+        public void CanAccessCurrentRepeatCount()
+        {
+            Assert.That(_fixtureContext.CurrentRepeatCount, Is.EqualTo(0), "expected value to default to zero");
+            _fixtureContext.CurrentRepeatCount++;
+            Assert.That(_fixtureContext.CurrentRepeatCount, Is.EqualTo(1), "expected value to be able to be incremented from the TestExecutionContext");
```
CharliePoole: Second assert is a test of C#!

bicarbon8: I put it there just to ensure that the property remains able to be set (i.e. so someone doesn't change this in the future without causing a failure). Will happily change it to some other mechanism to verify that if you'd like

src/NUnitFramework/framework/Internal/PropertyNames.cs
bicarbon8: While `CurrentRepeatCount` as a Property is not my favourite because, to me it doesn't seem as clear as `CurrentTestExecutionIteration` (also not great as it is quite a mouthful), I've updated the code to use `CurrentRepeatCount` and to expose the value in the _TestContext_ by way of the _TestExecutionContext_. Hopefully this change is as you wanted. 😄

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
bicarbon8: reverted most of the changes here and am now incrementing a `CurrentRepeatCount` property in the _TestExecutionContext_ if the test is to be retried

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
CharliePoole: Not sure what you are suggesting might happen... if the property were not settable, then NUnit would not compile. It's just an automatic property, right? The test does no harm, however.

src/NUnitFramework/framework/Internal/PropertyNames.cs
CharliePoole: Well, I only asked if you liked it! 😄 

`CurrentTestExectionIteration` sounds (to me) like we are re-running __all__ the tests, not just one, although I can see how it could mean that too. `CurrentRepeatCount` could be similarly ambiguous, I admit, but at least it matches the existing `Repeat` saved in the test by the `RepeatAttribute`. `CurrentRepetition` might work as well if you want something shorter.

As you have it now is basically as I imagined it. @jnm2 since you have been looking at this would you care to review too?

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
bicarbon8: @CharliePoole, you are correct, in the current implementation (the code I've added to _RetryAttribute_) it would not compile if the setter were removed and in thinking of scenarios where it could be made to still compile, I've convinced myself that it is *VERY* unlikely that someone would go so far as to make such a change unintentionally, so, you're right, that this isn't really needed. Happy to remove it if you would like. As you said though, it does no harm, so either way is ok, I suppose.

CharliePoole:
If you change it now, GitHub will invalidate my approval and we'll have to start over. Currently waiting for a second team approval per our standards.

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
jnm2: This test does more than ensure that you can access the count. it also ensures that custom attributes written from scratch (which we have seen) are also able to mutate that count.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -428,6 +428,12 @@ public IPrincipal CurrentPrincipal
         /// </summary>
         public bool IsSingleThreaded { get; set; }
 
+        /// <summary>
+        /// The number of times the current test has been scheduled for execution.
+        /// Currently only being executed in a test using the <see cref="RetryAttribute"/>
+        /// </summary>
+        public int CurrentRepeatCount { get; set; }
```
jnm2: I feel less than comfortable making the setter public. I'd prefer `{ get; private set; }` with a public `void IncrementCurrentRepeatCount()` method and an *internal* `void ResetCurrentRepeatCount()` method, if we need one. Requesting changes just to discuss this. Thoughts?

jnm2: This would also semantically give us the opportunity in the future to do other work during `IncrementCurrentRepeatCount()` such as raising an event for runners to react to.

CharliePoole: But we don't do other work and aren't even considering it, AFAIK. If we decided to issue an event on retry (which I feel is far-fetched anyway) we would hardly do it as a side effect of incrementing a counter! This is just one of many property setters in `TestExecutionContext`. Shouldn't we present a consistent interface to those who use the class?

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
CharliePoole: I have no problem keeping the test, but it doesn't show that at all, since no custom attribute is involved. For example, change the setter to `internal` and the test will still pass, but a custom attribute would be unable to use it.

You bring up a good point though. Maybe we need a set of custom attribute tests using a phony attribute that accesses everything we expose for them to use. Separate issue, however.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
jnm2: Even if we are deciding now that we aren't going to raise an event, I still want the user code to be forced to use an API indicating that it is repeating the test rather (semantic) than an API which directly affects the counter (implementation). In that regard, `IncrementCurrentRepeatCount` (which I suggested to illustrate the direction I wanted to move in) is a poor name. I'd prefer `BeginTestRepeat()`.

I do agree about consistency so that it doesn't get lost in the other properties, but in that case I really dislike the fact that it is a loose counter. What about a domain object?

```cs
public class TestExecutionContext
{
    public TestRepeatTracker CurrentRepeat { get; }
}
public sealed class TestRepeatTracker
{
    public int Number { get; }
    public void StartNew();
}
```

User attributes call TestExecutionContext.CurrentRepeat.StartNew() and user tests log TestExecutionContext.CurrentRepeat.Number. Resetting is done by internally setting TestExecutionContext.CurrentRepeat` to a new instance.

To me this solution more closely mirrors the problem space. Offering a loose counter variable also solves the problem but it's a leakier abstraction with more implied, unexpressed expectations. How should code react to the possibility that unrelated code might set `-1` or count backwards? I'm afraid of subtle misunderstandings of the unspoken convention, such as what 0 or 1 mean, so I'm trying to encode the expectations into the API.
Once we ship a loose variable and people start depending on being able to set it without following a particular convention, we'll end up being stuck with it even if we want to add a feature on top of it in the future unless we break people.

Now that's my point of view but I don't hold it strongly and don't want to impose if no one is enamored of domain objects to maintain.

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
jnm2: Oh! I forgot about InternalsVisibleTo. Quite right and an interesting topic.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
CharliePoole: @jnm2 I think you might be misinterpreting what happens here - or if not it won't hurt for me to spell it out...

The __attribute__ doesn't itself increment anything. What it does is provide a __command__ that retries the test. The framework simply checks to see if the attribute wants to provide a command and then uses it. Some attributes instantiate an external command while others define it as a nested class. 

That command becomes a temporary part of the execution term and it actually executes the test. So, in other words, it doesn't retry the test as a side effect of incrementing. Rather, the increment is the side effect and isn't used by NUnit in any other way.

To put it another way, the command doesn't __tell__ NUnit to retry the test, it simply __does__ it. As far as NUnit is concerned, there is only one test execution. and the core parts of the framework don't even know the test was retried. So there is no API that causes a test to be repeated.,, we're just recording some new information.

As a minor detail... I'm not understanding why you would ever want to reset the value. That would seem to be an error if it has once been incremented. I could imagine that you might want to allow it to __only__ be incremented rather than set, similar to how we handle assert counts.

jnm2: I was counting the command code as attribute code, but I'll be more precise. I do understand the picture. I'm imagining the user attribute's command's code doing something vaugely like this:

```cs
for (var i = 0; i < max; i++)
{
    innerCommand.Execute(); // Actually perform the execution/repeat
    TestExecutionContext.CurrentContext.CurrentRepeat.StartNew(); // Report new repeat
}
```

I prefer that to `TestExecutionContext.CurrentContext.CurrentRepeatCount = i + 1;`.

You're right, there's no reason to reset since the context is no reused between tests. I wasn't sure.

jnm2: Speaking of assert counts, I wonder if it would be interesting to keep this count on the current result rather than the context? The current result could link to a list of results from the previous repeats. Then the .Count of that collection would be the source of truth for the repeat count property which users could log.

CharliePoole: @jnm2 Regarding the code for a command... I would imagine that the increment or start new would have to take place before executing the inner command. Otherwise, the test would not be able to access the value from the TestContext. But otherwise... yes that makes sense if we want NUnit to know that repeats are being performed. But what would NUnit do with that info?

Regarding assert counts, there's a chicken and egg problem here. At any point where the user code can access the count, then the count is still unfinished. That is, the user can issue an assert at any of those points. Keeping it in the context was intended to signal that it's provisional.

Right now we throw away results from earlier repetitions.

jnm2: In the initial execution of the inner command, no repeats have begun yet the way I was thinking. The second time the inner commend is executed is the first time it is repeated. But I can adopt another terminology.

What would NUnit do with that info? At the least, NUnit could increment the repeat count property. The motivation of having the API step back and express *intent* rather than *implementation* (~like turn signal rather than changing lanes) was more for the benefit of the API consumers than for NUnit. When semantics are only represented as implementation details, the API is more error-prone for the users.
I went on to imagine how the API shape could potentially yield NUnit richer opportunities but even without that I'd still like the API shape.

Now I'm wondering if we can fit it together with test results and make it possible to see all results when diagnosing a problem. It would be cool not to throw it away. Assuming we stop throwing it away, there is where our repeat count would most naturally come from. I'm thinking that's better than what I proposed first.

CharliePoole:
@jnm2 I'm going to step back and go a bit meta on you here.

I think your ideas around repeating tests are great. I'd like to keep them in mind for future work, whether on NUnit's next major release or in a new framework.

However, I do get a little distracted when we branch out from solving a simple, small problem to discussing larger issues of API design, all in the same discussion.

We have an existing design in which commands created by attributes can arbitrarily repeat tests, cancel tests and modify results. If we can't trust the same command to increment a counter, that seems odd.

If you are saying "Let's redesign how attributes and commands participate in the execution of tests" I could agree that is a worthy topic to cover someplace, sometime. I don't think it's something to do while solving this relatively small issue.

Neither do I think it's worthwhile to try to predict in advance how we might do it in the future, in order to make the immediate fix "compatible" with that direction. The future developers (either our future selves or some other people) may decide on a different future than we would imagine today.

What I do think we can do here is come up with a clean way to leave behind a counter that shows how many times the test has been executed. We can make it a settable property, as it is in the PR or a read-only property that is allowed to be incremented only, as we do with assert counts. I'm happy either way. I'm not happy with going beyond the current job of providing a counter - not in this issue, anyway.

jnm2:
Not a problem at all. I can be happy with that.

> If we can't trust the same command to increment a counter, that seems odd.

I'm thinking of the combination of NUnit's built-in commands, custom user commands managing the counter, and user code interpreting the counter. I can imagine all three of these being in play at once. It's more believable to me that they can unify around something like a collection of results from previous repeats than something like a loose int counter. But I'm also sympathetic to just getting this done without the API spiraling into grandiosity! 😆

How far do you want to go towards this API, knowing that I won't really object if you think the int counter has more merit?

```cs
TestExecutionContext.Current.Repeat.StartNew();
TestExecutionContext.Current.Repeat.Number
TestContext.Current.RepeatNumber
```

CharliePoole:
@jnm2 I'd prefer just to finish this issue and discuss a significant API change separately... involving all team members and any users who are interested. 

`<meta>`
The way to do that, IMO, is for somebody who really wants to see it happen to create a discussion issue and carry water for the change. That is, make sure the discussion happens, that everyone is involved, etc. It doesn't seem fair to the user who requested a simple integer to be made available that he should need to get involved in a major discussion about APIs in order to reach that goal.

What's in scope for __this__ issue is the minimum amount of work needed to give the requesting user some equivalent of `TestContext.RepeatCount`, so it can be logged. The operative word here, IMO, is __minimum__.
`</meta>`

WRT __this__ issue and the three alternatives you suggest.

`TestContext.RepeatNumber` would be a simple readonly property to get the repeat count or number. It's a change to the `TestContext` API and is well in scope. We should discuss the best name (I like RepeatCount myself) and whether it starts from zero or 1. The only question is how we get it.

`TestExecutionContext.RepeatNumber` would be the backing storage for `TestContext.RepeatNumber`. It's within our scope here as well. Taken alone, it would have to be a read/write property. It could also be readonly if paired with an IncrementRepeatCount method. As part of `TestExecutionContext`, this would be a hidden implementation from the user (test writer) perspective but part of the API for those who extend NUnit. That's a super-important distinction in my mind.

`TestExecutionContext.StartNew` seems to be both out of scope and a complete reversal of how NUnit works. (The fact that it's a reversal doesn't mean it's bad, just that we need to have lots more discussion somewhere else about it.) NUnit does not start, finish or otherwise carry out repetitions of tests. It's not built in. It can only be done by commands provided through an attribute. SO a command that tells NUnit to start a new repetition would presumably mean completely changing who is responsible for handling repetitions. That's a big deal. I don't want to argue for or against it in the context of solving this issue because it's a small issue we can easily resolve without dealing with all that. In the right context, I'm glad to talk about completely rewriting the framework if you like. 😄

jnm2:
\<meta>
I'm agreeable to that. On the other side, I feel that there too should be a burden of carrying water for an API discussion before adding the new simple public API that this PR implements. If we don't consider alternatives and end up adding whatever seems simplest at the time, we're possibly painting ourselves into a corner. Just because there is a lack of time and effort to think through API design, does that mean that making a long-term commitment to the initial API or simplest API we hit upon is the right decision?

For example, imagine that we add a feature to track the results of past repeats. Do we make a breaking change and remove the public setter or Increment method, or do we allow that at any time the reported repeat count may have nothing to do with the number of reported repeat results? Either option is a potential burden for us and time wasted for NUnit users, and only because we ran out of time and shipped the simplest API.
\</meta>

Meta aside, I'm not too concerned about this particular instance. So I'm down to asking whether the Increment method or the public setter is more in line with where NUnit has been and where we want to be.

CharliePoole:
I agree with you on APIs but I guess I didn't see this as (initially) an API discussion. We know from prior conversations that you construe "API" much more broadly than I do. In this case, the only __general and public__ API is `TestContext`. I don't think we have any disagreement as to how that should be handled.

`TestExecutionContext` may be said to provide an API, for sure, but saying that gives a completely different meaning to the term. It's a class labelled for internal use, which puts general users on notice that they shouldn't use it.  We use it ourselves, but if we change it we can change our calls. It is possible that custom attribute writers may also use it, but they can deal with it just as we do. Since custom attributes can't be used across different versions of the framework there is nothing to prevent changing their implementation if the framework changes.

*****

Please, let's finish this PR before getting into any larger discussions. Just tell the guy what you want him to change and we can get it merged. That's what PRs are for. I've said I'm equally happy with a public setter or a public method that increments the count. I'd be unhappy with a method whose name suggests it's doing anything more than incrementing the count because that's just bad naming.

jnm2:
@CharliePoole `Internal`– that's a good point. I'm more okay with things being lax when they are internal because there breaking changes can be par for the course.

Issue #2413, opened 2017-09-04
Snapshot testing NUnit's external API


jnm2:
Per positive feedback on https://github.com/nunit/nunit/issues/2410#issuecomment-326976168.

- [ ] Add a test that generates a pseudo-C# `pubapi.txt` file off all public and protected symbols in NUnit Framework except the ones in the Internal namespace and fails if it is not identical to the `pubapi.approved.cs` file which is committed to the repo.
    In order to pass the test, the contents of `pubapi.approved.txt` will have to be replaced with the contents of the newly generated `pubapi.txt`. When a changed `pubapi.approved.txt` file shows up in a PR, we must have a compatibility discussion before approving the PR.
    The file will be generated such that any time there is red in the git diff, it represents either a source- or binary-breaking change.

- [ ] Decide whether to .gitignore `pubapi.txt` or keep it beside `pubapi.approved.txt`. The benefit to committing it also is that we could generate it on build rather than on test run and that way you'd have a shorter feedback loop if you make a public API change because it would show up in your git status.

Assigning to myself since I've put a lot of time into exploring this concept, but if someone else really want to claim this they can feel free to edit the assignment and comment.

ChrisMaddock:
The difficult part of this will be https://github.com/nunit/nunit/issues/1426 - it's not totally clear (to me, at least) - what is considered part of our public API, and what isn't. I don't either the 'public keyword' or 'public keyword except the internal namespace' are sufficient definitions - but I'm not really the person to know what is!

As my other comment, in the next major version, it would be great to make as much properly internal as we want to.

CharliePoole:
Well, #2410 had positive feedback, including mine, but also had negative feedback, which you seem to be ignoring. You have not clarified what you consider to be the public API at all. It is not and has never been the `public` keyword. We made all methods public if we wanted to test them, because:
1. The founders came from languages that did not have levels of visibility.
2. We are a test application and wanted to test our stuff
3. There was no supported way except reflection to test non-public members and even after Microsoft invented a way (InternalsVisibleTo) many of us thought that was an ugly kluge and a much greater violation of good coding practice than using public methods

I'm not saying that can't and shouldn't change. I recognize that for the majority of you younger folks working on NUnit now that older point of view doesn't make sense at all. So change it already! But change it in a careful, orderly way:

[...]

PR #2652, opened 2018-01-10
Add ApiContractGenerator MSBuild tool

jnm2:
Closes https://github.com/nunit/nunit/issues/2413

As you edit the NUnit Framework code and Visual Studio or build.cake builds, [this package](https://github.com/nunit/nunit/pull/2652/files#diff-d659fe06375fffbb7706a8687de4f407R11) extracts the <s>public API</s> external shape<sup>1</sup> to a readable diff-friendly file for each output DLL. The default configuration puts these files at `{Project}/../pubapi/{AssemblyName}.{framework}.pubapi.cs` [edit: [changed](https://github.com/nunit/nunit/pull/2652#issuecomment-356718668)] so that basic syntax highlighting works in all editors without interfering with the compilation context of any project.

If by chance or intent your edits have an impact on the <s>public API</s> external shape<sup>1</sup>, the diff is immediately reflected in Git's pending changes. If the alteration was not an accident, you commit the changed <s>.pubapi.cs</s> .shape.cs files along with the source change so that the diff shows up in PR reviews. This helps us be intentional about what sorts of breaking changes we might be approving. (Down the road, I'd like to see if I can make it autogenerate code samples that NUnit users might have written which illustrate the break. I am happy to provide code samples by hand for now.)

If any change in a <s>.pubapi.cs</s> .shape.cs file is excluded from a PR, CI will fail. (This is [detected by NUnit's Cake script](https://github.com/nunit/nunit/pull/2652/files#diff-4d3634747586b964df9f7712bc3ec669R161) when any <s>.pubapi.cs</s> .shape.cs file changes as a result of building from source.) An error message will be shown explaining that those changes must be included in the PR and git history. The same message is shown as a warning instead of an error when you're building on your local machine to help save you time.

Because of this enforced consistency at the PR and branch level, <s>public API</s> external shape changes can't slip by without showing up as modified <s>.pubapi.cs</s> .shape.cs files in reviews. The best part though is that it becomes super easy to get a great diff of NUnit's <s>public API</s> external shape between any two points in history, using GitHub, Visual Studio, or any git diff tool! This makes summarizing changes for a new release a breeze.

<sup>1</sup> <s>‘public API’</s> ‘external shape’ is currently configured to consist of all types and members which are visible to other assemblies, [minus anything in the `NUnit.Framework.Internal` namespace](https://github.com/nunit/nunit/pull/2652/files#diff-d659fe06375fffbb7706a8687de4f407R25).

###### Technical details: Types in that ignored namespace are still included if they are exposed by a type or member outside that namespace in such a way that changes to the internal type could cause source-level breaking changes for NUnit users who only name types outside that namespace; i.e., without having “NUnit.Framework.Internal” written anywhere in their code.

**This tool is to serve the NUnit team, not the other way around.** Let's trial this and if anything is not to your liking, please do make it known.

The tool lives at https://github.com/jnm2/ApiContractGenerator and https://www.nuget.org/packages/ApiContractGenerator.MSBuild. Issues and contributions are welcome there.


CharliePoole:
I have no problem with the concept or implementation. I do take issue (again!) with the terminology. The __public api__ of NUnit has been well defined up to now and did not mean "all public methods minus those in the internals namespace." Rather, it was a specific set of interfaces, which we documented.  Our commitment is that we will not break those things.

Over time, some users have pushed back against this, insisting that anything we __allow__ them to access (by making it public) should be considered as a promise to maintain that "interface." This change in terminology appears to say that we will now commit to keeping anything stable based on a technical definition of accessibility, rather than a list of classes and methods.

I'd like to see us do one of two things... or better yet both of them...

1. Not refer to "all public methods minus the internal namespace as an API.

2. Start coming up with updated, precise definitions of our __exact__ APIs - the things we support.

As an old COM guy, I was totally brainwashed with the idea that there are no "accidental features" What we support has to be defined and published. If we have failed to do that well, lets do it... using this tool if it helps... but let's __not__ define everything as an "API" while we are doing it.

At a minimum, we have three APIs, based on who uses them and for what. We should spell them out better, building on the documentation that already exists. We should not talk as if there is "an" API because that just encourages users to expect stability where we don't plan to have it.

It could well be our goal to have no public methods __except__ for those that form part of our APIs, but we are nowhere near there yet.

ChrisMaddock:
Very excited by this idea. 👍 

I'd like to think about the workflow a bit more - am I right that the current structure requires a new contributor to either know they need to generate the file, or get a CI failure? Can we alter this process, so the file is generated automatically, and the user would have to explicitly remove it, to get a CI failure?

How about:
1. File is rebuilt on a post-build step
2. Big and obvious explanation-comment at top of each file (Scenario: "This strange file keeps appearing in my change set - why?!")
3. This process is documented in the CONTIBUTING.md and BUILD.md and all over the place.

@CharliePoole - I agree we should work out what we intend to be public or otherwise. That said - after a couple of years working on NUnit, I'm still not sure of what should and shouldn't be public myself! We've also been trying to work on this since April 2016 (https://github.com/nunit/nunit/issues/1426), and haven't really gotten far - of course, in the face of other priorities.

I'd like us to get things locked down to be either public or internal - this helps me as a contributor understand which classes I'm free to change, and which may affect user's code. That helps make the project more accessible - which we all want. That said, I don't see this as something we're likely to sit down and work out class by class anytime soon - so I think it could be good to use this tool to help us with that process. 

Joseph's tool will now show us the public API changes for any PR. I'd hope that would give us a good opportunity to say "Hang on, we don't actually want those classes to be public" - and make those changes gradually over time.

ChrisMaddock:
Oh wait, sorry, I mis-read. What does "As you edit" mean - is it literally as you type, or on build?

jnm2:
@CharliePoole I'm with you there! We need to have that discussion sometime soon. I'm looking forward to it along with Chris. This PR is the first stepping stone I had in my mind toward that goal.
Yet I need to use *some* term to refer to all types and members visible outside the assembly without using eight words each time I refer to it. The term that everyone I've heard uses for this is ‘public API shape.’ I'm happy to adopt new terminology but that would that be?
Until one of us can think of a replacement term for that common one, maybe I should start saying:

 - ‘de-facto API’ for the thing that actually is the bare metal IL interface that applications are actually programing against when they reference NUnit and **which is the realm of breaking changes**, both important and unimportant
 - ’high-level API’ for the thing that transcends the IL implementation details, the intended concept which is (at times imperfectly) translated into IL

Whatever the term is, Chris said it well. This tool is specifically aimed at maintaining the IL interface, not our high-level public API. Ignoring a namespace named `Internal` is a helpful rule to cut down on noise since we know we are never going to care about most changes in there. If there are other such patterns that the tool can automatically apply to cut down even more noise, I want to add those too. But since the tool is about keeping an eye on potential unknowns in the IL interface, it comes with a blacklist on the `Internal` namespace, not a whitelist on the three high-level APIs.

jnm2:
@ChrisMaddock Good ideas!

Not as you type, but yes, as you build. VS does background builds when you open the project and at times as you type, if I'm not mistaken.

The .pubapi.cs files automatically appear and stay up-to-date every time nunit.framework.csproj is built [due to the MSBuild integration](https://github.com/jnm2/ApiContractGenerator/blob/aa3c085fdb9c978b252b630187325d6eeb504235/src/ApiContractGenerator.MSBuild/ApiContractGenerator.MSBuild.targets#L22). The only initiative that a new contributor must take is to commit any changes in these files along with the source edits that caused them when submitting a PR.

If they don't, the PR build will fail with [this message](https://github.com/nunit/nunit/pull/2652/files#diff-4d3634747586b964df9f7712bc3ec669R165):

```
Public API changes should be committed as part of the PR. Files which changed during the build:
src/NUnitFramework/pubapi/nunit.framework.net20.pubapi.cs
src/NUnitFramework/pubapi/nunit.framework.net35.pubapi.cs
src/NUnitFramework/pubapi/nunit.framework.net40.pubapi.cs
src/NUnitFramework/pubapi/nunit.framework.net45.pubapi.cs
src/NUnitFramework/pubapi/nunit.framework.netstandard1.6.pubapi.cs
src/NUnitFramework/pubapi/nunit.framework.netstandard2.0.pubapi.cs
```

If you only build with build.cake, you'll also see the same message as a warning rather than an error but only on builds which change the files again. Subsequent builds won't detect a change since the detection is source-control-agnostic. (We could improve that.) And if you stay in VS, you'll never see a warning because you aren't running NUnit's build.cake. So:

 1. Already done as you figured out from my ambiguous original explanation.

 2. Big and obvious explanation comment. I like those! Though the diff that the contributor's git tool of choice shows them will be thousands of lines down in the middle of the file. Do you think it would occur to the contributor to try to find the top of the file?

 3. Adding instructions to the build and contributing documents is a perfect suggestion which I should have thought of.

ChrisMaddock:
> Big and obvious explanation comment. I like those! Though the diff that the contributor's git tool of choice shows them will be thousands of lines down in the middle of the file. Do you think it would occur to the contributor to try to find the top of the file?

I thought of this too, but wasn't sure of a better idea. :-(

CharliePoole:
An API (Application Programming Interface) is a set of publicly accessible methods and types that are published (advertised) as supported by a particular application. Essentially, it's that stuff you guarantee will work. I believe that's the generally accepted meaning and I hope we can agree on that much.

Each application, of course, will define it's own API. NUnit has done so for the past 17 years. In NUnit V2 days, we said something like: "The API we support is what we document. Nothing else forms part of the API and nothing else is guaranteed." That was quite clear, I think. If something was reported as not working, which was not documented, we either added it to the docs or told the user it was an implementation detail.

Back in those days, however, we only had one API. The one that is directed at test writers. We still have that today and generally the same rule applies: If it's in the docs it's supported,, if it isn't it found then you don't know if it will remain stable.

One problem we had with NUnit V2, however, was that there was no API for running tests. People writing their own runners used whatever they could find in the source and complained if we changed it. We promised to resolve that with NUnit 3 and we did so. There is a program execution API, which is also documented.

The remaining issue, as far as I can see, is the API that is used by writers of custom attributes and other kinds of framework extensions. That never got documented - or only parts of it did. It's hard because such extensions may need to do some of the same things that we do internally. For example, an alternative RetryAttribute, with different repetition logic, would need to implement an interface and possibly make some changes to the test, test execution context, result, etc. The interface itself is documented but the things that they "might" want to do are not. I was scheduled at one point to do it but I'm leaving and nobody else has stepped up. IMO... this small segment of the potential API is the source of the problem... everything else is pretty well documented.

So, what's my problem with this PR? Basically, nothing as far as the basic idea goes. I am concerned about a term only. I've expressed that to @jnm2, who has several times told me he doesn't think it's important. Unfortunately, I do, so I am obligated to keep speaking up.

As written, this PR creates a new file, visible to users called `publicAPI.cs`. This enables a user to say... Oh, all this stuff is in your public API, so you have to keep it working. It even appears to be consistent with our older policy that the API is what we document and then goes ahead and documents a bunch of stuff as part of the API based on analysis of the code. But the entire premise of our past practice was that stuff you could only find by examining the code was __not__ our API.

It seems to me that it would have been very easy to use a different name - short or long, who cares? - for the file in order to satisfy my repeated complaints about the term API. That wasn't done. I'm highlighting the 2% of this PR that I don't like because I think it's a __big__ mistake and appears to reverse all of our prior policies. 

All of you guys have repeatedly said "This is just a tool" and it doesn't mean we are redefining the API. However, it __looks like__ you are and if it didn't look like it you would not need to explain that it doesn't mean what it looks like.

Frankly, I'm feeling very ignored here. I asked - months ago - for a very simple change in terminology. @jnm2 is already quite aware of everything I've said in this post and has chosen to ignore the request. I'm __not__ using my annoyance to formally disapprove the PR because I don't think that's appropriate for somebody in my position as soon to be gone. But I do feel I have to shout out that you guys are making a large mistake in changing the policy or, at least, in making it look like you are changing it. A big company like Microsoft can live with API creep, where things get locked in unintentionally because users started to make use of them. I don't think a project of our size can live that way.

CharliePoole:
Telling users they have to  commit the new "API" file is a mistake. What they have to do is figure out whether the change is acceptable. As a new user, they have to confer with the team in order to figure that out. IMO, the place to handle this is in CI. An otherwise valid change that appears to affect the so-called "API" should get an extra level of review before it is allowed to be merged, IMO.

CharliePoole:
Can this tool work with a whitelist rather than a blacklist?

There are two ways to go with defining the missing pieces of our API.

1. Assume everything public is part of the API and then back things out. This is the naive approach, as taken by many users who don't understand NUnit.

2. List those things that are currently supported and add more when justified by a user request. That's the way we have always worked in the past.

jnm2:
@CharliePoole I apologize for causing you to feel ignored. That was my mistake then. When I submitted this PR last night, I did so with you in mind thinking that you were comfortable with the terms it used. I'm obviously wrong, dropping the ball on that since our last discussion of five months ago. I am not interested in ignoring this knowledge! I appreciate your long-sufferingness with my haphazard juggling of time and notes.

Besides me forgetting that, you may have also seen me avoid dealing with that issue of determining what should or should not be considered public API. That discussion is nebulous to me and apparently controversial and as the new guy I'm going to have to lean very hard on you and Rob. It's full of convention that I'm not familiar with. I really don't think I can make headway on that on my own. This does not mean I'm pretending it doesn't matter. It does mean I don't have a plan for it–other than to get this tool in and then start bugging you and Rob to provide a proposal of how to define and document the three public APIs.

Your concerns about this PR come down to removing the word API from anything this tool does at present, right? Let's do it. Do you have a suggestion of what term to use in messages and documentation in place of ‘public API’? Should we say ‘IL interface’? It's the best I have right now but surely meaningless to a new contributor compared to the generally-used term. Maybe `external shape` or `IL contract`?

The file path of the diff-able files could be `{Project}/../externalshape/nunit.framework.net20.shape.cs`. How does that strike you?

> Telling users they have to commit the new "API" file is a mistake. What they have to do is figure out whether the change is acceptable. As a new user, they have to confer with the team in order to figure that out. IMO, the place to handle this is in CI. An otherwise valid change that appears to affect the so-called "API" should get an extra level of review before it is allowed to be merged, IMO.

I like the direction you're taking it. It's very similar to where I was trying to get us. I thought it could be difficult for the contributor to confer with the team without actually showing us a diff of some kind. Committing and pushing to the PR in progress seems the easiest to me because a readable diff is right there in the PR; a CI failure doesn't give us a diff. We'd have to pull and build locally. What do you think about that?

What if we say to commit the external shape file if you intend to change the external shape, but put [wip] in the PR title and confer with the team then? So long as CI passes, that way we're all looking at the same thing.

> Can this tool work with a whitelist rather than a blacklist? 

Not currently. The tool is aimed at better-safe-than-sorry and shows you *everything*, leaving it to you to decide if the change is in the high-level API or not, or if the change should be moved into or out of the public API. I don't really want to lose this catch-all feature. In NUnit's case, the tool is not about defining the public API. Its scope is limited to raising de facto C#-source-breaking or IL-binary-breaking changes to attention as only part of the considerations in actually maintaining NUnit's high-level API.

(Source-breaking changes are unequivocally defined by the C# language spec and binary-breaking changes are unequivocally defined by the CLR (CLI) spec. This data must inform decisions about NUnit's high-level public API though we are not bound by the data to make any particular decision about the public API.)

Would you be willing to try blacklist approach and propose a particular adjustment once we see what comes up?

> There are two ways to go with defining the missing pieces of our API.
> 
> 1. Assume everything public is part of the API and then back things out. This is the naive approach, as taken by many users who don't understand NUnit.
> 
> 2. List those things that are currently supported and add more when justified by a user request. That's the way we have always worked in the past.

So to reiterate: this tool is not concerned with *defining* any part of NUnit's high-level public API. Whether we go with the first or second listed way to define the API, this tool which operates at a much lower and more limited level should not be affected by that decision. According to my understanding.

CharliePoole:
@jnm2 First a brief clarification and apology... I can see from re-reading my post that you might conclude I was blaming you for my feeling of being ignored. I'm sorry as that was not what I meant. The feeling applies to the entire team and is one of the main things that moved me from envisioning a reduced participation in NUnit development to leaving entirely. As the newest core team guy, this negative environment can't be blamed on you and it wasn't my intention to do so. I won't say any more about this in this public forum but I wanted to say at least that much.

That said... I thought I made it pretty clear in the past that all you needed to do was stop talking about an "API" in order to have my 100% support for what you are doing. I was surprised you didn't do it, since it seemed like something that ought to be rather unimportant from the point of view of the PR while it was very important to me. Accommodating one another like that is how we get along.

I won't repeat my earlier long note but will only add that you could use the term API any way you wanted, in my view, __if NUnit had never used it in a specific way__ in the past. But having used it, any change you make in how it's used now looks like you are making a policy change and policy changes shouldn't be made accidentally. This is particularly important since we have previously committed to not changing things that break an "API"! Of course, I'm the guy who made these commitments in the past, so it's OK to blame me if you don't like them. However, I think you should meet them anyway for the sake of the users who believed in them!

Of all the names you have come up with, I like "shape" best because it has no pre-existing context. The more "meaningful" names like "IL" come with prior interpretations. Just as companies try to choose names without baggage attached for reasons of trademark law, we should choose a name for this file and talk about it in a way that doesn't invite users to make their own interpretations. If folks dislike "shape", then I'd suggest "surface" or even "visible-surface".

BTW, I did already understand that you don't intend to be __defining__ the NUnit API. You made that clear months ago, which is why I supported this tool. I hope you can see my point that by creating and publishing (in the repo) a `publicAPI.cs` file, you do __appear__ to be redefining it from the outside.

CharliePoole:
Aside from my major (naming) issue, I agree with @ChrisMaddock that defining the process is the most important part of this. Should we spend some time on that? What should a team-member do? What should a contributor do (if anything)?

CharliePoole:
@jnm2 Re blacklisting vs whitelisting...

Blacklisting will give you false negatives (i.e. you should not change this because it's in the public API)

Whitelisting will give you false positives until we are sure the whitelist is complete (i.e. you might change something that users are depending on because it's not whitelisted.)

In either case, we need a way to make changes to the output of the app, don't we? Or are you envisioning that the only way to make a change is to change the visibility of the code?

jnm2:
I added a commit which updates the PR to use the term ‘external shape’ rather than ‘public API.’

[Here is how you configure the output path](https://github.com/nunit/nunit/pull/2652/files#diff-d659fe06375fffbb7706a8687de4f407R24), by the way (ApiContractGenerator docs are a TODO). I picked `{Project}\..\externalshape\nunit.framework.net20.shape.cs` but it's easy to customize if anyone has a suggestion.

jnm2:
> Accommodating one another like that is how we get along.

@CharliePoole Thanks for your comments. I'm with you there. I feel bad about using the term; can't believe I forgot something so obvious. Appreciate the good will.

Yes, let's define the process. I'll start on a draft for CONTRIBUTING.md.

I'm not sure if I was using the terms blacklist and whitelist the same way. I want this tool to err on the side of detecting too much rather than not detecting enough. I'd rather provide ways to manage knowns (e.g. ignore namespace X) than be blind to unknowns by default. We can easily direct a contributor to add `;NUnit.Framework.NewNamespaceToIgnore` to [this line](https://github.com/nunit/nunit/pull/2652/files#diff-d659fe06375fffbb7706a8687de4f407R25) if we realize that it will be annoying to be notified about external shape changes in that namespace.

jnm2:
@ChrisMaddock Right now there is no capacity for ApiContractGenerator to add arbitrary comments at the top, but it would be very easy. Another option would be adding a `readme.txt` in the folder with them which points to CONTRIBUTING.md. I'm ambivalent. Which direction should I go?

ChrisMaddock:
Between comment and readme I’m fairly apathetic. Both? 😛

Been thinking about this a bit more - I’m still not sure I’m comfortable of the workflow of the user needing to commit an autogenerated file with their changes. My main concern I think is that it’s just different from normal convention. If I was to contribute to a new project today, went to stage my changes, and saw an unrecognised file there - my instinct would just be to revert it. (That happens quite often imo, e.g. things that should be gitignored, Rebracer/ReSharper ‘upgrading’ settings files...) I’d then PR, get a CI failure, and then have to go back and work out how to regenerate the file. Fine if you know the deal, but I’m a little concerned about the new contributor experience.

A question, what do we lose if we do this entirely within CI, and don’t commit anything to the repo? Cake could build master and the PR, generate and diff the files? 

That would rely on reviewers looking for the CI report - but I think that’s still an improvement on our current review process of hoping we notice these changes.

An even better solution could be to have a bot post the diff on each new PR - but I imagine that’s a bit more work!!

CharliePoole:
With the change in terminology, I'm now with Chris on this - I think it all depends on the process we set up.

I'm thinking we should somehow distinguish between contributors and team members. The latter can be expected to understand the tradeoffs and come up with a recommendation that something either be changed in the visible app surface or not. That seems to be asking too much of casual contributors. I feel that they should just be asked to refer the issue to a team member for resolution if it comes up.

Doing it this way means that branch builds should pass even if they result in a change, whereas master builds after merge should fail. I'm not sure which way PR builds should go.

jnm2:
> A question, what do we lose if we do this entirely within CI, and don’t commit anything to the repo? Cake could build master and the PR, generate and diff the files?

We lose the ability to see and comment on a nice diff in the PR. And ApiContractGenerator.MSBuild would no longer be source-control-agnostic.

Personally I'd prefer a bot that noticed the .shape.cs change and tagged the PR and left an explanation comment. The simplicity of the shape file appeals strongly to me. No one has to run a special tool to see the diff between commits and branches; everyone has it built into their favorite editor. But I am hearing you about the new contributor experience.

Given that I am not able to justify creating a bot right now, the choice before us seems to be between having a diff of the shape in with the PR for comments, and saving new contributors (who changed the external shape and didn't notice the CONTRIBUTING section) from having to notice the CI failure which points them to CONTRIBUTING and asks them to push one more commit.

Let's say we don't have the PR build fail. A new contributor adds a helper method in a medium-sized PR. None of us notices an actual API change. We approve and merge. The PR closes. CI starts building the merge commit.
Suddenly the master branch is failing and every subsequent PR fails as well. The only remedy is to pull master and submit a second PR to correct the shape file. I don't think this gains us anything.
Or let's say we don't have any builds fail. The shape files will continually desync with the source history. This doesn't gain us anything either.

Or let's say the shape file is not committed to source control at all. Then the tool needs to be redesigned to be able to read commits in the .git database which are not checked out on disk. We either need a bot or this tool becomes an on-demand tool rather than a way to keep us from missing potential breaking changes. Maybe it becomes part of Rob's release process. I don't like this; the feedback loop should be much closer to the actual change. If we don't look at the warnings before the PR is merged, the effectual procrastination will make any correction much harder to deal with a month or four later.

I returned to gaming out these thoughts over and over five months ago, so the picture is clear in my head. That doesn't mean it's the right picture for NUnit though. Taking no action is a venerated option. 😜

CharliePoole:
I would definitely not want to tell new contributors to commit the changed shapes file. I'd rather tell them to request feedback about __whether__ to commit the changes or not make the change at all.

I see what you are saying about possibly not realizing there was a problem till after the commit. What if we only warned in the branch build but failed the PR build? That would give the team member working with the new contributor a warning while also giving the contributor the benefit of at least one build that passes.

CharliePoole:
BTW, just in case we are not on the same wavelength, I'm assuming that every break in the visible surface interface will lead to some extended discussion in the PR or issue in question. Otherwise, it doesn't make much sense to do this.

jnm2:
> I would definitely not want to tell new contributors to commit the changed shapes file. I'd rather tell them to request feedback about whether to commit the changes or not make the change at all.

When they do request feedback, wouldn't it be nicer to discuss and comment on the .shape.cs diff in the PR? I'm not sure yet what we gain on the other side if we avoid that. Or are you saying they shouldn't commit anything (even the source code) until we discuss it, to save them time? But that also costs them time. They have to be sure they clearly communicate the details of what it was they were about do to without sharing an actual diff with us. Minor inefficiency on both sides.

> What if we only warned in the branch build but failed the PR build? That would give the team member working with the new contributor a warning while also giving the contributor the benefit of at least one build that passes.

Sure, I like this. As for the regular team members' branch builds, we'll all know what the .shape.cs changes mean and won't need the branch build to fail in order to stay on top of it.

> BTW, just in case we are not on the same wavelength, I'm assuming that every break in the visible surface interface will lead to some extended discussion in the PR or issue in question. Otherwise, it doesn't make much sense to do this.

Yes! 👍

ChrisMaddock:
> Or let's say the shape file is not committed to source control at all. Then the tool needs to be redesigned to be able to read commits in the .git database which are not checked out on disk. 

Can you explain what I'm missing here? Would it not be possible to checkout the repo, and build and generate the file - or am I missing how this works? 

I agree we want this as part of the PR process - the point of release is far too late, and our releases are time consuming enough as it is. 

One last thought - I suppose we wouldn't actually need a bot - we could fairly trivially have CI run the diff and post it back to the PR as a comment, I believe. So, for example, - the tool diff as a Cake task. Experienced contributors could use it as a development tool, run the diff to review their change before push. New contributors wouldn't know it was there. For all PRs, Cake could run the diff, and post the change back as a comment - as Joseph has been doing for a few PRs recently. We would however lose the line-by-line commenting option.

I'm interested to hear what other people think about an autogenerated file in the repo, which needs committing by all contributors. Maybe I'm over-worrying about how new contributors would handle it.

CharliePoole:
Suppose we just viewed this as a separate test type from unit tests, analogous to how acceptance tests are usually treated. This particular test is of the "golden master" type, with the master being the already committed version of the shape file.

The "test" generates a new file, which is supposed to match the master. If it doesn't match, that's a failure or warning as the case may be. However, the user doesn't commit the changed file. Rather we discuss the failure and decide whether it's a "Real" failure or if we want to change the shape file. That change could be a __true__ change (reflecting the newly generated file) or an exclusion (removing the changed entry from the master entirely. (I understand we don't yet have a mechanism for that, but we could create one.)

The decision as to whether to change the master is a team decision, obviously.

Would that work? Could the generation be run as a separate target in the command-line build rather than as a part of the VS build?

sharwell:
I did not read the entire thread, but another option is the PublicApiAnalyzer. This analyzer is build-enforced, and allows for the separation into "shipped" (stable) and "unshipped" (public but not stable) APIs. This approach is used by dotnet/roslyn, and I've started to use it for a few other cases as well. Let me know if you are interested in seeing how it would apply here.

jnm2:
@sharwell Things we hope to put emphasis on detecting are class modifiers, AttributeUsage or Obsolete changes, namespace definitions/removals, etc. The goal is that anything that has the potential to cause a source-level or binary-level break should show up in the diff. That way we can consciously decide whether each break is acceptable. (In the future, I want to make the tool generate sample code that would demonstrate the break. Right now I'm doing this manually, when I notice it.)

This tool also affords us the ability to ignore the `NUnit.Framework.Internal` namespace except for what can cause breaking changes by being transitively referenced from outside that namespace.

All—I *am* coming back to this after I'm finished helping with the 3.10 adapter work.

jnm2:
> One last thought - I suppose we wouldn't actually need a bot - we could fairly trivially have CI run the diff and post it back to the PR as a comment, I believe. So, for example, - the tool diffs as a Cake task. Experienced contributors could run it manually as a development tool, run the diff to review their change before push. New contributors wouldn't know it was there. For all PRs, Appveyor could run the diff, and post the change back as a comment - as Joseph has been doing for a few PRs recently. We would however lose the line-by-line commenting option.

The more I think about it, the more I like this. Writing a small service would enable a phenomenal experience. Do you know of a good way to host a service like this?

For myself, I would still want the local git diff to update as the project builds. Having used the local-file-changes method for a while at work and in other projects, I am still enjoying that too much.
I wonder how easy opt-in/out could be. MSBuild takes environment variables as project properties. The Cake script could have a target to opt in or out by setting a user-level environment variable.

ChrisMaddock:
> The more I think about it, the more I like this. Writing a small service would enable a phenomenal experience. Do you know of a good way to host a service like this?

I was meaning we could have it posted back by the Appveyor build. 🙂 

> I wonder how easy opt-in/out could be. MSBuild takes environment variables as project properties. The Cake script could have a target to opt in or out by setting a user-level environment variable.

Sounds good to me!

jnm2:
@ChrisMaddock Whoa... I didn't think of that! Do you know of something like this in action that I can look at?

ChrisMaddock:
Nope - I wanted to write one to maintain our samples before, but never got round to it. 👎

I was approaching it using Octokit. I even started on a [Cake Addin](https://github.com/ChrisMaddock/Cake.GitHub). 😆 It never got off the ground really - mainly because the Octokit API was clean enough that it hardly seemed worth it. Like all good side projects, I spent time making the logo, and never wrote the code...

I imagine it wouldn't be too difficult to store an Octokit access key in Appveyor, and use Octokit to comment back on the PR. I remember posting an issue being fairly trivial (see the addin code) - I don't imagine it would be much more to post comments to a PR.

We'd need an NUnitBot account! 😄

ChrisMaddock:
Oh! Looks like Appveyor even has a thing to do this directly!

https://www.appveyor.com/docs/notifications/#github-pull-request

jnm2:
This (unexpected) contributor seems to have figured it out without me documenting or saying anything: https://github.com/OmerMor/AsyncBridge/pull/16/files#diff-5ab2dbc87b8ac8102e4140f6efa1ce00

jnm2:
Not sure I have the time to finish this. Let's look into PublicApiAnalyzer in the future.

PR #2655, opened 2018-01-13
Use automatic documentation file path

jnm2:
Just a csproj simplification. I just noticed they finally added support for this, who knows how long ago now!


jnm2:
Thanks for providing a second verification.

Btw, I checked and this has been supported since VS 2017's initial release. `https://github.com/dotnet/sdk/pull/349`

PR #2657, opened 2018-01-13
Adding testname attribute in test case xml

Shyam-Gupta:
If TestName attribute is not null or empty then add it in test case attributes. This value will be read by nunit VS adapter. If value of this attribute is present then a test property with name 'AdjustedFQN' will get added in TestCase object. This will help Test Window in determining correct source information for the test method having TestName attribute.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Tests/Test.cs
+++ b/src/NUnitFramework/framework/Internal/Tests/Test.cs
@@ -182,6 +182,15 @@ namespace NUnit.Framework.Internal
         }
 
         /// <summary>
+        /// Gets the value associated with TestName attribute of the test.
+        /// Returns null if the test is not implemented as a method.
+        /// </summary>
+        public virtual string TestName
+        {
+            get { return null; }
+        }
+
+        /// <summary>
         /// The arguments to use in creating the test or empty array if none required.
         /// </summary>
         public abstract object[] Arguments { get; }
@@ -392,6 +401,8 @@ namespace NUnit.Framework.Internal
             if (this.ClassName != null)
                 thisNode.AddAttribute("classname", this.ClassName);
             thisNode.AddAttribute("runstate", this.RunState.ToString());
+            if(this.TestName != null)
+                thisNode.AddAttribute("testname", this.TestName);
 
             if (Properties.Keys.Count > 0)
                 Properties.AddToXml(thisNode, recursive);
--- a/src/NUnitFramework/framework/Internal/Tests/TestMethod.cs
+++ b/src/NUnitFramework/framework/Internal/Tests/TestMethod.cs
@@ -121,6 +121,11 @@ namespace NUnit.Framework.Internal
 
             PopulateTestNode(thisNode, recursive);
 
+            if (!string.IsNullOrEmpty(this.parms.TestName))
+            {
+                thisNode.AddAttribute("TestName", this.parms.TestName);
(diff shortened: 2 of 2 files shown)


CharliePoole:
This makes no sense at all. The value of `parms.TestName` is internal and contains special codes. It's external effect is shown in the __actual__ test name, which is exposed publicly but exposing this is unnecessary and dangerous.

If I understand correctly, the desire is to know whether the user has over-ridden the default naming used by NUnit. This is fragile because new ways to do the same thing may be added at any point. However, if it were really needed, then we already have a way to pass the information through properties.

My feeling is that it's not needed based on what I know of the framework and the adapter. If the user wants this info, it should be justified in a framework issue and, if accepted, a team member with knowledge of framework internals should actually implement it.

CharliePoole:
I should add: I do not think we should put special fields in the framework to help a particular vendor unless there is a more general need for the feature. If there is (and I don't think that has been demonstrated here) it should be implemented in a general way.

I see no reason why we need to expose the fact that the user typed "MyTestName" or "{m}_WithSpecialArg" as the test name. That's implementation detail, which could change at any time.

Shyam-Gupta:
At present, the test case xml does not have information about whether user has overridden the test method name specified in the code. Rather the fully qualified name of the test case has the overridden name which seems incorrect as it should be what is specified in the code. The DisplayName field can be set to what user wants it to be. Due to this missing information Test Window is not able to find source information for the test method. (refer: [bug 423](https://github.com/nunit/nunit3-vs-adapter/issues/423)

jnm2:
Instead of doing this, can the adapter just always send the fully qualified method name according to the standard CLR spec? `Namespace.GenericType~1+NestedType.SomeMethod~3(T4, T1, System.String)` 
That would be generally useful for anyone who wants to associate a test with source code.

Whatever it is that uniquely connects a test case with source information, I'd prefer it not to be affected by whether TestName is overridden (or my `ArgDisplayNames` which is ready to merge...)

jnm2:
If that doesn't meet your goal, can you help me understand what the goal is since my head isn't in the game yet?

CharliePoole:
@jnm2 Yes! Your `ArgDisplayNames` change is precisely the sort of thing I was thinking of as something that "could change at any time"

On nunit/nunit3-vs-adapter#423, I already proposed something similar to what you suggested, although the exact details are different. I think the key is for the adapter to always do the same thing - no matter what the user has done. Rather than having an "alternate" method when the user has modified the display name, make that "alternate" the only way of doing it.

As I suggested on nunit/nunit3-vs-adapter#423, all the needed info is not yet available to give to VS, but we could __make it available__ if the MS folks would take the time to specify it.

Some discussion along these lines was started with one of the MS PMs, but nothing came of it. IF we are going to add more info to the test, a complete and proper name, such as you suggest makes sense, either as one thing or in parts. Right now, we give a class name and method name. The class name contains the type parameters and actual arguments of the class. The method name, however, is bare and we might have to enhance that. Personally, since it's currently only provided for use in finding the source code, I wouldn't have a problem with that but if anybody else saw a problem we could just add a field.

Perhaps the discussion on this needs to go back to nunit/nunit3-vs-adapter#423.

Shyam-Gupta:
@CharliePoole We are closing this PR for time being. We will think through on the information required by VS and will come up with a more cleaner approach soon.

Issue #2662, opened 2018-01-16
NullReferenceException after parallel tests have finished executing


sanoursa:
Using NUnit 3.9.0, we have enabled `Parallelizable(ParallelScope.Fixtures)`, and while the tests all run correctly, about 30% of the time the test suite fails because NUnit's background thread throws a NullReferenceException with the following stack:

```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object. at 
NUnit.Framework.Internal.Execution.WorkItemQueue.get_IsEmpty() at 
NUnit.Framework.Internal.Execution.WorkShift.get_HasWork() at 
NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.SelectNextShift() at 
NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.OnEndOfShift(WorkShift endingShift) at 
NUnit.Framework.Internal.Execution.WorkShift.<StartWorkers>b__26_1(TestWorker s, WorkItem ea) at 
NUnit.Framework.Internal.Execution.TestWorker.TestWorkerThreadProc() at 
System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx) at 
System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx) at 
System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state) at 
System.Threading.ThreadHelper.ThreadStart() 
```

Is this a known issue? Is there anything I can do to gather more information to help diagnose the problem?

jnm2:
Huh, thought that stack trace was one of the two fixed in 3.9 by https://github.com/nunit/nunit/issues/2428.

sanoursa:
@CharliePoole I tried the latest package from the dev feed, and still hit the same exception. Is there any extra logging that I can enable to help diagnose this?

Also I've noticed that I hit the exception very reliably on a machine with 4 cores, and very infrequently on a machine with 2 cores.

CharliePoole:
It's tricky because we end up with a list of queues that includes null values. That causes an exception much later but I can't see how it happens.

sanoursa:
I'm still trying to understand the code, but is it possible that there's a race between WorkItemQueue.Save() calling InitializeQueues(), and the background thread calling IsEmpty?

[...]

PR #2668, opened 2018-01-18
Fully initialize the new array of queues before assigning it

sanoursa:
Fixes issue #2662.

Without this fix, the issue repros about 80% of the time on a 4 proc machine. With this fix, I have now done over 20 test runs, and all succeeded.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
@@ -135,11 +135,12 @@ namespace NUnit.Framework.Internal.Execution
 
         private void InitializeQueues()
         {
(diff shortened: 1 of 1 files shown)


src/NUnitFramework/framework/Internal/Execution/WorkItemQueue.cs
```diff
@@ -135,11 +135,12 @@ public WorkItemQueue(string name, bool isParallel)
 
         private void InitializeQueues()
         {
-            _innerQueues = new ConcurrentQueue<WorkItem>[PRIORITY_LEVELS];
+            ConcurrentQueue<WorkItem>[] newQueues = new ConcurrentQueue<WorkItem>[PRIORITY_LEVELS];
```
CharliePoole: I'd typically use `var` here but I won't ask for a change.

PR #2671, opened 2018-01-20
Ensure that FailureSite.Child is used where appropriate.

### LINKED ISSUE #2670: Invalid assemblies no longer give an error message
@CharliePoole commented on [Tue Jan 16 2018](https://github.com/nunit/nunit-console/issues/350)

In the NUnit3 console runner 3.7, the argument `x.dll` gives an error message indicating the file was not found. Typing `x.junk` gives a message saying that the file type is not known.

In the current master build, the run fails without specifying a message in both cases.

---

@CharliePoole commented on [Thu Jan 18 2018](https://github.com/nunit/nunit-console/issues/350#issuecomment-358854571)

As discussed in PR #177, assembly-level errors like file not found or invalid file type stopped displaying in the console runner with the merging of #326. This is due to the fact that the assembly level normally does not have a site associated with it.

An easy fix was tried but has the drawback that it causes all suites containing a test with a warning, either directly or indirectly, to display as well. That problem is due to the fact that warnings never have a site associated with them either.

This will be a two part fix. The framework has to be updated so that a warning result caused by a child test uses `FailureSite.Child` In fact, it's essential that any failure or warning of a suite caused by a child test use that value explicitly.

However, even when the framework is updated, the console has to be able to deal with older framework versions that do not set `FailureSite` correctly. That's what the current issue has to do. Specifically, this issue needs to...

1. Ensure that assembly errors, failures and warnings are shown in the errors, failures and warnings report

2. Ensure that warnings are also shown in the report __only once__, that is, not repeated for the suites that contain them

I'm replicating this same issue to the framework repository for the work that is needed there.

CharliePoole:
Fixes #2670 

This PR makes sure that a suite with a Warning result due to a child test is always given `FailureSite.Child` just as is done for Failed results. By doing this, we allow runners to display the warning where it occurs without needing to parser the wording of the message.

### DIFF
--- a/src/NUnitFramework/framework/Interfaces/ResultState.cs
+++ b/src/NUnitFramework/framework/Interfaces/ResultState.cs
@@ -135,7 +135,7 @@ namespace NUnit.Framework.Interfaces
         public readonly static ResultState ChildFailure = ResultState.Failure.WithSite(FailureSite.Child);
 
         /// <summary>
-        /// A suite failed because one or more child tests failed or had errors
+        /// A suite failed because one or more child tests had warnings
         /// </summary>
         public readonly static ResultState ChildWarning = ResultState.Warning.WithSite(FailureSite.Child);
 
--- a/src/NUnitFramework/framework/Interfaces/ResultState.cs
+++ b/src/NUnitFramework/framework/Interfaces/ResultState.cs
@@ -135,6 +135,11 @@ namespace NUnit.Framework.Interfaces
         public readonly static ResultState ChildFailure = ResultState.Failure.WithSite(FailureSite.Child);
 
         /// <summary>
+        /// A suite failed because one or more child tests failed or had errors
+        /// </summary>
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/framework/Interfaces/ResultState.cs
```diff
@@ -134,6 +134,11 @@ public ResultState(TestStatus status, string label, FailureSite site)
         /// </summary>
         public readonly static ResultState ChildFailure = ResultState.Failure.WithSite(FailureSite.Child);
 
+        /// <summary>
+        /// A suite failed because one or more child tests failed or had errors
```
mikkelbu: Shouldn't we change this comment, so that it mentions warning? E.g. 
`/// A suite failed because one or more child tests had warnings`
Currently, it is just the same comment as for `ChildFailure`

CharliePoole: Fixed

PR #2672, opened 2018-01-20
Enable easy NUnit source-stepping with GitLinked PDBs in main NuGet package

### LINKED ISSUE #2421: Publishing symbols with releases
We should provide debug symbols with every release so that people can step into the source code, set breakpoints, view variables, etc.

Can we start publishing PDBs with our NuGet package? We could make it part of our Cake script to index the PDBs with https://github.com/GitTools/GitLink so that the source code is pulled from GitHub when people step into the framework dll.

https://oren.codes/2015/09/23/enabling-source-code-debugging-for-your-nuget-packages-with-gitlink/

jnm2:
Closes #2421.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ NUnit is a unit-testing framework for all .NET languages. Initially ported from 
 - [Downloads](#downloads)
 - [Documentation](#documentation)
 - [Contributing](#contributing)
+- [Debugger source-stepping](#debugger-source-stepping)
 - [License](#license)
 - [NUnit Projects](#nunit-projects)
 
@@ -34,6 +35,12 @@ For more information on contributing to the NUnit project, please see [CONTRIBUT
 NUnit 3.0 was created by [Charlie Poole](https://github.com/CharliePoole), [Rob Prouse](https://github.com/rprouse), [Simone Busoli](https://github.com/simoneb), [Neil Colvin](https://github.com/oznetmaster) and numerous community contributors. A complete list of contributors since NUnit migrated to GitHub can be [found on GitHub](https://github.com/nunit/nunit/graphs/contributors).
 
 Earlier versions of NUnit were developed by Charlie Poole, James W. Newkirk, Alexei A. Vorontsov, Michael C. Two and Philip A. Craig.
+
+## Debugger source-stepping ##
+
+The NUnit and NUnitLite NuGet packages contain a source-indexed PDB for each binary from this repository.
+If you’re in the middle of a debugging session and realize you’d like to be able to step into NUnit code,
+set breakpoints and watch variables, [follow these steps](https://github.com/nunit/docs/wiki/Debugger-Source-Stepping).
 
 ## License ##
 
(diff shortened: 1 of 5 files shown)


build.cake
```diff
@@ -313,48 +319,49 @@ Task("CreateImage")
         }
     });
 
-Task("PackageFramework")
-    .Description("Creates NuGet packages of the framework")
+#tool GitLink
```
ChrisMaddock: How about moving `#tool` up top with the other `#tool`? Or moving the NUnit.Console `#tool` down to the test target - so the two are consistent. (But my personal style preference would be the former. 😄)

DEBUGGING.md
```diff
@@ -0,0 +1,45 @@
+# Debugging NUnit
```
ChrisMaddock: I love that you've added this file! 😄 One question, do you think it maybe belongs in the docs repo perhaps, rather than in the main repo root? Two reasons:
1. Root folder size.
2. This can be made generic if this goes well, and we add the same feature to the console.

README.md
```diff
@@ -9,6 +9,7 @@ NUnit is a unit-testing framework for all .NET languages. Initially ported from
 ## Table of Contents ##
 
 - [Downloads](#downloads)
+- [Debugger source-stepping](#debugger-source-stepping)
```
ChrisMaddock: I like this, but disagree with listing it above `Documentation` - in terms of user-priority. How about moving it to between contributing and licence? (Would also be nice in that this is very relevant to people thinking of contributing.)

jnm2:
Important comment 😃: I've also tested this in VS2017 and confirms that it works, via NuGet package and following DEBUGGING.md, without having any NUnit source on your machine.

DEBUGGING.md
jnm2: I would love to! Should that be done before or after this PR?

build.cake
jnm2: @nunit/framework-team Looks like we have two opposite personal preferences and no strong opinions here– should I move it to the top?

CharliePoole: Top feels right to me. Didn't see any competing comment.

jnm2: :+1: The opposite personal preference is showcased in the PR 😁

DEBUGGING.md
ChrisMaddock: Sorry - I meant in the docs **instead** of the nunit repo.

We only have key contribution docs at the moment in the repo root. This is a great doc - but I think it should be linked from the readme to docs, rather than living in the repo root itself - I don't see it as being on the same level as README, BUILDING and CONTRIBUTING. (Although a link from contributing as well might be a good idea!)

jnm2: I think I understood– you meant, move it to https://github.com/nunit/docs/wiki? Should I wait till this PR is approved and then move it, or let this PR be merged, then move to the wiki, then remove from this repo with a second PR?

ChrisMaddock: Ah, sorry - I thought you'd misunderstood, as I didn't see why you'd want to merge it, and then remove it! But I guess you're thinking of review purposes. 😄

Whatever's easier. Consider the doc reviewed by me! 👍

jnm2:
Review feedback has been addressed. Ready for final review!

Issue #2659, opened 2018-01-14
Naming Errors


CharliePoole:
I'm getting "Naming Errors" in my build of the framework. I don't know of any time when we all agreed that Visual Studio would be set up to do that. I assume it was done as a PR, which only requires two team members to review it, but I think stuff that constrains us all to work in a certain way requires either unanimity or a directive from the project lead. At a minimum, I think it requires discussion.

mikkelbu:
My guess is that the "Naming Errors" are due to #2320 (but it is only a guess as I'm not quite what Naming Errors cover). As far as I can remember the rules defined in the editorconfig file should just be the standards from https://github.com/nunit/docs/wiki/Coding-Standards.

CharliePoole:
Right. But up to now, as far as I was aware, the guidelines in editor config were not enforced in any way.  In fact, that lack of enforcement is what convinced me it was OK to do. But I'm seeing errors right in the VS error window along with compile errors, although they don't seem to show up on every compile. I'd appreciate if somebody can explain how this happened.

My own view has always been that having rules enforced by software is an anti-pattern in any team.

CharliePoole:
You're right about #2329, which fixed issue #2319.

The issue was originally written to enforce byte order marks. It seemed like a good idea and I for one didn't give it any further attention. Rob added a comment saying he wanted the coding standards to be enforced to the extent that they were in the editor config and I guess that was done.

In our governance structure, the project lead has the power to set rules like that. I have no qualms about it. He can decide whether to take advice from others or just make his own choices. At a minimum, however, I think the lead ought to announce the rules he is setting, not slip them in through a software fix.

It's that kind of slipstream promulgation of rules that makes me say automated rule enforcement is an antipattern in most teams.

I'm closing this issue as won't fix and chalking it up to experience.

rprouse:
@CharliePoole you can hide the `editorconfig` warnings from the error window by switching the dropdown to `build only`

![image](https://user-images.githubusercontent.com/493828/34922618-555ce200-f960-11e7-943d-f6a0f6bea706.png)

I do that for every legacy solution, VS2017 added way too much noise in intellisense warnings.

CharliePoole:
When it lists it shows these things as errors, not warnings. I guess I'll wait to see if my CI builds can pass. Note that if the errors lead contributors to make wider changes in the source, they will be doing something we have always asked that they __not__ do as part of a PR.

rprouse:
[...]

PR #2678, opened 2018-01-24
Changed .editorconfig errors to warnings

jnm2:
Fixes https://github.com/nunit/nunit/issues/2659

Doesn't look like the VS folks put much time towards .editorconfig since the last time we looked, but they are making up for it in other areas IMO so I'm just fine with that.


jnm2:
@ChrisMaddock Done.

ChrisMaddock:
Thanks! 😄

Issue #1944, opened 2016-12-29
Removing Compact Framework workarounds


jnm2:
Browsing through the code, I noticed some references to compromises made in order to support CF, or documentation of CF quirks:

- [x] <s>[/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs#L260](https://github.com/nunit/nunit/blob/d01935d8620f8f125bd1a77f89cb1fc8fd370910/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs#L260)</s>

   > [@CharliePoole] Just a comment - somebody wanted to use methods that are not in CF - note that "original" in this context doesn't mean original NUnit code, just that person's original idea of how to code. No change except maybe the useless comment.

- [x] [/src/NUnitFramework/framework/Interfaces/ITestResult.cs#L95](https://github.com/nunit/nunit/blob/d01935d8620f8f125bd1a77f89cb1fc8fd370910/src/NUnitFramework/framework/Interfaces/ITestResult.cs#L95)

   > [@CharliePoole] Useless comment - we haven't supported CF 1.0 for MANY years.

- [x] <s>[/src/NUnitFramework/framework/Internal/Reflect.cs#L162](https://github.com/nunit/nunit/blob/d01935d8620f8f125bd1a77f89cb1fc8fd370910/src/NUnitFramework/framework/Internal/Reflect.cs#L162)</s>

   > [@CharliePoole] Another comment. This one could be acted on by changing the code when somebody is actually working in the file.

- [ ] <s>[/src/NUnitFramework/framework/Internal/StackFilter.cs#L96](https://github.com/nunit/nunit/blob/d01935d8620f8f125bd1a77f89cb1fc8fd370910/src/NUnitFramework/framework/Internal/StackFilter.cs#L96)</s>

   > [@CharliePoole] I think the comment may be wrong and that other platforms may need that code. It was probably true when written.

- [x] <s>[/src/NUnitFramework/framework/Internal/TestNameGenerator.cs#L165](https://github.com/nunit/nunit/blob/d01935d8620f8f125bd1a77f89cb1fc8fd370910/src/NUnitFramework/framework/Internal/TestNameGenerator.cs#L165)</s>

   > [@CharliePoole] This would be a reasonable change if we know all platforms support TryParse. No sense changing when we aren't working in the file though.

[...]

PR #2679, opened 2018-01-24
Remove outdated Compact Framework comments

jnm2:
Fixes some of #1944.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
@@ -73,8 +73,6 @@ namespace NUnit.Framework.Constraints.Comparers
 
         private static IList<KeyValuePair<Type, MethodInfo>> GetEquatableGenericArguments(Type type)
         {
-            // NOTE: Original implementation used Array.ConvertAll and
-            // Array.FindAll, which don't exist in the compact framework.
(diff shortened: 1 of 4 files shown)


src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
```diff
@@ -73,8 +73,6 @@ private static MethodInfo FirstImplementsIEquatableOfSecond(Type first, Type sec
 
         private static IList<KeyValuePair<Type, MethodInfo>> GetEquatableGenericArguments(Type type)
         {
-            // NOTE: Original implementation used Array.ConvertAll and
-            // Array.FindAll, which don't exist in the compact framework.
```
CharliePoole: I originally left this in just to remind us in case we wanted to go back to the original implementation. Just saying... I'm OK with removing it.

jnm2: :+1: I like this more verbose code because perf is better without using delegates.

jnm2:
I apologize, @CharliePoole. I forgot the [comment in Reflect.cs](https://github.com/nunit/nunit/pull/2679/files#diff-15b9851c9669e69061a6267c56c44569). (It would have dismissed your review whether I amended or not.)

jnm2:
CI failed and I discovered that netstandard1.6 didn't resolve the xml cref. Fixed.

Issue #2685, opened 2018-01-25
Remove Resharper .DotSettings and Rebracer file in favour of .editorconfig


ChrisMaddock:
We currently have three different files providing formatting guidelines - which I don't believe are all maintained.

The most recent work has been done in .editorconfig, I believe. Do we have any need to maintain all three files, or could we remove the Rebracer and ReSharper options?

I'd like to apply the same change to both the nunit and nunit-console repo's, if we can make a decision.

rprouse:
Agreed. Rebracer isn't needed in VS 2017 which is now required to build so can be dropped. We also haven't updated Resharper in years other than your automatic changes.

jnm2:
I was just looking at that. I think we should remove them but be open to adding *minimal* ReSharper settings back in to quiet down ReSharper's default complaints in our codebase for things that .editorconfig doesn't control yet.

ChrisMaddock:
That sounds like a good idea to me. 😄

zastrowm:
Note that ReSharper also has support for configuring it via .editorconfig: https://www.jetbrains.com/help/resharper/EditorConfig_Index.html

ChrisMaddock:
Thanks @zastrowm!

Looks like we've made a decision. Should anyone new to contributing fancy picking this up, to clarify, the task is to remove the Rebracer.xml and NUnit.sln.DotSettings files from the repository. Pull Requests Welcome!

pandualuvala:
Can i take this task please let me know. Thanks

ChrisMaddock:
Yes please @pandualuvala! 😃 let us know if we need to clarify anything.

@nunit-team: I can’t remove the helpwanted label on my phone - would appreciate if someone else could now this issue is claimed.

OsirisTerje:
@ChrisMaddock I think it is wise to keep the resharper file there, with "the minimal settings to quiet it down", as you mentioned above.  Otherwise anyone using resharper will very soon do refactorings based on those complaints to the code base, and you will have very intensive pull requests. 
(cc: @pandualuvala )

ChrisMaddock:
My concern was that the resharper settings are probably out of sync with the editorconfig settings. If we think that we still need them, then I guess we should go through and work out what we can remove, and what’s being delegated to editorconfig.

@pand - ok, in that case, can you just remove Rebracer.xml for now? Looks like we’ll need to look at the ReSharper file in more detail. 🙂

pandualuvala:
[...]

PR #2687, opened 2018-01-26
Removed not used rebracer.xml file settings

pandualuvala:
removed the Rebracer.xml settings file based on issue #2685


rprouse:
Looks good, but can you remove the `Nunit.dotsettings` file also? Thanks.

mikkelbu:
@rprouse But @ChrisMaddock only asked for the Rebracer.xml in https://github.com/nunit/nunit/issues/2685#issuecomment-360794664?

ChrisMaddock:
Yup - please see Terje's comment on that thread, Rob.

I'm not sure myself how much ReSharper does by default which will go against what we currently have - so maybe better to leave it in and investigate further if Terje thinks it might cause a problem.

rprouse:
I missed that last comment, thanks. I have approved and merged. Thanks for another contribution  @pandualuvala.

PR #2689, opened 2018-01-27
Don't suppress output under Assert.Throws

### LINKED ISSUE #2688: Assert.Throws swallows console output
```
[TestFixture] // NUnit ver. 3.9
public class IssueTest
{
    [Test]
    public void AssertThrowsConsoleIssue()
    {
        Console.WriteLine(1);
        Assert.Throws<Exception>(() =>
        {
            Console.WriteLine(2);
            throw new Exception("test");
        });
        Console.WriteLine(3);
    }
}
```

I expect an output to be

```
1
2
3
```

But in reality it is

```
1
3
```

From my investigation `Assert.ThrowsAssync()` does not have described issue, only `Assert.Throws()` has.

Original question on stackoverflow: https://stackoverflow.com/questions/48453402/nunit-assert-throws-swallows-console-output

CharliePoole:
Fixes #2688

The fix takes any saved output in the isolated context and writes it to the original context.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -30,17 +30,15 @@ using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Threading;
-using NUnit.Framework.Interfaces;
-#if NET20
 using NUnit.Compatibility;
-#endif
+using NUnit.Framework.Interfaces;
 
 namespace NUnit.Framework.Internal
 {
     /// <summary>
     /// The TestResult class represents the result of a test.
     /// </summary>
-    public abstract class TestResult : ITestResult
+    public abstract class TestResult : LongLivedMarshalByRefObject, ITestResult
     {
         #region Fields
 
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -563,6 +563,7 @@ namespace NUnit.Framework.Internal
             /// </summary>
             public void Dispose()
             {
+                _originalContext.OutWriter.Write(CurrentContext.CurrentResult.Output);
                 CurrentContext = _originalContext;
             }
         }
--- a/src/NUnitFramework/tests/Assertions/AssertThrowsTests.cs
+++ b/src/NUnitFramework/tests/Assertions/AssertThrowsTests.cs
@@ -37,18 +37,34 @@ namespace NUnit.Framework.Assertions
         }
 
         [Test]
-        public void ThrowsDoesNotDiscardOutput()
+        public void AssertThrowsDoesNotDiscardOutput()
         {
             Console.WriteLine(1);
             Assert.Throws<Exception>(() =>
             {
                 Console.WriteLine(2);
+                TestContext.WriteLine(3);
                 throw new Exception("test");
             });
-            Console.WriteLine(3);
+            Console.WriteLine(4);
 
-            Assert.That(TestExecutionContext.CurrentContext.CurrentResult.Output, 
-                Is.EqualTo("1" + Environment.NewLine + "2" + Environment.NewLine + "3" + Environment.NewLine));
+            var NL = Environment.NewLine;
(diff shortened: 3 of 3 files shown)


ChrisMaddock:
Based on your comment at https://github.com/nunit/nunit/issues/2688#issuecomment-360919201 - shouldn't we handle everything else on the TestResult in the same way? TestAttachments seem very similar - what about assertion counts? Are nested assertions inside an `Assert.Throws` a supported feature?

CharliePoole:
It's a good question and suggests that maybe we shouldn't even do this.

My assumption was that the SUT is actually doing the writing and that the code on SO was only an example. When I look at the other things we are throwing away, I don't find anything else the SUT would do.

I'll ask about the use case on the issue.

We allow any code under Throws, so there could be an assert, but we intentionally throw it away, keeping only the exception. We use this in testing NUnit.

ChrisMaddock:
‘SUT’ Sorry?

It does look like a potential rabbit hole. Although I would expect the code sample to work too. Maybe there’s somewhere of implementing a ‘deferred context’ - which automatically passes up everything, without needing each item specified explicitly in the dispose...

CharliePoole:
System under test.

The code you are testing or a library it calls may write to the console.

It may.

ChrisMaddock:
Thanks. As you’d probably guessed - I’d assumed the opposite!

CharliePoole:
Based on the comments on the issue itself, I think this is the correct course of action. I'm particularly influenced by the difference in how we are handling various output types...

Type of Output | Without PR | With PR
----------|----------------|-----------
stdout | Not Shown | Shown
stderr   | Shown | Shown
progress | Shown | Shown

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -30,17 +30,15 @@
...
-#endif
+using NUnit.Framework.Interfaces;
 
 namespace NUnit.Framework.Internal
 {
     /// <summary>
     /// The TestResult class represents the result of a test.
     /// </summary>
-    public abstract class TestResult : ITestResult
+    public abstract class TestResult : LongLivedMarshalByRefObject, ITestResult
```
jnm2: What prompted this change?

ChrisMaddock: I had the same question. 🙂

ChrisMaddock:
Thanks for the table Charlie - I agree that should be fixed. I think it makes more sense when thinking of `Console.Out()` than `TestContext.Write()`.

On the latter - I could see it being confusing (although niche) that `TestContext.XXX` won't work entirely inside an `Assert.Throws()`. That seems like a separate issue at this point however.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
CharliePoole: One of the tests in `LowTrustFIxture` was throwing an exception when trying to access the test result across a remoting boundary. If we ever implement a facility to isolate a test or fixture in a separate AppDomain, we'll probably find more things like this to fix.

CharliePoole:
@ChrisMaddock Actually, the same fix took care of the TestContext.Xxx issues as well. Output either goes immediately to the runner as an event or is saved as part of the result. It's the saved result that was a problem.

I'm taking this as approved based on your comment.

ChrisMaddock:
Sorry, that wasn’t what I meant - I understand that console.out and testcontext.out go the same way.

My concern was that we’ve now made testcontext.out work, but something like testcontext.addattachment still won’t work - which seems inconsistent.

But I agree this good to get merge - and ‘less inconaistent’ than we were before with stdout and stderr differing in behaviour.

CharliePoole:
Yes, my first reaction was that everything should work. It's the technical guy's point of view. But then thinking about it, Assert.Throws is really intended to run methods from the system being tested, not to run tests themselves. __We__ run tests, of course, but that's because our SUT __is__ a test framework.

As far as I can figure out, writing to console is the only thing that non-test code can do that ends up in a test result. So I'm happy to leave it as is - at least until / unless some other use case comes up.

Issue #2628, opened 2017-12-19
Error during installing tools when running build script


OmicronPersei:
When I ran `.\build.ps1 -t test45`, I saw

> Error finding repository for 'https://ci.appveyor.com/nuget/nunit': An error occurred while retrieving package metadata for 'NUnit.ConsoleRunner.3.7.0' from source 'NUnit AppVeyor CI'.
  A task was canceled.

CC @jnm2

OmicronPersei:
I should note the build and tests all performed otherwise fine.

jnm2:
@rprouse Is this needed?

https://github.com/nunit/nunit/blob/e465d6966971cdcdd7859439902707e2c56414cb/NuGet.config#L4

OmicronPersei:
Also noticed another one when doing `build.ps1 -t clean`:
> Error finding repository for 'https://www.myget.org/F/nunit/api/v2': An error occurred while retrieving package metadata for 'NUnit.ConsoleRunner.3.7.0' from source 'NUnit MyGet Feed'.
  A task was canceled.

jnm2:
This might not be our bug:

 - https://github.com/cake-build/cake/issues/1937
 - https://github.com/cake-build/cake/issues/1877

rprouse:
@jnm2 that line isn't needed for NUnit as far as I can think of. We've copied that `NuGet.config` file from repo to repo. We add in the various AppVeyor feeds so we can easily test builds in different repos, so we just include them all for convenience.

jnm2:
If anyone would like to contribute this PR before I get to it, please go ahead.

jnm2:
@rprouse– I took from that that we're okay with the nuget.configs diverging between repos and going ahead with removing the line. Let me make sure; is that what you meant?

rprouse:
I am fine with them diverging. There is no reason to reference the NUnit CI packages from NUnit.

pandualuvala:
Hello can i take this Task? Thanks

jnm2:
Yes, it's yours! Let us know if you need anything.

PR #2690, opened 2018-01-30
update removed the reference for NUnit CI packages from NUnit.config …

pandualuvala:
update removed the reference for NUnit CI packages from NUnit.config file based on issue #2628


rprouse:
Build failure is the unrelated .NET 3.5 bug so I will override. It is also such a small change, I am not going to ask for a second review.

Thanks for the help 👍

pandualuvala:
Thank you @rprouse will pick other issue tomorrow.

PR #2692, opened 2018-01-31
Extracted a PropertyBagAdapter on the TestContext.Properties property.

### LINKED ISSUE #1578: TestContext.CurrentTest exposes too much internal info
The following problems should be fixed... and the code examined for any others...
1. The Properties property exposes IPropertyBag, which is intended as an extensibility interface, not one for users. Among other things, it allows setting properties, which is not a good idea and potentially confusing since they would not be persisted. The internal PropertyBag should be wrapped by an adapter object, just as we wrap the test and the result.
2. The Properties indexer returns an IList. We should determine if this can be made type-safe. If it cannot be done at the present time (IIRC, we have both string and int properties at a minimum) then this should be spun out as a separate future issue.
3. The IList returned by the indexer gives too much access. An IEnumerable would be better. Alternatively, another layer of adapter with enumeration and Count available would do the job.

Hard to know how to prioritize this. I'm marking it Low for now, but if we see other issues arising that relate to this problem we should increase it.

OmicronPersei:
Closes #1578 .

### DIFF
--- a/src/NUnitFramework/framework/TestContext.cs
+++ b/src/NUnitFramework/framework/TestContext.cs
@@ -22,6 +22,7 @@
 // ***********************************************************************
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
@@ -418,11 +419,11 @@ namespace NUnit.Framework
             }
 
             /// <summary>
-            /// The properties of the test.
+            /// A shallow copy of the properties of the test.
             /// </summary>
-            public IPropertyBag Properties
+            public PropertyBagAdapter Properties
             {
-                get { return _test.Properties; }
+                get { return new PropertyBagAdapter(_test.Properties); }
             }
 
             /// <summary>
@@ -547,6 +548,83 @@ namespace NUnit.Framework
 #endregion
         }
 
-#endregion
+        #endregion
+
+        #region Nested PropertyBagAdapter Class
+
+        /// <summary>
+        /// <see cref="PropertyBagAdapter"/> adapts an <see cref="IPropertyBag"/>
+        /// for consumption by the user.
+        /// </summary>
+        public class PropertyBagAdapter
+        {
+            private IPropertyBag _source;
+            
+            /// <summary>
+            /// Construct a <see cref="PropertyBagAdapter"/> from a source
+            /// <see cref="IPropertyBag"/>.
+            /// </summary>
+            public PropertyBagAdapter(IPropertyBag source)
+            {
+                _source = source;
+            }
+
+            /// <summary>
+            /// Get the first property at <paramref name="key"/>, if it can be found, otherwise
+            /// returns null.
+            /// </summary>
+            public string Get(string key)
+            {
+                return _source.Get(key)?.ToString() ?? null;
+            }
+
+            /// <summary>
+            /// Returns if <paramref name="key"/> is found in this
+            /// <see cref="PropertyBagAdapter"/>.
+            /// </summary>
+            public bool ContainsKey(string key)
+            {
+                return _source.ContainsKey(key);
+            }
+
+            /// <summary>
+            /// Returns an <see cref="IEnumerable{String}"/> of properties
+            /// at the <paramref name="key"/>.
+            /// </summary>
+            public IEnumerable<string> this[string key]
+            {
+                get
+                {
+                    var collection = new List<string>();
+                    foreach(var obj in _source[key])
+                    {
+                        collection.Add(obj.ToString());
+                    }
+
+                    return collection;
+                }
+            }
+
+            /// <summary>
+            /// Returns the count of elements at the <paramref name="key"/>.
+            /// </summary>
+            public int Count(string key)
+            {
+                return _source[key].Count;
+            }
+
+            /// <summary>
+            /// Returns an <see cref="ICollection{String}"/> of the property keys.
+            /// </summary>
+            public ICollection<string> Keys
+            {
+                get
+                {
+                    return _source.Keys;
+                }
+            }
+        }
+
+        #endregion
     }
 }
--- a/src/NUnitFramework/framework/TestContext.cs
+++ b/src/NUnitFramework/framework/TestContext.cs
@@ -570,16 +570,16 @@ namespace NUnit.Framework
             }
 
             /// <summary>
-            /// Get the first property at <paramref name="key"/>, if it can be found, otherwise
+            /// Get the first property with the given <paramref name="key"/>, if it can be found, otherwise
             /// returns null.
             /// </summary>
-            public string Get(string key)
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/framework/TestContext.cs
```diff
@@ -547,6 +548,83 @@ public int InconclusiveCount
...
+            /// Get the first property at <paramref name="key"/>, if it can be found, otherwise
+            /// returns null.
+            /// </summary>
+            public string Get(string key)
+            {
+                return _source.Get(key)?.ToString() ?? null;
+            }
+
+            /// <summary>
+            /// Returns if <paramref name="key"/> is found in this
```
jnm2: Could `Indicates whether` be clearer?

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -547,6 +548,83 @@ public int InconclusiveCount
...
+            /// <summary>
+            /// Returns the count of elements at the <paramref name="key"/>.
+            /// </summary>
+            public int Count(string key)
+            {
+                return _source[key].Count;
+            }
+
+            /// <summary>
+            /// Returns an <see cref="ICollection{String}"/> of the property keys.
```
jnm2: Sadly XML crefs do not make it possible to specify closed generic types. All this is doing is aliasing `T` to `String`, as though they had picked  `String` rather than `T` as the name of the generic type parameter. (I forget, IntelliSense may display this as `T` anyway.)
I'd say `a collection` since the return type is already highlighted on the line above in the IntelliSense tooltip.

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -547,6 +548,83 @@ public int InconclusiveCount
...
+                    {
+                        collection.Add(obj.ToString());
+                    }
+
+                    return collection;
+                }
+            }
+
+            /// <summary>
+            /// Returns the count of elements at the <paramref name="key"/>.
```
jnm2: (just a suggestion) `at the` -> `with the given`?

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -547,6 +548,83 @@ public int InconclusiveCount
...
+            /// <see cref="PropertyBagAdapter"/>.
+            /// </summary>
+            public bool ContainsKey(string key)
+            {
+                return _source.ContainsKey(key);
+            }
+
+            /// <summary>
+            /// Returns an <see cref="IEnumerable{String}"/> of properties
+            /// at the <paramref name="key"/>.
```
jnm2: (just a suggestion) `at the` -> `with the given`?

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -547,6 +548,83 @@ public int InconclusiveCount
...
+            /// Construct a <see cref="PropertyBagAdapter"/> from a source
+            /// <see cref="IPropertyBag"/>.
+            /// </summary>
+            public PropertyBagAdapter(IPropertyBag source)
+            {
+                _source = source;
+            }
+
+            /// <summary>
+            /// Get the first property at <paramref name="key"/>, if it can be found, otherwise
```
jnm2: (just a suggestion) `at` -> `with the given`?

OmicronPersei:
Looks like Travis is failing on this:
>4) Failed : NUnit.Framework.Internal.Execution.TestWorkerTests.BusyExecuteIdleEventsCalledInSequence

I'll restart it for a second time.  :(

src/NUnitFramework/tests/Internal/PropertyBagAdapterTests.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            _source.Add("key", "val1");
+            _source.Add("key", "val2");
+            _source.Add("meaningOfLife", 42);
+
+            _adapter = new PropertyBagAdapter(_source);
+        }
+
+        [Test]
+        public void PropertyBagAdapter_Get_CanAccesKeysFromSourceIPropertyBag()
```
jnm2: Typo in 'Access'

src/NUnitFramework/tests/Internal/PropertyBagAdapterTests.cs
```diff
@@ -0,0 +1,80 @@
...
+        public void PropertyBagAdapter_UpdatesWhenSourcePropertyBagUpdates()
+        {
+            _source.Add("newKey", "newVal");
+
+            Assert.IsTrue(_adapter.ContainsKey("newKey"));
+            Assert.AreEqual("newVal", _adapter.Get("newKey"));
+            CollectionAssert.AreEquivalent(new string[] { "newVal" }, _adapter["newKey"]);
+            Assert.AreEqual(1, _adapter.Count("newKey"));
+            CollectionAssert.AreEquivalent(new string[] { "key", "meaningOfLife", "newKey" }, _adapter.Keys);
+        }
```
jnm2: All the tests are good, but this one is the best! 😃

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -547,6 +548,83 @@ public int InconclusiveCount
...
+            /// returns null.
+            /// </summary>
+            public string Get(string key)
+            {
+                return _source.Get(key)?.ToString() ?? null;
```
jnm2: Only real concern: I do think this should stay `object`.

src/NUnitFramework/tests/TestContextTests.cs
```diff
@@ -221,10 +222,19 @@ public void SetUpCanAccessTestId()
...
+            Assert.That(TestContext.CurrentContext.Test.Properties.Get("Answer"), Is.EqualTo("42"));
+        }
+
+        [Test]
+        public void TestGettingPropertiesReturnsNewObjEveryTime()
+        {
+            PropertyBagAdapter adapter1 = CurrentContext.Test.Properties;
+            PropertyBagAdapter adapter2 = CurrentContext.Test.Properties;
+
+            Assert.AreNotEqual(adapter1, adapter2);
```
jnm2: If this changed, would we care?

OmicronPersei: Probably not, and I'd probably argue for exactly that reason, this test should be removed.  Originally I had `CurrentContext.Test.Properties` return a new `PropertyBagAdapter` every time because, upon construction, it created a shallow-copy of the current state of the given `IPropertyBag`, instead of what it does now, holding a reference to the source `IPropertyBag`.  In that context, it made sense to ensure that calling `.Properties` always returned a new object, probably not anymore.

CharliePoole: It's a bit academic since `TestContext.CurrentContext` always returns a new object.

Issue #2699, opened 2018-02-03
Warnings are not consistently displayed


CharliePoole:
Working on #2695 I realized I'd opened a rather large can of works around the display of warnings. I'm using this issue to try to summarize the problems and figure out a direction.

As background: warnings were added after multiple asserts and rely on the infrastructure created for multiple asserts. That's why they work well in conjunction with failures but not with other terminating events.

In this context, a terminating event for a test is one of the following:
 * An Error, i.e. an exception that is not one NUnit itself understands
 * A failure, caused by Assert.Fail or any assert that fails.
 * An Ignored event, caused by Assert.Ignore
 * An Inconclusive event, caused by Assert.Inconclusive or the failure of an Assume
 * A Pass event, caused by Assert.Pass

A Warning, of course, is not a terminating event.

The following table shows the problems I see when several warnings are followed by a terminating event __outside__ of a Multiple Assert block. As in...

```C#
Assert.Warn("Warning1");
Assert.Warn("Warning2");
Assert.Fail("Terminating Event!");
```

The action varies according to the nature of the terminating event.

Event | Displays | Test Result | Notes
------|----------|------------|-------
Error | Warnings plus error | Error     | Correct behavior
Fail    | Warnings plus failure | Failure  | Correct behavior
Ignore | Ignore message    | Ignored | Should display warnings
Inconclusive | Inconclusive message| Inconclusive | Should display warnings
Pass | Pass message | Pass | Should display warnings. Not clear whether this should override the warnings and Pass. Options: (1) Record as error (2)  record as warning (3) add an extra warning and pass.

If the terminating event occurs within a multiple assert block, the actions are slightly different...

Event | Displays | Test Result | Notes
------|----------|------------|-------
Error | Warnings plus error | Error     | Correct behavior
Fail    | Warnings plus failure (at end of block) | Failure  | Correct behavior
Ignore | Ignore message  | Error | Should display warnings. See discussion below.
Inconclusive | Inconclusive message| Error | Should display warnings. See discussion below.
Pass | Pass message | Error | Should display warnings. See discussion below.

Three of these events cause an error result. Assert.Pass, Assume failures, Assert.Inconclusive and Assert.Pass are not permitted inside a multiple assert block. This seems arbitrary in retrospect. I see two general approaches:
1. Let the events terminate execution immediately, within the block.
2. Save the results till the end of the block.

[...]

PR #2700, opened 2018-02-03
Add TestResult to MultipleAssertException

### LINKED ISSUE #2695: MultipleAssertException doesn't provide proper details on failures
When using `Assert.Multiple` and assertion failures happen inside the lambda, a `MultipleAssertException` is thrown by the framework.

**This exception has absolutely no details on the actual failures that happened**, which drastically impacts the debugging experience: users have to wait for the run to finish and then inspect the Test Explorer window to see what errors have occurred.

This is where it is being thrown currently:
`https://github.com/nunit/nunit/blob/efd115b26c8f5d4ebe6d52f5108097ceb0c77ffb/src/NUnitFramework/framework/Assert.cs#L340`
```
            if (context.MultipleAssertLevel == 0 && context.CurrentResult.PendingFailures > 0)
                throw new MultipleAssertException();
```

This results in the following message:
> "One or more failures in Multiple Assert block:"

This gives no hint to what happened and leaves the developer wondering how to see the actual failures. I think this is a really bad practice with exceptions in general: they should provide as much detail as humanly possible.

I'm aware that the display on Test Explorer has full details, but NUnit shouldn't assume that all developers _run_ tests instead of _debugging_ them.

I propose the following:
1. **That failures be integrated as inner exceptions**
This would be similar to how an `AggregateException` works: have all failures as child exceptions so that people debugging the application can see what errors occurred. Maybe using an `AggregateException` would be ideal, but I heard you had some issues with that because the type doesn't exist in all versions of the framework. Instead of just inspecting the `PendingFailures` counter like above, we could convert any `AssertionResult` into `AssertionException`s and add them to the parent `MultipleAssertException` before throwing it. I'd even suggest ditching the `AssertionResult` model itself in favor of directly handling `AssertionException` objects, but that's probably a bigger discussion that I don't want to get into right now.

2. **Create an user-friedly `.ToString` implementation on `MultipleAssertException`**
In combination with # 1, the exception's `ToString` should provide a summary of failures. That would allow developers to see what errors were reported easily, and only have to inspect the inner exceptions in case further details are required. Maybe ommiting the stacktrace for each error on this summary would be a good idea, for example.

This lack of debugging support had a very significant impact when migrating to `Assert.Multiple` on our team: we used to have a custom method to perform this task which reported a much cleaner error in debug mode showing all assertions that failed, and when we switched to `Assert.Multiple` there was justified backlash.

CharliePoole:
This fixes #2695 by including the test result and combined failure/warning messages in the MultipleAssert exception for use in debugging. In addition, the same thing is done for AssertionException.

Additional tests were added to highlight problems with warning messages but the tests are currently ignored since they don't pass. See issue #2699

It's clear that more work is needed in the recording of multiple results and in generally cleaning up the `TestResult` class. This PR should serve as a good start.

### DIFF
--- a/src/NUnitFramework/framework/Exceptions/MultipleAssertException.cs
+++ b/src/NUnitFramework/framework/Exceptions/MultipleAssertException.cs
@@ -45,11 +45,15 @@ namespace NUnit.Framework
         /// used normally, when exiting the multiple assert block with failures.
         /// Not used internally but provided to facilitate debugging.
         /// </summary>
-        /// <param name="result">The current result</param>
-        public MultipleAssertException(ITestResult result)
-            : base(result.Message)
+        /// <param name="testResult">
+        /// The current result, up to this point. The result is not used 
+        /// internally by NUnit but is provided to facilitate debugging.
+        /// </param>
+        public MultipleAssertException(ITestResult testResult)
+            : base(testResult?.Message)
         {
-            TestResult = result;
+            Guard.ArgumentNotNull(testResult, "testResult");
+            TestResult = testResult;
         }
 
         /// <param name="message">The error message that explains 
@@ -70,7 +74,7 @@ namespace NUnit.Framework
 #endif
 
         /// <summary>
-        /// Gets the ResultState provided by this exception
+        /// Gets the <see cref="ResultState"/> provided by this exception.
         /// </summary>
         public override ResultState ResultState
         {
@@ -78,8 +82,8 @@ namespace NUnit.Framework
         }
 
         /// <summary>
-        /// Gets the ITestResult of this test at the point the exception was thrown
+        /// Gets the <see cref="ITestResult"/> of this test at the point the exception was thrown,
         /// </summary>
-        public ITestResult TestResult { get; private set; }
+        public ITestResult TestResult { get; }
     }
 }
--- a/src/NUnitFramework/framework/Exceptions/MultipleAssertException.cs
+++ b/src/NUnitFramework/framework/Exceptions/MultipleAssertException.cs
@@ -36,9 +36,21 @@ namespace NUnit.Framework
     public class MultipleAssertException : ResultStateException
     {
         /// <summary>
-        /// Default Constructor (normally used)
+        /// Default Constructor
         /// </summary>
-        public MultipleAssertException() : base("One or more failures in Multiple Assert block:") { } 
+        public MultipleAssertException(string message) : base(message) { }
+
+        /// <summary>
+        /// Construct based on the TestResult so far. This is the constructor
+        /// used normally, when exiting the multiple assert block with failures.
+        /// Not used internally but provided to facilitate debugging.
+        /// </summary>
+        /// <param name="result">The current result</param>
+        public MultipleAssertException(ITestResult result)
+            : base(result.Message)
+        {
+            TestResult = result;
+        }
 
         /// <param name="message">The error message that explains 
         /// the reason for the exception</param>
@@ -64,5 +76,10 @@ namespace NUnit.Framework
         {
             get { return ResultState.Failure; }
         }
+
+        /// <summary>
+        /// Gets the ITestResult of this test at the point the exception was thrown
+        /// </summary>
+        public ITestResult TestResult { get; private set; }
     }
 }
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -496,24 +496,15 @@ namespace NUnit.Framework.Internal
             if (ex is NUnitException || ex is TargetInvocationException)
                 ex = ex.InnerException;
 
-            if (ex is AssertionException)
-            {
-                AssertionResults.Add(new AssertionResult(AssertionStatus.Failed, ex.Message, StackFilter.DefaultFilter.Filter(ex.StackTrace)));
-            }
-
             if (ex is ResultStateException)
-            {
-                string message = ex is MultipleAssertException
-                    ? CreateLegacyFailureMessage()
-                    : ex.Message;
-
-                string stackTrace = StackFilter.DefaultFilter.Filter(ex.StackTrace);
-
-                SetResult(((ResultStateException)ex).ResultState, message, stackTrace);
-            }
+                SetResult(
+                    ((ResultStateException)ex).ResultState,
+                    ex.Message, 
+                    StackFilter.DefaultFilter.Filter(ex.StackTrace));
 #if !NETSTANDARD1_6
             else if (ex is System.Threading.ThreadAbortException)
-                SetResult(ResultState.Cancelled,
+                SetResult(
+                    ResultState.Cancelled,
                     "Test cancelled by user",
                     ex.StackTrace);
 #endif
@@ -594,7 +585,8 @@ namespace NUnit.Framework.Internal
         }
 
         /// <summary>
-        /// Determine result after test has run to completion.
+        /// Update overall test result, including legacy Message, based
+        /// on AssertionResults that have been saved to this point.
         /// </summary>
         public void RecordTestCompletion()
         {
@@ -639,6 +631,27 @@ namespace NUnit.Framework.Internal
         public void RecordAssertion(AssertionStatus status, string message)
         {
             RecordAssertion(status, message, null);
+        }
+
+
+        /// <summary>
+        /// Creates a failure message incorporating failures
+        /// from a Multiple Assert block for use by runners
+        /// that don't know about AssertionResults.
+        /// </summary>
+        /// <returns>Message as a string</returns>
+        public string CreateLegacyFailureMessage()
+        {
+            var writer = new StringWriter();
+
+            if (AssertionResults.Count > 1)
+                writer.WriteLine("Multiple failures or warnings in test:");
+
+            int counter = 0;
+            foreach (var assertion in AssertionResults)
+                writer.WriteLine(string.Format("  {0}) {1}", ++counter, assertion.Message));
+
+            return writer.ToString();
         }
 
         #endregion
@@ -734,25 +747,6 @@ namespace NUnit.Framework.Internal
             }
 
             return attachmentsNode;
-        }
-
-        /// <summary>
-        /// Creates a failure message incorporating failures
-        /// from a Multiple Assert block for use by runners
-        /// that don't know about AssertionResults.
-        /// </summary>
-        /// <returns>Message as a string</returns>
-        private string CreateLegacyFailureMessage()
-        {
-            var writer = new StringWriter();
-
-            writer.WriteLine("\n  One or more failures in Multiple Assert block:");
-
-            int counter = 0;
-            foreach (var assertion in AssertionResults)
-                writer.WriteLine(string.Format("  {0}) {1}", ++counter, assertion.Message));
-
-            return writer.ToString();
         }
 
         #endregion
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -640,7 +640,7 @@ namespace NUnit.Framework.Internal
         /// that don't know about AssertionResults.
         /// </summary>
         /// <returns>Message as a string</returns>
-        public string CreateLegacyFailureMessage()
+        private string CreateLegacyFailureMessage()
         {
             var writer = new StringWriter();
 
--- a/src/NUnitFramework/tests/Assertions/AssertFailTests.cs
+++ b/src/NUnitFramework/tests/Assertions/AssertFailTests.cs
@@ -135,7 +135,7 @@ namespace NUnit.Framework.Assertions
             });
 
             // Ensure that no spurious info was recorded from the assertion
-            Assert.That(TestExecutionContext.CurrentContext.CurrentResult.AssertionResults.Count, Is.EqualTo(0));
+            Assert.That(TestExecutionContext.CurrentContext.CurrentResult.AssertionResults, Is.Empty);
         }
 
         [Test]
--- a/src/NUnitFramework/tests/Assertions/AssertFailTests.cs
+++ b/src/NUnitFramework/tests/Assertions/AssertFailTests.cs
@@ -73,7 +73,7 @@ namespace NUnit.Framework.Assertions
 
             Assert.AreEqual(ResultState.Failure, result.ResultState);
 
-            Assert.AreEqual(result.AssertionResults.Count, 1);
+            Assert.AreEqual(1, result.AssertionResults.Count);
             var assertion = result.AssertionResults[0];
             Assert.That(assertion.Status, Is.EqualTo(AssertionStatus.Failed));
         }
@@ -88,7 +88,7 @@ namespace NUnit.Framework.Assertions
             Assert.AreEqual(ResultState.Failure, result.ResultState);
             Assert.AreEqual("MESSAGE", result.Message);
 
-            Assert.AreEqual(result.AssertionResults.Count, 1);
+            Assert.AreEqual(1, result.AssertionResults.Count);
             var assertion = result.AssertionResults[0];
             Assert.That(assertion.Status, Is.EqualTo(AssertionStatus.Failed));
             Assert.That(assertion.Message, Is.EqualTo("MESSAGE"));
@@ -104,13 +104,13 @@ namespace NUnit.Framework.Assertions
             Assert.AreEqual(ResultState.Failure, result.ResultState);
             Assert.AreEqual("MESSAGE: 2+2=4", result.Message);
 
-            Assert.AreEqual(result.AssertionResults.Count, 1);
+            Assert.AreEqual(1, result.AssertionResults.Count);
             var assertion = result.AssertionResults[0];
             Assert.That(assertion.Status, Is.EqualTo(AssertionStatus.Failed));
             Assert.That(assertion.Message, Is.EqualTo("MESSAGE: 2+2=4"));
         }
 
-        [Test]
+        [Test, Ignore("Currently Fails: Must use Assert.Catch or Assert.Throws")]
         public void CatchingAssertionExceptionMakesTestPass()
         {
             try
@@ -121,6 +121,30 @@ namespace NUnit.Framework.Assertions
             {
                 // Eat the exception
             }
+
+            // Ensure that no spurious info was recorded from the assertion
+            Assert.That(TestExecutionContext.CurrentContext.CurrentResult.AssertionResults.Count, Is.EqualTo(0));
+        }
+
+        [Test]
+        public void AssertCatchMakesTestPass()
+        {
+            Assert.Catch(() =>
+            {
+                Assert.Fail("This should not be seen");
+            });
+
+            // Ensure that no spurious info was recorded from the assertion
+            Assert.That(TestExecutionContext.CurrentContext.CurrentResult.AssertionResults.Count, Is.EqualTo(0));
+        }
+
+        [Test]
+        public void AssertThrowsMakesTestPass()
+        {
+            Assert.Throws<AssertionException>(() =>
+            {
+                Assert.Fail("This should not be seen");
+            });
 
             // Ensure that no spurious info was recorded from the assertion
             Assert.That(TestExecutionContext.CurrentContext.CurrentResult.AssertionResults.Count, Is.EqualTo(0));
--- a/src/NUnitFramework/tests/Assertions/AssertWarnTests.cs
+++ b/src/NUnitFramework/tests/Assertions/AssertWarnTests.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2009 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -21,6 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+using System;
 using NUnit.Framework.Interfaces;
 using NUnit.TestData;
 using NUnit.TestUtilities;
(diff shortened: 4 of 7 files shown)


src/NUnitFramework/framework/Exceptions/MultipleAssertException.cs
```diff
@@ -64,5 +76,10 @@ public override ResultState ResultState
         {
             get { return ResultState.Failure; }
         }
+
+        /// <summary>
+        /// Gets the ITestResult of this test at the point the exception was thrown
```
julealgon: Missing period at the end.

src/NUnitFramework/framework/Exceptions/MultipleAssertException.cs
```diff
@@ -36,9 +36,21 @@ namespace NUnit.Framework
...
+        public MultipleAssertException(string message) : base(message) { }
+
+        /// <summary>
+        /// Construct based on the TestResult so far. This is the constructor
+        /// used normally, when exiting the multiple assert block with failures.
+        /// Not used internally but provided to facilitate debugging.
+        /// </summary>
+        /// <param name="result">The current result</param>
+        public MultipleAssertException(ITestResult result)
+            : base(result.Message)
```
julealgon: Missing argument validation. If `result` is null, this will throw. You'll probably need some inline static validation to keep using it as an argument to the base. Something like this:
```
: base(EnsureNotNull(result).Message)
```

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -498,24 +498,15 @@ public void RecordException(Exception ex)
...
-            }
+                SetResult(
+                    ((ResultStateException)ex).ResultState,
+                    ex.Message, 
+                    StackFilter.DefaultFilter.Filter(ex.StackTrace));
 #if !NETSTANDARD1_6
             else if (ex is System.Threading.ThreadAbortException)
-                SetResult(ResultState.Cancelled,
+                SetResult(
+                    ResultState.Cancelled,
```
julealgon: Unrelated formatting change, I'd suggest not performing it as part of this PR to avoid behavior+refactor changes in the same boat.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -498,24 +498,15 @@ public void RecordException(Exception ex)
...
-                    : ex.Message;
-
-                string stackTrace = StackFilter.DefaultFilter.Filter(ex.StackTrace);
-
-                SetResult(((ResultStateException)ex).ResultState, message, stackTrace);
-            }
+                SetResult(
+                    ((ResultStateException)ex).ResultState,
+                    ex.Message, 
+                    StackFilter.DefaultFilter.Filter(ex.StackTrace));
```
julealgon: I'd suggest keeping the scope (`{}`) for readability (since the if block still has more than one line).

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -498,24 +498,15 @@ public void RecordException(Exception ex)
...
-                string message = ex is MultipleAssertException
-                    ? CreateLegacyFailureMessage()
-                    : ex.Message;
-
-                string stackTrace = StackFilter.DefaultFilter.Filter(ex.StackTrace);
-
-                SetResult(((ResultStateException)ex).ResultState, message, stackTrace);
-            }
+                SetResult(
+                    ((ResultStateException)ex).ResultState,
```
julealgon: This looks unnecessarily complex. I'd suggest either changing this logic into `as`->`!= null` (to avoid the code duplication), or even better, rely on the new switch with pattern matching.

I know this is probably out of scope for this PR though.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -643,6 +635,27 @@ public void RecordAssertion(AssertionStatus status, string message)
...
         }
 
+
+        /// <summary>
+        /// Creates a failure message incorporating failures
+        /// from a Multiple Assert block for use by runners
+        /// that don't know about AssertionResults.
+        /// </summary>
+        /// <returns>Message as a string</returns>
+        public string CreateLegacyFailureMessage()
```
julealgon: Why did we make this `public` now? I don't see any extra usages on the PR.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -643,6 +635,27 @@ public void RecordAssertion(AssertionStatus status, string message)
...
+        {
+            var writer = new StringWriter();
```
julealgon: Isn't this `IDisposable`? I'd strongly suggest wrapping in an `using` block.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -643,6 +635,27 @@ public void RecordAssertion(AssertionStatus status, string message)
...
+
+            if (AssertionResults.Count > 1)
+                writer.WriteLine("Multiple failures or warnings in test:");
```
julealgon: Looks like we are changing the message here. Is it really related to these changes? I'd suggest not performing such a change as part of this PR.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -643,6 +635,27 @@ public void RecordAssertion(AssertionStatus status, string message)
...
+
+            int counter = 0;
+            foreach (var assertion in AssertionResults)
```
julealgon: Consider using the `Select` overload that returns the index instead of maintaining a separate `counter` variable.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -643,6 +635,27 @@ public void RecordAssertion(AssertionStatus status, string message)
...
+                writer.WriteLine(string.Format("  {0}) {1}", ++counter, assertion.Message));
```
julealgon: Can we use interpolation here?

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
```diff
@@ -104,13 +104,13 @@ public void AssertFailWorksWithMessageAndArgs()
             Assert.AreEqual(ResultState.Failure, result.ResultState);
             Assert.AreEqual("MESSAGE: 2+2=4", result.Message);
 
-            Assert.AreEqual(result.AssertionResults.Count, 1);
+            Assert.AreEqual(1, result.AssertionResults.Count);
```
julealgon: These inverted arguments are existing technical debt and not really part of this change, so I'd suggest fixing them in a separate PR.

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
```diff
@@ -104,13 +104,13 @@ public void AssertFailWorksWithMessageAndArgs()
...
             var assertion = result.AssertionResults[0];
             Assert.That(assertion.Status, Is.EqualTo(AssertionStatus.Failed));
             Assert.That(assertion.Message, Is.EqualTo("MESSAGE: 2+2=4"));
         }
 
-        [Test]
+        [Test, Ignore("Currently Fails: Must use Assert.Catch or Assert.Throws")]
```
julealgon: As mentioned in the issue, I'm personally not fond of this breaking change. Even though NUnit has this goal of separating test termination from the failures, it just results in unintuitive behavior to the end user.

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
```diff
@@ -125,5 +125,29 @@ public void CatchingAssertionExceptionMakesTestPass()
...
+        [Test]
+        public void AssertCatchMakesTestPass()
+        {
+            Assert.Catch(() =>
+            {
+                Assert.Fail("This should not be seen");
+            });
+
+            // Ensure that no spurious info was recorded from the assertion
+            Assert.That(TestExecutionContext.CurrentContext.CurrentResult.AssertionResults.Count, Is.EqualTo(0));
```
julealgon: Can we use `IsEmpty` instead of checking the `Count` property directly? We have an assertion for empty collections right?

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
```diff
@@ -125,5 +125,29 @@ public void CatchingAssertionExceptionMakesTestPass()
...
+        [Test]
+        public void AssertThrowsMakesTestPass()
+        {
+            Assert.Throws<AssertionException>(() =>
+            {
+                Assert.Fail("This should not be seen");
+            });
+
+            // Ensure that no spurious info was recorded from the assertion
+            Assert.That(TestExecutionContext.CurrentContext.CurrentResult.AssertionResults.Count, Is.EqualTo(0));
```
julealgon: Same here.

src/NUnitFramework/tests/Assertions/AssertWarnTests.cs
```diff
@@ -51,5 +52,47 @@ public void AssertWarnWorksWithMessageAndArgs()
...
+
+        [Test]
+        public void WarningsAreDisplayedWithFailure()
+        {
+            ITestResult result = TestBuilder.RunTestCase(
+                typeof(WarningFixture),
+                "TwoWarningsAndFailure");
+
+            Assert.That(result.ResultState, Is.EqualTo(ResultState.Failure));
+            Assert.That(result.AssertionResults.Count, Is.EqualTo(3));
```
julealgon: Similarly here, don't we have an assertion to check collection counts directly, `Has.Exactly`?

CharliePoole:
@julealgon The description of the PR explains how I broadened the scope, mostly from necessity because simply adding a populated result to the exception resulted in other problems. I'll point out the details inline.

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
CharliePoole: Agreed. OTOH, being able to catch an exception thereby preventing a failure is purely an accident of implementation, even though it's an "obvious" implementation. Note that you have __never__ been above to prevent a multiple assert failure by catching an exception and - of course - you can't prevent warnings that way either.

The only alternative I can see is that asserts not use the current result at all, but just make and update a copy of it, which they then throw. That seems like quite a lot of work and I'd want to apply it to all types of exceptions uniformly if we did it.

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
CharliePoole: I thought of it as "number of assertion results is zero" and wrote it that way. Style, I guess.

src/NUnitFramework/tests/Assertions/AssertWarnTests.cs
CharliePoole: My personal style is to try to limit myself to the most basic assertions when testing NUnit. Otherwise, I have to worry about what I'm testing in order to avoid using the feature that I am currently testing.

src/NUnitFramework/framework/Exceptions/MultipleAssertException.cs
julealgon: > Maybe base(result?.Message) and a Guard in the constructor.

That works but the validation will be misleading as you'd be validating the internal `Message` argument and not the `result` itself. A static validation check in this class can properly indicate that `result` itself is `null`.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
julealgon: > Personally, I'm not a fan of that. I think the indentation tells the story well enough. Some people fear errors due to indentation that doesn't actually match the nesting, but I haven't run into problems with it myself.

This is absolutely fine and more of a standardization thing honestly. I know StyleCop has a rule for this but as long as you are not following it in this repo it should be OK to ignore it (after all, the `else` block below suffers from the same issue).

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
julealgon: > I thought of it as "number of assertion results is zero" and wrote it that way. Style, I guess.

On this particular one, I disagree that it is a matter of style. The collection constraints provide better semantic and were created for this exact purpose. Additionally, they also help with reducing the amount of child member access/law of demeter violation (which is already insane over here with 4 levels).

src/NUnitFramework/tests/Assertions/AssertWarnTests.cs
julealgon: > My personal style is to try to limit myself to the most basic assertions when testing NUnit. Otherwise, I have to worry about what I'm testing in order to avoid using the feature that I am currently testing.

Now this is interesting and I see where you are coming from. Going by that logic, I'd even consider using exclusively the barebones `Assert.IsTrue`.

CharliePoole: So it's a balance between good messages and simplicity. To have good messages with IsTrue you have to add a message with each assert.

As a matter of general interest, the nunit 3 syntax came from nunitlite, which I boots trapped with about a half dozen assertions.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
jnm2: I mean... maybe. StringWriter is a minimal wrapper around a StringBuilder. So long as the instance doesn't leave this scope this doesn't bother me either way. Of course then, why not use a StringBuilder directly?

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
CharliePoole: @julealgon Which of the two approaches were you giving the thumbs up to?

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
jnm2: @julealgon I favor your approach because it yields a clearer failure message, but I think this is a matter of style no different than choosing between `collection.Count == 0` and `collection.Any()`.

CharliePoole:
@julealgon Some of your comments are what I'd call "polishing" the code. I like them but I'll wait to see if the overall approach is satisfactory to everyone before spending time. Don't like to "polish" and then discard!

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
CharliePoole: @jnm2 Sold! Based on the failure message.

I still hold back on a lot of Linq stuff for the same reason you like it. It just doesn't speak to me. I have to think twice to remember what each method does, whereas I know immediately what count == 0 means. That's particularly true with a method like Any() - a name that has its own use in NUnit. 

I wish we had a bit more specialization. Then, the people who did most of the work in a particular area of the code could decide on stylistic issues to their liking. Those who only dabbled would use that style for consistency. Of course, I was the guy for a long time in this particular area... 😄

src/NUnitFramework/framework/Internal/Results/TestResult.cs
julealgon: I know it's not that big of a deal here, but it is a known good practice to always dispose `IDisposable` instances. There is just no reason not to in this case as well.

I don't mind if you want to change it into a `StringBuilder` though.

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
jnm2: Ironically, contrary to what I just said, I still instinctively use `.Count == 0` rather than `.Any()`. I'm still trying to kick the habit as a premature optimization.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
jnm2: The argument in favor of doing this is that StringWriter may change to do something like lease a StringBuilder and return it to the pool on Dispose. Even though I can imagine the framework compat council vetoing that because people have not always disposed it, you are right about this.

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
julealgon: > @julealgon Which of the two approaches were you giving the thumbs up to?

I was just agreeing with your comment in general, in that I understand the huge challenge it would be to do something different right now, even though I'm still not fond of the proposed approach.

As I said, it is infinitely more intuitive to control failures _and_  test termination via the same exceptions, but it looks like the entire team is set on the idea of having these concerns separated. I'm leaving this one up to you guys.

CharliePoole:
I'll have to refresh my memory on this one to see what changes were asked in the reviews. IIRC they were all incidental.

rprouse:
The changes were incidental (that is why I approved), but there were a few that you mentioned were good ideas. Too many comments to read through 😄

CharliePoole:
Rebased on current master.

rprouse:
AppVeyor build failure is one of the unrelated .NET 3.5 failures. I am good to override and merge if you are done @CharliePoole

src/NUnitFramework/framework/Exceptions/MultipleAssertException.cs
```diff
@@ -45,11 +45,15 @@ public class MultipleAssertException : ResultStateException
         /// used normally, when exiting the multiple assert block with failures.
         /// Not used internally but provided to facilitate debugging.
         /// </summary>
-        /// <param name="result">The current result</param>
-        public MultipleAssertException(ITestResult result)
-            : base(result.Message)
+        /// <param name="testResult">
+        /// The current result, up to this point. The result is not used
```
julealgon: Did you forget to remove this from the ctor?

src/NUnitFramework/framework/Exceptions/MultipleAssertException.cs
```diff
@@ -45,11 +45,15 @@ public class MultipleAssertException : ResultStateException
...
-            : base(result.Message)
+        /// <param name="testResult">
+        /// The current result, up to this point. The result is not used 
+        /// internally by NUnit but is provided to facilitate debugging.
+        /// </param>
+        public MultipleAssertException(ITestResult testResult)
+            : base(testResult?.Message)
         {
-            TestResult = result;
+            Guard.ArgumentNotNull(testResult, "testResult");
```
julealgon: Can we use `nameof(testResult)` instead?

julealgon: I'm still not convinced this is a good approach, because argument validation is being performed _after_ calling the base constructor. In these cases, inline static validation like what I posted earlier seems like the better way to go.

CharliePoole: Didn't see this one before merging. I considered creating some new utilities along the line of what you suggested to ensure the arg is not null but decided it was out of scope. The call to the base is made to be OK, but the derived constructor will then throw with a proper message. That seems good enough for now.

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
jnm2: @CharliePoole Since we've been telling folks not to rely on this behavior, shouldn't we switch this from TODO to by-design (remove Ignore and negate the test)?

src/NUnitFramework/tests/Assertions/AssertWarnTests.cs
```diff
@@ -51,5 +52,47 @@ public void AssertWarnWorksWithMessageAndArgs()
...
+            Assert.That(result.Message, Contains.Substring("First warning"));
+            Assert.That(result.Message, Contains.Substring("Second warning"));
+            Assert.That(result.Message, Contains.Substring("This fails"));
+        }
+
+        [Test, Ignore("Currently Fails: Ignored message is displayed without the warnings")]
```
jnm2: @CharliePoole Could you start an issue and gather the information into one place so we can understand how to get to work on these new TODO items?

CharliePoole: @jnm2 I take it that by TODOs you mean the two ignored tests in this class? Do we know that the tests still don't work?

jnm2: They both still fail on master if I unignore them.

jnm2: Yes. They both still fail on master if I unignore them.

jnm2: @CharliePoole Three TODOs were introduced by this PR. https://github.com/nunit/nunit/pull/2924 handles one which is not related to these two, as far as I can see. I'd appreciate having the information about how to follow up in a new assignable issue, if you're willing and any of this is still fresh in your mind.

src/NUnitFramework/tests/Assertions/AssertFailTests.cs
jnm2: Fixed by https://github.com/nunit/nunit/pull/2924.

src/NUnitFramework/tests/Assertions/AssertWarnTests.cs
CharliePoole: I'm totally confused now! AFAIK a TODO is a formatted comment that starts `// TODO`, which is specially handled by Visual Studio. I also understand that "todo" is a generic phrase for something that needs to be done. Can you tell me what you mean by it? Where are the three todos? Can I read them?

Earlier in the thread, you asked me to create an issue, which is a term we do have a common understanding of (I think), but now you seem to have resolved the issues in a PR.

I'm almost 100% sure that you are using a term you think I'll understand 😈 but it's not working!

jnm2: I'm so sorry! Replace "TODO" with "Ignore" and the above should make sense.
I had internalized that "Ignore" means "use sparsely and over short periods of time," but I hadn't internalized that this was different than a TODO.

PR #2702, opened 2018-02-07
Switch editorconfig warnings to suggestions

### LINKED ISSUE #2698: Syntax suggestions errors as warnings
Looks like my VS2017 is treating some of these as errors, I want to say our `.editorconfig` settings are probably causing this?  Are these intended by chance?

image

rprouse:
Fixes #2698 

See the issue for more info, but I have confirmed that even though `editorconfig` had various rules explicitly set to warnings, they were showing up as errors in both the editor and in the error list. Switching to Suggestion works. It isn't as obvious, but we have so much code that doesn't match that it is probably the best solution for now.

They now show up as grey blocks in the scroll bar and with grey dots under the suggestion.

image

The suggestions also appear in the error list as messages now.

image

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -49,8 +49,8 @@ csharp_indent_case_contents = true
 csharp_indent_switch_labels = true
 
 # https://github.com/nunit/docs/wiki/Coding-Standards#naming
-dotnet_style_predefined_type_for_locals_parameters_members = true:warning
-dotnet_style_predefined_type_for_member_access = true:warning
+dotnet_style_predefined_type_for_locals_parameters_members = true:suggestion
+dotnet_style_predefined_type_for_member_access = true:suggestion
 
 # The first matching rule wins, more specific rules at the top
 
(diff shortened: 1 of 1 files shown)


rprouse:
Suggested several reviewers, but any of you can review and merge.

rprouse:
I only saw it when opening files with certain types of errors like name simplification. A full rebuild doesn't show anything. I searched for the problem on Google, but didn't come up with anything.

ChrisMaddock:
Is it possible related to us having 'warnings as errors' in the csproj's? I don't know why that would vary per machine however. Certain VS versions having bugs/changes in behaviour?

mikkelbu:
@ChrisMaddock I think you are right. At least it fits with the experiments I've made the last hour. Also it sounds like this is expected behaviour in https://github.com/dotnet/roslyn/issues/21332#issuecomment-321245896

rprouse:
I am okay with the warning being mentioned on 84. We would only ever switch it if the codebase was all updated.

PR #2703, opened 2018-02-09
Add AnyOfConstraint to test if item is equal to any of provided values

Dreamescaper:
Tests whether actual item is equal to any of provided expected values.

Usage:
`Assert.That("1", Is.AnyOf("1", "2", "3"))`

### DIFF
--- a/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
@@ -21,7 +21,6 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-
 using System;
 using System.Collections;
 using System.Collections.Generic;
@@ -47,6 +46,10 @@ namespace NUnit.Framework.Constraints
         /// <param name="expected">Expected collection</param>
         public AnyOfConstraint(IEnumerable expected) : base(expected)
         {
+            Guard.ArgumentNotNull(expected, nameof(expected));
+            Guard.ArgumentValid(expected.Cast<object>().Any(),
+                $"{nameof(AnyOfConstraint)} requires non-empty expected collection!", nameof(expected));
+
             _expected = expected;
         }
 
@@ -69,7 +72,7 @@ namespace NUnit.Framework.Constraints
         public override string ToString()
         {
             var expectedStrings = _expected.Cast<object>().Select(MsgUtils.FormatValue).ToArray();
-            return $"<{DisplayName.ToLower()} {string.Join(", ", expectedStrings)}>";
+            return $"<{DisplayName.ToLower()} {string.Join(" ", expectedStrings)}>";
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
@@ -107,7 +107,6 @@ namespace NUnit.Framework.Constraints
         /// Flag the constraint to use the supplied IComparer object.
         /// </summary>
         /// <param name="comparer">The IComparer object to use.</param>
-        /// <returns>Self.</returns>
         public AnyOfConstraint Using(IComparer comparer)
         {
             _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
@@ -118,7 +117,6 @@ namespace NUnit.Framework.Constraints
         /// Flag the constraint to use the supplied IComparer object.
         /// </summary>
         /// <param name="comparer">The IComparer object to use.</param>
-        /// <returns>Self.</returns>
         public AnyOfConstraint Using<T>(IComparer<T> comparer)
         {
             _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
@@ -129,7 +127,6 @@ namespace NUnit.Framework.Constraints
         /// Flag the constraint to use the supplied Comparison object.
         /// </summary>
         /// <param name="comparer">The Comparison object to use.</param>
-        /// <returns>Self.</returns>
         public AnyOfConstraint Using<T>(Comparison<T> comparer)
         {
             _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
@@ -140,7 +137,6 @@ namespace NUnit.Framework.Constraints
         /// Flag the constraint to use the supplied IEqualityComparer object.
         /// </summary>
         /// <param name="comparer">The IEqualityComparer object to use.</param>
-        /// <returns>Self.</returns>
         public AnyOfConstraint Using(IEqualityComparer comparer)
         {
             _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
@@ -151,7 +147,6 @@ namespace NUnit.Framework.Constraints
         /// Flag the constraint to use the supplied IEqualityComparer object.
         /// </summary>
         /// <param name="comparer">The IComparer object to use.</param>
-        /// <returns>Self.</returns>
         public AnyOfConstraint Using<T>(IEqualityComparer<T> comparer)
         {
             _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
--- a/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
@@ -29,21 +29,17 @@ using System.Linq;
 namespace NUnit.Framework.Constraints
 {
     /// <summary>
-    /// <see cref="AnyOfConstraint"/> is used to determine whether item is equal to any of parameters
+    /// <see cref="AnyOfConstraint"/> is used to determine whether the value is equal to any of the expected values.
     /// </summary>
     public class AnyOfConstraint : Constraint
     {
         private readonly IEnumerable _expected;
-
-        /// <summary>
-        /// The NUnitEqualityComparer in use for this constraint
-        /// </summary>
         private readonly NUnitEqualityComparer _comparer = NUnitEqualityComparer.Default;
 
         /// <summary>
         /// Construct a <see cref="AnyOfConstraint"/>
         /// </summary>
-        /// <param name="expected">Expected collection</param>
+        /// <param name="expected">Collection of expected values</param>
         public AnyOfConstraint(IEnumerable expected) : base(expected)
         {
             Guard.ArgumentNotNull(expected, nameof(expected));
@@ -68,7 +64,6 @@ namespace NUnit.Framework.Constraints
         /// <summary>
         /// Returns the constraint DisplayName followed by arguments within angle brackets.
         /// </summary>
-        /// <returns></returns>
         public override string ToString()
         {
             var expectedStrings = _expected.Cast<object>().Select(MsgUtils.FormatValue).ToArray();
@@ -80,8 +75,6 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         /// <typeparam name="TActual">Actual item type</typeparam>
         /// <param name="actual">Actual item</param>
-        /// <returns>A <see cref="ConstraintResult"/> indicating whether item is present 
-        /// in expected collection.</returns>
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
             var tolerance = Tolerance.Default;
--- a/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2007 Charlie Poole, Rob Prouse
+// Copyright (c) 2018 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
@@ -25,6 +25,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace NUnit.Framework.Constraints
 {
@@ -41,10 +42,10 @@ namespace NUnit.Framework.Constraints
         private readonly NUnitEqualityComparer _comparer = NUnitEqualityComparer.Default;
 
         /// <summary>
-        /// Construct a OneOfConstraint
+        /// Construct a <see cref="AnyOfConstraint"/>
         /// </summary>
         /// <param name="expected">Expected collection</param>
-        public AnyOfConstraint(IEnumerable expected)
+        public AnyOfConstraint(IEnumerable expected) : base(expected)
         {
             _expected = expected;
         }
@@ -59,6 +60,16 @@ namespace NUnit.Framework.Constraints
             {
                 return "any of " + MsgUtils.FormatValue(_expected);
             }
+        }
+
+        /// <summary>
+        /// Returns the constraint DisplayName followed by arguments within angle brackets.
+        /// </summary>
+        /// <returns></returns>
+        public override string ToString()
+        {
+            var expectedStrings = _expected.Cast<object>().Select(MsgUtils.FormatValue).ToArray();
+            return $"<{DisplayName.ToLower()} {string.Join(", ", expectedStrings)}>";
         }
 
         /// <summary>
@@ -158,12 +169,6 @@ namespace NUnit.Framework.Constraints
         public AnyOfConstraint Using<T>(Func<T, T, bool> comparer)
         {
             _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
-            return this;
-        }
-
-        internal AnyOfConstraint Using(EqualityAdapter adapter)
-        {
-            _comparer.ExternalComparers.Add(adapter);
             return this;
         }
 
--- a/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
@@ -29,9 +29,9 @@ using System.Collections.Generic;
 namespace NUnit.Framework.Constraints
 {
     /// <summary>
-    /// <see cref="ItemFromConstraint"/> is used to determine whether item is present in a collection
+    /// <see cref="AnyOfConstraint"/> is used to determine whether item is equal to any of parameters
     /// </summary>
-    public class ItemFromConstraint : Constraint
+    public class AnyOfConstraint : Constraint
     {
         private readonly IEnumerable _expected;
 
@@ -44,7 +44,7 @@ namespace NUnit.Framework.Constraints
         /// Construct a OneOfConstraint
         /// </summary>
         /// <param name="expected">Expected collection</param>
-        public ItemFromConstraint(IEnumerable expected)
+        public AnyOfConstraint(IEnumerable expected)
         {
             _expected = expected;
         }
@@ -57,7 +57,7 @@ namespace NUnit.Framework.Constraints
         {
             get
             {
-                return "item from " + MsgUtils.FormatValue(_expected);
+                return "any of " + MsgUtils.FormatValue(_expected);
             }
         }
 
@@ -87,7 +87,7 @@ namespace NUnit.Framework.Constraints
         /// <summary>
         /// Flag the constraint to ignore case and return self.
         /// </summary>
-        public ItemFromConstraint IgnoreCase
+        public AnyOfConstraint IgnoreCase
         {
             get
             {
@@ -101,7 +101,7 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         /// <param name="comparer">The IComparer object to use.</param>
         /// <returns>Self.</returns>
-        public ItemFromConstraint Using(IComparer comparer)
+        public AnyOfConstraint Using(IComparer comparer)
         {
             _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
@@ -112,7 +112,7 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         /// <param name="comparer">The IComparer object to use.</param>
         /// <returns>Self.</returns>
-        public ItemFromConstraint Using<T>(IComparer<T> comparer)
+        public AnyOfConstraint Using<T>(IComparer<T> comparer)
         {
             _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
@@ -123,7 +123,7 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         /// <param name="comparer">The Comparison object to use.</param>
         /// <returns>Self.</returns>
-        public ItemFromConstraint Using<T>(Comparison<T> comparer)
+        public AnyOfConstraint Using<T>(Comparison<T> comparer)
         {
             _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
@@ -134,7 +134,7 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         /// <param name="comparer">The IEqualityComparer object to use.</param>
         /// <returns>Self.</returns>
-        public ItemFromConstraint Using(IEqualityComparer comparer)
+        public AnyOfConstraint Using(IEqualityComparer comparer)
         {
             _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
@@ -145,7 +145,7 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         /// <param name="comparer">The IComparer object to use.</param>
         /// <returns>Self.</returns>
-        public ItemFromConstraint Using<T>(IEqualityComparer<T> comparer)
+        public AnyOfConstraint Using<T>(IEqualityComparer<T> comparer)
         {
             _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
@@ -155,13 +155,13 @@ namespace NUnit.Framework.Constraints
         /// Flag the constraint to use the supplied boolean-returning delegate.
         /// </summary>
         /// <param name="comparer">The supplied boolean-returning delegate to use.</param>
-        public ItemFromConstraint Using<T>(Func<T, T, bool> comparer)
+        public AnyOfConstraint Using<T>(Func<T, T, bool> comparer)
         {
             _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
-        internal ItemFromConstraint Using(EqualityAdapter adapter)
+        internal AnyOfConstraint Using(EqualityAdapter adapter)
         {
             _comparer.ExternalComparers.Add(adapter);
             return this;
--- a/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
@@ -994,6 +994,11 @@ namespace NUnit.Framework.Constraints
         /// <param name="expected">Expected values</param>
         public Constraint AnyOf(params object[] expected)
         {
+            if (expected == null)
+            {
+                expected = new object[] { null };
+            }
+
             return Append(new AnyOfConstraint(expected));
         }
 
--- a/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
@@ -986,24 +986,15 @@ namespace NUnit.Framework.Constraints
 
         #endregion
 
-        #region Item From
+        #region AnyOf
 
         /// <summary>
-        /// Returns a constraint that tests if an item is present in the collection
+        /// Returns a constraint that tests if an item is equal to any of parameters
         /// </summary>
-        /// <param name="expected">Expected collection</param>
-        public Constraint ItemFrom(IEnumerable expected)
+        /// <param name="expected">Expected values</param>
+        public Constraint AnyOf(params object[] expected)
         {
-            return Append(new ItemFromConstraint(expected));
-        }
-
-        /// <summary>
-        /// Returns a constraint that tests if an item is present in the collection
-        /// </summary>
-        /// <param name="expected">Expected collection</param>
-        public Constraint ItemFrom(params object[] expected)
-        {
-            return Append(new ItemFromConstraint(expected));
+            return Append(new AnyOfConstraint(expected));
         }
 
         #endregion
--- a/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
@@ -206,7 +206,7 @@ namespace NUnit.Framework.Constraints
         #endregion
 
         #region One
-       
+
         /// <summary>
         /// Returns a <see cref="ItemsConstraintExpression"/>, which will
         /// apply the following constraint to a collection of length one, succeeding
@@ -219,7 +219,7 @@ namespace NUnit.Framework.Constraints
                 builder.Append(new ExactCountOperator(1));
                 return new ItemsConstraintExpression(builder);
             }
-            
+
         }
 
         #endregion
@@ -986,5 +986,26 @@ namespace NUnit.Framework.Constraints
 
         #endregion
 
+        #region Item From
+
+        /// <summary>
+        /// Returns a constraint that tests if an item is present in the collection
+        /// </summary>
+        /// <param name="expected">Expected collection</param>
+        public Constraint ItemFrom(IEnumerable expected)
+        {
+            return Append(new ItemFromConstraint(expected));
+        }
+
+        /// <summary>
+        /// Returns a constraint that tests if an item is present in the collection
+        /// </summary>
+        /// <param name="expected">Expected collection</param>
+        public Constraint ItemFrom(params object[] expected)
+        {
+            return Append(new ItemFromConstraint(expected));
+        }
+
+        #endregion
     }
 }
--- a/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
@@ -994,6 +994,12 @@ namespace NUnit.Framework.Constraints
         /// <param name="expected">Expected values</param>
         public Constraint AnyOf(params object[] expected)
         {
+            if (expected.Length == 0)
+            {
+                throw new ArgumentException(nameof(expected),
+                    "Expected values are required for AnyOf constraint!");
+            }
+
             return Append(new AnyOfConstraint(expected));
         }
 
--- a/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
@@ -994,12 +994,6 @@ namespace NUnit.Framework.Constraints
         /// <param name="expected">Expected values</param>
         public Constraint AnyOf(params object[] expected)
         {
-            if (expected.Length == 0)
-            {
-                throw new ArgumentException(nameof(expected),
-                    "Expected values are required for AnyOf constraint!");
-            }
-
             return Append(new AnyOfConstraint(expected));
         }
 
--- a/src/NUnitFramework/framework/Is.cs
+++ b/src/NUnitFramework/framework/Is.cs
@@ -494,6 +494,12 @@ namespace NUnit.Framework
         /// <param name="expected">Expected values</param>
         public static AnyOfConstraint AnyOf(params object[] expected)
         {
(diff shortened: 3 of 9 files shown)


CharliePoole:
I'm not seeing what value this adds, since it only allows you to specify what you can already do, except by reversing the order of the apps.

Dreamescaper:
Well, when I use
`Assert.That(expectedValues, Does.Contain(actualValue))`
in case of failure it will log something like this:
```
Expected: does contain *actualValue*
Actual: *expectedValues*
```

Which is really confusing, since `actual`  and `expected` are switched places

jnm2:
I like this, but I much prefer the preposition 'in.' `Is.In(2, 3, 5)`, `Has.All.In(2, 3, 5)`, etc.

[We ask](https://github.com/nunit/nunit/blob/master/CONTRIBUTING.md#fixing-bugs-and-adding-features) that you start a discussion issue separately from providing the implementation so that reviews on the implementation aren't mixed in with the initial design discussion.

Dreamescaper:
Sure! 
Created issue: https://github.com/nunit/nunit/issues/2704

CharliePoole:
@Dreamescaper OK. I got confused by the description as being about containment. It's really about being equal to one of several possible expected values... even though we may use containment as an implementation tool.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
```diff
@@ -0,0 +1,172 @@
...
+        /// <summary>
+        /// Test whether item is present in expected collection
+        /// </summary>
+        /// <typeparam name="TActual">Actual item type</typeparam>
+        /// <param name="actual">Actual item</param>
+        /// <returns>A <see cref="ConstraintResult"/> indicating whether item is present 
+        /// in expected collection.</returns>
+        public override ConstraintResult ApplyTo<TActual>(TActual actual)
+        {
+            var tolerance = Tolerance.Default;
```
CharliePoole: Why not Tolerance.Exact? Since you decided not to support tolerance (Within, etc.) on this constraint, there's no way to specify a tolerance with the exception of using the `DefaultFloatingPointTolerance` attribute on the test. So all matches will end up being exact __except__ for floating point. Is that your intention?

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -986,5 +986,17 @@ public Constraint Exist
 
         #endregion
 
+        #region AnyOf
+
+        /// <summary>
+        /// Returns a constraint that tests if an item is equal to any of parameters
+        /// </summary>
+        /// <param name="expected">Expected values</param>
+        public Constraint AnyOf(params object[] expected)
```
CharliePoole: We're breaking new ground. This is only the second constraint to take more than one argument and the first to accept a variable number of arguments, modeled as a params array. That leads to a number of things we should discuss.

1. Should `AnyOf()` be allowed to compile? We can avoid that by specifying the first argument separately as discussed on the issue.

2. Should `AnyOf(42)` be allowed to compile? We can avoid that by specifying the first two arguments separately.

3. Should we force all arguments to the same type by use of a generic? If so, should there be constraints? A generic without no constraints will allow anything, so it's the same as specifying `object`. However, we could limit this in some ways, depending on how we want to guide folks in using the syntax.

src/NUnitFramework/framework/Is.cs
```diff
@@ -486,5 +486,18 @@ public static RangeConstraint InRange(object from, object to)
 
         #endregion
 
+        #region Any Of
+
+        /// <summary>
+        /// Returns a constraint that tests if an item is equal to any of parameters
+        /// </summary>
+        /// <param name="expected">Expected values</param>
+        public static AnyOfConstraint AnyOf(params object[] expected)
```
CharliePoole: Same as above, regarding the form of the constructor.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
```diff
@@ -0,0 +1,172 @@
...
+                return this;
+            }
+        }
+
+        /// <summary>
+        /// Flag the constraint to use the supplied IComparer object.
+        /// </summary>
+        /// <param name="comparer">The IComparer object to use.</param>
+        /// <returns>Self.</returns>
+        public AnyOfConstraint Using(IComparer comparer)
```
CharliePoole: Do you have a use case in mind for "Using?" We shouldn't add it "just in case" it is needed because we have learned we can always add things but pretty much __never__ can remove them!

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
```diff
@@ -0,0 +1,172 @@
...
+
+        /// <summary>
+        /// The Description of what this constraint tests, for
+        /// use in messages and in the ConstraintResult.
+        /// </summary>
+        public override string Description
+        {
+            get
+            {
+                return "any of " + MsgUtils.FormatValue(_expected);
```
CharliePoole: I had to do quite a bit of code searching to figure out what the description would look like and particularly whether it would be truncated. It appears that it will be in the form `any of <1, 42, 99, 1234, 4321>` and no truncation will occur.

I'm thinking we should
1. Decide on the format. You might think that array formats are already decided by `MsgUtils` but in this case we use a `params` array to model separate arguments. Maybe we want `1 OR 42 OR 99` or something equally silly.
2. Add a little comment here if we keep the present format so folks won't have to search.
3. Add tests of the description.
4. Add tests of the error message.

src/NUnitFramework/tests/Constraints/AnyOfConstraintTests.cs
```diff
@@ -0,0 +1,59 @@
...
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+namespace NUnit.Framework.Constraints
+{
+    [TestFixture]
+    public class AnyOfConstraintTests
```
CharliePoole: These are not just Constraint tests. They are testing the constraint as used __through__ the `Is.AnyOf` syntax. That's not a bad thing to do, but we generally test constraints and syntax elements separately. I would not get rid of these, but I'd like to add the separate tests.

In particular, testing a constraint separately allows us to easily write tests of the description and the error message. In many cases (not this one, of course) we have alternative syntax that can lead to the same constraint. So to test the syntax, we can simply verify that it produces the proper constraints, which we do by examining the string representation of the generated constraint expression.

src/NUnitFramework/tests/Constraints/AnyOfConstraintTests.cs
```diff
@@ -0,0 +1,59 @@
...
+            Assert.That("aB", Is.AnyOf("a", "B", "ab").IgnoreCase);
+        }
+
+        [Test]
+        public void ItemIsAbsent_WithEqualityComparer()
+        {
+            Func<object, object, bool> comparer = Equals;
+            TestDelegate testAction = () => Assert.That("1", Is.AnyOf(1, 2, 3));
+
+            Assert.That(testAction, Throws.TypeOf(typeof(AssertionException)));
```
CharliePoole: You should check the message here.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
```diff
@@ -0,0 +1,172 @@
+// ***********************************************************************
+// Copyright (c) 2007 Charlie Poole, Rob Prouse
```
jnm2: 2018

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
```diff
@@ -0,0 +1,172 @@
...
+    {
+        private readonly IEnumerable _expected;
+
+        /// <summary>
+        /// The NUnitEqualityComparer in use for this constraint
+        /// </summary>
+        private readonly NUnitEqualityComparer _comparer = NUnitEqualityComparer.Default;
+
+        /// <summary>
+        /// Construct a OneOfConstraint
```
jnm2: AnyOfConstraint- this is a good candidate for being `<see cref="AnyOfConstraint"/>`.

src/NUnitFramework/tests/Constraints/AnyOfConstraintTests.cs
```diff
@@ -0,0 +1,59 @@
+// ***********************************************************************
+// Copyright (c) 2009 Charlie Poole, Rob Prouse
```
jnm2: 2018

src/NUnitFramework/tests/Constraints/AnyOfConstraintTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        [Test]
+        public void ItemIsPresent_IgnoreCase()
+        {
+            Assert.That("aB", Is.AnyOf("a", "B", "ab").IgnoreCase);
+        }
+
+        [Test]
+        public void ItemIsAbsent_WithEqualityComparer()
+        {
+            Func<object, object, bool> comparer = Equals;
```
jnm2: Line appears unused?

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
jnm2: Never having figured out how to use the `ref tolerance` parameter, to me this code reads as "I don't care about tolerance; give me what people want by default."

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
jnm2: It did cross my mind that we might want to consider having these equality-related constraints inherit a common base.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
jnm2: `any of <2, 3, 5>` seems to be most consistent by the rest of NUnit. :+1:

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
jnm2: 1 and 2: Either of these would prevent use of  `.AnyOf(expectedList.ToArray())`. Also, haven't seen collection parameters split out like this before in .NET and it's not appealing to me.

3: If we make it generic, it'll be harder to use because C# as of 7.2) doesn't have the "best common type" proposal implemented. It would force you to write `Is.AnyOf<Animal>(new Cat(), new Dog())` rather than `Is.AnyOf(new Cat(), new Dog())`. I'm not seeing a benefit to that right now.

1. and 2. can be covered with an analyzer shipped in the main package. In fact, I'm going to request that we do exactly this: start an absolutely minimal analyzer that does exactly one thing really well (this diagnostic) and put it in the `\analyzers` folder in our main NuGet package. Once  @mikkelbu thinks the NUnit.Analyzers package is at the right point, we can move it there and have our NUnit package take a dependency on NUnit.Analyzers.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
CharliePoole: Speaking as the devil's advocate... It's the standard representation for collections. Here, a user can use the syntax without ever thinking about arrays or collections. In some ways it's like the Between syntax.

Another weird idea... what if the argument were a tuple?

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
CharliePoole: @jnm2 As I indicated in the issue, I'm happy to lose that capability. I'd even be happy to drop the `params` array argument and have four constructors with 2, 3, 4 and 5 args respectively. Or more if you like. The point is that the expected value isn't intended to be an open-ended collection but a set of possible values, which can easily be limited to some small number. This isn't a collection constraint, but a different variety of equality constraint.

BTW, the use of "split args" used to be very common. We removed it from the `Constraint` constructor only five years ago. In fact, a single `params` constructor has always worked quite well in C# but sometimes interoperates poorly when there are non-params overloads to choose as well.

If you adamantly oppose the "split args" approach, should we give an error when there is no argument specified? When only one is specified? I know you are thinking about ways you would use the syntax to build up other things, but you can do that by constructing the constraint directly and not using the user-oriented syntax.

Regarding (3), your example seems artificial to me. In any case, you can use the constraint directly if you have a real situation like that.

WRT to an analyzer, that's cool but IMO should not hold up this PR.

CharliePoole:
@Dreamescaper Suggest you fix the requested changes where @jnm2 and I are __not__ in disagreement. Eventually, we'll agree or a third team member will help tip the balance. 😈

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
jnm2: What about when `expectedList` is a method parameter which is specified alongside each test case, like I brought up in the issue?
For that matter, what's so harmful about a "is contained in this collection" constraint? Data-driven integration tests are common.

I see no need to give a runtime error when zero or one arguments are specified. An analyzer warning seems appropriate to me (but I agree it should not hold up the PR).

I'm sorry it seems artificial! I run into it often. How about this then: `Is.AnyOf<int?>(null, 1)`

CharliePoole: I agree that there are some problems with generic. In particular, it can give rise to error messages that are hard for the user to figure out. OTOH, your example would be a heck of a lot clearer if you had written `Is.Null.Or.EqualTo(1)`. 😄 

Mainly, I'm trying to make the overall point that a given syntactic element does not have to do every possible thing we can think of. The whole reason that syntax is a separate mini-layer within NUnit is so that you can add lots of syntactic elements that call upon common constraints in differing ways. I am seeing this one as simply being an alternative to use of multiple equality constraints when there are 2 to 5 or so possible values.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
jnm2: I like the similarity to collections and do see `EqualTo.Or.EqualTo...` as semantically identical to talking about a collection.
Anyway, the user will be typing values separated by comma and space, so let's keep at least that aspect. The prior syntax `<or <equal 2> <or <equal 3> <equal 5>>>` might suggest that what we really want is `<any of 2, 3, 5>`?

Tuple... interesting. I don't know?

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
jnm2: With `Is.Null.Or.EqualTo(1)` vs `Is.AnyOf(null, 1)`, the latter tempts me. But I get your point.

> I am seeing this one as simply being an alternative to use of multiple equality constraints when there are 2 to 5 or so possible values.

What if I have a whitelist of ten types discovered through reflection and I want to assert that a particular type must be among them?

I'm trying to make the overall point that we should not *raise* the complexity of this thing simply to speculatively deny people certain patterns. That seems almost on the level of raising the complexity to enable speculative patterns.

CharliePoole: OK... so "some small number" rather than "five". My point is that it's illogical to use this in a more general way as we do in collection constraints. I doubt that you'll have a list of 10,000 types that it might be!

I do agree that we should not raise the complexity of __what the user does__ arbitrarily. Raising the complexity of an implementation detail is a different matter, IMO.

I don't think you and I are going to see eye to eye on this. I'm ~~happy~~ willing to go with your view on the question of generics. I do think we should have an error for zero arguments. NUnit already uses runtime errors in lots of places where we hope to add analysis in the future. Users can always disable the analyzer. If we don't do that, then `AnyOf()` will always fail, which means `Not.AnyOf()` will always succeed! Both need to be wrong! I'm OK to allow just one arg, since that's just a degenerate case.

CharliePoole: @Dreamescaper After @jnm2's mixing of `ToString` and `Description` I realized you had not overridden `ToString`. My guess is that the base class will result in `<anyof <System.Object[]>>` which is not very useful. Best way to deal with this is to write the Syntax tests (see Syntax directory for examples) and then figure out what you want it to be. It needs to be something we can use in a test to distinguish one constraint from another clearly.

jnm2: > I do agree that we should not raise the complexity of what the user does arbitrarily. Raising the complexity of an implementation detail is a different matter, IMO.

Yes. But taking the example which you said was legitimate, specifying the expected values via method parameter, we are forcing complexity onto the user. The user wants to be able to do `Is.AnyOf(myArrayParameter)`, not:

```cs
myArrayParameter.Length == 1
    ? (Constraint)Is.EqualTo(myArrayParameter[0])
    : Is.AnyOf(myArrayParameter[0], myArrayParameter[1], myArrayParameter.Skip(2).ToArray())
```
(Ironically, the Animal/Cat/Dog problem shows up here; the Constraint cast is necessary.)

One last attempt to get on the same page, and then I'm happy to concede the decision: Given your statement that passing an expectedValues parameter is a legitimate pattern, I feel like you would agree that there should then be an easy way to use that parameter with Is.AnyOf.

While zero items is a degenerate case as well as the single case, I'm okay with an error. The nice thing about runtime errors is that we can start strict and loosen if need be.

Sorry about mixing up the description! `2 or 3 or 5` is marginally verbose but fine if we don't want to do `any of 2, 3, 5`.

CharliePoole: For your example...

```C#
Assert.That(someValue, new AnyOfConstraint(myArrayParameter);
```

It's definitely not as nice, I agree, but not nearly as bad as the example you gave. Straw man? 😜 

Re Description, I mostly want to make sure that we make an informed decision. Having a test will give us a place to review the decision. Thanks for pointing out that `ToString()` needed to be overridden!

jnm2: No, that's fair! Straw man wasn't my intent. I must have been distracted. So if I was in that position, I'd end up writing a temporary `IsAnyOf(params object[])` method and start an enhancement request. 😊

CharliePoole: It __is__ pretty interesting that the existence of the fluent syntax gives rise to a desire to be able to use it for everything! It may even be that most people don't realize that the second arg to `Assert.That` is simply a constraint, which can be created any way they like.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
Dreamescaper: oopsie :)

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
Dreamescaper: Considering  "is contained in this collection" constraint and usage for arbitrary collections. Not sure if that makes a lot of sense for unit tests, but might be useful for integration testing, where we can't control everything. Something like "get possible values in one place -> verify actual values in another".

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
Dreamescaper: Well, I think I simply forgot about tolerance-related methods, so I'm fine adding them.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
Dreamescaper: Again, not sure if that is really needed for unit tests (same as .Using for .Equals, in fact). But might be useful for ui tests, For example, when example, when we add a lot of entries of 3 types, and that type is displayed with some formatting (uppercase/with index/etc), and formatting itself is not the point of the test.

jnm2: That makes me think of another use case for a collection-friendly `Is.AnyOf`: reusing values defined in a static readonly variable. Otherwise maintenance would be a pain.

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
Dreamescaper: @CharliePoole what would you suggest as ToString result? Currently all the collection constraints (it is still a collection constraint underneath) show  <.. Type[]>, not actual values

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
Dreamescaper: @CharliePoole what would you suggest as ToString result? Currently all the collection constraints (it is still a collection constraint underneath) show  <.. Type[]>, not actual values

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
CharliePoole: So you are you (and @jnm2) good with

1. Applying any command-line default tolerance to a floating point value, rather than zero?
2. __Not__ having any tolerance allowed for any other Types?

BTW, `Tolerance.Exact` is like the "default default". `Tolerance.Default` is changeble using the command-line as well as the attribute.

I'm OK with this if you guys are...

jnm2: I want this to behave however `EqualTo().Or.EqualTo()` and the collection constraints behave when you don't specify a tolerance.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
CharliePoole: OK, you talked me into it! I'd still like an error for an empty array however.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
CharliePoole: Actually, it's not a `CollectionConstraint` in the sense I meant. It inherits from `Constraint`. Our `CollectionConstraints` apply to actual values that are collections. 

The only other multiple argument constraint we have is `RangeConstraint`, which gives you something like `<range 5 42>`. While we could live with `<oneof Type[]>` I'd rather see the list because it would make for better testing.

src/NUnitFramework/tests/Constraints/AnyOfConstraintTests.cs
CharliePoole: Syntax folder.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
CharliePoole: Then of course you want Default. I think it deserves a mention in the docs, because I don't think it's necessarily intuitive in the same way it is for `EqualTo`, particularly since we don't allow changing it.

BTW, if we __did__ want to allow changing, I'd argue that this should be a variant constructor for the `EqualConstraint`, with all the same features, rather than creating a common base.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
jnm2: Great! Okay, let's do that error.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
Dreamescaper: I think we have same behavior in CollectionItemsEqualConstraint - Tolerance.Default used, and no way to change that.
While I'm not sure that behavior that you described is very intuitive, I'd say that having it inconsistent across constraints is even worse.

Dreamescaper:
Updated PR based on comments.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
CharliePoole: We probably should have called the existing `Tolerance.Default` something else, like `Tolerance.Unspecified`. It seems like there is some reason to go through all the constraints and look at their consistency in this regard. A separate, much bigger issue, of course.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        /// <summary>
+        /// Returns the constraint DisplayName followed by arguments within angle brackets.
+        /// </summary>
+        /// <returns></returns>
+        public override string ToString()
+        {
+            var expectedStrings = _expected.Cast<object>().Select(MsgUtils.FormatValue).ToArray();
+            return $"<{DisplayName.ToLower()} {string.Join(", ", expectedStrings)}>";
```
CharliePoole: While it will do no real harm to have the commas, this is different from the formatting of any other constraint. See `RangeConstrant` for example. I'd prefer to have them consistent.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
CharliePoole: `CollectionItemsEqualConstraint` is an abstract class. It's up to the individual constraints to determine if a tolerance makes sense in their context. That said, a quick review indicates to me they don't do a good job of it. As mentioned above, I guess our use of tolerance deserves some kind of overall review that we can't deal with in this issue.

Mainly, I brought this up to ensure that your not including `Within` was intentional. Either decision is OK with me.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// </summary>
+        private readonly NUnitEqualityComparer _comparer = NUnitEqualityComparer.Default;
+
+        /// <summary>
+        /// Construct a <see cref="AnyOfConstraint"/>
+        /// </summary>
+        /// <param name="expected">Expected collection</param>
+        public AnyOfConstraint(IEnumerable expected) : base(expected)
+        {
+            _expected = expected;
```
CharliePoole: Need a guard here for the enumerable not being empty. Constraints may be - often are - used independently of the syntax elements that we create to invoke them.

src/NUnitFramework/framework/Is.cs
```diff
@@ -486,5 +486,24 @@ public static RangeConstraint InRange(object from, object to)
...
+        {
+            if (expected.Length == 0)
+            {
+                throw new ArgumentException(nameof(expected),
+                    "Expected values are required for AnyOf constraint!");
```
CharliePoole: This might be optional if you add the guard to the constructor of the constraint. It would be better to use our Guard class for this in any case, just as a matter of clarity.

src/NUnitFramework/tests/Constraints/AnyOfConstraintTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        [SetUp]
+        public void SetUp()
+        {
+            theConstraint = new AnyOfConstraint(new[] { 1, 2, 3 });
+            expectedDescription = "any of < 1, 2, 3 >";
+            stringRepresentation = "<anyof 1, 2, 3>";
+        }
+
+        private static object[] SuccessData = new object[] { 1, 2, 3 };
+        private static object[] FailureData = new object[] { new object[] { 4, "4" }, new object[] { "A", "\"A\"" } };
```
CharliePoole: Just a side comment... not for this PR... Somebody might go through all these and use a `tuple` for the failure data some time. It would be lots clearer to read then.

src/NUnitFramework/tests/Constraints/AnyOfConstraintTests.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        [Test]
+        public void ItemIsPresent_WithEqualityComparer()
+        {
+            Func<string, string, bool> comparer = (expected, actual) => actual.Contains(expected);
+            var anyOf = new AnyOfConstraint(new[] { "A", "B", "C" }).Using(comparer);
+            Assert.That(anyOf.ApplyTo("1. A").Status, Is.EqualTo(ConstraintStatus.Success));
+        }
+    }
+}
```
CharliePoole: Still no tests of failure message. You changed the one failure test you had to a success, rather than adding a check of the error message.

src/NUnitFramework/tests/Syntax/AnyOfTests.cs
```diff
@@ -0,0 +1,48 @@
...
+// ***********************************************************************
+
+namespace NUnit.Framework.Syntax
+{
+    public class AnyOfTests : SyntaxTest
+    {
+        [SetUp]
+        public void SetUp()
+        {
+            parseTree = "<anyof 1, 2, 3>";
```
CharliePoole: See note above about the commas.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
Dreamescaper: I haven't added `Within` because I wasn't able to come up with a use case. If you think it's worth it - I'll add

src/NUnitFramework/tests/Constraints/AnyOfConstraintTests.cs
Dreamescaper: @CharliePoole
Both failure and error message tests are defined in ConstraintTestBase.

CharliePoole: Fair enough.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
CharliePoole: Nope, like I said, I just wanted it thought about rather than missed. I'm a big fan of explicit decisions and it's sometimes hard just looking at the code to know how a decision was reached.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
jnm2: And `Guard.NotNull` too.

src/NUnitFramework/tests/Constraints/AnyOfConstraintTests.cs
jnm2: > Somebody might go through all these and use a tuple for the failure data some time. It would be lots clearer to read then.

So we can use C# 7 finally? 😍

CharliePoole: @jnm2 I thought you guys already decided that. Don't look to me for an OK! 😕

oznetmaster: Or me :(

jnm2: The last discussion I remember was whether to allow C# 6. I imagine that means a lot of work for Neil, doesn't it?

oznetmaster: I have a Roslyn program which converts C# 6 to C# 3 (since virtually all of the C# 6 features map to the older CLR).  C# 7 introduces a number of new features which cannot be mapped in that way (such as ref variables).

I would not strongly object to C# 6, although I personally find most of its syntactic sugaring to be undesirable, and in fact we do not allow a lot of it in our coding standards because it makes the code less maintainable.

jnm2: @oznetmaster We currently use C# 6. Too bad there's no way to compile using Roslyn and map the IL instead of mapping syntax?

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
Dreamescaper: I wouldn't add that at this point. 
I don't think anyone would really need that (especially considering no one asked for that in CollectionEquivalent constraint, for example).
Besides, full Within functionality (with all those seconds, days, percents etc) messes IntelliSense quite a lot.

As you already said, it's much easier to add functionality if needed than remove if not.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
```diff
@@ -0,0 +1,180 @@
...
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// <see cref="AnyOfConstraint"/> is used to determine whether item is equal to any of parameters
```
jnm2: 💭 Something like `whether the value is equal to any of the expected values.` might sound a little smoother.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
```diff
@@ -0,0 +1,180 @@
...
+    /// </summary>
+    public class AnyOfConstraint : Constraint
+    {
+        private readonly IEnumerable _expected;
+
+        /// <summary>
+        /// The NUnitEqualityComparer in use for this constraint
+        /// </summary>
```
jnm2: 💭 Instead of asking for `cref`, I'm thinking we could remove the XML comment.

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
```diff
@@ -0,0 +1,180 @@
...
+        private readonly NUnitEqualityComparer _comparer = NUnitEqualityComparer.Default;
+
+        /// <summary>
+        /// Construct a <see cref="AnyOfConstraint"/>
+        /// </summary>
+        /// <param name="expected">Expected collection</param>
```
jnm2: 💭 How about `Collection of expected values` in case it sounds like we're expecting `actual` to be a collection?

src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
```diff
@@ -0,0 +1,180 @@
...
+            get
+            {
+                return "any of " + MsgUtils.FormatValue(_expected);
+            }
+        }
+
+        /// <summary>
+        /// Returns the constraint DisplayName followed by arguments within angle brackets.
+        /// </summary>
+        /// <returns></returns>
```
jnm2: 💭 `<returns>` is never shown inside the IDE so [we like to leave it out](https://github.com/nunit/docs/wiki/Best-practices-for-XML-documentation#returns-and-value) (or put it in `<summary>` if it's important).
(7 occurrences)

src/NUnitFramework/framework/Is.cs
```diff
@@ -486,5 +486,18 @@ public static RangeConstraint InRange(object from, object to)
...
 
+        #region Any Of
+
+        /// <summary>
+        /// Returns a constraint that tests if an item is equal to any of parameters
+        /// </summary>
+        /// <param name="expected">Expected values</param>
+        public static AnyOfConstraint AnyOf(params object[] expected)
+        {
+            return new AnyOfConstraint(expected);
```
jnm2: ⚠ When people use `Is.AnyOf(null)`, we should use `new object[] { null }` like we do for all our other params APIs.
This doesn't apply to the `AnyOfConstraint(IEnumerable)` constructor itself.

Can we have tests for `Is.AnyOf(null)` and `Builder.AnyOf(null)` in AnyOfTests, please?

CharliePoole: `Builder.AnyOf` isn't a real thing. It's just a construct used in our syntax tests.

CharliePoole: Proper handling of null passed to a method taking a params array as it's argument is very difficult. We need to specify exactly what should happen in a test, as you say.

jnm2: @CharliePoole `Builder.AnyOf` is how you'd write the test for `ConstraintExpression.AnyOf` which needs separate handling from `Is.AnyOf`.

CharliePoole: Right, in the common base syntax tests. You'd just need to add a case that has null as an accepted value and specify what the ToString() should look like. We should do that for sure.

Dreamescaper: Added null value handling, and tests to Syntax tests (AnyOfTests -> AnyOf_NullValue_Tests).

btw, is that common approach to define new tests by inheriting from base class with defined tests, and to change only setUp method? Seems like I hate that, haha :))))

Dreamescaper:
Updated PR with xml docs fixes, null argument handling and tests for null argument

src/NUnitFramework/framework/Is.cs
jnm2: @Dreamescaper I've seen variations on that, yes. What do you think could be improved?

jnm2:
Your contribution is greatly appreciated! 🎉

PR #2711, opened 2018-02-14
NUnitLite: Add support for --nocolor option

hazzik:


jnm2:
Thanks @hazzik! Is there any way we can get a test guaranteeing this behavior?

hazzik:
@jnm2 I was thinking about this, but I'm not experienced with the NUnit's internal tests suit. Also, I was unable to find the suitable test suite to put this additional test into. Any suggestions welcome.

PR #2725, opened 2018-02-27
Enable 'strict' compilation flag

sharwell:
This flag turns on a few correctness warnings in the C# compiler which could not be enabled by default due to backwards compatibility requirements.

💡 Please do not rebase or squash this pull request on merge.

### DIFF
--- a/src/NUnitFramework/Common.props
+++ b/src/NUnitFramework/Common.props
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <LangVersion>6</LangVersion>
+    <Features>strict</Features>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>


rprouse:
@sharwell I don't disagree with the idea, but does this cause additional warnings when building? If so and if they are too numerous, we might like to fix the warnings along with this change.

sharwell:
@rprouse The project uses `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>`, so I was assuming the build would fail if there were any.

rprouse:
@jnm2 or @ChrisMaddock can you give this a quick second review?

sharwell:
@jnm2 Search for references of `FeatureStrictEnabled`:

http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp/Compilation/CSharpCompilation.cs,045da13303eb1a22

jnm2:
Here we go:
`https://github.com/dotnet/roslyn/issues/7086`
`https://github.com/dotnet/roslyn/issues/13223`

jnm2:
@sharwell Interesting, thanks. What prompted you to want to enable strict mode?

sharwell:
I would summarize it as "Disables intentional deviations in the compiler from the language specification that preserve the behavior of bugs in the legacy compiler".

sharwell:
> What prompted you to want to enable strict mode?

I was comparing your Common.props with a Directory.Build.props I have in another project, and noticed the omission.

rprouse:
Thanks Sam. Running into you twice in one day. What are the odds 😉

Issue #2579, opened 2017-11-28
AppVeyor Test Failures under .NET 3.5


rprouse:
These test failures under .NET 3.5 are getting consistent enough that we are getting few passing builds and we had [one build time out at 60 minutes](https://ci.appveyor.com/project/CharliePoole/nunit/build/3.10.0-ci-04854-pr-2380/messages) in the .NET 3.5 tests.

In PR #2380 I am going to `#if net35` and increase the timeout as a temporary workaround, but we should look at this. I am curious why we are having so many .NET 3.5 threading issues, but no other target.

```
1) Failed : NUnit.Framework.Internal.Execution.EventQueueTests+SetWaitHandle_Enqueue_AsynchronousTest.SetWaitHandle_Enqueue_Asynchronous
Test exceeded Timeout value of 1000ms
```

```
1) Failed : NUnit.Framework.Internal.Execution.EventQueueTests+DequeueBlocking_StopTest.DequeueBlocking_Stop
Test exceeded Timeout value of 1000ms
```

```
7) Cancelled : NUnit.Framework.Internal.Execution.EventQueueTests+SetWaitHandle_Enqueue_AsynchronousTest.SetWaitHandle_Enqueue_Asynchronous
Test cancelled by user
   at NUnit.Framework.Internal.Reflect.InvokeMethod(MethodInfo method, Object fixture, Object[] args) in C:\projects\nunit\src\NUnitFramework\framework\Internal\Reflect.cs:line 232
   at NUnit.Framework.Internal.Commands.TestMethodCommand.RunNonAsyncTestMethod(TestExecutionContext context) in C:\projects\nunit\src\NUnitFramework\framework\Internal\Commands\TestMethodCommand.cs:line 109
   at NUnit.Framework.Internal.Commands.TestMethodCommand.Execute(TestExecutionContext context) in C:\projects\nunit\src\NUnitFramework\framework\Internal\Commands\TestMethodCommand.cs:line 64
   at NUnit.Framework.Internal.Commands.BeforeTestCommand.Execute(TestExecutionContext context) in C:\projects\nunit\src\NUnitFramework\framework\Internal\Commands\BeforeTestCommand.cs:line 50
   at NUnit.Framework.Internal.Commands.BeforeAndAfterTestCommand.Execute(TestExecutionContext context) in C:\projects\nunit\src\NUnitFramework\framework\Internal\Commands\BeforeAndAfterTestCommand.cs:line 59
```

```
[...]

PR #2726, opened 2018-02-27
Replace the ConcurrentQueue and SpinWait compatibility classes

rprouse:
This is intended to address #2579 by replacing the thread related compatibility classes with the latest copies from Mono. I don't say fix above because I would like to leave #2579 open until we have run in CI for a few weeks without failure.

As commented in #2579, I managed to reproduce some of the errors by running the .NET 3.5 tests at the same time in multiple console windows. After making these changes, I have not been able to reproduce. I will rebuild the AppVeyor build a few times to double check.

My main concern with these changes are licensing. The source originally came from the .NET Framework reference source (via Mono) which is available on GitHub at https://github.com/Microsoft/referencesource. The readme there states that all of the source is now under MIT license unless otherwise noted in the headers of the files. The problem is, as far as I can tell, none of the headers in any of those files say MIT license, they all say Copyright Microsoft. My thinking is that if the source wasn't MIT as the root license and README say, then it would explicitly say so. To back that up, Mono is using the source and Mono is MIT. I also took my copies from Mono.

@jnm2 you've been tapped into the open sourcing of .NET more than I have. What have you read or seen? I can also contact the .NET Foundation.

Back to technical issues, I did make a couple of small changes to the files. I will add inline comments for those.

### DIFF
--- a/src/NUnitFramework/framework/Compatibility/System.Collections.Concurrent/ConcurrentQueue.cs
+++ b/src/NUnitFramework/framework/Compatibility/System.Collections.Concurrent/ConcurrentQueue.cs
@@ -1,32 +1,31 @@
+#if NET20 || NET35
+#pragma warning disable 0420
+
+// ==++==
+//
+//   Copyright (c) Microsoft Corporation.  All rights reserved.
+//
+// ==--==
+// =+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+=+
+//
 // ConcurrentQueue.cs
 //
-// Copyright (c) 2008 Jérémie "Garuma" Laval
+// <OWNER>Microsoft</OWNER>
 //
-// Permission is hereby granted, free of charge, to any person obtaining a copy
-// of this software and associated documentation files (the "Software"), to deal
-// in the Software without restriction, including without limitation the rights
-// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-// copies of the Software, and to permit persons to whom the Software is
-// furnished to do so, subject to the following conditions:
+// A lock-free, concurrent queue primitive, and its associated debugger view type.
 //
-// The above copyright notice and this permission notice shall be included in
-// all copies or substantial portions of the Software.
-//
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
-// THE SOFTWARE.
-//
-//
-#if NET20 || NET35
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/framework/Compatibility/System.Threading/SpinWait.cs
```diff
@@ -1,139 +1,363 @@
...
-#else
-                    Thread.Yield ();
-#endif
-                else
-                    // Multi-CPU system might be hyper-threaded, let other thread run
-                    Thread.SpinWait (Math.Min (ntime, maxTime) << 1);
-			}
-		}
+                }
+                else // if ((yieldsSoFar % SLEEP_0_EVERY_HOW_MANY_TIMES) == (SLEEP_0_EVERY_HOW_MANY_TIMES - 1))
```
rprouse: This is the only change of substance. `Thread.Yield()` isn't available in .NET 2.0/3.5. The original code used `Platform.Yield()` for those platforms, but that code is not available in Mono or in the reference source. The comments above say that yielding is preferable, but it still calls `Sleep(0)` for correctness, so I just made that the default.

src/NUnitFramework/framework/Compatibility/System.Collections.Concurrent/ConcurrentQueue.cs
```diff
@@ -36,31 +35,64 @@ namespace System.Collections.Concurrent
...
-
-		Node head = new Node ();
-		Node tail;
-		int count;
+    [ComVisible(false)]
+    [DebuggerDisplay("Count = {Count}")]
+    [DebuggerTypeProxy(typeof(SystemCollectionsConcurrent_ProducerConsumerCollectionDebugView<>))]
+    [HostProtection(Synchronization = true, ExternalThreading = true)]
+    [Serializable]
+    public class ConcurrentQueue<T> : IProducerConsumerCollection<T>//, IReadOnlyCollection<T>
```
rprouse: Second change from Mono, we don't use the `IReadOnlyCollection<T>` interface, so I didn't bother to bring it across.

jnm2:
The little I know is that Mono wanted to use the reference source for years and a few years ago they were finally legally allowed to. If Mono followed their legal responsibilities, I'd think all the copyright messages you need to preserve are already in the files. in the Mono repo. I don't know whether or not we need to add anything to our COPYRIGHT file.

I'm very curious if the latest (MIT) sources from corefx resolve the issues as well? I know they've hit a lot of these foundational implementations really hard for performance last year.

rprouse:
I originally tried with the CoreFx source, but it quickly spread out to way too much code to import including `PlatformHelpers`, `Interlocked.MemoryBarrier()`, and many others. It looked like too big a change.

Speaking of `PlatformHelpers`, one other change I made was to comment out the `Contract` code which are essentially `Debug.Asserts`.

rprouse:
@jnm2 that is also my understanding of the Mono copyright issues and I why I believe it is okay to use the code, but as always, IANAL. 😄

rprouse:
I kept an AppVeyor browser tab open today and kept kicking of builds of this PR. There are at least 10 successful builds and no failures so far. That isn't conclusive, but it is a good sign 👍

rprouse:
@CharliePoole what do you think about the copyright issue? Do you think that Microsoft has stated their reference source is MiT and the fact that Mono is using it as such is good enough despite the fact Microsoft hasn't updated any of their copyright headers?

rprouse:
@oznetmaster, this PR is proposing to pull in the latest version of `ConcurrentQueue` from the Mono project. We have been having numerous failed test runs and occasionally hung test runs in .NET 3.5. Looking through the code in question, the only differences are the `ConcrrentQueue`, `SpinWait` classes that were pulled in for 2.0 and 3.5 from Mono and the fact that we use `ReaderWriterLock` instead of `ReaderWriterLockSlim`. Based on that observation, I went back to Mono to see if they had updates or fixes to either class and found that they had fully replaced them with the versions from the reference source.

CharliePoole:
Saying who holds the copyright doesn't contradict the license. Some folks would say they ought not to include "All Rights Reserved" but I don't think it's intended to mean anything different and IANAL anyway.

What they are apparently referring to are files like https://github.com/Microsoft/referencesource/blob/4fe4349175f4c5091d972a7e56ea12012f1e7170/System.Core/Microsoft/Scripting/Utils/IRuntimeVariables.cs where a completely different license is specified.

So I don't think there's a problem with the files you want to use unless they have a different license called out. In that case, we would have to think about how to handle it.

PR #2739, opened 2018-03-10
Include type in invalid PropertyConstraint error message.

### LINKED ISSUE #2405: Improve PropertyConstraint error output
I normally get far too distracted by interesting but niche things. This is one that bites me almost daily. 

I use PropertyConstraint a lot - this is the error message that's currently shown when the property can't be found on the object.

image

I hit this regularly - either because my constraint is not actually on the object I think it is, because I go the property name wrong, or because I'm trying to look at an [explicit interface implementation](https://github.com/nunit/nunit/issues/1394).

I want to improve this so it provides a proper error message rather than stack trace. What I'm interested in seeing is:

1. The property name I was looking for. (I almost always use `nameof()` here - maybe the string passed in wasn't what I thought)
1. The type of the object I'm looking for the property on.

I'll PR this at some point - any other suggestions before I do? I considered listing all available property names - but decided against it as it could get noisy quickly.

ChrisMaddock:
Fixes #2405 

We talked a lot about this issue, but I think got a bit lost in the semantics. This simple string change is all I needed to fix my issue of not being sure what type I am testing against.

Personally, I have no concern as to whether this is reported as an error or failure - although think the current behaviour here is perhaps 'more semantically accurate' with our current architecture, so don't feel the need to rewrite anything! 😄

### DIFF
--- a/src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
@@ -66,7 +66,7 @@ namespace NUnit.Framework.Constraints
 
             // TODO: Use an error result here
             if (property == null)
-                throw new ArgumentException(string.Format("Property {0} was not found", name), "name");
+                throw new ArgumentException($"Property {name} was not found on {actualType}.", "name");
 
             propValue = property.GetValue(actual, null);
             var baseResult = BaseConstraint.ApplyTo(propValue);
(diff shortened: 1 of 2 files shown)


ChrisMaddock:
Restarting Appveyor - it timed out

ChrisMaddock:
@rprouse @CharliePoole - Do either of you have access to prod the stuck `appveyor/pr` build? I normally can, but for some reason don't have the button on that one.

image

Issue #325, opened 2014-11-02
Add RegexConstraint to Compact Framework build


CharliePoole:
The constraint is currently #if'd out of NETCF builds. It should be included, as well as its tests and the associated "Matches" syntax.

PR #2740, opened 2018-03-10
Add NUnitCallContext to clean-up CallContext around FrameworkController actions

### LINKED ISSUE #2614: TestExecutionContext.CurrentContext is saved in Remoting CallContext between test runs
Discovered from https://github.com/nunit/nunit-console/issues/325.

TestExecutionContext.CurrentContext is currently saved in `Remoting.Messaging.CallContext` as below:

```
        private static readonly string CONTEXT_KEY = "NUnit.Framework.TestContext";

        public static TestExecutionContext CurrentContext
        {
            [SecuritySafeCritical]
            get
            {
                var context = CallContext.GetData(CONTEXT_KEY) as TestExecutionContext;

                if (context == null)
                {
                    context = new AdhocContext();
                    CallContext.SetData(CONTEXT_KEY, context);
                }

                return context;
            }

            [SecuritySafeCritical]
            private set
            {
                if (value == null)
                    CallContext.FreeNamedDataSlot(CONTEXT_KEY);
                else
                    CallContext.SetData(CONTEXT_KEY, value);
            }
        }
```

It appears that this CallContext is still accessible past the end of a single run, and across app domains. The result of this is that when the Console runs two tests assemblies sequentially (in the same process), the second test suite attempts to access the final TestExecutionContext of the previous assembly, and triggers an `AppDomainUnloadedException`.

I've tested this theory by including the below line at the end of `RunTestsAction` - i.e. as the last thing done by the framework in a test run. This appeared to clear the CallContext, and mean a new TestExecutionContext was created for the second assembly, as we'd want.

`CallContext.FreeNamedDataSlot("NUnit.Framework.TestContext");`

I know absolutely nothing about CallContext's, and the docs seem 'sparse'! From what I can see, `TestExecutionContext` implements `ILogicalThreadAffinative` - which is specifically to allow it to be accessed across AppDomains. What's the purpose of this, is it to permit user-created AppDomains? 

How should we handle this issue? I don't think the framework currently has any sort of 'clean-up' - is that the correct solution here?

ChrisMaddock:
Fixes #2614. This is intended to ensure the Remoting.CallContext is cleaned up correctly after each call to a framework action. The CallContext should be cleared after each run (or other framework call) - remembering the fact that the assembly may be loaded just the once, and multiple calls to it made. This is to ensure that TestExecutionContext objects which are stored in the CallContext can not leak from one run to the next.

Is this the correct level to implement this fix? I've had a look into this - the Engine calls the framework via the `FrameworkController.*Action` methods where I've added the current fix, however NUnitLite and the Xamarin runner both use the NUnitTestAssemblyRunner, which is a level lower. How does this fit into our intended architecture - should we be applying this fix in the NUnitTestAssemblyRunner - or should these two runners not be using the NUnitTestAssembly runner, and instead be using the FrameworkController class? Our [docs](https://github.com/nunit/docs/wiki/Framework-Api) suggest things should be going through the FrameworkController class. (This is more a semantics issue, as the Xamarin runner will never run full framework assemblies, where CallContext is relevant, and NUnitLite rarely runs more that one assembly in the same process.)

@yaakov-h - Would you grab the framework package from the appveyor build, and check this fix does solve your issue reported in https://github.com/nunit/nunit-console/issues/325, and we've diagnosed correctly?

### DIFF
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -473,15 +473,21 @@ namespace NUnit.Framework.Api
 
 #endregion
 
+#region Nested Action Classes
+
 #region TestContollerAction
 
         /// <summary>
         /// FrameworkControllerAction is the base class for all actions
         /// performed against a FrameworkController.
         /// </summary>
-        public abstract class FrameworkControllerAction : LongLivedMarshalByRefObject { }
+        public abstract class FrameworkControllerAction : LongLivedMarshalByRefObject
+        {
+        }
 
-        #region LoadTestsAction
+#endregion
+
+#region LoadTestsAction
 
         /// <summary>
         /// LoadTestsAction loads a test into the FrameworkController
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -483,6 +483,22 @@ namespace NUnit.Framework.Api
         /// </summary>
         public abstract class FrameworkControllerAction : LongLivedMarshalByRefObject
         {
+            /// <summary>
+            /// Wrap in NUnitCallContext executes the action within an NUnitCallContext
+            /// which ensures System.Runtime.Remoting.Messaging.CallContext is cleaned up
+            /// suitably at the end of the test run. This method only has an effect running
+            /// the full .NET Framework.
+            /// </summary>
+            /// <param name="action"></param>
+            protected void WrapInNUnitCallContext(Action action)
+            {
+#if !(NET20 || NET35 || NET40 || NET45)
+                action();
+#else
+                using (new NUnitCallContext())
+                    action();
+#endif
+            }
         }
 
 #endregion
@@ -501,7 +517,7 @@ namespace NUnit.Framework.Api
             /// <param name="handler">The callback handler.</param>
             public LoadTestsAction(FrameworkController controller, object handler)
             {
-                controller.LoadTests((ICallbackEventHandler)handler);
+                WrapInNUnitCallContext(() => controller.LoadTests((ICallbackEventHandler)handler));
             }
         }
 
@@ -522,7 +538,7 @@ namespace NUnit.Framework.Api
             /// <param name="handler">The callback handler.</param>
             public ExploreTestsAction(FrameworkController controller, string filter, object handler)
             {
-                controller.ExploreTests((ICallbackEventHandler)handler, filter);
+                WrapInNUnitCallContext(() => controller.ExploreTests((ICallbackEventHandler)handler, filter));
             }
         }
 
@@ -544,7 +560,7 @@ namespace NUnit.Framework.Api
             /// <param name="handler">A callback handler used to report results</param>
             public CountTestsAction(FrameworkController controller, string filter, object handler)
             {
-                controller.CountTests((ICallbackEventHandler)handler, filter);
+                WrapInNUnitCallContext(() => controller.CountTests((ICallbackEventHandler)handler, filter));
             }
         }
 
@@ -565,7 +581,7 @@ namespace NUnit.Framework.Api
             /// <param name="handler">A callback handler used to report results</param>
             public RunTestsAction(FrameworkController controller, string filter, object handler)
             {
-                controller.RunTests((ICallbackEventHandler)handler, filter);
+                WrapInNUnitCallContext(() => controller.RunTests((ICallbackEventHandler)handler, filter));
             }
         }
 
@@ -586,7 +602,7 @@ namespace NUnit.Framework.Api
             /// <param name="handler">A callback handler used to report results</param>
             public RunAsyncAction(FrameworkController controller, string filter, object handler)
             {
-                controller.RunAsync((ICallbackEventHandler)handler, filter);
+                WrapInNUnitCallContext(() => controller.RunAsync((ICallbackEventHandler)handler, filter));
             }
         }
 
@@ -609,7 +625,7 @@ namespace NUnit.Framework.Api
             /// <remarks>A forced stop will cause threads and processes to be killed as needed.</remarks>
             public StopRunAction(FrameworkController controller, bool force, object handler)
             {
-                controller.StopRun((ICallbackEventHandler)handler, force);
+                WrapInNUnitCallContext(() => controller.StopRun((ICallbackEventHandler)handler, force));
             }
         }
 
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -473,37 +473,15 @@ namespace NUnit.Framework.Api
 
 #endregion
 
-#region Nested Action Classes
-
 #region TestContollerAction
 
         /// <summary>
         /// FrameworkControllerAction is the base class for all actions
         /// performed against a FrameworkController.
         /// </summary>
-        public abstract class FrameworkControllerAction : LongLivedMarshalByRefObject
-        {
-            /// <summary>
-            /// Wrap in NUnitCallContext executes the action within an NUnitCallContext
-            /// which ensures System.Runtime.Remoting.Messaging.CallContext is cleaned up
-            /// suitably at the end of the test run. This method only has an effect running
-            /// the full .NET Framework.
-            /// </summary>
-            /// <param name="action"></param>
-            protected void WrapInNUnitCallContext(Action action)
-            {
-#if !(NET20 || NET35 || NET40 || NET45)
-                action();
-#else
-                using (new NUnitCallContext())
-                    action();
-#endif
-            }
-        }
+        public abstract class FrameworkControllerAction : LongLivedMarshalByRefObject { }
 
-#endregion
-
-#region LoadTestsAction
+        #region LoadTestsAction
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


src/NUnitFramework/tests/Api/FrameworkControllerCallContextTests.cs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        private static IEnumerable<Action> FrameworkActions
+        {
+            get
+            {
+                var settings = new Dictionary<string, object>
+                {
+                    { FrameworkPackageSettings.RunOnMainThread, true }
+                };
```
ChrisMaddock: This is to ensure we're testing worse-case-scenario. The original issue was found when `TestExecutionContext` was created in a TestCaseSource, in test exploring code which always runs on the main thread - however, it could be that the entire test run is carried out on the main thread - and we should be protecting against that case too.

src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
```diff
@@ -591,25 +592,6 @@ private static string GetSkipReason(TNode result)
             return propNode == null ? null : propNode.Attributes["value"];
         }
 
-#endregion
-
-#region Nested Callback Class
-
```
ChrisMaddock: This has just been extracted to a TestUtility to share the code.

ChrisMaddock:
Something in the test here is leaving temporary files in the work directory, which I'll need to look into. Otherwise this is ready for review.

src/NUnitFramework/framework/Api/FrameworkController.cs
```diff
@@ -481,6 +481,22 @@ private static void AddDictionaryEntries(TNode settingNode, IDictionary entries)
         /// </summary>
         public abstract class FrameworkControllerAction : LongLivedMarshalByRefObject
         {
+            /// <summary>
+            /// Wrap in NUnitCallContext executes the action within an NUnitCallContext
+            /// which ensures System.Runtime.Remoting.Messaging.CallContext is cleaned up
+            /// suitably at the end of the test run. This method only has an effect running
+            /// the full .NET Framework.
+            /// </summary>
+            /// <param name="action"></param>
```
jnm2: ❔ Can you remove the `<param>` xml and use cref for `NUnitCallContext`? Maybe the summary could be started, `Executes`? Maybe `the` after `ensures`, and cref instead of `System.Runtime.Remoting.Messaging.CallContext`?

src/NUnitFramework/framework/Internal/NUnitCallContext.cs
```diff
@@ -0,0 +1,61 @@
...
+        private readonly object _oldContext;
+        public const string TestExecutionContextKey = "NUnit.Framework.TestExecutionContext";
+
+        // This method invokes security critical members on the 'System.Runtime.Remoting.Messaging.CallContext' class.
+        // Callers of this method have no influence on how these methods are used so we define a 'SecuritySafeCriticalAttribute'
+        // rather than a 'SecurityCriticalAttribute' to enable use by security transparent callers.
+        [SecuritySafeCritical]
+        public NUnitCallContext()
+        {
+            _oldContext = CallContext.GetData("NUnit.Framework.TestExecutionContext") as TestExecutionContext;
```
jnm2: Is the safe cast still necessary?

src/NUnitFramework/tests/TestUtilities/CallbackEventHandler.cs
```diff
@@ -0,0 +1,40 @@
+// ***********************************************************************
+// Copyright (c) 2014 Charlie Poole, Rob Prouse
```
jnm2: 2018

src/NUnitFramework/framework/Api/FrameworkController.cs
ChrisMaddock: Sure - good shout. 😊

src/NUnitFramework/tests/TestUtilities/CallbackEventHandler.cs
ChrisMaddock: This code was just moved, so I took the copyright date from the original file. 😊

CharliePoole:
Can you hold on while I review code? I don't understand what happened to the older checks that used to clear the call context. This PR doesn't replace them, so do we have multiple levels of checks or did somebody remove them earlier?

ChrisMaddock:
Charlie - sorry, just switched the labels while you were posting! 😆Sure - noted!

I couldn’t see any existing checks myself - but am not too sure at what level in the architecture to expect them.

CharliePoole:
@ChrisMaddock I can guarantee they were once present, but only on the test run entries, which were the only ones that created a TestExecutionContext until we added it to the load / discover entries.

I'm at the cafe now, I'll look at the repository when I get home.

jnm2:
Ah, here? https://github.com/nunit/nunit/blob/f108ab9ca6b5881b7aaad7bc858291cce7d8beed/src/NUnitFramework/framework/Internal/TestExecutionContext.cs#L220

jnm2:
Ah, are you talking about this?

https://github.com/nunit/nunit/blob/f108ab9ca6b5881b7aaad7bc858291cce7d8beed/src/NUnitFramework/framework/Internal/TestExecutionContext.cs#L220

Used by:

ttps://github.com/nunit/nunit/blob/f108ab9ca6b5881b7aaad7bc858291cce7d8beed/src/NUnitFramework/framework/Internal/TestExecutionContext.cs#L533-L569

CharliePoole:
@jnm2 No... Setting CurrentContext to null is __how__ you are supposed to free the slot, so that other classes don't know about the key. However, I'm looking for the code that actually does the freeing when about to return from a remoting call - the only time it's actually necessary.

I'm home now... let me check it out.

CharliePoole:
Interesting. It's all gone. I didn't take time to try to figure out when or by whom it was removed, but I think I understand why. 

The interface I created to clear the `TestExecutionContext` from the `CallContext` was that you set

```C#
TestExecutionContext.CurrentContext = null;
```

I guess that was a bit too "clever". People probably didn't realize why the context was being set to null and it got removed. I should have used something like a `ClearCurrentContext` method, which would be harder to miss.

Anyway... comments on the current PR...

It bothers me that code outside `TestExecutionContext` will now need to know how the `CurrentContext` is stored in order to free it. IIRC, it was previously done at the point where the initial context is created, in the dispatcher, by use of the set-to-null trick. IMO, we should continue to do it there, by calling some new method on `TestExecutionContext`. I like having a disposable class that does it automatically at the end of a `using` block. That could still be done in the dispatcher. NOTE: It has to be done in all dispatchers. 

This would have the added effect of handling with all runners, no matter how they interact with the framework.

mikkelbu:
@CharliePoole I don't know much about `CurrentContext` or the history behind this, but I did a little digging in the code base and found https://github.com/nunit/nunit/commit/76588be886cd2d65da836959f1a4efe13abba7a1#diff-11a1802be2fa3b7e8bb8b478b2f00c99L296 where we removed `TestExecutionContext.ClearCurrentContext` from the code base.

CharliePoole:
Looks like I didn't see the side effect of removing it... probably due to the obscurity of the API. We should probably have such a method that actually contains the code to free the data slot rather than relying on the ugly side effect of assigning null to CurrentContext.

ChrisMaddock:
Sorry, just getting back to this.

@CharliePoole - agree with your thoughts that `FrameworkController` needing to know about the CallContext isn't ideal. I'm not completely convinced that doing this in the Dispatcher's would be preferable however - those classes are still a level outside from the execution context, and in feels a little bit like making the situation worse - by requiring this context clearing knowledge in every future implementation of `IWorkItemDispatcher`.

I haven't tried writing this in code yet - do you think it would be feasible to make `TestExecutionContext` an IDisposable, and have it clear the context up after itself? I'm not totally clear on the lifecycle of this class.

The old solution was in the FrameworkController. Another option would be moving the clearing a level lower to NUnitTestAssemblyRunner, which would ensure the context is cleared by all our current runners - but won't spread the check out across multiple implementations of IWorkItemDispatcher.

yaakov-h:
Thanks; I'll try run a functional test tomorrow.

yaakov-h:
With the binaries from AppVeyor, this PR does solve the original issue reported in #325.

🎉

ChrisMaddock:
Great, thanks @yaakov-h!

@CharliePoole - any thoughts on the ideas above? 😊

CharliePoole:
I like the idea of moving it to NUnitTestAssemblyRunner if that can be managed cleanly.

The big problem I see with putting the implementation in TestExecutionContext is that we want to free any TEC That we are no longer using but we __don't__ want to free the slot unless the top-level context is being eliminated. Ideally, the code that frees the slot should be in the same class as the code that created the slot in the first place, but that doesn't seem to work very neatly.

ChrisMaddock:
Thanks Charlie. I’ll revisit this some point soon.

jnm2:
Is this PR still in need of a review? It's tagged with two tags.

ChrisMaddock:
Nope, sorry, this ones back with me now.

ChrisMaddock:
Ok, this has sat here for too long!

As per review comments, I've pushed the CallContext knowledge down into NUnitTestAssemblyRunner, which does seem like a good fit. 

This is ready for another review. Apologies it's been stood a long time, but hopefully the top post is enough to remind people what the issue is. 🙂

PR #2741, opened 2018-03-10
Dogfood NUnit Console Socket Exception fix

ChrisMaddock:
@BlythMeister added a temporary fix for the SocketExceptions we've been seeing in the NUnit Console in https://github.com/nunit/nunit-console/pull/372.

Can we dogfood that? This will hopefully also make our CI more stable.


ChrisMaddock:
Until we next update the console in build.cake as we normally would do Was my plan! Unless we discover any issues which would need us to change that.

PR #2744, opened 2018-03-10
Exposed NUnitEqualityComparer ctor and obsoleted Default

### LINKED ISSUE #2591: NUnitEqualityComparer.Default should be replaced with new NUnitEqualityComparer()
In https://github.com/nunit/nunit/pull/2290, we made the constructor internal ~~and introduced a Default static property which returns a new instance every time~~. I failed to notice this when reviewing that PR. 

Edit: Thanks for the correction @mikkelbu. The Default property has been in NUnitEqualityComparer since v2.5.1!

I believe that static properties should return the same instance every time or at least *semantically be able* to return the same instance every time. If the returned instance is immutable, it should be a static property. If it's mutable, it must create a new instance every time and therefore should be a method. Rather than having a static method delegate to the default constructor, I'd prefer re-exposing the default constructor.

I would like to propose the following:

```diff
namespace NUnit.Framework.Constraints
{
    public class NUnitEqualityComparer
    {
+       public NUnitEqualityComparer();

+       [Obsolete("Deprecated. Use the default constructor instead.", error: true)]
        public static NUnitEqualityComparer Default { get; }
    }
}
```

Should we make the obsoletion a warning or error? I prefer error because it helpfully points people to the necessary refactor while flushing out all uses but I could go either way.

jnm2:
Closes #2591

### DIFF
--- a/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
@@ -33,7 +33,7 @@ namespace NUnit.Framework.Constraints
     public class AnyOfConstraint : Constraint
     {
         private readonly object[] _expected;
-        private readonly NUnitEqualityComparer _comparer = NUnitEqualityComparer.Default;
+        private readonly NUnitEqualityComparer _comparer = new NUnitEqualityComparer();
 
         /// <summary>
         /// Construct a <see cref="AnyOfConstraint"/>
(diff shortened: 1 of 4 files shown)


jnm2:
Good thought. I can take care of that.

rprouse:
Looks good to me too. Just waiting on CI.

rprouse:
Travis failure was just an `apt-get` failure. Restarted.

jnm2:
Docs updated. https://github.com/nunit/docs/wiki/Breaking-Changes/376972f6797a4c76424e16986fed247eb817a296

jnm2:
@rprouse You know, I think there were more breaking changes in 3.7–3.9. Do you mind if I do an API diff and fill out https://github.com/nunit/docs/wiki/Breaking-Changes? I can do a PR against my fork of the wiki before copying it over to the docs wiki.

rprouse:
Sure, go ahead. I would be interested in seeing what you find.

PR #2745, opened 2018-03-10
Use NUnit equality when the element type is not sealed

### LINKED ISSUE #2627: Breaking change in CollectionAssert.AllItemsAreUnique with NUnit 3.9
After upgrading to NUnit Framework 3.9 we have a failing test that checks if all items of an object collection are unique. Example:
```
var collection = new Collection<object> {42, null, 42f};
Assert.That(
    () => CollectionAssert.AllItemsAreUnique(collection),
    Throws.TypeOf<AssertionException>());
```
A float value of `42f` now is considered to be different to an integer value `42`.
Is this intentional or an undesirable side effect of maybe #2551?

mikkelbu:
Fixes #2627


mikkelbu:
Restarting Travis

PR #2748, opened 2018-03-12
Enable netcoreapp1.1 tests on non-Windows platforms

### LINKED ISSUE #2611: Enable .NET Standard 1.6 tests on non-Windows
This will require installing two versions of the .NET Core SDK on Travis which may not be straightforward.
We need both the latest 1.1.x (because our tests target netcoreapp1.1) and the latest 2.0.x. I don't expect any 1.6 tests to fail once https://github.com/nunit/nunit/pull/2609 is merged, but since we use preprocessor and a whole different set of BCL libraries, there's always the chance.

The alternative would be to stop targeting netcoreapp1.1 in any test projects, but that would leave us without a way to test the netstandard1.6 NUnit DLL unless we added additional test csproj files which target netcoreapp2.0 and reference the netstandard1.6 framework DLL directly rather than referencing it via the framework project reference.

jnm2:
Fixes #2611. (Not priority for release; I verified last week that no tests failed.)

This is the simplest change I know of to enable these tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: csharp
 sudo: false
 
-# Need 2.*, should keep up to date with security releases
+# Should keep up to date with security releases
 dotnet: 2.0.3
 
 # Mono: Need 5.2+ for the included MSBuild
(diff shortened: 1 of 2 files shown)


jnm2:
Got a hang in Travis. Mac job succeeded, the Linux job hung doing TestNetStandard20:

```
=> NUnit.Framework.TextOutputTests.TestContextWriteLine_WritesToResult
Should go to the result


No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

I restarted. I can't think of any way it would be related to this PR.

PR #2752, opened 2018-03-13
Merge release 3.10

rprouse:
Merges the changes from the 3.10 release and updates the version to 3.11 for ongoing development.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,69 @@
+### NUnit 3.10 - March 12, 2018
+
+This release adds a .NET Standard 2.0 version of the framework which re-enables
+most of the features that have been missing in our earlier .NET Standard builds
+like parallelism, timeouts, directory and path based asserts, etc. It also contains
+numerous bug fixes and smaller enhancements. We've improved our XML docs,
+fixed performance issues and added more detail to Multiple Asserts.
+
+#### Issues Resolved
+
+ * 1373 Setting with a null value
+ * 1382 Use array argument contents in name of parameterized tests rather than just array type.
+ * 1578 TestContext.CurrentTest exposes too much internal info
+ * 1678 Result Message: OneTimeSetUp: Category name must not contain ',', '!', '+' or '-'
+ * 1944 Removing Compact Framework workarounds
+ * 1958 System.Reflection.TargetInvocationException after run finished
+ * 2033 Nameof refactor
+ * 2202 Best practices for XML doc comments
+ * 2325 Retry attribute doesn't retry the test.
+ * 2331 Repo does not build in VS without running `build -t build` first
+ * 2405 Improve PropertyConstraint error output
+ * 2421 Publishing symbols with releases
+ * 2494 CollectionAssert.AllItemsAreUnique() very slow
+ * 2515 Retarget Solution to use the New CSPROJ Format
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -1,3 +1,80 @@
+### NUnit 3.10 - March 12, 2018
+
+This release adds a .NET Standard 2.0 version of the framework which re-enables
+most of the features that have been missing in our earlier .NET Standard builds
+like parallelism, timeouts, directory and path based asserts, etc. It also contains
+numerous bug fixes and smaller enhancements. We've improved our XML docs,
+fixed performance issues and added more detail to Multiple Asserts.
+
+This release also contains source linked PDB files allowing developers to debug
```
jnm2: Technically GitLink is not the same mechanism as Source Link (sourcelink.json in the PDB) so that might cause people to think the wrong thing.

CHANGES.md
```diff
@@ -1,3 +1,80 @@
...
+into the NUnit Framework. This allows you to track down errors or see how the
+framework works.
+
+In order to support the .NET Standard 2.0 version, the NUnit project switched to
+the new CSPROJ format and now requires Visual Studio 2017 to compile. This only
+effects people contributing to the project. NUnit still supports building and
+compiling your tests in older .NET IDE's and NUnit still supports older versions
```
jnm2: Double checked the rule. Usually we only use an apostrophe with an acronym if there is a period after every letter (I.D.E.'s) and otherwise only use an apostrophe after an acronym to make it possessive (IDE's); plural would just be IDEs. (English, right? 😐)

src/NUnitFramework/FrameworkVersion.cs
```diff
@@ -26,5 +26,5 @@
 //
 // Current version for the NUnit Framework
 //
-[assembly: AssemblyVersion("3.10.0.0")]
-[assembly: AssemblyFileVersion("3.10.0.0")]
+[assembly: AssemblyVersion("3.11.0.0")]
+[assembly: AssemblyFileVersion("3.11.0.0")]
```
jnm2: Sometime can we start doing:

```cs
[assembly: AssemblyVersion("3.11.0")]
[assembly: AssemblyInformationalVersion("3.11.0")]
```

That will result in a .NET assembly version of 3.11.0.0 and Windows file version of 3.11.0.0 but a Windows product version of 3.11.0. (Can be seen in file properties.)

At work, we also have the build script update the file in CI before building to this effect:

```cs
[assembly: AssemblyInformationalVersion("3.11.0-ci-12345")]
```

Since the Windows product version is an arbitrary string, this works great. Some people also put `+{commit hash}` at the end.

CHANGES.md
rprouse: I am okay to update if the user behavior is likely to be different or confusing. What text would you suggest? Source linked with GitLink?

CHANGES.md
rprouse: I will update :stuck_out_tongue_winking_eye:

CHANGES.md
jnm2: The terminology I've seen is source indexing and source-indexed PDBs. That's the name of the TFS build task and I see that at https://github.com/GitTools/GitLink.

PR #2754, opened 2018-03-13
Made NuGet .props compatible with older MSBuild

### LINKED ISSUE #2753: NuGet props causes MSBuild < 15 to fail
MSBuild older than v15 will crash because we don't declare an XML namespace:

https://github.com/nunit/nunit/blob/36f193c13e86acab35f5c685f60a32f1ed71e8a5/nuget/framework/NUnit.props#L1

This is causing @marcelopontipedra these errors:

image
image

Change was made in https://github.com/nunit/nunit/pull/2594.

I wonder how hard it would be to set up an integration test against MSBuild 4.0.

jnm2:
Fixes #2753 

Per https://msdn.microsoft.com/library/dd576348.aspx

⚠ Based to merge into the `release` branch.

### DIFF
--- a/nuget/framework/NUnit.props
+++ b/nuget/framework/NUnit.props
@@ -1,4 +1,4 @@
-<Project>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <ItemGroup>
     <ProjectCapability Include="TestContainer" />


jnm2:
Did a quick 'n dirty test which verified that `C:\Windows\Microsoft.NET\Framework\v4.0.30319\MSBuild.exe` is happy with this change.

rprouse:
Looks good. Now I have to figure out how to do a hotfix release off these branches. Please stand by 😄

PR #2767, opened 2018-03-21
[Do not merge] Cross-cutting tests proof-of-concept

jnm2:
While [adding unit tests](https://github.com/nunit/nunit/pull/2765/files#diff-da11b68293e8c077e88bb97c497712e0R49) on NUnitComparer and then [duplicating the data](https://github.com/nunit/nunit/pull/2765/files#diff-60c7f8eb1c14bc5b8d779475070e614fR55) to keep our duplicated data in sync for NUnitEqualityComparer, I wished we could somehow reduce duplication and be a bit more exhaustive at the same time. I imagined what it would look like if I was able to add the four tests you see in the PR in `CharComparisons`.

This idea might come to nothing, but it was fun enough to play with for a bit and I'm interested in your reactions.

I ran into inconsistent async handling a few times last year and my mind kept going to the idea of cross-cutting guarantees to hold all those details consistent across the various APIs NUnit offers. So, if you end up not liking this for comparison tests, maybe you'd not mind something like this when I do the async PR?

### DIFF
--- a/src/NUnitFramework/tests/Crosscutting/CharComparisons.cs
+++ b/src/NUnitFramework/tests/Crosscutting/CharComparisons.cs
@@ -1,0 +1,29 @@
+namespace NUnit.Framework.Crosscutting
+{
+    public sealed class CharComparisons : CrosscuttingComparisonTests
+    {
+        [Test]
+        public void EqualChars()
+        {
+            AssertEquality((char)4, (char)4);
+        }
+
+        [Test]
+        public void CharEqualToInt()
+        {
+            AssertEquality((char)4, 4);
+        }
+
+        [Test]
+        public void CharGreaterThanInt()
+        {
+            AssertFirstIsGreater((char)4, 2);
+        }
+
+        [Test]
+        public void IntGreaterThanChar()
+        {
+            AssertFirstIsGreater(4, (char)2);
+        }
+    }
+}
--- a/src/NUnitFramework/tests/Crosscutting/CrosscuttingComparisonTests.ComparisonBehavior.cs
+++ b/src/NUnitFramework/tests/Crosscutting/CrosscuttingComparisonTests.ComparisonBehavior.cs
@@ -1,0 +1,37 @@
+﻿using System;
+
+namespace NUnit.Framework.Crosscutting
+{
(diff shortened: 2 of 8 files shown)


ChrisMaddock:
The concept, I'm totally keen on - I'd love for us to have more 'exhaustive' testing in this way, rather than the current case by case idea. This also bites us sometimes with chained syntax - I'd love to see us have a more exhaustive way of testing combinations there.

What's going on with the code here, I'm a little lost by, just looking at it on GitHub! Could you explain what's going on here that can't be achieved by building `TestCaseSource` comparisons in this the more traditional NUnit way?

jnm2:
Yes! Just curious, have you had any thoughts towards what some of those chained combinations might look like?

This probably could be done using TestCaseSource pointing to a member in the base class and inverting the flow of data. The benefit would be that you'd get `<Behavior description>` as the parent node and `<API 1>`, `<API 2>`, etc as child nodes which can pass and fail separately. What I really want is for [this code](https://github.com/nunit/nunit/pull/2767/files#diff-362ef3c6bceef1066dc3f574b3bd10e5R26) to be turned into a syntax that builds an `IEnumerable<TestCaseData>`.
Thanks for prompting. I'm going to play with it a bit more and see if I can make it nicer! 😁

ChrisMaddock:
> Yes! Just curious, have you had any thoughts towards what some of those chained combinations might look like?

Absolutely none! 😁

rprouse:
This looks interesting, although I must admit it made my eyes cross a little trying to review on GitHub, I think I'll have to pull the branch on this one to review and understand properly 😕 😄

jnm2:
Sorry about that! Let me refactor per Chris' idea before you spend too much time on that. It's all plumbing except the Char test class and the RunComparisons method, but it'll change shape.

jnm2:
Closing. I settled on an approach I like much better: https://github.com/nunit/nunit/pull/2818/commits/e0223181bbdf7e8888c4fe4522b04b73c9d0666e#diff-f64fee158fbb3c99982be6ce21408c48R38

I hope to use that pattern to deduplicate data and get more rigorous coverage in various places, eventually.

Issue #2774, opened 2018-03-24
SynchronizationContexts and awaiting


jnm2:
Going for a high-level view. I've been talking about finishing off the async PR for too long. I'm going to spend some time in airports/planes tomorrow and do a fresh spike. The PRs can be small and incremental, tackling a section at a time.

I'll be eager for any @nunit/framework-team comments on any part, though I know this is not most people's favorite thought experiment. 🙃 If you get a chance, please especially glance at the ❔s.

### Specially-recognized synchronization contexts

If `SynchronizationContext.Current` is an instance of `WindowsFormsSynchronizationContext` (Windows Forms control has been created) or `DispatcherSynchronizationContext` (WPF), NUnit should not call `.Wait()` on async methods or delegates or there will be a guaranteed deadlock. This is https://github.com/nunit/nunit/issues/2123. Sam ran into this a couple days ago. I've been making my test methods non-async and using [this workaround](https://gist.github.com/jnm2/4643df73fdda0efe1df3a03ad0188c46) for a long time.

### Default synchronization contexts

If `SynchronizationContext.Current` is null or an instance of `SynchronizationContext`, and the current apartment state is STA, we should install our own [`SingleThreadedSynchronizationContext`](https://github.com/nunit/nunit/issues/1200#issuecomment-312120851). Otherwise, awaits will continue on non-STA threads. This is https://github.com/nunit/nunit/issues/1200.

❔ Should we consider always installing a `SingleThreadedSynchronizationContext` so that all awaits in test methods and delegates keep continuing on the original thread, even if it isn't STA? If people don't like this, they can use `.ConfigureAwait(false)`. (You do have a habit of using that, right? 😉)

### Custom synchronization contexts

❔ Should we create a framework extension point so that people can override the waiting logic with custom code? Say, for example, some custom native message loop like certain DirectX libraries ship?
Probably not unless asked, but we would want to do this internally to some extent and make our four SynchronizationContext strategies pluggable.

### Fallthrough

In all other cases, we have no option but to block-wait. If continuations need to be pumped, there's no way NUnit can pump them and the tests will deadlock. If on the other hand they are scheduled on threadpool threads, they'll execute off-thread and finally complete the async task. This is what we currently do in all cases.

### Fallthrough—awaitable spec

This is https://github.com/nunit/nunit/issues/2286.

[...]

PR #2776, opened 2018-03-25
Fixed deadlocks on known synchronization contexts

### LINKED ISSUE #2123: Task.Run inside a test will result in deadlock if a control was created previously
Hi,

I'm not sure whether an issue with NUnit but the following test will always result in a deadlock:

```
[Test]
public async Task Deadlock()
{
    new Control();

    await Task.Run(() => { int x = 42; });
}
```

Version used: **3.6.1**

Thanks,
Costin

jnm2:
Fixes #2123, re: https://github.com/nunit/nunit/issues/2774

The first commit enabled me to quickly test because I was without internet, so build.cake couldn't pull the console runner from NuGet, and VS required my license to be refreshed.

I left some notes in the source about the later general-awaitable PR.

All reviews welcome, as usual. @sharwell, since you're knowledgeable in this area and you hit the bug, would you please review?

### DIFF
--- a/src/NUnitFramework/framework/Internal/AwaitAdapter.cs
+++ b/src/NUnitFramework/framework/Internal/AwaitAdapter.cs
@@ -1,0 +1,114 @@
+// ***********************************************************************
+// Copyright (c) 2018 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+#if ASYNC
+using System;
+using System.Runtime.CompilerServices;
+using System.Security;
+using System.Threading.Tasks;
+
+namespace NUnit.Framework.Internal
+{
+    internal abstract class AwaitAdapter
+    {
+        public abstract bool IsCompleted { get; }
+        public abstract void OnCompleted(Action action);
+        public abstract void BlockUntilCompleted();
+
+        public static AwaitAdapter FromAwaitable(object awaitable)
+        {
+            if (awaitable == null) throw new ArgumentNullException(nameof(awaitable));
+
+            var task = awaitable as Task;
+            if (task == null)
+                throw new NotImplementedException("Proper awaitable implementation to follow.");
+
+#if NET40
+            // TODO: use the general reflection-based awaiter if net40 build is running against a newer BCL
+            return new Net40BclTaskAwaitAdapter(task);
+#else
+            return new TaskAwaitAdapter(task);
+#endif
+        }
+
+#if NET40
+        private sealed class Net40BclTaskAwaitAdapter : AwaitAdapter
+        {
+            private readonly Task _task;
+
+            public Net40BclTaskAwaitAdapter(Task task)
+            {
+                _task = task;
+            }
+
+            public override bool IsCompleted => _task.IsCompleted;
+
+            public override void OnCompleted(Action action)
+            {
+                if (action == null) return;
+
+                // Normally we would call TaskAwaiter.UnsafeOnCompleted (https://source.dot.net/#System.Private.CoreLib/src/System/Runtime/CompilerServices/TaskAwaiter.cs)
+                // We will have to polyfill on top of the TPL API.
+                // Compare TaskAwaiter.OnCompletedInternal from Microsoft.Threading.Tasks.dll in Microsoft.Bcl.Async.nupkg.
+
+                _task.ContinueWith(_ => action.Invoke(), TaskScheduler.FromCurrentSynchronizationContext());
+            }
+
+            public override void BlockUntilCompleted()
+            {
+                // Normally we would call TaskAwaiter.GetResult (https://source.dot.net/#System.Private.CoreLib/src/System/Runtime/CompilerServices/TaskAwaiter.cs)
+                // We will have to polyfill on top of the TPL API.
+                // Compare TaskAwaiter.ValidateEnd from Microsoft.Threading.Tasks.dll in Microsoft.Bcl.Async.nupkg.
+
+                try
+                {
+                    _task.Wait(); // Wait even if the task is completed so that an exception is thrown for cancellation or failure.
+                }
+                catch (AggregateException ex) when (ex.InnerExceptions.Count == 1) // Task.Wait wraps every exception
+                {
+                    ExceptionHelper.Rethrow(ex.InnerException);
+                }
+            }
+        }
+#else
+        private sealed class TaskAwaitAdapter : AwaitAdapter
+        {
+            private readonly TaskAwaiter _awaiter;
+
+            public TaskAwaitAdapter(Task task)
+            {
+                _awaiter = task.GetAwaiter();
+            }
+
+            public override bool IsCompleted => _awaiter.IsCompleted;
+
+            [SecuritySafeCritical]
+            public override void OnCompleted(Action action) => _awaiter.UnsafeOnCompleted(action);
+
+            // Assumption that GetResult blocks until complete is only valid for System.Threading.Tasks.Task.
+            public override void BlockUntilCompleted() => _awaiter.GetResult();
+        }
+#endif
+    }
+}
+#endif
--- a/src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
+++ b/src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
@@ -1,0 +1,136 @@
+// ***********************************************************************
+// Copyright (c) 2018 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+#if ASYNC
+using System;
+using System.Threading;
+
+#if NET40 || NET45
+using System.Windows.Forms;
+using System.Windows.Threading;
+#endif
+
+namespace NUnit.Framework.Internal
+{
+    internal abstract class MessagePumpStrategy
+    {
+        public abstract void WaitForCompletion(AwaitAdapter awaitable);
+
+        public static MessagePumpStrategy FromCurrentSynchronizationContext()
+        {
+#if NET40 || NET45
+            var context = SynchronizationContext.Current;
+
+            if (context is WindowsFormsSynchronizationContext)
+                return WindowsFormsMessagePumpStrategy.Instance;
+
+            if (context is DispatcherSynchronizationContext)
+                return WpfMessagePumpStrategy.Instance;
+#endif
+
+            return NoMessagePumpStrategy.Instance;
+        }
+
+        private sealed class NoMessagePumpStrategy : MessagePumpStrategy
+        {
+            public static readonly NoMessagePumpStrategy Instance = new NoMessagePumpStrategy();
+            private NoMessagePumpStrategy() { }
+
+            public override void WaitForCompletion(AwaitAdapter awaitable)
+            {
+                awaitable.BlockUntilCompleted();
+            }
+        }
+
+#if NET40 || NET45
+        private sealed class WindowsFormsMessagePumpStrategy : MessagePumpStrategy
+        {
+            public static readonly WindowsFormsMessagePumpStrategy Instance = new WindowsFormsMessagePumpStrategy();
+            private WindowsFormsMessagePumpStrategy() { }
+
+            public override void WaitForCompletion(AwaitAdapter awaitable)
+            {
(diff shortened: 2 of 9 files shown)


src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
```diff
@@ -0,0 +1,121 @@
...
+                context.Post(
+                    state => ((AwaitAdapter)state).OnCompleted(Application.Exit),
+                    state: awaitable);
+
+                Application.Run();
+            }
+        }
+
+#if !NET20
+        private sealed class WpfMessagePumpStrategy : MessagePumpStrategy
```
sharwell: :bulb: `DispatcherMessagePumpStrategy`

src/NUnitFramework/tests/SynchronizationContextTests.cs
```diff
@@ -0,0 +1,134 @@
...
+using NUnit.TestData;
+using NUnit.TestUtilities;
+
+#if ASYNC
+using System.Threading.Tasks;
+#endif
+
+namespace NUnit.Framework
+{
+    public static class SynchronizationContextTests
```
sharwell: :bulb: Several of these tests would benefit from assertions relating to the thread(s) on which code is executed.

src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
jnm2: You're right, but Dispatcher is such a generic name. Is it used for anything but WPF?

jnm2:
@sharwell SingleThreadedSyncronization is coming in the next PR.

jnm2:
WIP is for the net40 tests till I get a chance to look at this.

src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
CharliePoole: Worse than generic... it has a specific use already in the codebase. The WorkItemDispatcher is the thing that dispatches the work items that constitute the implementation of a test. We have three different dispatchers that do this in different ways.

jnm2: This is a private nested class, so if you're seeing the name you're also seeing that it's coupled to DispatcherSynchronizationContext. I'm less worried there and more just intent on... um, bikeshedding. 😎

jnm2:
Yikes, caught the problem. Application.Exit and Dispatcher.ExitAllFrames were being called on the awaiter's continuation thread rather than being posted back to the synchronization context. That's pretty bad.

jnm2:
Fixed the WaitForCompletion implementations and added thread assertions per @sharwell's review.

jnm2:
Oh great, Mono is doing this in CI:

```
The type initializer for 'System.Windows.Forms.WindowsFormsSynchronizationContext' threw an exception.
  ----> System.TypeInitializationException : The type initializer for 'System.Windows.Forms.XplatUI' threw an exception.
  ----> System.ArgumentNullException : Could not open display (X-Server required. Check your DISPLAY environment variable)
Parameter name: Display
```

And for `new DispatcherSynchronizationContext()`:
```
System.NotImplementedException : The method or operation is not implemented.
```

jnm2:
I see that CI will pass. Ready for review.

jnm2:
@rprouse Is this PR close enough to something you'd approve that I can start the next PR (continuing on the STA thread after an await) based on this one?

src/NUnitFramework/framework/Internal/AwaitAdapter.cs
```diff
@@ -0,0 +1,111 @@
...
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+
+#if ASYNC
```
rprouse: The `AwaitAdapter` is only used in `AsyncInvocationRegion` which is only used when `ASYNC` is defined, so why isn't the whole class wrapped in `#if ASYNC` and why the `#ifdef` for other platforms that don't support async? Future PR?

rprouse: By other platform `#ifdef`s, I mean in the following files which could also be wrapped in an `#if ASYNC`.

src/NUnitFramework/framework/Internal/AwaitAdapter.cs
```diff
@@ -0,0 +1,111 @@
...
+            {
+                if (action == null) return;
+
+                // Mimick TaskAwaiter.UnsafeOnCompleted
+                _task.ContinueWith(_ => action.Invoke(), TaskScheduler.FromCurrentSynchronizationContext());
+            }
+
+            public override void BlockUntilCompleted()
+            {
+                // Mimick TaskAwaiter.GetResult
```
rprouse: Can you point to documentation on this and the mimic UnsafeOnCompleted above to help bring us up to speed?

src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
```diff
@@ -0,0 +1,139 @@
...
+
+namespace NUnit.Framework.Internal
+{
+    internal abstract class MessagePumpStrategy
+    {
+        public abstract void WaitForCompletion(AwaitAdapter awaitable);
+
+        public static MessagePumpStrategy FromCurrentSynchronizationContext()
+        {
+#if NET20 || NET35 || NET40 || NET45
```
rprouse: Once again, why the support for platforms that don't support async?

src/NUnitFramework/testdata/On.cs
```diff
@@ -0,0 +1,52 @@
...
+// ***********************************************************************
+
+using System;
+using System.Threading;
+
+namespace NUnit.TestData
+{
+    public static class On
+    {
+        public static IDisposable Dispose(Action action)
```
rprouse: Some doc comments to explain what this is and what it does so people don't have to read through the code? Cool little helper, didn't you have this in another PR awhile ago?

src/NUnitFramework/tests/Extensions.cs
```diff
@@ -0,0 +1,38 @@
...
+// ***********************************************************************
+
+using System;
+using NUnit.Framework.Interfaces;
+
+namespace NUnit.Framework
+{
+    internal static class Extensions
+    {
+        public static void AssertPassed(this ITestResult result)
```
rprouse: Another useful utility, but we should add doc comments 😉

src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
jnm2: Originally this file wasn't related to C#'s async. Now that I've changed it to receive an object so that the message pump strategy can decide whether to use a callback or to block, and I called the object AwaitAdapter, it really makes sense to make this only show up along with async support. I'll use `ASYNC`.

src/NUnitFramework/framework/Internal/AwaitAdapter.cs
jnm2: 🗨 https://github.com/nunit/nunit/pull/2776#discussion_r177935936

src/NUnitFramework/testdata/On.cs
jnm2: 😊 Yes I did, https://github.com/nunit/nunit-console/blob/master/src/NUnitEngine/nunit.engine.tests/Helpers/On.cs

jnm2:
Awesome, thanks! I will add documentation.

src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
jnm2: Once I add `#if ASYNC` at the top, here I'll need to choose between doing `#if !(NETSTANDARD1_6 || NETSTANDARD2_0)` and `#if NET40 || NET45`.

src/NUnitFramework/framework/Internal/AwaitAdapter.cs
jnm2: A future PR might remove the need for most of our #if ASYNC, but that should be done atomically.

jnm2:
@rprouse Ready for review. Boarding a plane so I won't have internet for a few hours.

jnm2:
False alarm, introduced a build error with the fuller polyfill.

jnm2:
Ready now.

jnm2:
@rprouse Sorry, I forgot to re-ping you.

rprouse:
@jnm2 I'm booked up today. I'll try to take a look tomorrow.

src/NUnitFramework/framework/Internal/AwaitAdapter.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            var task = awaitable as Task;
+            if (task == null)
+                throw new NotImplementedException("Proper awaitable implementation to follow.");
+
+#if NET40
+            // TODO: use the general reflection-based awaiter if net40 build is running against a newer BCL
+            return new Net40BclTaskAwaitAdapter(task);
+#else
+            return new TaskAwaitAdapter(task);
```
rprouse: What about `ValueTask`?

src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        public static MessagePumpStrategy FromCurrentSynchronizationContext()
+        {
+#if NET40 || NET45
+            var context = SynchronizationContext.Current;
+
+            if (context is WindowsFormsSynchronizationContext)
+                return WindowsFormsMessagePumpStrategy.Instance;
+
+            if (context is DispatcherSynchronizationContext)
```
rprouse: Will we need to be adding additional message pump strategies for other platforms like Xamarin, possibly Unity?

src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
```diff
@@ -0,0 +1,136 @@
...
+                return WpfMessagePumpStrategy.Instance;
```
rprouse: Is this used for UWP also?

src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
rprouse: If so, do we need issues to track these or do we wait for user requests?

src/NUnitFramework/testdata/On.cs
```diff
@@ -0,0 +1,59 @@
...
+    /// <summary>
+    /// Enables the <see cref="On.Dispose"/> syntax.
+    /// </summary>
+    public static class On
+    {
+        /// <summary>
+        /// Wraps an action so that it is executed when the returned object is disposed.
+        /// This disposal is thread-safe and the action will be executed at most once.
+        /// </summary>
+        public static IDisposable Dispose(Action action)
```
rprouse: Didn't you try to get this into an earlier pull requests? 😄 

Nice little utility 👍

src/NUnitFramework/framework/Internal/AwaitAdapter.cs
jnm2: I'm hoping to add support for awaitables (versus just Task as we have supported so far) in the third PR.

src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
jnm2: @rprouse I don't know enough about Xamarin and Unity to know whether they run message pumps.
https://developer.xamarin.com/api/type/Android.OS.Looper/ suggests that they do, but unless we can test these in CI, I'm not inclined to do any work on them. I spent an hour trying to figure out why VS starts the Visual Studio Android emulator but never finishes deploying a Xamarin Android project to it, so suffice it to say that I'm not in great shape for that on my main machine.

jnm2: \<rant>
The state of Android emulation on Windows is unreasonable. HAXM is no good because it requires me to shut down my Hyper-V virtual machines, one of which is a server, not to mention Docker. ARM is ten times slower since it's not hardware accelerated. The Visual Studio Android Emulator is perfect because it's accelerated and works without disabling Hyper-V, but I've never gotten it to work on my new machine and they only have up to API level 23 images.
\</rant>

src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
jnm2: No, but TIL UWP has its own: https://docs.microsoft.com/uwp/api/Windows.UI.Core.CoreDispatcher

jnm2:
> Looks good, although the message pump strategies are a bit over my head

I have failed 😭 lol.

### Background

In a nutshell: -> **We are forced to run an async thing synchronously.**
(That is what we are adapting here. In general you *never ever* do this, but since NUnit is a framework and is the entry point to all user code, it can be excused. The entirety of NUnit's threading model expects us to block the worker thread until the test is finished, whether or not it is asynchronous.)

So let's say you do the obvious thing. You adapt async to sync by having NUnit's worker thread call .`Wait()` (or `.GetAwaiter.GetResult()`. The worker thread blocks until the awaitable is completed.
So what happens when some awaitables (like `Task`) schedule part of their execution by posting it to some kind of message queue to execute on the original thread? Well, the original thread is still blocked on `.Wait()` and won't wake up until the whole task is completed—and at the same time, the task can't complete unless the original thread _does_ wake up, pulls the continuation out of the queue, executes it, and goes back to waiting for the task to signal that it is totally finished executing. That can't happen via `Wait()`, so they wait for each other forever. There is a deadlock.

### Solution

-> **Instead of `Wait()` I call `Application.Run()`.**
It runs a message loop, so any time an async thing schedules part of its execution on the Windows Forms thread while we're waiting, it gets executed as intended. Voila, no deadlocks!

`Application.Run()` does not return to the caller until `Application.Exit()` is called. Therefore:
-> **instead of `task.Wait()`, we want to call `task.GetAwaiter().OnCompleted(Application.Exit)`**.
That means as soon as the task is complete it puts the action `Application.Exit()` in the queue to be executed by the message loop inside Application.Run(). As soon as the loop pulls that action out of the queue and executes it, that signals it to shut itself down instead of waiting for the next message. Thus, it returns back to the NUnit worker thread that called it.

That's all there is to the general idea! Of course, Application.Run/Exit is specific to Windows Forms so this must be repeated for each message pump's API.

### Edge cases

So why don't you see `.GetAwaiter().OnCompleted(Application.Exit)` anywhere? Well, if the thing you are awaiting is not a Task—for example, if it's a ConfiguredTaskAwaitable from `task.ConfigureAwait(false)`—there is no guarantee which thread `Application.Exit` will execute on. That's what [this `else`](https://github.com/nunit/nunit/pull/2776/files#diff-308ac1319e4d291401e5ccd48a2b3651R129) handles.

The only other catch is explained (clearly, I hope) by [this comment](https://github.com/nunit/nunit/pull/2776/files#diff-308ac1319e4d291401e5ccd48a2b3651R106).

### Bonus details: What is a `SynchronizationContext` even

It's per-thread global state. It's the API which says very generally, "Here’s how to get back to the context you started from."

If you are awaiting an HTTP download, you've stopped executing on any threads at all. You've just left enough information behind so that, when the data arrives and the underlying async Windows API starts executing on a background thread, that background thread has enough information to know 1) where to stash that data and 2) what action to invoke to execute the next piece of the higher-level async method.

Problem: you can't always execute the next piece of the async method right away. We're on a background thread. if we executed the rest of the async method after the `await`, we might try to do something with UI and Windows Forms will throw a helpful exception because we're on the wrong thread.

So: when the UI thread gets to an `await` in your async method, and records where it is, and quits executing the async method so that it can go back to pumping messages, _rather than_ leaving the background thread an action which _directly_ executes the next part of the async method after the `await`, it leaves behind an action which calls the equivalent of `capturedSynchronizationContext.Post(_ => executeNextPartOfAsyncMethod.Invoke(), null);`.

 1. The background I/O thread executes the action it was given.
 2. The action contains ~`capturedSynchronizationContext.Post(_ => innerAction.Invoke(), null)`, so when the background thread executes it, `capturedSynchronizationContext` puts `innerAction` in a thread message queue.
 3. `Application.Run()`, in a galaxy far, far away, wakes up from waiting on the same queue, pulls out the message, and executes `innerAction`.

jnm2:
@rprouse I really do want to make sure that if something happens to me, I've left behind enough information that the team feels like they can own this stuff. Did the above help answer any questions about why I wrote exactly what I wrote?

jnm2:
@rprouse 

> I think @sharwell's comment about checking the threads is a good idea.

I meant to address in https://github.com/nunit/nunit/pull/2776/commits/38470e67b42cc91958de89f2c5a6e0e7a43da6a5. Is there maybe room for more checking?

src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
jnm2: 😍 https://blogs.msdn.microsoft.com/visualstudio/2018/05/08/hyper-v-android-emulator-support/

PR #2780, opened 2018-03-27
Only set thread's current culture when required

### LINKED ISSUE #2777: DefaultThreadCurrentUICulture and CurrentUICulture
I just found out that there is a problem with the change of the DefaultThreadCurrentUICulture that happened from 3.10.
Normally when we set the DefaultThreadCurrentUICulture, it defines the current culture and current UI culture of a new application thread. 

Until version 3.9, it works fine, but from version 3.10.0, setting DefaultThreadCurrentUICulture does not update CurrentUICulture.

```
// up to version 3.9
string oldCurrentUiCulture = CultureInfo.CurrentUICulture.Name; // fr-FR
CultureInfo.DefaultThreadCurrentUICulture = new CultureInfo("en-GB");
string newCurrentUiCulture = CultureInfo.CurrentUICulture.Name; // en-GB

// until version 3.10
string oldCurrentUiCulture = CultureInfo.CurrentUICulture.Name; // fr-FR
CultureInfo.DefaultThreadCurrentUICulture = new CultureInfo("en-GB");
string newCurrentUiCulture = CultureInfo.CurrentUICulture.Name; // fr-FR
```

My target framework is a .Net Core 2.0.

jnm2:
Fixes #2777

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
@@ -475,8 +475,11 @@ namespace NUnit.Framework.Internal.Execution
         {
             thread = new Thread(() =>
             {
-                thread.CurrentCulture = Context.CurrentCulture;
-                thread.CurrentUICulture = Context.CurrentUICulture;
+                if (thread.CurrentCulture != Context.CurrentCulture)
+                    thread.CurrentCulture = Context.CurrentCulture;
+
+                if (thread.CurrentUICulture != Context.CurrentUICulture)
+                    thread.CurrentUICulture = Context.CurrentUICulture;
 #if THREAD_ABORT
                 lock (threadLock)
                     nativeThreadId = ThreadUtility.GetCurrentThreadNativeId();
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -383,7 +383,8 @@ namespace NUnit.Framework.Internal
             {
                 _currentCulture = value;
 #if !NETSTANDARD1_6
-                Thread.CurrentThread.CurrentCulture = _currentCulture;
+                if (Thread.CurrentThread.CurrentCulture != _currentCulture)
+                    Thread.CurrentThread.CurrentCulture = _currentCulture;
 #endif
             }
         }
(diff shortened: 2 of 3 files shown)


jnm2:
Need some sleep tonight. Some of this might have to wait until I'm on a faster machine. Both VS and ReSharper's test runners are not working and build.cake is taking way too much time. This is the first time I've felt the Surface Pro 4 was slow.

jnm2:
@nunit/framework-team Ready for reviews. CI has passed on everything except OSX which is backed up again.

CharliePoole:
I didn't review when I looked at it because I couldn't see the purpose.

jnm2:
@CharliePoole The purpose is to fix #2777 which fails because CurrentCulture stops inheriting the value from DefaultThreadCurrentCulture when we do (effectively) thread.CurrentCulture = thread.CurrentCulture.

CharliePoole:
@jnm2 So it sounds like you are saying that Thread.CurrentCulture is not just a simple get/set property but returns the DefaultThreadCurrentCulture if the thread culture has not been set. Right?

If so, that's a bit weird! It seems like a user setting DefaultThreadCurrentCulture could influence __any__ other tests that are subsequently run on the same thread, which is a very bad thing. However, it's probably no worse than setting Thread.CurrentCulture, which we also ask users not to do. After each test, we actually want to regain control of the CurrentCulture and CurrentUICulture __no matter what the user has done__. Will your change in any way prevent that as far as you know? 

The following things should be true:
1. Using the SetCulture and SetUICulture attributes should change the culture.
2. The change should be for the entire scope of the test, whether it's a single case or a suite.
3. User setting culture or UI culture should also last only for the scope of the test, unless changed again by the user.
4. [NEW] User setting DefaultThreadCurrentCulture should have the same scope as (3). I believe that setting the CurrentCulture and CurrentUICulture unconditionally is what currently guarantees this. I'm not 100% clear that it's still guaranteed with the conditional test. Do you think it is?

CharliePoole:
A simpler comment: what if we just document that the CurrentCulture and CurrentUICulture of the thread are __always__ set by NUnit. That was the intent in the first place. 

With that, we can add that setting DefaultThreadCurrentCulture only affects new threads you create, not the threads that NUnit is responsible for.

In fact, that makes sense to me. Why should the user be able to change the default culture of a thread already created by somebody else?

ravichandranjv:
If I am not interrupting, it may be helpful and useful to add a use case for a proposed fix. This would smoothen out worry wrinkles of having a fix that may not be used or may break (not just for this fix) other code. If there is no specific use case, maybe just documenting it may suffice?

jnm2:
@CharliePoole That is very clear reasoning and I like it.

jnm2:
@ravichandranjv You are certainly not interrupting. I can't think of a use case, so until one comes to light (and is important enough to change our current design) we should just document that NUnit sets the culture on all worker threads.

PR #2796, opened 2018-04-05
Removed ITypeInfo abstraction

### LINKED ISSUE #2790: Removing ITypeInfo abstraction
It appears that both NUnit's code and the ITypeInfo abstraction itself are tightly coupled to `System.Type`.
This does not appear to have the potential to be an area of volatility, so the abstraction appears to be pure overhead. That's definitely the feeling I get each time I've worked with it. It's not useful for cross-platform BCL differences since we use extension methods in namespaces for that.

How bad of a change would it be to remove TypeWrapper and ITypeInfo and run directly against the reflection API?

jnm2:
Closes #2790. The first five commits are preliminary cleanup that I didn't know I had to do at first. The last commit is the actual removal of the abstraction.

`IMethodInfo` was doing double duty, tracking the fixture type along with the MethodInfo. This is because the same MethodInfo may be used for various points in the fixture's inheritance hierarchy. A long time ago, we used MethodInfo.ReflectedType to track this. MethodInfo.ReflectedType is deprecated and was [never really intended to be used for important domain logic](https://web.archive.org/web/20160429093051/http://blogs.msdn.com/b/kingces/archive/2005/08/01/446247.aspx). I made a new struct type, [`FixtureMethod`](https://github.com/nunit/nunit/pull/2796/files#diff-af7fa7cf599785bbbcc3fbf55d841c4aR29), which tracks where we are in the fixture's inheritance hierarchy.

The upshot of this is that most interfaces' `IMethodInfo` parameters became `FixtureMethod`, `ITypeInfo` became `Type`, and `IParameterInfo` became `ParameterInfo`. I'll do a full API comparison, but needless to say, removing these interfaces is a totally breaking change. I've spent so much time on this that I'm becoming quite partial to the organization I did 😛, but I'll leave it to @rprouse to decide how tolerant we should be of shipping a change like this. After all, it was done once when ITypeInfo first showed up... 😁


src/NUnitFramework/tests/Extensions.cs
```diff
@@ -1,5 +1,5 @@
-﻿// ***********************************************************************
-// Copyright (c) 2015 Charlie Poole, Rob Prouse
+// ***********************************************************************
+// Copyright (c) 2018 Charlie Poole, Rob Prouse
```
jnm2: Git is confused. These files shouldn't bear any resemblance to each other.

src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
```diff
@@ -99,415 +97,127 @@ public static Delegate CreateDelegate(this MethodInfo method, Type type)
...
-        /// Returns an array of generic arguments for the give type
+        /// Returns an array of <see cref="Type"/> objects that represent the type arguments of a closed generic type or the type parameters of a generic type definition.
         /// </summary>
-        /// <param name="type"></param>
-        /// <returns></returns>
         public static Type[] GetGenericArguments(this Type type)
         {
-            var info = type.GetTypeInfo();
-            return info.GenericTypeArguments.Concat(info.GenericTypeParameters).ToArray();
+            return type.GetTypeInfo().GetGenericArguments();
```
rprouse: We have an extension for Type.GetTypeInfo() for .NET 2.0 and other frameworks that don't support it that just returns Type for those frameworks. Because of that, we added GetTypeInfo to nearly every instance of type. So, now that PCL is gone, do we even need these extensions?

src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
```diff
@@ -99,415 +97,127 @@ public static Delegate CreateDelegate(this MethodInfo method, Type type)
...
-        /// <returns></returns>
-        public static PropertyInfo GetProperty(this Type type, string name, BindingFlags flags)
+        public static PropertyInfo GetProperty(this Type type, string name, BindingFlags bindingAttr)
         {
-            bool declaredOnly = flags.HasFlag(BindingFlags.DeclaredOnly);
-            return type.GetRuntimeProperties()
-                .Where(prop => declaredOnly ? prop.DeclaringType.Equals(type) : true)
-                .ApplyBindingFlags(flags)
-                .Where(p => p.Name == name)
-                .FirstOrDefault();
```
rprouse: All of my beautiful Linq code gone! I worked so hard on this 😛

src/NUnitFramework/framework/Interfaces/FixtureMethod.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            return obj is FixtureMethod && Equals((FixtureMethod)obj);
+        }
+
+        /// <summary>
+        /// Returns the hash code for this instance.
+        /// </summary>
+        public override int GetHashCode()
+        {
+            var hashCode = 662238274;
```
rprouse: Can you enlighten me on this method of generating a hash code? What is the significance of the number? I see this in Roslyn, but initializing with -1923861349.

src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
jnm2: @rprouse We do not, but they are a convenience if you don't like typing `GetTypeInfo()`. I don't have a strong opinion either way on whether we keep these shortcuts.

src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
jnm2: 😢 On one hand, code is a liability; on the other hand, I know how you feel!

src/NUnitFramework/framework/Interfaces/FixtureMethod.cs
jnm2: This was me using Roslyn 15.6.5, Ctrl-dotting on the struct declaration and selecting "[Generate Equals and GetHashCode](https://docs.microsoft.com/visualstudio/ide/reference/generate-equals-gethashcode-methods)." I don't know why it picked the constants it did, though I'm curious why I can't find discussion on it. Here's the [evolution of the generating code](https://github.com/dotnet/roslyn/commits/master/src/Workspaces/Core/Portable/Shared/Extensions/ICodeDefinitionFactoryExtensions_CreateGetHashCodeMethod.cs) in Roslyn.

jnm2: @rprouse It looks like the initial hashcode is derived from the names of the members involved, so you'd expect it to be different most of the type: https://github.com/dotnet/roslyn/blob/b6f7194660ebf549038a0aecd9f3b4ad3dd4d0ac/src/Workspaces/Core/Portable/Shared/Extensions/ICodeDefinitionFactoryExtensions_CreateGetHashCodeMethod.cs#L63-L73

src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
rprouse: I worked for ages getting all of these reflection API's running the same as the built in ones, but I will be happy to drop the complexity 😄

src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
rprouse: I don't have a strong opinion either, it just seems like these extension methods are now very thin wrappers and that the standard now is to call `.GetTypeInfo()` on types, so might be more expected. Let's merge as is and get a feel for the shape of the code.

rprouse:
Looks good, but there is a merge conflict with your previous PR.

src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
jnm2: 👍 They did a 180 with `GetTypeInfo()` in .NET Standard 2.0, so I'm wondering if it won't end up being considered legacy.

src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs
CharliePoole: I have to say that this is a path I followed for many years myself... trying to hide complexity in order to make life simpler for the average developer. I eventually came to believe that sometimes the complexity needs to be visible, so that people will see it's there and take it into account. But deciding on a case by case basis is rather hard!

src/NUnitFramework/framework/Interfaces/FixtureMethod.cs
jnm2: Cyrus says he was just copying the logic used by the C# compiler to implement GetHashCode for anonymous types, so the discussion would have all happened in the development of C# 3. =D

jnm2:
@rprouse Created `tests/Extensions.cs` in both PRs. The rebase is ready to be reviewed now; the only change is that file.

jnm2:
One of the AppVeyor builds went over a net35 timeout:

```cs
1) Failed : NUnit.Framework.Internal.Execution.EventQueueTests+DequeueBlocking_StopTest.DequeueBlocking_Stop
Test exceeded Timeout value of 2000ms
```

Restarted it.

PR #2801, opened 2018-04-12
Issue #2427: PropertyConstraint throws away the more helpful message …

### LINKED ISSUE #2427: PropertyConstraint throws away the more helpful message in the base constraint result
## Problem

PropertyConstraint keeps the base constraint description but throws away the actual comparison message and just prints the actual value.

Starting with a familiar example, `Is.EquivalentTo`:

```cs
var actual = new SomeClass { SomeProperty = { 2, 3, 5, 7 } };
Assert.That(actual.SomeProperty, Is.EquivalentTo(new List<int> { 2, 3, 5, 8 }));
```

>   Expected: equivalent to < 2, 3, 5, 8 >
>   But was:  < 2, 3, 5, 7 >
>   Missing (1): < 8 >
>   Extra (1): < 7 >

But as soon as you compose it with `PropertyConstraint`:

```cs
var actual = new SomeClass { SomeProperty = { 2, 3, 5, 7 } };
Assert.That(actual, Has.Property("SomeProperty").EquivalentTo(new List<int> { 2, 3, 5, 8 }));
```

>   Expected: equivalent to < 2, 3, 5, 8 >
>   But was:  < 2, 3, 5, 7 >

In that example, just printing the actual value is still pretty okay.
But when the value is a complex object that doesn't `.ToString()` to something you care about, you get something like this:

>   Expected: property SomeProperty [long description from custom constraint]
>   But was:  \<Appearance>

```cs
Assert.That(new SomeClass(), Has.Property("SomeComplexThing")
    .SomeCustomConstraint(expectedComplexThing));
```

## Proposal

Here you can see the constraint, which understands the semantics of the actual value, getting thrown away. Only the actual value and success bool are retained: 

https://github.com/nunit/nunit/blob/bf8f2dc863036655dc37583dd5b7408b356b6026/src/NUnitFramework/framework/Constraints/PropertyConstraint.cs#L73-L74

Rather than the existing code below, I think we should have an internal-for-now  `ChainedConstraintResult` class which delegates all its writing to the base constraint result:

```cs
internal class ChainedConstraintResult : ConstraintResult
{
    private readonly ConstraintResult _baseResult;

    public ChainedConstraintResult(IConstraint constraint, ConstraintResult baseResult) : base(constraint, baseResult.ActualValue, baseResult.Status)
    {
        _baseResult = baseResult;
    }

    public override void WriteActualValueTo(MessageWriter writer)
    {
        _baseResult.WriteActualValueTo(writer);
    }
}
```

PropertyConstraint would then do this:

```cs
var baseResult = BaseConstraint.ApplyTo(propValue);
return new ChainedConstraintResult(this, baseResult);
```

nisha-kaushik:
Fixes #2427

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
@@ -51,20 +51,20 @@ namespace NUnit.Framework.Constraints
             _tallyResult = tallyResult;
         }
 
-         /// <summary>Write the custom actual value for a failing constraint. <see cref="CollectionEquivalentConstraint"/>.</summary>
+         /// <summary>Write the custom extra chainable message to the failure message for a failing constraint. <see cref="CollectionEquivalentConstraint"/>.</summary>
         /// <param name="writer">The <see cref="MessageWriter"/> to write the failure message to.</param>
-        public override void WriteActualValueTo(MessageWriter writer)
-        {
-            base.WriteActualValueTo(writer);
-           
+        public override void WriteExtraChainableMessage(MessageWriter writer)
+        {            
+            base.WriteExtraChainableMessage(writer);
+
             if (_tallyResult.MissingItems.Count > 0)
             {
                 int missingItemsCount = _tallyResult.MissingItems.Count;
 
                 string missingStr = $"Missing ({missingItemsCount}): ";
                 missingStr += MsgUtils.FormatCollection(_tallyResult.MissingItems, 0, MaxDifferingElemsToWrite);
-                writer.WriteLine();
-                writer.WriteMessage(missingStr);
+
+                writer.WriteMessageLine(missingStr);
             }
 
             if (_tallyResult.ExtraItems.Count > 0)
@@ -73,8 +73,8 @@ namespace NUnit.Framework.Constraints
 
                 string extraStr = $"Extra ({extraItemsCount}): ";
                 extraStr += MsgUtils.FormatCollection(_tallyResult.ExtraItems, 0, MaxDifferingElemsToWrite);
-                writer.WriteLine();
-                writer.WriteMessage(extraStr);                
+
+                writer.WriteMessageLine(extraStr);
             }
         }
     }
--- a/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
@@ -51,12 +51,10 @@ namespace NUnit.Framework.Constraints
             _tallyResult = tallyResult;
         }
 
-         /// <summary>Write the custom extra chainable message to the failure message for a failing constraint. <see cref="CollectionEquivalentConstraint"/>.</summary>
+         /// <summary>Write the custom additional chainable message to the failure message for a failing constraint. <see cref="CollectionEquivalentConstraint"/>.</summary>
         /// <param name="writer">The <see cref="MessageWriter"/> to write the failure message to.</param>
-        public override void WriteExtraChainableMessage(MessageWriter writer)
-        {            
-            base.WriteExtraChainableMessage(writer);
-
+        public override void WriteAdditionalLinesTo(MessageWriter writer)
+        { 
             if (_tallyResult.MissingItems.Count > 0)
             {
                 int missingItemsCount = _tallyResult.MissingItems.Count;
--- a/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
@@ -51,21 +51,20 @@ namespace NUnit.Framework.Constraints
             _tallyResult = tallyResult;
         }
 
-        /// <summary>Write the custom actual value for a failing constraint. <see cref="CollectionEquivalentConstraint"/>.</summary>
+         /// <summary>Write the custom actual value for a failing constraint. <see cref="CollectionEquivalentConstraint"/>.</summary>
         /// <param name="writer">The <see cref="MessageWriter"/> to write the failure message to.</param>
         public override void WriteActualValueTo(MessageWriter writer)
         {
-            //Write the expected/actual message first.
             base.WriteActualValueTo(writer);
-
+           
             if (_tallyResult.MissingItems.Count > 0)
             {
                 int missingItemsCount = _tallyResult.MissingItems.Count;
 
                 string missingStr = $"Missing ({missingItemsCount}): ";
                 missingStr += MsgUtils.FormatCollection(_tallyResult.MissingItems, 0, MaxDifferingElemsToWrite);
-
-                writer.WriteMessageLine(missingStr);
+                writer.WriteLine();
+                writer.WriteMessage(missingStr);
             }
 
             if (_tallyResult.ExtraItems.Count > 0)
@@ -74,8 +73,8 @@ namespace NUnit.Framework.Constraints
 
                 string extraStr = $"Extra ({extraItemsCount}): ";
                 extraStr += MsgUtils.FormatCollection(_tallyResult.ExtraItems, 0, MaxDifferingElemsToWrite);
-
-                writer.WriteMessageLine(extraStr);
+                writer.WriteLine();
+                writer.WriteMessage(extraStr);                
             }
         }
     }
--- a/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2017 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -51,13 +51,13 @@ namespace NUnit.Framework.Constraints
             _tallyResult = tallyResult;
         }
 
-        /// <summary>Write the custom failure message for this object's <see cref="CollectionEquivalentConstraint"/>.</summary>
+        /// <summary>Write the custom actual value for a failing constraint. <see cref="CollectionEquivalentConstraint"/>.</summary>
         /// <param name="writer">The <see cref="MessageWriter"/> to write the failure message to.</param>
-        public override void WriteMessageTo(MessageWriter writer)
+        public override void WriteActualValueTo(MessageWriter writer)
         {
             //Write the expected/actual message first.
-            base.WriteMessageTo(writer);
-            
+            base.WriteActualValueTo(writer);
+
             if (_tallyResult.MissingItems.Count > 0)
             {
                 int missingItemsCount = _tallyResult.MissingItems.Count;
--- a/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
@@ -51,7 +51,7 @@ namespace NUnit.Framework.Constraints
             _tallyResult = tallyResult;
         }
 
-         /// <summary>Write the custom additional chainable message to the failure message for a failing constraint. <see cref="CollectionEquivalentConstraint"/>.</summary>
+        /// <summary>Write any additional lines (following <c>Expected:</c> and <c>But was:</c>) for a failing constraint.</summary>
         /// <param name="writer">The <see cref="MessageWriter"/> to write the failure message to.</param>
         public override void WriteAdditionalLinesTo(MessageWriter writer)
         { 
--- a/src/NUnitFramework/framework/Constraints/PropertyConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/PropertyConstraintResult.cs
@@ -1,7 +1,7 @@
 namespace NUnit.Framework.Constraints
 {
     /// <summary>
-    /// Contain the result of matching a <see cref="PropertyConstraint"/> against an actual value.
+    /// Contains the result of matching a <see cref="PropertyConstraint"/> against an actual value.
     /// </summary>
     internal sealed class PropertyConstraintResult : ConstraintResult
     {
--- a/src/NUnitFramework/framework/Constraints/PropertyConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/PropertyConstraintResult.cs
@@ -1,0 +1,42 @@
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// Contain the result of matching a <see cref="PropertyConstraint"/> against an actual value.
+    /// </summary>
+    public sealed class PropertyConstraintResult : ConstraintResult
+    {
+        private readonly ConstraintResult _baseResult;
+
+        #region Constructors
+
+        /// <summary>
+        /// Constructs a <see cref="PropertyConstraintResult"/> for a particular <see cref="PropertyConstraint"/>.
+        /// </summary>
+        /// <param name="constraint">The Constraint to which this result applies.</param>
+        /// <param name="baseResult">The base result with actual value to which the Constraint was applied.</param>       
+        public PropertyConstraintResult(IConstraint constraint, ConstraintResult baseResult) : base(constraint, baseResult.ActualValue, baseResult.Status)
+        {
+            _baseResult = baseResult;
+        }
+
+        #endregion
+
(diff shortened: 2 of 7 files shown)


jnm2:
Edit to add a [closing link](https://help.github.com/articles/closing-issues-using-keywords/) back to the issue.

jnm2:
There are a number of tests that began failing because of the changes to CollectionEquivalentConstraintResult. We will have to figure out a strategy for this.

The problem is that we can't [do as I suggested](https://github.com/nunit/nunit/issues/2427#issuecomment-328219335) and delegate `PropertyConstraintResult.WriteActualValueTo` to the base result's `WriteActualValueTo`, because the base result message is customized in `WriteMessageTo`.

We need some middle way for CollectionEquivalentConstraintResult to add `Missing:` etc that is neither included in `WriteActualValueTo` nor `WriteMessageTo`. It looks like we need a new API.

@CharliePoole Can I pick your brain here, since you were in the issue discussion?

CharliePoole:
I don't understand what you're saying here. The message generated by a constraint failure is intended to include things like "Missing" - that's how all of them work. I don't think it's the API that's failing here, it's the inheritance hierarchy used for collection constraints, which assumes they all work the same way. I think we talked about this already but if it's not clear, I'm happy to look closer and give some specific advice.

Note that inheriting from CollectionConstraint is not true `<isa>` inheritance, it's just a convenient way to share implementation, which you can drop if it isn't working for you.

CharliePoole:
It would help to know what failing tests the change is intended to fix.

jnm2:
On one hand, we can't have the collection constraint result write the `Missing:` test in `WriteMessageTo` because PropertyConstraintResult cannot display `Missing:` without displaying the wrong `Expected:` and `Actual:` lines.

On the other hand, we can't allow the collection constraint result to add `Missing:` in `WriteActualValueTo` because it's abuse of the API (it's not part of the actual value), and it causes [these test failures](https://ci.appveyor.com/project/CharliePoole/nunit/build/3.11.0-ci-05436-pr-2801#L338):

<details>

```
Errors, Failures and Warnings
1) Failed : NUnit.Framework.Assertions.CollectionAssertTest.EquivalentFailOne
  String lengths are both 122. Strings differ at index 74.
  Expected: "...", "y", "x" >\r\n  Missing (1): < "z" >\r\n  Extra (1): < "x" >\r\n"
  But was:  "...", "y", "x" >  Missing (1): < "z" >\r\n  Extra (1): < "x" >\r\n\r\n"
  ---------------------------^
   at NUnit.Framework.Assertions.CollectionAssertTest.EquivalentFailOne() in C:\projects\nunit\src\NUnitFramework\tests\Assertions\CollectionAssertTest.cs:line 367
2) Failed : NUnit.Framework.Assertions.CollectionAssertTest.EquivalentFailTwo
  String lengths are both 122. Strings differ at index 74.
  Expected: "...", "y", "z" >\r\n  Missing (1): < "x" >\r\n  Extra (1): < "z" >\r\n"
  But was:  "...", "y", "z" >  Missing (1): < "x" >\r\n  Extra (1): < "z" >\r\n\r\n"
  ---------------------------^
   at NUnit.Framework.Assertions.CollectionAssertTest.EquivalentFailTwo() in C:\projects\nunit\src\NUnitFramework\tests\Assertions\CollectionAssertTest.cs:line 383
3) Failed : NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestOneExtraOneMissing
  String lengths are both 128. Strings differ at index 74.
  Expected: "..."one" >\r\n  Missing (1): < "two" >\r\n  Extra (1): < "three" >\r\n"
  But was:  "..."one" >  Missing (1): < "two" >\r\n  Extra (1): < "three" >\r\n\r\n"
  ---------------------^
   at NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestOneExtraOneMissing() in C:\projects\nunit\src\NUnitFramework\tests\Constraints\CollectionEquivalentConstraintResultTests.cs:line 118
4) Failed : NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestOneExtraValue
  String lengths are both 109. Strings differ at index 81.
  Expected: "...t was:  < "three", "one", "two" >\r\n  Extra (1): < "three" >\r\n"
  But was:  "...t was:  < "three", "one", "two" >  Extra (1): < "three" >\r\n\r\n"
  -----------------------------------------------^
   at NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestOneExtraValue() in C:\projects\nunit\src\NUnitFramework\tests\Constraints\CollectionEquivalentConstraintResultTests.cs:line 102
5) Failed : NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestOneExtraValueDueToUnexpectedRepeatedValue
  String lengths are both 105. Strings differ at index 79.
  Expected: "...  But was:  < "one", "two", "two" >\r\n  Extra (1): < "two" >\r\n"
  But was:  "...  But was:  < "one", "two", "two" >  Extra (1): < "two" >\r\n\r\n"
  -------------------------------------------------^
   at NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestOneExtraValueDueToUnexpectedRepeatedValue() in C:\projects\nunit\src\NUnitFramework\tests\Constraints\CollectionEquivalentConstraintResultTests.cs:line 133
6) Failed : NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestOneMissingValue
  String lengths are both 93. Strings differ at index 65.
  Expected: "...", "two" >\r\n  But was:  < "one" >\r\n  Missing (1): < "two" >\r\n"
  But was:  "...", "two" >\r\n  But was:  < "one" >  Missing (1): < "two" >\r\n\r\n"
  -------------------------------------------------^
   at NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestOneMissingValue() in C:\projects\nunit\src\NUnitFramework\tests\Constraints\CollectionEquivalentConstraintResultTests.cs:line 58
7) Failed : NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestOnlyDisplaysUpToTenDifferences
  String lengths are both 241. Strings differ at index 139.
  Expected: "..., "six", "seven", "eight"... >\r\n  Extra (11): < "one", "tw..."
  But was:  "..., "six", "seven", "eight"... >  Extra (11): < "one", "two"..."
  --------------------------------------------^
   at NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestOnlyDisplaysUpToTenDifferences() in C:\projects\nunit\src\NUnitFramework\tests\Constraints\CollectionEquivalentConstraintResultTests.cs:line 148
8) Failed : NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestTwoExtraValues
  String lengths are both 125. Strings differ at index 89.
  Expected: "... "two", "three", "four" >\r\n  Extra (2): < "three", "four" >\r\n"
  But was:  "... "two", "three", "four" >  Extra (2): < "three", "four" >\r\n\r\n"
  ---------------------------------------^
   at NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestTwoExtraValues() in C:\projects\nunit\src\NUnitFramework\tests\Constraints\CollectionEquivalentConstraintResultTests.cs:line 87
9) Failed : NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestTwoMissingValues
  String lengths are both 98. Strings differ at index 63.
  Expected: "...wo" >\r\n  But was:  <empty>\r\n  Missing (2): < "one", "two" >\r\n"
  But was:  "...wo" >\r\n  But was:  <empty>  Missing (2): < "one", "two" >\r\n\r\n"
  ------------------------------------------^
   at NUnit.Framework.Constraints.CollectionEquivalentConstraintResultTests.TestTwoMissingValues() in C:\projects\nunit\src\NUnitFramework\tests\Constraints\CollectionEquivalentConstraintResultTests.cs:line 73
10) Failed : NUnit.Framework.Constraints.CollectionEquivalentConstraintTests.FailureMessageWithHashSetAndArray
  String lengths are both 188. Strings differ at index 122.
  Expected: "...bra", "presto", "hocusfocus" >\r\n  Missing (1): < "hocuspoc..."
  But was:  "...bra", "presto", "hocusfocus" >  Missing (1): < "hocuspocus..."
  --------------------------------------------^
   at NUnit.Framework.Constraints.CollectionEquivalentConstraintTests.FailureMessageWithHashSetAndArray() in C:\projects\nunit\src\NUnitFramework\tests\Constraints\CollectionEquivalentConstraintTests.cs:line 277
11) Failed : NUnit.Framework.Constraints.CollectionEquivalentConstraintTests.TestConstraintResultMessageDisplaysMissingAndExtraElements
  String lengths are both 128. Strings differ at index 74.
  Expected: "..."one" >\r\n  Missing (1): < "two" >\r\n  Extra (1): < "three" >\r\n"
  But was:  "..."one" >  Missing (1): < "two" >\r\n  Extra (1): < "three" >\r\n\r\n"
  ---------------------^
   at NUnit.Framework.Constraints.CollectionEquivalentConstraintTests.TestConstraintResultMessageDisplaysMissingAndExtraElements() in C:\projects\nunit\src\NUnitFramework\tests\Constraints\CollectionEquivalentConstraintTests.cs:line 226
```

</details>

This makes me thing we need a different way to compose the results of chained constraints.

CharliePoole:
I'm in no way suggesting you change `WriteActualValue` It should write the actual value 'cause that's what it does. Property constraint has to produce a message that's understandable on it's own. As I understand it, that's what this issue is about, no?

PropertyConstraint gets to decide what it's message looks like. As far as I can see, it is currently written on the assumption that the base constraint being applied to the value is an EqualConstraint. I get that from the fact that it uses the Expected: <value> But was: <value> format. It shouldn't be doing that at all.

I got confused by all the references to collection constraint, but the issue shows that collection constraint is already producing a decent message. PropertyConstraint is simply ignoring it. Why not just use it? Why get into the structure of various collection constraints?

jnm2:
@CharliePoole 

> I got confused by all the references to collection constraint, but the issue shows that collection constraint is already producing a decent message. PropertyConstraint is simply ignoring it. Why not just use it? Why get into the structure of various collection constraints?

We can't do that either or we lose the `Expected: property SomeProperty` from the message. Having that in the message is the entire reason I use `Has.Property`.

My point is that PropertyConstraint and its chained constraint need to collaborate to produce the `Expected:`, `Missing:` lines. The current API doesn't allow for such collaboration.

src/NUnitFramework/framework/Constraints/PropertyConstraintResult.cs
```diff
@@ -0,0 +1,42 @@
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// Contain the result of matching a <see cref="PropertyConstraint"/> against an actual value.
+    /// </summary>
+    public sealed class PropertyConstraintResult : ConstraintResult
```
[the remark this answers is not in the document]
nisha-kaushik: Yes, We can keep internal. I will edit the comments too.

nisha-kaushik:
Can't we go with current approach, tests were failing because of New Line in the end. 
Or we need another method for adding missing/extra values to the message.

jnm2:
@nisha-kaushik I know why you had to go the direction you did, but we don't like the idea of putting 'Missing:' in the actual value because that's not what it is. It spreads out the formatting logic at the wrong level of abstraction.

I think we need a new virtual method which allows you to combine any chainable constraint's results.

nisha-kaushik:
Got it.I added a new virtual method which allows to add extra message to the failure message and combine with chainable constraints. Will this work for us?

Also ,can you please suggest a nice name for the method, 'WriteExtraChainableMessage' is something I used.

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
```diff
@@ -51,13 +51,12 @@ public class CollectionEquivalentConstraintResult : ConstraintResult
...
+         /// <summary>Write the custom extra chainable message to the failure message for a failing constraint. <see cref="CollectionEquivalentConstraint"/>.</summary>
         /// <param name="writer">The <see cref="MessageWriter"/> to write the failure message to.</param>
-        public override void WriteMessageTo(MessageWriter writer)
-        {
-            //Write the expected/actual message first.
-            base.WriteMessageTo(writer);
-            
+        public override void WriteExtraChainableMessage(MessageWriter writer)
+        {            
+            base.WriteExtraChainableMessage(writer);
```
jnm2: Since `base` is known to be `Constraint`, this line is guaranteed to be a no-op and should be removed.

src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
```diff
@@ -276,6 +277,7 @@ private void WriteExpectedLine(object expected, Tolerance tolerance)
             Write(Pfx_Expected);
             Write(MsgUtils.FormatValue(expected));
             if (_sameValDiffTypes) {
+
```
jnm2: Would you mind undoing this blank line? Instead of fixing this in a new commit, if you amend the same commit you introduced it in, that keeps the git blame history clean.

src/NUnitFramework/tests/Constraints/PropertyTests.cs
```diff
@@ -183,6 +184,27 @@ public void MultipleProperties()
...
+
+        [Test]
+        public void FailureMessageContainsChainedConstraintMessage()
+        {
+            var inputObject = new { Foo = new List<int> { 2, 3, 5, 7 } };
+
+            //Property Constraint Message with chained Equivalent Constraint.
+            var constraint = ((IResolveConstraint)Has.Property("Foo").EquivalentTo(new List<int> { 2, 3, 5, 8 })).Resolve();
+
+            //Assert.That(inputObject, constraint);
```
jnm2: Debug code to remove?

jnm2:
Some thoughts on the name:
`WriteMessageTo` and `WriteActualValueTo` are chainable in the same way that `WriteExtraChainableMessage` is, so let's remove that word.

What do you think of `WriteAdditionalLinesTo`? We'd have the actual line, the expected line, and any additional lines like the `Missing:` and `Extra:` lines. /cc @nunit/framework-team

nisha-kaushik:
Ok noted. Will make the changes.

src/NUnitFramework/framework/Constraints/PropertyConstraintResult.cs
```diff
@@ -0,0 +1,40 @@
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// Contain the result of matching a <see cref="PropertyConstraint"/> against an actual value.
```
jnm2: For a separate PR since it needs to be done across all our constraint results:
ℹ Should be `Contains`, as in, “PropertyConstraintResult contains the result…”
Further, does PropertyConstraintResult *contain* the result, or *is* it the result?

src/NUnitFramework/framework/Constraints/PropertyConstraintResult.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        private readonly ConstraintResult _baseResult;
+
+        #region Constructors
+
+        /// <summary>
+        /// Constructs a <see cref="PropertyConstraintResult"/> for a particular <see cref="PropertyConstraint"/>.
+        /// </summary>
+        /// <param name="constraint">The Constraint to which this result applies.</param>
+        /// <param name="baseResult">The base result with actual value to which the Constraint was applied.</param>
```
jnm2: I have had an XML docs sweep on my todo list...

src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
```diff
@@ -51,13 +51,10 @@ public class CollectionEquivalentConstraintResult : ConstraintResult
             _tallyResult = tallyResult;
         }
 
-        /// <summary>Write the custom failure message for this object's <see cref="CollectionEquivalentConstraint"/>.</summary>
+         /// <summary>Write the custom additional chainable message to the failure message for a failing constraint. <see cref="CollectionEquivalentConstraint"/>.</summary>
```
jnm2: ⚠ This line is indented too far, and we should reflect the name change since we no longer call it ‘chainable message.’
Also, the end of the line renders as, “…for a failing constraint. [CollectionEquivalentConstraint](#example).” I'd be fine removing either `constraint. ` or the `<see>`.

Maybe something like this?


```cs
/// Write any additional lines (following <c>Expected:</c> and <c>But was:</c>) for a failing constraint.
```

nisha-kaushik:
Done, Please let me know if something else is needed.

nisha-kaushik:
Thank you so much, looking forward for more contribution in future :)

PR #2727, opened 2018-02-28
Avoid treating warnings as errors inside the IDE

sharwell:
This change helps improve developer inner loop efficiency by revealing but not _demanding attention_ for warnings that occur during normal iterative development. Anything missed will still be caught by the CI and command line builds.

💡 Please do not rebase or squash this pull request on merge.

PR #2803, opened 2018-04-12
#2759 -  Test fails with "No arguments were provided" error when no values returned from IParameterDataSource

### LINKED ISSUE #2759: Test fails with "No arguments were provided" error when no values returned from IParameterDataSource
We have a test that takes two parameters, seeded with IParameterDataSource attributes.
One parameter is an enum and always has values, second parameter is a string and could have 0 to many values based on test run configuration.

Problem is that when this set is empty, nunit (both VS adapter and console runner) report failed/invalid test with "No arguments were provided" message.
For now I found workaround for it - return empty string as an extra value and return from test if such value passed to test method.

I think such tests should not fail test run (allways, or maybe add a boolean property to IParameterDataSource interface or new method attribute).

DaiMichael:
Fixes #2759

**Description** 
We have a test that takes two parameters, seeded with IParameterDataSource attributes.
One parameter is an enum and always has values, second parameter is a string and could have 0 to many values based on test run configuration.

Problem is that when this set is empty, nunit (both VS adapter and console runner) report failed/invalid test with "No arguments were provided" message.

**Fix**
- Check for DataAttribute paramters explicitly and call paramitized method
- Tests for different possible DataAttribute's

**Evidence**
- No breaks in current tests.
- New tests pass.


src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
```diff
@@ -132,11 +132,30 @@ public Test BuildFrom(FixtureMethod method, Test suite)
                     tests.Add(test); 
             }
 
-            return hasBuildersSpecified && method.Method.GetParameters().Length > 0 || tests.Count > 0
+            return hasBuildersSpecified && method.Method.GetParameters().Length > 0 || tests.Count > 0 ||
+                   HasMethodGotDataAttributeParameters(method) || 
+                   tests.Count > 0
```
CharliePoole: Two problems with checking here...

1) If there are data attributes, then there should already be a builder, one of the combining attributes, either specified or supplied by NUnit. Rather than adding a check, it would be better to figure out why there is no combining attribute for the data.

2. tests.Count > 0 is repeated twice.

CharliePoole:
Please give PR a title that says what it does. That way people don't have to look up the issue number. Issue number belongs in a comment, usually the initial one, preceded by Fixes so Github will figure it out.

DaiMichael:
Apologies, I just left default, updated and I hope okay.

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
jnm2: @CharliePoole On this line, for the tests that have zero cases, `hasBuildersSpecified` is false. `builders.Count` had been empty because we only look for builders from the method attributes, not parameter attributes.

> Rather than adding a check, it would be better to figure out why there is no combining attribute for the data.

I don't understand what this means. `hasBuildersSpecified` is false because there were no ITestBuilder attributes on the method. There is a `CombinatorialAttribute`, but that has never had any influence over what we do here unless it creates at least two tests.

CharliePoole: @jnm2 I'll wait till I'm actually in the code to reply... I've been doing too much today off the top of my head.

rprouse:
The builds are failing because the XML document comments have not been updated or added. Every public method, property, etc in the framework must have proper XML document comments so that Intellisense works. I expect your changes are compiling with warnings although I thought warnings should fail your build?

Errors on Travis are,

```
Internal/Builders/DefaultTestCaseBuilder.cs(146,26): error CS1572: XML comment has a param tag for 'method', but there is no parameter by that name [/home/travis/build/nunit/nunit/src/NUnitFramework/framework/nunit.framework.csproj]
Internal/Builders/DefaultTestCaseBuilder.cs(148,72): error CS1573: Parameter 'fixtureMethod' has no matching param tag in the XML comment for 'DefaultTestCaseBuilder.HasMethodGotDataAttributeParameters(FixtureMethod)' (but other parameters do) [/home/travis/build/nunit/nunit/src/NUnitFramework/framework/nunit.framework.csproj]
Internal/Builders/DefaultTestCaseBuilder.cs(146,26): error CS1572: XML comment has a param tag for 'method', but there is no parameter by that name [/home/travis/build/nunit/nunit/src/NUnitFramework/framework/nunit.framework.csproj]
Internal/Builders/DefaultTestCaseBuilder.cs(148,72): error CS1573: Parameter 'fixtureMethod' has no matching param tag in the XML comment for 'DefaultTestCaseBuilder.HasMethodGotDataAttributeParameters(FixtureMethod)' (but other parameters do) [/home/travis/build/nunit/nunit/src/NUnitFramework/framework/nunit.framework.csproj]
Internal/Builders/DefaultTestCaseBuilder.cs(146,26): error CS1572: XML comment has a param tag for 'method', but there is no parameter by that name [/home/travis/build/nunit/nunit/src/NUnitFramework/framework/nunit.framework.csproj]
Internal/Builders/DefaultTestCaseBuilder.cs(148,72): error CS1573: Parameter 'fixtureMethod' has no matching param tag in the XML comment for 'DefaultTestCaseBuilder.HasMethodGotDataAttributeParameters(FixtureMethod)' (but other parameters do) [/home/travis/build/nunit/nunit/src/NUnitFramework/framework/nunit.framework.csproj]
Internal/Builders/DefaultTestCaseBuilder.cs(146,26): error CS1572: XML comment has a param tag for 'method', but there is no parameter by that name [/home/travis/build/nunit/nunit/src/NUnitFramework/framework/nunit.framework.csproj]
Internal/Builders/DefaultTestCaseBuilder.cs(148,72): error CS1573: Parameter 'fixtureMethod' has no matching param tag in the XML comment for 'DefaultTestCaseBuilder.HasMethodGotDataAttributeParameters(FixtureMethod)' (but other parameters do) [/home/travis/build/nunit/nunit/src/NUnitFramework/framework/nunit.framework.csproj]
Internal/Builders/DefaultTestCaseBuilder.cs(146,26): error CS1572: XML comment has a param tag for 'method', but there is no parameter by that name [/home/travis/build/nunit/nunit/src/NUnitFramework/framework/nunit.framework.csproj]
Internal/Builders/DefaultTestCaseBuilder.cs(148,72): error CS1573: Parameter 'fixtureMethod' has no matching param tag in the XML comment for 'DefaultTestCaseBuilder.HasMethodGotDataAttributeParameters(FixtureMethod)' (but other parameters do) [/home/travis/build/nunit/nunit/src/NUnitFramework/framework/nunit.framework.csproj]
Internal/Builders/DefaultTestCaseBuilder.cs(146,26): error CS1572: XML comment has a param tag for 'method', but there is no parameter by that name [/home/travis/build/nunit/nunit/src/NUnitFramework/framework/nunit.framework.csproj]
Internal/Builders/DefaultTestCaseBuilder.cs(148,72): error CS1573: Parameter 'fixtureMethod' has no matching param tag in the XML comment for 'DefaultTestCaseBuilder.HasMethodGotDataAttributeParameters(FixtureMethod)' (but other parameters do) [/home/travis/build/nunit/nunit/src/NUnitFramework/framework/nunit.framework.csproj]
```

jnm2:
@rprouse

> although I thought warnings should fail your build?

It's because of https://github.com/nunit/nunit/pull/2727.

DaiMichael:
First off, a bit embarrased about the duplicate test in the `if` statement, came from the rebase / merge and I didn't spot it.
Docs... meh, I'm not used to doing the documents on the API in work, I didn't even think about changing it after I changed the signature.. will know now.

Here is the latest push I did, I haven't done a pull request as I see there is disagreement about the change.

https://github.com/nunit/nunit/commit/d86a98318ada1e3536e323bc83628621e12984a5

With regards to the implementation, when I first commented to Rob, I had two ways to fix
1) The way I have, which kind of feels dirty in adding an additional "random" test.
2) Duplicate the `CombinatorialStrategy` and change the behaviour on the following line :

https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/Builders/CombinatorialStrategy.cs#L52

The problem with this solution is I would need to return a test which would be defined as "Skipped", this isn'the behaviour Rob wanted which left me with the easier fix in solution 1). FYI - Currently in the `CombinatorialStrategy`, the current behaviour will return false on L51 as there is no data returned (ala test) and returns no test cases, this then uses `DefaultTestCaseBuilder.BuildSingleTestMethod` which gives the failure message in the issue.

Rob had said he wanted the test to just not run, not ignore or skipped.

I'm new here :) and happy to leave the discussion of merits to more informed people on the code base. but I hope that explains why I took the slightly less clean path.

Let me know if you would like me to do a fresh pull request with the updated commit, make changes etc.

rprouse:
> Rob had said he wanted the test to just not run, not ignore or skipped.

I wanted the test marked as a warning like was done in the linked PR

> Here is the latest push I did, I haven't done a pull request as I see there is disagreement about the change.

Pushing changes automatically updates this PR, so there is no need to create another.

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
jnm2: It's just that I'm not following what you're saying. I debugged this branch and could be missing the obvious? Were you thinking we need a deeper refactor?

rprouse:
@jnm2 I might be mistaken, I thought we switched it to a warning so that users know there are no tests, but I might be wrong?

CharliePoole:
@jnm2 I'm not looking at the code right now but I believe the intent is to warn it there is no data for the method after considering all sources.

jnm2:
Oh! My mistake! I just discovered that none of the three test runner UIs I use display warnings, so... that's embarrassing. :D

CharliePoole:
Even if it's an error, it shouldn't be an error on the individual data source but only if none of them produce data.

jnm2:
We should be seeing a warning for `TestWithZeroTestSourceCasesShouldPassWithoutRequiringArguments`, and [I don't](https://ci.appveyor.com/project/CharliePoole/nunit/build/3.11.0-dev-05478). There also aren't any tests on the behavior that I can find, beyond making sure it doesn't fail.

I don't remember knowing about warnings, but I personally would also ask for a way to opt out of them. It's too common with integration tests.

CharliePoole:
Also note that a warning in discovery would be a first for us. It might call for a new Result State.

jnm2:
Anyway, can we safely say that enabling warnings would be a separate discussion and separate effort to @DaiMichael's PR?

CharliePoole:
I actually wasn't sure if @rprouse meant TestStatus.Warning or just the normal "warnings" we give instead of silently ignoring malformed tests... which are really errors of the Failure:Invalid kind.

jnm2:
I don't think @DaiMichael has a clear idea of what to do, nor to be honest am I totally able to follow. Since we don't warn for zero-case parameterized methods, can we merge this code? If not, can you help me understand what the approach is that @DaiMichael should be taking?

CharliePoole:
I don't have a clear idea of what you're trying to do and have not reviewed the code. Because we are trying to limit my role here, I only typically comment where it sounds like something is being missed.

In this case, your comment seemed to suggest that each data source would give it's own error or warning, if it found no data. IMO, if that's the intent, it's a bad idea. And factually, it's not the way anything has ever worked before. The error or warning is supposed to be at the test level not the data source level. I can give a method multiple data sources, expecting at least one of them to find data.

Regarding warnings... in the past we always said that not reliable errors are there as a warning to the user. If you want to make this one an __actual__ Warning, that seems reasonable but it's kind of a big deal and probably deserves to be separate.

jnm2:
@CharliePoole I certainly don't mind you commenting, but I'm a little confused; I'm referring to your [line comment](https://github.com/nunit/nunit/pull/2803#discussion_r181237276) when you requested changes in this PR. You're requesting code changes but I don't understand the request, as I [mentioned](https://github.com/nunit/nunit/pull/2803#discussion_r181490955).

The warnings discussion is separate. Since we don't do warnings today, can we keep this PR focused on making IParameterDataSource behave consistently with TestCaseSource and discuss adding warnings to both in a separate PR?

CharliePoole:
Sorry, that was two weeks ago and I thought we were well past it and discussing something else now. I'll look back to see if the change was done and post another comment.

However, on your second point: we should __never__ add warnings or errors when no data is found by either TestCaseSource OR ValueSource. The warnings (or errors) should happen at the level of the test. I realize I'm repeating myself, which probably doesn't make the comment any more understandable, but I don't know how else to say it.... The code for a data attribute must not detect whether it found any data and give an error or warning. Not having data is __normal__ for the attribute. It's only __abnormal__ for the test method as a whole!

Examples:
```C#
// Not an error because one set of args is provided
[TestCase(1, 2, 3)]
[TestCaseSource("X")] // X has no data!
public void SomeMethod(int x, int y, int z) { ... }

// This should be an error, since there are no test cases at all!
[TestCaseSource("X")] // Has no data!
[TestCaseSource("Y")] // Has no data
public void AnotherMethd(int x) { ... }
```

jnm2:
@CharliePoole 

> The warnings (or errors) should happen at the level of the test.

You expressed that very clearly earlier. I'm sorry I didn't respond to that. (I'm still wondering if we can skip talking about warnings in this PR altogether.)

jnm2:
@CharliePoole Thanks, specifically your point 1 which I asked about a couple weeks ago.

> The warnings (or errors) should happen at the level of the test.

You expressed that very clearly earlier. I'm sorry I didn't respond to that, and used imprecise wording while asking if we can safely skip talking about warnings in this PR altogether for the sake of unblocking @DaiMichael.

CharliePoole:
However, the comment applies even if we are talking about errors only.

jnm2:
We're talking about removing the current error, caused by `DefaultTestCaseBuilder` noticing the method has no cases, right?

CharliePoole:
That's what it started as. It felt  like the discussion had expanded and I was responding. By all means, let's get back to the actual issue.

jnm2:
Yes!
Since the PR is two weeks old, I'm hoping to get this first-time contributor's awesome work into our codebase. Just based on the current code I see, I would approve and merge. Here's what I'm waiting for:

 1. A team member (you) still has the PR marked as requesting changes. Would you be able to re-review since your point 2 is addressed, and approve the PR if your point 1 is sufficiently addressed at this time? If @DaiMichael's implementation would still benefit from changes, would you walk him/us through that outstanding concern, please?

 2. Rob asked for a warning to be reported in place of the error, thinking that's what we were already doing with TestCaseSource. I've just done a test and confirmed that no warnings show up in NUnit Console for any test method which uses an empty TestCaseSource. @rprouse, if I'm not missing anything, is this something that we can discuss after merging the PR?

Thanks guys! 😃

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
```diff
@@ -132,11 +132,30 @@ public Test BuildFrom(FixtureMethod method, Test suite)
                     tests.Add(test); 
             }
 
-            return hasBuildersSpecified && method.Method.GetParameters().Length > 0 || tests.Count > 0
+            return hasBuildersSpecified && method.Method.GetParameters().Length > 0 || 
+                   tests.Count > 0 ||
+                   HasMethodGotDataAttributeParameters(method)
```
jnm2: @DaiMichael I have a couple of ideas. Would you mind either:

 - Getting rid of the colloquialism ‘got,’ maybe using `MethodHasDataAttributeParameters`
 - Or, adding `using System.Linq;` and using `method.Method.GetParameters().Any(p => p.HasAttribute<DataAttribute>())` for this line.

Either one is good by me. I really appreciate it!

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
```diff
@@ -132,11 +132,30 @@ public Test BuildFrom(FixtureMethod method, Test suite)
...
+        /// Checks to see if the method has arguments with DataAttribute, 
+        /// if the argument has no data, the test should not run rather than fail.
+        /// </summary>
+        /// <param name="fixtureMethod">The method for which a test is to be built.</param>
+        /// <returns>True if the method signature contains DataAttribute arguments</returns>
+        private bool MethodHasDataAttributeParameters(FixtureMethod fixtureMethod)
+        {
+            foreach (var parameter in fixtureMethod.Method.GetParameters())
+            {
+                var dataAttributes = parameter.GetAttributes<DataAttribute>(true);
```
CharliePoole: This should be for `IParameterDataSource` and should use inherited=false because that's what NUnit does in all other places where parameters are used.

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
```diff
@@ -132,11 +132,30 @@ public Test BuildFrom(FixtureMethod method, Test suite)
                     tests.Add(test); 
             }
 
-            return hasBuildersSpecified && method.Method.GetParameters().Length > 0 || tests.Count > 0
+            return hasBuildersSpecified && method.Method.GetParameters().Length > 0 || 
+                   tests.Count > 0 ||
+                   MethodHasDataAttributeParameters(method)
```
CharliePoole: DataAttribute is a class that was somewhat important in earlier versions of NUnit. Currently, it's a holdover and is not used in the code at all. We left it in the code to avoid needlessly breaking anything. 

This change, if it were to be implemented would lock in `DataAttribute` as required for all data classes that apply to parameters. User classes would be required to inherit from it and any existing user attributes that __do not__ inherit from it would break.

I don't particularly like this approach of scanning for attributes (actually re-scanning since we do it elsewhere) but if it's the only solution to the problem, you should look for the `IParameterDataSource` interface rather than `DataAttribute`.

CharliePoole: I think what we need to ask here is "Why is `hasBuilderSpecified` insufficient?" Of course, the answer is that we give the method a hidden `CombinatorialAttribute` if it doesn't already have one of the combining attributes on it. We do so __whether or not__ any parameter data sources are provided. If we are going to check for parameter data sources, perhaps we should do it up front and only add the unspecified `CombinatorialAttribute` if it's __really__ needed. Then `hasBuilderSpecified` could be set correctly.

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
```diff
@@ -132,11 +132,30 @@ public Test BuildFrom(FixtureMethod method, Test suite)
...
+        /// </summary>
+        /// <param name="fixtureMethod">The method for which a test is to be built.</param>
+        /// <returns>True if the method signature contains DataAttribute arguments</returns>
+        private bool MethodHasDataAttributeParameters(FixtureMethod fixtureMethod)
+        {
+            foreach (var parameter in fixtureMethod.Method.GetParameters())
+            {
+                var dataAttributes = parameter.GetAttributes<DataAttribute>(true);
+                if (dataAttributes.Length > 0)
+                    return true;
```
CharliePoole: You are returning true if __any__ parameters have a parameter data source specified. Is that what we want in the case where the user has mistakenly added a source to one parameter and not to another? I haven't worked out the logic of this, but we want to be sure that an error message is still given in that case.

jnm2:
@DaiMichael 

> > Note that targeting a particular level of C# does not mean that all features are available. Features are limited based on the framework targeted as well. For example, the engine currently targets .NET 2.0, which means that System.Linq extensions cannot be used in the code
> 
> So as it wasn't a big deal, I wrote it as the function, I'm happy to leave it. Perhaps the Linq coding standard is now defunct?

When it says, “For example, the engine,” by ‘engine’ it’s referring to a difference codebase, https://github.com/nunit/nunit-console. The framework depends on a .NET 2.0 polyfill library called NUnit.System.Linq. But you arrived at a good outcome.

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
jnm2: @DaiMichael Also, these lines could be refactored to use `HasAttribute` rather than `GetAttributes….Length != 0`.

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
jnm2: I like this approach. If I understand correctly, we can implement this by moving `var hasBuildersSpecified =` to after the `if (needCombinatorial)` statement.

@CharliePoole Beyond moving that statement: we shouldn't be checking for parameter attributes as builders, should we? Discovering parameter attributes as builders seems like a bigger change which would be outside the scope of this PR.

jnm2:
@CharliePoole Thanks for the review. I've found Github's UI to have some irritating usability problems too.

Dunno if this helps, but what helps me is to remember for line comments to avoid the green button.

CharliePoole:
@jnm2 Easy to say, but the big green button sucks you in! 😀 

And once you have started a review, you have to finish it else all your comments remain pending! My preferred approach is for reviews to be non-automated - but that ship has sailed. 😞

DaiMichael:
@CharliePoole @jnm2 

Okay - I have made a few changes... hope they are inline with what was discussed.

I now check for `IParameterDataSource` rather than `DataAttribute` and unit tests are updated accordingly.
I have refactored the method to check `CombinatorialAttribute` needs to be added taking into account if there is parametized data.

Method "feels" better to me, hope you agree. 

All tests pass.

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
```diff
@@ -106,24 +106,8 @@ public Test BuildFrom(FixtureMethod method, Test suite)
...
-
-            // We could check to see if here are any data attributes specified
-            // on the parameters but that's what CombinatorialAttribute does
-            // and it simply won't return any cases if it finds nothing.
-            // TODO: We need to add some other ITestBuilder than a combinatorial attribute
-            // because we want the attribute to generate an error if it's present on
-            // a generic method.
-            if (needCombinatorial)
+            // See if we need a CombinatorialAttribute added for parametarized data
+            if (MethodHasDataAttributeParameters(method) && !HasCombinatorialStrategy(builders))
```
CharliePoole: I would have put the loop for HasCombinatorialStrategy inline here, but that's a matter of style. Alternatively, a Linq expression would be OK here in my view, since it would be a very simple one.

[You asked about the standard earlier... my own general rule is that the code has to be understandable by __the team that you have__! IOW, there's no general principle you can apply without knowing who the developers are.]

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
```diff
@@ -132,11 +116,42 @@ public Test BuildFrom(FixtureMethod method, Test suite)
...
                 : BuildSingleTestMethod(method, suite);
         }
 
+        /// <summary>
+        /// Checks to see if the method has arguments with DataAttribute, 
+        /// if the argument has no data, the test should not run rather than fail.
+        /// </summary>
+        /// <param name="fixtureMethod">The method for which a test is to be built.</param>
+        /// <returns>True if the method signature contains DataAttribute arguments</returns>
+        private bool MethodHasDataAttributeParameters(FixtureMethod fixtureMethod)
```
CharliePoole: Can we change the method name and comments to reflect that we are looking for `IParameterDataSource` and not `DataAttribute`?

src/NUnitFramework/tests/Attributes/DataAttributeTests.cs
```diff
@@ -0,0 +1,81 @@
+// ***********************************************************************
+// Copyright (c) 2007 Charlie Poole, Rob Prouse
```
CharliePoole: 2007?

src/NUnitFramework/tests/Attributes/DataAttributeTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+using System;
+using System.Collections;
+using System.Reflection;
+using NUnit.Framework.Interfaces;
+
+namespace NUnit.Framework.Attributes
+{
+    [TestFixture]
+    public class DataAttributeTests
```
CharliePoole: Can we give this a more meaningful name? Are they all tests of how missing data is handled?

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
```diff
@@ -132,11 +116,42 @@ public Test BuildFrom(FixtureMethod method, Test suite)
                     tests.Add(test); 
             }
 
-            return hasBuildersSpecified && method.Method.GetParameters().Length > 0 || tests.Count > 0
-                ? BuildParameterizedMethodSuite(method, tests)
+            return builders.Count > 0 && method.Method.GetParameters().Length > 0 || tests.Count > 0
```
CharliePoole: That works out very nicely now!

NUnit.sln.DotSettings
```diff
@@ -1,175 +0,0 @@
-﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
jnm2: Could you put this back, please? 😉

src/NUnitFramework/framework/Internal/Builders/DefaultTestCaseBuilder.cs
```diff
@@ -132,11 +116,42 @@ public Test BuildFrom(FixtureMethod method, Test suite)
                     tests.Add(test); 
             }
 
-            return hasBuildersSpecified && method.Method.GetParameters().Length > 0 || tests.Count > 0
-                ? BuildParameterizedMethodSuite(method, tests)
+            return builders.Count > 0 && method.Method.GetParameters().Length > 0 || tests.Count > 0 
+                ? BuildParameterizedMethodSuite(method, tests)
```
jnm2: I wouldn't have asked if I wasn't requesting other changes, but while you're at it, could you also remove the spaces being added to the ends of these two lines? Visual Studio hides this from you (without an [extension](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.TrailingWhitespaceVisualizer)), but GitHub and the git CLI don't look as clean.

DaiMichael:
@CharliePoole @jnm2 

In for a penny in for a pound... I inlined one function so thought I'd be consistent and do the other. 
Test class renamed to `ParameterDataSourceTests`
Reverted .DotSettings

FYI - I suspect the ,DotSettings went missing as I was having issues with VS and I reset R# and VS settings earlier, I've been compiling Roslyn and it isn't playing nicely on my PC.

I should add the functions (now inline), I often do it that way as the function name can document what you are doing without the implementation, handy when staff changes regularly on projects. Obviously it is a style thing etc...

Hope this is good now!

DaiMichael:
@CharliePoole @jnm2 

Fixed the comments.

jnm2:
Thanks for getting this done, @DaiMichael!

PR #2804, opened 2018-04-13
Made fields readonly where possible

jnm2:
`readonly` improves code quality and clarity of intent.


rprouse:
I added a short sentence about `readonly` to the coding standard.

jnm2:
Someday if we ever end up using C# 7.2+ I'd like to make most of our structs `readonly struct` as well.

PR #2805, opened 2018-04-13
Removed boxing and eager range creation

jnm2:
Fixes: https://github.com/nunit/nunit/blob/b648b8a32f8641452a13103f06a6924b8a39c385/src/NUnitFramework/framework/Attributes/RangeAttribute.cs#L44-L45

Rather than building up an array of boxed values and then converting the values (reboxing them), values are now only generated on demand *and* now generated natively matching the parameter type rather than having to convert each value in the sequence.

I structured it this way so that other things can be moved here in the future, such as [RandomAttribute's generic specializations](https://github.com/nunit/nunit/blob/d4686bbfba471337c646ab2592a60d2713e59109/src/NUnitFramework/framework/Attributes/RandomAttribute.cs#L168-L193).

### DIFF
--- a/src/NUnitFramework/framework/Attributes/RangeAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RangeAttribute.cs
@@ -175,9 +175,26 @@ namespace NUnit.Framework
         {
             var from = ParamAttributeTypeConversions.Convert(_from, parameter.ParameterType);
             var to = ParamAttributeTypeConversions.Convert(_to, parameter.ParameterType);
-            var step = ParamAttributeTypeConversions.Convert(_step, parameter.ParameterType);
 
-            return ValueGenerator.Create(parameter.ParameterType).GenerateRange(from, to, step);
+            var valueGenerator = ValueGenerator.Create(parameter.ParameterType);
+
+            ValueGenerator.Step step;
+            if (!valueGenerator.TryCreateStep(_step, out step))
+            {
+                // ValueGenerator.CreateStep is responsible to enable incrementing bytes by (int)(-1),
+                // or perhaps in the future a DateTime by a TimeSpan, but the responsibility to convert
+                // attribute values from Double to Decimal is in ParamAttributeTypeConversions.
+                // (ParamAttributeTypeConversions tries to simulate what would happen in C# if the
+                // literal syntax used for the attribute value had been used as a parameter argument
+                // in a direct call to the test method.)
+                object stepValueToRequire;
+                if (!ParamAttributeTypeConversions.TryConvert(_step, parameter.ParameterType, out stepValueToRequire))
+                    stepValueToRequire = _step;
+
+                step = valueGenerator.CreateStep(stepValueToRequire);
+            }
+
+            return valueGenerator.GenerateRange(from, to, step);
         }
     }
 }
--- a/src/NUnitFramework/framework/Attributes/RangeAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RangeAttribute.cs
@@ -24,6 +24,7 @@
 using System;
 using System.Collections;
 using System.Reflection;
+using NUnit.Framework.Constraints;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
 
@@ -195,6 +196,12 @@ namespace NUnit.Framework
             }
 
             return valueGenerator.GenerateRange(from, to, step);
+        }
+
+        /// <summary>Returns a string that represents the current object.</summary>
+        public override string ToString()
+        {
+            return $"{MsgUtils.FormatValue(_from)} .. {MsgUtils.FormatValue(_step)} .. {MsgUtils.FormatValue(_to)}";
         }
     }
 }
--- a/src/NUnitFramework/framework/Attributes/RangeAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RangeAttribute.cs
@@ -182,15 +182,21 @@ namespace NUnit.Framework
             ValueGenerator.Step step;
             if (!valueGenerator.TryCreateStep(_step, out step))
             {
-                // ValueGenerator.CreateStep is responsible to enable incrementing bytes by (int)(-1),
-                // or perhaps in the future a DateTime by a TimeSpan, but the responsibility to convert
-                // attribute values from Double to Decimal is in ParamAttributeTypeConversions.
-                // (ParamAttributeTypeConversions tries to simulate what would happen in C# if the
-                // literal syntax used for the attribute value had been used as a parameter argument
-                // in a direct call to the test method.)
+                // ValueGenerator.CreateStep has the responsibility to enable Byte values to be incremented
+                // by the Int32 value -1. Or perhaps in the future, DateTime values to be incremented by a TimeSpan.
+                // It handles scenarios where the incrementing type is fundamentally different in its most natural form.
+
+                // However, ParamAttributeTypeConversions has the responsibility to convert attribute arguments
+                // that are only of a different type due to IL limitations or NUnit smoothing over overload differences.
+                // See the XML docs for the ParamAttributeTypeConversions class.
+
                 object stepValueToRequire;
                 if (!ParamAttributeTypeConversions.TryConvert(_step, parameter.ParameterType, out stepValueToRequire))
+                {
+                    // This will cause CreateStep to throw the same exception as it would throw if TryConvert
+                    // succeeded but the value generator still didn’t recognize the step value.
                     stepValueToRequire = _step;
+                }
 
                 step = valueGenerator.CreateStep(stepValueToRequire);
             }
--- a/src/NUnitFramework/framework/Attributes/RangeAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RangeAttribute.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
@@ -24,8 +24,6 @@
 using System;
 using System.Collections;
 using System.Reflection;
-
-using NUnit.Compatibility;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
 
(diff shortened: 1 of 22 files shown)


jnm2:
@nunit/framework-team Ready for review!

src/NUnitFramework/framework/Internal/ValueGenerator.cs
```diff
@@ -0,0 +1,140 @@
...
+            step = null;
+            return false;
+        }
+    }
+
+    internal abstract partial class ValueGenerator
+    {
+        public abstract IEnumerable GenerateRange(object start, object end, Step step);
+
+        private static readonly MethodInfo GenericCreateMethod =
```
mikkelbu: This should be lowercase accordingly to .editorconfig?

src/NUnitFramework/tests/Attributes/RangeAttributeTests.cs
```diff
@@ -22,478 +22,489 @@
...
+            typeof(double),
+            typeof(decimal)
+        };
+
+        private static readonly Type[] Int32RangeConvertibleToParameterTypes = { typeof(int), typeof(sbyte), typeof(byte), typeof(short), typeof(decimal) };
+        private static readonly Type[] UInt32RangeConvertibleToParameterTypes = { typeof(uint) };
+        private static readonly Type[] Int64RangeConvertibleToParameterTypes = { typeof(long) };
+        private static readonly Type[] UInt64RangeConvertibleToParameterTypes = { typeof(ulong) };
+        private static readonly Type[] SingleRangeConvertibleToParameterTypes = { typeof(float) };
+        private static readonly Type[] DoubleRangeConvertibleToParameterTypes = { typeof(double), typeof(decimal) };
```
mikkelbu: Shouldn't these start with a lowercase? (also it took me some time to find the connection between these and `ParamAttributeTypeConversions.TryConvert` 😄 .

src/NUnitFramework/tests/Attributes/RangeAttributeTests.cs
```diff
@@ -22,478 +22,489 @@
...
 
-        private void MethodWithFloatRangeAndZeroStep([Range(0.7f, 1.2f, 0.0f)] float x) { }
-
         [Test]
-        public void FloatRangeAndStep_Reversed()
+        public static void MaxValueRange_Decimal()
         {
-            Assert.Throws<ArgumentException>(() => CheckValuesWithinTolerance(nameof(MethodWithFloatRangeAndStep_Reversed), 0.7f, 0.9f, 1.1f));
-        }
+            const double step = (double)(((decimal)LargestDoubleConvertibleToDecimal - (decimal)NextLargestDoubleConvertibleToLowerDecimal) / 2);
```
mikkelbu: If we extracted `(decimal)NextLargestDoubleConvertibleToLowerDecimal` into a variable (with a shorter name) then perhaps the test would be easier to read?

src/NUnitFramework/tests/Attributes/RangeAttributeTests.cs
```diff
@@ -22,478 +22,489 @@
...
         }
-        
-        private void MethodWithIntRangeAndDecimalArgument([Range(1, 3)] decimal x) { }
 
         [Test]
-        public void CanConvertDoubleRangeToDecimal() 
+        public static void MinValueRange_Decimal()
         {
-            CheckValues(nameof(MethodWithDoubleRangeAndDecimalArgument), 1.0M, 1.1M, 1.2M);
+            const double step = (double)((-(decimal)LargestDoubleConvertibleToDecimal - -(decimal)NextLargestDoubleConvertibleToLowerDecimal) / 2);
```
mikkelbu: If we extracted `(decimal)NextLargestDoubleConvertibleToLowerDecimal` into a variable (with a shorter name) then perhaps the test would be easier to read? (also I actually had to debug the tests to have an idea of the value of `step` 😄 .

jnm2:
@mikkelbu Wow, very helpful feedback! I really appreciate this!

> This should be lowercase accordingly to .editorconfig?

The .editorconfig doesn't check private fields because we have too much variation. I'm used to `static readonly` and `const` fields using PascalCase but we don't [mention any exceptions](https://github.com/nunit/docs/wiki/Coding-Standards#naming), so I'll go ahead and make them `_camelCase`.

More to follow!

jnm2:
Going to wait on  https://github.com/nunit/nunit/issues/2822 for the naming thing.

src/NUnitFramework/framework/Internal/ValueGenerator.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        public sealed class ComparableStep<TStep> : Step where TStep: IComparable
+        {
+            private readonly TStep _step;
+            private readonly Func<T, TStep, T> _apply;
+
+            public ComparableStep(TStep value, Func<T, TStep, T> apply)
+            {
+                if (apply == null) throw new ArgumentNullException(nameof(apply));
+                _step = value;
```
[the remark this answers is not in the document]
jnm2: This code is agnostic about whether `null` might be a valid value. That's something the caller can decide.

src/NUnitFramework/framework/Internal/ValueGenerator.Int16ValueGenerator.cs
```diff
@@ -0,0 +1,79 @@
...
+            }
+
+            public override bool TryCreateStep(object value, out ValueGenerator.Step step)
+            {
+                if (value is short)
+                {
+                    step = new ComparableStep<short>((short)value, (prev, stepValue) => unchecked((short)(prev + stepValue)));
+                    return true;
+                }
+
```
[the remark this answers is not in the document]
jnm2: @mikkelbu 📄 Great question. Going to document this in the code for sure!

ByteValueGenerator is unusual in this regard because we allow byte parameter ranges to start high and end low, and internally the step is represented as (int)(-1) since it can't be represented as a `byte`. `-1` can be converted natively to `Int16`, `SByte` and `Decimal` so we don't need to do anything special for them.

src/NUnitFramework/framework/Attributes/RangeAttribute.cs
```diff
@@ -218,7 +174,34 @@ public RangeAttribute(float from, float to, float step)
...
+            {
+                // ValueGenerator.CreateStep is responsible to enable incrementing bytes by (int)(-1),
+                // or perhaps in the future a DateTime by a TimeSpan, but the responsibility to convert
+                // attribute values from Double to Decimal is in ParamAttributeTypeConversions.
+                // (ParamAttributeTypeConversions tries to simulate what would happen in C# if the
+                // literal syntax used for the attribute value had been used as a parameter argument
+                // in a direct call to the test method.)
+                object stepValueToRequire;
+                if (!ParamAttributeTypeConversions.TryConvert(_step, parameter.ParameterType, out stepValueToRequire))
+                    stepValueToRequire = _step;
```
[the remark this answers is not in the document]
jnm2: Yes. This will cause CreateStep to throw the same exception as it would throw if TryConvert succeeded but the value generator still didn’t recognize the step value. 📄

src/NUnitFramework/framework/Internal/ValueGenerator.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+
+        public abstract class Step
+        {
+            public abstract bool IsPositive { get; }
+            public abstract bool IsNegative { get; }
+        }
+
+        public abstract Step CreateStep(object value);
+        public abstract bool TryCreateStep(object value, out Step step);
```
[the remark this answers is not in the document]
jnm2: 📄

src/NUnitFramework/framework/Attributes/RangeAttribute.cs
```diff
@@ -218,7 +174,34 @@ public RangeAttribute(float from, float to, float step)
...
-            return ParamAttributeTypeConversions.ConvertData(_data, parameter.ParameterType);
+            var from = ParamAttributeTypeConversions.Convert(_from, parameter.ParameterType);
+            var to = ParamAttributeTypeConversions.Convert(_to, parameter.ParameterType);
+
+            var valueGenerator = ValueGenerator.Create(parameter.ParameterType);
+
+            ValueGenerator.Step step;
+            if (!valueGenerator.TryCreateStep(_step, out step))
+            {
+                // ValueGenerator.CreateStep is responsible to enable incrementing bytes by (int)(-1),
```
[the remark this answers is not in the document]
jnm2: It's a bit awkward, perfect feedback. 📄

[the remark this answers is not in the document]
jnm2: It's a bit awkward, perfect feedback. 📄

> ValueGenerator.CreateStep has the responsibility to enable Byte values to be incremented by the Int32 value -1. Or perhaps in the future, DateTime values to be incremented by a TimeSpan. It handles scenarios where the incrementing type is naturally fundamentally different.
> 
> However, ParamAttributeTypeConversions has the responsibility to convert attribute arguments that are only of a different type due to limitations on what can be represented as a constant in IL. E.g., Double values that would have been Decimal values except that Decimal values cannot be used as attribute arguments. (See the XML docs for the ParamAttributeTypeConversions class.)

src/NUnitFramework/tests/Attributes/RangeAttributeTests.cs
jnm2: I'll add a comment pointing to the XML docs I'm adding on ParamAttributeTypeConversions. 📄

src/NUnitFramework/tests/Attributes/RangeAttributeTests.cs
jnm2: I don't have the slightest clue what the value of `step` is. 😳

src/NUnitFramework/tests/Attributes/RangeAttributeTests.cs
jnm2: I ended up doing something similar. Does https://github.com/nunit/nunit/pull/2805/commits/e5cc4c8d26b42ee8d50efeaf7e5f5615deb4ab89 make it easy to read?

jnm2:
@mikkelbu 

> So far I've only added a lot of questions and comments. My main "concern" (it sounds worse than it is) is that there are a lot of code duplication between the individual classes deriving from `ValueGenerator<T>`. E.g. `Int16ValueGenerator` and `Int32ValueGenerator` have the same implementation of `GenerateRange` (except for the signature), and the twp `TryCreateStep`s are also quite similar. 
> 
> Can we reduce the duplication by extracting the common logic (or introducing a "ValueGenerator"-hierarchy)?

Yes! Since I had to extract the add operation out of it, I suddenly realized that the only operation left is comparison.

> For `DecimalValueGenerator` we throw an exception if the step does not monotonically increase, but here it seems that we allow that `next == current` ? (creating an infinite loop)

> Not the same handling as `DecimalValueGenerator` wrt. overflow. Why is that?

This is because all the integers used `unchecked` so that no OverflowException would be thrown, which is more expensive. This check was originally to detect the overflow and exit. When I got to Decimal, rather than removing the `if` statement which would have been the consistent thing to do, I turned it into a diagnostic check without adding diagnostic checks to all the others.

So see the last commit, where I make things simple the way they should have been.

Ready for review again. Great comments, keep them coming please!

src/NUnitFramework/framework/Internal/ValueGenerator.cs
```diff
@@ -0,0 +1,253 @@
...
+namespace NUnit.Framework.Internal
+{
+    internal abstract class ValueGenerator<T> : ValueGenerator
+    {
+        private static Exception CreateNotSupportedException(string description)
+        {
+            return new NotSupportedException($"{typeof(T)} is using the default value generator which does not support {description}.");
+        }
+
+        public virtual int Compare(T x, T y)
```
mikkelbu: Do you already have an idea for which types we would need to override this? (or is it just to make it possible to extend this class later on).

jnm2: Also, this is a correct but less efficient default implementation. Each comparer could override this with `=> x.CompareTo(y);` for a performance boost since neither `!typeof(IComparable<T>).IsAssignableFrom(typeof(T))` nor `Comparer<T>.Default.Compare(x, y)` are detected as constant by the JIT. Because I would rather keep it simple and wait to add code till a performance need is identified, I didn't override it. So this is kind of a note for the future.

As to whether I should have left the method sealed, maybe? But if we ever create a value generator for things that aren't IComparable, it'll need to be overridden. I don't know, do you have a feeling one way or the other?

jnm2: This is a correct but less efficient default implementation. Each comparer could override this with `=> x.CompareTo(y);` for a performance boost since neither `!typeof(IComparable<T>).IsAssignableFrom(typeof(T))` nor `Comparer<T>.Default.Compare(x, y)` are detected as constant by the JIT. Because I would rather keep it simple and wait to add code till a performance need is identified, I didn't override it. So this is kind of a note for the future.

As to whether I should have left the method sealed until such need arises, maybe? If we ever create a value generator for things that aren't IComparable, it'll need to be overridden. I don't know, do you have a feeling one way or the other?

mikkelbu: I think it is fine as it is. I was just wondering when I saw `virtual`.

jnm2: Okay, thanks for commenting. Sounds like I'm good to merge then. 👍

src/NUnitFramework/tests/Attributes/RangeAttributeTests.cs
jnm2: Looping back to say that the outcome of https://github.com/nunit/nunit/issues/2822 was to allow this naming convention.

PR #2807, opened 2018-04-13
Removed accidental call to ValueType.GetHashCode while overriding

jnm2:
I was sure I had deleted this line. The point of overriding GetHashCode is to avoid incurring the performance hit of ValueType.GetHashCode.


rprouse:
Waiting for CI 👍

PR #2814, opened 2018-04-18
Remove public marker types

jnm2:
Gets rid of some of our public surface area that was implementation-detaily, `NUnitNullType` and `NonmatchingType`. No behavioral changes intended.

The whitespace diff was so extreme that I fixed it in an initial commit so that you can skip that and focus on the real changes.


mikkelbu:
@jnm2 It looks correct [on the commits on master](https://github.com/nunit/nunit/commits/master), but [on the PR](https://github.com/nunit/nunit/pull/2814/commits) it look like it is sorting the commits by time? I think that this explains why [Why are my commits in the wrong order?](https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/)

PR #2816, opened 2018-04-18
Scoped the modifiers to convey intended uses (NUnit.Framework.Constraints)

jnm2:
In the same vein as https://github.com/nunit/nunit/pull/2804.

ℹ This is not why I think `sealed` is important, but a side effect of this is that recent JITs generate faster code when calling methods of sealed classes.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionTally.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionTally.cs
@@ -28,10 +28,10 @@ namespace NUnit.Framework.Constraints
 {
     /// <summary><see cref="CollectionTally"/> counts (tallies) the number of occurrences 
     /// of each object in one or more enumerations.</summary>
-    public class CollectionTally
+    public sealed class CollectionTally
     {
         /// <summary>The result of a <see cref="CollectionTally"/>.</summary>
-        public class CollectionTallyResult
+        public sealed class CollectionTallyResult
         {
             /// <summary>Items that were not in the expected collection.</summary>
             public List<object> ExtraItems { get; set; }
--- a/src/NUnitFramework/framework/Constraints/CollectionTally.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionTally.cs
@@ -34,16 +34,16 @@ namespace NUnit.Framework.Constraints
         public sealed class CollectionTallyResult
         {
             /// <summary>Items that were not in the expected collection.</summary>
-            public List<object> ExtraItems { get; set; }
+            public List<object> ExtraItems { get; }
 
             /// <summary>Items that were not accounted for in the expected collection.</summary>
-            public List<object> MissingItems { get; set; }
+            public List<object> MissingItems { get; }
 
-            /// <summary>Constructs an empty <see cref="CollectionTallyResult"/>.</summary>
-            public CollectionTallyResult()
+            /// <summary>Initializes a new instance of the <see cref="CollectionTallyResult"/> class with the given fields.</summary>
+            public CollectionTallyResult(List<object> missingItems, List<object> extraItems)
             {
-                ExtraItems = new List<object>();
-                MissingItems = new List<object>();
+                MissingItems = missingItems;
+                ExtraItems = extraItems;
             }
         }
 
@@ -54,11 +54,9 @@ namespace NUnit.Framework.Constraints
         {
             get
             {
-                return new CollectionTallyResult()
-                {
-                    MissingItems = new List<object>(_missingItems),
-                    ExtraItems = new List<object>(_extraItems)
-                };
+                return new CollectionTallyResult(
+                    new List<object>(_missingItems),
+                    new List<object>(_extraItems));
             }
         }
 
--- a/src/NUnitFramework/framework/Constraints/ConstraintBuilder.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintBuilder.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
@@ -30,18 +30,18 @@ namespace NUnit.Framework.Constraints
     /// ConstraintBuilder maintains the stacks that are used in
     /// processing a ConstraintExpression. An OperatorStack
     /// is used to hold operators that are waiting for their
-    /// operands to be reorganized. a ConstraintStack holds 
+    /// operands to be reorganized. a ConstraintStack holds
     /// input constraints as well as the results of each
     /// operator applied.
     /// </summary>
-    public class ConstraintBuilder : IResolveConstraint
+    public sealed class ConstraintBuilder : IResolveConstraint
     {
         #region Nested Operator Stack Class
 
         /// <summary>
         /// OperatorStack is a type-safe stack for holding ConstraintOperators
         /// </summary>
-        public class OperatorStack
+        private sealed class OperatorStack
         {
             private readonly Stack<ConstraintOperator> stack = new Stack<ConstraintOperator>();
 
@@ -96,7 +96,7 @@ namespace NUnit.Framework.Constraints
         /// <summary>
         /// ConstraintStack is a type-safe stack for holding Constraints
         /// </summary>
-        public class ConstraintStack
+        public sealed class ConstraintStack
         {
             private readonly Stack<IConstraint> stack = new Stack<IConstraint>();
             private readonly ConstraintBuilder builder;
@@ -121,7 +121,7 @@ namespace NUnit.Framework.Constraints
 
             /// <summary>
             /// Pushes the specified constraint. As a side effect,
-            /// the constraint's Builder field is set to the 
+            /// the constraint's Builder field is set to the
             /// ConstraintBuilder owning this stack.
             /// </summary>
             /// <param name="constraint">The constraint to put onto the stack</param>
@@ -186,7 +186,7 @@ namespace NUnit.Framework.Constraints
 
             // Reduce any lower precedence operators
             ReduceOperatorStack(op.LeftPrecedence);
-            
+
             ops.Push(op);
             lastPushed = op;
         }
--- a/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
@@ -45,7 +45,7 @@ namespace NUnit.Framework.Constraints
         /// <summary>
         /// The ConstraintBuilder holding the elements recognized so far
         /// </summary>
-        protected ConstraintBuilder builder;
+        protected readonly ConstraintBuilder builder;
 
         #endregion
 
@@ -54,9 +54,8 @@ namespace NUnit.Framework.Constraints
         /// <summary>
         /// Initializes a new instance of the <see cref="ConstraintExpression"/> class.
         /// </summary>
-        public ConstraintExpression()
+        public ConstraintExpression() : this(new ConstraintBuilder())
         {
-            this.builder = new ConstraintBuilder();
         }
 
         /// <summary>
@@ -66,6 +65,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="builder">The builder.</param>
         public ConstraintExpression(ConstraintBuilder builder)
         {
+            Guard.ArgumentNotNull(builder, nameof(builder));
             this.builder = builder;
         }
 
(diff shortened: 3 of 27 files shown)


src/NUnitFramework/framework/Assume.cs
```diff
@@ -34,7 +34,7 @@ namespace NUnit.Framework
     /// result. If an assumption is not met, the test
     /// should produce an inconclusive result.
     /// </summary>
-    public class Assume
+    public abstract class Assume
```
jnm2: This would be `static` (= `sealed abstract`) but we want to allow inheritance in these particular classes as an extensibility mechanism since C# does not allow static extension members.

src/NUnitFramework/framework/Attributes/RepeatAttribute.cs
```diff
@@ -72,7 +72,7 @@ public TestCommand Wrap(TestCommand command)
         /// <summary>
         /// The test command for the RepeatAttribute
         /// </summary>
-        public class RepeatedTestCommand : DelegatingTestCommand
+        private sealed class RepeatedTestCommand : DelegatingTestCommand
```
jnm2: This class provides no new surface area. Extending code would use `RepeatAttribute.Wrap` to obtain these instances and use composition to delegate to them.

src/NUnitFramework/framework/Compatibility/LongLivedMarshalByRefObject.cs
```diff
@@ -30,11 +30,11 @@ namespace NUnit.Compatibility
     /// A MarshalByRefObject that lives forever
     /// </summary>
 #if NETSTANDARD1_6
-    public class LongLivedMarshalByRefObject
+    public abstract class LongLivedMarshalByRefObject
     {
     }
 #else
-    public class LongLivedMarshalByRefObject : MarshalByRefObject
+    public abstract class LongLivedMarshalByRefObject : MarshalByRefObject
```
jnm2: ⚠

src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -28,22 +28,22 @@ namespace NUnit.Framework.Constraints
...
-        public class CollectionTallyResult
+        public sealed class CollectionTallyResult
         {
             /// <summary>Items that were not in the expected collection.</summary>
-            public List<object> ExtraItems { get; set; }
+            public List<object> ExtraItems { get; }
 
             /// <summary>Items that were not accounted for in the expected collection.</summary>
-            public List<object> MissingItems { get; set; }
+            public List<object> MissingItems { get; }
```
jnm2: Rather than allowing two ways to mutate MissingItems, I went along with the way we were currently doing it.

src/NUnitFramework/framework/Constraints/ConstraintBuilder.cs
```diff
@@ -34,14 +34,14 @@ namespace NUnit.Framework.Constraints
     /// input constraints as well as the results of each
     /// operator applied.
     /// </summary>
-    public class ConstraintBuilder : IResolveConstraint
+    public sealed class ConstraintBuilder : IResolveConstraint
```
jnm2: As far as I can tell, there is no useful way to inherit from this class. If a class is not specifically designed with inheritance in mind, we should signal that like we do with readonly on fields. Inheritance comes with often-unintended consequences. (A simple example is that straight-forward custom equality comparisons stop being symmetric).

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -45,7 +45,7 @@ public class ConstraintExpression
         /// <summary>
         /// The ConstraintBuilder holding the elements recognized so far
         /// </summary>
-        protected ConstraintBuilder builder;
+        protected readonly ConstraintBuilder builder;
```
jnm2: Noticed that I missed this in the `readonly` PR. Maybe I should put this in its own PR?

src/NUnitFramework/framework/Constraints/Tolerance.cs
```diff
@@ -324,20 +324,20 @@ private Range PercentRange(object value)
         /// Tolerance.Range represents the range of values that match
         /// a specific tolerance, when applied to a specific value.
         /// </summary>
-        public class Range
+        public struct Range
```
jnm2: Since the purpose is to be a value bag rather than to have identity.

src/NUnitFramework/framework/Internal/TestListener.cs
```diff
@@ -21,46 +21,28 @@
...
 namespace NUnit.Framework.Internal
 {
     /// <summary>
-    /// TestListener provides an implementation of ITestListener that
-    /// does nothing. It is used only through its NULL property.
+    /// Provides an implementation of <see cref="ITestListener"/> through
+    /// its <see cref="NULL"/> property that does nothing.
     /// </summary>
-    public class TestListener : ITestListener
+    public static class TestListener
```
jnm2: The side effect here is that .ToString() gives `NullTestListener` and you check `== TestListener.NULL` rather than `is TestListener`. I was motivated to do this since `is TestListener` doesn't read as though you are checking for a *null* test listener.

ChrisMaddock:
Eee there's a lot to public classes here! 🙀 

What was your logic as to where to make changes here - are you applying this to everything that 'would' be internal (let's avoid the debate on terminology!), or just everything which isn't extended within the framework itself? Or is it classes that make no logical sense to extend? Is the aim also to restrict what we currently expose, or just the semantics/JIT benefits? 

Main question: how do we know this isn't going to break builds, and if we don't know that, what's the justification that makes it worthwhile? 😄

jnm2:
@ChrisMaddock My motivation was to improve the signal to noise ratio especially when browsing our APIs but also when working in the codebase. That includes signalling our semantics and reducing anything we expose which is not designed for public consumption.

I didn't want to really tackle the visibility aspects much in this PR. I'm thinking that should be done namespace by namespace, or even smaller.
The logic in this PR is *mostly* that only classes that are designed for inheritance are left unsealed. It's documentation letting you know that there wouldn't have been anything useful to gain by inheriting. (Or in the case of `abstract`/`static`, by instantiating.)

ChrisMaddock:
Thanks for clarifying! 🙂 

I completely agree this is where we _want_ to be. My concern is how we get there, causing the least disruption to legacy builds. This sort of incremental change, and picking off the low-hanging fruit may be the right way - I guess the other option is bulk making things private. The only issue is that whenever we start touching things like this, we invariably seem to end up breaking _some_ unusual use-case implemented 10 years ago! Maybe it's more down to us to say 'Sorry - that's unsupported' - at least this will move us in the right direction to need to say that less often!

There's a lot of changes here. I'll try and look through at some point.

src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -28,22 +28,22 @@ namespace NUnit.Framework.Constraints
...
 
             /// <summary>Constructs an empty <see cref="CollectionTallyResult"/>.</summary>
-            public CollectionTallyResult()
+            public CollectionTallyResult(List<object> missingItems, List<object> extraItems)
             {
-                ExtraItems = new List<object>();
-                MissingItems = new List<object>();
+                MissingItems = missingItems;
+                ExtraItems = extraItems;
```
ChrisMaddock: Why this change? And if this change, does the doccoment need updating too - as surely this allows you to potentially create a non-empty CollectionTallyResult

ChrisMaddock:
Suggestion for next time - would you mind breaking wide-spread changes like this up into multiple PRs? That would allow us to review/merge 20 at a time or so - most of my time is available in dribs and drabs, and this is a difficult PR to part-review. Thanks! 😄

jnm2:
> Suggestion for next time - would you mind breaking wide-spread changes like this up into multiple PRs? That would allow us to review/merge 20 at a time or so - most of my time is available in dribs and drabs, and this is a difficult PR to part-review. Thanks! 😄

Good idea. Maybe per namespace? It's not too late for me to break this one up...

src/NUnitFramework/framework/Constraints/CollectionTally.cs
jnm2: It allowed
```cs
return new CollectionTallyResult()
{
    MissingItems = new List<object>(_missingItems),
    ExtraItems = new List<object>(_extraItems)
};
```
to become:
```cs
return new CollectionTallyResult(
    new List<object>(_missingItems),
    new List<object>(_extraItems));
```
rather than:
```cs
var temp = new CollectionTallyResult();
temp.MissingItems.AddRange(_missingItems);
temp.ExtraItems.AddRange(_extraItems);
return temp;
```

Thanks for catching the doc comment! What do you think, is this too far? Should we go back to allowing the list instances (themselves mutable) to be swapped out, even though we only used that for initialization? What would it mean to later on set them to null or a new list, as opposed to clearing them?

ChrisMaddock:
If you wouldn't mind - that would be very helpful for me. I don't know most of these type names well enough to provide any sort of useful review just from looking at the diff - and I rarely have time to look up 155 different classes in the codebase! 

@rprouse or @CharliePoole may be happier to review this as is though, based on their familiarity of the codebase however!

jnm2:
I would not mind and I'm happy you brought it up! I'll make this PR just about NUnit.Framework.Constraints since that's where you've commented so far.

jnm2:
> The only issue is that whenever we start touching things like this, we invariably seem to end up breaking some unusual use-case implemented 10 years ago! Maybe it's more down to us to say 'Sorry - that's unsupported' - at least this will move us in the right direction to need to say that less often!

There's part of me that wants the pain to be over sooner rather than later. But that doesn't mean this is the right approach. If anything just bothers you, I'd rather not have it in the PR.

jnm2:
PR is updated! Tell me what you think of the list of commits.
I'm going to wait to submit the others until people have had a chance to digest this one.

src/NUnitFramework/framework/Constraints/FloatingPointNumerics.cs
```diff
@@ -45,9 +44,8 @@ namespace NUnit.Framework.Constraints
     ///     as low as 0.0000001 for small numbers or as high as 10.0 for large numbers.
     ///   </para>
     /// </remarks>
-    public class FloatingPointNumerics
+    internal static class FloatingPointNumerics
```
jnm2: Due to the inaccessible constructor, this class was already effectively `static` from the outside.

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -66,6 +65,7 @@ public ConstraintExpression()
         /// <param name="builder">The builder.</param>
         public ConstraintExpression(ConstraintBuilder builder)
         {
+            Guard.ArgumentNotNull(builder, nameof(builder));
             this.builder = builder;
```
jnm2: All the code in this class assumes that `builder` is not null. Now that the field is readonly and this guard is in place, that is finally a safe assumption.

src/NUnitFramework/framework/Constraints/CollectionTally.cs
ChrisMaddock: I understand, thanks. Sounds sensible - maybe just the comment to change. 😄

src/NUnitFramework/framework/Constraints/ConstraintBuilder.cs
ChrisMaddock: Sounds ok to me

src/NUnitFramework/framework/Constraints/ConstraintBuilder.cs
```diff
@@ -96,7 +96,7 @@ public ConstraintOperator Pop()
         /// <summary>
         /// ConstraintStack is a type-safe stack for holding Constraints
         /// </summary>
-        public class ConstraintStack
+        public sealed class ConstraintStack
```
ChrisMaddock: Is there a reason to not make this internal in-line with the change to OperatorStack?

src/NUnitFramework/framework/Constraints/FloatingPointNumerics.cs
ChrisMaddock: It looks like lots of this class isn't used by the framework. I'm not sure on it's origins - are the public methods designed to be used by end users, or has internal usage of it just lapsed? 

The former would be a reason not to make it internal, of course. `static` seems fine to me.

src/NUnitFramework/framework/Constraints/ConstraintBuilder.cs
jnm2: ConstraintOperator has `public abstract void Reduce(ConstraintBuilder.ConstraintStack stack)`.
My bad, `OperatorStack` could have been `private`. I'll change it.

jnm2:
@ChrisMaddock Forgot to ping you after the updates, sorry.

ChrisMaddock:
Looks good. My only concern is not understanding the origin/purpose of the FloatingPointNumerics class. (It being tested so heavily but unused makes me wonder if it was previously available to users, and legacy test suites may still rely on it.) @rprouse or @CharliePoole may know better there. 🙂

CharliePoole:
`FloatingPointNumerics` is used by `Numerics` and comes into play whenever you test floating point values for equality `.Within(n).Ulps`. That's fairly esoteric but you'd be surprised at how much discussion of it there was when I implemented it initially. The class was never aimed at direct user usage.

mikkelbu:
@jnm2 It looks as though AppVeyor is confused (perhaps due to a force push?). 
image

jnm2:
Thanks! I went to https://ci.appveyor.com/project/CharliePoole/nunit/history, used the browser to find the first occurrence of "2816", opened the build and clicked "Re-build PR." Looks like that has fixed it.

PR #2818, opened 2018-04-21
Implemented the ability to await in STA tests 🎉

### LINKED ISSUE #1200: async test + Apartment(ApartmentState.STA) => await not returning on STA thread

jnm2:
Fixes #1200, re: https://github.com/nunit/nunit/issues/2774

Whew! This was fun and the collateral was worth doing, but this was a longer time coming than I expected!

The heart and soul of this change is the 8th (last) commit. Its motivating tests are the 6th commit.

The 7th commit took a lot of thought to get right while keeping the complexity down. It refactors AsyncInvocationRegion—which as I've noticed many times since I arrived at this project, had become neither async, an invocation, nor a region 😆—due to the necessity for it to control the invocation itself rather than being handed the return value from the invocation. I did my best to come up with a descriptive name, `AsyncToSyncAdapter`, since it adapts async execution to synchronous execution.

The preceding commits were things which were all in the critical path, so for simplicity they are in the same PR. They should each be viewed separately.

@rprouse Would it be a good idea for me to do a walkthrough explanation of the 8th commit for posterity, or do #1200 and #2774 explain it well enough?

/cc @sharwell


rprouse:
@jnm2 I am not neglecting this, I just sold a house AND I'm swamped at work right now. I will review when my head is clearer.

jnm2:
@rprouse I appreciate that. Don't worry about it! I also have my hands full. I'd rather have your clear-headed review. 😜 And I don't mind having an excuse to do better documentation or write this one up. 

(I'm thinking: I've been programming lazily and hoping people will just tell me where my code is hard to understand so that I can improve it. But that may not be fair on you...)

jnm2:
I'm adding a simple ManualResetEventSlim polyfill in this one, but I built a [commit](https://github.com/nunit/nunit/commit/213a47156267a97d7b34540b846ad5f628c76b38) on top of it with full-blown lock-free CountdownEvent and spinning/lazy ManualResetEventSlim to see if that improves the throughput and gets rid of all the performance issues on net35 CI.

src/NUnitFramework/framework/Assert.Exceptions.cs
```diff
@@ -41,25 +41,7 @@ public static Exception Throws(IResolveConstraint expression, TestDelegate code,
...
-                    }
-                    catch (Exception e)
-                    {
-                        caughtException = e;
-                    }
-                }
-            }
-            else
-#endif
+            Guard.ArgumentNotAsyncVoid(code, nameof(code));
```
rprouse: Shouldn't we be guarding that code shouldn't be `async` period?

src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
```diff
@@ -0,0 +1,150 @@
...
+            return IsAsyncOperation(@delegate.GetMethodInfo());
+        }
+
+        private static bool IsTaskType(Type type)
+        {
+            for (; type != null; type = type.GetTypeInfo().BaseType)
+            {
+                if (type.GetTypeInfo().IsGenericType
+                    ? type.GetGenericTypeDefinition().FullName == "System.Threading.Tasks.Task`1"
+                    : type.FullName == "System.Threading.Tasks.Task")
```
rprouse: What about `ValueTask`?

jnm2: I'm thinking of this as the second async PR, and had planned ValueTask along with general awaitable support in the third PR.

src/NUnitFramework/framework/Assert.Exceptions.cs
jnm2: `code` is `TestDelegate` which is declared as `delegate void`, so if it's ever async, it's async void. I should add a comment to make that clear.

jnm2: `code` is `TestDelegate` which is declared as `delegate void`, so if it's ever async, it's async void. I added a comment to make that clear.

jnm2:
> Looks good to me although it is a lot to digest. I've asked a couple of questions, so I am not approving yet. @ChrisMaddock or @mikkelbu are either of you up to a second review? This is a large change, so the more eyes the better.

I would like a second one. Rob, from your review, do you think it would benefit from me commenting the code?

ChrisMaddock:
I can try and have a look next week - but I know very little about .NET async, so am probably the wrong person for it!

jnm2:
@ChrisMaddock Thanks for being willing to look! I can think of two-way benefits that can come from reviewing even so. You might catch a typo or think of an NUnit async constraint I didn't cover, and this way you have a general sense of where code is and what it's called. And if you find yourself curious about async, I enjoy explaining things like that. (Whether I'm good at explaining it might be another question. 😄)

jnm2:
@ChrisMaddock I wrote it up for you anyway! 😄

#### All you need to know about async for this PR:

 - When you use `await foo`, your method returns back to the caller _without_ running the rest of the method. 

 - Instead, your method wraps the rest of itself in an `Action` so that it can ask `foo` to schedule that `Action` to run when `foo` is completed.

 - When `foo` is a `Task`, `foo` avoids invoking the rest-of-the-method `Action` _directly_ when it completes because `foo` might be being completed by a different thread. (Think TaskCompletionSource.SetResult.)
   Instead, `foo` uses `SynchronizationContext.Current.Post` to schedule the `Action` so that the rest of the method executes in the same context as the part of the method before the `await`. (This is why after awaiting in Windows Forms, the rest of your method executes on the main UI thread instead of executing on the thread that called TaskCompletionSource.SetResult.)

#### And all you need to know about SynchronizationContext:

 - The default synchronization context doesn't synchronize anything. Everything you `Post` to it it immediately turns around and gives to `ThreadPool.QueueUserWorkItem`.  This is why if you are on an STA thread without a synchronization context and you await, the rest of your method runs on a random **non-STA** background thread. **This is the problem this PR fixes.**

 - Context means whatever you want it to, if you are the implementer of a SynchronizationContext. People ask you to run stuff by calling your implementation of Post. You get to decide if you want to run each action now or later, or eleven times, or not at all, or get a particular thread to do it, etc.

 - To fix the problem, I just implemented a SynchronizationContext which [throws everything in a queue](https://github.com/nunit/nunit/pull/2818/commits/4be14d688bb6e5948157062a58eb4e2db2eebbb3#diff-db1f4b93b0fc60264c8b3eec019cf2f6R45). Later, while NUnit's STA worker thread is waiting for the async method to finish, rather than blocking the whole time, the STA worker thread [gets busy executing everything in the queue](https://github.com/nunit/nunit/pull/2818/commits/4be14d688bb6e5948157062a58eb4e2db2eebbb3#diff-db1f4b93b0fc60264c8b3eec019cf2f6R106) until the test method is finished.


The rest is plumbing: [AsyncToSyncAdapter (formerly AsyncInvocationRegion) gets](https://github.com/nunit/nunit/pull/2818/commits/4be14d688bb6e5948157062a58eb4e2db2eebbb3#diff-6d621f6b800a90cf2858a8cd264b2957R85) a [SingleThreadedTestMessagePumpStrategy](https://github.com/nunit/nunit/pull/2818/commits/4be14d688bb6e5948157062a58eb4e2db2eebbb3#diff-308ac1319e4d291401e5ccd48a2b3651R40) and calls [WaitForCompletion](https://github.com/nunit/nunit/pull/2818/commits/4be14d688bb6e5948157062a58eb4e2db2eebbb3#diff-308ac1319e4d291401e5ccd48a2b3651R136) which is implemented exactly like the Windows Forms and WPF strategies.

And to get the test method to `await` using our synchronization context instead of the default one, we have to [set the synchronization context](https://github.com/nunit/nunit/pull/2818/commits/4be14d688bb6e5948157062a58eb4e2db2eebbb3#diff-6d621f6b800a90cf2858a8cd264b2957R113) before invoking the test method.

jnm2:
@rprouse One of these days I'm going to write it up properly in the PR body. 😆

ChrisMaddock:
Thanks Joseph, that looks great! I'll try and take a look at this next week. 😄

jnm2:
I see this PR solves the same problem UIFact does, while https://github.com/nunit/nunit/pull/2776 solves what WinFormsFact and WpfFact do. I took a more hands-off approach. E.g. rather than installing a WindowsFormsSynchronizationContext, it just detects the one you have.

src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            return false;
+        }
+
+#if ASYNC
+        private const string TaskWaitMethod = "Wait";
+        private const string TaskResultProperty = "Result";
+        private const string VoidTaskResultType = "VoidTaskResult";
+        private const string SystemAggregateException = "System.AggregateException";
+        private const string InnerExceptionsProperty = "InnerExceptions";
```
ChrisMaddock: Are these strings mostly just used the once? Is there a reason not to inline them?

src/NUnitFramework/tests/AsyncExecutionApiAdapter.cs
```diff
@@ -0,0 +1,176 @@
...
+        {
+            new TaskReturningTestMethodAdapter(),
+            new AssertThrowsAsyncAdapter(),
+            new AssertDoesNotThrowAsyncAdapter(),
+            new AssertCatchAsyncAdapter(),
+            new ThrowsNothingConstraintAdapter(),
+            new ThrowsConstraintAdapter(),
+            new ThrowsExceptionConstraintAdapter(),
+            new NormalConstraintAdapter()
+        };
```
ChrisMaddock: I like these! 😄

src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
jnm2: This code was [moved](https://github.com/nunit/nunit/pull/2818/files/4be14d688bb6e5948157062a58eb4e2db2eebbb3#diff-c4545d57570ce4b1269a9f0adf8a4b18L99) without changing it.
In the next PR when I add support for general C# awaitables (and maybe F# too), I'll be pulling this stuff out anyway.

jnm2: Spot on. I'm going to leave it since this code was [moved](https://github.com/nunit/nunit/pull/2818/files/4be14d688bb6e5948157062a58eb4e2db2eebbb3#diff-c4545d57570ce4b1269a9f0adf8a4b18L99) without changing it, and in the next PR when I add support for general C# awaitables (and maybe F# too), I'll be pulling this stuff out anyway.

src/NUnitFramework/tests/AsyncExecutionApiAdapter.cs
jnm2: Oops, I just realized I forgot to cover setups and teardowns!

jnm2: Yeah, isn't this better than what I tried to do the last time? 😆
Oops, I just realized I forgot to cover setups and teardowns!

src/NUnitFramework/tests/TestUtilities/DummyConstraint.cs
```diff
@@ -0,0 +1,15 @@
+using NUnit.Framework.Constraints;
```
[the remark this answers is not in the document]
jnm2: 👍 Fixed.

jnm2:
@rprouse, @ChrisMaddock I rebased to fix the missing header Chris found, and then I added the last four commits containing some new simple tests and refactors. As soon as any team member reapproves, I'm happy for this to be merged.

jcansdale:
I was getting some strange results when using NUnit to test some related functionality in TerstDriven.Net (see https://github.com/jcansdale/TestDriven.Net-Issues/issues/110 for the gory details).

It turns out the following test fails in NUnit `3.10.1`:
```cs
[Test]
[Apartment(ApartmentState.STA)]
public async Task ApartmentStateAfterYield()
{
    await Task.Yield();
    Assert.That(Thread.CurrentThread.GetApartmentState(), Is.EqualTo(ApartmentState.STA));
}
```

This test passes when using the `nunit.framework` from this PR.

This PR would get a ✔️ from me. Thanks for your work on this @jnm2. I still find the task architecture a little mind bending and your write-ups have been appreciated. 👍

jnm2:
@jcansdale 

> This PR would get a ✔️ from me. Thanks for your work on this @jnm2. I still find the task architecture a little mind bending and your write-ups have been appreciated. 👍

Thanks! I see you're a member here, which is awesome... I don't know but I don't think there would be a problem with you approving the PR if you wanted?
My explanations are only oversimplifications of what I've absorbed from smarter folks years ago. I want to democratize `async`. 😁 

> BTW, I think this would make a great blog post and something for the NUnit twitter. What do you reckon @jnm2? 😄

Haha, I had just offered to help with this. There are at least three more fundamental changes beyond these two which I want to get in for 3.11:

 - support for any C# awaitables (last item in https://github.com/nunit/nunit/issues/2774)
 - maybe support for F#'s async
 - save and restore the synchronization context along with the test execution environment so that one test can never affect another (https://github.com/nunit/nunit/issues/2821)
 - Very hopeful: introduce an `AsyncAssert` class (https://github.com/nunit/nunit/issues/2843)

And a sweep of a dozen potentially-related existing issues.

jnm2:
However, @rprouse, don't wait for my my slow pace with these if you want to release 3.11 sooner. Master has been continuously deliverable as far as I know.

jcansdale:
> Thanks! I see you're a member here, which is awesome... I don't know but I don't think there would be a problem with you approving the PR if you wanted?

I've approved from an API consumer PoV. I'd love to review the code itself, but work, family etc. 😕

jnm2:
@jcansdale Understandable, no worries. Thanks for your feedback!

rprouse:
LGTM, let's get this merged 😄

jnm2:
If you'd like to take advantage of this early, look for version `3.11.0-dev-05580` or greater when you merge this with your nuget.config:

```xml
<configuration>
  <packageSources>
    <add key="NUnit MyGet" value="https://www.myget.org/F/nunit/api/v3/index.json" />
  </packageSources>
</configuration>
```

Belorus:
Hi @jnm2 

I tried package 05600 with your changes.
https://github.com/nunit/nunit/issues/2851 - here i get 
System.InvalidOperationException : Shutting down SingleThreadedTestSynchronizationContext with work still in the queue.

This behavior makes perfect sense in very strict env where you wish to avoid async void methods anywhere in your code.

It would be really amazing to make a new configurable behavior that executes all delegates in SynchronizaitonContext that are left. This will make testing of `async void`  methods possible.

WDYT ?

Belorus:
Also i've noticed that 
`
[assembly: Parallelizable(ParallelScope.Fixtures)]
[assembly: Apartment(ApartmentState.STA)]
`
is 3-5 times slower than just
`
[assembly: Parallelizable(ParallelScope.Fixtures)]
`

CharliePoole:
@Belorus Since there's only one STA, your first example is single-threaded for all practical purposes. If you had some non-STA fixtures, they would run on parallel. Are you sure all your tests need the STA?

Belorus:
n@CharliePoole Well, what my example needs is a SynchronizationContext that will run all delegates in its queue before letting test finish. This means that all async void methods have finished their work.

This sounds for me as a good behavior for most cases, if you expect people testing async void methods.

From my POV its absolutely legal to create 8 threads each with its own SynchronizationContext. I don't see reason to execute such tests sequentially.

CharliePoole:
Because STA is an old COM concept it may not be what you want. @jnm2 did all the recent work on synchronization contexts and will probably have to help us out here.

jnm2:
@Belorus Thanks for the questions! I replied about async void over where you initially asked: https://github.com/nunit/nunit/issues/2851#issuecomment-392324978.

> Also i've noticed that 
> ```
> [assembly: Parallelizable(ParallelScope.Fixtures)]
> [assembly: Apartment(ApartmentState.STA)]
> ````
> is 3-5 times slower than just
> ```
> [assembly: Parallelizable(ParallelScope.Fixtures)]
> ```

I would expect NUnit to be creating eight STA threads for you. So long as nothing COM-sensitive crosses from one thread to another, that's the ideal situation.
(Easy to avoid- don't put anything COM-sensitive into state that is shared between the threads, such as an instance field of a fixture with ParallelScope.Children.)

Issue #1792, opened 2016-09-14
TestCaseSource: "Too many arguments provided" for IEnumerable<> parameter type


ChrisMaddock:
@debugosaurus commented on [Wed Sep 14 2016](https://github.com/nunit/docs/issues/115)

It appears that NUnit has started treating Array and Enumerable TestCaseSource parameters differently. I noticed this behaviour when upgrading from 3.0 to 3.4.1.0 (sorry that covers a lot of ground).

Say I declare some test data:

```
protected static IEnumerable<IEnumerable<string>> EnumerableOfStrings => new[] 
{ 
    new[] { "one", "two" }
};
```

The following test will pass:

```
[Test, TestCaseSource(nameof(EnumerableOfStrings))]
public void AcceptsArrayOfStrings(string[] array)
{
}
```

However this test will fail with the message **"Too many arguments provided"**:

```
[Test, TestCaseSource(nameof(EnumerableOfStrings))]
public void AcceptsEnumerableOfStrings(IEnumerable<string> enumerable)
{
}
```

It seems with the second test that NUnit is attempting to unbox the enumerable and pass each element as an argument rather than passing the enumerable as a single parameter.

---

@ChrisMaddock commented on [Wed Sep 14 2016](https://github.com/nunit/docs/issues/115#issuecomment-247175801)

I can reproduce this - agreed it seems to be a bug. (Great bug report, by the way! :+1:)

I'll also move this issue over to the NUnit framework repo - this ones just for doc issues. :smile:

CharliePoole:
Not sure I agree this is a bug - or at least not the same bug you see. It seems to me that the source has to be __either__ for a method that takes a string or for one that takes an IEnumerable<string>. Therefore, both test methods can't be good. One has to be invalid, while the other is valid. I recall that we made some changes to the code to make it work as it does. Are you suggesting that the first one should give an error of "Not enough arguments" or that both should work?

This has always been an ambiguous situation for NUnit: only one argument presented as a single array. Should we expect methods to take a string or an array of strings? Is there some way we can make this clearer? It would be interesting to see what are the minimal syntax changes to the TestCaseSource that would cause NUnit to switch over to the alternate interpretation. What happens if we specify either string[] or object[]?

ChrisMaddock:
Not sure I'm following you here Charlie, I'm afraid.

> It seems to me that the source has to be either for a method that takes a string or for one that takes an IEnumerable

[...]

PR #2824, opened 2018-04-25
Only introspect one-dimensional arrays

### LINKED ISSUE #2823: SetUp failed for test fixture - Array was not a one-dimensional array.  Issue seems related to byte[,] method parameters
I'm in the process of updating from `3.8.1` to `3.10.1` and I'm having an issue where testfixtures that pass in 3.8.1, fail to "SetUp" in 3.10.1.

R# test runner refuses to even display the fixture sometimes, and when it does, test results are always "inconclusive, test not run".

I only discovered this issue for this particular fixture because i received the stacktrace (below) when running `dotnet test` from the command line.

Here's the stacktrace:

```
SetUp failed for test fixture MyFixture
System.ArgumentException: Array was not a one-dimensional array.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Array.GetValue(Int32 index)
   at NUnit.Framework.Internal.TestNameGenerator.NameFragment.GetDisplayString(Object arg, Int32 stringMax) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\TestNameGenerator.cs:line 241
   at NUnit.Framework.Internal.TestNameGenerator.ArgListFragment.GetText(MethodInfo method, Object[] arglist) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\TestNameGenerator.cs:line 558
   at NUnit.Framework.Internal.TestNameGenerator.GetDisplayName(TestMethod testMethod, Object[] args) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\TestNameGenerator.cs:line 93
   at NUnit.Framework.Internal.Builders.NUnitTestCaseBuilder.BuildTestMethod(IMethodInfo method, Test parentSuite, TestCaseParameters parms) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Builders\NUnitTestCaseBuilder.cs:line 89
   at NUnit.Framework.TestCaseSourceAttribute.<BuildFrom>d__22.MoveNext() in C:\src\nunit\nunit\src\NUnitFramework\framework\Attributes\TestCaseSourceAttribute.cs:line 142
   at NUnit.Framework.Internal.Builders.DefaultTestCaseBuilder.BuildFrom(IMethodInfo method, Test parentSuite) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Builders\DefaultTestCaseBuilder.cs:line 144
   at NUnit.Framework.Internal.Builders.NUnitTestFixtureBuilder.AddTestCasesToFixture(TestFixture fixture) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Builders\NUnitTestFixtureBuilder.cs:line 183
   at NUnit.Framework.Internal.Builders.NUnitTestFixtureBuilder.BuildFrom(ITypeInfo typeInfo, ITestFixtureData testFixtureData) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Builders\NUnitTestFixtureBuilder.cs:line 158
   at NUnit.Framework.TestFixtureAttribute.<BuildFrom>d__48.MoveNext() in C:\src\nunit\nunit\src\NUnitFramework\framework\Attributes\TestFixtureAttribute.cs:line 227
   at NUnit.Framework.Internal.Builders.DefaultSuiteBuilder.BuildFrom(ITypeInfo typeInfo) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Builders\DefaultSuiteBuilder.cs:line 80
```

I'm not sure if this modified example will actually have the issue or not, but my fixture looks similar to this:

Edit (simplifying example)
```
public class FastUnitTest : BaseCategoryAttribute { }

[TestFixture, FastUnitTest]
public class MyFixture
{

    // note my test has no `[SetUp]`, but seems to be referenced in the stacktrace above

        private static object[] testDataByteArray = new object[]
        {
            new object[]
            {
                "test",
                "another deal",
                new byte[] {1, 2, 3, 4}
            }
        };

        // this test passes with both 3.8.1 and 3.10.1
        [Test]
        [TestCaseSource(nameof(testDataByteArray))]
        public void TestThing(string thing, string anotherString, byte[] multiByte)
        {
            Assert.Pass();
        }

        private static object[] testDataMultiDimensionalByteArray = new object[]
        {
            new object[]
            {
                "test",
                "another deal",
                new byte[,] { { 1, 2 }, { 2, 3 } }
            }
        };

        // this test passes only on 3.8.1, it fails (to even set up the fixture) with the stacktrace above of 3.10.1
        [Test]
        [TestCaseSource(nameof(testDataMultiDimensionalByteArray))]
        public void TestThingMultiDimension(string thing, string anotherString, byte[,] multiByte)
        {
            Assert.Pass();
        }
}
```

Any thoughts?

mikkelbu:
I've augmented the existing tests even though we do not introspect multi-dimensional arrays, but this keeps the related tests together.

Fixes #2823

### DIFF
--- a/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
@@ -349,7 +349,7 @@ namespace NUnit.Framework.Attributes
                 Assert.That(suite.Tests[1].Name, Is.EqualTo(@"MethodWithArrayArguments([])"));
                 Assert.That(suite.Tests[2].Name, Is.EqualTo(@"MethodWithArrayArguments([1, Int32[], 4])"));
                 Assert.That(suite.Tests[3].Name, Is.EqualTo(@"MethodWithArrayArguments([1, 2, 3, 4, 5, ...])"));
-                Assert.That(suite.Tests[4].Name, Is.EqualTo(@"MethodWithArrayArguments(System.Byte[,])"));
+                Assert.That(suite.Tests[4].Name, Is.EqualTo(@"MethodWithArrayArguments([System.Byte[,]])"));
             });
         }
 
--- a/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
@@ -341,7 +341,7 @@ namespace NUnit.Framework.Attributes
             TestSuite suite = TestBuilder.MakeParameterizedMethodSuite(
                 typeof(TestCaseSourceAttributeFixture), nameof(TestCaseSourceAttributeFixture.MethodWithArrayArguments));
 
-            Assert.That(suite.TestCaseCount, Is.EqualTo(4));
+            Assert.That(suite.TestCaseCount, Is.EqualTo(5));
 
             Assert.Multiple(() =>
             {
@@ -349,6 +349,7 @@ namespace NUnit.Framework.Attributes
                 Assert.That(suite.Tests[1].Name, Is.EqualTo(@"MethodWithArrayArguments([])"));
                 Assert.That(suite.Tests[2].Name, Is.EqualTo(@"MethodWithArrayArguments([1, Int32[], 4])"));
                 Assert.That(suite.Tests[3].Name, Is.EqualTo(@"MethodWithArrayArguments([1, 2, 3, 4, 5, ...])"));
+                Assert.That(suite.Tests[4].Name, Is.EqualTo(@"MethodWithArrayArguments(System.Byte[,])"));
             });
         }
 
--- a/src/NUnitFramework/tests/Attributes/ValueSourceTests.cs
+++ b/src/NUnitFramework/tests/Attributes/ValueSourceTests.cs
@@ -233,7 +233,8 @@ namespace NUnit.Framework.Attributes
             new object[] { 1, "text", new object() },
             new object[0],
             new object[] { 1, new int[] { 2, 3 }, 4 },
-            new object[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 }
+            new object[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 },
+            new object[] { new byte[,] { { 1, 2 }, { 2, 3 } } }
         };
 
         [Test]
@@ -242,7 +243,7 @@ namespace NUnit.Framework.Attributes
             TestSuite suite = TestBuilder.MakeParameterizedMethodSuite(
                 GetType(), nameof(MethodWithArrayArguments));
 
-            Assert.That(suite.TestCaseCount, Is.EqualTo(4));
+            Assert.That(suite.TestCaseCount, Is.EqualTo(5));
 
             Assert.Multiple(() =>
             {
@@ -250,6 +251,7 @@ namespace NUnit.Framework.Attributes
                 Assert.That(suite.Tests[1].Name, Is.EqualTo(@"MethodWithArrayArguments([])"));
                 Assert.That(suite.Tests[2].Name, Is.EqualTo(@"MethodWithArrayArguments([1, Int32[], 4])"));
                 Assert.That(suite.Tests[3].Name, Is.EqualTo(@"MethodWithArrayArguments([1, 2, 3, 4, 5, ...])"));
+                Assert.That(suite.Tests[4].Name, Is.EqualTo(@"MethodWithArrayArguments([System.Byte[,]])"));
             });
         }
     }
--- a/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
+++ b/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
@@ -221,7 +221,7 @@ namespace NUnit.Framework.Internal
                     : Convert.ToString(arg, System.Globalization.CultureInfo.InvariantCulture);
 
                 var argArray = arg as Array;
-                if (argArray != null)
+                if (argArray != null && argArray.Rank == 1)
                 {
                     if (argArray.Length == 0)
                         display = "[]";
--- a/src/NUnitFramework/testdata/TestCaseSourceAttributeFixture.cs
+++ b/src/NUnitFramework/testdata/TestCaseSourceAttributeFixture.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using NUnit.Framework;
 
 namespace NUnit.TestData.TestCaseSourceAttributeFixture
@@ -147,7 +148,7 @@ namespace NUnit.TestData.TestCaseSourceAttributeFixture
         {
         }
 
-        [TestCaseSource(nameof(ComplexArrayBasedTestInput))]
+        [TestCaseSource(nameof(ComplexArrayBasedTestInputTestCases))]
         public void MethodWithArrayArguments(object o)
         {
         }
@@ -184,7 +185,7 @@ namespace NUnit.TestData.TestCaseSourceAttributeFixture
             }
         }
 
-        public static object[] ComplexArrayBasedTestInput = new[]
+        static object[] ComplexArrayBasedTestInput = new[]
         {
             new object[] { 1, "text", new object() },
             new object[0],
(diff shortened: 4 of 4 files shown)


src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -341,14 +341,15 @@ public void TestNameIntrospectsArrayValues()
...
-            Assert.That(suite.TestCaseCount, Is.EqualTo(4));
+            Assert.That(suite.TestCaseCount, Is.EqualTo(5));
 
             Assert.Multiple(() =>
             {
                 Assert.That(suite.Tests[0].Name, Is.EqualTo(@"MethodWithArrayArguments([1, ""text"", System.Object])"));
                 Assert.That(suite.Tests[1].Name, Is.EqualTo(@"MethodWithArrayArguments([])"));
                 Assert.That(suite.Tests[2].Name, Is.EqualTo(@"MethodWithArrayArguments([1, Int32[], 4])"));
                 Assert.That(suite.Tests[3].Name, Is.EqualTo(@"MethodWithArrayArguments([1, 2, 3, 4, 5, ...])"));
+                Assert.That(suite.Tests[4].Name, Is.EqualTo(@"MethodWithArrayArguments(System.Byte[,])"));
```
mikkelbu: Compare this result to the last result in ValueSourceTests.cs. Is this due to #1792?

src/NUnitFramework/tests/Attributes/ValueSourceTests.cs
```diff
@@ -242,14 +243,15 @@ public void TestNameIntrospectsArrayValues()
...
-            Assert.That(suite.TestCaseCount, Is.EqualTo(4));
+            Assert.That(suite.TestCaseCount, Is.EqualTo(5));
 
             Assert.Multiple(() =>
             {
                 Assert.That(suite.Tests[0].Name, Is.EqualTo(@"MethodWithArrayArguments([1, ""text"", System.Object])"));
                 Assert.That(suite.Tests[1].Name, Is.EqualTo(@"MethodWithArrayArguments([])"));
                 Assert.That(suite.Tests[2].Name, Is.EqualTo(@"MethodWithArrayArguments([1, Int32[], 4])"));
                 Assert.That(suite.Tests[3].Name, Is.EqualTo(@"MethodWithArrayArguments([1, 2, 3, 4, 5, ...])"));
+                Assert.That(suite.Tests[4].Name, Is.EqualTo(@"MethodWithArrayArguments([System.Byte[,]])"));
```
mikkelbu: Last result

CharliePoole: That looks wrong to me since the arg type seems to read as "a single-dimensioned array containing a System.Byte[,]" rather than as above where it's just a System.Byte[,].

jnm2: @CharliePoole Isn't it, though? `new object[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 }` vs `new object[] { new byte[,] { { 1, 2 }, { 2, 3 } } }`.

jnm2: @CharliePoole Isn't it a single-dimensioned array containing a System.Byte[,], though?
`new object[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 }` vs
`new object[] { new byte[,] { { 1, 2 }, { 2, 3 } } }`.

CharliePoole: Isn't the new object[] simply the array of arts rather. Than one of the arts? That's what I assumed from your comments. I can't see the code  on my phone.

src/NUnitFramework/tests/Attributes/ValueSourceTests.cs
```diff
@@ -233,7 +233,8 @@ public void MethodWithArrayArguments([ValueSource(nameof(ComplexArrayBasedTestIn
             new object[] { 1, "text", new object() },
             new object[0],
             new object[] { 1, new int[] { 2, 3 }, 4 },
-            new object[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 }
+            new object[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 },
+            new object[] { new byte[,] { { 1, 2 }, { 2, 3 } } }
```
jnm2: @mikkelbu Since NUnit is inconsistent in how it handles splatting, and for better reading comprehension, could you put them on the same level?
```cs
new object[] { new object[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 } },
new object[] { new byte[,] { { 1, 2 }, { 2, 3 } } }
```

jnm2: @CharliePoole Yes. This is probably going to tie into an existing PR and several other issues though, addressing consistency between TestCaseSource and ValueSource. Splatting means applying an array or tuple so that its elements become arguments to some method call.

CharliePoole: Yes, it does. 😄

At the level where we are working, there's always an explicit object array that holds the args, but not necessarily at the user level. Key issue is how we interpret a __user-specified__ array, particularly when passed to a method that accepts a single object arg. Is the outer object array just a holder for the arg or is it the arg itself?

That said, this is not the issue here. We are trying to figure out how to represent a __single__ argument.

In any case, I'd love to see an issue that highlights any inconsistency you have found, especially in the key case of an object array passed to a method taking a single object. Our intent, obviously, has always been to be consistent but since the code is necessarily in multiple places, it has a tendency to degrade.

jnm2: I was confused. TestCaseSource enumerates argument lists; ValueSource enumerates arguments.

What's weird is that TestCaseSource passes the *argument list* `new object[] { 1, 2, 3 }` as the *argument* `new object[] { 1, 2, 3 }` rather than giving the error that there are not three parameters in the test method.

If it gave an error, that would make it symmetric with ValueSource and consistent when wrapped with TestCaseData.

<details>

```cs
using NUnit.Framework;

public static class Repro
{
    public static readonly IEnumerable<object[]> ArrayPerArgumentList = new[]
    {
        new object[] { 1, 2, 3 },
        new object[] { new object[] { 1, 2, 3 } }
    };

    [TestCaseSource(nameof(ArrayPerArgumentList))]
    public static void TestA(object anythingGoes)
    {
        // WithTestCaseSource([1, 2, 3]) ✔ (❕)
        // WithTestCaseSource([1, 2, 3]) ✔
    }

    public static readonly IEnumerable<TestCaseData> TestCaseDataPerArgumentList = new[]
    {
        new TestCaseData(new object[] { 1, 2, 3 }),
        new TestCaseData(new object[] { new object[] { 1, 2, 3 } })
    };

    [TestCaseSource(nameof(TestCaseDataPerArgumentList))]
    public static void TestB(object anythingGoes)
    {
        // WithTestCaseSource(1,2,3)     ❌ Too many arguments provided, provide at most 1 arguments.
        // WithTestCaseSource([1, 2, 3]) ✔
    }
}
```

</details>

But I'm not sure whether this is a good or bad thing.

CharliePoole: So when you use TestCaseSource with an object[], NUnit has to figure out what you meant.

If you pass `new object[] { 1, 2, 3 }` to a method taking a single int, it should create two test cases. But if you pass the same thing to a method taking a single object or object[], then that's just one test case. If you want it to be three test cases, then you would have to use `TestCaseData` instead of `object[]`. `TestCaseData` always makes it unambiguous.

Probably, as a separate issue, we should review all the things we allow the user to pass in as data and how we handle it. We were probably too generous in what we originally accepted, although I'm not sure if we can get away from it now other than through documentation.

Ahha! I just opened up your hidden example, and I see I'm telling you what you already know!

jnm2:
@mikkelbu Would you be happy to have this PR merged as-is?

jnm2:
Sounds good. Up to you, but if you put [wip] in the title, the WIP bot integration will make sure someone doesn't accidentally think that you are ready for merge.

Issue #2826, opened 2018-04-26
Is.EquivalentTo is slow on large ordered collections when running on .NET Framework


ggeurts:
After updating from NUnit 3.9.0 to 3.10.1 the timeout limit on tests of a custom ordered set implementation is broken for larger sets (1000 items). The tests look like this one:

```C#
		[Test, TestCaseSource(nameof(TestCases)), Timeout(100)]
		public void CanAddItems(List<int> items)
		{
			var tree = CreateCollection<int>();
			foreach (var item in items)
			{
				tree.Add(item);
			}
			Assert.That(tree, Is.Ordered & Is.EquivalentTo(items));
		}
```

It turns out that the Is.EquivalentTo constraint caused delays of more than 100ms when running tests for .NET Framework, whereas they pass when running on .NET Core 2.0.

ggeurts:
Here is a full repro:
```C#
#if (!NETSTANDARD1_3 && !NETSTANDARD1_6)
        [Test, Timeout(100)]
        public void LimitPerformanceDegradationForLargeOrderedCollections()
        {
            var list = new List<int>();
            for (var i = 1000; i > 0; i--) list.Add(i);

            var collection = new List<int>(list.OrderBy(i => i));
            Assert.That(collection, Is.Ordered & Is.EquivalentTo(list));
        }
#endif

```

ggeurts:
This issue seems duplicate of  #2799.

jnm2:
You're right. Sorry you ran into this! I'll see if I can help that issue along.

Issue #2799, opened 2018-04-09
CollectionAssert.AreEquivalent is extremely slow


nvborisenko:
```csharp
var lines1 = new List<string> { };
var lines2 = new List<string> { };

for (int i = 0; i < 10000; i++)
{
  lines1.Add(i.ToString());
  lines2.Add(i.ToString());
}

CollectionAssert.AreEquivalent(lines1, lines2);
```

Verification of 10000 string elements takes 10 seconds. Verification of 100k elements may take more than 30 minutes (cannot wait until it finishes). This behavior takes a place in NUnit **3.10** and not reproducible in **3.9**

CharliePoole:
Several of us did changes to collection assert in that period. I'll take a look to see if one of mine caused this.

mikkelbu:
Actually, I think that my change in nunit/nunit#2600 has caused the problem (I've not had time to check it yet, as I'm halfway a VS upgrade). In the case above we will have to run through the entire remaining collection for every match we want to remove. Before my change we would find each match immediately as the first element.

jnm2:
The ideal situation might be to implement `IEqualityComparer` in all our internal comparer classes so they can be used with `HashSet` and `Dictionary`. `AreEquivalent` would use `HashSet.SetEquals` with [O(n + m)](https://msdn.microsoft.com/en-us/library/bb346516(v=vs.110).aspx#Anchor_2) (the counts of the two collections).

jnm2:
That sounds like a good idea. Either way, the collection equivalence constraint itself will probably need to use HashSet.SetEquals eventually to get to the standard O(n+m) people expect. It won't be able to do that unless all our internal NUnit equality comparers implement IEqualityComparer.

jnm2:
That's an interesting question. Is it even meaningful to test for set equality with tolerance? I don't know what set equality means without transitive element equality comparisons, and if you have tolerance, you don't have transitive equality. My attempt at searching didn't turn anything up. It would be cool to know.

We'd only need to adapt to IEqualityComparer when using a hash table. We'd only use a hash table for operations that compare two collections. A hash table in theory requires transitive equality comparisons and transitive hash codes. This is what the IEqualityComparer contract demands, so if it's inherently possible to use a hash table for the comparison, it's inherently possible to adapt our element comparer to IEqualityComparer.

[...]

PR #2828, opened 2018-04-26
Fix for #2799 and #2826: slow collection equivalence tests.

ggeurts:


jnm2:
@ggeurts Thanks for volunteering! We really appreciate it!

It looks like you opened a second PR by mistake. Just in case you didn't know, it's typical to push or force-push to the original PR branch to add or replace the commits while the PR is still open.
I'll close this one. See you over there: https://github.com/nunit/nunit/pull/2830

PR #2830, opened 2018-04-26
Fix for #2799 and #2826: slow collection equivalence tests.

### LINKED ISSUE #2799: CollectionAssert.AreEquivalent is extremely slow

ggeurts:
Fixes #2799

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionTally.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionTally.cs
@@ -30,6 +30,8 @@ namespace NUnit.Framework.Constraints
     /// of each object in one or more enumerations.</summary>
     public sealed class CollectionTally
     {
+        private static readonly object NullKey = new object();
+
         /// <summary>The result of a <see cref="CollectionTally"/>.</summary>
         public sealed class CollectionTallyResult
         {
@@ -47,8 +49,6 @@ namespace NUnit.Framework.Constraints
             }
         }
 
-        private readonly NUnitEqualityComparer comparer;
-
         /// <summary>The result of the comparison between the two collections.</summary>
         public CollectionTallyResult Result
         {
@@ -60,7 +60,7 @@ namespace NUnit.Framework.Constraints
             }
         }
 
-        private readonly List<object> _missingItems = new List<object>();
+        private readonly MissingItemCollection _missingItems;
 
         private readonly List<object> _extraItems = new List<object>();
 
@@ -69,31 +69,14 @@ namespace NUnit.Framework.Constraints
         /// <param name="c">The expected collection to compare against.</param>
         public CollectionTally(NUnitEqualityComparer comparer, IEnumerable c)
         {
-            this.comparer = comparer;
-
-            foreach (object o in c)
-                _missingItems.Add(o);
-        }
-
-        private bool ItemsEqual(object expected, object actual)
-        {
-            Tolerance tolerance = Tolerance.Default;
-            return comparer.AreEqual(expected, actual, ref tolerance);
+            _missingItems = new MissingItemCollection(comparer, c);
         }
 
         /// <summary>Try to remove an object from the tally.</summary>
         /// <param name="o">The object to remove.</param>
         public void TryRemove(object o)
         {
-            for (int index = _missingItems.Count - 1; index >= 0; index--)
-            {
-                if (ItemsEqual(_missingItems[index], o))
-                {
-                    _missingItems.RemoveAt(index);
-                    return;
-                }
-            }
-
+            if (_missingItems.Remove(o)) return;
             _extraItems.Add(o);
         }
 
@@ -103,6 +86,93 @@ namespace NUnit.Framework.Constraints
         {
             foreach (object o in c)
                 TryRemove(o);
+        }
+
+        private class MissingItemCollection : IEnumerable<object>
+        {
+            private readonly Dictionary<object, int> _items;
+            private readonly int _count;
+
+            public MissingItemCollection(NUnitEqualityComparer comparer, IEnumerable items)
+            {
+                _items = new Dictionary<object, int>(new TallyEqualityComparer(comparer));
+
+                foreach (object item in items)
+                {
+                    var key = item ?? NullKey;
+
+                    int occurrenceCount;
+                    if (_items.TryGetValue(key, out occurrenceCount))
+                    {
+                        _items[key] = occurrenceCount + 1;
+                    }
+                    else
+                    {
+                        _items.Add(key, 1);
+                    }
+
+                    _count++;
+                }
+            }
+
+            public bool Remove(object item)
+            {
+                var key = item ?? NullKey;
+
+                int occurrenceCount;
+                if (_items.TryGetValue(key, out occurrenceCount))
+                {
+                    if (occurrenceCount > 1)
+                    {
+                        _items[key] = occurrenceCount - 1;
+                    }
+                    else
+                    {
+                        _items.Remove(key);
+                    }
+                    return true;
+                }
+
+                return false;
+            }
+
+            public IEnumerator<object> GetEnumerator()
+            {
+                foreach (var pair in _items)
+                {
+                    var item = pair.Key == NullKey ? null : pair.Key;
+                    for (var i = 0; i < pair.Value; i++)
+                    {
+                        yield return item;
+                    }
+                }
+            }
+
+            IEnumerator IEnumerable.GetEnumerator()
+            {
+                return GetEnumerator();
+            }
+        }
+
+        private class TallyEqualityComparer : EqualityComparer<object>
+        {
+            private readonly NUnitEqualityComparer _comparer;
+            private Tolerance _tolerance = Tolerance.Default;
+
+            public TallyEqualityComparer(NUnitEqualityComparer comparer)
+            {
+                _comparer = comparer;
+            }
+
+            public override bool Equals(object x, object y)
+            {
+                return _comparer.AreEqual(x, y, ref _tolerance);
+            }
+
+            public override int GetHashCode(object obj)
+            {
+                return _comparer.GetHashCode(obj);
+            }
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/Comparers/ChainComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/ChainComparer.cs
@@ -1,0 +1,127 @@
+// ***********************************************************************
+// Copyright (c) 2018 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System.Collections;
+using System.Collections.Generic;
+
+namespace NUnit.Framework.Constraints.Comparers
+{
+    /// <summary>
+    /// An <see cref="IEqualityComparer"/> that also supports equality comparisons
+    /// with a <see cref="Tolerance"/>.
+    /// </summary>
+    internal abstract class ChainComparer : IEqualityComparer
+    {
+        /// <summary>
+        /// Indicates whether this comparer can compare a given object
+        /// to other objects.
+        /// </summary>
+        public virtual bool CanCompare(object obj)
+        {
+            return false;
+        }
+
+        /// <summary>Determines whether the specified objects are equal.</summary>
+        /// <param name="x">The first object to compare.</param>
+        /// <param name="y">The second object to compare.</param>
+        /// <returns>true if the specified objects are equal; otherwise, false.</returns>
+        public new bool Equals(object x, object y)
+        {
+            var tolerance = Tolerance.Default;
+            return Equals(x, y, ref tolerance) ?? false;
+        }
+
+        /// <summary>
+        /// Compares two objects for equality within a tolerance.
+        /// </summary>
+        /// <param name="x">The first object to compare.</param>
+        /// <param name="y">The second object to compare.</param>
+        /// <param name="tolerance">The comparison tolerance .</param>
+        /// <returns>true if the specified objects are equal within the given tolerance; otherwise, false.</returns>
+        public abstract bool? Equals(object x, object y, ref Tolerance tolerance);
+
+        /// <summary>Returns a hash code for the specified object.</summary>
+        public virtual int GetHashCode(object obj)
+        {
+            return obj == null ? 0 : 1;
+        }
+    }
+
+    /// <summary>
+    /// An <see cref="IEqualityComparer{T}"/> that also supports equality comparisons
+    /// with a <see cref="Tolerance"/>.
+    /// </summary>
+    internal abstract class ChainComparer<T> : ChainComparer, IEqualityComparer<T>
+    {
+        /// <summary>
+        /// Indicates whether this comparer can compare a given object
+        /// to other objects.
+        /// </summary>
+        public override bool CanCompare(object obj)
+        {
+            return obj is T;
+        }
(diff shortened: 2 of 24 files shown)


src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -106,5 +91,26 @@ public void TryRemove(IEnumerable c)
...
+            }
+
+            public override bool Equals(object x, object y)
+            {
+                return _comparer.AreEqual(x, y, ref _tolerance);
+            }
+
+            public override int GetHashCode(object obj)
+            {
+                return Default.GetHashCode(obj);
```
jnm2: @ggeurts There are two problems with this GetHashCode implementation.

The first is that it does not account for things like `.IgnoreCase` and `.Using(myCustomComparer)`. This means it breaks the IEqualityComparer contract: if `Equals(x, y)` is true, then GetHashCode(x) *must* equal `GetHashCode(y)`. When this contract is broken, hash tables like Dictionary which rely on this will give false negatives. This will cause bad results.

The second problem is deeper. We can't use `IEqualityComparer` at all if `NUnitEqualityComparer` has a nonzero tolerance. It'll give broken results because the hashcode for that is incorrect. We'll have to fall back to the old behavior if there is a tolerance. Were you following the discussion where we were talking about tolerances and transitivity?

jnm2: @ggeurts If writing correct, custom GetHashCode methods for every `NUnit.Framework.Constraints.Comparers` comparer was not what you had in mind when you started this PR, that's on me. I apologize for failing to lay out more concrete expectations. If you aren't interested, don't feel bad. If you are interested, I'm happy to help.

CharliePoole:
I'm assuming this is the active one of the two PRs on same subject. Right???

src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -47,22 +47,20 @@ public CollectionTallyResult()
             }
         }
 
-        private readonly NUnitEqualityComparer comparer;
-
-        /// <summary>The result of the comparison between the two collections.</summary>
+        /// <summary>The result of the comparision between the two collections.</summary>
```
CharliePoole: "Introduce misspelling" refactoring. 😀

src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -104,5 +87,92 @@ public void TryRemove(IEnumerable c)
             foreach (object o in c)
                 TryRemove(o);
         }
+
+        private class MissingItemCollection : IEnumerable<object>
```
jnm2: ❔ Why introduce this class rather than keeping _items and _count as fields of CollectionTally?

src/NUnitFramework/framework/Constraints/Comparers/ChainComparer.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// <param name="x">The first object to compare.</param>
+        /// <param name="y">The second object to compare.</param>
+        /// <param name="tolerance">The comparison tolerance .</param>
+        /// <returns>true if the specified objects are equal within the given tolerance; otherwise, false.</returns>
+        public abstract bool? Equals(object x, object y, ref Tolerance tolerance);
+
+        /// <summary>Returns a hash code for the specified object.</summary>
+        public virtual int GetHashCode(object obj)
+        {
+            return obj == null ? 0 : 1;
```
jnm2: I think this should be `return 0;` in case a null reference is considered equal to a certain object.

src/NUnitFramework/framework/Constraints/Comparers/ChainComparer.cs
```diff
@@ -0,0 +1,127 @@
...
+using System.Collections;
+using System.Collections.Generic;
+
+namespace NUnit.Framework.Constraints.Comparers
+{
+    /// <summary>
+    /// An <see cref="IEqualityComparer"/> that also supports equality comparisons
+    /// with a <see cref="Tolerance"/>.
+    /// </summary>
+    internal abstract class ChainComparer : IEqualityComparer
```
jnm2: Can you help me understand why we're implementing both non-generic and generic versions of `IEqualityComparer`?

src/NUnitFramework/framework/Constraints/Comparers/CharsComparer.cs
```diff
@@ -21,36 +21,30 @@
...
-            char c2 = caseInsensitive ? Char.ToLower(yChar) : yChar;
-
+            char c1 = _caseInsensitive ? char.ToLower(x) : x;
+            char c2 = _caseInsensitive ? char.ToLower(y) : y;
             return c1 == c2;
         }
+
+        public override int GetHashCode(char ch)
+        {
+            return _caseInsensitive ? char.ToLower(ch) : ch;
```
jnm2: IIRC, ToLower trips up with Turkish because they have a letter that can be uppercased twice and lowercased twice.

When I last checked, the only correct way to get a non-ordinal hashcode for a char was to allocate a single-char string and use `StringComparer.CurrentCultureIgnoreCase.GetHashCode`.

src/NUnitFramework/framework/Constraints/Comparers/HashCodeBuilder.cs
```diff
@@ -0,0 +1,182 @@
...
+
+using System.Collections;
+
+namespace NUnit.Framework.Constraints.Comparers
+{
+    /// <summary>
+    /// Helper class that supports the implementation of <see cref="object.GetHashCode"/>
+    /// method overrides.
+    /// </summary>
+    internal struct HashCodeBuilder
```
jnm2: I like this approach, but I agree with the design decisions in https://github.com/dotnet/corefx/issues/14354 and wonder if we could create an API more similar to https://docs.microsoft.com/en-gb/dotnet/api/system.hashcode?view=netcore-2.1.

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -172,31 +192,81 @@ public bool AreEqual(object x, object y, ref Tolerance tolerance, bool topLevelC
...
+        private IEnumerable<ChainComparer> SelectCollectionComparers()
+        {
+            return Comparers.Where(c => !(c is ArraysComparer) && !(c is EquatablesComparer));
+        }
+
+        private void SetProperty<T>(ref T field, T value)
+        {
+            if (_isFrozen)
+            {
+                throw new InvalidOperationException("Comparer is read-only after first comparison operation.");
```
jnm2: ❤

src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -69,31 +69,14 @@ public CollectionTallyResult Result
...
-            for (int index = _missingItems.Count - 1; index >= 0; index--)
-            {
-                if (ItemsEqual(_missingItems[index], o))
-                {
-                    _missingItems.RemoveAt(index);
-                    return;
-                }
-            }
-
+            if (_missingItems.Remove(o)) return;
```
jnm2: If we are doing a non-transitive equality comparison, such as when the tolerance is non-zero, we can't use _missingItems because _missingItems is based on a hash table (Dictionary).

jnm2: In fact, the `TryRemove(object)` paradigm is fundamentally broken if we are using a non-transitive comparison, so you're taking us from one kind of broken code to another kind of broken code. @nunit/framework-team Is this okay? Maybe we can deal with non-transitivity properly in a separate effort since it looks like it has always been broken.

ggeurts: The current collection equality compare operations do not support non-transitive comparisons. If transitive collection comparisons were required, interval trees might be used to improve performance.

src/NUnitFramework/framework/Constraints/CollectionTally.cs
ggeurts: In introduced this class to minimise changes to the rest of the CollectionTally code.

src/NUnitFramework/framework/Constraints/Comparers/ChainComparer.cs
ggeurts: The non-generic `ChainComparer` is the default base class for special-case NUnit comparisons. The generic `ChainComparer<T>` is used as base class for special-case comparisons that involve comparing two values of the same type (such as `DateTime` and `string`) and reduces boiler-plate code for these comparisons.

src/NUnitFramework/framework/Constraints/Comparers/CharsComparer.cs
ggeurts: Ok, the Turkish char comparison case should be added to the unit tests in that case. The new  implementation maintains behavior of the current one.

src/NUnitFramework/framework/Constraints/CollectionTally.cs
jnm2: @rprouse, @CharliePoole: I feel like we need more input than just me here. Is it okay for EquivalentTo with a tolerance to go from one kind of broken to the same kind of broken but with different side effects?

src/NUnitFramework/framework/Constraints/Comparers/CharsComparer.cs
jnm2: Oh, sorry! I missed that we were already doing that. :+1:

src/NUnitFramework/framework/Constraints/CollectionTally.cs
jnm2: It feels like we're introducing a second collection tally class though. I'm not comfortable with it.

src/NUnitFramework/framework/Constraints/CollectionTally.cs
CharliePoole: @jnm2 Since you were going in a different direction, I had bowed out of this one. I can take another look but it will be easier for me (maybe others too) if you summarize how you see the existing collection tally as "broken" and what new firm of "broken" you see as coming in to replace it. Could be the limitations of following this on my cell, but I can't see what you are getting at.

If my memory is correct, we don't support tolerance in equivalence operations right now. Are you trying to add that support while working on performance improvement? If so, I'd say let's pick one goal at a time to work on even if we have to visit the same code twice.

jnm2: @CharliePoole I wasn't aware of the tolerance issue until https://github.com/nunit/nunit/issues/2799#issuecomment-379943712, and we discussed what that would look like in the comments immediately following that.

I don't believe I said that CollectionTally is broken; "EquivalentTo with a tolerance" is what I said was broken. Our equivalence implementation takes optimization shortcuts which are invalid when a tolerance is specified. We don't report an error if a user tries to do it. We silently return the wrong answer instead. Here is an example of something that fails which should succeed:

```cs
Assert.That(
    new[] { 2f, 1f },
    Is.EquivalentTo(new[] { 3f, 2f })
        .Using<float>((a, b) => Math.Abs(a - b) <= 1) );
```

Are you open to the idea making it an error to use collection equivalence with a non-transitive comparison? I like that as I said above because it allows us to implement that separately in the future, if at all.
If you think there might be folks doing `Is.EquivalentTo(...).Using(...)` today though, is it okay to take that semi-functional behavior away from them? Should we provide a thorough fix?

CharliePoole: @jnm2 You mentioned me and asked for input. I haven't been following this, so I'm just trying to clarify what you see as broken.  CollectionTally was mentioned previously so I guessed that was it.

We decided not to support tolerance in assertions that compare collections as sets because it's a pretty difficult problem of optimization with the potential of not converging to a solution. It's possible to reverse the decision but I wouldn't call that broken.

Another past decision was to allow folks to specify a tolerance in cases where we will not actually use it. It's similar to how you can specify case insensitive and then do a numeric test. That decision could also be changed or we could simply stop allowing it on certain constraints.q

I thought that this issue was aimed at improving efficiency without changing behavior. I think that's a better approach so since you asked I'm telling you, even though it's you who will have to move forward with the consequences.

Totally on the meta side but related... I was just in a very interesting session with Kent Beck around the question of what happens as diffs become smaller and smaller. At some point. He thin

jnm2: @CharliePoole This is why EquivalentTo, in combination with some of its `Using` methods, is broken:

- `Is.EquivalentTo(...).Using<double>((a, b) => Math.Abs(a - b) <= 1)` doesn't cause an error.
  I'd expect an InvalidOperationException if we decided not to support collection asserts with tolerances.

- `Is.EquivalentTo(...).Using<double>((a, b) => Math.Abs(a - b) <= 1)` doesn't ignore the custom comparer, either. It returns answers that are different from `Is.EquivalentTo(...)`, answers which are not only different from `Is.EquivalentTo(...)` but also which are incorrect for `Is.EquivalentTo(...).Using(...)`.

It's not documented to the user as such, but this constitutes undefined behavior. Our optimizations make it very order-sensitive. It will usually appear to the user that `Is.EquivalentTo(...).Using(...)` is working as intended while the user is setting up the test.

Wouldn't the right thing be to throw InvalidOperationException?
If we don't, this PR will have the side effect of changing the behavior of `Is.EquivalentTo(...).Using(...)` from something which appears to work as intended but doesn't, to something which still appears to work but fails to work as intended in a different way.

CharliePoole: Yes I said that allowing using is a potential bug. It's a bug because the user can simulate use of a tolerance as you do. I suspect that it will take a bit of discussion to select the best way to fix it.

As a bug, I feel it needs an issue, decision and fix. If I were doing it, I'd fix it first, followed by the performance fix. It doesn't seem like it should take long once you decide. It's a choice between disallowing it at compile time or throwing an exception later.

If that's your sole concern, I'm with you. It sounded like you were faulting us for not implementing equivalence with tolerance, which I think was correct.

jnm2: @CharliePoole I'm sorry I came across as faulting you for anything, particularly a decision I agree with. I referred to tolerance when I meant the more inclusive term 'non-transitivity,' so I probably confused both of us!

> I thought that this issue was aimed at improving efficiency without changing behavior. I think that's a better approach.

You mentioned that you thought this issue was aimed at improving efficiency without changing behavior. (Specifically, the behavior of `Is.EquivalentTo(...).Using(...)` will be scrambled around with a hash table.)
My point is that this issue *can't* improve efficiency without changing behavior, unless this PR waits for a separate PR to throw InvalidOperationException. This is my sole concern.

The meta comment is interesting to me. If you didn't use branches, would you still keep the master branch continuously deliverable?

jnm2: @CharliePoole I'm sorry I came across as faulting you for anything, and ironically a decision I agree with. I referred to tolerance when I meant the more inclusive term 'non-transitivity,' so I probably confused both of us!

You mentioned that you thought this issue was aimed at improving efficiency without changing behavior. 
This is my sole concern: this issue *can't* improve efficiency without changing behavior, unless this PR waits for a separate PR to throw InvalidOperationException. (Specifically, the behavior of `Is.EquivalentTo(...).Using(...)` will be scrambled around with a hash table.)

The meta comment is interesting to me. If you didn't use branches, would you still keep the master branch continuously deliverable?

jnm2: Here's the discussion issue: https://github.com/nunit/nunit/issues/2870

CharliePoole: @jnm2 More of a problem of clarity for me and it's clear now. nowv

CharliePoole: Regarding the meta comment, I don't think any of us fully understood but I think so. If you had been there and asked he'd have likely said "Go figure that out" 'cause that's who he is.

His idea includes some special tools that don't exist yet as well as putting an AST in the repo rather than text, so it's kind of futuristic.

My idea is a simpler subset... what if diffs were so small that you could automatically commit them once the tests passed?

jnm2: If you automatically commit when tests pass, are there still reviews and discussions?

jnm2: I might decide NUnit should have a brand new feature or remove an API. I'd still seek out the team's opinions and then want to link the implementation back to the discussion, right?

CharliePoole: I'd sure think so.

jnm2: If reviews were eliminated, how would the team handle mistakes if my commit didn't quite match the discussion? Master would then not be deliverable.

CharliePoole: Of course any change can be reversed.

However. You aren't getting in the "what if" spirit. The question was how small would a change have to be to eliminate all review. Or what would have to be true about the change?

What about a small refactoring? A large one? A single new failing test plus code to make it pass? A bunch of new tests that improve coverage and pass without changes?

Finally, if we could do a lot of work more simply and quickly, how would it affect the way we release?

What if is powerful. Somebody was the first to sdk What if each dev could have his own private repo for example.

jnm2: I'll keep thinking about it.

rprouse:

PR #2833, opened 2018-04-28
 Use longer BusyExecIdle to avoid CI failures 

### LINKED ISSUE #2831: Regular "BusyExecIdle after 200 milliseconds delay" CI failures
This is responsible for the last couple dozen times I read through the AppVeyor logs to figure out why a branch was failing to build. Always seems to be net35.

```
1) Failed : NUnit.Framework.Internal.Execution.TestWorkerTests.BusyExecuteIdleEventsCalledInSequence
  Expected: "BusyExecIdle" after 200 milliseconds delay
  But was:  <string.Empty>
   at NUnit.Framework.Internal.Execution.TestWorkerTests.BusyExecuteIdleEventsCalledInSequence() in C:\projects\nunit\src\NUnitFramework\tests\Internal\TestWorkerTests.cs:line 67
```

Can we change this to `10_000`? If it succeeds it's instantaneous. If it fails, all we care about is not hanging for minutes on end.

jnm2:
Fixes #2831

### DIFF
--- a/src/NUnitFramework/tests/Internal/TestWorkerTests.cs
+++ b/src/NUnitFramework/tests/Internal/TestWorkerTests.cs
@@ -64,7 +64,8 @@ namespace NUnit.Framework.Internal.Execution
             _worker.Start();
             _queue.Start();
 
-            Assert.That(() => sb.ToString(), Is.EqualTo("BusyExecIdle").After(200));
+            Assert.That(() => sb.ToString(), Is.EqualTo("BusyExecIdle").After(
+                delayInMilliseconds: 10000, pollingInterval: 200));
         }
 
         private void FakeMethod()


src/NUnitFramework/tests/Internal/TestWorkerTests.cs
```diff
@@ -56,15 +56,20 @@ public void BusyExecuteIdleEventsCalledInSequence()
...
 
-            _queue.Enqueue(work);
-            _worker.Start();
-            _queue.Start();
+                _queue.Enqueue(work);
+                _worker.Start();
+                _queue.Start();
 
-            Assert.That(() => sb.ToString(), Is.EqualTo("BusyExecIdle").After(200));
+                if (!countdown.Wait(10000)) Assert.Fail("Timed out.");
```
jnm2: If I had just changed `After`, we would have an unnecessary 10 second wait on success. (I also avoid [Timeout] on tests when possible because thread aborts introduce inherent unreliability.)

ChrisMaddock: Would this work using the polling functionality of DelayedConstraint? (E.g. After(10000,200))

jnm2: I wasn't aware of that. Probably! Should  I try it?

jnm2: Seems to work great, thanks!

PR #2834, opened 2018-04-30
#2798 - Show actual count on Has.Exactly(x).Items

### LINKED ISSUE #2798: [Request] Show actual count value when test fail on Has.Exactly(x).Items
When I test that a enumeration have the correct count of members, so it do it with

```csharp
    /// <summary>
    /// Represents a PCAN filter mode
    /// </summary>
    public enum PeakCanMode : byte
    {
        /// <summary>
        /// Mode is Standard (11-bit identifier)
        /// </summary>
        PcanModeStandard = 0x_00,

        /// <summary>
        /// Mode is Extended (29-bit identifier)
        /// </summary>
        PcanModeExtended = 0x_02
    }
```

```csharp
      ///<summary>
      /// Test that the enumeration have the correct count of elements
      ///</summary>
      [Test]
      [Description("Test that the enumeration have the correct count of members")]
      public void HaveTheCorrectMemberCount()
          => Assert.That(Enum.GetNames(typeof(PeakCanMode)), Has.Exactly(3).Items);
```

When the count is wrong i get
> Meldung:   Expected: exactly 3 items
>   But was:  < "PcanModeStandard", "PcanModeExtended" >

but it would be nicer, when I would get
> Meldung:   Expected: exactly 3 items
>   But was:  actual 2 items < "PcanModeStandard", "PcanModeExtended" >

Yes I know, I can add a extra message as error description with the count, but i think it would be nicer
```csharp
      ///<summary>
      /// Test that the enumeration have the correct count of elements
      ///</summary>
      [Test]
      [Description("Test that the enumeration have the correct count of members")]
      public void HaveTheCorrectMemberCount()
          => Assert.That(Enum.GetNames(typeof(PeakCanMode)),
                         Has.Exactly(3).Items);
                         $"Actual {Enum.GetNames(typeof(EsaDeviceCommand)).Length} items");

```

TobiasSekan:
Fixes #2798

### DIFF
--- a/src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.ObjectModel;
 using NUnit.Framework.Internal;
 
 namespace NUnit.Framework.Constraints
@@ -64,26 +65,33 @@ namespace NUnit.Framework.Constraints
         /// Apply the item constraint to each item in the collection,
         /// succeeding only if the expected number of items pass.
         /// </summary>
-        /// <param name="actual"></param>
-        /// <returns></returns>
+        /// <param name="actual">The value to be tested</param>
+        /// <returns>A ConstraintResult</returns>
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
             var enumerable = ConstraintUtils.RequireActual<IEnumerable>(actual, nameof(actual));
+            var itemList = new Collection<object>();
+            var matchCount = 0;
 
-            int count = 0;
-            if (_itemConstraint == null)
+            foreach (var item in enumerable)
             {
-                foreach (object item in enumerable)
-                    count++;
-            }
-            else
-            {
-                foreach (object item in enumerable)
+                if (_itemConstraint != null)
+                {
                     if (_itemConstraint.ApplyTo(item).IsSuccess)
-                        count++;
+                        matchCount++;
+                }
+                else
+                {
+                    matchCount++;
+                }
+
+                // We intentionally add one item too many because we use it to trigger
+                // the ellipsis when we call "MsgUtils.FormatCollection" later on.
+                if (itemList.Count <= ExactCountConstraintResult.MaxDisplayCount)
+                    itemList.Add(item);
             }
 
-            return new ConstraintResult(this, actual, count == _expectedCount);
+            return new ExactCountConstraintResult(this, actual, matchCount == _expectedCount, matchCount, itemList);
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
@@ -1,0 +1,80 @@
+// ***********************************************************************
+// Copyright (c) 2018 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System.Collections.Generic;
+
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// Contain the result of matching a <see cref="ExactCountConstraint"/> against an actual value.
+    /// </summary>
+    internal sealed class ExactCountConstraintResult : ConstraintResult
+    {
+        /// <summary>
+        /// The maximum count of list elements that are shown on the constraint result
+        /// </summary>
+        internal const int MaxDisplayCount = 10;
+
+        /// <summary>
+        /// The count of matched items of the <see cref="ExactCountConstraint"/>
+        /// </summary>
+        private readonly int _matchCount;
+
+        /// <summary>
+        /// A list with maximum count (+1) of items of the <see cref="ExactCountConstraint"/>
+        /// (The maximum count is set in <see cref="MaxDisplayCount"/>)
+        /// </summary>
+        private readonly ICollection<object> _itemList;
+
+        /// <summary>
+        /// Constructs a <see cref="ConstraintResult"/> for a <see cref="ExactCountConstraint"/>.
+        /// </summary>
+        /// <param name="constraint">The Constraint to which this result applies.</param>
+        /// <param name="actualValue">The actual value to which the Constraint was applied.</param>
+        /// <param name="isSuccess">If true, applies a status of Success to the result, otherwise Failure.</param>
+        /// <param name="matchCount">Count of matched items of the <see cref="ExactCountConstraint"/></param>
+        /// <param name="itemList">A list with maximum count (+1) of items of the <see cref="ExactCountConstraint"/></param>
+        internal ExactCountConstraintResult(IConstraint constraint, object actualValue, bool isSuccess, int matchCount, ICollection<object> itemList)
+            : base(constraint, actualValue, isSuccess)
+        {
+            _matchCount = matchCount;
+            _itemList = itemList;
+        }
+
+        /// <summary>
+        /// Write the actual value for a failing constraint test to a MessageWriter.
+        /// </summary>
+        /// <param name="writer">The writer on which the actual value is displayed</param>
+        public override void WriteActualValueTo(MessageWriter writer)
+        {
+            if (_itemList == null || _itemList.Count == 0)
+            {
+                writer.Write("no items");
+                return;
+            }
+
+            writer.Write(_matchCount != 1 ? "{0} items " : "{0} item ", _matchCount);
+            writer.Write(MsgUtils.FormatCollection(_itemList, 0, MaxDisplayCount));
+        }
+    }
+}
--- a/src/NUnitFramework/tests/Constraints/ExactCountConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/ExactCountConstraintTests.cs
@@ -42,7 +42,7 @@ namespace NUnit.Framework.Constraints
         {
             var expectedMessage =
                 TextMessageWriter.Pfx_Expected + "no item equal to \"Charlie\"" + Environment.NewLine +
-                TextMessageWriter.Pfx_Actual + "< \"Charlie\", \"Fred\", \"Joe\", \"Charlie\" >" + Environment.NewLine;
+                TextMessageWriter.Pfx_Actual + "2 items < \"Charlie\", \"Fred\", \"Joe\", \"Charlie\" >" + Environment.NewLine;
             var ex = Assert.Throws<AssertionException>(() => Assert.That(names, new ExactCountConstraint(0, Is.EqualTo("Charlie"))));
             Assert.That(ex.Message, Is.EqualTo(expectedMessage));
         }
@@ -59,7 +59,7 @@ namespace NUnit.Framework.Constraints
         {
             var expectedMessage =
                 TextMessageWriter.Pfx_Expected + "exactly one item equal to \"Charlie\"" + Environment.NewLine +
-                TextMessageWriter.Pfx_Actual + "< \"Charlie\", \"Fred\", \"Joe\", \"Charlie\" >" + Environment.NewLine;
+                TextMessageWriter.Pfx_Actual + "2 items < \"Charlie\", \"Fred\", \"Joe\", \"Charlie\" >" + Environment.NewLine;
             var ex = Assert.Throws<AssertionException>(() => Assert.That(names, new ExactCountConstraint(1, Is.EqualTo("Charlie"))));
             Assert.That(ex.Message, Is.EqualTo(expectedMessage));
         }
@@ -96,7 +96,7 @@ namespace NUnit.Framework.Constraints
         {
             var expectedMessage =
                 TextMessageWriter.Pfx_Expected + "exactly 2 items equal to \"Fred\"" + Environment.NewLine +
-                TextMessageWriter.Pfx_Actual + "< \"Charlie\", \"Fred\", \"Joe\", \"Charlie\" >" + Environment.NewLine;
+                TextMessageWriter.Pfx_Actual + "1 item < \"Charlie\", \"Fred\", \"Joe\", \"Charlie\" >" + Environment.NewLine;
             var ex = Assert.Throws<AssertionException>(() => Assert.That(names, new ExactCountConstraint(2, Is.EqualTo("Fred"))));
             Assert.That(ex.Message, Is.EqualTo(expectedMessage));
         }
@@ -118,7 +118,7 @@ namespace NUnit.Framework.Constraints
         {
             var expectedMessage =
                 TextMessageWriter.Pfx_Expected + "exactly one item" + Environment.NewLine +
-                TextMessageWriter.Pfx_Actual + "< \"Charlie\", \"Fred\", \"Joe\", \"Charlie\" >" + Environment.NewLine;
+                TextMessageWriter.Pfx_Actual + "4 items < \"Charlie\", \"Fred\", \"Joe\", \"Charlie\" >" + Environment.NewLine;
             var ex = Assert.Throws<AssertionException>(() => Assert.That(names, new ExactCountConstraint(1)));
             Assert.That(ex.Message, Is.EqualTo(expectedMessage));
         }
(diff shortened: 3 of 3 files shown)


mikkelbu:
Hi @TobiasSekan. I've just edited your description such that this PR auto-closes #2798 when it is merged to master.

src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
```diff
@@ -296,6 +294,18 @@ private void WriteExpectedLine(object expected, Tolerance tolerance)
         private void WriteActualLine(ConstraintResult result)
         {
             Write(Pfx_Actual);
+
+            // Write the actual item count, when ActualValue is a collection
+            if (result.Name == "ExactCount" && result.ActualValue is IEnumerable)
```
jnm2: So that every possible constraint doesn't end up putting code in this shared location, we use custom constraint results for this type of thing. In other words, you'd create a private sealed class `ExactCountConstraint.ExactCountConstraintResult`, override `WriteActualValueTo`, and use it [here](https://github.com/nunit/nunit/blob/d411123f1c705cc7ad147c67d2420d97b0706626/src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs#L86). Have you tried that?

CharliePoole: @TobiasSekan Emphasizing @jnm2 's point...

`TextMessageWriter` knows how to put out a message in text form based on what the constraint asked to have done. Constraints know what the `IMessageWriter` interface can do and call the appropriate methods. The message writer doesn't know anything about constraints and shouldn't. It only varies what is put out based on the particular type of output. So, a hypothetical `HtmlMessageWriter` would put put different characters for the same call from what a `TextMessageWriter` does.

A constraint, like `ExactCountConstraint` decides what it's output will look like. If you examine constraint code in NUnit, you will see that many different kinds of output are produced. Some constraints do this by direct use of the message writer, while others do it by creating a custom constraint result, as suggested by @jnm2. A custom result is needed when the output is significantly different from the standard or particularly complex.

I suggest you look at the code for various constraints, especially those that output lists of collection elements to see how we do it and modify the `ExactCountConstraint` output, leaving `TextMessageWriter` as it is.

TobiasSekan:
Hello,

I haven't notice this, I have search for a possibility to override the result, but I have only found the `Display` setter.

I currently not happy with the for-loop over the array object, but I have not idea for a better solution to count the element inside a none generic `IEnumerable`

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
```diff
@@ -102,6 +101,8 @@ public override string Description
                 return _itemConstraint != null ? PrefixConstraint.FormatDescription(descriptionPrefix, _itemConstraint) : descriptionPrefix;
             }
         }
+
+
```
jnm2: Would you please remove these blank lines? (A habit I've gotten into is to view the diff before I commit so that I don't accidentally add things like this.)

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// </summary>
+        /// <param name="writer">The writer on which the actual value is displayed</param>
+        public override void WriteActualValueTo(MessageWriter writer)
+        {
+            var elementList = _actualValue as IEnumerable;
+            if (elementList == null)
+            {
+                writer.Write("no items");
+                return;
+            }
```
jnm2: One benefit of making this class private nested is that you would then have an absolute guarantee that `_actualValue` is an instance of `IEnumerable`. If you keep the class internal, I'd ask that you put `ConstraintUtils.RequireActual<IEnumerable>(actualValue, nameof(actualValue));` in the *constructor*.

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    /// <summary>
+    /// Contain the result of matching a <see cref="ExactCountConstraint"/> against an actual value.
+    /// </summary>
+    public sealed class ExactCountConstraintResult : ConstraintResult
+    {
+        /// <summary>
+        /// The actual value to which the Constraint was applied.
+        /// </summary>
+        private readonly object _actualValue;
```
jnm2: Redundant since the base constructor sets the `ActualValue` property to the same value.

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
```diff
@@ -0,0 +1,68 @@
...
+                writer.Write("no items");
+                return;
+            }
+
+            var writeCounter  = 0;
+            var stringBuilder = new StringBuilder();
+
+            stringBuilder.AppendFormat(itemCounter > 1 ? "{0} items <" : "{0} item <", itemCounter);
+
+            foreach (var element in (IEnumerable)_actualValue)
```
jnm2: Maybe use `elementList` instead of a second cast?

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
```diff
@@ -0,0 +1,68 @@
...
+            }
+
+            if (itemCounter == 0)
+            {
+                writer.Write("no items");
+                return;
+            }
+
+            var writeCounter  = 0;
+            var stringBuilder = new StringBuilder();
```
jnm2: It's more efficient and simpler code to read if you write directly to the writer (which may already be wrapping a StringBuilder) than to start a buffer here.

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            stringBuilder.AppendFormat(itemCounter > 1 ? "{0} items <" : "{0} item <", itemCounter);
+
+            foreach (var element in (IEnumerable)_actualValue)
+            {
+                writeCounter++;
+                stringBuilder.AppendFormat(writeCounter != itemCounter ? " \"{0}\"," : " \"{0}\" >", element);
```
jnm2: It's more efficient and simpler code to read if you write directly to the writer (which is already doing its own buffering) than to introduce the string builder. This should be `writer.Write(format, element);`.

jnm2:
> I currently not happy with the for-loop over the array object, but I have not idea for a better solution to count the element inside a none generic IEnumerable

If you're not happy, you *could* add `using System.Linq;` and use `enumerable.Cast<object>().Count()`. That has the optimization of using `ICollection<>.Count` if available, too, so it's both simpler and faster.
However, the loop as you have it is also fine with me.

src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
```diff
@@ -21,9 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-using System;
 using System.Collections;
-using System.Globalization;
```
jnm2: Please amend this change too so that your PR doesn't show up in the history of this file.

src/NUnitFramework/framework/Constraints/ConstraintResult.cs
```diff
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
```
jnm2: Please revert and amend so that your PR doesn't show up in the history of this file.

jnm2:
@CharliePoole Good catch, thanks!

jnm2:
I'd say not! What's a good alternative?

CharliePoole:
Here's a straw man to start with...

If the test being made (nested constraint) is equality just say 
    Expected n1 items equal to whatever but was  n2 items.

For any other test (greater than, etc.) say the same thing but then follow with the actual values, possibly on multiple lines, but don't list more than some hard-coded number.

src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
TobiasSekan: Oh :roll_eyes: , i forgot that, my used plug-in, **Power Commands for VS** automatic remove unused **using** on save

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
```diff
@@ -0,0 +1,68 @@
+using System.Collections;
+using System.Text;
+
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// Contain the result of matching a <see cref="ExactCountConstraint"/> against an actual value.
+    /// </summary>
+    public sealed class ExactCountConstraintResult : ConstraintResult
```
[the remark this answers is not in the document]
TobiasSekan: :+1: -> internal
**private** is impossible for all objects on-top level of a namespace

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
TobiasSekan: :+1:

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
TobiasSekan: :+1:

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
TobiasSekan: :+1:

TobiasSekan:
@CharliePoole I can't use the original counter, because he only count the matched objects

```csharp
        {
            var enumerable = ConstraintUtils.RequireActual<IEnumerable>(actual, nameof(actual));

            int count = 0;
            if (_itemConstraint == null)
            {
                foreach (object item in enumerable)
                    count++;
            }
            else
            {
                foreach (object item in enumerable)
                    if (_itemConstraint.ApplyTo(item).IsSuccess)
                        count++;
            }

            return new ExactCountConstraintResult(this, actual, count == _expectedCount);
        }
```

CharliePoole:
There are two paths, with and without an item constraint. In each path the actual count is known, even though it has a different meaning in each.

CharliePoole:
Don't revert changes in a PR, instead use local reset or interactive rebate followed by a forced push. While we don't do forced pushed to master, we're OK with them on feature branches.

TobiasSekan:
1. Is there a way to roll back to a commit of the PR on Github or should I close this PR and make a new PR? 
2. I will looking for using the counter until the next week

CharliePoole:
Removing commits is very easy. Assuming you use command-line git, look at the help for the reset command. Examples:

Removing the last commit entirely, losing the changes...

    git reset --hard HEAD~1

Losing the last three commits...

    git reset --hard HEAD~3

Rolling back the last commit, but keeping the changes so that you can continue to revise

   git reset HEAD~1

Using reset to remove the __second-to-last commit__ only, keeping the last one...

    git log // write down the comment on the commit you want to keep
    git reset HEAD~1
    git stash // Save changes you want to re-commit
    git --hard reset HEAD~1 // Kill previous next-to-last, current last commit
    git stash pop // Get the changes back
    git commit -m "Comment you wrote down"

That last one can be done equally well (maybe even easier) using git rebase but IMO just using reset initially is a good way to get a grasp of how git commits work. Also, I think rebase takes a bit more effort to master.

Any of these would, of course, be followed by `git push -f`.

Any of these could be combined with changes to the code before re-commiting.

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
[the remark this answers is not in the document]
jnm2: That's why I suggested private *nested*. 😉 I had in mind `ExactCountConstraint.ExactCountConstraintResult`. Not a problem though.

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
jnm2: This fix looks like it might have been missed with the rest you did. I'm fan of rebasing/amending/squashing within a PR, so feel free to make your commits all nice and shiny if you feel like it.

jnm2:
Just saw Charlie's very useful comments. I'm totally on board with what he's saying.

I'd add to that: `git commit --amend` combines your staged changes into the most recent commit rather than adding a new commit.
I find [interactive rebase](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history#interactive-rebase) the easiest of all for everything else. You can change the order, squash things together, drop commits altogether, edit any commits, insert new commits, etc. For interactive rebase, you probably don't want the default editor which is vim. I like Notepad++ because of how fast it opens. Setup:

```
git config --global core.editor "'C:/Program Files/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"
```

Recent installs of Git for Windows also offer to use Visual Studio Code as the editor. It's an Electron app and a bit slow to open for my taste, but still a solid option. Setup:

```
git config --global core.editor "code --new-window --wait"
```

jnm2:
Just saw Charlie's very useful comments. I'm totally on board with what he's saying.

I'd add to that: `git commit --amend` combines your staged changes into the most recent commit rather than adding a new commit.
Besides a standalone `reset` command, I find [interactive rebase](https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history#interactive-rebase) the easiest of all for everything else. You can change the order, squash things together, drop commits altogether, edit any commits, insert new commits, etc. For interactive rebase, you probably don't want the default editor which is vim. I like Notepad++ because of how fast it opens. Setup:

```
git config --global core.editor "'C:/Program Files/Notepad++/notepad++.exe' -multiInst -notabbar -nosession -noPlugin"
```

Recent installs of Git for Windows also offer to use Visual Studio Code as the editor. It's an Electron app and a bit slow to open for my taste, but still a solid option. Setup:

```
git config --global core.editor "code --new-window --wait"
```

And, we appreciate the effort you're putting into this PR! Great stuff!

TobiasSekan:
@jnm2 ping

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            if (_elementList == null)
+            {
+                writer.Write("no items");
+                return;
+            }
+
+            writer.Write(_matchCount != 1 ? "{0} items <" : "{0} item <", _matchCount);
+
+            var writeCounter  = 1;
```
jnm2: I find the aligned `=` a little odd. The only problem I see is that when future contributors edit nearby, their editor will cause them grief by reformatting to remove the extra spaces and they'll have chose between painstakingly putting them back, or including changes over unrelated lines in the git blame history.

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            foreach (var element in _elementList)
+                writer.Write(writeCounter++ < _itemCount ? " \"{0}\"," : " \"{0}\" >", element);
```
jnm2: I apologize that this is only occurring to me just now; we try to respect the IEnumerable contract whose premise is that if you call GetEnumerator a second time, the sequence may have completely changed from the first enumeration. We try not to enumerate twice for the sake of correctness.

This is only for display purposes though, so I'm on the fence. @oznetmaster? @nunit/framework-team?

jnm2: It will drive logs on build servers though and may be the key to rare, hard-to-debug issue. I'm feeling less good about the double enumeration the more I think about real-world scenarios I've been in.

jnm2: @nunit/framework-team Final thought: maybe we should merge as-is and open an issue for adding cross-cutting double-enumeration-detecting tests. If we care about this, we should have tests that fail if we do double enumeration.

CharliePoole: As per my original comment, a constraint result should not re-enumerate the values.  It's supposed to __be__ a result, not __calculate__ one. The result is just an aggregation of data that can display itself in a meaningful way. The constraint initialized the result with the info needed, which I had understood to be the expected and actual counts.

CharliePoole: If we decided to display all the matching items , I missed that. It seems excessive but if we want to do it it's up to the constraint to save them.

jnm2: @CharliePoole I would like to see the elements printed out no matter what kind of constraint the element constraint is. But as you said, there might be thousands and we should certainly limit that.

@TobiasSekan Let's buffer the items into a list at the same time that we cound them, but stop adding to the list after there are already maybe ten items. Then pass that list to [`MsgUtils.FormatCollection`](https://github.com/nunit/nunit/blob/5c6413314bbbbbde29e2cdd7f851c4108554a2de/src/NUnitFramework/framework/Constraints/MsgUtils.cs#L145).

TobiasSekan:
Should i add a additional information that only 10 items are shown, when there are more that 10 items?

CharliePoole:
In other messages, we show that a list or a string is incomplete by use of `...`.

src/NUnitFramework/tests/Constraints/ExactCountConstraintTests.cs
```diff
@@ -136,5 +136,50 @@ public void FailsWhenNotUsedAgainstAnEnumerable()
             TestDelegate act = () => Assert.That(notEnumerable, new ExactCountConstraint(1));
             Assert.That(act, Throws.ArgumentException.With.Message.Contains("IEnumerable"));
         }
+
+        [Test]
+        [Description("Test that the" + nameof(ExactCountConstraint) + "return all ten elements of the given list without a triple dot")]
```
jnm2: This forms the string `Test that theExactCountConstraintreturn all…`. `return` should be `returns`, as in, "it returns."

CharliePoole: Also note missing space.

TobiasSekan:
Ping - need review :-)

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
```diff
@@ -0,0 +1,88 @@
...
+{
+    /// <summary>
+    /// Contain the result of matching a <see cref="ExactCountConstraint"/> against an actual value.
+    /// </summary>
+    internal sealed class ExactCountConstraintResult : ConstraintResult
+    {
+        /// <summary>
+        /// The maximum count of list elements that are shown on a constrain result, when the constrain condition not match
+        /// </summary>
+        internal static int MaxDisplayCount = 10;
```
jnm2: `const`?

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
```diff
@@ -0,0 +1,88 @@
...
+
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// Contain the result of matching a <see cref="ExactCountConstraint"/> against an actual value.
+    /// </summary>
+    internal sealed class ExactCountConstraintResult : ConstraintResult
+    {
+        /// <summary>
+        /// The maximum count of list elements that are shown on a constrain result, when the constrain condition not match
```
jnm2: XML docs are optional on internal things, but here's some nitpicks if you'd like to add them:
- `constrain result` should be `constraint result`, probably?
- Isn't this the maximum count *always*, not the maximum count only when the constraint condition doesn't match?

src/NUnitFramework/framework/Constraints/ExactCountConstraintResult.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// </summary>
+        internal static int MaxDisplayCount = 10;
+
+        /// <summary>
+        /// The list with all items of the <see cref="ExactCountConstraint"/>
+        /// </summary>
+        private readonly IEnumerable _elementList;
```
jnm2: No longer used, can be removed.

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
```diff
@@ -69,21 +70,32 @@ public ExactCountConstraint(int expectedCount, IConstraint itemConstraint)
...
+                }
             }
             else
             {
-                foreach (object item in enumerable)
+                foreach (var item in enumerable)
+                {
                     if (_itemConstraint.ApplyTo(item).IsSuccess)
-                        count++;
+                        matchCount++;
```
jnm2: The loop is getting big enough that I'd suggest moving the `_itemConstraint == null` check into the loop so that there is only one loop.

src/NUnitFramework/framework/Constraints/ExactCountConstraint.cs
```diff
@@ -69,21 +70,32 @@ public ExactCountConstraint(int expectedCount, IConstraint itemConstraint)
...
+
+                    if (itemList.Count <= ExactCountConstraintResult.MaxDisplayCount)
+                        itemList.Add(item);
```
jnm2: I was confused for a while, so it might be good to add a comment saying that we intentionally add one item too many to trigger the ellipsis when we later on call `FormatCollection`.

TobiasSekan:
ping @jnm2

rprouse:
Thanks for your contribution.

PR #2839, opened 2018-05-01
Test of Extending syntax helpers

### LINKED ISSUE #2836: NUnit.Framework.Does cannot be extended
The documentation for custom constraints demonstrates incorporating custom constraints into the fluent syntax like this:

> Provide a static class patterned after NUnit's Is class, with properties or methods that construct your custom constructor. If you like, you can even call it Is and extend NUnit's Is, provided you place it in your own namespace and avoid any conflicts.

That works fine for NUnit.Framework.Is, but NUnit.Framework.Does is marked as static, and so cannot be extended in the same way.

Given that it serves basically the same role and is only really there to make the assertion read more nicely, I think it should also be extendable, for the exact same reasons that Is is extendable.

CharliePoole:
Fixes #2836

### DIFF
--- a/src/NUnitFramework/framework/Does.cs
+++ b/src/NUnitFramework/framework/Does.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2014 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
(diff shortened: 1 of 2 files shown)


rprouse:
VSTS isn't building automatically or reporting, I am going to look into that, but for now I made it no longer required. Travis OSX is failing for what looks to be infrastructure problems, so I am going to override this time.

PR #2847, opened 2018-05-09
Support continued inheritance of Assert and related classes

### LINKED ISSUE #2842: Supporting inheritance of Assert and related classes
In https://github.com/nunit/nunit/issues/2836, the subject came up of allowing `Assert` and other `*Assert` classes to be extended through inheritance the way we do for `Is`, `Iz`, `Does`, `Has`, `Throws`, and `Contains`.

A past team, before NUnit 3, had decided that `Assert` should be inheritable:

https://github.com/nunit/nunit/blob/d411123f1c705cc7ad147c67d2420d97b0706626/src/NUnitFramework/framework/Assert.cs#L53-L62

@CharliePoole asked for us to reconsider these classes, keeping in mind that, “The current team has never actually **decided** that we want people to be able to do this.”

➡ What do I mean by “supporting inheritance of Assert,” since it is currently inheritable? From my best understanding of https://github.com/nunit/nunit/issues/2836, I mean keeping it in the list of classes we test to ensure inheritance which we started in https://github.com/nunit/nunit/pull/2839.

jnm2:
Fixes #2842.

### DIFF
--- a/src/NUnitFramework/framework/Assert.Comparisons.cs
+++ b/src/NUnitFramework/framework/Assert.Comparisons.cs
@@ -25,7 +25,7 @@ using System;
 
 namespace NUnit.Framework
 {
-    public partial class Assert
+    public abstract partial class Assert
     {
         #region Greater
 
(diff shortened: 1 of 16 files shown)


jnm2:
No problem. I'm always fighting tooling to keep it DRY, anyway. Good thought!

jnm2:
Updated the partial class definitions to be identical.

Issue #2786, opened 2018-03-28
Timeout value not resetting on Retry of failed test


db798:
Not sure if this has been raised before, could find a similar issue. 

I have a test with both Retry and Timeout attribute but when a test is retried the timeout isn't being reset to 0.

```
using NUnit.Framework;
using System;
using System.Threading;

namespace RetryTimeoutTest
{
    public class Class1
    {
        private static int count1 = 0;
        private static int count2 = 0;

        [Test]
        [Retry(3)]
        public void Test1()
        {
            count1++;
            Console.WriteLine(count1);

            Thread.Sleep(2000);

            if (count1 < 2)
            {
                Console.WriteLine("In failing assert");
                Assert.True(false);
            }
            else
            {
                Console.WriteLine("In passing assert");
                Assert.True(true);
            }
        }

        [Test]
        [Retry(3)]
        [Timeout(3000)]
        public void Test2()
        {
            count2++;
            Console.WriteLine(count2);

            Thread.Sleep(2000);

            if (count2 < 2)
            {
                Console.WriteLine("In failing assert");
                Assert.True(false);
            }
            else
            {
                Console.WriteLine("In passing assert");
                Assert.True(true);
            }
        }
    }
}
```

**Expected Result:** Both tests to pass after a retry

**Actual Result** Test1 passes as expected but Test2 fails with message _Test exceeded Timeout value of 3000ms_. 

It looks like the timeout isn't being reset when the test starts the second time, is this the expected result or should a reset be performed for each retry of the tests. 

_**ps** sorry about the code sample, doesn't look to be formatting everything_

_@ChrisMaddock edit: fixed formatting_

ChrisMaddock:
This sounds to me like an edge case we haven't thought of. I agree I'd expect it to work as you say.

Will just wait for someone else on @nunit/framework-team to confirm!

jnm2:
Yes, I can't think of any good way to use timeouts with retries the way it currently.

We'll have to make sure the timeout command wraps the test command before any attributes get a chance to wrap it. Would that change any behavior in combination with other attributes?

CharliePoole:
I think you can consider this confirmed. Rationale: Timeout works at the WorkItem level. Retry works at the command level. Since the work item executes the command chain, the single timeout value applies to the aggregate of the repetitions. 

Changing this would probably mean re-implementing timeout at the command level, since attributes don't generally get to play at the WorkItem level.

DaiMichael:
[...]

PR #2849, opened 2018-05-09
Timeout value not resetting on Retry of failed test #2786

DaiMichael:
Fixed `Timeout` & `Retry` issue, applied to `Repeat` using the suggested `IRepeatTest` interface solution. Now warns if both `Retry` and `Repeat` on one test.

Added new tests and all current tests past.


src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
```diff
@@ -153,6 +154,14 @@ private TestCommand MakeTestCommand()
                     command = new TimeoutCommand(command, timeout);
 #endif
 
+                // warn if we have more than one IRepeatTest (Retry / Repeat) Attribute
+                if (method.GetAttributes<IRepeatTest>(true).Length > 1)
+                    return new WarnCommand(_testMethod, "Test case has both RetryAttribute and RepeatAttribute, which could give unpredictable results.");
```
CharliePoole: This puts a small bit more information into workitem than we would like it to have. Can we just say that multiple attributes that repeat a test may cause problems? Remember, a user attribute could implement the interface.

DaiMichael:
@CharliePoole - Can do, you are lucky not to have seen v1... I got it to `string.Join` the attrributes for each on on there! and decided that was silly, even if correct. 🤦‍♂️

src/NUnitFramework/framework/Internal/Commands/WarnCommand.cs
```diff
@@ -0,0 +1,59 @@
...
+
+using System;
+using NUnit.Framework.Interfaces;
+
+namespace NUnit.Framework.Internal.Commands
+{
+    /// <summary>
+    /// This innerCommand will add a warning to the test result
+    /// </summary>
+    public class WarnCommand : TestCommand
```
CharliePoole: Clever! I see pros and cons to this approach, which we should discuss.

src/NUnitFramework/framework/Internal/Commands/WarnCommand.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <summary>
+        /// Overridden to simply set the CurrentResult to the
+        /// appropriate warn state with message.
+        /// </summary>
+        /// <param name="context">The execution context for the test</param>
+        /// <returns>A TestResult</returns>
+        public override TestResult Execute(TestExecutionContext context)
+        {
+            TestResult testResult = context.CurrentResult;
+            testResult.SetResult(ResultState.Warning, _message);
```
CharliePoole: So if the test had failed, we would change it to a warning and if more the failure?

src/NUnitFramework/framework/Internal/Commands/WarnCommand.cs
DaiMichael: 😨 ... This was an interesting decision... I saw the `SkipCommand` but there was nothing else that did what I needed. I looked at other ways to add a warning but couldn't see any other examples in the code to template. I didn't want to pollute code with a check (if `(method.GetAttributes<IRepeatTest>(true).Length > 1)`) like this anywhere else and it seemed to nicely where it was at the end of the command chain.

It didn't break any current tests so I couldn't see (with my limited knowledge) any downside but I was wondering if I would be asked to explain myself 😉

DaiMichael: 😨 ... This was an interesting decision... I saw the `SkipCommand` but there was nothing else that did what I needed. I looked at other ways to add a warning but couldn't see any other examples in the code to template. I didn't want to pollute code with a check ('if (method.GetAttributes<IRepeatTest>(true).Length > 1)`) like this anywhere else and it seemed to nicely where it was at the end of the command chain.

It didn't break any current tests so I couldn't see (with my limited knowledge) any downside but I was wondering if I would be asked to explain myself 😉

src/NUnitFramework/framework/Internal/Commands/WarnCommand.cs
DaiMichael: hmm.... The test could have failed due to the warning of more than one repeatable attributes.. 

FYI - This is a devils advocate answer.

We could check current state of test and if fail leave as failed.

CharliePoole:
We have never had warnings like this before as far as I remember. So we keep a record, I'll comment further on issue #2786 itself rather than here. I think the issue is a bit easier to find for anyone wanting to look at this in the future.

CharliePoole:
Looks good to me!

DaiMichael:
Hmm... CI keeps timing out? I can't see a reason either, do I need to do anything?

CharliePoole:
Someone will restart it soon. I can't do it from my phone.

rprouse:
Build restarted. It was the unrelated .NET 3.5 timeout issue.

src/NUnitFramework/tests/Attributes/RepeatableTestsWithTimeoutAttributesTests.cs
```diff
@@ -0,0 +1,110 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal.Commands;
+using NUnit.TestUtilities;
+
+#if (!NETCOREAPP1_1 || !NETCOREAPP2_0) && THREAD_ABORT
```
jnm2: `(!NETCOREAPP1_1 || !NETCOREAPP2_0)` is always true. It's logically identical to `!(NETCOREAPP1_1 && NETCOREAPP2_0)` and there's no instance where a build has both `NETCOREAPP1_1` and `NETCOREAPP2_0`.

src/NUnitFramework/tests/Attributes/RepeatableTestsWithTimeoutAttributesTests.cs
```diff
@@ -0,0 +1,110 @@
...
+    [TestFixture]
+    public class RepeatableTestsWithTimeoutAttributesTests
+    {
+        private int retryOnlyCount;
+        private int retryTimeoutCount;
+        private int repeatTimeoutCount;
+        
+        [Test]
+        [Retry(3)]
+        public void ShouldPassAfter3Retires()
```
jnm2: Typo `Retires` (2 places)

src/NUnitFramework/tests/Attributes/RepeatableTestsWithTimeoutAttributesTests.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            retryOnlyCount++;
+            System.Threading.Thread.Sleep(20);
```
jnm2: Why the sleep since there's no timeout?

src/NUnitFramework/tests/Attributes/RepeatableTestsWithTimeoutAttributesTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            repeatTimeoutCount++;
+
+            System.Threading.Thread.Sleep(35);
+            Assert.True(repeatTimeoutCount >= 1);
+        }
+
+        [Test]
+        public void ShouldFailOnMultipleRepeatableAttributes()
+        {
+            var testCase = TestBuilder.MakeTestCase(GetType(), "TestMethodForRepeatAndRetryExpectedFail");
```
jnm2: We use C# 6, so please use `nameof` for refactor safety. (2 places)

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
```diff
@@ -153,6 +154,10 @@ private TestCommand MakeTestCommand()
                     command = new TimeoutCommand(command, timeout);
 #endif
 
+                // Add wrappers for repetable tests after timeout so the timeout is reset on each instance
+                foreach (var repeatableAttribute in method.GetAttributes<IRepeatTest>(true))
+                    command = repeatableAttribute.Wrap(command);
```
jnm2: What happens when an attribute implements both `IWrapSetUpTearDown` and `IRepeatTest`? I think it'll get applied both here and in the earlier loop. We should add a test on this behavior.

src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
```diff
@@ -102,6 +104,20 @@ public TestMethod BuildTestMethod(FixtureMethod method, Test parentSuite, TestCa
...
 
+        /// <summary>
+        /// Checks to see if we have valid combinations of attributes.
+        /// </summary>
+        /// <param name="testMethod">The TestMethod to be checked. If it
+        /// is found to be non-runnable, it will be modified.</param>
+        /// <returns>True if the method signature is valid, false if not</returns>
+        private static bool CheckTestMethodAttributes(TestMethod testMethod)
+        {
+            if (testMethod.Method.GetAttributes<IRepeatTest>(true).Length > 1)
```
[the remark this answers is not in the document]
CharliePoole: First... I'd always advise test-writers not to write that! 😸 

If they did, it should run three times. But I think @jnm2 is asking whether the base class repeat will be honored if there is no repeat on the override. I'd say no.

Of course, in this PR, we should not change any behavior except what it aims to change - i.e. fix the issue. Any other change of behavior we make here is a "slipstream" upgrade at best, a breaking change without notice at worst.

We could probably use a set of tests that deal with overrides, just to document the current behavior and trigger a review if they start to fail.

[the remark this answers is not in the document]
jnm2: I'm actually asking if `DerivedTest.SomeTest` should repeat at all in:

```cs
    [TestFixture]
    public class BaseTest
    {
        [Test]
        [Repeat(2)]
        public virtual void SomeTest()
        {
            ...
        }
    }

    [TestFixture]    
    public class DerivedTest : BaseTest
    {
        [Test]
        public override void SomeTest()
        {
            ...
        }
    }
```

But I missed the fact that [we already repeat it](https://github.com/nunit/nunit/pull/2849/files#diff-1372200a9e9129d2d04701246dbff8b3R137) for our current RepeatAttribute and of course I'm not interested in expanding this PR to change existing behavior, like Charlie confirms.

src/NUnitFramework/framework/Interfaces/ICommandWrapper.cs
```diff
@@ -63,4 +63,13 @@ public interface IWrapTestMethod : ICommandWrapper
     public interface IWrapSetUpTearDown : ICommandWrapper
     {
     }
+
+    /// <summary>
+    /// Objects implementing this interface are used to wrap
+    /// tests that can repeat. The implemeting object runs
+    /// each time before the test is repeats.
```
jnm2: If I understood properly, an object implementing `IRepeatTest` is *not* invoked each time that the test is repeated. It's invoked once and may decide to invoke the inner command any number of times. Am I thinking correctly?

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
```diff
@@ -153,6 +154,10 @@ private TestCommand MakeTestCommand()
                     command = new TimeoutCommand(command, timeout);
 #endif
 
+                // Add wrappers for repeatable tests after timeout so the timeout is reset on each instance
```
jnm2: I apologize for not catching earlier... the term `instance` isn't as clear as saying `on each repeat`.

src/NUnitFramework/tests/Attributes/RepeatableTestsWithTimeoutAttributesTests.cs
```diff
@@ -0,0 +1,170 @@
...
+        [Retry(3)]
+        public void ShouldPassAfter3Retries()
+        {
+            retryOnlyCount++;
+            Assert.True(retryOnlyCount >= 2);
+        }
+
+        [Test]
+        [Retry(3), Timeout(30)]
+        public void ShouldPassAfter3RetiresAndTimeoutIsResetEachTime()
```
jnm2: Fyi, you only got one of the two instances I mentioned of the typo `Retires` (here's the other) 😊

src/NUnitFramework/tests/Attributes/RepeatableTestsWithTimeoutAttributesTests.cs
```diff
@@ -90,21 +90,81 @@ public void ShouldFailOnMultipleRepeatableAttributesIncludingCustom()
...
 
+        [Test]
+        public void ShouldFailWithRepeatableAndSetupTearDownCustomAttributes()
+        {
+            CustomRepeaterAndSetupAndTearDown.AttributeCallCount = 0;
+            var testCase = TestBuilder.MakeTestCase(GetType(), nameof(TestMethodForRepeatableAndSetupTearDownCustomAttribute));
+            var workItem = TestBuilder.CreateWorkItem(testCase);
+            TestBuilder.ExecuteWorkItem(workItem);
+
+            Assert.That(CustomRepeaterAndSetupAndTearDown.AttributeCallCount, Is.EqualTo(2));
```
jnm2: The title of the test is "ShouldFail" but I'm not seeing where a failure is tested for. Does it make any sense for the attribute to wrap the command at two different levels? I don't think so, because it can't know which of the two reasons `Wrap` is being called.

src/NUnitFramework/tests/Attributes/RepeatableTestsWithTimeoutAttributesTests.cs
```diff
@@ -90,21 +90,81 @@ public void ShouldFailOnMultipleRepeatableAttributesIncludingCustom()
...
+    [TestFixture]
+    public class DerivedRepeatableTestFixtureWithNoRepeat : BaseRepeatableTestFixture
+    {
+        [Test]
+        public override void ShouldBeOveridden()
+        {
+            RepeatCount++;
+            Assert.True(RepeatCount == 1);
+        }
+    }
```
jnm2: These will make some awesome tests! Are they used in a test yet?

src/NUnitFramework/framework/Interfaces/ICommandWrapper.cs
CharliePoole: Answering from memory... the wrapper object (IRepeatTest) is supposed to set up a TestCommand that is added to the chain of commands used in running the test. The command runs once, invoking chained commands as many times as it likes.

src/NUnitFramework/tests/Attributes/RepeatableTestsWithTimeoutAttributesTests.cs
DaiMichael: I only wrote that test last night and I can't understand what I was thinking a case was... I have removed it.

src/NUnitFramework/framework/Interfaces/ICommandWrapper.cs
```diff
@@ -66,8 +66,8 @@ public interface IWrapSetUpTearDown : ICommandWrapper
 
     /// <summary>
     /// Objects implementing this interface are used to wrap
-    /// tests that can repeat. The implemeting object runs
-    /// each time before the test is repeats.
+    /// tests that can repeat. The implemeting command is run once,
+    /// invoking the chained commands as many times as defined.
```
jnm2: Sooo close! I have to point out the typo `implemeting` here. Also, "as defined" sounds to me a little more vague than saying "invoking the chained commands any number of times."

src/NUnitFramework/tests/Attributes/RepeatableTestsWithTimeoutAttributesTests.cs
jnm2: I think this is actually a valuable test and we should keep testing it. I was wondering though what the test outcome should be. I think we need a new error mode where we mark the test invalid if the same class implements `IRepeatTest` and `IWrapSetUpTearDown`. The test would be the same as the above test, but you would assert that the test case is marked as not runnable.

DaiMichael:
@jnm2 Regarding the following

> I think this is actually a valuable test and we should keep testing it. I was wondering though what the test outcome should be. I think we need a new error mode where we mark the test invalid if the same class implements `IRepeatTest` and `IWrapSetUpTearDown`. The test would be the same as the above test, but you would assert that the test case is marked as not runnable.

I got a bit confused why I did this, I could have sworn I'd seen a comment from @CharliePoole about doing this test, then couldn't find the comment anywhere... Hence I removed it as for the following.

If I keep the test as is, this means I "should" for completeness add additional tests to include `IWrapTestMethod` as well. There could be a combination of `IWrapTestMethod`, `IWrapSetUpTearDown` and `IRepeatTest`. If all three are implemented, then I would expect a execute `count == 3` ala current test.

I'm not sure what the outcome should be? I don't see a problem with doing this, even if it is an unusual "pattern", although I could see unexpected behavior, i.e. repeat a setup method.

So happy to put test back in and do a couple more for completeness but I would have to see what current outcome is and put that as expected, so as not to change behavior. Do you agree?

DaiMichael:
@jnm2 - I also don't understand why build keep failing, this PR has consistently failed with no errors but timeouts. Obviously all builds fine locally for me.

jnm2:
@DaiMichael Good point about `IWrapTestMethod`. Even if current behavior is to wrap twice when the attribute implements `IWrapTestMethod` and `IWrapSetUpTearDown` (which I think is a problem and should change), I don't want to be consistent with that problem when we add `IRepeatTest`. I'd much rather mark the test invalid with the message that the same builder cannot implement `IRepeatTest` if it implements `IWrapSetUpTearDown` or `IWrapTestMethod`.

The reason (again) is that there's no way to know if you're being called as a test repeater or as a setup/teardown wrapper or as an after-setup, pre-teardown wrapper.

CharliePoole:
If this is a real problem, then it has been a problem before the addition of IRepeatTest. So...

1. Should we be adding to the problem? Can't we just make IRepeatTest a separate interface?

2. Can we fix the original problem or would you consider that an unacceptable breaking change?

Obviously, in designing the interface, I didn't consider it a problem. I thought having one interface under several names was a sufficient signal to an attribute implementor that they should pick just one. But if this now appears to you as a bad choice, why not just change it rather than introducing a complication?

jnm2:
@CharliePoole 

> Can't we just make IRepeatTest a separate interface?

Sure, that would solve the problem by allowing the same builder to implement the two separate interface methods explicitly. However, let's say we did make them separate interfaces. Even then, if both `IRepeatTest.Wrap` and `IWrapSetUpTearDown.Wrap` are implemented implicitly, you'll _again_ have no way to know what the function of the `Wrap` method should be for a given call.
So in any case, I'd still advocate a check to enforce that either you implement no more than one of these interfaces, or that `IRepeatTest.Wrap` is not mapped to the same method as `IWrapSetUpTearDown.Wrap`.

So for that reason, making `IWrapSetUpTearDown` no longer inherit from `ICommandWrapper` would not solve the problem. It would just be a breaking change.

CharliePoole:
To be clear, I'm not advocating for a separate interface. My primary argument is that this is __not really a problem__. I'll go into that further below.

However, given our different views of how custom extensions should be treated, you may decide to protect against the problem you perceive. If so, I'd say you should make the interface separate. True, if you implement two interfaces, each having a method with the same name, you have to implement them explicitly. But any dev who doesn't know that shouldn't even be __using__ NUnit, much less extending it!!!.

It might be a useful exercise to try to imagine practical use cases for an attribute that implemented each possible combination of these interfaces. If there is no good reason to do that, then our historical approach to the problem is to tell people "Don't do that." 😄 

On the general issue of protecting ourselves against bad custom attributes... The past attitude was that such extensions are, for all practical purposes, a part of NUnit itself. There are uncounted ways in which a custom attribute can completely break NUnit because they have access to internals. (That's why we used an Internal namespace at the beginning, rather than `internal` visibility.) Not that I want to spend the time, but I imagine I could take a few hours and come up with about 50 ways a custom attribute can break us. So the question arises: "Is this case special in some way or are you going to try to catch all those 50 potential problems as well?"

To me, the simpler approach is to say: "Writing custom attributes is a highly specialized activity requiring detailed internal knowledge of NUnit. Anyone who does so is taking a risk. Obviously they should have tests and should not deploy a custom attribute if it breaks things. We trust them to do that, just as we trust ourselves and our own tests WRT the internals of NUnit itself."

I can see a bit of logic for making this interface a special case, without trying to change our entire outlook on customization. You may feel that having separate interfaces invites implementing them separately. My own feeling was that having separate interfaces __derived from a common interface without adding anything__ clearly communicates that you should only pick one of them. The only way to resolve that is to ask multiple people.

OTOH, if you are __really__ trying to change the entire view that NUnit takes of custom attributes - that is by trying to protect NUnit against errors other than thrown exceptions - then I don't think it should be slid into a small issue like this. It would be a major redirection and it's implementation would take serious review of all our interfaces and analysis of the risks. If I were staying around, I'd argue against spending our time that way. 😟

CharliePoole:
@jnm2 I think I"ve said what I need to say on this issue as a departing active member. I'll leave it to you to decide where you're going with it.

jnm2:
I'm in favor of guardrails and failing early and descriptively to the extent that it is possible. There's also a nice benefit that people won't find a way to rely on undocumented behavior to achieve a goal but will instead open an issue for discussion about a better way to support their scenario.

The lack of checks appears to me as a burden representing future problems. I'm quite willing to learn and adapt, but this is where I'm starting from. It's great when you explain things like this so that I can see from your point of view. I'd like to have more conversation on why we can get away with saving time by not thinking through such corner cases. But I hear you; it's not going to be a priority.

Since I don't desire to waste time or change NUnit's entire view on this, let's go with having no test, @DaiMichael.

CharliePoole:
When it comes to customizations like this, my attitude has been "would we check against a similar error in a different part of the framework itself?" That's not the only attitude possible, of course. 

But since it's no longer just a matter of my personal attitude, it seems to me that discussion and possibly documentation are called for on the part of the active members. While discussions between you and me are always interesting, I believe it's in the interest of the project in the long term that the discussion be more generalized. I'm afraid that when the two of us get going, that tends not to happen. 😅

src/NUnitFramework/framework/Interfaces/ICommandWrapper.cs
```diff
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2011-2015 Charlie Poole, Rob Prouse
```
jnm2: Optional: 2011–2018

DaiMichael:
@jnm2 

Might as well finish it off 😆 Are we meant to update the copyright date when we touch a file?

Not sure what to pickup next? anything you want looking at?

jnm2:
> Are we meant to update the copyright date when we touch a file?

For major changes, I believe so. See https://github.com/nunit/docs/wiki/Coding-Standards#notes, n3.

> Not sure what to pickup next? anything you want looking at?

You're awesome! Have you looked at https://github.com/nunit/nunit/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22 yet? There are more issues we need to get to the `help wanted` stage, too.

mikkelbu:
@jnm2 Looking at this it seems that the label is wrong, and that the PR is only awaiting a second review?

jnm2:
@mikkelbu Oh dear... thanks.

jnm2:
...and oops.

mikkelbu:
One of the new tests failed
```
Test20
...
Errors, Failures and Warnings
1) Failed : NUnit.Framework.Attributes.RepeatableTestsWithTimeoutAttributesTests.ShouldPassAfter2RepeatsAndTimeoutIsResetEachTime
Test exceeded Timeout value of 50ms
```

jnm2:
We need much larger timeout values. Tests should be written to take no time at all if they pass and timeouts simply save CI from running for an hour before cancelling.

DaiMichael:
@jnm2 @mikkelbu - Wotcha, not been about for a month or two.

This one is tricky as I need to make the test timeout for the reset to occur. I've adjusted the test to have a bit more "room" without a massive slow down on CI. 

Usually in these cases where I need to have a Sleep / etc... As long as there aren't lots of tests I don't worry too much. If we need lots of the then it is more of a worry or we might need to create a "Long Running" set of tests.

mikkelbu:
I had to restart Travis two times due to other errors.
```
1) Failed : NUnit.Framework.Api.FrameworkControllerTests.ExploreTestsAction_FilterCategory_ReturnsTests
  Expected: not null and String starting with "ID"
  But was:  "7680"
   at NUnit.Framework.Api.FrameworkControllerTests.ExploreTestsAction_FilterCategory_ReturnsTests() in /home/travis/build/nunit/nunit/src/NUnitFramework/tests/Api/FrameworkControllerTests.cs:line 316
```
and
```
1) Failed : NUnit.Framework.Api.FrameworkControllerTests.RunTestsAction_AfterLoad_ReturnsRunnableSuite("<filter/>")
  Expected: not null and String starting with "ID"
  But was:  "8508"
  at NUnit.Framework.Api.FrameworkControllerTests.RunTestsAction_AfterLoad_ReturnsRunnableSuite (System.String filter) [0x00000] in <9278af9c5d3e455d8858b76816878cf6>:0 
```

DaiMichael:
@rprouse - I'd like to close this PR please, I have created a new PR #3007 for the same issue with the same code. 

As I explain in the new PR, I started this a long time ago when I was new to github and I had a cross contaminated branch with another, now accepted PR. I also need to do a rebase posy Joesph's changes and it was a bit of a mess. It was easier to create a new branch and merge the handful of class changes in by hand. Should all be good in the new PR #3007 and merge easily.

rprouse:
@DaiMichael no problem, you can always close yourself with a comment, but I've done so for you. Thanks for your continued work....

PR #2856, opened 2018-05-12
Show non-matching item for AllItems/NoItem constraints

### LINKED ISSUE #2854: Has.All.../Has.None... - show non-matching items in error message
Currently NUnit only logs actual collection in error message. That's fine if we have 3 items, but not really great for 200, especially due to that only several first items are displayed.
Would be really great to log non-matching items as well, something like:
```
Expected: all items instance of <System.Char>
Actual: < 'a', \"b\", 'c' >
Non-Matching: < \"b\" > 
```

That would require to traverse through all collection though, even if the first checked item does not match, but I suppose that's OK, considering that it might happen only for failing test.

Dreamescaper:
Fixes #2854 

Added additional line in message showing non-matching item:
```
Expected: all items instance of <System.Char>
Actual: < 'a', \"b\", 'c' >
Non-matching item at index [1]: "b"      <---- ADDED
```

### DIFF
--- a/src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
@@ -1,0 +1,40 @@
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// Provides a <see cref="ConstraintResult"/> for the constraints 
+    /// that are applied to each item in the collection
+    /// </summary>
+    public class EachItemConstraintResult : ConstraintResult
+    {
+        private readonly object _nonMatchingItem;
+        private readonly int _nonMatchingItemIndex;
+
+        /// <summary>
+        /// Constructs a <see cref="EachItemConstraintResult" /> for a particular <see cref="Constraint" />.
+        /// Only used for Failure
+        /// </summary>
+        /// <param name="constraint">The Constraint to which this result applies.</param>
+        /// <param name="actualValue">The actual value to which the Constraint was applied.</param>
+        /// <param name="nonMatchingItem">Actual item that does not match expected condition</param>
+        /// <param name="nonMatchingIndex">Non matching item index</param>
+        public EachItemConstraintResult(IConstraint constraint, object actualValue, object nonMatchingItem, int nonMatchingIndex)
+            : base(constraint, actualValue, false)
+        {
+            _nonMatchingItem = nonMatchingItem;
+            _nonMatchingItemIndex = nonMatchingIndex;
+        }
+
+        /// <summary>
+        /// Write constraint description, actual items, and non-matching item
+        /// </summary>
+        /// <param name="writer">The MessageWriter on which to display the message</param>
+        public override void WriteMessageTo(MessageWriter writer)
+        {
+            base.WriteMessageTo(writer);
+
+            var nonMatchingStr = $"  Non-matching item at index [{_nonMatchingItemIndex}]:  "
+                + MsgUtils.FormatValue(_nonMatchingItem);
+            writer.WriteLine(nonMatchingStr);
+        }
+    }
+}
--- a/src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
@@ -1,20 +1,43 @@
+// ***********************************************************************
+// Copyright (c) 2018 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
 namespace NUnit.Framework.Constraints
 {
     /// <summary>
     /// Provides a <see cref="ConstraintResult"/> for the constraints 
     /// that are applied to each item in the collection
     /// </summary>
-    public class EachItemConstraintResult : ConstraintResult
+    internal sealed class EachItemConstraintResult : ConstraintResult
     {
         private readonly object _nonMatchingItem;
         private readonly int _nonMatchingItemIndex;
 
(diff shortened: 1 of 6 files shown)


Dreamescaper:
Here I created EachItemConstraintResult, but used it only for Failure result, leaving simple ConstraintResult for Success, to avoid unnessessary checks in error message.
Hope that's OK.

src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
```diff
@@ -0,0 +1,40 @@
+namespace NUnit.Framework.Constraints
```
ChrisMaddock: This file needs to copyright header. Could you just copy it from another file, and change the year to 2018?

src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    /// <summary>
+    /// Provides a <see cref="ConstraintResult"/> for the constraints 
+    /// that are applied to each item in the collection
+    /// </summary>
+    public class EachItemConstraintResult : ConstraintResult
```
ChrisMaddock: Could this be an internal class? 🙂

src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
```diff
@@ -0,0 +1,40 @@
...
+            _nonMatchingItemIndex = nonMatchingIndex;
+        }
+
+        /// <summary>
+        /// Write constraint description, actual items, and non-matching item
+        /// </summary>
+        /// <param name="writer">The MessageWriter on which to display the message</param>
+        public override void WriteMessageTo(MessageWriter writer)
+        {
+            base.WriteMessageTo(writer);
```
ChrisMaddock: Is there a reason not to use the `WriteAdditionalLinesTo(MessageWriter writer)` method here, like the other ConstraintResults? e.g. https://github.com/nunit/nunit/blob/ffcfba9bdd85792f6aa32dada77a4b044857b4bf/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs#L56

src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            var nonMatchingStr = $"  Non-matching item at index [{_nonMatchingItemIndex}]:  "
```
ChrisMaddock: This is just the **first** non-matching item, right? Would it be good to say that in the message? (i.e. "First non-matching item at index...")

src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// that are applied to each item in the collection
+    /// </summary>
+    public class EachItemConstraintResult : ConstraintResult
+    {
+        private readonly object _nonMatchingItem;
+        private readonly int _nonMatchingItemIndex;
+
+        /// <summary>
+        /// Constructs a <see cref="EachItemConstraintResult" /> for a particular <see cref="Constraint" />.
+        /// Only used for Failure
```
ChrisMaddock: I wouldn't normally be too worried, but given I've suggested other changes anyway, would you mind just tidying up the full-stops in the doc-comments? There's just a mix of some ending with full stops and some not at the moment. 😄

ChrisMaddock:
> Here I created EachItemConstraintResult, but used it only for Failure result, leaving simple ConstraintResult for Success, to avoid unnessessary checks in error message.
Hope that's OK.

Also - thanks for asking this, I couldn't make my mind up myself! 😅 I understand that there's nothing to be gained from using EachItemConstraintResult for the success case, but it does feel a little murky in terms of the architecture to return different a different ConstraintResult for success and failure. Anyone on @nunit/framework-team have a stronger opinion here?

src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
ChrisMaddock: (And sealed too, I believe!)

Dreamescaper: Can't see why not.
I tried to make it consistent with other classes, which are public (CollectionEquivalentConstraintResult, EqualConstraintResult)

src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
Dreamescaper: This method added only a week ago.
Will update accordingly.

Dreamescaper:
> Also - thanks for asking this, I couldn't make my mind up myself! 😅 I understand that there's nothing to be gained from using EachItemConstraintResult for the success case, but it does feel a little murky in terms of the architecture to return different a different ConstraintResult for success and failure. Anyone on @nunit/framework-team have a stronger opinion here?

No problem, will update.

src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
ChrisMaddock: So it was! I didn't realise! 😅

src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
ChrisMaddock: Ah, ok! Historically, lots of classes in NUnit have been public, when actually there was no need for them to be visible to end users. That's unfortunately meant however that people have begun to rely on the behaviour/structure of classes we didn't intend to be public, and made it difficult for us to make changes.

With 'new code' - we're working on only exposing what we need to. 🙂

jnm2: Yes, `internal` or even better IMO, nested within the constraint class and `private`. Makes it super clear how it can be accessed and why.

jnm2:
Returning the same derived type no matter what is the way we have always done it, but it doesn't cause any problems. I understand where you're coming from and kind of like it.

Bottom line: do whatever you think is simplest. If this is a private nested class, you won't have to worry about validation inside the constraint result class.

src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
Dreamescaper: Can't make it nested, as it used both by AllItemsConstraint and NoItemConstraint

jnm2: We don't support C# 7.2 or I'd ask if it makes sense to nest in the base class and make it `private protected`.

src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
Dreamescaper: I wanted to look in other framework classes, which approach is commonly used - with or without full-stops.
Seems like I unconsciously chose correct approach - just mix them whatever you like :))))

Dreamescaper:
Updated PR after review.

Concerning using EachItemConstraintResult for failure only case, and why I decided to do that in first place.
If we want to make it work for success as well, I can see the following options:
- to have single constructor with `status` parameter. I don't like that, because I'd need to put null/default values for item value and index in Constraint.
- to have two constructors - one for success case (without item value and index parameters), one for failure case - both without `status` parameter. That's better, but it's not fully readable - it's not obvious in code whether I create success or failure result.

If we strongly think that we should use same ConstraintResult for both success and failure - I'd rather have single private constructor, and two static methods Success and Failure.

As for now I left it as before, let me know what you think about that. Thanks!

src/NUnitFramework/framework/Constraints/EachItemConstraintResult.cs
ChrisMaddock: Haha - yes, room for improvement in the codebase there too! 😉

ChrisMaddock:
Yeah...I agree it wouldn't particular particularly nice that way either. The static `Success`, `Failure` methods were what I was thinking too - but it's a trade-off either way. I have no strong opinion either way, so let's go with how it is!

Thanks for making the changes - I'll merge now. And thanks for the contribution! 😄

PR #2859, opened 2018-05-15
Lock-free CountdownEvent and lazy ManualResetEventSlim polyfills

jnm2:
Maybe this will help us achieve performance parity for our net35 tests?


jnm2:
Thrilling. [VSTS's failure](https://github.com/nunit/nunit/blob/09bdce775b46fef62b01c16024c873465830c67a/src/NUnitFramework/tests/Internal/EventQueueTests.cs#L195) is a 30ms timeout. This is the first time this error has happened as far as I can tell.
Going to poke it and see what happens.

jnm2:
This time VSTS is the only one failing again, but it's a new one: `DequeueBlocking_StopTest.DequeueBlocking_Stop`. At face value this seems to be degrading performance.

jnm2:
Actually, the first error is not a new error. It's happening at https://github.com/nunit/nunit/pull/2863#issuecomment-389688796.

jnm2:
@rprouse Would you be interested in this in principle, or meh?

rprouse:
@jnm2 I would be interested, but don't we need more to utilize these new classes in .NET 2.0/3.5?

jnm2:
@rprouse Nope, this is self-contained. I removed overloads that took CancellationToken, for instance.

jnm2:
Was playing around today and have an idea that might fix our hangs. We should not be replacing `Thread.Yield` with `Thread.Sleep(0)` since Sleep(0) often does not yield and allow other threads to get work done. `Thread.Sleep(1)` seems like the best replacement available.

http://joeduffyblog.com/2006/08/22/priorityinduced-starvation-why-sleep1-is-better-than-sleep0-and-the-windows-balance-set-manager/

Going to open a separate PR though, then base this on that.

CharliePoole:
I agree and also have a vague memory of our having decided this some time in the past.

jnm2:
PR: #3019

jnm2:
@oznetmaster Yes, in `EventPump`. See https://github.com/nunit/nunit/pull/3019#issuecomment-421201479. There was already a Sleep(0) in SpinWait at the time.

oznetmaster:
@jnm2 that was the article I quoted three years ago to get all the sleep(0) changed to sleep(1).

The reasoning has not changed, as far as I know.

jnm2:
@oznetmaster Sadly, it's evidently not the cause of the hang. (https://github.com/nunit/nunit/pull/3019#issuecomment-421205698) Gotta start from scratch. Any ideas?

jnm2:
What do you think, should I rebase this and get it in or close?

jnm2:
I don't think net35 is worth investing this amount of code.

PR #2860, opened 2018-05-15
Consistent netstandard1.6 behavior (System.Diagnostics.Process)

jnm2:
The recommendation from the corefx team is that we should expect .NET Standard platform extensions like System.Diagnostics.Process to just work.

### DIFF
--- a/nuget/framework/nunit.nuspec
+++ b/nuget/framework/nunit.nuspec
@@ -30,6 +30,7 @@ Supported platforms:
       <group targetFramework="net45" />
       <group targetFramework="netstandard1.6">
         <dependency id="NETStandard.Library" version="1.6.1" />
+        <dependency id="System.Diagnostics.Process" version="4.3.0" />
       </group>
       <group targetFramework="netstandard2.0">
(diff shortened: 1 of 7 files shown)


ChrisMaddock:
Do these packages work on our Xamarin platforms? Last time we tried this with Threading and Loader, neither did. (Although do all our xamarin targets use 2.0 now? I lose track)

jnm2:
I'll run some manual tests. If they don't, the intended solution is feature detection rather than limiting the .NET Standard 1.6 build to the lowest common denominator.

jnm2:
I'll run some manual tests. If they don't, the intended solution is feature detection rather than limiting the .NET Standard 1.6 build to the lowest common denominator. There are a couple scenarios where our .NET Standard 1.6 build may run on .NET Framework and .NET Core, as well.

ChrisMaddock:
Sounds good! We really need proper integration testing here - I just don't want to accidentally break the package for Xamarin, like we have done before!

jnm2:
.NET Core 1.x is nearly at end of life, so I can't imagine this would bring enough value to folks to be worth the effort.

PR #2861, opened 2018-05-15
Consistent netstandard1.6 behavior (System.Xml.XmlDocument)

jnm2:
Same as https://github.com/nunit/nunit/pull/2860.


jnm2:
.NET Core 1.x is nearly at end of life, so I can't imagine this would bring enough value to folks to be worth the effort.

PR #2873, opened 2018-05-28
Assert.That with a Throws constraint does not provide as much …  …info as Assert.Throws

### LINKED ISSUE #2146: Assert.That with a Throws constraint does not provide as much info as Assert.Throws
When using `Assert.Throws` and an unexpected exception is thrown nunit provides information about the unexpected exception. 

```csharp
[Test]
public void UnexpectedExceptionWithInfo()
{
  Assert.Throws<IOException>(() =>
  {
    throw new FormatException("This is a format exception.");
  });
}
```

This test expects an `IOException` to be thrown but instead a `FormatException` is thrown. This provides the following output:

```
 Expected: <System.IO.IOException>
  But was:  <System.FormatException: This is a format exception.
   at Testit.Tests.<>c.<UnexpectedExceptionWithInfo>b__0_0() in Program.cs:line 23
   at NUnit.Framework.Assert.Throws(IResolveConstraint expression, TestDelegate code, String message, Object[] args)>

at Testit.Tests.UnexpectedExceptionWithInfo() in Program.cs:line 21
```

It is providing information about the unexpected `FormatException` (message + stack trace). The problem is that in our codebase we are using `Assert.That` with a `Throws.TypeOf<>` constraint like this:

```
[Test]
public void UnexpectedExceptionWithoutInfo()
{
  Assert.That(() =>
  {
    throw new FormatException("This is a format exception.");
  }, Throws.TypeOf<IOException>());
}
```

But when this example encounters the unexpected `FormatException` the following is output:

```
  Expected: <System.IO.IOException>
  But was:  <System.FormatException>

at Testit.Tests.UnexpectedExceptionWithoutInfo() in Program.cs:line 30
```

Note here that the information about the unexpected exception (message + stack trace) is gone. But exactly those two exception details would be relavant for debugging the failed test.

Is this behaiviour intended? And if yes are there easy ways to fix this issue?

My current workaround is to use a `ThrowsConstraint` with an `ExceptionTypeConstraint` as base constraint: 

```
[Test]
public void UnexpectedExceptionFixedInfo()
{
  Assert.That(() =>
  {
    throw new FormatException("This is a format exception.");
  }, new ThrowsConstraint(new ExceptionTypeConstraint(typeof(IOException))));
}
```

ZenonBuratta:
Added a type check into the TypeConstraint ApplyTo method in order to set the actual value as the object rather than the type when the type is an exception.

Fixes #2146

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionEquivalentConstraintResult.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2017 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -51,13 +51,10 @@ namespace NUnit.Framework.Constraints
             _tallyResult = tallyResult;
         }
 
-        /// <summary>Write the custom failure message for this object's <see cref="CollectionEquivalentConstraint"/>.</summary>
+        /// <summary>Write any additional lines (following <c>Expected:</c> and <c>But was:</c>) for a failing constraint.</summary>
         /// <param name="writer">The <see cref="MessageWriter"/> to write the failure message to.</param>
-        public override void WriteMessageTo(MessageWriter writer)
-        {
-            //Write the expected/actual message first.
-            base.WriteMessageTo(writer);
-            
+        public override void WriteAdditionalLinesTo(MessageWriter writer)
+        { 
             if (_tallyResult.MissingItems.Count > 0)
             {
                 int missingItemsCount = _tallyResult.MissingItems.Count;
--- a/src/NUnitFramework/framework/Constraints/ConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintResult.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2011 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -51,7 +51,7 @@ namespace NUnit.Framework.Constraints
     }
 
     /// <summary>
-    /// Contain the result of matching a <see cref="Constraint"/> against an actual value.
+    /// Contains the result of matching a <see cref="Constraint"/> against an actual value.
     /// </summary>
     public class ConstraintResult
     {
@@ -144,6 +144,15 @@ namespace NUnit.Framework.Constraints
         public virtual void WriteMessageTo(MessageWriter writer)
         {
             writer.DisplayDifferences(this);
+        }
+
+        /// <summary>
+        /// Write some additional failure message.
+        /// </summary>
+        /// <param name="writer">The MessageWriter on which to display the message</param>
+        public virtual void WriteAdditionalLinesTo(MessageWriter writer)
+        {
+            //By default it does not write anything to writer but can be overriden in classes where needed. 
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
@@ -70,7 +70,7 @@ namespace NUnit.Framework.Constraints
 
             propValue = property.GetValue(actual, null);
             var baseResult = BaseConstraint.ApplyTo(propValue);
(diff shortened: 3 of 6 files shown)


mikkelbu:
> @mikkelbu I gather you're shepherding this one through?
I can try. I've edited the description to close the issue and will take a look at the code now.

jnm2:
@mikkelbu I'm not sure if you're talking about me, but since the merge commit has no effect on the PR diff, it is fine to stay as far as I know.

src/NUnitFramework/framework/Constraints/TypeConstraint.cs
```diff
@@ -61,6 +61,12 @@ protected TypeConstraint(Type type, string descriptionPrefix)
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
             actualType = actual == null ? null : actual.GetType();
+
+            if (actual is Exception)
```
jnm2: Like you, I see this as a code smell. But I don't have a better suggestion and I can see us opening this up to more types in the future.

src/NUnitFramework/tests/Constraints/ThrowsConstraintTests.cs
```diff
@@ -53,7 +53,7 @@ public void SetUp()
     }
 
     [TestFixture]
-    public class ThrowsConstraintTest_InstanceOfType : ConstraintTestBase
+    public class ThrowsConstraintTest_InstanceOfType : ThrowsConstraintTestBase
```
jnm2: @ZenonBuratta Can you help me understand this change?

src/NUnitFramework/tests/Syntax/ThrowsTests.cs
```diff
@@ -153,5 +192,6 @@ public MyClass(string s)
                     throw new ArgumentNullException();
             }
         }
+
```
jnm2: ℹ It's too late to fix now unless you amend the same commit that introduced this. For the future, it's good practice to review the diff in Visual Studio or the git CLI before each commit to catch stuff like this. Makes for a weird git blame history.

mikkelbu: If we squash merge then it would be ok to revert this with a separate commit?

src/NUnitFramework/tests/Constraints/ThrowsConstraintTests.cs
mikkelbu: I also wondered about this, so I took a look at the code. All other classes in this file inherits from `ThrowsConstraintTestBase`, and the only difference between `ConstraintTestBase` and `ThrowsConstraintTestBase` is the following:
https://github.com/nunit/nunit/blob/7b87bdc7b48ee1b13023dd282efea449c2d66b81/src/NUnitFramework/tests/Constraints/ConstraintTestBase.cs#L72-L74
vs
https://github.com/nunit/nunit/blob/f94ea919fe344fc656c84052d885cbd3c4aedc71/src/NUnitFramework/tests/Constraints/ThrowsConstraintTests.cs#L131-L133
Hence the assert is changed from `Is.EqualTo` to `Does.StartWith`. Perhaps a bit overkill to have two so similar classes, but that is another issue 😄.

ZenonBuratta: Yes @mikkelbu. The change of base class is the ThrowsConstraintTests uses a Does.StartWith. This is because of the new stack trace that comes with the exception and due to not wanting to make the tests become fragile, I only want to check the message.

src/NUnitFramework/tests/Syntax/ThrowsTests.cs
ZenonBuratta: Happy to do this if you want?

jnm2: True, yes!

mikkelbu: @ZenonBuratta Please do 👍. Not that this is a major problem, especially in the end of a test class, but still nice to have the history as clean as possible.

jnm2:
@ZenonBuratta If you leave the PR open, it will pick up pushes (including force pushes).

jnm2:
Oops, it looks like you force-pushed an empty branch :-D It says it'll be able to be reopened after you push some commits.

ZenonBuratta:
Yeah I have destroyed this PR. I will create a new clean one as I don't want to merge from my forks master. Big mistake, wasn't thinking when I did it.

jnm2:
Ah, I see. No problem!

ZenonBuratta:
@jnm2 @mikkelbu I have opened PR https://github.com/nunit/nunit/pull/2880

PR #2878, opened 2018-05-30
Implement prefiltering in NUnitLite

### LINKED ISSUE #2876: Implement Discovery-time filtering for NUnitLite
Issue #2867 asks for us to avoid running TestCaseSource code when the generated tests would be excluded by the filter that has been provided.

As a start on this, this issue will make changes to the framework and implement discover-time filtering that is available when running NUnitLite. This is a handy way to work since the framework and nunitlite are all part of the same repository and solution.

CharliePoole:
Fixes #2876

Allows use of pre-filters that apply before the tests are discovered.

### DIFF
--- a/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
+++ b/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
@@ -48,9 +48,6 @@ namespace NUnit.Framework.Api
         /// </summary>
         readonly ISuiteBuilder _defaultSuiteBuilder;
 
-        /// <summary>
-        /// The PreFilter used by this builder
-        /// </summary>
         private PreFilter _filter;
 
         #endregion
@@ -174,7 +171,7 @@ namespace NUnit.Framework.Api
 
                 _filter = new PreFilter();
                 if (options.ContainsKey(FrameworkPackageSettings.LOAD))
-                    foreach (string filterText in (IList<string>)options[FrameworkPackageSettings.LOAD])
+                    foreach (string filterText in (IList)options[FrameworkPackageSettings.LOAD])
                         _filter.Add(filterText);
 
                 var fixtures = GetFixtures(assembly);
@@ -238,7 +235,7 @@ namespace NUnit.Framework.Api
             var result = new List<Type>();
 
             foreach (Type type in assembly.GetTypes())
-                if (_filter.Match(type))
+                if (_filter.IsMatch(type))
                     result.Add(type);
 
             return result;
--- a/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
+++ b/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
@@ -48,6 +48,11 @@ namespace NUnit.Framework.Api
         /// </summary>
         readonly ISuiteBuilder _defaultSuiteBuilder;
 
+        /// <summary>
+        /// The PreFilter used by this builder
+        /// </summary>
+        private PreFilter _filter;
+
         #endregion
 
         #region Constructor
@@ -167,10 +172,12 @@ namespace NUnit.Framework.Api
                     }
                 }
 
-                IList fixtureNames = null;
+                _filter = new PreFilter();
                 if (options.ContainsKey(FrameworkPackageSettings.LOAD))
-                    fixtureNames = options[FrameworkPackageSettings.LOAD] as IList;
-                var fixtures = GetFixtures(assembly, fixtureNames);
+                    foreach (string filterText in (IList<string>)options[FrameworkPackageSettings.LOAD])
+                        _filter.Add(filterText);
+
+                var fixtures = GetFixtures(assembly);
 
                 testAssembly = BuildTestAssembly(assembly, suiteName, fixtures);
             }
@@ -187,12 +194,12 @@ namespace NUnit.Framework.Api
 
         #region Helper Methods
 
-        private IList<Test> GetFixtures(Assembly assembly, IList names)
+        private IList<Test> GetFixtures(Assembly assembly)
         {
             var fixtures = new List<Test>();
             log.Debug("Examining assembly for test fixtures");
 
-            var testTypes = GetCandidateFixtureTypes(assembly, names);
+            var testTypes = GetCandidateFixtureTypes(assembly);
 
             log.Debug("Found {0} classes to examine", testTypes.Count);
 #if LOAD_TIMING
@@ -207,10 +214,11 @@ namespace NUnit.Framework.Api
                 // of DefaultSuiteBuilder.CanBuildFrom cannot invoke any user code.
                 if (_defaultSuiteBuilder.CanBuildFrom(testType))
                 {
+                    // We pass the filter for use in selecting methods of the type.
                     // Any exceptions from this call are fatal problems in NUnit itself,
                     // since this is always DefaultSuiteBuilder and the current implementation
                     // of DefaultSuiteBuilder.BuildFrom handles all exceptions from user code.
-                    Test fixture = _defaultSuiteBuilder.BuildFrom(testType);
+                    Test fixture = _defaultSuiteBuilder.BuildFrom(testType, _filter);
                     fixtures.Add(fixture);
                     testcases += fixture.TestCaseCount;
                 }
@@ -225,22 +233,13 @@ namespace NUnit.Framework.Api
             return fixtures;
         }
 
-        private IList<Type> GetCandidateFixtureTypes(Assembly assembly, IList names)
+        private IList<Type> GetCandidateFixtureTypes(Assembly assembly)
         {
-            var types = assembly.GetTypes();
-
-            if (names == null || names.Count == 0)
-                return types;
-
             var result = new List<Type>();
 
-            foreach (Type type in types)
-                foreach (string name in names)
-                    if (type.FullName == name || type.FullName.StartsWith(name + "."))
-                    {
-                        result.Add(type);
-                        break;
-                    }
+            foreach (Type type in assembly.GetTypes())
+                if (_filter.Match(type))
+                    result.Add(type);
 
             return result;
         }
--- a/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
+++ b/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
@@ -234,20 +234,13 @@ namespace NUnit.Framework.Api
 
             var result = new List<Type>();
 
-            foreach (string name in names)
-            {
-                Type fixtureType = assembly.GetType(name);
-                if (fixtureType != null)
-                    result.Add(fixtureType);
-                else
-                {
-                    string prefix = name + ".";
-
-                    foreach (Type type in types)
-                        if (type.FullName.StartsWith(prefix))
-                            result.Add(type);
-                }
-            }
+            foreach (Type type in types)
+                foreach (string name in names)
+                    if (type.FullName == name || type.FullName.StartsWith(name + "."))
+                    {
+                        result.Add(type);
+                        break;
+                    }
 
             return result;
         }
--- a/src/NUnitFramework/framework/Interfaces/IFixtureBuilder.cs
+++ b/src/NUnitFramework/framework/Interfaces/IFixtureBuilder.cs
@@ -28,6 +28,9 @@ namespace NUnit.Framework.Interfaces
 {
     using Internal;
 
+    // TODO: These methods should really return IEnumerable<TestFixture>, 
+    // but that requires changes to the Test hierarchy.
+
     /// <summary>
     /// The IFixtureBuilder interface is exposed by a class that knows how to
     /// build test fixtures from a specified type. In general, it is exposed
@@ -40,7 +43,20 @@ namespace NUnit.Framework.Interfaces
         /// Builds any number of test fixtures from the specified type.
         /// </summary>
         /// <param name="type">The type to be used as a fixture.</param>
-        // TODO: This should really return a TestFixture, but that requires changes to the Test hierarchy.
         IEnumerable<TestSuite> BuildFrom(Type type);
+    }
+
+    /// <summary>
+    /// The IFixtureBuilder2 interface extends IFixtureBuilder by allowing
+    /// use of a PreFilter, which is used to select methods as test cases.
+    /// </summary>
+    public interface IFixtureBuilder2 : IFixtureBuilder
+    {
+        /// <summary>
+        /// Builds any number of test fixtures from the specified type.
+        /// </summary>
+        /// <param name="type">The type to be used as a fixture.</param>
+        /// <param name="filter">PreFilter to be used to select methods.</param>
+        IEnumerable<TestSuite> BuildFrom(Type type, PreFilter filter);
     }
 }
--- a/src/NUnitFramework/framework/Internal/Builders/DefaultSuiteBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/DefaultSuiteBuilder.cs
@@ -83,7 +83,7 @@ namespace NUnit.Framework.Internal.Builders
         /// <param name="type">The Type to be used as a suite.</param>
         /// <param name="filter">A PreFilter for selecting methods.</param>
         /// <returns></returns>
-        public TestSuite BuildFrom(Type type, PreFilter filter)
+        public TestSuite BuildFrom(Type type, IPreFilter filter)
         {
             var fixtures = new List<TestSuite>();
 
--- a/src/NUnitFramework/framework/Internal/Builders/DefaultSuiteBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/DefaultSuiteBuilder.cs
@@ -73,6 +73,18 @@ namespace NUnit.Framework.Internal.Builders
         /// <param name="type">The type to be used as a suite.</param>
         public TestSuite BuildFrom(Type type)
         {
+            return BuildFrom(type, PreFilter.Empty);
+        }
+
+        /// <summary>
+        /// Builds a single test suite from the specified type, subject
+        /// to a filter that decides which methods are included.
+        /// </summary>
+        /// <param name="type">The Type to be used as a suite.</param>
+        /// <param name="filter">A PreFilter for selecting methods.</param>
+        /// <returns></returns>
+        public TestSuite BuildFrom(Type type, PreFilter filter)
+        {
             var fixtures = new List<TestSuite>();
 
             try
@@ -80,8 +92,13 @@ namespace NUnit.Framework.Internal.Builders
                 IFixtureBuilder[] builders = GetFixtureBuilderAttributes(type);
 
                 foreach (var builder in builders)
-                    foreach (var fixture in builder.BuildFrom(type))
+                {
+                    // See if this is an enhanced attribute, accepting a filter
+                    var builder2 = builder as IFixtureBuilder2;
+
+                    foreach (var fixture in builder2?.BuildFrom(type, filter) ?? builder.BuildFrom(type))
                         fixtures.Add(fixture);
+                }
 
                 if (type.GetTypeInfo().IsGenericType)
                     return BuildMultipleFixtures(type, fixtures);
@@ -89,7 +106,7 @@ namespace NUnit.Framework.Internal.Builders
                 switch (fixtures.Count)
                 {
                     case 0:
-                        return _defaultBuilder.BuildFrom(type);
+                        return _defaultBuilder.BuildFrom(type, filter);
                     case 1:
                         return fixtures[0];
                     default:
@@ -107,6 +124,7 @@ namespace NUnit.Framework.Internal.Builders
                 return fixture;
             }
         }
+
         #endregion
 
         #region Helper Methods
--- a/src/NUnitFramework/framework/Internal/PreFilter.cs
+++ b/src/NUnitFramework/framework/Internal/PreFilter.cs
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2014-2018 Charlie Poole, Rob Prouse
+// Copyright (c) 2018 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
@@ -24,19 +24,20 @@
 using System;
 using System.Collections.Generic;
 using System.Reflection;
+using NUnit.Framework.Interfaces;
 
 namespace NUnit.Framework.Internal
 {
     /// <summary>
-    /// The PreFilter class implements a simplified filter for use in deciding which
+    /// Implements a simplified filter for use in deciding which
     /// Types and Methods should be used to generate tests. It is consructed with a 
     /// list of strings, each of which may end up being interpreted in various ways.
     /// </summary>
-    public class PreFilter
+    internal class PreFilter : IPreFilter
     {
         private static readonly char[] ARG_START = new char[] { '(', '<' };
 
-        private List<FilterElement> _filters = new List<FilterElement>();
+        private readonly List<FilterElement> _filters = new List<FilterElement>();
 
         /// <summary>
         /// Return a new PreFilter, without elements, which is considered
@@ -48,7 +49,7 @@ namespace NUnit.Framework.Internal
         }
 
         /// <summary>
-        /// Return true if the filer is empty, in which case it
+        /// Return true if the filter is empty, in which case it
         /// always succeeds. Technically, this is just a filter and
         /// you can add elements but it's best to use Empty when
         /// you need an empty filter and new when you plan to add.
@@ -64,6 +65,24 @@ namespace NUnit.Framework.Internal
         /// <param name="filterText"></param>
         public void Add(string filterText)
         {
+            // First, make sure the new filter is not redundant
+            foreach (var filter in _filters)
+            {
+                if (filter.Text == filterText)
+                    return;
+
+                if (filterText.StartsWith(filter.Text + "."))
+                    return;
+            }
+
+            var newFilter = new FilterElement(filterText);
+
+            // Check to see if it makes any of the existing
+            // filter elements redundant.
+            for (int index = _filters.Count - 1; index >= 0; index--)
+                if (_filters[index].Text.StartsWith(filterText + "."))
+                    _filters.RemoveAt(index);
+
             _filters.Add(new FilterElement(filterText));
         }
 
@@ -71,7 +90,7 @@ namespace NUnit.Framework.Internal
         /// Use the filter on a Type, returning true if the type matches the filter
         /// and should therefore be included in the discovery process.
         /// </summary>
-        public bool Match(Type type)
+        public bool IsMatch(Type type)
         {
             if (IsEmpty)
                 return true;
@@ -91,7 +110,7 @@ namespace NUnit.Framework.Internal
         /// Use the filter on a Type, returning true if the type matches the filter
         /// and should therefore be included in the discovery process.
         /// </summary>
-        public bool Match(Type type, MethodInfo method)
+        public bool IsMatch(Type type, MethodInfo method)
         {
             if (IsEmpty)
                 return true;
@@ -127,7 +146,7 @@ namespace NUnit.Framework.Internal
 
         private class FilterElement
         {
-            public FilterElementType ElementType = FilterElementType.Unknown;
+            private FilterElementType ElementType = FilterElementType.Unknown;
             public string Text;
             public string ClassName;
             public string MethodName;
--- a/src/NUnitFramework/framework/Internal/PreFilter.cs
+++ b/src/NUnitFramework/framework/Internal/PreFilter.cs
@@ -1,0 +1,208 @@
+// ***********************************************************************
+// Copyright (c) 2014-2018 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace NUnit.Framework.Internal
+{
+    /// <summary>
+    /// The PreFilter class implements a simplified filter for use in deciding which
+    /// Types and Methods should be used to generate tests. It is consructed with a 
+    /// list of strings, each of which may end up being interpreted in various ways.
+    /// </summary>
+    public class PreFilter
+    {
+        private static readonly char[] ARG_START = new char[] { '(', '<' };
+
+        private List<FilterElement> _filters = new List<FilterElement>();
+
+        /// <summary>
+        /// Return a new PreFilter, without elements, which is considered
+        /// empty and always matches.
+        /// </summary>
+        public static PreFilter Empty
+        {
+            get { return new PreFilter(); }
+        }
+
+        /// <summary>
+        /// Return true if the filer is empty, in which case it
+        /// always succeeds. Technically, this is just a filter and
+        /// you can add elements but it's best to use Empty when
+        /// you need an empty filter and new when you plan to add.
+        /// </summary>
+        public bool IsEmpty
+        {
+            get { return _filters.Count == 0; }
+        }
+
+        /// <summary>
+        /// Add a new filter element to the filter
+        /// </summary>
+        /// <param name="filterText"></param>
+        public void Add(string filterText)
+        {
+            _filters.Add(new FilterElement(filterText));
+        }
+
+        /// <summary>
+        /// Use the filter on a Type, returning true if the type matches the filter
+        /// and should therefore be included in the discovery process.
+        /// </summary>
+        public bool Match(Type type)
+        {
+            if (IsEmpty)
+                return true;
+
+            string typeName = type.FullName;
+
+            foreach (FilterElement filter in _filters)
+            {
+                if (filter.Match(type))
+                    return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Use the filter on a Type, returning true if the type matches the filter
+        /// and should therefore be included in the discovery process.
+        /// </summary>
+        public bool Match(Type type, MethodInfo method)
+        {
+            if (IsEmpty)
+                return true;
+
+            int filterCount = 0;
+            foreach (FilterElement filter in _filters)
+            {
+                if (type.FullName == filter.ClassName)
+                {
+                    ++filterCount;
+                    if (method.Name == filter.MethodName)
+                        return true;
+                }
+            }
+
+            // If no match, we succeed only if there were no method filters for this fixture
+            return filterCount == 0;
+        }
+
+        #region Nested FilterElementType Enumeration
+
+        private enum FilterElementType
+        {
+            Unknown,
+            Namespace,
+            Fixture,
+            Method
+        }
+
+        #endregion
+
+        #region Nested FilterElement Class
+
+        private class FilterElement
+        {
+            public FilterElementType ElementType = FilterElementType.Unknown;
+            public string Text;
+            public string ClassName;
+            public string MethodName;
+
+            public FilterElement(string text)
+            {
+                int argStart = text.IndexOfAny(ARG_START);
+                Text = argStart > 0
+                    ? text.Substring(0, argStart)
+                    : text;
+
+                // Filter may be the fullname of a test method, if so extract class name
+                int lastdot = Text.LastIndexOf('.');
+                if (lastdot > 0)
+                {
+                    ClassName = Text.Substring(0, lastdot);
+                    MethodName = Text.Substring(lastdot + 1);
+                }
+            }
+
+            public bool Match(Type type)
+            {
+                switch(ElementType)
+                {
+                    default:
+                    case FilterElementType.Unknown:
+                        return MatchUnknownElement(type);
+                    case FilterElementType.Fixture:
+                        return MatchFixtureElement(type);
+                    case FilterElementType.Namespace:
+                        return MatchNamespaceElement(type);
+                    case FilterElementType.Method:
+                        return MatchMethodElement(type);
+                }
+            }
+
+            private bool MatchUnknownElement(Type type)
+            {
+                if (MatchFixtureElement(type))
+                {
+                    ElementType = FilterElementType.Fixture;
+                    return true;
+                }
+
+                if (MatchNamespaceElement(type))
+                {
+                    ElementType = FilterElementType.Namespace;
+                    return true;
+                }
+
+                if (MatchMethodElement(type))
+                {
+                    ElementType = FilterElementType.Method;
+                    return true;
+                }
+
+                return false;
+            }
+
+            private bool MatchFixtureElement(Type type)
+            {
+                return type.FullName == Text;
+            }
+
+            private bool MatchNamespaceElement(Type type)
+            {
+                return type.FullName.StartsWith(Text + ".");
+            }
+
+            private bool MatchMethodElement(Type type)
+            {
+                return type.FullName == ClassName;
+            }
+        }
+
+        #endregion
+    }
+}
--- a/src/NUnitFramework/framework/nunit.framework.csproj
+++ b/src/NUnitFramework/framework/nunit.framework.csproj
@@ -6,6 +6,11 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net20|AnyCPU'">
+    <DebugType>full</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+  </PropertyGroup>
+
   <ItemGroup Condition="'$(TargetFramework)' == 'net20'">
     <PackageReference Include="NUnit.System.Linq" Version="0.6.0" />
   </ItemGroup>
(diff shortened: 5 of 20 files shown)


CharliePoole:
@jnm2 I believe you had some integration tests with significant timing issues in test case source. Are you able to use NUnitLite to run them and see whether the pre-filtering helps? Next step would be to make it automatic, at least in some cases, rather than manually specified.

jnm2:
@CharliePoole Yes, one TestCaseSource in particular causes a four-second delay. It could be simulated effectively with Thread.Sleep, but I will pull this code and give it a whirl.

jnm2:
@CharliePoole Is the TestCaseSource skipping implemented in this code? I'm overlooking it.

CharliePoole:
It doesn't require any special code to skip it. If a pre-filter is set, the LOAD setting causes only those namespaces and fixtures to be used in creating tests. For unselected fixtures, no tests are created so the related TestCaseSource is never invoked.

You don't see a change that makes it happen because the LOAD setting has been present forever. What I have done is exposed it via a command-line option and provided a bit more information in the output, including how long the discovery took.

The intent of this commit is to give us a chance to play with using a simple (only fixtures and namespaces) pre-filter. I think it will give us the performance boost we want. Then we have to decide where to take it next, which I'm inclined to say should be to interpret a normal filter and create a pre-filter from it.

Note that my sample runs prove that you don't have to remove the original filters. Running a redundant pre-filter and filter pair as in the third row of my table does no harm. That changes the original problem from one of transforming the original filter to simply generating a supplementary pre-filter for efficiency, which is easier.

CharliePoole:
With the last commit, this is basically in the shape I want it but is lacking support for pre-filtering individual methods.

That is, if you indicate you want to run a single method, via the `--test` or `--prefilter`, __all__ the tests in the same fixture are generated, including execution of any `TestCaseSource` code. We would rather generate __only__ the specified test method, obviously.

Another missing feature: we generate pre-filters from the `--test` option, but not from the `--where` option.

@jnm2 Have you tried this out on your tests?

src/NUnitFramework/framework/nunit.framework.csproj
```diff
@@ -6,6 +6,11 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net20|AnyCPU'">
+    <DebugType>full</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+  </PropertyGroup>
```
CharliePoole: I had to do this to be able to debug. I can remove it if somebody can tell me why I'm unable to debug (break) using the portable pdb output, which is the default.

src/NUnitFramework/nunitlite/nunitlite.csproj
```diff
@@ -5,6 +5,11 @@
     <RootNamespace>NUnitLite</RootNamespace>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|net20|AnyCPU'">
+    <DebugType>full</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+  </PropertyGroup>
```
CharliePoole: Same as above. I was unable to break at a breakpoint without this change.

CharliePoole:
Ready for review and merge.

What's in it:

1. --prefilter option to allow loading tests based on namespace, class or method. If you use this, then no code is executed for the tests that are excluded.

2. Automatic prefiltering if you use --test. Note that --test can specify a lower level of detail than --prefilter, so there is some added selection that remains after pre-filtering. This may lead to a few test case sources executing even though the specific cases are not selected. I'm guessing most people will stick with this option.

3. You can specify both --prefilter and --test but you will only get those tests that are selected by __both__ options.

4. The nunitlite console output has a bit more info about discovery and the use of prefilters.

What's not here

* No automatic translation of the --where clause to a prefilter. That's theoretically possible and desirable, but I think there's enough here to merge this PR and deal with the --where clause as a later issue. Note that this means `--test X.Y.Z` is pre-filtered but `--where test=X.Y.Z` is not.

oznetmaster:
Does the --testlist option also invoke the automatic prefilter?

If it doesn't, it probably should.

CharliePoole:
Yes, the `CommandLineOptions` class makes both `--test` and `--testlist` basically look the same to the rest of the code.

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+                if (filter.Match(type))
+                    return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Use the filter on a Type, returning true if the type matches the filter
+        /// and should therefore be included in the discovery process.
```
oznetmaster: Comment should be modified to indicate it is a filter on a type and a method.

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+            }
+
+            private bool MatchNamespaceElement(Type type)
+            {
+                return type.FullName.StartsWith(Text + ".");
+            }
+
+            private bool MatchMethodElement(Type type)
+            {
+                return type.FullName == ClassName;
```
oznetmaster: I am confused.  What happened to MethodName?

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: This particular call relates only to the Type

CharliePoole:
The PreFilter is a bit odd because it can represent a namespace, type or method and we only learn which it is in the course of trying to apply it.

This application takes place in two stages. First the framework has to decide which Types to examine. In the course of successfully matching a type, a filter is characterized as one of the three kinds. If it has a method name, that is saved for use later.

The second stage is handled by attributes, including custom attributes. They are passed the pre-filter for use in selecting methods to be examined.

This is all very ad-hoc. It has to be due to the lack of structure in the input.

jnm2:
@CharliePoole

Here's my repro with the same characteristics as the real-world problem. With `--test=FastTests` using NUnitLite 3.10.1, `FastTest` is not entered until after ten seconds have passed and `SlowTestCaseSource` is fully enumerated:

<details>

```cs
using System.Linq;
using System.Threading;
using NUnit.Framework;

public static class FastTests 
{
    [Test]
    public static void FastTest()
    {
    }
}

public static class TestsWithSlowSource
{
    public static IEnumerable<TestCaseData> SlowTestCaseSource()
    {
        Thread.Sleep(10_000);
        return from n in Enumerable.Range(1, 10_000) select new TestCaseData(n);
    }

    [TestCaseSource(nameof(SlowTestCaseSource))]
    public static void Foo(int n)
    {
    }
}
```

</details>

Your builds at https://ci.appveyor.com/project/CharliePoole/nunit/build/3.11.0-ci-05630-pr-2878/artifacts work as expected.

I'm excited about this and looking forward to integration with Test Explorer and ReSharper.

src/NUnitFramework/nunitlite/CommandLineOptions.cs
```diff
@@ -169,6 +169,7 @@ private static IEnumerable<string> GetArgsFromFile(string filename)
         public string InputFile { get; private set; }
 
         public IList<string> TestList { get; } = new List<string>();
+        public IList<string> PreFilters { get; } = new List<string>();
```
jnm2: The whole list is the prefilter, so singular?

src/NUnitFramework/framework/nunit.framework.csproj
jnm2: I have a feeling this will be a problem. Are you running VS older than 2015 Update 2?

src/NUnitFramework/tests/Api/DefaultTestAssemblyBuilderTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            var result = _builder.Build(_mockAssemblyPath, settings);
+
+            Assert.That(result.IsSuite);
+            Assert.That(result, Is.TypeOf<TestAssembly>());
+            Assert.That(result.Name, Is.EqualTo(MOCK_ASSEMBLY_FILE));
+            Assert.That(result.RunState, Is.EqualTo(Interfaces.RunState.Runnable));
+
+            return result.TestCaseCount;
+        }
+
```
jnm2: Blank line before a closing brace

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -49,6 +49,11 @@ public class DefaultTestAssemblyBuilder : ITestAssemblyBuilder
         /// </summary>
         readonly ISuiteBuilder _defaultSuiteBuilder;
 
+        /// <summary>
+        /// The PreFilter used by this builder
```
jnm2: It doesn't like like the XML comment adds any info beyond what you see by looking at the field name.

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -164,10 +169,12 @@ private TestSuite Build(Assembly assembly, string assemblyPath, IDictionary<stri
                     }
                 }
 
-                IList fixtureNames = null;
+                _filter = new PreFilter();
                 if (options.ContainsKey(FrameworkPackageSettings.LOAD))
-                    fixtureNames = options[FrameworkPackageSettings.LOAD] as IList;
-                var fixtures = GetFixtures(assembly, fixtureNames);
+                    foreach (string filterText in (IList<string>)options[FrameworkPackageSettings.LOAD])
```
jnm2: We're going from requiring `IList` to requiring `IList<string>`. E.g. ArrayList would have worked before but now would be ignored. Is there any chance of this being a problem?

src/NUnitFramework/framework/Interfaces/IFixtureBuilder.cs
```diff
@@ -40,7 +43,20 @@ public interface IFixtureBuilder
...
         /// <param name="type">The type to be used as a fixture.</param>
-        // TODO: This should really return a TestFixture, but that requires changes to the Test hierarchy.
         IEnumerable<TestSuite> BuildFrom(Type type);
     }
+
+    /// <summary>
+    /// The IFixtureBuilder2 interface extends IFixtureBuilder by allowing
+    /// use of a PreFilter, which is used to select methods as test cases.
+    /// </summary>
+    public interface IFixtureBuilder2 : IFixtureBuilder
```
jnm2: What do you think of `IFilterableFixtureBuilder`?

src/NUnitFramework/framework/Internal/Builders/DefaultSuiteBuilder.cs
```diff
@@ -80,16 +92,21 @@ public TestSuite BuildFrom(Type type)
                 IFixtureBuilder[] builders = GetFixtureBuilderAttributes(type);
 
                 foreach (var builder in builders)
-                    foreach (var fixture in builder.BuildFrom(type))
+                {
+                    // See if this is an enhanced attribute, accepting a filter
+                    var builder2 = builder as IFixtureBuilder2;
```
jnm2: `var filterableBuilder`?

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+using System.Reflection;
+
+namespace NUnit.Framework.Internal
+{
+    /// <summary>
+    /// The PreFilter class implements a simplified filter for use in deciding which
+    /// Types and Methods should be used to generate tests. It is consructed with a 
+    /// list of strings, each of which may end up being interpreted in various ways.
+    /// </summary>
+    public class PreFilter
```
jnm2: `sealed` since there's nothing to override?

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace NUnit.Framework.Internal
+{
+    /// <summary>
+    /// The PreFilter class implements a simplified filter for use in deciding which
+    /// Types and Methods should be used to generate tests. It is consructed with a 
+    /// list of strings, each of which may end up being interpreted in various ways.
```
jnm2: What do you think of starting it `A simplified filter` and not capitalizing types and methods?
(Typo `consructed`)

> each of which may end up being interpreted in various ways.

I'm not sure I know what this is saying. Could you throw in an example of what you're thinking?

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+    /// </summary>
+    public class PreFilter
+    {
+        private static readonly char[] ARG_START = new char[] { '(', '<' };
```
jnm2: Would it be okay with you if this was `ArgStart`? Also you can remove the `new char[]` if you want.

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        private List<FilterElement> _filters = new List<FilterElement>();
```
jnm2: `readonly`

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        /// <summary>
+        /// Return a new PreFilter, without elements, which is considered
+        /// empty and always matches.
+        /// </summary>
+        public static PreFilter Empty
+        {
+            get { return new PreFilter(); }
```
jnm2: What do you think about making this into a method like `CreateEmpty()` since you're receiving an instance which is not _inherently_ empty (it can be mutated)?

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+        public bool IsEmpty
+        {
+            get { return _filters.Count == 0; }
+        }
+
+        /// <summary>
+        /// Add a new filter element to the filter
+        /// </summary>
+        /// <param name="filterText"></param>
+        public void Add(string filterText)
```
jnm2: Empty XML tag can be removed. What about the name `elementText`?

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+        public static PreFilter Empty
+        {
+            get { return new PreFilter(); }
+        }
+
+        /// <summary>
+        /// Return true if the filer is empty, in which case it
+        /// always succeeds. Technically, this is just a filter and
+        /// you can add elements but it's best to use Empty when
+        /// you need an empty filter and new when you plan to add.
```
jnm2: Typo `filer`. What if the first sentence was `Indicates whether the filter is empty.`?
Reference to Empty should be `<see>` to be refactor-safe.

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+        /// </summary>
+        /// <param name="filterText"></param>
+        public void Add(string filterText)
+        {
+            _filters.Add(new FilterElement(filterText));
+        }
+
+        /// <summary>
+        /// Use the filter on a Type, returning true if the type matches the filter
+        /// and should therefore be included in the discovery process.
```
jnm2: Capitalizing `Type` looks odd to me. <see ref="Type"/> would render as `System.Type` but it would make just as much sense to me to use the common noun `type`. Could also start `Returns true if …`.  Same with the other overload.

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+        /// </summary>
+        public bool Match(Type type)
```
jnm2: Can we do `IsMatch` since it returns a bool? Same with the other overload?

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+            {
+                if (type.FullName == filter.ClassName)
+                {
+                    ++filterCount;
+                    if (method.Name == filter.MethodName)
+                        return true;
+                }
+            }
+
+            // If no match, we succeed only if there were no method filters for this fixture
```
jnm2: If the test list contains both `Namespace.MyClass` and `Namespace.MyClass.MyMethod` for whatever reason, this don't won't find any methods in `MyClass` except `MyMethod`. Is there something that handles this at an earlier stage?

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+        private class FilterElement
+        {
+            public FilterElementType ElementType = FilterElementType.Unknown;
+            public string Text;
+            public string ClassName;
+            public string MethodName;
+
+            public FilterElement(string text)
+            {
+                int argStart = text.IndexOfAny(ARG_START);
```
jnm2: `ARG_START` looks for `<`, but `MatchFixtureElement` only works with the ECMA `` `1[`` format since you're comparing Type.FullName. Which format do we expect in existing places? Also, what about nested classes- do we expect users to type `+` or `.` for the separator?

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+        #endregion
+
+        #region Nested FilterElement Class
+
+        private class FilterElement
+        {
+            public FilterElementType ElementType = FilterElementType.Unknown;
+            public string Text;
+            public string ClassName;
+            public string MethodName;
```
jnm2: Can these all be `private` and the three strings `readonly` (and the class `sealed` 😀)?

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+                    return true;
+                }
+
+                if (MatchMethodElement(type))
+                {
+                    ElementType = FilterElementType.Method;
+                    return true;
+                }
+
+                return false;
```
jnm2: Nice strategy.

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+            }
+
+            private bool MatchFixtureElement(Type type)
+            {
+                return type.FullName == Text;
```
jnm2: What about nested fixtures? The containing class would be treated as a namespace. I'd expect the fixture with Type.FullName `Namespace.MyClass+MyNestedFixture` to be included if I specify the filter `Namespace.MyClass`.

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -0,0 +1,208 @@
...
+                Text = argStart > 0
+                    ? text.Substring(0, argStart)
+                    : text;
+
+                // Filter may be the fullname of a test method, if so extract class name
+                int lastdot = Text.LastIndexOf('.');
+                if (lastdot > 0)
+                {
+                    ClassName = Text.Substring(0, lastdot);
+                    MethodName = Text.Substring(lastdot + 1);
```
jnm2: Why calculate and store MethodName?

src/NUnitFramework/nunitlite.tests/MakeRunSettingsTests.cs
```diff
@@ -29,56 +29,31 @@ namespace NUnitLite.Tests
...
-        public void WhenLoadIsSpecified_RunSettingsIncludeIt()
-        {
-            var options = new NUnitLiteOptions("test.dll", "--prefilter=A.B.C");
-            var settings = TextRunner.MakeRunSettings(options);
-
-            Assert.That(settings.ContainsKey("LOAD"));
-            Assert.AreEqual(new string[] { "A.B.C" }, settings["LOAD"]);
-        }
+        //[Test]
+        //public void WhenTestFixtureIsSpecified_RunSettingsIncludeIt)
```
jnm2: Leftovers?

src/NUnitFramework/tests/Api/DefaultTestAssemblyBuilderTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            var result = _builder.Build(_mockAssemblyPath, settings);
+
+            Assert.That(result.IsSuite);
+            Assert.That(result, Is.TypeOf<TestAssembly>());
+            Assert.That(result.Name, Is.EqualTo(MOCK_ASSEMBLY_FILE));
+            Assert.That(result.RunState, Is.EqualTo(Interfaces.RunState.Runnable), (string)result.Properties.Get(PropertyNames.SkipReason));
+
+            return result.TestCaseCount;
+        }
+
```
jnm2: Blank line before a closing brace

src/NUnitFramework/framework/nunit.framework.csproj
CharliePoole: VS2017 15.5.1

src/NUnitFramework/nunitlite/CommandLineOptions.cs
CharliePoole: In this code, it's just a list of names, similar to `TestList`. Later all the entries get added to a `PreFilter`  the same way that the `TestList` items get added to a `TestFilter`. So maybe we should call this a `FilterList`?

src/NUnitFramework/nunitlite/TextRunner.cs
```diff
@@ -252,6 +252,17 @@ private int Execute()
 
         #region Helper Methods
 
+        private void LoadTests(IDictionary<string, object> runSettings)
+        {
+            DateTime startTime = DateTime.UtcNow;
+            long startTicks = Stopwatch.GetTimestamp();
+            _runner.Load(_testAssembly, runSettings);
+            DateTime endTime = DateTime.UtcNow;
+            double duration = (double)(Stopwatch.GetTimestamp() - startTicks) / Stopwatch.Frequency;
```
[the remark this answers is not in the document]
CharliePoole: I copied our existing logic for start and end of test. At the time I implemented it (which was quite a long time ago) there were known problems with the StopWatch instance and the static method was the only approach that worked reliably. I don't know what a new review of the class would show and I really didn't want to undertake it.

If it seems like a problem (it's the same logic we have used for ages) I can try some tests and correct the elapsed time if it comes out negative.

src/NUnitFramework/nunitlite/TextUI.cs
```diff
@@ -181,6 +181,24 @@ public void DisplayRuntimeEnvironment()
 
         #endregion
 
+        #region Test Discovery Report
+
+        public void DisplayDiscoveryReport(DateTime startTime, DateTime endTime, double duration)
```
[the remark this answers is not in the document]
CharliePoole: We use seconds everywhere else, so I think it's best if duration continues to have that meaning. We can rename to give the units if you like.

src/NUnitFramework/tests/Api/DefaultTestAssemblyBuilderTests.cs
```diff
@@ -0,0 +1,75 @@
...
+using NUnit.Framework.Internal;
+using NUnit.Tests;
+using NUnit.Tests.Assemblies;
+using NUnit.Tests.Singletons;
+
+namespace NUnit.Framework.Api
+{
+    public class DefaultTestAssemblyBuilderTests
+    {
+        private const string MOCK_ASSEMBLY_FILE = "mock-assembly.dll";
```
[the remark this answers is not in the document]
CharliePoole: IIRC all caps is in our coding standard. Has it changed? I definitely like caps at the head of a test class because it shows me what all the constant parameters of the test are at one glance.

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
CharliePoole: Well, yes. But it's required. 😸 What would you like it to say?

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
CharliePoole: It's a problem if you consider this a public API. I don't, because it's something that can only be accidentally discovered by virtue of reading pretty deeply in the framework code. Bit if you want it to be treated as an API, that's possible and we should then use ILIst.

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: Why? It's a private nested class that exists solely at the convenience of the `PreFilter` class. Only reason it's not a struct is because of the rules for struct constructors.

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: I almost always use the name of the class at the start of a class comment, probably because of the generation of documentation from XML on some past projects. But if you just think of intellisense,leaving it out makes sense.

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: Hmmm... I'm not a big fan of sealing classes as a default. I wonder if the whole thing should be internal, however.

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: As prior, I thought that caps was a standard for const and static readonly fields used as constants. Regarding the second sentence: yes, I do know that. 😈

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: Right. Although since it's a List, it gets changed all the time anyway.

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: Logically, I agree. However, we use Empty this way in two or three other classes, so I was trying for consistency. See TestFilter.Empty.

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: Good.

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: Sure.

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: Good point! Note that the user may type it in as a prefilter or as a test (which gets converted to a prefilter), so it could be either `<` or `[`. For nested classes, I think we're in luck because we use + in all cases. Finally, I would never expect the user to type in a backslash.

If you have any ideas for test cases, I'll add them.

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: So it can be used later on, in the second phase, when we decide what methods to include.

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: Try this with `--test`. I don't believe it works that way. NUnit has never treated nested classes as nested. That is, we do not consider the class in which it's nested to be a "parent" in any sense. That's why the V2 gui lists `Namespace.MyClass` and `Namespace.MyClass+MyNestedFixture` as entirely separate entries. 

We've had a rather big feature in our list for years to __start__ treating nested classes in this way, but we've never taken it on.

OTOH If I'm wrong and `--test` works the way you say, then we should make `--prefilter` work that way as well.

src/NUnitFramework/nunitlite.tests/MakeRunSettingsTests.cs
CharliePoole: Yup!

src/NUnitFramework/nunitlite/TextRunner.cs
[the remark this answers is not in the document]
jnm2: Since it's copy and paste, could we move it to a common helper method? That will also resolve what was bothering me: that this doesn't say what it does, it just does it in a complex way.
I doubt that a test will get it to come out negative unless the system performance counters report in extremely high accuracy and the machine has been running for an unusual amount of time.

It's probably too long ago, but if you recall what the unreliability was, I love to know about that kind of thing.

src/NUnitFramework/nunitlite/TextUI.cs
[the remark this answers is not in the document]
jnm2: Yes please!

src/NUnitFramework/tests/Api/DefaultTestAssemblyBuilderTests.cs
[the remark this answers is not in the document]
jnm2: Just checked to confirm; the coding standard hasn't mentioned constants as far back as the current file goes. I believe we want to be more permissive of preferences now, too, so I'm happy with keeping it.

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
jnm2: It looks like it should not be required because it's on a private field. ❔

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
jnm2: I'm not the most familiar, so just confirming this so I can keep up: you don't expect the constant `FrameworkPackageSettings.LOAD` to be used by third parties integrating with NUnit in such a way that their values could end up here?

src/NUnitFramework/framework/Interfaces/IFixtureBuilder.cs
jnm2: I read something (who knows where) and have avoided suffixing `2` or `Ex` (which winapi does) ever since. Perhaps `IFixtureBuilder2` says, "this would be `IFixtureBuilder` if there was any way to do that. Perhaps `IFilterableFixtureBuilder` says, "this is why you would use this kind of `IFixtureBuilder`."
Not sure which of the two we things we want to be saying.

src/NUnitFramework/framework/Internal/PreFilter.cs
jnm2: `internal` makes me feel best of all. 😇

src/NUnitFramework/framework/Internal/PreFilter.cs
jnm2: It's nice to lock it down so that you know that if you take a reference to the current list instance in the field, you won't miss any changes. Otherwise there are two ways it could change and the combination is more complex to consider.

src/NUnitFramework/framework/Internal/PreFilter.cs
jnm2: That's reasonable, even more so if the class is internal. 😇

src/NUnitFramework/framework/Internal/PreFilter.cs
jnm2: @CharliePoole Because all those things make reading comprehension faster by cutting down on what-ifs that must be mentally tracked.
I thought about suggesting `struct`, but the problem is you'd have to copy it back into the `_filters` list each time it's used since it's mutable or else `ElementType` will just stay 0 the whole time.

jnm2: @CharliePoole Because all those things make reading comprehension faster by cutting down on what-ifs that must be mentally tracked. Scanning the mutability of the fields is for me lot like scanning the UPPER_CASE constant parameters is for you.

I thought about suggesting `struct`, but the problem is you'd have to copy it back into the `_filters` list each time it's used since it's mutable or else `ElementType` will just stay 0 the whole time.

src/NUnitFramework/framework/nunit.framework.csproj
jnm2: That's perplexing. What was the startup project/executable at the time?

src/NUnitFramework/framework/Internal/PreFilter.cs
jnm2: You're correct; NUnit does not treat containing types as namespaces.

src/NUnitFramework/nunitlite/TextRunner.cs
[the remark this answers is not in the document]
CharliePoole: A common method makes sense.

IIRC there were two basic problems: (1) when running on multiple processors, the same processor has to be used for starting and stopping the stopwatch, depending on the BIOS implementation and (2) when there is no hires counter, the system time is used internally, which is a problem because the time can be changed, perhaps even by the test we are running (apparently, this was a "won't fix" bug report to Microsoft in 2012).

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
CharliePoole: I don't. However, in an earlier discussion on the issue itself, both you and @ChrisMaddock asked me to keep the existing `LOAD` setting to avoid even the small risk of breaking anything. I suppose keeping `IList` would be consistent with that approach.

src/NUnitFramework/framework/nunit.framework.csproj
CharliePoole: Either the nunitlite test exe or the nunitlite runner - IIRC both had the problem.

CharliePoole:
@jnm2 Seems like I lost track of this PR and your latest comments. I'll go through them and update as needed. Then we can decide whether to abandon it per comments you made on the issue.

jnm2:
It's really confusing to me that you're bringing up abandoning the PR, and I replied to your other similar comment on the discussion issue: https://github.com/nunit/nunit/issues/2876#issuecomment-401411733

CharliePoole:
You made comments about what I see as an unrelated issue, which would preclude merging this PR. However, the PR is out of date WRT your earlier comments - my fault. Therefore, I thought I should bring it up to date and then you could decide what to do with it.

Note: abandoning a PR is not a big deal. Abandoning the actual work is bigger, but even that's not the end of the world. Code is generally disposable if it's going in a direction you don't want to go.

Let's discuss on the issue.

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: Has to be public I'm afraid, as it's used in the interface.

src/NUnitFramework/framework/Internal/PreFilter.cs
CharliePoole: I think you're right. I'll check into it further.

src/NUnitFramework/framework/Internal/PreFilter.cs
jnm2: Oh... we're trying to go the other direction with referencing internal types from our API. This code would require folks to open our internal namespace in order to use the APi. 

Seems like a great opportunity to decide how to solve this. Is that something you're up for?

CharliePoole: I'll finish up with fixing the older comments and then come back to this. I think we have to start with the interfaces themselves, make sure they are properly public interfaces and then decide how to handle the filter argument.

src/NUnitFramework/nunitlite/TextRunner.cs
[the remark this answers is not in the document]
CharliePoole: Hmm... So this is in nunitlite. The other uses are in nunit.framework. We can have a utility method in the framework, which is called by nunitlite. Are we OK with that? Or should we work toward separating nunitlite from relying on helper methods within the framework? I can see doing it either way, since we rebuild nunitlite for each version of the framework. OTOH, if we want to aim toward a more independent nuntlite, we would have to change this at a later time, along with other places where we do something similar.

src/NUnitFramework/framework/Interfaces/IFixtureBuilder.cs
jnm2: @nunit/framework-team ping?

CharliePoole:
@jnm2 This is ready for re-review

src/NUnitFramework/tests/Api/DefaultTestAssemblyBuilderTests.cs
```diff
@@ -24,32 +24,13 @@
...
 using NUnit.Framework.Internal;
 using NUnit.Tests;
 using NUnit.Tests.Assemblies;
 using NUnit.Tests.Singletons;
-using NUnit.Compatibility;
-using NUnit.TestData;
 
 namespace NUnit.Framework.Api
 {
-    [TestFixture] // [TestFixture] is needed so that ClassesCanBeImpliedToBeFixtures is effective.
```
jnm2: @CharliePoole I think it's important to preserve this comment on whatever class contains `ClassesCanBeImpliedToBeFixtures`.

src/NUnitFramework/tests/Internal/DefaultSuiteBuilderTests.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace NUnit.Framework.Internal
+{
+    [TestFixture]
+    public static class DefaultSuiteBuilderTests
+    {
+        [Test]
+        public static void ClassesCanBeImpliedToBeFixtures()
+        {
+            Assert.That(new Internal.Builders.DefaultSuiteBuilder().CanBuildFrom(typeof(ImpliedFixture)));
```
jnm2: This tells us that it *says* it can build a fixture, but can we does what it actually *does*? I.e, that the test suite it returns contains a fixture for the type `DefaultSuiteBuilderTests`?

jnm2:
Pulling this from https://github.com/nunit/nunit/pull/2878#discussion_r194238150 since GitHub is hiding that section by default:

> > > <sub>(@jnm2)</sub>
> > > What do you think of `IFilterableFixtureBuilder`?
> >
> > <sub>(@CharliePoole)</sub>
> > That works, for sure. `IFixtureBuilder2` follows a long tradition of expanding interfaces going back to COM days. I'd like to know what other @nunit/framework-team members think.
>
> <sub>(@jnm2)</sub>
> I read something (who knows where) and have avoided suffixing `2` or `Ex` (which winapi does) ever since. Perhaps `IFixtureBuilder2` says, "this would be `IFixtureBuilder` if there was any way to do that. Perhaps `IFilterableFixtureBuilder` says, "this is why you would use this kind of `IFixtureBuilder`."
Not sure which of the two we things we want to be saying.

I'm feeling hesitant about introducing `IFixtureBuilder2` to our public interfaces and we'd really appreciate hearing from @nunit/framework-team.

src/NUnitFramework/tests/Api/DefaultTestAssemblyBuilderTests.cs
CharliePoole: I'll add it if you like. However, I believe we have other of tests that would not work correctly without specifying `[TestFixture]`. So far, we have just done it without comment, just as we avoid using any of the features we are testing in the tests themselves.

src/NUnitFramework/tests/Internal/DefaultSuiteBuilderTests.cs
CharliePoole: I usually try to write the smallest possible micro-test for the thing I'm testing. IImplyFixture is only used within the `CanBuildFrom` method. The `BuildFrom` method does the same thing no matter how the fixture is identified.

Adding a call to `BuildFrom` would reassure you but would also be a test of something else that was not impacted by this issue. 

I think your comment reflects the part that we don't have any other tests of DefaultSuiteBuilder, and you'd like to see more. I could do that as a part of this PR even though it goes against my "smallest possible diff" initiative. 😄 

Your comment and my approach __may__ also reflect a different view of the nature of these tests: black box versus white box. Since I __know__ that the implementation only makes use of the interface in the `CanBuild` method, I can write a micro-test for that alone. I assume (blindly and wrongly) that some functional test will catch less localized errors. IOW, so long as we don't have separate functional and integration tests, there is always a temptation to write unit tests that test a few things at once.

That said, I'm glad to write a few more tests if you like!

CharliePoole:
Regarding interface naming...

If I see `IFixtureBuilder` / `IFilterableFixtureBuilder` I tend to say "OK, it's optional, I can choose to support filtering or not. Some custom attributes will then be written to not support filters and we will get issues that ask why NUnit won't filter in such and such a case.

If I see `IFixtureBuilder` / `IFixtureBuilder2` my reaction is "I see there is an old implementation and a newer one. I should use the newer one. I will implement the newer one and filtering will work. 

There is also the old convention of `IFilterBuilderEx`, which falls somewhere between the other two. I don't like it, since there may be a `3` some day.

I think I would react differently if the `IFixtureBuilder2` implementation had to do any work to make filtering work. All it does is pass the argument along, however. I tried to figure out a design where the filter would be saved somewhere globally, but it ended up quite messy so I abandoned it.

IIRC, we used to have a number of `2` interfaces in V2. I can't find any others in the current framework, so we're not bound by any precedent here. If we decide to use a new name, then I suggest we deprecate the old one.

An alternative, of course, is to just add the method to the old interface and use reflection to test for its presence. I avoided that in this PR because testing interfaces is both easier and what interfaces are for. 😸

src/NUnitFramework/tests/Api/DefaultTestAssemblyBuilderTests.cs
jnm2: If this is generally known by the team (I missed it) then it makes sense not to mention it. The cost of adding it doesn't seem great, unless it would further cause us to assume (as I already did) that all other `[TestFixture]` attributes could be removed.

src/NUnitFramework/tests/Internal/DefaultSuiteBuilderTests.cs
jnm2: Sure, splitting it into smaller microtests seems even better to me! I would like that, thank you!

jnm2:
You make good points.

If IFilterableFixtureBuilder inherits from the other one, and we deprecate the other one, we won't have the issue of folks thinking it's optional. Once you deprecate though, it does make you wonder what the third iteration of this interface would be named. So I see the benefit of IFixtureBuilder2. This is all so messy.

And, sadly, adding the method to the old interface would cause the runtime to refuse to load assemblies that had been compiled against NUnit 3.10 if NUnit 3.11 is already loaded in its place. The binary breaking change in action.

As an aside, this is why many libraries opt for abstract classes over interfaces because there, adding methods is not a breaking change. A new runtime and language feature called Default Interface Methods is planed for C# 8 where you can add a method to an interface and provide a default implementation to address this specific issue of having no way to evolve interfaces over time.

oznetmaster:
I prefer the IFixtureBuilder2 naming, at least until C# 8.

In my CF build, I have introduced a number of I...2 interfaces for exactly the reasons @CharliePoole discussed.

src/NUnitFramework/nunitlite.tests/MakeRunSettingsTests.cs
jnm2: @CharliePoole Oops, @oznetmaster [noticed](https://github.com/nunit/nunit/commit/d09def3521dca9be6fbd84fc483c555c4955afe3#r29649322) that the lines made it into master!

CharliePoole: Yes... I responded. I guess they can wait till somebody is in this file again.

oznetmaster:
I just want to say how much this speeds things up.  For example, running a single one of the NUnit tests on my CF build on a very slow ARM processor previously took about 1 minute.  Now it takes about 5 seconds!

This is a great addition to NUnitLite!

CharliePoole:
Cool! I'm hoping someone will make it available in the other runners.

jnm2:
@CharliePoole I'm typing up an issue as we speak. :D

ChrisMaddock:
I always click the default option. Disabling sounds good to me.

CharliePoole:
Me too. At some point, the default changed to Squash and I assumed that was a team decision that  I had missed.

mikkelbu:
Sounds good to me.

PR #2880, opened 2018-05-31
Assert.That with a Throws constraint provide more information

### LINKED ISSUE #2146: Assert.That with a Throws constraint does not provide as much info as Assert.Throws

ZenonBuratta:
Fix issue by checking the type of object in TypeConstraint.ApplyTo, if it is an exception, set the actual value to the exception object instead of the type.
Fix broken unit tests and add a couple of extra tests

Fixes #2146

### DIFF
--- a/src/NUnitFramework/framework/Constraints/TypeConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/TypeConstraint.cs
@@ -61,6 +61,10 @@ namespace NUnit.Framework.Constraints
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
             actualType = actual == null ? null : actual.GetType();
+
+            if (actual is Exception)
+                return new ConstraintResult(this, actual, this.Matches(actual));
+
             return new ConstraintResult(this, actualType, this.Matches(actual));
         }
 
--- a/src/NUnitFramework/tests/Assertions/TypeAssertTest.cs
+++ b/src/NUnitFramework/tests/Assertions/TypeAssertTest.cs
@@ -76,8 +76,8 @@ namespace NUnit.Framework.Assertions
         public void IsNotInstanceOfFails()
         {
             var expectedMessage =
-                "  Expected: not instance of <System.Exception>" + System.Environment.NewLine + 
-                "  But was:  <System.ArgumentException>" + System.Environment.NewLine;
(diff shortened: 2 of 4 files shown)


mikkelbu:
Thanks for the PR @ZenonBuratta. The contribution is much appreciated 👍 .

ZenonBuratta:
@mikkelbu Thanks, time to find another :)

PR #2881, opened 2018-05-31
Fix for #2837 DictionaryContainsKeyConstraint behaviour ignores the dictionary key comparer.

### LINKED ISSUE #2837: DictionaryContainsKeyConstraint behaviour is inconstant with Dictionary.ContainsKey when the dictionary uses a custom Comparer
If you have a dictionary that was created with a custom comparer then the DictionaryContainsKeyConstraint does not work as expected.

```
var dictionary = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase) { { "Hello", "World" }, { "Hola", "Mundo" } };
Assert.That(dictionary, Does.ContainKey("hello"));
```

DaiMichael:
fix #2837
Now use the underlying dictionary comparer.
Deprecated the IgnoreCase and Using() methods but use old method to maintain backwards compatibility.
Added unit tests.


DaiMichael:
Hmm... so the build is failing as I `Deprecated` 2 calls. I have left the unit tests in place as it uses the older method for backwards compatibility. But... the tests fail on a `Deprecated` function usage? So so I turn delete the tests of will making them 'Explicit' work ?

CharliePoole:
If you are obsoleting something, you have to obsolete anything that calls it, including test methods.

I assume there has some prior discussion leading to deprecating these features, because we don't do that lightly. But even if we deprecate something, we are still supporting it so we have to have tests.

DaiMichael:
@CharliePoole Thanks for letting me know about marking the tests as well, I have put the tests back in now.

Yes there was a discussion with @jnm2 and @ChrisMaddock about deprecating the methods but they are still support with backwards compatibility.

jnm2:
> If you are obsoleting something, you have to obsolete anything that calls it, including test methods.

Nice, I didn't even know that worked! I've always wrapped with `#pragma warning disable`/`restore CS0612`.

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -65,32 +71,138 @@ public override string Description
         /// </summary>
         protected object Expected { get; }
 
+        /// <summary>
+        /// Flag the constraint to ignore case and return self.
+        /// </summary>
+        [Obsolete("Deprecated, use Does.ContainKey")]
```
jnm2: Given that they have just typed `Does.ContainKey(...).IgnoreCase`, they're going to think, "I already am using `Does.ContainKey`" Could we be more specific about what they need to do if they want to do an case-insensitive key assert, e.g. `Has.Property("Keys").With.One.EqualTo(...).IgnoreCase`?

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -33,6 +37,8 @@ namespace NUnit.Framework.Constraints
     /// </summary>
     public class DictionaryContainsKeyConstraint : CollectionItemsEqualConstraint
     {
+        private bool _deprecatedFlag = false;
```
jnm2: I'd rather put 'is' in the name of a bool property than 'flag.' `_isLegacyMode` or `_isDeprecatedMode`?

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -65,32 +71,138 @@ public override string Description
...
-            return false;
+            throw new ArgumentException("Not a collection supporting ContainsKey method.");
         }
 
         /// <summary>
         /// Flag the constraint to use the supplied predicate function
         /// </summary>
         /// <param name="comparison">The comparison function to use.</param>
         /// <returns>Self.</returns>
+        [Obsolete("Deprecated, use Does.ContainKey")]
```
jnm2: Same as https://github.com/nunit/nunit/pull/2881/files#r192375027

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -65,32 +71,138 @@ public override string Description
...
 
+            _deprecatedFlag = true;
             base.Using(EqualityAdapter.For(invertedComparison));
             return this;
         }
+
+        /// <summary>
+        /// Flag the constraint to use the supplied Comparison object.
+        /// </summary>
+        /// <param name="comparer">The IComparer object to use.</param>
```
jnm2: > The IComparer object to use.

> Comparison<T> comparer

This is a problem 😆 was the base class also wrong here?

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -65,32 +71,138 @@ public override string Description
...
+        {
+            _deprecatedFlag = true;
+            return base.Using(comparer);
+        }
+
+        /// <summary>
+        /// Checks if the key is contained in a "keyed item container".
+        /// </summary>
+        /// <param name="actual">Keyed container.</param>
+        /// <returns>method to call.</returns>
```
jnm2: We [recommend](https://github.com/nunit/docs/wiki/Best-practices-for-XML-documentation#returns-and-value) that you delete the IDE's auto-inserted `<return>` elements because they don't show up in intellisense and we don't do anything else with the XML docs. (If we did, I'd ask for the word 'The' to precede it.)

Also, I think the best option is to rename the parameter `actual` to `keyedItemContainer` and [delete the `<param>`](https://github.com/nunit/docs/wiki/Best-practices-for-XML-documentation#param-and-typeparam).

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
jnm2: > The IComparer object to use.

👀

> `Comparison<T> comparer`

This is a problem 😆
Was the base class also wrong here?

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
jnm2: Given that they have just typed `Does.ContainKey(...).IgnoreCase`, they're going to think, "I already am using `Does.ContainKey`." Could we be more specific about what they need to do if they want to do an case-insensitive key assert, e.g. `Has.Property("Keys").With.One.EqualTo(...).IgnoreCase`?

It would read better to replace the comma with either a semicolon or a period followed by starting a new sentence.

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
jnm2: Same comments as the other Obsolete attribute.

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -65,32 +71,138 @@ public override string Description
...
+                            .FirstOrDefault(m => m != null);
+
+            return method;
+        }
+
+        /// <summary>
+        /// Looks for a base type that implements ContainsKey method
+        /// </summary>
+        /// <param name="type">Type to look for method ContainsKey</param>
+        /// <returns>Returns the method to call.</returns>
```
jnm2: Before I nitpick at this one too, I should point out that leaving the private methods without XML docs is also a good option.

The summary makes it look like it only searches base types, but it also searches the most derived type as well as all implemented interfaces. Then you'd say, "... that implements **a** \<c>ContainsKey\</c> method." (Including the indefinite article, the `<c>`, and the end mark.

The `<param>` has some awkward wording. You can say, "The type to search for X," because it makes sense to say, "Search the type for X." You can't say, "The type to look for X," because it doesn't make sense to say, "Look the type for X." 'Look' is an [intransitive](https://en.wikipedia.org/wiki/Intransitive_verb) verb.
Better would be, "The type which will be searched for a \<c>ContainsKey\</c> method." That way you're also describing the effect the parameter has, not just what the parameter *is*.

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -65,32 +71,138 @@ public override string Description
...
+
+                if (tKeyGenericArg != null)
+                {
+                    method = definition
+                             .GetMethods(BindingFlags.Instance | BindingFlags.Public)
+                             .FirstOrDefault(m =>
+                                 m.ReturnType == typeof(bool)
+                                 && m.Name == "Contains"
+                                 && m.GetParameters().Any()
+                                 && m.GetParameters().First()?.ParameterType == tKeyGenericArg);
```
jnm2: `m.GetParameters().First()` can't return null, so `?.` is out of place there. However, these two lines could be merged to `m.GetParameters().FirstOrDefault()?.ParameterType ==`.

jnm2:
@DaiMichael Would you please [edit "Fixes #2837"](https://help.github.com/articles/closing-issues-using-keywords/) into the PR comment at the top (not the title)?

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -65,32 +71,138 @@ public override string Description
...
+            }
+
+            return method;
+        }
+
+        /// <summary>
+        /// Returns all the base types of the class
+        /// </summary>
+        /// <param name="type">Type to search for base types implemeted.</param>
+        /// <returns>Base types / interfaces implemented by the class</returns>
```
jnm2: This only returns base types, not implemented interfaces. I'd encourage the removal of the `<returns>`.

For the `<param>`: 'implemented' is the term for interfaces and abstract base classes. The general term for base types is 'inherited.' Here I'd say, "The type whose base types will be listed."

The `<summary>` should have an end mark.

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -65,32 +71,138 @@ public override string Description
...
 
-            foreach (object obj in dictionary.Keys)
-                if (ItemsEqual(obj, Expected))
-                    return true;
+            var method = GetContainsKeyMethod(actual);
+            if (method != null)
+                return (bool)method.Invoke(actual, new[] { Expected });
 
-            return false;
+            throw new ArgumentException("Not a collection supporting ContainsKey method.");
```
jnm2: Needs an indefinite article before 'ContainsKey method.' I don't want to use the word 'collection' because that communicates that we are checking whether it's a collection, and we're not actually doing that.
How about, "The actual value must have a ContainsKey or Contains(TKey) method."?

Also, it's missing `nameof(actual)`.

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -84,35 +85,521 @@ public void WorksWithNonGenericDictionary()
...
+
+        [Test]
         public void SucceedsWhenKeyIsPresentUsingContainKeyWhenDictionaryUsingCustomComparer()
         {
             var dictionary = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase) { { "Hello", "World" }, { "Hola", "Mundo" } };
             Assert.That(dictionary, Does.ContainKey("hola"));
         }
+
+        [Test]
+        public void SucceedsWhenKeyIsPresentUsingContainKeyWhenUsingLookupCustomComparer()
```
jnm2: Typo, `ContainKey` should be `ContainsKey` (6 places)

These are really excellent tests, by the way!

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -84,35 +85,521 @@ public void WorksWithNonGenericDictionary()
...
+            Assert.That(lookup, Does.ContainKey("catherine"));
+        }
+
+        [Test]
+        public void SucceedsWhenKeyIsNotPresentUsingContainKeyUsingLookupDefaultComparer()
+        {
+            var list = new List<string> { "ALICE", "BOB", "CATHERINE" };
+            ILookup<string, string> lookup = list.ToLookup(x => x);
+
+            Assert.That(lookup, !Does.ContainKey("alice"));
```
jnm2: I did not know we declared a custom `!` operator! 😆

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -84,35 +85,521 @@ public void WorksWithNonGenericDictionary()
...
+        [Test]
+        public void SucceedsWhenKeyIsNotPresentUsingContainKeyUsingKeyedCollectionDefaultComparer()
+        {
+            var list = new TestKeyedCollection {"ALICE", "BOB", "CALUM"};
+            
+            Assert.That(list, !Does.ContainKey("alice"));
+        }
+
+        [Test]
+        public void SucceedsWhenKeyIsPresentUsingContainKeyUsingHastableCustomComparer()
```
jnm2: Typo `Hastable` (2 places)

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -84,35 +85,521 @@ public void WorksWithNonGenericDictionary()
...
+        {
+            var table = new Hashtable(StringComparer.OrdinalIgnoreCase) { { "ALICE", "BOB" }, { "CALUM", "DENNIS" } };
+
+            Assert.That(table, Does.ContainKey("alice"));
+        }
+
+        [Test]
+        public void SucceedsWhenKeyIsPresentUsingContainKeyUsingHastableDefaultComparer()
+        {
+            var table = new Hashtable {{ "ALICE", "BOB"}, {"CALUM", "DENNIS"} };
```
jnm2: Inconsistent whitespace `{{`

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -84,35 +85,521 @@ public void WorksWithNonGenericDictionary()
...
+
+            Assert.That(table, !Does.ContainKey("calum"));
+        }
+
+        [Test]
+        public void ShouldCallContainsKeysMethodWithTKeyParameterOnNewMethod()
```
jnm2: `ContainsKeys` should be `ContainsKey` (7 places)

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -84,35 +85,521 @@ public void WorksWithNonGenericDictionary()
...
+            Assert.That(dictionary, Does.ContainKey(20));
+            Assert.That(dictionary, !Does.ContainKey(10));
+        }
+
+        [Test]
+        public void ShouldCallContainsKeysMethodOnPlainDictionary()
+        {
+            var dictionary = new TestPlainDictionary(99);
+
+            Assert.Catch<ArgumentException>(() => Assert.That(dictionary, Does.ContainKey(99)));
```
jnm2: ℹ Another option would be to use constraint syntax:
```cs
Assert.That(() => Assert.That(dictionary, Does.ContainKey(99)), Throws.ArgumentException);
```

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -84,35 +85,521 @@ public void WorksWithNonGenericDictionary()
...
+        }
+
+        [Test]
+        public void ShouldCallContainsKeysMethodOnObject()
+        {
+            var poco = new TestPlainObjectContainsKey("David");
+
+            Assert.Catch<ArgumentException>(() => Assert.That(poco, Does.ContainKey("David")));
```
jnm2: Why is this throwing an ArgumentException? This one was supposed to work, I think.

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -84,35 +85,521 @@ public void WorksWithNonGenericDictionary()
...
+        }
+
+        [Test]
+        public void ShouldCallContainsMethodOnObject()
+        {
+            var poco = new TestPlainObjectContains("Peter");
+
+            Assert.Catch<ArgumentException>(() => Assert.That(poco, Does.ContainKey("Peter")));
```
jnm2: We should have a test on a `TestPlainObjectContains<TKey>` class where this succeeds, too.

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -84,35 +85,521 @@ public void WorksWithNonGenericDictionary()
...
+        public void ShouldCallContainsKeysMethodOnReadOnlyInterface()
+        {
+            var dictionary = new TestReadOnlyDictionary("BOB");
+
+            Assert.That(dictionary, Does.ContainKey("BOB"));
+            Assert.That(dictionary, !Does.ContainKey("ALICE"));
+        }
+
+        [Test]
+        public void ShouldCallContainsKeysMethodOnSet()
```
jnm2: `ShouldThrowWhenUsedWithISet`

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -84,35 +85,521 @@ public void WorksWithNonGenericDictionary()
...
+        [Test]
+        public void ShouldCallContainsKeysMethodOnObject()
+        {
+            var poco = new TestPlainObjectContainsKey("David");
+
+            Assert.Catch<ArgumentException>(() => Assert.That(poco, Does.ContainKey("David")));
+        }
+
+        [Test]
+        public void ShouldCallContainsMethodOnObject()
```
jnm2: `ShouldThrowWhenUsedOnObjectWithoutProperMethod`?

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -84,35 +85,521 @@ public void WorksWithNonGenericDictionary()
...
+            {
+                get { throw new NotImplementedException(); }
+                set { throw new NotImplementedException(); }
+            }
+
+            public ICollection<int> Keys { get; }
+            public ICollection<string> Values { get; }
+        }
+
+        public class TestPlainDictionary : IDictionary
```
jnm2: `Plain` should be `NonGeneric`

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
ChrisMaddock: Wow, me neither! 😄

ChrisMaddock:
One thought on the messages in the `[Obsolete]` attributes - I wonder if a significant use case would be people using these methods as they've encountered the bug we're trying to fix, and used this as a workaround.

e.g. Initially they'd have tried:

```
var dict = new Dictionary<string,int>(new CaseInsensitiveComparer())
dict["Key"] = 0;
Assert.That(dict, Contains.Key("key"));
```

They'd have then found that didn't work, so switched to:

```
Assert.That(dict, Contains.Key("key").Using(new CaseInsensitiveComparer()));
```

So, upgrading to include this fix, they'd get an `Obsolete` warning - but the fix they actually need is just to remove the `Using()` entirely!

Based on that, is it possible to include in these obsoletion methods some details of what the change made actually is? How about something like this...

> DictionaryContainsKeyConstraint now uses the custom comparer of the dictionary under test. To test with a alternative comparer, consider using a collection constraint on the set of keys.

I'm sure @jnm2 will be able to refine that message into something more intelligible! 😄

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -65,32 +71,138 @@ public override string Description
...
+                var definition = type.GetGenericTypeDefinition();
+                var tKeyGenericArg = definition.GetGenericArguments().FirstOrDefault(typeArg => typeArg.Name == "TKey");
+
+                if (tKeyGenericArg != null)
+                {
+                    method = definition
+                             .GetMethods(BindingFlags.Instance | BindingFlags.Public)
+                             .FirstOrDefault(m =>
+                                 m.ReturnType == typeof(bool)
+                                 && m.Name == "Contains"
```
ChrisMaddock: Why's this just looking for `Contains` and not `ContainsKey` sorry? I presume there's a case I'm not thinking of!

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
ChrisMaddock: > These are really excellent tests, by the way!

Completely agree! 😄

DaiMichael:
@jnm2 @ChrisMaddock 

yikes... I really need to read the doc/xml better before commits... I'm just not used to it, I never expose my code as an API in work, we are rather fast and loose in my industry, bugs are challenges for the user to work around :wink: (and that is pretty much true! as long as it doesn't cost them money).

On the Duck Typing, I did consider making into a seperate helper where it could be used else where, but did first principle, i.e. it is the only case at the moment so leave like that and will extract if we need else where at a later date. I'd need to write a bunch of code and tests which might not ever be needed. 

On shadow functions (new-ing), my only concern at the moment, is I've only done it to the ones we test, there are others in the base class that couuld be called and they would fail using the new code path. 

@jnm2 - Like the 
> Before I nitpick at this one too
 after about "20" change suggestions 🤣 

Will crack on with all above, it reads like a good novel !

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
DaiMichael: >Was the base class also wrong here?

Ctrl C and Ctrl V at its finest, I can change in base class.

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
jnm2: `ContainsKey` would have already been found by the previous section of code. We need to pick up *some* `Contains` methods as well for the same of classes like https://docs.microsoft.com/dotnet/api/system.collections.objectmodel.keyedcollection-2.contains.

jnm2:
@ChrisMaddock This is entirely intelligible!

> DictionaryContainsKeyConstraint now uses the comparer which the dictionary is based on. To test using a comparer which the dictionary is not based on, use a collection constraint on the set of keys.

@DaiMichael 

> On the Duck Typing, I did consider making into a seperate helper where it could be used else where, but did first principle, i.e. it is the only case at the moment so leave like that and will extract if we need else where at a later date. I'd need to write a bunch of code and tests which might not ever be needed.

I agree with this.

> On shadow functions (new-ing), my only concern at the moment, is I've only done it to the ones we test, there are others in the base class that could be called and they would fail using the new code path by default.

Which methods are those?

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -65,32 +71,138 @@ public override string Description
...
+            if (actual == null) throw new ArgumentNullException(nameof(actual));
+            var instanceType = actual.GetType();
+
+            var method = FindContainsKeyMethod(instanceType)
+                         ?? instanceType
+                            .GetInterfaces()
+                            .Concat(GetBaseTypes(instanceType))
+                            .Select(FindContainsKeyMethod)
+                            .FirstOrDefault(m => m != null);
+
```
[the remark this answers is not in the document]
jnm2: Thanks for bringing that to my attention. I'm really confused why [this method](https://github.com/nunit/nunit/blob/46b62b6dc4087efd0b4758ea039f3e07e7a7b911/src/NUnitFramework/framework/Internal/TypeHelper.cs#L316-L338) exists because Type.GetInterfaces returns all interfaces implemented by the type, which by definition includes all interfaces implemented by all base types. For example, try this in LINQPad:


<details>

```cs
void Main()
{
    typeof(B).GetInterfaces().Dump();
}

public class A : IEquatable<A>
{
    public bool Equals(A other)
    {
    }
}

public class B : A, IComparable<B>
{
    public int CompareTo(B other)
    {
    }
}
```

</details>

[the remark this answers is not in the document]
jnm2: I was confused at first, but `GetDeclaredInterfaces` only returns interfaces implemented by the most-derived type and intentionally excludes all interfaces implemented by base types. We actually want `Type.GetInterfaces()` here so that we catch all interface implementations.

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
DaiMichael: This is correct functionality... although we could extend the test scenario (read first and decide).

Basically in `CollectionConstraint.ApplyTo<>(...)` we call a method `ConstraintUtils.RequiresActual<IEnmerable>(...)` This throws the `ArgumentException` as our `TestNonGenericObjectContrainsKey` (new name not PlainObject as before) doesn't implement this interface. This is meant to be a "Collection" of some type, so it seems perfectly valid to throw.

We could make this `TestNonGenericObjectContrainsKey` impl IEnumerable and then obviously we should be good? Could create a new test with new object inherits this class and which has IEnumerable. I think both seem valid?

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
DaiMichael: > 

This is correct functionality... although we could extend the test scenario (read first and decide).

Basically in CollectionConstraint.ApplyTo<>(...) we call a method ConstraintUtils.RequiresActual<IEnmerable>(...) This throws the ArgumentException as our TestNonGenericObjectContrainsKey (new name not PlainObject as before) doesn't implement this interface. This is meant to be a "Collection" of some type, so it seems perfectly valid to throw.

We could make this TestNonGenericObjectContrainsKey impl IEnumerable and then obviously we should be good? Could create a new test with new object inherits this class and which has IEnumerable. I think both seem valid?

DaiMichael:
@jnm2 > Which methods are those?

These are in `CollectionItemsEqualConstraint` 

#### Shadowed (new-ed) and marked `[Obselete]`
```
CollectionItemsEqualConstraint` IgnoreCase
CollectionItemsEqualConstraint Using<T>(Comparison<T> comparison)
```

#### Not shadowed (new-ed) and `public` with no tests for `DictionaryContainsKeyConstraintTests`
```
CollectionItemsEqualConstraint Using(IComparer comparer)
CollectionItemsEqualConstraint Using<T>(IComparer<T> comparer)
CollectionItemsEqualConstraint Using(IEqualityComparer comparer)
CollectionItemsEqualConstraint Using<T>(IEqualityComparer<T> comparer)
CollectionItemsEqualConstraint Using<T>(Func<T, T, bool> comparer)
```

I accept that I could be missing something here....... 😕

jnm2:
Nope, you're not missing anything. We need to shadow all of them. :-(

jnm2:
@DaiMichael 

> Fix for #2837

This doesn't actually work. See https://help.github.com/articles/closing-issues-using-keywords/?

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
jnm2: @DaiMichael I had listed this scenario in https://github.com/nunit/nunit/issues/2837#issuecomment-390790330, so I expected it to work. You raise an interesting point about the RequiresActual. I'll ask in the discussion thread.

DaiMichael:
@jnm2 

> Nope, you're not missing anything. We need to shadow all of them. :-(

This why I mentioned way back when making them virtual in the base class, I didn't really explain why, my bad..  I'll need to add tests for each one as well, even though they aren't there at the moment, which is fine, keeps me off the streets 🚔

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
```diff
@@ -109,11 +109,11 @@ public CollectionItemsEqualConstraint Using<T>(IComparer<T> comparer)
         /// <summary>
         /// Flag the constraint to use the supplied Comparison object.
         /// </summary>
-        /// <param name="comparer">The IComparer object to use.</param>
+        /// <param name="comparison">The IComparer object to use.</param>
```
jnm2: Can you replace `IComparer` with `<see cref="Comparison{T}" />` here?

DaiMichael:
Went for the change to allow any object with `Contains(TKey)` or `ContainsKey`. Seems reasonable to what we have been doing so far.

Bunch of other minor fixes for typo and clarity stuff.
Added / Updated Tests.

Glad when I picked this up, I thought "Should be straight forward with no major changes"  :laughing:

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -106,6 +106,8 @@ protected override bool Matches(IEnumerable actual)
             throw new ArgumentException("Not a collection supporting ContainsKey method.");
         }
 
+        #region Shadow CollectionItemsEqualConstraint Methods
```
jnm2: Technically they are 'shadowing methods,' not 'shadow methods.'

DaiMichael:
@jnm2 This is down as awaiting-contributor but should it not be awaiting-review ?

jnm2:
@DaiMichael Dear me, thanks.

@nunit/framework-team We gave the labels a try, what do you think overall? I'm noticing a number of times where they end up incorrect because there's nothing automating them.

ChrisMaddock:
I think the labels are great - so much so, I just added them to nunit-console this morning. 😄 

Bit more than I can look through right now on this PR, but I'll try to get back to it soon!

ChrisMaddock:
This looks good - thanks for sticking with it, @DaiMichael. Sorry it's taken a while to get merged - hope it hasn't put you off further contributions! 😄

DaiMichael:
@ChrisMaddock - I can take a look at this, I suspect this will have broken my current issue (#1638) fix on my branch as well. My github skills need honing anyway.

ChrisMaddock:
Brilliant - thanks @DaiMichael!

DaiMichael:
New PR on #2967

PR #2882, opened 2018-06-04
Provide better error message for CollectionOrderedConstraint

### LINKED ISSUE #2855: Is.Ordered - show item where order breaks in error message
Currently NUnit only logs actual collection in error message. That's fine if we have 3 items, but not really great for 200, especially due to that only several first items are displayed.
Would be really great to show where order actually breaks somehow:

Something like this, maybe?
```
Expected: collection ordered
Actual: < "1", "2", "4", "3" >
Order breaks at index [2] with value "4" 
```

Besides, I would show collection with start index of `breakingIndex - 1` in actual, so that this item would always be displayed even for huge collections.

(Probably something trickier, like
```
int startIndex = breakingIndex > 8 ? breakingIndex - 1 : 0;
```
)

Dreamescaper:
fixes #2855 

Added line to error message with item at which ordering 
```
Expected: collection ordered
But was:  < "x", "z", "y" >
Ordering breaks at index [2]:  "y"    <----------- ADDED LINE
```

Besides, actual collection logging changed a bit, so it would always show item at which ordering breaks.
So if we have collection of 100 elements, and ordering breaks at index 91, it would look like that:
```
Expected: collection ordered
But was:  < 83, 84, 85, 86, 87, 88, 89, 1000, 91, 92... >
Ordering breaks at index [91]:  91
```

Open questions:
1. `CollectionOrderedConstraint` has `CollectionConstraint` as base class, but currently it's not used at all. I decided to leave it as is (as technically removing that could be a breaking change), but maybe you have other opinion.

2. I would prefer to put ellipsis at the beginning of actual collection, if it starts not from index 0 
(like `< ...83, 84, 85, 86, 87, 88, 89, 1000, 91, 92... >`). 
I could add it to MsgUtils.FormatCollections (if `start` is not 0), but that would require to change Missing/Extra items logging logic for CollectionEquivalent constraint. Do you have any objections or suggestions?

### DIFF
--- a/src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
@@ -152,12 +152,12 @@ namespace NUnit.Framework.Constraints
                 else if (failurePoint.ActualHasData)
                 {
                     writer.Write("  Extra:    ");
-                    writer.WriteCollectionElements(SkipItems(actual, failurePoint.Position), 0, 3);
+                    writer.WriteCollectionElements(actual.Skip(failurePoint.Position), 0, 3);
                 }
                 else
                 {
                     writer.Write("  Missing:  ");
-                    writer.WriteCollectionElements(SkipItems(expected, failurePoint.Position), 0, 3);
+                    writer.WriteCollectionElements(expected.Skip(failurePoint.Position), 0, 3);
                 }
             }
         }
@@ -242,18 +242,6 @@ namespace NUnit.Framework.Constraints
             return null;
         }
 
-        private static IEnumerable SkipItems(IEnumerable enumerable, long skip)
-        {
-            var iterator = enumerable.GetEnumerator();
-            while (skip-- > 0)
-            {
-                iterator.MoveNext();
-            }
-            while (iterator.MoveNext())
-            {
-                yield return iterator.Current;
-            }
-        }
         #endregion
 
         #region DisplayEnumerableDifferences
--- a/src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2011 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -63,7 +63,7 @@ namespace NUnit.Framework.Constraints
         /// Construct an EqualConstraintResult
         /// </summary>
         public EqualConstraintResult(EqualConstraint constraint, object actual, bool hasSucceeded)
-            : base(constraint, actual, hasSucceeded) 
+            : base(constraint, actual, hasSucceeded)
         {
             this.expectedValue = constraint.Arguments[0];
             this.tolerance = constraint.Tolerance;
@@ -152,12 +152,12 @@ namespace NUnit.Framework.Constraints
                 else if (failurePoint.ActualHasData)
                 {
                     writer.Write("  Extra:    ");
-                    writer.WriteCollectionElements(actual, failurePoint.Position, 3);
+                    writer.WriteCollectionElements(SkipItems(actual, failurePoint.Position), 0, 3);
                 }
                 else
                 {
                     writer.Write("  Missing:  ");
-                    writer.WriteCollectionElements(expected, failurePoint.Position, 3);
+                    writer.WriteCollectionElements(SkipItems(expected, failurePoint.Position), 0, 3);
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


jnm2:
> I decided to leave it as is (as technically removing that could be a breaking change)

Yes, we wouldn't consider moving something like that without a big benefit to balance it out.

> I would prefer to put ellipsis at the beginning of actual collection, i

I was going to suggest the same thing in the discussion issue. (Which incidentally is a good place for discussions like this. 😉)

Dreamescaper:
Well, question whether ellipsis should be present or not should be in issue discussion, but how to implement it is fine here, right? :)

I suppose in order to do that I'd need either to add in to MsgUtils, and rework CollectionEquivalent, or simply ignore MsgUtils, and construct message in constraint itself.

jnm2:
Yes, it is!

I like the idea of adding to MsgUtils, maybe as a new overload with a new `int skip` parameter?

rprouse:
I like the idea of this, but the implementation needs some cleaning up. I don't like that you are no longer using the protected `Matches`, but calling your own version in `ApplyTo`. I also don't like that you return a different result class for success and for failure. I would prefer if you saved the information you need as fields and didn't have your overridden match.

jnm2:
@Dreamescaper Checking in, can we help with anything?

Dreamescaper:
@jnm2 
Sorry for delay! I've made changes according to PR comments.

I haven't added ellipsis at start in error message, if several items are skipped. So currently it's `< 83, 84, 85, 86... >`, not `< ...83, 84, 85, 86... `. Not sure about best way to proceed here.

How it currently works in MsgUtils. There's method 
`public static string FormatCollection(IEnumerable collection, long start, int max)`.
It prints up `max` number of items, starting from `start` index.

I can see the following options:

1) Add ellipsis to FormatCollection method, if `start` is not zero. There is only one using of that method with non-zero `start` argument (CollectionEquivalentConstraintResult). That means I'd have to use something different there - something like linq's `Skip` method, but custom, supporting `long` argument type.

2) In added CollectionOrderedConstraintResult - don't use MsgUtils' method, build actual result message there instead.

3) Screw it, it's fine now.

jnm2:
@Dreamescaper I like your option 1, but wouldn't it make sense to change `CollectionEquivalentConstraintResult` to use a leading `...` for nonzero start indexes too?

Dreamescaper:
@jnm2 
UPDATE: missed it - it's not `CollectionEquivalentConstraintResult`, it's `EqualConstraintResult.DisplayCollectionDifferences`.

Thing is it used there to log actual collection, but to log Extra/Missing items (don't fully understand logic here, but anyway), so leading ellipsis is not an option:
```

                else if (failurePoint.ActualHasData)
                {
                    writer.Write("  Extra:    ");
                    writer.WriteCollectionElements(actual, failurePoint.Position, 3);
                }
                else
                {
                    writer.Write("  Missing:  ");
                    writer.WriteCollectionElements(expected, failurePoint.Position, 3);
                }
```
(`WriteCollectionElements` calls `MsgUtils.FormatCollection` underneath)

Dreamescaper:
@jnm2 
Committed changes with leading ellipsis. Let me know if there's fancier way :)

src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
```diff
@@ -241,6 +241,19 @@ private static object GetValueFromCollection(ICollection collection, int index)
 
             return null;
         }
+
+        private static IEnumerable SkipItems(IEnumerable enumerable, long skip)
+        {
+            var iterator = enumerable.GetEnumerator();
+            while (skip-- > 0)
+            {
+                iterator.MoveNext();
```
jnm2: `.Cast<object>().Skip(...)` is simple and available here on all target platforms. But if we go ahead with this, I think it's contractually invalid to call `MoveNext` if it returns false, so we should yield break. And maybe move it to an internal extension methods class.

jnm2:
Otherwise, looks good but we should have @nunit/framework-team confirm the decision and implementation.

src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
jnm2: And *technically* we should be doing `using (iterator as IDisposable)` around it too, to match the behavior of `foreach`.

Dreamescaper: 1) Skip is not available, as it doesn't accept long argument (failurePoint.Position).
(just curious - do we actually support collections with more than int.MaxValue elements?)

2) While I think it makes sense to add `yield break` (or some other if condition) to avoid unneeded calls, could you suggest where could I read that it's invalid? That sounds strange to me a bit (since in that case I cannot determine lookng at enumerator object whether I'm allowed to call MoveNext)?

3) Could you suggest where could I put it? Or create new class?

jnm2: 1. Oh my, I suppose we would support enumerables longer than int.MaxValue!

2. I just checked myself; I was wrong. Calling `MoveNext` again after the last call returned false [is actually legal](https://docs.microsoft.com/dotnet/api/system.collections.ienumerator.movenext#remarks). (I was thinking of Current, which [should not be called](https://docs.microsoft.com/dotnet/api/system.collections.generic.ienumerator-1.current) after MoveNext returns false.)
Either way, it doesn't make sense to me to ignore the return value of MoveNext.

3. Probably the `NUnit.Framework.Extensions` class.

Dreamescaper: Updated.

PR #2894, opened 2018-06-19
Move netstandard1.6 build to netstandard1.4

### LINKED ISSUE #2887: NETStandard 1.3 support dropped in NUnit 3.10
Hello,
The support for .NET Standard 1.3 seems to be dropped in NUnit 3.10, and now the lowest .NET Standard is 1.6, which does not work with Universal when targeting RS1 or RS2 (the latest version that works is 1.4). I might have missed it, but I did not find any notes on why .NET Standard 1.3 support is dropped. So, just wondering what options there might be to for RS1/RS2? Could we bring back earlier .NET Standard support?

jnm2:
Fixes #2887. The first commit adds a netstandard1.4 build and the second commit removes the netstandard1.6 build.

Rather than adding back all our previous .NET Standard 1.3 reflection polyfills, the .NET Standard 1.4 build takes a dependency on System.Reflection.TypeExtensions.


nuget/framework/nunit.nuspec
```diff
@@ -28,8 +28,9 @@ Supported platforms:
       <group targetFramework="net35" />
       <group targetFramework="net40" />
       <group targetFramework="net45" />
-      <group targetFramework="netstandard1.6">
+      <group targetFramework="netstandard1.4">
         <dependency id="NETStandard.Library" version="1.6.1" />
+        <dependency id="System.Reflection.TypeExtensions" version="4.4.0" exclude="compile" />
```
jnm2: `exclude="compile"` means users' test projects won't reference assemblies in the `System.Reflection.TypeExtensions` package because of the NUnit package.

ChrisMaddock:
I commented on the issue re: the package dependency, to keep the conversation in one place. 🙂

Issue #2889, opened 2018-06-13
Multidimensional array argument contents aren't used in name of parameterized tests.


Brar:
@CharliePoole wrote in #2597:

> What if my argument is a multidimensional array?

I just stumbled over this while implementing some advanced testing for [Npgsql](https://github.com/npgsql/npgsql/).

As I already had implemented it for my tests, I created this [Gist](https://gist.github.com/Brar/260a670863121bb0da616976d45681c4) that shows a possible implementation of multidimensional array formatting for NUnit.

If you think that something like this would be accepted into NUnit I am willing to work on a pull request.

jnm2:
I would like something like this, though I'm not sure about the `A` part yet.

@nunit/framework-team what do you think?

Brar:
> I'm not sure about the A part yet.

I absolutely share this feeling.
If we'd omit it, there would be no way to tell the difference between multidimensional and jagged arrays.

Initially I used 'new ' but it takes up 4 characters for very little use and we don't use C# semantics anyways.

I'm happy to use whatever you like best.

jnm2:
We appreciate that! Let's see what we end up with. To play devil's advocate, what's the worst thing that would happen if multidimensional and jagged arrays resulted in the same test name?

Brar:
> To play devil's advocate, what's the worst thing that would happen if multidimensional and jagged arrays resulted in the same test name?

Haha, that's simple.
If you have a test that systematically tests different types of arguments and want to test multidimensional arrays and jagged arrays, how would you tell which test is failing?

I'll prepare a pull request, so that we can discuss code.

jnm2:
One way is by specifying your own TestName. In general we try to optimize make the 99% case the nicest if possible (for example, no `A`) while leaving a way to accomplish the 1% case without making it disproportionately difficult.

A better way to ask my question would be, how often are people going to be sending both multidimensional and jagged arrays to the same test method parameter?

PR #2895, opened 2018-06-19
 Enable SetCulture and SetUICulture attributes for .NET-Standard 1.4 and fix tests that are failing in non-english locales

Brar:
While drafting an initial pull request for #2889 I recognized some tests that are failing on my system due to locale issues (I'm on a de-DE locale).

I looked into it and found that SetCulture/SetUICulture aren't used for NUnits own tests probably because those attributes are disabled for .Net-Standard ~~1.6~~ 1.4.

This isn't really necessary and probably is a leftover from older compatibility requirements (perhaps PORTABLE or NETSTANDARD1_3).

This pull request  enables SetCultureAttribute and SetUICultureAttribute for .NET-Standard ~~1.6~~ 1.4 which means that it now works on every platform that NUnit currently supports.
It also uses those attributes to fix some tests that are currently failing in non-english locales.

As this is pretty trivial I've opted for one pull request but left two separate commits to split the changes into logical units.

### DIFF
--- a/src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
+++ b/src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
@@ -94,7 +94,7 @@ namespace NUnit.Framework.Api
             Assert.That(result.TestCaseCount, Is.EqualTo(MockAssembly.Tests));
         }
 
-        [Test]
+        [Test, SetUICulture("en-US")]
         public void Load_FileNotFound_ReturnsNonRunnableSuite()
         {
             var result = _runner.Load(MISSING_FILE, EMPTY_SETTINGS);
@@ -108,7 +108,7 @@ namespace NUnit.Framework.Api
                 Does.StartWith(COULD_NOT_LOAD_MSG));
         }
 
-        [Test]
+        [Test, SetUICulture("en-US")]
         public void Load_BadFile_ReturnsNonRunnableSuite()
         {
             var result = _runner.Load(BAD_FILE, EMPTY_SETTINGS);
@@ -263,7 +263,7 @@ namespace NUnit.Framework.Api
             Assert.That(ex.Message, Is.EqualTo("The Run method was called but no test has been loaded"));
         }
 
-        [Test]
+        [Test, SetUICulture("en-US")]
         public void Run_FileNotFound_ReturnsNonRunnableSuite()
         {
             _runner.Load(MISSING_FILE, EMPTY_SETTINGS);
@@ -315,7 +315,7 @@ namespace NUnit.Framework.Api
             CheckParameterOutput(result);
         }
 
-        [Test]
+        [Test, SetUICulture("en-US")]
         public void Run_BadFile_ReturnsNonRunnableSuite()
         {
             _runner.Load(BAD_FILE, EMPTY_SETTINGS);
@@ -377,7 +377,7 @@ namespace NUnit.Framework.Api
             Assert.That(ex.Message, Is.EqualTo("The Run method was called but no test has been loaded"));
         }
 
-        [Test]
+        [Test, SetUICulture("en-US")]
         public void RunAsync_FileNotFound_ReturnsNonRunnableSuite()
         {
             _runner.Load(MISSING_FILE, EMPTY_SETTINGS);
@@ -394,7 +394,7 @@ namespace NUnit.Framework.Api
                 Does.StartWith(COULD_NOT_LOAD_MSG));
         }
 
-        [Test]
+        [Test, SetUICulture("en-US")]
         public void RunAsync_BadFile_ReturnsNonRunnableSuite()
         {
             _runner.Load(BAD_FILE, EMPTY_SETTINGS);
--- a/src/NUnitFramework/tests/Assertions/AssertThrowsAsyncTests.cs
+++ b/src/NUnitFramework/tests/Assertions/AssertThrowsAsyncTests.cs
@@ -211,7 +211,7 @@ namespace NUnit.Framework.Assertions
                 "  But was:  <System.Exception: my message" + Environment.NewLine));
         }
 
-        [Test]
+        [Test, SetUICulture("en-US")]
         public void DerivedExceptionThrown()
         {
             var ex = CatchException(() => Assert.ThrowsAsync<Exception>(AsyncTestDelegates.ThrowsArgumentException));
@@ -222,7 +222,7 @@ namespace NUnit.Framework.Assertions
             CheckForSpuriousAssertionResults();
         }
 
-        [Test]
+        [Test, SetUICulture("en-US")]
         public void DerivedExceptionThrownAsync()
         {
             var ex = CatchException(() => Assert.ThrowsAsync<Exception>(AsyncTestDelegates.ThrowsArgumentExceptionAsync));
--- a/src/NUnitFramework/tests/Assertions/AssertThrowsTests.cs
+++ b/src/NUnitFramework/tests/Assertions/AssertThrowsTests.cs
@@ -168,7 +168,7 @@ namespace NUnit.Framework.Assertions
             CheckForSpuriousAssertionResults();
         }
 
-        [Test]
+        [Test, SetUICulture("en-US")]
         public void DerivedExceptionThrown()
         {
             var ex = CatchException(() => 
--- a/src/NUnitFramework/tests/Assertions/TypeAssertTest.cs
+++ b/src/NUnitFramework/tests/Assertions/TypeAssertTest.cs
@@ -72,7 +72,7 @@ namespace NUnit.Framework.Assertions
             Assert.IsNotInstanceOf<System.Int32>("abc123");
         }
 
-        [Test]
+        [Test, SetUICulture("en-US")]
         public void IsNotInstanceOfFails()
         {
             var expectedMessage =
--- a/src/NUnitFramework/tests/Constraints/ThrowsConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/ThrowsConstraintTests.cs
@@ -118,7 +118,7 @@ namespace NUnit.Framework.Constraints
             }
         }
 
-        [Test, TestCaseSource("FailureData")]
+        [Test, TestCaseSource("FailureData"), SetUICulture("en-US")]
         public void FailsWithBadValues(object badValue, string message)
         {
             string NL = Environment.NewLine;
--- a/src/NUnitFramework/framework/Attributes/SetCultureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/SetCultureAttribute.cs
@@ -21,7 +21,6 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-#if !NETSTANDARD1_4
 using System;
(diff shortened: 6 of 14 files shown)


Brar:
@ChrisMaddock thanks for your feedback

> Please could you also re-enable the tests for the attributes?

Funny that I didn't think about this.
I already had a look at it but I'm getting a few seemingly unrelated test failures when testing for dotnetcore1.1 (OneTimeSetUp failures so probably some misconfiguration).
It might take a little time until I've found out what's wrong.

> Can you clarify why these tests were failing in de-DE?

Well, it's mostly related to assertions of English exception messages so it seemed like preatty clear cases to me but I'l review my changes test by test to be sure there isn't something else going on.

Brar:
@ChrisMaddock @jnm2 
Ok, I just found out that the baseline [now](https://github.com/nunit/nunit/commit/3dfb4fca636e6b4b22b508f4679e214bd88dca45) is .Net-Standard 1.4

I'll reevaluate my changes with 1.4.
Are there plans to go even lower than this?

Brar:
> I'll reevaluate my changes with 1.4.

There's no Problem with [.Net-Standard 1.4](https://docs.microsoft.com/de-de/dotnet/api/system.globalization.cultureinfo.currentculture?view=netstandard-1.4#System_Globalization_CultureInfo_CurrentCulture)

Rebased and renamed now.

mikkelbu:
> Are there plans to go even lower than this?

Not as far as I know.

Ps. We should probably also see if this solves (part of) #2228 at some time. At least they seems closely related.

jnm2:
@Brar To confirm what @mikkelbu said, going lower than .NET Standard 1.4 [does not gain us any interesting capabilities](https://docs.microsoft.com/dotnet/standard/net-standard) since we already ship builds net20—net45.

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
```diff
@@ -95,6 +95,7 @@ public void Load_GoodFile_ReturnsRunnableSuite()
         }
 
         [Test]
+        [SetUICulture("en-US")]
```
jnm2: Once we enable the SetCulture tests, I'm really excited about this PR and I'll want to get it in.

However, I would like us to review each of these changes on a case-by-case basis with this question in mind: is the right fix to change the framework to use an ordinal comparison? For example, with files, I would say that should absolutely be an OrdinalIgnoreCase comparison, never culture-dependent.

Brar: >For example, with files, I would say that should absolutely be an OrdinalIgnoreCase comparison, never culture-dependent.

It's not about ordinal comparison but mostly about exception messages that are localized.

This test (SetUICulture commented out):

```csharp
private const string COULD_NOT_LOAD_MSG = "Could not load";
private const string MISSING_FILE = "junk.dll";
private static readonly IDictionary<string, object> EMPTY_SETTINGS = new Dictionary<string, object>();

[Test]
//[SetUICulture("en-US")]
public void Load_FileNotFound_ReturnsNonRunnableSuite()
{
    var result = _runner.Load(MISSING_FILE, EMPTY_SETTINGS);

    Assert.That(result.IsSuite);
    Assert.That(result, Is.TypeOf<TestAssembly>());
    Assert.That(result.Name, Is.EqualTo(MISSING_FILE));
    Assert.That(result.RunState, Is.EqualTo(Interfaces.RunState.NotRunnable));
    Assert.That(result.TestCaseCount, Is.EqualTo(0));
    Assert.That(result.Properties.Get(PropertyNames.SkipReason),
        Does.StartWith(COULD_NOT_LOAD_MSG));
}
```

produces the following test failure:

```
1) Failed : NUnit.Framework.Api.TestAssemblyRunnerTests.Load_FileNotFound_ReturnsNonRunnableSuite
  Expected: String starting with "Could not load"
  But was:  "Die Datei oder Assembly "junk.dll" oder eine Abhängigkeit davon wurde nicht gefunden. Das System kann die angegebene Datei nicht finden.
  ----> Das System kann die angegebene Datei nicht finden. (Ausnahme von HRESULT: 0x80070002)"
   bei NUnit.Framework.Api.TestAssemblyRunnerTests.Load_FileNotFound_ReturnsNonRunnableSuite() in C:\Users\Brar\Documents\GitHub\nunit\src\NUnitFramework\tests\Api\TestAssemblyRunnerTests.cs:Zeile 110.
```

While we could certainly assert on the non-localized elements that are contained in both strings it wouldn't be much more than the file name in this case.

With `[SetUICulture("en-US")]` (funny that it's still partially localized):
```
Could not load file or assembly 'junk.dll' or one of its dependencies. Das System kann die angegebene Datei nicht finden.
  ----> Das System kann die angegebene Datei nicht finden. (Exception from HRESULT: 0x80070002)
```

Without `[SetUICulture("en-US")]`:
```
Die Datei oder Assembly "junk.dll" oder eine Abhängigkeit davon wurde nicht gefunden. Das System kann die angegebene Datei nicht finden.
  ----> Das System kann die angegebene Datei nicht finden. (Ausnahme von HRESULT: 0x80070002)
```

Brar: BTW: Can anyone tell me why/how the test cited above (and quite a few others) aren't executed with `.\build.ps1 -t TestNetStandard14`?

I get
```
Test Count: 5064, Passed: 4939, Failed: 0, Warnings: 0, Inconclusive: 109, Skipped: 16
    Skipped Tests - Ignored: 3, Explicit: 13, Other: 0
```
for `.\build.ps1 -t Test45` and
```
  Test Count: 2719, Passed: 2699, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 20
    Skipped Tests - Ignored: 14, Explicit: 6, Other: 0
```
for `.\build.ps1 -t TestNetStandard14`.

There are no `#if` definitions disabling the aforementioned test for a certain target and I have even confirmed (via ildasm) that it's in the compiled `nunit.framework.tests.dll` for `netcoreapp1.1`, yet it isn't executed.

jnm2: > It's not about ordinal comparison but mostly about exception messages that are localized.

Thanks for confirming. Each of these cases is indeed about localized exception messages, and this is the right fix.

> BTW: Can anyone tell me why/how the test cited above (and quite a few others) aren't executed

That's concerning; taking a look.

jnm2: The tests were not running for netstandard1.6 prior to my PR. I think I know which PR this was, though

jnm2: Confirmed. https://github.com/nunit/nunit/pull/2796 introduced the regression.

src/NUnitFramework/tests/Assertions/AssertThrowsAsyncTests.cs
```diff
@@ -212,6 +212,7 @@ public void BaseExceptionThrownAsync()
         }
 
         [Test]
+        [SetUICulture("en-US")]
         public void DerivedExceptionThrown()
```
Brar: Failure reason: Test asserts a constant English exception message that doesn't match the localized exception message.

src/NUnitFramework/tests/Assertions/AssertThrowsAsyncTests.cs
```diff
@@ -223,6 +224,7 @@ public void DerivedExceptionThrown()
         }
 
         [Test]
+        [SetUICulture("en-US")]
         public void DerivedExceptionThrownAsync()
```
Brar: Failure reason: Test asserts a constant English exception message that doesn't match the localized exception message.

src/NUnitFramework/tests/Assertions/AssertThrowsTests.cs
```diff
@@ -169,6 +169,7 @@ public void BaseExceptionThrown()
         }
 
         [Test]
+        [SetUICulture("en-US")]
         public void DerivedExceptionThrown()
```
Brar: Failure reason: Test asserts a constant English exception message that doesn't match the localized exception message.

src/NUnitFramework/tests/Assertions/TypeAssertTest.cs
```diff
@@ -73,6 +73,7 @@ public void IsNotInstanceOf()
         }
 
         [Test]
+        [SetUICulture("en-US")]
         public void IsNotInstanceOfFails()
```
Brar: Failure reason: Test asserts a constant English exception message that doesn't match the localized exception message.

src/NUnitFramework/tests/Constraints/ThrowsConstraintTests.cs
```diff
@@ -119,6 +119,7 @@ public void SucceedsWithGoodValues(object value)
         }
 
         [Test, TestCaseSource("FailureData")]
+        [SetUICulture("en-US")]
         public void FailsWithBadValues(object badValue, string message)
```
Brar: Failure reason: Test asserts a constant English exception message that doesn't match the localized exception message.

Brar:
> Forgot, we are going to need to remove every #if !NETCOREAPP1_1 that surrounds a usage of SetCulture or SetUICulture.

This was done by my latest commit although I'cant confirm that all newly enabled tests are actually working due to #2896

ChrisMaddock:
Thanks for running through all these @Brar. I’d thought some of these messages were hard coded strings in the framework - I guess I was wrong!

I surprised so many have sneaked in. 🙁 I guess ideally, we’d run our tests in a couple of different cultures.

jnm2:
> This was done by my latest commit although I'cant confirm that all newly enabled tests are actually working due to #2896

Yes, can this PR wait until we get a fix merged for https://github.com/nunit/nunit/issues/2896? I do apologize.

Brar:
> Yes, can this PR wait until we get a fix merged for #2896?

Certainly. There's no need to hurry.

jnm2:
I've verified that no tests fail when rebasing on top of https://github.com/nunit/nunit/pull/2899, so as soon as you edit `CultureAttributeData`, this can be merged without waiting.

Brar:
> One last file: CultureAttributeData.cs. I'm actually in favor of removing the entire contents of the #if because all targets build and test just fine without it.

Ok, I've amended the last commit with this (after verifying that no test uses those fixtures and there seems to be no need for such a test using them as NUnit doesn't do anything about invalid cultures).

Also I've amended `Fix tests that are failing in non-english locales` so that the `SetUICulture` attribute is in the same line as the `Test`attribute as it appeared to me that this is the de-facto coding standard in your code base.

Finally I've reworded `
Enable SetCulture and SetUICulture attributes for .NET-Standard 1.6` to `
Enable SetCulture and SetUICulture attributes for .NET-Standard 1.4`.

@jnm2 I think this should be ready to go now.

jnm2:
@Brar Just waiting for a review from a second @nunit/framework-team member now.

ChrisMaddock:
Looks good to me. Thanks for this, @Brar! 🎉

mikkelbu:
@rprouse As far as I can tell this was part of the 3.11 release, but not marked with the milestone and also not mentioned in the release notes.

ChrisMaddock:
Sorry - my fault!

PR #2910, opened 2018-06-25
Added information and warning properties to TestContext  #2908

### LINKED ISSUE #2908: TestContext should have test-output properties for Information and Warning
TestContext currently have two properties for test-output, Error and Progress.  
1. Both write to the StdError, but only Error is a real error, Progress is informational and should be written to StdOut
2. There should be two more properties Information and Warning, allowing the user to easily add context information in the test that appears in the test output. 

Note:  I am fully aware that parallel execution makes a mess out of such messages, but this should be up to the user.  If the user have a need for these messages they should be able to use them, for diagnostic purposes or for any reason they like.  

I am adding a corresponding issue to the NUnit3TestAdapter to respond correctly to these 4 streams.  For the adapter it doesn't matter which stream is being used (StdError or StdOut) as it only see the named "streams" and the value being sent.

OsirisTerje:
Closes #2908

### DIFF
--- a/src/NUnitFramework/tests/TextOutputTests.cs
+++ b/src/NUnitFramework/tests/TextOutputTests.cs
@@ -92,6 +92,8 @@ namespace NUnit.Framework
             // Test purely for display purposes
             TestContext.Progress.WriteLine("TestContext.Progress displays immediately");
             TestContext.Error.WriteLine("TestContext.Error displays immediately as well");
+            TestContext.Information.WriteLine("TestContext.Information displays immediately as well");
+            TestContext.Warning.WriteLine("TestContext.Warning displays immediately as well");
             Console.Error.WriteLine("Console.Error also displays immediately");
             Console.WriteLine("This line is added to the result and displayed when test ends");
             Console.WriteLine("As is this line");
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/tests/TextOutputTests.cs
```diff
@@ -92,6 +92,8 @@ public void MultipleWrites()
             // Test purely for display purposes
             TestContext.Progress.WriteLine("TestContext.Progress displays immediately");
             TestContext.Error.WriteLine("TestContext.Error displays immediately as well");
+            TestContext.Information.WriteLine("TestContext.Information displays immediately as well");
+            TestContext.Warning.WriteLine("TestContext.Warning displays immediately as well");
```
CharliePoole: Most runners won't actually display these.

CharliePoole:
I've discussed on the issue why I think this feature is a mistake. Since I no longer formally review changes to the framework, I won't make an exception here by requesting changes. I'll just say that this requires changes to every runner and invites mire issue reports about why it sort of looks like conditional logging but really isn't.

CharliePoole:
No issue reference.

rprouse:
This has been dormant for three years, closing.

PR #2912, opened 2018-06-26
Make DataAttribute obsolete and unnecessary

### LINKED ISSUE #2829: Obsoletion warning for DataAttribute
[Quoting Charlie,](
https://github.com/nunit/nunit/pull/2803/files#r184545757)

> DataAttribute is a class that was somewhat important in earlier versions of NUnit. Currently, it's a holdover and is not used in the code at all. We left it in the code to avoid needlessly breaking anything.

This seems like a good thing to document in the source. Can we document this using an obsoletion warning so that people who use it become aware?

```cs
[Obsolete("Holdover from NUnit v2. Please inherit from Attribute or PropertyAttribute and " +
          "implement IParameterDataSource.", error: false)]
```

YuriyOborozhnyi:
Close #2829.

I've found and edited all usages of `DataAttribute` in this repository. However I it is mentioned in https://github.com/nunit/docs/wiki/Active-Attributes and https://github.com/nunit/docs/wiki/Addin-Replacement-in-the-Framework as well. Should we updated these documentation pages too or not?

### DIFF
--- a/src/NUnitFramework/framework/Attributes/AttributeHierarchy.txt
+++ b/src/NUnitFramework/framework/Attributes/AttributeHierarchy.txt
@@ -5,10 +5,10 @@ NUnitAttribute
 		TestCaseActionAttribute       : 
 		TestSuiteActionAttribute      : 
 	DataAttribute
-		ValuesAttribute               : IParameterDataSource
-			RandomAttribute           : IParameterDataSource
-		RangeAttribute                : IParameterDataSource
-		ValueSourceAttribute          : IParameterDataSource
+	ValuesAttribute                   : IParameterDataSource
+	RandomAttribute                   : IParameterDataSource
+	RangeAttribute                    : IParameterDataSource
+	ValueSourceAttribute              : IParameterDataSource
 	DatapointAttribute
 	DatapointSourceAttribute
 		DatapointsAttribute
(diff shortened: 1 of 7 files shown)


src/NUnitFramework/framework/Attributes/AttributeHierarchy.txt
```diff
@@ -5,10 +5,10 @@ NUnitAttribute
 		TestCaseActionAttribute       : 
 		TestSuiteActionAttribute      : 
 	DataAttribute
```
ChrisMaddock: Did you intent to remove DataAttribute here? 🙂

ChrisMaddock:
I also reran the failed CI build. That was caused by a flakey test, I'm afraid!

src/NUnitFramework/framework/Attributes/AttributeHierarchy.txt
YuriyOborozhnyi: I just left it there since the attribute isn't removed actually, but marked as obsolete.
On the other hand it sounds reasonable to remove this from `AttributeHierarchy`, because it is just NUnit 2 holdover.

So what do you think?

ChrisMaddock: What you've done seems sensible to me. 🙂 We can remove it when we actually remove it from the codebase.

YuriyOborozhnyi:
@ChrisMaddock, thank you so much! I tried to figure out why AppVeyor build failed, but that test run OK on my machine.

PR #2913, opened 2018-06-27
Change TestAssembly to only take in assembly name, clean up the usage of path and name

### LINKED ISSUE #2898: AssemblyPath contains invalid charaters
When running test on UWP, we ran into an ArgumentException (Illegal characters in path) during BuildTestAssembly in DefaultTestAssemblyBuilder.cs. It seems to be complaining about assemblyPath, which is "\<unknown\>" instead of the real assembly path. Here are the related codes:
https://github.com/nunit/nunit/blob/e94c6d9cd4655c63c514ce354f7e8d451dc0118a/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs#L83
https://github.com/nunit/nunit/blob/e94c6d9cd4655c63c514ce354f7e8d451dc0118a/src/NUnitFramework/framework/Internal/AssemblyHelper.cs#L56
This issue was introduced in NUnit 3.7. Prior to 3.7, it calls `AssemblyHelper.GetAssemblyName(assembly).FullName` to get the assemblyPath, which works fine. Now, it all uses `Assembly.GetAssemblyPath(assembly)`. According to [.NET API](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.module.fullyqualifiedname?view=netframework-4.7.2#System_Reflection_Module_FullyQualifiedName), `FullyQualifiedName` for the module might be "\<unknown\>" in certain situations. Switching back to GetAssemblyName simply fixes the problem here.

I may be wrong, but it seems to me that the only use of assemblyPath is to get the name of the assembly later. So why not just getting the name instead of the path here?

jrjrguo:
The motivation of this change is to fix #2898 . The assemblyPath of an assembly might be <unknown> when running uwp tests. The fix is to use the assemblyName instead in those cases.
The main improvement is to TestAssembly so that it only takes in assembly name, since that's the only thing it actually uses. It was a bit misleading before, which is probably why assemblyPath and assemblyName are a little misused in multiple classes. Now the caller should only pass in name to TestAssembly.

### DIFF
--- a/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
+++ b/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
@@ -81,17 +81,17 @@ namespace NUnit.Framework.Api
 #endif
 
             string assemblyPath = AssemblyHelper.GetAssemblyPath(assembly);
-            string assemblyName = assemblyPath.IndexOfAny(Path.GetInvalidPathChars()) == -1
-                ? Path.GetFileName(assemblyPath)
-                : AssemblyHelper.GetAssemblyName(assembly).FullName;
+            string suiteName = assemblyPath.Equals("<Unknown>")
+                ? AssemblyHelper.GetAssemblyName(assembly).FullName
+                : Path.GetFileName(assemblyPath);
 
-            return Build(assembly, assemblyName, options);
+            return Build(assembly, suiteName, options);
         }
 
         /// <summary>
-        /// Build a suite of tests given the filename of an assembly
+        /// Build a suite of tests given the name or the location of an assembly
         /// </summary>
-        /// <param name="assemblyNameOrPath">The filename or path of the assembly from which tests are to be built</param>
+        /// <param name="assemblyNameOrPath">The name or the location of the assembly.</param>
         /// <param name="options">A dictionary of options to use in building the suite</param>
         /// <returns>
         /// A TestSuite containing the tests found in the assembly
@@ -120,7 +120,7 @@ namespace NUnit.Framework.Api
             return testAssembly;
         }
 
-        private TestSuite Build(Assembly assembly, string assemblyName, IDictionary<string, object> options)
+        private TestSuite Build(Assembly assembly, string suiteName, IDictionary<string, object> options)
         {
             TestSuite testAssembly = null;
 
@@ -172,11 +172,11 @@ namespace NUnit.Framework.Api
                     fixtureNames = options[FrameworkPackageSettings.LOAD] as IList;
                 var fixtures = GetFixtures(assembly, fixtureNames);
 
-                testAssembly = BuildTestAssembly(assembly, assemblyName, fixtures);
+                testAssembly = BuildTestAssembly(assembly, suiteName, fixtures);
             }
             catch (Exception ex)
             {
-                testAssembly = new TestAssembly(assemblyName);
+                testAssembly = new TestAssembly(suiteName);
                 testAssembly.MakeInvalid(ExceptionHelper.BuildMessage(ex, true));
             }
 
@@ -258,9 +258,9 @@ namespace NUnit.Framework.Api
         // Process class is used, so we can safely satisfy the link demand with a 'SecuritySafeCriticalAttribute' rather
         // than a 'SecurityCriticalAttribute' and allow use by security transparent callers.
         [SecuritySafeCritical]
-        private TestSuite BuildTestAssembly(Assembly assembly, string assemblyName, IList<Test> fixtures)
+        private TestSuite BuildTestAssembly(Assembly assembly, string suiteName, IList<Test> fixtures)
         {
-            TestSuite testAssembly = new TestAssembly(assembly, assemblyName);
+            TestSuite testAssembly = new TestAssembly(assembly, suiteName);
 
             if (fixtures.Count == 0)
             {
--- a/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
+++ b/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
@@ -81,42 +81,46 @@ namespace NUnit.Framework.Api
 #endif
 
             string assemblyPath = AssemblyHelper.GetAssemblyPath(assembly);
-            return Build(assembly, assemblyPath, options);
+            string assemblyName = assemblyPath.IndexOfAny(Path.GetInvalidPathChars()) == -1
+                ? Path.GetFileName(assemblyPath)
+                : AssemblyHelper.GetAssemblyName(assembly).FullName;
+
+            return Build(assembly, assemblyName, options);
         }
 
         /// <summary>
         /// Build a suite of tests given the filename of an assembly
         /// </summary>
-        /// <param name="assemblyName">The filename of the assembly from which tests are to be built</param>
+        /// <param name="assemblyNameOrPath">The filename or path of the assembly from which tests are to be built</param>
         /// <param name="options">A dictionary of options to use in building the suite</param>
         /// <returns>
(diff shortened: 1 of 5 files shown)


src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -81,42 +81,46 @@ public ITest Build(Assembly assembly, IDictionary<string, object> options)
...
+                : AssemblyHelper.GetAssemblyName(assembly).FullName;
+
+            return Build(assembly, assemblyName, options);
         }
 
         /// <summary>
         /// Build a suite of tests given the filename of an assembly
         /// </summary>
-        /// <param name="assemblyName">The filename of the assembly from which tests are to be built</param>
+        /// <param name="assemblyNameOrPath">The filename or path of the assembly from which tests are to be built</param>
```
jnm2: I think this should actually say `the assembly name or filename` or `the name or location of the assembly`.

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -81,42 +81,46 @@ public ITest Build(Assembly assembly, IDictionary<string, object> options)
 #endif
 
             string assemblyPath = AssemblyHelper.GetAssemblyPath(assembly);
-            return Build(assembly, assemblyPath, options);
+            string assemblyName = assemblyPath.IndexOfAny(Path.GetInvalidPathChars()) == -1
```
jnm2: I'm almost happier if we specially recognize `<Unknown>` inside `AssemblyHelper.GetAssemblyPath` and return null.

If there are invalid path characters for any other reason, wouldn't we'd want to know about it right away and not just treat it as though it was the result of a call to `Assembly.ManifestModule.FullyQualifiedName` returning its documented `<Unknown>` value?

src/NUnitFramework/framework/Internal/Tests/TestAssembly.cs
```diff
@@ -39,25 +39,25 @@ public class TestAssembly : TestSuite
...
         /// <summary>
         /// Initializes a new instance of the <see cref="TestAssembly"/> class
-        /// specifying the Assembly and the path from which it was loaded.
+        /// specifying the Assembly and the name of the assembly.
         /// </summary>
         /// <param name="assembly">The assembly this test represents.</param>
-        /// <param name="path">The path used to load the assembly.</param>
-        public TestAssembly(Assembly assembly, string path)
-            : base(path)
+        /// <param name="name">The name of the assembly.</param>
```
jnm2: To be more direct about the fact that this parameter just controls a display name, could we say, "The desired name for the test suite?"

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -81,42 +81,46 @@ public ITest Build(Assembly assembly, IDictionary<string, object> options)
...
-                testAssembly = new TestAssembly(assemblyName);
+                testAssembly = new TestAssembly(Path.GetFileName(assemblyNameOrPath));
                 testAssembly.MakeInvalid(ExceptionHelper.BuildMessage(ex, true));
             }
 
             return testAssembly;
         }
 
-        private TestSuite Build(Assembly assembly, string assemblyPath, IDictionary<string, object> options)
+        private TestSuite Build(Assembly assembly, string assemblyName, IDictionary<string, object> options)
```
jnm2: I'm thinking this would be most clear as `string suiteName`, and `suiteName` should also be the name of the `BuildTestAssembly` parameter.

src/NUnitFramework/tests/Internal/SetUpFixtureTests.cs
```diff
@@ -16,7 +16,8 @@ namespace NUnit.Framework.Internal
     [TestFixture]
     public class SetUpFixtureTests
     {
-        private static readonly string testAssembly = AssemblyHelper.GetAssemblyPath(typeof(NUnit.TestData.SetupFixture.Namespace1.SomeFixture).GetTypeInfo().Assembly);
+        private static readonly string ASSEMBLY_PATH = AssemblyHelper.GetAssemblyPath(typeof(NUnit.TestData.SetupFixture.Namespace1.SomeFixture).GetTypeInfo().Assembly);
+        private static readonly string ASSEMBLY_NAME = System.IO.Path.GetFileName(ASSEMBLY_PATH);
```
jnm2: Assembly name isn't the same thing as assembly filename, but either way I'm thinking `SUITE_NAME` is what we're really testing.

src/NUnitFramework/framework/Internal/Tests/TestAssembly.cs
```diff
@@ -39,25 +39,25 @@ public class TestAssembly : TestSuite
...
 
         /// <summary>
         /// Initializes a new instance of the <see cref="TestAssembly"/> class
-        /// for a path which could not be loaded.
+        /// specifying the name of the assembly.
         /// </summary>
-        /// <param name="path">The path used to load the assembly.</param>
-        public TestAssembly(string path) : base(path)
+        /// <param name="name">The name of the assembly.</param>
+        public TestAssembly(string name) : base(name)
```
jnm2: I'd almost ask for this to be `suiteName` too for full clarity, but it seems more consistent to align with the base class's parameter name and property.

jnm2:
@jrjrguo This is super, thanks for working on it! We like to ask that the PR comment contains the GitHub keyword "Fixes #2898" or "Closes #2898" so that the issue is specially linked and [autoclosed](https://help.github.com/articles/closing-issues-using-keywords/).

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
jrjrguo: Ah yes, that makes sense since <unknown> is a special case. I'll make the change to only handle <unknown>.

jrjrguo:
@jnm2 I totally agree with your comments on renaming and comment phrasing. Keeping `name` in TestAssembly and renaming `assemblyName` to `suiteName` in BuildTestAssembly is a nice transition from name/path to a desired name. Let me know what do you think!

src/NUnitFramework/framework/Internal/Tests/TestAssembly.cs
```diff
@@ -39,25 +39,25 @@ public class TestAssembly : TestSuite
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="TestAssembly"/> class
-        /// specifying the Assembly and the path from which it was loaded.
+        /// specifying the Assembly and the name of the assembly.
```
jnm2: `and the name for the test suite` perhaps, to match the parameter documentation?

src/NUnitFramework/framework/Internal/Tests/TestAssembly.cs
```diff
@@ -39,25 +39,25 @@ public class TestAssembly : TestSuite
...
         {
             this.Assembly = assembly;
-            this.Name = Path.GetFileName(path);
+            this.Name = name;
         }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="TestAssembly"/> class
-        /// for a path which could not be loaded.
+        /// specifying the name of the assembly.
```
jnm2: This constructor is specifically used when there is no Assembly object because the load failed. Maybe a way to keep reflecting that would be `specifying a suite name for an assembly which failed to load.` or something similar?

jnm2:
@jrjrguo That's the plan! ℹ See https://github.com/nunit/nunit/issues/2898: 

image

Some of our regular reviewers are unusually busy right now.

jrjrguo:
Awesome, thanks guys!

Issue #352, opened 2014-11-10
Test with infinite loop in TearDown cannot be aborted


CharliePoole:
Pursuant to a discussion on issue #326 I added some tests in order to see the behavior of tests that caused a timeout because of a hung SetUp or a hung TearDown. An infinite loop in the SetUp is handled the same as one in the test but a loop in TearDown causes the entire run to hang.

The cause of this phenomenon is that we run TearDown out of a finally clause. Code in a finally clause is not interupted by a ThreadAbortException.

The same behavior occurs in NUnit V2. See https://github.com/nunit/nunitv2/issues/26

It appears that there are three options:
1. Don't run TearDown in a finally clause.
2. Don't use Thread.Abort to cancel a test that times out.
3. Have NUnit host the CLR and thereby gain the ability to escalate aborts.

All of these are fairly big deals, so some design work is needed.

CharliePoole:
I'm moving this to Beta2 since it's low priority. That said, we should give it a bit of focus after beta 1 to decide whether it's feasible to fix it or if we should just document it. As commented above, fixing it means calling teardowns in some other way than in a finally clause. It seems like there should be a way to do this and that it would improve the design if we did so. Finally clauses should be for _our_ cleanup, not for the users'.

rprouse:
As part of #867, I am commenting out the ignored test `TimeoutTests.TearDownTimesOutAndNoFurtherTearDownIsRun` which fails because of this issue. When this issue is fixed, that test should be uncommented.

jnm2:
This seems like low-hanging fruit. I'd be happy to move this out of the `finally`:

https://github.com/nunit/nunit/blob/13b47765b9951ae5edc3253e4f9116db5f7d542d/src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs#L69-L73

It would require replacing the `finally` with `catch (Exception ex)` and rethrowing the exception (if any) after executing the finally block.

jnm2:
Or we could just do `AfterTestCommand` and assume there will be no exception lol:

like https://github.com/nunit/nunit/blob/dd73ab001cbbdc8f911d457b976a7f982b3c11b0/src/NUnitFramework/framework/Internal/Commands/AfterTestCommand.cs#L48-L50

CharliePoole:
@jnm2 Except... I think it's being fixed with #2920, using the simplest of the approaches I suggested.

jnm2:
Oh good!

[...]

PR #2920, opened 2018-07-06
Executing AfterTest actions under a ThreadAbortException safe zone

### LINKED ISSUE #2761: Infinite loop in nunit 3.9
I am running a test suite with nunit 3.9 using nunit3-console, that randomly blocks at the end of the run. All of the tests derive from a base class that is declared with `[Parallelizable(ParallelScope.All)]` attribute. 

I managed to break into nunit and the problem is that sometimes the NonParallel worker is stuck inside the `OnEndOfShift()` method of `ParallelWorkItemDispatcher`. The `_topLevelWorkItem` state is `WorkItemState.Complete`, and after all of the shifts have been run their queues are empty. No other worker is changing the state of the `_topLevelWorkItem`. Given this situation, the following code inside `OnEndOfShift()` is an endless loop:

```
// Shift has ended but all work may not yet be done
while (_topLevelWorkItem.State != WorkItemState.Complete)
{
    // This will return null if all queues are empty.
    nextShift = SelectNextShift();
    if (nextShift != null)
    {
        ShiftStarting?.Invoke(nextShift);
        nextShift.Start();
        return;
    }
}
```

The name of the `_topLevelWorkItem` is the name of the dll containing all the tests.
I was not able to recreate the problem with a small set of tests but I can make investigations about variable values, execution stacks, or order of events, if you need info.

dukearena:
Closes #2761

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
@@ -52,11 +52,25 @@ namespace NUnit.Framework.Internal.Commands
             if (Test.Fixture == null)
                 Test.Fixture = context.TestObject;
 
-            try
+            ThreadAbortSafeRun(context, () =>
             {
                 BeforeTest(context);
-
                 context.CurrentResult = innerCommand.Execute(context);
+            });
+
+            if (context.ExecutionStatus != TestExecutionStatus.AbortRequested)
+            {
+                ThreadAbortSafeRun(context, () => { AfterTest(context); });
+            }
+
+            return context.CurrentResult;
+        }
+
+        private static void ThreadAbortSafeRun(TestExecutionContext context, Action action)
+        {
+            try
+            {
+                action();
             }
             catch (Exception ex)
             {
@@ -66,13 +80,6 @@ namespace NUnit.Framework.Internal.Commands
 #endif
                 context.CurrentResult.RecordException(ex);
             }
-            finally
-            {
-                if (context.ExecutionStatus != TestExecutionStatus.AbortRequested)
-                    AfterTest(context);
-            }
-
-            return context.CurrentResult;
         }
         
         /// <summary>
--- a/src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
@@ -52,7 +52,7 @@ namespace NUnit.Framework.Internal.Commands
             if (Test.Fixture == null)
                 Test.Fixture = context.TestObject;
 
-            ThreadAbortSafeRun(context, () =>
+            RunTestMethodInThreadAbortSafeZone(context, () =>
             {
                 BeforeTest(context);
                 context.CurrentResult = innerCommand.Execute(context);
@@ -60,13 +60,13 @@ namespace NUnit.Framework.Internal.Commands
 
             if (context.ExecutionStatus != TestExecutionStatus.AbortRequested)
             {
-                ThreadAbortSafeRun(context, () => { AfterTest(context); });
+                RunTestMethodInThreadAbortSafeZone(context, () => { AfterTest(context); });
             }
 
             return context.CurrentResult;
         }
 
-        private static void ThreadAbortSafeRun(TestExecutionContext context, Action action)
+        private static void RunTestMethodInThreadAbortSafeZone(TestExecutionContext context, Action action)
         {
             try
             {


src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
```diff
@@ -52,11 +52,25 @@ public override TestResult Execute(TestExecutionContext context)
...
+
+            if (context.ExecutionStatus != TestExecutionStatus.AbortRequested)
+            {
+                ThreadAbortSafeRun(context, () => { AfterTest(context); });
+            }
+
+            return context.CurrentResult;
+        }
+
+        private static void ThreadAbortSafeRun(TestExecutionContext context, Action action)
```
CharliePoole: Can we find a name for this that makes it clear it is only intended for running test methods under the execution context? I realize it's private to the class, but the name makes it sound more general than it really is, since it records test results in the context.

Since I don't actually have a name to suggest, you can take this as a suggestion and leave as is if you can't think of a better name. :-)

I think this may resolve #352 as well.

dukearena: What about one of these?
- `RunWithContextInThreadAbortSafeZone`
- `TestMethodRunInThreadAbortSafeZone`
- `TestMethodContextRunInThreadAbortSafeZone`
- `TestContextInThreadAbortSafeZone`

CharliePoole:
Personally, I like the `TestMethodRunInThreadAbortSafeZone` best of the four. That also suggests `RunTestInThreadAboutSafeZone`. Take your pick.

dukearena:
@CharliePoole I opted for `RunTestMethodInThreadAbortSafeZone` since it's not running a test but only one part (a method).

CharliePoole:
`<meta>`
So I approved this with a suggestion of an optional change. @dukearena made the change, causing GitHub to cancel my approval. I approved it again.

As programmers, we put up with this nonsense because we understand how hard it is for GitHub to implement an automated approval system that will mimic real people reading and acting on the comments. It would take a fairly advanced AI approach.

But there __is__ an alternative, which is to allow folks to opt out of the GitHub automated approval system entirely and use their brains!
`</meta>`

src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
```diff
@@ -52,11 +52,25 @@ public override TestResult Execute(TestExecutionContext context)
...
+
+            if (context.ExecutionStatus != TestExecutionStatus.AbortRequested)
+            {
+                RunTestMethodInThreadAbortSafeZone(context, () => { AfterTest(context); });
+            }
+
+            return context.CurrentResult;
+        }
+
+        private static void RunTestMethodInThreadAbortSafeZone(TestExecutionContext context, Action action)
```
jnm2: It took me a a few minutes to figure out what "thread-abort-safe zone" means, but it looks like it means that it's not inside a finally block. It could be called inside a finally block higher up the call stack though, in which case it would not be a thread-abort-safe zone despite the name. Am I understanding this properly?

jnm2:
@CharliePoole I found VSTS interesting. It doesn't switch the review state automatically, and it has different options:

image

CharliePoole:
@jnm2 I believe you are. 😄 

@dukearena essentially made two logical changes:

1. Removed teardown from the finally block and put it in it's own try block.

2. Encapsulated the try / catch logic in a method.

Since the method is private, it can't be called from higher up.

jnm2:
@CharliePoole It can indirectly be called higher up. If someone calls `BeforeAndAfterTestCommand.Execute` within a finally block, or any command's Execute method higher up the chain, this method will not be a thread-abort-safe zone.

jnm2:
@CharliePoole It can indirectly be called higher up. If someone calls `BeforeAndAfterTestCommand.Execute` within a finally block, or any command's Execute method higher up the chain within a finally block, `RunTestMethodInThreadAbortSafeZone` will not be a thread-abort-safe zone. It's effectively a no-op except for the catch. I think it should be named after the function of the catch, which is what it adds on top of the delegate. Would that make sense?

CharliePoole:
@jnm2 Approve with suggestions is a great improvement.

Essentially, however, I think automated approval systems are evil. They automate the process of preventing changes to the code base by people who are presumably trusted. By automating it, they make it even easier for one dev to hold up all changes by a particular person.

This is exactly the pattern that has been widely discussed lately around the issue of inclusiveness and diversity in agile teams. Of course, it's possible to arbitrarily block somebody without the aid of automation, but the problem becomes much more obvious when it has to be done personally.

jnm2:
Wow, I have a lot to learn in this area, but that seems bizarre to me! If someone is so ill-willed that they're arbitrarily blocking someone's PR, why are they allowed to stay with the project?

CharliePoole:
So a wrapping test command, either one of ours or one created as part of a custom attribute could conceivably break things by calling Execute within a finally block.

That really sounds like a very strange thing to do, however, and it's a risk I'd be willing to take, especially as it removes a problem that's been around for four years! If we allow people to modify how NUnit works with customizations, it's probably an unavoidable risk as well.

I'm OK with the current name, which is not exposed outside the class.

jnm2:
If you're okay with it then I'm okay with it, but it would help me if it had a name that explains what it does (add a catch), rather than specifying what it doesn't do (add a finally block). I'll have to look at it a little closer when I get a chance.

CharliePoole:
@jnm2 If you google "code review and diversity" you'll find a lot of stuff.

My guess is that such behavior is usually disguised. The victim who "just happens to be" a woman, TG person, etc., is painted as not careful enough, not following standards etc. Of course, the bias doesn't need to be gender-based or anything like that: it could just be a personal bias or pure competitiveness. Either way it's bad for a team.

Some people handle it through blind review, but that seems counter to a transparent team-based approach to me. IMO, it just has to be unacceptable behavior in a team and the review process has to make any systematic bias obvious to other team members. (If you are a real team, then making a problem highly visible usually leads to solving it.)

Happy to discuss more and send links offline if you're interested.

CharliePoole:
Looks like an Appveyor hang. Restarted.

dukearena:
Thanks @CharliePoole @jnm2 , I'm very happy to have contributed :)

PR #2921, opened 2018-07-07
Re-enabled teardown abort test

### LINKED ISSUE #352: Test with infinite loop in TearDown cannot be aborted

jnm2:
Closes #352 by reversing https://github.com/nunit/nunit/commit/f9b4e2d22e99731788bf7609c51ee1bf7fa927dc to bring back the test which validates the fix in https://github.com/nunit/nunit/pull/2920.

I verified that the test still hung before @dukearena's change. It now passes.

### DIFF
--- a/src/NUnitFramework/tests/Attributes/TimeoutTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TimeoutTests.cs
@@ -74,6 +74,7 @@ namespace NUnit.Framework.Attributes
         }
 
         [Test]
+        [Platform(Exclude = "Mono", Reason = "Test never aborts on Mono (tested 5.4–5.12)")]
         public void TearDownTimesOutAndNoFurtherTearDownIsRun()
         {
             TimeoutFixture fixture = new TimeoutFixtureWithTimeoutInTearDown();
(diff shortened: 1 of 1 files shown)


jnm2:
@mikkelbu How interesting! Good idea, I will add `--labels=before`. After that I might also try Mono 5.12.

jnm2:
@mikkelbu Confirmed, they both get stuck on `TearDownTimesOutAndNoFurtherTearDownIsRun` with Mono 5.10.

jnm2:
Problem is exactly the same with Mono 5.12.

jnm2:
@mikkelbu Confirmed, they both get stuck on `TearDownTimesOutAndNoFurtherTearDownIsRun` with Mono 5.4–5.12.

I excluded the test from Mono since Mono is clearly deviating from .NET Framework. I'm not sure if there'd be any ROI on looking into it unless someone reports it.

PR #2924, opened 2018-07-08
Update AssertionException-catching test to match current design

jnm2:
Followup from https://github.com/nunit/nunit/pull/2700#pullrequestreview-135191516.

Converts a TODO item to reflect that this is by design.

### DIFF
--- a/src/NUnitFramework/testdata/AssertFailFixture.cs
+++ b/src/NUnitFramework/testdata/AssertFailFixture.cs
@@ -45,5 +45,18 @@ namespace NUnit.TestData
         {
             Assert.Fail("MESSAGE: {0}+{1}={2}", 2, 2, 4);
         }
+
+        [Test]
+        public void HandleAssertionException()
+        {
+            try
+            {
+                Assert.Fail("Custom message");
+            }
+            catch
+            {
+                // Eat the exception
+            }
+        }
     }
 }
--- a/src/NUnitFramework/tests/Assertions/AssertFailTests.cs
+++ b/src/NUnitFramework/tests/Assertions/AssertFailTests.cs
@@ -110,20 +110,16 @@ namespace NUnit.Framework.Assertions
             Assert.That(assertion.Message, Is.EqualTo("MESSAGE: 2+2=4"));
         }
 
-        [Test, Ignore("Currently Fails: Must use Assert.Catch or Assert.Throws")]
-        public void CatchingAssertionExceptionMakesTestPass()
+        [Test]
+        public void TestStillFailsIfAssertionExceptionIsHandled()
         {
-            try
-            {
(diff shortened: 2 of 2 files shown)


CharliePoole:
I don't see how this is a TODO item. The test says that the current behavior us correct.

jnm2:
@CharliePoole It was a TODO item because it had [Ignore], and I converted that TODO item into a test of our intended design. Sorry for the ambiguous wording.

mikkelbu:
> I don't see how this is a TODO item. The test says that the current behavior us correct.

But before this change the test is ignored, so somewhat similar to a TODO?

CharliePoole:
If this is how y'all want it to work, then I'm cool with it. It's the way I thought it should work.

However, I left this as an Ignore because the team didn't agree with me at the time. Otherwise, I would have removed it.

CharliePoole:
This is as good a place as any to add that I think using the term TODO for Ignored tests is a bad idea. Using Ignore as a todo item is something __some__ teams do. It's a practice I have always discouraged.

Of course, teams are free to use Ignore any way they like, with any meaning they like. For the NUnit team itself, I think we have to remember that we are serving as a model of how to use NUnit, so it's best to use it as designed. The `IgnoreAttribute` was designed to temporarily mark a test that would otherwise fail, while somebody worked on other things that were needed in order to make it pass. The operative word was always temporary and the key problem is how to manage a team so that it is really only temporary.

These days, we have warnings. That's what I would tend to use if some feature was not working as expected because the work was not yet done. Others have suggested an attribute for known failing tests, so that they would succeed, but that seems like the wrong direction to go.

jnm2:
I'm confused. I thought I heard the association of Ignore = TODO from you?

CharliePoole:
I'm getting more twitter-like in some of my responses, so I may not have gotten across the short-term part. I've always taught in my NUnit classes that Ignore should be used sparsely and over short periods only. It's probably a weakness of NUnit, that keeping the number and duration down has to be done through a manual team process.

As a guideline, I've always talked about one Ignore per 1000 tests. NUnit has usually been below that except that we cheat a bit by use of `PlatformAttribute` to skip tests on Mono and/or Linux. In principle, `[Platform]` should be used to exclude tests that are __never__ intended to work on the platform, not just those that don't work currently.

Personally, I might put "Fix that ignored test" on my TODO list, but I wouldn't call the test or the attribute a TODO. This is possibly too nit-picky on my part. 😄 

Speaking of TODO's we also have a number of `// TODO` entries in the code, which have never been done. ☹️

CharliePoole:
BTW I've always argued that our docs should be about how NUnit works, not about the best way to use it. I have said that the latter belongs in books or articles about NUnit and always had the intention of writing something like that. It's on my todo list. 😃

PR #2925, opened 2018-07-10
Removed leftovers from nunit/nunit#2878

jnm2:
@CharliePoole, micro-PR 😁

Thanks to @oznetmaster for [spotting](https://github.com/nunit/nunit/commit/d09def3521dca9be6fbd84fc483c555c4955afe3#r29649322).


OmicronPersei:
Do we / should we wait for VSTS to complete successfully before merging?

PR #2919, opened 2018-07-06
SetArgDisplayNames

jnm2:
Closes #2536.

PR #2927, opened 2018-07-11
Fixed clerical errors in nunit/nunit#2919

jnm2:
Another clerical error spotted thanks to @oznetmaster.

### DIFF
--- a/src/NUnitFramework/framework/Internal/TestParameters.cs
+++ b/src/NUnitFramework/framework/Internal/TestParameters.cs
@@ -32,8 +32,6 @@ namespace NUnit.Framework.Internal
     /// </summary>
     public abstract class TestParameters : ITestData, IApplyToTest
     {
-        #region Constructors
-
         /// <summary>
         /// Default Constructor creates an empty parameter set
         /// </summary>
@@ -94,10 +92,6 @@ namespace NUnit.Framework.Internal
             Array.Copy(args, Arguments, numArgs);
         }
 
-        #endregion
-
-        #region ITestData Members
-
         /// <summary>
         /// The RunState for this set of parameters.
         /// </summary>
@@ -134,10 +128,6 @@ namespace NUnit.Framework.Internal
         /// </summary>
         public IPropertyBag Properties { get; }
 
-        #endregion
-
-        #region IApplyToTest Members
-
         /// <summary>
         /// Applies ParameterSet values to the test itself.
         /// </summary>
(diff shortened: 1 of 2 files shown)


jnm2:
Found another clerical error, okay to go with this lopsided branch name?

jnm2:
@oznetmaster Regions wrapping a single member feel tedious to me, and my personal preference would be to remove them all. But if you feel strongly that we should have them, I can do that.

oznetmaster:
I have never been a big fan of regions.  Not sure how useful they are.  If we are going to continue to use them, then this should be changed for consistency and clarity.  If not, then lets remove them in every touched source file.

Whether the region contains a single member or one hundred members should not be relevant.

jnm2:
I'm happy with that. Especially the latter option. 😃

oznetmaster:
Sounds like we need a (hopefully) brief discussion on the issue of regions.  For now, I think that it should be added as a new region and the previous one restored.  Once a consensus has been reached, changes can be made as appropriate.

oznetmaster:
I see you went the other direction :)  Not sure this should be a unilateral decision?

oznetmaster:
Now, I wonder if regions around interface implementations is perhaps a useful thing?

This is why I was suggesting some discussion before actual implementation...

jnm2:
I implement interfaces fairly frequently and have only used regions if I had an unwieldy number of explicit implementations to get out of the way. I'm happy to wait for discussion on this though. @nunit/framework-team and @rprouse?

mikkelbu:
I'm not a fan of regions (and usually, I've seen them used as an attempt at structuring a class that do too much), also I don't feel they provide much help when reading the code, neither on screen or on paper, so I don't mind if we remove them.

ChrisMaddock:
Yup - I’m all for getting rid.

rprouse:
I used to be a huge fan of regions, but these days with better editors and navigation I find that they get in the way more than they help so I avoid them. I also hate how adding an `#if` causes all regions in a file to get un-indented. Yet another strike against them.

I don't think we should do a massive removal, but I'm also reluctant to add more. It sounds like most of the team is no longer in favor of them either. Shall we update the coding standard to reflect that or does anyone feel otherwise?

jnm2:
I'm more in favor of building up a repo-specific coding standard than adding to the central one, unless uniformity across repos is very important.

oznetmaster:
I am not proposing massive removals.

Just remove them when a source file is touched for some other reason.

Issue #281, opened 2014-10-19
XSD for Version 3 output format


oznetmaster:
Is there an XSD for the 3.0 output format?  If not, should there be one?

ravichandranjv:
Unable to attach file for some reason.

The suggestion to use a tool to generate the XSD from the XML file is not so simple because there are some restrictions that need to be taken care of in the output XML file as specifying only certain values to be present for some elements.

@CharliePoole The "Description" value is missing in the TestResult format described in the github so the below XSD is still not complete.. 

Plus, there are many other maintenance related points like global or local change impact of the schema in future that may impact the XML validation, which should be discussed here. 

This schema I have tested for only two elements - **TestRun** (test-run) and **Environment** but still would like someone to test if the schema validates correctly.

Assumption: the root element is named TestRun without the '-' as in the below XML, which is a modified file of the actual TestResults.xml. 

**The XML**

<?xml version="1.0" encoding="utf-8" standalone="no"?>
<TestRun xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.nunit.org testresults.xsd">
id="2" testcasecount="2" total="2" passed="2" failed="0" inconclusive="0" skipped="0" asserts="0" start-time="2015-10-07 13:30:06Z" end-time="2015-10-07 13:30:07Z" >
<environment nunit-version="3.0.5715.30860" clr-version="4.0.30319.42000" os-version="Microsoft Windows NT 6.1.7601 Service Pack 1" platform="Win32NT" cwd="E:\nunit3_tests" machine-name="JV-PC" user="Jv" user-domain="Jv-PC" culture="en-US" uiculture="en-US" os-architecture="x86" />
</TestRun>

**The XSD**

<?xml version="1.0"?>
<xsd:schema xmlns:xsd="http://www.w3.org/2001/XMLSchema"
 xmlns:nunit="http://www.nunit.org"
 elementFormDefault="unqualified"
 attributeFormDefault="unqualified">
<xsd:element name="TestRun" />
 <xsd:complexType name="etr">
	<xsd:sequence>
	 <xsd:element ref="Environment" minOccurs="1" maxOccurs="unbounded" />
	 <xsd:element ref="TestSuite" minOccurs="1" maxOccurs="unbounded" />
	 <xsd:element ref="CommandLine" minOccurs="1" maxOccurs="unbounded" />
	 <xsd:element ref="Settings" minOccurs="1" maxOccurs="unbounded" />
[...]

Issue #270, opened 2014-10-14
OneTimeSetUp Failure results in too much output


CharliePoole:
When OneTimeSetUp fails for a fixture, the console reports (both nunit-console and nunitlite) list the failure together with any message and stacktrace. Then it lists every test case within that fixture as failing, using the message from the setup failure but without a stack trace.

This is too much info. Obviously, what has to be fixed is the failure in OneTimeSetUp.

We should continue to mark the individual cases as failed and count them in our tally of failed cases, but they don't need to be listed in the output.

Up for discussion:

Should the failed test case have a message at all? The message takes up space in the XML and says the same thing as the actual failure. The test case will have attributes result="Failed" and site="Parent" to distinguish it. OTOH, having some kind of message may help in transforming the XML if the transform is only looking at the <test-case> elements. If we are unsure, we can leave it for now and ask for feedback in the next alpha.

Issue #13, opened 2013-10-19
Standardize commandline options for nunitlite runner


CharliePoole:
The NUnitLite integrated runner supports some of the same features as the full nunit-console but uses different options to access them. To the extent possible, where the features are common, we should use the same options in both programs.

Ideally, NUnitLite commandline option processing should be converted to use Mono.Options at the same time. However, it turns out that mono.options won't compile in the compact framework, making it unsuitable for NUnitLite.

We'll need to give more thought to a common approach to commandline interpretation across all nunit projects.

From https://bugs.launchpad.net/nunit-3.0/+bug/1225544

PR #2929, opened 2018-07-13
Added NUnit XML schemas

jnm2:
Covers the framework side of https://github.com/nunit/nunit-console/issues/13.
Also https://github.com/nunit/docs/issues/270.

Fair warning: I haven't touched XSD since ~2013.

If you're validating the framework's load/explore results, you'd use Test.xsd.
For the framework's run results, TestResult.xsd.
For an NUnitLite run, NUnitLite-Run.xsd. This will probably be largely copied to the console repository.
Since the test filter format is driven by the code shipped in the NUnit package, not NUnitLite, that resides in its own TestFilter.xsd.

Once this is merged, I'll update the wiki (https://github.com/nunit/docs/issues/269) while this is still fresh in my mind. Questions, maybe for future PRs:

1. I did not add CI tests. Are we interested in that? I manually validated using Visual Studio and TestResult.xml from the bin folder. I played with it a bit and everything I've tried validates (and has cool intellisense).

2. Do we want to eventually put these on nunit.org similar to http://nunit.org/files/testresult_schema_25.txt? Maybe host the XSD at the url `https://nunit.org/schemas/TestResult` etc, in case we ever use that as an XML namespace? (All four files would have to be in the same namespace for our current test run XML though, because there can only be one unqualified namespace.)

3. I like the comments for easy reference to the source. Would we want to leave them there, or have a post-process step where we lint the XML to publish from the repo?

4. Do we want to distribute these in the package .zip or .nupgs? Could NuGet be a way to include the framework-driven XSD in other engine repo?

/cc @rprouse, @OsirisTerje, @oznetmaster


jnm2:
@pammann-work: Could you comment on whether the proposed XML schema definitions would be useful to you in this form?

rprouse:
CI failed installing .NET Core. Restarted.

rprouse:
CI failed again with "The command "sudo apt-get install dotnet-sharedframework-microsoft.netcore.app-1.1.6" failed and exited with 1 during ." We should give it some time before restarting again.

The XSDs look good as far as I care to read them 😄 

I like the idea of hosting the schemas at https://nunit.org/schemas and using that for the schema uri. If we are going to do that, keeping them in sync with this repo could be error prone. If that is the case, should they be maintained there? Future unit tests could pull from that repo.

As for packaging the schemas in the nuget packages, would anybody find that useful? What would they be used for?

jnm2:
Thinking of the sync issue, I'm a bit leery now of introducing an XML namespace unless we intend all new schemas to be backwards-compatible. A namespace would kinda conflict with the disclaimer I put at the top of each of the files. But we could still host at that URL.

I was thinking of the concept of maintaining them as close as possible to the XML-generating source code so they can be updated in the same PR.

If we put it in a nupkg, perhaps the console repo's `Console-Run.xsd` could import `TestResult.xsd` and `TestFilter.xsd` from the associated nupkg so that the schema is updated in sync with the framework dependency?

All of this is just brainstorming. 😃

ChrisMaddock:
Nice one, Joesph! 😄 This IS a long standing issue!

Agreed it would be good to have some testing in, and keep this tied to code changes. It’s ripe for becoming outdated!

Some thoughts - logically the TestResult from the framework unit tests should contain a lot of the xml content. Maybe we should validate that every CI run, and include a couple of sample ‘failed results’ in the repo to also test? Not too sure on the best approach here...

Happy to merge this as is however though, and handle these bits later on. (I haven’t looked to be xsd itself, but Robs approved!)

CharliePoole:
In V2, we had tests that validated the output XML against the XSD.

jnm2:
Heh, caught a missed element by running against the result XML for mock-assembly.

jnm2:
Added tests checking the validity of each schema and also checking whether the framework and runner output match the schemas.

@CharliePoole: That's the very next thing on my list!

jnm2:
Ah, got a question. https://github.com/nunit/docs/wiki/Test-Filters states that the root element must be `<filter>`, but today the root `<filter>` element is inserted by the runner and not the framework. `ITestFilter.ToXml` outputs a `<dummy>` root element, not a `<filter>` element., so I excluded the root `<filter>` element from the schema file.

Inconsistently, `TestFilter.FromXml` requires the `<filter>` root node, meaning that `ToXml`/`FromXml` does not roundtrip, but thankfully `TestFilter` is in the `Internal` namespace.

Should `<filter>` be a required root node in TestFilter.xsd, an optional root node, or be left as-is (disallowed, except to represent our special-cased EmptyFilter)?

jnm2:
Oh goody, my new TestFilter.xsd causes an internal .NET framework NullReferenceException, but only on 2.0 and 3.5:

```
System.NullReferenceException : Object reference not set to an instance of an object.
   at System.Xml.Schema.XmlSchemaValidator.CheckElementValue(String stringValue)
   at System.Xml.Schema.XmlSchemaValidator.ValidateAtomicValue(String stringValue, XmlSchemaSimpleType& memberType)
   at System.Xml.Schema.XmlSchemaValidator.InternalValidateEndElement(XmlSchemaInfo schemaInfo, Object typedValue)
   at System.Xml.XsdValidatingReader.ProcessEndElementEvent()
   at System.Xml.XsdValidatingReader.Read()
```

jnm2:
Here are the gory details: https://social.msdn.microsoft.com/Forums/en-us/53be44de-30b2-4d18-968d-d3414d0783b1
Updated to exclude the buggy frameworks. We only really need to test the XSD on one (working) framework, since the XSD itself transcends a particular framework.

jnm2:
Another inconsistency with the docs. Marking [wip] until I'm finished thoroughly comparing them.

jnm2:
Got everything squared away with https://github.com/nunit/docs/wiki/Test-Filters. Only thing that remains is to update it with new filter element types.

Next question: is the `test-run` format supported to be framework-driven or runner-driven?

Here it's represented to be framework-driven: https://github.com/nunit/docs/wiki/Test-Result-XML-Format

If we want this format to be runner-driven, we should document the `<test-run>` element in a separate document for NUnitLine and NUnit Console. If we want it to be framework-driven, then runners have no freedom here and we *should* move the code from NUnitLite and NUnitConsole which generates `<test-run>` back into the framework, I think.

@nunit/framework-team Need input!

ChrisMaddock:
This sounds to me like the perfect reason to have an xsd. 😄 

In my mind, the runners should generate the test-run element to a 'standard'. (test-runs contain multiple assemblys - I guess we could have functionality in the framework where we pass in multiple assembly results to be combined into a test-run, but it seems a little convoluted...)

I think the XSD should define the format of test-run, as this should be consistent regardless of runner. The runner should remain responsible for generating the test-run node however - this is a little duplication, but the functionality appears to logically live in the runner, in my mind at least! 

This would also be a great used of the xsd - we currently have three (active) duplicates of this code, in NUnit Console, NUnitLite and nunit.xamarin. If each of these was able to validate the results to writes against an xsd, that should help avoid any inconsistencies! 

The only problem that seems to leave to me is the issue that it's not possible to get a 'complete' xml file from purely the framework. That's not too much of a concern in my mind - we've discussed elsewhere how that Framework API should remain low-level.

yetibrain:
In the <test-run> element, there is an attribute named testcasecount. Why is there no assemblycount attribute in case the test ran with multiple assemblies?

jnm2:
@yetibrain From what I could infer, `testcasecount` is redundant info which can be gathered by scanning the XML, as would be `assemblycount`. It's a matter of convenience. So if there was demand for `assemblycount` I'm sure we'd consider it.

yetibrain:
@jnm2 I see. It can be obtained by an appropriate xpath expression using count.

jnm2:
@ChrisMaddock This is great reasoning. I merged NUnitLite-Run.xsd into TestResult.xsd.

Another question: How important is validating which element is the root element?
Right now the possible root elements for XML that will validate against TestResult.xsd are:

- `<test-run>`
  (the only valid root according to the spec)
- `<test-suite>`, `<test-case>`
  (would complicate the XSD to disallow, and then to validate the framework's run results, we'd need to wrap them with dummy elements)
- `<filter>`
  (would complicate the XSD to disallow)

By complicate, I mean we'd need a new Shared.xsd to contain only group and type definitions and then each top-level XSD would declare only a single element at the root level.

ChrisMaddock:
I feel like `<test-run>` is the only valid root. 🤔 And I think it's potentially one of the most important things to validate, as it's easy to get wrong - from memory, nunit.xamarin existed with a `<test-suite>` root element for a while.

Is the complication caused by wanting to also validate the framework xml independently, which would have a `<test-suite>` root? If that's the case, should we make any validation part of the engine? That would cover everything except NUnitLite - which could be treated as a special-case, in that our overall architecture generally intends runners to make use of the engine.

jnm2:
Sounds good.

> Is the complication caused by wanting to also validate the framework xml independently, which would have a <test-suite> root? If that's the case, should we make any validation part of the engine? 

No, it was caused by me avoiding a `Shared.xsd` and instead importing one "top-level" .xsd into another. I liked your argument that the format is framework-driven and the XSD which binds the behavior of the various runners belongs in one central place.

jnm2:
New question! Given this disclaimer at the top of every .xsd:

```xml
  <xs:annotation>
    <xs:documentation>This is document is descriptive, rather than a guarantee of future decisions.</xs:documentation>
  </xs:annotation>
```

Is it worth descriptively restricting the `<test-run>` `id` to be `2` as https://github.com/nunit/docs/wiki/Test-Result-XML-Format#test-run points out?

> The unique ID of this test. For the test-run, it is hard-coded as "2".

If not, I'd really like to remove it from the wiki since we wouldn't want people depending on the value `2`.

jnm2:
@nunit/framework-team All root elements are now enforced and I finished comparing with the documenting wiki pages. I think this is ready for review.

Specifically: are you happy with the way the now-five .xsd files are factored?

src/NUnitFramework/framework/Schemas/TestResult.xsd
```diff
@@ -0,0 +1,232 @@
...
+          <xs:attribute name="os-version" type="xs:string" use="required" />
+          <xs:attribute name="platform" type="xs:string" />
+          <xs:attribute name="cwd" type="xs:string" use="required" />
+          <xs:attribute name="machine-name" type="xs:string" />
+          <xs:attribute name="user" type="xs:string" />
+          <xs:attribute name="user-domain" type="xs:string" />
+          <xs:attribute name="culture" type="xs:string" use="required" />
+          <xs:attribute name="uiculture" type="xs:string" use="required" />
+          <xs:attribute name="os-architecture" type="xs:string" use="required" />
+        </xs:complexType>
```
ChrisMaddock: I _think_ the xamarin runner uses slightly different fields in its top level test-run properties? Thats possibly the wrong thing to do, but if we’re going to publish this as ‘what will exist’ - I think we should check. (On my phone right now, sorry!)

src/NUnitFramework/nunitlite.tests/SchemaTests.cs
```diff
@@ -0,0 +1,73 @@
...
+using NUnit.TestUtilities;
+
+namespace NUnitLite.Tests
+{
+    public static class SchemaTests
+    {
+        [Test]
+        public static void TestResultSchemaIsValid()
+        {
+            Assert.Multiple(() => SchemaTestUtils.AssertValidXsd("TestResult.xsd"));
```
ChrisMaddock: Does this path need to be qualified with TestDirectory - for when WorkingDirectory and TestDirectory are different?

ChrisMaddock: (Same comment in other places!)

src/NUnitFramework/SchemaTestUtils.cs
```diff
@@ -0,0 +1,99 @@
...
+namespace NUnit.TestUtilities
+{
+    internal static class SchemaTestUtils
+    {
+        public static string GetSchemasPath() => Path.Combine(TestContext.CurrentContext.TestDirectory, "Schemas");
+
+        public static XmlSchema AssertValidXsd(string schemaFile)
+        {
+            using (var testXsd = File.OpenRead(Path.Combine(GetSchemasPath(), schemaFile)))
+            {
```
ChrisMaddock: Ah - this is why my previous comment isn’t an issue - it’s done at this level instead. 👍🏻

jnm2:
Oops! Right after merging I saw I forgot to follow up on @ChrisMaddock's comment:

> I _think_ the xamarin runner uses slightly different fields in its top level test-run properties? Thats possibly the wrong thing to do, but if we’re going to publish this as ‘what will exist’ - I think we should check. (On my phone right now, sorry!)

Great thought! I'll start a new PR. Before I do that, questions for @nunit/framework-and-engine-team:

1. Should we make arbitrary attributes legal on the `<test-run>` element (or other elements)? The Xamarin runner adds a `xamarin-runner-version` which will fail to validate unless it is explicitly and separately namespaced. Or should we require metadata like this to be in a `<properties>` os similar structure which we'd invent for the purpose?

2. These `<test-run>` attributes are missing in the Xamarin runner output: `random-seed`, `name`, `fullname`. The `<filter>` child element is also missing. Should we make these optional in the XSD or should be have the Xamarin runner output them?

CharliePoole:
Background to help you decide...

NUnit has always used loosely defined XML fragments. Where we read XML files, we generally use a default when an attribute is missing. We explicitly ignore any elements or attributes that we don't recognize, allowing third parties to add their own elements. Filter element is a great example. If no filter was specified, then there could be no filter element or an empty element.

This effort at standardizing all the XML using schemas could be viewed as changing all that and making the XML definition very rigid. Or you might just standardize it a little, if that's what you want to do. My own thought, not being part of the current push for schemas, is that it's probably a mistake to tie down too much those bits of XML that don't go outside of NUnit, but are used only to communicate between the framework and engine.

ChrisMaddock:
@Charlie - this PR is only about standardising the end test run result, isn’t it - which is external to nunit? Of course, the external xml is closely linked to the internal xml.

@joseph
1. I think allowing arbitrary additional attributes is sensible. I think it’s good to specify what’s expected to definitely be there - and allow additions - especially if we’re thinking in terms of future features. Xml is ‘extensible’, after all 😉

2. The long term goal for the xamarin runner is to use the engine to generate the xml, so I suggest our approach is to make xamarin consistent with the engine, whilst allowing features such as xamarin-runner-version which are probably useful to be additional. 😊

CharliePoole:
@ChrisMaddock I thought I had seen some discussion about other stuff, but maybe it isn't on this PR.

jnm2:
@ChrisMaddock There are a couple of confused GitHubbers right now that'll be following our conversation 😂

You would have extensibility either way by namespacing the extending attribute, but I don't think any of us is interested in tangling with locking it down like that! I agree.

Should I go through and explicitly allow any attributes on all elements or just test-run for now?

src/NUnitFramework/framework/Schemas/TestFilterDefinitions.xsd
```diff
@@ -0,0 +1,78 @@
...
+      <xs:element name="cat" type="ValueMatchFilterType" />
+
+      <!-- NUnit.Framework.Internal.Filters.ClassNameFilter.ElementName -->
+      <xs:element name="class" type="ValueMatchFilterType" />
+
+      <!-- NUnit.Framework.Internal.Filters.FullNameFilter.ElementName -->
+      <xs:element name="test" type="ValueMatchFilterType" />
+
+      <!-- NUnit.Framework.Internal.Filters.IdFilter.ElementName -->
+      <xs:element name="id" type="ValueMatchFilterType" />
```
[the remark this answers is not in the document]
jnm2: I see:

https://github.com/nunit/nunit/blob/f609611e5fc3980078376615fc6b016cc6cac1c2/src/NUnitFramework/framework/Internal/Filters/IdFilter.cs#L46-L48

The current behavior of the framework is to ignore the `IsRegex` property inherited from `ValueMatchFilter`. It doesn't error if it is set. I'm not opposed to making it invalid XML so long as we make sure the framework is changed to prevent `IdFilter.IsRegex` from being written out into the XML.

CharliePoole:
It's not clear here if you are talking about the representation of the filter in use for a run, which appears in the XML result file, or the representation that is sent to the framework to select tests. If a user assumes that what they see in the result file is a valid format to send to the framework, they could be mistaken. That's why we provide factory methods for creating filters.

It's likely that I caused the confusion here by over-documenting current behavior in the wiki. 😞 However, it seems to me that an XSD should not describe current behavior, but the limits of acceptable behavior. In the case of a filter used with a test run, it could be any string representation that the framework (any framework) is able to accept. For example, some framework might accept filter expressions as used by Visual Studio - for that matter, a future NUnit framework could easily accept them. In that case, I'd expect the filter in the result file to reflect that.

jnm2:
@CharliePoole 

>  If a user assumes that what they see in the result file is a valid format to send to the framework, they could be mistaken. That's why we provide factory methods for creating filters.

My ears couldn't help but perk up because of https://github.com/nunit/nunit/issues/2931. We don't have factory methods for users, do we?

jnm2:
@CharliePoole 

>  If a user assumes that what they see in the result file is a valid format to send to the framework, they could be mistaken. That's why we provide factory methods for creating filters.

My ears couldn't help but perk up because of https://github.com/nunit/nunit/issues/2931. We don't have factory methods for users, do we? 

> It's not clear here if you are talking about the representation of the filter in use for a run, which appears in the XML result file, or the representation that is sent to the framework to select tests.

We're using the same XSD to validate both, and I imagine we'd want to use the same wiki page for both. In this case, the question is whether `re="1"` should appear on `<id>` since the ignores the attribute by design.

CharliePoole:
The engine provides a service for use in creating filters. It is, however, a limitation of the engine that it only produces filters for the NUnit 3 framework. It's up to drivers for other frameworks to translate the provided filter to something understandable by the framework. In any case, users of that service don't need to know the format of a filter.

Issue #2931 is about providing some factory methods for those who do not want to use the engine. By definition, such users are only concerned with a single framework but the engine is concerned with more than that. In general, users of the engine only need to know that there is some sort of representation that is sent to the framework.

Where are we validating the filters sent to the framework? I thought  you were only validating the output result file?

Regarding `<id re="1" ...>` : where do we generate that? It's an error if we do and we should fix it.

jnm2:
> Where are we validating the filters sent to the framework? I thought you were only validating the output result file?

The framework itself validates nothing. We have tests validating sample test run, sample test load, sample test exploration, and sample test filter XML.

I'm not sure what it would take to make us generate `<id re="1" ...>` besides writing `new IdFilter("x") { IsRegex = true }.AddToXml(...)`. I only know, as I said, that the framework ignores it. Do we want to lock down the XSD? If so, it would be nice to have IdFilter stop inheriting from ValueMatchFilter. (Which would be nice anyway.)

src/NUnitFramework/framework/Schemas/TestFilterDefinitions.xsd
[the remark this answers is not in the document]
mikkelbu: I think I got it backwards (that the framework cannot parse e.g. `Id != ...` or `Id =~ ...` and that it would not generate a `<id re="1">...` element as the code is now).

I think we should keep the schema as it is. Then we can always improve on it later on, if we agree to enforce a more strict schema.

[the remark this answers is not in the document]
jnm2: Okay, sounds good.

CharliePoole:
So where are the tests of "sample test filter XML?" I'm trying to understand.

No opinion on what you should   "lock down" in the filter XSD. Locking down in that case doesn't seem to have the same meaning as locking down the outputs consumed by users. I'm not clear that it has any operational meaning for users.

jnm2:
Sorry, here:

https://github.com/nunit/nunit/blob/f609611e5fc3980078376615fc6b016cc6cac1c2/src/NUnitFramework/tests/Api/SchemaTests.cs#L194-L238

CharliePoole:
So, correct me if I'm wrong, but those appear to be testing the XSD itself by using it to verify certain hard-coded XML strings that either pass or fail validation. That's a good unit test for an XSD.

I was expecting some kind of test of the NUnit code, as we did for NUnit V2. That is, run some tests, look at the output and validate that it meets the XSD.

But for that kind of test, you need to have output produced by NUnit. The XSD then serves as both a test NUnit and as documentation for what NUnit produces. I don't see how you can have that sort of test in the framework for filters, since the framework consumes filters rather than producing them.

jnm2:
@CharliePoole Yes, this is what was indented. Should we be going further?

jnm2:
@CharliePoole Yes, this is what was intended. Should we be going further?

CharliePoole:
I'm probably not the right person to comment. I see value in an XSD used to test and support the code. The filter XSD seems to have no such purpose so it's not clear to me how it is useful.

ravichandranjv:
I managed to trace the XSD issue to "XSD for Version 3 output format" [#281](https://github.com/nunit/nunit/issues/281) opened in 2014 (and not [#270](https://github.com/nunit/docs/issues/270), which iis related to documentation) and got moved to nunit-console [#13](https://github.com/nunit/nunit-console/issues/13) ! 

Reading through I got a little confused because of Test-Filter.xsd. 

Should it not be part of or as a separate issue or is the confusion only caused by the title of **this** issue, which simply says "Adding XML schemas to nunit" ? 

@jnm2 - why would one want to test if a schema is valid?
[Test] 
         public static void TestFilterSchemaMatches() 
         { 

Maybe something like **XSD for Test-Result output format** would narrow down the context? 

Because from its origin, if I recall correctly, the requirement was specifically for validating the output XML against some XSD because of some css or some reporting tool's requirement that could use a testresult.xml as source data.

CharliePoole:
I highly recommend everyone in the discussion reading @ronjeffries latest article: https://ronjeffries.com/articles/018-01ff/no-estimates-logic/

Don't be confused by the title, it's really about how a project decides whether some artifact or activity is waste and, if it is, whether it's necessary waste. I think you'll find a few sentences in there where you can substitute "XSD" without spoiling the overall intent.

We need to be hard-nosed about defining what is waste. I learned from Kent Beck that __even tests__ are waste, but __necessary__ waste since we don't know any better way to build good working code. He compares tests to the scaffolding that is used to build a building. When it's done, we tear the scaffolding down. (IIRC, he said this at the Sheffield XP Conference in 2005, but don't hold me to it.)

Anyway, __some__ XSD files may be actual __product__, for example the one that describes our test result file for consumers that need an XSD. Others may serve as scaffolding. Yet others may be something we are doing out of pure reflex or because somebody told us you should always have an XSD. I think it would be good to dig deep and figure that out.

jnm2:
@CharliePoole 

> I'm probably not the right person to comment. I see value in an XSD used to test and support the code. The filter XSD seems to have no such purpose so it's not clear to me how it is useful.

The purpose of this XSD is documentation to assist those that want to integrate with NUnit framework. That's why I added this in each file:

```xsd
  <xs:annotation>
    <xs:documentation>This is document is descriptive, rather than a guarantee of future decisions.</xs:documentation>
  </xs:annotation>
```

The XSD is downstream from the framework; if they conflict, the default assumption is that the XSD is wrong, not the framework's output.

PR #2934, opened 2018-07-17
Ensures that exported types are not within System.* namespace.

### LINKED ISSUE #2883: Our public ConcurrentQueue causes type conflicts
If you try to use NUnit on net35 with another library such as AsyncBridge which ships a `ConcurrentQueue` polyfill, the project won't compile because of the type collision between the two dependencies.

This began in 3.10. We should make sure all our polyfill classes are internal since they are sharing the System namespace.

The same problems occur with NUnit.System.Linq, and `System.Action` etc, but I'm not sure we can do anything about that because it has shipped with NUnit for so long. If we kept it in the nupkg copying to the test project output folder but kept the test project from referencing it, it would solve the problem but it might also break people who have come to rely on NUnit.System.Linq always coming with NUnit.

OmicronPersei:
Closes #2883 .

### DIFF
--- a/src/NUnitFramework/tests/Internal/NamespaceTests.cs
+++ b/src/NUnitFramework/tests/Internal/NamespaceTests.cs
@@ -1,0 +1,40 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using NUnit.Framework.Api;
+
+namespace NUnit.Framework.Internal
+{
+    [TestFixture]
+    public class NamespaceTests
+    {
+        [Test]
+        public void AllExportedNameSpacesAreNotSystem()
+        {
+#if !NETCOREAPP1_1
+            var exportedTypes = Assembly.GetAssembly(typeof(FrameworkController)).GetExportedTypes();
+#else
+            var exportedTypes = typeof(FrameworkController).GetTypeInfo().Assembly.GetExportedTypes();
+#endif
+
+            var exportedTypesWhitelist = new[] 
+            {
+                typeof(System.Web.UI.ICallbackEventHandler)
+            };
+
+            Assert.Multiple(() =>
+            {
+                foreach (var type in exportedTypes.Except(exportedTypesWhitelist))
+                {
+                    Assert.IsFalse(
+                        type.Namespace.StartsWith("System", StringComparison.CurrentCultureIgnoreCase),
+                        $"Type {type.FullName} is in the System namespace but should not be.");
+                }
+            });
+
+
+        }
+    }
+}
--- a/src/NUnitFramework/framework/Compatibility/System.Collections.Concurrent/ConcurrentQueue.cs
+++ b/src/NUnitFramework/framework/Compatibility/System.Collections.Concurrent/ConcurrentQueue.cs
@@ -43,7 +43,7 @@ namespace System.Collections.Concurrent
     [DebuggerTypeProxy(typeof(SystemCollectionsConcurrent_ProducerConsumerCollectionDebugView<>))]
     [HostProtection(Synchronization = true, ExternalThreading = true)]
     [Serializable]
-    public class ConcurrentQueue<T> : IProducerConsumerCollection<T>//, IReadOnlyCollection<T>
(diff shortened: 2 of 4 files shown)


src/NUnitFramework/tests/Internal/NamespaceTests.cs
```diff
@@ -0,0 +1,40 @@
...
+    public class NamespaceTests
+    {
+        [Test]
+        public void AllExportedNameSpacesAreNotSystem()
+        {
+#if !NETCOREAPP1_1
+            var exportedTypes = Assembly.GetAssembly(typeof(FrameworkController)).GetExportedTypes();
+#else
+            var exportedTypes = typeof(FrameworkController).GetTypeInfo().Assembly.GetExportedTypes();
+#endif
```
ChrisMaddock: I'm not sure - but does the second version of this not work on all platforms? 🙂

ChrisMaddock: Just looked deeper, looks like we have an extension method implementation of GetTypeInfo() in ReflectionExtensions.cs to help with this case? I thought the idea of that is that it would allow us to use the same API across all platforms - I’m not 100% sure if it covers this case or not though, sorry! 😊

ChrisMaddock: If I'm wrong, lets get this merged. 😄

src/NUnitFramework/tests/Internal/NamespaceTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        public void AllExportedNameSpacesAreNotSystem()
+        {
+            var exportedTypes = typeof(FrameworkController).GetTypeInfo().Assembly.GetExportedTypes();
+
+            var exportedTypesWhitelist = new[] 
+            {
+                typeof(System.Web.UI.ICallbackEventHandler)
+            };
+
+            Assert.Multiple(() =>
```
jnm2: I'm increasingly enjoying this approach.

src/NUnitFramework/tests/Internal/NamespaceTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                foreach (var type in exportedTypes.Except(exportedTypesWhitelist))
+                {
+                    Assert.IsFalse(
+                        type.Namespace.StartsWith("System", StringComparison.CurrentCultureIgnoreCase),
+                        $"Type {type.FullName} is in the System namespace but should not be.");
```
jnm2: 50: Names used to address code elements should not be culture-dependent. I'd use `Ordinal` or `OrdinalIgnoreCase`.

51: I wouldn't say that the affected types should not be in the System namespace but rather that they should not be publicly visible.

src/NUnitFramework/tests/Internal/NamespaceTests.cs
jnm2: I've been starting to think we should honestly put our TypeExtensions in System.Reflection so that we don't need both namespaces imported. Or maybe namespaceless. It's so finicky.

ChrisMaddock: Great - thanks for that! I wasn't sure myself! 😅

src/NUnitFramework/tests/Internal/NamespaceTests.cs
```diff
@@ -47,8 +47,8 @@ public void AllExportedNameSpacesAreNotSystem()
                 foreach (var type in exportedTypes.Except(exportedTypesWhitelist))
                 {
                     Assert.IsFalse(
-                        type.Namespace.StartsWith("System", StringComparison.CurrentCultureIgnoreCase),
-                        $"Type {type.FullName} is in the System namespace but should not be.");
+                        type.Namespace.StartsWith("System", StringComparison.OrdinalIgnoreCase),
+                        $"Type {type.FullName} is publically visible in the System namespace but should not be.");
```
jnm2: The typical spelling is `publicly`.

OmicronPersei: Gah!

Issue #2911, opened 2018-06-26
Output of all runs of tests is not visible when Retry attribute is used


oviradoi:
In NUnit versions 3.5 and older, if you had a test with a Retry attribute, the output from all runs of the test would be printed to the console.

Starting from NUnit 3.6, only the output of the last run of the test will be written to the console.

For example, let's take this test:

    [TestFixture]
    class Fixture
    {
        private static int _times = 1;

        [Test, Retry(3)]
        public void Test()
        {
            TestContext.WriteLine("Retry nr " + _times);
            _times++;
            Assert.Fail();
        }
    }

In NUnit versions 3.5 and lower, it would print:

    Retry nr 1
    Retry nr 2
    Retry nr 3

But in NUnit versions 3.6 and newer, it prints:

    Retry nr 3

Is there any way to also get the output of the previous runs to be written to the console?

CharliePoole:
Prior to 3.6, standard output would go directly to the runner. In the case of the console runner, it all gets displayed on the console.

With 3.6, standard output becomes part of the test result. Since `[Retry]` discards all the results except for the final one, that output is discarded.

If you want a message to appear even if the the result is discarded, use `TestContext.Progress`.

oviradoi:
Thanks, using `TestContext.Progress` worked for us.

PR #2944, opened 2018-07-26
Concatenate retried tests TestContext Output

kallekrantz:
Hi all,

So we recently discovered that somewhere between nunit 3.0.0 and the current version of nunit, test output was no longer kept while using retries (instead only the final test output was printed). I decided to dig in and fix it immediately instead of filing an issue, hope that's okay!

As this is the first time contributing, I'm more than happy to accept feedback or proposals for changes. Thanks in advance!

### DIFF
--- a/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RetryAttribute.cs
@@ -109,8 +109,12 @@ namespace NUnit.Framework
                     // Clear result for retry
                     if (count > 0)
                     {
+                        var carryOverTestOutput = context.CurrentResult.Output;
                         context.CurrentResult = context.CurrentTest.MakeTestResult();
                         context.CurrentRepeatCount++; // increment Retry count for next iteration. will only happen if we are guaranteed another iteration
+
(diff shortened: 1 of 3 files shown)


mikkelbu:
Hi @kallekrantz 
Thanks for the contribution, but from nunit/nunit#2911 it looks like this is intended behaviour. Does the proposed usage of `TestContext.Progress` also work for you?

CharliePoole:
Creating an issue first gives us the opportunity to point out potential conflicts in what you want nunit to do before you invest in code changes. In this case, I would point you to #2911. 

As you see there, the change was intentional but there is a different type of output that is displayed for all repetitions.

CharliePoole:
@mikkelbu looks like we cross

kallekrantz:
Oh shit, serves me right for thinking I searched enough on google and in the issue tracker! Thanks for the info and sorry for the spam.

I'm actually not sure, we're using the nunit runner for running our SpecFlow tests and although Progress looks interesting I'm not sure how much control I have over where the runner is outputting to. I'll close this, and try to make the Specflow runner do the correct thing instead. 

Thanks for the quick response and once again sorry!

CharliePoole:
No problem. It's easier to find it if you know it's there. :smiley

PR #2952, opened 2018-08-08
Corrected TestFixtureSourceAttribute Summary Description

sboptimal:
Changed summary description to match class name definition

PR #2954, opened 2018-08-10
Attribute Summary Consistency Changes

sboptimal:
I have been through and modified all attributes to use consistent wording for their XML summaries.  There were a couple of grammatical changes in there so it is worth checking all files to see if you are happy with the wording.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/ApartmentAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/ApartmentAttribute.cs
@@ -29,7 +29,7 @@ using NUnit.Framework.Internal;
 namespace NUnit.Framework
 {
     /// <summary>
-    /// <see cref="ApartmentAttribute"/> marks a test that must run in a particular threading apartment state, causing it
+    /// Marks a test as needing to be run in a particular threading apartment state. This will cause it
     /// to run in a separate thread if necessary.
     /// </summary>
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = false, Inherited = true)]
--- a/src/NUnitFramework/framework/Attributes/ApartmentAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/ApartmentAttribute.cs
@@ -29,7 +29,7 @@ using NUnit.Framework.Internal;
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Marks a test that must run in a particular threading apartment state, causing it
+    /// <see cref="ApartmentAttribute"/> marks a test that must run in a particular threading apartment state, causing it
     /// to run in a separate thread if necessary.
     /// </summary>
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = false, Inherited = true)]
--- a/src/NUnitFramework/framework/Attributes/CultureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/CultureAttribute.cs
@@ -29,7 +29,7 @@ using NUnit.Framework.Internal;
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Marks a test fixture or method as applying to a specific Culture.
+    /// Marks an assembly, test fixture or test method as applying to a specific Culture.
     /// </summary>
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method | AttributeTargets.Assembly, AllowMultiple = false, Inherited=false)]
     public class CultureAttribute : IncludeExcludeAttribute, IApplyToTest
--- a/src/NUnitFramework/framework/Attributes/CultureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/CultureAttribute.cs
@@ -29,7 +29,7 @@ using NUnit.Framework.Internal;
 namespace NUnit.Framework
 {
     /// <summary>
-    /// CultureAttribute is used to mark a test fixture or an
+    /// <see cref="CultureAttribute"/> marks a test fixture or an
     /// individual method as applying to a particular Culture only.
     /// </summary>
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method | AttributeTargets.Assembly, AllowMultiple = false, Inherited=false)]
--- a/src/NUnitFramework/framework/Attributes/CultureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/CultureAttribute.cs
@@ -29,8 +29,7 @@ using NUnit.Framework.Internal;
 namespace NUnit.Framework
 {
     /// <summary>
-    /// <see cref="CultureAttribute"/> marks a test fixture or an
-    /// individual method as applying to a particular Culture only.
+    /// Marks a test fixture or method as applying to a specific Culture.
     /// </summary>
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method | AttributeTargets.Assembly, AllowMultiple = false, Inherited=false)]
     public class CultureAttribute : IncludeExcludeAttribute, IApplyToTest
--- a/src/NUnitFramework/framework/Attributes/DatapointSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/DatapointSourceAttribute.cs
@@ -27,11 +27,11 @@ using System.Collections.Generic;
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Used to mark a field, property or method providing a set of datapoints to 
-    /// be used in executing any theories within the same fixture that require an 
-    /// argument of the Type provided. The data source may provide an array of
+    /// <see cref="DatapointSourceAttribute"/> marks a field, property or method as providing 
+    /// a set of datapoints to be used in executing any theories within the same fixture that 
+    /// require an argument of the Type provided. The data source may provide an array of
     /// the required Type or an <see cref="IEnumerable{T}"/>.
-    /// Synonymous with DatapointsAttribute.
+    /// Synonymous with <see cref="DatapointsAttribute"/>.
     /// </summary>
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
     public class DatapointSourceAttribute : NUnitAttribute
--- a/src/NUnitFramework/framework/Attributes/DatapointSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/DatapointSourceAttribute.cs
@@ -27,11 +27,10 @@ using System.Collections.Generic;
 namespace NUnit.Framework
 {
     /// <summary>
-    /// <see cref="DatapointSourceAttribute"/> marks a field, property or method as providing 
-    /// a set of datapoints to be used in executing any theories within the same fixture that 
-    /// require an argument of the Type provided. The data source may provide an array of
-    /// the required Type or an <see cref="IEnumerable{T}"/>.
-    /// Synonymous with <see cref="DatapointsAttribute"/>.
+    /// Marks a field, property or method as providing a set of datapoints for use 
+    /// in executing any theories within the same fixture that require an argument 
+    /// of the Type provided. The data source may provide an array of the required 
+    /// Type or an <see cref="IEnumerable{T}"/>. Synonymous with <see cref="DatapointsAttribute"/>.
     /// </summary>
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
     public class DatapointSourceAttribute : NUnitAttribute
--- a/src/NUnitFramework/framework/Attributes/DatapointSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/DatapointSourceAttribute.cs
@@ -29,7 +29,7 @@ namespace NUnit.Framework
     /// <summary>
     /// Marks a field, property or method as providing a set of datapoints for use 
     /// in executing any theories within the same fixture that require an argument 
-    /// of the Type provided. The data source may provide an array of the required 
+    /// of the provided type. The data source may provide an array of the required 
     /// Type or an <see cref="IEnumerable{T}"/>. Synonymous with <see cref="DatapointsAttribute"/>.
     /// </summary>
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
--- a/src/NUnitFramework/framework/Attributes/DescriptionAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/DescriptionAttribute.cs
@@ -27,8 +27,7 @@ using NUnit.Framework.Internal;
 namespace NUnit.Framework
 {
     /// <summary>
-    /// <see cref="DescriptionAttribute"/> provides the descriptive text about a 
-    /// test case or fixture.
+    /// Provides the descriptive text relating to the test case or fixture.
     /// </summary>
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = false, Inherited=false)]
     public sealed class DescriptionAttribute : PropertyAttribute
--- a/src/NUnitFramework/framework/Attributes/DescriptionAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/DescriptionAttribute.cs
@@ -27,7 +27,7 @@ using NUnit.Framework.Internal;
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Provides the descriptive text relating to the test case or fixture.
+    /// Provides the descriptive text relating to the assembly, test fixture or test method.
     /// </summary>
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = false, Inherited=false)]
     public sealed class DescriptionAttribute : PropertyAttribute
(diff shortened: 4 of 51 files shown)


sboptimal:
Not sure why the AppVeyor build is failing.  Looking at the error I don't think it is related to my pull request.  If it is let me know and I will fix.

CharliePoole:
Appreciated as your work is, I'm not in favor of a chef to the same attribute that is being documented. It takes space and doesn't seem to have any benefit for the user.

Sbosanquet:
Hi @CharliePoole, in an earlier correction to an incorrect comment it was requested that I add it in for that one. From a consistency point of view they should either all have them or not. It's easy enough to remove the cref and attribute name completely from the summary if that's what is needed.  Either way, I feel they need to all be consistent.

CharliePoole:
Sorry, I haven't been following along here but just happened to glance in. When you have a `cref` to the same class you are documenting, what do you see when you hover? What when you click? Does it seem useful?

As the retiring member, I'd say pay more attention to other folks comments than mine, but please do think this through in terms of what the user sees.

CharliePoole:
So it is useful at the point of reference but not at the point of definition. I never thought of it that way, which is why I hadn't generally use it. I imagine the definition also appears when hovering over a list of options provided by Intellisense, so I can see the fix makes sense.

Thanks for the example!

ChrisMaddock:
Just FYI - this will need a review from a second team member - so you may prefer to wait for both sets of feedback before making any further changes. 🙂

sboptimal:
@ChrisMaddock ok, not a problem.  I have corrected those you highlighted.  Amazing how they slip through.  If you want to check the TimeoutAttribute description I have removed the reference to a new thread as requested.

src/NUnitFramework/framework/Attributes/ApartmentAttribute.cs
```diff
@@ -29,7 +29,7 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Marks a test that must run in a particular threading apartment state, causing it
+    /// <see cref="ApartmentAttribute"/> marks a test that must run in a particular threading apartment state, causing it
```
jnm2: While I would use crefs any time a code element is named, I wouldn't name the attribute here. [The BCL doesn't do this](https://github.com/dotnet/corefx/blob/d4541b21dcb0cb15c5ec49c38354bde46e3eaea4/src/System.ComponentModel.Primitives/src/System/ComponentModel/DescriptionAttribute.cs#L7-L11) and the reason that occurs to me is that the name is already right above the text, so starting off the sentence with the same name is visually redundant:

image

I know it's confusing that I didn't ask to entirely remove the existing content in https://github.com/nunit/nunit/pull/2952, but it did cross my mind. :D I love that you're shooting for consistency. That's something we could really use. If you're willing to help us in this direction, I'd prefer to start the sentence with the class name being an implicit subject noun like I see in other projects.

jnm2: Checking with @nunit/framework-team, is view you'd share?

sboptimal: @jnm2 not a problem.  Lets get it sorted.  As I said to Charlie, I'm willing to redo them so they are all have the same structure and consistency.  It isn't a problem.  I'll wait until the general consensus is back and then we can go from there.

ChrisMaddock: > Checking with @nunit/framework-team, is this a view you'd share?

I'm apathetic. Happy to accept this, as @sboptimal's put the work in to tidy things up, but I have no strong opinion on whether we're consistent one way, or the other. 🙂

jnm2: Okay :D Just to clarify, there's a ton in this PR that I want to merge whichever way we go. I'm proposing that in addition to these changes (and as a reversal of only some of them) we delete the attribute names as the first words of the summaries and start with the next word.

ChrisMaddock: Sounds good to me!

mikkelbu: I don't feel strongly about this. If @sboptimal doesn't minding doing the additional changes then I don't see any reason not to do this. Otherwise, I can also live with the current solution.

sboptimal: Ok, so just to confirm:
1. The Cref at the start of each summary will be removed along with the attribute name
1. All attributes will now start with something like: Marks..., Applies..., Supplies... etc.  Basically, whatever their first word is after the attribute name.

Old:
```
    /// <summary>
    /// <see cref="RandomAttribute"/> supplies a set of random values
    /// to a single parameter of a parameterized test.
    /// </summary>
```

New:
```
    /// <summary>
    /// Supplies a set of random values to a single parameter of a parameterized test.
    /// </summary>
```

Looking at the BCL link you sent through earlier.  It seems they wrap the summary comments in a <para> tag for formatting reasons I guess.  Do you want your attributes doing the same way as i'll do them at the same time if so?

jnm2: @sboptimal That's what I had in mind, yes. We don't want to use `<para>` tags line the BCL did in that example, though. We're targeting only IntelliSense with our XML docs and we only use `<para>` to wrap at least two paragraphs.

sboptimal: Ok, not a problem.  I'll try to get it done this week.

sboptimal:
Sorry for the delay.  So as discussed last week here is the new batch of changes, Have a read through and check your are happy.  I'll correct any that need changing.

src/NUnitFramework/framework/Attributes/MaxTimeAttribute.cs
```diff
@@ -29,7 +29,7 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Specifies the maximum time (in milliseconds) for a test case to succeed.
```
ChrisMaddock: Why this change, sorry?

src/NUnitFramework/framework/Attributes/OneTimeSetUpAttribute.cs
```diff
@@ -26,8 +26,8 @@ namespace NUnit.Framework
 	using System;
 
 	/// <summary>
-	/// Attribute used to identify a method that is called once
-	/// to perform setup before any child tests are run.
+	/// Identifies a method to be called once, prior to the child tests executing, 
+	/// to perform any setup requirements for the proceding tests.
```
ChrisMaddock: *preceding - but I think this is the wrong word anyway. Maybe you mean 'descendant' tests - but is that perhaps overcomplicating the explanation? Can you clarify what you wanted to change about the previous description here, and maybe we can come up with a better summary?

src/NUnitFramework/framework/Attributes/NonTestAssemblyAttribute.cs
```diff
@@ -27,12 +27,11 @@
...
 {
     /// <summary>
-    /// The NonTestAssemblyAttribute may be used by third-party frameworks
-    /// or other software that references the NUnit framework but does not
-    /// contain tests. Applying the attribute indicates that the assembly
-    /// is not a test assembly and may prevent errors if certain runners
-    /// attempt to load the assembly. Note that recognition of the attribute
-    /// depends on each individual runner.
+    /// Used by third-party frameworks, or other software, that reference 
+    /// the NUnit framework but does not contain any tests. Applying the
```
ChrisMaddock: does -> do

src/NUnitFramework/framework/Attributes/RepeatAttribute.cs
```diff
@@ -36,8 +36,7 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// RepeatAttribute may be applied to test case in order
-    /// to run it multiple times.
+    /// Specified that a test should be run multiple times.
```
ChrisMaddock: *Specifies

src/NUnitFramework/framework/Attributes/SingleThreadedAttribute.cs
```diff
@@ -28,11 +28,11 @@
...
-    /// SingleThreadedAttribute applies to a test fixture and indicates
-    /// that all the child tests must be run on the same thread as the
-    /// OneTimeSetUp and OneTimeTearDown. It sets a flag in the
-    /// TestExecutionContext and forces all tests to be run sequentially
-    /// on the current thread. Any ParallelScope setting is ignored.
+    /// Marks a test fixture as requiring all child tests must be run on the 
+    /// same thread as the OneTimeSetUp and OneTimeTearDown. A flag in the
+    /// <see cref="TestExecutionContext"/> is set and forces all tests to be run 
+    /// sequentially on the current thread. Any <see cref="ParallelScope"/> 
+    /// setting is ignored.
```
ChrisMaddock: "A flag in the TestExecutionContext is set and forces..." doesn't sound right. Should this be "which forces" now, instead?

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
```diff
@@ -32,8 +32,7 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// TestCaseAttribute is used to mark parameterized test cases
-    /// and provide them with their arguments.
+    /// Marks the method as a parameterized test cases and provide them with their arguments.
```
ChrisMaddock: Pluralization problems in this one. ('Marks' vs 'provide')

src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
```diff
@@ -33,8 +33,7 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// TestCaseSourceAttribute indicates the source to be used to
-    /// provide test cases for a test method.
+    /// Identifies the source to be used as the test cases for a test method.
```
ChrisMaddock: Not sure about the re-wording here - the source and the test cases are different things, really. Can we switch this one back to the old version? `Identifies the source to be used to provide test cases for a test method.`

src/NUnitFramework/framework/Attributes/TimeoutAttribute.cs
```diff
@@ -30,10 +30,10 @@
...
 {
     /// <summary>
-    /// Used on a method, marks the test with a timeout value in milliseconds. 
-    /// The test will be run in a separate thread and is cancelled if the timeout 
-    /// is exceeded. Used on a class or assembly, sets the default timeout 
-    /// for all contained test methods.
+    /// Applies a timeout to a test. When used on a method it marks the test 
+    /// with a timeout value in milliseconds. The test is cancelled if the 
+    /// timeout is exceeded. When used on a class or assembly it sets the 
+    /// default timeout for all contained test methods.
```
ChrisMaddock: I see how you've tried to reword here - the original in this case was ambiguous, I think. The timeout is always in milliseconds, whether the attribute is on a method/class/assembly.

sboptimal:
@ChrisMaddock I have been through and made some changes.  Hopefully should make it easier for the other guys to review if you are happy with one set.  I have rolled back some of the wording as requested.  Others I have corrected as per your suggestions.  If you can have a quick scan when you get chance.

src/NUnitFramework/framework/Attributes/TimeoutAttribute.cs
```diff
@@ -30,10 +30,9 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Used on a method, marks the test with a timeout value in milliseconds. 
-    /// The test will be run in a separate thread and is cancelled if the timeout 
-    /// is exceeded. Used on a class or assembly, sets the default timeout 
-    /// for all contained test methods.
+    /// Applies a timeout in milliseconds to a test. 
+    /// When applied to a method the test is cancelled if the timeout is exceeded. 
+    /// When applied to a class or assembly it sets the default timeout for all contained test methods.
```
jnm2: There should be a comma after each “When” phrase.

src/NUnitFramework/framework/Attributes/CultureAttribute.cs
```diff
@@ -29,8 +29,7 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// CultureAttribute is used to mark a test fixture or an
-    /// individual method as applying to a particular Culture only.
+    /// Marks a test fixture or method as applying to a specific Culture.
```
mikkelbu: Should this also mention assembly, since the attribute has `AttributeTargets.Assembly`?

src/NUnitFramework/framework/Attributes/DatapointSourceAttribute.cs
```diff
@@ -27,11 +27,10 @@
...
 {
     /// <summary>
-    /// Used to mark a field, property or method providing a set of datapoints to 
-    /// be used in executing any theories within the same fixture that require an 
-    /// argument of the Type provided. The data source may provide an array of
-    /// the required Type or an <see cref="IEnumerable{T}"/>.
-    /// Synonymous with DatapointsAttribute.
+    /// Marks a field, property or method as providing a set of datapoints for use 
+    /// in executing any theories within the same fixture that require an argument 
+    /// of the Type provided. The data source may provide an array of the required
```
mikkelbu: For `DatapointsAttribute` we have `... of the provided Type. ...` and not `...of the Type provided. ...` as here. I prefer the `DatapointsAttribute` variant.

src/NUnitFramework/framework/Attributes/DescriptionAttribute.cs
```diff
@@ -27,8 +27,7 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Attribute used to provide descriptive text about a 
-    /// test case or fixture.
+    /// Provides the descriptive text relating to the test case or fixture.
```
mikkelbu: What about assembly?

src/NUnitFramework/framework/Attributes/ExplicitAttribute.cs
```diff
@@ -28,10 +28,9 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// ExplicitAttribute marks a test or test fixture so that it will
-    /// only be run if explicitly executed from the GUI or command line
-    /// or if it is included by use of a filter. The test will not be
-    /// run simply because an enclosing suite is run.
+    /// Marks a test or test fixture such that it will only run if explicitly
```
mikkelbu: What about assembly?

src/NUnitFramework/framework/Attributes/IgnoreAttribute.cs
```diff
@@ -29,9 +29,7 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Attribute used to mark a test that is to be ignored.
-    /// Ignored tests result in a warning message when the
-    /// tests are run.
+    /// Marks a test as being ignored. Ignored tests result in a warning message when the tests are run.
```
mikkelbu: The other places we have also mentioned test fixture (and perhaps also assembly)

src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
```diff
@@ -29,8 +29,7 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// PlatformAttribute is used to mark a test fixture or an
-    /// individual method as applying to a particular platform only.
+    /// Marks a test fixture or method as applying to a specific platform.
```
mikkelbu: Assembly?

src/NUnitFramework/framework/Attributes/SetCultureAttribute.cs
```diff
@@ -28,11 +28,9 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Sets the current Culture for the duration of a test.
-    /// <para>
-    /// It may be specified at the level of a test or a fixture.
-    /// The culture remains set until the test or fixture completes and is then reset to its original value.
-    /// </para>
+    /// Sets the current Culture for the duration of a test. The culture remains set until 
+    /// the test or fixture completes and is then reset to its original value.
```
mikkelbu: Assembly?

src/NUnitFramework/framework/Attributes/SetUICultureAttribute.cs
```diff
@@ -28,11 +28,9 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Sets the current UI Culture for the duration of a test.
-    /// <para>
-    /// It may be specified at the level of a test or a fixture.
-    /// The UI culture remains set until the test or fixture completes and is then reset to its original value.
-    /// </para>
+    /// Sets the current UI Culture for the duration of a test. The UI culture remains set until
+    /// the test or fixture completes and is then reset to its original value.
```
mikkelbu: Assembly?

src/NUnitFramework/framework/Attributes/SetUpAttribute.cs
```diff
@@ -26,8 +26,7 @@ namespace NUnit.Framework
     using System;
 
     /// <summary>
-    /// Attribute used to identify a method that is called 
-    /// immediately before each test is run.
+    /// Identifies a method to be called prior to the running of each test.
```
mikkelbu: I prefer "Identifies a method to be called immediately before each test is run." to match the TearDownAttribute.

src/NUnitFramework/framework/Attributes/TestOfAttribute.cs
```diff
@@ -33,7 +33,7 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Indicates which class the test or test fixture is testing
+    /// Indicates the method or class the test or test fixture is testing
```
mikkelbu: Assembly? We should also end the sentence with a comma.

jnm2: @mikkelbu You meant, end the sentence with a period?

mikkelbu: Yes :)

PR #2961, opened 2018-08-16
Improve error message on EmptyConstraint 

### LINKED ISSUE #2928: Improve error message on EmptyConstraint
When EmptyConstraint is used on the incorrect type, it throws the following exception:

```
throw new System.ArgumentException("The actual value must be a string or a non-null IEnumerable or DirectoryInfo", nameof(actual));
```

When chaining constraints, I sometimes hit this error message because of problems in my chaining, where I'm actually testing against a different type to the one I think I am. It would be useful if this error message included the type of `actual` which caused the exception - so the user can better debug where the problem in their chaining lies.

Ideally, whoever contributes this functionality would also add a corresponding unit test! There may already be something there - I haven't checked. 🙂

nisha-kaushik:
Fix #2928 [](https://github.com/nunit/nunit/issues/2928)


src/NUnitFramework/framework/Constraints/EmptyConstraint.cs
```diff
@@ -49,16 +52,33 @@ public override string Description
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
             // NOTE: actual is string will fail for a null typed as string
+            Type actualType = actual as Type;
+            if (actual == null)
+                actualType = GetDeclaredType(actual);
```
roederja2: Why not just `typeof(TActual)` ?

nisha-kaushik: Noted, I will change. Thanks.

src/NUnitFramework/framework/Constraints/EmptyConstraint.cs
```diff
@@ -48,17 +50,21 @@ public override string Description
         /// <returns>True for success, false for failure</returns>
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
-            // NOTE: actual is string will fail for a null typed as string
+            // NOTE: actual is string will fail for a null typed as string           
+            Type actualType = typeof(TActual);           
+            if (actualType == null)
```
roederja2: I don't think this is correct now. `actualType` will never be null.

nisha-kaushik:
Can someone help with Appveyor build, the tests are working for me on local and without the changes also,build is failing :(

src/NUnitFramework/framework/Constraints/EmptyConstraint.cs
```diff
@@ -48,11 +50,13 @@ public override string Description
         /// <returns>True for success, false for failure</returns>
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
-            // NOTE: actual is string will fail for a null typed as string
+            // NOTE: actual is string will fail for a null typed as string           
+            Type actualType = typeof(TActual);
```
roederja2: If you have the variable you should use it everywhere tyepof(TActual) is used i.e. line 56

ChrisMaddock:
@nisha-kaushik - Unfortunately I think the CI tests are just flaky tests, and nothing related to your changes. Re-running the CI builds now.

src/NUnitFramework/framework/Constraints/EmptyConstraint.cs
```diff
@@ -48,11 +50,13 @@ public override string Description
...
-            // NOTE: actual is string will fail for a null typed as string
-            if (typeof(TActual) == typeof(string))
+            // NOTE: actual is string will fail for a null typed as string           
+            Type actualType = typeof(TActual);
+            
+            if (actualType == typeof(string))
                 realConstraint = new EmptyStringConstraint();
             else if (actual == null)
-                throw new System.ArgumentException("The actual value must be a string or a non-null IEnumerable or DirectoryInfo", nameof(actual));
+                throw new System.ArgumentException($"The actual value of type - {actualType} must be a string or a non-null IEnumerable or DirectoryInfo", nameof(actual));
```
ChrisMaddock: The changes all look good! I think this error message is getting a bit long-winded however - could I suggest rewording it to something a bit more natural? How about something like...

```
$"The actual value must be a string, non-null IEnumerable or DirectoryInfo. The value passed was of type {actualType}."
```

ChrisMaddock:
Just FYI @nisha-kaushik - this also needs a second review from another member of @nunit/framework-team, before it's good to merge - you may wish to wait for the second set of feedback, before making any more required changes. 😄

src/NUnitFramework/framework/Constraints/EmptyConstraint.cs
jnm2: Yes. For reference, [this is message](https://sharplab.io/#v2:EYLgHgbALANALiATgVwHYB9kGcCWqDmABAMoCeWcApgLYDcAsAFBMAOywANjgMaEUCGcHoW4d+WLIQAKiAPb5E/akwDeTQhsJtOwgUN4A3WTgAmhALL88ACgCU6zWsaaXhAMKzUWWR0oA6AHVEHCoAGTxKa1RKAHdCAEFEfGRqSlQ4AHlkTIAzACV+AkoAUTBuShYhT2sAIhZ+RWpCVCVKGphCKAAmDprUiX58Nts/c0oBodsGZ00AXyZZoA) the `ArgumentOutOfRangeException` `actualValue` parameter causes:

> Actual value was 42.

jnm2:
@nisha-kaushik We appreciate your work!

PR #2966, opened 2018-08-19
Restore synchronization context after executing each test case and test case source

### LINKED ISSUE #2821: Save and restore the SynchronizationContext before and after each test case
One of hundreds of tests is causing System.Windows.Forms.Control.CreateHandle() which installs a WindowsFormsSynchronizationContext. When the test worker runs the next and that test happens to be async, the test deadlocks due to the fix for https://github.com/nunit/nunit/issues/2123 not being released yet. The problem is, even with the deadlock fix in NUnit, I don't want the next test running with a WindowsFormsSynchronizationContext.

Rather than locating the misbehaving test(s) through trial and error I'd rather have a failure or at least a warning of executing the work item caused an overall change to the thread's SynchronizationContext.Current. That would cause a nice todo list to fall out of my tests.

Or would we prefer to save and restore `SynchronizationContext.Current` with the other global state we save and restore?

Either way, we'd have to make sure that we take the snapshot before any ITestActions and other setups start and only compare or restore after they all tear down.

jnm2:
Closes #2821. I separated out a 'Fixed whitespace' commit so that you can ignore it.

### DIFF
--- a/src/NUnitFramework/testdata/SynchronizationContextFixture.cs
+++ b/src/NUnitFramework/testdata/SynchronizationContextFixture.cs
@@ -1,0 +1,26 @@
+using System.Threading;
+using NUnit.Framework;
+
+namespace NUnit.TestData
+{
+    public static class SynchronizationContextFixture
+    {
+        public static void TestCasesThatSetSynchronizationContext([Range(1, 2)] int testCase)
+        {
+            if (SynchronizationContext.Current is DummySynchronizationContext)
+            {
+                Assert.Fail();
+            }
+
+            SynchronizationContext.SetSynchronizationContext(new DummySynchronizationContext());
+        }
+
+        private sealed class DummySynchronizationContext : SynchronizationContext
+        {
+            public override SynchronizationContext CreateCopy()
+            {
+                return new DummySynchronizationContext();
+            }
+        }
+    }
+}
--- a/src/NUnitFramework/testdata/SynchronizationContextFixture.cs
+++ b/src/NUnitFramework/testdata/SynchronizationContextFixture.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using System.Threading;
 using NUnit.Framework;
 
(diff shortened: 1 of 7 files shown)


jnm2:
@CharliePoole You've recently edited these files. Can you double-check that my approach of saving and restoring the entire previous execution context in `WorkItem.RunOnCurrentThread` makes sense?

jnm2:
@CharliePoole You've recently edited these files. Can you double-check that my approach of saving and restoring the entire previous execution context in `WorkItem.RunOnCurrentThread` makes sense?

Now that I implemented this for test case sources, it's a bigger question. I believe we actually do want to be restoring the culture etc between test case source executions. If TestExecutionContext is doing too much for that context, I might split out the non-test-related parts.

jnm2:
I suppose this answers my question: 

```
Error : NUnit.Framework.Attributes.TestCaseSourceTests.TestNameIntrospectsArrayValues
System.NullReferenceException : Object reference not set to an instance of an object.
   at NUnit.Framework.Assert.Multiple(TestDelegate testDelegate) in /Users/travis/build/nunit/nunit/src/NUnitFramework/framework/Assert.cs:line 323
```

I'll split it out.

jnm2:
@nunit/framework-team Everything ready for review!

src/NUnitFramework/testdata/SynchronizationContextFixture.cs
```diff
@@ -0,0 +1,55 @@
...
+            Assert.That(succeed);
+        }
+
+        [TestCaseSource(nameof(TestCaseSourceThatSetsSynchronizationContext2))]
+        public static void TestMethodWithSourceThatSetsSynchronizationContext2(bool succeed)
+        {
+            Assert.That(succeed);
+        }
+
+        public static void TestCasesThatSetSynchronizationContext([Range(1, 2)] int testCase)
```
ChrisMaddock: Heh, clever! 🙂

Issue #2963, opened 2018-08-16
Flakey tests in FrameworkControllerTests 


ChrisMaddock:
Several tests in FrameworkControllerTests have recently failed in a flaky manner on the CI builds. We need to look into why that is, and make a fix.

The failures don't appear to be platform-specific, here are a few sample failure messages:

```
Errors, Failures and Warnings
1) Failed : NUnit.Framework.Api.FrameworkControllerTests.RunTestsAction_AfterLoad_ReturnsRunnableSuite("<filter/>")
  Expected: not null and String starting with "ID"
  But was:  "8601"
  at NUnit.Framework.Api.FrameworkControllerTests.RunTestsAction_AfterLoad_ReturnsRunnableSuite (System.String filter) [0x00000] in <1df9fb68b85648378ea6dd2e640b0ed4>:0 
```
```
1) Failed : NUnit.Framework.Api.FrameworkControllerTests.ExploreTestsAction_FilterCategory_ReturnsTests
  Expected: not null and String starting with "ID"
  But was:  "7680"
   at NUnit.Framework.Api.FrameworkControllerTests.ExploreTestsAction_FilterCategory_ReturnsTests() in /home/travis/build/nunit/nunit/src/NUnitFramework/tests/Api/FrameworkControllerTests.cs:line 316
```
```
1) Failed : NUnit.Framework.Api.FrameworkControllerTests.RunTestsAction_AfterLoad_ReturnsRunnableSuite("<filter/>")
  Expected: not null and String starting with "ID"
  But was:  "8508"
  at NUnit.Framework.Api.FrameworkControllerTests.RunTestsAction_AfterLoad_ReturnsRunnableSuite (System.String filter) [0x00000] in <9278af9c5d3e455d8858b76816878cf6>:0 
```

As always, contributions welcome!

CharliePoole:
@ChrisMaddock Are you saying the same tests sometimes pass for the same code? I had seen this error and figured someone had messed up the setting of the prefix for the ids but if it's only failing intermittently, that explanation wouldn't apply.

ChrisMaddock:
@CharliePoole Pass - I haven't had a chance to even look at the code, and see what these tests do! I just wanted to record the error messages as an issue.

mikkelbu:
And again (from https://travis-ci.org/nunit/nunit/jobs/416897254)
```
[...]

PR #2967, opened 2018-08-19
Fixes #2837 DictionaryContainsKeyConstraint behaviour ignores the dictionary key comparer.

### LINKED ISSUE #2837: DictionaryContainsKeyConstraint behaviour is inconstant with Dictionary.ContainsKey when the dictionary uses a custom Comparer

DaiMichael:
@ChrisMaddock @jnm2 

I've had to create a new branch, there were only 3 files that were changed in the original PR. My branch was too messed up which was my fault 100% and it was just me being unfarmiliar with git when I started. I might need to do the same for #2786 (it might be ok) but I will check this myself.

I've made the change so that is fixed for NetStandard1.4, it was a bit of a trial and confusion rained heavily as I needed `System.Reflection.TypeExtensions` which was refereneced but... missing what I thought was there. After a lot of head scratching and drinking neat vodka 😉, I ended up ILSpying and finding out 1.4 uses a degraded version... Anyway I have a fix and it passes all unit tests.

I don't get any other breaks and I ran the tests a number of times to see if they were threading/sync issue but nothing.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
@@ -263,9 +263,8 @@ namespace NUnit.Framework.Constraints
 
 #if NETSTANDARD1_4
         
-        //
         // This is for missing MetadataToken in NetStandard 1.4, matched by method signature
-        //
+
         private static bool Matched(MethodInfo methodInfo, MethodInfo matchedMethodInfo)
         {
             return methodInfo.Name == matchedMethodInfo.Name &&
@@ -273,7 +272,7 @@ namespace NUnit.Framework.Constraints
                    methodInfo.IsHideBySig == matchedMethodInfo.IsHideBySig &&
                    methodInfo.IsGenericMethod == matchedMethodInfo.IsGenericMethod &&
                    methodInfo.ReturnParameter.ParameterType == matchedMethodInfo.ReturnParameter.ParameterType &&
-                   Matched(methodInfo.GetParameters(), matchedMethodInfo.GetParameters())&&
+                   Matched(methodInfo.GetParameters(), matchedMethodInfo.GetParameters()) &&
                    Matched(methodInfo.GetGenericArguments(), matchedMethodInfo.GetGenericArguments());
         }
 
--- a/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2015 Charlie Poole, Rob Prouse
+// Copyright (c) 2018 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
@@ -23,6 +23,10 @@
 
 using System;
 using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using NUnit.Compatibility;
 using NUnit.Framework.Internal;
 
 namespace NUnit.Framework.Constraints
@@ -33,6 +37,8 @@ namespace NUnit.Framework.Constraints
     /// </summary>
     public class DictionaryContainsKeyConstraint : CollectionItemsEqualConstraint
     {
+        private bool _isDeprecatedMode = false;
+
         /// <summary>
         /// Construct a DictionaryContainsKeyConstraint
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -66,31 +72,242 @@ public override string Description
...
+                var tKeyGenericArg = definition.GetGenericArguments().FirstOrDefault(typeArg => typeArg.Name == "TKey");
+
+                if (tKeyGenericArg != null)
+                {
+                    method = definition
+                             .GetMethods(BindingFlags.Instance | BindingFlags.Public)
+                             .FirstOrDefault(m =>
+                                 m.ReturnType == typeof(bool)
+                                 && m.Name == "Contains"
+                                 && m.GetParameters().FirstOrDefault()?.ParameterType == tKeyGenericArg);
```
jnm2: We should check `...&& !m.IsGenericMethod && m.GetParameters().Length == 1 && m.GetParameters()[0]...`

DaiMichael:
@jnm2 I have done changes as per request.

> I'm going to look for a simpler solution for netstandard1.4. Something about that doesn't feel right.

Agreed, but I couldn't find any help with matching methods without using the `MethodInfo.MetadataToken`, this isn't implemented in Standard 1.4. I matched what was needed for this particular case, I haven't tested to see if it would work for all situations. It does test most things are unique in a functional sense (i.e. Parameters / ReturnType / Name / Attributes)  but still.....

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -66,31 +72,242 @@ public override string Description
...
+
+                    if (method != null)
+                    {
+#if NETSTANDARD1_4
+                        method = methods.Single(m => Matched(method, m));
```
jnm2: As far as I can see, .NET Standard 1.4 just puts us in a really uncomfortable spot.

I'd take a different approach. I wouldn't even bother doing `var definition = type.GetGenericTypeDefinition();` on .NET Standard 1.4. May as well copy the whole block and have .NET Standard 1.4 operate on the instantiated `type` instead of `definition`. Would that make sense?

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -66,31 +73,242 @@ public override string Description
...
+            return method;
+        }
+
+        private static MethodInfo FindContainsKeyMethod(Type type)
+        {
+            var methods = type.GetMethods(BindingFlags.Instance | BindingFlags.Public);
+            var method = methods.FirstOrDefault(m =>
+                m.ReturnType == typeof(bool)
+                && m.Name == "ContainsKey"
+                && m.GetParameters().Length == 1);
```
jnm2: We should make sure `!m.IsGenericMethod` here, too.

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
jnm2: I can help with the coding if that's clearer than me trying to explain it. :D

DaiMichael:
@jnm2 Change in 😟 

1) I don't check for for generic but thinking about it perhaps I should, I'm just doing some playing but as this is inside `if (method == null && type.GetTypeInfo().IsGenericType)` , this means we should be looking for a `IsGenericMethod`....

2) That broken test #2963 has started affecting this PR. It seems to happen in side Core2.0

3) How are people debugging Standard1.4 / 2.0.. My R# plugin is unhappy and reports errors and doesn't run the tests. I remember in days of old.... like 10yrs ago there was a gui for nUnit and you could attach to it but now there is only the console. Do you debug by launching the console via VS and then setting which test to run in properties ?

DaiMichael:
@jnm2 - You wrote this and I shamelessly took your example code way back when we started this issue fix... I'm only really delving in to what it is doing now so, is the new way for 1.4 Standard good enough for all frameworks and get ride of the conditional code?

I also think the other way (none 1.4), if we keep it, should check for `IsGenericMethod` as again, it is inside the `if` that checks we are dealing with generics. Unless I'm missing something....

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -66,31 +73,250 @@ public override string Description
...
+                        method = methods.Single(m => m.MetadataToken == method.MetadataToken);
+                    }
+                }
+#else
+                    method = methods.FirstOrDefault(m => 
+                        m.ReturnType == typeof(bool)
+                        && m.Name == "Contains"
+                        && m.IsGenericMethod
+                        && m.GetParameters().Length == 1
+                        && m.GetParameters()[0].ParameterType == type.GetGenericTypeDefinition());
```
jnm2: Great! Indentation is off and `ParameterType == type.GetGenericTypeDefinition()` isn't what we want (parameter type could be `object` if `type.GetGenericTypeDefinition()` is `Dictionary<,>`. Probably don't need a check?

DaiMichael:
I have fixed but my first point stands from above, is the 1.4 code good enough for all code paths?

#2963 seems to come in and go, locally at least...

ChrisMaddock:
Hey @DaiMichael - thanks for refactoring this one.

One of the CI failures looks to be the flakey tests we have at the moment (😞), but there are two test failures in .NET Standard 1.4 which seem to be from this change. Would you mind taking a look?

```
1) Failed : NUnit.Framework.Constraints.DictionaryContainsKeyConstraintTests.FailsWhenNotUsedAgainstADictionary
  Expected: <System.ArgumentException> and property Message String containing "ContainsKey"
  But was:  "Object of type 'System.String' cannot be converted to type 'System.Collections.Generic.KeyValuePair`2[System.String,System.String]'."
   at NUnit.Framework.Constraints.DictionaryContainsKeyConstraintTests.FailsWhenNotUsedAgainstADictionary() in /home/travis/build/nunit/nunit/src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs:line 76

2) Failed : NUnit.Framework.Constraints.DictionaryContainsKeyConstraintTests.SucceedsWhenKeyIsPresentUsingContainsKeyWhenUsingKeyedCollectionCustomComparer
  Expected: dictionary containing key "calum"
  But was:  < "ALICE", "BOB", "CALUM" >
   at NUnit.Framework.Constraints.DictionaryContainsKeyConstraintTests.SucceedsWhenKeyIsPresentUsingContainsKeyWhenUsingKeyedCollectionCustomComparer() in /home/travis/build/nunit/nunit/src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs:line 143
```

jnm2:
@DaiMichael Looking at the error @ChrisMaddock mentioned, it looks like we'll have to check the parameter type of the Add method on .NET Standard 1.4 after all and make sure `ParameterType.IsInstanceOfType(expectedValue)` or something like that.

DaiMichael:
@jnm2 @ChrisMaddock 

>@DaiMichael Looking at the error @ChrisMaddock mentioned, it looks like we'll have to check the parameter type of the Add method on .NET Standard 1.4 after all and make sure ParameterType.IsInstanceOfType(expectedKey) or something like that.

Argggh.... This is proving to be a bit of a nightmare... I did some code to get the Add() and take the parameter out... This didn't work for `ILookup` or `IReadOnlyDictionary`. I then played with returning `Contains` and doing a bit of filtering and this failed the `List<KeyValuePair<>>` tests....

Options.....
1) I've currently checked in my original flaky which no one is 100% happy with solution (matching the function sigs). It works (I know that a "average solution might want to be avoided) and in "theory" should always work on the tests as they aren't a moving target, i.e. defined Interfaces on these are pretty old in .NET.

2) For `NetStandard1.4` we automatically turn on the `_isDeprecatedMode` and people using this out of date, functionally stunted Standard (😆) have to use the old method, but we have marked it 'Obsolete' which is perhaps a little unfair.

I'm leaning to option 1 tbh but I'm happy for someone else to have a crack. We could delete some of the unit tests to make it easier 😉 

FYI - Apologies for pushing some broken tests, my bad... it was late last night but I could have sworn they  all passed, ho humm. My penance was fixing broken references and nuget in work.

DaiMichael:
@ChrisMaddock @jnm2 

Tagged as awaiting contributer but it is awaiting review if we are happy with my Standard 1.4 change.

ChrisMaddock:
Thanks for the ping Dai - bit much to review on my phone right now, but I’ve switched the label!

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -66,31 +73,245 @@ public override string Description
...
+        // This is for missing MetadataToken in NetStandard 1.4, matched by method signature
+        //
+        private static bool Matched(MethodInfo methodInfo, MethodInfo matchedMethodInfo)
+        {
+            return methodInfo.Name == matchedMethodInfo.Name &&
+                   methodInfo.Attributes == matchedMethodInfo.Attributes &&
+                   methodInfo.IsHideBySig == matchedMethodInfo.IsHideBySig &&
+                   methodInfo.IsGenericMethod == matchedMethodInfo.IsGenericMethod &&
+                   methodInfo.ReturnParameter.ParameterType == matchedMethodInfo.ReturnParameter.ParameterType &&
+                   Matched(methodInfo.GetParameters(), matchedMethodInfo.GetParameters())&&
```
jnm2: Space before `&&`

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -66,31 +73,245 @@ public override string Description
...
+            }
+
+            return method;
+        }
+
+#if NETSTANDARD1_4
+        
+        //
+        // This is for missing MetadataToken in NetStandard 1.4, matched by method signature
+        //
```
jnm2: We don't have spacing with empty `//` lines anywhere else, so this sticks out. Dunno if it matters at all.

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -66,31 +73,244 @@ public override string Description
...
-                if (ItemsEqual(obj, Expected))
-                    return true;
+                return false;
+            }
+
+            var method = GetContainsKeyMethod(actual);
+            if (method != null)
+                return (bool)method.Invoke(actual, new[] { Expected });
+
+            throw new ArgumentException("The actual value must have a ContainsKey or Contains(TKey) method.");
```
ChrisMaddock: Can we include `typeof(actual)` in this message somewhere? My reason for asking, is that I would normally hit error messages like this when I'm chaining a long list of constraints together, and I've called this constraint on a type different to the one I think I have. We've just added the type retrospectively to a couple of other constraints - so it would seem a shame to miss the opportunity on new code going into the codebase!

DaiMichael: @ChrisMaddock - Amended exception message for your perusal 😄 .

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -101,7 +101,7 @@ private bool Matches(object actual)
             if (method != null)
                 return (bool)method.Invoke(actual, new[] { Expected });
 
-            throw new ArgumentException("The actual value must have a ContainsKey or Contains(TKey) method.");
+            throw new ArgumentException($"The {actual.GetType().Name} value must have a ContainsKey or Contains(TKey) method.");
```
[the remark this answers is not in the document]
DaiMichael: @jnm2 - Fine by me, rather it done right, nice helper class.

mikkelbu:
Strange. Travis has completed successfully (several hours ago), but the status here is "Pending — The Travis CI build is in progress".

jnm2:
Hitting https://github.com/nunit/nunit/issues/2963 on just the Linux build, restarted.

PR #2971, opened 2018-08-21
Fixes #2970 InvalidCastException @ NUnit.Framework.TestFixtureSourceAttribute.BuildFrom

### LINKED ISSUE #2970: InvalidCastException @ NUnit.Framework.TestFixtureSourceAttribute.BuildFrom
Given the following code:

    using System;
    using NUnit.Framework;

    public class TypedTestFixture<T> : NUnit.Framework.Internal.TestParameters, ITestFixtureData
    {
        public TypedTestFixture(params object[] arguments)
            : base(arguments)
        {
            this.TypeArgs = new[] { typeof(T) };
        }

        public Type[] TypeArgs { get; }
    }

    public class Source
    {
        public static readonly ITestFixtureData[] Fixtures =
        {
            new TypedTestFixture<object>(),
            new TypedTestFixture<int>(),
        };
    }

    [TestFixtureSource(typeof(Source), nameof(Source.Fixtures))]
    public class TestFixture<T>
    {
        public TestFixture() { }

        [Test]
        public void Foo() { }
    }

Test-detection fails and no test is run because an `InvalidCastException` occurs:

> "Unable to cast object of type 'Erowa.EMC.ConfigurationTool.Configuration.Editing.NameValidation.TypedTestFixture`1[Erowa.EMC.Domain.Entities.PalletTypeSettingsEntity]' to type 'NUnit.Framework.Internal.TestFixtureParameters'."} 

and stack trace:

> "   at NUnit.Framework.TestFixtureSourceAttribute.<BuildFrom>d__17.MoveNext() in C:\\src\\nunit\\nunit\\src\\NUnitFramework\\framework\\Attributes\\TestFixtureSourceAttribute.cs:line 114"

which is [here](https://github.com/nunit/nunit/blob/368a65da746c506dd211679c51a9e38d7e7b3597/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs#L112)

BrunoJuchli:
Fixes #2970

I've done three things:

- removed the unnecessary cast from NUnit.Framework.TestFixtureSourceAttribute.BuildFrom
- removed internal setter from NUnit.Framework.Internal.TestFixtureParameters.TypeArgs - as there's no usages of the setter (apart from being set by ctor)
- added test cases, including one for type parameter deduction from ctor arguments (=> goal: unit tests show what's supported and what not).

Question: how can / should I run the tests included in the nunit solution locally? I've not found any documentation on this.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
@@ -25,7 +25,6 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Reflection;
-using NUnit.Compatibility;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
 using NUnit.Framework.Internal.Builders;
@@ -109,7 +108,7 @@ namespace NUnit.Framework
         {
             Type sourceType = SourceType ?? type;
 
-            foreach (TestFixtureParameters parms in GetParametersFor(sourceType))
+            foreach (ITestFixtureData parms in GetParametersFor(sourceType))
                 yield return _builder.BuildFrom(type, PreFilter.Empty, parms);
         }
 
--- a/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
@@ -106,10 +106,7 @@ namespace NUnit.Framework
         /// <param name="type">The type to be used as a fixture.</param>
         public IEnumerable<TestSuite> BuildFrom(Type type)
         {
-            Type sourceType = SourceType ?? type;
-
-            foreach (ITestFixtureData parms in GetParametersFor(sourceType))
-                yield return _builder.BuildFrom(type, PreFilter.Empty, parms);
+            return BuildFrom(type, PreFilter.Empty);
         }
 
         #endregion
@@ -125,7 +122,7 @@ namespace NUnit.Framework
         {
             Type sourceType = SourceType ?? type;
 
-            foreach (TestFixtureParameters parms in GetParametersFor(sourceType))
+            foreach (ITestFixtureData parms in GetParametersFor(sourceType))
                 yield return _builder.BuildFrom(type, filter, parms);
         }
 
--- a/src/NUnitFramework/testdata/TestFixtureSourceData.cs
+++ b/src/NUnitFramework/testdata/TestFixtureSourceData.cs
@@ -375,11 +375,11 @@ namespace NUnit.TestData.TestFixtureSourceData
     }
 
     [TestFixtureSource(typeof(GenericFixtureWithTypeAndConstructorArgsSource), "Source")]
-    public class GenerixFixtureSourceWithTypeAndConstructorArgs<T>
+    public class GenericFixtureSourceWithTypeAndConstructorArgs<T>
     {
         private readonly T _arg;
 
-        public GenerixFixtureSourceWithTypeAndConstructorArgs(T arg)
+        public GenericFixtureSourceWithTypeAndConstructorArgs(T arg)
         {
             _arg = arg;
         }
--- a/src/NUnitFramework/testdata/TestFixtureSourceData.cs
+++ b/src/NUnitFramework/testdata/TestFixtureSourceData.cs
@@ -23,6 +23,7 @@
 
 using System.Collections;
 using NUnit.Framework;
+using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
 using System;
 using System.Collections.Generic;
@@ -92,9 +93,9 @@ namespace NUnit.TestData.TestFixtureSourceData
     }
 
     [TestFixtureSource("StaticProperty")]
-    public class StaticProperty_InheritedClass : StaticProperty_SameClass 
+    public class StaticProperty_InheritedClass : StaticProperty_SameClass
     {
-        public StaticProperty_InheritedClass (string arg) : base(arg, "StaticPropertyInClass") { }
+        public StaticProperty_InheritedClass(string arg) : base(arg, "StaticPropertyInClass") { }
     }
 
     [TestFixtureSource("StaticMethod")]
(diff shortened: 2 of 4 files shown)


src/NUnitFramework/testdata/TestFixtureSourceData.cs
```diff
@@ -353,6 +354,69 @@ public void SomeTest()
...
+            {
+                TypeArgs = new[] { typeof(T) };
+            }
+
+            public Type[] TypeArgs { get; }
+        }
+    }
+
+    [TestFixtureSource(typeof(GenericFixtureWithTypeAndConstructorArgsSource), "Source")]
+    public class GenerixFixtureSourceWithTypeAndConstructorArgs<T>
```
jnm2: Typo `Generix`

jnm2:
Thanks for taking the time! Getting @nunit/framework-team's attention since there isn't a discussion issue.

> Question: how can / should I run the tests included in the nunit solution locally? I've not found any documentation on this.

Perhaps we should link this from the readme: https://github.com/nunit/nunit/blob/master/BUILDING.md#build-script

jnm2:
@BrunoJuchli Looking at the order of the commits, did you check that the tests do fail prior to your changes?

jnm2:
@BrunoJuchli I noticed that this test is failing 

```
NUnit.Framework.Attributes.TestFixtureSourceTests.CanRunGenericFixtureSourceWithProperTypeAndConstructorArgsProvided
>   Expected: Runnable
>   But was:  NotRunnable
>    at NUnit.Framework.Attributes.TestFixtureSourceTests.CanRunGenericFixtureSourceWithProperTypeAndConstructorArgsProvided() in C:\projects\nunit\src\NUnitFramework\tests\Attributes\TestFixtureSourceTests.cs:line 164
```

For suggestions on how to run tests locally other than BUILDING.md, see https://github.com/nunit/nunit/issues/2973. :D

mikkelbu:
> Thanks for taking the time! Getting @nunit/framework-team's attention since there isn't a discussion issue.

Isn't the (discussion) issue nunit/nunit#2970 (or am I overlooking something)?

@BrunoJuchli I've edited your message to mention #2970, so that the issue and PR will be linked (and also such that the PR will auto close the issue when merged). Mentioning it in the titel will not work as far as I know.

src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
```diff
@@ -109,7 +108,7 @@ public IEnumerable<TestSuite> BuildFrom(Type type)
         {
             Type sourceType = SourceType ?? type;
 
-            foreach (TestFixtureParameters parms in GetParametersFor(sourceType))
+            foreach (ITestFixtureData parms in GetParametersFor(sourceType))
```
mikkelbu: Shouldn't we also remove the same unnecessary cast below? (https://github.com/nunit/nunit/pull/2971/files#diff-a7991f3c3308d3394ecebfc7b72ba252R128)

src/NUnitFramework/testdata/TestFixtureSourceData.cs
```diff
@@ -353,6 +354,69 @@ public void SomeTest()
...
+                : base(arguments)
+            {
+                TypeArgs = new[] { typeof(T) };
+            }
+
+            public Type[] TypeArgs { get; }
+        }
+    }
+
+    [TestFixtureSource(typeof(GenericFixtureWithTypeAndConstructorArgsSource), "Source")]
```
mikkelbu: What about `"Source"` => `nameof(GenericFixtureWithTypeAndConstructorArgsSource.Source)`?

src/NUnitFramework/testdata/TestFixtureSourceData.cs
```diff
@@ -353,6 +354,69 @@ public void SomeTest()
...
+    public class GenericFixtureWithConstructorArgsSource
+    {
+        public static readonly TestFixtureData[] Source =
+        {
+            new TestFixtureData(5),
+            new TestFixtureData(new object())
+        };
+    }
+
+    [TestFixtureSource(typeof(GenericFixtureWithConstructorArgsSource), "Source")]
```
mikkelbu: Same comment as above

src/NUnitFramework/testdata/TestFixtureSourceData.cs
BrunoJuchli: I implemented it like that at first but then I wasn't sure whether that would be supported by the compiler c# level.
All other examples in the same file use constant strings instead of the  `nameof(...)` operator.
If `nameof` is fine I'll create a separate pull request to modify all string references to  types/members to use `nameof(...)` - that would make sense, right?

src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
BrunoJuchli: @mikkelbu 
I guess that's why the test is failing. Will have a look.
Thanks!

src/NUnitFramework/testdata/TestFixtureSourceData.cs
BrunoJuchli: @jnm2 thanks!

-----------

I'd be happy to provide a PR to change string-constants to `nameof(..)`, only adjust the newly introduced usages, or open a new issue to discuss the topic.
However, I'm not in a position to decide which is the right way to go so I'll refrain from doing anything regarding this topic until I get instructions ;-)

BrunoJuchli:
@jnm2  @mikkelbu 
Thanks for all your feedback. I've incorporated it, apart from the `nameof(...)` operator (see below).
IMO it should be good to merge now.

Regarding `nameof(..)` I see the following options:

a) leave it as it is
b) apply `nameof(...)` for the whole solution (separate PR)
c) apply `nameof(...)` just to new code

I personally would favor b) for improved refactor-safety. But I'll leave the decision to you guys, of course.

ChrisMaddock:
Yep, I agree

mikkelbu:
Hi @nunit/framework-team can one of you do a second review, so that we can merge this in, as it is closing an issue which is part of the 3.11 milestone. Then we can always create a subsequent task wrt. `nameof(..)`.

src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
```diff
@@ -126,7 +122,7 @@ public IEnumerable<TestSuite> BuildFrom(Type type, IPreFilter filter)
...
```
jnm2: Thanks so much for doing this! This has bugged me for a while. I thought it was due to having to use the type name instead of `var` which silently introduces a possible invalid cast exception when the method return type is changed.

PR #2983, opened 2018-08-26
Allow Warn.If and Warn.Unless to run in a Multiple assert block

### LINKED ISSUE #2979: Warn.If in Assert.Multiple
This test
`
[Test]
public void Test()
{
  Assert.Multiple(() => Warn.If(true));
}
`
gives this result: System.Exception : Warn.If may not be used in a multiple assertion block.

But the documentation of [Multiple Asserts](https://github.com/nunit/docs/wiki/Multiple-Asserts) says:

> 7. Use of Warnings (Assert.Warn, Warn.If, Warn.Unless) is permitted inside a multiple assert block. Warnings are reported normally along with any failures that occur inside the block.

So either the documentaion is worng or there is a worng check in the code.

CharliePoole:
Fixes #2979 

`Warn.If` and `Warn.Unless` were modified to run in multiple assert block.

New tests were added

Existing tests were improved and refactored.

### DIFF
--- a/src/NUnitFramework/tests/Assertions/AssertMultipleTests.cs
+++ b/src/NUnitFramework/tests/Assertions/AssertMultipleTests.cs
@@ -25,6 +25,7 @@ using System;
 using NUnit.Framework.Interfaces;
 using NUnit.TestData.AssertMultipleData;
 using NUnit.TestUtilities;
+using AM = NUnit.TestData.AssertMultipleData.AssertMultipleFixture;
 
 namespace NUnit.Framework.Assertions
 {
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/tests/Assertions/AssertMultipleTests.cs
```diff
@@ -25,101 +25,89 @@
...
-            CheckResult(methodName, ResultState.Failure, assertionMessageRegex);
+            CheckResult(methodName, ResultState.Failure, asserts, assertionMessages);
         }
 
-        [TestCase("ExceptionThrown")]
-        [TestCase("ExceptionThrownAfterTwoFailures", "Failure 1", "Failure 2", "Simulated Error")]
-        public void AssertMultipleErrorTests(string methodName, params string[] assertionMessageRegex)
+        [TestCase(nameof(AM.ThreeAssertWarns), 0, "WARNING1", "WARNING2", "WARNING3")]
+        [TestCase(nameof(AM.ThreeWarnIf_TwoFail), 3, "WARNING1", "WARNING3")]
+        [TestCase(nameof(AM.ThreeWarnUnless_TwoFail), 3, "WARNING1", "WARNING3")]
```
mikkelbu: A question about Warn (which I've not used much - if at all). So `Assert.Warn` does not count as an assertion, but `Warn.If` and `War.Unless` do?

mikkelbu: I just checked it, and so it is.

CharliePoole: Same is true of `Assert.Fail`, `Assert.Pass` and `Assert.Inconclusive`.  I actually expected `Assert.Warn` to be counted so I went through the same research myself. 😄

Issue #2903, opened 2018-06-23
Package acceptance tests


jnm2:
Proof-of-concept: https://github.com/nunit/nunit/compare/jnm2/nupkg_tests

`.\build.ps1 -t acceptance` tests the full restore, build, and NUnitLite run of a `uap10.0` project which pulls our .NET Standard 1.4 builds of NUnit and NUnitLite. The pattern can easily be followed to extend the concept to test against our other runners on various platforms.

This will allow us to move ahead with greater confidence when adding new package dependencies. Eventually, it will help us spot incompatibilities introduced between the other repos and this one.

### Questions

- Our CI is already starting to be uncomfortably slow to me. How do you feel about not testing this in CI?
@rprouse, how do you feel about us leaving it to you to discover what we broke, as part of the release process? Maybe we could do a nightly or weekly slow build if we fix the slow NuGet restore and figure out the developer mode issue?

- I put acceptance tests under `/tests` and added `/acceptance.cake` and `/lib.cake`. I also put a utility I wrote for this project under `/src/AppxRunner` since it is used by `acceptance.cake`. I'm very open do better names and better locations, if you have any suggestions.

/cc @nunit/framework-team 

### ℹ NuGet Restore

One of the things it does to accomplish this is /tests/nuget.config:

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <config>
    <clear />
    <add key="globalPackagesFolder" value="Isolated package cache" />
  </config>
  <packageSources>
    <add key="Build script package output" value="..\package" />
  </packageSources>
</configuration>
```

This makes it safe to do a NuGet restore of freshly built NUnit 3.11 packages without leaking them onto your user cache. The initial restore takes a long time. There might be a way to improve this by combining with the existing cache while guaranteeing that we don't trash it.

### ℹ Developer mode

Building and restoring works without this. To actually run NUnitLite on UAP, you need to enable this setting:

![image](https://user-images.githubusercontent.com/8040367/41805502-b3ca2a3c-7679-11e8-880c-d0a471b20955.png)

ChrisMaddock:
Drive-by-phone-post just to say that this is incredibly exciting! (And also, now we’ve got this far...I wonder how far off we are running the actual nunit tests on a UWP runner?! One day! 😜)

[...]

PR #3003, opened 2018-09-05
Add Platform detection to .NET Standard 2.0 build

### LINKED ISSUE #2432: Ability to exclude/include the platform .NET Core
The .NET Standard tests may be executed against a variety of runtimes, so ideally I would write this:

```cs
[Test]
[Platform(Exclude = "netcore", Reason = "BeginInvoke causes PlatformNotSupportedException")]
public static void WarningInBeginInvoke()
{
    new Action(() => Assert.Warn("(Warning message)"))
        .BeginInvoke(null, null)
        .AsyncWaitHandle.WaitOne();
}
```

Using `#if !(NETSTANDARD1_3 || NETSTANDARD1_6)` is not ideal because the .NET Standard tests can (and in theory should) be run against .NET Framework and Mono as well.

Unless we've got a new include/exclude attribute lined up to implement, can we just add the platform to the existing platform attribute?

MatthewBeardmore:
Closes #2432.

Implements platform detection for .NET Standard 2.0. .NET Standard 1.4 did not have the APIs available to implement this feature, so .NET Core 1.1 and 1.0 apps will not be able to be detected. Adds NETCORE as a new identifier for runtime detection and adds .NET Core version detection. 

Adds unit tests for .NET Core detection, enables the existing platform detection unit tests for .NET Core 2.0 testing, and fixes a hidden failing unit test now that .NET Core support was enabled (InvalidFilePathsThrowsArgumentException has a change in behavior vs .NET Framework). 

Fixes a couple of existing issues with platform detection that I ran into while creating the unit tests with runtime support. First issue was that .NET Framework 4.5 was not able to be used as a exclude/include option as the CLRVersion was not set to a detected version when using InitFromClrVersion. This change allows for "net-4.5" to be used as a runtime identifier. Secondly, checking whether a runtime supported a different runtime checked that the major and minor versions were greater than each other. However, this failed in cases such as checking that .Net Core 3.0 would support .Net Core 2.2. I'm unsure if this was intentional, but it previously worked with .NET Framework due to the differences in CLR versions up until 4.0+. If this is intentional, I'll add a comment around that section of the code and change the unit tests.

### DIFF
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -372,7 +372,7 @@ namespace NUnit.Framework.Api
 #else
             env.AddAttribute("clr-version", Environment.Version.ToString());
 #endif
-#if !PLATFORM_DETECTION
+#if NETSTANDARD1_4 || NETSTANDARD2_0
             env.AddAttribute("os-version", System.Runtime.InteropServices.RuntimeInformation.OSDescription);
 #else
             env.AddAttribute("os-version", OSPlatform.CurrentPlatform.ToString());
--- a/src/NUnitFramework/framework/Internal/PlatformHelper.cs
+++ b/src/NUnitFramework/framework/Internal/PlatformHelper.cs
@@ -248,7 +248,7 @@ namespace NUnit.Framework.Internal
                     isSupported = IntPtr.Size == 4;
                     break;
 
-#if NET40 || NET45
+#if NET40 || NET45 || NETSTANDARD2_0
                 // We only support bitness tests of the OS in .NET 4.0 and up
                 case "64-BIT-OS":
                     isSupported = Environment.Is64BitOperatingSystem;
--- a/src/NUnitFramework/framework/Internal/PlatformHelper.cs
+++ b/src/NUnitFramework/framework/Internal/PlatformHelper.cs
@@ -327,7 +327,7 @@ namespace NUnit.Framework.Internal
         {
             if (versionSpecification != null)
             {
-                throw new PlatformNotSupportedException("Detection versions of .NET Core is not supported - " + runtime.ToString() + "-" + versionSpecification);
+                throw new PlatformNotSupportedException($"Detecting versions of .NET Core is not supported - {runtime.ToString()}-{versionSpecification}");
             }
 
             RuntimeFramework target = new RuntimeFramework(runtime, RuntimeFramework.DefaultVersion);
--- a/src/NUnitFramework/framework/Internal/PlatformHelper.cs
+++ b/src/NUnitFramework/framework/Internal/PlatformHelper.cs
@@ -295,7 +295,7 @@ namespace NUnit.Framework.Internal
                     return IsRuntimeSupported(RuntimeType.Net, versionSpecification);
 
                 case "NETCORE":
-                    return IsRuntimeSupported(RuntimeType.NetCore, versionSpecification);
+                    return IsNetCoreRuntimeSupported(RuntimeType.NetCore, versionSpecification);
 
                 case "SSCLI":
                 case "ROTOR":
@@ -319,6 +319,18 @@ namespace NUnit.Framework.Internal
                 : new Version(versionSpecification);
 
             RuntimeFramework target = new RuntimeFramework(runtime, version);
+
+            return _rt.Supports(target);
+        }
+
+        private bool IsNetCoreRuntimeSupported(RuntimeType runtime, string versionSpecification)
+        {
+            if (versionSpecification != null)
+            {
+                throw new PlatformNotSupportedException("Detection versions of .NET Core is not supported - " + runtime.ToString() + "-" + versionSpecification);
+            }
+
+            RuntimeFramework target = new RuntimeFramework(runtime, RuntimeFramework.DefaultVersion);
 
             return _rt.Supports(target);
         }
--- a/src/NUnitFramework/framework/Internal/PlatformHelper.cs
+++ b/src/NUnitFramework/framework/Internal/PlatformHelper.cs
@@ -54,7 +54,7 @@ namespace NUnit.Framework.Internal
         /// Comma-delimited list of all supported Runtime platform constants
         /// </summary>
         public static readonly string RuntimePlatforms =
-            "Net,SSCLI,Rotor,Mono,MonoTouch";
+            "Net,NetCore,SSCLI,Rotor,Mono,MonoTouch";
 
         /// <summary>
         /// Default constructor uses the operating system and
@@ -293,6 +293,9 @@ namespace NUnit.Framework.Internal
             {
                 case "NET":
                     return IsRuntimeSupported(RuntimeType.Net, versionSpecification);
+
+                case "NETCORE":
+                    return IsRuntimeSupported(RuntimeType.NetCore, versionSpecification);
 
                 case "SSCLI":
                 case "ROTOR":
--- a/src/NUnitFramework/framework/Internal/RuntimeFramework.cs
+++ b/src/NUnitFramework/framework/Internal/RuntimeFramework.cs
@@ -110,21 +110,8 @@ namespace NUnit.Framework.Internal
             }
             else if (isNetCore)
             {
-#if NETSTANDARD2_0
-                var assembly = typeof(System.Runtime.GCSettings).GetTypeInfo().Assembly;
-                var assemblyPath = assembly.CodeBase.Split(new[] { '/', '\\' }, StringSplitOptions.RemoveEmptyEntries);
-                int netCoreAppIndex = Array.IndexOf(assemblyPath, "Microsoft.NETCore.App");
-                if (netCoreAppIndex > 0 && netCoreAppIndex < assemblyPath.Length - 2)
-                {
-                    string version = assemblyPath[netCoreAppIndex + 1];
-                    Version frameworkVersion;
-                    if (Version.TryParse(version, out frameworkVersion))
-                    {
-                        major = frameworkVersion.Major;
-                        minor = frameworkVersion.Minor;
-                    }
-                }
-#endif
+                major = 0;
+                minor = 0;
             }
             else /* It's windows */
 #if NETSTANDARD2_0
@@ -406,7 +393,7 @@ namespace NUnit.Framework.Internal
         private static bool IsNetCore()
         {
 #if NETSTANDARD2_0
-            //Mono versions will throw a TypeLoadException when attempting to run the internal method, so we wrap it in a try/catch 
+            // Mono versions will throw a TypeLoadException when attempting to run the internal method, so we wrap it in a try/catch 
             // block to stop any inlining in release builds and check whether the type exists
             Type runtimeInfoType = Type.GetType("System.Runtime.InteropServices.RuntimeInformation,System.Runtime.InteropServices.RuntimeInformation", false);
             if (runtimeInfoType != null)
--- a/src/NUnitFramework/framework/Internal/RuntimeFramework.cs
+++ b/src/NUnitFramework/framework/Internal/RuntimeFramework.cs
@@ -27,6 +27,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 #if NETSTANDARD2_0
 using System.Runtime.Versioning;
 #endif
@@ -413,10 +414,11 @@ namespace NUnit.Framework.Internal
             return false;
         }
 
+        [MethodImpl(MethodImplOptions.NoInlining)]
         private static bool IsNetCore_Internal()
         {
 #if NETSTANDARD2_0
-            //Mono versions will throw a TypeLoadException when attempting to run any method that uses RuntimeInformation
+            // Mono versions will throw a TypeLoadException when attempting to run any method that uses RuntimeInformation
             // so we wrap it in a try/catch block in IsNetCore to catch it in case it ever gets this far
             if (System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription.StartsWith(".NET Core", StringComparison.OrdinalIgnoreCase))
             {
--- a/src/NUnitFramework/framework/Internal/RuntimeFramework.cs
+++ b/src/NUnitFramework/framework/Internal/RuntimeFramework.cs
@@ -27,6 +27,9 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+#if NETSTANDARD2_0
+using System.Runtime.Versioning;
+#endif
 using Microsoft.Win32;
 
 namespace NUnit.Framework.Internal
@@ -46,7 +49,9 @@ namespace NUnit.Framework.Internal
         /// <summary>Mono</summary>
         Mono,
         /// <summary>MonoTouch</summary>
-        MonoTouch
+        MonoTouch,
+        /// <summary>Microsoft .NET Core</summary>
+        NetCore
     }
 
     /// <summary>
@@ -75,12 +80,15 @@ namespace NUnit.Framework.Internal
             Type monoTouchType = Type.GetType("MonoTouch.UIKit.UIApplicationDelegate,monotouch");
             bool isMonoTouch = monoTouchType != null;
             bool isMono = monoRuntimeType != null;
+            bool isNetCore = !isMono && !isMonoTouch && IsNetCore();
 
             RuntimeType runtime = isMonoTouch
                 ? RuntimeType.MonoTouch
                 : isMono
                     ? RuntimeType.Mono
-                    : RuntimeType.Net;
+                    : isNetCore
+                        ? RuntimeType.NetCore
+                        : RuntimeType.Net;
 
             int major = Environment.Version.Major;
             int minor = Environment.Version.Minor;
@@ -97,6 +105,23 @@ namespace NUnit.Framework.Internal
                         minor = 5;
                         break;
                 }
+            }
+            else if (isNetCore)
+            {
+#if NETSTANDARD2_0
+                string frameworkName = Assembly.GetEntryAssembly()?.GetCustomAttribute<TargetFrameworkAttribute>()?.FrameworkName;
+                if (frameworkName != null)
+                {
+                    string[] versionSplit = frameworkName.Split(new string[1] { "=v" }, StringSplitOptions.RemoveEmptyEntries);
+                    Version frameworkVersion;
+                    if (versionSplit.Length == 2 &&
+                        Version.TryParse(versionSplit[1], out frameworkVersion))
+                    {
+                        major = frameworkVersion.Major;
+                        minor = frameworkVersion.Minor;
+                    }
+                }
+#endif
             }
             else /* It's windows */
             if (major == 2)
@@ -173,6 +198,10 @@ namespace NUnit.Framework.Internal
             if (version.Major > 0) // 0 means any version
                 switch (Runtime)
                 {
+                    case RuntimeType.NetCore:
+                        ClrVersion = new Version(4, 0, 30319);
+                        break;
+
                     case RuntimeType.Net:
                     case RuntimeType.Mono:
                     case RuntimeType.Any:
@@ -222,6 +251,10 @@ namespace NUnit.Framework.Internal
             ClrVersion = version;
             if (Runtime == RuntimeType.Mono && version.Major == 1)
                 FrameworkVersion = new Version(1, 0);
+            if (Runtime == RuntimeType.Net && version.Major == 4 && version.Minor == 5)
+                ClrVersion = new Version(4, 0, 30319);
+            if (Runtime == RuntimeType.NetCore)
+                ClrVersion = new Version(4, 0, 30319);
         }
 
         #endregion
@@ -352,12 +385,46 @@ namespace NUnit.Framework.Internal
             if (!VersionsMatch(ClrVersion, target.ClrVersion))
                 return false;
 
-            return FrameworkVersion.Major >= target.FrameworkVersion.Major && FrameworkVersion.Minor >= target.FrameworkVersion.Minor;
+            if (FrameworkVersion.Major > target.FrameworkVersion.Major)
+                return true;
+            return FrameworkVersion.Major == target.FrameworkVersion.Major && FrameworkVersion.Minor >= target.FrameworkVersion.Minor;
         }
 
         #endregion
 
         #region Helper Methods
+
+        private static bool IsNetCore()
+        {
+#if NETSTANDARD2_0
+            //Mono versions will throw a TypeLoadException when attempting to run the internal method, so we wrap it in a try/catch 
+            // block to stop any inlining in release builds and check whether the type exists
+            Type runtimeInfoType = Type.GetType("System.Runtime.InteropServices.RuntimeInformation,System.Runtime.InteropServices.RuntimeInformation", false);
+            if (runtimeInfoType != null)
+            {
+                try
+                {
+                    return IsNetCore_Internal();
+                }
+                catch { }
+            }
+#endif
+
+            return false;
+        }
+
+        private static bool IsNetCore_Internal()
+        {
+#if NETSTANDARD2_0
+            //Mono versions will throw a TypeLoadException when attempting to run any method that uses RuntimeInformation
+            // so we wrap it in a try/catch block in IsNetCore to catch it in case it ever gets this far
+            if (System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription.StartsWith(".NET Core", StringComparison.OrdinalIgnoreCase))
+            {
+                return true;
+            }
+#endif
+            return false;
+        }
 
         private static bool IsRuntimeTypeName(string name)
         {
--- a/src/NUnitFramework/framework/Internal/RuntimeFramework.cs
+++ b/src/NUnitFramework/framework/Internal/RuntimeFramework.cs
@@ -30,8 +30,9 @@ using System.Reflection;
 using System.Runtime.CompilerServices;
 #if NETSTANDARD2_0
 using System.Runtime.Versioning;
-#endif
+#else
 using Microsoft.Win32;
+#endif
 
 namespace NUnit.Framework.Internal
 {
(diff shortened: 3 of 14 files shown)


src/NUnitFramework/framework/Internal/RuntimeFramework.cs
```diff
@@ -352,13 +385,47 @@ public bool Supports(RuntimeFramework target)
...
+            //Mono versions will throw a TypeLoadException when attempting to run the internal method, so we wrap it in a try/catch 
+            // block to stop any inlining in release builds and check whether the type exists
+            Type runtimeInfoType = Type.GetType("System.Runtime.InteropServices.RuntimeInformation,System.Runtime.InteropServices.RuntimeInformation", false);
+            if (runtimeInfoType != null)
+            {
+                try
+                {
+                    return IsNetCore_Internal();
+                }
+                catch { }
```
jnm2: The usual method to prevent inlining is `[MethodImpl(MethodImplOptions.NoInlining)]`. Would we be able to do that instead?

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
```diff
@@ -352,13 +385,47 @@ public bool Supports(RuntimeFramework target)
...
+            }
+#endif
+
+            return false;
+        }
+
+        private static bool IsNetCore_Internal()
+        {
+#if NETSTANDARD2_0
+            //Mono versions will throw a TypeLoadException when attempting to run any method that uses RuntimeInformation
```
jnm2: Space after `//`

src/NUnitFramework/framework/nunit.framework.csproj
```diff
@@ -15,6 +15,10 @@
     <PackageReference Include="NUnit.System.Linq" Version="0.6.0" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Microsoft.Win32.Registry" Version="4.4.0" />
```
jnm2: I didn't see the registry being used. Do we need this WIndows compat pack package?

src/NUnitFramework/tests/TestContextTests.cs
```diff
@@ -400,7 +400,16 @@ public void FilePathAndDescriptionDoesNotThrow()
 #endif
         public void InvalidFilePathsThrowsArgumentException(string filePath)
         {
+#if NETCOREAPP2_0
+            //Due to changes in .NET Core 2.0 around file path verification, this test does not match
```
jnm2: Space after `//`

jnm2:
Would you please edit the PR body to add "Closes #2432" (https://help.github.com/articles/closing-issues-using-keywords/)?

> Secondly, checking whether a runtime supported a different runtime checked that the major and minor versions were greater than each other. However, this failed in cases such as checking that .Net Core 3.0 would support .Net Core 2.2. I'm unsure if this was intentional, but it previously worked with .NET Framework due to the differences in CLR versions up until 4.0+. If this is intentional, I'll add a comment around that section of the code and change the unit tests.

I'm not sure here. @nunit/framework-team?

src/NUnitFramework/framework/nunit.framework.csproj
MatthewBeardmore: Unfortunately, yes. That code is used to detect the .NET Framework version by looking in the registry in RuntimeFramework.cs. We need it for .NET Standard as well in case the .NET standard framework is running on .NET Framework instead of .NET Core (although, I'm not sure that would ever happen).

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
MatthewBeardmore: That would help - I'll make that change.

src/NUnitFramework/framework/nunit.framework.csproj
ChrisMaddock: Hmm...not a nice dependency for a .NET Standard build!

As far as I can see (please doubel check my logic!), in RuntimeFramework, the registry lookup is only used for determining if the current framework is any of .NET 2.0/3.0/3.5. Given the .NET Standard build will never be running on any of those platforms, could we pre-processor out the `Registry.OpenSubkey` in NET_STANDARD20, then remove this dependency? And of course throw an exception should the 'impossible' ever happen? 🙂

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
ChrisMaddock: Does this change mean the try and catch should also be removed, and comments updated? Or am I missing the point here? This isn't something I'm familiar with! 😅

src/NUnitFramework/framework/Api/FrameworkController.cs
```diff
@@ -372,7 +372,7 @@ public static TNode InsertEnvironmentElement(TNode targetNode)
 #else
             env.AddAttribute("clr-version", Environment.Version.ToString());
 #endif
-#if !PLATFORM_DETECTION
+#if NETSTANDARD1_4 || NETSTANDARD2_0
```
ChrisMaddock: I _think_ this preprocessor should stay in sync with whatever's wrapping the OSPlatform class (currently PLATFORM_DETECTION). My _guess_ is that this currently works because of the `Microsoft.Win32.Registry` - but will fail to compile once that's removed. 

OSPlatform also uses Registry lookup (seemingly to determine Windows versions...). I think this is the right solution, it will just need updating if the dependency is removed. Either that, or the introduction of a REGISTRY_LOOKUP pre-processor - although that's possibly overkill! 😄

src/NUnitFramework/tests/Attributes/TestCaseAttributeTests.cs
```diff
@@ -395,6 +396,84 @@ public void CanExcludePlatform()
...
+                Assert.That(testCase2.RunState, Is.EqualTo(RunState.Runnable));
+                Assert.That(testCase3.RunState, Is.EqualTo(RunState.Skipped));
+            }
+            else
+            {
+                Assert.That(testCase1.RunState, Is.EqualTo(RunState.Skipped));
+                Assert.That(testCase2.RunState, Is.EqualTo(RunState.Runnable));
+                Assert.That(testCase3.RunState, Is.EqualTo(RunState.Runnable));
+            }
+        }
```
ChrisMaddock: Nice tests! 👍

src/NUnitFramework/tests/TestContextTests.cs
```diff
@@ -400,7 +400,16 @@ public void FilePathAndDescriptionDoesNotThrow()
 #endif
         public void InvalidFilePathsThrowsArgumentException(string filePath)
         {
+#if NETCOREAPP2_0
```
ChrisMaddock: Good find!

This test is only intended to test that a garbage path is checked for, and thrown out. The test below checks the behaviour for non-existent files already.

I think a better change here would be to make the test path something which is invalid on .NET Core too - i.e. change it to `bad|path.png`?

MatthewBeardmore:
Since I can't reply to all the comments from Chris, I'll put them here.

The registry is a nasty addition sadly, which I'm really not sure whether it's possible to remove. It can easily be ifdef-ed out of the RuntimeFramework.cs file by the method you put in above. However, it is also used in OSPlatform.cs, which I'm really not seeing a way of getting rid of at this time. It's used in the private static Version GetWindows81PlusVersion(Version version) method, which gets whether the OS is Windows 8, 8.1, or 10. This would still be needed in the .NET Standard version I think, based on whether the app that is running it has the manifest file set to the right version (or not).

For the try/catch addition, it'll get run under older versions of Mono, as when it tries to compile the IsNetCore_Internal method, it will throw an exception as it can't find the System.Runtime.InteropServices assembly, which will be caught in that try/catch.

As for the #if NETSTANDARD1_4 || NETSTANDARD2_0 change, this would actually work as-is without my change. I can revert that, but I figured that I'd keep it working as-is, as those two platforms were both previously executing that code path.

For InvalidFilePathsThrowsArgumentException, I can change that - I just didn't want to change the intention of the unit test, so I left the input argument as-is. I can make that change if you want.

ChrisMaddock:
> For InvalidFilePathsThrowsArgumentException, I can change that - I just didn't want to change the intention of the unit test, so I left the input argument as-is. I can make that change if you want.

Yes please! (It's actually a test I originally wrote, so I can 'ok' the intention! 😂)

> For the try/catch addition, it'll get run under older versions of Mono, as when it tries to compile the IsNetCore_Internal method, it will throw an exception as it can't find the System.Runtime.InteropServices assembly, which will be caught in that try/catch.

👍 Cool, thanks for clarifying!

> As for the #if NETSTANDARD1_4 || NETSTANDARD2_0 change, this would actually work as-is without my change. I can revert that, but I figured that I'd keep it working as-is, as those two platforms were both previously executing that code path.

Ah - I'd mis-interpreted this bit, thinking that the .NET Standard functionality was the 'degraded version' for when the OSPlatform class wasn't available, but looks like it may be intended the other way around - as OSDescription isn't available on the older .NET Frameworks we target. Looks like your current way is good! :+1:

> The registry is a nasty addition sadly, which I'm really not sure whether it's possible to remove. It can easily be ifdef-ed out of the RuntimeFramework.cs file by the method you put in above. However, it is also used in OSPlatform.cs, which I'm really not seeing a way of getting rid of at this time. It's used in the private static Version GetWindows81PlusVersion(Version version) method, which gets whether the OS is Windows 8, 8.1, or 10. This would still be needed in the .NET Standard version I think, based on whether the app that is running it has the manifest file set to the right version (or not).

I'd missed OSPlatform was included in this, as it didn't show up in the diff. 

In my eyes, it's a big dependency for a very minor piece of functionality - I would prefer we offered a degraded-functionality PlatformAttribute in the .NET Standard builds, than added the dependency. _However_ - I wonder if there's a way we can achieve the Win 8 functionality in OSPlatform in .NET Standard via the System.Runtime.InteropServices.RuntimeInformation.OSDescription class that the NUnitLite output uses? 

Pinging @nunit/framework-team on this last point to see if anyone disagrees with my thoughts! 🙂

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
jnm2: Yes, I would expect that we would remove it. @MatthewBeardmore Was the `try` actually used to catch an exception or just for the side effects on inlining with the current JIT?

src/NUnitFramework/framework/nunit.framework.csproj
jnm2: Right, the .NET Standard 1.4 build shouldn't be checking for any versions of .NET Framework lower than 4.6.1.

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
```diff
@@ -98,6 +107,23 @@ public sealed class RuntimeFramework
                         break;
                 }
             }
+            else if (isNetCore)
+            {
+#if NETSTANDARD2_0
+                string frameworkName = Assembly.GetEntryAssembly()?.GetCustomAttribute<TargetFrameworkAttribute>()?.FrameworkName;
```
jnm2: This tells you the version of .NET Core that the entry assembly was compiled against (e.g. 1.0) but not the version it's running on (could be any other version).

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
```diff
@@ -98,6 +107,23 @@ public sealed class RuntimeFramework
...
+                if (frameworkName != null)
+                {
+                    string[] versionSplit = frameworkName.Split(new string[1] { "=v" }, StringSplitOptions.RemoveEmptyEntries);
```
jnm2: ℹ `new[] { "=v" }` is an option

jnm2:
I don't mind the dependency itself as much as I mind that we're using the registry, but if it's really useful in the netstandard1.4 build and it's used safely, I'm fine with it.

MatthewBeardmore:
The registry currently is not used in the .NET Standard 1.4 build in this branch - only in .NET Standard 2.0 as the OSPlatform.cs and RuntimeFramework.cs classes are now being built after enabling PLATFORM_DETECTION.

MatthewBeardmore:
If we want to remove the registry dependency from .NET Standard 2.0, that is easily enough done - we just won't be calling the GetWindows81PlusVersion version, which means that for projects that don't have a manifest file that says it supports Windows 10, it will always be returning Windows 8. I personally am fine with that, as the APIs are designed to return that for compatibility based on the manifest file.

Let me know what you'd like me to do.

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
MatthewBeardmore: That's true... I didn't realize that. I'm not sure of a way to get the running .NET Core version then. I'll investigate that issue further.

jnm2:
@MatthewBeardmore Sorry, thinking about it, it's better for the .NET Standard builds to behave the same  when running on .NET Framework as the .NET Framework build. So I think I'm happy with this dependency. @ChrisMaddock, would you be okay with that?

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
MatthewBeardmore: So looking into how other projects do it, BenchmarkDotNet had the following issue: https://github.com/dotnet/BenchmarkDotNet/issues/448 - however, it doesn't work for docker or self-contained distributions (which is a big issue for me - all of my .net core projects use self-contained) - see the open https://github.com/dotnet/BenchmarkDotNet/issues/788. So I'm not sure there is a really good way to detect it at runtime.

jnm2: I see. Can we detect *whether* it's .NET Core while sidestepping the whole version issue, just so we don't restrict our future options? So `netcore` would work, but `netcore2.1` would not be recognized?

MatthewBeardmore: We can do that, yeah. Is that the route we want to go (limit to only netcore - no version detection) or do we want to implement the BenchmarkDotNet way that will at least detect the version in some cases like they do?

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
MatthewBeardmore: From my comment below:

For the try/catch addition, it'll get run under older versions of Mono, as when it tries to compile the IsNetCore_Internal method, it will throw an exception as it can't find the System.Runtime.InteropServices assembly, which will be caught in that try/catch.

src/NUnitFramework/framework/nunit.framework.csproj
MatthewBeardmore: None of this code will be running with .NET Standard 1.4. It doesn't have the APIs available to do platform detection. I can remove the registry check from RuntimeFramework.cs where it checks for frameworks < .NET 4, which will never happen, but OSPlatform.cs still needs the registry. If we aren't removing the registry, the code will be more clear if we don't #if out that section of code (imho).

ChrisMaddock:
Sorry, I'm not convinced on this one. 😞 

As the nunit.framework sits next to the production assembly, doesn't adding this dependency potentially open us up to a world of binding redirection problems? What happens if the user's production code references a different version of the Microsoft.Win32.Registry package than the NUnit framework does? Do we end up in a situation where the user is restricted in which package versions they can use in production, because of the version of NUnit they are using? Alternatively - they require binding redirects...which potentially could cause problems if there were any breaking changes in the Registry package. This sounds like we're about to introduce another problem like https://github.com/nunit/nunit3-vs-adapter/issues/366, which @jnm2 has just done a lot of work to resolve - unless I'm missing why that's not the case here?

If adding this dependency had substantial benefit, I'd be more willing to consider it - but just to distinguish in between Win8.1 and Win 10 in the Platform attribute doesn't justify that, in my opinion! Below are the solutions I'd personally prefer us to look at, in preference order:

1. Investigate using `Runtime.OSDescription` in `OSPlatform.GetWindows81PlusVersion` instead of the existing registry lookup. This would be for the .NET Standard builds only, .NET Framework 4.5 and below don't have `Runtime.OSDescription`, so would need to keep the existing functionality
1. Instead provide a degraded PlatformAttribute in .NET Standard - i.e `[Platform(Win)]` is valid, `[Platform(Win10)]` is not valid.

Of course, this is a team project, so I'll happily step aside if people disagree with me! 😄 But I'm currently unconvinced that adding this dependency is a good idea, I'm afraid.

MatthewBeardmore:
I'm not convinced either that adding the registry is a good idea for .NET Standard. However, I think there is an option 3 to your list:
3) Provide a slightly degraded PlatformAttribute for .NET Standard that doesn't use the registry, but will still return correct Windows version information assuming the manifest file is set up correctly, otherwise, it will mark > Windows 8 as Windows 8.

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
jnm2: I'm not sure, if the check would fail if the tests are compiled to a self-contained app or run in Docker. @nunit/framework-team?

CharliePoole:
As developers, we are experts in developing. We are not experts in predicting what users need, perhaps surprisingly, given that those users are other developers. 😕 

So how about doing the simpler solution, as suggested by @MatthewBeardmore and waiting to see if anybody actually needs us to deal with the extremely limited corner case.

jnm2:
@ChrisMaddock I don't feel strongly, but this may relieve some of the worry: The reason why the dependency is a problem with nunit/nunit3-vs-adapter#366 is that Mono.Cecil is *not* a dependency that the project system knows about when compiling, but rather is sneakily replaced behind the compiler's back.

But in this scenario, not even binding redirects would be necessary. Because the test project references the production project, both the test project and the production project will compile and run against the very same `Microsoft.Win32.Registry` assembly, guaranteed.

If I remember right, all of our pain with dependencies has been either that we bundled dependency assemblies instead of referencing a NuGet package containing the dependency assemblies, or that certain .NET Standard BCL packages had buggy packaging when the .NET team told us to assume will never happen again (and report if it does).

jnm2:
So, in theory, adding the dependency (as a package reference in the nuspec, *not* bundled) occurs to me as the least work and least pain. But if we can achieve parity through option 1 then that's a tie for me, and 2 is a close second so long as `win10` errors rather than silently skipping tests.

Option 3 would really bother me since there would be subtle behavioral changes that could cause important tests to be silently skipped when people switch to .NET Standard. I'f be okay with option 3 only if the .NET Framework builds started doing the same thing.

CharliePoole:
@jnm2 Yes. Action could be to mark the test as non-runnable in that case.

ChrisMaddock:
@jnm2 I dug into this some more. Could something like this case not still bite us, if I'm understanding correctly?

image

https://docs.microsoft.com/en-us/nuget/consume-packages/dependency-resolution#cousin-dependencies

The reference added here will work with package versions 4.4.0 and above. What if user code requires specifically version 4.3.0?

-----

On the alternatives, I agree 'silent failure' is bad, which is what I also understood option 3 to be. Charlie - you're option sounds nice, but sounds like the option which requires the most work to determine those valid/invalid states - I'd be quite happy for us just to block any version specific windows name. We could then 'enhance' at a later date if there's demand. 🙂

jnm2:
@ChrisMaddock You would resolve the cousin dependency error by directly referencing the version you want from `MyApp`. This is something that happens in my work projects not infrequently. To quote the text under the image:

> In these situations, the top-level consumer (the application or package) should add its own direct dependency on Package B so that the Nearest Wins rule applies.

ChrisMaddock:
Fair enough. This is much more your area than mine, so I’ll trust your judgement that we don’t need to worry about it! 😊

jnm2:
I want to add dependencies https://github.com/nunit/nunit/pull/2860 and https://github.com/nunit/nunit/pull/2861, too, but I was waiting for https://github.com/nunit/nunit/issues/2903. :D

MatthewBeardmore:
So what is the final desired result? It sounds like we're okay with the registry requirement, so is there any other changes that need to be made?

src/NUnitFramework/framework/nunit.framework.csproj
jnm2: The .NET Standard 2.0 build shouldn't be checking for any versions of .NET Framework lower than 4.6.1 either, since those frameworks can't run it.

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
jnm2: I'd prefer only catching exceptions we specifically know about, like `TypeLoadException`.

jnm2:
Let's make sure @rprouse is okay with the nuspec dependency too.

src/NUnitFramework/framework/nunit.framework.csproj
CharliePoole: It shouldn't be actually checking, but itt has to return not supported for those frameworks that can't run it in order for the Platform attribute to work correctly. We used to do something like this in the CF build... we conditionally included code to reject everything except CF.

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
```diff
@@ -407,7 +415,7 @@ private static bool IsNetCore()
                 {
                     return IsNetCore_Internal();
                 }
-                catch { }
+                catch(TypeLoadException) { }
```
jnm2: Space after `catch` , please, sorry!

src/NUnitFramework/framework/nunit.framework.csproj
jnm2: @MatthewBeardmore So instead of adding the registry dependency, why not out `#if !NETSTANDARD2_0` around this, since there's no way .NET Framework's 2.0 CLR could be running the .NET Standard build?

https://github.com/nunit/nunit/blob/49ded54b0f68f333399b8931e6e0a9151df92ee7/src/NUnitFramework/framework/Internal/RuntimeFramework.cs#L104-L123

`throw PlatformNotSupportedException();` would be fine.

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
```diff
@@ -352,13 +394,48 @@ public bool Supports(RuntimeFramework target)
...
         }
 
         #endregion
 
         #region Helper Methods
 
+        private static bool IsNetCore()
+        {
+#if NETSTANDARD2_0
+            //Mono versions will throw a TypeLoadException when attempting to run the internal method, so we wrap it in a try/catch
```
jnm2: Since we're keeping this, space after `//` please (sorry again)

ChrisMaddock:
The .NET Core Version thing is still to be resolved too, right? Picking up on this conversation:

> I see. Can we detect whether it's .NET Core while sidestepping the whole version issue, just so we don't restrict our future options? So netcore would work, but netcore2.1 would not be recognized?

 > @MatthewBeardmore
MatthewBeardmore 2 days ago
We can do that, yeah. Is that the route we want to go (limit to only netcore - no version detection) or do we want to implement the BenchmarkDotNet way that will at least detect the version in some cases like they do?

>  @jnm2
jnm2 2 days ago  Member
I'm not sure, if the check would fail if the tests are compiled to a self-contained app or run in Docker. nunit/framework-team?

I agree, if we can't reliably determine .NET Core versions, lets not offer that functionality, and just limit it to detecting non-version-specific netcore. 🙂

src/NUnitFramework/tests/Internal/PlatformDetectionTests.cs
```diff
@@ -383,37 +383,13 @@ public void DetectMonoTouch()
...
+                new RuntimeFramework(RuntimeType.NetCore, new Version(0, 0, 0)),
+                "NetCore");
         }
-
-        [Test]
-        public void DetectNetCore21()
-        {
-            CheckRuntimePlatforms(
-                new RuntimeFramework(RuntimeType.NetCore, new Version(2, 1, 0)),
-                "NetCore,NetCore-2.0,NetCore-2.1");
```
jnm2: Would you please add a test verifying that `NetCore-2.0` errors, rather than skipping or matching?

jnm2:
See `https://github.com/dotnet/corefx/issues/31002` for .NET Core version detection.

src/NUnitFramework/tests/Internal/PlatformDetectionTests.cs
ChrisMaddock: I would expect PlatformHelper to throw an InvalidPlatformException, which will be caught by the platformattribute and mark the test not-runnable. There's an example of that for another case - it should be possible just to extend that to cover this case too.

That means that a user will get clear feedback if they try and create a test with the NetCore-2.0, when it is not valid.

MatthewBeardmore:
I looked into that - unfortunately, it doesn't return what you would expect - it returns "win-x64", "linux-x64", or something like that and not the versioning of .NET Core.

src/NUnitFramework/tests/Internal/PlatformDetectionTests.cs
MatthewBeardmore: Sounds good, I'll make that happen.

src/NUnitFramework/framework/nunit.framework.csproj
MatthewBeardmore: It wouldn't be possible to get into that section, which is why I've added an #ifdef around that section. The registry is needed to detect Windows 8+ versioning if the manifest is not set up with a GUID that marks it as supporting Windows 8 and higher, which we can't remove without a difference between the .NET Standard and .NET Framework builds.

jnm2: @MatthewBeardmore You're quite right, not sure what I was thinking at the time. Thanks for changing this section!

src/NUnitFramework/framework/Internal/PlatformHelper.cs
```diff
@@ -319,6 +322,18 @@ private bool IsRuntimeSupported(RuntimeType runtime, string versionSpecification
 
             return _rt.Supports(target);
         }
+
+        private bool IsNetCoreRuntimeSupported(RuntimeType runtime, string versionSpecification)
+        {
+            if (versionSpecification != null)
+            {
+                throw new PlatformNotSupportedException("Detection versions of .NET Core is not supported - " + runtime.ToString() + "-" + versionSpecification);
```
jnm2: `Detection` → `Detecting`

(If you like, you can use a C# 6 interpolated string.)

src/NUnitFramework/tests/Internal/ThreadUtilityTests.cs
```diff
@@ -21,7 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-#if PARALLEL && PLATFORM_DETECTION
+#if PARALLEL && PLATFORM_DETECTIONN && THREAD_ABORT
```
jnm2: Just caught `PLATFORM_DETECTIONN` now.

jnm2: https://github.com/nunit/nunit/pull/3224

PR #3007, opened 2018-09-06
fixes #2786 Timeout value not resetting on Retry of failed test.

### LINKED ISSUE #2786: Timeout value not resetting on Retry of failed test

DaiMichael:
This is a new PR for #2786, my previous branch had some issues that was too painful to resolve, this was my fault when I created it. It also needed rebasing as there have been a lot of changes since that creation.

I have created a new branch which doesn't need rebasing and I have merged just my changes. This should be clean for the PR. No other changes have been made since this was accepted previously, other than I updated a couple of the copyright dates in the headers. 😄

### DIFF
--- a/src/NUnitFramework/framework/Attributes/RepeatAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RepeatAttribute.cs
@@ -40,7 +40,7 @@ namespace NUnit.Framework
     /// to run it multiple times.
     /// </summary>
     [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
-    public class RepeatAttribute : PropertyAttribute, IWrapSetUpTearDown
+    public class RepeatAttribute : PropertyAttribute, IRepeatTest
     {
         private readonly int _count;
 
@@ -53,7 +53,7 @@ namespace NUnit.Framework
             _count = count;
         }
 
-        #region IWrapSetUpTearDown Members
+        #region IRepeatTest Members
 
         /// <summary>
         /// Wrap a command and return the result.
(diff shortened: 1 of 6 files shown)


DaiMichael:
Sorry... Should add the PR to close is #2849 (previous PR for this issue).

CharliePoole:
Looks good to me. It's too bad the command has to be generated in SimpleWorkItem but there's really no choice. It would, of course, be better if we could handle both timeout and repetition as part of the normal command chain but this has been a long-standing problem with timeout.

jnm2:
Yes, thank you @DaiMichael!

PR #3009, opened 2018-09-09
Fix failing CI Builds by upgrading to NUnit Console 3.10.0

ChrisMaddock:
All framework CI builds are currently failing, as the framework tests currently use `NUnit.ConsoleRunner 3.9.0-dev-03938`, and this version can't be found. 

As far as I can see (publicly), it looks like only the last 20 builds are available on MyGet. I'm not sure if that's intentional or not - but probably means we shouldn't be using myget in this case! @rprouse - I think you're the only person with myget access, could you check this? Looks like there are [package retention settings](https://docs.myget.org/docs/reference/package-retention) that may be set 'wrong'.

To fix the CI, this upgrades us to 3.9.0, from NuGet.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,4 +1,4 @@
-#tool nuget:https://www.myget.org/F/nunit/api/v3/index.json?package=NUnit.ConsoleRunner&version=3.9.0-dev-03938
+#tool nuget:?package=NUnit.ConsoleRunner&version=3.9.0
 #tool GitLink
 
 //////////////////////////////////////////////////////////////////////


rprouse:
We could increase the number of versions that we retain on MyGet, but I originally decreased it because we hit our storage limit and new uploads were failing. We could also contact MyGet about an OpenSource plan....

ChrisMaddock:
This looks like it would meet our needs - just needs to be renewed annually, unfortunately. I’m presuming we’d get suitable alerts however. 

https://www.myget.org/opensource

Happy to help set it up, if you want. In my mind this is something we can improve - ideally we want people to be using the dev builds, and that’s more difficult if a certain version is only available for a limited time.

ChrisMaddock:
@rprouse I can't find it either - I just Googled for it last time.

Here's what used to be there. Maybe worth contacting the MyGet people directly?
https://web.archive.org/web/20170323224409/https://www.myget.org/opensource

ChrisMaddock:
@rprouse Looks like MyGet was taken over a few weeks ago. Wonder if it's related. 😞 

https://blog.myget.org/post/2018/08/31/assembla-acquisition.html

rprouse:
I hope that doesn't mean they are backing away from providing support for opensource projects. I posted a comment on the blog post asking so we'll see.

Meanwhile, i looked at Azure DevOps Artifacts, but they are only for the team.

Issue #3016, opened 2018-09-12
Broadcasting messages through engine


nvborisenko:
Continuing https://github.com/nunit/nunit-console/issues/433

Sometimes a test needs send a message which is not kind of test output.

It would be nice to have something like:
```csharp
TestContext.BroadcastMessage(Type fromType, Type destinationType, string message);
```

- `destinationType` can be null, what means that message is broadcast to everybody
- `fromType` is for understanding who sent a message

And any logger can handle it via xml:
```xml
<broadcast-message from="fullTypeName" to="fullTypeName" testId="123" time="....Z">
  <![CDATA[ This is message content ]]>
</broadcast-message>
```

- `to` attribute is optional
- `testId` attribute is optional due a message may be broadcast from empty context
- `time` attribute is DateTime when message was sent

@CharliePoole please welcome here to review

CharliePoole:
@nvborisenko Not sure why you created a new issue rather than using the old one.

I write some guidelines for that issue as requested by @ChrisMaddock but I don't see the comment posted! I'll check further after a little more sleep. :smiley

ChrisMaddock:
@CharliePoole - this comment? https://github.com/nunit/nunit-console/issues/433#issuecomment-419655440

I saw it, and it looked ideal to me. 🙂

CharliePoole:
@ChrisMaddock Strange. I can't see it in OctoDroid at all but if I switch to the browser I do.

nvborisenko:
I created this issue in correct repository, previous one is in `nunit-console`.

I am ready to provide a pull request, let's just finalize specification for xml events.

CharliePoole:
Interesting... I thought of it as an engine issue, but it looks like all the actual code will be in the framework. We could have continued there, so long as the PR is here, but this is OK as well.

Comments on the proposed design...

1. I originally suggested a method in TestContext, but I'm having second thoughts. Is this really a facility to be used by test writers or is it really for use in custom attributes, etc.? Discussion leading up to this point has tended to be about needing to send a message to an engine extension, so I assumed a custom attribute would be distributed with such an extension. If it is really only for use in and with extensions, then I'd put it in `TestExecutionContext` without any access from `TestContext`. If it's really for users (non-extenders) then a frontend in `TestContext` is needed.

[...]

Issue #2960, opened 2018-08-16
Planning for 3.11 Release


rprouse:
We are past due for a 3.11 release. I went through the issues and cleaned up. We only have 33 completed issues so far for 3.11, so I don't think there is a huge rush, so I was thinking mid to late September for a release. @nunit/framework-team if there is anything you are working on or planning on working on that you would like to see in 3.11, please add to the milestone. I removed a couple of long running issues/PRs from the milestone. If work is ongoing and you'd like to get them in, feel free to re-add.

rprouse:
@agray I'm still planning on getting the release out by the end of the month if not a bit earlier. I've been doing some cleanup and other admin duties now that I've got the console release out. Are their particular issues that you are waiting for? If so, those are the sort of thing I would like to highlight in the release notes.

agray:
Parallel test execution in the same assembly like JUnit.

CharliePoole:
Could you clarify? We already do parallel execution on assembly. What features are implied by "like JUnit"?

jnm2:
@agray I think NUnit 3 has had that for quite a while?

agray:
When I have a data driven test fixture tests won't run in parallel.
Implemention is at github.com/sauceforge/saucery3.

CharliePoole:
@agray Is this related to #2481? It doesn't appear that anyone has worked on that or even confirmed it to this point.

agray:
It doesn't hang on my machine. It would be good if someone else could
confirm it.

Getting NUnit 3 to behave like JUnit 4 wrt parallelism has been the goal of
Saucery3 all along.

agray:
Charlie informed me that the parallism in NUnit 3 was across different
assemblies not in the same assembly.

In JUnit 4 parallism works great in a single data driven test fixture. For
an implementation of this see github.com/sauceforge/SauceryJ

I'm yet to be able to get NUnit 3 to work the same way.  As soon as I try
to add parallelism attribute to the TestFixture it hangs.

Regards,

Andrew

rprouse:
@agray there may be some fixes and improvements in the upcoming release of NUnit, but no major changes to parallelism or capabilities. I would confirm your requirements and expectations using the NUnit Console, then if you are not seeing the same behaviour in Saucery, we can help you figure out what you are doing differently.

CharliePoole:
[...]

PR #3018, opened 2018-09-13
Added SendMessage method to TestExecutionContext

nvborisenko:
Added `SendMessage` method to `TestExecutionContext` for sending messages from tests to listeners.

Discussion: https://github.com/nunit/nunit/issues/3016


CharliePoole:
This looks very good to me but I'd like to see some tests of the new code added.

nvborisenko:
@CharliePoole done. Please review.

src/NUnitFramework/tests/Interfaces/TestMessageTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        [TestCase("destination", "text", "testId")]
+        [TestCase("destination", "text", null)]
+        [TestCase(null, "text", "testId")]
+        public void ToXml_IncludeAttributesInProperFormatting(string destination, string message, string testId)
+        {
+            var testMessage = new TestMessage(destination, message, testId);
+            var expected = new StringBuilder("<test-message");
+
+            if (destination != null)
```
CharliePoole: If we allow the destination to be null, how will any extensions decide whether the message is for them? I know the original spec talks about broadcast, but that doesn't make much sense to me unless we require all extensions to respond to broadcasts, which would be a breaking change.

If you have a particular usage in mind, let us know. Otherwise, I think we should require a destination.

[Good thing about tests: I didn't notice this originally. 😄]

@jnm2 Opinion on this?

@nvborisenko I'm OK with merging this once we resolve the issue about a required destination. FYI, it takes two of the team members agreeing in order to merge, so someone else will need to add their voice.

CharliePoole: OK. It's ready as far as I'm concerned. We'll wait for the second review.

nvborisenko: BTW travis build returns unpredictable results. And I guess somebody is working on setting up AzureDevops pipelines.

CharliePoole:
I'm on my phone and can't kick the build. Next person in will do it.

nvborisenko:
@jnm2 I will appreciate if you review this PR. Thank you in advance.

CharliePoole:
Restarted Travis, which was failing for an unrelated test. @nunit/framework-team this needs a second reviewer.

src/NUnitFramework/framework/Interfaces/TestMessage.cs
```diff
@@ -0,0 +1,97 @@
...
+        public string TestId { get; }
+
+        /// <summary>
+        /// Convert the TestMessage object to an XML string
+        /// </summary>
+        public string ToXml()
+        {
+            TNode tnode = new TNode("test-message", Text, true);
+
+            if (Destination != null)
```
jnm2: This always succeeds unless the constructor is changed. Should we allow null destinations in the constructor?
/cc @nunit/framework-team

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -484,9 +484,19 @@ private TestExecutionContext CreateIsolatedContext()
             return context;
         }
 
-#endregion
+        /// <summary>
+        /// Send a test message to listeners
```
jnm2: `Sends` and please add a period.

Should we mention that it goes in the test result XML, or is that overwhelmingly obvious?

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -484,9 +484,19 @@ private TestExecutionContext CreateIsolatedContext()
...
+        /// </summary>
+        /// <param name="destination">Destination of test message to be handled by listeners</param>
```
jnm2: How about `A name recognized by the intended listeners.` or something like that?

src/NUnitFramework/framework/Internal/TestListener.cs
```diff
@@ -50,6 +50,12 @@ public class TestListener : ITestListener
         /// <param name="output">A TestOutput object containing the text to display</param>
         public void TestOutput(TestOutput output) { }
 
+        /// <summary>
+        /// Called when a test produces message to be sent to listeners
```
jnm2: <code>produces *a* message</code>

nvborisenko:
@jnm2 added one more commit to follow your review. Please validate, thank you again in advance.

jnm2:
@nvborisenko Do you mind dropping the two `One more commit to trigger travis` commits since we don't normally squash PRs?

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
```diff
@@ -475,6 +475,14 @@ public async Task AsyncTestCanAccessOutWriter()
 
 #endregion
 
+#region SendMessage
+        [Test]
+        public void CanSendMessage()
+        {
+            TestExecutionContext.CurrentContext.SendMessage("destination", "message");
```
jnm2: Is there a way to verify the result of this call?

src/NUnitFramework/nunitlite/TeamCityEventListener.cs
```diff
@@ -104,6 +104,12 @@ public void TestFinished(ITestResult result)
         /// <param name="output">A TestOutput object containing the text to display</param>
         public void TestOutput(TestOutput output) { }
 
+        /// <summary>
+        /// Called when a test produces message to be sent to listeners
```
jnm2: <code>produces *a* message</code>

Change also needed in all ITestListener implementations.

jnm2:
I did not realize that `TestOutput` was the odd one out having `ToXml` return `string` instead of `TNode`, or I wouldn't have asked you to make that change. Sorry about that!

@nunit/framework-team Are you okay with `TestMessage` returning `TNode` like the ToXml method on  `ITestFilter`, `ITest`, `ITestResult`, `IPropertyBag` but unlike `TestOutput` which does not implement an interface?

src/NUnitFramework/framework/Interfaces/TestMessage.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            if (destination == null)
+            {
+                throw new ArgumentNullException(nameof(destination));
+            }
+
+            if (text == null)
+            {
+                throw new ArgumentNullException(nameof(text));
+            }
```
jnm2: I opened https://github.com/nunit/nunit/issues/3029 because we have asked for the `Guard` class to be used in the past, but that wouldn't necessarily be my preference.

src/NUnitFramework/framework/Interfaces/TestMessage.cs
```diff
@@ -0,0 +1,109 @@
...
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Diagnostics;
+
+namespace NUnit.Framework.Interfaces
+{
+    /// <summary>
+    /// The TestMessage class holds a message sent by a test to all listeners
```
jnm2: The style we're moving to is to start the sentence with an implied subject: `Holds a message sent by a test to all listeners.`

(Otherwise I'd ask for a `<see cref` to make the name refactor-safe, and then it creates visual duplication both in intellisense and in the source code.)

src/NUnitFramework/framework/Interfaces/TestMessage.cs
```diff
@@ -0,0 +1,109 @@
...
+        /// supplied parent node..
+        /// </summary>
+        /// <param name="parentNode">The parent node.</param>
+        /// <param name="recursive">If true, descendant results are included</param>
+        public TNode AddToXml(TNode parentNode, bool recursive)
+        {
+            TNode tnode = parentNode.AddElementWithCDATA("test-message", Message);
+
+            if (Destination != null)
+                tnode.AddAttribute("destination", Destination);
```
jnm2: Why is Destination checked for null here since Message is not?

src/NUnitFramework/framework/Interfaces/TestMessage.cs
```diff
@@ -0,0 +1,109 @@
...
+            return Destination + ": " + Message;
+        }
+
+        /// <summary>
+        /// A message to send to listeners
+        /// </summary>
+        public string Message { get; }
+
+        /// <summary>
+        /// Destination of a message.
```
jnm2: It's referring to this message so it would be <code>Destination of *the* message</code>. Same for TestId.

src/NUnitFramework/framework/Interfaces/TestMessage.cs
```diff
@@ -0,0 +1,109 @@
...
+                throw new ArgumentNullException(nameof(text));
+            }
+
+            Destination = destination;
+            Message = text;
+            TestId = testId;
+        }
+
+        /// <summary>
+        /// Converts <see cref="TestMessage"/> object to string
```
jnm2: Should be <code>Converts *the*</code>

CharliePoole:
It's a heck of a lot more complicated that way, and a string is actually what we end up with no matter how you implement it.

For the more complicated formats, we needed a way to easily manipulate the XML adding various attributes and elements, and working directly with a string was too messy. We were supporting platforms that didn't all have the same native XML implementation so I wrote TNode as a common way to do simple things with XML rather than using conditional logic for the various implementations.

In the case of TextOutput, two factors militated against using TNode:
1. Some tests produce lots of output, so it needed to be efficient.
2. The format is extremely simple.

You may notice that we do the same thing for our test and suite start notifications.

jnm2:
@CharliePoole Thanks!

@nvborisenko I apologize. Would you undo the ToXml change I mistakenly asked for earlier?

src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
```diff
@@ -125,6 +125,32 @@ public override void Send(ITestListener listener)
...
+        /// Initializes a new instance of the <see cref="TestMessageEvent"/> class.
+        /// </summary>
+        /// <param name="testMessage">The test message object.</param>
+        public TestMessageEvent(TestMessage testMessage)
+        {
+            _testMessage = testMessage;
+        }
+
+        /// <summary>
+        /// Calls SenMessage on the specified listener.
```
jnm2: Typo `SenMessage`—good reason to use `<see cref`.

nvborisenko:
@jnm2 Done.

jnm2:
@nvborisenko 

The test is currently failing:

```
Failed : NUnit.Framework.Internal.TestExecutionContextTests.CanSendMessage
  Expected: 1
  But was:  4165
  at NUnit.Framework.Internal.TestExecutionContextTests.CanSendMessage ()
```

That'll need to be fixed before we merge.

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
```diff
@@ -475,6 +475,25 @@ public async Task AsyncTestCanAccessOutWriter()
...
+
+#if !NETCOREAPP1_1
+            var listener = TestExecutionContext.CurrentContext.Listener as QueuingEventListener;
+            Assert.That(listener.Events.Count, Is.EqualTo(1));
```
jnm2: Should probably assert that the count is equal to one greater than the previous count?

Also, can you mark this test `[NonParallelizable]` to draw attention to the fact that this method cannot be safely run in parallel with other tests?

nvborisenko: It seems we cannot verify queue of the events.

Moreover it seems `SendMessage` method doesn't add TestEvent immediately into the queue. https://ci.appveyor.com/project/CharliePoole/nunit/build/3.11.0-ci-05907-pr-3018#L1001

Currently I removed all assertions which are related to the queue. Only one way I see that's we can dequeue TestEvents until we face our specific TestEvent, this is kind of awaiting. But this way may affect others tests which are working with listener's queue. Need your input here.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -484,9 +484,19 @@ private TestExecutionContext CreateIsolatedContext()
...
 
-#endregion
+        /// <summary>
+        /// Sends a message from test to listeners. This message is not kind of test output and doesn't go to test result.
+        /// </summary>
+        /// <param name="destination">A name recognized by the intended listeners.</param>
+        /// <param name="message">A message to be sent</param>
+        public void SendMessage(string destination, string message)
+        {
+            Listener?.SendMessage(new TestMessage(destination, message, CurrentContext.CurrentTest.Id));
```
CharliePoole: This is an instance method - as it should be. So the user already called the static `CurrentContext` to get the context that is to be used. In that scenario, your use of it is merely redundant.

But consider a unit test, where we create our own `TestExecutionContext` so as not to interfere with other tests. In that case, your code is not redundant but causes an error because the __actual__ current context will be used instead of the one we created for testing!

src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs
CharliePoole: This test merely verifies that the call does not throw. It doesn't indicate what happens as a result. I believe you need to follow a strategy similar to `TextOutputTests`, where the class itself implements a listener and verifies that it receives the proper message. That test would work within it's own, specially created context, in order to avoid polluting the actual test run with a fake message to "destination". 

This is probably better done in `SendMessageTests` rather than here.

CharliePoole:
In the latest commit, you are going to a lot of trouble to make sure that the intercepted calls to the listener continue to be processed normally.

But in a unit test, that's exactly the opposite of what we want. You are sending a fake message, which is meaningless, so we don't want it to ever leave the test environment. We should isolate this and __avoid__ having it continue to the actual listener. 

Please question me if I haven't explained that well! :smile:

nvborisenko:
@CharliePoole Thanks, I think I caught it. Please see my last commit.

nvborisenko:
Thanks guys for your support. As I understand this feature will be available in NUnit.Framework 3.11.0. Do we have some approximate estimations when this package will be released?

mikkelbu:
Hi @nvborisenko 
Both the milestone and #2960 says in the end of this month, so I guess that this is still the plan.

rprouse:
> Both the milestone and #2960 says in the end of this month, so I guess that this is still the plan.

That is correct, hopefully this weekend or early next week.

Thanks for the contribution, great work.

PR #3019, opened 2018-09-14
Attempt to fix hangs by using Sleep(1) instead of Sleep(0) for the polyfill's replacement of Thread.Yield

jnm2:
tl;dr `Thread.Sleep(0)` hogs the CPU and is a terrible replacement for `Thread.Yield()`. `Thread.Sleep(1)` is probably the best we can do on .NET CLR 2.0.

See the comment in the (small) diff. I'm no expert here, so I could be missing something.
Sources:

- http://joeduffyblog.com/2006/08/22/priorityinduced-starvation-why-sleep1-is-better-than-sleep0-and-the-windows-balance-set-manager/

  > In our example above when Sleep(0) is used, we hope this will unstick the machine and let the producer produce and finally the consumer to consume. Indeed with some testing, we see it unstick after a little more than 3 seconds. This is still long enough, however, to kill performance on a server application, cause a noticeable perf degradation on the client, and destroy responsiveness in a GUI app.

- https://blogs.msdn.microsoft.com/khen1234/2006/02/02/sleeping-vs-yielding/

/cc @rprouse @oznetmaster @CharliePoole

### DIFF
--- a/src/NUnitFramework/framework/Compatibility/System.Threading/SpinWait.cs
+++ b/src/NUnitFramework/framework/Compatibility/System.Threading/SpinWait.cs
@@ -142,14 +142,19 @@ namespace System.Threading
                 {
                     Thread.Sleep(1);
                 }
-                else // if ((yieldsSoFar % SLEEP_0_EVERY_HOW_MANY_TIMES) == (SLEEP_0_EVERY_HOW_MANY_TIMES - 1))
+                else if ((yieldsSoFar % SLEEP_0_EVERY_HOW_MANY_TIMES) == (SLEEP_0_EVERY_HOW_MANY_TIMES - 1))
                 {
                     Thread.Sleep(0);
                 }
-                //else
-                //{
-                //    Thread.Yield();
-                //}
+                else
+                {
+                    // Polyfill note:
+                    // Thread.Yield() is not available in the 2.0 CLR, and the native implementation
+                    // is much more nuanced than p/invoking SwitchToThread.
+                    // Replace with Thread.Sleep(1) since Thread.Sleep(0) kills performance by often not actually yielding.
+                    // http://joeduffyblog.com/2006/08/22/priorityinduced-starvation-why-sleep1-is-better-than-sleep0-and-the-windows-balance-set-manager/
+                    Thread.Sleep(1);
+                }
             }
             else
             {


jnm2:
I have this branch and a clone of master side by side on my machine running net35 tests in an infinite loop to see if the hang reproduces.

After ten minutes, neither one had hung, so I ran four `while (true) { }` LINQPAD queries to burn CPU **and immediately got a hang from `master` while this PR branch is still going strong.**

jnm2:
I have this branch and a clone of master side by side on my machine running net35 tests in an infinite loop to see if the hang reproduces.

After ten minutes, neither one had hung, so I ran four `while (true) { }` LINQPad queries to burn CPU and **immediately** got a hang from `master`, while this PR branch is **still going strong.** This is exciting to me.

Let's merge now and leave EventQueue alone?

jnm2:
I have this branch and a clone of master side by side on my machine running net35 tests in an infinite loop to see if the hang reproduces.

I have to retract my earlier comment—I had the two windows switched. 🤦‍♂️

I'm going to try replacing EventQueue's Sleep(0) with Sleep(1) on net35- and Yield on net40+.

jnm2:
I went ahead and replaced all Sleep(0) with Sleep(1). I have that and and a clone of master side by side on my machine running net35 tests in an infinite loop to see if the hang reproduces.

So far the master branch has not hung, and this branch (plus using Sleep(1) in EventQueue) has hung twice. I'm not on the right track with Sleep(0) here.

jnm2:
Master branch finally hung after 54 runs back to back. Wonder if it'll happen with tracing on.

jnm2:
Hey, came back to it and the 26th run has hung, complete with trace.

jnm2:
Complete trace from the hanging net35 run:

[InternalTrace.16552.log](https://github.com/nunit/nunit/files/2381731/InternalTrace.16552.log)
[InternalTrace.19284.nunit.framework.tests.dll.log](https://github.com/nunit/nunit/files/2381732/InternalTrace.19284.nunit.framework.tests.dll.log)
[nunit-agent_19284.log](https://github.com/nunit/nunit/files/2381733/nunit-agent_19284.log)

jnm2:
Complete trace from the hanging net35 run:

[InternalTrace.16552.log](https://github.com/nunit/nunit/files/2381731/InternalTrace.16552.log)
[InternalTrace.19284.nunit.framework.tests.dll.log](https://github.com/nunit/nunit/files/2381732/InternalTrace.19284.nunit.framework.tests.dll.log) ⬅ the interesting one
[nunit-agent_19284.log](https://github.com/nunit/nunit/files/2381733/nunit-agent_19284.log)

For reference, here's one of the 25 runs that did not hang:
[InternalTrace.13064.nunit.framework.tests.dll.log](https://github.com/nunit/nunit/files/2381741/InternalTrace.13064.nunit.framework.tests.dll.log)

@CharliePoole I'm not spotting anything by comparing the two traces. What would you look for?

mintsoft:
Whoa, this will absolutely hammer parallel performance. I keep an eye on NUnit looking out for any regression around this because it took so long to get it fixed! Just to clarify the problem that was resolved with https://github.com/nunit/nunit/commit/7bdec91908ede8e9e78a6ef737581b9855dc4aca 

The duration that `Thread.Sleep(x)` sleeps for isn't the number of ms that are passed in, it's the smallest clock resolution larger than the value passed in. So the default clockres is `15.625ms` meaning that if you sleep for `1`ms, you might as well sleep for `15.625ms`. 

It's definitely worth noting that visual studio changes the `clockres`, so for performance testing you need to use a completely clean machine and confirm using the sysinternals clockres tool (https://download.sysinternals.com/files/ClockRes.zip) that nothing has changed your kernel settings; otherwise it'll work fine on your machine and be terrible on CI environments and the like :(

jnm2:
@mintsoft This would only affect tests targeting net20 and net35 since Thread.Yield is not a option and we actually need to force a yield (which Sleep(0) does not). However, it didn't fix the hang, so I'm starting from scratch anyway.

jnm2:
Adding here in case it's relevant. Got a hang on a CI build at work and I had the trace on since this is the fifth hang in the last couple weeks. The last thing reported was that Parallel workers 1, 2 and 4 were all executing OneTimeTearDowns.

(I don't think this is related, but more info on that: there's no OneTimeTearDown attribute, but three classes inherit from the same base class and the base class is IDisposable. So three derived fixture instances were being disposed when we stopped hearing from workers 1, 2, and 4.)

Parallel worker 3 last showed up halfway up the trace, executing a parameterized test case along with the other workers.

CharliePoole:
@jnm2 Looked at the traces you linked.

If you look at line 10981 in the log for the hung test run, you see the entries

```
[48] TestWorker: NonParallelWorker executing TestFixture1
[59] TestWorker: NonParallelWorker executing TestFixture2
```

So, two nonparallel workers are running on two separate threads. The corresponding code in the non-hanging run starts at line 10980. TestFixture1 runs to completion and then, further down, TestFixture2 runs, as expected, on the same thread.

It looks to me as if the situation may arise from our running second-level tests (e.g. loading and running mock-assembly) using the same dispatcher as we do the primary tests. We are probably just not isolating some of these tests enough and causing the problem.

indy-singh:
> tl;dr `Thread.Sleep(0)` hogs the CPU and is a terrible replacement for `Thread.Yield()`. `Thread.Sleep(1)` is probably the best we can do on .NET CLR 2.0.
> 
> See the comment in the (small) diff. I'm no expert here, so I could be missing something.
> Sources:
> 
> * http://joeduffyblog.com/2006/08/22/priorityinduced-starvation-why-sleep1-is-better-than-sleep0-and-the-windows-balance-set-manager/
>   > In our example above when Sleep(0) is used, we hope this will unstick the machine and let the producer produce and finally the consumer to consume. Indeed with some testing, we see it unstick after a little more than 3 seconds. This is still long enough, however, to kill performance on a server application, cause a noticeable perf degradation on the client, and destroy responsiveness in a GUI app.
> * https://blogs.msdn.microsoft.com/khen1234/2006/02/02/sleeping-vs-yielding/
> 
> /cc @rprouse @oznetmaster @CharliePoole

My understanding is that the effect of using `Thread.Sleep(0)` is dependent on the version of Windows used.

Post-Windows XP:-

> A value of zero causes the thread to relinquish the remainder of its time slice to **any other thread that is ready to run**. If there are no other threads ready to run, the function returns immediately, and the thread continues execution.

Pre-Windows XP:-

> A value of zero causes the thread to relinquish the remainder of its time slice to **any other thread of `equal priority` that is ready to run**. If there are no other threads of equal priority ready to run, the function returns immediately, and the thread continues execution. This behavior changed starting with Windows Server 2003.

Sources: https://stackoverflow.com/questions/16584358/is-thread-sleep1-special & https://docs.microsoft.com/en-gb/windows/desktop/api/synchapi/nf-synchapi-sleep

jnm2:
@indy-singh Good, thanks for bringing that to our attention. I understand that Windows XP has been out of support for a while, so I'm hesitant to encourage its use.

jnm2:
I've lost track of what I was doing here, but it looks like this didn't really make any progress towards fixing the hangs.

oznetmaster:
However, Sleep(1) is still the correct polyfill for Thread.Yield().  Sleep(0) is just plain bad.

This is documented in many places, including a number of Microsoft blogs and articles.

jnm2:
I think you're right even though it hurts performance for anyone using the .NET 3.5 build of NUnit. (@mintsoft, are you using the .NET 3.5 build?) Plus, hurt performance is better than hangs on .NET 3.5 and there may still be a slight chance that Sleep(0) exacerbates that.

@rprouse What do you think, should we go ahead and get rid of Sleep(0) from the codebase even though it doesn't stop the hangs, just so the code is correct?

CharliePoole:
FWIW, @oznetmaster is correct. What's more, I thought we had already decided to do that in the earlier discussion.

jnm2:
@CharliePoole There are two places where we are using Thread.Sleep(0). One is resolved in this PR (the Sleep(0) it leaves is following the newer .NET Framework design) and the other is EventQueue, where you [approved](https://github.com/nunit/nunit/pull/2233#issuecomment-310475607) a change from Sleep(1) to Sleep(0) and mentioned that you'd like to see the sleep removed at some future point.

CharliePoole:
@jnm2 I'm confused. I thought you were closing this PR without merging, in which case it won't resolve anything.

WRT the `Sleep(0)` in `EventQueue`, I'd say the same thing today if asked. Nothing has been done AFAIK.

jnm2:
👍 I meant "would be resolved" since I was planning to reopen once I knew it was only this instance of Sleep(0) that we wanted to change.

rprouse:
I'm fine to continue with this but you said way back in 2018 that this didn't solve the problem for you as is. Is this PR still useful for the comments and as something to build on?

jnm2:
@rprouse I was doing it in an attempt to fix the occasional hangs. I think I confirmed that it did not. I'm currently trying out the MyGet build from https://github.com/nunit/nunit/pull/3281 because those changes make a lot of sense to me as a fix for the hangs. I'm not planning on doing anything more to pursue the hang in this thread. If Charlie hasn't already fixed it, I'd open a new issue if I started digging in more.

mintsoft:
> I think you're right even though it hurts performance for anyone using the .NET 3.5 build of NUnit. (@mintsoft, are you using the .NET 3.5 build?) Plus, hurt performance is better than hangs on .NET 3.5 and there may still be a slight chance that Sleep(0) exacerbates that.

@jnm2 nope, .net 4 all the way

jnm2:
@mintsoft This pull request only affects our .NET 3.5 DLL, so I'm thinking that there's no way this change could cause issues for you?

jnm2:
@rprouse I think I have the hang figured out in the other thread and it has nothing to do with this thread. I'm ready for this to be merged whenever you are.

mintsoft:
> @mintsoft This pull request only affects our .NET 3.5 DLL, so I'm thinking that there's no way this change could cause issues for you?

Sure that's probably true, it probably _would_ negatively affect anyone using .NET 3.5 with a clean CI environment though?

jnm2:
My inclination would be to say the loss of performance goes with the territory of .NET 3.5 since the 11-year-old framework is missing the APIs needed to get this right otherwise. But I'll defer to others on the correctness issue. I'm just reading docs; they are the ones who have experience with it.

mintsoft:
> My inclination would be to say the loss of performance goes with the territory of .NET 3.5 since the 11-year-old framework is missing the APIs needed to get this right otherwise. 

Good point, absolutely fair imo!

PR #3021, opened 2018-09-16
Updated nUnit-VS-Adapter to 3.10.0

### LINKED ISSUE #3020: Upgrade nunit-vs-adapter to 3.10 for nUnit 3.11
Should we upgrade nunit-vs-adapter to latest version 3.10. I have been having issues with running tests for .Net Core and using this fixes my issues for this reason :

https://github.com/nunit/nunit3-vs-adapter/issues/325

I'm happy to fix and PR if agreed.

DaiMichael:
Fixes #3020, I had considered updating the ConsoleRunner to version 3.10.0-dev-4086 but as it was only added 9 days ago, I'll leave at the moment but happy to do so in this PR or in another issue. I know PR's are probably a bit of work for you but on the whole I prefer to commit work seperatly so roll backs are clean.

Anyway reference updated to 3.10.0

FYI - @jnm2 - Thanks for the help on getting to debug core, I had tried that and a miriade of other things. Turns out unistalling a whole bunch of old preview core sdk's, then I did a clean install and worked from there. Your help meant I knew it was environmental rather than me being a 🤡. This was the final fix 😄

### DIFF
--- a/src/NUnitFramework/tests/nunit.framework.tests.csproj
+++ b/src/NUnitFramework/tests/nunit.framework.tests.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFrameworks>net20;net35;net40;net45;netcoreapp1.1;netcoreapp2.0</TargetFrameworks>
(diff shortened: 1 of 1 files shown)


mikkelbu:
Merged. Thanks for the contribution @DaiMichael  👍 .

PR #3028, opened 2018-09-21
Try to fix travis builds

nvborisenko:
Sometimes travis build is timed out. Following [this](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) I added `travis_wait` for build command.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,4 @@ matrix:
       osx_image: xcode9.1
 
 script:
-  - ./build.sh --target=Travis --configuration=Release
+  - travis_wait ./build.sh --target=Travis --configuration=Release


CharliePoole:
Note this comment on the same Travis doc page linked above:

```
We recommend careful use of travis_wait, as overusing it can extend your build time when there could be a deeper underlying issue. 
```

I'm not sure this is a great idea. Recent successful builds have taken around  18 minutes. This includes the compile, half a dozen test steps and the final creation of artifacts. Each of those steps produces output. Before doing this, I would want to see a few sample failures that indicate we ran out of time in a build that would have succeeded if it had more time. Waiting longer to find out that we are in an infinite loop or that something has crashed doesn't seem like a good idea.

nvborisenko:
Agree with you. Just trying to understand whether the issue is really related to console output. It's more realistic that deadlock occurs sometimes, in this case travis doesn't see output and abort job.

nvborisenko:
The issue is not related to console output. This build on [vsts](https://nunit.visualstudio.com/NUnit/_build/results?buildId=808&view=logs) also timed out. So the issue is related to something like infinite loop or deadlock.

PR #3036, opened 2018-09-30
Apartment state can't be used for .NET Standard 2.0 tests

### LINKED ISSUE #3035: Apartment state can't be used for .NET Standard 2.0 tests
Setting apartment state is available in .NET Standard 2.0, but isn't available for NUnit's .NET Standard 2.0's build.

MatthewBeardmore:
Closes #3035.

Enables apartment state for .NET Standard 2.0 builds. All tests pass under Windows, Unix may have some failing tests, but I think I've added platform includes in all the places they might.


jnm2:
@MatthewBeardmore In order to enable this, we are going to need to catch the PlatformNotSupportedException and set the test result to Error. Otherwise the entire test run gets skipped because of one test.

Something that I expect will help with precisely this is https://www.nuget.org/packages/Microsoft.DotNet.Analyzers.Compatibility/. Would you be so kind as to add this to Directory.Build.props?

```xml
    <PackageReference Include="
Microsoft.DotNet.Analyzers.Compatibility" Version="0.2.12-alpha" PrivateAssets="all" />
```

(More info at https://docs.microsoft.com/dotnet/standard/analyzers/portability-analyzer. I want to add all these analyzers to NUnit at some point.)

jnm2:
Oops, fixed link.

jnm2:
Specifically, we should expect `SetApartmentState` to throw PlatformNotSupportedException anywhere it shows up in the code base, and catch it and set the test as errored:

https://github.com/nunit/nunit/blob/45db3dcd7b10dfdb7ede8c1f793f7a5e81fdb51f/src/NUnitFramework/framework/Internal/Execution/TestWorker.cs#L155-L164

Let us know if you need help with this.

MatthewBeardmore:
I've captured all the PNSEs and it'll mark all of the tests that have apartment state's set as failing now with an informative message now, but I'm not really sure how to test this behavior in the unit tests.

jnm2:
I see your commit messages about the analyzer. I'm sorry! I forgot about the deprecation checking. That is something we'll have to do in a separate PR.

Hmm, macOS and Linux builds are stalling both the Azure and Travis build servers during the netstandard2.0 run.

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
```diff
@@ -157,7 +175,18 @@ public void Start()
             _workerThread = new Thread(new ThreadStart(TestWorkerThreadProc));
             _workerThread.Name = Name;
 #if APARTMENT_STATE
-            _workerThread.SetApartmentState(WorkQueue.TargetApartment);
+#if NETSTANDARD2_0
+            if (OSPlatform.CurrentPlatform.IsWindows)
+            {
+#endif
+                _workerThread.SetApartmentState(WorkQueue.TargetApartment);
```
jnm2: Rather an OS detection, the .NET team intends folks to use feature detection—triggering and catching PlatformNotSuportedException, and maybe caching the fact that the feature is not available.

jnm2: ~~Rather than OS detection, the .NET team intends folks to use feature detection—triggering and catching PlatformNotSupportedException, and maybe caching the fact that the feature is not available.~~

Well, I thought so, but https://blogs.msdn.microsoft.com/dotnet/2017/10/31/introducing-api-analyzer/#user-content-discovering-cross-platform-issues has `RuntimeInformation.IsOSPlatform(OSPlatform.Windows)`.

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
```diff
@@ -131,9 +136,22 @@ private void TestWorkerThreadProc()
                     // worrying about competing workers.
                     Busy(this, _currentWorkItem);
 
-                    // Because we execute the current item AFTER the queue state
-                    // is saved, its children end up in the new queue set.
-                    _currentWorkItem.Execute();
+#if NETSTANDARD2_0
+                    if (_queueNotSupportedDueToApartmentState)
+                    {
+                        string msg = "Apartment state cannot be set on non-windows platforms.";
```
jnm2: `Windows` is a brand name, so we should follow the brand's capitalization.

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItemDispatcher.cs
```diff
@@ -60,7 +60,16 @@ public void Start(WorkItem topLevelWorkItem)
 
 #if APARTMENT_STATE
             if (topLevelWorkItem.TargetApartment == ApartmentState.STA)
+            {
+#if NETSTANDARD2_0
+                if (!OSPlatform.CurrentPlatform.IsWindows)
+                {
+                    topLevelWorkItem.MarkNotRunnable("Apartment state cannot be set on non-windows platforms.");
```
jnm2: ‘Windows’ is a brand name, so we should follow the brand's capitalization. 

How about `on platforms other than Windows.`?

(Two places.)

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItemDispatcher.cs
```diff
@@ -60,7 +60,16 @@ public void Start(WorkItem topLevelWorkItem)
 
 #if APARTMENT_STATE
             if (topLevelWorkItem.TargetApartment == ApartmentState.STA)
+            {
+#if NETSTANDARD2_0
+                if (!OSPlatform.CurrentPlatform.IsWindows)
```
jnm2: This is a lot more expensive than `RuntimeInformation.IsOSPlatform(OSPlatform.Windows)` (needs `using OSPlatform = System.Runtime.InteropServices.OSPlatform`) or catching PlatformNotSupportedException.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -248,6 +248,16 @@ public virtual void Execute()
                     Test.RequiresThread ? "RequiresThread" : "different Apartment");
 
 #if APARTMENT_STATE
+#if NETSTANDARD2_0
+                if (!OSPlatform.CurrentPlatform.IsWindows)
+                {
+                    string msg = "Apartment state cannot be set on non-windows platforms.";
```
jnm2: Since we aren't checking `WorkQueue.TargetApartment != ApartmentState.MTA` before checking if we're on Windows, this means that no workers are started.

https://github.com/nunit/nunit/blob/cbeffad907943d4219a2f0cdcea55830d267225e/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs#L169-L172

I think this is the reason that the CI builds are hanging when not on Windows.

A good question is whether we should be setting the apartment state to MTA at all or leaving it Unknown and skipping the call to SetApartmentState.

jnm2:
@nunit/framework-team Right now we don't distinguish between initializing COM to MTA and leaving COM uninitialized. Every work item queue is set to either STA or MTA.

If @MatthewBeardmore adds a check to skip calling SetApartmentState when the queue's apartment state is set to MTA and we're on Windows, then 

|                         | Windows                | Other platforms         |
|-------------------------|------------------------|-------------------------|
| No `[ApartmentState]`   | Initializes COM to MTA | Does not initialize COM |
| `[ApartmentState(MTA)]` | Initializes COM to MTA | Does not initialize COM |
| `[ApartmentState(STA)]` | Initializes COM to STA | Errors                  |

Would this scenario be better? This is what I would expect:

|                         | Windows                     | Other platforms         |
|-------------------------|-----------------------------|-------------------------|
| No `[ApartmentState]`   | **Does not initialize COM** | Does not initialize COM |
| `[ApartmentState(MTA)]` | Initializes COM to MTA      | **Errors**              |
| `[ApartmentState(STA)]` | Initializes COM to STA      | Errors                  |

It would require starting a new queue to run tests that require CoInitializing to MTA, but that should be rare. Most tests don't care about COM.

ChrisMaddock:
Sounds reasonable to me - but we should make sure this is highlighted in the release notes. 👍

jnm2:
@MatthewBeardmore Okay, are you up for implementing scenario 2?

We're going to need a new dedicated MTA versions of everything STA here:

https://github.com/nunit/nunit/blob/cbeffad907943d4219a2f0cdcea55830d267225e/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs#L73-L109

An MTA version of this test will be needed:

https://github.com/nunit/nunit/blob/cbeffad907943d4219a2f0cdcea55830d267225e/src/NUnitFramework/tests/Internal/Execution/ParallelExecutionTests.cs#L425-L434

The existing tests in `ParallelExecutionTests` will continue checking that tests without an Apartment attribute use a non-apartment-specific queue.

jnm2:
@MatthewBeardmore There are a lot of tests. If this is bigger than you were anticipating, I'm comfortable merging as-is (scenario 1) and implementing scenario 2 separately.

Also, new MTA versions of existing STA tests and necessary updates to:

- https://github.com/nunit/nunit/blob/cbeffad907943d4219a2f0cdcea55830d267225e/src/NUnitFramework/tests/Attributes/ApartmentAttributeTests.cs
- https://github.com/nunit/nunit/blob/cbeffad907943d4219a2f0cdcea55830d267225e/src/NUnitFramework/tests/WorkItemTests.cs

- Each hit of STA in https://github.com/nunit/nunit/blob/cbeffad907943d4219a2f0cdcea55830d267225e/src/NUnitFramework/tests/Attributes/ApplyToTestTests.cs

- https://github.com/nunit/nunit/blob/cbeffad907943d4219a2f0cdcea55830d267225e/src/NUnitFramework/tests/Attributes/RequiresThreadAttributeTests.cs#L45-L59

- https://github.com/nunit/nunit/blob/cbeffad907943d4219a2f0cdcea55830d267225e/src/NUnitFramework/tests/Attributes/TimeoutTests.cs#L138-L142

- https://github.com/nunit/nunit/blob/cbeffad907943d4219a2f0cdcea55830d267225e/src/NUnitFramework/tests/Attributes/SingleThreadedFixtureTests.cs#L81-L99

- There are a few places where `ApartmentState.MTA` but it should now be `ApartmentState.Unknown`:

  https://github.com/nunit/nunit/blob/cbeffad907943d4219a2f0cdcea55830d267225e/src/NUnitFramework/tests/Internal/WorkShiftTests.cs#L57-L64

  Maybe make the ApartmentState parameter default to Unknown so that we can remove the `#if APARTMENT_STATE` and use `new WorkItemQueue(name, true)`?

MatthewBeardmore:
After looking at what you're proposing - I don't think I am. I don't have a good "big-picture" view of NUnit or how all of the test workers/queues are set up yet.

jnm2:
@MatthewBeardmore That is no problem at all. I opened https://github.com/nunit/nunit/issues/3041 to track moving from scenario 1 to scenario 2 above.

⚠ I think it would be a good thing to get this in the same release as https://github.com/nunit/nunit/pull/3036 so that we don't ship with `[ApartmentState(MTA)]` being a no-op on platforms other than Windows and then fix it six months later. @rprouse Is it safe for me to milestone this one as 3.12.0?

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
```diff
@@ -131,9 +139,22 @@ private void TestWorkerThreadProc()
...
```
jnm2: Message should be updated consistently with the other message updates.

src/NUnitFramework/framework/nunit.framework.csproj
```diff
@@ -28,6 +28,10 @@
     <Reference Include="System.Windows.Forms" />
     <Reference Include="WindowsBase" Condition="'$(TargetFramework)' != 'net20'" />
   </ItemGroup>
+
+  <ItemGroup>
+    <!--<PackageReference Include="Microsoft.DotNet.Analyzers.Compatibility" Version="0.2.12-alpha" PrivateAssets="all" />-->
+  </ItemGroup>
```
jnm2: Could you go ahead and remove this change, since it needs its own PR?

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -248,6 +251,19 @@ public virtual void Execute()
                     Test.RequiresThread ? "RequiresThread" : "different Apartment");
 
 #if APARTMENT_STATE
+#if NETSTANDARD2_0
+                if (!RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Windows))
+                {
+                    if(targetApartment != ApartmentState.Unknown)
```
jnm2: Could you do this check the other way around (targetApartment field first, then `IsOSPlatform`)?

Also, we need a space after the `if` keyword.

CI is still hanging except on Windows. Can you try using the check `targetApartment == ApartmentState.STA` instead of `targetApartment != ApartmentState.Unknown`?

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
```diff
@@ -157,7 +178,21 @@ public void Start()
             _workerThread = new Thread(new ThreadStart(TestWorkerThreadProc));
             _workerThread.Name = Name;
 #if APARTMENT_STATE
-            _workerThread.SetApartmentState(WorkQueue.TargetApartment);
+#if NETSTANDARD2_0
+            if (!RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Windows))
+            {
+                if (WorkQueue.TargetApartment != ApartmentState.STA)
```
jnm2: Could you do this check the other way around (WorkQueue.TargetApartment property first, then `IsOSPlatform`)?

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItemDispatcher.cs
```diff
@@ -60,7 +63,16 @@ public void Start(WorkItem topLevelWorkItem)
 
 #if APARTMENT_STATE
             if (topLevelWorkItem.TargetApartment == ApartmentState.STA)
+            {
+#if NETSTANDARD2_0
+                if (!RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Windows))
+                {
+                    topLevelWorkItem.MarkNotRunnable("Apartment state cannot be set on platforms other than Windows.");
```
jnm2: @CharliePoole Is marking the top-level work item as not runnable going to report a not-runnable error for every test that has an apartment state set?

I feel like this is going to short circuit and not report any test results. Maybe this is the reason macOS and Linux builds are hanging?

jnm2:
@MatthewBeardmore I decided to comment out the `#if NETSTANDARD2_0` and the Windows check and debug the net45 build. This makes the net45 build hang on Windows just like it's hanging on Linux and macOS.

Let's work this way until Windows doesn't hang, then put back the Windows check and the `#if NETSTANDARD2_0`. After that we should be able to expect that Linux and macOS will not hang, either.

CharliePoole:
Unless I misunderstand something, the PR seems to conflate using .NET Core with the windows platform. All the error messages state that apartment state is only supported on windows, which has never been true using mono.

jnm2:
@CharliePoole Mono meaningfully distinguishes between STA and MTA? I thought COM was a Windows-only technology?

jnm2:
I think the hang is because `_topLevelWorkItem.State` is `Running`.

https://github.com/nunit/nunit/blob/cbeffad907943d4219a2f0cdcea55830d267225e/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs#L313

jnm2:
@MatthewBeardmore I found a fix! Do you mind me pushing to this PR branch?

jnm2:
@CharliePoole That is a good point. Then, so that our .NET Standard 2.0 build doesn't block Mono from setting the apartment state on Linux, we should stop doing OS checks and check for PlatformNotSupportedException. Getting tests on this would be a good thing, but we aren't set up to test .NET Standard builds against Mono.

How about this message: "The apartment state cannot be set on this platform." Too vague?

jnm2:
@MatthewBeardmore I don't have a good way to debug on Unix either, so I'm just commenting out the Windows check. :-}

src/NUnitFramework/tests/Attributes/ApartmentAttributeTests.cs
```diff
@@ -47,6 +53,9 @@ public void TestWithRequiresSTARunsInSTA()
         [Test]
 #if THREAD_ABORT
         [Timeout(10000)]
+#endif
+#if PLATFORM_DETECTION && NETCOREAPP2_0
```
jnm2: `NETCOREAPP2_0` implies `PLATFORM_DETECTION`, so I'd prefer the `PLATFORM_DETECTION` check to be removed for all of these. If `NETCOREAPP2_0` ever stops implying `PLATFORM_DETECTION`, build failures are exactly what we want.

src/NUnitFramework/tests/Attributes/ApartmentAttributeTests.cs
```diff
@@ -47,6 +53,9 @@ public void TestWithRequiresSTARunsInSTA()
...
+        [Platform(Include = "Win")]
```
jnm2: All of these need to change to `"Win, Mono"` in case we ever run .NET Standard tests on Mono because Mono simulates apartment state.

CharliePoole:
@jnm2 Just saying "on this platform" seems like a good compromise. Maybe with a page in the docs that explains all the intricacies around ApartmentState.

jnm2:
@MatthewBeardmore To see how we're doing on Linux and macOS, click the Details link next to the Travis row in the checks section at the bottom of the PR and the click a build job to see a live log.

I wonder if switching to catching PlatformNotSupportedException instead of preemptively checking the OS will fix the problem where `[Platform]` isn't suppressing the remaining failing tests?

jnm2:
Are you up for switching over to catching PNSE instead of checking the OS? If so, I'll stay out of your way for now on your branch.

MatthewBeardmore:
Not immediately. I'll see if I have time tonight or tomorrow. The only problem I had with PNSE is that I couldn't determine a good way that would work with different languages to accurately determine which PNSE was being thrown.

jnm2:
Also: rather than merging before pushing, you can do `git pull --rebase` to replay your commits on top of whatever commits I had pushed. This keeps it linear and a little easier to follow.

MatthewBeardmore:
This should be the last of the changes. I switched it over to PNSE and turned off the failing unit tests on Unix+NetCore and added two tests that only run on Unix+NetCore to verify that those tests should be failing.

MatthewBeardmore:
So no tests are failing, but the run is failing? Any ideas on this?

jnm2:
@MatthewBeardmore Thanks for sticking at this one!

@nunit/framework-team I think I lean towards wanting failure instead of skipping when attempting to set apartment state on .NET Core off Windows, but we'd have to come up with a way to layer properly so that PlatformAttribute's skip overrides ApartmentAttribute's failure. What is your preference?

MatthewBeardmore:
That would be my opinion too, but I figured that should go in a different issue given it wasn't a trivial change.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

jnm2:
@MatthewBeardmore Since 3.11 is no longer around the corner, I'm more comfortable leaving the master branch in a state that I wouldn't consider deliverable. I don't think of this as ideal, but you have been through a lot in this one. @nunit/framework-team Okay with merging and committing to fixing before 3.12?

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
```diff
@@ -157,7 +158,13 @@ public void Start()
...
             _workerThread.Name = Name;
 #if APARTMENT_STATE
-            _workerThread.SetApartmentState(WorkQueue.TargetApartment);
+            try
+            {
+                _workerThread.SetApartmentState(WorkQueue.TargetApartment);
+            }
+            catch (PlatformNotSupportedException)
+            {
+            }
```
ChrisMaddock: I'm not sure silently catching the exception here is ok. I think this will mean that whenever tests are running in parallel, then STA tests will end up running on a thread which is not STA, with no error. 

I'm not sure of the correct solution to this problem, without digging into the code more I'm afraid - this isn't part of the code I'm too familiar

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
```diff
@@ -480,7 +481,18 @@ private void RunOnSeparateThread()
...
+                thread.SetApartmentState(apartment);
+            }
+            catch (PlatformNotSupportedException)
+            {
+                string msg = "Apartment state cannot be set on this platform.";
+                log.Error(msg);
+                Result.SetResult(ResultState.Skipped, msg);
+                WorkItemComplete();
+                return;
+            }
```
ChrisMaddock: Note the comments around line 220 of this class:
```
            // 2. If the test needs to run in a different apartment.
            // This should not normally occur when using the parallel
            // dispatcher because tests are dispatches to a queue that
            // matches the requested apartment. Under the SimpleDispatcher
            // (--workers=0 option) it occurs routinely whenever a
            // different apartment is requested.
```

This looks like a good fix for sequential tests, but I don't think it covers parallel tests.

ChrisMaddock:
> @nunit/framework-team I think I lean towards wanting failure instead of skipping when attempting to set apartment state on .NET Core off Windows, but we'd have to come up with a way to layer properly so that PlatformAttribute's skip overrides ApartmentAttribute's failure. What is your preference?

> That would be my opinion too, but I figured that should go in a different issue given it wasn't a trivial change.

> @nunit/framework-team Okay with merging and committing to fixing before 3.12?

I agree with all of these. It seems sensible to switch to failure once we can provide a workaround. We should document this, however - and I agree, _ideally_ it would go in before 3.12. 

I do think we need to resolve the problems mentioned above however - if what I've written is correct (always a change it's not!), then we currently have inconsistent behaviour between sequential and parallel tests.

CharliePoole:
Some thoughts...

ApartmentStateAttribute should be unusable on platforms where it is not defined. This will inconvenience users who build for multiple platforms, since they will need to use conditional compilation but it's the easiest way to give an error at compile time.

If the attribute is unavailable, then the TargetApartment for the work item should not be set. No work items should ever be queued to the parallel or non-parallel STA queues. If desired those queues could simply be omitted from the parallel dispatcher along with the shifts that run them. However, they could be created for now and simply not used until we decide to do another PR.

I don't think we should catch PlatformException when we try to set ApartmentState. Rather, we should ensure it never happens in our code.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
jnm2: This was catching the parallel case you mentioned in your other comment.

src/NUnitFramework/framework/Internal/Execution/TestWorker.cs
jnm2: There's probably a better way to do this, but this was the smallest change I could think of at the time. It works for both sequential and parallel. What happens is that the work item starts with the wrong apartment state and errors itself out when it finds out. (Should be commented.)

jnm2:
@CharliePoole If we don't add `ApartmentAttribute` to the .NET Standard build, folks on .NET Core will not be able to use `[Apartment]` on .NET Core, which is what https://github.com/nunit/nunit/issues/3035 is asking.

The only alternative is adding a new .NET Core 2.1 build target. But even there, we still need to catch PlatformNotSupportedException (or do OS checks) in case the .NET Core tests are not running on Windows. We can't escape having to do dynamic capability checks at runtime. Also, we can't do OS checks in the .NET Standard build or we'll block Mono from simulating apartment state on macOS and Linux.

Is avoiding PlatformNotSupportedException really worth adding a seventh build target?

CharliePoole:
@jnm2 Right... if we need to use the same build on different platforms then of course we can't do a compile-time check.

src/NUnitFramework/framework/Internal/Execution/WorkItem.cs
ChrisMaddock: Ah, I understand - `needsNewThread` is being used to create a new thread for tests which would normally be run directly on the Worker. Thanks for explaining - I agree it could do with a comment, I think it really adds a third case to the above! 🙂

ChrisMaddock:
Thanks for the explanation @jnm2. This looks good to me as a first pass, although I think we should aim to not have that STA queue at all, when it's not available. But that sounds good to wait for a separate PR.

One last thing that's not clear to me - do we really need Skipped? Do WorkItems which have been marked by a platform attribute to not run really get as far as `WorkItem.Execute()`? I would have thought they would have been filtered out much earlier - but I guess I'm wrong!

jnm2:
@ChrisMaddock The hard part is, how do we go through the tree of tests and mark them all invalid? We're getting this for free right now by letting it go so far as to try to execute each one, but we'd need a secondary mechanism to walk the tree if we didn't even queue the work.

MatthewBeardmore:
I would have thought so too @ChrisMaddock, but they get that far, while apartment state causes it to fail far sooner - the opposite of what I would have personally expected.

Not having the STA queue would be good, and I'll see if I can work on that with a separate issue and pull request sometime soon.

CharliePoole:
WorkItems for Skipped tests have to execute in order to create a test result, which will have an outcome of Skipped. The code of the test itself is never executed, so it could be run before the point where a new thread is created in the correct apartment.

Code could be changed so that Skipped tests are run before the need for a new thread is considered. I think this would be best done before any other changes.

ChrisMaddock:
You learn something every day! 😅

With all of that, is this ready to merge? @jnm2 - will you hit the button if you’re happy? 😄

radiy:
Hello,

I am porting some winforms tests to .net core 3 and ApartmentState.STA blocks me from moving forward.
Can I help somehow to move this forward?

ChrisMaddock:
@jnm2 - this one looked good to me - are you happy for it to be merged? 😊

batzen:
Any news on this one?
It blocks me from using nunit for my .net core 3.0 WPF projects.

jnm2:
I'm so sorry! How did I totally fail to see the mentions. 🤦‍♂️ Looks like it's ready to merge, so I'll get it done for 3.12.

jnm2:
@MatthewBeardmore Thanks for your hard work!

Issue #3013, opened 2018-09-11
Provide a way to use a custom IWorkItemDispatcher


cwensley:
Currently there is only three options for dispatchers when using `NUnitTestAssemblyRunner`:

1. MainThreadWorkItemDispatcher
2. ParallelWorkItemDispatcher
3. SimpleWorkItemDispatcher

For UI-based applications, it would be nice to be able to invoke the tests in the UI thread, while running the unit tests in a background thread.  Currently, the only solution is to do an Invoke in each test itself, which is onerous (and messy to catch assertion exceptions and re-throw them after the invoke).

Alternatively, it would be nice to have an attribute that could specify that a test (or test fixture) should be run using `SynchronizationContext.Send()` from the `SynchronizationContext.Current` of the UI thread.

CharliePoole:
The dispatcher is global and used for the entire run so it's not suitable to replace for a single test or fixture.

@jnm2 Didn't you do some work on this problem?

cwensley:
@CharliePoole ah good to know, thanks!  I was thinking that I would be able to implement said attribute using a custom dispatcher, but perhaps that isn't the right place.  

Any hints of how/where something like that might be implemented would be very helpful.

jnm2:
This seems related to https://github.com/nunit/nunit/issues/2917 (which I need to get to) and https://github.com/nunit/nunit/issues/1200 (which is about to be released in 3.11).

@cwensley Would `[Apartment(STA)]` solve your problem when you use 3.11 with https://github.com/nunit/nunit/issues/1200? I typically do UI right in the test thread, sometimes in parallel.

cwensley:
@jnm2, Hm, with #1200 it appears I would need to return `Task` and use async/await?  I have no awaitable code, and it appears that the test itself is still initially called in the non-UI thread.

Also, it appears `[Apartment(ApartmentState.STA)]` isn't supported in netstandard, so I can't use it anyway.

#2917 is definitely what I'm after.

Thanks for the quick responses, guys.

cwensley:
Hmm.  Just poking around more, and it looks like creating an attribute that implements `IWrapTestMethod` can potentially be used for this scenario.   I'll test it out and report back.

cwensley:
Ok, it works!  I've used this code:

```c#
[System.AttributeUsage(AttributeTargets.Method, Inherited = true, AllowMultiple = false)]
[...]

Issue #2917, opened 2018-07-02
Extensibility of MessagePumpStrategy


garuma:
We have been maintaining a fork (https://github.com/mono/guiunit) and a not-so-big fork (https://github.com/garuma/GuiUnitNg) of NUnit to allow us to run unit-tests in the context of a GUI mainloop. 

(For context, we previously logged an enhancement idea for this: https://github.com/nunit/nunit/issues/1227)

Looking at the recent code addition of `MessagePumpStrategy` (and rework of `AsyncInvocationRegion`) it seems like this is pretty much tackling the same thing.

Unfortunately the current implementation will only allow the hardcoded loops that are available on .NET (Winforms + WPF) with no way to extend this to Gtk+, Mac or others.

Of course the goal is not to add those implementation necessarily in the core of NUnit but at least allow an extension point of some sort where users can inject their own message loop for their toolkit of choice.

Additionally, I'm not quite sure what's the strategy to be able to "proxy" the test method execution to said main loop (i.e. up to the first `await`) so that `SynchronizationContext.Current` is valid and in practice only `ApartmentState.STA` seems to trigger a synchronization context to be set.

rprouse:
When we first added the MessagePumpStrategy, I was concerned that we won't be able to cover all of the platforms. I like the idea of making them plugable though. Given an interface, it shouldn't be too hard to add, we just need to figure out how to expose it. @jnm2 what do you think?

jnm2:
This is something I brought up as part of https://github.com/nunit/nunit/issues/2774, but we were waiting for actual use cases, so this is exciting!

We'd want it to apply to async setup and teardown as well as async test methods, maybe even the proposed async test case sources. So what's the ideal way to annotate this, an attribute which could be applied to the fixture or test method or test case source, implementing an IMessagePumpStrategy interface?

Right now the message pump strategy is selected at the last minute depending on the current synchronization context. However, we may want to give the straw-man IMessagePumpStrategy an install method and an uninstall method so that you can opt to install your own synchronization context? And then the rest of the interface would be what MessagePumpStrategy currently has?

decriptor:
Any word on this?

jnm2:
I'm hoping to hear from others as many ideas as we can collect, and my free time is coming back soon!

jnm2:
I'm currently hoping to provide for this scenario by implementing:

- https://github.com/nunit/nunit/issues/2220, which gives you control of the async-to-sync adaptation, so you can start and stop the message pump
[...]

Issue #2220, opened 2017-06-06
Apply an IWrapSetUpTearDown attribute to all tests in a fixture


yaakov-h:
Similar to #2015, I'm trying to perform some setup and teardown code before the test's own setup and teardown code is run, and apply this to multiple test fixtures via an attribute.

I have tried a custom attribute which implements `IWrapSetUpTearDown`, but it's not called if I apply it to the entire test class, only if I apply it to each individual test. `ITestAction` is not suitable, as it does not wrap setup and teardown.

Is there any way to do this with the current NUnit framework, or if not, can we please do this in a future version of the framework? 😄 

Example:

<details>
Code:

```
using System;
using NUnit.Framework;
using NUnit.Framework.Interfaces;
using NUnit.Framework.Internal;
using NUnit.Framework.Internal.Commands;

namespace NUnitWrap
{
    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method | AttributeTargets.Interface | AttributeTargets.Assembly, Inherited = true, AllowMultiple = true)]
    sealed class SomeTestAttribute : Attribute, IWrapSetUpTearDown
    {
        public string Comment { get; set; }

        public TestCommand Wrap(TestCommand command)
            => new SomeTestCommand(command) { Comment = Comment };

        sealed class SomeTestCommand : DelegatingTestCommand
        {
            public SomeTestCommand(TestCommand innerCommand)
                : base(innerCommand)
            {
            }

            public string Comment { get; set; }

            public override TestResult Execute(TestExecutionContext context)
            {
                context.OutWriter.WriteLine($"Context: Before test - {Comment}");
                try
                {
                    return innerCommand.Execute(context);
                }
                finally
                {
                    context.OutWriter.WriteLine($"Context: After test - {Comment}");
                }
            }
        }
    }

    [SomeTest(Comment = "applied to class")]
    class WrappingTest
    {
        [Test]
        [SomeTest(Comment = "applied to method")]
        public void Test() => Assert.Pass();
    }
}
```

Output:

```
NUnit Console Runner 3.6.0
Copyright (C) 2017 Charlie Poole

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.15063.0
  CLR Version: 4.0.30319.42000

Test Files
    nunitwrap.dll

=> NUnitWrap.WrappingTest.Test
Context: Before test - applied to method
Context: After test - applied to method

Run Settings
    DisposeRunners: True
[...]

PR #3037, opened 2018-10-01
[RFC] Add extensibility to the async pipeline to allow custom implementation of waiting strategy

garuma:
/cc @jnm2 

Related to:
 - https://github.com/nunit/nunit/issues/2220
 - https://github.com/nunit/nunit/issues/3013
 - https://github.com/nunit/nunit/issues/2917

This PR is an example of extensibility mechanism for the async machinery allowing external people to customize its behavior via attributes.

By leveraging the existing `IWrapSetUpTearDown` interface, one can now fully execute unit-test inside any sort of mainloop (GUI-based or not).

For instance here is the set of custom attribute created to support running tests with [XWT](https://github.com/mono/xwt):

```csharp
// Run wrapper
[System.AttributeUsage (AttributeTargets.Class, Inherited = true, AllowMultiple = false)]
public class XwtCommandWrapperAttribute : Attribute, IWrapSetUpTearDown
{
	public TestCommand Wrap (TestCommand command) => new RunOnUICommand (command);
	class RunOnUICommand : DelegatingTestCommand
	{
		public RunOnUICommand (TestCommand innerCommand)
			: base (innerCommand)
		{
		}

		public override TestResult Execute (TestExecutionContext context)
		{
			TestResult result = null;
			var evt = new ManualResetEventSlim ();
			Xwt.Application.Invoke (() => {
				try {
					result = innerCommand.Execute (context);
				} catch (Exception e) {
					Console.WriteLine ("Catastrophic fail", e);
				}
				evt.Set ();
			});
			evt.Wait ();
			return result;
		}
	}
}

// Custom wait strategy
[System.AttributeUsage (AttributeTargets.Class, Inherited = true, AllowMultiple = false)]
public class XwtAsyncWaitStrategyAttribute : Attribute, IAsyncCompletionStrategy
{
	public void WaitForCompletion (IAwaiterObject awaitable)
	{
		while (!awaitable.IsCompleted)
			Xwt.Application.MainLoop.DispatchPendingEvents ();
	}
}
```

Additionally, this PR adds a `GetCombinedCustomAttributes` method to support looking up attribute definitions on both the test method and the containing fixture so that those two attributes above can be applied at the fixture level thus ensuring all the tests automatically inherit the behavior.

(Note that this way of doing things work for our own test suite with minimal changes to the nunit codebase but I don't have any strong feeling about going in that specific direction)

### DIFF
--- a/src/NUnitFramework/framework/Interfaces/IAsyncCompletionStrategy.cs
+++ b/src/NUnitFramework/framework/Interfaces/IAsyncCompletionStrategy.cs
@@ -1,0 +1,46 @@
+// ***********************************************************************
+// Copyright (c) 2014 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Collections.Generic;
+using NUnit.Framework.Internal.Commands;
+
+namespace NUnit.Framework.Interfaces
+{
+    /// <summary>
+    /// Implement this interface to customize the waiting behavior
+    /// of the test adapter for <c>async</c> test methods
+    /// </summary>
+    public interface IAsyncCompletionStrategy
+    {
+        /// <summary>
+        /// This method will be called by the adapter to wait for the completion
+        /// of an asynchronous test
+        /// </summary>
+        /// <remarks>
+        /// Implementers must ensure that the awaitable object provided has
+        /// completed before returning from this method
+        /// </remarks>
+        void WaitForCompletion(IAwaiterObject awaitable);
+    }
+}
--- a/src/NUnitFramework/framework/Interfaces/IAwaiterObject.cs
+++ b/src/NUnitFramework/framework/Interfaces/IAwaiterObject.cs
@@ -1,0 +1,42 @@
+// ***********************************************************************
+// Copyright (c) 2014 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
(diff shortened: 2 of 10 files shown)


src/NUnitFramework/framework/Interfaces/IAsyncCompletionStrategy.cs
```diff
@@ -0,0 +1,48 @@
...
+using System;
+using System.Collections.Generic;
+using NUnit.Framework.Internal.Commands;
+
+namespace NUnit.Framework.Interfaces
+{
+#if ASYNC
+    /// <summary>
+    /// Implement this interface to customize the waiting behavior
+    /// of the test adapter for <c>async</c> test methods
```
CharliePoole: "Test adapter" has a very specific meaning in the NUnit world. It's something outside the framework that __adapts__ the representation of tests in NUnit to some other form. The most obvious example is the VS Test Adapter.
In this context (i.e. within the framework itself) it has no particular meaning for me. Sorry not to suggest an alternative, but I don't know what you are trying to say.

CharliePoole: Looking down farther, I see somebody has added an `AwaitAdapter` in the past... I still don't like the term but I see that you aren't the one who introduced it. :confused:

garuma: Yep, "adapter" there refers to `AwaitAdapter` and `AsyncToSyncAdapter`. I can change to something else if the wording is troublesome.

jnm2:
Thanks @garuma. This PR is a great proof of concept because it shows that it doesn't take much code at all to get something basic working. I apologize for letting the time slip by. I want to work on this in two phases:

- Allowing user code to take control of test execution (https://github.com/nunit/nunit/issues/2220)
- Enabling sync-over-async strategies for message pumps other than .NET's built-in ones (https://github.com/nunit/nunit/issues/2917)

I'm going to comment in each of those places, for now.

jnm2:
@garuma More on https://github.com/nunit/nunit/issues/2917#issuecomment-430070519. I would like to avoid having this ever be necessary for an NUnit extender to implement:

```cs
// Custom wait strategy
[System.AttributeUsage (AttributeTargets.Class, Inherited = true, AllowMultiple = false)]
public class XwtAsyncWaitStrategyAttribute : Attribute, IAsyncCompletionStrategy
{
    public void WaitForCompletion (IAwaiterObject awaitable)
    {
        while (!awaitable.IsCompleted)
            Xwt.Application.MainLoop.DispatchPendingEvents ();
    }
}
```

Instead, you'd modify your first code sample to do something like this:

```c#
// Run wrapper
[System.AttributeUsage(AttributeTargets.Class, Inherited = true, AllowMultiple = false)]
public sealed class XwtCommandWrapperAttribute : Attribute, IWrapSetUpTearDown
{
    public TestCommand Wrap (TestCommand command) => new RunOnUICommand (command);

    private sealed class RunOnUICommand : DelegatingTestCommand
    {
        public RunOnUICommand(TestCommand innerCommand)
            : base (innerCommand)
        {
        }

        public override TestResult Execute(TestExecutionContext context)
        {
            TestResult result = null;
            
            Xwt.Application.Invoke(() =>
            {
                try
                {
                    result = innerCommand.Execute(context);
                }
                catch (Exception e)
                {
                    Console.WriteLine("Catastrophic fail", e);
                }
                
                Xwt.Application.Exit();
            });
            
            Xwt.Application.Run();
            
            return result;
        }
    }
}
```

This avoids the anti-pattern of calling `DispatchPendingEvents()` in a busy loop which I carefully avoided for the [Windows Forms](https://github.com/nunit/nunit/blob/6af1a1536e2ad86d3974b1167cb3c984b0c23f83/src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs#L85-L100) and [WPF](https://github.com/nunit/nunit/blob/6af1a1536e2ad86d3974b1167cb3c984b0c23f83/src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs#L118-L126) message loops.

Rather, the underlying framework should be using MsgWaitForMultipleObjectsEx or similar so that no CPU is consumed until an event occurs such as an event loop message or a synchronization context post.

jnm2:
@garuma I'm sorry for letting this go so long. I've started looking at it again.

I'm opposed to a busy loop, effectively doing this:

```cs
while (!awaitable.IsCompleted)
    Xwt.Application.MainLoop.DispatchPendingEvents();
```

It's far better to call Run and allow the UI stack to do a MsgWaitForMultipleObjectsEx or equivalent call so that the thread sleeps until the next UI message or SynchronizationContext work arrives in the UI message queue:

```cs
var synchronizationContext = new XwtSynchronizationContext();
SynchronizatonContext.Current = synchronizationContext;
synchronizationContext.Post(OnPost, state: ...);
Xwt.Application.Run();

void OnPost(object state)
{
    // (get synchronizationContext and awaitable from state)
    awaitable.OnCompleted(() =>
    {
        synchronizationContext.Post(_ => Xwt.Application.Exit(), state: null);
    });
}
```

I'm going to bring up an API that gets all the tricky parts of this correct so that you only have to point it at the methods `Xwt.Application.Run()` and `Xwt.Application.Exit()`.

jnm2:
(moved my comment to https://github.com/nunit/nunit/issues/2917#issuecomment-511069863)

OsirisTerje:
@jnm2 What is the state of this now, and the corresponding issues?   This is now 5 years old, but it may still have merit, right?

jnm2:
The problem is that a DoEvents loop doesn't behave the same as an Application.Run() call, with a queued action and queued Exit() following the action. I wanted the API to be a different shape so that people don't go down the DoEvents loop route.

OsirisTerje:
@jnm2 Do you believe this can be made to work, or should this PR be abandoned ?  It is now 5 years old......

CharliePoole:
This seems like a reason for concern...

> This branch is [1 commit ahead of](https://github.com/garuma/nunit/compare/nunit%3Anunit%3Amaster...mainloop-extensibility), [1326 commits behind](https://github.com/garuma/nunit/compare/mainloop-extensibility...nunit%3Anunit%3Amaster) nunit/nunit:master.

OsirisTerje:
I'll close this now.  
With the execution hooks we now have extensions points for Setup Teardown. I am not sure at all if that would help here, but if this is to be implemented, it has to be based on the current code base.

PR #3044, opened 2018-10-06
Lazy description creation in ComparisonConstraints

### LINKED ISSUE #2975: ComparisonConstraints are allocating string on construction
While investigating our test suite's memory usage, NUnit also came up from some paths where a lot of parsed data was processed through loop that checked against LessThanOrEqualConstraint. If you look how all ComparisonConstraint inheritors are implemented, they all allocate Description when constructed even though I would presume it's not needed unless assertion fails?

https://github.com/nunit/nunit/blob/6f806c61eb84f1fed0b47f9b90f1d1e21ec72e14/src/NUnitFramework/framework/Constraints/GreaterThanOrEqualConstraint.cs#L35-L38

I would suggest that these would be either lazy evaluated or evaluated every time the getter is called, like for example in EqualConstraint:

https://github.com/nunit/nunit/blob/6f806c61eb84f1fed0b47f9b90f1d1e21ec72e14/src/NUnitFramework/framework/Constraints/EqualConstraint.cs#L399-L421

madetara:
Fixes #2975 
Moved description creation to a base class.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
@@ -87,6 +87,18 @@ namespace NUnit.Framework.Constraints
             {
                 System.Text.StringBuilder sb = new System.Text.StringBuilder(DisplayMessageBase);
                 sb.Append(MsgUtils.FormatValue(_expected));
+                
+                if (_tolerance != null && !_tolerance.IsUnsetOrDefault)
+                {
+                    sb.Append(" within ");
+                    sb.Append(MsgUtils.FormatValue(_tolerance.Amount));
+
+                    if (_tolerance.Mode == ToleranceMode.Percent)
+                    {
+                        sb.Append(" percent");
+                    }
+                }
+                
                 return sb.ToString();
             }
         }
@@ -147,7 +159,6 @@ namespace NUnit.Framework.Constraints
                 throw new InvalidOperationException("Within modifier may appear only once in a constraint expression");
 
             _tolerance = new Tolerance(amount);
-            Description += " within " + MsgUtils.FormatValue(amount);
             return this;
         }
 
@@ -162,7 +173,6 @@ namespace NUnit.Framework.Constraints
             get
             {
                 _tolerance = _tolerance.Percent;
-                Description += " percent";
                 return this;
             }
         }
--- a/src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
@@ -22,6 +22,7 @@
 // ***********************************************************************
 
 using System;
+using System.Text;
 using System.Collections;
 using System.Collections.Generic;
 
@@ -162,7 +163,7 @@ namespace NUnit.Framework.Constraints
             if (comparisonText == null)
                 throw new ArgumentNullException(nameof(comparisonText), "Comparison text can not be null");
             
-            System.Text.StringBuilder sb = new System.Text.StringBuilder(comparisonText);
+            StringBuilder sb = new StringBuilder(comparisonText);
             sb.Append(MsgUtils.FormatValue(_expected));
                 
             if (_tolerance != null && !_tolerance.IsUnsetOrDefault)
--- a/src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
@@ -155,7 +155,7 @@ namespace NUnit.Framework.Constraints
         /// Provides standard description of what the constraint tests
         /// based on comparison text.
         /// </summary>
-        /// <param name="comparisonText">Describes, what constraint tests, throws <see cref="ArgumentNullException"/>
+        /// <param name="comparisonText">Describes the comparison being tested, throws <see cref="ArgumentNullException"/>
         /// if null</param>
         /// <exception cref="ArgumentNullException">Is thrown when null passed to a method</exception>
         protected string DefaultDescription(string comparisonText)
--- a/src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
@@ -78,37 +78,6 @@ namespace NUnit.Framework.Constraints
         }
 
         /// <summary>
-        /// The Description of what this constraint tests, for
-        /// use in messages and in the ConstraintResult.
-        /// </summary>
-        public override string Description
-        {
-            get
-            {
-                System.Text.StringBuilder sb = new System.Text.StringBuilder(DisplayMessageBase);
-                sb.Append(MsgUtils.FormatValue(_expected));
-                
-                if (_tolerance != null && !_tolerance.IsUnsetOrDefault)
-                {
-                    sb.Append(" within ");
-                    sb.Append(MsgUtils.FormatValue(_tolerance.Amount));
-
-                    if (_tolerance.Mode == ToleranceMode.Percent)
-                    {
-                        sb.Append(" percent");
-                    }
-                }
-                
-                return sb.ToString();
-            }
-        }
-
-        /// <summary>
-        /// Protected field overriden by derived class to provide basis for Description construction
-        /// </summary>
-        protected abstract string DisplayMessageBase { get; }
-
-        /// <summary>
         /// Protected function overridden by derived class to actually perform the comparison
         /// </summary>
         protected abstract bool PerformComparison(ComparisonAdapter comparer, object actual, object expected, Tolerance tolerance);
@@ -177,6 +146,39 @@ namespace NUnit.Framework.Constraints
             }
         }
 
+        #endregion
+        
+        #region Protected Methods
+        
+        /// <summary>
+        /// Provides standard description of what the constraint tests
+        /// based on comparison text.
+        /// </summary>
+        /// <param name="comparisonText">Describes, what constraint tests, throws <see cref="ArgumentNullException"/>
+        /// if null</param>
+        /// <exception cref="ArgumentNullException">Is thrown when null passed to a method</exception>
+        protected string DefaultDescription(string comparisonText)
+        {
+            if (comparisonText == null)
+                throw new ArgumentNullException(nameof(comparisonText), "Comparison text can not be null");
+            
+            System.Text.StringBuilder sb = new System.Text.StringBuilder(comparisonText);
+            sb.Append(MsgUtils.FormatValue(_expected));
+                
+            if (_tolerance != null && !_tolerance.IsUnsetOrDefault)
+            {
+                sb.Append(" within ");
+                sb.Append(MsgUtils.FormatValue(_tolerance.Amount));
+
+                if (_tolerance.Mode == ToleranceMode.Percent)
+                {
+                    sb.Append(" percent");
+                }
+            }
+                
+            return sb.ToString();
+        }
+        
         #endregion
     }
 }
--- a/src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
@@ -78,6 +78,25 @@ namespace NUnit.Framework.Constraints
         }
 
         /// <summary>
+        /// The Description of what this constraint tests, for
+        /// use in messages and in the ConstraintResult.
+        /// </summary>
+        public override string Description
+        {
+            get
+            {
+                System.Text.StringBuilder sb = new System.Text.StringBuilder(DisplayMessageBase);
+                sb.Append(MsgUtils.FormatValue(_expected));
+                return sb.ToString();
+            }
+        }
+
+        /// <summary>
+        /// Protected field overriden by derived class to provide basis for Description construction
+        /// </summary>
+        protected abstract string DisplayMessageBase { get; }
+
+        /// <summary>
         /// Protected function overridden by derived class to actually perform the comparison
         /// </summary>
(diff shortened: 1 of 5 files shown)


src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
```diff
@@ -77,6 +77,37 @@ public override ConstraintResult ApplyTo<TActual>(TActual actual)
...
+                }
+                
+                return sb.ToString();
+            }
+        }
+
+        /// <summary>
+        /// Protected field overriden by derived class to provide basis for Description construction
+        /// </summary>
+        protected abstract string DisplayMessageBase { get; }
```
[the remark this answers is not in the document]
CharliePoole: How flexible does it have to be? It's the base class for four of our classes... the four comparison constraints. It's not a class we offer for extension as far as I can see.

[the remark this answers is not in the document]
jnm2: How will folks know they're not intended to subclass `ComparisonConstraint`?

Typically that design decision is implemented using `internal abstract` (or better, `private protected abstract` in C# 7.0) so that the only assemblies that can extend it are assemblies with internal access to this one.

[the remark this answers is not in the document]
CharliePoole: This is a general discussion you and I have been having for a few years now and it continues to infect discussions of particular changes, where the person performing the work if the change just needs to be told what to do.

We won't resolve our different views of what constitutes an API here - probably ever - so let's not go there on this issue. 

Instead, why not just have this developer change the signature of the base class?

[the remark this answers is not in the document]
jnm2: "Infect" is not a diplomatic word. 😆

By changing the signature, you mean asking for `internal virtual` instead of `protected abstract`? Okay, but it's not my place (or desire) to make unilateral decisions. I would like to solicit thoughts like yours from the team. Definitions of APIs aside, is there something which differentiates ComparisonConstraint from Constraint in such a way that

- we expect no users to have inherited from `ComparisonConstraint`
- or, we want users who inherit to be required by the compiler to figure out what we did and change their programs when they upgrade?

Decisions can certainly be made by default but to me it's worth being sure that we like the outcome of this decision.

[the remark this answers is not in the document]
CharliePoole: @jnm2 You should see the words I rejected. :smiley:

Seriously when we (you and I) have these broader discussions, the rest of the team loses interest in the thread. That's why I call it an infection. So let's just not do that.

As you well know, I no longer vote or take part in discussions of where nunit will go in the future, because I'm not a part of that future. I do try to strike a balance by giving you all some info from the past where it seems necessary from my perspective. It may be that that's not working, in which case I'll stop.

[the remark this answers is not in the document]
jnm2: And yet I specifically asked _not_ to have a broader discussion about APIs. Can we focus on this single, direct outcome of `protected abstract`? Like I asked above: 1) Will it affect anyone? 2) If so, do we like the effect, or should we use a less obvious alternative like `internal` or `virtual`?

@nunit/framework-team If your preference is that I skip asking this kind of end user cause-effect question and simply request our (awesome) contributor to take what I think is the safer route, I can sure do that.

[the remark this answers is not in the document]
madetara: Thank you for the detailed discussion. 
It was my bad not paying attention to the class being public and possibility (even slightest) of it being used in someone else's code. I guess @jnm2 has the point here, and it will be better solution to make it virtual in order to not break things up.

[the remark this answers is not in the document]
CharliePoole: If you mean defining the `DescriptionBase` property as protected abstract, it seems fine to me as he has done it. You seemed to be critiquing that change because that's where your comment took off from.

If you are suggesting changing the signature of `ComparisonConstraint`, that's what seems to me to be out of scope here and part of a broader discussion of how we handle classes that are intended only to be inherited from by other classes of ours. That's the broader discussion I thought you were trying to initiate, even while saying you didn't want to initiate it.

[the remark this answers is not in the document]
jnm2: @CharliePoole Yes, I'm only talking about `DescriptionBase`. I'm not sure what "changing the signature of ComparisonConstraint" means; did it sound like I was suggesting making the whole class internal? That's wouldn't compile because public classes can only inherit from other public classes. My bad if that wasn't clear.

@madetara There's two things I saw. One was that custom inheritors would have to implement that method even though they're currently overriding `Description` if they are setting the description at all.  Different team members have expressed a desire to minimize required changes like this unless the custom implementation is already buggy in the previous release.

The other thing is that we can't remove `DescriptionBase` in the future, even if we find out down the road that we need something else and `DescriptionBase` becomes vestigial. This isn't a huge problem, just worth a passing thought.

A tactic that doesn't require derived classes to change their understanding of how the base class works might be something like adding an internal or protected (sealed) member `string DefaultDescription(string comparisonText)` to ComparisonConstraint. Derived classes could then override Description with `=> DefaultDescription("greater than")`, etc.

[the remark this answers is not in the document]
mikkelbu: > @nunit/framework-team If your preference is that I skip asking this kind of end user cause-effect question and simply request our (awesome) contributor to take what I think is the safer route, I can sure do that. So far I've asked for your input because I like hearing from you and I'm just one member here.

Personally, I value that you catch these kinds of issues (I think that I would have overlooked it). I'm happy with you requesting the contributor to do the changes, then we can always discuss if we disagree. At least for me I think that I would look just as closely on a normal PR review comment as on the original comment https://github.com/nunit/nunit/pull/3044#discussion_r223183731, but of course it also depends on scale/complexity of the changes (the larger the more need for discussion and consensus).

Ps. I like the last variant proposed (which is also implemented now).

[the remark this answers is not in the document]
jnm2: @mikkelbu Okay, sounds like a plan! Thanks!

src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
```diff
@@ -143,11 +142,43 @@ public ComparisonConstraint Percent
...
+        /// </summary>
+        /// <param name="comparisonText">Describes, what constraint tests, throws <see cref="ArgumentNullException"/>
+        /// if null</param>
+        /// <exception cref="ArgumentNullException">Is thrown when null passed to a method</exception>
+        protected string DefaultDescription(string comparisonText)
+        {
+            if (comparisonText == null)
+                throw new ArgumentNullException(nameof(comparisonText), "Comparison text can not be null");
+            
+            System.Text.StringBuilder sb = new System.Text.StringBuilder(comparisonText);
```
jnm2: We would typically add `using System.Text;` it the namespace is used more than once.

src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
```diff
@@ -143,11 +142,43 @@ public ComparisonConstraint Percent
...
+            sb.Append(MsgUtils.FormatValue(_expected));
```
jnm2: 💭 Side note: we should think about having MsgUtils write directly into a StringBuilder.

src/NUnitFramework/framework/Constraints/GreaterThanConstraint.cs
```diff
@@ -21,22 +21,37 @@
...
 {
     /// <summary>
     /// Tests whether a value is greater than the value supplied to its constructor
     /// </summary>
     public class GreaterThanConstraint : ComparisonConstraint
     {
+        /// <summary>
+        /// The Description of what this constraint tests, for
+        /// use in messages and in the ConstraintResult.
+        /// </summary>
```
jnm2: I would recommend not putting XML comments on private members because the end users will never see it.
(Same in other files.)

src/NUnitFramework/framework/Constraints/GreaterThanConstraint.cs
```diff
@@ -21,22 +21,37 @@
...
+        public override string Description
         {
-            Description = "greater than " + MsgUtils.FormatValue(expected);
+            get
+            {
+                if (System.String.IsNullOrEmpty(_description))
```
jnm2: Please use the `string` keyword rather than the `String` class name. But before you do that, why isn't this `_description == null`?
(Same in other files.)

src/NUnitFramework/framework/Constraints/GreaterThanOrEqualConstraint.cs
```diff
@@ -28,13 +28,30 @@ namespace NUnit.Framework.Constraints
...
+        /// use in messages and in the ConstraintResult.
+        /// </summary>
+        public override string Description
         {
-            Description = "greater than or equal to " + MsgUtils.FormatValue(expected);
+            get
+            {
+                if (System.String.IsNullOrEmpty(_description))
+                    _description = DefaultDescription("greater than or equal to ");
+                return _description;
```
jnm2: 💡 If you wanted to, you could put `_description` in the base class (still private) and move the lazy initialization logic into `DefaultDescription`. I don't have an opinion one way or the other on this.

src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
madetara: But it's only used once in `ComparisonConstraint`, from what I see. Thanks for the tip though.

src/NUnitFramework/framework/Constraints/GreaterThanConstraint.cs
madetara: It seemed more "clear" to me to use `string.IsNullOrEmpty` rather than `_desciption == null`, which is probably better from the point of performance.

src/NUnitFramework/framework/Constraints/GreaterThanOrEqualConstraint.cs
madetara: Since the base concept here is overriding the `Description`, it seems, in my opinion, more clear to have corresponding field into the inheritor, rather than the base class, even though it has a smell of code duplication.

src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
madetara: Didn't quite get the idea.
If you are talking about adding a method to `MsgUtils`, which interacts with `StringBuilder` directly, I can work on it after closing that issue.

src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
jnm2: I count it as used twice in `System.Text.StringBuilder sb = new System.Text.StringBuilder(...);`.

src/NUnitFramework/framework/Constraints/GreaterThanConstraint.cs
madetara: Thanks for pointing it out. 
Why private members in `ComparisonConstraint` are documented then? Not arguing, just trying to understand the documenting process, to avoid such mistakes later on.

src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
jnm2: Sorry, I should have @'ed the team. This is an idea for the future that I didn't mean to ask you to do. (Unless you feel like it.)

src/NUnitFramework/framework/Constraints/GreaterThanConstraint.cs
jnm2: If you use IsNullOrEmpty, it signals to the reader that there is a reason that both states (null and empty) might show up. I think we should decide that only one of those states is treated as 'uninitialized' so that there's less to keep in your head as you read through the code.

src/NUnitFramework/framework/Constraints/GreaterThanConstraint.cs
jnm2: > Why private members in ComparisonConstraint are documented then?

I'm not sure why that happened. They shouldn't be documented if they are directed at end users. Occasionally an XML summary makes sense on a private member of a large class, but only if it's information directed at users who are editing the class rather than end users.

src/NUnitFramework/framework/Constraints/GreaterThanConstraint.cs
madetara: Thank you for your advices. I appreciate it. Will try to follow them too in my next contributions.

src/NUnitFramework/framework/Constraints/ComparisonConstraint.cs
```diff
@@ -143,11 +143,43 @@ public ComparisonConstraint Percent
...
 
         #endregion
+        
+        #region Protected Methods
+        
+        /// <summary>
+        /// Provides standard description of what the constraint tests
+        /// based on comparison text.
+        /// </summary>
+        /// <param name="comparisonText">Describes, what constraint tests, throws <see cref="ArgumentNullException"/>
```
jnm2: Stray comma in here I forgot to comment on earlier. How about: `Describes the comparison being tested,`?

madetara: That sounds smoother to me. English isn't my native, so any language advices are appreciated

Issue #3042, opened 2018-10-04
RequiresThreadAttribute allows ApartmentState.Unknown, unlike ApartmentAttribute


jnm2:
Does it make sense to allow `[RequiresThread(ApartmentState.Unknown)]`?

ChrisMaddock:
I guess it makes sense if https://github.com/nunit/nunit/issues/3041 is differentiating between MTA and 'UnInitialized' - and in which case is should be 'Uninitialized' instead?

These are hierachical, right? Presumably I would want this if I wanted to make an entire assembly as STA, and then certain fixtures as 'not-STA'?

Is there a cost to permitting it? Other than the face I don't like the "Unknown" name? 😄

CharliePoole:
Note that in older versions of .NET ApartmentState.Unknown meant "not yet set." If you did nothing, the apartment used was MTA.

jnm2:
I left out part of the question, sorry. Given that `[RequiresThread]` is always allowed, should `[RequiresThread(ApartmentState.Unknown)]` be allowed? If so, should we allow `[Apartment(ApartmentState.Unknown)]`?

We can't change Unknown to Uninitialized unless we declare our own enum. :D

@CharliePoole I can't think of any use cases for running tests that care about the apartment state and need it to be uninitialized. The only reason it's worth differentiating between leaving COM uninitialized and setting it to MTA is just so we can error in one case and not the other on platforms other than Windows.

ChrisMaddock:
I understand, sorry. Yes - I agree these two attributes should be consistent.

This is a good issue for first-time contributors! To anyone interested:

- Take a look at the existing check in ApartmentAttribute: 
https://github.com/nunit/nunit/blob/5163de63d94174f2857ac197248b1306164ee5d8/src/NUnitFramework/framework/Attributes/ApartmentAttribute.cs#L44

To close this issue, we should introduce a similar check in RequiresThreadAttribute. And of course, a unit test to support the behaviour!

petmat:
I'll take a look at this if it's ok!

jnm2:
Great, let us know if you need anything!

mikkelbu:
Closed by #3058

oznetmaster:
Not sure why this was necessary at all?  In fact, I think it is incorrect to change it.

"Note: This attribute, used with or without an ApartmentState argument will always result in creation of a new thread. To create a thread only if the current ApartmentState is not appropriate, use the ApartmentAttribute."

[...]

PR #3058, opened 2018-10-09
Add guard to not allow ApartmentState.Unknown for RequiresThreadAttribute

petmat:
This is a fix for issue #3042 

RequiresThreadAttribute did not use to throw if used like this `[RequiresThread(ApartmentState.Unknown)]` which was inconsistent with ApartmentAttribute.

With this fix both attributes throw if ApartmentState.Unknown is used.

I also included a unit test to make sure RequiresThreadAttribute works as expected.


CharliePoole:
This looks good to me, provided that you have verified the exception is properly caught and handled by NUnit. Test should be reported as not runnable.

petmat:
I had a bug in the test I created. It was not inside the `#if APARTMENT_STATE ... #endif` block. Now it's fixed. Is there something else still missing? I'm not entirely sure what you meant by reporting the test as not runnable?

ChrisMaddock:
Hey @petmat - thanks for making this change!

The way you've done the test is all I'd originally imagined, but I think Charlie makes a good point that we could test this in a stronger way. The way we normally do that, is create a 'mock test' in the test-data project in the solution, and then use the `TestBuilder` class to create the actual test in the unit test.

Have a look at the below test for the AuthorAttribute. I think we'd want to do something very similar here, except of course we'd be asserting that the test state is `RunState.NotRunnable` instead.

https://github.com/nunit/nunit/blob/cefc0c3a86c065daa4130aa255b1384db51b2b08/src/NUnitFramework/tests/Attributes/AuthorTests.cs#L44-L49

I _think_ this would work, but haven't tried it! Let us know if you hit any problems!

CharliePoole:
You __could__ write a unit test as @ChrisMaddock describes. We do that a lot and it does make for a good test.

I, however, was just looking for you to test manually and tell us the runner reported that test as invalid (not runnable).

Issue #2488, opened 2017-10-12
Improve performance of discovery/execution


rprouse:
@OsirisTerje commented on [Thu Sep 28 2017](https://github.com/nunit/nunit3-vs-adapter/issues/392)

It has been found that discovery of test cases are slowed down by the processing of traits.

rprouse:
This is a copy of an issue in the adapter. We've found that some of our internal XML processing code is causing performance problems in larger test suites. Some fixes have been made in the adapter, but improvements can be made in the framework too. This needs investigation and performance testing to determine what can be done.

See https://github.com/nunit/nunit3-vs-adapter/issues/392

ChrisMaddock:
This is related to the issue below that I wanted to look at a few days ago. 🙂 At least for the 'run individual test' workflow - not overall performance.

https://github.com/nunit/nunit/issues/3#issuecomment-334953472

rprouse:
@navin22 will add some performance issues he has found to this issue.

navin22:
I see the following as the top time consuming calls in the latest release of 3.9.0-dev-04564 version

Cross App Domain calls during the events NUnit.Framework.Internal.Execution.TestFinishedEvent.Send & NUnit.Framework.Internal.Execution.TestStartedEvent.Send.

![image](https://user-images.githubusercontent.com/19592945/31708479-b628b260-b40c-11e7-9845-5576d4ffb897.png)

I have uploaded the perfview traces for more analysis [here](https://1drv.ms/u/s!Ah2_fRsQIuFtgTk_uV4j0LPMvY6B).

PR #3061, opened 2018-10-14
Improve performance of discovery/execution #2488

MatthewBeardmore:
Related to #2488 

Changes the regex that stripped out invalid XML characters to an equivalent method that will not allocate more strings unless there is an invalid character in the string. This was a performance issue for me when attempting to run a single test in a project with 20,000 tests with the VS test runner as this method took nearly 8s before (now down to < 1s with my changes).

### DIFF
--- a/src/NUnitFramework/framework/Interfaces/TNode.cs
+++ b/src/NUnitFramework/framework/Interfaces/TNode.cs
@@ -22,6 +22,7 @@
 // ***********************************************************************
 
 using System;
+using System.Text;
 using System.Text.RegularExpressions;
 using System.Xml;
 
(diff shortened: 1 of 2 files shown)


MatthewBeardmore:
Not sure why the CI failed - given that it just got stuck on the Mac builds it looks like. Any ideas?

src/NUnitFramework/framework/Interfaces/TNode.cs
```diff
@@ -318,17 +319,64 @@ private static NodeList ApplySelection(NodeList nodeList, string xpath)
...
+                }
+                else if (!(0x0 <= c && c <= 0x8) &&
+                    c != 0xB &&
+                    c != 0xC &&
+                    !(0xE <= c && c <= 0x1F) &&
+                    !(0x7F <= c && c <= 0x84) &&
+                    !(0x86 <= c && c <= 0x9F) &&
+                    !(0xD800 <= c && c <= 0xDFFF) &&
+                    c != 0xFFFE &&
+                    c != 0xFFFF)
```
mikkelbu: Would it be possible to add some comments to this part of the if-statement? Comparing with https://www.w3.org/TR/xml/#charsets I think that I get some of the cases, but every time I need to look at this code I'll need to look up what e.g. `0x85` is, or why the range `0x86 <= c && c <= 0x9F` is special

src/NUnitFramework/framework/Interfaces/TNode.cs
```diff
@@ -400,7 +448,7 @@ public bool Pass(TNode node)
             }
         }
 
-        #endregion
+#endregion
```
mikkelbu:
```suggestion
        #endregion
```
To match the `#region` start (I know that the VS tooling for this is lacking :()

mikkelbu:
Ps. I've also restarted Travis

src/NUnitFramework/framework/Interfaces/TNode.cs
MatthewBeardmore: Hopefully these comments should be enough for you - let me know if there's anything else you'd like.

src/NUnitFramework/framework/Interfaces/TNode.cs
```diff
@@ -318,15 +319,69 @@ private static NodeList ApplySelection(NodeList nodeList, string xpath)
...
+                }
+                else
+                {
+                    // We keep the builder null so that we don't allocate a string
+                    // when doing this conversion until we encounter a unicode character.
+                    // Then, we allocate the rest of the string and escape the invalid
+                    // character.
+                    if (builder == null)
+                    {
+                        builder = new StringBuilder();
```
mikkelbu: My last comment/question (sorry about not thinking about this earlier): will there be any noticeable difference if we use one of the other overloads of the constructor to minimize expanding the `StringBuilder`? (especially when `str` is a larger string). We could do something like
```suggestion
                        builder = new StringBuilder(str.Length);
```
or also avoiding the loop (hope I get the boundaries right)
```suggestion
                        builder = new StringBuilder(str.Substring(0, index)), str.Length);
```

MatthewBeardmore: I considered it, but since we don't know how large the string will be, I didn't figure there would be much of a performance benefit, as whatever capacity we passed in would have to immediately be changed anyway.

mikkelbu:
We need just another review from the @nunit/framework-team then this can be merged.

PR #3062, opened 2018-10-14
Timeout failures on MacOS #3014

### LINKED ISSUE #3014: Timeout failures on MacOS
After setting up the macOS build on VSTS, I am seeing the following test failures occasionally,

```
1) Failed : NUnit.Framework.Attributes.RepeatableTestsWithTimeoutAttributesTests.ShouldPassAfter2RepeatsAndTimeoutIsResetEachTime
Test exceeded Timeout value of 85ms

2) Failed : NUnit.Framework.Attributes.RepeatableTestsWithTimeoutAttributesTests.ShouldPassAfter3RetriesAndTimeoutIsResetEachTime
Test exceeded Timeout value of 85ms
```

https://dev.azure.com/nunit/NUnit/_build/results?buildId=731&view=logs

MatthewBeardmore:
Closed #3014 

Converts the timeout tests from using sleeping to test the behavior via test to a test that checks how the test would have executed without running it to verify that the behavior will work.

### DIFF
--- a/src/NUnitFramework/tests/Attributes/RepeatableTestsWithTimeoutAttributesTests.cs
+++ b/src/NUnitFramework/tests/Attributes/RepeatableTestsWithTimeoutAttributesTests.cs
@@ -22,9 +22,11 @@
 // ***********************************************************************
 
 using System;
+using System.Reflection;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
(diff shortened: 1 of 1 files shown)


rprouse:
The helper tests would be better in TestData, but that would require updates beyond this, so I'm okay with leaving them here. The MacOS timeouts have been causing a ton of problems for us, so I'd really like to get this in. Thanks for tackling it.

Issue #3063, opened 2018-10-15
Test filters passed into NUnit programmatically are not escaped and cause invalid XML characters to never match


MatthewBeardmore:
When NUnit VS adapter passes a test filter generated by VS into NUnit, the filter contains an invalid XML character (but valid UTF16 character) - for example, '\x01'. When read into TestFilter.LoadXML, it throws an exception and the test can never be run. Even if the XML character is escaped in the VS test adapter, it would still fail in FullNameFilter, as that expects the unescaped UTF16 character at that time.

jnm2:
> When NUnit VS adapter passes a test filter generated by VS into NUnit, the filter contains an invalid XML character (but valid UTF16 character) - for example, '\x01'.

This should be fixed in the adapter. No good can come of constructing and passing invalid XML between systems, and we don't want to hide the symptoms of a more fundamental problem. my preference is for the framework to fail fast and loudly until the adapter sends it well-formed XML.

jnm2:
More info:

The root of the problem starts in the framework when test names contain control characters. The framework is failing to remove the control characters from the test name values it encodes in XML. It must remove them because XML completely bans them and does not even permit them to be escaped using XML entities.

In practice, if the binary data is important, folks use base64 for this. Since the name is just for display purposes, I suggest we strip out the control characters by replacing them with our standard C-style display format: `\0`, \u0001`, etc.

jnm2:
I'm slowly catching up to where you all are!
I believe this problem should be solved in three places:

1. The framework should prevent test names from containing control characters by replacing them with C-style escaping for display purposes. There should be no such thing as `ITest.FullName` containing control characters.
   ➡ All test names can be meaningfully seen and entered, and both XML output and filtering automatically work for free. Each runner doesn't have to reinvent ways to allow users to see and enter control characters in test names.
   ➡ Tests with parameters `"\u0001"` (control char) and `"\\u0001"` (literal text `\u0001`) would both be given the same name, literally the text `Foo("\u0001")`. For default name generation, I think this is ideal.

2. No need for filters to ever represent control characters, because no test names have control characters. The engine's filter builder should throw an exception if you pass it values that contain control characters. (It's currently also buggy because it doesn't handle UTF16 surrogate pairs. Both issues at https://github.com/nunit/nunit-console/issues/491).

[...]

PR #3064, opened 2018-10-15
Addresses issue 3063.

MatthewBeardmore:
Closes nunit/nunit3-vs-adapter#484 and #3063

When NUnit VS adapter passes a test filter generated by VS into NUnit, the filter contains an invalid XML character (but valid UTF16 character) - for example, '\x01'. When read into TestFilter.LoadXML, it throws an exception and the test can never be run. Even if the XML character is escaped in the VS test adapter, it would still fail in FullNameFilter, as that expects the unescaped UTF16 character at that time.

We will first need to either (enforce) escaping from any applications calling into NUnit programmatically, or do the escaping ourselves when we read in a string (in FrameworkController.RunTests) so that it always will be read in as valid XML. Secondly, we will need to address the FullNameFilter, and my two proposals are copied from the other issue below:

My proposals from the other issue are as follows:
1: Unescape the characters that we escaped in the XML to make it load, so that FullNameFilter would match correctly.
2. Escape all FullNameFilter options that come in to match what is able to be loaded from the XML.

I think 2 is more feasible, as unescaping is a more difficult task and we know that the test filters will be coming in as XML (which will be escaped if they load properly), so the escaping in the filter will make it so that they will match between the two in (I believe) all cases.

I implemented escaping all filter strings when they come in if they contain non XML compliant characters and proposal 2 for how to handle the full name filtering. The only concern I have at this time is that until the escaping is faster (from a separate pull request), this may take a hit on performance as the regex is not a fast operation.

### DIFF
--- a/src/NUnitFramework/tests/Internal/Filters/FullNameFilterTests.cs
+++ b/src/NUnitFramework/tests/Internal/Filters/FullNameFilterTests.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2015 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -24,6 +24,7 @@
 using System;
 using System.Collections.Generic;
 using System.Text;
+using NUnit.TestUtilities;
 
 namespace NUnit.Framework.Internal.Filters
 {
@@ -61,6 +62,7 @@ namespace NUnit.Framework.Internal.Filters
             Assert.False(_filter.Pass(_anotherFixture));
         }
 
+        [Test]
         public void ExplicitMatchTest()
         {
             Assert.That(_filter.IsExplicitMatch(_topLevelSuite));
@@ -68,6 +70,41 @@ namespace NUnit.Framework.Internal.Filters
             Assert.False(_filter.IsExplicitMatch(_dummyFixture.Tests[0]));
 
             Assert.False(_filter.IsExplicitMatch(_anotherFixture));
+        }
+
+        [Test]
+        public void UTF8CharacterMatchFromXML()
+        {
+            TestFilter filter = TestFilter.FromXml("<filter><test>NUnit.Framework.Internal.Filters.FullNameFilterTests.UTF8CharacterMatchFromXMLTestMethod(&quot;ABC\u0001&quot;)</test></filter>");
+            var test = TestBuilder.MakeTestFromMethod(GetType(), nameof(UTF8CharacterMatchFromXMLTestMethod));
+
+            Assert.IsTrue(test.HasChildren);
+            Assert.IsTrue(test.Tests[0].Arguments[0] is string);
+            Assert.IsTrue((string)test.Tests[0].Arguments[0] == "ABC\x01");
+            Assert.IsTrue(filter.Match(test.Tests[0]));
+        }
+
+        [Test]
+        public void ArgumentsMatch()
+        {
+            TestFilter filter = TestFilter.FromXml("<filter><test>NUnit.Framework.Internal.Filters.FullNameFilterTests.ArgumentsMatchTestMethod(1)</test></filter>");
+            var test = TestBuilder.MakeTestFromMethod(GetType(), nameof(ArgumentsMatchTestMethod));
+
+            Assert.IsTrue(test.HasChildren);
+            Assert.IsTrue(test.Tests[0].Arguments[0] is int);
+            Assert.IsTrue((int)test.Tests[0].Arguments[0] == 1);
+            Assert.IsTrue(filter.Match(test.Tests[0]));
+        }
+
+        [TestCase("ABC\x01", ExpectedResult = "ABC%01")]
+        public string UTF8CharacterMatchFromXMLTestMethod(string s)
+        {
+            return string.Empty;
+        }
+
+        [Test]
+        public void ArgumentsMatchTestMethod([Values(1, 2)] int s)
+        {
         }
     }
 }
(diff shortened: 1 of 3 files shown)


ChrisMaddock:
@MatthewBeardmore thank you for all these fixes! Apologies for just labelling them up at the moment, will get back to have a proper look through at some point soon!

MatthewBeardmore:
No rush - I'm just having fun working on them!

src/NUnitFramework/tests/Internal/Filters/FullNameFilterTests.cs
```diff
@@ -61,13 +62,51 @@ public void PassTest()
...
         }
 
+        [Test]
         public void ExplicitMatchTest()
         {
             Assert.That(_filter.IsExplicitMatch(_topLevelSuite));
             Assert.That(_filter.IsExplicitMatch(_dummyFixture));
-            Assert.False(_filter.IsExplicitMatch(_dummyFixture.Tests[0]));
+            if (((FullNameFilter)_filter).IsRegex)
+                Assert.That(_filter.IsExplicitMatch(_dummyFixture.Tests[0]));
```
MatthewBeardmore: I had to change this to work, but even after reverting my changes, I don't see how this test ever worked with the regex test case. Any thoughts?

ChrisMaddock:
Thanks for the detailed explanation @MatthewBeardmore, I think I've got my head around the issue!

I agree with the bugs you've identified, but afraid I'm not completely sure I agree with the fixes. Whilst the filters passed into the framework are just of type `string`, they are defined to be xml strings - which makes me think that the framework should expect them to be valid xml strings. If we start escaping the filter strings in the framework, then we might end up with doubly-escaped strings, for applications which are interacting with the framework 'correctly'.

As you say, this would "(enforce) escaping from any applications calling into NUnit programmatically", however - the recommended approach for the general case, is for runners not to call directly into the framework, and instead use the NUnit Engine to run tests, which provides a 'friendly' interface to the framework. In this case, I think the NUnit Engine should be taking care of escaping the runners test filter to valid xml - in fact, it already has a method to do so as below. (However, note that this method is: A) Lacking the particular invalid characters in this case, and B) may not actually be used by the adapter right now - I'm not sure!)

https://github.com/nunit/nunit-console/blob/af89a437153955fb3aea0ea168965f596a6f3066/src/NUnitEngine/nunit.engine/Services/TestFilterBuilder.cs#L82-L90

Now! Going back to the 'other half' - I think you're right, that the FullNameFilter should be unescaping the xml. However, I'm not sure the `Match()` method is the right place for that - I would think that the FillNameFilter object should be created with the correctly escaped value in the first place. (Also means we just escape the value once, rather than once per test.) To my mind, that should instead be done when the FullNameFilter is created, in the `TestFilter.FromXml()` method. (You'll see from the code below that there's a number of other filters which should probably have the same fix.)

https://github.com/nunit/nunit/blob/be72472d073e39e820898c95040ad37eedec2fba/src/NUnitFramework/framework/Internal/TestFilter.cs#L149-L199

What do you think - is there anything I've missed about the problem, which might not work here? It's also probably worth waiting for someone else on @nunit/framework-team to take a look at this...they may disagree!

jnm2:
If the framework is ever given invalid XML, my preference is that it should fail fast and loudly. Glossing over invalid XML only hides the symptoms of a more fundamental misunderstanding.

jnm2:
Here's the real culprit:

https://github.com/nunit/nunit-console/blob/af89a437153955fb3aea0ea168965f596a6f3066/src/NUnitEngine/nunit.engine/Services/TestFilterBuilder.cs#L82-L90

NUnit.Engine has an invalid XmlEscape implementation. There's also a low-hanging performance win to grab here too if we use XmlWriter instead of string.Replace five (should be at least thirty-seven) times.

jnm2:
I forgot something important: XML bans control characters and won't even allow you to escape them. I updated https://github.com/nunit/nunit-console/issues/491.

This requires the framework to remove control characters from test names before putting them in test XML or test result XML. It should probably replace them with C-style escaping such as `\0` since the test names are for display purposes.

src/NUnitFramework/tests/Internal/Filters/FullNameFilterTests.cs
```diff
@@ -61,13 +62,51 @@ public void PassTest()
...
+            var test = TestBuilder.MakeTestFromMethod(GetType(), nameof(ArgumentsMatchTestMethod));
+
+            Assert.IsTrue(test.HasChildren);
+            Assert.IsTrue(test.Tests[0].Arguments[0] is int);
+            Assert.IsTrue((int)test.Tests[0].Arguments[0] == 1);
+            Assert.IsTrue(filter.Match(test.Tests[0]));
+        }
+
+        [TestCase("ABC\x01")]
+        public void UTF8CharacterMatchFromXMLTestMethod(string s)
```
jnm2: All .NET strings are UTF16 and it doesn't make sense to talk about UTF8 characters within them. `"\x01"` is a control character.

jnm2:
I'm sorry it's taking me so long to catch up to where you all are. I think we should take a different approach than allowing invalid XML to be passed around. Could you please see https://github.com/nunit/nunit/issues/3063#issuecomment-432788447?

MatthewBeardmore:
Taking a different approach is fine by me - I'm just used to forcing "compatibility at any cost" in my work, which is why I figured the framework would be the place for a fix. If there's a better place to put this (most likely multiple places), then that's just fine by me. I like @jnm2's comment on #3036 on where to fix this, I think that makes the most sense across all of the products.

rprouse:
This is dormant, closing

PR #3071, opened 2018-10-19
Flakey tests in FrameworkControllerTests #2963

MatthewBeardmore:
Moves setting the test prefix id into each test that requires testing that functionality, and adds a try/finally block to reset it to the previous value.

Fixes https://github.com/nunit/nunit/issues/2963

### DIFF
--- a/src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
+++ b/src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
@@ -43,6 +43,7 @@ namespace NUnit.Framework.Api
         private const string MISSING_NAME = "junk";
         private const string EMPTY_FILTER = "<filter/>";
         private const string FIXTURE_CAT_FILTER = "<filter><cat>FixtureCategory</cat></filter>";
+        private const string TEST_PREFIX = "ID";
 
         private static readonly string MOCK_ASSEMBLY_NAME = typeof(MockAssembly).GetTypeInfo().Assembly.FullName;
         private static readonly string EXPECTED_NAME = MOCK_ASSEMBLY_FILE;
@@ -79,8 +80,14 @@ namespace NUnit.Framework.Api
         [SetUp]
         public void CreateController()
         {
-            _controller = new FrameworkController(MOCK_ASSEMBLY_PATH, "ID", _settings);
+            _controller = new FrameworkController(MOCK_ASSEMBLY_PATH, Test.IdPrefix, _settings);
             _handler = new CallbackEventHandler();
+        }
+
+        [OneTimeTearDown]
+        public void TearDown()
+        {
+            Test.IdPrefix = "";
         }
 
         #region SettingsElement Tests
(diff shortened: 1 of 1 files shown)


CharliePoole:
The finally blocks set the prefix to the empty string rather than resetting it to the previous value.

ChrisMaddock:
This fixes https://github.com/nunit/nunit/issues/2963. @MatthewBeardmore - looks good to me, just needs a second review for another framework team member. 🙂

src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
```diff
@@ -246,38 +262,55 @@ public void ConstructController()
...
-            Assert.That(result.Attributes["name"], Is.EqualTo(EXPECTED_NAME));
-            Assert.That(result.Attributes["runstate"], Is.EqualTo("Runnable"));
-            Assert.That(result.Attributes["testcasecount"], Is.EqualTo(MockAssembly.Tests.ToString()));
-            Assert.That(result.SelectNodes("test-suite").Count, Is.EqualTo(0), "Load result should not have child tests");
+            var originalPrefix = Test.IdPrefix;
+            var controller = new FrameworkController(MOCK_ASSEMBLY_PATH, TEST_PREFIX, _settings);
+            try
+            {
+                new FrameworkController.LoadTestsAction(controller, _handler);
+                var result = TNode.FromXml(_handler.GetCallbackResult());
```
jnm2: As little as possible should be in the `try` block. It matches the exact relevance of the `finally` block to the contents of the `try` block, which is good for clarity. Could you end the `try` right after `_handler.GetCallbackResult()`?

src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
```diff
@@ -76,13 +80,25 @@ public void TestOutOfFixtureCategory()
...
         {
-            _controller = new FrameworkController(MOCK_ASSEMBLY_PATH, "ID", _settings);
+            _controller = new FrameworkController(MOCK_ASSEMBLY_PATH, Test.IdPrefix, _settings);
             _handler = new CallbackEventHandler();
         }
 
+        [OneTimeTearDown]
+        public void TearDown()
+        {
+            Test.IdPrefix = _originalPrefix;
```
jnm2: Is there any scenario in which `Test.IdPrefix` has been changed, the finally blocks do not execute, and `TearDown` can run?

MatthewBeardmore: I don't think so, but I'd rather be safe than sorry, which is why I added it. It should be able to be removed safely though.

src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
MatthewBeardmore: I wasn't sure where the cut off was, so I put it across the entire method, as it's more clear to someone looking at the method what it was doing (at least, in my opinion).

jnm2: All those Assert statements are running with the wrong `Test.IdPrefix`. Could you move at least those out of the `try`?

src/NUnitFramework/tests/Api/FrameworkControllerTests.cs
jnm2: Its presence here implies that it is doing something. Would you be willing to remove it?

jnm2:
@MatthewBeardmore Can we do anything to help with this one?

rprouse:

PR #3081, opened 2018-10-30
Byte ranges no longer fail when they end on a boundary

### LINKED ISSUE #3079: Regression from 3.10 to 3.11: Range in bytes
Hello,

I have tests which define a Range in bytes. After the update from NUnit 3.10 to 3.11 those tests were not executable any more. I changed the data type form byte to int and the tests were executable again.

If the Range was defined in bytes, I got the attached error.
Example: public void test([Range(0x00, 0x0F)] byte value)
[git3.11.log](https://github.com/nunit/nunit/files/2525229/git3.11.log)

Best regards,
Martin

jnm2:
Fixes #3079.

### DIFF
--- a/src/NUnitFramework/framework/Internal/ValueGenerator.ByteValueGenerator.cs
+++ b/src/NUnitFramework/framework/Internal/ValueGenerator.ByteValueGenerator.cs
@@ -31,7 +31,7 @@ namespace NUnit.Framework.Internal
             {
                 if (value is byte)
                 {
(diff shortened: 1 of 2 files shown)


mikkelbu:
@jnm2 A test is failing on Travis and AppVeyor, but strangely Azure Pipelines are "succeeding **with issues**", and the issue is that the test is also failing there. @rprouse We probably need to make Azure Pipelines fail when one of the steps fail.

jnm2:
Oops, missed one. There were several tests that would have required their own PRs because it would require expanding functionality beyond what we have always had.

Issue #3059, opened 2018-10-11
Constraint Throws.Exception does not work with async return value


origameg:
When using the `Throws.Exception` constraint to test an async delegate with a return value, the test incorrectly fails even when an Exception should have been thrown:
```csharp
[Test]
public void ThrowsExceptionConstraint_AsyncReturnValue_Fails()
{
    Assert.That(async () => { await Task.Delay(200); throw new Exception(); return true; },
        Throws.Exception);
}
```
```
Failed   ThrowsExceptionConstraint_AsyncReturnValue_Fails
Error Message:
   Expected: an exception to be thrown
  But was:  no exception thrown
```

The constraint appears to work fine for async delegates that return `Task` instead of `Task<T>` or if additional constraints are supplied.
```csharp
[Test]
public void ThrowsExceptionConstraint_AsyncTask_Passes()
{
    Assert.That(async () => { await Task.Delay(200); throw new Exception(); },
        Throws.Exception);
    // Passes
}

[Test]
public void ThrowsExceptionConstraint_AsyncReturnValue_AdditionalConstraint_Passes()
{
    Assert.That(async () => { await Task.Delay(200); throw new Exception("foo"); },
        Throws.Exception.Message.Contains("foo"));
    // Passes
}
```
I'm aware that it is not recommended to use the `Throws.Exception` constraint on its own. However, it's not completely uncommon, and the false test failure is pretty misleading.

_I've reproduced this on:_
- NUnit versions: _3.11.0, 3.10.1_
- Test runners: _nunit3-console v3.9.0, ReSharper 2018.1.2_
- .NET versions: _Core 2.0, Core 2.1, Framework 4.6.1_

mikkelbu:
We should probably generalise the logic (see https://github.com/nunit/nunit/pull/2547/files#diff-528b529b1298bbb4bd387f02090a3de3R105) from #2547 to also handle `Task<T>`.

PR #3087, opened 2018-11-08
3059 - Fix assert throws for async Task<>

stevenaw:
Update `ThrowsExceptionConstraint` to detect if anonymous `async Task<>` is passed.

Fixes https://github.com/nunit/nunit/issues/3059

### DIFF
--- a/src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
@@ -98,7 +98,12 @@ namespace NUnit.Framework.Constraints
         public override ConstraintResult ApplyTo<TActual>(ActualValueDelegate<TActual> del)
         {
 #if ASYNC
-            if (typeof(TActual) == typeof(Task)) return ApplyTo(new AsyncTestDelegate(() => (Task)(object)del.Invoke()));
+            if (typeof(TActual) == typeof(Task))
+                return ApplyTo(new AsyncTestDelegate(() => (Task)(object)del.Invoke()));
+            
+            var typeWrapper = new TypeWrapper(typeof(TActual));
+            if (typeWrapper.IsGenericType && typeWrapper.GetGenericTypeDefinition() == typeof(Task<>))
+                return ApplyTo(new AsyncTestDelegate(() => (Task)(object)del.Invoke()));
 #endif
             return ApplyTo(new TestDelegate(() => del.Invoke()));
         }
--- a/src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
@@ -26,6 +26,7 @@ using NUnit.TestUtilities;
 
 #if ASYNC
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 #endif
 
@@ -65,19 +66,39 @@ namespace NUnit.Framework.Constraints
         };
 
 #if ASYNC
+        private static async Task YieldAsync()
+        {
+#if NET40
+            await TaskEx.Yield();
+#else
+            await Task.Yield();
+#endif
+        }
+
+
         [Test]
         public static void CatchesAsyncException()
         {
             Assert.That(async () =>
             {
-#if NET40
-                await TaskEx.Yield();
-#else
-                await Task.Yield();
-#endif
+                await YieldAsync();
                 throw new Exception();
             }, Throws.Exception);
         }
+        
+        [Test]
+        public static void CatchesAsyncOfTException()
+        {
+            Assert.That(async () =>
+            {
+#pragma warning disable CS0162
+                await YieldAsync();
+                throw new Exception();
+                return 2;
+#pragma warning restore CS0162
+            }, Throws.Exception);
+        }
+
 #endif
     }
 }
--- a/src/NUnitFramework/tests/TestUtilities/AsyncTestDelegates.cs
+++ b/src/NUnitFramework/tests/TestUtilities/AsyncTestDelegates.cs
@@ -1,4 +1,4 @@
-﻿#if ASYNC
+#if ASYNC
 
 using System;
 using System.Threading.Tasks;


src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
```diff
@@ -98,7 +98,12 @@ public override ConstraintResult ApplyTo<TActual>(TActual actual)
         public override ConstraintResult ApplyTo<TActual>(ActualValueDelegate<TActual> del)
         {
 #if ASYNC
-            if (typeof(TActual) == typeof(Task)) return ApplyTo(new AsyncTestDelegate(() => (Task)(object)del.Invoke()));
+            if (typeof(TActual) == typeof(Task))
+                return ApplyTo(new AsyncTestDelegate(() => (Task)(object)del.Invoke()));
+            
+            var typeWrapper = new TypeWrapper(typeof(TActual));
```
stevenaw: I'm unsure if this is the best way to do this, but I see this pattern in other places in the codebase

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
```diff
@@ -65,19 +66,39 @@ public void SucceedsWithNonVoidReturningFunction()
         };
 
 #if ASYNC
+        private static async Task YieldAsync()
```
stevenaw: I DRYed this out within this fixture, based on what I saw here: https://github.com/nunit/nunit/blob/cefc0c3a86c065daa4130aa255b1384db51b2b08/src/NUnitFramework/tests/Internal/TestExecutionContextTests.cs#L1002

Though this means we have 2 of these functions within the test suite. Is there a place I should move this?

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
```diff
@@ -65,19 +66,39 @@ public void SucceedsWithNonVoidReturningFunction()
...
         }
+        
+        [Test]
+        public static void CatchesAsyncOfTException()
+        {
+            Assert.That(async () =>
+            {
+#pragma warning disable CS0162
+                await YieldAsync();
+                throw new Exception();
```
mikkelbu: Personally, I would prefer something like `if (DateTime.MinValue < DateTime.Now) throw new Exception();` or equivalent and the remove the `pragma warning`, but that is just my personal opinion (and actually I don't know how we do it in the rest of the codebase).

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
mikkelbu: We could place it in the folder `nunit/src/NUnitFramework/tests/TestUtilities/`, but I don't think that there is a file that makes sense (`AsyncTestDelegates.cs` is probably the most relevant). And I think it would be overkill to create a new file just for this simple method. So I don't mind the minor duplication for now.

src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
```diff
@@ -98,7 +98,12 @@ public override ConstraintResult ApplyTo<TActual>(TActual actual)
         public override ConstraintResult ApplyTo<TActual>(ActualValueDelegate<TActual> del)
         {
 #if ASYNC
-            if (typeof(TActual) == typeof(Task)) return ApplyTo(new AsyncTestDelegate(() => (Task)(object)del.Invoke()));
+            if (typeof(TActual) == typeof(Task))
```
jnm2: This (pre-existing) check is actually brittle. Folks are free to derive from Task and we would fail to recognize that.

I think this should be `if (typeof(Task).IsAssignableFrom(del.GetMethod("Invoke").ReturnType)) { ...`.
By doing this, the following code should work:

```c#
object obj = new Func<Task>(async () => await ...);
Assert.That(obj, Throws...);

// Task<T> derives from Task, so this is handled automatically.
obj = new Func<Task<T>>(async () => await ...);
Assert.That(obj, Throws...);
```

src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
jnm2: ℹ `TypeWrapper` shouldn't be used except as an exchange type when using the framework interfaces to communicate with non-framework code.

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
jnm2: What do you think of the name `YieldsAsync`?

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
```diff
@@ -65,19 +66,39 @@ public void SucceedsWithNonVoidReturningFunction()
...
-#else
-                await Task.Yield();
-#endif
+                await YieldAsync();
                 throw new Exception();
             }, Throws.Exception);
         }
+        
+        [Test]
+        public static void CatchesAsyncOfTException()
```
jnm2: `Async` here refers to the way in which the exception is thrown (asynchronously, not synchronously), so we wouldn't treat it as a type and say `AsyncOfT`. Maybe `CatchesAsynchronousTaskOfTException`, since `CatchesSynchronousTaskOfTException` is also a test we should have at some point?

jnm2: To contrast synchronously and asynchronously thrown exceptions:

```c#
async Task ThrowsExceptionAsynchronously() => throw new Exception();
var task = ThrowsExceptionAsynchronously(); // No exception thrown on this line
await task; // Exception thrown on this line

Task AlsoThrowsExceptionAsynchronously() => Task.FromException(new Exception);
var task = AlsoThrowsExceptionAsynchronously(); // No exception thrown on this line
await task; // Exception thrown on this line

Task ThrowsExceptionSynchronously() => throw new Exception();
var task = ThrowsExceptionSynchronously(); // Exception thrown on this line
```

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
jnm2: Why is the `return 2` there though? It's unreachable.

src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
jnm2: Ah, missed that `del` is `ActualValueDelegate<TActual>` and not `TActual`. The above is code that would go in the `ApplyTo<TActual>(TActual actual)` overload.

For this overload, the check would simplify to `typeof(Task).IsAssignableFrom(typeof(Actual))` since `del.GetMethod("Invoke").ReturnType` will always be `typeof(Actual)` in this overload.

src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
stevenaw: Ah, thanks for the explanation.

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
stevenaw: That was more just to make the compiler happy, since the specific test case was for a `Task<T>` which also throws. I've restructured it a bit now to not require the pragmas.

Is there an established way we do this in nunit?

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
stevenaw: Thanks for the suggestions. I've renamed it to `CatchesAsyncTaskOfTException` so it's consistent naming with the existing `CatchesAsyncException` function

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
jnm2: Now, now I see why the `return 2;` was there, so that the generic type argument would be inferred to be `Task<int>` instead of `Task`. I think the idiomatic C# would be to stop relying on inference and specify the type argument explicitly:

```cs
Assert.That<Task<int>>(async () =>
{
    await YieldAsync();
    throw new Exception();
}, Throws.Exception);
```

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
```diff
@@ -65,19 +66,39 @@ public void SucceedsWithNonVoidReturningFunction()
...
+            Assert.That(async () =>
+            {
+#pragma warning disable CS0162
+                await YieldAsync();
+                throw new Exception();
+                return 2;
+#pragma warning restore CS0162
+            }, Throws.Exception);
+        }
+
```
jnm2: (While you're changing stuff, there wasn't a blank line here before.)

src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
jnm2: Here is an example of why we should always use `typeof(Task).IsAssignableFrom`: `https://github.com/nunit/nunit/issues/3093#issuecomment-441184381`

@stevenaw If you do `if (typeof(Task).IsAssignableFrom(typeof(TActual)))`, a single `if` will take care of `Task`, Task<T>`, and `MyCustomTask : Task<T> { }`.

stevenaw: Thanks @jnm2 . I think you might be looking at an old commit, I did that update here: https://github.com/nunit/nunit/pull/3087/commits/926a3b5e4423c76c0ee122cf035887af0c9984f1#diff-528b529b1298bbb4bd387f02090a3de3R101

:)

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
stevenaw: Removed 👍

src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
jnm2: Oh, perfect! Sorry about that! 😃

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
```diff
@@ -45,6 +46,7 @@ public void SetUp()
         [Test]
         public void SucceedsWithNonVoidReturningFunction()
         {
+
```
jnm2: I'm sorry I didn't notice this earlier. I don't think we should add this blank line.

jnm2:
Awesome! Other than one nit, this is ready to go though as soon as @mikkelbu re-reviews! 🎉

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
stevenaw: Oops, no, that's my fault. I accidentally added that just now. Thanks for catching, removing now.

mikkelbu:
I'll take a look at this tonight (3-4 hours), but I doubt that I can find anything that has escaped Joseph.

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
```diff
@@ -68,13 +69,30 @@ public void SucceedsWithNonVoidReturningFunction()
...
-            Assert.That(async () =>
+            Assert.That(async () => await AsyncTestDelegates.ThrowsArgumentExceptionAsync(), Throws.Exception);
+        }
+        
+        [Test]
+        public static void CatchesAsyncTaskOfTException()
+        {
+            Assert.That<Task<int>>(async () =>
+            {
+                await AsyncTestDelegates.Delay(500);
```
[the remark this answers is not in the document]
stevenaw: Sorry, I hadn't considered the impact on the test time. Fixed!

src/NUnitFramework/tests/Constraints/ThrowsExceptionConstraintTests.cs
```diff
@@ -68,13 +69,30 @@ public void SucceedsWithNonVoidReturningFunction()
...
+        [Test]
+        public static void CatchesSyncException()
+        {
+            Assert.That(() => AsyncTestDelegates.ThrowsArgumentException(), Throws.Exception);
+        }
+
+        [Test]
+        public static void CatchesSyncTaskOfTException()
+        {
+            Assert.That<int>(() =>
```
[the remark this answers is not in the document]
stevenaw: Ah, you are right. Thanks for spotting @jnm2

PR #3091, opened 2018-11-19
End support for .NET Framework 2.0

### LINKED ISSUE #3070: End support for .NET Framework 2.0 (released in 2005)
Supporting a minimum of net20 instead of net35 increases the complexity of our development. We have NUnit.System.Linq and define our own System.Action and write `NET35 || NET20` where we could have `NET35`. We take extra time waiting for tests. And we have more work to do on just net20: https://github.com/nunit/NUnit.System.Linq/issues/12

To quote @NN--- from https://github.com/nunit/NUnit.System.Linq/issues/12#issuecomment-430979252:

> If I have library for .NET 2.0 , then tests should be .NET 2.0 .
> I don't think there is anyone who still uses 2.0.
> The only issue is that many libraries have support for all .NET versions starting from 2.0.

Maybe if NUnit stops supporting net20, it could give other libraries the nudge they need to drop net20 as well. If a net20 project is still in development, it should upgrade to a newer .NET Framework and fix any bugs. I expect bugs in practice to be extremely rare. If a net20 project is not still in development, it shouldn't have a reason to upgrade to newer NUnit framework or runners.

I'm still in favor of supporting net35 (released in 2008) because I know of real-world projects that run using CLR v2 (supports up to net35) and I wouldn't be confident running tests for that on the CLR v4 engine. Also, VSTest is still shipping with a net35 runner. However, I do wonder if dropping this support could have a beneficial ripple effect.

Lastly, the .NET Framework 2.0 product is no longer supported by its own manufacturer:

> Support for .NET Framework 2.0 ended on July 12, 2011. .NET 3.5 SP1 is the only supported service pack level after this date. We strongly encourage customers to migrate to .NET Framework 3.5 SP1. For more information, please visit the .NET Framework Support Lifecycle Policy FAQ 

[https://support.microsoft.com/lifecycle/search?alpha=.NET Framework 2.0](https://support.microsoft.com/lifecycle/search?alpha=.NET%20Framework%202.0)

jnm2:
Closes #3070.


rprouse:
Well, this has been open 13 days now and it has been quiet here and in the issue. Next stop the engine so we can consolidate our usage of Mono.Cecil? 😸

jnm2:
Once the engine drops net20, we can also archive the NUnit.System.Linq repo.

mikkelbu:
> Once the engine drops net20, we can also archive the NUnit.System.Linq repo.

@jnm2 Actually, as far as I can tell, we don't use NUnit.System.Linq in the engine, or am I missing something?

jnm2:
@mikkelbu Oh, it's probably NUnitLite I'm thinking of.

Issue #3093, opened 2018-11-22
Tests having TaskLike objects as their return type throws Exception


miyaniv:
After upgrading NUnit.3.10.1 to NUnit.3.11.0 tests which returned TaskLike objects have failed.

Message: System.InvalidOperationException: The delegate did not return a Task.

StackTrace:
AsyncToSyncAdapter.Await(Func`1 invoke)
TestMethodCommand.RunTestMethod(TestExecutionContext context)

The tests have passed before the upgrade without fail.

I looked at the code and it throws this exception if returnType FullName is Task's type name.

A TaskLike object is defined by a certain attribute which force the Task interface.
The attribute:
System.Runtime.CompilerServices.AsyncMethodBuilderAttribute

The comment in the code applies this feature is soon to come however I couldn't find an open issue about it.
I would like to know when to expect a resolution.

This issue is important because many of our tests fail due to this behaviour.

mikkelbu:
Hi @miyaniv 
The issue to watch is this one - I guess - nunit/nunit#2286, but I don't know the status of it. Probably @jnm2 will have more information.

jnm2:
Oh, yes, I would like to get this done.

To be technical about it, we ignore the `AsyncMethodBuilderAttribute` attribute (except that we use it to error on async void methods). Is we depended on this attribute, we wouldn't recognize this as a tasklike method:

```cs
[Test]
public Task Test1() => AssertFoo(() => 42);

private async Task AssertFoo(Func<int> foo)
{
    // ...
}
```

Same for ValueTask.

Instead, what we'll do is recognize the C# awaitable pattern as defined in the C# 5.0 spec and unaltered as of C# 7.3: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#awaitable-expressions

> I would like to know when to expect a resolution.
>
> This issue is important because many of our tests fail due to this behaviour.

We're folks like yourself who contribute in our spare time. I can't promise, but I can say that I have prototypes and it makes sense to finish them soon now that we've gotten our second request for this. Thanks for letting us know that this one matters to you!

miyaniv:
Thank you for your quick response

Let me give you a little brief about the Task which may explain why our Tests were fine before and why the workaround is a bit complex for us.

I wrote a Task called ContextTask.
A Regular Task copies its context with **shallow copy** which means not all context is being past to the new Task. 
ContextTask **deep copies** that context.
ContextTask is basically a Task for all that matters. 
[...]

Issue #3023, opened 2018-09-18
NUnit runner fails when test method returns ValueTask<>


toolmagic:
**Versions**: 

Nuget packages:
NUnit 3.10.1
NUnit3TestAdapter 3.10.0
System.Threading.Tasks.Extensions 4.5.1

Environment:
Windows 10.1803 17134.286 
Visual Studio 2017 15.8.4

**Steps to Reproduce**:

1. Create C# Class Library, .NET 4.7.2
2. Add Nuget references for NUnit, NUnit3TestAdapter and System.Threading.Tasks.Extensions
3. Add Test Class:
``` csharp
using System.Threading.Tasks;
using NUnit.Framework;

namespace ClassLibrary1
{
    public class Class1
    {
        [Test(/*ExpectedResult = 100*/)]
        public async ValueTask<int> AsyncGetValueTest()
        {
            return await Task.FromResult(100);
        }
    }
}
```
4. Run test "AsyncGetValueTest"

**Expected Behavior**:
NUnit says "Missed ExpectedResult in Test attribute" or "Test methods with return type *ValueTask<>* are not supported"

**Actual Behavior**:
NUnit fails with an internal error:
System.NullReferenceException : Object reference not set to an instance of an object.
   at NUnit.Framework.Internal.AsyncInvocationRegion.AsyncTaskInvocationRegion.WaitForPendingOperationsToComplete(Object invocationResult)
  at C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\AsyncInvocationRegion.cs:line 109
  at NUnit.Framework.Internal.Commands.TestMethodCommand.RunAsyncTestMethod(TestExecutionContext context)
 at C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Commands\TestMethodCommand.cs:line 96

toolmagic:
@mikkelbu , exactly. It's better to say "a feature is not supported" until the feature is implemented.

jnm2:
Fixed in the prerelease feed and included in 3.12 which is just around the corner!

PR #3095, opened 2018-11-24
Support custom awaitables

### LINKED ISSUE #2168: ThrowsAsync reports OperationCanceledException as TaskCanceledException

### LINKED ISSUE #2286: Add support for custom Task (i.e. ValueTask)
Can you add support for custom Tasks, i.e. ValueTask (new in C# 7.0).

Right now looking at Assert.ThrowsAsync.

### LINKED ISSUE #3222: Our build script tests hang when run with Mono on Windows
When I run the equivalent of test45 using Mono 5.20 on Windows just like we do on Linux, the test run hangs consistently. It looks similar to what we've seen intermittently in CI.

Steps to reproduce from a PowerShell prompt on Windows:

- `choco install mono -y` if you don't have it already

- `.\build.ps1 -t build`

- ```
  &"C:\Program Files\Mono\bin\mono.exe" `
    tools/NUnit.ConsoleRunner.3.9.0/tools/nunit3-console.exe `
    bin/Release/net45/nunit.framework.tests.dll `
    --result=test-results/net45/nunit.framework.tests.xml `
    --process=InProcess
  ```

The test run has never completed for me.

jnm2:
Closes #2286, #3023, #3093.
Fixes #2168 and properly fixes #3222.

### New features

This enables test and setup methods (and any other methods in user code which NUnit directly invokes) to return objects other than Task which the C# language knows how to await, according to the awaitable expressions section of the C# 5 spec which has not been changed as of 7.3.

It does not matter whether the `async` keyword is used, so both of these will work:

```cs
[Test]
public async ValueTask Test1()
{
    await Task.Yield();
}

[Test]
public ValueTask Test2() => Test1();
```

Awaitable objects (consumable via the `await` keyword since C# 5) go beyond tasklike objects (produceable via the `async` keyword since C# 7). It doesn't matter whether you can use a type as an `async` method builder return type; it only matters whether you can `await` it:

```cs
[Test]
public System.Runtime.CompilerServices.YieldAwaitable Test3() => Task.Yield();
```

#### Behavioral changes

`Assert.That` now treats all delegates that return an awaitable type the same way it has treated delegates returning Task, by waiting for the result:

```cs
Assert.That(() => foo.ValueTaskReturningMethod(), Throws.InstanceOf<FooException>());
```

### No new APIs

This PR does not add new APIs. Therefore, `Assert.ThrowsAsync` still requires our `AsyncTestDelegate` type to be passed which requires the delegate to return `Task`. 

If we wanted to support this, we would have to add an `Assert.ThrowsAsync(Func<object>)` overload. This is as specific as you can be in C# while still allowing custom awaitable return types. We would have to error at runtime (and via analyzer) if the returned value was not in fact awaitable.

Folks using ValueTask have reasonably easy ways to handle this already which work in all versions of NUnit:

```c#
[Test]
public void Test4() => Assert.ThrowsAsync(() => foo.ValueTaskReturningMethod().AsTask());

[Test]
public void Test5() => Assert.ThrowsAsync(async () => await foo.ValueTaskReturningMethod());
```

As well as constraint syntax as of NUnit 3.12, as mentioned above:

```c#
[Test]
public void Test7() => Assert.That(foo.ValueTaskReturningMethod, Throws.Exception);
```

### Codebase changes

Because awaitability is pattern-based rather than type-based, I was able to delete almost all occurrences of `#if ASYNC` (there are only five places left in the framework). Async behavior is now automatically effective in all builds, including `net20` and `net35`.

The remaining occurrences have less to do with async behavior and more to do with whether the .NET 4.0 Task Parallel Library framework types are available. For example, whether we define `AsyncTestDelegate` which has a return value of `Task`.
In an [earlier discussion](https://github.com/orgs/nunit/teams/framework-team/discussions/3?from_comment=2#discussion-3-comment-2) we decided on changing `ASYNC` to `TASK_PARALLEL_LIBRARY_API`.


jnm2:
@mikkelbu and @rprouse, would you both be willing to review if you can find the time?

jnm2:
Travis is showing this failure on the four Mono test runs:

> Cancelled : NUnit.Framework.Constraints.DelayedConstraintTests.ThatBlockingDelegateWhichFailsWithoutPolling_FailsAfterDelay
Test cancelled by user
  at NUnit.Framework.Internal.Reflect.InvokeMethod

mikkelbu:
@jnm2 I can take a look at it in the beginning of next week, but I'm not strong in async + await, so I'll take it as a "learning session" 😄.

jnm2:
@mikkelbu I'm sorry, I was thinking you would be interested in this one, but now I'm not sure what made me think that. I'd covet a quick review but don't trouble yourself too much, please!

jnm2:
Other DelayedConstraintTests tests are causing continual failures on master with AzureDevops macOS builds. It's not necessarily unique to this PR.

jnm2:
Resolved merge conflicts.

jnm2:
Resolved merge conflicts and copied a commit from https://github.com/nunit/nunit/pull/3096 to fix a problem with `master`.

jnm2:
Rebased to consume a commit from https://github.com/nunit/nunit/pull/3096 to fix a problem with `master`.

src/NUnitFramework/framework/Internal/AwaitAdapter.cs
```diff
@@ -34,81 +30,47 @@ internal abstract class AwaitAdapter
...
+            // we still need to be able to await it to preserve NUnit behavior on machines
+            // which have a max .NET Framework version of 4.0 installed, such as the default
+            // for versions of Windows earlier than 8.
+            var task = awaitable as System.Threading.Tasks.Task;
+            if (task != null) return Net40BclTaskAwaitAdapter.Create(task);
+#endif
 
-            // Assumption that GetResult blocks until complete is only valid for System.Threading.Tasks.Task.
-            public override void BlockUntilCompleted() => _awaiter.GetResult();
+            throw new NotSupportedException("NUnit can only await objects which follow the C# specification for awaitable expressions.");
```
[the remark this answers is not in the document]
jnm2: No regression; we only ever supported `System.Threading.Tasks.Task`, which is a subset of all objects which can be awaited in C#.

src/NUnitFramework/framework/Internal/DefaultBlockingAwaitAdapter.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// </summary>
+    internal abstract class DefaultBlockingAwaitAdapter : AwaitAdapter
+    {
+        private volatile ManualResetEventSlim _completedEvent;
+
+        public sealed override void BlockUntilCompleted()
+        {
+            if (IsCompleted) return;
+
+            var completedEvent = _completedEvent; // Volatile read (would be Volatile.Read if not for net40 support)
```
stevenaw: Question (rather than a suggestion), but why not wrap this in a `#if NET_40` and then use `Volatile.Read()` in an `#else` block?

jnm2: Using `Volatile.Read` would be redundant so long as `_completedEvent` is declared with the `volatile` keyword, and in fact, the compiler would warn that a `volatile` field is unsafely used as a `ref` parameter. So I could do that, but I would also have to add `#if` around the `volatile` keyword and I didn't feel like the effort was worth it since the behavior is the same.

stevenaw: Oh ok, that makes sense. Thanks for the explanation!

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -375,5 +375,107 @@ private static bool IsNullable(Type type)
...
+
+        /// <summary>
+        /// Same as <c>GetMethod(<paramref name="name"/>, <see cref="BindingFlags.Public"/> |
+        /// <see cref="BindingFlags.Instance"/>, <see langword="null"/>, <paramref name="parameterTypes"/>,
+        /// <see langword="null"/>)</c> except that it also chooses only non-generic methods.
+        /// Useful for avoiding the <see cref="AmbiguousMatchException"/> you can have with <c>GetMethod</c>.
+        /// </summary>
+        internal static MethodInfo GetNonGenericPublicInstanceMethod(this Type type, string name, Type[] parameterTypes)
+        {
+            for (var currentType = type; currentType != null; currentType = currentType.GetTypeInfo().BaseType)
```
rprouse: Minor, but this could be a `foreach(var currentType in type.TypeAndBaseTypes())` reusing your code from above. it might read better?

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -375,5 +375,107 @@ private static bool IsNullable(Type type)
...
+        internal static object DynamicInvokeWithTransparentExceptions(this Delegate @delegate)
+        {
+            try
+            {
+                return @delegate.DynamicInvoke();
+            }
+            catch (TargetInvocationException ex)
+            {
+                ExceptionHelper.Rethrow(ex.InnerException);
+                throw null; // Rethrow’s return type would be `never` if C# could express that.
```
rprouse: Just checking if I understand this, the `throw null` causes the `RecordException` to return null if there is no inner exception, correct?

rprouse:
I also kicked off a new Travis build which failed again and the Azure DevOps build no longer exists. You might want to merge latest from master to get the various new build changes?

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: Oh, I like that! Much better.

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: `throw null;` is there to satisfy C#'s flow control but it will never be hit unless `ExceptionHelper.Rethrow` fails to throw an exception. (I think its actual effect is identical to `throw new NullReferenceException();`.)

Maybe better would be:

```cs
throw new InvalidOperationException("ExceptionHelper.Rethrow failed to throw an exception.");
```

jnm2:
Thanks Rob! I'll rebase and add some more direct tests when I get a chance. The rest of my day is pretty full but I will take opportunities to work on it as they come and tomorrow evening is mostly free.

src/NUnitFramework/framework/Internal/DefaultBlockingAwaitAdapter.cs
jnm2: I forgot: recent versions of the C# compiler suppress the warning for `ref` arguments to Interlocked and Volatile methods.

jnm2:
@rprouse I rebased and then added four commits.

jnm2:
@rprouse I rebased and then added three commits.

rprouse:
`ThatBlockingDelegateWhichFailsWithoutPolling_FailsAfterDelay` is failing on Linux and Mac for .NET 3.5, 4.0 and 4.5.

```
Error message
Test cancelled by user

Stack trace
  at NUnit.Framework.Internal.Reflect.InvokeMethod (System.Reflection.MethodInfo method, System.Object fixture, System.Object[] args) [0x00017] in <519db8877aff400fad56aeeaa1258048>:0 
  at NUnit.Framework.Internal.MethodWrapper.Invoke (System.Object fixture, System.Object[] args) [0x00006] in <519db8877aff400fad56aeeaa1258048>:0 
  at NUnit.Framework.Internal.Commands.TestMethodCommand.InvokeTestMethod (NUnit.Framework.Internal.TestExecutionContext context) [0x00011] in <519db8877aff400fad56aeeaa1258048>:0 
  at NUnit.Framework.Internal.Commands.TestMethodCommand.RunTestMethod (NUnit.Framework.Internal.TestExecutionContext context) [0x0003d] in <519db8877aff400fad56aeeaa1258048>:0 
  at NUnit.Framework.Internal.Commands.TestMethodCommand.Execute (NUnit.Framework.Internal.TestExecutionContext context) [0x00000] in <519db8877aff400fad56aeeaa1258048>:0 
  at NUnit.Framework.Internal.Commands.BeforeAndAfterTestCommand+<>c__DisplayClass1_0.<Execute>b__0 () [0x00016] in <519db8877aff400fad56aeeaa1258048>:0 
  at NUnit.Framework.Internal.Commands.BeforeAndAfterTestCommand.RunTestMethodInThreadAbortSafeZone (NUnit.Framework.Internal.TestExecutionContext context, System.Action action) [0x00000] in <519db8877aff400fad56aeeaa1258048>:0 
```

jnm2:
I see. Starting on it now.

jnm2:
The problem is introduced in https://github.com/nunit/nunit/pull/3095/commits/04eaebdb1402dd07df1b899f7e735c903e349c99. Windows Subsystem for Linux is super helpful.

jnm2:
Not making any headway yet, but I have another three hours and I'm not giving up.

jnm2:
The failure ("Test cancelled by user") is something I've seen intermittently in our master branch for as long as I can remember. I'm struggling to figure out how my PR is interacting with it, but the good news is that it'll force us to fix it. The bad news is that I don't know how long it's going to take to unravel this. You might have to release without my PR.

What's clear according to my testing is that [this line](https://github.com/nunit/nunit/blob/9f2d8ea7c35518ee28f1255bc21aeca6d84057b3/src/NUnitFramework/tests/Constraints/DelayedConstraintTests.cs#L261) is throwing ThreadAbortException, handled [here](https://github.com/nunit/nunit/blob/9f2d8ea7c35518ee28f1255bc21aeca6d84057b3/src/NUnitFramework/framework/Internal/Results/TestResult.cs#L593) (verified by changing the message and seeing it change in the repro).

What's not clear is why a thread abort is happening at all or why it's only happening on Mono.

Similar thing here, no answer: https://stackoverflow.com/questions/18982951/can-waithandle-waitone-throw-a-threadabortexception
https://stackoverflow.com/questions/7629986/why-am-i-getting-thread-was-being-aborted-in-asp-net seems to indicate that this happens due to AppDomain shutdown. Could it be that the NUnit runner is shutting down the AppDomain before the test has finished delaying?

jnm2:
No repro on Ubuntu 18.04.1 LTS in Hyper-V with Mono 5.18.1 [which our CI build currently gets](https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/linux/Ubuntu1604-README.md).

What could the difference be between my Ubuntu machine and CI?

jnm2:
Ooh! I spoke too soon, caught it on `test40` even though `test45` passed!

jnm2:
Not related probably, but debugging the NUnit Console master branch against this PR branch in MonoDevelop (on Mono 5.18.1) shows that `_currentWorkItem` is null on this line:

https://github.com/nunit/nunit/blob/8a6245b90276fcab21a11bf65b06bbcdd562d67b/src/NUnitFramework/framework/Internal/Execution/TestWorker.cs#L135-L137

This can't be good, but I thought we had fixed that.

jnm2:
Also, it's difficult to break on exceptions on Mono because the framework is constantly triggering System.Runtime.Remoting.RemotingException "Cannot resolve method NUnit.Engine.RunTestsCallbackHandler:RaiseCallbackEvent" on this line:

https://github.com/nunit/nunit/blob/8a6245b90276fcab21a11bf65b06bbcdd562d67b/src/NUnitFramework/framework/Internal/TestProgressReporter.cs#L94

Both the framework and the console repos are up to date and compiled in debug mode.

jnm2:
"Cannot resolve method NUnit.Engine.RunTestsCallbackHandler:RaiseCallbackEvent" stops happening if I check out the `v3.10` tag from nunit-console, so something has been introduced between 3.10 and master. I think by sticking with 3.10 I can leave that problem for some other time.

jnm2:
Reproduces consistently from the command line but not when MonoDevelop is attached.

jnm2:
Since there doesn't seem to be a way to reproduce this in any debugger, I'm going by trial and error and deleting as many tests as I can without losing the repro.

jnm2:
Down to: MaxTimeTests, ThreadingTests, TimeoutTests, and DelayedConstraintTests.

jnm2:
The repro is preserved if you delete all tests except for this one:

https://github.com/nunit/nunit/blob/57238982a64ff29ed738ed8aa719ee6a32649774/src/NUnitFramework/tests/Attributes/TimeoutTests.cs#L129-L138

And any one of these:

https://github.com/nunit/nunit/blob/57238982a64ff29ed738ed8aa719ee6a32649774/src/NUnitFramework/tests/Constraints/DelayedConstraintTests.cs#L183-L245

jnm2:
Got the repro down to this (the only two tests in my nunit.framework.tests project):

```cs
[Test]
public void A()
{
    TimeoutTestCaseFixture fixture = new TimeoutTestCaseFixture();
    TestSuite suite = TestBuilder.MakeFixture(fixture);
    TestMethod testMethod = (TestMethod)TestFinder.Find("TestTimeOutElapsed", suite, false);
    ITestResult result = TestBuilder.RunTest(testMethod, fixture);
    Assert.That(result.ResultState, Is.EqualTo(ResultState.Failure));
    Assert.That(result.Message, Does.Contain("100ms"));
}

[Test]
public void B()
{
    Assert.Throws<AssertionException>(() =>
    {
        Assert.That(() =>
        {
            Thread.Sleep(100);
            return false;
        }, Is.True);
    });
}
```

jnm2:
And with the above code, the debugger can repro.

jnm2:
It looks like the problem is that Mono postpones the ThreadAbortException for a while if you catch it and neither rethrow nor call ResetAbort:

https://github.com/nunit/nunit/blob/57238982a64ff29ed738ed8aa719ee6a32649774/src/NUnitFramework/framework/Internal/Reflect.cs#L272-L278

This seems to have fixed it:

```cs
try
                {
                    return method.Invoke(fixture, args);
                }
                {
                    throw new NUnitException("Rethrown", e.InnerException);
                }
                catch (Exception e)

#if THREAD_ABORT
                    when (!(e is System.Threading.ThreadAbortException))
#endif
                {
                    throw new NUnitException("Rethrown", e);
                }
```

jnm2:
Guess what? This also fixes the TimeoutTests that we've been excluding on Mono for the last ten years! 🎉

@nunit/framework-team Ready for review!

jnm2:
Glad to help!

Issue #3053, opened 2018-10-09
Conversion from TestCase string parameter to DateTimeOffset


Thorarin:
`TestCaseAttribute` includes behavior to perform conversion of some parameter types to the desired target type. Amongst others, conversion from `String` to `DateTime` and from `String` to `TimeSpan` are supported. This is useful, because no constants of these target types can be defined in the `TestCaseAttribute`.

It would be nice if `DateTimeOffset` could be added could be added to the list of supported conversions.

For example:

```C#
[TestCase("2018-10-09 15:15:00+02:00")]
public void MyTest(DateTimeOffset foo)
{
}
```

`Convert.ChangeType` does not appear to work for `DateTimeOffset`, so the code would be similar to the code for `TimeSpan`.

I wouldn't mind creating a PR for this if that helps; it seems fairly straightforward to add.

NUnit version: latest master

jnm2:
I like this. If I understand correctly, `Convert` was a legacy class to help folks migrate to .NET from older VB. Type converters were the modern (.NET Framework 1.0) answer. If we simply get the default type converter for the parameter type and attempted to convert the string that way, would it behave the way you'd like?

@nunit/framework-team What do you think about TypeConverter? Needs `https://www.nuget.org/packages/System.ComponentModel.TypeConverter/` for the .NET Standard 1.4 build, and the .NET team really intends us to use packages like that and report issues if any still exist.

oznetmaster:
Convert is used for anything that implement sthe IConvertible interface.

TypeConverter is for anything else.

I think we should fully support both.

jnm2:
Good. Yes, I'm not proposing to stop using `Convert`. That would almost surely break someone.

stevenaw:
@Thorarin Were you still working on this, or do you mind if give it a try?

jnm2:
@stevenaw I do believe this item is free to work on, since we haven't put it up for grabs yet. All the feedback so far is positive, so let's go ahead.

@rprouse Does it seem fair to go ahead and assign @stevenaw?

PR #3098, opened 2018-11-26
3053 - Conversion from TestCase string parameter to DateTimeOffset

stevenaw:
Fixes https://github.com/nunit/nunit/issues/3053

Updates `TestCaseAttribute` to convert `String` parameter to `DateTimeOffset`
I noticed the existing type conversion logic wasn't applied to params arrays, so I expanded it to work there too (I added a test for the existing TimeSpan logic for that too).

#### NET Standard 1.4
- `System.ComponentModel.TypeConverter`

### Scope Creep
I also noticed that the `ValuesAttribute` handles most type conversions the `TestCaseAttribute` does, so I added this `string -> DateTimeOffset` conversion for it too. I also consolidated some of the type conversion logic between them (`TestCaseAttribute` now uses `ParamAttributeTypeConversions` too). In the process, I had to update `ParamAttributeTypeConversions` to map values to nullable primitives, it looks like previously we only mapped null to nullable primitives.

One potential change to existing behaviour is that `TestCaseAttribute` previously called `TimeSpan.Parse()` without providing a culture, whereas it looked like most other conversions used `CultureInfo.InvariantCulture` (including when `ValuesAttribute` converts a string to a `TimeSpan`. So in the process of centralizing the type conversion logic, `TestCaseAttribute` now also uses `CultureInfo.InvariantCulture` when converting a string to a `TimeSpan`.

In keeping with comment https://github.com/nunit/nunit/issues/3053#issuecomment-428203802, `TimeSpan` now also gets converted using `TypeConverter` instead of `Convert`, as it does not implemement `IConvertible`.

Since this is a bit more than just "Conversion from TestCase string parameter to DateTimeOffset" covered in this PR, let me know if a second GH issue and/or PR should be required for the broader changes.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
@@ -269,7 +269,7 @@ namespace NUnit.Framework
                 // Special handling for ExpectedResult (see if it needs to be converted into method return type)
                 object expectedResultInTargetType;
                 if (parms.HasExpectedResult
-                    && PerformSpecialConversion(parms.ExpectedResult, method.ReturnType.Type, out expectedResultInTargetType))
+                    && ParamAttributeTypeConversions.TryConvert(parms.ExpectedResult, method.ReturnType.Type, out expectedResultInTargetType))
                 {
                     parms.ExpectedResult = expectedResultInTargetType;
                 }
@@ -288,7 +288,7 @@ namespace NUnit.Framework
                             if (!lastParameterType.IsInstanceOfType(parms.Arguments[argsProvided - 1]))
                             {
                                 Array array = Array.CreateInstance(elementType, 1);
-                                var argValue = CoalesceParameterValue(parms.Arguments[argsProvided - 1], elementType);
+                                var argValue = ParamAttributeTypeConversions.Convert(parms.Arguments[argsProvided - 1], elementType);
 
                                 array.SetValue(argValue, 0);
                                 parms.Arguments[argsProvided - 1] = array;
@@ -304,7 +304,7 @@ namespace NUnit.Framework
                             Array array = Array.CreateInstance(elementType, length);
                             for (int i = 0; i < length; i++)
                             {
-                                var argValue = CoalesceParameterValue(parms.Arguments[argsNeeded + i - 1], elementType);
+                                var argValue = ParamAttributeTypeConversions.Convert(parms.Arguments[argsNeeded + i - 1], elementType);
                                 array.SetValue(argValue, i);
                             }
 
@@ -361,15 +361,6 @@ namespace NUnit.Framework
             return parms;
         }
 
-        private static object CoalesceParameterValue(object arg, Type targetType)
-        {
-            object argAsTargetType;
-            if (PerformSpecialConversion(arg, targetType, out argAsTargetType))
-                return argAsTargetType;
-            else
-                return arg;
-        }
-
         /// <summary>
         /// Performs several special conversions allowed by NUnit in order to
         /// permit arguments with types that cannot be used in the constructor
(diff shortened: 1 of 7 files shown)


src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
```diff
@@ -104,22 +115,33 @@ public static bool TryConvert(object value, Type targetType, out object converte
...
-                convertedValue = System.Convert.ChangeType(value, targetType, System.Globalization.CultureInfo.InvariantCulture);
+                Type convertTo = targetType.GetTypeInfo().IsGenericType && targetType.GetGenericTypeDefinition() == typeof(Nullable<>) ?
+                    targetType.GetGenericArguments()[0] : targetType;
+
+                convertedValue = System.Convert.ChangeType(value, convertTo, CultureInfo.InvariantCulture);
+                return true;
+            }
+
+            var converter = TypeDescriptor.GetConverter(targetType);
+            if (converter.CanConvertFrom(value.GetType()))
```
stevenaw: I'd appreciate some input on this especially. Is it better to use `CanConvertFrom`, or to have a specific list of what should be supported (like this: https://github.com/nunit/nunit/pull/3098/files#diff-64bcafa7925c912b792f366a7971facfR123)

src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
```diff
@@ -86,8 +88,17 @@ public static object Convert(object value, Type targetType)
...
+        /// permit arguments with types that cannot be used in the constructor
+        /// of an Attribute such as TestCaseAttribute or to simplify their use.
         /// </summary>
+        /// <param name="value">The value to be converted</param>
+        /// <param name="targetType">The target <see cref="Type"/> in which the <paramref name="value"/> should be converted</param>
+        /// <param name="convertedValue">If conversion was successfully applied, the <paramref name="value"/> converted into <paramref name="targetType"/></param>
+        /// <returns>
+        /// <c>true</c> if <paramref name="value"/> was converted and <paramref name="convertedValue"/> should be used;
+        /// <c>false</c> is no conversion was applied and <paramref name="convertedValue"/> should be ignored
+        /// </returns>
```
stevenaw: This xmldoc comment was copied over from `TestCaseAttribute`'s "PerformSpecialConversion"
https://github.com/nunit/nunit/pull/3098/files#diff-012b7fc50160d70b26586addebf7ce1fL395

src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
ChrisMaddock: I don't see the problem this would create as written, but would appreciate @nunit/framework-team's thoughts on this!

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
```diff
@@ -301,7 +303,10 @@ private TestCaseParameters GetParametersForTestCase(IMethodInfo method)
                             int length = argsProvided - argsNeeded + 1;
                             Array array = Array.CreateInstance(elementType, length);
                             for (int i = 0; i < length; i++)
-                                array.SetValue(parms.Arguments[argsNeeded + i - 1], i);
+                            {
+                                var argValue = ParamAttributeTypeConversions.Convert(parms.Arguments[argsNeeded + i - 1], elementType);
+                                array.SetValue(argValue, i);
+                            }
```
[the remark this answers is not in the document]
stevenaw: Sorry, let me wrap my head back into this.
I "think" this was to support conversion of `string` -> `DateTimeOffset` if the receiving function expected an array of `DateTimeOffset`s, but you are absolutely correct that it would also apply to non-`DateTimeOffset` parameters too. Would this make more sense to split off into a separate issue/PR?

stevenaw:
Thanks for the review @ChrisMaddock !
I gave a quick reply to a question, but it might be a few days until I can parse and respond to the rest of the review. Thanks for pointing out the nuspec thing though!

ChrisMaddock:
Don’t worry - it’s taken me over a month to get round to looking at this! 😅 enjoy New Years!

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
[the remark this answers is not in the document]
ChrisMaddock: Splitting it out sounds really helpful, if that’s ok with you! 😊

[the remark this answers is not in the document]
stevenaw: Slowly chipping away at the PR feedback, but this piece is done. Changes are backed out.

src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
```diff
@@ -104,22 +115,33 @@ public static bool TryConvert(object value, Type targetType, out object converte
 
             bool convert = false;
 
-            if (targetType == typeof(short) || targetType == typeof(byte) || targetType == typeof(sbyte))
+            if (targetType == typeof(short) || targetType == typeof(byte) || targetType == typeof(sbyte) || targetType == typeof(long?) ||
+                targetType == typeof(short?) || targetType == typeof(byte?) || targetType == typeof(sbyte?) || targetType == typeof(double?))
```
jnm2: I'd prefer to use `Nullable.GetUnderlyingType(targetType)` to remove redundancy.

src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
jnm2: I'd like to support all types this way, too.

Should GetConverter be given `Nullable.GetUnderlyingType(targetType)` if it's not null?

src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
```diff
@@ -104,22 +115,33 @@ public static bool TryConvert(object value, Type targetType, out object converte
...
+            else if (targetType == typeof(DateTime) || targetType == typeof(DateTime?))
             {
                 convert = value is string;
             }
 
             if (convert)
             {
-                convertedValue = System.Convert.ChangeType(value, targetType, System.Globalization.CultureInfo.InvariantCulture);
+                Type convertTo = targetType.GetTypeInfo().IsGenericType && targetType.GetGenericTypeDefinition() == typeof(Nullable<>) ?
+                    targetType.GetGenericArguments()[0] : targetType;
```
jnm2: Why not `Nullable.GetUnderlyingType`?

stevenaw:
Thanks for the feedback @jnm2 !
I'm on vacation right now, but should be able to get to this within a week or two

src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
```diff
@@ -23,9 +23,12 @@
 
 using System;
 using System.Collections;
+using System.ComponentModel;
+using System.Globalization;
+
+#if !(NET35 || NET40 || NET45)
```
stevenaw: Removed the unused usings, and wrapped the platform-specific usings in a check

src/NUnitFramework/tests/Attributes/RangeAttributeTests.cs
```diff
@@ -52,7 +52,7 @@ public partial class RangeAttributeTests
         };
 
         // See XML docs for the ParamAttributeTypeConversions class.
-        private static readonly Type[] Int32RangeConvertibleToParameterTypes = { typeof(int), typeof(sbyte), typeof(byte), typeof(short), typeof(decimal) };
+        private static readonly Type[] Int32RangeConvertibleToParameterTypes = { typeof(int), typeof(sbyte), typeof(byte), typeof(short), typeof(decimal), typeof(long), typeof(double) };
```
stevenaw: **NOTE: this change is required based on this line**: https://github.com/nunit/nunit/pull/3098/files#diff-64bcafa7925c912b792f366a7971facfR121

This was a hidden inconsistency between our attributes. `TestCaseAttribute` allowed `int -> long` and `int -> double` but `RangeAttribute` did not explicitly support the conversion. Now they are allowed on both (in nullable and non-nullable forms)

stevenaw:
Thanks for taking the time to review this earlier @jnm2 
I'm unable to update the labels on this PR, but if you have time to re-review, I think I've incorporated all your suggestions.

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
```diff
@@ -263,13 +263,13 @@ private TestCaseParameters GetParametersForTestCase(IMethodInfo method)
                 IParameterInfo[] parameters = method.GetParameters();
                 int argsNeeded = parameters.Length;
                 int argsProvided = Arguments.Length;
-
+
```
jnm2: I'd rather this wasn't in the git history for this line. :D

src/NUnitFramework/tests/Attributes/RangeAttributeTests.cs
jnm2: This is good, we want this. The idea behind this list of special-cased conversions is to emulate C#'s implicit casts.

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
stevenaw: Oops, good catch. Thanks @jnm2 , I've removed the change here

jnm2:
@rprouse Can we get this in 3.12?

mikkelbu:
I'll probably have time to look at this during the weekend.

stevenaw:
Thanks!

PR #3105, opened 2018-12-05
Add tests for use of ApartmentState.Unknown in RequiresThreadAttribute

ChrisMaddock:
Tests for #3058. I wasn't sure of behaviour of myself, so wrote a test to check that, and it made sense to keep it!

Given that ApartmentAttribute has the same behaviour, I also added a test in the same format, and remove the previous test which tested at a lower level, for consistencies sake.

### DIFF
--- a/src/NUnitFramework/tests/Attributes/ApartmentAttributeTests.cs
+++ b/src/NUnitFramework/tests/Attributes/ApartmentAttributeTests.cs
@@ -24,17 +24,24 @@
 #if APARTMENT_STATE
 using System;
 using System.Threading;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+using NUnit.TestData;
+using NUnit.TestUtilities;
 
 namespace NUnit.Framework.Attributes
 {
     [TestFixture]
     public class ApartmentAttributeTests : ThreadingTests
     {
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/tests/Attributes/ApartmentAttributeTests.cs
```diff
@@ -24,17 +24,24 @@
...
     [TestFixture]
     public class ApartmentAttributeTests : ThreadingTests
     {
+#if APARTMENT_STATE
         [Test]
-        public void ApartmentStateUnknownThrowsException()
+        public void ApartmentStateUnknownIsNotRunnable()
         {
-            Assert.That(() => new ApartmentAttribute(ApartmentState.Unknown), Throws.ArgumentException);
+            var testSuite = TestBuilder.MakeFixture(typeof(ApartmentDataRequiresThreadAttribute));
```
mikkelbu: Isn't this the wrong type?
```suggestion
            var testSuite = TestBuilder.MakeFixture(typeof(ApartmentDataApartmentAttribute));
```

ChrisMaddock: Ugh - good spot! Thanks!

PR #3108, opened 2018-12-06
Declare default Program delivered with NUnitLite Nuget package as static class

### LINKED ISSUE #3107: Declare class in Program.cs provided with NUnitLite Nuget package static
The default Program class provided with the NUnitLite Nuget package 3.11.0 is not static.
This causes warnings/errors on some static code analysis like SonarQube (classes only containing static members should be declared static).

As the file is always replaced/recreated when installing the Nuget package, you cannot simple fix the problem.

BTW, the documentation in the Wiki is correctly creating a static class:
https://github.com/nunit/docs/wiki/NUnitLite-Runner

ahoisl:
Fixes #3107

### DIFF
--- a/nuget/nunitlite/Program.cs
+++ b/nuget/nunitlite/Program.cs
@@ -25,7 +25,7 @@ using NUnitLite;
 
 namespace NUnitLite.Tests
 {
-    public class Program
+    public static class Program
     {
         /// <summary>
         /// The main program executes the tests. Output may be routed to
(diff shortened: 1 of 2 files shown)


CharliePoole:
LG 2  me as well.

ahoisl:
I did convert the former VB class to a module.
Just noticed that this is probably a breaking change if somebody inherited from the Program class (both in C# and VB). I'm not able to decide whether this is acceptable nor not

CharliePoole:
It's probably OK, given that this is intended as an example.

PR #3121, opened 2018-12-22
Fix EqualConstraint failure message for enumerables of different size

### LINKED ISSUE #3094: Bad error message if collections have different types
While framework supports checking equality of collections of different type, error message is not very informative in case of failure:

```
Expected is <System.String[1]>, actual is <System.Linq.Enumerable+SelectListIterator`2[NamespaceA.SomeType,System.String]>
  Values differ at index [0]
```

Possible options to fix:
a) always log collections as items (e.g. "< 1,2,3... >"), ignoring type
b) log actual and expected item where failed

(Personally I prefer option a) )

Dreamescaper:
(fixes #3094 )

Uninformative message occurs when there are non-collection arguments, and quantity differs:
```
        [Test]
        public void DifferentCollectionType()
        {
            var expected = new int[] { 1, 2, 3 };
            var actual = expected.Take(2);

            Assert.That(actual, Is.EqualTo(expected));
        }
```

Before fix message:
```
Expected is <System.Int32[3]>, actual is <System.Linq.Enumerable+ListPartition`1[System.Int32]>
  Values differ at index [2]
```

After fix:
```
Expected is <System.Int32[3]>, actual is <System.Linq.Enumerable+ListPartition`1[System.Int32]>
Values differ at index [2]
Missing: 3
```

In PR were added overloads for messages methods, accepting parameter `hasValue`, which in case of `false` means that no value present at all (not even null).

### DIFF
--- a/src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
@@ -271,13 +271,11 @@ namespace NUnit.Framework.Constraints
                         ++depth);
                 else if (failurePoint.ActualHasData)
                 {
-                    writer.Write("  Extra:    ");
-                    writer.WriteValue(failurePoint.ActualValue);
+                    writer.Write($"  Extra:    < {MsgUtils.FormatValue(failurePoint.ActualValue)}, ... >");
                 }
                 else
                 {
-                    writer.Write("  Missing:  ");
-                    writer.WriteValue(failurePoint.ExpectedValue);
+                    writer.Write($"  Missing:  < {MsgUtils.FormatValue(failurePoint.ExpectedValue)}, ... >");
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


ChrisMaddock:
> I think I would prefer to see a message that suggests the single value may not be the only missing value, **as we do for collections**, but I can do without it.

@CharliePoole - can you clarify here? I'm not sure of the message you mean?

CharliePoole:
If memory serves :smile: we do something like...

```
Missing: 3, 4, 5, ...
```

(maybe with angle brackets)

The `...` tells you that there's more stuff not shown.

As indicated, my approval isn't conditioned on this... just an idea.

Dreamescaper:
So in this case - something like this
`Missing: < 3, ... >`
?

CharliePoole:
That would suit me, but I'll go with whatever we all like. Just "3" is a major improvement over what we have now.

Dreamescaper:
Updated

PR #3123, opened 2018-12-23
Fix NRE

### LINKED ISSUE #2696: Getting WorkerId fails in debug
@botinko commented on [Fri Feb 02 2018](https://github.com/nunit/nunit-vs-adapter/issues/166)

Hi! 

1, Visual Studio 2017 Version 15.5.2
2. NUnit test adapter V 3.9 
3, Nunit 3.9 (installed via nuget) 

When I try to debug test from Test Explorer, getting TestContext.CurrentContext.WorkerId throws null reference exception. 
When I debug test in Rider or via Resharper in VS, then all is ok. 

---

@OsirisTerje commented on [Fri Feb 02 2018](https://github.com/nunit/nunit-vs-adapter/issues/166#issuecomment-362671063)

You placed this in the NUnit 2 adapter repo, I'll move it to the NUnit3 adapter repo.

Confirmed.  I get the same.  It seems however, to be an issue in NUnit itself.  
image

It fails in debug
It works with a non-debug run. 

@rprouse Can you have a look?
I'll move this to the NUnit repo.

CharliePoole:
Fixes #2696 

I merely added the obvious `?.` so that null is returned when there is no WorkerId in the context. The issue seems to have degenerated into a broad discussion of who should want to use the WorkerId and why but the NRE is still there.

If somebody wants more done on this issue, then I suggest removing "fixes" above so it doesn't close but merging this trivial change so there's no more NRE.

### DIFF
--- a/src/NUnitFramework/framework/TestContext.cs
+++ b/src/NUnitFramework/framework/TestContext.cs
@@ -121,7 +121,7 @@ namespace NUnit.Framework
         /// </summary>
         public string WorkerId
         {
-            get { return _testExecutionContext.TestWorker.Name; }
+            get { return _testExecutionContext.TestWorker?.Name; }
         }
 #endif
 


ChrisMaddock:
LGTM. I agree that it's sufficient to fix the NRE for now - if someones looking for the value not to be null, then that's a separate discussion, and dependent on the particular circumstances. 👍

PR #3131, opened 2018-12-30
Only create PlatformHelper when required/Don't throw for MonoTouch

### LINKED ISSUE #3068: First Chance Exception in RuntimeFramework
I've started seeing a first chance exception from line 82 below. Looking at the code, it makes sense that this might be suppressed by adding the `false` parameter, to suppress errors due to missing dependencies.

https://github.com/nunit/nunit/blob/a5edc87c34c25676063c84fe8f902544b12da7b2/src/NUnitFramework/framework/Internal/RuntimeFramework.cs#L80-L85

What's unclear at the moment is why I've never seen this before - this codes been in place for 3 years. Filing an issue rather than a PR, and I'll aim to dig into it soon. I suspect it's an implementation detail difference between .NET Core and .NET Framework

Call is coming via the creation of a PlatformHelper in TestCaseAttribute.BuildFrom. Incidentally, we should skip that call if the IncludePlatform and ExcludePlatform properties are empty - creating a PlatformHelper calculates the current framework, and all available frameworks, so ain't cheap.

Using NUnit 3.11, .NET Standard 2.0 running on .NET Core 2.1 on Windows.

ChrisMaddock:
Fixes #3068. Two changes:

1. I was hitting a first-chance exception when evaluating the current framework. I couldn't actually replicate this when I came back to it, I imagine the issue may have been a dodgy MonoTouch installation on my work machine. Either was that shouldn't cause an exception here, so I've err'd on the side of caution with these low-popularity platform `GetType()` calls.

2. More importantly, the first TestCase attribute in an assembly was previously required to work out the current runtime framework and OS platform, regardless of whether the information was required or not. This isn't a cheap operation - so it now only happens if the IncludePlatform/ExcludePlatform properties are in use.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
@@ -454,7 +454,6 @@ namespace NUnit.Framework
             TestMethod test = new NUnitTestCaseBuilder().BuildTestMethod(method, suite, GetParametersForTestCase(method));
 
 #if PLATFORM_DETECTION
-
             if (IncludePlatform != null || ExcludePlatform != null)
             {
                 if (test.RunState == RunState.NotRunnable || test.RunState == RunState.Ignored)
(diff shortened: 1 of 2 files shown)


ChrisMaddock:
Thanks!

PR #3133, opened 2018-12-30
[WIP] Remove TypeHelper.ConvertArgumentList

### LINKED ISSUE #3125: TestCaseSource "Single" cannot be converted to "int" issue
I don't know if this is an issue or a feature :) but a TestCaseSource cannot be used in two methods with different parameter types, in the following case, the second test fails with  Object of type 'System.Single' cannot be converted to type 'System.Int32'

`class Test
   {
      private static readonly TestCaseData[] Cases =
      {
         new TestCaseData("", 0),
         new TestCaseData("", 1)
      };
      
      public void TestA(string a, float b)
      {
      }

      public void TestB(string a, int b)
      {
      }
   }`

mikkelbu:
The added tests failed with the message "Object of type 'System.Single'
cannot be converted to type 'System.Int32'. The removal of
`TypeHelper.ConvertArgumentList` solves the problem.

Fixes #3125

### DIFF
--- a/src/NUnitFramework/framework/Internal/TypeHelper.cs
+++ b/src/NUnitFramework/framework/Internal/TypeHelper.cs
@@ -222,45 +222,6 @@ namespace NUnit.Framework.Internal
         }
 
         /// <summary>
-        /// Convert an argument list to the required parameter types.
-        /// Currently, only widening numeric conversions are performed.
-        /// </summary>
-        /// <param name="arglist">An array of args to be converted</param>
-        /// <param name="parameters">A ParameterInfo[] whose types will be used as targets</param>
-        public static void ConvertArgumentList(object[] arglist, IParameterInfo[] parameters)
-        {
-            System.Diagnostics.Debug.Assert(arglist.Length <= parameters.Length);
-
-            for (int i = 0; i < arglist.Length; i++)
-            {
-                object arg = arglist[i];
-
-                if (arg is IConvertible)
-                {
-                    Type argType = arg.GetType();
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/framework/Internal/TypeHelper.cs
```diff
@@ -221,45 +221,6 @@ public static bool IsNumeric(Type type)
...
-                    if (argType != targetType && IsNumeric(argType) && IsNumeric(targetType))
-                    {
-                        if (targetType == typeof(double) || targetType == typeof(float))
-                            convert = arg is int || arg is long || arg is short || arg is byte || arg is sbyte;
-                        else
-                            if (targetType == typeof(long))
-                                convert = arg is int || arg is short || arg is byte || arg is sbyte;
-                            else
-                                if (targetType == typeof(short))
-                                    convert = arg is byte || arg is sbyte;
```
jnm2: This section mimics a subset of the C# language's implicit conversions. Can you check to see if these conversions still work after your change? I think [ParamAttributeTypeConversions](https://github.com/nunit/nunit/blob/db5bee0f8bb74d68202377917f6d958c0e139bfd/src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs#L107-L118) may be handling it but I'm not sure if it acts exactly the same.

It'll make me feel better since it's been this way for seven years: https://github.com/nunit/nunit/commit/57f64ef3a3b0c52a73df09512687d11c421d60a9#diff-7842b7101011893b22318d35b285ed47R337

rprouse:

PR #3134, opened 2018-12-30
Require non-whitespace string as test name

### LINKED ISSUE #3038: TestName in TestCase attribute not validated to be not empty
Hi everyone! Thank you for NUnit!

Using NUnit 3.10.1

The usage of NUnit in our project in rather advanced. We use many publicly available features as well as some undocumented ones ;)

And we have a homebrew integration of NUnit and ReportPortal (by EPAM) which attaches itself to NUnit engine like a parasite, for the information about currently running tests to be transfered to RP in real time (not after the test run is completed) even when running tests in debug mode from VS. We use the information supplied by NUnit such as test names, descriptions, categories, ignore reasons etc to make test reports as easy to read as possible.

Several days ago reports at RP started indicating integration problems which occur very rarely. I took a deeper look and found errors at RP logs saying that test name constraint "not empty" fails test report creation. Further research pointed at a new test in our project which had its attributes defined like this:

```csharp
[Test]
[TestCase(TestName = "")]
[Category("DB")]
```

We use this workaround with TestCase for having a human-readable test name supplied by NUnit instead of coding a custom attribute. And in many cases we use TestCase attribute for its intended purpose as well.

So obviously one programmer forgot to type in the value for TestName. The integration code was looking at ITest.Name property which appeared to be an empty string. And an empty string in its turn appeared to be unacceptable as a test name for RP.

If you run this test with Resharper, it displays a test icon with nothing next to it (no name), which means Resharper relies on the same NUnit functionality.

I've added a workaround:
```csharp
internal static string GetName(this ITest test) 
    => string.IsNullOrWhiteSpace(test.Name) ? test.MethodName : test.Name;
```

But the expected behavior is one of these:
1. NUnit exception during initialization, like it happens when parameters count/types of the test method do not match with those of TestCase attribute
1. If TestName is empty then provide a default generated value instead, like as if there was no TestName property defined at all

Thanks in advance! :)

mikkelbu:
If the test name is specified - and different from null - then it must
be a non-whitespace string.

Fixes #3038

### DIFF
--- a/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
@@ -228,6 +228,9 @@ namespace NUnit.Framework.Internal.Builders
                 parameters = testMethod.Method.GetParameters();
             }
 
+            if (parms != null && parms.TestName != null && parms.TestName.Trim() == "")
+                return MarkAsNotRunnable(testMethod, "Test has no name");
+
             if (arglist != null && parameters != null)
                 TypeHelper.ConvertArgumentList(arglist, parameters);
 
--- a/src/NUnitFramework/testdata/TestNameFixture.cs
+++ b/src/NUnitFramework/testdata/TestNameFixture.cs
@@ -1,0 +1,55 @@
+// ***********************************************************************
+// Copyright (c) 2018 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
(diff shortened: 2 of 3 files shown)


src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
```diff
@@ -228,6 +228,9 @@ private static bool CheckTestMethodSignature(TestMethod testMethod, TestCasePara
                 parameters = testMethod.Method.GetParameters();
             }
 
+            if (parms != null && parms.TestName != null && parms.TestName.Trim() == "")
+                return MarkAsNotRunnable(testMethod, "Test has no name");
```
ChrisMaddock:
```suggestion
                 return MarkAsNotRunnable(testMethod, "Test name cannot be all white-space or empty.");
```

I'm not sure on "Test has no name" - how about something more explicit like this instead? 🙂

Issue #3156, opened 2019-02-07
UnexpectedExceptionTests should tolerate Mono on Azure DevOps Ubuntu


jnm2:
The Linux build failed on all the Mono tests:

```
1) Failed : NUnit.Framework.Internal.UnexpectedExceptionTests.FailRecordsInnerExceptionsAsPartOfAggregateException
  Expected string length 153 but was 203. Strings differ at index 53.
  Expected: "...on : Outer Aggregate Exception\n  ----> System.Exception : ..."
  But was:  "...on : Outer Aggregate Exception (Inner Exception 1 of 2) (I..."
  --------------------------------------------^
  at NUnit.Framework.Internal.UnexpectedExceptionTests.FailRecordsInnerExceptionsAsPartOfAggregateException () [0x0005e] in <2086766ee22743e488729c0438b8b918>:0 

2) Failed : NUnit.Framework.Internal.UnexpectedExceptionTests.FailRecordsNestedInnerExceptionAsPartOfAggregateException
  Expected string length 145 but was 163. Strings differ at index 53.
  Expected: "...on : Outer Aggregate Exception\n  ----> System.Exception : ..."
  But was:  "...on : Outer Aggregate Exception (Inner Exception)\n  ----> S..."
  --------------------------------------------^
  at NUnit.Framework.Internal.UnexpectedExceptionTests.FailRecordsNestedInnerExceptionAsPartOfAggregateException () [0x0005e] in <2086766ee22743e488729c0438b8b918>:0 
```

https://nunit.visualstudio.com/NUnit/_build/results?buildId=1308&view=logs&jobId=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&taskId=8ef82b3b-1feb-5bbd-06f6-b1f7b5467f03&lineStart=708&lineEnd=720&colStart=1&colEnd=196

These tests need to be made less specific so that they tolerate platform differences.

mikkelbu:
Looks like it behaves like the case for `NETCOREAPP1_1 || NETCOREAPP2_0` in the code. What if we changed the assertion to assert that:
* The message starts with `System.AggregateException : Outer Aggregate Exception`
* The message ends with 
```             
"  ----> System.Exception : Inner Exception 1 of 2" + Environment.NewLine +
"  ----> System.Exception : Inner Exception 2 of 2";
```

jnm2:
@mikkelbu That sounds good.

PR #3158, opened 2019-02-07
Catch exceptions preventing return of awaitable object

### LINKED ISSUE #3138: Assert.Ignore breaks when a Task is returned w/o using async/await
Writing a test that invokes `Assert.Ignore()` and it fails with the following exception:

<details><summary>Stacktrace</summary>

```
NUnit.Framework.Internal.NUnitException : Rethrown
  ----> NUnit.Framework.IgnoreException : should not run
   at NUnit.Framework.Internal.Reflect.InvokeMethod(MethodInfo method, Object fixture, Object[] args) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Reflect.cs:line 281
   at NUnit.Framework.Internal.AsyncToSyncAdapter.Await(Func`1 invoke) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\AsyncToSyncAdapter.cs:line 77
   at NUnit.Framework.Internal.Commands.TestMethodCommand.RunTestMethod(TestExecutionContext context) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Commands\TestMethodCommand.cs:line 84
--IgnoreException
   at NUnit.Framework.Assert.Ignore(String message, Object[] args) in C:\src\nunit\nunit\src\NUnitFramework\framework\Assert.cs:line 214
   at TestNUnit.Class1.Failing() in C:\Users\Sean\Desktop\TestNUnit\TestNUnit\Class1.cs:line 16
```
</details>
<br>

The failing test looks as the following:

```c#
        [Test]
        public Task Failing()
        {
            Assert.Ignore("should not run");

            return Task.FromResult(1);
        }
```

Changing the test to include the `async` keyword with `await` fixes this test:

```c#
        [Test]
        public async Task Working()
        {
            Assert.Ignore("should not run");

            await Task.FromResult(1);
        }
```

This fails with NUnit 3.11.0, but works with earlier versions.
Tested with .NET Framework 4.5.2

jnm2:
Fixes #3138.

The problem was was an NUnitException with an InnerException of _another_ NUnitException with an InnerException of a Success/Ignore/InconclusiveException. `ValidateAndUnwrap` only unwraps one level, so `if (ex is ResultStateException)` never happens because the ResultStateException it's trying to find is actually `ex.InnerException` at that point.

https://github.com/nunit/nunit/blob/b5c1d549938476ff06f118557b29778b36593549/src/NUnitFramework/framework/Internal/Results/TestResult.cs#L578-L585

https://github.com/nunit/nunit/blob/b5c1d549938476ff06f118557b29778b36593549/src/NUnitFramework/framework/Internal/Results/TestResult.cs#L560-L568

The problem was caused in https://github.com/nunit/nunit/commit/b83f556cc41b3c12a6e1a7729d1e05e5e140f252#diff-6325ab2a4b2ce8d3a6fd52a0f58f75f5L98 by me keeping the `catch` block but failing to notice that it used to only apply to the `WaitForPendingOperationsToComplete` part and not the `Reflect.InvokeMethod` part which already wraps.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/TestMethodCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TestMethodCommand.cs
@@ -79,14 +79,7 @@ namespace NUnit.Framework.Internal.Commands
 #if ASYNC
             if (AsyncToSyncAdapter.IsAsyncOperation(testMethod.Method.MethodInfo))
             {
-                try
-                {
-                    return AsyncToSyncAdapter.Await(() => InvokeTestMethod(context));
-                }
-                catch (Exception e)
-                {
-                    throw new NUnitException("Rethrown", e);
-                }
+                return AsyncToSyncAdapter.Await(() => InvokeTestMethod(context));
             }
 #endif
             return InvokeTestMethod(context);
--- a/src/NUnitFramework/testdata/AsyncRealFixture.cs
+++ b/src/NUnitFramework/testdata/AsyncRealFixture.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
@@ -29,6 +29,9 @@ using NUnit.Framework;
 #if NET40
 using Task = System.Threading.Tasks.TaskEx;
 #endif
+
+// Don’t warn when async is used without await
+#pragma warning disable CS1998
 
 namespace NUnit.TestData
 {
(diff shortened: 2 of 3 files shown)


src/NUnitFramework/framework/Internal/Commands/TestMethodCommand.cs
```diff
@@ -79,14 +79,7 @@ private object RunTestMethod(TestExecutionContext context)
 #if ASYNC
             if (AsyncToSyncAdapter.IsAsyncOperation(testMethod.Method.MethodInfo))
             {
-                try
-                {
-                    return AsyncToSyncAdapter.Await(() => InvokeTestMethod(context));
-                }
-                catch (Exception e)
-                {
-                    throw new NUnitException("Rethrown", e);
```
jnm2: I'd like to deprecate `NUnitException` entirely in the future.

jnm2:
Build failure is unrelated. See https://github.com/nunit/nunit/issues/3156.

src/NUnitFramework/framework/Internal/Commands/TestMethodCommand.cs
mikkelbu: So not wrapping the exception in a `NUnitException` solves the problem?
(Actually, I got most information from the title of the second commit `Fix twice-wrapped NUnitException to fix Assert.Pass, Ignore, and Inconclusive in non-async Task-returning methods` 😄).

CharliePoole: NUnitException is an implementation detail that's never supposed to escape from the framework, so there shouldn't be anything to deprecate.

It was used internally to indicate that the wrapped exception has special meaning for NUnit and shouldn't be handled as an exception. It's predecessor in V2 was used more heavily than it is today so you could phase it out of the implementation, but there doesn't seem to be a lot of benefit in putting effort into removing it.

jnm2: @mikkelbu Correct. It was an NUnitException with an InnerException of an NUnitException with the InnerException of the Success/Ignore/InconclusiveException. `ValidateAndUnwrap` only unwraps one level, so `if (ex is ResultStateException)` fails because the ResultStateException is actually `ex.InnerException` at that point.

https://github.com/nunit/nunit/blob/b5c1d549938476ff06f118557b29778b36593549/src/NUnitFramework/framework/Internal/Results/TestResult.cs#L578-L585

https://github.com/nunit/nunit/blob/b5c1d549938476ff06f118557b29778b36593549/src/NUnitFramework/framework/Internal/Results/TestResult.cs#L560-L568

The problem was caused in https://github.com/nunit/nunit/commit/b83f556cc41b3c12a6e1a7729d1e05e5e140f252#diff-6325ab2a4b2ce8d3a6fd52a0f58f75f5L98 by me keeping the `catch` block but failing to notice that it used to only apply to the `WaitForPendingOperationsToComplete` part and not the `Reflect.InvokeMethod` part which already wraps.

jnm2:
Rebuilt the PR, and all checks now pass.

src/NUnitFramework/framework/Internal/Commands/TestMethodCommand.cs
jnm2: @mikkelbu I moved the explanation to the PR body. Is there anything else I can clarify?

PR #3160, opened 2019-02-09
nunit.org url to https in nuspecs

304NotModified:


ChrisMaddock:
Thanks for picking up on this @304NotModified! Looks great, I'll merge now.

We could do with doing this across all our packages, e.g. the NUnit Console and the Adapter as well. Any you'd be willing to help with would be very welcome! 😄

PR #3162, opened 2019-02-10
Update links to HTTPS

jnm2:
I liked https://github.com/nunit/nunit/pull/3160 so much that I did all the rest.

### DIFF
--- a/src/NUnitFramework/framework/Compatibility/System.Collections/CollectionDebuggerView.cs
+++ b/src/NUnitFramework/framework/Compatibility/System.Collections/CollectionDebuggerView.cs
@@ -7,7 +7,7 @@
 // Authors:
 //	Marek Safar  <marek.safar@gmail.com>
 //
-// Copyright (C) 2009 Novell, Inc (http://www.novell.com)
+// Copyright (C) 2009 Novell, Inc (https://www.novell.com)
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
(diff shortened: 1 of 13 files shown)


src/NUnitFramework/framework/Compatibility/System.Collections/CollectionDebuggerView.cs
```diff
@@ -7,7 +7,7 @@
 // Authors:
 //	Marek Safar  <marek.safar@gmail.com>
 //
-// Copyright (C) 2009 Novell, Inc (http://www.novell.com)
```
ChrisMaddock: D'ya think the two links in copyright notices should be left as is? Otherwise everything else in this PR makes me very happy! 😄

jnm2: I don't know why not, but I can change if there's a concern. Both forms redirect to their new domain name.

ChrisMaddock: I'm not too worried, I'd just normally err on the side of safety myself and not touch copyright notices - as I don't really understand the details! I'll approve, and leave it with you to decide whether to merge as is, or make the change. 🙂

jnm2: If it's up to me, I'm going to act as though this is fine because I don't want to be in a world where it isn't. I guess we'll find out if that's realistic. 😃

PR #3172, opened 2019-02-21
Copy test fixtures with filters

### LINKED ISSUE #3171: Incorrect type for Test Fixtures when using running explore with a filter
Running nunit.framework v 3.11

Related to [nunit/nunit-console/issues/409](https://github.com/nunit/nunit-console/issues/409), and introduced by  #2094, test fixtures are improperly marked with type="TestSuite" when running explore with a filter. 

This is an example of an explore output with no filter:
```xml
<?xml version="1.0" encoding="utf-8" standalone="no"?>
<test-run id="2" testcasecount="219">
  <test-suite type="Assembly" id="0-1269" name="My.Tests.dll" fullname="My.Tests.Tests.dll" runstate="Runnable" testcasecount="219">
    <properties>
      <property name="Category" value="SmokeTest" />
      <property name="ParallelScope" value="Fixtures" />
      <property name="_PID" value="7132" />
      <property name="_APPDOMAIN" value="domain-" />
    </properties>
    <test-suite type="TestSuite" id="0-1270" name="My" fullname="Apt" runstate="Runnable" testcasecount="219">
          <test-suite type="TestSuite" id="0-1273" name="Tests" fullname="My.Tests" runstate="Runnable" testcasecount="219">
            <test-suite type="TestFixture" id="0-1000" name="ResxTest" fullname="My.Tests.ResxTest" classname="My.Tests.ResxTest" runstate="Runnable" testcasecount="2">
              <properties>
                <property name="Category" value="Resources" />
              </properties>
...
```
Where as this is an example with a filter 'cat == Resources'
```xml
<?xml version="1.0" encoding="utf-8" standalone="no"?>
<test-run id="2" testcasecount="219">
  <test-suite type="Assembly" id="0-1269" name="My.Tests.dll" fullname="My.Tests.Tests.dll" runstate="Runnable" testcasecount="219">
    <test-suite type="TestSuite" id="0-1270" name="My" fullname="Apt" runstate="Runnable" testcasecount="219">
          <test-suite type="TestSuite" id="0-1273" name="Tests" fullname="My.Tests" runstate="Runnable" testcasecount="219">
            <test-suite type="TestSuite" id="0-1000" name="ResxTest" fullname="My.Tests.ResxTest" runstate="Runnable" testcasecount="2">
              <properties>
                <property name="Category" value="Resources" />
              </properties>
...
```

The linked issue also mentions that properties are missing which is true, but I want to limit the scope of this issue since it seems like a simpler fix.

cjconrad2:
Fixes #3171 

When reconstructing test suites with filters, check if the given suite is test fixture and construct that instead.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Tests/ParameterizedFixtureSuite.cs
+++ b/src/NUnitFramework/framework/Internal/Tests/ParameterizedFixtureSuite.cs
@@ -43,10 +43,10 @@ namespace NUnit.Framework.Internal
         }
 
         /// <summary>
-        /// Copy constructor style to create a filtered copy of the given parameterized fixture suite
+        /// Creates a copy of the given suite with only the descendants that pass the specified filter.
         /// </summary>
-        /// <param name="suite">Parameterized Fixture Suite to copy</param>
-        /// <param name="filter">Filter to be applied</param>
+        /// <param name="suite">The <see cref="ParameterizedFixtureSuite"/> to copy.</param>
+        /// <param name="filter">Determines which descendants are copied.</param>
         public ParameterizedFixtureSuite(ParameterizedFixtureSuite suite, ITestFilter filter)
             : base(suite, filter)
         {
@@ -66,9 +66,9 @@ namespace NUnit.Framework.Internal
         }
 
         /// <summary>
-        /// Overriden to return a Parameterized Fixture Suite
+        /// Creates a filtered copy of the test suite.
         /// </summary>
-        /// <param name="filter">Filter to apply</param>
+        /// <param name="filter">Determines which descendants are copied.</param>
         public override TestSuite Copy(ITestFilter filter)
         {
             return new ParameterizedFixtureSuite(this, filter);
--- a/src/NUnitFramework/framework/Internal/Tests/ParameterizedFixtureSuite.cs
+++ b/src/NUnitFramework/framework/Internal/Tests/ParameterizedFixtureSuite.cs
@@ -43,6 +43,16 @@ namespace NUnit.Framework.Internal
         }
 
         /// <summary>
+        /// Copy constructor style to create a filtered copy of the given parameterized fixture suite
+        /// </summary>
+        /// <param name="suite">Parameterized Fixture Suite to copy</param>
+        /// <param name="filter">Filter to be applied</param>
+        public ParameterizedFixtureSuite(ParameterizedFixtureSuite suite, ITestFilter filter)
+            : base(suite, filter)
+        {
+        }
+
+        /// <summary>
         /// Gets a string representing the type of test
         /// </summary>
         public override string TestType
@@ -53,6 +63,15 @@ namespace NUnit.Framework.Internal
                     ? "GenericFixture"
                     : "ParameterizedFixture";
             }
+        }
+
+        /// <summary>
+        /// Overriden to return a Parameterized Fixture Suite
+        /// </summary>
+        /// <param name="filter">Filter to apply</param>
+        public override TestSuite Copy(ITestFilter filter)
+        {
+            return new ParameterizedFixtureSuite(this, filter);
         }
     }
 }
--- a/src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
+++ b/src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
@@ -117,7 +117,9 @@ namespace NUnit.Framework.Internal
                 {
                     if(child.IsSuite)
                     {
-                        TestSuite childSuite = new TestSuite(child as TestSuite, filter);
+                        TestSuite childSuite = child.GetType() == typeof(TestFixture)
+                            ? new TestFixture(child as TestSuite, filter)
+                            : new TestSuite(child as TestSuite, filter);
                         childSuite.Parent    = this;
                         this.tests.Add(childSuite);
                     }
--- a/src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
+++ b/src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
@@ -99,10 +99,10 @@ namespace NUnit.Framework.Internal
         }
 
         /// <summary>
-        /// Copy constructor style to create a filtered copy of the given test suite
+        /// Creates a copy of the given suite with only the descendants that pass the specified filter.
         /// </summary>
-        /// <param name="suite">Test Suite to copy</param>
-        /// <param name="filter">Filter to be applied</param>
+        /// <param name="suite">The <see cref="TestSuite"/> to copy.</param>
+        /// <param name="filter">Determines which descendants are copied.</param>
         public TestSuite(TestSuite suite, ITestFilter filter)
             : base(suite.Name)
         {
@@ -117,7 +117,7 @@ namespace NUnit.Framework.Internal
                 {
                     if(child.IsSuite)
                     {
-                        TestSuite childSuite = ((TestSuite) child).Copy(filter);
+                        TestSuite childSuite = ((TestSuite)child).Copy(filter);
                         childSuite.Parent    = this;
                         this.tests.Add(childSuite);
                     }
@@ -180,10 +180,9 @@ namespace NUnit.Framework.Internal
         }
 
         /// <summary>
-        /// Creates a filtered copy of the test suite
+        /// Creates a filtered copy of the test suite.
         /// </summary>
-        /// <param name="filter">Filter to apply</param>
-        /// <returns></returns>
+        /// <param name="filter">Determines which descendants are copied.</param>
         public virtual TestSuite Copy(ITestFilter filter)
         {
             return new TestSuite(this, filter);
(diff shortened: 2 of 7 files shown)


src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
```diff
@@ -117,7 +117,9 @@ public TestSuite(TestSuite suite, ITestFilter filter)
                 {
                     if(child.IsSuite)
                     {
-                        TestSuite childSuite = new TestSuite(child as TestSuite, filter);
+                        TestSuite childSuite = child.GetType() == typeof(TestFixture)
```
mikkelbu: Personally, I think we should go for a more general solution for all classes deriving from `TestSuite` and not only handle `TestFixture` as a special case. We could introduce a copy constructor for each deriving class and also an instance method that returns an instance of the same type as the instance, e.g. in `TestFixture`
```c#
public void Copy(TestSuite suite, ITestFilter filter)
{
    return new TestFixture(suite, filter);
}
```
then we could replace the code with
```c#
TestSuite childSuite = child.Copy(child as TestSuite, filter);
```

jnm2: I agree. We should do whatever it takes to avoid a type test.

src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
```diff
@@ -117,7 +117,9 @@ public TestSuite(TestSuite suite, ITestFilter filter)
...
+                            ? new TestFixture(child as TestSuite, filter)
+                            : new TestSuite(child as TestSuite, filter);
```
jnm2: These should be direct casts. Direct casts fail closer to the source of the problem and are information-rich. Safe casts are only useful if `null` is a valid value; otherwise, they produce a much harder to understand null reference exception at a later point in time.

cjconrad2: Agreed. With the new copy method, these casts go away anyway.

src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
cjconrad2: Makes sense. I've updated with something similar to @mikkelbu's proposal.

src/NUnitFramework/framework/Internal/Tests/ParameterizedFixtureSuite.cs
```diff
@@ -42,6 +42,16 @@ public ParameterizedFixtureSuite(ITypeInfo typeInfo) : base(typeInfo.Namespace,
             _genericFixture = typeInfo.ContainsGenericParameters;
         }
 
+        /// <summary>
+        /// Copy constructor style to create a filtered copy of the given parameterized fixture suite
```
jnm2: I want to move in the direction of using end marks (period, question mark or colon + list) for all sentences in user-facing messages.

You'd call this a 'copy-style constructor.' 'Copy constructor style' doesn't mean anything to me. Maybe:

```suggestion
        /// Creates a copy of the given suite with only the descendants that pass the specified filter.
```

(Same comment for all occurrences.)

src/NUnitFramework/framework/Internal/Tests/ParameterizedFixtureSuite.cs
```diff
@@ -42,6 +42,16 @@ public ParameterizedFixtureSuite(ITypeInfo typeInfo) : base(typeInfo.Namespace,
...
+        /// </summary>
+        /// <param name="suite">Parameterized Fixture Suite to copy</param>
```
jnm2: I'd like to either use `The <see cref="ParameterizedFixtureSuite"/> to copy.` or `The suite to copy.`

(Same comment for all occurrences.)

src/NUnitFramework/framework/Internal/Tests/ParameterizedFixtureSuite.cs
```diff
@@ -42,6 +42,16 @@ public ParameterizedFixtureSuite(ITypeInfo typeInfo) : base(typeInfo.Namespace,
...
+        /// <param name="filter">Filter to be applied</param>
```
jnm2:
```suggestion
        /// <param name="filter">Determines which descendants are copied.</param>
```

(Same comment for all constructors _and_ Copy methods.)

src/NUnitFramework/framework/Internal/Tests/ParameterizedFixtureSuite.cs
```diff
@@ -54,5 +64,14 @@ public override string TestType
                     : "ParameterizedFixture";
             }
         }
+
+        /// <summary>
+        /// Overriden to return a Parameterized Fixture Suite
+        /// </summary>
+        /// <param name="filter">Filter to apply</param>
```
jnm2: `Overriden` is a typo, but I'd also recommend copying the entire base method's documentation exactly.

src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
```diff
@@ -117,7 +117,7 @@ public TestSuite(TestSuite suite, ITestFilter filter)
                 {
                     if(child.IsSuite)
                     {
-                        TestSuite childSuite = new TestSuite(child as TestSuite, filter);
+                        TestSuite childSuite = ((TestSuite) child).Copy(filter);
```
jnm2: I think our style guide recommends the traditional C# style of no space after direct cast. ReSharper makes Visual Studio start adding the space unless we add `csharp_space_after_cast = false` to our .editorconfig. Ignore the fact that you're seeing some unusual styles in the parts of the file you aren't touching. :D

I'd be grateful if you took this change in the meantime:

```suggestion
                        TestSuite childSuite = ((TestSuite)child).Copy(filter);
```

src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
```diff
@@ -179,6 +179,16 @@ public void Add(Test test)
             tests.Add(test);
         }
 
+        /// <summary>
+        /// Creates a filtered copy of the test suite
+        /// </summary>
+        /// <param name="filter">Filter to apply</param>
+        /// <returns></returns>
```
jnm2: No `<returns>`, please. (https://github.com/nunit/docs/wiki/Best-practices-for-XML-documentation#returns-and-value)

```suggestion
```

cjconrad2:
Fixed doc comments and styling errors. I wrote tests that confirm that the proper type is maintained when calling copy. I believe we have coverage of the TestSuite copy constructor from the original PR (https://github.com/nunit/nunit/pull/2094/files), but let me know if there is anything else you think I should cover!

src/NUnitFramework/tests/Internal/TestSuiteCopyTests.cs
```diff
@@ -0,0 +1,69 @@
...
+    // Tests that Copy is properly overridden for all types extending TestSuite - #3171
+    public class TestSuiteCopyTests
+    {
+        [Test]
+        public void CopyTestSuiteReturnsCorrectType()
+        {
+            TestSuite testSuite =
+                new TestSuite(new TypeWrapper(typeof(NUnit.TestData.ParameterizedTestFixture)));
+            var copiedTestSuite = testSuite.Copy(TestFilter.Empty);
+            Assert.AreEqual(copiedTestSuite.GetType(), typeof(TestSuite));
```
jnm2: Expected and actual are backwards in each test.

This is exactly why I dislike the Assert.AreEqual syntax. Elsewhere we use `Assert.That` and I personally find that more intuitive: `Assert.That(copy, Is.TypeOf<TestSuite>());`

jnm2: @cjconrad2 Sorry to bug you since I'm sure you're busy, but I want to make sure that my requested change was clear. Either swapping the AreEqual arguments in each test or switching to Assert.That syntax is fine.

cjconrad2: Sorry about the delay all, and thank you for your patience! I had a long vacation and I have spent the last week or so catching up. I updated the tests and decided that I also prefer the Assert.That syntax more. Thanks for suggesting that! Hopefully we can merge soon

jnm2: The delay is no problem for us. I wanted to make sure I wasn't the cause of the delay, but I'm sorry I interrupted your vacation!

cjconrad2:
Hmm looks like the build failed with a timeout. I think it is likely transient. I'm sure there a way to rerun it without committing again, but I cannot seem to find it. Any ideas?

jnm2:
@cjconrad2 Only team members can restart Travis builds or jobs at the moment, but thanks for pointing it out. I've just restarted the failing job.

cjconrad2:
That makes sense. Thanks @jnm2!

Issue #2562, opened 2017-11-15
Support generic type parameters for TestCaseSourceAttribute


DustinKingen:
I am working on some test cases that may be better designed using generics. I am trying to use `TestCaseSource` with many different combinations of types.

The problem that I am encountering is the use of reflection in my tests to invoke a generic test method that I use for object validation.

```csharp
[TestFixture]
public class SomeTests
{
    public static IEnumerable<TestCaseData> TestCases()
    {
        yield return
            new TestCaseData(
                typeof(Foo),
                typeof(Bar),
                Generator.Foo,
                Assertions.Foo);
        yield return
            new TestCaseData(
                typeof(Bar),
                typeof(Foo),
                Generator.Bar,
                Assertions.Bar);
    }

    [TestCaseSource(nameof(TestCases))]
    public void CheckSomeTypes(
        Type sourceType,
        Type destinationType,
        TSource source,
        Action<object, object> assert)
    {
        GetType()
            .GetMethod(nameof(CheckSomeTypesImpl), BindingFlags.Public | BindingFlags.Instance)
            .MakeGenericType(new[] { sourceType, destinationType })
            .Invoke(this, new object[] { source, assert });
    }

    public void CheckSomeTypesImpl<TSource, TDestination>(
        TSource source,
        Action<TSource, TDestination> assert)
    {
        var destination = Mapper.Map<TSource, TDestination>(source);

        // Assert are equiviliant
        assert(source, destination);
    }
}
```

However, I believe the following design may add value in this instance:

```csharp
[TestFixture]
public class SomeTests
{
    public static IEnumerable<TestCaseData> TestCases()
    {
        yield return
            new TestCaseData(
                typeof(Foo),
                typeof(Bar),
                Generator.Foo,
                Assertions.Foo);
        yield return
            new TestCaseData(
                typeof(Bar),
                typeof(Foo),
                Generator.Bar,
                Assertions.Bar);
    }

    [TestCaseSource(nameof(TestCases))]
    public void CheckSomeTypes<TSource, TDestination>(
        TSource source,
        Action<TSource, TDestination> assert)
    {
        var destination = Mapper.Map<TSource, TDestination>(source);

        // Assert are equiviliant
        assert(source, destination);
    }
}

public class Foo
{
    public string Qux { get; set; }
}

public class Bar
{
    public string Qux { get; set; }
}

public static class Generator
{
    public static Foo Foo =>
        new Foo { Qux = "TestTestTest" };
[...]

PR #3177, opened 2019-02-28
added TestCaseGeneric and TestCaseSourceGeneric attributes with tests

wernervn:
I had a requirement for a generic test case and came across [#2562](https://github.com/nunit/nunit/issues/2562) and [#1215](https://github.com/nunit/nunit/issues/1215). Further search brought me to a StackOverflow [answer](https://stackoverflow.com/questions/2364929/nunit-testcase-with-generics/43339950#43339950) by user [taffer](https://stackoverflow.com/users/5114784/taffer).

All credit to [taffer ](https://stackoverflow.com/users/5114784/taffer) for the code.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestCaseGenericAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseGenericAttribute.cs
@@ -40,6 +40,15 @@ namespace NUnit.Framework.Attributes
 
         /// <summary>
         /// Construct a TestCaseGenericAttribute with a list of arguments.
+        /// </summary>
+        /// <param name="arg"></param>
+        public TestCaseGenericAttribute(object arg)
+            : base(arg)
+        {
+        }
+
+        /// <summary>
+        /// Construct a TestCaseGenericAttribute with a list of arguments.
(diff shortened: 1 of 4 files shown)


rprouse:
I would really like to see this feature in NUnit, but...

**First** and foremost, we cannot take code that you did not write originally. Stack Overflow posts are not under MIT license but under Creative Commons with Attribution. We would need to attribute the changes both in the code and in our releases. See the [StackOverflow license info](https://stackoverflow.com/help/licensing). If you could get in touch with taffer on Stack Overflow and get him to give his permission to use his code under MIT license, then I would reconsider. I would need him to acknowledge that here though.

**Second** I would prefer to see the functionality incorporated into the existing `TestCase` and `TestCaseSource` attributes rather than adding new attributes. I think that could be done fairly easily by checking if the test method is generic.

**Third** Your tests test the happy path, but they do not prove that new tests were discovered correctly. Notice how we add tests like this to the `TestData` project then use that in our tests to count the number of tests that were created.

I really appreciate the effort that you put into this, but unfortunately we can't take the PR as is. If you want to rewrite from scratch in the existing attributes, we would be happy to accept that. Thanks.

PR #3187, opened 2019-03-11
Don't truncate individual arguments unless specified

### LINKED ISSUE #3181: Template Based Test Naming - Incorrect truncation for individual arguments
[Thought this could be a documentation issue - apparently not](https://github.com/nunit/docs/issues/300)

Apparently individual arguments should not truncate when not specifying a truncation length, but they do.  There is no test that covers this.

[Line to be changed](https://github.com/nunit/nunit/blob/8c8cc6c8ff8b24002a94a9457c4a10a4db9f26d2/src/NUnitFramework/framework/Internal/TestNameGenerator.cs#L157)
Should be 
`fragments.Add(new ArgumentFragment(index, 0));`

The additional test is simple to add but to get the tests to run require some effort due to missing .Net frameworks/ .NET Core as well as an msbuild error.  

@CharliePoole Please can you have a look

tonyhallett:
The new test, that was missed before, is in the same format as the others. It is probably better to just have {0} instead of {m}({0}).  I have not updated the copyright as the change seems trivial. 

Closes #3181

### DIFF
--- a/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
+++ b/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
@@ -154,7 +154,7 @@ namespace NUnit.Framework.Internal
                     case "{8}":
                     case "{9}":
                         int index = token[1] - '0';
-                        fragments.Add(new ArgumentFragment(index, 40));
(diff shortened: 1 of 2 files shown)


jnm2:
This is very much appreciated, @tonyhallett! Thank you!

PR #3190, opened 2019-03-16
Timeoutattribute on netcore

### LINKED ISSUE #1638: TimeoutAttribute not available when targeting netcoreapp framework

mikkelbu:
Continuation of #3027.

Fixes #1638

### DIFF
--- a/src/NUnitFramework/tests/Attributes/TimeoutTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TimeoutTests.cs
@@ -187,7 +187,7 @@ namespace NUnit.Framework.Attributes
         [Timeout(50)]
         public void TestTimeoutAndReportsTimeoutFailure()
         {
-            while (true) { }
+            Thread.Sleep(Timeout.Infinite);
         }
 
         [Test, Timeout(100)]
--- a/src/NUnitFramework/tests/Attributes/TimeoutTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TimeoutTests.cs
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2012 Charlie Poole, Rob Prouse
+// Copyright (c) 2012-2018 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
@@ -21,7 +21,6 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-#if PLATFORM_DETECTION && THREAD_ABORT
 using System;
 using System.Linq;
 using System.Threading;
@@ -36,6 +35,7 @@ namespace NUnit.Framework.Attributes
     [NonParallelizable]
     public class TimeoutTests : ThreadingTests
     {
+#if PLATFORM_DETECTION && THREAD_ABORT
         [Test, Timeout(500)]
         public void TestWithTimeoutRunsOnSameThread()
         {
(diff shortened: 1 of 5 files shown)


mikkelbu:
The only thing left is the comment in https://github.com/nunit/nunit/pull/3027#issuecomment-426842718.

jnm2:
@mikkelbu Ah, because Task.Run causes the execution context to be [captured](https://github.com/dotnet/corefx/blob/cd45097d0066c936676ce1f35958e715dc250a44/src/Common/src/CoreLib/System/Threading/Tasks/TaskContinuation.cs#L533-L551) and [restored](https://github.com/dotnet/corefx/blob/cd45097d0066c936676ce1f35958e715dc250a44/src/Common/src/CoreLib/System/Threading/Tasks/Task.cs#L2382-L2400).

CurrentCulture is [internally implemented as an AsyncLocal](https://github.com/dotnet/corefx/blob/cd45097d0066c936676ce1f35958e715dc250a44/src/Common/src/CoreLib/System/Globalization/CultureInfo.cs#L124-L135) and AsyncLocals are one of the main things contained in the execution context. (On .NET Core, ExecutionContext's [instance state](https://github.com/dotnet/corefx/blob/cd45097d0066c936676ce1f35958e715dc250a44/src/Common/src/CoreLib/System/Threading/ExecutionContext.cs#L30-L33) is pretty much nothing but managing AsyncLocals.)

jnm2:
@mikkelbu ^ That's why the `TestWithTimeoutCurrentContextIsNotAnAdhocContext` is passing too, because `TestExecutionContext.CurrentContext` is [implemented using an AsyncLocal](https://github.com/nunit/nunit/blob/b83f789461eb2c192b22ff3b660df66808cf1c5d/src/NUnitFramework/framework/Internal/TestExecutionContext.cs#L148) except in our .NET Framework builds, and I think logical call context data is also managed within the execution context on .NET Framework.

I forgot all this in @DaiMichael's PR. I'm sorry, I hope I haven't caused you to spend too much time trying to get it to fail. The tests are really important to me in case we change from Task.Run part. Thanks for getting this done!

src/NUnitFramework/tests/Attributes/TimeoutTests.cs
```diff
@@ -167,6 +181,49 @@ public void TimeoutWithMessagePumpShouldAbort()
             Assert.That(result.ResultState, Is.EqualTo(ResultState.Failure));
             Assert.That(result.Message, Is.EqualTo("Test exceeded Timeout value of 500ms"));
         }
+#endif
+
+#if !THREAD_ABORT
+        [Timeout(50)]
+        public void TestTimeoutAndReportsTimeoutFailure()
+        {
+            while (true) { }
```
jnm2: This is a busy loop that will eat CPU for the entire rest of the run of our test suite since we can't abort it. Would you change this to `Thread.Sleep(Timeout.Infinite);` please?

src/NUnitFramework/tests/Attributes/TimeoutTests.cs
```diff
@@ -167,6 +181,49 @@ public void TimeoutWithMessagePumpShouldAbort()
...
+        public void TestTimeoutDoesNotStopCompletion()
+        {
+            Thread.Sleep(20);
+            Assert.True(true);
+        }
+
+        [Timeout(100)]
+        public void TestTimeoutWhichThrowsTestException()
+        {
+            throw new ArgumentException($"{nameof(ArgumentException)} was thrown.");
```
jnm2: What should happen if an exception is thrown in the thread after we've stopped waiting for it due to timeout? I feel like we should add that exception message as an additional error to the test's result even though we've already reported the timeout failure. We should then re-report the test status to the engine so that the exception message is reflected in Test Explorer etc right away. Would this make sense to do?

Maybe something for a future PR.

mikkelbu: I think we should leave this for a future PR. I don't think that this will occur very often, and I think this could be difficult to implement, so I would rather have this PR shipped (and also version 3.12 😄).

src/NUnitFramework/tests/Attributes/TimeoutTests.cs
mikkelbu: Done

jnm2:
@mikkelbu and @DaiMichael, thank you very much!

Issue #3157, opened 2019-02-07
Test Explorer will not show our tests unless the test project stops targeting net35


jnm2:
Affects Visual Studio 2017 Update 9 and Visual Studio 2019 previews.

If `net35` is present at all, even if it's not the first item, no tests show up in Test Explorer.

https://github.com/nunit/nunit/blob/b5c1d549938476ff06f118557b29778b36593549/src/NUnitFramework/tests/nunit.framework.tests.csproj#L4

https://github.com/nunit/nunit/blob/b5c1d549938476ff06f118557b29778b36593549/src/NUnitFramework/nunitlite.tests/nunitlite.tests.csproj#L4

The Tests output is this:

> Microsoft.VisualStudio.TestPlatform.ObjectModel.TestPlatformException: Framework35 is not supported. For projects targeting .Net Framework 3.5, please use Framework40 to run tests in CLR 4.0 "compatibility mode".

```
   at Microsoft.VisualStudio.TestPlatform.CommandLine.TestPlatformHelpers.TestRequestManager.UpdateRunSettingsIfRequired(String runsettingsXml, List`1 sources, String& updatedRunSettingsXml)
   at Microsoft.VisualStudio.TestPlatform.CommandLine.TestPlatformHelpers.TestRequestManager.RunTests(TestRunRequestPayload testRunRequestPayload, ITestHostLauncher testHostLauncher, ITestRunEventsRegistrar testRunEventsRegistrar, ProtocolConfig protocolConfig)
   at Microsoft.VisualStudio.TestPlatform.Client.DesignMode.DesignModeClient.<>c__DisplayClass19_0.<StartTestRun>b__0()
```

Since Test Explorer is the only remotely convenient way to debug a single test (unless I install ReSharper into this instance of VS), this is extremely frustrating.

I don't think there's anything we can do about it besides report it to Microsoft. Visual Studio should be applying `Framework40` itself; it's not us who are passing `Framework35` to VSTest.

@OsirisTerje, is this something we should tag a VSTest team member for?

jnm2:
@OsirisTerje ping.

icnocop:
Is https://github.com/Microsoft/vstest/issues/1896 related?

mikkelbu:
@jnm2 Accordingly to Microsoft/vstest#1896: "This issue has been resolved in VS 2017 v15.9.10 (#1945).", and it again works on my machine.

jnm2:
[...]

PR #3205, opened 2019-04-09
Added TestCaseData.Combinatorial() method

jerryk414:
When generating test cases, it is extremely convenient to be able to use the Combinatorial strategy for parameterized tests.

Unfortunately, this concept was not available when using TestCaseSource / TestCaseData.  This change is meant to make this functionality available in that situation.

I originally rolled out my own Combinatorial method, not thinking about how the logic already exists elsewhere.  I eventually modified that logic to be re-usable, and just re-implemented it for TestCaseSource / TestCaseData.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Builders/CombinatorialStrategy.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/CombinatorialStrategy.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2008 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -40,7 +40,16 @@ namespace NUnit.Framework.Internal.Builders
         /// <returns>The test cases.</returns>
         public IEnumerable<ITestCaseData> GetTestCases(IEnumerable[] sources)
         {
-            List<ITestCaseData> testCases = new List<ITestCaseData>();
+            return GetCombinations<ITestCaseData>(sources, (o) => new TestCaseParameters(o));
+        }
+
+        /// <summary>
+        /// Returns an <see cref="IEnumerable{T}"/> representing every single possible combination of the given parameters
+        /// </summary>
+        /// <returns>Every combination of parameters</returns>
+        public static IEnumerable<T> GetCombinations<T>(IEnumerable[] sources, Func<object[], T> handle)
+        {
+            List<T> testCases = new List<T>();
             IEnumerator[] enumerators = new IEnumerator[sources.Length];
             int index = -1;
 
@@ -58,7 +67,7 @@ namespace NUnit.Framework.Internal.Builders
                 for (int i = 0; i < sources.Length; i++)
                     testdata[i] = enumerators[i].Current;
 
-                TestCaseParameters parms = new TestCaseParameters(testdata);
+                T parms = handle(testdata);
                 testCases.Add(parms);
 
                 index = sources.Length;
--- a/src/NUnitFramework/framework/TestCaseData.cs
+++ b/src/NUnitFramework/framework/TestCaseData.cs
@@ -22,6 +22,8 @@
 // ***********************************************************************
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
 
@@ -198,6 +200,92 @@ namespace NUnit.Framework
             this.RunState = RunState.Ignored;
             this.Properties.Set(PropertyNames.SkipReason, reason);
             return this;
+        }
+
+        #endregion
+        
+        #region Methods
+
+        /// <summary>
+        /// Returns a list of <see cref="TestCaseData"/> objects representing all possible combinations of the given parameters.
+        /// </summary>
+        /// <param name="values">Each array in this object array is to represent a parameter for the method being tested</param>
+        /// <returns>An <see cref="IEnumerable{T}"/> of <see cref="TestCaseData"/> objects that represents all possible combinations for the given parameters</returns>
+        /// <remarks>
+        /// This method will generate a collection of <see cref="TestCaseData"/> instances where the collection represents every possible combination that could be made
+        /// with the given parameters.  That is to say, with the following code:
+        /// <code>
+        /// return TestCaseData.Combinatorial(
+        ///             new[] { true, false }.Cast&lt;object&gt;(),
+        ///             new[] { 0, 1, 10 }.Cast&lt;object&gt;());
+        /// </code>
+        /// 
+        /// The following test data will be returned:
+        /// <list type="bullet">
+        /// <item><description>TestCaseData(true, 0)</description></item>
+        /// <item><description>TestCaseData(true, 1)</description></item>
+        /// <item><description>TestCaseData(true, 10)</description></item>
+        /// <item><description>TestCaseData(false, 0)</description></item>
+        /// <item><description>TestCaseData(false, 1)</description></item>
+        /// <item><description>TestCaseData(false, 10)</description></item>
+        /// </list>
+        /// </remarks>
+        /// <example>
+        /// <code>
+        /// public class TestClass
+        /// {
+        ///     #region TestCaseSource
+        /// 
+        ///     private static IEnumerable ExampleTest_TestCases
+        ///     {
+        ///         return TestCaseData.Combinatorial(
+        ///             new[] { true, false }.Cast&lt;object&gt;(),
+        ///             new[] { 0, 1, 10 }.Cast&lt;object&gt;());
+        ///     }
+        /// 
+        ///     #endregion
+        ///     [Test, TestCaseSource(typeof(TestClass), nameof(ExampleTest_TestCases))]
+        ///     public void ExampleTest(bool condition, int input)
+        ///     {
+        ///          // Execute test
+        ///     }
+        /// }
+        /// </code>
+        /// </example>
+        public static IEnumerable<TestCaseData> Combinatorial(params IEnumerable<object>[] values)
+        {
+            // Let's first figure out how many combinations we're going to be creating (the count of each list multiplied together)
+            int combinations = values.Aggregate(1, (x, y) => x * y.Count());
+
+            // If there are no combinations to be made, just return an empty list
+            if (combinations == 0)
+            {
+                return Enumerable.Empty<TestCaseData>();
+            }
+
+            // Iterate through every list of objects and do the magic
+            List<List<object>> result = Enumerable.Range(0, combinations).Select(i => new List<object>()).ToList();
+            int repetition = combinations;
+            foreach (IEnumerable<object> value in values)
+            {
+                int index = 0;
+
+                repetition = repetition / value.Count();
+
+                while (index < combinations)
+                {
+                    foreach (object o in value)
+                    {
+                        for (int i = 0; i < repetition; i++)
+                        {
+                            result[index].Add(o);
+                            index++;
+                        }
+                    }
+                }
+            }
+
+            return result.Select(i => new TestCaseData(i.ToArray()));
         }
 
         #endregion
--- a/src/NUnitFramework/framework/TestCaseData.cs
+++ b/src/NUnitFramework/framework/TestCaseData.cs
@@ -22,10 +22,12 @@
 // ***********************************************************************
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
+using NUnit.Framework.Internal.Builders;
 
 namespace NUnit.Framework
 {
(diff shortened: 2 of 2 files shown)


CharliePoole:
This PR seems to provide existing functionality using a different syntax. Having different ways to do the same thing can be an advantage but can also sometimes be confusing to users.

In the current code, we generally use TestCase/TestCaseSource when the specific combinations of arguments are desired and Value/ValueSource when we want the combinations to be generated. This allows alternate algorithms for combining the values, including possible custom strategies.

Can you say a little about why this might be a better approach than the existing one or a useful supplement to it? One way to explain might be to think of how we could provide guidance to users on which feature to use when.

jerryk414:
@CharliePoole 
So ValueSource is convenient and can give you this same functionality, in a way, but it doesn't give you the granularity and control that using TestCaseSource does.

So here are some examples of use cases:

```csharp
public class TestClass
{
#region TestCaseSource

protected static IEnumerable TESTCASESOURCENAME

     [Test, TestCaseSource(typeof(TestClass), nameof(TestMethod_TestCases))]
     public void TestMethod(int input, int condition)
     {
          // Some test code
     }
}
```

jerryk414:
@CharliePoole 
So ValueSource is convenient and can give you this same functionality, in a way, but it doesn't give you the granularity and control that using TestCaseSource does.

So in our specific implementation of this that we've been using for a couple of years now, we actually derive a class from TestCaseData and add some additional logic to that class. For example, we have the following:

1.  Code to automatically append a unique ID to the end of each test name so that, in instances where we use TestCaseSource and pass in complex objects, the VS Test Explorer does not run into any issues with ambiguous named tests.
2.  Code to automatically strip out the unique identifiers added onto Moq objects, so if we pass along Moq's as test parameters, the tests can be executed by name.  Moq'd objects are generated on the fly and named with a unique name each time - so if you select a test by name in the VS Test Explorer that contains a Moq as a parameter, the name will change when you try to actually run the test - the Test Explorer then complains about not being able to find the test you want to run.
3.  Code to expand IEnumerable test parameters so it is easier to view the information in the Test Explorer.  By default, when test cases contain test case data of an IEnumerable type, it will display the collection type, instead of the contents of the collection.  We've added code to recursively expand the contents so it's easier to tell which test we are running in the Test Explorer.
4.  Code to handle issues with the maximum name length.  VS Test Explorer's maximum name length is 449 characters - we added code in our implementation to handle this situation.
5.  Code to make it easier to pull back the original arguments off of the TestCaseData object.

_The specifics of what's above doesn't matter_.  The point is, we can't be the only ones who do this.

So here are some examples of use cases:

```csharp
namespace TestNameSpace
{
    public class TestClass
    {
        #region TestCaseSource

        protected static IEnumerable Example1_TestCases
        {
            get
            {
                foreach (TestCaseData data in TestCaseData.Combinatorial(
                    new[] { 0, 1, 10 },
                    new[] { true, false },
                    new[] { true, false }))
                {
                    yield return data;
                }

                // Include a test case independent of the combinatorial logic
                yield return new TestCaseData(-500, true, true);
            }
        }

        #endregion     
        [Test, TestCaseSource(typeof(TestClass), nameof(Example1_TestCases))]
        public void Example1(long input, bool condition1, bool condition2)
        {
            // Some test code
        }

        #region TestCaseSource

        protected static IEnumerable Example2_TestCases
        {
            get
            {
                foreach (TestCaseData data in TestCaseData.Combinatorial(
                    new[] { 0, 1, 10 },
                    new[] { true, false },
                    new[] { true, false }))
                {
                    // Ignore an invalid test case generated via the combinatorial logic
                    if (data.OriginalArguments[0].Equals(0) && 
                        data.OriginalArguments[1].Equals(true) && 
                        data.OriginalArguments[2].Equals(true))
                        continue;

                    yield return data;
                }
            }
        }

        #endregion     
        [Test, TestCaseSource(typeof(TestClass), nameof(Example2_TestCases))]
        public void Example2(long input, bool condition1, bool condition2)
        {
            // Some test code
        }
        
        #region TestCaseSource

        protected static IEnumerable Example3_TestCases
        {
            get
            {
                // Include one set of parameters to combine
                foreach (TestCaseData data in TestCaseData.Combinatorial(
                    new[] { 0, 1, 10 },
                    new[] { true, false },
                    new[] { true, false }))
                {
                    yield return data;
                }
                
                // And include a second set of parameters to combine
                foreach (TestCaseData data in TestCaseData.Combinatorial(
                    new[] { -1, -10 },
                    new[] { false },
                    new[] { true, false }))
                {
                    yield return data;
                }
            }
        }

        #endregion     
        [Test, TestCaseSource(typeof(TestClass), nameof(Example2_TestCases))]
        public void Example3(long input, bool condition1, bool condition2)
        {
            // Some test code
        }

        #region TestCaseSource

        protected static IEnumerable Example4_TestCases
        {
            get
            {
                foreach (TestCaseData data in TestCaseData.Combinatorial(
                    new[] { 0, 1, 10 },
                    new[] { true, false },
                    new[] { true, false }))
                {
                    // Use the TestCaseData methods
                    yield return data
                        .Returns(500)
                        .SetName($"SomeUniqueName({ string.Join(", ", data.OriginalArguments.Select(i => i.ToString())) } ");
                }
            }
        }

        #endregion     
        [Test, TestCaseSource(typeof(TestClass), nameof(Example2_TestCases))]
        public int Example4(long input, bool condition1, bool condition2)
        {
            // Some test code
        }
    }
}
```

CharliePoole:
Interesting... Some comments on each of the examples...

Example 1... 

"Standard" way to do this is to use Value or ValueSource together with an additional TestCase attribute to add the special case. 

Example 2...

This can't be done automatically with the current features. That is, there is no way to not generate a certain combination. You could use an Assume.That to reject the case but that's really not quite the same thing.

Example 3...

Multiple combinatorial sets would be a new feature.

Example 4...

Yes. When you use Value or ValueSource you don't have the do things like specify an expected result or use other properties of TestCaseData.


I can see that this approach adds quite a lot of value. I have some recommended changes, but first let's see what the @nunit/framework-team  thinks about the idea.

(Normally, we'd be into code review of the PR here, but since we haven't had the "do we want this feature" discussion ahead of time, please bear with us as we catch up. We appreciate the contribution.)

jerryk414:
Not a problem.  This is something we've used internally for quite some time, and its proven to be extremely useful.

Additionally, I think exposing the method of generating combinations would be valuable in instances where a consumer is using a custom implementation of TestCaseData, like us.  That is to say, without it being exposed, we will have to either implement it again ourselves, or generate a collection of TestCaseData objects, then send those into a constructor for our custom implementation to create a collection of our custom TestCaseData objects.

CharliePoole:
That makes sense. I'd like to see some more votes for accepting the feature before getting much into the code details but IMO it would be nice if all combining strategies could be made available, not just Combinatorial.

src/NUnitFramework/framework/Internal/Builders/CombinatorialStrategy.cs
```diff
@@ -40,7 +40,16 @@ public class CombinatorialStrategy : ICombiningStrategy
...
         {
-            List<ITestCaseData> testCases = new List<ITestCaseData>();
+            return GetCombinations<ITestCaseData>(sources, (o) => new TestCaseParameters(o));
+        }
+
+        /// <summary>
+        /// Returns an <see cref="IEnumerable{T}"/> representing every single possible combination of the given parameters
+        /// </summary>
+        /// <returns>Every combination of parameters</returns>
+        public static IEnumerable<T> GetCombinations<T>(IEnumerable[] sources, Func<object[], T> handle)
```
jnm2: `handle` would be clearer to me if it was called `resultSelector` or something similar.

src/NUnitFramework/framework/Internal/Builders/CombinatorialStrategy.cs
```diff
@@ -40,7 +40,16 @@ public class CombinatorialStrategy : ICombiningStrategy
...
         public IEnumerable<ITestCaseData> GetTestCases(IEnumerable[] sources)
         {
-            List<ITestCaseData> testCases = new List<ITestCaseData>();
+            return GetCombinations<ITestCaseData>(sources, (o) => new TestCaseParameters(o));
+        }
+
+        /// <summary>
+        /// Returns an <see cref="IEnumerable{T}"/> representing every single possible combination of the given parameters
+        /// </summary>
+        /// <returns>Every combination of parameters</returns>
```
jnm2: Leaving this method internal is an option I'd consider.
If you keep it public, I'd like to consistently use full sentences (including end marks) for XML documentation. Also, we recommend not adding `<returns>` since it isn't shown in intellisense and our XML docs aren't reused for anything else.

jerryk414: The argument for leaving it public would be that, in cases where TestCaseData is extended upon in a derived class, you will not be able to use TestCaseData.Combinatorial() for that derived class since it would be returning TestCaseData instances and not the instance of the derived class.  So, in order to get Combinatorial() behavior in a derived TestCaseData instance, you would have to add your own static method that calls CombinatorialStrategy.GetCombinations().

I guess an alternative to solving this would be to expose an overload for TestCaseData.Combinatorial() as a protected static method where it would be something like this:

```csharp
public class TestCaseData : TestCaseParameters
{
...
     protected static IEnumerable<T> Combinatorial(IEnumerable[] values, Func<object[], T> handle)
     {
          return CombinatorialStrategy.GetCombinations<T>(values, handle);
     }
...
}
```
And then in the derived class, you can use the ```new``` keyword to hide away the old Combinatorial method and expose the new one, like so:
```csharp
public class DerivedTestCaseData : TestCaseData
{
...
     public static new IEnumerable<DerivedTestCaseData> Combinatorial(IEnumerable[] values)
     {
          return TestCaseData.Combinatorial<DerivedTestCaseData>(values, (o) => new DerivedTestCaseData(o));
     }
...
}
```

If you'd prefer, i can go ahead and make those changes, please let me know though.

src/NUnitFramework/framework/Internal/Builders/CombinatorialStrategy.cs
jerryk414: I personally do not have any strong opinions on this, if you want ```resultSelector``` then that is fine by me.

src/NUnitFramework/framework/Internal/Builders/CombinatorialStrategy.cs
jerryk414: I agree that there should be some unit tests.  I ran into some issues even trying to run existing unit tests in the solution, though.

I can try again once i get a response on the other items of note here.

jnm2: I like it public then. That makes sense. If it's a public API though, we don't want it in the Internal namespace. To me, it also doesn't seem very discoverable where it is.

When we decide how to expose this API, I'd like to do it in a consistent way that achieves the goal @CharliePoole mentioned:

> IMO it would be nice if all combining strategies could be made available, not just Combinatorial.

If we make it public in this pull request, it will probably take longer to merge since @nunit/framework-team needs to decide what the full picture will look like. Are you okay with that?

jnm2:
@jerryk414 

> I agree that there should be some unit tests. I ran into some issues even trying to run existing unit tests in the solution, though. IIRC, it had something to do with .NET 3.5 causing issues - i can't remember any specifics though.

This sounds like https://github.com/nunit/nunit/issues/3157 which was fixed. We want Test Explorer to be a way to run our tests, so let us know if it's still not working. Whether or not you run tests in the IDE for fast inner loop development, we recommend running `build -t test` or `build -t test45` before committing, just so you see the failures and successes you expect.

jnm2:
@nunit/framework-team and @jerryk414: What do you think, should we plan to expose pairwise and sequential generators as well as combinatorial?

For TestCaseData and TestFixtureData, do we want folks to use APIs like `TestCaseData.Combinatorial(...)`, `TestCaseData.Pairwise(...)`, `TestFixtureData.Combinatorial(...)`, etc?

We could tackle custom classes later. For a user's `CustomTestCaseData` class, I'm thinking a protected methods `TestParameters.Combinatorial`, `TestParameters.Pairwise`, etc would be the most easily discovered. The user would expose Combinatorial and Pairwise for their `CustomTestCaseData` class this way:

```cs
public class CustomTestCaseData : TestCaseData
{
    // ...

    public static IEnumerable<CustomTestCaseData> Combinatorial(params IEnumerable[] valuesByParameter)
    {
        return Combinatorial(valuesByParameter, args => new CustomTestCaseData(args));
    }

    public static IEnumerable<CustomTestCaseData> Pairwise(params IEnumerable[] valuesByParameter)
    {
        return Pairwise(valuesByParameter, args => new CustomTestCaseData(args));
    }

    public static IEnumerable<CustomTestCaseData> Sequential(params IEnumerable[] valuesByParameter)
    {
        return Sequential(valuesByParameter, args => new CustomTestCaseData(args));
    }
}
```

src/NUnitFramework/framework/Internal/Builders/CombinatorialStrategy.cs
jnm2: @jerryk414 What do you think of the name `IEnumerable[] valuesByParameter`?

CharliePoole: I don't see the point of introducing a static method here. Like all statics, it ain't pretty, and it doesn't seem to do anything different from `new CombinatorialStrategy().GetTestCases()`. Am I missing something?

I see a great advantage as well in sticking with the use of the existing internal interface. TestCaseData could have an internal method that takes an `ICombiningStrategy`, which would be used in turn by the public `Combinatorial` method.

I also dislike changing the internal structure of just one of our combining strategies, which I believe currently maintain the same structure.

rprouse:
This has gone dormant and nobody has picked it up and completed it, so I'm closing.

PR #3206, opened 2019-04-16
Fix for IRepeatTest attributes bypass always the IMethodInfo interface

qlikTERror:
When looking for IRepeatTest instead of doing it through the IMethodInfo (and by inheritage IReflectionInfo) interface it, it was done directly through the MethodInfo.

This was not allowing a full customization of the MethodWrapper class to add custom attributes.

The default implemementation of the GetCustomAttributes directly calls MethodInfo.GetAttributes and should not effect current usage.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
@@ -114,7 +114,7 @@ namespace NUnit.Framework.Internal.Builders
         /// <returns>True if the method signature is valid, false if not</returns>
         private static bool CheckTestMethodAttributes(TestMethod testMethod)
         {
-            if (testMethod.Method.MethodInfo.GetAttributes<IRepeatTest>(true).Length > 1)
+            if (testMethod.Method.GetCustomAttributes<IRepeatTest>(true).Length > 1)
                 return MarkAsNotRunnable(testMethod, "Multiple attributes that repeat a test may cause issues.");
 
             return true;
--- a/src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
@@ -152,7 +152,7 @@ namespace NUnit.Framework.Internal.Execution
                     command = new TimeoutCommand(command, timeout);
 
                 // Add wrappers for repeatable tests after timeout so the timeout is reset on each repeat
-                foreach (var repeatableAttribute in method.MethodInfo.GetAttributes<IRepeatTest>(true))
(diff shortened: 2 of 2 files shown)


qlikTERror:
Unfortunately yes, @mikkelbu. I was in need of a kind of RetryOnExceptionAttribute, and I didn't want to modify every single test method of our integration tests to add it. When I found that IRepeatTest was not applicable to Class or Assembly, I decided digged the code and decided to use ITestAction to Wrap the MethodWrapper of each ITest that was a TestMethod. The wrapper would reimplement the IReflectionInfo.GetCustomAttributes to add the RetryOnExceptionAttribute even if it was not on the original Test method. And here came out the problem... IRepeatTest were read directly with the MethodInfo, instead of through IMethodInfo and the MethodWrapper.

qlikTERror:
@mikkelbu do you know if the CLA can be signed by one of our legal representative?

CharliePoole:
AFAIK the CLA has to be signed by each individual who pushes a change through a PR. I guess it's unlikely that any of your lawyers will be doing that. :smile: 

If you did the change on your own time, then it has nothing to do with your company. You can just sign it yourself.

If you did the code on company time, then they may have rights to it. So we ask that you sign a document that says, among other things, that you have permission to contribute the code from your company. Obviously, you have to first make sure you have that permission, which usually comes from your own management but may also involve legal. If you know you have that permission, then the CLA becomes not such a big deal.

That's my opinion, anyway.

PS: FWIW, some earlier versions of NUnit (around 2.0 through 2.2) were completely created by guys working for a single company. They made sure that they worked on their own time and that the company was not claiming any rights to it. That's why you can use it today!

CharliePoole:
Noting we still don't have a CLA on this...

qlikTERror:
@CharliePoole I signed the CLA after fetching the requested permissions from my empoloyer.

@jnm2 I really don't know where to begin to write a tests for them. I think I will require some directions and helps.

Apart this I used the IMethodInfo exception to create a working cross-cutting attribute. Wrapping the original MethodWrapper with mine allowed me to return more Attributes than the one specified on the Test method. To do this I use the ITestAction interface to use the BeforeTest (during the TestFixture) to wrap all the methods that with my custom IMethodInfo that will return my RetryOnExceptionAttribute too when calling for GetCustomAttributes (or IsDefined). Moreover I am recording the Exception thrown in my IMethodInfo.Invoke.

And like this I solved the problem of not having cross-cutting attributes :D

qlikTERror:
@jnm2 any way you can instruct me how to write the test you require for this?

jnm2:
@qlikTERror Thank you very much for this contribution!

qlikTERror:
Glad to contribute :D

jnm2:
@qlikTERror If you'd like to use it sooner than our next release, you can look for the latest package prerelease at https://www.myget.org/F/nunit/api/v3/index.json.

PR #3211, opened 2019-04-21
Fix logging

ChrisMaddock:
See https://github.com/nunit/nunit-console/pull/612. There was only this one case to fix on the framework. 🙂

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
@@ -185,13 +185,13 @@ namespace NUnit.Framework.Internal.Execution
                     }
                     catch (Exception ex)
                     {
-                        log.Error( "Exception in event handler\r\n {0}", ex );
+                        log.Error("Exception in event handler {0}", ExceptionHelper.BuildStackTrace(ex));
                     }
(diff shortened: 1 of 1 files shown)


src/NUnitFramework/framework/Internal/Execution/EventPump.cs
```diff
@@ -191,7 +191,7 @@ private void PumpThreadProc()
             }
             catch (Exception ex)
             {
-                log.Error( "Exception in pump thread", ex );
+                log.Error("Exception in pump thread {0}", ex);
```
mikkelbu: Any reason you are not using `ExceptionHelper` in this repo (similar to the changes in the console-repo)?

And should we also make line 188 consistent with this line and remove `\r\n` ?

mikkelbu:
I've created #3212 to solve the problem with the Linux build on Travis.

src/NUnitFramework/framework/Internal/Execution/EventPump.cs
ChrisMaddock: Done! 🙂

ChrisMaddock:
Thanks Mikkel, changes made. 😄

PR #3212, opened 2019-04-21
Bump Ubuntu version

mikkelbu:
The Linux build has started to fail in a similar manner to nunit/nunit-console#611, so we bump the Ubuntu version.


mikkelbu:
See e.g. https://travis-ci.org/nunit/nunit/builds/522781717 where I've triggered a build on master.

PR #3214, opened 2019-04-23
Supporting Fields in Is.Ordered.By

### LINKED ISSUE #3082: Is.Ordered.By
Is.Ordered.By only works with Properties but it would be nice if it also works with fields.

`Assert.That(array, Is.Ordered.By("PropertyName"));`

FeatureRequest:
`Assert.That(array, Is.Ordered.By("FieldName"));`

Perhaps it would be nice to can use internal Properties / Fields also.

heshamamin:
This PR resolves #3082 .
I've added support for Fields in Is.Ordered.By, however I think internal properties and fields need more discussion.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
@@ -223,15 +223,9 @@ namespace NUnit.Framework.Constraints
                         {
                             string propertyName = step.PropertyName;
 
-                            PropertyInfo previousProp = previous.GetType().GetProperty(propertyName);
-                            if (previousProp == null)
-                                throw new ArgumentException($"Property {propertyName} not found at index {_breakingIndex - 1}", nameof(actual));
-                            var previousValue = previousProp.GetValue(previous, null);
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/framework/Constraints/CollectionOrderedConstraint.cs
```diff
@@ -297,6 +291,26 @@ private void CreateNextStep(string propertyName)
...
+                FieldInfo field = item.GetType().GetField(propertyName);
+                if (field == null)
+                {
+                    throw new ArgumentException($"No property or field with name {propertyName} was found at index {index}", nameof(actual));
+                }
+                value = field.GetValue(item);
+            }
+            else
+            {
+                value = property.GetValue(item, null);
```
jnm2: Can you flip this around so that it's nearer to the `GetProperty` and `property == null` lines, to keep related things close together? Same for the `field.GetValue` line?

CharliePoole: I'd reverse the test and do an early return as well, eliminating a level of nesting.

CharliePoole:
@jnm2 Please make sure you still like it and merge. :smile:

jnm2:
Excellent. Thanks for this, @heshamamin!

PR #3216, opened 2019-04-24
Restore full assembly paths in test XML

### LINKED ISSUE #3085: XML Test-Suite Assembly does not contain DLL path anymore
In XML Result the attribute fullname of test-suite with the type "Assembly", does not contain an absolute filepath of the dll anymore.
NUnit-Framework 3.10.1:
```
<test-suite type='Assembly' id='0-1063' name='cgw_kammer.dll' fullname='C:\Users\thejoklla\source\repos\cgw_kammer\cgw_kammer\bin\Debug\cgw_kammer.dll' 
```
NUnit-Framework 3.11.0:
```
<test-suite type="Assembly" id="0-1063" name="cgw_kammer.dll" fullname="cgw_kammer.dll"
```
Tested with NUnit.Engine.Api 3.9.0

jnm2:
Fixes #3085.

/cc @CharliePoole


jnm2:
@CharliePoole It's picked up indirectly by the SetUpFixtureTests changes I made, but I suppose we should have dedicated test names for it.

PR #3217, opened 2019-04-26
Allow static SetUpFixture classes

### LINKED ISSUE #3166: Allow static SetUpFixture classes
As long as the setup and teardown methods are static, no instance should be created and therefore `static` and `abstract` classes and classes without constructors should work, just like OneTimeTearDown with a non-[SetUpFixture] class.

```cs
[SetUpFixture]
public static class AcceptanceTestsTeardownFixture
{
    [OneTimeTearDown]
    public static void OneTimeTearDown()
    {
        AcceptanceTests.OnGlobalTeardown();
    }
}
```

It currently fails all tests with:

> OneTimeSetUp: NUnit.VisualStudio.TestAdapter.Tests.Acceptance.AcceptanceTestsTeardownFixture is an abstract class

stevenaw:
This PR adds support for `SetupFixtureAttribute` on static classes + some tests.

Fixes #3166


stevenaw:
Thanks!

PR #3218, opened 2019-04-27
Remove todos from the code base

mikkelbu:


src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -128,9 +128,6 @@ public IList<EqualityAdapter> ExternalComparers
             get { return externalComparers; }
         }
 
-        // TODO: Define some sort of FailurePoint struct or otherwise
```
mikkelbu: The `FailurePoint` class is defined below in the file.

src/NUnitFramework/framework/Internal/Commands/DelegatingTestCommand.cs
```diff
@@ -38,9 +38,9 @@ public abstract class DelegatingTestCommand : TestCommand
 #pragma warning restore IDE1006
 
         /// <summary>
-        /// TODO: Documentation needed for constructor
+        /// Initializes a new instance of the <see cref="DelegatingTestCommand"/> class.
         /// </summary>
-        /// <param name="innerCommand"></param>
+        /// <param name="innerCommand">The inner command.</param>
```
jnm2: If it was me, I'd either make this more descriptive (explaining what "inner command" means) or delete the `<param>`.

PR #3225, opened 2019-04-30
If you can’t beat ’em, join ’em

### LINKED ISSUE #3222: Our build script tests hang when run with Mono on Windows

jnm2:
Fixes #3222. It would be nice to figure out why thread aborting hasn't worked for us on Mono [in the last ten years](https://github.com/nunit/nunit/commit/fc042531ab73cfebc0c1b71e1bb26ff2bc2483dc), but I should have excluded Mono in #2023 for my test as well, since it's just a fancier version of:

https://github.com/nunit/nunit/blob/9f2d8ea7c35518ee28f1255bc21aeca6d84057b3/src/NUnitFramework/tests/Attributes/TimeoutTests.cs#L65-L76

https://github.com/nunit/nunit/blob/9f2d8ea7c35518ee28f1255bc21aeca6d84057b3/src/NUnitFramework/testdata/TimeoutFixture.cs#L49-L67

### DIFF
--- a/src/NUnitFramework/tests/Attributes/TimeoutTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TimeoutTests.cs
@@ -50,6 +50,9 @@ namespace NUnit.Framework.Attributes
         }
 
 #if PLATFORM_DETECTION && THREAD_ABORT
+        // Most recent version of Mono tested: 5.20.1
+        private const string MonoFailsToAbortThreadReason = "ThreadAbortException is never thrown on Mono";
+
         [Test, Timeout(500)]
         public void TestWithTimeoutRunsOnSameThread()
         {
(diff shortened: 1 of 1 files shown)


jnm2:
@CharliePoole I'd sure like to eventually. @rprouse is waiting for my custom awaiters PR and I'm trying to only do things that get me closer to figuring out where its unexpected thread abort is coming from.

jnm2:
> Funny thing... I just assumed you knew it had been a problem and were ultra-determined to fix it!

Guess what? I was ultra-determined to fix #3095 and in fixing it, I accidentally fixed all these tests too! It's a nice reward for a frustrating and time-consuming search. In addition to learning a few things.

PR #3234, opened 2019-05-06
AndConstraint should log additional information from failed constraint

### LINKED ISSUE #3233: AndConstraint should write additional information from failed constraint
```
Assert.That(new[]{ 1,2 }, Is.Ordered & Is.EquivalentTo(new[] { 1, 2, 3 }));
```

Expected message:
```
  Expected: collection ordered and equivalent to < 1, 2, 3 >
  But was:  < 1, 2 >
  Missing (1): < 3 >
```

Actual message:
```
  Expected: collection ordered and equivalent to < 1, 2, 3 >
  But was:  < 1, 2 >
```

Dreamescaper:
Fix #3233

### DIFF
--- a/src/NUnitFramework/framework/Constraints/AndConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/AndConstraint.cs
@@ -76,7 +76,7 @@ namespace NUnit.Framework.Constraints
             private readonly ConstraintResult rightResult;
 
             public AndConstraintResult(AndConstraint constraint, object actual, ConstraintResult leftResult, ConstraintResult rightResult)
-                : base(constraint, actual, leftResult.IsSuccess && rightResult.IsSuccess) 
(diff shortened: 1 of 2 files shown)


Dreamescaper:
@jnm2
Will as tests a bit later.
Not sure if that's applicable for NotConstraint and OrConstraint though.

NotConstraint fails when inner constraint passes, and they do not write additional info when passed.

With OrConstraint situation is different. Both left and right constraint have to be failed. What should be written then? Additional info from both of them, left then right?

CharliePoole:
There's no reason to add similar but different work to treasure issue. I'd say create s new issue no matter who will do the work.

jnm2:
Sounds good, let's keep it separate. https://github.com/nunit/nunit/issues/3235

Dreamescaper:
Added tests

PR #3238, opened 2019-05-13
Added prerelease nuget.config boilerplate for easy copy/paste

jnm2:
I keep looking for the last place I wrote this by hand so that I can reuse it, so if you don't think this explanation is too much clutter for the readme...

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,18 @@ NUnit is a unit-testing framework for all .NET languages. Initially ported from 
 
 ## Downloads ##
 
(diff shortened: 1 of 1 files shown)


jnm2:
That's fair. I haven't used the NuGet Package Manager Console in VS much yet. I'll try sharing the link when suggesting our prerelease feed instead.

mikkelbu:
The Package Manager Console was just an example :). I think I use .NET CLI the most, e.g. `dotnet add package NUnit --version 3.12.0-dev-06237 --source https://www.myget.org/F/nunit/api/v3/index.json`.

image

PR #3241, opened 2019-05-14
Fix Randomizer.NextDecimal bias

### LINKED ISSUE #3228: Modulo bias is present in Randomizer.NextDecimal(decimal)
Using the modulo operator means that some consecutive amount of the lowest numbers in the range are twice as likely as the remaining numbers in the range:

https://github.com/nunit/nunit/blob/8a6245b90276fcab21a11bf65b06bbcdd562d67b/src/NUnitFramework/framework/Internal/Randomizer.cs#L575-L578

We should be doing something more along the lines of this method further down:

https://github.com/nunit/nunit/blob/8a6245b90276fcab21a11bf65b06bbcdd562d67b/src/NUnitFramework/framework/Internal/Randomizer.cs#L602-L609

My preference would be to initialize `low`, `mid`, and `high` using uniform random bits, then mask out the most significant bits which are guaranteed to put the number out of range, then return the resulting value if it is within the range and otherwise start back at the beginning.

The chance that another iteration will be necessary will always be less than 50% so long as the high bits are masked out. The chance that two iterations will be required is > 25%, three is > 12.5%, and so on.

It should be possible to write a test that shows the current bias and only passes once the bias is removed.

jnm2:
Fixes #3228

### DIFF
--- a/src/NUnitFramework/framework/Internal/Randomizer.cs
+++ b/src/NUnitFramework/framework/Internal/Randomizer.cs
@@ -592,17 +592,17 @@ namespace NUnit.Framework.Internal
                     {
                         low = RawInt32();
                         mid = RawInt32();
-                        high = RawInt32() & (int)MaskOutBitsGuaranteedToExceedMaximum(maximum: (uint)parts[2] - 1);
+                        high = RawInt32() & (int)MaskToRemoveBitsGuaranteedToExceedMaximum(maximum: (uint)parts[2] - 1);
                     }
                     else if (parts[1] != 0)
                     {
                         low = RawInt32();
-                        mid = RawInt32() & (int)MaskOutBitsGuaranteedToExceedMaximum(maximum: (uint)parts[1] - 1);
+                        mid = RawInt32() & (int)MaskToRemoveBitsGuaranteedToExceedMaximum(maximum: (uint)parts[1] - 1);
                         high = 0;
                     }
                     else
                     {
-                        low = RawInt32() & (int)MaskOutBitsGuaranteedToExceedMaximum(maximum: (uint)parts[0] - 1);
+                        low = RawInt32() & (int)MaskToRemoveBitsGuaranteedToExceedMaximum(maximum: (uint)parts[0] - 1);
                         mid = 0;
                         high = 0;
                     }
@@ -647,7 +647,7 @@ namespace NUnit.Framework.Internal
             return unchecked(raw % range + min);
         }
 
-        private static uint MaskOutBitsGuaranteedToExceedMaximum(uint maximum)
+        private static uint MaskToRemoveBitsGuaranteedToExceedMaximum(uint maximum)
         {
             // http://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2 but
             // without the value-- and value++
--- a/src/NUnitFramework/framework/Internal/Randomizer.cs
+++ b/src/NUnitFramework/framework/Internal/Randomizer.cs
@@ -579,35 +579,59 @@ namespace NUnit.Framework.Internal
 
             unchecked
             {
+                // Flooring sets the exponent (scale) to zero. This is necessary even if the number is already an
+                // integer so that the low, mid and high parts of the mantissa get shifted by the appropriate power of
+                // ten until they only contain the integral part of the number.
+                // May as well set it back to the parameter so that the `result < max` lower down is doing less work.
                 max = decimal.Floor(max);
-                var parts = decimal.GetBits(max);
-                var scale = (byte)(parts[3] >> 16);
-                if (scale != 0) throw new InvalidOperationException("Decimal.Floor returned a value whose scale was not 0.");
+
+                var parts = DecimalParts.FromValue(max);
+
+                // If scale is not zero, that means that low, mid, and high are shifted to make room for digits from the
+                // fractional part of the number. We relied on Decimal.Floor to prevent this.
+                Guard.OperationValid(parts.Scale == 0, "Decimal.Floor returned a value whose scale was not 0.");
 
                 while (true)
                 {
+                    // Fill all 96 bits with uniformly-distributed randomness except for the bits that must be zero in
+                    // order to stay under the exclusive maximum.
                     int low, mid, high;
 
-                    if (parts[2] != 0)
+                    if (parts.High != 0)
                     {
+                        var inclusiveMaximum = parts.High - 1;
+
                         low = RawInt32();
                         mid = RawInt32();
-                        high = RawInt32() & (int)MaskToRemoveBitsGuaranteedToExceedMaximum(maximum: (uint)parts[2] - 1);
+                        high = RawInt32() & (int)MaskToRemoveBitsGuaranteedToExceedMaximum(inclusiveMaximum);
                     }
-                    else if (parts[1] != 0)
+                    else if (parts.Mid != 0)
                     {
+                        var inclusiveMaximum = parts.Mid - 1;
+
                         low = RawInt32();
-                        mid = RawInt32() & (int)MaskToRemoveBitsGuaranteedToExceedMaximum(maximum: (uint)parts[1] - 1);
+                        mid = RawInt32() & (int)MaskToRemoveBitsGuaranteedToExceedMaximum(inclusiveMaximum);
                         high = 0;
                     }
                     else
                     {
-                        low = RawInt32() & (int)MaskToRemoveBitsGuaranteedToExceedMaximum(maximum: (uint)parts[0] - 1);
+                        var inclusiveMaximum = parts.Low - 1;
+
+                        low = RawInt32() & (int)MaskToRemoveBitsGuaranteedToExceedMaximum(inclusiveMaximum);
                         mid = 0;
                         high = 0;
                     }
 
                     var result = new decimal(low, mid, high, false, 0);
+
+                    // By masking out the random bits which would put the result at or over the exclusive maximum, the
+                    // chance of having to loop and try again is strictly less than 50% in the worst-case scenario. The
+                    // best-case scenario is 0% chance, and an average is 25% chance per iteration.
+
+                    // For the worst-case scenario of 50% chance per iteration, the total chance of having to iterate
+                    // twice is < 25%. Three times, 12.5%. And so on.
+                    // For the average scenario of 25% chance per iteration, the total chance of having to iterate twice
+                    // is < 6.25%. Three times, 1.5625%. And so on.
                     if (result < max) return result;
                 }
             }
--- a/src/NUnitFramework/framework/Internal/Randomizer.cs
+++ b/src/NUnitFramework/framework/Internal/Randomizer.cs
@@ -563,10 +563,14 @@ namespace NUnit.Framework.Internal
         /// </summary>
         public decimal NextDecimal()
         {
-            int low = Next(0, int.MaxValue);
-            int mid = Next(0, int.MaxValue);
-            int high = Next(0, int.MaxValue);
-            return new Decimal(low, mid, high, false, 0);
+            var bytes = new byte[sizeof(int) * 3];
+            NextBytes(bytes);
+
+            var low = BitConverter.ToInt32(bytes, 0);
+            var mid = BitConverter.ToInt32(bytes, 4);
+            var high = BitConverter.ToInt32(bytes, 8);
+
+            return new decimal(low, mid, high, false, 0);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/NUnitFramework/framework/Internal/Randomizer.cs
```diff
@@ -563,18 +563,67 @@ public string GetString()
...
+            var mid = BitConverter.ToInt32(bytes, 4);
+            var high = BitConverter.ToInt32(bytes, 8);
+
+            return new decimal(low, mid, high, false, 0);
         }
 
         /// <summary>
         /// Returns a random decimal between positive zero and the specified maximum.
         /// </summary>
         public decimal NextDecimal(decimal max)
```
CharliePoole: I think we need to document the algorithm in a comment.

src/NUnitFramework/framework/Internal/Randomizer.cs
```diff
@@ -563,18 +563,67 @@ public string GetString()
...
         {
-            return NextDecimal() % max;
```
CharliePoole: I'm almost ready to bet that we have other uses of modulo with other numeric types. Do we need to test them too?

jnm2: Yes. Should I open a separate issue for non-decimal types rather than do everything in this PR?

src/NUnitFramework/framework/Internal/Randomizer.cs
jnm2: Good idea.

jnm2:
> I think this may be the first test we have written of the distribution of random values. Do we need others? Should they be separate from our general unit tests?

It is. I'm also realizing it could be a lot more thorough; each bit of the same 10,000 decimals can be separately checked for flat distribution. Also, I should test 96 maximums, one for each possible number of significant bits.

Because we were using `Next(0, int.MaxValue)`, the `low`, `mid`, and `high` ints used to create the decimals never had the high bit set. That meant `NextDecimal()` never generated a number between `decimal.MaxValue / 2` and `decimal.MaxValue` as well as having some gaps.

Because the tests run so fast (10 ms on my machine), I'm not thinking of a benefit to keeping distribution tests in a separate test run. Maybe a separate test class?

jnm2:
I need to fix this too: 10,000 isn't enough iterations for a 0.01 precision.

It isn't true randomness if there isn't *some* chance that the distribution of a particular random sequence is biased, so test failure is always possible with these tests. It's just a matter of calculating the likelihood and making it sufficiently small in the same way that guid collisions are sufficiently negligible.

Maybe the test could keep running up to a time limit before failing, just as long as there is a minimum number of iterations.

src/NUnitFramework/framework/Internal/Randomizer.cs
CharliePoole: I'm for small PRs!

CharliePoole:
I wasn't thinking of efficiency so much as the fact that this is a bit out of our wheelhouse. We're more users of pseudo-random number generation than creators, even though some of us may have a degree of experience with them.

Ideally, we would be using somebody else's library and testing library. Ideally, we 
might allow users to replace the underlying sequence generator.

So for me, separation implies a possible future division of labor.

But for now, a separate class is fine. :smile:

jnm2:
It seems to me that by synthesizing our own NextDecimal methods at all, we're placing ourselves firmly in the pseudorandom value creators' seat. Our implementation is either good or poor and if we have it at all, I'd like it to be good.

CharliePoole:
Agreed. I'm speculating about whether we can get out of that position. I do know there are some very good generalized suites for testing random number generators. I also know that our synthesized values can never get better than the underlying generator, which adds a bit of complexity.

src/NUnitFramework/framework/Internal/Randomizer.cs
jnm2: I just checked and I don't see anything else in Randomizer.cs that is using modulo without also correcting for its bias.

jnm2:
@CharliePoole Ready for review again.

CharliePoole:
Seems like I have to start over from the beginning because of the force push.

jnm2:
If you click this link above, it shows you the diff:

image

jnm2:
That's not as helpful as I thought since you can't go commit by commit. I'm sorry.

jnm2:
The original review is also not lost. I clicked Resolve to collapse it:

image

CharliePoole:
GitHub says that @jnm2 dismissed my stale review. Not sure but I think that means the forced push. It's useful to be able to look back at what one reviewed and asked for as new requests for review arrive. We should try to avoid situations where re-review is required and the reviewer can no longer access the old review.

jnm2:
I'll prefer to be forward-only in the future to make reviewing easier. I think I will still track which things I have finished by collapsing with the Resolve button though.

It's not possible for review comments to be lost. Any new push, even a non-force push, will dismiss all reviews as stale. Dismissing a review only clears the status. It doesn't remove any information that had been added to the thread.

CharliePoole:
As a reviewer, I want to click on my review and see what I wrote. If GitHub says it can't find the information, then my review is "lost" to me even if there is some further search or steps I don't know about that would allow me to find them.

Part of this problem, of course, is in the policy that requires a re-review for every change that is made. That's because we (1) have a strict policy of two committers to approve a review and (2) have decided to enforce that policy through GitHub rather than socially.

jnm2:
GitHub doesn't say that it can't find the information since that's not what 'dismiss' means and there is UI in the same location as the original information saying "click to show."

Both of these effects (dismissing reviews on non-force pushes and collapsing addressed comments) would still be in effect even if all the GitHub requirements were turned off, I think.

CharliePoole:
Trust me... that __is__ what GitHub UI said when I clicked. :smile: 

The previous review is gone - either really gone or gone for practical purposes, since I can't find it. I figure if someone with my experience can't immediately find the information, then it's gone for most people. This isn't a criticism, by the way, of anything you did and I hope it did not sound that way.

jnm2:
That's a mystery. Earlier on desktop and now on my phone I can see everything that has been said in the previous review.

Thanks! I don't take it as a criticism. I'm commenting to clarify that I don't see any action the NUnit team could take to avoid these outcomes (besides report them as bugs).

oznetmaster:
For some reason I cannot determine, on my CF build, the NextDecimalNotBiased test fails for values 0, 32, and 64.  All other values work correctly.

For those three values, the result 0.0d.

Any thoughts as to where to look?

The NextDecimalWithMaximumNotBiased test passes as well.

CharliePoole:
@jnm2 Are you referring to the code comments? I can see those. It's the review itself I can't find. For example, did I previously approve it or ask for changes? What reasons did I give?

IMO, the action that could be taken would be to not rely on GitHub software to enforce our cooperative agreements. Then no bug reports or fixes would be needed. However, this is probably better not discussed here.

jnm2:
That means that `Random.Next(int.MinValue, int.MaxValue)` is returning numbers whose least significant bit is never set.

For CF then I'd recommend RawInt32 to be implemented using NextBytes and BitConverter.ToInt32. NextBytes takes a 1 random sample per byte on .NET Framework, whereas `Next(int.MinValue, int.MaxValue)` takes two samples.

You would find the sourceof.net implementation interesting for `Next(int, int)`. It has a special case that I'm guessing CF does not have. I'm going to assume CF takes only one sample.

jnm2:
@oznetmaster Hey, where did your comment go?

oznetmaster:
It is still there

CharliePoole:
@jnm2 Yes, and that implies any tests we write have to  take into account the underlying Random distribution, which can differ according to platform and possibly even version. Wow!

jnm2:
@oznetmaster I got mixed up, sorry. [Here's the netfx implementation.](https://referencesource.microsoft.com/#mscorlib/system/random.cs,ddd1e9e09c70bab5)
Does the NextBytes route seem workable?

oznetmaster:
I am testing it now

jnm2:
@CharliePoole 

> Are you referring to the code comments? I can see those. It's the review itself I can't find. For example, did I previously approve it or ask for changes? What reasons did I give?

It is here: https://github.com/nunit/nunit/pull/3241#pullrequestreview-237151229 (the reasons you gave are right above the code comments)
You originally did a comment review rather than approving or ask for changes. I noticed that right away. If you had requested changes, that line would still be showing a red +/- icon and would say "CharliePoole requested changes" rather than "CharliePoole left a comment."

oznetmaster:
@jnm2 That fixed the problem.

Does making that change standard make the code more platform independent in the long run?

RawUInt32 and RawUInt64 both use that code, so why not RawInt32?

CharliePoole:
@jnm2 So I think you are saying that when I do a review and check "Comment"
it's not recorded as a review at all. That would explain why I couldn't
find it for sure. What I was looking for was something exactly like the box
in which I entered the review in the first place, because (as I've often
said) I implement a stateless protocol with respect to comments and
reviews. Each time I look I'm starting over because I comment on too many
things for too many people to remember state for each one. :smile:

jnm2:
@CharliePoole It is recorded just like the other two kinds in all the same ways, including in the summary at the bottom of the page before the PR is merged. It just uses a different icon (one of three) and a different phrase (one of three) underneath the text "CharliePoole reviewed 7 days ago."

jnm2:
@oznetmaster I'd ordinarily not support it because it's a bug if a framework only returns even integers when you ask for a range. However, it just struck me that `Next(int.MinValue, int.MaxValue)` excludes the number `int.MaxValue`! It's almost impossible to detect this in a test (chance is one in four billion) but it will force me to switch to NextBytes as the next most obvious implementation.

PR incoming!

oznetmaster:
@jnm2 The CF NUnit framework is not only returning even integers if a range is specified.  All 96 tests are being run correctly.  It was only the 0, 32, 64 tests that were failing.

oznetmaster:
As a point of information, I just ran the RandomizerBiasTests (fixed as above) on the CF framework 100 times, and it passed every time.

jnm2:
What do you mean by the 0, 32, and 64 tests failing? That would be three of the 96 tests, corresponding to bits 0, 32, and 64, wouldn't it?

oznetmaster:
Yes.  Test cases 0, 32, and 64 (generated by the Range attribute).

Once the changes were made (as above) all the test cases pass all the time.

jnm2:
Right, so the only way that could be is if CF's `Next(int.MinValue, int.MaxValue)` fails to set the least significant bit (ones place). Does that make sense?

oznetmaster:
CF's Next:

```
public virtual int Next(int minValue, int maxValue)
{
    if (minValue > maxValue)
    {
        throw new ArgumentOutOfRangeException();
    }
    int num = maxValue - minValue;
    if (num >= 0)
    {
        return (((int) (this.Sample() * num)) + minValue);
    }
    long num2 = maxValue - minValue;
    return (((int) ((long) (this.Sample() * num2))) + minValue);
}
```

jnm2:
> For those three values, the result 0.0d.

This means that bits 0, 32, and 64 were set 0 out of 100,000 times. Those bits came directly from bit 0 of `Next(int.MinValue, int.MaxValue)`; therefore, bit 0 of `Next(int.MinValue, int.MaxValue)` was set 0 out of 300,000 tries.

[Here's the netfx implementation](https://referencesource.microsoft.com/#mscorlib/system/random.cs,ddd1e9e09c70bab5) I linked above:

```cs
      public virtual int Next(int minValue, int maxValue) {
          if (minValue>maxValue) {
              throw new ArgumentOutOfRangeException("minValue",Environment.GetResourceString("Argument_MinMaxValue", "minValue", "maxValue"));
          }
          Contract.EndContractBlock();
          
          long range = (long)maxValue-minValue;
          if( range <= (long)Int32.MaxValue) {  
              return ((int)(Sample() * range) + minValue);
          }          
          else { 
              return (int)((long)(GetSampleForLargeRange() * range) + minValue);
          }
      }

      private double GetSampleForLargeRange() {
          // The distribution of double value returned by Sample 
          // is not distributed well enough for a large range.
          // If we use Sample for a range [Int32.MinValue..Int32.MaxValue)
          // We will end up getting even numbers only.
 
          int result = InternalSample();
          // Note we can't use addition here. The distribution will be bad if we do that.
          bool negative = (InternalSample()%2 == 0) ? true : false;  // decide the sign based on second sample
          if( negative) {
              result = -result;
          }
          double d = result;
          d += (Int32.MaxValue - 1); // get a number in range [0 .. 2 * Int32MaxValue - 1)
          d /= 2*(uint)Int32.MaxValue - 1  ;              
          return d;
      }
```

This is why I'm so sure that CF is faulty, but you can remove all doubt by testing whether this ever completes:

```cs
var rng = new Random();

while ((rng.Next(int.MinValue, int.MaxValue) & 1) == 0)
{
}
```

When I run the above code on .NET Framework but using the CF implementation you posted, it never completes.

PR #3258, opened 2019-05-23
Fix wrong bias test iteration count resulting in high probability of false negative

### LINKED ISSUE #3253: Chance of failure in random bias tests is not sufficiently low for CI
If I knew what I was doing with the binomial confidence interval calculator, 100,000 iterations per test and the ±0.01% tolerance around 50% (same number of ones and zeros) ~~should yield around a 99.9999% success rate, or one failure in a million.~~ **Edit: I rounded to 0.01 from 0.0077 and didn't realize that that rounding actually made it better than one failure in a *billion*, which is 0.0097.**

I've seen around ten failures in CI since the PR was merged six hours ago. That means that **either I don't know what I'm doing with the math (send help!)** or CI has a much worse source of randomness than what I saw at my local machine when I left my computer running them with `[Repeat(10000)]`.

### Options

1. Make the bias tests explicit or skip them when an environment variable named `CI` is present
   - Explicit is not as nice unless we put a note in Randomizer.cs and RandomizerTests.cs to remember to manually run RandomizerBiasTests.cs. 

2. See what other people do and maybe get a more standard bias test

3. Get someone with more experience to look at what I've done

@nunit/framework-team Do other options come to mind? I don't really want to delete them, and in fact I want to add 10-100 times more of them at some point.

jnm2:
Fixes #3253.
Also added a thorough explanation and stopped rounding `0.00965802342765787` to `0.01`.

### DIFF
--- a/src/NUnitFramework/tests/Internal/RandomizerBiasTests.cs
+++ b/src/NUnitFramework/tests/Internal/RandomizerBiasTests.cs
@@ -58,7 +58,7 @@ namespace NUnit.Framework.Internal
             const decimal wholeRange = decimal.MaxValue * (2 / 3m);
             const decimal halfRange = wholeRange / 2;
 
-            const int totalCount = 10000;
+            const int totalCount = 100000;
             var countInTopHalf = 0;
 
             for (var i = 0; i < totalCount; i++)
--- a/src/NUnitFramework/tests/Internal/RandomizerBiasTests.cs
+++ b/src/NUnitFramework/tests/Internal/RandomizerBiasTests.cs
@@ -51,7 +51,7 @@ namespace NUnit.Framework.Internal
         negatives would then only occur on average only one month out of every 286 years.
 
         One in a billion, or 99.9999999% confidence, seems desirable. To obtain this confidence level, we need a
-        sufficiently high trial count or a sufficiently low tolerance. I chose a trial count of 100000 because
+        sufficiently high trial count or a sufficiently high tolerance. I chose a trial count of 100000 because
         that takes only about 10ms on my machine. That results in a confidence interval (see Wikipedia link above)
         of ~±0.0097 around 0.5. 0.5 of the trials are expected to succeed, where success means something
         like "the third bit is set"—something which we expect to happen half the time if the generator is not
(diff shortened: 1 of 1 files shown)


src/NUnitFramework/tests/Internal/RandomizerBiasTests.cs
```diff
@@ -58,18 +108,17 @@ public static void NextDecimalWithMaximumIsNotBiased()
             const decimal wholeRange = decimal.MaxValue * (2 / 3m);
             const decimal halfRange = wholeRange / 2;
 
-            const int totalCount = 10000;
```
jnm2: @rprouse Now that we require VS2017+, is there any reason not to remove the <LangVersion> restriction so that we can beginning using C# 7? Specifically here, I wanted to use `10_000` and that would have helped me and maybe reviewers of the first PR to spot the inconsistency with the other method's `100_000`.

rprouse: I think we should keep the `<LangVersion>` to make sure language features that aren't supported in VS 2017 don't creep in, but I'm fine with updating it to 7 from 6. I'm not sure, but I think we may require 15.3+ which would allow 7.1, but our `BUILDING.md` just says VS 2017, so 7 is safest.

src/NUnitFramework/tests/Internal/RandomizerBiasTests.cs
```diff
@@ -25,18 +25,68 @@ namespace NUnit.Framework.Internal
...
+        99.9999% ^ ‭291,000‬, or ~74.75%. The chance of there being at least one false negative per month is then
+        100% - 74.75%, or 25.25‬%, or one false negative on average every four months. This seems too frequent.
+
+        Rather than expecting at least one false negative every four months, we can drive that chance down by
+        choosing the per-trial confidence to be 99.9999999%, or a one-in-a-billion chance of false negative.
+        This brings the per-month chance of all trials having no false negative to 99.97%. One or more false
+        negatives would then only occur on average only one month out of every 286 years.
+
+        One in a billion, or 99.9999999% confidence, seems desirable. To obtain this confidence level, we need a
+        sufficiently high trial count or a sufficiently low tolerance. I chose a trial count of 100000 because
```
jnm2:
```suggestion
        sufficiently high trial count or a sufficiently high tolerance. I chose a trial count of 100000 because
```

jnm2:
@mikkelbu No problem. I can't say I know what I'm doing either, but at least my entire thought process is written out for someone to debug if needed.

PR #3261, opened 2019-05-26
Workaround for lost MethodInfo.Invoke stack trace beginning in Mono 5.20

jnm2:
Fixes #3257. This is a regression in Mono 5.20 which is interfering with @oznetmaster's work and which will be on Azure Pipelines any minute as far as I know, affecting all our builds. (5.20 has been out for a month and a half and AzDO is still on 5.18.)

I stepped through the code on MonoDevelop on Ubuntu. The stack trace is present when the exception is initially thrown:

image

On the very next step when you press F10, the step out of MethodInfo.Invoke, the stack trace has been overwritten with useless info:

image

Full text of the thrown exception's StackTrace after stepping out of MethodInfo.Invoke:

image

Issue #3257, opened 2019-05-22
Running under mono


oznetmaster:
trying to run the .NET 4.5 build under mono, executing nunitlite.tests.exe results in:

```
mono nunitlite.tests.exe
NUnitLite 3.13.0 (.NET Framework 4.5 Debug)
Copyright (c) 2019 Charlie Poole, Rob Prouse

Runtime Environment
   OS Version: Unix 4.4.0.148
  CLR Version: 4.0.30319.17020

Test Files
    /media/sf_D_DRIVE/Projects/github/nunit/bin/debug/net45/nunitlite.tests.exe

Test Discovery
  Start time: 2019-05-22 11:19:56Z
    End time: 2019-05-22 11:19:56Z
    Duration: 0.023 seconds

System.TypeLoadException: Could not load type 'NUnit.Framework.Internal.TestSuiteResult' from assembly 'nunit.framework, Version=3.13.0.0, Culture=neutral, PublicKeyToken=2638cd05610744eb'.
  at NUnit.Framework.Internal.Execution.WorkItemBuilder.CreateWorkItem (ITest test, ITestFilter filter, Boolean recursive) <0x41997a00 + 0x00143> in <filename unknown>:0
  at NUnit.Framework.Api.NUnitTestAssemblyRunner.RunAsync (ITestListener listener, ITestFilter filter) <0x41994900 + 0x0010f> in <filename unknown>:0
  at NUnit.Framework.Api.NUnitTestAssemblyRunner.Run (ITestListener listener, ITestFilter filter) <0x41994880 + 0x0001b> in <filename unknown>:0
  at NUnitLite.TextRunner.RunTests (NUnit.Framework.Internal.TestFilter filter, IDictionary`2 runSettings) <0x41994610 + 0x00059> in <filename unknown>:0
  at NUnitLite.TextRunner.Execute () <0x41984560 + 0x0071f> in <filename unknown>:0
```

I have no idea what to do to debug this.  This test exe runs correctly under .NET 4.5 on windows.

oznetmaster:
Never mind.
Mono runtime was apparently still at 4.0 on machine.  Updated to latest, and all is ok.

oznetmaster:
Well, I spoke to soon.  Tried running full framework tests, and got:
```
 mono nunitlite-runner.exe nunit.framework.tests.dll
NUnitLite 3.13.0 (.NET Framework 4.5 Debug)
Copyright (c) 2019 Charlie Poole, Rob Prouse

[...]

PR #3263, opened 2019-05-26
Enable and begin using C# 7.0 syntax (requires capable compiler)

jnm2:
There is one commit to enable C# 7 and then four small cleanup commits.

@rprouse at https://github.com/nunit/nunit/pull/3258#discussion_r286954570:

> I think we should keep the `<LangVersion>` to make sure language features that aren't supported in VS 2017 don't creep in, but I'm fine with updating it to 7 from 6. I'm not sure, but I think we may require 15.3+ which would allow 7.1, but our `BUILDING.md` just says VS 2017, so 7 is safest.

### DIFF
--- a/src/NUnitFramework/Directory.Build.props
+++ b/src/NUnitFramework/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <LangVersion>6</LangVersion>
+    <LangVersion>7</LangVersion>
     <Features>strict</Features>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
--- a/src/NUnitFramework/framework/Compatibility/System.Threading/SpinWait.cs
+++ b/src/NUnitFramework/framework/Compatibility/System.Threading/SpinWait.cs
@@ -283,7 +283,7 @@ namespace System.Threading
     /// </summary>
     internal static class PlatformHelper
     {
-        private const int PROCESSOR_COUNT_REFRESH_INTERVAL_MS = 30000; // How often to refresh the count, in milliseconds.
+        private const int PROCESSOR_COUNT_REFRESH_INTERVAL_MS = 30_000; // How often to refresh the count, in milliseconds.
         private static volatile int s_processorCount; // The last count seen.
         private static volatile int s_lastProcessorCountRefreshTicks; // The last time we refreshed.
 
--- a/src/NUnitFramework/nunitlite.tests/NUnit2XmlOutputWriterTests.cs
+++ b/src/NUnitFramework/nunitlite.tests/NUnit2XmlOutputWriterTests.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
@@ -127,16 +127,14 @@ namespace NUnitLite.Tests
         public void TestResults_HasValidDateAttribute()
         {
             string dateString = RequiredAttribute(topNode, "date");
-            DateTime date;
-            Assert.That(DateTime.TryParse(dateString, out date), "Invalid date attribute: {0}", dateString);
+            Assert.That(DateTime.TryParse(dateString, out _), "Invalid date attribute: {0}", dateString);
         }
 
         [Test]
         public void TestResults_HasValidTimeAttribute()
         {
             string timeString = RequiredAttribute(topNode, "time");
-            DateTime time;
-            Assert.That(DateTime.TryParse(timeString, out time), "Invalid time attribute: {0}", timeString);
+            Assert.That(DateTime.TryParse(timeString, out _), "Invalid time attribute: {0}", timeString);
         }
 
         [Test]
@@ -204,10 +202,9 @@ namespace NUnitLite.Tests
         [Test]
         public void TestSuite_HasValidTimeAttribute()
         {
-            double time;
             var timeString = RequiredAttribute(suiteNode, "time");
             // NOTE: We use the TryParse overload with 4 args because it's supported in .NET 1.1
-            var success = double.TryParse(timeString,System.Globalization.NumberStyles.Float,System.Globalization.NumberFormatInfo.InvariantInfo, out time);
+            var success = double.TryParse(timeString, System.Globalization.NumberStyles.Float, System.Globalization.NumberFormatInfo.InvariantInfo, out _);
             Assert.That(success, "{0} is an invalid value for time", timeString);
         }
 
--- a/src/NUnitFramework/testdata/TestCaseAttributeFixture.cs
+++ b/src/NUnitFramework/testdata/TestCaseAttributeFixture.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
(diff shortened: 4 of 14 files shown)


CharliePoole:
So it's eventually clear in the release notes, shouldn't the title of the issue be __Require__ C# 7.0?

jnm2:
@CharliePoole What is (already) required is the Visual Studio 2017 SDK. No requirements are changing.

jnm2:
I broke a ValueTuple test. Fixing now.

jnm2:
Fixed a broken test. Ready for review.

CharliePoole:
So no existing (or even hypothetical) compiler other than VS can build the framework now? I hadn't really thought about it that way before.

In any case, the code now uses C# 7 features. So the PR requires C# 7 features. The initial commit __enabled__ those features but subsequent commits __require__ them if the code is to compile.

I'm not objecting to the change, but saying you are "enabling" something that is really mandatory sounds like we have something to hide.

jnm2:
@CharliePoole The `dotnet` command line and `msbuild` both ship with the ability to compile C# 7.0 code. This is the case whether you use Visual Studio 2017+, Visual Studio Code (OmniSharp), and MonoDevelop (Mono itself supports C# 7.0).

The only way you could get yourself into a situation where C# 7.0 wouldn't compile for you is if you are on Windows and aren't using Visual Studio 2017 or later, which is explicitly required by us for building our codebase on Windows, or not on Windows and you're using a version of Mono earlier than 5.0.0 which is when they started shipping the open-source Roslyn compiler which came with C# 7 support at the time.

Cake (our `build.sh` dependency) won't even run unless you have Mono 5.12+.

jnm2:
Let's boil it down to this. Supported methods of building from the CLI *before* this pull request:

### Windows

### CLI builds

- `build.ps1` - requires Visual Studio 2017+ to be installed, so C# 7 capability is already guaranteed
- `dotnet build` - requires .NET Core 2.0, so C# 7 capability is already guaranteed

### IDEs

- Visual Studio 2017+ (C# 7 capability is already guaranteed)
- Visual Studio Code with OmniSharp (C# 7 capability is already guaranteed)
- Others such as Rider: not necessarily supported by us, but in general, C# 7 capability is already guaranteed

### Other OSs

#### CLI builds

- `build.sh` - requires Mono 5.12 or Cake won't run, so C# 7 capability is already guaranteed
- `dotnet build` - requires .NET Core 2.0, so C# 7 capability is already guaranteed

### IDEs

- MonoDevelop: not explicitly supported by us and I don't know what version we'd go back to. (C# 7.0 is supported from 5.0.0 on.)
- Visual Studio Code with OmniSharp (C# 7 capability is already guaranteed)
- Others such as Rider: not explicitly supported by us, but in general, C# 7 capability is already guaranteed

jnm2:
So nothing is being made mandatory except upgrading MonoDevelop to at least 5.0.0 if you use that as your IDE. In practice we see that people end up on 5.20 or whatever their latest release is.

CharliePoole:
@jnm2 I am frankly stunned that a simple suggestion to change one word in the __description__ (not the code) is getting such pushback from you. For whatever reason, you are answering as if I had objected to the PR itself and not just made a suggestion as to your wording. All your reasoning does is explain __why it's OK to require C# 7.0__ but we do actually require it, as I said in the first place.

I really don't want to spend much more time on one word, but I'd kind of like to know why you __don't__ want to say "Requires" here. What's your downside?

jnm2:
To me, it sounds confusing. We require things like underscores in field names. We won't require anyone to use C# 7.0. 'Enable' seems to much better communicate both what contributors now can do as well as the obvious implication we now have C# 7.0 code and therefore a C# 7.0 compiler is needed.

Would this be a good compromise?

> Enable C# 7.0 (requires capable compiler)

jnm2:
I apologize for the pushback too. It sounded like you thought it was possible that a user of our codebase could be disrupted by these changes. I want to make sure our expectations are all on the same page since my intention is that this will not disrupt anyone.

CharliePoole:
I don't think "Require C# 7.0" remotely means "Require people to use C# 7.0 features wherever possible." This has nothing to do with writing code for NUnit. We use standards for that purpose. This has to do with someone compiling our existing code.

If you really want to stress that we are "not actually changing anything" you can say something like "Clarify requirement for a C# 7.0 capable compiler" 

The audience I'm concerned about is someone who decides next month or year to compile NUnit locally. When they look back at past changes, I'd like them to know they need a C# 7.0 enabled compiler. We can't tell them what compiler that will be, since it's the future.

People working in the codebase as part of our project or separately have other sources of information regarding what features to use.

jnm2:
It's not just clarifying the existing requirement- it's also loosening the restriction that only C# 6 may be used. The big thing that's happening is that we're enabling and using C# 7, and a side effect is that we now have two reasons for our existing requirements instead of just one (which was the new-style csproj).
Can I put the more significant thing first?

How about:

> Permit and depend on C# 7.0 syntax

When future contributors are looking through, could they be more likely to see something in README.md or BUILDING.md than they are to spot something deep in a list of PRs in an old release notes section? I'd be happy to clarify in places like those if you think that would help them get up and running quickly.

CharliePoole:
I see the difference in perspective. If contributors are __permitted__ to use C# 7.0 features, then folks building NUnit are __required__ to have a compiler that can build those features.

With that in mind, please pick whatever description makes sense to you.

jnm2:
Okay. With what I just picked, I hope it communicates that our focus was on unblocking contributors and improving the codebase while still drawing attention to the fact that the codebase now requires a 7+ compiler.

src/NUnitFramework/framework/Constraints/ComparisonAdapter.cs
```diff
@@ -167,7 +167,7 @@ public override int Compare(object expected, object actual)
                 if (!TypeHelper.TryCast(actual, out actualCast))
                     throw new ArgumentException($"Cannot compare to {actual?.ToString() ?? "null"}");
 
-                return comparison.Invoke((T)expected, (T)actual);
+                return comparison.Invoke(expectedCast, actualCast);
```
mikkelbu: Since we are using C# 7, can't we move the declaration of `expectedCast` and `actualCast` into the `TypeHelper.TryCast`?

jnm2: 👍 I wasn't thinking, but I meant to take care of this in a future PR where I do `out var` everywhere. This commit will be overwritten by that anyway, so I'll remove it from this PR.

jnm2:
@mikkelbu Azure Pipelines just picked up Mono 5.20, so this PR can't merge before https://github.com/nunit/nunit/pull/3261.

oznetmaster:
I understand the desire to use C# 7 in future work.

However, I do not understand the rush to retrofit existing, working, code to use C# 7.

We have never encouraged the reformatting of existing source code to meet "new" coding standards.

Why go back and change existing code to use C# 7, just because we now can?

oznetmaster:
I will also point out that many developers/organizations do not encourage/allow the use of many C# 7 syntactic features because it is felt that they actually make code maintenance and review more difficult, not easier.

So have the ability to use them, and actually using them, could be two different things.

jnm2:
@oznetmaster For digit separators: because not having them bit me recently and destabilized the master branch.

For tuples: improved code clarity, but it's not important.

For discards/out var (out var is not in this PR): I have an ongoing annoyance at the way the code has had to be just by looking at these two things. However, I get that this carries zero weight with anyone else and it's not important.

oznetmaster:
I agree that in specific cases, it does improve things.

Digit separators could help for readability of large constants, although I personally have never had the issue.

I do find the ability to name elements and decompose tuples does really clarify the code.  I hate Item1, Item2, ...  

Discards, probably not so much.  Having a variable named dummyxxx may actually be clearer then a discard when reading the code.

out var is mixed bag.  I like the simplification, but also find myself look for the variable definition later in the code when the variable is sued, and I cannot find it.  So I do not use out var if the variable usage is not very localized.

jnm2:
> Discards, probably not so much. Having a variable named dummyxxx may actually be clearer then a discard when reading the code.

I often do `paramName: out _` to make it really clear what is being discarded (because I often use named arguments in general for readability), but what I really dislike is having to declare a variable on the previous line because of it.

Otherwise it sounds like this PR would fit your definition of improvements then?

oznetmaster:
My only additional comment is about discards.

For all of the methods where I would use discards a lot, like the TryParse methods, I have "wrappers" that I call Isxxxx (string s) returning a bool, which internally calls the TryParse method for that type, and returns the result, discarding the output value.

This reduces the call overhead, and actually clarifies the readability of the code, which the discard does not.

Otherwise, I agree with the concept.

rprouse:
The Travis failures are just Travis crashing again, but the Linux failures on Azure DevOps look like real test failures. Can you take a look?

jnm2:
@rprouse They kinda look real, but it's just because Azure Pipelines's Ubuntu image just picked up Mono 5.20. This PR can't merge before #3261 (thus the failure).

rprouse:
Looks like the Azure DevOps failures are the ones discussed in #3257. Should we fix them separately?

rprouse:
I've merged your Mono fix so I am going to override and do the same with this PR.

oznetmaster:
Actually, as I look at the use of the number separator, I actually find it more difficult to read the numbers with the separator then without it.

Not sure we should make this usage universal.

Perhaps some other comments would be useful.

PR #3265, opened 2019-05-27
Add tests that IRepeatTest attributes are resoved via IMethodInfo

### LINKED ISSUE #3264: Test that IRepeatTest only gets attributes via the IMethodInfo interface
Waiting for https://github.com/nunit/nunit/pull/3206 to go in.

jnm2:
Closes #3264 which tracks the followup to https://github.com/nunit/nunit/pull/3206 of adding tests.


jnm2:
It's a surprise to me that a user was able to find the ITypeInfo abstraction useful, and even so it's only as a workaround.

PR #3272, opened 2019-05-30
Use out declarations

jnm2:
(Failure is not related; https://github.com/nunit/nunit/pull/3271 needs to be merged first.)

My aim is to improve readability. This is a way to declare things closer to their usage where they are more relevant. 
If you see an instance where the code is less understandable than it was before, please comment on it.

/cc @oznetmaster

### DIFF
--- a/src/NUnitFramework/tests/TestUtilities/Comparers/ObjectToStringComparer.cs
+++ b/src/NUnitFramework/tests/TestUtilities/Comparers/ObjectToStringComparer.cs
@@ -37,10 +37,8 @@ namespace NUnit.TestUtilities.Comparers
         int System.Collections.IComparer.Compare(object x, object y)
         {
             WasCalled = true;
-            int intX = 0;
-            int intY = 0;
 
-            if(int.TryParse(x.ToString(), out intX) && int.TryParse(y.ToString(), out intY))
+            if (int.TryParse(x.ToString(), out var intX) && int.TryParse(y.ToString(), out var intY))
             {
                 return intX.CompareTo(intY);
             }
--- a/src/NUnitFramework/framework/Attributes/RangeAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RangeAttribute.cs
@@ -177,8 +177,7 @@ namespace NUnit.Framework
 
             var valueGenerator = ValueGenerator.Create(parameter.ParameterType);
 
-            ValueGenerator.Step step;
-            if (!valueGenerator.TryCreateStep(_step, out step))
+            if (!valueGenerator.TryCreateStep(_step, out var step))
             {
                 // ValueGenerator.CreateStep has the responsibility to enable Byte values to be incremented
                 // by the Int32 value -1. Or perhaps in the future, DateTime values to be incremented by a TimeSpan.
@@ -188,8 +187,7 @@ namespace NUnit.Framework
                 // that are only of a different type due to IL limitations or NUnit smoothing over overload differences.
                 // See the XML docs for the ParamAttributeTypeConversions class.
 
-                object stepValueToRequire;
-                if (!ParamAttributeTypeConversions.TryConvert(_step, parameter.ParameterType, out stepValueToRequire))
+                if (!ParamAttributeTypeConversions.TryConvert(_step, parameter.ParameterType, out var stepValueToRequire))
                 {
                     // This will cause CreateStep to throw the same exception as it would throw if TryConvert
                     // succeeded but the value generator still didn’t recognize the step value.
--- a/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
@@ -267,9 +267,8 @@ namespace NUnit.Framework
                 parms = new TestCaseParameters(this);
 
                 // Special handling for ExpectedResult (see if it needs to be converted into method return type)
-                object expectedResultInTargetType;
                 if (parms.HasExpectedResult
-                    && ParamAttributeTypeConversions.TryConvert(parms.ExpectedResult, method.ReturnType.Type, out expectedResultInTargetType))
+                    && ParamAttributeTypeConversions.TryConvert(parms.ExpectedResult, method.ReturnType.Type, out var expectedResultInTargetType))
                 {
                     parms.ExpectedResult = expectedResultInTargetType;
                 }
@@ -369,8 +368,7 @@ namespace NUnit.Framework
             {
                 object arg = arglist[i];
                 Type targetType = parameters[i].ParameterType;
-                object argAsTargetType;
-                if (ParamAttributeTypeConversions.TryConvert(arg, targetType, out argAsTargetType))
+                if (ParamAttributeTypeConversions.TryConvert(arg, targetType, out var argAsTargetType))
                 {
                     arglist[i] = argAsTargetType;
                 }
--- a/src/NUnitFramework/framework/Compatibility/System.Collections.Concurrent/ConcurrentQueue.cs
+++ b/src/NUnitFramework/framework/Compatibility/System.Collections.Concurrent/ConcurrentQueue.cs
@@ -321,9 +321,7 @@ namespace System.Collections.Concurrent
             try
             {
                 //store head and tail positions in buffer,
-                Segment head, tail;
-                int headLow, tailHigh;
-                GetHeadTailPositions(out head, out tail, out headLow, out tailHigh);
+                GetHeadTailPositions(out var head, out var tail, out var headLow, out var tailHigh);
 
                 if (head == tail)
                 {
@@ -399,9 +397,7 @@ namespace System.Collections.Concurrent
             get
             {
                 //store head and tail positions in buffer,
-                Segment head, tail;
-                int headLow, tailHigh;
-                GetHeadTailPositions(out head, out tail, out headLow, out tailHigh);
+                GetHeadTailPositions(out var head, out var tail, out var headLow, out var tailHigh);
 
                 if (head == tail)
                 {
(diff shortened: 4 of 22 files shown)


oznetmaster:
Or perhaps a ValueTuple or Tuple?

jnm2:
@mikkelbu That's a good idea, I agree.

I wouldn't like Tuple because of my strong aversion to the names Item1, Item2, etc. In order to use C# tuples in any public method we'd need to drop our `net35` and `net40` builds because we definitely don't want to be defining our own public ValueTuple types.

oznetmaster:
Since we have decided that c# 7 is the standard, there is no reason to use Item1, Item2 ....

public (string s, int i, double d) method (bool escape) is perfectly valid, and desirable.

oznetmaster:
And as I previously indicated, I dislike the use of out declarations.

jnm2:
@oznetmaster The reason to use Item1, Item2 would be if we used Tuple (System.Tuple) which requires that. I'm explaining why that suggestion doesn't appeal to me.

C# language tuples such as `(string s, int i, double d)` require ValueTuple to be defined in an accessible location which requires either dropping `net40` and `net35` support since the other option is defining ValueTuple in NUnit framework ourselves—for various reasons, not something we want to get into. :)

oznetmaster:
I did it in CF.  It is not a big deal to add ValueTuple support.  CF obviously does not have the c#7 syntax support, but my CF NUnit passes all the ValueTuple based tests.

jnm2:
The first challenge is deciding how much of https://github.com/dotnet/coreclr/blob/b1a841249bd23d94a5c88684d57a0dbdd92cac86/src/System.Private.CoreLib/shared/System/ValueTuple.cs we want to or have to strip out including related types like `IValueTupleInternal`. So there's a one-time addition of boilerplate which isn't that bad.

The next real issue is that we must keep it internal, which means the compiler won't let us use it in public or protected APIs. If we expose it as public, people will be getting more than they bargained for, similar to the System.Linq issue which motivated us to drop net20. Along the same lines, we'd never take a dependency on the async polyfill in the framework library even though we do in the test project.

If we can't use it in public/protected APIs, I'd rather not use it in internal calls either. Dealing with ValueTuple in the framework library doesn't seem worth it when we can have a consistent approach by writing a six-liner struct.

oznetmaster:
This change seems to have been done with little or no discussion.

Reviewing the actual changes, I find them significantly more difficult to understand and maintain.

Even more so when the out declaration variable is referenced later in the code, and one goes to find where it is defined.

Am I the only one who actually works in an environment where the use of the out declaration construct is not allowed as an inferior coding practice?

jnm2:
@oznetmaster I'm sorry, I'll try to do better next time. What would you suggest in order to get the discussion you'd like?

> Am I the only one who actually works in an environment where the use of the out declaration construct is not allowed as an inferior coding practice?

This is quite a strong view, and that is fine. To be honest I wouldn't expect many folks to think this way based on the circles of discussion I've been exposed to.

oznetmaster:
Not everyone is jumping up and down about this feature.  For example (from articles on the subject):

> NB! Use this trick if you are using out variables near where they are defined. If you also need this variable in other blocks that follow the declaring block then better go with a traditional out variable.

> Inline declarations of out variables might lead us to cleaner code, but we may lose readability if these out variables are also used in other code blocks.

If the variable is perhaps used only once, and in the next statement, it is a useful concept. If the variable is used anywhere away from that close context, it is a maintenance disaster.

We just chose to not allow its use, so that the question of "how far away" was not one that anyone had to worry about.

jnm2:
If a variable is ever used far away from its declaration, that's a maintenance burden. No distances were lengthened in this PR. I think the idea remains to be supported that the burden is worsened by declaring it in an `if` statement. This is in line with the general direction of the C# language, since the language designers decided on inline declarations as a cornerstone of pattern matching.

oznetmaster:
Actually, I was slightly mistaken.

We allow the use of the inline declaration if the method is called within an if statement, and the variable is only used in that same if statement.

CharliePoole:
@jnm2 FWIW my impression is that @oznetmaster is correct about the change having been made with little or no discussion.

I followed the entire discussion although I didn't chime in. You had one team member approve the change and one who stated his disagreement, but did not formally ask for any modifications using GitHub review. Having one approval made it allowable (according to GitHub) to merge the change, but left that second member's disagreement off to one side.

Based on that, my own answer to how to have a better discussion would be "First, have a discussion." Assume people have a good reason to disagree and work to understand it. Don't make it a debate. Try to find out what it would take to get each person's agreement. Basic stuff really.

And you guys did, in fact, have a discussion on this PR. A really good one. Unfortunately it was about the tangentially related topic of tuples. :smile:

Hmm... I wonder if tangential discussions might not get in the way of discussing the actual PR. :smiling_imp:

jnm2:
That makes sense, I moved too fast. My thought process was that I CC'd Neil specifically to ask his opinion on the code and he commented adding onto Mikkel's comment. I took that to mean there was no disagreement, but what I should have done is waited and clarified. Apologies again! Thanks for chiming in.

My intent with every change was to improve readability which is a function not only of the code itself but of the team. So Neil, can you help me see how declaring variables in an `if` statement is a maintenance burden? I have some reasons (in my previous comment) that make me think it's not, and by the distance metric you mentioned, it seems like at least I haven't made the distances worse than they already were?

oznetmaster:
I said that declaring them in the if statement made sense if they were subsequently used in the same if statement:
`if (Int32.TryParse (s, out int val) && val > 23)`
for example.

jnm2:
Ah, in the condition.
So now we're at a point where I'm saying the whole C# feature makes sense and you're saying it rarely makes sense. I'm interested in why. I think the default position is to assume the C# language team generally takes an evidence-based approach when they add things like this, but regardless, I know what I specifically like about it: there's no need to take up a previous statement which breaks the flow while typing and requires you to use an explicit type declaration. Those are both things I find repetitive and onerous, i.e. a maintenance burden, mostly during initial typing and during refactoring.

On the other hand, C# was a great language even in version 2, so it's not like I'm going to die if I can't use the new syntax.

That's why I think it makes more sense to declare online. Why do you think it makes more sense to declare on a previous line?

rprouse:
I have trouble seeing how this is harder to maintain. In every case, the variable was declared on the line before it was used. To me, it is no harder to read and maintain than switching;

```c#
int i;
i = 0;
```

To `int i = 0;`

Some people used to like all of their variables declared at the beginning of a method. Opinions changed over time and most people declare variables as close to where they are used as possible now. To me, this is the same sort of thing. There will be friction and some people may prefer the other style, but it is a matter of style and team preferences. I like the new style, others don't. That is okay.

As for switching `out` variables over to `ValueTuples`, I'd agree except I'd prefer not to pull in more 'Compatibility' classes than we need to. I think it makes sense to pull them in for functionality like threading and concurrency improvements, but not for stylistic choices.

Back to maintainability, I also find the statement "If the variable is used anywhere away from that close context, it is a maintenance disaster." a bit strong. The declaration of the variable moved one line in all cases and even if it is a bit harder to see, in both Visual Studio and VS Code, F12 jumps you to the definition of the variable from anywhere making it easy to find.

The one thing that I do think might have been a mistake (but once again, it is a style choice) is switching the out variables to use `var` because it is difficult to deduce the types in this usage. Even then, hovering over the variable anywhere in the code shows me the type in VS Code

image

and in Visual Studio

image

Which means I can fairly easily know the type and never actually have to find the definition.

Lastly, none of this is external code or APIs, so it is just the people working on the code that must work with it. I realize that you are a special case @oznetmaster because you are maintaining a port of the code, but we're spending way more characters discussing this than were even changed 😉

oznetmaster:
I, like many others, do a large amount of code review on my iPad (or even iPhone), so these tools are not available.

Good old fashioned reading and scrolling are pretty much it.

mikkelbu:
Or reading/reviewing the code on GitHub.

oznetmaster:
Yes, that too!

oznetmaster:
@rprouse I am actually maintaining two ports of the code now!

The one on CF is not even relevant, since that is C# 3.

The one on a Linux platform is relevant, but my comments still apply.

rprouse:
@oznetmaster, a total aside, but may I ask which Linux platform you are maintaining a port for and why you cannot use our repo as-is or the releases we distribute? Is it to compile packages for that platform, or is it a platform that compiles all packages from source? Either way, do our build scripts not work on that platform?

oznetmaster:
Because, as in the case of the CF build, which is no standard CF, but a Crestron based modified CF, the Linux environment is also a very specific Crestron based framework.

Much less modification required then in the CF case, but still will not run a standard NUnit build inside of the Crestron framework.

jnm2:
I heard this month that GitHub was integrating go-to-definition, but I can't find that now. Anyway, let's assume that you're reading code without any IDE features like that. It's likely that the first assignment is more interesting to follow than the declaration itself. Would placing the declarations on the previous line (which also moves them further from their usage) improve that experience, and how would it do that?

oznetmaster:
One reason (of many) is that a separate declaration allows the declaration to have a comment, which is not really possible/readable inline.

CharliePoole:
Seems to prove you can only discuss this usefully on a case by case basis.

jnm2:
Yes, that would be helpful. @oznetmaster (or anyone), could you comment on the line in https://github.com/nunit/nunit/pull/3272/files that is the best example of how maintainability decreased, and we can go from there?

src/NUnitFramework/tests/TestUtilities/Comparers/ObjectToStringComparer.cs
```diff
@@ -37,10 +37,8 @@ public class ObjectToStringComparer : System.Collections.IComparer
         int System.Collections.IComparer.Compare(object x, object y)
         {
             WasCalled = true;
-            int intX = 0;
-            int intY = 0;
 
-            if(int.TryParse(x.ToString(), out intX) && int.TryParse(y.ToString(), out intY))
+            if (int.TryParse(x.ToString(), out var intX) && int.TryParse(y.ToString(), out var intY))
```
CharliePoole: I'd use `int` here rather than `var`, but I don't think you have made it any less clear WRT the use of `out` either here or anywhere else.

IMO the high level of discussion comes from the fact that by making a global stylistic change you __appear__ to be defining a new standard of how to code, even though it isn't actually proposed as a standard and (again IMO) shouldn't be. Just a comment.

jnm2: `var` back to `int`: do you think it's worth a new PR? Happy to do that.

I'm glad you're making that distinction. This is not a standard. Along similar lines, this way future writers seeing this code nearby would hopefully conclude that they now have either option open to them rather than just the old way.

CharliePoole: I don't think it's worth it, personally, but up to you.

PR #3277, opened 2019-06-06
Add method to set Until date when ignoring TestCaseData

### LINKED ISSUE #3275: Enable setting IgnoreUntilDate in TestCaseData.Ignore
Currently, there is no easy method to set an `IgnoreUntilDate` property on a test case when using `TestCaseData`. Similar to the optional `Until` property on the `IgnoreAttribute`, an overload of the `Ignore(string)` method can be added to that takes in a reason string as well as an until date as a string. The signature would be something along the lines of:
```
public TestCaseData Ignore(string reason, string until);
```
Here's an example usage
```
private static IEnumerable<TestCaseData> _cases =  new [] { new TestCaseData("test").Ignore("Skip this test", until: "2019-06-30") };
```

abborg:
Resolves #3275 

Provide a simple method to set the IgnoreUntilDate on a test case defined by TestCaseData. The behavior is the same as `IgnoreAttribute` and only applies the ignored status if the test is not already set to `NotRunnable` and the `Until` date has not been reached yet.


src/NUnitFramework/framework/TestCaseData.cs
```diff
@@ -200,6 +201,40 @@ public TestCaseData Ignore(string reason)
...
+        /// <summary>
+        /// Ignores this TestCase, specifying the reason until the specified date.
+        /// </summary>
+        /// <param name="reason">The reason.</param>
+        /// <param name="until">
+        /// The date in the future to stop ignoring the test as a string in UTC time.
+        /// For example for a date and time, "2014-12-25 08:10:00Z" or for just a date,
+        /// "2014-12-25". If just a date is given, the Ignore will expire at midnight UTC.
+        /// </param>
+        /// <returns></returns>
```
mikkelbu: Major nit pick. Just remove this when it is empty.

src/NUnitFramework/testdata/TestCaseSourceAttributeFixture.cs
```diff
@@ -60,7 +60,11 @@ private static IEnumerable IgnoredSource
             {
                 return new object[] {
                     new TestCaseData(1),
-                    new TestCaseData(2).Ignore("Don't Run Me!")
+                    new TestCaseData(2).Ignore("Don't Run Me!"),
+                    new TestCaseData(3).Ignore("Ignore Me Until The Future", "4242-01-01"),
+                    new TestCaseData(4).Ignore("Ignore Me Until The Future", "4242-01-01 00:00:00"),
+                    new TestCaseData(5).Ignore("Ignore Me Until The Future", "4242-01-01 00:00:00Z"),
```
mikkelbu: Should we also have a test where the time is not `00:00:00(Z)`, just to ensure that time is not ignored.

abborg:
Thank you for reviewing the changes and providing feedback @mikkelbu! I'll make the amendments you noted.

rprouse:
This looks good to me too, but I agree with @CharliePoole about discussing the API. I am fine with adding an overload with a second parameter as it matches the `Ignore` attribute. I am not sure if the second parameter should be a string though. The `Ignore` attribute uses string as a workaround because attribute parameters must be `const`, but we don't have the same limitation with `TestCaseData`. 

I think the `until` parameter should be strongly typed as a `DateTime`. The advantage is that date format mistakes will be caught at compile time. The disadvantage is that constructing a new `DateTime` is more typing than a simple string. As it is, the code takes a string, tries to convert to a `DateTime` then converts back to a string in a known format. It also fails silently if the string parameter is in a bad format.

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
```diff
@@ -250,6 +259,24 @@ public string Category
             }
         }
 
+        /// <summary>
+        /// Gets and sets the ignore until date for this test case.
+        /// </summary>
+        public string Until
```
jnm2: Could this be `DateTimeOffset`?

@nunit/framework-team what do you think?

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
```diff
@@ -387,6 +414,20 @@ public IEnumerable<TestMethod> BuildFrom(IMethodInfo method, Test suite)
         {
             TestMethod test = new NUnitTestCaseBuilder().BuildTestMethod(method, suite, GetParametersForTestCase(method));
 
+            if (_untilDate.HasValue)
+            {
+                if (_untilDate > DateTimeOffset.UtcNow)
```
jnm2: Nitpick: `_untilDate.Value`

(Not necessary, but I'd write `if (_untilDate != null && _untilDate.Value > DateTimeOffset.UtcNow)`.)

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
```diff
@@ -387,6 +414,20 @@ public IEnumerable<TestMethod> BuildFrom(IMethodInfo method, Test suite)
...
+                {
+                    test.RunState = RunState.Ignored;
+                    string reason = string.Format("Ignoring until {0}. {1}", _untilDate.Value.ToString("u"), IgnoreReason);
```
jnm2: Please use the C# 6 interpolated string feature:

```suggestion
                    string reason = $"Ignoring until {_untilDate.Value:u}. {IgnoreReason}";
```

Is there anywhere else that we're building an "Ignoring until" message? If so, we should deduplicate as much as possible into one of our formatting helpers.

oznetmaster: “Please use the C# 6 interpolated string feature:”

Why?

This is not part of our code no standards, as far as I know.

Many coding shops have rejected the use of the interpolated string feature.

There is no reason to impose it here.

jnm2: @oznetmaster I don't agree with your assumptions and the options you're effectively leaving seem frustrating to me. Out of courtesy to Andrew, this is not a good place to have a long conversation. Would you be willing to discuss at https://github.com/orgs/nunit/teams/framework-team or over email? It would be appreciated.

@abborg No worries about the interpolated string :D

src/NUnitFramework/framework/Internal/IgnoredTestCaseData.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <returns>A modified TestCaseData.</returns>
+        public TestCaseData Until(DateTimeOffset datetime)
+        {
+            if (this._prevRunState != RunState.NotRunnable)
+            {
+                if (datetime > DateTimeOffset.UtcNow)
+                {
+                    this.RunState = RunState.Ignored;
+                    string reason = (string)this.Properties.Get(PropertyNames.SkipReason);
+                    reason = string.Format("Ignoring until {0}. {1}", datetime.ToString("u"), reason);
```
jnm2: We need to deduplicate this code. I just realized that this is a copy of the code in IgnoreAttribute.

@nunit/framework-team Any thoughts on where we should keep common code like this?

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
rprouse: I would agree that this is not the place for discussions like this. I'd like to keep PRs on-point and not distract from the code being discussed. Deeper coding philosophical issues can be discussed in other channels.

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
rprouse: I don't think this can be a `DateTimeOffset` because parameters in an attribute constructor must be `const`. Same as in the `IgnoreAttribute`.

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
rprouse: Very nitpick (even if it is the sort of thing I nitpick about) 😝

src/NUnitFramework/framework/Internal/IgnoredTestCaseData.cs
rprouse: I agree, but this PR has been out there for many months. Other than this, I think it looks good and can be merged. Maybe we leave the refactoring to another PR?

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
jnm2: Oh, of course... thanks!

src/NUnitFramework/framework/Internal/IgnoredTestCaseData.cs
jnm2: @rprouse That sounds good!

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
jnm2: Oh sorry, quite right!

rprouse:
@jnm2 would you still like to see changes to this? If not, I'll resolve conflicts and merge. This has been out there too long. I apologize for that @abborg.

jnm2:
@rprouse The copy/paste is the only outstanding thing that bothers me. https://github.com/nunit/nunit/pull/3277#discussion_r327378300

rprouse:
This PR has been superseded by #3544.

Issue #3276, opened 2019-06-06
Aborting test run with `force` true causes apparent hang


CharliePoole:
Found this in the GUI, see TestCentric/testcentric-gui#337.

There are a number of outstanding issues that may be related, but I decided to create this one anyway since it's specific to the `force` argument being true.

When we cancel with `force` false, everything works as expected. All ongoing tests and suites terminate and signal completion. The run is signaled as complete.

When `force` is true, the tests are aborted and signaled as complete but all ongoing suites are merely abandoned. No event is received for them so they appear to still be running. No run finished event is ever received.

Note that our existing tests around this all pass because the run is actually aborted. It just doesn't look that way to any runner that relies on event notifications, which is why I referred to an "apparent hang" in the title.

Existing issues, which may be related to this: #2947, #2282, #166 (five years old!). #166 is assigned to @oznetmaster, the others are unassigned.

oznetmaster:
As @CharliePoole pointed out, I first discussed this years ago.

Forced stop actually worked correctly when we had a separate thread for each test instance.  It needed a few tweaks, perhaps.

It stopped working when the decision was made to stop using a different thread for each instance.

I pointed that out at the time, but the comments from others indicated that no one really thought that was an important consideration.

CharliePoole:
Actually, all the tests __are__ being stopped and tearing is skipped as required. Problem is that no events are being received to record termination of the suites.

CharliePoole:
Hmm... looking back at #166 that's the same behavior we are seeing now. Tests are cancelled but we get no notifications. I left a comment for @oznetmaster on #166 to please remove himself if he isn't actually working on it.

CharliePoole:
OK, I think I see one source of the problem. `WorkItem`'s Cancel method is cancelling the thread on which Cancel is called. The intent is clearly that it would be cancelling the thread on which it is executing, but the method is now called from various places, on various threads. Cancelling the thread of your caller can't be good! :cry: We want to cancel the threads executing tests!

This explains why, after the initial tests are cancelled, additional code that is intended to clean up after the threads runs into the `ThreadAbortException`. We need to explicily abort the threads of the executing workers, not whatever thread happens to make the request.

CharliePoole:
I noticed that I self-assigned this in error last year, while I was traveling in Europe!

CharliePoole:
@nunit/framework-team 

This issue has come up again in TestCentric/testcentric-engine#39.

[...]

PR #3281, opened 2019-06-13
Fix forced StopRun

CharliePoole:
A start on issue #3276.

I made this a WIP PR with failing tests so we could all see where the problem is and what might be involved in fixing it. Also, to get some ideas for moving further. Here's what I know.

1. When a runner uses `StopRun(true)` to force cancelling the run, the system hangs.

2. We never knew this for two reasons:
  * We don't use this call in the console runner
  * Our tests of this feature were never actually testing anything because the `StopRun` call occured before the run even got started. I used a `SpinWait.SpinUntil` to fix that and get the tests to fail.

3. All running test cases were properly terminated. That code seems solid.

4. The containing fixtures for such test cases, however, are never terminated. That's because we enqueue a special `WorkItem` for one time teardown after the test workers have stopped processing the queue. I fixed this in `CompositeWorkItem`.

5. However, any such special one time teardown work items that were enqueued __before__ we stopped the run will never be executed. I believe this is true whether the stop is forced or cooperative but I don't have a test for it because it involves a very narrow window and a race condition.

6. Higher level suites, such as namespaces and the assembly itself are never terminated. That's why we get the hang. I've taken the code to the point where the terminating fixture __tries__ to notify the higher level suite but an exception gets thrown because of the pending Thread abort.

Here's what I believe...

1. We may be able to craft some special-purpose code to ensure that all in-flight suites terminate. Such code is likely IMO to lead to more technical debt in the dispatcher.

2. We probably need a full re-design of the parallel dispatcher that takes into account all the various addons that have happened since the first version. That version managed test cases very explicitly but allowed suites and fixtures to manage themselves WRT dependencies on completing children. I've always had a sense that we would eventually go to a more explicit model in the dispatcher, where every work item was tracked, along with it's dependencies, from the time of creation until it completed. Maybe that time is now.

I'm setting the issue to the side for now while I work on a workaround on the GUI side but if anyone has ideas about this I can come back to it and give them a try.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
@@ -382,8 +382,10 @@ namespace NUnit.Framework.Internal.Execution
         /// </summary>
         private void OnAllChildItemsCompleted()
         {
-            var teardown = new OneTimeTearDownWorkItem(this);
-            Context.Dispatcher.Dispatch(teardown);
+            if (Context.ExecutionStatus == TestExecutionStatus.AbortRequested)
+                WorkItemComplete();
+            else
+                Context.Dispatcher.Dispatch(new OneTimeTearDownWorkItem(this));
         }
 
         private readonly object cancelLock = new object();
--- a/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
@@ -28,6 +28,7 @@ using System.Reflection;
 using NUnit.Compatibility;
 using NUnit.Framework.Internal.Commands;
 using NUnit.Framework.Interfaces;
+using System.Diagnostics;
 
 namespace NUnit.Framework.Internal.Execution
 {
@@ -459,9 +460,6 @@ namespace NUnit.Framework.Internal.Execution
             {
                 lock (_teardownLock)
                 {
-                    //if (Test.Parent != null && Test.Parent.Name.EndsWith("nunit.framework.tests.dll"))
-                    //    System.Diagnostics.Debugger.Launch();
-
                     if (Test.TestType == "Theory" && Result.ResultState == ResultState.Success && Result.PassCount == 0)
                         Result.SetResult(ResultState.Failure, "No test cases were provided");
 
@@ -483,6 +481,20 @@ namespace NUnit.Framework.Internal.Execution
             /// PerformWork is not used in CompositeWorkItem
             /// </summary>
             protected override void PerformWork() { }
+
+#if PARALLEL
+            /// <summary>
+            /// WorkItemCancelled is called directly by the parallel dispatcher
+            /// when a test suite is left hanging after a forced StopRun. We
+            /// simulate WorkItemComplete() but without the ripple effect to
+            /// higher level suites, since we are controlling it all directly.
+            /// </summary>
+            internal void WorkItemCancelled()
+            {
+                Result.SetResult(ResultState.Cancelled, TestResult.USER_CANCELLED_MESSAGE);
+                _originalWorkItem.WorkItemComplete();
+            }
+#endif
         }
 
         #endregion
--- a/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
@@ -170,6 +170,8 @@ namespace NUnit.Framework.Internal.Execution
         /// </summary>
         private void PumpThreadProc()
         {
+            log.Debug("Starting EventPump");
+
             //ITestListener hostListeners = CoreExtensions.Host.Listeners;
             try
             {
@@ -188,6 +190,8 @@ namespace NUnit.Framework.Internal.Execution
                         log.Error("Exception in event handler {0}", ExceptionHelper.BuildStackTrace(ex));
                     }
                 }
+
+                log.Debug("EventPump Terminating");
             }
             catch (Exception ex)
             {
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -44,7 +44,7 @@ namespace NUnit.Framework.Internal.Execution
         private WorkItem _topLevelWorkItem;
         private readonly Stack<WorkItem> _savedWorkItems = new Stack<WorkItem>();
 
-        private readonly List<WorkItem> _workItemsInProcess = new List<WorkItem>();
+        private readonly List<CompositeWorkItem> _activeWorkItems = new List<CompositeWorkItem>();
 
         #region Events
 
@@ -226,8 +226,14 @@ namespace NUnit.Framework.Internal.Execution
         {
             log.Debug("Using {0} strategy for {1}", strategy, work.Name);
 
-            _workItemsInProcess.Add(work);
-            work.Completed += (s, e) => _workItemsInProcess.Remove((WorkItem)s);
+            // Currently, we only track CompositeWorkItems - this could be expanded
+            var composite = work as CompositeWorkItem;
+            if (composite != null)
+                lock (_activeWorkItems)
+                {
+                    _activeWorkItems.Add(composite);
+                    composite.Completed += OnWorkItemCompletion;
+                }
 
             switch (strategy)
             {
@@ -268,13 +274,17 @@ namespace NUnit.Framework.Internal.Execution
                 SpinWait.SpinUntil(() => _topLevelWorkItem.State == WorkItemState.Complete, WAIT_FOR_FORCED_TERMINATION);
 
                 // Notify termination of any remaining in-process suites
-                int index = _workItemsInProcess.Count;
-                while (index > 0)
+                lock (_activeWorkItems)
                 {
-                    var work = _workItemsInProcess[--index] as CompositeWorkItem;
-                    
-                    if (work != null && work.State == WorkItemState.Running)
-                        new CompositeWorkItem.OneTimeTearDownWorkItem(work).WorkItemCancelled();
+                    int index = _activeWorkItems.Count;
+
+                    while (index > 0)
+                    {
+                        var work = _activeWorkItems[--index];
+
+                        if (work.State == WorkItemState.Running)
+                            new CompositeWorkItem.OneTimeTearDownWorkItem(work).WorkItemCancelled();
+                    }
                 }
             }
         }
@@ -324,9 +334,20 @@ namespace NUnit.Framework.Internal.Execution
             }
         }
 
-#endregion
+        #endregion
 
-#region Helper Methods
+        #region Helper Methods
+
+        private void OnWorkItemCompletion(object sender, EventArgs args)
+        {
+            var work = (CompositeWorkItem)sender;
+
+            lock (_activeWorkItems)
+            {
+                _activeWorkItems.Remove(work);
+                work.Completed -= OnWorkItemCompletion;
+            }
+        }
 
         private void OnEndOfShift(WorkShift endingShift)
         {
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -27,6 +27,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
+using NUnit.Framework.Interfaces;
 
 namespace NUnit.Framework.Internal.Execution
 {
@@ -38,8 +39,12 @@ namespace NUnit.Framework.Internal.Execution
     {
         private static readonly Logger log = InternalTrace.GetLogger("Dispatcher");
 
+        private const int WAIT_FOR_FORCED_TERMINATION = 5000;
+
         private WorkItem _topLevelWorkItem;
         private readonly Stack<WorkItem> _savedWorkItems = new Stack<WorkItem>();
+
+        private readonly List<WorkItem> _workItemsInProcess = new List<WorkItem>();
 
         #region Events
 
@@ -221,6 +226,9 @@ namespace NUnit.Framework.Internal.Execution
         {
             log.Debug("Using {0} strategy for {1}", strategy, work.Name);
 
+            _workItemsInProcess.Add(work);
+            work.Completed += (s, e) => _workItemsInProcess.Remove((WorkItem)s);
+
             switch (strategy)
             {
                 default:
@@ -254,6 +262,21 @@ namespace NUnit.Framework.Internal.Execution
         {
             foreach (var shift in Shifts)
                 shift.Cancel(force);
+
+            if (force)
+            {
+                SpinWait.SpinUntil(() => _topLevelWorkItem.State == WorkItemState.Complete, WAIT_FOR_FORCED_TERMINATION);
+
+                // Notify termination of any remaining in-process suites
+                int index = _workItemsInProcess.Count;
+                while (index > 0)
+                {
+                    var work = _workItemsInProcess[--index] as CompositeWorkItem;
+                    
+                    if (work != null && work.State == WorkItemState.Running)
+                        new CompositeWorkItem.OneTimeTearDownWorkItem(work).WorkItemCancelled();
+                }
+            }
         }
 
         private readonly object _queueLock = new object();
--- a/src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
+++ b/src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
@@ -249,9 +249,9 @@ namespace NUnit.Framework.Api
             var explorer = _runner.ExploreTests(filter);
             Assert.That(explorer.TestCaseCount, Is.EqualTo(_runner.CountTestCases(filter)));
         }
-#endregion
+        #endregion
 
-#region Run
+        #region Run
 
         [Test]
         public void Run_AfterLoad_ReturnsRunnableSuite()
@@ -364,9 +364,9 @@ namespace NUnit.Framework.Api
                 Does.StartWith("Could not load"));
         }
 
-#endregion
+        #endregion
 
-#region RunAsync
+        #region RunAsync
 
         [Test]
         public void RunAsync_AfterLoad_ReturnsRunnableSuite()
@@ -447,7 +447,7 @@ namespace NUnit.Framework.Api
 
         #endregion
 
-#region StopRun
+        #region StopRun
 
 #if THREAD_ABORT // Can't stop run on platforms without ability to abort thread
         [Test]
@@ -507,9 +507,9 @@ namespace NUnit.Framework.Api
         }
 #endif
 
-#endregion
+        #endregion
 
-#region ITestListener Implementation
+        #region ITestListener Implementation
 
         void ITestListener.TestStarted(ITest test)
         {
@@ -569,9 +569,9 @@ namespace NUnit.Framework.Api
 
         }
 
-#endregion
+        #endregion
 
-#region Helper Methods
+        #region Helper Methods
 
         private ITest LoadMockAssembly()
         {
@@ -603,6 +603,6 @@ namespace NUnit.Framework.Api
                 "Parameter Y = 7" + Environment.NewLine));
         }
 
-#endregion
+        #endregion
     }
 }
--- a/src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
+++ b/src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
@@ -39,6 +39,7 @@ using NUnit.Framework.Internal.Filters;
 namespace NUnit.Framework.Api
 {
     // Functional tests of the TestAssemblyRunner and all subordinate classes
+    [NonParallelizable]
     public class TestAssemblyRunnerTests : ITestListener
     {
         private const string MOCK_ASSEMBLY_FILE = "mock-assembly.dll";
@@ -85,11 +86,6 @@ namespace NUnit.Framework.Api
             _failCount = 0;
             _skipCount = 0;
             _inconclusiveCount = 0;
-        }
-
-        public void DestroyRunner()
-        {
-            
         }
 
         #region Load
@@ -458,9 +454,11 @@ namespace NUnit.Framework.Api
         {
             new TestCaseData(0, false).SetName("{m}(Simple dispatcher, cooperative stop)"),
             new TestCaseData(0, true).SetName("{m}(Simple dispatcher, forced stop)"),
+#if PARALLEL
             new TestCaseData(2, false).SetName("{m}(Parallel dispatcher, cooperative stop)"),
 #if !NETCOREAPP2_0 // Hangs the CI build
             new TestCaseData(2, true).SetName("{m}(Parallel dispatcher, forced stop)")
+#endif
 #endif
         };
 
--- a/src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
+++ b/src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
@@ -26,6 +26,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
+using System.Text;
 using System.Threading;
 using NUnit.Compatibility;
 using NUnit.Framework.Interfaces;
@@ -72,6 +73,8 @@ namespace NUnit.Framework.Api
         private int _skipCount;
         private int _inconclusiveCount;
 
+        private Dictionary<string, bool> _activeTests;
+
         [SetUp]
         public void CreateRunner()
         {
@@ -86,6 +89,8 @@ namespace NUnit.Framework.Api
             _failCount = 0;
             _skipCount = 0;
             _inconclusiveCount = 0;
+
+            _activeTests = new Dictionary<string, bool>();
         }
 
         #region Load
@@ -442,8 +447,9 @@ namespace NUnit.Framework.Api
 
         #endregion
 
-        #region StopRun
+#region StopRun
 
+#if THREAD_ABORT // Can't stop run on platforms without ability to abort thread
         [Test]
         public void StopRun_WhenNoTestIsRunning_DoesNotThrow([Values] bool force)
         {
@@ -454,11 +460,9 @@ namespace NUnit.Framework.Api
         {
             new TestCaseData(0, false).SetName("{m}(Simple dispatcher, cooperative stop)"),
             new TestCaseData(0, true).SetName("{m}(Simple dispatcher, forced stop)"),
-#if PARALLEL
+#if PARALLEL // Currently, all THREAD_ABORT platforms are also PARALLEL, but just in case...
             new TestCaseData(2, false).SetName("{m}(Parallel dispatcher, cooperative stop)"),
-#if !NETCOREAPP2_0 // Hangs the CI build
             new TestCaseData(2, true).SetName("{m}(Parallel dispatcher, forced stop)")
-#endif
 #endif
         };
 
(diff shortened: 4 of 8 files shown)


rprouse:
As you say, there is already a lot of technical debt in the dispatcher and the queues. The idea of rewriting it scares me though. It took us quite a few releases to get it working and somewhat stable.

CharliePoole:
Maybe a feature switch? Create alt parallel dispatcher, make it pass the
same tests and eventually switch over.

CharliePoole:
OK, with the latest commit I expected to announce this was fixed. However the tests hang on all platforms in CI, while passing on my local machine.

@rprouse @jnm2 @mikkelbu @ChrisMaddock can you see anything that would leave them hanging? My assumption is that it's in the new `StopRun` tests.

Today is a driving day, moving from Umbria into Tuscany (Chianti), so I'll check back in the evening.

CharliePoole:
Temporarily commenting out the test of `StopRun(true)` allows the tests to complete successfully. Putting it back now.

CharliePoole:
OK, tests are now passing, including forced cancellation. Ready for review and/or discussion.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -382,8 +383,10 @@ private void OnChildItemCompleted(object sender, EventArgs e)
         /// </summary>
         private void OnAllChildItemsCompleted()
         {
-            var teardown = new OneTimeTearDownWorkItem(this);
-            Context.Dispatcher.Dispatch(teardown);
+            if (Context.ExecutionStatus == TestExecutionStatus.AbortRequested)
```
CharliePoole: If we're aborting, the one time teardown won't run so we don't enqueue it. Instead, we just mark the item as complete.

Come to think, maybe this should be a comment in the code.

src/NUnitFramework/framework/Internal/Execution/EventPump.cs
```diff
@@ -188,6 +190,8 @@ private void PumpThreadProc()
                         log.Error("Exception in event handler {0}", ExceptionHelper.BuildStackTrace(ex));
                     }
                 }
+
+                log.Debug("EventPump Terminating");
```
CharliePoole: These helped me figure out the problem. Leaving them in.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
```diff
@@ -254,6 +268,25 @@ public void CancelRun(bool force)
...
+
+                    while (index > 0)
+                    {
+                        var work = _activeWorkItems[--index];
+
+                        if (work.State == WorkItemState.Running)
+                            new CompositeWorkItem.OneTimeTearDownWorkItem(work).WorkItemCancelled();
+                    }
+                }
+            }
```
CharliePoole: This is the core of the fix. Dispatcher observes that the test suite threads didn't take care of completing themselves so it does it for them. IMO, active management by the dispatcher like this is really what we need, simplifying the code in the work items and centralizing the logic in the dispatcher. could be done incrementally or in one big bang by use of a feature switch.

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
```diff
@@ -433,71 +445,76 @@ public void RunAsync_BadFile_ReturnsNonRunnableSuite()
                 Does.StartWith("Could not load"));
         }
 
-#endregion
+        #endregion
```
jnm2: This doesn't line up with other region directives in the file anymore.

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
```diff
@@ -546,9 +569,9 @@ public void SendMessage(TestMessage message)
 
         }
 
-        #endregion
+#endregion
 
-        #region Helper Methods
+#region Helper Methods
```
jnm2: Somehow these traveled the opposite direction.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
```diff
@@ -221,6 +226,15 @@ private void Dispatch(WorkItem work, ParallelExecutionStrategy strategy)
         {
             log.Debug("Using {0} strategy for {1}", strategy, work.Name);
 
+            // Currently, we only track CompositeWorkItems - this could be expanded
+            var composite = work as CompositeWorkItem;
+            if (composite != null)
```
jnm2: You mentioned potentially wanting to use pattern matching. Maybe you already thought about it, but if you like you can do `if (work is CompositeWorkItem composite)`.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
```diff
@@ -221,6 +226,15 @@ private void Dispatch(WorkItem work, ParallelExecutionStrategy strategy)
...
+                lock (_activeWorkItems)
+                {
+                    _activeWorkItems.Add(composite);
+                    composite.Completed += OnWorkItemCompletion;
```
jnm2: Would it be possible to run this line before taking the lock?

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
```diff
@@ -254,6 +268,25 @@ public void CancelRun(bool force)
...
+                SpinWait.SpinUntil(() => _topLevelWorkItem.State == WorkItemState.Complete, WAIT_FOR_FORCED_TERMINATION);
+
+                // Notify termination of any remaining in-process suites
+                lock (_activeWorkItems)
+                {
+                    int index = _activeWorkItems.Count;
+
+                    while (index > 0)
+                    {
+                        var work = _activeWorkItems[--index];
```
jnm2: (Not important) Moving the decrement to a `for` loop is easier for me personally to parse than using the value of a decrement expression.

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
CharliePoole: I stopped cleaning these up when VS shifts them, because I understood they were all being removed at some point. Is that not the case?

jnm2: @ChrisMaddock offered to remove them all from the https://github.com/nunit/nunit-console repo in https://github.com/nunit/nunit-console/issues/635. If you like I can open an issue and probably mark it `good first issue` to do this same in this codebase.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
CharliePoole: I believe it is, but __only__ because we know how the containing method is being called. And it could eventually be called incorrectly. If that happened, we might end up with a race condition where the completion event that removes the item got called before the item was added. This way seemed safer.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
CharliePoole: Definitely stylistic, I think. My preference comes from having started writing pdp-11 assembly code, where auto-decrementing registers are a thing. Also, very popular in K&R C. :-)

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
CharliePoole: I've only paid attention to the new feature as it applies to catch statements. Didn't know the above syntax was possible. I'm not sure how much I like it as far as clarity goes.

Seems like moving the declaration inside the `if` saves a line and makes `composite` visible only in the following block - both good things. But having the declaration on the right-hand side of the `is` operator makes it much less visible and potentially confusing.

I'd have to study up on more examples of this kind of usage to get comfortable with it.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
jnm2: That sounds best.

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
CharliePoole: I wouldn't like to have a global edit issue out there as a `good first issue` unless we 100% agree that every single `#region` directive should be removed. As I expressed elsewhere, I'd rather not see global changes at all. And that's __even__ (in case it wasn't clear) if I agree with the change that is being made. I think global changes for conformity or consistency are evil!. :smiling_imp: 

OTOH if we had a standard that said NEVER use `#region` or even AVOID it, I'd be removing them in every file I touched, just like I add `_` in front of names. I think that's a better way to establish a team coding standard than doing blanket changes because it gets everyone's hands dirty, so to speak.

src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
jnm2: No problem. The C# team made the decision in the case of the `if` statement only,due to the prevalence of feedback they collected, that `composite` would be visible (but not definitely assigned) the same as it would be if it was declared on the previous line.

More at https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching#the-is-type-pattern-expression

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
jnm2: If we're discussing personal preferences, far from being evil, I think consistency is lovely because it puts the focus on the logic itself.

> I stopped cleaning these up when VS shifts them, because I understood they were all being removed at some point. Is that not the case?

If you don't want us to remove them as a global edit, and you don't want to remove them as you touch these files, would you mind making the indentation consistent where you are touching them?

CharliePoole: Sorry, I thought we were done and merged. 

To be clear, I __would__ want to remove them in this PR if that were the team standard. I'd want to have the wording first... is it NEVER or AVOID?

I do admit to not wanting to spend time on something the IDE keeps changing on me.

CharliePoole: Self-consistency is great. Imposed consistency (by or on others) not so great. Consistency that arises out of a team is best of all but hardest to achieve.

jnm2: @rprouse Do you want to make region removal a team standard so that Charlie feels comfortable removing them in this PR? Or would you want to do the same thing Chris did in nunit-console?

CharliePoole: @jnm2 That would be "imposed" of course. That is, whether Rob sets it as a team standard OR decides to make the change himself as Chris did, it would still be happening without a discussion. So be it.

But even so it still requires a clear statement of what the standard actually is. I'm using the MS terminology above: ALWAYS, USUALLY, AVOID, NEVER, which I think is a nice way to clarify a standard. It can be done in other ways, of course, but putting it into words is often a way to find out how clear your thinking is on the issue.

Sorry for my confusion about this PR, which I thought we had merged already. It was the related GUI PR that I merged. Obviously, this one is unmerged,  even though approved. Since you're concerned about the `#region` issue, I'll wait to merge. Gotta say though, I think that being able to cancel tests ought to have priority over an issue like this one.

jnm2: I'm finding this very complex and trying to do the complexity justice, but there are foure paths:

- Squashing out the region indentation to fix the inadvertent changes. I avoid asking for this because it's a bunch of additional steps.
- Removing the regions in this PR that were touched. You said you'd do this if we had a standard with the terminology AVOID or NEVER, which means this option is now blocked unless we wait for a team discussion.
- Leaving the autoindentation thrashing added in this PR, and opening an issue so we remember to remove all regions. You said you thought we were already going to remove all regions, and then that you would not like me to open an issue to track this and that you're against global changes.
- Leaving the autoindentation thrashing added in this PR and forgetting about it. 

Of *course* being able to cancel tests ought to have priority over formatting. By attaching complexity, which I don't agree with but which we're both loathe to turn into a metadiscussion, you have systematically left no path open other than disregarding my comment. This is fine; I wouldn't have approved if I thought it was important. I just wish you didn't make it sound like I'm suggesting that we prioritizing style discussions over real bug fixes.

CharliePoole: Well Sh*t, I typed a comment and then unresolved the conversation and my comment disappeared. I __hate__ this IDE. I'll start over when I calm down!

CharliePoole: So a couple of points...

 * Editing the `#region` statements that were changed by VS. You seem to have misunderstood me. I said I __couldn't__ do it because the PR was merged. I was under the impression you had merged it and I wasn't looking at it's status. I actually mentioned this in two comments. I was about to remind you of it again when I noticed that the PR had __not__ been merged. So I made the change you asked for as I would naturally do for any reasonable change request.

I __thought__ a global decision had been made to remove `#region` statements. You corrected me and said it was only made for nunit-console... at least that's what I think you said. But then you asked me to make the change in the files I had touched anyway. That was a really strange request, particularly since it's a change I would argue against in any discussion we might have.

That said, my views about `#region` are pretty flexible compared to my views about how decisions ought to get made and documented.

Finally, it's not __I__ who "makes it sound like" you prioritize style discussions over bug fixes. I'm just echoing back what I get from what your comments focus on and from your having referred the PR to @rprouse  to resolve.

CharliePoole:
@jnm2 We're waiting for @rprouse to respond to your request for guidance about `#region` statements. Fortunately, I can use this build for my own purposes while we wait to resolve that, so I'm not blocked.

oznetmaster:
Once again, I agree with @CharliePoole about the pattern matching.  Just because the C# team decided to add a feature does not necessarily make it desirable.

As with the out declaration, having the variable declared on the right hand side of the is operator reduces visibility and increases the difficulty of maintenance.

We have also decided to not allow its use as part of out programming standards for those reasons.

Again, this, like so many other things, is just a preference.

CharliePoole:
@oznetmaster Because I worked as an external, short term coach, I have interacted with a lot of teams. One of the biggest things I've had to learn and get them to learn is that there are lots of things that don't need to be standardized.

One team felt they had to make a statement, one way or another, as to whether the ternary `?:` operator could be used in their code. Getting them to simply __not say anything about it__ was incredibly hard, but in the end that's what they did.

Then, in code review, they gradually developed a feel for when it's use was clear and when it obfuscated the intention of the code. 

Of course, even worse than a team setting too many standards, is a manager who does it. :smile:

oznetmaster:
I really do not want to see these changes held up for a discussion of the #region issue.

Perhaps I am being premature, but I am going to approve the merge.

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
jnm2: It sounds like at least some of the conversation was talking past each other. We have some very different sets of expectations about what things mean. As far as my comments focusing on style, I only continued replying because you seemed to be suggesting options or requirements that needed a response. It was frustrating to hear that from your point of view I was putting style above getting this PR through. If you were saying I should just drop it, I apologize for missing that.

I did not refer this PR for Rob to resolve; I asked him question about our future plans to see which course to take out of the options you were presenting. This is again frustrating; you're attributing things to me that I was not aware of. Why must our conversations be this way?

jnm2:
@CharliePoole I did not (knowingly) ask for this PR to wait for Rob, which I also clarified just now in https://github.com/nunit/nunit/pull/3281#discussion_r296919081. Like I said earlier in this PR, holding up a bugfix for style reasons is something I strongly disagree with.

CharliePoole:
@oznetmaster I'm a bit dismayed. Our long-standing tradition has been that if anyone expresses an objection we don't move ahead unless the project lead resolves it. We give everybody very broad power to merge things because we trust them to voluntarily hold back in using that power when someone has raised an objection. I was recently in a conversation where you were the person whose objections had been ignored and I took the same view there.

All that said, @jnm2 is saying he wasn't intending to raise an objection to the PR so I'm going to just forget about this and chalk it up to experience.

CharliePoole:
@jnm2 You did say that you would not hold up a bug fix for style reasons. As people often do, you seemed to be doing the opposite of your words.

In any case, @oznetmaster has merged the fix. I had meant to put back [WIP] but didn't.

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
jnm2: Let me swallow my frustration and put this on myself, since as far as I know that's where the problem is happening. It's probably fair to say this would be compatible with your perspective on the conversation. 😊

Do you have any tips—besides trying to be underfoot less—for how I can better listen or better communicate my thoughts? If I don't figure something out that I can act on to improve my communication, I'm afraid I face two unpleasant options: engage and frustrate you, or withdraw at a hint of disagreement which seems almost passive aggressive and is in any case naturally hard for my personality. Both options seem like things you wouldn't want either.

jnm2:
> You did say that you would not hold up a bug fix for style reasons. As people often do, you seemed to be doing the opposite of your words.

It is possible to hold a consistent perspective of the entire conversation in which I was trying to avoid holding up the PR. I'm not sure it's worth me walking back through the conversation at this point. Please do give me the benefit of the doubt if it doesn't make sense to you.

oznetmaster:
@CharliePoole I have always been under the impression, perhaps incorrectly, that style issues were there for discussion, but that they, in and of themselves, were not a reason for merging or not merging changes.  I was of the same impression as @jnm2 on this issue.

The above comments have left me even more confused on this issue.

jnm2:
@oznetmaster I wouldn't merge. This would only be for my own style suggestions.

oznetmaster:
I think I am lost in this discussion :(

I will clarify once.

I do not think that bug fixes should be held up for style issues in the bug fix PR.

I do not think that PRs should be created/merged solely for style issues without a full discussion, and approval of the project leader.

jnm2:
@oznetmaster I wouldn't merge. This would only be for my own style suggestions. I *would* take explicit approval literally as you did, though. That is the reason I initially and later used the explicit Approve option to communicate that I was happy for the current code to go in. I'm glad you acted on it.

For future reference, I will explicitly request changes using the review mechanism if I wan't to block the PR.

oznetmaster:
On a more positive note, @CharliePoole I merged your changes unchanged into my CF version, and they worked perfectly, passing all of the TestAssemblyRunnerTests on first try.

Good work!

jnm2:
I edited poorly, I'm sorry. What I meant to say was that I wouldn't merge based on whether or not an issue had to do with style, but rather based on whether or not anyone was blocking or approving the PR using the review mechanism.

In this case, I approved to explicitly signal that I thought the code was good to merge. Charlie thought so too, because he submitted it as such, and you were the third person to think so. Therefore merging was fine from my understanding.

oznetmaster:
@jnm2 merging seemed to be the correct thing to do, from my understanding as well.

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
CharliePoole: I didn't think I was presenting options or making things more complex. I just saw myself as giving you reasons why I should not remove the #region statements, something which I both disagree with on it's merits and don't think should be part of this PR in any case. I thought that was pretty clear.

My suggestion is not to bring things up in a PR unless they impact the PR's approval. If you pull the project lead in, that's presumably to resolve a disagreement between us about the PR... because we are in a code review of the PR, not a general discussion! When you're the reviewer, you're the authority, and the project lead is an even bigger authority.

Frankly, at this point, I'm shocked that the PR has been merged without any discussion of the actual code changes. I'd like to think it was so good that nobody could find anything to change, but that's unlikely. I feel like all the attention went to the #region stuff and not to the meat of the actual change.

CharliePoole:
@jnm2 That may be another difference. I dislike the GitHub approval mechanism and don't treat it as "ruling" if there are any contradictory comments at all. I may be an anachronism in this.

Easiest for me would be if there were no idle chatter in a PR discussion. We have so many places to talk about things, why clutter PRs with things that are not part of the approval process. Also, after I've spent a lot of effort getting everything to work, I'd kind of like the discussion to be about my work, not something extraneous.

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
jnm2: My comment asking whether we could move the event subscription out of the lock doesn't count? All my attention did not go to the regions. They were an afterthought. There was nothing fundamental I saw that should change.

jnm2: I'll keep those things in mind for your next PR, and I'll clarify to counter the presumption if I do want to tag Rob (I don't know the reason I've tagged him was ever in order to arbitrate a disagreement). Thanks for your advice!

My comment asking whether we could move the event subscription out of the lock doesn't count? All my attention did not go to the regions. They were an afterthought. There was nothing fundamental I saw that should change.

oznetmaster:
I will repeat for clarity.  I saw @CharliePoole asking for @rprouse input as a request for discussion solely on the #region issue.

@jnm2 had approved the request.  I had merged the code into my CF build and ran the tests and they ran perfectly.  Therefore, it seemed that approving the request was the appropriate thing to do.  I basically ignored the discussion of stylistic issues in deciding to approve the request.  They did not seem relevant to the issue the PR addressed.

CharliePoole:
@oznetmaster Neither you not I are in a position any longer to decide how this project conducts reviews. But I'm the author of a PR and I'm certainly going to speak up if the way it's done affects me negatively.

Having some general discussion in the PR thread is bearable if every comment is so labelled. For example "OK to merge this but I think we should... whatever it is." Without that the proposer can't be sure what to do. In my view, it's easier to start a new conversation about the unrelated matter.

I find this particularly hard personally because GitHub will let me do anything, with or without approval, on any of the NUnit projects. I'm constantly watching myself to be sure I don't overstep and do something inappropriate. I hope that helps explain my shock at your merging the PR, a shock which I'm long over.

oznetmaster:
@CharliePoole If you felt that this merge was incorrectly done, why didn't you just roll it back?

I am sure that @jnm2 and I would have not taken personal offense at that action, given your distress at it having been done.

You still can, if you feel that it is not ready to be merged after all.

Do you feel that the code was not actually ready to be merged, even though @jnm2 and I reviewed the code, and I actually merged it into my CF version (which is historically problematic with test execution parallelism), and fully tested it with your new tests, and it worked perfectly?

It seems that you are upset about the lack of additional comments on the code itself.  I certainly had none, and as usual, I had to carefully review every line, and migrate it to CF and C# 3.  I had no further comments, except some possible additional style ones that were unrelated to the bug being fixed, and given the existing discussion, seemed unnecessary for now.

CharliePoole:
@oznetmaster I have a sense you are not reading my words. I thought I pretty clearly said I was surprised at your taking that liberty, when even I didn't feel free to merge. I also said I was over it.

I didn't merge or rollback, because I don't think that's within my rights according to the team's protocols. Again, as you say I _could__ have done it, that is, GitHub would let me do it, but that's not the point because we have only very broad control over what people can do in GitHub.

As I also said, the __only__ issue I felt was remaining was that I had been asked (not by you, obviously) to remove all the `#region` directives. I didn't want to make that change. However, I could not just ignore it so long as @jnm2 continued to want it done because those are the rules we have agreed to play by.

Lack of comments is not an issue. The fact that about 90% of the discussion had to do with #region directives is. Again, that's clearly not on you. I'm sorry, but also surprised that, you took it that way.

This is the second comment I've written to try to make it clear that __you__ are not the source of the problem I'm having with the review process. If this doesn't make it clear, it will still have to be enough. My issues with others have moved to private conversation and won't reappear here.

jnm2:
The hangs we're seeing at work still repro with https://www.myget.org/feed/nunit/package/nuget/NUnit/3.13.0-dev-06339.

CharliePoole:
I'd expect that unless they are hangs that occur when forcibly cancelling the run, which our console runner never does. I tried to be very careful that the new behavior should __only__ come into play when `StopRun(true)` was called.

jnm2:
Oh, I thought the console did pass `true`. Thanks!

CharliePoole:
The console never calls `StopRun` at all. The user has no way to cancel a run other than ctrl-C. In a traditional DOS environment, we would handle the break key to stop the run, but we didn't see the need to do so here.

oznetmaster:
@CharliePoole What I am confused about is that as far as I know, I was following our protocol exactly in approving the merge. @jnm2 appears to believe that was the case as well.

If that was not the case, then I need to understand what I did wrong.  Has the protocol changed?  Was there something I should have seen to stop me from approving the merge (in the past, it was "do not merge" in the title or comment)?

That is the only reason I am following this up.  No other agenda.

CharliePoole:
@oznetmaster For my view if what the protocol ought to be, see my initial comment to you. Clearly that's not the protocol any longer and @jnm2 seemed to agree with your understanding of it. I no longer have a voice in deciding what it actually is and I'm actually as confused as you are, so it's better if I stop expressing opinions on it. For the last word @rprouse would be the person to ask.

jnm2:
Part of it is that I view a reviewer's choice of approval or requesting changes as a highly visible, carefully chosen statement, equivalent to the boilerplate statement they'd otherwise have to write.

Of course I'll do my best to read the comments and notice if they seem to have changed their minds in the meantime. But without a clear, explicit statement in balance against the clear, explicit choice of approval or requesting changes, I'm likely going to respect their initial strong signal via review mechanism at face value exactly as though they had worded it manually.

jnm2:
Part of it is that I view a reviewer's choice of approval or requesting changes as a highly visible, carefully chosen statement, equivalent to the boilerplate statement they'd otherwise have to write.

Of course I'll do my best to read the comments and notice if they seem to have changed their minds in the meantime. But without a clear, explicit statement in balance against the clear, explicit choice of approval or requesting changes, I'm likely going to respect their initial strong signal via review mechanism at face value exactly as though they had worded it manually. If a subsequent statement is such that I'm not sure if they've changed their mind, I'll ask first.

So maybe the confusion is due to neither Neil nor I reading into my comments that I thought the PR should be held up over this, while Charlie did read that into my comments. In this case it seems to be a simple misunderstanding, the kind which is guaranteed to happen in the future by the nature of communication. I will (and do) strive to be better at avoiding such misunderstandings. In the grand scheme, i view such misunderstandings like a stubbed toe. Annoying in the moment, but ultimately we have all options still in front of us and can put it in perspective.

CharliePoole:
@jnm2 Of course that's a valid way to view it and I have no doubt that's how you look at reviews you get and how you try to do reviews.

For other people, however, a mix of approval with suggested changes may have another meaning. Perhaps the comments, since they have more effort behind them, seem more important than the simple selection of a dropdown.

One side issue with the approval versus change request is that we don't all always use care in picking one of them. You hit a button without thinking and you are in a review, where you intended a comment. Whatever you do, GitHub doesn't let you change it so you have to either do a new review or use your words to say what you want to have happen. I favor using your words.

I favor this both in my personal preference and in the style of teamwork I teach XP teams. You are right to say that communication is fraught with possible mistakes. That's why we need to do it a lot. When people have a chance to check a box instead of doing actual communication with team members, they usually choose the checkbox. I try to remove that option in my teams, which is why I don't use any automated approval mechanism.

GitHub's mechanism doesn't even match our protocol, which requires two approvals of every merge. In the case where the submitter is a team member, we only need one (more) approval so GitHub works for us. In the case where it's an outside contribution, we are forced back to actually looking at the details.

That said, I'm willing to follow whatever the team policy is on how reviews are conducted. Now I know your personal preference but I don't know any more than I did before about what that policy is. If it's "Whenever somebody selects approve for merge, then you may merge, no matter what other comments there are" then I can do that. If it's "Read the comments and make a judgement as to whether each item is something that is intended to hold up the merge" I don't think that's a reasonable expectation. None of my stuff will get merged in that case because I will always err on the side of __not__ merging so as not to overstep my bounds.

PR #3291, opened 2019-06-24
Use pattern matching and throw expressions

jnm2:
This pull request uses the C# 7.0 pattern matching feature to reduce repetition and clarify code. I wrote these changes in the original pull request before splitting into multiple pull requests.

The pattern matching language feature is built around the concept of declaring new variables in a very similar manner to `out` declarations which we are still discussing in https://github.com/nunit/nunit/pull/3272. Therefore, I applied the **awaiting:discussion** label so that we don't merge until we've discussed whether we as a team want all, some, or none of these changes. cc: @rprouse 

@oznetmaster Like the other PR, I'm tagging you to get your feedback on the code itself since you've mentioned disliking certain (most?) uses of it. I'd decidedly rather not introduce changes that make you hate the codebase.

### DIFF
--- a/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
+++ b/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
@@ -138,8 +138,7 @@ namespace NUnit.Framework.Api
 
                 if (options.ContainsKey(FrameworkPackageSettings.TestParametersDictionary))
                 {
-                    var testParametersDictionary = options[FrameworkPackageSettings.TestParametersDictionary] as IDictionary<string, string>;
-                    if (testParametersDictionary != null)
+                    if (options[FrameworkPackageSettings.TestParametersDictionary] is IDictionary<string, string> testParametersDictionary)
                     {
                         foreach (var parameter in testParametersDictionary)
                             TestContext.Parameters.Add(parameter.Key, parameter.Value);
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -421,8 +421,7 @@ namespace NUnit.Framework.Api
 
             if (value != null)
             {
-                var dict = value as IDictionary;
-                if (dict != null)
+                if (value is IDictionary dict)
                 {
                     AddDictionaryEntries(setting, dict);
                     AddBackwardsCompatibleDictionaryEntries(setting, dict);
--- a/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
@@ -196,8 +196,7 @@ namespace NUnit.Framework
                             // 3. An array was passed, it may be an object[]
                             //    or possibly some other kind of array, which
                             //    TestCaseSource can accept.
-                            var array = item as Array;
-                            if (array != null)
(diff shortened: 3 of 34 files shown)


CharliePoole:
I'd like to see a meta-discussion (although I usually hate those) either here or elsewhere about making global stylistic changes to the codebase. 

Why?
1. Because we have repeatedly said in the past - both among ourselves and to potential contributors - that we don't do that.
2. It spends energy without getting any actual features added or bugs fixed.
3. It bypasses the process of agreeing on the standards to be followed, instead establishing a _de facto_ standard, often in cases where we normally would decide on a case by case basis.
4. It tends to gives precedenc to one person's view, the person who sees the value in standardizing for it's own sake.
5. It propagates the fiction that there is a "best way" to code.
6. It can easily lead to thrashing in the code base as others change things back to the style that they prefer.

As for the feature, I think it's cool and I'll use it in my code where I would otherwise have a test on the exception, __except where I think the test is actually clearer__.

oznetmaster:
I totally agree with @CharliePoole

jnm2:
I'm not heavily invested, and having two of you who see this as problematic means it's probably not worth doing. To a few points, quickly:

1. I don't think we should view this as an end in itself. It seems good to be free to change this, or stick with it, based on the underlying reasoning.
2. This represents very little energy (5-10 minutes on my end) which would not have gone into an actual feature or bugfix.
3. [ignore, markdown autonumbering requires an item 3]
4. I want our codebase to give preference not to one person's view but definitely to the _team's_ view. This pull request is an efficient way to find out the preference we have as a team.
5. If I didn't think this syntax was an improvement, I wouldn't want to waste your time. :)

jnm2:
@nunit/framework-team I'm leaning towards closing since Charlie and Neil aren't in favor, but I'll keep it open for a few more days in case you would like to share your thoughts.

There are a couple refactorings which really could improve the code, unrelated to pattern matching. I'll submit them as tiny PRs later.

mikkelbu:
I don't mind global changes so much, and I also very much like consistency, but I agree with Charlie - and Neil - that things should be reviewed on a case by case basis.

Most of the changes in this makes the code easier to read *for me*, but there are some changes where the resulting expressions are quite hard to read - at least at a glance. Places where I prefer the original multiline version. It can be that the line becomes very long - with a "span" between important concepts, like `if` and `testParametersDictionary` in
```csharp
if (options[FrameworkPackageSettings.TestParametersDictionary] is IDictionary<string, string> testParametersDictionary)
```
or it can be that so much is happening on one line that it is hard to read the code, like
```csharp
var parms = item as ITestFixtureData
    ?? new TestFixtureParameters(item as object[] ?? new object[] { item });
```

This is probably also why I don't feel like this PR is trying to establishing a de facto standard, as I would not follow this standard in all cases.

Ps. There are also places where the diff is so large, like in EnumerablesComparer.cs, that I would rather keep the code as it is in fear for introducing errors, and instead make some smaller incremental changes.

ChrisMaddock:
I'm hoping to contribute less in the framework repo going forward, but I will chime in to say I personally like PRs such as this.

If someone is prepared to spend time doing the work, I prefer to work in a cleaner and more standardised codebase. Reviewing is a problem for significant changes, as are the frequency of large formatting PRs. I think this particular one is reasonable on both counts. 

Everything is case-by-case, of course. In this particular case, if this PR was to the console, I would merge this.

jnm2:
Thanks for chiming in, everyone! I'm going to close since it doesn't have full support.

PR #3298, opened 2019-07-02
Fix exception message formatting for Assert.Multiple

### LINKED ISSUE #3302: Incorrect formatting of failure message if test fails with Assert.Multiple
```csharp
Assert.Multiple(() =>
            {
                Assert.AreEqual(2, 3, "The first failed assert");
                Assert.AreEqual(4, 5, "The second failed assert");
                throw new Exception("And exception and the end.");
            });
```

XML report contains failure message with missing whitespace:
image

nvborisenko:
This PR fixes #3302.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -511,7 +511,7 @@ namespace NUnit.Framework.Internal
             if (AssertionResults.Count > 0 && result.ResultState == ResultState.Error)
             {
                 // Add pending failures to the legacy result message
-                Message += " " + CreateLegacyFailureMessage();
+                Message += Environment.NewLine + Environment.NewLine + CreateLegacyFailureMessage();
 
                 // Add to the list of assertion errors, so that newer runners will see it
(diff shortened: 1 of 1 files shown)


src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -511,7 +511,7 @@ public void RecordException(Exception ex)
             if (AssertionResults.Count > 0 && result.ResultState == ResultState.Error)
             {
                 // Add pending failures to the legacy result message
-                Message += CreateLegacyFailureMessage();
+                Message += " " + CreateLegacyFailureMessage();
```
CharliePoole: See my comment on the issue. This is obviously better than what we have, but I'm not sure it's enough.

CharliePoole: I agree with @ChrisMaddock that the simpler fix is best. Personally, I'd expect to see the legacy failure message under the original message with a blank line separating, i.e. 2 newlines rather than a space added.

jnm2: I would like to see two newlines as well.

Issue #3295, opened 2019-06-28
Reproducible hang with no user code when shutting down threads


jnm2:
I'm finally tackling a longstanding hang issue with our CI builds at work. What makes me sure that this is an NUnit bug is that whenever the tests hang and I attach a debugger, there is no user code in any thread's call stack. There are always between 5 and 7 worker threads, blocked on the following call:

```
mscorlib.dll!System.Threading.ManualResetEventSlim.Wait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken)
nunit.framework.dll!NUnit.Framework.Internal.Execution.WorkItemQueue.Dequeue()
nunit.framework.dll!NUnit.Framework.Internal.Execution.TestWorker.TestWorkerThreadProc()
mscorlib.dll!System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state, bool preserveSyncCtx)
mscorlib.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state, bool preserveSyncCtx)
mscorlib.dll!System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext executionContext, System.Threading.ContextCallback callback, object state)
mscorlib.dll!System.Threading.ThreadHelper.ThreadStart()
[Native to Managed Transition]
```

The fact that there are always fewer than 9 threads would seem to indicate that NUnit is in the process of shutdown from the normal eight plus an STA thread every time this happens. At least two worker threads have shut down in every case I've seen so far.

I have a project that reproduces the problem around 2% of the time. It's still 39 tests. I'm in the process of trying to remove things.

jnm2:
I really hope this isn't environmental. I was making steady progress removing things without losing the repro, but now I can't get it to repro anymore.

jnm2:
Okay, it's not that I'm losing the repro, thankfully. It's just getting rarer. Last cycle took an hour of back-to-back runs to reproduce.

jnm2:
Still removing minor parts of the remaining 39 tests but it's hard to do much without losing the repro. Verifying whether I still have it takes anywhere from a few seconds to not much more than an hour max. But an hour is a long time to wait for any feedback on what I just tried.

jnm2:
[...]

PR #3300, opened 2019-07-03
Prevent crash when exceptions are thrown by Exception members

### LINKED ISSUE #3296: ExceptionHelper.GetExceptionMessage(Exception ex) should tolerate exceptions from exceptions
We have a test which fails with buggy exception (ex.Message throws exception). When nunit tries to get exception details for test outcome messages, it leads to test runner crash, which results in test run stop (still visual studio test runner shows progress animation, but no test executed anymore).

Stack trace of original exception:
```cs
   at System.Collections.ArrayList.get_Item(Int32 index)
   at Sybase.Data.AseClient.AseErrorCollection.get_Message()
   at Sybase.Data.AseClient.AseException.get_Message()
   at NUnit.Framework.Internal.ExceptionHelper.GetExceptionMessage(Exception ex)
   at NUnit.Framework.Internal.ExceptionHelper.BuildMessage(Exception exception, Boolean excludeExceptionNames)
   at NUnit.Framework.Internal.TestResult.ExceptionResult..ctor(Exception ex, FailureSite site)
   at NUnit.Framework.Internal.TestResult.RecordException(Exception ex)
   at NUnit.Framework.Internal.Execution.SimpleWorkItem.PerformWork()
   at NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread()
   at NUnit.Framework.Internal.Execution.WorkItem.Execute()
   at NUnit.Framework.Internal.Execution.SimpleWorkItemDispatcher.Dispatch(WorkItem work)
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.RunChildren()
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.PerformWork()
   at NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread()
   at NUnit.Framework.Internal.Execution.WorkItem.Execute()
   at NUnit.Framework.Internal.Execution.SimpleWorkItemDispatcher.Dispatch(WorkItem work)
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.RunChildren()
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.PerformWork()
   at NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread()
   at NUnit.Framework.Internal.Execution.WorkItem.Execute()
   at NUnit.Framework.Internal.Execution.SimpleWorkItemDispatcher.Dispatch(WorkItem work)
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.RunChildren()
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.PerformWork()
   at NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread()
   at NUnit.Framework.Internal.Execution.WorkItem.Execute()
   at NUnit.Framework.Internal.Execution.SimpleWorkItemDispatcher.Dispatch(WorkItem work)
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.RunChildren()
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.PerformWork()
   at NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread()
   at NUnit.Framework.Internal.Execution.WorkItem.Execute()
   at NUnit.Framework.Internal.Execution.SimpleWorkItemDispatcher.Dispatch(WorkItem work)
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.RunChildren()
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.PerformWork()
   at NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread()
   at NUnit.Framework.Internal.Execution.WorkItem.Execute()
   at NUnit.Framework.Internal.Execution.SimpleWorkItemDispatcher.Dispatch(WorkItem work)
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.RunChildren()
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.PerformWork()
   at NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread()
   at NUnit.Framework.Internal.Execution.WorkItem.Execute()
   at NUnit.Framework.Internal.Execution.SimpleWorkItemDispatcher.RunnerThreadProc()
   at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()
```

Note that we cannot fix this exception class, because it is a third party library, which hardly be fixed by it's vendor.

Nunit version: 3.12.0 (used through nuget)

jnm2:
Fixes #3296.

This would also improve the test results in the https://github.com/nunit/nunit/issues/3295 situation if the NUnit assembly didn't have partial trust turned on.

### DIFF
--- a/src/NUnitFramework/framework/Assert.cs
+++ b/src/NUnitFramework/framework/Assert.cs
@@ -404,7 +404,29 @@ namespace NUnit.Framework
         private static readonly StackFilter SystemEnvironmentFilter = new StackFilter(@" System\.Environment\.");
 
         private static string GetStackTrace() =>
-            StackFilter.DefaultFilter.Filter(SystemEnvironmentFilter.Filter(Environment.StackTrace));
+            StackFilter.DefaultFilter.Filter(SystemEnvironmentFilter.Filter(GetEnvironmentStackTraceWithoutThrowing()));
+
+        /// <summary>
+        /// If <see cref="Exception.StackTrace"/> throws, returns "SomeException was thrown by the
+        /// Environment.StackTrace property." See also <see cref="ExceptionExtensions.GetStackTraceWithoutThrowing"/>.
+        /// </summary>
+#if !(NET35 || NETSTANDARD1_4)
+        // https://github.com/dotnet/coreclr/issues/19698 is also currently present in .NET Framework 4.7 and 4.8. A
+        // race condition between threads reading the same PDB file to obtain file and line info for a stack trace
+        // results in AccessViolationException when the stack trace is accessed even indirectly e.g. Exception.ToString.
+        [System.Runtime.ExceptionServices.HandleProcessCorruptedStateExceptions]
+#endif
+        private static string GetEnvironmentStackTraceWithoutThrowing()
+        {
+            try
+            {
+                return Environment.StackTrace;
+            }
+            catch (Exception ex)
+            {
+                return ex.GetType().Name + " was thrown by the Environment.StackTrace property.";
+            }
+        }
 
         private static void IncrementAssertCount()
         {
--- a/src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
--- a/src/NUnitFramework/framework/ExceptionExtensions.cs
+++ b/src/NUnitFramework/framework/ExceptionExtensions.cs
@@ -1,0 +1,102 @@
+// ***********************************************************************
+// Copyright (c) 2019 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Collections;
+
+namespace NUnit.Framework
+{
+    internal static class ExceptionExtensions
+    {
+        /// <summary>
+        /// If <see cref="Exception.StackTrace"/> throws, returns "SomeException was thrown by the Exception.StackTrace
+        /// property." See also <see cref="Assert.GetEnvironmentStackTraceWithoutThrowing"/>.
+        /// </summary>
+#if !(NET35 || NETSTANDARD1_4)
+        // https://github.com/dotnet/coreclr/issues/19698 is also currently present in .NET Framework 4.7 and 4.8. A
+        // race condition between threads reading the same PDB file to obtain file and line info for a stack trace
+        // results in AccessViolationException when the stack trace is accessed even indirectly e.g. Exception.ToString.
+        [System.Runtime.ExceptionServices.HandleProcessCorruptedStateExceptions]
+#endif
+        public static string GetStackTraceWithoutThrowing(this Exception exception)
+        {
+            if (exception is null) throw new ArgumentNullException(nameof(exception));
+
+            try
+            {
+                return exception.StackTrace;
+            }
+            catch (Exception ex)
+            {
+                return ex.GetType().Name + " was thrown by the Exception.StackTrace property.";
+            }
+        }
+
+        /// <summary>
+        /// If <see cref="Exception.Message"/> throws, returns "SomeException was thrown by the Exception.Message
+        /// property."
+        /// </summary>
+#if !(NET35 || NETSTANDARD1_4)
+        // https://github.com/dotnet/coreclr/issues/19698 is also currently present in .NET Framework 4.7 and 4.8. A
+        // race condition between threads reading the same PDB file to obtain file and line info for a stack trace
+        // results in AccessViolationException when the stack trace is accessed even indirectly e.g. Exception.ToString.
+        [System.Runtime.ExceptionServices.HandleProcessCorruptedStateExceptions]
+#endif
+        public static string GetMessageWithoutThrowing(this Exception exception)
+        {
+            if (exception is null) throw new ArgumentNullException(nameof(exception));
+
+            try
+            {
+                return exception.Message;
+            }
+            catch (Exception ex)
+            {
+                return ex.GetType().Name + " was thrown by the Exception.Message property.";
+            }
+        }
+
+        /// <summary>
+        /// If <see cref="Exception.Data"/> throws, returns "SomeException was thrown by the Exception.Data property."
+        /// </summary>
+#if !(NET35 || NETSTANDARD1_4)
+        // https://github.com/dotnet/coreclr/issues/19698 is also currently present in .NET Framework 4.7 and 4.8. A
+        // race condition between threads reading the same PDB file to obtain file and line info for a stack trace
+        // results in AccessViolationException when the stack trace is accessed even indirectly e.g. Exception.ToString.
+        [System.Runtime.ExceptionServices.HandleProcessCorruptedStateExceptions]
+#endif
+        public static Result<IDictionary> GetDataWithoutThrowing(this Exception exception)
+        {
+            if (exception is null) throw new ArgumentNullException(nameof(exception));
+
+            try
+            {
+                return Result.Success(exception.Data);
+            }
+            catch (Exception ex)
+            {
+                return ex.GetType().Name + " was thrown by the Exception.Data property.";
+            }
+        }
+    }
+}
--- a/src/NUnitFramework/framework/ExceptionExtensions.cs
+++ b/src/NUnitFramework/framework/ExceptionExtensions.cs
@@ -95,7 +95,7 @@ namespace NUnit.Framework
             }
             catch (Exception ex)
             {
-                return ex.GetType().Name + " was thrown by the Exception.Data property.";
+                return Result.Error<IDictionary>(ex.GetType().Name + " was thrown by the Exception.Data property.");
             }
         }
     }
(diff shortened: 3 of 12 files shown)


src/NUnitFramework/framework/ExceptionExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+    {
+        /// <summary>
+        /// If <see cref="Exception.StackTrace"/> throws, returns "SomeException was thrown by the Exception.StackTrace
+        /// property." See also <see cref="Assert.GetEnvironmentStackTraceWithoutThrowing"/>.
+        /// </summary>
+#if !(NET35 || NETSTANDARD1_4)
+        // https://github.com/dotnet/coreclr/issues/19698 is also currently present in .NET Framework 4.7 and 4.8. A
+        // race condition between threads reading the same PDB file to obtain file and line info for a stack trace
+        // results in AccessViolationException when the stack trace is accessed even indirectly e.g. Exception.ToString.
+        [System.Runtime.ExceptionServices.HandleProcessCorruptedStateExceptions]
```
jnm2: These `[HandleProcessCorruptedStateExceptions]` attributes have no effect because .NET Framework ignores them unless we disable partial trust. I added them for consistency since we have them other places.

src/NUnitFramework/framework/Result.cs
```diff
@@ -0,0 +1,84 @@
...
+namespace NUnit.Framework
+{
+    internal static class Result
+    {
+        public static Result<T> Success<T>(T value) => Result<T>.Success(value);
+
+        public static Result<T> Error<T>(string message) => Result<T>.Error(message);
+    }
+
+    internal struct Result<T>
```
jnm2: Nice ol' manual discriminated union I've typed many times.

src/NUnitFramework/tests/Internal/Results/TestResultApiTests.cs
```diff
@@ -75,5 +76,55 @@ public void DoesNotThrowForMissingInnerException()
...
+                    + "BrokenException was thrown by the Exception.Data property."));
+
+                Assert.That(result, Has.Property("StackTrace").StartWith(
+                    "BrokenException was thrown by the Exception.StackTrace property."));
+            }
+        }
+
+        private sealed class BrokenException : Exception
+        {
+            public override string Message => throw this;
```
jnm2: I couldn't figure out which was more evil, `throw this` or `throw new BrokenException();`

Both certainly got the point across 😁

image

ReSharper: "The active test run was aborted. Reason: test host process crashed."

src/NUnitFramework/framework/ExceptionExtensions.cs
CharliePoole: Maybe we need a more general discussion of supporting (or not) partial trust.

src/NUnitFramework/framework/ExceptionExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// If <see cref="Exception.Message"/> throws, returns "SomeException was thrown by the Exception.Message
+        /// property."
+        /// </summary>
+#if !(NET35 || NETSTANDARD1_4)
+        // https://github.com/dotnet/coreclr/issues/19698 is also currently present in .NET Framework 4.7 and 4.8. A
+        // race condition between threads reading the same PDB file to obtain file and line info for a stack trace
+        // results in AccessViolationException when the stack trace is accessed even indirectly e.g. Exception.ToString.
+        [System.Runtime.ExceptionServices.HandleProcessCorruptedStateExceptions]
+#endif
+        public static string GetMessageWithoutThrowing(this Exception exception)
```
jnm2: My idea was that by making these extension methods, we would notice them in intellisense and choose them over the properties in future code.

jnm2: (Meant to add this comment earlier, but I missed that there's no 'Add single comment' button in the mobile verson, so it was a pending review comment. Lame!)

src/NUnitFramework/framework/ExceptionExtensions.cs
jnm2: Yes! I first asked this question at https://github.com/nunit/nunit/issues/3295#issuecomment-507503500 but I'm half afraid it's too buried. Maybe I should move it to its own thread.

CharliePoole: Probably. We do bury stuff sometimes.

jnm2:
I just realized I didn't find all implicit calls to Exception.ToString() which can throw directly or indirectly via Message or StackTrace. `Assert.That(new BrokenException(), ...)` kills the test host. More coming...

jnm2:
@nunit/framework-team Alright, ready to go again now that `Assert.That` is hardened. If you can think of any other way Exception.ToString() might ever be called, please mention it!

src/NUnitFramework/tests/Internal/UnexpectedExceptionTests.cs
```diff
@@ -100,24 +97,125 @@ public void FailRecordsNestedInnerExceptionAsPartOfAggregateException()
...
+            var result = (ITestResult)null;
+
+            var ex = RecordPossiblyDangerousException(() =>
+                result = RunDataTestCase(nameof(UnexpectedExceptionFixture.AssertThatWithRecursivelyThrowingExceptionAsActual)));
+
+            Assert.That(ex is null); // Careful not to pass ex to Assert.That and crash the test run rather than failing
+
+            Assert.That(result, Has.Property("Message").StartWith(
+                "  Expected: null" + Environment.NewLine
+                + "  But was:  <! RecursivelyThrowingException was thrown by RecursivelyThrowingException.ToString() !>"));
```
jnm2: @nunit/framework-team Do you like the way this looks? I added `!` and an extra space on each side to indicate that this is not a normal situation.

src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
```diff
@@ -67,7 +67,7 @@ public void ApplyToTest(Test test)
                 catch (InvalidPlatformException ex)
                 {
                     test.RunState = RunState.NotRunnable;
-                    test.Properties.Add(PropertyNames.SkipReason, ex.Message);
+                    test.Properties.Add(PropertyNames.SkipReason, ex.GetMessageWithoutThrowing());
```
CharliePoole: Discuss: Should we use this technique for our own thrown exceptions? Or should we trust our own code and use this for outside exceptions? Which approach will make it easier to catch problems? Which approach is clearer? My reaction here was to go and try to figure out why you thought this might throw an exception, which is of course a waste of time.

src/NUnitFramework/tests/Internal/Results/TestResultApiTests.cs
```diff
@@ -22,6 +22,7 @@
 // ***********************************************************************
 
 using System;
+using System.Collections;
```
CharliePoole: Not used?

src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
jnm2: I didn't think about this, but of course this wouldn't throw an exception. It's not as though we were accessing ex.StackTrace here which might throw. I'll undo instances of GetMessageWithoutThrowing where the exception could not have been thrown by user code.

src/NUnitFramework/tests/Internal/UnexpectedExceptionTests.cs
mikkelbu: I think it is fine (and I cannot come up with anything better)

src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
mikkelbu: I would lean towards handling all the exceptions the same, as this would also work in the future if exceptions are changed. I see it similar to how we often have guards in the beginning of methods even though we now can be sure that the guards are satisfied by examining all call-sites.
But I can see Charlies' point that it can confuse the reader, if he/she tries to figure out why we think that the code might throw an exception.

CharliePoole: Something like GetMessageFromUserException might give a clearer signal.

src/NUnitFramework/framework/Assert.cs
```diff
@@ -404,7 +404,29 @@ private static void IssueWarning(string message)
...
+#endif
+        private static string GetEnvironmentStackTraceWithoutThrowing()
+        {
+            try
+            {
+                return Environment.StackTrace;
+            }
+            catch (Exception ex)
+            {
+                return ex.GetType().Name + " was thrown by the Environment.StackTrace property.";
```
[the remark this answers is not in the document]
jnm2: I thought about it. I'm sure this is safe from user code throwing an exception since the C# rules bind this call to `System.Object.GetType()` which is sealed.

I can't think of any normal exceptions the BCL itself could throw when GetType is called. A type load exception would have prevented the exception from even being instantiated. That leaves exceptions where the BCL is seriously broken, such as the race condition when creating stack traces that hasn't been patched yet in .NET 4.7.2 and 4.8. Those we should only plan for when we have a specific case because there are just too many ways things could go wrong in theory all over the place if the BCL is unreliable.

src/NUnitFramework/framework/ExceptionExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            if (exception is null) throw new ArgumentNullException(nameof(exception));
+
+            try
+            {
+                return Result.Success(exception.Data);
+            }
+            catch (Exception ex)
+            {
+                return ex.GetType().Name + " was thrown by the Exception.Data property.";
```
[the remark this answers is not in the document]
jnm2: Yeah, not my favorite. But I do find the explicitness of the `Result.Error` part to be desirable. I could make the syntax `Result.Error(...)` work with a new error type that implicitly converts to the generic type... 🤔

src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
jnm2: @CharliePoole I like the way you think. The reason I didn't go with a name like that is that I like the concept that the extension methods are divorced from the methodology they are used in. The name would indicate that they don't work just as well for non-user exceptions and that feels to me like it a confusing message.

jnm2: @mikkelbu Yes. After Charlie pointed it out, I also personally preferred to switch back to `.Message` so that the code isn't misleading (which maybe you saw). Just to confirm, sounds like you're happy with that for now?

src/NUnitFramework/framework/ExceptionExtensions.cs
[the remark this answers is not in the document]
jnm2: @mikkelbu What do you think of `Result.ErrorResult`, better or worse? Should we just go with `Result.Error<IDictionary>(`?

src/NUnitFramework/framework/Attributes/PlatformAttribute.cs
mikkelbu: Yes. I'm happy with that :).

src/NUnitFramework/framework/ExceptionExtensions.cs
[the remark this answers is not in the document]
mikkelbu: @jnm2 I don't mind `Result.Error<IDictionary>(...)`, but I'm not sure I follow you with respect to `Result.ErrorResult`. Was that to make `Result.ErrorResult` do "work with a new error type that implicitly converts to the generic type", as you mention above?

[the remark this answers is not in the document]
jnm2: I added the nested type `Result.ErrorResult` so that it is possible to write `return Result.Error(...)` instead of `return Result.Error<IDictionary>(...)` but I'm not sure whether it is worth the fanciness of making the Result API more complex. Do you have a preference?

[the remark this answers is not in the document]
mikkelbu: Sorry, I missed that you had added a commit. Since we are only using `Result.Error(...)` in one location then I prefer the more verbose version where we are explicit about the type. If we had a lot of usage then I think that the "fanciness" would be more warranted.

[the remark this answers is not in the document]
jnm2: Sounds sensible. Pushed a different commit, ready for review then!

src/NUnitFramework/framework/Internal/ExceptionHelper.cs
```diff
@@ -110,41 +110,25 @@ public static string BuildMessage(Exception exception, bool excludeExceptionName
...
+            StringBuilder sb = new StringBuilder(exception.GetStackTraceWithoutThrowing());
 
             foreach (Exception inner in FlattenExceptionHierarchy(exception))
             {
                 sb.Append(Environment.NewLine);
                 sb.Append("--");
                 sb.Append(inner.GetType().Name);
                 sb.Append(Environment.NewLine);
-                sb.Append(GetSafeStackTrace(inner));
+                sb.Append(exception.GetStackTraceWithoutThrowing());
```
mikkelbu:
```suggestion
                sb.Append(inner.GetStackTraceWithoutThrowing());
```

jnm2: Thanks Mikkel, I appreciate that!

jnm2:
Force pushed to fix @mikkelbu's find and then again to fix the merge conflict.

PR #3305, opened 2019-07-04
Remove unused methods

jnm2:
(None of these changes can be detected by users of the NUnit library.)

### DIFF
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -568,7 +568,7 @@ namespace NUnit.Framework.Internal
             public AdhocContext()
             {
                 var type = GetType();
-                var method = type.GetMethod("AdhocTestMethod", BindingFlags.NonPublic | BindingFlags.Instance);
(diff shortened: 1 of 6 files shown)


src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -568,7 +568,7 @@ public class AdhocContext : TestExecutionContext
             public AdhocContext()
             {
                 var type = GetType();
-                var method = type.GetMethod("AdhocTestMethod", BindingFlags.NonPublic | BindingFlags.Instance);
+                var method = type.GetMethod(nameof(AdhocTestMethod), BindingFlags.NonPublic | BindingFlags.Instance);
```
jnm2: This lets ReSharper know that `AdhocTestMethod` is not unused.

jnm2:
Note to self: https://github.com/nunit/nunit/compare/master...jnm2:refactor_provider_cache?expand=1 depends on this

jnm2:
@ChrisMaddock Do you think we should remove the `awaiting` label at the same time as applying the `closed` label?

ChrisMaddock:
Can do...I was just being lazy. 🤷‍♂️😊

PR #3309, opened 2019-07-04
Simplify code in EventListenerTextWriter

jnm2:
The first commit fixes the whitespace in EventListenerTextWriter. Probably best to avoid including this in the diffs you look at.

The remaining commits show each step in the process of removing duplicated code in EventListenerTextWriter.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
@@ -113,9 +113,8 @@ namespace NUnit.Framework.Internal.Execution
         {
             if (value != null)
             {
-                IFormattable f = value as IFormattable;
-                var stringValue = f != null
-                    ? f.ToString(null, FormatProvider)
+                var stringValue = value is IFormattable formattable
+                    ? formattable.ToString(null, FormatProvider)
                     : value.ToString();
 
                 if (TrySendToListener(stringValue))
--- a/src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
@@ -119,11 +119,8 @@ namespace NUnit.Framework.Internal.Execution
         /// </summary>
         public override void Write(object value)
         {
-            if (value != null)
-            {
-                if (TrySendToListener(FormatForListener(value)))
-                    return;
-            }
+            if (value != null && TrySendToListener(FormatForListener(value)))
+                return;
 
             _defaultWriter.Write(value);
         }
--- a/src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
@@ -34,21 +34,21 @@ namespace NUnit.Framework.Internal.Execution
     /// listener is active in the context, or if there is no context,
     /// the output is forwarded to the supplied default writer.
     /// </summary>
-	public class EventListenerTextWriter : TextWriter
-	{
+    public class EventListenerTextWriter : TextWriter
+    {
         private readonly TextWriter _defaultWriter;
-		private readonly string _streamName;
+        private readonly string _streamName;
 
         /// <summary>
         /// Construct an EventListenerTextWriter
         /// </summary>
         /// <param name="streamName">The name of the stream to use for events</param>
         /// <param name="defaultWriter">The default writer to use if no listener is available</param>
-		public EventListenerTextWriter( string streamName, TextWriter defaultWriter )
-		{
-			_streamName = streamName;
+        public EventListenerTextWriter( string streamName, TextWriter defaultWriter )
+        {
+            _streamName = streamName;
             _defaultWriter = defaultWriter;
-		}
+        }
 
         /// <summary>
         /// Get the Encoding for this TextWriter
@@ -61,7 +61,7 @@ namespace NUnit.Framework.Internal.Execution
             if (context == null || context.Listener == null)
                 return false;
 
-            context.Listener.TestOutput(new TestOutput(text, _streamName, 
+            context.Listener.TestOutput(new TestOutput(text, _streamName,
                 context.CurrentTest?.Id, context.CurrentTest?.FullName));
 
             return true;
--- a/src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
@@ -114,15 +114,9 @@ namespace NUnit.Framework.Internal.Execution
             if (value != null)
             {
                 IFormattable f = value as IFormattable;
-                string stringValue;
-                if (f != null)
-                {
-                    stringValue = f.ToString(null, FormatProvider);
-                }
-                else
-                {
-                    stringValue = value.ToString();
-                }
+                var stringValue = f != null
+                    ? f.ToString(null, FormatProvider)
+                    : value.ToString();
 
                 if (TrySendToListener(stringValue))
                     return;
--- a/src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
@@ -55,6 +55,14 @@ namespace NUnit.Framework.Internal.Execution
         /// </summary>
         public override Encoding Encoding { get; } = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
 
+        private string FormatForListener(object value)
+        {
+            return
+                value is null ? string.Empty :
+                value is IFormattable formattable ? formattable.ToString(null, FormatProvider) :
+                value.ToString();
+        }
+
         private bool TrySendToListener(string text)
         {
             var context = TestExecutionContext.CurrentContext;
@@ -113,11 +121,7 @@ namespace NUnit.Framework.Internal.Execution
         {
             if (value != null)
             {
-                var stringValue = value is IFormattable formattable
-                    ? formattable.ToString(null, FormatProvider)
-                    : value.ToString();
-
-                if (TrySendToListener(stringValue))
+                if (TrySendToListener(FormatForListener(value)))
                     return;
             }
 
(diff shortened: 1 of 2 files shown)


oznetmaster:
Why??

You really think this code is easier to read and maintain?

I have no problem with removing/refactoring duplicated code. 

It is the rest of it that concerns me.

oznetmaster:
Once again, just because the C# team has decided to add some new feature to C# does not mean that it is good/maintainable coding practice to use it.

There is much controversy and disagreement over many of the features.

I certainly see no need to start retrofitting existing code to use them, for no other reason then they exist.

I am sure that this is one of those divisive issues.

jnm2:
@oznetmaster 

> Why??
>
> You really think this code is easier to read and maintain?

Are you referring to the first commit, "This is the way I wanted to write it originally," or the second commit, "Less verbosity"?

For the first, it makes sense to me that I should be allowed to contribute the code I originally wanted to contribute and for which I asked that we allow C# 7.

For the second, it's changing three lines to one line. If this one commit is that big of a deal to you, I'm happy to drop it but very confused at your strong reaction.

jnm2:
@oznetmaster 

> Why??
>
> You really think this code is easier to read and maintain?

Are you referring to the first commit, "This is the way I wanted to write it originally," or the second commit, "Less verbosity"?

For the first, it makes sense to me that I should be allowed to contribute the code I originally wanted to contribute and for which I asked that we allow C# 7. In the case of the switch especially, it's now three times as readable in my opinion.

For the second, it's changing three lines to one line. If this one commit is that big of a deal to you, I'm happy to drop it but very confused at your strong reaction. This isn't "many of the features." This is a single C# 6 feature, the `?.` operator. We use it many other places in NUnit already.

> I certainly see no need to start retrofitting existing code to use them, for no other reason then they exist.

I object to this characterization. This is not happening in any line of this PR.

jnm2:
Not only that, but what was specifically rejected was the mass changes for the sake of changes. Not C# 7 syntax. I fully intend to use it judiciously in all future PRs of any nature.

jnm2:
Not only that, but what was specifically rejected was the mass changes for the sake of changes. Not C# 7 syntax. I fully intend to use it judiciously in all future PRs of any nature.

Since you aren't commenting on the code itself, I'm going to bring it into the discussion:

```cs
case Status.ShuttingDown when _timeSinceShutdown.Elapsed >= _shutdownTimeout:
case Status.ShutDown:
    throw ErrorAndGetExceptionForShutdownTimeout();
```

```cs
var context = SynchronizationContext.Current as SingleThreadedTestSynchronizationContext
    ?? throw new InvalidOperationException("This strategy must only be used from a SingleThreadedTestSynchronizationContext.");
```

```cs
(context.TestObject as IDisposable)?.Dispose();
```

jnm2:
Oops, didn't mean to click Close. Sorry!

jnm2:
@oznetmaster Please clarify if you'd like me to drop the second commit and/or move the first commit to its own PR.

jnm2:
> There is much controversy and disagreement over many of the features.

Since you make this point, please mention specifically which lines of code in this PR you believe to be controversial to @nunit/framework-team. I listened to the feedback from all of you in the two C# 7 PRs and made sure that this PR doesn't fall into any specific issue that anyone mentioned. If I missed something, **please** point to a particular line of code. That's much less confounding than a blanket statement that any C# 7 is bad.

(For one thing, we decided as a team to allow C# 7 syntax. For another, the syntax itself had positive comments in other PRs from multiple members. Lastly, the vehemence against C# 7 in the general programming world which you've cited a few times is not visible in the slice of the world that I have access to. I'll take it into consideration, but on a personal level I have no option but to consider it in proportion to the other evidence around me.)

Please also clarify if you'd like me to drop the second commit and/or move the first commit to its own PR.

oznetmaster:
For example, I asked a number of my colleagues about:
```
                var context = SynchronizationContext.Current as SingleThreadedTestSynchronizationContext;

                if (context == null)
                    throw new InvalidOperationException("This strategy must only be used from a SingleThreadedTestSynchronizationContext.");
```
vs.
```
                var context = SynchronizationContext.Current as SingleThreadedTestSynchronizationContext
                    ?? throw new InvalidOperationException("This strategy must only be used from a SingleThreadedTestSynchronizationContext.");

``` 
and I could not find a single one who did not think that the second one was inferior to the first for readability and maintainability.

Obviously, we work in totally different programming environments.

jnm2:
@oznetmaster I'm willing to drop that to keep the peace, but since nothing tells me *why* this is less readable or maintainable, I will certainly use it in future PRs. That makes sense, right?

CharliePoole:
@jnm2 I vowed to keep out of this set of refactoring PRs but I'll just butt in to say that the standard for what you might do in a PR for a different purpose is (and should be) biased towards your preferences. The person who does the fix should get to decide, within limits of a standard, how to do it. Other folks may have opinions, which you should pay attention to, but you get to decide.

On the other hand, when you do a global fix just for its own sake, you open a can of worms. Everyone won't agree that what you are doing is desirable and, by definition, it's not necessary.

Whether (for example) using something other than a value after `??` is clear or unclear is a matter of opinion. (*) It's not worth putting forth arguments as to why your opinion is right, because there is no right in matters of style. There's only agreement or non-agreement.

I understand that putting forth your own view strongly can sometimes work to get a productive discussion going. So far, I'm not seeing that happening here. Usually, we are better able to do this in issues than PRs I think.

(*) My opinion, it's absolutely not.

jnm2:
@CharliePoole Thanks, those are good points.

@CharliePoole and @oznetmaster I hope it's clear that there are no global refactorings being discussed or in progress. Today's set of refactoring PRs has nothing to do with code style, with the exception of the two commits which I have removed. I may bring one back in a separate PR because the `goto case` is really harder to read than the C# 7 when clause, but it's definitely separable from this PR.

jnm2:
(ready for review of the remaining code)

CharliePoole:
@jnm2 Sorry, since I wasn't reviewing this one, I actually only read @oznetmaster 's comment and took it to be a continuation of our standards discussions. I see now it's a specific item you felt needed to be refactored. In doing so, you used certain features as is your choice as the person doing it. IMO anyone who reviews should specifically indicate lines you changed and say something like "This seems confusing to me...." 

Sorry I missed that. However, I think most of us are out here like stateless services, responding to the specific comments we see when we see them and therefore sometimes missing some context.

Now that I'm clear about what you are doing, I would suggest adding (in future refactoring PRs) a sentence or two that says why you are doing this particular thing right now. I don't think the title is sufficient. Removing duplication is good as is simplifying code. But if you were not already working in `EventListenerTextWriter`, there was some motivation as to why you wanted to do this right now. IMO it helps to tell us, even if all there is to tell is "I got tired of seeing this" or "I'm finding this code hard to follow." Then anyone who reviews can feel that they are helping you do something you want to accomplish.

At least, this is what I would try to do and recommend to anyone else. YMMV.

src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
```diff
@@ -34,34 +34,42 @@ namespace NUnit.Framework.Internal.Execution
...
 
         /// <summary>
         /// Get the Encoding for this TextWriter
         /// </summary>
         public override Encoding Encoding { get; } = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
 
+        private string FormatForListener(object value)
+        {
+            return
+                value is null ? string.Empty :
```
CharliePoole: I have always lined up and indented the ternary op like so...

```c#
return value is null
    ? string.Empty
    : value is IFormattable formattable
        ? formattable.ToString(null, FormatProvider)
        : value.ToString();
```

I'm not sure if we have a standard, but that seems much clearer to me.

I'm as yet unclear about the inline declaration of `formattable`. I suspect I'll like it after I get used to it though.

src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
```diff
@@ -264,25 +256,8 @@ public override void WriteLine(string value)
...
-                    if (!TrySendLineToListener(f.ToString(null, FormatProvider)))
-                        _defaultWriter.WriteLine(value);
-                }
-                else
-                {
-                    if (!TrySendLineToListener(value.ToString()))
-                        _defaultWriter.WriteLine(value);
-                }
-            }
+            if (!TrySendLineToListener(FormatForListener(value)))
```
CharliePoole: I like how it cleans up the duplication.

jnm2:
@CharliePoole 

> Now that I'm clear about what you are doing, I would suggest adding (in future refactoring PRs) a sentence or two that says why you are doing this particular thing right now. I don't think the title is sufficient. 

Thanks, this seems like a great idea. In the case of these six PRs, these are places where I ran across the code for unrelated reasons and the code raised a flag in my mind that it could be cleaner or in some cases where the implementation could be improved.

src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
jnm2: I go back and forth between that formatting and this one depending on how I want to look at it: flat and linear like a switch statement, or as a tree. The way it is now, it lines up all the cases after `?` characters and ends all the lines with an outcome. I usually switch to the tree structure when it can't be thought of as flat, but I see that situation as more complex. Happy to change this though.

src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
CharliePoole: Only a suggestion. I could try to look at it as a switch statement and see how that works.

src/NUnitFramework/framework/Internal/Execution/EventListenerTextWriter.cs
jnm2: > Only a suggestion. I could try to look at it as a switch statement and see how that works.

Okay, I might lean towards leaving it as-is then.

Speaking of switching, here is C# 8's [switch expression](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#switch-expressions)—not sure yet if I like it as much as my flat ternary though:

```c#
return value switch
{
    null => string.Empty,
    IFormattable formattable => formattable.ToString(null, FormatProvider),
    _ => value.ToString()
};
```

PR #3310, opened 2019-07-04
Increase timeout for CI and use events rather than spin waits in tests

### LINKED ISSUE #3307: Sporadic GetResultIsNotCalledUntilContinued failure
I've seen this a few times in CI builds:

```
1) Failed : NUnit.Framework.AwaitableReturnTypeTests("ReturnsCustomTask").GetResultIsNotCalledUntilContinued
GetResult was not called after the continuation passed to OnCompleted was invoked.
   at NUnit.Framework.AwaitableReturnTypeTests.GetResultIsNotCalledUntilContinued() in C:\projects\nunit\src\NUnitFramework\tests\AwaitableReturnTypeTests.cs:line 83
```

jnm2:
Fixes #3307. Can't repro, but I think it was as simple as the 1000 timeout being too short for CI. Added a commit to increase to the standard 10_000.


ChrisMaddock:
Let's try it. 🙂

PR #3317, opened 2019-07-12
Fix check for parameterless delegate

### LINKED ISSUE #3315: Assert.DoesNotThrow() stopped working as it was previously
Some specs of environment first.
Windows 10 1809 build 17763.557
VS 2017 v15.9.12
.Net Framework Version: 4.7.03190
Test project is targeted for .Net Core 2.0.3
NUnit package version: 3.12.0 (latest for now)
NUnit3TestAdapter package version 3.12.0

Let's say we have an extension method like:

```
public static void Run<T>(this IEnumerable<T> sequence)     {
     foreach (var item in sequence);
}
```

and I used to use it like
  `Assert.DoesNotThrow(someEnumerable.Run)`

because IDE (or resharper, whatever)  suggests me use it instead of 
  `Assert.DoesNotThrow(()=>someEnumerable.Run())`

And everything was fine but then I had to update because another bug was fixed it started failing my tests with the exception
"System.ArgumentException : The actual value must be a parameterless delegate but was TestDelegate.
Parameter name: parameterName"

What is that? Do you even test the changes you make, guys?

Here is the test:

```
[Test]
public void NunitDoesNotThrowTest()
{
  TestDelegate fullform = () => Enumerable.Repeat(1, 10).Run(); // it is your TestDelegate. working
  TestDelegate shortForm = Enumerable.Repeat(1, 10).Run; // this is to. should be working because IT IS your TestDelegate (acording to VS) you want me to pass into DoesNotThrow() method!
  Assert.DoesNotThrow(() => Assert.DoesNotThrow(fullform));
  Assert.DoesNotThrow(() => Assert.DoesNotThrow(shortForm));
}
```

jnm2:
Fixes #3315 

I'm not sure how I overlooked the difference between checking the delegate signature and checking the method signature.

A delegate can have one fewer parameter than the method by binding a target value to the first parameter of a static method. (That's what we're seeing in #3315.) A delegate can also have one more parameter than the parameterize the `this` instance of a parameterless method. There is a test for each of these things.

### DIFF
--- a/src/NUnitFramework/framework/Internal/ExceptionHelper.cs
+++ b/src/NUnitFramework/framework/Internal/ExceptionHelper.cs
@@ -73,7 +73,7 @@ namespace NUnit.Framework.Internal
 
         /// <summary>
         /// Builds up a message, using the Message field of the specified exception
-        /// as well as any InnerExceptions. Optionally excludes exception names, 
+        /// as well as any InnerExceptions. Optionally excludes exception names,
         /// creating a more readable message.
         /// </summary>
         /// <param name="exception">The exception.</param>
@@ -212,7 +212,7 @@ namespace NUnit.Framework.Internal
             Guard.ArgumentNotNull(parameterlessDelegate, parameterName);
 
             Guard.ArgumentValid(
-                parameterlessDelegate.GetMethodInfo().GetParameters().Length == 0,
+                parameterlessDelegate.GetType().GetMethod("Invoke").GetParameters().Length == 0,
                 $"The actual value must be a parameterless delegate but was {parameterlessDelegate.GetType().Name}.",
                 nameof(parameterName));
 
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/framework/Internal/ExceptionHelper.cs
```diff
@@ -212,9 +212,9 @@ internal static Exception RecordException(Delegate parameterlessDelegate, string
             Guard.ArgumentNotNull(parameterlessDelegate, parameterName);
 
             Guard.ArgumentValid(
-                parameterlessDelegate.GetMethodInfo().GetParameters().Length == 0,
+                parameterlessDelegate.GetType().GetMethod("Invoke").GetParameters().Length == 0,
                 $"The actual value must be a parameterless delegate but was {parameterlessDelegate.GetType().Name}.",
-                nameof(parameterName));
+                parameterName);
```
jnm2: I was really on a roll that day, wasn't I?

PR #3321, opened 2019-07-13
Keep dependencies up to date

jnm2:


src/NUnitFramework/tests/nunit.framework.tests.csproj
```diff
@@ -17,8 +17,8 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' != 'net35' and '$(TargetFramework)' != 'net40'">
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.5.0" />
-    <PackageReference Include="NUnit3TestAdapter" Version="3.10.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.7.0" />
```
jnm2: All later packages cause our NUnitLite tests to fail on netcoreapp2.0 and I'm not ready to dig into that.

.editorconfig
```diff
@@ -12,6 +12,9 @@ indent_style = space
 insert_final_newline = true
 charset = utf-8
 
+[*.{sln,*proj}]
+charset = utf-8-bom
```
mikkelbu: The commit message says "VS needs projects to use byte order marks". I tried to google for this, but found answers in both directions.
Is the reason the same as in DotNetAnalyzers/StyleCopAnalyzers#1142? (since the first comment in https://marketplace.visualstudio.com/items?itemName=SergeyVlasov.FixFileEncoding#review-details seems to indicate that some people want sln, csproj ... files without BOM).

src/NUnitFramework/tests/nunit.framework.tests.csproj
mikkelbu: We should probably create an issue for this either here or in https://github.com/microsoft/vstest)

.editorconfig
jnm2: Ah. If you ever edit the solution or project file, VS forces a resave with BOM and so git shows an edit on the top row back and forth for no good reason.

jnm2: Fwiw I strongly disagree with byte order marks. But so long as tooling forces thrashing like this on us, I'd rather not ask everyone to keep an eye on what VS does and squash it out of their PRs.

mikkelbu: Ok. I read "VS **needs** projects to use byte order marks" as this change was required for VS to work, but I'm all for reducing noise in reviews.

jnm2: That wasn't clear wording, I'm sorry!

src/NUnitFramework/tests/nunit.framework.tests.csproj
jnm2: Opened https://github.com/nunit/nunit/issues/3327 (sorry that's all I can do right now)

jnm2: Opened https://github.com/nunit/nunit/issues/3327 (sorry that's all I can do right now)

PR #3366, opened 2019-09-01
Timeout no longer enforced with debugger attached

### LINKED ISSUE #3054: Don't enforce [Timeout] when debugger is attached
Greetings NUnit team!

I was just debugging within a nunit test that has a Timeout attribute. Unfortunately it was aborted by timeout while debugging. I have found this existing issue #1479 that was already fixed by #1785 in 2016.
For me it seems the fix got lost with commit 1a164db06b1a91b0d82f5481286b960cc6db2644 .

So is there something wrong with my little research result or was the feature lost without intention?

### LINKED ISSUE #3282: TimeoutAttribute makes all Assertions count as failure
Hello

TimeoutAttribute(via TimeoutCommand) wraps the tests in a Task.Wait call under netcore (because Thread.Abort is no longer supported API), here: https://github.com/nunit/nunit/blob/d8490ecd6171538fa0ba9cb5d38948bde5ebd61e/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs#L102

Using any of the 'flow-control' assertions like Assert.Ignore will throw their respective special-case Exception (here IgnoreException), which Task.Wait will wrap in an AggregateException, which is not handled by ValidateAndUnwrap here
https://github.com/nunit/nunit/blob/c3caee3dcbc4704cd9a2c4ae426789869d5d092c/src/NUnitFramework/framework/Internal/Results/TestResult.cs#L564

This will result in the test counting as full-blown failure due to unhandled exception.

This simple test will do it:
```        
[Test, Timeout(1234)]
public void TimeoutBug()
{
    Assert.Ignore("a");
}
```

I believe it should be safe to change ValidateAndUnwrap to handle AggregateExceptions the same way it does for TargetInvocationExceptions. It may also be beneficial to dive into the exception recursively instead of only once.

zdenek-jelinek:
Fixes #3054, fixes #3282.
* Adjusted TimeoutCommand to no longer cancel tests when running with debugger attached.
* Introduced IDebugger and DebuggerProxy into a new namespace NUnit.Framework.Internal.Abstractions - I could not find any precedent for the concept and I can imagine other abstractions such as IClock to also appear over time so these could be put there as well.
* Introduced new API overloads for creating tests with ability to supply custom IDebugger, defaulting to production DebuggerProxy otherwise.
* Adjusted .NET Standard implementation of TimeoutCommand to emulate async/await exception propagation instead of Task.Wait/AggregateException.
* Added unit test coverage for all possible test outcomes when using Timeout.


src/NUnitFramework/framework/Internal/Abstractions/DebuggerProxy.cs
```diff
@@ -0,0 +1,38 @@
+// ***********************************************************************
+// Copyright (c) 2019 Charlie Poole
```
ChrisMaddock, line 2:
```suggestion
// Copyright (c) 2019 Charlie Poole, Rob Prouse
```

src/NUnitFramework/framework/Internal/Abstractions/DebuggerProxy.cs
```diff
@@ -0,0 +1,38 @@
...
+// ***********************************************************************
+
+using System.Diagnostics;
+
+namespace NUnit.Framework.Internal.Abstractions
+{
+    /// <summary>
+    /// A production <see cref="IDebugger"/> implementation that delegates directly to .NET's <see cref="Debugger"/>.
+    /// </summary>
+    public class DebuggerProxy : IDebugger
```
ChrisMaddock, line 31: Could these two new classes be internal, so they're not visible to consumers of the framework? I guess they maybe can't be, due to the being used in the signatures of other classes which are public?

src/NUnitFramework/framework/Internal/Abstractions/IDebugger.cs
```diff
@@ -0,0 +1,36 @@
+// ***********************************************************************
+// Copyright (c) 2019 Charlie Poole
```
ChrisMaddock, line 2:
```suggestion
// Copyright (c) 2019 Charlie Poole, Rob Prouse
```

src/NUnitFramework/framework/Internal/Abstractions/IDebugger.cs
```diff
@@ -0,0 +1,36 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+namespace NUnit.Framework.Internal.Abstractions
+{
+    /// <summary>
+    /// A layer of abstraction around <see cref="System.Diagnostics.Debugger"/> to facilitate testing.
+    /// </summary>
+    public interface IDebugger
```
ChrisMaddock, line 29: Same comment as above

zdenek-jelinek:
Hello @ChrisMaddock and thank you for bringing this up, for your suggestions and for going through the code. I will be more than happy to do any changes required.

I have made two more commits that address the issues you have raised.

src/NUnitFramework/framework/Internal/Abstractions/DebuggerProxy.cs
zdenek-jelinek: I only did that to keep in line with the rest of the framework classes which all seem to be public. I was also afraid of having explosion of "internalization" for many types and methods but it turned out to be ok.

zdenek-jelinek:
I had some thoughts about the CI because of the flaky behavior and I think the issue was caused by the test being run which had 100ms timeout and contained 200ms sleep.

My reasoning is that the VM was probably loaded with some other tasks, and since the termination of the pending test is scheduled through a Timer on .NET 4.0, it happened that the timer callback to abort the test was scheduled but ran after the test actually completed. I have reproduced the issue with identical stack trace as the CI error on my x64 Win 10 using timeout equal to the sleep period.

As a fix, I have applied a timeout of 50ms and sleep period of 500ms which were being used in the tests originally.

Please note that I have near zero experience with Mono or OS X and would be very happy for someone more experienced to check it too.

src/NUnitFramework/TestBuilder.cs
```diff
@@ -107,16 +108,29 @@ public static WorkItem CreateWorkItem(Test test)
...
         public static WorkItem CreateWorkItem(Test test, object testObject)
         {
-            var context = new TestExecutionContext();
-            context.TestObject = testObject;
-            context.Dispatcher = new SuperSimpleDispatcher();
+            return CreateWorkItem(test, testObject, null);
+        }
 
-            return CreateWorkItem(test, context);
+        internal static WorkItem CreateWorkItem(Test test, object testObject, IDebugger debugger)
```
jnm2, line 114: The namespace betrays the fact that this class is not part of the framework (in spite of the file path). This is utility code used in NUnit's self-tests. Could you change the three added methods to public?

src/NUnitFramework/framework/Internal/Abstractions/DebuggerProxy.cs
```diff
@@ -0,0 +1,38 @@
...
+// ***********************************************************************
+
+using System.Diagnostics;
+
+namespace NUnit.Framework.Internal.Abstractions
+{
+    /// <summary>
+    /// A production <see cref="IDebugger"/> implementation that delegates directly to .NET's <see cref="Debugger"/>.
+    /// </summary>
+    internal class DebuggerProxy : IDebugger
```
jnm2, line 31: (by design, it wouldn't make sense to inherit from this class)

```suggestion
    internal sealed class DebuggerProxy : IDebugger
```

src/NUnitFramework/TestBuilder.cs
jnm2: Actually, could you add optional parameters rather than adding overloads since there are no backwards compatibility constraints due to this class not being shipped?

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -48,11 +50,15 @@ public class TimeoutCommand : BeforeAndAfterTestCommand
         /// </summary>
         /// <param name="innerCommand">The inner command</param>
         /// <param name="timeout">Timeout value</param>
-        public TimeoutCommand(TestCommand innerCommand, int timeout) : base(innerCommand)
+        /// <param name="debugger">An <see cref="IDebugger"/> instance</param>
+        internal TimeoutCommand(TestCommand innerCommand, int timeout, IDebugger debugger) : base(innerCommand)
```
jnm2, line 54: This is a problem. The TimeoutCommand type is not usable at all unless its constructor is usable. We may as well make the whole type internal. @nunit/framework-team Since this is under the Internal namespace, are you okay with that?

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -99,7 +110,12 @@ public override TestResult Execute(TestExecutionContext context)
         {
             try
             {
-                if (!Task.Run(() => context.CurrentResult = innerCommand.Execute(context)).Wait(_timeout))
+                var testExecution = RunTestWithTimeoutAsync(context);
+                if (_debugger.IsAttached)
+                {
+                    AwaitCompletionWithoutExceptionWrapping(testExecution);
+                }
```
jnm2, line 117: This wasn't the most straightforward for me to see because the pieces are spread out, but functionally it's solid.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -114,6 +130,29 @@ public override TestResult Execute(TestExecutionContext context)
...
+            AwaitWithoutExceptionWrapping(testExecution, _timeout);
+
+            return testExecution;
+        }
+
+        private static void AwaitWithoutExceptionWrapping(Task<TestResult> testExecution, int timeoutMilliseconds)
+        {
+            var timeoutOrCompletion = Task
+                .WhenAny(testExecution, Task.Delay(timeoutMilliseconds))
+                .Unwrap();
```
jnm2, line 147: No action needed, but I want to comment on the preexisting code. The recommended approach for timeouts is WhenAny with `new CancellationTokenSource(timeout).Token`.

The CancellationTokenSource approach could then evolve to us using a timer just like the `THREAD_ABORT` section. When the timer ticks and debugger is not attached at that point, we call CancellationTokenSource.Cancel().

src/NUnitFramework/TestBuilder.cs
zdenek-jelinek: The internal is required because the IDebugger type is internal. I can change that but I think maybe it's better to have this method internal instead of IDebugger public?
Regarding the overloads, I chose them because it felt more obvious to me than having IDebugger = null which then defaults to non-null instance but I am happy to make the change, I can also see that multitude of overloads is not very obvious either.

jnm2: Ah, right. Could you make the class internal then? This file is only included in nunit.framework.tests.csproj and nunitlite.tests.csproj.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
CharliePoole: @jnm2 (I know I'm not on the team but I hope this may be useful)

The design principle behind custom attributes was that users would implement interfaces and not reuse code. Based on that principle, I would say this is __not__ breaking.

OTOH the framework has accepted bugs in recent years where people re-used such classes, even though they were not designed to be reused. And we know from experience that some people do re-use classes rather than just implement the interfaces. So your guess is as good as mine... maybe better.

Of course, this is just one instance of a big disagreement we've kicked around for a few years: is it OK to break an accessible class if people were not "supposed" to use it by some definition or another?

Anyway, it's out of my wheelhouse now, but that's how I see the choice.

jnm2: That's a good summary. If we think people may be using this class directly, we will have to preserve the `public TimeoutCommand(TestCommand innerCommand, int timeout)` constructor since I can't think of a way this class could be used from the outside without calling this constructor. If we don't think people may be using this class directly, or we are willing to break their code, then there's no reason to keep the class public.

For a member like this constructor which is essential to the use of the class itself, the member needs to be as accessible as the whole class. To do otherwise is a design smell.

ChrisMaddock: Ahhh, this old chestnut. 🙂 

My view: TimeoutCommand is new-ish (2017, https://github.com/nunit/nunit/pull/2106) - and has been in the Internal namespace its entire existence. I agree the Internal namespace isn't anywhere near an ideal solution, but I do think it's a policy that we can stand by. If we go the other way and make the new DebuggerProxy classes public, we're just making the overall problem worse - so I think we should take a risk on making TimeoutCommand internal, as the lesser evil.

@jnm2 /@CharliePoole /@zdenek- : Can we make come to a decision between the four of us,  so this doesn't end up stuck again and we can get it merged? 😄 Who wants to disagree with me? 😁

ChrisMaddock:
Thanks for taking a look at this Joseph - appreciate the expert's opinion! 😄 

---

@zdenek-jelinek - in case you're interested, bit of background around out public vs internal conversation. NUnit was originally written without any use of access modifiers like internal. 20 years later, this has lead to lots of people relying on lots of classes we didn't intend them too, and problems every time we want to make a change to method signatures, just like this PR. It's something we're trying to gradually clean up, but it's not an easy task, when you have thousands of users, a 20 year old code base, and no idea who's relying on what! 😅

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
jnm2: Looks like GitHub's autocomplete messed up on you there. /cc @zdenek-jelinek

I don't disagree. It's worth pointing out that we can add this internal constructor and leave the existing public constructor in place. We wouldn't be talking about making IDebugger public necessarily.

ChrisMaddock: Maybe that's the best solution - it's low-cost, and zero risk.

jnm2: If we plan to clean up in v4 by making the class internal, we could also signal that now with an obsoletion message.

CharliePoole: @ChrisMaddock As a user of the framework, it's a don't care for me. I think your point of view makes sense and I'd support it if I had a vote, which I don't. Sorry, but as we discussed privately, I take formal roles very seriously.

ChrisMaddock: I agree - although I think that could be applied much more widely across a lot of classes in this area. Perhaps out of scope for this PR, and a separate project? 🙂

jnm2: Yes. So I have to say I want to make the class internal, but it's hard to argue with low cost and zero risk and I can't disagree with that. So what we're asking @zdenek-jelinek to do is to restore the original public constructor with its signature unchanged. It would call the internal constructor so that it would continue to work as it had previously for users depending on that constructor signature.

zdenek-jelinek: Just one more question from my side: How does an internal consumer (i.e. person working on NUnit framework) know which constructor to use? Should I [Obsolete("Will be removed in the next major release")] the public ctor, or just leave it as it were?

CharliePoole:
This discussion has really gotten down into the weeds and I was only involved in a small part of it, but I just got curious and looked at the current code. Maybe I'm missing something, but why is this not simply a change to SimpleWorkItem, which decides whether there should be a TimeoutCommand at all? It curretnly decides how to build the command chain and only inserts a TimeoutCommand if running under the debugger.

Of course, I wrote the command structure so that the commands are regenerated every time they are needed. That was an upfront decision - for reasons - and may have been changed in the years since. So feel free to point out what I'm missing. :-)

zdenek-jelinek:
That's a very good point. What I have done here is based on my very limited understanding of NUnit internals and the original pre-command implementation where this worked. It is true that adding a simple debugger attached check in SimpleWorkItem would be much simpler.

The benefits of the current implementation, as I see them:
1. Timeout applies based on debugger state at the time of timeout, not at the time of start. This allows for attaching into a previously running test without it being canceled mid-session. Or detaching from a test and observing the timeout.
1. Smaller difference between debugger attached/detached behavior - we still get a new thread spawned in .NET Core even if timeout does not apply. 

I can see myself appreciating both of these cases but I cannot talk for the NUnit community.

CharliePoole:
Interesting. I see how you made the first one possible by checking the interface at the time the test would normally time out. The second point, about running in a separate thread, I'd view as something we should try to correct in .NET Core. 

Anyway, thank... just curious.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
jnm2: @zdenek-jelinek Good point. Either that or an XML doc comment, but let's use do the obsoletion warning.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -131,27 +152,16 @@ public override TestResult Execute(TestExecutionContext context)
...
-        private static void AwaitWithoutExceptionWrapping(Task<TestResult> testExecution, int timeoutMilliseconds)
+        private static void Wait(Task task, int timeoutMilliseconds)
         {
-            var timeoutOrCompletion = Task
-                .WhenAny(testExecution, Task.Delay(timeoutMilliseconds))
-                .Unwrap();
-
-            AwaitCompletionWithoutExceptionWrapping(timeoutOrCompletion);
-        }
+            var timeoutOrCompletion = Task.WhenAny(task, Task.Delay(timeoutMilliseconds));
```
siprbaum, line 162: This is not new in your change, but why is this this using this Task.WhenAny and that additinal Task.Delay here instead of directly doing `Task.Wait(timeoutMilliseconds)` here?

zdenek-jelinek: `Task.Wait` throws `AggregateException` if the task in question throws during the wait. In that case, there would be a need for try/catch and `ExceptionDispatchInfo` so that it is possible to rethrow the actual exception thrown by the test.

siprbaum: I was not thinking about error handling when looking at that code but now that you spelled it out explicitely it makes sense! 
Thx for the explanation. It is really appreciated.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
zdenek-jelinek: When revisiting the code, I figured I have issues understanding it either. I have changed the code a bit, I'm still not entirely happy with it but I hope it's easier to understand now. I hope you don't mind that I change it.

src/NUnitFramework/TestBuilder.cs
```diff
@@ -179,7 +183,12 @@ public static ITestResult RunTest(Test test)
 
         public static ITestResult RunTest(Test test, object testObject)
         {
-            return ExecuteWorkItem(CreateWorkItem(test, testObject));
+            return RunTest(test, testObject, null);
+        }
+
+        internal static ITestResult RunTest(Test test, object testObject, IDebugger debugger)
```
jnm2, line 189:
```suggestion
        public static ITestResult RunTest(Test test, object testObject, IDebugger debugger)
```

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
```diff
@@ -35,16 +36,20 @@ namespace NUnit.Framework.Internal.Execution
...
+        private readonly IDebugger _debugger;
+
         readonly TestMethod _testMethod;
 
         /// <summary>
         /// Construct a simple work item for a test.
         /// </summary>
         /// <param name="test">The test to be executed</param>
         /// <param name="filter">The filter used to select this test</param>
-        public SimpleWorkItem(TestMethod test, ITestFilter filter)
```
jnm2, line 45: The same things apply to `SimpleWorkItem` that applied to `TimeoutCommand`. Either this will break people and we should leave the public constructor in place, or it won't break people and therefore making the whole class internal won't break people either since I believe the class is unusable if the constructor is unusable.

For consistency we should do what we did for TimeoutCommand.

src/NUnitFramework/framework/Internal/Execution/WorkItemBuilder.cs
```diff
@@ -41,11 +51,11 @@ public static class WorkItemBuilder
         /// <param name="filter">The filter to be used in selecting any child Tests.</param>
         /// <param name="recursive">True if child work items should be created and added.</param>
         /// <returns></returns>
-        static public WorkItem CreateWorkItem(ITest test, ITestFilter filter, bool recursive = false)
```
jnm2, line 44: This is another breaking change. For consistency we should do what we did for TimeoutCommand.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -99,12 +120,28 @@ public override TestResult Execute(TestExecutionContext context)
         {
             try
             {
-                if (!Task.Run(() => context.CurrentResult = innerCommand.Execute(context)).Wait(_timeout))
+                var testExecution = RunTestOnSeparateThread(context);
+
+                Wait(testExecution, _timeout);
+
+                if (testExecution.IsCompleted)
                 {
```
jnm2, line 128: Since we're changing the implementation from `.Wait(_timeout)`, I would suggest this:

```cs
if (Task.WaitAny(new[] { testExecution }, _timeout) != -1)
{
```

GetAwaiter().GetResult() on Task.WhenAny is not better than .Wait because Task.WhenAny is guaranteed to complete with TaskStatus.RanToCompletion. (https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.whenany?view=netcore-3.1#System_Threading_Tasks_Task_WhenAny_System_Threading_Tasks_Task___)

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -99,12 +120,28 @@ public override TestResult Execute(TestExecutionContext context)
...
-                        $"Test exceeded Timeout value {_timeout}ms.",
-                        FailureSite.Test));
+                    context.CurrentResult = testExecution.GetAwaiter().GetResult();
+                }
+                else
+                {
+                    if (_debugger.IsAttached)
+                    {
+                        // Wait for the test to complete
+                        context.CurrentResult = testExecution.GetAwaiter().GetResult();
```
jnm2, line 136: Consider merging the duplicate blocks: `f (Task.WaitAny(new[] { testExecution }, _timeout) != -1 || _debugger.IsAttached)`

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
zdenek-jelinek: Thank you, I did not realize I could use `WaitAny` for this.

zdenek-jelinek:
Thank you and thanks to everybody for the reviews, discussion and feedback!

PR #3377, opened 2019-09-24
Better error messages for generic test with type parameters

### LINKED ISSUE #3215: Better error message for generic test where type parameter cannot be determined
This is a corner case that I only ran into in due to work on the analyzers in nunit/nunit.analyzers#54, and my guess is that very few will run into this problem. Nevertheless, I think we could improve the error message, and fix seems simple and safe.

Given a generic test method with no parameters we get a failure that is difficult to interpret for the users (see last example below for more detail). E.g. the test
```csharp
        [Test]
        public void TestWithGenericTypeThatFailsWithUnusableMessage<T>()
        {
        }
```
fails with `System.InvalidOperationException : Late bound operations cannot be performed on types or methods for which ContainsGenericParameters is true`.

I propose that we change the code at https://github.com/nunit/nunit/blob/db5bee0f8bb74d68202377917f6d958c0e139bfd/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs#L221-L229 so that we also return ` MarkAsNotRunnable(testMethod, "Unable to determine type arguments for method")` if `arglist == null` instead of failing hard later on.

**Context**
The following test is executed without problem, as NUnit determines that `int` can be used as `T`.
```csharp
        [TestCase(1, ExpectedResult = 1)]
        public T TestWithGenericTypeThatWorks<T>(T arg1)
        {
            return arg1;
        }
```
But if we cannot determine the type of `T` using the parameters of the method
```csharp
        [TestCase(1, ExpectedResult = 1)]
        public T TestWithGenericTypeThatFailsWithUsableMessage<T>(int arg1)
        {
            return default(T);
        }
```
then we mark the test as not runnable with the message `Unable to determine type arguments for method`.

However, if the method does not have any parameters
```csharp
        [Test]
        public void TestWithGenericTypeThatFailsWithUnusableMessage<T>()
        {
        }
```
then we fail with the message `System.InvalidOperationException : Late bound operations cannot be performed on types or methods for which ContainsGenericParameters is true.` and a stacktrace
```
at System.Reflection.RuntimeMethodInfo.ThrowNoInvokeException()
   at System.Reflection.RuntimeMethodInfo.InvokeArgumentsCheck(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at NUnit.Framework.Internal.Reflect.InvokeMethod(MethodInfo method, Object fixture, Object[] args) in C:\src\nunit\nunit\src\NUnitFramework\framework\Internal\Reflect.cs:line 266
```

Augure34:
Fixes #3215


jnm2:
@Augure34 Thanks for working on this! I agree with @mikkelbu. If we don't specifically test what we care about, we might inadvertently roll back your changes. Let us know if you need any help.

Augure34:
@jnm2 Of course, I'll have a look at it (sorry had a busy week)

jnm2:
No rush btw. We know what it's like to be busy!

Augure34:
I pushed a few tests to cover the changes. I have a few specific questions though (in case someone has a couple minutes):

- Should I check if the error message is exactly the one I set? It feels strange to do so, I would usually create error message instead of working with plain strings
- I've mirrored similar test behaviors that included a count of test runs. Is that necessary? Are there cases where we want to make sure that no other tests data is added?

Thank you for your feedbacks :)

rprouse:
This has been awaiting contributor for a couple of years. Closing.

PR #3379, opened 2019-09-28
Fix issues with Pass method on TestFilter's (issue #3328 )

### LINKED ISSUE #3328: Problems when using a mixture of Not and Or filters in NUnit framework 3.12.0
In short, the NUnit 3 framework does not select the correct tests when using a filter like `!(test!='Some.Namespace'|...)`.

To give a minimal example, suppose that I have an assembly (referencing NUnit framework 3.12.0) with 1 test, called `Namespace.Fixture.Test`, without any attributes (and hence no categories). Running the tests with the filter `!(test!='Namespace.Fixture'|cat=='X')` will execute 0 tests. However, if it references NUnit framework version 2.6.4, it runs the test `Namespace.Fixture.Test`.

This is, of course, a very contrived example, and an easy workaround is to apply De Morgan and use the filter `test=='Namespace.Fixture'&cat!='X'`.

I believe that this relates to issue #3231.

I have tested this with version 3.10.0 of the console runner, installed with Chocolatey, as well as version 3.7.0 of the NUnit V2 Framework Driver, installed the same way. Moreover, I have added [FilterTest.zip](https://github.com/nunit/nunit/files/3427218/FilterTest.zip) with projects producing assemblies `NUnit2FilterTest.dll` and `NUnit3FilterTest.dll` containing the test mentioned above and referencing NUnit framework 2.6.4 and NUnit 3.12.0, respectively.  After building the projects, the problem can be reproduced with:
```
nunit3-console.exe NUnit2\bin\Debug\NUnit2FilterTest.dll --where "!(test!='Namespace.Fixture'|cat=='X')"
nunit3-console.exe NUnit3\bin\Debug\NUnit3FilterTest.dll --where "!(test!='Namespace.Fixture'|cat=='X')"
```

okknudsen:
This fixes #3328 

The problem was the difference in whether we are using a filter to exclude tests or include them when calling `Pass`. In the former case, we want to look back through parents only, but in the latter case, we want to look forward through children as well.

This fix makes sure that this decision is taken by the inner most filters (leafs I guess) by applying De Morgan to ensure that negations are pushed through `AndFilters` and `NotFilters`.

If I forgot something, please let me know. This is my first time contributing.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Filters/NotFilter.cs
+++ b/src/NUnitFramework/framework/Internal/Filters/NotFilter.cs
@@ -47,20 +47,13 @@ namespace NUnit.Framework.Internal.Filters
 
         /// <summary>
         /// Determine if a particular test passes the filter criteria.
-        /// 
-        /// Overriden in NotFilter so that
-        /// 1. Two nested NotFilters are simply ignored
-        /// 2. Otherwise, we only look at the test itself and parents, ignoring
-        /// any child test matches.
         /// </summary>
         /// <param name="test">The test to which the filter is applied</param>
+        /// <param name="negated">If set to <c>true</c> we are carrying a negation through</param>
         /// <returns>True if the test passes the filter, otherwise false</returns>
-        public override bool Pass(ITest test)
+        public override bool Pass(ITest test, bool negated)
         {
-            var secondNotFilter = BaseFilter as NotFilter;
-            return secondNotFilter != null
-                ? secondNotFilter.BaseFilter.Pass(test) 
-                : !BaseFilter.Match (test) && !BaseFilter.MatchParent (test);
+            return BaseFilter.Pass(test, !negated);
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Internal/TestFilter.cs
+++ b/src/NUnitFramework/framework/Internal/TestFilter.cs
@@ -65,6 +65,24 @@ namespace NUnit.Framework.Internal
         /// <returns>True if the test passes the filter, otherwise false</returns>
         public virtual bool Pass(ITest test)
         {
+            return Pass(test, false);
+        }
+
+        /// <summary>
+        /// Determine if a particular test passes the filter criteria. The default
+        /// implementation checks the test itself, its parents and any descendants.
+        ///
+        /// Derived classes may override this method or any of the Match methods
+        /// to change the behavior of the filter.
+        /// </summary>
+        /// <param name="test">The test to which the filter is applied</param>
+        /// <param name="negated">If set to <c>true</c> we are carrying a negation through</param>
+        /// <returns>True if the test passes the filter, otherwise false</returns>
+        public virtual bool Pass(ITest test, bool negated)
+        {
+            if (negated)
+                return !Match(test) && !MatchParent(test);
+
             return Match(test) || MatchParent(test) || MatchDescendant(test);
         }
 
@@ -211,7 +229,7 @@ namespace NUnit.Framework.Internal
                 return true;
             }
 
-            public override bool Pass( ITest test )
+            public override bool Pass( ITest test, bool negated )
             {
                 return true;
             }
--- a/src/NUnitFramework/tests/Internal/Filters/DeMorganOnFiltersTest.cs
+++ b/src/NUnitFramework/tests/Internal/Filters/DeMorganOnFiltersTest.cs
@@ -1,0 +1,100 @@
+// ***********************************************************************
+// Copyright (c) 2019 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System.Collections.Generic;
+using NUnit.Framework.Interfaces;
+
+namespace NUnit.Framework.Internal.Filters {
+    public class DeMorganOnFiltersTest : TestFilterTests
+    {
+        private static List<TestFilter[]> filterPairs;
+
+        static DeMorganOnFiltersTest()
+        {
+            var filterParts = new TestFilter[]
+            {
+                new CategoryFilter("Dummy"),
+                new MethodNameFilter("Test1"),
+                new PropertyFilter("Priority", "High"),
+                new PropertyFilter("Priority", "Low")
+            };
+
+            filterPairs = new List<TestFilter[]>();
+            foreach (var part1 in filterParts)
+            foreach (var part2 in filterParts)
+            {
+                var and = new AndFilter(part1, new NotFilter(part2));
+                var or = new OrFilter(new NotFilter(part1), part2);
+                filterPairs.Add(new TestFilter[2] {new NotFilter(and), or});
+                filterPairs.Add(new TestFilter[2] {and, new NotFilter(or)});
+            }
+        }
+
+        private IEnumerable<ITest> GetTests()
+        {
+            var q = new Queue<ITest>();
+            q.Enqueue(_topLevelSuite);
+            while (q.Count > 0)
+            {
+                var test = q.Dequeue();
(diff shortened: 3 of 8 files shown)


src/NUnitFramework/framework/Internal/Filters/NotFilter.cs
```diff
@@ -54,13 +54,11 @@ public NotFilter( TestFilter baseFilter)
         /// any child test matches.
         /// </summary>
         /// <param name="test">The test to which the filter is applied</param>
+        /// <param name="negated">If set to <c>true</c> we are carrying a negation through</param>
         /// <returns>True if the test passes the filter, otherwise false</returns>
-        public override bool Pass(ITest test)
+        public override bool Pass(ITest test, bool negated = false)
```
CharliePoole: This is very clever and obviously central to the algorithm. Others should decide whether this is a breaking change. Obviously, it changes the interface by changing a method, rather than merely adding one. Are there .NET languages for which the use of a default value is non-compliant?

okknudsen: No problem @CharliePoole :)

unfortunately, I don't know about the compliance, but I will be happy to change it to a separate method if this is preferable. This was actually my thought to begin with, until I stated making the change. I went with this solution because I thought it might be more conceptually clear. However, this is probably not much, and might even entirely be my subjective view.

okknudsen:
Ohh sorry @CharliePoole . I noticed that I have forgotten to remove some documentation, and pushed a commit, but I'm a bit new to the pull request model, so I did not realize that it would take effect here. It seems like it made me dismissed your review, if I understand it correctly.

CharliePoole:
Yeah, GitHub can't tell whether you made an important change or not. I re-approved. Note that it takes a second approval before merging.

src/NUnitFramework/framework/Internal/Filters/NotFilter.cs
jnm2: If there is an already-compiled assembly out there that calls one of these `Pass` methods, this change will result in MissingMethodException because the compiled binary has a hard reference to the `bool Pass(ITest)` signature and the .NET runtime doesn't have the ability to forward that to the new `bool Pass(ITest, bool)` signature.

Adding an interface method is also a breaking change for anyone out there who implements ITestFilter themselves. Once published, interfaces can never be changed without breaking somebody: https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md#members. This was the motivation for the C# 8 Default Interface Members feature which only works on the .NET Core 3.0+ runtime, but that's not useful to us unless we stop supporting .NET Framework.

Can we think up a way around this? The .NET Framework used the tactic of adding virtual methods to abstract base classes implementing the existing functions, such as `DbConnection`, since `IDbConnection` can never be changed.

We've had several complaints about changes like this, so I want to either do something non-breaking or wait until NUnit 4 to fix the bug.

jnm2: If there is an already-compiled assembly out there that calls one of these `Pass` methods, this change will result in MissingMethodException because the compiled binary has a hard reference to the `bool Pass(ITest)` signature and the .NET runtime doesn't have the ability to forward that to the new `bool Pass(ITest, bool)` signature.

Adding an interface method is also a breaking change for anyone out there who implements ITestFilter themselves. Once published, interfaces can never be changed without breaking somebody: https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md#members. This was the motivation for the C# 8 Default Interface Members feature which only works on the .NET Core 3.0+ runtime, but that's not useful to us unless we stop supporting .NET Framework.

Can we think up a way around this? The .NET Framework used the tactic of adding virtual methods to abstract base classes implementing the existing functions, such as `DbConnection`, since `IDbConnection` can never be changed.

We've had several complaints about changes like this, so I want to either do something non-breaking or wait until NUnit 4 to fix the bug.

src/NUnitFramework/framework/Internal/TestFilter.cs
```diff
@@ -62,9 +62,13 @@ public bool IsEmpty
         /// to change the behavior of the filter.
         /// </summary>
         /// <param name="test">The test to which the filter is applied</param>
+        /// <param name="negated">If set to <c>true</c> we are carrying a negation through</param>
         /// <returns>True if the test passes the filter, otherwise false</returns>
-        public virtual bool Pass(ITest test)
+        public virtual bool Pass(ITest test, bool negated = false)
```
jnm2: Could we revert the change to the interface and the change to this method and add a new virtual `bool Pass(ITest, bool)` method? That would enable libraries that call the existing methods to continue working against both past and future versions of NUnit at the same time.

https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md#members

src/NUnitFramework/framework/Internal/Filters/NotFilter.cs
CharliePoole: @jnm2 In spite of the Microsoft guidelines, I think we could add a member here without much impact.

Of course it is possible that somebody has taken our code and used it in unexpected ways. However...

1. There is no way to create a usable custom filter without modifying the framework. In practice this means that any new filter has to be implemented by us and is peculiar to the version of the framework that contains support for it. That is to say, there is no extensibility mechanism for filters.

2. The filter interface is really internal, as are almost all of those in the `Interfaces` directory. I believe they were once in the `Internal` directory and were moved to reduce clutter - probably not a good decision. It's primary use is to allow easy testing.

3. The API for the framework is reflection-based rather than interface-based anyway.

4. Microsoft's rules are based on their bad experience. Someone in the company has said that anything you make accessible to users immediately becomes a feature that must be supported. We can all sympathize with that! However, I don't think we need to support unreasonable usage of accessible members, i.e. usage that goes beyond the fairly obvious intention of the interface, which in this case is to get the framework to filter tests.

5. This __will__ impact anyone who has modified the source of the framework for their own purposes, adding new filters. But only if they decide to make the same change themselves, which seems fine to me.

`<meta>`
It would be great if we could have a written, approved and published standard around breaking changes. We've been pushing around the same arguments for a couple of years now without coming to much conclusion. The key thing to define IMO is what is meant by a "published API".
`</meta>`

jnm2: Okay, if we are sure that no one will be implementing it, then adding a new method will be fine. I haven't tried to look for a problem with that, but I trust you.

> The API for the framework is reflection-based rather than interface-based anyway.

That's only the API for driving the framework from an engine. The Assert class and everything users write `using NUnit.Framework` for are another API, one which we care about because users reasonably expect that we will not break their libraries or code (unless maybe under Internal, but even there we tell them that is their only option sometimes and we still shouldn't break that).

CharliePoole: As I've written before we have a number of APIs. I believe each of them has to have its own standard. Here, I was talking about the "Framework API" for calling the framework to run tests. The name comes from our docs.

Asserts, etc. are part of the test-writing API, which we don't call out as such, both obviously exists. There is also an extensibility API, which is partially documented.

jnm2: Right, so then I'm not sure how point 3 has any bearing on this PR and the complaints I mentioned that we've been getting when we change APIs that we ship out of this repo.

CharliePoole: Only bearing was if you were considering this as part of that API.

jnm2: Okay, no. Since few people maintain NUnit engines/drivers, most of my thought goes to the user-testing and extensibility APIs that we get the complaints about.

src/NUnitFramework/framework/Internal/TestFilter.cs
okknudsen: Hi @jnm2, the change can indeed be made by adding a virtual method to the `TestFilter` class instead, so that no changes are made to the interface.

I have implemented the change and can update the PR if you would like, but as I am new to PR's could you tell me if I should just push a new commit to the branch? I'm unsure how it plays together with all the references to the code in these comments.

CharliePoole: If you just push, then the PR is updated. But if you are reversing a lot of changes, it's cleaner to first do a reset to the prior HEAD followed by a forced push. (If that's new to you, the simpler way us OK.)

okknudsen: I did a force push. I'm comfortable enough with Git. It is mostly the Github part that's new to me :)

src/NUnitFramework/tests/Internal/Filters/MockTestFilter.cs
```diff
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
```
okknudsen: Ohh, I might have removed a byte order mark here.

src/NUnitFramework/framework/Internal/TestFilter.cs
okknudsen: I did not realize that this was waiting on me because I was supposed to have pressed resolve. Sorry :(

okknudsen:
All right @CharliePoole, thanks. I will remember that for next time.
@jnm2, I have made changes so that `ITestFilter` stays unaltered.

src/NUnitFramework/tests/Internal/Filters/DeMorganOnFiltersTest.cs
```diff
@@ -0,0 +1,100 @@
...
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System.Collections.Generic;
+using NUnit.Framework.Interfaces;
+
+namespace NUnit.Framework.Internal.Filters {
+    public class DeMorganOnFiltersTest : TestFilterTests
```
rprouse: Nicely named 😄

PR #3382, opened 2019-09-29
Obsolete ListMapper

### LINKED ISSUE #3381: Obsolete ListMapper
`ListMapper.Property` throws NullReferenceException if you pass a collection with a null element to `new ListMapper`. I was going to fix it to throw a proper ArgumentException, then I realized that it didn't seem very useful and in fact is not used (or tested) at all.

Since it's connected to the AssertionHelper class which is already obsoleted, could we delete it? Or should we mark it obsolete instead? We shouldn't delete it if we think someone out there is using it, because they might have a binary dependency on it. I'm having a hard time picturing someone using it, given what it does.

One benefit to deleting it is that intellisense won't suggest it if you type `List` without angle brackets and press Ctrl+. to add a using directive.

```diff
namespace NUnit.Framework
{
    [Obsolete("The AssertionHelper class has been deprecated and will be removed in a future "
        + "release. Please consider using the NUnit.StaticExpect NuGet package instead.")]
    public class AssertionHelper
    {
-        public ListMapper Map(ICollection original);
    }

-   public class List
-   {
-       public static ListMapper Map(ICollection actual);
-   }

-   public class ListMapper
-   {
-       public ListMapper(ICollection original);

-       public ICollection Property(string name);
-   }
}
```

jnm2:
Closes #3381 (see discussion).

### DIFF
--- a/src/NUnitFramework/framework/List.cs
+++ b/src/NUnitFramework/framework/List.cs
@@ -8,10 +8,10 @@
 // distribute, sublicense, and/or sell copies of the Software, and to
 // permit persons to whom the Software is furnished to do so, subject to
 // the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be
 // included in all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
 // EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
 // MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/framework/List.cs
```diff
@@ -21,25 +21,30 @@
...
-		public static ListMapper Map( ICollection actual )
-		{
-			return new ListMapper( actual );
-		}
-	}
+    /// <summary>
+    /// The List class is a helper class with properties and methods
+    /// that supply a number of constraints used with lists and collections.
+    /// </summary>
+    [Obsolete("The ListMapper class has been deprecated and will be removed in a future release. "
```
CharliePoole: "The List class"?

src/NUnitFramework/framework/List.cs
```diff
@@ -21,25 +21,30 @@
...
+        + "Please use the extension method System.Linq.Enumerable.Select instead.")]
+    public class List
+    {
+        /// <summary>
+        /// List.Map returns a ListMapper, which can be used to map
+        /// the original collection to another collection.
+        /// </summary>
+        /// <param name="actual"></param>
+        /// <returns></returns>
+        [Obsolete("The ListMapper class has been deprecated and will be removed in a future release. "
```
CharliePoole: If the class is obsolete, is it necessary to mark each method as well? This applies in several other places as well.

jnm2: This is a good point. We don't need duplicate warnings showing up.

The only way it would be useful to mark the method is if a third-party library does this:

```cs
public class ListEx : NUnit.Framework.List
{
}
```

If someone using that library does `ListEx.Map`, they wouldn't see an obsoletion warning. Seems far-fetched, will remove.

jnm2: Or, if the third-party library does this:

```cs
public class Foo
{
    public List List { get; }
}
```

And someone using the library does `new Foo(...).List.Map("foo")`. Still seems unlikely.

Issue #3320, opened 2019-07-13
New targets to replace EOL platforms


jnm2:
Microsoft stopped supporting .NET Core 2.0 on October 1, 2018 and .NET Core 1.0 and 1.1 on June 27, 2019. (https://dotnet.microsoft.com/platform/support/policy/dotnet-core)

Any version of .NET Standard lower than 2.0 is not needed because the only difference moving to 2.0 makes is that you have to move from UWP 10.0 to UWP 10.0.16299. (https://docs.microsoft.com/en-us/dotnet/standard/net-standard) There should be no reason _not_ to move to UWP 10.0.16299 at this point because it targets every version of Windows 10 that Microsoft supports.
@jrjrguo You asked for UWP 10.0 support about a year ago and we shipped it pretty soon after that. I remember from your GitHub profile that you work for Microsoft, so I'm assuming you have no need to support versions of Windows 10 released more than two years ago and therefore no need to target the old UWP 10.0. What are your thoughts?

### Things I would like to do

@nunit/framework-team What is your input on this?

1. [**Done**] Stop running our tests using netcoreapp1.1 and start using netcoreapp2.1, netcoreapp2.2, and netcoreapp3.0 (preview 7 is go-live and should be out next week)

2. [**Done**] Remove the NUnit framework netstandard1.4 target which simplifies our test matrix

3. Add a new netcoreapp2.1 target to NUnit framework so that TimeoutAttribute can actually use threads instead of tasks on .NET Core
  Now that I think of it, this would open the door to using AsyncLocal and other things not present in net45.

Whatever we decide here will affect https://github.com/nunit/nunit/issues/3223 where I want to begin testing every assembly we ship on every runtime we support.

jnm2:
I don't personally like the idea of requiring contributors to have preview bits on their machines in order to run tests, so until we autodetect and skip that (#3223) let's just add up to netcoreapp2.2.

rprouse:
I am fine with all of this, the only question is, if we want to add a `netcoreapp2.1` target and bring in support for `AsyncLocal`, should we also add a .NET 4.6+ target for the same thing on the full framework? Are there other constructs in 4.6+ that we should be supporting but aren't?

jnm2:
@rprouse Perhaps yes? I'm not sure yet. AsyncLocal is just something we use internally in our .NET Standard builds IIRC, but it can't hurt to have more uniform behavior by adding a `net46` target.

[...]

PR #3383, opened 2019-09-29
Drop netstandard1.4 and stop testing on end-of-life versions of .NET Core

jnm2:
Related to https://github.com/nunit/nunit/issues/3320.

The.NET Standard 1.4 build is gone and the .NET Standard 2.0 build is no longer tested on .NET Core 2.0 which is out of support. Instead, it is tested on .NET Core 2.1, 2.2, *and* 3.0.


jnm2:
@rprouse Did you remove the Travis validation build, or is Travis maybe not responding to webhooks right now? I started getting only four checks in the last few updates to this PR, but I can start builds manually.

jnm2:
I actually cannot figure out how to get Ubuntu on Travis to work without a large amount of further effort.

> ```
> Tool 'cake.tool' (version '0.35.0') was successfully installed.
> dotnet-cake: command not found
> ```

@rprouse Can you confirm whether we are dropping Travis? If so, I may as well delete the yaml because it won't run in its current state in this branch.

CharliePoole:
I use the cake tool for my Travis builds of TestCentric. My script allows for the fact that the tool is installed in a hidden subdirectory of your home directory.

See... https://github.com/TestCentric/testcentric-gui/blob/master/.travis.yml#L18

jnm2:
That's the same line I started with, I think. Can you see a difference, excluding the latest commit?

CharliePoole:
Looks the same to me except for versions.

rprouse:
I removed Travis as a required build, but I don't think I removed it completely. I'm not sure why it isn't showing up. I'd be fine with dropping it though since we are building Linux on Azure DevOps and that has been much more reliable than Travis.

jnm2:
@rprouse I deleted the Travis file like I mentioned above because leaving it alone would have still left it broken, but I can do something else if you prefer!

PR #3400, opened 2019-10-14
Contains.Key now working for IDictionary

### LINKED ISSUE #3331: Contains.Key no longer working for IDictionary
I had some tests that were checking for a key in the exception data property that have stopped working since updating the version of nunit.

Narrowing it down I believe it's because it only implements the non generic `IDictionary` and maybe a change has been made that hasn't taken that into account. This has stopped working since `3.11`

It can be replicated fairly easily.
`Assert.That(new Exception().Data, Contains.Key("hello"));`

Previously this would return
```
Expected: dictionary containing key "hello"
But was:  <empty>
```
but now returns
```
System.ArgumentException : The ListDictionaryInternal value must have a ContainsKey or Contains(TKey) method.
```

stevenaw:
Fixes #3331 
This PR fixes a regression in `3.11` where using a DictionaryContainsKeyConstraint on a non-generic dictionary would fail. All existing tests pass (except for one which was testing that this specific case threw an exception).

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
@@ -93,19 +93,27 @@ namespace NUnit.Framework.Constraints
         {
             if (_isDeprecatedMode)
             {
-                var dictionary = ConstraintUtils.RequireActual<IDictionary>(actual, nameof(actual));
-                foreach (object obj in dictionary.Keys)
-                    if (ItemsEqual(obj, Expected))
-                        return true;
-
-                return false;
+                var requiredDictionary = ConstraintUtils.RequireActual<IDictionary>(actual, nameof(actual));
+                return MatchesKeysIteratively(requiredDictionary);
             }
 
             var method = GetContainsKeyMethod(actual);
             if (method != null)
                 return (bool)method.Invoke(actual, new[] { Expected });
 
+            if (actual is IDictionary dictionary)
+                return MatchesKeysIteratively(dictionary);
+
             throw new ArgumentException($"The {TypeHelper.GetDisplayName(actual.GetType())} value must have a ContainsKey or Contains(TKey) method.");
+        }
+
+        private bool MatchesKeysIteratively(IDictionary dictionary)
+        {
+            foreach (var obj in dictionary.Keys)
+                if (ItemsEqual(obj, Expected))
+                    return true;
+
+            return false;
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
@@ -93,27 +93,19 @@ namespace NUnit.Framework.Constraints
         {
             if (_isDeprecatedMode)
             {
-                var requiredDictionary = ConstraintUtils.RequireActual<IDictionary>(actual, nameof(actual));
-                return MatchesKeysIteratively(requiredDictionary);
+                var dictionary = ConstraintUtils.RequireActual<IDictionary>(actual, nameof(actual));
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -448,6 +449,7 @@ public class TestNonGenericDictionary : IDictionary
             public TestNonGenericDictionary(int key)
             {
                 _key = key;
+                Keys = new[] { key };
```
stevenaw: Note to reviwer: Our custom `IDictionary` implementation for testing was previously throwing a `NullReferenceException` if you tried to reference the `Keys` property of it.

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -189,7 +190,7 @@ public void ShouldCallContainsKeysMethodOnPlainDictionary()
         {
             var dictionary = new TestNonGenericDictionary(99);
 
-            Assert.Catch<ArgumentException>(() => Assert.That(dictionary, Does.ContainKey(99)));
```
stevenaw: Note to reviewer: This test was previously detecting that we were throwing an exception when using an `IDictionary`. I've updated it to pass. I didn't add any additional tests because we already have "Should fail" tests in place for non-dictionaries:
https://github.com/nunit/nunit/blob/a5adf028c101283beecad0b49104468efe75b90f/src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs#L68-L76

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -231,7 +231,7 @@ private static MethodInfo FindContainsKeyMethod(Type type)
             var methods = type.GetMethods(BindingFlags.Instance | BindingFlags.Public);
             var method = methods.FirstOrDefault(m =>
                 m.ReturnType == typeof(bool)
-                && m.Name == "ContainsKey"
+                && (m.Name == "ContainsKey" || (m.Name == "Contains" && m.DeclaringType == typeof(IDictionary)))
```
stevenaw: So as not to get the `Contains` method from other non-generic collections

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyConstraintTests.cs
```diff
@@ -189,7 +190,7 @@ public void ShouldCallContainsKeysMethodOnPlainDictionary()
...
+            Assert.That(dictionary, Does.ContainKey(99));
```
mikkelbu: Should we also have a line
```csharp
Assert.That(dictionary, !Does.ContainKey(35));
```
so that the test have the same structure as `ShouldCallContainsKeysMethodOnDictionary`

stevenaw: We should, thanks for catching this. I've just pushed an update for this

mikkelbu:
Looks good to me, but as usual we need a second review before merging.

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
jnm2: Since ` == typeof(IDictionary)` shows that you have this method in mind in particular:

```suggestion
                && (m.Name == "ContainsKey" || (m.Name == nameof(IDictionary.Contains) && m.DeclaringType == typeof(IDictionary)))
```

stevenaw:
Good idea @jnm2 , thanks! I've incorporated your suggestion.

jnm2:
@stevenaw Thanks for your help!

Issue #52, opened 2013-11-03
Self-contained item in array causes stack overflow


CharliePoole:
See https://bugs.launchpad.net/nunit-3.0/+bug/1051847

Fix needs to be checked to see if it applies to nunit 3.0 and, if so, ported.

constructor-igor:
NUnit v3Beta4Console crashed on test:

```c#
    [TestFixture]
    public class NUnitIssue52
    {
        class SelfContainer : IEnumerable
        {
            public IEnumerator GetEnumerator() { yield return this; }
        }
        [Test]
        public void SelfContainedItemFoundInArray()
        {
            var item = new SelfContainer();
            var items = new SelfContainer[] { new SelfContainer(), item };

            // work around
            Assert.True(((ICollection<SelfContainer>)items).Contains(item));

            // causes StackOverflowException
            Assert.Contains(item, items);
            Console.WriteLine("test completed");
        }
    }
```

![image](https://cloud.githubusercontent.com/assets/1849690/10331725/39a225ca-6ce0-11e5-8c2c-5081d9886da9.png)

CharliePoole:
Does a timeout prevent the crash and report the test as failing?

constructor-igor:
at least 4 seconds (on my computer) between message "... to StackOverflowException" and crash dialog.

CharliePoole:
There's no doubt it's a bug. Did you come across it in your normal work or just by trying to make it happen following this issue report? In other words, is there a reason to push up the priority?

constructor-igor:
I found the crash during the issue's investigation. So, I don't see a reason to push up the priority.
On the other hand, if nunit console catches StackOverflowException exception, why it cannot process it without crash?

CharliePoole:
.NET 2.0 and later won't let you continue after a stack overflow.

stevenaw:
I tested this last night and can confirm its still an issue in master (in the EnumerableComparer). I'd be happy to take this on to try and resolve.

stevenaw:
Thanks for merging the PR @mikkelbu 

While that fixed a recursion issue within the comparers, I discovered during development that there's a separate overflow condition specific to unequal dictionaries that occurs when trying to output the key-values after a failed assertion. I opted not to tackle it in the original PR because it was a much trickier change to make without it becoming a breaking change. Do you think it's worth keeping this issue open until that is fixed too, or should that be split off into a separate issue?

mikkelbu:
[...]

PR #3406, opened 2019-10-20
Self-contained item in array no longer causes stack overflow

stevenaw:
Partial resolution to #52 

This PR fixes the _comparer_ side of the stack overflow issue noted. However, a second issue is still outstanding in `MsgUtils.FormatValue()` for dictionary types. I've setup separate test cases + tests in `NUnitEqualityComparerTests` to call this out. This means that self-referential array + object graphs will now work for `Assert.Contains()` and the like, but not for self-referential dictionaries.

I've decided not to address the `MsgUtils` / `Dictionary` side in this PR because I am having a hard time ensuring that the required changes are made in a 100% backwards-compatible way (both behavioural + binary). For that reason, as well as to mitigate risk, I'd like to propose reviewing this PR as-is, and to defer adding Dictionary support for a follow-up PR.


src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -154,10 +154,31 @@ public IList<FailurePoint> FailurePoints
...
+        public bool AreEqual(object x, object y, ref Tolerance tolerance, bool topLevelComparison)
+        {
+            return AreEqual(x, y, ref tolerance, new ComparisonState()
+            {
+                TopLevelComparison = topLevelComparison,
+                Comparisons = new List<ComparisonState.Comparison>()
+            });
+        }
+
+        internal bool AreEqual(object x, object y, ref Tolerance tolerance, ComparisonState state)
```
stevenaw: Note to reviewer: All the internal comparer signatures were updated to receive a `ComparisonState` object instead of a `bool`. `ComparisonState` is an internal struct which tracks, among other things, the existing "topLevelComparison" state as well as a list of objects already compared to each other.

To ensure backwards-compatibiliity, the public API of NUnitEqualityComparer was left as-is, with an internal overload added to receive this new object.

src/NUnitFramework/tests/Constraints/NUnitEqualityComparerTests.cs
```diff
@@ -306,6 +308,89 @@ public void IEnumeratorIsDisposed()
...
+            equalityComparer.ExternalComparers.Add(new DetectRecursionComparer(30));
+
+            Assert.DoesNotThrow(() => equalityComparer.AreEqual(x, y, ref tolerance));
+        }
+
+        public static IEnumerable<TestCaseData> GetRecursiveComparerTestCases()
+        {
+            // Separate from 'GetRecursiveContainsTestCases' until a stackoverflow issue in 
+            // 'MsgUtils.FormatValue()' can be fixed for the below cases
+            foreach (var testCase in GetRecursiveContainsTestCases())
```
stevenaw: As noted in the PR description, a separate stack overflow issues exists in `MsgUtils.FormatValue` for dictionary types that I'd like to defer fixing for a follow-up PR

src/NUnitFramework/tests/Constraints/NUnitEqualityComparerTests.cs
```diff
@@ -306,6 +308,89 @@ public void IEnumeratorIsDisposed()
...
+            Assert.That(y, Contains.Item(x));
+            Assert.That(y, Does.Contain(x));
+        }
+
+        [TestCaseSource(nameof(GetRecursiveComparerTestCases))]
+        public void SelfContainedItemDoesntRecurseForever<T>(T x, ICollection y)
+        {
+            var equalityComparer = new NUnitEqualityComparer();
+            var tolerance = Tolerance.Default;
+            equalityComparer.ExternalComparers.Add(new DetectRecursionComparer(30));
```
stevenaw: A naive + simple way to detect stack overflows when unit testing the `NUnitEqualityComparer` directly. For each call to the same instance of `NUnitEqualityComparer`, use this external comparer to hook into a check of the current execution stack depth.

src/NUnitFramework/framework/Constraints/Comparers/DictionariesComparer.cs
```diff
@@ -54,8 +54,12 @@ internal DictionariesComparer(NUnitEqualityComparer equalityComparer)
                 return false;
 
             foreach (object key in xDictionary.Keys)
-                if (!_equalityComparer.AreEqual(xDictionary[key], yDictionary[key], ref tolerance, topLevelComparison: false))
+            {
+                state.TopLevelComparison = false;
+                if (!_equalityComparer.AreEqual(xDictionary[key], yDictionary[key], ref tolerance, state))
```
jnm2: I find it very much simpler to think about when the state type is readonly because the syntax then shows that shared state is not being mutated, e.g. `state.WithTopLevelComparison(false)`.

(Shared state isn't being mutated here because `state` is a value type and `TopLevelComparison` is backed by a value-typed field, but I'm still happier when the syntax makes it obvious.)

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
```diff
@@ -0,0 +1,36 @@
...
+        public bool TopLevelComparison { get; set; }
+        public List<Comparison> Comparisons { get; set; }
+
+        public void RecordComparison(object x, object y)
+        {
+            Comparisons.Add(new Comparison()
+            {
+                X = x,
+                Y = y
+            });
```
jnm2: Even though `Comparison` is a struct, please make X and Y get-only properties and use `new Comparison(x, y)`. Guarantees of immutability reduce the cognitive load on the reviewers and future maintainers of the code.

src/NUnitFramework/framework/Constraints/Comparers/DictionariesComparer.cs
jnm2: I find it very much simpler to think about when the state type is readonly because the syntax then shows that shared state is not being mutated, e.g. `state.WithTopLevelComparison(false)`.

(Shared state isn't being mutated here because `state` is a value type and `TopLevelComparison` is backed by a value-typed field, but I'm still happier when the syntax makes it obvious.)

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            public object X { get; set; }
+            public object Y { get; set; }
+        }
+
+        /// <summary>
+        /// Flag indicating whether or not this is the top level comparison.
+        /// </summary>
+        public bool TopLevelComparison { get; set; }
+        public List<Comparison> Comparisons { get; set; }
```
jnm2: Please make this a private readonly field to make it easier to think about the possible mutations and to streamline the usability of this struct for future users.

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
```diff
@@ -0,0 +1,36 @@
...
+        public void RecordComparison(object x, object y)
+        {
+            Comparisons.Add(new Comparison()
+            {
+                X = x,
+                Y = y
+            });
+        }
+
+        public bool HasCompared(object x, object y)
```
jnm2: (optional) `DidCompare`?

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -167,20 +184,26 @@ public bool AreEqual(object x, object y, ref Tolerance tolerance, bool topLevelC
...
+                bool? result = comparer.Equal(x, y, ref tolerance, state);
                 if (result.HasValue)
                     return result.Value;
             }
 
             return x.Equals(y);
         }
 
+
+
```
jnm2: Was this intentional?

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -154,7 +154,24 @@ public IList<FailurePoint> FailurePoints
...
+        /// <summary>
+        /// Compares two objects for equality within a tolerance.
+        /// </summary>
+        public bool AreEqual(object x, object y, ref Tolerance tolerance, bool topLevelComparison)
+        {
+            return AreEqual(x, y, ref tolerance, new ComparisonState()
+            {
+                TopLevelComparison = topLevelComparison,
+                Comparisons = new List<ComparisonState.Comparison>()
+            });
```
jnm2: A constructor should initialize all required fields. It wouldn't make sense for `Comparisons` to be null, so I'd suggest a `bool` constructor that creates the list.

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -167,20 +184,26 @@ public bool AreEqual(object x, object y, ref Tolerance tolerance, bool topLevelC
             if (object.ReferenceEquals(x, y))
                 return true;
 
+            if (!state.TopLevelComparison && state.HasCompared(x, y))
+                return false;
+            state.RecordComparison(x, y);
```
jnm2: What happens if you only check `state.HasCompared(x, y)`, effectively replacing `TopLevelComparison`?

jnm2: The call pattern strongly suggests to me something like this, in the same vein as `bool HashSet<>.Add`:

```cs
if (!state.RecordComparison(x, y)) return false;
```

jnm2: Now I'm worried about something. And it's connected to `List<Comparison>` not being an immutable stack. (I really dislike mutable state. It's hard to think about properly.)

What about this scenario?

```cs
var equalInstance1 = new[] { 1 };
var equalInstance2 = new[] { 1 };

var x = new[] { equalInstance1, equalInstance1 };
var y = new[] { equalInstance2, equalInstance2 };

Assert.True(equalityComparer.AreEqual(x, y, ref tolerance));
```

If I understand the code properly, `equalInstance1` and `equalInstance2` will be compared once and then the pair will be recorded. Then they will be compared a second time and the line above will return `false` when it should return `true`.

A fix for this would be to use ImmutableStack or ImmutableQueue instead of List, and each time you chain to a comparer that is a level deeper (vertically, not horizontally), you do `comparisons.Push(new Comparison(x, y))`. A singly-linked list would work well since we can't reference System.Collections.Immutable. This is my favorite approach because it means that `state` is completely immutable, just like the original `topLevelComparison` was, and super easy to think about.

stevenaw: Oh good catch! When I was writing this I was mostly concerned with duplicates down the call stack, but I'd completely forgotten about allowing horizontal duplicates. I'm going to have to play around a bit more with this.

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
stevenaw: Oops. Nope. Thanks :)

stevenaw:
@jnm2 Thanks for the review. I've pushed changes to address _most_ of your feedback, but the horizontal duplicates issue is still an outstanding bug.

On the general theme of using readonly structs, I think I saw a conversation in a ticket a while back about which C# language features to target in the csproj file. `nunit.framework.csproj` currently targets `LangVersion` 7.0, but readonly structs are available in 7.2. The "Building.md" guide mentions Visual Studio 2017 as a prerequisite, of which some versions support newer language features and some do not. Am I correct in thinking we are specifically targeting 7.0 in order to support building on all versions of VS 2017? Or are we ok to target 7.2 in order to enforce readonly structs at the compiler level (instead of by convention).

jnm2:
Cool! By "readonly struct" I meant that the fields (incl. backing fields) are readonly. That's the important part to me. It would then make sense to put `readonly` on the struct itself if we used C# 7.2

I'm not sure which update of Visual Studio 2017 began supporting C# 7.2. @rprouse, would it be okay to require the very latest update, Update 9, if using VS2017? It's been out since November 2018. I don't think it's even possible to install older updates.

stevenaw:
> Cool! By "readonly struct" I meant that the fields (incl. backing fields) are readonly. That's the important part to me. It would then make sense to put readonly on the struct itself if we used C# 7.2.

Yep, I thought that's what you meant. I think we're on the same page. I did some digging and it looks like C# 7.2 support was added in 15.5 (released Dec 4, 2017).

As far as the rest of the changes, I think I've incorporated most of the ideas now.

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
```diff
@@ -0,0 +1,62 @@
...
+        private ComparisonState(bool topLevelComparison, Comparison[] comparisons)
+        {
+            TopLevelComparison = topLevelComparison;
+            _comparisons = comparisons;
+        }
+
+        public ComparisonState PushComparison(bool topLevelComparison, object x, object y)
+        {
+            var comparisons = new Comparison[_comparisons.Length + 1];
+            _comparisons.CopyTo(comparisons, 0);
```
rprouse: what is your thinking of using an array and copying every time you add to it instead of a collection class like List?

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
```diff
@@ -0,0 +1,62 @@
...
+            _comparisons = new Comparison[0];
+        }
+
+        private ComparisonState(bool topLevelComparison, Comparison[] comparisons)
+        {
+            TopLevelComparison = topLevelComparison;
+            _comparisons = comparisons;
+        }
+
+        public ComparisonState PushComparison(bool topLevelComparison, object x, object y)
```
rprouse: You only ever pass in `false` for `topLevelComparison`, so is this needed? Don't we only start with a top-level and all pushes are therefore children?

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
```diff
@@ -0,0 +1,62 @@
...
+            return new ComparisonState(
+                topLevelComparison,
+                comparisons
+            );
+        }
+
+        public bool DidCompare(object x, object y)
+        {
+            for(var i = 0; i < _comparisons.Length; i++)
+                if (_comparisons[i].X == x && _comparisons[i].Y == y)
```
rprouse: Maybe I'm missing something, but since we are only looking for recursion, wouldn't it be more efficient here to only test for reference equality, `Object.ReferenceEquals(_comparisons[0].X, x)`?

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
jnm2: Here's the reason (with example) of why something like this is needed: https://github.com/nunit/nunit/pull/3406#discussion_r339322348.

It would be nice to avoid array copies, especially if there are ever graphs that go more than three or four levels deep. In order to not hold up the PR, I'd be happy to implement an ImmutableStack based on an singly-linked list instead of using array storage. Unless implementing a minimal ImmutableStack sounds interesting to you, @stevenaw :D

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
jnm2: Yes! It seems like it could be calculated as `comparisons.Length == 0` or something like that.

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
jnm2: The `==` operator checks reference equality if one of the operands is a reference type. It can't be overloaded in this scenario since both operands are `object`. `ReferenceEquals` might be clearer though?

jnm2: (The reference equality of `==` on `object` is also a pitfall. `obj == "foo"` is checking reference equality, which is different than `obj as string == "foo"` which uses the `==(string, string)` overload.)

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
stevenaw: Agreed, this is ideal. The only wrinkle to making this calculated is backwards compatibility. This call is downstream from a public API where this boolean can be passed in as an argument:
https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs#L157

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
stevenaw: Sorry, I must've misunderstood your initial suggestion when you mentioned the immutablestack. By all means go for it :) My free time over the coming free weeks is going to be a little more restricted, so I welcome any help or collaboration you want in this area.
Within this PR, I guess the main requirements would be ability to push, and ability to enumerate all items. Thanks @jnm2 !

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
stevenaw: Thanks for the suggestion! I'll try and get this done as soon as possible.

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
stevenaw: Oh, but that is a good point. I can omit passing it here, since it would always be false deeper down the stack

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
jnm2: No worries, I tried to leave that open-ended. If you like, you can pull or cherry-pick this commit into your PR, or I can do a separate PR after this one. Up to you.

https://github.com/stevenaw/nunit/compare/issue52/cyclical-graph...nunit:examples/cyclical_graph_immutablestack

stevenaw: Thanks! I've done a fast-forward merge of your branch into mine and run some tests. Pushed + just waiting for CI to finish

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
stevenaw: Done. I ended up defaulting to a hard-coded `false` on push. I'm a little concerned that making this fully calculated could break someone

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
stevenaw: Converted to `Object.ReferenceEquals()`

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
```diff
@@ -0,0 +1,59 @@
...
+                false,
+                _comparisons.Push(new Comparison(x, y))
+            );
+        }
+
+        public bool DidCompare(object x, object y)
+        {
+            foreach (var comparison in _comparisons)
+                if (Object.ReferenceEquals(comparison.X, x) &&
+                    Object.ReferenceEquals(comparison.Y, y))
```
jnm2: Keywords for built-in types are recommended over type names. If you prefer, `ReferenceEquals` doesn't need to be qualified with `Object.` or `object.` because all types inherit from object.

src/NUnitFramework/framework/Constraints/Comparers/ComparisonState.cs
jnm2: I agree that we don't want to make any behavioral changes and I guess there's no way to avoid it because the top-level flag would sometimes be false even when the stack is empty. We could emulate it by pushing a `new object()` comparison, but that seems less straightforward.

stevenaw:
Thanks @jnm2 ! You're right, I converted the prop to a field thinking it would help performance. If you're ok to wait to merge until the weekend (pending second review) I should have some time by EOD Friday to address your last two comments

jnm2:
@stevenaw There's absolutely no rush. You can edit `[WIP]` into the PR title so that we don't accidentally merge before you're finished.

stevenaw:
Thanks for the suggestion! I wouldn't really consider this a WIP so much as that I don't like leaving unaddressed feedback. Once something gets merged, there's always the chance things like that won't get fixed later.

PR #3413, opened 2019-11-01
Show names of parameters

### LINKED ISSUE #3389: Show names of parameters #feature
Hi!
Maybe I don’t know something, but to my surprise, in the test window (VS), I could not display the parameter names, in xunit they are displayed. This is solved simply enough to enable it by default. I would like to have such an opportunity at least optionally.

```csharp
            public override string GetText(MethodInfo method, object[] arglist)
            {
                var sb = new StringBuilder();

                if (arglist != null)
                {
                    sb.Append('(');
                    var parameters = method.GetParameters();
                    for (int i = 0; i < arglist.Length; i++)
                    {
//sample code, nothing more.
                        if (i > 0) sb.Append(", ");
                        var paramName = parameters.Length > i ? parameters[i].Name : "?";
                        sb.Append(paramName);
                        sb.Append(": ");
                        sb.Append(GetDisplayString(arglist[i], Math.Max(1, _maxStringLength - paramName.Length - 3)));
                    }

                    sb.Append(')');
                }

                return sb.ToString();
            }
```
The result is pleasing.
What do you think about it?
(Above is a sample code for adding default functionality)

изображение

lobster2012-user:
TestNameGenerator.cs: show names of parameters. Fixes #3389 

(first try)

### DIFF
--- a/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
+++ b/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
@@ -143,6 +143,9 @@ namespace NUnit.Framework.Internal
                     case "{a}":
                         fragments.Add(new ArgListFragment(0));
                         break;
+                    case "{p}":
+                        fragments.Add(new ParamArgListFragment(0));
+                        break;
                     case "{0}":
                     case "{1}":
                     case "{2}":
@@ -158,12 +161,14 @@ namespace NUnit.Framework.Internal
                         break;
                     default:
                         char c = token[1];
-                        if (token.Length >= 5 && token[2] == ':' && (c == 'a' || char.IsDigit(c)))
+                        if (token.Length >= 5 && token[2] == ':' && (c == 'a' || c == 'p' || char.IsDigit(c)))
                         {
                             if (int.TryParse(token.Substring(3, token.Length - 4), out var length) && length > 0)
                             {
                                 if (c == 'a')
                                     fragments.Add(new ArgListFragment(length));
+                                else if (c == 'p')
+                                    fragments.Add(new ParamArgListFragment(length));
                                 else // It's a digit
                                     fragments.Add(new ArgumentFragment(c - '0', length));
                                 break;
@@ -580,6 +585,42 @@ namespace NUnit.Framework.Internal
                 return _index < args.Length
                     ? GetDisplayString(args[_index], _maxStringLength)
                     : string.Empty;
+            }
+        }
+
+        private class ParamArgListFragment : NameFragment
+        {
+            private readonly int _maxStringLength;
+
+            public ParamArgListFragment(int maxStringLength)
+            {
+                _maxStringLength = maxStringLength;
+            }
+
+            public override string GetText(MethodInfo method, object[] args)
+            {
+                var sb = new StringBuilder();
+                args = args ?? new object[0];
+
+                var parameters = method.GetParameters();
+
+                sb.Append('(');
+                var count = Math.Max(args.Length, parameters.Length);
+
+                for (int i = 0; i < count; i++)
+                {
+                    if (i > 0) sb.Append(", ");
+
+                    var paramName = GetDisplayString(parameters.Length > i ? parameters[i].Name : "?", _maxStringLength).Trim('\"');
+                    sb.Append(paramName);
+                    sb.Append(": ");
+                    sb.Append(GetDisplayString(args.Length > i ? args[i] : "?", _maxStringLength));
+                }
+
+                sb.Append(')');
+
+
+                return sb.ToString();
             }
         }
 
--- a/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
+++ b/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
@@ -609,16 +609,26 @@ namespace NUnit.Framework.Internal
 
                 for (int i = 0; i < count; i++)
                 {
+                    if(i < parameters.Length)
+                    {
+                        if (parameters[i].IsOptional)
+                        {
+                            if(i >= args.Length)
+                            {
+                                continue;
+                            }
+                        }
+                    }
+
                     if (i > 0) sb.Append(", ");
 
-                    var paramName = GetDisplayString(parameters.Length > i ? parameters[i].Name : "?", _maxStringLength).Trim('\"');
+                    var paramName = i < parameters.Length ? parameters[i].Name : "?";
                     sb.Append(paramName);
                     sb.Append(": ");
-                    sb.Append(args.Length > i ? GetDisplayString(args[i], _maxStringLength) : "?");
+                    sb.Append(i < args.Length ? GetDisplayString(args[i], _maxStringLength) : "?");
                 }
 
                 sb.Append(')');
-
 
                 return sb.ToString();
             }
--- a/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
+++ b/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
@@ -600,36 +600,28 @@ namespace NUnit.Framework.Internal
             public override string GetText(MethodInfo method, object[] args)
             {
                 var sb = new StringBuilder();
-                args = args ?? new object[0];
 
-                var parameters = method.GetParameters();
-
-                sb.Append('(');
-                var count = Math.Max(args.Length, parameters.Length);
-
-                for (int i = 0; i < count; i++)
+                if (args != null)
                 {
-                    if(i < parameters.Length)
+                    sb.Append('(');
+
+                    var parameters = method.GetParameters();
+                    
+                    for (int i = 0; i < args.Length; i++)
                     {
-                        if (parameters[i].IsOptional)
+                        if (i > 0) sb.Append(", ");
+
+                        if (i < parameters.Length)
                         {
-                            if(i >= args.Length)
-                            {
-                                continue;
-                            }
+                            sb.Append(parameters[i].Name);
+                            sb.Append(": ");
                         }
+
+                        sb.Append(GetDisplayString(args[i], _maxStringLength));
                     }
 
-                    if (i > 0) sb.Append(", ");
-
-                    var paramName = i < parameters.Length ? parameters[i].Name : "?";
-                    sb.Append(paramName);
-                    sb.Append(": ");
-                    sb.Append(i < args.Length ? GetDisplayString(args[i], _maxStringLength) : "?");
+                    sb.Append(')');
                 }
-
-                sb.Append(')');
-
                 return sb.ToString();
             }
         }
--- a/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
+++ b/src/NUnitFramework/framework/Internal/TestNameGenerator.cs
@@ -614,7 +614,7 @@ namespace NUnit.Framework.Internal
                     var paramName = GetDisplayString(parameters.Length > i ? parameters[i].Name : "?", _maxStringLength).Trim('\"');
                     sb.Append(paramName);
                     sb.Append(": ");
-                    sb.Append(GetDisplayString(args.Length > i ? args[i] : "?", _maxStringLength));
+                    sb.Append(args.Length > i ? GetDisplayString(args[i], _maxStringLength) : "?");
                 }
 
                 sb.Append(')');
--- a/src/NUnitFramework/tests/Internal/TestNameGeneratorTests.cs
+++ b/src/NUnitFramework/tests/Internal/TestNameGeneratorTests.cs
@@ -89,9 +89,9 @@ namespace NUnit.Framework.Internal
             return new TestNameGenerator(pattern).GetDisplayName(_simpleTest, args);
         }
 
-        [TestCase("{m}{p}", new object[] { 1 }, ExpectedResult = "TestMethod(a: 1, b: ?)")]
-        [TestCase("{m}{p}", new object[] { 1, 2 }, ExpectedResult = "TestMethod(a: 1, b: 2)")]
-        [TestCase("{m}{p}", new object[] { 1, 2, 3 }, ExpectedResult = "TestMethod(a: 1, b: 2, ?: 3)")]
+        [TestCase("{m}{p}", new object[] { 1 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: ?)")]
+        [TestCase("{m}{p}", new object[] { 1, 2 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: 2)")]
+        [TestCase("{m}{p}", new object[] { 1, 2, 3 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: 2, ?: 3)")]
         public string ParameterizedTestsWithArgs(string pattern, object[] args)
         {
             return new TestNameGenerator(pattern).GetDisplayName(_simpleTestWithArgs, args);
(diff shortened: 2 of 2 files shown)


lobster2012-user:
I’ll update it soon.

lobster2012-user:
Done for review, but I do not see some tests.

mikkelbu:
@lobster2012-user Please note that the failing Linux tests in "NUnit Framework CI" are not due to your changes, I just provoked the same errors on master by triggering a build. It seems that Azure pipelines have updated their Linux image.

src/NUnitFramework/tests/Internal/TestNameGeneratorTests.cs
```diff
@@ -86,6 +89,14 @@ public string ParameterizedTests(string pattern, object[] args)
             return new TestNameGenerator(pattern).GetDisplayName(_simpleTest, args);
         }
 
+        [TestCase("{m}{p}", new object[] { 1 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: ?)")]
```
jnm2: If the `b` parameter was optional, I think I'd expect `TestMethodWithArgs(a: 1)`.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
```diff
@@ -583,6 +588,42 @@ public override string GetText(MethodInfo method, object[] args)
...
+                var parameters = method.GetParameters();
+
+                sb.Append('(');
+                var count = Math.Max(args.Length, parameters.Length);
+
+                for (int i = 0; i < count; i++)
+                {
+                    if (i > 0) sb.Append(", ");
+
+                    var paramName = GetDisplayString(parameters.Length > i ? parameters[i].Name : "?", _maxStringLength).Trim('\"');
```
jnm2: GetDisplayString is intended for values and I don't think it's a good fit for parameter names.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
```diff
@@ -583,6 +588,42 @@ public override string GetText(MethodInfo method, object[] args)
...
+                    sb.Append(paramName);
+                    sb.Append(": ");
+                    sb.Append(args.Length > i ? GetDisplayString(args[i], _maxStringLength) : "?");
+                }
+
+                sb.Append(')');
+
+
```
jnm2: Was the double-spacing intentional?

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
lobster2012-user: I would use a raw value with no length limit for the parameter name

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
lobster2012-user: No, I’ll fix it.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
jnm2: Ah, right. I'm okay with not applying the length limit, but if we do, we should extract the length-limiting behavior out of GetDisplayString and call that directly from here rather than going through GetDisplayString.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
```diff
@@ -583,6 +588,52 @@ public override string GetText(MethodInfo method, object[] args)
...
+                for (int i = 0; i < count; i++)
+                {
+                    if(i < parameters.Length)
+                    {
+                        if (parameters[i].IsOptional)
+                        {
+                            if(i >= args.Length)
+                            {
+                                continue;
```
jnm2: Not `break`?
Also, the if statements can be joined with `&&`.

lobster2012-user: I was thinking over a 'break'.
If possible, specify a non-optional parameter between optional, this is not suitable.
At the moment there is no such possibility, perhaps it never will be.
I can change to break.

lobster2012-user: >the if statements can be joined with &&.
I prefer brackets, I can also change.

jnm2: That's a good thing to consider, but I think the `break;` also pairs well with the `i > args.Length`.

I'd be happy if we disregarded `IsOptional`, even, for this purpose, unless you think that could be a source of confusion.

lobster2012-user: If an optional parameter is not specified, then its default value can be displayed.
For tests, I never used the optional parameters, so I do not have a clear opinion.

lobster2012-user: Perhaps I did not quite understand your point.

jnm2: By disregarding IsOptional, I meant, let's pretend for the sake of formatting that all the parameters are optional if `args` is too short.

CharliePoole:
What happens if the user specifies test case arts incorrectly? I.e. too many, too few, wrong type. It would be useful to see what is displayed using {p} vs {a}. Runners need to display test names for invalid tests.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
CharliePoole: Test names, IMO, should reflect what the user specified foe arguments. Unspecified arts shouldn't appear. Consequently, as @jnm2 points out, it doesn't matter if they are optional.

lobster2012-user:
> 
> 
> What happens if the user specifies test case arts incorrectly? I.e. too many, too few, wrong type. It would be useful to see what is displayed using {p} vs {a}. Runners need to display test names for invalid tests.

https://github.com/nunit/nunit/issues/3389#issuecomment-539128290
```csharp
[TestCase("{m}{p}", new object[] { 1 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: ?)")]
[TestCase("{m}{p}", new object[] { 1, 2 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: 2)")]
[TestCase("{m}{p}", new object[] { 1, 2, 3 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: 2, c: 3)")]
[TestCase("{m}{p}", new object[] { 1, 2, 3, 4 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: 2, c: 3, ?: 4)")]

private void TestMethodWithArgs(Int32 a, Int32 b, Int32 c = 0) { }
```
I won’t say anything about the wrong types.

CharliePoole:
Seems like an unnecessary change in behavior compared to {a}.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
lobster2012-user: https://user-images.githubusercontent.com/9452368/66336959-d4484780-e946-11e9-9e65-b17b03be715a.png

In the current implementation, all arguments will be displayed as `?` Unfortunately.

lobster2012-user:
> 
> 
> Seems like an unnecessary change in behavior compared to {a}.
```csharp
[TestCase("{m}{a}", new object[] { 1, 2 }, ExpectedResult = "TestMethod(1,2)")]
[TestCase("{m}{p}", new object[] { 1, 2 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: 2)")]
```

lobster2012-user:
I did not quite understand the question. 
We are talking about the futility of PR or display '?'

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
CharliePoole: It's a non-obvious problem, but I'd hazard a guess that we've spent more effort on invalid test display over the years than valid ones! In early days, invalid tests just disappeared, which gives you a false positive. In this case, the ? Gives a warning so it's not terrible, but it is different from the default display. @jnm2 what do you think?

lobster2012-user: @CharliePoole 
@jnm2 
In the latest version.
1. If the optional parameter is not set, it is not displayed.
2. If the required parameter is not specified, "?" is displayed instead of value.
3. If there are too many arguments, then the names of non-existent parameters are replaced with "?"

Which of these do you not like and what to change?

CharliePoole:
> > Seems like an unnecessary change in behavior compared to {a}.
> 
> ```cs
> [TestCase("{m}{a}", new object[] { 1, 2 }, ExpectedResult = "TestMethod(1,2)")]
> [TestCase("{m}{p}", new object[] { 1, 2 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: 2)")]
> ```

This is not a change in behavior - except for the intended change of introducing the parameter name. The change occurs in error cases only, i.e. if the number of args doesn't match correctly.

So (working out of my head, here so I may have missed something) if the method takes three args, then we have

> ```cs
> [TestCase("{m}{a}", new object[] { 1, 2 }, ExpectedResult = "TestMethod(1,2)")]
> [TestCase("{m}{p}", new object[] { 1, 2 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: 2, c: ?)")]
> ```

In both cases, I believe NUnit will display the same error message for the invalid test case. I can see some advantage to the `?` (it highlights the fact that there's an error in the params) but it is different from what people have been seeing up to now.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
CharliePoole: In the case of {a}

1. If an optional parameter is not set, it is not displayed.

2. If a required parameter is not set, it is not displayed.

3. If there are too many arguments, they are displayed.

If we want {p} error handling to conform closely to that for {a}, then I'd change 2 and 3. In the case of 3, I'd fall back on merely displaying the value.

As usual, I'm afraid it takes seeing an example to enable me to zero in on what I'd like to see.

lobster2012-user:
If I understand correctly, in order not to change the behavior, you suggest replacing max with args.Length.  I understand the reason for your point of view, I will not insist on my own, although I like my option more.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
lobster2012-user: What about
```csharp
[TestCase("{m}{p}", new [] { 1 }, Ex.Result = "Test.Args(a: 1)")]
[TestCase("{m}{p}", new [] { 1, 2 }, Ex.Result = "Test.Args(a: 1, b: 2)")]
[TestCase("{m}{p}", new [] { 1, 2, 3 }, Ex.Result = "Test.Args(a: 1, b: 2, c: 3)")]
[TestCase("{m}{p}", new [] { 1, 2, 3, 4 }, Ex.Result = "Test.Args(a: 1, b: 2, c: 3, ?: 4)")]

private void TestMethodWithArgs(Int32 a, Int32 b, Int32 c = 0) { }
```

lobster2012-user: or

[TestCase("{m}{p}", new object[] { 1 }, ExpectedResult = "TestMethodWithArgs(a: 1)")]
        [TestCase("{m}{p}", new object[] { 1, 2 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: 2)")]
        [TestCase("{m}{p}", new object[] { 1, 2, 3 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: 2, c: 3)")]
        [TestCase("{m}{p}", new object[] { 1, 2, 3, 4 }, ExpectedResult = "TestMethodWithArgs(a: 1, b: 2, c: 3, 4)")]

CharliePoole:
I think the issue of compatibility has to be considered - i.e. compatibility from the point of view of what users are used to seeing. However, I don't see it as a clear right/wrong choice, so I'd like to hear what @jnm2 and other team members think before you spend time making any changes.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
CharliePoole: Personally, I prefer the version without the question mark.

The issue I see is that the naming of a test has never depended on the number of args that the method takes. With {a} it is solely dependent on what args the user supplies, whether they are correct, too few or too many. With {p} we obviously have to look at the names of the method parameters but I'd rather not see us change anything more than is needed. I also find it a bit surprising to see `?` used for both the name and the value.

@jnm2 Do you have any opinion one way or the other here?

jnm2:
I was asking for the same thing as @CharliePoole; I think sticking to `args.Length` without Math.Max is more consistent and works better for us.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
jnm2: @CharliePoole Yes, my opinion is the same as yours.

src/NUnitFramework/framework/Internal/TestNameGenerator.cs
```diff
@@ -583,6 +588,44 @@ public override string GetText(MethodInfo method, object[] args)
...
+                _maxStringLength = maxStringLength;
+            }
+
+            public override string GetText(MethodInfo method, object[] args)
+            {
+                var sb = new StringBuilder();
+
+                if (args != null)
+                {
+                    sb.Append('(');
```
jnm2: Why does it make sense to skip adding `()` when args is null but not when args is empty?

lobster2012-user: See ArgListFragment.
I changed the logic, which coincided with {a}. At the beginning, I always added brackets.

CharliePoole: There are two cases in which args will be null...

1. This is a simple test, specified by the `TestAttribute`, so no arguments are supplied at all.

2. A perverse use of `TestCaseSourceAttribute` returns null rather than an empty list.

The first case would seem to be the primary usage. If parentheses were added, the majority of test names in most applications would probably change, which would be bad.

Leaving aside the above "perverse" case, bear in mind that multiple sources, like `[TestCase]` and `[TestCaseSource]` maybe specified. Some of them might be empty. It's desirable that the names of the tests should be as similar as possible, given differences in arguments.

jnm2: Good, thanks!

lobster2012-user:
@jnm2 @CharliePoole Thank you too.
The only thing that remains a mystery to me is how people learn about the new template.  By default, it is not enabled.

jnm2:
I'd like to make this the default template in v4. I'll open an issue.

CharliePoole:
@lobster2012-user Most people don't know about templates at all until they ask a question and get pointed to the docs. We will, of course, need to update the docs to reflect this new template.

Feature discoverability has been a big issue in software for as long as I've been programming. Modern-day apps have moved to pushing all the new features at you rather aggressively. I hope we don't go that way.

Looking at the docs, can we make this feature more discoverable? The template description is hidden in a secondary page currently linked from the `TestCaseData` page, IIRC.

jnm2:
(I also updated https://github.com/nunit/docs/wiki/Template-Based-Test-Naming just now. Thanks for bringing it up!)

lobster2012-user:
I think that a short description of features with links can be included in the project template(  rich xml comments with urls) or you can include  readme file(txt, md, html) in the package.  Is there a need for this, perhaps.

PR #3416, opened 2019-11-03
Set culture on tests where comparing case of US English strings

ChrisMaddock:
Fixes #3414 (Pending Azure CI run on PR!).

An OS upgrade on Azure changed the default culture our tests ran in from en-US to en-US-POSIX. Our tests shouldn't be sensitive to this sort of change. 🙂

Issue #3414, opened 2019-11-02
Azure pipelines are failing on Linux


mikkelbu:
Between Friday 1st and Saturday 2nd Azure pipelines have started failing on Linux. Examining the logs the it seems that we have started using a new linux kernel in Ubuntu (from 4.15... to 5.0...). I have tried to verify this on https://github.com/microsoft/azure-pipelines-image-generation, but I cannot find the details.
EDIT: Apparently, we are now using Ubuntu 18.04 and not 16.04 - https://github.com/microsoft/azure-pipelines-image-generation/issues/506#issuecomment-548918601.

Before - https://nunit.visualstudio.com/NUnit/_build/results?buildId=2244
```
Runtime Environment
   OS Version: Linux 4.15.0.1059 
  CLR Version: 4.0.30319.42000
```
After - https://nunit.visualstudio.com/NUnit/_build/results?buildId=2253
```
Runtime Environment
   OS Version: Linux 5.0.0.1020 
  CLR Version: 4.0.30319.42000
```

Strangely this makes some of our test fail (all seem related to casing)
```
Errors, Failures and Warnings

1) Failed : NUnit.Framework.Constraints.CollectionEquivalentConstraintTests.EquivalentHonorsUsing
  Expected: True
  But was:  False

2) Failed : NUnit.Framework.Constraints.ContainsConstraintTests.HonorsIgnoreCaseForString
  Expected: True
  But was:  False

3) Failed : NUnit.Framework.Constraints.DictionaryContainsKeyConstraintTests.UsingIsHonored
  Expected: dictionary containing key "HELLO"
  But was:  < ["Hello", "World"], ["Hola", "Mundo"] >

4) Failed : NUnit.Framework.Constraints.DictionaryContainsValueConstraintTests.UsingIsHonored
  Expected: dictionary containing value "UNIVERSE"
  But was:  < ["Hello", "World"], ["Hi", "Universe"], ["Hola", "Mundo"] >

5) Failed : NUnit.Framework.Constraints.EqualConstraintTests+UsingModifier.UsesProvidedLambda_StringArgs
  String lengths are both 5. Strings differ at index 0.
  Expected: "HELLO"
  But was:  "hello"
  -----------^

6) Failed : NUnit.Framework.Constraints.EqualConstraintTests+UsingModifier.UsesProvidedPredicateForItemComparison
  Expected and actual are both <System.String[3]>
  Values differ at index [0]
  String lengths are both 4. Strings differ at index 0.
[...]

PR #3417, opened 2019-11-03
Print cultures using display name rather than ToString()

ChrisMaddock:
When working on #3414 - I got a little confused as the test results were displaying `culture=""`. Turns out the empty string is used when you ToString the Invariant Culture.

This doesn't seem like the most friendly way to record culture information - even though the four char codes are nice to work with. I've switched to `DisplayName`, which prints a friendly name instead (e.g. `English (United Kingdom)`. 

What do people think? I had to change a test for this, but I don't think the behaviour of the test was particularly important. I didn't change the NUnit 2 result writer, which has the same issue.

CI should be fixed after #3416 is merged. 👍

### DIFF
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -365,8 +365,8 @@ namespace NUnit.Framework.Api
             env.AddAttribute("machine-name", Environment.MachineName);
             env.AddAttribute("user", Environment.UserName);
             env.AddAttribute("user-domain", Environment.UserDomainName);
-            env.AddAttribute("culture", CultureInfo.CurrentCulture.ToString());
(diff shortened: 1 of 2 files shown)


jnm2:
Won't this break everyone that parses that field?

ChrisMaddock:
This is what I wasn't sure about. I can't think of a workflow where you would parse the final TestResult.xml file, and then create a culture based off the results. The only use I could think of is runners that display it as a string for info purposes, in which case this should matter (and be an improvement, for the invariant case).

But I'm not sure if there's something I might not be considering. Were you thinking of anything in particular? 🤔

jnm2:
No, but I have post-processed the XML before and was picturing a script that filtered the tests based on culture equalling known IDs, or something like that.

ChrisMaddock:
That's fair. Let's leave it then. One for v4, maybe! 😄

PR #3419, opened 2019-11-05
Update nuspec description for framework

### LINKED ISSUE #3411: Update nuspec file to mention support for NET Standard 2.0+
I think the nuspec file for nunit.dll needs to have the description updated to reflect current supported platforms. It mentions `.NET Standard 1.4+` but I think the actual support is `.NET Standard 2.0+`.
https://github.com/nunit/nunit/blob/master/nuget/framework/nunit.nuspec#L21

stevenaw:
Fixes #3411 

This PR updates the description to indicate the framework targets .NET Standard 2.0+
It also drops the mention of .NET Core (not sure if this is desired for searchability reasons or not)

### DIFF
--- a/nuget/framework/nunit.nuspec
+++ b/nuget/framework/nunit.nuspec
@@ -18,8 +18,7 @@ This package includes the NUnit 3 framework assembly, which is referenced by you
 
 Supported platforms:
 - .NET Framework 3.5+
-- .NET Standard 1.4+
-- .NET Core</description>
+- .NET Standard 2.0+</description>
     <releaseNotes>This package includes the NUnit 3 framework assembly, which is referenced by your tests. You will need to install version 3 of the nunit3-console program or a third-party runner that supports NUnit 3 in order to execute tests. Runners intended for use with NUnit 2.x will not run NUnit 3 tests correctly.</releaseNotes>
     <language>en-US</language>
     <tags>nunit test testing tdd framework fluent assert theory plugin addin</tags>


nuget/framework/nunit.nuspec
```diff
@@ -18,8 +18,7 @@ This package includes the NUnit 3 framework assembly, which is referenced by you
 
 Supported platforms:
 - .NET Framework 3.5+
-- .NET Standard 1.4+
-- .NET Core</description>
```
stevenaw: As noted in the description, this isn't really a supported target for the framework. Though I'm not sure if this is desired to leave in for clarity (or searchability ?)

jnm2: Not listing .NET Core is more accurate. Since there's a Dependencies section on the nuget.org pack, this list is duplicate information which usually only makes a difference in a negative way when it gets out of date. @nunit/framework-team  what are your thoughts?

CharliePoole: If it's duplicate info that has to be kept up to date, I'd favor removal.

jnm2: Also, .NET Core 3.1 is the last version before they drop the word "Core" and go to ".NET 5."

stevenaw: Oh, very good points. So as far as current state goes, does this make sense as-is? Or should we remove this part of the description entirely?

mikkelbu: I think it looks fine. @jnm2 do you agree?

jnm2: Yes.

Issue #2574, opened 2017-11-24
Instance-per-test-case feature


jnm2:
Everyone's mental model appears to be that there is an instance created per test case when parallelizing within a fixture (https://github.com/nunit/nunit/issues/2105, https://github.com/nunit/nunit/issues/2252, https://github.com/nunit/nunit/issues/2573) even though this has never been the case. I tend to solve this problem by always using static classes and defining a disposable nested fixture class which provides a richer experience in some ways ([example](https://github.com/nunit/nunit/issues/2105#issuecomment-325347460)), but that might not be what we want for everyone. The slight downside here is in terminology, that the static class is what NUnit considers to be the fixture but the real fixture is the nested class.

It's not an option to make instance-per-test-case the default because that breaks non-parallel fixtures which rely on one test being able to access the state from another test. People using the `[Order]` attribute are probably the only people it would affect, but it would totally break any order-dependent test I can think of.

I like Charlie's proposal:

>  * An assembly level attribute that allows setting the way test fixtures are created to be single instance or instance per fixture.
>  * Single Instance would work just like now
>  * Instance per fixture would create a new fixture for each test
>  * EITHER Perform one time setup and teardown activities for each test (i.e. no longer "one time") OR cause OneTimeSetUp and OneTimeTearDown to be flagged as errors.
> 
> Possible extensions:
>  * Make this class level as well, allowing individual fixtures to have different life cycles.
>  * Have a third assembly-level setting that automatically runs separate instances for any fixture containing one or more tests that are run in parallel. (Tricky to always know in advance)
> 
> I'd say do the basics first, however.

Right now it's guaranteed to be a bug if there is mutable instance state and a fixture runs in parallel with itself, so we could have an Auto option that switches to instance-per-test-case if test cases from the fixture are configured to run in parallel and **avoid any breaking changes**. I think Auto has the potential to be confusing, but it might also just be the most sensible default so that everyone won't be adding `[assembly: InstancePerTestCase]` as a matter of routine.

I'm in favor of guardrails. OneTimeSetUp and OneTimeTearDown should be flagged as errors if the fixture does an instance per test case. Perhaps unless they are static?

CharliePoole:
Currently, you __can__ share state between parallel tests provided (1) you set the state in the constructor or one-time setup and (2) don't change it in any test. You can even get by in some cases if the state is set to a constant in the test or setup, provided you never set it to any other value. The latter is icky, even though I think we have some NUnit tests that do it.

[...]

PR #3427, opened 2019-11-16
introduce FixtureLifeCycle

avilv:
see issue #2574

### DIFF
--- a/src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
@@ -50,7 +50,7 @@ namespace NUnit.Framework
         public LifeCycle LifeCycle { get; }
 
         /// <summary>
-        /// Overridden to set TestFixture LifeCycle
+        /// Overridden to set a TestFixture's <see cref="LifeCycle"/>.
         /// </summary>
         public override void ApplyToTest(Test test)
         {
@@ -62,6 +62,5 @@ namespace NUnit.Framework
 
             base.ApplyToTest(test);
         }
-
     }
 }
--- a/src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
@@ -1,0 +1,75 @@
+// ***********************************************************************
+// Copyright (c) 2019 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+
+namespace NUnit.Framework
+{
+    /// <summary>
+    /// Specify the lifecycle of a Fixture
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
+    public sealed class FixtureLifeCycleAttribute : PropertyAttribute
+    {
+        /// <summary>
+        /// Construct a FixtureLifeCycleAttribute using default LifeCycle.SingleInstance.
+        /// </summary>
+        public FixtureLifeCycleAttribute() : this(LifeCycle.SingleInstance) { }
+
+
+        /// <summary>
+        /// Construct a FixtureLifeCycleAttribute with a specified lifeCycle.
+        /// </summary>
+        /// <param name="lifeCycle">The Lifecycle associated with this attribute.</param>
+        public FixtureLifeCycleAttribute(LifeCycle lifeCycle) : base()
+        {
+            LifeCycle = lifeCycle;
+        }
+
+        /// <summary>
+        /// Defines the lifecycle for this test fixture or assembly.
+        /// </summary>
+        public LifeCycle LifeCycle { get; }
+
+        /// <summary>
+        /// Overridden to set TestFixture LifeCycle
+        /// </summary>
+        /// <param name="test"></param>
+        public override void ApplyToTest(Test test)
+        {
+            if (test is TestFixture)
+            {
+                ((TestFixture)test).LifeCycle = LifeCycle;
+            }
+
+            base.ApplyToTest(test);
+        }
+
+    }
+
+}
--- a/src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
@@ -31,39 +31,33 @@ using NUnit.Framework.Internal;
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Specify the lifecycle of a Fixture
+    /// Specify the life cycle of a Fixture
     /// </summary>
     [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
     public sealed class FixtureLifeCycleAttribute : PropertyAttribute
     {
         /// <summary>
-        /// Construct a FixtureLifeCycleAttribute using default LifeCycle.SingleInstance.
+        /// Construct a FixtureLifeCycleAttribute with a specified <see cref="LifeCycle"/>.
         /// </summary>
-        public FixtureLifeCycleAttribute() : this(LifeCycle.SingleInstance) { }
-
-        /// <summary>
-        /// Construct a FixtureLifeCycleAttribute with a specified lifeCycle.
-        /// </summary>
-        /// <param name="lifeCycle">The Lifecycle associated with this attribute.</param>
-        public FixtureLifeCycleAttribute(LifeCycle lifeCycle) : base()
+        public FixtureLifeCycleAttribute(LifeCycle lifeCycle)
         {
             LifeCycle = lifeCycle;
         }
 
         /// <summary>
-        /// Defines the lifecycle for this test fixture or assembly.
+        /// Defines the life cycle for this test fixture or assembly.
         /// </summary>
         public LifeCycle LifeCycle { get; }
 
         /// <summary>
         /// Overridden to set TestFixture LifeCycle
         /// </summary>
-        /// <param name="test"></param>
         public override void ApplyToTest(Test test)
         {
-            if (test is TestFixture)
+            var testFixture = test as TestFixture;
+            if (testFixture != null)
             {
-                ((TestFixture)test).LifeCycle = LifeCycle;
+                testFixture.LifeCycle = LifeCycle;
             }
 
             base.ApplyToTest(test);
--- a/src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
@@ -41,7 +41,6 @@ namespace NUnit.Framework
         /// </summary>
         public FixtureLifeCycleAttribute() : this(LifeCycle.SingleInstance) { }
 
-
         /// <summary>
         /// Construct a FixtureLifeCycleAttribute with a specified lifeCycle.
         /// </summary>
@@ -71,5 +70,4 @@ namespace NUnit.Framework
         }
 
     }
-
 }
--- a/src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
@@ -34,7 +34,7 @@ namespace NUnit.Framework
     /// Specify the life cycle of a Fixture
     /// </summary>
     [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
-    public sealed class FixtureLifeCycleAttribute : PropertyAttribute
+    public sealed class FixtureLifeCycleAttribute : NUnitAttribute, IApplyToTest
     {
         /// <summary>
         /// Construct a FixtureLifeCycleAttribute with a specified <see cref="LifeCycle"/>.
@@ -52,15 +52,13 @@ namespace NUnit.Framework
         /// <summary>
         /// Overridden to set a TestFixture's <see cref="LifeCycle"/>.
         /// </summary>
-        public override void ApplyToTest(Test test)
+        public void ApplyToTest(Test test)
         {
             var testFixture = test as TestFixture;
             if (testFixture != null)
             {
                 testFixture.LifeCycle = LifeCycle;
             }
-
-            base.ApplyToTest(test);
         }
     }
 }
--- a/src/NUnitFramework/framework/Attributes/LifeCycle.cs
+++ b/src/NUnitFramework/framework/Attributes/LifeCycle.cs
@@ -24,7 +24,7 @@
 namespace NUnit.Framework
 {
     /// <summary>
-    /// Specifies the lifecycle for a fixture.
+    /// Specifies the life cycle for a test fixture.
     /// </summary>
     public enum LifeCycle
     {
--- a/src/NUnitFramework/framework/Attributes/LifeCycle.cs
+++ b/src/NUnitFramework/framework/Attributes/LifeCycle.cs
@@ -38,5 +38,4 @@ namespace NUnit.Framework
         /// </summary>
         InstancePerTestCase
     }
-
 }
--- a/src/NUnitFramework/framework/Attributes/LifeCycle.cs
+++ b/src/NUnitFramework/framework/Attributes/LifeCycle.cs
@@ -1,0 +1,42 @@
+// ***********************************************************************
+// Copyright (c) 2019 Charlie Poole, Rob Prouse
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+// 
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+namespace NUnit.Framework
+{
+    /// <summary>
+    /// Specifies the lifecycle for a fixture.
+    /// </summary>
+    public enum LifeCycle
+    {
+        /// <summary>
+        /// A single instance is created and shared for all test cases.
+        /// </summary>
+        SingleInstance,
+
+        /// <summary>
+        /// A new instance is created for each test case.
+        /// </summary>
+        InstancePerTestCase
+    }
+
+}
--- a/src/NUnitFramework/framework/Internal/Commands/ConstructFixtureCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/ConstructFixtureCommand.cs
@@ -49,7 +49,6 @@ namespace NUnit.Framework.Internal.Commands
                 currentTest = currentTest.Parent;
             }
 
-
             Guard.ArgumentValid(testSuite is TestSuite, "ConstructFixtureCommand must reference a TestSuite", nameof(innerCommand));
 
             BeforeTest = (context) =>
--- a/src/NUnitFramework/framework/Internal/Commands/ConstructFixtureCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/ConstructFixtureCommand.cs
@@ -21,6 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+using System;
 using NUnit.Framework.Interfaces;
 
 namespace NUnit.Framework.Internal.Commands
@@ -37,7 +38,19 @@ namespace NUnit.Framework.Internal.Commands
         public ConstructFixtureCommand(TestCommand innerCommand)
             : base(innerCommand)
         {
-            Guard.ArgumentValid(Test is TestSuite, "ConstructFixtureCommand must reference a TestSuite", nameof(innerCommand));
+            TestSuite testSuite = null;
+            TestFixture testFixture = null;
+
+            ITest currentTest = Test;
+            while (currentTest != null && (testSuite == null || testFixture == null))
+            {
+                testSuite = testSuite ?? currentTest as TestSuite;
+                testFixture = testFixture ?? currentTest as TestFixture;
+                currentTest = currentTest.Parent;
+            }
+
+
+            Guard.ArgumentValid(testSuite is TestSuite, "ConstructFixtureCommand must reference a TestSuite", nameof(innerCommand));
 
             BeforeTest = (context) =>
             {
@@ -45,12 +58,20 @@ namespace NUnit.Framework.Internal.Commands
 
                 if (typeInfo != null)
                 {
-                    // Use preconstructed fixture if available, otherwise construct it
                     if (!typeInfo.IsStaticClass)
                     {
-                        context.TestObject = Test.Fixture ?? typeInfo.Construct(((TestSuite)Test).Arguments);
-                        if (Test.Fixture == null)
+                        if (testFixture != null && testFixture.LifeCycle == LifeCycle.InstancePerTestCase)
+                        {
+                            context.TestObject = typeInfo.Construct(testSuite.Arguments);
                             Test.Fixture = context.TestObject;
+                        }
+                        else
+                        {
+                            // Use preconstructed fixture if available, otherwise construct it
+                            context.TestObject = Test.Fixture ?? typeInfo.Construct(testSuite.Arguments);
+                            if (Test.Fixture == null)
+                                Test.Fixture = context.TestObject;
+                        }
                     }
                 }
             };
--- a/src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
@@ -49,7 +49,7 @@ namespace NUnit.Framework.Internal.Commands
                 currentTest = currentTest.Parent;
             }
 
-            Guard.ArgumentValid(testSuite is TestSuite, "ConstructFixturePerTestCaseCommand must reference a TestSuite", nameof(innerCommand));
+            Guard.ArgumentValid(testSuite != null, "ConstructFixturePerTestCaseCommand must reference a TestSuite", nameof(innerCommand));
 
             BeforeTest = (context) =>
             {
--- a/src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
@@ -55,7 +55,7 @@ namespace NUnit.Framework.Internal.Commands
             {
                 ITypeInfo typeInfo = Test.TypeInfo;
 
-                if (context.TestObject == null && typeInfo != null && !typeInfo.IsStaticClass)
+                if (typeInfo != null && !typeInfo.IsStaticClass)
                 {
                     context.TestObject = typeInfo.Construct(testSuite.Arguments);
                     Test.Fixture = context.TestObject;
--- a/src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2017 Charlie Poole, Rob Prouse
+// Copyright (c) 2019 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
--- a/src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
@@ -21,6 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+using System;
 using NUnit.Framework.Interfaces;
 
 namespace NUnit.Framework.Internal.Commands
@@ -28,30 +29,36 @@ namespace NUnit.Framework.Internal.Commands
     /// <summary>
     /// ConstructFixtureCommand constructs the user test object if necessary.
     /// </summary>
-    public class ConstructFixtureCommand : BeforeTestCommand
+    public class ConstructFixturePerTestCaseCommand : BeforeTestCommand
     {
         /// <summary>
-        /// Constructs a OneTimeSetUpCommand for a suite
+        /// Constructs a fixture per test case
         /// </summary>
         /// <param name="innerCommand">The inner command to which the command applies</param>
-        public ConstructFixtureCommand(TestCommand innerCommand)
+        public ConstructFixturePerTestCaseCommand(TestCommand innerCommand)
             : base(innerCommand)
         {
-            Guard.ArgumentValid(Test is TestSuite, "ConstructFixtureCommand must reference a TestSuite", nameof(innerCommand));
+            TestSuite testSuite = null;
+            TestFixture testFixture = null;
+
+            ITest currentTest = Test;
+            while (currentTest != null && (testSuite == null || testFixture == null))
+            {
+                testSuite = testSuite ?? currentTest as TestSuite;
+                testFixture = testFixture ?? currentTest as TestFixture;
+                currentTest = currentTest.Parent;
+            }
+
+            Guard.ArgumentValid(testSuite is TestSuite, "ConstructFixturePerTestCaseCommand must reference a TestSuite", nameof(innerCommand));
 
             BeforeTest = (context) =>
             {
                 ITypeInfo typeInfo = Test.TypeInfo;
 
-                if (typeInfo != null)
+                if (context.TestObject == null && typeInfo != null && !typeInfo.IsStaticClass)
                 {
-                    // Use preconstructed fixture if available, otherwise construct it
-                    if (!typeInfo.IsStaticClass)
-                    {
-                        context.TestObject = Test.Fixture ?? typeInfo.Construct(((TestSuite)Test).Arguments);
-                        if (Test.Fixture == null)
-                            Test.Fixture = context.TestObject;
-                    }
+                    context.TestObject = typeInfo.Construct(testSuite.Arguments);
+                    Test.Fixture = context.TestObject;
                 }
             };
         }
--- a/src/NUnitFramework/framework/Internal/Commands/DisposeFixtureCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/DisposeFixtureCommand.cs
@@ -1,4 +1,4 @@
-// ***********************************************************************
+﻿// ***********************************************************************
 // Copyright (c) 2017 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -41,16 +41,15 @@ namespace NUnit.Framework.Internal.Commands
         public DisposeFixtureCommand(TestCommand innerCommand)
             : base(innerCommand)
         {
+            Guard.OperationValid(Test is IDisposableFixture, "DisposeFixtureCommand does not apply to a " + Test.GetType().Name);
+
             AfterTest = (context) =>
             {
                 try
                 {
                     IDisposable disposable = context.TestObject as IDisposable;
                     if (disposable != null)
-                    {
                         disposable.Dispose();
-                        context.TestObject = null;
-                    }
                 }
                 catch (Exception ex)
                 {
--- a/src/NUnitFramework/framework/Internal/Commands/DisposeFixtureCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/DisposeFixtureCommand.cs
@@ -41,7 +41,9 @@ namespace NUnit.Framework.Internal.Commands
         public DisposeFixtureCommand(TestCommand innerCommand)
             : base(innerCommand)
         {
-            Guard.OperationValid(Test is IDisposableFixture, "DisposeFixtureCommand does not apply to a " + Test.GetType().Name);
+            Guard.OperationValid(
+                Test is IDisposableFixture || Test?.Parent is IDisposableFixture, 
+                $"DisposeFixtureCommand does not apply neither to {Test.GetType().Name}, nor to {Test.Parent?.GetType().Name ?? "it's parent (null)"}");
 
             AfterTest = (context) =>
             {
--- a/src/NUnitFramework/framework/Internal/Commands/DisposeFixtureCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/DisposeFixtureCommand.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2017 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
@@ -41,7 +41,6 @@ namespace NUnit.Framework.Internal.Commands
         public DisposeFixtureCommand(TestCommand innerCommand)
             : base(innerCommand)
         {
-            Guard.OperationValid(Test is IDisposableFixture, "DisposeFixtureCommand does not apply to a " + Test.GetType().Name);
 
             AfterTest = (context) =>
             {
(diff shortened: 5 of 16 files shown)


src/NUnitFramework/testdata/LifeCycleFixture.cs
```diff
@@ -0,0 +1,100 @@
...
+
+namespace NUnit.TestData.LifeCycleTests
+{
+
+
+    [TestFixture]
+    public class LifeCycleFixtureTestCountIsAlwaysOne
+    {
+        public int Count { get; set; }
+        public LifeCycleFixtureTestCountIsAlwaysOne()
```
CharliePoole: We don't normally create explicit default constructors, since we get them anyway from the compiler.

src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
```diff
@@ -0,0 +1,75 @@
...
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+
+namespace NUnit.Framework
+{
+    /// <summary>
+    /// Specify the lifecycle of a Fixture
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
+    public sealed class FixtureLifeCycleAttribute : PropertyAttribute
```
CharliePoole: It seems to me that making this a `PropertyAttribute` rather than setting the context complicates things more than is necessary. Commands, which are intended to be as independent of the tests as reasonably possible, have to examine the test, sometimes even the parent tests, in order to do different things based on life cycle. In the original design, which has admittedly changed on an adhoc basis as problems arose, all the "intelligence" about tests was incorporated in the point where the command chains were generated, not internally in the commands themselves. Each command was intended to do one thing. What we have here is a new "thing" to do, so maybe we need a new command.

CharliePoole: To clarify... I'm OK with a PropertyAtribute if the property is interrogated when the command chains are created, to control what commands are in the chain. Less OK with adding another branch inside an existing command. I'd like to hear what others think about this before deciding, however.

CharliePoole: Alternative to a `PropertyAttribute` is one that implements `IApplyToContext`. Then the commands would still have branches, but they would be context-based and would not require interrogating the test that created the command.

src/NUnitFramework/framework/Internal/Commands/ConstructFixtureCommand.cs
```diff
@@ -37,20 +38,40 @@ public class ConstructFixtureCommand : BeforeTestCommand
         public ConstructFixtureCommand(TestCommand innerCommand)
             : base(innerCommand)
         {
-            Guard.ArgumentValid(Test is TestSuite, "ConstructFixtureCommand must reference a TestSuite", nameof(innerCommand));
```
CharliePoole: This line is what triggered my comments above... ConstructFixtureCommand was designed to work on suites, represented by a `CompositeWorkItem`. That's in part reflected in the fact that it's a `BeforeTestCommand` rather than a `BeforeAndAfterTestCommand`. Repurposing it to work on both suites and test cases introduces a bit more complexity.

I recognize that it's a bit hard to see that we have two different sorts of commands, since they all work against `Tests` and dynamically determine if the test is a suite or not. That's because of the history... we first tried to make one type of command work in both environments, but it turned out that `CompositeWorkItems` needed two chains of commands, whereas `SimpleWorkItems` only need one. It was my hope to come back and refactor this at some point, but it never happened. :disappointed:

src/NUnitFramework/framework/Internal/Commands/DisposeFixtureCommand.cs
```diff
@@ -41,15 +41,17 @@ public class DisposeFixtureCommand : AfterTestCommand
         public DisposeFixtureCommand(TestCommand innerCommand)
             : base(innerCommand)
         {
-            Guard.OperationValid(Test is IDisposableFixture, "DisposeFixtureCommand does not apply to a " + Test.GetType().Name);
```
CharliePoole: Another point where the intelligence is moved __into__ the command rather than resting in the command chain generation logic.

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
```diff
@@ -140,6 +140,13 @@ private TestCommand MakeTestCommand()
                 foreach (var attr in method.GetCustomAttributes<IApplyToContext>(true))
                     command = new ApplyChangesToContextCommand(command, attr);
 
+                // Add construct and optionally dispose command in case of instance per test case.
+                if (parentFixture?.LifeCycle == LifeCycle.InstancePerTestCase)
+                {
+                    command = new ConstructFixtureCommand(command);
+                    if (typeof(IDisposable).IsAssignableFrom(Test.TypeInfo.Type))
+                        command = new DisposeFixtureCommand(command);
+                }
```
CharliePoole: Also good! I'd just prefer you had a new command rather than modifying behavior of `ConstructFixtureCommand`. Obviously, we can't have two `ConstructFixtureCommand`s, so some renaming may be in order.

At some point, I'd like to do a pass through the commands to make them more atomic. I can see other places where what I call "policy logic" has slipped into the commands.

src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
avilv: this makes sense, i switched to a simple NUnitAttribute + IApplyToContext, but im having trouble with two parts

1)
right now i'm using:
```
            var fixture = TestBuilder.MakeFixture(typeof(LifeCycleFixtureTestCountIsAlwaysOne));
            var attr = new FixtureLifeCycleAttribute(LifeCycle.SingleInstance);
```
and attr.ApplyToTest(fixture); to apply the attribute to the fixture, but now im not sure how i can apply a context attribute to a fixture.. i'm probably missing something since i haven't seen it done anywhere in the test code

2)
 i still have to change the command chain construction, in SimpleWorkItem MakeTestCommand() since the IApplyToContext is already executed and i can identify the LifeCycle attribute.
but i still need to make the same change (i assume) in  MakeOneTimeSetUpCommand,  but the context is not applied there yet.. so i have no way to know if the fixture is marked as an InstancePerTestCase fixture

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
avilv: agreed, naming things is one of the hardest things in programming ;)
perhaps ConstructInstancePerTestCaseCommand ? i'm not sure if i'm allowed to change the original ConstructFixtureCommand to ConstructSingleInstanceCommand

src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
CharliePoole: As you know, `IApplyToContext` has one method, `ApplyToContext`. That method modifies the context of the test where it's applied. NUnit's standard command chain call `ApplyToContext` at the appropriate place, when you have used an attribute that implements the interface.

I assume that your attribute will set a new property in the `TestExecutionContext`. Once that is set, all the contexts of any descendant test cases will have the property set. That leaves nothing for you to do except to recognize the context in your own commands.

Many of NUnit's own attributes work this way. See, for example, `SetCultureAttribute` and `SingleThreadedAttribute`. Our tests for such attributes are all in `AppllyToContextTests`.

avilv: right now there's code in in `CompositeWorkItem` `MakeOneTimeSetUpCommand()`  that adds `ConstructFixtureCommand` to the chain:

```csharp
private TestCommand MakeOneTimeSetUpCommand(List<SetUpTearDownItem> setUpTearDown, List<TestActionItem> actions)
{
            TestCommand command = new EmptyTestCommand(Test);

            // Add Action Commands
            int index = actions.Count;
            while (--index >= 0)
                command = new BeforeTestActionCommand(command, actions[index]);

            if (Test.TypeInfo != null)
            {
                // Build the OneTimeSetUpCommands
                foreach (SetUpTearDownItem item in setUpTearDown)
                    command = new OneTimeSetUpCommand(command, item);

                // Construct the fixture if necessary
                if (!Test.TypeInfo.IsStaticClass)
                    command = new ConstructFixtureCommand(command); 
            }

            // Prefix with any IApplyToContext items from attributes
            foreach (var attr in Test.GetCustomAttributes<IApplyToContext>(true))
                command = new ApplyChangesToContextCommand(command, attr); 

            return command;
}

```

this is called from `PerformWork()` **InitializeSetUpAndTearDownCommands**
```
        protected override void PerformWork()
        {
            if (!CheckForCancellation())
                if (Test.RunState == RunState.Explicit && !Filter.IsExplicitMatch(Test))
                    SkipFixture(ResultState.Explicit, GetSkipReason(), null);
                else
                    switch (Test.RunState)
                    {
                        default:
                        case RunState.Runnable:
                        case RunState.Explicit:
                            // Assume success, since the result will otherwise
                            // default to inconclusive.
                            Result.SetResult(ResultState.Success);

                            if (Children.Count > 0)
                            {
                                InitializeSetUpAndTearDownCommands();

                                PerformOneTimeSetUp();
```
all the `IApplyToContext`'s attributes `ApplyToContext` are being executed by `PerformOneTimeSetUp` which happens after **InitializeSetUpAndTearDownCommands**
so at this point, Context.LifeCycle (my new property - added to TestExecutionContext) is not applied yet

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
rprouse: I agree with @CharliePoole here. I think the fact that you are trying to make the `ConstructFixtureCommand` do two things is cascading out into other complexity. Walking up from the test to the suite isn't appropriate there because the class wasn't intended to work on individual tests.

I'm not sure that you need to rename `ConstructFixtureCommand` as that has always been our default behavior. Maybe a new command `ConstructFixturePerTestCaseCommand`?

src/NUnitFramework/testdata/LifeCycleFixture.cs
```diff
@@ -0,0 +1,100 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using NUnit.Framework;
+
+namespace NUnit.TestData.LifeCycleTests
+{
+
+
```
rprouse: Minor nit, but please watch for these extra, unneeded blank lines in the code. Maybe it is a bit OCD of me, but I think it looks sloppy 😄

src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
rprouse: @CharliePoole, you've been following this PR more closely than I have. Do you think that the new instance per fixture command should be created in the `MakeOneTimeSetUpCommand` rather than down in the `SimpleWorkItem`?

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
avilv: i agree @rprouse  - `ConstructFixturePerTestCaseCommand` sounds good, all that's left is the when, see above since im having trouble trying to control it via `Context` (not applied soon enough in the command chain to control via `MakeOneTimeSetUpCommand` )

src/NUnitFramework/testdata/LifeCycleFixture.cs
avilv: haha, will do ;)

CharliePoole:
@avilv @rprouse I'm in the middle of some other work, but I'll come back again on this tomorrow. I think I want to take a step back and look at the entire flow of control through the command chains.

avilv:
hey @CharliePoole @rprouse , i cleaned up some of the minor issues and separated the logic to a `ConstructFixturePerTestCaseCommand` command - until i get some feedback about a way of implementing this via an IApplyContext  approach

src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        /// <summary>
+        /// Construct a FixtureLifeCycleAttribute using default LifeCycle.SingleInstance.
+        /// </summary>
+        public FixtureLifeCycleAttribute() : this(LifeCycle.SingleInstance) { }
```
jnm2: I don't think it would be useful or clear to use `[FixtureLifeCycle]` with no argument.

src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        /// <summary>
+        /// Construct a FixtureLifeCycleAttribute with a specified lifeCycle.
+        /// </summary>
+        /// <param name="lifeCycle">The Lifecycle associated with this attribute.</param>
+        public FixtureLifeCycleAttribute(LifeCycle lifeCycle) : base()
```
jnm2: It's unusual to explicitly specify the default `: base()` call and we don't do it elsewhere.

src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
```diff
@@ -0,0 +1,73 @@
...
+    [AttributeUsage(AttributeTargets.Assembly | AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
+    public sealed class FixtureLifeCycleAttribute : PropertyAttribute
+    {
+        /// <summary>
+        /// Construct a FixtureLifeCycleAttribute using default LifeCycle.SingleInstance.
+        /// </summary>
+        public FixtureLifeCycleAttribute() : this(LifeCycle.SingleInstance) { }
+
+        /// <summary>
+        /// Construct a FixtureLifeCycleAttribute with a specified lifeCycle.
```
jnm2: (nit) Plain english is less awkward: `with a specified life cycle.` If desired, this could also be `with a specified <see cref="LifeCycle"/>.`

If the parameter name is referenced, it should be used as a proper noun, e.g.: `with the value specified by <paramref cref="lifeCycle"/>`. I would personally go with one of the first two options.

src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// </summary>
+        /// <param name="lifeCycle">The Lifecycle associated with this attribute.</param>
```
jnm2: (nit) Unless there is something that cannot be communicated through the parameter name itself, I would remove the param-specific docs. If you keep the param tag, it should describe the parameter's effect like the description property summary which is excellent: `Defines the lifecycle for this test fixture or assembly.` A big-picture kind of message belongs here because it this where the user is making the decision of which value to choose.

src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        /// <summary>
+        /// Overridden to set TestFixture LifeCycle
+        /// </summary>
+        /// <param name="test"></param>
+        public override void ApplyToTest(Test test)
+        {
+            if (test is TestFixture)
+            {
+                ((TestFixture)test).LifeCycle = LifeCycle;
```
jnm2: (nit) I'd prefer to avoid the double type check pattern. You can use C# 7:

```cs
if (test is TestFixture fixture)
    fixture.LifeCycle = LifeCycle;
```

or C# 1+:

```cs
var fixture = test as TestFixture;
if (fixture != null)
    fixture.LifeCycle = LifeCycle;
```

src/NUnitFramework/framework/Attributes/LifeCycle.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <summary>
+        /// A single instance is created and shared for all test cases.
+        /// </summary>
+        SingleInstance,
+
+        /// <summary>
+        /// A new instance is created for each test case.
+        /// </summary>
+        InstancePerTestCase
+    }
```
jnm2: This documentation is great!

src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
```diff
@@ -0,0 +1,67 @@
+// ***********************************************************************
+// Copyright (c) 2017 Charlie Poole, Rob Prouse
```
jnm2:
```suggestion
// Copyright (c) 2019 Charlie Poole, Rob Prouse
```

src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+
+using System;
+using NUnit.Framework.Interfaces;
+
+namespace NUnit.Framework.Internal.Commands
+{
+    /// <summary>
+    /// ConstructFixtureCommand constructs the user test object if necessary.
+    /// </summary>
+    public class ConstructFixturePerTestCaseCommand : BeforeTestCommand
```
jnm2: Is it possible to make this `internal sealed`?

src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            ITest currentTest = Test;
+            while (currentTest != null && (testSuite == null || testFixture == null))
+            {
+                testSuite = testSuite ?? currentTest as TestSuite;
+                testFixture = testFixture ?? currentTest as TestFixture;
+                currentTest = currentTest.Parent;
+            }
+
+            Guard.ArgumentValid(testSuite is TestSuite, "ConstructFixturePerTestCaseCommand must reference a TestSuite", nameof(innerCommand));
```
jnm2: This can just be a null check since `testSuite` is statically known to either be `TestSuite` or null. My null check of choice is `testSuite is object` but you can also do `testSuite != null`.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -214,7 +214,8 @@ private TestCommand MakeOneTimeSetUpCommand(List<SetUpTearDownItem> setUpTearDow
                     command = new OneTimeSetUpCommand(command, item);
 
                 // Construct the fixture if necessary
-                if (!Test.TypeInfo.IsStaticClass)
+                if (!Test.TypeInfo.IsStaticClass &&
+                    (!(Test is TestFixture && ((TestFixture)Test).LifeCycle == LifeCycle.InstancePerTestCase)))
```
jnm2: 💡 Simpler? (again below)

```cs
if (!Test.TypeInfo.IsStaticClass && (Test as TestFixture)?.LifeCycle != LifeCycle.InstancePerTestCase)
```

src/NUnitFramework/framework/Internal/Execution/SimpleWorkItem.cs
```diff
@@ -140,6 +140,13 @@ private TestCommand MakeTestCommand()
                 foreach (var attr in method.GetCustomAttributes<IApplyToContext>(true))
                     command = new ApplyChangesToContextCommand(command, attr);
 
+                // Add construct and optionally dispose command in case of instance per test case.
```
jnm2:
```suggestion
                // Add a construct command and optionally a dispose command in case of instance per test case.
```

Helps me understand better.

src/NUnitFramework/framework/Internal/Tests/TestFixture.cs
```diff
@@ -31,6 +31,11 @@ namespace NUnit.Framework.Internal
     /// </summary>
     public class TestFixture : TestSuite, IDisposableFixture
     {
+        /// <summary>
+        /// The LifeCycle specified for the current test fixture.
```
jnm2: Another example of where English words seem more consistent and less awkward: `The life cycle specified for the current test fixture.`

(There's wouldn't really be much reason to use `<see cref="LifeCycle"/>` since the property type is already being shown.)

src/NUnitFramework/testdata/LifeCycleFixture.cs
```diff
@@ -0,0 +1,92 @@
...
+            Assert.AreEqual(1, Count);
+        }
+
+        [Test]
+        public void CountIsAlwaysOne_2()
+        {
+            Count++;
+            Assert.AreEqual(1, Count);
+        }
+
```
jnm2: @rprouse What do you think about adding a warning with a Ctrl+. fix when there is a blank line before a closing brace or after an opening brace?

src/NUnitFramework/testdata/LifeCycleFixture.cs
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    [TestFixture]
+    [FixtureLifeCycle(LifeCycle.InstancePerTestCase)]
+    public class LifeCycleFixtureInstancePerTestCaseDispose : IDisposable
```
jnm2: The verb `Dispose` should become an adjective at the front of the name:

```suggestion
    public class DisposableLifeCycleFixtureInstancePerTestCase : IDisposable
```

src/NUnitFramework/testdata/LifeCycleFixture.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+        public void Dispose()
+        {
+            DisposeCalls++;
+        }
+    }
+    
+    [TestFixture]
+    [FixtureLifeCycle(LifeCycle.InstancePerTestCase)]
+    public class LifeCycleFixtureInstancePerTestCaseRepeat
```
jnm2:
```suggestion
    public class RepeatingLifeCycleFixtureInstancePerTestCase
```

src/NUnitFramework/tests/Attributes/LifeCycleAttributeTests.cs
```diff
@@ -0,0 +1,86 @@
...
+            var fixture = TestBuilder.MakeFixture(typeof(LifeCycleFixtureTestCountIsAlwaysOne));
+            var attr = new FixtureLifeCycleAttribute(LifeCycle.SingleInstance);
+            attr.ApplyToTest(fixture);
+
+            ITestResult result = TestBuilder.RunTest(fixture);
+            Assert.AreEqual(result.Children.Count(), 2);
+            var childTests = result.Children.ToArray();
+
+            Assert.That(childTests.Any(x => x.ResultState == ResultState.Success));
+            Assert.That(childTests.Any(x => x.ResultState == ResultState.Failure));
```
jnm2: I'd avoid `ToArray()`. Option:

```cs
Assert.That(
    result.Children.Select(t => t.ResultState),
    Is.EquivalentTo(new[] { ResultState.Success, ResultState.Failure });
```

src/NUnitFramework/tests/Attributes/LifeCycleAttributeTests.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        [Test]
+        public void InstancePerTestCaseShouldDisposeForEachTestCase()
+        {
+            LifeCycleFixtureInstancePerTestCaseDispose.DisposeCalls = 0;
+            var fixture = TestBuilder.MakeFixture(typeof(LifeCycleFixtureInstancePerTestCaseDispose));
+            ITestResult result = TestBuilder.RunTest(fixture);
+            Assert.That(result.ResultState.ToString(), Is.EqualTo("Passed"));
```
jnm2: `ToString()` isn't the best for equality tests.

src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
avilv: i guess, but i'm just following the current convention - ConstructFxitureCommand and all other commands are all public right now

avilv:
@jnm2  - done i went through the code again to make sure i didn't miss anything, hoping we can get this merged soon

src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
jnm2: Yes, that makes sense. My memory is saying that we've wanted to move away from adding public types in the Internal namespace from now on. @nunit/framework-team Does that sound right?

src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+            while (currentTest != null && (testSuite == null || testFixture == null))
+            {
+                testSuite = testSuite ?? currentTest as TestSuite;
+                testFixture = testFixture ?? currentTest as TestFixture;
+                currentTest = currentTest.Parent;
+            }
+
+            Guard.ArgumentValid(testSuite != null, "ConstructFixturePerTestCaseCommand must reference a TestSuite", nameof(innerCommand));
+
+            BeforeTest = (context) =>
```
CharliePoole: @avilv Is there some reason that you didn't use a BeforeAndAfter command? That way, the cleanup of the fixture instance could be handled automatically without needing a separate command to do it.

avilv: i thought about it but, but i figured since the convention right now is to have a construct and a dispose command i should try and stick to it.. i don't mind changing it though

src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        /// <summary>
+        /// Defines the life cycle for this test fixture or assembly.
+        /// </summary>
+        public LifeCycle LifeCycle { get; }
+
+        /// <summary>
+        /// Overridden to set a TestFixture's <see cref="LifeCycle"/>.
+        /// </summary>
+        public override void ApplyToTest(Test test)
```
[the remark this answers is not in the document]
avilv: my bad, i was basing my changes on ParallelizeAttribute at first, i changed it to NUnitAttribute, IApplyToTest with my latest commit

src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
avilv: this.. is a good question, i agree on the second approach, as soon as we come to an agreement ill add the test cases and implementation for review if needed.

src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
CharliePoole: That's the architecture for commands that relate to classes, because classes have two different command sequences, one for before and one for after.

But for commands that relate to methods, we generally use beforeandafter commands when there are symmetric things to do because there is only one command sequence constructed. I think this is a bit confusing because we are constructing a class as part of the execution of a method.

avilv: ah.. that makes sense, so say i add the dispose logic in AfterTest, now its doing two things - 
should it still be called `ConstructFixturePerTestCaseCommand` ? 
or perhaps `HandleFixturePerTestCaseCommand`

CharliePoole: Or just `FixturePerTestCaseCommand`?

Anyway, the advantage of Before + After is that the system keeps track for you. After doesn't run unless before ran. If you use two commands, you're supposed to do that yourself.

avilv: done

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -243,7 +243,8 @@ private TestCommand MakeOneTimeTearDownCommand(List<SetUpTearDownItem> setUpTear
                 command = new OneTimeTearDownCommand(command, item);
 
             // Dispose of fixture if necessary
-            if (Test is IDisposableFixture && typeof(IDisposable).IsAssignableFrom(Test.TypeInfo.Type))
+            if (Test is IDisposableFixture && typeof(IDisposable).IsAssignableFrom(Test.TypeInfo.Type) &&
+                (Test as TestFixture)?.LifeCycle != LifeCycle.InstancePerTestCase)
                 command = new DisposeFixtureCommand(command);
```
CharliePoole: Here's a problem I didn't see at first. Let's say the fixture is `LifeCycle.InstancePerTestCase`. In that case, we won't construct it here. Now, let's say that the `OneTimeSetUp` is an instance method that sets some property in the fixture. It's going to throw an NRE.

So, it's really not a matter of choice whether we run `OneTimeSetUp` and `OneTimeTearDown` for each instance. It has to be done that way... unless there is something I'm missing.

I suggest adding a test case that does exactly that. Let it set some property of the fixture to a value. You should see the crash. Then you should fix it. :smiling_imp:

avilv: i would really like to expand the test coverage on this
right now i'm basically hard-coding the LifeCycle flag to InstancePerTestCase then running the entire suite to check for bugs in this mode
since this affects everything, it needs to be tested, well - everywhere
i need to find an approach of testing this without duplicating every test code that was written
any ideas? like a global way to run all the tests with single then instance-per-test-case
what i'm going to do right now is simply find all the tests that fail and shouldn't with the InstancePerTestCase life cycle flag and add them to LifeCycleAttributeTests (basically duplicating the code.. )

but i know that's not ideal

avilv: also about one time, i have a less obtrusive approach to fixing it - i can let MakeOneTimeSetUpCommand still construct the first instance, basically leaving it unchanged
FixturePerTestCaseCommand can then only construct the instance if its null

what do you think? currently it fixes all the failing test cases

avilv: so i just had some free time and i pushed the latest change-set , i added more test cases went through the entire suite (with instance-per-test-case forced on) and made sure it makes sense on all test cases.

the only thing i would like to suggest is.. maybe we should throw an exception when `InstancePerTestCase` is used with an instance `OneTimeSetup/TearDown` method - this can lead to unexpected results since the instance is no longer shared

what do you think ?

src/NUnitFramework/framework/Internal/Commands/ConstructFixturePerTestCaseCommand.cs
jnm2: Sounds good.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
jnm2: @CharliePoole 

> So, it's really not a matter of choice whether we run OneTimeSetUp and OneTimeTearDown for each instance. It has to be done that way... unless there is something I'm missing.

I do not believe that we should run OneTime methods more than one time. There's an inherent contradiction and we should error out the entire fixture and let the user decide how to resolve it, e.g. by moving the code into SetUp and TearDown. This is what I proposed in https://github.com/nunit/nunit/issues/2574.

src/NUnitFramework/framework/Attributes/FixtureLifeCycleAttribute.cs
jnm2: I prefer the second approach too, with the exception that non-static "OneTime" methods cause the fixture to error and require the user to remove the "OneTime" part or choose another resolution.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
CharliePoole: @jnm2 
Of course, that's up to you guys. I have no dog in this fight and was just trying to give you some information that might help.

It does seem to me that there are two contradictory things here and both of them can't be satisfied.

1. OneTimeSetUp definitely sounds like it should only run one time. :smile: 

2. If OneTimeSetUp is an instance method that changes state, it has to be run multiple times in order for the feature to be backward compatible.

Joseph, I'm used to the argument going the other way, with you being in favor of very strong backward compatibility. Why not here?

jnm2: > Joseph, I'm used to the argument going the other way, with you being in favor of very strong backward compatibility. Why not here?

The answer is that they opted in to a brand new feature by starting to use this new attribute. No one's code can break as a result of merely updating the referenced version of NUnit.

> UPDATE: OK, your inline comment clarified that for me. So if people use this feature, they have to have static onetime setup and teardown.

Or they have to change OneTimeSetUp to just SetUp, if that works for them. It might not be the same resolution for all situations.

avilv: i guess we cant have our cake and eat it too, if you use this flag, something has to change in your codebase if you use OneTimeSetUp, i cant think of a reason of even using OneTimeSetUp when its per-instance anyway

either way, whatever you decide im here to do it ;)

manfred-brands: I would like this for C#8 nullability #3613 where the Constructor becomes SetUp and Dispose becomes the TearDown method. This allows then to use non-nullable fields and either far less forgive (!) operators in the code or the removal of #pragma. It also makes Roslyn Analyzer Disposal checks relevant.
method
OneTimeSetUp is sometimes used to create immutable data used by all tests or instantiate auxiliary classes needed. If that functionality can be maintained using a static method and fields would be great. A matching static OneTimeTearDown for cleaning would be also be needed.
If you need the https://github.com/nunit/nunit.analyzers to check for this, I'm willing to work on that there.

rprouse: @manfred-brands you could just use a static constructor for the one time setup in the scenario you described. I'm uncomfortable with the `OneTimeSetup` and `OneTimeTearDown` being allowed here because people may expect them to behave like the instance versions. Do you have a scenario for a static `OneTimeTearDown`?

rprouse:
@avilv I sincerely apologize for the long delay in reviewing this. It is a big change and to be honest, the number of scenarios we need to take into account scares me, so I've been putting off doing an in-depth review for too long. Pullng your code and writing additional unit tests for things that I am concerned about helped and uncovered a few issues. I am pretty impressed however with how close it is, so good work.

I fully understand if you've left this behind and don't want to continue working on it. The PR has been open a long time and there have been a lot of comments to keep up with. You did pick off a very meaty issue 😈 

Can you let me know if you would like to continue working on this? If you do, I will leave it with you and promise to respond much more quickly going forward. If you don't, then I will try to find someone else to pick up the torch. Either way, thanks for all your work so far.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
manfred-brands: @rprouse We use `OneTimeSetUp` for expensive setups, like creating files(databases) which are then used by multiple tests. The `OneTimeTearDown` would then delete these. Although the tests would work if those methods are called for every test, they would certainly increase the time for the test run. So the behaviour would be the same. But it would not be  OneTime. The runner could verify that it is a static method which in turn means it can only use static fields etc.

manfred-brands:
> `OneTimeSetup` and `OneTimeTearDown` are called once before/after all of the tests. This will confuse people, so they should cause errors.

Isn't that what they should do? The documentation says:
_Identifies a method that is called once to perform setup before any child tests are run._ and
_Identifies a method to be called once after all the child tests have run._

Yes, if those method set member fields. then 2nd test will fail as it isn't set. Hence my previous comment verifying these methods are `static`.

rprouse:
@manfred-brands I'd be okay if it only allows static OneTimeSetup and teardown, but this doesn't do that yet either. I'd like to make it hard for users to make mistakes like setting instance variables.

CharliePoole:
This is extremely hard to do as a non-breaking change.

In my experimental microtesting framework, I disallowed all types of SetUp and TearDown. If I later wanted to support integration or system tests, I could introduce them without breaking anything by defining a different fixture type. Then users would have to clearly state what type of test they were writing.

However, NUnit 3 has passed that point already. The existing TestFixture has to support all sorts of testing. The user makes separate choices about whether to use static or instance members, one-time setup, standard setup, etc... and now... life-cycle. The fact that these are separate suggests that the user may pick any combination of those choices, which is scary from a support point of view. It seems more robust to define two bundles of supported features and have the user select which one to use.

Bundle 1: Existing TestFixture capabilities at least for the life of NUnit 3.
Bundle 2: Fixture per test case [+ Setup/TearDown [+ Static OneTimeSetUp/TearDown]]

The distinction could be signaled by the user in one of three (that I can think of) ways:
1. Separate attribute per bundle
2. Separate attribute for lifecycle
3. Property on TestFixtureAttribute

Alternative 1 is most likely impractical before NUnit 4.

Alternative 2 is what the current PR does. It's main drawback is with parameterized and generic fixtures. All instances are forced to use the same lifecycle because of how .NET identifies attributes.

Alternative 3 doesn't have that drawback and is also easily convertible to a future implementation of alternative 1.

rprouse:
> 1. Separate attribute per bundle
> 2. Separate attribute for lifecycle
> 3. Property on TestFixtureAttribute
> 
> Alternative 1 is most likely impractical before NUnit 4.
> 
> Alternative 2 is what the current PR does. It's main drawback is with parameterized and generic fixtures. All instances are forced to use the same lifecycle because of how .NET identifies attributes.
> 
> Alternative 3 doesn't have that drawback and is also easily convertible to a future implementation of alternative 1.

I agree that alternative 1 is inpractical for now.

I don't think that the many users would want to use a different lifecycle for the same parameterized or generic fixture so I don't think it is a major drawback. To me, the lifecycle you choose guides how you write your tests.

Alternative 3 has the drawback that the TestFixtureAttribute then needs to be applied to every class, whereas alternative 2 can be applied at the assembly level so that you can select the lifecycle for all of your tests easily.

CharliePoole:
@rprouse Yes, I do see your point about applying at the assembly level.

If you allow that, I hope it will come with lots of warnings. I think we misled many users with the easy availability of parallelism at the assembly level. It seems to me that many folks assumed we would be doing something to __make__ their tests work in parallel. At least I've had to point out quite a lot that all the attribute does is __promise__ that the tests are parallelizable!

For V4, I hope you'll give option 1 some consideration. Most non-unit tests should probably share the fixture and run steps sequentially. OTOH, web tests may be an exception and they are, of course, the most common sort of non-unit tests these days. I can imagine having different fixture types for unit and functional tests and maybe even a separate type for web tests.

avilv:
hey just letting you guys know, i'm still alive, willing to make changes to the PR as needed
just to add my 2 cents to the discussion
its a complicated issue, i don't see a way to support the current paradigm, a lot of things will work different with this attribute
but when i went into NUnit i was sure this is how it works by default, and having to support parallelism in my tests was nothing less than a 'breaking change', and a lot of boilerplate, essentially fighting the framework i'm using

i see this as a 'NUnit Next' trigger, lots of warning clauses, what you have currently *might* work as expected, might not, but for me at least, would help tremendously 
at work we use a fork of NUnit with this change for a lot of integration tests that needed minimal changes
again this is just my subjective experience

i'm willing to make any change needed to make this better/safer/more easy to use, less to abuse

gleb-osokin:
Hi @avilv, I've created a PR to your branch, that addresses the issues, mentioned by @rprouse .
Please, see them here: https://github.com/avilv/nunit/pull/1

This is really one feature, that I'd love to see in NUnit.

rprouse:
@gleb-osokin at first pass, that looks great. Thanks for helping out.

@avilv if you merge @gleb-osokin's PR, it will be included in this PR. Either that, or @gleb-osokin can create a new PR to NUnit and we can close this one. Either way, your commits will still be in the history and you will still show up in the contributors page.

I am going to pull the updated branch and run it against another test project that I'm building for this. I'd like to test a mix of features and see how they behave before we merge, call it a quasi-integration test. 😸

avilv:
> @gleb-osokin at first pass, that looks great. Thanks for helping out.
> 
> @avilv if you merge @gleb-osokin's PR, it will be included in this PR. Either that, or @gleb-osokin can create a new PR to NUnit and we can close this one. Either way, your commits will still be in the history and you will still show up in the contributors page.
> 
> I am going to pull the updated branch and run it against another test project that I'm building for this. I'd like to test a mix of features and see how they behave before we merge, call it a quasi-integration test. 😸

merged, here's hoping we pass all tests, and great job @gleb-osokin for helping out

rprouse:
I've tested the following in combination with the new `FixtureLifeCycle`;

- [x] Every test method gets a new class instance
- [x] `Dispose` happens in the correct order
- [x] `OneTimeSetup` and `OneTimeTearDown` must be static and is only called once
- [x] `SetUp` and `TearDown` are called before and after each method
- [x] `Order` attribute is respected
- [ ] Parallel tests with a constructor, constructor is called before every test
- [ ] Parallel tests with `SetUp` is called before every test and every test has it's own instance variables

Unfortunately, the Parallel tests failed. Luckily the fix is simple. I will add it to this PR, but first I want to see about adding unit tests for that. I tested all of the above and my fix in an integration test project.

rprouse:
I cannot reproduce the problem with parallel from the NUnit tests themselves. I even copied the tests from my integration tests verbatim into the NUnit test project and they pass there but fail in my integration project. It is late and I can't see what the difference is, so I'm going to leave it until tomorrow.

src/NUnitFramework/framework/Internal/Commands/FixturePerTestCaseCommand.cs
```diff
@@ -0,0 +1,71 @@
...
+                currentTest = currentTest.Parent;
+            }
+
+            Guard.ArgumentValid(testSuite != null, "FixturePerTestCaseCommand must reference a TestSuite", nameof(innerCommand));
+
+            ITypeInfo typeInfo = testSuite.TypeInfo;
+
+            BeforeTest = (context) =>
+            {
+                if (typeInfo != null && !typeInfo.IsStaticClass && testSuite.Fixture == null)
```
rprouse, line 56: So, it turns out the reason I am having problems is that the behavior is inconsistent. I get different behavior for the tests on different runs and if I run the tests in the debugger or straight from the command line. In the debugger, the tests always pass, but on the command line they **usually** fail. It sounds like a classic threading problem, but I can't figure out what is going on because they always behave correctly in the debugger. 🤕 

From what I have managed to see in the debugger, I have a theory. In some of the test runs, the `testSuite.Fixture` is not null, so we don't actually construct the object below, instead we use the previously constructed one. I believe that this works fine single threaded, but when we are running in parallel, the first test sets the **one** `testSuite.Fixture` on line 60 below, then subsequent parallel tests don't enter this block because it is no longer `null`.

@avilv do you remember why you test if the Fixture is null and set it below? It also seems wrong to set it back to `null` in the `AfterTest` if it wasn't set here.

rprouse:
My fixes with comments are in a PR against @avilv's fork at https://github.com/avilv/nunit/pull/2

rprouse (approved):
I am ready to approve this. It will need one more approval from the @nunit/framework-team

gleb-osokin:
@CharliePoole, can you please check if your request for changes has been addressed? I think it is also a blocker for this PR.

CharliePoole (approved):

CharliePoole:
@gleb-osokin That's not quite true. A PR requires approval from two team members to be merged, not two people. (I'm not a team member.) 

However, since I raised an objection, I approved it now so it's clear that I'm satisfied. (Sorry, I actually try to only make comments rather than request changes so this does not happen.) Anyway, Rob has approved and I think you need another team member to approve and merge it.

gleb-osokin:
@mikkelbu do you think you can review this PR if you have time? Would be nice to finally finish this one before it's one year anniversary.

janissimsons:
Does this mean that feature #2574 is now implemented and ready to be used? 
When an official release is planned to ship this feature?

rprouse:
@janissimsons sorry for the delay responding, I don't check on closed PRs often. This will be in the 3.13 release which I expect to complete within the week.

src/NUnitFramework/framework/Internal/Commands/FixturePerTestCaseCommand.cs
```diff
@@ -0,0 +1,69 @@
...
+
+namespace NUnit.Framework.Internal.Commands
+{
+    /// <summary>
+    /// ConstructFixtureCommand constructs the user test object if necessary.
+    /// </summary>
+    public class FixturePerTestCaseCommand : BeforeAndAfterTestCommand
+    {
+        /// <summary>
+        /// Handles the construction and disposement of a fixture per test case
```
jnm2, line 35: XML docs here and on the class look like they were copied from somewhere else and don't apply.

src/NUnitFramework/framework/Internal/Execution/IMethodValidator.cs
```diff
@@ -0,0 +1,39 @@
...
+// ***********************************************************************
+
+using System.Reflection;
+
+namespace NUnit.Framework.Internal.Execution
+{
+    /// <summary>
+    /// Validates method to execute.
+    /// </summary>
+    public interface IMethodValidator
```
jnm2, line 31: An alternative to adding new public types would be to pass around `Action<MethodInfo> methodValidator` instead of `IMethodValidator methodValidator`, but this is the Internal namespace so it should be fine.

PR #3450, opened 2019-12-27
Nullability annotations for the Assert class

jnm2:
Annotating any more than a class at a time will be overwhelming to review. I'm starting with the most-seen types.

The intent with all annotations is to reflect current behavior. If `null` is tolerated, I'm marking the parameter as nullable. If it causes ArgumentNullException, NullReferenceException, etc, I'm marking as non-nullable.

### DIFF
--- a/src/NUnitFramework/Directory.Build.props
+++ b/src/NUnitFramework/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <LangVersion Condition="'$(MSBuildProjectExtension)' == '.csproj'">7</LangVersion>
+    <LangVersion Condition="'$(MSBuildProjectExtension)' == '.csproj'">8</LangVersion>
     <Features>strict</Features>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
--- a/src/NUnitFramework/framework/Assert.Conditions.cs
+++ b/src/NUnitFramework/framework/Assert.Conditions.cs
@@ -21,6 +21,8 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+#nullable enable
+
 using System.Collections;
 using NUnit.Framework.Constraints;
 using System;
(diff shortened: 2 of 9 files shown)


src/NUnitFramework/framework/Assert.Conditions.cs
```diff
@@ -32,32 +34,29 @@ public abstract partial class Assert
         #region True
 
         /// <summary>
-        /// Asserts that a condition is true. If the condition is false the method throws
-        /// an <see cref="AssertionException"/>.
+        /// Asserts that a condition is true. Returns without throwing an exception when inside a multiple assert block.
```
jnm2: Folks are going to ask why we don't add `[DoesNotReturnIf(...)]` and other postconditions unless we document that no postconditions are possible due to the multiple assert feature.

src/NUnitFramework/framework/Assert.cs
```diff
@@ -313,8 +314,10 @@ public static void Contains(object expected, ICollection actual)
         /// <param name="testDelegate">A TestDelegate to be executed in Multiple Assertion mode.</param>
         public static void Multiple(TestDelegate testDelegate)
         {
-            TestExecutionContext context = TestExecutionContext.CurrentContext;
-            Guard.OperationValid(context != null, "There is no current test execution context.");
+            if (testDelegate is null) throw new ArgumentNullException(nameof(testDelegate));
+
+            TestExecutionContext context = TestExecutionContext.CurrentContext
+                ?? throw new InvalidOperationException("There is no current test execution context.");
```
jnm2: `Guard.OperationValid` is not recognized as contributing to the null check, so this avoids having to suppress a warning in our own code with `!`.

src/NUnitFramework/Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <LangVersion Condition="'$(MSBuildProjectExtension)' == '.csproj'">7</LangVersion>
+    <LangVersion Condition="'$(MSBuildProjectExtension)' == '.csproj'">8</LangVersion>
```
stevenaw: I'm probably not a good reviewer of the nullability stuff, but this change caught my eye. Would this require VS2019 as a build tool? If so, we should update BUILDING.md, as it still calls out VS2017 as a req.

jnm2: Yes, thank you!

src/NUnitFramework/framework/Assert.cs
rprouse: Interesting. I wonder if there is a way to signal that?

jnm2: Wait a minute, I wasn't thinking properly when I said that. We can annotate our `Guard.OperationValid` parameter with `[DoesNotReturnIf(false)]` to achieve the desired effect. I'll fix this with my next PR.

jnm2: Wait, now I remember why I didn't annotate it. I have to declare an internal copy of DoesNotReturnIfAttribute. I can do that though.

PR #3465, opened 2020-02-03
Prevent ExecutionContext from flowing between test cases

### LINKED ISSUE #3283: ExecutionContext is flowed between unrelated tests
Same as https://github.com/nunit/nunit/issues/2821, but for the ExecutionContext rather than the SynchronizationContext. (.NET tracks them individually.)

One test should not be able to affect another test simply because the test execution scheduler happens to run them on the same worker thread. This has unstable results depending on how many worker threads there are and on how NUnit decides to order them over time. But more importantly, it's an invisible factor that influences every test so that you may not be actually testing what you think you are testing.

Each test is inheriting an unrelated ExecutionContext from the last test the worker happened to pick up. This can cause some tests to fail when they should pass. Much worse, as in some cases just now for me, it can silently render some tests ineffective (passing whether or not they should fail).

### Repro

Some tests designed to demonstrate the problem by failing:

```cs
public static class SystemUnderTest
{
    private static readonly AsyncLocal<int> ValueFlowedWithExecutionContext = new AsyncLocal<int>();

    public static int Increment() => ++ValueFlowedWithExecutionContext.Value;
}

public static class Tests
{
    // Only one test succeeds because the .NET ExecutionContext is flowed between tests
    // in whatever order they happen to execute.

    [Test]
    public static void Test1()
    {
        Assert.That(SystemUnderTest.Increment(), Is.EqualTo(1));
    }

    [Test]
    public static void Test2()
    {
        Assert.That(SystemUnderTest.Increment(), Is.EqualTo(1));
    }

    [Test]
    public static void Test3()
    {
        Assert.That(SystemUnderTest.Increment(), Is.EqualTo(1));
    }
}
```

### Workarounds

The easiest workaround I'm aware of is to put `[RequireThread]` on every test method. This comes with a performance hit. It's also easy to forget if it breaks your test by causing it to pass rather than by causing it to fail.

A more efficient workaround would be:

```cs
private static void DoInIsolatedExecutionContext(Action action)
{
    using (var copy = ExecutionContext.Capture().CreateCopy())
    {
        ExecutionContext.Run(copy, state => ((Action)state).Invoke(), state: action);
    }
}

[Test]
public static void TestN()
{
    DoInIsolatedExecutionContext(() =>
    {
        Assert.That(SystemUnderTest.Increment(), Is.EqualTo(1));
    });
}
```

### Concerns

Would some NUnit users be making use of this behavior as a feature when all tests execute on a single thread and they are constrained to execute in order?

```cs

public static class Tests
{
    [Test, Order(1)]
    public static void Test1()
    {
        Assert.That(SystemUnderTest.Increment(), Is.EqualTo(1));
    }

    [Test, Order(2)]
    public static void Test2()
    {
        // Build on the previous test (where ‘previous’ is codified via OrderAttribute)
        Assert.That(SystemUnderTest.Increment(), Is.EqualTo(2));
    }

    [Test, Order(3)]
    public static void Test3()
    {
        // Build on the previous test (where ‘previous’ is codified via OrderAttribute)
        Assert.That(SystemUnderTest.Increment(), Is.EqualTo(3));
    }
}
```

The SynchronizationContext is no longer flowed between tests as of 3.11 (https://github.com/nunit/nunit/issues/2821) which is very similar though it affects different things. No one has reported noticing the difference.

jnm2:
Fixes #3283, plus:

- ValuesSource is now isolated the same way as TestCaseSource (missed in https://github.com/nunit/nunit/pull/2966).

- Setup and teardown methods were being executed out of order on .NET 3.5 and 4.0. They were being sorted using a BaseTypesFirstComparer implementation that broke the `IComparer<>.Compare` invariants of reflexivity, symmetry, and transitivity. This means that any .NET sorting method that used it was specified to have undefined behavior.
  I fixed it in this PR because the differing order of execution depending on target framework made it impossible to write [`ExecutionContextFlowFixture.cs`](https://github.com/nunit/nunit/pull/3465/files#diff-9621d3098df833cc8f0b0fb0713e4406) clearly.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Reflect.cs
+++ b/src/NUnitFramework/framework/Internal/Reflect.cs
@@ -67,34 +67,23 @@ namespace NUnit.Framework.Internal
         /// <returns>The array of methods found</returns>
         public static MethodInfo[] GetMethodsWithAttribute(Type fixtureType, Type attributeType, bool inherit)
         {
-            List<MethodInfo> list = new List<MethodInfo>();
-
-            var flags = AllMembers | (inherit ? BindingFlags.FlattenHierarchy : BindingFlags.DeclaredOnly);
-            foreach (MethodInfo method in fixtureType.GetMethods(flags))
+            if (!inherit)
             {
-                if (method.IsDefined(attributeType, inherit))
-                    list.Add(method);
+                return fixtureType
+                   .GetMethods(AllMembers | BindingFlags.DeclaredOnly)
+                   .Where(method => method.IsDefined(attributeType, inherit: false))
+                   .ToArray();
             }
 
-            list.Sort(new BaseTypesFirstComparer());
+            var methodsByDeclaringType = fixtureType
+                .GetMethods(AllMembers | BindingFlags.FlattenHierarchy) // FlattenHierarchy is complex to replicate by looping over base types with DeclaredOnly.
+                .Where(method => method.IsDefined(attributeType, inherit: true))
+                .ToLookup(method => method.DeclaringType);
 
-            return list.ToArray();
-        }
-
-        private class BaseTypesFirstComparer : IComparer<MethodInfo>
-        {
-            public int Compare(MethodInfo m1, MethodInfo m2)
-            {
-                if (m1 == null || m2 == null) return 0;
-
-                Type m1Type = m1.DeclaringType;
-                Type m2Type = m2.DeclaringType;
-
-                if (m1Type == m2Type) return 0;
-                if (m1Type.IsAssignableFrom(m2Type)) return -1;
-
-                return 1;
-            }
+            return fixtureType.TypeAndBaseTypes()
+                .Reverse()
+                .SelectMany(declaringType => methodsByDeclaringType[declaringType])
+                .ToArray();
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Internal/Reflect.cs
+++ b/src/NUnitFramework/framework/Internal/Reflect.cs
@@ -58,13 +58,13 @@ namespace NUnit.Framework.Internal
         #region Get Methods of a type
 
         /// <summary>
-        /// Examine a fixture type and return an array of methods having a
-        /// particular attribute. The array is order with base methods first.
+        /// Returns all methods declared by the specified fixture type that have the specified attribute, optionally
+        /// including base classes. Methods from a base class are always returned before methods from a class that
+        /// inherits from it.
         /// </summary>
-        /// <param name="fixtureType">The type to examine</param>
-        /// <param name="attributeType">The attribute Type to look for</param>
-        /// <param name="inherit">Specifies whether to search the fixture type inheritance chain</param>
-        /// <returns>The array of methods found</returns>
+        /// <param name="fixtureType">The type to examine.</param>
+        /// <param name="attributeType">Only methods to which this attribute is applied will be returned.</param>
+        /// <param name="inherit">Specifies whether to search the fixture type inheritance chain.</param>
         public static MethodInfo[] GetMethodsWithAttribute(Type fixtureType, Type attributeType, bool inherit)
         {
             if (!inherit)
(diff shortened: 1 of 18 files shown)


src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -67,33 +67,29 @@ public static class Reflect
...
-            List<MethodInfo> list = new List<MethodInfo>();
-
             var flags = AllMembers | (inherit ? BindingFlags.FlattenHierarchy : BindingFlags.DeclaredOnly);
-            foreach (MethodInfo method in fixtureType.GetMethods(flags))
-            {
-                if (method.IsDefined(attributeType, inherit))
-                    list.Add(method);
-            }
 
-            list.Sort(new BaseTypesFirstComparer());
```
jnm2: There are half a dozen other instances of `List<>.Sort` that I didn't touch that are suspect as well. We should not be using `List<>.Sort` or `Array.Sort<>` if there is a chance that two different items could compare as equal.

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -67,33 +67,29 @@ public static class Reflect
...
-                Type m2Type = m2.DeclaringType;
+            private BaseTypesFirstComparer() { }
 
-                if (m1Type == m2Type) return 0;
-                if (m1Type.IsAssignableFrom(m2Type)) return -1;
+            public int Compare(Type x, Type y)
+            {
+                if (x == y) return 0;
+                if (x is null) return -1;
+                if (y is null) return 1;
```
jnm2: Previously, this method returned `0` when either parameter was null. This would have made the comparison no longer transitive. When A < B, the sorting method could compare and find that A = null and B = null and would conclude that A = B. Inconsistent values returned from this method means undefined behavior for the sorting method.

https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.icomparer-1.compare#remarks

> Comparing `null` with any reference type is allowed and does not generate an exception. A null reference is considered to be less than any reference that is not null.

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -67,33 +67,29 @@ public static class Reflect
...
 
-                return 1;
+                if (x.IsSubclassOf(y)) return -1;
+                if (y.IsSubclassOf(x)) return 1;
```
jnm2: `IsAssignableFrom` searches more than just base types.

src/NUnitFramework/tests/Assertions/AdhocTestExecutionTests.cs
```diff
@@ -47,7 +47,7 @@ public void CanCallAssertWithoutTestExecutionContext(MethodInfo method)
                 throw testException;
         }
 
-#if !(NET35 || NET40 || NET45)
+#if !(NET35 || NET40 || NET45 || NET46) // NET46 should be removed when the framework project adds a net46 target
```
jnm2: We could consider adding a net46 target to the framework so that it can use AsyncLocal.

https://github.com/nunit/nunit/blob/30df2d13c0f67eeb6b3dc7f56eebd7c5cfb5c7de/src/NUnitFramework/framework/Internal/TestExecutionContext.cs#L144-L195

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: I've just realized that we still have the intransitivity problem. Imagine these types:

```cs
class A { }
class B : A { }
class C { }
```

This comparer should consider A < B, A == C, and B == C. "Base types first" is an intransitive comparison. We can't implement it using any of the BCL sorting methods because they rely on the assumption that `IComparer<>.Compare` is always transitive.

jnm2: Alternatively, we could refrain from getting the methods out of order and having to sort them in the first place. I'll implement this.

src/NUnitFramework/tests/Assertions/AdhocTestExecutionTests.cs
rprouse: Agreed. I've been looking for features in 4.6, 4.7 and 4.8 that would justify any of those targets but couldn't think of any. This might be one...

jnm2:
Thanks Rob!

PR #3486, opened 2020-03-02
Addition of DictionaryContainsKeyValuePairConstraint

### LINKED ISSUE #3470: Assertion for key-value-pair
I was looking for a good way to check key-value-pairs. I know about `DictionaryContainsKeyConstraint` and `DictionaryContainsValueConstraint`, but have not found any option to combine them on the same item. Chaining it with `And` will also operate on the asserted value and not on the item identified by the key.

```cs
public void Merge_NoConflicts()
{
    var mainDict = new Dictionary<string, int>
    {
        { "Test", 1 }
    };
    var otherDict = new Dictionary<string, int>
    {
        { "Test2", 42 }
    };

    var result = mainDict.Merge(otherDict);

    // --- Variant 1 (full checks, ugly to read) ---
    Assert.That(result, Contains.Key("Test"));
    Assert.That(result["Test"], Is.EqualTo(1));
    Assert.That(result, Contains.Key("Test2"));
    Assert.That(result["Test2"], Is.EqualTo(42));

    // --- Variant 2 (better but throws KeyNotFoundException, would like outlike like SomeItemsConstraint) ---
    Assert.That(result["Test"], Is.EqualTo(1));
    Assert.That(result["Test2"], Is.EqualTo(42));

    // --- Variant 3 (works, better than 1 but but uglier to read than 2) ---
    Assert.That(result, Contains.Item(new KeyValuePair<string, int>("Test", 1)));
    Assert.That(result, Contains.Item(new KeyValuePair<string, int>("Test2", 42)));

    // --- Variant 4 (looks correct, but ContainValue is called on result, so no relation to the pair) ---
    Assert.That(result, Contains.Key("Test").And.ContainValue(1));
    Assert.That(result, Contains.Key("Test2").And.ContainValue(42));

    // --- Variant 5 (some API I would prefer but can't find) ---
    Assert.That(result, Contains.Key("Test").WithValue(1));
    Assert.That(result, Contains.Key("Test2").WithValue(42));
}
```

It would be pretty easy to implement an `DictionaryContainsKeyValueConstraint` that implements the `Matches` method using `TryGet`. It's also easy to add the `WithValue` method to `DictionaryContainsKeyConstraint` and have it work with all notations (`Contains.Key`, `Does.ContainKey`, ...). But before doing a PR, I wanted to know if there already is a good way of doing it and just not finding it.

Falco20019:
Fixes #3470

Just a first idea to resolve it. Not sure if I implemented `WithValue` in `DictionaryContainsKeyConstraint` correctly. Also not sure if using `KeyValuePair<TKey, TValue>` is better or using the approach already used for properties (`ResolvableConstraintExpression`?) is better. This seemed to have worked for `Does.ContainKey.WithValue` and also `Does.Not.ContainKey.WithValue`.

Looking forward to feedback :)

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
@@ -89,7 +89,7 @@ namespace NUnit.Framework.Constraints
                 builder.Append(this);
             }
 
-            var constraint = new DictionaryContainsKeyValuePairConstraint(Expected, expected);
+            var constraint = new DictionaryContainsKeyValuePairConstraint(Expected, expectedValue);
             builder.Append(constraint);
             return constraint;
         }
--- a/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
@@ -77,6 +77,24 @@ namespace NUnit.Framework.Constraints
         protected object Expected { get; }
 
         /// <summary>
+        /// Returns a new DictionaryContainsKeyValuePairConstraint checking for the
+        /// presence of a particular key-value-pair in the dictionary.
+        /// </summary>
+        public DictionaryContainsKeyValuePairConstraint WithValue(object expected)
+        {
+            var builder = this.Builder;
+            if (builder == null)
+            {
+                builder = new ConstraintBuilder();
+                builder.Append(this);
+            }
+
+            var constraint = new DictionaryContainsKeyValuePairConstraint(Expected, expected);
+            builder.Append(constraint);
+            return constraint;
+        }
+
+        /// <summary>
         /// Flag the constraint to ignore case and return self.
         /// </summary>
         [Obsolete(ComparerMemberObsoletionMessage)]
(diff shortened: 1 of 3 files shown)


Falco20019:
@rprouse / @nunit/framework-team
Any news on the second review?

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -76,6 +76,24 @@ public override string Description
...
+            var builder = this.Builder;
+            if (builder == null)
+            {
+                builder = new ConstraintBuilder();
+                builder.Append(this);
+            }
+
+            var constraint = new DictionaryContainsKeyValuePairConstraint(Expected, expected);
+            builder.Append(constraint);
+            return constraint;
```
mikkelbu, line 94: I was a bit unsure about this part, but it seems to handle `Assert.That(dictionary, Does.Not.ContainKey(...).WithValue....` nicely

Falco20019: Also not sure what's the best way to solve it since I'm not too deeply into the code base. I found that approach in another class (not sure anymore where it was). So if there's a better way, let me know!

mikkelbu:
Thanks for supplying this PR @Falco20019, and sorry about the slow feedback from our side.

Falco20019:
Thanks for updating the PR. I totally forgot that I had open suggestions.

Issue #3376, opened 2019-09-23
Nullable Reference Types annotations


jnm2:
tl;dr Now that C# 8 is out, people will be using the Nullable Reference Types language feature. We should consider putting the compiler annotations on our library. That would require moving the codebase to C# 8. To reap the benefits in our own codebase, we could also use https://github.com/tunnelvisionlabs/ReferenceAssemblyAnnotator.

### What it is

People who put `<Nullable>enable</Nullable>` in their projects or `#nullable enable` in their source files are then allowed to use the syntax `ReferenceType?` where nulls are permitted. In typical software, nulls end up being actually intended in only a small percentage of the usages of reference types.

The compiler then produces warnings (we have /warnaserror) when a thing that might be null is treated as though it can't be null, like when dotting off which could cause an NRE or when passing to a method which could cause an ArgumentNullException, or when setting a field or property which only causes a problem at a later point in time.

This goes a long way toward solving the billion-dollar problem of nullability blindness in the type system.

https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references
https://devblogs.microsoft.com/dotnet/nullable-reference-types-in-csharp/
https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/nullable-reference-types

Things like `string.IsNullOrEmpty` are annotated with new compiler attributes such as `[NotNullWhen(false)]` on its parameter so that the compiler can recognize that it's a null check. These attributes can be defined by anyone as internal so that they don't conflict with other assemblies. .NET Core 3.0 is the first version of the BCL that adds these attributes as public classes. Referencing https://github.com/tunnelvisionlabs/ReferenceAssemblyAnnotator would define these internally for us without having to maintain the internal attribute definitions ourselves.

### Reasons to do it

- It would be nice to annotate our library so that people who use NUnit are warned if they forget to check things like `TestContext.CurrentContext` for null or if they pass things that could be null to our APIs. It's being up-front about the types that each API accepts in the same way that declaring a parameter `IFoo foo` instead of `object foo` is being more up-front and helpful.

- This would aid the mission of maintaining high-quality internal code written by many different people. We've had our share of null reference exceptions that nothing had forced us to notice at compile time.

- It's not a breaking change.

### Reasons not to do it

[...]

PR #3488, opened 2020-03-03
Annotate ITestData and implementations

jnm2:
This is a central point around which a lot of interfaces and implementations turn.
Contributes to https://github.com/nunit/nunit/issues/3376.

#### Cleanup

1. Two files had line endings that didn't match

#### Annotation

2. Annotate ITestData and implementations
3. Enable NRTs for usages of ITestData to validate annotations

#### Fix

4. Fix NUnit crash found by nullability warnings

#### Refactoring

5. Always initialize TestParameters.Arguments because past behavior is inconsistent with ITestData usage
6. Move TestMethod.NoArguments to TestParameters.NoArguments
7. Refactor InitializeArguments into a more analysis-friendly constructor (fixes remaining warnings)

#### Improvement

8. Add null guards that should have been present anyway to fix warnings in VS prior to 16.5


src/NUnitFramework/framework/Internal/PropertyBag.cs
```diff
@@ -1,177 +1,178 @@
...
-        /// <summary>
-        /// Adds a key/value pair to the property set
-        /// </summary>
-        /// <param name="key">The key</param>
-        /// <param name="value">The value</param>
-        public void Add(string key, object value)
-        {
-            Guard.ArgumentNotNull(value, "value");
-
+// ***********************************************************************
```
rprouse: Is this a change in line endings?

jnm2: Ah... yes. Oops. Can I push to split this into two commits?

jnm2:
ExpectedResultDoesNotBlockApplyToTestAttributes is failing, looking into it.

src/NUnitFramework/framework/Internal/PropertyBag.cs
jnm2: TestCaseAttribute has the same problem.

src/NUnitFramework/framework/Internal/TestParameters.cs
```diff
@@ -40,6 +40,7 @@ public abstract class TestParameters : ITestData, IApplyToTest
         public TestParameters()
         {
             RunState = RunState.Runnable;
+            InitializeArguments(new object[0]);
```
jnm2: This is the change causing ExpectedResultDoesNotBlockApplyToTestAttributes to fail.

jnm2: And this is the conflict:

https://github.com/nunit/nunit/blob/22ea163f1dbbecdb71b88270434c948f6e381bbd/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs#L53-L65

jnm2: parms.Arguments == null is being used as in indirect way to tell whether to overwrite `parms` properties with method-level attributes. There needs to either be a direct way or the merge order needs to be swapped.

jnm2: `parms.Arguments.Length == 0` seems to satisfy all the tests, so I'll leave it at that for simplicity. I guess that would have been the original code except for the inconsistent behavior of some TestParameters constructors leaving Arguments null.

src/NUnitFramework/framework/Internal/PropertyBag.cs
jnm2: @rprouse I now have an initial commit converting these two files to CRLF, but I'm wondering if that was the wrong thing to do. Shouldn't Git be autoconverting for all files unless specified otherwise in .gitattributes?

PR #3493, opened 2020-03-07
Annotate builder and reflection interfaces

jnm2:
Contributes to https://github.com/nunit/nunit/issues/3376.

#### Builder interfaces

1. 🏷 Annotated builder interfaces
2. 🏷 Annotate Internal/Builders folder to validate interface annotations

#### Reflection interfaces

3. 🏷 Annotate reflection interfaces and implementations
4. 🐛 Fix NRE in Reflect class found by nullability warnings
5. 🏷 Annotate TypeHelper to validate implementations
6. 🐛 Fix NRE in TypeHelper class found by nullability warnings

#### Tooling workarounds

7. 🔇 Suppression needed until Visual Studio 16.5 is out


jnm2:
@rprouse I resolved the merge conflict with the line-ending fix.

mikkelbu:
@jnm2 I know we have already merged a handful of "Annotate..." PRs (and I know that you have worked more with nullable than I 😄), but should we be worried by https://github.com/xunit/assert.xunit/pull/36#issuecomment-572719747, or can we assume that people are not consuming our source ? or if they are then we require C# 8.

ChrisMaddock:
Another question to add to Mikkel's - I'm not too familiar with the effects of these changes - is there anything we should particularly be looking out for in review? The compiler should catch any problems really, right? 🤔

jnm2:
Consuming a library as source is not something I've had to deal with when annotating a library. You're right. With the changes already merged to master, people can't use NUnit as source unless they set LangVersion >= 8. In our last release, we required LangVersion >= 7 in projects that use NUnit as source. We've moved things around several times like when we moved away from legacy csproj and I'm not aware of once receiving a comment from anyone using NUnit as source.

@nunit/framework-team We haven't released yet, so we have some options:

- **Don't annotate**
  Requires C# >= 7 for anyone that includes NUnit source directly in their project.

- **Annotate simply**
  Requires C# >= 8 for anyone that includes NUnit source directly in their project.

- **Use preprocessor so that all signatures have an annotated and unannotated version**, a la https://github.com/xunit/assert.xunit/pull/36#issuecomment-572719747
  I'd take a look, but this sounds painful both for the initial implementation and also for just reading the code.
  Requires C# >= 7 for anyone that includes NUnit source directly in their project.
  Folks that include NUnit source directly in their project will need to define an extra symbol in order to opt in to nullability annotations.

jnm2:
> Another question to add to Mikkel's - I'm not too familiar with the effects of these changes - is there anything we should particularly be looking out for in review? The compiler should catch any problems really, right? 🤔

Right. The main thing to look out for is to try to prove me wrong every time I use `!`, just like you would if I added `Debug.Assert(x != null);`.

ChrisMaddock:
Hmm...XUnit facilitate consuming as source much more than we do, with the assertions separated out into a separate repo for submodule use, and the separate NuGet package. I'd say we should go for this as we have been doing, and anyone who is consuming as source needs to update.

Our primary userbase are consuming as a library, and this feature will be a big benefit to them. I don't think we should hold that back, for a situation we haven't committed to supporting.

ChrisMaddock:
> Right. The main thing to look out for is to try to prove me wrong every time I use !, just like you would if I added Debug.Assert(x != null);.

Thanks. 🙂 I'll let you into a secret, I'm behind the times, and don't even know what that syntax means! Guess I've got some reading to do. 🤓

jnm2:
@ChrisMaddock This is a 7-minute read with the null-forgiveness operator (unofficially the "dammit operator") `!` introduced a page and a half down: https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references

jnm2:
Oh, and my tl;dr: think of `!` as a more focused `#pragma warning disable`.

src/NUnitFramework/framework/Internal/Builders/PairwiseStrategy.cs
```diff
@@ -301,11 +302,11 @@ public bool IsTupleCovered( FeatureTuple tuple )
...
         internal class PairwiseTestCaseGenerator
         {
-            private FleaRand _prng;
+            private FleaRand? _prng;
 
-            private int[] _dimensions;
+            private int[]? _dimensions;
 
-            private List<FeatureTuple>[][] _uncoveredTuples;
+            private List<FeatureTuple>[][]? _uncoveredTuples;
```
ChrisMaddock, line 310: Ok - let me check I'm understanding this correctly. You've used the `!` operator in this class, because these vars are created as null, but they're actually only used/accessible when you call `GetTestCases`, which initialises them all? (Which arguably also makes this a bit of a code smell, and suggests these values should be passed around as parameters, instead?)

jnm2: Yes, exactly! The `!` makes it more obvious where the previously implicit assumptions are.

jnm2: That makes them a nice target for refactoring later. I learned the hard way to add `!` first to show the world as it has been before doing that though, just to make it possible for reviewers to see what triggered the refactoring.

ChrisMaddock: I definitely agree with that approach. 🙂

jnm2:
Ugh, pushed to the wrong PR branch. Reverted. Sorry for losing your review, @ChrisMaddock!

jnm2:
No changes, force pushed the same commit that was the tip before I accidentally pushed to the wrong branch.

What do you think about me actually dropping the final commit that I added before VS 16.5 was released? VS got smarter and fixed a lot of bugs in nullability warnings from 16.3-16.5. Also, I think 16.6 is around the corner.

ChrisMaddock:
Hmmm does the bug affect versions 16.0-16.4? I feel like it's work keeping in - it's fair for us to require people to be on the latest version, but I think anything we can do to ease the pain for people a minor version or two behind is valuable. I don't expect I'm on the latest version myself, at the moment! 😅 

One thing I didn't clock before, can we comment the reason for it in the code? Just so the explanation isn't lost in years down the line, when this looks like an unnatural thing to do. 🙂

jnm2:
Okay, I'll keep it in. There could be warnings that affect 16.0-16.3 that I never saw because the oldest version I had when I wrote this was 16.4.

I'll push a commit that documents the fixes the way I normally do.

jnm2:
> image

One day I will learn not to type push commands into a console window without looking at which tab I'm on...

mikkelbu:
A little late to the party, by I also agree that the following is the right approach
> **Annotate simply**
Requires C# >= 8 for anyone that includes NUnit source directly in their project.

jnm2:
Good, thanks for the input anyway.

PR #3494, opened 2020-03-07
.NET Core 2.2 and 3.0 are no longer supported and we should test 3.1

jnm2:
Table at https://dotnet.microsoft.com/platform/support/policy/dotnet-core#lifecycle

Relevant part:

| Version        | Original Release Date  | Support Level  | End of Support    |
|:---------------|:-----------------------|:---------------|:------------------|
| .NET Core 3.1  | December 3, 2019       | LTS            | December 3, 2022  |
| .NET Core 3.0  | September 23, 2019     | EOL            | March 3, 2020     |
| .NET Core 2.2  | December 4, 2018       | EOL            | December 23, 2019 |
| .NET Core 2.1  | May 30, 2018           | LTS            | August 21, 2021   |

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -45,12 +45,6 @@ jobs:
       packageType: runtime
       version: 2.1.x
 
-  - task: UseDotNet@2
-    displayName: 'Install .NET Core runtime 2.2'
-    inputs:
-      packageType: runtime
-      version: 2.2.x
-
   - bash: sudo apt-get install fsharp
     displayName: Install F# for Mono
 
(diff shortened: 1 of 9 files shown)


saxenark:
Does nunit latest version 3.12 support .net core 3.1?

jnm2:
@saxenark Yes, NUnit has supported every version of .NET Core as far as I'm aware. This pull request demonstrates NUnit running on .NET Core 3.1 in continuous integration on Windows, macOS, and Ubuntu.

Also, I believe .NET Core is still fairly aggressively backwards-compatible (though not as much as .NET Framework was). I expect current and past versions of NUnit to run on .NET 5, 6, 7, and so on, with only the most minor of road bumps if any.

Here's our docs page on how to use NUnit with .NET Core: https://github.com/nunit/docs/wiki/.NET-Core-and-.NET-Standard

PR #3495, opened 2020-03-07
Require test run id to be 2, per docs

jnm2:
https://github.com/nunit/docs/wiki/Test-Result-XML-Format#test-run

> - **Attributes:**
>   - **id** The unique ID of this test. For the test-run, it is hard-coded as "2".

### DIFF
--- a/src/NUnitFramework/framework/Schemas/TestResult.xsd
+++ b/src/NUnitFramework/framework/Schemas/TestResult.xsd
@@ -189,7 +189,13 @@
         <xs:element name="filter" type="TestFilterType" />
         <xs:group ref="ContainedTestGroup" minOccurs="0" maxOccurs="unbounded" />
       </xs:sequence>
-      <xs:attribute name="id" type="xs:string" use="required" />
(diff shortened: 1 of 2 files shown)


jnm2:
Found this old branch and thought I had never pushed it, but GitHub helpfully brought up Charlie's commit comments and now I know why this commit isn't in master.

CharliePoole:
Funny that nobody ever changed the docs. IMO the word "currently" added to the sentence would clarify it.

jnm2:
@CharliePoole Are you okay with the idea of removing the mention of the value "2" from the docs? The conclusions I imagine the reader could come to range from "this is a distraction" to "I'll depend on this."

CharliePoole:
That's probably the best idea.

jnm2:
Done: https://github.com/nunit/docs/wiki/Test-Result-XML-Format/_compare/427e2237c6de086eb5807339d517ceecfa03f047...1af10f1ab99fd8502edde30d13bb2dbdb9425eb4

PR #3496, opened 2020-03-07
Adding data dictionary should not add a trailing newline

jnm2:


jnm2:
Thanks! I noticed the formatting looked off in test runners compared to test results without the extra newline.

PR #3501, opened 2020-03-13
Issue 2477

### LINKED ISSUE #2477: Parameterized fixture with Explicit attribute can not be run when selected by name
Given code like this...

```C#
    [TestFixtureSource(nameof(GetParallelSuites)), Explicit]
    public class ParallelExecutionTests
    {
        ...
    }
```

Running the console with a filter `--where test==Namespace.Of.ParallelExecutionTests` causes no tests to run. I think that's because we are running the suite that includes all the instances of the fixture but the attribute is applied to each fixture.

AFAIR this works correctly in the case of methods, so we should look at how we do it there.

Workaround for the above, BTW, is to use `--where test=~Namespace.Of.ParallelExecutionTests`.

Poimen:
Resolves #2477

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
@@ -27,6 +27,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Reflection;
+using NUnit.Compatibility;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
 using NUnit.Framework.Internal.Builders;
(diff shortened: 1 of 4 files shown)


src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
```diff
@@ -123,8 +124,16 @@ public IEnumerable<TestSuite> BuildFrom(ITypeInfo typeInfo, IPreFilter filter)
...
+            fixtureSuite.ApplyAttributesToTest(typeInfo.Type.GetTypeInfo());
+
             foreach (ITestFixtureData parms in GetParametersFor(sourceType))
-                yield return _builder.BuildFrom(typeInfo, filter, parms);
+            {
+                TestSuite fixture = _builder.BuildFrom(typeInfo, filter, parms);
+                fixtureSuite.Add(fixture);
+            }
+
+            yield return fixtureSuite;
```
jnm2, line 136: Instead of doing this inside TestFixtureSourceAttribute, could you implement it in the caller of this method, `DefaultSuiteBuilder.BuildFrom`? I'm thinking that way the fix applies to all IFixtureBuilder attributes and not just TestFixtureSourceAttribute. I think this might automatically fix it for TestFixtureAttribute without changes too.

PR #3502, opened 2020-03-13
Unique Constraint failure message only outputs failed items

### LINKED ISSUE #3279: Improve failure message from UniqueItemsConstraint
Given the test below, then I expect that is is easy to see the item that is not unique.

```csharp
[TestFixture]
public class Class1
{
    [Test]
    public void Test()
    {
        int[] actual = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 12, 13 };
        Assert.That(actual, Is.Unique);
    }
}
```
But this gives
```
  Message:   Expected: all items unique
      But was:  < 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 12, 13 >
```

If the list is even longer then it can be that we cannot see the duplicated element
```csharp
[Test]
public void Test()
{
    int[] actual = { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 12, 13 };
    var longList = actual.ToList();
    longList.AddRange(Enumerable.Range(0, 20));
    Assert.That(longList, Is.Unique);
}
```
Gives 
```
  Message:   Expected: all items unique
      But was:  < 1, 2, 3, 4, 5, 6, 7, 8, 9, 10... >
```

stevenaw:
Fixes #3279 

This PR updates the constraint result of a unique check to only output the _distinct_ non-unique items from a collection, rather than the entire thing. This means that, no matter how many times a given value occurs, it will only appear once in the output (please let me know if this is desired).
It will output all different non-unique values, rather than just the first one found

### DIFF
--- a/src/NUnitFramework/framework/Constraints/MsgUtils.cs
+++ b/src/NUnitFramework/framework/Constraints/MsgUtils.cs
@@ -53,6 +53,11 @@ namespace NUnit.Framework.Constraints
     internal static class MsgUtils
     {
         /// <summary>
+        /// Default amount of items used by <see cref="FormatCollection"/> method.
+        /// </summary>
+        internal const int DefaultMaxItems = 10;
+
+        /// <summary>
         /// Static string used when strings are clipped
         /// </summary>
         private const string ELLIPSIS = "...";
@@ -96,7 +101,7 @@ namespace NUnit.Framework.Constraints
 
             AddFormatter(next => val => val is char ? string.Format(Fmt_Char, val) : next(val));
 
-            AddFormatter(next => val => val is IEnumerable ? FormatCollection((IEnumerable)val, 0, 10) : next(val));
+            AddFormatter(next => val => val is IEnumerable ? FormatCollection((IEnumerable)val) : next(val));
 
             AddFormatter(next => val => val is string ? FormatString((string)val) : next(val));
 
@@ -161,7 +166,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="collection">The collection containing elements to write.</param>
         /// <param name="start">The starting point of the elements to write</param>
         /// <param name="max">The maximum number of elements to write</param>
-        public static string FormatCollection(IEnumerable collection, long start, int max)
+        public static string FormatCollection(IEnumerable collection, long start = 0, int max = DefaultMaxItems)
         {
             int count = 0;
             int index = 0;
--- a/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
@@ -151,24 +151,21 @@ namespace NUnit.Framework.Constraints
 
         private static ICollection NonUniqueItemsInternal<T>(IEnumerable<T> actual, Func<T,T> hashValueFactory)
         {
-            var hash = new Dictionary<T, int>();
+            var hash = new Dictionary<T, bool>();
             var nonUniques = new List<T>();
 
             foreach (T item in actual)
             {
                 var itemToHash = hashValueFactory(item);
 
-                if (!hash.TryGetValue(itemToHash, out var itemCount))
+                if (!hash.TryGetValue(itemToHash, out var knownNonUnique))
                 {
-                    hash.Add(itemToHash, 1);
+                    hash.Add(itemToHash, false);
                 }
-                else
+                else if (!knownNonUnique)
                 {
-                    hash[itemToHash] = ++itemCount;
-                    if (itemCount == 2)
-                    {
-                        nonUniques.Add(item);
-                    }
+                    hash[itemToHash] = true;
+                    nonUniques.Add(item);
                 }
             }
 
--- a/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
@@ -21,6 +21,8 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+#nullable enable
+
 using System;
 using System.Collections;
 using System.Collections.Generic;
@@ -106,7 +108,7 @@ namespace NUnit.Framework.Constraints
             return nonUniques;
         }
 
-        private ICollection TryFastAlgorithm(IEnumerable actual)
+        private ICollection? TryFastAlgorithm(IEnumerable actual)
         {
             // If the user specified any external comparer with Using, exit
             if (UsingExternalComparer)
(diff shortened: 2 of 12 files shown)


src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -21,6 +21,8 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+#nullable enable
```
stevenaw: Because of all the other `#nullable` changes lately

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -91,58 +129,50 @@ private bool OriginalAlgorithm(IEnumerable actual)
...
+        private static ICollection StringsUniqueIgnoringCase(IEnumerable<string> actual)
+            => NonUniqueItemsInternal(actual, v => v.ToLower());
 
-            return true;
-        }
+        private static ICollection CharsUniqueIgnoringCase(IEnumerable<char> actual)
+            => NonUniqueItemsInternal(actual, char.ToLower);
 
-        private static bool CharsUniqueIgnoringCase(IEnumerable<char> actual)
+        private static ICollection NonUniqueItemsInternal<T>(IEnumerable<T> actual, Func<T,T> hashValueFactory)
```
stevenaw: We can't pass a `IEqualityComparer<T>` in here and use that when constructing the `Dictionary<T,int> below, as the linux tests seem to fail when uniquely detecting case insensitive strings. The only way I could get that to work was to first do a `ToLower()` and storing that in the dictionary (this was also what the previous version of the code did).

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -91,58 +129,47 @@ private bool OriginalAlgorithm(IEnumerable actual)
...
-            foreach (T item in actual)
-            {
-                if (!hash.Add(item))
-                    return false;
-            }
-
-            return true;
-        }
-
-        private static bool StringsUniqueIgnoringCase(IEnumerable<string> actual)
```
stevenaw: I ended up genericizing these

src/NUnitFramework/framework/Constraints/MsgUtils.cs
```diff
@@ -52,6 +52,11 @@ namespace NUnit.Framework.Constraints
     /// </summary>
     internal static class MsgUtils
     {
+        /// <summary>
+        /// Default amount of items used by <see cref="FormatCollection"/> method.
+        /// </summary>
+        private const int DefaultMaxItems = 10;
```
stevenaw, line 58: Changes to this file cherry-picked from #3505 via https://github.com/nunit/nunit/pull/3505/commits/dda80d6822f15909374e3cd9e0c11e54f66be9c5

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -52,33 +54,69 @@ public override string Description
...
+        {
+            IEnumerable enumerable = ConstraintUtils.RequireActual<IEnumerable>(actual, nameof(actual));
+
+            var nonUniqueItems = TryFastAlgorithm(enumerable) ?? OriginalAlgorithm(enumerable);
+
+            return new UniqueItemsContstraintResult(this, actual, nonUniqueItems);
         }
 
-        private bool OriginalAlgorithm(IEnumerable actual)
+        private ICollection OriginalAlgorithm(IEnumerable actual)
```
mikkelbu, line 71: I haven't tested this - having some computer problems 😄 - but couldn't this be simplified into ?
```csharp
        private ICollection OriginalAlgorithm(IEnumerable actual)
        {
            var processedItems = new List<object>();
            var nonUniques = new List<object>();

            foreach (object o1 in actual)
            {
                var isNonUnique = false;
                var knownNonUnique = false;

                foreach (object o2 in processedItems)
                {
                    if (ItemsEqual(o1, o2))
                    {
                        isNonUnique = true;
                        knownNonUnique = nonUniques.Where(o2 => ItemsEqual(o1, o2)).Any();
                        break;
                    }
                }

                if (!isNonUnique)
                    processedItems.Add(o1);
                else if (!knownNonUnique)
                    nonUniques.Add(o1);
            }

            return nonUniques;
        }
```

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
mikkelbu: I guess the `ToLower()` will give us problems in some locale, but this was the same issue before. Is the test failure the same kind as we corrected in #3416, meaning that they could be solved by annotating the tests?

src/NUnitFramework/tests/Constraints/UniqueItemsConstraintTests.cs
```diff
@@ -77,7 +93,16 @@ public void PerformanceTests(IEnumerable values, bool ignoreCase)
                     Assert.That(values, Is.Unique.IgnoreCase);
                 else
                     Assert.That(values, Is.Unique);
-            }, HelperConstraints.HasMaxTime(100));
+            }, HelperConstraints.HasMaxTime(150));
```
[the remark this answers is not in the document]
stevenaw: Thanks for spotting this! This was failing at one point, but I can't remember if it was on my machine or CI, or during which of a few approaches I've taken. I've tried reverting it and it now passes on my local machine. I'll try pushing the rest of the feedback and see how it goes

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -91,58 +129,46 @@ private bool OriginalAlgorithm(IEnumerable actual)
...
-            return true;
-        }
+        private static ICollection CharsUniqueIgnoringCase(IEnumerable<char> actual)
+            => NonUniqueItemsInternal(actual, char.ToLower);
 
-        private static bool CharsUniqueIgnoringCase(IEnumerable<char> actual)
+        private static ICollection NonUniqueItemsInternal<T>(IEnumerable<T> actual, Func<T,T> hashValueFactory)
         {
-            var hash = new HashSet<char>();
+            var hash = new HashSet<T>();
```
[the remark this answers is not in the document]
stevenaw: Updated!

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: That worked, thanks!

stevenaw:
Thanks for the suggestions @mikkelbu 
There's a few history issues I'm working through, plus it'll take a bit to ensure the builds work. Marking this as a WIP until it's sorted

src/NUnitFramework/tests/Constraints/UniqueItemsConstraintTests.cs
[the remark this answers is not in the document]
stevenaw: Ok, I can confirm that this was a CI thing. These tests pass in ~24 ms on my local machine, but around approaching the 100ms threshold on Azure (a recent failure took 101ms).

stevenaw:
I'm going to recreate this PR. GitHub is showing some diffs in files that I hadn't touched

Issue #3423, opened 2019-11-12
TestResult.cs casts ITestResult to TestResult


tommikiviniemi-srs:
Hi.

The iteration of the "IEnumerable< ITestResult > Children" member in TestResult.cs on line 407 is unnecessarily casting the ITestResult to TestResult, making it hard/impossible to keep extended children in the list.

Could the enumeration be fixed to iterate on the ITestResult interface, without the unnecessary casting? There are other places in the code base where similar unnecessary casts take place, but this one place is a real pain for me.

Thanks.

CharliePoole:
IIRC, we did a pass to fix this at one time, but probably only caught explicit casts, not the implied cast you point out. At some point, I believe we decided to require all results to derive from the TestResult abstract class, but I don't think we have been consistent about it. IMO we should either lose the interface and require the class or remove all these casts, which are confusing at best.

As for this change, we would probably want to search for uses of non-interface methods of TestResult and ensure we are not breaking anything.

tommikiviniemi-srs:
Thanks, that would be great. I derived my own classes from the interface directly, which is why I ran into this problem. The abstract class on the other hand has several methods marked as non-virtual so I can't extend from it properly either.

Thanks for consideration,
-Tommi

PR #3503, opened 2020-03-13
Remove implicit cast from ITestResult to TestResult

stevenaw:
Relates to #3423 

There was an implicit cast in `TestResult.AddToXml` that could cause exceptions to be thrown when used with custom-derived implementations of `ITestResult`. This is a short-term fix to avoid the crash, but doesn't address the main point from @CharliePoole here: https://github.com/nunit/nunit/issues/3423#issuecomment-552968884


ChrisMaddock:
> Personally, I also think we should remove the same unnecessary casts in the test projects (both explicit casts and implicit casts in connection with `foreach), just to minimise the confusion. What do @nunit/framework-team think?

Absolutely, in my opinion. (Although I'm no longer framework team!) 

Thanks for sorting this, @stevenaw!

PR #3505, opened 2020-03-15
Better failure messages for Subset and Superset constraints

Dreamescaper:
Provide extra/missing items to `CollectionSubsetConstraint` and `CollectionSupersetConstraint` failure messages.

E.g.
```
  Expected: subset of < 1, 2, 3, 4, 5 >
  But was:  < 1, 2, 2, 2, 5 >
  Extra items: < 2, 2 >        <------ ADDED
```

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionSupersetConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionSupersetConstraint.cs
@@ -21,8 +21,12 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+#nullable enable
+
 using System;
 using System.Collections;
+using System.Collections.Generic;
+using NUnit.Framework.Internal;
 
 namespace NUnit.Framework.Constraints
 {
@@ -33,6 +37,7 @@ namespace NUnit.Framework.Constraints
     public class CollectionSupersetConstraint : CollectionItemsEqualConstraint
     {
         private readonly IEnumerable _expected;
+        private List<object>? _missingItems;
 
         /// <summary>
         /// Construct a CollectionSupersetConstraint
@@ -72,7 +77,19 @@ namespace NUnit.Framework.Constraints
             CollectionTally tally = Tally(actual);
             tally.TryRemove(_expected);
 
-            return tally.Result.ExtraItems.Count == 0;
+            _missingItems = tally.Result.ExtraItems;
+
+            return _missingItems.Count == 0;
+        }
+
+        /// <summary>
+        /// Test whether the constraint is satisfied by a given value.
+        /// </summary>
+        public override ConstraintResult ApplyTo<TActual>(TActual actual)
+        {
+            IEnumerable enumerable = ConstraintUtils.RequireActual<IEnumerable>(actual, nameof(actual));
+            bool matches = Matches(enumerable);
+            return new CollectionSupersetConstraintResult(this, actual, matches, _missingItems);
         }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


src/NUnitFramework/framework/Constraints/MsgUtils.cs
```diff
@@ -161,7 +166,7 @@ public static string FormatValue(object val)
         /// <param name="collection">The collection containing elements to write.</param>
         /// <param name="start">The starting point of the elements to write</param>
         /// <param name="max">The maximum number of elements to write</param>
-        public static string FormatCollection(IEnumerable collection, long start, int max)
+        public static string FormatCollection(IEnumerable collection, long start = 0, int max = DefaultMaxItems)
```
Dreamescaper: Decided to add default parameter, since it is used in most cases with those arguments.
Let me know if there are any objections.

stevenaw: Personally, I really like this. It makes it easy to ensure future calls to this will always respect the defaults unless the author explicitly decides not to.
Is there any way you could expose just this change in a separate commit on your branch? I'd like to graft it onto my own branch for https://github.com/nunit/nunit/pull/3502 to help simplify things there

Dreamescaper: Updated PR to have separate commit for MsgUtils change. Still, not sure it's worth reusing it until this change is approved.

ChrisMaddock: I really like this change! Can I suggest updating all the places this method is currently used to make use of it too? The vast majority are currently passing in zero and ten.

src/NUnitFramework/framework/Constraints/CollectionSupersetConstraint.cs
```diff
@@ -72,7 +77,19 @@ protected override bool Matches(IEnumerable actual)
             CollectionTally tally = Tally(actual);
             tally.TryRemove(_expected);
 
-            return tally.Result.ExtraItems.Count == 0;
+            _missingItems = tally.Result.ExtraItems;
```
ChrisMaddock, line 80: Is this correct? It's a little confusing to read...maybe worth a comment? 🙂

src/NUnitFramework/framework/Constraints/MsgUtils.cs
Dreamescaper: Done!

src/NUnitFramework/framework/Constraints/CollectionSupersetConstraint.cs
mikkelbu: The reason is that we have 
```
CollectionTally tally = Tally(actual);
tally.TryRemove(_expected);
```
in this file, whereas we have 
```
CollectionTally tally = Tally(_expected);
tally.TryRemove(actual);
```
in CollectionSubsetConstraint.cs. I'm a bit tired, so I cannot determine if we can use `MissingItems` here and then just swap arund `_expected` and `actual` (I'm mostly worried about duplicates).

Dreamescaper: Added comment.

@mikkelbu 
Seems like no issues with duplicates, but it affect order of parameters custom comparer lambda - `.Using<int, string>((i, s) => ..)` VS `.Using<string, int>((s, i) => ..)`.
Currently this order differs between `CollectionSupersetConstraint` and `CollectionSubsetConstraint`.
While this looks like a bug, and would make sense to make it consistent, that would be a breaking change, so I'm not sure what would be better here.

CharliePoole:
@mikkelbu Regarding your concern about use of the word "set"... 

Much of this comes from our early implementation of `CollectionAssert`, which was intended to mimic the MsTest class of the same name. That class still contains the method `IsSubsetOf`, which they define as

```
Tests whether one collection is a subset of another collection and throws an exception if any element in the subset is not also in the superset
```

So it never had any reference to "sets" in the mathematical sense. Using "subset" and "superset" in this way is pretty common in normal language as well.

src/NUnitFramework/framework/Constraints/CollectionSupersetConstraint.cs
mikkelbu: If it will be a breaking change, then I think the current solution - with the comment - is the right one 👍

PR #3506, opened 2020-03-16
ValueTuple tests now running if not targeting NET35

stevenaw:
The value tuple tests were setup to only run for NET45. Seems like they pass for me in the tests as far back as NET40, due to the inclusion of `System.ValueTuple` as a test dependency. This PR switches up the conditional compilation so that they run `#if !NET35` instead of the previous `#if NET45`.

I'm unsure if it can be expanded further so that the ValueTuple + Tuple tests are run on NET35 too.


ChrisMaddock:
Good catch, @stevenaw! 🙂

PR #3510, opened 2020-03-22
issue 3508 Avoid to initialize Randomizer before InitialSeed has been set

### LINKED ISSUE #3508: Randomizer in NUnitTestCaseBuilder is always initialized with InitialSeed set to 0
Hi,

I found an issue where TestContext.CurrentContext.Random is not random between each session (always same list of numbers returned).
I investigated the issue and found that the Seed is always the same.
Here is my analysis:

When doing var runner =  new NUnitTestAssemblyRunner(new DefaultTestAssemblyBuilder());
runner.Load(assemblyPath, new ListDictionary());

Then TestContext.CurrentContext.Random is always giving the same list of numbers because same seed is used.
This is due because creation of object DefaultTestAssemblyBuilder is creating the NUnitTestCaseBuilder._randomizer before the InitialSeed being set

callstack of the randomizer creation:
NUnit.Framework.Internal.Randomizer.SeedGenerator : 69
NUnit.Framework.Internal.Randomizer.CreateRandomizer()
NUnit.Framework.Internal.Builders.NUnitTestCaseBuilder.NUnitTestCaseBuilder()
NUnit.Framework.Internal.Builders.DefaultTestCaseBuilder.DefaultTestCaseBuilder() 
NUnit.Framework.Internal.Builders.NUnitTestFixtureBuilder.NUnitTestFixtureBuilder() 
NUnit.Framework.Internal.Builders.DefaultSuiteBuilder.DefaultSuiteBuilder() 
NUnit.Framework.Api.DefaultTestAssemblyBuilder.DefaultTestAssemblyBuilder() 

So InitialSeed used is always 0

The InitialSeed is only set when doing the runner.Load with the InitialSeed possibly available in the settings.

I propose to only create the NUnitTestCaseBuilder._randomizer only when necessary : in NUnitTestCaseBuilder.BuildTestMethod

This way Load has been called first and InitialSeed has been set before building the test tree

fduvoux:
Fixes #3508

### DIFF
--- a/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
@@ -33,7 +33,16 @@ namespace NUnit.Framework.Internal.Builders
     /// </summary>
     public class NUnitTestCaseBuilder
     {
-        private readonly Randomizer _randomizer = Randomizer.CreateRandomizer();
+        private Randomizer? _randomizer = null;
+        private Randomizer Randomizer 
+        {
+            get
+            {
+                return _randomizer ??= Randomizer.CreateRandomizer();
+            }
+
+        }
+
         private readonly TestNameGenerator _nameGenerator;
 
         /// <summary>
@@ -55,7 +64,7 @@ namespace NUnit.Framework.Internal.Builders
         {
             var testMethod = new TestMethod(method, parentSuite)
             {
-                Seed = _randomizer.Next()
+                Seed = Randomizer.Next()
             };
 
             CheckTestMethodAttributes(testMethod);


ChrisMaddock:
Hi @fduvoux - thanks for digging in to this.

I'm sorry, I'm struggling to follow your logic, from what you posted in #3508. From what I can see, the initial seed is always set to something random in the static constructor of the Randomizer class, which will be called before `CreateRandomizer()` in the NUnitTestCaseBuilder class.

Can you help explain to me what I'm missing here?

fduvoux:
Ok I missed something, I was on a old version of NUnit3.x and this has been fixed since.
Sorry about that

ChrisMaddock:
No worries - thanks for digging into things anyway! 😊

PR #3516, opened 2020-03-29
Public API Analyzers

### LINKED ISSUE #2413: Snapshot testing NUnit's external API

jnm2:
Closes #2413
Supersedes https://github.com/nunit/nunit/pull/2652

This is an analyzer which is used by various .NET community libraries as well as in the various .NET Core and Roslyn repositories that ship libraries. It's a little more manual to use than the features we were discussing in #2413, but unless one of us has time to build those things from scratch very soon, this is better than having nothing.

If you add, modify, or remove one or more types or members that are visible to assemblies that reference NUnit, a warning shows up in the error list with a light bulb action prompting you to add a line to PublicAPI.Unshipped.txt. The 'Fix all in solution' option should always be used or else you'll have to do this action four times due to there being four separate API documents, one for each target framework. The warning causes the build to fail in CI if missed, and the changes to the PublicAPI.Unshipped.txt file will catch everyone's attention in PR reviews.

Instructions: https://github.com/dotnet/roslyn-analyzers/blob/master/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md

@rprouse: During each release, we will have to manually delete the entire contents of each of the eight Shipped/Unshipped files, then use the light bulb fix on the warnings in the error list to repopulate the Unshipped files, then cut the contents of each Unshipped file and paste 

Roslyn has [this unofficial script](https://github.com/dotnet/roslyn/blob/master/scripts/PublicApi/mark-shipped.ps1) to do some of this, but by looking at it, I'm pretty sure it won't work for lines where we are replacing a null-oblivious API with an annotated one.

@sharwell Can you confirm that the best way to handle releases as we incrementally annotate oblivious APIs is to delete the contents of all eight PublicAPI files, repopulate Unshipped, and cut and paste to Shipped? See https://github.com/nunit/nunit/commit/44c9dc61d384bf3eb985f6b7d5e690651dd44d27 for example. There are no `*REMOVE*` lines for them, and those are not new APIs but rather existing APIs that are now annotated. We can't just paste the lines from Unshipped to Shipped, even assuming we have no `*REMOVE*` lines to deal with.


src/NUnitFramework/framework/PublicAPI/net35/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,303 @@
+*REMOVED*NUnit.Framework.Constraints.NUnitEqualityComparer.AreEqual(object x, object y, ref NUnit.Framework.Constraints.Tolerance tolerance, bool topLevelComparison = true) -> bool
```
jnm2: This is no good. We will almost certainly break someone. We need to revert this before our next release.

stevenaw: This was probably me in #3406
I had to update the signature, and since I've had issues with unexpectedly breaking changes of default method args in the past, I thought it safest at the time to split that function up into two overloads.

Sorry, I hadn't expected it to be considered breaking

jnm2: No, sorry on my end... I wasn't looking at what we replaced it with. I don't think this is either a binary break or a source break; we made the parameter no longer be a default parameter, which is a source break but not a binary break, and then we added an overload to avoid the source break. The subtlety is that you've switched overloads on recompile, but in this case that's exactly what we want anyway.

So we're good!

rprouse:
@jnm2 just letting you know I've seen this but I don't have time to deal with it right now. let's discuss...

jnm2:
No problem. I'm working through my OSS backlog and this was on it, but let me know what you want to do. If it's not something we want, I'm just as happy to cross it off without implementing it. Did "let's discuss" mean here?

ChrisMaddock:
I always worried about the "entry-barrier" for new contributors to adding in something like this, but I think I've now come to the conclusion that this is the lesser evil, and this would be really helpful for all the problems we face on releases.

I'll leave you guys to work out what's best for the framework, but I'd like to put this in for the nunit.engine.api assembly at some point. 🙂 Thanks for the example, if nothing else, Joseph!

CharliePoole:
This is a great idea. I'm going to use it in my own stuff! I was always concerned about the more hard-coded approaches, because they are harder to adjust according to specific project goals. This approach seems to do more by doing less along with being configurable. And the configuration can be used as documentation for what kinds of changes a project promises to avoid.

sharwell:
> the best way to handle releases as we incrementally annotate oblivious APIs is to delete the contents of all eight PublicAPI files, repopulate Unshipped, and cut and paste to Shipped

No. The code fix for annotating oblivious APIs will apply those changes directly to the shipped file.

sharwell:
I really wish the API files used a more friendly syntax but I just haven't had the chance to implement that yet. Fixing the syntax isn't too hard but keeping all the code fixes working is.

rprouse (dismissed):
I've added your note about the release to our release documents. Next time I do the release I'll fill in any additional details that are needed.

Thanks for doing this...

rprouse:
@jnm2 I've approved. I assume you will clean up and get the builds passing?

jnm2:
If I merge with master locally and run `.\build.ps1 -t build`, everything is fine. When AppVeyor runs the same build target, it shows hundreds of build failures of the form:

> error RS0017: Symbol '~static NUnit.Framework.Assume.That<TActual>(NUnit.Framework.Constraints.ActualValueDelegate<TActual> del, NUnit.Framework.Constraints.IResolveConstraint expr, System.Func<string> getExceptionMessage) -> void' is part of the declared API, but is either not public or could not be found

jnm2:
If I merge with master locally, run `git clean -xdf`, and run `.\build.ps1 -t build`, everything is fine. When AppVeyor runs the same build target, it shows hundreds of build failures of the form:

> error RS0017: Symbol '~static NUnit.Framework.Assume.That<TActual>(NUnit.Framework.Constraints.ActualValueDelegate<TActual> del, NUnit.Framework.Constraints.IResolveConstraint expr, System.Func<string> getExceptionMessage) -> void' is part of the declared API, but is either not public or could not be found

@sharwell Is this something you've seen in the past?

sharwell:
I'd try putting this in the code to see if it's caused by a version difference:

```
#error version
```

jnm2:
I'm so embarrassed... I'm on the wrong PR branch. Sorry Sam!

jnm2:
@sharwell Now what I have is no errors or warnings in VS, but tons of RS0017 when building from the CLI.

VS: Compiler version: '3.6.0-3.20218.3 (37ad245f)'. Language version: 8.0.	
CLI: Compiler version: '3.5.0-beta4-20153-05 (20b9af91)'. Language version: 8.0.

jnm2:
@sharwell Okay, fixed that by switching to VS 16.5.4 instead of the latest 16.6 preview. Is this known or should I report somewhere with a repro?

rprouse:
@jnm2 want to push a null-op to kick off new builds on the latest VS images and see what happens?

OsirisTerje:
@jnm2 What is the current state of this?

PR #3523, opened 2020-04-12
Unique Constraint failure message only outputs failed items

### LINKED ISSUE #3279: Improve failure message from UniqueItemsConstraint

stevenaw:
Remake of #3502 
Fixes #3279 

This PR updates the constraint result of a unique check to only output the _distinct_ non-unique items from a collection, rather than the entire thing. This means that, no matter how many times a given value occurs, it will only appear once in the output (**_please let me know if this is desired_**).
It will output all different non-unique values, rather than just the first one found

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
@@ -37,7 +37,7 @@ namespace NUnit.Framework.Constraints
         /// <summary>
         /// The NUnitEqualityComparer in use for this constraint
         /// </summary>
-        private readonly NUnitEqualityComparer _comparer = new NUnitEqualityComparer();
+        protected internal NUnitEqualityComparer Comparer { get; } = new NUnitEqualityComparer();
 
         /// <summary>
         /// Construct an empty CollectionConstraint
@@ -57,7 +57,7 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         protected bool IgnoringCase
         {
-            get { return _comparer.IgnoreCase; }
+            get { return Comparer.IgnoreCase; }
         }
 
         /// <summary>
@@ -65,7 +65,7 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         protected bool UsingExternalComparer
         {
-            get { return _comparer.ExternalComparers.Count > 0; }
+            get { return Comparer.ExternalComparers.Count > 0; }
         }
 
         #endregion
@@ -79,7 +79,7 @@ namespace NUnit.Framework.Constraints
         {
             get
             {
-                _comparer.IgnoreCase = true;
+                Comparer.IgnoreCase = true;
                 return this;
             }
         }
@@ -90,7 +90,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="comparer">The IComparer object to use.</param>
         public CollectionItemsEqualConstraint Using(IComparer comparer)
         {
-            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            Comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
@@ -100,7 +100,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="comparer">The IComparer object to use.</param>
         public CollectionItemsEqualConstraint Using<T>(IComparer<T> comparer)
         {
-            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            Comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
@@ -110,7 +110,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="comparison">The Comparison object to use.</param>
         public CollectionItemsEqualConstraint Using<T>(Comparison<T> comparison)
         {
-            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparison));
+            Comparer.ExternalComparers.Add(EqualityAdapter.For(comparison));
             return this;
         }
 
@@ -120,7 +120,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="comparer">The IComparer object to use.</param>
         public CollectionItemsEqualConstraint Using(IEqualityComparer comparer)
         {
-            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            Comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
@@ -130,7 +130,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="comparer">The IComparer object to use.</param>
         public CollectionItemsEqualConstraint Using<T>(IEqualityComparer<T> comparer)
         {
-            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            Comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
@@ -140,13 +140,13 @@ namespace NUnit.Framework.Constraints
         /// <param name="comparer">The supplied boolean-returning delegate to use.</param>
         public CollectionItemsEqualConstraint Using<T>(Func<T, T, bool> comparer)
         {
-            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            Comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
         internal CollectionItemsEqualConstraint Using(EqualityAdapter adapter)
         {
-            _comparer.ExternalComparers.Add(adapter);
+            Comparer.ExternalComparers.Add(adapter);
             return this;
         }
 
@@ -158,7 +158,7 @@ namespace NUnit.Framework.Constraints
         protected bool ItemsEqual(object x, object y)
         {
             Tolerance tolerance = Tolerance.Default;
-            return _comparer.AreEqual(x, y, ref tolerance);
+            return Comparer.AreEqual(x, y, ref tolerance);
         }
 
         /// <summary>
@@ -167,7 +167,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="c">The collection to be included in the tally</param>
         protected CollectionTally Tally(IEnumerable c)
         {
-            return new CollectionTally(_comparer, c);
+            return new CollectionTally(Comparer, c);
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
@@ -37,7 +37,7 @@ namespace NUnit.Framework.Constraints
         /// <summary>
         /// The NUnitEqualityComparer in use for this constraint
         /// </summary>
-        protected internal NUnitEqualityComparer Comparer { get; } = new NUnitEqualityComparer();
+        private protected NUnitEqualityComparer Comparer { get; } = new NUnitEqualityComparer();
 
         /// <summary>
         /// Construct an empty CollectionConstraint
--- a/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
@@ -37,7 +37,7 @@ namespace NUnit.Framework.Constraints
         /// <summary>
         /// The NUnitEqualityComparer in use for this constraint
         /// </summary>
-        private protected NUnitEqualityComparer Comparer { get; } = new NUnitEqualityComparer();
+        private readonly NUnitEqualityComparer _comparer = new NUnitEqualityComparer();
 
         /// <summary>
         /// Construct an empty CollectionConstraint
@@ -57,7 +57,7 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         protected bool IgnoringCase
         {
-            get { return Comparer.IgnoreCase; }
+            get { return _comparer.IgnoreCase; }
         }
 
         /// <summary>
@@ -65,7 +65,7 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         protected bool UsingExternalComparer
         {
-            get { return Comparer.ExternalComparers.Count > 0; }
+            get { return _comparer.ExternalComparers.Count > 0; }
         }
 
         #endregion
@@ -79,7 +79,7 @@ namespace NUnit.Framework.Constraints
         {
             get
             {
-                Comparer.IgnoreCase = true;
+                _comparer.IgnoreCase = true;
                 return this;
             }
         }
@@ -90,7 +90,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="comparer">The IComparer object to use.</param>
         public CollectionItemsEqualConstraint Using(IComparer comparer)
         {
-            Comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
@@ -100,7 +100,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="comparer">The IComparer object to use.</param>
         public CollectionItemsEqualConstraint Using<T>(IComparer<T> comparer)
         {
-            Comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
@@ -110,7 +110,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="comparison">The Comparison object to use.</param>
         public CollectionItemsEqualConstraint Using<T>(Comparison<T> comparison)
         {
-            Comparer.ExternalComparers.Add(EqualityAdapter.For(comparison));
+            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparison));
             return this;
         }
 
@@ -120,7 +120,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="comparer">The IComparer object to use.</param>
         public CollectionItemsEqualConstraint Using(IEqualityComparer comparer)
         {
-            Comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
@@ -130,7 +130,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="comparer">The IComparer object to use.</param>
         public CollectionItemsEqualConstraint Using<T>(IEqualityComparer<T> comparer)
         {
-            Comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
@@ -140,13 +140,13 @@ namespace NUnit.Framework.Constraints
         /// <param name="comparer">The supplied boolean-returning delegate to use.</param>
         public CollectionItemsEqualConstraint Using<T>(Func<T, T, bool> comparer)
         {
-            Comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
+            _comparer.ExternalComparers.Add(EqualityAdapter.For(comparer));
             return this;
         }
 
         internal CollectionItemsEqualConstraint Using(EqualityAdapter adapter)
         {
-            Comparer.ExternalComparers.Add(adapter);
+            _comparer.ExternalComparers.Add(adapter);
             return this;
         }
 
@@ -158,7 +158,7 @@ namespace NUnit.Framework.Constraints
         protected bool ItemsEqual(object x, object y)
         {
             Tolerance tolerance = Tolerance.Default;
-            return Comparer.AreEqual(x, y, ref tolerance);
+            return _comparer.AreEqual(x, y, ref tolerance);
         }
 
         /// <summary>
@@ -167,7 +167,7 @@ namespace NUnit.Framework.Constraints
         /// <param name="c">The collection to be included in the tally</param>
         protected CollectionTally Tally(IEnumerable c)
         {
-            return new CollectionTally(Comparer, c);
+            return new CollectionTally(_comparer, c);
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
@@ -45,12 +45,9 @@ namespace NUnit.Framework.Constraints.Comparers
             string xString = (string)x;
             string yString = (string)y;
 
-            bool caseInsensitive = _equalityComparer.IgnoreCase;
+            var comparison = _equalityComparer.IgnoreCase ? StringComparison.CurrentCulture : StringComparison.CurrentCultureIgnoreCase;
 
-            string s1 = caseInsensitive ? xString.ToLower() : xString;
-            string s2 = caseInsensitive ? yString.ToLower() : yString;
-
-            return s1.Equals(s2);
+            return string.Equals(xString, yString, comparison);
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
@@ -45,7 +45,7 @@ namespace NUnit.Framework.Constraints.Comparers
             string xString = (string)x;
             string yString = (string)y;
 
-            var comparison = _equalityComparer.IgnoreCase ? StringComparison.CurrentCulture : StringComparison.CurrentCultureIgnoreCase;
+            var comparison = _equalityComparer.IgnoreCase ? StringComparison.CurrentCultureIgnoreCase : StringComparison.CurrentCulture;
 
             return string.Equals(xString, yString, comparison);
         }
--- a/src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
@@ -45,9 +45,12 @@ namespace NUnit.Framework.Constraints.Comparers
             string xString = (string)x;
             string yString = (string)y;
 
-            var comparison = _equalityComparer.IgnoreCase ? StringComparison.CurrentCultureIgnoreCase : StringComparison.CurrentCulture;
+            bool caseInsensitive = _equalityComparer.IgnoreCase;
 
-            return string.Equals(xString, yString, comparison);
+            string s1 = caseInsensitive ? xString.ToLower() : xString;
+            string s2 = caseInsensitive ? yString.ToLower() : yString;
+
+            return s1.Equals(s2);
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
@@ -151,24 +151,21 @@ namespace NUnit.Framework.Constraints
 
         private static ICollection NonUniqueItemsInternal<T>(IEnumerable<T> actual, Func<T,T> hashValueFactory)
         {
-            var hash = new Dictionary<T, int>();
+            var hash = new Dictionary<T, bool>();
             var nonUniques = new List<T>();
 
             foreach (T item in actual)
             {
                 var itemToHash = hashValueFactory(item);
 
-                if (!hash.TryGetValue(itemToHash, out var itemCount))
+                if (!hash.TryGetValue(itemToHash, out var knownNonUnique))
                 {
-                    hash.Add(itemToHash, 1);
+                    hash.Add(itemToHash, false);
                 }
-                else
+                else if (!knownNonUnique)
                 {
-                    hash[itemToHash] = ++itemCount;
-                    if (itemCount == 2)
-                    {
-                        nonUniques.Add(item);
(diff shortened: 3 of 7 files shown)


src/NUnitFramework/tests/Constraints/UniqueItemsConstraintTests.cs
```diff
@@ -77,7 +93,17 @@ public void PerformanceTests(IEnumerable values, bool ignoreCase)
                     Assert.That(values, Is.Unique.IgnoreCase);
                 else
                     Assert.That(values, Is.Unique);
-            }, HelperConstraints.HasMaxTime(100));
+            }, HelperConstraints.HasMaxTime(150));
```
stevenaw, line 96: As noted in https://github.com/nunit/nunit/pull/3502#discussion_r405096102
This was exceeding the time on CI (on my machine it only took ~24ms). It's worth noting that, with this PR, we now have to process the entire collection is order to know all non-unique values. Previously, it would failfast on the first non-unique.

src/NUnitFramework/tests/Constraints/UniqueItemsConstraintTests.cs
```diff
@@ -41,13 +43,19 @@ public void SetUp()
...
 
         static object[] SuccessData = new object[] { new int[] { 1, 3, 17, -2, 34 }, new object[0] };
-        static object[] FailureData = new object[] { new object[] { new int[] { 1, 3, 17, 3, 34 }, "< 1, 3, 17, 3, 34 >" } };
+        static object[] FailureData = new object[] { new object[] {
+            new int[] { 1, 3, 17, 3, 34 }, 
+            "< 1, 3, 17, 3, 34 >" + Environment.NewLine + "  Not unique items: < 3 >" }
+        };
 
-        [Test]
+        [Test, SetCulture("en-US")]
```
stevenaw, line 51: As noted in https://github.com/nunit/nunit/pull/3502#discussion_r405144602
The class has been updated to use a `IEqualityComparer<T>`:
https://github.com/nunit/nunit/pull/3523/files#diff-f6b841c6c7eed9838e33cf9082f26944R147

Which caused some funny behaviour on Azure Linux (which uses the `en-POSIX-us` culture). See this for more info: https://github.com/nunit/nunit/pull/3416

src/NUnitFramework/tests/Constraints/UniqueItemsConstraintTests.cs
```diff
@@ -21,16 +21,18 @@
...
 using System.Linq;
 using NUnit.TestUtilities;
 using NUnit.TestUtilities.Collections;
+using static NUnit.Framework.Constraints.UniqueItemsConstraint;
 
 namespace NUnit.Framework.Constraints
 {
     [TestFixture]
-    public class UniqueItemsTests : ConstraintTestBase
+    public class UniqueItemsConstraintTests : ConstraintTestBase
```
stevenaw, line 35: Updated class name to match file name

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -52,97 +55,112 @@ public override string Description
...
-                if (!hash.Add(item))
-                    return false;
-            }
-
-            return true;
-        }
+        private static ICollection<T> ItemsUnique<T>(IEnumerable<T> actual)
+            => NonUniqueItemsInternal(actual, EqualityComparer<T>.Default);
 
-        private static bool StringsUniqueIgnoringCase(IEnumerable<string> actual)
```
stevenaw, line 118: These have all been genericized

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -174,5 +192,39 @@ private Type GetGenericTypeArgument(IEnumerable actual)
 
             return null;
         }
+
+        private sealed class CaseInsensitiveCharComparer : IEqualityComparer<char>
+        {
```
stevenaw, line 197: I can't find a way to do a case-insensitive culture-aware char comparison. I'd been hoping to avoid converting each char in the collection to a 1-length string

mikkelbu: @jnm2 Do you have input on this. You usually have a good grab regarding case-insensitive comparison and best practises

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -52,97 +55,112 @@ public override string Description
...
+                var unknownNonUnique = false;
+
+                foreach (var o2 in processedItems)
+                {
                     if (ItemsEqual(o1, o2))
-                        return false;
-                list.Add(o1);
+                    {
+                        isUnique = false;
+                        unknownNonUnique = !nonUniques.Any(o2 => ItemsEqual(o1, o2));
```
ChrisMaddock, line 87: I suspect this will be contributing to the performance test needing tweaking. Here, you're checking the entire list of every non unique item for a match, for every new item you find. Why not just add them all to a list, and then `.Distinct()` at the end? 🙂

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -174,5 +192,39 @@ private Type GetGenericTypeArgument(IEnumerable actual)
...
+                return char.ToLower(x) == char.ToLower(y);
+            }
+
+            public int GetHashCode(char obj)
+            {
+                return char.ToLower(obj).GetHashCode();
+            }
+        }
+
+        internal sealed class UniqueItemsContstraintResult : ConstraintResult
```
ChrisMaddock, line 209:
```suggestion
        internal sealed class UniqueItemsConstraintResult : ConstraintResult
```

Small typo. 🙂 Good thinking making it internal sealed though!

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -174,5 +192,39 @@ private Type GetGenericTypeArgument(IEnumerable actual)
...
+            {
+                NonUniqueItems = nonUniqueItems;
+            }
+
+            public override void WriteAdditionalLinesTo(MessageWriter writer)
+            {
+                if (this.Status == ConstraintStatus.Failure)
+                {
+                    writer.Write("  Not unique items: ");
+                    var output = MsgUtils.FormatCollection(NonUniqueItems);
```
ChrisMaddock, line 224:
```suggestion
                    var output = MsgUtils.FormatCollection(NonUniqueItems, 10);
```

Another performance suggestion - at the moment, you're building a list of all the non-unique items in a collection, then only displaying the first 10. (That's the default maximum amount shown by `FormatCollection`, from memory.) How about setting that value from the constraint - which them means in `NonUniqueItemsInternal` you can just find the first 10 non-unique items, and then stop looking. That will help in cases where you have lots of non-unique items, like the performance test in question!

src/NUnitFramework/tests/Constraints/UniqueItemsConstraintTests.cs
ChrisMaddock: Made a couple of suggestions on this above. From memory - we added these tests as we accidentally made this too slow a few years ago - it would be great if we can avoid slowing things down as much as possible with this change. 🙂

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -52,97 +55,112 @@ public override string Description
...
-            return true;
-        }
+        private static ICollection<T> ItemsUnique<T>(IEnumerable<T> actual)
+            => NonUniqueItemsInternal(actual, EqualityComparer<T>.Default);
 
-        private static bool StringsUniqueIgnoringCase(IEnumerable<string> actual)
-        {
-            var hash = new HashSet<string>();
+        private static ICollection<string> StringsUniqueIgnoringCase(IEnumerable<string> actual)
+            => NonUniqueItemsInternal(actual, StringComparer.CurrentCultureIgnoreCase);
```
ChrisMaddock, line 140: Nice - our string comparer wasn't very culture friendly before! I think this is a separate bug fixed here, two for the price of one!

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
ChrisMaddock: Or - taking inspiration from the version before, add them to a Hashmap. On false, add the non-unique items to a second hashmap, and check the size of that second hash map. End the method whenever you've run out of items to check, or found the first 10 non-unique items - whichever comes first?

stevenaw: Thanks @ChrisMaddock ! Things are a bit busy for me right now, but I'll slowly try and poke away at this over the next little bit.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: thanks!

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: Turns out I had a bit more time than expected. 😄 

I tried using `HashSet<T>` or `Distinct<T>()` here, but ran into issues because this path deals with NUnitEqualityComparer (including the scenario of having user-defined comparers or equality between different types, like `42` and `42f`). Trying to wrap `ItemsEqual` into a `IEqualityComparer<object>` didn't seem to pan out perfectly, since objects with equality (like `42` and `42f`) had different default hash codes and often ended up in different hash buckets. This meant "equal" objects weren't flagged, and uniques ended up getting missed.

I think I did make some progress by adding a fastpath for when all the items implement `IComparable`, which let me then use a `SortedSet<T>`, resulting in O(n log n) instead of the original O(n^2). The performance test cases now pass 😁

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: Thanks! I hadn't considered failing fast after 10 non-uniques. I've added this in to all the places in the class :)

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
```diff
@@ -37,7 +37,7 @@ public abstract class CollectionItemsEqualConstraint : CollectionConstraint
         /// <summary>
         /// The NUnitEqualityComparer in use for this constraint
         /// </summary>
-        private readonly NUnitEqualityComparer _comparer = new NUnitEqualityComparer();
+        protected internal NUnitEqualityComparer Comparer { get; } = new NUnitEqualityComparer();
```
stevenaw, line 40: In order to wrap this in an `IComparer` implementation in a child class (UniqueItemsConstraint)

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: No problem! To be honest though, I'm not the best at culture stuff. I went with CurrentCulture here since I think that matches the existing behaviour of ToLower(), but I'm unsure if we should instead be using InvariantCulture or Ordinal for the ignore case.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -52,97 +56,172 @@ public override string Description
...
+            var nonUniqueItems = GetNonUniqueItems(enumerable);
+
+            return new UniqueItemsConstraintResult(this, actual, nonUniqueItems);
         }
 
-        private bool OriginalAlgorithm(IEnumerable actual)
+        private ICollection OriginalAlgorithm(IEnumerable actual)
         {
-            var list = new List<object>();
+#if !NET35
```
stevenaw, line 75: The `IComparable` fastpath uses a `SortedList<T>`, which doesn't exist in NET35

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
jnm2: `StringComparer.CurrentCultureIgnoreCase`?

jnm2: So here's the problem: culture-sensitive comparisons are broken when you make them operate on individual chars. How is `"ß".Equals("ss", StringComparison.CurrentCulture)` being true going to be reflected when you aren't passing the whole string for the comparison?

Can we get out of comparing a char at a time?

src/NUnitFramework/tests/Constraints/UniqueItemsConstraintTests.cs
jnm2: I think `SetCulture("")` sets it to invariant, but I forget.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
jnm2: In situations where operating on individual paired chars makes sense (i.e. where `x.ToString().Equals(y.ToString(), ...)` would be a valid thing to do), .NET Core 2.1 adds an Equals extension method on ReadOnlySpan<char> instances with StringComparison. That lets you avoid the .ToString() allocation. But as mentioned above, operating on individual chars is probably not a valid way to be doing any kind of linguistic comparison.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
jnm2: If the items implement IComparable, how do you know that the implementation is consistent with the custom equality comparer that was set in the assertion?

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
jnm2: One last twist is that `char` in .NET doesn't represent what you'd think of as a character. It's UTF-16, it either represents one "character" or half of a "character" (surrogate pairs).

src/NUnitFramework/tests/Constraints/UniqueItemsConstraintTests.cs
jnm2: Similar issues on slowness: https://github.com/nunit/nunit/issues/2494 (fixed), https://github.com/nunit/nunit/issues/2799 (not fixed)

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: Thanks @jnm2 !

For the IComparable path, `ItemEquals` is used to check if the items are equal first (which will use NUnitEqualityComparer, including any "ExternalComparers") and if they are unequal, _then_ try to apply some sort of sort based on the IComparable. I think I recently added a string-specific case to handle if `IgnoreCase` is specified, though I should update that to fall back to string checking _after_ the call to the NUnitEqualityComparer:

https://github.com/nunit/nunit/pull/3523/files#diff-f6b841c6c7eed9838e33cf9082f26944R268-R271

Sorry, I'm unsure if this answers your question

stevenaw: On the note of checking for distincts while processing or at the end, I did a quick benchmark using int arrays:

<pre>
|                  Method |     N |         Mean |       Error |      StdDev | Ratio | RatioSD |
|------------------------ |------ |-------------:|------------:|------------:|------:|--------:|
|                  UseSet |    10 |     314.2 ns |     2.22 ns |     2.08 ns |  1.00 |    0.00 |
| UseHashSet_PostDistinct |    10 |     373.6 ns |     6.36 ns |    12.10 ns |  1.18 |    0.05 |
|                         |       |              |             |             |       |         |
|                  UseSet |   100 |   2,562.1 ns |    12.98 ns |    10.84 ns |  1.00 |    0.00 |
| UseHashSet_PostDistinct |   100 |   2,886.5 ns |    56.10 ns |    62.36 ns |  1.13 |    0.02 |
|                         |       |              |             |             |       |         |
|                  UseSet | 10000 | 309,340.6 ns | 2,704.61 ns | 2,397.57 ns |  1.00 |    0.00 |
| UseHashSet_PostDistinct | 10000 | 326,946.5 ns | 6,145.89 ns | 6,036.08 ns |  1.06 |    0.02 |
</pre>

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
jnm2: Can this be `private protected`?

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
jnm2: Ah, I follow now. Thanks! For what types does IComparable/SortedSet lend a solution that IEquatable/HashSet doesn't? `42` vs `42f` should be a problem for both, right?

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: Ah ok, thanks for explaining! I must admit, I don't have much experience with internationalization at application level. I'll push a few updates over the next little bit to try and make this more robust

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: It has less to do with the types themselves and most to do with the containing data structure. Based on the pre-PR state of the constraint, a HashSet-based "fast-path" was used if:
- No "ExternalComparer" was specified
- A generic container (ex: List<T>) was passed
- typeof(T) is sealed
- T is not a type with special handling by NUnit (list here: https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs#L153-L158)

So if a collection of type `List<object>` or `IList` were passed, it would have to go down the "slow path", regardless of containing data (even if they were all the same data type, like `int`). The IComparable/SortedSet was my attempt to make this "slow path" a little more intelligent by checking if the data is sortable (from which it becomes much simpler to identify duplicate data, even if the container has multiple different data types).

I'm still not sure if this is the absolute best way to have approached it, so I'd welcome any feedback or suggestions on it. :)

stevenaw: As part of my latest push, I've added tests for this "medium path", which uses a container which only implements the non-generic `ICollection`

src/NUnitFramework/tests/Constraints/UniqueItemsConstraintTests.cs
stevenaw: I just tested this, and it works. Thanks!

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
jnm2: What are the real-world cases in particular that this optimization is addressing? I wouldn't have expected List<object> containing all elements of the same type to come up often. Is it mainly about `typeof(T)` being sealed?

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
ChrisMaddock: CurrentCulture is the right choice here - we need to use CurrentCulture, as users can set the culture on any individual test themselves with the `SetCulture` attribute. 🙂

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: Sorry, yes, I oversimplified with my example of `List<object>`.

One (pre-existing) condition which will cause the fastpath to not be taken is if `typeof(T)` is unsealed. So if one passes a `List<T>` where T is unsealed but `typeof(T)` implements IComparable (or all contained items derived from T happen to implement IComparable), then this new "medium path" would be taken instead. Similarly, if a non-generic collection like `IList` or `ICollection` were passed, but all items implement IComparable, it would take the medium path.

That last case was the important one for me, as my day job involves maintaining a codebase originating from the .NET20 days, with many collections of type `List<object>` (originally these were  `IList`, but they were updated to `List<object>` during a past upgrade since both enumerate `object`s).

I know it's a bit ugly to have used the NET35 conditional compilation. If you think this optimization is safe and makes sense to keep, I'm happy to try a few different other implementations.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: Oh right, of course! :)

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
stevenaw: Oh, sorry. I just checked MSDN. I had thought this would available to types that were both derived and within the assembly (rather than either). I'll update this next chance I get.

src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
```diff
@@ -45,12 +45,9 @@ internal StringsComparer(NUnitEqualityComparer equalityComparer)
...
             string yString = (string)y;
 
-            bool caseInsensitive = _equalityComparer.IgnoreCase;
+            var comparison = _equalityComparer.IgnoreCase ? StringComparison.CurrentCulture : StringComparison.CurrentCultureIgnoreCase;
 
-            string s1 = caseInsensitive ? xString.ToLower() : xString;
-            string s2 = caseInsensitive ? yString.ToLower() : yString;
-
-            return s1.Equals(s2);
+            return string.Equals(xString, yString, comparison);
```
stevenaw, line 50: So that `"ß".Equals("ss", StringComparison.CurrentCulture) == true`

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: I've done some digging and it seems like I've run into an existing bug in that the `StringsComparer` is [not culture-safe](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs#L48-L53), as demonstrated by the below:

```csharp
[Test, SetCulture("de-DE")]
        public void CultureAware()
        {
            string x = "ss";
            string y = "ß";

            Assert.IsTrue(string.Equals(x, y, StringComparison.CurrentCulture));  // passes
            Assert.AreEqual(x, y);   // fails
        }
```

This same issue is causing failures for the medium/slow paths when the framework has to fall back to using the NUnitEqualityComparer to check for uniques:
https://github.com/nunit/nunit/pull/3523/files#diff-a31a83d72a32eb0e9538e5ca251a3876R100-R113

I can fix this easy enough by updating StringsComparer to use `CurrentCulture/CurrentCultureIgnoreCase`, but it seems we have an existing test which relies on using an OrdinalComparison. This test will fail if run using either `CurrentCulture/CurrentCultureIgnoreCase`:
https://github.com/nunit/nunit/blob/ad943eb/src/NUnitFramework/tests/Assertions/StringAssertTests.cs#L155-L162

What would the best approach here be?

ChrisMaddock: Eesh, this has turned into a can of worms, hasn't it! What do you think @jnm2 - 'fix' it to be culture sensitive, and document the change? At least that would be more consistent across the library.

That would mean to specifically compare for different encodings, you'd have to specifically put in an Ordinal string comparer. That seems sensible, but is of course a change in behaviour...

stevenaw: Definitely a good exercise in corner cases. 😄 

While digging a bit deeper, I've also noticed that the "different encoding" behaviour itself is a bit inconsistent on the current `master` branch as the test only passes when `IgnoreCase` is false.

```csharp
[Test]
public void DifferentEncodingsOfSameStringAreNotEqual()
{
    string input = "Hello World";
    byte[] data = System.Text.Encoding.Unicode.GetBytes( input );
    string garbage = System.Text.Encoding.UTF8.GetString( data, 0, data.Length);

    Assert.That(input, Is.Not.EqualTo(garbage));             // Passes
    Assert.That(input, Is.Not.EqualTo(garbage).IgnoreCase);  // Fails
}
```

stevenaw: It looks like this encoding test is actually quite a long-standing change (going back to [NUnit 2.5.1](https://github.com/nunit/nunitv2/commit/641f0e14631a62ab7fe138ed3618a6c99a888def), and referencing a bug tracker ID I'm not familiar with). I'm unsure when case insensitivity was added in relation to it, but I'm thinking it might be worth rolling back all culture-aware changes made in this PR (and going back to using `ToLower()`) and instead spinning off a separate ticket to discuss explicitly moving to `CurrentCulture/CurrentCultureIgnoreCase`.

Thoughts?

ChrisMaddock: Would that be a lot of work? I think that's possibly a good idea...the culture issue doesn't really feel like one to be tackled bit-by-bit. 😢

jnm2: Yes, let's try to keep the issue separate from this PR so long as we aren't regressing the master branch in any way.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
jnm2: This would make a nice comment on the same line.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
jnm2: My outstanding concern is that the implementation of IComparable.CompareTo may contradict NUnit's equality comparer.

What do you think about the approach of scanning every item in the collection to determine if NUnit equality *would* delegate to the type's .Equals method and then using a hash set if so? Would that address your scenario?

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: Done (at least as far as this branch goes). Will make an issue shortly

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: I'm not entirely sure I understand, but do you mean apply the check here to every item's type in the `IEnumerable`?
https://github.com/nunit/nunit/blob/15175cba113b3937862d2f258320429658ce7440/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs#L82

If so, that's a really good idea. I can give that a try.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: I've created https://github.com/nunit/nunit/issues/3549

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
stevenaw: Thanks for the suggestion! This worked. I've done this in the [TryInferFastPath](https://github.com/nunit/nunit/pull/3523/files#diff-f6b841c6c7eed9838e33cf9082f26944R104) function below.

src/NUnitFramework/tests/Constraints/UniqueItemsConstraintTests.cs
```diff
@@ -79,5 +143,175 @@ public void PerformanceTests(IEnumerable values, bool ignoreCase)
...
+
+            var items = new SimpleObjectCollection((IEnumerable<object>)allData);
+            var constraint = new UniqueItemsConstraint();
+            var stopwatch = new Stopwatch();
+
+            stopwatch.Start();
+            constraint.ApplyTo(items);
+            stopwatch.Stop();
+
+            Assert.That(stopwatch.ElapsedMilliseconds, Is.GreaterThanOrEqualTo(50));
```
stevenaw, line 191: I had a hard time figuring out how to determine if the fast path wasn't being taken without doing a test like this.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
rprouse: I agree with @jnm2 let's keep all this separate. Thanks for creating the issue @stevenaw. I'm going to mark this conversations resolved.

Issue #3526, opened 2020-04-26
Add .NET Core 2.1 build of Framework


ChrisMaddock:
There's currently a problem with running tests on .NET Core, that they can only be run successfully if the test assembly and runner are located in the same directory. This hasn't been too much of a problem yet, as for the adapter that's always the case, but it is a problem for users of the engine  (https://github.com/nunit/nunit-console/issues/710), and will be for the .NET Core Console.

We currently have some reflection-based code which attempts to access .NET Core specific functionality from .NET Standard to correctly load the assemblies - although it's not working for the engine's case, due to the above assumption. It's also tricky in terms of maintenance - and at this point, I suggest we should just add a .NET Core build and remove the reflection.

https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/AssemblyHelper.cs#L90

I'm suggesting this as an additional build rather than a replacement for .NET Standard 2.0, to continue supporting our Xamarin users. I'm suggesting .NET Core 2.1 as the oldest current LTS version that I think the required AssemblyLoadContext APIs are available on.

This is a bit of a blocker for the console, so I'm happy to try and look at it. (Although anyone else who'd like to take this on would be very welcome! 😄)

CharliePoole:
I'm in favor of this feature (I think it's a feature rather than a bug, btw) but I have a bit of a side comment, which might impact the console project...

"There's currently a problem with running tests on .NET Core, that they can only be run successfully if the test assembly and runner are located in the same directory. "

I'm currently running .NET Core tests using the .NET Standard 2.0 engine. I find I need to have the nunit framework in the same directory, but nothing else. We should probably go elsewhere to discuss why this is working for me.

ChrisMaddock:
You're probably right there - it's been a long time since I last tried to debug this. 🙂

ChrisMaddock:
I've come to the conclusion that this is the wrong approach, and we should instead be approaching this at the point we load the framework, in the engine.

This issue is superceded by a number of PR's attached to https://github.com/nunit/nunit-console/issues/710

PR #3530, opened 2020-05-02
Add .NET Core 2.1 build of Framework and NUnitLite

ChrisMaddock:
Contributes to #3526 

This PR adds the new build, but makes no functionality changes yet.

Will keep these as two separate PR's to make this easier to review - leaving this in draft from until I've confirmed the functionality change does what we need it to do. 🙂

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -32,7 +32,8 @@ var AllFrameworks = new string[]
     "net45",
     "net40",
     "net35",
-    "netstandard2.0"
+    "netstandard2.0",
+    "netcoreapp2.1"
 };
 
 var NetCoreTests = new String[]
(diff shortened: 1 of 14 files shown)


build.cake
```diff
@@ -347,10 +368,9 @@ Task("PackageZip")
             GetFiles(CurrentImageDir + "bin/net35/**/*.*") +
             GetFiles(CurrentImageDir + "bin/net40/**/*.*") +
             GetFiles(CurrentImageDir + "bin/net45/**/*.*") +
-            GetFiles(CurrentImageDir + "bin/netstandard1.4/**/*.*") +
             GetFiles(CurrentImageDir + "bin/netstandard2.0/**/*.*") +
-            GetFiles(CurrentImageDir + "bin/netcoreapp1.1/**/*.*") +
-            GetFiles(CurrentImageDir + "bin/netcoreapp2.0/**/*.*");
```
mikkelbu, line 353: Are this changes also correcting some older changes, where we overlooked this section of the code?

build.cake
```diff
@@ -242,6 +243,26 @@ foreach (var runtime in new[] { "netcoreapp2.1", "netcoreapp3.1" })
     testNetStandard20.IsDependentOn(task);
 }
 
+var testNetCore21 = Task("TestNetCore21")
```
mikkelbu, line 246: I must be overlooking something, but how is this different from the above section for NetStandard20? The only differences I found diffing was in strings for names and descriptions, but how do we know that the tests are run against the .NET Core 2.1 version of the framework?

ChrisMaddock:
Thanks for looking through this Mikkel. I think I've come to the conclusion that I was taking the wrong approach here, and we actually want to approach this problem at the point we load the framework assembly in the engine. 

Thanks for the review, regardless however. 🙂

PR #3535, opened 2020-05-08
Fix accidental use of 'false negative' instead of 'false positive'

jnm2:
I was confused :)

### DIFF
--- a/src/NUnitFramework/tests/Internal/RandomizerBiasTests.cs
+++ b/src/NUnitFramework/tests/Internal/RandomizerBiasTests.cs
@@ -29,9 +29,9 @@ namespace NUnit.Framework.Internal
         private const double ConfidenceInterval = 0.00965802342765787;
 
         /* Random generators wouldn’t be truly random if it was impossible to hit a highly biased sequence.
-        These tests fail as a false negative whenever an actually-*unbiased* generator provides a sequence at
+        These tests fail as a false positive whenever an actually-*unbiased* generator provides a sequence at
         random that turns out to be biased this time.
-        What we can do is quantify the *chance* of such a false negative and choose that chance to be sufficiently
+        What we can do is quantify the *chance* of such a false positive and choose that chance to be sufficiently
         small to not bother us with spurious failures, while maintaining a precise enough signal to fail loud
         and clear if the generator is biased.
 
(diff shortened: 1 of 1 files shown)


CharliePoole:
If a test passes, i.e. it asserts conditions that are true, that's a positive result. If it fails, that's negative. So if it passes incorectly, that's a false positive and if it fails incorrectly, that's a false negative.

The reason this is confusing (I think) is that in software testing positive is good, negative is bad. OTOH in most medical tests, like testing for the Corona virus, positive as bad and negative is good.

In software, a false positive is the worst! We think it's working and it isn't. False negatives are bad too, but they just waste time rather than shipping errors.

In testing for disease, a false positive is bad but a false negative is really really bad.

It's no wonder you're confused. :smiley_cat:

jnm2:
@CharliePoole Oh no, are you sure? That means I just switched it the wrong way. I did a search and didn't find much, but here's someone's usage on Wikipedia that treats "test failures" and "false positives" as synonyms:

> It is also essential to implement a sustainable process for ensuring that test case failures are reviewed regularly and addressed immediately.[14] If such a process is not implemented and ingrained into the team's workflow, the application will evolve out of sync with the unit test suite, increasing false positives and reducing the effectiveness of the test suite.

https://en.wikipedia.org/wiki/Unit_testing#Limitations_and_disadvantages

jnm2:
Oh, here we go via https://stackoverflow.com/questions/43726965/test-failures-false-positive-or-false-negative:

> Simply put, false positives are test instances that fail without there being a defect in the application under test, i.e., the test itself is the reason for the failure.

(https://www.ontestautomation.com/on-false-negatives-and-false-positives/)

> False positive:– Test is marked as failed even in reality it is passed, or the functionality works properly

(
https://narainko.wordpress.com/2012/08/26/understanding-false-positive-and-false-negative/)

CharliePoole:
Well, the SO example actually has conflicting answers. I'm surprised to see Gerard quoted that way. The blog is by somebody I don't know, but he uses medical tests as a point of comparison, where negative == GOOD (no disease) and I've always been told that the good vs bad thing depended on how the hypothesis was stated.

Wikipedia (https://en.wikipedia.org/wiki/False_positives_and_false_negatives#False_negative_error) equates false positive and false negative to type 1 and type 2 errors respectively. This ties it in to the particular hypothesis you are testing.

In unit tests, we make Assertions of various hypotheses, like assert that adding 2 and 2 gives 4. If the hypothesis passes, that's positive. Hypothetically, I guess there oculd be some kind of testing where we asserted that an error exists. That would reverse the meaning.

At least, that's how I've used the terms in unit testing for lots of years.

Still, it's important to know how folks in general understand it and there seems to be a mix. I'm putting out the question to my twitter peeps.

jnm2:
I was thinking not in terms of a medical virus but of computer virus scanning. Close enough though? 😀

To make matters worse, the words 'positive' and 'negative' are overloaded between the terms 'false positive/negative' and 'positive/negative test.' I'd say that a positive test has the hypothesis that the software correctly accepts valid input, and a negative test has the hypothesis that the software correctly rejects invalid input.

I think your reasoning makes more sense. But I'm afraid of being misunderstood no matter what I say now. Maybe this is a word like 'comprise' that can have either one meaning or the exact opposite meaning based on common usage and I should just strike it from my vocabulary.

How about 'false failure' and 'false success'?

jnm2:
I've also used the terms 'broken' for either one and 'silently broken' or 'ineffective' for false success.

CharliePoole:
I'm used to having to explain these terms to people. It never occurred to me before that maybe I just shouldn't use them!!! Or at least use them coupled with a better explanatory phrase.

PR #3536, opened 2020-05-09
Reduce newly added API surface

jnm2:
New API members showed up after merging #3486. Public APIs constrain our future decisions, so we should only add ones that are necessary.

/cc @Falco20019 so that we aren't blocking something you need.


Falco20019:
Thanks for the adjustment @jnm2 , I think it's a good idea.

jnm2:
Thanks @Falco20019!

PR #3541, opened 2020-05-10
Annotate remaining Interfaces types

jnm2:
Contributes to #3376.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Randomizer.cs
+++ b/src/NUnitFramework/framework/Internal/Randomizer.cs
@@ -21,6 +21,8 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+#nullable enable
+
 using System;
 using System.Collections;
 using System.Collections.Generic;
@@ -61,7 +63,7 @@ namespace NUnit.Framework.Internal
 
         // Static Random instance used exclusively for the generation
         // of seed values for new Randomizers.
-        private static Random _seedGenerator;
+        private static Random _seedGenerator = null!; // Initialized by the InitialSeed setter called by the static constructor.
 
         /// <summary>
         /// Initial seed used to create randomizers for this run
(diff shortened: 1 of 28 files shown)


jnm2:
I was going to annotate TestContext in another branch, but because that process informed a change needed in this branch and there are hardly any changes to do for TestContext, I pushed those commits here instead. Ready for review.

jnm2:
CI failed and I fixed a mistake in "Inline method to get rid of nullability warning" where I forgot to remove the `return` keyword when I copied the method contents.

src/NUnitFramework/framework/Internal/Randomizer.cs
```diff
@@ -61,7 +63,7 @@ static Randomizer()
 
         // Static Random instance used exclusively for the generation
         // of seed values for new Randomizers.
-        private static Random _seedGenerator;
+        private static Random _seedGenerator = null!; // Initialized by the InitialSeed setter called by the static constructor.
```
rprouse, line 66: `null!` is new to me. Can you explain? I tried to Google it but I couldn't find it. Thx.

jnm2: Ah, the `!` part means "suppress the nullability warning." I think of the null forgiveness operator `!` as being an inline pragma.

`= null!` tells the compiler, "trust me, this field is always initialized before being accessed." It's necessary in this case because the static constructor doesn't directly set the field, and the compiler has not been given the smarts to analyze across more than one method body at a time. But I believe this is being worked on, so one day we will probably be able to remove the `= null!`.

Another approach would be to refactor so that `_seedGenerator` is directly initialized in the static constructor.

rprouse: Thanks. You have your one review if you want to merge. 👍

jnm2: Cool, I do want to get this in because TestContext is a user-facing piece.

PR #3544, opened 2020-05-11
Add method to set Until date when ignoring TestCaseData

### LINKED ISSUE #3275: Enable setting IgnoreUntilDate in TestCaseData.Ignore

rprouse:
Fixes #3275 

I've resolved the merge conflicts in #3277 and de-duplicated the code as requested by @jnm2. I couldn't push to @abborg's fork, so I'm opening this PR to replace @abborg's.

From @abborg's PR;

Provide a simple method to set the IgnoreUntilDate on a test case defined by TestCaseData. The behavior is the same as `IgnoreAttribute` and only applies the ignored status if the test is not already set to `NotRunnable` and the `Until` date has not been reached yet.


src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
```diff
@@ -1,440 +1,481 @@
...
-                }
-            }
-
-            yield return test;
-        }
-
-        #endregion
-    }
-}
+// ***********************************************************************
```
rprouse, line 1: I originally merged this in the web interface. Not sure what is going on, probably line-feeds. I'll check and update so we get a proper diff.

src/NUnitFramework/framework/Attributes/IgnoreAttribute.cs
```diff
@@ -85,8 +86,7 @@ public void ApplyToTest(Test test)
                     if (_untilDate.Value > DateTime.Now)
                     {
                         test.RunState = RunState.Ignored;
-                        string reason = string.Format("Ignoring until {0}. {1}", _untilDate.Value.ToString("u"), _reason);
-                        test.Properties.Set(PropertyNames.SkipReason, reason);
+                        test.AddIgnoreUntilReason(_untilDate.Value, _reason);
```
rprouse, line 89: @jnm2 I couldn't include the `RunState` in the de-duplication easily. The interface is `get` only and the derived classes have a wide range of private and public sets.

src/NUnitFramework/framework/Interfaces/ICommonTestData.cs
```diff
@@ -0,0 +1,49 @@
...
+
+#nullable enable
+
+namespace NUnit.Framework.Interfaces
+{
+    /// <summary>
+    /// Represents data that is common between an <see cref="ITest"/>
+    /// and <see cref="ITestData"/> to allow for de-duplication of code.
+    /// </summary>
+    public interface ICommonTestData
```
rprouse, line 32: The places we wanted to de-duplicate are either an `ITest` or an `ITestData`, so I couldn't add an extension method for both of those. I pulled the common properties into this common base inteface so that we could start de-duplicating more of the messages with extension methods. I thought extension methods were a good option for common code now that we no longer have .NET 2.0 targets.

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
rprouse: Looking at the history, it looks like you fixed the line endings in this file a couple of months ago @jnm2. I checked and the line endings are all still CRLF, so I'm not sure why it isn't showing a diff. You'll have to take my word that it is the same as your previously reviewed @jnm2.

src/NUnitFramework/framework/Interfaces/ITest.cs
```diff
@@ -77,21 +77,11 @@ public interface ITest : IXmlNodeBuilder
         /// </summary>
         IMethodInfo? Method { get; }
 
-        /// <summary>
-        /// Gets the RunState of the test, indicating whether it can be run.
-        /// </summary>
-        RunState RunState { get; }
```
jnm2, line 83: If I remember right, moving interface methods down to a base interface causes MissingMethodException for already-compiled libraries that referenced the previous version of NUnit when they run against the new version of NUnit. E.g. https://stackoverflow.com/questions/55061714/missingmethodexception-after-extracting-base-interface.

The primary example is a NuGet test helper library that itself references on NUnit. The user installs the newest version of the helper library and the newest version of NUnit into their project, and then the helper library calls into `ITest.RunState` and the CLR says "I can't find that." (The CLR doesn't search base interfaces because of the diamond dependency problem. At least, this was the state of things before .NET Core added Default Interface Methods.)

Less of a hassle (because it's actionable) is that this change will also cause the following code to stop compiling:

```cs
class MyTest : ITest
{
    // ...

    // The compiler will error until ITest is changed to ICommonTestData:
    RunState ITest.RunState => _runState;
}
```

jnm2: What do you think about making `AddIgnoreUntilReason` an extension method on `IPropertyBag` so it's `test.Properties.AddIgnoreUntilReason`?

src/NUnitFramework/framework/Attributes/IgnoreAttribute.cs
jnm2: Thanks, I didn't realize that at the time.

src/NUnitFramework/framework/Interfaces/ICommonTestData.cs
jnm2: If I remember right, moving interface methods down to a base interface causes MissingMethodException for already-compiled libraries that referenced the previous version of NUnit when they run against the new version of NUnit. E.g. https://stackoverflow.com/questions/55061714/missingmethodexception-after-extracting-base-interface.

The primary example is a NuGet test helper library that itself references on NUnit. The user installs the newest version of the helper library and the newest version of NUnit into their project, and then the helper library calls into `ITest.RunState` which existed at the point in time when the helper library was compiled. The newest version doesn't have that method though because it moved to `ICommonTestData.RunState`. So the CLR says "I can't find that." (The CLR doesn't search base interfaces because of the diamond dependency problem. At least, this was the state of things before .NET Core added Default Interface Methods.)

Less of a hassle (because it's actionable) is that this change will also cause the following code to stop compiling:

```cs
class MyTest : ITest
{
    // ...

    // The compiler will error until ITest is changed to ICommonTestData:
    RunState ITest.RunState => _runState;
}
```

What do you think about making `AddIgnoreUntilReason` an extension method on `IPropertyBag` so it's `test.Properties.AddIgnoreUntilReason`?

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
jnm2: Just tried this and it worked:
image

src/NUnitFramework/framework/Interfaces/ICommonTestData.cs
rprouse: That sounds like a good alternative. I'll do it tomorrow. Thanks.

rprouse: @jnm2 I've reverted the interface change and switched the extension to iPropertyBag. This is ready for review again.

PR #3527, opened 2020-04-26
Draft experiment: Docs via docfx

SeanKilleen:
Inspired by discussion at https://github.com/nunit/docs/issues/337

Exploring the option of generating world-class docs in a maintainable and welcoming way using docfx.

## What other projects are using docfx? 

* [Akka.NET](https://getakka.net/)'s docs are built using docfx.

## Work Remaining

Wiki Content to Port:

* [x] NUnit Docs
* [x] NUnit Xamarin Runners
* [x] VS Test Adapter
* [x] VS Test Generator
* [x] NUnit Analyzers
* [x] Developer Info

As part of the initial effort: 

* [x] Get a docfx build going
* [x] Generate API documentation for publicly documented NUnit API surface
* [x] Allow browsing of multiple versions? (if docfx allows this well)
* [ ] Add contribution guide for docs
* [ ] Add readme for building docs
* [x] Add logos etc. as appropriate.
* [x] Check wiki references (e.g. links within `[[ ]]`) to point them to the right spot.
* [x] Ensure all warnings are resolved
* [ ] Check wiki repo to ensure all content has migrated
* [x] Breadcrumb navigation
* [x] Add copyright / license to footer
* [ ] Add home page
* [ ] Check links for pointers to wiki and change to markdown references

Decisions to be made by NUnit team representatives:

* [ ] Decide where docs should live (whether to have this live at `nunit/docs` or somewhere other than in this folder.)
* [ ] Whether to include generated source code docs
* [ ] If there's an attribute name they'd like to use to hide certain code from docs publication
* [ ] Whether additional articles (e.g. VS test runner) should be captured in this site as well or done separately.
* [ ] Decide what to do about potential 404s (URLs will be changing)

Once those decisions have been made: 

* [ ] Exclude any internal public methods using filter config and an attribute (maybe a new `MeantForInternalConsumptionAttribute` or something along those lines?)
* [ ] Generate docs for additional APIs outside of this repo?
* [ ] Update build to utilize console docs builder
* [ ] Update azure pipelines to build and push docs.
* [ ] Ensure wiki content changes since this PR are reflected in the new docs

Formatting after full contents have been ported:

* [ ] Ensure all pages have title frontmatter (for title bar)
* [ ] Ensure all pages have top headings (for web page -- use title if possible?)
* [ ] Call out info, warnings, cautions, etc. using docfx syntax
* [x] Replace `C#` code snippet language with `csharp`
* [ ] Clean up the navigation
* [ ] Check alignment of code samples
[...]

PR #3550, opened 2020-05-16
Issue 3475 reformat xml comments

### LINKED ISSUE #3475: Our XML comments are using <code> (block element) instead of <c> (inline element)
Example:

https://github.com/nunit/nunit/blob/30df2d13c0f67eeb6b3dc7f56eebd7c5cfb5c7de/src/NUnitFramework/framework/Assert.Conditions.cs#L218-L222

Visual Studio has recently begun taking the XML comment formatting tags seriously, so it may make a difference.

I would recommend that we add the design-time tool https://www.nuget.org/packages/DotNetAnalyzers.DocumentationAnalyzers to our Directory.Build.props and follow all its prompts. It's been awesome in every project where I've seen it. I know it catches this particular case.

IrisClasson:
fixes #3475 

As recommended in the issue I added the documentation analyzers to Directory.Build.props. There are only beta releases for this package, and there hasn't been an update for a while.

I've gone through the rules one by one and updated the XML comments and double checked the end result.

Let me know if there is anything more I can do, or if something looks strange.

### DIFF
--- a/src/NUnitFramework/Directory.Build.props
+++ b/src/NUnitFramework/Directory.Build.props
@@ -32,4 +32,8 @@
     <EmbedAllSources>true</EmbedAllSources>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="DotNetAnalyzers.DocumentationAnalyzers" Version="1.0.0-beta.59"/>
+  </ItemGroup>
+
 </Project>
--- a/src/NUnitFramework/framework/Assert.Conditions.cs
+++ b/src/NUnitFramework/framework/Assert.Conditions.cs
@@ -209,7 +209,7 @@ namespace NUnit.Framework
         #region NotNull
 
         /// <summary>
-        /// <para>Verifies that the object that is passed in is not equal to</para> <code>null</code><para>. Returns without throwing an
+        /// <para>Verifies that the object that is passed in is not equal to <see langword="null"/>. Returns without throwing an
         /// exception when inside a multiple assert block.</para>
         /// </summary>
         /// <param name="anObject">The object that is to be tested</param>
@@ -221,7 +221,7 @@ namespace NUnit.Framework
         }
 
         /// <summary>
-        /// <para>Verifies that the object that is passed in is not equal to</para> <code>null</code><para>. Returns without throwing an
+        /// <para>Verifies that the object that is passed in is not equal to <see langword="null"/>. Returns without throwing an
         /// exception when inside a multiple assert block.</para>
         /// </summary>
         /// <param name="anObject">The object that is to be tested</param>
@@ -231,7 +231,7 @@ namespace NUnit.Framework
         }
 
         /// <summary>
-        /// <para>Verifies that the object that is passed in is not equal to</para> <code>null</code><para>. Returns without throwing an
+        /// <para>Verifies that the object that is passed in is not equal to <see langword="null"/>. Returns without throwing an
         /// exception when inside a multiple assert block.</para>
         /// </summary>
         /// <param name="anObject">The object that is to be tested</param>
@@ -243,7 +243,7 @@ namespace NUnit.Framework
         }
 
         /// <summary>
-        /// <para>Verifies that the object that is passed in is not equal to</para> <code>null</code><para>. Returns without throwing an
+        /// <para>Verifies that the object that is passed in is not equal to <see langword="null"/>. Returns without throwing an
         /// exception when inside a multiple assert block.</para>
         /// </summary>
         /// <param name="anObject">The object that is to be tested</param>
@@ -257,7 +257,7 @@ namespace NUnit.Framework
         #region Null
 
         /// <summary>
-        /// <para>Verifies that the object that is passed in is equal to</para> <code>null</code><para>. Returns without throwing an
+        /// <para>Verifies that the object that is passed in is equal to <see langword="null"/>. Returns without throwing an
         /// exception when inside a multiple assert block.</para>
         /// </summary>
         /// <param name="anObject">The object that is to be tested</param>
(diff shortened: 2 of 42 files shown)


src/NUnitFramework/Directory.Build.props
```diff
@@ -32,4 +32,8 @@
     <EmbedAllSources>true</EmbedAllSources>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="DotNetAnalyzers.DocumentationAnalyzers" Version="1.0.0-beta.59"/>
```
jnm2, line 36: Sorry for not mentioning. This should have `PrivateAssets="all"` in case we ever use `dotnet pack` to create our NuGet packages. Otherwise the NUnit package will have a .nuspec dependency on `DotNetAnalyzers.DocumentationAnalyzers`, causing any project that references NUnit to transitively reference DocumentationAnalyzers.

jnm2: Tracked at https://github.com/nunit/nunit/pull/3551.

oznetmaster:
I cannot find anything in the .NET documentation rules that suggests/requires using a `<para>` tag inside another xml comment element when there is only a single paragraph of text in that element.

In fact, the rules state "You use the <para> tag to format the content within its parent tag. `<para>` is usually used inside a tag, such as `<remarks>` or `<returns>`, **to divide text into paragraphs**."

Why were the `<para>` tags added to all the single paragraph xml comment elements?

jnm2:
@oznetmaster This is explained in the commit messages. I suggested that we delegate expertise in this matter to the community DocumentationAnalyzers project. Another option would have been to add a suppression of DOC100 to the .editorconfig file. I don't feel strongly.

@sharwell may be able to explain the rationale of https://github.com/DotNetAnalyzers/DocumentationAnalyzers/blob/master/docs/DOC100.md.

oznetmaster:
I prefer to utilize the Microsoft .NET Documentation rules/guidelines.

If the DocumentationAnalyzers project has decided to implement its own set of rules that are different, that is its choice, but that does not mean that we need to blindly follow them.

Requiring a para tag in single paragraph comments makes no sense whatsoever, and makes them less readable in the code, and requires more maintenance for no apparent reason.

jnm2:
I found the rationale and discussion: https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/601

Looks like the two reasons are `<inheritdoc/>`, which we might use, and HTML generators, which we are actively planning to use: https://github.com/nunit/nunit/pull/3527

I don't mind disabling this safety check though if the extra tags are worth going to the trouble of removing now.

oznetmaster:
I notice that many of the people who commented on this also thought that it was a bad idea.  For example:

``
But as I've never seen it so far, I'm still leaning toward not wanting this rule since adding <para> tags to a single paragraph (even a few word sentence) of a doc section hurts readability where it matters most (in the code). Document formatting tools should be able to do whatever is necessary when tags are missing from a single paragraph. Adding <para> when multiple paragraphs are in use makes sense though.
``

``
I'm actually 👎 in having this rule in the main StyleCop.Analyzers. Perhaps Sandcastle.Analyzers, but I don't think the majority of developers would benefit from this rule. In fact, if not using Sandcastle, I feel this makes the code worse.
``

I think we should take them out, and disable the error or whatever.

CharliePoole:
As a visiting coach, there's a thing I've run into over the years... static analysis software generally comes with default rules plus the ability to disable them but once it's in use those defaults become the de facto standard.

oznetmaster:
I do not think that will become the case here.  Too many objections to this.  All over the web.

src/NUnitFramework/framework/Assert.Conditions.cs
```diff
@@ -339,8 +339,8 @@ public static void IsNaN(double? aDouble, string? message, params object?[]? arg
         }
 
         /// <summary>
-        /// Verifies that the double that is passed in is an <code>NaN</code> value. Returns without throwing an
-        /// exception when inside a multiple assert block.
+        /// <para>Verifies that the double that is passed in is an <c>NaN</c> value. Returns without throwing an
+        /// exception when inside a multiple assert block.</para>
```
sharwell, line 343: 💭 Interesting. The `<para>` warning in this case was actually a red herring. The use of `<code>` made it _think_ there were multiple paragraphs when in reality it should have used `<c>` for just one paragraph.

CharliePoole:
> I do not think that will become the case here. Too many objections to this. All over the web.

Au contraire, it seems like that's exactly what has already happened here. :wink:

oznetmaster:
No.  In fact, @jnm2 has even indicated that he has no problem disabling the check and removing the extra tags.

I know that my programming shop will NEVER adapt this.  I have talked to many others who feel the same way.

CharliePoole:
@oznetmaster My comment wasn't about anything Joseph is or isn't doing but was a general one. My experience has led me to feel that folks will generally go with the delivered defaults in static analysis tools. It seems that may have happened here although I wasn't following along with the discussion until your comment caught my eye.

src/NUnitFramework/framework/Assert.Conditions.cs
jnm2: Thanks for pointing that out, that's helpful.

jnm2:
Only took ten minutes, mostly because Visual Studio's "configure severity in editorconfig" fix froze for minutes and didn't result in any changes when it unfroze. I had to type the .editorconfig change by hand.

src/NUnitFramework/framework/Assert.Conditions.cs
CharliePoole: Nothing to do with this issue, but that's actually an odd comment. The lack of an exception in a multiple assert block is a feature of __all__ asserts, not just this one.

jnm2: @CharliePoole I added that comment to all assertion methods because of how often the comes up in support questions that these methods can be depended on to throw AssertionException for failures.

CharliePoole: Ah! I see you did that quite a while back, in fact.

PR #3553, opened 2020-05-19
Enable message pump strategies to light up on any framework that has the correct types

### LINKED ISSUE #3552: MessagePumpStrategy does not work for WPF on netcoreapp3.0 and upwards
I am trying port our applications to netcoreapp3.1.
During this i also tried to run our unit tests for that target framework.
To my surprise the tests deadlocked as soon as they use async or things like `Assert.CatchAsync`.
The issue boils down to `MessagePumpStrategy` which had special cases for windows forms and WPF on net framework 4.

Is there a reason why there are no special cases for netcoreapp3.0 and upwards or was it just an oversight and those should be there?

jnm2:
Fixes #3552

### DIFF
--- a/src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
+++ b/src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
@@ -22,8 +22,11 @@
 // ***********************************************************************
 
 using System;
+using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Security;
 using System.Threading;
+using NUnit.Compatibility;
 
 #if NET40 || NET45
 using System.Windows.Forms;
@@ -43,15 +46,9 @@ namespace NUnit.Framework.Internal
             if (context is SingleThreadedTestSynchronizationContext)
                 return SingleThreadedTestMessagePumpStrategy.Instance;
 
-#if NET40 || NET45
-            if (context is WindowsFormsSynchronizationContext)
-                return WindowsFormsMessagePumpStrategy.Instance;
-
-            if (context is DispatcherSynchronizationContext)
-                return WpfMessagePumpStrategy.Instance;
-#endif
-
-            return NoMessagePumpStrategy.Instance;
+            return WindowsFormsMessagePumpStrategy.GetIfApplicable()
+                ?? WpfMessagePumpStrategy.GetIfApplicable()
+                ?? NoMessagePumpStrategy.Instance;
         }
 
         private sealed class NoMessagePumpStrategy : MessagePumpStrategy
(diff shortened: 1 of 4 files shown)


src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
```diff
@@ -65,18 +62,52 @@ public override void WaitForCompletion(AwaitAdapter awaiter)
...
+
+                    var applicationRun = (Action)applicationType
+                        .GetMethod("Run", BindingFlags.Public | BindingFlags.Static | BindingFlags.DeclaredOnly, null, Type.EmptyTypes, null)
+                        .CreateDelegate(typeof(Action));
+
+                    var applicationExit = (Action)applicationType
+                        .GetMethod("Exit", BindingFlags.Public | BindingFlags.Static | BindingFlags.DeclaredOnly, null, Type.EmptyTypes, null)
+                        .CreateDelegate(typeof(Action));
+
+                    _instance = new WindowsFormsMessagePumpStrategy(applicationRun, applicationExit);
```
jnm2, line 94: There is no need for racing threads to agree upon a single WindowsFormsMessagePumpStrategy instance, so Interlocked.CompareExchange wouldn't add value.

batzen:
@jnm2 I can confirm that my tests don't deadlock anymore after using the version from this PR.
Thank you so much!

mikkelbu:
@jnm2 I can take a look at this PR this weekend. Currently, it does not build on macOS and Linux with the failure
```
...
Windows is required to build Windows desktop applications.
...
```

batzen:
Looking at the code the changes in the project file don't seem necessary as the problem/issue is solved by using reflection.

jnm2:
@batzen The changes (or something like them) are needed in order to make CI fail if support for .NET Core was revoked.

batzen:
What do you mean with"revoked"? If we need something to make the build fail if something in .net changes i could provide some tests through a separate PR.

jnm2:
@batzen I mean, I wanted to have a test that would fail before making any implementation changes. That way if we accidentally revert the behavior in the implementation (say, adding incorrect preprocessor back in), we will notice what we are breaking.

jnm2:
I think Microsoft.NET.Sdk.WindowsDesktop is a no-go. When I have the time, I'll try to see what the next best way is to reference System.Windows.Forms and WindowsBase in the test project for .NET Core 3.1.

If necessary we can continue with no tests, but then there's nothing to tell us if we won't accidentally break one of the things that these tests guarantee on just .NET Core.

src/NUnitFramework/tests/SynchronizationContextTests.cs
```diff
@@ -104,7 +104,7 @@ public static void AsyncDelegatesAreExecutedOnStaThread(AsyncExecutionApiAdapter
             });
         }
 
-#if NETFRAMEWORK
+#if !NETCOREAPP2_1
```
jnm2, line 107: This change is what is causing the CI build failure. `System.Windows.Threading.DispatcherSynchronizationContext` and `System.Windows.Threading.DispatcherSynchronizationContext` aren't available to compile or run against in .NET Core 3.1 unless we use the Windows desktop SDK (which fails the build on other OSs) or come up with a workaround.

src/NUnitFramework/tests/nunit.framework.tests.csproj
```diff
@@ -38,6 +38,9 @@
 
   <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">
     <Reference Include="System.Web" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework' or '$(TargetFramework)' != 'netcoreapp3.1'">
     <Reference Include="System.Windows.Forms" />
     <Reference Include="WindowsBase" />
   </ItemGroup>
```
jnm2, line 46: These are currently producing missing reference warnings during compilation.

batzen:
I know how we can referecence the correct sdk only on windows. Will post code later or sometime during the weekend.

jnm2:
@batzen That would be great! I reset the branch back to the version that has the two csprojs on the Windows desktop SDK. If you'd like to have a commit with your name on it, submit a PR to this branch in my repo (https://github.com/jnm2/nunit/tree/ui_message_pump_lightup) and I'll fast-forward merge to bring it into this PR. Or if you want to just paste an example, that's great too.

batzen:
@jnm2 Created a PR in your fork.

jnm2:
@batzen Thanks so much for solving that problem for us, and @mikkelbu thanks for reviewing!

batzen:
@jnm2 you are welcome. thanks for solving the issue. Especially in such a future proof way.

jnm2:
I forgot, but CI didn't. The same changes are needed for nunitlite-runner.csproj since it is the host process for the CI tests. On it.

PR #3555, opened 2020-05-20
expose Test on TestAdapter

SimonCropp:
so i can avoid the reflection here https://github.com/VerifyTests/Verify/blob/master/src/Verify.NUnit/Verifier.cs


CharliePoole:
Exposing NUnit's internal representation of a test would mean that representation could not be changed without breaking users. Perhaps you could ask for the specific info you need to be exposed, rather than the entire representation of the test.

SimonCropp:
replaced by https://github.com/nunit/nunit/pull/3557

PR #3557, opened 2020-05-21
expose TypeInfo

SimonCropp:
@CharliePoole is this better?

### DIFF
--- a/src/NUnitFramework/framework/TestContext.cs
+++ b/src/NUnitFramework/framework/TestContext.cs
@@ -417,6 +417,24 @@ namespace NUnit.Framework
                 get { return _test.Arguments; }
             }
 
+            /// <summary>
+            /// The Type of the test.
+            /// Returns null if no fixture type is associated with it.
+            /// </summary>
+            public Type? TypeInfo
+            {
+                get { return _test.TypeInfo?.Type; }
+            }
+
+            /// <summary>
+            /// The MethodInfo of the test.
+            /// Returns null if the test is not implemented as a method
+            /// </summary>
+            public MethodInfo? MethodInfo
+            {
+                get { return _test.Method?.MethodInfo; }
+            }
+
             #endregion
         }
 


CharliePoole:
Less is better, for sure. But still, the TestAdapter class is there precisely to __avoid__ giving you details of the internal implementation, which may change in the future.

Actually, I don't work on the framework any longer, so I'll defer to the @nunit/framework-team members, for whom I intended the [original comment](https://github.com/nunit/nunit/pull/3555#issuecomment-631858261). They'll figure out how much they want to expose.

FYI, dropping the old PR in favor of a new one left no connection to the original, which is why I made a link. Discussing __whether__ to do something may work better in an issue you create and leaves behind an easier-to-find record for folks who come along later.

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -417,6 +417,24 @@ public PropertyBagAdapter Properties
                 get { return _test.Arguments; }
             }
 
+            /// <summary>
+            /// The Type of the test.
+            /// Returns null if no fixture type is associated with it.
+            /// </summary>
+            public Type? TypeInfo
```
jnm2, line 424: I would rather expose this as `FixtureType` or `Type`. However, can you explain at a higher level why ClassName doesn't give you the information you need?

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -417,6 +417,24 @@ public PropertyBagAdapter Properties
...
+            {
+                get { return _test.TypeInfo?.Type; }
+            }
+
+            /// <summary>
+            /// The MethodInfo of the test.
+            /// Returns null if the test is not implemented as a method
+            /// </summary>
+            public MethodInfo? MethodInfo
```
jnm2, line 433: Same question here. I see you're using this [in your code](https://github.com/VerifyTests/Verify/blob/19983c954581557dd7fea0df3b489a6c4083fcce/src/Verify.NUnit/Verifier.cs#L22) to pass to TestNameBuilder.GetUniqueTestName. It sounds like a more direct feature for NUnit to expose might be a guaranteed unique test name property which might address a more generalizable need.

SimonCropp: I am not sure if Verifies unique test name would be useful to Nunit? it is currently https://github.com/VerifyTests/Verify/blob/master/src/Shared/TestNameBuilder.cs

but either way, people can extend nunit in many ways, and exposing the current type and method info enables those scenarios

SimonCropp:
@CharliePoole 

> dropping the old PR in favor of a new one left no connection to the original, which is why I made a link. Discussing whether to do something may work better in an issue you create and leaves behind an easier-to-find record for folks who come along later

i dont understand. the link back to the other PR is clearly visible in this PR

image

src/NUnitFramework/framework/TestContext.cs
jnm2: Do you need that specific implementation, or could you also find it useful if we exposed something similar to VSTest's FQN (fully-qualified name)? Not saying we should do that, but just as an example.

SimonCropp: > Do you need that specific implementation 

yep

exposing an FQN for the would probably work. but i need the full methodinfo, so it would seem strange to expose the methodinfo and not the typeinfo.

CharliePoole:
@SimonCropp Regarding the links between PRs, none was visible when I posted my comment, which was a very short while after you posted yours. Maybe latency?

CharliePoole:
@jnm2 If the intention is to expose more info for extenders, isn't that at least in part why there's an internal namespace? Everything needed is available through TestExecutionContext. But TestContext is for general usrs, i.e. writers of tests.

jnm2:
@CharliePoole I didn't think the internal namespace was for extenders. If I'm understanding correctly, extenders are usually people who publish libraries as binaries to use along with NUnit. There is more of a practical requirement for NUnit APIs to remain stable for extenders than for regular users because of all the combinations that become possible between published versions of NUnit and published versions of the extending library. The internal namespace doesn't fit the bill as a thing that remains more stable than our other APIs, at least not the way I'm remembering us talking about it.

If we've lost an important distinction between things in the Internal namespace that need to be ultra-stable and things that we can change for our convenience, maybe we should go through the namespace and document which is which. I'll need help if that's what we think we should do, because I don't know which parts are extender APIs and which parts are not intended to be APIs. (Unless anything public is automatically for extenders to use?)

src/NUnitFramework/framework/TestContext.cs
jnm2: The FQN would cover both the type and method name. Is there something that you need the full methodinfo for that the FQN (or some guaranteed unique name) doesn't give you?

SimonCropp: >  Is there something that you need the full methodinfo for that the FQN (or some guaranteed unique name) doesn't give you?

the parameter infos.

CharliePoole:
@jnm2 I don't think we have lost the distinction, rather we never made it very clearly. I hope that can happen at some point.

When you extend the framework, as opposed to the engine, you are limited to extending a single version. You generally have to reference it to extend it and create a new version of your extension each time there's a new version of NUnit. So I actually think framework extenders require less stability than general users, third-party runner authors or engine extenders.

BTW I'm not __advocating__ for framework extensions to only be for a particular version. I am merely saying it's a built-in limitation of how it works. Most people who create custom attributes or assertions do it for their own work rather than for publication, which makes it less of a problem to have to rebuild under each new release.

src/NUnitFramework/framework/TestContext.cs
jnm2: VSTest's FQN includes the parameter names for the method as well as (IIRC) parameter values. Assuming you had a test name like this: `A.B.C.Method(x: 1, y: "2")`, would you still need access to the System.Reflection side of the test?

jnm2:
> You generally have to reference it to extend it and create a new version of your extension each time there's a new version of NUnit. 

This is not the default pressure of the .NET ecosystem. As far as I know, there was never a good way for extension authors to enforce this in general. (This isn't a problem that is unique to NUnit.) In the specific case of distributing via NuGet package, we would have to spell out that package authors must use an exact-version .nuspec dependency on NUnit. Otherwise, mixed versions are what the IDE and dotnet CLI tooling will give you by default.

Exact-version dependencies tend to annoy the users of the extension who now can't use a prerelease of NUnit or, in general, update to use the latest bugfix or feature because the extension author may not be at the right place in their release cycle to immediately react to the new NUnit version. (Or the extension author may not have free time at the right time.)

src/NUnitFramework/framework/TestContext.cs
SimonCropp: a test name like that wont help me. i would still need to reflect into the internals of nunit

CharliePoole:
Joseph, I feel like you're missing the point here. I completely agree with you about how extensions ought to work ideally. I'm describing the limitations of what we have, which I doubt will be overcome during the NUnit 3 lifecycle.

SimonCropp:
any chance of getting this in 3.13?

SimonCropp:
any way i can help with this? perhaps get on a call to discuss?

SimonCropp:
should i close this?

CharliePoole:
@SimonCropp I'm commenting as a fellow user here. An @nunit/framework-team member will need to respond to you.

jnm2:
@SimonCropp Oh, hi. I'm sorry I lost track of this one. We have at least one user that implements IMethodInfo themselves. What would your library need if there *was* no MethodInfo? I didn't end up understanding what the MethodInfo/ParameterInfo would be used for. Could you give us a picture of that?

src/NUnitFramework/framework/TestContext.cs
jnm2: Continuing at https://github.com/nunit/nunit/pull/3557#issuecomment-731905892

SimonCropp:
@jnm2 essentially i want to avoid these hacks https://github.com/VerifyTests/Verify/blob/master/src/Verify.NUnit/Verifier.cs#L15-L36

jnm2:
https://github.com/VerifyTests/Verify/blob/master/src/Verify.NUnit/Verifier.cs#L31 could be `var test = TestExecutionContext.CurrentTest;` btw.

So, now I think there is an opportunity to expose something better than MethodInfo:

1. I think the user will expect the test name to follow the `TestName` property on TestCase attribute or TestCaseData objects rather than coming from the method.
2. TestCaseData has SetArgDisplayNames which allows users to customize the way arguments are displayed. The main thing I use it for is to hide extra wiring parameters that are derivative of the main parameters.

I would recommend using ITest.Name instead of method.Name, and I'd prefer to expose the name pieces you need as new things on ITest rather than breaking the abstraction and assuming that all ITest instances always have a method.

PR #3577, opened 2020-06-07
throws async should return a task

SimonCropp:
There seems to be quite a bit of legacy code around converting async to sync APIs. give nunit now supports tests returning a task, shouldnt most of that code be removed?


Temtaime:
Why closed?

Issue #2950, opened 2018-08-06
Ability to pass TestFixtureData to TestCaseSource


paylett:
This is a feature request. It would be great if there was a way to use [TestFixtureSource] to provide test fixtures, and then have the TestFixtureData instance for the current fixture instance available when dynamically generating test cases using [TestCaseSource].

To illustrate, my use case is as follows:
I am testing a report SQL generation engine. I wish to test the preparation of multiple reports over several customers. I wish to use TestFixtureSource to provide a list of customers, producing one fixture per customer; and then use TestCaseSource to provide a list of reports to test for that customer. The list of reports varies from customer to customer, and I wish to test that each report generates valid SQL. (Presently I am iterating over both the customers, and the reports for each, as a single TestCaseSource operation - with no generation of features. But it is desirable to group the reports, one fixture per customer, for UI and reporting convenience).

It appears that this would be possible by passing 'suite' from TestCaseSourceAttribute.BuildFrom to GetTestCasesFor, to GetTestCaseSource, then onto the Invoked callback in any of a number of ways, but I didn't want to just dive into this presumptuously.

Some suggestions regarding the API:

1. test to see if the first parameter of the TestCaseSource method accepts a TestFixtureData, and if it does, then prefix that to any other MethodParams.
2. provide a wrapping object that can pass both the TestFixtureData and MethodParams
3. provide an interface for the SourceType to implement
4. a new property on the TestCaseSourceAttribute to signal that the fixture data should be passed.

Fwiw, I found someone else was also looking for a similar feature [over here on StackOverflow](https://stackoverflow.com/questions/41719951/how-to-access-test-fixture-parameters-from-function-generating-test-case-paramet).

Thanks for considering. :)

ChrisMaddock:
Interestingly, I just came across the same problem. My case is I'm generating thousands of test cases based on a single excel document - and want to run this class on multiple excel documents.

I'll work around it by adding an extra loop in my TestCaseSource, but I loose all separation that way. Anyone any better ideas for a workaround?

CharliePoole:
I think it would help if somebody wrote the pseudo-code they would like to work. I think I understand but I'm not sure.

ChrisLusena:
This would solve my problem for the epic #7 as well.  I have the same case as https://github.com/nunit/nunit/issues/2707#issuecomment-421885014 . 
If I have the parameters to the constructor of the test fixture then I can generate test cases I want. 

So the test case generation process would go something like:
[...]

Issue #2707, opened 2018-02-11
Not able to use TestFixtureSource values in TestCaseSource object.


amittrehan44:
I am trying to use TestFixtureSource values and pass it in TestCaseSource but null value is passed.

namespace HybridFramework
{
    
   [TestFixtureSource("FixtureArgs")]
    public class ParameterExperiment
    {   
        static string eq1;

        static object[] FixtureArgs = {
        new object[] { "Question"},
        new object[] { "Answer"}
		};

        public ParameterExperiment(string eq1)
        {
            ParameterExperiment.eq1 = eq1;

        }

        [TestCaseSource("getFruit")]
        public void TestMethod(string fruit)
        {
            // TODO: Add your test code here
            Assert.Pass("Your first passing test"+fruit + eq1);
        }

        static object[] getFruit =
       {
        new object[] {"Apple" },
        new object[] { "Orange" },
        new object[] { ParameterExperiment.eq1 }
        };

    }
}

TestFixture "ParameterExperiment" has TestFixtureSource as "FixtureArgs" with two parameters 1)Question and 2)Answer

TestCase "TestMethod" has TestCaseSource as "getFruit" with three parameters
1)Apple
2)Orange
3)ParameterExperiment.eq1 (TestFixture Parameter)

Issue: If I am using eq1(TestFixture Parameter) in test method I am able to see TestFixture Parameter value passed
but if I am using eq1 in getFruit, null value is passed.

**Expected**:
ParameterExperiment("Question")
- TestMethod("Apple")
- TestMethod("Orange")
- TestMethod("Question")

ParameterExperiment("Answer")
- TestMethod("Apple")
- TestMethod("Orange")
- TestMethod("Answer")

**Actual**:
ParameterExperiment("Question")
- TestMethod("Apple")
- TestMethod("Orange")
- TestMethod(null)

ParameterExperiment("Answer")
- TestMethod("Apple")
- TestMethod("Orange")
- TestMethod(null)

CharliePoole:
The sequence of events is as follows:

1. The fixture is discovered and the value of it's argument is saved.
2. Each method is discovered and the value if its argument is saved.
3. You run the tests
4. Your fixture constructor is called with the saved value of its argument passed to it.
5. Your constructor saves that argument in a field.
6. Each test method is called and the saved value of its argument is passed to it.

So the test method gets the argument that was saved at step 2. This is simply how it works.

[...]

PR #3593, opened 2020-07-12
TestFixtureSource arguments usage in the TestCaseSource method

skomialek:
Hi,

I have found a need to use some information specific to `TestFixture` I'm in during `TestCaseSource` generation code.
It might be something that is being discussed in #2950 

The solution presented here allows to reference arguments passed to TestFixtureData in the TestCaseSource attribute. 

Please see sample usage in `TestCaseSourceWithTestFixtureArgumentsTests`.

The fixtures are created with data as follows 

```c#
        public static IEnumerable<TestFixtureData> TestFixturesSource
        {
            get
            {
                for (int maxArgs = 0; maxArgs <= NUMBER_OF_ARGS; maxArgs++)
                {
                    string[] args = FixtureArgumentsSource.Take(maxArgs).ToArray();

                    yield return new TestFixtureData(maxArgs, args).SetArgDisplayNames($"{maxArgs}_arguments_test");

                }
            }
        }
```

So that each fixture contains different number of arguments. In order to make test case source aware one can ask for the `maxArgs` value by referencing `TestFixtureArgumentRef.Arg0`.

```c#
        [Test]
        [TestCaseSource(nameof(TestCaseDataWithFixtureContext), new object[]{ TestFixtureArgumentRef.Arg0 })]
        public void FixtureArgumentTest(int argumentIndex, int numberOfArguments)
        {
            Assert.AreEqual(FixtureNumberOfArgs, numberOfArguments);
            Assert.AreEqual(FixtureArgs[argumentIndex], FixtureArgumentsSource[argumentIndex]);
        }

        public static IEnumerable<TestCaseData> TestCaseDataWithFixtureContext(int numberOfArgsFixtureWasCreatedWith)
        {
            for (int i = 0; i < numberOfArgsFixtureWasCreatedWith; i++)
            {
                yield return new TestCaseData(i, numberOfArgsFixtureWasCreatedWith)
                    .SetArgDisplayNames($"Number of args : {i + 1} of {numberOfArgsFixtureWasCreatedWith} ");
            }
        }

```

In this case the `TextFixtureArgumentRef.Arg0` gets replaced with respective test fixture argument before the `TestCaseDataWithFixtureContext` method is called, which allows to generate test cases specific for given fixture.

The `TextFixtureArgumentRef` enumeration allows addressing up to 10 parameters (0-9). IF there is a need for more one can cast any integer value like `(TextFixtureArgumentRef) 22`  which will simply get replaced by value of fixture argument at index 22.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
@@ -35,6 +35,15 @@ using NUnit.Framework.Internal.Builders;
 
 namespace NUnit.Framework
 {
+    public enum TestFixtureArgumentPlaceholder
+    {
+        Arg0,
+        Arg1,
+        Arg2,
+        Arg3
+    }
+
+
     /// <summary>
     /// Indicates the source to be used to provide test fixture instances for a test class.
     /// </summary>
@@ -135,7 +144,7 @@ namespace NUnit.Framework
         {
             int count = 0;
 
-            foreach (TestCaseParameters parms in GetTestCasesFor(method))
+            foreach (TestCaseParameters parms in GetTestCasesFor(method, suite))
             {
                 count++;
                 yield return _builder.BuildTestMethod(method, suite, parms);
@@ -158,13 +167,13 @@ namespace NUnit.Framework
         #region Helper Methods
 
         [SecuritySafeCritical]
-        private IEnumerable<ITestCaseData> GetTestCasesFor(IMethodInfo method)
+        private IEnumerable<ITestCaseData> GetTestCasesFor(IMethodInfo method, Test? suite)
         {
             List<ITestCaseData> data = new List<ITestCaseData>();
 
             try
             {
-                IEnumerable? source = ContextUtils.DoIsolated(() => GetTestCaseSource(method));
+                IEnumerable? source = ContextUtils.DoIsolated(() => GetTestCaseSource(method, suite));
 
                 if (source != null)
                 {
@@ -235,7 +244,7 @@ namespace NUnit.Framework
             return data;
         }
 
-        private IEnumerable? GetTestCaseSource(IMethodInfo method)
+        private IEnumerable? GetTestCaseSource(IMethodInfo method, Test? suite)
         {
             Type sourceType = SourceType ?? method.TypeInfo.Type;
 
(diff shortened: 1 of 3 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=3593) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=3593) before we can accept your contribution.<br/><br/>:x: skomialek [sign now](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=3593)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/nunit/nunit?pullRequest=3593) it.</sub>

skomialek:
Hi @rprouse / @jnm2 / @mikkelbu  ,

I'm not sure If I'm missing something in the process on how to contribute to NUnit, but this PR has been sitting here for a while.

Any input here would be appreciated.

rprouse:
Sorry @skomialek this is totally on us. Everyone on the team has been focused on our day jobs and haven't been as involved as we need to be. We'll do our best to give this the review it deserves as soon as we can. Please bear with us and thank you for your contribution.

skomialek:
Thank you @rprouse  for the update. I completely understand. I just wanted to make sure that I have not forgotten to press any button or follow some process before/after submitting the PR. As long as everything is in place, there is no rush - the PR is not going anywhere. 

PS. I'm not sure I like the sound of "give this the review it deserves" :D

MagnumOpusX:
Any idea when this feature will be available in nunit. I have similar requirement and this would help a lot.

rprouse:
@skomialek I admit that I've been avoiding this PR. The syntax of this doesn't feel right to me but I'm having trouble putting why it makes me feel that way into words and I also don't have alternatives. We need to be very careful when accepting new features because we need to continue to support them in the future. We have several features in the framework that I regret accepting. They confuse users and we end up having to spend too much time answering questions or correcting misconceptions.

I can say that an arbitrary limit of 10 parameters called `Arg0` to `Arg9` seems very unintuitive to me 😄 

Can you give me a more concrete example of the problem this solves for you? I can see it in the abstract, but I'd like more of a real world example. I'd also like to see an idea of how you would propose we document this feature so that users understand it and it is clear what problem it will solve for them.

@nunit/framework-team does anyone else have opinions on, or better syntax ideas for this feature?

skomialek:
Thank you for the feedback. I completely understand feature management problem. 

Let me start with the problem explanation which in my case is simple test organization. 
I generate tests dynamically for the chain of builds that verify dependency for each of them.

So I do have multi-layer structure where for each build I need to validate bunch of registered dependencies.

```cs
foreach(var buildId in buildIds)
{
    foreach(var dependencyId in GetDependencies(buildId)
    {
        Assert.IsTrue(IsValid(dependencyId));
    } 
}
```

I do have multiple options here:

- one test iterating all the cases - this is the least favorable option as the first run will not report all failures making the turnaround time for fixing problems very long.
- using test case source - this is much better - I get one test per dependency, but because of the volume, it is becoming a very long, flat list... it is also hard to search / filter.
- using fixture of every build and then generate tests for every dependency using test data source. This is the scenario that this code enables me to do - generate tests in context of argument passed to the fixture generation. I'm getting nicely organized results grouped by build (one fixture per build) and then have specific tests generated for every dependency to be verified.

This way I get nice filtering, reporting and overview of the tests :) 

The use can be generalized to any two (or more) layer hierarchy (enumerating folders and files, databases and tables etc.).
Some sample scenario descriptions from others:

https://github.com/nunit/nunit/issues/2950#issue-347934665
https://github.com/nunit/nunit/issues/2707#issuecomment-421885014

You can also refer to the sample from this comment : https://github.com/nunit/nunit/issues/2950#issuecomment-662011712 

About the implementation of `Arg0` ... `Arg9` - it actually does support any number of arguments by casting int to the enum type, but I agree it might feel awkward. 

I wast trying to get a very explicit declaration of intent here and not to work with 'magic index' approach like negative numbers etc. Unfortunately, because we are in the context of attribute the nicer approach of having dedicated class : 

```cs
public class FixtureArgument
{
    public int Index { get; }

    public FixtureArgument(int index)
    {
    }

    public static FixtureArgument At(int index)
    {
        return new FixtureArgument(index);
    }
}
```

Will not work because the only things we can pass to the attribute constructor is constant expressions which can be evaluated at compile time.

image

So the solution is driven by language limitation, but if there is anything that could potentially improve this, I'm all in. Enum based solution is where my C# superpowers ended :)

Dreamescaper:
Just an idea. We could add some attribute like TestFixtureSourceArgumentsAttribute, which could be applied to method's `object[]` parameter.


So it would look something like that:
```csharp
        [TestCaseSource(nameof(TestCaseDataWithFixtureContext))]
        public void FixtureArgumentTest(int argumentIndex, int numberOfArguments)
        {
        }

        public static IEnumerable<TestCaseData> TestCaseDataWithFixtureContext([TestFixtureSourceArguments] fixtureArguments)
        {
            int numberOfArgsFixtureWasCreatedWith = (int)fixtureArguments[0];

            for (int i = 0; i < numberOfArgsFixtureWasCreatedWith; i++)
            {
                yield return new TestCaseData(i, numberOfArgsFixtureWasCreatedWith)
                    .SetArgDisplayNames($"Number of args : {i + 1} of {numberOfArgsFixtureWasCreatedWith} ");
            }
        }
```

Wdyt?

skomialek:
@Dreamescaper  , this is definitely an option that opens more door and I must say I like the approach, One could even take it to the level of using attributes to point to specific argument

```cs
        [TestCaseSource(nameof(TestCaseDataWithFixtureContext))]
        public void FixtureArgumentTest(int argumentIndex, int numberOfArguments)
        {
        }

        public static IEnumerable<TestCaseData> TestCaseDataWithFixtureContext([TestFixtureSourceArgument("argumentIndex")] int argumentIndex, [TestFixtureSourceArgument] numberOfArguments)
        {
            int numberOfArgsFixtureWasCreatedWith = numberOfArguments;

            for (int i = 0; i < numberOfArgsFixtureWasCreatedWith; i++)
            {
                yield return new TestCaseData(i, numberOfArgsFixtureWasCreatedWith)
                    .SetArgDisplayNames($"Number of args : {i + 1} of {numberOfArgsFixtureWasCreatedWith} ");
            }
        }
```

Where the default constructor can just match arguments by name (if named the same). That would eliminate the need for user addressing parameters by their index (which is also the downside in my suggestion that I'm not really happy with). 

The reason I selected  the path of indexed access is already exiting feature of test case source with arguments, which passes parameters array in the attribute constructor. 

I have seen advantage of being able to 'inherit' some parameter yet still being able to use test case sources with specific parameters. This way it becomes a part of already established method rather than another way of doing similar thing. Since NUnit is a framework, I believe flexibility of mixing the two approaches might also solve some potential scenarios and I don't really see a reason to force user to choose either you use feature A or B. Obviously, this is my personal opinion.

PR #3602, opened 2020-07-24
Issue 3452

### LINKED ISSUE #3452: Assertions that use an existing Regex
As far as I can tell, there are no assertions, in either model, that allow the use of an existing Regex in the assertion.

All of the Regex assertions take a string containing the regex pattern.

It would be quite useful to have constraints which take an existing Regex in place of the string pattern, but function identically.

Poimen:
Fixes #3452

### DIFF
--- a/src/NUnitFramework/framework/Constraints/RegexConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/RegexConstraint.cs
@@ -29,28 +29,67 @@ namespace NUnit.Framework.Constraints
     /// RegexConstraint can test whether a string matches
     /// the pattern provided.
     /// </summary>
-    public class RegexConstraint : StringConstraint
+    public class RegexConstraint : Constraint
     {
+        private Regex _regex;
+
+        /// <summary>
+        /// The Description of what this constraint tests, for
+        /// use in messages and in the ConstraintResult.
+        /// </summary>
+        public override string Description
+        {
+            get
+            {
+                var description = $"String matching pattern \"{_regex}\"";
+                var caseInsensitive = (_regex.Options & RegexOptions.IgnoreCase) == RegexOptions.IgnoreCase;
+                if (caseInsensitive)
+                {
+                    description += ", ignoring case";
+                }
+                return description;
+            }
+        }
+
         /// <summary>
         /// Initializes a new instance of the <see cref="RegexConstraint"/> class.
         /// </summary>
         /// <param name="pattern">The pattern.</param>
-        public RegexConstraint(string pattern) : base(pattern) 
+        public RegexConstraint(string pattern) : base(pattern)
         {
-            descriptionText = "String matching";
+            _regex = new Regex(pattern);
         }
 
         /// <summary>
-        /// Test whether the constraint is satisfied by a given value
+        /// Initializes a new instance of the <see cref="RegexConstraint"/> class.
         /// </summary>
-        /// <param name="actual">The value to be tested</param>
-        /// <returns>True for success, false for failure</returns>
-        protected override bool Matches(string actual)
+        /// <param name="regex">The Regex pattern object.</param>
+        public RegexConstraint(Regex regex) : base(regex.ToString())
         {
-            return actual != null && Regex.IsMatch(
-                    actual,
-                    this.expected,
-                    this.caseInsensitive ? RegexOptions.IgnoreCase : RegexOptions.None);
+            _regex = regex;
+        }
+
+        /// <summary>
+        /// Modify the constraint to ignore case in matching.
+        /// </summary>
+        public RegexConstraint IgnoreCase
+        {
+            get
+            {
+                _regex = new Regex(_regex.ToString(), _regex.Options | RegexOptions.IgnoreCase);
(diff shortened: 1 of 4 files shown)


CharliePoole:
@Poimen 
The CI tests uncovered the backward-compatibility issue I mentioned. My tc-lite implementation did not need to support the `IgnoreCase` modifier, because I relied on `RegexOptions` for that purpose. An NUnit implementation has to keep those tests passing or wait till V4 when it's OK to have breaking changes.

BTW, It's a good idea to run the CI  tests locally, from the command-line, before doing a PR.

```
build -t Test
```

The first time you do it, you'll probably have to do a bit of work to ensure that all the right platforms are installed.

Poimen:
Thanks - I have never had much success with that script and Rider (for whatever reason) crashes Gnome if I run all the tests :see_no_evil: 

Still, agreed, that tests should be run locally before pushing to the CI. Hopefully that test should be fixed now.

CharliePoole:
What about existing users with...

```c#
Assert.That(phrase, Does.Match("Make.*tests.*pass").IgnoreCase);
```

Since the CI is passing, there is apparently no current test of that syntax.

Poimen:
At this point I want to shoot the poor CI :rofl: 

It is failing on:
```
2020-07-24T16:55:36.6396506Z 1) Failed : NUnit.Framework.Constraints.RegexConstraintTests.RegExCaseMatchFails
2020-07-24T16:55:36.6397464Z   Expected: <NUnit.Framework.AssertionException> and property Message equal to "  Expected: String matching pattern "make.*tests.*fail"
2020-07-24T16:55:36.6398161Z   But was:  "Make your tests fail before passing!"
2020-07-24T16:55:36.6398549Z "
2020-07-24T16:55:36.6398960Z   But was:  "  Expected: String matching "make.*tests.*fail"
2020-07-24T16:55:36.6399491Z   But was:  "Make your tests fail before passing!"
2020-07-24T16:55:36.6399867Z "
2020-07-24T16:55:36.6400463Z    at NUnit.Framework.Constraints.RegexConstraintTests.RegExCaseMatchFails() in /home/vsts/work/1/s/src/NUnitFramework/tests/Constraints/RegexConstraintTests.cs:line 79
2020-07-24T16:55:36.6401047Z 
```
Which I'm guessing is some new line character issue as this test run fine locally (on Ubuntu 18.04). The assertion is running against:
```csharp
const string expectedErrorMessage = "  Expected: String matching pattern \"make.*tests.*fail\"\n  But was:  \"Make your tests fail before passing!\"\n";
```

Are these exception message handling tests required?

@CharliePoole for your question:
```csharp
Assert.That(phrase, Does.Match("Make.*tests.*pass").IgnoreCase);
```
is still retained as a use case.

I have added new test into the mix with checking both versions:
```csharp
const string testPhrase = "Make your tests fail before passing!";

Assert.That(testPhrase, Does.Match(testMatcher).IgnoreCase);
Assert.That(testPhrase, Does.Match(new Regex(testMatcher)).IgnoreCase);
```
So this should cover both use cases?

CharliePoole:
You can't test messages with platform-specific line termination. I usually define `string NL = Environment.NewLine` and interpolate that variable in the expected messages. (To tell the truth, I often forget but get reminded when I switch from one OS  to the other.)

It looks like the message is being generated for the subordinate constraint rather than for the full expression. I would check and see how the other constraints handle this and make `RegexConstraint` consistent.

CharliePoole:
@Poimen I don't think you can use `Assert.That` for the outer assertion in this context. It may be a bug. Try using

```c#
var ex = Assert.Throws<AssertionException> (() Assert.That...etc. );
Assert.That(ex.Message, etc. );
```

Poimen:
I have pushed up the changes as per the suggestions :+1:

CharliePoole:
Nice! Looks like that does it. Testing error messages is really tedious, isn't it? :smiling_imp: 

If you're ready, you should remove the WIP from the issue title. Then the merge has to be approved by two team members.

Poimen:
Oh, yeah. Testing error messages...my favourite past time... :rofl:

Poimen:
@CharliePoole Thanks for the support :+1:

src/NUnitFramework/framework/Constraints/RegexConstraint.cs
```diff
@@ -29,28 +29,67 @@ namespace NUnit.Framework.Constraints
...
+        /// <param name="actual">The string to be tested.</param>
+        /// <returns>True for success, false for failure.</returns>
+        public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
-            return actual != null && Regex.IsMatch(
-                    actual,
-                    this.expected,
-                    this.caseInsensitive ? RegexOptions.IgnoreCase : RegexOptions.None);
+            return new ConstraintResult(this, actual,
+                actual != null && actual is string && _regex.Match(actual.ToString()).Success);
```
Dreamescaper, line 92:
```suggestion
                actual is string actualString  && _regex.Match(actualString).Success);
```

(`is string` handles null case as well)

Issue #3589, opened 2020-06-29
File headers, copyrights, and licenses


jnm2:
I've heard a mix of things that I'm still trying to clarify:
- Year ranges are extraneous and possibly invalidate the copyright notice.
- A year in a file should never be updated or it possibly invalidates the copyright notice.
- A starting year that is later than the entire project's initial publish, even for new files, possibly invalidates the copyright notice.
- A license notice is not required to be in source files in order for the license to cover those files.

Things I'm pretty sure of from reading the linked https://www.copyright.gov references just now:
- A copyright notice is not required to be in source files in order for the copyright to cover those files.
- `©` is the preferred symbol, but `(c)` is explicitly listed as an acceptable alternative marking.
- `Copyright` and `©` are redundant with each other but it doesn't invalidate the notice to use both.

Things I know for sure:
- Many open-source projects don't have headers.
- Many others use one-liners tell you that the copyright and license info for the file is in the repository root. (Notably, all projects copyright by .NET Foundation.)
- I can't stand scrolling past, adding, or updating file headers, or asking for contributors to do these things in PRs.

It bothers me a lot, especially the not knowing if what we're doing is even legally achieving what we think it is. I'm willing to get solid legal answers to all of this if needed. First, is there any way we could be comfortable with the idea of not using file headers at all? What would it take to get to that point? If that's not possible, can we do the next most light-weight thing which is a single-line header that never changes?

### References:

The GitHub URLs are intentionally within backticks to prevent spamming the threads with crossreferences.

- https://www.copyright.gov/circs/circ03.pdf
- https://www.copyright.gov/comp3/chap2200/ch2200-notice.pdf
- "Use legally-correct copyright notices" `https://github.com/DotNetAnalyzers/StyleCopAnalyzers/pull/1661`
- "File headers: allow pattern matching or multiple allowable headers to support headers that must include a date" `https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/1357`

CharliePoole:
Lots of conflicting info out there, for sure. There are a few good reasons for that, I believe:
 1. Statements are not always from lawyers.
 2. People are from various countries.
 3. Even sticking with lawyers and US law, many of these opinions are predictions of how a court would interpret the law. Lots of copyright stuff as applied to Open Source has actually never been adjudicated and existing case law is the whole basis of our legal system.

So there are no guarantees, each project has to use its best judgement and then stick with a consistent approach.

[...]

PR #3603, opened 2020-07-26
Add missing license in file. Add missing "readonly" modifiers on immutable structs

stevenaw:
A few things I noticed that I missed in a previous PR (#3406 , which addressed #52 )
A new file was added (ComparisonState.cs) but I had forgotten to add the license header at the top of the file. This PR adds it with the copyright date of that changeset (2019).

Additionally, a common theme in the recommended approach to that PR was to make the structs immutable, however the `readonly` keyword on the struct itself was omitted. This PR adds that in to ensure that the types _remain_ immutable by default, even if new members are added.

NOTE: I ran into some issues when trying to make the `ImmutableStruct` from that PR `readonly`: https://github.com/nunit/nunit/issues/3604


stevenaw:
File header relevant to discussion in #3589

PR #3609, opened 2020-07-30
Issue 2211

### LINKED ISSUE #2211: Request: Add support of indexers to the PropertyConstraint
At the moment it is only possible to test values of normal properties and no way to test indexer values. I need to write a test that looks something like this:

```c#
  [TestFixture]
  public class TestClass
  {
    class Example
    {
      private Dictionary<string, string> dictionary = new Dictionary<string, string>();

      public string this[string key]
      {
        get { return dictionary[key]; }
        set { dictionary[key] = value; }
      }
    }

    [Test]
    public void Test()
    {
      var obj = new Example
      {
        ["TEST1"] = "1",
        ["TEST2"] = "2",
      };

      Assert.That(obj, Has.Property("Item", "TEST1").EqualTo("1").And.Property("Item", "TEST2").EqualTo("2"));
    }
  }
```

Poimen:
Fixes #2211


rprouse (dismissed):
Looks good, thanks for the help.

src/NUnitFramework/framework/Constraints/IndexerConstraint.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// </summary>
+        /// <param name="actual">The value to be tested</param>
+        public override ConstraintResult ApplyTo<TActual>(TActual actual)
+        {
+            Guard.ArgumentNotNull(actual, nameof(actual));
+
+            if (!(actual is Type actualType))
+            {
+                actualType = actual.GetType();
+            }
```
jnm2, line 65: Since you're rewriting it from what PropertyConstraint has, `as`+coalesce is an often-overlooked way to simplify:

```suggestion
            var actualType = actual as Type ?? actual.GetType();
```

(same in both files)

src/NUnitFramework/framework/Constraints/IndexerConstraint.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            var indexerInfo = Reflect.GetIndexerByName(actualType, _argumentTypes);
+            if (indexerInfo?.GetGetMethod() == null)
+            {
+                throw new ArgumentException($"Indexer with {StringifyArguments()} was not found on {actualType}.");
+            }
```
jnm2, line 71: Applying the same logic:

```suggestion
            var getter = Reflect.GetIndexerByName(actualType, _argumentTypes)?.GetGetMethod()
                ?? throw new ArgumentException($"Indexer with {StringifyArguments()} was not found on {actualType}.");
```

(same in the other file, minus the `throw` part)

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -463,5 +463,32 @@ internal static bool IsVoidOrUnit(Type type)
...
+
+        /// <summary>
+        /// Returns the property information for a given named indexer.
+        /// </summary>
+        /// <param name="type">Type to reflect on for indexer.</param>
+        /// <param name="indexerTypes">List of indexer types that matches the indexer type order.</param>
+        /// <returns>The return value from the invoked method</returns>
+        public static PropertyInfo? GetIndexerByName(Type type, Type[] indexerTypes)
+        {
+            const string defaultFrameworkIndexerName = "Item";
```
jnm2, line 475: The default indexer name varies. It's `"Item"` when there is no DefaultMemberAttribute applied to the class.

You can write a test for this by putting `[IndexerName("NonDefaultIndexerName")]` on a C# indexer. The C# compiler removes this attribute during emit and emits `[DefaultMember("NonDefaultIndexerName")]` on the class instead.

([sharplab link](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAGAAkIEYA6AJQFcA7DbAWxioGEIWAHPGKAGUBAN2xgYAZwDcAWABQCwgCZSHBQG8FpHaQDaASQYATGAgEA5AIZsAFACILEBgBEYAMyt1cGI6fNQ1mz2AJQAutq6hADMpNhMpBgAFtiSevEYpAhhpAC8AHxZcvIAvkA))

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -463,5 +463,32 @@ internal static bool IsVoidOrUnit(Type type)
...
+
+            var props = type.GetProperties(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance).ToArray();
+
+            var indexerInfo = props.FirstOrDefault(p => p.Name == defaultFrameworkIndexerName && ParametersEqual(p.GetIndexParameters(), indexerTypes));
+            return indexerInfo;
```
jnm2, line 480: Did you consider using using `type.GetProperty(indexerName, bindingFlags, binder: null, returnType: null, indexerTypes, modifiers: null)`?

jnm2: It would probably be best to use the same shadowing behavior that PropertyConstraint uses, eventually, but I don't think this is needed immediately.

src/NUnitFramework/framework/Constraints/IndexerConstraint.cs
Poimen: I had something similar - reversed in commit 96154ca6402c692cec5e0ae6947f9922e72799fe. The build failed for earlier framework versions so went with the longer version.

Please confirm you would like the simplification and I'll see if I can dig up those errors?

src/NUnitFramework/framework/Internal/Reflect.cs
Poimen: I had known about this, but with this comment:
https://github.com/nunit/nunit/issues/2211#issuecomment-305782721

I can't remember why I moved away from handling that condition. I'll have to look back into what was going on there. Let me check my reference test and get back on this.

src/NUnitFramework/framework/Internal/Reflect.cs
Poimen: Yes, it was for multiple indexers that have similar parameters. It was returning the incorrect indexer (or throwing) as far as I can recall at this time. I'll have to have another look into this and confirm.

src/NUnitFramework/framework/Constraints/IndexerConstraint.cs
jnm2: I have only a slight preference, so don't put too much effort into it. I'm not spotting anything that wouldn't work back to .NET Framework 1.0.

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: Oh, I'm sorry for the confusion in terms. https://github.com/nunit/nunit/issues/2211#issuecomment-305782721 refers to default versus named *indexers*. A default indexer looks like `obj[args]`. A named indexer is not a concept in C# but it would look like `obj.SomeProperty[args]`. In VB, the default indexer is declared with the keyword `Default` rather than being given a name. We decided to only implement assertions on default indexers.

My comment here in this PR is on a different, independent topic. This topic is the *name* of a default indexer. The default name of a default indexer is `Item`, but you can declare a *default* indexer with a *non-default* name. This isn't something we should just choose to ignore; it's an actual correctness issue that would lead to bad behavior if we ignore the DefaultMember attribute.

src/NUnitFramework/framework/Constraints/IndexerConstraint.cs
jnm2: I spotted what was framework-dependent in the commit you mention: your use of the `PropertyInfo.GetMethod` property which first appears in .NET Framework 4.5. That's not contained in my suggestions in this PR.

Poimen: :man_facepalming:  Not sure what I was thinking - I'm all for simplification. I'll push up the changes.

src/NUnitFramework/framework/Internal/Reflect.cs
Poimen: So I've gone back to the tests. The `type.GetProperty` throws:
```
System.Reflection.AmbiguousMatchException
```
when used on a class that has something like:
```c#
public string this[int x] => // ...
public string this[string x] => // ...
public string this[int x, int y] => // ..
```

So in the event of multiple indexers in a class, the `GetProperties` needs to be iterated through.

Poimen: Ok, not sure what I did wrong, but this is working now..as you would expect.

src/NUnitFramework/framework/Constraints/IndexerConstraint.cs
Poimen: Pushed up simplified implementation

src/NUnitFramework/framework/Internal/Reflect.cs
Poimen: I've added some more tests and implementation around the handling `[IndexerName("NonDefaultIndexerName")]`

I've changed the reflection code now to return the `MethodInfo` rather than the `PropertyInfo`. The method now attempts get default indexer first, otherwise it goes into a property search.

src/NUnitFramework/framework/Internal/Reflect.cs
Poimen: So this was interesting:
> It would probably be best to use the same shadowing behavior that PropertyConstraint uses

I had expected that it would also need this. However, I wrote some tests around shadow indexers and they behave a little differently. It looks to be pulling through the shadow indexers fine. 

If you could have a look at the tests in case I have missed something there.

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -463,5 +463,43 @@ internal static bool IsVoidOrUnit(Type type)
...
+        private static PropertyInfo? GetDefaultIndexer(Type type, Type[] indexerTypes, BindingFlags bindingFlags)
+        {
+            const string defaultFrameworkIndexerName = "Item";
+            return type.GetProperty(defaultFrameworkIndexerName, bindingFlags, null, null, indexerTypes, null);
+        }
+
+        private static PropertyInfo? GetNamedIndexer(Type type, Type[] indexerTypes, BindingFlags bindingFlags)
+        {
+            var props = type.GetProperties(bindingFlags).ToArray();
+            return props.FirstOrDefault(p => ParametersEqual(p.GetIndexParameters(), indexerTypes));
```
jnm2, line 491: This would not be correct. This would find `non-default indexers`. The correct way to find the *default indexer* that has a *non-default name* would be to check for `DefaultMemberAttribute` on `type`.

If `DefaultMemberAttribute` is present, we should not search for an indexer with a metadata name of `"Item"`. Instead, we should search for the member name specified by the DefaultMemberAttribute.

jnm2 (changes requested):

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: > Ok, not sure what I did wrong, but this is working now..as you would expect.

I think what is different is that this time you are using an overload of GetProperty that specifies the argument types.

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: There's a difference between each of these things:

- A default indexer with the default name ("Item"). Because this is a default indexer, you can call it in C# or VB without using the property name. No `DefaultMemberAttribute` is present.
- A default indexer with a non-default name (specified in C# using `IndexerNameAttribute` on the default indexer). Because this is a default indexer, you can call it in C# or VB without using the property name. A `DefaultMemberAttribute` is present.
- A so-called *named indexer*. Because this is not a default indexer, you *can't* call it in C# or VB without using its metadata name. No `DefaultMemberAttribute` is present.

jnm2: I can give code examples of the three things if it helps.

jnm2: Therefore we should have a method called `Reflect.GetDefaultIndexer` and no method referring to named indexers. It should do one search for the default indexer using the name it got from DefaultMemberAttribute, or using the name `Item` only if there is no DefaultMemberAttribute on the declaring type.

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: > If you could have a look at the tests in case I have missed something there.

This isn't quite what I was talking about. I was talking about mirroring what PropertyConstraint does, doing something analogous at each step where applicable. PropertyConstraint uses Reflect.GetUltimateShadowingProperty. Here are the tests that show why a method like that was needed: https://github.com/nunit/nunit/pull/2423/files#diff-246abb5484e1277f022d8d83ba9a6439dcc5d992f5d473f84e43c0a1c503de62

I don't think this is something we need to deal with in this PR, but I would expect that we would have the same thing happen with indexers that is shown in those tests. The fact that `PropertyConstraint` calls `Reflect.GetUltimateShadowingProperty` is a strong hint that `IndexerConstraint` should be calling a method called `Reflect.GetUltimateShadowingDefaultIndexer`.

Again, let's worry about this in a separate PR because this PR is complex enough, and I think the chance of hitting the bug is so tiny.

src/NUnitFramework/framework/Internal/Reflect.cs
Poimen: Ok, so I misunderstood the _ignoring_ of 3 ... not that I realised 3 was a real thing. 

Could you please clarify your last statement -
> It should do one search for the default indexer using the name it got from DefaultMemberAttribute

Is there any preference on searching default `Item` before `DefaultMemberAttribute` or the other way round?

It is just a point of wondering if there is a critical difference in the direction of the search. I would expect (in terms of user cases) having a default overload using `Item` would be far more common, thus search for it first?

src/NUnitFramework/framework/Internal/Reflect.cs
Poimen: Absolutely - I was expecting the need for `GetUltimateShadowingDefaultIndexer`.

So the added the tests to do the shadowing (not sure the tests are even correct) should have failed - they are similar to the property shadow tests. But I suspect the error you highlighted with the default/non-default indexer may be playing a role here.

If this does cause a problem once the other issue is resolved, then I'll remove the tests - but, agreed, not going to push deeply into this.

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: Specifically, there is one search, one name, that can possibly be a default indexer for a given class. An indexer named `Item` is not a default indexer if `DefaultMemberAttribute` specifies another name. Such an indexer falls into category 3, an indexer whose name must be given when you call it. Not having to use the name when you call it is what makes a thing a default indexer rather than a named indexer.

So it would be wrong to use `Item` if DefaultMemberAttribute specifies another name.

Poimen: Pushed up the implementation to go grab the `DefaultMemberAttribute` method name off the type provided. This now attempts to get the attribute, if not available, defaults back to `Item`. 

Using that name - a search is done using `type.GetProperty(name, ...)`

Poimen:
Applied fixes based on review feedback :+1:

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -463,5 +463,29 @@ internal static bool IsVoidOrUnit(Type type)
...
+
+        /// <summary>
+        /// Returns the invokable method for a given named indexer.
+        /// </summary>
+        /// <param name="type">Type to reflect on for indexer.</param>
+        /// <param name="indexerTypes">List of indexer types that matches the indexer type order.</param>
+        /// <returns>Invokable indexer method</returns>
+        public static MethodInfo? GetDefaultIndexer(Type type, Type[] indexerTypes)
+        {
+            const BindingFlags bindingFlags = BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance;
```
jnm2, line 475: Completely off topic, but one of the things being considered for future C# is the ability to target-type the name lookup so that you can do `type.GetProperty(indexerName, NonPublic | Public | Instance)` etc.

src/NUnitFramework/tests/Constraints/IndexerConstraintTests.cs
```diff
@@ -0,0 +1,177 @@
...
+            public int DummyParam { get; set; }
+
+            public string this[int x] => x == 42 ? "Answer to the Ultimate Question of Life, the Universe, and Everything" : "Still calculating";
+
+            public string this[string x] => "Second indexer";
+
+            protected string this[int x, int y] => "Third indexer";
+        }
+
+        private class ClassWithShadowIndexer : IndexerTester
```
jnm2, line 129: Nitpick: The `new` member in the derived class is usually called the *shadowing* or *hiding* indexer and the base member is the *shadowed* or *hidden* member.

Also, I only recently learned that the C# lingo is primarily 'hide,' not 'shadow' which is terminology I guess I carried over from my VB days a decade ago. https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/new-modifier

jnm2 (approved):
This is incredibly thorough. Thank you for your efforts and patience!

src/NUnitFramework/tests/Constraints/IndexerConstraintTests.cs
Poimen: Oh, yes :man_facepalming: I even read the error message:
```
because it hides inherited member
```
and didn't twig :rofl:

Poimen:
Updated those minor aspects :+1:

Dreamescaper (commented):
Could you add tests for negative asserts?
e.g.
`Assert.That(tester, Has.No.Item(typeof(object)))`
`Assert.That(tester, Has.No.Item(6 * 8).EqualTo(42))`


Negative asserts are a bit strange though. E.g. what should be the result of the following assert?
`Assert.That(new[] {1,2,2}, Has.No.Item(3));`
I mean, it has not item with an index of 1...

Besides, I'd say that `Has.Item` is still very confusable with `Contains.Item`.
@jnm2 
I saw that you mentioned in the issue in favor of `Has.Index`, any reason it was decided to go with `Has.Item` ?

src/NUnitFramework/tests/Constraints/IndexerConstraintTests.cs
```diff
@@ -0,0 +1,177 @@
...
+            var tester = new DerivedClassWithoutNamedIndexer();
+
+            Assert.That(tester, Has.Item(42));
+            Assert.That(tester, Has.Item(42).EqualTo("A Named Int Indexer from shadow"));
+            Assert.That(tester, Has.Item(42, 43).EqualTo("A Named Int Int Indexer"));
+        }
+
+        private class IndexerTester
+        {
+            public int DummyParam { get; set; }
```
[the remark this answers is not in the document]
Poimen: While playing with the reflection code, those where to ensure it was searching for indexers only and not picking up other fields. Since then the search has been improved as not to be "implement" a coded search pattern. Hence, we could remove these if you would like?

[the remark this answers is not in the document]
Dreamescaper: Got that!
No, it makes sense, I suppose.

jnm2:
@Dreamescaper That's a good question. I don't remember my thought process 2.5 years ago and didn't explain why I started preferring Has.Item at https://github.com/nunit/nunit/issues/2211#issuecomment-386774663.

If I was to invent a reason, I think you would say an instance *has* an item *at* an index, not that an instance *has* an index.

What do you think about this?

Dreamescaper:
@jnm2 
Oops, haven't noticed that it was 2.5 years ago.
That makes sense, although still quite confusing :).

Anyway, apart from that, LGTM.

jnm2:
It's good you brought it up. I share your concern now/again. People will do `Assert.That(collection, Has.Item("x"));` and not be too happy to sort through the resulting message that no indexer accepting `"x"` could be found.

Going to move this quick to the discussion issue and hopefully get a quick resolution.

jnm2:
I was in the middle of typing `Has.ItemAtIndex` as an example of an off-the-wall suggestion and suddenly realized that I like `Has.ItemAt` very much.

@Poimen Heads up that whatever is decided at https://github.com/nunit/nunit/issues/2211#issuecomment-741274964 could cause a simple rename operation to be needed in all the places where you wrote `Item`. Hopefully that seems pretty easy to do in the IDE.

Poimen:
@jnm2 `Ctrl+R+R` was invented for days like this :rofl:

jnm2 (changes requested):
Per https://github.com/nunit/nunit/issues/2211#issuecomment-741818483, we've decided to go with `ItemAt`. Changing all applicable occurrences of `Item` to `ItemAt` is the only thing I see that is needed before merging. Thanks for all your help so far!

Poimen:
@jnm2 @Dreamescaper In terms of negative testing - I'm still updating for it. 

For error messaging, the property constraint gives the message:
```
  Expected: not property DummyParam
  But was:  <NUnit.Framework.Constraints.IndexerConstraintTests+IndexerTester>
```

For the indexer:
```c#
Assert.That(tester, Has.No.Item("wow"));
```
gives:
```
  Expected: not Default indexer accepting arguments < "wow" >
  But was:  <NUnit.Framework.Constraints.IndexerConstraintTests+IndexerTester>
```
I haven't given any effort to customising the `Expected: not Default indexer` message as this might be part of the expectation, etc.

I am assuming this is expected?

jnm2:
@Poimen Yes, that's expected. We don't customize other such combinations.

Poimen:
Added some negative tests and rename to `ItemAt` :+1:

Dreamescaper:
@Poimen 
I'm trying to understand how this test works :)
```csharp
Assert.That(new[] { 1, 2, 2 }, Has.No.ItemAt(3));
```

It only checks for actual type, not the value, right? So in this case we simply check that int array has no int indexer. But it does, so why it passes?

src/NUnitFramework/tests/Constraints/IndexerConstraintTests.cs
```diff
@@ -106,7 +126,7 @@ public void CanMatchShadowIndexer()
         }
 
         [Test]
-        public void CanMatchShadowNamedIndexer()
+        public void CanMatchHiddenNamedIndexer()
```
jnm2, line 129: These two tests are demonstrating matching the *hiding* members, not the *hidden* members, right?

jnm2 (approved):

src/NUnitFramework/tests/Constraints/IndexerConstraintTests.cs
Poimen: errr, yes :man_facepalming:

jnm2:
Whether positive or negative, this does like it would be checking the count rather than checking for the presence of an indexer:

```cs
Assert.That(new[] { 1, 2, 2 }, Has.ItemAt(3));
Assert.That(new[] { 1, 2, 2 }, Has.No.ItemAt(3));
```

Perhaps we should remove the Exists constraint until we have more time to think about it. Maybe we'd want `Has.No.Indexer<T1, T2, ...>()` for that.

Poimen:
Oops...left that one in there... :innocent: 

The reason it passes is that even:
```c#
Assert.That(new[] { 1, 2, 2 }, Has.ItemAt(0));
```
will fail.

This is because the property `Item` is not actually present on the type `System.Int32[]`. `IEnumerable` versions passes when used like:
```c#
Assert.That(new[] { 1, 2, 2 }.ToList(), Has.ItemAt(3));
```

However, more interestingly, this fails (which I hadn't considered before):
```c#
Assert.That(new[] { 1, 2, 2 }.ToList(), Has.No.ItemAt(3).EqualTo(345));
```

with:
```
NUnit.Framework.Internal.NUnitException : Rethrown
  ----> System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
```
I'm not sure if when used outside the framework if the `.No.` would catch this exception or not.


So, to address the first issue - we reflect on the type then grab the `Item` property. On `new[]{...}` the type is `System.Int32[]` which does not have an `Item` property, so it fails incorrectly.

I am contemplating that the indexer exists checker, in this case, would need to check if the `actualType.IsArray` and then check against the `actualType.GetElementType()` for the argument type match. Also, then match a single argument. That should resolve the test mismatch?

For the index out of range exception, I'm not totally sure of how the `.No.` is intended to function. Was this meant to be a catch to exception thrown?

I am leaning towards this being an "indexer" style exception, so when it attempts to grab the value:
```c#
var indexReturnedValue = Reflect.InvokeMethod(getMethod, actual, _arguments);
return BaseConstraint.ApplyTo(indexReturnedValue);
```
place a `try/catch` around that (Also noting that `actualType` in this instance above would also need the same treatment as exists) - however, at which you would return a `ConstraintResult` for the index out of range exception?

src/NUnitFramework/tests/Constraints/IndexerConstraintTests.cs
Poimen: This does sound strange - `CanMatchHidingNamedIndexer`. Does this make sense?

jnm2:
`.No.` would not catch an exception, and that's a good thing because the constraint model only throws exceptions if it's buggy or if the user passed an invalid argument which a constraint could not be created from.

If the indexer throws an exception, we should behave the same as PropertyConstraint if a property throws an exception. I don't think we handle it IIRC. It's not what is being asserted by the constraint; the `Throws` etc constraints are the only ones that are user-exception-aware.

I think we'll have to special-case arrays since they are a special case in IL. Instead of emitting `callvirt` to an indexer, the compiler emits `ldelem` for arrays. If the `actual` type is found to be assignable to `Array`, we could switch the actual type to `System.Collections.IList` to allow us to still use an indexer get accessor to obtain the array element.

jnm2 (dismissed):
Given the three thumbs up on https://github.com/nunit/nunit/issues/2211#issuecomment-742587300, let's make `ItemAt` stop returning a resolvable constraint expression so that this does not compile:

```cs
Assert.That(actual, Has.ItemAt(42));
```

But this does:

```cs
Assert.That(actual, Has.ItemAt(42).EqualTo(43));
```

Let us know if you need any help, as always!

src/NUnitFramework/framework/Internal/Reflect.cs
jnm2: Perfect!

jnm2:
I filed a separate issue (https://github.com/nunit/nunit/issues/3690) pending design approval for a `Has.Indexer` API to bring IndexerExistsConstraint into existence. There are no expectations on you for that one.

Poimen:
I think  understood the new requirement. I had a little fun with the `Array` handling. It should hopefully be implemented correctly. :+1:

src/NUnitFramework/framework/Constraints/IndexerConstraint.cs
```diff
@@ -59,12 +59,22 @@ public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
             Guard.ArgumentNotNull(actual, nameof(actual));
 
+            object indexedValue;
             var actualType = actual as Type ?? actual.GetType();
 
-            var getMethod = Reflect.GetDefaultIndexer(actualType, _argumentTypes) ?? throw new ArgumentException($"Default indexer accepting arguments {MsgUtils.FormatCollection(_arguments)} was not found on {actualType}."); ;
+            if (actualType.IsArray || typeof(Array).IsAssignableFrom(actualType))
```
Dreamescaper, line 65: Do you need a second condition here? Isn't `actualType.IsArray` enough?

src/NUnitFramework/framework/Constraints/IndexerConstraint.cs
```diff
@@ -59,12 +59,22 @@ public override ConstraintResult ApplyTo<TActual>(TActual actual)
...
+            {
+                var array = actual as Array;
+                indexedValue = array?.GetValue((int)_arguments[0]);
```
Dreamescaper, line 68:
```suggestion
                var array = (Array)actual;
                indexedValue = array.GetValue((int)_arguments[0]);
```
I think it is certain at this point that an item should be an array, so I'd prefer it to fail if we're wrong instead of returning null.

Dreamescaper, line 68: Also, it would be great to check array rank and handle multi-dimensional arrays.

Dreamescaper:
Thanks a lot!
I've only left a couple of nitpicks.

src/NUnitFramework/framework/Constraints/IndexerConstraint.cs
Poimen: Sure - we can do that. It'll fail with a cast exception, so was erring on the side of caution. Point taken that it is unlikely to ever happen given the precondition check :+1:

Poimen: Ah, in my haste I missed the API for:
```c#
public object GetValue(params int[] indices);
```
will update :+1:

Poimen: Ah, right, now I remember. Can't do:
```c#
var array = (Array)actual;
```
because:
```
CS0030: Cannot convert type 'TActual' to 'System.Array'
```

Poimen:
Updated as per review :+1:

Dreamescaper (approved):
LGTM

rprouse (approved):
Looks good, thanks for the help.

PR #3620, opened 2020-08-11
Removes PInvoke to GetVersionEx() that causes problems on AOT systems

### LINKED ISSUE #3618: NUnit has a P/Invoke whose native function doesn't exist on all platforms
Repro project: [nunitlite-3.12.0-c582df3.zip](https://github.com/nunit/nunit/files/5056519/nunitlite-3.12.0-c582df3.zip)

To reproduce, download and extract, and then execute `msbuild /p:Platform=iPhone /r`

This results in a few errors:

```
  Undefined symbols for architecture arm64:
    "_GetVersionEx", referenced from:
        wrapper_managed_to_native_NUnit_Framework_Internal_OSPlatform_GetVersionEx_NUnit_Framework_Internal_OSPlatform_OSVERSIONINFOEX_ in nunit.framework.dll.o
  ld: symbol(s) not found for architecture arm64

MTOUCH : error MT5210: Native linking failed, undefined symbol: _GetVersionEx. Please verify that all the necessary frameworks have been referenced and native libraries are properly linked in. [/Users/rolf/Projects/nunitlite-3.12.0/nunitlite-3.12.0.csproj]
```

[Full build log](https://gist.github.com/rolfbjarne/4d9348922057c8ec27cef70c7f98aa20)

This is the P/Invoke in question: https://github.com/nunit/nunit/blob/35242d47ac4669998a0fbf69787914598b0b7664/src/NUnitFramework/framework/Internal/OSPlatform.cs#L201-L202

This is clearly a Windows-only P/Invoke, since `GetVersionEx` is a Windows-only function.

For Xamarin.iOS apps it's highly recommended to not have P/Invokes to functions that don't exist on the platform, because otherwise we have to apply a workaround that has a performance impact on the application.

Would it be possible to rework this code somehow so that the NUnit assemblies that are used for Xamarin.iOS (and tvOS/watchOS) don't have this P/Invoke?

rprouse:
Fixes #3618 

The PInvoke causes problems on iPhone. This switches the .NET Standard build to look in the registry to determine if the OS is a Windows Server or Client OS. This registry entry is valid on modern Windows OS's that support .NET Standard 2.0 versions of .NET.

### DIFF
--- a/src/NUnitFramework/framework/Internal/OSPlatform.cs
+++ b/src/NUnitFramework/framework/Internal/OSPlatform.cs
@@ -38,10 +38,6 @@ namespace NUnit.Framework.Internal
     [SecuritySafeCritical]
     public class OSPlatform
     {
-        readonly PlatformID _platform;
-        readonly Version _version;
-        readonly ProductType _product;
-
         #region Static Members
         private static readonly Lazy<OSPlatform> currentPlatform = new Lazy<OSPlatform> (() =>
         {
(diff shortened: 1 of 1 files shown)


src/NUnitFramework/framework/Internal/OSPlatform.cs
```diff
@@ -217,16 +249,13 @@ public OSPlatform(PlatformID platform, Version version)
...
         }
 
         /// <summary>
         /// Get the platform ID of this instance
         /// </summary>
-        public PlatformID Platform
-        {
-            get { return _platform; }
-        }
+        public PlatformID Platform { get; }
```
rprouse, line 258: Switching these three properties to be non-backed is just a refactor that made it easier for me to see where and how these were used.

src/NUnitFramework/framework/Internal/OSPlatform.cs
```diff
@@ -179,6 +181,35 @@ public enum ProductType
...
+        private static ProductType GetProductType()
+        {
+            try
+            {
+                using (var key = Registry.LocalMachine.OpenSubKey(@"SOFTWARE\Microsoft\Windows NT\CurrentVersion"))
+                {
+                    if (key != null)
+                    {
+                        var installationType = key.GetValue("InstallationType") as string;
+                        switch(installationType)
```
jnm2, line 194: If desired: `switch (key?.GetValue("InstallationType") as string)`

PR #3624, opened 2020-08-20
Equal constraint using comparer

### LINKED ISSUE #1897: EqualTo().Using() prevents caller from comparing strings to anything else
Given a suitable `DateTimeComparer` class that converts `IConvertible` objects to `DateTime`, I expect this code to not throw an assertion:
```
Assert.That(new DateTime(2016, 8, 2), Is.EqualTo("2 August 2016")
                                        .Using(DateTimeComparer.Instance));
```
But it does, because `Using(DateTimeCompare.Instance)`  returns either a `ComparerAdapter` or an `EqualityComparerAdapter`, neither of which override the default implementation of `CanCompare`.

Since the person writing the test is taking responsibility for comparing the actual and expected values (due to calling `Using`), `CanCompare` should always return `true` for any `ExternalComparer` added when calling `Using`.

The original issue and MCVE is described at [Stackoverflow](http://stackoverflow.com/q/38711296/3195526).

manfred-brands:
Fixes #1897

Changes EqualityAdapter.CanCompare if one of the arguments is `string` and the other is not `IEnumerable`.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/EqualityAdapter.cs
+++ b/src/NUnitFramework/framework/Constraints/EqualityAdapter.cs
@@ -49,13 +49,8 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         public virtual bool CanCompare(object x, object y)
         {
-            if (x is string && y is string)
-                return true;
-
-            if (x is IEnumerable || y is IEnumerable)
-                return false;
-
-            return true;
+            return (x is string || !(x is IEnumerable)) &&
+                   (y is string || !(y is IEnumerable));
         }
 
         #region Nested IComparer Adapter
--- a/src/NUnitFramework/tests/Constraints/EqualConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/EqualConstraintTests.cs
@@ -558,6 +558,38 @@ namespace NUnit.Framework.Constraints
             }
 
             [Test]
+            public void UsesProvidedEqualityComparerForExpectedIsString()
+            {
+                var comparer = new ObjectToStringEqualityComparer();
+                Assert.That(4, Is.EqualTo("4").Using(comparer));
+                Assert.That(comparer.WasCalled, "Comparer was not called");
+            }
+
+            [Test]
+            public void UsesProvidedEqualityComparerForActualIsString()
+            {
+                var comparer = new ObjectToStringEqualityComparer();
+                Assert.That("4", Is.EqualTo(4).Using(comparer));
+                Assert.That(comparer.WasCalled, "Comparer was not called");
+            }
+
+            [Test]
+            public void UsesProvidedComparerForExpectedIsString()
+            {
+                var comparer = new ObjectToStringComparer();
+                Assert.That(4, Is.EqualTo("4").Using(comparer));
+                Assert.That(comparer.WasCalled, "Comparer was not called");
+            }
+
+            [Test]
+            public void UsesProvidedComparerForActualIsString()
+            {
+                var comparer = new ObjectToStringComparer();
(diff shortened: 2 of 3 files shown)


src/NUnitFramework/framework/Constraints/EqualityAdapter.cs
```diff
@@ -193,6 +199,11 @@ class EqualityComparerAdapter<T> : GenericEqualityAdapter<T>
         {
             private readonly IEqualityComparer<T> comparer;
 
+            public override bool CanCompare(object x, object y)
+            {
+                return TypeHelper.CanCast<T>(x) && TypeHelper.CanCast<T>(y);
+            }
```
mikkelbu, line 205: What is the reason for introducing this override? (As far as I can tell it is overriding the same implementation in the parent class `GenericEqualityAdapter<T>`)

src/NUnitFramework/framework/Constraints/EqualityAdapter.cs
```diff
@@ -224,6 +235,11 @@ class ComparerAdapter<T> : GenericEqualityAdapter<T>
         {
             private readonly IComparer<T> comparer;
 
+            public override bool CanCompare(object x, object y)
+            {
+                return TypeHelper.CanCast<T>(x) && TypeHelper.CanCast<T>(y);
+            }
```
mikkelbu, line 241: Same comment as above

mikkelbu (approved):
Looks good to me, but we will need a second review before merging this (as per the rules of the repository).

src/NUnitFramework/framework/Constraints/EqualityAdapter.cs
```diff
@@ -52,6 +52,12 @@ public virtual bool CanCompare(object x, object y)
...
                 return true;
 
+            if (x is string && !(y is IEnumerable) ||
+                y is string && !(x is IEnumerable))
+            {
+                return true;
+            }
+
             if (x is IEnumerable || y is IEnumerable)
                 return false;
```
jnm2, line 62: If desired, the whole method impl could be:

```cs
return (x is string || !(x is IEnumerable))
    && (y is string || !(y is IEnumerable));
```

jnm2 (dismissed):

src/NUnitFramework/framework/Constraints/EqualityAdapter.cs
manfred-brands: > What is the reason for introducing this override? (As far as I can tell it is overriding the same implementation in the parent class `GenericEqualityAdapter<T>`)

Initially I was overriding all `CanCompare` operations as per idea in the original ticket. These were indeed useless.
I have added a commit to remove the new overloads again. Thanks for detecting this.

src/NUnitFramework/framework/Constraints/EqualityAdapter.cs
manfred-brands: Removed.

src/NUnitFramework/framework/Constraints/EqualityAdapter.cs
manfred-brands: @jnm2 I like it, added a commit to the branch doing so.

jnm2 (approved):
Thanks for working on this!

mikkelbu (approved):
Also looks good to me. Thanks for doing this @manfred-brands. I'll wait with the merge in case you want to clean up the commits as in the Analyzer project.

manfred-brands:
Thanks @mikkelbu. Cleaned up commits.

mikkelbu (approved):

PR #3635, opened 2020-09-17
Don't use .NET 5, the RC requires a preview version of Visual Studio

rprouse:
If you have the RC of .NET 5 installed but haven't updated Visual Studio to the latest preview the projects won't load because .NET 5 RC1 requires MSBuild 16.8.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,7 +6,7 @@ branches:
 
 install:
   - ps: Invoke-WebRequest 'https://dot.net/v1/dotnet-install.ps1' -OutFile 'dotnet-install.ps1'
-  - ps: .\dotnet-install.ps1 -Version 5.0.100-preview.6.20318.15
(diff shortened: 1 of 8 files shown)


rprouse:
Looks like Appveyor only has .NET 5 preview 6 installed.

rprouse:
@jnm2 you switched the build machines to use .NET 5 preview 6. Do you have an issue with backing this down to 3.1 until the final version of .NET 5 is out?

jnm2 (dismissed):
That sounds best.

jnm2 (approved):

jnm2:
`net5.0` will need to be removed from the TargetFrameworks element in each project.

rprouse:
@jnm2 I removed the 5.0 targets and I also had to pin the FSharp.Core version because they released a 5.0 beta a couple of weeks ago that doesn't support .NET FX and it was getting pulled in.

jnm2 (approved):

PR #3642, opened 2020-10-06
Comparer: Preference IComparable<T> over IComparable

### LINKED ISSUE #3641: Type implementing IComparable<float> (or any IComparable) fails comparison.
Giving the following code:

```
private struct C : IComparable<float>
{
    public C(float v) => this.V = v;
    public float V { get; }
    public int CompareTo(float other) => this.V.CompareTo(other);
}

public void TestMethod()
{
    C x = new C(1.1f);
    float y = 1.0f;

    Assert.That(x.CompareTo(y), Is.GreaterThan(0));
    Assert.That(x,Is.GreaterThan(y));
}
```
The last test fails because `NunitComparer.Compare` checks `y is IComparable` before checking for `xType.CompareTo(yType)` and hence calls the `Single.CompareTo(Object)` method instead of `C.CompareTo(float)` method.

Can I switch the order such that the two tests on `IComparable` are last thereby preferencing `IComparable<T>` over `IComparable`?
I you think there are no unforeseen issues, I will raise a PR.

manfred-brands:
Fixes #3641

### DIFF
--- a/src/NUnitFramework/framework/Constraints/NUnitComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/NUnitComparer.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections;
+using System.Linq;
 using System.Reflection;
 using NUnit.Compatibility;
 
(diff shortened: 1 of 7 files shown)


src/NUnitFramework/framework/Constraints/NUnitComparer.cs
```diff
@@ -58,23 +59,29 @@ public int Compare(object x, object y)
...
+
             MethodInfo method = xType.GetMethod("CompareTo", new Type[] { yType });
-            if (method != null)
+            if (method != null && !IsIComparable(method))
                 return (int)method.Invoke(x, new object[] { y });
 
             method = yType.GetMethod("CompareTo", new Type[] { xType });
-            if (method != null)
+            if (method != null && !IsIComparable(method))
                 return -(int)method.Invoke(y, new object[] { x });
```
jnm2, line 77: I'm not sure this prefers `IComparable<object>` over `IComparable`, especially when explicitly implemented, but we might not care about that.

jnm2 (approved):

rprouse (approved):

PR #3646, opened 2020-10-12
IStructuralEquatable support

### LINKED ISSUE #3447: Is.EqualTo(...).Using(StructuralComparisons.StructuralEqualityComparer or StructuralComparer) not working
[`System.Collections.Immutable.ImmutableArray`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutablearray-1?view=netcore-3.1) implements both reference equality and structural equality via [`IStructuralEquatable`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.istructuralequatable?view=netcore-3.1).

I want to be able to check two arrays for structural equality, but only some of the overloads of [`EqualConstraint.Using`](https://github.com/nunit/nunit/blob/cefc0c3a86c065daa4130aa255b1384db51b2b08/src/NUnitFramework/framework/Constraints/EqualConstraint.cs#L308) work as expected.

The two overloads that take `IEqualityComparer` or `IComparer` don't use the correct comparison, whereas the versions that take delegates do.

The attached code (which was built with packages **NUnit 3.12.0** and **System.Collections.Immutable 1.7.0** fails the last two tests.

```C#
using System;
using System.Collections;
using System.Collections.Immutable;
using NUnit.Framework;

namespace NUnitStructural
{
    [TestFixture]
    public class Test
    {
        [Test]
        public void ImmutableArrayNotReferenceEquality()
        {
            var x = ImmutableArray.Create(1, 2, 3);
            var y = ImmutableArray.Create(1, 2, 3);

            Assert.That(x != y);
        }

        [Test]
        public void ImmutableArrayWithStructuralEqualityComparison()
        {
            var x = ImmutableArray.Create(1, 2, 3);
            var y = ImmutableArray.Create(1, 2, 3);

            Comparison<ImmutableArray<int>> comparison = (a, b) => StructuralComparisons.StructuralComparer.Compare(a, b);
            Assert.That(x, Is.EqualTo(y).Using(comparison));
        }

        [Test]
        public void ImmutableArrayWithStructuralEqualityComparerEquals()
        {
            var x = ImmutableArray.Create(1, 2, 3);
            var y = ImmutableArray.Create(1, 2, 3);

            var del = new Func<ImmutableArray<int>, ImmutableArray<int>, bool>((a, b) => StructuralComparisons.StructuralEqualityComparer.Equals(a, b));
            Assert.That(x, Is.EqualTo(y).Using<ImmutableArray<int>>(del));
        }

        [Test]
        public void ImmutableArrayWithStructuralComparerEquals()
        {
            var x = ImmutableArray.Create(1, 2, 3);
            var y = ImmutableArray.Create(1, 2, 3);

            var del = new Func<ImmutableArray<int>, ImmutableArray<int>, bool>((a, b) => StructuralComparisons.StructuralComparer.Compare(a, b) == 0);
            Assert.That(x, Is.EqualTo(y).Using<ImmutableArray<int>>(del));
        }

        [Test]
        public void ImmutableArrayWithStructuralEqualityComparer()
        {
            var x = ImmutableArray.Create(1, 2, 3);
            var y = ImmutableArray.Create(1, 2, 3);

            Assert.That(x, Is.EqualTo(y).Using(StructuralComparisons.StructuralEqualityComparer));
        }

        [Test]
        public void ImmutableArrayWithStructuralComparer()
        {
            var x = ImmutableArray.Create(1, 2, 3);
            var y = ImmutableArray.Create(1, 2, 3);

            Assert.That(x, Is.EqualTo(y).Using(StructuralComparisons.StructuralComparer));
        }
    }
}

```

stevenaw:
Fixes #3447 

It looks like the reason for the failures here was that, as an implementer of IEquatable<>, ImmutableArray was not deferring comparisons down to the enumerable-level. IEquatable<> in turn attempted reference equality on the boxed objects of ImmutableArray (a struct).

I'd welcome some feedback on the implementation here, as I'm not entirely sure the best solution.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
@@ -48,6 +48,9 @@ namespace NUnit.Framework.Constraints.Comparers
             Type xType = x.GetType();
             Type yType = y.GetType();
 
+            if (DoesUseStructuralEquality(xType) && DoesUseStructuralEquality(yType))
+                return null;
+
             MethodInfo equals = FirstImplementsIEquatableOfSecond(xType, yType);
             if (equals != null)
                 return InvokeFirstIEquatableEqualsSecond(x, y, equals);
@@ -57,6 +60,15 @@ namespace NUnit.Framework.Constraints.Comparers
                 return InvokeFirstIEquatableEqualsSecond(y, x, equals);
 
             return null;
+        }
+
+        private static bool DoesUseStructuralEquality(Type type)
+        {
+            foreach(var @interface in type.GetInterfaces())
+                if (@interface.FullName == "System.Collections.IStructuralComparable" || @interface.FullName == "System.Collections.IStructuralEquatable")
+                    return true;
+
+            return false;
         }
 
         private static MethodInfo FirstImplementsIEquatableOfSecond(Type first, Type second)
--- a/src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
@@ -22,10 +22,7 @@
 // ***********************************************************************
 
 using System;
-using System.Collections;
-using System.Collections.Generic;
 using System.Reflection;
-using NUnit.Compatibility;
 
 namespace NUnit.Framework.Constraints.Comparers
 {
@@ -49,10 +46,6 @@ namespace NUnit.Framework.Constraints.Comparers
             Type xType = x.GetType();
             Type yType = y.GetType();
 
-            if (x is IEnumerable && DoesUseStructuralEquality(xType)
-                && y is IEnumerable && DoesUseStructuralEquality(yType))
-                return null;
-
             MethodInfo equals = FirstImplementsIEquatableOfSecond(xType, yType);
             if (equals != null)
                 return InvokeFirstIEquatableEqualsSecond(x, y, equals);
@@ -62,18 +55,6 @@ namespace NUnit.Framework.Constraints.Comparers
                 return InvokeFirstIEquatableEqualsSecond(y, x, equals);
 
             return null;
-        }
-
-        private static bool DoesUseStructuralEquality(Type type)
-        {
-            if (!type.IsValueType)
-                return false;
-
-            foreach(var @interface in type.GetInterfaces())
-                if (@interface.FullName == "System.Collections.IStructuralComparable" || @interface.FullName == "System.Collections.IStructuralEquatable")
-                    return true;
-
-            return false;
         }
 
         private static MethodInfo FirstImplementsIEquatableOfSecond(Type first, Type second)
--- a/src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
@@ -88,7 +88,7 @@ namespace NUnit.Framework.Constraints.Comparers
             return pair.Value;
         }
 
-        private static IList<KeyValuePair<Type, MethodInfo>> GetEquatableGenericArguments(Type type)
+        private static List<KeyValuePair<Type, MethodInfo>> GetEquatableGenericArguments(Type type)
         {
             var genericArgs = new List<KeyValuePair<Type, MethodInfo>>();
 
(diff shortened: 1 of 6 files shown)


src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
```diff
@@ -48,6 +49,10 @@ internal EquatablesComparer(NUnitEqualityComparer equalityComparer)
             Type xType = x.GetType();
             Type yType = y.GetType();
 
+            if (x is IEnumerable && DoesUseStructuralEquality(xType)
+                && y is IEnumerable && DoesUseStructuralEquality(yType))
+                return null;
```
stevenaw, line 54: Skip using EquatablesComparer if each type fulfills these criteria:
1. Implements IEnumerable
2. Implements either IStructuralComparable or IStructuralEquatable
3. Is a value type

I'm unsure if this is overkill. Would it be enough to just check for `IEnumerable` here, or would that be overly broad and break something? Alternately, would it be better to instead check if the types are valuetypes that implement `IStructuralEquatable` (though I suspect this might break the "defer to enumerables" idea)?

src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
```diff
@@ -71,7 +88,7 @@ private static MethodInfo FirstImplementsIEquatableOfSecond(Type first, Type sec
             return pair.Value;
         }
 
-        private static IList<KeyValuePair<Type, MethodInfo>> GetEquatableGenericArguments(Type type)
+        private static List<KeyValuePair<Type, MethodInfo>> GetEquatableGenericArguments(Type type)
```
stevenaw, line 91: `List<T>.GetEnumerator()` returns a value type, which avoid an allocation at the caller.

src/NUnitFramework/tests/nunit.framework.tests.csproj
```diff
@@ -32,6 +32,7 @@
 
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.7.0" />
     <PackageReference Include="NUnit3TestAdapter" Version="3.13.0" />
+    <PackageReference Include="System.Collections.Immutable" Version="1.5.0" />
```
stevenaw, line 35: I started to hit issues when building with Cake with version >= 1.6 (similar to what was documented here: https://github.com/Azure/azure-functions-vs-build-sdk/issues/353). VS builds fine with the latest version of this (1.7.1).

src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
jnm2: I like the idea that we would give IStructural interfaces priority over IEnumerable. Not sure that being a value type should make a difference; is there are reason you can think of?

src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
```diff
@@ -59,6 +64,18 @@ internal EquatablesComparer(NUnitEqualityComparer equalityComparer)
             return null;
         }
 
+        private static bool DoesUseStructuralEquality(Type type)
+        {
+            if (!type.IsValueType)
+                return false;
```
jnm2, line 70: Why would a reference type not use structural equality? For example, C# 9 records are emitted as ordinary classes, but they implement structural equality as indeed any class can if it wishes.

src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
stevenaw: Good points thanks @jnm2 . I think I'd been looking for a "quick and easy" fix here, but I see it's a little more nuanced. Especially with having IStructural take priority over IEnumerable (similar to how EquatablesComparer takes priority).

Re: value types, I had originally added that in error. I noticed MSDN primarily showed value types as implementing this, but of course both structs and classes can implement interfaces.

I've stepped back and given this a different approach now, with a new `StructuralComparer` type which, when invoked, will use NUnitEqualityComparer for all comparisons (including Tolerance).

src/NUnitFramework/framework/nunit.framework.csproj
```diff
@@ -11,6 +11,10 @@
     <PackageReference Include="jnm2.ReferenceAssemblies.net35" Version="1.0.1" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' != 'net35' and '$(TargetFramework)' != 'net40'">
```
stevenaw, line 14: `IStructuralEquatable` is not available to NET4 on NuGet via `System.Runtime`. Instead, it is just a part of the framework

rprouse (approved):

PR #3647, opened 2020-10-14
Fix exception under blazor 5

Temtaime:
See https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/#dealing-with-apis-that-are-unsupported-in-blazor-webassembly

### DIFF
--- a/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
+++ b/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
@@ -258,7 +258,12 @@ namespace NUnit.Framework.Api
 
             testAssembly.ApplyAttributesToTest(assembly);
 
-            testAssembly.Properties.Set(PropertyNames.ProcessId, System.Diagnostics.Process.GetCurrentProcess().Id);
+            try
+            {
+                testAssembly.Properties.Set(PropertyNames.ProcessId, System.Diagnostics.Process.GetCurrentProcess().Id);
+            }
+            catch (PlatformNotSupportedException)
+            { }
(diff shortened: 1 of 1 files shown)


src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -258,7 +258,12 @@ private TestSuite BuildTestAssembly(Assembly assembly, string assemblyNameOrPath
 
             testAssembly.ApplyAttributesToTest(assembly);
 
-            testAssembly.Properties.Set(PropertyNames.ProcessId, System.Diagnostics.Process.GetCurrentProcess().Id);
+            try
+            {
+                testAssembly.Properties.Set(PropertyNames.ProcessId, System.Diagnostics.Process.GetCurrentProcess().Id);
```
jnm2, line 263: Also starting in .NET 5, we should use the much less wasteful `Environment.ProcessId`. Do you know if Blazor still implements `Environment.ProcessId` to throw the same exception?

jnm2 (approved):

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
jnm2: I guess we could also call https://docs.microsoft.com/en-us/dotnet/api/system.operatingsystem.isbrowser?view=net-5.0 inside the `#if NET5_0` section even if Environment.ProcessId does throw, rather than catching an exception.

mikkelbu: I think I would prefer that solution, as I think we should avoid catching exceptions with empty error handlers if it is possible.

rprouse (approved):

PR #3654, opened 2020-10-25
Fix for #3592

### LINKED ISSUE #3592: Add classname and methodname to the start-test event
The test case finished event includes the classname and methodname.  The start-test event needs the same.  The lack of this is the reason the NUnit3TestAdapter cannot skip the discovery phase.

OsirisTerje:
Fixes #3592

### DIFF
--- a/src/NUnitFramework/framework/Internal/TestProgressReporter.cs
+++ b/src/NUnitFramework/framework/Internal/TestProgressReporter.cs
@@ -64,14 +64,9 @@ namespace NUnit.Framework.Internal
             var parent = GetParent(test);
             try
             {
-                string report = string.Format(
-                    "<{0} id=\"{1}\" parentId=\"{2}\" name=\"{3}\" fullname=\"{4}\" type=\"{5}\"/>",
-                    startElement,
-                    test.Id,
-                    parent != null ? parent.Id : string.Empty,
-                    FormatAttributeValue(test.Name),
-                    FormatAttributeValue(test.FullName),
-                    test.TestType);
+                string report = test is TestSuite 
+                    ? $"<{startElement} id=\"{test.Id}\" parentId=\"{(parent != null ? parent.Id : string.Empty)}\" name=\"{FormatAttributeValue(test.Name)}\" fullname=\"{FormatAttributeValue(test.FullName)}\" type=\"{test.TestType}\"/>" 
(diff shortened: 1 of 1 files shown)


src/NUnitFramework/framework/Internal/TestProgressReporter.cs
```diff
@@ -64,14 +64,9 @@ public void TestStarted(ITest test)
...
-                    "<{0} id=\"{1}\" parentId=\"{2}\" name=\"{3}\" fullname=\"{4}\" type=\"{5}\"/>",
-                    startElement,
-                    test.Id,
-                    parent != null ? parent.Id : string.Empty,
-                    FormatAttributeValue(test.Name),
-                    FormatAttributeValue(test.FullName),
-                    test.TestType);
+                string report = test is TestSuite 
+                    ? $"<{startElement} id=\"{test.Id}\" parentId=\"{(parent != null ? parent.Id : string.Empty)}\" name=\"{FormatAttributeValue(test.Name)}\" fullname=\"{FormatAttributeValue(test.FullName)}\" type=\"{test.TestType}\"/>" 
+                    : $"<{startElement} id=\"{test.Id}\" parentId=\"{(parent != null ? parent.Id : string.Empty)}\" name=\"{FormatAttributeValue(test.Name)}\" fullname=\"{FormatAttributeValue(test.FullName)}\" type=\"{test.TestType}\" classname=\"{FormatAttributeValue(test.ClassName ?? "")}\" methodname=\"{FormatAttributeValue(test.MethodName ?? "")}\"/>";
```
mikkelbu, line 69: Looking at the entire `TestStarted` method I think we should remove `startElement` and just state the elements explicitly
```suggestion
                string report = test is TestSuite 
                    ? $"<start-suite id=\"{test.Id}\" parentId=\"{(parent != null ? parent.Id : string.Empty)}\" name=\"{FormatAttributeValue(test.Name)}\" fullname=\"{FormatAttributeValue(test.FullName)}\" type=\"{test.TestType}\"/>"
                    : $"<start-test id=\"{test.Id}\" parentId=\"{(parent != null ? parent.Id : string.Empty)}\" name=\"{FormatAttributeValue(test.Name)}\" fullname=\"{FormatAttributeValue(test.FullName)}\" type=\"{test.TestType}\" classname=\"{FormatAttributeValue(test.ClassName ?? "")}\" methodname=\"{FormatAttributeValue(test.MethodName ?? "")}\"/>";
```
also I'm considering whether we should wrap these lines, as the longest is more than 350 characters.

mikkelbu (commented):
I think we at least should extend the test `TestStarted_ReportAttributes` to also capture the new properties.

rprouse (approved):

PR #3655, opened 2020-10-26
1546 - Equatable comparer works regardless of runtime

### LINKED ISSUE #1546: NUnitEqualityComparer.GetEquatableGenericArguments should explicitly order arguments
`NUnitEqualityComparer.GetEquatableGenericArguments(...)` returns arguments in order from the base class up to the derived class, but that behaviour is not guaranteed, it just happens to work. This may not work on all platforms.

See the discussion in #1544 for more information.

stevenaw:
Fixes #1546 

As noted in the original issue, GetInterfaces() is not guaranteed to return interfaces in a particular order across all runtimes. This PR replaces it with `FindInterfaces()` and `GetInterfaceMap()` to always get the single most derived implementation of a particular interface for the provided type

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
@@ -67,33 +67,30 @@ namespace NUnit.Framework.Constraints.Comparers
             {
                 if (implementation.Argument.IsAssignableFrom(second))
                 {
-                    if (mostDerived.Argument == null)
+                    if (mostDerived.Argument == null || mostDerived.Argument.IsAssignableFrom(implementation.Argument))
                         mostDerived = implementation;
-                    else if (mostDerived.Argument.IsAssignableFrom(implementation.Argument))
-                    {
-                        // Look for method with most-derived of type arg which can be assigned to 'second'
-                        // If there end up being multiple of same type, look for most derived declaring type
-                        if (mostDerived.Argument != implementation.Argument
-                            || mostDerived.Argument.IsAssignableFrom(implementation.Argument))
-                        mostDerived = implementation;
-                    }
                 }
             }
 
             return mostDerived.Method;
         }
 
-        private static IEnumerable<EquatableMethodImpl> GetEquatableImplementations(Type type)
+        private static EquatableMethodImpl[] GetEquatableImplementations(Type type)
         {
-            foreach (Type @interface in type.GetInterfaces())
+            static bool IsIEquatableOfT(Type t, object filter) => t.IsGenericType && t.GetGenericTypeDefinition().Equals(typeof(IEquatable<>));
+            
+            var interfaces = type.FindInterfaces((t,f) => IsIEquatableOfT(t,f), string.Empty);
+            var implementations = new EquatableMethodImpl[interfaces.Length];
+
+            for(var i = 0; i < interfaces.Length; i++)
             {
-                if (@interface.GetTypeInfo().IsGenericType && @interface.GetGenericTypeDefinition().Equals(typeof(IEquatable<>)))
-                {
-                    var iMap = type.GetInterfaceMap(@interface);
-                    foreach (var method in iMap.TargetMethods)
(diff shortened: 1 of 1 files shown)


src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
```diff
@@ -61,35 +61,53 @@ internal EquatablesComparer(NUnitEqualityComparer equalityComparer)
 
         private static MethodInfo FirstImplementsIEquatableOfSecond(Type first, Type second)
         {
-            var pair = new KeyValuePair<Type, MethodInfo>();
+            var mostDerived = default(EquatableMethodImpl);
 
-            foreach (var xEquatableArgument in GetEquatableGenericArguments(first))
-                if (xEquatableArgument.Key.IsAssignableFrom(second))
-                    if (pair.Key == null || pair.Key.IsAssignableFrom(xEquatableArgument.Key))
```
stevenaw, line 68: Original PR used KeyValuePair<> for storing all this. During development I created a custom struct because an earlier approach required a third piece of data to store. I eventually got it working with only `Type` and `MethodInfo` again, but I kept the custom struct since I found it easier to read this way.

Please let me know if you'd like me to revert it back to using `KeyValuePair<Type, MethodInfo>`.

src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
```diff
@@ -61,35 +61,53 @@ internal EquatablesComparer(NUnitEqualityComparer equalityComparer)
...
+        private static EquatableMethodImpl[] GetEquatableImplementations(Type type)
         {
-            var genericArgs = new List<KeyValuePair<Type, MethodInfo>>();
+            static bool IsIEquatableOfT(Type t, object filter) => t.IsGenericType && t.GetGenericTypeDefinition().Equals(typeof(IEquatable<>));
+            
+            var interfaces = type.FindInterfaces((t,f) => IsIEquatableOfT(t,f), string.Empty);
+            var implementations = new EquatableMethodImpl[interfaces.Length];
 
-            foreach (Type @interface in type.GetInterfaces())
+            for(var i = 0; i < interfaces.Length; i++)
```
rprouse, line 85: Generally I'd prefer to see a Linq `Select(i => new EquatableMethodImpl(...)).ToArray()` but this is okay, it wasn't Linq before 😉

rprouse (approved):
We'll need a second review from a team member.

src/NUnitFramework/framework/Constraints/Comparers/EquatablesComparer.cs
```diff
@@ -67,33 +67,30 @@ private static MethodInfo FirstImplementsIEquatableOfSecond(Type first, Type sec
...
             {
-                if (@interface.GetTypeInfo().IsGenericType && @interface.GetGenericTypeDefinition().Equals(typeof(IEquatable<>)))
-                {
-                    var iMap = type.GetInterfaceMap(@interface);
-                    foreach (var method in iMap.TargetMethods)
-                        yield return new EquatableMethodImpl(method, method.GetParameters()[0].ParameterType, iMap.TargetType);
-                }
+                var iMap = type.GetInterfaceMap(interfaces[i]);
+                var method = iMap.TargetMethods[0];
```
oznetmaster, line 88: This is incorrect.  It should be:
```
var method = iMap.InterfaceMethods[0];
```

jnm2: Noting the followup at https://github.com/nunit/nunit/issues/3683#issuecomment-739479852:

> It appears that the problem is in the third part implementation of InterfaceMap I was using.
> 
> You may remember that all of my work is on WindowsCE and the Compact Framework.
> 
> Since the Compact Framework does not support InterfaceMap, I had to use a third party implementation of it.
> 
> Apparently, that implementation is faulty.
> 
> I am writing my own implementation, which should solve my problem.
> 
> Sorry for the noise here.

PR #3666, opened 2020-11-13
Support .Net 5.0

### LINKED ISSUE #3650: Build issue with the latest .NET SDK 5.0.100-rc.2
If I install latest version .NET SDK 5.0.100-rc.2 I am not able to compile the project, because NuGetRestore downloads FSharp.Core v5.0.0 in `nunit.testdata.fsharp` project. It errors out with the following message: "Package FSharp.Core 5.0.0 is not compatible with net45"

If I install version .NET SDK 5.0.100-preview.6 (which is listed in `global.json`) NuGetRestore downloads FSharp.Core v4.7.2 which lets me compile the solution without any errors.

rprouse:
Fixes #3650

Fixes for building on the final version of .NET 5.0

- Global.json updated
- Target FSharp.Core 4.7.2 and disable automatically targeting 5.0.0.
- The framework define `NETCOREAPP5_0` was switched to `NET5_0`
- Our WPF test code needs to target `net5.0-windows`
- There are some changes in nullability
- Updates Cake build to use the new bin directories
- Updates the CI build definitions
- Update build requirements in `BUILDING.md`

### DIFF
--- a/src/NUnitFramework/framework/Internal/TypeHelper.cs
+++ b/src/NUnitFramework/framework/Internal/TypeHelper.cs
@@ -198,7 +198,7 @@ namespace NUnit.Framework.Internal
             if (type1.IsAssignableFrom(type2)) { bestCommonType = type1; return true; }
             if (type2.IsAssignableFrom(type1)) { bestCommonType = type2; return true; }
 
-            bestCommonType = null;
+            bestCommonType = typeof(object);
             return false;
         }
 
--- a/src/NUnitFramework/framework/nunit.framework.csproj
+++ b/src/NUnitFramework/framework/nunit.framework.csproj
@@ -6,6 +6,11 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" />
+    <PackageReference Include="jnm2.ReferenceAssemblies.net35" Version="1.0.1" />
+  </ItemGroup>
+
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.Win32.Registry" Version="4.5.0" />
   </ItemGroup>
--- a/src/NUnitFramework/nunitlite-runner/nunitlite-runner.csproj
+++ b/src/NUnitFramework/nunitlite-runner/nunitlite-runner.csproj
@@ -8,8 +8,15 @@
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" Condition="'$(UseWindowsFormsAndWPF)' != 'true'" />
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk.WindowsDesktop" Condition="'$(UseWindowsFormsAndWPF)' == 'true'" />
 
-  <PropertyGroup>
+  <PropertyGroup Condition="'$(OS)' == 'Windows_NT'">
+    <TargetFrameworks>net35;net40;net45;netcoreapp2.1;netcoreapp3.1;net5.0-windows</TargetFrameworks>
+  </PropertyGroup>
+  
+  <PropertyGroup Condition="'$(OS)' != 'Windows_NT'">
     <TargetFrameworks>net35;net40;net45;netcoreapp2.1;netcoreapp3.1;net5.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <PropertyGroup>
     <OutputType>Exe</OutputType>
     <RootNamespace>NUnitLite</RootNamespace>
   </PropertyGroup>
--- a/src/NUnitFramework/nunitlite-runner/nunitlite-runner.csproj
+++ b/src/NUnitFramework/nunitlite-runner/nunitlite-runner.csproj
@@ -11,7 +11,7 @@
   <PropertyGroup Condition="'$(OS)' == 'Windows_NT'">
     <TargetFrameworks>net35;net40;net45;netcoreapp2.1;netcoreapp3.1;net5.0-windows</TargetFrameworks>
   </PropertyGroup>
-  
+
   <PropertyGroup Condition="'$(OS)' != 'Windows_NT'">
     <TargetFrameworks>net35;net40;net45;netcoreapp2.1;netcoreapp3.1;net5.0</TargetFrameworks>
   </PropertyGroup>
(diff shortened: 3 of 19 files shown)


src/NUnitFramework/framework/Internal/TypeHelper.cs
```diff
@@ -198,7 +198,7 @@ public static bool TryGetBestCommonType(Type? type1, Type? type2, [NotNullIfNotN
             if (type1.IsAssignableFrom(type2)) { bestCommonType = type1; return true; }
             if (type2.IsAssignableFrom(type1)) { bestCommonType = type2; return true; }
 
-            bestCommonType = null;
+            bestCommonType = typeof(object);
```
rprouse, line 201: This cannot be `null` if `type1` and `type2` are not null which could be the case. It was causing a compile error, so I looked at the calling code. If we ever get here and return `false`, we immediately set another value to the `out` parameter.

src/NUnitFramework/nunitlite-runner/nunitlite-runner.csproj
```diff
@@ -8,8 +8,15 @@
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" Condition="'$(UseWindowsFormsAndWPF)' != 'true'" />
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk.WindowsDesktop" Condition="'$(UseWindowsFormsAndWPF)' == 'true'" />
 
-  <PropertyGroup>
+  <PropertyGroup Condition="'$(OS)' == 'Windows_NT'">
+    <TargetFrameworks>net35;net40;net45;netcoreapp2.1;netcoreapp3.1;net5.0-windows</TargetFrameworks>
```
rprouse, line 12: We need the windows target for our WinForms and WPF tests.

rprouse:
AppVeyor and Azure DevOps builds are failing because they haven't updated the build images yet to use VS 16.8.

AppVeyor has a VS 2019 Preview image that should work, but Azure DevOps doesn't, so we may as well wait a bit. I'll check back in a couple of days.

rprouse:
- AppVeyor [has been updated](https://www.appveyor.com/updates/2020/11/14/)
- For Azure DevOps, [.NET 5.0 will be installed by default in two weeks](https://github.com/actions/virtual-environments/issues/1891#issuecomment-726802027), but we are already installing it manually, so that part should be okay. So now we are waiting on VS 16.8 which is being tracked here [https://github.com/actions/virtual-environments/issues/2022]. They are hoping for sometime this week?

The workarounds to get Azure DevOps building are all Windows only, so we need to wait. I must say, I'm a little bit disappointed in Microsoft right now. IMHO they should have had this out with the release and then retagged the current version to Visual Studio 2019 Previous like Appveyory does.

mikkelbu (approved):
As far as I could tell the changes looks good to me, but I think a second pair of eyes on this would be useful.

rprouse:
The Windows images on Azure DevOps have been updated and are passing now, but the Linux and Mac images still have 16.6 on them, so they are way behind. According to https://github.com/actions/virtual-environments/issues/2082 at least Mac should roll out tomorrow.

rprouse:
@jnm2 can you take a look at this PR when you get a chance, especially my change for nullability?

src/NUnitFramework/framework/Internal/TypeHelper.cs
jnm2: Oh, I missed this when I annotated. My preferred fix would be to add `[MaybeNullWhen(false)]` to the parameter, if that works alongside NotNullIfNotNull, but this is also fine. Either way, if this method returns false, the out parameter should not be read by the caller.

jnm2 (approved):

rprouse:
The Mac images have rolled out but it is still failing. I'll try again tomorrow and if it is still failing I'll see if I can find a workaround.

jnm2:
I think we might need a newer Mono version. The build failures are on Linux and macOS when "Microsoft (R) Build Engine version 16.6.0 for Mono" runs into the target framework `net5.0`.

jnm2:
Or, we could start building with .NET Core instead of Mono's msbuild. .NET Framework reference assemblies are automatically included starting in the .NET 5 SDK so that building for .NET Framework work cross-platform. You'll need https://www.nuget.org/packages/jnm2.ReferenceAssemblies.net35 though because they decided not to support 3.5 in particular.

rprouse:
> I think we might need a newer Mono version. The build failures are on Linux and macOS when "Microsoft (R) Build Engine version 16.6.0 for Mono" runs into the target framework net5.0 and thinks it needs a .NET Framework targeting pack.

That was my thinking and I was hoping that the updated build images would support it, but Mono hasn't seen an [update since May](https://www.mono-project.com/docs/about-mono/releases/), so I doubt an update would help.

I'll look into switching to a .NET Core build. Thanks for linking to your package @jnm2 .

rprouse:
The work to update Mono from 16.6 isn't even on a preview channel yet and breaking several other projects like Omnisharp, so I don't think we can wait. https://github.com/mono/mono/issues/20250

src/NUnitFramework/framework/nunit.framework.csproj
```diff
@@ -6,6 +6,11 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" />
```
jnm2, line 10: ℹ Microsoft.NETFramework.ReferenceAssemblies is implicitly referenced when needed starting with SDK 5.0.100, so we can remove this if require that minimum SDK version on all platforms.

src/NUnitFramework/nunitlite.tests/nunitlite.tests.csproj
```diff
@@ -1,16 +1,28 @@
...
+  </PropertyGroup>
+
   <ItemGroup>
     <ProjectReference Include="..\mock-assembly\mock-assembly.csproj" />
     <ProjectReference Include="..\nunitlite\nunitlite.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" />
+    <PackageReference Include="jnm2.ReferenceAssemblies.net35" Version="1.0.1" />
```
jnm2, line 23: I would consider placing this in Directory.Build.props rather than in each csproj.

jnm2 (approved):
Looks like the build failure is an transient network issue.
https://nunit.visualstudio.com/NUnit/_build/results?buildId=3386&view=logs&j=0da5d1d9-276d-5173-c4c4-9d4d4ed14fdb&t=4b52221a-dfc4-5cf4-1545-f434ad87af34&l=10

rprouse:
As usual when we try to support newer .NET Frameworks, we got backed into a corner. We currently use Mono to compile on Linux and Mac, but the Mono project doesn't support building .NET 5.0. I checked the Mono project and they haven't started work on it and judging by their recent commit history, I don't think it will be available anytime soon.

I tried switching to building with `dotnet build` which nearly worked except the `Microsoft.BCL.*` assemblies that we use for `async` support in .NET 4.0 have MSBuild targets that don't work with `dotnet build`.

To get this working for the release, I'm not building or testing .NET 5 on Linux or Mac. For the 3.14 release, we should consider dropping the .NET 4.0 target and adding a .NET 5.0 target. .NET 4.0 was EOL in Jan 2016, nearly five years ago.

Issue #3473, opened 2020-02-19
Assert.NotNull should have a ValidatedNotNullAttribute on the anObject parameter


Corniel:
Take the following example:

``` C#
public static void MySmartAssert(MyComplexModel model)
{
    Assert.NotNull(model, nameof(model));
   if (model.WhatEver)
   {
       Assert.Fail("What ever!");
   }
}
```

Static code analyzers like Sonar will generate a warning on the `model.WhatEver` call, arguing that model could be `null`. Since the early days of FxCop the `[ValidatedNotNullAttribute]` was introduced to help with the analysis, saying: if a parameter was decorated with that, it could be safely assumed that no null reference exception could occur on later references to it.

It would be nice if `Assert.IsNotNull`  and `Assert.NotNull` would have such an attribute on the `object AnObject` parameter.

CharliePoole:
I understand the perceived need, but this seems like a bit of a slippery slope, in several ways...

1. Many nunit assertions and constraints could take a null argument. As a test framework, nunit looks for such stuff and doesn't normally throw an NRE when it gets one. Rather, it fails a test. As an example, what about `Assert.That(model, Is.Not.Null)`? Does it pass the analyzer? I'm afraid we would have to add the attribute to lots of methods.

2. Assuming we added it wherever needed, which analyzers will it satisfy. Are there analyzers that want something else? Do we need to support them as well?

3. As a specific example, if I understand the issue correctly, this approach apparently won't work for Roslyn-based analyzers. See https://github.com/dotnet/roslyn-analyzers/issues/2215

I guess I'm saying that NUnit needs to figure out how much of a burden this could eventually become before taking the first step toward supporting it.

Meanwhile, can't you define the attribute in your own test code and use it where you need it?

jnm2:
There's another problem that means it's not obviously the right annotation. If the parameter is null and you are inside a multiple-assert block, the method will return and execution will continue.

Corniel:
@CharliePoole : Let met start with point 3: (Sonar Analyzer)[https://github.com/SonarSource/sonar-dotnet] is one of the bigger Roslyn analyzers out there. The adopted this because the classic FxCop (built by Microsoft) introduced this. I'm therefor under the impression that a common/advised way doing so. So from that perspective I don't think it is a slippery slope.

[...]

PR #3677, opened 2020-11-27
Add attributes to Assert.NotNull method to tell compiler input is checked for null.

### LINKED ISSUE #3376: Nullable Reference Types annotations

manfred-brands:
Partially fixes #3376

### DIFF
--- a/src/NUnitFramework/framework/Assert.Conditions.cs
+++ b/src/NUnitFramework/framework/Assert.Conditions.cs
@@ -26,6 +26,7 @@
 using System.Collections;
 using NUnit.Framework.Constraints;
 using System;
+using System.Diagnostics.CodeAnalysis;
 
 namespace NUnit.Framework
 {
@@ -41,7 +42,7 @@ namespace NUnit.Framework
         /// <param name="args">Array of objects to be used in formatting the message</param>
         public static void True(bool? condition, string? message, params object?[]? args)
         {
-            Assert.That(condition, Is.True ,message, args);
+            Assert.That(condition, Is.True, message, args);
         }
 
         /// <summary>
@@ -61,7 +62,7 @@ namespace NUnit.Framework
         /// <param name="condition">The evaluated condition</param>
         public static void True(bool? condition)
         {
-            Assert.That(condition, Is.True ,null, null);
+            Assert.That(condition, Is.True, null, null);
         }
 
         /// <summary>
@@ -70,7 +71,7 @@ namespace NUnit.Framework
         /// <param name="condition">The evaluated condition</param>
         public static void True(bool condition)
         {
-            Assert.That(condition, Is.True ,null, null);
+            Assert.That(condition, Is.True, null, null);
         }
 
         /// <summary>
@@ -81,7 +82,7 @@ namespace NUnit.Framework
         /// <param name="args">Array of objects to be used in formatting the message</param>
         public static void IsTrue(bool? condition, string? message, params object?[]? args)
         {
-            Assert.That(condition, Is.True ,message, args);
+            Assert.That(condition, Is.True, message, args);
         }
 
         /// <summary>
@@ -92,7 +93,7 @@ namespace NUnit.Framework
         /// <param name="args">Array of objects to be used in formatting the message</param>
         public static void IsTrue(bool condition, string? message, params object?[]? args)
         {
-            Assert.That(condition, Is.True ,message, args);
+            Assert.That(condition, Is.True, message, args);
         }
 
         /// <summary>
@@ -101,7 +102,7 @@ namespace NUnit.Framework
         /// <param name="condition">The evaluated condition</param>
         public static void IsTrue(bool? condition)
         {
-            Assert.That(condition, Is.True ,null, null);
+            Assert.That(condition, Is.True, null, null);
         }
 
         /// <summary>
@@ -110,7 +111,7 @@ namespace NUnit.Framework
         /// <param name="condition">The evaluated condition</param>
         public static void IsTrue(bool condition)
         {
-            Assert.That(condition, Is.True ,null, null);
+            Assert.That(condition, Is.True, null, null);
         }
 
         #endregion
@@ -126,7 +127,7 @@ namespace NUnit.Framework
         /// <param name="args">Array of objects to be used in formatting the message</param>
         public static void False(bool? condition, string? message, params object?[]? args)
         {
-            Assert.That(condition, Is.False ,message, args);
+            Assert.That(condition, Is.False, message, args);
         }
 
         /// <summary>
@@ -138,7 +139,7 @@ namespace NUnit.Framework
         /// <param name="args">Array of objects to be used in formatting the message</param>
         public static void False(bool condition, string? message, params object?[]? args)
         {
-            Assert.That(condition, Is.False ,message, args);
+            Assert.That(condition, Is.False, message, args);
         }
 
         /// <summary>
@@ -147,7 +148,7 @@ namespace NUnit.Framework
         /// <param name="condition">The evaluated condition</param>
         public static void False(bool? condition)
         {
-            Assert.That(condition, Is.False ,null, null);
+            Assert.That(condition, Is.False, null, null);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


Dreamescaper:
Please take a look at this comment.
https://github.com/nunit/nunit/issues/3376#issuecomment-733099803

Such attributes were never added, because NUnit does _not_ guarantees that it will fail in case of null argument - because it might be inside Assert.Multiple block.

manfred-brands:
@Dreamescaper Yes, I know NUnit doesn't throw inside an `Assert.Multiple`. The use case for that would be to test multiple variables for different values. I don't see a use case for an `Assert.Multiple` where one would assert a variable to be not null and then use it.

src/NUnitFramework/framework/Assert.Conditions.cs
```diff
@@ -201,33 +202,35 @@ public static void IsFalse(bool? condition)
...
 
         #endregion
 
         #region NotNull
 
+#pragma warning disable CS8777 // Parameter must have a non-null value when exiting.
+
         /// <summary>
         /// Verifies that the object that is passed in is not equal to <see langword="null"/>. Returns without throwing an
         /// exception when inside a multiple assert block.
```
jnm2, line 216: Note this documentation I added to explain why we decided not to add `[NotNull]`.

jnm2 (commented):
The team needs to have a conversation if we want to change the decision made at https://github.com/nunit/nunit/issues/3473#issuecomment-592254847. That's something I'm open to thinking about. There would then be a lot of new annotations to add, including `[MaybeNull]` on the Assert.Null methods, `[DoesNotReturn]` on Assert.Fail, and multiple `[NotNullIfNotNull(...)]` on equality asserts.

jnm2:
@rprouse Where would you recommend that we have this discussion?

rprouse:
@jnm2 let's have a side discussion in Slack then bring it back here after we chat.

src/NUnitFramework/framework/AssertionExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <typeparam name="T">The type of instance being handled.</typeparam>
+        /// <param name="source">The source instance to check for null.</param>
+        /// <returns>The source instance, or throws when null.</returns>
+        public static T AssertNotNull<T>(this T? source)
+            where T : class
+        {
+            Assert.NotNull(source);
+
+            return source;
+        }
```
jnm2, line 22: I just noticed these new extension method Assert APIs. This is something we should also discuss. If we add these, we should add `T : struct` copies of these methods.

manfred-brands: If we change the language version to C#9 we can drop the `T : class` constraint and it will work for both.

jnm2: The return type would not be as convenient if we did that though. `T?` means "defaultable" rather than "nullable" if there is no class or struct constraint.

Having to do `nullableValueType.AssertNotNull().Value` would be unpleasant to me since we can avoid making the `.Value` part (logically an assertion itself)  necessary.

manfred-brands: I stand corrected. Thanks. Disappointing that even though the return type is `T` it still returns `T?` I thought C#9 fixed this. 

As constraints are still not part of the signature we cannot have two versions, one with `where T : class` and one with `where T : struct`. However, I managed to make the 2nd by explicitly declaring it as:

        public static T AssertNotNull<T>(this Nullable<T> source)
            where T : struct

jnm2: We actually can have both versions because the signature is still different for the `T : struct` version when the parameter is `T?`.

The reason that C# 9 didn't fix this the way people hoped was that there is no sequence of bytes the compiler can emit that would work that way. The .NET runtime would have to gain a new feature and have a new way to encode ECMA-335 metadata that would not be compatible with previous runtimes and versions.

manfred-brands: You are correct again. Thanks.

jnm2: (The ability to *resolve* compiler overloads like this is a newish C# compiler feature IIRC, but IIRC it was backported to all previous language versions. Don't quote me on this.)

jnm2: No worries, glad I can help.

rprouse:
I had a long discussion with @jnm2 about this in Slack. I'm okay with changing the legacy asserts in this way, just not adding new ones, so that part isn't a problem. My issue is that this is a partial fix that cannot have a complete fix within the framework. If we do this for `Assert.NotNull` in the framework, then people will expect it to work for `Assert.That(actual, Is.NotNull)` and might also expect it to work in multiple assert blocks.

I think that we could handle all of these things in the Analyzer project, so it is the best place for that. Unfortunately, there are also some technical difficulties there, so let's leave this PR open for now and see how the work goes in the Analyzer project.

https://github.com/nunit/nunit.analyzers/issues/157

I'll post this comment to the issue too so that if we close this PR, it will also be tracked there.

manfred-brands:
@rprouse I have implemented the DiagnosticSuppressor in NUnit.Analyzers and it works a charm.

PR #3686, opened 2020-12-03
Initial implementation for looking for property defined in interface

### LINKED ISSUE #1394: Has.Property cannot see explicit interface implementation properties
Using nunit.framework 3.2 and nunit3-console 3.0.1.

Code:

```
using NUnit.Framework;

namespace NUnitHasProperty
{
    public interface IFoo
    {
        string Description { get; }
    }

    public class ImplicitFoo : IFoo
    {
        public string Description => "Foo";
    }

    public class ExplicitFoo : IFoo
    {
        string IFoo.Description => "Foo";
    }

    [TestFixture(typeof(ImplicitFoo))]
    [TestFixture(typeof(ExplicitFoo))]
    public class HasPropertyTestCase<TFoo>
        where TFoo : IFoo, new()
    {
        [Test]
        public void DescriptionIsFoo()
        {
            IFoo thing = new TFoo();
            Assert.That(thing, Has.Property(nameof(IFoo.Description)).EqualTo("Foo"));
        }
    }
}
```

Results:

```
NUnit Console Runner 3.0.5813

Runtime Environment
   OS Version: Microsoft Windows NT 10.0.10240.0
  CLR Version: 4.0.30319.42000

Test Files
    NUnitHasProperty.dll


Errors and Failures

1) Error : NUnitHasProperty.HasPropertyTestCase<ExplicitFoo>.DescriptionIsFoo
System.ArgumentException : Property Description was not found
Parameter name: name
   at NUnit.Framework.Constraints.PropertyConstraint.ApplyTo[TActual](TActual actual)
   at NUnitHasProperty.HasPropertyTestCase`1.DescriptionIsFoo() in C:\temp\NUnitHasProperty\NUnitHasProperty\HasPropertyTestCase.cs:line 29

Run Settings
    WorkDirectory: C:\temp\NUnitHasProperty\NUnitHasProperty\bin\Debug
    NumberOfTestWorkers: 12

Test Run Summary
    Overall result: Failed
   Tests run: 2, Passed: 1, Errors: 1, Failures: 0, Inconclusive: 0
  Start time: 2016-04-06 04:58:54Z
    End time: 2016-04-06 04:58:54Z
    Duration: 0.085 seconds

```

MichalBrylka:
Fixes #1394


src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
```diff
@@ -23,6 +23,7 @@
 
 using System;
 using System.Reflection;
+
```
jnm2, line 26: We don't format our code this way; could you please undo this line?

I think we should consider adding https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/formatting-rules#organize-using-directives to our .editorconfig file because Ctrl+R, Ctrl+G is a useful habit to clean up leftovers, but it behaves differently for different people if our .editorconfig doesn't specify what to do.

src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
```diff
@@ -56,21 +57,33 @@ public override ConstraintResult ApplyTo<TActual>(TActual actual)
...
+            const BindingFlags bindingFlags = BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance;
 
-            Type actualType = actual as Type;
-            if (actualType == null)
-                actualType = actual.GetType();
+            PropertyInfo property = Reflect.GetUltimateShadowingProperty(typeof(TActual), name, bindingFlags);
 
-            PropertyInfo property = Reflect.GetUltimateShadowingProperty(actualType, name,
-                BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance);
+            if (property == null && typeof(TActual).IsInterface && typeof(TActual).GetInterfaces() is { } interfaces && interfaces.Length > 0)
```
jnm2, line 64: This null check can be removed because Type.GetInterfaces() is documented to return an empty array rather than null, following the general pattern.
```suggestion
            if (property == null && typeof(TActual).IsInterface && typeof(TActual).GetInterfaces() is var interfaces && interfaces.Length > 0)
```

jnm2: I actually see the length check as redundant with the foreach loop, so I'd loop over `typeof(TActual).GetInterfaces()` directly.

src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
jnm2: Same for the two blank lines added in the other file.

src/NUnitFramework/tests/Constraints/PropertyTests.cs
```diff
@@ -206,4 +208,68 @@ public void FailureMessageContainsChainedConstraintMessage()
...
+        {
+            public int SomeProperty { get; set; } = 42;
+        }
+
+
+        private int[] _array;
+        private PropertyConstraint _countPropertyConstraint;
+
+        [SetUp]
+        public void BeforeEveryTest()
```
jnm2, line 237: I really like this name.

src/NUnitFramework/tests/Constraints/PropertyTests.cs
```diff
@@ -206,4 +208,68 @@ public void FailureMessageContainsChainedConstraintMessage()
...
+        {
+            _array = new[] { 1, 2, 3 };
+            _countPropertyConstraint = new PropertyConstraint(nameof(IList<object>.Count), new EqualConstraint(_array.Length));
+        }
+
+        [Test]
+        public void PropertyExists_PropertyDefinedInInterface_ShouldNotExist_WhenNotExplicitlyDefined()
```
jnm2, line 244: The word 'explicitly' could be confusing to future readers because arrays explicitly rather than implicitly implement the Count property. What do you think about:
```suggestion
        public void ExplicitlyImplementedProperty_ShouldNotExist_ViaImplementingType()
```

src/NUnitFramework/tests/Constraints/PropertyTests.cs
```diff
@@ -206,4 +208,68 @@ public void FailureMessageContainsChainedConstraintMessage()
             Assert.IsTrue(textMessageWriter.ToString().Contains("Extra"));
         }
     }
+
+    [TestFixture(TestOf = typeof(PropertyConstraint))]
+    public class PropertyHierarchyTests
```
jnm2, line 213: I like the idea of either having the tests in the same test class or starting this class name with `PropertyConstraint` and using the refactoring to move it to its own file. Maybe `TypeHierarchy` too instead of just `Hierarchy` to be extra clear. What do you think?

src/NUnitFramework/tests/Constraints/PropertyTests.cs
jnm2: (We don't usually use the same prefix on a bunch of tests like `PropertyExists_`. Do you feel we would lose too much if we don't use it?)

jnm2 (dismissed):
Really clear, good stuff.

src/NUnitFramework/tests/Constraints/PropertyTests.cs
```diff
@@ -206,4 +208,68 @@ public void FailureMessageContainsChainedConstraintMessage()
...
+
+
+        private int[] _array;
+        private PropertyConstraint _countPropertyConstraint;
+
+        [SetUp]
+        public void BeforeEveryTest()
+        {
+            _array = new[] { 1, 2, 3 };
+            _countPropertyConstraint = new PropertyConstraint(nameof(IList<object>.Count), new EqualConstraint(_array.Length));
```
jnm2, line 240: Can you do a `_countPropertyExistsConstraint` in each test showing that `PropertyExistsConstraint` also behaves the same?

`((IResolveConstraint)Has.Property("X").EqualTo(42)).Resolve()` returns PropertyConstraint, whereas `((IResolveConstraint)Has.Property("X")).Resolve()` returns PropertyExistsConstraint.

src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
MichalBrylka: sure, done. I'd be cautious with MS declarations especially given you're probably compiling against multiple legacy targets, but I assume this time we can trust them ;-)

src/NUnitFramework/tests/Constraints/PropertyTests.cs
MichalBrylka: I'm ok with suggested name. You're right, "PropertyExists_" might have been an exaggeration

src/NUnitFramework/tests/Constraints/PropertyTests.cs
MichalBrylka: Moving to separate file is probably cleaner approach

src/NUnitFramework/tests/Constraints/PropertyTests.cs
MichalBrylka: ok, I can add additional preparation and assertion for each test that demonstrate property existence

MichalBrylka: now we have same logic in 2 different classes (PropertyConstraint and PropertyExistsConstraint). My rule of thumb says "extract common code paths when it appears in 3 or more places" - so I'd let this slip unless we have similar logic in more places (i.e. places that could also benefit from new reasoning about properties)

MichalBrylka:
Please check if that's to you liking now

rprouse (approved):
Looks good and all of @jnm2's change requests have been addressed, so approving.

PR #3697, opened 2020-12-23
Improve TextMessageWriter output for numeric values

### LINKED ISSUE #3563: [Suggestion] Improve TextMessageWriter output for numeric values
When comparing double values it is sometimes difficult to _see_ why the test failed. Example:
```cs
        [Test]
        public void CompareDoubleValues()
        {
            double actual = -25.148640873240467;
            double expected = -24.685244717451383;
            const double tolerance = 0.4;
            Assert.That(actual, Is.EqualTo(expected).Within(tolerance));
        }
```
results in the following message:
```
  Expected: -24.685244717451383d +/- 0.40000000000000002d
  But was:  -25.148640873240467d
```
It is immediately obvious, that the numbers differ. And you trust nunit to properly handle the tolerance. However, to fix the issue you have to understand _what's wrong with my number?_ Is my number too large? Oh wait, they are signed. It's actually too small. Was my tolerance reasonable? Well, what is the actual difference?

**Proposed Message**
If we have two values and a tolerance, it would be great to see the actual _difference_ as well. In my oppinion, that would improve the readability significantly for weird numbers or tiny tolerances.
```
  Expected: -24.685244717451383d +/- 0.40000000000000002d
  But was:  -25.148640873240467d  (differs by -0.463396155789084)
```
Of course I can output the difference on my own, e.g. using nunit's message argument. However, I think that extra work could be done once in nunit instead of _every time_ on the developer side.

**Remarks**
`nunit 3.12` (latest at the time of writing)
Visual Studio 2017 professional + ReSharper to run the tests
or dotnet core with `dotnet test`

I wanted to create a PR, but VS2017 cannot compile the solution.
The relevant parts of the code should be 
https://github.com/nunit/nunit/blob/fd6347133590b28b43de02953af216a47301c4a1/src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs#L96
and
https://github.com/nunit/nunit/blob/fd6347133590b28b43de02953af216a47301c4a1/src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs#L166-L175
but I might be wrong, since I couldn't step into the code.

Detection of numerical values is probably required to work with `object` arguments - there is nice code available in [Numerics.cs](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Constraints/Numerics.cs). This class could be extended with methods like `GetDifference(object expected, object actual, Tolerance tolerance)`.

PS: Thanks for this awesome test framework!

Antash:
Closes #3563

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Numerics.cs
+++ b/src/NUnitFramework/framework/Constraints/Numerics.cs
@@ -418,5 +418,75 @@ namespace NUnit.Framework.Constraints
             return Convert.ToInt32(expected).CompareTo(Convert.ToInt32(actual));
         }
         #endregion
+
+        #region Numeric Difference
+
+        /// <summary>
+        /// Calculates the difference between 2 values.
+        /// </summary>
+        /// <param name="expected">The expected value</param>
+        /// <param name="actual">The actual value</param>
+        /// <param name="tolerance">A reference to the tolerance in effect</param>
+        /// <returns>The difference between the values</returns>
+        public static object Difference(object expected, object actual, Tolerance tolerance)
+        {
+            switch (tolerance.Mode)
+            {
+                case ToleranceMode.Linear:
+                    return DifferenceAbsolute(expected, actual);
+                case ToleranceMode.Percent:
+                    return DifferencePercent(expected, actual);
+                default:
+                    throw new InvalidOperationException("Cannot calculate a difference for specified tolerance mode");
+            }
+        }
+
+        private static object DifferencePercent(object expected, object actual)
+        {
+            if (!IsNumericType(expected) || !IsNumericType(actual))
+                throw new ArgumentException("Both arguments must be numeric");
+
+            if (IsFloatingPointNumeric(expected) || IsFloatingPointNumeric(actual))
+                return (Convert.ToDouble(expected) - Convert.ToDouble(actual)) / Convert.ToDouble(expected) * 100;
+
+            if (expected is decimal || actual is decimal)
+                return (Convert.ToDecimal(expected) - Convert.ToDecimal(actual)) / Convert.ToDecimal(expected) * 100;
+
+            if (expected is ulong || actual is ulong)
+                return (Convert.ToUInt64(expected) - Convert.ToUInt64(actual)) / (double)Convert.ToUInt64(expected) * 100;
+
+            if (expected is long || actual is long)
+                return (Convert.ToInt64(expected) - Convert.ToInt64(actual)) / (double)Convert.ToInt64(expected) * 100;
+
+            if (expected is uint || actual is uint)
+                return (Convert.ToUInt32(expected) - Convert.ToUInt32(actual)) / (double)Convert.ToUInt32(expected) * 100;
+
+            return (Convert.ToInt32(expected) - Convert.ToInt32(actual)) / (double)Convert.ToInt32(expected) * 100;
+        }
+
+        private static object DifferenceAbsolute(object expected, object actual)
+        {
+            if (!IsNumericType(expected) || !IsNumericType(actual))
+                throw new ArgumentException("Both arguments must be numeric");
+
+            if (IsFloatingPointNumeric(expected) || IsFloatingPointNumeric(actual))
+                return Convert.ToDouble(expected) - Convert.ToDouble(actual);
+
+            if (expected is decimal || actual is decimal)
+                return Convert.ToDecimal(expected) - Convert.ToDecimal(actual);
+
+            if (expected is ulong || actual is ulong)
+                return Convert.ToUInt64(expected) - Convert.ToUInt64(actual);
+
+            if (expected is long || actual is long)
+                return Convert.ToInt64(expected) - Convert.ToInt64(actual);
+
+            if (expected is uint || actual is uint)
+                return Convert.ToUInt32(expected) - Convert.ToUInt32(actual);
+
+            return Convert.ToInt32(expected) - Convert.ToInt32(actual);
+        }
+
+        #endregion
     }
 }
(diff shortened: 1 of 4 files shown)


src/NUnitFramework/tests/Constraints/NumericsTest.cs
```diff
@@ -83,6 +84,21 @@ public void CanMatchDecimalWithPercentage()
             Assert.IsTrue(Numerics.AreEqual(10000m, 10500m, ref tenPercent));
         }
 
+        [Test]
+        public void CanCalculateAbsoluteDifference()
+        {
+            Assert.AreEqual(Numerics.Difference(10000m, 9500m, absoluteTolerance), 500m);
+            Assert.AreEqual(Convert.ToDouble(Numerics.Difference(0.1, 0.05, absoluteTolerance)), 0.05, 0.00001);
+            Assert.AreEqual(Convert.ToDouble(Numerics.Difference(0.1, 0.15, absoluteTolerance)), -0.05, 0.00001);
```
mikkelbu, line 92: I think the "expected" and the "actual" arguments have been swapped in this test (and in the test below). The signature of `Assert.AreEqual` is `Assert.AreEqual(double expected, double actual, double tolerance);`. On could also use the constraint based model instead, as this avoids the confusion of `Assert.AreEqual`. E.g. the first assertion could be written as `Assert.That(Numerics.Difference(10000m, 9500m, absoluteTolerance), Is.EqualTo(500m))`.

Also I reviewed this file first in the PR and was a bit confused how the tolerance would affect the the difference, e.g. what is the difference between `10000m` and `9500m` when the tolerance is `0.1` ? (I now know that we only use the mode of the tolerance, so I think it would make more sense just to pass in `ToleranceMode` and not `Tolerance` to `Difference`.

src/NUnitFramework/framework/Constraints/Numerics.cs
```diff
@@ -418,5 +418,75 @@ public static int Compare(object expected, object actual)
...
+        #region Numeric Difference
+
+        /// <summary>
+        /// Calculates the difference between 2 values.
+        /// </summary>
+        /// <param name="expected">The expected value</param>
+        /// <param name="actual">The actual value</param>
+        /// <param name="tolerance">A reference to the tolerance in effect</param>
+        /// <returns>The difference between the values</returns>
+        public static object Difference(object expected, object actual, Tolerance tolerance)
```
mikkelbu, line 431: I think it makes more sense to use `ToleranceMode` and not `Tolerance` (see comment in `NumericsTest.cs`). Could this method be internal?

src/NUnitFramework/framework/Constraints/Numerics.cs
```diff
@@ -418,5 +418,75 @@ public static int Compare(object expected, object actual)
...
+                case ToleranceMode.Linear:
+                    return DifferenceAbsolute(expected, actual);
+                case ToleranceMode.Percent:
+                    return DifferencePercent(expected, actual);
+                default:
+                    throw new InvalidOperationException("Cannot calculate a difference for specified tolerance mode");
+            }
+        }
+
+        private static object DifferencePercent(object expected, object actual)
```
mikkelbu, line 444: I was comparing this method to the method below, and I think it makes more sense just to have one method that handles both cases (rather than two the have more or less the same flow). This method could be passed e.g. a boolean to determine if the result should be absolute or in percent. Then the case for decimal would become:
```csharp
...
            if (expected is decimal || actual is decimal)
            {
                var difference = Convert.ToDecimal(expected) - Convert.ToDecimal(actual);
                return isAbsolute ? difference : difference / Convert.ToDecimal(expected) * 100;
            }
...
```

mikkelbu (commented):
Thanks for the contribution @antash. Overall, I think the code looks good, but I've some minor comments. Also I would like to manually test the code some more before approving, but this would be easiest when #3698 is merged and this PR is build, so I have a nuget-package to test (as currently my dev-machine is in a non-working state wrt. compiling this project 😄).

src/NUnitFramework/tests/Constraints/EqualConstraintTests.cs
```diff
@@ -463,6 +465,19 @@ public void FailsOnSinglesOutsideOfRelativeTolerance(object value)
...
+
+            [TestCase(1.21)]
+            [TestCase(1.19)]
+            public void FailsOnDoublesOutsideOfAbsoluteTolerance(object value)
+            {
+                const double tolerance = 0.001;
+                var ex = Assert.Throws<AssertionException>(() => Assert.That(value, new EqualConstraint(1.2).Within(tolerance)));
+                Assert.That(ex.Message, Does.Contain($"+/- {MsgUtils.FormatValue(tolerance)}"));
+                var expectedPercentDiff = 1.2 - (double)value;
+                Assert.That(ex.Message, Does.Contain($"{MsgUtils.FormatValue(expectedPercentDiff)}"));
```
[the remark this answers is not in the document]
Antash: good catch, thanks!

src/NUnitFramework/tests/Constraints/NumericsTest.cs
Antash: Agree! Refactored.

src/NUnitFramework/framework/Constraints/Numerics.cs
Antash: done

src/NUnitFramework/framework/Constraints/Numerics.cs
Antash: Methods merged, looks bulky a bit but at the same time more convenient. I think the best way is to rework all this stuff using generics.

rprouse:
@Antash I've created a PR in your repo to pull in the build fixes that @mikkelbu made. If you merge that, it will update this PR and get the builds passing for you. https://github.com/Antash/nunit/pull/1

rprouse (approved):
Looks good, you will need to merge my PR and get the build passing though.

Issue #3715, opened 2021-01-11
FixtureLifeCycle(LifeCycle.InstancePerTestCase) Not working with TestFixtureSource


xadvfh:
I'm using the new `FixtureLifeCycle(LifeCycle.InstancePerTestCase)` with `TestFixtureSource` and am running into an issue where a new instance is not created.

At first, I thought this might be an issue with WireMock.Net but it works fine without `TestFixtureSource`.  It also works fine when using an `IDisposable` class to simulate setup/teardown per test as @jnm2 has mentioned in the past.

In my tests, I am using [WireMock.Net](https://github.com/WireMock-Net/WireMock.Net) to create a new instance of the mock server for each test.  Before `FixtureLifeCycle`, I was starting the server in the constructor of an `IDisposable` class and stopping the server in the `Dispose` method.

In my tests below I'm just writing out the URL of the mock server to show that they are using the same url:port.  Each test should start a server on a unique port number but both tests are using the same port.

```c#
using NUnit.Framework;
using System;
using System.Collections;
using System.Linq;
using WireMock.Server;

namespace NUnitTestProject1
{
    [TestFixture]
    [Parallelizable(ParallelScope.All)]
    [FixtureLifeCycle(LifeCycle.InstancePerTestCase)]
    [TestFixtureSource(typeof(BrowserData))]
    public class TestBase
    {
        protected WireMockServer Server { get; set; }
        protected string driverId;

        public TestBase(string driver)
        {
            driverId = driver;
        }

        [SetUp]
        public void BeforeTest()
        {
            Server = WireMockServer.Start();
        }


        [TearDown]
        public void AfterTest()
        {
            Server?.Stop();
        }
    }


    public class Tests : TestBase
    {
        public Tests(string driver) : base(driver)
        {
        }

        [Test]
        public void Test1()
        {
            TestContext.Out.WriteLine(Server.Urls.First());
        }

        [Test]
        public void Test2()
        {
            TestContext.Out.WriteLine(Server.Urls.First());
        }
    }


    public class BrowserData : IEnumerable
    {
        private readonly string[] _drivers;
        public BrowserData()
        {
            var driverId = "LocalChrome";
            _drivers = Array.ConvertAll(driverId.Split(",", StringSplitOptions.RemoveEmptyEntries), d => d.Trim());
        }

        public IEnumerator GetEnumerator()
        {
            return _drivers.GetEnumerator();
        }
    }
}
```

gleb-osokin:
[...]

PR #3720, opened 2021-01-11
Fix FixtureLifeCycle application

### LINKED ISSUE #3716: Assembly level FixtureLifeCycle(LifeCycle.InstancePerTestCase) doesn't work
LifeCycle.InstancePerTestCase is not applied when set at the assembly level. 

Here's a sample test that fails:

```csharp
using NUnit.Framework;

[assembly: FixtureLifeCycle(LifeCycle.InstancePerTestCase)]

[TestFixture]
public class AssemblyLevelFixtureLifeCycleTest
{
    private int _value;

    [Test]
    public void Test1() => Assert.AreEqual(0, _value++);

    [Test]
    public void Test2() => Assert.AreEqual(0, _value++);
}

gleb-osokin:
Fixes #3716 and #3715 

I was not sure how to better test the assembly-level attribute without creating a dedicated project - current test only checks full .NET Framework (although the fix itself should work everywhere). I tried to use Roslyn for .NET Standard, but there were some dependencies conflicts, unfortunately, which I am not ready to resolve at the moment.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestFixtureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestFixtureAttribute.cs
@@ -267,6 +267,7 @@ namespace NUnit.Framework
         public IEnumerable<TestSuite> BuildFrom(ITypeInfo typeInfo, IPreFilter filter)
         {
             var fixture = _builder.BuildFrom(typeInfo, filter, this);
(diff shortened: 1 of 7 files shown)


rprouse (approved):
Thanks for taking care of this so quickly.

PR #3733, opened 2021-01-18
Don't manually install .NET 5.0

### LINKED ISSUE #3729: AppVeyor builds failing
The AppVeyor builds have starting failing again, so the solution in #3698 was only a temporary solution, as the failure is similar to the last failures (that netcoreapp3.1 builds fail with `error NETSDK1127: The targeting pack Microsoft.NETCore.App is not installed. Please restore and try again.` / `error NETSDK1127: The targeting pack Microsoft.WindowsDesktop.App is not installed. Please restore and try again.`).

I can see that a new AppVeyor image has been released yesterday - https://www.appveyor.com/updates/2021/01/16/ - which among other contains Visual Studio 2019 version 16.8.4 (see appveyor/ci#3603) which includes `.NET SDK 5.0.102`. So I think the problem is that our install scripts install an old version `5.0.101` on top of a newer version. At least https://github.com/dotnet/runtime/issues/45755#issuecomment-760929942 indicates that there could be issues in this area (but why we are the only one experiencing it I don't know
> @marcpopMSFT @vitek-karas @dsplaisted have we heard of such issues in the past where installing previous shipped SDKs change the behavior of the latest one installed and causes the discussed restore issues?

I'll see if removing the manual installation of `5.0.101` solves the problem (at least it should already be installed in the images we use).

rprouse:
Fixes #3729

A newer version of .NET 5.0 is installed on the latest AppVeyor build images so attempting to install an older version causes the build to fail.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,10 +4,6 @@ branches:
   except:
     - /travis-.*/
 
-install:
-  - ps: Invoke-WebRequest 'https://dot.net/v1/dotnet-install.ps1' -OutFile 'dotnet-install.ps1'
-  - ps: .\dotnet-install.ps1 -Version 5.0.101
-
 build_script:
   - ps: .\build.ps1 -Target "Appveyor" -Configuration "Release"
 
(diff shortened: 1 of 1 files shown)


rprouse:
Builds are passing, so I'm going to self approve and merge to get the builds working again as soon as possible.

PR #3734, opened 2021-01-18
Don't install .NET Core 5.x SDK on Azure Pipelines

### LINKED ISSUE #3724: Test start and end time should end with Z
The test start and end time are UTC and should be formatted as such so that they are parsed as UTC by the engine and runners. This causes problems in the Adapter with Explicit tests that are set to `DateTime.MinTime`. They are getting parsed as local which makes the date negative throwing an exception. 

See nunit/nunit3-vs-adapter#811

```
<NUnitTestEvent>
<test-case id='0-1001' name='ExplicitTest' fullname='NUnit3VSIssue811.Explicit.ExplicitTest' methodname='ExplicitTest' classname='NUnit3VSIssue811.Explicit' runstate='Runnable' seed='1980958818' result='Skipped' label='Explicit' site='Parent' start-time='0001-01-01T00:00:00.0000000' end-time='0001-01-01T00:00:00.0000000' duration='0.000000' asserts='0' parentId='0-1000'>
   <reason>
      <message><![CDATA[OneTimeSetUp: https://github.com/nunit/nunit3-vs-adapter/issues/811]]></message>
   </reason>
</test-case>
</NUnitTestEvent>
```

rprouse:
Fixes #3724

Removes the install of .NET Core 5.0.101 which is already installed for each OS. This will prevent errors installing an older version if they get updated in the future.

I left the 2.1 and 3.1 runtime installs on Linux and MacOS because the docs do not specify that they are installed.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -12,12 +12,6 @@ jobs:
     vmImage: windows-latest
   steps:
 
-  - task: UseDotNet@2
-    displayName: 'Install .NET Core SDK'
-    inputs:
-      version: 5.0.101
-      performMultiLevelLookup: true
-
   - powershell: .\build.ps1 --target=Test --test-run-name=Windows
     displayName: Build, test, and publish results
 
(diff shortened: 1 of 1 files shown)


jnm2:
Oh, good! I guess this means they fixed the lookup issue where installing 2.1 or 3.1 would cause it to no longer be able to find the preinstalled 5.0.

PR #3741, opened 2021-01-23
Fix date on Ignored and Explicit Tests

### LINKED ISSUE #2339: Wrong date format in Ignored TestFixtures
@alexamz commented on [Mon Jul 31 2017](https://github.com/nunit/nunit-console/issues/274)

Hi all,

First of all, many thanks to all the contributors to this great and useful project!

We have problems with the start-time and end-time dates in Ignored TestFixtures.
We are using the last version of the NUnit console, the 3.7.0.

This is an example of a test-case header in the xml generated by NUnit Console with a wrong formatted date:

>     <test-case id="0-1071" name="AfterStart_AClientMayConnect" fullname="Port.TCP.UnitTests.TcpServerFactoryTests.AfterStart_AClientMayConnect" methodname="AfterStart_AClientMayConnect" classname="Port.TCP.UnitTests.TcpServerFactoryTests" runstate="Runnable" seed="604943940" result="Skipped" label="Ignored" site="Parent" start-time="0001-01-01 00:00:00Z" end-time="0001-01-01 00:00:00Z" duration="0.000000" asserts="0">

On this example we can see that start-time and end-time dates have a wrong date  value "0001-01-01 00:00:00Z ". I would expect to have the date of the execution even if they are skipped. Moreover, if you try to publish these dates to TFS API it will fail because of the date value.

This issue is very easy to reproduce , just adding a [Ignore("some ignore reasons")] on a [TextFixture]. And running Nunit Console to generate the xml result output.

Many thanks!

Have a nice day

---

@ChrisMaddock commented on [Mon Jul 31 2017](https://github.com/nunit/nunit-console/issues/274#issuecomment-319094815)

The code for this lives in the NUnit Framework. I'll move this issue over, and comment further over there. 🙂

### LINKED ISSUE #3724: Test start and end time should end with Z

rprouse:
Continues the work started by @Poimen in ##3706 superseding that PR.

Fixes #2339
Fixes #3724

This fixes the Unit Tests for the fix completed by @Poimen, adds additional tests and fixes a bug in the NUnit2 result writer that these changes exposed.

The unit tests cover the changes but I've also tested by examining the test results and manually verifying the start/end/duration of ignored and explicit tests.

### DIFF
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -23,6 +23,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Reflection;
@@ -212,6 +213,18 @@ namespace NUnit.Framework.Internal
         /// The time the current test started in Ticks
         /// </summary>
         public long StartTicks { get; set; }
+
+        /// <summary>
+        /// The current duration of the test taken
+        /// </summary>
+        public double Duration
+        {
+            get
+            {
+                var tickCount = Stopwatch.GetTimestamp() - StartTicks;
+                return (double)tickCount / Stopwatch.Frequency;
+            }
+        }
 
         /// <summary>
         /// Gets or sets the current test result
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -215,7 +215,7 @@ namespace NUnit.Framework.Internal
         public long StartTicks { get; set; }
 
         /// <summary>
-        /// The current duration of the test taken
+        /// Gets the elapsed time for running the test in seconds
         /// </summary>
         public double Duration
         {
--- a/src/NUnitFramework/mock-assembly/MockAssembly.cs
+++ b/src/NUnitFramework/mock-assembly/MockAssembly.cs
@@ -114,7 +114,7 @@ namespace NUnit.Tests
         public class MockTestFixture
         {
             public const int Tests = 13;
-            public const int Suites = 1;
+            public const int Suites = 2;
 
             public const int Passed = 2;
 
@@ -155,10 +155,9 @@ namespace NUnit.Tests
             [Test, Ignore("Ignore Message")]
             public void IgnoreTest() { }
 
-            [Test, Ignore("Ignore testcase")]
-            [TestCase(1)]
-            [TestCase(2)]
-            [TestCase(3)]
+            [TestCase(1, Ignore = "Ignore testcase")]
+            [TestCase(2, Ignore = "Ignore testcase")]
+            [TestCase(3, Ignore = "Ignore testcase")]
             public void SkippedTest(int a)
             {
                 Assert.Pass();
--- a/src/NUnitFramework/mock-assembly/MockAssembly.cs
+++ b/src/NUnitFramework/mock-assembly/MockAssembly.cs
@@ -22,6 +22,7 @@
 // ***********************************************************************
 
 using System;
+using System.Collections.Generic;
 using System.Reflection;
 using NUnit.Framework;
 using NUnit.Framework.Internal;
@@ -155,13 +156,18 @@ namespace NUnit.Tests
             [Test, Ignore("Ignore Message")]
             public void IgnoreTest() { }
 
-            [TestCase(1, Ignore = "Ignore testcase")]
-            [TestCase(2, Ignore = "Ignore testcase")]
-            [TestCase(3, Ignore = "Ignore testcase")]
-            public void SkippedTest(int a)
+            [TestCaseSource(nameof(SkippedTestCaseData))]
+            public void SkippedTest(int _)
             {
                 Assert.Pass();
             }
+
+            public static IEnumerable<TestCaseData> SkippedTestCaseData => new[]
+            {
+                new TestCaseData(1).Ignore("Ignore testcase"),
+                new TestCaseData(2).Ignore("Ignore testcase"),
+                new TestCaseData(3).Ignore("Ignore testcase")
+            };
 
             [Test, Explicit]
             public void ExplicitTest() { }
--- a/src/NUnitFramework/mock-assembly/MockAssembly.cs
+++ b/src/NUnitFramework/mock-assembly/MockAssembly.cs
@@ -113,12 +113,12 @@ namespace NUnit.Tests
         [Category("FixtureCategory")]
         public class MockTestFixture
         {
-            public const int Tests = 10;
+            public const int Tests = 13;
             public const int Suites = 1;
 
             public const int Passed = 2;
 
-            public const int Skipped_Ignored = 1;
+            public const int Skipped_Ignored = 4;
             public const int Skipped_Explicit = 1;
             public const int Skipped = Skipped_Ignored + Skipped_Explicit;
 
@@ -154,6 +154,15 @@ namespace NUnit.Tests
 
             [Test, Ignore("Ignore Message")]
             public void IgnoreTest() { }
+
+            [Test, Ignore("Ignore testcase")]
+            [TestCase(1)]
+            [TestCase(2)]
+            [TestCase(3)]
+            public void SkippedTest(int a)
+            {
+                Assert.Pass();
+            }
 
             [Test, Explicit]
             public void ExplicitTest() { }
--- a/src/NUnitFramework/nunitlite.tests/NUnit2XmlOutputWriterTests.cs
+++ b/src/NUnitFramework/nunitlite.tests/NUnit2XmlOutputWriterTests.cs
@@ -113,7 +113,7 @@ namespace NUnitLite.Tests
         [TestCase("errors", MockTestFixture.Failed_Error)]
         [TestCase("failures", MockTestFixture.Failed_Other)]
         [TestCase("inconclusive", MockTestFixture.Inconclusive)]
-        [TestCase("not-run", MockTestFixture.Skipped+MockTestFixture.Failed_NotRunnable-MockTestFixture.Skipped_Explicit)]
+        [TestCase("not-run", MockTestFixture.Skipped+MockTestFixture.Failed_NotRunnable)]
         [TestCase("ignored", MockTestFixture.Skipped_Ignored)]
         [TestCase("skipped", MockTestFixture.Skipped-MockTestFixture.Skipped_Ignored-MockTestFixture.Skipped_Explicit)]
         [TestCase("invalid", MockTestFixture.Failed_NotRunnable)]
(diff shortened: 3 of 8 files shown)


src/NUnitFramework/mock-assembly/MockAssembly.cs
```diff
@@ -113,12 +113,12 @@ public static void Main(string[] args)
         [Category("FixtureCategory")]
         public class MockTestFixture
         {
-            public const int Tests = 10;
-            public const int Suites = 1;
+            public const int Tests = 13;
+            public const int Suites = 2;
```
rprouse, line 117: Adds three tests with the test cases below and one additional suite because the test cases form a suite.

src/NUnitFramework/nunitlite.tests/NUnit2XmlOutputWriterTests.cs
```diff
@@ -113,7 +113,7 @@ public void TestResults_AssemblyPathIsCorrect()
         [TestCase("errors", MockTestFixture.Failed_Error)]
         [TestCase("failures", MockTestFixture.Failed_Other)]
         [TestCase("inconclusive", MockTestFixture.Inconclusive)]
-        [TestCase("not-run", MockTestFixture.Skipped+MockTestFixture.Failed_NotRunnable-MockTestFixture.Skipped_Explicit)]
+        [TestCase("not-run", MockTestFixture.Skipped+MockTestFixture.Failed_NotRunnable)]
```
rprouse, line 116: This didn't match the code but happened to pass.

src/NUnitFramework/nunitlite/OutputWriters/NUnit2XmlOutputWriter.cs
```diff
@@ -89,7 +89,7 @@ private void InitializeXmlFile(ITestResult result)
             xmlWriter.WriteAttributeString("total", summary.TestCount.ToString());
             xmlWriter.WriteAttributeString("errors", summary.ErrorCount.ToString());
             xmlWriter.WriteAttributeString("failures", summary.FailureCount.ToString());
-            var notRunTotal = summary.SkipCount + summary.FailureCount + summary.InvalidCount;
+            var notRunTotal = summary.IgnoreCount + summary.ExplicitCount + summary.SkipCount + summary.InvalidCount;
```
rprouse, line 92: The old code didn't make sense to me. Why are failed tests not run? This seems more correct to me.

rprouse (commented):
Adding a few notes to help with reviewing. The first three files were changed by @Poimen, most of the remainder are my changes.

src/NUnitFramework/nunitlite/OutputWriters/NUnit2XmlOutputWriter.cs
OsirisTerje: The ResultSummary class has a method called ```NotRunCount``` which looks like:  
```
 public int NotRunCount
        {
            get { return InvalidCount + SkipCount + IgnoreCount + ExplicitCount;  }
        }
```
Seems like the same as is expressed here. Perhaps better to use that?

rprouse: Good idea @OsirisTerje, I made the change.

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -213,6 +214,18 @@ private set
         /// </summary>
         public long StartTicks { get; set; }
 
+        /// <summary>
+        /// The current duration of the test taken
```
mikkelbu, line 218: This sounds odd. Should this be `The current duration of the test` ?

src/NUnitFramework/nunitlite.tests/NUnit3XmlOutputWriterTests.cs
```diff
@@ -208,6 +208,55 @@ public void TestSuite_HasValidEndTimeAttribute()
...
+            string endTimeStr = RequiredAttribute(testCase, "end-time");
+
+            Assert.That(startTimeStr, Does.EndWith("Z"), "Explicit start-time is not UTC");
+            Assert.That(endTimeStr, Does.EndWith("Z"), "Explicit end-time is not UTC");
+
+            Assert.IsTrue(DateTime.TryParse(startTimeStr, out var startTime));
+            Assert.IsTrue(DateTime.TryParse(endTimeStr, out var endTime));
+
+            Assert.That(startTime, Is.InRange(testRunStartTime, testRunEndTime), "Explicit test cases should be set to approximately the start time of test suite");
+            Assert.That(endTime, Is.InRange(testRunStartTime, testRunEndTime), "Explicit test cases should be set to approximately the start time of test suite");
```
mikkelbu, line 257: Should this be the end time?
```suggestion
            Assert.That(endTime, Is.InRange(testRunStartTime, testRunEndTime), "Explicit test cases should be set to approximately the end time of test suite");
```
or perhaps `The end time of explicit test cases should be set to approximately the end time of test suite` (and similar change for start time.

src/NUnitFramework/nunitlite.tests/NUnit3XmlOutputWriterTests.cs
```diff
@@ -208,6 +208,55 @@ public void TestSuite_HasValidEndTimeAttribute()
...
+                Assert.IsTrue(DateTime.TryParse(startTimeStr, out var startTime));
+                Assert.IsTrue(DateTime.TryParse(endTimeStr, out var endTime));
+
+                Assert.That(startTime, Is.InRange(testRunStartTime, testRunEndTime), "Ignored test cases should be set to approximately the start time of test suite");
+                Assert.That(endTime, Is.InRange(testRunStartTime, testRunEndTime), "Ignored test cases should be set to approximately the start time of test suite");
+            }
+        }
+
+        [Test]
+        public void ExplicitTest_HasValidEndTimeAttribute()
```
mikkelbu, line 237:
```suggestion
        public void ExplicitTest_HasValidStartAndEndTimeAttributes()
```

src/NUnitFramework/nunitlite.tests/NUnit3XmlOutputWriterTests.cs
```diff
@@ -208,6 +208,55 @@ public void TestSuite_HasValidEndTimeAttribute()
             Assert.That(success, "{0} is an invalid value for end time", endTimeString);
         }
 
+        [Test]
+        public void IgnoredTestCases_HaveValidEndTimeAttribute()
```
mikkelbu, line 212:
```suggestion
        public void IgnoredTestCases_HaveValidStartAndEndTimeAttributes()
```

src/NUnitFramework/nunitlite.tests/NUnit3XmlOutputWriterTests.cs
```diff
@@ -208,6 +208,55 @@ public void TestSuite_HasValidEndTimeAttribute()
...
+        {
+            DateTime.TryParse(RequiredAttribute(topNode, "start-time"), out var testRunStartTime);
+            DateTime.TryParse(RequiredAttribute(topNode, "end-time"), out var testRunEndTime);
+
+            var testCaseNodes = suiteNode.SelectNodes("test-suite[@name='SkippedTest']/test-case");
+            Assert.That(testCaseNodes, Is.Not.Null);
```
mikkelbu, line 218: Should we also assert that Count is larger than 0 (otherwise this test could pass without actually checking the attributes).

src/NUnitFramework/nunitlite.tests/NUnit3XmlOutputWriterTests.cs
```diff
@@ -208,6 +208,55 @@ public void TestSuite_HasValidEndTimeAttribute()
...
+                string endTimeStr = RequiredAttribute(testCase, "end-time");
+
+                Assert.That(startTimeStr, Does.EndWith("Z"), "Ignored start-time is not UTC");
+                Assert.That(endTimeStr, Does.EndWith("Z"), "Ignored end-time is not UTC");
+
+                Assert.IsTrue(DateTime.TryParse(startTimeStr, out var startTime));
+                Assert.IsTrue(DateTime.TryParse(endTimeStr, out var endTime));
+
+                Assert.That(startTime, Is.InRange(testRunStartTime, testRunEndTime), "Ignored test cases should be set to approximately the start time of test suite");
+                Assert.That(endTime, Is.InRange(testRunStartTime, testRunEndTime), "Ignored test cases should be set to approximately the start time of test suite");
```
mikkelbu, line 232: See comment below about wording the message differently

src/NUnitFramework/nunitlite.tests/TextUIReportTests.cs
```diff
@@ -101,7 +101,13 @@ public void TestsNotRunTest()
                 "Tests Not Run\n\n" +
                 "1) Ignored : NUnit.Tests.Assemblies.MockTestFixture.IgnoreTest\n" +
                 "Ignore Message\n\n" +
-                "2) Explicit : NUnit.Tests.Assemblies.MockTestFixture.ExplicitTest\n\n"));
+                "2) Ignored : NUnit.Tests.Assemblies.MockTestFixture.SkippedTest(1)\n"+
+                "Ignore testcase\n\n"+
+                "3) Ignored : NUnit.Tests.Assemblies.MockTestFixture.SkippedTest(2)\n"+
+                "Ignore testcase\n\n"+
+                "4) Ignored : NUnit.Tests.Assemblies.MockTestFixture.SkippedTest(3)\n"+
+                "Ignore testcase\n\n"+
```
mikkelbu, line 109: Are we here (implicitly) testing the ordering of attributes on a method? We often remind users that the attributes are unordered, so one cannot depend on the ordering. Should we use a TestCaseSource to make the ordering deterministic? (although this will be a bit more verbose)

mikkelbu (dismissed):
Looks good. I have a handful of comments/suggestions/questions, but nothing major.

rprouse:
All excellent suggestions @mikkelbu, thanks for a very thorough review. I've incorporated all of your suggestions.

PR #3744, opened 2021-01-26
Apply FixtureLifeCycle to parametrized tests

### LINKED ISSUE #3743: Parametrized tests do not respect FixtureLifeCycle.InstancePerTestCase
This applies to (at least) TestCase, TestCaseSource, Theory and Values attributes.

Simple test example that fails:

```csharp
    public class FixtureWithTestCases
    {
        private int _counter;

        [TestCase(0)]
        [TestCase(1)]
        [TestCase(2)]
        [TestCase(3)]
        public void Test(int _)
        {
            Assert.AreEqual(0, _counter++);
        }
    }
```

gleb-osokin:
Fixes #3743


src/NUnitFramework/framework/Internal/Extensions/TestExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using NUnit.Framework.Interfaces;
+
+namespace NUnit.Framework.Internal.Extensions
+{
+    internal static class TestExtensions
+    {
+        public static bool HasLifeCycle(this ITest test, LifeCycle lifeCycle)
```
gleb-osokin, line 30: This method has a drawback that it will, perhaps, skip the overriden LifeCycle.SingleInstance fixture within parent LifeCycle.InstancePerTestCase.

Surprisingly, I've added some tests exactly for this case and they pass. Perhaps this is because nested fixture doesn't have wrapping fixture as parent, but I didn't investigate that.

Better solution would be to add something like new LifeCycle.Inherit and make it default one, but that would be a breaking change, so I left it like this for now.

rprouse:
I thought I had tested this in the original PR. I guess not ☹️ 

This looks good, but I'm going to pull this and test it locally to make sure nothing else gets broken.

Thanks for tackling this. I will get it reviewed soon.

rprouse (approved):
I've added unit tests for this and all of the other recently reported `FixtureLifecycle` bugs to a test solution that I created when first testing this feature. With your changes, all tests pass including the assembly level that you were worried about. Thanks for the help.

PR #3748, opened 2021-01-30
Create NuGet source packages, snupkg

### LINKED ISSUE #3745: After upgrading to NUnit 3.13.0 the debugger enters NUnit code despite having checked "Enable Just My Code"
This issue was originally reported by Mike in https://groups.google.com/g/nunit-discuss/c/f2yV4G_kdi8/m/ypWZfLSEDgAJ

> I recently started having an issue when debugging tests after upgrading to the new 3.13.0 version.  Using Visual Studio, it treats nunit.framework.dll as "User Code" and steps into each step of code. This did not happen when using 3.12.0, and I've uninstalled, reinstalled, deleted the NuGet package numerous times. When I downgrade back to 3.12.0, it's resolved, and then if I try and upgrade again -- same problem.

I guess the problem was introduced by #3464, and @jnm2 proposed the following as a solution

> I'm in favor of moving to the snupkg format then, since that uses the well-known nuget.org symbol server that has been built into VS for a while.

EDIT: Also reported to Microsoft in https://developercommunity2.visualstudio.com/t/The-debugger-enters-code-in-PackageRefer/1321266

rprouse:
Fixes #3745

- Partially reverts #3464
- Build portable PDBs
- Create snupkg files for uploading symbols
- Use sourcelink for the PDBs
- Upload snupkg to AppVeyor

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,4 +1,5 @@
 #tool NUnit.ConsoleRunner&version=3.10.0
+#tool GitLink&version=3.1.0
 
 //////////////////////////////////////////////////////////////////////
 // ARGUMENTS
@@ -160,6 +161,9 @@ DotNetCoreBuildSettings CreateDotNetCoreBuildSettings() =>
 MSBuildSettings CreateMsBuildSettings()
 {
     var settings = new MSBuildSettings { Verbosity = Verbosity.Minimal, Configuration = configuration };
+    
+    // Only needed when packaging	
+    settings.WithProperty("DebugType", "pdbonly");
 
     if (IsRunningOnWindows())
     {
@@ -274,14 +278,17 @@ var FrameworkFiles = new FilePath[]
     "mock-assembly.dll",
     "mock-assembly.exe",
     "nunit.framework.dll",
+    "nunit.framework.pdb",
     "nunit.framework.xml",
     "nunit.framework.tests.dll",
     "nunit.testdata.dll",
     "nunitlite.dll",
+    "nunitlite.pdb",
     "nunitlite.tests.exe",
     "nunitlite.tests.dll",
     "slow-nunit-tests.dll",
     "nunitlite-runner.exe",
+    "nunitlite-runner.pdb",
     "nunitlite-runner.dll",
     "Microsoft.Threading.Tasks.dll",
     "Microsoft.Threading.Tasks.Extensions.Desktop.dll",
@@ -329,10 +336,22 @@ Task("CreateImage")
             CopyDirectory(sourceDir, targetDir);
         }
     });
+    
+Task("GitLink")	
+    .IsDependentOn("CreateImage")	
+    .Description("Source-indexes PDBs in the images directory to the current commit")	
+    .Does(() =>	
+    {	
+        var settings = new GitLink3Settings	
+        {	
+            BaseDir = PROJECT_DIR	
+        };	
+        GitLink3(GetFiles($"{CurrentImageDir}**/*.pdb"), settings);	
+    });
 
 Task("PackageFramework")
     .Description("Creates NuGet packages of the framework")
-    .IsDependentOn("CreateImage")
+    .IsDependentOn("GitLink")
     .Does(() =>
     {
         CreateDirectory(PACKAGE_DIR);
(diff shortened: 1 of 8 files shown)


nuget/nunitlite/nunitlite.snupkg.nuspec
```diff
@@ -0,0 +1,33 @@
...
+    </packageTypes>
+    <authors>Charlie Poole, Rob Prouse</authors>
+    <owners>Charlie Poole, Rob Prouse</owners>
+    <license type="file">LICENSE.txt</license>
+    <projectUrl>https://nunit.org</projectUrl>
+    <repository type="git" url="https://github.com/nunit/nunit"/>
+    <iconUrl>https://cdn.rawgit.com/nunit/resources/master/images/icon/nunit_256.png</iconUrl>
+    <icon>icon.png</icon>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>Symbols for debugging into the NUnitLite assemblies.</description>
```
mikkelbu, line 18: Should we also add a `<language>en-US</language>` for consistency (as we have in nuget/framework/nunit.snupkg.nuspec)?

nuget/nunitlite/nunitlite.snupkg.nuspec
```diff
@@ -0,0 +1,33 @@
...
+    <copyright>Copyright (c) 2021 Charlie Poole, Rob Prouse</copyright>
+  </metadata>
+  <files>
+    <file src="..\..\nuget\icon.png" />
+    <file src="bin/net35/nunitlite.pdb" target="lib\net35" />
```
mikkelbu, line 23: Do we also need a `<file src="LICENSE.txt" />` or is this implicit from `<license type="file">LICENSE.txt</license>` ?

nuget/framework/nunit.snupkg.nuspec
```diff
@@ -0,0 +1,29 @@
...
+    <iconUrl>https://cdn.rawgit.com/nunit/resources/master/images/icon/nunit_256.png</iconUrl>
+    <icon>icon.png</icon>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>Symbols for debugging into the NUnit Framework assemblies.</description>
+    <language>en-US</language>
+    <copyright>Copyright (c) 2021 Charlie Poole, Rob Prouse</copyright>
+  </metadata>
+  <files>
+    <file src="..\..\nuget\icon.png" />
+    <file src="bin/net35/nunit.framework.pdb" target="lib\net35" />
```
mikkelbu, line 24: Same comment as for the other file

> Do we also need a <file src="LICENSE.txt" /> or is this implicit from <license type="file">LICENSE.txt</license> ?

mikkelbu (approved):
LGTM. I don't know that much about this area, but I compare the files with the exising nuspec files and with the change in #3464

nuget/nunitlite/nunitlite.snupkg.nuspec
rprouse: I don't think it matters, but added.

nuget/nunitlite/nunitlite.snupkg.nuspec
rprouse: You can't include anything but PDBs in the `snupkg`, so license and the icon aren't included and not working. I removed the icon and changed the license to `<license type="expression">MIT</license>`. I don't think much of this matters since these packages are not displayed.

mikkelbu:
I looked at the build log and remembered that Claire had a comment https://github.com/nunit/nunit/pull/3464#issuecomment-604465470 . This could indicate that this is not working correctly.

rprouse:
Thanks for the info @mikkelbu. I've updated to use sourcelink, but I am running into problems. If I build using Visual Studio, the PDBs are properly sourcelinked, but if I build using Cake, they are not. I'll probably need to continue this tomorrow.

build.cake
```diff
@@ -341,13 +347,35 @@ Task("PackageFramework")
...
+            BasePath = CurrentImageDir,
+            OutputDirectory = PACKAGE_DIR,
+            Symbols = true,
+            // snupkg is not yet supported by Cake, https://github.com/cake-build/cake/issues/2362
+            ArgumentCustomization = args => args.Append("-SymbolPackageFormat snupkg")
+        };
+
+        NuGetPack("nuget/framework/nunit.snupkg.nuspec", settings);
+        NuGetPack("nuget/nunitlite/nunitlite.snupkg.nuspec", settings);
+    });
```
jnm2, line 377: Too bad the same .nuspec files can't be reused and a single `nuget pack` command be run. I don't know enough about it yet to know if that's attainable when using a nuspec.

src/NUnitFramework/framework/nunit.framework.csproj
```diff
@@ -4,8 +4,15 @@
...
     <RootNamespace>NUnit.Framework</RootNamespace>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <DebugType>portable</DebugType>
+    <DebugSymbols>true</DebugSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
+  </ItemGroup>
```
jnm2, line 14: What do you think about moving these six new lines out of each csproj into the Directory.Build.Props file?

jnm2 (approved):
> Reinstate gitlink

Gitlink is a different, obsolete technology. This PR is introducing the use of Source Link for the first time. :+1:

If it was me, I'd use `<EmbedAllSources>true</EmbedAllSources>` rather than Source Link especially now that the PDBs are only downloaded when requested, but Source Link is the official recommendation usually.

If you *do* use Source Link, I would recommend adding `<EmbedUntrackedSources>true</EmbedUntrackedSources>`. (https://github.com/dotnet/sourcelink/blob/master/README.md#using-source-link-in-net-projects)

build.cake
rprouse: It may be possible. That was my first approach and I got errors, so I just went with this. If we can clean it up later I'd be cool, but I just want to get the release out. I'd hoped to have it done today.

src/NUnitFramework/framework/nunit.framework.csproj
rprouse: I had them there, then decided we only want to do this for the assemblies that we distribute, but it is cleaner, so I am moving them back. Thanks

PR #3753, opened 2021-02-01
Update version to 4.0 and package to 4.0.0-alpha-1

rprouse:
Preparation for working on NUnit 4.0 by updating the version. Any bug fixes for 3.x should be made against the `v3.13-dev` branch.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -17,8 +17,8 @@ var ErrorDetail = new List<string>();
 // SET PACKAGE VERSION
 //////////////////////////////////////////////////////////////////////
 
-var version = "3.13.2";
-var modifier = "";
+var version = "4.0.0";
+var modifier = "-alpha-1";
 
 var dbgSuffix = configuration == "Debug" ? "-dbg" : "";
 var packageVersion = version + modifier + dbgSuffix;
--- a/build.cake
+++ b/build.cake
@@ -57,6 +57,7 @@ var NUNITLITE_RUNNER_DLL = "nunitlite-runner.dll";
 
 // Test Assemblies
 var FRAMEWORK_TESTS = "nunit.framework.tests.dll";
+var EXECUTABLE_NUNITLITE_TEST_RUNNER_EXE = "nunitlite-runner.exe";
 var EXECUTABLE_NUNITLITE_TESTS_EXE = "nunitlite.tests.exe";
 var EXECUTABLE_NUNITLITE_TESTS_DLL = "nunitlite.tests.dll";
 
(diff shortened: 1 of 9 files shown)


rprouse:
So, it looks like we have something hardcoded somewhere to look for the 3.x version, probably to distinguish from 2.x? I need to check if it is in the framework or in the engine. I suspect the later because NUnitLite runse. This would suck because we'll need to coordinate an engine/console release. CC @ChrisMaddock 

```
1) Invalid : D:\a\1\s\bin\Release\net40\nunit.framework.tests.dll
No suitable tests found in 'D:\a\1\s\bin\Release\net40\nunit.framework.tests.dll'.
Either assembly contains no tests or proper test driver has not been found.
```

rprouse:
The code in question is in `NUnit3DriverFactory.cs`,

```c#
public bool IsSupportedTestFramework(AssemblyName reference)
{
    return NUNIT_FRAMEWORK.Equals(reference.Name, StringComparison.OrdinalIgnoreCase) && reference.Version.Major == 3;
}
```

For now while we decide how we want to handle this, I might switch the framework tests over to running with the framework `nunitlite-runner.exe`

mikkelbu (approved):
The change looks fine, and I think it is fine to run the tests using nunitlite for the time being.

build.cake
```diff
@@ -202,7 +203,8 @@ Task("Test45")
     {
         var runtime = "net45";
         var dir = BIN_DIR + runtime + "/";
-        RunNUnitTests(dir, FRAMEWORK_TESTS, runtime, ref ErrorDetail);
+        RunTest(dir + EXECUTABLE_NUNITLITE_TEST_RUNNER_EXE, dir, FRAMEWORK_TESTS, dir + "nunit.framework.tests.xml", runtime, ref ErrorDetail);
+        //RunNUnitTests(dir, FRAMEWORK_TESTS, runtime, ref ErrorDetail);
```
rprouse, line 207: Run tests with NUnitLite runner. I left the old code commented out so we could switch back easily.

src/NUnitFramework/nunitlite-runner/nunitlite-runner.csproj
```diff
@@ -1,22 +1,6 @@
-﻿<Project>
-  <PropertyGroup Condition="'$(OS)' == 'Windows_NT' and '$(TargetFramework)' != 'netcoreapp2.1'">
-    <UseWindowsFormsAndWPF>true</UseWindowsFormsAndWPF>
```
rprouse, line 3: @jnm2 this was causing no tests to be run and the app returning immediately. I think you added it so you could test the WPF and WinForms code, but when I ran at the CLI it returned immediately with no output. Even `nunitlite-runner.exe --help` and `nunitlite-runner.exe --version` didn't output anything. I don't think that was intended, so I've reverted for now. Maybe we need to wrap the tests that you intended to run in NUnitLite?

PR #3760, opened 2021-02-06
Remove .NET 3.5 and 4.0 targets

### LINKED ISSUE #3708: Drop .NET 4.0 Build Target
.NET 4.0 has been EOL for around five years, so I'd like to drop support for it. This would allow us to clean up our async code.

### LINKED ISSUE #3758: Drop .NET 3.5 Build Targets
Drop the .NET 3.5 build targets for the 4.0 release. Users can remain on 3.13.x releases with bugfixes or target their tests to 4.5+ if they want to upgrade.

rprouse:
Fixes #3708
Fixes #3758

- Removes .NET 3.5 and 4.0 targets from the build scripts, nuspecs and the project files
- Remove .NET 3.5 and 4.0 pragmas and code
- Remove TASK_PARALLEL_LIBRARY_API pragma as it is now supported on all platforms
- Remove NET45 pragmas that are no longer needed


rprouse:
Sorry for such a large review guys. At least it is mainly just code deletion. 😺

build.cake
```diff
@@ -367,12 +336,8 @@ Task("PackageZip")
 
         var zipFiles =
             GetFiles(CurrentImageDir + "*.*") +
-            GetFiles(CurrentImageDir + "bin/net35/**/*.*") +
-            GetFiles(CurrentImageDir + "bin/net40/**/*.*") +
             GetFiles(CurrentImageDir + "bin/net45/**/*.*") +
-            GetFiles(CurrentImageDir + "bin/netstandard1.4/**/*.*") +
             GetFiles(CurrentImageDir + "bin/netstandard2.0/**/*.*") +
-            GetFiles(CurrentImageDir + "bin/netcoreapp1.1/**/*.*") +
```
stevenaw, line 375: Good catch

stevenaw (commented):
Phew, that was a big review. Thanks for putting this all together @rprouse 
Most everything made sense, but I noticed a few things while searching the codebase for `3.5` on your branch that might still be lingering. GitHub doesn't make it easy (to my knowledge) to comment on lines too far from a changed line, but I think there's still a few things that could be removed in `RuntimeFramework` and `PlatformDetectionTests`:
- https://github.com/nunit/nunit/blob/ad49f27294bd0f2677d8699756c6ccb10df600f8/src/NUnitFramework/framework/Internal/RuntimeFramework.cs#L113-L114
- https://github.com/nunit/nunit/blob/ad49f27294bd0f2677d8699756c6ccb10df600f8/src/NUnitFramework/tests/Internal/PlatformDetectionTests.cs#L274

EDIT: Though not sure if we care about those within this PR if this is focused just on the compile-time stuff

rprouse:
Thanks for a thorough review @stevenaw, I really appreciate it. I've resolved all of your suggestions except the `DetectNet35()` test. There are many OS and runtime tests in that class that are no longer supported by NUnit, so I plan to do a full pass at the `Platform` attribute and related classes soon and clean them up in a separate PR.

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -34,7 +34,7 @@
 using System.Security;
 using NUnit.Framework.Internal.Abstractions;
 
-#if NET35 || NET40 || NET45
+#if NET45
```
jnm2, line 37: Possibly was meant to be `#if NETFRAMEWORK` (a bunch of other places as well)

src/NUnitFramework/tests/nunit.framework.tests.csproj
```diff
@@ -1,52 +1,28 @@
...
-    <TargetFrameworks>net35;net40;net45;net46;netcoreapp2.1;netcoreapp3.1;net5.0</TargetFrameworks>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(OS)' != 'Windows_NT'">
-    <TargetFrameworks>net35;net45;net46;netcoreapp2.1;netcoreapp3.1;net5.0</TargetFrameworks>
-  </PropertyGroup>
-
   <PropertyGroup>
     <RootNamespace>NUnit.Framework</RootNamespace>
+    <TargetFrameworks>net45;net46;netcoreapp2.1;netcoreapp3.1;net5.0</TargetFrameworks>
```
jnm2, line 5: Feels like something I would place above `RootNamespace`, but I don't have a rationale other than a vague sense of the primacy of the `TargetFrameworks` element.

jnm2 (approved):
Feels good!

stevenaw:
Thanks @rprouse for the quick response. That makes sense about tackling `Platform` separately

stevenaw (approved):

src/NUnitFramework/tests/nunit.framework.tests.csproj
rprouse: Done

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
rprouse: Good point @jnm2, most instances should be NETFRAMEWORK and I've changed as appropriate.

rprouse:
Damn, the latest builds started failing because .NET 5.0 is not installed on Linux or OSX on Azure. We just removed the install because newer versions were installed on those platforms! 👿

jnm2:
@rprouse I've had issues where simply installing an older runtime causes the preinstalled runtimes to no longer be found, so once I install one, I have to install all. Could that be it?

rprouse:
@jnm2 we removed the install of 5.0 a few weeks ago for just that reason, a newer version came pre-installed on the build images. Now it appears that it isn't installed anymore!

Issue #3588, opened 2020-06-28
Audit string equality comparisons and consider changes for v4


jnm2:
Many of .NET's System.String APIs default to CurrentCulture comparisons. Unintentional linguistic comparisons cause correctness and performance problems. @jskeet has found a couple of significant speed issues caused by NUnit using EndsWith without specifying StringComparison.Ordinal: https://github.com/nunit/nunit/pull/3583, https://github.com/nunit/nunit/issues/1555.

~~If you think it's a good idea, I would like to reference https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers. I would silence all rules using .editorconfig which we aren't prepared to fix and would specifically leave on these rules:~~

**Update:** In the current .NET SDK, we would instead do `<AnalysisLevel>6.0-minimal</AnalysisLevel>` or similar and turn *on* these rules.

https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1310 (Specify StringComparison for correctness)
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1309 (Use ordinal StringComparison)

For each case where we can't fix the CA warning due to backwards compatibility concerns, I'd prefer to use `#pragma warning disable` with a comment noting why. For v4 we could consider removing these exceptions.

stevenaw:
I did a cursory check for `EndsWith` a week or two ago, and I think most usages of that check for a constant string of symbols all in the ASCII range (similar to the `EndsWith("[]")` case covered in #3583  ). It's possible that most of these could be updated in a non-breaking way.

OsirisTerje:
Seems most have been done here, or are there more left?  
@stevenaw

stevenaw:
Good catch @OsirisTerje !
I think this was an issue we'd originally thought there to be more follow-up to do but I think that it can be closed. There may be a bit of release notes cleanup needed here too. I can try and tackle this in the next few days

PR #3770, opened 2021-02-15
Update internal string comparisons to use Ordinal when safe

stevenaw:
Contributes to #3588 

This PR tackles the low-hanging fruit of StartsWith/EndsWith/Contains of string comparisons within the framework. It focuses on cases where the argument is a string literal with all characters lying within the standard ASCII range.

The majority of the cases deal with either:
- Inspection of type/member names
- Detection of line breaks

### DIFF
--- a/src/NUnitFramework/framework/Constraints/PredicateConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/PredicateConstraint.cs
@@ -52,7 +52,7 @@ namespace NUnit.Framework.Constraints
             get
             {
                 var name = predicate.GetMethodInfo().Name;
-                return name.StartsWith("<")
+                return name.StartsWith("<", StringComparison.Ordinal)
                     ? "value matching lambda expression"
                     : "value matching " + name;
             }
--- a/src/NUnitFramework/framework/Interfaces/TNode.cs
+++ b/src/NUnitFramework/framework/Interfaces/TNode.cs
@@ -274,7 +274,7 @@ namespace NUnit.Framework.Interfaces
             Guard.ArgumentNotNullOrEmpty(xpath, nameof(xpath));
             if (xpath[0] == '/')
                 throw new ArgumentException("XPath expressions starting with '/' are not supported", nameof(xpath));
-            if (xpath.IndexOf("//") >= 0)
+            if (xpath.IndexOf("//", StringComparison.Ordinal) >= 0)
                 throw new ArgumentException("XPath expressions with '//' are not supported", nameof(xpath));
 
             string head = xpath;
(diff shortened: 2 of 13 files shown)


src/NUnitFramework/framework/Constraints/PredicateConstraint.cs
```diff
@@ -52,7 +52,7 @@ public override string Description
             get
             {
                 var name = predicate.GetMethodInfo().Name;
-                return name.StartsWith("<")
+                return name.StartsWith("<", StringComparison.Ordinal)
```
mikkelbu, line 55: I guess we could use the char-overload here?
```suggestion
                return name.StartsWith('<')
```

src/NUnitFramework/framework/Interfaces/TNode.cs
```diff
@@ -410,7 +410,7 @@ public NodeFilter(string xpath)
                 int lbrack = xpath.IndexOf('[');
                 if (lbrack >= 0)
                 {
-                    if (!xpath.EndsWith("]"))
+                    if (!xpath.EndsWith("]", StringComparison.Ordinal))
```
mikkelbu, line 413: I guess we could use the char overload here?

src/NUnitFramework/nunitlite/CommandLineOptions.cs
```diff
@@ -459,7 +459,8 @@ protected virtual void ConfigureOptions(bool allowInputFile)
 
         private bool LooksLikeAnOption(string v)
         {
-            return v.StartsWith("-") || v.StartsWith("/") && Path.DirectorySeparatorChar != '/';
+            return v.StartsWith("-", StringComparison.Ordinal)
+                || (v.StartsWith("/", StringComparison.Ordinal) && Path.DirectorySeparatorChar != '/');
```
mikkelbu, line 463:
```suggestion
            return v.StartsWith('-') || (v.StartsWith('/') && Path.DirectorySeparatorChar != '/');
```

mikkelbu (approved):
Looks good @stevenaw. I've added some suggestions, but overall it looks sound.

stevenaw:
Oh, thanks @mikkelbu ! Using the char overload makes sense, I'm not sure how I missed that the first time

stevenaw:
@mikkelbu Thanks again for your review. I've tried updating the single-length string StartsWith/EndsWith to use the char overload, but I think that's only a netstandard 2.1+ API. I receive a compiler error on the current build targets of net45 + netstandard2.

mikkelbu:
@stevenaw You are right (just took at look at the documentation). I'll merge this now. Thanks for the contribution 👍

PR #3775, opened 2021-02-23
add ClearOutput method to TestResult

antonBrits:
Adds the option to be able to clear testresult output.

**Use case:** 
The idea is based off of how Azure devops infers test results and only attaches logs for failed tests. 

So my use case would be in some teardown, clear output if test passes:
        [TearDown]
        public void BaseTearDown()
        {
            var result = TestExecutionContext.CurrentContext.CurrentResult;
            if (Equals(result.ResultState, ResultState.Success))
                result.ClearOutput();
        }

When working with very large (100s of MB!) reports, or tests with large outputs (up to 10MB, 400 lines, per test) , any sort of processing of the results can be very slow and often leads to memory issues.
So this would allow you to suppress/clear output for a test/fixture at any point _before_ it gets written to the results.
Saving a lot of MB in space and a lot of time in processing reports

Other use cases could be suppressing output from third party libraries, repeated test setups etc.

These changes don't affect any existing functionality and its completely up to the user to decide what logging is (or not) important to them.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -656,6 +656,18 @@ namespace NUnit.Framework.Internal
             return writer.ToString();
         }
 
+
+        /// <summary>
+        /// Clears any text output written to this result.
+        /// </summary>
+        public void ClearOutput()
+        {
+            lock (OutWriter)
+            {
+                _output.Clear();
+            }
+        }
+
         #endregion
 
         #region Helper Methods
--- a/src/NUnitFramework/tests/Internal/Results/TestResultOutputTests.cs
+++ b/src/NUnitFramework/tests/Internal/Results/TestResultOutputTests.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2014 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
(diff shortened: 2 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=3775) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=3775) before we can accept your contribution.<br/><br/>:x: antonBrits [sign now](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=3775)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/nunit/nunit?pullRequest=3775) it.</sub>

rprouse:
@antonBrits thanks for the contribution. We usually prefer if an issue is created for new features so we can discuss them as a team before you do the work coding.  We need to be careful before adding new features because once accepted we need to support them for a long time. We've accepted features in the past without thinking too deeply and they have ended up costing us a lot of time answering questions when people use them incorrectly or think they should work differently.

@nunit/framework-team what do people think about this feature? On the surface, I can't see an issue with the idea but it feels wrong to me. Can anyone think of ways that this would be used incorrectly that would cause people to report issues?

antonBrits:
sorry wasnt 100% sure on the process, @rprouse happy to move this to an issue and can create PR later if accepted?

rprouse:
Let's see what the team says first.

Dreamescaper:
I would prefer something like that to be a parameter at TestAdapter level, or even at trx logger level (which is used by Azure DevOps).

Unfortunately, my knowledge of test adapters or test loggers is very limited, so I have no idea if it is possible.

rprouse:
> I would prefer something like that to be a parameter at TestAdapter level, or even at trx logger level (which is used by Azure DevOps).

It could also be an attribute on tests?

rprouse:
@antonBrits I'm going to close this PR. While I can see a use for this I think it is too limited in scope to accept into the framework and incur the long time cost of maintaining it and explaining it to users. Don't let this discourage you, we appreciate contributions, but adding new features has a fairly high bar to meet.

PR #3779, opened 2021-02-27
Obsolete AreEqual methods with nullable numeric arguments for 3.13

### LINKED ISSUE #3449: Assert.AreEqual overloads for nullable double are not useful
We've had these four methods as far back as NUnit source history goes. The `double?` overloads are the same as the first two overloads except that they throw InvalidOperationException if you pass `null`. Since they can't be used to pass `null`, they don't do anything that the safer `double` overloads don't do.

Should we obsolete them or fix them? What would fixing them look like?

```cs
namespace NUnit.Framework
{
    public abstract class Assert
    {
        public static void AreEqual(double expected, double actual, double delta, string message, params object[] args);
        
        public static void AreEqual(double expected, double actual, double delta);
        
        public static void AreEqual(double expected, double? actual, double delta, string message, params object[] args);
        
        public static void AreEqual(double expected, double? actual, double delta);
    }
}
```

Antash:
Closes #3449

### DIFF
--- a/src/NUnitFramework/framework/Assert.Equality.cs
+++ b/src/NUnitFramework/framework/Assert.Equality.cs
@@ -66,10 +66,9 @@ namespace NUnit.Framework
         /// <param name="delta">The maximum acceptable difference between the the expected and the actual</param>
         /// <param name="message">The message to display in case of failure</param>
         /// <param name="args">Array of objects to be used in formatting the message</param>
+        [System.Obsolete("This method will be removed in NUnit 4.0.")]
         public static void AreEqual(double expected, double? actual, double delta, string? message, params object?[]? args)
         {
-            // TODO: https://github.com/nunit/nunit/issues/3449
-            //                                    ↓↓↓↓↓↓↓
             AssertDoublesAreEqual(expected, actual!.Value, delta, message, args);
         }
 
@@ -80,10 +79,9 @@ namespace NUnit.Framework
         /// <param name="expected">The expected value</param>
         /// <param name="actual">The actual value</param>
         /// <param name="delta">The maximum acceptable difference between the the expected and the actual</param>
+        [System.Obsolete("This method will be removed in NUnit 4.0.")]
         public static void AreEqual(double expected, double? actual, double delta)
         {
-            // TODO: https://github.com/nunit/nunit/issues/3449
-            //                                    ↓↓↓↓↓↓↓
             AssertDoublesAreEqual(expected, actual!.Value, delta, null, null);
         }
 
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/framework/Assert.Equality.cs
```diff
@@ -80,10 +79,9 @@ public static void AreEqual(double expected, double? actual, double delta, strin
         /// <param name="expected">The expected value</param>
         /// <param name="actual">The actual value</param>
         /// <param name="delta">The maximum acceptable difference between the the expected and the actual</param>
+        [System.Obsolete("This method is going to be removed in NUnit 4.0")]
```
jnm2, line 82: Could you say `will be` and add a period to make it a full sentence? (Full sentence because that's what it would have to be if we add a second sentence, as IIRC we do elsewhere.)

jnm2 (changes requested):

jnm2 (approved):

Dreamescaper (approved):

mikkelbu (approved):
Looks good @Antash. I was going to merge this, but the CI is failing (as handling of .NETCore 5 have changed yet again on Azure DevOps). I've made another PR that should fix this issue #3782.

jnm2:
I queued a new build for the merge of this PR on Azure Pipelines.

mikkelbu:
@jnm2 I looks like the new build does not build a merge against the HEAD of the branch. Azure DevOps builds `Merge e326073 into b91d97f` and `b91d97f` is the commit before the CI fix on `v3.13-dev`. I'll just try to mark the PR as a draft and back again to see if that tricks the CI.

rprouse:
I think we need to add the v-3.13 branch to the build script, https://github.com/nunit/nunit/blob/v3.13-dev/azure-pipelines.yml#L5

mikkelbu:
> I think we need to add the v-3.13 branch to the build script, https://github.com/nunit/nunit/blob/v3.13-dev/azure-pipelines.yml#L5

I think adding the v-3.13 branch to the build script will just run a build on the branch when we merge to it (or commit on it), so I don't think it solves this issue, but there is no harm in adding this. I'll make a PR. To rerun the check we can probably either close and reopen the PR (or force push or push an empty commit).

mikkelbu:
Closing and reopening PR to force new checks

mikkelbu:
It doesn't look like the branch build on appveyor ever started, and my closing/reopening only retriggered all the other checks (and I can only do a PR build or a master build in appveyor)

rprouse:
I am going to override and merge. Thanks for the contribution.

PR #3780, opened 2021-02-27
Remove AreEqual methods with nullable numeric arguments

### LINKED ISSUE #3449: Assert.AreEqual overloads for nullable double are not useful

Antash:
Closes #3449

### DIFF
--- a/src/NUnitFramework/framework/Assert.Equality.cs
+++ b/src/NUnitFramework/framework/Assert.Equality.cs
@@ -57,36 +57,6 @@ namespace NUnit.Framework
             AssertDoublesAreEqual(expected, actual, delta, null, null);
         }
 
-        /// <summary>
-        /// Verifies that two doubles are equal considering a delta. If the expected value is infinity then the delta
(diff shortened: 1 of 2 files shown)


jnm2 (approved):
Thanks for getting this done!

mikkelbu (approved):
LGTM. Thanks for the contribution @Antash 👍 . I'll merge it in now.

PR #3781, opened 2021-02-28
Add DefaultConstraint

### LINKED ISSUE #3457: Add DefaultConstraint
Sometimes it is needed to verify value type against default value. 
Common use case - verify that value in rest response is not empty, e.g. Guid (not empty) or enums (not Undefined value).

Current alternative - using EqualConstraint (e.g. `Is.Not.EqualTo(Guid.Empty)`, `Is.Not.EqualTo(default(MyEnumTypeName))`), which is not always convenient.

Details:
1. If actual is non-nullable value type - should pass if `actual == default`.

2. If actual is nullable value type - should pass if `actual` is equal to default value of underlying type, so the following code could be possible:
```
MyValueType? actual = ...;
Assert.That(actual, Is.Not.Null & Is.Not.Default);
```

Open question - what the expected behavior if nullable `actual` is null? Technically it is default value for nullable value type, but I think would make sense to fail it in this case, i.e. only account for underlying type default value.

3. Not sure about the behavior for reference types. Not sure it should be allowed at all. Another option - allow (passing when null), and provide roslyn analyzer with warning if used against reference type.

Dreamescaper:
Fixes #3457 .

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DefaultConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DefaultConstraint.cs
@@ -1,0 +1,27 @@
+using System.Runtime.CompilerServices;
+
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// Tests that the actual value is default value for type.
+    /// </summary>
+    public class DefaultConstraint : Constraint
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DefaultConstraint"/> class.
+        /// </summary>
+        public DefaultConstraint()
+        {
+            this.Description = "default";
+        }
+
+        /// <summary>
+        /// Applies the constraint to an actual value, returning a ConstraintResult.
+        /// </summary>
+        public override ConstraintResult ApplyTo<TActual>(TActual actual)
+        {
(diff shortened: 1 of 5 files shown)


src/NUnitFramework/framework/Constraints/DefaultConstraint.cs
```diff
@@ -0,0 +1,35 @@
...
+            this.Description = "default";
+        }
+
+        /// <summary>
+        /// Applies the constraint to an actual value, returning a ConstraintResult.
+        /// </summary>
+        public override ConstraintResult ApplyTo<TActual>(TActual actual)
+        {
+            var actualType = actual?.GetType();
+
```
manfred-brands, line 25: What is the reason for not using `typeof(TActual)`?
Is this because a lot of thing in NUnit get cast to `object` and `TActual` is not relevant?
The problem this leads to is that `int?` v = 0 has an actual type of `int` and returns true for `Is.Default`

Dreamescaper: You're right, I decided simply to use default(TActual) after discussion in original issue.

rprouse (approved):

jnm2:
Looks superb!

Issue #3786, opened 2021-03-05
NUnit with dotnet test results in O(n^2) filtering complexity


pakrym:
The usual pattern for `dotnet test` when invoking test is to pass an Or filter with a FullNameFilter for every test it wants to run.

To check if the Or filter matches the test we need to iterate over every child filter and do it for every test in the project that results in O(n^2) time complexity. It's not noticeable on small projects but is very noticeable on large ones (9000+ tests). 

One solution is to reduce the OrFilter that only has FullNameFilter to a HashSet-based filter that would reduce overall complexity to O(N). I prototyped it in https://github.com/nunit/nunit/pull/3787 

For real-world data, the https://github.com/azure/azure-sdk-for-net/blob/1814567d1f6cfb4905539e71d54377ab46ed6737/sdk/storage/Azure.Storage.Blobs takes 80 sec to run but only 40 with the fix from the https://github.com/nunit/nunit/pull/3787.

Considering tests themselves take ~ 25-30 sec to run it's a 5X filter time improvement.

mikkelbu:
Hi @pakrym 

I thought that your issue sounded like something I had seen somewhere, and after some digging I found nunit/nunit3-vs-adapter#660. Close to the end of this issue it is mentioned that version 4 of the adapter solves this issue (or at least to some degree). Would it be possible for you try out the prerelease of the adapter to see how it affects the performance?

That being said I think your contribution still provides value. @nunit/framework-team do you agree?

pakrym:
I'm happy to try the 4.0.0.

On the other hand, this might improve the performance enough that AssemblySelectLimit can be increased to a much higher number without a performance impact.

pakrym:
Ping @jnm2 @rprouse, happy to implement the optimization if you can comment with the way you'd like to see it.

rprouse:
@pakrym sorry for the delay. I would appreciate the help with this. I will try and take a closer look today or tomorrow.

rprouse:
@pakrym I've looked at your code so far and I'm fine with the approach you are taking. Thanks for the help.

kasobol-msft:
@rprouse My team suffers from this problem. I.e. we have large test suite that takes long time to spin. Could this be prioritized? What's the timeline to release nunit with this fix?

rprouse:
[...]

PR #3787, opened 2021-03-05
[Prototype] Optimize an Or filter that contains only FullNameFilters

pakrym:
This is a prototype of a fix for the https://github.com/nunit/nunit/issues/3786

Please tell me if this is the approach you are willing to take and I'll finish the implementation and tests.


src/NUnitFramework/framework/Internal/Filters/OrFullNameFilter.cs
```diff
@@ -0,0 +1,60 @@
...
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
+
+using System.Collections.Generic;
+using NUnit.Framework.Interfaces;
+
+namespace NUnit.Framework.Internal.Filters
+{
+    internal class OrFullNameFilter : TestFilter
```
pakrym, line 29: Another option is to encapsulate this behavior in the OrFilter itself.

src/NUnitFramework/framework/Internal/Tests/TestMethod.cs
```diff
@@ -135,7 +142,7 @@ public override TNode AddToXml(TNode parentNode, bool recursive)
         /// <value>A list of child tests</value>
         public override IList<ITest> Tests
         {
-            get { return new ITest[0]; }
+            get { return _emptyArray; }
```
pakrym, line 145: When filtering a large number of tests this allocates a lot of throwaway objects.

mikkelbu: Makes sense 👍

rprouse:
@pakrym sorry for the delay. This approach looks good to me.

src/NUnitFramework/framework/Internal/Filters/CompositeFilter.cs
```diff
@@ -46,16 +47,7 @@ public CompositeFilter()
...
         public CompositeFilter( params TestFilter[] filters )
         {
-            Filters = new List<TestFilter>(filters);
-        }
-
-        /// <summary>
-        /// Adds a filter to the list of filters
-        /// </summary>
-        /// <param name="filter">The filter to be added</param>
-        public void Add(TestFilter filter)
```
pakrym, line 56: We are adding logic to filter construction that relies on filters not changing after creation.

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -373,12 +373,13 @@ private void OnRunCompleted(object sender, EventArgs e)
         private int CountTestCases(ITest test, ITestFilter filter)
         {
             if (!test.IsSuite)
-                return 1;
+                return filter.Pass(test) ? 1: 0;
```
pakrym, line 376: Pass would walk both parents and children of the node. This optimization makes it so we only call Pass on the leaf nodes avoiding repeated enumeration and evaluation of descendants.

src/NUnitFramework/framework/Internal/Execution/WorkItemBuilder.cs
```diff
@@ -56,42 +56,52 @@ static public WorkItem CreateWorkItem(ITest test, ITestFilter filter, bool recur
         /// <param name="filter">The filter to be used in selecting any child Tests.</param>
         /// <param name="debugger">An <see cref="IDebugger" /> instance.</param>
         /// <param name="recursive">True if child work items should be created and added.</param>
+        /// <param name="root">True iа work item needs to be created unconditionally, if false  null would be returned for tests that don't match the filter</param>
         /// <returns></returns>
-        internal static WorkItem CreateWorkItem(ITest test, ITestFilter filter, IDebugger debugger, bool recursive = false)
+        internal static WorkItem CreateWorkItem(ITest test, ITestFilter filter, IDebugger debugger, bool recursive = false, bool root = true)
```
pakrym, line 61: Similar to the `CountTestCases` this change makes it so we call `filter.Pass()` only on non-suite nodes. We use the presence of children as an indicator that the node needs to be created.

pakrym: Except for the case where this is the first invocation of the method and WI always needs to be created.

src/NUnitFramework/framework/Internal/Filters/OrFilter.cs
```diff
@@ -53,10 +75,37 @@ internal class OrFilter : CompositeFilter
...
+                if (negated)
+                    return !Match(test) && !MatchParent(test);
+
+                return Match(test) || MatchParent(test) || MatchDescendant(test);
+            }
+
             if (negated)
-                return Filters.All(f => f.Pass(test, negated));
+            {
+                foreach (var filter in Filters)
```
pakrym, line 89: Avoid lambda allocation here and below.

pakrym:
Before time to first test ~4min

```
TpTrace Verbose: 0 : 27624, 1, 2021/04/13, 19:23:50.423, 12790168432201, testhost.dll, Version: 16.8.0

TpTrace Verbose: 0 : 27624, 9, 2021/04/13, 19:27:13.296, 12792197119808, testhost.dll, TestExecutionRecorder.RecordStart: Starting test: Azure.Storage.Blobs.Test.AppendBlobClientTests(False,V2019_02_02).AppendBlockAsync.
```

After:
Time to first test 6 secs.
```
TpTrace Verbose: 0 : 57308, 1, 2021/04/13, 19:44:28.071, 12802544872994, testhost.dll, Version: 16.8.0

TpTrace Verbose: 0 : 57308, 13, 2021/04/13, 19:44:34.169, 12802605823457, testhost.dll, TestExecutionRecorder.RecordStart: Starting test: Azure.Storage.Blobs.Test.AppendBlobClientTests(False,V2019_02_02).AppendBlockAsync.

File: src/NUnitFramework/framework/Internal/TestFilter.cs [hunk 7]
```diff
@@ -123,11 +124,15 @@ public bool MatchParent(ITest test)
...
         {
-            if (test.Tests == null)
+            var tests = test.Tests;
+            if (tests == null)
                 return false;
 
-            foreach (ITest child in test.Tests)
+            // Use for-loop to avoid allocating the enumerator
+            int count = tests.Count;
+            for (var index = 0; index < count; index++)
```
rprouse, line 133: Another interesting performance change that I wouldn't have thought about. I'll take your word that you benchmarked each of these changes 😄

rprouse (approved):
Great work @pakrym, thanks for tackling this. Those are some impressive performance gains.



I'd like to get one more set of eyes on this before we merge. @jnm2 or @mikkelbu do either of you have time to take a look? It is easier to review with whitespace ignored.

I am going to assign this to myself to cherry pick to master once it is merged.

mikkelbu:
I can take a look at tonight (aka in 8 hours time)

src/NUnitFramework/framework/Internal/Execution/WorkItemBuilder.cs
```diff
@@ -56,42 +56,52 @@ static public WorkItem CreateWorkItem(ITest test, ITestFilter filter, bool recur
         /// <param name="filter">The filter to be used in selecting any child Tests.</param>
         /// <param name="debugger">An <see cref="IDebugger" /> instance.</param>
         /// <param name="recursive">True if child work items should be created and added.</param>
+        /// <param name="root">True iа work item needs to be created unconditionally, if false  null would be returned for tests that don't match the filter</param>
```
mikkelbu, line 59:
```suggestion
        /// <param name="root"><see langword="true"/> if work item needs to be created unconditionally, if <see langword="false"/> <see langword="null"/> will be returned for tests that don't match the filter.</param>
```
and perhaps we should break the line somewhere

src/NUnitFramework/framework/Internal/Execution/WorkItemBuilder.cs
```diff
@@ -56,42 +56,52 @@ static public WorkItem CreateWorkItem(ITest test, ITestFilter filter, bool recur
...
         /// <returns></returns>
-        internal static WorkItem CreateWorkItem(ITest test, ITestFilter filter, IDebugger debugger, bool recursive = false)
+        internal static WorkItem CreateWorkItem(ITest test, ITestFilter filter, IDebugger debugger, bool recursive = false, bool root = true)
         {
+            // Run filter on leaf nodes only
+            // use the presense of leaf nodes as an indicator that parent need to be created
```
mikkelbu, line 64:
```suggestion
            // use the presence of leaf nodes as an indicator that parent need to be created
```

src/NUnitFramework/framework/Internal/Execution/WorkItemBuilder.cs
```diff
@@ -56,42 +56,52 @@ static public WorkItem CreateWorkItem(ITest test, ITestFilter filter, bool recur
...
+            CompositeWorkItem work = root ? new CompositeWorkItem(suite, filter): null;
 
             if (recursive)
             {
                 int countOrderedItems = 0;
 
                 foreach (var childTest in suite.Tests)
                 {
-                    if (filter.Pass(childTest))
+                    var childItem = CreateWorkItem(childTest, filter, debugger, recursive: true, root: false);
```
mikkelbu, line 84: Perhaps we should pass in `recursive` instead of `true`, to be more robust if we at some point make a small change to the expression in the if-statement.

mikkelbu (commented):
I only managed to review 6 out of 8 files today, but I'll proceed tomorrow.

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
CharliePoole: @jnm2 I haven't reviewed the PR but caught your mention of me and wanted to reply.

Yes, it does apply to all nodes. It probably not clear these days because the docs say that the filter applies to "tests". People accustomed to the Microsoft runners tend to think that a "test" means what we call a "test case".

The existing logic relies on the fact that the check is made at the top level only. Since we never call CountTestCases on a node we don't want to have included in the count, the filter check seems redundant... that is, __unless__ the other changes in this PR cause this private member to be called in some recursive way.

Although I haven't reviewed this, note that the purpose of the change is to make a particular type of filter -- a large number of test name filters joined by or -- more efficient. Of course, that's a filter, which no one familiar with how NUnit works should ever construct, but it's also how Microsoft constructs filters. They do that precisely because they want it to work with every runner possible. I've suggested before that the vs adapter is the place to make such a change because the adapter is the sole piece of code, which combines knowledge of how NUnit works with knowledge of how Microsoft works.

That said, it's true that the filtering logic could be more efficient. :smile:

CharliePoole: Wondering how you can filter looking only at leaf nodes and still handle Categories correctly.  That is, you can deduce the name of all the suites that contain a test case from the test case itself, but you can't do that with categories.

pakrym: >Wondering how you can filter looking only at leaf nodes and still handle Categories correctly.

`.Pass()` walks parents of the test as well as the node itself and the descendants. So calling it on the leaf node would check it's parents as well.

pakrym: >If we pass in a negative filter, then we will only match the current ITest and the parents, so I that setting this change will make the performance worse as far as I can tell (as we could short-circuit the checks if some of the parents fail the check).

Correct, but re-walking parents is comparatively cheap (there are usually few levels) to walking the descendants. This seems like a tradeoff worth making.

CharliePoole: That's OK then. I was going by your description of the change... i.e. that you would only look at leaf nodes. You're still looking at them, but indirectly through `Pass`.

I remember we had a reason to look at both ancestors and descendants of each node in the original V2 code. We wanted to avoid doing both but for some reason couldn't. Wish I could remember the reason!!! In any case, if you can make all tests pass, that reason might be obsolete. Still, if I can figure out what it was, I'll post here again.

mikkelbu: > Although I haven't reviewed this, note that the purpose of the change is to make a particular type of filter -- a large number of test name filters joined by or -- more efficient.

As far as I can tell, the changes made in this file and in all the other files in this PR, except for OrFilter.cs, improves the general performance (of filters) and are not related to the particular filter created by vsTest.

> In any case, if you can make all tests pass, that reason might be obsolete

I think this change is sound, but making all test pass is a very low bar at the moment, as we only have one test that tests `CountTestCases` with a non empty filter. So I would be much more confident if we had some more tests of this (I would also like to test this some more manually). Unfortunately, my entire weekend is packed, so I don't think I'll have much time for this before next week.

src/NUnitFramework/framework/Internal/Filters/OrFilter.cs
```diff
@@ -53,10 +75,37 @@ internal class OrFilter : CompositeFilter
...
+                {
+                    if (!filter.Pass(test, negated: true))
```
mikkelbu, line 91: I think I would prefer to use the variable `negated` rather than `true`.

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
pakrym: >  So I would be much more confident if we had some more tests of this

I'll add some.

src/NUnitFramework/framework/Internal/TestFilter.cs
```diff
@@ -172,16 +177,20 @@ public static TestFilter FromXml(TNode node)
...
                 case "or":
-                    var orFilter = new OrFilter();
+                    List<TestFilter> orChildFilters = new List<TestFilter>();
+
                     foreach (var childNode in node.ChildNodes)
-                        orFilter.Add(FromXml(childNode));
-                    return orFilter;
+                        orChildFilters.Add(FromXml(childNode));
+
+                    return new OrFilter(orChildFilters.ToArray());
```
stevenaw, line 193: Would it make sense to add an internal `IList<TestFilter>` overload on the constructors of `AndFilter`, `OrFilter`, and `CompositeFilter`? I am thinking it may save on the need to call `ToArray()` here + above since the `List<TestFilter>` can then be passed right through into the `ReadOnlyCollection<TestFilter>()` constructor

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
pakrym: >So I would be much more confident if we had some more tests of this

Added CountTestCases tests that match the assembly, category, fixture, and individual tests.

src/NUnitFramework/framework/Internal/TestFilter.cs
pakrym: I'm reluctant to do more API changes that slightly required. The ToArray call while unfortunate is comparably cheap to other things that happen during the test execution.

stevenaw: Sure, that makes sense. Thanks for considering the feedback and for your response @pakrym .
I'm excited to see this change land, it's a huge improvement.

rprouse:
Changes look good. @nunit/framework-team any final reviews before I merge?

mikkelbu:
I'm in the process of manually testing some larger examples on `master` and on this branch to see if I get the same results. Hopefully, it will not take more than 30 minues.

mikkelbu:
I created the following tests (100 classes containing 100 tests each)

```csharp
    [Category("TestClass0")]
    [Category("Test0")]
    public class Test0
    {
        [Category("Test0")]
        [Test]
        public void Test_0()
...
        [Category("Test99")]
        [Test]
        public void Test_99()
        { }
    }
...
    [Category("TestClass99")]
    [Category("Test99")]
    public class Test99
    {
...
    }
````
and ran the following tests
```csharp
            Assert.That(_runner.CountTestCases(new OrFilter(
                new CategoryFilter("Test19"),
                new CategoryFilter("Test22"))), Is.EqualTo(396));

            Assert.That(_runner.CountTestCases(new OrFilter(
                new CategoryFilter("Test0"),
                new CategoryFilter("Test1"),
                new CategoryFilter("Test2")
                )), Is.EqualTo(591));

            Assert.That(_runner.CountTestCases(new OrFilter(
                new MethodNameFilter("Test_37"),
                new CategoryFilter("Test2")
                )), Is.EqualTo(298));

            Assert.That(_runner.CountTestCases(new OrFilter(
                new ClassNameFilter("NUnit.Tests.Test2") ,
                new ClassNameFilter("NUnit.Tests.Test3")
                )), Is.EqualTo(200));

            Assert.That(_runner.CountTestCases(new NotFilter(new OrFilter(
                new CategoryFilter("Test2"),
                new CategoryFilter("Test3")
                ))), Is.EqualTo(9637));
```
and got the same results on NUnit `master` and on this branch.

mikkelbu (approved):

PR #3794, opened 2021-03-12
Replace Gitter badge with Slack invite link

jnm2:
We discussed moving recently, and we already have better availability on Slack. Is it time to publicize this change better? Should we mention the change anywhere else?


CharliePoole:
@jnm2 Point of Order here, since you have asked for the Core Team to review. Although GitHub would mark this as OK to merge if one of us approved it, it's really a framework issue, and I think the Framework Team needs to approve it. Only three of the six Core Team members are on the Framework Team and those of us who are not should stay out of it unless it somehow becomes rephrased as a more general issue.

jnm2:
Well, now I don't know what to do, and I don't have enough extra time right now to think about it. I don't know that this is a good fit to tag as either core team or framework team. Who is responsible for Gitter overall? Do I have to obtain an answer to that question before I know how and where to open an issue/PR?

rprouse (approved):
Looks good to me. This is in the framework repo and the framework team isn't hanging out in Gitter anymore, so I think this is a good idea.

CharliePoole:
I can only give you my answer as I understand it: I don't believe anybody is responsible for Gitter "overall". I could be wrong, I may have missed some discussion. Is that what you meant by "we discussed moving recently?"

I think it's fine if you want to change what you are using. Some other NUnit projects are also using the same Gitter room. IMO if you just want to change what NUnit Framework is doing, then Rob's OK is all you need. Reaching out tor the Core team made me think you were trying to change more than just the framework.

Charlie

jnm2:
Okay, that approach makes sense to me. I'll submit PRs repo by repo and we probably don't need to specially coordinate that, so I won't need to tag any teams.

PR #3804, opened 2021-03-29
Fix IDisposable call for InstancePerTestCase with parametrized test cases

### LINKED ISSUE #3773: IDisposable not working with InstancePerTestCase
```
using System;
using System.Buffers;
using NUnit.Framework;

namespace Foo
{
    [TestFixture(256)]
    [TestFixture(1024)]
    [Parallelizable(scope: ParallelScope.All)]
    public class FixtureTests : IDisposable
    {
        private byte[] _buffer;

        public FixtureTests(int bufferSize) => _buffer = ArrayPool<byte>.Shared.Rent(bufferSize);

        public void Dispose() => ArrayPool<byte>.Shared.Return(_buffer);

        [TestCase(0)]
        [TestCase(256)]
        public void Foo_Test(int size)
        {
            Assert.AreEqual(size, size);
        }
    }
}
```

I would expect the above to create and dispose the fixture per test case but it is resulting in an exception from NUnit

```
Failed Foo_Test(0) [127 ms]
  Error Message:
   System.InvalidOperationException : DisposeFixtureCommand does not apply neither to TestMethod, nor to ParameterizedMethodSuite
  Stack Trace:
     at NUnit.Framework.Guard.OperationValid(Boolean condition, String message)
   at NUnit.Framework.Internal.Commands.DisposeFixtureCommand..ctor(TestCommand innerCommand)
   at NUnit.Framework.Internal.Execution.SimpleWorkItem.MakeTestCommand()
   at NUnit.Framework.Internal.Execution.SimpleWorkItem.PerformWork()
```

isn't this expected to work this way?

gleb-osokin:
Fixes #3773

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/DisposeFixtureCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/DisposeFixtureCommand.cs
@@ -1,7 +1,6 @@
-﻿// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System;
-using System.Collections.Generic;
 using NUnit.Framework.Interfaces;
 
 namespace NUnit.Framework.Internal.Commands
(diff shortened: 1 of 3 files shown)


Dreamescaper (approved):
LGTM.
Thanks for contribution!

rprouse (approved):

PR #3810, opened 2021-04-06
Enable deterministic build

304NotModified:
Current result:

image

Desired result:

Deterministic Build, see https://github.com/clairernovotny/DeterministicBuilds

Preview NuGet packages: https://ci.appveyor.com/project/304NotModified/nunit/builds/38580291/artifacts

And green check :)

image

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -159,12 +159,13 @@ DotNetCoreBuildSettings CreateDotNetCoreBuildSettings() =>
 
 MSBuildSettings CreateMsBuildSettings()
 {
-    var settings = new MSBuildSettings { 
-        // EmbedUntrackedSources and ContinuousIntegrationBuild for deterministic build
-        ArgumentCustomization = args => args.Append("-p:EmbedUntrackedSources=true -p:ContinuousIntegrationBuild=true"),
-        Verbosity = Verbosity.Minimal, 
-        Configuration = configuration
-    };
+    var settings = new MSBuildSettings { Verbosity = Verbosity.Minimal, Configuration = configuration };
+
+    if (BuildSystem.IsRunningOnAppVeyor)
+    {
+        // Extra arguments for NuGet package creation on AppVeyor: EmbedUntrackedSources and ContinuousIntegrationBuild for deterministic build
+        settings.ArgumentCustomization = args => args.Append("-p:EmbedUntrackedSources=true -p:ContinuousIntegrationBuild=true");
+    }
 
     if (IsRunningOnWindows())
     {
--- a/build.cake
+++ b/build.cake
@@ -159,7 +159,12 @@ DotNetCoreBuildSettings CreateDotNetCoreBuildSettings() =>
 
 MSBuildSettings CreateMsBuildSettings()
 {
-    var settings = new MSBuildSettings { Verbosity = Verbosity.Minimal, Configuration = configuration };
+    var settings = new MSBuildSettings { 
+        // EmbedUntrackedSources and ContinuousIntegrationBuild for deterministic build
+        ArgumentCustomization = args => args.Append("-p:EmbedUntrackedSources=true -p:ContinuousIntegrationBuild=true"),
+        Verbosity = Verbosity.Minimal, 
+        Configuration = configuration
+    };
 
     if (IsRunningOnWindows())
     {
--- a/build.cake
+++ b/build.cake
@@ -161,9 +161,9 @@ MSBuildSettings CreateMsBuildSettings()
 {
     var settings = new MSBuildSettings { Verbosity = Verbosity.Minimal, Configuration = configuration };
 
-    if (BuildSystem.IsRunningOnAppVeyor)
+    if (!BuildSystem.IsLocalBuild)
     {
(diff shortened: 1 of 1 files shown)


304NotModified:
PS: I assumed the latest dev branch, but not sure, can't find that info - checked https://github.com/nunit/nunit/blob/master/CONTRIBUTING.md

build.cake
```diff
@@ -159,7 +159,12 @@ DotNetCoreBuildSettings CreateDotNetCoreBuildSettings() =>
 
 MSBuildSettings CreateMsBuildSettings()
 {
-    var settings = new MSBuildSettings { Verbosity = Verbosity.Minimal, Configuration = configuration };
+    var settings = new MSBuildSettings { 
+        // EmbedUntrackedSources and ContinuousIntegrationBuild for deterministic build
+        ArgumentCustomization = args => args.Append("-p:EmbedUntrackedSources=true -p:ContinuousIntegrationBuild=true"),
```
jnm2, line 164: I usually put EmbedUntrackedSources in Directory.Build.props, but this works too.

For ContinuousIntegrationBuild: we sometimes need to debug specifically the builds created by this build script. Since this setting messes with debuggability, https://github.com/clairernovotny/DeterministicBuilds recommends that we make this conditional on environment variables like `CI`  or `TF_BUILD`. However, we can't make it conditional if the NuGet package that we publish is built locally and no special switch is passed to the script.

@rprouse Can you confirm whether there's a way we can condition this so that what ends up on nuget.org has ContinuousIntegrationBuild set, but local builds via build.cake do not have this set? Or for now, should we forgo debuggability of all local builds created via build.cake, or maybe just all builds with a release configuration?

304NotModified: > I usually put EmbedUntrackedSources in Directory.Build.props, but this works too.

yes that's fine too indeed

> For ContinuousIntegrationBuild: we sometimes need to debug specifically the builds created by this build script. Since this setting messes with debuggability, https://github.com/clairernovotny/DeterministicBuilds recommends that we make this conditional on environment variables like CI or TF_BUILD. However, we can't make it conditional if the NuGet package that we publish is built locally and no special switch is passed to the script.

I could also use some env variable or pass it from the .appveyor config. But I assumed the build.cake is only used at the CI and not locally - maybe that's a bad assumption?

jnm2: I have personally used it a number of times to debug release builds to track down elusive problems like thread aborts.

mikkelbu: I also (often) use it locally.

In the cake script we already use `BuildSystem.IsRunningOnAppVeyor` to make special `packageVersion`. As far as I can see - https://cakebuild.net/api/Cake.Common.Build/BuildSystem/ - there is no `BuildSystem.IsRunningOnCI`, so we can either check for both `IsRunningOnAppVeyor` and `IsRunningOnAzurePipelines`, but I would prefer to negate `IsLocalBuild` to determine if we are running in CI.

304NotModified: Is the package also created on Azure? Or only on AppVeyor?

304NotModified: I think IsRunningOnAppVeyor is the best, as the .cake file already uses that condition for the package version. So that's a consistent solution

mikkelbu: I just took a closer look. According to https://docs.nunit.org/articles/developer-info/Packaging-the-Framework.html we use the packages from Azure DevOps in the releases - and we use the packages from AppVeyor in the prereleases - so I think we should handle both cases.

304NotModified: Fixed and pushed

304NotModified: If a squash is needed, let me know. :)

jnm2: @304NotModified As Mikkel is saying, IsRunningOnAppVeyor will only be true for our prereleases (MyGet) and Azure DevOps is where the nuget.org releases come from. I also agree with his suggestion of using `!IsLocalBuild`. That would be less brittle if we move something to GitHub Actions etc.

304NotModified: ow that is unexpected! So the prerelease is build on another system than the file. Sounds like an issue to me.

304NotModified: Updated, but not sure if I could validate the Azure Devops output

rprouse: @304NotModified we originally used AppVeyor for the majority of our bulds, but then began migrating over to Azure DevOps because it builds on multiple OS's. We haven't gotten around to switching the deployment of our nightly builds as we are a small volunteer team with limited time. Pre-release is a bit of a misnomer, if we were going to release an official pre-release, it would be build on Azure DevOps and deployed to NuGet.org.

So, the priority for me for this PR would be to apply this on Azure DevOps.

rprouse:
It is fine to merge this to the 3.13 dev branch, but our default dev branch is actually master. We are working on the 4.0 release, but merging bug fixes and changes like this back to this 3.13 branch for any point releases. We haven't fully setup this branch properly for PR builds yet so I will need to Admin merge it and then we can cherry pick to master.

build.cake
304NotModified: ok clear, thanks :)

jnm2 (approved):
Thanks for the help!

mikkelbu (approved):

304NotModified:
FYI|

looks great!

image

PR #3819, opened 2021-04-16
Add GitHub Actions to build on Windows, Linux and Mac

### LINKED ISSUE #3812: Add GitHub Actions
I'd like to see us using GitHub Actions to build this project so that everything is in one place. Once we do that we can make a decision on which pipelines we will maintain going forward and which will be our primary pipeline.

@OsirisTerje linked this project that runs GitHub Actions locally for testing, https://github.com/nektos/act

rprouse:
Fixes #3812

Currently just builds on all three platforms and saves the packages from Windows. It does not publish the packages to any NuGet package feed. In the future, we may do that.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -1,10 +1,6 @@
 name: Continuous Integration
 
-on:
-  push:
-    branches: [ master, release, v3.13-dev ]
-  pull_request:
-    branches: [ master, release, v3.13-dev ]
+on: [push, pull_request]
 
 jobs:
   build-windows:
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -97,7 +97,7 @@ jobs:
     - name: Upload test results
       uses: actions/upload-artifact@v2
       with:
-        name: Test results (Linux)
+        name: Test results (macOS)
         path: test-results
       # Use always() to always run this step to publish test results when there are test failures
       if: ${{ always() }}
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -1,0 +1,107 @@
+name: Continuous Integration
+
+on:
+  push:
+    branches: [ master, release, v3.13-dev ]
+  pull_request:
+    branches: [ master, release, v3.13-dev ]
+
+jobs:
+  build-windows:
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Build and Test
+      run: .\build.ps1 --target=Test --test-run-name=Windows
+
+    - name: Package
+      run: .\build.ps1 --target=Package
+
+    - name: Upload build artifacts
+      uses: actions/upload-artifact@v2
+      with:
+        name: Package
+        path: package
+
+    - name: Upload test results
+      uses: actions/upload-artifact@v2
+      with:
+        name: Test results (Windows)
+        path: test-results
+      # Use always() to always run this step to publish test results when there are test failures
+      if: ${{ always() }}
+
+  build-linux:
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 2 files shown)


rprouse:
This is looking good now. Marking ready for review and will make Required once it is merged.

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,113 @@
...
+
+jobs:
+  build-windows:
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Build and Test
+      run: .\build.ps1 --target=Test --test-run-name=Windows
```
manfred-brands, line 23: Any reason why you not use 'dotnet cake' to match the other two OS?

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,113 @@
...
+  push:
+    branches:
+      - master
+      - release
+      - 'v3.13-dev'
+  pull_request:
+    branches:
+      - master
+      - release
+      - 'v3.13-dev'
```
jnm2, line 13: Is there any reason to block CI from running if there was ever a pull request to another branch?
There might be other better-looking ways to do it, but `pull_request:` by itself with nothing following it enables the pull request trigger without putting filters on it. Maybe `branches: [ '*' ]` underneath if that looks too weird.

jnm2 (approved):

jnm2:
One thing I'd like to do eventually is add binlog artifacts. That makes it easy to drop into https://msbuildlog.com if you have it installed to diagnose weird CI issues if they come up. It captures all environment variables though, so it doesn't work if there are decrypted secrets in there.

Here's what it looks like to create binlogs (three of them in this case): https://github.com/shouldly/shouldly/blob/master/build.ps1#L15-L29

And upload them: https://github.com/shouldly/shouldly/blob/037a2b9c118aa266a84988d9adf0301c4ca1ad5a/.github/workflows/CI.yml#L29-L34

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,113 @@
...
+        name: Package
+        path: package
+
+    - name: Upload test results
+      uses: actions/upload-artifact@v2
+      with:
+        name: Test results (Windows)
+        path: test-results
+      # Use always() to always run this step to publish test results when there are test failures
+      if: ${{ always() }}
```
jnm2, line 40: `if: always()` has been working for me.

Looks like the `${{` escape is only needed if you aren't already implicitly in an expression context or if you have operators: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#about-contexts-and-expressions

rprouse:
@manfred-brands thanks for the review. I apologize, I mis-tagged you, but I appreciate your help 👍🏻

jnm2:
Just thought of one more thing to recommend:

```yml
env:
  DOTNET_NOLOGO: true
```

.github/workflows/continuous_integration.yml
rprouse: I took this directly from the GitHub docs, but good to know.

.github/workflows/continuous_integration.yml
rprouse: You are probably right. I might change this in my next PR.

rprouse:
@manfred-brands thanks for the review. I apologize, I mis-tagged you

.github/workflows/continuous_integration.yml
jnm2: I wonder if `pull_request` with no `:` would have worked.

jnm2: > if there was ever a pull request to another branch?

It sounds unusual, but I've contributed directly to the topic branches of another person's PRs on GitHub. PR inception. I've also contributed to a compiler feature branch which is not something NUnit currently does.

Issue #3825, opened 2021-04-18
CollectionAssert.AreEquivalent is extremely slow in NUnit version 3.13.1


NightElfik:
I have recently updated NUnit to version 3.13.1 and some of our tests started to be extremely slow (~1000x). I have identified the culprit as `CollectionAssert.AreEquivalent`. The test case below takes 25 seconds to finish.

```csharp
[Test]
public void NUunitRepro() {
	var dict = new Dictionary<int, string>();
	for (int i = 0; i < 10000; ++i) {
		dict.Add(i, i.ToString());
	}
	CollectionAssert.AreEquivalent(dict.Values.AsEnumerable(),
		Enumerable.Range(0, 10000).Select(j => j.ToString()));
}
```

In comparison, when using `HashSet` for equivalence assertion, the test time is  ~10 ms:

```csharp

[Test]
public void NUunitRepro_UsingSet() {
	var dict = new Dictionary<int, string>();
	for (int i = 0; i < 10000; ++i) {
		dict.Add(i, i.ToString());
	}

	var actual = new HashSet<KeyValuePair<int, string>>(dict);
	NUnit.Framework.Assert.IsTrue(actual.SetEquals(Enumerable.Range(0, 10000)
		.Select(j => new KeyValuePair<int, string>(j, j.ToString()))));
}
```

rprouse:
Adding the confirm label but I think this is likely caused by changes in that area of the code. I just want to walk through it in the debugger.

stevenaw:
Thanks @NightElfik You'd mentioned a recent version update. Can I ask which version you were using previously (was it < 3.10)?
I'm wondering if this could be a duplicate of https://github.com/nunit/nunit/issues/2799

rprouse:
Testing the `NUnitRepro` test in .NET 5.0 with various versions of NUnit;

- 3.9 => 27ms
- 3.10.1 => 6.7 sec
- 3.11 => 6.8 sec
- 3.12 => 6.9 sec
- 3.13 => 17.8 sec
- 3.13.1 => 17.8 sec

So, this does look like a duplicate of #2799 but it was made even worse in the 3.13 release.

I'm going to close this issue as a duplicate and add this info to #2799

PR #3831, opened 2021-04-21
Performance improvements comparing equivalence of large collections

### LINKED ISSUE #2598: Is.Not.EquivalentTo is extremely slow

### LINKED ISSUE #2799: CollectionAssert.AreEquivalent is extremely slow

rprouse:
Fixes #2799
See #2598 and #3825

When comparing equivalence of large collections, we found that removing from the end of a collection is faster to fix #2598, but this makes comparing two collections in the same order O(n^2).

This PR maintains that fix, but sorts the collections if possible to keep the time as close to O(n) as possible.

For this to work, the types in the collection must implement `IComparable` and the collection must be strongly typed so that we know that every item in the collection is the same type.

I do not use `IComparable<T>` and sort using `IList<T>.Sort()` because it is difficult to get an `IEnumerable` into an `IList<T>`. I took the easy route and use `ArrayList.Sort()`.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionTally.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionTally.cs
@@ -2,6 +2,7 @@
 
 using System.Collections;
 using System.Collections.Generic;
+using NUnit.Framework.Internal.Extensions;
 
 namespace NUnit.Framework.Constraints
 {
@@ -28,18 +29,33 @@ namespace NUnit.Framework.Constraints
 
         private readonly NUnitEqualityComparer comparer;
 
+        private bool _sorted = false;
+
         /// <summary>The result of the comparison between the two collections.</summary>
         public CollectionTallyResult Result
         {
             get
             {
-                return new CollectionTallyResult(
-                    new List<object>(_missingItems),
-                    new List<object>(_extraItems));
+                var missingItems = new List<object>(_missingItems.Count);
+                foreach (var o in _missingItems)
+                    missingItems.Add(o);
+
+                List<object> extraItems = new List<object>();
+                if(_sorted)
+                {
+                    for (int index = _extraItems.Count - 1; index >= 0; index--)
+                        extraItems.Add(_extraItems[index]);
+                }
+                else
+                {
+                    extraItems.AddRange(_extraItems);
+                }
+
+                return new CollectionTallyResult(missingItems, extraItems);
             }
         }
 
-        private readonly List<object> _missingItems = new List<object>();
+        private readonly ArrayList _missingItems = new ArrayList();
 
         private readonly List<object> _extraItems = new List<object>();
 
@@ -52,6 +68,9 @@ namespace NUnit.Framework.Constraints
 
             foreach (object o in c)
                 _missingItems.Add(o);
+
+            if (c.IsSortable())
+                _missingItems.Sort();
         }
 
         private bool ItemsEqual(object expected, object actual)
@@ -80,8 +99,25 @@ namespace NUnit.Framework.Constraints
         /// <param name="c">The objects to remove.</param>
         public void TryRemove(IEnumerable c)
         {
-            foreach (object o in c)
-                TryRemove(o);
+            if (c.IsSortable())
+            {
+                var remove = new ArrayList();
+                foreach (object o in c)
+                    remove.Add(o);
+
+                remove.Sort();
+                _sorted = true;
+
+                // Reverse so that we match removing from the end,
+                // see issue #2598 - Is.Not.EquivalentTo is extremely slow
+                for (int index = remove.Count - 1; index >= 0; index--)
+                    TryRemove(remove[index]);
+            }
+            else
+            {
+                foreach (object o in c)
+                    TryRemove(o);
+            }
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/CollectionTally.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionTally.cs
@@ -40,7 +40,7 @@ namespace NUnit.Framework.Constraints
                 foreach (var o in _missingItems)
                     missingItems.Add(o);
 
-                List<object> extraItems = new List<object>();
+                List<object> extraItems = new List<object>(_extraItems.Count);
                 if (_sorted)
                 {
                     for (int index = _extraItems.Count - 1; index >= 0; index--)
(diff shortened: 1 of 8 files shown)


src/NUnitFramework/framework/Constraints/Numerics.cs
```diff
@@ -39,6 +39,7 @@ public static bool IsFloatingPointNumeric(Object obj)
             {
                 if (obj is System.Double) return true;
                 if (obj is System.Single) return true;
+                if (obj is System.Decimal) return true;
```
rprouse, line 42: Unrelated. I was using `IsNumeric` originally and noticed that we missed this.

src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -28,18 +29,33 @@ public CollectionTallyResult(List<object> missingItems, List<object> extraItems)
...
-                    new List<object>(_missingItems),
-                    new List<object>(_extraItems));
+                var missingItems = new List<object>(_missingItems.Count);
+                foreach (var o in _missingItems)
+                    missingItems.Add(o);
+
+                List<object> extraItems = new List<object>();
+                if(_sorted)
+                {
+                    for (int index = _extraItems.Count - 1; index >= 0; index--)
```
rprouse, line 46: All of the extra backwards iterations of collections is extra complexity but the only way I could not revert the performance gains @mikkelbu made fixing #2598

src/NUnitFramework/framework/Internal/Extensions/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace NUnit.Framework.Internal.Extensions
+{
+    internal static class IEnumerableExtensions
+    {
+        public static bool IsSortable(this IEnumerable collection)
+        {
+            if (collection is null)
+                return false;
+
+            if (collection is StringCollection)
```
rprouse, line 16: I think this is the only non-generic built in collection that is strongly typed?

src/NUnitFramework/tests/Constraints/CollectionEqualsTests.cs
```diff
@@ -41,6 +41,15 @@ public void CanMatchAnArrayWithACollection()
             Assert.That(array, Is.EqualTo(collection));
         }
 
+        [Test]
+        public void CollectionsInDifferentOrderArNotEqual()
+        {
+            IList expected = new List<int> { 1, 2, 3 };
+            IList actual = new List<int> { 3, 2, 1 };
+
+            Assert.That(actual, Is.Not.EqualTo(expected));
```
rprouse, line 50: Just checking that sorting did not change behaviour.

src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -28,18 +29,33 @@ public CollectionTallyResult(List<object> missingItems, List<object> extraItems)
...
+                {
+                    extraItems.AddRange(_extraItems);
+                }
+
+                return new CollectionTallyResult(missingItems, extraItems);
             }
         }
 
-        private readonly List<object> _missingItems = new List<object>();
+        private readonly ArrayList _missingItems = new ArrayList();
```
mikkelbu, line 58: Why this change

src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -28,18 +29,33 @@ public CollectionTallyResult(List<object> missingItems, List<object> extraItems)
...
         public CollectionTallyResult Result
         {
             get
             {
-                return new CollectionTallyResult(
-                    new List<object>(_missingItems),
-                    new List<object>(_extraItems));
+                var missingItems = new List<object>(_missingItems.Count);
+                foreach (var o in _missingItems)
+                    missingItems.Add(o);
```
mikkelbu, line 41: Couldn't this be `missingItems.AddRange(_missingItems);` ?

src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -28,18 +29,33 @@ public CollectionTallyResult(List<object> missingItems, List<object> extraItems)
...
+
+                List<object> extraItems = new List<object>();
```
mikkelbu, line 43:
```suggestion
                List<object> extraItems = new List<object>(_extraItems.Count);
```

mikkelbu (approved):
LGTM and definitely an improvement

src/NUnitFramework/framework/Constraints/CollectionTally.cs
mikkelbu: I should have read the description before reviewing

> I do not use IComparable<T> and sort using IList<T>.Sort() because it is difficult to get an IEnumerable into an IList<T>. I took the easy route and use ArrayList.Sort().

src/NUnitFramework/framework/Constraints/CollectionTally.cs
mikkelbu: Probably not, since `_missingItems` is an `ArrayList`

Issue #3843, opened 2021-04-29
IDisposable & InstancePerTestCase: Object created for OneTimeSetUp is not disposed


mvdfugro:
Hi folks,

While trying to better understand the fixture lifecycle, I noticed that for a test fixture with two tests, the constructor is called three times, but the disposer is only called twice. I used the following code:

```C#
    [TestFixture, FixtureLifeCycle(LifeCycle.InstancePerTestCase)]
    public class SingleUseLifecycle : IDisposable
    {
        private static void Log(object obj, string value)
        {
            File.AppendAllText("c:\\temp\\SingleUseLifecycle.log", $"{DateTime.Now}\t{obj?.GetHashCode()}\t{value}\n");
        }
        
        public SingleUseLifecycle() => Log(this, "Constructor");
        public void Dispose() => Log(this, "Disposer");

        [OneTimeSetUp] public static void OneTimeSetUp() => Log(null, "OneTimeSetUp");
        [OneTimeTearDown] public static void OneTimeTearDown() => Log(null, "OneTimeTearDown");
        
        [SetUp] public void SetUp() => Log(this, "SetUp");
        [TearDown] public void TearDown() => Log(this, "TearDown");
        
        [Test] public void Test1() => Log(this, "Test1");
        [Test] public void Test2() => Log(this, "Test2");
    }
```

Which results in the following log:
```
29-4-2021 21:29:18	51781231	Constructor
29-4-2021 21:29:18		OneTimeSetUp
29-4-2021 21:29:18	463695	Constructor
29-4-2021 21:29:18	463695	SetUp
29-4-2021 21:29:18	463695	Test1
29-4-2021 21:29:18	463695	TearDown
29-4-2021 21:29:18	463695	Disposer
29-4-2021 21:29:18	26430654	Constructor
29-4-2021 21:29:18	26430654	SetUp
29-4-2021 21:29:18	26430654	Test2
29-4-2021 21:29:18	26430654	TearDown
29-4-2021 21:29:18	26430654	Disposer
29-4-2021 21:29:18		OneTimeTearDown
```

Objects 463695 and 26430654 are correctly constructed and disposed, but object 51781231 is not.

Object 51781231 gets constructed to run the OneTimeSetUp; a breakpoint in the constructor of the test fixture shows the following stack trace:
```
   at NUnitLifecycle.SingleUseLifecycle..ctor() in C:\Development\NUnitLifecycle\Class1.cs:line 36
[...]

PR #3844, opened 2021-04-30
Do not construct object for OneTimeSetUp in InstancePerTestCase lifecycle

mvdfugro:
Test + (hopefully to-the-point) solution for #3843 . May need some extra unit tests?

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
@@ -199,7 +199,7 @@ namespace NUnit.Framework.Internal.Execution
                     command = new OneTimeSetUpCommand(command, item);
 
                 // Construct the fixture if necessary
-                if (!Test.TypeInfo.IsStaticClass)
+                if (!Test.TypeInfo.IsStaticClass && !Test.HasLifeCycle(LifeCycle.InstancePerTestCase))
                     command = new ConstructFixtureCommand(command);
             }
 
--- a/src/NUnitFramework/testdata/AssemblyLevelLifeCycleFixture.cs
+++ b/src/NUnitFramework/testdata/AssemblyLevelLifeCycleFixture.cs
@@ -1,7 +1,7 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 namespace NUnit.TestData.LifeCycleTests
-{
+{ 
     public static class AssemblyLevelFixtureLifeCycleTest
     {
         public const string Code = @"
@@ -10,7 +10,7 @@ namespace NUnit.TestData.LifeCycleTests
             [assembly: FixtureLifeCycle(LifeCycle.InstancePerTestCase)]
 
             [TestFixture]
-            public class FixtureUnderTest
+            public class FixtureUnderTest : NUnit.TestData.LifeCycleTests.BaseLifeCycle
             {
                 private int _value;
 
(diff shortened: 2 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=3844) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=3844) before we can accept your contribution.<br/><br/>:x: mvdfugro [sign now](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=3844)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/nunit/nunit?pullRequest=3844) it.</sub>

mvdfugro:
Thanks for the quick review! I'm currently a bit swamped but hope to be able to write some additional tests in the next week or so.

mvdfugro:
In the above, the first three commits rearrange the tests to fully test the lifecycle in a single test (rather than testing bits and pieces in various tests).
The fourth adds a test to verify that inheritance works as expected.

The rest of the commits use inheritance to make sure the basic invariants are verified in every test. I'm not sure whether that's acceptable for test within NUnit or whether it is preferred to have duplication to have simpler tests -- I'm happy to just push the first four commits otherwise.

Adding the BaseLifeCycle did surface a bug (this is why the tests are failing -- I've kept a failing test to show the issues):

* The InstancePerTestCase annotation does not seem to be 'inherited' by nested classes. (The original test cases actually didn't test anything as they (confusingly) did not run -- `TestBuilder.MakeFixture(typeof(LifeCycleWithNestedOverridingFixture))` doesn't actually instantiate any nested test classes.)

I also found something I'm confused by but which might be intended behavior:

* The test with the TestFixtureSource annotation calls OneTimeSetUp _twice_. This happens for both InstancePerTestCase & SingleInstance so it is consistent -- I did find it somewhat surprising. Is this intended behavior?



The assembly level tests are a bit out of sync with the main tests (i.e. not every test in `NUnit.TestData.LifeCycleTests` has an friend in `NUnit.TestData.LifeCycleTests.AssemblyLevelFixtureLifeCycleTest`). I'm not entirely sure how useful it is to test all combinations of attributes there?
My suggestion would be to remove those tests (as they are nominally covered in the regular tests), and instead focus on tests that verify the behavior of overrides of `FixtureLifeCycle(LifeCycle.InstancePerTestCase)`. I just realized those are missing entirely so I'll add those in any case.

mvdfugro:
Pushed additional assembly-attribute-tests to verify the overriding behavior. Seems to work as expected! 👍

mvdfugro:
Dug into the wrapping class issue as well, but I'm not entirely sure how it's supposed to work:

- In `NUnitTestFixtureBuilder.BuildFrom`, `fixture.ApplyAttributesToTest(new AttributeProviderWrapper<FixtureLifeCycleAttribute>(typeInfo.Type.GetTypeInfo().Assembly));` is called. This correctly applies attributes on inherited classes but doesn't so do for wrapped classes. The AttributeProviderWrapper could recurse into parent wrappers as well.
  - In the end, this is what sets the `LifeCycle` flag on the fixture, so this seems like a natural choice.
-  `TestExtensions.HasLifeCycle` recurses through `Test.Parent`, but that is only set for _tests_, not for _fixtures_. I can't quite figure out how `TestExtensions.HasLifeCycle` is supposed to work -- it seems to search for the first use of 'InstancePerTestCase' rather than looking for the first defined lifecycle... but that would imply that you could never override from `InstancePerTestCase` back to `SingleInstance`.

mvdfugro:
Rebased onto latest master, and added `Ignore` attributes to the two tests that fail because of #3888. I'll move the comment above to that bug for separate discussion; this now just resolves the original original issue + reorganizes / simplifies the test case setup.

rprouse (approved):
That is an impressive set of tests, nicely done.

mvdfugro:
Hi @rprouse, is there anything I still need to do for this to get merged?

rprouse:
@mvdfugro I was hoping to get a second review, but I'm happy with the work and the tests you've added so I'll merge. Thanks for your contribution.

mvdfugro:
Thanks!

PR #3845, opened 2021-04-30
Changed summary comments on Warn.If overloads to correctly reflect…

### LINKED ISSUE #3811: Incorrect summary comments on Warn.If overloads
I believe a couple `Warn.If` overloads have incorrect summary comments.

https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Warn.cs#L335

https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Warn.cs#L335

The summary comments on those overloads read:

> Asserts that a condition is true. If the condition is false, a warning is issued.

However, the method bodies read:

```
Warn.If(condition, Is.True, ...);
```

They log a warning if the condition is true.

The summary comment seems to be correct on other overloads:

> Asserts that a condition is false. If the condition is true a warning is issued.

hyp-do:
Fixes #3811

 that it asserts false and warns if true

### DIFF
--- a/src/NUnitFramework/framework/Warn.cs
+++ b/src/NUnitFramework/framework/Warn.cs
@@ -327,7 +327,7 @@ namespace NUnit.Framework
         #region Boolean
 
         /// <summary>
-        /// Asserts that a condition is true. If the condition is false, a warning is issued.
+        /// Asserts that a condition is false. If the condition is true, a warning is issued.
         /// </summary>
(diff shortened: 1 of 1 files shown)


mikkelbu (approved):
Looks good to me. I've triggered the build and also changed the first description, so that it auto-closes #3811 when merged to master.

mikkelbu:
Merging this now. Thanks for the contribution @hyp-do 👍

PR #3857, opened 2021-05-17
Add support for Theories in nested test classes

### LINKED ISSUE #3856: Theories in nested Testfixtures
To add a better structure to our tests we use nested test fixtures.

```
public class OuterTestFixture{
    public class InnerTestFixture{
        [Test]
        public void ActualTest()
    }
}
```

We recently started using theories for our tests and ran into the issue that having nested classes makes the use of theories unnecessarily complicated. We have to duplicate data points and data point sources in every nested class to get the theories to recognize them. 

It would be great if theories could recognize data points from their declaring types. To avoid issues with backward compatibility this could be activated through a flag on the `Theory` attribute.

Crown0815:
fixes #3856

We already have a working solution for the issue described in #3856 which I adapted to the existing NUnit Theory attribute.
Please let me know what I can improve to make this a viable contribution.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TheoryAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TheoryAttribute.cs
@@ -36,9 +36,9 @@ namespace NUnit.Framework
         /// <summary>
         /// Construct the attribute, specifying a combining strategy and source of parameter data.
         /// </summary>
-        public TheoryAttribute() : base(
+        public TheoryAttribute(bool searchInDeclaringTypes = false) : base(
             new CombinatorialStrategy(),
-            new ParameterDataProvider(new DatapointProvider(), new ParameterDataSourceProvider()))
+            new ParameterDataProvider(new DatapointProvider(searchInDeclaringTypes), new ParameterDataSourceProvider()))
         {
         }
     }
--- a/src/NUnitFramework/framework/Internal/Builders/DatapointProvider.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/DatapointProvider.cs
@@ -20,9 +20,9 @@ namespace NUnit.Framework.Internal.Builders
         private readonly bool _searchInDeclaringTypes;
         
         /// <summary>
-        /// Creates new DatapointProvider
+        /// Creates a new DatapointProvider.
         /// </summary>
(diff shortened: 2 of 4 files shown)


mikkelbu (commented):
@nunit/framework-team I don't see a problem with adding this flag, as the addition is compatible with existing code. Do any of you have concerns?

@Crown0815 I think we should also add some data in `https://github.com/nunit/nunit/blob/master/src/NUnitFramework/testdata/TheoryFixture.cs` with a `Theory(true)` and a `Theory` attribute in the same class and then a test in src/NUnitFramework/tests/Attributes/TheoryTests.cs that tests that the boolean affects the search.

Crown0815:
@mikkelbu I made the requested changes. 

Apparently, some of the tests I added are failing, but I am not able to get them to run locally. Could you please explain what I have to do so I can run these tests locally. They are not discovered by VSTest.

mikkelbu:
> Apparently, some of the tests I added are failing, but I am not able to get them to run locally. Could you please explain what I have to do so I can run these tests locally. They are not discovered by VSTest.

Currently, there are some issues running the tests in VS directly. There is a possible solution here
https://github.com/nunit/nunit/discussions/3861#discussioncomment-867700. Alternatively, you can run `build --target=Test --configuration=Release` from powershell to run our cake scripts - see https://github.com/nunit/nunit/blob/master/BUILDING.md.

The tests are failing as reflection is trying to access the field `i0` on `NestedWhileSearchingInDeclaringType`. I guess we need to return a pair/tuple from `GetMembersFromType` - the `Type` and the `MemberInfo`, so we know which type has the `MemberInfo`. We should then use this type in all the places where we call `ProviderCache.GetInstanceOf`, so we use the right `Type`.

Crown0815:
@mikkelbu I could not get the local tests to run, but your proposed changes fixed failing tests. I considered a specialized class to encapsulate some of the operations executed on the member but then went with the proposed Tuple since it was a quick fix.

Please let me know what you think.

rprouse (approved):
Looks good to me. Do you want to see any other changes @mikkelbu ?

mikkelbu (approved):
@Crown0815 I'm sorry. I completely lost track of this during the summer. I'll merge the changes now.

PR #3881, opened 2021-06-20
Collection Superset/Subset/Equivalent works with Tuple+ValueTuple for all cases

### LINKED ISSUE #3841: Breaking change: Is.SupersetOf with ValueTuple requires IComparable in NUnit 3.13.2
The following test passed in 3.13.1 and earlier, but fails in 3.13.2.

This is a simplified repro, of course. ValueTuple equality is useful for simplifying more complex comparisons.

```
[Test]
public void Test()
{
    var a = new S { C = 'a' };
    var b = new S { C = 'b' };
    var c = new S { C = 'c' };
    var d = new S { C = 'd' };

    var actual = new[] { (a, 1), (b, 2), (c, 3), (d, 4) };
    var expected = new[] { (b, 2), (c, 3) };
    Assert.That(actual, Is.SupersetOf(expected));
}

class S
{
    public char C;
}
```

Here is the stack trace:
```
System.InvalidOperationException: Failed to compare two elements in the array.
 ---> System.ArgumentException: At least one object must implement IComparable.
   at System.Collections.Comparer.Compare(Object a, Object b)
   at System.Collections.Generic.ObjectComparer`1.Compare(T x, T y)
   at System.ValueTuple`2.CompareTo(ValueTuple`2 other)
   at System.ValueTuple`2.System.IComparable.CompareTo(Object other)
   at System.Collections.Comparer.Compare(Object a, Object b)
   at System.Array.SorterObjectArray.InsertionSort(Int32 lo, Int32 hi)
   at System.Array.SorterObjectArray.IntroSort(Int32 lo, Int32 hi, Int32 depthLimit)
   at System.Array.SorterObjectArray.IntrospectiveSort(Int32 left, Int32 length)
   at System.Array.SorterObjectArray.IntrospectiveSort(Int32 left, Int32 length)
   at System.Array.Sort(Array keys, Array items, Int32 index, Int32 length, IComparer comparer)
   at System.Collections.ArrayList.Sort(Int32 index, Int32 count, IComparer comparer)
   at System.Collections.ArrayList.Sort()
   at NUnit.Framework.Constraints.CollectionTally..ctor(NUnitEqualityComparer comparer, IEnumerable c)
   at NUnit.Framework.Constraints.CollectionItemsEqualConstraint.Tally(IEnumerable c)
   at NUnit.Framework.Constraints.CollectionSupersetConstraint.Matches(IEnumerable actual)
   at NUnit.Framework.Constraints.CollectionSupersetConstraint.ApplyTo[TActual](TActual actual)
```

Note that ValueTuple's CompareTo works as expected:
```
(a, 1).CompareTo((a, 1)) // 0
```
but trying to use this on the assertion does not help. This still fails, presumably because it's not using this delegate as the comparer for the sorting.
```
Assert.That(actual, Is.SupersetOf(expected).Using<(S, int)>((t1, t2) => t1.CompareTo(t2)));
```

Implementing `IComparable` on the inner type works, but having to implement this all over the place just to make an assertion in the unit tests work, when that assertion used to work fine until a patch change in NUnit is not very nice. The objects are already reference equal.
```
class S : IComparable
{
    public char C;
    public int CompareTo(object s) => C.CompareTo(((S)s).C);
}
```

Also note that `HashSet<T>.IsSupersetOf` still works as expected, so this works:
```
Assert.That(new HashSet<(S, int)>(actual).IsSupersetOf(new HashSet<(S, int)>(expected)));
```
but one of the reasons NUnit is so good is that `Assert.That(actual, Is.SupersetOf(expected))` generates more useful messages when the assertion fails than "expected true but was false".

Using anonymous types rather than ValueTuples also still works as expected.

stevenaw:
Fixes #3841 

CollectionTally could fail if a ValueTuple  is passed where one of the contained item's types doesn't implement IComparable. This PR provides an exception-based fallback for cases like this which can't easily be identified upfront.

This is branched from/to 3.13.x, but should likely later be cherry-picked onto the main development branch

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionTally.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionTally.cs
@@ -21,6 +21,7 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using NUnit.Framework.Internal.Extensions;
@@ -50,6 +51,7 @@ namespace NUnit.Framework.Constraints
 
         private readonly NUnitEqualityComparer comparer;
 
+        private readonly bool _isSortable;
         private bool _sorted = false;
 
         /// <summary>The result of the comparison between the two collections.</summary>
@@ -91,13 +93,28 @@ namespace NUnit.Framework.Constraints
                 _missingItems.Add(o);
 
             if (c.IsSortable())
-                _missingItems.Sort();
+            {
+                _isSortable = TrySort(_missingItems);
+            }
         }
 
         private bool ItemsEqual(object expected, object actual)
         {
             Tolerance tolerance = Tolerance.Default;
             return comparer.AreEqual(expected, actual, ref tolerance);
+        }
+
+        private static bool TrySort(ArrayList items)
+        {
+            try
+            {
+                items.Sort();
+                return true;
+            }
+            catch (InvalidOperationException e) when (e.InnerException is ArgumentException ae && ae.Message.Contains(nameof(IComparable)))
+            {
+                return false;
+            }
         }
 
         /// <summary>Try to remove an object from the tally.</summary>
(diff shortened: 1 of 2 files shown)


stevenaw:
It looks like the LargeStringCollection test timed out on macOS .NET 3.5. I'm unsure if this is just a slow VM or a slow test issue.

Dreamescaper:
I don't like this try-catch hack, but unfortunately I have no idea how we could resolve it otherwise :(

Dreamescaper (approved):

mikkelbu (approved):
I proposed this kind of solution, but I can clearly understand why @Dreamescaper think this feels like a try-catch hack, but alas I also cannot think of a better solution given the current context. So I'm going to approve and merge this. 

Thanks for the contribution @stevenaw 👍

stevenaw:
Thanks @mikkelbu !
PR to apply this to 4.x has been made here: #3895

Issue #3273, opened 2019-05-31
System.InvalidOperationException: Invalid attempt made to decrement the event's count below zero.


RalfKoban:
Hi,

I'm using NUnit 3.12 and NCrunch 3.28.0.6 on a .NET Framework 4.8 C# solution in VS 2017 (v15.9.12).
NCrunch is running in Auto-churn mode - that is, the tests get executed in an endless loop.

After some time (maybe half a minute or so) it happens that some of the tests fail with following exception:

``` C#
An exception was thrown in the Task Environment: System.InvalidOperationException: Invalid attempt made to decrement the event's count below zero.
   at System.Threading.CountdownEvent.Signal()
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.OnChildItemCompleted(Object sender, EventArgs e) in D:\a\1\s\src\NUnitFramework\framework\Internal\Execution\CompositeWorkItem.cs:line 374
   at System.EventHandler.Invoke(Object sender, EventArgs e)
   at NUnit.Framework.Internal.Execution.WorkItem.WorkItemComplete() in D:\a\1\s\src\NUnitFramework\framework\Internal\Execution\WorkItem.cs:line 378
   at NUnit.Framework.Internal.Execution.SimpleWorkItem.PerformWork() in D:\a\1\s\src\NUnitFramework\framework\Internal\Execution\SimpleWorkItem.cs:line 79
   at NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread() in D:\a\1\s\src\NUnitFramework\framework\Internal\Execution\WorkItem.cs:line 521
   at NUnit.Framework.Internal.Execution.CompositeWorkItem.RunChildren() in D:\a\1\s\src\NUnitFramework\framework\Internal\Execution\CompositeWorkItem.cs:line 296
   at NUnit.Framework.Internal.Execution.WorkItem.RunOnCurrentThread() in D:\a\1\s\src\NUnitFramework\framework\Internal\Execution\WorkItem.cs:line 521
   at NUnit.Framework.Internal.Execution.TestWorker.TestWorkerThreadProc() in D:\a\1\s\src\NUnitFramework\framework\Internal\Execution\TestWorker.cs:line 142
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[...]

PR #3883, opened 2021-06-22
Fixed a race condition that could cause NUnit to launch unsanctioned background test execution threads destabilising the test environment.

remcomulder:
In our repeated hammering of NCrunch/NUnit in our CI, we've picked up on an intermittent issue that has probably been in the NUnit code base for a while now.

When the WorkShift starts up worker threads in a first-start situation, the worker threads can end the shift before the _firstStart flag is set in the WorkShift.Start() method.  This causes the first-start code to repeat more than once in the WorkShift, spawning background threads that destabilise the test environment.

The simple solution is to make sure the flag is set to false prior to kicking off background threads.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
@@ -132,12 +132,13 @@ namespace NUnit.Framework.Internal.Execution
             IsActive = true;
 
             if (_firstStart)
+            {
+                _firstStart = false;
                 StartWorkers();
+            }
 
             foreach (var q in Queues)
                 q.Start();
-
-            _firstStart = false;
         }
 
         private void StartWorkers()


stevenaw:
Thanks @remcomulder !
Out of curiosity, does this issue resemble what you hit? https://github.com/nunit/nunit/issues/3273

Also, can you please confirm which version of the framework you saw this on?

remcomulder:
I would say there is definitely a chance that #3273 is caused by this issue.  I saw all kinds of crazy issues appear when that background thread kicked off.

I observed the issue appearing under .NET 4.7 and .NET Core 2.0.  In theory, it will appear on all platforms.  For the problem to show, there needs to be a context switch during the execution of WorkShift.Start(), which will generally only happen when the system is under load.  Our CI system almost never kicked this up, but I could reproduce it on demand when thrashing a suite of NUnit tests under load with a 138 core grid and NCrunch churn mode.

My guess is that there is a minority of NUnit users being driven crazy by this.  It was quite tough to track it down.

Dreamescaper:
It looks fine by me.
I'm going to merge it soon, unless there are any objections.

Dreamescaper (approved):

src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
```diff
@@ -132,12 +132,13 @@ public void Start()
             IsActive = true;
 
             if (_firstStart)
+            {
+                _firstStart = false;
```
jnm2, line 136: Is there any chance of a race between threads for this check? If so, we could do:

```cs
// _firstStart will need to be an int with 0 for false and 1 for true
if (Interlocked.Exchange(ref _firstStart, 0) == 1)
{
```

jnm2 (approved):
Thanks so much!

src/NUnitFramework/framework/Internal/Execution/WorkShift.cs
remcomulder: From what I can tell, as long as the flag gets set before the background thread is kicked off, there should be no threading risk here.

mikkelbu:
Is it on purpose that this has been merged to `master` and not to the `v3.13-dev` branch? I don't think that there is a timeline for releasing from `master`.

jnm2:
I think the pattern has been to backport to the v3 branch after merging to main: 3837

PR #3889, opened 2021-06-25
Apply attributes recursively for nested classes

### LINKED ISSUE #3888: FixtureLifeCycle not applied correctly to nested test classes
Found during work on #3844.

Reproduction testcase:
```C#
    [TestFixture]
    public class Nested
    {
        [TestFixture]
        public class NestedAgain
        {
            int x;

            [Test, Order(1)]
            public void Test1()
            {
                Assert.That(x++, Is.EqualTo(0));
            }

            [Test, Order(2)]
            public void Test2()
            {
                Assert.That(x++, Is.EqualTo(0));
            }
        }
    }
```

Expected result: Test passes
Actual result: 
```
_localtests.Nested+NestedAgain.Test2

  Expected: 0
  But was:  1
```

Verified on 3.13.1 and 'latest master' (50938798c1d9608a62149701a2c1f7705303a4b9)

mvdfugro:
This resolves #3888 by applying _all_ attributes recursively for nested classes.

Although this is a nice solution in terms of consistency, it also has a somewhat higher risk of breaking backwards compatibility (in all test cases in the world there is probably _someone_ depending on the non-inheritance of attributes). So maybe this is an OK solution for 4.0+ but not for backporting to 3.x. An interesting example of this is the `Ignore` attribute:

```C#
    [TestFixture]
    [Ignore("testing ignore a fixture")]
    public class FixtureUsingIgnoreAttribute
    {
        [Test]
        public void Success()
        {
        }

        public class SubFixture
        {
            [Test]
            public void Success()
            {
                // this is currently a test that runs, but gets ignored with this change
            }
        }
    }
```

A safer (but less consistent) solution would filter the recursive application for specific attributes (maybe based on an explicit interface?)

Note that branch is 'on top of' #3843 as it depends (somewhat) on the new test system added there. The framework changes should apply to master without any changes.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/SetUpFixtureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/SetUpFixtureAttribute.cs
@@ -38,7 +38,7 @@ namespace NUnit.Framework
                     fixture.MakeInvalid(reason);
             }
 
-            fixture.ApplyAttributesToTest(typeInfo.Type.GetTypeInfo());
+            fixture.ApplyAttributesToTest(typeInfo.Type);
 
             return new TestSuite[] { fixture };
         }
--- a/src/NUnitFramework/framework/Attributes/TestFixtureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestFixtureAttribute.cs
@@ -247,7 +247,7 @@ namespace NUnit.Framework
         {
             var fixture = _builder.BuildFrom(typeInfo, filter, this);
             fixture.ApplyAttributesToTest(new AttributeProviderWrapper<FixtureLifeCycleAttribute>(typeInfo.Type.GetTypeInfo().Assembly));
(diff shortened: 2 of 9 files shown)


rprouse:
It looks like this has the changes from your other PR mixed up in it. I am going to hold off reviewing until that is merged.

mvdfugro:
Indeed this is based on the other PR (as otherwise I had no failing tests to fix ;-)). I'll rebase this on master once the other request is merged back there.

mvdfugro:
Rebased on latest master. Now the review only contains the three relevant commits :-)

Fundamentally, two changes have been made:
- Attributes are now applied recursively (and I've also added a test for the Ignore attribute),
- HasLifeCycle has been simplified to just use the first TestFixture found. I believe this is what the actual behavior already was (as there only is a single TestFixture in the ITest.Parent hierarchy) but this more clearly communicates that this is not where the inheritance happens. See the discussion on #3888

rprouse (approved):

PR #3895, opened 2021-07-11
Patch fix for #3841 (collection tally tuples) to 4.x branch

### LINKED ISSUE #3841: Breaking change: Is.SupersetOf with ValueTuple requires IComparable in NUnit 3.13.2

stevenaw:
Patches the changes from #3881 which fixes #3841 from 3.13.x onto 4.x

### DIFF
--- a/src/NUnitFramework/tests/Constraints/CollectionSupersetConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/CollectionSupersetConstraintTests.cs
@@ -69,6 +69,42 @@ namespace NUnit.Framework.Constraints
             }
         }
 
+#if !NET35
+        [Test]
+        public void WorksOnValueTuples()
+        {
+            var actual = new[] { ('a', 1), ('b', 2), ('c', 3), ('d', 4) };
+            var expected = new[] { ('b', 2), ('c', 3) };
+
+            var constraint = new CollectionSupersetConstraint(expected);
+            var constraintResult = constraint.ApplyTo(actual);
+
+            Assert.That(constraintResult.IsSuccess, Is.True);
+        }
+
+        [Test]
+        public void WorksOnValueTuples_OneTypeIsntIComparer()
+        {
+            var a = new S { C = 'a' };
+            var b = new S { C = 'b' };
+            var c = new S { C = 'c' };
+            var d = new S { C = 'd' };
+
+            var actual = new[] { (a, 1), (b, 2), (c, 3), (d, 4) };
+            var expected = new[] { (b, 2), (c, 3) };
+
+            var constraint = new CollectionSupersetConstraint(expected);
+            var constraintResult = constraint.ApplyTo(actual);
+
+            Assert.That(constraintResult.IsSuccess, Is.True);
+        }
+        
+        private class S
+        {
+            public char C;
+        }
+#endif
+
         public class IgnoreCaseDataProvider
         {
             public static IEnumerable TestCases
--- a/src/NUnitFramework/tests/Constraints/CollectionSupersetConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/CollectionSupersetConstraintTests.cs
@@ -71,6 +71,35 @@ namespace NUnit.Framework.Constraints
 
 #if !NET35
         [Test]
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/tests/Constraints/CollectionSupersetConstraintTests.cs
```diff
@@ -69,6 +69,71 @@ public void HonorsIgnoreCase(IEnumerable expected, IEnumerable actual)
             }
         }
 
+#if !NET35
```
rprouse, line 72: No more NET35 in this release, so not needed.

rprouse (approved):

src/NUnitFramework/tests/Constraints/CollectionSupersetConstraintTests.cs
stevenaw: Oops. Thanks for catching that!

PR #3920, opened 2021-08-21
Fix comparisons of high precision decimals

### LINKED ISSUE #3898: NUnit 3.13.2 : LessThanOrEqualTo fails on a case which should succeed
image

Here is a code example :
```
[Test, Repeat(1000)]
public void Test()
{
}
```

Best regards

stevenaw:
Fixes #3898 

NOTE: I'm unsure the history behind treating `decimal` separate from `IsFloatingPointNumeric`, but recently adding it to `IsFloatingPointNumeric` broke high-precision comparisons. I've removed it again to try and preserve previous behaviour.
I also noticed that `IsFixedPointNumeric` would return true when given decimal as a boxed value (the `object` overload) but `false` when given decimal in the `Type` overload. This PR changes this to now return `true` for both cases

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Numerics.cs
+++ b/src/NUnitFramework/framework/Constraints/Numerics.cs
@@ -60,7 +60,6 @@ namespace NUnit.Framework.Constraints
             {
                 if (obj is System.Double) return true;
                 if (obj is System.Single) return true;
-                if (obj is System.Decimal) return true;
             }
             return false;
         }
@@ -71,7 +70,6 @@ namespace NUnit.Framework.Constraints
             {
                 if (type == typeof(double)) return true;
                 if (type == typeof(float)) return true;
-                if (type == typeof(decimal)) return true;
             }
             return false;
         }
--- a/src/NUnitFramework/framework/Constraints/Numerics.cs
+++ b/src/NUnitFramework/framework/Constraints/Numerics.cs
@@ -105,6 +105,7 @@ namespace NUnit.Framework.Constraints
             {
                 if (type == typeof(byte)) return true;
                 if (type == typeof(sbyte)) return true;
+                if (type == typeof(decimal)) return true;
                 if (type == typeof(int)) return true;
                 if (type == typeof(uint)) return true;
                 if (type == typeof(long)) return true;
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/tests/Constraints/NumericsTest.cs
```diff
@@ -58,6 +58,58 @@ public void CanMatchDecimalWithoutToleranceMode()
...
+            });
+        }
+
+        [Test]
+        public void TreatFixedPointsSameForInstanceOrType()
+        {
+            var values = new object[] {
+                (byte)1,
+                (sbyte)1,
+                (decimal)1,
```
stevenaw, line 96: These tests were more about documenting the existing implementation than any particular behaviour. Perhaps it's overkill to include them?

src/NUnitFramework/framework/Constraints/Numerics.cs
```diff
@@ -107,6 +105,7 @@ internal static bool IsFixedPointNumeric(Type type)
             {
                 if (type == typeof(byte)) return true;
                 if (type == typeof(sbyte)) return true;
+                if (type == typeof(decimal)) return true;
```
jnm2, line 108: `Decimal` is not fixed point. It's a base 10 floating-point number. (Floating point means it contains an exponent part that specifies where the decimal point goes.) `Decimal` is also correctly described this way at e.g. https://docs.microsoft.com/en-us/dotnet/api/system.decimal.

The method names should change or the caller should do something else to differentiate between the two binary floating-point types and the decimal floating-point type.

jnm2 (changes requested):

src/NUnitFramework/framework/Constraints/Numerics.cs
stevenaw: Yeah. I'm also surprised it was not treated as an explicit floating type by these functions in nunit prior to 3.13.2

Instead of adding it here, should we remove it from the object overload of IsFixedPointNumeric in a separate PR (a breaking change).

jnm2: Oh yes, I think we should. It's a bugfix.

stevenaw: Hm, actually, as I think about this more I may be able to fix the precision issue while keeping `decimal` in the floating point check and without adding this line.
This should allow the final changeset for this fix to more easily be cherry picked into v4 branch

PR #3926, opened 2021-09-06
Add support for .NET5.0 OSPlatformAttribute

### LINKED ISSUE #3866: SupportedOSPlatform
Are there plans to filter tests on the .NET5.0 SupportedOSPlatform attribute?
At the moment, I have to put two attributes, one to shut up the compiler, the other for nunit to only run this for the matching Platform.

In addition the platform name for the .NET5.0 differs from the supported ones for .NUnit Platform

```
    [Platform("win")]
    [SupportedOSPlatform("windows")]
```

If acceptable, I can raise a PR for supporting SupportedOSPlatform on .NET5.0 only.

manfred-brands:
Fixes #3866 

Draft because somehow 'build -t test' doesn't find any tests on my machine.


src/NUnitFramework/framework/Attributes/OSPlatformTranslator.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Reflection;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+
+namespace NUnit.Framework
+{
+    /// <summary>
+    /// Marks an assembly, test fixture or test method as applying to a specific platform.
+    /// </summary>
+    public static class OSPlatformTranslator
```
Dreamescaper, line 16: Probably should be internal

src/NUnitFramework/framework/Attributes/OSPlatformTranslator.cs
```diff
@@ -0,0 +1,113 @@
...
+    {
+        /// <summary>
+        /// Converts a .NET 5+ SupportedOSPlatformAttribute into an NUnit PlatformAttribute
+        /// </summary>
+        /// <param name="possibleSupportedOSPlatformAttribute"></param>
+        /// <returns></returns>
+        public static object Translate(object possibleSupportedOSPlatformAttribute)
```
Dreamescaper, line 23: Personally I would prefer something like that:
```csharp
public static bool TryTranslate(object possibleSupportedOSPlatformAttribute, out PlatformAttribute platformAttribute);
```

That would make it more obvious what this method returns.

And then use it smth like that:
```csharp
OSPlatformTranslator.TryTranslate(attribute, out var platformAttribute) ? platformAttribute : attribute
```

(Your option is more concise though, so I don't insist to hard)

Dreamescaper (commented):
Generally looks fine by me. 
I've only left a minor nitpicks.

src/NUnitFramework/framework/Attributes/OSPlatformTranslator.cs
manfred-brands: Completely agree. Done.

src/NUnitFramework/framework/Attributes/OSPlatformTranslator.cs
manfred-brands: Because of #3941 I have to changes this, combining multiple `OSPlatformAttribute` into a single `PlatformAttribute`

manfred-brands:
@mikkelbu MacOS builds fail with "__The repository primary signature validity period has expired__".  Also fails in 'master'.

src/NUnitFramework/framework/Attributes/OSPlatformTranslator.cs
stevenaw: Thanks for explaining this, I was going to ask why the need to handle `PlatformAttribute` in the translator. Are you saying that this PR will also fix https://github.com/nunit/nunit/issues/3941 ?

manfred-brands: @stevenaw This PR will as a side effect combine multiple Platform attributes into one, thereby 'fixing' #3941. It depends if the owners want that or change the AllowMultiple flag.
The main reason I combine it here is for code that currently use both `SupportOSPlatform` and `Platform`. The first for the .NET compiler, the 2nd for nunit.

mikkelbu:
I also noticed the problems with builds which was also why I manually triggered https://nunit.visualstudio.com/NUnit/_build/results?buildId=4256&view=results, but I've not had time to look at it the last couple of week.

Regarding failures like this I proposed to run the builds every day in #3928, but so far I've not heard any feedback.

src/NUnitFramework/framework/Attributes/OSPlatformTranslator.cs
stevenaw: @manfred-brands Sorry it's taken me a few days to get back to you. That makes a lot of sense, thanks!

src/NUnitFramework/framework/Attributes/OSPlatformTranslator.cs
```diff
@@ -0,0 +1,141 @@
...
+        public static IEnumerable<object> Translate(IEnumerable<object> allAttributes)
+        {
+            var includes = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            var excludes = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+            foreach (var attribute in allAttributes)
+            {
+                if (attribute is PlatformAttribute platformAttribute)
+                {
+                    Add(includes, platformAttribute.Include);
```
stevenaw, line 30: `HashSet<T>` defines a `UnionWith` function to merge another `IEnumerable<T>` to itself. Would it work to use that here?

src/NUnitFramework/tests/Attributes/OSPlatformAttributeTests.cs
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+#if NET5_0
```
stevenaw, line 2: Instead of explicitly targeting NET5 by name, can we invert the check to be "NOT {UNSUPPORTED}". That way these tests will be included in other future build targets, NET6+.

For example something like (untested):
```csharp
#if !(NET45 | NETCOREAPP)
// net 5 code here
#endif
```

stevenaw (changes requested):
@manfred-brands Thanks for investigating + putting this PR up! I have one suggestion that may help with ensuring these tests continue to run in the future

src/NUnitFramework/tests/Attributes/OSPlatformAttributeTests.cs
manfred-brands: I would have liked to use `#if NET5_0_OR_GREATER` but it wasn't available in the SDK used by NUnit. Is there a reason we are not using the latest minor version 5.0.400?

stevenaw: That's a good question! I'm not sure

mikkelbu: Usually, we have only bumped these when necessary to make it possible for people to contribute even if they are not using the latest SDKs. And often we only updated if some CI broke.

But I cannot see a large problem with bumping to 5.0.400 (except that I need to update my VS installation). @rprouse do you have any oppinon on this?

rprouse: I am fine to bump the minor version, we should be running the latest bug and security fixes.

src/NUnitFramework/framework/Attributes/OSPlatformTranslator.cs
manfred-brands: Replaced own loop with call to `UnionWith`.

src/NUnitFramework/tests/Attributes/OSPlatformAttributeTests.cs
manfred-brands: Actually the bump by @mikkelbu to 5.0.202 added NET5_0_OR_GREATER already. Although I couldn't find it in any release notes.
Rebased the branch and updated the code to use '#if NET5_0_OR_GREATER' to fix @stevenaw concern.

src/NUnitFramework/framework/Attributes/OSPlatformTranslator.cs
mikkelbu: I think it is fix to fix #3941 and keep the AllowMultiple as it is.

src/NUnitFramework/tests/Attributes/OSPlatformAttributeTests.cs
stevenaw: Thanks @manfred-brands , that looks great! 👍

stevenaw (approved):

stevenaw:
Thanks for your contribution @manfred-brands ! We'll need a second approval before the PR can be merged.

PR #3929, opened 2021-09-11
Fix high precision decimal calculations in v3.13 (#3898)

### LINKED ISSUE #3898: NUnit 3.13.2 : LessThanOrEqualTo fails on a case which should succeed

stevenaw:
Fixes #3898 
Fixes high-precision decimal calculations by internally using `decimal` when appropriate.

I tried a few approaches here, and the one I found worked best while maintaining full backwards compatibility is to compare `expected` and `actual` as decimals when at least one of the values is decimal and BOTH are within the valid values for decimals (between `decimal.MinValue` and `decimal.MaxValue`).

I also tried to treat all floating point operations as `decimal` internally when expected and actual are within valid range, however this caused some a change in failure messages when comparing two `double` or `float` values (or a double to an int, etc) with a tolerance since the number of decimal places in the "+/-" portion of the error message changed. This approach broke some existing tests in nunit, and likely would've broken others as well.

NOTE: After merging, this should also be ported to the main development branch too

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Numerics.cs
+++ b/src/NUnitFramework/framework/Constraints/Numerics.cs
@@ -402,18 +402,11 @@ namespace NUnit.Framework.Constraints
             if (!IsNumericType(expected) || !IsNumericType(actual))
                 throw new ArgumentException("Both arguments must be numeric");
 
-            if (IsFloatingPointNumeric(expected) || IsFloatingPointNumeric(actual))
-            {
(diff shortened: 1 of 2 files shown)


stevenaw:
Oh that's a good point @realfaisalk thanks for pointing that out! That could probably be made clearer. Would you be willing to file a GitHub issue about it?

PR #3931, opened 2021-09-15
Fix high precision decimal calculations in v4 (#3898)

### LINKED ISSUE #3898: NUnit 3.13.2 : LessThanOrEqualTo fails on a case which should succeed

stevenaw:
Fixes #3898 
Backport of #3929

Allows high-precision decimal comparisons to be treated internally as `decimal`


mikkelbu (approved):
Looks exactly the same as were merged to the v3.13-dev branch, so it looks good 👍 .

stevenaw:
Thanks @mikkelbu !

PR #3950, opened 2021-10-01
fix: Correct builds on CI

mikkelbu:
According to NuGet/Announcements#56 .NET SDK 5.0.202 will have NuGet package verification disabled on Linux and macOS.


rprouse (approved):
Thanks for researching this!

Issue #3899, opened 2021-07-27
Allow randomizing 'Guid' test arguments with [Random]


julealgon:
`[Random]` currently supports several primitive types (mostly numeric ones), but it does not work with `Guid`. To provide different values to a `Guid` argument, one has to manually provide them either via `[TestCaseSource]` or via `[ValueSource]`.

In my experience, being able to easily provide different `Guid` values to a test is fairly useful. Currently, I'm doing this by using a `ValueSource`, which caused a [different problem](https://github.com/nunit/nunit/issues/3894), forcing me to resort to hardcoded `Guid`s instead of using `Guid.NewGuid`.

I'm aware that `Guid` randomization is different than the standard one using the `Random` class, so I'm not sure if it would be possible to have the same "seed" in discovery and execution phases to build the unique test names with the built-in values (probably not).

julealgon:
The last part of my post is probably not true as per [this comment from @jnm2](https://github.com/nunit/nunit/issues/3894#issuecomment-881877904) :
> In the meantime, that suggests the workaround you could use. Instead of Guid.NewGuid(), use `TestContext.CurrentContext.Random.NextGuid()`.

I guess you could then easily leverage this for generating the random `Guid` values in a consistent manner like any other random values are created today.

mikkelbu:
Closed by #3951

PR #3951, opened 2021-10-02
Allow randomizing 'Guid' test arguments with [Random]

Greybird:
Hi,

This PR should solve #3899.

I am not particularly satisfied about 
1. having to throw `NotSupportedException`s, but I chose to stay the closest possible to the current design. Very open to suggestion on a better design (interface for range support ?)
2. not having a test for checking that using `min` and/or `max` with `Guid` parameters would actually be refused through the `Guard`, but I did not find a test for the other one checking about having enough different values in the range when using `Distinct`. If I missed something, again, would be very happy to enhance the proposal.

If this PR is accepted, the [documentation for random attribute](https://docs.nunit.org/articles/nunit/writing-tests/attributes/random.html) will need to be updated for , though a PR on [nunit/docs, random.md](https://github.com/nunit/docs/blob/master/docs/articles/nunit/writing-tests/attributes/random.md)

For the sake of completeness, and even if this is outside the responsibility of nunit, I'm also getting a warning in Visual Studio, coming from a [Jetbrains rule](https://www.jetbrains.com/help/rider/NUnit.ParameterTypeIsNotCompatibleWithAttribute.html), which would require an update.
image

### DIFF
--- a/src/NUnitFramework/framework/Attributes/RandomAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RandomAttribute.cs
@@ -166,6 +166,8 @@ namespace NUnit.Framework
                     _source = new SByteDataSource(_count);
                 else if (parmType == typeof(decimal))
                     _source = new DecimalDataSource(_count);
+                else if (parmType == typeof(Guid))
+                    _source = new GuidDataSource(_count);
                 else if (parmType.GetTypeInfo().IsEnum)
                     _source = new EnumDataSource(_count);
                 else // Default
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=3951) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=3951) before we can accept your contribution.<br/><br/>:x: Greybird [sign now](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=3951)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/nunit/nunit?pullRequest=3951) it.</sub>

PR #3952, opened 2021-10-02
fix: Remove install script and entry classes

### LINKED ISSUE #1428: NUnitLite package always installs both Program.cs and Program.vb
The NUnitLite package has an install script which looks like it should be delete either of these files if they do not match the project type. It doesn't seem to be working. Haven't really looked in to why, yet.

mikkelbu:
Fixes #1428 by removing the install script and also the entry points, as they have not been applied for quite some time.

I'll update the documentation when this is merged.


stevenaw (approved):
Thanks @mikkelbu 👍

Issue #3410, opened 2019-10-31
Consider deprecating NUnitEqualityComparer.AreEqual optional bool parameter


stevenaw:
The NUnitEqualityComparer contains an optional parameter in the public API to specify if the comparison is top-level or not.
https://github.com/nunit/nunit/blob/8de9af1d695d42d2d85604a100ff9f6c00e5728f/src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs#L157

This has been converted to separate overloads in a subsequent PR. https://github.com/nunit/nunit/pull/3406

That PR also gets the code to the point where the boolean parameter _could_ be calculated (but doing so could break consumers who explicitly pass false).

We can't just remove it outright, since it's a part of the public API and could be used by consumers (ex: a custom wrapper over the comparer), but I think we should consider marking the overload with the bool as obsolete, to be removed in a later version.

mikkelbu:
I think it is fine to mark it obsolete in the 3.13.x branch and then correct it on master.

seanpoulter:
It appears to me that #3410 can now we be closed. What do you think? We've had two PRs merged in October that marked AsEqual as obsolete then removed it.

mikkelbu:
@seanpoulter Thanks for the heads up. You are right. This should be closed by the two PRs.

PR #3959, opened 2021-10-11
Marked 'NUnitEqualityComparer.AreEqual(object, object, Tolerance, bool)' as obsolete

x789:
This is PR 1/2 that solves issue #3410.

Changes:
* Marked the method `NUnitEqualityComparer.AreEqual(object, object, Tolerance, bool)` as obsolete.
* Changed the only (internal) caller of that method `AreEqual(object, object, Tolerance)`, so there are no internal references to the deprecated method.


mikkelbu:
@x789 As far as I can tell we are starting to use version 1.3.0 of the cake.tool. Prior we used version 1.2.0 on this branch. Version 1.3.0 were uploaded to nuget for approx 7 days ago. 

The builds are working on master, but there we also have #3835. @rprouse can you see any problem in backporting #3835 to the `v3.13-dev` branch? (and perhaps change the version 1.1.0 to 1.3.0)

PR #3962, opened 2021-10-14
Ensure that AfterTest always runs in AfterTestCommand

### LINKED ISSUE #3953: Dispose is not called on test fixtures with LifeCycle.InstancePerTestCase without TearDown method
# Issue

If I run the following test fixture, the `test_that_dispose_was_called` passes because the  `Dispose` method is called as expected.

```csharp
public class Test_fixture : IDisposable
{
    private static bool _disposeWasCalled;

    [Test, Order(0)]
    public void test_running_first() => Assert.Pass("Some test running before");

    [Test, Order(1)]
    public void test_that_dispose_was_called() => Assert.That(_disposeWasCalled);

    [TearDown]
    public static void TearDown(){}

    public void Dispose() => _disposeWasCalled = true;
}
```
image

However, if I remove the `TearDown` method, the `Dispose` method is no longer called and `_wasDisposed` remains false.

```csharp
public class Test_fixture : IDisposable
{
    private static bool _disposeWasCalled;

    [Test, Order(0)]
    public void test_running_first() => Assert.Pass("Some test running before");

    [Test, Order(1)]
    public void test_that_dispose_was_called() => Assert.That(_disposeWasCalled);

    public void Dispose() => _disposeWasCalled = true;
}
```
image

# Expected

I would have expected `Dispose` to be called after every test if the fixture has the attribute `FixtureLifeCycle(LifeCycle.InstancePerTestCase)` set on it, independent of the presence of a `TearDown` method. This is expectation is based on [the documentation of the FixtureLifeCycleAttribute](https://docs.nunit.org/articles/nunit/writing-tests/attributes/fixturelifecycle.html), which states that `IDisposable` fixtures are disposed after each test.

manfred-brands:
Fixes #3953

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/AfterTestCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/AfterTestCommand.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2017 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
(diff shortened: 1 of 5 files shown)


manfred-brands:
Created #3963 for master.

PR #3978, opened 2021-11-07
Prevent partial updates to collectiontally's _missingItems

### LINKED ISSUE #3976: Equivalency fallback for non-IComparable types can leave CollectionTally in corrupt state
When checking 2 arrays for equivalency it should not be required for the object to be comparable, equatable should be enough.
Take a look at following tests that demonstrate the issue. The last test should not have failed.

```
        [Test]
        public void ObjectArrayEquivalency()
        {
            var message3 = new object();
            var message4 = new object();
            var array = new[]
            {
                message3,
                message4
            };

            //Succeeds
            Assert.That(array, Is.EquivalentTo(new[]
            {
                message4,
                message3
            }));
        }

        [Test]
        public void TuplesWithComparableArrayEquivalency()
        {
            var message3 = DateTime.Now;
            var message4 = DateTime.Now.AddDays(1);
            var array = new[]
            {
                new Tuple<int, object, CancellationToken>(1, message3, CancellationToken.None),
                new Tuple<int, object, CancellationToken>(1, message4, CancellationToken.None)
            };

            //Succeeds
            Assert.That(array, Is.EquivalentTo(new[]
            {
                new Tuple<int, object, CancellationToken>(1, message4, CancellationToken.None),
                new Tuple<int, object, CancellationToken>(1, message3, CancellationToken.None)
            }));
        }



        [Test]
        public void TuplesWithObjectArrayEquivalency()
        {
            var message3 = new object();
            var message4 = new object();
            var array = new[]
            {
                new Tuple<int, object, CancellationToken>(1, message3, CancellationToken.None),
                new Tuple<int, object, CancellationToken>(1, message4, CancellationToken.None)
            };

            //Fails on System.Array.Sort: Failed to compare two elements in the array. At least one object must implement IComparable.
            Assert.That(array, Is.EquivalentTo(new[]
            {
                new Tuple<int, object, CancellationToken>(1, message4, CancellationToken.None),
                new Tuple<int, object, CancellationToken>(1, message3, CancellationToken.None)
            }));
        }
```

I'm using the following nuget package versions:
```
<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
<PackageReference Include="NUnit" Version="3.13.2" />
<PackageReference Include="NUnit3TestAdapter" Version="4.0.0" />
```

When bumping the versions a bit lower to earlier versions (where I knew my test did not fail) the problem is not there.
These versions do not have this issue:
```
<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
<PackageReference Include="NUnit" Version="3.11.0" />
<PackageReference Include="NUnit3TestAdapter" Version="3.13.0" />
```

stevenaw:
Fixes #3976 

This fixes an error which can come up when `ArrayList.Sort()` throws an exception when trying to sort Tuples which contain non-IComparable members, preventing comparison of equivalent collections. The exception was thrown mid-swap of items, leaving the ArrayList in a corrupt state.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionTally.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionTally.cs
@@ -89,8 +89,7 @@ namespace NUnit.Framework.Constraints
         {
             this.comparer = comparer;
 
-            foreach (object o in c)
-                _missingItems.Add(o);
+            _missingItems = ToArrayList(c);
 
             if (c.IsSortable())
             {
@@ -141,9 +140,7 @@ namespace NUnit.Framework.Constraints
         {
             if (_isSortable && c.IsSortable())
             {
-                var remove = new ArrayList();
-                foreach (object o in c)
-                    remove.Add(o);
+                var remove = ToArrayList(c);
 
                 if (TrySort(ref remove))
                 {
@@ -169,6 +166,18 @@ namespace NUnit.Framework.Constraints
                 foreach (object o in c)
                     TryRemove(o);
             }
+        }
+
+        private static ArrayList ToArrayList(IEnumerable items)
+        {
+            if (items is ICollection ic)
+                return new ArrayList(ic);
+
+            var list = new ArrayList();
+            foreach (object o in items)
+                list.Add(o);
+
+            return list;
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/CollectionTally.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionTally.cs
@@ -94,7 +94,7 @@ namespace NUnit.Framework.Constraints
 
             if (c.IsSortable())
             {
-                _isSortable = TrySort(_missingItems);
+                _isSortable = TrySort(ref _missingItems);
             }
         }
 
@@ -104,8 +104,9 @@ namespace NUnit.Framework.Constraints
             return comparer.AreEqual(expected, actual, ref tolerance);
         }
 
-        private static bool TrySort(ArrayList items)
+        private static bool TrySort(ref ArrayList items)
         {
+            var original = (ArrayList)items.Clone();
             try
             {
                 items.Sort();
@@ -113,6 +114,7 @@ namespace NUnit.Framework.Constraints
             }
             catch (InvalidOperationException e) when (e.InnerException is ArgumentException ae && ae.Message.Contains(nameof(IComparable)))
             {
+                items = original;
                 return false;
             }
         }
@@ -143,7 +145,7 @@ namespace NUnit.Framework.Constraints
                 foreach (object o in c)
                     remove.Add(o);
 
-                if (TrySort(remove))
+                if (TrySort(ref remove))
                 {
                     _sorted = true;
 
--- a/src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintTests.cs
@@ -27,6 +27,7 @@ using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading;
 using NUnit.Framework.Internal;
 using NUnit.TestUtilities.Collections;
 using NUnit.TestUtilities.Comparers;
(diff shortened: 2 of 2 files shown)


src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -104,15 +104,17 @@ private bool ItemsEqual(object expected, object actual)
             return comparer.AreEqual(expected, actual, ref tolerance);
         }
 
-        private static bool TrySort(ArrayList items)
+        private static bool TrySort(ref ArrayList items)
         {
+            var original = (ArrayList)items.Clone();
```
manfred-brands, line 109: I wonder if it would be cheaper to check if all elements are IComparable iso making a Clone.

stevenaw: Hi @manfred-brands !
I'm not terribly happy with this clone approach myself. This is a bit of a tricky edge case where, in the case of a `Tuple<T1, T2, etc>` the tuple itself may implement IComparable but will throw an exception when you try to call it. This is because the tuple's properties must all implement IComparable for the tuple itself to be sortable but the Tuple doesn't actually perform this check until runtime. Rather than for us to do this type check recursively on each item, a decision was made in https://github.com/nunit/nunit/pull/3831 to assume the happypath and catch the exception and fallback if it can't be sorted.

The thorny part of this, however, is that the in-place Array.Sort() sorting algorithm used by ArrayList + List<T> seems to leave the array in an inconsistent state if an exception is thrown in this case in .NET 4.5+. I'm still working through a few options but it is looking like the simplest option is something which avoids using Array.Sort() either directly or indirectly.

src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -168,5 +167,17 @@ void TryRemoveSlow(IEnumerable c)
                     TryRemove(o);
             }
         }
+
+        private static ArrayList ToArrayList(IEnumerable items)
+        {
+            if (items is ICollection ic)
```
stevenaw, line 173: minor optimization to avoid O(n) capacity checks and potential resizing of the ArrayList. Two common collection types (arrays and list<T>) implement this

stevenaw:
Description of the root cause for prosperity:

It looks like the array is being left in an in-between state when the first of a tuple's properties is IComparable and the second isn't and sorting requires using the second property to tie-break. This causes an exception to be thrown at runtime during sorting. For example, the following will throw and leave the array in an inconsistent state:
```
var message3 = new object();
var message4 = new object();

var actual = new[]
{
  Tuple.Create(1, message3),
  Tuple.Create(2, message3),
  Tuple.Create(1, message4),
  Tuple.Create(2, message4)
};
```

but switching the order of the tuple's properties will cause it to throw earlier in the sort algo and be left in an appropriate state:

```
var message3 = new object();
var message4 = new object();

var actual = new[]
{
  Tuple.Create(message3, 1),
  Tuple.Create(message3, 2),
  Tuple.Create(message4, 1),
  Tuple.Create(message4, 2)
};
```

This only happens in .NET 4.5+ and not .NET 4.

src/NUnitFramework/framework/Constraints/CollectionTally.cs
manfred-brands: Thanks for the explanation.
This ticket seems to have stalled, even if not ideal, this at least would fix the issue were are currently seeing after upgrading to 3.12.2

manfred-brands (approved):
Can we get this in to fix the bug?

mikkelbu (approved):
LGTM - although I share your concern about the clone. Thanks for doing this @stevenaw 👍

stevenaw:
Thanks @mikkelbu @manfred-brands 
I think we'd all prefer a different approach. I'll create an issue to more proactively detect and handle this

src/NUnitFramework/framework/Constraints/CollectionTally.cs
hazzik: This has broken our custom collection type: https://github.com/nhibernate/nhibernate-core/blob/master/src/NHibernate/Collection/Generic/SetHelpers/SetSnapShot.cs

https://github.com/nhibernate/nhibernate-core/pull/3088

stevenaw: Sorry about that @hazzik 
I think a fix for this was done on the 3.13.x branch in https://github.com/nunit/nunit/pull/4097, but it hasn't been released yet.
To be sure, would you mind opening an issue with the details specific to what you're hitting? At a minimum, perhaps it could begin a conversation about timelines for the next 3.13.x that

hazzik: @stevenaw, ok, sounds good. I think #4097 covers it. Thanks

Issue #3984, opened 2021-11-11
Add net6.0 targets


stevenaw:
We should add a net6.0 build target to the test projects so that we can test against the latest LTS runtime. We should add this to both 4.0 + 3.13 branches

PR #3986, opened 2021-11-11
Drop netcore2.1 as a target

### LINKED ISSUE #3980: Drop netcoreapp2.1 targets
It is out of support and generating build warnings.

```txt
C:\Program Files\dotnet\sdk\5.0.402\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.EolTargetFrameworks.targets(28,5): warning NETSDK1138: The target framework 'netcoreapp2.1' is out of supp
ort and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy. [C:\Src\NUnit\nunit\src\NUnitFra
mework\mock-assembly\mock-assembly.csproj]
```

manfred-brands:
Fixes #3980

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -48,11 +48,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Setup .NET Core SDK 2.1.x
-      uses: actions/setup-dotnet@v1
-      with:
-        dotnet-version: '2.1.x'
-
     - name: Setup .NET Core SDK 3.1.x
       uses: actions/setup-dotnet@v1
       with:
@@ -85,11 +80,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-
-    - name: Setup .NET Core SDK 2.1.x
-      uses: actions/setup-dotnet@v1
-      with:
-        dotnet-version: '2.1.x'
 
     - name: Setup .NET Core SDK 3.1.x
       uses: actions/setup-dotnet@v1
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -38,7 +38,6 @@ As the NUnit solution targets multiple frameworks, a single build will generate 
           net5.0
           net45
           net46
-          netcoreapp2.1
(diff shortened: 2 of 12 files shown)


manfred-brands:
The macOS failure on XML files seems unrelated.

jnm2 (approved):
It could be convenient to do https://github.com/nunit/nunit/issues/3984 at the same time since the overall number of things would remain the same.

manfred-brands:
In light of @CharliePoole's comments, I drop this. Replaced with #3988 which add .net6.
That PR also suppresses 'out of support' warnings.

CharliePoole:
Looking at #3988, I think you may have misunderstood me. My point was that I think you need to keep the out of support second-level test assemblies, _even though you are dropping 2.1 support from the framework itself_. If I'm not mistaken, dropping the actual support is a decision @rprouse has already made.

manfred-brands:
The framework itself is targeting `net45` and `netstandard2.0` and doesn't need dropping anything.
I kept `netcoreapp2.1` in the actual tested assembly: `nunit.framework.tests` which is the only assembly actually tested.
All other assemblies are referenced by this one and only need targeting the same as the framework itself.

The only place I wasn't sure was `nunitlite` where I left `netcoreapp2.1` unless someone tells me it can go from there.
But as I'm new to nunit internals, I might have it wrong.

CharliePoole:
OK, I'd say wait for Rob to to tell you what he wants to do. I thought that he wanted them all gone and I was only arguing for keeping the second-level tests - those that are run by your own tests - in order to be able to prove that a user with a .NET Core 2.1 test can still use NUnit. IMO it's insufficient to just state that analytically. Ya gotta test it!

All that said, I'm not a team member and it's just one person's opinion. :-)

rprouse:
I'd like to drop support for compiling test assemblies to 2.1, so I'd like to see this reopened.

manfred-brands:
Rebased and also removed installing netcore 2.1 from github actions.

rprouse (approved):

Issue #3867, opened 2021-06-01
nunit Framework tests do not run with "dotnet test" nor inside VS2019 (Windows)


FrankRay78:
**Expected:** as per the NUnit documentation [here](https://docs.nunit.org/articles/nunit/getting-started/dotnet-core-and-dotnet-standard.html), I should be able to run the unit tests from the command line using `dotnet test`

![image](https://user-images.githubusercontent.com/52075808/120346968-c7ceea80-c2f3-11eb-9f50-e3fdfad338a1.png)

**Actual:** No tests are ever found

#

**Environment**
Dell XPS 15, 32BG ram
Running `clrver` outputs

```
Versions installed on the machine:
v2.0.50727
v4.0.30319
```
.Net 5.0 SDK has been installed manually

#

**Steps to reproduce**

1. Open VS2019 and clone the nunit repo into a clean directory
2. Open windows PowerShell and navigate to the directory
3. Run `build --target=Test --configuration=Release`
4. NuGet packages restored, all dlls are built, all unit tests pass (for all framework targets)

![image](https://user-images.githubusercontent.com/52075808/120344050-2181e580-c2f1-11eb-8445-4cfe9f85ff8a.png)

5. Navigate to the bin\Release\net5.0 directory
6. Run `dotnet test nunit.framework.tests.dll --framework:net5.0`

![image](https://user-images.githubusercontent.com/52075808/120344475-863d4000-c2f1-11eb-994f-5027e1d2c85a.png)

7. Open the solution in VS2019 and set the test target architecture to x64

![image](https://user-images.githubusercontent.com/52075808/120344833-dcaa7e80-c2f1-11eb-99d8-8637fca13212.png)

8. Save all and close VS2019
9. Repeat steps 3, 5 and 6 above

![image](https://user-images.githubusercontent.com/52075808/120345359-50e52200-c2f2-11eb-9ff4-4b668cf3a96b.png)

10. Specify the path to the TestAdapters located here:

![image](https://user-images.githubusercontent.com/52075808/120345554-7bcf7600-c2f2-11eb-9c1d-67640496ef64.png)

11. Run the following command `dotnet test nunit.framework.tests.dll /TestAdapterPath:C:\Users\frank\Documents\Professional\Source\nunit-VS2019-2\bin\Release\net5.0 --framework:net5.0`

![image](https://user-images.githubusercontent.com/52075808/120345751-aa4d5100-c2f2-11eb-95ee-cac2af7ac0a0.png)

[...]

PR #3988, opened 2021-11-13
Issue/3984: Add NET6.0 support

### LINKED ISSUE #3980: Drop netcoreapp2.1 targets

### LINKED ISSUE #3984: Add net6.0 targets

manfred-brands:
Fixes #3984 
Fixes #3980 by suppressing warning.

Changing the SDK in global.json to 6.0.100 causes restore error with fsharp.

> D:\3rd\nUnit\nunit\src\NUnitFramework\testdata.fsharp\nunit.testdata.fsharp.fsproj : error NU1604: Project dependency System.ValueTuple does not contain an inclusive lower bound. Include a lower bound in the dependency version to ensure consistent restore results.

This only occurs on the net45 build.

The obj/project.assets.json contains:
```
"System.ValueTuple": {
  "target": "Package",
   "version": "(, )",
   "generatePathProperty": true
}
```

While previously the version was listed as `"version": "[4.4.0, )"`

`dotnet restore` only succeeds once I set the target to `net471` or greater.
Before that it thinks it needs `System.ValueTuple` version `4.3.0` 

For now I worked around it by ignoring the error.

Other changes:
* Most non-runtime test libraries only need to be compiled for netstandard2.0 as they are only executed in the context of the project that references them.
* Not sure what runtimes (.dll/.exe) need including in nunitlite. Previously net5.0 wasn't added.
* The cake script tries to explicitly run `msbuild` on windows. This fails on my machine as it can't find VS2022 preview and hence uses the wrong version of `msbuild`. As `dotnet build` works on all platforms, I added a separate commit to drop it. The build already uses `dotnet` for restore.

### DIFF
--- a/src/NUnitFramework/Directory.Build.props
+++ b/src/NUnitFramework/Directory.Build.props
@@ -17,11 +17,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <DefineConstants Condition="'$(TargetFramework)' != 'netstandard2.0'
-                            and '$(TargetFramework)' != 'netcoreapp2.1'
-                            and '$(TargetFramework)' != 'netcoreapp3.1'
-                            and '$(TargetFramework)' != 'net5.0'
-                            and '$(TargetFramework)' != 'net5.0-windows'">$(DefineConstants);THREAD_ABORT</DefineConstants>
+    <DefineConstants Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">$(DefineConstants);THREAD_ABORT</DefineConstants>
   </PropertyGroup>
 
   <!-- We always want a good debugging experience in tests -->
--- a/src/NUnitFramework/Directory.Build.props
+++ b/src/NUnitFramework/Directory.Build.props
@@ -6,7 +6,8 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\nunit.snk</AssemblyOriginatorKeyFile>
     <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
-
+    <NUnitLibraryFrameworks>net45;netstandard2.0</NUnitLibraryFrameworks>
+    <NUnitRuntimeFrameworks>net45;netcoreapp3.1;net5.0;net6.0</NUnitRuntimeFrameworks>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <OutputPath>..\..\..\bin\$(Configuration)\</OutputPath>
   </PropertyGroup>
--- a/src/NUnitFramework/mock-assembly/mock-assembly.csproj
+++ b/src/NUnitFramework/mock-assembly/mock-assembly.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net45;netcoreapp3.1</TargetFrameworks>
+    <TargetFrameworks>net45;netstandard2.0</TargetFrameworks>
     <RootNamespace>NUnit.Tests</RootNamespace>
   </PropertyGroup>
 
--- a/src/NUnitFramework/mock-assembly/mock-assembly.csproj
+++ b/src/NUnitFramework/mock-assembly/mock-assembly.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net45;netcoreapp2.1</TargetFrameworks>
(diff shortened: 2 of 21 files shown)


src/NUnitFramework/Directory.Build.props
```diff
@@ -17,11 +18,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <DefineConstants Condition="'$(TargetFramework)' != 'netstandard2.0'
-                            and '$(TargetFramework)' != 'netcoreapp2.1'
-                            and '$(TargetFramework)' != 'netcoreapp3.1'
-                            and '$(TargetFramework)' != 'net5.0'
-                            and '$(TargetFramework)' != 'net5.0-windows'">$(DefineConstants);THREAD_ABORT</DefineConstants>
+    <DefineConstants Condition="$(TargetFramework.StartsWith('net4'))">$(DefineConstants);THREAD_ABORT</DefineConstants>
```
jnm2, line 21: Option: `Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'"`

The parallel of `#if NETFRAMEWORK`

src/NUnitFramework/mock-assembly/mock-assembly.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net45;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>net45;netstandard2.0</TargetFrameworks>
```
jnm2, line 4: I don't think .NET Standard tests are supported. We should try to test as close to possible what we support, and we tell people to make their test assemblies target concrete platforms (which is any target except .NET Standard).

https://docs.nunit.org/articles/nunit/getting-started/dotnet-core-and-dotnet-standard.html#why-cant-my-tests-target-net-standard

jnm2 (changes requested):

rprouse (approved):
Other than @jnm2's comment on mockassembly, looks good.

src/NUnitFramework/mock-assembly/mock-assembly.csproj
manfred-brands: @jnm2 You would be right if this assembly actually was a test assembly. However the only test assembly is `nunit.framework.tests`, i.e. an assembly that references the `Microsot.NET.Test.Sdk` and `NUnit3TestAdapter`. That is also the only one tested in `build.cake`. All others, including the `mock-assembly` are referenced in that one.

My understanding is that the `TargetFramework` for a referenced assembly defines the API available on that platform, Only once it is executed, the `RunTimeFramework` comes into play.
There is no conditional compilation in the source code to behave differently between `netstandard2.0` and later.

If my understanding is wrong, please correct me.

src/NUnitFramework/Directory.Build.props
manfred-brands: Thanks, code updated.

src/NUnitFramework/mock-assembly/mock-assembly.csproj
jnm2: Maybe it doesn't matter, but I thought the point of mock-assembly and the testdata assembly was to mimic a real test assembly as closely as possible. `build.cake` only invokes `nunit.framework.tests`, but then `nunit.framework.tests` takes on the same role as `build.cake` as a test executor to test execution of tests.

manfred-brands: Ok. 
To ensure that these satellite assemblies use the same set of frameworks as the main `nunit.framework.tests`, I have added two properties to `Directory.Build.props`:
* `NUnitLibraryFrameworks`, used for libraries.
* `NUnitRuntimeFrameworks`, used for all test assemblies and nunitlite-runner.

This should get consistency and prevents diverging.
The only difference is that in build.cake the frameworks are still listed separately.
It would be good to fix #3867 so we could simply run `dotnet test` to test on all supported runtime frameworks.

manfred-brands: @jnm2 As you seem to be happy, can you approve please?

jnm2 (approved):
Thank you!

PR #4002, opened 2021-11-23
Only use decimal comparison if one of the operands is decimal

### LINKED ISSUE #3999: Numeric comparison fails when it should succeed.
@stevenaw This fails on 3.13-dev, but succeeds on 3.13.2, it was introduced with #3898 

```
[Test]
public void Decimal()
{
    const double value = 1.0000000000000038d;

    Assert.Multiple(() =>
    {
        Assert.That(value > 1.0, Is.True, "> 1.0");
        Assert.That(value, Is.GreaterThan(1.0), "Is.GreaterThan(1.0)");
    });
}
```

This fails with:

```
Message: 
  Is.GreaterThan(1.0)
  Expected: greater than 1.0d
  But was:  1.0000000000000038d
```

The error is in the conversion to `decimal`. Despite the type [claiming to have 28 digits of precision](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types),
`Convert.ToDecimal(value)` results in `1` dropping all fractional digits.

Even though the comparison uses `decimal` the error message uses the original values.

manfred-brands:
Fixes #3999

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Numerics.cs
+++ b/src/NUnitFramework/framework/Constraints/Numerics.cs
@@ -407,34 +407,22 @@ namespace NUnit.Framework.Constraints
             if (!IsNumericType(expected) || !IsNumericType(actual))
                 throw new ArgumentException("Both arguments must be numeric");
 
-            // Treat as decimal if one is decimal and other can be treated as decimal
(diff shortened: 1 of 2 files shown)


stevenaw (approved):
Looks great, thanks again @manfred-brands 
A second approval will be needed for merge

rprouse (approved):
Thanks.

PR #4003, opened 2021-11-23
Don't convert to decimal if double values are out of range.

### LINKED ISSUE #4000: OverflowException comparing large double values
@stevenaw This fails on 3.13-dev, but succeeds on 3.13.2, it was introduced with #3898 

```
[Test]
public void DecimalRange()
{
    const double Maximum = 1e30;
    const double Value = 0.0099999999999988987;

    Assert.That(Value, Is.LessThan(Maximum));
}
```

It fails with:
```
  Message: 
System.OverflowException : Value was either too large or too small for a Decimal.

  Stack Trace: 
Decimal.ctor(Double value)
IConvertible.ToDecimal(IFormatProvider provider)
Convert.ToDecimal(Object value)
Numerics.Compare(Object expected, Object actual)
NUnitComparer.Compare(Object x, Object y)
```

The last change for testing range only test range if either of the parameters is `decimal`.
If both are `double` no range checking is done and we get this error.

I will raise a PR for this shortly.

manfred-brands:
Fixes #4000 for master

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Numerics.cs
+++ b/src/NUnitFramework/framework/Constraints/Numerics.cs
@@ -392,7 +392,15 @@ namespace NUnit.Framework.Constraints
                 return Convert.ToDecimal(expected).CompareTo(aDec);
 
             if (IsFloatingPointNumeric(expected) || IsFloatingPointNumeric(actual))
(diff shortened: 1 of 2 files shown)


jnm2 (approved):

mikkelbu (approved):
LGTM

Issue #3961, opened 2021-10-14
OneTimeTearDown runs on a new thread with mismatched Thread Name and Worker Id


EraserKing:
I'm not sure if I'm relying on some behavior which is not guaranteed, so I'm not 100% sure if this is an issue.

I would like to have parallel testing on a class to speed up.
To make sure each worker initialize a separate instance of the specific class, I created a wrapper which returns the instance of the class under the current thread (based on *Thread.CurrentThread.Name*), or initialize a new instance if not initialized yet - so in the test class I would just call the wrapper without considering which thread I'm now on and which instance is the one I should get among a set of instances. It it something like:
```csharp
public static class TestObjectWrapper
{
    public static TestObject Get()
    {
        if (!testObjects.ContainsKey(Thread.CurrentThread.Name))
        {
            testObjects.AddOrUpdate(Thread.CurrentThread.Name, new TestObject(), (o, n) => new TestObject());
        }
        return testObjects[Thread.CurrentThread.Name];
    }

    private static ConcurrentDictionary<string, TestObject> testObjects = new ConcurrentDictionary<string, TestObject>();
}
```
So no matter where I am, I only need to call `TestObjectWrapper.Get()`, I can get an instance.

The reason why I rely on the thread name is because the wrapper is inside another project without NUnit reference, and it's also used in some other places without NUnit (so I cannot take *TestContext.CurrentContext.WorkerId* instead). Altogether, I read from stackoverflow that the thread name is identical to the worker id, which is something like *ParallelWorker#8*, or *NonParallelWorker*, or *null* (under Debug mode).

Under most scenarios it works quite well, when I only enable test fixture level parallelism (*[Parallelizable(ParallelScope.Fixtures)]*), but I found an unexpected behavior for OneTimeTearDown method in the test fixtures.
In the OneTimeTearDown, I'm hoping to get the class instance to do some clean up. So I have some code like:
```csharp
[Parallelizable(ParallelScope.Fixtures)]
public class TestClass
{
        // Actual test code omitted
        [OneTimeTearDown]
        public void OneTimeTearDown()
        {
            TestObjectWrapper.Get().Shutdown();
        }
}
```
Obviously I'm expecting getting the same instance used by test methods in this test fixture, but later I found there's high chance that this method is exectued on a completely new thread, which caused the original instance not terminated but creates a new one instead.

[...]

Issue #3426, opened 2019-11-16
ParallelScope.Fixtures - (Sometimes) Different Threads Perform OneTimeSetUp & TearDown


Vemahk:
NUnit: v3.12.0
NUnit.ConsoleRunner: v3.10.0
NUnit3TestAdapter: v3.15.1
All of the above installed via NuGet.

Hi all,

There is a long and complicated story as to why I was doing what I was doing when I discovered this. The gist of it is that the tests I am working with seed data to the database at the beginning of the test and are trying to remove the data during teardown, but in order to not have the tests' fixtures step on each other while running in parallel, I decided to mark any seeded data by the thread ID that created it so the teardown would only remove data that the thread had created (and nothing else). However, this little bit appeared to me that for whatever reason, the same thread that ran the OneTimeSetUp (and all the tests, mind you) was not always the same thread that performed the OneTimeTearDown (which is where the data deletion was taking place).

I figured it was strange that all the tests were being run by the same thread (since I was only parallelizing by fixtures and not by all), but then the TearDown was being handled by a different thread.

This isn't at all vital to what we were doing- I was only testing to see if it could be done, and I think I may know a way of engineering around it to accomplish what I was trying to do. I did want to drop this by here for someone to pick up though in case it was a problem that needed to be addressed..

I have some sample code that I have been able to get to consistently (in so far as it decides to choose another thread to perform TearDown) throw an error.

```
using NUnit.Framework;
using System;
using System.Threading;

namespace NUnitTester
{
    [TestFixture]
    [Parallelizable(ParallelScope.Fixtures)]
    public class WhenRunningTestsInParallel_Fixtures
    {
        private int _startingThread;

        [OneTimeSetUp]
        public void FixtureSetup()
        {
            ThreadLog("Entering FixtureSetup");

            _startingThread = Thread.CurrentThread.ManagedThreadId;
        }

        [OneTimeTearDown]
        public void FixtureTeardown()
        {
            ThreadLog("Entering FixtureTeardown");

            Assert.That(Thread.CurrentThread.ManagedThreadId, Is.EqualTo(_startingThread), "The threads running setup and teardown are different.");
        }

        [SetUp]
        public void TestSetup()
        {
            ThreadLog("Entering TestSetup");
        }

        [TearDown]
        public void TestTeardown()
        {
            ThreadLog("Entering TestTeardown");
        }

        [TestCase(0)]
        [TestCase(1)]
        [TestCase(2)]
        [TestCase(3)]
[...]

Issue #2610, opened 2017-12-11
SingleThreadedAttribute ignored on OneTimeTearDown with ParallelScope set to All or Fixtures


jtomkiew-ITM:
NUnit 3.9.0

`OneTimeTearDown` from TestFixture with `[SingleThreaded]` and `[Parallelizable(ParallelScope.Fixtures)]` (or `ParallelScope.All`) attributes is **not always** run on the same thread as previous test methods (`OneTimeSetup`, `SetUp`, `Test`, `TearDown`). Thread differs, but the `NUnit.Framework.TestContext.CurrentContext.WorkerId` stays the same.

I would expect for `SingleThreaded` attribute to force usage of the same thread on whole fixture.

The same code works as expected on NUnit 3.7.1 and stops working on NUnit 3.8.0.

Tested with R# and nunit3-console 3.7.0 test runners.

#### Sample code:

```cs
using System.Threading;
using NUnit.Framework;

namespace Tests.nunittest
{
    [TestFixture, SingleThreaded]
    [Parallelizable(ParallelScope.Fixtures)] // or ParallelScope.All
    internal class TestClass
    {
        private Thread _setupThread;

        [OneTimeSetUp]
        public void OTSetup()
        {
            _setupThread = Thread.CurrentThread;
        }

        [SetUp]
        public void Setup()
        {
            Assert.That(Thread.CurrentThread, Is.EqualTo(_setupThread));
        }

        [Test]
        public void Test()
        {
            Assert.That(Thread.CurrentThread, Is.EqualTo(_setupThread));
        }

        [TearDown]
        public void Teardown()
        {
            Assert.That(Thread.CurrentThread, Is.EqualTo(_setupThread));
        }

        [OneTimeTearDown]
        public void OTTeardown()
        {
            Assert.That(Thread.CurrentThread, Is.EqualTo(_setupThread)); // failure, not always
        }
    }
}
```

#### AssemblyInfo.cs :
```cs
[assembly: LevelOfParallelism(4)]
```

CharliePoole:
It's possible that the WorkerId is actually wrong, and is merely copied from the initial worker id. Running on a different thread would normally indicate we are running on a different worker.

I marked this confirm, but it's very credible because we __do__ enqueue the OneTimeTearDown for execution and I don't believe we check whether the fixture is SIngleThreaded before doing it.

Brian-Watson:
[...]

PR #4004, opened 2021-11-25
Force the OneTimeTearDown to be executed on the current thread if STA

EraserKing:
This aims to be a workaround for #3961 

When both `SingleThreaded` and `Parallizable` attributes are attached to the test fixture, `SingleThreadedAttribute.ApplyToContext` may be executed later than expected (even after all test methods in the test fixtures are executed), so it leads to the following result:

1. The parent context still has `IsSingleThreaded` false, and all the test methods are executed on this context: (`targetApartment` and `currentApartment` are both `Unknown`, in `WorkItem.Execute`), so `WorkItem.RunOnCurrentThread` is always called.

2. Then, when the `OneTimeTearDown` of test fixture is executed, a new `OneTimeTearDownWorkItem` is created based on the current work item (see `CompositeWorkItem.OnAllChildItemsCompleted`), which *inherits* the parent `ExecutionStrategy`. At that time, since the `Context.IsSingleThreaded` may still haven't been set to true, the `ExecutionStrategy` will be `Parallel`.

3. At this time, not sure about the call stack, `SingleThreadedAttribute.ApplyToContext` is called and `IsSingleThreaded` set to true.

4. When the one time tear down work item is actually executed, though `Context.IsSingleThreaded` is now true, but the `ParallelExecutionStrategy` is already set in step 2, so it will still fall into the `ParallelExecutionStrategy.Parallel` branch and executed by a parallel queue, which may happen on another thread (see `ParallelWorkItemDispatcher.Dispatch`).

6. The PR adds a workaround, when it's going to execute the work item, double-check if `Context.IsSingleThreaded` is set to true, and if so, force it to run on the current thread.

The clean fix should be investiage why step 3 is delayed but that's beyond my ability. Thanks.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -191,24 +191,32 @@ namespace NUnit.Framework.Internal.Execution
                     composite.Completed += OnWorkItemCompletion;
                 }
 
-            switch (strategy)
+            // The attribute SingleThreaded may not be applied to the content in time, so here as a workaround, check for the flag again
+            if (work.Context.IsSingleThreaded)
             {
-                default:
-                case ParallelExecutionStrategy.Direct:
-                    work.Execute();
-                    break;
-                case ParallelExecutionStrategy.Parallel:
-                    if (work.TargetApartment == ApartmentState.STA)
-                        ParallelSTAQueue.Enqueue(work);
-                    else
-                        ParallelQueue.Enqueue(work);
-                    break;
-                case ParallelExecutionStrategy.NonParallel:
-                    if (work.TargetApartment == ApartmentState.STA)
-                        NonParallelSTAQueue.Enqueue(work);
-                    else
-                        NonParallelQueue.Enqueue(work);
-                    break;
+                work.Execute();
+            }
+            else
+            {
+                switch (strategy)
+                {
+                    default:
+                    case ParallelExecutionStrategy.Direct:
+                        work.Execute();
+                        break;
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=4004) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=4004) before we can accept your contribution.<br/><br/>:x: EraserKing [sign now](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=4004)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/nunit/nunit?pullRequest=4004) it.</sub>

stevenaw:
NOTE to self or other reviewers. This may also fix:
- #2610 
- #3426

OsirisTerje:
@stevenaw You looked at this earlier.  It doesn't come with any tests, but do (or did) you see any possible side-effects of this?  If not, any reason for not merging it in?

stevenaw (approved):
I hadn't had the chance to pull and run the code, but the changes "look" safe to me and appear to be backwards compatible too. I'd be fine if it were merged in.

I think I'd mostly avoided any explicit action on it earlier out of a desire to dig into if there was also any relation to the changes in the underlying bug fixed by #4099 , but that investigation doesn't warrant holding this up.

Not sure if we also prefer Rob review since he self-assigned, but I can approve from my own involvement in this. I understand he's a little busy right now. Otherwise, feel free to merge this in @OsirisTerje 🙂

PR #4007, opened 2021-11-27
Eagerly detect sortable types for equivalency tests in 3.13.x

### LINKED ISSUE #3998: Eagerly determine when a set is unsortable
While trying to optimize collectiontally by pre-sorting items in https://github.com/nunit/nunit/issues/3841, an edge case was hit where a Tuple<T1, T2> may implement IComparable even if it's underlying properties didn't. This could cause an exception at runtime. A workaround was done in the PR (https://github.com/nunit/nunit/pull/3881) to catch the exception and fallback to a slower path.

https://github.com/nunit/nunit/issues/3976 was subsequently filed around an edge-case of this edge-case, where a tuple which has its first property implement IComparable and it's second doesn't could throw an exception at such a point that it left the array in a corrupt state. The workaround in PR https://github.com/nunit/nunit/pull/3978 was to copy the array prior to sorting.

We should consider proactively detecting these cases and falling back to the slow path. This will have a few benefits:
1. Simpler code
2. Likely lower memory overhead due to not needing to copy the array internally
3. Better surfacing of issues in calling code by allowing exceptions in user-defined IComparers to halt the comparison entirely

stevenaw:
Fixes #3998

This turned out much simpler than I expected. This PR will eagerly determine if an array can be sorted without issue for equivalency tests. This PR targets 3.13, but I'll cherry-pick the commits to 4.0 once the implementation is finalized.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Extensions/TypeExtensions.cs
+++ b/src/NUnitFramework/framework/Internal/Extensions/TypeExtensions.cs
@@ -8,5 +8,20 @@ namespace NUnit.Framework.Internal.Extensions
     {
         public static bool ImplementsIComparable(this Type type) =>
             type?.GetInterface("System.IComparable") != null;
+
+        public static bool IsSortable(this Type type)
+        {
+            if (!type.ImplementsIComparable())
+                return false;
+
+            if (TypeHelper.IsTuple(type) || TypeHelper.IsValueTuple(type))
+            {
+                foreach (var typeArg in type.GetGenericArguments())
+                    if (!typeArg.IsSortable())
+                        return false;
+            }
+
+            return true;
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


src/NUnitFramework/framework/Internal/Extensions/TypeExtensions.cs
```diff
@@ -8,5 +8,20 @@ internal static class TypeExtensions
...
+
+        public static bool IsSortable(this Type type)
+        {
+            if (!type.ImplementsIComparable())
+                return false;
+
+            if (TypeHelper.IsTuple(type) || TypeHelper.IsValueTuple(type))
+            {
+                foreach (var typeArg in type.GetGenericArguments())
+                    if (!typeArg.IsSortable())
```
stevenaw, line 20: Recursively check + confirm that all of a tuple's properties can reasonably be expected to be sortable.

manfred-brands (approved):
Looks good.
Good example of recursion making things simple.

rprouse (approved):
Nice

PR #4014, opened 2021-12-03
DictionaryContainsKeyValuePairConstraint supports generic IDictionary

### LINKED ISSUE #3964: DictionaryContainsKeyValuePairConstraint doesn't work with `IDictionary<TKey, TValue>`
`DictionaryContainsKeyValuePairConstraint` fails when actual object implements only `IDictionary<TKey, TValue>`, but not `IDictionary`:

```csharp
[Test]
void Test()
{
    var dictionary = new TestDictionary();
    Assert.That(dictionary, new DictionaryContainsKeyValuePairConstraint(1, "hello"));
}

class TestDictionary : IDictionary<int, string>
{
    // Dictionary implementation
}

// Test fails with next error:
// 1) Error : NUnit.Framework.Constraints.DictionaryContainsKeyValuePairConstraintTests.ShouldCallContainsKeysMethodOnDictionary
// System.ArgumentException : Expected: IDictionary But was: TestDictionary (Parameter 'actual')
```

**NUnit version:** 3.13.2
**Runner:** NUnit3TestAdapter

It looks like the bug is here: https://github.com/nunit/nunit/blob/2c21bff1d8891c2df3cad1f5ed34551e1c4edb9e/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs#L43

louis-z:
- Adapt `DictionaryContainsKeyValuePairConstraint` so that it works with types that implement
  - `IDictionary<TKey, TValue>` but
  - not `IDictionary`
- For backward compatibility, keep throwing same exception with same message as before when a failures occurs

Fixes #3964

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs
@@ -1,6 +1,9 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
+using System;
 using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
 using NUnit.Framework.Internal;
 
 namespace NUnit.Framework.Constraints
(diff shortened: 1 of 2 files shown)


mikkelbu (commented):
Looks good to me @louis-z. Thanks for the contribution. We will need one more review before merging this.

louis-z:
> Looks good to me @louis-z. Thanks for the contribution. We will need one more review before merging this.

No problem, I was glad to help. 😉 
I just added an assertion to my test, to make sure that looking up a non-existent KVP also works.

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs
```diff
@@ -40,12 +43,31 @@ public override string Description
...
+            {
+                foreach (var entry in dictionary)
+                    if (ItemsEqual(entry, _expected))
+                        return true;
+                return false;
+            }
+
+            // If 'actual' implements IDictionary<TKey, TValue>, construct an 'expected' KeyValuePair<TKey, TValue>
+            // and look it up by iterating using IEnumerable
+            if (actual.GetType().GetInterfaces().Any(i => i.FullName.StartsWith("System.Collections.Generic.IDictionary`2[")))
```
Dreamescaper, line 59: I'd suggest smth like that instead:
```csharp
i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IDictionary<,>)
```

louis-z: That's a good idea. Just made the change. Thanks @Dreamescaper.

Dreamescaper (approved):

Issue #1821, opened 2016-10-03
Responding to cooperative cancellation within test fixture


JackUkleja:
I am currently using NUnit for writing system tests. 

Because these tests are long running I need the ability to cancel the test runs at will. NUnit already supports this but as far as I can tell there is no way within a test fixture itself to know the test run has been cancelled.

The reason I need to know is because I want to be able to clean up system resources that were enrolled in running the system tests e.g. stop/uninstall Windows services.

Leaving aside whether this is the right tool for the job (it works well for our purposes / build chain etc.) is it possible to do this with NUnit?

It does appear there is mention of a "cooperative cancel" design within NUnit but I cannot figure out at what level this API is exposed. It seems maybe it's only exposed at the test runner level and is not exposed to tests themselves? I can understand this might be by-design due to keeping tests ignorant of the framework, but then again TestContext is available so maybe this is not hard and fast rule.

CharliePoole:
The notion of "cooperative cancellation" in NUnit refers to cooperation between the runner and the framework itself. The framework stops running any new tests and waits for outstanding tests to finiish. As tests finish, all teardowns and one time teardowns are executed, giving you the opportunity to clean up any resources. The teardowns are not intended to know that the run was cancelled and should be coded so that they work whether the run was cancelled or the test simply completed normally.

There is also a "forced cancel", which actually aborts the threads that are running tests. No teardowns are run in this case as it's considered a last resort. Currently, there is no way to abort a running test and still run the teardowns.

The API for running tests is part of the NUnit test engine. It supports this type of cancellation for use in runners that provide access to it. The nunit console runner doesn't use it at all. We anticipate using it in the new GUI, just as we did in the V2 GUI.

None of this is currently accessible to tests. There is a flag in the execution context that indicates a cancel has been requested. I suppose we could expose this in TestContext for any tests that wanted to examine it periodically.

The key issue for this problem is to define how you would like to tell the test to cancel as a user.

JackUkleja:
I guess the ideal solution for me would simply be if the tear downs were called irrespective of whether the run finished normally or was cancelled. My apologies if this is already how it works - I've not had a chance to verify it.

I don't need to take specific action on a cancellation vs normal termination, I just need to make sure that the same clean up is run in either case.

[...]

PR #4025, opened 2021-12-27
Feature/4021 cancellationtoken

### LINKED ISSUE #3859: Proper test result exception message when hitting `TimeoutAttribute`
Hey guys,

lately I was expetimenting with the `TimeoutAttribute` and ran into a few timeouts for my tests. I only saw a `Exception doesn't have a stacktrace` as a test result message, which confused me at first sight. Then I realized, the test were hitting the timeout.

I tried to dig into the code to see, if implementing a timeout exception message is possible, but havn't found anything so far.

So I'd like to suggest a better test result message when hitting the timeout.

Code to reproduce:
```
[Test]
[Timeout(1000)]
public async Task Timeout()
{
    // Arrange

    // Act
    await Task.Delay(2000).ConfigureAwait(false);

    // Assert
    Assert.Pass();
}
```

Thanks and kind regards

### LINKED ISSUE #4021: Dangerous behaviour of TimeoutAttribute on NET Core and later
Due to the [removal of `Thread.Abort` in later versions of .NET](https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/5.0/thread-abort-obsolete), the `TimeoutAttribute` no longer stops the long running test.

This means the test keeps running in the background, despite that the nunit runner moves to the next test.
As an example, see the following testfixture:

```
internal class TimeOut
{
    private string _logFilePath;

    [OneTimeSetUp]
    public void OneTimeSetup()
    {
        _logFilePath = Path.Combine(TestContext.CurrentContext.TestDirectory, "Timeout.txt");
        File.Delete(_logFilePath);
    }

    private void WriteLine(string text)
    {
        File.AppendAllText(_logFilePath, text + Environment.NewLine);
    }

    [TearDown]
    public void Teardown()
    {
        WriteLine("Teardown, result: " + TestContext.CurrentContext.Result.Outcome);
    }

    [Test]
    [Order(1)]
    [Timeout(110)]
    public void WithTimeout()
    {
        LongTest("WithTimeout");
    }

    [Test]
    [Order(2)]
    public void NoTimeout()
    {
        LongTest("NoTimeout");
    }

    private void LongTest(string testName)
    {
        try
        {
            for (int i = 0; i < 5; i++)
            {
                WriteLine($"{testName}: Iteration {i}");
                Thread.Sleep(100);
            }
            Assert.Pass("Success");
        }
        catch (Exception ex)
        {
            WriteLine(ex.Message);
        }
        finally
        {
            WriteLine("Task finished");
        }
    }
}
```

When running this on NET48, the `net48/Timeout.txt' file contains:

```
WithTimeout: Iteration 0
WithTimeout: Iteration 1
Thread was being aborted.
Task finished
Teardown, result: Failed:Cancelled
NoTimeout: Iteration 0
NoTimeout: Iteration 1
NoTimeout: Iteration 2
NoTimeout: Iteration 3
NoTimeout: Iteration 4
Task finished
Teardown, result: Passed
```

This shows the 1st test being aborted, and its `Teardown` method running, before the 2nd test starts.

When running on .NET5, the `net5.0/Timeout.txt` contains:

```
WithTimeout: Iteration 0
WithTimeout: Iteration 1
NoTimeout: Iteration 0
WithTimeout: Iteration 2
NoTimeout: Iteration 1
WithTimeout: Iteration 3
NoTimeout: Iteration 2
WithTimeout: Iteration 4
NoTimeout: Iteration 3
Task finished
NoTimeout: Iteration 4
Task finished
Teardown, result: Passed
```

This shows that the `WithTimeout` test runs to the end but with no `Teardown` running.
At the same time, the 2nd test is running in parallel, starting after the timeout of the 1st test.

There are multiple issues:
1. `TearDown` is not run when the test 'times out'
2. The test keeps running, which on tests with shared instance fields could be problematic, possibly interfering with other tests.

If I add a `_counter` field, set to `0` in `SetUp`, increment it in the for-loop and `Assert.That(_counter, Is.EqualTo(5));`, in the actual test, the net5.0 run fails with:
```
 Failed NoTimeout [507 ms]
  Error Message:
     Expected: 5
  But was:  8

  Stack Trace:
     at Tests.TimeOut.NoTimeout() in C:\Users\m.brands\source\Tests\TimeOut.cs:line 53
```

The question is what can be done about it.
1. Mark the `TimeoutAttribute` as obsolete except for the build targeting net framework?
2. Add an analyzer rule in nunit.analyzers warning about the drawbacks on non-framework builds.
3. Can we fix the behaviour to what it was before. Not likely.
4. Add a `CancellationToken` to `TestContext`, set to either `CancellationToken.None` if no timeout set, or a token that will be cancelled after the time out, see below. This approach will require updating end-users tests to 'the new way'.
```
using CancellationTokenSource cancellationTokenSource = new CancellationTokenSource(110);
await LongTest("WithTimeout", cancellationTokenSource.Token);
Assert.That(_counter, Is.EqualTo(5));
```

manfred-brands:
Fixes #3859 by setting both _message_ and _label_. 
Fixes #4021

It doesn't fix the current behaviour of `TimeoutAttribute` but allows for a _modern_ alternative.

Two commits:
1. Add CancellationToken to Test(Execution)Context which will be cancelled on Timeout.
2. Allow CancellationToken to be injected into Test methods

### DIFF
--- a/src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
+++ b/src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
@@ -43,6 +43,24 @@ namespace NUnit.Framework.Internal
             return IsAsyncOperation(@delegate.GetMethodInfo());
         }
 
+#if !THREAD_ABORT
+        public static bool AcceptsCancellationToken(MethodInfo method)
+        {
+            ParameterInfo[] parameters = method.GetParameters();
+            return parameters.Length > 0 && parameters[parameters.Length - 1].ParameterType == typeof(CancellationToken);
+        }
+
+        public static bool LastArgumentIsCancellationToken(object[] argumentList)
+        {
+            if (argumentList == null || argumentList.Length == 0)
+                return false;
+
+            object lastArgument = argumentList[argumentList.Length - 1];
+
+            return lastArgument != null && lastArgument.GetType() == typeof(CancellationToken);
+        }
+#endif
+
         public static object Await(Func<object> invoke)
         {
             Guard.ArgumentNotNull(invoke, nameof(invoke));
--- a/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
@@ -214,6 +214,15 @@ namespace NUnit.Framework.Internal.Builders
             else if (parms == null || !parms.HasExpectedResult)
                 return MarkAsNotRunnable(testMethod, "Method has non-void return value, but no result is expected");
 
+#if !THREAD_ABORT
+            if (AsyncToSyncAdapter.AcceptsCancellationToken(testMethod.Method.MethodInfo) &&
+                !AsyncToSyncAdapter.LastArgumentIsCancellationToken(arglist))
+            {
+                // Implict CancellationToken argument
+                argsProvided++;
+            }
+#endif
+
             if (argsProvided > 0 && maxArgsNeeded == 0)
                 return MarkAsNotRunnable(testMethod, "Arguments provided for method with no parameters");
 
--- a/src/NUnitFramework/framework/Internal/Commands/TestMethodCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TestMethodCommand.cs
@@ -21,7 +21,10 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
+#if !THREAD_ABORT
 using System;
+using System.Threading;
+#endif
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal.Builders;
 
@@ -33,8 +36,8 @@ namespace NUnit.Framework.Internal.Commands
     /// </summary>
     public class TestMethodCommand : TestCommand
     {
-        private readonly TestMethod testMethod;
-        private readonly object[] arguments;
+        private readonly TestMethod _testMethod;
+        private readonly object[] _arguments;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="TestMethodCommand"/> class.
@@ -42,8 +45,8 @@ namespace NUnit.Framework.Internal.Commands
         /// <param name="testMethod">The test.</param>
         public TestMethodCommand(TestMethod testMethod) : base(testMethod)
         {
-            this.testMethod = testMethod;
-            this.arguments = testMethod.Arguments;
+            this._testMethod = testMethod;
+            this._arguments = testMethod.Arguments;
         }
 
         /// <summary>
@@ -64,8 +67,8 @@ namespace NUnit.Framework.Internal.Commands
 
             object result = RunTestMethod(context);
 
-            if (testMethod.HasExpectedResult)
-                NUnit.Framework.Assert.AreEqual(testMethod.ExpectedResult, result);
+            if (_testMethod.HasExpectedResult)
+                Assert.AreEqual(_testMethod.ExpectedResult, result);
 
             context.CurrentResult.SetResult(ResultState.Success);
 
@@ -77,7 +80,7 @@ namespace NUnit.Framework.Internal.Commands
 
         private object RunTestMethod(TestExecutionContext context)
         {
-            var methodInfo = MethodInfoCache.Get(testMethod.Method);
+            var methodInfo = MethodInfoCache.Get(_testMethod.Method);
 
             if (methodInfo.IsAsyncOperation)
             {
@@ -89,7 +92,21 @@ namespace NUnit.Framework.Internal.Commands
 
         private object InvokeTestMethod(TestExecutionContext context)
         {
-            return testMethod.Method.Invoke(context.TestObject, arguments);
+		    var methodInfo = MethodInfoCache.Get(_testMethod.Method);
+		
+            object[] arguments = _arguments;
+
+#if !THREAD_ABORT
+            if (AsyncToSyncAdapter.AcceptsCancellationToken(_testMethod.Method.MethodInfo) &&
+                !AsyncToSyncAdapter.LastArgumentIsCancellationToken(_arguments))
+            {
+                arguments = new object[_arguments.Length + 1];
+                Array.Copy(_arguments, arguments, _arguments.Length);
+                arguments[_arguments.Length] = context.CancellationToken;
+            }
+#endif
+
+            return _testMethod.Method.Invoke(context.TestObject, arguments);
         }
     }
 }
--- a/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
@@ -124,24 +124,47 @@ namespace NUnit.Framework.Internal.Commands
         {
             try
             {
+                using var cancellationTokenSource = new CancellationTokenSource();
+
+                context.CancellationToken = cancellationTokenSource.Token;
                 var testExecution = RunTestOnSeparateThread(context);
-                if (Task.WaitAny(new Task[] { testExecution }, _timeout) != -1
+                var tasks = new Task[] { testExecution };
+                if (Task.WaitAny(tasks, _timeout) != -1 
                     || _debugger.IsAttached)
                 {
                     context.CurrentResult = testExecution.GetAwaiter().GetResult();
                 }
                 else
                 {
-                    string message = $"Test exceeded Timeout value of {_timeout}ms";
+                    // Try to cancel the task and see if it completes quickly.
+                    cancellationTokenSource.Cancel();
+                    if (Task.WaitAny(tasks, 10) != -1)
+                    {
+                        context.CurrentResult = testExecution.GetAwaiter().GetResult();
+                    }
+                    else
+                    {
+                        string message = $"Test exceeded Timeout value of {_timeout}ms";
 
-                    context.CurrentResult.SetResult(
-                        new ResultState(TestStatus.Failed, message),
-                        message);
+                        context.CurrentResult.SetResult(
+                            new ResultState(TestStatus.Failed, message),
+                            message);
+                    }
                 }
+            }
+            catch (OperationCanceledException)
+            {
+                string message = $"Test exceeded Timeout value of {_timeout}ms";
+                context.CurrentResult.SetResult(new ResultState(
+                    TestStatus.Failed, message, FailureSite.Test), message);
             }
             catch (Exception exception)
             {
                 context.CurrentResult.RecordException(exception, FailureSite.Test);
+            }
+            finally
+            {
+                context.CancellationToken = CancellationToken.None;
             }
 
             return context.CurrentResult;
@@ -149,7 +172,7 @@ namespace NUnit.Framework.Internal.Commands
 
         private Task<TestResult> RunTestOnSeparateThread(TestExecutionContext context)
         {
-            return Task.Run(() => innerCommand.Execute(context));
+            return Task.Run(() => innerCommand.Execute(context), context.CancellationToken);
         }
 #endif
     }
--- a/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
+++ b/src/NUnitFramework/framework/Internal/TestExecutionContext.cs
@@ -345,6 +345,13 @@ namespace NUnit.Framework.Internal
         /// </summary>
         public int TestCaseTimeout { get; set; }
 
+#if !THREAD_ABORT
+        /// <summary>
+        /// Gets or sets the <see cref="CancellationToken"/> for the test case.
+        /// </summary>
+        public CancellationToken CancellationToken { get; internal set; } = CancellationToken.None;
+#endif
+
         /// <summary>
         /// Gets a list of ITestActions set by upstream tests
         /// </summary>
@@ -452,9 +459,7 @@ namespace NUnit.Framework.Internal
         /// </summary>
         public void IncrementAssertCount(int count)
         {
-            // TODO: Temporary implementation
-            while (count-- > 0)
-                Interlocked.Increment(ref _assertCount);
+            Interlocked.Add(ref _assertCount, count);
         }
 
         /// <summary>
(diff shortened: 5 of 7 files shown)


manfred-brands:
If accepted requires update to nunit.analyzers rule NUnit1027 which checks parameters, to allow for CancellationToken parameter, which must be last.

rprouse (commented):
This looks good but I'm not setup for a proper review right now. I'll have to get back to it.

src/NUnitFramework/tests/Attributes/TimeoutTests.cs
```diff
@@ -392,8 +394,92 @@ public void TimeoutCausesOtherwisePassingTestToFailWithoutDebuggerAttached()
...
+
+            Assert.That(result.ResultState.Status, Is.EqualTo(TestStatus.Failed));
+            Assert.That(result.ResultState.Site, Is.EqualTo(FailureSite.Test));
+            Assert.That(result.Message, Is.EqualTo($"Test exceeded Timeout value of {SampleTests.Timeout}ms"));
+        }
+
+        [Test]
+        public void TestWithCancellationTokenThatTimesOutIsRanToCompletionWhenDebuggerIsAttachedBeforeTimeOut()
+        {
+            // give
```
mikkelbu, line 469:
```suggestion
            // given
```

mikkelbu:
I've tried to make the failing CIs run, but I'm not able to force trigger either. So we probably need to force push a new commit (or force push the latest).

src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
```diff
@@ -203,6 +203,14 @@ private static bool CheckTestMethodSignature(TestMethod testMethod, TestCasePara
             else if (parms == null || !parms.HasExpectedResult)
                 return MarkAsNotRunnable(testMethod, "Method has non-void return value, but no result is expected");
 
+#if !THREAD_ABORT
+            if (AsyncToSyncAdapter.AcceptsCancellationToken(testMethod.Method.MethodInfo))
+            {
+                // Implict CancellationToken argument
+                argsProvided++;
```
mikkelbu, line 210: Will this also work if the user is already supplying the CancellationToken or will we then get a off-by-one error. E.g. if the user now have something like (probably not very likely - but could we break somebody?)

```csharp
    [TestCaseSource(nameof(TestCases))]
    public void DivideTest(int n, CancellationToken token)
    {
...
    }

    static object[] TestCases=
    {
        new object[] { 5, someCancellationToken },
        new object[] { 12, anotherCancellationToken }
    };
```

manfred-brands: Theoretically yes. I will update the code to check if the actual parameters already has a CancellationToken.
I'm on leave without a computer but will do this next weekend when I get back.

manfred-brands:
The comment change triggered a build.

manfred-brands:
Thanks @mikkelbu You were right. I added a test similar to yours and updated the code to ensure it works.

src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
```diff
@@ -43,6 +43,24 @@ public static bool IsAsyncOperation(Delegate @delegate)
             return IsAsyncOperation(@delegate.GetMethodInfo());
         }
 
+#if !THREAD_ABORT
```
Dreamescaper, line 46: I wouldn't restrict it for Core only.
It is possible that my tests will have multiple TargetFrameworks - for both .NET Framework and .NET Core. I would prefer to be able to have CancellationToken parameter, even if it no-op for .NET Framework.

manfred-brands: For .NET Framework we can do the same and potentially drop the `Thread.Abort` in case the test method expects a `CancellationToken` or turn this into a two prong approach. Do `CancellationToken` first so _behaving_ code works, followed by a `Thread.Abort` a bit later.

stevenaw:
Hi @manfred-brands Just wanted to circle back to this. I've been waiting to get involved in this PR myself until I had a decent amount of time to sink into understanding it. I think this would be a great fix to eventually land into a release as this seems to be a recurring issue for people

I hope to be able to get to it within the coming weeks. Can you help me understand if it's in need of a review or if it's awaiting changes?

manfred-brands:
> I hope to be able to get to it within the coming weeks. Can you help me understand if it's in need of a review or if it's awaiting changes?

Hi @stevenaw this is not a fix for the `TimeOutAttribute` on non-Framework runtimes.
I'm now leaning to introduce a new `CancelAfterAttribute` instead for this to make it clear it is something completely different.

The current code works, I just rebased it today on v13.3-dev.
It is still using the `TimeoutAttribute`.

The changes are the introduction of a `CancellationToken` to the `TestContext`.
The tests can either get this from the `TestContext` or specify an additional test method parameter.

src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
stevenaw: It may just be that I haven't had my coffee yet, but I'd like to be clear on the suggestion here.

If I understand your suggestion here @manfred-brands , are you saying we should always allow an extra `CancellationToken` param to a test method and only use cooperative cancellation on netstandard builds of the framework? And that a follow-up change could use `Thread.Abort` to cancel tests on netfx builds of the framework?

If so, I agree with that approach.

stevenaw: Actually I'm curious about how it could behave if a netstandard build were used in a netfx context

src/NUnitFramework/framework/Internal/Commands/TestMethodCommand.cs
```diff
@@ -33,17 +36,17 @@ namespace NUnit.Framework.Internal.Commands
     /// </summary>
     public class TestMethodCommand : TestCommand
     {
-        private readonly TestMethod testMethod;
-        private readonly object[] arguments;
+        private readonly TestMethod _testMethod;
+        private readonly object[] _arguments;
```
stevenaw, line 40: 👍Thanks for this clean-up

src/NUnitFramework/framework/Internal/Commands/TestMethodCommand.cs
```diff
@@ -89,7 +92,21 @@ private object RunTestMethod(TestExecutionContext context)
 
         private object InvokeTestMethod(TestExecutionContext context)
         {
-            return testMethod.Method.Invoke(context.TestObject, arguments);
+		    var methodInfo = MethodInfoCache.Get(_testMethod.Method);
```
stevenaw, line 95: nitpick: Very minor whitespace issue, I think there's two tabs at the beginning of the line throwing this off.
Given this uses `MethodInfoCache` this likely accidentally came into the recent rebasing

```suggestion
            var methodInfo = MethodInfoCache.Get(_testMethod.Method);
```

src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
manfred-brands: > Actually I'm curious about how it could behave if a netstandard build were used in a netfx context

NuGet will pick the netfx build for a netfx context.
As said I would rather have a `CancelAfterAttribute` working for both and mark the `TimeoutAttribute` as Framework only.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -124,32 +124,55 @@ public override TestResult Execute(TestExecutionContext context)
         {
             try
             {
+                using var cancellationTokenSource = new CancellationTokenSource();
+
+                context.CancellationToken = cancellationTokenSource.Token;
                 var testExecution = RunTestOnSeparateThread(context);
-                if (Task.WaitAny(new Task[] { testExecution }, _timeout) != -1
```
stevenaw, line 128: question: I know this builds off the existing code pattern to use `Task.WaitAny(Task[], int)`, but since we're adding a CTS here would it make sense to instead use `CTS.CancelAfter(int)` here?

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -345,6 +345,13 @@ internal int AssertCount
         /// </summary>
         public int TestCaseTimeout { get; set; }
 
+#if !THREAD_ABORT
+        /// <summary>
+        /// Gets or sets the <see cref="CancellationToken"/> for the test case.
+        /// </summary>
+        public CancellationToken CancellationToken { get; internal set; } = CancellationToken.None;
```
stevenaw, line 352: Similar to @Dreamescaper 's question above, would this mean we'd have a different API shape for netstandard + netfx builds of the framework here and in `TestContext` below?  Or would we want to expose this in all builds and just have it be a (temporary?) no-op when run in a netfx context?

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -452,9 +459,7 @@ public void IncrementAssertCount()
         /// </summary>
         public void IncrementAssertCount(int count)
         {
-            // TODO: Temporary implementation
-            while (count-- > 0)
-                Interlocked.Increment(ref _assertCount);
+            Interlocked.Add(ref _assertCount, count);
```
stevenaw, line 462: 💯

src/NUnitFramework/tests/Attributes/TimeoutTests.cs
```diff
@@ -244,6 +248,23 @@ public void TestThatTimesOutIsRanToCompletionWhenDebuggerIsAttachedBeforeTimeOut
...
+        private static readonly CancellationToken CancelledToken = new CancellationToken(true);
+        private static readonly CancellationToken PendingToken = new CancellationToken(false);
+
+        static readonly object[] TokenTestCases = new object[]
+        {
+            new object[] { true, CancelledToken },
+            new object[] { false, PendingToken },
+        };
+
+        [TestCaseSource(nameof(TokenTestCases))]
```
stevenaw, line 261: 👍thanks for adding this, there can be some interesting edge cases when dealing with parameterized tests.
Out of curiousity, does `TestCaseAttribute` work with this pattern of injecting a `CancellationToken` too?

Separate question: `TestCaseAttribute` also supports targeting test methods with a `params[]` array as the last method argument. Since can't have _both_ a `CancellationToken` and a `params object[]` as the last args, how would cooperative cancellation look there? Would a test have to read the `CancellationToken` from the `TestContext` instead?

Something like this pseudocode?
```cs
[TestCase(1,2,3, ExpectedResult = 6)]
public async Task<int> ParameterizedSum(params int[] numbers) {
  return await TestApi.SumAsync(numbers, TestContext.Current.CancellationToken);`
}
```

stevenaw (commented):
Thanks for having started on this @manfred-brands as well as the incremental cleanup!
I asked a few questions, mostly around cross-platform behaviours and edge cases with parameterized tests

src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
stevenaw: Ok, that's good about NuGet. I'd been concerned that some historical upgrade paths could inadvertently pull in a netstandard build but it sounds like it's not possible.

stevenaw:
> Hi @stevenaw this is not a fix for the `TimeOutAttribute` on non-Framework runtimes. I'm now leaning to introduce a new `CancelAfterAttribute` instead for this to make it clear it is something completely different.

Sorry, I'd thought this was the fix for non-framework runtimes. Can you help me understand how it's different?

goblinmaks:
@manfred-brands What is next step for this PR ?

manfred-brands:
> @manfred-brands What is next step for this PR ?

Good question.
We cannot "fix" timeout attribute for .NET Core as there is no Thread.Abort to stop any runaway test.
Personally, I don't like reporting an error and then let the test continue running in the background. There will be too many test that will fail a subsequent test after that.
So for new code, I suggest to deprecate TimeOutAttribute and add a CancelAfterAttribute and the user needs to respect the passed in CancellationToken.

If that is acceptable for the @nunit/framework-team, I can update this PR to introduce that feature.

stevenaw:
> We cannot "fix" timeout attribute for .NET Core as there is no Thread.Abort to stop any runaway test.
Personally, I don't like reporting an error and then let the test continue running in the background. There will be too many test that will fail a subsequent test after that.
So for new code, I suggest to deprecate TimeOutAttribute and add a CancelAfterAttribute and the user needs to respect the passed in CancellationToken.

@manfred-brands Thanks for following up on this. The lack of `Thread.Abort()` on .NETCore is a real issue. Perhaps it's just lack of familiarity on my part, but would having the Timeout command cancel the CTS not allow for cooperative cancellation if a test opts in to using the newly-exposed token (at least for long-running tests which use operations which support CTs, if the test has opted into use the provided one)? Or are you saying we can't fix it to "just work" without also requiring each test to opt in by handling the token?

I also wouldn't mind hearing what some of the other reviewers on the PR think.

goblinmaks:
@rprouse @jnm2 @mikkelbu @CharliePoole 
FYI

NickStrupat:
https://github.com/nunit/nunit/issues/1821

manfred-brands:
> I think I missed something; why does adding the `CancellationToken` depend on having a solution for the lack of `Thread.Abort` in .NET Core+?

Correct. Hence my suggestion for a `CancelAfterAttribute` iso using the existing `TimeoutAttribute`
But we have not heard anything from the @nunit/framework-team if this is acceptable.
I will update the PR to only add the new attribute and functionality without affecting the existing timeout behaviour.
Then end-user can switch to the new cooperative cancellation if they want to.

stevenaw:
@manfred-brands Just doing a bit of cleanup here. I'm closing this as superseded by https://github.com/nunit/nunit/pull/4386 based on your comment in https://github.com/nunit/nunit/issues/4447#issuecomment-1661516745 but feel free to reopen this if you wish

manfred-brands:
Superseded by #4386

PR #4031, opened 2022-01-19
Added support for non-enumerable collections.

### LINKED ISSUE #4030: IsEmpty doesn't work with new .NET6 PriorityQueue
The PriorityQueue class has a `Count` property, but is not `IEnumerable`.

The `EmptyConstraint` should verify that `actual is IEnumerable` instead of assuming it.
If not `IEnumerable` it can check for a `Count` property.

manfred-brands:
Fixes #4030

Allows for 'collections' with a .Count property, such as the .net 6 PriorityQueue.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CountZeroConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CountZeroConstraint.cs
@@ -42,11 +42,11 @@ namespace NUnit.Framework.Constraints
         }
 
         /// <summary>
-        /// Checks if the specified <paramref name="type"/> has a Count property.
+        /// Checks if the specified <paramref name="type"/> has a int Count property.
         /// </summary>
         /// <param name="type">Type to check.</param>
-        /// <returns><see langword="true"/> when <paramref name="type"/> has a 'Count' property, <see langword="false"/> otherwise.</returns>
-        public static bool HasCountProperty(Type type) => type.GetProperty(CountPropertyName) != null;
+        /// <returns><see langword="true"/> when <paramref name="type"/> has a 'int Count' property, <see langword="false"/> otherwise.</returns>
+        public static bool HasCountProperty(Type type) => type.GetProperty(CountPropertyName)?.PropertyType == typeof(int);
 
         /// <summary>
         /// Test whether the constraint is satisfied by a given value
--- a/src/NUnitFramework/framework/Constraints/CountZeroConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CountZeroConstraint.cs
@@ -1,5 +1,5 @@
 // ***********************************************************************
-// Copyright (c) 2007 Charlie Poole, Rob Prouse
+// Copyright (c) 2020 Charlie Poole, Rob Prouse
 //
 // Permission is hereby granted, free of charge, to any person obtaining
 // a copy of this software and associated documentation files (the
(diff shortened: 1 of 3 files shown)


src/NUnitFramework/framework/Constraints/CountZeroConstraint.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            get { return "<empty>"; }
+        }
+
+        /// <summary>
+        /// Checks if the specified <paramref name="type"/> has a Count property.
+        /// </summary>
+        /// <param name="type">Type to check.</param>
+        /// <returns><see langword="true"/> when <paramref name="type"/> has a 'Count' property, <see langword="false"/> otherwise.</returns>
+        public static bool HasCountProperty(Type type) => type.GetProperty(CountPropertyName) != null;
```
mikkelbu, line 49: Should we also validate if the type of the property is an `int`. Otherwise I think the cast `ApplyTo` would fail.

mikkelbu (approved):
Only a single comment. Otherwise it looks good 👍

src/NUnitFramework/framework/Constraints/EmptyConstraint.cs
```diff
@@ -51,18 +51,20 @@ public override string Description
...
-                throw new System.ArgumentException($"The actual value must be a string, Guid, non-null IEnumerable or DirectoryInfo. The value passed was of type {actualType}.", nameof(actual));
             else if (actual is System.IO.DirectoryInfo)
                 realConstraint = new EmptyDirectoryConstraint();
-            else
+            else if (actual is System.Collections.IEnumerable)
                 realConstraint = new EmptyCollectionConstraint();
+            else if (CountZeroConstraint.HasCountProperty(actualType))
+                realConstraint = new CountZeroConstraint();
+            else
+                throw new ArgumentException($"The actual value must be a string, Guid, have a Count property, non-null IEnumerable or DirectoryInfo. The value passed was of type {actualType}.", nameof(actual));
```
stevenaw, line 67: Thanks for tackling this @manfred-brands . I recently hit this myself with `NameValueCollection` in a netfx project.

If I'm reading the diff right, this should better advise on future edge cases where a non-null yet incompatible type is passed in as it should avoid a cryptic NRE.

manfred-brands: Yes, previously if always assumed IEnumerble and that resulted in a runtime exception when casting later.

src/NUnitFramework/framework/Constraints/CountZeroConstraint.cs
manfred-brands: Added a check on `PropertyType` and added new unit test. Initially I contemplated allowing more types, but I don't think it is worth it.

src/NUnitFramework/framework/Constraints/EmptyConstraint.cs
manfred-brands: Added test case with `NameValueCollection`.

manfred-brands:
I see two approvals, but the PR is not merged, it anything holding this up?

mikkelbu:
Thanks for the ping @manfred-brands . I have been too busy to keep track of the issues and PRs lately. I'll merge it now. Can you also provide a PR for master?

stevenaw:
Thanks @mikkelbu 
Sorry @manfred-brands I had to step away from my computer before merging after my approval. Thanks for following up 👍

PR #4033, opened 2022-01-20
Recognized private members in base class for Source attributes

### LINKED ISSUE #4032: Tests won't run with an abstract base class that has a TestCaseFixtureSource
When reporting a bug, please provide the following information to speed up triage:

* NUnit and NUnit3TestAdapter versions
NUnit 3.12.2 NUnit3TestAdapter 4.2.0

* A short repro, preferably attached or pointing to a git repo or gist
[Repo Here](https://github.com/thomhurst/FailingAbstractTestCaseFixtureSourceNUnitTestProject) - Can't run tests when inheriting from an abstract base class that has a TestCaseFixtureSource

* What .net platform and version is being targeted
.NET 6

image

manfred-brands:
Fixes #4032

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
@@ -198,8 +198,8 @@ namespace NUnit.Framework
             if (SourceName == null)
                 return Reflect.Construct(sourceType) as IEnumerable;
 
-            MemberInfo[] members = sourceType.GetMember(SourceName,
-                    BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance | BindingFlags.FlattenHierarchy);
+            MemberInfo[] members = sourceType.GetMemberIncludingFromBase(SourceName,
+                    BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance);
 
             if (members.Length == 1)
             {
--- a/src/NUnitFramework/framework/Attributes/TypeGetMemberExtension.cs
+++ b/src/NUnitFramework/framework/Attributes/TypeGetMemberExtension.cs
@@ -1,0 +1,30 @@
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace NUnit.Framework
+{
+    internal static class TypeGetMemberExtension
+    {
+        /// <summary>
+        /// Searches for the specified members, using the specified binding constraints.
+        /// </summary>
+        /// <remarks>
+        /// Similar to <see cref="Type.GetMember(string, BindingFlags)"/> but also finds private static members from the base clas.
+        /// </remarks>
+        /// <param name="type">The type to find the member from.</param>
+        /// <param name="name">The string containing the name of the members to get.</param>
+        /// <param name="flags">A bitwise combination of the enumeration values that specify how the search is conducted.</param>
+        /// <returns></returns>
+        public static MemberInfo[] GetMemberIncludingFromBase(this Type type, string name, BindingFlags flags)
+        {
+            var members = new List<MemberInfo>();
+            do
+            {
+                members.AddRange(type.GetMember(name, flags));
+            }
+            while ((type = type.BaseType) != null);
+            return members.ToArray();
+        }
+    }
+}
--- a/src/NUnitFramework/tests/Attributes/TestFixtureSourceTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TestFixtureSourceTests.cs
@@ -44,9 +44,11 @@ namespace NUnit.Framework.Attributes
         [TestCase(typeof(SourceReturnsFixtureParameters))]
         [TestCase(typeof(ExtraTestFixtureAttributeIsIgnored))]
         [TestCase(typeof(TestFixtureMayUseMultipleSourceAttributes))]
+        [TestCase(typeof(DerivedClassUsingBaseClassDataSource))]
         public void CheckArgument(Type type)
         {
             var result = TestBuilder.RunTestFixture(type);
+            Assert.That(result.HasChildren, Is.True);
             Assert.That(result.ResultState, Is.EqualTo(ResultState.Success));
         }
 
--- a/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
@@ -243,8 +243,8 @@ namespace NUnit.Framework
             if (SourceName == null)
                 return Reflect.Construct(sourceType, null) as IEnumerable;
 
-            MemberInfo[] members = sourceType.GetMember(SourceName,
-                    BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance | BindingFlags.FlattenHierarchy);
+            MemberInfo[] members = sourceType.GetMemberIncludingFromBase(SourceName,
+                    BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance);
 
             if (members.Length == 1)
             {
--- a/src/NUnitFramework/framework/Attributes/ValueSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/ValueSourceAttribute.cs
@@ -103,8 +103,8 @@ namespace NUnit.Framework
                     ?? throw new InvalidDataSourceException($"The value source type '{sourceType}' does not implement IEnumerable.");
             }
 
-            MemberInfo[] members = sourceType.GetMember(SourceName,
-                BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance | BindingFlags.FlattenHierarchy);
+            MemberInfo[] members = sourceType.GetMemberIncludingFromBase(SourceName,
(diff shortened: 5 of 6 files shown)


mikkelbu (approved):
This of course changes the behaviour a little, but I don't think that this change will break anyone, as I think very few will have a method/field/property in the base class with the same name as the source.
@nunit/framework-team do you agree?

stevenaw:
> This of course changes the behaviour a little, but I don't think that this change will break anyone, as I think very few will have a method/field/property in the base class with the same name as the source. @nunit/framework-team do you agree?

I'm a little hesitant about this myself. While it seems like a simple fix to look for private members, I'm concerned it may also introduce a lot of unexpected edge cases. For example, what if a base class and a child class each define identically named methods with identical arguments? Non-private members would have this caught by the C# compiler as an error but the guard rails are off if we try to go around that.

I'm a little rusty on how/if private members can be used in other inheritance situations (like a virtual test method which gets overridden by a child class) but I see there's a bit of a discussion in the originating issue that this works with protected or public members and that perhaps an analyzer could help guide around this. I'm wondering if a in-framework fix could instead provide a more helpful error message if no public or protected members could be found.

manfred-brands:
>  For example, what if a base class and a child class each define identically named methods with identical arguments? 

In the previous case, the `GetMember` would have passed the child class parameters to both the child and the base class and never the base class parameter to the base class test. Most likely causing the base class test to fail.
So we have no real 'hiding' of naming in base classes. It is more like the child class can 'override' a `static` member.

This is actually used in various NUnit Framework test where the base class (`ConstraintTestBase`) depends on a `SuccessData` source which it doesn't define itself. The child class (`EmptyConstraintTest`) defines this, but the compiler complains it isn't used.
Looks like a candidate for [abstract static members](https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/static-abstract-interface-methods)

In the changes for this PR it would test nothing :-( as `GetMemberIncludingFromBase` returns 2 members and that is translated to `null` and silently ignored! Personally I would prefer an error if the requested data source cannot be (uniquely) found.

To return previous behaviour, I will update the `GetMemberIncludingFromBase` method to bail if it finds a member. So that if the member is defined in the child class it will use that regardless of what is defined in the base class.

stevenaw:
> This is actually used in various NUnit Framework test where the base class (ConstraintTestBase) depends on a SuccessData source which it doesn't define itself.

Hm, you are quite right, I hadn't thought about it this way. I suppose this is mostly to facilitate using `nameof()`, which those tests predate.

It still "feels" more right to me to expect a member which is declared in a base class and potentially defined or overridden in a child class to require "protected" or "public" visibility much the way the C# compiler enforces member visibility, but I can also see it being a bit awkward when the resolution (via `nameof()`) and invocation (via the test runner) don't both occur at compile-time.

I still think we may want to improve error messages to guide a user to use protected visibility over private when using `nameof()` in an abstract class, but perhaps that can just be done via an analyzer (I think @mikkelbu has already made the corresponding GH issue). Automagically resolving private members here definitely seems like a nice usability feature.

Thanks @manfred-brands (and thanks for the latest changes to help ensure this is non-breaking)

manfred-brands:
> It still "feels" more right to me to expect a member which is declared in a base class and potentially defined or overridden in a child class to require "protected" or "public" visibility.

There are two ways of 'overriding' a `static` member.
# If the member is `private`, the derived class can have a member with the same name without getting any warnings.
# If the member is `protected`, the derived class must use the `new` keyword to replace it.

The initial ticket has both the attribute and the member applied to the base class only. The fact is hidden from the child class which doesn't need to know. Therefore `private` accessibility sounds right. But NUnit operates on the child class and will find the member in the derived class first which might not be what is wanted.

Thinking more about this, I now tend to agree with you. If the member would be `protected` then a developer making a child class would get a warning if it accidentally uses the same name for a member. If deliberate, they can use the `new` keyword to state that intention. We do need to verify the latter works.

> I still think we may want to improve error messages to guide a user to use protected visibility over private when using nameof() in an abstract class.

I agree that we would need error messages, both analyzer and runtime, if the child class has a member with the same name as a private member in a base class if such a member is used in any of the Source attributes.

If nunit itself would use nunit.analyzer (which has been discussed), then they would have a warning in `ConstraintTestBase` that the `"SuccessData"` string passed to the `TestCaseSource` doesn't exist, unless we don't warn on `abstract` classes and only check the child which would be ideal as we otherwise have no way to enforce the child class defining the member.

What is the underlying reason for the `if (members.Length == 1)` tests in the `...SourceAttribute` classes. We neither get an error if the member is not defined nor if it defined twice. Both in my opinion are errors.

stevenaw:
Ah, yes, you're quite right. Somehow I had missed the "static" part of this scenario. You're absolutely right. Sorry about that @manfred-brands

manfred-brands:
@stevenaw I was not shouting, I thought I was numbering, but they became big bold headings. :-(

stevenaw:
Not to worry, no offense taken at all @manfred-brands . I hadn't realized when making my earlier comments that we had also been considering static overriding, but of course your suggestions make sense in that context, and I am glad the conversation was steered back in that direction 🙂

src/NUnitFramework/tests/Attributes/TestFixtureSourceTests.cs
```diff
@@ -44,9 +44,12 @@ public class TestFixtureSourceTests
         [TestCase(typeof(SourceReturnsFixtureParameters))]
         [TestCase(typeof(ExtraTestFixtureAttributeIsIgnored))]
         [TestCase(typeof(TestFixtureMayUseMultipleSourceAttributes))]
+        [TestCase(typeof(DerivedClassUsingBaseClassDataSource))]
+        [TestCase(typeof(BaseClassUsingDerivedClassDataSource))]
```
stevenaw, line 48: 👍 Thanks for the thorough test cases!

src/NUnitFramework/framework/Attributes/TypeGetMemberExtension.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace NUnit.Framework
+{
+    internal static class TypeGetMemberExtension
```
stevenaw, line 7: Thanks for making these extensions internal visibility by default.

Suggestion: I think there may be an existing + recently created [`TypeExtensions`](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/Extensions/TypeExtensions.cs) class which lives in `Internal/Extensions`. Would it make sense for this new method to live there too?

src/NUnitFramework/framework/Attributes/TypeGetMemberExtension.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <param name="flags">A bitwise combination of the enumeration values that specify how the search is conducted.</param>
+        /// <returns></returns>
+        public static MemberInfo[] GetMemberIncludingFromBase(this Type type, string name, BindingFlags flags)
+        {
+            var members = new List<MemberInfo>();
+            do
+            {
+                members.AddRange(type.GetMember(name, flags));
+            }
+            while (members.Count == 0 && (type = type.BaseType) != null);
```
stevenaw, line 26: Suggestion:  `NUnit.Framework.Internal.Reflect` also defines a few extension methods over the `Type` type, including this "walking the class hierarchy" behaviour. Would it make sense to use [`TypeAndBaseTypes`](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/Reflect.cs#L291-L297) here instead?

I know this part changed recently, but if I'm reading the current version of the code right, we also only walk the hierarchy until the first time that `GetMember` returns something. Coupled with the above suggestion, would it be simpler to change this from a dowhile to a simple foreach + return like the below pseudocode?

```
foreach currentType in type.TypeAndBaseTypes()
  members = type.GetMember(name, flags)
  if members.Count != 0
    return members
```

src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
```diff
@@ -198,8 +198,8 @@ public IEnumerable<ITestFixtureData> GetParametersFor(Type sourceType)
             if (SourceName == null)
                 return Reflect.Construct(sourceType) as IEnumerable;
 
-            MemberInfo[] members = sourceType.GetMember(SourceName,
-                    BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Static | BindingFlags.Instance | BindingFlags.FlattenHierarchy);
```
stevenaw, line 202: Thanks for catching this. It's a little awkward for our purposes that `FlattenHierarchy` doesn't handle private members

stevenaw (changes requested):
Thanks @manfred-brands ! Everything looks great. I left a few suggestions for a few existing classes or extensions which could help what you're trying to do here.

src/NUnitFramework/tests/Attributes/TestFixtureSourceTests.cs
manfred-brands: Test Driven Development. First create a test case that shows the problem, then fix it.

src/NUnitFramework/framework/Attributes/TypeGetMemberExtension.cs
manfred-brands: I had looked for extension classes, but missed this one. Method moved and Tests added.

stevenaw (approved):

stevenaw:
LGTM! Thanks for your contribution @manfred-brands 👍

Issue #4009, opened 2021-11-29
Performance degradation on many tests


timcassell:
I've been noticing performance degradation when running many tests (thousands in my project) for a while. It's especially noticeable on my long-running tests, but the test times also show degradation on quick tests also.

I have setup and teardowns for every test.

At first I thought it was because I have objects pooling and the pool could be causing a leak and interfering with the performance, but I still see the same degradation with the pool disabled.

![image](https://user-images.githubusercontent.com/35501420/143837499-37054505-371d-4537-a9b3-5ed2127b89de.png)

![image](https://user-images.githubusercontent.com/35501420/143837063-6ffd6f9c-5549-4e8b-b417-57f31a7af45a.png)

When I run all tests in a group, the last sub-group takes `1.4s`. Then when I run just the last sub-group independently, it takes `480ms`, nearly 1/3 of the time. The more tests I run at a time, the longer it takes the last test to complete. This is a big problem for me with over 3600 tests, half of them are long-running (the images are only showing the quick tests).

I was using NUnit 3.13.0, and I just updated to NUnit 3.13.2 and saw the same behavior on both versions.

Any ideas?

[Edit] I should add that my library is targeting netstandard2.0, and I'm running the tests in net472 and net5.0 (I see the same degradation in both runtimes).

mikkelbu:
@timcassell Do you see the same degradation when you run the tests using the NUnit console - https://github.com/nunit/nunit-console/releases/tag/v3.12 - and/or NUnitLite - see e.g. https://docs.nunit.org/articles/nunit/running-tests/NUnitLite-Runner.html

When you are running the tests in VS you are also depending on VS, VSTest, and the NUnit 3 Test Adapter, so the above suggestion is to determine where the problem is. Note that both the NUnit console and the NUnit 3 Test Adapter uses the NUnit engine.

timcassell:
@mikkelbu Interestingly, I got the opposite effect with the nunit3-console. Though I could only run the net472 tests with it, not net5.0.

```
nunit3-console ProtoPromiseTests.dll --where "namespace == ProtoPromiseTests.APIs"

SequenceTests duration="0.539729"
```
```
nunit3-console ProtoPromiseTests.dll --where "class = ProtoPromiseTests.APIs.SequenceTests"

SequenceTests duration="0.606604"
```

And I'm not sure if the nunit3-console executes the tests in the same order as the VS test runner...

lahma:
[...]

Issue #3601, opened 2020-07-22
Nunit 3.10.1.0 + ReSharper 2020.1.4: Test execution delayed by ~60seconds 


h4kun4m4t4t4:
Hi,

first of all, I'm not sure whether NUnit or Resharper might be the problem. (But problems started after upgrading to latest resharper)

When I start the execution of a test via resharper test runner, the test shows pending for about 60 seconds before even starting the relevant test code. I tracked it down to be a loading/performance issue at **RuntimeFramework.cs** 
`Type monoTouchType = Type.GetType("MonoTouch.UIKit.UIApplicationDelegate,monotouch");`
The call to the above takes about 60 seconds and will then continue normal test execution.

Does anyone know what the problem might be or how I can track it further down?

What I did so far:

- Execution with MS Test and NUnitTestadapter: works fine

- setup a new Solution: no problem of test execution

- When at the above position in debugger -> execute at immediate window (without the "**,monotouch**"):  `Type.GetType("MonoTouch.UIKit.UIApplicationDelegate");` --> returns immediately

CharliePoole:
The key bit you have left out is what kind of an application you are targeting. :smiley: That line of code will execute differently depending on whether you are using MonoTouch or not.

To be more precise, it will execute differently depending on whether an assembly containing that UI delegate is loaded or not.

h4kun4m4t4t4:
Thanks for the info.
The application is written in C# and targeting .Net Framework 4.7. 
In fact in our company we have different Solutions with the above settings... some show the delay and some not.

I did non hear about MonoTouch until yesterday, so I do not think it is involved at all. The call always returns "Null".

In the meantime I believe it might be something how the resharper testrunner orchestrates NUnit. Maybe it is something in the environment. But currently I'm lost and do not know what to look for.

ChrisMaddock:
I suspect a PR I did 2 years ago is the reason for the change between 3.10 and 3.12 --> https://github.com/nunit/nunit/pull/3131

I have no idea what might be the cause of your current issue, I'm afraid.

h4kun4m4t4t4:
Morning, 
I'm currently back to my old setup **Nunit 3.10.1.0 + ReSharper 2019.2.4**. This works fine so far.

Last Friday I was experimenting and tried to create a simple test project, to be able to reproduce the problem.... It is not that easy :-/.
In the end I took one of my company projects and stripped down all unnecessary code parts. 
[...]

PR #4034, opened 2022-01-25
Improve method discovery and filtering performance

lahma:
First of all, sorry for not pinging beforehand and discussing with you about the change. I got a bit carried away while profiling and worked with the fixes as I went on.

## TLDR;

80k parametrized test case suite (loading suite, counting test cases) down from almost 11 minutes to less than one second when using Rider. So for this specific case it's __over 700x improvement__.

## Backstory

I'm generating test source code for running [ECMAScript test suite](https://github.com/tc39/test262). This test suite is heavy with 86k+ test cases (about 40k files which are tested in both strict and non-strict mode) and they are ideal use case for using `TestCaseAttribute` defining the files for each method (where method is usually a JavaScript target like `Array.prototype.filter`, having many test cases, e.g. files). 

## Problem

I got the generation working and then started to run them with Rider... or actually I didn't as the whole unit testing sub-system froze when I tried to run the full test suite. I could run small subsets without the IDE's unit testing session hanging in waiting state. The pending to start state before tests would actually run could take minutes. Seems that Rider has different approach than VS (based on earlier issue it sends more like full names) to sending tests - Rider uses OR with all the ids - 80k ids that is. Which will cause the inclusion check algorithm to loop with a large set of data using O(n^2) .

[Here's example of generated test case](https://gist.github.com/lahma/ec546e690aebfed4d186f3a02c80b037) and how I then later profiled NUnit with benchmark test case run in console, using data from Rider invocation (`FrameworkController` parameters). I tweaked it a bit on the way, but results are calculated using the same code.

It seems that it's enough to have a test case such as below to make Rider freeze with unit testing operations. This should mimic the situation quite well be creating the 86 test cases we are testing here.

```csharp
public class TestCaseTest
{
    [Test]
    public void Test([Range(1, 86_000)] int i)
    {
    }
}
```

## Results before

Processing 86607 test cases. Loading them and then discovery without filter and then counting with id filter containing every id found. Ryzen 5950X with 32GB RAM. Almost 11 minutes to load tests and go trough them using a OR'ed filter. ExploreTests is using empty filter in Rider's call chain.

```
LoadTests took: 00:00:27.5665738 // total 86k tests, mostly from TestCase attributes
ExploreTests took: 00:00:00.0037093 // empty filter
CountTests took: 00:10:09.3055713 // 86k OR Id filter items - yes, 10 MINUTES
```

## Fixes

Fixes consist of two parts:

### Loading tests: reducing memory allocations and re-calculations when building test cases

It's very costly to do reflection and inspecting same targets multiple times. For a method having 20 test case attributes all introspection against the test method would be done 20 times, when only parameters for the test case in construction will vary.

Now caching static information about `IMethodInfo` for subsequent accesses.

### Filtering tests: combine `OR` filter with same-kind, non-regexp items to single hash-based InFilter

This builds on the earlier https://github.com/nunit/nunit/issues/3786 and its PR https://github.com/nunit/nunit/pull/3787 . Now detecting more filter types and target fields and creating efficient hash set for that case. Created new `InFilter` type dedicated to matching.

/cc @pakrym 

## Results after

```
LoadTests took: 00:00:00.6497215 // total 86k tests, mostly from TestCase attributes
ExploreTests took: 00:00:00.0030948 // empty filter
CountTests took: 00:00:00.1815572 // 86k OR Id filter items
```

/cc @AndreyAkinshin for Rider performance in Unit testing (VS has different filter logic for full test suite I guess)

Related issues: https://github.com/nunit/nunit/issues/3601 https://github.com/nunit/nunit/issues/3786 https://github.com/nunit/nunit/issues/4009 https://github.com/nunit/nunit3-vs-adapter/issues/660

### DIFF
--- a/src/NUnitFramework/framework/Internal/AttributeProviderWrapper.cs
+++ b/src/NUnitFramework/framework/Internal/AttributeProviderWrapper.cs
@@ -23,12 +23,11 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Reflection;
 
 namespace NUnit.Framework.Internal
 {
-    internal class AttributeProviderWrapper<T> : ICustomAttributeProvider
+    internal sealed class AttributeProviderWrapper<T> : ICustomAttributeProvider
         where T : Attribute
     {
         private readonly ICustomAttributeProvider _innerProvider;
@@ -52,14 +51,27 @@ namespace NUnit.Framework.Internal
 
         public bool IsDefined(Type attributeType, bool inherit)
         {
-            return GetCustomAttributes(attributeType, inherit).Any();
+            return GetCustomAttributes(attributeType, inherit).Length > 0;
         }
 
-        private static T[] GetFiltered(IEnumerable<object> attributes)
+        private static T[] GetFiltered(object[] attributes)
         {
-            return attributes
-                   .OfType<T>()
-                   .ToArray();
+            List<T> filtered = null;
+            foreach (var attribute in attributes)
+            {
+                if (attribute is T t)
+                {
+                    filtered ??= new List<T>();
+                    filtered.Add(t);
+                }
+            }
+
+            return filtered?.ToArray() ??
+#if NETSTANDARD2_0_OR_GREATER
+                Array.Empty<T>();
+#else
+                new T[0];
+#endif
         }
     }
 }
--- a/src/NUnitFramework/framework/Internal/Builders/MethodInfoCache.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/MethodInfoCache.cs
@@ -1,0 +1,62 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using NUnit.Framework.Interfaces;
+
+namespace NUnit.Framework.Internal.Builders
+{
+    /// <summary>
+    /// Caches static information for IMethodInfo to reduce re-calculations and memory allocations from reflection.
+    /// </summary>
+    internal static class MethodInfoCache
+    {
+        // we would otherwise do a lot of attribute allocations when repeatedly checking the same method for example
+        // in case of building TestCaseAttribute-based tests
+#if NET35
+        private static readonly Dictionary<IMethodInfo, TestMethodMetadata> MethodMetadataCache = new Dictionary<IMethodInfo, TestMethodMetadata>();
+#else
+        private static readonly ConcurrentDictionary<IMethodInfo, TestMethodMetadata> MethodMetadataCache = new ConcurrentDictionary<IMethodInfo, TestMethodMetadata>();
+#endif
+
+        /// <summary>
+        /// Returns cached metadata for method instance.
+        /// </summary>
+        internal static TestMethodMetadata Get(IMethodInfo method)
+        {
+#if NET35
+            lock (MethodMetadataCache)
+            {
+                if (!MethodMetadataCache.TryGetValue(method, out var metadata))
+                    MethodMetadataCache.Add(method, metadata = new TestMethodMetadata(method));
+
+                return metadata;
+            }
+#else
+            return MethodMetadataCache.GetOrAdd(method, m => new TestMethodMetadata(m));
+#endif
+        }
+
+        /// <summary>
+        /// Memoization of TestMethod information to reduce subsequent allocations from parameter and attribute information.
+        /// </summary>
+        internal sealed class TestMethodMetadata
+        {
+            public TestMethodMetadata(IMethodInfo method)
+            {
+                Parameters = method.GetParameters();
+                IsAsyncOperation = AsyncToSyncAdapter.IsAsyncOperation(method.MethodInfo);
+                IsVoidOrUnit = Reflect.IsVoidOrUnit(method.ReturnType.Type);
+
+                RepeatTestAttributes = method.GetCustomAttributes<IRepeatTest>(true);
+                TestBuilderAttributes = method.GetCustomAttributes<ITestBuilder>(false);
+            }
+
+            public IParameterInfo[] Parameters { get; }
+            public IRepeatTest[] RepeatTestAttributes { get; }
+            public bool IsAsyncOperation { get; }
+            public bool IsVoidOrUnit { get; }
+            public ITestBuilder[] TestBuilderAttributes { get; }
+        }
+    }
+}
--- a/src/NUnitFramework/framework/Internal/Builders/MethodInfoCache.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/MethodInfoCache.cs
@@ -48,15 +48,29 @@ namespace NUnit.Framework.Internal.Builders
                 IsAsyncOperation = AsyncToSyncAdapter.IsAsyncOperation(method.MethodInfo);
                 IsVoidOrUnit = Reflect.IsVoidOrUnit(method.ReturnType.Type);
 
+                // TODO could probably go trough inherited and non inherited in two passes instead of multiple
+
+                // inherited
                 RepeatTestAttributes = method.GetCustomAttributes<IRepeatTest>(true);
+                WrapTestMethodAttributes = method.GetCustomAttributes<IWrapTestMethod>(true);
+                WrapSetupTearDownAttributes = method.GetCustomAttributes<IWrapSetUpTearDown>(true);
+                ApplyToContextAttributes = method.GetCustomAttributes<IApplyToContext>(true);
+
+                // non-inherited
                 TestBuilderAttributes = method.GetCustomAttributes<ITestBuilder>(false);
+                TestActionAttributes = method.GetCustomAttributes<ITestAction>(false);
             }
 
             public IParameterInfo[] Parameters { get; }
-            public IRepeatTest[] RepeatTestAttributes { get; }
             public bool IsAsyncOperation { get; }
             public bool IsVoidOrUnit { get; }
+
+            public IRepeatTest[] RepeatTestAttributes { get; }
             public ITestBuilder[] TestBuilderAttributes { get; }
+            public IWrapTestMethod[] WrapTestMethodAttributes { get; }
+            public ITestAction[] TestActionAttributes { get; }
+            public IWrapSetUpTearDown[] WrapSetupTearDownAttributes { get; }
+            public IApplyToContext[] ApplyToContextAttributes { get; }
         }
     }
 }
--- a/src/NUnitFramework/framework/Internal/Filters/CompositeFilter.cs
+++ b/src/NUnitFramework/framework/Internal/Filters/CompositeFilter.cs
@@ -21,7 +21,6 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using NUnit.Framework.Interfaces;
--- a/src/NUnitFramework/framework/Internal/Filters/InFilter.cs
+++ b/src/NUnitFramework/framework/Internal/Filters/InFilter.cs
@@ -16,16 +16,19 @@ namespace NUnit.Framework.Internal.Filters
     {
         private readonly Func<ITest, string?> _selector;
         private readonly HashSet<string?> _values;
+        private readonly string _xmlElementName;
 
         /// <summary>
         /// Constructs new InFilter.
         /// </summary>
         /// <param name="selector">Selector to get value to check.</param>
         /// <param name="values">Valid values for the filter.</param>
-        public InFilter(Func<ITest, string?> selector, IEnumerable<string?> values)
+        /// <param name="xmlElementName">The XML element name the original filter used.</param>
+        private InFilter(Func<ITest, string?> selector, IEnumerable<string?> values, string xmlElementName)
         {
             _selector = selector;
             _values = new HashSet<string?>(values);
+            _xmlElementName = xmlElementName;
         }
 
         /// <summary>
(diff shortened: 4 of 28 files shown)


src/NUnitFramework/framework/Internal/Builders/MethodInfoCache.cs
```diff
@@ -0,0 +1,62 @@
...
+
+                return metadata;
+            }
+#else
+            return MethodMetadataCache.GetOrAdd(method, m => new TestMethodMetadata(m));
+#endif
+        }
+
+        /// <summary>
+        /// Memoization of TestMethod information to reduce subsequent allocations from parameter and attribute information.
```
ChrisMaddock, line 41:
```suggestion
        /// Memorization of TestMethod information to reduce subsequent allocations from parameter and attribute information.
```

ChrisMaddock (commented):
The performance numbers intrigued me on this one so I had a look through. This looks brilliant!

I couldn't find any problem with the code here, and the numbers are really impressive. The only thing that jumped out to me was where this would benefit from some additional testing around the new logic in `TestFilter.FromXml` for when an InFilter should and shouldn't be built - more to protect against future code changes in this area. (I wonder if it may also be worth breaking that logic out into a separate method?)

Looks like a great improvement @lahma! 👍

src/NUnitFramework/framework/Internal/Builders/MethodInfoCache.cs
lahma: I think I got it right https://en.wikipedia.org/wiki/Memoization , but I can change if you want to 🙂

lahma:
I'll improve test coverage and check if breaking into separate method is easy.

src/NUnitFramework/framework/Internal/AttributeProviderWrapper.cs
```diff
@@ -52,14 +51,27 @@ public object[] GetCustomAttributes(bool inherit)
...
+                    filtered.Add(t);
+                }
+            }
+
+            return filtered?.ToArray() ??
+#if NETSTANDARD2_0_OR_GREATER
+                Array.Empty<T>();
+#else
+                new T[0];
+#endif
```
siprbaum, line 74: When glancing over the code I really got confused  about this change, because from a readibility point of view it is definitely worse than before 😞 Took me quite a few brain cycles to see that it just replaces the linq usage and that it is correct (which it is, as far as I can tell!).

So at least I would have expected a seperate commit doing this "get rid of linq" change,  accompanied with the motivating performance numbers in the commit message telling WHY this  decrease in readability is needed, e.g. because it improves the performance. Would make it easier to swallow that bitter pill here ;-)
Right now it is embedded in the "Cache method metadata" commit, which this change above has nothing to do with.

Please take this comment with a grain of salt, as this is just a "drive by review", as I am not a contributer to this project and therefore I don't think my opinion really matters here...

lahma: Fair enough, optimizations tend to get a bit verbose and harder to read. "Cache method metadata" might not tell the store, but I put it under the "Reduce allocations" sub-heading under PR 😉 So it was one of the things that contributed to performance improvements, I should have included the rough allocation numbers in hindsight though.

This method was called quite a lot and it usually returned empty array at the end.  It was hot w.r.t to allocations so it was part of the cleanup for the biggest offenders. I might create a separate PR for other long-hanging fruit but this already got a bit out of hand... 

Thanks for the drive-by, all feedback is always appreciated!

lahma:
@ChrisMaddock Thank you for the review, I've addressed your feedback in latest commit (except for the grammar), by moving optimization step into `InFilter` nullables were introduced so I had to tweak signatures a bit.

src/NUnitFramework/framework/Internal/Builders/MethodInfoCache.cs
ChrisMaddock: Oops, sorry! New word to me, and my browsers spellchecker! 😆

ChrisMaddock:
Nice refactor! 👍 That structure looks all a bit cleaner to me, I like it!

lahma:
So it's been a month, all horrible things going on aside, is there something I can do to push this forward?

stevenaw:
Sorry it's taken so long to get to this @lahma 
We usually like to have two sets of reviews for changes. I hope to be able to get to this myself this weekend or sometime next week

src/NUnitFramework/framework/Internal/Filters/InFilter.cs
```diff
@@ -0,0 +1,100 @@
...
+            // eagerly build expected value collection
+            var expectedValues = new List<string?>(orFilter.Filters.Count);
+
+            // use for for faster traversal without boxed enumerator
+            for (var i = 0; i < orFilter.Filters.Count; i++)
+            {
+                var filter = orFilter.Filters[i];
+
+                // make sure invariants are valid
+                commonType ??= filter.GetType();
```
mikkelbu, line 49: Would it make any difference if we do this initialization outside the loop ? E.g. remove this line and change the declaration to
```
Type? commonType = hashCommonTypeAndNonRegex ? orFilter.Filters[0] : null;
```

src/NUnitFramework/framework/Internal/Filters/InFilter.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        public override bool Match(ITest test)
+        {
+            return _values.Contains(_selector(test));
+        }
+
+        public override TNode AddToXml(TNode parentNode, bool recursive)
+        {
+            throw new NotImplementedException();
```
mikkelbu, line 97: I think this will give us problems in runners that also write the filters in the XML (e.g. NUnitLite, the console etc). I guess we could write the filter as a `OrFilter` (https://docs.nunit.org/articles/nunit/technical-notes/usage/Test-Filters.html), but then we need to store the original OrFilter also (or at least have more information).

src/NUnitFramework/tests/Internal/Filters/InFilterTests.cs
```diff
@@ -0,0 +1,83 @@
...
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+// ***********************************************************************
```
mikkelbu, line 22:
```suggestion
// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
```

mikkelbu (commented):
This looks very promissing. I think we need to handle the `NotImplementedException` in InFilter.cs. The rest of the comments are minor issues and questions/comments.

Great work @lahma

src/NUnitFramework/framework/Internal/AttributeProviderWrapper.cs
```diff
@@ -23,12 +23,11 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Reflection;
 
 namespace NUnit.Framework.Internal
 {
-    internal class AttributeProviderWrapper<T> : ICustomAttributeProvider
+    internal sealed class AttributeProviderWrapper<T> : ICustomAttributeProvider
```
stevenaw, line 30: Thanks for this 😄

src/NUnitFramework/framework/Internal/Filters/CompositeFilter.cs
```diff
@@ -21,7 +21,6 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-using System;
```
stevenaw, line 24: Thanks for doing opportunistic cleanup like this too

stevenaw (approved):
This looks great thanks! I asked a few questions but no real additional comments or concerns beyond @mikkelbu 's. Nice job.

src/NUnitFramework/tests/Internal/Filters/OrFilterTests.cs
```diff
@@ -241,7 +241,7 @@ public void BuildFromXmlFullName()
             TestFilter filter = TestFilter.FromXml(
                 $"<filter><or><test>{DUMMY_CLASS}</test><test>{ANOTHER_CLASS}</test></or></filter>");
 
-            Assert.That(filter, Is.TypeOf<OrFilter>());
+            Assert.That(filter, Is.TypeOf<InFilter>());
```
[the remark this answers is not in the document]
lahma: I moved this to `InFilterTests.cs`, not sure about create `OrFilter` part, it's `or` in XML so test case makes sure it's transformed?

src/NUnitFramework/framework/Internal/Builders/MethodInfoCache.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// </summary>
+        internal sealed class TestMethodMetadata
+        {
+            public TestMethodMetadata(IMethodInfo method)
+            {
+                Parameters = method.GetParameters();
+                IsAsyncOperation = AsyncToSyncAdapter.IsAsyncOperation(method.MethodInfo);
+                IsVoidOrUnit = Reflect.IsVoidOrUnit(method.ReturnType.Type);
+
+                // TODO could probably go trough inherited and non inherited in two passes instead of multiple
```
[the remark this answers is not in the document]
lahma: I think it could be a separate PR, the performance is the same compared to current, this is a not that current code (even before the PR) loops same collections multiple times, which is unnecessary, but it would complicate this PR, I think.

src/NUnitFramework/framework/Internal/Filters/InFilter.cs
lahma: That would change the semantics, the next line ensure that every filter in list has the same type, which is a requirement.

`hashCommonTypeAndNonRegex &= commonType == filter.GetType();`

src/NUnitFramework/framework/Internal/Filters/InFilter.cs
lahma: My thinking was that this is dynamic, e.g. translated runtime so it would not matter. I probably don't understand the full ramifications, should this be addressed in this PR, will this be breaking things?

src/NUnitFramework/framework/Internal/AttributeProviderWrapper.cs
```diff
@@ -52,14 +51,27 @@ public object[] GetCustomAttributes(bool inherit)
 
         public bool IsDefined(Type attributeType, bool inherit)
         {
-            return GetCustomAttributes(attributeType, inherit).Any();
+            return GetCustomAttributes(attributeType, inherit).Length > 0;
         }
 
-        private static T[] GetFiltered(IEnumerable<object> attributes)
+        private static T[] GetFiltered(object[] attributes)
```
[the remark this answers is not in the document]
lahma: Yes it's always a bit bad for perf to hide an array behind an interface. They never really are interface-based so a slower wrapper will be introduced. CLR is highly optimized when it sees an actual array. Not a matter when called a couple of times but when in thousands, it adds up.

src/NUnitFramework/framework/Internal/AttributeProviderWrapper.cs
lahma: Ideally all private/internal things should be sealed, it allows CLR some more room to optimize. This is what the coreclr/runtime did, made everything sealed they could. Might be a micro-optimization, but can be important to things like comparers/equality checkers.

src/NUnitFramework/framework/Internal/Filters/CompositeFilter.cs
lahma: I try to do this only for files I touch, this was a bit on the fence as the addition of `sealed` was not required for the functionality 🤷🏻‍♂️

lahma:
@stevenaw thank you for your review. I responded to comments, tried to improve the code based on suggestion, but leaving the question about `AddToXml` open. It also seems that CI is now broken due to using some obsolete build infrastructure 😬

stevenaw:
Thanks @lahma 
The build should now be fixed of you merge latest into your branch. I'm less familiar with the implications of the xml line of questions, so I think Mikkel would know better about that

lahma:
@stevenaw @mikkelbu I've rebased this branch against `v3.13-dev` and implemented the `AddToXml` including test coverage. Now `AddToXml` generates same XML as the original `OrFilter` that was transformed into `InFilter`.

src/NUnitFramework/tests/Internal/Filters/OrFilterTests.cs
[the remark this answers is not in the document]
mikkelbu: I think this is fine. The second bullet was a mistake on my part. I just wanted to be sure we had tests of XML => OrFilter, but this is covered by `BuildFromXml` and `BuildFromXmlFullNameRegex`.

src/NUnitFramework/framework/Internal/Filters/InFilter.cs
mikkelbu: Isn't the semantics the same?

In both cases `commonType` will be the type of the first element in `orFilter.Filters`. I was just wondering if it would make any difference to set `commonType` outside the loop rather than in each iteration checking if `commonType` is `null` and in that case assign it to the type of the current filter (meaning that this will only be set in the first iteration, but we will check in each iteration)

mikkelbu: This was mostly a question as I was curious whether it made any difference (my feeling is that it don't, so I don't think this should hold back this PR). Thanks for the great work @lahma :+1 and thanks for having patience with us :)

mikkelbu:
Thanks for the work @lahma :+1. I'll be merging it now.

lahma:
Thanks to everyone involved and helping out iterating the final version!

PR #4035, opened 2022-01-30
Reduce allocations caused by string.Split

lahma:
## Problem

When you have > 80k test cases mostly via `TestCaseAttribute`  and utilize categories for them, the category information processing (string splitting) is causing a lot of allocations. Below is a profile from a test run against the test suite. `IsAsyncOperation` was fixed in the other PR.

image

So about 15% of memory allocations are coming from this setter.

## Fix

I've created `StringTokenEnumerator` that allows foreach'ing against a string with given separator without separate array allocations. There's a convenience extension method to call this using string `myString.Tokenize(',')` . It's always faster compared to `string.Split` and doesn't allocate for empty or single category at all like the split does (string.Split has to create a new string array). Added benefit is that for empty content property bad adding isn't called either.

With these changes the `set_Category` vanishes from allocation traces altogether.

## Benchmark

```csharp
[MemoryDiagnoser]
public class SplitBenchmark
{
    [Params("","single", "category1,category2,category3")]
    public string Input { get; set; }

    [Benchmark]
    public int SplitWithString()
    {
        return Input.Split(new char[] { ',' }).Length;
    }

    [Benchmark]
    public int SplitWithEnumerator()
    {
        int count = 0;
        foreach (var item in new StringTokenEnumerator(Input, ','))
        {
            count++;
        }

        return count;
    }
}
```

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
.NET SDK=6.0.101
  [Host]     : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT
  DefaultJob : .NET 6.0.1 (6.0.121.56705), X64 RyuJIT


```
|              Method |                Input |      Mean |     Error |    StdDev |  Gen 0 | Allocated |
|-------------------- |--------------------- |----------:|----------:|----------:|-------:|----------:|
|     **SplitWithString** |                     **** |  **8.102 ns** | **0.0279 ns** | **0.0372 ns** | **0.0038** |      **64 B** |
| SplitWithEnumerator |                      |  2.675 ns | 0.0050 ns | 0.0047 ns |      - |         - |
|     **SplitWithString** | **categ(...)gory3 [29]** | **44.790 ns** | **0.2100 ns** | **0.1640 ns** | **0.0119** |     **200 B** |
| SplitWithEnumerator | categ(...)gory3 [29] | 43.417 ns | 0.4088 ns | 0.3413 ns | 0.0072 |     120 B |
|     **SplitWithString** |               **single** | **15.026 ns** | **0.0534 ns** | **0.0473 ns** | **0.0038** |      **64 B** |
| SplitWithEnumerator |               single |  9.322 ns | 0.0169 ns | 0.0159 ns |      - |         - |

### DIFF
--- a/src/NUnitFramework/framework/Internal/StringTokenEnumerator.cs
+++ b/src/NUnitFramework/framework/Internal/StringTokenEnumerator.cs
@@ -1,0 +1,122 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+namespace NUnit.Framework.Internal
+{
+    /// <summary>
+    /// Allows enumerating sub-strings from a string based on separator char.
+    /// Basically a less allocating version of string.Split.
+    /// </summary>
+    internal struct StringTokenEnumerator : IEnumerator<string>, IEnumerable
+    {
+        private int _currentStartIndex;
+        private readonly string _target;
+        private readonly char _separator;
+        private readonly bool _returnEmptyTokens;
+
+        /// <summary>
+        /// Constructs a new enumerator against given target with given separator.
+        /// </summary>
+        /// <param name="target">Target string to enumerate.</param>
+        /// <param name="separator">Separator between tokens to return.</param>
+        /// <param name="returnEmptyTokens">Whether empty tokens should be returned</param>
+        public StringTokenEnumerator(string target, char separator, bool returnEmptyTokens)
+        {
+            _currentStartIndex = 0;
+            _target = target;
+            _separator = separator;
+            _returnEmptyTokens = returnEmptyTokens;
+            Current = null;
+        }
+
+        /// <inheritdoc />
+        public bool MoveNext()
+        {
+            if (!_returnEmptyTokens && _target.Length == 0)
+            {
+                return false;
+            }
+
+            while (FindNextToken())
+            {
+                if (!_returnEmptyTokens && Current!.Length == 0)
+                {
+                    // check for the next one
+                    continue;
+                }
+
+                break;
+            }
+
+            return Current != null;
+        }
+
+        private bool FindNextToken()
+        {
+            int start = _currentStartIndex;
+
+            if (start > _target.Length)
+            {
+                // we've jumped to the end
+                Current = null;
+                return false;
+            }
+
+            var endIndex = _target.IndexOf(_separator, start);
+            if (endIndex == -1)
+            {
+                Current = _target.Substring(start);
+                // signal that we have no more to give, next loop will return false
+                _currentStartIndex = _target.Length + 1;
+                return true;
+            }
+
+            Current = _target.Substring(start, endIndex - start);
+            _currentStartIndex = endIndex + 1;
+            return true;
+        }
+
+        /// <inheritdoc />
+        public void Reset()
+        {
+            throw new NotSupportedException("Resetting enumerator is not supported");
+        }
+
+        /// <inheritdoc />
+        public void Dispose()
+        {
+        }
+
+        /// <inheritdoc />
+        public string Current { get; private set; }
+
+        object IEnumerator.Current => Current;
+
+        /// <summary>
+        /// Return unboxed enumerator when type is well known, do not remove.
+        /// </summary>
+        public StringTokenEnumerator GetEnumerator() => this;
+
+        IEnumerator IEnumerable.GetEnumerator() => this;
+    }
+
+    internal static class StringExtensions
+    {
+        /// <summary>
+        /// Tokenizes given string with given separator.
+        /// </summary>
+        /// <param name="s">Input string.</param>
+        /// <param name="separator">Separator to use.</param>
+        /// <param name="returnEmptyTokens">Whether to return empty token (length 0), defaults to false.</param>
+        /// <returns></returns>
+        [MethodImpl(256)] // aggressive inlining, not available for old frameworks via enum
+        internal static StringTokenEnumerator Tokenize(this string s, char separator, bool returnEmptyTokens = false)
+        {
+            return new StringTokenEnumerator(s, separator, returnEmptyTokens);
+        }
+    }
+}
--- a/src/NUnitFramework/tests/Internal/StringTokenEnumeratorTests.cs
+++ b/src/NUnitFramework/tests/Internal/StringTokenEnumeratorTests.cs
@@ -1,0 +1,59 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using System;
+using System.Collections.Generic;
+
+namespace NUnit.Framework.Internal
+{
+    public class StringTokenEnumeratorTests
+    {
+        [Test]
+        public void Empty()
+        {
+            var tokens = Enumerate("", ',');
+            Assert.AreEqual(0, tokens.Count);
+        }
+
+        [Test]
+        public void Whitespace()
+        {
+            var tokens = Enumerate("  ", ',');
+            Assert.AreEqual(1, tokens.Count);
+            Assert.AreEqual("  ", tokens[0]);
+        }
+
+        [Test]
+        public void Tokens()
+        {
+            var tokens = Enumerate("a,b,c", ',');
+            Assert.AreEqual(3, tokens.Count);
+            Assert.AreEqual("a", tokens[0]);
+            Assert.AreEqual("b", tokens[1]);
+            Assert.AreEqual("c", tokens[2]);
+        }
+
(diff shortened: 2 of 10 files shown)


src/NUnitFramework/framework/Internal/StringTokenEnumerator.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// </summary>
+        /// <param name="target">Target string to enumerate.</param>
+        /// <param name="separator">Separator between tokens to return.</param>
+        public StringTokenEnumerator(string target, char separator)
+        {
+            _currentStartIndex = 0;
+            _target = target;
+            _separator = separator;
+            Current = null;
+            Current = null;
```
mikkelbu, line 30:
```suggestion
            Current = null;
```

src/NUnitFramework/framework/Internal/StringTokenEnumerator.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+        }
+
+        /// <inheritdoc />
+        public string Current { get; private set; }
+
+        object IEnumerator.Current => Current;
+
+        /// <summary>
+        /// Return unboxed enumerator then type is well known, do not remove.
```
mikkelbu, line 75: Should "then" be "when" ?
```suggestion
        /// Return unboxed enumerator when type is well known, do not remove.
```

src/NUnitFramework/framework/Internal/StringTokenEnumerator.cs
```diff
@@ -0,0 +1,91 @@
...
+                _currentStartIndex = endIndex + 1;
+            }
+
+            return true;
+        }
+
+        /// <inheritdoc />
+        public void Reset()
+        {
+            _currentStartIndex = -2;
```
mikkelbu, line 61: This seem wrong. After a call to `Reset` the following calls to `MoveNext` will fail as we cannot have a negative start. I think we either should return `0` instead of `-2` or perhaps throw `NotSupportedException` - I prefer to return `0`.

From the documentation https://docs.microsoft.com/en-us/dotnet/api/system.collections.ienumerator.reset?view=net-6.0
> The [Reset](https://docs.microsoft.com/en-us/dotnet/api/system.collections.ienumerator.reset?view=net-6.0) method is provided for COM interoperability. It does not necessarily need to be implemented; instead, the implementer can simply throw a [NotSupportedException](https://docs.microsoft.com/en-us/dotnet/api/system.notsupportedexception?view=net-6.0).

src/NUnitFramework/tests/Internal/StringTokenEnumeratorTests.cs
```diff
@@ -0,0 +1,45 @@
...
+        [Test]
+        public void Tokens()
+        {
+            var tokens = Enumerate("a,b,c", ',');
+            Assert.AreEqual(3, tokens.Count);
+            Assert.AreEqual("a", tokens[0]);
+            Assert.AreEqual("b", tokens[1]);
+            Assert.AreEqual("c", tokens[2]);
+        }
+
```
mikkelbu, line 33: Can we also add some tests where we have longer strings between the tokens and also some where the tokens are next to each other.

EDIT: I made the following tests and one of them fail as the string ends with a token
```csharp
        [TestCase("aaa,,cccdd", ',')]
        [TestCase("somestring;another", ';')]
        [TestCase(",abc,def,,somelongerstring,", ',')] // Fails
        public void CompareAgainstStringSplit(string input, char separator)
        {
            var tokens = Enumerate(input, separator);
            var tokens2 = input.Split(separator);
            Assert.That(tokens, Is.EqualTo(tokens2));
        }
```
The failing testcase fails with 
```
  Expected is <System.String[6]>, actual is <System.Collections.Generic.List`1[System.String]> with 5 elements
  Values differ at index [5]
  Missing:  < <string.Empty> >
```

Dreamescaper:
Do those 12mb of allocations cause any significant issues? I'm hesitant, because this code brings significant additional complexity.
FWIW, I would simply replace all `.Split(new[] { '*' })` with `.Split('*')` (it wouldn't affect Framework builds in any way, but will slightly decrease allocations for Core).

src/NUnitFramework/framework/Internal/StringTokenEnumerator.cs
lahma: Oops fixed

src/NUnitFramework/framework/Internal/StringTokenEnumerator.cs
lahma: Ah, yes, I did some expirements and this was left. Thanks for checking for requirements, will go with NotSupportedException.

src/NUnitFramework/framework/Internal/StringTokenEnumerator.cs
lahma: Yes, it should.

src/NUnitFramework/tests/Internal/StringTokenEnumeratorTests.cs
lahma: Great find, I've modified the logic to account for this and also support to parametrize whether to return empty tokens. By default empty ones are not returned, adjusted this test case to test both.

lahma:
@mikkelbu thanks for the thorough review, I've hopefully addressed your concerns in the latest commit.

@Dreamescaper these allocations cause significant issues with large test suites containing categories. I hope the complexity is inside the single enumerator and covered by tests.

And it's 12GB, not 12MB. I don't know why they are using comma as thousand separator in dotTrace when using my Finnish locale (should be space), below a screenshot with allocations in descending order.

image

stevenaw:
Thanks @lahma 
The gains in the benchmark look nice, but I share where @Dreamescaper is coming from, at least as it pertains to the .NET Standard build of the framework. Out of curiousity, how much does this save on a test suite of your size?

It looks like your benchmarks target .NET 6, which supports the `String.Split(char, StringSplitOptions = StringSplitOptions.None)` overload. How do the benchmarks look if that were added? If it's comparable to `StringTokenEnumerator` would it make sense to use the built-in `String.Split(char)` overload in the netstandard build of nunit and `StringTokenEnumerator` in the .NET framework build?

NOTE: _If_ that route makes sense, it may be easiest to define a `string.split(char)` extension function within nunit on netframework builds which would simply return your `StringTokenEnumerator` for enumerating, similar to some of the other classes in the internal "compatibility" namespace (ex: [ReflectionExtensions.cs](https://github.com/nunit/nunit/blob/v3.13-dev/src/NUnitFramework/framework/Compatibility/ReflectionExtensions.cs) in the `v3.13-dev` branch)

lahma:
@stevenaw thank you for the feedback. I think it would make more sense run further benchmarks for alternatives after the fate of https://github.com/nunit/nunit/pull/4034 has been decided. This string-split PR doesn't make much difference as long as the test suite for described use case takes over 10 minutes to get prepared.

lahma:
I re-did the tests and now actually obeying the default configuration where NUnit is built with debug mode (optimize = false). My version can be even two times slower than any .NET library version in NUnit's debug mode, with release mode it's slightly faster and of course doesn't allocate. I cannot compete against .NET BCL release build binaries with NUnit that is compiled with `<Optimize>false</Optimize>`.

I don't fully understand why NUnit *consumer* would need to step into NUnit code when debugging or running tests. This was changed in https://github.com/nunit/nunit/pull/3464 which basically makes NUnit library quite a lot slower when evaluating things. One would think that having symbol packages around etc would allow sufficient debugging experience - if NUnit debugging is part of debugging consumer-side code.

So I'm closing this as I cannot add much value with current build configuration.

stevenaw:
Thanks @lahma for investigating.
I agree with you 100% about debugging options for consumers, and I think symbols are also likely a better option. If you don't mind, I'd actually like to keep this open, or somehow on the radar, if there's a chance it's faster on release mode than any alternatives.

I'll try to start a discussion about consumer debugging options and experience when I get back to a computer later today.

lahma:
Thank you @stevenaw for chiming in,  I've reopened and let's see what future holds.

lahma:
As an additional note, if NUnit would take additional dependency on Microsoft.Extensions.Primitives, it would have [similar functionality baked-in](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Primitives/src/StringTokenizer.cs). Also would bring StringSegment which is like a span, but not so strict usage-wise.

SimonCropp:
i know this is a delayed request. but should this be retargeted to master for a v4 release?

lahma:
I've rebased and re-targeted this against master. Contents should be the same (use `Tokenize` when can, `string.Split` with single char argument otherwise).

stevenaw (approved):
Thanks for maintaining this branch until we got to this @lahma .
LGTM. Thanks for the contribution!

PR #4040, opened 2022-02-16
Master issue/4030 empty constraint

### LINKED ISSUE #4030: IsEmpty doesn't work with new .NET6 PriorityQueue

manfred-brands:
Fixes #4030 in master

Same as #4031


stevenaw (approved):
Thanks once again @manfred-brands !

PR #4042, opened 2022-02-28
Minimize + Centralize empty array declarations

### LINKED ISSUE #4041: Minimze empty array allocations via centralized helper for pre-net46
net462 + netstandard2 define a static `Array.Empty<T>()` helper which returns a cached empty array. There's a few places in the code where we `#ifdef` this in where it's easy but there's other places where this approach would be cumbersome. It might help to both cleanup the code + give a minor perf boost (via reduced allocations) by creating a helper to handle this.

This would be most useful for v3.13.x but could also have a place in v4 if we continue to target net45 there (https://github.com/nunit/nunit/issues/4036). If net45 support _is_ dropped in v4, then we should instead update all callsites to just call `Array.Empty<T>` directly instead

stevenaw:
Fixes #4041

This PR introduces a helper to cache empty array declarations. It will use `Array.Empty<T>` if possible, falling back to a custom implementation when that's not available.

This should be ported to v4 unless we decide to drop support for net45 there.

### DIFF
--- a/src/NUnitFramework/framework/Internal/ArrayHelper.cs
+++ b/src/NUnitFramework/framework/Internal/ArrayHelper.cs
@@ -1,0 +1,31 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+#nullable enable
+
+using System;
+using System.Runtime.CompilerServices;
+
+namespace NUnit.Framework.Internal
+{
+    internal static class ArrayHelper
+    {
+#if NET45_OR_GREATER
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+#endif
+        public static T[] Empty<T>()
+        {
+#if NETSTANDARD2_0_OR_GREATER
+            return Array.Empty<T>();
+#else
+            return InternalArrayHelper<T>.Empty;
+#endif
+        }
+
+#if !NETSTANDARD2_0_OR_GREATER
+        private static class InternalArrayHelper<T>
+        {
+            public static readonly T[] Empty = new T[0];
+        }
+#endif
+    }
+}
--- a/src/NUnitFramework/framework/Attributes/SetUpFixtureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/SetUpFixtureAttribute.cs
@@ -78,7 +78,7 @@ namespace NUnit.Framework
                     return false;
                 }
 
-                if (!typeInfo.HasConstructor(new Type[0]))
+                if (!typeInfo.HasConstructor(ArrayHelper.Empty<Type>()))
                 {
                     reason = string.Format("{0} does not have a default constructor", typeInfo.FullName);
                     return false;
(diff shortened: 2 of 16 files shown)


manfred-brands (changes requested):
The idea and changes are good.
You forgot one in `framework` directory in `DefaultSuiteBuilder.cs`: ` IFixtureBuilder[] attrs = new IFixtureBuilder[0];`
I assume you deliberately ignored zero array allocations in the `testdata` and `tests` directories.

manfred-brands (approved):
I would not do more here and for main move to `Array.Empty` once we have a minimum target of net462

src/NUnitFramework/framework/Internal/ArrayHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System;
+using System.Runtime.CompilerServices;
+
+namespace NUnit.Framework.Internal
+{
+    internal static class ArrayHelper
+    {
+#if NET45_OR_GREATER
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
mikkelbu, line 13: Only comment from me. Would it make any difference to have `[MethodImpl(256)]` as in https://github.com/nunit/nunit/pull/4035/files#diff-a8ef56950756871bf762f8132c715423e246747709de814a3734d7bb00377448 ? or does that make any difference ?

mikkelbu (approved):
Only one comment, but I'm also happy to merge it as it is

src/NUnitFramework/framework/Internal/ArrayHelper.cs
stevenaw: Thanks @mikkelbu To be honest, I'm not sure about difference.

`[MethodImpl(256)]` and `MethodImpl(MethodImplOptions.AggressiveInlining)` are semantically identical since `MethodImplOptions.AggressiveInlining = 256`, with the exception that using 256 explicitly allows the code to compile on build targets which predate the enum value (introduced in .NET 4.5). Though the JIT still needs to support reading the flag from the IL for it to make a difference, and I'm not sure when that was added.

If older JITs (ex: .NET 3.5 / CLR 2) don't recognize the flag then the main difference I could see it making is if a .NET 4.5+ app were to reference a .NET 3.5 or .NET4 build of the framework. Which is possible, though I'd think it rare given nuget "should" likely serve the .NET4.5 build of the framework to the app when restored. In that case, I think the JIT should still inline it even if we don't explicitly tell it because it is a small, simple method non-polymorphic method (a few of the heuristics are listed [here](https://devblogs.microsoft.com/premier-developer/a-common-execution-path-optimization/)).

Happy to make the change though if we think it could help keep the code cleaner or more readable. I may put it behind a named constant if we do use it in multiple places in the codebase. I should be able to get to coding it later tonight. 😃

mikkelbu: Then I think we should merge it as it is :)

stevenaw: Thanks @mikkelbu !

PR #4050, opened 2022-03-02
4036 drop net45

### LINKED ISSUE #4036: Drop net45 build target in nunit4
The framework library is currently built for net45 + netstandard2.0, but net45 has hit EOL. Within a few months time, by April 26, 2022, net462 will be the minimum supported runtime by Microsoft (excluding 3.5 SP1) https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/

I'm finding I'm having a hard time installing the required dependencies on Windows 11 to build the net45 version of the framework. It may be worth considering targeting minimum net462 in the next semver-major release 

@nunit/framework-team thoughts?

stevenaw:
Fixes #4036 

Updates the framework to target net462 where it had been targeting net45 prior

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -37,8 +37,7 @@ As the NUnit solution targets multiple frameworks, a single build will generate 
        Debug\
           net6.0
           net5.0
-          net45
-          net46
+          net462
           netcoreapp3.1
           netstandard2.0
 ```
@@ -132,7 +131,7 @@ Feature constants are defined in [Directory.Build.props](src/NUnitFramework/Dire
 - `THREAD_ABORT` enables timeouts and forcible cancellation
 
 Platform constants are defined by convention by the csproj SDK, one per target framework.
-For example, `NET45`, `NETSTANDARD2_0`, `NETCOREAPP2_1`, and so on.
+For example, `NET462`, `NETSTANDARD2_0`, `NETCOREAPP2_1`, and so on.
 It is most helpful to call out which platforms are the exception in rather than the rule
 in a given scenario. Keep in mind the effect the preprocessor would have on a newly added platform.
 
@@ -140,15 +139,15 @@ For example, rather than this code:
 
 ```cs
 #if NETSTANDARD2_0 || NETSTANDARD2_1
-// Something that .NET Framework 4.5 can't do
+// Something that .NET Framework can't do
 #endif
 ```
 
 Consider this:
 
 ```cs
-#if !NET45
-// Something that .NET Framework 4.5 can't do
+#if !NETFRAMEWORK
+// Something that .NET Framework can't do
 #endif
 ```
 
--- a/build.cake
+++ b/build.cake
@@ -30,7 +30,7 @@ var packageVersion = version + modifier + dbgSuffix;
 // Equivalent of NUnitLibraryFrameworks in Directory.Build.props
 var LibraryFrameworks = new string[]
 {
-    "net45",
+    "net462",
     "netstandard2.0"
 };
 
@@ -164,13 +164,13 @@ Task("CheckForError")
     .Description("Checks for errors running the test suites")
     .Does(() => CheckForError(ref ErrorDetail));
 
-Task("Test45")
-    .Description("Tests the .NET 4.5 version of the framework")
+Task("TestNetFramework")
+    .Description("Tests the .NET Framework version of nunit framework")
     .IsDependentOn("Build")
     .OnError(exception => { ErrorDetail.Add(exception.Message); })
     .Does(() =>
     {
-        var runtime = "net45";
+        var runtime = "net462";
         var dir = BIN_DIR + runtime + "/";
         RunTest(dir + EXECUTABLE_NUNITLITE_TEST_RUNNER_EXE, dir, FRAMEWORK_TESTS, dir + "nunit.framework.tests.xml", runtime, ref ErrorDetail);
         //RunNUnitTests(dir, FRAMEWORK_TESTS, runtime, ref ErrorDetail);
@@ -536,7 +536,7 @@ Task("Rebuild")
 Task("Test")
     .Description("Builds and tests all versions of the framework")
     .IsDependentOn("Build")
-    .IsDependentOn("Test45")
+    .IsDependentOn("TestNetFramework")
     .IsDependentOn("TestNetStandard20");
 
 Task("Package")
--- a/src/NUnitFramework/framework/Internal/PlatformHelper.cs
+++ b/src/NUnitFramework/framework/Internal/PlatformHelper.cs
@@ -214,16 +214,15 @@ namespace NUnit.Framework.Internal
                 case "MACOSX":
                     isSupported = _os.IsMacOSX;
                     break;
+
                 // These bitness tests relate to the process, not the OS.
-                // We can't use Environment.Is64BitProcess because it's
-                // only supported in NET 4.0 and higher.
                 case "64-BIT":
                 case "64-BIT-PROCESS":
-                    isSupported = IntPtr.Size == 8;
+                    isSupported = Environment.Is64BitProcess;
                     break;
                 case "32-BIT":
                 case "32-BIT-PROCESS":
-                    isSupported = IntPtr.Size == 4;
+                    isSupported = !Environment.Is64BitProcess;
                     break;
 
                 case "64-BIT-OS":
--- a/src/NUnitFramework/framework/Internal/TestParameters.cs
+++ b/src/NUnitFramework/framework/Internal/TestParameters.cs
@@ -13,12 +13,7 @@ namespace NUnit.Framework.Internal
     /// </summary>
     public abstract class TestParameters : ITestData, IApplyToTest
     {
-        internal static readonly object[] NoArguments =
-#if NET45 // New in net46
-            new object[0];
-#else
-            Array.Empty<object>();
-#endif
+        internal static readonly object[] NoArguments = Array.Empty<object>();
 
         /// <summary>
         /// Default Constructor creates an empty parameter set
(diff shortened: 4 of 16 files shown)


src/NUnitFramework/framework/Internal/PlatformHelper.cs
```diff
@@ -214,16 +214,15 @@ public bool IsPlatformSupported( string platform )
                 case "MACOSX":
                     isSupported = _os.IsMacOSX;
                     break;
+
                 // These bitness tests relate to the process, not the OS.
-                // We can't use Environment.Is64BitProcess because it's
-                // only supported in NET 4.0 and higher.
```
stevenaw, line 219: Not specific to 4.6.2 per se, but something that might've been missed as part of removing .NET 4.0 support

src/NUnitFramework/framework/Internal/TestParameters.cs
```diff
@@ -13,12 +13,7 @@ namespace NUnit.Framework.Internal
     /// </summary>
     public abstract class TestParameters : ITestData, IApplyToTest
     {
-        internal static readonly object[] NoArguments =
-#if NET45 // New in net46
```
stevenaw, line 17: This crossed my radar because of the macro here, but I'll create a separate ticket to track a comprehensive adoption of Array.Empty<T>

src/NUnitFramework/framework/Internal/Tests/TestSuite.cs
```diff
@@ -124,8 +124,7 @@ public void Sort()
 
                 foreach (Test test in Tests)
                 {
-                    TestSuite? suite = test as TestSuite;
-                    if (suite != null)
+                    if (test is TestSuite suite)
```
stevenaw, line 127: Bit of an opportunistic update that the compiler warned me about

src/NUnitFramework/tests/Internal/RuntimeFrameworkTests.cs
```diff
@@ -317,6 +317,7 @@ public override string ToString()
             new FrameworkData(RuntimeType.NetFramework, new Version(3,5), new Version(2,0,50727), "net-3.5", "Net 3.5"),
             new FrameworkData(RuntimeType.NetFramework, new Version(4,0), new Version(4,0,30319), "net-4.0", "Net 4.0"),
             new FrameworkData(RuntimeType.NetFramework, new Version(4,5), new Version(4,0,30319), "net-4.5", "Net 4.5"),
+            new FrameworkData(RuntimeType.NetFramework, new Version(4,6), new Version(4,0,30319), "net-4.6", "Net 4.6"),
```
stevenaw, line 320: The current code doesn't support reading "build" numbers from the provided `Version` so I left it like this for now but I'd appreciate some feedback on this area please.

src/NUnitFramework/tests/Internal/RuntimeFrameworkTests.cs
```diff
@@ -26,17 +26,17 @@ public void CanGetCurrentFramework()
...
 
-#if NET45
+#if NETFRAMEWORK
         [Test]
         public void TargetFrameworkIsSetCorrectly()
         {
             // We use reflection so it will compile and pass on Mono,
             // including older versions that do not have the property.
-            var prop = typeof(AppDomainSetup).GetProperty("FrameworkName");
+            var prop = typeof(AppDomainSetup).GetProperty("TargetFrameworkName");
```
stevenaw, line 35: I was a little confused by this, it looks like this prop is actually named `TargetFrameworkName` in the .NET Framework, so I update this test accordingly

stevenaw:
@nunit/framework-team I've taken a first-run at making this update. CI is currently a bit of an issue, but I'd appreciate some early feedback on some of the issues noted above please.

build.cake
```diff
@@ -164,13 +164,13 @@ Task("CheckForError")
     .Description("Checks for errors running the test suites")
     .Does(() => CheckForError(ref ErrorDetail));
 
-Task("Test45")
-    .Description("Tests the .NET 4.5 version of the framework")
+Task("TestNetFramework")
```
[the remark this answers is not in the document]
manfred-brands: I do like it as we want to test the .net framework version. This way you don't have to update if once we move the minimum target to 4.8

src/NUnitFramework/tests/Internal/RuntimeFrameworkTests.cs
manfred-brands: So how did this test ever pass?
I see the test uses `Assume.That(prop, Is.Not.Null);` which marks the test as inconclusive instead of failed.
Never heard of that class. We would use `Assert.That(prop, Is.Not.Null);` which would cause the test to fail.

Looking at the last test run on master: `Test Count: 5823, Passed: 5641, Failed: 0, Warnings: 0, Inconclusive: 120, Skipped: 62` There are 120 inconclusive tests. How many of these are hidden failures because of failing assumptions?
Not sure what the idea is/was behind `Assume`.

manfred-brands: This test is now failing in the Linux build: 
```
Failed : NUnit.Framework.Internal.RuntimeFrameworkTests.TargetFrameworkIsSetCorrectly
  Expected: String starting with ".NETFramework,Version=v4."
  But was:  null
```

This test should never run on Linux as Linux doesn't support .NET Framework.

stevenaw: Thanks for checking this out :D 
I'll have to take a closer look, but prior to my change (on my local windows) the behaviour I saw was:
- It was wrapped in a `#if NET45`
- It was "Inconclusive" because the prop lookup was returning null on Windows
Now:
- It is wrapped in a `#if NETFRAMEWORK`
- It is passing on Windows
- It is (apparently) running + failing on linux
Will take a bit of digging. I'm very curious about this myself too

stevenaw: Alrighty, after some digging this appears to be a mono-specific issue like the comment indicates. Tricky thing is, mono can be installed on Windows in additional to *nix (though our CI doesn't test mono on Windows). I suspect that this test had been "inconclusive" for a while though I assume it must've worked at some point in the past. I can't find a previous version of the .NET Framework which has a `AppDomainSetup.FrameworkName` property, however.

At any rate, either of the two options seem to work:

- Use `[Platform("Win")]`
  - This works for us since the netfx tests only mono on linux/mac
- Use the `Assume.That(Is.Not.Null)` approach on the _return value_ of the prop invocation rather than the prop returned via reflection
  - This allows us to work-around the runtime differences without getting into OS differences, but has the potential to silently turn "inconclusive" again in the future

I've left each of these approaches as separate commits in the history (the last 2 commits in this PR). No strong preference from me on either, other than that I slightly prefer the second because it allows us to test the `RuntimeFramework` class without relying on an attribute which _also_ relies on the class under test.

stevenaw: Wait, this doesn't actually test nunit code from what I can tell. Oops

stevenaw: Alrighty, I have removed these tests completely. My impression is that they were added to test netfx-specific functionality introduced in NET45 (which we now no longer support). I'd appreciate hearing any feedback the team may have about this.

BUILDING.md
```diff
@@ -56,17 +55,15 @@ The tests that should be run in the solution are grouped by project name:
...
+You should then be able to run the unit tests directly from within your development IDE of choice against one or all target frameworks (as you would when using NUnit in any other development project). For example, this is what it looks like in JetBrains Rider (2021.1.2) when right clicking on the `AssertEqualsTests` TextFixture:
 
 ![image](https://user-images.githubusercontent.com/52075808/121511286-61775580-c9e0-11eb-8e1e-ff44d0d8873d.png)
 
-Because NUnit solution targets multiple frameworks, JetBrains Rider knows this and offers the option to run the tests against a specific framework and/or all target frameworks.
-
-Unfortunately, there is currently a known issue ([#3008](https://github.com/nunit/nunit/issues/3008)) preventing the tests from being run in Visual Studio and JetBrains Rider IDEs. Equally there is also a known issue ([#3867](https://github.com/nunit/nunit/issues/3867)) preventing the tests from being run from the command line using `dotnet test`
```
stevenaw, line 65: I've moved this into a "Known Issues and Workarounds" section below instead of including it under "Tests"

stevenaw:
Latest merge conflicts have been resolved. This should now be ready for review.

manfred-brands (approved):
Nice work.

stevenaw:
Thanks @manfred-brands !

BUILDING.md
```diff
@@ -56,17 +55,15 @@ The tests that should be run in the solution are grouped by project name:
 
 Other test projects contain tests designed to fail purposely for integration tests.
 
-Normally you should be able to run the unit tests directly from within your development IDE of choice (as you would when using NUnit in any other development project). For example, this is what it looks like in JetBrains Rider (2021.1.2) when right clicking on the `AssertEqualsTests` TextFixture:
+You should then be able to run the unit tests directly from within your development IDE of choice against one or all target frameworks (as you would when using NUnit in any other development project). For example, this is what it looks like in JetBrains Rider (2021.1.2) when right clicking on the `AssertEqualsTests` TextFixture:
```
mikkelbu, line 58: I initially, thought that it was odd to write "You should then be able to run the unit tests directly from within...." and then later have "Known Issues and Workarounds", but looking at the resulting page I think this is fine

mikkelbu (approved):
LGTM

stevenaw:
Thanks @mikkelbu ! 
I had tried to retain the original wording as much as possible on the documentation updates, but I'd also be happy to make any further changes. Since we're a few days past net45 EOL, I'd like to go ahead and merge this today, but I can happily do a follow-up PR to adjust any wording in the README.

PR #4056, opened 2022-03-09
Cache RegEx in ValueMatchFilter

### LINKED ISSUE #4055: Use static Regex.IsMatch in ValueMatchFilter to take advantage of caching
The static Regex functions internally cache parsed regex instances based on the provided pattern on .NET 5+. ValueMatchFilter can be used to filter tests based on a regex and will apply the same regex to all potential candidates to see if it matches. This means that on a large test suite there could be measurable savings by using an overload which caches the regex.

Something like this:
https://github.com/nunit/nunit/blob/master/src%2FNUnitFramework%2Fframework%2FInternal%2FFilters%2FValueMatchFilter.cs#L43

```diff
-return input != null && new Regex(ExpectedValue).IsMatch(input);
+return input != null && Regex.IsMatch(input, ExpectedValue);
```

manfred-brands:
Fixes #4055

### DIFF
--- a/src/NUnitFramework/framework/Internal/Filters/IdFilter.cs
+++ b/src/NUnitFramework/framework/Internal/Filters/IdFilter.cs
@@ -15,7 +15,7 @@ namespace NUnit.Framework.Internal.Filters
         /// Construct an IdFilter for a single value
         /// </summary>
         /// <param name="id">The id the filter will recognize.</param>
-        public IdFilter(string id) : base (id) { }
+        public IdFilter(string id) : base(id, false) { }
 
         /// <summary>
         /// Match a test against a single value.
--- a/src/NUnitFramework/framework/Internal/TestFilter.cs
+++ b/src/NUnitFramework/framework/Internal/TestFilter.cs
@@ -156,7 +156,7 @@ namespace NUnit.Framework.Internal
             {
                 case "filter":
                 case "and":
-                    List<TestFilter> childFilters = new List<TestFilter>();
+                    var childFilters = new List<TestFilter>(node.ChildNodes.Count);
 
                     foreach (var childNode in node.ChildNodes)
                         childFilters.Add(FromXml(childNode));
@@ -164,7 +164,7 @@ namespace NUnit.Framework.Internal
                     return new AndFilter(childFilters.ToArray());
 
                 case "or":
-                    List<TestFilter> orChildFilters = new List<TestFilter>();
+                    var orChildFilters = new List<TestFilter>(node.ChildNodes.Count);
 
                     foreach (var childNode in node.ChildNodes)
                         orChildFilters.Add(FromXml(childNode));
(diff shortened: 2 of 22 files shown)


src/NUnitFramework/framework/Internal/Filters/IdFilter.cs
```diff
@@ -15,7 +15,7 @@ internal class IdFilter : ValueMatchFilter
         /// Construct an IdFilter for a single value
         /// </summary>
         /// <param name="id">The id the filter will recognize.</param>
-        public IdFilter(string id) : base (id) { }
+        public IdFilter(string id) : base(id, false) { }
```
stevenaw, line 18: 👍

src/NUnitFramework/framework/Internal/TestFilter.cs
```diff
@@ -156,15 +156,15 @@ public static TestFilter FromXml(TNode node)
             {
                 case "filter":
                 case "and":
-                    List<TestFilter> childFilters = new List<TestFilter>();
+                    var childFilters = new List<TestFilter>(node.ChildNodes.Count);
```
stevenaw, line 159: Nice touch to pre-initialize this. Thanks!

nitpick: Since we have to convert it to an array a few lines below, would it make sense to make this an array here instead of a list to save an array copy operation? Similar suggest for the "or" below.

src/NUnitFramework/framework/Internal/TestFilter.cs
```diff
@@ -178,27 +178,27 @@ public static TestFilter FromXml(TNode node)
                     return new IdFilter(node.Value);
 
                 case "test":
-                    return new FullNameFilter(node.Value) { IsRegex = isRegex };
+                    return new FullNameFilter(node.Value, isRegex);
```
stevenaw, line 181: 👍

stevenaw (approved):
Nicely done, thanks @manfred-brands !
Very minor nitpick suggestion related to pre-initializing the and/or filters, but it's also fine to merge as-is too. Thanks!

src/NUnitFramework/framework/Internal/TestFilter.cs
manfred-brands: I agree with your nitpick, I should have looked how the list was used after.
Changed code to create array, extracted to a function so I only needed one copy.

manfred-brands:
@stevenaw Can you look at `TestFilter.TopLevel` property. I can't find any real usages.
There is code commented out in `NotFilter` and a test that checks the "TopLevel" property.
Can all of that be deleted as part of this PR?

stevenaw:
> @stevenaw Can you look at `TestFilter.TopLevel` property. I can't find any real usages. There is code commented out in `NotFilter` and a test that checks the "TopLevel" property. Can all of that be deleted as part of this PR?

Thanks for noticing this 😊 I agree, it looks unused and appears to have been that way since a beta of nunit 3. That predates my involvement in the project, however, so I'd appreciate a second opinion. @CharliePoole any concerns with removing the code?

CharliePoole:
@stevenaw The commenting out was part of a fairly large change I made back in 2015. The idea was that we were going to re-visit the question of how we wanted top-level selected tests (those actually selected by the user) and `ExplicitAttribute` to interact. I don't think we ever resolved it and, of course, I eventually left the project so was no longer pushing to resolve it.

My 2 cents: If you find that Explicit is now working the way you want it to work (and/or as it's documented to work) I'd remove the property and all the commented out code that references it, possibly in a separate PR (but that's up to you guys).

manfred-brands:
@stevenaw I will create a new issue and PR once this one is in to remove the TopLevel property.

stevenaw:
Sounds good, thanks @CharliePoole and @manfred-brands 👍

rprouse (approved):

stevenaw:
LGTM. Thanks @manfred-brands !

PR #4066, opened 2022-03-16
Use pattern matching in the comparers

### LINKED ISSUE #4065: Use pattern matching in the comparers
Most of the comparers perform type-checking using the following pattern:

```csharp
if (x is DesiredType)
  ((DesiredType)x).DoSomething()
```

Newer versions of C# added the ability to avoid the second cast by defining a temp variable.
```csharp
if (x is DesiredType dt)
  dt.DoSomething()
```

This helps simplify code and provides a small perf boost by avoiding an unneccessary cast. This could also minimize future changes if the comparers ever get a generic overload as pattern matching can play nicer than casts in that case

stevenaw:
Fixes #4065

A small clean-up PR on an old part of the codebase to opt into newer language features with the possibility of small perf boost

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
@@ -16,7 +16,7 @@ namespace NUnit.Framework.Constraints
     public class DictionaryContainsKeyConstraint : CollectionItemsEqualConstraint
     {
         private const string ContainsMethodName = "Contains";
-        private bool _isDeprecatedMode = false;
+        private readonly bool _isDeprecatedMode = false;
 
         /// <summary>
         /// Construct a DictionaryContainsKeyConstraint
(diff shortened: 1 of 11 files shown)


src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs
```diff
@@ -16,7 +16,7 @@ namespace NUnit.Framework.Constraints
     public class DictionaryContainsKeyConstraint : CollectionItemsEqualConstraint
     {
         private const string ContainsMethodName = "Contains";
-        private bool _isDeprecatedMode = false;
+        private readonly bool _isDeprecatedMode = false;
```
stevenaw, line 19: A few non-pattern-matching things like this were also updated in this PR, each of which should give more info to the JIT and potentially a small perf boost

rprouse (approved):

manfred-brands (approved):
Nice work

PR #4069, opened 2022-03-19
Remove references to TopLevel in comments and tests and remove TopLevel property

### LINKED ISSUE #4058: Remove TopLevel property from ValueMatchFilter
As Found in #4056

This property was left behind and is only accessed in commented out code and one unit test (using reflection).

Phyrik:
Fixes #4058

Let me know if there's any references I missed but all the tests are passing so I think it should be all done


manfred-brands (approved):
Looks good. Thanks @Phyrik 
We need a 2nd approval.

rprouse (approved):

PR #4073, opened 2022-03-20
Use pattern matching instead of casting in Constraints namespace

### LINKED ISSUE #4072: Use pattern matching in Constraints to avoid unecessary casts
Sibling to https://github.com/nunit/nunit/issues/4065
Use pattern matching in everything under `Constraints` namespace to avoid unnecessary casts or comparisons to give a minor perf boost.

stevenaw:
Fixes #4072
Use pattern matching in everything under Constraints namespace to avoid unnecessary casts or comparisons to give a minor perf boost.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionTally.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionTally.cs
@@ -28,7 +28,7 @@ namespace NUnit.Framework.Constraints
             }
         }
 
-        private readonly NUnitEqualityComparer comparer;
+        private readonly NUnitEqualityComparer _comparer;
 
         private readonly bool _isSortable;
         private bool _sorted = false;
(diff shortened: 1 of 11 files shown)


src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -28,7 +28,7 @@ public CollectionTallyResult(List<object> missingItems, List<object> extraItems)
             }
         }
 
-        private readonly NUnitEqualityComparer comparer;
+        private readonly NUnitEqualityComparer _comparer;
```
stevenaw, line 31: Not technically a pattern matching update, but this makes this class-level variable naming consistent with below (and removes a compiler INFO message in the process)

src/NUnitFramework/framework/Constraints/Constraint.cs
```diff
@@ -156,20 +154,19 @@ protected virtual string GetStringRepresentation()
...
-                sb.Append(_displayable(arg));
+                sb.Append(Displayable(arg));
             }
 
             sb.Append(">");
 
             return sb.ToString();
-        }
-
-        private static string _displayable(object o)
```
stevenaw, line 167: Similar as above. Moved to a static local function and renamed to fix a few compiler warnings. There's pattern matching updates within it too

rprouse (approved):
Looks good, thanks.

PR #4076, opened 2022-03-21
Remove unnecessary comparer allocations from NUnitEqualityComparer

### LINKED ISSUE #4075: Remove unnecessary allocations in NUnitEqualityComparer
The `NUnitEqualityComparer` on every instantiation creates a list if `IChainComparer` instances.
The only reason these are instances is because the constructor passes in the current instance of  `NUnitEqualityComparer`
Refactor to static methods where `this` is passed as an extra argument to the `Equal` method instead.

This should reduce pressure on the garbage collector.

manfred-brands:
Fixes #4075


mikkelbu (approved):
Looks good to me (and I think I were the one that introduced the allocations).

PR #4079, opened 2022-03-22
TargetFrameworkIdentifier is not set when Directory.Directory.Build.props is parsed

### LINKED ISSUE #4078: THREAD_ABORT not properly set
This was detected when backporting this to 3.13-dev.

As part of #3984 the following line was replaced in `Directory.Build.props`

```xml
<DefineConstants Condition="'$(TargetFramework)' != 'netstandard2.0'
                            and '$(TargetFramework)' != 'netcoreapp2.1'
                            and '$(TargetFramework)' != 'netcoreapp3.1'
                            and '$(TargetFramework)' != 'net5.0'
                            and '$(TargetFramework)' != 'net5.0-windows'">$(DefineConstants);THREAD_ABORT</DefineConstants>
```

With:

```xml
 <DefineConstants Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">$(DefineConstants);THREAD_ABORT</DefineConstants>
```

Unfortunately `TargetFrameworkIdentifier` is not yet set when `Directory.Build.props` is parsed, but only later in `Microsoft.NET.TargetFrameworkInference.targets` 

Then net effect is that `THREAD_ABORT` is never set.
This is not detected by unit test because we don't test `Thread.Abort` unless the symbol is set.

manfred-brands:
Fixes #4078

### DIFF
--- a/src/NUnitFramework/Directory.Build.props
+++ b/src/NUnitFramework/Directory.Build.props
@@ -18,7 +18,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <DefineConstants Condition="'$(TargetFrameworkIdentifier)' == '.NETFramework'">$(DefineConstants);THREAD_ABORT</DefineConstants>
+    <DefineConstants Condition="$(TargetFramework.StartsWith('net4'))">$(DefineConstants);THREAD_ABORT</DefineConstants>
   </PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


mikkelbu (approved):
LGTM. Thanks for noticing this @manfred-brands 👍

PR #4088, opened 2022-03-27
Constraints + AssemblyHelper use in-place string comparisions

### LINKED ISSUE #4086: Perform case-insensitive string comparisons in-place
NUnit currently uses `ToLower()` to do case-insensitive strings comparisons. We should consider moving towards in-place comparisons using the `Equals()` `StartsWith()` and `EndsWith()` overloads which take in a `stringcomparison` instead. This should help reduce memory overhead, especially on large suites where GC overhead can become a factor.

NOTE: I attempted this at one point in the past but I recall the potential for breaking behaviour on _either_ Windows or Linux on .netcore <= 3.1. I'm wondering if it's worth revisiting this now that netcore3.1 will be hitting EOL in the foreseeable future and .NET 5+ supports the same ICU tables as linux now (https://docs.microsoft.com/en-us/dotnet/core/extensions/globalization-icu)

stevenaw:
Fixes #4086 
Uses in-place comparisons instead of calling `ToLower()`. According to MSDN docs, the existing calls to `ToLower()` use a culture-aware comparer by default so I used `StringComparison.CurrentCultureIgnoreCase` for case-insensitive in-place comparisons to try and preserve existing behaviour.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
@@ -1,5 +1,7 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
+#nullable enable
+
 using System;
 
 namespace NUnit.Framework.Constraints.Comparers
@@ -14,10 +16,8 @@ namespace NUnit.Framework.Constraints.Comparers
             if (!(x is string xString) || !(y is string yString))
                 return null;
 
-            if (equalityComparer.IgnoreCase)
-                return xString.Equals(yString, StringComparison.CurrentCultureIgnoreCase);
-
-            return xString.Equals(yString);
+            var stringComparison = equalityComparer.IgnoreCase ? StringComparison.CurrentCultureIgnoreCase : StringComparison.Ordinal;
+            return xString.Equals(yString, stringComparison);
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
@@ -14,12 +14,10 @@ namespace NUnit.Framework.Constraints.Comparers
             if (!(x is string xString) || !(y is string yString))
                 return null;
 
-            bool caseInsensitive = equalityComparer.IgnoreCase;
+            if (equalityComparer.IgnoreCase)
+                return xString.Equals(yString, StringComparison.CurrentCultureIgnoreCase);
 
-            string s1 = caseInsensitive ? xString.ToLower() : xString;
-            string s2 = caseInsensitive ? yString.ToLower() : yString;
-
-            return s1.Equals(s2);
+            return xString.Equals(yString);
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
@@ -263,7 +263,7 @@ namespace NUnit.Framework.Constraints
                 if (obj is null)
                     return 0;
                 else if (_ignoreCase)
-                    return obj.ToLower().GetHashCode();
+                    return StringComparer.CurrentCultureIgnoreCase.GetHashCode(obj);
                 else
                     return obj.GetHashCode();
             }
--- a/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
@@ -254,10 +254,8 @@ namespace NUnit.Framework.Constraints
 
             public bool Equals(string x, string y)
             {
-                if (_ignoreCase)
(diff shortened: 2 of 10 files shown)


src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
```diff
@@ -14,12 +14,10 @@ internal static class StringsComparer
             if (!(x is string xString) || !(y is string yString))
                 return null;
 
-            bool caseInsensitive = equalityComparer.IgnoreCase;
+            if (equalityComparer.IgnoreCase)
+                return xString.Equals(yString, StringComparison.CurrentCultureIgnoreCase);
```
manfred-brands, line 18: I think it would be nicer to do something like:
```csharp
StringComparison stringComparison = equalityComparer.IgnoreCase ? StringComparison.CurrentCultureIgnoreCase : StringComparison.CurrentCulture;
return xString.Equals(yString, stringComparison)
```
Or in one go:
```csharp
return xString.Equals(yString, equalityComparer.IgnoreCase ? StringComparison.CurrentCultureIgnoreCase : StringComparison.CurrentCulture)
```

Same for all other cases. It is less code and show the same path, just another comparison flag.

manfred-brands (commented):
Suggestion to explicitly state comparer for none ignore case scenarios to have the same execution path.

src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
stevenaw: Thanks @manfred-brands ! I've pushed some updates for this. One note: I had to use Ordinal instead of CurrentCulture for the case-sensitive path to match existing behaviour

manfred-brands: The tricky part is that the defaults are inconsistent. Indeed .Equals uses Ordinal, but EndsWith uses CurrentCulture. Hence there is an Analyzer that suggests to always specify a Comparison.

manfred-brands: Ordinal is the preferred if we can use it.

stevenaw: Right, I'd forgotten about the different default. Thanks for noticing! I'll do another pass next chance I get. I think there's a case to be made for having all comparisons consistent, but I think that's a different convo for a different issue

manfred-brands (approved):
Thanks. Nothing further

src/NUnitFramework/tests/Constraints/EndsWithConstraintTests.cs
```diff
@@ -24,6 +24,15 @@ public void SetUp()
             new TestCaseData( "say hello to Fred", "\"say hello to Fred\"" ),
             new TestCaseData( string.Empty, "<string.Empty>" ),
             new TestCaseData( null , "null" ) };
+
+        [Test]
+        public void RespectsCulture()
+        {
+            var constraint = new EndsWithConstraint("r\u00E9sum\u00E9");
+
+            var result = constraint.ApplyTo("re\u0301sume\u0301");
```
stevenaw, line 33: I'm very glad to have found this. It took me a long time to find a single, clear test case which could be applied to demonstrate all the test scenarios in this PR and which worked in both NLS + ICU on all runtimes _locally_, and it looks like the CI agrees too

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -254,18 +254,16 @@ public NUnitStringEqualityComparer(bool ignoreCase)
...
+                var stringComparison = _ignoreCase ? StringComparison.CurrentCultureIgnoreCase : StringComparison.Ordinal;
+                return x.Equals(y, stringComparison);
             }
 
             public int GetHashCode(string obj)
             {
                 if (obj is null)
                     return 0;
                 else if (_ignoreCase)
-                    return obj.ToLower().GetHashCode();
```
stevenaw, line 268: Not only does this allocate, I found through testing that this code was incorrect and could erronously put some things which should be the same into different buckets.

src/NUnitFramework/framework/Internal/AssemblyHelper.cs
```diff
@@ -159,7 +159,7 @@ public static Assembly Load(string nameOrPath)
 
         private static bool IsFileUri(string uri)
         {
-            return uri.ToLower().StartsWith(UriSchemeFile);
+            return uri.StartsWith(UriSchemeFile, StringComparison.CurrentCultureIgnoreCase);
```
jnm2, line 162: I would like to change behavior here to OrdinalIgnoreCase if we can without breaking anyone. CurrentCulture and InvariantCulture are slower and semantically correct for linguistic operations, not for canonical identifiers.

jnm2 (approved):

src/NUnitFramework/framework/Internal/AssemblyHelper.cs
stevenaw: I think this should be a fairly safe change. I nearly did this myself before submitting the initial PR, but I opted for not doing it mostly out of abundance of caution.

manfred-brands (approved):

PR #4092, opened 2022-04-06
Remove unneccessary casts (and a few other minor cleanup)

stevenaw:
A bit of a broad cleanup within the framework to remove unnecessary casts where I found readability wasn't degraded.
While going through I also made a few class-level fields readonly or simplified null checks where I noticed the compiler advising about it. I tried to keep this cleanup focused on changes with a (minor) performance benefit and stayed away from purely stylistic changes.

No corresponding GH issue but I will make one if we prefer, please let me know

### DIFF
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -205,7 +205,7 @@ namespace NUnit.Framework.Api
 
         class ActionCallback : ICallbackEventHandler
         {
-            Action<string> _callback;
+            private readonly Action<string> _callback;
 
             public ActionCallback(Action<string> callback)
             {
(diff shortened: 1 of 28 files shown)


src/NUnitFramework/framework/Api/FrameworkController.cs
```diff
@@ -385,18 +385,14 @@ private static void AddSetting(TNode settingsNode, string name, object value)
             TNode setting = new TNode("setting");
             setting.AddAttribute("name", name);
 
-            if (value != null)
+            if (value is IDictionary dict)
```
stevenaw, line 388: The only not wholly trivial change here. Since this was inside a null check but `is` also performs a null check I removed the outer null check to reduce nesting.

src/NUnitFramework/framework/Internal/AwaitAdapter.cs
```diff
@@ -37,8 +37,7 @@ public static AwaitAdapter FromAwaitable(object awaitable)
             // special quality of blocking until complete in GetResult.
             // As long as the pattern-based adapters are reflection-based, this
             // is much more efficient as well.
-            var task = awaitable as System.Threading.Tasks.Task;
-            if (task != null) return TaskAwaitAdapter.Create(task);
+            if (awaitable is System.Threading.Tasks.Task task) return TaskAwaitAdapter.Create(task);
```
manfred-brands, line 40: I would prefer the `return` statement to be on the next line, but that is outside your scope.

manfred-brands (approved):
Nice work.
More readable and hopefully faster as it doesn't have to do the same work twice.

src/NUnitFramework/framework/Internal/AwaitAdapter.cs
mikkelbu: I agree with @manfred-brands. I sometimes overlook `return´ statements when they are on the same line.

mikkelbu (approved):

src/NUnitFramework/framework/Internal/AwaitAdapter.cs
stevenaw: Thanks! Makes sense to me. Updated

mikkelbu (approved):
LGTM. Thanks for the PR @stevenaw

PR #4097, opened 2022-04-15
3.13.3 Regression - CollectionTally correctly handles ImmutableDictionary again

### LINKED ISSUE #4095: Is.SupersetOf and Is.SubsetOf no longer work with IImmmutableDictionary<TKey,TValue> in NUnit 3.13.3
The following test passes in 3.13.2, but it fails in 3.13.3:
```csharp
using System.Collections.Immutable;
using NUnit.Framework;

namespace testimmutabledictionary;

[TestFixture]
public class Class1
{
    [Test]
    public void TestImmutableDictionary()
    {
        var numbers = Enumerable.Range(1, 3);
        var test1 = numbers.ToImmutableDictionary(t => t);
        var test2 = numbers.ToImmutableDictionary(t => t);

        Assert.That(test1, Is.SubsetOf(test2));
    }
}
```

Output:
```
  Failed TestImmutableDictionary [54 ms]
  Error Message:
     Expected: subset of < [1, 1], [2, 2], [3, 3] >
  But was:  < [1, 1], [2, 2], [3, 3] >
  Extra items: < [1, 1], [2, 2], [3, 3] >

  Stack Trace:
     at testimmutabledictionary.Class1.TestImmutableDictionary() in C:\danl\testimmutabledictionary\Class1.cs:line 16


Failed!  - Failed:     1, Passed:     0, Skipped:     0, Total:     1, Duration: 54 ms - testimmutabledictionary.dll (net6.0)
```

I notice that if both calls are instead `ToDictionary`, the test passes, but if either one is `ToImmmutableDictionary`, it fails.
I also noticed that using any of `ToImmutableList`, `ToImmutableArray`, or `ToImmutableHashSet` works (even when paired with non-immutable counterparts), so this seems limited to immutable dictionaries in particular.

project files attached:
[testimmutabledictionary.zip](https://github.com/nunit/nunit/files/8454051/testimmutabledictionary.zip)

stevenaw:
Fixes #4095

Allows CollectionTally to work with ImmutableDictionary properly again by avoiding a bulk-add for the ICollection case

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionTally.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionTally.cs
@@ -170,10 +170,8 @@ namespace NUnit.Framework.Constraints
 
         private static ArrayList ToArrayList(IEnumerable items)
         {
-            if (items is ICollection ic)
-                return new ArrayList(ic);
+            var list = items is ICollection ic ? new ArrayList(ic.Count) : new ArrayList();
 
-            var list = new ArrayList();
             foreach (object o in items)
                 list.Add(o);
 
--- a/src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintTests.cs
@@ -24,6 +24,9 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+#if !(NET35 || NET40)
+using System.Collections.Immutable;
+#endif
 using System.Collections.Specialized;
 using System.Diagnostics;
 using System.Linq;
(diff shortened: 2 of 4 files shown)


src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -170,10 +170,8 @@ void TryRemoveSlow(IEnumerable c)
 
         private static ArrayList ToArrayList(IEnumerable items)
         {
-            if (items is ICollection ic)
-                return new ArrayList(ic);
+            var list = items is ICollection ic ? new ArrayList(ic.Count) : new ArrayList();
```
stevenaw, line 173: `ImmutableDictionary.CopyTo` will convert the underlying type which is why the ctor which took in an ICollection was failing, but we can at least initialize the ArrayList to the proper size if it is known

manfred-brands (approved):
LGTM

mikkelbu (approved):
LGTM.

PR #4113, opened 2022-05-05
WIP: Attempt on Generic Numerical EqualConstraint

manfred-brands:
Various unit tests fail.


stevenaw:
@manfred-brands Are you open to a PR into this branch with some contributing fixes, along with some resolved conflicts? I was able to find a bit of time to look at this today and have it down to 2 failures.
If you're open to collaborating, what are your thoughts on us using a shared branch within the NUnit repo itself?

manfred-brands:
@stevenaw I sure am. Every time I got stuck. So collaboration sound great.

stevenaw:
Thanks @manfred-brands . This is indeed a tricky problem to solve, thanks for having gotten it started.

I've opened up https://github.com/nunit/nunit/pull/4456 which I think you should also be able to push or PR to. I'll close this PR then if you're alright with that.

PR #4114, opened 2022-05-06
Improve XML documentation for `TestCaseSourceAttribute`

### LINKED ISSUE #4112: Update documentation to clarify passing parameters to test cases
I am using NUnit 3.13.3 and NUnit3TestAdapter 4.2.1 with .NET Framework 4.7.2 and Visual Studio 2019.

I am not able to pass arrays or lists. I have the method return type set to an array/list, and so is the return value and the parameter type. When I try, I get one of these error messages:

```
MyMethodTest(MyType) Failed	60 ms
System.ArgumentException : Object of type 'MyType' cannot be converted to type 'MyType[]'.

MyMethodTest(MyType) Failed	60 ms
System.ArgumentException : Object of type 'MyType' cannot be converted to type 'System.Collections.Generic.List`1[MyType]'.
```

I also tried setting the return type to `TestCaseData` and wrapping in `new TestCaseData(retval)` but that doesn't work either. The only solution I have found is to set the return type to `TestCaseData[]` and wrap like this:

```cs
return new TestCaseData[] { new TestCaseData(retval) };
```

This issue seems to be the same as https://github.com/nunit/nunit/issues/1327 and it looks like there has been two rounds of attempted fixes but it still doesn't work.

aaronfranke:
Closes #4112 

This is similar to some of the content on this page, but with some of the wording changed for brevity: https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcasesource.html

I also hit the automatic format button and fixed the indentation of an `#endregion` to match the others in the file.

Thanks to @mikkelbu for the help.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
@@ -15,6 +15,12 @@ namespace NUnit.Framework
 {
     /// <summary>
     /// Indicates the source to be used to provide test fixture instances for a test class.
+    /// <list>
+    /// <listheader>The name parameter is a <see cref="string"/> representing the name of the source used to provide test cases. It has the following characteristics:</listheader>
+    /// <item>It must be a static field, property, or method in the same class as the test case.</item>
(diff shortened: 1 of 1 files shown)


aaronfranke:
@mikkelbu Can you review this?

PR #4118, opened 2022-05-28
Ensure consistent messages for TimeoutCommand for both .NETFramework and .NETCore

### LINKED ISSUE #3859: Proper test result exception message when hitting `TimeoutAttribute`

manfred-brands:
Fixes #3859

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
@@ -101,7 +101,11 @@ namespace NUnit.Framework.Internal.Commands
                 // If the timer cancelled the current thread, change the result
                 if (_commandTimedOut)
                 {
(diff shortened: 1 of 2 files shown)


stevenaw (approved):
Thanks again @manfred-brands ! Would you be able to setup a PR to main too?

manfred-brands:
> Thanks again @manfred-brands ! Would you be able to setup a PR to main too?

Done as #4122

PR #4120, opened 2022-06-01
Enumerate enumerables to determine emptiness

### LINKED ISSUE #4111: `Is.Empty` constraint on complex collections might hide bugs
With the recent update to NUnit 3.13.3, one of our unit tests started failing. The test failed because of a "mistake" in one of our collections. We build a *soft deleting* collection with a restore functionality. You can imagine the collection looking like this:

```csharp
public class SoftDeletingCollection<T> : IEnumerable<T>
{
    private readonly List<Deletable<T>> _inner = new();

    public void Add(T item)
    {
        _inner.Add(new Deletable<T>(item, false));
    }

    // Remove only soft deletes items so that they can be restored later
    public void Remove(T item)
    {
        var deletable = _inner.Single(x => x.Represents(item));
        deletable.IsDeleted = true;
    }

    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
    public IEnumerator<T> GetEnumerator() => new SoftDeletingEnumerator<T>(_inner.GetEnumerator());
}
```
(the real remove logic is more complex, but the shown one is sufficient for my point)

The `Deletable<T>` class is simply a container holding the value and `IsDeleted` flag:

```csharp
private class Deletable<T>
{
    public Deletable(T Value, bool IsDeleted)
    {
        this.Value = Value;
        this.IsDeleted = IsDeleted;
    }

    public T Value { get; }
    public bool IsDeleted { get; set; }

    public bool Represents(T value) => EqualityComparer<T>.Default.Equals(Value, value);
}
```

The `SoftDeletingEnumerator<T>` is crucial since it is responsible for hiding deleted members during enumeration:

```csharp
private class SoftDeletingEnumerator<T> : IEnumerator<T>
{
    private readonly IEnumerator<Deletable<T>> _inner;
    public SoftDeletingEnumerator(IEnumerator<Deletable<T>> inner) => _inner = inner;

    public bool MoveNext()
    {
        // This while loop skips deleted items during enumeration
        while (_inner.MoveNext())
            if (!_inner.Current.IsDeleted) return true;
        return false;
    }

    public T Current => _inner.Current.Value;
    object IEnumerator.Current => Current;
    public void Reset() => _inner.Reset();
    public void Dispose() { }
}
```

Now I test that the `SoftDeletingCollection<T>` `Is.Empty` after I remove the last item:

```
[Test]
public void Test_without_count_property()
{
    var instance = new SoftDeletingCollection<int> { 1 };
    instance.Remove(1);
    Assert.That(instance, Is.Empty);    // Passes
}
```

However, if we now introduce a `Count` property to the collection (e.g., to implement `IReadOnlyCollection<T>` instead of `IEnumerable<T>`)  and make a small mistake such as forgetting to deduce *deleted* items

```csharp
public class SoftDeletingCollection<T> : IEnumerable<T>
{
    ...
    public int Count => _inner.Count; // this ingores the soft deletion
    ...
}
```

the test fails with the message

```
Expected: <empty>
  But was:  <empty>
```

which is **very confusing**.

In our particular case, we had a collection such as `SoftDeletingCollection` and the test was passing until NUnit `3.13.3`, so to figure out what was the problem was quite a journey.

But the worst part about this is that we could make the test pass and still have a bug in our collection by **fixing** the `Count` implementation (i.e., `... Count => _inner.Where(x => !x.IsDeleted).Count();`) but breaking the `SoftDeletingEnumerator`

```csharp
private class SoftDeletingEnumerator<T> : IEnumerator<T>
{
    ...
    public bool MoveNext()
    {
        return _inner.MoveNext();    // this will enumerate deleted items as well
    }
    ...
}
```

---

I think that `Is.Empty` should rather check the enumeration than the `Count` property. To evaluate the `Count` the `Has.Count` expression is far more explicit. In the case described above, the new behavior of `Is.Empty` will lead to false positive test results and bugs in production code.

It might be valuable to *include* a check of the `Count` in the `Is.Empty` constraint, but I am not convinced that `Count` should be the deciding property for emptiness.

Crown0815:
The usage of the `Count` property can lead to unexpected results
if the `Count` is not in sync with the enumeration.

It makes sense to use the `Count` on a `ICollection` since the interface
clearly connects the `Count` with the enumeration. On any `IEnumerable`,
this assumption is not always correct, though.

fixes #4111

### DIFF
--- a/src/NUnitFramework/framework/Constraints/EmptyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/EmptyConstraint.cs
@@ -31,7 +31,7 @@ namespace NUnit.Framework.Constraints
         {
             // NOTE: actual is string will fail for a null typed as string
             Type actualType = actual?.GetType() ?? typeof(TActual);
-            
+
             if (actualType == typeof(string))
                 realConstraint = new EmptyStringConstraint();
             else if (actual is Guid || actualType == typeof(Guid?))
@@ -40,10 +40,10 @@ namespace NUnit.Framework.Constraints
                 realConstraint = new EmptyDirectoryConstraint();
             else if (actual is System.Collections.ICollection)
                 realConstraint = new EmptyCollectionConstraint();       // Uses ICollecion.Count
-            else if (CountZeroConstraint.HasCountProperty(actualType))  // For Collections that have Count but are not ICollection
-                realConstraint = new CountZeroConstraint();
             else if (actual is System.Collections.IEnumerable)          // Enumerates whole collection
                 realConstraint = new EmptyCollectionConstraint();
+            else if (CountZeroConstraint.HasCountProperty(actualType))  // For Collections that have Count but are not ICollection
+                realConstraint = new CountZeroConstraint();
             else
                 throw new ArgumentException($"The actual value must be not-null, a string, Guid, have an int Count property, IEnumerable or DirectoryInfo. The value passed was of type {actualType}.", nameof(actual));
 
--- a/src/NUnitFramework/tests/Constraints/EmptyConstraintTest.cs
+++ b/src/NUnitFramework/tests/Constraints/EmptyConstraintTest.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.IO;
 using NUnit.TestUtilities;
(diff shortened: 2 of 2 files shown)


manfred-brands (approved):
Thanks @Crown0815, changes look good with a new test case added.

stevenaw (approved):
LGTM, thanks for your contribution @Crown0815 !

PR #4127, opened 2022-06-11
Update codebase to use Array.Empty<T>

### LINKED ISSUE #4051: Update codebase to use Array.Empty<T>
After #4050 merges and #4036 resolves, net462 will be the minimum target framework. We should update the code in the main branch to use Array.Empty<T> after this. A complimentary framework-focused PR + ticket were created to update the v3.13 branch previously in #4041 and #4042

mjedrzejek:
Update codebase to use Array.Empty<T>

Fixes #4051

### DIFF
--- a/src/NUnitFramework/tests/Assertions/ConditionAssertTests.cs
+++ b/src/NUnitFramework/tests/Assertions/ConditionAssertTests.cs
@@ -130,8 +130,8 @@ namespace NUnit.Framework.Assertions
         public void IsEmpty()
         {
             Assert.IsEmpty( "", "Failed on empty String" );
-            Assert.IsEmpty( new int[0], "Failed on empty Array" );
-            Assert.IsEmpty((IEnumerable)new int[0], "Failed on empty IEnumerable");
+            Assert.IsEmpty(Array.Empty<int>(), "Failed on empty Array" );
+            Assert.IsEmpty((IEnumerable)Array.Empty<int>(), "Failed on empty IEnumerable");
 
             Assert.IsEmpty( new ArrayList(), "Failed on empty ArrayList" );
             Assert.IsEmpty( new Hashtable(), "Failed on empty Hashtable" );
@@ -210,7 +210,7 @@ namespace NUnit.Framework.Assertions
             var expectedMessage =
                 "  Expected: not <empty>" + Environment.NewLine +
                 "  But was:  <empty>" + Environment.NewLine;
-            var ex = Assert.Throws<AssertionException>(() => Assert.IsNotEmpty( new int[0] ));
+            var ex = Assert.Throws<AssertionException>(() => Assert.IsNotEmpty( Array.Empty<int>() ));
             Assert.That(ex.Message, Is.EqualTo(expectedMessage));
         }
 
@@ -220,7 +220,7 @@ namespace NUnit.Framework.Assertions
             var expectedMessage =
                 "  Expected: not <empty>" + Environment.NewLine +
                 "  But was:  <empty>" + Environment.NewLine;
-            var ex = Assert.Throws<AssertionException>(() => Assert.IsNotEmpty((IEnumerable)new int[0]));
+            var ex = Assert.Throws<AssertionException>(() => Assert.IsNotEmpty((IEnumerable)Array.Empty<int>()));
             Assert.That(ex.Message, Is.EqualTo(expectedMessage));
         }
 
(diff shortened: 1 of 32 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=4127) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=4127) before we can accept your contribution.<br/><br/>:x: mjedrzejek [sign now](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=4127)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/nunit/nunit?pullRequest=4127) it.</sub>

304NotModified (approved):

manfred-brands (changes requested):
This is only a subset of the zero element array allocations.
There are many more.
You can find them with a regular expression: `new [a-zA-Z_0-9]*\[0\]`

mjedrzejek:
@manfred-brands thanks!
I used the regular expression and added more changes to the MR. I was not able to replace 5 occurrences where empty array is created as a parameter for attribute - if I try to change the code there, I am getting a compiler error and I was not able to find a workaround (error: CS0182 - 
An attribute argument must be a constant expression, typeof expression or array creation expression of an attribute parameter type).

src/NUnitFramework/tests/Assertions/ConditionAssertTests.cs
```diff
@@ -220,7 +220,7 @@ public void IsNotEmptyFailsOnEmptyIEnumerable()
             var expectedMessage =
                 "  Expected: not <empty>" + Environment.NewLine +
                 "  But was:  <empty>" + Environment.NewLine;
-            var ex = Assert.Throws<AssertionException>(() => Assert.IsNotEmpty((IEnumerable)new int[0]));
+            var ex = Assert.Throws<AssertionException>(() => Assert.IsNotEmpty((IEnumerable)Array.Empty<int>()));
```
manfred-brands, line 223: Even though your change is correct, a better version here would be:

```suggestion
            var ex = Assert.Throws<AssertionException>(() => Assert.IsNotEmpty(Enumerable.Empty<int>()));
```

You will also have to add a `using System.Linq;` to the file.

src/NUnitFramework/tests/Assertions/ConditionAssertTests.cs
```diff
@@ -130,8 +130,8 @@ public void IsNaNFails()
         public void IsEmpty()
         {
             Assert.IsEmpty( "", "Failed on empty String" );
-            Assert.IsEmpty( new int[0], "Failed on empty Array" );
-            Assert.IsEmpty((IEnumerable)new int[0], "Failed on empty IEnumerable");
+            Assert.IsEmpty(Array.Empty<int>(), "Failed on empty Array" );
+            Assert.IsEmpty((IEnumerable)Array.Empty<int>(), "Failed on empty IEnumerable");
```
manfred-brands, line 134: The intent of the test would make the below a better version:
```suggestion
            Assert.IsEmpty(Enumerable.Empty<int>(), "Failed on empty IEnumerable");
```

manfred-brands (changes requested):
Technically the conversion are done correctly.
Looking at the code, two of them should actually be `Enumerable.Empty<int>()`.
Please update those two please.

manfred-brands:
For next time, note that there is a 'Quickfix' for CA1825, that does the simple conversion for you.
It also know that you cannot convert the ones in attributes.

Go to one place where there was an empty array allocation and select `Ctrl+.` or select the lightbulb.
image

Unfortunately that quickfix doesn't add `using` statements, nor does the `Enumerable.Empty` conversion.

src/NUnitFramework/tests/Assertions/ConditionAssertTests.cs
mjedrzejek: Done

src/NUnitFramework/tests/Assertions/ConditionAssertTests.cs
mjedrzejek: Done

mikkelbu (approved):
LGTM. Thanks for the work @mjedrzejek :+1

PR #4137, opened 2022-06-25
Cache method discovery main

manfred-brands:
@stevenaw I cannot push to your branch, so here are my commits for your PR.


stevenaw:
Thanks for your time and attention to this @manfred-brands ! I really appreciate it. I hadn't realized behaviour between v3 + v4 had started to diverge so much from already.

I'm not sure the best way to allow multiple people to contribute to the same branch, but PRing like this into a common one seems to have worked well here

Issue #4124, opened 2022-06-10
TestContext to implement IProgress<T> interface


vpenades:
My proposal is that TestContext to implement these interfaces:

```c#
class TestContext: IProgress<int>, IProgress<string>
{
    void IProgress<int>.Report(int value) { TestExecutionContext.CurrentContext.ProgressWriter.WriteLine("{value}%"); }

    void IProgress<string>.Report(string value) { TestExecutionContext.CurrentContext.OutWriter.WriteLine(value); }
}

```

Reasoning:

I have several methods that expect an IProgress<int> and | or IProgress<string> argument, like this:

```c#
public static void DoSomeBackgroundWork(IProgress<int> percentProgress, IProgress<string> statusProgress);
```

So I have to write a wrapper class to redirect IProgress<T> to TestContext.CurrentContext

Since `System.IProgress<T>` is a BCL type, it can help bridge these use cases without requiring cross referencing assemblies.

If there's interest I could submit a PR.

stevenaw:
I like this idea. It seems minimally intrusive yet provides a standard way to hook into incremental progress updates.

stevenaw:
@nunit/framework-team what are your thoughts?

rprouse:
I don't hate the idea, but adding it using the string and int interfaces is rather arbitrary as is treating the int version as a percentage. The next user might want an int progress to indicate the number of steps not a percentage. 

In other words, the proposed solution would solve your problem but not in a generic enough way to be useful to most users. I think that writing the wrapper class for your needs is appropriate.

Sorry about the confusion about the process and the PR. Thanks for the suggestion, but I'm going to close this as won't do.

PR #4139, opened 2022-06-27
Added IProgress<> interfaces

vpenades:
Following #4124, adding this interfaces allows previosly existing APIs that support IProgress<T> to interact with nunit seamlessly


rprouse:
Closing as won't do. See #4124

Issue #4149, opened 2022-08-04
Distribute optimized framework builds with easy debugging


stevenaw:
As noted in https://github.com/nunit/nunit/pull/4035#issuecomment-1200160651
In https://github.com/nunit/nunit/pull/3464 we stopped using sourcelink in favour of distributing a debug build of the framework with embedded symbols + source, however it seems this can make some operations noticeably slower in larger suites. I'm wondering what thoughts or options there are to distribute an optimized build while still allow a good debugging experience.

@nunit/framework-team @nunit/core-team Thoughts?

jnm2:
I'm happy to go back as soon as there is evidence we can share to back this up. When @sharwell switched us to publishing debug builds, we simply didn't have that evidence yet.

stevenaw:
@lahma are you able to share any of the numbers you saw in #4035 ?

lahma:
@stevenaw I'll put this on my TODO list, need to dig up my test suites for benchmarking test discovery etc...

lahma:
OK here are the results. Note I can offer a PR if you want basic benchmark project setup. This relies on having esprima-dotnet's large test suite built and ready as DLL. This is the runner code I extracted by debugging what Rider IDE was doing and sending.

The only change in running the tests is that I flip the value of `nunit.framework.csproj` attribute `<Optimize>false</Optimize>` value between true and false.

Testing done in `v13.3-dev` branch, under NET 6 runtime.

## Benchmark

* __CountTestsEmptyFilter__: 60737 matches
* __CountTestsWithCategoryFilter__: 2307 matches
* __CountTestsWithOrFilterAll__: 60737 matches (the IDE scenario where IDE runner sends all ids which was optimized with OrFilter)
* __LoadTests__: 341 results

```csharp
using System.Collections;
using BenchmarkDotNet.Attributes;
using NUnit.Framework.Api;
using NUnit.Framework.Interfaces;
using NUnit.Framework.Internal;
using NUnit.Framework.Internal.Filters;

namespace NUnit.Framework;

[InProcess] // need to use so that we get correct test loading
[MemoryDiagnoser]
public class FrameworkControllerBenchmark
{
    private FrameworkController _controller;
    private List<string> _testIds;
    private string _orFilter;

    [GlobalSetup]
    public void GlobalSetup()
    {
        const string workDirectory = "C:\\Work\\jint\\Jint.Tests.Test262\\bin\\Release\\net6.0";
        var settings = new Hashtable
        {
[...]

PR #4151, opened 2022-08-07
Add benchmark project

lahma:
[Per discussion](https://github.com/nunit/nunit/issues/4149#issuecomment-1207303422), here's and PR offer/sample how to run benchmarks in NUnit code base. I think this might be a worthy addition as it allows contributors to drop their benchmarks into this new project and validate improvements with hard numbers.

Includes one benchmark for `FrameworkController` that works against NUnit's own test suite.

/cc @stevenaw

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -189,3 +189,4 @@ project.lock.json
 *.out
 .ionide
 .fake
+BenchmarkDotNet.Artifacts*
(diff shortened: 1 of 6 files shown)


rprouse (approved):
I pulled the code and ran the benchmarks. I like it 👍🏻

stevenaw (approved):
My thanks as well @lahma for setting this up. Sorry it's taken a while to get this merged.
This will provide a great basis to allow for more performance-minded changes in the future

PR #4175, opened 2022-09-13
enable nullable in DefaultTestAssemblyBuilder

SimonCropp:


src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -117,34 +117,36 @@ private TestSuite Build(Assembly assembly, string assemblyNameOrPath, IDictionar
...
                     // Deserializes the way old runners understand.
-                    
-                    options.TryGetValue(FrameworkPackageSettings.TestParameters, out object testParameters);
-                    string parametersString = testParameters as string;
-                    if (!string.IsNullOrEmpty(parametersString))
+
+                    if (options.TryGetValue(FrameworkPackageSettings.TestParameters, out var testParameters))
                     {
-                        foreach (string param in parametersString.Split(new[] { ';' }))
+                        var parametersString = (string)testParameters;
```
manfred-brands, line 123: If this can be null, you need a `?` in the type name.
```suggestion
                        var parametersString = (string?)testParameters;
```
Or if the parameter might not be a string at all, keep the original code, but change the type:
```csharp
                        string? parametersString = testParameters as string;
```

manfred-brands (changes requested):
One small remark.

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
SimonCropp: if it isnt a string, shouldnt we throw an exception?

SimonCropp: side note. where is this feature documented?

manfred-brands: The original code didn't throw.

I don't know if it is documented, I couldn't find any setters. Some comments indicate it is for backwards compatibility of old runners.

SimonCropp:
remark addressed

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -27,8 +29,6 @@ public class DefaultTestAssemblyBuilder : ITestAssemblyBuilder
         /// </summary>
         readonly ISuiteBuilder _defaultSuiteBuilder;
 
-        private PreFilter _filter;
```
stevenaw, line 30: Thanks for reworking this a bit so that we could keep this non-nullable

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -117,34 +117,36 @@ private TestSuite Build(Assembly assembly, string assemblyNameOrPath, IDictionar
...
+                    if (options.TryGetValue(FrameworkPackageSettings.TestParameters, out var testParameters))
                     {
-                        foreach (string param in parametersString.Split(new[] { ';' }))
+                        var parametersString = (string?)testParameters;
+                        if (!string.IsNullOrEmpty(parametersString))
                         {
-                            int eq = param.IndexOf('=');
-                        
-                            if (eq > 0 && eq < param.Length - 1)
+                            foreach (var param in parametersString!.Split(';'))
```
stevenaw, line 126: I'm unsure, but I don't think this will route to the `Split(char, StringSplitOptions = None)` overload one might expect, since that's only on NETStandard2.1+. I think it will still route to the params overload. Though I like how this allows the codebase to automatically target the newer API if a netstandard21 or higher version of the framework were ever released

stevenaw (approved):
Thanks for this contribution @SimonCropp ! This looks good to me but I'd appreciate another pair of eyes

rprouse (approved):

PR #4176, opened 2022-09-13
FrameworkController.Settings cant be null

SimonCropp:
(i could be mistaken, but) I dont think FrameworkController.Settings can ever be null. it is always initialized in the constructor


SimonCropp:
can we get a merge?

stevenaw (approved):
No concerns here, thanks @SimonCropp !

SimonCropp:
@stevenaw thanks

PR #4178, opened 2022-09-13
enable nullable for RuntimeFramework

SimonCropp:


src/NUnitFramework/framework/Internal/RuntimeFramework.cs
```diff
@@ -226,21 +219,18 @@ public static RuntimeFramework CurrentFramework
         /// <summary>
         /// The framework version for this runtime framework
         /// </summary>
-        public Version FrameworkVersion { get; private set; }
+        public Version FrameworkVersion { get; private set; } = null!;
```
manfred-brands, line 222: This is pure lying, saying that null is not null.

There are three ways to solve this issue:
1.  Suppress the message in the constructor:
```csharp
#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
        public RuntimeFramework( RuntimeType runtime, Version version)
#pragma warning restore CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
```

2. Do it declarative way, by adding `MemberNotNull` attributes to the two `InitFrom... Version` methods. This however requires include a reference to the `nullable` nuget package and increase the `LangVersion` to 9.0.
```csharp
        [MemberNotNull(nameof(FrameworkVersion))]
        [MemberNotNull(nameof(ClrVersion))]
        private void InitFromFrameworkVersion(Version version)
```

3.  Rewrite to code so that the two properties are set in the constructor itself.
```csharp
            (FrameworkVersion, ClrVersion) = version.Build < 0 ?
                InitFromFrameworkVersion(version) :
                InitFromClrVersion(version);
```
With a matching declaration and updated code for:
```csharp
        private (Version frameworkVersion, Version clrVersion) InitFromFrameworkVersion(Version version)
        {
            Version frameworkVersion, clrVersion;

            frameworkVersion = clrVersion = version;
```
etc.

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
```diff
@@ -226,21 +219,18 @@ public static RuntimeFramework CurrentFramework
...
 
         /// <summary>
         /// The CLR version for this runtime framework
         /// </summary>
-        public Version ClrVersion { get; private set; }
+        public Version ClrVersion { get; private set; } = null!;
```
manfred-brands, line 227: Same remark.

manfred-brands (changes requested):
Fix the non-initialized warning in one of the three suggested ways,

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
SimonCropp: updated with option 2

src/NUnitFramework/framework/Compatibility/System.Diagnostics.CodeAnalysis/NullableAttributes.cs
```diff
@@ -1,110 +0,0 @@
-// Source: https://github.com/dotnet/runtime/blob/4c71c6cf112d5cdfe5d39485505f1b2991c6e0d5/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/NullableAttributes.cs
```
stevenaw, line 1: Thanks for noticing that this could be cleaned up!

src/NUnitFramework/framework/Internal/RuntimeFramework.cs
```diff
@@ -210,13 +214,7 @@ private void InitFromClrVersion(Version version)
...
         /// for the framework that is currently in use.
         /// </summary>
-        public static RuntimeFramework CurrentFramework
-        {
-            get
-            {
-                return currentFramework.Value;
-            }
-        }
+        public static RuntimeFramework CurrentFramework => currentFramework.Value;
```
stevenaw, line 217: Thanks for the modernization here too

src/NUnitFramework/framework/nunit.framework.csproj
```diff
@@ -8,9 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" />
-  </ItemGroup>
-
-  <ItemGroup>
+    <PackageReference Include="Nullable" Version="1.3.1" PrivateAssets="all" />
```
stevenaw, line 11: Oh, now I understand. We've swapped out our own "shape" and copy for this. I'm unsure what the decision process is for taking on new dependencies, but I do like your inclusion of `PrivateAssets` here.
@jnm2 I think you did the initial pass in making some things nullable. Can I ask you your thoughts on this?

rprouse: Ah, this one confused me. I had to go out to the NuGet package and read the description to understand, "This package does not contain any compiled binaries, but instead adds the attribute classes as C# source code to your project." A note as such in the PR would have been helpful.

rprouse (approved):

PR #4183, opened 2022-09-13
enable nullable for MessagePumpStrategy

SimonCropp:


manfred-brands (approved):
Not a fan of the '!' operator on the GetMethod calls.
But let us assume that MS won't rename/remove those methods.

manfred-brands:
Yes that is a good suggestion.

src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
```diff
@@ -171,9 +173,7 @@ private sealed class SingleThreadedTestMessagePumpStrategy : MessagePumpStrategy
 
             public override void WaitForCompletion(AwaitAdapter awaiter)
             {
-                var context = SynchronizationContext.Current as SingleThreadedTestSynchronizationContext;
-
-                if (context == null)
+                if (!(SynchronizationContext.Current is SingleThreadedTestSynchronizationContext context))
                     throw new InvalidOperationException("This strategy must only be used from a SingleThreadedTestSynchronizationContext.");
```
jnm2, line 177: My preference
```suggestion
var context = SynchronizationContext.Current as SingleThreadedTestSynchronizationContext
    ?? throw new InvalidOperationException("This strategy must only be used from a SingleThreadedTestSynchronizationContext.");
```

Or, consider `is not` instead of `!(`

jnm2 (approved):

jnm2:
> Not a fan of the '!' operator on the GetMethod calls. But let us assume that MS won't rename/remove those methods.

I am actually a fan of this and have been using it with the reflection API since C# 8. I think this is dead center in what the `!` operator was intended for.

SimonCropp:
can we get a ruling on `!`?

jnm2:
@manfred-brands I'm not against making helpers that throw with nice messages instead of returning null, but I wouldn't personally do this in place of using `!` every time I use `!`, and I don't see reflection as a special case. So I'm not thrilled about maintaining new helpers. What do you think?

manfred-brands:
@jnm2 I'm ok with leaving as it is here. 

Taking it on a case by case basis. 

I have seen code where developers were doing it to silence the compiler even though it clearly was possible for those variables to be null.

PR #4198, opened 2022-10-01
add some missing process dispose

SimonCropp:


jnm2:
For some reason I thought Process.GetCurrentProcess() didn't need to be disposed. The .NET API may differ in this regard from the Windows API itself [which says](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentprocess#remarks):

> The pseudo handle need not be closed when it is no longer needed. Calling the [CloseHandle](https://learn.microsoft.com/en-us/windows/desktop/api/handleapi/nf-handleapi-closehandle) function with a pseudo handle has no effect. If the pseudo handle is duplicated by [DuplicateHandle](https://learn.microsoft.com/en-us/windows/desktop/api/handleapi/nf-handleapi-duplicatehandle), the duplicate handle must be closed.

jnm2 (approved):

mikkelbu (approved):
LGTM

PR #4199, opened 2022-10-01
avoid duplicate dictionary lookups

SimonCropp:
minor perf improvement, and IMO improves readability

### DIFF
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -124,8 +124,8 @@ namespace NUnit.Framework.Api
                 }
                 else
                 {
-                    var workDirectory = Settings.TryGetValue(FrameworkPackageSettings.WorkDirectory, out var workDirectoryValue) ?
-                        (string)workDirectoryValue :
+                    var workDirectory = Settings.TryGetValue(FrameworkPackageSettings.WorkDirectory, out var workDirectoryValue)
+                        ? (string)workDirectoryValue :
                         Directory.GetCurrentDirectory();
                     using var process = Process.GetCurrentProcess();
                     var id = process.Id;
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -114,17 +114,19 @@ namespace NUnit.Framework.Api
             var newSettings = settings as IDictionary<string, object>;
             Settings = newSettings ?? settings.Cast<DictionaryEntry>().ToDictionary(de => (string)de.Key, de => de.Value);
 
-            if (Settings.ContainsKey(FrameworkPackageSettings.InternalTraceLevel))
+            if (Settings.TryGetValue(FrameworkPackageSettings.InternalTraceLevel, out var traceLevelValue))
             {
-                var traceLevel = (InternalTraceLevel)Enum.Parse(typeof(InternalTraceLevel), (string)Settings[FrameworkPackageSettings.InternalTraceLevel], true);
+                var traceLevel = (InternalTraceLevel)Enum.Parse(typeof(InternalTraceLevel), (string)traceLevelValue, true);
 
-                if (Settings.ContainsKey(FrameworkPackageSettings.InternalTraceWriter))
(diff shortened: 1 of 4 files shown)


src/NUnitFramework/framework/Api/FrameworkController.cs
```diff
@@ -114,17 +114,19 @@ private void Initialize(string assemblyNameOrPath, IDictionary settings)
...
+                    InternalTrace.Initialize((TextWriter)textWriterValue, traceLevel);
+                }
                 else
                 {
-                    var workDirectory = Settings.ContainsKey(FrameworkPackageSettings.WorkDirectory)
-                        ? (string)Settings[FrameworkPackageSettings.WorkDirectory]
-                        : Directory.GetCurrentDirectory();
+                    var workDirectory = Settings.TryGetValue(FrameworkPackageSettings.WorkDirectory, out var workDirectoryValue) ?
+                        (string)workDirectoryValue :
+                        Directory.GetCurrentDirectory();
```
jnm2, line 129: It's harder for me to scan code when the operators aren't at the start of the line. Was this part of the change intentional?

jnm2 (approved):

src/NUnitFramework/framework/Api/FrameworkController.cs
SimonCropp: nope. will move to start

jnm2: Can you move the `:` back too?

src/NUnitFramework/framework/Api/FrameworkController.cs
```diff
@@ -114,17 +114,19 @@ private void Initialize(string assemblyNameOrPath, IDictionary settings)
...
+                    InternalTrace.Initialize((TextWriter)textWriterValue, traceLevel);
+                }
                 else
                 {
-                    var workDirectory = Settings.ContainsKey(FrameworkPackageSettings.WorkDirectory)
-                        ? (string)Settings[FrameworkPackageSettings.WorkDirectory]
-                        : Directory.GetCurrentDirectory();
+                    var workDirectory = Settings.TryGetValue(FrameworkPackageSettings.WorkDirectory, out var workDirectoryValue)
+                        ? (string)workDirectoryValue :
+                        Directory.GetCurrentDirectory();
```
mikkelbu, line 129:
```suggestion
                        ? (string)workDirectoryValue
                        : Directory.GetCurrentDirectory();
```

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -310,14 +315,14 @@ private void CreateTestExecutionContext(ITestListener listener)
...
-            if (Settings.ContainsKey(FrameworkPackageSettings.DefaultUICulture))
-                Context.CurrentUICulture = new CultureInfo((string)Settings[FrameworkPackageSettings.DefaultUICulture], false);
-            if (Settings.ContainsKey(FrameworkPackageSettings.StopOnError))
-                Context.StopOnError = (bool)Settings[FrameworkPackageSettings.StopOnError];
+            if (Settings.TryGetValue(FrameworkPackageSettings.DefaultTimeout, out var timeout))
+                Context.TestCaseTimeout = (int)timeout;
+            if (Settings.TryGetValue(FrameworkPackageSettings.DefaultCulture, out var culture))
+                Context.CurrentCulture = new((string)culture, false);
+            if (Settings.TryGetValue(FrameworkPackageSettings.DefaultUICulture, out var uiCulture))
+                Context.CurrentUICulture = new((string)uiCulture, false);
```
mikkelbu, line 323: I don't think we use these inferred constructors anywhere else (at least at the moement), and I find the code harder to read when there is no type on either side of the assignment, so I think we should keep the old constructors (at least for now).
```suggestion
                Context.CurrentCulture = new CultureInfo((string)culture, false);
            if (Settings.TryGetValue(FrameworkPackageSettings.DefaultUICulture, out var uiCulture))
                Context.CurrentUICulture = new CultureInfo((string)uiCulture, false);
```

mikkelbu (approved):
Looks good to me. I only have a couple of minor comments.

stevenaw (approved):

stevenaw:
LGTM too. Thanks @SimonCropp 👍

PR #4202, opened 2022-10-01
remove GetTypeInfo

SimonCropp:
it should no longer be required

### DIFF
--- a/src/NUnitFramework/framework/Extensions.cs
+++ b/src/NUnitFramework/framework/Extensions.cs
@@ -3,7 +3,6 @@
 using System;
 using System.Collections;
 using System.Reflection;
-using NUnit.Compatibility;
 
 namespace NUnit.Framework
 {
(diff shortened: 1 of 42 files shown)


stevenaw (approved):

SimonCropp:
can we get a merge here?

src/NUnitFramework/framework/Extensions.cs
```diff
@@ -39,7 +39,7 @@ public static bool HasAttribute<T>(this Type type, bool inherit)
 
         public static T[] GetAttributes<T>(this Type type, bool inherit) where T : class
         {
-            return ((ICustomAttributeProvider)type.GetTypeInfo()).GetAttributes<T>(inherit);
+            return ((ICustomAttributeProvider)type).GetAttributes<T>(inherit);
```
jnm2, line 42: Are these casts still required after removing GetTypeInfo?

src/NUnitFramework/framework/Internal/Tests/Test.cs
```diff
@@ -335,7 +335,7 @@ public void ApplyAttributesToTest(ICustomAttributeProvider provider)
         public void ApplyAttributesToTest(Type type)
         {
             foreach (var t in GetNestedTypes(type).Reverse()) 
-                ApplyAttributesToTest((ICustomAttributeProvider) t.GetTypeInfo());
+                ApplyAttributesToTest((ICustomAttributeProvider) t);
```
jnm2, line 338: This cast is probably showing as unnecessary in the IDE now?

src/NUnitFramework/framework/Extensions.cs
SimonCropp: still required since they force the path tp go into `this ICustomAttributeProvider attributeProvider` path

TBH the whole reflection codebase need some re-work. but i would rather attempt that in another pr

src/NUnitFramework/framework/Internal/Tests/Test.cs
SimonCropp: still required since it forces a different path

SimonCropp:
any more feedback on this one?

SimonCropp:
bump

stevenaw:
@jnm2 Thanks for your review earlier. Do you mind if I merge?

stevenaw:
I'm going to go ahead and merge this. This would be great to get this and a few others in before month's end.
Thanks once again for the contribution @SimonCropp

PR #4203, opened 2022-10-01
fix null reference in TestProgressReporter

SimonCropp:
so this one is a bug. every time TestProgressReporter is called from this path, a null ref would be thrown. it is hidden since all methods of TestProgressReporter do a catch

### DIFF
--- a/src/NUnitFramework/framework/Api/FrameworkController.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkController.cs
@@ -199,7 +199,7 @@ namespace NUnit.Framework.Api
         /// <returns>The XML result of the test run</returns>
         public string RunTests(string filter)
         {
-            TNode result = Runner.Run(new TestProgressReporter(null), TestFilter.FromXml(filter)).ToXml(true);
+            TNode result = Runner.Run(TestListener.NULL, TestFilter.FromXml(filter)).ToXml(true);
             return InsertChildElements(result).OuterXml;
         }
 


jnm2 (approved):

SimonCropp:
should this be assigned to the milestone?

jnm2:
I don't see recent examples of us putting PRs in milestones, but maybe we should start?

SimonCropp:
some examples 
image

SimonCropp:
although i think the "enable more nullable" should be rolled up into one issue in the milestone. having every nullable PR in the milestone might be too much noise? the nullable PRs could be linked the issue?

jnm2:
That makes sense. I'm not sure who looks at the milestone and when. @nunit/framework-team?

SimonCropp:
on my projects i use milestones to track what i want in the (eventual) release notes

stevenaw:
Personally, I like the idea, especially when there isn't a corresponding issue to track the change. Which, then, perhaps it's easiest then to just always do it to avoid having to think? I've probably missed doing it for 1 or 2 recent PRs as sometimes it's from my phone.

I think I've seen @rprouse  use  them for generating release notes, but it can also be a handy way to see as-yet-unreleased changes too.

SimonCropp:
here is my workflow for "milestone to release notes"

```
name: on-tag-do-release
on:
  push:
    tags:
      - '[0-9]+\.[0-9]+\.[0-9]+'
jobs:
  release:
    runs-on: windows-latest
    steps:
    - name: Run
      run: |
        dotnet tool install --global GitReleaseManager.Tool
        tag="${GITHUB_REF:10}"
        owner="${GITHUB_REPOSITORY%/*}"
        repo="${GITHUB_REPOSITORY#*/}"
        dotnet-gitreleasemanager create -m ${tag} --token ${{secrets.GITHUB_TOKEN}} -o ${owner} -r ${repo}
        dotnet-gitreleasemanager close -m ${tag} --token ${{secrets.GITHUB_TOKEN}} -o ${owner} -r ${repo}
        dotnet-gitreleasemanager publish -t ${tag} --token ${{secrets.GITHUB_TOKEN}} -o ${owner} -r ${repo}
      shell: bash
```

rprouse:
Thanks @SimonCropp, I like that. Currently I run a program that I wrote that pulls the issues and PRs marked with the milestone and creates the Markdown for the changelog, then do some hand editing to clean it up.

PR #4204, opened 2022-10-01
remove some redundant parameters

SimonCropp:


src/NUnitFramework/framework/Api/FrameworkController.cs
```diff
@@ -558,7 +553,7 @@ public class StopRunAction : FrameworkControllerAction
             /// <remarks>A forced stop will cause threads and processes to be killed as needed.</remarks>
             public StopRunAction(FrameworkController controller, bool force, object handler)
             {
-                controller.StopRun((ICallbackEventHandler)handler, force);
+                controller.StopRun(force);
```
jnm2, line 556: These changes won't change current behavior, though the mere existence of the `object handler` parameter makes me wonder if there's a bug in the current behavior.

SimonCropp: > wonder if there's a bug in the current behavior

yeah i thought the same. but i lack the history so if there is a bug, i could not see it

stevenaw: I got curious and dug into this. It looks like `FrameworkController.StopRun()` would at one point raise a callback event
https://github.com/nunit/nunit/commit/362f7a3f7074ff520ad811ba06883b3595aa0685#diff-a4de5abf6ba13245db808314f98df69cd318b068215f50932b0d50abdd938c43R203
```csharp
private void StopRun(ICallbackEventHandler handler, bool force, int timeout)
        {
            //handler.RaiseCallbackEvent(Runner.StopRun(force, timeout).ToString());
            Runner.StopRun(force);
            handler.RaiseCallbackEvent(Runner.WaitForCompletion(timeout).ToString());
        }
```
It looks like it was removed at some point many (> 7) years ago though I haven't dug further to an exact commit

stevenaw (approved):

CharliePoole:
When called from the .NET Framework, the API consists of a set of constructors. Since the constructors always succeed, the result of the action can only be returned by an event of some kind. That's why the handler is passed. This could be done differently nowadays, but the API dates back to .NET 1.1.

(Note that a different interface is used with .NET Core.)

At some time in the past, this return value apparently stopped being used for StopRun. That means we have no way of knowing whether StopRun succeeded. I have had to deal with that problem in both the GUI and the Console runner but I didn't know why. This may be the reason.

src/NUnitFramework/framework/Api/FrameworkController.cs
SimonCropp: ok i reverted this part of the PR given it needs more work

SimonCropp:
rebased

mikkelbu (approved):
LGTM. Thanks for the contribution @SimonCropp

PR #4205, opened 2022-10-01
Reduce the visibility of 'Numerics' to 'internal'

### LINKED ISSUE #3932: Make `Numerics` class internal
Most of NUnit's numeric-related work (comparisons, differences, equality) is exposed via constraints or the classic assertion model, but the work itself is done by a helper class: [Numerics](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Constraints/Numerics.cs).

This helper class has public visibility, which means API changes to it require extra caution to ensure we don't break anyone. It might be helpful to look at if we can make this an internal class. This would be a breaking change, but I suspect it shouldn't be very disruptive. I haven't seen any usages of it in public repos based on a glance on https://grep.app

x789:
Reduced visibility as discussed in #3932

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Numerics.cs
+++ b/src/NUnitFramework/framework/Constraints/Numerics.cs
@@ -8,7 +8,7 @@ namespace NUnit.Framework.Constraints
     /// <summary>
     /// The Numerics class contains common operations on numeric values.
     /// </summary>
-    public static class Numerics
+    internal static class Numerics
     {
         #region Numeric Type Recognition
         /// <summary>


manfred-brands (approved):

jnm2:
> This PR closes https://github.com/nunit/nunit/issues/3932

This causes that issue to close when this PR is merged even if the other outstanding PR required by the issue is not merged.

x789:
I changed the description of the PR. It is now save to merge it without closing the issue.

jnm2 (approved):

jnm2:
Thanks for your help!

PR #4172, opened 2022-09-07
change ITypeInfo.Namespace to be nullable

SimonCropp:
since in .net namespace can be null

PR #4104, opened 2022-04-23
Handle Exceptions thrown during discovery, such as in Custom Attributes

manfred-brands:
Fixes #4096 
Fixes #4107

Tests with Exceptions are Marked As Invalid and show up as an Error.
Fixtures with Exceptions are non-runnable and don't show up with `dotnet test`.

One alternative option would be to make a special fixture with one runable method that fails.

PR #4206, opened 2022-10-01
Mark 'Numerics' as obsolete (for public use)

x789:
As discussed in #3932

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/NumericsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/NumericsComparer.cs
@@ -30,10 +30,12 @@ namespace NUnit.Framework.Constraints.Comparers
     {
         public bool? Equal(object x, object y, ref Tolerance tolerance, ComparisonState state)
         {
+#pragma warning disable CS0618 // 'Numerics' is only marked as obsolete for public use
             if (!Numerics.IsNumericType(x) || !Numerics.IsNumericType(y))
                 return null;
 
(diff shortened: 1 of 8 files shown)


jnm2 (approved):

x789:
I have no clue why [the build is failing](https://github.com/nunit/nunit/actions/runs/3166225898). I suspect it has something to do with the configuration of the build agent.
At the time of the last merge to v3.13-dev [a successful build was possible](https://github.com/nunit/nunit/actions/runs/2478493187). When I forked the code a few days later, [the build failed](https://github.com/x789/nunit/actions/runs/3164653353/jobs/5153097316).

The build works on my machine, if I comment out the F# part.  (Win10x64 with NET3.5, 4.6.1, 4.6.2, 4.7, 4.8, 4.8.1, 6.0 SDKs/Targeting-Packs)

Since I cannot find an explicit reference to `System.ValueTuple 4.4.0` in the code, I suspect that the reference is implicitly due to a (non) installed framework.
@manfred-brands: when you created PR #4104, everything worked fine. Do you have any idea what I can do so that the build does not fail?

mikkelbu:
My best guess (just a quick googling is that the reason for the builds to fail is dotnet/sdk#24747. I've not looked at the solutions yet, but it seems like the problem mention in https://github.com/nunit/nunit/pull/4172#pullrequestreview-1098930085 has resurfaced.

x789:
Thank you @mikkelbu. Changing the `PackageReference` repaired the build-error.

But unfortunately, the macOS-build fails because of a trust-settings issue.

Issue #4187, opened 2022-09-21
 CollectionAssert.DoesNotContain for a list of item


WindingWinter:
I'm looking for a  CollectionAssert.DoesNotContain for a list of item, ie, something like the below:

```
       // all items in excludedList shall not contain in the existingList
        private static void CollectionAssertDoesNotContain<T>(List<T> existingList, IEnumerable<T> excludedList)
        { 
            foreach (var obj in excludedList)
            {
                CollectionAssert.DoesNotContain(existingList, obj);
            }
        }
```

Do we already have such a function?

stevenaw:
Great question @WindingWinter !
While the superset / subset methods could validate if ALL items are present, I haven't found something which works for the ANY case quite so easily. Something like this would be possible:
```csharp
            Assert.That(a.Union(b).ToArray(), Has.Length.EqualTo(a.Length + b.Length));
```
However you would of course lose the item-level failure reporting of which items appear in `existingList`. I'm going to label this as an idea for consideration for inclusion, though I would welcome if anyone could show a built way to do this without requiring the explicit loop

tbenne10:
I ran into this same issue last week - CollectionAssert does not contain any methods that perform evaluation for disjoint sets. 

I have added in changes and tests to include ContainsAny(IEnumerable actual, IEnumerable expected) and  DoesNotContainAny(IEnumerable actual, IEnumerable expected) methods. Noting this in case someone wanted to pick it up, I will have the build complete and ready for PR within a day.

CharliePoole:
Historically, CollectionAssert has never been the primary way of doing assertions on collections in NUnit. I added the class many years ago, when the NUnit constraint model was brand new. It was intended to allow folks familiar with the MSTest CollectionAssert to use the same syntax. We consistently turned down requests to enhance CollectionAssert for many years, instead adding new constraints, which would work with Assert. CollectionAssert, in fact, was at one time listed to be removed in the next major version. If you decide to change the policy of freezing CollectionAssert, I suggest you add a new constraint for use by Assert and make use of it in the CollectionAssert implementation.

stevenaw:
I am going to close this due to inactivity, as well as based on Charlie's comment above regarding a past decision to freeze that API.
Note that this is possible to do in the constraint model today based on the following example:

```csharp
int[] array = [1, 2, 3];
int[] actual = [4, 5, 6];

Assert.That(actual, Has.None.AnyOf(array));
```

stevenaw:
[...]

Issue #3688, opened 2020-12-06
Clarifying the status of the classic model assertions


jnm2:
Do we document anywhere that we consider the classic model to be legacy? We bring it up in GitHub issues, but people seem to be surprised to learn it, e.g. https://github.com/nunit/nunit/issues/3644#issuecomment-739532642.

https://docs.nunit.org/articles/nunit/writing-tests/assertions/assertions.html hints but doesn't come right out and say it.

1. Is "legacy" the word we want to use?
2. Do we want more folks to know about this designation?
3. Which pages would be good places to get folks' attention?

mikkelbu:
1. I think legacy has a negative sound, so in the analyzers we also use "constraint model" vs "classic model" (and I prefer this to legacy). A comment on nunit/nunit.analyzers#9 states that they where called legacy asserts at some time
> There was a time, about 7 years ago, when I wanted to get rid of Classic asserts. In fact, I was calling them legacy asserts back then. User feedback strongly favored keeping them and so I dropped the goal of dropping them. :-)

2. We can make it more clear, but I can understand why some use the classical model for simple asserts like `Assert.IsTrue(expression)` or `Assert.IsNull(expression)` rather than the longer constraint variant.

3. I think the page that you already link to and perhaps a new page under "Assertion Models" so that one does not end up on "classical models" when expanding the item.

Ps. It also seems like we have much more documentation (at least in the index) for the classical model compared to the constraint model, so I think we should add all the constraint pages, like https://docs.nunit.org/articles/nunit/writing-tests/constraints/AnyOfConstraint.html, to an item under "Constraint model"
![image](https://user-images.githubusercontent.com/1007631/101342144-71bc4580-3882-11eb-9025-bb66cad311b6.png)

OsirisTerje:
About the documentation we have:  I notice that many of the examples we show use the legacy/classic model.  It might be helpful to change those to use the constraint model.

stevenaw:
This came up at work yesterday. Another thing which could be helpful to document when we do this is to describe the reasons why the constraint model is recommended.

stevenaw:
This came up in a recent PR again. @CharliePoole gave the helpful context in the originating issue that the `CollectionAssert` class was at one point considered for removal. Given the maturity of the analyzers and fixers, I'm wondering if this is something we should now consider. Perhaps just marking it as `Obsolete` could also communicate long-term intent.

[...]

PR #4211, opened 2022-10-04
Added ContainsAny and DoesNotContainAny to CollectionAssert

tbenne10:
Mentioned in #4187 - Added feature to allow CollectionAssert to verify if a  IEnumerable contains any or does not contain any elements from a provided IEnumerable

### DIFF
--- a/src/NUnitFramework/framework/Is.cs
+++ b/src/NUnitFramework/framework/Is.cs
@@ -373,6 +373,19 @@ namespace NUnit.Framework
 
         #endregion
 
+        #region ContainedIn
+
+        /// <summary>
+        /// Returns a constraint that tests whether the actual value
+        /// contains a value from the collection supplied as an argument.
+        /// </summary>
+        public static CollectionContainsAnyConstraint ContainedIn(IEnumerable expected)
+        {
+            return new CollectionContainsAnyConstraint(expected);
+        }
+
+        #endregion
+
         #region SubsetOf
 
         /// <summary>
--- a/src/NUnitFramework/framework/CollectionAssert.cs
+++ b/src/NUnitFramework/framework/CollectionAssert.cs
@@ -294,6 +294,30 @@ namespace NUnit.Framework
         }
         #endregion
 
+        #region ContainsAny
+        /// <summary>
+        /// Asserts that collection contains any element of another collection.
+        /// </summary>
+        /// <param name="collection">IEnumerable of objects to be considered</param>
+        /// <param name="expected">IEnumerable of Objects to be found within collection</param>
+        public static void ContainsAny(IEnumerable collection, IEnumerable expected)
+        {
+            ContainsAny(collection, expected, string.Empty, null);
+        }
+
+        /// <summary>
+        /// Asserts that collection contains any element of another collection.
+        /// </summary>
+        /// <param name="collection">IEnumerable of objects to be considered</param>
+        /// <param name="expected">IEnumerable of Objects to be found within collection</param>
+        /// <param name="message">The message that will be displayed on failure</param>
+        /// <param name="args">Arguments to be used in formatting the message</param>
+        public static void ContainsAny(IEnumerable collection, IEnumerable expected, string message, params object[] args)
+        {
+            Assert.That(collection, Is.ContainedIn(expected), message, args);
+        }
+        #endregion
+
         #region DoesNotContain
 
         /// <summary>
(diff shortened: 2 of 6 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=4211) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=4211) before we can accept your contribution.<br/><br/>:x: tbenne10 [sign now](https://cla.dotnetfoundation.org/nunit/nunit?pullRequest=4211)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/nunit/nunit?pullRequest=4211) it.</sub>

MoritzHayden (approved):
@tbenne10 Thank you for adding this!

OsirisTerje:
@stevenaw Can you have a look at this?

SimonCropp (approved):

OsirisTerje (changes requested):
I am struggling a bit with the syntax of these.  
The first looks fine:   I would write  `Assert.That(collectionA,Does.ContainsAny(collectionB)`.  
The second becomes strange, because I would expect to be writing  `Assert.That(collectionA,Does.Not.ContainsAny(collectionB)`.  So when the keyword I DoesNotContainsAny it looks like a legacy collection assert, and not a constraint based one.

mr-russ:
I also recommend Charlie Pool's comment on #4187 , and decide if we are going down the road of new constraint.

src/NUnitFramework/framework/Is.cs
```diff
@@ -373,6 +373,19 @@ public static CollectionEquivalentConstraint EquivalentTo(IEnumerable expected)
 
         #endregion
 
+        #region ContainedIn
+
+        /// <summary>
+        /// Returns a constraint that tests whether the actual value
+        /// contains a value from the collection supplied as an argument.
+        /// </summary>
+        public static CollectionContainsAnyConstraint ContainedIn(IEnumerable expected)
```
stevenaw, line 382: Thanks for adding this. I often also use the `Contains` class for this sort of check too. Would it be possible to also expose a `Contains.All(IEnumerable)` function there too?

stevenaw (changes requested):
Thanks for taking initiative on this @tbenne10 
Apologies for a lack of up-front clarity on this. As a team we're trying to adopt a constraint-first approach to new features. As a team we also need to figure out how to make this more well-known, which is on us. We have https://github.com/nunit/nunit/issues/3688 to track doing that.

After this PR was created, Charlie has also weighed in in the issue about a previous bit of discussion to freeze the `CollectionAssert` API and that the team would need to decide if a change in direction were to happen. I think that would be an important conversation to have, but it wouldn't need to block adding this in as constraints.

Would it be possible to adjust the added tests to target the new constraint instead of `CollectionAssert` itself? In lieu of a change in direction about `CollectionAssert` it may also help to remove any changes to `CollectionAssert` in the PR.

src/NUnitFramework/framework/Is.cs
tbenne10: Anytime - just want to clarify, do you use `Contains` with iteration to perform the checks you need? Just want to make sure we are on the same page since they perform different actions. 

The naming might have been a bit misleading here - `Contains.All(IEnumerable)` already exists but is instead named SupersetOf(IEnumerable) within this class.

tbenne10:
Thank you all for your input. I will remove changes to `CollectionAssert` and retarget the tests per @stevenaw's comment. 

Before I complete updates, I wanted to circle back and check with everyone if we want to support adding methods in `Assert.cs` utilizing the constrains within this MR. If so I will add those with naming suggestions from @OsirisTerje unless anyone has differing naming suggestions.

src/NUnitFramework/framework/Is.cs
stevenaw: @tbenne10 Thanks for asking, I appreciate you coming back to this PR after all this time.

You make a good point. I may've mistyped my earlier suggestion in haste. I'd suggest we disregard my suggestion around `Contains.All()` in the interest of keeping this PR simple. Having thought it through, I agree with @OsirisTerje 's suggestion that it might make sense to expose your new constraint as a `Does.ContainAny()` method, which could then be easily inverted to be `Does.Not.ContainAny()` using the already-existing `Not` constraint.


I think my original comment came from a place of concern around ambiguity of the name `Is.ContainedIn` in relation to some other API operations when both arguments are collections. If I understand your PR right, the following will pass the first assertion but fail the second and third because they apply fundamentally different checks, though it might not be clear simply by reading them that this will happen. 

```csharp
var a = new[] { 1, 2, 3 };
var b = new[] { 2, 4 };

Assert.That(a, Is.ContainedIn(b));
Assert.That(a, Is.SubsetOf(b));
Assert.That(b, Contains.All(b));
```

OsirisTerje:
@tbenne10 Seems this has been dormant for some time.  Do you think you have time to get it up to shape?

OsirisTerje:
@tbenne10 Do you think you can resolve the remaining comments?

mithileshz:
@tbenne10 @OsirisTerje would you mind if I take over and complete the changes on this?

OsirisTerje:
@mithileshz Are you working on this?

OsirisTerje:
Seems no one wants to finish this. Closing for now.  
Feel free to reopen only if you really want to do something here.

PR #4215, opened 2022-10-07
remove some unused variables

SimonCropp:


src/NUnitFramework/framework/Constraints/CollectionConstraint.cs
```diff
@@ -35,10 +36,7 @@ protected static bool IsEmpty(IEnumerable enumerable)
             if (enumerable is ICollection collection)
                 return collection.Count == 0;
 
-            foreach (object o in enumerable)
-                return false;
-
-            return true;
+            return !enumerable.Cast<object>().Any();
```
jnm2, line 39: I wonder if `object _` would have worked to resolve the unused variable notification. This is a nice simplification, though.

It's probably not worth making more complex to squeeze out perf:
```cs
var enumerator = enumerable.GetEnumerator();
using (enumerator as IDisposable)
    return enumerator.MoveNext();
```

src/NUnitFramework/tests/Internal/EventQueueTests.cs
```diff
@@ -145,7 +145,7 @@ public class SetWaitHandle_Enqueue_AsynchronousTest : ProducerConsumerTest
 #endif
             public void SetWaitHandle_Enqueue_Asynchronous()
             {
-                using (AutoResetEvent waitHandle = new AutoResetEvent(false))
+                using (new AutoResetEvent(false))
```
jnm2, line 148: Why create `AutoResetEvent` at all?

SimonCropp: no idea. it was added by @CharliePoole 9 years ago https://github.com/nunit/nunit/commit/4bbf615e01bc0d6078eac6dce07d4f0aac5c03ec

SimonCropp: seems the handle was previously passed into `SetWaitHandleForSynchronizedEvents` https://github.com/nunit/nunit/blame/024bab1a77aee1e56a8851ac3bf5997c578071be/NUnitFramework/src/tests/Internal/EventQueueTests.cs#L208-L215

i assume that functionality has been refactored away and this test is no longer relevant?

stevenaw: I think you may be correct.

This historical comment here implies it was used for synchronous events: https://github.com/nunit/nunit/blob/96cb0664cd505beaddfdf0e300c1dc82951e2139/src/NUnitFramework/framework/Internal/Execution/EventQueue.cs#L99

Which were removed (also by @CharliePoole ) in https://github.com/nunit/nunit/pull/1526

stevenaw (approved):

src/NUnitFramework/tests/Internal/EventQueueTests.cs
jnm2: That sounds accurate.

SimonCropp: test removed

SimonCropp:
any more feedback?

mikkelbu (approved):

PR #4217, opened 2022-10-07
throws should return not null

SimonCropp:


manfred-brands (commented):
The method does return `null` if no exception is thrown inside an `Assert.Multiple()`

jnm2:
See https://github.com/nunit/nunit/issues/3376#issuecomment-983257658. Since we had a design discussion which settled there, we'd have to reopen it and decide again.

manfred-brands:
See also https://github.com/nunit/nunit.analyzers/issues/333

PR #4218, opened 2022-10-07
Add NotNull to NotNull

SimonCropp:


manfred-brands (commented):
Note that `[NotNull]` is not true of the code is called from inside an `Assert.Multiple()`.
These cases are correctly detected by the NUnit.Analyzer.

jnm2:
See https://github.com/nunit/nunit/issues/3376#issuecomment-983257658. Since we had a design discussion which settled there, we'd have to reopen it and decide again.

manfred-brands:
See also https://github.com/nunit/nunit.analyzers/issues/157

PR #4219, opened 2022-10-07
make classes static where possible

SimonCropp:


src/NUnitFramework/framework/Assume.cs
```diff
@@ -16,7 +16,7 @@ namespace NUnit.Framework
     /// result. If an assumption is not met, the test
     /// should produce an inconclusive result.
     /// </summary>
-    public class Assume
+    public static class Assume
```
jnm2, line 19: We shouldn't make public classes outside the Internal namespace static. A publicized technique for "extending" the NUnit assertion classes is to do this:

```cs
public class Assert : NUnit.Framework.Assert
{
    public static void Foo(...) { }
}
```

jnm2: We could consider making them `abstract` to break code which does `new Assert()` and I thought we had done so.

jnm2 (changes requested):

src/NUnitFramework/framework/Assume.cs
SimonCropp: so revert the Assume and Warn changes?

src/NUnitFramework/framework/Assume.cs
```diff
@@ -16,7 +16,7 @@ namespace NUnit.Framework
     /// result. If an assumption is not met, the test
     /// should produce an inconclusive result.
     /// </summary>
-    public class Assume
+    public abstract class Assume
```
rprouse, line 19: Why `abstract` and not `static`? I don't see anything that makes this `abstract` and it is used as-is? Same with the other usages of `abstract`.

rprouse: Nevermind, I see Joseph's comment. That is a good point.

rprouse (approved):

Issue #4158, opened 2022-08-19
SetupFixture should have an `AttributeUsage` of `Inherited = false`


stevenaw:
Based on the conversation in https://github.com/nunit/nunit/issues/3937, `SetupFixture` isn't intended to be used in a base class, however there is nothing stopping it from being added to a base class or even an abstract base class. One source of confusion here could be that the attribute is explicitly marked with `Inherited = true` prop in the `AttributeUsage`. It might help avoid confusion to make this explicitly false.

x789:
I can take care of it.

Shall I also make this change in the v3 branch?
Normally, I would say that setting `Inherited` to `false` is a breaking change. But if I understood the discussion in #3937 correctly, the attribute does not work with derived classes, therefore it should be okay in this context.

stevenaw:
Thanks once again for tackling this!

PR #4222, opened 2022-10-08
Indicate that `SetupFixtureAttribute` should not be used on base-classes

x789:
As discussed in #4158.


jnm2 (approved):
Thanks for your help!

stevenaw (approved):
Thanks for your contribution @x789 !

PR #4223, opened 2022-10-08
Indicate that SetupFixtureAttribute should not be used on base-classes

x789:
As discussed in #4158. This PR contains the changes for NUnit v3.


jnm2 (approved):

jnm2:
The build failures are because the `v3.13.-dev` branch is currently in a bad state.

stevenaw (approved):
Thanks!
I'm a little puzzled by the build failures myself though they're definitely not related to your changes. Will take a bit of digging to sort out

x789:
@mikkelbu gave a hint how to fix the `System.ValueTuple` build-error.

jnm2:
Maybe we don't need that package reference at all? Eventually it'll be bringing the package version backwards.

PR #4227, opened 2022-10-10
enable nullable in filters

SimonCropp:


src/NUnitFramework/framework/Internal/Filters/ValueMatchFilter.cs
```diff
@@ -1,7 +1,6 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
-using System;
-using System.Collections.Generic;
+#nullable enable
```
stevenaw, line 3: I'm having a hard time submitting this comment as a suggestion, but I think this adds duplicate `#nullable enable` in the file (on line 7), though for consistency it might be good to remove that one and keep this to have them all above the `using`s

src/NUnitFramework/framework/Internal/Filters/OrFilter.cs
```diff
@@ -111,20 +108,13 @@ public override bool Match( ITest test )
         /// <returns>True if any of the component filters explicit match, otherwise false</returns>
         public override bool IsExplicitMatch( ITest test )
         {
-            foreach( TestFilter filter in Filters )
-                if ( filter.IsExplicitMatch( test ) )
-                    return true;
-
-            return false;
+            return Filters.Any(filter => filter.IsExplicitMatch(test));
```
stevenaw, line 111: Context: Some parts of the framework were updated recently to move away from linq to avoid closure allocations. I can't recall where, but based on the file history I don't believe it was the filters here so I think this should be ok.

stevenaw (commented):
Thanks @SimonCropp this looks good to me, just one minor nitpick about a duplicate `#nullable` directive in a file.

SimonCropp:
should be all good now

stevenaw:
Thanks! LGTM 🚀

PR #4232, opened 2022-10-11
remove LOAD_TIMING

SimonCropp:
i assume the LOAD_TIMING condition was put in due to the perception that a stopwatch is slow. stopwatch is actually very efficient. the cost is the time for 2xDateTime.Now + calculating the diff between them. in the context of loading and parsing assemblies that time is insignificant. So if we want to know that time, we may as well always log it

thoughts?


mikkelbu (approved):
LGTM

PR #4247, opened 2022-10-30
thoughts on Implicit usings?

SimonCropp:


manfred-brands:
I see you put a lot of work in this again, but my thoughts on implicit using is: don't use it.
I like to see from a file what it is pulling in.
Now we have a `GlobalUsings.cs` referencing everything. It will become a file were lines are not removed when no longer needed as nobody knows where that using statement is actually used.

SimonCropp:
fair enough

Issue #4255, opened 2022-11-16
InternalTrace.Initialize fails with Nullref exception


OsirisTerje:
When testing the 4.3.1 NUnit3Testadapter I'm experiencing crashes in the InternalTrace.Initialize, which raises a nullref exception.  

The stacktrace and messages are
```txt
SendMessage: Current directory: D:\repos\NUnit\nunit3-vs-adapter\src\NUnitTestAdapterTests\bin\Debug\netcoreapp3.1
SendMessage:    Exception thrown executing tests in D:\repos\NUnit\nunit3-vs-adapter\src\NUnitTestAdapterTests\bin\Debug\netcoreapp3.1\mock-assembly.dll
SendMessage: NUnit.Engine.NUnitEngineException: An exception occurred in the driver while loading tests.
 ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at NUnit.Framework.Internal.InternalTrace.Initialize(String logName, InternalTraceLevel level)
   at NUnit.Framework.Api.FrameworkController.Initialize(String assemblyNameOrPath, IDictionary settings)
   at NUnit.Framework.Api.FrameworkController..ctor(String assemblyNameOrPath, String idPrefix, IDictionary settings)
   at NUnit.Framework.Api.FrameworkController..ctor(Assembly assembly, String idPrefix, IDictionary settings)
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor, Boolean wrapExceptions)
   at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.RuntimeType.CreateInstanceImpl(BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture)
   at System.Activator.CreateInstance(Type type, BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture, Object[] activationAttributes)
   at System.Activator.CreateInstance(Type type, Object[] args)
   at NUnit.Engine.Drivers.NUnitNetStandardDriver.CreateObject(String typeName, Object[] args)
   at NUnit.Engine.Drivers.NUnitNetStandardDriver.Load(String testAssembly, IDictionary`2 settings)
[...]

PR #4256, opened 2022-11-17
Fix for #4255, InternalTrace Initialize crash

OsirisTerje:
See #4255

### DIFF
--- a/src/NUnitFramework/framework/Internal/Logging/InternalTrace.cs
+++ b/src/NUnitFramework/framework/Internal/Logging/InternalTrace.cs
@@ -28,7 +28,7 @@ namespace NUnit.Framework.Internal
         /// <summary>
         /// Gets a flag indicating whether the InternalTrace is initialized
         /// </summary>
-        public static bool Initialized { get; private set; }
+        public static bool Initialized => traceWriter != null;
 
         /// <summary>
         /// Initialize the internal trace facility using the name of the log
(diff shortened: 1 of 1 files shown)


CharliePoole (commented):
This seems to work for your particular use case in the adapter but changes the nature of the check previously made from ensuring that Initialization is only done once to ensuring that initialization is only done once with a level > `Off`.

I think the original intent (don't initialize twice) is what should be done but that the code error causing the exception should be removed. Possibly, the `TraceWriter` should not be initialized at all in this method, but that would require more work.

CharliePoole:
It seems to me that any code (whether the adapter or the engine) that calls `InitializeTrace` more than once is in error. While the exception should be fixed, it should not, in my view, keep us from figuring out why that call is made twice and fixing it.

OsirisTerje:
As mentioned for the issue, the adapter is actually calling the assembly twice in its own testing.  You can check this by running the red tests in the adapter, branch Fixtmppath. 

I can't see that the change really "changes the nature" of this.  If the method is called twice, the else clause simply ignores the call, just as it did before.

CharliePoole:
@OsirisTerje I can see your point. The change you made is very small, I agree, but it's a small change in the semantics of the check being made. If I were re-doing this, I think I'd just throw an exception on the second initialization.

rprouse (approved):

PR #4260, opened 2022-11-27
PropertyConstraint on Type

### LINKED ISSUE #2436: Properties on System.Type cannot be used with either property constraint
Given:

```cs
Assert.That(
    new[] { typeof(TopLevelType), typeof(TopLevelType.NestedType) },
    Has.None.With.Property("IsNested").EqualTo(true));

public static class TopLevelType
{
    public static class NestedType
    {
    }
}
```

I get:
> System.ArgumentException : Property IsNested was not found

Just as though the `.None.With` was not there!

But it gets weirder. If I use any type other than System.Type, it behaves:

```cs
Assert.That(
    new[] { new FakeTypeForTesting(false), new FakeTypeForTesting(true) },
    Has.None.With.Property("IsNested").EqualTo(true));

public sealed class FakeTypeForTesting
{
    public bool IsNested { get; }

    public FakeTypeForTesting(bool isNested)
    {
        IsNested = isNested;
    }
}
```

Gives what it should:

>   Expected: no item property IsNested equal to True
>   But was:  < \<FakeTypeForTesting>, \<FakeTypeForTesting> >

### LINKED ISSUE #4259: Bug: Assert.That(IEnumerable<Type>, Has.All.Property(nameof(Type.Namespace)) fails.
Originally posted as https://stackoverflow.com/questions/74475185.
Using NUnit v3.13.2, .NET Framework 4.8, running using ReSharper test runner.  I'll try to find some time to run this with just nunit console runner and/or actually debug this myself.

I want to assert that all the types I am interested in are in the expected namespace using NUnit.
```
IReadOnlyCollection<Type> types = GetInterestingTypes(); // Method defined elsewhere.
Assert.That(types, Has.All.Property(nameof(Type.Namespace)).StartsWith("MyNamespace"));
```
This fails with `ArgumentException Property Namespace not found on ...`.
(Note the `StartsWith` is irrelevant.)

However, the equivalent non-collection assert `Has.Property(nameof(Type.Namespace)).StartsWith("MyNamespace")` succeeds.
I assume that this is a bug, if not, let's make sure the documentation is clear about how this is supposed to work.

Minimal reproducible example:
```
using System;
using System.Linq;
using System.Reflection;
using System.Runtime.CompilerServices;

using NUnit.Framework;

namespace MyNamespace
{
    internal class MyTests
    {
        [Test]
        public void TestsWithinNamespace()
        {
            var types = new[] {typeof(Foo), typeof(NestedNamespace.Bar)};

            foreach (var type in types)
            {
                // Passes
                Assert.That(type, Has.Property(nameof(Type.Namespace)).StartsWith("MyNamespace"));
            }

            // Fails
            Assert.That(types, Has.All.Property(nameof(Type.Namespace)).StartsWith("MyNamespace"));
        }
    }

    public class Foo { }

    namespace NestedNamespace
    {
        public class Bar { }
    }
}
```
When running this test, we get

```
System.ArgumentException : Property Namespace was not found on MyNamespace.Foo.
Parameter name: name
   at NUnit.Framework.Constraints.PropertyConstraint.ApplyTo[TActual](TActual actual)
   at NUnit.Framework.Constraints.AllItemsConstraint.ApplyTo[TActual](TActual actual)
   at MyNamespace.MyTests.TestsWithinNamespace()
```

Potentially relevant comment from Ralf on the stackoverflow question:
Debugging through NUnit i would assume a bug here. When using Property on the AllItemConstraint there is an extra handling when the thing in there is already a type. It does not create a type on the type that you would need for your code to work. Now it searches "Namespace" on Foo and not on the type object of Foo.

manfred-brands:
Fixes #4259 
Fixes #2436

### DIFF
--- a/src/NUnitFramework/Directory.Build.props
+++ b/src/NUnitFramework/Directory.Build.props
@@ -12,6 +12,7 @@
     <NUnitRuntimeFrameworks>$(OlderNUnitRuntimeFrameworks);$(NewerNunitRuntimeFrameworks)</NUnitRuntimeFrameworks>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <OutputPath>..\..\..\bin\$(Configuration)\</OutputPath>
+    <CheckEolTargetFramework>false</CheckEolTargetFramework>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(BuildingInsideVisualStudio)' != 'true'">
--- a/src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
@@ -71,13 +71,21 @@ namespace NUnit.Framework.Constraints
 
             if (property == null)
             {
-                Type actualType = actual as Type ?? actual.GetType();
+                if (actual is Type actualType)
+                {
+                    property = Reflect.GetUltimateShadowingProperty(actualType, name, bindingFlags);
+                }
 
-                property = Reflect.GetUltimateShadowingProperty(actualType, name, bindingFlags);
-
-                // TODO: Use an error result here
                 if (property == null)
-                    throw new ArgumentException($"Property {name} was not found on {actualType}.", nameof(name));
+                {
+                    actualType = actual.GetType();
+
+                    property = Reflect.GetUltimateShadowingProperty(actualType, name, bindingFlags);
+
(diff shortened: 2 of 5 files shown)


SimonCropp (approved):

stevenaw (approved):
Thanks again for the contribution @manfred-brands . LGTM 🚀
Appreciate the adapter update in this PR too.

PR #4278, opened 2023-01-15
add [NotNull] to Assert.NotNull

SimonCropp:
so the compiler knows the value is not null when the mothod returns (doesnt throw)

### DIFF
--- a/src/NUnitFramework/framework/Assert.Conditions.cs
+++ b/src/NUnitFramework/framework/Assert.Conditions.cs
@@ -5,6 +5,7 @@
 using System.Collections;
 using NUnit.Framework.Constraints;
 using System;
+using System.Diagnostics.CodeAnalysis;
 
 namespace NUnit.Framework
 {
@@ -187,6 +188,7 @@ namespace NUnit.Framework
 
         #region NotNull
 
+#pragma warning disable CS8777
         /// <summary>
         /// Verifies that the object that is passed in is not equal to <see langword="null"/>. Returns without throwing an
         /// exception when inside a multiple assert block.
@@ -194,9 +196,9 @@ namespace NUnit.Framework
         /// <param name="anObject">The object that is to be tested</param>
         /// <param name="message">The message to display in case of failure</param>
         /// <param name="args">Array of objects to be used in formatting the message</param>
-        public static void NotNull(object? anObject, string? message, params object?[]? args)
+        public static void NotNull([NotNull] object? anObject, string? message, params object?[]? args)
         {
-            Assert.That(anObject, Is.Not.Null ,message, args);
+            Assert.That(anObject, Is.Not.Null, message, args);
         }
 
         /// <summary>
@@ -204,7 +206,7 @@ namespace NUnit.Framework
         /// exception when inside a multiple assert block.
         /// </summary>
         /// <param name="anObject">The object that is to be tested</param>
-        public static void NotNull(object? anObject)
+        public static void NotNull([NotNull]object? anObject)
         {
             Assert.That(anObject, Is.Not.Null ,null, null);
         }
@@ -216,7 +218,7 @@ namespace NUnit.Framework
         /// <param name="anObject">The object that is to be tested</param>
         /// <param name="message">The message to display in case of failure</param>
         /// <param name="args">Array of objects to be used in formatting the message</param>
-        public static void IsNotNull(object? anObject, string? message, params object?[]? args)
+        public static void IsNotNull([NotNull] object? anObject, string? message, params object?[]? args)
         {
             Assert.That(anObject, Is.Not.Null ,message, args);
         }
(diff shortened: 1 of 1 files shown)


SeanKilleen (approved):
Makes sense to me. Took me a second to grasp the purpose, even though it was right there in the PR description. 

To elaborate, for those like me who are a catching up a little on nullability stuff: at the conclusion of this method, if it doesn't throw, we are sure that the input is not null. The `[NotNull]` postcondition is a helpful compiler hint for down the line.

manfred-brands (commented):
This is not correct. Even the XML doc comment states it. If called from inside an Assert.Multiple the assert returns even if the value is null. This does seem to make the whole method a bit useless.

SeanKilleen:
@manfred-brands thank you, I'd missed the part about the assert.multiple. removing my approval so others can consider in more detail and I'll be more cautious in the future.

stevenaw:
Thanks for noticing this @SimonCropp 
@jnm2 did much of the original attribute decorating and I believe as @manfred-brands mentioned, the `Assert.Multiple()` scenario meant this attribute usage had to be omitted. There's a bit of a history discussed in https://github.com/nunit/nunit/issues/3376 and https://github.com/nunit/nunit/pull/3677, which I _think_ ended up with this instead being tackled as an analyzer in https://github.com/nunit/nunit.analyzers/issues/157

SeanKilleen:
Makes sense. @SimonCropp maybe you could turn this into a comment that documents said reasoning?

jnm2:
The NUnit.Analyzers invested heavily in diagnostic suppressions to make this a smooth experience. I wonder if it's time to make the NUnit package bring in the analyzers without people having to opt in.

SimonCropp:
thanks for the clarification peoples. and sorry for the noise, i am still getting familiar with the codebase.

@SeanKilleen yep. good idea. will do

and i am sure this discussion has been had, but was a diff api considered? perhaps

```
    Assert.Multiple(_ =>
    {
        _.AreEqual(5.2, result.RealPart, "Real Part");
        _.AreEqual(3.9, result.ImaginaryPart, "Imaginary Part");
    });
```

jnm2:
@Jarek300 Using the NUnit.Analyzers package addresses this use case, and there's a reason we don't want to use `[NotNull]` as discussed above.

mikkelbu:
I'm closing this (like we closed #4218).

@SimonCropp  Regarding why the API looks the way it does, then I think this design was made before most people were part of this project (or at least me) almost 6 years ago. Perhaps @rprouse can remember the details - see e.g. https://github.com/nunit/nunit/issues/391 for context and perhaps the Multiple-Asserts-Spec can be found on the old wiki (if necessary).

Jarek300:
The issue is that asserts IsNotNull/NotNull don't suppress nullability warnings in the next lines.

How NUnit.Analyzers addresses this?

It compatibility with VS 2017 the only reason not to simply add [NotNull] attribute? If yes then when compatibility with VS2017 can be dropped?

jnm2:
Also, compatibility with 2017 was not the reason. We do use annotations where they are accurate. The reason was correctness (these Assert methods do not always throw) as the XML docs on these methods explain.

Jarek300:
I'm not going to keep persuading you because you have already decided. I will only tell why I'm not going to use your solution.

Nullable types are one of the most important new features. Not only in C# but also in many other languages. Every library sooner or later must correctly support it. And here, all what is needed, is to suppress one warning. 

I tried to add NUnit.Analyzers and it created thousands of new warning. Generally that Assert.AreEqual is not fashionable and that Assert.That is all the rage. So now I would have to suppress warnings from NUnit.Analyzers in all my test projects to suppress one warning from compiler.

Instead of using NUnit.Analyzers, I just have added to my code class AssertThat with one function IsNotNull. And instead of Assert.IsNotNull I write  AssertThat.IsNotNull. Not pretty but good enough and better than using NUnit.Analyzers. Maybe one day NUnit will have something similar.

SeanKilleen:
@Jarek300 I just want to clarify what's being said above because I fear you may have taken the wrong impression from it.

* Where NUnit is able, as you can see, we attempt to add the appropriate compiler hints around nullable checks. Because, as you state, we also believe that nullable types and nullable checks are important.
* As is stated in this thread, suppressing this one warning, in this one case, actually causes a problem, because it is not always correct. If this assertion is inside a block of multiple assertions, it is not guaranteed to be correct. We do not want to tell the compiler things that we do not know to be true/correct.
* Because of the above, this warning _has_ to be included in the NUnit analyzers package, which has the ability to take context into account such as whether the code is running in an Assert.Multiple block. This way, we can emit the correct nullability information.

> I tried to add NUnit.Analyzers and it created thousands of new warning. Generally that Assert.AreEqual is not fashionable and that Assert.That is all the rage

I know the warnings can be daunting whenever you first add an analyzer package. But they're there because the authors of a library are trying to guide toward better outcomes. In this case, the analyzer documentation comes with both an explanation ([example](https://docs.nunit.org/articles/nunit-analyzers/NUnit2005.html)) and a codefix that will automatically change it for you. I know when changing my own assertions to this format I found at least 5 assertions that were wrong, and I've been doing this for a while.

> I will only tell why I'm not going to use your solution.

You are free to use any solution you like. 👍 I'd offer you a refund except it's already free.

PR #4295, opened 2023-02-22
Remove Azure DevOps pipeline and update GitHub Action

### LINKED ISSUE #4293: .NET5 Test Suite is failing on Azure DevOps for Windows
It looks like the test suite is having issues targeting .NET5 for Window on Azure DevOps. Multiple recent PRs have failed on the .NET5 build with the following ([example link](https://nunit.visualstudio.com/NUnit/_build/results?buildId=5961&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=abae1f68-3c73-5bff-491f-f2b908580ce6&l=1701))

> ========================================
TestNetStandard20 on net5.0
##[error]Unable to process command '##vso[results.publish type=NUnit;mergeResults=true;config=Release;runTitle='net5.0/Windows';publishRunAttachments=true;]' successfully. Please reference documentation (http://go.microsoft.com/fwlink/?LinkId=817296)

The build output indicates that .NET5 isn't installed by default, and the image documentation supports this (https://github.com/actions/runner-images/blob/main/images/win/Windows2019-Readme.md#net-core-tools)

The Linux and MacOS builds appear fine since it seems we install the runtime  as part of the build. While normally we may consider manually installing it on the Windows image too, .NET5 has been out of support by Microsoft for about 9 months, since May 10, 2022 (https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core). We should consider if we should explicitly target it on the Windows build or, perhaps, to instead remove it as a test suite target.

Any change here would need to be included in both main + v3.13 branches

rprouse:
- Removes the Azure DevOps pipeline (fixes #4293) 
- Updates the GitHub Actions to give them names so they can be added to the branch protection rules


stevenaw (approved):
LGTM, thanks @rprouse

donatocayurin:
Thank you very much ❤️🙏

PR #4296, opened 2023-02-22
Remove Azure DevOps pipeline and update GitHub Action

rprouse:
Back-port #4295 to v3.13-dev, see #4293


stevenaw (approved):
LGTM, diff matches from PR #4295

Issue #3858, opened 2021-05-19
Running tests with ITestAction attributes leaks memory


MatteS75:
Running this example:

`    [SomeTestAttr]
    [TestFixtureSource(nameof(FixtureArgs))]
    public class Tests1
    {
        private static object[] FixtureArgs = Enumerable.Range(0, 100)
            .Select(a => new object[] {$"a{a:000}"})
            .ToArray();

        public Tests1(string arg)
        {
        }

        [Test]
        public void Test1([Range(0, 10)] int x)
        {
            Thread.Sleep(100);
            Assert.Pass();
        }
    }

    public class SomeTestAttrAttribute : Attribute, ITestAction
    {
        private object _data = Enumerable.Range(0, 1000000).ToArray();
        
        public void BeforeTest(ITest test)
        {
        }

        public void AfterTest(ITest test)
        {
        }

        public ActionTargets Targets { get; }
    }
`

From what we can see new attributes are created for each fixture, but never released.

This example is using TestFixtureSource to create many fixtures, but from what we can see the behavior is the same with multiple fixture classes.

stevenaw:
This may've been "fixed" in https://github.com/nunit/nunit/pull/4034 through the introduction of caching the attribute instances

lahma:
The earlier cache only was effective against attributes on methods. Created https://github.com/nunit/nunit/pull/4300 to address caching of test level metadata.

PR #4300, opened 2023-02-26
Add TestMetadataCache to improve CompositeWorkItem build speed

lahma:
Addresses high memory usage reported in #3858 . Adds `TestCache` the same way there is `MethodInfoCache` and caches the action attributes just like in method info cache - other cached items could be added later if needed. 

``` ini

BenchmarkDotNet=v0.13.5, OS=Windows 11 (10.0.22621.1343/22H2/2022Update/SunValley2)
.NET SDK=6.0.114
  [Host]     : .NET 6.0.14 (6.0.1423.7309), X64 RyuJIT AVX2
  DefaultJob : .NET 6.0.14 (6.0.1423.7309), X64 RyuJIT AVX2


```

## Before

|          Method |     Mean |   Error |  StdDev |     Gen0 |     Gen1 |     Gen2 | Allocated |
|---------------- |---------:|--------:|--------:|---------:|---------:|---------:|----------:|
| TestActionUsage | 161.4 ms | 3.19 ms | 5.76 ms | 750.0000 | 500.0000 | 500.0000 | 386.22 MB |

## After

|          Method |     Mean |     Error |    StdDev |     Gen0 |     Gen1 | Allocated |
|---------------- |---------:|----------:|----------:|---------:|---------:|----------:|
| TestActionUsage | 8.975 ms | 0.1141 ms | 0.1068 ms | 281.2500 | 140.6250 |   4.71 MB |

### DIFF
--- a/src/NUnitFramework/benchmarks/nunit.framework.benchmarks/CompositeWorkItemBenchmark.cs
+++ b/src/NUnitFramework/benchmarks/nunit.framework.benchmarks/CompositeWorkItemBenchmark.cs
@@ -1,0 +1,86 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using BenchmarkDotNet.Attributes;
+using NUnit.Framework.Api;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+using NUnit.Framework.Internal.Abstractions;
+using NUnit.Framework.Internal.Execution;
+
+namespace NUnit.Framework;
+
+[MemoryDiagnoser]
+public class CompositeWorkItemBenchmark
+{
+    private Dictionary<string, object> _loaderOptions;
+
+    [GlobalSetup]
+    public void GlobalSetup()
+    {
+        _loaderOptions = new Dictionary<string, object>
+        {
+            { "ProcessModel", "InProcess" },
+            { "DomainUsage", "None" },
+            { "ShadowCopyFiles", false },
+            { "TestParametersDictionary", "" },
+            { "NumberOfTestWorkers", 0 },
+            { "SynchronousEvents", "false" },
+            { "RandomSeed", 878248866 },
+            { "LOAD", new List<string> { "NUnit.Framework.TestWithTestActionAttribute" } },
+            { "WorkDirectory", Path.GetTempPath() },
+        };
+    }
+
+    [Benchmark]
+    public void TestActionUsage()
+    {
+        DefaultTestAssemblyBuilder builder = new DefaultTestAssemblyBuilder();
+        var test = builder.Build(typeof(TestWithTestActionAttribute).Assembly, _loaderOptions);
+        WorkItem workItem = WorkItemBuilder.CreateWorkItem(test, TestFilter.Empty, new DebuggerProxy(), recursive: true);
+        var context = new TestExecutionContext();
+        context.Dispatcher = new SuperSimpleDispatcher();
+        workItem.InitializeContext(context);
+        workItem.Execute();
+    }
+}
+
+[SomeTestAttr]
+[TestFixtureSource(nameof(_fixtureArgs))]
+public class TestWithTestActionAttribute
+{
+    private static object[] _fixtureArgs = Enumerable.Range(0, 100)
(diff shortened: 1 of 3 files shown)


mikkelbu (commented):
Thanks for providing this @lahma. I'm a bit tired today, so I cannot determine if we will run into some of the same problems as we did for the other cache - see e.g. nunit/nunit#4104 for the solution. 

I'll need to examine this some more

lahma:
> Thanks for providing this @lahma. I'm a bit tired today, so I cannot determine if we will run into some of the same problems as we did for the other cache - see e.g. #4104 for the solution.

Thanks for the review. At least in this case the usage in in single point so if the attribute fails at first use it should throw at that point. Consecutive calls of course get the cached value but like discussed in one of the threads that kind of stateful attribute usage would be considered invalid.

OsirisTerje:
@mikkelbu Have you had time to look into this?   Would be nice to have it merged in.

src/NUnitFramework/framework/Internal/Builders/TestCache.cs
```diff
@@ -0,0 +1,92 @@
...
+                TestActionAttributes = GetActionsForType(testType);
+            }
+
+            public ITestAction[] TestActionAttributes { get; }
+
+            private static ITestAction[] GetActionsForType(Type type)
+            {
+                var actions = new List<ITestAction>();
+
+                if (type != null && type != typeof(object))
```
SimonCropp, line 77: can we waely exist here and do return an array.empty

SimonCropp:
why is this targeting v3. given we are prepping for v4, i would think this could target master?

src/NUnitFramework/framework/Internal/Builders/TestCache.cs
lahma: Done

lahma:
I've changed this to target master.

stevenaw:
Thanks for rebasing this and resolving conflicts @lahma . You were too quick for me to ask. 😁

src/NUnitFramework/benchmarks/nunit.framework.benchmarks/CompositeWorkItemBenchmark.cs
```diff
@@ -0,0 +1,107 @@
...
+    public TestWithTestActionAttribute(string arg) { }
+
+    [Test]
+    public void Test1([Range(0, 10)] int x) { }
+}
+
+[AttributeUsage(AttributeTargets.Class)]
+public class SomeTestAttrAttribute : Attribute, ITestAction
+{
+    private object _data = Enumerable.Range(0, 1000000).ToArray();
```
manfred-brands, line 85: I presume this is to make this a slow attribute in order to see the difference in caching.

manfred-brands (changes requested):
Nice work, more caching.

A few minor comments added.

This does change the behaviour of the actions, but only for badly designed ones that don't put re-initialize stuff in the `BeforeTest` method. More an item to mention in any release notes.

src/NUnitFramework/framework/Internal/Builders/TestCache.cs
```diff
@@ -0,0 +1,78 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace NUnit.Framework.Internal.Builders
+{
+    /// <summary>
+    /// Caches static information for ITestAction to reduce re-calculations and memory allocations from reflection.
+    /// </summary>
+    internal static class TestCache
```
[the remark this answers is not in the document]
lahma: I've renamed the class to `TestMetadataCache` and the field to `Cache`

lahma:
@manfred-brands I've done the fixes you requested, I squashed to single commit keep cleaner, PR title and commit text now reflect the new name `TestMetadataCache`.

manfred-brands (approved):
Thanks for the quick reaction. Nothing more from me

OsirisTerje:
Ok, so we can merge @manfred-brands ?

manfred-brands:
@OsirisTerje I was sitting in the bus when @lahma made his latest changes. Merged now.

PR #4304, opened 2023-03-01
Add support for the ref and in keywords in ValuesAttribute

### LINKED ISSUE #3872: Add support for `ref bool`, `ref bool?`, `in bool`, and `in bool?` when using `NUnit.Framework.ValuesAttribute`
The following type of tests are not executed because they do not produce any data for `[Values]` attribute when parameters are `in bool`, `in bool?`, `ref bool`, or `ref bool?`.
```
public void SomeParameterizedTest([Values] in bool someFlag)
{
  // Test case code here ...
}
```
The test is not reported as failure, or ignored but it is silently not executed.

If the described behavior is not a design decision I would like to add support for the four scenarios (already implemented on my fork at commit [d6f38fadc1c790695720bddde233c4b0fcdb02f4](https://github.com/lsolano/nunit/commit/d6f38fadc1c790695720bddde233c4b0fcdb02f4)). 

As a side note, it will be very helpful to have some kind of warning about those situations.

sdelarosbil:
This closes #3872 

It should allow for `ref`, `ref?`, `in` and `in?` types to be used with the parameterless version of the attribute for both bool and enum types.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/ValuesAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/ValuesAttribute.cs
@@ -102,15 +102,16 @@ namespace NUnit.Framework
                 Array.Copy(enumValues, enumValuesWithNull, enumValues.Length);
                 return enumValuesWithNull;
             }
-            if (targetType.IsEnum)
+            Type actualType = targetType.IsByRef ? targetType.GetElementType() : targetType;
+            if (actualType.IsEnum)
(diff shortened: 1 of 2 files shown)


stevenaw (approved):
LGTM, thanks for your contribution @sdelarosbil ! Appreciate your writing some tests for the change too. Sorry it's taken a few weeks to get to this review

Issue #3761, opened 2021-02-06
Drop pre-conversion of test case/fixture params where .NET supports it implicitly


stevenaw:
NUnit has historically handled and pre-converted some type conversions when passing parameters to test cases/fixtures. These include both native-supported implicit conversions of primitives as per the ECMA spec (ex: `int` -> `long` or `short` -> `int`) and some special case conversions for ease-of-use (like `string` -> `DateTime`). Preconversion of primitives within NUnit of types also implicitly converted by netfx + netcore has also meant some bugs and added complexity:
- https://github.com/nunit/nunit/issues/3125
- https://github.com/nunit/nunit/issues/2638
- https://github.com/nunit/nunit/issues/3219

I think it should be considered to drop this preconversion of NUnit-provided ECMA-defined implicit conversions. This would _not_ include removing ease-of-use custom conversions like `string` -> `DateTime` which _should_ make it a relatively transparent change to consumers. One point of discussion is that this preconversion was originally added for Compact Framework, and that removing it could disrupt any CF consumers (if that is still supported)

I'd like to hear some thoughts on this.

stevenaw:
A recent pull request (#4307 ) for a bugfix caused me to think back to this issue.
@nunit/framework-team @nunit/core-team @oznetmaster Is there interest in this issue proceeding or should it be closed?

PR #4307, opened 2023-03-01
Fix open generic type deductions

### LINKED ISSUE #4243: Type args are not deduced correctly for parameterized fixtures
NUnit 3.13.1

Consider the following test class (from your own codebase):

```
[Category("Generics")]
[TestFixture(100.0, 42)]
[TestFixture(42, 100.0)]
public class DeduceTypeArgsFromArgs<T1, T2>
{
    T1 t1;
    T2 t2;

    public DeduceTypeArgsFromArgs(T1 t1, T2 t2)
    {
        this.t1 = t1;
        this.t2 = t2;
    }

    [TestCase(5, 7)]
    public void TestMyArgTypes(T1 t1, T2 t2)
    {
        Assert.That(t1, Is.TypeOf<T1>());
        Assert.That(t2, Is.TypeOf<T2>());
    }
}
```

Expected type args:
1. `System.Double`, `System.Int32`
2. `System.Int32`, `System.Double`

Actual type args:
1. `System.Double`, `System.Double`
2. `System.Double`, `System.Double`

P.S. I fixed it (and wrote tests) but could not commit to a new branch in your repository so as I could make a pull request. How do I do that?
```
remote: Permission to nunit/nunit.git denied to estrizhok.
fatal: unable to access 'https://github.com/nunit/nunit.git/': The requested URL returned error: 403
```

sdelarosbil:
Closes #4243 

When I investigated why it was resolving to double, double, I found that it would try to get the best common type twice for each type args because of this `IsGenericParameter` condition rather than match them via their underlying type. I found it odd: why would it try to check matches between the first arg with the second type param and the second arg with the first one?

I ran some git blame and I found that this if was added as a result of #358 and while I am not too familiar with the project's history, I did some research and it seems NETCF support was removed in this repository a long time ago. It seems to me that it was added due to generics being weirder to support on that platform, but since it isn't supported anymore, it seems that it should be removed if it causes problems which I think it does here.

### DIFF
--- a/src/NUnitFramework/framework/Internal/TypeHelper.cs
+++ b/src/NUnitFramework/framework/Internal/TypeHelper.cs
@@ -265,7 +265,7 @@ namespace NUnit.Framework.Internal
                 {
                     for (int j = 0; j < arglist.Length; j++)
                     {
-                        if (typeParameters[i].IsGenericParameter || parameters[j].ParameterType.Equals(typeParameters[i]))
+                        if (parameters[j].ParameterType.Equals(typeParameters[i]))
                         {
                             if (!TypeHelper.TryGetBestCommonType(
                                 typeArgs[i],
--- a/src/NUnitFramework/tests/Internal/DeduceTypeArgsFromArgs.cs
+++ b/src/NUnitFramework/tests/Internal/DeduceTypeArgsFromArgs.cs
@@ -23,6 +23,8 @@ namespace NUnit.Framework.Internal
         {
             Assert.That(t1, Is.TypeOf<T1>());
             Assert.That(t2, Is.TypeOf<T2>());
+            Assert.That(t1 is int || t2 is int);
+            Assert.That(t1 is double || t2 is double);
         }
     }
 }


stevenaw:
Thanks for noticing this @sdelarosbil and appreciate the summary of your investigation. On a project as long-lived as this one, many of the current maintainers were involved in some of the earlier work.

To my understanding though, you are right about the historical compact framework support having introduced some extra code. I believe the codebase still contains many places where we pre-convert different arguments as some historically-supported platforms like compact framework didn't do this at runtime themselves. We have an overarching issue https://github.com/nunit/nunit/issues/3761 to track removal of them all if the team decides to.

I'm actually a little unsure about the current state of support for CF, though I believe at a minimum a separate fork is maintain by @oznetmaster for it. It might be good to hear his thoughts on this change if it would also fix this issue. For a bugfix like this, would you be able to please add a test as well? It would help prevent future accidental functional regressions.

rprouse:
This looks good but can you add some unit tests?

sdelarosbil:
Done, I found a way to ensure the existing test cover this case.

OsirisTerje (approved):
LGTM now.

PR #4316, opened 2023-03-22
Handle wraparound for Random and Distinct

### LINKED ISSUE #4308: Random attribute with Distinct and wide range causes test to disappear
# Steps to reproduce

Create a new NUnit test project with this test class:

```csharp
using NUnit.Framework;

namespace Repro;

public class Tests
{
    [Test]
    public void Test1(
        [Random(int.MinValue / 2, int.MaxValue / 2 + 1, count: 10, Distinct = true)] int i)
    {
        Assert.Pass();
    }
}
```

Run it with `dotnet test`.

# Expected behaviour
The test runs and passes ten time.

# Actual behaviour
No tests run.

```
No test is available in [redacted]\Repro\Repro\bin\Debug\net7.0\Repro.dll.
Make sure that test discoverer & executors are registered and platform & framework version settings are appropriate
and try again.
```

# More information
This seems to be related to the requested range of the random values. If you remove the `+ 1` the tests execute as expected:

```csharp
[Test]
public void Test1(
    [Random(int.MinValue / 2, int.MaxValue / 2, count: 10, Distinct = true)] int i)
{
    Assert.Pass();
}
```

Alternatively, if you leave out the `Distinct` requirement, the tests also run as expected:

```csharp
[Test]
public void Test1(
    [Random(int.MinValue / 2, int.MaxValue / 2 + 1, count: 10)] int i)
{
    Assert.Pass();
}
```

System information:
- Windows 11 Pro 22H2
- .NET 7

mr-russ:
Closes #4308

### DIFF
--- a/src/NUnitFramework/testdata/RandomAttributeFixture.cs
+++ b/src/NUnitFramework/testdata/RandomAttributeFixture.cs
@@ -33,6 +33,11 @@ namespace NUnit.TestData.RandomAttributeTests
             previousIntValues.Add(x);
         }
 
+        [Test]
+        public void RandomInt_IntRange_DistinctNoWrap([Random(int.MinValue / 2, int.MaxValue / 2 + 1, COUNT, Distinct = true)] int x)
+        {
+            Assert.That(x, Is.InRange(int.MinValue / 2, int.MaxValue / 2 + 1));
+        }
         #endregion
 
         #region Unsigned Int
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/testdata/RandomAttributeFixture.cs
OsirisTerje, line 1: Shouldn't these tests be in the framework.tests project?

mr-russ: That's a good question. I followed the existing pattern for RandomAttribute tests.  And this was the way they were tested.  That file appears in nunit.testdata project in the GUI.   This is a new way to run tests as it's all wrapped up in a Test builder.

OsirisTerje: I was under the impression that was "data" for checking that the tests could be run.  I have the same in the adapter  repo.  But tests that just checks some functionality would (afaic see) be simpler if implemented in the framework.tests project.  I am not 100% sure here, so 
@rprouse  should chime in here.  Also, afaic see, the tests testdata has to be called from framework test project.   But looking at your code, it seems to be part of an existing fixture, so I assume you have checked it actually runs?

OsirisTerje (approved):
LGTM, ref comments

stevenaw (approved):
LGTM thanks for your contribution @mr-russ !
I've pulled the code and the new tests pass locally for me

stevenaw:
@nunit/framework-team @OsirisTerje Any objections to this being backported to 3.13? It seems fairly safe

PR #4317, opened 2023-03-23
Ensure Console.Out is not disposed and is restored if it was replaced during the test run

### LINKED ISSUE #4319: TextRunner accidentally disposes System.Out
It is possible to replace `System.Out` with your own implementation of `TextWriter`.
One reason why you would want to do that is if you want to both capture all output to a stream and write it to console.

I was using this approach and I stumbled upon an issue when using `TextRunner`.
When using `TextRunner`, and not providing a file, `TextRunner` defaults to `System.Out`.
When `TextRunner` is done running it disposes the underlying stream that it was writing to. This is correct behaviour when using files, but not desired behaviour when using `System.Out`.
See: [source code
](https://github.com/nunit/nunit/blob/92180f13381621e308b01f0abd1a397cc1350c12/src/NUnitFramework/nunitlite/TextRunner.cs#L104)

I modified `ConsoleColorWriter` so `Dispose` is a NOP. (The base class disposes the writer provided by `ConsoleColorWriter` = `System.Out`).

I've also added a unit test to verify that the fix works. If you want a code example of what crashes, check the unit test.

normj:
I'm on the AWS .NET team. We got a report from a  user that used `NUnitLite` and got an object disposed exception when using this library in their [AWS Lambda](https://aws.amazon.com/lambda/) functions. Lambda is AWS's serverless compute offering.  

In Lambda we redirect `Console.Out` to our own `StreamWriter` for logging. `NUnitLite` is wrapping `Console.Out` with its own `ColorConsoleWriter` and when it is done it disposes `ColorConsoleWriter` and disposes its inner `TextWriter` which would normally be the console stdout that ignores the dispose call but instead calls our `TextWriter` to dispose. With our `TextWriter` disposed any calls to `Console.Out` now fail.

We have [fixed Lambda](https://github.com/aws/aws-lambda-dotnet/pull/1464) to ignore the dispose call but I though this PR would be helpful for others to ensure `NUnitLite` doesn't accidently dispose of somebody else's custom `Console.Out`.  

I also made sure if `Console.SetOut` or `Console.SetError` are used to be sure to restore to the original `TextWriter`s after the run is complete.

Fixes #4319

### DIFF
--- a/src/NUnitFramework/nunitlite.tests/ExtendedTextWrapperTests.cs
+++ b/src/NUnitFramework/nunitlite.tests/ExtendedTextWrapperTests.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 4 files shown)


OsirisTerje:
1. @normj   Would be nice to see a test around this.
2. @stevenaw @normj Should this also be backported to 3.13 ?

stevenaw:
I think this could be helpful for 3.13. Redirecting or splitting these outputs has been a bit of a common request in issues and discussions the last few months.

OsirisTerje:
@normj Would you mind creating another PR for merging into the https://github.com/nunit/nunit/tree/v3.13-dev branch?

normj:
:man_facepalming: Not adding tests on a PR for a test framework

@OsirisTerje I added a test to cover the `ColorConsoleWriter`. I made sure the test failed without this PRs code changes for `ColorConsoleWriter`. I'm not sure how to add a test for the change in the `Execute` method of `TestRunner` since that is the code executing the tests. Appreciate any suggestions there.

 I'm happy to make a make a backporting PR to the `v3.13-dev` branch once I know I got this PR right.

PR #4318, opened 2023-03-24
Fixed issue where TextRunner accidentally disposes System.Out

jeroenvervaeke:
It is possible to replace `System.Out` with your own implementation of `TextWriter`.
One reason why you would want to do that is if you want to both capture all output to a stream and write it to console.

I was using this approach and I stumbled upon an issue when using `TextRunner`.
When using `TextRunner`, and not providing a file, `TextRunner` defaults to `System.Out`.
When `TextRunner` is done running it disposes the underlying stream that it was writing to. This is correct behaviour when using files, but not desired behaviour when using `System.Out`.
See: [source code
](https://github.com/nunit/nunit/blob/92180f13381621e308b01f0abd1a397cc1350c12/src/NUnitFramework/nunitlite/TextRunner.cs#L104)

I modified `ConsoleColorWriter` so `Dispose` is a NOP. (The base class disposes the writer provided by `ConsoleColorWriter` = `System.Out`).

I've also added a unit test to verify that the fix works. If you want a code example of what crashes, check the unit test.

### DIFF
--- a/src/NUnitFramework/nunitlite.tests/TextRunnerTests.cs
+++ b/src/NUnitFramework/nunitlite.tests/TextRunnerTests.cs
@@ -1,0 +1,57 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using System;
+using System.IO;
+using System.Text;
+using NUnit.Framework;
+
+namespace NUnitLite.Tests
+{
+    [TestFixture]
+    public class TextRunnerTests
+    {
+        [Test]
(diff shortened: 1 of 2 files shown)


jeroenvervaeke:
@dotnet-policy-service agree

stevenaw:
Appreciate your time and effort on this @jeroenvervaeke . I'm going to close this one as a duplicate of Norm's as I suspect the detection of `Console.Out` being changed in that PR may make it a safer fix with fewer chance of edge cases. Thanks again for having looked into and submitted a fix for this in your spare time.

Issue #3432, opened 2019-11-21
Assert.That is blocking and might lead to deadlock when used with WCF.


MikiKaufhold:
### Problem 1)
Suppose you are asserting an asynchronous WCF service call - for the sake of simple example just its completion (for complete implementation see [NUnitUsingWCFClientDeadlockReproducer.cs.txt](https://github.com/nunit/nunit/files/3873451/NUnitUsingWCFClientDeadlockReproducer.cs.txt))

```
Assert.That(async () =>
{
  await AsyncCall();
  return true;
}, Is.True);
```
In certain cases this lead to deadlock…

### Analysis 1)
There is a [great article where Alois explains a case of similar WCF deadlock](https://aloiskraus.wordpress.com/2018/07)

> The problem is that WCF runs asynchronous method completions on the WCF channel dispatcher which seems to be single threaded just like a UI application with a message pump.
> …
>  A sync/async mixture of remoted methods will likely cause deadlocks

Interestingly, our WCF service has only Async methods…
However, `Assert.That` ends up with a blocking `Task.Wait()` call and (likely) causes the deadlock.

### Solution 1)
To resolve this we have introduced an async variant for `Assert.That` using `DelayedConstraint` that looks like this:
`public static Task DelayedAsync<T>(Func<Task<T>> actualValueDelegate,
													IResolveConstraint expression,
													int delayInMilliseconds)`
 (for complete implementation see [DelayedAsync.cs.txt](https://github.com/nunit/nunit/files/3873527/DelayedAsync.cs.txt))

### Problem 2) - Considering upgrade of NUnit from 2.6.6 -> 3.11
We are running into the issue generally known as "Catching AssertionException fails tests" #2758 but also mentioned in #2043, #2040, #2007.

### Analysis 2)
As suggested in those threads,
`using (new NUnit.Framework.Internal.TestExecutionContext.IsolatedContext())` in the most inner DelayedAsync call resolves the issue.

However, since catching AssertionExceptions is not recommended practise and  TestExecutionContext is Internal, I am hesitant to adopt this as a final solution.

Instead I would wish for a native awaitable Assert.That that would implement [**async all the way** pattern as recommended by Stephen Cleary](https://docs.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming).

Looking forward to your feedback and suggestions.

jnm2:
[...]

Issue #2843, opened 2018-05-02
Replacing ThrowsAsync with a composable async alternative


jnm2:
Today's Assert.ThrowsAsync and related methods do **sync-over-async**. This means they convert an async operation into a synchronous one. In other words, they block until the task is complete and return `void` or the return value directly, rather than returning an awaitable type themselves such as `Task ThrowsAsync` or `Task<Exception> CatchAsync`.

#### What does the problem look like?

```cs
public class Assert
{                Problem: not async Task
                    ⬇
    public static void ThrowsAsync(AsyncTestDelegate code) { … }

                Problem: no async Task ThatAsync
                    ⬇
    public static void That(ActualValueDelegate<Task> code) { … }
}

      Problem: not async Task
[Test]   ⬇
public void TestMethod()
{
  Problem: no `await`
    ⬇
    Assert.ThrowsAsync(async () =>
    {
        await Task.Delay(10_000); // Fiddle while Rome burns, why don’t ya 😇
    });

  Problem: no `await`
    ⬇
    Assert.That(async () => await …, Throws.InstanceOf<FooException>());
}
```

#### Why is this a problem?

Waiting synchronously for an async operation is a antipattern with real-world consequences. The consequences range from losing scalability (if you're doing `async` right, [there is no thread](https://blog.stephencleary.com/2013/11/there-is-no-thread.html)) to outright deadlocks and other logic errors.

Well, in 3.11, I fixed the deadlock issue if you instantiate Windows Forms or WPF controls in your tests. So what then, **why not** keep doing sync-over-async in `Assert`?

 - First, deadlocks are still a threat. I only fixed deadlocks on *recognized* message-pumping SynchronizationContexts, and even that hasn't been stress-tested for reentry with assert delegates nested inside assert delegates. We shouldn't *have* to go to the length of testing sync-over-async-over-sync-over-async; our APIs shouldn't be allowing it like this.

 - Second, without a good reason, it goes in the face of hard-learned good practices with async/await. The async-await paradigm is complex enough for those who are still becoming familiar with it that we should not be requiring them to code their tests around this antipattern. `Assert.*Async` is doing nothing which justifies an exception. It's one thing for NUnit as a _framework_ to do sync-over-async with the entry point, the test method itself—this is exactly like C# 7.1's `async Task Main`—but it's quite another thing for NUnit as a _library_, in `Assert.*Async`, to be doing sync-over-async as well.

#### What does the solution look like?

[...]

PR #4322, opened 2023-03-31
ThatAsync and MultipleAsync

uecasm:
Because #2843 and #3432 made me sad.

This introduces `Assert.ThatAsync` (which accepts `Is` or `Throws` constraints as usual) and `Assert.MultipleAsync` (which can contain both `ThatAsync` and synchronous assertions), and does all the right things with them (at least as far as the test method boundary), provided that you `await` them inside an `async Task` test.

Not included, but I was pondering whether it might be a good idea: constraint `Is.Canceled` and/or `Throws.Canceled` which is equivalent to `Throws.InstanceOf<OperationCanceledException>()`.

### DIFF
--- a/src/NUnitFramework/framework/Assert.ThatAsync.cs
+++ b/src/NUnitFramework/framework/Assert.ThatAsync.cs
@@ -1,0 +1,84 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+#nullable enable
+
+using System.Runtime.ExceptionServices;
+using System.Threading.Tasks;
+using System;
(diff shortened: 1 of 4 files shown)


uecasm:
@dotnet-policy-service agree

OsirisTerje:
@stevenaw I have read through the earlier discussions on this issue, and the implementation here looks very straightforward.  I can't see any side-effects nor breaking changes here.  Do you agree?

@jnm2  You had a long thread on this topic many years ago.  Do you agree this PR looks ok?

OsirisTerje (approved):
LGTM
And this is additions, nothing existing is changed, afaic see.

PR #4325, opened 2023-04-03
Backport PR 4317 - Ensure Console.Out is not disposed and is restored

normj:
PR backporting https://github.com/nunit/nunit/pull/4317

In AWS Lambda we redirect `Console.Out` to our own `StreamWriter` for logging. `NUnitLite` is wrapping `Console.Out` with its own `ColorConsoleWriter` and when it is done it disposes `ColorConsoleWriter` and disposes its inner `TextWriter` which would normally be the console stdout that ignores the dispose call but instead calls our `TextWriter` to dispose. With our `TextWriter` disposed any calls to `Console.Out` now fail.

I also made sure if `Console.SetOut` or `Console.SetError` are used to be sure to restore to the original `TextWriter`s after the run is complete.

### DIFF
--- a/src/NUnitFramework/nunitlite.tests/ExtendedTextWrapperTests.cs
+++ b/src/NUnitFramework/nunitlite.tests/ExtendedTextWrapperTests.cs
@@ -1,4 +1,4 @@
-﻿// ***********************************************************************
+// ***********************************************************************
 // Copyright (c) 2015 Charlie Poole, Rob Prouse
 //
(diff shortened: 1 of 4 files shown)


OsirisTerje (approved):
LGTM

OsirisTerje:
Thanks, @normj !!
Appreciate the work and the quick response!

normj:
Awesome, great experience contributing to the library.

PR #4330, opened 2023-04-04
RetryUntilFailureAttribute

barakadax:
Added attribute to run test until failure and added test to prove it's working

### DIFF
--- a/src/NUnitFramework/framework/Attributes/RetryUntilFailureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RetryUntilFailureAttribute.cs
@@ -1,0 +1,75 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+#nullable enable
+
+using System;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+using NUnit.Framework.Internal.Commands;
+
+namespace NUnit.Framework.Attributes
+{
+    /// <summary>
+    /// Specifies that a test method should be rerun until failure.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
+    public class RetryUntilFailureAttribute : NUnitAttribute, IRepeatTest
+    {
+
+        #region IRepeatTest Members
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


OsirisTerje:
An attribute like this may cause the test to run forever, causing the test system to hang seemingly.  It may also end up with different runtime issues.  The other Retry attribute is for retrying a maximum number of times if it fails.  That is safe to do.  
This one could have disastrous effects if used with no way to stop it, without killing the test process.  It should at least have a timeout or maximum number.  
You could however look at extending NUnit with this functionality, which would solve your special problem, and you could do whatever you would like, you can have a look here:  https://docs.nunit.org/articles/nunit/extending-nunit/ICommandWrapper-Interface.html

barakadax:
Because I kind of need the infinite run for cases my bug won't replicate for a week or so I listened to you and added max retires mechanism to my code.

barakadax:
@dotnet-policy-service agree

stevenaw:
Thanks for suggesting this change @barakadax 
I'm away from a computer the next few days and I'd appreciate being able to look at this on a screen larger than my phone. Out of curiosity, can you explain a bit more about how you use this? I'd be curious about how a test that could truly run forever could work in a suite combined with any other tests as I think it could prevent a test report from being written.

stevenaw:
If this is added, I think it might make more sense to modify the existing Retry attribute as we're effectively just making the existing "tryCount" optional, though I think it would also make sense for this to be explicitly opt-in instead of using a default constructor to trigger the behaviour. Though I'd like to better understand the use case before actually suggesting any changes. It would also be good to give the team a chance to weigh in on if this feature should be supported before you go to the effort of making more changes

barakadax:
I don't plan ever to merge code with this attribute, you're right about writing test logs and not stopping the tests from running, but I would use this to test flakiness, I have code that fails once a day/week and keep it running might make it happen more so I can debug easier, instead of adding for each test a loop with try catch and variable I think this is more suitable solution and might help others in the future solve flakiness

OsirisTerje:
I suggested making this into an extension instead.  That would make more sense to me.  THat also means the package with this extension is gone, the attribute will cause a compiler error, and thus one is more sure this is not going into test-production code.

stevenaw:
Oh thanks @OsirisTerje I may have missed that point. Sorry about that

PR #4334, opened 2023-04-08
remove Microsoft.NETFramework.ReferenceAssemblies

SimonCropp:
i dont think it is required?


manfred-brands (approved):
As it compiles without it, it seems that this indeed is not needed.

PR #4335, opened 2023-04-08
remove System.Runtime

SimonCropp:
i dont think it is required


manfred-brands (approved):
Already pulled in by default.

mikkelbu (approved):
LGTM

PR #4336, opened 2023-04-08
remove WindowsBase

SimonCropp:


manfred-brands (approved):
We should indeed do without this.

PR #4337, opened 2023-04-08
update System.Security.Principal.Windows and Microsoft.Win32.Registry

SimonCropp:


manfred-brands (approved):
This changes the dependencies.
If a end-user's code depends on the older version, their code no longer compiles without updating.
However as this is for the v4 I don't see this as a problem.

PR #4338, opened 2023-04-08
FSharp.Core 7.0.200

SimonCropp:


stevenaw (approved):
Thanks @SimonCropp !
I recall us being constrained to a particular version in the past, though it looks like CI is passing all tests with this change. Perhaps the issue has since resolved (or was specific to a particular CI we no longer use)

PR #4340, opened 2023-04-08
make indents consistent

SimonCropp:


manfred-brands:
I'm all for consistency. The best way to ensure that it is done once and will stay that way is for builds to fail if not properly indented. This I have done previously using  StyleCop. For any repository with different contributors it is the only way to keep consistent look and feel.

@SimonCropp were these the only files inconsistent or just some files you were working on?

OsirisTerje:
Just to mention it:  @jnm2 became fed-up with commenting on my small "mistakes" of that sort in the adapter, and added in Stylecop there.  Using that and also a good editorconfig file solved those issues, so I am all for adding Stylecop here too.

manfred-brands:
@SimonCropp @OsirisTerje I have added a commit to enable StyleCop, but only have enabled a single rule: SA1137.
This showed a few more places where indentation was off which I have fixed in a 2nd commit.

If you're happy with this, then after the merge of this branch we can enable other rules, one at a time in separate PRs until we have a consistent code base.

SimonCropp:
LGTM. although i cant approve my own PR

manfred-brands (approved):

mikkelbu (approved):
LGTM

PR #4341, opened 2023-04-08
update sdk and langversion

SimonCropp:


OsirisTerje:
I am not sure this will work with appveyor, and we still need that, since the pushing to myget goes from them.

manfred-brands:
With appveyor you are dependent on them what version of .net sdk is installed. Instead of demanding a very specific version, it is better to use a global.json containing 7.0.100 with a roll forward of latest feature. The global.json must match the SDK installed with .GitHub/workflow which is 7.0.x.

@OsirisTerje can we not do a dotnet push to myget from the GitHub action instead and fase out appveyor. It would be better if we could limit to one ci/cd eco system.

OsirisTerje:
@manfred-brands 190% agree!   We just need to get that done too.

OsirisTerje (changes requested):
We need to get the appveyor build decommissioned, and get the push to myget from our github action builds.

OsirisTerje:
@SimonCropp Can you fix this?

Issue #4264, opened 2022-12-02
Assert.Multiple method should fail only if a contained assertion failed


rockstome:
Here is code

```
namespace Tests
{
    public class NUnit
    {
        [TearDown]
        public void TearDown()
        {
            System.Console.WriteLine("Teardown 1");
            Assert.Multiple(() =>
            {
                Assert.That(true);
            });
            System.Console.WriteLine("Teardown 2");
        }

        [Test]
        public void Test()
        {
            System.Console.WriteLine("Test 1");
            Assert.That(false);
            System.Console.WriteLine("Test 2");
        }
    }
}
```

What is very strange for me, line where "Teardown 2" should be printed is not executed. Why is this happening? Assertion in teordown is not throwing any exception

stevenaw:
Thanks @rockstome this one is particularly curious. It looks like `Assert.Multiple` is detecting if any assertion failed in the test and not just what was contained in the multiple block itself. For example, the following will output `"Test 1"` and `"Test 2"` but not `"Test 3"`

```csharp
[Test]
public void Test()
{
  System.Console.WriteLine("Test 1");
  
  try
  {
      Assert.That(false);
  }
  catch { }
  
  System.Console.WriteLine("Test 2");
  Assert.Multiple(() => { });
  System.Console.WriteLine("Test 3");
}
```

That said, I'm not sure if I'd call this a bug per se or that it's working by-design. One almost has to "force" this to fail wholly within a test by suppressing a previous assertion failure from outside the multiple block like in the example. I don't know if it was ever considered how it would behave with assertions in a `TearDown` but based on [rprouse's comment here](https://github.com/nunit/nunit/issues/3957#issuecomment-941727061) asserting from a teardown is something that's discouraged.

Your example is obviously a minimal repro which is appreciated, but to help understand how you originally hit this: is there something you're needing to assert in your TearDown that can't be handled elsewhere?

manfred-brands:
Changing your Test to not fail, causes both `Test 2` and `Teardown 2` to be printed.
Making your Test fail again and replacing the `Assert.Multiple` with the `Assert` stops printing `Test 2` as expected, but still prints `Teardown 2`.
It therefore looks as if the failure status is not reset and causes `Assert.Multiple` to report the same failure again.
Assert.Multiple reports if there are any failures, regardless if they existed before calling Assert.Multiple:
```csharp
            if (context.MultipleAssertLevel == 0 && context.CurrentResult.PendingFailures > 0)
```

[...]

PR #4344, opened 2023-04-10
Add a test to verify Assert.Multiple only throws for failures in current scopes

### LINKED ISSUE #4331: Add testing to "Accumulate further failures if any on AssertMultiple instead of throwing"
Need to add proper testing to verify PR #4313 , issue #4264

The test to be added, so that it works, is https://github.com/nunit/nunit/issues/4264#issuecomment-1336297152

stevenaw:
Resolves #4331 
Based on the repro from https://github.com/nunit/nunit/issues/4264#issuecomment-1336297152


mikkelbu (approved):
LGTM

PR #4133, opened 2022-06-21
Port caching of method discovery to main

stevenaw:
Fixes #4053
Fixes #4096 in main
Fixes #4107 in main
Fixes #4426

Migrates:
- #4034
- #4104

PR #4347, opened 2023-04-11
WIP: Revert "remove System.Runtime"

stevenaw:
This reverts commit 1981e2b510bbdd00a071b2aa4e430ce7b99d92e1. We found it was needed at runtime in the `AsyncToSyncAdapter.IsAsyncOperation`.

https://github.com/nunit/nunit/blob/aeed2b0/src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs#L12-L18

Making this as a draft for now for discussion

### DIFF
--- a/src/NUnitFramework/framework/nunit.framework.csproj
+++ b/src/NUnitFramework/framework/nunit.framework.csproj
@@ -8,6 +8,7 @@
 
   <ItemGroup>
     <PackageReference Include="Nullable" Version="1.3.1" PrivateAssets="all" />
+    <PackageReference Include="System.Runtime" version="4.3.1" />
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">


stevenaw:
@OsirisTerje @manfred-brands @SimonCropp 
Making this as a draft PR for now to continue discussion from https://github.com/nunit/nunit/pull/4133#issuecomment-1502937881

@jnm2 tagging you too as I think you wrote this originally. Based on the original commit it sounds like this might prevent "types other than Task from being awaited" https://github.com/nunit/nunit/commit/8df37f2024108b8d53d5dbbd7a8d3e518ad2e54a. Can you confirm? I'd like to add a test for this if we don't have one already.

stevenaw:
Ok thanks @manfred-brands I suppose I got a little concerned about your comment that "on some platforms the AsyncStateMachineAttribute type is not found".🙂 Did you find this broke for you, or was the null check simply to remove a nullability compiler  warning?

OsirisTerje:
@manfred-brands I also read you as "Not working" on certain platforms.  If that is the case, it may not turn up in the unit tests we have now.  So, please elaborate and enlighten us !

manfred-brands:
The appveyor build for the caching PR was failing with NullReference exceptions. The other builds were green. I don't know what is different with the appveyor build.

OsirisTerje:
@manfred-brands Ahh, the appveyor builds have been more sensitive and the servers they run on are not as often updated, as the others are.  Afaik they are Linux servers, not sure what kind of Linux though.

stevenaw:
Digging into this a bit more, it looks like that line which threw a nullref was originally added to support custom awaitables in #3095 . It looks like we have some tests for this, which appear to be passing. For example: https://github.com/nunit/nunit/blob/master/src/NUnitFramework/tests/AwaitableReturnTypeTests.cs

The CI tests seem fine, but the nullref exception is concerning. Still, I'm happy to close this if we feel it's not needed.

@OsirisTerje @manfred-brands what should we do here?

manfred-brands:
I found the offending build on [appveyor](https://ci.appveyor.com/project/CharliePoole/nunit/builds/46739429)
The test that had the `NullReferenceException` are in NUnitLite's nunit test which I presume don't have any asynchronous tests:
```
Invalid : NUnitLite.Tests.CommandLineTests.GetArgsFromFiles_FailureTests
Failure building Test
System.ArgumentNullException : Value cannot be null.
Parameter name: attributeType
   at System.Reflection.RuntimeMethodInfo.IsDefined(Type attributeType, Boolean inherit)
   at NUnit.Framework.Internal.AsyncToSyncAdapter.IsAsyncOperation(MethodInfo method)
   at NUnit.Framework.Internal.Builders.MethodInfoCache.TestMethodMetadata..ctor(IMethodInfo method)
   at NUnit.Framework.Internal.Builders.MethodInfoCache.<>c.<Get>b__1_0(IMethodInfo m)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at NUnit.Framework.Internal.Builders.MethodInfoCache.Get(IMethodInfo method)
   at NUnit.Framework.Internal.Builders.DefaultTestCaseBuilder.BuildFrom(IMethodInfo method, Test parentSuite)
```

No other tests seem to have been failing.
So I still think the code as is is fine, so we can close this PR.

stevenaw:
Ok thanks! Closing it now

Issue #4254, opened 2022-11-13
Revisit CollectionTally performance for homogenous collections


stevenaw:
Constraints which use the `CollectionTally` class to do their work will currently take O(n * m) time, which can turn into O(n^2) when comparing same-sized collections as is common with collection equivalency. As a result, a few issues have been filed over the years about performance on large collections.

Some ideas were discussed in https://github.com/nunit/nunit/issues/2799 mostly geared at using `HashSet` and a comparer when possible to bring it down to O(n + m) time though this is easiest when the collections being compared all have the same type and tolerance isn't required for comparison. The homogenous types scenario is likely a fairly common one. Some CI tests in the `LargeStringCollection` family would also benefit from this as they can occasionally exceed what we've determined to be an "acceptable" timeout. 

We should look at what we can do to help improve this. Some previous work was done to speed up the unique constraint which may be reusable here

stevenaw:
The idea from #3985 could also likely help here.

stevenaw:
@OsirisTerje I am tentatively putting this in 4.6 as I've been able to demonstrate some good improvements for collections of certain types such as `int` and `byte`. I'm still wrapping up some parts of it such as understanding if it's just _blittable_ types which show the gains, as well as adjusting and refining tests

eXpl0it3r:
This seems to introduce a regression / breaking change with implicit type conversion / cross-type coercion.

Is this intentional? If so, should this be documented in the changelog?

Works on 4.5.1
```c#
var listOfLongs = new List<long>{1L, 2L, 3L};
Assert.That(listOfLongs, Is.EquivalentTo(new[] { 1, 2, 3 }));
```

Fails on 4.6.0 with
```
Assert.That(listOfLongs, Is.EquivalentTo(new[] { 1, 2, 3 }))
  Expected: equivalent to < 1, 2, 3 >
  But was:  < 1, 2, 3 >
  Missing (3): < 1, 2, 3 >
  Extra (3): < 1, 2, 3 >
```

Workaround / Fix: Use the same type on both sides
```c#
var listOfLongs = new List<long>{1L, 2L, 3L};
Assert.That(listOfLongs, Is.EquivalentTo(new[] { 1L, 2L, 3L }));
```

<details>
<summary>Claude Analysis</summary>
  Here's what changed in CollectionTally<T> (a new class added in 4.6.0 for performance):

  // 4.6.0 - new path for the fallback case (T=object, expected=int[])
[...]

PR #4348, opened 2023-04-11
Convert ArrayList usage to List<T> in CollectionTally

stevenaw:
Contributes to #4254 and #53

Something I've been wanting to do for a while, that is easier now that the framework isn't built against some older TFMs.

We've historically used `ArrayList` to track missing items in the CollectionTally class in order to use the in-place sorting. Both [`List<T>`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.sort?view=net-7.0#system-collections-generic-list-1-sort) and [`ArrayList`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.arraylist.sort?view=net-7.0#system-collections-arraylist-sort) use the same in-place sorting algo though `List<T>` supports using a generic underlying collection while `ArrayList` does not. This "generics all the way down" option should eventually make things easier for us to bring "generics all the way up" and start supporting generic constraints.

`CollectionTally` has recently been a [source](https://github.com/nunit/nunit/pull/3881) of [type-specific](https://github.com/nunit/nunit/pull/4097) [edgecases](https://github.com/nunit/nunit/pull/3978) driven partly by a few attempts to [improve](https://github.com/nunit/nunit/pull/3834) [performance](https://github.com/nunit/nunit/issues/2799). I'm hopeful that moving to a generic underlying collection could help avoid some of these issues while also opening doors to implementing `IEqualityComparer` on our internal classes to better integrate with some of the BCL sorting functions as [suggested here](https://github.com/nunit/nunit/issues/2799#issuecomment-379866290).

I've kept the implementation simple for now, but perhaps in the future a `CollectionTally<T>` could be created

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionTally.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionTally.cs
@@ -41,7 +41,7 @@ namespace NUnit.Framework.Constraints
                 foreach (var o in _missingItems)
                     missingItems.Add(o);
 
-                List<object> extraItems = new List<object>(_extraItems.Count);
+                var extraItems = new List<object>(_extraItems.Count);
                 if (_sorted)
                 {
                     for (int index = _extraItems.Count - 1; index >= 0; index--)
@@ -56,7 +56,7 @@ namespace NUnit.Framework.Constraints
             }
         }
 
-        private readonly ArrayList _missingItems = new ArrayList();
+        private readonly List<object> _missingItems = new List<object>();
 
         private readonly List<object> _extraItems = new List<object>();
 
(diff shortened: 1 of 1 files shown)


src/NUnitFramework/framework/Constraints/CollectionTally.cs
```diff
@@ -126,9 +126,9 @@ void TryRemoveSlow(IEnumerable c)
             }
         }
 
-        private static ArrayList ToArrayList(IEnumerable items)
+        private static List<object> ToList(IEnumerable items)
         {
-            var list = items is ICollection ic ? new ArrayList(ic.Count) : new ArrayList();
+            var list = items is ICollection ic ? new List<object>(ic.Count) : new List<object>();
 
             foreach (object o in items)
```
OsirisTerje, line 133: Can you use AddRange here?

OsirisTerje (approved):
Nice! Good to get rid of ArrayList!

src/NUnitFramework/framework/Constraints/CollectionTally.cs
stevenaw: Not yet, unfortunately. AddRange takes an `IEnumerable<T>` but the non-generic CollectionTally currently only receives an `IEnumerable`

stevenaw:
Thanks for the review @OsirisTerje !

PR #4350, opened 2023-04-13
Change release builds to be optimized

lahma:
As discussed in #4149 , seems like a good time with 4.0 nearing to try out again distributing optimized release builds. With this PR removing the forced no-optimize and BenchmarkDotNet project proves that artifacts can now be consumed as release builds (no longer need to disable validator).

Quick benchmark test for changes. Shows > 10% time reduction in best cases (very feature-limited benchmarks of course).

``` ini

BenchmarkDotNet=v0.13.5, OS=Windows 11 (10.0.22621.1555/22H2/2022Update/SunValley2)
12th Gen Intel Core i9-12900H, 1 CPU, 20 logical and 14 physical cores
.NET SDK=7.0.203

```

## Before (master)

|                       Method |        Mean |       Error |      StdDev |      Gen0 |     Gen1 |     Gen2 |  Allocated |
|----------------------------- |------------:|------------:|------------:|----------:|---------:|---------:|-----------:|
|        CountTestsEmptyFilter |    138.5 us |     1.89 us |     1.77 us |    4.1504 |        - |        - |    51.6 KB |
| CountTestsWithCategoryFilter |  2,758.3 us |    52.83 us |    49.41 us |  500.0000 |        - |        - | 6172.07 KB |
|    CountTestsWithOrFilterAll |  4,806.7 us |    58.17 us |    51.56 us |  312.5000 | 125.0000 |  23.4375 | 3903.07 KB |
|                    LoadTests | 66,056.0 us | 1,224.24 us | 1,202.37 us | 2125.0000 | 875.0000 | 250.0000 | 24510.9 KB |

## After (this PR)

|                       Method |        Mean |       Error |    StdDev |      Gen0 |      Gen1 |     Gen2 |   Allocated |
|----------------------------- |------------:|------------:|----------:|----------:|----------:|---------:|------------:|
|        CountTestsEmptyFilter |    116.8 us |     1.72 us |   1.61 us |    4.1504 |         - |        - |     51.6 KB |
| CountTestsWithCategoryFilter |  2,332.5 us |    28.56 us |  26.72 us |  500.0000 |         - |        - |  6172.07 KB |
|    CountTestsWithOrFilterAll |  3,921.2 us |    76.52 us |  71.58 us |  312.5000 |  125.0000 |  23.4375 |  3903.08 KB |
|                    LoadTests | 61,599.0 us | 1,057.11 us | 988.82 us | 2222.2222 | 1000.0000 | 333.3333 | 24131.66 KB |

### DIFF
--- a/src/NUnitFramework/tests/Constraints/DefaultConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/DefaultConstraintTests.cs
@@ -20,8 +20,7 @@ namespace NUnit.Framework.Constraints
         [Test] public void Success_Int() => AssertSuccess(default(int));
         [Test] public void Success_NullableInt() => AssertSuccess(default(int?));
         [Test] public void Success_Long() => AssertSuccess(default(long));
-        [Test] public void Success_DefaultStructWithOverriddenEquals() => AssertSuccess(default(StructWithStrangeEquals));
-        [Test] public void Success_ImmutableArray() => AssertSuccess(default(StructWithStrangeEquals));
+        [Test] public void Success_ImmutableArray() => AssertSuccess(default(ImmutableArray<int>));
 
         [Test] public void Failure_NewObject() => AssertFailure(new object());
         [Test] public void Failure_EmptyString() => AssertFailure(string.Empty);
@@ -29,7 +28,6 @@ namespace NUnit.Framework.Constraints
         [Test] public void Failure_BoxedInt_Zero() => AssertFailure((object)0);
         [Test] public void Failure_Int_NonZero() => AssertFailure(1);
         [Test] public void Failure_Double_NaN() => AssertFailure(double.NaN);
-        [Test] public void Failure_NonDefaultStructWithOverriddenEquals() => AssertFailure(new StructWithStrangeEquals { SomeField = 1, EqualsResult = true });
         [Test] public void Failure_ImmutableArray_Empty() => AssertFailure(ImmutableArray.Create<int>());
 
         private void AssertSuccess<T>(T actual)
@@ -47,16 +45,6 @@ namespace NUnit.Framework.Constraints
         {
             var constraintResult = TheConstraint.ApplyTo(actual);
             Assert.That(!constraintResult.IsSuccess);
-        }
-
-        private struct StructWithStrangeEquals
-        {
-            public int SomeField { get; set; }
-            public bool EqualsResult { get; set; }
-
-            public override bool Equals(object obj) => EqualsResult;
-
-            public override int GetHashCode() => 0;
         }
     }
 }
--- a/src/NUnitFramework/Directory.Build.props
+++ b/src/NUnitFramework/Directory.Build.props
@@ -24,7 +24,6 @@
 
   <!-- We always want a good debugging experience in tests -->
   <PropertyGroup>
-    <Optimize>false</Optimize>
     <DebugType>portable</DebugType>
     <DebugSymbols>true</DebugSymbols>
     <SymbolPackageFormat>snupkg</SymbolPackageFormat>
--- a/src/NUnitFramework/benchmarks/nunit.framework.benchmarks/Program.cs
+++ b/src/NUnitFramework/benchmarks/nunit.framework.benchmarks/Program.cs
@@ -1,32 +1,6 @@
-﻿// ***********************************************************************
-// Copyright (c) 2022 Charlie Poole
-//
-// Permission is hereby granted, free of charge, to any person obtaining
-// a copy of this software and associated documentation files (the
-// "Software"), to deal in the Software without restriction, including
-// without limitation the rights to use, copy, modify, merge, publish,
-// distribute, sublicense, and/or sell copies of the Software, and to
-// permit persons to whom the Software is furnished to do so, subject to
-// the following conditions:
-//
-// The above copyright notice and this permission notice shall be
-// included in all copies or substantial portions of the Software.
-//
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
-// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
-// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
-// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
-// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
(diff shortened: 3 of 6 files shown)


manfred-brands (approved):
You are on a roll.
Did you check how it affects end user debugging?

lahma:
I had to change to draft as there seems to be some tests checking that optimizations are off.

> You are on a roll. Did you check how it affects end user debugging?

As discussed in original issue, I would assume debugging experience is the same for NUnit developers when developing in debug mode (IDE setting). I don't see a normal use case of consumers of the NUnit library for debugging say, why Assert behaves they way it does that often. I myself am accustomed to stepping into 3rd party code and if optimized, it's "normal" for debugger stepping behaving a bit wonky. IDEs usually download the symbols etc and use source link to get the code in question.

manfred-brands:
@lahma Any idea why the build failed?

lahma:
So test projects can also be optimized. Had to fix two things:

* `SlowPath_TakenWhenSpecialTypes` verifies code path selection via stopwatch (lovely), so lowered the value from 50ms to 30ms as optimized code is fast(!)
* `DefaultConstraintTests.Success_DefaultStructWithOverriddenEquals()` has different `ToString` on full framework for some reason under optimized build but this is the behavior I believe NUnit consumer will see anyway because the type in question will be optimized based on their project configuration

src/NUnitFramework/tests/Constraints/DefaultConstraintTests.cs
```diff
@@ -20,8 +20,10 @@ public DefaultConstraintTests()
         [Test] public void Success_Int() => AssertSuccess(default(int));
         [Test] public void Success_NullableInt() => AssertSuccess(default(int?));
         [Test] public void Success_Long() => AssertSuccess(default(long));
+#if !NETFRAMEWORK // the error message is different between class framework and newer .NET runtimes with optimized builds
```
stevenaw, line 23: Curious. Do you recall how it's different? I could see stack traces changes, but I'm not sure what else could.
I'm a little concerned that this loses some test coverage on netfx runs. Is there a way we could keep the coverage of the the `DefaultStructWithOverriddenEquals` scenario here on optimized builds?

stevenaw: Oops, I see now that you explained the difference. I'll have to take a closer look here. I'm still curious if we can avoid losing the netfx coverage here

lahma: I was a bit surprised about this too but I would believe this is quite the corner case, I don't think it will much difference when deciphering a failed Assert in test log. As said, I would believe they were already getting the same message if they were using optimized code against the assert.

stevenaw: I've pulled your branch and this test seems to pass for me locally on all platforms, including net462 if I remove the conditional compilation. Does it matter if the `ToString()` differs?

lahma: @stevenaw here's the error if the assert is removed: https://github.com/nunit/nunit/actions/runs/4739738110/jobs/8414851760?pr=4350#step:8:423

```
Errors, Failures and Warnings

1) Failed : NUnit.Framework.Constraints.DefaultConstraintTests.Success_DefaultStructWithOverriddenEquals
  Expected: default
  But was:  NUnit.Framework.Constraints.DefaultConstraintTests+StructWithStrangeEquals
   at NUnit.Framework.Constraints.DefaultConstraintTests.AssertSuccess[T](T actual)
   at NUnit.Framework.Constraints.DefaultConstraintTests.Success_DefaultStructWithOverriddenEquals()

Run Settings
    Number of Test Workers: 2
    Work Directory: D:\a\nunit\nunit\src\NUnitFramework\tests\bin\Release\net462
    Internal Trace: Off
```

manfred-brands: @lahma Your comment says the error message is different, but these tests are supposed to pass, meaning error message is not relevant.
I checked out your branch and removed the `#if !NETFRAMEWORK` and I don't see any test failing.
I pushed your branch without that code to https://github.com/nunit/nunit/pull/4357 to see of it fails on the build server.

manfred-brands: So it does fail on the build server, but not on my machine?
```
dotnet test --configuration=Release --framework=net462 \
--filter="NUnit.Framework.Constraints.DefaultConstraintTests"
```
gives:
`Passed!  - Failed:     0, Passed:    16, Skipped:     0, Total:    16, Duration: 15 ms - nunit.framework.tests.dll (net462)`

lahma: @manfred-brands yes this seems to be somehow machine dependent, should I remove the problematic altogether then?

manfred-brands: The existing implementation calls [RuntimeHelper.Equals](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.runtimehelpers.equals?view=netframework-4.6.2) which is the correct way as it ignores `Equals` overrides: `if both are the same value type and the values of their underlying memory are equal`.
So why does it fail on your machine and the build server, but only if the tests are optimized. As the previous un-optimized version passes.

lahma: To make things even stranger, if I create a new .NET Console application with target `net462`, optimize = true and run in it in release mode: the code passes when I run it in console and in debugger I can see assert is fine.

```c#

var defaultConstraintTests = new DefaultConstraintTests();
defaultConstraintTests.Success_DefaultStructWithOverriddenEquals();
```

image

manfred-brands: I have upgraded my SDK to the same as the build server used, but it is still not failing for me. 
This is scary, code that works only on some machines.
@stevenaw Does this test fail for you?

lahma: I removed the test case but kept the struct as it's referenced in equality test that isn't broken. I force pushed the version - not sure how you want your history but single commit probably cleaner as I guess you don't squash merge commits by default.

manfred-brands: That test case is covered by the ImmutableArray case and the weird Equatable implementation will confuse any reader. Please remove it.

Yes, single commit is preferred by me

lahma: Done and force-pushed.

manfred-brands:
Thanks @lahma

PR #4352, opened 2023-04-14
Convert more files to nullable

### LINKED ISSUE #3376: Nullable Reference Types annotations

manfred-brands:
Fixes #3376


src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
```diff
@@ -236,22 +236,25 @@ private IEnumerable<ITestCaseData> GetTestCasesFor(IMethodInfo method)
                 var field = member as FieldInfo;
                 if (field != null)
                     return field.IsStatic
-                        ? (MethodParams == null ? (IEnumerable)field.GetValue(null)
+                        ? (MethodParams == null ? (IEnumerable?)field.GetValue(null)
```
OsirisTerje, line 239: Here we return a nullable collection, which doesn't really make sense, since it is a collection, and can be empty, which is the same logically (IMHO).  I know the return value is set to nullable, and changing that might have ripple effects through the code, so perhaps not the time to tackle that too.

manfred-brands: We are calling into end-user methods, so we have no idea what they are doing and would need guards anyhow. 
The alternative would be to add an null-suppression (!) operator and cast to `IEnumerable`. 
This method is allowed to return null which is checked against in `GetTestCasesFor`, I therefore choose to cast to `IEnumerable?`

OsirisTerje (approved):
Yet another awesome job!

PR #4354, opened 2023-04-18
Optimize filter operations

lahma:
I was investigating how to make filtering and test case traversal faster as this is what I believe directly correlates with user experience and lag when things are being calculated and discovered. I tried to stay on filtering side for now, but I have other interesting branches as work in progress too 😉 

The `CountTestsWithCategoryFilter` numbers pretty nicely sum up the pitfalls of using LINQ in capturing context memory-wise (from 68MB to 12KB).

* change `CompositeFilter.Filters` to be an array for fast traversal - internal type so signature change should be OK
* replace LINQ usage with non-allocating plain loops
* add `TryGet` to `PropertyBag` which will not allocate on retrieval in contrast to indexer, name OK?
* detect empty filter from string without the need to parse the XML

I'm adding some comments to PR to clarify my suggestions/reasons behind the changes.

Following testing has been done against `Jint.Tests.Test262` test project as the source for discovery, which has the > 60k test cases to traverse.

Please also consider the delta of change instead of the original nano/milliseconds as I have quite beefy desktop - so halving the wait time on older hardware / CI pipelines can shine.

``` ini
BenchmarkDotNet=v0.13.5, OS=Windows 11 (10.0.22621.1631/22H2/2022Update/SunValley2)
.NET SDK=7.0.203

```

## NUnit.Framework.FrameworkControllerBenchmark

| **Diff**|Method|Mean|Error|Allocated|
|------- |-------|-------:|-------|-------:|
| Old |CountTestsEmptyFilter|960.0 μs|4.16 μs|47.38 KB|
| **New** |	| **736.3 μs (-23%)** | **9.67 μs** | **1 B (-100%)** |
| Old |CountTestsWithCategoryFilter|40,061.2 μs|764.54 μs|68642.27 KB|
| **New** |	| **15,705.2 μs (-61%)** | **191.34 μs** | **12974 B (-100%)** |
| Old |CountTestsWithOrFilterAll|72,789.7 μs|548.23 μs|35458.02 KB|
| **New** |	| **71,194.9 μs (-2%)** | **399.37 μs** | **36272121 B (0%)** |
| Old |LoadTests|337,160.2 μs|3,950.07 μs|208299.28 KB|
| **New** |	| **332,760.1 μs (-1%)** | **4,875.89 μs** | **213225540 B (0%)** |

### DIFF
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -344,6 +344,7 @@ namespace NUnit.Framework.Api
             if (!test.IsSuite)
                 return filter.Pass(test) ? 1: 0;
 
+            // Use for-loop to avoid allocating the enumerator
             int count = 0;
             var tests = test.Tests;
             for (var i = 0; i < tests.Count; i++)
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -339,15 +339,16 @@ namespace NUnit.Framework.Api
             _runComplete.Set();
         }
 
-        private int CountTestCases(ITest test, ITestFilter filter)
+        private static int CountTestCases(ITest test, ITestFilter filter)
         {
             if (!test.IsSuite)
                 return filter.Pass(test) ? 1: 0;
 
             int count = 0;
-            foreach (ITest child in test.Tests)
+            var tests = test.Tests;
+            for (var i = 0; i < tests.Count; i++)
             {
-                count += CountTestCases(child, filter);
+                count += CountTestCases(tests[i], filter);
             }
 
             return count;
--- a/src/NUnitFramework/framework/Attributes/TestFixtureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestFixtureAttribute.cs
@@ -192,8 +192,7 @@ namespace NUnit.Framework
             get
             {
                 //return Properties.Get(PropertyNames.Category) as string;
-                var catList = Properties[PropertyNames.Category];
-                if (catList == null)
+                if (!Properties.TryGet(PropertyNames.Category, out var catList))
                     return null;
 
                 switch (catList.Count)
--- a/src/NUnitFramework/framework/Interfaces/IPropertyBag.cs
+++ b/src/NUnitFramework/framework/Interfaces/IPropertyBag.cs
@@ -77,7 +77,15 @@ namespace NUnit.Framework.Interfaces
         bool ContainsKey(string key);
 
         /// <summary>
-        /// Gets or sets the list of values for a particular key
+        /// Tries to retrieve list of values.
+        /// </summary>
+        /// <param name="key">The key for which the values are to be retrieved</param>
+        /// <param name="values">Values, if found</param>
+        /// <returns>true if found</returns>
+        bool TryGet(string key, out IList values);
+
+        /// <summary>
+        /// Gets or sets the list of values for a particular key, initializes new list behind the key if not found.
         /// </summary>
         /// <param name="key">The key for which the values are to be retrieved or set</param>
         IList this[string key] { get; set; }
--- a/src/NUnitFramework/framework/Internal/Execution/WorkItemBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/WorkItemBuilder.cs
@@ -98,11 +98,11 @@ namespace NUnit.Framework.Internal.Execution
                 var xKey = int.MaxValue;
                 var yKey = int.MaxValue;
 
-                if (x.Test.Properties.ContainsKey(PropertyNames.Order))
-                    xKey = (int)x.Test.Properties[PropertyNames.Order][0];
+                if (x.Test.Properties.TryGet(PropertyNames.Order, out var xOrders))
+                    xKey = (int)xOrders[0];
 
-                if (y.Test.Properties.ContainsKey(PropertyNames.Order))
-                    yKey = (int)y.Test.Properties[PropertyNames.Order][0];
+                if (y.Test.Properties.TryGet(PropertyNames.Order, out var yOrders))
+                    yKey = (int)yOrders[0];
 
                 return xKey.CompareTo(yKey);
             }
--- a/src/NUnitFramework/framework/Internal/Filters/AndFilter.cs
+++ b/src/NUnitFramework/framework/Internal/Filters/AndFilter.cs
@@ -30,6 +30,8 @@ namespace NUnit.Framework.Internal.Filters
         /// <returns>True if all the component filters pass, otherwise false</returns>
         public override bool Pass( ITest test, bool negated )
         {
+            // Use foreach-loop against array instead of LINQ for best performance
+
             if (negated)
             {
                 foreach (var filter in Filters)
@@ -61,6 +63,7 @@ namespace NUnit.Framework.Internal.Filters
         /// <returns>True if all the component filters match, otherwise false</returns>
         public override bool Match( ITest test )
         {
+            // Use foreach-loop against array instead of LINQ for best performance
             foreach (var filter in Filters)
             {
                 if (!filter.Match(test))
@@ -79,6 +82,7 @@ namespace NUnit.Framework.Internal.Filters
         /// <returns>True if all the component filters explicit match, otherwise false</returns>
         public override bool IsExplicitMatch( ITest test )
         {
+            // Use foreach-loop against array instead of LINQ for best performance
             foreach (var filter in Filters)
             {
                 if (!filter.IsExplicitMatch(test))
--- a/src/NUnitFramework/framework/Internal/Filters/AndFilter.cs
+++ b/src/NUnitFramework/framework/Internal/Filters/AndFilter.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 #nullable enable
-using System.Linq;
 using NUnit.Framework.Interfaces;
 
 namespace NUnit.Framework.Internal.Filters
@@ -32,9 +31,27 @@ namespace NUnit.Framework.Internal.Filters
         public override bool Pass( ITest test, bool negated )
         {
             if (negated)
-                return Filters.Any(f => f.Pass(test, negated));
+            {
+                foreach (var filter in Filters)
+                {
+                    if (filter.Pass(test, negated))
+                    {
+                        return true;
+                    }
+                }
 
-            return Filters.All(f => f.Pass(test, negated));
+                return false;
+            }
+
+            foreach (var filter in Filters)
+            {
+                if (!filter.Pass(test, negated))
+                {
+                    return false;
+                }
+            }
+
+            return true;
         }
 
         /// <summary>
@@ -44,7 +61,15 @@ namespace NUnit.Framework.Internal.Filters
         /// <returns>True if all the component filters match, otherwise false</returns>
         public override bool Match( ITest test )
         {
-            return Filters.All(filter => filter.Match(test));
+            foreach (var filter in Filters)
+            {
+                if (!filter.Match(test))
+                {
+                    return false;
+                }
+            }
+
+            return true;
         }
 
         /// <summary>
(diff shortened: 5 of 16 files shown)


src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -339,15 +339,16 @@ private void OnRunCompleted(object sender, EventArgs e)
...
 
-        private int CountTestCases(ITest test, ITestFilter filter)
+        private static int CountTestCases(ITest test, ITestFilter filter)
         {
             if (!test.IsSuite)
                 return filter.Pass(test) ? 1: 0;
 
             int count = 0;
-            foreach (ITest child in test.Tests)
+            var tests = test.Tests;
```
lahma, line 348: Tests is interface distpatch to get a collection which shouldn't change, calling only once

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -339,15 +339,16 @@ private void OnRunCompleted(object sender, EventArgs e)
...
+            for (var i = 0; i < tests.Count; i++)
```
lahma, line 349: `for` loop against IList is the best choice as no need for boxed enumerator (IList source isn't known and direct struct enumerator from `GetEnumerator` isn't being used).

src/NUnitFramework/framework/Attributes/TestFixtureAttribute.cs
```diff
@@ -192,8 +192,7 @@ public bool Explicit
             get
             {
                 //return Properties.Get(PropertyNames.Category) as string;
-                var catList = Properties[PropertyNames.Category];
-                if (catList == null)
+                if (!Properties.TryGet(PropertyNames.Category, out var catList))
```
lahma, line 195: indexer allocates a new list on query, so return value would never be null anyway, `TryGet` now allows to return fast

src/NUnitFramework/framework/Interfaces/IPropertyBag.cs
```diff
@@ -77,7 +77,15 @@ public interface IPropertyBag : IXmlNodeBuilder
         bool ContainsKey(string key);
 
         /// <summary>
-        /// Gets or sets the list of values for a particular key
+        /// Tries to retrieve list of values.
+        /// </summary>
+        /// <param name="key">The key for which the values are to be retrieved</param>
+        /// <param name="values">Values, if found</param>
+        /// <returns>true if found</returns>
+        bool TryGet(string key, out IList values);
```
lahma, line 85: not sure about the method name, it isn't a dictionary of key-value

src/NUnitFramework/framework/Interfaces/IPropertyBag.cs
```diff
@@ -77,7 +77,15 @@ public interface IPropertyBag : IXmlNodeBuilder
...
+
+        /// <summary>
+        /// Gets or sets the list of values for a particular key, initializes new list behind the key if not found.
```
lahma, line 88: added warning about the behavior...

src/NUnitFramework/framework/Internal/Execution/WorkItemBuilder.cs
```diff
@@ -98,11 +98,11 @@ public int Compare(WorkItem x, WorkItem y)
                 var xKey = int.MaxValue;
                 var yKey = int.MaxValue;
 
-                if (x.Test.Properties.ContainsKey(PropertyNames.Order))
```
lahma, line 101: no double-lookup (Contains, Get) necessary when using `TryGet`

src/NUnitFramework/framework/Internal/Filters/AndFilter.cs
```diff
@@ -32,9 +31,27 @@ internal sealed class AndFilter : CompositeFilter
         public override bool Pass( ITest test, bool negated )
         {
             if (negated)
-                return Filters.Any(f => f.Pass(test, negated));
```
lahma, line 35: these LINQ `Any` can be a beast as it used to create `Func<>`  instance, new versions are more verbose, but a lot faster

src/NUnitFramework/framework/Internal/Filters/CompositeFilter.cs
```diff
@@ -26,13 +25,13 @@ public CompositeFilter()
...
         {
-            Filters = new ReadOnlyCollection<TestFilter>(filters);
+            Filters = filters;
         }
 
         /// <summary>
         /// Return a list of the composing filters.
         /// </summary>
-        public IList<TestFilter> Filters { get; }
+        public TestFilter[] Filters { get; }
```
lahma, line 34: array is fastest to traverse and supports `foreach` as the fastest alternative, `internal` type so signature change shouldn't matter

src/NUnitFramework/framework/Internal/Filters/InFilter.cs
```diff
@@ -37,17 +37,15 @@ private InFilter(Func<ITest, string?> selector, IEnumerable<string?> values, str
...
+            var hashCommonTypeAndNonRegex = orFilter.Filters.Length > 0;
             Type? commonType = null;
 
             // eagerly build expected value collection
-            var expectedValues = new List<string?>(orFilter.Filters.Count);
+            var expectedValues = new List<string?>(orFilter.Filters.Length);
 
             // use for for faster traversal without boxed enumerator
-            for (var i = 0; i < orFilter.Filters.Count; i++)
+            foreach (var filter in orFilter.Filters)
```
lahma, line 47: `foreach` is now even faster as it operates against array, for would need `uint` casting for loop variable `i` and other magic to avoid bounds checking

src/NUnitFramework/framework/Internal/PreFilter.cs
```diff
@@ -200,7 +199,7 @@ private bool MatchMethodElement(Type type)
             private bool MatchSetUpFixture(Type type)
             {
                 return IsSubNamespace(type.Namespace) &&
-                       type.GetCustomAttributes(typeof(SetUpFixtureAttribute), true).Any();
+                       type.GetCustomAttributes(typeof(SetUpFixtureAttribute), true).Length > 0;
```
lahma, line 202: `Any` against array needs to box via `IEnumerable`, old-school is better here

src/NUnitFramework/framework/Internal/PropertyBag.cs
```diff
@@ -131,14 +142,14 @@ public TNode AddToXml(TNode parentNode, bool recursive)
         {
             TNode properties = parentNode.AddElement("properties");
 
-            foreach (string key in Keys)
+            foreach (var pair in inner)
```
lahma, line 145: far better to enumerate dictionary directly than first getting keys and then getting value behind every key

src/NUnitFramework/framework/Internal/TestFilter.cs
```diff
@@ -208,7 +208,7 @@ private static TestFilter[] GetChildNodeFilters(TNode node)
         /// returns true when called. It never matches explicitly.
         /// </summary>
         [Serializable]
-        private class EmptyFilter : TestFilter
+        private sealed class EmptyFilter : TestFilter
```
lahma, line 211: always seal classes when possible, maybe someone more familiar with dependencies should investigate what could be sealed or even made `internal`

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -512,13 +512,13 @@ public bool ContainsKey(string key)
             {
                 get
                 {
-                    var list = new List<object>();
-                    foreach(var item in _source[key])
+                    if (_source.TryGet(key, out var values))
```
lahma, line 515: fast case for not finding anything now doesn't allocate a List, using `IEnumerable` API as promised

src/NUnitFramework/framework/Internal/TestFilter.cs
```diff
@@ -111,8 +111,14 @@ protected virtual bool MatchDescendant(ITest test)
         /// </summary>
         public static TestFilter FromXml(string xmlText)
         {
-            if (string.IsNullOrEmpty(xmlText))
-                xmlText = "<filter />";
+            const string emptyFilterXml1 = "<filter />";
+            const string emptyFilterXml2 = "<filter/>";
+
+            // check for fast cases
+            if (string.IsNullOrEmpty(xmlText) || xmlText.Length < 11 && xmlText is emptyFilterXml1 or emptyFilterXml2)
```
lahma, line 118: Known empty filter

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
OsirisTerje: This code is operating only on the test (ITest ).  Shouldn't it then be moved to the class implementing that?

src/NUnitFramework/framework/Internal/Filters/AndFilter.cs
```diff
@@ -54,9 +79,13 @@ public override bool Match( ITest test )
         /// <returns>True if all the component filters explicit match, otherwise false</returns>
         public override bool IsExplicitMatch( ITest test )
         {
-            foreach( TestFilter filter in Filters )
-                if ( !filter.IsExplicitMatch( test ) )
+            foreach (var filter in Filters)
```
OsirisTerje, line 82: There are 4 nearly identical loops here.  Can that be simplified?

src/NUnitFramework/framework/Internal/PreFilter.cs
OsirisTerje: But the next dev coming in here will simplify this by moving it to Any again, 3 years down the road or some.   If old-school, it can still be added to some function, generic or standard, with a comment to explain why it is like that.

src/NUnitFramework/framework/Internal/TestFilter.cs
```diff
@@ -111,8 +111,14 @@ protected virtual bool MatchDescendant(ITest test)
         /// </summary>
         public static TestFilter FromXml(string xmlText)
         {
-            if (string.IsNullOrEmpty(xmlText))
-                xmlText = "<filter />";
+            const string emptyFilterXml1 = "<filter />";
```
OsirisTerje, line 114: These two looked very identical to me, until I noticed the extra space there.  Perhaps name them emptyFilterXmlWithSpace  and .....WithoutSpace or something like that?

OsirisTerje (changes requested):
Some loops are moved to foreach, others are moved to for.  I see that all is done for a good reason, but without something that explains why, it will be "improved" later back to probably the most "modern" (aka foreach).

src/NUnitFramework/framework/Internal/Filters/AndFilter.cs
lahma: I think that might become hard to read if we try parametrize the logic, all the Pass == false -> return true inverting etc

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
lahma: That would require adding a new member to `ITest` which would also take filter to allow counting

src/NUnitFramework/framework/Internal/PreFilter.cs
lahma: added comment

src/NUnitFramework/framework/Internal/TestFilter.cs
lahma: changed to more descriptive names

lahma:
Thank you for the review @OsirisTerje ! I tried to address the issues by adding comments why `for`/`foreach` was selected. Some of the remarks would probably require refactoring the class hierarchy / introducing new members so I didn't go there (yet at least).

OsirisTerje:
@lahma 
Introducing new members - feel free to do that!  
Rearranging the hierarchy, well, we do have tests, don't we?  Might work.

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
OsirisTerje: Yes, and ITestFilter is also a public interface, so that would be fine I think.

OsirisTerje (approved):

OsirisTerje:
@lahma Do you want to make more changes here, or is it ready for merge?

lahma:
@OsirisTerje I would be glad if this was good for now, maybe refactoring is for another time. Especially nice would be sealing the types that are not expected to be actually inherited outside of the framework - but probably a case of another PR.

OsirisTerje:
If you like, you could create an issue, label it with is:refactoring, and just list the things you think should be done later.

PR #4356, opened 2023-04-19
Allows Is.AnyOf to be called with arrays or other collections

### LINKED ISSUE #4355: Allow Is.AnyOf to be called with arrays or other collections
Follow-up to a great docs observation on https://github.com/nunit/docs/pull/750 and https://github.com/nunit/nunit.analyzers/issues/533

The `Is.AnyOf` only works if the members are specified as individual values, not if passed as an array or list.

Example Gotcha:

```csharp
[Test]
public void CanBeUsedWithParameters()
{
    Assert.That(42, Is.AnyOf(0, -1, 42, 100)); // Passes
}

[Test]
public void CantBeUsedWithAnArray()
{
    var array = new[] { 0, -1, 42, 100 };
    Assert.That(42, Is.AnyOf(array)); // Fails
}

[Test]
public void CantBeUsedWithAList()
{
    var list = new List<int>() { 0, -1, 42, 100 };
    Assert.That(42, Is.AnyOf(list)); // Fails
}
```

It was suggested to make an Analyzer to warn about this, but it seems more prudent to fix it in the framework, so that these expected cases work.

manfred-brands:
Fixes #4355

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Constraint.cs
+++ b/src/NUnitFramework/framework/Constraints/Constraint.cs
@@ -2,6 +2,7 @@
 
 using NUnit.Framework.Internal;
 using System;
+using System.Collections;
 
 namespace NUnit.Framework.Constraints
 {
@@ -139,16 +140,16 @@ namespace NUnit.Framework.Constraints
         }
 
         /// <summary>
-        /// Returns the string representation of this constraint
+        /// Returns the string representation of this constraint and the passed in arguments
         /// </summary>
-        protected virtual string GetStringRepresentation()
+        protected string GetStringRepresentation(IEnumerable arguments)
         {
             System.Text.StringBuilder sb = new System.Text.StringBuilder();
 
             sb.Append("<");
             sb.Append(DisplayName.ToLower());
 
-            foreach (object arg in Arguments)
+            foreach (object arg in arguments)
             {
                 sb.Append(" ");
                 sb.Append(Displayable(arg));
@@ -165,6 +166,12 @@ namespace NUnit.Framework.Constraints
                 else return string.Format(System.Globalization.CultureInfo.InvariantCulture, "{0}", o);
             }
         }
+
+        /// <summary>
+        /// Returns the string representation of this constraint
+        /// </summary>
+        protected virtual string GetStringRepresentation()
+            => GetStringRepresentation(Arguments);
 
         #endregion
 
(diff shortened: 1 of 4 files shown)


SeanKilleen:
That's....better, yeah. 😆

src/NUnitFramework/framework/Constraints/Constraint.cs
```diff
@@ -139,16 +140,16 @@ public override string ToString()
         }
 
         /// <summary>
-        /// Returns the string representation of this constraint
+        /// Returns the string representation of this constraint and the passed in arguments
         /// </summary>
-        protected virtual string GetStringRepresentation()
+        protected string GetStringRepresentation(IEnumerable arguments)
```
stevenaw, line 145: question: Since `Constraint` is public, this overload technically becomes a new "public" API as a custom derived Constraint may try to call this. I usually default to `protected` in this case out of habit, but figured I'd ask: would `private protected` be wanted here instead?

stevenaw (approved):
Nicely done, this looks good to me. Thanks @manfred-brands 
I'd like to leave it open for another team member to take a look too, just in case I'm missing anything. I left a question about member visibility

src/NUnitFramework/framework/Constraints/Constraint.cs
manfred-brands: I had to look up `private protected` because it logically didn't make sense. Since when does `private` mean `internal`?
But that is because `protected internal` means either `derived (anywhere)` or `direct calls from the assembly` instead of `only derived classes inside the assembly`

I think this can be `protected` in case an external `Constraint` has a single `IEnumerable` argument which by default gets formatted as just the type name.
There are at least two existing constraints: `CollectionSubsetConstraint` and `CollectionSupersetConstraint` which could benefit from this call as their current (tested against) string representation is: `<s...setof System.Int32[]>`

mikkelbu (approved):
Also looks good to me, and @manfred-brands reasoning in the PR and the issue seems valid (at least he has given it more thought/examination than I have).

OsirisTerje (approved):
LGTM too.

ivan-gurin:
Hey how about https://github.com/nunit/nunit/blob/097a1be976fb6bfff6a2a3a9d8b24dd5010fbcdb/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs#L870? 
I want Has.All.AnyOf() :)

PR #4360, opened 2023-04-21
Optimize progress reporting by reusing StringBuilder

lahma:
When lookin at memory allocations during Jint.Tests.Test262 run, seems that most of the allocations come from NUnit itself (12%).  Reporting test progress generates a lot of string so there's some room for improvement. Using reused `StringBuilder` to report progress.

The targets are `OuterXml` which always creates a new `StringBuilder` and `TestStarted` which manually concatenates a larger string. Now new `GetOuterXml()` allows providing reusable `StringBuilder`. Introduced new internal `XmlExtensions` that allows writing things in concise manner.

The root cause is more on the [NUnit.Console side](https://github.com/nunit/nunit-console/pull/1338), but this should also help a bit.

image

### DIFF
--- a/src/NUnitFramework/framework/Internal/PropertyBag.cs
+++ b/src/NUnitFramework/framework/Internal/PropertyBag.cs
@@ -144,13 +144,16 @@ namespace NUnit.Framework.Internal
             // enumerating dictionary directly with struct enumerator which is fastest
             foreach (var pair in inner)
             {
-                foreach (var value in pair.Value)
+                // Use for-loop to avoid allocating the enumerator
+                var list = pair.Value;
+                var propertyCount = list.Count;
+                for (var i = 0; i < propertyCount; i++)
                 {
                     TNode prop = properties.AddElement("property");
 
                     // TODO: Format as string
                     prop.AddAttribute("name", pair.Key);
-                    prop.AddAttribute("value", value.ToString()!);
+                    prop.AddAttribute("value", list[i]!.ToString()!);
                 }
             }
 
--- a/src/NUnitFramework/framework/Internal/TestProgressReporter.cs
+++ b/src/NUnitFramework/framework/Internal/TestProgressReporter.cs
@@ -1,7 +1,10 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System;
+using System.IO;
+using System.Text;
 using System.Web.UI;
+using System.Xml;
 using NUnit.Framework.Interfaces;
 
 namespace NUnit.Framework.Internal
@@ -11,11 +14,14 @@ namespace NUnit.Framework.Internal
     /// the async callbacks that are used to inform the client
     /// software about the progress of a test run.
     /// </summary>
-    public class TestProgressReporter : ITestListener
+    public sealed class TestProgressReporter : ITestListener
     {
-        static readonly Logger log = InternalTrace.GetLogger("TestProgressReporter");
+        private static readonly Logger log = InternalTrace.GetLogger("TestProgressReporter");
 
-        private readonly ICallbackEventHandler handler;
+        private readonly ICallbackEventHandler _handler;
+
+        // this reporter is being called synchronously so we can reuse shared string builder
+        private readonly StringBuilder _stringBuilder = new();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="TestProgressReporter"/> class.
@@ -23,7 +29,7 @@ namespace NUnit.Framework.Internal
         /// <param name="handler">The callback handler to be used for reporting progress.</param>
         public TestProgressReporter(ICallbackEventHandler handler)
         {
-            this.handler = handler;
+            _handler = handler;
         }
 
         #region ITestListener Members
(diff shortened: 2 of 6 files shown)


src/NUnitFramework/framework/Internal/TestProgressReporter.cs
```diff
@@ -30,6 +31,9 @@ public TestProgressReporter(ICallbackEventHandler handler)
 
         #region ITestListener Members
 
+        [ThreadStatic]
+        private static StringBuilder? _testStartedStringBuilder;
```
stevenaw, line 35: This feels like a good step forward here, thanks! I don't have a great perfect model of the framework's approach to parallelization but this feels safe since we internally only use it on-stack.

I _think_ this will still create a builder per thread. This change makes me curious if this opens the door to take this further in future PRs by making a formal pooling mechanism with rent/release mechanics so it's only _at most_ a builder per thread in the worst case (ie, when all worker threads happen to be processing a `TestStarted` at the same time).

src/NUnitFramework/framework/Internal/TestProgressReporter.cs
```diff
@@ -39,19 +43,39 @@ public void TestStarted(ITest test)
             var parent = GetParent(test);
             try
             {
-                string report;
+                var report = _testStartedStringBuilder ??= new StringBuilder();
```
stevenaw, line 46: question: Should we set an initial capacity? I think it'll default to `16` but in either path below we likely require more

src/NUnitFramework/framework/Internal/TestProgressReporter.cs
```diff
@@ -39,19 +43,39 @@ public void TestStarted(ITest test)
...
+                report.Clear();
                 if (test is TestSuite)
                 {
+                    report
+                        .Append("<start-suite id=\"").Append(test.Id).Append('"')
```
stevenaw, line 51: comment: There was some duplicate of attributes in both paths here, but this move to StringBuilder also probably opens some opportunities for deduplication of XML attribute writing too in the future. Thanks!

src/NUnitFramework/framework/Internal/TestProgressReporter.cs
lahma: I think the easiest approach might be just taking the source from .NET: https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/Text/StringBuilderCache.cs , it's MIT licensed so should not be a problem and already battle-tested. Here I just tried to avoid bringing too much new stuff in, but I agree that general solution would be better.

src/NUnitFramework/framework/Internal/TestProgressReporter.cs
lahma: With MS version it would be 16, without initial the call after first will have larger StringBuilder with this original code.

src/NUnitFramework/framework/Internal/TestProgressReporter.cs
lahma: Seems that NUnit would require larger than 360 chars default as it's building XML, Jint is using a pool which is not bound to threads which might work better here: https://github.com/sebastienros/jint/blob/main/Jint/Pooling/StringBuilderPool.cs

src/NUnitFramework/framework/Internal/TestProgressReporter.cs
lahma: Fixed in the latest force-push, adding common once before checking the type of requested output

OsirisTerje:
@lahma Some conflicts here now

manfred-brands:
@lahma I merged your other PR so could you resolve the merge conflicts in this one please?

lahma:
@manfred-brands rebased and even got a green build!

OsirisTerje:
@manfred-brands @stevenaw Is this ready to be merged now?  Any comments actually unresolved?

src/NUnitFramework/framework/XmlExtensions.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        /// <summary>
+        /// Checks that CDATA section contains safe content and if not, escapes it.
+        /// </summary>
+        internal static void WriteCDataSafe(this XmlWriter writer, string text)
+        {
+            if (text is null)
+                throw new ArgumentNullException(nameof(text));
+
+            text = EscapeInvalidXmlCharacters(text);
```
stevenaw, line 37: question: I noticed we added the call to `EscapeInvalidXmlCharacters`. Were there any other changes as part of moving this into an extensions file? It's tough to tell for sure

src/NUnitFramework/framework/Internal/TestProgressReporter.cs
```diff
@@ -119,29 +171,28 @@ public void SendMessage(TestMessage message)
...
-        private static string FormatAttributeValue(string original)
+        private StringBuilder GetStringBuilder()
         {
-            return original
-                .Replace("&", "&amp;")
-                .Replace("\"", "&quot;")
-                .Replace("'", "&apos;")
-                .Replace("<", "&lt;")
-                .Replace(">", "&gt;");
+            _stringBuilder.Clear();
```
stevenaw, line 194: 👍

src/NUnitFramework/framework/Internal/TestProgressReporter.cs
```diff
@@ -11,19 +14,22 @@ namespace NUnit.Framework.Internal
     /// the async callbacks that are used to inform the client
     /// software about the progress of a test run.
     /// </summary>
-    public class TestProgressReporter : ITestListener
+    public sealed class TestProgressReporter : ITestListener
     {
-        static readonly Logger log = InternalTrace.GetLogger("TestProgressReporter");
+        private static readonly Logger log = InternalTrace.GetLogger("TestProgressReporter");
```
stevenaw, line 19: Thanks for making the visibility explicit

src/NUnitFramework/framework/Internal/TestProgressReporter.cs
```diff
@@ -11,19 +14,22 @@ namespace NUnit.Framework.Internal
     /// the async callbacks that are used to inform the client
     /// software about the progress of a test run.
     /// </summary>
-    public class TestProgressReporter : ITestListener
+    public sealed class TestProgressReporter : ITestListener
```
stevenaw, line 17: I understand how sealing a class can help performance, but I'm a little cautious about this one. I think `ITestEventListener` was removed as an extension point in the framework but is still an extension point within the engine. It's plausible that someone may've extended `TestProgressReporter` in their codebase to do this.
https://docs.nunit.org/articles/nunit-engine/extensions/creating-extensions/Event-Listeners.html?q=ITestListener

@OsirisTerje you have more experience at this level, but am I reading the docs right? I'm a little confused as we also have this, but I believe this page only indicates the removal as a framework extension point: https://docs.nunit.org/articles/nunit/technical-notes/usage/Addin-Replacement-in-the-Framework.html?q=ITestListener

src/NUnitFramework/framework/Internal/TestProgressReporter.cs
```diff
@@ -34,27 +40,54 @@ public TestProgressReporter(ICallbackEventHandler handler)
...
+                var isSuite = test is TestSuite;
+                writer.WriteStartElement(isSuite ? "start-suite" : "start-test");
+
+                writer.WriteAttributeString("id", test.Id);
+                writer.WriteAttributeString("parentId", parent != null ? parent.Id : string.Empty);
+                writer.WriteAttributeStringSafe("name", test.Name);
+                writer.WriteAttributeStringSafe("fullname", test.FullName);
+                writer.WriteAttributeStringSafe("type", test.TestType);
+
+                if (isSuite)
```
stevenaw, line 73: Thank you 😄 
I hadn't expected my earlier comment about this to make it into the PR but it's a welcome addition

manfred-brands:
@OsirisTerje I did not review as @stevenaw had some comments.

stevenaw (approved):
Thanks @lahma ! I've been revisiting this PR periodically hoping to have had the time to take it all in and process it. I think I've finally been able to give it the attention it deserves.

This looks great to me, thank you once again for the careful profiling. I've left one question requesting clarification from someone more in-the-know about the sealing of `TestProgressListener` and would appreciate a second pair of eyes in general, but overall this looks great ✅

stevenaw:
> @OsirisTerje I did not review as @stevenaw had some comments.

@manfred-brands seems we were both looking at this at the same time 😄 No obligation, but I'd appreciate a second pair of eyes if you're able.

src/NUnitFramework/framework/Internal/PropertyBag.cs
```diff
@@ -144,13 +144,16 @@ public TNode AddToXml(TNode parentNode, bool recursive)
...
+                var list = pair.Value;
+                var propertyCount = list.Count;
+                for (var i = 0; i < propertyCount; i++)
                 {
                     TNode prop = properties.AddElement("property");
 
                     // TODO: Format as string
                     prop.AddAttribute("name", pair.Key);
-                    prop.AddAttribute("value", value.ToString()!);
+                    prop.AddAttribute("value", list[i]!.ToString()!);
```
manfred-brands, line 156:
```suggestion
                    prop.AddAttribute("value", list[i].ToString()!);
```

list elements are not-null. Null suppression operator not needed.

src/NUnitFramework/framework/Internal/TestProgressReporter.cs
manfred-brands: We should enable the StyleCop rule and do it for all fields.

manfred-brands (approved):
Only a tiny extra '!' null suppression operator that isn't needed.
I'll remove it on my next nullable change.

src/NUnitFramework/framework/Internal/TestProgressReporter.cs
lahma: Sorry for being a little late for the game. Would v4 be good time to seal (separate PR) all things that imaginable that doesn't break the current upstream (adapters, runners) and then if someone cries out just open in small point releases, opening sealed classes later isn't a breaking change but sealing too late is...

src/NUnitFramework/framework/XmlExtensions.cs
lahma: No other changes than ensuring the common fast path for "no need to encode anything".

src/NUnitFramework/framework/Internal/TestProgressReporter.cs
lahma: I'm also partly confused about usage of regions (I personally don't like them) and braces (I usually always use them), maybe they could be part of the rule too if either one should not be used.

PR #4362, opened 2023-04-23
Optimize TNode XML reading

lahma:
`XmlDocument` is heavy and by using just `XmlReader` and reading information that is interesting in a forward-only manner, things get faster and less memory-consuming. `TNode` also always created properties and child nodes collections allocating memory even though they might not be needed. `TNode` also allowed something that I call "picking the pockets" as collections allowed direct `.Add` called, now there's separate `InsertChildNode` and `AddChildNode` which also check for backing store initialization. So this PR targets the read part of XML.
 
* change `TNode` to read from `XmlReader`
* change `TNode`'s `_childNodes` and `_attributes` to be lazily initialized only when needed
* create struct-based views for attributes and child nodes which hide nullability of backing stores

# Benchmark results

These are against 60k test suite Jint.Tests.Test262.

``` ini

BenchmarkDotNet=v0.13.5, OS=Windows 11 (10.0.22621.1631/22H2/2022Update/SunValley2)
.NET SDK=7.0.203

```

## NUnit.Framework.FrameworkControllerBenchmark

| **Diff**|Method|Mean|Error|Allocated|
|------- |-------|-------:|-------|-------:|
| Old |CountTestsEmptyFilter|91.78 μs|0.527 μs|0 B|
| **New** |	| **90.52 μs (-1%)** | **0.415 μs** | **0 B** |
| Old |CountTestsWithCategoryFilter|573.45 μs|0.812 μs|12953 B|
| **New** |	| **564.48 μs (-2%)** | **0.842 μs** | **10865 B (-16%)** |
| Old |CountTestsWithOrFilterAll|3,545.07 μs|3.139 μs|3957323 B|
| **New** |	| **1,717.27 μs (-52%)** | **12.752 μs** | **1285461 B (-68%)** |
| Old |LoadTests|60,392.06 μs|141.555 μs|24623708 B|
| **New** |	| **60,387.40 μs (0%)** | **101.571 μs** | **24623267 B (0%)** |

### DIFF
--- a/src/NUnitFramework/framework/Interfaces/TNode.cs
+++ b/src/NUnitFramework/framework/Interfaces/TNode.cs
@@ -1,7 +1,12 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.IO;
+using System.Runtime.CompilerServices;
 using System.Text;
 using System.Xml;
 
@@ -15,8 +20,17 @@ namespace NUnit.Framework.Interfaces
     /// </summary>
     // ReSharper disable once InconsistentNaming
     // Disregarding naming convention for back-compat
-    public class TNode
+    [DebuggerDisplay("{OuterXml}")]
+    public sealed class TNode
     {
+        private static readonly XmlWriterSettings _xmlWriterSettings = new()
+        {
+            ConformanceLevel = ConformanceLevel.Fragment
+        };
+
+        internal List<TNode>? _childNodes;
+        internal Dictionary<string, string>? _attributes;
+
         #region Constructors
 
         /// <summary>
@@ -26,8 +40,6 @@ namespace NUnit.Framework.Interfaces
         public TNode(string name)
         {
             Name = name;
-            Attributes = new AttributeDictionary();
-            ChildNodes = new NodeList();
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


OsirisTerje:
@lahma  Seems you have some conflicts here now.

OsirisTerje (approved):
LGTM, you just need to fix the conflicts

lahma:
Rebased, I believe these two of my PR will probably nicely conflict with each other so if either one gets merged, I'll rebase the other.

src/NUnitFramework/framework/Interfaces/TNode.cs
```diff
@@ -115,14 +124,75 @@ public string OuterXml
...
+            TNode? root = null;
+            TNode? current = null;
+            var parents = new Stack<TNode>();
+
+            while (!reader.EOF)
             {
-                throw new ArgumentException("Cannot parse text as Xml");
+                if (reader.NodeType == XmlNodeType.Element)
+                {
+                    // keep track of previous which will be promoted to parent stack of reader depth changes
```
mikkelbu, line 141: Should this be "if"?
```suggestion
                    // keep track of previous which will be promoted to parent stack if reader depth changes
```

mikkelbu (approved):
LGTM

src/NUnitFramework/framework/Interfaces/TNode.cs
lahma: Force-pushed typo fix to keep single commit PR.

manfred-brands (approved):
Nice work using structs as wrappers for a possible empty list.

PR #4366, opened 2023-04-24
Add NUnit.Analyzer to nunit.framework.tests

### LINKED ISSUE #4364: Add NUnit.Analyzer to our test projects
Mainly as it is needed for #3376 we should add the NUnit.Analyzers to our test projects.

### LINKED ISSUE #4365: TestBuilder.RunTestCase changes CurrentContext
Whilst adding the nunit.analyzer to the framework tests, some `Assert.Multiple` tests were failing.

```csharp
ITestResult result = TestBuilder.RunTestCase(typeof(AssertMultipleFixture), methodName);

Assert.That(result.StackTrace, Is.Not.Null); // Added by me to show failure: Passes

Assert.Multiple(() =>
{
    Assert.That(result.ResultState, Is.EqualTo(expectedResultState), "ResultState");
    Assert.That(result.AssertCount, Is.EqualTo(expectedAsserts), "AssertCount");
    Assert.That(result.AssertionResults, Has.Count.EqualTo(assertionMessageRegex.Length), "Number of AssertionResults");
});

Assert.That(result.StackTrace, Is.Not.Null); // Added by me to show failure: Fails
```

The `Assert.Multiple` is modifiying the local `result` instance and because it succeeds resets the `StackTrace` member.
The error message is not `expect not null`, but the error message from the test case run inside the `RunTestCase`.

The `TestBuilder.RunTestCase` creates a new `TestContext` but does not restore it at exit and as `result` is just a reference to `CurrentContect.Result` it is the same variable.

```csharp
var context = TestContext.CurrentContext;

ITestResult result = TestBuilder.RunTestCase(typeof(AssertMultipleFixture), methodName);

Assert.That(TestContext.CurrentContext, Is.EqualTo(context));
```

That means that the test run and the test verifying the result run in the same context modifying the same `result`.

manfred-brands:
Fixes #4364 
Fixes #4365 

Convert appropriate classic Assert to Constraints except in the area explicitly testing those classic Assertions.

### DIFF
--- a/src/NUnitFramework/TestBuilder.cs
+++ b/src/NUnitFramework/TestBuilder.cs
@@ -165,15 +165,29 @@ namespace NUnit.TestUtilities
 
         public static ITestResult ExecuteWorkItem(WorkItem work)
         {
-            work.Execute();
+            var savedContext = TestExecutionContext.CurrentContext;
 
-            // TODO: Replace with an event - but not while method is static
-            while (work.State != WorkItemState.Complete)
+            try
             {
-                Thread.Sleep(1);
+                return ExecuteUntilComplete(work);
+            }
+            finally
+            {
+                savedContext.EstablishExecutionEnvironment();
             }
 
-            return work.Result;
+            static ITestResult ExecuteUntilComplete(WorkItem work)
+            {
+                work.Execute();
+
+                // TODO: Replace with an event - but not while method is static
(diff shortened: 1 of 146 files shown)


OsirisTerje (approved):
Awesome Job!
Looked through the top 30%, then just skimmed the rest.

PR #4382, opened 2023-05-04
adding better error message for missing generic parameter

### LINKED ISSUE #3215: Better error message for generic test where type parameter cannot be determined

ovebastiansen:
fixes #3215

got full file conflicts if I tried reusing the branch and pushing it on top of master from prev PR, so started over

adding tests for generic methods that cannot resolve type
not the best unittest, but had to use a the runfixture method to get the buildup happening and "crashing" in the correct place

### DIFF
--- a/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
@@ -203,9 +203,9 @@ namespace NUnit.Framework.Internal.Builders
             if (argsProvided > maxArgsNeeded)
                 return MarkAsNotRunnable(testMethod, $"Too many arguments provided, provide at most {maxArgsNeeded} arguments.");
 
-            if (testMethod.Method.IsGenericMethodDefinition && arglist != null)
+            if (testMethod.Method.IsGenericMethodDefinition)
             {
-                if (!new GenericMethodHelper(testMethod.Method.MethodInfo).TryGetTypeArguments(arglist, out var typeArguments))
+                if (arglist == null || !new GenericMethodHelper(testMethod.Method.MethodInfo).TryGetTypeArguments(arglist, out var typeArguments))
                     return MarkAsNotRunnable(testMethod, "Unable to determine type arguments for method");
 
                 testMethod.Method = testMethod.Method.MakeGenericMethod(typeArguments);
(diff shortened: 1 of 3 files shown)


OsirisTerje (approved):
LGTM

Issue #2783, opened 2018-03-27
uint attribute parameter conversion fails with leading zeros


ChrisMaddock:
In the following example:

```
        [TestCase(0x00FFFFFF, TestName = "{m}(Clear)")]
        [TestCase(0xFFFFFFFF, TestName = "{m}(White)")]
        public void UIntTest(uint stopColor)
        {
        }
```

The second test case passes, where as the first failes with the below Exception:

```
System.ArgumentException : Object of type 'System.Int32' cannot be converted to type 'System.UInt32'.
   at System.RuntimeType.TryChangeType(Object value, Binder binder, CultureInfo culture, Boolean needsSpecialCast)
   at System.RuntimeType.CheckValue(Object value, Binder binder, CultureInfo culture, BindingFlags invokeAttr)
   at System.Reflection.MethodBase.CheckArguments(Object[] parameters, Binder binder, BindingFlags invokeAttr, CultureInfo culture, Signature sig)
   at System.Reflection.RuntimeMethodInfo.InvokeArgumentsCheck(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at NUnit.Framework.Internal.Reflect.InvokeMethod(MethodInfo method, Object fixture, Object[] args)
```

This is NUnit 3.10.0

CharliePoole:
You would probably have to use the debugger to confirm, but here's what I think happens...

First Case:
 * Compiler interprets argument as an Int32
 * TestAttribute performs no special conversions for unsigned targets because user is assumed to use either a variable of the correct type or a constant with `U` suffix.
 * An error ensues when we try to call the method.

Second Case:
 * Compiler interprets the argument as an unsigned int (guessing here)
 * TestAttribute does no conversion since arg is the correct type
 * It works correctly

So essentially, NUnit is doing nothing to change the arguments here but is relying on .NET to figure out any automatic conversions when calling the method. That's what happens in the majority of cases, in fact.

Workaround would be to use some syntax that let's compiler know it's dealing with an unsigned. There are several possible options including casting and use of a defined constant.

[...]

PR #4383, opened 2023-05-05
WIP: Try to convert int to uint in testcase parameters

stevenaw:
Fixes https://github.com/nunit/nunit/issues/2783

We noticed that passing an `int` in a `TestCase` into a `uint` would fail. This was originally by design as:

> WRT TestCaseAttribute args, we compromised but we only step in if you have a target Type that cannot be expressed as an attribute argument

However it was noticed that this was inconsistent with how we do equality conversions. Equality conversions do allow for implicit `int` -> `uint` conversions.

> A thought: maybe we should do all the conversions that we already do on numerics (e.g. for equality). That would probably feel more consistent. If so, we should do them where they will work for all data attributes rather than just for TestCaseAttribute.

> Since we are already converting int to uint when comparing for equality, that answers my question since I think we should make all type conversions use the same logic. And we wouldn't want to remove that feature when unifying.


stevenaw:
Hm. The build failed, seeming universally. Yet "it worked on my machine". Converting this to a draft for now

PR #4386, opened 2023-05-06
Add support for CancelAfterAttribute

### LINKED ISSUE #4385: Add support for Test Cancellation
This is to augment the existing TimeoutAttribute
The latter which has problems on .NET Core version of NUnit #4021

manfred-brands:
Fixes #4385 #1821 

This is a cooperative multitasking alternative to TimeoutAttribute

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/CancelAfterCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/CancelAfterCommand.cs
@@ -1,0 +1,78 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using System;
+using System.Threading;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal.Abstractions;
+
+namespace NUnit.Framework.Internal.Commands
+{
+    /// <summary>
+    /// <see cref="CancelAfterCommand"/> creates a timer in order to cancel
+    /// a test if it exceeds a specified time and adjusts
+    /// the test result if it did time out.
+    /// </summary>
+    public class CancelAfterCommand : DelegatingTestCommand
+    {
+        private readonly int _timeout;
+        private readonly IDebugger _debugger;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="TimeoutCommand"/> class.
+        /// </summary>
+        /// <param name="innerCommand">The inner command</param>
+        /// <param name="timeout">Timeout value</param>
+        /// <param name="debugger">An <see cref="IDebugger" /> instance</param>
+        internal CancelAfterCommand(TestCommand innerCommand, int timeout, IDebugger debugger) : base(innerCommand)
+        {
+            _timeout = timeout;
+            _debugger = debugger;
+
+            Guard.ArgumentValid(innerCommand.Test is TestMethod, "CancelAfterCommand may only apply to a TestMethod", nameof(innerCommand));
+            Guard.ArgumentValid(timeout > 0, "Timeout value must be greater than zero", nameof(timeout));
+            Guard.ArgumentNotNull(debugger, nameof(debugger));
+        }
+
+        /// <summary>
+        /// Runs the test, saving a TestResult in the supplied TestExecutionContext.
+        /// </summary>
+        /// <param name="context">The context in which the test should run.</param>
+        /// <returns>A TestResult</returns>
+        public override TestResult Execute(TestExecutionContext context)
+        {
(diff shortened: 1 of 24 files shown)


manfred-brands:
Convert to draft to see why the Unix based build were hanging.

manfred-brands:
This introduces a new `CancelAfter` attribute.
The framework will automatically pass in an `CancellatonToken` argument if the test expects one, alternatively that is available under `TestExecutionContext.CancellationToken`

Note that this is cooperative and only works if test code (or code called from there) passes in and obeys this token.

stevenaw:
Congrats on finding the underlying linux issue @manfred-brands . I'll try to take a look within the week

src/NUnitFramework/framework/Internal/Commands/CancelAfterCommand.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// <summary>
+        /// Runs the test, saving a TestResult in the supplied TestExecutionContext.
+        /// </summary>
+        /// <param name="context">The context in which the test should run.</param>
+        /// <returns>A TestResult</returns>
+        public override TestResult Execute(TestExecutionContext context)
+        {
+            // Because of the debugger possibly attaching after the test method is started
+            // we use a 2-prong approach where we only cancel the test's cancellation token
+            // if the debugger is not attached when the timer expires.
```
stevenaw, line 44: Thanks for adding this. Good thinking to have this caveat about the debugger

src/NUnitFramework/testdata/TimeoutFixture.cs
```diff
@@ -27,7 +27,7 @@ public void TearDown()
         [Test, Timeout(50)]
         public void InfiniteLoopWith50msTimeout()
         {
-            while (true) { }
+            Thread.Sleep(5000);
```
stevenaw, line 30: 👍 thank you for this update too

src/NUnitFramework/tests/Attributes/CancelAfterTests.cs
```diff
@@ -0,0 +1,318 @@
...
+            Assert.That(cancellationToken, Is.EqualTo(TestContext.CurrentContext.CancellationToken));
+        }
+
+        [TestCaseSource(nameof(Arguments))]
+        [CancelAfter(500)]
+        public void TestWithCancelAfterAttributeAndTestCaseSourceHasCancellationToken(int value, CancellationToken cancellationToken)
+        {
+            Assert.That(value, Is.EqualTo(1));
+            Assert.That(cancellationToken, Is.Not.EqualTo(CancellationToken.None));
+            Assert.That(cancellationToken, Is.EqualTo(TestContext.CurrentContext.CancellationToken));
```
stevenaw, line 289: 👍

src/NUnitFramework/testdata/TimeoutFixture.cs
stevenaw: nitpick: Actually, I feel like this may no longer do what the test method name says. Is it worth keeping keeping one busy wait in place to simulate a CPU-bound long running test?

stevenaw (approved):
Thanks @manfred-brands nicely done getting this complete. This looks good to me at first glance, though there is enough here and I am sleepy enough that I would appreciate a second pair of eyes from either myself later or another team member.

@OsirisTerje or @jnm2 Are either of you are up to a review?

src/NUnitFramework/testdata/TimeoutFixture.cs
manfred-brands: I mainly made this change to prevent the builds hanging and taking up unnecessary resources.
Yes, they probably should be renamed.
I don't see a point in making a CPU-bound versions. It should not make a difference in 'interuptability'.

manfred-brands:
ping @nunit/framework-team Any one else for a 2nd review?

OsirisTerje:
@manfred-brands There are a lot of changes/additions here.  I can't really wrap my brain around it.  Is there anything in particular in the code you're worried about?

manfred-brands:
> @manfred-brands There are a lot of changes/additions here. I can't really wrap my brain around it. Is there anything in particular in the code you're worried about?

@OsirisTerje Not really worried, but could you look at the changes to `TestResult.cs` and `SimpleWorkItem.cs` as they deal with existing code.

manfred-brands:
@nunit/framework-team Does anyone has any objections to me merging this?

OsirisTerje:
Suggest you just merge it, and then we check the dev build afterwards.  I could not see anything that raised any worries here.

OsirisTerje (approved):
LGTM

PR #4389, opened 2023-05-13
WIP: Async TestCaseSource

### LINKED ISSUE #2729: Proposal: Async test case sources
For integration tests, it seems almost all of my test case sources would benefit from being allowed to use `await` syntax. I've been converting async code to `.GetAwaiter().GetResult()` every time I move it into a test case source.

These test case sources are following the principle of doing the absolute minimum amount of work needed to discover the correct number and correct display names of test cases before running any tests.
Many of the objects they use to determine the test-discovery data provide no synchronous API.

Since I'm familiar with the async execution parts of NUnit, I'm imagining this would be a few line change (besides the tests) and no maintenance overhead beyond what we're already doing today.

```cs
public static async Task<IEnumerable<TestCaseData>> AllGizmoConfigurations()
{
    var dataController = await base.LoadController();
    return from gizmo in await dataController.LoadGizmos()
           from configuration in gizmo.Configurations.AsIndexed()
           select new TestCaseData(gizmo, configuration.Value)
               .SetName("{m}(" + gizmo.Name + "[" + configuration.Index + "])")
}

[TestCaseSource(nameof(AllGizmoConfigurations))]
public static void Gizmo_configuration_is_valid(Gizmo gizmo, GizmoConfiguration configuration)
{
    // ...
}
```

stevenaw:
Fixes #2729
TODO: ValueSource, TestFixtureSource, make an issue to relax the analyzer


OsirisTerje (approved):
LGTM

stevenaw:
Thanks!

PR #4396, opened 2023-05-26
Removed some using warnings that keeps popping up as errors in vs

ovebastiansen:
Annoying list of warnings/errors that I wanted to just get rid of


stevenaw (approved):
Thanks for catching and fixing this!

PR #4398, opened 2023-05-31
Remove partial trust support

### LINKED ISSUE #3301: [HandleProcessCorruptedStateExceptions] has no effect unless we disable partial trust
@nunit/framework-team This question can stand on its own, even though https://github.com/nunit/nunit/issues/3295 is dependent on the decision we make here.

### Conflict

@ggeurts asked for partial trust support and added it in https://github.com/nunit/nunit/pull/1775 which shipped in NUnit Framework 3.5.

Later, @endrih asked for NUnit to handle exceptions such as AccessViolationException and added it in https://github.com/nunit/nunit/pull/2512 which shipped in NUnit Framework 3.9. However, we all missed the fact that it has no effect under partial trust:

> The HandleProcessCorruptedStateExceptionsAttribute attribute is ignored when it is encountered in partially trusted or transparent code, because a trusted host should not allow an untrusted add-in to catch and ignore these serious exceptions.

(https://docs.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.handleprocesscorruptedstateexceptionsattribute#remarks)

### Looking ahead

The .NET team decided not to implement partial trust on .NET Core. The partial trust attributes are no-ops on .NET Core. (See e.g. `https://github.com/dotnet/corefx/issues/12592`.)
.NET 5 is where the .NET community is headed and it will be based on .NET Core. (https://devblogs.microsoft.com/dotnet/introducing-net-5/) Therefore, the future of .NET does not have a place for partial trust. In contrast, exceptions such as AccessViolationException will likely always be something we'll want to catch.

In addition, CPU exploits such as Spectre have made it clear that security boundaries cannot be more granular than shared memory address space. Effectively that means that less secure code must be run in a separate process, period. .NET Core embraced this perspective from the beginning when the team decided that AppDomains were too fraught with issues to carry them forward from .NET Framework.

It's worth pointing out that if we merge https://github.com/nunit/nunit-console/pull/660, we won't need `[HandleProcessCorruptedStateExceptions]` when running under console 3.11 and newer since the config element opts in globally. We'll still need `[HandleProcessCorruptedStateExceptions]` when running under any other test host process with its own config, such as VSTest, ReSharper's test host, NCrunch, etc.

### Action

Should we keep the partial trust attributes? If so, how long? E.g. until v4 of NUnit?

If we keep support for partial trust, I'd like to put comments on each current and future occurrence of `[HandleProcessCorruptedStateExceptions]` noting that on .NET Framework it will only come into effect if we stop supporting partial trust.

If we don't keep support and @ggeurts or others still need it, they will be unable to upgrade to newer NUnit versions.

stevenaw:
Fixes #3301 

This largely was a find-remove of the following attributes usage and comments without much else needed. 
- `SecurityCriticalAttribute`
- `SecuritySafeCriticalAttribute`
- `AllowPartiallyTrustedCallers`

The `LowTrustFixture` tests were updated to now expect a `MemberAccessException` to be thrown as a way of verifying we continue to not support partial trust, though I'm on the fence about removing them outright.

### DIFF
--- a/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
+++ b/src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Reflection;
-using System.Security;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
 using NUnit.Framework.Internal.Builders;
@@ -209,11 +208,6 @@ namespace NUnit.Framework.Api
             return result;
         }
 
-        // This method invokes members on the 'System.Diagnostics.Process' class and must satisfy the link demand of
-        // the full-trust 'PermissionSetAttribute' on this class. Callers of this method have no influence on how the
-        // Process class is used, so we can safely satisfy the link demand with a 'SecuritySafeCriticalAttribute' rather
-        // than a 'SecurityCriticalAttribute' and allow use by security transparent callers.
-        [SecuritySafeCritical]
         private TestSuite BuildTestAssembly(Assembly assembly, string assemblyNameOrPath, IList<Test> fixtures)
         {
             TestSuite testAssembly = new TestAssembly(assembly, assemblyNameOrPath);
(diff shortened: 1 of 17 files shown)


OsirisTerje (approved):
LGTM

stevenaw:
Thanks!

Issue #4415, opened 2023-07-10
Remove use of params for messages


OsirisTerje:
Ref issue #4413  and Issue #3936 - introduction of  CallerArgumentExpressions (CAE).

Since we need to have the CAE at the end of the parameter list, we can't have the params there.  The use of params for messages are now superseded by the use of string interpolation.

The analyzer should warn about this usage and suggest conversion to an interpolated string, see https://github.com/nunit/nunit.analyzers/issues/562

Issue #4413, opened 2023-06-30
Assert.That methods should autogenerate message, if null


FrankDzaebel:
if we use something like:  `Assert.That(actualBool, Is.EqualTo(expectedBool));`  
when they do not match we get the output:
``` c#
  Expected: True
  But was:  False
```
With the new [CallerArgumentExpressionAttribute](https://learn.microsoft.com/de-de/dotnet/api/system.runtime.compilerservices.callerargumentexpressionattribute?view=net-7.0) it would be easy to get:
``` c#
  actualBool
  Expected: True
  But was:  False
```
This would be easier to understand in unit tests.   ... and extremly easy to implement.

FrankDzaebel:
"_easy to implement_" for example: 
``` c#
    public static void That<TActual>(
        TActual actual,
        IResolveConstraint expression,
        string? message,
        [CallerArgumentExpression(nameof(actual))] string? actualExpression = null,
        params object?[]? args)
    {
        message ??= actualExpression;
        IConstraint constraint = expression.Resolve();
        Assert.IncrementAssertCount();
        ConstraintResult result = constraint.ApplyTo<TActual>(actual);
        if (result.IsSuccess)
            return;
        Assert.ReportFailure(result, message, args);
    }
```

manfred-brands:
That way you still need to supply a message. We could put the attribute on the `message` parameter instead.
See discussions at #3936

OsirisTerje:
I just had to play around with this, and it actually looks pretty good!
Just to play around I made a new class Verify, with the That method, a bit modified, as I wanted both the actual value and the constraint to be displayed, otherwise it looked like it was missing something.  I also used a bit more complex constraints in order to check out if this actually worked or not.

Test code:
```cs
[Test]
    public void SingleTest()
    {
        int nunitAge = 23;
        Verify.That(nunitAge, Is.GreaterThan(42).And.LessThan(85), null);
    }
```
I got results like:
```txt
  Message: 
  Assert.That(nunitAge, Is.GreaterThan(42).And.LessThan(85))
  Expected: greater than 42 and less than 85
  But was:  23
```
NUnit constraints doers a great job of making a meaningful message, but it is nice to get the complete expression out too.
And I really like that it actually gets the complete constraint out.  

If I change the test to use double values, then it makes even more sense:
```cs
[Test]
    public void SingleTest2()
    {
        double nunitAge = 23.4;
[...]

Issue #3936, opened 2021-09-19
Is there any way we could make use of CallerArgumentExpressionAttribute?


Dreamescaper:
C#10 has a nice addition of [CallerArgumentExpressionAttribute](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/caller-argument-expression.md), which we could use to print `actual` argument expression in error messages.

I can see two problems though:
- We need to add optional `string` parameter to methods. But we already have overloads with `message` string parameter.
- We have methods with `params` parameter. And we can't combine `params` and optional parameters in the same method.

Is there any way we could make use of this attribute?

manfred-brands:
We cannot do this with breaking binary compatibility, but the attribute is only effective at compile time anyhow.
For source code we have currently two methods for each test, e.g.:
```
Assert.True(bool condition)
Assert.True(bool condition, string? message, param object?[]? args)
```
We can replace the first with:
```
Assert.True(bool condition, [CallerArgumentExpression("condition")] string? message = null)
```

It will not be confused with the 2nd. Similar like `Console.WriteLine` has different overloads.

Dreamescaper:
Problem is that existing code like that:
```csharp
Assert.True(myCondition, "MyCondition should certainly be true!");
```
will pick up the first method now.
> Assert.True(bool condition, [CallerArgumentExpression("condition")] string? conditionExpression = null)

manfred-brands:
Which is correcy. The CallerArgumentExpression only evaluates to the condition string when not specified:

```
Assert.True(args.Length == 0);
Assert.True(args.Length == 0, "args");
Assert.True(args.Length == 0, "{0} != 0", args.Length);
```

Gives:
```
Expected 'true' for 'args.Length == 0'
Expected 'true' for 'args'
Expected 'true' for '1 != 0'
```

Dreamescaper:
Problem is that existing method usages rarely provide `actual` expression, but usually some failure description.
So for my example
```csharp
Assert.True(myCondition, "MyCondition should be true when the sky is blue!");
```

it will print 
> Expected 'true' for 'MyCondition should be true when the sky is blue!'

which is probably not what's expected.

manfred-brands:
Yes, this only helps where there was no message given previously.

manfred-brands:
This one helps with your situation:
```
public static void True(bool condition, string? userMessage = null, [CallerArgumentExpression("condition")] string? conditionMessage = null)
```

[...]

PR #4419, opened 2023-07-15
Issue4415

OsirisTerje:
Removing params from framework, and handle params in classic.

See #4415 

Note:  This PR builds on top of the PR #4417 .  See commits after #ccd58ab

### DIFF
--- a/.runsettings
+++ b/.runsettings
@@ -1,0 +1,123 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- For more information, see https://docs.microsoft.com/en-us/visualstudio/test/configure-unit-tests-by-using-a-dot-runsettings-file?view=vs-2019  -->
+
+<RunSettings>
+  <RunConfiguration>
+    <!-- See https://github.com/microsoft/vstest-docs/blob/master/docs/configure.md -->
+    <!-- 1. Test related settings -->
+    <!-- [x86] | x64: architecture of test host -->
+    <!--<TargetPlatform>x86</TargetPlatform>-->
+
+    <!-- Framework35 | [Framework40] | Framework45 -->
+    <!--<TargetFrameworkVersion>Framework40</TargetFrameworkVersion>-->
+
+    <!-- Path to Test Adapters -->
+    <!--<TestAdaptersPaths>%SystemDrive%\Temp\foo;%SystemDrive%\Temp\bar</TestAdaptersPaths>-->
+    <TreatTestAdapterErrorsAsWarnings>false</TreatTestAdapterErrorsAsWarnings>
+
+    <!-- Path relative to solution directory -->
+    <ResultsDirectory>.\TestResults</ResultsDirectory>
+
+    <!-- 0 = As many processes as possible, limited by number of cores on machine, 1 = Sequential (1 process), 2-> Given number of processes up to limit by number of cores on machine-->
+    <MaxCpuCount>0</MaxCpuCount>
+
+    <!-- TestSessionTimeout was introduced in Visual Studio 2017 version 15.5 -->
+    <!-- Specify timeout in milliseconds. A valid value should be >= 0. If 0, timeout will be infinity-->
+    <TestSessionTimeout>100000</TestSessionTimeout>
+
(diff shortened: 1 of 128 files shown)


manfred-brands:
@OsirisTerje It would have been nicer if you had made a PR to merge into Issue4416 iso master.

Observations:
* `CollectionAssert` should derive from `AssertBase` and not `Framework.Assert`.
* In various overrides you removed the `null` parameter for the `params args`. This cause the compiler to pass an empty array as `args` and therefore missing the optimization in `ConvertMessageWithArgs` to not call `Format`. This again can cause an exception if the _message_ parameter contains a `{...}` value that not has a matching argument or is not even a number. This can be partly fixed by changing the condition in `ConvertMessageWithArgs` into `(args is null || args.Length == 0)`. But passing `null` is more efficient. But see next point.
* In the classic versions with no argument you pass in `""`.  This can be removed as there is a `That` without arguments.
* In some classic versions that call other classic versions you called `ConvertMessageWithArgs` prematurely resulting in it being called twice.
* In framework versions you now pass in `""` iso `null`. As [SA1122](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1122.md) says this means the compiler creates lots of embedded empty strings.
* You forget to remove `params` from `Assert.ThatAsync`.
* You probably wanted that analyzer fix to convert format specification strings to formattablestrings.

I think I have addressed all of the above in my review commit.

One remark is regarding optimization. Assuming NUnit test pass in 99% of the cases, the string will be formatted unnecessary. And that operation is quite expensive. This is not only the case for the classic versions with `param args` but also for `FormattableString` which by default is formatted before passing to the `Assert.That`.
The first we can fix by using the overload with `Func<string?> getExceptionMessage` parameter. Addressed in 2nd commit.

The latter can be fixed by adding an overload accepting `DefaultInterpolatedStringHandler ` but this requires .NET6.0
I think it is a worthwhile addition, but means we have to add a NET6.0 build (and possibly drop netstandard2.0)
I'm happy to add that if you agree.

OsirisTerje:
@manfred-brands Awesome - thanks a lot for the comments !!  
I'll change it to merge to 4416.

OsirisTerje:
@manfred-brands 

> The latter can be fixed by adding an overload accepting DefaultInterpolatedStringHandler  but this requires .NET6.0
I think it is a worthwhile addition, but means we have to add a NET6.0 build (and possibly drop netstandard2.0)
I'm happy to add that if you agree.

What if we add the Net6 build, but for now keep the netstandard 2.0 and add a compiler directive for the code for netstd 2. ?  That way we can keep the netcore3.1 in play a little bit longer.  (Would have loved some metrics on the framework usage though...)

OsirisTerje:
@manfred-brands 
I just did the same (removed params) for the Assume and Warn classes, and removed and fixed corresponding tests.  

If you have time, (after the DefaultInterpolatedStringHandler  )   you could start up with the CallerArgumentExpression. 
We have the issues #4413 and older #3936.  I  tested out the CAE in the experimental code there.

We should consider  the sequence of the messages, we have 3 types,  user added message,  result message and then CAE message.  I believe the sequence could be like that too, or possible CAE message as 2nd.

stevenaw:
Still on vacation but I've been following from my phone.

I like the NET6 target as it would also allow for many more internal optimizations by allowing us to use more newer span-based APIs when possible, among other newer performance-based APIs. In there past one noted concern was keeping on top of which newer runtimes to target as the framework has historically had a slower release cadence than the newer .NET LTS schedule but I think we've agreed in another issue that the quicker framework release cadence we're aiming for can help here 🙂

manfred-brands:
I started on the .NET6 target. Have to work through some Obsolete and non-available APIs.
That is the problem with NetStandard. It compiles fine, but results in `PlatformNotSupported` exceptions at runtime. We do catch those, but for later targets I have to make the calls conditional.
`dotnet test` works now, but `nunitlist-runner` cannot load the dll. Any ideas?
Will keep you posted.

@stevenaw Once we have the build in place we can see what optimizations we can apply.

OsirisTerje:
Merging this since #4420 handles net6 target

PR #4420, opened 2023-07-17
Add support for native .NET-6.0 target

manfred-brands:
Adds both `net6.0` and `net6.0-windows` support.

Some remarks:
* Converted multiple conditional sections of code to get OSDescription into one property in `OSPlatform`
* Removed `HandleProcessCorruptedStateExceptions` attribute referencing https://github.com/dotnet/coreclr/issues/19698 as that issue has been closed for two years (.net framework 4.8 security update)
* Added some conditional code on `OperatingSystem.IsWindows()` to allow `Registry` and `ApartmentState` calls. These are only supported on `-windows` targets.


manfred-brands:
@OsirisTerje I don't know what is wrong with the appveyor build complaining about formatting.
I updated the `.github/workflows/continuous_integration.yml` to run on all PRs.

OsirisTerje:
@manfred-brands I've looked at it, but can't see what makes it complain on the warnings.  But, the appveyor build have been troublesome, and we discussed keeping to just the github builds.  So I dont think we should waste too much time on the appveyor.

OsirisTerje:
@manfred-brands Do we really need both netX plain and netX-windows targets?

OsirisTerje:
> @OsirisTerje I don't know what is wrong with the appveyor build complaining about formatting.
I updated the .github/workflows/continuous_integration.yml to run on all PRs.

It seems we might have some end-of-line issues.  

If we run :

`git add --renormalize .`

a bunch of files are corrected.  We do have the .gitattributes set to auto for cs files, so we're good there, but some files have probably been altered at some time. 

However, checking the list of normalized files with the appveyor error list doesnt match.  But the coordinates for the errors matches end-of-lines.......

manfred-brands:
> Actually, I think I need this explained. I can't really see why we need the specific windows target. The code exist in the std net6, and there is no conditional switches on this. What am I missing here?

I thought that the code only existed in the `net6.0-windows` target, but after investigation, I think you are right.
The attributes are there to warn the developer that this will only work on Windows.
So indeed we might not need this `-windows` target at all as long as we deal with the expected exception (or call it conditionally)
I'll update the code.

OsirisTerje:
Awesome!    Dealing with the exception sounds great!

OsirisTerje:
@manfred-brands Based on the #4424 discussion, I think we safely can say that we drop Netstandard 2.0 and Netcore 3.1, and set the baseline now at net.6.   When a couple more have had their say, I'll write up that policy and get it in to the docs :-)

manfred-brands:
Replaced by #4428

PR #4422, opened 2023-07-18
WIP: dotnet format whitespace

OsirisTerje:


manfred-brands (approved):
Unfortunately doesn't seem to fix the appveyor build.

OsirisTerje:
No it doesn't do it fully, only fixed around half of the errors. 
Believe the rest is pure line ending issues, but not sure how to fix those in a batch.  The gitattributes should take care of it, but it doesnt.  And I suspect there is something with the editorconfig, but currently it escapes me.  And it works on the github linux build, so not quite sure what is going on here.

PR #4427, opened 2023-07-21
WIP:  Caller argument expression

OsirisTerje:
Issue #4413 and #3936

PR #4428, opened 2023-07-22
WIP: Issue4421 net6

### LINKED ISSUE #4421: Add support for native .NET-6.0 target
To track the work in #4420

manfred-brands:
Fixes #4421 

This is a net6 only solution for conditional formatting of messages.

I found a way to also do this for net462.
Will push that later.
This is parked for comparison

### DIFF
--- a/src/NUnitFramework/framework/Assert.That.cs
+++ b/src/NUnitFramework/framework/Assert.That.cs
@@ -8,6 +8,11 @@ using NUnit.Framework.Internal;
 // Disabled because of the CallerArgumentExpression attributes which are only for the compiler.
 #pragma warning disable CS1573 // Parameter has no matching param tag in the XML comment (but other parameters do)
 
+#if NET6_0_OR_GREATER
+// Disabled because of InterpolatedStringHandlerArgument attributes.
+#pragma warning disable IDE0060 // Remove unused parameter
+#endif
+
 namespace NUnit.Framework
 {
     /// <summary>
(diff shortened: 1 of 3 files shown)


OsirisTerje:
Awesome!  

About Assert and Assert.That files:  Be careful with too many changes there, as I am changing a lot there for the CallerArgumentExpression in #4427 
These changes also affects a lot of the tests, as many of them check on the assert messages.

manfred-brands:
> About Assert and Assert.That files: Be careful with too many changes there, as I am changing a lot there for the CallerArgumentExpression in #4427
> These changes also affects a lot of the tests, as many of them check on the assert messages.

I saw you creating that and it will certainly conflict.  I thought it was on my list after this. You were too keen.
Especially with the conditional formatting it needs work.

manfred-brands:
The code example from #4429:
```csharp
Assert.That(1 + 1, Is.EqualTo(2), $"{text} {formatCounter}");
```

Gets for .net6.0+ compiled into:

```csharp
AssertingInterpolatedStringHandler<int> assertingHandler = new AssertingInterpolatedStringHandler<int>(1, 2, num, resolveConstraint, out isEnabled);
if (isEnabled)
{
	assertingHandler.AppendFormatted("String was formatted");
	assertingHandler.AppendLiteral(" ");
	assertingHandler.AppendFormatted(formatCounter);
}
Assert.That(actual, expression, assertingHandler);
```

This is normally done to prevent the boxing of `struct` objects, but as these are only evaluated on failure, performance is not an issue and I think this solution is too complex.
But it was a nice exercise.

manfred-brands:
This solution is too complex, it requires the `Assert` to be done inside the `InterpolatedStringHandler` constructor only so the `isEnabled` out parameter can be set.
Limited benchmarking only show a slight improvement on the `FormattableString` solution:
```
|                       Method |      Mean | Ratio | Allocated | Alloc Ratio |
|----------------------------- |----------:|------:|----------:|------------:|
|        UseFormatAndArguments |  3.981 ns |  1.00 |      56 B |        1.00 |
|       UsePreformattedMessage | 27.599 ns |  6.95 |     104 B |        1.86 |
|        UseFormattableMessage |  9.216 ns |  2.32 |      88 B |        1.57 |
| UseInterpolatedStringHandler |  7.813 ns |  1.96 |         - |        0.00 |
```

The weirdness of doing the assertion in a string handler constructor and the fact that it doubles the code base with conditional compiles make it not worthwhile to consider.

manfred-brands:
Not worthwhile.

PR #4429, opened 2023-07-22
Issue4421 all target frameworks

### LINKED ISSUE #4421: Add support for native .NET-6.0 target

manfred-brands:
Also fixed #4421 

This version defers formatting interpolatable string to until the test fails for all frameworks.
It adds an overload to all `Assert`, `Assume`, `Warn` methods with a `FormattableString` parameter.
Code like:
```csharp
Assert.That(1 + 1, Is.EqualTo(2), $"{text} {formatCounter}");
```
Gets compiled into:
```csharp
Assert.That(2, Is.EqualTo(2), FormattableStringFactory.Create("{0} {1}", "String was formatted", formatCounter));
```

This works even for .NET 4.6.2

For a  .NET6 specific solution see (#4428). I'm not sure if that is worth it.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -7,10 +7,6 @@ on:
       - release
       - 'v3.13-dev'
   pull_request:
-    branches:
-      - master
-      - release
-      - 'v3.13-dev'
 
 jobs:
   build-windows:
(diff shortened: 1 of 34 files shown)


manfred-brands:
@OsirisTerje This also ensures that classic.test are tested as part of the "build.cake" script.

OsirisTerje:
I assume this one can be closed now?
Outdated by #4430

manfred-brands:
Superceded by #4430

PR #4435, opened 2023-07-26
All Assert messages must be non-null

### LINKED ISSUE #4432: Fix null message in internal static string? ExtendedMessage(string methodName, string? message, string actualExpression, string constraintExpression)
Ref comment in PR #4430 

> The message comes from the _getExceptionMessageoverloads which are declared asFunc<string?>. Also there are some NUnit tests (e.g. LifeCycleAttributeTest) that pass in result.Messagewhich is also declaredstring?`.
If we want to change all that, I would do that in a separate PR.

manfred-brands:
Fixes #4432 

We need to use `NUnitString = default` for the cases where the user does not specify an error.
This will call the `struct` hidden default constructor that initializes all fields with null.
Although the user cannot specify a null string (as there is no converter for it), the compiler can.

### DIFF
--- a/src/NUnitFramework/framework/NUnitString.cs
+++ b/src/NUnitFramework/framework/NUnitString.cs
@@ -14,13 +14,14 @@ namespace NUnit.Framework
     /// </remarks>
     public readonly struct NUnitString
     {
+        // Keep as nullable, as default(NUnitString) will set this to null.
         private readonly string? _message;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="NUnitString"/> class.
         /// </summary>
         /// <param name="message">The message formattable to hold.</param>
-        public NUnitString(string? message)
+        public NUnitString(string message)
         {
             _message = message;
         }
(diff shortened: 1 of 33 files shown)


manfred-brands:
@OsirisTerje Is there something wrong with the myget feed?
Do we even need it inside the repo?

OsirisTerje:
@manfred-brands Seems like myget.org is down.
I believe we only need it for the appveyor, but not sure, never actually looked into it.

OsirisTerje:
I don't really see myget.org to be needed in the project, suggest you just comment it out.

src/NUnitFramework/framework/NUnitString.cs
```diff
@@ -43,6 +44,6 @@ public NUnitString(string? message)
         public static implicit operator NUnitString(string? message) => new(message);
```
OsirisTerje, line 44: This is the one that causes the build errors.  
Do we need the nullable string in here?  (Was a comment on some other PR I think, and perhaps we need to tackle the Result class too? )

OsirisTerje (approved):

src/NUnitFramework/framework/NUnitString.cs
OsirisTerje: Looked very quickly at it, seems like it can be non-null since you have fixed the nullability in the Result class, if I am not missing something.

manfred-brands: Somehow I missed this one. Fixed.

PR #4436, opened 2023-07-26
Added PropertiesComparer

### LINKED ISSUE #4394: Expand NUnitEquality to use Equality of all Properties
We have several class that we don't want to implement `IEquatable<>` for the sake of NUnit tests.

To compare these we have created a `ReflectionAssert.PropertiesEqual` class.
This class gets all the properties of a class and calls the method recursively on those.
If any properties implements `IEquatable<>` it will call the `Equals` method instead.
It would be a nice feature to add for NUnit4 so that we can compare more classes.

manfred-brands:
Fixes #4394 

We constantly have to make special Validate methods or implement IEquatable on classes that should not be comparable.
However we want to be able to compare them in NUnit, e.g. to check if serialization roundtrips.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
-    <add key="NUnit MyGet Feed" value="https://www.myget.org/F/nunit/api/v3/index.json" />
+    <!-- <add key="NUnit MyGet Feed" value="https://www.myget.org/F/nunit/api/v3/index.json" /> -->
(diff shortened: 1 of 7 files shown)


OsirisTerje (approved):
Looks good.
Could perhaps had an explicit test for the PropertiesComparer, but I do see that it is perhaps sufficiently covered by many of other tests.

manfred-brands:
> Could perhaps had an explicit test for the PropertiesComparer

None of the other Comparers have their own tests ...
I did add a specific new test for a class that doesn't override `Equals`, which without this comparer fails and now passes.

Issue #4434, opened 2023-07-25
Fixing the classic asserts


OsirisTerje:
The classic asserts are currently a bit hard to work with, since the tests need some attributes from the Framework namespace, and the Assert keyword is in both the Framework and the Framework.Classic namespace, One need to add some extra using (or qualify the attributes for each usage),  line 2 and 3 below.

See
<img width="307" alt="image" src="https://github.com/nunit/nunit/assets/203432/72a08c9e-56f4-4d77-bd55-ea23d193e088">

Thoughts @manfred-brands @stevenaw ?

manfred-brands:
A few potential options:

* Rename the classic `Assert` class into `ClassicAssert`
* Add all attributes (as derived classes) also to the Classic namespace.

My preference would be the first.

OsirisTerje:
@manfred-brands 

> Rename the classic Assert class into ClassicAssert

Since the namespace is `NUnit.Framework.Classic`, the FQN would be rather verbose with `Classic` duplicated.  Is this ok?  

I would like to have it in its own namespace, but if we change the name itself, we could place it in the NUnit.Framework namespace.  It does make me shudder, but it might be easier for the user.

manfred-brands:
You could use `NUnit.Framework.Deprecated` as the namespace.

With the same namespace users have to make less changes if they really want to keep the classic asserts.
But do we want to make it easy to keep the old or easy to convert to the new?

OsirisTerje:
It is a balance, I think.  We should not scare people away from using version 4, but since it do have breaking changes they should expect some rewrites to be needed. 

It should be easy to convert to the new, and that's where the analyzer comes in.  It do make it very easy, although now we have to make it recognize the new classic ones. 

Deprecated sounds a bit rough, but it also makes for moving more over to the same.  Perhaps we could name it `NUnit.Framework.Legacy` ?

manfred-brands:
Deprecated still sounds better than Obsolete ...

Legacy sounds fine by me.

manfred-brands:
Before changing it, would a single alias like below work?
`using Assert = NUnit.Framework.Classic.Assert`

That way they can keep the existing `using NUnit.Framework` for the `Test` etc. attributes.

OsirisTerje:
Good point -  I think we can do both.   We change the names as in the PR, but we add this possibility in the documentation (even in the readme file for the package when we get that done). 

```cs
using Assert = NUnit.Framework.Legacy.ClassicAssert
```

[...]

PR #4438, opened 2023-07-26
Changed project to be named Legacy and Assert to be called ClassicAssert

OsirisTerje:
Ref issue   #4434


manfred-brands:
Fixed a few namespaces.

manfred-brands (approved):
If we go ahead with this, then the changes are fine.
See my last comment in #4434

PR #4442, opened 2023-07-28
Derive all runtimes in cake script from Directory.Build.props

stevenaw:
As we have been talking about targeting different runtimes for tests or the framework distributable, I recall there having been an idea mentioned in a previous PR to simplify or outright infer some of the runtimes we target. I thought I would make an attempt at it to see thoughts.

The cake script currently loops over multiple netcore runtimes for tests but assumes there will only be one netfx runtime test target. I've left that assumption in place but it would be easy enough to change it to a set-based approach there too.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -24,24 +24,6 @@ var dbgSuffix = configuration == "Debug" ? "-dbg" : "";
 var packageVersion = version + modifier + dbgSuffix;
 
 //////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


stevenaw:
Interesting. The myget part of the appveyor build completed successfully here

OsirisTerje:
Seems they've got myget up again, but we should get it up on github packages instead.

OsirisTerje (approved):
lgtm

Issue #53, opened 2013-11-03
Generic Constraint Model


CharliePoole:
For NUnit v3, a fully generic constraint system (i.e., the Constraint and ConstraintExpression classes having a generic type parameter T) would be desirable in order to be able to leverage types known at compile time when writing constraints. Constraint methods such as Equal and GreaterTo could use that type parameter in their signature, so invalid comparisons would be detected at compile time. Constraints such as Property could provide support for naming symbols using expressions (Expression&lt;Func&lt;T, TR&gt;&gt;). For collection constraints, a TElement generic parameter knowing the element type of the collection should also be added.

The classic constraint builders Is and Has would instantiate the T (and TElement) with System.Object, so that "Assert.That (myValue, Is.EqualTo (10))" would cause an EqualConstraint&lt;object&gt; to be passed to Assert.That and "Assert.That (myValue, Has.Member (10))" would cause a CollectionContainsConstraint&lt;object&gt; (derived from Constraint&lt;T&gt; with T being IEnumerable&lt;object&gt; in this case) to be passed to Assert.That.

Is.TypeOf&lt;T&gt;, Is.AssignableTo&lt;T&gt;, and Is.InstanceOf&lt;T&gt; would return their respective constraints parameterized with the given T, so that "Assert.That (myValue, Is.TypeOf&lt;MyClass&gt;().With.Property (v => v.P).EqualTo (42))" would work.

For full type inference support, a new Assert.That syntax could be provided that binds the T to the actual type of the value passed to the Assert method:
public static void That&lt;T&gt; (T actualValue, Func&lt;ResolveConstraintBuilder&lt;T&gt;, IResolveConstraint&lt;T&gt;&gt; constraintDelegate)
and
public static void That&lt;TElement&gt; (IEnumerable&lt;TElement&gt; actualValue, Func&lt;ResolveCollectionConstraintBuilder&lt;TElement&gt;, IResolveConstraint&lt;IEnumerable&lt;TElement&gt;&gt;&gt; constraintDelegate).

This would allow code such as the following to be written in a typesafe way without needing to repeat the type:
Assert.That (myVariable, v => v.EqualTo (42));
Assert.That (myCollection, c => c.EquivalentTo (new[] { 1, 2, 3 }));
Assert.That (myCollection, c => c.Member (3));

From https://bugs.launchpad.net/nunit-3.0/+bug/901212

stevenaw:
[...]

Issue #3985, opened 2021-11-11
Cache comparer when comparing two IEnumerable<T>


stevenaw:
The arrays and enumerables comparer currently will iterate over and evaluate each item in an array independently and determine the comparer for each. I imagine most usages of this API will involve comparing two arrays of identical types, and likely using a generic collection such as `Array<T>` or `List<T>`. We should consider detecting if both `expected` and `actual` implement `IEnumerable<T>` and doing a one-time determination of the comparer to use over the collection based on the type of `T`.

We could consider taking this one step further and using `SequenceEquals()` if both arrays are of a primitive built-in type.

louis-z:
Although the issue does not have the `help wanted` label, I'd be willing to take a stab at it, if that's alright.

I was considering first tackling arrays, perhaps by modifying `ArraysComparer.Equal` so that - under certain conditions (e.g. no tolerance) - it uses `IStructuralEquatable.Equals`. Any thoughts on that approach?

stevenaw:
Hi @louis-z sorry it's taken a bit to get back to you on your question. You're most welcome to pick this up, but first we may want to confirm a bit of the approach. There's probably a few ways to go about tackling this and I'm not sure which would be best/easiest. 

My first thoughts are that since the ArrayComparer delegates to [EnumerablesComparer](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Constraints/Comparers/EnumerablesComparer.cs), the most general place to tackle this may be inside of there so that all collections would benefit from this. We may also want to find some way to reuse nunit's internal comparers so that any other customizations, like `IgnoreCase`, could also be considered by this optimization.

Within there we may want to then:
1. Examine if x and y implement IEnumerable<T> and if T is the same for both
2. Add a function to NUnitEqualityComparer to determine + return the internal `IChainComparer` to use for `T`
3. Use the comparer returned from step 2 when enumerating x and y instead of the class-level `_equalityComparer`

Alternately, another option could be to keep the changes contained wholly within the `NUnitEqualityComparer` and have it store a cache of resolved `IChainComparer`s for all types in a lookup table like `Dictionary<Type, IChainComparer>`, though since that's public I'd be a little concerned about unexpected side effects if that comparer is somehow used across multiple assertions (each with their own potential customizations to comparison logic).

@nunit/framework-team what are your thoughts? What could this be the best and simplest way to approach this?

PR #4456, opened 2023-08-10
Generic EqualConstraint

stevenaw:
Moved the work from @manfred-brands into a shared branch in the nunit repo for collaboration
Contributes to https://github.com/nunit/nunit/issues/53


stevenaw:
There were two failures at the time I had to stop working on this yesterday, shown below.

The second error, where two ValueTuples which should be equal but aren't reporting as such, was failing when trying to use the default `Equals()` function here: https://github.com/nunit/nunit/blob/149d5b45a01486f5379526b9c1200304ac6fd7e5/src/NUnitFramework/framework/Constraints/EqualConstraint%7BT%7D.cs#L296

The first error, it appears an AssertionException _should_ be thrown but isn't as it's going into the fallback comparer in that case.

```
Errors, Failures and Warnings

1) Failed : NUnit.Framework.Tests.Constraints.EqualConstraintTests.SameValueDifferentTypeExactMessageMatch
  Assert.That(ex?.Message, Does.Contain("  Expected: 0 (Int32)" + NL + "  But was:  0 (IntPtr)" + NL))
  Expected: String containing "  Expected: 0 (Int32)
  But was:  0 (IntPtr)
"
  But was:  "  Assert.That(new IntPtr(0), Is.EqualTo(0))
  Expected: 0
  But was:  0
"
   at NUnit.Framework.Tests.Constraints.EqualConstraintTests.SameValueDifferentTypeExactMessageMatch() in C:\Users\sweer\Documents\Development\OSS-stevenaw\nunit\src\NUnitFramework\tests\Constraints\EqualConstraintTests.cs:line 838

2) Failed : NUnit.Framework.Tests.Constraints.ValueTupleEqualityTests.ValueTupleElementsAreComparedUsingNUnitEqualityComparer
  Assert.That(a, Is.EqualTo(b))
  Expected: (<empty>)
  But was:  (<empty>)
   at NUnit.Framework.Tests.Constraints.ValueTupleEqualityTests.ValueTupleElementsAreComparedUsingNUnitEqualityComparer() in C:\Users\sweer\Documents\Development\OSS-stevenaw\nunit\src\NUnitFramework\tests\Constraints\ValueTupleEqualityTests.cs:line 74
```

stevenaw:
Update: I tried my hand at an `IResolvableConstraint<T>` and `IConstraint<T>` in order to support a signature like `Assert.That(TActual, IResolvableConstraint<TActual>)` so that we could get compile-time wiring to an overload when the expected and actual are the same type.

The implementation kept them distinct from the existing non-generic versions in order to avoid lots of internal rework but caused innumerable overload resolution conflicts for the compiler.

OsirisTerje:
@stevenaw Do you think it is possible to get through those  overload resolution conflicts ?

OsirisTerje:
Merged with main.  Got a compiler error at line 1048
image
There is no Using method that will accept two type parameters, and hard to see what would do it.  The lambda does two things, converts the the actual and expected values to the same type, and does a comparison.

When adding an object cast for the int, causing it to use the old EqualConstraint, the tests works.  Can be a temporarily fix. 

Running the tests, it fails `SameValueDifferentTypeExactMessageMatch`.  The messages does not match. This is the same as you have above (case 1). 

The second case seems to work.

The comparer function only take one generic type. That will not work when different types are present.  Can we change that?

OsirisTerje:
This is the type dependency diagram for the EqualConstraint<T>
image

I believe it will be hard to start with this one, without fixing the dependencies it has first.  Re-reading #53 @CharliePoole also has some valid points.  Not sure if we can make this work without starting with the Constraint class itself.

manfred-brands:
> There is no Using method that will accept two type parameters, and hard to see what would do it. The lambda does two things, converts the the actual and expected values to the same type, and does a comparison.

Actually there is in _main_: [Using<TActual, TExpected>](https://github.com/nunit/nunit/blob/main/src/NUnitFramework/framework/Constraints/EqualConstraint.cs#L370..L381)

This allows for comparing different types.

manfred-brands:
> I believe it will be hard to start with this one, without fixing the dependencies it has first. Re-reading #53 @CharliePoole also has some valid points. Not sure if we can make this work without starting with the Constraint class itself.

Yes, the `TApply` in `IConstraint` must be a type parameter on the type, not on the `Apply` method.
I haven't looked for a while, but I remember that gave a lot of problems as in the `EqualConstaint` the `TApply` could be different from `T`. This would still allow calling it with a different parameter.

stevenaw:
Sorry, a combination of busy work schedule and jetlag kept me from chiming in earlier. 
I agree this is a tough one to solve, and that the best chance could be if it were possible to do it in pieces. 

I do think that the best chance at making this happen could be to first focus on equality tests as @manfred-brands did here as they are likely the most-used and also serve as a basis for other collection-oriented comparisons.

My earlier comment in this PR had been coming from an attempt at only doing generics when the compiler can resolve both `expected` and `actual` to be the same type as this would save us having to handle the convenience conversions NUnit supports.The overload resolution issue I hit stemmed from having separate generic and non-generic interfaces implemented by the same constraint. I've since wondered if implicit conversion operators could've helped but haven't tried yet. 

Unfortunately I don't know if I have a large enough chunk of time this summer to pursue it in a meaningful way given the size we suspect it requires and would welcome help or experimentation from others. I'll also continue to poke at it as well as time allows to try different approaches here

manfred-brands:
I found a `WIP` commit in my branch that actually introduces `IConstraint<T>`.
I need to dig into this to see whether it was a deep hole I couldn't get out of or whether it is promising.

stevenaw:
I've gotten much closer on this failure:

> 1) Failed : NUnit.Framework.Tests.Constraints.EqualConstraintTests.SameValueDifferentTypeExactMessageMatch

It looks like there's some special formatting and handling in the non-generic `EqualConstrainResult` that this test relies on. I think the simplest solution is just to use `EqualConstrainResult` instead. There's a few minor things we'd have to adapt from our generic constraint to be able to pass in, but it should be fairly simple

stevenaw:
That error message discrepancy failure is now fixed. I'll have to do some digging on the others but I see it's using the built-in `Equals()` method when trying to compare valuetuples with tolerance. I'm not sure but it may be that we will need to apply some generic version of the NUnitEqualityComparer to ensure NUnit's equality rules are applied here.

stevenaw:
@manfred-brands @OsirisTerje The tests are now passing 😃 
I had to tie it into the default comparers to keep some things backwards compatible in particular around tolerance, but I've added a generic overload there to forward the generic type definitions down. I think it may also make it easier for a future iteration to allow all types to use the generic `EqualTo()` rather than just `struct, IEquatable<T>`.

I'm leaving this as a `WIP` for now so that we can dissect it and figure out if this is a good direction.

@nunit/framework-team open to thoughts?

src/NUnitFramework/framework/Constraints/EqualConstraint{T}.cs
```diff
@@ -0,0 +1,362 @@
...
+
+        #endregion
+
+        #region Constructor
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EqualConstraint"/> class.
+        /// </summary>
+        /// <param name="expected">The expected value.</param>
+        public EqualConstraint(T expected)
+            : base(expected)
```
manfred-brands, line 35: Wouldn't we not loose all 'generic' benefits of not boxing values when passing it to the non-generic base class as `object`.

stevenaw: To be honest, I can't recall why I added that. Potentially it may, but only for the things that directly reference base class members rather than overriding generic equivalents. 

From my past attempts I found one of the challenges with making generic constraints "work" was internal pieces relying heavily on either `Constraint` or `IConstraint` as a base. I liked here how you were able to inherit from `Constraint` here while still overriding all the important pieces to get generic typing at the `ApplyTo()` level.

I've begun wondering if a transition approach to getting generic constraints into the codebase could be something like this PR where we still inherit from `Constraint` but override as much as possible - or at least where important - to enable generic typing. That would at least allow us to have the constraints "chain" together at the assertion level for strong-typing there without having to do all the plumbing in one go. One way that could work is if we added an `IConstraint<T>` which defines strongly-typed public members to carry generic typing through successive chained constraints while still inheriting from the base `Constraint` class that all the plumbing uses.

All just ideas. I was pleasantly surprised to see how easily generic typing could start to apply down at the comparer level without duplicating codepaths there

stevenaw: Looking at this again, I think I better understand your question and I can't recall why I added this. I can try removing

src/NUnitFramework/framework/Constraints/EqualConstraint{T}.cs
```diff
@@ -0,0 +1,362 @@
...
+            {
+                var value = _fallbackComparer.AreEqual(_expected, actual, ref _tolerance);
+                return new EqualConstraintResult(this, GetConstraintResultData(), actual, value);
+            }
+
+            throw new ArgumentException("Wrong type", nameof(actual));
+
+            EqualConstraintResult.ResultData GetConstraintResultData() => new()
+            {
+                // TODO: Handle string-specific cases separately?
```
stevenaw, line 312: After some experimentation, I believe we'll have to add these here onto the generic constraint too. The below test will pass with NUnit 4.2.1 but will give a compiler error presently

```
[Test]
public void TestSomething()
{
    var a = (2, 5, "HELLO");
    var b = (2, 5, "hello");

    Assert.That(a, Is.EqualTo(b).IgnoreCase);
}
```

src/NUnitFramework/framework/Constraints/EqualConstraint{T}.cs
stevenaw: It appears it's used by the base class for XML generation, but I'll push a change shortly that at least cleans up a few references within this class.

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -176,6 +176,18 @@ public bool AreEqual(object? x, object? y, ref Tolerance tolerance)
         {
             EqualMethodResult result = AreEqual(x, y, ref tolerance, new ComparisonState(true));
 
+            return ProcessEqualResult(x, y, result);
+        }
+
+        internal bool AreEqual<T>(T? x, T? y, ref Tolerance tolerance)
```
stevenaw, line 182: Something like this, if left in the final design, could also lend itself well to #3985

src/NUnitFramework/framework/Is.cs
```diff
@@ -141,6 +141,15 @@ public abstract class Is
 
         #region EqualTo
 
+        /// <summary>
+        /// Returns a constraint that tests two items for equality
+        /// </summary>
+        public static EqualConstraint<T> EqualTo<T>(T expected)
```
stevenaw, line 147: Something for consideration. Adding this overload will cause at least some call sites previously resolving to the `object` overload to now resolve here. The difference in return type between these methods _could_ then be breaking unless our new `EqualConstraint<T>` were to subclass `EqualConstraint` itself in the short term.

stevenaw: Or perhaps the reverse
```
EqualConstraint : EqualConstraint<object>
```

manfred-brands: Yes, I had thought of that as well.

stevenaw: Thanks! With "deep coding" time harder to come by at the moment, these comments were a way of documenting where I had to leave off for the next time myself or someone else continues to chip away. 

Glad to hear we've been thinking about this the same way!

stevenaw:
I think this is fairly close. I've removed the generic constraints now and all tests are passing but one, `CanCompareUncomparableTypes`
https://github.com/nunit/nunit/blob/0928fb1088ac43a04a6b34adefb3a764fda74971/src/NUnitFramework/tests/Constraints/EqualConstraintTests.cs#L835

The issue is that for the following line:
```csharp
var comparer = new ConvertibleComparer();
Assert.That(2 + 2, Is.EqualTo("4").Using(comparer));
```

The custom comparer is not running as `CanCompare()` [returns false here](https://github.com/nunit/nunit/blob/0928fb1088ac43a04a6b34adefb3a764fda74971/src/NUnitFramework/framework/Constraints/EqualityAdapter.cs#L156):
```
public override bool CanCompare(object x, object y)
            {
                return TypeHelper.CanCast<T>(x) && TypeHelper.CanCast<T>(y);
            }
```

Given that `T` has changed from `object` to `string` and `int` can cast to `object` but not to `string`

stevenaw:
All the tests are passing and I don't believe there are any breaking changes in the mix.
This is ready for review

manfred-brands (changes requested):
@stevenaw Thanks for all the effort but I'm afraid that the result isn't much better than what we have already. Sorry.

The changes in `Numerics` and `NUnitEqualityComparer` make the compiled IL code worse due to all the boxing back to `object`.

Even though it add some sort of type safety into `EqualConstraint` it still has the big drawback that every possible modifier is allowed on every type.
I think that the generic approach isn't going to work.

Maybe something like the below is what we need;
* `Is.EqualTo(string)`  -> `EqualStringContraint`
* `Is.EqualTo(double)` -> `EqualNumericsConstraint`
* Is.Equal(DateTimeOffset)` -> `EqualDateTimeOffsetConstraint`
* `Is.EqualTo(object)` -> `EqualConstraint` (existing for all others)

This allows dropping specific properties that don't make sense for the type at hand, such as 'Ticks' for tolerance on string.
I'll put up a PR for you to get my idea.

stevenaw:
Thanks @manfred-brands 
I'm fine with type-specific type-safe constraints if we are ok with the increased API complexity, but I suspect we'll have to keep these "catch all" properties for cases like valuetuples or collections.

The boxing issue in this PR should be easy enough to overcome by simply carrying the generic typing down further. I hadn't done that yet in order to get feedback on the higher level constraint design

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -193,7 +205,7 @@ public bool AreEqual(object? x, object? y, ref Tolerance tolerance)
             static string GetType(object? x) => x?.GetType().FullName ?? "null";
```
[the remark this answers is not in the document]
stevenaw: Thanks for catching this. We may need to keep the `GetType()` rather than switch to `typeof()` as this is called from more than just the equal constraint, but a generic version would certainly be a welcome addition.

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -193,7 +205,7 @@ public bool AreEqual(object? x, object? y, ref Tolerance tolerance)
...
         }
 
-        internal EqualMethodResult AreEqual(object? x, object? y, ref Tolerance tolerance, ComparisonState state)
+        internal EqualMethodResult AreEqual<T>(T? x, T? y, ref Tolerance tolerance, ComparisonState state)
```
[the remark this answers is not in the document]
stevenaw: Oh right. I'd been thinking this would be no different than boxing at the assert level but I "think" what you're saying is that in some cases, like collections, this would be called O(n) times?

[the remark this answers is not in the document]
stevenaw: Oh wait. I got it now

stevenaw:
Thanks for the advice @CharliePoole 

I had tried that as well in an early attempt some months ago and, I admit, I had stopped at that point once it became apparent that it must be generic "all the way down". I am now tempted to give it another try as it seems I had been on the right path after all 🙂

Perhaps some of the understanding I've gained from this and other attempts could help. There have been a few by me as well!

src/NUnitFramework/framework/Constraints/EqualConstraint{T}.cs
```diff
@@ -0,0 +1,307 @@
...
+            AdjustArgumentIfNeeded(ref expected);
+            _expected = expected;
+        }
+        #endregion
+
+        #region Constraint Modifiers
+        /// <summary>
+        /// Flag the constraint to ignore case and return self.
+        /// </summary>
+        public new EqualConstraint<T> IgnoreCase => (EqualConstraint<T>)base.IgnoreCase;
```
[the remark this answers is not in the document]
stevenaw: I'm also not a fan. Having to duplicate all these with `new` is not great for long term maintainability either.

stevenaw:
Closing this for now. Lots has been learned from this and, ultimately, I think I've come to the conclusion mentioned above that this must be done at a more base level within the constraint class and builder itself

OsirisTerje:
> I am now tempted to give it another try as it seems I had been on the right path after all 

You will continue with this, right !

And then, wounldn't this work (in this PR) then be what we need here, or will this also be .... idk... simpler ?

stevenaw:
I'll be continuing with it 🙂
Tackling it at the base level should make the changes to EqualsConstraint itself simpler as we shouldn't (I think) need to `new` all these definitions. 

The eventual PR itself though "may" be larger in scope, depending on how opt-in a generic constraint builder could be

stevenaw:
Coming back to this, @OsirisTerje @manfred-brands 

should we reopen this to enable merging just the generic equal constraint early while the larger `Constraint<T>` work happens? It would allow us to start releasing something with a strongly-typed public API sooner so that consumers could benefit sooner while the rest of the rework happens.

OsirisTerje:
That makes sense to me.
Things don't need to be perfect.

stevenaw:
Ok, will do.

"Not perfect" no, my concern had initially been that getting this merged early when we knew we wanted a larger rework could constrain future direction but I think it's _possible_ that the scope of changes needed for the broader `Contraint<T>` work could be better for an eventual major version bump anyways - at which point minor breaking challenges would be more acceptable.

As far as these changes, there were a few points about `NumericComparer` from @manfred-brands that I'll address in the process of resurrecting the PR. I think the simplest way is to just revert those and then tackle them after we also make a few higher-level constraints generic too. The numerics _will_ likely need to be tackled all at once, I suspect, given that they all potentially call each other.

PR #4402, opened 2023-06-15
Add PartitionFilter to run a subset of tests based off a hash/partition key

Plasma:
# Overview
Added a new filter, PartitionFilter, that allows you to filter to a subset of arbitrary tests based on a partition number and count.

This is helpful when you may want to run a subset of tests (eg, across 3 machines - or partitions), each with a separately assigned partition number and fixed partition count

Fixes #4391 and is a port of changes for master branch (See #4392)

PR #4392, opened 2023-05-22
Add PartitionFilter to run a subset of tests based off a hash/partition key

Plasma:
# Overview
Added a new filter, PartitionFilter, that allows you to filter to a subset of arbitrary tests based on a partition number and count.

This is helpful when you may want to run a subset of tests (eg, across 3 machines - or partitions), each with a separately assigned partition number and fixed partition count

Fixes #4391

PR #4457, opened 2023-08-10
Add PartitionFilter to run a subset of tests based off a hash/partition key

### LINKED ISSUE #4391: Hash/Partition based Test Filter
# Suggestion
Implement a Hash/Partition based Test Filter, so that tests match the filter based on a hash and modulo comparison. This is separate and different to parallel execution attribute.

This would allow a subset of tests to be ran (and others excluded) based on a partition number, in order to partition the running of the same (large, slow) test suite across multiple machines by selectively filtering to a subset of tests based on a partition count and number.

## Use case
Using GitHub Actions to run nunit tests, it would be great to use the `matrix` option in GitHub Actions to run multiple copies of the test process across multiple machines at once in order to speed up test runs for large test suites easily.

A filter parameter that allowed me to specify the number of partitions (eg, 10) and the partition number to filter tests for (eg, 3) would make it very easy to partition the running of tests in GitHub Actions and other runners where sub-jobs are supported.

## Usage example
``--filter <partition/max partitions>``

Algorithm may be something like ``IsMatch == sha1(fully qualified test name) % max partitions == partition``

## Contribution
I'm happy to contribute a PR if this sounds like a reasonable contribution.

stevenaw:
Recreation of PR https://github.com/nunit/nunit/pull/4402 from source commits in order for CI to cooperate. Previous CI failures seemed to be a transitory timeout, but it's over 30 days since so it can't be rerun in Github Actions. EDIT: It's since been updated to work with the recent repo restructuring.

Original text:

# Overview
Added a new filter, PartitionFilter, that allows you to filter to a subset of arbitrary tests based on a partition number and count.

This is helpful when you may want to run a subset of tests (eg, across 3 machines - or partitions), each with a separately assigned partition number and fixed partition count

Fixes #4391 and is a port of changes for master branch (See #4392)


stevenaw:
Putting this up under the original commit seems to have re-started the original PR's job. Once the build passes I'll close this PR and track things back under the original

stevenaw:
@Plasma It seems a few things have moved namespaces since you started this. Given it's been a while, I took the liberty of syncing it back up and updating the partition tests with recalculated partition numbers. I'd appreciate if you had time for a review of the changes, but there's no obligation on your part to do so.

Plasma:
Thanks @stevenaw that looks reasonable to me

manfred-brands (approved):
Code looks good.
I don't know if the first 4 bytes of sha256 results in a uniform distribution.
if you calculated the partition number for each test in some project do all partitions have a similar number of tests?

src/NUnitFramework/framework/Internal/Filters/PartitionFilter.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        /// <summary>
+        /// Computes an unsigned integer hash value based upon the provided string
+        /// </summary>
+        private static uint ComputeHashValue(string name)
+        {
+            using var hashAlgorithm = SHA256.Create();
+
+            // SHA256 ComputeHash will return 32 bytes, we will use the first 4 bytes of that to convert to an unsigned integer
+            var hashValue = hashAlgorithm.ComputeHash(Encoding.UTF8.GetBytes(name));
```
manfred-brands, line 110: I wonder how expensive this is and if we can not simply use `name.GetHashCode()`

stevenaw: That's a good point, and not something I thought of in the PR to 3.14
Thanks for raising it @manfred-brands 

I'm going to merge this in order to sync the changes between the branches but this is certainly a possibility worth a future change

Plasma: I've put together a dotnetfiddle to check, https://dotnetfiddle.net/xvn7zM

Both implementations give roughly the same distribution, speed is faster in GetHashCode by ~19x but trivial in terms of timing (50ms for 1m hashes in GetHashCode, 950ms for SHA256).

While both are evenly distributed, I note the partition assigned via GetHashCode changes on each iteration (hashing the memory address of the string I assume) where-as the SHA256 of the same text is stable between iterations.

manfred-brands: Thanks @Plasma
For .NET6 and later the performance difference can be limited to a factor 10 when using `var hashValue = SHA256.HashData(Encoding.UTF8.GetBytes(name));` instead.
The same improvement can be obtained by creating the SHA256 instance once instead of for ever test name.

Maybe the changing distribution of `GetHashCode` is a good thing so tests are more randomly placed on different build runners instead of always on the same.

Plasma: Thanks @manfred-brands good to know RE: HashData method for .NET 6 onwards. I didn't create a single field/instance within the method as the attribute isn't `IDisposable` and thought it may be too much of a change.

RE: `GetHashCode` I lean towards having stability between runs (assuming the partition count, test names and NUnit version didn't change of course) as being important, for example, re-trying a failed GH Action job for a partition of tests that failed should re-run the same tests next time (which isn't promised with `GetHashCode`), and we can rely upon the hashing function fairly splitting tests between partitions.

PSanetra:
@stevenaw @Plasma is it possible to use this filter when the tests are executed via `dotnet test` in nunit 4 beta 1?

stevenaw:
@PSanetra Sorry we're a little behind on the documentation side of this.

While yes I do believe this is technically within beta 1 of nunit 4 framework, it won't be easily usable until the command-line parsing piece merges and eventually makes its way into the Nunit Adapter package called by `dotnet test`.

Glad to see this new feature is of interest to you!

PSanetra:
@stevenaw is there some issue to track implemenation of a NUnit4TestAdapter that supports this feature?

stevenaw:
Thanks for putting this back on our radar @PSanetra 
There's a bit of discussion in the main issue (https://github.com/nunit/nunit/issues/4391) about this need but I hadn't seen an issue filed yet to add it to the adapter. I've added https://github.com/nunit/nunit3-vs-adapter/issues/1138 just now

PR #4467, opened 2023-09-14
Issue2466 propagate stack trace to child result

### LINKED ISSUE #2466: Missing stack trace when exception occurs during OneTimeSetUp
@ericnewton76 commented on [Sat Dec 17 2016](https://github.com/nunit/nunit3-vs-adapter/issues/284)

image

Missing the stack trace for the error if it bubbles up from a OneTimeSetUp method.

---

@ericnewton76 commented on [Sat Dec 17 2016](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-267782375)

I have forked vs-adapter and I'm looking through the code, and this type of scenario isn't tested.

https://github.com/ericnewton76/nunit3-vs-adapter/commit/09901cd4d26afbb4aa6dc02f28ed79c5abb9556c
I added a class to mock-assembly project (at the end of MockAssembly.cs) but i'm not sure how to run the tests against this new mock assembly.   

Additionally I wonder if nunit itself has the stack trace for the exception during OneTimeSetUp and if the vs-adapter is displaying all that it has.

---

@ericnewton76 commented on [Sat Dec 17 2016](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-267783470)

Ok, attempted to run build script, because I see it includes test running, but MSBuild giving me some kind of VS2017 sxs installation problem, so dont install VS2017 yet.  The VS developers are lamenting the fact that VS2015 cant build VS2017 at the moment.  I'm guessing this is related to my problem. with getting msbuild to build successfully.  https://github.com/github/VisualStudio/issues/655#issuecomment-261895992

---

@OsirisTerje commented on [Sun Dec 18 2016](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-267813809)

I can repro this, and I see we don't add the OneTimeSetup crash information to the output.  The information comes from the TestFixture node, and the site info is "Child". 
 @CharliePoole @rprouse : I see the Setup and Teardown have its own site names, but OneTimeSetup and OneTimeTearDown are not listed there.  And I can't see any of these site names in the output result file.  I only see Child and Parent.   

---

@CharliePoole commented on [Sun Dec 18 2016](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-267816835)

One time setup is the parent setup. It's difficult to know what to do here since VS has no place to report suite results but only test cases. We could report the suite result on each test case but that's potentially a lot of duplication. I compromised by reporting only the message but the stack trace could be reported as well.

---

@ericnewton76 commented on [Sun Dec 18 2016](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-267831160)

How does nunit handle an exception in onetimesetup?  Does it basically mark the whole test fixture as invalid, or does it keep retrying onetimesetup for each test in the fixture (Which to me would makes most sense)

---

@rprouse commented on [Sun Dec 18 2016](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-267833767)

It marks the whole fixture as failed, it doesn't retry. We assume that your setup code is deterministic and will fail consistently.

---

@OsirisTerje commented on [Sun Dec 25 2016](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-269140754)

Is the output I see in the console output (testresult.xml) identical to what the adapter gets as a response?

---

@CharliePoole commented on [Mon Dec 26 2016](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-269141326)

Yes, the adapter gets the entire XML result in pieces as each test completes. (It could also get the completed result but it doesn't need it since the info is given to VS as it arrives.) The result of a test suite is pretty much ignored, since it's not associated with any of the tests that have been discovered by VS. You can see that happen here: https://github.com/nunit/nunit3-vs-adapter/blob/master/src/NUnitTestAdapter/NUnitEventListener.cs#L148

The result of a test case is handled in the method right above that one. It records what is contained in the test case result.

---

@OsirisTerje commented on [Mon Dec 26 2016](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-269141549)

Thanks!  And what is the real meaning of the term/attribute "site" ?   It can take values like Child and Parent, but also SetUp and TearDown and the like ? 

---

@CharliePoole commented on [Mon Dec 26 2016](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-269142030)

In the framework, it's an enum. We use the enum names as attribute values. The default is "Test" which is what usually happens. The code in the adapter only logs a message for setup and teardown suite failures, since those don't get reflected anywhere else.

Note that the one time setup error message was being shown in the original issue description. That indicates it must be coming in the test case failure. You want to examine whether that's happening with the stack trace too. I suspect not.

---

@CharliePoole commented on [Mon Dec 26 2016](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-269142109)

BTW these attributes should all be documented here: https://github.com/nunit/docs/wiki/Test-Result-XML-Format

---

@carlin-q-scott commented on [Wed Apr 19 2017](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-295403429)

So if I want a OneTimeSetup that gives me sufficient detail to figure out the cause of the failure, should I do something like this?

```
static bool oneTimeSetupRan;

[SetUp]
public void OneTimeSetup()
{
    if (!oneTimeSetupRan)
    {
         oneTimeSetupRan = true;
    }
}
```

---

@CharliePoole commented on [Wed Apr 19 2017](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-295461226)

@carlin-q-scott That would definitely simulate the one-time setup for a fixture. It's a pity to have to go to that trouble though.

---

@CharliePoole commented on [Wed Apr 19 2017](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-295462238)

@nunit/vs-extensions-team What do we want to do here? Is it simply a limitation of the VS interface? Should we try to report the stack trace for each error? That would require our remembering the error info for each one time setup failure and reporting it for the corresponding test cases. Non-trivial, especially in the presence of multiple fixtures running in parallel.

---

@blorger commented on [Fri Aug 18 2017](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-323297593)

This is not just limitation of VS interface. NUnit simply does not provide call stack information. Problem remains if you execute tests with NUnit console, no call stack on console output or in result file.
Why not simply change content of a message. Provide entire exception description rather than just exception message. This way there will be no need to debug a test or modify it (move one time setup code inside test, or wrap each setup method in try/catch block to log exception ...) just to get information where failure occurs.

---

@OsirisTerje commented on [Sun Oct 01 2017](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-333383025)

If I have an exception in the OneTimeSetup, for each test it is reported, like this:
OneTimeSetUp: ClassLibrary11.WhateverException : Exception of type 'ClassLibrary11.WhateverException' was thrown.

The stack trace is coming with the suite finished event a bit later, but then the test results have already been sent.  

I think it would be better just to return the full exception stacktrace with each failing tests.  It would only slow down failing tests, so that should be no big matter.  And then no changed code is needed for the adapter. 

@rprouse / @CharliePoole  If you agree, can we add an issue on NUnit for this, or move this to NUnit ? 

---

@CharliePoole commented on [Sun Oct 01 2017](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-333388234)

If we want to make it look a certain way for VS, then I think we should make it look that way in the adapter. NUnit itself attaches the error to the test that failed, which was the fixture. That's a pretty fundamental thing about how NUnit looks at tests: everything is a test - test cases, methods, classes, namespaces and the assembly. VS looks at the tests as a list of cases. Since the VS view is simpler, it should be possible to make things appear that way if it's really desired.

---

@OsirisTerje commented on [Sun Oct 01 2017](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-333400566)

Yes, we can do this in the adapter, but NUnit **is** reporting this per test case,  just with less information.  It has split the information of the exception between the test case and the fixture, with the test case reporting the exception and the message,  and the fixture reporting the same plus the stacktrace. 

---

@CharliePoole commented on [Mon Oct 02 2017](https://github.com/nunit/nunit3-vs-adapter/issues/284#issuecomment-333433563)

Add the issue to NUnit in that case, so we can discuss it there. I think it should be a new issue, since up to now this has been about how the adapter displays what it gets from NUnit, rather than changing what NUnit delivers.

We have a similar problem in the console runner, because the console runner historically only counted test cases and only displayed test case failures and errors. Over time, we started displaying more info. But the console has access to the entire XML, so modifying the output was easier.

manfred-brands:
Fixes #2466

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,2 +1,2 @@
 @echo off
-powershell ./build.ps1 %CAKE_ARGS% %*
+powershell -ExecutionPolicy ByPass -NoProfile ./build.ps1 %CAKE_ARGS% %*
--- a/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
@@ -115,7 +115,7 @@ namespace NUnit.Framework.Internal.Execution
                                         case TestStatus.Skipped:
                                         case TestStatus.Inconclusive:
                                         case TestStatus.Failed:
-                                            SkipChildren(this, Result.ResultState.WithSite(FailureSite.Parent), "OneTimeSetUp: " + Result.Message);
+                                            SkipChildren(this, Result.ResultState.WithSite(FailureSite.Parent), "OneTimeSetUp: " + Result.Message, Result.StackTrace);
                                             break;
                                     }
                                 }
@@ -313,14 +313,14 @@ namespace NUnit.Framework.Internal.Execution
         private void SkipFixture(ResultState resultState, string message, string stackTrace)
         {
             Result.SetResult(resultState.WithSite(FailureSite.SetUp), message, StackFilter.DefaultFilter.Filter(stackTrace));
-            SkipChildren(this, resultState.WithSite(FailureSite.Parent), "OneTimeSetUp: " + message);
+            SkipChildren(this, resultState.WithSite(FailureSite.Parent), "OneTimeSetUp: " + message, stackTrace);
         }
 
-        private void SkipChildren(CompositeWorkItem workItem, ResultState resultState, string message)
+        private void SkipChildren(CompositeWorkItem workItem, ResultState resultState, string message, string stackTrace)
         {
             foreach (WorkItem child in workItem.Children)
             {
-                SetChildWorkItemSkippedResult(child.Result, resultState, message);
+                SetChildWorkItemSkippedResult(child.Result, resultState, message, stackTrace);
                 _suiteResult.AddResult(child.Result);
 
                 // Some runners may depend on getting the TestFinished event
@@ -328,13 +328,13 @@ namespace NUnit.Framework.Internal.Execution
                 Context.Listener.TestFinished(child.Result);
 
                 if (child is CompositeWorkItem)
-                    SkipChildren((CompositeWorkItem)child, resultState, message);
+                    SkipChildren((CompositeWorkItem)child, resultState, message, stackTrace);
             }
         }
 
-        private void SetChildWorkItemSkippedResult(TestResult result, ResultState resultState, string message)
+        private void SetChildWorkItemSkippedResult(TestResult result, ResultState resultState, string message, string stackTrace)
         {
-            result.SetResult(resultState, message);
+            result.SetResult(resultState, message, stackTrace);
             result.StartTime = Context.StartTime;
             result.EndTime = DateTime.UtcNow;
             result.Duration = Context.Duration;
--- a/src/NUnitFramework/tests/Attributes/OneTimeSetUpTearDownTests.cs
+++ b/src/NUnitFramework/tests/Attributes/OneTimeSetUpTearDownTests.cs
@@ -149,6 +149,28 @@ namespace NUnit.Framework.Attributes
         }
 
         [Test]
+        public void FailedSetUpStacktracePropogatesToTestResult()
+        {
+            SetUpAndTearDownFixture fixture = new SetUpAndTearDownFixture();
+            fixture.ThrowInBaseSetUp = true;
+            ITestResult result = TestBuilder.RunTestFixture(fixture);
+
+            Assert.That(result.ResultState.Site, Is.EqualTo(FailureSite.SetUp));
+            Assert.That(result.StackTrace, Is.Not.Null);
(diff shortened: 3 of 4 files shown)


OsirisTerje (approved):
LGTM

PR #4475, opened 2023-09-29
Improve debugging experience

RenderMichael:
These changes make types more transparent when developing against them. Attached images show the results of these additions.

image

image

Since I noticed it, I made Tolerance.Range `readonly`.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionTally.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionTally.cs
@@ -2,6 +2,8 @@
 
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
+
 using NUnit.Framework.Internal.Extensions;
 
 namespace NUnit.Framework.Constraints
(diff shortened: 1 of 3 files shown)


OsirisTerje (approved):
Looks good to me.
I can't see it has any adverse effects. 
Overriding ToString is something to be careful about, but on Tolerance I can't see it raising any alarm.
@manfred-brands @stevenaw  Comments?
@RenderMichael  Normally we want PRs to be responses to an issue, and also that any PR should have tests included.  In this case the changes are small, and seem harmless, so.....   I do see the changes don't trigger any of the existing CI tests, but would be nice if you could confirm that running   `build --target Acceptance` is all green.

RenderMichael:
> Normally we want PRs to be responses to an issue, and also that any PR should have tests included.
@OsirisTerje That's completely fair, I can add the screenshotted code as an actual test. Let me know if it's worth it.

OsirisTerje:
@RenderMichael It actually makes sense, to get tests matching the screenshots.  This issue will drown over time, but the tests will stay - and then serve as documentation.  :-)

OsirisTerje:
Awesome!  Thanks!

PR #4478, opened 2023-10-02
Add the ability to use params arguments properly when using a parametrised test fixture

### LINKED ISSUE #1459: Support for params keyword in parameterized test fixtures
Below is an example where the `params` keyword can be used in v2 to eliminate the need to have one constructor/TestFixtureSetUp/OneTimeSetUp for each TestFixture

```
    [TestFixture(1, 1)]
    [TestFixture(3, 1, 2)]
    [TestFixture(6, 1, 2, 3)]
    class Test
    {
        private readonly int total;
        private readonly int[] integers;

        public Test(int total, params int[] integers)
        {
            this.total = total;
            this.integers = integers;
        }

        [Test]
        public void Can_add_numbers()
        {
            Assert.That(integers.Sum(), Is.EqualTo(total));
        }
    }
```

Whilst this kind of test works in v2, when run in v3 (tried in version 3.2.1 of console runner) it seems an exception is thrown within NUnit:

```
1) Invalid : ConsoleApplication1.Test(1,1)
No suitable constructor was found

2) Invalid : ConsoleApplication1.Test(3,1,2)
No suitable constructor was found

3) Invalid : ConsoleApplication1.Test(6,1,2,3)
No suitable constructor was found
```

Related discussion:
https://groups.google.com/forum/#!topic/nunit-discuss/iPuVKU6bFws

Shiney:
Added some tests for this and made the methods that call the constructors on TestFixtures able to allow params arguments. This doesn't really address the duplication with similar code for calling test methods but that didn't seem particularly quick to also fix as part of this.

Fixes #1459

### DIFF
--- a/src/NUnitFramework/framework/Internal/Reflect.cs
+++ b/src/NUnitFramework/framework/Internal/Reflect.cs
@@ -81,20 +81,32 @@ namespace NUnit.Framework.Internal
                 throw new InvalidTestFixtureException(type.FullName + " does not have a suitable constructor");
             ParameterInfo[] parameterInfos = ctor.GetParameters();
 
-            // TOOD is it worht sharing code with GetParametersForTestCase
             if (parameterInfos.Length > 0)
             {
                 ParameterInfo parameterInfo = parameterInfos.Last();
                 if (parameterInfo.HasAttribute<ParamArrayAttribute>(false))
                 {
-                    //TODO need a test that we only do this if the last parameter passed in is the right type
-                    int paramsOffset = ctor.GetParameters().Length - 1;
-                    var paramArray = Array.CreateInstance(parameterInfo.ParameterType.GetElementType(), argTypes.Length - ctor.GetParameters().Length + 1);
-                    for (int i = 0; i < paramArray.Length; i++)
+                    if (arguments.Length == parameterInfos.Length 
+                        && parameterInfo.ParameterType.IsAssignableFrom(arguments[parameterInfos.Length - 1]?.GetType()))
                     {
-                        paramArray.SetValue(arguments[i + paramsOffset], i); 
+                        // Don't convert arguments as there was already an array we could use.
                     }
-                    arguments = arguments.Take(parameterInfos.Length - 1).Concat(new object[] { paramArray }).ToArray();
+                    else
+                    {
+                        Type? elementType = parameterInfo.ParameterType.GetElementType();
+                        if (elementType is null)
+                        {
+                            throw new InvalidTestFixtureException(type.FullName + " params argument did not have an element type");
+                        }
+                        int paramsOffset = ctor.GetParameters().Length - 1;
+                        var paramArray = Array.CreateInstance(elementType, argTypes.Length - ctor.GetParameters().Length + 1);
+                        for (int i = 0; i < paramArray.Length; i++)
+                        {
+                            paramArray.SetValue(arguments[i + paramsOffset], i);
+                        }
+                        arguments = arguments.Take(parameterInfos.Length - 1).Concat(new object[] { paramArray }).ToArray();
+                    }
+
                 }
             }
 
@@ -132,7 +144,6 @@ namespace NUnit.Framework.Internal
         /// </summary>
         internal static bool ParametersMatch(this ParameterInfo[] pinfos, Type?[] ptypes)
         {
-            
             bool hasParamsArgument = pinfos.Length > 0 && pinfos[pinfos.Length -1].HasAttribute<ParamArrayAttribute>(false);
 
             if (hasParamsArgument)
@@ -151,7 +162,7 @@ namespace NUnit.Framework.Internal
                 if (hasParamsArgument && i == pinfos.Length - 1)
                 {
                     var elementType = pinfos[i].ParameterType.GetElementType();
-                    if(elementType is not null)
+                    if (elementType is not null)
                     {
                         bool allParamArgsMatched = true;
                         for (int j = i; j < ptypes.Length; j++)
--- a/src/NUnitFramework/framework/Internal/Reflect.cs
+++ b/src/NUnitFramework/framework/Internal/Reflect.cs
@@ -79,6 +79,24 @@ namespace NUnit.Framework.Internal
             ConstructorInfo? ctor = GetConstructors(type, argTypes).FirstOrDefault();
             if (ctor is null)
                 throw new InvalidTestFixtureException(type.FullName + " does not have a suitable constructor");
+            ParameterInfo[] parameterInfos = ctor.GetParameters();
+
+            if (parameterInfos.Length > 0)
+            {
+                ParameterInfo parameterInfo = parameterInfos.Last();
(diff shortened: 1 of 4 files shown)


Shiney:
Will fix some of these build issues later

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -78,6 +78,35 @@ public static object Construct(Type type, object?[]? arguments)
...
+                    }
+                    else
+                    {
+                        Type? elementType = parameterInfo.ParameterType.GetElementType();
+                        if (elementType is null)
+                        {
+                            throw new InvalidTestFixtureException(type.FullName + " params argument did not have an element type");
+                        }
+                        int paramsOffset = ctor.GetParameters().Length - 1;
+                        var paramArray = Array.CreateInstance(elementType, argTypes.Length - ctor.GetParameters().Length + 1);
```
manfred-brands, line 101: You are calling the `ctor.GetParameters()` twice again, just to get its length. However the return value of this method was previously assigned to `parametersInfos`. So please replace the function calls with references to the variable.

src/NUnitFramework/testdata/TestFixtureSourceData.cs
```diff
@@ -216,6 +216,32 @@ private static IEnumerable MyData()
...
+        public int[] Parameters { get; }
+
+        [Test]
+        public void Test()
+        {
+            Assert.That(Parameters, Is.Not.Null);
+            for (int i = 0; i < Parameters.Length; i++)
+                Assert.That(Parameters[i], Is.EqualTo(i + 1));
+        }
+        public static IEnumerable MyData()
```
manfred-brands, line 237: Please make the `MyData` method private in order not to raise `NUnit1028`: Only test methods should be public.

src/NUnitFramework/tests/Attributes/ParameterizedTestFixtureTests.cs
```diff
@@ -199,4 +201,56 @@ public void MakeSureTypeIsInSystemNamespace()
...
+    [TestFixture("Zero")]
+    [TestFixture("One", 1)]
+    [TestFixture("Many", 1, 2, 3, 4)]
+    [TestFixture(1.5, 8.2)]
+    [TestFixtureSource(nameof(SourceData))]
+    public class ParameterizedTestFixtureWithParamsArgument
+    {
+#pragma warning disable NUnit1028 // The non-test method is public
+        public static IEnumerable SourceData()
+#pragma warning restore NUnit1028 // The non-test method is public
```
manfred-brands, line 214: Instead if suppressing the errors, make the `SourceData` method `private`.

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -113,11 +142,42 @@ internal static IEnumerable<ConstructorInfo> GetConstructors(Type type, Type?[]
...
+                    var elementType = pinfos[i].ParameterType.GetElementType();
+                    if (elementType is not null)
+                    {
+                        bool allParamArgsMatched = true;
+                        for (int j = i; j < ptypes.Length; j++)
+                        {
+                            if (!ptypes[j].CanImplicitlyConvertTo(elementType))
+                            {
+                                allParamArgsMatched = false;
+                                break;
```
manfred-brands, line 171: This can be simplified by simply calling `return false` here. Then you don't need the `allParamArgsMatched` variable at all.
Keep the comment below in mind so you also deal with the case where the last parameter is an array instead of a set of values.

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -113,11 +142,42 @@ internal static IEnumerable<ConstructorInfo> GetConstructors(Type type, Type?[]
...
+                            }
+                        }
+                        if (allParamArgsMatched)
+                        {
+                            continue;
+                        }
+                    }
+                }
+
                 if (!ptypes[i].CanImplicitlyConvertTo(pinfos[i].ParameterType))
```
manfred-brands, line 181: If you do this test first, and only if it fails AND we are dealing with the last `params` parameter, do the test above.

manfred-brands (changes requested):
Thanks @Shiney for your contribution.
I only have a few small comments.

src/NUnitFramework/tests/Attributes/ParameterizedTestFixtureTests.cs
Shiney: Oh didn't know that worked that's cool

src/NUnitFramework/framework/Internal/Reflect.cs
Shiney: Sorry I've tried switching it to do it this way round and it doesn't make the code easier because we need to special case it for the example of a params argument that ends up being passed no parameters at all

src/NUnitFramework/framework/Internal/Reflect.cs
Shiney: Unfortunately I don't think this ends up being easier due to the problems with switching the order

Shiney:
Hi @manfred-brands  I tried to do your comments about swapping the order of the checks but it didn't seem to end up making the code much nicer in my opinion because it ended up having to do all sorts of extra special code to deal with the case where zero parameters provided were used for the params argument

src/NUnitFramework/framework/Internal/Reflect.cs
manfred-brands: @Shiney I looked at it a bit more and you are correct.

src/NUnitFramework/framework/Internal/Reflect.cs
manfred-brands: After loading the code in VS, I concur

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -113,11 +142,42 @@ internal static IEnumerable<ConstructorInfo> GetConstructors(Type type, Type?[]
...
 
             for (int i = 0; i < pinfos.Length; i++)
             {
+                if (hasParamsArgument && i == pinfos.Length - 1)
+                {
+                    var elementType = pinfos[i].ParameterType.GetElementType();
+                    if (elementType is not null)
+                    {
+                        bool allParamArgsMatched = true;
+                        for (int j = i; j < ptypes.Length; j++)
```
manfred-brands, line 166:
```suggestion
                        for (int j = i; j < ptypes.Length && allParamArgsMatched; j++)
```

manfred-brands:
@Shiney I tried to push my review changes, but you didn't allow maintainers to update the fork.
Could you please address the other outstanding issues.

Shiney:
Yes will do this soon

Shiney:
I believe I've fixed all the comments and have ticked the Allow edits by maintainers button just in case

manfred-brands:
Thanks @Shiney. I have fixed some spacing issues as earlier this week we enabled enforcing these rules.

manfred-brands (approved):
Nothing further.

PR #4479, opened 2023-10-04
We have no other build for deploying dev packages, so reenabling this…

OsirisTerje:
… for now.


OsirisTerje:
Appveyor parsing error, even if it is identical afaic see to the former one.

PR #4483, opened 2023-10-07
Enable equality/inequality constraints against non-seekable steams

### LINKED ISSUE #4476: Add support to `StreamsComparer` for non-seekable streams
For example, `ZipArchiveEntry.Open()` opens unseekable steams

```csharp
[Test]
public void ZipArchiveTest()
{
    const string fileContentString = "Hello!";
    using var expectedFile = new MemoryStream(Encoding.UTF8.GetBytes(fileContentString));
 
    using var archiveContents = new MemoryStream();
    using ZipArchive actualArchive = GetZipArchiveFromSomewhere(archiveContents); // opened in read mode
    ZipArchiveEntry entry = actualArchive.Entries[0];

    using Stream entryStream = entry.Open(); // an archive in read mode returns a DeflateStream, which is unseekable
    Assert.That(entryStream, Is.EqualTo(expectedFile)); // System.ArgumentException : Stream is not seekable (Parameter 'actual')

    static ZipArchive GetZipArchiveFromSomewhere(MemoryStream archiveContents)
    {
        using (var archive = new ZipArchive(archiveContents, ZipArchiveMode.Create, leaveOpen: true))
        {
            ZipArchiveEntry demoFile = archive.CreateEntry("Hello entry");

            using Stream entryStream = demoFile.Open();

            using var entryFs = new StreamWriter(entryStream);
            entryFs.Write(fileContentString);
        }

        return new ZipArchive(archiveContents, ZipArchiveMode.Read, leaveOpen: false);
    }
}
```

I can work on a PR, let me know if this is something worth pursuing.

edit: improved the repro to be all in-memory

RenderMichael:
Fixes #4476 

Instead of using the stream's length as a stopping condition, we check that the amount of bytes read in the previous read operation is greater than 0.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
@@ -39,9 +39,12 @@ namespace NUnit.Framework.Constraints.Comparers
 
             try
             {
-                if (bothSeekable)
+                if (xStream.CanSeek)
                 {
                     binaryReaderExpected.BaseStream.Seek(0, SeekOrigin.Begin);
+                }
+                if (yStream.CanSeek)
+                {
                     binaryReaderActual.BaseStream.Seek(0, SeekOrigin.Begin);
                 }
 
@@ -73,9 +76,12 @@ namespace NUnit.Framework.Constraints.Comparers
             }
             finally
             {
-                if (bothSeekable)
+                if (xStream.CanSeek)
                 {
                     xStream.Position = expectedPosition;
+                }
+                if (yStream.CanSeek)
+                {
                     yStream.Position = actualPosition;
                 }
             }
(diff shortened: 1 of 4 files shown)


RenderMichael:
For the inequality tests, I made the expected and actual streams of the same size. I can double-up each inequality test where the stream lengths aren't equal, let me know if it's worth it.

Same with streams whose contents exceed the `BUFFER_SIZE` value of 4096. I manually tested against all 4 permutations of the two variables, and they all passed, and I can add tests for it if it's deemed valuable enough.

src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
```diff
@@ -23,31 +23,39 @@ internal static class StreamsComparer
...
                 throw new ArgumentException("Stream is not readable", "actual");
-            if (!xStream.CanSeek)
-                throw new ArgumentException("Stream is not seekable", "expected");
-            if (!yStream.CanSeek)
-                throw new ArgumentException("Stream is not seekable", "actual");
 
-            if (xStream.Length != yStream.Length) return false;
+            bool bothSeekable = xStream.CanSeek && yStream.CanSeek;
+
+            if (bothSeekable && xStream.Length != yStream.Length) return false;
```
stevenaw, line 29: Thanks, this is a nice early bailout optimization

stevenaw (approved):
Thanks @RenderMichael this looks great! I've pulled the changes and they work great. I also can't seem to find any edge cases you haven't already accounted for :)

PR #4488, opened 2023-10-09
Fixes #2841

### LINKED ISSUE #2841: DelayedConstraint calls delegate twice
NUnit 3.10.1, simple test case shown below. The IsComplete is called twice.

```
    [TestFixture()]
    class MyTests
    {
        private int c = 0;


        [Test]
        public void TestConstraint()
        {
            var constraint = Is.True.After(10).Minutes.PollEvery(5000).MilliSeconds;
            Assert.That(IsComplete, constraint);
        }


        public bool IsComplete()
        {
            TestContext.Progress.WriteLine("InComplete");
            if ( c == 5 ) return true;
            c++;
            return false;
        }
    }
```

OsirisTerje:
Added fix from @mikkelbu, comment https://github.com/nunit/nunit/issues/2841#issuecomment-386111091  
Added test to verify it

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
@@ -72,6 +72,9 @@ namespace NUnit.Framework.Constraints
             /// <returns></returns>
             public WithRawPollingInterval PollEvery(int milliSeconds)
             {
+                if (milliSeconds < 0)
+                    throw new ArgumentException("Interval cannot be negative");
+
                 _parent.PollingInterval = new Interval(milliSeconds).InMilliseconds;
                 return new WithRawPollingInterval(_parent);
             }
@@ -102,6 +105,9 @@ namespace NUnit.Framework.Constraints
             /// <returns></returns>
             public WithRawPollingInterval PollEvery(int milliSeconds)
             {
+                if (milliSeconds < 0)
+                    throw new ArgumentException("Interval cannot be negative");
+
                 _parent.PollingInterval = new Interval(milliSeconds).InMilliseconds;
                 return new WithRawPollingInterval(_parent);
             }
@@ -195,6 +201,8 @@ namespace NUnit.Framework.Constraints
         {
             if (delayInMilliseconds < 0)
                 throw new ArgumentException("Cannot check a condition in the past", nameof(delayInMilliseconds));
+            if (pollingIntervalInMilliseconds < 0)
+                throw new ArgumentException("Interval cannot be negative", nameof(pollingIntervalInMilliseconds));
 
             DelayInterval = new Interval(delayInMilliseconds).InMilliseconds;
             PollingInterval = new Interval(pollingIntervalInMilliseconds).InMilliseconds;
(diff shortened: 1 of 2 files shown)


OsirisTerje:
MacOS wasnt happy with the test:
```
1) Failed : NUnit.Framework.Tests.Constraints.DelayedConstraintTests.ThatPollingCallsDelegateCorrectNumberOfTimes
  Assert.That(PollCount, Is.EqualTo(4).After(100, 25))
  Expected: 4 after 100 milliseconds delay
  But was:  3
   at NUnit.Framework.Tests.Constraints.DelayedConstraintTests.ThatPollingCallsDelegateCorrectNumberOfTimes() in /Users/runner/work/nunit/nunit/src/NUnitFramework/tests/Constraints/DelayedConstraintTests.cs:line 241
```

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -227,6 +227,7 @@ public override ConstraintResult ApplyTo<TActual>(TActual actual)
                 {
                     if (nextPoll > now)
                         ThreadUtility.BlockingDelay((int)TimestampDiff(delayEnd < nextPoll ? delayEnd : nextPoll, now).TotalMilliseconds);
+                    now = Stopwatch.GetTimestamp();
```
manfred-brands, line 230: This is correct, as otherwise *nextPoll* would be the same as when going into the method.
A nicer solution would be to only calculate *nextPoll* before the delay, removing two lines of code.
```csharp
while ((now = Stopwatch.GetTimestamp()) < delayEnd)
{
    long nextPoll = TimestampOffset(now, PollingInterval.AsTimeSpan);
    ThreadUtility.BlockingDelay((int)TimestampDiff(delayEnd < nextPoll ? delayEnd : nextPoll, now).TotalMilliseconds);

    ConstraintResult result = BaseConstraint.ApplyTo(actual);
    if (result.IsSuccess)
        return new DelegatingConstraintResult(this, result);
}
```

The test on `if (nextPoll > now)` is also not needed assuming `PollingInterval` is greater than 0.
The latter should be checked in the constraint constructor.
The constructor currently only tests `if (delayInMillisecond < 0)`, but should also not allow a negative *pollingIntervalInMilliseconds*.

manfred-brands (approved):
The change is correct, I only have a suggestion to simplify the code.
And adding a check on *pollingIntervalInMilliseconds > 0* would be nice.

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
OsirisTerje: CAn you add this as a suggested commit?

manfred-brands: No, github doesn't let me suggest changes to non-changed lines.
I can push the change to your branch if you like.

OsirisTerje: Please do :-)

manfred-brands: Change pushed. I have also added checks on pollInterval

mikkelbu (approved):
LGTM

PR #4492, opened 2023-10-10
Add type validation to regex constraint

### LINKED ISSUE #4484: String not-regex constraint succeeds non-`string` actual value
The following code passes
```csharp
[Test]
public void StringMatchesRegex()
{
    // lang=regex
    string regex = "[0-9]";
    string? actual = null;
    Assert.That(actual, Does.Not.Match(regex));
}
```

Should the `RegexConstraint` throw an `ArgumentException` for non-string values (like `null`)?

If so, that a 2-line change (plus tests) and I can spin up the PR myself.

RenderMichael:
Fixes #4484 
tests like
```csharp
Assert.That(1, Does.Not.Match("[A-Z]"));
Assert.That((string?)null, Does.Not.Match("[A-Z]"));
```
will no longer pass.

Bonus performance for converting `regex.Match(...).Success` to `regex.IsMatch(...)`

### DIFF
--- a/src/NUnitFramework/framework/Constraints/RegexConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/RegexConstraint.cs
@@ -72,7 +72,7 @@ namespace NUnit.Framework.Constraints
         {
             string actualString = ConstraintUtils.RequireActual<string>(actual, nameof(actual), allowNull: false);
 
-            return new ConstraintResult(this, actual,_regex.IsMatch(actualString));
+            return new ConstraintResult(this, actual, _regex.IsMatch(actualString));
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/RegexConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/RegexConstraint.cs
@@ -3,6 +3,8 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Text.RegularExpressions;
 
+using NUnit.Framework.Internal;
+
 namespace NUnit.Framework.Constraints
 {
     /// <summary>
@@ -68,8 +70,9 @@ namespace NUnit.Framework.Constraints
         /// <returns>True for success, false for failure.</returns>
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
-            return new ConstraintResult(this, actual,
-                actual is string actualString && _regex.Match(actualString).Success);
+            string actualString = ConstraintUtils.RequireActual<string>(actual, nameof(actual), allowNull: false);
+
+            return new ConstraintResult(this, actual,_regex.IsMatch(actualString));
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/framework/Constraints/RegexConstraint.cs
```diff
@@ -68,8 +70,9 @@ public RegexConstraint IgnoreCase
         /// <returns>True for success, false for failure.</returns>
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
-            return new ConstraintResult(this, actual,
-                actual is string actualString && _regex.Match(actualString).Success);
+            string actualString = ConstraintUtils.RequireActual<string>(actual, nameof(actual), allowNull: false);
+
+            return new ConstraintResult(this, actual,_regex.IsMatch(actualString));
```
manfred-brands, line 75: Nitpick. There should be a space after a `,`. 

Our .editorconfig shows:  `csharp_space_after_comma = true`
I thought it should have raised some warning, but it doesn't seem to.
Even adding `EnforceCodeStyleInBuild` doesn't seem to do anything.
Only enabling the StyleCop rule does so:
```none
# Commas should be spaced correctly
dotnet_diagnostic.SA1001.severity = warning
```

I will create a future PR to enable this so we get consistent code.

src/NUnitFramework/tests/Constraints/RegexConstraintTests.cs
```diff
@@ -10,6 +10,17 @@ public class RegexConstraintTests
     {
         private static readonly string NL = Environment.NewLine;
 
+        [Test]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Assertion", "NUnit2024", Justification = "Testing this negative case")]
```
manfred-brands, line 14: There seem to be no other `SuppressMessage` calls in the code base.
Please replace it with:
```csharp
#pragma warning disable NUnit2024 // Wrong actual type used with String Constraint
...
#pragma warning restore NUnit2024 // Wrong actual type used with String Constraint
```

src/NUnitFramework/tests/Constraints/RegexConstraintTests.cs
```diff
@@ -10,6 +10,17 @@ public class RegexConstraintTests
...
+        public void RegexTypeMatches()
+        {
+            Assert.Multiple(() =>
+            {
+                Assert.That(() => Assert.That(1, Does.Match("[A-Z]")), Throws.ArgumentException);
+                Assert.That(() => Assert.That((string?)null, Does.Match("[A-Z]")), Throws.ArgumentException);
```
manfred-brands, line 20:
```suggestion
                Assert.That(() => Assert.That(default(string), Does.Match("[A-Z]")), Throws.ArgumentException);
```

manfred-brands (changes requested):
Thanks @RenderMichael Your code changes are functionally fine.
My nitpicks are purely to get a more consistent code base.

RenderMichael:
@manfred-brands thanks for the review! feedback merged, PTAL

manfred-brands (approved):
Thanks for that. Nothing further.

PR #4495, opened 2023-10-12
Consistent code style

### LINKED ISSUE #4494: Enforce consistent code style
Even though we have code style defined in .editorconfig the style is not enforced in the build.
This leads to PR which only need changing because of code style, wasting both contributor and reviewer time.

manfred-brands:
Fixes #4494 

This PR enabled Style rules, on commit per rule.
Some rules had no associated changes to the code, other did, but in all not much.

Main changes are in code which uses `if (condition) statement` on a single line.
Personally I don't find those readable.


SeanKilleen (approved):
👍 nice!

As an aside, do we have "format on save" turned on, and will it respect these rules? Or do we have to make an editor config match up with them to achieve that?

PR #4499, opened 2023-10-13
Mark TimeoutAttribute Obsolete for non .NET Framework builds.

### LINKED ISSUE #4297: Empty <failure /> element on <test-case /> failed due to [Timeout]
nunit.framework: 3.13.3

In version 3.12.0 you made it possible to use [Timeout] attribute in tests targeting .NETCore. However, reporting of an error caused by this attribute was and still is different from when the same error occurs in tests targeting .NETFramework:

Stripped of irrelevant for this issue attributes, the reports look as follows:

In `net7.0`
```
<test-case id="0-1001" 
           fullname="TestProject1.Tests.Test000" 
           runstate="Runnable" 
           result="Failed" 
           asserts="0"
           label="Test exceeded Timeout value 1000ms.">
  <properties>
    <property name="Timeout" value="1000" />
  </properties>
  <failure />
</test-case> 

```

In `net48`:
```
<test-case id="0-1001"
           fullname="TestProject1.Tests.Test000"
           runstate="Runnable" 
           result="Failed"
           asserts="0">
  <properties>
    <property name="Timeout" value="1000" />
  </properties>
  <failure>
    <message><![CDATA[Test exceeded Timeout value of 1000ms]]></message>
  </failure>
</test-case> 
```

Note how `net7.0` has an empty <failure /> element and a new label attribute not seen in `net48` variant.

Would you consider unifying how the failure is reported across different target frameworks?

manfred-brands:
Fixes #4297 

Align implementation/tests as much as possible.

Also marks the TimeoutAttribute obsolete for non-framework builds.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TimeoutAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TimeoutAttribute.cs
@@ -12,6 +12,9 @@ namespace NUnit.Framework
     /// When applied to a class or assembly, the default timeout is set for all contained test methods.
     /// </summary>
     [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Assembly, AllowMultiple = false, Inherited = false)]
+#if !NETFRAMEWORK
+    [Obsolete(".NET No longer supports aborting threads as it is not a safe thing to do. Update your tests to use CancelAfterAttribute instead")]
+#endif
     public class TimeoutAttribute : PropertyAttribute, IApplyToContext
     {
         private readonly int _timeout;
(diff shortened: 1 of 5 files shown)


OsirisTerje (approved):
LGTM

stevenaw (approved):
LGTM too thanks @manfred-brands

gedefran:
@manfred-brands this commit seems to break TestContext in my teardown methods, they are now getting the wrong results object. Pretty sure it's because of your change in TimeoutCommand.cs creating its own test results, and then it gets blindly overwritten in the call to Execute. Let me know if that makes sence of if I should create an issue.

stevenaw:
@gedefran Would you be able to file a new issue please? It would help us if you could provide a small repro of your issue as well as information on which version of nunit you are using and which runtime

gedefran:
No problem, will get to it tonight.

PR #4507, opened 2023-10-15
Set Tolerance as not nullable

### LINKED ISSUE #4506: Should Tolerance be a record type?
1) The Tolerance class is not something that is changed after creation.
2) It should not be null.

The code here in the Equal Result constraint causes a warning, and the parameter is `Tolerance?` but it is initialized as Default, and possibly later set to another Tolerance (through new), so it can't be null.
image

It doesn't seem to change at all, so could it be a record type?

@manfred-brands @stevenaw  Comments?

Are there other classes that are better suited as records?

OsirisTerje:
Fixes #4506 
Changed Tolerance to record type
Fixed nullability

### DIFF
--- a/src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
@@ -73,7 +73,7 @@ namespace NUnit.Framework.Constraints
                 DisplayEnumerableDifferences(writer, expectedEnumerable, actualEnumerable, depth);
             else if (expected is Stream expectedStream && actual is Stream actualStream)
                 DisplayStreamDifferences(writer, expectedStream, actualStream, depth);
-            else if (_tolerance is not null)
+            else if (_tolerance != Tolerance.Default)
                 writer.DisplayDifferences(expected, actual, _tolerance);
             else
                 writer.DisplayDifferences(expected, actual);
--- a/src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
+++ b/src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
@@ -73,10 +73,8 @@ namespace NUnit.Framework.Constraints
                 DisplayEnumerableDifferences(writer, expectedEnumerable, actualEnumerable, depth);
             else if (expected is Stream expectedStream && actual is Stream actualStream)
                 DisplayStreamDifferences(writer, expectedStream, actualStream, depth);
-            else if (_tolerance != Tolerance.Default)
-                writer.DisplayDifferences(expected, actual, _tolerance);
             else
-                writer.DisplayDifferences(expected, actual);
+                writer.DisplayDifferences(expected, actual, _tolerance);
         }
 
         #region DisplayStringDifferences
--- a/src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
@@ -130,19 +130,6 @@ namespace NUnit.Framework.Internal
         }
 
         /// <summary>
-        /// Display Expected and Actual lines for given values. This
-        /// method may be called by constraints that need more control over
-        /// the display of actual and expected values than is provided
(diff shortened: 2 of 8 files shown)


src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
```diff
@@ -73,7 +73,7 @@ private void DisplayDifferences(MessageWriter writer, object? expected, object?
                 DisplayEnumerableDifferences(writer, expectedEnumerable, actualEnumerable, depth);
             else if (expected is Stream expectedStream && actual is Stream actualStream)
                 DisplayStreamDifferences(writer, expectedStream, actualStream, depth);
-            else if (_tolerance is not null)
+            else if (_tolerance != Tolerance.Default)
```
manfred-brands, line 76: There is no longer a need for this test and you can simply call:
`writer.DisplayDifferences(expected, actual, _tolerance);` unconditionally.

src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
```diff
@@ -139,7 +139,7 @@ private void ResolveTypeNameDifference(object expected, object actual, out strin
         /// <param name="actual">The actual value causing the failure</param>
         public override void DisplayDifferences(object? expected, object? actual)
         {
-            DisplayDifferences(expected, actual, null);
+            DisplayDifferences(expected, actual, Tolerance.Default);
```
manfred-brands, line 142: With the above change, this method can be removed.

src/NUnitFramework/tests/Constraints/ToleranceTests.cs
```diff
@@ -111,5 +111,16 @@ public void ToStringTests()
                 Assert.That(Is.EqualTo(5).Within(2).Minutes.Tolerance.ToString(), Is.EqualTo("00:02:00"));
             });
         }
+
+        [Test]
+        public void EqualityTests()
+        {
+            var tol1 = Tolerance.Default;
+            var tol2 = Tolerance.Default;
+            Assert.That(tol1, Is.EqualTo(tol2));
```
manfred-brands, line 120: This could be an `Is.SameAs` test as there is only one `Default` and there is no way a user can create another instance as that constructor is `private`

manfred-brands (changes requested):
Some suggested simplifications

src/NUnitFramework/tests/Constraints/ToleranceTests.cs
manfred-brands: And that is already tested, see `TestToleranceDefaultIsSameAs`, so this test can be removed.

src/NUnitFramework/tests/Constraints/ToleranceTests.cs
```diff
@@ -3,124 +3,112 @@
 using System;
 using NUnit.Framework.Constraints;
 
-namespace NUnit.Framework.Tests.Constraints
+namespace NUnit.Framework.Tests.Constraints;
```
manfred-brands, line 6: I'm not in favour of changing to file scoped namespace as part of a normal PR.
We have always said that layout changes should be done as separate PRs.

I don't mind changing globally if that is what we want. I like consistency and if we move it should be a separate PR with a rule enforcing this for future developments.

OsirisTerje: Here I actually don't agree with you.  The "Boy Scout Rule",  clean up as you go - leave the camp site in a better shape before you leave, is a better way, imho.  

Globally doing large scale code changes just for pretty-ness have been done the last couple of years, making it now very hard to see  who actually did a real functional change, git blame is now worthless. 

Small layout changes that comes as part of the suggestions should be ok to do, that be spaces, braces or similar small stuff.  File scoped namespaces is to me just a removal of braces. 

Also, we don't need to fix all the guidelines by ourselves, imho we can just "follow the flow", and accept the changes ind defaults that comes as time goes by.

Refactoring code however, should be done separately, because it just may have some other consequences.

PR #4515, opened 2023-10-21
Unwrap TargetInvocationException

### LINKED ISSUE #4481: `Does.ContainKey()` throws an unintuitive exception for a null key
```csharp
Assert.That(new Dictionary<string, string>(), Does.ContainKey(null));
```
This throws the following exception
```
 System.Reflection.TargetInvocationException : Exception has been thrown by the target of an invocation.
  ----> System.ArgumentNullException : Value cannot be null. (Parameter 'key')
```

It would be nice to unwrap this exception in here:
https://github.com/nunit/nunit/blob/32b83e7ca8b9f96bab09d319fbc0a3a0da166601/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyConstraint.cs#L85-L88

to 
```csharp
if (method is not null)
{
    try
    {
        return (bool)method.Invoke(actual, new[] { Expected })!;
    }
    catch (TargetInvocationException ex)
    {
        ExceptionDispatchInfo.Throw(ex.InnerException!);
    }
}
```

manfred-brands:
Fixes #4481 

Code for 
```cshap
Assert.That(new Dictionary<string, string>(), Does.ContainKey(null!));
```

Now results in:
```
  Message: 
    System.ArgumentNullException : Value cannot be null. (Parameter 'key')

  Stack Trace: 
    Dictionary`2.FindValue(TKey key)
    Dictionary`2.ContainsKey(TKey key)
```

There are two methods in Reflect.cs ending in `WithTransparentExceptions`.
These are used in `Assert.Throws` and equivalents and were not modified to _unwrap_. 
Unwrapping all `InnerException`s would make it impossible to catch a `TargetInvocationException` thrown by user code.

This means that to validate that the `Does.ContainKey(null!)` throws an `ArgumentNullException` we
first have to test for the `TargetInvocationException`.
Beside testing NUnit, not many people will `Assert` an `Assert`.

```csharp
Assert.That(() => Assert.That(new Dictionary<string, string>(), Does.ContainKey(null!)),
            Throws.TargetInvocationException.With.InnerException.InstanceOf<ArgumentNullException>());
```

### DIFF
--- a/src/NUnitFramework/framework/Internal/Builders/DefaultSuiteBuilder.cs
+++ b/src/NUnitFramework/framework/Internal/Builders/DefaultSuiteBuilder.cs
@@ -91,9 +91,7 @@ namespace NUnit.Framework.Internal.Builders
             }
             catch (Exception ex)
             {
-                if (ex is System.Reflection.TargetInvocationException)
-                    ex = ex.InnerException!;
-                var fixture = new TestFixture(typeInfo, ex);
+                var fixture = new TestFixture(typeInfo, ex.Unwrap());
 
                 return fixture;
             }
--- a/src/NUnitFramework/framework/Internal/ExceptionHelper.cs
+++ b/src/NUnitFramework/framework/Internal/ExceptionHelper.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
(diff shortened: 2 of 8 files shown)


OsirisTerje (approved):
I believe this looks good now.

OsirisTerje:
Feel free to merge

PR #4516, opened 2023-10-21
Tolerance is not supported for IEquatable types

### LINKED ISSUE #4514: If Tolerance is used a different comparer is used
As a result of the changes for #4436 and #4312, we are now in a situation that if a type implements `IEquatable<T>` and no `Tolerance` is specified, the type's `Equal(T)` method is called.

Howevever, if `Tolerance` is specified, #4312 would throw an `InvalidOperationException`
After the change #4436, the `PropertiesComparer.Equals` method is called which calls `Equals` on all properties, passing in the specified `Tolerance`.

I wanted to align NUnit.Analyzer with the new `PropertiesComparer`, relaxing the {_Within does not make sense_ rule](https://github.com/nunit/nunit.analyzers/issues/617), when I stumbled over this odd behaviour.

What behaviour do we want if a type implements `IEquatable<T>` (or overrides `Equals(object)`) and `Tolerance` is specified:
  1.  `throw InvalidOperationException`?
  2. Fallback to `PropertiesComparer` changing the comparison used and using the `Tolerance`

I'm leaning to 1. Which requires a minor change to NUnit4.

manfred-brands:
Fixes #4514 

It previously fell back to property comparer if tolerance was used not even calling the type's defined comparison method.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/ArraysComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/ArraysComparer.cs
@@ -9,10 +9,10 @@ namespace NUnit.Framework.Constraints.Comparers
     /// </summary>
     internal static class ArraysComparer
     {
-        public static bool? Equal(object x, object y, ref Tolerance tolerance, ComparisonState state, NUnitEqualityComparer equalityComparer)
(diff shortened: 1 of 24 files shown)


manfred-brands:
This PR also make treatment of Tolerance more consistent.

It will complain about 'string', 'char', 'enum' and 'IEquatable'
For Tuples and user defined types, if one member supports Tolerance,
the whole Tuple/Class/Struct does.

OsirisTerje (approved):

PR #4519, opened 2023-10-27
Allow null parameters to Contains

### LINKED ISSUE #4518: NUnit4: Inconsistent nullability for Contains
When compiling my work's big project, I get nullability compile errors, due to some overloads not accepting `null`.

We have: `Does.Contain(object?)` but `Does.Contain(string)` and `ConstraintExpression.Contains(object)`

This mean one can code: `Does.Contain(null)` but not `Does.Not.Contain(null)` nor `Does.Contain(default(string)`.

This was introduced by me when adding nullability.
As we don't know if the collection at hand is nullable we must allow `null`.
A possible NUnit.Analyzer rule could check if the collection is declared to have a non-nullable member type.

manfred-brands:
Fixes #4518


mikkelbu (approved):
LGTM. Thanks for the report and the fix @manfred-brands

Issue #3043, opened 2018-10-05
AsyncEnumerable support for TestCaseSource, ValueSource, and TestFixtureSource


jnm2:
The .NET Core API proposal for IAsyncDisposable and IAsyncEnumerable is up:
`https://github.com/dotnet/corefx/issues/32640`

The matching C# 8.0 features `foreach async` and `using async` and async iterators will land in a few months.

- Right now we call IDisposable.Dispose on fixtures. Would we want to recognize and await `IAsyncDisposable.DisposeAsync` on fixtures?

- Would we want `Assert.That(asyncSequence, Is.EquivalentTo(expectedAsyncSequence))` to light up and just work? I lean towards thinking we should require `await AsyncAssert.That(` (once we add those APIs) in order to light this one up since it's not an entry point into user code from the framework the way Dispose/DisposeAsync is, and adapting async to sync is an ugly thing for user code work with.

The C# 8 language feature is shape-based, just like the existing async and foreach pattern recognition since C# 5 and 1. We can implement this using reflection without depending on .NET Core or C# 8.

stevenaw:
@nunit/framework-team @nunit/core-team 
Any objections to if I were to try to complete this for v4?

manfred-brands:
None here

stevenaw:
Yep, that was my thinking @manfred-brands . Disposable first as it's likely simpler and applicable to more types.

Curious, @nunit/framework-team , how do we feel about taking on a dependency on `Microsoft.Bcl.AsyncInterfaces` on the framework netfx build to support this? The original issue description described a shape-based solution but this issue is also long-standing enough that it predates that package's existence.

Our own tests here would also likely be simplified by taking on this dependency as well, but that can happen independent of any framework changes.

manfred-brands:
At my work we added that dependency on multitargeting projects with no issues.

jnm2:
It will mean that people who declare their own IAsyncDisposable interface in the system names pace (or use an alternate package) will not be supported. That could be fine.

stevenaw:
Thanks @jnm2 !
Shape -based is fine with me too, and sounds like there could be a compatibility case for it too.

stevenaw:
@nunit/framework-team 
Looking for confirmation of a design decision or two.

I assume that in the case that a class has both a Dispose() and DisposeAsync() method that we should invoke only the async one and rely on the user to delegate any sync disposal internally in the DisposeAsync() implementation?
[...]

PR #4530, opened 2023-10-29
Async disposal of test fixtures

stevenaw:
Contributes to #3043 

Adds support for async disposal of test fixtures

@nunit/framework-team This has been a "20 minutes at a time" sort of change over the last several weeks. I welcome a closer scrutiny from the team over any design or implementation decisions here

### DIFF
--- a/src/NUnitFramework/tests/Attributes/OneTimeSetUpTearDownTests.cs
+++ b/src/NUnitFramework/tests/Attributes/OneTimeSetUpTearDownTests.cs
@@ -378,8 +378,11 @@ namespace NUnit.Framework.Tests.Attributes
         {
             var fixture = new AsyncDisposableFixture();
             TestBuilder.RunTestFixture(fixture);
+
+            var expected = new[] { nameof(AsyncDisposableFixture.OneTimeSetUp), nameof(AsyncDisposableFixture.OneTimeTearDown), nameof(AsyncDisposableFixture.DisposeAsync) };
+
             Assert.That(fixture.DisposeCalled, Is.EqualTo(1));
-            Assert.That(fixture.Actions, Is.EqualTo(new object[] { "OneTimeSetUp", "OneTimeTearDown", "DisposeAsync" }));
+            Assert.That(fixture.Actions, Is.EqualTo(expected));
         }
 
         [Test]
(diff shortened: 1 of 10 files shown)


src/NUnitFramework/tests/Attributes/OneTimeSetUpTearDownTests.cs
```diff
@@ -372,6 +372,34 @@ public void DisposeCalledOnceWhenFixtureImplementsIDisposableAndHasTestCases()
...
             Assert.That(fixture.DisposeCalled, Is.EqualTo(1));
         }
+
+        [Test]
+        public void AsyncDisposeCalledOnceWhenFixtureImplementsIAsyncDisposable()
+        {
+            var fixture = new AsyncDisposableFixture();
+            TestBuilder.RunTestFixture(fixture);
+            Assert.That(fixture.DisposeCalled, Is.EqualTo(1));
+            Assert.That(fixture.Actions, Is.EqualTo(new object[] { "OneTimeSetUp", "OneTimeTearDown", "DisposeAsync" }));
```
manfred-brands, line 382: I'm not a fan of hard-coded strings. They are brittle in case they are renamed in the fixture called.
My preference would be `nameof(AsyncDisposableFixture.OneTimeSetup)` etc.
I know these strings are all over nunit test code, but maybe we cannot introduce more.

src/NUnitFramework/tests/Attributes/OneTimeSetUpTearDownTests.cs
```diff
@@ -372,6 +372,34 @@ public void DisposeCalledOnceWhenFixtureImplementsIDisposableAndHasTestCases()
...
+        }
+
+        [Test]
+        public void AsyncDisposeCalledOnceWhenFixtureImplementsIAsyncDisposableThroughInheritance()
+        {
+            var fixture = new InheritedAsyncDisposableFixture();
+            TestBuilder.RunTestFixture(fixture);
+            Assert.That(fixture.DisposeCalled, Is.EqualTo(1));
+            Assert.That(fixture.Actions, Is.EqualTo(new object[] { "OneTimeSetUp", "OneTimeTearDown", "DisposeAsync" }));
```
manfred-brands, line 391: Suggest to use `nameof`

src/NUnitFramework/tests/Attributes/OneTimeSetUpTearDownTests.cs
```diff
@@ -372,6 +372,34 @@ public void DisposeCalledOnceWhenFixtureImplementsIDisposableAndHasTestCases()
...
+        }
+
+        [Test]
+        public void AsyncDisposePrioritizedWhenSyncAndAsyncDispose()
+        {
+            var fixture = new AsyncAndSyncDisposableFixture();
+            TestBuilder.RunTestFixture(fixture);
+            Assert.That(fixture.DisposeCalled, Is.EqualTo(1));
+            Assert.That(fixture.Actions, Does.Contain("DisposeAsync"));
+            Assert.That(fixture.Actions, Does.Not.Contain("Dispose"));
```
manfred-brands, line 401: Suggest to use `nameof`

manfred-brands (approved):
@stevenaw I think it looks good. All Disposing is dealt with with a set of helpers, instead of spread out over the code.

I have some small nitpicks, because I'm pedantic.

stevenaw:
> @stevenaw I think it looks good. All Disposing is dealt with with a set of helpers, instead of spread out over the code.
> 
> I have some small nitpicks, because I'm pedantic.

Thanks @manfred-brands appreciate the feedback! I think it's good to go now

PR #4533, opened 2023-10-31
Allow null string message to be passed to Assert.That

### LINKED ISSUE #4532: Assert.That no longer allows 'null' message
I encountered NUnit extension methods that have a signature with: `string? message = null` which eventually calls into some version of `Assert.That(actual, constaint, message)`. These no longer compile.

manfred-brands:
Fixes #4532

### DIFF
--- a/src/NUnitFramework/framework/NUnitString.cs
+++ b/src/NUnitFramework/framework/NUnitString.cs
@@ -21,7 +21,7 @@ namespace NUnit.Framework
         /// Initializes a new instance of the <see cref="NUnitString"/> class.
         /// </summary>
         /// <param name="message">The message formattable to hold.</param>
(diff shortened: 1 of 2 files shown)


stevenaw (approved):
LGTM. Good catch @manfred-brands

manfred-brands:
> LGTM. Good catch @manfred-brands

I only noticed it because some my work's NUnit extensions didn't compile anymore with version 4.

OsirisTerje:
@manfred-brands Awesome!  Very good that you found this before real release.  The more we manage to test the beta, the more likely it is we wont have any hotfixes.

OsirisTerje:
@manfred-brands Could be interesting if you wrote up your experiences, could be useful for creating a migration guide.

PR #4538, opened 2023-11-06
AsyncEnumerable support for TestCaseSource, ValueSource, and TestFixtureSource

### LINKED ISSUE #3391: TestCaseSource to recognize new async streams
Would be nice to have support new .net core 3.0 async streams feature.

```c#
[Test, TestCaseSource(nameof(EnumerateAsync))]
public void AsyncStreamTest(object val) {
    Assert.IsNotNull(val);
}

public static async IAsyncEnumerable<object> EnumerateAsync()  {
    yield return await Create();
    yield return await Create();
}

public static Task<object> Create() {
    return Task.FromResult(new object());
}
```

Currently it says:
```
  X AsyncStreamTest [< 1ms]
  Error Message:
   System.InvalidCastException : Unable to cast object of type '<EnumerateAsync>d__1' to type 'System.Collections.IEnumerable'.
  Stack Trace:
     at NUnit.Framework.TestCaseSourceAttribute.GetTestCaseSource(IMethodInfo method)
   at NUnit.Framework.TestCaseSourceAttribute.GetTestCasesFor(IMethodInfo method)
```

stevenaw:
Fixes #3391 
Relates to #3043 

TODO:
- [x] TestCaseSource
- [x] ValueSource
- [x] TestFixtureSource
- [x] Make an issue to relax the analyzer (https://github.com/nunit/nunit.analyzers/issues/634)
- [x] Make an issue for the docs repo (https://github.com/nunit/docs/issues/834)

Preliminary questions:
- Should we cache the shape info on netfx side? We do this for awaitables because we expect a low number of different types with a potentially high number of invocations each. Conversely, the async disposal of TestFixtures doesn't cache it because it's likely to be a much lower ratio there.
- How thorough should error reporting be on shape-based checks (do we expect someone to try to implement a custom IAsyncEnumerable or IAsyncEnumerator outside of `Microsoft.Bcl.AsyncInterfaces` and to do it in a way that differs in some way from the standard? I'm particularly thinking about either return types or how DisposeAsync requires a separate look-up from the IAsyncDisposable interface from which IAsyncEnumerator extends.


manfred-brands:
> Should we cache the shape info on netfx side

Try initially without, then check to see if it is a bottleneck. Don't optimize prematurely.

> do we expect someone to try to implement a custom IAsyncEnumerable or IAsyncEnumerator outside of Microsoft.Bcl.AsyncInterfaces and to do it in a way that differs in some way from the standard

An interface defines a standard behaviour not just signature. If their `Enumerator` doesn't enumerate, then they likely have other problems.

stevenaw:
> Try initially without, then check to see if it is a bottleneck. Don't optimize prematurely.

Works for me. FWIW I'd expect a relatively small number of types for this case but it's always something we can adjust later, as you say.

> An interface defines a standard behaviour not just signature. If their Enumerator doesn't enumerate, then they likely have other problems.

Yeah. Initially I'd been thinking about what if a custom shim were to flatten the inheritance hierarchy so that DisposeAsync() was in the same interface type as the IAsyncEnumerator<T> members such as an `IDisposableAsyncEnumerator<T>` type, but I suspect that could cause other issues in such a case anyways. Also happy to keep it simple here too and code against and support the standard interfaces.

stevenaw:
My latest changes appear to have hit a slow mac VM, but they're otherwise ready.

> 3-2) Failed : NUnit.Framework.Tests.Constraints.CollectionEquivalentConstraintTests.LargeStringCollectionsInSameOrder
2023-11-08T02:45:21.1495100Z LargeStringCollectionsInSameOrder took 650 ms.
2023-11-08T02:45:21.1496500Z   at NUnit.Framework.Tests.Constraints.CollectionEquivalentConstraintTests.LargeStringCollectionsInSameOrder () [0x00000] in <1d4d35f5ec3f47379ae62d255b6c312c>:0

src/NUnitFramework/framework/Internal/AsyncEnumerableAdapter.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    internal static partial class AsyncEnumerableAdapter
+    {
+        public static IEnumerable? CoalesceToEnumerable(object? enumerable)
+        {
+            if (enumerable is null)
+            {
+                return null;
+            }
+            else if (TryGetAsyncBlockingEnumerable(enumerable, out var result))
```
manfred-brands, line 17: As `IEnumerable` is more likely, I would check that first. Especially as the .NET Framework version of this method uses reflection.

The code now relies on a runtime InvalidCastException if the method at hand doesn't return an `IEnumerable` at all.
Maybe you should make that more explicit?

src/NUnitFramework/framework/Internal/AsyncEnumerableAdapter.NetCore.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            if (enumerable is IAsyncEnumerable<object> asyncEnumerable)
+            {
+                // Allow for lazily enumeration
+                result = new AsyncEnumerableWrapper(asyncEnumerable);
+                return true;
+            }
+
+            result = default;
+            return default;
```
manfred-brands, line 22:
```suggestion
            return false;
```

It might be the same value, but reads more logical.

src/NUnitFramework/framework/Internal/AsyncEnumerableAdapter.NetCore.cs
```diff
@@ -0,0 +1,63 @@
...
+                _asyncEnumerator = asyncEnumerator;
+            }
+
+            public object? Current => _asyncEnumerator.Current;
+
+            public void Dispose()
+                => AsyncToSyncAdapter.Await(() => _asyncEnumerator.DisposeAsync());
+
+            public bool MoveNext()
+                => AsyncToSyncAdapter.Await<bool>(() => _asyncEnumerator.MoveNextAsync());
```
manfred-brands, line 56: This is an expensive wrap to use for every element, but I don't see a better easy option. 
Unless we say NUnit should do proper async iso AsyncOverSync. That that might have to wait til NUnit 5.x

src/NUnitFramework/framework/Internal/AsyncEnumerableAdapter.NetFramework.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            var moveNextMethod = enumeratorType.GetMethod("MoveNextAsync");
+            if (moveNextMethod is null)
+                return false;
+
+            var asyncDisposableType = enumeratorType.GetInterface("System.IAsyncDisposable");
+            if (asyncDisposableType is null)
+                return false;
+
+            var asyncDisposeMethod = asyncDisposableType!.GetMethod("DisposeAsync");
```
manfred-brands, line 40:
```suggestion
            var asyncDisposeMethod = asyncDisposableType.GetMethod("DisposeAsync");
```

No need for the null suppression operator.

src/NUnitFramework/framework/Internal/AsyncEnumerableAdapter.NetFramework.cs
```diff
@@ -0,0 +1,112 @@
...
+                _asyncEnumerator = asyncEnumerator;
+            }
+
+            public object? Current => _shape.CurrentProperty.GetValue(_asyncEnumerator);
+
+            public void Dispose()
+                => AsyncToSyncAdapter.Await(() => _shape.DisposeAsyncMethod.Invoke(_asyncEnumerator, null));
+
+            public bool MoveNext()
+                => AsyncToSyncAdapter.Await<bool>(() => _shape.MoveNextAsyncMethod.Invoke(_asyncEnumerator, null));
```
manfred-brands, line 89: MethodInfo.Invoke is sloow. As the signature is known, we could use delegates:
```
Job=.NET Framework 4.6.2  Runtime=.NET Framework 4.6.2

| Method               | Mean       | Ratio | Allocated | Alloc Ratio |
|--------------------- |-----------:|------:|----------:|------------:|
| DirectMethodCall     |   1.145 ns |  0.01 |         - |        0.00 |
| MethodInfoInvoke     | 102.406 ns |  1.00 |     112 B |        1.00 |
| CreateDelegateInvoke |   2.052 ns |  0.02 |         - |        0.00 |
```

Something like:

```csharp
private delegate ValueTask<bool> MoveNextAsyncDelegate(object @this);

...
MoveNextAsync = (MoveNextAsyncDelegate)moveNextAsyncMethod.CreateDelegate(typeof(MoveNextAsyncDelegate));
...

=> AsyncToSyncAdapter.Await<bool>(() => _shape.MoveNextAsync(_asyncEnumerator));
```

Although all this is probably a mute point as long as we need to use the `AsyncToSyncAdapter.Await`

manfred-brands (changes requested):
A few small changes and some big questions for some other time.

src/NUnitFramework/framework/Internal/AsyncEnumerableAdapter.cs
stevenaw: > As IEnumerable is more likely, I would check that first.

Thanks, will do. The current ordering was a conscious decision, though perhaps misguided, based on the discussion in the AsyncDispose PR. I think it's unlikely a type will implement both interfaces, in particular based on that I can't even find (m)any concrete implementations in the BCL - they're predominantly compiler-generated via `yield return` in an async method.

stevenaw: > The code now relies on a runtime InvalidCastException if the method at hand doesn't return an IEnumerable at all.
Maybe you should make that more explicit?

I actually meant to do this but had forgotten! Thanks

src/NUnitFramework/framework/Internal/AsyncEnumerableAdapter.NetFramework.cs
stevenaw: I like both of these ideas, thanks, though I agree top-level async may need to wait 😄 

I think I'd mistaken the overhead of `CreateDelegate()` with the overhead of using `Reflection.Emit`-based codegen but I can see that it's comparatively comparatively quite cheap even on [netfx](https://referencesource.microsoft.com/#mscorlib/system/delegate.cs,715).

stevenaw: I have a few ideas here that'll take a bit to experiment with. I think changes for the rest of the suggestions have been pushed

manfred-brands (approved):
Thanks, nothing else.

stevenaw:
Thanks @manfred-brands . I've pushed the last of my planned changes here - in particular CreateDelegate usage for `Current` and `MoveNextAsync()` since they'll be invoked more than once against a given instance.

Something I had thought about exploring but ultimately didn't is around caching. Given that most `IAsyncEnumerable<T>` implementations appear to be likely compiler-generated, I _think_ it means it would be a different type per method, however given the reflection look-ups are against the interface itself there may be grounds for caching _those_. Something for us to consider later.

vjred:
@stevenaw , @manfred-brands This seems to have regressed in `4.2.0` or `4.2.1` for `TestCaseSource` attribute. We are getting `NotSupportedException` when using `IAsyncEnumerable` for `TestCaseSourceAttribute`. Its failing on [this line](https://github.com/nunit/nunit/blob/main/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs#L152). Could you please confirm and fix if possible?

stevenaw:
@vjred It's tough for us to triage the issue in this context. Could you please file a new issue and include a code snippet which can reproduce the issue you see? The snippet should ideally be as small as possible and run as-is.

vjred:
> @vjred It's tough for us to triage the issue in this context. Could you please file a new issue and include a code snippet which can reproduce the issue you see? The snippet should ideally be as small as possible and run as-is.

Thanks @stevenaw, I have created #4802 with the steps to reproduce.

PR #4540, opened 2023-11-06
Updated cake to version 3.1.0 and corrected breaking changes cake code

### LINKED ISSUE #4465: Update cake version in build scripts
There's a few draft PRs in progress right now that have had to upgrade cake in the build scripts as part of other changes. It might be nice to upgrade cake separately as part of this issue to make those and other future updates simpler

OsirisTerje:
While correcting the script after the update, I noticed that we do have some tasks that seems to not be used,  Publish Test results is one - that one seems to be oriented to Azure Pipelines.  And then SignPackages.  This one I believed we should have there, but I guess we might need to update it, and also ensure we have a valid certificate.
On the other hand, nuget does an automatic repository signing for the packages. So, do we need more?

Fixes #4465

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - master
       - release
-      - 'v3.13-dev'
+      - 'v3'
   pull_request:
 
 jobs:
--- a/build.cake
+++ b/build.cake
@@ -228,7 +228,7 @@ Task("CreateImage")
 
         var imageBinDir = Directory(CurrentImageDir) + Directory("bin");
 
-        CreateDirectory(imageBinDir.ToString());
+        CreateDirectory(imageBinDir);
         Information("Created imagedirectory at:" + imageBinDir.ToString());
         var directories = new String[]
         {
--- a/build.cake
+++ b/build.cake
@@ -1,5 +1,3 @@
-#tool NUnit.ConsoleRunner&version=3.12.0
-
 //////////////////////////////////////////////////////////////////////
 // ARGUMENTS
 //////////////////////////////////////////////////////////////////////
@@ -20,7 +18,7 @@ var ErrorDetail = new List<string>();
 var version = "4.0.0";
 var modifier = "-beta.1";
 
-var dbgSuffix = configuration == "Debug" ? "-dbg" : "";
+var dbgSuffix = configuration.ToLower() == "debug" ? "-dbg" : "";
 var packageVersion = version + modifier + dbgSuffix;
 
 //////////////////////////////////////////////////////////////////////
@@ -97,7 +95,7 @@ Task("NuGetRestore")
     .Description("Restores NuGet Packages")
     .Does(() =>
     {
-        DotNetCoreRestore(SOLUTION_FILE);
+        DotNetRestore(SOLUTION_FILE);
     });
 
 //////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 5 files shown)


.github/workflows/continuous_integration.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches:
       - master
       - release
-      - 'v3.13-dev'
+      - 'v3'
```
manfred-brands, line 8: Our `release` branch is pointing to `v3.13.2`. Should that be moved or deleted?

build.cake
```diff
@@ -109,15 +107,15 @@ Task("Build")
...
     .Does(() =>
     {
-        DotNetCoreBuild(SOLUTION_FILE, CreateDotNetCoreBuildSettings());
+        DotNetBuild(SOLUTION_FILE, CreateDotNetCoreBuildSettings());
     });
 
-DotNetCoreBuildSettings CreateDotNetCoreBuildSettings() =>
-    new DotNetCoreBuildSettings
+DotNetBuildSettings CreateDotNetCoreBuildSettings() =>
+    new DotNetBuildSettings
```
manfred-brands, line 114: For consistency, drop `Core` from the method name as well.

build.cake
```diff
@@ -109,15 +107,15 @@ Task("Build")
...
     {
         Configuration = configuration,
         NoRestore = true,
-        Verbosity = DotNetCoreVerbosity.Minimal
+        Verbosity = DotNetVerbosity.Minimal
```
manfred-brands, line 118: I had added the below to ensure the version is part of the dll.
```cake
MSBuildSettings = new DotNetMSBuildSettings
{
    Version = packageVersion
}
```

This was previously done on AppVeyor builds only, but that was removed.

src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintTests.cs
```diff
@@ -287,8 +287,10 @@ public void WorksWithNonIComparableTuples()
 
         // The following tests are each running in 14ms to 46ms on my machine. Based on that,
         // warn at 100ms and fail at 500ms
-        private const int LARGE_COLLECTION_WARN_TIME = 100;
-        private const int LARGE_COLLECTION_FAIL_TIME = 500;
+        // Seems to be slower on MacOs on build on github actions, so increasing this with 50%
+
+        private const int LARGE_COLLECTION_WARN_TIME = 100 * 3 / 2;
+        private const int LARGE_COLLECTION_FAIL_TIME = 500 * 3 / 2;
```
manfred-brands, line 293: If only MaCOS is the problem, don't run them there:
`[Platform(Exclude = "MACOSX", Reason = "MacOS github runner is very slow")]`

manfred-brands (changes requested):
Some suggestions.

.github/workflows/continuous_integration.yml
OsirisTerje: The release branch is not in direct parent path behind v3, so I would prefer to just skip it.

OsirisTerje: Deleted the rule now, and also deleted the branch itself

build.cake
OsirisTerje: Done

build.cake
OsirisTerje: Done

src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintTests.cs
OsirisTerje: Done

build.cake
OsirisTerje: But it has a property of that type, so moved there

manfred-brands (approved):
Thanks. Nothing further from me.

stevenaw (approved):
Thanks @OsirisTerje , especially for the added macos fix. Looks good.

PR #4541, opened 2023-11-07
Issue4539 upgrade n unit analyzers

### LINKED ISSUE #4539: Upgrade to latest analyzers package and fix the warnings
We've recently released v3.9.0 of the analyzers. The framework repo still targets 3.6.1
I tested an upgrade recently and found a few things getting flagged which should be fixed

manfred-brands:
Fixes #4539 

I have enabled central package management so that all projects always will use the same version of packages.
It ensure when upgrading we only need to update the version in one place.

### DIFF
--- a/src/NUnitFramework/testdata/SetUpFixtureData.cs
+++ b/src/NUnitFramework/testdata/SetUpFixtureData.cs
@@ -68,7 +68,7 @@ namespace NUnit.Framework.Tests.TestUtilities
             /// </summary>
             public void Verify()
             {
-                EventMatcher eventMatcher = _eventMatchers.Dequeue();
+                EventMatcher? eventMatcher = _eventMatchers.Dequeue();
                 int item = 0;
 
                 foreach (string actualEvent in _actualEvents)
@@ -77,8 +77,7 @@ namespace NUnit.Framework.Tests.TestUtilities
                     {
                         Assert.Fail($"More events than expected were recorded. Current event: {actualEvent} (Item {item})");
                     }
-
-                    if (!eventMatcher.MatchEvent(actualEvent, item++))
+                    else if (!eventMatcher.MatchEvent(actualEvent, item++))
                     {
                         eventMatcher = _eventMatchers.Count > 0 ? _eventMatchers.Dequeue() : null;
                     }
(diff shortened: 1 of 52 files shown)


mikkelbu (commented):
I'm halfway through the review and I'll review the rest in a couple of hours. So far all look great 👍

src/NUnitFramework/testdata/SetUpFixtureData.cs
```diff
@@ -107,10 +106,10 @@ public static void RegisterEvent(string evnt)
         /// <param name="expectedEvents">The expected events.</param>
         public static void Verify(params string[] expectedEvents)
         {
+            int item = 0;
```
mikkelbu, line 109: Nice find

src/NUnitFramework/tests/Constraints/MsgUtilTests.cs
```diff
@@ -124,23 +124,19 @@ public static void FormatValue_CharTest(char c, string expected)
             Assert.That(MsgUtils.FormatValue(c), Is.EqualTo(expected));
         }
 
-        [TestCase(null, null, "[null, null]")]
-        [TestCase(null, "Second", "[null, \"Second\"]")]
```
mikkelbu, line 128: I guess these are removed (and the same comment for the test below) as the key must not be `null`?

src/NUnitFramework/framework/Internal/ValueTaskAwaitAdapter.cs
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
```
mikkelbu, line 1: I guess this -ValueTaskAwaitAdapter - should be mentioned in the ticket or PR?

mikkelbu (approved):
LGTM. Only minor comments/suggestions

src/NUnitFramework/tests/Constraints/MsgUtilTests.cs
manfred-brands: Yes, indeed. `Dictionary<TKey, TValue> where TKey : notnull`.

src/NUnitFramework/framework/Internal/ValueTaskAwaitAdapter.cs
manfred-brands: It has nothing to do with the ticket, I will extract it into a separate PR.

src/NUnitFramework/testdata/SetUpFixtureData.cs
manfred-brands: Yes, it would only claim to fail on parameter 1, regardless what parameter failed.
Introduced 9 years ago to add a message in case of failure. I guess it never failed.

src/NUnitFramework/framework/Internal/ValueTaskAwaitAdapter.cs
manfred-brands: Done. Removed from this PR and Created #4543

PR #4543, opened 2023-11-08
Add ValueTaskAwaitAdapter

manfred-brands:
This is the equivalent of the `TaskAwaitAdapter` but for `ValueTask` as returned by `DisposeAsync`

Addition to #4530


stevenaw:
Thanks for getting this together @manfred-brands 
I should be able to review over the next few days

stevenaw (approved):
Thanks @manfred-brands no concerns. @jnm2 wrote much of this original awaiting code, so I'm tagging him for any thoughts.

Assuming this is for performance reasons, have you benchmarked this against the existing ReflectionAdapter-based case for `ValueTask` ?

Issue #4554, opened 2023-11-22
SetUpFixture has stopped working going from NUnit 3.13.3 to 3.14.0


nicholashead:
Hi folks, we just updated our unit test project to use the 3.14.0 NuGet package, and now our tests are failing because it appears that the SetUpFixture marked classes are no longer running. Has anyone else seen this behavior? Thanks.

stevenaw:
Thanks for the report @nicholashead 
Do you have a small sample repro you can share which shows the issue?

OsirisTerje:
Just tested it here:  https://github.com/nunit/nunit.issues/tree/main/Issue4554
And it works there.

nicholashead:
Sorry, I'll try to come up with a repro. Downgrading to 3.13.3 fixed it for us, in our somewhat medium/large sized solution.

OsirisTerje:
@nicholashead You can use the repro attempt I added above, and just modify that to make it "not work".

nicholashead:
I was able to identify what the cause is for us - this change - https://github.com/nunit/nunit/issues/4158 - to make the attribute not inheritable broke things. We had a global namespace class that is meant to apply an inherited class's SetUpFixture logic. This works in 3.13.3 but won't in 3.14.0. For us, I think we can just ditch the "global wrapper" class we made. But this may trip some other folks up. Happy to close this since it appears this is an intentional change to lock down the API a little. Sorry for the noise!

OsirisTerje:
Interesting, but not sure I follow the logic.  You had your own setupfixtureattribute, that derived from the nunit setupfixture.  Was that so that it should be easier to apply this other places, other assemblies or namespaces?  
And, you solved it by marking that class as a Setupfixture instead ?

@stevenaw Comments?

nicholashead:
> Interesting, but not sure I follow the logic. You had your own setupfixtureattribute, that derived from the nunit setupfixture. Was that so that it should be easier to apply this other places, other assemblies or namespaces? And, you solved it by marking that class as a Setupfixture instead ?
> 
> @stevenaw Comments?

We have multiple test projects, and one "shared" test project (which is referenced by all the other test projects). We wanted a class in our shared project marked with `[SetUpFixture]` to run for all of our test projects.

Pseudo code Example of what's in the shared project:

```cs
namespace SharedTestLogic {
    [SetUpFixture]
    public class SharedSetup() {
        [OneTimeSetUp]
        public void Setup()
        {
            DoSomething();
        } 
[...]

Issue #3937, opened 2021-09-22
Test run will not start when using [SetUpFixture] as base class


RenePeuser:
What we want to test:
* We want to run Integrationtest with an AspNet Core API - And we will not start and tear down our API every test scenario - Just start once run all test then tear down
* Another aspect would be nice to avoid use static properties in the class which use the [SetUpFixture]
* Would be nice to derive from the a base class which use [SetUpFixture] + [OneTimeSetUp]

What is working
* Like in your description a standalone class for "OneTimeSetup" works - As long you dont derive from this class

What is not working (or maybe not expected behavior)
[Sample](https://softwaresolutions-my.sharepoint.com/:i:/g/personal/rene_peuser_softwaresolutions_onmicrosoft_com/ERKfKMhVADREiqo66UeBZl4B2OMPA7ZzR5mHEJLizUsxZQ?e=m3bxuS)

```csharp
[SetUpFixture]
public class TestEnvironment
{
    [OneTimeSetUp]
    public void Setup()
    {
        Console.WriteLine(nameof(Setup));
    }

    [OneTimeTearDown]
    public void Teardown()
    {
        Console.WriteLine(nameof(Teardown));
    }
}
```

Current behavior:
* The test runner will not run the tests (sample below)*

```csharp
[SetUpFixture]
public abstract class TestEnvironment
{
    protected ApiEnvironment ApiEnvironment {get; private set;}

    [OneTimeSetUp]
    public void Setup()
    {
        Console.WriteLine(nameof(Setup));
        // ApiEnvironment = new ApiEnvironment()....
    }

    [OneTimeTearDown]
    public void Teardown()
    {
        Console.WriteLine(nameof(Teardown));
        // ApiEnvironment?.Dispose();
    }
}

public class Test_A : TestEnvironment
{
    [Test]
    public void Test_1()
    {
        Assert.Pass();
        // APIEnvironemt.Client....
    }

    [Test]
    public void Test_2()
    {
        Assert.Pass();
        // APIEnvironemt.Client....
    }
}
```

RenePeuser:
Hi something new about this issue ?

CharliePoole:
FWIW a SetUpFixture was never intended to be used as a base class. In the past year I have noticed several people doing that so I updated the docs for that purpose.

denis-sklyarov:
I develop integrated tests and i need to execute some code exactly after all test cases. Is there are wa y to do this using nunit 3 ?

[...]

PR #4555, opened 2023-11-23
(For Discussion) Revert "Indicate that SetupFixtureAttribute should not be used on base-classes"

stevenaw:
Relates to https://github.com/nunit/nunit/issues/4554 https://github.com/nunit/nunit/issues/3937 https://github.com/nunit/nunit/issues/4158

Based on the discussion in https://github.com/nunit/nunit/issues/4554 and https://github.com/nunit/nunit/issues/3937 it seems as though nunit has traditionally not supported `SetupFixture` to work on a base class, though it doesn't seem to appear in the docs. The last time I can find an [explicit discussion](https://github.com/nunit/nunit/issues/3937#issuecomment-1081302248) it appears that this was intended to not work. It appears to have accidentally started working at some point, likely in 3.13.3.

We explicitly turned off the `Inherited` flag on the attribute in 3.14 and 4.0, which inadvertently seem to have broken some users of 3.x who had built an inheritance structure around `SetupFixture`.

I've expressed hesitation about reverting the change in 4.0 unless we decide as a team it _should_ work but in the interest of compatibility I've opened this PR for us to consider reverting it on 3.x. There's pros and cons to merging this. Pros would be minimizing accidental breaking changes in 3.x while the cons would be explicitly reenabling a behaviour we seem to want to discourage.

@OsirisTerje @jnm2 I'd like to hear your thoughts on this. I'm also happy to have this PR close without merging if we want to actively discourage this pattern in 3.x. The workaround for users who hit this in 3.14 is simply to add the attribute at each level of their inheritance hierarchy.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/SetUpFixtureAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/SetUpFixtureAttribute.cs
@@ -39,7 +39,7 @@ namespace NUnit.Framework
     /// <see cref="OneTimeTearDownAttribute" /> methods for all the test fixtures
     /// under a given namespace.
     /// </summary>
-    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
     public class SetUpFixtureAttribute : NUnitAttribute, IFixtureBuilder
     {
         #region ISuiteBuilder Members
--- a/src/NUnitFramework/tests/Attributes/SetUpFixtureAttributeTests.cs
+++ b/src/NUnitFramework/tests/Attributes/SetUpFixtureAttributeTests.cs
@@ -88,7 +88,7 @@ namespace NUnit.Framework.Attributes
             var usageAttrib = Attribute.GetCustomAttribute(typeof(SetUpFixtureAttribute), typeof(AttributeUsageAttribute)) as AttributeUsageAttribute;
 
             Assert.NotNull(usageAttrib);
-            Assert.False(usageAttrib.Inherited);
+            Assert.True(usageAttrib.Inherited);
         }
 
         private static class StaticSetupClass


manfred-brands:
In my opinion having a class execute things without having to specify `SetupFixture` on that class could lead to this being run unexpectedly.

OsirisTerje:
I had to ask  @CharliePoole about this, and I got this answer back from him, which I think clears up the confusion.  

> So before I get further into this, I'm wondering if my original point was perhaps misunderstood...
>
> The idea wasn't that SetUpFixture classes can't be inherited from but that they can't be inherited
> by TestFixture classes. I wonder if that's clear in all the current discussion.
> 
> Simple explanation: some attributes provide a class with an identity so the class `isa` TestFixture
> or `isa` SetUpFixture. These are two different things and behave differently. Originally, in the framework,
> I think we felt everyone understood that and would not use inheritance to make the same class be both a
> SetUpFixture and a TestFixture (or TestFixture base class).
> 
> So... a TestFixture is a class, which contains tests.
> 
> A SetupFixture is a class without any tests, which provides onetimesetup for some group of TestFixtures.
> 
> By mixing the two concepts, various things can go wrong. Most obviously, there's the fact that `OneTimeSetUp`
> and `OneTimeTearDown` in a SetUpFixture has different semantics from the same attribute in a TestFixture.
> 
> I can go more into detail but please let me know if I'm on the right track in terms of what people are thinking
> the "don't inherit" warning was intended to mean.
> 
> Charlie

From a maintainability view, it would be good to always add the attribute to inherited classes, but - it is not a requirement.  It could be an advice.

More important, attempts to add a TestFixture attribute to a class derived from a SetupFixture should be as prohibited as we can. I would assume this could be done by the Analyzer.

And based on what Charlie writes above, we should revert this, because there is nothing in the framework that disallows you from inheriting from a SetupFixture.

OsirisTerje (approved):

manfred-brands:
Thanks for clarifying. Yes then it should we reverted. Although we can add a rule to the analyzer to warn at edit time, the framework should also throw an error at runtime if any class with tests derives from a SetUpFixture annotated class

OsirisTerje:
> , the framework should also throw an error at runtime if any class with tests derives from a SetUpFixture annotated class

Yes, this makes sense, we should do that.  But, we should still have it in the analyzer so the developer gets an "early warning", or "early error".

manfred-brands:
> > , the framework should also throw an error at runtime if any class with tests derives from a SetUpFixture annotated class
> 
> Yes, this makes sense, we should do that. But, we should still have it in the analyzer so the developer gets an "early warning", or "early error".

Yes, I meant we need to do both.

OsirisTerje:
I'll also move Charlie's comment into the documentation for SetupFixture.  It is a very good clarification.

stevenaw:
Thanks for helping confirm that @OsirisTerje 
Sounds like we should also revert in 4.0. I can get that PR setup too

stevenaw:
Hm. The PR doesn't seem to be getting picked up by CI at all, I don't see anything in the list of "All workflows" in Github actions.
I'm unsure if the branch name prefix `discussion-` is complicating anything here, but I'll try remaking this PR with a new branch name in a bit.

OsirisTerje (approved):

PR #4603, opened 2024-01-18
#4602 WpfMessagePumpStrategy

### LINKED ISSUE #4602: WpfMessagePumpStrategy - change from Dispatcher.Run to Dispatcher.PushFrame
Currently when awaiting an awaitable test nunit post to the context and calls Dispatcher.Run. In the Post a continuation is scheduled which calls ExitAllFrames on the dispatcher. This leads to some problems if there is already a frame running. 

My idea (and I have a working version) is to create a DispatcherFrame and call PushFrame on the Dispatcher. If the posted action is completed just set frame.Continue to false. This will cause the PushFrame to return and will not disturb a previously pushed frame. Would you guys be interested in a PR for this?

soerendd:
Replace Dispatcher.Run with PushFrame as Dispatcher.Run needs to exit all frames to return properly which causes issues if there is already an outer frame.

Should fix #4602

### DIFF
--- a/src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
+++ b/src/NUnitFramework/framework/Internal/MessagePumpStrategy.cs
@@ -110,13 +110,18 @@ namespace NUnit.Framework.Internal
         {
             private static WpfMessagePumpStrategy? _instance;
 
-            private readonly Action _dispatcherRun;
-            private readonly Action _dispatcherExitAllFrames;
+            private readonly MethodInfo _dispatcherPushFrame;
+            private readonly MethodInfo _dispatcherFrameSetContinueProperty;
+            private readonly Type _dispatcherFrameType;
 
-            private WpfMessagePumpStrategy(Action dispatcherRun, Action dispatcherExitAllFrames)
+            private WpfMessagePumpStrategy(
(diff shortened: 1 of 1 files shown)


jnm2 (approved):
Thanks so much! LGTM

jnm2:
If it's possible to write a test that fails for the previous strategy and succeeds here, that would be great.

soerendd:
> If it's possible to write a test that fails for the previous strategy and succeeds here, that would be great.

Sure thing. I checked if there is a test for the message pump strategies. There are none.

I will check where to add this test

soerendd:
> > If it's possible to write a test that fails for the previous strategy and succeeds here, that would be great.
> 
> Sure thing. I checked if there is a test for the message pump strategies. There are none.
> 
> I will check where to add this test

@jnm2  I'm sorry but I haven't found a place where to put tests for that. The reason is that it would require a reference to System.Windows.Threading (wpf) and I do not see an easy way to mock this properly.

OsirisTerje:
@soerendd The builds fails due to a mismatch on sdks.  This is fixed in the master branch, so if you just merge in the master branch and push that up, the builds should work just fine.

soerendd:
@OsirisTerje Merged and pushed.

PR #4607, opened 2024-01-21
chore: Refer to global.json for SDK version

### LINKED ISSUE #4606: GitHub Actions fail on `master` for Windows and Linux builds
The builds are failing on `master` as our `global.json` expects version "8.0.100" for the SDK and the build agents - for Windows and Linux - now install 8.0.101. I guess we would like to update the `global.json` at some point, but I'm unsure if we can expect all contributors to be updating immediately when a new version comes out, so perhaps we should initially just supply a more specific version than `8.0.x` in `setup-dotnet` and then bump the `global.json` and the CI agents later ?

image

mikkelbu:
Fixes #4606

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -17,13 +17,11 @@ jobs:
     - name: ⤵️ Checkout Source
       uses: actions/checkout@v3
 
-    - name: 🛠️ Setup .NET 
+    - name: 🛠️ Setup .NET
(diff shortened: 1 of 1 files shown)


manfred-brands:
We now no longer install .net 6.0 ourselves.
The tests pass because it is pre-installed on GitHub runners. Assuming this will stay the case while it is supported by MS I'm ok with this change.

OsirisTerje (approved):
lgtm

PR #4608, opened 2024-01-24
Issues/4572 properties comparer

### LINKED ISSUE #4572: Make new PropertiesComparer optional
I updated NUnit to bersion 4.0 and run my tests like usual. After doing the required migrations some tests that I didn't even touched start to fail. After some investigation I discovered that this is related to the new PopertiesComparer added by #4394.

The affected tests compare classes via `Is.EqualTo()` and before they were compared by reference. But due to some objects having the exact same properties the new comparer now consideres them equal, failing my tests.

The new comparer is used by default and it looks like it cannot be disabled in the current version. 

Example of  a Test that worked before but fails in version 4:
```
private class TestClass 
{
    public int Prop { get; set; } = 0;
}

[Test]
public void CompareObjects()
{
    TestClass a = new() { Prop = 2 };
    TestClass b = new() { Prop = 2 };

    Assert.That(b, Is.Not.EqualTo(a));
}
```

### LINKED ISSUE #4591: Parameter count mismatch with indexer
Hi there,
for code as simple as this I get an error. (see the stacktrace)

```csharp
[Test]
public void Test_Accessor()
{
  var a = new X();
  var b = new X();

  Assert.That( a, Is.Not.EqualTo( b ) );
}
		
public class X
{
  private readonly Guid _ID = Guid.NewGuid();

  public Guid? this[ Guid characteristicUuid ] => _ID;
}
```

*Stack trace*
```
System.Reflection.TargetParameterCountException : Parameter count mismatch.
   at System.Reflection.RuntimePropertyInfo.GetValue(Object obj, BindingFlags invokeAttr, Binder binder, Object[] index, CultureInfo culture)
   at System.Reflection.RuntimePropertyInfo.GetValue(Object obj, Object[] index)
   at NUnit.Framework.Constraints.Comparers.PropertiesComparer.Equal(Object x, Object y, Tolerance& tolerance, ComparisonState state, NUnitEqualityComparer equalityComparer)
   at NUnit.Framework.Constraints.NUnitEqualityComparer.AreEqual(Object x, Object y, Tolerance& tolerance, ComparisonState state)
   at NUnit.Framework.Constraints.NUnitEqualityComparer.AreEqual(Object x, Object y, Tolerance& tolerance)
   at NUnit.Framework.Constraints.EqualConstraint.ApplyTo[TActual](TActual actual)
   at NUnit.Framework.Constraints.NotConstraint.ApplyTo[TActual](TActual actual)
   at NUnit.Framework.Assert.That[TActual](TActual actual, IResolveConstraint expression, NUnitString message, String actualExpression, String constraintExpression)
   at My.TestClass.Test_Accessor() in C:\TestClass.cs:line 007
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
```

Is this a bug?

Best regards.

manfred-brands:
Fixes #4572 
Fixes #4591

By default behaviour is as per 3.14
A new `UsingPropertiesComparer()` can be used as an opt-in.

Also fixes when using indexers with the property comparer.
Also fixes nested references.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System;
+using System.Linq;
 using System.Reflection;
 
 namespace NUnit.Framework.Constraints.Comparers
@@ -29,9 +30,10 @@ namespace NUnit.Framework.Constraints.Comparers
             }
 
             PropertyInfo[] properties = xType.GetProperties(BindingFlags.Instance | BindingFlags.Public);
-            if (properties.Length == 0 || properties.Length >= 32)
+            if (properties.Length == 0 || properties.Length >= 32 || properties.Any(p => p.GetIndexParameters().Length > 0))
             {
                 // We can't compare if there are no (or too many) properties.
+                // We also can't deal with indexer properties as we don't know the range of valid values.
                 return EqualMethodResult.TypesNotSupported;
             }
 
--- a/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/AnyOfConstraint.cs
@@ -130,6 +130,19 @@ namespace NUnit.Framework.Constraints
             return this;
         }
 
+        /// <summary>
+        /// Enables comparing of instance properties.
+        /// </summary>
+        /// <remarks>
+        /// This allows comparing classes that don't implement <see cref="IEquatable{T}"/>
+        /// without having to compare each property separately in own code.
+        /// </remarks>
+        public AnyOfConstraint UsingPropertiesComparer()
+        {
+            _comparer.CompareProperties = true;
+            return this;
+        }
+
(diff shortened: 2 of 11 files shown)


src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
```diff
@@ -29,9 +30,10 @@ public static EqualMethodResult Equal(object x, object y, ref Tolerance toleranc
             }
 
             PropertyInfo[] properties = xType.GetProperties(BindingFlags.Instance | BindingFlags.Public);
-            if (properties.Length == 0 || properties.Length >= 32)
+            if (properties.Length == 0 || properties.Length >= 32 || properties.Any(p => p.GetIndexParameters().Length > 0))
```
stevenaw, line 33: Good idea! I hadn't thought about that

stevenaw:
Thanks @manfred-brands ! I tried to do a quick review just now before heading out of town early tomorrow.
This looks good to me, however I could only do a quick pass. It could benefit from a closer look.

OsirisTerje (approved):
Looks good to me.  Good solution!

PR #4617, opened 2024-02-07
Switch to using license expression

lahma:
By using license expression it will show up correctly on nuget.org and helps license validators/checkers to know what is the exact license type. Usually you have this in csproj file producing the nuspec, but I checked what kind of file is generated with `<PackageLicenseExpression>MIT</PackageLicenseExpression>`. The package will also still contain the license file.

Currently on [nuget.org](https://www.nuget.org/packages/NUnit):

image

Ideally like for [Newtonsoft.Json](https://www.nuget.org/packages/Newtonsoft.Json):

image

### DIFF
--- a/nuget/nunitlite/nunitlite.nuspec
+++ b/nuget/nunitlite/nunitlite.nuspec
@@ -6,7 +6,7 @@
     <version>$version$</version>
     <authors>Charlie Poole, Rob Prouse</authors>
     <owners>Charlie Poole, Rob Prouse</owners>
-    <license type="file">LICENSE.txt</license>
+    <license type="expression">MIT</license>
     <projectUrl>https://nunit.org</projectUrl>
     <repository type="git" url="https://github.com/nunit/nunit"/>
     <icon>icon.png</icon>
--- a/nuget/framework/nunit.nuspec
+++ b/nuget/framework/nunit.nuspec
@@ -6,7 +6,7 @@
     <version>$version$</version>
     <authors>Charlie Poole, Rob Prouse</authors>
     <owners>Charlie Poole, Rob Prouse</owners>
-    <license type="file">LICENSE.txt</license>
+    <license type="expression">MIT</license>
     <projectUrl>https://nunit.org</projectUrl>
     <repository type="git" url="https://github.com/nunit/nunit"/>
     <icon>icon.png</icon>


nuget/nunitlite/nunitlite.nuspec
```diff
@@ -6,7 +6,8 @@
     <version>$version$</version>
     <authors>Charlie Poole, Rob Prouse</authors>
     <owners>Charlie Poole, Rob Prouse</owners>
-    <license type="file">LICENSE.txt</license>
+    <license type="expression">MIT</license>
+    <licenseUrl>https://licenses.nuget.org/MIT</licenseUrl>
```
OsirisTerje, line 10: Is this url correct?  Without it, the license points to this URL anyway, see NUnit3TestAdapter, and it then doesn't have the fields filled out.   Do you have an another package as an example?

OsirisTerje (changes requested):
Not sure this is the way to go, and if we do it, it must have the year info and owner info.

lahma:
Sorry, I need to get back to this a bit later. Generally what you see is generated and "machine readable" so what should be advertised as the license and as an agreement.

Based on the general license notes the names of people seem to be a bit auxiliary to the actual license. They seem to mainly provide context about the work that has been done when permissive licenses are being used. So for example GitHub shows info as "MIT license" regardless of who has been named the actual person that did the hard work.

From NUnit repo the license "headers" GitHub provides:

image

And the right-hand bar

image

I hope I don't indicate I want to change the license, I would want just to provide the general go/no-go decision for anyone browsing the nuget.org. OK fair, I'm trying to make the license checkers happy 😉

lahma:
I force-pushed a fix to point to GitHub repository's license file. This hopefully still keeps license analyzers happy.

lahma:
...and it doesn't work...

```
PackageFramework
Attempting to build package from 'nunit.temp.nuspec'.
The licenseUrl and license elements cannot be used together.
An error occurred when executing task 'PackageFramework'.
```

So package needs to have just the expression and license can be part of the regular root package location, which it already is.

OsirisTerje:
About the license, it is not about showing the MIT license symbol, but when you go to that license, it has template fields that are not filled in - for year and names.

lahma:
Yes I understand the concern, but Microsoft has the same situation: https://www.nuget.org/packages/Microsoft.Extensions.DependencyInjection/8.0.0 , same kind of template link there but the actual package contains `LICENSE.txt` : https://nuget.info/packages/Microsoft.Extensions.DependencyInjection/8.0.0 .

I believe the expression of indicator of license format and restrictions and then the actual license is/can be part of the package and of course part of original repository which holds more details. So as a consumer if I see that it's MIT/Apache-2.0/BSD I know where I'm getting into and I know I just need to be ready to show the project's license which includes the actual authors and copyrights but I don't need to worry about redistribution rights etc.

lahma:
For reference I have [this fairly popular NuGet package](https://www.nuget.org/packages/Quartz) with same kind of template shown on NuGet.org, I don't mind - I still have the "actual license" that follows the template and has my name and year on it.

lahma:
And as an another example, [NUnit3TestAdapter uses the expression and doesn't package the license with the package](https://nuget.info/packages/NUnit3TestAdapter/4.5.0).

OsirisTerje:
Got me there :-)

OsirisTerje (approved):
Have no choice anymore :-)

lahma:
Thanks for accepting the change,  I hope it will be for the common good 🙂

PR #4618, opened 2024-02-11
Add support for DateTime comparison with TimeSpan tolerance

### LINKED ISSUE #4600: Add `DateTime`/`TimeSpan` support for inequality tolerance
It would be nice if the following code worked:

``` csharp
public static DateTime LastDateAccessed { get; set; } // simulate a remote value

[Test]
public void TestMe()
{
    DateTime beforeUpdate = DateTime.UtcNow;

    // Do operation
    LastDateAccessed = DateTime.UtcNow.AddMilliseconds(-50); // some kind of timing imperfection on the server

    // This fails because of the timing issue
    Assert.That(LastDateAccessed, Is.GreaterThan(beforeUpdate));

    // It would be nice to do this
    Assert.That(LastDateAccessed, Is.GreaterThan(beforeUpdate).Within(TimeSpan.FromSeconds(1)));
}
```

For consistency, this should probably be done on all constraints that extend [`ComparisonConstraint`](https://github.com/search?q=repo%3Anunit%2Fnunit+%22%3A+ComparisonConstraint%22&type=code).

I am willing to take up the task. I just want to make sure this is kind of change would be accepted.

RenderMichael:
Fixes #4600

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Tolerance.cs
+++ b/src/NUnitFramework/framework/Constraints/Tolerance.cs
@@ -278,6 +278,11 @@ namespace NUnit.Framework.Constraints
                 return new Range(v - amount, v + amount);
             }
 
+            if (Amount is TimeSpan amountDateTime && value is DateTime timeSpan)
+            {
+                return new Range(timeSpan - amountDateTime, timeSpan + amountDateTime);
+            }
+
             throw new InvalidOperationException("Cannot create range for a non-numeric value");
         }
 
--- a/src/NUnitFramework/framework/Constraints/Tolerance.cs
+++ b/src/NUnitFramework/framework/Constraints/Tolerance.cs
@@ -278,9 +278,9 @@ namespace NUnit.Framework.Constraints
                 return new Range(v - amount, v + amount);
             }
 
-            if (Amount is TimeSpan amountDateTime && value is DateTime timeSpan)
+            if (Amount is TimeSpan interval && value is DateTime dateTime)
             {
-                return new Range(timeSpan - amountDateTime, timeSpan + amountDateTime);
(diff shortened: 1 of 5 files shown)


RenderMichael:
I had to pull out the test data into a source method because `DateTime`/`TimeSpan` are not allowed in attributes.

Also DateTime's formatting in the error messages is different from the normal `ToString()` is for the existing `double` and `int` tests, so I called `MsgUtils.FormatValue` directly like the exception message formatting itself does.

manfred-brands (changes requested):
Thanks @RenderMichael.

The code changes are nice and simple.

However as it also affects other areas where `Tolerance` is used, could you also please update the other constraint tests: `GreaterThanOrEqualConstraintTests`,  `LessThanConstraintTests` and `LessThanOrEqualConstraintTests` by adding use cases for `TimeSpan`.

RenderMichael:
Thanks for the feedback @manfred-brands 

The PR has been updated, I also added an equality test to the existing Greater Than tests.

manfred-brands (approved):
Thanks again @RenderMichael for adding those extra tests.

Nothing more from me.

RenderMichael:
Sorry for the update after the fact, I noticed I used some terrible variable names in the tolerance method. Fixed with no functional changes

stevenaw (approved):
Thanks for the contribution @RenderMichael ! Appreciate the final cleanup you did too :)

stevenaw:
I'm going to try converting this to a draft and back to see if it un-sticks the WIP check

OsirisTerje:
@stevenaw Merged.

stevenaw:
Thanks @OsirisTerje

PR #4620, opened 2024-02-15
Add explicit generic type specification to TestCase

stevenaw:
Fixes https://github.com/nunit/nunit/issues/1215

Adds support for explicitly specifying a generic type for a test case method. I've chosen to keep it simple and let the runtime itself ensure that cases like this work or not:
```
[TestCase("2", TypeArgs = new[] { typeof(int) })]
public void ExplicitTypeArgsWithIncompatibleParameters<T>(T input)
```

I'd like to hear thoughts on if we wanted to allow the framework to eagerly detect and mark these as `NonRunnable`. My own suggestion is to propose we defer it to another PR

### DIFF
--- a/global.json
+++ b/global.json
@@ -2,6 +2,6 @@
   "sdk": {
     "version": "8.0.100",
     "allowPrerelease": false,
-    "rollForward": "disable"
+    "rollForward": "latestFeature"
   }
 }
--- a/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
@@ -248,6 +248,11 @@ namespace NUnit.Framework
         public string? ExcludePlatform { get; set; }
 
         /// <summary>
+        /// Comma-delimited list of platforms to run the test for
+        /// </summary>
+        public Type[]? TypeArgs { get; set; } = null;
+
+        /// <summary>
         /// Gets and sets the category for this test case.
         /// May be a comma-separated list of categories.
         /// </summary>
@@ -300,7 +305,10 @@ namespace NUnit.Framework
                 int argsNeeded = parameters.Length;
                 int argsProvided = Arguments.Length;
 
-                parms = new TestCaseParameters(this);
+                parms = new TestCaseParameters(this)
+                {
+                    TypeArgs = TypeArgs
+                };
 
                 // Special handling for ExpectedResult (see if it needs to be converted into method return type)
                 if (parms.HasExpectedResult
--- a/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
@@ -248,7 +248,9 @@ namespace NUnit.Framework
         public string? ExcludePlatform { get; set; }
 
         /// <summary>
-        /// Comma-delimited list of platforms to run the test for
+        /// Get or set the type arguments for a generic test method.
+        /// If not set explicitly, the generic types will be inferred
+        /// based on the test case parameters.
         /// </summary>
         public Type[]? TypeArgs { get; set; } = null;
 
--- a/src/NUnitFramework/framework/Internal/TestCaseParameters.cs
+++ b/src/NUnitFramework/framework/Internal/TestCaseParameters.cs
@@ -81,6 +81,11 @@ namespace NUnit.Framework.Internal
 
         #endregion
 
-        internal Type[]? TypeArgs { get; set; } = null;
+        /// <summary>
+        /// Get or set the type arguments for a generic test method.
+        /// If not set explicitly, the generic types will be inferred
+        /// based on the test case parameters.
+        /// </summary>
+        public Type[]? TypeArgs { get; set; } = null;
     }
 }
--- a/src/NUnitFramework/framework/Internal/TestCaseParameters.cs
+++ b/src/NUnitFramework/framework/Internal/TestCaseParameters.cs
@@ -80,5 +80,7 @@ namespace NUnit.Framework.Internal
         public bool HasExpectedResult { get; set; }
 
         #endregion
+
+        internal Type[]? TypeArgs { get; set; } = null;
     }
 }
--- a/src/NUnitFramework/testdata/TestCaseAttributeFixture.cs
+++ b/src/NUnitFramework/testdata/TestCaseAttributeFixture.cs
@@ -114,5 +114,10 @@ namespace NUnit.TestData.TestCaseAttributeFixture
         public void MethodWithArrayArguments(object o)
         {
         }
+
+        [TestCase("doesn't work", TypeArgs = new[] { typeof(int) })]
+        public void MethodWithIncompatibleTypeArgs<T>(T input)
+        {
+        }
     }
 }
--- a/src/NUnitFramework/testdata/TestCaseAttributeFixture.cs
+++ b/src/NUnitFramework/testdata/TestCaseAttributeFixture.cs
@@ -116,7 +116,7 @@ namespace NUnit.TestData.TestCaseAttributeFixture
         }
 
         [TestCase("doesn't work", TypeArgs = new[] { typeof(int) })]
-        public void MethodWithIncompatibleTypeArgs<T>(T input)
+        public static void MethodWithIncompatibleTypeArgs<T>(T input)
         {
         }
     }
--- a/src/NUnitFramework/testdata/TestCaseSourceAttributeFixture.cs
+++ b/src/NUnitFramework/testdata/TestCaseSourceAttributeFixture.cs
@@ -142,6 +142,11 @@ namespace NUnit.TestData.TestCaseSourceAttributeFixture
         {
         }
 
+        [TestCaseSource(nameof(IncompatibleGenericTypeAndArgumentTestCases))]
+        public void MethodWithIncompatibleGenericTypeAndArgument(object o)
+        {
+        }
+
         private static IEnumerable ExceptionSource
         {
             get
@@ -188,6 +193,14 @@ namespace NUnit.TestData.TestCaseSourceAttributeFixture
         {
             foreach (var argumentValue in ComplexArrayBasedTestInput)
                 yield return new TestCaseData(args: new[] { argumentValue });
+        }
+
+        public static IEnumerable<TestCaseData> IncompatibleGenericTypeAndArgumentTestCases()
+        {
+            yield return new TestCaseData("doesn't work")
+            {
+                TypeArgs = new[] { typeof(string) }
+            };
         }
 
         #region Test name tests
--- a/src/NUnitFramework/testdata/TestCaseSourceAttributeFixture.cs
+++ b/src/NUnitFramework/testdata/TestCaseSourceAttributeFixture.cs
@@ -143,7 +143,7 @@ namespace NUnit.TestData.TestCaseSourceAttributeFixture
         }
 
         [TestCaseSource(nameof(IncompatibleGenericTypeAndArgumentTestCases))]
-        public void MethodWithIncompatibleGenericTypeAndArgument(object o)
+        public static void MethodWithIncompatibleGenericTypeAndArgument<T>(T o)
         {
         }
 
@@ -199,7 +199,7 @@ namespace NUnit.TestData.TestCaseSourceAttributeFixture
         {
             yield return new TestCaseData("doesn't work")
             {
-                TypeArgs = new[] { typeof(string) }
+                TypeArgs = new[] { typeof(int) }
             };
         }
 
--- a/src/NUnitFramework/tests/Attributes/TestCaseAttributeTests.cs
+++ b/src/NUnitFramework/tests/Attributes/TestCaseAttributeTests.cs
@@ -756,21 +756,22 @@ namespace NUnit.Framework.Tests.Attributes
             Assert.That(typeof(T), Is.EqualTo(typeof(int)));
         }
 
-        [TestCase(2, TypeArgs = new[] { typeof(int) })]
-        public void ExplicitTypeArgsWithUnrelatedParameters<T>(int input)
+        [TestCase("2", TypeArgs = new[] { typeof(long) })]
+        public void ExplicitTypeArgsWithUnrelatedParameters<T>(string input)
         {
-            Assert.That(typeof(T), Is.EqualTo(typeof(int)));
-            Assert.That(input, Is.EqualTo(2));
+            Assert.That(typeof(T), Is.EqualTo(typeof(long)));
+            Assert.That(input, Is.EqualTo("2"));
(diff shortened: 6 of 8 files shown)


global.json
```diff
@@ -2,6 +2,6 @@
   "sdk": {
     "version": "8.0.100",
     "allowPrerelease": false,
-    "rollForward": "disable"
+    "rollForward": "latestFeature"
```
stevenaw, line 5: Unrelated to the issue at hand, but I needed to change this to be able to build under the following configuration:
image

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
```diff
@@ -247,6 +247,11 @@ public bool Explicit
         /// </summary>
         public string? ExcludePlatform { get; set; }
 
+        /// <summary>
+        /// Comma-delimited list of platforms to run the test for
```
stevenaw, line 251: Note to self: Fix this copy/paste error

global.json
manfred-brands: You can set `"rollForward": "latestPatch"` to allow 8.0.1xx.
Now it also allows 8.0.2xx and MS is has in the past broken things when adding new "features". (6.0.300)
But I haven't found anything breaking with 8.0.200

src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
manfred-brands: Yes, copy it from TestFixtureAttribute's TypeArgs property instead.

src/NUnitFramework/testdata/TestCaseAttributeFixture.cs
```diff
@@ -114,5 +114,10 @@ public void MethodWithExcludeRuntime(int num)
         public void MethodWithArrayArguments(object o)
         {
         }
+
+        [TestCase("doesn't work", TypeArgs = new[] { typeof(int) })]
```
manfred-brands, line 118: I feel an Analyzer rule request coming up.

src/NUnitFramework/tests/Attributes/TestCaseAttributeTests.cs
```diff
@@ -748,5 +749,38 @@ public async Task<T> TestWithAsyncGenericReturnType<T>(T arg1)
...
 
         #endregion
+
+        [TestCase(TypeArgs = new[] { typeof(int) })]
+        public void ExplicitTypeArgsWithoutParameters<T>()
+        {
+            Assert.That(typeof(T), Is.EqualTo(typeof(int)));
+        }
+
+        [TestCase(2, TypeArgs = new[] { typeof(int) })]
```
manfred-brands, line 759: Doesn't this already without having to add `TypeArgs` as the `TryGetTypeArguments` determines the types from the actual parameters?

src/NUnitFramework/tests/Attributes/TestCaseAttributeTests.cs
```diff
@@ -748,5 +749,38 @@ public async Task<T> TestWithAsyncGenericReturnType<T>(T arg1)
...
+        public void ExplicitTypeArgsWithUnrelatedParameters<T>(int input)
```
manfred-brands, line 760: I think the name of the method is wrong as `2` is `typeof(int)`
This seems more the test that belongs to the `ExplicitTypeArgsWithCompatibleParameters` method below.

Can you add an practical example where the default determined type argument is wrong?
The only one I can think is a class which has an (implicit) conversion from string, e..g. DateTime.

manfred-brands (changes requested):
See comments requesting a practical example where current code without TypeArgs won't work, but it does now.

src/NUnitFramework/tests/Attributes/TestCaseAttributeTests.cs
stevenaw: Very true. My thinking behind the name was that the generic method type and the argument type are "unrelated" from the compiler's perspective - as opposed to if `input` were type `T` as well. Having them still be the same underlying type leaves something to be desired though.

My thinking behind this test and the one above where `input` _is_ of type `T` is that they represent a change in responsibilities internally. Previously, or presently but without passing `TypeArgs`, it is up to the framework to infer what generic types to use for the method and it does this through inference based on the values passed into the test case as arguments. Now, when `TypeArgs` is specified, the responsibility for that shifts to the user as we will always use their specified types for the generic method - even if it conflicts with the values they try to explicitly pass in! This last scenario is the `MethodWithIncompatibleTypeArgs<T>(T input)` test added.

I'll try to revisit some of these - in particular the types we pass - to make it a bit clearer and robust.

global.json
stevenaw: Thanks! I hadn't realized there was a precedent there.

stevenaw:
Thanks for syncing your changes here @OsirisTerje 

@manfred-brands as I've been revisiting the test scenarios, I've realized it's likely very minimal extra work to make this work for TestCaseSource - something Charlie had also originally advocated for in the issue.
I have had a few more pockets of time than expected, so I think the next push will likely include support for TestCaseAttribute as well.

src/NUnitFramework/framework/Internal/TestCaseParameters.cs
```diff
@@ -80,5 +80,12 @@ public TestCaseParameters(ITestCaseData data) : base(data)
         public bool HasExpectedResult { get; set; }
 
         #endregion
+
+        /// <summary>
+        /// Get or set the type arguments for a generic test method.
+        /// If not set explicitly, the generic types will be inferred
+        /// based on the test case parameters.
+        /// </summary>
+        public Type[]? TypeArgs { get; set; } = null;
```
stevenaw, line 89: I was surprised at just _how_ little was required to make it work for TestCaseSource after TestCase was done. If I had made this public originally instead of private then it would've "just worked"

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -480,6 +480,83 @@ public void TestNameIntrospectsArrayValues()
...
+
+            Assert.That(test.RunState, Is.EqualTo(RunState.Runnable));
+            var exception = Assert.Throws<NUnitException>(() => test.Method!.Invoke(test.Parent, test.Arguments));
+
+            Assert.That(exception.InnerException, Has.Message.EqualTo("Object does not match target type."));
+        }
+
+        [TestCaseSource(nameof(ExplicitTypeArgsWithGenericConstraintSatisfiedTestCases))]
+        public void ExplicitTypeArgsWithGenericConstraintSatisfied<T1, T2>(T1 a, T2 b)
+            where T1 : IComparer<T2>
```
stevenaw, line 538: Scenarios like this are why I'd prefer to leave it to the runtime to validate the arguments satisfy the generic types as I suspect it could get tricky quickly.

src/NUnitFramework/testdata/TestCaseSourceAttributeFixture.cs
```diff
@@ -142,6 +142,11 @@ public void MethodWithArrayArguments(object o)
         {
         }
 
+        [TestCaseSource(nameof(IncompatibleGenericTypeAndArgumentTestCases))]
+        public void MethodWithIncompatibleGenericTypeAndArgument(object o)
```
manfred-brands, line 146: This is not a generic method, and `NUnitTestCaseBuilder` doesn't even use the `TypeArgs`.

src/NUnitFramework/testdata/TestCaseSourceAttributeFixture.cs
```diff
@@ -190,6 +195,14 @@ private static IEnumerable<TestCaseData> ComplexArrayBasedTestInputTestCases()
                 yield return new TestCaseData(args: new[] { argumentValue });
         }
 
+        public static IEnumerable<TestCaseData> IncompatibleGenericTypeAndArgumentTestCases()
+        {
+            yield return new TestCaseData("doesn't work")
+            {
+                TypeArgs = new[] { typeof(string) }
+            };
```
manfred-brands, line 203: These seem compatible types to me. `doesn't work` is a `string`.

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -480,6 +480,83 @@ public void TestNameIntrospectsArrayValues()
...
+            var suite = TestBuilder.MakeParameterizedMethodSuite(
+                typeof(TestCaseSourceAttributeFixture),
+                nameof(TestCaseSourceAttributeFixture.MethodWithIncompatibleGenericTypeAndArgument));
+
+            var test = suite.Tests[0];
+
+            Assert.That(test.RunState, Is.EqualTo(RunState.Runnable));
+            var exception = Assert.Throws<NUnitException>(() => test.Method!.Invoke(test.Parent, test.Arguments));
+
+            Assert.That(exception.InnerException, Has.Message.EqualTo("Object does not match target type."));
```
manfred-brands, line 533: This is wrong. The test now fails because `test.Parent` is of type `ParameterizedMethodSuite` and not `TestCaseSourceAttributeFixture`
The error talks about `target` type which is the `this` argument, not about parameters.
You have to call `TestBuilder.RunTest` and check the results which shows that the test fails.
However, in case of mismatch of type arguments, the test is marked as non-runnable as you don't have to run it at al.

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
manfred-brands: Yes, the Analyzer is also skipping `TestCaseDate`

src/NUnitFramework/tests/Attributes/TestCaseAttributeTests.cs
```diff
@@ -756,21 +756,22 @@ public void ExplicitTypeArgsWithoutParameters<T>()
...
+        [TestCase(2, TypeArgs = new[] { typeof(long) }, ExpectedResult = typeof(long))]
+        [TestCase(2L, TypeArgs = new[] { typeof(long) }, ExpectedResult = typeof(long))]
+        [TestCase(2, ExpectedResult = typeof(int))]
+        [TestCase(2L, ExpectedResult = typeof(long))]
+        public Type GenericMethodAndParameterWithExplicitOrImplicitTyping<T>(T input)
+            => typeof(T);
 
         [Test]
-        public void ExplicitTypeArgsWithIncompatibleParametersFailsAtRuntime()
+        public void ExplicitTypeArgsWithUnassignableParametersFailsAtRuntime()
```
manfred-brands, line 774: This test has the same issue as mentioned before. Don't run the test, the discovery turns the test into non-runnable.

manfred-brands (changes requested):
Some of the tests are wrong and asserting failures that are the result of that instead of what really should be happening.

src/NUnitFramework/testdata/TestCaseSourceAttributeFixture.cs
stevenaw: Thanks. I have no idea what happened here

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
stevenaw: Sorry, maybe there's been a misunderstanding on my end. I like your suggestion as it avoids the need to explicitly check exception status. I'll push some changes shortly.

> However, in case of mismatch of type arguments, the test is marked as non-runnable as you don't have to run it at al.

I may still not be understanding this suggestion. Are you saying that mismatched values and generic types should fail earlier than test execution time? If so, I think I'm having a hard time reconciling that with "let the (dotnet) runtime sort it out" when `TypeArgs` are explicitly passed.

src/NUnitFramework/testdata/TestCaseAttributeFixture.cs
stevenaw: True, analyzers are a great validator for what would otherwise be runtime checks. What do you think about having a generic attribute to extend this on modern (NET8+) runtimes after we get the framework targeting that TFM?

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
manfred-brands: > Are you saying that mismatched values and generic types should fail earlier than test execution time

Yes. If NUnit cannot determine the TypeArgs. See [NUnitTestCaseBuilder](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs#L222-L223). If NUnit cannot determine the TypeArguments to use, the test is not runnable.

> "let the (dotnet) runtime sort it out" when TypeArgs are explicitly passed.

If `TypeArgs` is passed, the above `TryGetTypeArguments` is not called and NUnit first try to do some conversion in `TypeHelper.ConvertArgumentList` but after that passes it on to the runtime. The conversions done by NUnit are the ones normally done by the compiler when passing an `int` to a `double` parameter.

src/NUnitFramework/testdata/TestCaseAttributeFixture.cs
manfred-brands: Do we really need to wait for .NET8? It is a C#11 feature and it seems to work in .NET6.0 but fails in .NET Framework. with a runtime error: `Generic types are not valid`

```csharp
    /// <summary>
    /// Marks a method as a parameterized test suite and provides arguments for each test case.
    /// </summary>
    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true, Inherited = false)]
    public class TestCaseAttribute<T> : TestCaseAttribute
    {
        /// <summary>
        /// Construct a TestCaseAttribute with a list of arguments.
        /// </summary>
        public TestCaseAttribute(T argument)
            : base(new object?[] { argument })
        {
            TypeArgs = new[] { typeof(T) };
        }
    }
```

And equivalent version with 2, 3, 4, etc. parameters.

With a usage :

```
        [TestCase(2, TypeArgs = new[] { typeof(double) }, ExpectedResult = typeof(double))]
        [TestCase<double>(2, ExpectedResult = typeof(double))]
        public Type GenericMethodAndParameterWithExplicitOrImplicitTyping<T>(T input)
            => typeof(T);
```

manfred-brands (approved):
Changes and tests look fine now.
I'll leave it up to you if you want to add the Generic `TestCaseAttribute`s

stevenaw:
Thanks @manfred-brands !
That's great to hear it'll also work with NET 6. I'd thought it needed runtime support but happy to see that may not be the case.
If you're alright with it, I'd like to tackle generic attribute support in a separate issue rather than rushing into this one in a potentially half-baked manner.

I'm going to merge this now for release -related timing (FYI @OsirisTerje ) and will create those follow up issues later today

PR #4628, opened 2024-02-17
Make builds deterministic with ContinuousIntegrationBuild flag

lahma:
Currently [NuGet Package Explorer gives a warning](https://nuget.info/packages/NUnit/4.0.1) about missing compilation flags. This can be fixed by setting [ContinuousIntegrationBuild](https://cakebuild.net/api/Cake.Common.Tools.DotNet.MSBuild/DotNetMSBuildSettings/29A05B4C) to true. Also updated workflow definition to use new version of tasks to get rid of warnings during CI runs.

image

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -116,7 +116,7 @@ DotNetBuildSettings CreateDotNetBuildSettings() =>
         Configuration = configuration,
         NoRestore = true,
         Verbosity = DotNetVerbosity.Minimal,
-        MSBuildSettings = new DotNetMSBuildSettings { Version = packageVersion }
+        MSBuildSettings = new DotNetMSBuildSettings { Version = packageVersion, ContinuousIntegrationBuild = true }
      };
 
 //////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


lahma:
After the changes now complaints for download CI artifacts
image

OsirisTerje:
This flag should be set when build on the server, but not locally.  So it needs to be dropped in as a parameter, with default false.

build.cake
```diff
@@ -116,7 +116,7 @@ DotNetBuildSettings CreateDotNetBuildSettings() =>
         Configuration = configuration,
         NoRestore = true,
         Verbosity = DotNetVerbosity.Minimal,
-        MSBuildSettings = new DotNetMSBuildSettings { Version = packageVersion }
+        MSBuildSettings = new DotNetMSBuildSettings { Version = packageVersion, ContinuousIntegrationBuild = true }
```
OsirisTerje, line 119: This flag should be set when build on the server, but not locally. So it needs to be dropped in as a parameter, with default false.

OsirisTerje (changes requested):

build.cake
lahma: I changed the flag to be based on [BuildSystem.GitHubActions.IsRunningOnGitHubActions](https://cakebuild.net/api/Cake.Common.Build.GitHubActions/IGitHubActionsProvider/3F54CBE7)

OsirisTerje (approved):

OsirisTerje:
Thanks @lahma !

PR #4629, opened 2024-02-17
Updatepackages and fix warnings

OsirisTerje:


src/NUnitFramework/Directory.Packages.props
```diff
@@ -3,21 +3,18 @@
...
   <ItemGroup>
     <PackageVersion Include="System.Collections.Immutable" Version="6.0.0" />
     <PackageVersion Include="System.ValueTuple" version="4.5.0" />
   </ItemGroup>
-
   <!-- General Packages -->
   <ItemGroup>
-    <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="1.1.1" />
-    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.7.2" />
+    <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="8.0.0" />
```
mikkelbu, line 13: This seems like a major bump, but I guess there are no breaking changes?

src/NUnitFramework/tests/Attributes/CancelAfterTests.cs
```diff
@@ -294,7 +294,7 @@ public void TestWithCancelAfterAttributeAndTestCaseSourceHasCancellationToken(in
 #pragma warning restore NUnit1003 // The TestCaseAttribute provided too few arguments
 #pragma warning restore NUnit1027 // The test method has parameters, but no arguments are supplied by attributes
 #pragma warning restore NUnit1029 // The number of parameters provided by the TestCaseSource does not match the number of parameters in the Test method
-
+#pragma warning disable NUnit1029
```
mikkelbu, line 297: It seems odd to disable this when we have just restore it on the line above. Also, I would have thought that this work now?

mikkelbu (approved):
LGTM. Only a couple of comments/questions.

Also have you considered to place
```
  <ItemGroup>
    <PackageReference Update="Microsoft.CodeAnalysis.CSharp.CodeStyle">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
    </PackageReference>
  </ItemGroup>
```
in some Directory.Build.props, so we don't have to have the same definition in a lot of places. I've checked if the packages is included in every csproj.

Also the same comment for the NUnit.Analyzers - although it seems like it is used in different folders

src/NUnitFramework/benchmarks/nunit.framework.benchmarks/nunit.framework.benchmarks.csproj
```diff
@@ -19,4 +19,11 @@
       <ProjectReference Include="..\..\tests\nunit.framework.tests.csproj" />
     </ItemGroup>
 
+    <ItemGroup>
+      <PackageReference Update="Microsoft.CodeAnalysis.CSharp.CodeStyle">
+        <PrivateAssets>all</PrivateAssets>
+        <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      </PackageReference>
+    </ItemGroup>
+
```
manfred-brands, line 28: There is no need to add this at all as the package is already added in `Directory.Build.props`
Same hold for all other instances of this you added.

src/NUnitFramework/mock-assembly/mock-assembly.csproj
```diff
@@ -4,11 +4,20 @@
     <TargetFrameworks>$(NUnitRuntimeFrameworks)</TargetFrameworks>
     <RootNamespace>NUnit.Tests</RootNamespace>
     <IsTestProject>false</IsTestProject>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <NoWarn>1591</NoWarn>
```
manfred-brands, line 8: What is the reason for adding this? 
Is it so that the XML analyzer run as I don't see much use in creating XML docs for an nunit test.

src/NUnitFramework/nunit.framework.legacy.tests/AssertEqualsTests.cs
```diff
@@ -1,5 +1,7 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
+#pragma warning disable NUnit2005
+
```
manfred-brands, line 4: This is already excluded by the .editorconfig you added so manual pragma's are not needed.
We should also delete the pre-existing `#pragma` then from this file.

src/NUnitFramework/nunitlite.tests/nunitlite.tests.csproj
```diff
@@ -23,7 +25,17 @@
   </ItemGroup>
 
  <ItemGroup>
-    <PackageReference Include="NUnit.Analyzers" />
+    <PackageReference Include="NUnit.Analyzers">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
```
manfred-brands, line 31: This is not needed. The Analyzer package is marked as a `developmentDependency` which effectively [does the same](https://docs.microsoft.com/en-us/nuget/reference/nuspec#developmentdependency) without having to specify it.

manfred-brands (changes requested):
Most of your changes to the .csproj files are not needed.

src/NUnitFramework/benchmarks/nunit.framework.benchmarks/nunit.framework.benchmarks.csproj
lahma: For general purpose packages one could also use [Global Package References](https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management#global-package-references), which are automatically private assets.

src/NUnitFramework/Directory.Packages.props
OsirisTerje: Didn't look like no. And this is just for development ("There are no supported framework assets in this package.") , and we build using the 8 version.

src/NUnitFramework/tests/Attributes/CancelAfterTests.cs
OsirisTerje: No, it doesn't.  It still fails, and I thought to create an issue for it, but from your comment - was it supposedly fixed ?

OsirisTerje: but I'll delete 296 and 297

src/NUnitFramework/benchmarks/nunit.framework.benchmarks/nunit.framework.benchmarks.csproj
OsirisTerje: @manfred-brands Right you are.  I added them using the Nuget wizard, and didn't look closely there.  My bad.  Good you're awake :-)  

@lahma CPR is already there.

src/NUnitFramework/nunitlite.tests/nunitlite.tests.csproj
OsirisTerje: Awesome!

src/NUnitFramework/mock-assembly/mock-assembly.csproj
OsirisTerje: Yes, without that one we get a warning like the one below
image

There is a roslyn issue on this, and an explanation for the behaviour - which affects the IDE0005 (using statement analysis)  :  https://github.com/dotnet/roslyn/issues/41640#issuecomment-588403027

manfred-brands (approved):
There were more places where the NUnit.Analyzer reference was unnecessarily updated.

There are a few left-in suppressions that can be removed now that the analyzer package is updated. 

I have pushed some commit for you.

PR #4631, opened 2024-02-19
Publishing dev builds to myget.org

OsirisTerje:
I've created a build action to publish to myget.
I set the prefix for dev builds to `alpha` to get the versioning to behave well.  The suffix we had earlier using `dev` would be above a possible `beta`.  
Further, I used the `MinVer` client to get a reasonable numbering scheme.  

This build triggers on main, master, release, and v3.  I added the main now, because we have a task to move the master branch to main asap, I believe just after we have released 4.1 should be good. 

We can then use a copy of this builddef for publishing to nuget.org, but just set it without any triggers, so we do it manually when we feel ready.

There are now packages built from this in https://www.myget.org/feed/nunit/package/nuget/NUnit and https://www.myget.org/feed/nunit/package/nuget/NUnitLite


.github/workflows/NUnit.Myget.Publish.yml
```diff
@@ -0,0 +1,65 @@
+name: NUnit.Myget.Publish
+env:
+  MinVerAutoIncrement: minor
+on:
+  push:
+    branches:
+      - main
```
stevenaw, line 7: 👍 Thanks

Issue #4588, opened 2023-12-13
Assert.ThatAsync doesn't seem to work with ValueTask


OsirisTerje:
Twitter user eduardo @ed_dev commented about this, and showed the following repro code:

![image](https://github.com/nunit/nunit/assets/203432/b01b538a-78b7-4afa-82fc-fb7876f92069)

See twitter post https://twitter.com/ed_dev_/status/1730951076730323419

eduherminio:
Adding workaround as well:

```csharp
await Assert.ThatAsync(() => Method(), Is.EqualTo(""));   // Doesn't work, would work if method returned Task<string>
await Assert.ThatAsync(async () => await Method(), Is.EqualTo(""));  // Works
```

OsirisTerje:
Another workaround is 
```cs
await Assert.ThatAsync(()=>Method().AsTask(),Is.EqualTo(""));
```

jnm2:
I'm not sure where, but I'd thought we had an existing thread on ValueTask. ValueTask is just one of dozens of awaitable types other than Task. `() => Task.Yield()` is an example, and so is UWP's IAsyncOperation.

For all of these, `async () => await` seems quite tolerable. I still think it's good advice to add `async`/`await` even when it's not strictly necessary: <https://blog.stephencleary.com/2016/12/eliding-async-await.html>

manfred-brands:
After re-reading [Stephen Toubs article](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/) I also think `ValueTask` will become more in use and a specific overload could be useful. Due to the limitations described in the article it could be that we have to call `AsTask()` and then call the original API.

OsirisTerje:
@jnm2 We do have other issues on ValueTask.  You authored one back in 2018.  I'll link these together. 

But this one was very specific, so I thought it was a valid one to keep.

OsirisTerje:
@jnm2  Here are all the async issues, open and closed.  I have had them collected in a projected.   https://github.com/nunit/nunit/issues?q=is%3Aissue+project%3Anunit%2Fnunit%2F1+

PR #4640, opened 2024-02-26
Properly handle generic ValueTask await adapter

### LINKED ISSUE #4639: `ValueTask` is not being properly consumed by the `AwaitAdapter`
The current implementation assumes `ValueTask<>` is derived from `ValueTask`:

https://github.com/nunit/nunit/blob/1b91bdd8623d3d65ed4533eb0b4b8d3a988430ea/src/NUnitFramework/framework/Internal/ValueTaskAwaitAdapter.cs#L12-L28

This looks like a copy-paste from the `Task` version:
https://github.com/nunit/nunit/blob/1b91bdd8623d3d65ed4533eb0b4b8d3a988430ea/src/NUnitFramework/framework/Internal/TaskAwaitAdapter.cs#L12-L28

However, `ValueTask<>` is a struct and cannot inherit from `ValueTask`, so this code was not working. Fortunately, `ValueTask<>`s were getting picked up by the `CSharpPatternBasedAwaitAdapter`, so nothing was working wrong, but this should nonetheless be fixed.

RenderMichael:
Fixes #4639

### DIFF
--- a/src/NUnitFramework/framework/Internal/ValueTaskAwaitAdapter.cs
+++ b/src/NUnitFramework/framework/Internal/ValueTaskAwaitAdapter.cs
@@ -16,7 +16,7 @@ namespace NUnit.Framework.Internal
         public static AwaitAdapter? TryCreate(object task)
         {
             Type taskType = task.GetType();
-            if (taskType.GetGenericTypeDefinition() == typeof(ValueTask<>))
+            if (taskType.IsGenericType && taskType.GetGenericTypeDefinition() == typeof(ValueTask<>))
             {
                 return (AwaitAdapter)typeof(GenericAdapter<>)
                     .MakeGenericType(taskType.GetGenericArguments()[0])
--- a/src/NUnitFramework/framework/Internal/ValueTaskAwaitAdapter.cs
+++ b/src/NUnitFramework/framework/Internal/ValueTaskAwaitAdapter.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System;
-using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 
@@ -11,21 +10,21 @@ namespace NUnit.Framework.Internal
     {
         public static AwaitAdapter Create(ValueTask task)
         {
-            var genericValueTaskType = task
-                .GetType()
-                .TypeAndBaseTypes()
-                .FirstOrDefault(t => t.IsGenericType && t.GetGenericTypeDefinition() == typeof(ValueTask<>));
+            return new NonGenericAdapter(task);
+        }
 
-            if (genericValueTaskType is not null)
+        public static AwaitAdapter? TryCreate(object task)
+        {
+            Type taskType = task.GetType();
+            if (taskType.GetGenericTypeDefinition() == typeof(ValueTask<>))
             {
-                var typeArgument = genericValueTaskType.GetGenericArguments()[0];
                 return (AwaitAdapter)typeof(GenericAdapter<>)
-                     .MakeGenericType(typeArgument)
-                     .GetConstructor(new[] { typeof(ValueTask<>).MakeGenericType(typeArgument) })!
-                     .Invoke(new object[] { task });
+                    .MakeGenericType(taskType.GetGenericArguments()[0])
+                    .GetConstructor(new[] { taskType })!
+                    .Invoke(new object[] { task });
             }
 
-            return new NonGenericAdapter(task);
+            return null;
         }
 
         private sealed class NonGenericAdapter : AwaitAdapter
--- a/src/NUnitFramework/framework/Internal/AwaitAdapter.cs
+++ b/src/NUnitFramework/framework/Internal/AwaitAdapter.cs
@@ -43,9 +43,9 @@ namespace NUnit.Framework.Internal
             if (awaitable is System.Threading.Tasks.ValueTask valueTask)
                 return ValueTaskAwaitAdapter.Create(valueTask);
 
-            // Await all the (C# and F#) things
-            var adapter =
-                CSharpPatternBasedAwaitAdapter.TryCreate(awaitable)
+            var adapter = ValueTaskAwaitAdapter.TryCreate(awaitable)
+                // Await all the (C# and F#) things
+                ?? CSharpPatternBasedAwaitAdapter.TryCreate(awaitable)
                 ?? FSharpAsyncAwaitAdapter.TryCreate(awaitable);
             if (adapter is not null)
                 return adapter;
--- a/src/NUnitFramework/testdata/AwaitableReturnTypeFixture.cs
+++ b/src/NUnitFramework/testdata/AwaitableReturnTypeFixture.cs
@@ -194,6 +194,34 @@ namespace NUnit.TestData
             }
         }
 
+        public ValueTask ReturnsValueTask()
+        {
+            _workload.BeforeReturningAwaitable();
+            _workload.BeforeReturningAwaiter();
+
+            var source = new TaskCompletionSource<object?>();
+
+            var complete = new Action(() =>
+            {
+                try
+                {
+                    _workload.GetResult();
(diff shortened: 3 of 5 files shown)


manfred-brands (approved):
Thanks for doing this.
Looks good.

manfred-brands:
@RenderMichael Any idea why the tests after 3 hours are still running?

RenderMichael:
Not sure why it manifested as a timeout, but there's a bunch of behavior which seemed to rely on the old implementation, in other works for `ValueTask<>` to be handled by `CSharpPatternBasedAwaitAdapter`.
image

I'll look deeper into this later today.

RenderMichael:
Ah never mind, I was completely misinterpreting those results. `taskType.GetGenericTypeDefinition()` was throwing for non-generic types, I forgot to put a `taskType.IsGenericType` before it. Fixed, the failing test passes for me locally.

I guess the failure was at an unfortunate location where it caused an infinite wait instead of an exception bubbling up.

OsirisTerje:
Good there are tests then!  

@RenderMichael Btw, are there enough tests to cover the changes done here ?

src/NUnitFramework/framework/Internal/ValueTaskAwaitAdapter.cs
```diff
@@ -11,21 +10,21 @@ internal static class ValueTaskAwaitAdapter
     {
         public static AwaitAdapter Create(ValueTask task)
         {
-            var genericValueTaskType = task
-                .GetType()
-                .TypeAndBaseTypes()
```
OsirisTerje, line 16: Why is it that this is no longer needed ?

OsirisTerje:
This is the coverage of the current code for the ValueTaskAwaitAdapter, and the GenericAdapter is not part of any tests.  ANd with the change in the PR, this part of the code is being changed.  Shouldn't we have a unit test around this code?  
image

src/NUnitFramework/framework/Internal/ValueTaskAwaitAdapter.cs
RenderMichael: This was originally taken from the Task await adapter, where for the generic task this method would return [Task<>, Task, object]. This was probably to get Task<> and Task in the same sequence, though in my opinion it was a bit excessive to begin with.

For ValueTask however, it doesn’t work to begin with, because ValueTask<> does not inherit from ValueTask, since they’re structs. So the handling has to be separate.

RenderMichael:
@OsirisTerje I’m happy to add a unit test here, note that while there’s no explicit test coverage, as you can see from the CI failures, this was implicitly tested by the rest of the suite.

OsirisTerje:
@jnm2 wrote the AsyncAdapters back in 2018, and he added a series of tests,  AsyncExecutionApiAdapterTests, AwaitableReturnTypeTests, and more - search for "await" in the test explorer.  @jnm2 might explain more here, but I got a feeling we should have added more tests here when ValueTask was introduced/changed.  I also see we have issue #4588 that is related to ValueTask.  And the old #2774 is still open.  

The issues related to ValueTask are old, so this is not new, but when we start changing things, that's when I at least start to be a bit worried :-) 

About implicitly tests, I am not sure if that was implicit or a side-effect, if it was implicit, it should have shown up in the coverage.  

@stevenaw @manfred-brands Comments ?

OsirisTerje (approved):
Thanks!   LGTM :-)

stevenaw (approved):
Thanks for noticing and fixing this @RenderMichael ! LGTM

jnm2:
I apologize for missing https://github.com/nunit/nunit/pull/4543. What is the reason that we don't use CSharpPatternBasedAwaitAdapter for ValueTask?

RenderMichael:
> What is the reason that we don't use CSharpPatternBasedAwaitAdapter for ValueTask

<details>

``` csharp

using NUnit.TestData;

namespace NUnit.Framework;

[MemoryDiagnoser(false)]
public class AwaitableBenchmarks
{
    public IEnumerable<object> ValueData()
    {
        yield return Task.Yield();
        yield return ValueTask.FromResult(30);
        yield return ValueTask.FromException<object>(new Exception());
        yield return new AwaitableReturnTypeFixture.CustomAwaitable();
        yield return new AwaitableReturnTypeFixture.CustomAwaitableWithImplicitOnCompleted();
        yield return new AwaitableReturnTypeFixture.CustomAwaitableWithImplicitUnsafeOnCompleted();
    }

    [Benchmark]
    [ArgumentsSource(nameof(ValueData))]
    public AwaitAdapter? After(object maybeTask)
    {
        var awaiter = ValueTaskAwaitAdapter.TryCreate(maybeTask)
            ?? CSharpPatternBasedAwaitAdapter.TryCreate(maybeTask)
            ?? FSharpAsyncAwaitAdapter.TryCreate(maybeTask);

        if (awaiter is not null)
        {
            return awaiter;
        }

        throw new InvalidOperationException();
    }

    [Benchmark(Baseline = true)]
    [ArgumentsSource(nameof(ValueData))]
    public AwaitAdapter? Before(object maybeTask)
    {
        var awaiter = CSharpPatternBasedAwaitAdapter.TryCreate(maybeTask)
            ?? FSharpAsyncAwaitAdapter.TryCreate(maybeTask);

        if (awaiter is not null)
        {
            return awaiter;
        }

        throw new InvalidOperationException();
    }
}
```

</details>

| Method |            maybeTask |     Mean |    Error |   StdDev | Ratio | RatioSD | Allocated | Alloc Ratio |
|------- |--------------------- |---------:|---------:|---------:|------:|--------:|----------:|------------:|
|  After |                      | 634.9 ns | 12.32 ns | 16.86 ns |  0.65 |    0.04 |     296 B |        1.32 |
| Before |                      | 964.9 ns | 19.30 ns | 44.34 ns |  1.00 |    0.00 |     224 B |        1.00 |
|        |                      |          |          |          |       |         |           |             |
|  After |                   30 | 596.2 ns | 11.38 ns | 17.37 ns |  0.59 |    0.03 |     288 B |        1.33 |
| Before |                   30 | 993.7 ns | 19.83 ns | 54.28 ns |  1.00 |    0.00 |     216 B |        1.00 |
|        |                      |          |          |          |       |         |           |             |
|  After | NUnit(...)table [57] | 938.2 ns | 18.77 ns | 28.66 ns |  1.02 |    0.06 |     240 B |        1.00 |
| Before | NUnit(...)table [57] | 924.4 ns | 18.22 ns | 32.39 ns |  1.00 |    0.00 |     240 B |        1.00 |
|        |                      |          |          |          |       |         |           |             |
|  After | NUnit(...)leted [80] | 926.4 ns | 10.37 ns |  8.66 ns |  0.98 |    0.03 |     240 B |        1.00 |
| Before | NUnit(...)leted [80] | 945.7 ns | 18.42 ns | 24.59 ns |  1.00 |    0.00 |     240 B |        1.00 |
|        |                      |          |          |          |       |         |           |             |
|  After | NUnit(...)leted [86] | 942.3 ns | 18.42 ns | 22.62 ns |  1.03 |    0.03 |     240 B |        1.00 |
| Before | NUnit(...)leted [86] | 916.5 ns | 17.82 ns | 19.81 ns |  1.00 |    0.00 |     240 B |        1.00 |
|        |                      |          |          |          |       |         |           |             |
|  After | Syste(...)table [46] | 908.2 ns | 17.76 ns | 24.31 ns |  1.03 |    0.03 |     208 B |        1.00 |
| Before | Syste(...)table [46] | 884.8 ns | 17.68 ns | 23.60 ns |  1.00 |    0.00 |     208 B |        1.00 |

I'm not versed enough in BenchmarkDotNet to give a better parameter name than the default `ToString()` it gives for `maybeTask`, but the first two cases are `ValueTask<>`s and the rest are custom awaiters. Note that this check happens after we check for `Task`, `Task<TResult>`, any task derived from `Task`, as well as the non-generic `ValueTask` (so in practice, I can't imagine the "custom awaiter test fixture" scenario happens very often)

I can't speak for why it was initially implemented, but it noticeably speeds up the ValueTask<> check (in exchange for higher allocations) without meaningfully regressing the other cases.

Issue #4598, opened 2024-01-11
DefaultTimeout in .runsettings + TearDown method seems to break test output


patrykpiotrmarek:
NUnit v4.0.1 (regression after upgrade from 3.x)

I have simple test class with `TestContext.WriteLine` in test method and `TearDown` method.
```
[TestFixture]
public class Tests
{
    [Test]
    public void Test1()
    {
        TestContext.WriteLine("line1");
        Assert.That(1, Is.EqualTo(0));
    }

    [TearDown]
    public void TearDown()
    {
    }
}
```
and .runsettings file with `DefaultTimeout` setting
```
<?xml version="1.0" encoding="utf-8"?>
<RunSettings>
  <NUnit>
    <DefaultTimeout>600000</DefaultTimeout>
  </NUnit>
</RunSettings>
```
When I run this code from dotnet test with `--settings` parameter, there is no output message other than assertion. 
```
PS C:\NUnitTest\NUnitTest\bin\Debug\net8.0> dotnet test NUnitTest.dll --settings .\nunit.runsettings
Microsoft (R) Test Execution Command Line Tool Version 17.8.0 (x64)
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
  Failed Test1 [20 ms]
  Error Message:
     Assert.That(1, Is.EqualTo(0))
  Expected: 0
  But was:  1

  Stack Trace:
     at NUnitTest.Tests.Test1() in C:\NUnitTest\NUnitTest\Tests.cs:line 12



Failed!  - Failed:     1, Passed:     0, Skipped:     0, Total:     1, Duration: 19 ms - NUnitTest.dll (net8.0)
```
Everything works fine after I remove `TearDown` method or `DefaultTimeout` setting
```
PS C:\NUnitTest\NUnitTest\bin\Debug\net8.0> dotnet test NUnitTest.dll --settings .\nunit.runsettings
Microsoft (R) Test Execution Command Line Tool Version 17.8.0 (x64)
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
line1

  Failed Test1 [17 ms]
  Error Message:
     Assert.That(1, Is.EqualTo(0))
  Expected: 0
  But was:  1

  Stack Trace:
     at NUnitTest.Tests.Test1() in C:\NUnitTest\NUnitTest\Tests.cs:line 12

[...]

PR #4646, opened 2024-03-02
WIP: Keep output writer when Timeout on modern .NET

stevenaw:
Relates to #4598 
Initial POC to try to get output printing when combining `Timeout` with `Teardown`

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
@@ -119,11 +119,9 @@ namespace NUnit.Framework.Internal.Commands
 
         private Task<TestResult> RunTestOnSeparateThread(TestExecutionContext context)
         {
-            var separateContext = new TestExecutionContext(context)
-            {
-                CurrentResult = context.CurrentTest.MakeTestResult()
-            };
-            return Task.Run(() => innerCommand.Execute(separateContext));
+            var separateContext = new TestExecutionContext(context);
+
+            return Task.Run(() => innerCommand.Execute(context));
         }
 #endif
     }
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -121,8 +121,9 @@ private Task<TestResult> RunTestOnSeparateThread(TestExecutionContext context)
         {
             var separateContext = new TestExecutionContext(context)
             {
-                CurrentResult = context.CurrentTest.MakeTestResult()
+                //CurrentResult = context.CurrentTest.MakeTestResult()
```
stevenaw, line 124: This resets the `OutWriter` as the new test result  will create it's own writer, which when assigned to `separateContext.CurrentResult` will then apply it to the `separateContext`'s outwriter based on https://github.com/nunit/nunit/blob/321db1ae2b67de0049ee9f534518369786d9b557/src/NUnitFramework/framework/Internal/TestExecutionContext.cs#L208-L217

stevenaw: @manfred-brands I'd like to pick up fixing this issue, but I'm looking for advice here. Do you recall why we create a new `CurrentResult` in the `TimeoutCommand`?

If we need it then I think the way to fix the issue here is to pass in `context.OutWriter` when creating it to avoid creating new one on initialization. That's a bit involved to do in a non-breaking way so I wanted to double-check first about what the commented out line here does

stevenaw:
Having explored a bit more, I'm feeling fairly confident that we can delete the line, though I'd welcome feedback on the approach. For now I'm going to proceed and try to get a test written for this

PR #4647, opened 2024-03-03
Don't generate output content twice in TestResult.AddToXml

lahma:
Extract the essential part from #4460 and skipping generating the output twice. Also checking against backing field `_assertionResults` instead of of interface-returning get-property to remove interface dispatch and enumerator boxing.


mikkelbu (approved):
LGTM. Thanks for the contribution @lahma 👍

PR #4648, opened 2024-03-03
Switch to using macos-14 instead of macos-latest for CI

### LINKED ISSUE #4381: MacOS build really slow for some tests
I had several PRs failing on MacOs with:

image

As the comment above the timeouts says the tests are much faster locally:

```
        // The following tests are each running in 14ms to 46ms on my machine. Based on that,
        // warn at 100ms and fail at 500ms
```

It this because the MacOS runner is underspec'd?

Should we increase the timeout, but then we won't detect deterioration under other OS?
Should we disable the test on MacOS as these are pure performance tests?

The warning time doesn't do much good as there is so much output of our test run that it will get lost and nobody notices.

lahma:
[github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/)

[And it's fast ](https://github.com/sebastienros/jint/pull/1769#issuecomment-1918470696)

Needed to install NET 6 SDK specifically as only NET 7 and NET 8 come OOTB. Setting CI friendly values for [DOTNET_NOLOGO](DOTNET_NOLOGO) and [DOTNET_CLI_TELEMETRY_OPTOUT](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_cli_telemetry_optout) (true).

Fixes https://github.com/nunit/nunit/issues/4649
fixes #4381 (hopefully) as MacOS build is now pretty much as fast as Ubuntu build:

image

### DIFF
--- a/.github/workflows/NUnit.CI.yml
+++ b/.github/workflows/NUnit.CI.yml
@@ -8,6 +8,10 @@ on:
       - release
       - 'v3'
   pull_request:
+  
+env:
+  DOTNET_NOLOGO: true                     # Disable the .NET logo
+  DOTNET_CLI_TELEMETRY_OPTOUT: true       # Disable sending .NET CLI telemetry  
 
 jobs:
   build-windows:
@@ -81,7 +85,7 @@ jobs:
 
   build-macos:
     name: MacOS Build
-    runs-on: macos-latest
+    runs-on: macos-14
 
     steps:
     - name: ⤵️ Checkout Source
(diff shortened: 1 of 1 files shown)


stevenaw:
Nice find, thanks @lahma ! Impressive to see the difference in macos timing between this PR and others where it's consistently over 5 minutes.

I'll try and give it a proper look over later today

stevenaw (approved):
LGTM thanks once again @lahma !

PR #4652, opened 2024-03-05
WIP: Additional Test Events for OneTimeSetUp / OneTimeTearDown

z002Holpp:
We implemented a first draft of a solution that provides additional test events for OneTimeSetUp and OneTimeTearDown.
The main uncertainty during the implementation was to find the right place where the events should be raised.

Please find more information in the corresponding issue.
This PR should serve more as a base for a technical discussion rather than being "production-ready".


src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -323,6 +323,10 @@ private TestExecutionContext CreateTestExecutionContext(ITest loadedTest, ITestL
             // Set the listener - overriding runners may replace this
             context.Listener = listener;
 
+            // Set the listener for extended test events like OneTimeSetUp / OneTimeTearDown
+            if (listener is ITestListenerExt testListenerExt)
```
OsirisTerje, line 327: Mixing behaviour.  Avoid this.

src/NUnitFramework/framework/Internal/Commands/SetUpTearDownItem.cs
```diff
@@ -92,11 +92,35 @@ private void RunSetUpOrTearDownMethod(TestExecutionContext context, IMethodInfo
...
+            }
+            finally
+            {
+                RaiseFinishedEvent(context, methodInfo);
+            }
+        }
+
+        private static void RaiseStartedEvent(TestExecutionContext context, MethodInfoCache.TestMethodMetadata methodInfo)
+        {
+            if (methodInfo.HasOneTimeSetUpAttribute)
```
OsirisTerje, line 112: Why do you check this?  You should not be here if you were not in a Setup/Teardown method.

I assume you try to separate between pure OneTimeSetup/TD and the test Setup/TD.  That is checked earlier.  It might be necessary to state that information inside the SetUpTearDownItem.

src/NUnitFramework/framework/Internal/Commands/SetUpTearDownItem.cs
```diff
@@ -92,11 +92,35 @@ private void RunSetUpOrTearDownMethod(TestExecutionContext context, IMethodInfo
...
+                context.ListenerExt.OneTimeSetUpStarted(context.CurrentTest);
+            else if (methodInfo.HasOneTimeTearDownAttribute)
+                context.ListenerExt.OneTimeTearDownStarted(context.CurrentTest);
+        }
+
```
OsirisTerje, line 117: Do you need this if inside here, you know when you call what you have, you can just pass that along, or even better drop this method all together and just call in in-place.

src/NUnitFramework/framework/Internal/Commands/SetUpTearDownItem.cs
```diff
@@ -92,11 +92,35 @@ private void RunSetUpOrTearDownMethod(TestExecutionContext context, IMethodInfo
...
+        private static void RaiseFinishedEvent(TestExecutionContext context, MethodInfoCache.TestMethodMetadata methodInfo)
+        {
+            if (methodInfo.HasOneTimeSetUpAttribute)
+                context.ListenerExt.OneTimeSetUpFinished(context.CurrentTest);
```
OsirisTerje, line 121: The ListenerExt may not exist, thus be null, and if it is null, it should not be called. 
   context.ListenerExt?.OneTimeSetUpFinished(context.CurrentTest);

OsirisTerje (changes requested):

src/NUnitFramework/framework/Internal/Commands/SetUpTearDownItem.cs
z002Holpp: Right. Updated in latest commit.
Changed to distinguish OneTimeSetUp and SetUp with the TestContext.

src/NUnitFramework/framework/Internal/Commands/SetUpTearDownItem.cs
z002Holpp: Right. Updated in latest commit.

src/NUnitFramework/framework/Internal/Commands/SetUpTearDownItem.cs
z002Holpp: Right. Updated in latest commit.

z002Holpp:
Will close the PR since we will continue with another, more leightweight proposal in https://github.com/nunit/nunit/issues/4744

PR #4663, opened 2024-03-16
Add support for Exceptions for Display Name

### LINKED ISSUE #4659: TestCaseSource that contains Exception with InnerException - not running tests
After updating NUnit 3 to NUnit 4 (4.1.0) my test cases with InnerException are not run. I'm using TestAdapter 4.5.0
```
public class Tests1
{
    [TestCaseSource(nameof(GetSomeExceptionCases))]
    public void NoMethod_GivenException_ShouldPassTest(Exception testCase) => testCase.Should().NotBeOfType<string>();

    public static TestCaseData[] GetSomeExceptionCases() =>
    [
        new TestCaseData(new ArgumentNullException()),
        new TestCaseData(new ArgumentNullException("message", new Exception())),
    ];
}
```
image

When I downgrade NUnit to 3.14.0, both tests are passing.
image

manfred-brands:
Fixes #4659 

There should probably be a more Generic filter in removing characters that VS Test Explorer cannot handle.

### DIFF
--- a/src/NUnitFramework/framework/Internal/DisplayName.cs
+++ b/src/NUnitFramework/framework/Internal/DisplayName.cs
@@ -215,8 +215,31 @@ namespace NUnit.Framework.Internal
                 else if (sb.Equals(sbyte.MinValue))
                     display = "sbyte.MinValue";
             }
+            else if (arg is Exception e)
+            {
+                display = FormatException(e);
+            }
 
             return display;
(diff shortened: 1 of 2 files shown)


mikkelbu (approved):
LGTM - and I agree with you comment on the PR that this could/should be extended to a more generic method at some point

Issue #4662, opened 2024-03-14
StringAssert.IsNullOrEmpty overloads


MgSam:
I like the `StringAssert` class because it provides more informative error messages over just doing something like `Assert.True()`. I'd like to suggest adding a few overloads to the `StringAssert` class:

```C#
StringAssert.IsNullOrEmpty()
StringAssert.IsNullOrWhiteSpace()
StringAssert.IsNotNullOrEmpty()
StringAssert.IsNotNullOrWhiteSpace()
```

These are very common things to want to check and right now the best way to do it is with `Assert.True()` which does not provide informative error messages by default.

OsirisTerje:
The StringAsserts are part of the legacy/classic assertions.  Although we don't expect to do anything more with these, what you suggest can be added, as they seem to be just additions, causing no breaking changes or changes to existing assertions.

All the StringAssert methods are implemented in terms of the constraint model, like:

```csharp
 public static void StartsWith(string expected, string actual, string message, params object?[]? args)
 {
      Assert.That(actual, Does.StartWith(expected), () => ConvertMessageWithArgs(message, args));
 }
```

And this is the way your suggestions should be implemented too.

Suggestion 1 and 3 can easily be added, as they are just simple translations like: 

```csharp
Assert.That(somestring,Is.Not.Null.And.Not.Empty);
```

PS:  Have you considered using the constraint syntax instead? 

No 2 and 4 include whitespace, and we currently don't have a constraint for whitespace.  (We could probably have that, as it is frequently used in code. ). A solution would be to use the Regex constraint. 

We do have an open issue on this, see #3918.  Perhaps due time to fix this  @stevenaw @manfred-brands ? 

The team will probably not do this (except for creating a whitespace constraint or modifier at some time), but we are very open to receive a Pull Request for StringAsserts.  That PR also has to contain unit tests for the asserts. It should be easy to add, and placed along with the other string assert tests.  

Since this adds functionality it has to go along with the same assertions being documented in the [docs repo](https://github.com/nunit/docs), under the section for [StringAssert](https://docs.nunit.org/articles/nunit/writing-tests/assertions/classic-assertions/String-Assert.html).

manfred-brands:
@OsirisTerje I wouldn't mind adding a `Whitespace` constraint and also implement #3918. 
I currently work around that issue using `Assert.That(NoSpaces(json), Is.EqualTo(NoSpaces(expectedJson))`

OsirisTerje:
[...]

PR #4664, opened 2024-03-16
Issue3918 white space

### LINKED ISSUE #3918: String comparison without whitespace
In some situations a string comparison without taking whitespace into account is desirable.

A use-case would be JSON string comparison, which should be whitespace agnostic but requires perfect whitespace matching when using the `EqualConstraint`.

# Current solution to the issue

To avoid this issue I have to write

```csharp
Assert.That(resultingJsonString.WithoutWhitespace(), Is.EqualTo(expectedJsonString.WithoutWhitespace()));
```

where `WithoutWhitespace()` is an extension method removing all whitespace characters.

# Prefered syntax

It would be more fluent to write

```csharp
Assert.That(resultingJsonString, Is.EqualTo(expectedJsonString).IgnoreWhitespace));
```

`IgnoreWhitespace` was inspired by the existing `IgnoreCase` configuration.

manfred-brands:
Fixes #3918 

Adds preparation work for #4662 

There are two new constraints:

One to match String.IsNullOrWhiteSpace . This I modeled on 'Empty':

```csharp
Assert.That(<s>, Is.WhiteSpace)
Assert.That(<s>, Is.Not.WhiteSpace)
```

The other is a modifier on existing constrains, which I modeled on `IgnoreCase`:

```csharp
Assert.That("Hello World"), Is.EqualTo("HelloWorld").IgnoreWhiteSpace)
```

Adding the constraint was easy, to get sensible error messages in case of failure where strings have different white space was harder. In case of mis-matched white-space, the message now contains two carrots, one for `expected` and one for `actual`:

```
Expected: "abc def", ignoring white-space
-----------------^
But was:  "a b c d e g"
---------------------^
```

### DIFF
--- a/src/NUnitFramework/framework/Constraints/MessageWriter.cs
+++ b/src/NUnitFramework/framework/Constraints/MessageWriter.cs
@@ -81,10 +81,12 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         /// <param name="expected">The expected string value</param>
         /// <param name="actual">The actual string value</param>
-        /// <param name="mismatch">The point at which the strings don't match or -1</param>
+        /// <param name="mismatchExpected">The point in <paramref name="expected"/> at which the strings don't match or -1</param>
+        /// <param name="mismatchActual">The point in <paramref name="actual"/> at which the strings don't match or -1</param>
         /// <param name="ignoreCase">If true, case is ignored in locating the point where the strings differ</param>
+        /// <param name="ignoreWhiteSpace">If true, white space is ignored in locating the point where the strings differ</param>
         /// <param name="clipping">If true, the strings should be clipped to fit the line</param>
-        public abstract void DisplayStringDifferences(string expected, string actual, int mismatch, bool ignoreCase, bool clipping);
+        public abstract void DisplayStringDifferences(string expected, string actual, int mismatchExpected, int mismatchActual, bool ignoreCase, bool ignoreWhiteSpace, bool clipping);
 
         /// <summary>
         /// Writes the text for an actual value.
--- a/src/NUnitFramework/framework/Constraints/MessageWriter.cs
+++ b/src/NUnitFramework/framework/Constraints/MessageWriter.cs
@@ -2,6 +2,7 @@
 
 using System.IO;
 using System.Collections;
+using System;
 
 namespace NUnit.Framework.Constraints
 {
@@ -81,12 +82,32 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         /// <param name="expected">The expected string value</param>
         /// <param name="actual">The actual string value</param>
+        /// <param name="mismatch">The point in <paramref name="expected"/> at which the strings don't match or -1</param>
+        /// <param name="ignoreCase">If true, case is ignored in locating the point where the strings differ</param>
+        /// <param name="clipping">If true, the strings should be clipped to fit the line</param>
+        public abstract void DisplayStringDifferences(string expected, string actual, int mismatch, bool ignoreCase, bool clipping);
+
+        /// <summary>
+        /// Display the expected and actual string values on separate lines.
+        /// If the mismatch parameter is >=0, an additional line is displayed
+        /// line containing a caret that points to the mismatch point.
+        /// </summary>
+        /// <param name="expected">The expected string value</param>
+        /// <param name="actual">The actual string value</param>
         /// <param name="mismatchExpected">The point in <paramref name="expected"/> at which the strings don't match or -1</param>
         /// <param name="mismatchActual">The point in <paramref name="actual"/> at which the strings don't match or -1</param>
         /// <param name="ignoreCase">If true, case is ignored in locating the point where the strings differ</param>
         /// <param name="ignoreWhiteSpace">If true, white space is ignored in locating the point where the strings differ</param>
         /// <param name="clipping">If true, the strings should be clipped to fit the line</param>
-        public abstract void DisplayStringDifferences(string expected, string actual, int mismatchExpected, int mismatchActual, bool ignoreCase, bool ignoreWhiteSpace, bool clipping);
+        public virtual void DisplayStringDifferences(string expected, string actual, int mismatchExpected, int mismatchActual, bool ignoreCase, bool ignoreWhiteSpace, bool clipping)
+        {
+            if (ignoreWhiteSpace && mismatchExpected != mismatchActual)
+            {
+                throw new NotImplementedException("Please override to show difference with 'ignoreWhiteSpace'");
+            }
+
+            DisplayStringDifferences(expected, actual, mismatchExpected, ignoreCase, clipping);
+        }
 
         /// <summary>
         /// Writes the text for an actual value.
--- a/src/NUnitFramework/framework/Constraints/MsgUtils.cs
+++ b/src/NUnitFramework/framework/Constraints/MsgUtils.cs
@@ -395,64 +395,83 @@ namespace NUnit.Framework.Constraints
         [return: NotNullIfNotNull("s")]
         public static string? EscapeControlChars(string? s)
         {
-            if (s is not null)
+            int index = 0;
+            return EscapeControlChars(s, ref index);
+        }
+
+        /// <summary>
+        /// Converts any control characters in a string
+        /// to their escaped representation.
+        /// </summary>
+        /// <param name="s">The string to be converted</param>
+        /// <param name="index">The index in the array of a specific spot, which needs to be updated when expanding.</param>
+        /// <returns>The converted string</returns>
+        [return: NotNullIfNotNull("s")]
+        public static string? EscapeControlChars(string? s, ref int index)
+        {
+            if (s is null)
+                return null;
+
+            int originalIndex = index;
+            StringBuilder sb = new(s.Length + 32);
+
+            for (int i = 0; i < s.Length; i++)
             {
-                StringBuilder sb = new StringBuilder();
-
-                foreach (char c in s)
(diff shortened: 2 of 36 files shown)


OsirisTerje:
Awesome!

OsirisTerje (approved):
This is awesome work!   Really like the changes in ConstraintExpression to the Append method, making it generic fix the typecasting that is all over the other constraints.  Very good!

manfred-brands:
@OsirisTerje The change to `Append` is an breaking change as it changes the signature of a public method in a public class.
I managed to keep the original and add the new one. Luckily the compiler is smart enough to distinguish.

Whilst updating the documentation, I noticed an issue with the reporting when texts are too long.
Put PR in draft until I sorted this out.

manfred-brands:
All issues fixed, PR ready.

src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
```diff
@@ -164,41 +164,43 @@ public override void DisplayDifferences(object? expected, object? actual, Tolera
...
             Write(MsgUtils.FormatValue(expected));
             if (ignoreCase)
                 Write(", ignoring case");
+            if (ignoreWhiteSpace)
+                Write(", ignoring white-space");
             WriteLine();
+            if (mismatchExpected >= 0 && mismatchExpected != mismatchActual)
+                WriteCaretLine(mismatchExpected);
             WriteActualLine(actual);
             //DisplayDifferences(expected, actual);
```
mikkelbu, line 201: Should we remove this line?

src/NUnitFramework/tests/Constraints/MsgUtilTests.cs
```diff
@@ -307,9 +315,9 @@ public static void EscapesNullControlChars()
         private const string S52 = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
 
         [TestCase(S52, 52, 0, S52, TestName = "NoClippingNeeded")]
-        [TestCase(S52, 29, 0, "abcdefghijklmnopqrstuvwxyz...", TestName = "ClipAtEnd")]
-        [TestCase(S52, 29, 26, "...ABCDEFGHIJKLMNOPQRSTUVWXYZ", TestName = "ClipAtStart")]
-        [TestCase(S52, 28, 26, "...ABCDEFGHIJKLMNOPQRSTUV...", TestName = "ClipAtStartAndEnd")]
+        [TestCase(S52, 26, 0, "abcdefghijklmnopqrstuvwxyz...", TestName = "ClipAtEnd")]
+        [TestCase(S52, 26, 26, "...ABCDEFGHIJKLMNOPQRSTUVWXYZ", TestName = "ClipAtStart")]
+        [TestCase(S52, 22, 26, "...ABCDEFGHIJKLMNOPQRSTUV...", TestName = "ClipAtStartAndEnd")]
```
mikkelbu, line 320: Is this related to you comment on the PR (pasted below)? If not then what is this change about?

> Whilst updating the documentation, I noticed an issue with the reporting when texts are too long.

src/NUnitFramework/framework/Constraints/MsgUtils.cs
```diff
@@ -395,64 +395,83 @@ public static string GetTypeRepresentation(object obj)
...
+            if (s is null)
+                return null;
 
-                        case '\x0085':
-                        case '\x2028':
-                        case '\x2029':
-                            sb.Append($"\\x{(int)c:X4}");
-                            break;
+            int originalIndex = index;
+            StringBuilder sb = new(s.Length + 32);
```
mikkelbu, line 416: Why 32 ?

src/NUnitFramework/framework/Constraints/MsgUtils.cs
```diff
@@ -536,99 +555,169 @@ public static int[] GetArrayIndicesFromCollectionIndex(IEnumerable collection, l
...
         /// <param name="s">The string to be clipped</param>
-        /// <param name="maxStringLength">The maximum permitted length of the result string</param>
+        /// <param name="clipLength">The length of the clipped string</param>
         /// <param name="clipStart">The point at which to start clipping</param>
         /// <returns>The clipped string</returns>
-        public static string ClipString(string s, int maxStringLength, int clipStart)
+        public static string ClipString(string s, int clipLength, int clipStart)
         {
-            int clipLength = maxStringLength;
             StringBuilder sb = new StringBuilder();
```
mikkelbu, line 563: Should we set this to be of a certain size - e.g. `s.length` + 6?

mikkelbu (approved):
Looks good to me. I mostly have minor questions and/or comments

src/NUnitFramework/framework/Constraints/MessageWriter.cs
```diff
@@ -81,10 +81,12 @@ public void WriteMessageLine(string message, params object?[]? args)
...
         /// <param name="expected">The expected string value</param>
         /// <param name="actual">The actual string value</param>
-        /// <param name="mismatch">The point at which the strings don't match or -1</param>
+        /// <param name="mismatchExpected">The point in <paramref name="expected"/> at which the strings don't match or -1</param>
+        /// <param name="mismatchActual">The point in <paramref name="actual"/> at which the strings don't match or -1</param>
         /// <param name="ignoreCase">If true, case is ignored in locating the point where the strings differ</param>
+        /// <param name="ignoreWhiteSpace">If true, white space is ignored in locating the point where the strings differ</param>
         /// <param name="clipping">If true, the strings should be clipped to fit the line</param>
-        public abstract void DisplayStringDifferences(string expected, string actual, int mismatch, bool ignoreCase, bool clipping);
+        public abstract void DisplayStringDifferences(string expected, string actual, int mismatchExpected, int mismatchActual, bool ignoreCase, bool ignoreWhiteSpace, bool clipping);
```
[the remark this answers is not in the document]
manfred-brands: Yes it would be. Thanks for noticing!

The problem then is that if I leave the original signature, I cannot do the message improvements.
Even adding a new `abstract` method would be breaking.
I think adding a a new `virtual` method it would be fine as it can be resolved without an implementation having to implement it. The default implementation would call into the original method unless:
```csharp
if (ignoreWhiteSpace && mismatchExpected != mismatchActual)
{
    throw new NotImplementedException("Please override to show difference with 'ignoreWhiteSpace'");
}
```

src/NUnitFramework/tests/Constraints/EqualConstraintTests.cs
```diff
@@ -65,6 +65,59 @@ public void DoesntRespectCultureWhenCasingMatters()
...
+                    "surname": "Smith"
+                  },
+                  {
+                    "name": "Jane",
+                    "surname": Doe"
+                  }
+                ]
+                """;
+            const string condensedJson = """
+                "persons":[{"name":"John","surname":"Smith"},{"name": "Jane","surname": Doe"}]
```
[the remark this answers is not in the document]
manfred-brands: No, it wasn't. I only noticed this myself when I had copied this example into the documentation and the strings were not formatted as expected. Initially I blamed it on the new `"""` string syntax until I spotted the missing `"`.

`"` added in two places.

src/NUnitFramework/framework/Internal/Execution/TextMessageWriter.cs
manfred-brands: I presume you mean the one commented out and not the one I just added?

I suspect this line was commented out when the `WriteCaretLine` was added, but all of this was already there in the first commit to the repo 15 years ago. I agree it should be removed.

src/NUnitFramework/tests/Constraints/MsgUtilTests.cs
manfred-brands: No, it isn't. The original code had one places determining if the actual visible string should be clipped and in the called method it actually reduced the `maxDisplayLength`. I moved all determination code into the first method `ClipWhenNeeded` and reduced the `ClipString` method to do only the clipping to the length passed in.
This means that the length passed into the `ClipString` method is 3 or 6 less than previously and hence the updated numbers in the test.

src/NUnitFramework/framework/Constraints/MsgUtils.cs
manfred-brands: Purely to prevent an re-allocation/copy in most cases when there are a few characters to escape.
It could be any number. I can make it the magic '42' (_“The Hitchhiker's Guide to the Galaxy”_) or 10.

src/NUnitFramework/framework/Constraints/MsgUtils.cs
manfred-brands: In case anyone ever changes the value of `ELLIPSIS`, I changed it to `s.Length + 2 * ELLIPSIS.Length`

src/NUnitFramework/framework/Constraints/MsgUtils.cs
OsirisTerje, line 416:
```suggestion
            const int Headroom = 42;
            StringBuilder sb = new(s.Length + Headroom);
```

manfred-brands: Done it both places.

manfred-brands:
@OsirisTerje I think so, but as I made changes after the review from @mikkelbu, I didn't know if he wants to look at it again.

OsirisTerje:
I'll merge this now.  @mikkelbu did approve earlier.  If he finds more to change, we can add a new PR to address that.  I can't see anything really.   
With this merged, we can then continue with #3918

PR #4665, opened 2024-03-16
Add Prefix in TestContext.AddTestAttachment for long file paths

### LINKED ISSUE #4353: TestContext.AddTestAttachment with long file paths
Hello! 

I have some tests that create different artefacts (log files, screenshots, etc.), and sometimes the full paths of the artefacts can be longer, than 255 symbols. The artefacts are still created and published in TeamCity, but I'm facing the issue, that NUnit can't handle such file paths when I'm using `TestContext.AddTestAttachment(filePath)`.
The problem is hidden inside the `AddTestAttachment` method:

```csharp
            if (!File.Exists(filePath))
                throw new FileNotFoundException("Test attachment file path could not be found.", filePath);

            var result = TestExecutionContext.CurrentContext.CurrentResult;
            result.AddTestAttachment(new TestAttachment(filePath, description));
```
When `filePath` is too long, `File.Exists(filePath)` returns `false` and I'm getting the `FileNotFoundException`. But in reality, the file exists, and I'm able to access it. 

Is there a way to bypass the `File.Exists` validation?

I tried to use the `"\\?\"` prefix as [suggested](https://learn.microsoft.com/en-gb/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#win32-file-namespaces) by Microsoft, but it didn't help.

Meet2rohit99:
Fixes #4353.

For .NetFramework, with file path greater than 260, TestContext.AddTestAttachment was throwing attachment not found exception. As [suggested](https://learn.microsoft.com/en-gb/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#win32-file-namespaces) prefixing "\\\\?\\" with absolute path works fine. The issue is with .NET Framework variants but not on .NET Core3.1, .NET6, or .NET7. Therefore adding prefix only in case of .NetFramework 

                  string prefix = string.Empty;
      #if NETFRAMEWORK
                  if (filePath.Length > 260)
                      prefix = @"\\?\";
      #endif
                  if (!File.Exists($"{prefix}{filePath}"))
                      throw new FileNotFoundException("Test attachment file path could not be found.", filePath);

### DIFF
--- a/src/NUnitFramework/framework/TestContext.cs
+++ b/src/NUnitFramework/framework/TestContext.cs
@@ -371,7 +371,7 @@ namespace NUnit.Framework
 
             string prefix = string.Empty;
 #if NETFRAMEWORK
-            if (filePath.Length > 260)
+            if (Environment.OSVersion.Platform == PlatformID.Win32NT & filePath.Length > 260)
                 prefix = @"\\?\";
 #endif
 
--- a/src/NUnitFramework/framework/TestContext.cs
+++ b/src/NUnitFramework/framework/TestContext.cs
@@ -369,7 +369,13 @@ namespace NUnit.Framework
             if (!Path.IsPathRooted(filePath))
                 filePath = Path.Combine(TestContext.CurrentContext.WorkDirectory, filePath);
 
-            if (!File.Exists(filePath))
+            string prefix = string.Empty;
+#if NETFRAMEWORK
+            if (filePath.Length > 260)
+                prefix = @"\\?\";
+#endif
+
+            if (!File.Exists($"{prefix}{filePath}"))
                 throw new FileNotFoundException("Test attachment file path could not be found.", filePath);
 
             var result = TestExecutionContext.CurrentContext.CurrentResult;
--- a/src/NUnitFramework/tests/TestContextTests.cs
+++ b/src/NUnitFramework/tests/TestContextTests.cs
@@ -25,13 +25,17 @@ namespace NUnit.Framework.Tests
         private string _tempFilePath;
         private string _tempLongFilePath;
 
-        private readonly string _longPathPrefix = @"\\?\";
+        private string _longPathPrefix;
 
         private const string TempFileName = "TestContextTests.tmp";
 
         [OneTimeSetUp]
         public void CreateTempFiles()
         {
+            _longPathPrefix = string.Empty;
+#if NETFRAMEWORK
+            _longPathPrefix = $@"\\?\";
+#endif
             _tempFilePath = Path.Combine(_workDirectory, TempFileName);
             File.Create(_tempFilePath).Dispose();
 
--- a/src/NUnitFramework/tests/TestContextTests.cs
+++ b/src/NUnitFramework/tests/TestContextTests.cs
@@ -23,33 +23,20 @@ namespace NUnit.Framework.Tests
         private readonly string _workDirectory = TestContext.CurrentContext.WorkDirectory;
 
         private string _tempFilePath;
-        private string _tempLongFilePath;
-
-        private string _longPathPrefix;
 
         private const string TempFileName = "TestContextTests.tmp";
 
         [OneTimeSetUp]
-        public void CreateTempFiles()
+        public void CreateTempFile()
         {
-            _longPathPrefix = string.Empty;
-#if NETFRAMEWORK
-            _longPathPrefix = $@"\\?\";
-#endif
             _tempFilePath = Path.Combine(_workDirectory, TempFileName);
             File.Create(_tempFilePath).Dispose();
-
-            /* Absolute file path length greater than 260 are considered long paths for .NET Framework
-             * Creating 261 length file path */
-            _tempLongFilePath = Path.Combine(_workDirectory, new string('A', (260 - _workDirectory.Length) - 4) + ".tmp");
-            File.Create($"{_longPathPrefix}{_tempLongFilePath}").Dispose();
         }
 
         [OneTimeTearDown]
-        public void RemoveTempFiles()
+        public void RemoveTempFile()
         {
             File.Delete(_tempFilePath);
-            File.Delete($"{_longPathPrefix}{_tempLongFilePath}");
         }
 
         [SetUp]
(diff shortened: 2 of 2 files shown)


Meet2rohit99:
> @dotnet-policy-service agree

@dotnet-policy-service agree

stevenaw:
Thanks for your contribution @Meet2rohit99 
The changes themselves look great! Would you be able to add a test for your change too?

Meet2rohit99:
> Thanks for your contribution @Meet2rohit99 The changes themselves look great! Would you be able to add a test for your change too?

Thanks @stevenaw! Added test.

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -369,7 +369,13 @@ public static void AddTestAttachment(string filePath, string? description = null
             if (!Path.IsPathRooted(filePath))
                 filePath = Path.Combine(TestContext.CurrentContext.WorkDirectory, filePath);
 
-            if (!File.Exists(filePath))
+            string prefix = string.Empty;
+#if NETFRAMEWORK
+            if (filePath.Length > 260)
+                prefix = @"\\?\";
```
manfred-brands, line 375: Whether this works or not depends on a registry setting which people (and the github runners) might not have set.
It also might need to be set in the manifest of the .exe running the tests.
See [Maximum File Path Limitation](https://learn.microsoft.com/en-au/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later) for details.

src/NUnitFramework/tests/TestContextTests.cs
```diff
@@ -386,6 +399,12 @@ public void NoneExistentFileThrowsFileNotFoundException()
             Assert.That(() => TestContext.AddTestAttachment("NotAFile.txt"), Throws.InstanceOf<FileNotFoundException>());
         }
 
+        [TestCase(IncludePlatform = "Net")]
+        public void LogFilePathDoesNotThrow()
+        {
+            Assert.That(() => TestContext.AddTestAttachment(_tempLongFilePath), Throws.Nothing);
```
manfred-brands, line 405: This should probably include an `Assume.That(RtlAreLongPathEnabled())`

```
[DllImport("ntdll", ExactSpelling = true)]
private static extern bool RtlAreLongPathsEnabled();
```

manfred-brands (changes requested):
See comments in PR.

src/NUnitFramework/framework/TestContext.cs
manfred-brands: It is also Windows specific, which requires a runtime test, e.g.: `OSPlatform.IsWindows10` or `OSPlatform.Platform == PlatformID.Win32NT`
For other OS there is no path length limit.

stevenaw: Thanks for catching this @manfred-brands . I'd forgotten about mono runtime on Linux/Mac

Meet2rohit99: Got it. As adding prefix for long filepaths is workaround. From Windows 10 version 1607, onwards long file path can be enabled through registry and manifest file. So do we really need to add this workaround in AddTestattachment ? Though we can add this along with runtime test for platform like below-
```
            string prefix = string.Empty;
#if NETFRAMEWORK
            if (Environment.OSVersion.Platform == PlatformID.Win32NT & filePath.Length > 260)
                prefix = @"\\?\";
#endif
```

src/NUnitFramework/tests/TestContextTests.cs
```diff
@@ -23,20 +23,33 @@ public class TestContextTests
...
+#if NETFRAMEWORK
+            _longPathPrefix = $@"\\?\";
+#endif
+            _tempFilePath = Path.Combine(_workDirectory, TempFileName);
             File.Create(_tempFilePath).Dispose();
+
+            /* Absolute file path length greater than 260 are considered long paths for .NET Framework
+             * Creating 261 length file path */
+            _tempLongFilePath = Path.Combine(_workDirectory, new string('A', (260 - _workDirectory.Length) - 4) + ".tmp");
+            File.Create($"{_longPathPrefix}{_tempLongFilePath}").Dispose();
```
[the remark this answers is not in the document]
Meet2rohit99: Thanks @manfred-brands! Definitely Line 42 to 45 can be moved inside test but as we need File.Delete($"{_longPathPrefix}{_tempLongFilePath}") in Teardown so need to declare _tempLongFilePath at class level.

src/NUnitFramework/tests/TestContextTests.cs
Meet2rohit99: As this test is targeted to test the workaround implemented via prefix handling in AddTestAttachment method for LongFilePath inputs that bypasses the role of flag `RtlAreLongPathEnabled`. so do we really need it. Please correct if i am misunderstanding anything.

manfred-brands: I'm not sure on long file names, except that there are many tools that cannot deal with them. So we try to avoid them.

As the test seems to pass on Windows, we might not need the `RtlAreLongPathEnabled`.

src/NUnitFramework/framework/TestContext.cs
manfred-brands: Currently the Framework test on Linux fails because of this `prefix`:
```
System.IO.DirectoryNotFoundException : Could not find a part of the path "/home/runner/work/nunit/nunit/src/NUnitFramework/tests/bin/Release/net462/\\?\/home/runner/work/nunit/nunit/src/NUnitFramework/tests/bin/Release/net462/AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA.tmp".
```

So yes, we do need the test on OS.

If you explanation is correct that long file names only work without the prefix if explicitly enabled, I would still add this workaround if that works for everybody.

src/NUnitFramework/tests/TestContextTests.cs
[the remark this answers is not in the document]
manfred-brands: You can use
```
try
{
   // Create & Test
}
finally
{
    File.Delete(...)
}

src/NUnitFramework/framework/TestContext.cs
Meet2rohit99: Incorporated Comments-Added OS test to add prefix in case of win32NT.

src/NUnitFramework/tests/TestContextTests.cs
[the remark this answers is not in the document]
Meet2rohit99: Thanks @manfred-brands . Incorporated review comments.

manfred-brands (approved):
Thanks. Looking good now. Builds are all green.

stevenaw (approved):
Thanks for your contribution @Meet2rohit99 ! This looks good to me too

PR #4668, opened 2024-03-20
Use MemoryExtensions in StreamsComparer for performance

### LINKED ISSUE #3829: Consider optimizing `StreamsComparer` for happy path
StreamsComparer currently compares the bytes of two streams and reports on the index of the first difference. It does this by reading the streams 4096 bytes at a time into buffers and comparing the arrays byte-by-byte.

Some of the new .NET APIs allow for vectorized equality checks on `Span<byte>`. It may be worth investigating if there's a performance benefit to check for buffer equality via `MemoryExtenions.SequenceEquals()` and falling back to a byte-by-byte check to determine the failure point if the two buffers are known to be unequal.

https://github.com/nunit/nunit/blob/c9fe0e704ce67ad70bf72ccdeae05d4105a2428c/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs#L57-L64

mithileshz:
Use MemoryExtensions to check that the expected and actual are equal.

Fixes #3829

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
@@ -62,6 +62,12 @@ namespace NUnit.Framework.Constraints.Comparers
                     readExpected = binaryReaderExpected.Read(bufferExpected, 0, BUFFER_SIZE);
                     readActual = binaryReaderActual.Read(bufferActual, 0, BUFFER_SIZE);
 
+                    if (MemoryExtensions.SequenceEqual<byte>(bufferExpected, bufferActual))
+                    {
+                        readByte += BUFFER_SIZE;
+                        continue;
+                    }
+
                     for (int count = 0; count < BUFFER_SIZE; ++count)
                     {
                         if (bufferExpected[count] != bufferActual[count])
--- a/src/NUnitFramework/Directory.Packages.props
+++ b/src/NUnitFramework/Directory.Packages.props
@@ -6,6 +6,7 @@
   <!-- Packages for used features -->
   <ItemGroup>
     <PackageVersion Include="System.Collections.Immutable" Version="6.0.0" />
+    <PackageVersion Include="System.Memory" Version="4.5.4" />
     <PackageVersion Include="System.ValueTuple" version="4.5.0" />
   </ItemGroup>
   <!-- General Packages -->
--- a/src/NUnitFramework/benchmarks/nunit.framework.benchmarks/StreamsComparerBenchmark.cs
+++ b/src/NUnitFramework/benchmarks/nunit.framework.benchmarks/StreamsComparerBenchmark.cs
@@ -1,0 +1,217 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using BenchmarkDotNet.Attributes;
+
+namespace NUnit.Framework
+{
+    public class StreamsComparerBenchmark
+    {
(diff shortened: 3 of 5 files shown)


OsirisTerje:
@stevenaw Could you review this please?

stevenaw (dismissed):
Looks great to me! Thanks for your contribution @mithileshz

src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
```diff
@@ -62,6 +62,12 @@ public static EqualMethodResult Equal(object x, object y, ref Tolerance toleranc
                     readExpected = binaryReaderExpected.Read(bufferExpected, 0, BUFFER_SIZE);
                     readActual = binaryReaderActual.Read(bufferActual, 0, BUFFER_SIZE);
 
+                    if (MemoryExtensions.SequenceEqual<byte>(bufferExpected, bufferActual))
+                    {
+                        readByte += BUFFER_SIZE;
```
stevenaw, line 67:
```suggestion
                        readByte += readActual;
```

Sorry, I had missed this in my first pass. It's a bit of a nitpick, but I think it may be safer here to increment by `readActual` here. It's possible for fewer than `BUFFER_SIZE` bytes to be read in and 99% of the time it likely won't matter as it's simply because it's near the end of the stream. For some streams like `NetworkStream` however, it may matter if the reason is that there simply aren't more bytes to read from the network that moment.

stevenaw (commented):
Thanks @mithileshz ! Sorry for the false approval earlier.
This looks really great, but I left one minor suggestion here out of abundance of caution.

src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
mithileshz: That makes sense to me. Should we apply the same change to line 85 below? https://github.com/nunit/nunit/pull/4668/files/8a0a34f1cc5d469714f099c314c56a0bae3fc44d#diff-57f5421e71da296e02ec0d48cfee1a9e89e53de267929ee779797c9d24448fbcR85

stevenaw: Thanks for pointing that out @mithileshz ! Good point, I can see why you approached your PR this way then. Given the current code also advances it by `BUFFER_SIZE` and has been fine for 15 years then I think the PR is also great in it's current state too.

Happy to merge it as-is or to wait if you'd prefer to push another changeset. How would you like to proceed?

stevenaw (approved):

src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
mithileshz: I'll leave the decision up to you as the maintainer of the repository :) 

If we go with the approach for changing it, I would think we need to add a test for it with the newer changes and verify it works for existing scenarios. I don't mind doing it but I would prefer to do it in a separate changeset in case the change causes a regression and would make it easy to roll back the change.

Let me know what you think

stevenaw: @mithileshz I'd like to merge this as-is, it's always nice when a PR lands :)
I really like your proposed approach though, and we'd definitely appreciate a follow-up PR if you wanted to do it. 🙂

stevenaw:
Thanks for your contribution in this PR @mithileshz !

PR #4672, opened 2024-03-25
Support And/Or linking with DictionaryContainsKeyValuePairConstraint

### LINKED ISSUE #4670: `.ContainKey().WithValue()` and `.Or`/`.And` interact incorrectly
Combining `Or`/`And` with the `DictionaryContainsKeyValuePairConstraint` produces incorrect results, see examples below.

The error messages in the failure cases imply that the expression is parsed in a weird/unintended way.

```csharp
[Test]
public void BugReport()
{
    var dictionary = new Dictionary<string, int>
    {
        { "a", 123 },
        { "b", 456 }
    };

    Assert.Multiple(() =>
    {
        Assert.That(dictionary, Does.ContainKey("a").WithValue(456).Or.ContainKey("b").WithValue(123), "Passes but should fail");
        Assert.That(dictionary, Does.ContainKey("a").WithValue(456).And.ContainKey("b").WithValue(456), "Passes but should fail");

        // Expected: dictionary containing key "c" or dictionary containing entry ["c", 456]
        // But was:  < ["a", 123], ["b", 456] >
        Assert.That(dictionary, Does.ContainKey("a").WithValue(123).Or.ContainKey("c").WithValue(456), "Fails but should pass");

        // Expected: dictionary containing key "c" and dictionary containing entry ["c", 456]
        // But was:  < ["a", 123], ["b", 456] >
        Assert.That(dictionary, Does.ContainKey("a").WithValue(123).And.ContainKey("c").WithValue(456), "Fails but for the wrong reason");
    });
}
```

Falco20019:
Fixes #4670

The implementation of #3486 did not combine the the key and value checks, which lead to issues when combining with `And` or `Or`. This PR adds the `InsteadOperator` with high precedence which allows to replace the `DictionaryContainsKeyConstraint` with the `DictionaryContainsKeyValuePairConstraint` since there is no other nice way to replace an entry on the `ConstraintStack` that would also work with correct precedence.

I initially tried to use the `AndOperator`, but this would fail in `Does.Not` cases as the `NotOperator` would apply to the key-check before the AND would be applied.

I also tried using `Contraint.With` (which is currently an alias for `And` and NOT using `WithOperator`). But that would also only bind the key on the right side and adds the value afterwards.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Constraint.cs
+++ b/src/NUnitFramework/framework/Constraints/Constraint.cs
@@ -262,6 +262,27 @@ namespace NUnit.Framework.Constraints
             }
         }
 
+        /// <summary>
+        /// Returns a ConstraintExpression by appending Instead
+        /// to the current constraint.
+        /// </summary>
+        internal ConstraintExpression Instead
+        {
+            get
+            {
+                ConstraintBuilder? builder = Builder;
+                if (builder is null)
+                {
+                    builder = new ConstraintBuilder();
+                    builder.Append(this);
+                }
+
+                builder.Append(new InsteadOperator());
+
+                return new ConstraintExpression(builder);
+            }
+        }
+
         #endregion
 
         #region After Modifier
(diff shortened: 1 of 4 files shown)


OsirisTerje (approved):
LGTM

OsirisTerje:
@Falco20019 Really appreciate the fast response !   Awesome work!

PR #4674, opened 2024-03-25
Add early return for Streams Equal comparison when they are empty

mithileshz:
When the streams are seekable and are empty, we can return they are equal and save the allocation of the byte arrays and the comparisons around them.


OsirisTerje (approved):
LGTM

@mithileshz Tip:  Note that we require issues for release notes.  This is small, but - just for the future :-)

mithileshz:
@OsirisTerje ah sorry! I'll keep that in mind for next time. Thank you!

OsirisTerje:
@mithileshz No problem :-)  We do accept PRs without too. It is just that we like to give credit to those who contribute PRs in the [release notes](https://docs.nunit.org/articles/nunit/release-notes/framework.html#enhancements-1) , but we do that through the issues and checks who have done the corresponding PR, using an automated process. So without issues you don't get that honorable mention you do deserve :-)

PR #4689, opened 2024-04-12
Permit DictionaryContainsKeyValuePairConstraint to take in a null value (fix CS8604)

andrewimcclement:
This matches the nullability of DictionaryEntry.

Context of the issue I hit:
```csharp
internal class Foo
{
    [Test]
    public void Bar()
    {
        var dictionary = new Dictionary<int, string?> { {0, null} };
        Assert.That(dictionary, Contains.Key(0).WithValue(null)); // This line raises CS8604.
    }
}
```

### DIFF
--- a/src/NUnitFramework/tests/Constraints/DictionaryContainsKeyValueConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/DictionaryContainsKeyValueConstraintTests.cs
@@ -10,10 +10,10 @@ namespace NUnit.Framework.Tests.Constraints
     [TestFixture]
     public class DictionaryContainsKeyValuePairConstraintTests
     {
-        [Theory]
-        public void SucceedsWhenKeyValuePairIsPresent(bool valueIsNull)
+        [TestCase("Universe")]
+        [TestCase(null)]
+        public void SucceedsWhenKeyValuePairIsPresent(string? expectedValue)
         {
-            var expectedValue = valueIsNull ? null : "Universe";
             var dictionary = new Dictionary<string, string?> { { "Hello", "World" }, { "Hi", expectedValue }, { "Hola", "Mundo" } };
 
             Assert.That(dictionary, new DictionaryContainsKeyValuePairConstraint("Hi", expectedValue));
@@ -39,10 +39,10 @@ namespace NUnit.Framework.Tests.Constraints
             Assert.That(act, Throws.Exception.TypeOf<AssertionException>());
         }
 
-        [Theory]
-        public void SucceedsWhenPairIsPresentUsingContainKeyWithValue(bool valueIsNull)
+        [TestCase("Mundo")]
+        [TestCase(null)]
+        public void SucceedsWhenPairIsPresentUsingContainKeyWithValue(string? expectedValue)
         {
-            var expectedValue = valueIsNull ? null : "Mundo";
             var dictionary = new Dictionary<string, string?> { { "Hello", "World" }, { "Hola", expectedValue } };
             Assert.That(dictionary, Does.ContainKey("Hola").WithValue(expectedValue));
         }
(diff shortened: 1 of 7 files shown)


manfred-brands (changes requested):
@andrewimcclement 
There is also a DictionaryContainsValueConstraint that needs updating.

Can you also update `Does.ContainValue`, `Contains. Value` and `ConstraintExpression.ContainValue` to allow `null`?

Could you also please add some unit test to verify that not only we now allow specifying `null` but that we also can find it.

andrewimcclement:
Thanks for the spots, I'll update the places you suggest. I wasn't really sure what to add for a unit test - it just seemed like any other value, i.e. covered by the existing tests. I'm happy to parameterise some of the existing tests to use either a null value or a non-null value maybe? I'll see how I go.

mikkelbu:
> ... I'm happy to parameterise some of the existing tests to use either a null value or a non-null value maybe? I'll see how I go.

That sounds good 👍

andrewimcclement:
Added a couple of tests. Shout if there is anything you want changed.

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyValueConstraintTests.cs
```diff
@@ -10,12 +10,13 @@ namespace NUnit.Framework.Tests.Constraints
     [TestFixture]
     public class DictionaryContainsKeyValuePairConstraintTests
     {
-        [Test]
-        public void SucceedsWhenKeyValuePairIsPresent()
+        [Theory]
+        public void SucceedsWhenKeyValuePairIsPresent(bool valueIsNull)
```
manfred-brands, line 14: Technically correct, but `[Theory]` isn't used a lot inside NUnit. It also cause the builder to look for members with a `[DataPoint]` or `[DataPointSource]` attribute in the class before dropping to the fallback for `bool`.
I wonder if the below is more in line with the rest and clearer as it doesn't require neither the developer not the engine looking for possible values in the class.
```csharp
[Test]
public void SucceedsWhenKeyValuePairIsPresent([Values] bool valueIsNull)
```

Or it might be even nicer to go for:
```csharp
[TestCase("Universe")]
[TestCase(default(string)]
public void SucceedsWhenKeyValuePairIsPresent(string? expectedValue)
```

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyValueConstraintTests.cs
```diff
@@ -38,11 +39,12 @@ public void FailsWhenValueIsMissing()
...
         }
 
-        [Test]
-        public void SucceedsWhenPairIsPresentUsingContainKeyWithValue()
+        [Theory]
+        public void SucceedsWhenPairIsPresentUsingContainKeyWithValue(bool valueIsNull)
         {
-            var dictionary = new Dictionary<string, string> { { "Hello", "World" }, { "Hola", "Mundo" } };
-            Assert.That(dictionary, Does.ContainKey("Hola").WithValue("Mundo"));
+            var expectedValue = valueIsNull ? null : "Mundo";
```
manfred-brands, line 45: Suggest:
```csharp
[TestCase("Mundo")]
[TestCase(default(string)]
public void SucceedsWhenPairIsPresentUsingContainKeyWithValue(string? expectedValue)
{
```

manfred-brands (commented):
Thanks @andrewimcclement.
The use of `Theory` threw me a little as we don't use that a lot.
The code is correct, but I think my suggestion will make it clearer and understood by more developers.

andrewimcclement:
Changed to using `[TestCase]`.

I will argue for [Theory] being a very useful attribute, but it matters more when there are 2+ parameters, when it saves you writing out [Values] or [ValueSource] for each parameter. When there is just one, it is purely a matter of style :) I agree using [TestCase] is more in keeping with the rest of the code - I'm simply in the habit of using [Theory] a lot.

andrewimcclement:
Thanks for the code reviews, it is much appreciated.

PR #4673, opened 2024-03-25
Issue4658 timeout

manfred-brands:
This PR adds support for proper Timeout to .NET 8.0

Issue #2397, opened 2017-08-29
Test timeout with teardown executed?


kdubau:
Hi folks, I've done a bit of searching and can't find a solution. Searching `timeout` in the issues here has some relevant results but issues are either closed without resolution (from years back) or still open. 

I have a simple question: for those of us that require a way to have a test timeout but _still execute the teardown_ what are our options? If a test case does some setup, and times out during the test, in many cases the teardown still needs to execute. That is, there is still some clean up that needs to happen so things are healthy for the next test. 

I've searched and pondered but haven't found a solution. Can you recommend something? Are there any extensions points maybe? It seems timeout functionality is of limited use if we cannot clean up after the timed out test. Arguably, cleaning up after a timed out test can be more important than even a successful test.

kdubau:
In addition to cleanup, the user might want to do some analysis after the time out. ie. What did it timeout? The user should be able to capture logs, machine state, etc.

ChrisMaddock:
I hit exactly the same problem a few months ago - @kdubau, did you come across this thread? https://groups.google.com/forum/#!topic/nunit-discuss/3-v-6bosbyM

I never found the time to address it - I agree it would be a good feature.

kdubau:
I actually didn't find that particular thread, but that summarizes my situation quite perfectly.

>My gut feeling is that Timeout should only apply to the test method, and neither setup nor teardown.

Same here!

>allow you to create a method that runs after a timeout, marking it with some special attribute.

Would work for me!

I would love to see this become a feature, but I need something ~now - any ideas for something temporary to hold me over? I can't use OneTimeTearDown, mostly for the same reasons you posted in that discussion thread.

CharliePoole:
I replied to @ChrisMaddock 's earlier discussion thread with two alternative ways it could be done. I had hoped somebody would pick one and create an issue for it. I sincerely hope this issue will not become another long discussion as to why it's needed, why the current design is bad,etc. 

If this is a request for a workaround (i.e. question) the answer is that there really is not one. We should close it. If we want to make this the issue that implements an enhancement, the person wanting to carry it out should pick one of the two approaches I suggested and propose doing it. Neither is something I would support as suitable for a new contributor because they both apply to the innermost workings of our test execution. I'm stepping back from the project myself but I'll be glad to consult with a team member who steps up to do this.

[...]

PR #4692, opened 2024-04-15
Apply test output from timedout tests on non-threadabort platforms

stevenaw:
Fixes https://github.com/nunit/nunit/issues/4598

@manfred-brands You mentioned [in the source issue](https://github.com/nunit/nunit/issues/4598#issuecomment-2026408377) that you may also be looking into this on your own branch.

I've put up this PR as I was able to get this working in a relatively self-contained way for the existing code which may be able to merge either independent of your larger rework or could perhaps be PR'd into it as well. This PR could also be discarded if you've already gotten it working on your end too

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
@@ -101,7 +101,7 @@ namespace NUnit.Framework.Internal.Commands
                 var testExecution = Task.Run(() => innerCommand.Execute(separateContext));
                 var timedOut = Task.WaitAny(new Task[] { testExecution }, _timeout) == -1;
 
-                separateContext.CurrentResult.ApplyOutput(context.CurrentResult);
+                context.CurrentResult.CopyOutputTo(separateContext.CurrentResult);
 
                 if (timedOut && !_debugger.IsAttached)
                 {
--- a/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
@@ -119,9 +119,12 @@ namespace NUnit.Framework.Internal.Commands
 
         private Task<TestResult> RunTestOnSeparateThread(TestExecutionContext context)
         {
-            var separateContext = new TestExecutionContext(context);
+            var separateContext = new TestExecutionContext(context)
+            {
+                CurrentResult = context.CurrentTest.MakeTestResult()
+            };
 
-            return Task.Run(() => innerCommand.Execute(context));
+            return Task.Run(() => innerCommand.Execute(separateContext));
         }
 #endif
     }
--- a/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
@@ -102,11 +102,10 @@ namespace NUnit.Framework.Internal.Commands
                 }
                 else
                 {
-                    string message = $"Test exceeded Timeout value of {_timeout}ms";
-
+                    context.CurrentResult.ApplyTimeoutResult(testExecution.Result);
                     context.CurrentResult.SetResult(
                         ResultState.Failure,
-                        message);
+                        $"Test exceeded Timeout value of {_timeout}ms");
                 }
             }
             catch (Exception exception)
--- a/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
@@ -94,18 +94,24 @@ namespace NUnit.Framework.Internal.Commands
         {
             try
             {
-                var testExecution = RunTestOnSeparateThread(context);
-                if (Task.WaitAny(new Task[] { testExecution }, _timeout) != -1
-                    || _debugger.IsAttached)
+                var separateContext = new TestExecutionContext(context)
                 {
-                    context.CurrentResult = testExecution.GetAwaiter().GetResult();
-                }
-                else
+                    CurrentResult = context.CurrentTest.MakeTestResult()
+                };
+                var testExecution = Task.Run(() => innerCommand.Execute(separateContext));
+                var timedOut = Task.WaitAny(new Task[] { testExecution }, _timeout) == -1;
+
+                separateContext.CurrentResult.ApplyOutput(context.CurrentResult);
+
+                if (timedOut && !_debugger.IsAttached)
                 {
-                    context.CurrentResult.ApplyTimeoutResult(testExecution.Result);
                     context.CurrentResult.SetResult(
                         ResultState.Failure,
                         $"Test exceeded Timeout value of {_timeout}ms");
+                }
+                else
+                {
+                    context.CurrentResult = testExecution.GetAwaiter().GetResult();
                 }
             }
             catch (Exception exception)
@@ -114,16 +120,6 @@ namespace NUnit.Framework.Internal.Commands
             }
 
             return context.CurrentResult;
-        }
-
-        private Task<TestResult> RunTestOnSeparateThread(TestExecutionContext context)
-        {
-            var separateContext = new TestExecutionContext(context)
-            {
-                CurrentResult = context.CurrentTest.MakeTestResult()
-            };
-
-            return Task.Run(() => innerCommand.Execute(separateContext));
         }
 #endif
     }
--- a/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
@@ -101,8 +101,6 @@ namespace NUnit.Framework.Internal.Commands
                 var testExecution = Task.Run(() => innerCommand.Execute(separateContext));
                 var timedOut = Task.WaitAny(new Task[] { testExecution }, _timeout) == -1;
 
-                context.CurrentResult.CopyOutputTo(separateContext.CurrentResult);
-
                 if (timedOut && !_debugger.IsAttached)
                 {
                     context.CurrentResult.SetResult(
@@ -111,6 +109,7 @@ namespace NUnit.Framework.Internal.Commands
                 }
                 else
                 {
+                    context.CurrentResult.CopyOutputTo(separateContext.CurrentResult);
                     context.CurrentResult = testExecution.GetAwaiter().GetResult();
                 }
             }
--- a/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
@@ -119,11 +119,9 @@ namespace NUnit.Framework.Internal.Commands
 
         private Task<TestResult> RunTestOnSeparateThread(TestExecutionContext context)
         {
-            var separateContext = new TestExecutionContext(context)
-            {
-                CurrentResult = context.CurrentTest.MakeTestResult()
-            };
-            return Task.Run(() => innerCommand.Execute(separateContext));
+            var separateContext = new TestExecutionContext(context);
+
+            return Task.Run(() => innerCommand.Execute(context));
         }
 #endif
     }
--- a/src/NUnitFramework/testdata/TimeoutFixture.cs
+++ b/src/NUnitFramework/testdata/TimeoutFixture.cs
@@ -154,4 +154,27 @@ namespace NUnit.TestData
         {
         }
     }
+
+    [TestFixture]
+    public class TimeoutWithSetupTestAndTeardownOutputFixture
+    {
+        [SetUp]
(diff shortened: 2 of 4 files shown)


src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -94,19 +94,24 @@ public override TestResult Execute(TestExecutionContext context)
...
-                {
-                    string message = $"Test exceeded Timeout value of {_timeout}ms";
+                    CurrentResult = context.CurrentTest.MakeTestResult()
+                };
+                var testExecution = Task.Run(() => innerCommand.Execute(separateContext));
+                var timedOut = Task.WaitAny(new Task[] { testExecution }, _timeout) == -1;
 
+                separateContext.CurrentResult.ApplyOutput(context.CurrentResult);
+
+                if (timedOut && !_debugger.IsAttached)
```
stevenaw, line 106: Note to reviewer: I had to move the `Task.WaitAny` to it's own line in order to apply the test output afterwards. I ended up inverting this check here as part of that as I found it read clearer this way but I'm happy to instead change it back to something like:
```
if (!timedOut || _debugger.IsAttached)
```

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -94,19 +94,24 @@ public override TestResult Execute(TestExecutionContext context)
...
-                }
-                else
-                {
-                    string message = $"Test exceeded Timeout value of {_timeout}ms";
+                    CurrentResult = context.CurrentTest.MakeTestResult()
+                };
+                var testExecution = Task.Run(() => innerCommand.Execute(separateContext));
+                var timedOut = Task.WaitAny(new Task[] { testExecution }, _timeout) == -1;
 
+                separateContext.CurrentResult.ApplyOutput(context.CurrentResult);
```
manfred-brands, line 104: Should this not be the other way around?
You want the _separateContext_ to go to the _context_ not the other way around?

src/NUnitFramework/tests/Attributes/TimeoutTests.cs
```diff
@@ -174,6 +175,27 @@ public void TestTimesOutAndTearDownIsRun()
...
+
+        [Test]
+        public void OutputIsCapturedOnTimedoutTestAfterTimeout()
+        {
+            var suiteResult = TestBuilder.RunTestFixture(typeof(TimeoutWithSetupAndOutputAfterTimeoutFixture));
+            var testMethod = suiteResult.Children.First();
+
+            Assert.That(testMethod.Output, Does.Contain("setup"));
+            Assert.That(testMethod.Output, Does.Contain("method output before pause"));
+            Assert.That(testMethod.Output, Does.Not.Contain("method output after pause"));
```
manfred-brands, line 196: Under .Net core that data will be written after the test has timed out and is likely not available.

src/NUnitFramework/testdata/TimeoutFixture.cs
```diff
@@ -108,4 +108,50 @@ public void TestTimeOutTestCase(int delay)
...
+        [Test, Timeout(2_000)]
+        public void Test2()
+        {
+            TestContext.WriteLine("method output");
+            Assert.That(1, Is.EqualTo(0));
+        }
+
+        [TearDown]
+        public void TearDown()
+        {
```
manfred-brands, line 154: The #4598 issue talks about `TearDown` output being lost. We need a test with that in.

manfred-brands (changes requested):
Not sure if this will work.
The problem is the test continuing to run and writing output to its context when at the same time the teardown is run.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
stevenaw: From memory (which may be flawed) but I think I saw in the debugger that these were getting _written_ to the original context but that it was getting overridden when we create the testResult from the separateContext (which is what would then get output).

I'll load this up and rerun it soon to confirm

stevenaw: I see the method wording here could be confusing too.
Leaving desired functionally aside for a moment, if the intention of this code were to take the output from `separateContext` and apply it to `context` perhaps the method name would be clearer if it were `ApplyOutputTo`?

manfred-brands: `CopyOutputTo` would be clearer. At the moment unless I got it wrong it does a `CopyOutputFrom`

stevenaw: Of course, thanks! That's also more consistent with some BCL methods like `Stream.CopyTo()`

src/NUnitFramework/testdata/TimeoutFixture.cs
stevenaw: Good point! I had changed the test from having a `TearDown` to a `SetUp` during development and missed re-adding a separate `TearDown` case. I'll add one.

I think the originating issue sought more to keep output from the test method itself in the presence of a `TearDown` rather than output from the `TearDown` itself but that will be a good test for me to add here to. One thing to note here though is that I think this was a NET6-specific bug and that presently `TearDown` is presently only run for timedout test cases on netfx (https://github.com/nunit/nunit/issues/2397). I verified the behaviour described there and it's consistent as far back as 3.12

stevenaw: Wait, sorry. @manfred-brands I think this does cover the example in the issue as it talks about test output from the test rather than the `TearDown`, but I'll still take another look and see if there's any edge cases I could add

src/NUnitFramework/tests/Attributes/TimeoutTests.cs
stevenaw: I think that's the intended behaviour, right? The `Does.Not` here ensures that any output written after the test times out isn't captured

manfred-brands: It is not the _intended_ behaviour, but the current behaviour on net 6.0`

stevenaw: I believe it's also the current behavior on other runtimes like netfx. Changing that would be a bigger body of work than I was looking to do in this PR, and one which we may wish to discuss further with the team in a separate issue. 

If we were to keep this PR focused on just resuming outputting of what is executed, would you prefer we leave this assertion in place knowing it may change if the behavior does, or should we remove the assertion?

stevenaw:
@manfred-brands I've been able to get back to coding this.

I've pushed the naming change to adjust `ApplyContext` -> `CopyOutputTo` to make it a bit clearer what's happening. My memory was indeed correct, the output was getting lost because the `currentContext` was getting written to while the `separateContext` owned the `TestResult` which was getting output.

To verify the 3.x functionality I'm trying to restore here, I've put up a branch with the same tests on the 3.x branch. CI is still verifying but they pass as-is locally for me.

I'm a little unclear on if there's further changes you'd like to see here, can you please take a look and let me know?

stevenaw:
@manfred-brands I know you've been busy between a few different PRs lately. Could I ask you for a bit of clarification on if you wanted further changes here? 

I'd been hoping to keep this PR focused to just restoring 3.x behavior but if it also starts discussion about other changes then I'm all for that discussion too.

src/NUnitFramework/testdata/TimeoutFixture.cs
```diff
@@ -108,4 +108,50 @@ public void TestTimeOutTestCase(int delay)
...
         }
     }
+
+    [TestFixture]
+    public class TimeoutWithSetupAndOutputAfterTimeoutFixture
+    {
+        [SetUp]
+        public void Setup()
+        {
+            TestContext.WriteLine("setup");
```
manfred-brands, line 118: This writes to `TestExecutionContext.CurrentContext.OutWriter`
The `separateContext` in `TimedOutCommand` does not change that.
Unfortunately NUnit uses `static` fields and sometimes parameters.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -94,19 +94,24 @@ public override TestResult Execute(TestExecutionContext context)
...
+                if (timedOut && !_debugger.IsAttached)
+                {
                     context.CurrentResult.SetResult(
                         ResultState.Failure,
-                        message);
+                        $"Test exceeded Timeout value of {_timeout}ms");
+                }
+                else
+                {
+                    context.CurrentResult = testExecution.GetAwaiter().GetResult();
```
manfred-brands, line 113: This is the place where `CurrentResult` gets overwritten and any output created by `SetUp` output would get lost.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -94,19 +94,24 @@ public override TestResult Execute(TestExecutionContext context)
...
-                }
-                else
-                {
-                    string message = $"Test exceeded Timeout value of {_timeout}ms";
+                    CurrentResult = context.CurrentTest.MakeTestResult()
+                };
+                var testExecution = Task.Run(() => innerCommand.Execute(separateContext));
+                var timedOut = Task.WaitAny(new Task[] { testExecution }, _timeout) == -1;
 
+                context.CurrentResult.CopyOutputTo(separateContext.CurrentResult);
```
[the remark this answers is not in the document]
manfred-brands: When putting a break point here, the `separatedContext.CurrentResult.Output` is empty and the text is in the `context.CurrentResult.Output`

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
manfred-brands: I moved the `CopyOutputTo` here to make it clear that this is the relevant location.

src/NUnitFramework/tests/Attributes/TimeoutTests.cs
manfred-brands: Leave it as is.

manfred-brands (commented):
@stevenaw So sorry for the delay, but I didn't get to it to pay it the attention needed.

Creating a _separateContext_ with a new `CurrentResult` does create a new `OutWriter`, but that one will not be used as nothing has access to this from an NUnit tests.
Therefore this `Output` will stay empty.
Then the overall result is set to the new (empty) result. This forces you to copy the `CurrentContext` output to _separateContext_

I think the _separateContext_ was created so any code executing after the test timed out won't affect status, but as it is not exposed it doesn't seem to do much.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
```diff
@@ -94,19 +94,23 @@ public override TestResult Execute(TestExecutionContext context)
         {
             try
             {
-                var testExecution = RunTestOnSeparateThread(context);
-                if (Task.WaitAny(new Task[] { testExecution }, _timeout) != -1
-                    || _debugger.IsAttached)
+                var separateContext = new TestExecutionContext(context)
```
manfred-brands, line 97: I think we need to use `IsolatedContext` which would already to the `CopyOutput` part.
But that relies on setting the `TestExecutionContect.CurrentContext`. 
How does that work with the _context_ parameter.

manfred-brands (commented):
The fact that we have both `CurrentContext` and a `context` parameter is confusing.

src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
manfred-brands: Actually no matter what you do here will help. The test will use whatever the current context is at the time, which might be the context of the next test running.

So we can only solve the case where the test doesn't time out.
So either your fix or not creating a separate context at all.

stevenaw:
Thanks for the prompt and detailed feedback @manfred-brands , appreciate it!
My turnaround time for coding activities is slower at the moment but I will take a look and push follow-ups when I am next able

stevenaw:
@manfred-brands I've read through your feedback. I'm not 100% sure, but were there further changes you wanted to see here?

Reminder that, when combined with your own PR https://github.com/nunit/nunit/pull/4673, this code will primarily only serve .NET6 which will be an increasingly small base over time

manfred-brands:
@stevenaw I think we can have leave it as is. It solves the problem when test with a timeout succeed in time.
When they don't, we know that on .NET Core it is a can of worms.

Regarding #4673 I have to sort out the deadlock in the CI build under linux, but have been to busy with work to dive into this, hopefully this weekend.

stevenaw:
Thanks for the confirmation @manfred-brands ! Appreciate your time and review here.

> have been to busy with work to dive into this, hopefully this weekend.

I completely understand, I'm also pretty busy in this realm at the moment too.

Issue #4653, opened 2024-03-05
Additional Test Events for OneTimeSetUp / OneTimeTearDown


z002Holpp:
Hi all,

For our higher-level System Tests, we need a mechanism that notifies us during the execution of a test run when certain parts of the test suite are executed. Regarding the begin and the end of a test case, NUnit already already provides corresponding test events. So we can implement a custom TestEventListener and attach some handlers to it that react when the corresponding events occure.
However, we need a similar mechanism also for OneTimeSetUp and OneTimeTearDown.

We implemented a first draft of a solution that provides additional test events for OneTimeSetUp and OneTimeTearDown.
Our main uncertainty during the implementation was to find the right place where the events should be raised.

The implementation / PR can be found at
https://github.com/nunit/nunit/pull/4652
with a brief introduction / explanation in
https://github.com/nunit/nunit/blob/c76d7d2200413d51ef307464562004617f19afec/AdditionalTestEvents.md

It would be a pleasure to discuss this approach with you and maybe some other experts in this round.
The mentioned PR should serve more as a base for a technical discussion rather than being "production-ready".

Thanks and best regards,
Stefan

OsirisTerje:
@manfred-brands @stevenaw  Comments?

stevenaw:
This is a neat idea and seems to cover a gap in the current event listeners. I acknowledge it may require breaking changes given the interface we'd have to modify.

@OsirisTerje @manfred-brands Thoughts or concerns?

OsirisTerje:
@stevenaw We don't need to change ITest.  ITest is really about the tests themselves, and I don't think it is a fit for these. Since these are changes to the OneTimeSetup/Teardown we can either use another interface or add a new one (if we are missing one, which I suspect).  That would move this more into an enhancement than a breaking change.

stevenaw:
Oh, great point!

TSAVogt:
Hi,
I am colleague of @z002Holpp and working in the same context.
There are two major challenges:
1. Finding the right place where to raise the event.
1. Designing a new interface for providing the needed information similar to ITest.

We would like to discuss 1. first to have the right location for such notification.

@z002Holpp already provided in [Additional Test Events](https://github.com/nunit/nunit/blob/c76d7d2200413d51ef307464562004617f19afec/AdditionalTestEvents.md) the three options:
 a. OneTimeSetUpCommand.cs
 b. SetUpTearDownItem.cs
 c. CompositeWorkItem.cs

[...]

PR #4698, opened 2024-04-21
Enabling extension of events

OsirisTerje:
If one is to add more events without touching the existing event system - which many may be dependent upon - these can be added to the code by following separate paths.  

In order to avoid code duplication where unnecessary, the changes in this PR make some non-functional changes to the existing code.  The classes are made generic, and some new interfaces are added to bind them together. 

A new event system can then be inserted into the code using these classes, but have their implementations of the critical classes, like Event, ITestListener, and QueuingEventListener.  

See flow of events [here](https://github.com/nunit/AnalysisAndDesign/blob/main/Eventflow.md)

UPDATE:

* Will allow externals to add extra events, like we have in the Issue #4653 .
* Will allow the NUnit Team to extend the events without breaking the "older" event system.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
@@ -1,5 +1,6 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
+using NUnit.Framework.Interfaces;
 using System;
 using System.Threading;
 
@@ -29,11 +30,30 @@ namespace NUnit.Framework.Internal.Execution
 
     /// <summary>
     /// EventPump pulls events out of an EventQueue and sends
-    /// them to a listener. It is used to send events back to
+    /// them to a ITestListener. It is used to send events back to
     /// the client without using the CallContext of the test
     /// runner thread.
     /// </summary>
-    public class EventPump<TEvent, TListener>
+    public class EventPump : EventPumpTemplate<Event, ITestListener>, IDisposable
+    {
+        /// <summary>
+        /// Constructor for standard EventPump
+        /// </summary>
+        /// <param name="eventListener">The EventListener to receive events</param>
+        /// <param name="events">The event queue to pull events from</param>
+        public EventPump(ITestListener eventListener, EventQueueTemplate<Event> events)
+            : base(eventListener, events, "Standard")
+        {
+        }
+    }
+
+    /// <summary>
+    /// EventPump template pulls events of any type out of an EventQueue and sends
+    /// them to any listener. It is used to send events back to
+    /// the client without using the CallContext of the test
+    /// runner thread.
+    /// </summary>
+    public class EventPumpTemplate<TEvent, TListener>
         where TEvent : IEvent<TListener>,
         IDisposable
     {
@@ -49,7 +69,7 @@ namespace NUnit.Framework.Internal.Execution
         /// <summary>
         /// The queue that holds our events
         /// </summary>
-        private readonly EventQueue<TEvent> _events;
+        private readonly EventQueueTemplate<TEvent> _events;
 
         /// <summary>
         /// Thread to do the pumping
@@ -71,7 +91,7 @@ namespace NUnit.Framework.Internal.Execution
         /// <param name="eventListener">The EventListener to receive events</param>
         /// <param name="events">The event queue to pull events from</param>
         /// <param name="name">Name of the thread and pump</param>
-        public EventPump(TListener eventListener, EventQueue<TEvent> events, string name = "Standard")
+        public EventPumpTemplate(TListener eventListener, EventQueueTemplate<TEvent> events, string name = "Standard")
         {
             _eventListener = eventListener;
             _events = events;
--- a/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
@@ -53,9 +53,8 @@ namespace NUnit.Framework.Internal.Execution
     /// the client without using the CallContext of the test
     /// runner thread.
     /// </summary>
-    public class EventPumpTemplate<TEvent, TListener>
-        where TEvent : IEvent<TListener>,
-        IDisposable
+    public class EventPumpTemplate<TEvent, TListener> : IDisposable
+        where TEvent : IEvent<TListener>
     {
         private static readonly Logger Log = InternalTrace.GetLogger("EventPump");
 
--- a/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/EventPump.cs
@@ -29,31 +29,31 @@ namespace NUnit.Framework.Internal.Execution
     }
 
     /// <summary>
-    /// EventPump pulls events out of an EventQueue and sends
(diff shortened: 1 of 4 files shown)


OsirisTerje:
- Naming of the "template" classes?  
- Should possibly be abstract ?  Would prevent from spreading template parameters all over the place.

src/NUnitFramework/framework/Internal/Execution/EventPump.cs
```diff
@@ -1,8 +1,8 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
+using NUnit.Framework.Interfaces;
```
manfred-brands, line 3: Do we use the System namespace first or pure alphabetical convention in NUnit? I thought it was the first.

src/NUnitFramework/framework/Internal/Execution/EventPump.cs
```diff
@@ -142,14 +166,14 @@ public void Stop()
         /// </summary>
         private void PumpThreadProc()
         {
-            Log.Debug("Starting EventPump");
+            Log.Debug($"Starting EventPump {Name}");
```
manfred-brands, line 169: This would duplicate the word `EventPump` as it is part of `Name`

src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
```diff
@@ -1,11 +1,25 @@
...
+    /// Interface for ALL event types that can be queued for processing.
+    /// </summary>
+    /// <typeparam name="TListener"></typeparam>
+    public interface IEvent<in TListener>
+    {
+        /// <summary>
+        /// The Send method is implemented by derived classes to send the event to the specified listener.
+        /// </summary>
+        /// <param name="listener">The listener.</param>
+        void Send(TListener listener);
```
manfred-brands, line 20: Can you explain this? Events sending themselves sounds not logical. What about multiple listeners?

src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
```diff
@@ -15,13 +29,20 @@ namespace NUnit.Framework.Internal.Execution
     /// or to queue them for forwarding on another thread or at
     /// a later time.
     /// </summary>
-    public abstract class Event
+    public abstract class Event : IEvent<ITestListener>, IDisposable
```
manfred-brands, line 32: Should an `Event` not be a pure data class?
Who would `Dispose` an event. Normally code creating would do that, but they don't know when it is received. A receiver can't do it as we don't know how many listeners there are.

src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
```diff
@@ -15,13 +29,20 @@ namespace NUnit.Framework.Internal.Execution
...
     {
         /// <summary>
         /// The Send method is implemented by derived classes to send the event to the specified listener.
         /// </summary>
         /// <param name="listener">The listener.</param>
         public abstract void Send(ITestListener listener);
+
+        /// <summary>
+        /// Not quite sure why this is needed, but the compiler complained.
```
manfred-brands, line 41: It is needed because the class is declared `IDisposable`. Like the `Send` method as a minimum you need an `abstract` one.

src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
```diff
@@ -15,13 +29,20 @@ namespace NUnit.Framework.Internal.Execution
...
+        /// </summary>
+        public void Dispose()
```
manfred-brands, line 43: If this is not `virtual` derived classes cannot implement it if needed.

manfred-brands (commented):
I'm confused about the `Event` class and put some questions in the PR.

src/NUnitFramework/framework/Internal/Execution/EventPump.cs
OsirisTerje: Didnt even notice,  good point. Fixed.

src/NUnitFramework/framework/Internal/Execution/EventPump.cs
OsirisTerje: Thanks!

src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
OsirisTerje: This is the way it has been for years. No change here.  I agree it doesn't look right, but I am not changing the logic here.
The code the Send method in a concrete event like TestStartedEvent looks like
```csharp
public override void Send(ITestListener listener)
{
    listener.TestStarted(_test);
}
```
where _test is an ITest.  So it is actually not sending itself, but the content it holds. An Event is just an envelope.

OsirisTerje: And yes, multiple listeners are not implemented.

src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
OsirisTerje: Yes, that was the intention, but the compiler complained with the declaration for EventQueue, and I just dropped in the IDisposable to fix it.  It should not be there, but I didn't find why it complained.

src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
OsirisTerje: Yes, I that was added AFTER I got a compilation error.  
Without it, I get this:
image

Suggestions?

src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
OsirisTerje: good point! Fixed.

src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
manfred-brands: In the EventPump declaration you put the `IDisposable` on the `where` clause instead of the class itself. You need to specify interface before where clause

src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
manfred-brands: Thanks for the explanation

src/NUnitFramework/framework/Internal/Execution/EventQueue.cs
OsirisTerje: Ahhrrg!!   Thanks @manfred-brands !   That was indeed the case!  I was surely blinded there! :-)

manfred-brands:
> * Naming of the "template" classes?

I don't like the `Template` suffix and suggest to drop it.
C# allows having `EventQueue` and `EventQueue<T>`

> * Should possibly be abstract ?  Would prevent from spreading template parameters all over the place.

If that is the intention yes. Also make then constructors `protected`

> * Should the concrete derived classes be sealed ?

Yes. `EventQueue` as a specific implementation.

manfred-brands (approved):
Code ok with the caveat my answers to your questions regarding naming and abstract/sealed classes

OsirisTerje:
Thanks @manfred-brands !   I'll do the latest changes you suggest.  I do agree with them all!

Issue #3767, opened 2021-02-11
Incorrect number of items listed in failure message


dahvyd:
The output of this code:
```
static void AssertHasHeaderWithValue(HttpResponseMessage response, string headerName, string value)
{
	Assert.That(response.Headers.TryGetValues(headerName, out var values), Is.True, "Header name: " + headerName);
	Assert.That(values, Has.Exactly(1).EqualTo("boop"));
}
```
Is:
```
  Message: 
      Expected: exactly one item equal to "boop"
      But was:  **0 items** < "no-store, no-cache" >
```
Seems like it should say "1 item" instead.

manfred-brands:
The 0 is referring to the number of matches, not the total number of entries in the collection.

dahvyd:
Hmm that isn't overly clear. Perhaps it should just list the items? Or say `0 matching items`?

OsirisTerje:
We can make the message more clear, as suggested.

PR #4702, opened 2024-05-02
Improve Assert.That error text message

iamdmitrij:
Fixes https://github.com/nunit/nunit/issues/3767.

`Assert.That` for collections, error text changes to:

```
  Message: 
      Expected: exactly one item equal to "boop"
      But was:  **0 matching items** < "no-store, no-cache" >
```

by adding `x matching item/s` to the message text.

- [x] Add `matching` keyword to error message
- [x] Unit tests
- [x] Unit test for `no matching items` missing use-case


OsirisTerje (approved):
lgtm

PR #4704, opened 2024-05-06
Specify units in `After(int delayInMilliseconds, int pollingInterval)`

RenderMichael:
Although milliseconds is the natural unit for an `int` delay, specifying can still be helpful just to be sure.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Constraint.cs
+++ b/src/NUnitFramework/framework/Constraints/Constraint.cs
@@ -304,7 +304,7 @@ namespace NUnit.Framework.Constraints
         /// and polling interval.
         /// </summary>
         /// <param name="delayInMilliseconds">The delay in milliseconds.</param>
-        /// <param name="pollingInterval">The interval at which to test the constraint.</param>
+        /// <param name="pollingInterval">The interval at which to test the constraint, in milliseconds.</param>
         /// <returns></returns>
         public DelayedConstraint After(int delayInMilliseconds, int pollingInterval)
         {


src/NUnitFramework/framework/Constraints/Constraint.cs
```diff
@@ -304,7 +304,7 @@ public DelayedConstraint.WithRawDelayInterval After(int delay)
         /// and polling interval.
         /// </summary>
         /// <param name="delayInMilliseconds">The delay in milliseconds.</param>
-        /// <param name="pollingInterval">The interval at which to test the constraint.</param>
+        /// <param name="pollingInterval">The interval at which to test the constraint, in milliseconds.</param>
```
manfred-brands, line 307: It would be more consistent to rename the parameter into `pollingIntervalInMilliseconds` to match the delay one.

RenderMichael: If that’s an acceptable break, I’ll make that change next time I’m in front of a keyboard

manfred-brands: Previously I would have said yes, but thinking off it, if someone did `.After(100, pollingInterval: 10)` it would be a breaking change. So therefore all we can change is the comments.

RenderMichael: If it’s valuable enough, this method can be obsoleted in favor of the enhanced syntax `.After(100).Milliseconds.PollEvery(10).Milliseconds`. It would be the perfect scenario for an analyzer and fixer.

stevenaw (approved):
Looks great to me, thanks for your contribution @RenderMichael !
I'll leave this unmerged for now to allow @manfred-brands the opportunity to review too as they had been involved in earlier discussions.

manfred-brands (approved):
Approved. 
Maybe we can add more type safe version in some other PR and then hide this in the next major release.

PR #4706, opened 2024-05-07
Fixes Issue #4705 - Wrong version numbers in dlls

### LINKED ISSUE #4705: The dll's in the release 4.1 has version 4.0.1
This is caused by the hardcoding of the version numbers in the CommonAssemblyInfo.cs file. 
We're now using Cake.MinVer to generate the correct versionnumber, and sending that through MSBuild.  
This doesn't work for the assemblies (, but it works for the package), since we don't generate the assembly info (GenerateAssebmlyInfo=false).  If we set this to true, MSBuild can't override them.  We can't set it all from Cake since some information there is target framework dependent.  We can't roll that out by compiling for each frameworks, because our projects have different target frameworks.

So the resolution is to generate an AssemblyInfo.g.cs file in the build.cake script, and run that before the build starts.  This file is then added to the framework, framework.legacy and the nunitlite projects.   The others don't need any version.

Thanks to @CharliePoole for finding this!

OsirisTerje:
Fixes #4705


src/NUnitFramework/FrameworkVersion.cs
```diff
@@ -1,9 +1,9 @@
...
 
-using System.Reflection;
+//using System.Reflection;
 
 //
 // Current version for the NUnit Framework
 //
-[assembly: AssemblyVersion("4.0.1.0")]
-[assembly: AssemblyFileVersion("4.0.1.0")]
+//[assembly: AssemblyVersion("4.0.1.0")]
```
manfred-brands, line 8: Is removing this not enough?
We pass _version_ to MSBuild

OsirisTerje: In order for MSBuild to override the version the GenerateAssemblyInfo has to be set to true.  If set to false, the msbuild properties are ignored.  
If set to true, we must move all other assembly instructiosn that today are in the CommonAssemblyInfo.cs file into the csproj files.  Those that don't vary with framework could go into the direcoty.build.props file, but Configuration has to go into every csproj we have, since they run on different sets of target frameworks.  
That is doable, but the approach of simply generating a single file from the cake script seemed like an easier approach.

src/AssemblyInfo.g.cs
stevenaw, line 1: I haven't been able to pull and run the changes yet, but why commit this file to source control if it's generated? Would this get automatically updated and flagged as a change on future changes?

OsirisTerje: Right you are, we don't need to.  Need to put it into gitignore

src/NUnitFramework/FrameworkVersion.cs
OsirisTerje: @manfred-brands I've been thinking, and realize I did a "Design by surprise" here :-(  .  Flat on my face here.  And thanks for the kick!  ( I just tend to like code generation. ) 
So, I removed the code generation, added conditionals into directory.msbuild.props (perhaps a a bit too many, but they dont do any harm, and then i avoided  more compiler errors).  I then also removed most of the assemblyinfo.cs information and moved that to the csproj files.
And voila, all builds as it should, versions are correct and everything looks like normal again.

build.cake
```diff
@@ -128,17 +128,26 @@ Task("Build")
         DotNetBuild(SOLUTION_FILE, CreateDotNetBuildSettings());
     });
 
-DotNetBuildSettings CreateDotNetBuildSettings() =>
-    new DotNetBuildSettings
+DotNetBuildSettings CreateDotNetBuildSettings() 
+{
+    var version = packageVersion.ToString(); 
+    var assemblyVersion = version.Substring(0, version.LastIndexOf('-')) + ".0";
```
manfred-brands, line 134: The `version` only has a `-` if there is one in the tag. It won't be there for release builds causing this to fail.
The [MinVerVersion](https://www.cakebuild.net/api/Cake.MinVer/MinVerVersion/) class has a property `FileVersion` that does seem to do what you want here.

manfred-brands: I'm not sure why you set `AssemblyVersion` to the same as `FileVersion`?
That way code doesn't see a difference between any packages created between releases.

src/NUnitFramework/Directory.Build.props
```diff
@@ -35,6 +39,41 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
   </PropertyGroup>
 
+    <!-- .NET Core 3.1 -->
+    <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+      <AssemblyConfiguration>.NET Core 3.1 $(Configuration)</AssemblyConfiguration>
+    </PropertyGroup>
+
```
manfred-brands, line 46: We no longer support .NET Core 3.1.

src/NUnitFramework/Directory.Build.props
```diff
@@ -35,6 +39,41 @@
...
+    <!-- .NET Framework 4.6.2 -->
+    <PropertyGroup Condition="'$(TargetFramework)' == 'net462'">
+      <AssemblyConfiguration>.NET Framework 4.6.2 $(Configuration)</AssemblyConfiguration>
+    </PropertyGroup>
+
+    <!-- .NET 5.0 -->
+    <PropertyGroup Condition="'$(TargetFramework)' == 'net5.0'">
+      <AssemblyConfiguration>.NET 5.0 $(Configuration)</AssemblyConfiguration>
+    </PropertyGroup>
+
```
manfred-brands, line 56: We no longer support .NET 5.0

src/NUnitFramework/Directory.Build.props
```diff
@@ -35,6 +39,41 @@
...
+    <!-- .NET 8.0 -->
+    <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
+      <AssemblyConfiguration>.NET 8.0 $(Configuration)</AssemblyConfiguration>
+    </PropertyGroup>
+
+    <!-- .NET 9.0 -->
+    <PropertyGroup Condition="'$(TargetFramework)' == 'net9.0'">
+      <AssemblyConfiguration>.NET 9.0 $(Configuration)</AssemblyConfiguration>
+    </PropertyGroup>
+
```
manfred-brands, line 76: Nice forward thinking, but to prevent use not setting this when .net10 comes out.
Can I suggest to use [Choose](https://learn.microsoft.com/en-us/visualstudio/msbuild/choose-element-msbuild?view=vs-2022) instead? This allows an `Otherwise` clause where you can fail the build.

```xml
<Choose>
  <When Condition="'$(TargetFramework)' == 'net462'">
    <PropertyGroup>
      <AssemblyConfiguration>.NET Framework 4.6.2 $(Configuration)</AssemblyConfiguration>
    </PropertyGroup>
  </When>
  <When Condition="'$(TargetFramework)' == 'net8.0'">
    ...
  </When>
  <Otherwise>
    <Error Text="MissingAssemblyConfiguration for $(TargetFramework)" />
  <Otherwise>
</Choose>
```

Alternatively add a conditional error:
```xml
<Error Condition="'$(AssemblyConfiguration)' == ''" Text="MissingAssemblyConfiguration for $(TargetFramework)" />
```

src/NUnitFramework/framework/nunit.framework.csproj
```diff
@@ -7,6 +7,13 @@
     <IsTestProject>false</IsTestProject>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net462'">
+    <Title>NUnit Framework (.NET Framework 4.6.2)</Title>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <Title>NUnit Framework (.NET 6.0)</Title>
+  </PropertyGroup>
```
manfred-brands, line 15: Maybe we can use the previously set `AssemblyConfiguration` in order not to duplicate things:

```suggestion
  <PropertyGroup>
    <Title>NUnit Framework $(AssemblyConfiguration)</Title>
  </PropertyGroup>
```

src/NUnitFramework/nunit.framework.legacy/nunit.framework.legacy.csproj
```diff
@@ -6,14 +6,16 @@
...
-
   <ItemGroup>
     <ProjectReference Include="..\framework\nunit.framework.csproj" />
   </ItemGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net462'">
+    <Title>NUnit Framework Classic (.NET Framework 4.6.2)</Title>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <Title>NUnit Framework Classic (.NET 6.0)</Title>
+  </PropertyGroup>
```
manfred-brands, line 17: Maybe we can use the previously set `AssemblyConfiguration` in order not to duplicate things:

```suggestion
  <PropertyGroup>
    <Title>NUnit Framework Classic $(AssemblyConfiguration)</Title>
  </PropertyGroup>
```

src/NUnitFramework/nunitlite/nunitlite.csproj
```diff
@@ -6,13 +6,16 @@
...
-  <ItemGroup>
-    <ProjectReference Include="..\framework\nunit.framework.csproj" />
-  </ItemGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net462'">
+    <Title>NUnitLite Runner (.NET Framework 4.6.2)</Title>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <Title>NUnitLite Runner (.NET 6.0)</Title>
+  </PropertyGroup>
+
```
manfred-brands, line 15: Maybe we can use the previously set `AssemblyConfiguration` in order not to duplicate things:

```suggestion
  <PropertyGroup>
    <Title>NUnitLite Runner $(AssemblyConfiguration)</Title>
  </PropertyGroup>
```

manfred-brands (changes requested):
@OsirisTerje Sorry for the delay, I was off sick and couldn't concentrate enough to do reviews.
I also like code generation, but I do think your second option is nicer.
I think we can remove a lot of conditionals in the other projects by utilizing the `AssemblyConfiguration` set in the `Directroy.Build.props`

build.cake
OsirisTerje: The Minver FileVersion and Assemblyversion is by default not quite what I believe we want.  We could do it way, but we have never discussed it.  So I therefore just made them equal.  Minver is however not in line with the guidance they point to themselves at https://github.com/adamralph/minver?tab=readme-ov-file#version-numbers to https://learn.microsoft.com/en-ca/dotnet/standard/library-guidance/versioning#version-numbers.

One option is to keep the assembly version the way it is now, 3 first digits set and 4th = 0.  Then let FileVersion be with the 4th digit, which is normally the number of commits from the last tag, otherwise 0.

I am open to anything here actually, as long as we set them to something meaningful, and which matches the package version.

src/NUnitFramework/Directory.Build.props
OsirisTerje: I tried to add this, which looked very nice, but ended with a msbuild error when running restore.  VS says there is a syntax  error here somewhere, but I can't see it.  Might be something small.  Have commented out for now, but would be great if you had a look at it @manfred-brands .

build.cake
OsirisTerje: This is the latest results:  
image

src/NUnitFramework/Directory.Build.props
manfred-brands: Unfortunately the `Error` task cannot be placed in a `Choose`.
I added a commit to go with option B.

If a non-supported target is used it fails build with:
```
src\NUnitFramework\Directory.Build.props(71,5): error : Missing AssemblyConfiguration for net6.0-windows7 [src\NUnitFramework\windows-tests\windows-tests.csproj::TargetFramework=net6.0-windows7]
src\NUnitFramework\Directory.Build.props(71,5): error : Missing AssemblyConfiguration for net8.0-windows10.0.19041.0 [src\NUnitFramework\windows-tests\windows-tests.csproj::TargetFramework=net8.0-windows10.0.19041.0]
```

I fixed those by use `StartsWith` in the Conditions.

src/NUnitFramework/tests/CakeTests.cs
```diff
@@ -0,0 +1,52 @@
...
+    public void ThatWeCanParseAssemblyVersionStrings(string versionString, string expected)
+    {
+        var cake = new CakeCode();
+        var parsedVersion = cake.ParseAssemblyVersion(versionString);
+        Assert.That(parsedVersion, Is.EqualTo(expected));
+    }
+}
+
+/// <summary>
+/// This class is copied into the cake script. It is just here to be tested.
```
manfred-brands, line 27: How do we keep this in sync so we are testing what is used?

src/NUnitFramework/tests/Internal/Results/TestResultTests.cs
```diff
@@ -32,8 +32,8 @@ protected static void ReasonNodeExpectedValidation(TNode testNode, string reason
         {
             TNode? reason = testNode.SelectSingleNode("reason");
             Assert.That(reason, Is.Not.Null);
-            Assert.That(reason.SelectSingleNode("message"), Is.Not.Null);
-            Assert.That(reason.SelectSingleNode("message").Value, Is.EqualTo(reasonMessage));
+            Assert.That(reason!.SelectSingleNode("message"), Is.Not.Null);
+            Assert.That(reason.SelectSingleNode("message")!.Value, Is.EqualTo(reasonMessage));
```
manfred-brands, line 36: Null Suppression operators are not needed as the NUnit.Analyzer suppresses these because they have been checked for `Is.Not.Null`

src/NUnitFramework/tests/WorkItemTests.cs
```diff
@@ -29,7 +29,7 @@ public void CreateWorkItems()
         [TearDown]
         public void DisposeWorkItems()
         {
-            _workItem.Dispose();
+            _workItem?.Dispose();
```
manfred-brands, line 32: Not needed as the NUnit.Analyzer suppresses this because it knows `_workItem` was initialized in the `SetUp` method

src/NUnitFramework/TestBuilder.cs
```diff
@@ -11,213 +11,214 @@
...
-        {
-            TestSuite suite = MakeFixture(fixture.GetType());
-            suite.Fixture = fixture;
-            return suite;
-        }
+    public static TestSuite MakeParameterizedMethodSuite(Type type, string methodName)
+    {
+        var suite = MakeTestFromMethod(type, methodName) as TestSuite;
+        Assert.That(suite, Is.Not.Null, "Unable to create parameterized suite - most likely there is no data provided");
+        return suite!;
```
manfred-brands, line 44: No need for null suppression operator. NUnit.Analyzer does the supressing.

src/NUnitFramework/TestBuilder.cs
```diff
@@ -11,213 +11,214 @@
...
+    public static TestMethod MakeTestCase(Type type, string methodName)
+    {
+        var test = MakeTestFromMethod(type, methodName) as TestMethod;
+        Assert.That(test, Is.Not.Null, $"Unable to create TestMethod from {methodName}");
 
-        public static TestMethod MakeTestCase(Type type, string methodName)
-        {
-            var test = MakeTestFromMethod(type, methodName) as TestMethod;
-            Assert.That(test, Is.Not.Null, $"Unable to create TestMethod from {methodName}");
+        return test!;
```
manfred-brands, line 52: No need for null suppression operator. NUnit.Analyzer does the suppressing.

src/NUnitFramework/TestBuilder.cs
```diff
@@ -11,213 +11,214 @@
...
-            var context = new TestExecutionContext
-            {
-                TestObject = testObject,
-                Dispatcher = new SuperSimpleDispatcher()
-            };
+    public static WorkItem CreateWorkItem(Test test, TestExecutionContext context, IDebugger? debugger = null)
+    {
+        var work = WorkItemBuilder.CreateWorkItem(test, TestFilter.Empty, debugger ?? new DebuggerProxy(), true);
+        Assert.That(work, Is.Not.Null);
+        work!.InitializeContext(context);
```
manfred-brands, line 101: No need for null suppression operator. NUnit.Analyzer does the suppressing.

src/NUnitFramework/TestBuilder.cs
```diff
@@ -11,213 +11,214 @@
...
+    }
 
-            return RunTest(testMethod, testObject);
-        }
+    public static ITestResult RunAsTestCase(Action action)
+    {
+        var method = action.GetMethodInfo();
+        Assert.That(method, Is.Not.Null);
+        Assert.That(method.DeclaringType, Is.Not.Null);
+        var testMethod = MakeTestCase(method.DeclaringType!, method.Name);
```
manfred-brands, line 154: No need for null suppression operator. NUnit.Analyzer does the suppressing.

manfred-brands (changes requested):
You added various unnecessary null suppression operators (!).
All of them are unnecessary as NUnit.Analyzer does the suppressing for us as it knows those cannot be null.

src/NUnitFramework/tests/CakeTests.cs
OsirisTerje: I looked up how to include source in cake, but didnt see how to go all the way there.  Just wanted now to make sure the code was tested - to avoid missing something.  But we should include the code.  I can ask the Cake people later.

src/NUnitFramework/tests/Internal/Results/TestResultTests.cs
OsirisTerje: I believe I added them when the analyzers dozed off, they do that from time to time - on my VS......

OsirisTerje: Hmm..  I still get the squibblies:  
image
The analyzer is loaded, I get the NUnit2045 on the line below. 

I see that this warning comes from Resharper.  That means that also Rider will give this warning.

manfred-brands: ReSharper/Rider don't know about NUnit. They raise their own. This is from pre-C# nullability. C# is now very good at checking this. We dropped ReSharper for my work as it no longer added value.
Can you disable nullability checking in ReSharper for a project with nullability enabled? Should it do so by default to prevent double reporting?
You could raise an issue with ReSharper if they allow suppressing their errors if a 3rd party analyzer knows more.

src/NUnitFramework/tests/CakeTests.cs
OsirisTerje: Done

src/NUnitFramework/Directory.Build.props
manfred-brands: It should be this one:
```
Author: Manfred Brands <manfred-brands@users.noreply.github.com>
Date: 14/05/2024 17:47:34
Commit hash: b6448846040142fa792d9d900ad9e9a63ab0ee5b

Raise Error when compiling for a TargetFramework with no AssemblyConfiguration

.NET 7.0 is out of support as of May 14th 2024.
```

It is part of your merge commit.

CakeScripts.Tests/VersionParsersTests.cs
```diff
@@ -0,0 +1,23 @@
...
+        Assert.That(parsedVersion, Is.EqualTo(expected));
+    }
+
+    [TestCase("1.2.3-beta.1", "1.2.3.0")]
+    [TestCase("1.2.3", "1.2.3.0")]
+    public void ThatWeCanParseAssemblyVersionStrings(string versionString, string expected)
+    {
+        var parsedVersion = VersionParsers.ParseAssemblyVersion(versionString);
+        Assert.That(parsedVersion, Is.EqualTo(expected));
+    }
```
manfred-brands, line 22: I think that you have `Assembly` and `File` version the wrong way around.
As per [documentation](https://learn.microsoft.com/en-us/troubleshoot/developer/dotnet/framework/general/assembly-version-assembly-file-version#assemblyversion) the `AssemblyVersion` must be unique

manfred-brands (changes requested):
Still some issues.

src/NUnitFramework/Directory.Build.props
OsirisTerje: Thanks!

src/NUnitFramework/TestBuilder.cs
OsirisTerje: Now then.....

CakeScripts.Tests/Usings.cs
```diff
@@ -0,0 +1,3 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+global using NUnit.Framework;
```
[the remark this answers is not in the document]
OsirisTerje: It's coming off the generated code when using "dotnet new .....", so I have adjusted to accepting it.

src/NUnitFramework/tests/CakeTests.cs
```diff
@@ -0,0 +1,3 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+namespace NUnit.Framework.Tests;
```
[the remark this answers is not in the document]
OsirisTerje: Done

src/NUnitFramework/tests/Internal/Results/TestResultTests.cs
```diff
@@ -33,7 +33,7 @@ protected static void ReasonNodeExpectedValidation(TNode testNode, string reason
             TNode? reason = testNode.SelectSingleNode("reason");
             Assert.That(reason, Is.Not.Null);
             Assert.That(reason.SelectSingleNode("message"), Is.Not.Null);
-            Assert.That(reason.SelectSingleNode("message").Value, Is.EqualTo(reasonMessage));
+            Assert.That(reason.SelectSingleNode("message")!.Value, Is.EqualTo(reasonMessage));
```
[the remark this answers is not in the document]
OsirisTerje: Done.  Btw, reported the warning as a bug in resharper.  It was fixed in 2021, but seems to have reverted back.  Can't see anything in my setup that should cause it..

src/NUnitFramework/tests/WorkItemTests.cs
```diff
@@ -64,7 +64,7 @@ public void CanStopRun()
 
             Assert.Multiple(() =>
             {
-                Assert.That(_workItem.State, Is.EqualTo(WorkItemState.Complete));
+                Assert.That(_workItem?.State, Is.EqualTo(WorkItemState.Complete));
```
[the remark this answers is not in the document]
OsirisTerje: Done

CakeScripts.Tests/VersionParsersTests.cs
OsirisTerje: @manfred-brands The AssemblyVersion should indicate the Major.Minor version (above also patch in included, but we're not quite semver yet), and not change for each build or revision.   The FileVersion (aka AssemblyFileVersion) should change for all patches and revisons.
To me it looks like that is what the code above do.
Or do you see something I don't (not surprised though).

manfred-brands: The code does what you say it does. But my interpretation of the linked documentation is that the AssembyVersion should be unique so the .NET Resolver find the correct instance.
If one builds against 1.2.3-beta.1 then later against 1.2.3-beta.2 which has a bugfix, how would the compiler (and developer) know that we have the correct version if both are versioned as 1.2.3.0.
Maybe one of our native English (@stevenaw ?) speakers can read the linked documentation.

OsirisTerje: >  In central or final release builds it makes more sense to change the AssemblyVersion and most keep the AssemblyFileVersion same as assembly version.

Thanks @CharliePoole !  I then think this is the appropriate solution for us.

And Charlie - this does fix the issue.   The issue arised because we had to manually set these numbers, and that was forgotten.  Now they are automatically set together with the package version, and will follow the package version, given we now have a resolution for how these numbers should be :-)

OsirisTerje: This is actually not that easy.  We want the AssemblyVersion to be (equal to or) higher than the beta releases.  If we add the beta number as the 4th digit, that will be 0 for release, which means lower than the beta ones.
Earlier when we just did this manually, we kept the File and Assembly version numbers equal and fixed through the beta period.  
image
and
image

So if we continue to follow this practice we would not create anything more than we potentially have had for years.

manfred-brands (approved):
I raised one issue in the build.cake script.
It was mentioned originally in my first review but got lost in the further discussions.
I pushed up a commit for that.

One thing though is that if I build this locally it build version: `4.0.0-alpha.88`
So maybe the code relies on annotated tags instead of normal ones.
`git describe` returns `4.0.0-alpha-244-ga604a39ca`

I do see confusing tags in the repository some have a `v` prefix others don't
```
4.0.0-alpha
4.1.0
4.1.0-alpha.0
v4.0.0
v4.0.0-beta.1
v4.0.1
```

Note that the CI build doesn't fetch any tags and creates version `0.0.0.0`:

image

I added a commit to fetch the tags to see what it does. You can revert this as it isn't needed for CI builds only for publications.
It seems to create 4.2.0 versions:

image

MinVer must pick up the 4.1.0 tag and create the next release.
There are various properties for [MinVer](https://github.com/adamralph/minver/blob/main/README.md) we can use to tailor this.

build.cake
```diff
@@ -128,17 +129,29 @@ Task("Build")
...
```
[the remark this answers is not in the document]
manfred-brands: Pushed up in latest commit.

build.cake
```diff
@@ -128,17 +129,29 @@ Task("Build")
...
+    var msBuildSettings = new DotNetMSBuildSettings {
+        ContinuousIntegrationBuild = BuildSystem.GitHubActions.IsRunningOnGitHubActions,
+        AssemblyVersion = VersionParsers.ParseAssemblyVersion(version),
+        FileVersion = VersionParsers.ParseAssemblyVersion(version),
```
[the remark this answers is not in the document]
manfred-brands: Pushed up in latest commit.

CakeScripts.Tests/CakeScripts.Tests.csproj
```diff
@@ -0,0 +1,25 @@
...
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.9.0" />
+    <PackageReference Include="NUnit" Version="4.1.0" />
+    <PackageReference Include="NUnit3TestAdapter" version="4.5.0" />
```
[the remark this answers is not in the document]
OsirisTerje: Weird....  it does on cmd line,  maybe I added this using VS, but it should be the same template

CakeScripts/VersionParsers.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+public static class VersionParsers
+{
+    public static string ParseAssemblyVersion(string version)
+    {
+        var dash = version.LastIndexOf('-');
+        if (dash > 0)
+        {
+            return string.Concat(version.Substring(0, dash), ".0");
```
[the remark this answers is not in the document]
OsirisTerje: Hmmm  I thought we had got rid of the 'v' prefix now.  Seems something is still adding them.
Cake 4 sounds like a good idea. Better than downgrading to net462

manfred-brands:
@OsirisTerje for some reason `dotnet test` no longer finds any tests in `master` after this commit got merged.

If I go a commit before this `dotnet test` finds test, afterwards not anymore.
I looked at the changes and this doesn't change anything except versions.
Not sure how the adapter finds tests, but the adapter is the same in all cases, so it must be something in NUnit itself.
I tried `dotnet test -- NUnit.DebugExecution=true` but your blog seems to indicate it only works with a debug version of the adapter and my PC refused to execute any non-signed PS1 scripts.

@stevenaw  Can you confirm this behaviour?

stevenaw:
I think I can confirm. Based on the following to test:
```
dotnet clean
dotnet build
dotnet test
```

I tried checking out the previous merge commit (`376b434`) and could run all the tests that way. Presently, it seems like only the cake tests are running. This is a subset of the output I see presently:

```
Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
No test is available in C:\Users\sweer\Documents\Development\OSS-stevenaw\nunit\src\NUnitFramework\tests\bin\Debug\net8.0\nunit.framework.tests.dll. Make sure that test discoverer & executors are registered and platform & framework version settings are appropriate and try again.

Additionally, path to test adapters can be specified using /TestAdapterPath command. Example  /TestAdapterPath:<pathToCustomAdapters>.
```

OsirisTerje:
I do see the same using `dotnet test`, and `build --target=test` (cake tests) still works. 

How can version number changes do this ???

mikkelbu:
> How can version number changes do this ???

The only thing I can thing of - without debugging into the code and I don't know much about versioning in .NET - is that the `IsSupportedTestFramework` - https://github.com/nunit/nunit-console/blob/6ab7411d454ab2563426fdb60a9cdff7be655bb4/src/NUnitEngine/nunit.engine.core/Drivers/NUnit3DriverFactory.cs#L21-L24 - cannot determine that the major version is above or equal to 3.

Looking at the product versions it seems that we now add a lot of "prerelease" information that might confuse the validation (comparing NUnit.4.2.0-alpha.0.17.nupkg and NUnit.4.2.0-alpha.0.21.nupkg from myget)
image
image

OsirisTerje:
Just checking:  Product Version == Informational Version, and should into be used in the comparison.
FileVersion is the one being returned in the check in the engine, so it **should** work (I know, famous last words.....)

Am I right or wrong here?

manfred-brands:
Ok, I figured it out.  @mikkelbu pointed me in the right direction.
When building nunit from within visual studio no version number is created at all.

I added the following to `Directory.Build.props`:

```xml
    <Version Condition="'$(Version)'==''">4.0.0.0</Version>
```

Now when compiling with `cake`, I get version `4.2.0.0` and when building inside VS, it is version `4.0.0.0`
I'll put a PR up shortly.

PR #4708, opened 2024-05-08
Backport tests from PR #4692 to 3.x

stevenaw:
Backport tests from https://github.com/nunit/nunit/pull/4692
To verify the functionality from 3.x that we're restoring. Relates to https://github.com/nunit/nunit/issues/4598

This PR also switches the macos image on github actions to target specifically `macos-12` in order to work around this NET5-related issue: https://github.com/dotnet/runtime/issues/48068. It seems this may be the first time we're running CI on the 3.x branch since `macos-latest` was changed to `macos-14` and the newer silicon architecture.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -107,7 +107,7 @@ jobs:
 
   build-macos:
     name: MacOS Build
-    runs-on: macos-latest
+    runs-on: macos-12
 
     steps:
     - name: ⤵️ Checkout Source
--- a/src/NUnitFramework/testdata/TimeoutFixture.cs
+++ b/src/NUnitFramework/testdata/TimeoutFixture.cs
@@ -21,7 +21,6 @@
 // WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 // ***********************************************************************
 
-#if THREAD_ABORT
 using System;
 using System.Runtime.InteropServices;
 using System.Threading;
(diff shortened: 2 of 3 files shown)


stevenaw:
@nunit/framework-team 
I've noticed net5 tests still running on the 3.x branch are now failing on macos. That runtime has been EOL for a while and is no longer used in main. Any objections to me removing it from 3.x either within this PR or separately?

OsirisTerje:
I believe we have stated that 3.X will continue to support these OOD frameworks. 
Can't we just disable that test from MacOS, and make a note of it ?

stevenaw:
Thanks @OsirisTerje I can try to take a closer look.

stevenaw:
Huh interesting. Thanks for having me take a closer look here @OsirisTerje 
I had thought it was a `NET5` issue but the real thing here was that `macos-latest` switched to macos-14 which uses newer Apple Silicon / M1+. It appears this was a known issue a few years back with NET5 not working there in all cases: https://github.com/dotnet/runtime/issues/48068

Switching the image from `macos-latest` down to `macos-12` seems to have solved it. `macos-12` is still supported and not yet deprecated based on https://github.com/actions/runner-images?tab=readme-ov-file.

I think this is rady to switch out of `Draft`

OsirisTerje:
@stevenaw Awesome!  And agree!

OsirisTerje (approved):
LGTM

stevenaw:
Thanks for merging!

Issue #4134, opened 2022-06-22
NUnit 3 console does not display logs from background thread from a library


Peter-K10:
Hi,

We switched from NUnit 2 to NUnit 3, but we have some issues with displaying logs from background thread in the console.
We have a dynamic library (written in C++, for which we created a CLI wrapper to be able to use it in C# projects) that we are testing with NUnit app. This library provides a logger object on the interface that can be used to display the logs from it, so we are using log4net to display the library logs to the console (using a custom ConsoleAppender) and to a file (using RollingFileAppender). 

We are using a custom log4net console appender in order to display the library log lines in real time in the console using the TestContext.Progress.Write() method as mentioned here https://github.com/nunit/nunit/issues/1139 (this is needed because NUnit 3 is displaying the console output only when the test is finished and this is not acceptable for us):
```
public class NUnit3ConsoleAppender : log4net.Appender.ConsoleAppender
{
  override protected void Append(log4net.Core.LoggingEvent loggingEvent)
   {
      TestContext.Progress.Write(RenderLoggingEvent(loggingEvent));
   }
}
```

In the app.config file we have:
```
<appender name="NUnit3ConsoleAppender" type="MyNamespace.NUnit3ConsoleAppender, MyNamespace">
...
<root>
  <level value="ALL" />
  <appender-ref ref="NUnit3ConsoleAppender" />
</root>
```

The main problem is that the **library log lines from a background thread are not displayed in the console**, only the library log lines from the main thread are displayed in the console. We have this issue when executing the tests with 'NUnit Console 3.15' and also with 'TestCentric 2.0.0-alpha6'.
Interesting is that the library log lines from the background thread correctly appear in the file where we are saving the output using log4net, they are only missing in the console.
Even if we don't use a custom log4net console appender but use the existing log4net console appender (appender name="ConsoleAppender" type="log4net.Appender.ConsoleAppender") which displayed the logs to the console after the NUnit test has finished, the library log lines from the background thread are still missing from the console.

With NUnit 2 (v2.6.4) we don't have this issue, only with NUnit 3.

We saw that NUnit 3 provides support for parallel test execution and that's why NUnit 3 is capturing the output and then prints it to the console when the test finishes running. That way, the output is matched with the test, not interleaved with other tests.
We do not want to execute the tests in parallel, so we tried adding [NonParallelizable] to the NUnit test and also [SingleThreadedAttribute] to the TestFixture, but it didn't help.

[...]

PR #4709, opened 2024-05-11
Issue #4686: Add ProgressTraceListener

### LINKED ISSUE #4686: Provide a ProgressTraceListener to redirect Trace output to the NUnit Progress output
This issue is based on discussions at https://github.com/TestCentric/testcentric-gui/issues/1043 and https://github.com/nunit/nunit3-vs-adapter/issues/718 but more belongs to NUnit itself.

Similar to [VS Test Adapter documentation](https://docs.nunit.org/articles/vs-test-adapter/Trace-and-Debug.html) which describes a workaround to activate Trace and Debug output, @CharliePoole has hinted "possible to create a listener which would capture the output and send it as a Progress event or write it to stderr, but I don't believe such a thing exists at this time" which I implemented at https://github.com/TestCentric/testcentric-gui/issues/1043#issuecomment-2043534958. It is simple and works.

Same as the [ConsoleTraceListener](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.consoletracelistener) class is provided by the .NET Framework, I propose the NUnit framework provides this NUnit specific `ProgressTraceListener` ready for use by the NUnit users. I don't ask NUnit to always have this listener active, it shall only provide the class and document its purpose and how to use it.

maettu-this:
As proposed in https://github.com/nunit/nunit/issues/4686, resolves #4686
Also fixes https://github.com/nunit/nunit/issues/4134.
Documentation update with separate docs PR.


maettu-this:
I guess I need some support here. The Windows and MacOS tests indicate "1 tests failed" but searching the log for neither "fail" nor "Trace" nor "Progress" sheds any light into the failure.

On Linux, one of the newly added tests indeed fails:
```
1) Failed : NUnit.Framework.Tests.Diagnostics.ProgressTraceListenerTests.TestDebugIsDirectedToOutput 
Assert.That(TestResultListener.Outputs, Has.Count.EqualTo(1)) 
Expected: property Count equal to 1 
But was: 0 
at NUnit.Framework.Tests.Diagnostics.ProgressTraceListenerTests.TestDebugIsDirectedToOutput() in /_/src/NUnitFramework/tests/Diagnostics/ProgressTraceListenerTests.cs:line 108 
```
But the question is, why does this behave differently on Linux?

maettu-this:
Review consideration regarding potential location of the listener:

- `NUnit.Framework.Diagnostics.ProgressTraceListener` (new namespace) pro: orthogonality with `
System.Diagnostics.ConsoleTraceListener`
- `NUnit.Framework.TestContext.ProgressTraceListener` (class within `TestContext`) pro: in proximity with `TestContext.Progress`
- `NUnit.Framework.ProgressTraceListener` (top-level) pro: simplicity

maettu-this:
@OsirisTerje thanks for re-triggering the workflows. I have found the culprit, of course the `System.Diagnostics.Debug` related tests must be `DEBUG` conditional. Please trigger the workflows once more.

OsirisTerje:
@stevenaw @manfred-brands  Is this ok with you guys?  It's a new namespace but it may be fitting.  I am good with this.

src/NUnitFramework/framework/Diagnostics/ProgressTraceListener.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// "directs tracing or debugging output".
+    /// </para><para>
+    /// This listener is provided by NUnit (i.e. the origin of
+    /// <see cref="TestContext.Progress"/>) same as the
+    /// <see cref="ConsoleTraceListener"/> is provided by .NET
+    /// (the origin of <see cref="Console"/>).
+    /// </para></remarks>
+    // Implemented following the ConsoleTraceListener implementation.
+#if !NET6_0_OR_GREATER
+    [HostProtection(Synchronization = true)]
```
stevenaw, line 44: @maettu-this I'm curious, is this needed? I haven't used this myself, but the documentation I can find indicates this could be [CAS-related](https://learn.microsoft.com/en-us/dotnet/api/system.security.permissions.hostprotectionattribute?view=net-8.0) which is something I think we tried to move away from in NUnit 4

stevenaw:
@OsirisTerje The new namespace is a good point. This appears to be the first time we're explicitly adding a new writer with external intent. Current writers are all in the `Internal` namespace. 

I don't have strong opinions on it either way so I think the namespacing in this PR is fine until we decide on long term plans for some of the other Internal-namespaced items.

src/NUnitFramework/framework/Diagnostics/ProgressTraceListener.cs
```diff
@@ -0,0 +1,60 @@
...
+{
+    /// <summary><para>
+    /// The ProgressTraceListener class allows directing tracing or
+    /// debugging output to <see cref="TestContext.Progress"/>.
+    /// </para><para>
+    /// To activate, place the following snippet into the one-time
+    /// set-up method of either a test's fixture or the set-up
+    /// fixture of a project:
+    /// <c>
+    /// System.Trace.Listeners.Add(new ProgressTraceListener());
```
manfred-brands, line 19:
```suggestion
    /// _progressTraceListener = new ProgressTraceListener();
    /// System.Trace.Listeners.Add(_progressTraceListener);
```

This better aligns with the suggested tear-down below.

src/NUnitFramework/framework/Diagnostics/ProgressTraceListener.cs
manfred-brands: Even the .NET Framework [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.security.permissions.hostprotectionattribute?view=netframework-4.6.2#remarks) says it has been deprecated. 
I do see `HostProtection` in the `ConsoleTraceListener`  .NET Framework reference [source](https://github.com/microsoft/referencesource/blob/master/System/compmod/system/diagnostics/ConsoleTraceListener.cs)

src/NUnitFramework/framework/Diagnostics/ProgressTraceListener.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </para><para>
+    /// Make sure to only add a listener once, e.g.:
+    /// <c>
+    /// if (!System.Trace.Listeners.OfType&lt;ProgressTraceListener&gt;().Any())
+    ///     System.Trace.Listeners.Add(new ProgressTraceListener());
+    /// </c>
+    /// </para><para>
+    /// Or again remove in the one-time tear-down, e.g.:
+    /// <c>
+    /// System.Trace.Listeners.Remove(_progressTraceListener);
```
manfred-brands, line 30:
```suggestion
    /// System.Trace.Listeners.Remove(_progressTraceListener);
    _progressTraceListener.Dispose();
```

The listener is `IDisposable`.

manfred-brands (changes requested):
A few small changes suggested to the XML documentation to better reflect what is needed and what is actually used in the `ProgressTraceListenerTests`

src/NUnitFramework/framework/Diagnostics/ProgressTraceListener.cs
maettu-this: @stevenaw and @manfred-brands, for this very reason I added the...

> // Implemented following the ConsoleTraceListener implementation.

...comment at line 42. I have neither used this attribute myself, but I am pretty sure the .NET framework guys have (or at least had) reasons for this.

src/NUnitFramework/framework/Diagnostics/ProgressTraceListener.cs
maettu-this: Well, the tear-down is only an alternative proposal:

> Or again remove...

The standard use can remain...

> System.Trace.Listeners.Add(new ProgressTraceListener());

...and let the listener be [closed by the framework](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.trace.close):

> Flushes the output buffer, and then closes the [Listeners].

I'll push another proposal to make the comment clearer.

src/NUnitFramework/framework/Diagnostics/ProgressTraceListener.cs
maettu-this: According to https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.textwritertracelistener.dispose, `Close()` should be used instead of `Dispose()`.

src/NUnitFramework/framework/Diagnostics/ProgressTraceListener.cs
manfred-brands: I prefer the user that creates the object is responsible for disposal.

src/NUnitFramework/framework/Diagnostics/ProgressTraceListener.cs
manfred-brands: We can leave it in for now.

manfred-brands (approved):
Thanks @maettu-this, looks good to me now.

src/NUnitFramework/framework/Diagnostics/ProgressTraceListener.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// "directs tracing or debugging output".
+    /// </para><para>
+    /// This listener is provided by NUnit (i.e. the origin of
+    /// <see cref="TestContext.Progress"/>) same as the
+    /// <see cref="ConsoleTraceListener"/> is provided by .NET
+    /// (the origin of <see cref="Console"/>).
+    /// </para></remarks>
+    // Implemented following the ConsoleTraceListener implementation.
+#if !NET6_0_OR_GREATER
+    [HostProtection(Synchronization = true)]
```
stevenaw, line 50: I think I'm unclear if you needed this for your specific hosting requirements but I think I would prefer we remove this unless you explicitly need this @maettu-this , at which point we should have a discussion before adding it. 

The attribute, and code access security in general are deprecated, and the documentation mentions 
> This attribute affects only unmanaged applications that host the common language runtime and implement host protection, such as SQL Server. If the code is run in a client application or on a server that is not host-protected, the attribute "evaporates"; it is not detected and therefore not applied. 

Though even the docs for hosting CLR in SQL Server also mention it is no longer supported there
https://learn.microsoft.com/en-us/sql/relational-databases/clr-integration/common-language-runtime-integration-overview?view=sql-server-ver16#code-access-security-no-longer-supported

Deprecated functionality has a way of being "sticky" and being hard to remove which may be why we still see it in reference source for the full framework. It was indeed sticky when we added CAS / partial trust support in NUnit as it took us 8 years and a major version before it could be removed.

stevenaw (changes requested):

src/NUnitFramework/framework/Diagnostics/ProgressTraceListener.cs
maettu-this: I'm fine with removing it, I'll push the removal in a minute.

stevenaw:
@OsirisTerje The nuget warnings are odd. I've tried to return the job. 

@maettu-this the other failures may just be related to unused using after having removed the attribute. Thank you for removing that! Seems like maybe just a bit more related cleanup

stevenaw:
Thanks @manfred-brands 
Been tough to push changes myself, most of these reviews have been from my phone. I see it building now, will approve and merge after it passes 🙂

stevenaw (approved):
The build is now passing!
Thank you for the contribution @maettu-this this looks good to me now too

PR #4712, opened 2024-05-15
PropertiesComparer: Show actual property and value that fails equality

### LINKED ISSUE #4710: Improve PropertiesComparer diagnostics
Can the `PropertiesComparer` return _which_ property was unequal? [It seems it currently fails without that information:](https://github.com/nunit/nunit/blob/4eaab2eef3713907ca37bfb2f7f47e3fc2785214/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs#L77)

```csharp
                        if (result == EqualMethodResult.ComparedNotEqual)
                            return result;
```

When a type has multiple properties, this only outputs the name of the type (or whatever `ToString()` happens to return), e.g.

```
Assert.That(result[0], Is.EqualTo(testData.ExpectedResult[0]).UsingPropertiesComparer())
  Expected: <MyTypeName>
  But was:  <MyTypeName>
```

manfred-brands:
Fixes #4710

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
@@ -39,6 +39,8 @@ namespace NUnit.Framework.Constraints.Comparers
 
             ComparisonState comparisonState = state.PushComparison(x, y);
 
+            string declaringTypeName = xType.Name;
+
             uint redoWithoutTolerance = 0x0;
             if (tolerance.HasVariance)
             {
@@ -50,9 +52,13 @@ namespace NUnit.Framework.Constraints.Comparers
 
                     EqualMethodResult result = equalityComparer.AreEqual(xPropertyValue, yPropertyValue, ref tolerance, comparisonState);
                     if (result == EqualMethodResult.ComparedNotEqual)
-                        return result;
-                    if (result == EqualMethodResult.ToleranceNotSupported)
+                    {
+                        return PropertyNotEqualResult(equalityComparer, i, declaringTypeName, property.Name, xPropertyValue, yPropertyValue);
+                    }
+                    else if (result == EqualMethodResult.ToleranceNotSupported)
+                    {
                         redoWithoutTolerance |= 1U << i;
+                    }
                 }
 
                 if (redoWithoutTolerance == (1U << properties.Length) - 1)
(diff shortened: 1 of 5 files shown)


src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -271,6 +274,11 @@ public sealed class FailurePoint
             /// </summary>
             public long Position;
 
+            /// <summary>
+            /// The name of the property.
+            /// </summary>
+            public string? PropertyName;
```
OsirisTerje, line 280: Could this be just a string, default empty string, to avoid nulls.

src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
```diff
@@ -17,8 +17,9 @@ public class EqualConstraintResult : ConstraintResult
         private readonly Tolerance _tolerance;
         private readonly bool _caseInsensitive;
         private readonly bool _ignoringWhiteSpace;
+        private readonly bool _comparingProperties;
         private readonly bool _clipStrings;
-        private readonly IList<NUnitEqualityComparer.FailurePoint> _failurePoints;
+        private readonly IList<NUnitEqualityComparer.FailurePoint>? _failurePoints;
```
OsirisTerje, line 22: Does this need to be null?  It could be just an empty list.  Would also simplify some code below.

src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
```diff
@@ -50,9 +52,13 @@ public static EqualMethodResult Equal(object x, object y, ref Tolerance toleranc
 
                     EqualMethodResult result = equalityComparer.AreEqual(xPropertyValue, yPropertyValue, ref tolerance, comparisonState);
                     if (result == EqualMethodResult.ComparedNotEqual)
-                        return result;
-                    if (result == EqualMethodResult.ToleranceNotSupported)
+                    {
+                        return PropertyNotEqualResult(equalityComparer, i, declaringTypeName, property.Name, xPropertyValue, yPropertyValue);
+                    }
+                    else if (result == EqualMethodResult.ToleranceNotSupported)
```
OsirisTerje, line 58: Could drop the 'else' here

OsirisTerje (approved):
Looks good, a few small nitpicks.

And, consider the nulls.

src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
manfred-brands: Yes. Removed.

src/NUnitFramework/framework/Constraints/EqualConstraintResult.cs
manfred-brands: This change was deliberate. Previously the `NUnitEqualityComparer` would allocate a new list for every `AreEqual` call, regardless if there were ever any failure points. Now it is only allocated when needed.
This should speed up all `AreEqual` calls, especially assuming that 99.9% succeed and only a fraction of the failing ones even have failure points.

However I have updated the code in this class to remove `null` and initialize it with `Array.Empty`.
Now various null checks can be removed. Thanks for that.

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
manfred-brands: It could, but it would mean that all places that create a FailurePoint get an (implicit) unnecessary extra initialization.
I find a check for `null` clearer than one for `string.IsNullOrEmpty`.
Also the compiler tells me when I don't to the `null` check, not when I miss the empty check.

Non-nullable is good for Collectons, so I updated the nullable failurePoint you mentioned where

PR #4286, opened 2023-02-03
WIP: Add Coverlet Coverage to builds

SeanKilleen:
Resolves #423.

Putting this out here in super-draft form just to get it started. 

* Updates Cake to v3 and adjusts cake file to new conventions since v1
* Introduces `Cake.Coverlet` as an add-in
* Adds Coverlet collector as part of the tool manifest
* Updates the testing task to use Cake command & coverlet settings

PR #4722, opened 2024-06-03
WIP: Add Coverlet Coverage to builds

SeanKilleen:
A continuation of #4286, which got stale.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -19,6 +19,12 @@
       "commands": [
         "SignClient"
       ]
+    },
+    "coverlet.console": {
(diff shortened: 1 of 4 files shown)


SeanKilleen:
@stevenaw FYI, I restarted things here. Rough weekend, didn't get nearly as much time as I thought I would. But at least it's no longer stale, so, small win? 😅 

Next up, I'll look through the cake file to understand where we're outputting our test assemblies and will figure out how to run coverlet on them (since we're not using `dotnet test` which makes sense, I'll probably look into how to use the Cake CLI add-in to do what we need to do.

PR #4727, opened 2024-06-09
Runs .NET Core Timeout test in IsolatedContext

### LINKED ISSUE #4723: CurrentContext.Result.Outcome.Status is inconclusive in TearDown if Timeout attribute is used in 4.x
Hi,

I know timeout attribute is deprecated in 4.x but it still seems to be a regression compared to 3.x.
It seems when using the Timeout attribute then `TestContext.CurrentContext.Result.Outcome.Status` is reported as `Inconclusive` in the Teardown method while it used to be reported as `Passed` before.

Example:
```
        [Test, Timeout(20_000)]  // Non Ok status will be Inconclusive
        public async Test_Timeout() {
              await Task.Delay(10_000);
              Assert.Pass();
        }

        [Test, CancelAfter(20_000)] // Ok status will be Passed
        public async Test_CancelAfter(CancellatonToken ct) {
              await Task.Delay(10_000, ct);
              Assert.Pass();
        }

        [Test ]  // Ok status will be Passed
        public async Test() {
              await Task.Delay(10_000);
              Assert.Pass();
        }
        [TearDown]
        public async Task Cleanup()
        {
                Console.WriteLine($"Test status = {TestContext.CurrentContext.Result.Outcome.Status}");
        }
```

Thank you!

manfred-brands:
Fixes #4723

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
@@ -94,23 +94,21 @@ namespace NUnit.Framework.Internal.Commands
         {
             try
             {
-                var separateContext = new TestExecutionContext(context)
+                using (new TestExecutionContext.IsolatedContext())
                 {
-                    CurrentResult = context.CurrentTest.MakeTestResult()
-                };
-                var testExecution = Task.Run(() => innerCommand.Execute(separateContext));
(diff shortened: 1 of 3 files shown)


stevenaw (approved):
Thanks @manfred-brands 
Good catch to have thought of this. Good to merge if you are.

Issue #3359, opened 2019-08-28
Add XML documentation on SetName letting folks know about SetArgDisplayNames


jnm2:
People seem to mainly use SetName to get argument values in the test name, e.g. https://github.com/nunit/nunit3-vs-adapter/issues/650. That's what I see most frequently.

SetArgDisplayNames is an even better fit for this use case since you don't have to specify `{m}` (or `{{m}}` in an interpolated string). It doesn't lose the namespace information, or whatever is causing `SetNames` to wreak havoc on prefiltering in the issues folks are opening over at the adapter repo. And it doesn't force `{m}` or `{0}` or a dozen other tokens to be replaced.

I think we should have an XML doc showing example syntax.

OsirisTerje:
@jnm2 Would also be good to add this to the nunit docs for TestCaseData which documents SetName, and doesn't mention SetArgDisplayNames. 

I also have some blogposts using SetName, and I will go about correcting them :-)

OsirisTerje:
@mikkelbu are you going to fix this, or should we let it just be a Good First Issue?

mikkelbu:
@OsirisTerje I've just unsigned as I have very little time (as usual)

PapisKang:
Hi. I m new overthere (new to Open Source too). 
I would like to take this one. 
I ve already red then Contributing and the best practices for XML doc and i think i get what i have to do.
But before i submit any PR, i would to discuss about it. 

1.  Pull Request on nunit 
I m willing to edit  the XML documentation on SetName by adding a remark tag
`/// <remarks>
/// Consider using <see cref="SetArgDisplayNames"/>for setting argument values in the test name. 
/// <see cref="SetArgDisplayNames"/> allows you to specify the display names for parameters directly without 
/// needing to use tokens like <c>{m}</c>.
/// </remarks>`

I want to add on SetArgDisplayNames XML documentation an example tag 
`   /// <example>
    /// Example usage:
    /// <code>
    /// TestCaseData testCase = new TestCaseData(args)
    ///     .SetArgDisplayNames("arg1DisplayName", "arg2DisplayName");
    /// </code>
    /// </example>`

2. Pull Request on docs 
I ll modify TestCaseData.md and add under "Returns"

` SetArgDisplayNames(params string[] displayNames): Sets the list of display names to use as the parameters in the test name.`

stevenaw:
Welcome @PapisKang !
[...]

PR #4732, opened 2024-06-14
Issue-3359 : Add XML documentation on SetName letting folks know abou…

PapisKang:
Fixes https://github.com/nunit/nunit/issues/3359
For setting argument values in the test name, poeple should consider using SetArgDisplayNames instead of SetNames. 
It allows to specify the display names for parameters directly without  needing to use tokens like {m} or {0}. This method preserves namespace information and avoids issues related to prefiltering that can occur with SetNames


OsirisTerje (approved):
lgtm

PR #4737, opened 2024-06-25
Change byte allocation to be on the stack instead of the heap for performance improvements

### LINKED ISSUE #4735: StreamComparer - Allocating the byte array on the stack reduces memory allocation by 96%
In .net 8, instead of allocating the byte array on the heap, if we were to allocate it on stack using `stackalloc` it can provide some performance improvements based on the amount of memory allocated and therefore the GC runs.

The changes are as follows (for [line 42](https://github.com/nunit/nunit/blob/78523ddfc127bade516c9012469c66807bf66c36/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs#L42)) 
```c#
byte[] bufferExpected = new byte[BUFFER_SIZE];
byte[] bufferActual = new byte[BUFFER_SIZE];
```
to 
```c#
Span<byte> bufferExpected = stackalloc byte[BUFFER_SIZE];
Span<byte> bufferActual = stackalloc byte[BUFFER_SIZE];
```

This would also require some changes on [line 68](https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs#L68) as `BinaryReader.Read(Span<>)` doesn't exist in .net framework (https://learn.microsoft.com/en-us/dotnet/api/system.io.binaryreader.read?view=net-8.0) - so was thinking of using the `#IF NETFRAMEWORK` flag to keep the current implementation for .net framework.

Here is the performance benchmark in .net 8 for both of these methods:
```
BenchmarkDotNet v0.13.12, Windows 10 (10.0.19045.4529/22H2/2022Update)
AMD Ryzen 5 5600X, 1 CPU, 12 logical and 6 physical cores
.NET SDK 8.0.300
  [Host] : .NET 8.0.5 (8.0.524.21615), X64 RyuJIT AVX2


| Method       | Size  | Mean     | Error     | StdDev    | Ratio | RatioSD | Gen0   | Gen1   | Allocated | Alloc Ratio |
|------------- |------ |---------:|----------:|----------:|------:|--------:|-------:|-------:|----------:|------------:|
| NewByteArray | 32768 | 1.784 μs | 0.0346 μs | 0.0517 μs |  1.00 |    0.00 | 0.5112 | 0.0114 |    8560 B |        1.00 |
| StackAlloc   | 32768 | 1.344 μs | 0.0267 μs | 0.0408 μs |  0.75 |    0.03 | 0.0191 |      - |     320 B |        0.04 |
```

<details>
<summary>Benchmark</summary>

```c#

BenchmarkRunner.Run<ByteAllocationBenchmark>();

[MemoryDiagnoser]
public class ByteAllocationBenchmark
{
	private const int BUFFER_SIZE = 4096;

	[Params(4096 * 8)]
	public int Size { get; set; }
	private Stream? XStream { get; set; }
	private Stream? YStream { get; set; }

	[GlobalSetup]
	public void Setup()
	{
		var buffer = new byte[BUFFER_SIZE];

		XStream = new MemoryStream(Size);
		for (var i = 0; i < Size; i += BUFFER_SIZE)
			XStream.Write(buffer, 0, BUFFER_SIZE);
		XStream.Seek(0, SeekOrigin.Begin);

		YStream = new MemoryStream(Size);
		for (var i = 0; i < Size; i += BUFFER_SIZE)
			YStream.Write(buffer, 0, BUFFER_SIZE);
		YStream.Seek(0, SeekOrigin.Begin);
	}

	[GlobalCleanup]
	public void Cleanup()
	{
		XStream?.Dispose();

		YStream?.Dispose();
	}

	[Benchmark(Baseline = true)]
	public bool NewByteArray()
	{
		var equal = Equal_Original(XStream!, YStream!, out _);
		return equal == EqualMethodResult.ComparedEqual;
	}

	[Benchmark]
	public bool StackAlloc()
	{
		var equal = Equal_Enhanced(XStream!, YStream!, out _);
		return equal == EqualMethodResult.ComparedEqual;
	}

	public static EqualMethodResult Equal_Original(Stream xStream, Stream yStream, out long? failurePoint)
	{
		failurePoint = null;
		bool bothSeekable = xStream.CanSeek && yStream.CanSeek;

		if (bothSeekable && xStream.Length != yStream.Length)
			return EqualMethodResult.ComparedNotEqual;

		byte[] bufferExpected = new byte[BUFFER_SIZE];
		byte[] bufferActual = new byte[BUFFER_SIZE];

		BinaryReader binaryReaderExpected = new BinaryReader(xStream);
		BinaryReader binaryReaderActual = new BinaryReader(yStream);

		long expectedPosition = bothSeekable ? xStream.Position : default;
		long actualPosition = bothSeekable ? yStream.Position : default;

		try
		{
			if (xStream.CanSeek)
			{
				binaryReaderExpected.BaseStream.Seek(0, SeekOrigin.Begin);
			}
			if (yStream.CanSeek)
			{
				binaryReaderActual.BaseStream.Seek(0, SeekOrigin.Begin);
			}

			int readExpected = 1;
			int readActual = 1;
			long readByte = 0;

			while (readExpected > 0 && readActual > 0)
			{
				readExpected = binaryReaderExpected.Read(bufferExpected, 0, BUFFER_SIZE);
				readActual = binaryReaderActual.Read(bufferActual, 0, BUFFER_SIZE);

				if (MemoryExtensions.SequenceEqual<byte>(bufferExpected, bufferActual))
				{
					readByte += BUFFER_SIZE;
					continue;
				}

				for (int count = 0; count < BUFFER_SIZE; ++count)
				{
					if (bufferExpected[count] != bufferActual[count])
					{
						failurePoint = readByte + count;
						return EqualMethodResult.ComparedNotEqual;
					}
				}
				readByte += BUFFER_SIZE;
			}
		}
		finally
		{
			if (xStream.CanSeek)
			{
				xStream.Position = expectedPosition;
			}
			if (yStream.CanSeek)
			{
				yStream.Position = actualPosition;
			}
		}

		return EqualMethodResult.ComparedEqual;
	}

	public static EqualMethodResult Equal_Enhanced(Stream xStream, Stream yStream, out long? failurePoint)
	{
		failurePoint = null;
		bool bothSeekable = xStream.CanSeek && yStream.CanSeek;

		if (bothSeekable && xStream.Length != yStream.Length)
			return EqualMethodResult.ComparedNotEqual;

		Span<byte> bufferExpected = stackalloc byte[BUFFER_SIZE];
		Span<byte> bufferActual = stackalloc byte[BUFFER_SIZE];

		BinaryReader binaryReaderExpected = new BinaryReader(xStream);
		BinaryReader binaryReaderActual = new BinaryReader(yStream);

		long expectedPosition = bothSeekable ? xStream.Position : default;
		long actualPosition = bothSeekable ? yStream.Position : default;

		try
		{
			if (xStream.CanSeek)
			{
				binaryReaderExpected.BaseStream.Seek(0, SeekOrigin.Begin);
			}
			if (yStream.CanSeek)
			{
				binaryReaderActual.BaseStream.Seek(0, SeekOrigin.Begin);
			}

			int readExpected = 1;
			int readActual = 1;
			long readByte = 0;

			while (readExpected > 0 && readActual > 0)
			{
				readExpected = binaryReaderExpected.Read(bufferExpected);
				readActual = binaryReaderActual.Read(bufferActual);

				if (MemoryExtensions.SequenceEqual<byte>(bufferExpected, bufferActual))
				{
					readByte += BUFFER_SIZE;
					continue;
				}

				for (int count = 0; count < BUFFER_SIZE; ++count)
				{
					if (bufferExpected[count] != bufferActual[count])
					{
						failurePoint = readByte + count;
						return EqualMethodResult.ComparedNotEqual;
					}
				}
				readByte += BUFFER_SIZE;
			}
		}
		finally
		{
			if (xStream.CanSeek)
			{
				xStream.Position = expectedPosition;
			}
			if (yStream.CanSeek)
			{
				yStream.Position = actualPosition;
			}
		}

		return EqualMethodResult.ComparedEqual;
	}

	/// <summary>
	/// Result of the Equal comparison method.
	/// </summary>
	public enum EqualMethodResult
	{
		/// <summary>
		/// Method does not support the instances being compared.
		/// </summary>
		TypesNotSupported,

		/// <summary>
		/// Method is appropriate for the data types, but doesn't support the specified tolerance.
		/// </summary>
		ToleranceNotSupported,

		/// <summary>
		/// Method is appropriate and the items are considered equal.
		/// </summary>
		ComparedEqual,

		/// <summary>
		/// Method is appropriate and the items are considered different.
		/// </summary>
		ComparedNotEqual,

		/// <summary>
		/// Method is appropriate but the class has cyclic references.
		/// </summary>
		ComparisonPending
	}
}
```
</details>

mithileshz:
Fixes #4735

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
@@ -74,7 +74,7 @@ namespace NUnit.Framework.Constraints.Comparers
 
                     if (MemoryExtensions.SequenceEqual<byte>(bufferExpected.AsSpan(0, readExpected), bufferActual.AsSpan(0, readActual)))
                     {
-                        readByte += readExpected;
+                        readByte += readActual;
                         continue;
                     }
 
--- a/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
@@ -39,8 +39,13 @@ namespace NUnit.Framework.Constraints.Comparers
                     return EqualMethodResult.ComparedEqual;
             }
 
+#if NETFRAMEWORK
             byte[] bufferExpected = new byte[BUFFER_SIZE];
             byte[] bufferActual = new byte[BUFFER_SIZE];
+#else
+            Span<byte> bufferExpected = stackalloc byte[BUFFER_SIZE];
+            Span<byte> bufferActual = stackalloc byte[BUFFER_SIZE];
+#endif
 
             BinaryReader binaryReaderExpected = new BinaryReader(xStream);
             BinaryReader binaryReaderActual = new BinaryReader(yStream);
@@ -65,8 +70,13 @@ namespace NUnit.Framework.Constraints.Comparers
 
                 while (readExpected > 0 && readActual > 0)
                 {
+#if NETFRAMEWORK
                     readExpected = binaryReaderExpected.Read(bufferExpected, 0, BUFFER_SIZE);
                     readActual = binaryReaderActual.Read(bufferActual, 0, BUFFER_SIZE);
+#else
+                    readExpected = binaryReaderExpected.Read(bufferExpected);
+                    readActual = binaryReaderActual.Read(bufferActual);
+#endif
 
                     if (MemoryExtensions.SequenceEqual<byte>(bufferExpected, bufferActual))
                     {
--- a/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
@@ -2,6 +2,9 @@
 
 using System;
 using System.IO;
+#if !NETFRAMEWORK
+using System.Buffers;
+#endif
 
 namespace NUnit.Framework.Constraints.Comparers
 {
@@ -43,8 +46,10 @@ namespace NUnit.Framework.Constraints.Comparers
             byte[] bufferExpected = new byte[BUFFER_SIZE];
             byte[] bufferActual = new byte[BUFFER_SIZE];
 #else
-            Span<byte> bufferExpected = stackalloc byte[BUFFER_SIZE];
-            Span<byte> bufferActual = stackalloc byte[BUFFER_SIZE];
+            ArrayPool<byte> arrayPool = ArrayPool<byte>.Shared;
+
+            byte[] bufferExpected = arrayPool.Rent(BUFFER_SIZE);
+            byte[] bufferActual = arrayPool.Rent(BUFFER_SIZE);
 #endif
 
             BinaryReader binaryReaderExpected = new BinaryReader(xStream);
@@ -70,13 +75,8 @@ namespace NUnit.Framework.Constraints.Comparers
 
                 while (readExpected > 0 && readActual > 0)
                 {
-#if NETFRAMEWORK
                     readExpected = binaryReaderExpected.Read(bufferExpected, 0, BUFFER_SIZE);
                     readActual = binaryReaderActual.Read(bufferActual, 0, BUFFER_SIZE);
-#else
-                    readExpected = binaryReaderExpected.Read(bufferExpected);
-                    readActual = binaryReaderActual.Read(bufferActual);
-#endif
 
                     if (MemoryExtensions.SequenceEqual<byte>(bufferExpected, bufferActual))
                     {
@@ -110,6 +110,11 @@ namespace NUnit.Framework.Constraints.Comparers
                 {
                     yStream.Position = actualPosition;
                 }
+
+#if !NETFRAMEWORK
+                arrayPool.Return(bufferExpected);
+                arrayPool.Return(bufferActual);
+#endif
             }
 
             return EqualMethodResult.ComparedEqual;
--- a/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
@@ -2,9 +2,7 @@
 
 using System;
 using System.IO;
-#if !NETFRAMEWORK
 using System.Buffers;
-#endif
 
 namespace NUnit.Framework.Constraints.Comparers
 {
@@ -42,24 +40,20 @@ namespace NUnit.Framework.Constraints.Comparers
                     return EqualMethodResult.ComparedEqual;
             }
 
-#if NETFRAMEWORK
-            byte[] bufferExpected = new byte[BUFFER_SIZE];
-            byte[] bufferActual = new byte[BUFFER_SIZE];
-#else
-            ArrayPool<byte> arrayPool = ArrayPool<byte>.Shared;
-
-            byte[] bufferExpected = arrayPool.Rent(BUFFER_SIZE);
-            byte[] bufferActual = arrayPool.Rent(BUFFER_SIZE);
-#endif
-
             BinaryReader binaryReaderExpected = new BinaryReader(xStream);
             BinaryReader binaryReaderActual = new BinaryReader(yStream);
 
             long expectedPosition = bothSeekable ? xStream.Position : default;
             long actualPosition = bothSeekable ? yStream.Position : default;
 
+            byte[]? bufferExpected = null;
+            byte[]? bufferActual = null;
+
             try
             {
+                bufferExpected = ArrayPool<byte>.Shared.Rent(BUFFER_SIZE);
+                bufferActual = ArrayPool<byte>.Shared.Rent(BUFFER_SIZE);
+
                 if (xStream.CanSeek)
                 {
                     binaryReaderExpected.BaseStream.Seek(0, SeekOrigin.Begin);
@@ -78,9 +72,9 @@ namespace NUnit.Framework.Constraints.Comparers
                     readExpected = binaryReaderExpected.Read(bufferExpected, 0, BUFFER_SIZE);
                     readActual = binaryReaderActual.Read(bufferActual, 0, BUFFER_SIZE);
 
-                    if (MemoryExtensions.SequenceEqual<byte>(bufferExpected, bufferActual))
+                    if (MemoryExtensions.SequenceEqual<byte>(bufferExpected.AsSpan(0, readExpected), bufferActual.AsSpan(0, readActual)))
                     {
-                        readByte += readActual;
+                        readByte += readExpected;
                         continue;
                     }
 
@@ -111,10 +105,11 @@ namespace NUnit.Framework.Constraints.Comparers
                     yStream.Position = actualPosition;
                 }
 
-#if !NETFRAMEWORK
-                arrayPool.Return(bufferExpected);
-                arrayPool.Return(bufferActual);
-#endif
+                if (bufferExpected is not null)
+                    ArrayPool<byte>.Shared.Return(bufferExpected);
+
+                if (bufferActual is not null)
+                    ArrayPool<byte>.Shared.Return(bufferActual);
             }
 
             return EqualMethodResult.ComparedEqual;
--- a/src/NUnitFramework/benchmarks/nunit.framework.benchmarks/ByteAllocationBenchmark.cs
+++ b/src/NUnitFramework/benchmarks/nunit.framework.benchmarks/ByteAllocationBenchmark.cs
@@ -206,10 +206,8 @@ namespace NUnit.Framework
             if (bothSeekable && xStream.Length != yStream.Length)
                 return EqualMethodResult.ComparedNotEqual;
 
-            var arrayPool = ArrayPool<byte>.Shared;
-
-            byte[] bufferExpected = arrayPool.Rent(BUFFER_SIZE);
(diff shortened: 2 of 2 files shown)


src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
```diff
@@ -39,8 +39,13 @@ public static EqualMethodResult Equal(object x, object y, ref Tolerance toleranc
                     return EqualMethodResult.ComparedEqual;
             }
 
+#if NETFRAMEWORK
             byte[] bufferExpected = new byte[BUFFER_SIZE];
             byte[] bufferActual = new byte[BUFFER_SIZE];
+#else
+            Span<byte> bufferExpected = stackalloc byte[BUFFER_SIZE];
+            Span<byte> bufferActual = stackalloc byte[BUFFER_SIZE];
```
OsirisTerje, line 47: The BUFFER_SIZE is 4096.  That is a bit more than what is recommended for stackalloc, which says up to 1KB.   Should this be a concern here?

OsirisTerje:
The StreamsComparer does not have direct unit tests.  It could be an idea to add this since it is being directly modified here.

OsirisTerje (changes requested):

src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
stevenaw: I agree, even 1KB is on the higher side as I've seen 256 bytes often as a limit to stackalloc.

One other option here could be `ArrayPool`.

stevenaw:
Thanks for adding me here @OsirisTerje 

@mithileshz Welcome back and thank you as well for noticing the room for improvement here 🙂
I'm away on vacation without a laptop but can participate better upon my return. For now I've left a suggestion above in response @OsirisTerje 's comment.

mithileshz:
> The StreamsComparer does not have direct unit tests. It could be an idea to add this since it is being directly modified here.

There are tests inside EqualConstraintTests inside the region StreamEquality. As this change doesn't change the behaviour of the StreamComparer, I didn't add any extra tests as those would have been covered by the tests in class?

src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
mithileshz: I'll look into `ArrayPool` - haven't used it before so will investigate and come back later today.

mithileshz:
@stevenaw @OsirisTerje 
ArrayPools are just as performant and don't have the issues that stack allocation would have. Here is the benchmark results. I have also updated the PR to reflect this change :)

```

```

OsirisTerje:
4 tests failing.

src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
```diff
@@ -100,6 +110,11 @@ public static EqualMethodResult Equal(object x, object y, ref Tolerance toleranc
                 {
                     yStream.Position = actualPosition;
                 }
+
+#if !NETFRAMEWORK
+                arrayPool.Return(bufferExpected);
+                arrayPool.Return(bufferActual);
```
stevenaw, line 116: @mithileshz 
Based on the info in the test failures, I am wondering if data from the pooled arrays is persisting across calls. Something you can try is to pass `true` as a second argument here when returning the arrays to the pool to force the content to be cleared between usages :)

src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
```diff
@@ -39,8 +42,15 @@ public static EqualMethodResult Equal(object x, object y, ref Tolerance toleranc
                     return EqualMethodResult.ComparedEqual;
             }
 
+#if NETFRAMEWORK
```
stevenaw, line 45: request:

Can we remove the need for `#if NETFRAMEWORK` here? `ArrayPool` is available for all of NUnit's build targets so there's no need to make this .NET Core-specific :)

src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
```diff
@@ -39,8 +42,15 @@ public static EqualMethodResult Equal(object x, object y, ref Tolerance toleranc
...
             byte[] bufferExpected = new byte[BUFFER_SIZE];
             byte[] bufferActual = new byte[BUFFER_SIZE];
+#else
+            ArrayPool<byte> arrayPool = ArrayPool<byte>.Shared;
+
+            byte[] bufferExpected = arrayPool.Rent(BUFFER_SIZE);
+            byte[] bufferActual = arrayPool.Rent(BUFFER_SIZE);
```
stevenaw, line 52: request:

Can we move this into the try/finally to help ensure we always return the array backto the pool even in the event an exception is thrown? This can help prevent memory leaks. It might look something like this in practice:

```csharp
byte[] bufferExpected = null;

try {
  bufferExpected = ArrayPool<byte>.Shared.Rent(BUFFER_SIZE);
  
  // other things
} finally {
  if (bufferExpected is not null) {
    ArrayPool<byte>.Shared.Return(bufferExpected, clearArray: true);
  }
}

**Review by stevenaw — CHANGES_REQUESTED** [review]
Great work @mithileshz , results are looking promising so far!
I've proactively left a few helpful pointers since you mentioned you were a little new to using `ArrayPool`.

File: src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs (hunk 3 above)
Comment (mithileshz) [reply src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs:45]: The reason I added it is because it gives an error on .net framework for [IDE0005 ](https://learn.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/style-rules/ide0005?pivots=lang-csharp-vb) because we are not using it

File: src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs (hunk 2 above)
Comment (mithileshz) [reply src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs:116]: Ok I think I have finally understood what happens.

Basically there is no gurantee that `ArrayPool.Shared.Rent(size)` gives you an array with values `0`. It is not a replacement for `new byte[size]` which means that when we use the binaryReader to read into these arrays (line 68ish), it will override the first _n_ items in the byte array but not the rest. This causes the `SequenceEqual` method to return false as it will read the whole byte array and compare the two. This causes the unit test to fail as the data is not the same.

Comment (stevenaw) [reply src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs:116]: Good sleuthing on the cause. I'm hesitant to suggest we move to a smaller buffer size here as 4096 bytes also aligns with many of the internal .NET and Windows buffer sizes, which carries it's own performance benefits.

I think `ArrayPool` could still be a viable option, we may just need to manually create spans over only the consumed part of the buffer and then pass that into `MemoryExtensions.SequenceEqual` instead. Something like the below works on my machine.

```diff
+if (MemoryExtensions.SequenceEqual<byte>(bufferExpected, bufferActual))
-if (MemoryExtensions.SequenceEqual(bufferExpected.AsSpan(0, readExpected), bufferActual.AsSpan(0, readActual)))
```

src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
stevenaw: Why not use it on .NET framework too? It should be available.

src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
mithileshz: Ah yes I didn't think about this solution!
It works with the above.
I've updated the PR and here's the benchmark:

```
| Method       | Size  | Mean     | Error     | StdDev    | Ratio | RatioSD | Gen0   | Gen1   | Allocated | Alloc Ratio |
|------------- |------ |---------:|----------:|----------:|------:|--------:|-------:|-------:|----------:|------------:|
| NewByteArray | 32768 | 1.829 μs | 0.0315 μs | 0.0768 μs |  1.00 |    0.00 | 0.5112 | 0.0114 |    8560 B |        1.00 |
| StackAlloc   | 32768 | 1.370 μs | 0.0266 μs | 0.0249 μs |  0.76 |    0.03 | 0.0191 | 0.0019 |     320 B |        0.04 |
| ArrayPool    | 32768 | 1.503 μs | 0.0203 μs | 0.0190 μs |  0.83 |    0.02 | 0.0191 | 0.0019 |     320 B |        0.04 |
```

src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
mithileshz: Done :)

src/NUnitFramework/framework/Constraints/Comparers/StreamsComparer.cs
mithileshz: Aah I was not sure if this was a .NET Core only feature! I've added it across the frameworks now

OsirisTerje (approved):
lgtm

stevenaw (approved):
Fantastic, looks like your change speeds things up _and_ lowers allocations.
LGTM, thanks for your contribution @mithileshz

Issue #4738, opened 2024-06-26
QoL suggestion: fail fast in Assert.Multiple under debugger


MaceWindu:
While `Assert.Multiple` is nice for failure reporting on test runs, it is quite bad when you try to debug test compared to regular assert.

- with simple assert you get exception on assertion failure and could debug it
- with `Assert.Multiple` you get exception at the end where you probably lost all related test context already and need to put manual breakpoints inside of delegate to be able to debug first failure (or comment `Assert.Multiple` wrap)

I propose to throw assert exception on first failure if process is under debug (`Debugger.IsAttached == true`) with some kind of switch for backward compatibility. E.g. for those who use tests with debbugger for any reason or just enjoy having troubles during debugging.

OsirisTerje:
Nice idea!  Could be enabled using a setting.  Would you like to do a PR for this ?

MaceWindu:
Sure https://github.com/nunit/nunit/pull/4739

Need some help with:

- what else is missing?
- default value true or false? I assume it should be false for backward compat.

OsirisTerje:
> default value true or false? I assume it should be false for backward compat.
Yes :-)

Can't think of anything else missing right now.

OsirisTerje:
Need to raise a separate documentation issue for this.

RenderMichael:
I am also interested in this.

Thanks to first-change exceptions, the debugger stops whenever anything throws, even if it is caught (unless Just My Code is enabled).

Therefore you could have something like

```csharp
if (DummyThrowForDebugger)
{
    try
    {
        throw new Exception();
    }
    catch (Exception)
    {
    }
}
```

And the debugger will still break here.

Alternatively, you could just call `System.Diagnostics.Debugger.Break();` directly.

This will enable debugger stopping without breaking `Assert.Multiple` functionality.

OsirisTerje:
@RenderMichael But does the PR, which is now merged,solve this for you?

RenderMichael:
It solves it in the same way commenting out `Assert.Multiple` would. We’d lose the second, third, etc. error message, and in order to debug through those, the first assertion needs to be commented out or resolved.

MaceWindu:
Such behavior also could be useful. @OsirisTerje any objections to convert new boolean option to enumeration like that:

```cs
enum MultipleAssertBehaviorUnderDebugger
{
    Default,
    ThrowOnFirstFailure,
    ThrowOnEachFailure
}
```

[...]

PR #4739, opened 2024-06-27
Disable Assert.Multiple under debugger

MaceWindu:
Initial implementation for https://github.com/nunit/nunit/issues/4738

### DIFF
--- a/src/NUnitFramework/framework/Assert.cs
+++ b/src/NUnitFramework/framework/Assert.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Threading.Tasks;
 using NUnit.Framework.Interfaces;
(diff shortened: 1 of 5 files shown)


src/NUnitFramework/framework/Assert.cs
```diff
@@ -318,8 +319,8 @@ private static void ReportFailure(string message)
             result.RecordAssertion(AssertionStatus.Failed, message, GetStackTrace());
             result.RecordTestCompletion();
 
-            // If we are outside any multiple assert block, then throw
-            if (TestExecutionContext.CurrentContext.MultipleAssertLevel == 0)
+            // If we are outside any multiple assert block or multiple asserts disabled, then throw
+            if (TestExecutionContext.CurrentContext.MultipleAssertLevel == 0 || (TestExecutionContext.CurrentContext.DisableMultipleAssertsUnderDebugger && Debugger.IsAttached))
```
MaceWindu, line 323: Check debugger status at exception throw location as user could connect/disconnect it during run

OsirisTerje:
The setting also needs to be implemented in the runsettings for the adapter. The adapter is responsible for setting it for the engine/framework.

OsirisTerje (approved):
Looks good to me.

This new setting needs to be documented.

PR #4743, opened 2024-07-03
WIP: HookExtension to support high level tests

TSAVogt:
see issue: https://github.com/nunit/nunit/issues/4744

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/SetUpTearDownItem.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/SetUpTearDownItem.cs
@@ -95,7 +95,6 @@ namespace NUnit.Framework.Internal.Commands
             {
                 context.HookExtension.AfterSetUp(setUpMethod.Name);
             }
-
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/NUnitFramework/framework/Internal/Commands/SetUpTearDownItem.cs
```diff
@@ -50,7 +50,51 @@ public void RunSetUp(TestExecutionContext context)
...
+                    RunSetUpOrTearDownMethod(context, setUpMethod);
+                }
+                finally
+                {
+                    CallAfterSetUpsHook(context, setUpMethod);
+                }
+            }
+        }
+
+        private void CallBeforeSetUpsHook(TestExecutionContext context, IMethodInfo setUpMethod)
```
OsirisTerje, line 66: These methods look identical, except for the two methods to be called.  
Suggest you send those two methods in as parameters to a single method,  to avoid having to check identical code later.  (DRY principle)

TSAVogt: This pull request I have closed because I missed to design the interface really minimal...
See [this](https://github.com/nunit/nunit/pull/4745) pull request that also does not have the additional methods but only the IHookExtension interface.
Sorry for confusion. Did not found a way to delete the pull request...

Issue #4744, opened 2024-07-03
HookExtension to support high level tests


TSAVogt:
To support tests of higher test level with e.g. advanced test protocols or test environment preparation we need a reliable way to get called synchronously before and after any test relevant method (all kind of setups and teardowns, test actions and test methods).
Therefore, we would like to introduce this HookExtension interface like this:

``` csharp
public interface IHookExtension
{
    // TODO: clarfy if parameter methodName is sufficient
    // or if the (additional?) TestExecutionContext parameter is better 
    void BeforeOneTimeSetUp(string methodName);
    void AfterOneTimeSetUp(string methodName);

    void BeforeSetUp(string methodName);
    void AfterSetUp(string methodName);

    void BeforeTest(string methodName);
    void AfterTest(string methodName);

    void BeforeTearDown(string methodName);
    void AfterTearDown(string methodName);

    void BeforeOneTimeTearDown(string methodName);
    void AfterOneTimeTearDown(string methodName);
}
```
A custom extension can be activated by e.g. an IApplyToContext assembly attribute:

``` csharp
public class TestEnvironment : NUnitAttribute, IApplyToContext
{
    public void ApplyToContext(TestExecutionContext context)
    {
        context.HookExtension = new TestEnvironmentHookExtension();
    }
}
```
For more implementation details see the draft pull request: https://github.com/nunit/nunit/pull/4743

This solution would have the advantage that relevant changes are only necessary on NUnit framework, are minimal invasive and enable users to handle or orchestrate all different types of extensions for higher test levels.

What do you think about this solution?

TSAVogt:
need some corrections first

TSAVogt:
Just learned that I should provide some usage in the nunit.issues repo. Will provide it... Stay tuned...

TSAVogt:
Show case is ready here: https://github.com/nunit/nunit.issues/tree/main/Issue4744

Looking forward your feedback!

RenderMichael:
Is it possible to add hooks on each assertion? I am imagining something like middleware that's executed on each `Assert`. Ideally this would not regress the performance of the non-customized case.

TSAVogt:
@RenderMichael: when we have finished the discussion on how to implement this featur, I will check if this mechanism can be also extended for your use case.

TSAVogt:
As a basis for how to implement this feature, I have also added a draft PR to show what the implementation looks like if we use C# events: https://github.com/nunit/nunit/pull/4769
[...]

PR #4745, opened 2024-07-04
WIP: Introduce HookExtension to support high level tests

TSAVogt:
see issue: https://github.com/nunit/nunit/issues/4744


TSAVogt:
Will follow up with event approach.

PR #4749, opened 2024-07-09
Rework assert.multiple behavior under debugger to throw on each failure

MaceWindu:
See discussion in #4738.

Instead of throwing on first failure, we will generate catched exception on each failure to allow user to debug it and still have multiple assert exception in the end

### DIFF
--- a/src/NUnitFramework/framework/Api/FrameworkPackageSettings.cs
+++ b/src/NUnitFramework/framework/Api/FrameworkPackageSettings.cs
@@ -88,9 +88,9 @@ namespace NUnit
         public const string StopOnError = "StopOnError";
 
         /// <summary>
-        /// If true, asserts in multiple asserts block will throw immediately.
+        /// If true, asserts in multiple asserts block will throw first-chance exception on failure.
         /// </summary>
-        public const string DisableMultipleAssertsUnderDebugger = "DisableMultipleAssertsUnderDebugger";
+        public const string ThrowOnEachFailureUnderDebugger = "ThrowOnEachFailureUnderDebugger";
 
         /// <summary>
         /// If true, use of the event queue is suppressed and test events are synchronous.
--- a/src/NUnitFramework/framework/Assert.cs
+++ b/src/NUnitFramework/framework/Assert.cs
@@ -321,8 +321,18 @@ namespace NUnit.Framework
             result.RecordTestCompletion();
 
             // If we are outside any multiple assert block or multiple asserts disabled, then throw
-            if (TestExecutionContext.CurrentContext.MultipleAssertLevel == 0 || (TestExecutionContext.CurrentContext.DisableMultipleAssertsUnderDebugger && Debugger.IsAttached))
-                throw new AssertionException(result.Message);
+            var debugMultiple = TestExecutionContext.CurrentContext.ThrowOnEachFailureUnderDebugger && Debugger.IsAttached;
+            if (TestExecutionContext.CurrentContext.MultipleAssertLevel == 0 || debugMultiple)
+            {
+                try
+                {
+                    throw new AssertionException(result.Message);
+                }
+                catch (AssertionException) when (debugMultiple && TestExecutionContext.CurrentContext.MultipleAssertLevel > 0)
+                {
+                    // we catch exception for multiple assert block to not change observed behavior but still allow user to break into debugger
+                }
+            }
         }
 
         private static void IssueWarning(string message)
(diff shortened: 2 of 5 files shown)


OsirisTerje:
@manfred-brands @stevenaw  Can you have a look?

src/NUnitFramework/framework/Api/FrameworkPackageSettings.cs
```diff
@@ -88,9 +88,9 @@ public static class FrameworkPackageSettings
         public const string StopOnError = "StopOnError";
 
         /// <summary>
-        /// If true, asserts in multiple asserts block will throw immediately.
+        /// If true, asserts in multiple asserts block will throw first-chance exception on failure.
         /// </summary>
-        public const string DisableMultipleAssertsUnderDebugger = "DisableMultipleAssertsUnderDebugger";
+        public const string ThrowOnEachFailureUnderDebugger = "ThrowOnEachFailureUnderDebugger";
```
stevenaw, line 93: I think I like this updated name more, thanks!

src/NUnitFramework/framework/Assert.cs
```diff
@@ -320,8 +320,18 @@ private static void ReportFailure(string message)
...
-            if (TestExecutionContext.CurrentContext.MultipleAssertLevel == 0 || (TestExecutionContext.CurrentContext.DisableMultipleAssertsUnderDebugger && Debugger.IsAttached))
-                throw new AssertionException(result.Message);
+            var debugMultiple = TestExecutionContext.CurrentContext.ThrowOnEachFailureUnderDebugger && Debugger.IsAttached;
+            if (TestExecutionContext.CurrentContext.MultipleAssertLevel == 0 || debugMultiple)
+            {
+                try
+                {
+                    throw new AssertionException(result.Message);
+                }
+                catch (AssertionException) when (debugMultiple && TestExecutionContext.CurrentContext.MultipleAssertLevel > 0)
```
stevenaw, line 330: @manfred-brands This looks functionally correct to me, but barring time to run this myself I may need to defer to your recent foray on multiple assertion exceptions (if you have time yourself, of course) 

@MaceWindu out of curiosity, why not have used `Debugger.Break()`?

manfred-brands (commented):
github doesn't allow me to comment on non-changes code.

@MaceWindu Should the `ThrowOnEachFailureUnderDebugger` property not be copied in the copy constructor?

Every child in a `CompositeWorkItem` aka a Fixture gets a new `TestExecutionContext`.
But the child will not have `ThrowOnEachFailureUnderDebugger` set, ever.

The only place where it is currently set is in the `NUnitTestAssemblyRunner`.

@OsirisTerje Not all the other properties are copied either.
`TestCaseTimeout` and `StopOnError` are, but `Current(UI)Culture` are not. Should they?

src/NUnitFramework/framework/Assert.cs
MaceWindu: Now I think it would be better to use Debugger.Break here

Pros:
- simplier/less code
- no need for user to setup debugger to break on handled exceptions

Cons:
- `result.Message` not explicitly available, user need to inspect it
- as @manfred-brands said - it will break always even if user don't want it. But in that case he shouldn't have enabled option I think
- need to rename setting again 😄 

> Not all the other properties are copied either.

That's why I didn't do it - I've checked how other settings handled

stevenaw: Thanks for the explanation @MaceWindu 😃 
It does make sense to me to prioritize user friendliness on a setting like this, which it sounds like was how you approached it too

OsirisTerje:
@MaceWindu  See comments from @manfred-brands - any upcoming action on those?


> @manfred-brands Not all the other properties are copied either.
TestCaseTimeout and StopOnError are, but Current(UI)Culture are not. Should they?

Good question, meaning, IDK.  I don't think they are being asked for, on the other side, it should not do any harm if they do. 
I also noticed that over time we have added properties to TestContext from underlying classes, and they have not been documented. (Did that now).,  So perhaps we should do a separate cleanup and documentation issue on that?

MaceWindu:
@OsirisTerje , I'will add copy to those properties. Just tested and can confirm options are not propagated to test context

MaceWindu:
BTW, CurrentCulture and CurrentUICulture not copied, as they are calculated properties over _sandboxedThreadState, which is copies already

OsirisTerje (approved):
Ok, seems all is done here.

stevenaw:
Thanks @OsirisTerje 
Had just been looking to approve this myself 👍

PR #4755, opened 2024-07-11
Generic test case attribute

manfred-brands:
Add support for:

* `[TestCase<double>(2)]`
* `new TestCaseData<double>(2)`

Both support up to 5 generic arguments.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs
@@ -476,4 +476,91 @@ namespace NUnit.Framework
 
         #endregion
     }
+
+#if NET6_0_OR_GREATER // Although this compiles for .NET Framework, it fails at runtime with a NotSupportedException : Generic types are not valid.
+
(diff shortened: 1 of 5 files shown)


OsirisTerje:
Seems to be a small timing issue on the Linmux build.  The time limit is set to 100 mS, actual is 125 mS.  So Linux is slower.....

manfred-brands:
> Seems to be a small timing issue on the Linux build. The time limit is set to 100 mS, actual is 125 mS. So Linux is slower.....
> 
> But the 100mS is a hardcoded limit, is this meant to be really faster, so this is just a high limit, or are there something behind this ?

Nothing in this PR affected this.
Maybe there is a difference with the runner we get now.
slower or with less memory. i think we had to disable the time test for MacOs.

OsirisTerje:
> i think we had to disable the time test for MacOs.

Yes, we can do the same here, or increase the timeout.  Whatever.

manfred-brands:
Triggered a re-build, it is green now.

OsirisTerje (approved):
lgtm

PR #4756, opened 2024-07-12
OSPlatforms attribute no longer gets translated into PlatformAttribute

### LINKED ISSUE #4577: Remove some version hard-coding in the OSPlatformTranslator
In https://github.com/nunit/nunit/pull/4574 and https://github.com/nunit/nunit/pull/4569 we added some hard-coded version numbers in the OSPlatformTranslator to fix a few things in NUnit 4.0.1.
It should be alright in the short-term as new major versions of Windows are relatively infrequent, however we should see if we can remove some of the hard-codings to minimize the chance that a new version of Windows will require a new version of the framework to work with it.

We should also ensure that the mapping code can account for any changes in the SDK for how a TFM such as `{version}-windows` may map to a compiler-generated `SupportOSPlatform` attribute that may come out of this issue: https://github.com/dotnet/sdk/issues/37220

One option could be to see if we can make use of `[Platform("Win")]` .

manfred-brands:
Fixes #4577 

OSPlatformAttribute no longer translated to old PlatformAttribute

Introduced a new NetPlatformAttribute (only so it can implement IApplyToTest)
OSPlatformAttribute is converted to NetPlatformAttribute
Actual platform name is passed verbatim to `System.OperatingSystem.IsOSPlatform`


src/NUnitFramework/framework/Attributes/NetPlatformAttribute.cs
```diff
@@ -0,0 +1,78 @@
...
+using NUnit.Framework.Internal;
+
+namespace NUnit.Framework
+{
+    /// <summary>
+    /// Marks an assembly, test fixture or test method as applying to a specific platform.
+    /// </summary>
+    /// <remarks>
+    /// This class is a replacement for the <see cref="PlatformAttribute"/> class,
+    /// the platform names are based on the values in the TargetFramrwork.
```
OsirisTerje, line 14:
```suggestion
    /// the platform names are based on the values in the TargetFramework.
```

src/NUnitFramework/framework/Internal/NetPlatformHelper.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            if (platform.Contains(','))
+                return IsPlatformSupported(platform.Split(','));
+
+            string platformName = platform.Trim();
+            ParseOSAndVersion(platformName, out string os, out Version? version);
+
+            if (version is not null && os.Equals("Windows", StringComparison.OrdinalIgnoreCase))
+            {
+                // Atual versions don't match public known names
```
OsirisTerje, line 43:
```suggestion
                // Actual versions don't match public known names
```

OsirisTerje (approved):

PR #4758, opened 2024-07-13
Introduced Assert.EnterMultipleScope

### LINKED ISSUE #4587: Feature request: Assert.Multiple() could return an IDisposable, avoiding passing an Action around.
Would it be possible to be able to write code like:
```
// Act
var result = GetResult();

// Assert
using (Assert.Multiple())
{
    Assert.That(result.A, Is.EqualTo("A"));
    Assert.That(result.B, Is.EqualTo("B"));
}
```

This would hopefully avoid capturing unnecessary context into the Action (because it wouldn't exist).  Async code becomes slightly simpler (no need to specify `async` again) also.

It might also make things like #4242 easier to get right.

Considerations:
- As written, this would not conflict, but we would certainly want to drive users towards this as a preferred option over the existing `Assert.Multiple(() => { ... });` option.  So we could consider using a different name to avoid ambiguity.

manfred-brands:
Fixes #4587

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/TestMethodCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TestMethodCommand.cs
@@ -47,7 +47,10 @@ namespace NUnit.Framework.Internal.Commands
             if (_testMethod.HasExpectedResult)
                 Assert.That(result, Is.EqualTo(_testMethod.ExpectedResult));
 
-            context.CurrentResult.SetResult(ResultState.Success);
+            if (context.MultipleAssertLevel > 0)
+                context.CurrentResult.SetResult(ResultState.Error, $"Test completed with {context.MultipleAssertLevel} active assertion scopes.");
+            else
+                context.CurrentResult.SetResult(ResultState.Success);
 
             if (context.CurrentResult.AssertionResults.Count > 0)
                 context.CurrentResult.RecordTestCompletion();
--- a/src/NUnitFramework/testdata/AssertMultipleData.cs
+++ b/src/NUnitFramework/testdata/AssertMultipleData.cs
@@ -525,6 +525,15 @@ namespace NUnit.TestData.AssertMultipleData
             outerScope.Dispose();
             innerScope.Dispose();
         }
+
+        [Test]
+        public void ScopeReleasedTwice()
+        {
+            using IDisposable scope = Assert.EnterMultipleScope();
+            Assert.That(Complex.RealPart, Is.EqualTo(5.2));
+            Assert.That(Complex.ImaginaryPart, Is.EqualTo(3.9));
+            scope.Dispose();
+        }
     }
 
     internal class ComplexNumber
--- a/src/NUnitFramework/testdata/AssertMultipleData.cs
+++ b/src/NUnitFramework/testdata/AssertMultipleData.cs
@@ -492,6 +492,39 @@ namespace NUnit.TestData.AssertMultipleData
                 }
             }
         }
+
+        [Test]
+        public void NonReleasedScope()
+        {
+            Assert.EnterMultipleScope();
+            Assert.That(Complex.RealPart, Is.EqualTo(5.2));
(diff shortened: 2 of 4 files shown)


manfred-brands:
This is now ready for review.

src/NUnitFramework/framework/Internal/Commands/TestMethodCommand.cs
```diff
@@ -47,7 +47,10 @@ public override TestResult Execute(TestExecutionContext context)
             if (_testMethod.HasExpectedResult)
                 Assert.That(result, Is.EqualTo(_testMethod.ExpectedResult));
 
-            context.CurrentResult.SetResult(ResultState.Success);
+            if (context.MultipleAssertLevel > 0)
```
stevenaw, line 50: question: Should we dispose of any open scopes here after reporting the error? I'm not sure if there could be any stability issues with leaving them around given they themselves contain a reference to a particular `TestExecutionContext`

stevenaw (commented):
Thanks @manfred-brands this looks great.
I have two questions about edge case behaviours on the new scope.

src/NUnitFramework/testdata/AssertMultipleData.cs
```diff
@@ -429,6 +472,59 @@ public void TwoNestedBlocks_TwoAssertsFail_Async()
...
+            Assert.EnterMultipleScope();
+            Assert.That(2 + 2, Is.EqualTo(4));
+
+            Assert.EnterMultipleScope();
+            Assert.That(Complex.RealPart, Is.EqualTo(5.2));
+            Assert.That(Complex.ImaginaryPart, Is.EqualTo(3.9));
+        }
+
+        [Test]
+        public void ScopeReleasedOutOfOrder()
```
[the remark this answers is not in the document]
manfred-brands: @stevenaw Good point!

I went for not throwing an error as multiple calls to `Dispose` should be safe as per the interface specification [remarks](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose?view=net-8.0#remarks).

> If an object's [Dispose](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose?view=net-8.0) method is called more than once, the object must ignore all calls after the first one. The object must not throw an exception if its [Dispose](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose?view=net-8.0) method is called multiple times.

The [CA2202](https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2019/code-quality/ca2202?view=vs-2019-archive&viewFallbackFrom=vs-2022) is deprecated and no longer in the latest analyzer nuget package.

src/NUnitFramework/framework/Internal/Commands/TestMethodCommand.cs
manfred-brands: We cannot as we don't have a reference to them. I don't think it is worth it to add them as references to the _context_.
Every test gets it [own new context](https://github.com/nunit/nunit/blob/main/src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs#L272) and the old scope and context will be garbage collected.

stevenaw: Ok sounds good. My thinking had indeed been probing if it made sense to track these on the context as a `Stack<AssertionScope>` instead of the `MultipleAssertLevel` counter, but as you say we can also fulfill the behaviour without doing those broader changes.

src/NUnitFramework/testdata/AssertMultipleData.cs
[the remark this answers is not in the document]
stevenaw: Thanks! I had somehow glossed over the part of the documentation where it said the rule was deprecated.

stevenaw (approved):
This looks good to me, thanks @manfred-brands .
Not sure if you were in a rush to merge or wanted to see if others had thoughts

OsirisTerje:
Lgtm, so I'll just merge it.

PR #4762, opened 2024-07-18
Testcontextsimplify

OsirisTerje:
Simplifies test context Out methods code, by changing from using explicit code per type to generic methods.


src/NUnitFramework/framework/TestContext.cs
```diff
@@ -240,85 +173,12 @@ public static void Write(string format, params object?[] args)
             Out.Write(format, args);
         }
 
-        /// <summary>Write a line terminator to the current result</summary>
-        public static void WriteLine()
```
stevenaw, line 244: @OsirisTerje I'm not 100% sure as I'm on my phone at the moment, but wouldn't this one still be needed?

NUnit.sln.DotSettings
```diff
@@ -42,8 +42,22 @@
...
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/UserRules/=03c58a20_002Df24c_002D4992_002Da01b_002Dd4dd88b40443/@EntryIndexedValue">&lt;Policy&gt;&lt;Descriptor Staticness="Static" AccessRightKinds="Private, Internal" Description="Non-visible static readonly and constant fields"&gt;&lt;ElementKinds&gt;&lt;Kind Name="READONLY_FIELD" /&gt;&lt;Kind Name="CONSTANT_FIELD" /&gt;&lt;/ElementKinds&gt;&lt;/Descriptor&gt;&lt;Policy Inspect="False" Prefix="" Suffix="" Style="AaBb" /&gt;&lt;/Policy&gt;</s:String>
+	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/UserRules/=15b5b1f1_002D457c_002D4ca6_002Db278_002D5615aedc07d3/@EntryIndexedValue">&lt;Policy&gt;&lt;Descriptor Staticness="Static" AccessRightKinds="Private" Description="Static readonly fields (private)"&gt;&lt;ElementKinds&gt;&lt;Kind Name="READONLY_FIELD" /&gt;&lt;/ElementKinds&gt;&lt;/Descriptor&gt;&lt;Policy Inspect="False" Prefix="" Suffix="" Style="AaBb" /&gt;&lt;/Policy&gt;</s:String>
```
manfred-brands, line 45: ReSharper should be able to use our .editorconfig which has the naming rules in it: https://www.jetbrains.com/help/resharper/Using_EditorConfig.html
So there should be no need for naming conventions in this file.

src/NUnitFramework/framework/TestContext.cs
```diff
@@ -142,79 +142,12 @@ public string TestDirectory
...
-        public static void Write(ulong value)
-        {
-            Out.Write(value);
-        }
+        /// <summary>
+        /// Write the string representation of an object to the current result
+        /// </summary>
+        /// <param name="val"></param>
+        /// <typeparam name="T"></typeparam>
+        public static void Write<T>(T val) => Out.Write(val);
```
manfred-brands, line 150: This does not what you intended it to do. This always calls `TextWriter.Write(object?)`.
It won't call different methods depending on `T`.
The original code had separate methods per type to prevent boxing.

src/NUnitFramework/framework/TestContext.cs
manfred-brands: Actually, even if the other code is _compile_ time compatible, it still isn't binary compatible.
If someone has a NUnitExtension library calling the old methods, those references no longer can be resolved.
This can only be done in a major version and as per my other comment generics here doesn't work as the code called is not generic.

src/NUnitFramework/tests/TestContextTests.cs
```diff
@@ -335,9 +335,15 @@ public async Task TestContextOut_ShouldFlowWithAsyncExecution()
         [Test]
         public async Task TestContextWriteLine_ShouldNotThrow_WhenExecutedFromAsyncMethod()
         {
-            Assert.DoesNotThrow(TestContext.WriteLine);
+            Assert.DoesNotThrow(() => TestContext.WriteLine(string.Empty));
             await YieldAsync();
-            Assert.DoesNotThrow(TestContext.WriteLine);
+            Assert.DoesNotThrow(() => TestContext.WriteLine(string.Empty));
```
manfred-brands, line 340: This change should not be done. It is the result of removing a method.

manfred-brands (changes requested):
As said in my individual comments, this change cannot be done.

NUnit.sln.DotSettings
OsirisTerje: First, this file should be in a separate PR, not here , but slipped through.

Do we want Resharper/Rider to write into the editorconfig with their specific (more finegrained) code and style rules?  
E.g.
```
# ReSharper custom properties for code formatting styles
resharper_csharp_brace_style=next_line
resharper_csharp_blank_lines_around_invocable=2

# ReSharper custom properties for code syntax styles
csharp_default_private_modifier=explicit
braces_for_ifelse=not_required

# ReSharper custom properties for code inspections
resharper_possible_null_reference_exception_highlighting=error
resharper_replace_with_string_is_null_or_empty_highlighting=none
```

Or should we just keep them in the dotsettings file?

At the same time, I - for me - don't want to spend too much time bothering about these settings.  I haven't even looked into what these "new" settings actually are.

The upgrade of this file happened automagically when I upgraded to the latest Resharper version.

The point of having this file here is to allow devs using Rider or Resharper to work on the code.  I am non-opiniated on what tools people use.

src/NUnitFramework/framework/TestContext.cs
OsirisTerje: Yes, it does, but it eliminates tons of lines of pretty much useless code but at the same time allows people to use these shortcut methods.  And performance issues should not be an issue when writing to the console, should it?

We have two console output streams (TextWriter), Out and Error.  These methods is just convenience methods for the Out stream.

The methods hides the fact that these go to the Out.  Not sure if that is good or actually bad.

Imho, I would have just killed all of these methods, but that would be a real breaking change.

PS:  If you look at the usage of these methods,  none are used in the Test parts of NUnit.  All usage goes to the string methods.  
Looking at the two most used libs, FluentAssertions and Specflow.Nunit,  none of them seems to use this either.

My goal is not to accomplish extraordinary feats, but to eliminate bloated code

But if you guys feel strongly about it, we'll just leave it as is.

src/NUnitFramework/tests/TestContextTests.cs
OsirisTerje: What does this code test?

src/NUnitFramework/framework/TestContext.cs
manfred-brands: I agree with you that it simplifies code and that performance is not really an issue here.

I also would say that they all could go so users explicitly need to use `Out` directly.

What we could do is make an analyzer rule for this with a CodeFix that inserts `.Out` between it. 
Once that is done we can Obsolete these methods and remove them. Technically the latter needs to be done at a major version.

OsirisTerje: Agree that is a good approach!  We need to create some issues for that then.

src/NUnitFramework/tests/TestContextTests.cs
manfred-brands: That is a whole other question.
It doesn't seem that any of the other `Write` methods are tested at all.

Issue #4761, opened 2024-07-16
`MultipleAssertException` should derive from `AssertionException`, or both should derive from a common e.g. `AssertionExceptionBase`


maettu-this:
https://github.com/nunit/nunit/issues/4701#issuecomment-2231695472 shows an issue with the current design, `MultipleAssertException` and `AssertionException` are currently not directly related to each other. From an OO point of view, `MultipleAssertException` is a specialization of `AssertionException`, I therefore expected to be able to catch both at once using `AssertionException`, but this is not the case.

Also, the naming seems somewhat inconsistent, I would have expected `MultipleAssertionException` rather than `MultipleAssertException`, or `AssertException` rather than `AssertionException`. Whatever way around, either both `Assertion` or `Assert`.

Potential designs:

- `AssertionExceptionBase` as base class. `AssertionException` and `MultipleAssertException` as derived classes. This would be non-breaking for sure.
- Change `AssertionException` to become the base class. `SingleAssertException` and `MultipleAssertException` as derived classes. In the majority of cases this should also be non-breaking. And this would resolve the naming inconsistency.
- `AssertionException` as base class. `MultipleAssertionException` as derived class. The latter renaming is breaking but could be made backward compatible with an `MultipleAssertException` alias.

Please consider refining.

OsirisTerje:
They both derive from `ResultStateException`.
The full hierarchy looks like this:
<img width="752" alt="image" src="https://github.com/user-attachments/assets/b3aa9c4f-fa65-41e8-bcd6-3ac754ab021d">

Making this hierarchy deeper is not something I believe will be an improvement, rather the opposite.

We can not easily change the names either, that would be a breaking change for many users, with very little benefit.  

In a code base like this we can't always go for what is currently regarded as optimal, but have to live with decisions made way back. 

In order to handle the issue of catching both, you can catch the ResultStateException and then filter on the two others.
```csharp
 catch (ResultStateException e) when (e is AssertionException or MultipleAssertException)
```

maettu-this:
@OsirisTerje, I understand the necessity of keeping compatibility, especially with long running projects. At the same time, long running projects also need to evolve where there is need.

Therefore, I am challenging the...

> change the names either, that would be a breaking change for many users

...with two experimental PRs for which I ask you to approve the workflows for. The two PRs contain minimal change and I think will pass verification. Reasoning:

- Not much changes for `MultipleAssertException`, it just derives from `AssertionException`, making it clearer this is a specialization of such.
[...]

PR #4763, opened 2024-07-21
WIP: #4761 Experiment1: MultipleAssertException deriving from AssertionException

maettu-this:
This is a first experiment on #4761 to see the impact on `MultipleAssertException` deriving from `AssertionException`.

### DIFF
--- a/.github/workflows/NUnit.CI.yml
+++ b/.github/workflows/NUnit.CI.yml
@@ -1,6 +1,7 @@
 name: Continuous Integration
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - main
--- a/.github/workflows/NUnit.CI.yml
+++ b/.github/workflows/NUnit.CI.yml
@@ -1,7 +1,6 @@
 name: Continuous Integration
 
 on:
-  workflow_dispatch:
   push:
     branches:
       - main
--- a/src/NUnitFramework/framework/Exceptions/MultipleAssertException.cs
+++ b/src/NUnitFramework/framework/Exceptions/MultipleAssertException.cs
@@ -10,7 +10,7 @@ namespace NUnit.Framework
     /// Thrown when an assertion failed.
     /// </summary>
     [Serializable]
-    public class MultipleAssertException : AssertionException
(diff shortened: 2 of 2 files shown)


OsirisTerje:
You can run these in your own fork. I believe that would be much easier for you.

maettu-this:
Hmm..., don't seem to be able to manually trigger [the workflow in my fork](
https://github.com/maettu-this/nunit/actions/workflows/NUnit.CI.yml). [Manually running a workflow](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow#configuring-a-workflow-to-run-manually) instructions don't help much either. Any clues?

OsirisTerje:
If you go to your Actions tab you should see this:
image

That alone should cause the CI action to run.

If you want to trigger it manually, add the workflow_dispatch trigger to the yaml file.  See https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow

maettu-this:
I already previously did enable the workflows. Tried a bunch of other things with no success. Then, based on [a post](https://stackoverflow.com/a/65389878/8543604), I suceeded by adding `workflow_dispatch:` to `NUnit.CI.yml` which I then had to push onto the forked `main`. Not exactly an ideal solution, especially it also triggered the `NUnit.Myget.Publish.yml` of which I don't know the impacts of. I could manually cancel that.

Summarizing:

- Yes, I could trigger the workflows on the fork.
- No, it is neither obvious nor seems the right way to do it.

Adding `workflow_dispatch:` to the original `NUnit.CI.yml` could be the better solution, but that is even less my turf to decide upon.

OsirisTerje:
When experimenting, which is just awesome that people do, you should NOT push this up to the NUnit repo.  It just creates unnecessary noise, and give us more work to maintain these things.

You should do that in your own sandbox.

If you don't like the way it works with the fork, then it is better to just clone the nunit repo, remove the remote, and add a new remote to your own repo.  
Then you can run as many experiments and build as you like.

maettu-this:
Sure, sorry for that, just wanted to share what issues I have been running into when trying to use the workflows on the fork.

OsirisTerje:
No problem, note that you can also use the [slack channels for discussions](https://nunit.slack.com), or also raise a discussion item here.

PR #4769, opened 2024-07-27
WIP: Introduce HookExtension to support high level tests Hook extensions by events

TSAVogt:
This is an alternative implementation of HookExtensions based on events.
Intension: be open for multiple users and also framework internal usage.
Related issue: https://github.com/nunit/nunit/issues/4744


stevenaw (commented):
Apologies that it took me a few extra days to review @TSAVogt . This looks good to me at a glance and based on other issues in the repo would be extremely helpful across a variety of cases. Thank you!

One question I'd have for other members of @nunit/framework-team is around the use of C# `event` keywords here. I don't think we have many other cases where we use events in this way, would a `List<SetUpTearDownHookHandler>` or perhaps `List<Action<TestExecutionContext, IMethodInfo>>` be better? 

@OsirisTerje I'm unsure if in your discussions you specifically talked about `event` usage or not

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -102,6 +102,10 @@ public TestExecutionContext(TestExecutionContext other)
             _priorContext = other;
 
             CurrentTest = other.CurrentTest;
+            if (other.HookExtension is object)
```
manfred-brands, line 105:
```suggestion
            if (other.HookExtension is not null)
```

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
```diff
@@ -397,6 +401,11 @@ public IPrincipal? CurrentPrincipal
         /// </summary>
         public int CurrentRepeatCount { get; set; }
 
+        /// <summary>
+        /// Hook Extension to support high level test extensions.
+        /// </summary>
+        public HookExtension? HookExtension { get; } = new();
```
manfred-brands, line 407: As this is likely only used by some testing code, remove the initializer.
Developers that use it must explicitly set the Hook.

src/NUnitFramework/framework/Internal/HookExtension.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// </summary>
+        public HookExtension()
+        {
+            BeforeAnySetUps = (context, method) => { };
+            AfterAnySetUps = (context, method) => { };
+            BeforeTest = (context, testMethod) => { };
+            AfterTest = (context, testMethod) => { };
+            BeforeAnyTearDowns = (context, method) => { };
+            AfterAnyTearDowns = (context, method) => { };
+        }
```
manfred-brands, line 24: Not needed, code below deals with null.

src/NUnitFramework/framework/Internal/HookExtension.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="HookExtension"/> class by copying hooks from another instance.
+        /// </summary>
+        /// <param name="other">The instance of <see cref="HookExtension"/> to copy hooks from.</param>
+        public HookExtension(HookExtension other) : this()
+        {
+            other.BeforeAnySetUps?.GetInvocationList()?.ToList().ForEach(d => BeforeAnySetUps += d as SetUpTearDownHookHandler);
```
manfred-brands, line 32: Are you only calling `ToList` so you can call `ForEach`?
According to [Nick Chapsas  ](https://www.youtube.com/watch?v=jUZ3VKFyB-A) the `ForEach` method is 3 times slower than iterating yourself and that is if you already have a list to begin with. It is worse when you first have to allocate one. `GetInvocationList()` returns an array and can be looped over directly.

manfred-brands (changes requested):
We have to remember that the delegates call into "end-user" code which can fail.
When using delegates, if the first callback fails, no other callback is called.
The normal behaviour do safely use this would be to get the delegate list and call each handler inside a `try/catch`.

manfred-brands:
> One question I'd have for other members of @nunit/framework-team is around the use of C# `event` keywords here. I don't think we have many other cases where we use events in this way, would a `List<SetUpTearDownHookHandler>` or perhaps `List<Action<TestExecutionContext, IMethodInfo>>` be better?

The `event` does give you the '+=' and '-=' code for free. If using a `List` you have to add that code.
The `event` allows you to call all delegate with one call, but as I said in the review this is bad as one handler could stop other handlers from being called. Especially if those are `CleanUp` handlers that close resources they might leak.

In my work code we have a `TryRaise` extension method which iterates over the delegates returned from `GetInvocationList()` and calls each one of them safely.

stevenaw:
Ok thanks that makes sense too @manfred-brands 
Mostly asked in case there was an unspoken preference within the team. I like the syntax sugar for adding and removing callbacks if we are fine with events.

src/NUnitFramework/framework/Internal/HookExtension.cs
TSAVogt: As I expect not hundreds or thousands registered hooks I did not care about performance and instead focused on readability and compact code.
Do you think this could be an error?

jnm2:
I added a [comment in the issue](https://github.com/nunit/nunit/issues/4744#issuecomment-2312744124) about an alternative to events for supporting async setup work being attached to these hooks.

TSAVogt:
@manfred-brands, @stevenaw: Regarding **error handling**, I had the following approach in my mind (synchronous events):

Errorhandling HookExtensions excalidraw

So the error behavior for before and after hooks is similar to failing SetUp or OneTimeSetUp methods in relation to tests. A failing hook will stop subsequent hooks of the same type from being executed, including the hooked method call (SetUp in the figure). Because of this failure in the Setup context, the test will not run either. To give the upcoming tests a change to execute, the TearDown including its hooks shall be called.

But this only makes sense if we have synchronous or a mix of synchronous / asynchronous hooks like @jnm2 has shown in his example: https://github.com/nunit/nunit/issues/4744#issuecomment-2312744124

=> let's first decide in the issue about the kind of hooks we should go for. Okay?

src/NUnitFramework/framework/Internal/TestExecutionContext.cs
TSAVogt: If the users get a setter method to do this, we need an additional check that they do not overwrite some already set hook extension with registered event handlers.
To avoid unnecessary object instantiations, I would propose storing the instance in a private member and instantiate on the first get access.

src/NUnitFramework/framework/Internal/HookExtension.cs
TSAVogt: But this can be also easily done be a small helper.
But we do not need the reverse order if we got for async hooks => wait for feedback on sync or async here....

TSAVogt:
I have just pushed the adaptions based on the proposal of @jnm2. The hook extensions now also support async event handlers. I have also added two basic tests, to show that the approach in principle works.
Currently open points:
- more test cases needed
- existing code needs better error handling (at the moment I have ignored some possible null references)
- TestUtilities for TestsUnderTest need some improvements
- as @jnm2 proposed: ensure that hooks are executed in the same thread as the test.

ckolumbus:
Hi all, is there anything missing in order to push this further? From the disucssions in the linked issue and here in the PR I'd think that this could make it out of draft status. Or do I overlook some open points that need further clarification?

OsirisTerje:
@TSAVogt needs to clarify if they want to continue this PR.  Otherwise, if someone else is willing to continue this work.

TSAVogt:
Hi @OsirisTerje ,
that's a good question. 
I will not be available most of the time next week. I try to contact you at the end of next week.
We currently have a very good overview over all needed changes and are on the way to collect additional feedback from our 200000 existing test cases challanging the hook extensions we are adding here... 
I would like to discuss what is the best approach to go ahead here: contribute the whole extensions by some more PRs or does an explicit branch, that you also mentioned in the past, makes more sense here...

TravellerD14:
@TSAVogt Please close this PR as we have a successor: https://github.com/nunit/nunit/pull/4986

OsirisTerje:
Superseded by #4986

PR #4772, opened 2024-07-30
Add support to tuples in inline values (#4771)

### LINKED ISSUE #4771: Support tuples in inline data
*Feature request*

Add support to passing tuples in inline parameter values. This would simplify passing complex values while using combinatorial and pairwise test cases generation.

Example usage:

```csharp
[Test]
public void Test
(
    [Values(new object[] {"a", 1}, new object[] {"b", 2})]
    (string, int) a,
    [Values(1, 2, 3)]
    int b
)
```

Currently it can be achieved with `ValueSource`. Note that this feature will also need an update to `NUnit1031` diagnostic.

MaxKot:
Fixes #4771 

Please note that merging this also requires an update to `NUnit1031` diagnostic in the analyzers repository.


src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
```diff
@@ -118,8 +118,66 @@ public static bool TryConvert(object? value, Type targetType, [NotNullWhen(true)
...
+                return true;
+            }
+
+            if (genericType.Assembly == typeof(ValueTuple).Assembly &&
+                genericType.Namespace == typeof(ValueTuple).Namespace &&
+                genericType.Name.StartsWith("ValueTuple`", StringComparison.Ordinal))
+            {
+                typeArgs = type.GetGenericArguments();
+                return true;
+            }
```
jnm2, line 177: There should probably be tests of tuples with more than seven elements, since the C# type `(T1, T2, T3, T4, T5, T6, T7, T8)` maps to the type `ValueTuple<T1, T2, T3, T4, T4, T5, T6, T7, ValueTuple<T8>>`.

(System.Tuple and System.ValueTuple require the TRest parameter to be a tuple, and this mapping is recursive to allow any number of elements.)

MaxKot: Thanks! Fixed in [14c321b](https://github.com/nunit/nunit/pull/4772/commits/14c321b53f3c91670f19f7a00aa5b1a408a3f36a).

OsirisTerje:
@manfred-brands @stevenaw  Can you have a look at this?   It's a bit of code here. Are we good with this?

manfred-brands (approved):
@MaxKot Thanks for your contribution.
It looks good.
I have made one possible suggestion to replace the string comparisons.

stevenaw (approved):
Thanks @MaxKot this looks great and seems to cover many of the edge cases I can think of. Thanks again for your patience on this review, as well as for your contribution.

Issue #4584, opened 2023-12-07
Nunit 4.0.x Test case selection is incorrect with certain test data


johnjaylward:
We have certain test cases that are using test data and expected results.

When the tests run in Microsoft Visual Studio 2022 (64-bit) Version 17.8.2, the tests get ignored.

I've created a sample project that recreates the issue: https://github.com/johnjaylward/Nunit4_TestcaseSelectionIssue

I managed to find a work around that does let the tests run by adding another Test with the same name, but does not use parameters.

In the test project the `SelectionFail` tests show the bug while the `SelectionSuccess` test shows the workaround.

![image](https://github.com/nunit/nunit/assets/383703/c626766e-5f5f-47d7-8e02-dd966d21b758)

OsirisTerje:
I tried your code, and I do see the same in Visual Studio.  
When I add a runsettings file, and dump the content from the framework/engine, I also see that no tests from SelectionFail appears. They are sent down from the explorer/adapter, but doesnt come back. 

It is something with the naming here, if I do a SetName on each data to a "normal" name, then it works.  

![image](https://github.com/nunit/nunit/assets/203432/3b1fdc1a-f5fe-4349-846d-59bd66eca37c)

I also see that it works from the command line.
Looking at the dump files, the dotnet test doesn't use any filter, as expected, in this case

```xml
<TestFilter>
   <filter/>
</TestFilter>
```

Using Test Explorer, it does send all down as a list of testcases, so we then get a rather ugly filter:
(just start of shown)

```xml
  <filter><or><test>Nunit4_TestcaseSelectionIssue.SelectionFail.TestFailure([
  1,
  2,
  3,
  4,
  5,
  6
],4)</test><test>Nunit4_TestcaseSelectionIssue.SelectionFail.TestFailure([
  1,
  2,
  3,
  4,
  5,
  6
],2)</test><test>Nunit4_TestcaseSelectionIssue.SelectionFail.TestFailure([
  1,
```

There is then something in the framework/engine that doesn't like these test names.  Why an extra test should make them appear again beats me right now.

As a workaround:  You can force it to not use any filter if you add `<AssemblySelectLimit>2</AssemblySelectLimit>` (with any number less than the number of tests you have) to your .runsettings file.  Then it will always run all tests.

johnjaylward:
ok, instead of using the default `.runsettings` as provided by MS, I used the one in this repo, and I got similar results to what you showed:

[...]

PR #4773, opened 2024-08-02
Test selection issue with normalized newlines

stevenaw:
Fixes https://github.com/nunit/nunit/issues/4584

~An initial pass to turn off line ending normalization while we decide on final approach. Still room for improvement and optimizations~

### DIFF
--- a/src/NUnitFramework/framework/Internal/DisplayName.cs
+++ b/src/NUnitFramework/framework/Internal/DisplayName.cs
@@ -223,6 +223,22 @@ namespace NUnit.Framework.Internal
             return display;
         }
 
+        public static string EscapeControlChars(string s)
+        {
+            if (MayNeedEscape(s))
+            {
+                // cleanup
+                var sb = new StringBuilder();
+                foreach (char c in s)
+                {
+                    sb.Append(EscapeControlChar(c));
+                }
+                s = sb.ToString();
+            }
+
+            return s;
+        }
+
         private static string FormatException(Exception ex)
         {
             var builder = new StringBuilder();
--- a/src/NUnitFramework/framework/Internal/DisplayName.cs
+++ b/src/NUnitFramework/framework/Internal/DisplayName.cs
@@ -223,22 +223,6 @@ namespace NUnit.Framework.Internal
             return display;
         }
 
-        public static string EscapeControlChars(string s)
-        {
-            if (MayNeedEscape(s))
-            {
-                // cleanup
-                var sb = new StringBuilder();
-                foreach (char c in s)
-                {
-                    sb.Append(EscapeControlChar(c));
-                }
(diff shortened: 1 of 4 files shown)


src/NUnitFramework/tests/Issue4584.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    [TestFixture]
+    public class Issue4584
+    {
+        #region Multiline selection failure
+        [TestFixture(Description = "https://github.com/nunit/nunit/issues/4584")]
+        public class SelectionFail
+        {
+            [TestCaseSource(nameof(Data)), Explicit]
+            public void TestFailure(JsonArray jsonArray, int maxSize)
```
stevenaw, line 17: I've been using this for investigation.
Eventually, if we do decide to make code changes, I will update this test to instead discover the copy that lives in `testdata` and validate it instead

manfred-brands: I have updated this with a test showing the filter issue and not depending on Json.

manfred-brands (approved):
Your fix makes the code compatible with v3 with the minimum amount of changes.

src/NUnitFramework/framework/Internal/DisplayName.cs
```diff
@@ -223,6 +223,22 @@ public static string GetValueString(object? arg, int stringMax)
             return display;
         }
 
+        public static string EscapeControlChars(string s)
```
stevenaw, line 226: @manfred-brands Was this still needed? I don't see it referenced

src/NUnitFramework/tests/Issue4584.cs
```diff
@@ -0,0 +1,36 @@
...
+using NUnit.Framework.Tests.TestUtilities;
+using NUnit.TestData.TestCaseSourceAttributeFixture;
+
+namespace NUnit.Framework.Tests
+{
+    [TestFixture]
+    public class Issue4584
+    {
+        [Test]
+        public void DiscoverAndExecution()
```
stevenaw, line 14: Looks great. Thanks @manfred-brands !

src/NUnitFramework/framework/Internal/DisplayName.cs
manfred-brands: @stevenaw I added it initially to allow escaping characters for all TestNames. Similar like what is done in case the parameter is a string. As we haven't decided on an approach, I added a commit to remove it.
That allows this branch to be merged and 4.2 to be released.

stevenaw: Thanks @manfred-brands . Definitely nice to get this merged in case it's blocking others from upgrading

PR #4779, opened 2024-08-11
Update the dependencies

### LINKED ISSUE #4778: Update dependencies
A few dependencies of the framework or its tests have had newer versions since released. It would be good to do an update, ideally prior to next release of the framework.

stevenaw:
Fixes #4778 
Updates dependencies of the framework and test projects

### DIFF
--- a/src/NUnitFramework/Directory.Packages.props
+++ b/src/NUnitFramework/Directory.Packages.props
@@ -6,30 +6,30 @@
   <!-- Packages for used features -->
   <ItemGroup>
     <PackageVersion Include="System.Collections.Immutable" Version="6.0.0" />
-    <PackageVersion Include="System.Memory" Version="4.5.4" />
+    <PackageVersion Include="System.Memory" Version="4.5.5" />
     <PackageVersion Include="System.ValueTuple" version="4.5.0" />
   </ItemGroup>
   <!-- General Packages -->
   <ItemGroup>
     <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="8.0.0" />
-    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.9.0" />
-    <PackageVersion Include="NUnit3TestAdapter" Version="4.5.0" />
-    <PackageVersion Include="FSharp.Core" Version="7.0.200" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.10.0" />
+    <PackageVersion Include="NUnit3TestAdapter" Version="4.6.0" />
+    <PackageVersion Include="FSharp.Core" Version="8.0.400" />
   </ItemGroup>
   <!-- Package to allow Nullability etc. in older SDKS -->
   <ItemGroup>
     <PackageVersion Include="IsExternalInit" Version="1.0.3" />
     <PackageVersion Include="Nullable" Version="1.3.1" />
     <PackageVersion Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
-    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
+    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
     <PackageVersion Include="TunnelVisionLabs.ReferenceAssemblyAnnotator" Version="1.0.0-alpha.160" />
   </ItemGroup>
   <!-- Analyzers -->
   <ItemGroup>
-    <PackageVersion Include="CSharpIsNullAnalyzer" Version="0.1.495" />
+    <PackageVersion Include="CSharpIsNullAnalyzer" Version="0.1.593" />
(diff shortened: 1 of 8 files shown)


src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
```diff
@@ -57,7 +57,7 @@ protected override bool Matches(string? actual)
         /// than <paramref name="comparisonType"/> was already set.</exception>
         public SubstringConstraint Using(StringComparison comparisonType)
         {
-            if (_comparisonType == null)
+            if (_comparisonType is null)
```
stevenaw, line 60: Flagged by updating `CSharpIsNullAnalyzer`. `_comparisonType` is a nullable enum which must now be detectable

src/NUnitFramework/nunit.framework.legacy.tests/nunit.framework.legacy.tests.csproj
```diff
@@ -16,7 +16,10 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" />
     <PackageReference Include="NUnit3TestAdapter" />
-    <PackageReference Include="NUnit.Analyzers" />
+    <PackageReference Include="NUnit.Analyzers">
+      <PrivateAssets>all</PrivateAssets>
```
stevenaw, line 20: Autogenerating by the VS tooling. I opted not to revert it as it may continue to keep happening on each update, but open to thoughts or ideas

src/NUnitFramework/tests/Assertions/AssertThrowsTests.cs
```diff
@@ -22,7 +22,7 @@ public void AssertThrowsDoesNotDiscardOutput()
             Assert.Throws<Exception>(() =>
             {
                 Console.WriteLine(2);
-                TestContext.WriteLine(3);
+                TestContext.Out.WriteLine(3);
```
stevenaw, line 25: Flagged by updating the NUnit analyzer.

src/NUnitFramework/tests/TextOutputTests.cs
```diff
@@ -139,14 +139,18 @@ public void TestContextOut_WritesToResult()
         [Test]
         public void TestContextWrite_WritesToResult()
         {
+#pragma warning disable NUnit1033 // The Write methods on TestContext will be marked as Obsolete and eventually removed
```
stevenaw, line 142: disabled as the name of the test explicitly checks the deprecated method. `TestContextOut_WritesToResult()` above is tested separately

src/NUnitFramework/nunit.framework.legacy.tests/nunit.framework.legacy.tests.csproj
manfred-brands: Because we use `Directory.Packages.props` there is no need to update the `.csproj` at all.
VS tooling doesn't seem to recognize that.
I normally just update the version number manually.

src/NUnitFramework/Directory.Packages.props
```diff
@@ -6,30 +6,30 @@
   <!-- Packages for used features -->
   <ItemGroup>
     <PackageVersion Include="System.Collections.Immutable" Version="6.0.0" />
```
stevenaw, line 8: There was something about higher versions of this package which caused an issue for the tests last time around too. For now I've left alone as it is only used by our tests

src/NUnitFramework/nunit.framework.legacy.tests/nunit.framework.legacy.tests.csproj
stevenaw: Happy to do that, thanks! I'll give this another go tomorrow evening again.

We should probably squash merge this when the time comes to keep the history clean

stevenaw: Updated. I've adjusted this and squashed on my end so that the PR itself is simpler.
Thanks for the suggestion!

OsirisTerje (approved):
This seems to be good to me.  I resolved the conversation comments since they all seemed to be done.  
Anything more @stevenaw ?  Can we merge it?

stevenaw:
Good on my end. Thanks @OsirisTerje

OsirisTerje:
@stevenaw Just noted these using dotnet outdated:
image

stevenaw:
Hmm. Thanks for noting that @OsirisTerje 
I would've thought some of those listed versions have been updated by the PR. I can take a closer look later on today.

stevenaw:
@OsirisTerje That command didn't work for me on the default SDK install, but I have tried using `dotnet list package --outdated` and it shows something different

image

Do you know if you're using `dotnet-outdated`? It looks like it should also support [centralized package management](https://github.com/dotnet-outdated/dotnet-outdated/pull/243) so I'm a little confused.

OsirisTerje:
Yes, I use `dotnet outdated`, and it seems it understands the CPM.  

I also tried the package list, and got these results  A bit different from yours,  do you run it on the framework directory ?
image

PR #4799, opened 2024-08-23
Fixes #4798

### LINKED ISSUE #4798: Add a test to ensure direct framework dependencies in csproj are reflected in nuspec
We currently have to manually ensure framework dependencies in the csproj are also reflected in nuspec. We should consider adding some sort of test coverage to validate this to give us a safety net for future changes.

OsirisTerje:
Fixes #4798 

I followed the idea from @stevenaw and made this into a test.
The test is in folder for SolutionTests, indicating that it doesn't test code, but tests the setup of the projects.

Writing code like this with XML parsing and so is not something I am very fond out, so I did take some "help" from the ChatGPT "friend".   The parsing code it did pretty well, but the logic - that was more of a struggle.  I believe I have managed to clean up most of that, after NNN iterations, but there certainly some more weirdoes around there.  Appreciate some good comments!

The first commit shows the error, so the action should go red.  
The next commit will fix that error, and should go green. 

The code takes the csproj as the "truth" and checks that the nuspec have the same setup.  Right now it says the System.Memory is missing in the .net6.0 dependency group in the nuspec, but we know that is should instead be placed under the .netframework condition in the csproj.  Some thinking is still needed from the devs :-)

### DIFF
--- a/src/NUnitFramework/tests/SolutionTests/NuspecDependenciesTests.cs
+++ b/src/NUnitFramework/tests/SolutionTests/NuspecDependenciesTests.cs
@@ -42,7 +42,7 @@ namespace NUnit.Framework.Tests.SolutionTests
         }
 
         [Test]
-        public void NUnitFrameworkNuspecContainsCorrectDependencies()
+        public void AllPackagesInCsprojIsInNuspec()
         {
             VerifyDependencies.ComparePackages(_csprojPackages, _nuspecPackages);
         }
--- a/src/NUnitFramework/tests/SolutionTests/NuspecDependenciesTests.cs
+++ b/src/NUnitFramework/tests/SolutionTests/NuspecDependenciesTests.cs
@@ -15,7 +15,9 @@ namespace NUnit.Framework.Tests.SolutionTests
     {
         private Dictionary<string, List<string>> _nuspecPackages;
         private Dictionary<string, List<string>> _csprojPackages;
+
         private const string Root = "../../../../../../";
+        private const string NotSpecified = nameof(NotSpecified);
 
         [OneTimeSetUp]
         public void OneTimeSetUp()
@@ -41,8 +43,6 @@ namespace NUnit.Framework.Tests.SolutionTests
         internal sealed class CsprojReader
         {
             private const string PathToFrameworkFolder = $"{Root}/src/NUnitFramework/";
-
-            public const string NotSpecified = nameof(NotSpecified);
 
             private string Xml { get; }
 
@@ -118,7 +118,7 @@ namespace NUnit.Framework.Tests.SolutionTests
                     foreach (var group in dependenciesSection.Elements(ns + "group"))
                     {
                         // Get the targetFramework attribute
-                        var framework = group.Attribute("targetFramework")?.Value ?? CsprojReader.NotSpecified;
+                        var framework = group.Attribute("targetFramework")?.Value ?? NotSpecified;
 
                         // Find all dependency elements in the current group
                         var dependencies = group.Elements(ns + "dependency")
@@ -149,7 +149,7 @@ namespace NUnit.Framework.Tests.SolutionTests
                 // Iterate through the frameworks in the csprojPackages dictionary
                 foreach (var csprojFramework in csprojPackages.Keys)
                 {
-                    if (csprojFramework == CsprojReader.NotSpecified)
+                    if (csprojFramework == NotSpecified)
                     {
                         TestContext.Out.WriteLine("Checking for packages that should be in all frameworks in the .nuspec file");
                         // Check if the packages from the csproj are present in all nuspec framework
--- a/src/NUnitFramework/tests/SolutionTests/NuspecDependenciesTests.cs
+++ b/src/NUnitFramework/tests/SolutionTests/NuspecDependenciesTests.cs
@@ -11,15 +11,17 @@ namespace NUnit.Framework.Tests.SolutionTests
     /// <summary>
     /// Checks that the csproj of the framework projects are reflected correctly in the nuspec files.
     /// </summary>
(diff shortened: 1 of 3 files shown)


OsirisTerje:
Ok... stylecop..   love you....  
That's for tomorrow.....

manfred-brands:
> Ok... stylecop.. love you.... That's for tomorrow.....

I have fixed this. StyleCop is your friend and for my work we have rules set to error so they will be fixed before CI kicks in. Initially it is a hassle for developers new to the code, but in the end they all love it as they don't have to do rework.
Anyhow, I accepted that people here wanted these as warnings as to not slow initial development.

I added a few more changes to my commit:
1. Ensure all paths use a forward / as separator so it works for both Windows and Unix like OS.
2. Ensure casing of file names is correct so it works on Case Sensitive file systems.
3. Moved your classes as nested inside the unit test. If reviewing my changes use the ignore whitespace option.
4. Small optimization for the double dictionary lookup.

src/NUnitFramework/tests/SolutionTests/NuspecDependenciesTests.cs
```diff
@@ -0,0 +1,174 @@
...
+
+                // Find the dependencies section
+                var dependenciesSection = doc.Descendants(ns + "dependencies").FirstOrDefault();
+                if (dependenciesSection is not null)
+                {
+                    // Iterate over all group elements within the dependencies block
+                    foreach (var group in dependenciesSection.Elements(ns + "group"))
+                    {
+                        // Get the targetFramework attribute
+                        var framework = group.Attribute("targetFramework")?.Value ?? "Both";
```
manfred-brands, line 108: What is "Both"? Is this the equivalent of no condition (NotSpecified)? Or is that not allowed in a nuspec?

src/NUnitFramework/tests/SolutionTests/NuspecDependenciesTests.cs
```diff
@@ -0,0 +1,174 @@
...
+        private const string Root = "../../../../../../";
+
+        [Test]
+        public void NUnitFrameworkNuspecContainsCorrectDependencies()
+        {
+            var nuspec = new NuspecReader("framework/nunit.nuspec");
+            var csproj = new CsprojReader("framework/nunit.framework.csproj");
+            var nuspecPackages = nuspec.ExtractNuspecPackages();
+            var csprojPackages = csproj.ExtractCsprojPackages();
+            VerifyDependencies.ComparePackages(csprojPackages, nuspecPackages);
```
manfred-brands, line 25: This checks that the nuspec contains all packages from the csproj.
Should we also do a reverse check to ensure that if we drop a package from the csproj it is also dropped from the nuspec?
Also do we need to add a separate test to ensure that the nuspec contains entries for all supported TargetFramework?

manfred-brands (commented):
@OsirisTerje Code looks good. I asked a few questions to see if we need more checks.

src/NUnitFramework/tests/SolutionTests/NuspecDependenciesTests.cs
OsirisTerje: Yes, that should be NotSpecified.

src/NUnitFramework/tests/SolutionTests/NuspecDependenciesTests.cs
OsirisTerje: Yes, agree!

manfred-brands:
> PS: Wish that StyleCop was a Roslyn analyzer with the possibility to auto fix all these things.

It is and has many CodeFixes

OsirisTerje:
But is that a separate analyzer to be installed, or do we have it already?

src/NUnitFramework/tests/SolutionTests/NuspecDependenciesTests.cs
OsirisTerje: But, that should really not happen, or.  That means we would have a section in the nuspec which would apply for all frameworks, if that is even allowed.....

manfred-brands:
@OsirisTerje Thanks for the excellent start, I added more functionality:

1. Parameterized the test so it now also checks `nunitlite.nuspec`
2. Augmented the test to also check the version number of the packages so that if we update in source the nuspec also gets updated.

I think we are done, but as I should not approve things I worked on @stevenaw can you do the final review please?

src/NUnitFramework/tests/SolutionTests/NuspecDependenciesTests.cs
```diff
@@ -0,0 +1,283 @@
...
+        [OneTimeSetUp]
+        public void OneTimeSetUp()
+        {
+            _nuspecPackages = NuspecReader.ExtractNuspecPackages(_nuspecPath);
+            _csprojPackageVersions = PackageVersionReader.ExtractPackageVersions(_propsPath);
+            _csprojPackages = CsprojReader.ExtractCsprojPackages(_csprojPath, _csprojPackageVersions);
+        }
+
+        [Test]
+        public void NUnitFrameworkNuspecContainsCorrectDependencies()
```
stevenaw, line 45:
```suggestion
        public void AllPackagesInCsprojIsInNuspec()
```

nitpick: Given we do the inverse test below named `AllPackagesInNuspecIsInCsproj`, it may be clearer to name this test in a similar way

stevenaw (approved):
Thanks @OsirisTerje @manfred-brands 
LGTM. Seems robust and no real concerns of any note on my end. Passes locally for me. One nitpick suggestion about test naming consistency and clarity

stevenaw:
Thanks for accepting the suggestion @OsirisTerje 
This looks good to me if it looks good to @manfred-brands

PR #4813, opened 2024-09-05
Add Constraint ApplyAsync

### LINKED ISSUE #4811: Assert.ThatAsync does not support polling
```csharp
        var i = 0;
        async Task<string> GetResult()
        {
            return new string('1', i++);
        }

        await Assert.ThatAsync(GetResult, Is.Not.Empty.After(1000, 10));
```

Expected Result:
GetResult function returns non-empty result on second execution. So assertion should pass, since it has polling interval.

Actual Result:
Nunit invokes GetResult only once, therefore it fails.
```
Assert.That(GetResult, Is.Not.Empty.After(1000, 10))
  Expected: not <empty> after 1000 milliseconds delay
  But was:  <string.Empty>
```

Dreamescaper:
Fixes #4811 .

This PR is created for futher discussions. 
Instead of ThatAsync awaiting the result and handling the exception, it delegates it to constraints - via ApplyAsync method. 
Therefore, DelayedConstraint supports async tasks as well.

Obviusly, this is a breaking change. It will break anyone who implements IConstraint directly (instead of inheriting from Constraint).
Workarounds are possible to avoid that breaking change, but I'm not sure how ofter this interface is implemented direclty.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
@@ -259,7 +259,6 @@ namespace NUnit.Framework.Constraints
             long now = Stopwatch.GetTimestamp();
             long delayEnd = TimestampOffset(now, DelayInterval.AsTimeSpan);
 
-            object? actual;
             if (PollingInterval.IsNotZero)
             {
                 long nextPoll = TimestampOffset(now, PollingInterval.AsTimeSpan);
@@ -269,11 +268,9 @@ namespace NUnit.Framework.Constraints
                         ThreadUtility.BlockingDelay((int)TimestampDiff(delayEnd < nextPoll ? delayEnd : nextPoll, now).TotalMilliseconds);
                     nextPoll = TimestampOffset(now, PollingInterval.AsTimeSpan);
 
-                    actual = InvokeDelegate(del);
-
                     try
                     {
-                        ConstraintResult result = BaseConstraint.ApplyTo(actual);
+                        ConstraintResult result = BaseConstraint.ApplyTo(del);
                         if (result.IsSuccess)
                             return new DelegatingConstraintResult(this, result);
                     }
@@ -286,8 +283,7 @@ namespace NUnit.Framework.Constraints
             if ((now = Stopwatch.GetTimestamp()) < delayEnd)
                 ThreadUtility.BlockingDelay((int)TimestampDiff(delayEnd, now).TotalMilliseconds);
 
-            actual = InvokeDelegate(del);
-            return new DelegatingConstraintResult(this, BaseConstraint.ApplyTo(actual));
+            return new DelegatingConstraintResult(this, BaseConstraint.ApplyTo(del));
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
@@ -338,7 +338,8 @@ namespace NUnit.Framework.Constraints
                 while ((now = Stopwatch.GetTimestamp()) < delayEnd)
                 {
                     if (nextPoll > now)
-                        ThreadUtility.BlockingDelay((int)TimestampDiff(delayEnd < nextPoll ? delayEnd : nextPoll, now).TotalMilliseconds);
+                        await Task.Delay(TimestampDiff(delayEnd < nextPoll ? delayEnd : nextPoll, now));
+
                     nextPoll = TimestampOffset(now, PollingInterval.AsTimeSpan);
 
                     try
@@ -353,8 +354,9 @@ namespace NUnit.Framework.Constraints
                     }
                 }
             }
+
             if ((now = Stopwatch.GetTimestamp()) < delayEnd)
-                ThreadUtility.BlockingDelay((int)TimestampDiff(delayEnd, now).TotalMilliseconds);
+                await Task.Delay(TimestampDiff(delayEnd, now));
 
             return new DelegatingConstraintResult(this, await BaseConstraint.ApplyToAsync(taskDel));
         }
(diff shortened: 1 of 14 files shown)


src/NUnitFramework/framework/Constraints/IConstraint.cs
```diff
@@ -61,6 +64,16 @@ public interface IConstraint : IResolveConstraint
...
+        /// <summary>
+        /// Applies the constraint to a delegate that returns the task.
+        /// The default implementation simply evaluates the delegate and awaits the task
+        /// but derived classes may override it to provide for delayed processing.
+        /// </summary>
+        /// <typeparam name="TActual"></typeparam>
+        /// <param name="delTask"></param>
+        /// <returns></returns>
+        Task<ConstraintResult> ApplyToAsync<TActual>(Func<Task<TActual>> delTask);
+
```
manfred-brands, line 76: Could we add this to an `IAsyncConstraint` interface?
When resolving the constraint in `ThatAsync` you can then test on `IAsyncConstraint`?
If not, throw a `NotSupportedException`?

That way it would be a non-breaking change.

manfred-brands (commented):
@Dreamescaper I like the idea and the implementation is relatively simple and even makes the `ThatAsync` simpler.
I had one question/suggestion to make this non-breaking.

src/NUnitFramework/framework/Constraints/IConstraint.cs
Dreamescaper: That makes sense.
I've made IAsyncConstraint internal for now - in case we'd need to more task-related methods later.

src/NUnitFramework/tests/Constraints/DelayedConstraintTests.cs
```diff
@@ -241,7 +241,7 @@ public void ThatPollingCallsDelegateCorrectNumberOfTimes()
             Assert.That(PollCount, Is.EqualTo(4).After(110, 25));
         }
 
-        [Test]
+        [Test, Platform(Exclude = "MACOSX", Reason = "Doesn't seem to work correctly with timing, something to ponder later")]
```
manfred-brands, line 244: Drop this as a re-run of the build caused MacOs to pass. The machine github uses for MacOs seems to be very slow.

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -324,6 +326,39 @@ public override ConstraintResult ApplyTo<TActual>(ref TActual actual)
...
+            long now = Stopwatch.GetTimestamp();
+            long delayEnd = TimestampOffset(now, DelayInterval.AsTimeSpan);
+
+            if (PollingInterval.IsNotZero)
+            {
+                long nextPoll = TimestampOffset(now, PollingInterval.AsTimeSpan);
+                while ((now = Stopwatch.GetTimestamp()) < delayEnd)
+                {
+                    if (nextPoll > now)
+                        ThreadUtility.BlockingDelay((int)TimestampDiff(delayEnd < nextPoll ? delayEnd : nextPoll, now).TotalMilliseconds);
```
manfred-brands, line 341: Should we not use non-blocking' delay in an `async` method?
```suggestion
                        await Task.Delay(TimestampDiff(delayEnd < nextPoll ? delayEnd : nextPoll, now));
```

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -324,6 +326,39 @@ public override ConstraintResult ApplyTo<TActual>(ref TActual actual)
...
+                            return new DelegatingConstraintResult(this, result);
+                    }
+                    catch (Exception)
+                    {
+                        // Ignore any exceptions when polling
+                    }
+                }
+            }
+            if ((now = Stopwatch.GetTimestamp()) < delayEnd)
+                ThreadUtility.BlockingDelay((int)TimestampDiff(delayEnd, now).TotalMilliseconds);
```
manfred-brands, line 357: Same here:
```suggestion
                await Task.Delay(TimestampDiff(delayEnd, now));
```

manfred-brands (changes requested):
Suggestion to replace `BlockingDelay` in `async` `DelayedConstraint` with `Task.Delay`

src/NUnitFramework/tests/Constraints/DelayedConstraintTests.cs
Dreamescaper: I can see that other tests have this exclude as well, so I assume that it does happen from time time. Still, removed.

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
Dreamescaper: Makes sense, done.

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
Dreamescaper: Makes sense, done.

src/NUnitFramework/tests/Constraints/DelayedConstraintTests.cs
Dreamescaper: ...and it failed again :)

manfred-brands (approved):
Thanks @Dreamescaper Looks good to me now.

stevenaw (approved):
LGTM too. Thanks again for another contribution @Dreamescaper !

OsirisTerje:
Just noticed:

> Obviusly, this is a breaking change. It will break anyone who implements IConstraint directly (instead of inheriting from Constraint).

Is this so?   
Then we need to update the release notes that this is a breaking change.

@Dreamescaper @manfred-brands @stevenaw

Dreamescaper:
@OsirisTerje 
No, @manfred-brands suggested a way to avoid it.

Dreamescaper:
@OsirisTerje 
On a second thought, there IS a breaking change. Just not the one I've described here.
I've mentioned it here:
https://github.com/nunit/nunit/pull/4908

PR #4815, opened 2024-09-05
Make DelayedConstraint work with ThrowsContraint

### LINKED ISSUE #4281: Throws and Delayed (.After) Constraints do not cooperate, resulting in incorrectly failing test
**NUnit Version:** 3.13.3 (latest stable as of this posting)

**Goal:** I am attempting to repeatedly check a method to Assert that it eventually throws an Exception within a certain amount of time. (Specifically, looking to get a 401 back from a webcall after a delete, where the deletion does not always occur instantly).

**Approach:** I would like to do this by using a constraint of the form `Throws.<...>.After.<...>`.

**Issue:** Unfortunately, the combination of these constraints results in the test failing due to the thrown Exception, rather than passing due to the Assert that the Exception will be thrown. It appears that that nunit gets snagged on the Delayed constraint, never making it to the Throws constraint to see that the Exception was actually the expected outcome.

Note that `.After` works fine with other constraints, such as `Is`. Similarly, `Throws` works fine when I leave off the `.After` constraint, simply checking that the method throws the first time it is called. 

**Example:** The commented-out counting code was implemented to simulate the repeated calls, e.g., 4 calls before the expected error is thrown. But the presence or absence of that code has no impact on the outcome.
```
    [Test]
    public void TestThrowsAfter()
    {
      Assert.That(() => ThrowingMethod(), Throws.TypeOf<ArgumentException>().After(5).Seconds.PollEvery(500));
    }

    private int count;
    private int ThrowingMethod()
    {
      Thread.Sleep(50);
      // Uncommenting this code has no effect on the outcome:
      //if (count < 5)
      //{
      //  return ++count;
      //}
      throw new ArgumentException("test");
    }
```

**Example Result:** When I run the above, I get a test failure, with the following readout:
```
System.ArgumentException : test
   at Tests.ThrowingMethod() in <throw line>
   at Tests.<TestThrowsAfter>b__48_0() in <Assert.That line>
   at NUnit.Framework.Constraints.DelayedConstraint.InvokeDelegate[T](ActualValueDelegate`1 del)
   at NUnit.Framework.Constraints.DelayedConstraint.ApplyTo[TActual](ActualValueDelegate`1 del)
   at NUnit.Framework.Assert.That[TActual](ActualValueDelegate`1 del, IResolveConstraint expr, String message, Object[] args)
   at NUnit.Framework.Assert.That[TActual](ActualValueDelegate`1 del, IResolveConstraint expr)
   at Tests.TestThrowsAfter() in <Assert.That line>
```

Dreamescaper:
Fixes #4281.

Previously, DelayedConstraint would execute the delegate by itself. However, underlying contraints might have additional logic for delegate handling - like ThrowsConstraint, which catches the exception.

Therefore, the fix is to pass the delegate to the constraint instead of executing it.

This fix has a change in the behavior. Currently, the test would fail if the delegate throws an exception. Now, since the delegate is invoked inside of try-catch block, it would be retried (for all constraints, not only Throws ones).
In my opinion, that's an expected behavior.
Still, let me know if you have any concerns.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
@@ -258,7 +258,6 @@ namespace NUnit.Framework.Constraints
             long now = Stopwatch.GetTimestamp();
             long delayEnd = TimestampOffset(now, DelayInterval.AsTimeSpan);
 
-            object? actual;
             if (PollingInterval.IsNotZero)
             {
                 long nextPoll = TimestampOffset(now, PollingInterval.AsTimeSpan);
@@ -268,11 +267,9 @@ namespace NUnit.Framework.Constraints
                         ThreadUtility.BlockingDelay((int)TimestampDiff(delayEnd < nextPoll ? delayEnd : nextPoll, now).TotalMilliseconds);
                     nextPoll = TimestampOffset(now, PollingInterval.AsTimeSpan);
 
-                    actual = InvokeDelegate(del);
-
                     try
                     {
-                        ConstraintResult result = BaseConstraint.ApplyTo(actual);
+                        ConstraintResult result = BaseConstraint.ApplyTo(del);
                         if (result.IsSuccess)
                             return new DelegatingConstraintResult(this, result);
                     }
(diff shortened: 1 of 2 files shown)


manfred-brands (approved):
@Dreamescaper Thanks. Looks good and a simple change.

stevenaw (approved):
Thanks for spotting this and contributing the fix @Dreamescaper !
I agree, the change in behaviour here is correcting an unwanted one (a bug) which is a good thing

PR #4820, opened 2024-09-07
WIP: Added an EqualStringConstraint

manfred-brands:
Contributes to #53, but instead of generic have more specific constraints for certain types.

See comment in #4456 

The idea is that with this we slowly can remove non-relevant properties (like IgnoreCase) from the `EqualConstraint` where they don't make sense as in `Assert.That(3.14, Is.EqualTo(2.72).IgnoreCase);`

The problem is with some  old constraint expected _untyped_ `IEnumerable` instead of `IEnumerable<string>`


manfred-brands:
Retracted in favour of #4847 and #4848

PR #4835, opened 2024-09-22
Added NUnit.Fw to the threads we're creating

### LINKED ISSUE #4832: Setting thread names to make debugging easier
Right now it's not easily distinguishable which threads are NUnit's, and specifically which are part of its MTA queues or STA queues.

image

OsirisTerje:
Fixes #4832

### DIFF
--- a/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
+++ b/src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
@@ -14,7 +14,7 @@ namespace NUnit.Framework.Internal.Execution
     {
         private static readonly Logger Log = InternalTrace.GetLogger("Dispatcher");
 
-        private const int WAIT_FOR_FORCED_TERMINATION = 5000;
+        private const int WaitForForcedTermination = 5000;
 
         private WorkItem? _topLevelWorkItem;
         private readonly Stack<WorkItem> _savedWorkItems = new();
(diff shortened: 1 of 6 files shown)


src/NUnitFramework/framework/Internal/Execution/ParallelWorkItemDispatcher.cs
```diff
@@ -65,17 +65,17 @@ private void InitializeShifts()
...
-            ParallelShift.Assign(new TestWorker(ParallelSTAQueue, "ParallelSTAWorker"));
+            ParallelShift.Assign(new TestWorker(ParallelSTAQueue, "NUnit.Fw.ParallelSTAWorker"));
 
-            var worker = new TestWorker(NonParallelQueue, "NonParallelWorker");
+            var worker = new TestWorker(NonParallelQueue, "NUnit.Fw.NonParallelWorker");
             worker.Busy += OnStartNonParallelWorkItem;
             NonParallelShift.Assign(worker);
 
-            worker = new TestWorker(NonParallelSTAQueue, "NonParallelSTAWorker");
+            worker = new TestWorker(NonParallelSTAQueue, "NUnit.Fw.NonParallelSTAWorker");
```
jnm2, line 78: What do you think about leaving the NUnit names for the workers the same, but prepending `NUnit.Fw` at the point of assigning `Thread.Name`?

OsirisTerje: Sure , that could also work. 
We're then assuming that TestWorker is not called from the outside, as it is a public method, even if it is in the Internal namespace.

jnm2: Just thinking, since `NUnit.Fw` was motivated by being able to distinguish threads, perhaps it "belongs" with Thread.Name and not in logs. But I don't feel strongly

OsirisTerje: Agree there.  I actually don't like that we're sending names of something around the way we do here and in some other places.  If the name should be close to where the thread are created, it should be in that parent class.  To me it looks like there should be different classes for each of these, and not a single class like now.
If we had a 
public class NonParallelTestWorker : TestWorker which handled the initialization, including name, this would resolve easily:
The current code is:
`var worker = new TestWorker(NonParallelQueue, "NUnit.Fw.NonParallelWorker");
worker.Busy += OnStartNonParallelWorkItem;
NonParallelShift.Assign(worker);`
could all be moved into NonParallelTestWorker .  Same with the OnStartNonParallelWorkItem, it only works on the TestWorker and the Workitem.

jnm2: I'd personally think of disassociating the class/log name from the thread name, since the thread names are more to look "pretty" for easy understanding when debugging. Though come to think of it, perhaps folks will want to correlate it with the log.

I'm good with your approach!

PR #4837, opened 2024-09-24
Compare records by properties when requested

### LINKED ISSUE #4836: PropertiesComparer doesn't work well with records
```csharp
record C1
{
    public string Name { get; set; }
    public int[] Collection { get; set; }
}

var v1 = new C1 { Name = "1", Collection = [1, 2, 3] };
var v2 = new C1 { Name = "1", Collection = [1, 2, 3] };

Assert.That(v1, Is.EqualTo(v2).UsingPropertiesComparer());
```

The assertion above fails, even thouht the same code works fine if I replace "record" with "class".
I assume that the issue is that record is IEquatable, therefore comparer doesn't attempt to compare it by properties.

I would suggest to ignore IEquatable for records in case `UsingPropertiesComparer` is used, and still compare it property by property.

Dreamescaper:
Fixes #4836 .

If CompareProperties is true, and actual value is record type, I ignore compiler-generated Equals methods, and force it to compare by properties instead.

In order to check that provided type is record, I verify that Equals method has CompilerGeneratedAttribute. Not sure if there's a better way.

I wanted to add a test for class with primary constructor, therefore I increased LangVersion to 12. Let me know if that causes any issues.

### DIFF
--- a/src/NUnitFramework/Directory.Packages.props
+++ b/src/NUnitFramework/Directory.Packages.props
@@ -30,7 +30,7 @@
     <PackageVersion Include="DotNetAnalyzers.DocumentationAnalyzers" Version="1.0.0-beta.59" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.CodeStyle" Version="4.10.0" />
     <PackageVersion Include="NUnit.Analyzers" Version="4.3.0" />
-    <PackageVersion Include="StyleCop.Analyzers" Version="1.2.0-beta.507" />
+    <PackageVersion Include="StyleCop.Analyzers" Version="1.2.0-beta.556" />
   </ItemGroup>
   <!-- Specific dependencies -->
   <ItemGroup>
--- a/src/NUnitFramework/framework/Internal/TypeHelper.cs
+++ b/src/NUnitFramework/framework/Internal/TypeHelper.cs
@@ -389,13 +389,10 @@ namespace NUnit.Framework.Internal
             return type.FullName ?? throw new InvalidOperationException("No name for type: " + type);
         }
 
-        private static readonly Type[] EqualsObjectParameterTypes = { typeof(object) };
-
-        internal static bool IsRecord(this Type type)
+        internal static bool HasCompilerGeneratedEquals(this Type type)
         {
-            // Check if Equals method has CompilerGenerated attribute
             var equalsMethod = type.GetMethod(nameof(type.Equals), BindingFlags.Instance | BindingFlags.Public,
-                                  null, EqualsObjectParameterTypes, null);
+                null, [type], null);
 
             return equalsMethod?.GetCustomAttribute<CompilerGeneratedAttribute>() is not null;
         }
--- a/src/NUnitFramework/framework/Internal/TypeHelper.cs
+++ b/src/NUnitFramework/framework/Internal/TypeHelper.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Text;
 using NUnit.Framework.Interfaces;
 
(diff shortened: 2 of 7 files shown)


src/NUnitFramework/framework/Internal/TypeHelper.cs
```diff
@@ -387,5 +388,16 @@ internal static string FullName(this Type type)
         {
             return type.FullName ?? throw new InvalidOperationException("No name for type: " + type);
         }
+
+        private static readonly Type[] EqualsObjectParameterTypes = { typeof(object) };
+
+        internal static bool IsRecord(this Type type)
```
manfred-brands, line 394: Suggest to rename to `HasCompilerGeneratedEquals`

src/NUnitFramework/tests/Assertions/AssertThatTests.cs
```diff
@@ -655,6 +664,8 @@ public void TestPropertyFailureSecondLevel()
              */
         }
 
+        private record Record(string Name, int[] Collection);
```
manfred-brands, line 667: I would like an example of a `record` where the user overrides `Equals`.
E.g. To use Case insensitive comparison of `Name`.
```csharp
private sealed record Person(string Name)
{
    public bool Equals(Person? other)
    {
        return string.Equals(Name, other?.Name, System.StringComparison.OrdinalIgnoreCase);
    }

    public override int GetHashCode()
    {
        return Name.GetHashCode(System.StringComparison.OrdinalIgnoreCase);
    }
}
```

src/NUnitFramework/framework/Internal/TypeHelper.cs
```diff
@@ -387,5 +388,16 @@ internal static string FullName(this Type type)
...
+        {
+            // Check if Equals method has CompilerGenerated attribute
+            var equalsMethod = type.GetMethod(nameof(type.Equals), BindingFlags.Instance | BindingFlags.Public,
```
manfred-brands, line 397: This tests for `Equals(object)`. It would be better to check for `Equals(T)`. See my `Person` record class. It defined `Equals(Person)` but not `Equals(object)` as the generated version is fine as it defers to `Equals(Person)`.

src/NUnitFramework/tests/Internal/TypeHelperTests.cs
```diff
@@ -84,5 +84,53 @@ private class Generic<T>
...
+                return obj is ClassWithOverriddenEquals other && other.Name == Name;
+            }
+
+            public override int GetHashCode()
+            {
+                return 539060726 + EqualityComparer<string?>.Default.GetHashCode(Name ?? string.Empty);
+            }
+        }
+
+        private record class RecordClass(string Name);
```
manfred-brands, line 125: Please add a record class overriding `Equals<T>`.

src/NUnitFramework/framework/Internal/TypeHelper.cs
manfred-brands: This is only relevant if the class has reference value properties (or an ImmutableArray).
Only for records where the default generated implementation would be wrong.

src/NUnitFramework/framework/Internal/TypeHelper.cs
manfred-brands: Or `HasNonValueCompilerGeneratedEquals`

manfred-brands (changes requested):
We should assume that the owner of the class knows best. If it explicitly implements `Equals(T)` it should be called regardless of properties comparer selection.

Also for 90% of record classes which only contain _value_ like properties the generated code is just fine.

src/NUnitFramework/Directory.Packages.props
```diff
@@ -30,7 +30,7 @@
     <PackageVersion Include="DotNetAnalyzers.DocumentationAnalyzers" Version="1.0.0-beta.59" />
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.CodeStyle" Version="4.10.0" />
     <PackageVersion Include="NUnit.Analyzers" Version="4.3.0" />
-    <PackageVersion Include="StyleCop.Analyzers" Version="1.2.0-beta.507" />
+    <PackageVersion Include="StyleCop.Analyzers" Version="1.2.0-beta.556" />
```
Dreamescaper, line 33: Older version complained about collection initializers

Dreamescaper:
I've updated the PR per your suggestions, thanks!

manfred-brands (approved):
Thanks @Dreamescaper, looks good now.

PR #4847, opened 2024-09-29
Added an EqualStringConstraint

### LINKED ISSUE #4875: String constraints should not allow using Within
Currently NUnit allows (as in compiles) statements like

```cs
Assert.That(somestring,Is.EqualTo("Don't panic").Within(42))  // reports NUnit2047 warning
```

The compiler should report this as an error.

manfred-brands:
@stevenaw This is for discussion, to show issues we face.

Contributes to #53 

I added a specific constraint for comparing strings (`EqualStringConstraint`)
The idea was that I then could remove the string specific modifiers (`IgnoreCase`, `IgnoreWhiteSpace`) from the actual `EqualConstraint`

This didn't work because the `EqualConstraint` modifiers, don't just apply to the type at hand, but to members of (nested) collection and dictionaries. Whether it applies is determined at runtime.

See the `CollectionEqualsTests`:
```csharp
[TestCaseSource(nameof(IgnoreCaseData))]
public void HonorsIgnoreCase(IEnumerable expected, IEnumerable actual)
{
    Assert.That(expected, Is.EqualTo(actual).IgnoreCase);
}

```

The only thing this `EqualsStringConstraint` brings us is that this constraint doesn't allow passing in an `Within` constraint.
So the one test for that:
```csharp
Assert.Throws<NotSupportedException>(() => Assert.That("abc", new EqualStringConstraint("abcd").Within(1)));
```

This no longer compiles. Hereby moving an error from runtime to compile time.

Fixes #4875

### DIFF
--- a/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
@@ -427,6 +427,14 @@ namespace NUnit.Framework.Constraints
             return Append(new EqualConstraint(expected));
         }
 
+        /// <summary>
+        /// Returns a constraint that tests two strings for equality
+        /// </summary>
+        public EqualStringConstraint EqualTo(string? expected)
+        {
+            return Append(new EqualStringConstraint(expected));
+        }
+
         #endregion
 
         #region SameAs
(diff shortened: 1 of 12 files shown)


OsirisTerje (approved):
lgtm

manfred-brands:

PR #4849, opened 2024-09-30
Micro-optimizations to name generation

stevenaw:
An ad-hoc PR.
Most of these changes are just updating to using `StringBuilder.Append(char)` where we can, but I was also able to update `TypeHelper.GetDisplayName(Type, object[])` to skip a branch in the loop since we know by that point in the method that there are at least 1 args.

### DIFF
--- a/src/NUnitFramework/framework/Internal/DisplayName.cs
+++ b/src/NUnitFramework/framework/Internal/DisplayName.cs
@@ -24,7 +24,7 @@ namespace NUnit.Framework.Internal
                 else
                 {
                     var builder = new StringBuilder();
-                    builder.Append("[");
+                    builder.Append('[');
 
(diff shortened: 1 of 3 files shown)


mikkelbu (approved):
LGTM. I guess there are tests covering `TypeHelper.GetDisplayName(Type, object[])` (I'm just reviewing in the browser, so a little hard to check)

stevenaw:
I can take another pass to review coverage too if you prefer? Let me know. 
We have "some" at least as they had failed for me at one point when I had accidentally deleted the line that adds the `,` between test parameters, but I'm not sure how comprehensive.
I'm always happy to take an opportunity to add missed coverage

manfred-brands (approved):
Thanks @stevenaw. Looks good to me.

Issue #3772, opened 2021-02-18
IsEquivalent on default ImmutableArray throws exception.


manfred-brands:
NUnit version 3.13.0

An ImmutableArray is a structure, but can contained a null array.
The following test fails with an InvalidOperationException: _This operation cannot be performed on a default instance of ImmutableArray<T>_

```
ImmutableArray<string> expected = ImmutableArray.Create("Test");
ImmutableArray<string> actual = default(ImmutableArray<string>);

Assert.That(actual, Is.EquivalentTo(expected));
```

Same for `Assert.That(actual, Is.Not.Empty);`

It is a pity that Microsoft doesn't treat this the same as `ImmutableArray<string>.Empty`.

mikkelbu:
It is a bit unfortunate that we need to have special cases for all these structures, but I think we should also handle this.

stevenaw:
Agreed, this is unfortunate.
Should `default(ImmutableArray<T>)` be treated as semantically equivalent to `Empty` or to `null`? NUnit makes a distinction between these for reference-type collections:

```csharp
string[] expected = Array.Empty<string>();
string[] actual = null;
Assert.That(actual, Is.Not.EqualTo(expected));
```

Though I would also find it unusual for the following test to pass:

```csharp
Assert.That(default(ImmutableArray<string>), Is.Null);
```

I wonder if this is case solved via a DefaultConstraint (https://github.com/nunit/nunit/issues/3457) ?

manfred-brands:
It looks like `Assert.That(default(ICollection), Is.Empty)` throws an `ArgumentException`. _The actual value must be a string, ..._
and `Assert.That(default(ICollection), Is.EquivalentTo(Array.Empty<string()))` throws with mesage _Expected IEnumerable But was: null_

So it looks like in all cases, one would need to check for `null` (or Default) first, so maybe this does not need changing in NUnit.
In my case that would mean inserting an `Assert.That(actualValue.IsDefault, Is.False)` or use the new DefaultConstraint when available.

jnm2:
I don't think NUnit should treat `default(ImmutableArray<T>)` as though it was `ImmutableArray<T>.Empty`. This will hide problems in my code.

stevenaw:
I'm unsure if there's still interest in this getting fixed, but I noticed that we solved a similar case in the `EqualConstraint` for `ArraySegment` in the past in #477

stevenaw:
[...]

PR #4850, opened 2024-09-30
Fix IsEquivalent on default ImmutableArray throws exception

mr-sergito:
Fixes the issue (#3772) when trying to compare equivalency of two `default(ImmutableArray)`s.

Added an early return at the `Matches` method so that any test which passes `EqualityComparer<T>.Default.Equals` also passes `Is.EquivalentTo()`.


Dreamescaper:
Could you please add a test for a case when two default ImmutableArrays are compared?

stevenaw (approved):
LGTM, thanks for your contribution @mr-sergito !
@jnm2 an oversight on my part during our discussion in the issue is how that the current lack of generic support in this constraint makes it harder to use `EqualityComparer<T>.Default` here.

PR #4853, opened 2024-10-06
Added EqualTimeBaseConstraint

### LINKED ISSUE #4877: Fix EqualTo modifiers for DateTime
The following below is currently allowed, we need to change this to be compiler errors.

So we allow: Assert.That(date1, Is.EqualTo(date2).Within(3).Seconds);
but not: Assert.That(date1, Is.EqualTo(date2).Seconds)
nor: Assert.That(date1, Is.EqualTo(date2).Within(timeSpan).Seconds)

Because the class with tolerance is different than the one without this also doesn't allow repeats like:
Assert.That(date1, Is.EqualTo(date2).Within(1).Hours.Seconds.Within(3))

manfred-brands:
Third PR working towards #53 and Related to #4847 and #4848

This constraint use classes (EqualTimeBasedConstraintWithNumericTolerance) to limit in what cases we can specify the `.Seconds` etc. modifiers.

So we allow: `Assert.That(date1, Is.EqualTo(date2).Within(3).Seconds);`
but not: `Assert.That(date1, Is.EqualTo(date2).Seconds)`
nor: `Assert.That(date1, Is.EqualTo(date2).Within(timeSpan).Seconds)`

Because the class with tolerance is different than the one without this also doesn't allow repeats like:
`Assert.That(date1, Is.EqualTo(date2).Within(1).Hours.Seconds.Within(3))`

Fixes #4877

### DIFF
--- a/src/NUnitFramework/framework/Constraints/ConstraintBuilder.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintBuilder.cs
@@ -176,6 +176,17 @@ namespace NUnit.Framework.Constraints
         }
 
         /// <summary>
+        /// Replaces the last pushed constraint with the specified constraint.
+        /// </summary>
+        /// <param name="constraint">The constraint to replace the lastPushed with.</param>
+        public void Replace(Constraint constraint)
+        {
+            _constraints.Pop();
+            _lastPushed = _ops.Top;
+            Append(constraint);
+        }
+
+        /// <summary>
         /// Sets the top operator right context.
         /// </summary>
         /// <param name="rightContext">The right context.</param>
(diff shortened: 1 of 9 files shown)


OsirisTerje (approved):
This makes sense to me.  AFAICS the Equal constraints have to separate for at least 1) each group of primitive types  (Numeric, Date/Time, bools,  char/strings), where the groups often have things in common  and then 2) when the types within the group differs in such a way it is better to seperate there too.  So to me this is on the right way.

OsirisTerje:
@manfred-brands All three PRs are approved.  Do you want to do more changes, or can they be merged ?

manfred-brands:
@OsirisTerje I think I extracted the common .Using in a base class, but cannot remember if all branches use that. I need to check/ensure that.

manfred-brands:

PR #4854, opened 2024-10-07
Add `IEnumeratable<KVP>` support for `ContainsKey(...).WithValue(...)`

### LINKED ISSUE #4852: Expand the usability of `Contains.Key(...).WithValue(...)`
I want to assert on HTTP response headers, but the `DictionaryContainsKeyValuePairConstraint` created by `Contains.Key(...).WithValue(...)` does not work on the type.

Currently the constraint checks for `IDictionary` and (despite the error message) also `IDictionary<TKey, TValue>`.

However, the `HttpHeaders` type implements `IEnumerable<KeyValuePair<string, IEnumerable<string>>>` (and of course the non-generic `IEnumerable` interface implicitly) and nothing else.

Would it be acceptable to expand the checks on this contraint to allow `IEnumerable<KeyValuePair<TKey, TValue>>` as well? I'm happy to work on it if the change would be accepted.

RenderMichael:
Fixes #4852

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using NUnit.Framework.Internal;
 
@@ -68,6 +69,33 @@ namespace NUnit.Framework.Constraints
                 }
 
                 return false;
+            }
+
+            // If 'actual' implements IEnumerable<KeyValuePair<TKey, TValue>>, we can find the value using that interface
+            var enumerableType = actual.GetType().GetInterface("IEnumerable`1");
+            if (enumerableType is not null)
+            {
+                var enumerableInnerType = enumerableType.GenericTypeArguments[0];
+                if (enumerableInnerType.IsGenericType && enumerableInnerType.GetGenericTypeDefinition() == typeof(KeyValuePair<,>))
+                {
+                    var keyGetter = enumerableInnerType.GetProperty("Key");
+                    var valueGetter = enumerableInnerType.GetProperty("Value");
+
+                    Debug.Assert(keyGetter is not null && valueGetter is not null);
+
+                    foreach (object item in (IEnumerable)actual)
+                    {
+                        var expectedKvp = new KeyValuePair<object, object?>(_expected.Key, _expected.Value);
+                        var actualKvp = new KeyValuePair<object, object?>(keyGetter.GetValue(item)!, valueGetter.GetValue(item));
+
+                        if (ItemsEqual(actualKvp, expectedKvp))
+                        {
+                            return true;
+                        }
+                    }
+
+                    return false;
+                }
             }
 
             throw new ArgumentException($"Expected: IDictionary But was: {actual.GetType()}", nameof(actual));
--- a/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs
@@ -58,13 +58,12 @@ namespace NUnit.Framework.Constraints
             if (ImplementsIEnumerableKeyValuePair(actual))
             {
                 var expected = new KeyValuePair<object, object?>(_expected.Key, _expected.Value);
+                var enumerable = (IEnumerable)actual;
 
-                foreach (object item in (IEnumerable)actual)
+                foreach (var item in enumerable)
                 {
                     if (ItemsEqual(item, expected))
-                    {
                         return true;
-                    }
                 }
 
                 return false;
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs
```diff
@@ -70,6 +71,33 @@ private bool Matches(object? actual)
                 return false;
             }
 
+            // If 'actual' implements IEnumerable<KeyValuePair<TKey, TValue>>, we can find the value using that interface
+            var enumerableType = actual.GetType().GetInterface("IEnumerable`1");
```
manfred-brands, line 75: Suggest to cache the result of `actual.GetType().GetInterfaces()` from line 60.

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs
```diff
@@ -70,6 +71,33 @@ private bool Matches(object? actual)
...
+
+                    foreach (object item in (IEnumerable)actual)
+                    {
+                        var expectedKvp = new KeyValuePair<object, object?>(_expected.Key, _expected.Value);
+                        var actualKvp = new KeyValuePair<object, object?>(keyGetter.GetValue(item)!, valueGetter.GetValue(item));
+
+                        if (ItemsEqual(actualKvp, expectedKvp))
+                        {
+                            return true;
+                        }
```
manfred-brands, line 94: Is any of this needed? The code in the existing code above, even though it checks for `IDictionary` could as well only check for `IEnumerable<KeyValuePair>` as that is all that it uses.

manfred-brands (changes requested):
The test that fails (`DictionaryContainsKeyValuePairConstraintTests.FailsWhenNotUsedAgainstADictionary`) is one that expects the code to fail if it isn't an `IDictionary`.
The test uses an List<KeyValuePair<,>>` which is the support you added.

Rewrite that test to pass or delete it as you already made a new test.

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs
manfred-brands: Not relevant if the original test on line 60 is changed to test for `IEnumerable<KeyValuePair<,>>` instead.

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs
RenderMichael: That's a good point. I changed the entire constraint to remove any dictionary-ness and instead make it work with any "key-value pair provider".

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs
```diff
@@ -41,7 +40,7 @@ public DictionaryContainsKeyValuePairConstraint(object key, object? value)
         private bool Matches(object? actual)
         {
             if (actual is null)
-                throw new ArgumentException("Expected: IDictionary But was: null", nameof(actual));
+                throw new ArgumentException("Expected: IEnumerable<KeyValuePair<,>> But was: null", nameof(actual));
```
manfred-brands, line 43: It also works for just `IDictionary`. 
So suggest to make it:
```suggestion
                throw new ArgumentException("Expected: IDictionary or IEnumerable<KeyValuePair<,>> But was: null", nameof(actual));
```

src/NUnitFramework/framework/Constraints/DictionaryContainsKeyValuePairConstraint.cs
```diff
@@ -70,7 +69,24 @@ private bool Matches(object? actual)
                 return false;
             }
 
-            throw new ArgumentException($"Expected: IDictionary But was: {actual.GetType()}", nameof(actual));
+            throw new ArgumentException($"Expected: IEnumerable<KeyValuePair<,>> But was: {actual.GetType()}", nameof(actual));
```
manfred-brands, line 72: Same here.
```suggestion
            throw new ArgumentException($"Expected type implementing IDictionary or IEnumerable<KeyValuePair<,>> But was: {actual.GetType()}", nameof(actual));
```

src/NUnitFramework/tests/Constraints/DictionaryContainsKeyValueConstraintTests.cs
```diff
@@ -222,7 +221,24 @@ public void FailsWhenNullDictionary()
 
             TestDelegate act = () => Assert.That(dictionary, new DictionaryContainsKeyValuePairConstraint("1", "World"));
 
-            Assert.That(act, Throws.ArgumentException.With.Message.Contains("Expected: IDictionary But was: null"));
+            Assert.That(act, Throws.ArgumentException.With.Message.Contains("Expected: IEnumerable<KeyValuePair<,>> But was: null"));
```
manfred-brands, line 224: Adjust to changes suggest above.

manfred-brands (changes requested):
@RenderMichael Most of it looks good now.
Only item I have is the dropping of `IDictionary` in the exceptions.
Please put that back in.

RenderMichael:
The error message has been fixed to refer to non-generic `IDictionary` again.

manfred-brands (approved):
@RenderMichael Thanks for your contribution. Looks good now.

PR #4863, opened 2024-10-26
chore: Attempt to reenable deterministic builds

mikkelbu:


stevenaw:
@mikkelbu Were you still working on this? The change you've made seem to work based on my own local testing using [NuGetPackageExplorer](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer).

mikkelbu:
@stevenaw I was working on this, but then I discovered that it seems like our CI builds are already deterministic (and then I forgot all about it).

So if I take the artifacts from e.g. https://github.com/nunit/nunit/actions/runs/12476664776/job/34821787291- (latest build on master) - https://github.com/nunit/nunit/actions/runs/12476664776/artifacts/2358138821, then `NUnit.4.4.0-alpha.0.1.nupkg` is `deterministic` according to `NuGet Package Explorer`, but if I download `nunit.4.3.2.nupkg` from https://www.nuget.org/packages/nunit, then it is not marked as deterministic and if I read #4911 correctly, then perhaps the problem is that the files we are uploading are a build done locally?

https://github.com/nunit/docs/blob/master/docs/articles/developer-info/Packaging-the-Framework.md says (but I'm unsure about the process)
> The release should not be built on a developers machine, it should be built by the build servers. The following steps are only for reproducing the steps locally.

@OsirisTerje can you confirm how we currently do releases?

Ps. The files on myget are even worse, but perhaps this is due to myget?
image

OsirisTerje:
@mikkelbu  @stevenaw  When we currently do a release now, we build it locally and upload that to nuget.  So that is the reason it is not marked as deterministic.  
The pre-release builds however, are built by the Publish action.  This was done quite recently.  And they should be deterministic, as you see from the 

Next up is the ordinary nuget release publishing.   

I am bit surprised however, that the Myget publish is not deterministic, but the CI build is.  They do things exactly the same way, as far as I can see.  I might be missing something ......

OsirisTerje:
Since version 4.4.0 we have deterministic builds.

PR #4870, opened 2024-11-01
4862: Remove Property Limit From PropertiesComparer

### LINKED ISSUE #4862: PropertiesComparer 32 Property Limit Error Message
When using `.UsingPropertiesComparer()` on a class with 32 or more properties the following error message is returned
```
System.NotSupportedException : No comparer found for instances of type 'x' and 'x'
```
(x is the same class in this case, just as an example)

This message doesn't actually tell the test writer what's going on as unless you look at the `PropertiesComparer.cs` in this repo there's no documentation on a max property limit. At least as far as I could find on [this page ](https://docs.nunit.org/articles/nunit/writing-tests/constraints/EqualConstraint.html)

As can be seen in [the file](https://github.com/nunit/nunit/blob/3d6b9e3b824da20e07e351be3a476fb2c31125d4/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs) this check could return it's own enum result for a better error message. 

Also, I'm wondering why there is a 32 property max (actually only 31 because a `>=` is used instead of a `>`)?
I see in that code there's an `int` whose 31 of 32 bits are being used as flags, but would that also work as a `long` so that a max of 64 properties is available instead? A list could also be used instead of bitwise operations so that there is no property limit

Cincidial:
Changes `PropertiesComparer` to use a [BitArray](https://learn.microsoft.com/en-us/dotnet/api/system.collections.bitarray?view=net-8.0) instead of a uint (only 32 bits).

Note that in the docs for `BitArray` above there are 2 methods which are only available from .NET 8 and onwards. As such they can't be used here
- [HasAnySet](https://learn.microsoft.com/en-us/dotnet/api/system.collections.bitarray.hasanyset?view=net-8.0#system-collections-bitarray-hasanyset)
- [HasAllSet](https://learn.microsoft.com/en-us/dotnet/api/system.collections.bitarray.hasallset?view=net-8.0#system-collections-bitarray-hasallset)

Instead of re-implementing these methods for use in this case, I opted to use a counter to count the properties that could not use the tolerance. If re-implementing these would be preferred let me know and I can add them in as extension methods?

Tests passed locally for Windows. I don't have another OS to test on.

Fixes #4862

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System;
+using System.Collections;
 using System.Linq;
 using System.Reflection;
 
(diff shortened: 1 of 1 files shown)


manfred-brands (approved):
@Cincidial Thanks for you contribution.

The changes look nice and simple.
I don't think it makes sense to make extension methods for the two missing methods.
These methods would have to iterate over all bits twice (All and Any) when it is simpler to maintain the count.

mikkelbu (approved):

PR #4872, opened 2024-11-02
Add support for Generic Nullable parameters

### LINKED ISSUE #4868: TestCase- and TestCaseSourceAttribute cannot handle generic test methods where method parameter is Nullable<T>
When using the TestCaseAttribute or TestCaseSourceAttribute on a test-method, the test-method is allowed to be generic, as long as the testcase-arguments provided for each testcase can be used by NUnit to resolve all the open generic type parameters of the test-method to generate and run a new test.

However, NUnit fails to resolve the generic type parameter **TValue** when that parameter is constrained to be a **struct** and the actual method parameter is declared as a **Nullable<TValue>** and the provided testcase-arguments are either of type **TValue** or **Nullable<TValue>**. For example, given the following test-method, NUnit will fail to run the method with the message '_Unable to determine type arguments for method_':

```
[TestCase(0)]
public void SomeTest<TValue>(TValue? value) where TValue : struct
{
    ...
}
```

The expected behavior is, of course, that NUnit will be able to resolve the desired type of **TValue** to **Int32** and then pass the test-case argument (0) as a value for the method parameter.

This bug was discovered using NUnit 4.1.0 but is still present in NUnit 4.2.2.

manfred-brands:
Fixes #4868 

The PR fixes this:
```csharp
[TestCase(0)]
[TestCase(1L)]
[TestCase(2UL)]
[TestCase(3F)]
[TestCase(4D)]
public void GenericNullable<TValue>(TValue? value)
    where TValue : struct, IConvertible
```

With the generic type determined correctly:

{0AFD890A-51E3-4C21-AE05-53C80F2A4F14}

### DIFF
--- a/src/NUnitFramework/framework/Internal/GenericMethodHelper.cs
+++ b/src/NUnitFramework/framework/Internal/GenericMethodHelper.cs
@@ -108,6 +108,10 @@ namespace NUnit.Framework.Internal
                             TryApplyArgType(genericArgTypes[i], argTypes[i]);
                     }
                 }
+                else if (Reflect.IsNullable(parmType) && !argType.IsClass)
+                {
+                    ApplyArgType(genericArgTypes[0], argType);
+                }
             }
         }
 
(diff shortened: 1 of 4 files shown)


OsirisTerje (approved):
lgtm

PR #4873, opened 2024-11-02
Add test coverage for Value(Source) attributes for generic methods

### LINKED ISSUE #150: Improve inference of type arguments in generic test methods
Currently the deduction of type arguments in NUnit is somewhat limited.

For example the following will work:

``` C#
public static ITestCaseData[] TestCasesForEqualitiy =
{
    new TestCaseData(1, 1).Returns(true), 
    new TestCaseData(1, 2).Returns(false), 
};

[Test]
[TestCaseSource("TestCasesForEqualitiy")]
public bool Equality<TElement>(TElement left, TElement right)
{
    return left.Equals(right);
}
```

While the following won't work:

``` C#
public static ITestCaseData[] TestCasesForEqualitiy =
{
    new TestCaseData(1, 1).Returns(true), 
    new TestCaseData(1, 2).Returns(false), 
};

[Test]
[TestCaseSource("TestCasesForEqualitiy")]
public bool Equality<TElement>(TElement? left, TElement? right)
{
    return left.Equals(right);
}
```

As far as I can see the modification would be local to `NUnit.Framework.Internal.Builders.NUnitTestCaseBuilder.GetTypeArgumentsForMethod(...)`.

I have coded a somewhat complex type inference logic for another project. I'll check into it and will happily supply it to NUnit project.

### LINKED ISSUE #4868: TestCase- and TestCaseSourceAttribute cannot handle generic test methods where method parameter is Nullable<T>

stevenaw:
Relates to https://github.com/nunit/nunit/issues/150 and https://github.com/nunit/nunit/issues/4868

@manfred-brands PR which fixed #4868 also fixed #150. This PR just adds some test coverage as #150 also expresses a desire for consistency between these attributes which provide values to generic methods. For this reason I'm only saying this PR "relates" and will update @manfred-brands 's PR to include a "fixes" annotation there.

This PR also adds a test for existing ValueTuple equality that I haven't merged in through other means yet.

### DIFF
--- a/src/NUnitFramework/tests/Attributes/ValueSourceTests.cs
+++ b/src/NUnitFramework/tests/Attributes/ValueSourceTests.cs
@@ -279,6 +279,21 @@ namespace NUnit.Framework.Tests.Attributes
                 Assert.That(suite.Tests[4].Name, Is.EqualTo(@"MethodWithArrayArguments([System.Byte[,]])"));
             });
         }
+
+        [Test]
+        public void GenericNullableTest<TValue>(
+            [ValueSource(nameof(GenericNullableSource))] TValue? value)
+            where TValue : struct, IConvertible
+        {
+            Assert.That(value, Is.Not.Null);
+            int index = value.Value.ToInt32(null);
+#pragma warning disable NUnit2021 // Incompatible types for EqualTo constraint
+            Assert.That(value.Value, Is.EqualTo(index));
+#pragma warning restore NUnit2021 // Incompatible types for EqualTo constraint
+            Assert.That(value.Value, Is.InstanceOf(GenericNullableSource[index].GetType()));
+        }
+
+        private static readonly object[] GenericNullableSource = [0, 1L, 2UL, 3F, 4D];
     }
 
     public class ValueSourceMayBeInherited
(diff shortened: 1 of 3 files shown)


OsirisTerje (approved):
lgtm

mikkelbu (approved):

PR #4881, opened 2024-11-08
Fixes #4876 - Tests with negative tolerance always fail without proper message

### LINKED ISSUE #4876: Equal.Within Fails With Negative Tolerance
## Summary

If asserting the equality of two floats within a tolerance (`Is.EqualTo.Within`), a negative tolerance leads to the test always failing. No warning is being reported, instead the test simply fails with the default error message. From the error message it is unclear that the reason for the test failing is the negative tolerance.

This can be an issue in the case of the tolerance being calculated, if the resulting tolerance can be negative.

## Example

Versions: .NET 8, NUnit 4.3.0 (although was also observed in 3.9).

Given a test which compares the equality of two floats with a negative tolerance as follows:

```
[Test]
public void NegativeToleranceTest()
{
    Assert.That(2f, Is.EqualTo(2f).Within(-0.1f), "Negative Tolerance");
}
```

This test, when being executed, reports the following message:

>  Negative Tolerance
>   Expected: 2.0f +/- -0.100000001f
>   But was:  2.0f
>   Off by:   0.0d

From a mathematical standpoint this message does not make sense, as `2 +/- -0.1 == 2 -/+ 0.1`. This makes it unnecessarily hard to find this issue.

## Proposed Solution

Introduce a warning (preferrably compile-time) for negative tolerances, at least for built-in number formats.

RaffaelCH:
Fixes issue #4876 

Currently executing a test with a negative Tolerance (e.g., `Is.Equal.Within(-1)`) leads to the test not passing without any errors. With these changes an ArgumentException is thrown at runtime instead.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Tolerance.cs
+++ b/src/NUnitFramework/framework/Constraints/Tolerance.cs
@@ -19,6 +19,7 @@ namespace NUnit.Framework.Constraints
         private const string ModeMustFollowTolerance = "Tolerance amount must be specified before setting mode";
         private const string MultipleToleranceModes = "Tried to use multiple tolerance modes at the same time";
         private const string NumericToleranceRequired = "A numeric tolerance is required";
+        private const string NoNegativeTolerance = "Tolerance amount must not be negative";
 
         /// <summary>
         /// Returns a default Tolerance object, equivalent to a default matching rules.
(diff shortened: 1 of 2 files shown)


RaffaelCH:
@dotnet-policy-service agree

OsirisTerje (approved):
LGTM

stevenaw (approved):
LGTM as well. Thanks for your contribution @RaffaelCH !

PR #4883, opened 2024-11-11
Async Test continuations now also run on the same thread when SIngleThreaded is used.

### LINKED ISSUE #4110: Support running single-threaded async tests on Linux
Currently the [Apartment(ApartmentState.STA)] annotation is not supported on Linux, which means that the SingleThreadedSynchronizationContext cannot be enabled on Linux and tests using the annotation are skipped with "Apartment state cannot be set on this platform.". There is as far as I can see no other way to enable the nunit's SingleThreadedSynchronizationContext. It also seems that MessagePumpStrategy only has hardcoded strategies for the default SynchronizationContextes and does not support a custom SynchronizationContext, making use of a custom single threaded SynchronizationContextes impossible.

manfred-brands:
Fixes #4110 

If the `TestContext.IsSingleThreaded` is set using the `[SingleThreaded]` attribute and an Async operation is executed a SingleThreadedSynchronizationContext is created.

### DIFF
--- a/src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
+++ b/src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
@@ -22,13 +22,19 @@ namespace NUnit.Framework.Internal
         }
 
         public static object? Await(Func<object?> invoke)
-            => Await<object?>(invoke);
+            => Await<object?>(default(TestExecutionContext), invoke);
 
         public static TResult? Await<TResult>(Func<object?> invoke)
+            => Await<TResult>(default(TestExecutionContext), invoke);
+
+        public static object? Await(TestExecutionContext? context, Func<object?> invoke)
+            => Await<object?>(context, invoke);
+
+        public static TResult? Await<TResult>(TestExecutionContext? context, Func<object?> invoke)
         {
             Guard.ArgumentNotNull(invoke, nameof(invoke));
 
-            using (InitializeExecutionEnvironment())
+            using (InitializeExecutionEnvironment(context))
             {
                 var awaiter = AwaitAdapter.FromAwaitable(invoke.Invoke());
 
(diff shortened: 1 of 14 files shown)


stevenaw (approved):
LGTM thanks @manfred-brands !
I'll request @jnm2 here for review too if they're available.

src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
```diff
@@ -22,13 +22,19 @@ public static bool IsAsyncOperation(Delegate @delegate)
         }
 
         public static object? Await(Func<object?> invoke)
-            => Await<object?>(invoke);
+            => Await<object?>(default(TestExecutionContext), invoke);
 
         public static TResult? Await<TResult>(Func<object?> invoke)
+            => Await<TResult>(default(TestExecutionContext), invoke);
```
jnm2, line 28: (optional) So as not to imply that TestExecutionContext is a struct:
```suggestion
            => Await<TResult>((TestExecutionContext)null, invoke);
```

src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
```diff
@@ -41,10 +47,10 @@ public static bool IsAsyncOperation(Delegate @delegate)
                 return (TResult?)awaiter.GetResult();
             }
         }
-
-        private static IDisposable? InitializeExecutionEnvironment()
+        private static IDisposable? InitializeExecutionEnvironment(TestExecutionContext? executionContext)
```
jnm2, line 50: Could this parameter be required? If not provided, the danger is that we may forget to pass it. Passing `context: null` explicitly in code helps raise awareness that we're omitting context.

jnm2 (approved):

src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
manfred-brands: I didn't want to many changes in existing code. I have looked at all usages and the remaining didn't seem to need it.
But I agree that the suggested code makes it clearer then using the overload.
When I'm back near a PC next week, I will remove the overload and change all calls.

src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
manfred-brands: Removed overload, so no longer relevant.

I don't associate `default(T)` with struct, just a different (nicer?) way to write `(T)null`

src/NUnitFramework/framework/Internal/AsyncToSyncAdapter.cs
manfred-brands: Changes made.

manfred-brands:
@OsirisTerje Are you ok for me to merge?

Issue #4857, opened 2024-10-13
Drop .NET 6 as a build target of the framework


stevenaw:
I'm creating this to start a discussion over timing and planning.

With .NET 9 about a month from release now, a new version of the SDK is also on the horizon. I _think_ they support building "current-3" TFMs though I am unable to track that down at the moment. This would mean that the upcoming release of the SDK could still build nunit if it targets .NET 6 as a TFM, however the following one to be released in Nov 2025 likely could not.  .NET 6 itself will hit EOL in Nov 2024.

We may want to discuss and consider a transition plan to releasing the framework against a higher TFM

manfred-brands:
@stevenaw I have a branch targeting .net 8.0 as part of the timeout issue which I still need resolving on Linux.

Adding an extra target increases the size of the nuget package.
If that target utilizes extra functionality, then that should be ok.
If there is no difference between the two targets we might as well only keep the lowest as it would support any higher frameworks. Caveat obsoleted apis such as binary serialization.

At work we use the .NET 8.0.4xx SDK to multi-target .NET 8.0 and .NET 4.8 Framework.
Newer SDKs introduced 2 things, language features and new core libraries.
Sometimes the new language features are just syntax and available on all targets, e.g. collection initializers, sometimes not e.g. static interface members
If targeting older frameworks it will use still use the older libraries.

We need to way up if the new features are worth the complexity of conditional code.
At work I have a "compatibility" library that consist of extension methods and some classes so that I can use some of the advantages of later targets without conditional code. Older targets will call into that library, newer ones directly into their libraries. The downside is that I cannot do this for neither properties nor static methods. The feature to allow that was shown for C#13 put was pulled from .NET9.

Having said that, I think our support statement is to support SDKs 6 months after MS although I can't seem to find that on the docs site. If correct we need .NET 6.0 until at least May 2025.

stevenaw:
Quite right! And I should clarify that the EOL date I mentioned above was Microsoft's, meaning NUnit's would be around May 2025. In either event, a transition process could be good to discuss as _eventually_ the .NET6 TFM will fail to build at some point in the future. Even if we happen to all be thinking the same transition plan, I figure it's good to confirm. On my end, my thinking is:

1. We could add a new TFM soon, likely NET8, perhaps as early as 4.3. This allows us to start taking advantage of newer features in the codebase sooner. There would be a short time where we may be targeting both NET6 + NET8.
[...]

PR #4884, opened 2024-11-13
Add a .NET 8.0 target

### LINKED ISSUE #4890: Add .net 8 as a target framework in the package
See #4857 

Add the .NET 8.0 target. So that if users that are not allowed to use end-of-life frameworks (e.g. .net 6) due to lack of security updates can use the .NET8.0 binaries.

manfred-brands:
Fixes #4890 
Works toward #4857

This adds the .NET 8.0 target. So that if users that are not allowed to use end-of-life frameworks due to lack of security updates can the .NET8.0 binaries.

### DIFF
--- a/src/NUnitFramework/framework/Internal/InvalidDataSourceException.cs
+++ b/src/NUnitFramework/framework/Internal/InvalidDataSourceException.cs
@@ -3,7 +3,9 @@
 namespace NUnit.Framework.Internal
 {
     using System;
+#if !NET8_0_OR_GREATER
     using System.Runtime.Serialization;
+#endif
 
     /// <summary>
     /// InvalidTestFixtureException is thrown when an appropriate test
@@ -36,6 +38,7 @@ namespace NUnit.Framework.Internal
         {
         }
 
+#if !NET8_0_OR_GREATER
         /// <summary>
         /// Serialization Constructor
         /// </summary>
@@ -43,5 +46,6 @@ namespace NUnit.Framework.Internal
             StreamingContext context) : base(info, context)
         {
         }
+#endif
     }
 }
--- a/src/NUnitFramework/nunitlite/Options.cs
+++ b/src/NUnitFramework/nunitlite/Options.cs
@@ -134,8 +134,12 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.IO;
+#if !NET8_0_OR_GREATER
 using System.Runtime.Serialization;
+#if !NET6_0_OR_GREATER
 using System.Security.Permissions;
+#endif
+#endif
 using System.Text;
 using System.Text.RegularExpressions;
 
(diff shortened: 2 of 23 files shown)


src/NUnitFramework/framework/Internal/InvalidDataSourceException.cs
```diff
@@ -3,7 +3,9 @@
 namespace NUnit.Framework.Internal
 {
     using System;
+#if !NET8_0_OR_GREATER
     using System.Runtime.Serialization;
+#endif
```
mikkelbu, line 8: Just suggestions. Should we move the usings outside the namespace? (and perhaps move `System.Runtime.Serialization`  down to the sole purpose, so we avoid having two `!NET8_0_OR_GREATER` blocks ?

EDIT: I can see that we do it differently in many of the `...Exception` files, so I'm also happy with the current solution (to keep the fix as small as possible)

src/NUnitFramework/nunitlite/Options.cs
```diff
@@ -134,8 +134,12 @@
 using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.IO;
+#if !NET8_0_OR_GREATER
 using System.Runtime.Serialization;
+#if !NET6_0_OR_GREATER
 using System.Security.Permissions;
+#endif
```
mikkelbu, line 141: Why is this need now ?

src/NUnitFramework/nunitlite/Options.cs
```diff
@@ -569,14 +573,17 @@ public OptionException(string message, string optionName, Exception innerExcepti
...
             : base(info, context)
         {
             _option = info.GetString("OptionName");
         }
+#endif
 
         public string OptionName => _option;
 
+#if !NET8_0_OR_GREATER
 #if !NET6_0_OR_GREATER
```
mikkelbu, line 587: Doesn't `#if !NET6_0_OR_GREATER` imply `#if !NET8_0_OR_GREATER` ? 
Or is this to to anticipate when we remove the .NET6 target ?

mikkelbu (approved):
Looks good - only comments and questions from me 👍

OsirisTerje (changes requested):
We need to add the .net 8 bits to the nunit.nuspec file, otherwise the bits don't go into the nupkg.

manfred-brands:
> We need to add the .net 8 bits to the nunit.nuspec file, otherwise the bits don't go into the nupkg.

Thanks @OsirisTerje 

I forgot and to ensure nobody will forget it again, I added a nunit test to verify proper file references in .nuspec files.
Similar to the one you added for dependencies.

The bulk of the size is the 3 copies of the .xml files 1.2MB per target.
They are 99% the same with a few TFM specific APIs.
I tired to put only one of the in lib/any but it then doesn't seem to be picked up.
Searching the net didn't result in any solutions.

OsirisTerje (approved):

OsirisTerje:
Lgtm,  @manfred-brands : can we merge ?

PR #4885, opened 2024-11-16
Ony Unwrap TargetInvocationException if raised from inside NUnit.Framework

### LINKED ISSUE #4718: Stacktrace calls lost on exception of type TargetInvocationException
Environment:
Runner: VS Test Adapter v4.5.0
Target Framework: .net8.0

There seems to be a difference in how NUnit 3.14 behaves compared to NUnit 4.1 regarding the exception `System.Reflection.TargetInvocationException`. In NUnit 4.1 only the inner exception gets logged and thus we lose the stacktrace of the outer exception resulting in the loss of valuable information.

Given the following simple example below I will provide the output for each version.

```csharp
namespace TestTargetInvocationException
{
    public class Tests
    {
        [Test]
        public void Test()
        {
            Foo();
        }

        public static void Foo()
        {
            typeof(Tests).GetMethod(nameof(Bar))?.Invoke(null, null);
        }

        public static void Bar()
        {
        }
    }
}
```

NUnit 3.14 output:
```
  Message: 
System.Reflection.TargetInvocationException : Exception has been thrown by the target of an invocation.
  ----> System.NotImplementedException : The method or operation is not implemented.

  Stack Trace: 
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
MethodBase.Invoke(Object obj, Object[] parameters)
Tests.Foo() line 13
Tests.Test() line 8
RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
--NotImplementedException
Tests.Bar() line 18
RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

NUnit 4.1 output:
```
  Message: 
System.NotImplementedException : The method or operation is not implemented.

  Stack Trace: 
Tests.Bar() line 18
RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

While the NUnit 4.1 output does look cleaner it misses the lines that lead to that exception being thrown.

manfred-brands:
Fixes #4718

### DIFF
--- a/src/NUnitFramework/framework/Internal/ExceptionHelper.cs
+++ b/src/NUnitFramework/framework/Internal/ExceptionHelper.cs
@@ -225,7 +225,8 @@ namespace NUnit.Framework.Internal
             }
 
             while (exception is TargetInvocationException targetInvocationException &&
-                targetInvocationException.InnerException is not null)
+                targetInvocationException.InnerException is not null &&
+                targetInvocationException.StackTrace?.Contains("NUnit.Framework") is true)
             {
                 exception = targetInvocationException.InnerException;
             }
--- a/src/NUnitFramework/testdata/UnexpectedExceptionFixture.cs
+++ b/src/NUnitFramework/testdata/UnexpectedExceptionFixture.cs
@@ -9,6 +9,12 @@ namespace NUnit.TestData.UnexpectedExceptionFixture
     public class UnexpectedExceptionFixture
     {
         [Test]
+        public void ThrowsException()
+        {
+            throw new Exception("Thrown Exception");
+        }
+
+        [Test]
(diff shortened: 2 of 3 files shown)


OsirisTerje (approved):
lgtm

PR #4888, opened 2024-11-23
Refactor TestAssert: Replace ToArray()[0] with First() for efficiency.

### LINKED ISSUE #4886: Optimize ToArray()[0] with First() in TestAssert class in TestAssert.cs
In the TestAssert class, specifically in methods like IsRunnable and IsNotRunnable, the code currently uses ToArray()[0] to access the first element in a collection (result.Children). This approach creates an array unnecessarily, when the First() LINQ method can be used instead to directly get the first element without the added overhead of creating an array.

**Current Code:**
```
if (result.HasChildren)
    result = result.Children.ToArray()[0];
```

**Proposed Improvement:**
Replace the ToArray()[0] with the more efficient First() method from LINQ to avoid unnecessary array allocation.
```
if (result.HasChildren)
    result = result.Children.First();
```

Saurav-shres:
This pull request refactors the TestAssert class, specifically the IsRunnable(Type type, string name, ResultState resultState) method. The change replaces the use of ToArray()[0] with the more efficient First() method from LINQ. This improves the performance by eliminating unnecessary array allocation and directly accessing the first element in the collection.

Fixes #4886

### DIFF
--- a/src/NUnitFramework/tests/TestUtilities/TestAssert.cs
+++ b/src/NUnitFramework/tests/TestUtilities/TestAssert.cs
@@ -36,7 +36,7 @@ namespace NUnit.Framework.Tests.TestUtilities
             object testObject = Reflect.Construct(type);
             ITestResult result = TestBuilder.RunTest(test, testObject);
             if (result.HasChildren) // In case it's a parameterized method
-                result = result.Children.ToArray()[0];
+                result = result.Children.First();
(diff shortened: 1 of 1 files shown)


Saurav-shres:
@dotnet-policy-service agree

stevenaw (approved):
Looks great to me, and all the tests are passing. Thanks for spotting this and for your contribution to fixing it @Saurav-shres

Issue #4806, opened 2024-09-01
Benchmarks fail to run


MaxKot:
*Bug*

Steps to reproduce:

1. Run nunit.framework.benchmarks in Release configuration.
2. Select ByteAllocationBenchmark, CompositeWorkItemBenchmark or StreamsComparerBenchmark

*Expected result*

Benchmarks running

*Actual result*

Benchmark error `CSC : error CS7027: Error signing output with public key from file '..\..\nunit.snk' -- File not found.`

*Additional steps*

3. Set `AssemblyOriginatorKeyFile` value relative to `$(MSBuildThisFileDirectory)` in src\NUnitFramework\Directory.Build.props and src\NUnitFramework\benchmarks\nunit.framework.benchmarks\nunit.framework.benchmarks.csproj

*Result after additional steps*

Benchmark error `\src\NUnitFramework\Directory.Build.props(72,5): error : Missing AssemblyConfiguration for net6.0`

*Addtional notes*

It appears that `TargetFramework` is empty when BenchmarkDotNet builds boilerplate code. Setting hard-coded `AssemblyConfiguration` value allows benchmarks to run without issue.

stevenaw:
Thanks for the diagnosis and report here @MaxKot !

stevenaw:
@OsirisTerje 
 We have a defined `4.4` milestone however I realize we may want a `4.3.1`. I'll make that now and assign this to it.

PR #4897, opened 2024-12-15
Get the benchmarks running again

stevenaw:
Fixes https://github.com/nunit/nunit/issues/4806

This PR gets the benchmarks running again by following the steps outlined in the issue. I was able to inject the assemblyconfiguration at runtime by defining a BenchmarkDotNet Job based on the default one

### DIFF
--- a/src/NUnitFramework/benchmarks/nunit.framework.benchmarks/nunit.framework.benchmarks.csproj
+++ b/src/NUnitFramework/benchmarks/nunit.framework.benchmarks/nunit.framework.benchmarks.csproj
@@ -6,9 +6,14 @@
         <ImplicitUsings>enable</ImplicitUsings>
         <LangVersion>latest</LangVersion>
         <Optimize>true</Optimize>
-        <AssemblyOriginatorKeyFile>..\..\..\nunit.snk</AssemblyOriginatorKeyFile>
         <RootNamespace>NUnit.Framework</RootNamespace>
     </PropertyGroup>
+
+    <ItemGroup>
+      <Compile Remove="BenchmarkDotNet.Artifacts\**" />
+      <EmbeddedResource Remove="BenchmarkDotNet.Artifacts\**" />
+      <None Remove="BenchmarkDotNet.Artifacts\**" />
+    </ItemGroup>
 
     <ItemGroup>
       <PackageReference Include="BenchmarkDotNet" />
--- a/src/NUnitFramework/benchmarks/nunit.framework.benchmarks/nunit.framework.benchmarks.csproj
+++ b/src/NUnitFramework/benchmarks/nunit.framework.benchmarks/nunit.framework.benchmarks.csproj
@@ -7,8 +7,6 @@
         <LangVersion>latest</LangVersion>
         <Optimize>true</Optimize>
         <RootNamespace>NUnit.Framework</RootNamespace>
-        <GenerateDocumentationFile>true</GenerateDocumentationFile>
-        <NoWarn>1591</NoWarn>
     </PropertyGroup>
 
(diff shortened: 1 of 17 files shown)


stevenaw:
This is ready for review. 

However, this could be taken one step further by updating the TFM from `net6.0` to something newer. Should it be updated to latest (NET 9.0) or latest LTS (net8.0)? Note that latest LTS will hit EOL later I believe

src/NUnitFramework/benchmarks/nunit.framework.benchmarks/nunit.framework.benchmarks.csproj
manfred-brands, line 1: Suggest to change target to net8.0 as being the currently support LTS and target framework

manfred-brands (changes requested):
I currently get compile time errors when DotNetBenchMark is compiling the code, i.e. when running.
MSBuild wants `GenerateDocumentationFile` be set. We have it set on all other projects, I think we should move it to `Directory.Build.props` instead. For some reason MS can only check for 'unused namespace' if documentation is enabled.
However setting that results in CS1591 errors as there is no documentation, we have disabled that in various other test projects.

stevenaw:
Thanks @manfred-brands .
Odd, I didn't get those on the current TFM, but I _had_ when moving to net8.0 initially.

I can make those updates tomorrow. It'll likely also require a `NoWarn` to suppress the warning for missing xmldocs on public members (1591 ?) similar to how our other test projects are already setup.

manfred-brands:
I made them here and can push them if you are ok with that.

stevenaw:
Sure! Feel free, thanks @manfred-brands

manfred-brands (approved):
I'm happy with this now. Verify it runs for you.

src/NUnitFramework/benchmarks/nunit.framework.benchmarks/nunit.framework.benchmarks.csproj
```diff
@@ -2,14 +2,21 @@
...
         <OutputType>Exe</OutputType>
-        <TargetFrameworks>net6.0</TargetFrameworks>
+        <TargetFrameworks>net8.0</TargetFrameworks>
         <ImplicitUsings>enable</ImplicitUsings>
         <LangVersion>latest</LangVersion>
         <Optimize>true</Optimize>
-        <AssemblyOriginatorKeyFile>..\..\..\nunit.snk</AssemblyOriginatorKeyFile>
         <RootNamespace>NUnit.Framework</RootNamespace>
+        <GenerateDocumentationFile>true</GenerateDocumentationFile>
+        <NoWarn>1591</NoWarn>
```
mikkelbu, line 11: Is this needed when we also have it in the Directory.Build.props ?

stevenaw:
Thanks @manfred-brands @mikkelbu 
I should be able to get to both of those in a few hours

src/NUnitFramework/benchmarks/nunit.framework.benchmarks/nunit.framework.benchmarks.csproj
manfred-brands: No, I must have left that in accidentally.

stevenaw:
The requested changes are done and I've confirmed that @manfred-brands 's changes also work for me. 

I'll wait a few days before merging in case anyone wants to take another look.

mikkelbu (approved):
Still looks good to me :). Thanks for the work @stevenaw and @manfred-brands 👍

stevenaw:
Thanks @mikkelbu !

PR #4908, opened 2024-12-24
Add ApplyToAsync override to ThrowsExceptionConstraint

### LINKED ISSUE #4907: Assert.ThatAsync fail to catch expected exception
Given a simple test to check that an async method throws a certain exception - 

```csharp
namespace TestFailToFail;

public class Tests
{
    [Test]
    public async Task Test1()
    {
        await Assert.ThatAsync(async () => throw new Exception("Ugh"), Throws.Exception);
    }

    [Test]
    public async Task Test2()
    {
        await Assert.ThatAsync(() => Task.FromException(new Exception("Ugh")), Throws.Exception);
    }

    [Test]
    public void Test3()
    {
        Assert.That(() => throw new Exception("Ugh"), Throws.Exception);
    }
}
```

With NUnit 4.2.2 - all tests pass (as expected)
With NUnit 4.3.0 (and also 4.3.1) - both `Test1` and `Test2` fail with the thrown exception. Only the sync version - `Test3` pass

```
 Test1
   Source: UnitTest1.cs line 6
   Duration: 8 ms

  Message: 
System.Exception : Ugh

  Stack Trace: 
<<Test1>b__0_0>d.MoveNext() line 8
--- End of stack trace from previous location ---
<<ThatAsync>b__0>d.MoveNext()
--- End of stack trace from previous location ---
Constraint.ApplyToAsync[TActual](Func`1 taskDel)
Assert.ThatAsync(AsyncTestDelegate code, IResolveConstraint constraint, NUnitString message, String actualExpression, String constraintExpression)
Tests.Test1() line 8
GenericAdapter`1.GetResult()
AsyncToSyncAdapter.Await[TResult](TestExecutionContext context, Func`1 invoke)
AsyncToSyncAdapter.Await(TestExecutionContext context, Func`1 invoke)
TestMethodCommand.RunTestMethod(TestExecutionContext context)
TestMethodCommand.Execute(TestExecutionContext context)
<>c__DisplayClass3_0.<PerformWork>b__0()
<>c__DisplayClass1_0`1.<DoIsolated>b__0(Object _)
ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
ContextUtils.DoIsolated(ContextCallback callback, Object state)
ContextUtils.DoIsolated[T](Func`1 func)
SimpleWorkItem.PerformWork()
```

manfred-brands:
Fixes #4907

### DIFF
--- a/src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/ThrowsExceptionConstraint.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System;
+using System.Threading.Tasks;
 using NUnit.Framework.Internal;
 
 namespace NUnit.Framework.Constraints
@@ -40,6 +41,14 @@ namespace NUnit.Framework.Constraints
         public override ConstraintResult ApplyTo<TActual>(ActualValueDelegate<TActual> del)
         {
             return ApplyTo((Delegate)del);
+        }
+
+        /// <inheritdoc/>
+        public override async Task<ConstraintResult> ApplyToAsync<TActual>(Func<Task<TActual>> actual)
+        {
+            var exception = await ExceptionHelper.RecordExceptionAsync(actual, nameof(actual));
+
+            return new ThrowsExceptionConstraintResult(this, exception);
         }
 
         #region Nested Result Class
--- a/src/NUnitFramework/tests/Assertions/AssertThatAsyncTests.cs
+++ b/src/NUnitFramework/tests/Assertions/AssertThatAsyncTests.cs
@@ -28,6 +28,18 @@ namespace NUnit.Framework.Tests.Assertions
         }
 
         [Test]
+        public async Task AssertionPasses_FaultedDelegate_ThrowsAnyException()
+        {
+            await Assert.ThatAsync(() => throw new Exception("Ugh"), Throws.Exception);
+        }
+
+        [Test]
+        public async Task AssertionPasses_FaultedTask_ThrowsAnyException()
+        {
+            await Assert.ThatAsync(() => Task.FromException(new Exception("Ugh")), Throws.Exception);
+        }
+
+        [Test]
         public async Task AssertionPasses_FaultedTask_ThrowsMatchingException()
         {
             await Assert.ThatAsync(() => Task.FromException(new InvalidOperationException()), Throws.InvalidOperationException);


stevenaw (approved):
LGTM. Thanks @manfred-brands ! Feel free to merge if you're ready.

Dreamescaper:
While changes look fine, I don't get why they are needed. The intention was that previous behavior would be preserved if ApplyToAsync is not overridden.

manfred-brands:
The default behaviour of `ApplyAsync` does fail in case of exceptions. That is fine for all normal constraints, just not for any `Throws` one. For the same reason there is an `ApplyAsync` overload for the other `Throw` constraint.

Dreamescaper:
@manfred-brands 
So, my intention with default `ApplyToAsync` was to preserve the existing behavior. And only to allow to change this behavior for Constraints that need that.
Obviously, I failed. If I wanted to preserve the behavior, I shoud've simply invoke sync `ApplyTo` method (which would do sync-over-async) from default `ApplyToAsync` implementation.

Now, I'm not sure what is the best cource of action here.
Sync-over-async is, well, bad, especially for a designated Async method.
OTOH, currently we have a breaking change. This issue might happen for user-created Contstraints.

manfred-brands:
> @manfred-brands So, my intention with default `ApplyToAsync` was to preserve the existing behavior of `ApplyTo` method. And only to allow to change this behavior for Constraints that need that. Obviously, I failed. If I wanted to preserve the behavior, I shoud've simply invoke sync `ApplyTo` method (which would do sync-over-async) from default `ApplyToAsync` implementation.
> 
> Now, I'm not sure what is the best cource of action here. Sync-over-async is, well, bad, especially for a designated Async method. OTOH, currently we have a breaking change. This issue might happen for user-created Contstraints.

Indeed calling `ApplyTo` would keep existing behaviour and making the `Assert.ThatAsync` a fake.
To then use proper `async` would require overloading `ApplyToAsync` for every `Constraint` including user ones.

The approach we have now will convert 99% of the constraints to proper `async` and only the ones that need it require an override. We forgot one because for some reason there are 3 `Throw` constraints.

On the other hand, how many people pass `AsyncDelegate` to constraints that only care about the actual result?
Why call: `await Assert.ThatAsync(SomeAsyncMethod, Is.EqualTo(expected))`
Instead of: `Assert.That(await SomeAsyncMethod(), Is.EqualTo(expected))`
There is more chance they have: `Assert.That(SomeAsyncMethod, Is.EqualTo(expected))` as that code worked before we had `Assert.ThatAsync`

But I have been wrong making assumptions about how code is being used.

The only cases where NUnit needs to call the delegate are `Throws` to be able to inspect the exception and `Delay` to call it more than once.

Yes, there could be user constraints requiring this behaviour.

I suggest we do:
1. Change the implementation in `Constraint` to: `return Task.FromResult(ApplyTo(async () => await taskDel());`
This will call the `ActualValueDelegate` overload which will use the `AsyncToSyncAdapter`.
This method has overloads in the 3 `Throw` and in the `Delay` constraints. The same constraints that have their own `ApplyToAsync` overload. I tested the above by removing that overload. We should add a special test constraint to verify the behaviour.
It will fix end-user constraints that override the `ApplyTo<TActual>(ActualValueDelegate<TActual> del)` method.
It will cause other constraints in my first example to call async-over-sync, but those should be nudged to change to `await` the result before calling `Assert`. Maybe I can create an analyzer rule for this.
1. Document that if a constraints overloads `ApplyTo<TActual>(ActualValueDelegate<TActual> del)` it must also override `ApplyToAsync<TActual>(Func<Task<TActual>> taskDel)` to support proper `async`.

PR #4922, opened 2025-01-10
DelayedConstraint shows full result of underlying constraint.

### LINKED ISSUE #4545: DelegatingConstraintResult does not use its `_innerResult` for `WriteMessageTo`
The `DelegatingConstraintResult` should override the `WriteMessageTo` to call the `_innerResult.WriteMessageTo` method, otherwise it's not writing the expected message to the output.

`WriteActualValueTo` and `WriteAdditionalLinesTo` are already properly forwarded to the `_innerResult`.

https://github.com/nunit/nunit/blob/6a8973732261cc459b7189ac9e2d92515951fb45/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs#L369-L382

### LINKED ISSUE #4909: Assertion failures with PropertiesComparer are hard to debug when combined with other modifiers
```csharp
        await Assert.ThatAsync(() => getExpectedResult()), Is.EqualTo(new ExpectedType
        {
            My = 3.548m,
            Type = 0.312m,
            With = new()
            {
                Lots = true,
                Of = "23456",
                Properties = new(2024, 05, 24)
            }
        }).UsingPropertiesComparer().After(10000, 1000));
```

Assertion above fails with a not really useful message:
```
      Assert.That(() => getExpectedResult(), Is.EqualTo(new ExpectedType
        {
            My = 3.548m,
            Type = 0.312m,
            With = new()
            {
                Lots = true,
                Of = "23456",
                Properties = new(2024, 05, 24)
            }
        }).UsingPropertiesComparer().After(10000, 1000))
  Expected: <My.Name.Space.ExpectedType> after 10000 milliseconds delay
  But was:  <My.Name.Space.ExpectedType>
```

I would be really great to improve it and print the failed property name (like is happens when using PropertiesComparer without After).

### LINKED ISSUE #4919: Assert.ThatAsync doesn't work correctly if calling nested Asserts
When working on #4909 I had a test like:

```
await Assert.ThatAsync(FailingDelayedAssertion, Throws.InstanceOf<AssertionException>()
    .With.Message.Contain("milliseconds delay")
    .With.Message.Contain("property ExpectedType.My:"));
```

Where `FailingDelayedAssertion` was a method calling `Assert.ThatAsync` and is throwing an `AssertionException`.
The above test is reports as failed, but no stack trace and no actual failure message relating to the above constraints.

If I change the text in the `Contain` causing the assert to fail, it then reports `Multiple failures`.
The failure in the original _called_ `Assert` and the failure in the _top-level_ `Assert`.

When changing this to the "pseudo-synchronous" version of `Assert`, the test passes.
```
Assert.That(FailingDelayedAssertion, Throws.InstanceOf<AssertionException>()
    .With.Message.Contain("milliseconds delay")
    .With.Message.Contain("property ExpectedType.My:"));
```

manfred-brands:
Fixes #4909 
Fixes #4919 
Fixes #4545

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
@@ -220,7 +220,7 @@ namespace NUnit.Framework.Constraints
         /// <summary>
         /// Gets text describing a constraint
         /// </summary>
-        public override string Description => $"{BaseConstraint.Description} after {DelayInterval} delay";
+        public override string Description => $"After {DelayInterval} delay";
 
         /// <summary>
         /// Test whether the constraint is satisfied by a given value
@@ -401,9 +401,11 @@ namespace NUnit.Framework.Constraints
                 _innerResult = innerResult;
             }
 
-            public override void WriteActualValueTo(MessageWriter writer) => _innerResult.WriteActualValueTo(writer);
-
-            public override void WriteAdditionalLinesTo(MessageWriter writer) => _innerResult.WriteAdditionalLinesTo(writer);
+            public override void WriteMessageTo(MessageWriter writer)
+            {
+                writer.WriteMessageLine(Description);
+                _innerResult.WriteMessageTo(writer);
+            }
         }
     }
 }
--- a/src/NUnitFramework/framework/Constraints/MsgUtils.cs
+++ b/src/NUnitFramework/framework/Constraints/MsgUtils.cs
@@ -5,6 +5,8 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
+using System.Linq;
+using System.Reflection;
 using System.Text;
 using NUnit.Framework.Internal;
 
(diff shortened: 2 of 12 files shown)


stevenaw (commented):
Thanks for the PR @manfred-brands 
I hadn't time to properly review tonight, but I suspect you may've fixed a _third_ bug too here 🙂

src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
```diff
@@ -401,9 +401,11 @@ public DelegatingConstraintResult(IConstraint constraint, ConstraintResult inner
                 _innerResult = innerResult;
             }
 
-            public override void WriteActualValueTo(MessageWriter writer) => _innerResult.WriteActualValueTo(writer);
-
-            public override void WriteAdditionalLinesTo(MessageWriter writer) => _innerResult.WriteAdditionalLinesTo(writer);
+            public override void WriteMessageTo(MessageWriter writer)
+            {
+                writer.WriteMessageLine(Description);
+                _innerResult.WriteMessageTo(writer);
```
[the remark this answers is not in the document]
manfred-brands: You are good at finding issues. Unknowingly I followed a similar solution as described in that ticket.

OsirisTerje (approved):
LGTM.

Checked the tests.

src/NUnitFramework/framework/Constraints/MsgUtils.cs
```diff
@@ -94,6 +96,55 @@ static MsgUtils()
             AddFormatter(next => val => val is Array valArray ? FormatArray(valArray) : next(val));
         }
 
+        /// <summary>
+        /// Try to format the properties of an object as a string.
+        /// </summary>
+        /// <param name="val">The object to format.</param>
+        /// <returns>Formatted string for all properties.</returns>
+        public static string FormatValueProperties(object? val)
```
stevenaw, line 104: nitpick: Was there a use case for making this public, or could it be made private to maximize our ability to evolve it without breaking others?

For example, there could be a time in the future where we'd want to pass in a precomputed collection of properties from elsewhere in the framework, or perhaps allowing this to recurse to child objects with a configurable "depth".

stevenaw (commented):
Thanks @manfred-brands ! This looks great to me too.
I left one minor suggestion to make the new `MsgUtils` method private unless there was a use case you knew of to keep it public

src/NUnitFramework/framework/Constraints/MsgUtils.cs
manfred-brands: > nitpick: Was there a use case for making this public, or could it be made private to maximize our ability to evolve it without breaking others?

It cannot be private as it is called from another class (`EqualConstraintResult`). I didn't want to introduce it in the formatter call chain for _all_ values as that will have side effects where properties are not to be printed, like DbContext.

The whole class `MsgUtils` is `internal` and therefore we can change the signature without breaking external code.

stevenaw: Oh! I had meant internal, but has missed that MsgUtils itself was already internal. Sorry about that

stevenaw (approved):

PR #4927, opened 2025-01-24
Fix numeric comparison when actual is declared as object.

### LINKED ISSUE #4917: Multiple unannounced breaking changes in minor version
Hi, we've recently finished a months old migration from 3.14 to 4.1, migrating 110K tests using `NUnit.Analyzers`. Today, I tried upgrading our version from 4.1 to 4.3.2 but to my distress I faced an important amount of compilation errors.

It seems like the new minor versions introduced several breaking changes that were not announced in the [changelog](https://docs.nunit.org/articles/nunit/release-notes/framework.html#nunit-42---august-22-2024). Here are a few simplified occurrences that I caught:
```csharp
// Example 1
Dictionary<long, float>? x = null;
Assert.That(x, Is.EqualTo(null).AsCollection);

// Example 2
Assert.That(0.0, Is.EqualTo(0L).Within(0.1));

// Example 3
Assert.That(true, Is.EqualTo(default));
```

This code builds with 4.1.0 but not 4.3.2 :(

They can all be fixed like the following but that will be weeks of extra work that I'm not sure we can afford.
```diff
// Example 1
Dictionary<long, float>? x = null;
- Assert.That(x, Is.EqualTo(null).AsCollection);
+ Assert.That(x, Is.Null);

// Example 2
- Assert.That(0.0, Is.EqualTo(0L).Within(0.1));
+ Assert.That(0.0, Is.EqualTo((double)0L).Within(0.1));

// Example 3
- Assert.That(true, Is.EqualTo(default));
+ Assert.That(true, Is.False);
```

manfred-brands:
Fixes #4917

And value types must be mismatched: double vs float.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/EqualConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/EqualConstraint.cs
@@ -156,6 +156,17 @@ namespace NUnit.Framework.Constraints
         }
 
         /// <summary>
+        /// Flag the constraint to use a tolerance defined elsewhere when determining equality.
+        /// </summary>
+        /// <param name="tolerance">Tolerance to be used</param>
+        /// <returns>Self.</returns>
+        internal EqualConstraint WithinConfiguredTolerance(Tolerance tolerance)
+        {
+            _tolerance = tolerance;
+            return this;
+        }
+
+        /// <summary>
         /// Flag the constraint to use a tolerance when determining equality.
         /// </summary>
         /// <param name="amount">Tolerance value to be used</param>
--- a/src/NUnitFramework/framework/Constraints/EqualNumericWithoutUsingConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/EqualNumericWithoutUsingConstraint.cs
@@ -155,7 +155,7 @@ namespace NUnit.Framework.Constraints
 
                 hasSucceeded = equatable.Equals(_expected);
             }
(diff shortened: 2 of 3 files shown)


stevenaw:
LGTM from my phone, thanks @manfred-brands 
I'll try and take a closer look from a computer soon

stevenaw (approved):
LGTM thanks @manfred-brands 
I'll leave this unmerged in case you wanted eyes from @mikkelbu specifically

src/NUnitFramework/framework/Constraints/EqualConstraint.cs
```diff
@@ -155,6 +155,17 @@ public EqualConstraint AsCollection
             }
         }
 
+        /// <summary>
+        /// Flag the constraint to use a tolerance defined elsewhere when determining equality.
+        /// </summary>
+        /// <param name="tolerance"></param>
+        /// <returns></returns>
```
mikkelbu, line 162:
```suggestion
        /// <param name="tolerance">Tolerance value to be used.</param>
        /// <returns>Self.</returns>
```

mikkelbu (approved):

src/NUnitFramework/framework/Constraints/EqualConstraint.cs
manfred-brands: Thanks, comments updated.

PR #4928, opened 2025-01-24
Fix test error in TestContextOneTimeTearDownTests

### LINKED ISSUE #4906: Test error in TestContextOneTimeTearDownTests.
This test have a set of Asserts in its OneTimeTearDown method.
These asserts show up as errors in Rider and Resharper, but in dotnet test, nor in Test Explorer in Visual Studio.

I have moved it into a separate repro test https://github.com/nunit/nunit.issues/tree/main/Issue4906  

There are several issues here:

1. The test fails, so we have an error we have not detected. 
This the essence of this Issue.
This test seems to have been written way back, in 2014/2015
The testfixture has one Explicit test, and checks that tthe Result.SkipCOunt is equalto 1. It is not.
image

2. (This pt should probably be transferred to the adapter) Rider/Resharper detects that this is a fault.  It is not reported in VS/dotnet , it stops in the adapter.  Enabling the dump I see it in the response from the engine/framework, but as it is test output, we have nowhere to put that output.  
I know that MSTest adds a kind of extra test node in these cases. We could do the same.  In NUnit all suites are also tests, and might have information.  It would make sense to be able to show that.  Rider does this too, so you see the node is failing. 
In VS it is shown in the Output/Test window:
image
Further, NUnitLite does show it, but it is suppressed somewhere in out NUnit build output.  I see it in the repro code I made.
 
Dump file output:
image

In the dump file we can also see that the skippedcount is 0, ref pt 1 here.

The repro is here: https://github.com/nunit/nunit.issues/tree/main/Issue4906

3. The repro has been enabled for NUNitLite.  
When running with VS the error is visible in the Output/Test as mentioned above.
When running with NUnitLite, it is not there, and debugging it, show SkippedCount = 1.  Which is correct.  I have no clue yet why it works when seperated out into this repro, and ONLY when running with NUnitLite. 

@manfred-brands @stevenaw Comments ?  @rprouse @CharliePoole   You guys wrote this test 10 years ago.  Does the console show this node?  Is this pre-adapter ?  Any idea why NUnitLite is confused ?

[UPDATE]
Just checked with the console:
image
It doesn't react to the test either.

But Rider does:
image

stevenaw:
Fixes #4906 

Re: my "filter" comments in the issue. I wasn't able to see anything of note based on an initial look.

### DIFF
--- a/src/NUnitFramework/tests/TestContextTests.cs
+++ b/src/NUnitFramework/tests/TestContextTests.cs
@@ -480,7 +480,7 @@ namespace NUnit.Framework.Tests
             Assert.That(tearDownResult, Is.Not.Null);
             Assert.Multiple(() =>
             {
-                Assert.That(tearDownResult!.Outcome, Is.EqualTo(result.ResultState));
+                Assert.That(tearDownResult.Outcome, Is.EqualTo(result.ResultState));
                 Assert.That(tearDownResult.PassCount, Is.EqualTo(result.PassCount));
                 Assert.That(tearDownResult.FailCount, Is.EqualTo(result.FailCount));
                 Assert.That(tearDownResult.SkipCount, Is.EqualTo(result.SkipCount));
--- a/src/NUnitFramework/tests/TestContextTests.cs
+++ b/src/NUnitFramework/tests/TestContextTests.cs
@@ -9,6 +9,7 @@ using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
 using NUnit.TestData.TestContextData;
 using NUnit.Framework.Tests.TestUtilities;
+using NUnit.TestData;
 
 namespace NUnit.Framework.Tests
 {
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/tests/TestContextTests.cs
```diff
@@ -459,35 +460,31 @@ public void TearDown()
...
-            Assert.That(context.Result.Outcome, Is.EqualTo(ResultState.Success));
-            Assert.That(context.Result.PassCount, Is.EqualTo(2));
-            Assert.That(context.Result.FailCount, Is.EqualTo(0));
-            Assert.That(context.Result.SkipCount, Is.EqualTo(1));
+            var tearDownResult = fixture.ResultInOneTimeTearDown;
+
+            Assert.That(tearDownResult, Is.Not.Null);
+            Assert.Multiple(() =>
+            {
+                Assert.That(tearDownResult!.Outcome, Is.EqualTo(result.ResultState));
```
manfred-brands, line 483: You shouldn't need the null suppression operator `!`.

manfred-brands (commented):
Looks good to me.
It proves that the result in the `OneTimeTearDown` matches the overall result.

manfred-brands (approved):

src/NUnitFramework/tests/TestContextTests.cs
stevenaw: I can try to take a closer look at this one tomorrow. I had added it to suppress a warning, but I can also try updating the referenced version of the analyzer package. I think I saw that we currently reference 4.3.0 whereas latest is 4.6.0

stevenaw: Aha. Appears I was still a few bugfix versions behind on Visual Studio. I've upgraded to latest and now no longer need the `!`. Thanks for pointing that out!

I've pushed the change to remove it and will merge once the builds pass

PR #4929, opened 2025-01-24
Fix DelayedConstraint calling delegates at double the rate.

### LINKED ISSUE #2841: DelayedConstraint calls delegate twice

manfred-brands:
Fixes #2841

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DelayedConstraint.cs
@@ -262,12 +262,10 @@ namespace NUnit.Framework.Constraints
 
             if (PollingInterval.IsNotZero)
             {
-                long nextPoll = TimestampOffset(now, PollingInterval.AsTimeSpan);
                 while ((now = Stopwatch.GetTimestamp()) < delayEnd)
                 {
-                    if (nextPoll > now)
-                        ThreadUtility.BlockingDelay((int)TimestampDiff(delayEnd < nextPoll ? delayEnd : nextPoll, now).TotalMilliseconds);
-                    nextPoll = TimestampOffset(now, PollingInterval.AsTimeSpan);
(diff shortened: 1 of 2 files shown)


manfred-brands:
Not sure what goes on with these build agents:
```
Assert.That(list.PollCount, Is.LessThanOrEqualTo(10 + 1))
  Expected: less than or equal to 11
  But was:  576
```

This could only happen if the Stopwatch doesn't increase and the `Thread.Sleep` doesn't actually sleep.

manfred-brands:
I don't like my hack for the github runners.
Maybe I missed something.

stevenaw:
@manfred-brands The last few weeks have been busy for me, but I'll try to get a review in of this in the coming days

stevenaw:
Thanks @manfred-brands . Changes themselves look good. Odd about the github runner. It looks like the run before adding that explicit loop only failed on linux, and the build came with a notice that `ubuntu-latest` would soon change to 24.02. That change has since happened so I'm going to try pushing a test branch with your last commit removed to see if that works now.

stevenaw (approved):
It looks like that last commit is still needed for GH runners to pass. I don't have any concerns with this PR as-is, however I'll abstain from merging myself in case someone else wants to take a look.
@manfred-brands feel free to merge as well if you're satisfied

PR #4934, opened 2025-02-03
Fix PropertiesComparer when properties are not comparable

### LINKED ISSUE #4933: Dictionary assertion falsely passes with PropertiesComparer when subtypes are involved
```csharp
class ClassA
{
    public string V1 { get; set; }
    public string V2 { get; set; }
}

class ClassB : ClassA
{
    public string V3 { get; set; }
}

[Test]
public async Task VerifySubtypesBehavior()
{
    var expected = new Dictionary<string, ClassA>
    {
        ["Key"] = new ClassA { V1 = "1", V2 = "1" }
    };

    var actual = new Dictionary<string, ClassA>
    {
        ["Key"] = new ClassB { V1 = "2", V2 = "2" }
    };

    Assert.That(actual, Is.EqualTo(expected).UsingPropertiesComparer()); // Passes!
}
```

Assertion shouldn't pass here, as have completely different values.
It works as expected when both dictionaries have ClassA values, but not when subtypes are involved.

Dreamescaper:
Fixes #4933 .

Turns out the problem is not Dictionary specific. It reproduces even for something like this:
```csharp
class TypeWithObjectProperty
{
     public object? Value { get; set; }
}

var actual = new TypeWithObjectProperty { Value = new ObjectA("WTF") };
var expected = new TypeWithObjectProperty { Value = new ObjectB(123.345) };

Assert.That(actual, Is.EqualTo(expected).UsingPropertiesComparer()); // PASS!!!
```

The problem is that PropertiesComparer didn't check for NotSupported result - it basically treated it like a successul comparision.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
@@ -53,12 +53,7 @@ namespace NUnit.Framework.Constraints.Comparers
 
                 EqualMethodResult result = equalityComparer.AreEqual(xPropertyValue, yPropertyValue, ref tolerance, comparisonState);
 
-                if (result == EqualMethodResult.TypesNotSupported)
-                {
-                    return EqualMethodResult.TypesNotSupported;
-                }
-
-                if (result == EqualMethodResult.ComparedNotEqual)
+                if (result == EqualMethodResult.ComparedNotEqual || result == EqualMethodResult.TypesNotSupported)
                 {
                     return PropertyNotEqualResult(equalityComparer, i, declaringTypeName, property.Name, xPropertyValue, yPropertyValue);
                 }
--- a/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
@@ -52,6 +52,12 @@ namespace NUnit.Framework.Constraints.Comparers
                 object? yPropertyValue = property.GetValue(y, null);
 
                 EqualMethodResult result = equalityComparer.AreEqual(xPropertyValue, yPropertyValue, ref tolerance, comparisonState);
+
+                if (result == EqualMethodResult.TypesNotSupported)
+                {
+                    return EqualMethodResult.TypesNotSupported;
+                }
+
                 if (result == EqualMethodResult.ComparedNotEqual)
                 {
                     return PropertyNotEqualResult(equalityComparer, i, declaringTypeName, property.Name, xPropertyValue, yPropertyValue);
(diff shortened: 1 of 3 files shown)


Dreamescaper:
@manfred-brands 
Could you please take a look?

Dreamescaper:
The only problem is that currently the message is not very readable:
```
{"No comparer found for instances of type 'TypeWithObjectProperty' and 'TypeWithObjectProperty'"}
```

src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
```diff
@@ -52,7 +52,8 @@ public static EqualMethodResult Equal(object x, object y, ref Tolerance toleranc
                 object? yPropertyValue = property.GetValue(y, null);
 
                 EqualMethodResult result = equalityComparer.AreEqual(xPropertyValue, yPropertyValue, ref tolerance, comparisonState);
-                if (result == EqualMethodResult.ComparedNotEqual)
+
+                if (result == EqualMethodResult.ComparedNotEqual || result == EqualMethodResult.TypesNotSupported)
```
manfred-brands, line 56: Should it not return the `TypesNotSupported`?
The idea is that if one comparer doesn't support items, it bubble up and the next comparer is called.

As the `PropertiesComparer` is the last, the result is likely the same.

Dreamescaper: Originally I returned it, but the error message is not very helpful. It prints root object type name, something like `Can't find comparer for types 'TypeWithObjectProperty' and 'TypeWithObjectProperty'`. In case of assertion failure it prints exact property, which is much more useful.

Plus if I return `TypesNotSupported`, it fails with `NotSupportedException`, `AssertionException` is probably more suitable here.

manfred-brands: It depends if we want to distinguish between comparing as different vs not being able to compare. 

What does normal comparison do when comparing instances of different types?

Dreamescaper: In regular case, it's `NotSupportedException`.
Which makes sense - as I have explicitly passed two incomparrable values to assert method, i.e. I have done something wrong.

In this case, however - I'm passing two `TypeWithObjectProperty` instances, that's a perfectly valid thing to do. I can't know if some deep down property values have different types - that's exactly what I'm testing. Therefore - it simply should be an assertion failure.

manfred-brands: One final thing. What does a regular compare on a dictionary with different element types return?

Dreamescaper: AssertionException:
Assert.That(actual, Is.EqualTo(expected))
  Expected: <NUnit.Framework.Tests.TypeWithObjectProperty>
  But was:  <NUnit.Framework.Tests.TypeWithObjectProperty>

manfred-brands: Thanks

Dreamescaper: Oh, Dictionary, not object. Still,
```
Assert.That(actual, Is.EqualTo(expected))
  Expected and actual are both <System.Collections.Generic.Dictionary`2[System.String,System.Object]> with 1 elements
```

stevenaw:
Thanks for another contribution @Dreamescaper 
Great test coverage.

PR #4936, opened 2025-02-10
Catch exceptions in AfterTestCommand.AfterTest

### LINKED ISSUE #2938: Test runner hangs after Assert.Fail in AfterTest
I'm running an NUnit 3.9 test suite with nunit3-console, with a TestAction which asserts on some state after the fixture finishes. When the assert fails, it throws an exception and the runner hangs indefinitely. Are AfterTest implementations not allowed to assert or throw?

```C#
using NUnit.Framework;
namespace Tests
{
    [TestFixture]
    [MyTestAction]
    public class MyTest
    {
        [Test]
        public void Pass() { }
    }

    public class MyTestActionAttribute : TestActionAttribute
    {
        public override void BeforeTest(ITest test)
        {
            base.BeforeTest(test);
        }

        public override void AfterTest(ITest test)
        {
            Assert.Fail("A test left resources open");
            base.AfterTest(test);
        }
    }
}
```

manfred-brands:
Fixes #2938

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/AfterTestCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/AfterTestCommand.cs
@@ -29,7 +29,10 @@ namespace NUnit.Framework.Internal.Commands
                 context.CurrentResult = innerCommand.Execute(context);
             });
 
-            AfterTest(context);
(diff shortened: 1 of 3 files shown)


OsirisTerje (approved):
lgtm

Issue #4937, opened 2025-02-13
`PartitionFilter.ComputeHasValue` throws exception when test name is greater than 4096 characters


WillRayAtPropertyMe:
We have unit tests in our project that contain test case data that can extend beyond 4000 characters. When running unit tests in a partition, the following exception is thrown (versions: Nunit 4.3.2 and NUnit3TestAdapter 5.0.0):

```
  Error Message:
   NUnit.Engine.NUnitEngineException: An exception occurred in the driver while exploring tests.
 ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.ArgumentException: The output byte buffer is too small to contain the encoded data, encoding codepage '65001' and fallback 'System.Text.EncoderReplacementFallback'. (Parameter 'bytes')
   at System.Text.Encoding.ThrowBytesOverflow()
   at System.Text.Encoding.ThrowBytesOverflow(EncoderNLS encoder, Boolean nothingEncoded)
   at System.Text.Encoding.GetBytesWithFallback(ReadOnlySpan`1 chars, Int32 originalCharsLength, Span`1 bytes, Int32 originalBytesLength, EncoderNLS encoder, Boolean throwForDestinationOverflow)
   at System.Text.Encoding.GetBytesWithFallback(Char* pOriginalChars, Int32 originalCharCount, Byte* pOriginalBytes, Int32 originalByteCount, Int32 charsConsumedSoFar, Int32 bytesWrittenSoFar, Boolean throwForDestinationOverflow)
   at System.Text.UTF8Encoding.GetBytes(ReadOnlySpan`1 chars, Span`1 bytes)
   at NUnit.Framework.Internal.Filters.PartitionFilter.ComputeHashValue(String name)
   at NUnit.Framework.Internal.Filters.PartitionFilter.ComputePartitionNumber(ITest value)
   at NUnit.Framework.Internal.Filters.PartitionFilter.Match(ITest test)
   at NUnit.Framework.Internal.TestFilter.MatchDescendant(ITest test)
   at NUnit.Framework.Internal.TestFilter.MatchDescendant(ITest test)
   at NUnit.Framework.Internal.TestFilter.MatchDescendant(ITest test)
   at NUnit.Framework.Internal.TestFilter.Pass(ITest test, Boolean negated)
   at NUnit.Framework.Internal.TestFilter.Pass(ITest test)
[...]

PR #4941, opened 2025-02-18
Added trimming to ComputeHashValue to handle names over 4096 bytes.

WillRayAtPropertyMe:
This PR addresses #4937 by adding in trimming for names above 4096 bytes, including a unit test to check for this case.


WillRayAtPropertyMe:
@dotnet-policy-service agree company="PropertyMe"

mikkelbu (approved):
I don't know much about `Encoder`, but as far as I can tell this looks sound.

Is it possible to have a test that it actually uses the first 4096 bytes in hashvalue (I guess it don't have to be exactly 4096)? AFAICT the added tests would also pass if the code was accidentally changed to e.g. only use the first couple bytes.

WillRayAtPropertyMe:
Hey @mikkelbu the two tests provided are both over 4096 bytes. Just to clarify, the concern is if those tests were changed to be less than 4096 bytes - they would still pass?

mikkelbu:
> Hey @mikkelbu the two tests provided are both over 4096 bytes. Just to clarify, the concern is if those tests were changed to be less than 4096 bytes - they would still pass?

Sorry about the long answer time. I was just wondering if it would make sense to have a test that checks that two names that are distinct (but after 4096 bytes) are considered equal by ComputeHashValue), but I don't think this add much.

PR #4944, opened 2025-02-22
Mark ApplyTo<TActual>(ref TActual actual) as obsolete

### LINKED ISSUE #4930: What to do with `IConstraint.Apply<TActual>(ref TActual)`
The `IConstraint` interface defines an overload where one can pass a actual value by reference.
Nothing in NUnit tests seems to call this as far as I can tell.

There is one overload in `DelayedConstraint` but that drops the `ref` prefix and would pass actual by value to the base constraint.

The only place where this might have been intended for is a `ref struct` that modifies itself, like in:

```csharp
private struct PretendList
{
    public int PollCount { get; private set; }

    public int Count
    {
        get
        {
            PollCount++;
            return 0;
        }
    }
}
```

Mutable structs are a bad concept unless one can guarantee it is always passed by ref.

The `DelayedConstraint` calls `BaseConstraint.Apply(actual)` without `ref`, it therefore uses copy semantics and the `PollCount` will always be `0`.
Even changing the code to pass `ref actual` doesn't fix it as the `PropertyConstraint` called by `Has.Count` looses the `ref`.
I wouldn't even know of passing `ref` to `PropertyInfo.GetValue` even works.

I suggest to remove the `Apply<TActual>(ref TActual)` overloads in the next major version as it is technically breaking.
But it is currently not working for any of our constraints as they all behave like `Apply<TActual>(TActual)` and any mutations are lost.

manfred-brands:
Fixes #4930

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Constraint.cs
+++ b/src/NUnitFramework/framework/Constraints/Constraint.cs
@@ -105,6 +105,7 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         /// <param name="actual">A reference to the value to be tested</param>
         /// <returns>A ConstraintResult</returns>
+        [Obsolete("This was never implemented and will be removed.")]
         public virtual ConstraintResult ApplyTo<TActual>(ref TActual actual)
         {
             return ApplyTo(actual);
--- a/src/NUnitFramework/framework/Constraints/IConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/IConstraint.cs
@@ -1,5 +1,7 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
+using System;
+
 namespace NUnit.Framework.Constraints
 {
     /// <summary>
@@ -59,6 +61,7 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         /// <param name="actual">A reference to the value to be tested</param>
         /// <returns>A ConstraintResult</returns>
+        [Obsolete("This was never implemented and will be removed.")]
         ConstraintResult ApplyTo<TActual>(ref TActual actual);
 
         #endregion


src/NUnitFramework/framework/Constraints/IConstraint.cs
```diff
@@ -59,6 +61,7 @@ public interface IConstraint : IResolveConstraint
         /// </summary>
         /// <param name="actual">A reference to the value to be tested</param>
         /// <returns>A ConstraintResult</returns>
+        [Obsolete("This never really worked")]
```
stevenaw, line 64:
```suggestion
        [Obsolete("This was never implemented and will be removed.")]
```

src/NUnitFramework/framework/Constraints/Constraint.cs
```diff
@@ -105,6 +105,7 @@ public virtual ConstraintResult ApplyTo<TActual>(ActualValueDelegate<TActual> de
         /// </summary>
         /// <param name="actual">A reference to the value to be tested</param>
         /// <returns>A ConstraintResult</returns>
+        [Obsolete("This never really worked")]
```
stevenaw, line 108:
```suggestion
        [Obsolete("This was never implemented and will be removed.")]
```

stevenaw (approved):
Thanks for fixing this @manfred-brands . I've left a small optional suggestion for what I think may be clearer wording.

src/NUnitFramework/framework/Constraints/IConstraint.cs
manfred-brands: Applied.

src/NUnitFramework/framework/Constraints/Constraint.cs
manfred-brands: Applied

manfred-brands:
@OsirisTerje Can we do something about the MacOS build? Why would a NuGet package download fine on WIndows/Linux but not on MacOS? Do we even need to sign just to test?

mikkelbu:
> @OsirisTerje Can we do something about the MacOS build? Why would a NuGet package download fine on WIndows/Linux but not on MacOS? Do we even need to sign just to test?

I've made a temporary fix that removes installing sign from the MacOs workflow - see #4950

mikkelbu (approved):

OsirisTerje:
@manfred-brands Sorry - I am out travelling, currently in Australia :-)  North of Brisbane.  But I see @mikkelbu  has removed the signing on mac (good).   

You say that the package may not work on Mac ?  

Isn't it only the build that fails?  The package should be identical for all targets.

manfred-brands:
> @manfred-brands Sorry - I am out travelling, currently in Australia :-) North of Brisbane. But I see @mikkelbu has removed the signing on mac (good).

There is still a lot of Australia north of Brisbane, but I presume you are at the sunshine coast (Noosa). Enjoy.

> You say that the package may not work on Mac ?
> 
> Isn't it only the build that fails? The package should be identical for all targets.

I wondered why the same package is fine on Windows/Linux and not on MacOS.

stevenaw:
I have to wonder if an update to the Mac image has compatibility issues. 

It looks like `macos-latest` was last updated Jan 20 based on https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md
Anecdotally (as I can't view the build history at the moment) I think I remember the first set of changes to PR #4925 passing when the PR was first made on Jan 17

mikkelbu:
> I have to wonder if an update to the Mac image has compatibility issues.
> 
> It looks like `macos-latest` was last updated Jan 20 based on https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md Anecdotally (as I can't view the build history at the moment) I think I remember the first set of changes to PR #4925 passing when the PR was first made on Jan 17

So far I've not seen any other reports of this issue (but I've not done a thorough search). I tried bumping to `macos-15` in #4949, but that also failed. As far as I can tell we started seeing these kind of errors on Feb 15 - https://github.com/nunit/nunit/actions/runs/13346353761

It also seems like the macos-14 image was updated on Jan 20 (see image - from https://github.com/nunit/nunit/actions/runs/13610136540/job/38046197396 -
image

But either signature expired a long time ago 
image

(But I don't think I'll spend more time on this now :) )

PR #4946, opened 2025-02-24
Add Support for MultipleOf, Even and Odd.

### LINKED ISSUE #4945: Integer `Assert.That(i, Is.Even);`/`Assert.That(i, Is.Odd);` and `Assert.That(i, Is.MultipleOf(step));`
Would it be possible to add `Assert.That(i, Is.Even);`/`Assert.That(i, Is.Odd);` for integers, as well as `Assert.That(i, Is.MultipleOf(step));`?

I am fully aware all this can be achieved by simple numerics. Still, it would be convenient and more explicitly visible what is intended:
```
var items = GetItems();
Assert.That(items.Length, Is.Even);
for (int i = 0; i < items.Length; i += 2)
{
    // Do something with item pairs.
}
```
```
var items = GetItems();
Assert.That(items.Length, Is.MultipleOf(3));
for (int i = 0; i < items.Length; i += 3)
{
    // Do something with item triples.
}
```

manfred-brands:
Fixes #4945

### DIFF
--- a/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
+++ b/src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
@@ -384,6 +384,25 @@ namespace NUnit.Framework.Constraints
 
         #endregion
 
+        #region MultipleOf, Even, Odd
+
+        /// <summary>
+        /// Returns a constraint that tests for a number to be a multiple of another.
+        /// </summary>
+        public MultipleOfConstraint MultipleOf(int multiple) => Append(new MultipleOfConstraint(multiple));
+
+        /// <summary>
+        /// Returns a constraint that tests for a number to be even (i.e. a multiple of two)
+        /// </summary>
+        public MultipleOfConstraint Even => MultipleOf(2);
+
+        /// <summary>
+        /// Returns a constraint that tests for a number to be odd.
+        /// </summary>
+        public MultipleOfConstraint Odd => Not.Even;
+
+        #endregion
+
         #region Empty
 
         /// <summary>
--- a/src/NUnitFramework/framework/Constraints/MultipleOfConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/MultipleOfConstraint.cs
@@ -5,7 +5,7 @@ using System;
 namespace NUnit.Framework.Constraints
 {
     /// <summary>
-    /// EvenConstraint tests that the actual value is an even integer
+    /// MultipleOfConstraint tests that the actual value is a multiple of a specified integer.
     /// </summary>
     public class MultipleOfConstraint : Constraint
     {
--- a/src/NUnitFramework/framework/Constraints/MultipleOfConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/MultipleOfConstraint.cs
@@ -1,0 +1,55 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using System;
+
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// EvenConstraint tests that the actual value is an even integer
+    /// </summary>
+    public class MultipleOfConstraint : Constraint
+    {
+        private readonly uint _multiple;
+
+        /// <summary>
+        /// Initializes an instance of the <see cref="MultipleOfConstraint"/>.
+        /// </summary>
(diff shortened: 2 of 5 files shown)


src/NUnitFramework/framework/Constraints/MultipleOfConstraint.cs
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using System;
+
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// EvenConstraint tests that the actual value is an even integer
```
mikkelbu, line 8:
```suggestion
    /// MultipleOfConstraint tests that the actual value is a multiple of a specified integer.
```

src/NUnitFramework/framework/Constraints/MultipleOfConstraint.cs
```diff
@@ -0,0 +1,55 @@
...
+                result = value % _multiple == 0;
+            }
+            else if (actual is uint || actual is ulong || actual is ushort || actual is byte)
+            {
+                ulong value = Convert.ToUInt64(actual);
+                result = value % _multiple == 0;
+            }
+            else
+            {
+                // Should this throw an exception instead?
```
mikkelbu, line 48: I'm on a phone, so cannot check, but how does this effect the resulting assertion? (and cannot remember much about the constraints)

maettu-this: Had a quick look at the `NaNConstraint`, which only applies to floating point types, that implementation is based on `ConstraintResult` with `bool isSuccess = false` in case of non-floating point types:
https://github.com/nunit/nunit/blob/main/src/NUnitFramework/framework/Constraints/NaNConstraint.cs#L23

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
```diff
@@ -384,6 +384,25 @@ public Constraint Matches<TActual>(Predicate<TActual> predicate)
...
+
+        /// <summary>
+        /// Returns a constraint that tests for a number to be a multiple of another.
+        /// </summary>
+        public MultipleOfConstraint MultipleOf(int multiple) => Append(new MultipleOfConstraint(multiple));
+
+        /// <summary>
+        /// Returns a constraint that tests for a number to be even (i.e. a multiple of two)
+        /// </summary>
+        public MultipleOfConstraint Even => MultipleOf(2);
```
thomhurst, line 397: This would fail for 0, but 0 is considered even.

https://en.wikipedia.org/wiki/Parity_of_zero

manfred-brands: No it does not. 0 is a multiple of 2. (0 * 2 => 0).
To verify this I added a test to the PR:
```csharp
Assert.That(0, Is.Even);
```

src/NUnitFramework/framework/Constraints/MultipleOfConstraint.cs
manfred-brands: Removed comment to keep behaviour where any non fixed point type is considers not a multiple of anything.

src/NUnitFramework/framework/Constraints/MultipleOfConstraint.cs
manfred-brands: Thanks, I initially had separate Even and Odd constraints until I changed them to be instantiations of MultipleOfConstraint instead.

src/NUnitFramework/framework/Constraints/ConstraintExpression.cs
thomhurst: Oh I thought I saw that the multiple of constraint had a guard where the input had to be greater than 0. But it was probably looking at the 2, not the 0, my bad.

stevenaw (approved):
Looks good to me. Thanks @manfred-brands

manfred-brands:
@nunit/framework-team What are we going to do with the MaxOS build? It is stopping PRs going in.

mikkelbu:
> @nunit/framework-team What are we going to do with the MaxOS build? It is stopping PRs going in.

I've made a temporary fix that removes installing sign from the MacOs workflow - see #4950

mikkelbu (approved):

PR #4950, opened 2025-03-01
fix: Attempt to remove sign from MacOS workflow

mikkelbu:


stevenaw (approved):
Thanks for the workaround @mikkelbu 
Approving from my phone

PR #4951, opened 2025-03-03
Use OverloadResolutionPriorityAttribute to prioritise generic Using overloads

### LINKED ISSUE #4916: Is.EqualTo(double).Using(EqualityComparer<double>) causes overload ambiguity error
Updated from NUnit 4.2.2 to 4.3.2 and now some tests fail to compile. 

We have tests like this where actual and expected are doubles and `DoubleComparer.Approximately()` is a method that returns a `EqualityComparer<double>`. 

```csharp
Assert.That(actual, Is.EqualTo(expected).Using<double>(DoubleComparer.Approximately(precision)));
```

This triggers this compilation error: 
```
error CS0121: The call is ambiguous between the following methods or properties: 'IEqualWithUsingConstraintExtensions.Using<T
Expected>(IEqualWithUsingConstraint<TExpected>, IEqualityComparer<TExpected>)' and 'IEqualWithUsingConstraintExtensions.Using<TExpected>(IEqualWithUsingConstraint<TExpected>, IEqualityComparer)'
```

Current workaround is to introduce an explicit cast in our tests:
```csharp
Assert.That(actual, Is.EqualTo(expected).Using((IEqualityComparer<double>)DoubleComparer.Approximately(precision)));
```

manfred-brands:
Fixes #4916 

Only if test-code compiled with 9.0.200 (or later)

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "9.0.100",
+    "version": "9.0.200",
     "allowPrerelease": false,
     "rollForward": "latestFeature"
   }
--- a/src/NUnitFramework/Directory.Build.props
+++ b/src/NUnitFramework/Directory.Build.props
@@ -1,7 +1,7 @@
 ﻿<Project>
 
   <PropertyGroup>
-    <LangVersion Condition="'$(MSBuildProjectExtension)' == '.csproj'">12</LangVersion>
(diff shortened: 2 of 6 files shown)


manfred-brands:
@jnm2 Did you want any other changes?

jnm2:
Looks good to me. I will just caution what I said earlier, which is that the tradeoff with this approach is that it has no effect for people who are using a langversion < 13.

PR #4957, opened 2025-03-08
Issue4954 equal to date time

### LINKED ISSUE #4954: Regression in `Is.EqualTo(DateTime)` constraint in 4.3.x for types that implement `IEquatable<DateTime>`
After upgrading from 4.2.2 to 4.3.2, I have a few regressions in tests that compare a type that implements `IEquatable<DateTime>` with an `Is.EqualTime(DateTime)` constraint which fails, and the `EqualTo(DateTime)` method is not called on the instance.

It seems to be due to #4882 (#4877) which adds a custom constraint implementation `EqualTimeBaseConstraint<T>`, which only tests if `actual` is a `T` (`DateTime` or `DateTimeOffset` in this case), but does not test if `TActual` implements `IEquatable<DateTime>` or `IEquatable<DateTimeOffset>` and returns false. Looking at other custom constraints, for example `EqualsStringWithoutUsingConstraints`, it tests for `IEqualtable<string>` as well.

Here is a small repro. The test passes in 4.2.2, but fails in 4.3.2. Neither `Equals(object)` nor `Equals(DateTime)` are called on `DateTimeLike` in 4.3.2.
```c#

public sealed class DateTimeLike : IEquatable<DateTime>
{

	public DateTime Value { get; init; }

	public bool Equals(DateTime other) => other.Equals(this.Value);

	public override bool Equals(object? obj) => obj is DateTime dt && Equals(dt);

	public override int GetHashCode() => this.Value.GetHashCode();

	public override string ToString() => this.Value.ToString(CultureInfo.InvariantCulture);

}

[TestFixture]
public class DateTimeEqualityFacts
{

	[Test]
	public void Test_IEquatableOfDateTime_Is_EqualTo()
	{
		var now = DateTime.Now;
		var item = new DateTimeLike { Value = now };

		Assert.That(item, Is.EqualTo(now));
		// 4.2.2: PASS
		// 4.3.2: FAIL
	}

}
```
Result:
```
  Failed: Assert.That(item, Is.EqualTo(now))
  Expected: 2025-03-07 11:37:37.0200461
  But was:  <03/07/2025 11:37:37>
```

manfred-brands:
Fixes #4954 

This PR also adds a reciproke test to EqualConstraint:

EqualConstraint now tries both x.Equals(y) and y.Equals(x)

This matches the behaviour of EquatablesConstraint which tries both:
x.IEquatable<y> and y.IEquatable<x>

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/EqualsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/EqualsComparer.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System;
-using System.Reflection;
 using NUnit.Framework.Internal;
 
 namespace NUnit.Framework.Constraints.Comparers
@@ -26,8 +25,8 @@ namespace NUnit.Framework.Constraints.Comparers
                 return EqualMethodResult.TypesNotSupported;
             }
 
-            bool xOverridesEqualsObject = OverridesEqualsObject(xType);
-            bool yOverridesEqualsObject = OverridesEqualsObject(yType);
+            bool xOverridesEqualsObject = TypeHelper.OverridesEqualsObject(xType);
+            bool yOverridesEqualsObject = TypeHelper.OverridesEqualsObject(yType);
 
             if (xOverridesEqualsObject || yOverridesEqualsObject)
             {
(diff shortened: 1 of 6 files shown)


OsirisTerje (approved):
Looks good to me, and all tests are passing, and it seems they are covering the code in question.

But I wonder if we should split up the tests,  the EqualConstraintTests are now 1500 lines long.  I was a bit hard to try to find the tests that actually was to be run for this.  There are multiple nested classes, and I am not sure we gain anything by that.  This is unrelated to this PR though, but something we could consider rectoring.

manfred-brands:
@OsirisTerje I agree with you that that file is too big.

In most, but not all cases, it tests the constraint and not what most users use the `Is.EqualTo`.
My bad when I replaced the constraint used by the latter I did not add more new tests as I thought it was tested. I did add the with this PR.

PR #4965, opened 2025-03-14
Ensure Builder is passed to new constraint when replacing constraints

### LINKED ISSUE #4964: Is.Not.EqualTo does not work correctly with Using comparer
If we apply custom comparer with `Using` to `Is.Not.EqualTo` then `Not` is ignored and it starts to work as `Is.EqualTo` with `Using`.

Test below fails but should pass:
```
[Test]
public void UsesProvidedIComparerWithNot()
{
    var comparer = new ObjectComparer();
    Assert.Multiple(() =>
    {
        Assert.That(2 + 3, Is.Not.EqualTo(4).Using(comparer));
        Assert.That(comparer.WasCalled, "Comparer was not called");
    });
}
```

It produces the followong error:
```
Assert.That(2 + 3, Is.Not.EqualTo(4).Using(comparer))
  Expected: 4
  But was:  5
```

If just remove `.Using(comparer)` it starts to work as expected.

Tested with latest NUnit 4.3.2 and .NET 8.

manfred-brands:
Fixes #4964


OsirisTerje (approved):
LGTM

PR #4972, opened 2025-03-23
Improve error message for mismatched tolerance + Use appropriate tolerance for PropertiesComparer

### LINKED ISSUE #4968: Asserts with UsingPropertiesComparer and Tolerance crashes unxpectedly if using TimeSpan and the class contains Properties with different types.
Example:
```C#
    [TestFixture]
    public class RepeatTest
    {
        public class ClassToCheck()
        {
            public int Id { get; set; } = 1;
            public string Name { get; set; } = "Name";
            public DateTimeOffset Now { get; } = DateTimeOffset.UtcNow;
        }


        [Test]
        public void Test()
        {
            var expected = new ClassToCheck();
            var actual = new ClassToCheck();
            Assert.That(actual, Is.EqualTo(expected).UsingPropertiesComparer().Within(TimeSpan.FromSeconds(1)));
        }
    }
```

I expect this test to pass since I'm trying to add a tolerance for DateTime comparison. NUnit, however, tries to apply this tolerance to the int field. As a result, the test fails with the following:
```
Message: 
  System.InvalidCastException : Unable to cast object of type 'System.TimeSpan' to type 'System.IConvertible'.

Stack Trace: 
  Convert.ToInt32(Object value)
  Numerics.AreEqual(Int32 expected, Int32 actual, Tolerance tolerance)
  Numerics.AreEqual(Object expected, Object actual, Tolerance& tolerance)
  NumericsComparer.Equal(Object x, Object y, Tolerance& tolerance, ComparisonState state, NUnitEqualityComparer equalityComparer)
  NUnitEqualityComparer.AreEqual(Object x, Object y, Tolerance& tolerance, ComparisonState state)
  PropertiesComparer.Equal(Object x, Object y, Tolerance& tolerance, ComparisonState state, NUnitEqualityComparer equalityComparer)
  NUnitEqualityComparer.AreEqual(Object x, Object y, Tolerance& tolerance, ComparisonState state)
  NUnitEqualityComparer.AreEqual(Object x, Object y, Tolerance& tolerance)
  EqualConstraint.ApplyTo[TActual](TActual actual)
  Assert.That[TActual](TActual actual, IResolveConstraint expression, NUnitString message, String actualExpression, String constraintExpression)
  RepeatTest.Test() line 26
  RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
  MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)

```

While I understand why it happened, an expected behavior would be to apply tolerance only to the applicable types. It however opens a rabbit hole (what if I want more than one tolerance for different types?). I still think this is a bug, may be it can be handled with a more user friendly error message.

manfred-brands:
Fixes #4968 

@OsirisTerje This PR contain 2 fixes:

1. The generic version of NumericEqualityComparer were calling the object overloads of Numerics.AreEqual due to mismatch of constraints.
2. Checks the type of the Tolerance.Amount before applying it. That way we get a NotSupported instead of a InvalidCastException

The latter fixes the issue #4968 where the user expected the `TimeSpan` tolerance to be only applied to the `DateTime` value because the numeric comparisons now reject the tolerance and the PropertiesComparer will retry that without the tolerance value.

Further it contains 2 commit with different versions of new API to specify multiple tolerance to the PropertiesComparer

1. `c => c.Within<int>(1).Within<DateTimeOffset>(TimeSpan.FromSeconds(1))`
2. `c => c.Within(1).Within(TimeSpan.FromSeconds(1))`

The first requires specifying the type the users want to apply the specified tolerance to.
As there are basically only two types, the seconds allows specifying multiple tolerances and based upon the type applies it to either DateTime like values or numeric values.

The latter option is more compact.
It doesn't allow different tolerances between floating and fixed point. For most practical tolerances (close to zero) the integer tolerance will be zero.
It doesn't allow different tolerance between `DateTime(Offset)` and `TimeSpan`. Not sure if that matters.

Let me know which option you prefer.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/DateTimeOffsetsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/DateTimeOffsetsComparer.cs
@@ -16,9 +16,13 @@ namespace NUnit.Framework.Constraints.Comparers
 
             bool result;
 
-            if (tolerance?.Amount is TimeSpan amount)
+            if (tolerance.Amount is TimeSpan amount)
             {
                 result = (xOffset - yOffset).Duration() <= amount;
+            }
+            else if (tolerance.Mode != ToleranceMode.Unset)
+            {
+                return EqualMethodResult.ToleranceNotSupported;
             }
             else
             {
--- a/src/NUnitFramework/framework/Constraints/Comparers/NumericsComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/NumericsComparer.cs
@@ -12,6 +12,9 @@ namespace NUnit.Framework.Constraints.Comparers
             if (!Numerics.IsNumericType(x) || !Numerics.IsNumericType(y))
                 return EqualMethodResult.TypesNotSupported;
 
+            if (!Numerics.IsNumericType(tolerance.Amount))
+                return EqualMethodResult.ToleranceNotSupported;
+
             return Numerics.AreEqual(x, y, ref tolerance) ?
                 EqualMethodResult.ComparedEqual : EqualMethodResult.ComparedNotEqual;
         }
--- a/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
@@ -93,11 +93,16 @@ namespace NUnit.Framework.Constraints.Comparers
 
                 Tolerance toleranceToApply = tolerance;
 
-                if (configuration.ToleranceForType is not null &&
-                    xPropertyValue is not null &&
-                    configuration.ToleranceForType.TryGetValue(xPropertyValue.GetType(), out Tolerance? typeTolerance))
+                if (tolerance.IsUnsetOrDefault)
                 {
-                    toleranceToApply = typeTolerance;
+                    if (xPropertyValue is TimeSpan or DateTime or DateTimeOffset)
+                    {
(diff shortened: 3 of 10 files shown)


OsirisTerje:
Awesome work!  

I like the terser syntax better, but since it can express less, then .......   Argh...   Is it possible to allow for both syntax'es ?

If not, I still think I would stay with 2. 

If I specify Within(0.6) would that translate to 0 or 1 for ints ?

manfred-brands:
Thanks @OsirisTerje 

> If I specify Within(0.6) would that translate to 0 or 1 for ints ?

I thought it was truncating, but it is ["rounding"](https://learn.microsoft.com/en-us/dotnet/api/system.convert.toint32?view=net-9.0#system-convert-toint32(system-double)):

> value, rounded to the nearest 32-bit signed integer. If value is halfway between two whole numbers, the even number is returned; that is, 4.5 is converted to 4, and 5.5 is converted to 6.

I am now allowing `Within(fixed)` and `Within(floating)` and `Within(TimeSpan)`
The tolerance value picked now depends on the type of the property at hand.
The only thing is that `Within(1)` will not apply to floating point numbers you need to use `Within(1.0)`.
I initially had fallbacks, but that would mean one has to use `Within(1).Within(0.0)` if the tolerance is only for integers and vv.


This now will give enough flexibility that I don't think we need option 1.

OsirisTerje (approved):
Lgtm

PR #4976, opened 2025-04-12
Implement NormalizeLineEndings

### LINKED ISSUE #4975: Support ignoring line ending format when comparing strings
## Context
When you compare multi-line strings in NUnit test code, you often need to normalize line endings beforehand. A simple `.Replace()` will do, but:
1. it's annoying to put it everywhere.
2. the string replacement causes unnecessary memory allocations even when the string doesn't contain any newlines.

## Proposal
A more user-friendly option would be handle this at the framework level, similar to the `IgnoreCase` modifier or the `IgnoreWhiteSpace` modifier which exist for some constraints.

## Example
```cs
[Test]
public void TestIgnoringLineEndingFormat(string actual)
{
    const string expected = """
        multi-line
        raw
        string
        literal
        """;
    Assert.That(actual, Is.EqualTo(expected).IgnoreLineEndingFormat);
}
```

Bartleby2718:
This fixes #4975.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
@@ -36,14 +36,19 @@ namespace NUnit.Framework.Constraints
         #region Protected Properties
 
         /// <summary>
-        /// Get a flag indicating whether the user requested us to ignore case.
+        /// Get a flag indicating whether the user requested that we ignore case.
         /// </summary>
         protected bool IgnoringCase => _comparer.IgnoreCase;
 
         /// <summary>
-        /// Get a flag indicating whether the user requested us to ignore white space.
+        /// Get a flag indicating whether the user requested that we ignore white space.
         /// </summary>
         protected bool IgnoringWhiteSpace => _comparer.IgnoreWhiteSpace;
+
+        /// <summary>
+        /// Get a flag indicating whether the user requested that we normalize newlines.
+        /// </summary>
+        protected bool NormalizingLineEndings => _comparer.NormalizeLineEndings;
 
         /// <summary>
         /// Get a flag indicating whether any external comparers are in use.
@@ -74,6 +79,18 @@ namespace NUnit.Framework.Constraints
             get
             {
                 _comparer.IgnoreWhiteSpace = true;
+                return this;
+            }
+        }
+
+        /// <summary>
+        /// Flag the constraint to normalize newlines and return self.
+        /// </summary>
+        public CollectionItemsEqualConstraint NormalizeLineEndings
+        {
+            get
+            {
+                _comparer.NormalizeLineEndings = true;
                 return this;
             }
         }
--- a/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
@@ -48,7 +48,7 @@ namespace NUnit.Framework.Constraints
         /// <summary>
         /// Get a flag indicating whether the user requested that line ending format be ignored.
         /// </summary>
-        protected bool NormalizingLineEndings => _comparer.NormalizeLineEndings;
+        protected bool IgnoringLineEndingFormat => _comparer.IgnoreLineEndingFormat;
 
         /// <summary>
         /// Get a flag indicating whether any external comparers are in use.
@@ -84,13 +84,13 @@ namespace NUnit.Framework.Constraints
         }
 
         /// <summary>
-        /// Flag the constraint to normalize newlines and return self.
+        /// Flag the constraint to ignore line ending format (\r vs. \n vs. \r\n) and return self.
         /// </summary>
-        public CollectionItemsEqualConstraint NormalizeLineEndings
+        public CollectionItemsEqualConstraint IgnoreLineEndingFormat
         {
             get
             {
-                _comparer.NormalizeLineEndings = true;
+                _comparer.IgnoreLineEndingFormat = true;
                 return this;
             }
         }
--- a/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
@@ -36,17 +36,17 @@ namespace NUnit.Framework.Constraints
         #region Protected Properties
 
         /// <summary>
-        /// Get a flag indicating whether the user requested that we ignore case.
+        /// Get a flag indicating whether the user requested that case be ignored.
         /// </summary>
         protected bool IgnoringCase => _comparer.IgnoreCase;
 
         /// <summary>
-        /// Get a flag indicating whether the user requested that we ignore white space.
+        /// Get a flag indicating whether the user requested that white space be ignored.
         /// </summary>
         protected bool IgnoringWhiteSpace => _comparer.IgnoreWhiteSpace;
 
         /// <summary>
-        /// Get a flag indicating whether the user requested that we normalize newlines.
+        /// Get a flag indicating whether the user requested that line ending format be ignored.
         /// </summary>
         protected bool NormalizingLineEndings => _comparer.NormalizeLineEndings;
 
--- a/src/NUnitFramework/framework/Constraints/Comparers/LineEndingNormalizingStringComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/LineEndingNormalizingStringComparer.cs
@@ -1,0 +1,184 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace NUnit.Framework.Constraints.Comparers;
+
+/// <summary>
+/// Normalizes different line endings (\r, \n, and \r\n) to \n before comparig strings.
+/// </summary>
+internal sealed class LineEndingNormalizingStringComparer : IEqualityComparer<string>
+{
+    /// <summary>
+    /// Gets a <see cref="LineEndingNormalizingStringComparer"/> that performs case-sensitive comparison using the rules
+    /// of the current culture after normalizing line endings.
+    /// </summary>
+    public static readonly LineEndingNormalizingStringComparer CurrentCulture = new(StringComparison.CurrentCulture);
+
+    /// <summary>
+    /// Gets a <see cref="LineEndingNormalizingStringComparer"/> that performs case-insensitive comparison using the rules
+    /// of the current culture after normalizing line endings.
+    /// </summary>
+    public static readonly LineEndingNormalizingStringComparer CurrentCultureIgnoreCase = new(StringComparison.CurrentCultureIgnoreCase);
+
+    /// <summary>
+    /// Gets a <see cref="LineEndingNormalizingStringComparer"/> that performs case-sensitive comparison using the rules
+    /// of the invariant culture after normalizing line endings.
+    /// </summary>
+    public static readonly LineEndingNormalizingStringComparer InvariantCulture = new(StringComparison.InvariantCulture);
+
+    /// <summary>
+    /// Gets a <see cref="LineEndingNormalizingStringComparer"/> that performs case-insensitive comparison using the rules
+    /// of the invariant culture after normalizing line endings.
+    /// </summary>
+    public static readonly LineEndingNormalizingStringComparer InvariantCultureIgnoreCase = new(StringComparison.InvariantCultureIgnoreCase);
+
+    /// <summary>
+    /// Gets a <see cref="LineEndingNormalizingStringComparer"/> that performs a case-sensitive ordinal comparison
+    /// after normalizing line endings.
+    /// </summary>
+    public static readonly LineEndingNormalizingStringComparer Ordinal = new(StringComparison.Ordinal);
+
+    /// <summary>
+    /// Gets a <see cref="LineEndingNormalizingStringComparer"/> that performs a case-insensitive ordinal comparison
+    /// after normalizing line endings.
+    /// </summary>
+    public static readonly LineEndingNormalizingStringComparer OrdinalIgnoreCase = new(StringComparison.OrdinalIgnoreCase);
+
+    private readonly StringComparison _comparisonType;
+
+    private LineEndingNormalizingStringComparer(StringComparison comparisonType)
+    {
+        _comparisonType = comparisonType;
+    }
+
+    public bool Equals(string? x, string? y)
+    {
+        if (ReferenceEquals(x, y))
+            return true;
+        if (x is null || y is null)
+            return false;
+
+        int xIndex = 0;
+        int yIndex = 0;
+
+        while (xIndex < x.Length && yIndex < y.Length)
+        {
+            // handle x first
+            char xChar = x[xIndex];
+            bool xIsNewLine = false;
+            if (xChar == '\r')
+            {
+                xIsNewLine = true;
+                // also handle \r\n
+                if (xIndex + 1 < x.Length && x[xIndex + 1] == '\n')
+                    xIndex++;
+            }
+            else if (xChar == '\n')
+            {
+                xIsNewLine = true;
+            }
+
+            // handle y next
+            char yChar = y[yIndex];
+            bool yIsNewLine = false;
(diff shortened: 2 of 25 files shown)


src/NUnitFramework/framework/Constraints/Comparers/LineEndingNormalizingStringComparer.cs
```diff
@@ -0,0 +1,184 @@
...
+using System.Text;
+
+namespace NUnit.Framework.Constraints.Comparers;
+
+/// <summary>
+/// Normalizes different line endings (\r, \n, and \r\n) to \n before comparig strings.
+/// </summary>
+internal sealed class LineEndingNormalizingStringComparer : IEqualityComparer<string>
+{
+    /// <summary>
```
Bartleby2718, line 12: Only two of these are used, but I just created singletons for all 6 `StringComparison` values.

src/NUnitFramework/framework/Constraints/EqualConstraint.cs
```diff
@@ -471,6 +491,9 @@ public override string Description
                 if (_comparer.IgnoreWhiteSpace)
                     sb.Append(", ignoring white-space");
 
+                if (_comparer.IgnoreWhiteSpace)
```
Bartleby2718, line 494: I guess I'll have to fix this copypasta. I can do this after the first round of reviews.

src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
```diff
@@ -31,7 +32,14 @@ public static bool Equals(string x, string y, bool ignoreCase, bool ignoreWhiteS
             }
             else
             {
-                return x.Equals(y, ignoreCase ? StringComparison.CurrentCultureIgnoreCase : StringComparison.Ordinal);
+                IEqualityComparer<string> comparer = (ignoreCase, NormalizeLineEndings) switch
+                {
+                    (true, true) => LineEndingNormalizingStringComparer.CurrentCultureIgnoreCase,
```
Bartleby2718, line 37: `CurrentCultureIgnoreCase` and `Ordinal` to be consistent with the existing code.

src/NUnitFramework/tests/TestUtilities/Collections/SimpleObjectCollection.cs
```diff
@@ -24,6 +24,8 @@ public SimpleObjectCollection(params object?[] source)
             _contents = new List<object?>(source);
         }
 
+        public override string ToString() => $"SimpleObjectCollection [{string.Join(", ", _contents)}]";
```
Bartleby2718, line 27: Updated because I found Test Explorer misleading without this.

src/NUnitFramework/framework/Constraints/UniqueItemsConstraint.cs
```diff
@@ -253,17 +253,19 @@ private sealed class NUnitStringEqualityComparer : IEqualityComparer<string>
 
             private readonly bool _ignoreCase;
             private readonly bool _ignoreWhiteSpace;
+            private readonly bool _NormalizeLineEndings;
```
Bartleby2718, line 256:
```suggestion
            private readonly bool _normalizeLineEndings;
```

src/NUnitFramework/framework/Constraints/EqualConstraint.cs
Bartleby2718, line 494: I guess I'll have to fix this copypasta. I can do this after the first round of reviews.
```suggestion
                if (_comparer.NormalizeLineEndings)
```

src/NUnitFramework/framework/Constraints/EqualStringWithoutUsingConstraint.cs
```diff
@@ -20,6 +20,7 @@ public class EqualStringWithoutUsingConstraint : Constraint
 
         private bool _caseInsensitive;
         private bool _ignoringWhiteSpace;
+        private bool _NormalizeLineEndings;
```
Bartleby2718, line 23: I guess I'll have to fix this copypasta. I can do this after the first round of reviews.
```suggestion
        private bool _normalizeLineEndings;
```

src/NUnitFramework/framework/Constraints/ContainsConstraint.cs
```diff
@@ -18,6 +18,7 @@ public class ContainsConstraint : Constraint
         private Constraint? _realConstraint;
         private bool _ignoreCase;
         private bool _ignoreWhiteSpace;
+        private bool _NormalizeLineEndings;
```
Bartleby2718, line 21: I guess I'll have to fix this copypasta. I can do this after the first round of reviews.

```suggestion
        private bool _normalizeLineEndings;
```

src/NUnitFramework/framework/Constraints/NUnitEqualityComparer.cs
```diff
@@ -63,6 +63,11 @@ public sealed class NUnitEqualityComparer
         /// </summary>
         private bool _ignoreWhiteSpace;
 
+        /// <summary>
+        /// If true, all string comparisons will be done after normalizing newlines
+        /// </summary>
+        private bool _NormalizeLineEndings;
```
Bartleby2718, line 69: I guess I'll have to fix this copypasta. I can do this after the first round of reviews.

```suggestion
        private bool _normalizeLineEndings;
```

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
```diff
@@ -36,15 +36,20 @@ protected CollectionItemsEqualConstraint(object? arg) : base(arg)
         #region Protected Properties
 
         /// <summary>
-        /// Get a flag indicating whether the user requested us to ignore case.
+        /// Get a flag indicating whether the user requested that we ignore case.
```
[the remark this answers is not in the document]
manfred-brands: Us/We as the software?
What about:
`Get a flag indicating whether the user requested to ignore case.`

src/NUnitFramework/framework/Constraints/Comparers/StringsComparer.cs
```diff
@@ -17,12 +18,12 @@ public static EqualMethodResult Equal(object x, object y, ref Tolerance toleranc
...
                 return EqualMethodResult.ToleranceNotSupported;
 
-            return Equals(xString, yString, equalityComparer.IgnoreCase, equalityComparer.IgnoreWhiteSpace) ?
+            return Equals(xString, yString, equalityComparer.IgnoreCase, equalityComparer.IgnoreWhiteSpace, equalityComparer.NormalizeLineEndings) ?
                 EqualMethodResult.ComparedEqual :
                 EqualMethodResult.ComparedNotEqual;
         }
 
-        public static bool Equals(string x, string y, bool ignoreCase, bool ignoreWhiteSpace)
+        public static bool Equals(string x, string y, bool ignoreCase, bool ignoreWhiteSpace, bool NormalizeLineEndings)
```
manfred-brands, line 26: Use camelCase for parameter names.
Did this not raise IDE1006 for you?
{93D7E055-CC6C-4BFC-A857-751090DD24EC}

src/NUnitFramework/framework/Constraints/EqualConstraint.cs
```diff
@@ -74,10 +74,18 @@ public EqualConstraint(object? expected)
         /// Gets a value indicating whether to compare ignoring white space.
         /// </summary>
         /// <value>
-        ///   <see langword="true"/> if comparing ignoreing white space; otherwise, <see langword="false"/>.
+        ///   <see langword="true"/> if comparing ignoring white space; otherwise, <see langword="false"/>.
         /// </value>
         public bool IgnoringWhiteSpace => _comparer.IgnoreWhiteSpace;
 
+        /// <summary>
+        /// Gets a value indicating whether to compare after normalizing newlines.
```
manfred-brands, line 82: The code isn't "normalizing newlines" or line endings, but "ignoring different line endings"
Normalize indicates it changes them before comparing which is something not done deliberately.
Maybe rename all "NormalizeLineEndings" into "IgnoreDifferentStyleLineEndings".
It also keeps the naming of the properties consistent where the modifiers ignores an aspect, being it case, or line endings or white space.

src/NUnitFramework/framework/Constraints/EqualConstraint.cs
manfred-brands: It does indicate that the change is not covered in nunit tests.
But neither are the original modifiers so it seems.

src/NUnitFramework/framework/Constraints/EqualStringWithoutUsingConstraint.cs
manfred-brands: Should have shown IDE1006 warning in your editor:
{7F4443CF-5250-4290-BD7F-C848A4AECFFD}

src/NUnitFramework/tests/Constraints/AnyOfConstraintTests.cs
```diff
@@ -44,6 +44,27 @@ public void ItemIsPresent_IgnoreCaseWhiteSpace()
...
+            Assert.That(anyOf.ApplyTo("a\rB\r\n").Status, Is.EqualTo(ConstraintStatus.Success));
+        }
+
+        [Test]
+        public void ItemIsPresent_IgnoreWhiteSpaceNormalizeLineEndings()
+        {
+            var anyOf = new AnyOfConstraint(new[] { "a", "B", "a b" }).IgnoreWhiteSpace;
+            Assert.That(anyOf.ApplyTo("ab").Status, Is.EqualTo(ConstraintStatus.Success));
+        }
+
```
manfred-brands, line 67: You didn't add the `NormalizeLineEndings` modifiers to proof that it doesn't break things.
Even though there are no line endings when white space is ignored.

manfred-brands (changes requested):
Thanks @Bartleby2718 for your contribution.
You already found your casing problem in parameters and fields.
It would have been easier if you had just fixed the instead of putting a comment on your own PR that you need to fix it.

The actual comparer looks good.
The main change I would like is to rename the word "Normalize" and "Normalizing"
One what is normal for one might not be normal for another.
But more to match the other modifiers that it "Ignore" an aspect.
It doesn't change any strings but ignores different style line endings.

src/NUnitFramework/framework/Constraints/EqualStringWithoutUsingConstraint.cs
Bartleby2718: I did a non-case-preserving string replacement globally right before pushing. It does flag that, but it doesn't block a build because [`EnforceCodeStyleInBuild`](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#enforcecodestyleinbuild) is not set.

src/NUnitFramework/framework/Constraints/CollectionItemsEqualConstraint.cs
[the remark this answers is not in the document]
Bartleby2718: Updated to use passive form instead

src/NUnitFramework/framework/Constraints/EqualConstraint.cs
Bartleby2718: We have `IgnoreCase`, not `IgnoreDifferentCases`, so I went with `IgnoreLineEndingFormat`.

src/NUnitFramework/framework/Constraints/EqualStringWithoutUsingConstraint.cs
Bartleby2718: Huh, interestingly https://github.com/nunit/nunit/blob/d70b7eab12c41da68ccc313fd5d3d5abcdf15fa4/src/NUnitFramework/Directory.Build.props#L9 _is_ set.

manfred-brands (changes requested):
Thanks @Bartleby2718. mostly fine.
A few things remain.

src/NUnitFramework/framework/Constraints/StringConstraint.cs
```diff
@@ -29,6 +29,11 @@ public abstract class StringConstraint : Constraint
         protected bool caseInsensitive;
 #pragma warning restore IDE1006
 
+        /// <summary>
+        /// Indicates whether tests should ignore line ending format (\r vs. \n vs. \r\n)
+        /// </summary>
+        protected bool IgnoringLineEndingFormat;
+
```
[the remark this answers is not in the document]
Bartleby2718: @manfred-brands I think I _do_ want to do this (and `IgnoringWhiteSpace`) _but in a separate PR_.

1. Is that okay?
2. If so, do you want me to keep this or remove this _in this PR_?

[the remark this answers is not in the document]
manfred-brands: If not done in this PR, this should be removed.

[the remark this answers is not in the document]
Bartleby2718: Actually, I see that this _is_ used in `ContainsConstraint`. Do you want me to keep this (since there _is_ one usage in https://github.com/nunit/nunit/pull/4976/files#diff-541252d3ddf1d5ca9f75af21eb3c62c40e5d3a00b2f8c7d4cc42919f7d7bca84R110) or remove it (because it's not used in _all_ subclasses)?

src/NUnitFramework/framework/Constraints/MessageWriter.cs
```diff
@@ -109,6 +109,30 @@ public virtual void DisplayStringDifferences(string expected, string actual, int
...
+        /// <param name="mismatchExpected">The point in <paramref name="expected"/> at which the strings don't match or -1</param>
+        /// <param name="mismatchActual">The point in <paramref name="actual"/> at which the strings don't match or -1</param>
+        /// <param name="ignoreCase">If true, case is ignored in locating the point where the strings differ</param>
+        /// <param name="ignoreWhiteSpace">If true, white space is ignored in locating the point where the strings differ</param>
+        /// <param name="ignoreLineEndingFormat">If true, line ending format is ignored in locating the point where the strings differ</param>
+        /// <param name="clipping">If true, the strings should be clipped to fit the line</param>
+        // TODO: consider updating visibility in v5
+        internal virtual void DisplayStringDifferences(string expected, string actual, int mismatchExpected, int mismatchActual, bool ignoreCase, bool ignoreWhiteSpace, bool ignoreLineEndingFormat, bool clipping)
+        {
+            if ((ignoreWhiteSpace || ignoreLineEndingFormat) && mismatchExpected != mismatchActual)
```
Bartleby2718, line 128: Imitated line 104

src/NUnitFramework/framework/Constraints/StringConstraint.cs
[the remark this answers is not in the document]
manfred-brands: It is not, it is set there on the `EqualConstraint` but that is not a subset of the `StringConstraint`.
It is set on the `SubstringConstraint`, but that one doesn't support it.
For now I changed it in the same way as the `IgnoreWhiteSpace` and removed it from `StringConstraint` until we have a case. But even then I prefer to add it to the subclass that uses it instead of the base class.

manfred-brands (approved):
Looking good now.

manfred-brands:
@Bartleby2718 There is a serious performance degradation:

```text
Warning : NUnit.Framework.Tests.Constraints.UniqueItemsConstraintTests.PerformanceTests_FastPath(System.Collections.Generic.List`1[System.String],True)
  Warn.Unless(() =>
            {
                if (ignoreCase)
                else
            }, HelperConstraints.HasMaxTime(100))
Elapsed time of 3696ms exceeds maximum of 100ms
   at NUnit.Framework.Tests.Constraints.UniqueItemsConstraintTests.PerformanceTests_FastPath(IEnumerable values, Boolean ignoreCase) in /_/src/NUnitFramework/tests/Constraints/UniqueItemsConstraintTests.cs:line 186
```

I have traced it down to a subtle line I missed in my review: `comparer.GetHashCode()` which should have been `comparer.GetHashCode(s)`
Putting all items in the same 'bucket' in a HashSet slowed down the test dramatically.

Issue #4861, opened 2024-10-22
Always repeat Test x times


cfuerbachersparks:
Currently only repeat until success `RetryAttribute` and repeat until failure `RepeatAttribute` are possible.
There should be a way to repeat a test always x times, for stress testing a system for example.

This would probably need the refactor mentioned in the `RepeatAttribute` [here](https://github.com/nunit/nunit/blob/a39e8297b4e9ad279679430021af7a83f3091d59/src/NUnitFramework/framework/Attributes/RepeatAttribute.cs#L3)

stevenaw:
I think this is a good idea to increase usability. We might want to consider exposing a property to configure this behavior with a default value compatible with the current behavior. Something like: 

```
[Retry(5, StopOnSuccess = true)]
```

I'm not sure if there's a better name for that property. @cfuerbachersparks @nunit/framework-team thoughts?

manfred-brands:
I think that `[Repeat(5, StopOnFailure=false)]` is a better   option than `Retry` which implies repeat until success.

stevenaw:
I'm not against that idea. I'd been trying to think about if we could avoid making a new attribute for this, though I agree with the sentiment that the `Repeat` name is clearer for this purpose.

Anecdotally, I suspect the "flaky test" use case met by `Retry` may be more common than the "load test" based on my perception of issues filed here, so I don't know about `Repeat` completely replacing `Retry`. My thinking is we'd want to keep something with defaults setup for that use case, in which case perhaps any new `Repeat` attribute wouldn't need a property to configure its behavior. It would always complete `n` times.

cfuerbachersparks:
https://github.com/cfuerbachersparks/nunit/tree/RepeatWithoutStop

moved to logic to the `RepeatAttribute` on a new branch

manfred-brands:
@cfuerbachersparks Can you please put up a PR where you want to merge that branch in your repository into this one.

PR #4979, opened 2025-04-19
Treat an exception as an assertion failure and always add it to the assertion list.

### LINKED ISSUE #4537: AssertException in TearDown hides Exception in Test
I know from other comments in issues that it is not recommended to do Assertions during TearDown. Nevertheless, I want to report the following to check if it is a known issue/by design or a bug.

Simple code example to demonstrate the issue:
```csharp
using NUnit.Framework;

namespace DemoTests; 

public class TearDownTests {
    [Test]
    public void Assertion() {
        Assert.IsTrue(false);
    }

    [Test]
    public void Exception() {
        throw new Exception("Test exception");
    }

    [Test]
    public void Pass() {
        
    }

    [TearDown]
    public void TearDown() {
        Assert.Warn("Warning");
    }
}
```

Result:

- Assertion() tests shows both failures and warnings
- Pass() shows Warning from TearDown
- Exception() only shows Warning and hides/forgets/overrides the Exception

This happens with any AssertException in TearDown. Of course, when I catch the exception and "rethrough" it using something like Assert.Fail(ex.Message), I see both failures as in Assertion() tests, but for me it is one of the big advantages of NUnit that I don't need to catch all my exceptions to let a test fail but that it fails automatically.

manfred-brands:
Fixes #4537 

There is a change in `RepeatAttribute` behaviour:

1. If an exception is thrown and `StopOnFailure` is false, the test continues executing.
2. Test result are reset between runs. So if the last run passes, the test succeeds. Previously it would fail.

The above are side effects of #4861 and only affect the code when `StopOnFailure` is set to false.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/RepeatAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RepeatAttribute.cs
@@ -94,18 +94,33 @@ namespace NUnit.Framework
 
                 while (count-- > 0)
                 {
-                    context.CurrentResult = innerCommand.Execute(context);
+                    try
+                    {
+                        context.CurrentResult = innerCommand.Execute(context);
+                    }
+                    // Commands are supposed to catch exceptions, but some don't
+                    // and we want to look at restructuring the API in the future.
+                    catch (Exception ex)
+                    {
+                        if (context.CurrentResult is null)
+                            context.CurrentResult = context.CurrentTest.MakeTestResult();
+                        context.CurrentResult.RecordException(ex);
+                    }
 
                     if (_stopOnFailure && context.CurrentResult.ResultState != ResultState.Success)
                         break;
 
-                    context.CurrentRepeatCount++;
+                    // Clear result for repeat
+                    if (count > 0)
+                    {
+                        context.CurrentResult = context.CurrentTest.MakeTestResult();
+                        context.CurrentRepeatCount++; // increment Retry count for next iteration. will only happen if we are guaranteed another iteration
+                    }
                 }
 
                 return context.CurrentResult;
             }
         }
-
         #endregion
     }
 }
--- a/src/NUnitFramework/framework/Attributes/RepeatAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/RepeatAttribute.cs
@@ -91,6 +91,7 @@ namespace NUnit.Framework
             public override TestResult Execute(TestExecutionContext context)
             {
                 int count = _repeatCount;
+                TestResult? overallResult = null;
 
                 while (count-- > 0)
                 {
@@ -107,8 +108,13 @@ namespace NUnit.Framework
                         context.CurrentResult.RecordException(ex);
                     }
 
-                    if (_stopOnFailure && context.CurrentResult.ResultState != ResultState.Success)
-                        break;
+                    if (context.CurrentResult.ResultState != ResultState.Success)
+                    {
+                        // If any repeat fails, the whole test is considered failed
+                        overallResult = context.CurrentResult;
+                        if (_stopOnFailure)
+                            break;
+                    }
 
                     // Clear result for repeat
                     if (count > 0)
@@ -116,6 +122,11 @@ namespace NUnit.Framework
                         context.CurrentResult = context.CurrentTest.MakeTestResult();
                         context.CurrentRepeatCount++; // increment Retry count for next iteration. will only happen if we are guaranteed another iteration
                     }
+                }
+
+                if (overallResult is not null)
+                {
+                    context.CurrentResult = overallResult;
                 }
 
                 return context.CurrentResult;
--- a/src/NUnitFramework/framework/Interfaces/AssertionResult.cs
+++ b/src/NUnitFramework/framework/Interfaces/AssertionResult.cs
@@ -30,7 +30,7 @@ namespace NUnit.Framework.Interfaces
         }
 
         /// <summary> Where did the failure occur.</summary>
-        public FailureSite Site;
+        public FailureSite Site { get; }
 
         /// <summary> The pass/fail status of the assertion</summary>
         public AssertionStatus Status { get; }
--- a/src/NUnitFramework/framework/Interfaces/AssertionResult.cs
+++ b/src/NUnitFramework/framework/Interfaces/AssertionResult.cs
@@ -14,11 +14,23 @@ namespace NUnit.Framework.Interfaces
         /// Construct an AssertionResult
         /// </summary>
         public AssertionResult(AssertionStatus status, string message, string? stackTrace)
+            : this(FailureSite.Test, status, message, stackTrace)
         {
+        }
+
+        /// <summary>
+        /// Construct an AssertionResult
+        /// </summary>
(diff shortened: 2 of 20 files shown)


src/NUnitFramework/framework/Attributes/RepeatAttribute.cs
```diff
@@ -94,18 +94,33 @@ public override TestResult Execute(TestExecutionContext context)
...
-                    context.CurrentResult = innerCommand.Execute(context);
+                    try
+                    {
+                        context.CurrentResult = innerCommand.Execute(context);
+                    }
+                    // Commands are supposed to catch exceptions, but some don't
+                    // and we want to look at restructuring the API in the future.
+                    catch (Exception ex)
+                    {
+                        if (context.CurrentResult is null)
```
OsirisTerje, line 105: CurrentResult is declared as non-nullable.
But TestExecutionContext doesn't explicitly set this to non-null.
Also see the comment on l.52 workitem.cs.  So, perhaps just accept this.......

src/NUnitFramework/framework/Interfaces/AssertionResult.cs
```diff
@@ -14,12 +14,24 @@ public class AssertionResult : IEquatable<AssertionResult>
...
+        public AssertionResult(FailureSite site, AssertionStatus status, string message, string? stackTrace)
+        {
+            Site = site;
             Status = status;
             Message = message;
             StackTrace = stackTrace;
         }
 
+        /// <summary> Where did the failure occur.</summary>
+        public FailureSite Site;
```
OsirisTerje, line 33:
```suggestion
        public FailureSite Site { get; }
```

src/NUnitFramework/framework/Internal/Commands/OneTimeTearDownCommand.cs
```diff
@@ -22,19 +20,7 @@ public OneTimeTearDownCommand(TestCommand innerCommand, SetUpTearDownItem setUpT
...
-            {
-                TestResult suiteResult = context.CurrentResult;
-
-                try
-                {
-                    setUpTearDownItem.RunTearDown(context);
-                }
-                catch (Exception ex)
-                {
-                    suiteResult.RecordTearDownException(ex);
```
OsirisTerje, line 35: Why should not this be run?

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -480,23 +471,54 @@ public void RecordException(Exception ex, FailureSite site)
...
+        public void RecordException(Exception ex, FailureSite site)
+        {
+            var result = new ExceptionResult(ex, site);
 
-            ResultState resultState = ResultState == ResultState.Cancelled
-                ? ResultState.Cancelled
-                : ResultState.Error;
-            if (Test.IsSuite)
-                resultState = resultState.WithSite(FailureSite.TearDown);
+            if (ex is NUnitException nUnitException && nUnitException.InnerException is ResultStateException ||
```
OsirisTerje, line 486:
```suggestion
          if (ex is NUnitException { InnerException: ResultStateException } or ResultStateException)
```

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -480,23 +471,54 @@ public void RecordException(Exception ex, FailureSite site)
...
+                ex is ResultStateException)
```
OsirisTerje, line 487:
```suggestion
              
```

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -676,6 +703,27 @@ private ResultState AssertionStatusToResultState(AssertionStatus status)
             }
         }
 
+        private AssertionStatus ResultStateToAssertionStatus(ResultState resultState)
+        {
+            if (resultState.Status == TestStatus.Failed)
+            {
+                if (resultState.Label == ResultState.Error.Label)
+                    return AssertionStatus.Error;
+                else if (resultState.Label == ResultState.Failure.Label)
```
OsirisTerje, line 712:
```suggestion
                if (resultState.Label == ResultState.Failure.Label)
```

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -676,6 +703,27 @@ private ResultState AssertionStatusToResultState(AssertionStatus status)
             }
         }
 
+        private AssertionStatus ResultStateToAssertionStatus(ResultState resultState)
+        {
+            if (resultState.Status == TestStatus.Failed)
```
OsirisTerje, line 708: Could perhaps be a switch statement

OsirisTerje (changes requested):
Just commented on some smaller coding stuff.  
The change seems to hit  rather large, but seems to be well covered by the tests.

src/NUnitFramework/framework/Interfaces/AssertionResult.cs
manfred-brands: Yes, thanks.

src/NUnitFramework/framework/Internal/Commands/OneTimeTearDownCommand.cs
manfred-brands: Because the exceptions are already caught in the `RunTearDown` method hence it would never get into the `catch` part.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
manfred-brands: If the user executes `Assert.Fail/Ignore/Success` in a SetUp or TearDown it will get here with a (derived) `ResultStateException`

src/NUnitFramework/framework/Internal/Results/TestResult.cs
manfred-brands: ok.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
manfred-brands: ok

src/NUnitFramework/framework/Attributes/RepeatAttribute.cs
manfred-brands: TestExecutionContext its properties is fake nullable. With a TODO:
```csharp
// TODO: Fix design where properties are not set at unknown times.

#pragma warning disable CS8618 // Non-nullable field must contain a non-null value when exiting constructor. Consider declaring as nullable.
```

It is literally copied from `RetryCommand.Execute`.
Looking at the code above, it assigns the result of the command to `context.CurrentResult`.
If the command fails, it is not assigned, unless it is assigned somewhere else previously which we cannot guarantee.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
manfred-brands: Not used to the fancy nested pattern, but I like it.

OsirisTerje (approved):

PR #4980, opened 2025-04-19
Add the missing Tests suffixes in some test files

Bartleby2718:
### Context
It looks like it is a convention in this repository to add the `Tests` suffix for test files, but some test files were missing the suffix.

### Changes
Renaming some files so that your search result becomes more accurate.


mikkelbu (approved):
LGTM

stevenaw (approved):
LGTM, thanks for noticing this and contributing a fix @Bartleby2718

PR #4997, opened 2025-06-13
Chore/update NUnit.Analyzer

manfred-brands:
Updating the NUnit.Analyzer to 4.9.1 to ensure we don't see any unexpected issues.


mikkelbu (approved):
Looks good to me - thanks for the bump and the corrections @manfred-brands

PR #4999, opened 2025-06-21
Issue 4998 Proposed Solution

### LINKED ISSUE #4998: `PropertyConstraint` Causes Unexpected Behavior Because the Property's Value Is `object` Instead of the Property’s Actual Type
### Overview:

### Example:

```csharp
[Test]
public void Example()
{
    var actual = new Foo
    {
        Bar = 0
    };

    Assert.Multiple(() =>
    {
        // passes

        // fails
    });
}
```
```
But was:  0
```

### Proposed Solution:

LeQuackers:
This is the proposed solution to resolve #4998

### DIFF
--- a/src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/PropertyConstraint.cs
@@ -1,6 +1,8 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System;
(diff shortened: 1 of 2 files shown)


LeQuackers:
@dotnet-policy-service agree

manfred-brands:
Thanks @LeQuackers 
Although your PR might work, it only address on specific entry point into `Default`.
I therefore going to close this PR in favour of mine that addresses the problem in the `DefaultConstraint`  itself therefore also solving other use cases.

PR #5000, opened 2025-06-21
In DefaultConstraint ignore TActual if it is object.

### LINKED ISSUE #4998: `PropertyConstraint` Causes Unexpected Behavior Because the Property's Value Is `object` Instead of the Property’s Actual Type

manfred-brands:
Fixes #4998 

There is one change of behaviour from previous:

```csharp
object instance = 0;
Assert.That(instance, Is.Not.Default);
```

Previously this passed because `instance != null`, now it fails because `object` is ignored.
Instead the underlying type is checked and `0 == default(int)`

### DIFF
--- a/src/NUnitFramework/framework/Constraints/DefaultConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/DefaultConstraint.cs
@@ -1,5 +1,6 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
+using System;
 using System.Runtime.CompilerServices;
 
 namespace NUnit.Framework.Constraints
@@ -17,7 +18,25 @@ namespace NUnit.Framework.Constraints
         /// </summary>
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
-            var isDefault = RuntimeHelpers.Equals(actual, default(TActual));
+            if (actual is null)
+            {
+                // Null is always considered the default value.
+                return new ConstraintResult(this, actual, true);
+            }
+
+            // We ignore TActual if it is object as that type is used a lot in non-generic contexts.
+            Type actualType = typeof(TActual) == typeof(object) ? actual.GetType() : typeof(TActual);
+            if (!actualType.IsValueType)
+            {
+                // For reference types, as actual is not null, it cannot be default.
+                return new ConstraintResult(this, actual, false);
+            }
+
+            object? defaultValue = typeof(TActual) == typeof(object)
+                ? Activator.CreateInstance(actualType) // Create a default instance of the actual type.
+                : default(TActual);
+
+            var isDefault = RuntimeHelpers.Equals(actual, defaultValue);
             return new ConstraintResult(this, actual, isDefault);
         }
     }
--- a/src/NUnitFramework/tests/Constraints/DefaultConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/DefaultConstraintTests.cs
@@ -23,11 +23,11 @@ namespace NUnit.Framework.Tests.Constraints
         [Test] public void Success_NullableInt() => AssertSuccess(default(int?));
         [Test] public void Success_Long() => AssertSuccess(default(long));
         [Test] public void Success_ImmutableArray() => AssertSuccess(default(ImmutableArray<int>));
+        [Test] public void Success_BoxedInt() => AssertSuccess((object)default(int));
 
         [Test] public void Failure_NewObject() => AssertFailure(new object());
         [Test] public void Failure_EmptyString() => AssertFailure(string.Empty);
         [Test] public void Failure_NullableInt_Zero() => AssertFailure((int?)0);
-        [Test] public void Failure_BoxedInt_Zero() => AssertFailure((object)0);
         [Test] public void Failure_Int_NonZero() => AssertFailure(1);
         [Test] public void Failure_Double_NaN() => AssertFailure(double.NaN);
         [Test] public void Failure_ImmutableArray_Empty() => AssertFailure(ImmutableArray.Create<int>());
@@ -47,6 +47,13 @@ namespace NUnit.Framework.Tests.Constraints
         {
             var constraintResult = TheConstraint.ApplyTo(actual);
             Assert.That(!constraintResult.IsSuccess);
+        }
+
+        [Test]
+        public void EqualDefault()
+        {
+            var instance = 0;
+            Assert.That(instance, Is.InstanceOf<int>().And.Default);
         }
     }
 }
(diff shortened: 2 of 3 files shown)


mikkelbu:
> Still, the fact that we had a test for the old behavior makes it seem as if the old behavior were expected. I haven't checked the history yet but do we know why that test was the way it was?

The test was added along with the constraint in nunit#3781. There were a little discussion in both the PR and the issue - most relevant https://github.com/nunit/nunit/issues/3457#issuecomment-787530002

stevenaw:
Thanks @mikkelbu 
Having experimented a bit with this in the code I think I can see where @jnm2 was coming from there. We've always tried to have consistency with our equality operations and those of the BCL - or at least if it's equals in the BCL is should be equal in NUnit too. I also think of the `DefaultConstraint` as checking equality against `default`.

I think the change to automatically unbox a value here could introduce an inconsistency in that regard. Consider the following test where the latter two fail because the equality check for an `object` is against null.
```csharp
[Test]
public void Test1()
{
    using (Assert.EnterMultipleScope())
    {
        var x = 0;
        object y = x;

        Assert.That(x == default);
        Assert.That(x == 0);
        Assert.That(y == default);
        Assert.That(y == null);
    }
}
```

Also of note here is that changing `y == default` to `y == default(int)` introduces a compiler error and requires that `y` also be cast to an `int` for the compiler to accept it. To me, this is similar to @jnm2 's suggestion that the generic `<object>` overload use null as the default value and that a cast to _explicitly_ unbox the value is required to get value semantics here

manfred-brands:
NUnit is full of `object`, basically anything that is not generic is `object`, regardless what it was previously.

Note the different result between:
```csharp
[Test]
public void EmptyStringHasDefaultLength()
{
    string s = string.Empty;

    using (Assert.EnterMultipleScope())
    {
        Assert.That(s.Length, Is.Default);
        Assert.That(s, Has.Length.Default);
    }
}
```

The first passes, the second fails with because regardless of property type, the `TActual` is `object`.
```text
  Assert.That(s, Has.Length.Default)
  Expected: property Length default
  But was:  0
```

Following the reasoning that if a instance is assigned to a variable of type object changes it behaviour, the below should fail as `object` does not have a `Length` property, but it passes:

```csharp
[Test]
public void StringHasLength()
{
    string s = string.Empty;
    object o = s;

    Assert.That(o, Has.Length);
}
```

There are other places where we ignore `TActual`:
* `EmptyConstraint` does `Type actualType = actual?.GetType() ?? typeof(TActual);`
* `TypeConstraint` does `actualType = actual?.GetType();`

Dreamescaper:
The current behavior was intended when `DefaultConstraint` was introduced. Please take a look at this comment and related discussions:
https://github.com/nunit/nunit/issues/3457#issuecomment-787484525

Dreamescaper:
I think this PR introduces additional inconsistency around nullable types:
```
int? number = 0;
object boxed = number;

Assert.That(boxed, Is.Default); // passes
Assert.That(number, Is.Default); // fails
```

manfred-brands:
> I think this PR introduces additional inconsistency around nullable types:
> 
> ```
> int? number = 0;
> object boxed = number;
> 
> Assert.That(boxed, Is.Default); // passes
> Assert.That(number, Is.Default); // fails
> ```

Agreed, but that it not much different from the existing inconsistency:

```csharp
int number = 0;
object boxed = number;

Assert.That(number, Is.Default); // passes
Assert.That(boxed, Is.Default); // fails;
```

The same value assigned to a different type changes behaviour.
Yes, technically `0` is not the default of `object` and I have agreed with the existing behaviour in the past.

If it wasn't for `object` being used for a lot in nunit itself.

* The OP problem of `PropertyConstraint` where `TActual` is `object` regardless of property type. 
* The `AllItemsConstraint`, `NoItemConstraint` where object is the type used for any member of a collection,
despite the user code declaring the property correctly.
```csharp
int[] items = [0, 0];
Assert.That(items[0], Is.Default);
Assert.That(items[1], Is.Default);
Assert.That(items, Has.All.Default); // fails
Assert.That(items, Has.None.Default); // passes
```

* The `ExactCountConstraint` and `SomeItemsConstraint``:
```csharp
int[] items = [0, 1];
Assert.That(items[0], Is.Default);
Assert.That(items[1], Is.Not.Default);
Assert.That(items, Has.One.Default); // fails
Assert.That(items, Has.Some.Default); // fails
```

Almost every use of the `Default` property in the `ConstraintExpression` doesn't work, except for `Not.Default`.
Should we drop `Default` from there so we only support `Is.Default` where default matches the declared TActual?

So which approach is the lesser of two evals?

stevenaw:
@manfred-brands I like the current behavior myself, FWIW. It's more generic-friendly, which is the direction we increasingly want to take the constraints.

I find it also makes it easy to opt into value semantics for the assertions that require it by just casting to unbox the value.

```
object x = 0;
Assert.That((int)x, Is.Default);
```

LeQuackers:
After seeing @stevenaw's / @jnm2's example the design behind `DefaultConstraint` makes sense in order to match BCL behavior, but as @manfred-brands pointed out there are lots of constraints that cast the actual value to an `object`, which ends up causing inconsistencies / unexpected behavior.

As an outside observer that uses NUnit, I'd expect these to pass:
```csharp
string s = string.Empty;
Assert.That(s, Has.Length.Default);

int[] items = [0, 1];
Assert.That(items, Has.One.Default);
Assert.That(items, Has.Some.Default);
```

Just spitballing some ideas:

1. Make `PropertyConstraint`, `AllItemsConstraint`, `SomeItemsConstraint`, `ExactCountConstraint`, etc., invoke `BaseConstraint.ApplyTo<T>(T)` with the actual value's type instead of `object`. It's messy, uses reflection, and is probably more work, but `DefaultConstraint` would still match BCL behavior and the base constraints would receive the correct type argument (which would allow the above example to pass), so it seems like a reasonable compromise.

2. Allow the actual value to be cast back to the expected type in the constraint expression somehow, e.g. having an `As(type)/Cast(type)` operator. I honestly don't like this approach because I'd expect the constraint expression to already know what type the property, item, etc. should be, but that's just my opinion. Example:
```csharp
// All of these would be expected to pass:

string s = string.Empty;
Assert.That(s, Has.Length.Cast(typeof(int)).Default);

int[] items = [0, 1];
Assert.That(items, Has.One.Cast(typeof(int)).Default);
Assert.That(items, Has.Some.Cast(typeof(int)).Default);

int number = 0;
object boxed = number;
Assert.That(number, Is.Default);
Assert.That(boxed, Is.Cast(typeof(int)).Default);
```

OsirisTerje:
This PR fixes #4998 but PR #5003 is also a fix for the same.  
@manfred-brands Which one should be merged ?

manfred-brands:
@OsirisTerje It seems that ignoring `object` is not the preferred solution, so I'm closing this.

PR #5001, opened 2025-06-21
Add attribute telling the debugger not to evaluate destructive properties

### LINKED ISSUE #4994: Fix or remove ToString in class ConstraintExpression
The ToString method Resolve the expression, this is a sideeffect, and causes the code to not work correctly when debugging - if you don't turn off Property evaluation (which is default on).

The ToString is not used in the code, it was used for legacy debugging back in the days.

It should either be fixed to not have any sideeffects (but not sure what it should display), or dimply be removed.

### LINKED ISSUE #4995: Check and fix properties for sideeffects
We have observed unwanted side effects in #4994.  We should also ensure that properties - in particular in the area of classes concerned with expressions and constraints, to NOT have side effects.

If any is found, they should be fixed-

manfred-brands:
Fixes #4994 
Fixes #4995


stevenaw:
Thanks for fixing this @manfred-brands 
This looks good to me. Just for the sake of considering all sides, do we think there's a chance anyone was reliant on our override of `ToString()` ?

EDIT: I agree the existing implementation can be problematic. I'm thinking instead about if it's worth a dedicated call-out in any release notes about it

@OsirisTerje Any concerns about this going into 4.4?

stevenaw (approved):

src/NUnitFramework/tests/Assertions/AssertThatTests.cs
```diff
@@ -1090,5 +1090,17 @@ public void CompareMismatchedDifferentTypes()
                           .And.Message.Contains("Expected: \"89030\"")
                           .And.Message.Contains("But was:  \"89031\""));
         }
+
+        [Test]
+        public void Abc()
```
mikkelbu: Can we give this test a better name?

manfred-brands:
> do we think there's a chance anyone was reliant on our override of `ToString()` ?

The constraint doesn't work correctly anymore after calling that method, so I would say not.

stevenaw:
Thanks @manfred-brands 
I agree it's pretty unlikely outside perhaps the testing of custom constraints, similar to the usage in NUnit's suite.

src/NUnitFramework/tests/Assertions/AssertThatTests.cs
manfred-brands: Rename to `TestToVerifyThatWhenSteppingThroughWithDebuggerThisDoesNotAlterBehaviour`

PR #5003, opened 2025-06-26
Second Proposed Solution for Issue 4998

### LINKED ISSUE #4998: `PropertyConstraint` Causes Unexpected Behavior Because the Property's Value Is `object` Instead of the Property’s Actual Type

LeQuackers:
This proposal is similar to the [first one](https://github.com/nunit/nunit/pull/4999); however, it's a bit cleaner and is applied to more constraints.

As of writing this, some constraints cast the actual value to an `object`, which ends up causing inconsistencies / unexpected behavior when interacting with other constraints, in particular `DefaultConstraint`. `DefaultConstraint` was designed to match BCL behavior, so modifying it has caused a lot of discussion. Although it's a roundabout way of doing it, this proposal maintains the `DefaultConstraint's` behavior and resolves #4998, so it seems like a reasonable compromise.

As of writing this the following constraints have been modified:
- `AllItemsConstraint`
- `ExactCountConstraint`
- `NoItemConstraint`
- `PropertyConstraint`
- `SomeItemsConstraint`

If I missed any, let me know and I'll get them in. Feel free to edit as needed.

### DIFF
--- a/src/NUnitFramework/framework/Constraints/AllItemsConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/AllItemsConstraint.cs
@@ -41,7 +41,7 @@ namespace NUnit.Framework.Constraints
             int index = 0;
             foreach (object item in enumerable)
             {
-                if (!BaseConstraint.ApplyTo(item).IsSuccess)
+                if (!Reflect.InvokeApplyTo(BaseConstraint, item?.GetType(), item).IsSuccess)
                 {
                     return new EachItemConstraintResult(this, actual, item, index);
                 }
--- a/src/NUnitFramework/framework/Internal/Reflect.cs
+++ b/src/NUnitFramework/framework/Internal/Reflect.cs
@@ -1,10 +1,13 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using System.Linq.Expressions;
 using System.Reflection;
+using NUnit.Framework.Constraints;
 
 namespace NUnit.Framework.Internal
 {
@@ -25,6 +28,8 @@ namespace NUnit.Framework.Internal
     /// </summary>
     public static class Reflect
     {
+        private static readonly ConcurrentDictionary<Type, Func<IConstraint, object?, ConstraintResult>> InvokeApplyToLookup = [];
+
         internal static readonly BindingFlags AllMembers = BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static;
 
         #region Get Methods of a type
@@ -272,6 +277,43 @@ namespace NUnit.Framework.Internal
             {
                 throw new NUnitException("Rethrown", e);
             }
+        }
+
+        /// <summary>
+        /// Invokes <see cref="IConstraint.ApplyTo{TActual}(TActual)"/> using the given <paramref name="constraint"/> and
+        /// <paramref name="type"/> as its generic type argument.
+        /// </summary>
+        /// <param name="constraint">The <see cref="IConstraint"/>.</param>
+        /// <param name="type">The generic type argument to pass to <see cref="IConstraint.ApplyTo{TActual}(TActual)"/>.</param>
+        /// <param name="actual">The actual value.</param>
+        /// <returns>A <see cref="ConstraintResult"/>.</returns>
+        public static ConstraintResult InvokeApplyTo(IConstraint constraint, Type? type, object? actual)
+        {
+            Guard.ArgumentNotNull(constraint, nameof(constraint));
+
+            type ??= typeof(object);
+            var actualType = actual?.GetType();
+            Guard.ArgumentValid(actualType is null
+                ? !type.IsValueType
+                : type.IsAssignableFrom(actualType),
+                "The actual value must be assignable to the provided type.",
+                nameof(type));
+
+            return InvokeApplyToLookup.GetOrAdd(type, BuildApplyToDelegate).Invoke(constraint, actual);
+        }
+
+        private static Func<IConstraint, object?, ConstraintResult> BuildApplyToDelegate(Type type)
+        {
+            var methodInfo = ((MethodCallExpression)((LambdaExpression)((IConstraint x) => x.ApplyTo(0))).Body).Method
+                .GetGenericMethodDefinition();
+
+            var targetParameter = Expression.Parameter(typeof(IConstraint), "constraint");
+            var actualParameter = Expression.Parameter(typeof(object), "actual");
+
+            return Expression.Lambda<Func<IConstraint, object?, ConstraintResult>>(
+                Expression.Call(targetParameter, methodInfo.MakeGenericMethod(type), Expression.Convert(actualParameter, type)),
+                [targetParameter, actualParameter])
+                .Compile();
         }
 
         #endregion
--- a/src/NUnitFramework/tests/Constraints/ConstraintCastingTests.cs
+++ b/src/NUnitFramework/tests/Constraints/ConstraintCastingTests.cs
@@ -1,0 +1,123 @@
+using System.Collections.Generic;
+using System.Linq;
+
+#pragma warning disable NUnit2046 // Use Constraint for better assertion messages in case of failure
+namespace NUnit.Framework.Tests.Constraints
+{
+    /// <summary>
+    /// A test fixture that ensures constraints cast their associated property,
+    /// collection item, etc. to its actual type instead of just <see cref="object"/>.
+    /// </summary>
+    [TestFixture]
+    public class ConstraintCastingTests
+    {
+        [Test]
+        public void LengthPropertyConstraint()
+        {
+            var defaultLength = string.Empty;
+            var nonDefaultLength = "1";
+
+            using (Assert.EnterMultipleScope())
+            {
+                Assert.That(defaultLength.Length, Is.Zero);
+                Assert.That(defaultLength.Length, Is.Default);
+                Assert.That(defaultLength, Has.Length.Default);
+                Assert.That(defaultLength, Has.Property("Length").Default);
+
+                Assert.That(nonDefaultLength.Length, Is.Not.Zero);
+                Assert.That(nonDefaultLength.Length, Is.Not.Default);
+                Assert.That(nonDefaultLength, Has.Length.Not.Default);
+                Assert.That(nonDefaultLength, Has.Property("Length").Not.Default);
+            }
+        }
+
+        [Test]
+        public void CountPropertyConstraint()
+        {
+            var defaultCount = new List<int>();
+            var nonDefaultCount = new List<int>() { 0 };
+
+            using (Assert.EnterMultipleScope())
+            {
+                Assert.That(defaultCount.Count, Is.Zero);
+                Assert.That(defaultCount.Count, Is.Default);
+                Assert.That(defaultCount, Has.Count.Default);
+                Assert.That(defaultCount, Has.Property("Count").Default);
+
+                Assert.That(nonDefaultCount.Count, Is.Not.Zero);
+                Assert.That(nonDefaultCount.Count, Is.Not.Default);
+                Assert.That(nonDefaultCount, Has.Count.Not.Default);
+                Assert.That(nonDefaultCount, Has.Property("Count").Not.Default);
+            }
+        }
+
+        [Test]
+        public void ExactConstraint()
+        {
+            var oneDefaultItems = new List<int>() { default };
+            var noDefaultItems = new List<int>() { 1 };
+
+            using (Assert.EnterMultipleScope())
+            {
+                Assert.That(oneDefaultItems.Single(), Is.Zero);
+                Assert.That(oneDefaultItems.Single(), Is.Default);
+                Assert.That(oneDefaultItems, Has.One.Default);
+                Assert.That(oneDefaultItems, Has.Exactly(1).Default);
+
+                Assert.That(noDefaultItems.Single(), Is.Not.Zero);
+                Assert.That(noDefaultItems.Single(), Is.Not.Default);
+                Assert.That(noDefaultItems, Has.One.Not.Default);
+                Assert.That(noDefaultItems, Has.Exactly(1).Not.Default);
+            }
+        }
+
+        [Test]
+        public void SomeItemsConstraint()
+        {
+            var someDefaultItems = new List<int>() { default, 1, 2, 3 };
+            var noDefaultItems = new List<int>() { 1, 2, 3 };
+
+            using (Assert.EnterMultipleScope())
+            {
+                Assert.That(someDefaultItems.Any(x => x == default), Is.True);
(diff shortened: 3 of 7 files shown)


LeQuackers:
@manfred-brands @mikkelbu @stevenaw @Dreamescaper
This is one of the ideas from the comment the other day; when you have time let me know what you think.

manfred-brands:
@LeQuackers Can you fix the build please?
Check all warnings in Visual Studio as they become errors in CI

LeQuackers:
@manfred-brands 
The warnings should be resolved. The `IEnumerable` in the collection constraints can return `null` items, but in some of them the item was cast to just `object` instead of `object?`. The nullable changes ended up flowing down to `EachItemConstraintResult` and `ExactCountConstraintResult`, which are a bit out of scope.

There was an unexpected issue with nullable value types; they're boxed to either their underlying value or null instead of itself ([docs](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types#boxing-and-unboxing)). That ended up adding more complexity than I'd like, but at the moment the unit tests are passing. Any ideas are welcome.

LeQuackers:
After thinking about it a little more, there could probably be some kind of inner
`ConstraintResult ApplyToEnumerable<TActual, TItem>(TActual actual, IEnumerable<TItem>)` method for the given collection constraint that does the actual iteration. If `TActual` has a single/"primary" `IEnumerable<T>` interface defined on it, then we could invoke it at runtime with something similar to `Reflect.InvokeApplyTo`; if not or if it's just an `IEnumerable`, we could call it normally with `object?` as the `TItem` type argument.

I think that'd clean up the nullable value type issue; I'll experiment with it tomorrow.

LeQuackers:
I implemented the idea; let me know if you have any questions/comments/concerns.

LeQuackers:
@manfred-brands
That should be the final simplification/cleanup; feel free to review the changes when you have time. let me know if you have any questions/comments/concerns.

src/NUnitFramework/framework/Constraints/AllItemsConstraint.cs
```diff
@@ -32,14 +34,26 @@ public AllItemsConstraint(IConstraint itemConstraint)
...
+        /// <inheritdoc cref="IConstraint.ApplyTo{TActual}(TActual)"/>
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
             var enumerable = ConstraintUtils.RequireActual<IEnumerable>(actual, nameof(actual));
+            var itemType = TypeHelper.FindPrimaryEnumerableInterfaceGenericTypeArgument(typeof(TActual));
 
+            return itemType is null
+                ? ApplyToCollection(actual, enumerable.Cast<object>())
+                : Reflect.InvokeApplyToCollection(this, actual?.GetType(), itemType, actual, enumerable);
+        }
```
manfred-brands: The `ApplyTo` is the same for all implementations.
But as there is no common base class for these constraints there is not much that can be done.

src/NUnitFramework/framework/Constraints/ICollectionConstraint.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using System.Collections.Generic;
+
+namespace NUnit.Framework.Constraints
+{
+    /// <summary>
+    /// Interface for collection constraints
+    /// </summary>
+    public interface ICollectionConstraint : IConstraint
```
manfred-brands: NUnit already has an `CollectionConstraint` class which does not implement this new interface.
Suggest to rename this to `IEnumerableConstraint`.

src/NUnitFramework/framework/Constraints/ICollectionConstraint.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        /// <summary>
+        /// Applies the constraint to a collection, returning a ConstraintResult.
+        /// </summary>
+        /// <param name="actual">The value that the collection originated from</param>
+        /// <param name="collection">The collection to be tested</param>
+        /// <returns>A ConstraintResult</returns>
+        ConstraintResult ApplyToCollection<TActual, TItem>(TActual actual, IEnumerable<TItem> collection);
```
manfred-brands: Rename to `ApplyToEnumerable` and use `enumerable` as parameter.

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -274,6 +281,83 @@ internal static bool CanImplicitlyConvertTo(this Type? from, Type to)
...
 
+        /// <summary>
+        /// Invokes <see cref="IConstraint.ApplyTo{TActual}(TActual)"/> using the given <paramref name="constraint"/> as the instance
+        /// with <paramref name="actualType"/> as its generic type argument.
+        /// </summary>
+        /// <param name="constraint">The <see cref="IConstraint"/>.</param>
+        /// <param name="actualType">The generic type argument to pass to <see cref="IConstraint.ApplyTo{TActual}(TActual)"/>.</param>
+        /// <param name="actual">The actual value.</param>
+        /// <returns>A <see cref="ConstraintResult"/>.</returns>
+        public static ConstraintResult InvokeApplyTo(IConstraint constraint, Type? actualType, object? actual)
```
manfred-brands: There is no need for `actualType` to be nullable.

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -274,6 +281,83 @@ internal static bool CanImplicitlyConvertTo(this Type? from, Type to)
...
+        {
+            Guard.ArgumentNotNull(constraint, nameof(constraint));
+
+            actualType ??= typeof(object);
```
manfred-brands: If `actualType` is not nullable, this line can go.

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -274,6 +281,83 @@ internal static bool CanImplicitlyConvertTo(this Type? from, Type to)
...
+                nameof(actual));
+
+            static ConstraintResult DelegateTemplate<T>(IConstraint constraint, object? actual)
+                => constraint.ApplyTo((T?)actual);
+
+            static Func<IConstraint, object?, ConstraintResult> BuildDelegate(Type type)
+                => ((Delegate)DelegateTemplate<object>).Method
+                .GetGenericMethodDefinition()
+                .MakeGenericMethod(type)
+                .CreateDelegate<Func<IConstraint, object?, ConstraintResult>>();
```
manfred-brands, line 308: CreateDelegate is expensive.
You do cache the result here, but one needs at least 10 calls to justify this over calling Method.Invoke.
For now lets leave it as it is.

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -274,6 +281,83 @@ internal static bool CanImplicitlyConvertTo(this Type? from, Type to)
...
+        /// using the given <paramref name="constraint"/> as the instance with <paramref name="actualType"/> and
+        /// <paramref name="itemType"/> as its generic type arguments.
+        /// </summary>
+        /// <param name="constraint">The <see cref="IConstraint"/>.</param>
+        /// <param name="actualType">The actual type argument to pass to <see cref="ICollectionConstraint.ApplyToCollection{TActual, TItem}(TActual, IEnumerable{TItem})"/>.</param>
+        /// <param name="itemType">The item type argument to pass to <see cref="ICollectionConstraint.ApplyToCollection{TActual, TItem}(TActual, IEnumerable{TItem})"/>.</param>
+        /// <param name="actual">The actual value.</param>
+        /// <param name="collection">The collection.</param>
+        /// <returns>A <see cref="ConstraintResult"/>.</returns>
+        public static ConstraintResult InvokeApplyToCollection(ICollectionConstraint constraint, Type? actualType, Type? itemType, object? actual, IEnumerable collection)
```
manfred-brands: Again `actualType` cannot be `null`, nor can `actual` (which is the same as _collection_).
All constraints that call this have a `Guard` on `IEnumerable` with _allowNull_ set to `false.
Neither are _actual_ nor _itemType_.

manfred-brands (approved):
I added one commit addressing the naming issue: _collection_ => _enumerable_
and simplifying the code a bit knowing that non of the parameters can be `null`.

src/NUnitFramework/framework/Constraints/ICollectionConstraint.cs
manfred-brands: Renamed by me.

src/NUnitFramework/framework/Constraints/ICollectionConstraint.cs
manfred-brands: Renamed by me.

src/NUnitFramework/framework/Internal/Reflect.cs
manfred-brands: Removed `actualType` completely.

src/NUnitFramework/framework/Internal/Reflect.cs
manfred-brands: Removed.

src/NUnitFramework/framework/Internal/Reflect.cs
manfred-brands: Simplified the code based upon this knowledge.

src/NUnitFramework/framework/Internal/Reflect.cs
LeQuackers: If you want to scrap the caching and just invoke the `MethodInfo` that's fine with me.

`Activator.CreateInstance(Type)` could probably also be used here, i.e. create some `new ApplyToInvoker<TActual>()` instance that'd call `ApplyTo<T>(T)` with the correct type argument, but I don't know the benchmarks for it. I'd assume it would be comparable to invoking the `MethodInfo`.

If [interceptors](https://github.com/dotnet/roslyn/blob/main/docs/features/interceptors.md) were out of the preview phase we could probably leverage those to inject the `ApplyToInvoker<TActual>` class during the constraint expression construction, but I think that's a long ways a way. Off the top of my head a rough example would probably look like this:
```csharp
public class Foo
{
	public int? Bar { get; set; }
}


...


var actual = new List<Foo>()
{
	new() { Bar = default },
	new() { Bar = 1 },
};

Assert.That(actual, Has.Some.Property("Bar").Default);


...


[InterceptsLocation(version: 1, data: "the location of Assert.That(...)")]
public static void InterceptorMethod<T>(T actual, ConstraintExpression constraintExpression, ...)
{
	// Nothing to do for ".Some"
	// ".Some" would have a declarationIndex of 0
	
	// Applies an invoker that calls `ApplyTo<Foo>(Foo)` for all the constraints added by ".Property(...)"
	constraintExpression.SetApplyToInvoker(declarationIndex: 1, new ApplyToInvoker<Foo>());
	
	// Applies an invoker that calls `ApplyTo<int?>(int?)` for all the constraints added by ".Default"
	constraintExpression.SetApplyToInvoker(declarationIndex: 2, new ApplyToInvoker<int?>());

	// Call original assertion here.
}
```

OsirisTerje:
Is this WIP or ready to be merged?

manfred-brands:
> Is this WIP or ready to be merged?

It is ready, but I wouldn't mind a 2nd review.

src/NUnitFramework/framework/Internal/TypeHelper.cs
```diff
@@ -332,6 +332,53 @@ public static bool IsTuple(Type type)
...
+        /// For the given <paramref name="type"/>, find the primary <see cref="IEnumerable{T}"/> interface declared
+        /// on it and return its generic type argument. If multiple <see cref="IEnumerable{T}"/> interfaces are declared,
+        /// return <see langword="null"/>.
+        /// </summary>
+        /// <param name="type">The given <see cref="Type"/>.</param>
+        /// <returns>
+        /// The generic type argument of the primary <see cref="IEnumerable{T}"/> interface declared on the given
+        /// <paramref name="type"/>, or <see langword="null"/> if zero interfaces are found or more than one interface is found.
+        /// </returns>
+        public static Type? FindPrimaryEnumerableInterfaceGenericTypeArgument(Type type)
```
OsirisTerje, line 345: Is there a test for this ?

(I see it is covered by something, but there are no explicit test for this one, afaic see)

src/NUnitFramework/framework/Internal/TypeHelper.cs
```diff
@@ -332,6 +332,53 @@ public static bool IsTuple(Type type)
...
+            {
+                var @interface = queue.Dequeue();
+
+                // if we run into multiple `IEnumerable<T>` that have the same T generic argument then that's fine,
+                // i.e. if the primary enumerable type is equal to the interface type we can skip this check.
+                if (primaryEnumerableType != @interface && predicate(@interface))
+                {
+                    if (found)
+                    {
+                        return null;
```
OsirisTerje, line 367: This case is not covered by any tests.

OsirisTerje (changes requested):

src/NUnitFramework/tests/Constraints/ConstraintCastingTests.cs
```diff
@@ -0,0 +1,203 @@
...
+
+#pragma warning disable NUnit2046 // Use Constraint for better assertion messages in case of failure
+namespace NUnit.Framework.Tests.Constraints
+{
+    /// <summary>
+    /// A test fixture that ensures constraints cast their associated property,
+    /// collection item, etc. to its actual type instead of just <see cref="object"/>.
+    /// </summary>
+    [TestFixture]
+    public class ConstraintCastingTests
```
OsirisTerje, line 15: The issue #4998 has some example code. 
Could this code be added as a Test here ?

```csharp

   var actual = new Foo
    {
        Bar = 0
    };

    Assert.Multiple(() =>
    {
        // passes

        // fails
    });
    
```

OsirisTerje (changes requested):

src/NUnitFramework/tests/Constraints/ConstraintCastingTests.cs
manfred-brands: The tests added `LengthPropertyConstraint` and `CountPropertyConstraint` cover the same functionality:
```csharp
Assert.That(defaultLength.Length, Is.Default);
Assert.That(defaultLength, Has.Length.Default);
Assert.That(defaultLength, Has.Property("Length").Default);
```

I don't see the need to add a `Foo` class just for the sake of it. I rather update #4998 to use `Length` as an example.

src/NUnitFramework/framework/Internal/TypeHelper.cs
LeQuackers: Sorry for the delay; I added some unit tests for the method.

src/NUnitFramework/framework/Internal/TypeHelper.cs
LeQuackers: This should be covered now by the added tests.

Issue #3947, opened 2021-09-30
Simply asserting ArgumentException.ParamName


heaths:
As an alternative to #2820, what about borrowing a concept from xUnit's `Assert.Throws<T>(string, Action) where T : ArgumentException` and add an overload to `Throws.InstanceOf<T>()` and `Throws.Typeof<T>()`:

```c#
InstanceOf<T>(string paramName, Action action) where T : ArgumentException;
TypeOf<T>(string paramName, Action action) where T : ArgumentException;
```

This has the benefit of being brief without adding an `ParamName(string)` on every `Constraint` or derivative. I'd be happy to submit a PR if acceptable.

stevenaw:
This is an interesting idea. I had recently been looking at ways of converting `Assert.Throws` syntax into equivalent constraint-based methods for the similar purpose of inspecting this property.

Something like this:
```csharp
var exception = Assert.Throws<ArgumentException>(() => throw new ArgumentException("message", "foo"));
Assert.That(exception.ParamName, Is.EqualTo("foo");
```

becomes much harder when using the `Throws` constraint as we don't currently strong-type the constraint. 
In addition to the above proposal, another syntax idea could be:

```csharp
Assert.That(() => { /*throw */ }, Throws.ArgumentException.Property(o => o.ParamName).EqualTo("foo"));
```

stevenaw:
I've put this into the 5.0 milestone as it will be much simpler once https://github.com/nunit/nunit/issues/5007 completes

PR #5008, opened 2025-07-04
Add generic type constraints

### LINKED ISSUE #5007: Create generic type constraints
Many type constraints currently have generic function overloads deferring to non-generic constraints, in particular for exception-related assertions. Adding generic constraints for these methods would be a nice step towards fulfilling the generic constraint model described in #53

stevenaw:
A bit more incremental work towards generic constraints.

Fixes #5007 
Contributes to #53

This PR adds generic classes and overloads for all things under the `TypeConstraint` hierarchy. Each generic type was extended from it's non-generic counterpart as a way of avoiding duplication of logic. The thought behind this decision is that it simplifies short-term future maintenance while still allowing us a way to establish API signatures with some generic support without having to make an entire generic constraint resolver... yet. The tradeoff would be that all type logic is deferred to non-generic functions, resulting in a negligible difference from ideal performance as we'd lose the ability for some JIT-time constant folding in some cases on some runtimes. I felt this was an acceptable tradeoff to allow another piece of the generic API work to proceed.

### DIFF
--- a/src/NUnitFramework/framework/Assert.Exceptions.Async.cs
+++ b/src/NUnitFramework/framework/Assert.Exceptions.Async.cs
@@ -117,7 +117,7 @@ namespace NUnit.Framework
         /// <param name="args">Arguments to be used in formatting the message</param>
         public static Exception? CatchAsync(AsyncTestDelegate code, string message, params object?[]? args)
         {
-            return ThrowsAsync(new InstanceOfTypeConstraint(typeof(Exception)), code, message, args);
+            return ThrowsAsync(new InstanceOfTypeConstraint<Exception>(), code, message, args);
         }
 
         /// <summary>
@@ -127,7 +127,7 @@ namespace NUnit.Framework
         /// <param name="code">A TestDelegate</param>
         public static Exception? CatchAsync(AsyncTestDelegate code)
         {
-            return ThrowsAsync(new InstanceOfTypeConstraint(typeof(Exception)), code);
+            return ThrowsAsync(new InstanceOfTypeConstraint<Exception>(), code);
         }
 
         /// <summary>
@@ -168,7 +168,7 @@ namespace NUnit.Framework
         public static TActual? CatchAsync<TActual>(AsyncTestDelegate code, string message, params object?[]? args)
             where TActual : Exception
         {
-            return (TActual?)ThrowsAsync(new InstanceOfTypeConstraint(typeof(TActual)), code, message, args);
+            return (TActual?)ThrowsAsync(new InstanceOfTypeConstraint<TActual>(), code, message, args);
         }
 
         /// <summary>
@@ -179,7 +179,7 @@ namespace NUnit.Framework
         public static TActual? CatchAsync<TActual>(AsyncTestDelegate code)
             where TActual : Exception
         {
-            return (TActual?)ThrowsAsync(new InstanceOfTypeConstraint(typeof(TActual)), code);
+            return (TActual?)ThrowsAsync(new InstanceOfTypeConstraint<TActual>(), code);
         }
 
         #endregion
--- a/src/NUnitFramework/framework/Assert.Exceptions.Async.cs
+++ b/src/NUnitFramework/framework/Assert.Exceptions.Async.cs
@@ -117,7 +117,7 @@ namespace NUnit.Framework
         /// <param name="args">Arguments to be used in formatting the message</param>
         public static Exception? CatchAsync(AsyncTestDelegate code, string message, params object?[]? args)
         {
-            return ThrowsAsync(new InstanceOfTypeConstraint<Exception>(), code, message, args);
+            return CatchAsync<Exception>(code, message, args);
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Assert.Exceptions.Async.cs
+++ b/src/NUnitFramework/framework/Assert.Exceptions.Async.cs
@@ -89,7 +89,7 @@ namespace NUnit.Framework
         public static TActual? ThrowsAsync<TActual>(AsyncTestDelegate code, string message, params object?[]? args)
             where TActual : Exception
         {
-            return (TActual?)ThrowsAsync(typeof(TActual), code, message, args);
+            return (TActual?)ThrowsAsync(new ExceptionTypeConstraint<TActual>(), code, message, args);
         }
 
         /// <summary>
--- a/src/NUnitFramework/framework/Assert.Exceptions.Async.cs
+++ b/src/NUnitFramework/framework/Assert.Exceptions.Async.cs
@@ -127,7 +127,7 @@ namespace NUnit.Framework
         /// <param name="code">A TestDelegate</param>
         public static Exception? CatchAsync(AsyncTestDelegate code)
         {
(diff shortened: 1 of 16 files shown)


OsirisTerje:
@stevenaw 
1. Are these changes covered by already existing tests ?

stevenaw:
Thanks @OsirisTerje It looks like copilot also caught a few things. I'll set this to WIP until I can address

stevenaw:
I'm moving this out of WIP status and marking it as Ready for Review.
Many of these changes are just converting from non-generic to generic equivalents and are covered under existing tests. I have added new constraint-level tests for the generic constraints introduced here to ensure consistent messaging and xml representation.

~~The failing MacOS build is a timeout issue which I'll rerun in the morning. Windows and Linux tests are passing~~
EDIT: MacOS tests have passed on rerun now too

OsirisTerje:
@stevenaw  Have you tested these changes against the issue #5011 repro code ?

Please merge with latest main to get @manfred-brands fixes in to this branch.

src/NUnitFramework/framework/Assert.Exceptions.cs
```diff
@@ -118,7 +118,7 @@ public abstract partial class Assert
         /// <param name="args">Arguments to be used in formatting the message</param>
         public static Exception? Catch(TestDelegate code, string message, params object?[]? args)
         {
-            return Throws(new InstanceOfTypeConstraint(typeof(Exception)), code, message, args);
+            return Throws(new InstanceOfTypeConstraint<Exception>(), code, message, args);
```
manfred-brands: This could call the generic overload:
```suggestion
            return Throws<Exception>(code, message, args);
```

src/NUnitFramework/framework/Assert.Exceptions.cs
```diff
@@ -128,7 +128,7 @@ public abstract partial class Assert
         /// <param name="code">A TestDelegate</param>
         public static Exception? Catch(TestDelegate code)
         {
-            return Throws(new InstanceOfTypeConstraint(typeof(Exception)), code);
+            return Throws(new InstanceOfTypeConstraint<Exception>(), code);
```
manfred-brands: Call the generic overload:
```suggestion
            return Catch<Exception>(code);
```

src/NUnitFramework/framework/Assert.Exceptions.Async.cs
```diff
@@ -127,7 +127,7 @@ public partial class Assert
         /// <param name="code">A TestDelegate</param>
         public static Exception? CatchAsync(AsyncTestDelegate code)
         {
-            return ThrowsAsync(new InstanceOfTypeConstraint(typeof(Exception)), code);
+            return ThrowsAsync(new InstanceOfTypeConstraint<Exception>(), code);
```
manfred-brands: Call generic version:
```suggestion
            return CatchAsync<Exception>(code);
```

src/NUnitFramework/framework/Assert.Exceptions.Async.cs
```diff
@@ -117,7 +117,7 @@ public partial class Assert
         /// <param name="args">Arguments to be used in formatting the message</param>
         public static Exception? CatchAsync(AsyncTestDelegate code, string message, params object?[]? args)
         {
-            return ThrowsAsync(new InstanceOfTypeConstraint(typeof(Exception)), code, message, args);
+            return ThrowsAsync(new InstanceOfTypeConstraint<Exception>(), code, message, args);
```
manfred-brands: Call generic overload:
```suggestion
            return CatchAsync<Exception>(code, message, args);
```

manfred-brands (commented):
@stevenaw I'm sorry, but as a standalone change, I don't see the benefits.

It is a breaking change as it changes the return type of various `Is.XXX` methods.
But it doesn't add any code that takes advantage of the generic type.
The generic overloads have no type specific properties. 
The constraints modified only care about the type, not about instances of the generic type where generics would add something,

We already had overloads `TException Assert.Throws<TException>()`

We don't have code that can do that in the constraint model:
`Throws.TypeOf<InconclusiveException>().With.Message.EqualTo("MESSAGE"))`

This falls back to reflection even though we know the type at compile time, but we loose it as soon as we add to the constraint expression.

To me adding generics only makes sense if it adds something. 
Either a restriction of illegal code such as passing Tolerance to a string comparison or IgnoreCase to a numeric comparsion.
Or if it allows less reflection and allows using the type in further constraints, such as: 
`Throws.TypeOf<InconclusiveException>().Where(x => x.Message.EqualTo("MESSAGE"))`

stevenaw:
@manfred-brands Thanks for catching some more cases where generic overloads could be called. I had tried to catch them all but seem to have missed some.

On the note of breaking, none of these changes should be breaking as the changed return types all extend the existing return types.

In terms of benefits, the goal here is to lay groundwork for future compile-time niceties like https://github.com/nunit/nunit/issues/3947#issuecomment-2270189739 within the 4.x version line. I suspect getting these generic APIs in place now will also make it easier to move towards compile-time generic type checking later on too.

@OsirisTerje I've been away visiting family but I can do that soon. I'll try to wrap this up soon, but please do not hold up a 4.4 release over this

stevenaw:
I can confirm the tests all pass after merging latest main in. I haven't tested it against the repro code directly yet

manfred-brands:
> On the note of breaking, none of these changes should be breaking as the changed return types all extend the existing return types.

That's what I previously thought, but was proven wrong in #5011

stevenaw:
Ah. That's a fair nuance. I _think_ it should still be safe here as I don't think any overloads were introduced which could affect method resolution. I'll also give the repro a try when I'm able.

EDIT: The runtime guide to breaking changes also indicates that certain edgecases may come up here. https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md#property-field-parameter-and-return-values 

I'll take a closer look at this

manfred-brands:
Your code is _compile time_ compatible but not _binary_ compatible.

Given the following test:
```csharp
[TestCase(3)]
[TestCase(3.3)]
public void TypeOf<T>(T value)
{
    Assert.That(value, Is.TypeOf<T>());
}
```

Compile it against nunit 4.3.2 and then copy the dll from this branch to the binaries. Running it then results in the expected binary compatibility issue:
```text
TypeOf<Int32>(3) (< 1ms): Error Message: System.MissingMethodException : Method not found: 'NUnit.Framework.Constraints.ExactTypeConstraint NUnit.Framework.Is.TypeOf()
TypeOf<Double>(3.3d) (< 1ms): Error Message: System.MissingMethodException : Method not found: 'NUnit.Framework.Constraints.ExactTypeConstraint NUnit.Framework.Is.TypeOf()
```

manfred-brands:
It falls under this subheading: https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md#signatures

image

stevenaw:
Ok thanks for confirming this one @manfred-brands 
I'll try to change the signature back within a few hours.

stevenaw:
I'm going to close this one for now as I think it's tough to get even the compile-time benefits from this without changing return types

PR #5010, opened 2025-07-14
Add Assert.Retry

dfev77:
Allow to retry an assert with a given timeout and retry interval. This is used for UI tests, where the application UI might take some time to update, subject to server processing and network.

As this might be seen as a minor use case, I was not sure if to add the assert method in the same class or create a separate one, AssertRetry.


dfev77:
@dotnet-policy-service agree

OsirisTerje:
Thanks for the PR.  

We would appreciate that PRs are a result of a raised, and discussed, issue.  

Further, we do have the Retry attribute, so not sure what your implementation do that is different from the Retry attribute.
Also, I do see you have changed the ReportFailure method.  I assume you have done so since you have changed the grammar we use.

dfev77:
Thank you for your feedback, I suppose I's too late to start that way 🙃
The intention was to contribute to the community and this wonderful project that we rely upon.

On what this brings, compared with RetryAttribute:
* RetryAttribute is applied to a test/fixture to re-trigger that until passes.
* the retry assert, is just a construct to retry an assert

Our use case is that we are writing UI system/integration tests, based on NUnit & [FlaUI](https://github.com/FlaUI/FlaUI). As these involve a number of other components & servers the effect of an action will only be reflected in the UI after a while. So instead of hard-coding  some sleep between actions & the asserts we rely on retrying the assert operation.

The ReportFailure was changes in order to reuse the formatting of the message printed on failure. When troubleshooting we want to be able to print out whatever happened during retries, so if the assert is not matched you are able (opt-in) to print out the failed assert message.

Other reasons to contribute rather then keep it private is that such a custom assert cannot be integrated with multiple assert by the end-dev as MultipleAssertLevel & ThrowOnEachFailureUnderDebugger (from TestExecutionContext) are internal and making them public would not be nice.

dfev77:
Any feedback from the team would be highly appreciated

stevenaw:
Thanks @dfev77 
I agree with @OsirisTerje , who is not just a team member but also the team lead, on their points. We welcome all contributions, however the path to landing code changes is smoothest when we can discuss them ahead of time before the effort is put into coding and integrating a changeset.

Is there a reason the built-in `Retry` attribute didn't meet your needs, and if so how did the changes here bridge any gap for you that the attribute didn't yet handle?

manfred-brands:
@dfev77 
I think that your retry functionality you created  is already there with the `DelayedConstraint` which is accessed using the `After` modifier:

```csharp
[Test]
public void GivenATrueConditionAfterRetry_Passes()
{
    Stopwatch stopwatch = Stopwatch.StartNew();
    Assert.That(() => stopwatch.Elapsed.TotalSeconds, Is.GreaterThan(2).After(3000).PollEvery(100));
}
```

This allows specifying a total timeout and a poll interval for the retry.
It allows different modifiers like:
```csharp
    Assert.That(() => stopwatch.Elapsed.TotalSeconds,
                Is.GreaterThan(2.0)
                  .After(3).Seconds
                  .PollEvery(100).MilliSeconds);
```

I could argue that the term `After` is misleading as it doesn't wait that time before testing.
Maybe we need an overload called `Before` indicate that it should pass _before_ the specified timeout.

OsirisTerje:
@dfev77  What @manfred-brands points out seems to solve your issue, right ?  Documentation here:  https://docs.nunit.org/articles/nunit/writing-tests/constraints/DelayedConstraint.html

@nunit/framework-team We could perhaps look at improving the documentation for the DelayedConstraint, also perhaps put a pointer from the docs on the Retry attribute to the delayedconstraint and explain how they can be used to solve issues like @dfev77 discuss.

@manfred-brands said:

> I could argue that the term After is misleading as it doesn't wait that time before testing.
> Maybe we need an overload called Before indicate that it should pass before the specified timeout.

Agree.  What about RetryUntil ?

dfev77:
@manfred-brands , @OsirisTerje yes, seems the thing I was looking for, thank you. Sorry for not spotting the pollingInterval/PollEvery earlier.

One small question, would you see it appropriate for DelayedConstraint to (opt-in) print the ContraintResult when retrying, if the constraint failed?
I would see this useful when dealing with unstable tests.

manfred-brands:
> One small question, would you see it appropriate for DelayedConstraint to (opt-in) print the ContraintResult when retrying, if the constraint failed?
> I would see this useful when dealing with unstable tests.

I'm not a fan of output. 
If the test eventually succeeds nobody will check it.
If the test keeps failing, the last error will be reported.

PR #5015, opened 2025-07-23
Use span-based stackalloc for random GUID generation on non-NETFRAMEWORK targets

### LINKED ISSUE #5014: Use span-based APIs to construct random GUIDs on non-framework codepath
We construct a small 16-byte array here to generate random bytes for a GUID.
https://github.com/nunit/nunit/blob/a865cf2/src/NUnitFramework/framework/Internal/Randomizer.cs#L680-L681

Span-based APIs are available here on non-framework builds which should allow for a small performance boost. The size of the allocation is also small enough that we could likely stackalloc too. For example (untested)

```csharp
public Guid NextGuid()
{
#if NETFRAMEWORK
  byte[] b = new byte[16];
#else
  Span<byte> b = stackalloc byte[16];
#endif

  NextBytes(b);

  // rest of the implementation
}
```

zaidonolsen:
- Updated Randomizer.NextGuid() to use stackalloc Span<byte> for random byte buffer on .NET Core and .NET 6+ targets, reducing heap allocations and improving performance.
- Existing array allocation is retained for .NET Framework builds.
- Added/updated unit tests to ensure correct behavior on all targets.

Fixes #5014

### DIFF
--- a/src/NUnitFramework/tests/Internal/RandomizerTests.cs
+++ b/src/NUnitFramework/tests/Internal/RandomizerTests.cs
@@ -661,6 +661,25 @@ namespace NUnit.Framework.Tests.Internal
             Assert.That(bytes[8] & 0xc0, Is.EqualTo(0x80));
         }
 
+#if !NETFRAMEWORK
+        [Test]
+        [Description("Test that NextGuid uses the span-based code path and produces valid GUIDs on non-NETFRAMEWORK targets")]
+        public void RandomGuidSpanApiWorks()
+        {
+            // This test is to ensure the span-based code path is exercised and produces valid results.
+            // It is functionally similar to RandomGuidsAreV4, but documents the intent for the span-based implementation.
+            var guid = _randomizer.NextGuid();
+            var bytes = guid.ToByteArray();
+            // Check the version (should be 4)
+            Assert.That(bytes[7] & 0xf0, Is.EqualTo(0x40));
+            // Check the variant (should be 0b10xxxxxx)
+            Assert.That(bytes[8] & 0xc0, Is.EqualTo(0x80));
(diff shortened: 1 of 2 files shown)


zaidonolsen:
@dotnet-policy-service agree

src/NUnitFramework/tests/Internal/RandomizerTests.cs
```diff
@@ -661,6 +661,25 @@ public void RandomGuidsAreV4()
             Assert.That(bytes[8] & 0xc0, Is.EqualTo(0x80));
         }
 
+#if !NETFRAMEWORK
```
manfred-brands: This test should not be added.
The whole idea is that optimizations in the library don't affect the tests. The normal test will exercise the new path.

manfred-brands (changes requested):
Thanks for your PR.
Please remove the extra test.

src/NUnitFramework/tests/Internal/RandomizerTests.cs
zaidonolsen: Thanks for the feedback. I have removed it.

zaidonolsen:
> Thanks for your PR. There are two more places allocating an array before calling NextBytes. Please update those as well.
> 
> Please remove the extra test.

I have updated the code. Thanks.

stevenaw (approved):
This looks good to me, thanks @zaidonolsen !
All github action workflows have run succesfully. The macos failure was an unrelated timeout which passed when run again

stevenaw:
@manfred-brands Any concerns on your side?

manfred-brands (approved):
Thanks, looks good now.
Northing further from me.

PR #5019, opened 2025-07-30
.net 6 cleanup

siva1160github:


OsirisTerje:
We're not doing this now.

OsirisTerje:
Appreciate the effort  @siva1160github , but we need to do this on a breaking change, and we're not quite there yet.  There is a lot more we want to do on the 5.X version, and we haven't started yet to create that list.

If you want to contribute, I would suggest you have a look at this list, there are 24 issues there, and some of these should be rather easy to fix.  https://github.com/nunit/nunit/issues?q=is%3Aissue%20state%3Aopen%20label%3Apri%3Alow

siva1160github:
Sure @OsirisTerje . Let me check those open issues.

PR #5020, opened 2025-08-05
chore: Remove TODOs related to old NUnit.Analyzer issues and bump package

mikkelbu:
I didn't add the NUnit.Analyzers to nunit.testdata as there were ~60 compile errors due to test cases that checks how the framework handles bad data. So I've left that as future work.


OsirisTerje (approved):
lgtm

Issue #5025, opened 2025-08-28
PropertiesComparer fails on properties with new modifier


lrampazzo:
Hi,

First of all, thank you for your hard work in maintaining this library.

In the latest release (4.4.0), I have started experiencing some issues with tests using the properties comparer with properties that use the `new` modifier. Here is a simple repro case:

```csharp
public class Base
{
	public object Metadata { get; init; }
}

public class Derived<T> : Base
{
	public new T Metadata { get; init; }
}

[TestFixture]
public class NewModifierIssue
{
        [Test]
	public void PropertiesComparer_NewModiferDuplicateKey()
	{
		// Arrange
		var item1 = new Derived<string>
		{
			Metadata = "value"
		};
		var item2 = new Derived<string>
		{
			Metadata = "value"
		};

		// Fails with System.ArgumentException : An item with the same key has already been added. Key: Metadata
		Assert.That(item1, Is.EqualTo(item2).UsingPropertiesComparer());
	}
}
```

The issues is that `Metadata` fields are clashing in the properties [lookup](https://github.com/nunit/nunit/blob/209bcbc95c0ed5688ed7ffbe634b22657f5ff43f/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs#L88). 

It may be an idea to create lookups as `<PropertyInfo.DeclaringType.FullName>.<PropertyInfo.Name>`? If you like this solution i would gladly create a PR for you.

manfred-brands:
@lrampazzo Thanks for your report and thanks for your appreciation.

I hadn't thought of the use of `new` for properties. 
Your suggestion will not work as the comparer allows comparing different types.
What needs to happen is that any `new` properties should replace the property it shadows.

I presume your example is contrived as accessing the property from the base class will return `null` because they are two different properties.

lrampazzo:
Hi @manfred-brands, you're right, it isn't a viable solution. 

I took some time to investigate and I've implemented a possible solution in #5028.

PR #5028, opened 2025-08-31
feat: add support for properties override via new modifier

lrampazzo:
This PR adds support for properties overridden with [`new `modifier](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/new-modifier) in PropertiesComparer.

When an overridden property is found (in `GetComparableProperties`) now it tries to resolve the correct property by checking the `DeclaringType` and finding the nearest in the type hierarchy. 

Resolves: #5025

### DIFF
--- a/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
+++ b/src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
@@ -33,7 +33,7 @@ namespace NUnit.Framework.Constraints.Comparers
                 return EqualMethodResult.TypesNotSupported;
             }
 
-            PropertyInfo[] xProperties = GetComparableProperties(xType);
+            PropertyInfo[] xProperties = GetPropertiesToCompare(xType);
             HashSet<string> xPropertyNames = new HashSet<string>(xProperties.Select(p => p.Name));
 
             PropertyInfo[] yProperties = xProperties;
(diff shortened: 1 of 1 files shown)


lrampazzo:
@dotnet-policy-service agree

src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
```diff
@@ -150,6 +151,36 @@ public static EqualMethodResult Equal(object x, object y, ref Tolerance toleranc
...
+                    {
+                        // If the existing property is declared in the current type, keep the existing one
+                        if (existing.DeclaringType == type)
+                            continue;
+
+                        if (
+                            // If the property is declared in the current type, use it
+                            property.DeclaringType == type
+                            // Use the one declared in the most derived type (longest inheritance chain)
+                            || property.DeclaringType.TypeAndBaseTypes().Count() > existing.DeclaringType.TypeAndBaseTypes().Count())
```
manfred-brands: We don't need to count all base types. A base class is assignable from a derived class.
So if `existing.DeclaringType!.IsAssignableFrom(property.DeclaringType))` then we need to replace it.

src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
```diff
@@ -150,6 +151,36 @@ public static EqualMethodResult Equal(object x, object y, ref Tolerance toleranc
 
             return EqualMethodResult.ComparedEqual;
 
+            PropertyInfo[] GetComparableProperties(Type type)
```
manfred-brands: I prefer the name `GetProperiesToCompare`

manfred-brands (approved):
I made two small changes as highlighted in my comments below.
Thanks for your work.

lrampazzo:
Hey @manfred-brands, probably I'm not enabled to merge this PR because I don't see the button

src/NUnitFramework/framework/Constraints/Comparers/PropertiesComparer.cs
lrampazzo: Neat!

PR #5029, opened 2025-09-06
Issue5026 inconclusive

### LINKED ISSUE #5026: Assume.That no longer works in Setup
Since the latest release (4.4.0) Assume.That no longer ignores tests when it's in a setup method, where it did ignore them prior (4.3.2).

```
namespace TestProject1
{
    public class Tests
    {
        [SetUp]
        public void Setup()
        {
            Assume.That(
                "1",
                Is.EqualTo("2"),
                "1 needs to be equal to 2");
        }

        [Test]
        public void Test1()
        {
            Assert.Pass();
        }
    }
}
```

Notice that on 4.4.0 the test passes but on 4.3.2 it is ignored.

manfred-brands:
Fixes #5026 

A possibly controversial change is the default state of s suite going from `Passed` to `Inconclusive`.
It should only affect suites with no tests of which there should be not many.

An alternative approach could be to replace `Inconclusive` with `Skipped` if an Assume fails.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/BeforeAndAfterTestCommand.cs
@@ -37,8 +37,12 @@ namespace NUnit.Framework.Internal.Commands
                 BeforeTest(context);
 
                 // If the BeforeTest method fails, we don't run the test
-                if (context.CurrentResult.ResultState.Status is not TestStatus.Failed and not TestStatus.Skipped)
+                // An failing Assume sets the Status to Inconclusive, but with a FailureSite of SetUp
+                if (context.CurrentResult.ResultState.Status is not TestStatus.Failed and not TestStatus.Skipped &&
+                    context.CurrentResult.ResultState.Site != FailureSite.SetUp)
+                {
                     context.CurrentResult = innerCommand.Execute(context);
+                }
             });
 
             if (context.ExecutionStatus != TestExecutionStatus.AbortRequested)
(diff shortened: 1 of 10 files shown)


stevenaw:
Thanks for doing this @manfred-brands 
FWIW I agree with that it would sense that a suite has a default result of Inconclusive as a lack of a status or result on a test logically means that its state is exactly that: inconclusive.

For the sake of discussion, do we expect this change to surprise any users in any way that might require a callout in the release notes?

manfred-brands:
> For the sake of discussion, do we expect this change to surprise any users in any way that might require a callout in the release notes?

It should only affect developers that make empty tests so they can say all their tests pass.

It will still not _fail_ a build, but our reporting about _inconclusive_ tests is non existing.
`Test summary: total: 5, failed: 0, succeeded: 0, skipped: 0, duration: 0.8s`
This is also what VS shows.
image

The `Microsoft.VisualStudio.TestPlatform.ObjectModel.TestOutcome` used in the adapter has no `Inconclusive` state and it is translated to `None`, which is not reported. 
@OsirisTerje Would translation to `Skipped` be a better option?
I can't find it, but I do think I mentioned it before, why is the sum of the numbers not matching the total.

I cannot find any documentation for that namespace.
There seems to be an `Microsoft.VisualStudio.TestTools.Common.TestOutcome` enum which has an `Inconclusive` state, but that project is listed as legacy.

OsirisTerje:
@manfred-brands   I missed this comment.

We're ignoring the `Inconclusive`, so it ends up as a Microsoft "None".  
They are visible though in the Test Explorer, and appears as "not run". 
In the `dotnet test` results however, they are as you show, simply gone.

We could change that, and add a runsetting so people can choose what to map it to. 

But, the Assume is also used in Theory tests,  and there they are meant to be ignored if they fail, as they represent a case that should not be tested further - so when an Assume is failing , you get an `Inconclusive` state.   If that was mapped to Failed, then the whole Theory test would fail incorrectly. 

An example of some results:
image

In the command run, it will show:

image

About the total number, I can't see that we are sending that to the testhost at all.  I believe they just calculate that based on the number of tests they receive, and they DO receive the inconclusive ones too, just with a state of None.

Since this issue is about setting the whole Test fixture to inconclusive, I am not sure how that ends up.  I haven't testet that yet.......  Are we sending out results from the framework/engine for those tests ?

Issue #5040, opened 2025-10-05
Add NET10 target for the framework tests


stevenaw:
.NET 10 will be coming out in about a month. It's been RC for a bit. We should start getting the tests running against it. Completing this would require installing [Visual Studio 2026 Insiders - Community Edition](https://visualstudio.microsoft.com/insiders/). ~~We should consider waiting to merge until .NET 10 hits stable so as to avoid forcing contributors to pull preview bits of IDEs.~~. We should be able to merge this anytime.

The previous issue where we did this for .NET 9 would be a good example to draw from for this: https://github.com/nunit/nunit/issues/4846

stevenaw:
I've labeled this as a `good first issue`. It would be a good candidate for anyone participating in Hacktoberfest.

zahran001:
I'm an open source noob - will it be okay if I start my self-study on this one?

mikkelbu:
@zahran001 Sure - I've assigned the issue to you - please ask questions if you have any. I guess you can get some inspiration from the addition to .NET8

@stevenaw Regarding Hacktoberfest, then will the candidate get credit for this if we first merge it in November?

stevenaw:
Welcome @zahran001 ! If you haven't yet, please check out https://github.com/nunit/nunit/blob/main/BUILDING.md and https://github.com/nunit/nunit/blob/main/CONTRIBUTING.md#fixing-bugs-and-adding-features as it will help guide you on how to get setup and, eventually, to submit a Pull Request. Note that for this particular issue you will need to install [Visual Studio 2026 Insiders - Community Edition](https://visualstudio.microsoft.com/insiders/) in order to locally run the tests in .NET 10.

@mikkelbu Good point. In the past there's been some leeway from DigitialOcean with merging PRs so long as it has been created in the month of October. I think it would still be fine to merge any PRs for this in October, especially if it would help earn badges or other things. I see they've brought back t-shirts this year 🙂

zahran001:
@stevenaw @mikkelbu 
Thank you so much for the heads up!!

PR #5041, opened 2025-10-08
Add NET10 target for the framework tests #5040

zahran001:
Fixes https://github.com/nunit/nunit/issues/5040

image

### DIFF
--- a/.github/workflows/NUnit.CI.yml
+++ b/.github/workflows/NUnit.CI.yml
@@ -30,7 +30,7 @@ jobs:
         dotnet-version: |
             6.0.x
             8.0.x
-            10.0.x
+            9.0.x
         
     - name: 🛠️ Install dotnet tools
       run: dotnet tool restore
@@ -75,7 +75,7 @@ jobs:
         dotnet-version: |
             6.0.x
             8.0.x
-            10.0.x
+            9.0.x
 
     - name: 🛠️ Install F#
       run: sudo apt-get install fsharp
@@ -111,7 +111,7 @@ jobs:
         dotnet-version: |
             6.0.x
             8.0.x
-            10.0.x
+            9.0.x
 
     - name: 🛠️ Install dotnet tools
       run: dotnet tool restore --tool-manifest ./.config-macos/dotnet-tools.json


zahran001:
@dotnet-policy-service agree

src/NUnitFramework/framework/Internal/StackFilter.cs
```diff
@@ -14,7 +14,7 @@ namespace NUnit.Framework.Internal
     public class StackFilter
     {
         private const string DEFAULT_TOP_OF_STACK_PATTERN = @" NUnit\.Framework\.(Assert|Assume|Warn|CollectionAssert|StringAssert|FileAssert|DirectoryAssert)\.";
-        private const string DEFAULT_BOTTOM_OF_STACK_PATTERN = @" System\.(Reflection|RuntimeMethodHandle|Threading\.ExecutionContext)\.";
+        private const string DEFAULT_BOTTOM_OF_STACK_PATTERN = @" System\.(Reflection|RuntimeMethodHandle|Threading)\.";
```
manfred-brands: I don't think this is correct. It seems unrelated to .NET 10.
It would mean we drop an System.Threading.Tasks etc from a Stacktrace.

.github/workflows/NUnit.CI.yml
```diff
@@ -30,6 +30,7 @@ jobs:
         dotnet-version: |
             6.0.x
             8.0.x
+            10.0.x
```
manfred-brands: Please add 9.0.x, It was previously found using the global.json file.
Suggest to drop 10.0.x That way if we move from rc to production we don't need to update this file.

.github/workflows/NUnit.CI.yml
```diff
@@ -74,6 +75,7 @@ jobs:
...
```
manfred-brands: Please add 9.0.x, It was previously found using the global.json file.
Suggest to drop 10.0.x That way if we move from rc to production we don't need to update this file.

.github/workflows/NUnit.CI.yml
```diff
@@ -109,6 +111,7 @@ jobs:
...
```
manfred-brands: Please add 9.0.x, It was previously found using the global.json file.
Suggest to drop 10.0.x That way if we move from rc to production we don't need to update this file.

manfred-brands (changes requested):
Thanks @zahran001 for your contribution.
See my comments.

src/NUnitFramework/framework/Internal/StackFilter.cs
zahran001: # net10.0 test failing

## This issue is most likely related to the System.Threading runtime frames emitted on .NET 10

image

image

.github/workflows/NUnit.CI.yml
zahran001: Addressed accordingly!

src/NUnitFramework/framework/Internal/StackFilter.cs
manfred-brands: I see that the raw stack trace on .NET Framework looks like:
```text
   at NUnit.Framework.Assert.Warn(String message) in D:\Development\3rd\NUnit\nunit\src\NUnitFramework\framework\Assert.cs:line 144
   at NUnit.TestData.WarningFixture.<>c__DisplayClass74_0.<WarningInThreadPoolQueueUserWorkItem>b__0(Object _) in D:\Development\3rd\NUnit\nunit\src\NUnitFramework\testdata\WarningFixture.cs:line 536
   at System.Threading.QueueUserWorkItemCallback.System.Threading.IThreadPoolWorkItem.ExecuteWorkItem()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
```

On .NET 6.0:
```text
   at NUnit.Framework.Assert.Warn(String message) in D:\Development\3rd\NUnit\nunit\src\NUnitFramework\framework\Assert.cs:line 144
   at NUnit.TestData.WarningFixture.<>c__DisplayClass74_0.<WarningInThreadPoolQueueUserWorkItem>b__0(Object _) in D:\Development\3rd\NUnit\nunit\src\NUnitFramework\testdata\WarningFixture.cs:line 536
   at System.Threading.QueueUserWorkItemCallback.<>c.<.cctor>b__6_0(QueueUserWorkItemCallback quwi)
   at System.Threading.ExecutionContext.RunForThreadPoolUnsafe[TState](ExecutionContext executionContext, Action`1 callback, TState& state)
   at System.Threading.QueueUserWorkItemCallback.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
   at System.Threading.Thread.StartCallback()
```

.NET 8/9:
```text
   at NUnit.Framework.Assert.Warn(String message) in D:\Development\3rd\NUnit\nunit\src\NUnitFramework\framework\Assert.cs:line 144
   at NUnit.TestData.WarningFixture.<>c__DisplayClass74_0.<WarningInThreadPoolQueueUserWorkItem>b__0(Object _) in D:\Development\3rd\NUnit\nunit\src\NUnitFramework\testdata\WarningFixture.cs:line 536
   at System.Threading.QueueUserWorkItemCallback.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
```

And on .NET 8/9/10:
```text
   at NUnit.Framework.Assert.Warn(String message) in D:\Development\3rd\NUnit\nunit\src\NUnitFramework\framework\Assert.cs:line 144
   at NUnit.TestData.WarningFixture.<>c__DisplayClass74_0.<WarningInThreadPoolQueueUserWorkItem>b__0(Object _) in D:\Development\3rd\NUnit\nunit\src\NUnitFramework\testdata\WarningFixture.cs:line 536
   at System.Threading.QueueUserWorkItemCallback.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
   at System.Threading.Thread.StartCallback()
```

The NUnit StackFilter doesn't filter out the `ExecutionContext` frames like .NET Core does, but stops the stack trace there which I think is too early.

The test asserts on number of lines which is framework dependent.
It should instead test that the items that are supposed to be filtered out are not there.

This is out of scope for this ticket. I will create a separate ticket.

src/NUnitFramework/Directory.Build.props
```diff
@@ -8,7 +8,7 @@
     <DisableImplicitNuGetFallbackFolder>true</DisableImplicitNuGetFallbackFolder>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <NUnitLibraryFrameworks>net462;net6.0;net8.0</NUnitLibraryFrameworks>
```
manfred-brands, line 10: @OsirisTerje Do we want to ship a .net10 build or only proof we can run with .net10?

Also will we drop .net6.0 at the same time we add .net10?

src/NUnitFramework/framework/Internal/StackFilter.cs
manfred-brands: I created #5043

src/NUnitFramework/Directory.Build.props
OsirisTerje: @manfred-brands We start with proof we can run with .net10.

Adding .net 10 target (later) is non-breaking, but removing .net 6 is breaking, so we can do the former first.

But adding .net 10 target should be done when we actually need it, due to something behaving different, or if it give us or the user some other benefit by having it.    

From what you write above, I read you as the #5043 is not directly related to .net 10, right ?

stevenaw:
Thanks for your contribution @zahran001 .
LGTM

manfred-brands:
@stevenaw I hoped that you had updated `global.json` to use the RTM SDK 10.0 instead of RC2 before merging.

stevenaw:
Thanks, I had forgotten. I'll do that shortly

PR #5044, opened 2025-10-16
Add PartitionFilter.TryCreate() test coverage

stevenaw:
I was browsing the codebase when I noticed there wasn't any direct test coverage over this function. So I've added it 😀

### DIFF
--- a/src/NUnitFramework/tests/Internal/Filters/PartitionFilterTests.cs
+++ b/src/NUnitFramework/tests/Internal/Filters/PartitionFilterTests.cs
@@ -106,5 +106,24 @@ namespace NUnit.Framework.Tests.Internal.Filters
             Assert.DoesNotThrow(() => _filter.ComputePartitionNumber(_testFixtureWithWithLongTestCaseNames.Tests[0]));
             Assert.DoesNotThrow(() => _filter.ComputePartitionNumber(_testFixtureWithWithLongTestCaseNames.Tests[1]));
         }
+
+        [TestCase("1/1n")]
+        [TestCase("1")]
+        public void TryCreateFailure(string input)
+        {
+            Assert.That(PartitionFilter.TryCreate(input, out var filter), Is.False);
+        }
+
+        [TestCase("1/2")]
+        [TestCase(" 1/2")]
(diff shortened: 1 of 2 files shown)


src/NUnitFramework/tests/Internal/Filters/PartitionFilterTests.cs
```diff
@@ -106,5 +106,24 @@ public void ComputeParitionNumberHandlesLongTestCaseName()
...
+
+        [TestCase("1/2")]
+        [TestCase(" 1/2")]
+        [TestCase("1/2 ")]
+        public void TryCreateSuccess(string input)
+        {
+            var result = PartitionFilter.TryCreate(input, out var filter);
+
+            Assert.That(result, Is.True);
+            Assert.That(filter!.PartitionNumber, Is.EqualTo(1));
```
stevenaw, line 125: question for the reviewer: I added `NotNullWhen` on the `TryCreate` method in the hopes that I wouldn't need to use the `!` operator here. I had thought the analyzer would already detect a situation like this where we reference it _after_ asserting that `result Is.True`. Am I misremembering or mis-decorating?

manfred-brands: Your annotation is correct and would have worked if you directly test the _result_ variable:
```csharp
if (result)
    Assert.That(filter.PartitionNumber, Is.EqualTo(1));
```

There is currently only an NUnit analyzer that detects `Assert.That(variable, Is.Not.Null)`
If you want one that recognizes the `NotNullWhen` and the `Is.True` you need to create an issue for that.

stevenaw: Thanks for confirming @manfred-brands . I must have been thinking of the `Is.Not.Null` case.

manfred-brands: To late now, but I thought of this on my way home, if you had added `Assert.That(filter, Is.Not.Null);` you would not have needed the null suppression operator.

PR #5049, opened 2025-11-12
Update build docs, global.json, and remove unused package from the test projects

stevenaw:
Relates to https://github.com/nunit/nunit/issues/5040

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -19,7 +19,7 @@ The development process for making code changes and submitting Pull Requests (PR
 
 There is a single Visual Studio solution, `nunit.sln`, which resides in the NUnit repository root.
 
-NUnit framework can be built from this solution using a .NET 6-capable IDE such as:
+NUnit framework can be built from this solution using a .NET-capable IDE such as:
 
 - [Visual Studio 2026](https://www.visualstudio.com/vs/) or newer (Windows)
 - [Visual Studio for Mac](https://www.visualstudio.com/vs/) (macOS).


BUILDING.md
```diff
@@ -21,13 +21,13 @@ There is a single Visual Studio solution, `nunit.sln`, which resides in the NUni
 
 NUnit framework can be built from this solution using a .NET 6-capable IDE such as:
 
-- [Visual Studio 2022](https://www.visualstudio.com/vs/) or newer (Windows)
+- [Visual Studio 2026](https://www.visualstudio.com/vs/) or newer (Windows)
```
stevenaw, line 24: Question for reviewer: Should we drop the explicit version reference to `2022` or `2026` here? The link will always allow the downloading of the latest version.

src/NUnitFramework/Directory.Packages.props
```diff
@@ -36,9 +36,6 @@
   <ItemGroup>
     <PackageVersion Include="BenchmarkDotNet" Version="0.15.0" />
   </ItemGroup>
-  <ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
```
stevenaw, line 39: This was unused and was prompting warnings by the SDK

BUILDING.md
OsirisTerje: As this is information for a new developer, I believe it is good to state that they should use the version we specify.  Otherwise they might just use the version they have, and then things may not work as expected.

stevenaw: Thanks, that makes sense to me.
Thanks for fixing the build as well. I'll rebase this PR to pull in your changes

src/NUnitFramework/Directory.Packages.props
OsirisTerje: Yes, can go, also the ref to System.Security.Principal.Windows.  It is also in testdata.csproj and that can also go.

BUILDING.md
OsirisTerje: Not sure what happened actually, it worked in the PR branch, and then failed on main.  So a little bit of "panic mode" here :-) 

You'll fix the removal of the System.Security.Principal.Windows too, right ?

stevenaw: > You'll fix the removal of the System.Security.Principal.Windows too, right ?

Fixed, both in packages.props and in the related csproj file.
Works on my machine ( 🙂 ) and I've just pushed the updates now too for build confirmation

BUILDING.md
```diff
@@ -21,13 +21,13 @@ There is a single Visual Studio solution, `nunit.sln`, which resides in the NUni
...
 - [Visual Studio for Mac](https://www.visualstudio.com/vs/) (macOS).
 - [Visual Studio Code](https://code.visualstudio.com/)
    - Compile using the [build script](#script-build) instead if using VS Code (see below for details).
 - [Jetbrains Rider](https://www.jetbrains.com/rider/)
 
-On all platforms, you will need to install [.NET 6.0 SDK](https://www.microsoft.com/net/download/windows) or newer. 
+On all platforms, you will need to install [the latest stable SDK](https://www.microsoft.com/net/download/windows).
```
manfred-brands, line 30: GitHub doesn't allow me to comment on line 22 which mentions NET-6. I suggest to drop the SDK remark there.

stevenaw: Good catch! I've pushed a change which updates line 22. I'm not sure if it's what you had in mind, but can you let me know what you think? You should hopefully be able to push suggestions too

manfred-brands (approved):
Thanks @stevenaw . Nothing further.

PR #5052, opened 2025-11-14
Allow string comparison specification for string constraints

### LINKED ISSUE #3922: Expose a way to provide 'StringComparison' value to 'StartsWith' assertions
The `StartsWith` validations will currently perform the comparisons using the standard .NET value for `StringComparison`, which relies on the current culture for checking the characters.

This has caused a test to break on my application when converting an unit test from a NET 4.8 project to a .NET 5.0 project, because of the changes described here:
- https://docs.microsoft.com/en-us/dotnet/standard/base-types/string-comparison-net-5-plus

The following assertion works in .NET 4.8, but fails in .NET 5.0:
```csharp
Assert.That("my string", Does.Not.StartWith("\0"));
```

Due to the changes done with string comparison, the null character `\0` is treated differently now and the result changes.

If I change the code to do the comparison using `Ordinal` setting, it works as expected:
```csharp
Assert.IsTrue(!"my string".StartsWith("\0", StringComparison.Ordinal));
```

However, NUnit does't provide any means of specifying the `StringComparison` value, so I'm stuck with having to use more general assertions that don't provide as good of an error message when failing.

I propose that either:
- a second, optional parameter/overload is added to `StartsWith` constraint to pass `StringComparison` setting
- something similar to `.Using` method on the equals constraint, where you can pass the `StringComparison` value separately like `Does.Not.StartWith("something").Using(StringComparison.Ordinal)`

nancyhuyn:
Fixes #3922

- Add `.Using(StringComparison)` and `.Using(CultureInfo)` modifiers to base `StringConstraint `
- Add implementations for `StartsWithConstraint `and `EndsWithConstraint`
- Update tests and `SubstringConstraint` accordingly

### DIFF
--- a/src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
@@ -21,16 +21,24 @@ namespace NUnit.Framework.Constraints
         }
 
         /// <summary>
+        /// Modify the constraint to the specified comparison.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Thrown when a comparison type different
+        /// than <paramref name="comparisonType"/> was already set.</exception>
+        public new SubstringConstraint Using(StringComparison comparisonType)
+        {
+            // This method is needed because of binary backward compatibility.
+            return (SubstringConstraint)base.Using(comparisonType);
+        }
+
+        /// <summary>
         /// Test whether the constraint is satisfied by a given value.
         /// </summary>
         /// <param name="actual">The value to be tested</param>
         /// <returns>True for success, false for failure</returns>
         protected override bool Matches(string? actual)
         {
-            if (actual is null)
-                return false;
-
-            return actual.IndexOf(expected) >= 0;
+            return actual is not null && actual.IndexOf(expected) >= 0;
         }
 
         /// <summary>
@@ -42,10 +50,7 @@ namespace NUnit.Framework.Constraints
         /// <returns>True for success, false for failure</returns>
         protected override bool Matches(string? actual, StringComparison stringComparison)
         {
-            if (actual is null)
-                return false;
-
-            return actual.IndexOf(expected, stringComparison) >= 0;
+            return actual is not null && actual.IndexOf(expected, stringComparison) >= 0;
         }
 
         /// <summary>
@@ -57,10 +62,8 @@ namespace NUnit.Framework.Constraints
         /// <returns>True for success, false for failure</returns>
         protected override bool Matches(string? actual, CultureInfo cultureInfo)
         {
-            if (actual is null)
-                return false;
-
-            return cultureInfo.CompareInfo.IndexOf(actual, expected, caseInsensitive ? CompareOptions.IgnoreCase : CompareOptions.None) >= 0;
+            return actual is not null && cultureInfo.CompareInfo.IndexOf(actual, expected,
+                caseInsensitive ? CompareOptions.IgnoreCase : CompareOptions.None) >= 0;
         }
     }
 }


nancyhuyn:
@dotnet-policy-service agree

manfred-brands:
@nancyhuyn Thanks for your contribution.

Instead of adding the `Using` to both `StartsWith` and `EndsWith`, we should add it to the base class `StringConstraint` instead.
That way all derived constraints will get a `Using` option.

You then can add the `_cultureInfo` to the base class and add an overload to `Match` which has a `CultureInfo` parameter when it is set.

Would you be so kind to implement it that way?

OsirisTerje:
@nancyhuyn Thanks for contributing!   We love that!

1.  The failed builds are only our code style errors, easy to fix:

```txt
Error: /Users/runner/work/nunit/nunit/src/NUnitFramework/tests/Constraints/EndsWithConstraintTests.cs(121,1): 

error SA1507: Code should not contain multiple blank lines in a row 
```

2,  Review The CoPilot suggestions, and commit them if they make sense to you.    If not, @manfred-brands will guide you there too :-)

src/NUnitFramework/tests/Constraints/StartsWithConstraintTests.cs
```diff
@@ -37,6 +39,103 @@ public void RespectsCulture()
...
+            Assert.DoesNotThrow(() =>
+            {
+                var newConstraint = startsWithConstraint.Using(StringComparison.CurrentCultureIgnoreCase).IgnoreCase;
+            });
+
+            var stringConstraint = (StringConstraint)new StartsWithConstraint("hello");
+            Assert.DoesNotThrow(() =>
+            {
+                var newConstraint = (StartsWithConstraint)stringConstraint.IgnoreCase;
+                newConstraint = newConstraint.Using(StringComparison.CurrentCultureIgnoreCase);
```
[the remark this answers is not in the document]
nancyhuyn: I think the current code is easier to read

src/NUnitFramework/tests/Constraints/StartsWithConstraintTests.cs
```diff
@@ -37,6 +39,103 @@ public void RespectsCulture()
...
+            });
+        }
+
+        [Test]
+        public void UseSameComparisonTypes_DoesNotThrowException()
+        {
+            var startsWithConstraint = new StartsWithConstraint("hello");
+            Assert.DoesNotThrow(() =>
+            {
+                var newConstraint = startsWithConstraint.Using(StringComparison.CurrentCultureIgnoreCase).IgnoreCase;
```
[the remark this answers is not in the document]
nancyhuyn: Needed to invoke IgnoreCase without warnings

src/NUnitFramework/framework/Constraints/EndsWithConstraint.cs
```diff
@@ -28,8 +47,38 @@ public EndsWithConstraint(string expected) : base(expected)
...
+            return this;
+        }
+
+        /// <summary>
+        /// Modify the constraint to use the specified culture info.
+        /// </summary>
+        public EndsWithConstraint Using(CultureInfo cultureInfo)
+        {
+            if (_cultureInfo is null)
+                _cultureInfo = cultureInfo;
```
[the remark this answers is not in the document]
nancyhuyn: Unlike for `comparisonType`, this is currently the only way to set `cultureInfo` and I don't foresee any issues even if you do overwrite it. 
It was a different case for `comparisonType` and `caseInsensitive` since the two could clash.

src/NUnitFramework/framework/Constraints/StartsWithConstraint.cs
```diff
@@ -28,8 +47,38 @@ public StartsWithConstraint(string expected) : base(expected)
...
+            return this;
+        }
+
+        /// <summary>
+        /// Modify the constraint to use the specified culture info.
+        /// </summary>
+        public StartsWithConstraint Using(CultureInfo cultureInfo)
+        {
+            if (_cultureInfo is null)
+                _cultureInfo = cultureInfo;
```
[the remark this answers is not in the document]
nancyhuyn: Unlike for comparisonType, this is currently the only way to set cultureInfo and I don't foresee any issues even if you do overwrite it.
It was a different case for comparisonType and caseInsensitive since the two could clash.

src/NUnitFramework/framework/Constraints/EndsWithConstraint.cs
```diff
@@ -28,8 +47,38 @@ public EndsWithConstraint(string expected) : base(expected)
...
         protected override bool Matches(string? actual)
         {
-            var stringComparison = caseInsensitive ? StringComparison.CurrentCultureIgnoreCase : StringComparison.CurrentCulture;
-            return actual is not null && actual.EndsWith(expected, stringComparison);
+            if (_cultureInfo is not null)
+            {
+                return actual is not null && actual.EndsWith(expected, caseInsensitive, _cultureInfo);
+            }
+
+            return actual is not null && actual.EndsWith(expected, _comparisonType ?? StringComparison.CurrentCulture);
```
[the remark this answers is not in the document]
nancyhuyn: This is already handled by calling `Using(StringComparison.CurrentCultureIgnoreCase)` when `IgnoreCase` is invoked.

src/NUnitFramework/framework/Constraints/StartsWithConstraint.cs
```diff
@@ -28,8 +47,38 @@ public StartsWithConstraint(string expected) : base(expected)
...
         protected override bool Matches(string? actual)
         {
-            var stringComparison = caseInsensitive ? StringComparison.CurrentCultureIgnoreCase : StringComparison.CurrentCulture;
-            return actual is not null && actual.StartsWith(expected, stringComparison);
+            if (_cultureInfo is not null)
+            {
+                return actual is not null && actual.StartsWith(expected, caseInsensitive, _cultureInfo);
+            }
+
+            return actual is not null && actual.StartsWith(expected, _comparisonType ?? StringComparison.CurrentCulture);
```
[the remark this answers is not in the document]
nancyhuyn: This is already handled by calling Using(StringComparison.CurrentCultureIgnoreCase) when IgnoreCase is invoked.

src/NUnitFramework/framework/Constraints/StringConstraint.cs
```diff
@@ -73,16 +93,44 @@ protected StringConstraint(string expected)
...
+
+            return this;
+        }
+
+        /// <summary>
+        /// Modify the constraint to use the specified culture info.
+        /// </summary>
+        public virtual StringConstraint Using(CultureInfo culture)
+        {
+            cultureInfo ??= culture;
```
manfred-brands: As we can multiple `Usings`, either the last one should win or we should throw an exception when setting it twice.
We should also not allow setting both `StringComparison` and `CultureInfo` as they might conflict.

src/NUnitFramework/framework/Constraints/StringConstraint.cs
```diff
@@ -73,16 +93,44 @@ protected StringConstraint(string expected)
...
+        /// <summary>
+        /// Modify the constraint to use the specified comparison.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Thrown when a comparison type different
+        /// than <paramref name="strComparison"/> was already set.</exception>
+        public virtual StringConstraint Using(StringComparison strComparison)
+        {
+            if (comparisonType is null)
+                comparisonType = strComparison;
+            else if (comparisonType != strComparison)
```
manfred-brands: Add test on `cultureInfo is not null` as we should not allow setting both.

src/NUnitFramework/framework/Constraints/StringConstraint.cs
```diff
@@ -38,6 +40,24 @@ public abstract class StringConstraint : Constraint
...
 #pragma warning restore IDE1006
 
+        /// <summary>
+        /// The comparison type to use for string comparisons
+        /// </summary>
+#pragma warning disable IDE1006
+        // ReSharper disable once InconsistentNaming
+        // Disregarding naming convention for back-compat
+        protected StringComparison? comparisonType;
+#pragma warning restore IDE1006
```
manfred-brands: This is a new field and should follow the guidelines as it doesn't have to be backward compatible.
We should make the field `private` and pass it through in a `Match` overload or wrap it in a non-nullable property.
That way not all derived implementation have to add null tests and potentially use different defaults.

Same for the _cultureInfo_ field below.

src/NUnitFramework/framework/Constraints/StringConstraint.cs
```diff
@@ -92,7 +140,7 @@ public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
             var stringValue = ConstraintUtils.RequireActual<string>(actual, nameof(actual), allowNull: true);
 
-            return new ConstraintResult(this, actual, Matches(stringValue));
+            return new ConstraintResult(this, actual, cultureInfo is not null ? Matches(stringValue, cultureInfo) : Matches(stringValue));
```
manfred-brands: What about when _comparisonType_ is selected.
Suggest to add an `Match` overload which also expects `StringComparison`

src/NUnitFramework/framework/Constraints/EndsWithConstraint.cs
```diff
@@ -20,16 +21,32 @@ public EndsWithConstraint(string expected) : base(expected)
...
+        /// using the specified <see cref="CultureInfo"/> and case sensitivity specified by the constraint.
+        /// If <paramref name="cultureInfo"/> is not null, the comparison uses culture-specific rules;
+        /// otherwise, it falls back to the default string comparison logic.
+        /// </summary>
+        /// <param name="actual">The string value to test.</param>
+        /// <param name="cultureInfo">The culture information to use for the comparison.</param>
+        /// <returns></returns>
+        protected override bool Matches(string? actual, CultureInfo cultureInfo)
+        {
+            if (cultureInfo is not null)
```
manfred-brands: _cultureInfo_ should never be null as the base class doesn't call this method if it is.

To protect from outside calls you could add `Guard.ArgumentNotNull(cultureInfo, nameof(cultureInfo));`

src/NUnitFramework/framework/Constraints/StringConstraint.cs
manfred-brands, line 1: Github doesn't allow me to comment on existing lines, however you must update the `Description` property to say something about the _comparisonType_ or _cultureInfo_ that has been set.

src/NUnitFramework/framework/Constraints/StringConstraint.cs
```diff
@@ -73,16 +93,44 @@ protected StringConstraint(string expected)
...
                 return this;
             }
         }
 
+        /// <summary>
+        /// Modify the constraint to use the specified comparison.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Thrown when a comparison type different
+        /// than <paramref name="strComparison"/> was already set.</exception>
+        public virtual StringConstraint Using(StringComparison strComparison)
```
manfred-brands: I dislike shorthand, please change `strComparison` to `comparisonType`.
Mainly to match the one that was in `SubstringContraint`.
Once you renamed the field to `_comparisonType` you won't have conflicts.

src/NUnitFramework/tests/Constraints/EndsWithConstraintTests.cs
```diff
@@ -36,6 +38,102 @@ public void RespectsCulture()
...
+                Assert.That(() => endsWithConstraint.Using(StringComparison.CurrentCulture).IgnoreCase,
+                    Throws.TypeOf<InvalidOperationException>());
+                Assert.That(() => endsWithConstraint.Using(StringComparison.InvariantCulture).IgnoreCase,
+                    Throws.TypeOf<InvalidOperationException>());
+                Assert.That(() => endsWithConstraint.Using(StringComparison.InvariantCultureIgnoreCase).IgnoreCase,
+                    Throws.TypeOf<InvalidOperationException>());
+                Assert.That(() => endsWithConstraint.Using(StringComparison.Ordinal).IgnoreCase,
+                    Throws.TypeOf<InvalidOperationException>());
+                Assert.That(() => endsWithConstraint.Using(StringComparison.OrdinalIgnoreCase).IgnoreCase,
+                    Throws.TypeOf<InvalidOperationException>());
```
manfred-brands: I wonder if these should throw.
Yes, they could specify `StringComparison.CurrentCultureIgnoreCase` but this could be the same effect.

src/NUnitFramework/tests/Constraints/EndsWithConstraintTests.cs
```diff
@@ -36,6 +38,102 @@ public void RespectsCulture()
...
+                Assert.That(() => ((SubstringConstraint)endsWithConstraint.IgnoreCase).Using(StringComparison.InvariantCulture),
+                    Throws.TypeOf<InvalidOperationException>());
+                Assert.That(() => ((SubstringConstraint)endsWithConstraint.IgnoreCase).Using(StringComparison.InvariantCultureIgnoreCase),
+                    Throws.TypeOf<InvalidOperationException>());
+                Assert.That(() => ((SubstringConstraint)endsWithConstraint.IgnoreCase).Using(StringComparison.Ordinal).IgnoreCase,
+                    Throws.TypeOf<InvalidOperationException>());
+                Assert.That(() => ((SubstringConstraint)endsWithConstraint.IgnoreCase).Using(StringComparison.OrdinalIgnoreCase).IgnoreCase,
+                    Throws.TypeOf<InvalidOperationException>());
+            });
```
manfred-brands: The casts are now unnecessary as the implementation is on the base class.
Again not sure if these should throw or only modify the `StringComparison`?

src/NUnitFramework/tests/Constraints/StartsWithConstraintTests.cs
manfred-brands, line 1: Same remarks as for EndsWith

OsirisTerje:
Awesome work @nancyhuyn !

src/NUnitFramework/framework/Constraints/EndsWithConstraint.cs
nancyhuyn: Good point. Given the way `Matches` is currently being used, I don't think the guard is necessary so I've chosen not to add it.

src/NUnitFramework/tests/Constraints/EndsWithConstraintTests.cs
nancyhuyn: Good point, I think it's better ease of use if this didn't throw and only updated the `StringComparison`.

Updated the implementation so that this doesn't throw anymore. Which is what made me decide that restricting to only one `Using` call might be better.

If we allowed for multiple `Usings` **and** this scenario didn't throw, then `.Using(StringComparison.Ordinal).Using(StringComparison.OrdinalIgnoreCase)` would be valid. Which I don't think reads very well, so only allowing one `Using` would remove that edge case.

src/NUnitFramework/framework/Constraints/StringConstraint.cs
nancyhuyn: After dwelling on this point, I think it may be better to only allow the `Using` modifier to get called once, like in `CollectionOrderedConstraint`. I don't foresee a situation where it makes sense to allow different methods of string comparison to be configured.

This does change the existing behaviour of `SubstringConstraint` since you could chain them in the old implementation. But I don't think it was necessary there either. Not sure if it was done that way for a reason or if this change is an issue. 

I've pushed the changes to only allow Using to get called once, let me know your thoughts.

src/NUnitFramework/framework/Constraints/StringConstraint.cs
```diff
@@ -92,14 +163,58 @@ public override ConstraintResult ApplyTo<TActual>(TActual actual)
...
+        public override string Description
+        {
+            get
+            {
+                string desc = $"{descriptionText} {MsgUtils.FormatValue(expected)}";
+                if (caseInsensitive)
+                    desc += ", ignoring case";
+                if (_comparisonType is not null)
+                    desc += $", with comparison type {_comparisonType}";
+                if (_cultureInfo is not null)
```
manfred-brands: Could use an `else` as both cases are mutually exclusive.

src/NUnitFramework/framework/Constraints/SubstringConstraint.cs
```diff
@@ -22,47 +21,46 @@ public SubstringConstraint(string expected) : base(expected)
...
-            if (_comparisonType is null)
-                _comparisonType = comparisonType;
-            else if (_comparisonType != comparisonType)
-                throw new InvalidOperationException("A different comparison type was already set.");
+            if (actual is null)
+                return false;
 
-            return this;
+            return cultureInfo.CompareInfo.IndexOf(actual, expected, caseInsensitive ? CompareOptions.IgnoreCase : CompareOptions.None) >= 0;
         }
```
manfred-brands, line 67: We cannot remove this method as it will result in:
```text
   System.MissingMethodException : Method not found: 'NUnit.Framework.Constraints.SubstringConstraint NUnit.Framework.Constraints.SubstringConstraint.Using(System.StringComparison)'.
```

We can add it like:
```csharp
        /// <summary>
        /// Modify the constraint to the specified comparison.
        /// </summary>
        /// <exception cref="InvalidOperationException">Thrown when a comparison type different
        /// than <paramref name="comparisonType"/> was already set.</exception>
        public new SubstringConstraint Using(StringComparison comparisonType)
        {
            // This method is needed because of binary backward compatibility.
            return (SubstringConstraint)base.Using(comparisonType);
        }
```

nancyhuyn: Yes that makes sense, apologies I missed it.
Thank you so much for you patience and ongoing cooperation @manfred-brands :blush:

src/NUnitFramework/framework/Constraints/StringConstraint.cs
manfred-brands: I agree.

Issue #3808, opened 2021-04-01
Mark Assert.IsNotNull param with NotNull attribute


pepone:
https://github.com/nunit/nunit/blob/af52b9a3b93edb5e54442ed6121b718532202a8f/src/NUnitFramework/framework/Assert.Conditions.cs#L219

```
public static void IsNotNull([NotNull]object? anObject, string? message, params object?[]? args)
```

I think this will be helpful as you don't need to use `!` after checking `IsNotNull`

Similarlly IsTrue/IsFalse can also take advantage of `DoesNotReturnIf` attribute

jnm2:
The lines just above that one explain why we didn't add `NotNull`. This method can return and continue executing the test when the value is null, due to the Multiple Asserts feature.

https://github.com/nunit/nunit/blob/af52b9a3b93edb5e54442ed6121b718532202a8f/src/NUnitFramework/framework/Assert.Conditions.cs#L212-L215

If you install the NUnit.Analyzers package, we suppress the warnings for you when you're not inside a Multiple Asserts block so that you don't have to use `!`.

Existing discussion: https://github.com/nunit/nunit/issues/3717

pepone:
thanks @jnm2 I didn't knew about Assert.Multiple

PR #5054, opened 2025-11-15
ClassicAssertExtensions reestablish the classic Assert as nunit.framework.Assert class members.

### LINKED ISSUE #4613: Can you please give us back `Assert.IsNotNull`, `Assert.IsFalse` etc.?
Hello, I see from v3.13.3 to v4.0.1 one of the breaking changes was to remove the `IsNotNull`, `IsFalse`, `IsTrue` etc. methods from the `Assert` class.

However, I have static classes named `Is` all through the code base I am working on (for various reasons that I will not go into) and they keep name-colliding with the `NUnit.Framework.Is`.

Please put the `Assert.IsNotNull`, `Assert.IsFalse` and their friends back.  Thank you.

### LINKED ISSUE #4693: Re-add "legacy" Assert methods
I know this is going to be controversial. However, I strongly recommend re-adding the "legacy" Assert/CollectionAssert/StringAssert methods (eg, Assert.Equal). 

I believe Assert.That is objectively worse in every respect. In pre 4.0 versions, when you type Assert and then a dot, you get Intelliense for all the possible overloads (Equals, LessThan, etc). They are simple and straightforward to use. You do not need to go look up documentation.

Assert.That is the complete opposite. You have to "know" that the second argument requires a magical "Is" or "Does" construct. If you want to do something more complex you may have to chain multiple extremely undiscoverable method calls like Is.Not.... You are basically required to implicitly understand lots of extra-linguistic constructs that NUnit has added.

I understand that some people find the "That" syntax more pleasant to read. But I think it is way more difficult to write, way less discoverable, and a huge downgrade in usability for this library. 

The main reason I and others have continued to prefer NUnit over XUnit is the superior simplicity and usability. Removing the easily discoverable methods we've all known and loved for decade(s) is a huge mistake, IMHO.

OsirisTerje:
Fixes #4693 
Fixes #4613

The introduction of the new extension mechanism in C14 allow extension methods on static classes.  Using this we can create extension methods on the Assert and Assume static classes that "moves" the classic asserts back.  The extension methods live in the legacy framework project, so they don't interfere with the nunit.framework project itself.  

This is what we really wanted when we had to separate the class asserts from the constraint model in NUnit V4.

1. This works for projects with C14/SDK10.  Trying to use these with <C14 give a compiler error clearly stating you need C14.
2. NUnit projects require C14/SDK10
3. Added verification tests

## Implementation

The extension mechanism for the Assert class is done like this  (prototype):

image

When it is run with a failing test, you can see  (1) it looks like it belong to the framework Assert class and 2) we see it as an constraint test - because it still IS a classic assert (which as known, calls the constraint model). 

image

And as (1) above, you can see it is a classic method:

image


src/NUnitFramework/nunit.framework.legacy.tests/AssertWrapperSmokeTests.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+        }
+
+        [Test]
+        public void Type_Asserts_Work()
+        {
+            var obj = new Derived();
+            Assert.IsInstanceOf<Derived>(obj);
+            Assert.IsNotInstanceOf<Base>(new object());
+            Assert.IsAssignableFrom<Derived>(new Base());
```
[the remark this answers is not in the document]
OsirisTerje: It works as is

src/NUnitFramework/nunit.framework.legacy.tests/AssertWrapperSmokeTests.cs
```diff
@@ -0,0 +1,124 @@
...
+            Assert.IsNotAssignableFrom<Base>(new Derived());
```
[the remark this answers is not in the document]
OsirisTerje: It works as is

src/NUnitFramework/nunit.framework.legacy/LegacyExtensions.Comparisons.cs
```diff
@@ -0,0 +1,144 @@
...
+    /// <summary>
+    /// C#14 static extension methods for comparison assertions (Greater, Less, etc.)
+    /// </summary>
+    public static class LegacyAssertComparisonExtensions
+    {
+        extension(Assert)
+        {
+            #region Greater
+
+            public static void Greater(int arg1, int arg2, string? message = null, params object[] args) => Legacy.ClassicAssert.Greater(arg1, arg2, message ?? string.Empty, args);
```
manfred-brands: 3.14 doesn't have nullable _message_. It has two overloads, one with and one without _message_ parameter.
Should we not do the same here?

src/NUnitFramework/nunit.framework.legacy/LegacyExtensions.Numeric.cs
```diff
@@ -0,0 +1,130 @@
...
+    /// C#14 static extension methods for numeric assertions
+    /// </summary>
+    public static class LegacyAssertNumericExtensions
+    {
+        extension(Assert)
+        {
+            #region Zero
+
+            public static void Zero(int actual) => Legacy.ClassicAssert.Zero(actual);
+            public static void Zero(int actual, string? message, params object?[]? args) => Legacy.ClassicAssert.Zero(actual, message ?? string.Empty, args);
```
manfred-brands: Interesting, 3.14 does have a nullable _message_ here.

src/NUnitFramework/nunit.framework.legacy/LegacyExtensions.Types.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        extension(Assert)
+        {
+            #region IsAssignableFrom
+
+            public static void IsAssignableFrom(Type expected, object? actual, string? message = null, params object?[]? args)
+                => Legacy.ClassicAssert.IsAssignableFrom(expected, actual, message ?? string.Empty, args);
+
+            public static void IsAssignableFrom(Type expected, object? actual)
+                => Legacy.ClassicAssert.IsAssignableFrom(expected, actual);
```
manfred-brands: The _message_ should not be optional, you already have the overload without a _message_ parameter.

src/NUnitFramework/nunit.framework.legacy/LegacyExtensions.cs
```diff
@@ -0,0 +1,356 @@
...
+    public static class LegacyAssertExtensions
+    {
+        extension(Assert)
+        {
+            #region AreEqual
+
+            /// <summary>
+            /// Verifies that two doubles are equal considering a delta.
+            /// </summary>
+            public static void AreEqual(double expected, double actual, double delta, string? message = null, params object?[]? args)
```
manfred-brands: _message_ should not be optional.

src/NUnitFramework/nunit.framework.legacy/nunit.framework.legacy.csproj
```diff
@@ -3,7 +3,8 @@
   <PropertyGroup>
     <TargetFrameworks>$(NUnitLibraryFrameworks)</TargetFrameworks>
     <RootNamespace>NUnit.Framework.Legacy</RootNamespace>
-    <NoWarn/>
+    <LangVersion>latest</LangVersion>
+    <NoWarn>CS1591;IDE0005</NoWarn>
```
manfred-brands: Why add IDE0005 (unnecessary using directives)? We should remove what is not needed.

src/NUnitFramework/nunit.framework.legacy/LegacyExtensions.Comparisons.cs
OsirisTerje: Yes, agree

src/NUnitFramework/nunit.framework.legacy/LegacyExtensions.Numeric.cs
OsirisTerje: oh ?

src/NUnitFramework/nunit.framework.legacy/nunit.framework.legacy.csproj
OsirisTerje: Agree

OsirisTerje:
@manfred-brands   I only wrote the first prototype, simplified.  Then I asked Copilot Pro to add the others, in the same manner.  

So it may have missed a few things.

We could use cursor or claude to verify, and prompt to make them 1:1, since that is what I understand you found.


I gave CoPilot your first reponse, and got this answer back:

> I can see the problem - we're using optional parameters (e.g., string? message = null, params object?[]? args) which creates only ONE method signature, but the reviewer wants separate overloads to match the legacy code 1:1.

So it has been trying to make one extension method covering multiple classic assert overloads.

I agree we should have 1:1, makes it less prone to an inadvertent bug popping up.

src/NUnitFramework/Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 ﻿<Project>
 
   <PropertyGroup>
-    <LangVersion Condition="'$(MSBuildProjectExtension)' == '.csproj'">13</LangVersion>
+    <LangVersion Condition="'$(MSBuildProjectExtension)' == '.csproj'">14</LangVersion>
```
stevenaw, line 4: Good catch. It looks like we missed this earlier.
Should we continue to target a specific version here, or is `latest` sufficient?

src/NUnitFramework/nunit.framework.legacy/LegacyExtensions.Collection.cs
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using System;
+using System.Collections;
+
+// ReSharper disable once CheckNamespace
+namespace NUnit.Framework
```
stevenaw: question: Can you help me remember how we want these to be opted into?

Will this be as a separate nuget package containing types in the _same namespace_ or will it be in the same assembly as the framework but within a _different namespace_ ?

stevenaw:
Thanks @OsirisTerje 
I was able to get through a few files before having to step away. I left a few questions

src/NUnitFramework/Directory.Build.props
stevenaw: Oh wait. I think I had just been thinking about test projects. The current setup will allow us to continue to use the latest language features in framework packages even if they don't target the latest TFMs, won't it?

manfred-brands: Correct. This is about language features only.

OsirisTerje: My thought was that when it says the exact number, it also serves as a documentation - and reminder for our selves.  What happened now when I moved to 14, one test broke because of `Reverse()` had changed (which @manfred-brands of course told me). If it had been latest, the build had just suddenly broken.

src/NUnitFramework/nunit.framework.legacy/LegacyExtensions.Collection.cs
OsirisTerje: It will be in the NUnit.Framework namespace, and connected to the Assert there, but the implementation is in the Framework.Legacy project.   So there is no opt-in, it will just be there, and pop up on Intellisense when you are on C# 14.   Pretty cool actually :-) 
Now we distribute the legacy together with the framework, so all should be good. 


This is what we hoped should be ready earlier for V4, but which was delayed by MS. Extension methods on static classes.  Now it finally is in. 

And we should of course mention and thank [Joseph](https://github.com/jnm2) - this is actually his invention, as he worked with the C# language group.

OsirisTerje:
@manfred-brands I believe there should be 1:1 now.  Also renamed from Legacy to Classic extensions. Sounds better.  Inheritdoc fixed, and warning suppressions removed.
Added also a document there  classic_extensions.md to verify we got them all.

src/NUnitFramework/Directory.Build.props
stevenaw: Thanks for explaining. I like the explicit documentation aspect of this.

src/NUnitFramework/nunit.framework.legacy/LegacyExtensions.Collection.cs
stevenaw: Ah yes, I remember this coming up now as part of our discussions for how to handle the original split. Thanks!

stevenaw:
@OsirisTerje Thanks for including me as a reviewer on this one. I'll slowly dig through it but please don't wait for an approval from me if we find it at a place that you and @manfred-brands are happy with it

OsirisTerje:
@stevenaw I think most importantly is to know that you and @manfred-brands agree this is fine to do.  Most of the code is generated, as I mentioned above, based on a "template", so this is more about the principle.   

I see this as a way to achieve both separation of concerns internally in the code base, and making it easier for the users to use whatever they like.   I know we discussed this way back when we did V4, but at that time we didn't have any better options.  With the new extension mechanism we've got what we need.

src/NUnitFramework/nunit.framework.legacy.tests/AssertWrapperSmokeTests.cs
manfred-brands: I think we can delete this file as you now have explicit tests.

src/NUnitFramework/nunit.framework.legacy/ClassicAssertExtensions.Collection.cs
```diff
@@ -0,0 +1,145 @@
...
+    /// <summary>
+    /// C#14 static extension methods for CollectionAssert methods on Assert
+    /// </summary>
+    public static partial class ClassicAssertExtensions
+    {
+        extension(Assert)
+        {
+            #region AllItemsAreInstancesOfType
+
+                /// <inheritdoc cref="Legacy.CollectionAssert.AllItemsAreInstancesOfType(IEnumerable, Type)"/>
```
manfred-brands: Copilot got carried away. These methods never existed as `Assert.AllItemsAreInstanceOfType`.

I don't think we should wrap `CollectionAssert` etc methods.

src/NUnitFramework/nunit.framework.legacy/ClassicAssertExtensions.Directory.cs
manfred-brands, line 1: Copilot got carried away. Drop this file.

src/NUnitFramework/nunit.framework.legacy/ClassicAssertExtensions.File.cs
manfred-brands, line 1: Copilot got carried away. Drop this file.

src/NUnitFramework/nunit.framework.legacy/ClassicAssertExtensions.String.cs
manfred-brands, line 1: Copilot got carried away. Drop this file.

src/NUnitFramework/nunit.framework.legacy/nunit.framework.legacy.csproj
```diff
@@ -3,7 +3,6 @@
   <PropertyGroup>
     <TargetFrameworks>$(NUnitLibraryFrameworks)</TargetFrameworks>
     <RootNamespace>NUnit.Framework.Legacy</RootNamespace>
-    <NoWarn/>
```
manfred-brands, line 6: Reset this as now it inherits from Directory.Build.props which as a NoWarn of 1591 we use for our test projects.

manfred-brands (approved):
@OsirisTerje Copilot got overzealous, it created Assert extensions for method that were never part of Assert such as CollectionAssert, StringAssert etc.

I have removed those and updated the .md document.
I have also addressed the formatting issues.

I compared the Assert with the extensions and both have 184 methods.

src/NUnitFramework/nunit.framework.legacy/nunit.framework.legacy.csproj
manfred-brands: Done.

JarJasAskom:
The new/old Assert.IsNotNull implementation is a significant step backward (pun intended). Please consider fixing this.

It would be beneficial to annotate the parameter anObject with [NotNull], as this aligns with modern C# nullable reference types (C# 14) and improves static analysis and null-safety guarantees.

For example:

```
        public static void IsNotNull([NotNull] object? anObject)
        {
            NUnit.Framework.Legacy.ClassicAssert.IsTrue(anObject != null);
            _ = anObject ?? throw new ArgumentNullException(nameof(anObject));
        }
```

manfred-brands:
@JarJasAskom This cannot be done as explained in various other  tickets asking for the same such as #3808 

The issue was solved for the `Assert.That(an object, Is.Not.Null);` by the NUnit.Analyzers.
Now that `Assert.IsNotNull` is back that analyzer could be updated to recognize that for users that don't want to convert to the constraint syntax.

Issue #5055, opened 2025-11-15
Convert sln file to slnx format


stevenaw:
Building and testing the framework now requires the latest stable bits (as of today) in order to run the .NET 10 tests and use C#14 language features. I think this also means that any software a person or CI may have would also support slnx format. Visual Studio 2026, [Rider 2024.3](https://blog.jetbrains.com/dotnet/2024/10/04/support-for-slnx-solution-files/) and the .NET 10 SDK itself would all be examples of software which could build and run the tests.

I believe this means we should also be able to convert the sln file to slnx format without asking anything extra of contributors.

PR #5056, opened 2025-11-15
Convert sln file to slnx format

stevenaw:
Fixes https://github.com/nunit/nunit/issues/5055

NOTE: 2 areas I didn't change are:
- https://github.com/nunit/nunit/blob/dff03c93d7a33d00188e00036ecc02a579469e51/.gitattributes#L12
- https://github.com/nunit/nunit/blob/dff03c93d7a33d00188e00036ecc02a579469e51/NUnit.sln.DotSettings

The former seems specific to MonoDevelop for which support has been discontinued. The last release for it was 7 years ago which also indicates it is unlikely to be able to build the latest code anyways. In practice I think this would have the effect of `* text=auto` taking over here for slnx files.

The latter I am unsure of if it needs to match exactly or not for Rider. If no one is familiar with that then I can try to install it and find out

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,6 +1,9 @@
 # Building NUnit
 
-NUnit consists of three separate layers: the Framework, the Engine and the Console Runner. The source code is kept in two GitHub repositories, https://github.com/nunit/nunit and https://github.com/nunit/nunit-console.
+NUnit consists of multiple components across separate layers: the Framework, the Engine, the Adapter, and the Console Runner. The source code is kept in three GitHub repositories
+- https://github.com/nunit/nunit
+- https://github.com/nunit/nunit-vs-adapter
+- https://github.com/nunit/nunit-console
 
 ## Source Code
 


BUILDING.md
```diff
@@ -1,6 +1,6 @@
 # Building NUnit
 
-NUnit 3 consists of three separate layers: the Framework, the Engine and the Console Runner. The source code is kept in two GitHub repositories, https://github.com/nunit/nunit and https://github.com/nunit/nunit-console.
+NUnit consists of three separate layers: the Framework, the Engine and the Console Runner. The source code is kept in two GitHub repositories, https://github.com/nunit/nunit and https://github.com/nunit/nunit-console.
```
OsirisTerje: We have the adapter too.....  on same layer, sort of, as the console.runner.

OsirisTerje:
@stevenaw  Just tried it with Rider, and that worked.

Also pretty sure the dotsettings file is fine.  If it wasn't I would have seen tons of warnings, which I don't.  So, it looks good.

OsirisTerje (approved):

OsirisTerje:
@stevenaw  When I try to switch branch in VS, having loaded the slnx, VS insists on creating  a sln file.  Do you see the same ?

stevenaw:
Thanks @OsirisTerje 
I didn't see that when switching branches using the Visual Studio Git Explorer. Did you switch outside of VS using the command line? I am wondering if some state is not transitioned correctly by Visual Studio in some flows.

BUILDING.md
stevenaw: Very good point. I'll update that in this PR

OsirisTerje:
It works outside VS,  it is the VS that does this.

stevenaw:
Sorry @OsirisTerje  I think I understand now. From your description it sounds as if VS is getting stuck in "sln" mode when switching to a branch with only a `sln` file.

I'm having a hard time reproducing. I've tried:
1. Open VS when branch was on `main`
2. Switch to `slnx-format` in VS
and I've tried the reverse
1. Open VS when branch was on `slnx-format`
2. Switch to `main` in VS

I'm on VS 2026 Community v18.0.0, which seems latest. Is it something you only see when _switching_ branches between this one and others, or do you see it when opening VS with the slnx already there?

image

nunit.slnx
```diff
@@ -0,0 +1,63 @@
+<Solution>
+  <Folder Name="/.config/">
+    <File Path=".config-macos/dotnet-tools.json" />
+    <File Path=".config/dotnet-tools.json" />
+  </Folder>
+  <Folder Name="/.github/" />
+  <Folder Name="/.github/workflows/">
+    <File Path=".github/workflows/NUnit.CI.yml" />
+    <File Path=".github/workflows/NUnit.Myget.Publish.yml" />
+    <File Path=".github/workflows/NUnit.Nuget.Publish.yml" />
```
manfred-brands: There is also a `.github` folder under `Solution Items`. I prefer to keep the latter.

nunit.slnx
```diff
@@ -0,0 +1,63 @@
+<Solution>
+  <Folder Name="/.config/">
+    <File Path=".config-macos/dotnet-tools.json" />
+    <File Path=".config/dotnet-tools.json" />
+  </Folder>
```
manfred-brands: This should also be under solution items so that it is less prominent.

stevenaw:
Thanks for catching those @manfred-brands . I think I'll also remove the `.gitattributes` line I called out initially as a setting for `.sln` files doesn't make much sense after this merges.

Are you able to reproduce what @OsirisTerje is seeing? This is a new format we'd be moving towards so I'm happy to move slowly here if it means we're sure there's no issues.

manfred-brands:
@stevenaw Switching from a branch with an ".slnx" to a branch with a ".sln" will confuse VS.
The solution it had no longer exists, so it wants to safe it on exit.

VS caches things in its .vs folder. 
I normally exit VS, switch branches and open it again, unless I know there are no new or deleted projects.

stevenaw:
Thanks @manfred-brands 
That was my suspicion as well. I've had similar issues in the past with switching branches during csproj format conversion and removing the `.vs` folder had fixed it then. That issue would only be present until the branch with the conversion merged.  Still, I didn't want to make assumptions since I couldn't repro quite what @OsirisTerje described.

@OsirisTerje Did you have any further concerns about what you saw?

OsirisTerje:
I'm on travel now, but I believe you're on the right track.  I get that moving from a branch with slnx to a branch that doesn't have it is what us causing this.  But to me this seems to be a bug in vs, since it tries to create a sln file in the source branch.  I'll try to reach out to the vs devs if this is what happens.

nunit.slnx
stevenaw: Thanks for noticing this. It turns out that the "Solution Items" one was also missing an entry. Consolidating them under Solution Items has also allowed us to fix it

stevenaw:
Thanks @manfred-brands 

I'm going to go ahead and merge this now as it sounds like we are all comfortable with the changes this PR makes to source and that we believe the issue Terje noticed is a bug in VS itself (and one that will be minimized by not having to switch between branches containing sln and slnx after this merges)

PR #5058, opened 2025-11-18
Allow Retry on listed Exceptions

### LINKED ISSUE #1388: Retry Decorator does not retry test if fail cause is an exception
In nunit 3.2...

When a test method has the Retry(int) decorator set and fails because of an uncaught exception (Happens quite often in my API integration tests and is expected) nunit does not retry the test. 

If the fail goes through an nunit assert it retries successfully.

### LINKED ISSUE #2785: Retry attribute should handle unexpected exceptions
Most users that use the `Retry` attribute do so because the code under test is flaky. More often than not, that code will throw an unexpected exception rather than returning incorrect data, but the Retry attribute only retries on assertion failures, not on errors. The retry should retry both.

See https://stackoverflow.com/questions/49520854/remove-failure-message-from-multiple-assertion-failure-in-nunit/49534616?noredirect=1#comment86075785_49534616 and https://testingrepository.com/retry-failed-tests-in-nunit/

manfred-brands:
Fixes #2785
Fixes #1388 

This PR allows specifying a list of exception types to be ignored by the RetryCommand.
Syntax:
```csharp
[Retry(3, typeof(TimeoutException), typeof(OperationCanceledException))]
```

### DIFF
--- a/src/NUnitFramework/testdata/RetryFixture.cs
+++ b/src/NUnitFramework/testdata/RetryFixture.cs
@@ -218,6 +218,17 @@ namespace NUnit.TestData.RepeatingTests
         }
 
         [Test]
+        [Retry(tryCount: 3, RetryExceptions = [typeof(TimeoutException)])]
+        public void OnlyRetriesOnAllowedException()
+        {
+            Count = TestContext.CurrentContext.CurrentRepeatCount;
+            if (Count == 0)
+                throw new TimeoutException("forcing a retry on allowed TimeoutException");
+            else if (Count == 1)
+                throw new OperationCanceledException("forcing a retry on allowed OperationCanceledException");
+        }
+
+        [Test]
         [Retry(tryCount: 3, RetryExceptions = [typeof(Exception)])]
         public void RetriesButEventuallyFails()
         {
--- a/src/NUnitFramework/tests/Attributes/RetryAttributeTests.cs
+++ b/src/NUnitFramework/tests/Attributes/RetryAttributeTests.cs
@@ -106,8 +106,19 @@ namespace NUnit.Framework.Tests.Attributes
             RepeatingTestsFixtureBase fixture = (RepeatingTestsFixtureBase)Reflect.Construct(typeof(RetryWithRetryExceptionFixture));
             ITestResult result = TestBuilder.RunTestCase(fixture, "RetriesOnAllowedException");
 
-            Assert.That(fixture.Count + 1, Is.EqualTo(fixture.TearDownResults.Count), "expected the CurrentRepeatCount property to be one less than the number of executions");
-            Assert.That(0, Is.EqualTo(result.FailCount), "expected that the test passed final retry");
+            Assert.That(result.ResultState, Is.EqualTo(ResultState.Success), "expected that the test passed final retry");
+            Assert.That(fixture.Count, Is.EqualTo(2), "expected that the test needed 3 tries");
+        }
+
+        [Test]
+        public void RetryOnAllowedStopsOnOtherException()
+        {
+            RepeatingTestsFixtureBase fixture = (RepeatingTestsFixtureBase)Reflect.Construct(typeof(RetryWithRetryExceptionFixture));
+            ITestResult result = TestBuilder.RunTestCase(fixture, "OnlyRetriesOnAllowedException");
+
+            Assert.That(result.ResultState, Is.EqualTo(ResultState.Error), "expected that the test failed after 2nd retry");
+            Assert.That(fixture.Count, Is.EqualTo(1), "expected that the test stopped after 2nd retry");
+            Assert.That(result.Message, Does.Contain(typeof(OperationCanceledException).Name), "expected that the final exception was the OperationCanceledException");
         }
 
         [Test]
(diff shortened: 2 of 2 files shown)


src/NUnitFramework/framework/Attributes/RetryAttribute.cs
```diff
@@ -96,6 +134,23 @@ public override TestResult Execute(TestExecutionContext context)
...
+                foreach (var retryException in _retryExceptions)
+                {
+                    if (retryException.IsAssignableFrom(exceptionType))
+                    {
+                        return true;
+                    }
+                }
+
+                return false;
+            }
```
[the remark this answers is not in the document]
manfred-brands: Not true. If there are no allowed exceptions, then indeed any exception at all should return `false`.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
```diff
@@ -70,21 +95,34 @@ public override TestResult Execute(TestExecutionContext context)
...
 
-                    if (context.CurrentResult.ResultState != ResultState.Failure)
+                    if (caughtException is not null and not ResultStateException)
+                    {
+                        if (count == 0 || IsRetryException(caughtException) is false)
+                        {
+                            context.CurrentResult.RecordException(caughtException);
+                            break;
+                        }
+                    }
```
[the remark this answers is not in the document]
manfred-brands: The result is reset for the next retry and no other code is executed that could interrogate that result.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
```diff
@@ -70,21 +95,34 @@ public override TestResult Execute(TestExecutionContext context)
...
+
+                    if (context.CurrentResult.ResultState != ResultState.Failure &&
+                        IsRetryException(context.CurrentResult.RecordedException) is false)
```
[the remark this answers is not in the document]
manfred-brands: This is correct. The loop should break out in case of failures as per previous behaviour.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
```diff
@@ -47,16 +59,29 @@ public TestCommand Wrap(TestCommand command)
...
+                : this(innerCommand, tryCount, Array.Empty<Type>())
+            {
+            }
+
+            /// <summary>
+            /// Initializes a new instance of the <see cref="RetryCommand"/> class.
+            /// </summary>
+            /// <param name="innerCommand">The inner command.</param>
+            /// <param name="tryCount">The maximum number of repetitions.</param>
+            /// <param name="retryExceptions">A list of exception types, that trigger a retry when thrown.</param>
```
CharliePoole, line 73: Suggest "... that trigger a retry when thrown, in addition to the normal behavior of retrying only on an assertion failure." or similar.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
```diff
@@ -15,14 +15,26 @@ namespace NUnit.Framework
...
+            : this(tryCount, Array.Empty<Type>())
+        {
+        }
+
+        /// <summary>
+        /// Construct a <see cref="RetryAttribute" />
+        /// </summary>
+        /// <param name="tryCount">The maximum number of times the test should be run if it fails.</param>
+        /// <param name="retryExceptions">A list of exception types, that trigger a retry when thrown.</param>
+        public RetryAttribute(int tryCount, params Type[] retryExceptions)
```
stevenaw: @manfred-brands  How do you feel about exposing these as a property instead of as a `params` here? Params in the constructor here could be limiting if we want to apply other changes to this attribute in the future.

Maybe something like

```csharp
public Type[] RetryExceptions { get;set; } = Array.Empty<Type>()
```

A params constructor could also cause users to think something like this would retry _only_ those exceptions and not other conditions like assertion failure.
```csharp
[Retry(3, typeof(TimeoutException), typeof(OperationCanceledException))]
```

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
manfred-brands: Thanks @CharliePoole. Done.

src/NUnitFramework/framework/Attributes/RetryAttribute.cs
manfred-brands: I had that initially, but found a `params` array easier to use. But if you prefer the other way, that is fine by me.

Now with collections expression it is simple to make arrays.

src/NUnitFramework/testdata/RetryFixture.cs
```diff
@@ -203,4 +203,26 @@ public void PassesOnLastRetry()
             }
         }
     }
+
+    public class RetryWithRetryExceptionFixture : RepeatingTestsFixtureBase
```
stevenaw, line 207: Should we add a test where an exception is thrown that's not in the retry list? For example:

```csharp


        [Test]
        [Retry(tryCount: 3, RetryExceptions = [typeof(TimeoutException)])]
        public void NoRetriesOnDisallowedException()
        {
            Count = TestContext.CurrentContext.CurrentRepeatCount;
            if (Count == 0)
                Assert.Fail();
            else if (Count == 1)
                throw new OperationCanceledException("forcing a retry on allowed OperationCanceledException");
        }
```

stevenaw (approved):
Thanks @manfred-brands Changes look good and robust. I tried running it on my own machine and couldn't break it :)

I appreciate you changing it to use a property. 

I left an optional suggestion for a test case to add, but nothing that "must" happen. Thanks again, and it looks good to merge on my end when you are ready.

OsirisTerje:
@manfred-brands Can we merge this ?

src/NUnitFramework/testdata/RetryFixture.cs
manfred-brands: Test added.

stevenaw:
Thanks for having added the extra test @manfred-brands

stevenaw:
I've added an issue on the docs repo to ensure we don't forget to share this: https://github.com/nunit/docs/issues/1098

PR #5065, opened 2025-11-26
Execution hooks: Explicit tests moved to test data

### LINKED ISSUE #5062: Tests marked Explicit for being used as Testdata should be moved to NUnit.TestData project
There are a lot of (new) Explicit tests, that are intended as testdata for other tests.
These tests are not run, but they are logged - and discovered, so they cost time, and also creates larger logs.

We have a separate project for Testdata,  NUnit.Testdata where we place such tests.  

Example:

Image

Image

An alternative approach, used in the adapter (acceptance tests project), keep the tests as source text in the C# files, and build them when they are to be used.

Image

This works fine, but will of course increase run time of the tests, which are high as it is, so the first approach should be chosen.

TravellerD14:
- All Explicit tests in Execution hooks tests are moved to test data project.
- Explicit attributes removed for test fixtures inside test data.
- Optimized moved fixtures and attributes in test data for better reuse and test readability.
- Consolidated multiple classes inside test data into single files to reduce number of files inside test data.

Fixes #5062

### DIFF
--- a/src/NUnitFramework/testdata/ExecutionHooks/AfterTestHookTestFixtures.cs
+++ b/src/NUnitFramework/testdata/ExecutionHooks/AfterTestHookTestFixtures.cs
@@ -13,7 +13,7 @@ namespace NUnit.TestData.ExecutionHooks
         public void SetUp() => TestLog.LogCurrentMethod();
 
         [Test]
-        [ActivateAfterTestHook]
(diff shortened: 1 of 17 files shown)


TravellerD14:
This PR resolves the issue with explicit tests in Execution hooks tests as mentioned in the issue https://github.com/nunit/nunit/issues/5062

Issue #3385, opened 2019-10-01
TestCaseSource doesn't convert IConvertible arguments


Denis535:
This example throws: _System.ArgumentException: Object of type 'System.String' cannot be converted to type 'System.DateTime'_.

    private static readonly TestCaseData[] Args = new TestCaseData[] {
        new TestCaseData( "2000-01-01 12:30:00", "2000-01-01 13:30:00" )
    };

    [TestCaseSource( nameof( Args ) )]
    public void Test(DateTime from, DateTime to) {
    }

CharliePoole:
It's not clear why you are passing a string rather than a DateTime.

Deni35:
Because the first is shorter. In my code I have more than 3 arguments, so I must write more long line.

Of course this is not a critical issue. But it would be more convenient to use shorter option.

    new TestCaseData( "2000-01-01 12:30:00", "2000-01-01 13:30:00" )
    new TestCaseData( DateTime.Parse( "2000-01-01 12:30:00" ), DateTime.Parse( "2000-01-01 13:30:00" ) )

CharliePoole:
Alternatively, write the test to take strings, which it then converts to DateTimes. That's less repetitious, but still longer. Howowever, I agree that the conversion would make sense.

Deni35:
I'd like to keep tests code as clean as possible.

I made some useful extension methods.

    new TestCaseData( "2000-01-01 12:30:00".ToDateTime(), "2000-01-01 13:30:00".ToDateTime(),  "01:00".ToTimeSpan() )

But if you make implicit converting it would be very well!

Deni35:
Why `TestParameters.Arguments` hasn't setter? Could you make setter for this property?

I tried to make SetArguments extension method but it's not possible(

CharliePoole:
There is no `Arguments` property in `TestParameters`. If you have run parameters, your test can access them by name but `TestParameters` has no connection at all with `TestCaseDataSource` or other ways of specifying arguments to test methods.

Deni35:
I'm confused.
![img](https://user-images.githubusercontent.com/7755015/66040419-09583280-e520-11e9-89ea-fba86063ce03.png)

CharliePoole:
It's understandable! So am I. :smile: 

It seems we have two classes with the same name, `NUnit.Framework.TestParameters` and `NUnit.Framework.Internal.TestParameters`! Not good!

[...]

Issue #3219, opened 2019-04-27
Consolidate type conversion code 


stevenaw:
There's currently a few places in the codebase that NUnit converts types for the purpose of input/output to tests. The ones I've noticed are:
- https://github.com/nunit/nunit/blob/9a98666/src/NUnitFramework/framework/Internal/TypeHelper.cs#L230
- https://github.com/nunit/nunit/blob/df130ce/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs#L395
- https://github.com/nunit/nunit/blob/3a8ee04/src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs#L91
- https://github.com/nunit/nunit/blob/061f435/src/NUnitFramework/framework/Internal/Reflect.cs#L223

Over time, some of these have drifted out of sync with each other (for example, present master has a `TestCaseAttribute` which will implicitly convert `int` -> `long`, but a `RangeAttribute` which does not).

Most of these conversion functions all focus on the `IConvertible` types, and mimic a subset of C#'s implicitly convertible types. Based on recollection from @CharliePoole this was likely added for CompactFramework, and "may" no longer be necessary. https://github.com/nunit/nunit/issues/3125#issuecomment-450432111

At present there's 2 open PRs that incidentally consolidate + remove some of these functions in order to address other issues.
- https://github.com/nunit/nunit/pull/3098
- https://github.com/nunit/nunit/pull/3133

I think if we consolidate other usages, it would both simplify the codebase, as well as facilitate future change (or removal of the implicit C# conversion code, if that's a goal)

PR #5066, opened 2025-11-26
Add support for generic types for params parameter.

### LINKED ISSUE #5061: TestCase (+ TestCaseSource) throw exception when passing arguments to params array in generic method
I've noticed that the following will throw an exception on NUnit 4.4 and below

```csharp
[TestCase(1, 2, 3, 4)]
public void HandlesParamsArrayForGenerics<T>(T s1, T s2, params T[] array)
{
}
```

> System.NotSupportedException : Type is not supported.
>
>  Stack Trace: 
Array.InternalCreate(RuntimeType elementType, Int32 rank, Int32* pLengths, Int32* pLowerBounds)
Array.CreateInstance(Type elementType, Int32 length)
TestCaseAttribute.GetParametersForTestCase(IMethodInfo method)

This can also now be reproduced in main with TestCaseSource as they use a common helper for this:

```csharp
 [TestCase(1, 2, 3, 4)]
 [TestCaseSource(nameof(ParamsArrayFourStringArguments))]
 public void HandlesParamsArrayForGenerics<T>(T s1, T s2, params T[] array)
 {
 }

 private static IEnumerable ParamsArrayFourStringArguments
 {
     get
     {
         yield return new TestCaseData("a", "b", "c", "d").SetArgDisplayNames("new TestCaseData(\"a\", \"b\", \"c\", \"d\")");
         yield return new string[] { "a", "b", "c", "d" };
     }
 }
```

The exception is currently thrown from https://github.com/nunit/nunit/blob/main/src/NUnitFramework/framework/Internal/TestCaseParameters.cs#L140

manfred-brands:
Fixes #5061

### DIFF
--- a/src/NUnitFramework/framework/Internal/Reflect.cs
+++ b/src/NUnitFramework/framework/Internal/Reflect.cs
@@ -90,13 +90,13 @@ namespace NUnit.Framework.Internal
 
             if (parameterInfos.Length > 0)
             {
-                arguments = PopulateOptionalArgsAndParamsArray(null, ctor, arguments, parameterInfos);
+                arguments = PopulateOptionalArgsAndParamsArray(ctor, null, arguments, parameterInfos);
             }
 
             return ctor.Invoke(arguments);
         }
 
-        internal static object?[] PopulateOptionalArgsAndParamsArray(Type[]? typeArgs, MethodBase method, object?[] arguments, ParameterInfo[] parameterInfos)
+        internal static object?[] PopulateOptionalArgsAndParamsArray(MethodBase method, Type[]? typeArgs, object?[] arguments, ParameterInfo[] parameterInfos)
         {
             ParameterInfo lastParameter = parameterInfos.Last();
             var hasParamsArray = lastParameter.ParameterIsParamsArray();
--- a/src/NUnitFramework/framework/Internal/TestCaseParameters.cs
+++ b/src/NUnitFramework/framework/Internal/TestCaseParameters.cs
@@ -117,7 +117,7 @@ namespace NUnit.Framework.Internal
             if (parameters.Length > 0)
             {
                 // Special handling for params and optional arguments
-                Arguments = Reflect.PopulateOptionalArgsAndParamsArray(TypeArgs, method.MethodInfo, Arguments, parameters);
+                Arguments = Reflect.PopulateOptionalArgsAndParamsArray(method.MethodInfo, TypeArgs, Arguments, parameters);
                 argsProvided = argsNeeded;
             }
 


manfred-brands:
@stevenaw I think I covered most cases.

To be decided if we want to allow all C# built-in conversions or only the once supported by NUnit's ParamAttributeTypeConversions.

If we want to change it, then some nunit test in Range need changing as the C# built-in rules allow more conversion.
Those test tests that certain combinations are explicitly not supported.

Alternatively we can call a separate method to convert params members.

stevenaw:
Thanks @manfred-brands ! I appreciate all the edge cases and coverage you've accounted for here.

I admit, it was unintentional of me to add the special conversions for TestCaseSourceAttribute as part of my other PR. My understanding was that those conversions were added historically as a convenience feature for TestCaseAttribute as only limited types are supported to specified _into_ the attribute constructor at compile time. For example, you can specify a double but not a decimal, so NUnit internally supported converting them. TestCaseSource, on the other hand, happens at runtime where users have much more control over the types they can pass into the test method. When faced with feature requests in the past to support implicit conversions for TestCaseSource (or even other scenarios in TestCase), we've tended to instead recommend users instead just pass the types they want to receive (ex: https://github.com/nunit/nunit/issues/3385).

I see my PR has muddied those waters now. Had I caught that I certainly would've suggested a discussion before making such a change, but perhaps we could have the discussion now as a team. Personally, I'd prefer to keep TestCaseSource simple and _undo_ the parameter conversion behaviour I'd inadvertently added to it. @nunit/framework-team what are your thoughts?

There is also another historical piece to this, to my understanding. Support for the framework to emulate the C# implicit conversion spec by performing a "preconversion" of the arguments was added in order to support certain runtimes which didn't do it themselves. I understand .NET Framework CE was one such example. I don't believe that is something that still needs to be supported and it's likely that at least some of that conversion code could be removed. I had tried to start a discussion about it in https://github.com/nunit/nunit/issues/3761

manfred-brands:
Thanks @stevenaw, I didn't know we have soo many places doing conversions #3219 
To prevent inconsistency they certainly should be de-duplicated.

Regarding the removal, I don't think we should.

Even though `TestCaseSource` is runtime, they are often created from compile time constants assigned to `object`.
The compiler will align: `new double[] { 1, 2.1 }` for both constants to be `double`, but not `new object[] { 1, 2.1 }`.
This alignment is needed if those parameters are to be passed to the same `params` array.

There is no type safety in `TestCaseParameters` and NUnit has to convert the instances to the expected type, allowing for built-in conversion, like `int` to `short`.
The alternative would be for the user to have to specify `new object[] { (short) 1, (short) 2 }` as there are no suffixes for `byte` and `short`.

From [Built-in numeric conversions](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/numeric-conversions):

* _Any [integral numeric type](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types) is implicitly convertible to any [floating-point numeric type](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types)._

* _A value of a constant expression of type int (for example, a value represented by an integer literal) can be implicitly converted to sbyte, byte, short, ushort, uint, ulong, nint, or nuint, if it's within the range of the destination type_

stevenaw:
That's a fair example. I had been thinking differently, and more along the lines that "if a user wants an array of decimals then they should pass an array of decimals" so that it just works. If the data passed to the test isn't implicitly compatible then it might be a mistake by the user that they should be alerted about. 

If there are truly ambiguous cases for a generic params array then there is the `TestCaseData.TypeArgs` property that we direct them to in an error message. Encouraging users to explicitly pass intended data types could also help the framework in the long-term if we further plumb generics in this area.

In other words, I agree that adding these conversions here _would_ improve consistency between `TestCase` and `TestCaseSource` in the short term. I'm just a little concerned it could make other things harder in the long term.

OsirisTerje:
@stevenaw Can this PR be approved ?
@manfred-brands Anything more you need to do here ?

stevenaw:
@OsirisTerje I had a few concerns about the argument conversion aspect. Historically we have avoided that in TestCaseSource and instead encouraged users to simply pass their intended types rather than NUnit taking over and converting anything. Parameter conversion was accidentally introduced in my previous PR https://github.com/nunit/nunit/pull/5057. My preference would actually be that we remove that parameter conversion from TestCaseSource instead of building upon it.

I think @manfred-brands and I have initially differing opinions about it and I'd welcome more thoughts on the long term direction we want to take the attribute as a team.

OsirisTerje:
@stevenaw 

>  I'd welcome more thoughts on the long term direction we want to take the attribute as a team.

Does this comment also apply to #3125 and the PR #5079 ?

Also, should we move that discussion  (with some clarifications) to a Discussion item ?

stevenaw:
@OsirisTerje 

> Does this comment also apply to #3125 and the PR #5079 ?

At a glance, yes. If single is not implicitly convertible to int then that example working would rely on NUnit doing the conversion internally. 

I'd hoped a quick discussion might be possible here to get clarity for this PR, but perhaps a dedicated discussion is better for visibility. I could spin one up in a few hours

One argument _for_ adding the conversions that I would also offer here is cross-attribute consistency.

stevenaw:
> I'd hoped a quick discussion might be possible here to get clarity for this PR, but perhaps a dedicated discussion is better for visibility. I could spin one up in a few hours

@OsirisTerje Just coming back to this with a quick question in lieu of starting a larger discussion. Despite my preference not to have added the conversions within my own PR, I've also felt that I would be alright to keep the conversions in if we felt as a team that we wanted to go in that direction. It sounds like you may like the idea of adding the conversions based on your changes for https://github.com/nunit/nunit/pull/5079. Would you prefer to keep it in?

manfred-brands:
@stevenaw I would like this merged together with #5079 before addressing #3219 as there would be conflicts.
So if you could do a review would be great.

stevenaw:
Sorry for the delays here @manfred-brands 
Given past discussions seemed hesitant to add the parameter conversion and that my own PR accidentally introduced it, I thought it would be good to hear from at least one other team member is on board with the change in direction here before I go too far with a thorough review.

stevenaw:
@manfred-brands I see now that the parameter discussion has been open for about two weeks. It doesn't seem like there's been any objections raised which I am inclined to also take as agreement about the approach. 

It might be a day or two until I can review but I'll take a close look at this when I'm able. Thanks again for your patience on this one

src/NUnitFramework/framework/Internal/TestCaseParameters.cs
```diff
@@ -118,15 +118,31 @@ internal void AdjustArgumentsForMethod(IMethodInfo method)
...
+                            else
+                            {
+                                Array array = CreateParamsArray(method, argsNeeded, argsProvided, elementType);
+                                array.SetValue(lastArgument, 0);
+                                Arguments[argsProvided - 1] = array;
+                            }
+                        }
+                        else
+                        {
+                            // Should we replace with empty array or an array with one null object or leave as is?
```
stevenaw: question: Is this to account for a case such as:

```csharp
[TestCase(1, 2, null)]
[TestCase("a", "b", null)]
public static void ReferenceParams<T>(T a, T b, params T[] rest)
{
    if (rest is null) {
      Console.WriteLine("null");
    } else {
      Console.WriteLine(rest.Length);
    }
}
```

If so then I think  `rest` should be a null array instead of an empty array or array with a null inside in order to mimic what the runtime seems to do in that case. Both of the below would output "null" instead of 0 or 1 if called like a regular C# method

```csharp

ReferenceParams(1, 2, null);
ReferenceParams("a", "b", null);
```

If this is a change in behaviour from my first PR then thank you for catching it :)

manfred-brands: Your PR and before including 3.14 does wrap `null` into `[null]`.
My remark was to question this behaviour.

If you change it up to only have a `params` field you get different behaviour in `nunit`:
```csharp
        [TestCase("a", "b")]
        [TestCase()]
        [TestCase([null])]
        [TestCase(null)]
        public void ReferenceParams(params string?[]? rest)
        {
            if (rest is null)
            {
                Console.WriteLine("null");
            }
            else
            {
                Console.WriteLine(rest?.Length);
            }
        }
```

This result in an output:
```text
2
0
1
1
```

The last two are considered the same and in `Test Explorer` it only shows 3 tests (but it runs 4)

If I execute this in normal C#, I get:
```text
2
0
1
null
```

I agree with you we should emulate the same behaviour.

Changing that breaks the `HandlesParamsArrayWithExplicitNullArgument` test you added.
As well as the test `GenericParams` I added, but where I also questioned the behaviour.

@OsirisTerje This is a breaking change from 3.14 and 4.4.0, but one that wasn't previously tested.

The pre-existing behaviour is weird:
```csharp
        [TestCase(null)]
        public static void MethodWithParamsArrayNullParameter(params int[]? a)
        {
        }
```

This actually passes an array with 1 element, being the integer 0.

The cause (accidental?) is the test that checks if the argument passed has the correct type, in case the user passes aan array.
As `null` doesn't have a type, this test fails and it creates a new array with 1 element.

This might have been an unexpected side-effect of the original code.

I have updated the code to match the C# normal conventions.

OsirisTerje:
> It sounds like you may like the idea of adding the conversions based on your changes for https://github.com/nunit/nunit/pull/5079. Would you prefer to keep it in?

Yes, I do.  I'm thinking that it should follow the C# implicit conversions. As we're untyped, we (imho) should provide that conversion ourselves.  Otherwise I feel it is a bit of a "surprise" there for the users.

E.g. all of these are valid in C#:

```csharp
   int x = 42;
   float y = 42;
   float z= x;
   double w = z;
```

and the total table should be like this: 

| From   | Implicitly Convertible To                                     |
|--------|---------------------------------------------------------------|
| sbyte  | short, int, long, float, double, decimal                      |
| byte   | short, ushort, int, uint, long, ulong, float, double, decimal |
| short  | int, long, float, double, decimal                             |
| ushort | int, uint, long, ulong, float, double, decimal                |
| int    | long, float, double, decimal                                  |
| uint   | long, ulong, float, double, decimal                           |
| long   | float, double, decimal                                        |
| ulong  | float, double, decimal                                        |
| char   | ushort, int, uint, long, ulong, float, double, decimal        |
| float  | double                                                        |

or expressed as @manfred-brands does at https://github.com/nunit/nunit/pull/5066#issuecomment-3584491951, ref MS links at the bottom there.

The best would of course be that we were typed too, so that we could take advantage of the implicit conversions.

stevenaw:
@OsirisTerje 
I agree about implicit conversions. If it "just works" for regular method invocations in C# to pass arguments of one type into a method expecting another without explicit conversion then NUnit should do the same. My concern is around when normally an _explicit_ conversion would be required.

For example, something like this will work in NUnit to allow a mechanism of defining a date at compile-time in a `TestCase`
```csharp
[TestCase("2025-10-10")]
public static void TestDate(DateTime d)
{
}
```

However the following code would not compile which is the root of my initial hesitation to propagate certain parameter conversion behaviour to TestCaseSource

```csharp
DateTime x = "2025-10-10";
// or
TestDate("2025-10-10");
```

Looking closer at the code same in https://github.com/nunit/nunit/issues/3125 I agree with your initial statement that that _should_ be allowed as passing `int` into a `float` _is_ implicitly convertible. The error message indicates somewhere are internally doing the reverse which is not allowed by the C# spec. I agree that that would be an internal bug then in the framework and should definitely be fixed so that a TestCaseSource can pass `int` to `float`.

OsirisTerje:
@stevenaw Hmm.....  I noted when doing the opposite (double into int) I got a NUnit1001 warning from the analyzer.

I was not even aware that one could convert from string to DateTime, and frankly - that conversion is not global, it is only InvariantCulture - a kind of US Light.

It is kind of cool - although very surprised to see that functionality.   
image

and it happens here:

image
The culture is **hardcoded**......

Looking at the other code above
image

It seems I was a bit too fast here, I "believed" I saw the intrinsic conversions, but when looking further I think I understand your concern.  These are ALSO converting from higher to lower (e.g. int to byte) , meaning information is lost.

A little AI table:
image

But a test will still convert from e.g. float -> double, so why do we then do the intrinsic conversions ?   

image

image

And with that, I am pretty much lost here......
----------

That said:

If I get what this issue is trying to do, is to introduce type safety in parameters.  Then we should not need any extra conversions for values that can be intrinsic'ly converted, right ?

OsirisTerje:
So, when looking at  @manfred-brands code, I really like the split of `BuiltInNumericalConversions ` and `NUnitConversions`.

(Not quite fully typesafe though, and still needing intrinsic conversions - ARE we really sure we need that? )

But I see the code with the hardcoded InvariantCulture is still there.  I assume that can be handled in a separate PR later.  No one outside US can be using that anyway with their default culture. 

The other NUnitConversions, I assume that makes it easier to write `short s = 42` which is valid C#, and code like `short s = 42000` is an error in C#.   I assume that is what we're trying to handle?

stevenaw:
Agreed with @OsirisTerje I like how we've separated that conversion split in this PR.
I ran out of time tonight to do this review but I should be able to do it tomorrow

src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
```diff
@@ -28,6 +29,42 @@ namespace NUnit.Framework.Internal
     /// </summary>
     internal static class ParamAttributeTypeConversions
     {
+        /// <summary>
+        /// Set of implicit conversion for built-in numeric types.
+        /// https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/numeric-conversions#implicit-numeric-conversions
+        /// </summary>
+        private static readonly Dictionary<Type, HashSet<Type>> BuiltInNumericalConversions = new()
```
stevenaw: We have a similar mapping here https://github.com/nunit/nunit/blob/e774ff3616ed609d3ffff66cb9bcb50b90cc4c74/src/NUnitFramework/framework/Internal/Reflect.cs#L196
Although that one is private and yours seems to have been updated to reflect newer types like `nuint`, etc.

I don't know the "best" place for this to live long term right but I suspect we'll find that out as we work through https://github.com/nunit/nunit/issues/3219. For now, how would you feel about updating that mapping in `Reflect.cs` with the new types you've noticed and to call `Reflect.CanImplicitlyConvertTo()` until we can sort out the consolidation?

src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
```diff
@@ -88,22 +125,13 @@ public static bool TryConvert(object? value, Type targetType, [NotNullWhen(true)
...
-                convert = value is double || value is string || value is int;
-            }
-            else if (underlyingTargetType == typeof(DateTime))
-            {
-                convert = value is string;
-            }
+            bool convert = HasNUnitConversion(valueType, underlyingTargetType);
+
+            //if (convert is false)
+            //    convert = HasImplicitConversion(valueType, underlyingTargetType);
```
stevenaw: I'm not sure, was this accidentally left commented out?

src/NUnitFramework/framework/Internal/TestCaseParameters.cs
```diff
@@ -118,16 +118,25 @@ internal void AdjustArgumentsForMethod(IMethodInfo method)
...
                     {
-                        if (!lastParameterType.IsInstanceOfType(Arguments[argsProvided - 1]))
+                        object? lastArgument = Arguments[argsProvided - 1];
+                        if (lastArgument is not null)
                         {
-                            Array array = Array.CreateInstance(elementType, 1);
-                            array.SetValue(Arguments[argsProvided - 1], 0);
-                            Arguments[argsProvided - 1] = array;
+                            Type lastArgumentType = lastArgument.GetType();
+                            if (lastArgumentType.IsArray)
```
stevenaw: good catch

src/NUnitFramework/framework/Internal/TestCaseParameters.cs
```diff
@@ -195,5 +201,96 @@ internal void AdjustArgumentsForMethod(IMethodInfo method)
...
+                        {
+                            elementType = TypeArgs[i];
+                            break;
+                        }
+                    }
+                }
+                else
+                {
+                    // Try to infer the type from the provided arguments
+                    elementType = DetermineBestElementType(Arguments, argsNeeded - 1);
```
stevenaw: We have a `TypeHelper.TryGetBestCommonType()` method already. Would that work here?

src/NUnitFramework/framework/Internal/TestCaseParameters.cs
```diff
@@ -195,5 +201,96 @@ internal void AdjustArgumentsForMethod(IMethodInfo method)
...
+
+            return array;
+        }
+
+        private Type DetermineParamsElementType(IMethodInfo method, int argsNeeded, Type elementType)
+        {
+            if (elementType.ContainsGenericParameters)
+            {
+                // Is the type specified?
+                if (TypeArgs is not null)
```
stevenaw: Thanks for considering this

src/NUnitFramework/framework/Internal/TestCaseParameters.cs
```diff
@@ -195,5 +201,96 @@ internal void AdjustArgumentsForMethod(IMethodInfo method)
...
+            Array array = Array.CreateInstance(elementType, length);
+
+            for (int i = 0; i < length; i++)
+            {
+                object? value = Arguments[argsNeeded + i - 1];
+                if (elementType != typeof(object) &&
+                    value is not null &&
+                    !elementType.IsAssignableFrom(value.GetType()))
+                {
+                    value = ParamAttributeTypeConversions.Convert(value, elementType);
```
stevenaw: question: I can see you'd added a lot of type calculations before the code gets to this point. Do we still need to convert here or will the array type already be the maximally compatible type such that we wouldn't need to convert the individual arguments anymore?
I'm unsure if there's an edge case I'm still missing... maybe implicit conversions?

stevenaw (commented):
Thanks @manfred-brands 
It's late here by the time I've been able to start this but I've tried to at least leave a few comments and questions to at least get the ball rolling. I understand the broader "conversions" discussion may still be going with @OsirisTerje

src/NUnitFramework/framework/Internal/TestCaseParameters.cs
manfred-brands: Thanks for pointing that out, I should have looked for it instead of implementing a new one.
But that method has yet another list of conversions.

src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
manfred-brands: Replaced `ConvertibleValueTypes` with ` BuiltInNumericalConversions`

src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs
manfred-brands: Removed commented out code.

src/NUnitFramework/framework/Internal/TestCaseParameters.cs
manfred-brands: Method now used `TypeHelper.TryGetBestCommonType`

src/NUnitFramework/framework/Internal/TestCaseParameters.cs
manfred-brands: All tests seem to work without the explicit conversion.

manfred-brands:
@stevenaw I had to fix rebase conflicts after the work you did.

I'm done now, can you please re-review.

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -90,16 +90,16 @@ public static object Construct(Type type, object?[]? arguments)
...
             {
-                arguments = PopulateOptionalArgsAndParamsArray(arguments, parameterInfos);
+                arguments = PopulateOptionalArgsAndParamsArray(null, ctor, arguments, parameterInfos);
             }
 
             return ctor.Invoke(arguments);
         }
 
-        internal static object?[] PopulateOptionalArgsAndParamsArray(object?[] arguments, ParameterInfo[] parameterInfos)
+        internal static object?[] PopulateOptionalArgsAndParamsArray(Type[]? typeArgs, MethodBase method, object?[] arguments, ParameterInfo[] parameterInfos)
```
stevenaw: nitpick: I would find it more natural for `MethodBase` to come first in the argument list. It looks like that's how `DetermineParamsElementType` has been setup already. Can we swap the ordering of these first two arguments here so that it's consistent?

Ex:
```csharp
        internal static object?[] PopulateOptionalArgsAndParamsArray(MethodBase method, Type[]? typeArgs, object?[] arguments, ParameterInfo[] parameterInfos)
```

stevenaw:
Thanks @manfred-brands 
Glad you could build off my other PR :)

I've left an initial nitpick for now. I'll try and perform a proper review within the next 48 hours or so

PR #5075, opened 2025-12-04
Use separate messages for posting work after Shutdown and for having work 10s after shutdown.

### LINKED ISSUE #3348: Incorrect exception thrown when using an old SynchronizationContext
Moving from NUnit 3.11 to 3.12, I've noticed some major problems with async tests.
One such problem is an incorrect error message being displayed when a test attempts to reuse an already shut down context.

My original tests works fine in 3.11, but my simple repro case throws InvalidOperationException, while in 3.12 for both cases, I get "Work posted to the synchronization context did not complete within ten seconds. Consider explicitly waiting for the work to complete."
However this error is thrown immediately, not after 10 seconds.

I've been reading #3209 and debugging my own tests. I haven't been able to reproduce the core issue (where my tests work fine in 3.11) in a simpler test case, but I have reproduced the incorrect error being thrown.

Expected (3.11.0):
System.InvalidOperationException : This SingleThreadedTestSynchronizationContext has been shut down.

Observed (3.12.0):
Work posted to the synchronization context did not complete within ten seconds. Consider explicitly waiting for the work to complete.

We have several thousand tests, which take a minute or so to execute, and in the original application, the sync context is saved so other threads know which is the UI one. However for these tests, the context now expires apparently. If I attempt to use an old one after 10s have elapsed since it was created, it throws the above message.

This is running tests under VS 2017 15.9.7.
NUnit 3.12.0
NUnit3TestAdapter 3.13.0
.NET Framework 4.5 runtime (net45)

[SynchronizationContextTest.zip](https://github.com/nunit/nunit/files/3494445/SynchronizationContextTest.zip)

manfred-brands:
Fixes #3348

### DIFF
--- a/src/NUnitFramework/framework/Internal/SingleThreadedSynchronizationContext.cs
+++ b/src/NUnitFramework/framework/Internal/SingleThreadedSynchronizationContext.cs
@@ -4,7 +4,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Threading;
-using NUnit.Framework.Interfaces;
 
 namespace NUnit.Framework.Internal
 {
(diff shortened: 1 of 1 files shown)


OsirisTerje (approved):

OsirisTerje:
Is the MAC build error related at all - or is it the "normal" failing of MAc we've had from time to time for a while ?

manfred-brands:
The PR only changes the message in the exception and the MacOS, although consistent is not in that area.
The MacOS error is on an `Assert.Inconclusive` which somehow becomes Failed, but only on MacOS?

manfred-brands:
I have rebased the branch to see if that makes a difference.

PR #5077, opened 2025-12-05
Allow specifying type PropertiesComparerConfiguration in more places.

### LINKED ISSUE #5073: Generically Typed UsingPropertiesComparer Consistency
The overload `UsingPropertiesComparer` taking a config delegate was added in 4.4.0 to `AnyOfConstraint`, `CollectionItemsEqualConstraint`, `EqualConstraint` and `SomeItemsConstraint`, with an untyped config `PropertiesComparerConfigurationUntyped`. But an overload that takes `PropertiesComparerConfiguration<T>` was only added to `EqualConstraint` and `SomeItemsConstraint`.

The documentation for these constraints claims they should be able to take `Func<PropertiesComparerConfiguration, PropertiesComparerConfiguration>`, however that doesn't appear to be the case. For `AnyOfConstraint` and `CollectionItemsEqualConstraint`, only the `PropertiesComparerConfigurationUntyped` variant is allowed, and for `EqualConstraint` and `SomeItemsConstraint` only the untyped and generic typed versions are allowed.

Can these be updated to take the base type `PropertiesComparerConfiguration` to allow all configuration types to be used for all constraints that support `UsingPropertiesComparer` as in the docs? Or at the very least, can the generically typed `PropertiesComparerConfiguration<T>` be added to the two constraints missing that currently?

manfred-brands:
Fixes #5073

### DIFF
--- a/src/NUnitFramework/framework/Internal/TypeHelper.cs
+++ b/src/NUnitFramework/framework/Internal/TypeHelper.cs
@@ -439,13 +439,13 @@ namespace NUnit.Framework.Internal
 
         internal static bool HasCompilerGeneratedEquals(this Type type)
         {
-            if (type.GetCustomAttribute<CompilerGeneratedAttribute>() is not null)
+            if (type.HasAttribute<CompilerGeneratedAttribute>(false))
                 return true;
 
             var equalsMethod = type.GetMethod(nameof(type.Equals), BindingFlags.Instance | BindingFlags.Public,
                 null, [type], null);
 
-            return equalsMethod?.GetCustomAttribute<CompilerGeneratedAttribute>() is not null;
+            return equalsMethod?.HasAttribute<CompilerGeneratedAttribute>(false) is true;
         }
 
         private static readonly Type[] EqualsObjectParameterTypes = { typeof(object) };


OsirisTerje (approved):
lgtm

src/NUnitFramework/framework/Internal/TypeHelper.cs
```diff
@@ -439,6 +439,9 @@ internal static string FullName(this Type type)
 
         internal static bool HasCompilerGeneratedEquals(this Type type)
         {
+            if (type.GetCustomAttribute<CompilerGeneratedAttribute>() is not null)
```
stevenaw: nit: This follows the pattern of [the code below,](https://github.com/nunit/nunit/pull/5077/files#diff-72ad15671f7b6e15bbb3cb97f0c67e77704dc715a5263d9d3d32221b2eec6857R448) but I think it should be simpler and marginally quicker to use `type.HasAttribute<CompilerGeneratedAttribute>(true)` to avoid having to construct the attribute instance itself.

Nothing that would prevent merging, just a nice-to-have

manfred-brands: Thanks, I updated both place to use `HasAttribute`.

PR #5078, opened 2025-12-05
Fix for #4513 based on the suggestions.

### LINKED ISSUE #4513: Randomizer.GetString can return strings that have invalid surrogate pairs
System.String is UTF-16, and not all sequences of `char` values are valid UTF-16. The current implementation disregards this:

https://github.com/nunit/nunit/blob/9b5ec78f28a00077ef4928f3a62a21a6853283e6/src/NUnitFramework/framework/Internal/Randomizer.cs#L502-L510

This is only a concern if `System.Char.IsSurrogate` returns true for any character in the `allowedChars` parameter. Possibly, this could be resolved by throwing NotSupportedException for such inputs.

Alternatively, `allowedChars` could be enumerated with https://learn.microsoft.com/en-us/dotnet/api/system.globalization.stringinfo.gettextelementenumerator (available since .NET Framework 1.1) in order to not split apart char sequences that need to stay together in order to be meaningful. This would be a slower code path, so it would be desirable to still check to see if `System.Char.IsSurrogate` returns true for any character in the `allowedChars` parameter.

(For some target frameworks, there may be a vectorized way to compute the equivalent of `allowedChars.Any(char.IsSurrogate)`, I'm not sure.)

OsirisTerje:
Fixes #4513 
Handles emojies too, have a fast and a safe path.

### DIFF
--- a/src/NUnitFramework/framework/Internal/Randomizer.cs
+++ b/src/NUnitFramework/framework/Internal/Randomizer.cs
@@ -500,7 +500,7 @@ namespace NUnit.Framework.Internal
         /// Returns a random string of the specified length using characters from the specified set.
         /// </summary>
         /// <param name="outputLength">The desired length of the string in char units.</param>
-        /// <param name="allowedChars">The characters to use in building the string</param>
+        /// <param name="allowedChars">The characters to use in building the string.</param>
         /// <returns>A random string.</returns>
         /// <exception cref="ArgumentException">
         /// Thrown if the requested outputLength cannot be satisfied with the provided characters


src/NUnitFramework/framework/Internal/Randomizer.cs
```diff
@@ -492,34 +495,127 @@ public T NextEnum<T>()
...
-        private const int MaxStackAllocSize = 256;
 
         /// <summary>
-        /// Generate a random string based on the characters from the input string.
+        /// Returns a random string of the specified length using characters from the specified set.
         /// </summary>
-        /// <param name="outputLength">desired length of output string.</param>
-        /// <param name="allowedChars">string representing the set of characters from which to construct the resulting string</param>
-        /// <returns>A random string of arbitrary length</returns>
+        /// <param name="outputLength">The desired length of the string in char units</param>
```
stevenaw: Thanks for adding 'in char units' here as it's an important caveat now :)

stevenaw: request: I think we've tried to standardize our xmldocs to always end in a period. Could we re-add them to the `<param>` contents please?

src/NUnitFramework/framework/Internal/Randomizer.cs
```diff
@@ -492,34 +495,127 @@ public T NextEnum<T>()
...
-            void FillSpan(Span<char> data, string allowedChars)
+#if NET6_0_OR_GREATER
+        private string GetStringUnsafeSpan(int outputLength, string allowedChars)
+        {
+            // Use stackalloc for small strings, heap for large ones
+            const int maxStackLimit = 256; // Reasonable stack limit
+
+            if (outputLength <= maxStackLimit)
             {
+                Span<char> data = stackalloc char[outputLength];
```
stevenaw, line 542: question: Did you prefer this over `string.Create()` to avoid the closure over the delegate?

src/NUnitFramework/framework/Internal/Randomizer.cs
```diff
@@ -492,34 +495,127 @@ public T NextEnum<T>()
...
-            var data = new char[outputLength];
 
+            // Fall back to array for very large strings
+            return GetStringUnsafeArray(outputLength, allowedChars);
+        }
+#endif
+
+        private string GetStringUnsafeArray(int outputLength, string allowedChars)
+        {
+            var data = new char[outputLength];
```
stevenaw, line 555: request: If we prefer to avoid the closure of sting.Create(), could we instead then use ArrayPool on non-netfx ?

Something like (untested):
```csharp
            char[]? data = null;
            try
            {
#if !NETFRAMEWORK
                data = ArrayPool<char>.Shared.Rent(outputLength);
#else
                data = new char[outputLength];
#endif
                for (int i = 0; i < outputLength; i++)
                    data[i] = allowedChars[Next(0, allowedChars.Length)];
                return new string(data, 0, outputLength);
            }
            finally
            {
#if !NETFRAMEWORK
                if (data is not null)
                {
                    ArrayPool<char>.Shared.Return(data);
                }
#endif
            }
```

src/NUnitFramework/framework/Internal/Randomizer.cs
OsirisTerje: Sure!   Happy to take commit suggestions though :-)

src/NUnitFramework/framework/Internal/Randomizer.cs
OsirisTerje: Probably, but I am on thin ice here.  If you would like to change this in this PR, I am just happy to see that.  
I assume it will give a performance boost, right ?

src/NUnitFramework/framework/Internal/Randomizer.cs
OsirisTerje: I was mostly thinking of performance. And the alloc is safe here.  
Do you prefer string.Create ?  As you say, the alloc does avoid closure and also seems easier in this case.

src/NUnitFramework/framework/Internal/Randomizer.cs
stevenaw: I can put some up tonight 😁

src/NUnitFramework/framework/Internal/Randomizer.cs
stevenaw: I haven't benchmarked it but I suspect you probably have the fastest implementation. 🙂

`String.Create()` _might_ be faster if we could use a static lambda but the need to reference instance member `Next()` eliminates that option and would likely require us the closure allocation.

What you have is also most readable.

src/NUnitFramework/framework/Internal/Randomizer.cs
stevenaw: It "should" give a boost. I'd be happy to push a commit with that change in a bit after I benchmark it to confirm

stevenaw: Interesting!

Part of my motivation behind my questions in this PR has been to capture some of the different things considered and ruled out as these areas of the code have tended to attract optimization attempts. And I was surprised.

The ArrayPool implementation does halve the allocations but the overall speed is not offset by that, at least on my machine while I was running the benchmarks. I ran them 4 times and they consistently showed results similar to below.

```
| Method    | N     | Mean        | Error    | StdDev   | Ratio | Gen0   | Code Size | Allocated | Alloc Ratio |
|---------- |------ |------------:|---------:|---------:|------:|-------:|----------:|----------:|------------:|
| Original  | 256   |    469.7 ns |  3.43 ns |  3.21 ns |  1.00 | 0.1707 |     638 B |    1072 B |        1.00 |
| ArrayPool | 256   |    494.2 ns |  1.62 ns |  1.35 ns |  1.05 | 0.0849 |   2,927 B |     536 B |        0.50 |
|           |       |             |          |          |       |        |           |           |             |
| Original  | 1000  |  1,752.0 ns | 18.01 ns | 15.04 ns |  1.00 | 0.6447 |     638 B |    4048 B |        1.00 |
| ArrayPool | 1000  |  1,889.3 ns |  4.39 ns |  3.67 ns |  1.08 | 0.3223 |   3,192 B |    2024 B |        0.50 |
|           |       |             |          |          |       |        |           |           |             |
| Original  | 10000 | 17,296.7 ns | 85.49 ns | 71.39 ns |  1.00 | 6.3477 |     638 B |   40048 B |        1.00 |
| ArrayPool | 10000 | 18,975.4 ns | 79.56 ns | 70.52 ns |  1.10 | 3.1738 |   3,192 B |   20024 B |        0.50 |
```

The allocation savings are naturally higher as the size of the output grows, but I would _guess_ that random strings generated for tests are also more likely to be on the shorter side. I am going to decline pushing any changes here @OsirisTerje . Thanks for entertaining my exploration here.

src/NUnitFramework/framework/Internal/Randomizer.cs
stevenaw:
```suggestion
        /// <param name="outputLength">The desired length of the string in char units.</param>
```

src/NUnitFramework/framework/Internal/Randomizer.cs
```diff
@@ -492,34 +495,127 @@ public T NextEnum<T>()
...
 
         /// <summary>
-        /// Generate a random string based on the characters from the input string.
+        /// Returns a random string of the specified length using characters from the specified set.
         /// </summary>
-        /// <param name="outputLength">desired length of output string.</param>
-        /// <param name="allowedChars">string representing the set of characters from which to construct the resulting string</param>
-        /// <returns>A random string of arbitrary length</returns>
+        /// <param name="outputLength">The desired length of the string in char units</param>
+        /// <param name="allowedChars">The characters to use in building the string</param>
```
stevenaw:
```suggestion
        /// <param name="allowedChars">The characters to use in building the string.</param>
```

src/NUnitFramework/framework/Internal/Randomizer.cs
```diff
@@ -492,34 +495,127 @@ public T NextEnum<T>()
...
+        /// <returns>A random string</returns>
```
stevenaw:
```suggestion
        /// <returns>A random string.</returns>
```

PR #5079, opened 2025-12-05
Fixes Issue #3125

### LINKED ISSUE #3125: TestCaseSource "Single" cannot be converted to "int" issue

OsirisTerje:
Fixes #3125

This PR fixes Issue 3125, which involves type conversions. 
The static class TypeHelper had a lot of static methods, which I realized was only working as extension methods.  I happily made them so  - it helped me understand the codebase, but noticed too late that a "small" syntactical sideeffect flood happened, as Resharper updated all usages of TypeHelper.  
So this become more changes not really related to the issue at hand - I am very sorry for doing that.

I have pushed up a repro for 3125 here https://github.com/nunit/nunit.issues/tree/main/Issue3125   - this issue (#3125) have been dormant for quite some time, but did have  quite a lot of comments.

I noticed that there is a discussion about type conversions at https://github.com/nunit/nunit/pull/5066#issuecomment-3618145162.  I am not sure if it also applies here.  

I do however, find the usage expected in 3125 to be valid though, as it is expected C# behaviour.

### DIFF
--- a/src/NUnitFramework/tests/Internal/TypeHelperTests.cs
+++ b/src/NUnitFramework/tests/Internal/TypeHelperTests.cs
@@ -170,7 +170,7 @@ namespace NUnit.Framework.Tests.Internal
                 .Select(p => new RuntimeParameterInfo(p))
                 .ToArray();
 
-            // Original arguments 
+            // Original arguments
             object?[] originalArgs = new object?[] { string.Empty, 0 };
 
             // Convert for TestA (should convert int to float)


src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
```diff
@@ -224,20 +250,28 @@ private static bool CheckTestMethodSignature(TestMethod testMethod, MethodInfoCa
...
             if (arglist is not null && parameters is not null)
-                TypeHelper.ConvertArgumentList(arglist, parameters);
+            {
+                // Convert the arguments for this specific test method
+                // Since we cloned parms earlier, this only affects this test method's copy
+                parms!.Arguments = TypeHelper.ConvertArgumentList(arglist, parameters);
+            }
 
-            return true;
+            return parms;
```
manfred-brands: The method always returns the parameter _parms_. It might as well return nothing.
Previously the (boolean) return type was ignored.

Even though you overwrite the `Arguments` property, it is the same instance.
Alternatively you can postpone cloning to here so only do it if really needed.

src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
```diff
@@ -224,20 +250,28 @@ private static bool CheckTestMethodSignature(TestMethod testMethod, MethodInfoCa
...
         }
 
         private static bool MarkAsNotRunnable(TestMethod testMethod, string reason)
```
manfred-brands: This method could now return `void` as its return value is no longer used.
Ok, it is used in `CheckTestMethodAttributes` but that methods own return type is ignored.
It probably makes this method unnecessary and we could replace calls to `testMethod.MakeInvalid` directly.

src/NUnitFramework/tests/Internal/TypeHelperTests.cs
```diff
@@ -146,5 +149,89 @@ public override int GetHashCode()
...
+            // Create test parameters using reflection from real methods
+            var methodA = typeof(TestMethods).GetMethod(nameof(TestMethods.TestA));
+            var methodB = typeof(TestMethods).GetMethod(nameof(TestMethods.TestB));
+
+            var paramA = methodA!.GetParameters()
+                .Select(p => new RuntimeParameterInfo(p))
+                .ToArray();
+            var paramB = methodB!.GetParameters()
+                .Select(p => new RuntimeParameterInfo(p))
+                .ToArray();
```
manfred-brands: My preference is to put the _null suppression_ operator on where it occurs, not where it is used:

```suggestion
            var methodA = typeof(TestMethods).GetMethod(nameof(TestMethods.TestA))!;
            var methodB = typeof(TestMethods).GetMethod(nameof(TestMethods.TestB))!;

            var paramA = methodA.GetParameters()
                .Select(p => new RuntimeParameterInfo(p))
                .ToArray();
            var paramB = methodB.GetParameters()
                .Select(p => new RuntimeParameterInfo(p))
                .ToArray();
```

manfred-brands: Or if it is inside a test as here, verify just in case:

```suggestion
            var methodA = typeof(TestMethods).GetMethod(nameof(TestMethods.TestA));
            var methodB = typeof(TestMethods).GetMethod(nameof(TestMethods.TestB));

            Assert.That(methodA, Is.Not.Null);
            Assert.That(methodB, Is.Not.Null);

            var paramA = methodA.GetParameters()
                .Select(p => new RuntimeParameterInfo(p))
                .ToArray();
            var paramB = methodB.GetParameters()
                .Select(p => new RuntimeParameterInfo(p))
                .ToArray();
```

src/NUnitFramework/tests/Internal/TypeHelperTests.cs
```diff
@@ -146,5 +149,89 @@ public override int GetHashCode()
...
+        private class FloatComparer : IComparer<float>
+        {
+            public static readonly FloatComparer Instance = new FloatComparer();
+
+            public int Compare(float x, float y)
+            {
+                return x.CompareTo(y);
+            }
+        }
+
```
manfred-brands: Not used, delete:
```suggestion
```

src/NUnitFramework/framework/Internal/TypeHelper.cs
```diff
@@ -171,223 +365,51 @@ public static bool TryGetBestCommonType(Type? type1, Type? type2, [NotNullIfNotN
...
         /// <param name="parameters">A ParameterInfo[] whose types will be used as targets</param>
-        public static void ConvertArgumentList(object?[] arglist, IParameterInfo[] parameters)
+        public static object?[] ConvertArgumentList(object?[] arglist, IParameterInfo[] parameters)
         {
             System.Diagnostics.Debug.Assert(arglist.Length <= parameters.Length);
 
-            for (int i = 0; i < arglist.Length; i++)
+            // Clone the array to avoid modifying the original when the same TestCaseData is used by multiple test methods
+            object?[] convertedArgs = new object?[arglist.Length];
+            Array.Copy(arglist, convertedArgs, arglist.Length);
```
manfred-brands: Together with my other remark, maybe we can do this lazily.
Only if we get inside the `if (convert)` and `convertedArgs is null` we make a copy.

src/NUnitFramework/framework/Internal/TypeHelper.cs
```diff
@@ -15,103 +15,297 @@ namespace NUnit.Framework.Internal
     /// </summary>
     public static class TypeHelper
     {
-        private const int STRING_MAX = 40;
-        private const int STRING_LIMIT = STRING_MAX - 3;
-        private const string THREE_DOTS = "...";
+        private const int StringMax = 40;
+        private const string ThreeDots = "...";
```
manfred-brands: `ThreeDots` is not used (and wasn't previously)

src/NUnitFramework/framework/Internal/TypeHelper.cs
```diff
@@ -15,103 +15,297 @@ namespace NUnit.Framework.Internal
...
-        public static string GetDisplayName(Type type)
+        /// <param name="type">The type to evaluate.</param>
+        extension(Type type)
         {
-            if (type.IsGenericParameter)
-                return type.Name;
+            /// <summary>
+            /// Determines whether the specified type represents a ValueTuple.
+            /// </summary>
+            /// <returns><c>true</c> if the type is a ValueTuple; otherwise, <c>false</c>.</returns>
```
manfred-brands: I get a DOC104 here.
```suggestion
            /// <returns><see langword="true"/> if the type is a <see cref="ValueTuple"/>; otherwise, <see langword="false"/>.</returns>
</returns>
```

src/NUnitFramework/framework/Internal/TypeHelper.cs
```diff
@@ -15,103 +15,297 @@ namespace NUnit.Framework.Internal
...
+            public bool IsValueTuple()
+            {
+                return type.IsTupleInternal("System.ValueTuple");
+            }
 
-            if (type.IsGenericType)
+            /// <summary>
+            /// Determines whether the specified type represents a Tuple.
+            /// </summary>
+            /// <returns><c>true</c> if the type is a Tuple; otherwise, <c>false</c>.</returns>
```
manfred-brands: Same DOC104 here:

```suggestion
            /// <returns><see langword="true"/> if the type is a <see cref="Tuple"/>; otherwise, <see langword="false"/>.</returns>
</returns>
```

manfred-brands (changes requested):
A few suggestions.
I don't know if doing the copies of `TestCaseParameters` and its Arguments lazily is worth the effort.

src/NUnitFramework/framework/Internal/TypeHelper.cs
OsirisTerje: Done

src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
OsirisTerje: Done

src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
OsirisTerje: Done, except postpone, will look into that

src/NUnitFramework/framework/Internal/TypeHelper.cs
OsirisTerje: Done

OsirisTerje:
@manfred-brands After the last few changes, including merging from main, one test fails here.   I does go through the TypeHelper ConvertArgumentList method, but can't see why it should break there.  It might be some of the other small changes, but this here was the latest change.
image

manfred-brands:
@OsirisTerje You seemed to have lost a change from me in
 `TypeHelper.HasCompilerGeneratedEquals()`
An anonymous type has a compiler generated Equals overload.
However the CompilerGeneratedAttribute is only on the class, not the method.
See my last changes in main.

OsirisTerje:
Thanks @manfred-brands !  That was the thing!!

src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
```diff
@@ -122,7 +122,25 @@ public IEnumerable<TestMethod> BuildFrom(IMethodInfo method, Test? suite)
...
             {
                 count++;
-                yield return _builder.BuildTestMethod(method, suite, parms);
+
+                // Clone the parameters to ensure each test method gets its own independent copy
+                // This prevents issues when the same TestCaseData is reused across multiple test methods
+                object?[] clonedArgs = new object?[parms.Arguments.Length];
+                Array.Copy(parms.Arguments, clonedArgs, parms.Arguments.Length);
+
+                var clonedParms = new TestCaseParameters(clonedArgs)
```
manfred-brands: I didn't know, but the `TestParameters` [constructor](https://github.com/nunit/nunit/blob/main/src/NUnitFramework/framework/Internal/TestParameters.cs#L66) already copies the arguments. In addition, in `ConvertArgumentList` where we DO modify the parameters we make another copy.
However, the `GetTestCasesFor` modifies the `Arguments`  (in `AdjustArgumentsForMethod`) before this part has a chance to clone them. This was introduced by @stevenaw in #5057 and the side effect missed in my reviews.

This is one of the reason at my work we now use immutable classes so nothing can modify them, only return updated clones.

There are 2 ways to fix this:
1. Move the clone into `GetTestCasesFor`. That way the originally returned instances are never modified. Downside is we might copy unnecessary. But as I don't know how many other places modify their input, it is the safest bet.
2. Update `AdjustArgumentsForMethod` to use the same lazy copy as `ConvertArgumentList`, however this method also changes the type of the `ExpectedResult`.

As I cannot guarantee nothing else modifies its inputs, I moved for option 1.

src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
```diff
@@ -224,20 +243,26 @@ private static bool CheckTestMethodSignature(TestMethod testMethod, MethodInfoCa
...
-
-            return true;
+            {
+                // Convert the arguments for this specific test method
+                // ConvertArgumentList returns a new array, so we update parms.Arguments
+                parms!.Arguments = TypeHelper.ConvertArgumentList(arglist, parameters);
+            }
         }
 
         private static bool MarkAsNotRunnable(TestMethod testMethod, string reason)
```
manfred-brands: Converted this method into `void`.

src/NUnitFramework/tests/Attributes/TestCaseSourceIssueTests.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+namespace NUnit.Framework.Tests.Attributes
+{
+    /// <summary>
+    /// Verify that Issue3125 https://github.com/nunit/nunit/issues/3125 is resolved.
+    /// Before 4.5 this will fail/crash due to arguments not being converted correctly.
+    /// </summary>
+    public class TestCaseSourceIssue3125Tests
```
manfred-brands, line 11: I augmented this class with more use cases that show other problems.

src/NUnitFramework/tests/Internal/TypeHelperTests.cs
```diff
@@ -146,5 +149,79 @@ public override int GetHashCode()
...
 
         #endregion
+
+        #region ConvertArgumentList
+
+        [Test]
+        public void ConvertArgumentList_WithMultipleTestMethodsUsingSameTestCaseData()
+        {
+            // Regression test for Issue #3125
+            // When the same TestCaseData is used by multiple test methods with different parameter types,
```
manfred-brands: As this test doesn't use `TestCaseData`, it doesn't actually proof it was fixed.
It is the unit test here that explicitly clones.
I'm happy to keep the test to show it can convert types, but maybe adjust the comment?

The other test added `TestCaseSourceIssue3125Tests` does verify the actual use case.

src/NUnitFramework/framework/Internal/Builders/NUnitTestCaseBuilder.cs
manfred-brands: Done.

src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs
OsirisTerje: Nice!

src/NUnitFramework/tests/Internal/TypeHelperTests.cs
OsirisTerje: Done

PR #5083, opened 2025-12-09
AssignableTo/From among value types or other implicit conversions

### LINKED ISSUE #5082: `AssignableTo/From` among value types
Referring to [AssignableToConstraint](https://docs.nunit.org/articles/nunit/writing-tests/constraints/AssignableToConstraint.html) and [AssignableFromConstraint](https://docs.nunit.org/articles/nunit/writing-tests/constraints/AssignableFromConstraint.html):

Shouldn't these constraints return true on compatible value types?
```
Assert.That(MethodReturningInt(), Is.AssignableTo<long>());
Assert.That(MethodReturningFloat(), Is.AssignableTo<double>());
```

If the above behavior is intended, this should be documented, e.g. "the constraint is limited to reference types".

And of course the behavior should be backed by test cases. Currently there seem no such, e.g. `AssignableFromTest`, `AssignableToTest`, `AssertionHelperTests` show very limited coverage.

stevenaw:
Fixes #5082 

`int` -> `long` is not "assignable" in the strictest CLR sense that `Type.IsAssignableTo` rules follow as the compiler will need to emit a call to _convert_ from the original type to the target type via the implicit conversion operator.

In practice though, no _explicit_ action is required to assign variables of one type to another for these cases, so we're used to seeing and thinking of code below as "assign an int to a long" so I think it makes sense to include the implicit operator as part of our to/from constraint.

```csharp
int x = 1;
long y = x;
```

### DIFF
--- a/src/NUnitFramework/tests/Constraints/AssignableFromConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/AssignableFromConstraintTests.cs
@@ -1,5 +1,6 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
+using System;
 using NUnit.Framework.Constraints;
 using static NUnit.Framework.Tests.Constraints.AssignableTestScenarios;
 
@@ -25,5 +26,34 @@ namespace NUnit.Framework.Tests.Constraints
             private static readonly object[] SuccessData = [new TTo(), new TFrom(), new object()];
             private static readonly object[] FailureData = [new TestCaseData(new B(), "<" + typeof(B).FullName + ">")];
         }
+
+        [TestCaseSource(nameof(SuccessCases))]
+        public static void CanAssignFrom(object? actual, Type type)
+        {
+            Assert.That(actual, Is.AssignableFrom(type));
+        }
+
+        private static readonly TestCaseData[] SuccessCases =
+        [
+            new TestCaseData(new object(), null),
+            new TestCaseData(42.0, typeof(int)),
+            new TestCaseData(42.0, typeof(double)),
+            new TestCaseData(42.0, typeof(float)),
+            new TestCaseData(new D1(), typeof(D2)),
+            new TestCaseData(new D1(), typeof(D3)),
+        ];
+
+        [TestCaseSource(nameof(FailureCases))]
+        public static void CanNotAssignFrom(object? actual, Type type)
+        {
+            Assert.That(actual, Is.Not.AssignableFrom(type));
+        }
+
+        private static readonly TestCaseData[] FailureCases =
+        [
+            new TestCaseData(42.0f, typeof(double)),
+            new TestCaseData(new D2(), typeof(D1)),
+            new TestCaseData(new D3(), typeof(D1)),
+        ];
     }
 }
--- a/src/NUnitFramework/tests/Constraints/AssignableToConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/AssignableToConstraintTests.cs
@@ -1,5 +1,6 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
+using System;
 using NUnit.Framework.Constraints;
 using static NUnit.Framework.Tests.Constraints.AssignableTestScenarios;
 
@@ -25,5 +26,37 @@ namespace NUnit.Framework.Tests.Constraints
             private static readonly object[] SuccessData = [new TTo(), new TFrom()];
             private static readonly object[] FailureData = [new TestCaseData(new object(), "<" + typeof(object).FullName + ">")];
         }
+
+        [TestCaseSource(nameof(SuccessCases))]
+        public static void CanAssignTo(object? actual, Type type)
+        {
+            Assert.That(actual, Is.AssignableTo(type));
+        }
+
+        private static readonly TestCaseData[] SuccessCases =
+        [
+            new TestCaseData(null, typeof(object)),
+            new TestCaseData(null, typeof(int?)),
+            new TestCaseData(42, typeof(int?)),
+            new TestCaseData(42, typeof(double)),
+            new TestCaseData(42, typeof(double?)),
+            new TestCaseData(42.0f, typeof(double)),
+            new TestCaseData(new D2(), typeof(D1)),
+            new TestCaseData(new D3(), typeof(D1)),
+        ];
+
+        [TestCaseSource(nameof(FailureCases))]
+        public static void CanNotAssignTo(object? actual, Type type)
+        {
+            Assert.That(actual, Is.Not.AssignableTo(type));
+        }
+
+        private static readonly TestCaseData[] FailureCases =
+        [
+            new TestCaseData(null, typeof(int)),
+            new TestCaseData(42.0, typeof(float)),
+            new TestCaseData(new D1(), typeof(D2)),
+            new TestCaseData(new D1(), typeof(D3)),
+        ];
     }
 }


src/NUnitFramework/framework/Constraints/AssignableFromConstraint.cs
```diff
@@ -25,7 +26,22 @@ public AssignableFromConstraint(Type type) : base(type, "assignable from ")
...
+            {
+                return true;
+            }
+
+            if (actual.GetType().IsAssignableFrom(expectedType))
+            {
+                return true;
+            }
+
+            return Reflect.CanImplicitlyConvertTo(expectedType, actual.GetType());
```
manfred-brands: This reads much better if used as an extension method.
The method also already checks your first condition so that can be removed here.

```suggestion
            return expectedType.CanImplicitlyConvertTo(actual.GetType());
```

src/NUnitFramework/framework/Constraints/AssignableToConstraint.cs
```diff
@@ -25,7 +26,22 @@ public AssignableToConstraint(Type type) : base(type, "assignable to ")
...
+            {
+                return true;
+            }
+
+            if (actual.GetType().IsAssignableTo(expectedType))
+            {
+                return true;
+            }
+
+            return Reflect.CanImplicitlyConvertTo(actual.GetType(), expectedType);
```
manfred-brands: Same here. It also means that you don't need the `TypeExtension.cs` file.
```suggestion
            return actual.GetType().CanImplicitlyConvertTo(expectedType);
```

src/NUnitFramework/tests/Constraints/AssignableToConstraintTests.cs
```diff
@@ -1,39 +1,29 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using NUnit.Framework.Constraints;
+using static NUnit.Framework.Tests.Constraints.AssignableTestScenarios;
 
 namespace NUnit.Framework.Tests.Constraints
 {
     [TestFixture]
-    public class AssignableToConstraintTests : ConstraintTestBase
+    public class AssignableToConstraintTests
```
manfred-brands: Can this one now also be `static` like `AssignableFromConstraint`?

src/NUnitFramework/tests/Constraints/AssignableFromConstraintTests.cs
```diff
@@ -1,39 +1,29 @@
...
+            [SetUp]
+            public void SetUp()
+            {
+                ExpectedDescription = $"assignable from <{typeof(TFrom)}>";
+                StringRepresentation = $"<assignablefrom {typeof(TFrom)}>";
+            }
 
-        private class D2 : D1
-        {
+            private static readonly object[] SuccessData = [new TTo(), new object()];
```
manfred-brands: Should this not include `new TFrom()` to proof that one can assign from an instance of the same type?

src/NUnitFramework/tests/Constraints/AssignableTestScenarios.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Collections.Generic;
+
+namespace NUnit.Framework.Tests.Constraints
+{
+    public static class AssignableTestScenarios
+    {
+        public static IEnumerable<TestFixtureData> GetAssignableTestScenarios()
+        {
+            yield return new TestFixtureData() { TypeArgs = [typeof(int), typeof(long)] };
+            yield return new TestFixtureData() { TypeArgs = [typeof(int), typeof(int)] };
```
manfred-brands: The `AssignableToConstraintTests` test already tests that one can assign `TTo` to `TTo`. 
So this use case with two the same types is not needed if the same is added to `AssignableFromConstraintTests` see my remark there.

manfred-brands (changes requested):
A few issues.

src/NUnitFramework/framework/Constraints/AssignableToConstraint.cs
stevenaw: >  It also means that you don't need the TypeExtension.cs file.

Thanks for noticing this. I had thought I hit a negative case during development where it was needed but perhaps that was just an artifact of figuring this one out. I'll take a look today and push a change accordingly.

src/NUnitFramework/tests/Constraints/AssignableToConstraintTests.cs
stevenaw: Thanks! Yes, another artifact of back-and-forth approaches during development.

For added context or documention too, I will add that I found wrapping the underlying `ConstraintValidation` fixture with a parent fixture like this instead of just having the generic test fixture on it's own made it easier to run all the scenarios at once.

src/NUnitFramework/framework/Internal/TestFixtureParameters.cs
```diff
@@ -53,7 +53,7 @@ public TestFixtureParameters(ITestFixtureData data) : base(data)
         /// <summary>
         /// Type arguments used to create a generic fixture instance
         /// </summary>
-        public Type[]? TypeArgs { get; }
+        public Type[]? TypeArgs { get; init; }
```
stevenaw, line 56: @manfred-brands @OsirisTerje 

One note about this and the `init` property we added to `RetryAttribute`.
Both of these changes will require calling code to target `LangVersion >= 9` in order to set these properties. Is this anything worth calling out in release notes? 

This would exclusively affect netfx test projects that either use their default lang version or explicitly target a version < 9 AND which also do not consume any sort of polyfill

stevenaw:
Thanks for your suggestions @manfred-brands 
I think all have been implemented now

src/NUnitFramework/framework/Internal/TestFixtureParameters.cs
manfred-brands: @stevenaw  It is not only the langversion, but also the special attribute.
I know we reference the 'IsExternalInit' nuget package for this which is source code only.
What happens if:
1. A .net framework package does not reference anything else
2. A .net framework package has its own copy of either IsExternalInit or otherwise.

Let me see if I can test this today.

stevenaw: I tested the latest alpha on an old test project and found that I just needed to change the langversion in that case. I was using the latest Visual Studio 2022. The project in question was not an sdk-formatted csproj (though I suspect it wouldn't matter)

I don't _think_ there was an `IsExternalInit` reference but I could be mistaken or missing something. I'm curious to hear how your investigation goes @manfred-brands

src/NUnitFramework/framework/Constraints/AssignableFromConstraint.cs
```diff
@@ -25,7 +26,17 @@ public AssignableFromConstraint(Type type) : base(type, "assignable from ")
...
+            {
+                return false;
+            }
+
+            if (expectedType == actualType)
+            {
+                return true;
+            }
+
+            return expectedType.CanImplicitlyConvertTo(actual.GetType());
```
manfred-brands: I didn't realize that _actualType_ is a field set in the base class. 
This then can now be:

```suggestion
            return expectedType.CanImplicitlyConvertTo(actualType);
```

src/NUnitFramework/framework/Constraints/AssignableToConstraint.cs
```diff
@@ -25,7 +26,17 @@ public AssignableToConstraint(Type type) : base(type, "assignable to ")
...
+            {
+                return false;
+            }
+
+            if (expectedType == actualType)
+            {
+                return true;
+            }
+
+            return actual.GetType().CanImplicitlyConvertTo(expectedType);
```
manfred-brands:
```suggestion
            return actualType.CanImplicitlyConvertTo(expectedType);
```

src/NUnitFramework/tests/Constraints/AssignableToConstraintTests.cs
```diff
@@ -1,39 +1,29 @@
...
-#pragma warning disable IDE0052 // Remove unread private members
-        private static readonly object[] SuccessData = new object[] { new D1(), new D2() };
-        private static readonly object[] FailureData = new object[]
-        {
-            new TestCaseData(new B(), "<" + typeof(B).FullName + ">")
-        };
-#pragma warning restore IDE0052 // Remove unread private members
-
-        private class B
+        [TestFixtureSource(typeof(AssignableTestScenarios), nameof(GetAssignableTestScenarios))]
```
manfred-brands, line 12: This doesn't allow testing `null` values and I find the double loop hard to see as they are in different files.
I find the following clearer:

```csharp
        [TestCaseSource(nameof(SuccessCases))]
        public static void CanAssignTo(object? actual, Type type)
        {
            Assert.That(actual, Is.AssignableTo(type));
        }

        private static readonly TestCaseData[] SuccessCases =
        [
            new TestCaseData(null, typeof(object)),
            new TestCaseData(null, typeof(int?)),
            new TestCaseData(42, typeof(int?)),
            new TestCaseData(42, typeof(double)),
            new TestCaseData(42, typeof(double?)),
            new TestCaseData(42.0f, typeof(double)),
            new TestCaseData(new D2(), typeof(D1)),
            new TestCaseData(new D3(), typeof(D1)),
        ];

        [TestCaseSource(nameof(FailureCases))]
        public static void CanNotAssignTo(object? actual, Type type)
        {
            Assert.That(actual, Is.Not.AssignableTo(type));
        }

        private static readonly TestCaseData[] FailureCases =
        [
            new TestCaseData(null, typeof(int)),
            new TestCaseData(42.0, typeof(float)),
            new TestCaseData(new D1(), typeof(D2)),
            new TestCaseData(new D1(), typeof(D3)),
        ];
```

Note that the current code allows assigning `42` to `int?` and to `double`, but not to `double?`.
The `Nullable` is only tested if value is `null`.

manfred-brands (approved):
@stevenaw rather than suggestion changes you apply, I made some changes directly.

I did them in separate commits so if you don't agree, you can revert and make your own version.

src/NUnitFramework/framework/Internal/TestFixtureParameters.cs
manfred-brands: I have create a library with a class with an `init` property. Compiled with later SDK, LangVersion 9 and `IsExternalInit`.
Then I made a consuming console application and referenced the library using a dll.

Compilation of the consuming application result an error that I need LangVersion 9.
I tried to go back to a dotnet version that doesn't know about `init`, i.e. 3.1 and compiled it using that.
This compiles fine without having to set a language version and it also runs.
Then I thought that it might see `init` as `set` and it would allow me change the property later, but no it didn't.
So even though `init` is a later language version, it seems to already supported by earlier runtime.

stevenaw:
Thanks @manfred-brands I like your simplifications, thanks for making them.
I'm happy to merge this if you are.

Each of the commits seem fairly self-contained and descriptive so I may avoid squashing unless you had a preference.

PR #5086, opened 2025-12-15
Support explicit TypeArgs specification on TestFixtureSource

### LINKED ISSUE #2958: Support Type Arguments with [TestFixtureSource]
Using [`[TestFixtureSource]`](https://github.com/nunit/nunit/blob/368a65da746c506dd211679c51a9e38d7e7b3597/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs) I'd like to provide type arguments alongside ctor arguments.

Example:

    public class When_entity_name_is_too_short<TEntity> : BaseFixture
        where TEntity : IHaveName
    {
        private TEntity _entity;

        public When_entity_name_is_too_short<TEntity>(TEntity entity, int minLength)
        {
            this._entity = entity;
        
            this._entity.Name = new String('-', minLength - 1);
        }
        
        [Test]
        public void Should_be_invalid()
        {
            this._entity.Should().BeInvalid();
        }
        
        [Test]
        public void Should_have_name_validation_error()
        {
            this._entity.Should().HavePropertyError(nameof(IHaveName.Name));
        }
    }

`[TestFixtureSource]` implements [`IFixtureBuilder2`](https://github.com/nunit/nunit/blob/d09def3521dca9be6fbd84fc483c555c4955afe3/src/NUnitFramework/framework/Interfaces/IFixtureBuilder.cs) which returns an [`IEnumerable<TestSuite>`](https://github.com/nunit/nunit/blob/42eba347d67ab441bc01903cedf1596c029c1e6b/src/NUnitFramework/framework/Internal/Tests/TestSuite.cs).
There's several constructors for `TestSuite` and at least [this one](https://github.com/nunit/nunit/blob/42eba347d67ab441bc01903cedf1596c029c1e6b/src/NUnitFramework/framework/Internal/Tests/TestSuite.cs#L81) takes the type of the fixture and arguments as parameters.
Hence the feature should generally be feasible.

As far as I can gather I can see the following options to add support for type arguments:

- make `[TestFixtureSource]` accept `ITestFixtureData` instead of just `TestFixtureData`
- make `TestFixtureData.TypeArgument` setter public. Makes the type externally mutable, though (why isn't the property currently read-only?)
- introduce a new ctor on `[TestFixtureSource`]: `public ctor(Type[] typeArguments, object[] arguments)`
--> this is a breaking change for all usages of `ctor(params object[] args)` where it was used like `new (Type[], object[])`. Don't think it was used very often like this, though...
--> the breaking change could be circumvented by creating a new type which itself holds on to the type arguments and the other arguments. That would be a bit ugly, though
- support `CustomAttribute : NUnitAttribute, IFixtureBuilder` which could be used instead of `TextFixtureSource` (I'm actually not certain if that's already the case).

I'd be happy to provide a PR including updated documentation ;-)

stevenaw:
Fixes #2958 

This PR uses the same approach for explicit specification of generic types as already exists for TestCaseSource.

This PR also updates a few lines in NUnitTestFixtureBuilder to use `Array.Copy()` instead of a manual iterative approach to rebuilding the arg arrays as the built-in method will be faster on modern runtimes thanks to some internal optimizations.

Documentation will need to be updated after this merges.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
@@ -76,7 +76,7 @@ namespace NUnit.Framework
         public string? Category { get; set; }
 
         /// <summary>
-        /// Type arguments used to create a generic fixture instance
+        /// Type arguments used to create a generic fixture instance.
         /// </summary>
         public Type[]? TypeArgs { get; set; }
 


src/NUnitFramework/framework/Attributes/TestFixtureSourceAttribute.cs
```diff
@@ -75,6 +75,11 @@ public TestFixtureSourceAttribute(Type sourceType)
         /// </summary>
         public string? Category { get; set; }
 
+        /// <summary>
+        /// Type arguments used to create a generic fixture instance
```
stevenaw:
```suggestion
        /// Type arguments used to create a generic fixture instance.
```

OsirisTerje:
@stevenaw Do you plan to have this into 4.5 ? :innocent:

stevenaw:
Only if the approvals and merge happen in time. 🙂
 I left the milestone off the issue for now so as not to hold up the release over this. I figure we can add that after if this makes it in.

stevenaw:
Thanks @manfred-brands 
I'll merge this now. Documentation issue has been created and linked

Issue #5089, opened 2025-12-17
Optional parameters aren't handled when using TestFixtureSource


stevenaw:
Optional argument support was added for TestCaseSource in https://github.com/nunit/nunit/issues/1281
TestFixtureSource should remain consistent and add support for optional arguments too

Issue #3434, opened 2019-11-24
Handling of passing null as element to a parametric testmethod that accepts a params


mikkelbu:
The following issue can up when I tried to add (i.e. extend some tests to solve nunit.analyzers#169)

Given
```csharp
[TestCase(1, null)]
[TestCase(1, "A")]
public void TestHandlingOfNullInParams(params int[] a) {
    Assert.That(a, Is.EquivalentTo(new[] { 1, 0 }));
}
```
The first test passes, but the second test fails with
```
Result Message:	System.InvalidCastException : Object cannot be stored in an array of this type.
Result StackTrace:	
at System.Array.InternalSetValue(Void* target, Object value)
   at System.Array.SetValue(Object value, Int32 index)
   at NUnit.Framework.TestCaseAttribute.GetParametersForTestCase(IMethodInfo method) in D:\a\1\s\src\NUnitFramework\framework\Attributes\TestCaseAttribute.cs:line 263
```
I would have expected them both to fail, as null cannot be put in an int array, but due to the note mentioned on https://docs.microsoft.com/en-us/dotnet/api/system.array.setvalue?view=netframework-4.8#System_Array_SetValue_System_Object_System_Int32_ the null is "converted" into a zero.

> If SetValue is used to assign `null` to an element of an array of value types, all fields of the element are initialized to zero. The value of the element is not a null reference, and cannot be found by searching for a null reference.

And we use `SetValue` twice in 
https://github.com/nunit/nunit/blob/46f8b2ae280a6a6150e0f0bf86718a783d8b0c9d/src/NUnitFramework/framework/Attributes/TestCaseAttribute.cs#L281-L307

So perhaps we should add special handling of placing null into value-type arrays?

CharliePoole:
Perhaps. Or perhaps just detect that you are not passing an int[] and give an error?

Phyrik:
Bit of an old issue but I'd like to work on it. Any consensus on which route to go with fixing this? I think a special case for setting null into value type arrays would be a good way to go

stevenaw:
Thanks for asking @Phyrik ! This one would likely be a great candidate for work once we decide on an approach.

[...]

Issue #2268, opened 2017-06-19
Standardise behaviour of params and optional arguments across parameterised attributes


ChrisMaddock:
NUnit has various parameterised arguments, which make various special conversions to adjust parameters to match method signatures, to allow invocation via reflection. The key ones I'm thinking of are `[TestCase()]`, `[TestCaseSource()]`, `[TestFixture()]` and `[TestFixtureSource()]`.

Support for params and optional arguments between these attributes is inconsistent. Over the years, this has raised various issues:
https://github.com/nunit/nunit/issues/1281 (open)
https://github.com/nunit/nunit/issues/1883 (open)
https://github.com/nunit/nunit/issues/1459 (open)
https://github.com/nunit/nunit/issues/1640
https://github.com/nunit/nunit/issues/1278
https://github.com/nunit/nunit/issues/1282
https://github.com/nunit/nunit/issues/257

We've talked in a couple of places ([link 1](https://github.com/nunit/nunit/pull/2104), [link 2](https://github.com/nunit/nunit/issues/1281), [link 3](https://github.com/nunit/nunit/pull/1850)) about harmonising this functionality, and having a single utility class, which resolves parameters passed to attributes, and ensures they are handled in a standard manner across all attributes. As I remember - this was something we agreed to go ahead with - @Yarmonov started on a PR in https://github.com/nunit/nunit/pull/2104. (This seems to have gone stale, unfortunately. Will check in on it shortly.) This would mean that the params keyword and optional arguments would be handled in a standard fashion across all of these attributes, meaning no unexpected surprises for the end-user. 

I wanted to create a master issue for this whole idea - as some of the existing open issues are tagged easyfix - and I didn't want anyone to end up implementing any more solutions individually, when we actually want to be sharing the existing functionality between attributes. If @nunit/framework-team agree - I'd like to close the open issues above, and use this as a single issue to reflect the overall plan.

CharliePoole:
Seems like this should be an Epic.

ChrisMaddock:
I would have thought an Epic implied we wanted people to take each issue on separately, where as my intention was that the functionality would be synchronized as one PR - rather than someone creating a new implementation for each attribute.

Yarmonov:
I have done such utility class that covers all attributes you've mentioned, all cases (optional or/and array params), with a lot of performance optimizations, unit tests and some slight design ideas. Halfway migrated the existing code to this new utility before got extremely occupied. My apologizes.

[...]

PR #5090, opened 2025-12-17
Add support for optional args to TestFixtureSource

stevenaw:
Fixes https://github.com/nunit/nunit/issues/5089
Fixes https://github.com/nunit/nunit/issues/3434

Relates to https://github.com/nunit/nunit/issues/2268
Relates to https://github.com/nunit/nunit/issues/1281

### DIFF
--- a/src/NUnitFramework/framework/Internal/TestCaseParameters.cs
+++ b/src/NUnitFramework/framework/Internal/TestCaseParameters.cs
@@ -122,7 +122,7 @@ namespace NUnit.Framework.Internal
             }
 
             // Special handling when sole argument is an object[]
-            if (argsNeeded == 1 && method.GetParameters()[0].ParameterType == typeof(object[]))
+            if (argsNeeded == 1 && parameters[0].ParameterType == typeof(object[]))
             {
                 if (argsProvided > 1 ||
                     argsProvided == 1 && Arguments[0]?.GetType() != typeof(object[]))


src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -90,7 +90,25 @@ public static object Construct(Type type, object?[]? arguments)
...
+                {
+                    // Add the optional args at the call-side where we haven't passed enough explicitly
+                    var endIdx = hasParamsArray ? parameterInfos.Length - 1 : parameterInfos.Length;
+
+                    var newArgs = new object?[endIdx];
+                    Array.Copy(arguments, newArgs, arguments.Length);
+
+                    for (var i = arguments.Length; i < endIdx && parameterInfos[i].IsOptional; i++)
+                    {
+                        newArgs[i] = parameterInfos[i].DefaultValue;
```
manfred-brands: The code in `TestCaseParameters.AdjustArgumentsForMethod` uses the special `Type.Missing`.

stevenaw:
Thanks @manfred-brands I like that idea. 
I'll be starting traveling tomorrow evening but I hope to be able to make that change before I do

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -89,34 +90,69 @@ public static object Construct(Type type, object?[]? arguments)
...
                     {
-                        // Don't convert arguments as there was already an array we could use.
+                        newArgs[i] = Type.Missing;
                     }
                     else
                     {
-                        Type? elementType = parameterInfo.ParameterType.GetElementType();
-                        if (elementType is null)
-                        {
-                            throw new InvalidTestFixtureException(type.FullName + " params argument did not have an element type");
```
stevenaw, line 105: Note: This was removed as part of the change. This check and exception type is specific to test fixtures, however on further investigation I have not been able to trigger it. It seems to need something like the below where the params keyword is used with a non-array type.
```
public static void (params int x);
```

This is invalid syntax in both C# and F#. The comment on [the commit](https://github.com/nunit/nunit/commit/e874a3ecb5dcfa7e321d0594aef5b36ebecd27cf#diff-b8b14777b50239a004ff75fa1b88d2736133b7f3769ec7be1aa963652fb379ea) which added it seems to indicate that it was for also passing an explicit array in as the params arg, however I was not able to get the tests to fail when removing the check, nor could I figure out how to add a new test to make it fail.

The simplest solution seemed to be to remove it.

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -89,34 +90,69 @@ public static object Construct(Type type, object?[]? arguments)
...
-                        }
-
-                        int paramsOffset = parameterInfos.Length - 1;
-                        var paramArray = Array.CreateInstance(elementType, argTypes.Length - parameterInfos.Length + 1);
-                        for (int i = 0; i < paramArray.Length; i++)
```
stevenaw, line 110: Note: This code was also modified in the final version to now use `Array.Copy()`

src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs
```diff
@@ -463,6 +463,13 @@ public void HandlesCancellationTokenAsLastArgument(int _, CancellationToken canc
             Assert.That(cancellationToken, Is.EqualTo(TestContext.CurrentContext.CancellationToken));
         }
 
+        [TestCaseSource(nameof(OneArg)), CancelAfter(5000)]
+        public void HandlesCancellationTokenWithDefaultArgumentAsLastArgument(int _, CancellationToken cancellationToken = default)
```
stevenaw:
```suggestion
        public void HandlesCancellationTokenWithDefaultValueAsLastArgument(int _, CancellationToken cancellationToken = default)
```

src/NUnitFramework/framework/Internal/Reflect.cs
stevenaw: Update: I've actually had to re-add this item-by-item construction. The previous code would allow for implicit conversion of values into the params array while `Array.Copy()` did not. `Array.Copy()` caused the following previously working and valid scenario to fail:

```csharp
[TestCase(1, 2)]
public static void HandlesImplicitlyConvertibleParamsValues(params long[] a)
{
    Assert.That(a, Is.EqualTo([1L, 2L]));
}
```

src/NUnitFramework/framework/Internal/Reflect.cs
```diff
@@ -89,34 +90,82 @@ public static object Construct(Type type, object?[]? arguments)
...
+                    {
+                        var arg = arguments[i + paramsOffset];
+                        var argType = arg?.GetType();
+
+                        //Only assign if we can convert the value to the element type
+                        if (!argType.CanImplicitlyConvertTo(elementType))
                         {
-                            paramArray.SetValue(arguments[i + paramsOffset], i);
+                            var sourceType = argType is null ? "null" : argType.FullName;
+                            throw new InvalidCastException($"Cannot convert {sourceType} to '{elementType}'");
```
stevenaw, line 158: This was explicitly thrown to maximize behavioral compatibility as this is what is thrown today by the runtime for these cases.

src/NUnitFramework/framework/Internal/Extensions/ArgumentsExtensions.cs
```diff
@@ -32,6 +41,14 @@ public static bool ParameterIsParamsArray(this IParameterInfo lastParameter)
...
         }
 
+        /// <summary>
+        /// Checks if the parameter is a 'params' array.
+        /// </summary>
+        public static bool ParameterIsParamsArray(this ParameterInfo lastParameter)
+        {
+            return lastParameter.ParameterType.IsArray && lastParameter.HasAttribute<ParamArrayAttribute>(false);
+        }
+
```
manfred-brands, line 51: This method was intended to be called in the two places currently having this spelled out.

manfred-brands (approved):
Good changes look good, including the additional tests.

I pushed up 2 small commits
* Use the new extension method where possible.
* Remove an unnecessary fetch of parameters as they are now in a local variable.

stevenaw:
Thanks @manfred-brands 
Good catch about the extension method

PR #5091, opened 2025-12-20
Simplify validation logic in ApartmentAttribute constructor

aritrikg:
This change simplifies the ApartmentAttribute constructor by replacing
Guard.ArgumentValid with a direct argument check and standard exception.

There is no functional behavior change. The intent is improved readability
and reduced indirection.

### DIFF
--- a/src/NUnitFramework/framework/Attributes/ApartmentAttribute.cs
+++ b/src/NUnitFramework/framework/Attributes/ApartmentAttribute.cs
@@ -1,26 +1,24 @@
-// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
-
 using System;
 using System.Threading;
 using NUnit.Framework.Internal;
 
 namespace NUnit.Framework
 {
-    /// <summary>
-    /// Marks a test as needing to be run in a particular threading apartment state. This will cause it
-    /// to run in a separate thread if necessary.
-    /// </summary>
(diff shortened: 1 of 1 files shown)


manfred-brands:
@aritrikg Thanks for your incentive.

Why did you delete all XML documentation?

There is a reason to have indirect method: 

Smaller code as the condition and the throwing is not repeated in every use case.

Microsoft has their own equivalent such as [ArgumentOutOfRangeException.ThrowIfEqual](https://learn.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception.throwifequal?view=net-10.0#system-argumentoutofrangeexception-throwifequal-1(-0-0-system-string))

These `ThrowIf` methods were only added in .net8
We could move over to them if we make C#14 extension methods to support code pre-net8

stevenaw:
@aritrikg  Thanks for your interest in contributing to Nunit! 

As @manfred-brands mentioned, there's a reason each of those particular pieces are in place. We have labeled a few predefined issues in our repo which are ideal for first time contributors if you're interested in taking a look :)

https://github.com/nunit/nunit/issues?q=state%3Aopen%20label%3A%22good%20first%20issue%22

OsirisTerje:
As @manfred-brands  and @stevenaw said:  The Guard pattern is a well known and much used pattern, it simplifies technical code.  Microsoft community have their own package for Guards too, see https://learn.microsoft.com/en-us/dotnet/communitytoolkit/diagnostics/guard

Issue #5088, opened 2025-12-16
Add a strongly-typed API for passing settings into the framework


stevenaw:
The framework currently expected settings to be passed in as a `Dictionary<string, object>`. The framework internally knows and expects certain types for each setting name. The API itself doesn't expose this and requires callers to know setting names and their expected formats or underlying types for settings to be received and processed. A clearer, strongly-typed API could improve the discoverability of which settings are supported by the framework and clearer communicate their types and intent.

This could be done in a non-breaking fashion through an overload to be opted-into slowly over time, or as part of a breaking change requiring consumers an explicit migration step at the time they upgrade to a new major version.

stevenaw:
I've put the `design` label on this as I think it needs some thought, discussion, and decision on how to best approach this

OsirisTerje:
Could this be a new API, so that we in effect do a versioning of the settings Api ?
That way we keep the old Api so that we don't break people, but let the old Api call the new Api after conversion. 
And mark it as deprecated.

stevenaw:
Agreed. I'd been trying to leave implementation details out of the initial proposal but that makes the most sense to me as well.
I don't know how often framework api consumers (other library or framework authors) update compared to application testers, but in the event that it is slower then I would suggest that there probably isn't an urgency to move quickly on the deprecation front.

Leaving at least 1-2 years AND at least one major version between deprecation and removal would hopefully allow for enough notice for any consuming library authors to transition with minimal friction. Improving error messages when invalid or incompatible setting are passed into the old API and internally adapted to the new API could be another way to guide transition in the interim.

stevenaw:
Two possible transition paths could be:

**Version 4 of the framework**:
- New API is introduced
- Old API is deprecated via the `[Obsolete]` attribute and an adapter class is used internally to allow framework internals to use the new API
- Any added input flexibility from https://github.com/nunit/nunit/issues/5087 stays in place
  - Optionally we could choose to introduce a warning if the input value requires type conversion

**Version 5 of the framework (option 1)**
- We drop the old API immediately if enough time has passed since it had been marked it as `[Obsolete]`
 
**Version 5 of the framework (option 2)**
- New API and old API surface remains unchanged
- Internally we remove conversions when adapting inputs to the new API, and instead issue an error at _runtime_ if an input value requires type conversion
- A follow-up version 6 could then drop the old API at which point type-unsafe values would cause errors at _compile time_.

PR #5101, opened 2026-01-06
Type coerce settings to expected types

### LINKED ISSUE #5087: InvalidCastException for arguments when running nunit console with .NET 8 or 10
I am getting an exceptions when running the NUnit console for a `net8.0-windows` or `net10.0-windows` compiled test assembly. I've considered the agents config for 8.0/10.0 being changed to `Microsoft.WindowsDesktop.App`

When debugging I observed following exception:

`DirectTestRunner.cs:
line 196:
System.InvalidCastException: Unable to cast object of type 'System.String' to type 'System.Int32'.
   at NUnit.Framework.Api.NUnitTestAssemblyRunner.GetLevelOfParallelism(ITest loadedTest)
   at NUnit.Framework.Api.NUnitTestAssemblyRunner.CreateTestExecutionContext(ITest loadedTest, ITestListener listener)
   at NUnit.Framework.Api.NUnitTestAssemblyRunner.RunAsync(ITestListener listener, ITestFilter filter)
   at NUnit.Framework.Api.NUnitTestAssemblyRunner.Run(ITestListener listener, ITestFilter filter)
   at NUnit.Framework.Api.FrameworkController.RunTests(Action1 callback, String filter)
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
   at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span1 copyOfArgs, BindingFlags invokeAttr)`

**The full command line being used**

`nunit3-console.exe DotNet8_10Test.dll --dispose-runners --noh --labels=BeforeAndAfter --trace=Verbose --debug-agent --agents=1 --workers=1 --timeout=120000 --debug-agent`

It seems the problematic arguments, which cause the exception are:

1. workers
2. timeout

**Without these arguments, nunit console runs without any failure.**

**Method of installation (e.g. NuGet, zip, msi)**
built/debugged source or chocolatey lib, result is the same

**Log files (Run the console with --trace=Debug to produce these)**

`Runtime Environment
   OS Version: Microsoft Windows NT 10.0.22631.0
   Runtime: .NET Framework CLR v4.0.30319.42000

Test Files
    DotNet8_10Test.dll

Unhandled Exception: NUnit.Engine.NUnitEngineException: Remote test agent exited with non-zero exit code -532462766
   at NUnit.Engine.Services.TestAgency.OnAgentExit(Process process, Guid agentId)
   at NUnit.Engine.Services.TestAgency.<>c__DisplayClass13_0.<GetAgent>b__0(Object sender, EventArgs e)
   at System.Diagnostics.Process.OnExited()
   at System.Diagnostics.Process.RaiseOnExited()
   at System.Threading._ThreadPoolWaitOrTimerCallback.PerformWaitOrTimerCallback(Object state, Boolean timedOut)

Errors, Failures and Warnings

1) Error : C:\00_git\eee4\binTests\NUnit\DotNet8_10Test.dll
System.Net.Sockets.SocketException : An existing connection was forcibly closed by the remote host
--SocketException
An existing connection was forcibly closed by the remote host
   at NUnit.Engine.Communication.Transports.Tcp.SocketReader.GetNextMessage()
   at NUnit.Engine.Communication.Transports.Tcp.TestAgentTcpProxy.TestRunResult(ITestEventListener listener)
   at NUnit.Engine.Runners.ProcessRunner.RunTests(ITestEventListener listener, TestFilter filter)

Test Run Summary
  Overall result: Failed
  Test Count: 0, Passed: 0, Failed: 0, Warnings: 0, Inconclusive: 0, Skipped: 0
  Start time: 2025-12-15 09:46:50Z
    End time: 2025-12-15 09:46:52Z
    Duration: 1.972 seconds`

**Version of the NUnit Engine/Console**
`3.21.0 (latest)`

**All versions of the NUnit Framework in use**
`4.4.0 (latest)`

**Whether the behaviour is any different when the tests are run using the --inprocess flag**
did not investigate

stevenaw:
Fixes #5087

Converts settings to expected types. An exception will be thrown if a setting with an incompatible format is provided in order to alert to incorrect values. This is to handle the case on NET 8+ using version 3.22 of the console runner where numeric or boolean settings passed as command line arguments are "passed through" by the engine to the framework as strings.

This is intended as a shorter-term solution until the proper APIs can be added to the framework to communicate the expected types for the framework's settings to runners.

### DIFF
--- a/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
+++ b/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
@@ -218,7 +218,15 @@ namespace NUnit.Framework.Api
             // Needs to be set for StopRun
             Context = context;
 
-            WrapInNUnitCallContext(() => StartRun(context, TopLevelWorkItem, listener));
+            try
+            {
+                WrapInNUnitCallContext(() => StartRun(context, TopLevelWorkItem, listener));
+            }
+            catch (Exception)
+            {
+                OnRunCompleted(this, EventArgs.Empty);
+                throw;
+            }
         }
 
         private static void OnUnhandledException(object sender, UnhandledExceptionEventArgs args)


src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
```diff
@@ -102,15 +103,15 @@ private TestSuite Build(Assembly assembly, string assemblyNameOrPath, IDictionar
 
             try
             {
-                if (options.TryGetValue(FrameworkPackageSettings.DefaultTestNamePattern, out object? defaultTestNamePattern))
-                    TestNameGenerator.DefaultTestNamePattern = (string)defaultTestNamePattern;
+                if (options.TryGetValue<string>(FrameworkPackageSettings.DefaultTestNamePattern, out var defaultTestNamePattern))
```
manfred-brands: If you declare the variable as `string` instead of `var` you don't need to specify the generic type as in `<string>` 

One problem with this approach is that it is up to caller to specify the type whilst in fact the type should be associated with the name.
This now allows one place to expect a string whilst another expects an integer.

Have a look at the solution in nunit-console:
https://github.com/nunit/nunit-console/blob/main/src/NUnitEngine/nunit.engine.api/SettingDefinitions.cs
https://github.com/nunit/nunit-console/blob/main/src/NUnitEngine/nunit.engine.api/PackageSettings.cs

Ideally this should be in a library used by both, but that would require creating a nunit-common repository.

src/NUnitFramework/framework/Internal/Extensions/IDictionaryExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            // Handle nulls
+            if (raw is null)
+            {
+                // If T is reference type or nullable, we can return default (null) and store null
+                var targetType = typeof(T);
+                if (!targetType.IsValueType || Nullable.GetUnderlyingType(targetType) is not null)
+                {
+                    settings[key] = null!;
```
manfred-brands: We only get here if there is a `null` value already stored, so no point storing it again.

src/NUnitFramework/framework/Api/DefaultTestAssemblyBuilder.cs
OsirisTerje: Do we need to keep settings in an object dictionary?  
Is this something that is a leftover from way back old days?
Why can't we just define a settings class with the correct property types and be done with it? 
Instances of that class can then be serialized (json or something even more compact) if we need to transport them over something. 

I have seen the code in the engine, and to me it is too much complexity for something that is very simple.

stevenaw: Overhauling this to define a settings class was what I'd had in mind for https://github.com/nunit/nunit/issues/5088
For now we are somewhat limited in how we can fix this on the framework side to allow for the latest versions of the console runner to run .NET 8+ tests as there's some parts of the public API which have the dictionary approach baked into their contract: https://github.com/nunit/nunit/blob/main/src/NUnitFramework/framework/Api/ITestAssemblyRunner.cs#L56

My preference here in order to allow the console runner to run .NET 8+ again in the short term would be a small focused fix without any external behavioural or contract changes and to reserve larger more robust changes for #5088 where we can make some more longer-term decisions about how we want this to look. Right now it sounds like the console runner effectively can't run .NET 8 or 10 tests at all, which seemed to me like the more urgent issue to fix as unobtrusively as possible.

OsirisTerje: The latest 3.22 engine version do run net8 tests, at least the engine does, as that is what the adapter use, and it also runs net 10 tests.  So I would assume that the console also do, or ?  @CharliePoole

CharliePoole: @OsirisTerje Yes, the console also runs net 10 tests, using a net 10 agent.

CharliePoole: @stevenaw I assume you mean that a full fix would have to wait for a 5.x framework. However, you could add code like that in the engine for use in interpreting the objects you receive correctly. See code for `SettingDefinition` / `SettingDefinitions` / `PackageSetting` / PackageSettings` in the engine.

Although I agree with @manfred-brands that a shared package would be a nicer solution, that will require changes on both sides, since the shared part should not include the specific setting names, which vary between the engine and the framework. So I'd start with a copy and modify approach anyway, since the required modifications would be a guide for how to repackage it.

You could also consider __adding an additional API_, a non-breaking change for those who don't use it.

stevenaw: Thanks, that's great to hear that the console runner can run NET10 then. I am wondering then if there's anything left to do for #5087 . Perhaps it and this PR can be closed and we can continue this design discussion in #5088 itself.

stevenaw: Actually, something I could try to do is an _internal_ settings class. That way we wouldn't lock ourselves in to anything before we can iron out a long term design for a type-safe API here.

That would probably be beneficial even if there were no issues running these tests or passing these settings on NET10 today

stevenaw:
I've decided to go back and simplify this further. The first commit introduced a "convert and cache" mechanism which would replace the string-formatted value in the dictionary with the underlying type. In practice this looks to have been overly complicated as we primarily only read each settings once per test run.

I've added a second commit to just convert the extracted values to their intended types _after_ retrieving it from the settings dictionary. I've added this as a second commit to make it easy to adjust as this PR evolves, though it will mean we'd likely prefer to squash merge when the time comes.

Three remaining settings will still require investigation as their expected underlying types are not primitives and will not easily convert from a string if that is how they are received.
- TestParametersDictionary (type-checked to a `Dictionary<string, string>`)
- LOAD (type-checked to a `IList`, nunitlite passes this in as a `List<string>`)
- InternalTraceWriter (type-checked to a `TextWriter`)

CharliePoole:
@stevenaw Each of the three setting you mention has special handling in the engine... but in some cases only in the V4 build.

* In V4 `TestParametersDictionary` is serialized between the engine and any agent as XML in a very simple format. Code in the agent then converts the XML to a Dictionary<string, string> before passing it to the framework. In V3.22 I believe this is now broken when running through an agent although not when in-process. It would be a little trickier to port this change from V4 to V3 because of the need to support both in- and out-of-process scenarios but not impossible.
* The situation with `LOAD` is similar. A string containing a semicolon-separated list of Type fullnames is passed in V4. I'm not sure if it works in V3. Actually, I was uncertain whether the framework still used it so didn't spend much  time on it.
* `InternalTraceWriter` is still unsupported in both engine versions. I have a TODO item to find out if anyone wants this to be supported.

I'm wondering... there are a number of issues similar to this and I'm not 100% sure which of them you are trying to solve. However, if the problem is something like "the engine sends us the wrong Types" perhaps the fix should be in the engine. I say this even though, as @OsirisTerje knows, I'm trying to reduce my role in the engine, especially V3. However, someone else may be able to make the necessary engine changes so that the existing framework code works while we concentrate on getting a better overall interface for V4 engine + V5 framework. Just a thought.

stevenaw:
Thanks @CharliePoole That is good news if those three settings are explicitly handled by the engine.
In this PR I'm primarily trying to solve the problem you've described here (emphasis my own)

> The problem arises when tests are run by a separate agent process. Then the test package and its settings have to be serialized and deserialized before it gets to the driver that creates the dictionary containing the settings. In the past, this worked because BinaryFormatter was used for serialization and deserialization and maintained type correctness across the boundary. The XML formatting we now use delivers everything as a string, of course.
> 
> In the V3 engine, I've dealt with that by simply making sure that the type is correct at every site where a setting is used. But _**since framework settings are not actually known to the engine they are simply passed through**_.

In other words, and if my understanding is right, I'm seeking to fix the V3 interaction on the framework side by allowing stringified versions of these settings to continue to be passed through and then converted by the framework into their intended typed.

@OsirisTerje 

I'm also in favour of an additional upstream type-check in the engine for those who can't or won't have updated to the latest version of the framework. I would also be fine with if we preferred to _only_ have the fix there and to close this PR without merging it.

CharliePoole:
@stevenaw That makes sense. Here's a bit more info for you regarding engine / framework interaction...

We have documented in some comments (I believe in both projects) that settings are supposed to be of Type string, int and bool. However, over time we added a few things that use other Types even though those comments still remain. What I have done in the V4 engine serialization to deal with this is on several levels...

1. If the name of a setting is known, then the correct Type is known, we will try to treat it as that type. If it actually came to the engine as a string we try to parse it or, for special cases, we decode the XML format used internally in the engine. This is why the V4 engine needs to know about some framework settings and their types, e.g. TestParametersDictionary.
2. If the name is not known it must be a string, int or bool. This gives a high probablility of success with unknown parameters but can fail in some situations, such as some sort of id like "12345", interpreted as an int but possibly intended as a string. Ideally, the engine should probably import all the setting definitions from the framework once you have something that specifies the expected Type.

In the longer term, I think it would be advantageous if all settings were defined as a string and the framework knew how to handle those strings, parsing or decoding them as required.

CharliePoole:
Whoops! I accidentlaly marked this ready for review and have now made it a draft, which I hope it was before my click. It's slightly scary when folks have the ability in GitHub to do things they are not supposed to do!!!

stevenaw:
Seems you were just a few days too quick for me @CharliePoole  😃

I've now been able to test this using the latest version of the console runner and am ready to switch this out of draft. My repro code has been uploaded to the Nunit issues repo.

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
manfred-brands: You missed this one.

stevenaw: Thanks for catching, that must've gotten missed when I adjusted the implementation. Interesting that the tests didn't catch it. 

I'll take a closer look this weekend and squash rebase the next push to remove the initial implementation from the history.

CharliePoole: @stevenaw The tests in `build.cake` appear to use nunitlite. Do you run tests through the console elsewhere?

stevenaw: Thanks @CharliePoole 
My typical flow is to run our automated within Visual Studio during development and then a full build and test from command-line prior to pushing changes. In this case I also ran some manual tests using the console runner.

The test I'd been surprised to see pass was a unit test added within this PR. It looks like it targeted one overload of `NUnitTestAssemblyRunner.Load()` while the setting I was changing was read from two separate overloads independently.

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -112,9 +112,9 @@ public ITest Load(string assemblyNameOrPath, IDictionary<string, object> setting
         {
             Settings = settings;
 
-            if (settings.TryGetValue(FrameworkPackageSettings.RandomSeed, out var randomSeedValue))
+            if (settings.TryGetValue(FrameworkPackageSettings.RandomSeed, out var randomSeed))
             {
-                Randomizer.InitialSeed = (int)randomSeedValue;
+                Randomizer.InitialSeed = Convert.ToInt32(randomSeed);
```
CharliePoole: General comment for all these conversions using `Convert.ToInt32`, `ConvertToBoolean`, etc.

We have historically allowed users to create their own settings and have, in fact, not prevented them from creating settings with the same names as our defined settings such as `RandomSeed`. This change will cause an error if such a setting cannot be converted to the correct type. IMO that's not a problem - it's fixing a bug rather than making a breaking change.

However, I wonder if it's not overly permissive to accept _any_ Type that is convertible to int. `Convert.ToInt32` has a pretty long list of overloads. Additionaly, for a reference type, it will handle `null`, returning a value of 0. The case you are trying to solve is exclusively limited to string values, so you could consider a more complex algorithm encapsulated in a method, that does something like...

```
// pseudocode
if the type is as expected
   just return the value
else if it's a non-null string
   use Convert.ToInt32 (or other method for some other target type)
```

As usual, just a comment rather than a review. :-)

OsirisTerje:
@stevenaw I have not dived this discussion - but is this ready for merging ?

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
stevenaw: Thanks @CharliePoole I appreciate your insights here and I share your concern about being overly-permissive. I'll push another commit shortly once I finish up with some testing.

stevenaw:
@OsirisTerje This is very close. I'm just in the process of finishing and testing a change which addresses Charlie's comment, in order to avoid us releasing something more permissive than required here.

stevenaw:
This is ready on my end.
@manfred-brands you had requested changes in relation to a missed conversion. I think that's now been addressed.

CharliePoole:
@nunit/framework-team Slightly out-of-band question... 

I only looked at enough code to understand this change but I saw enough that I suspect you don't pass any user defined settings to the TestContext for access by the user. Is that correct?

Why I care... I'm doing a bit of work in the engine to allow undefined settings to be passed through to the framework, while blocking any incorrect use of defined settings. I could simplify a bit if you don't want any settings except those defined in the `FrameworkSettings.cs` file. Alternatively, I could keep passing them for possible future use.

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -272,6 +272,38 @@ public void StopRun(bool force)
...
+        /// <param name="value">The setting value to convert.</param>
+        /// <returns>The requested setting value.</returns>
+        /// <exception cref="ArgumentException">Thrown when parsing is not supported for the requested type.</exception>
+        /// <exception cref="FormatException">Thrown when parsing failed to convert the input value to the requested type.</exception>
+        private static T ConvertSetting<T>(object value)
+        {
+            if (value is T variable)
+                return variable;
+
+            if (value is string && typeof(T) != typeof(string))
```
manfred-brands: There is no need for the 2nd part. If both _value_ and `T` are `string` it would be caught by the test above.

```suggestion
            if (value is string)
```

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -339,15 +371,15 @@ private TestExecutionContext CreateTestExecutionContext(ITest loadedTest, ITestL
 
             // Apply package settings to the context
             if (Settings.TryGetValue(FrameworkPackageSettings.DefaultTimeout, out var timeout))
-                context.TestCaseTimeout = (int)timeout;
+                context.TestCaseTimeout = ConvertSetting<int>(timeout);
             if (Settings.TryGetValue(FrameworkPackageSettings.DefaultCulture, out var culture))
                 context.CurrentCulture = new CultureInfo((string)culture, false);
```
manfred-brands: For consistency should we call `ConvertSetting<string>` here?
```suggestion
                context.CurrentCulture = new CultureInfo(ConvertSettng<string>(culture), false);
```

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
```diff
@@ -272,6 +272,38 @@ public void StopRun(bool force)
...
+        /// <exception cref="FormatException">Thrown when parsing failed to convert the input value to the requested type.</exception>
+        private static T ConvertSetting<T>(object value)
+        {
+            if (value is T variable)
+                return variable;
+
+            if (value is string && typeof(T) != typeof(string))
+            {
+                // Try to find a Parse method, such as may be from the IParsable interface
+                var parseMethod = typeof(T).GetMethod("Parse", [typeof(string), typeof(IFormatProvider)]);
```
manfred-brands: This doesn't work for `bool`.

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
```diff
@@ -82,6 +79,90 @@ public void Load_GoodFile_ReturnsRunnableSuite()
...
+                { FrameworkPackageSettings.SynchronousEvents, "true" },
+                { FrameworkPackageSettings.RandomSeed, "12345" },
+                { FrameworkPackageSettings.RunOnMainThread, "true" },
+                { FrameworkPackageSettings.DefaultCulture, "en-US" },
+                { FrameworkPackageSettings.DefaultUICulture, "en-US" }
+            };
+
+            ITest[] suites = [
+                _runner.Load(GetMockAssembly(), settings),
+                LoadMockAssembly(settings)
```
manfred-brands: Some settings are only retrieved during the `Run` phase, not the load phase.
This misses the fact that you can no longer convert `bool` values.

I made the `ConvertSettings` method internal and added the explicit tests to check conversions:

```csharp
        [Test]
        public void CanConvertBool()
        {
            Assert.That(NUnitTestAssemblyRunner.ConvertSetting<bool>("true"), Is.True);
            Assert.That(NUnitTestAssemblyRunner.ConvertSetting<bool>("False"), Is.False);
            Assert.That(NUnitTestAssemblyRunner.ConvertSetting<bool>(true), Is.True);
            Assert.That(NUnitTestAssemblyRunner.ConvertSetting<bool>(false), Is.False);
        }

        [Test]
        public void CanConvertInt()
        {
            Assert.That(NUnitTestAssemblyRunner.ConvertSetting<int>("42"), Is.EqualTo(42));
            Assert.That(NUnitTestAssemblyRunner.ConvertSetting<int>(42), Is.EqualTo(42));
        }

        [Test]
        public void PassesThroughString()
        {
            Assert.That(NUnitTestAssemblyRunner.ConvertSetting<string>("NUnit"), Is.EqualTo("NUnit"));
        }
```

manfred-brands (changes requested):
The `ConvertSettings` method needs work and the tests don't cover all variables.

src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs
stevenaw: Thanks for noticing this. Perhaps I had been "too clever" by trying to use a standard string "Parse()" function given we knew it would be a string here. This can be converted back to a typeconverter approach easily enough

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
stevenaw: Thanks! I can add these.
I'll also try to get some happy path tests to go through the full Load + Run cycle to capture behavioural contracts of these API functions as runners can expect them to behave.
Unhappy path, where validation fails, is a bit trickier which is where these will be handy

stevenaw:
@manfred-brands Thanks for your repeated reviews on this one earlier. I think this should be ready on my end and ready for another review when time allows for you.

src/NUnitFramework/tests/Api/TestAssemblyRunnerTests.cs
```diff
@@ -82,6 +94,74 @@ public void Load_GoodFile_ReturnsRunnableSuite()
...
+                { FrameworkPackageSettings.StopOnError, "x" },
+                { FrameworkPackageSettings.ThrowOnEachFailureUnderDebugger, "x" },
+                { FrameworkPackageSettings.SynchronousEvents, "x" },
+                { FrameworkPackageSettings.RandomSeed, "x" },
+                { FrameworkPackageSettings.RunOnMainThread, "x" },
+                { FrameworkPackageSettings.DefaultCulture, "x" },
+                { FrameworkPackageSettings.DefaultUICulture, "x" }
+            };
+
+            Assert.That(() => LoadMockAssembly(settings), Throws.Exception.TypeOf<FormatException>());
```
manfred-brands: This doesn't indicate that all those settings are actually checked.

I changed the test into:
```csharp
        [TestCase(FrameworkPackageSettings.NumberOfTestWorkers)]
        [TestCase(FrameworkPackageSettings.DefaultTimeout)]
        [TestCase(FrameworkPackageSettings.DebugTests)]
        [TestCase(FrameworkPackageSettings.PauseBeforeRun)]
        [TestCase(FrameworkPackageSettings.InternalTraceLevel)]
        [TestCase(FrameworkPackageSettings.StopOnError)]
        [TestCase(FrameworkPackageSettings.ThrowOnEachFailureUnderDebugger)]
        [TestCase(FrameworkPackageSettings.SynchronousEvents)]
        [TestCase(FrameworkPackageSettings.RandomSeed)]
        [TestCase(FrameworkPackageSettings.RunOnMainThread)]
        [TestCase(FrameworkPackageSettings.DefaultCulture)]
        [TestCase(FrameworkPackageSettings.DefaultUICulture)]
        public void Load_UnparseableSettings_ThrowsException(string parameter)
        {
            var settings = new Dictionary<string, object>
            {
                [parameter] = "x",
            };

            Assert.That(() => LoadMockAssembly(settings), Throws.Exception.TypeOf<FormatException>());
        }
```

And all but one test (RandomSeed) fail, meaning those values are not obtained during the `Load`.

Most settings are loaded indirectly from `RunAsync`.
If I add a call to `Run`, 8 of them pass.

Some fail  with a `CultureNotFoundException` because `x` is not a valid culture.

The `InternalTraceLevel` is only used in the `FrameworkController` which is not used here.

manfred-brands:
@stevenaw I had to cancel the build. `nunitlite` is hanging after one Target (net462) has been tested.
It also hangs on my PC when using nunit-lite.
Attaching then, it only shows `EventPump.Dequeue`
But it shows: 'Test Run Summary' so it knows it is finished?

Debugging midway doesn't work as test that it possibly affects behaviour.
It hangs in a `GetResult` as result of a ForcedStop. 

Main still runs, so it must be something here, but I can't see what changes in this branch would affect this.

stevenaw:
Thanks for the prompt review @manfred-brands 
I had run into hanging behavior earlier when passing in "true" for a debug setting on the happy path tests as it would suspend the test when the debugger attached.

Though this sounds a bit different if it's related to random. I'll take a closer look when I'm able. It's good that it's reproducible outside GH Actions.

stevenaw:
I have to pause investigating for now. I think I have narrowed this down to some interaction of these tests:
- `Load_InvalidTypeSettings_ThrowsException`
- `Load_UnparseableSettings_ThrowsException`

and what happens internally when invalid settings are passed. My unconfirmed suspicion is that the thrown exceptions may be interfering with our signalling to complete a test run, such as perhaps preventing `OnRunCompleted` from being called.

`DebugTests` and `PauseBeforeRun` both appear affected. I am able to get a consistent completed local run when only those test cases are commented out. Both of those settings happen to be used in [`NUnitTestAssemblyRunner.StartRun()`](https://github.com/nunit/nunit/blob/main/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs#L279)

Ex:
```
        [TestCase(FrameworkPackageSettings.NumberOfTestWorkers)]
        [TestCase(FrameworkPackageSettings.DefaultTimeout)]
        //[TestCase(FrameworkPackageSettings.DebugTests)]
#if NETFRAMEWORK
        //[TestCase(FrameworkPackageSettings.PauseBeforeRun)]
#endif
        [TestCase(FrameworkPackageSettings.StopOnError)]
        [TestCase(FrameworkPackageSettings.ThrowOnEachFailureUnderDebugger)]
        [TestCase(FrameworkPackageSettings.SynchronousEvents)]
        [TestCase(FrameworkPackageSettings.RandomSeed)]
        [TestCase(FrameworkPackageSettings.RunOnMainThread)]
```

manfred-brands:
> My unconfirmed suspicion is that the thrown exceptions may be interfering with our signalling to complete a test run, such as perhaps preventing OnRunCompleted from being called.

I think you are correct.

When the cake build hangs, the main thread is finished, but multiple EventPump threads are still running.
This indicates that the WorkComplete is not called.

I added one commit to do a try/catch around StartRun.

It runs locally now.

stevenaw:
Thanks @manfred-brands 
Glad that's all it was. Your changes look good to me. Any objections if we were to squash merge this now?

manfred-brands:
> Thanks @manfred-brands Glad that's all it was. Your changes look good to me. Any objections if we were to squash merge this now?

Go ahead.

stevenaw:
Merged. Thanks for your help on this one @manfred-brands

PR #5102, opened 2026-01-06
Add Enum.Parse<T>() polyfill for .NET Framework builds (#5100)

### LINKED ISSUE #5100: Add an `Enum.Parse<T>()` method for .NET Framework builds
There's a few places in the framework that we call a non-generic `Enum.Parse()` method where the typing is known at compile time: https://github.com/search?q=repo%3Anunit%2Fnunit+%22Enum.Parse%22+path%3A%22src%2FNUnitFramework%2Fframework%22&type=code A generic `Enum.Parse<T>(string, bool)` method exists on newer runtimes and provides better performance there if we can use it. We should add a polyfill for this as a static extension method to allow the framework to seamlessly use the new API.

We should do this as a static extension method over `Enum` living in the `NUnit.Compatibility` namespace within the directory https://github.com/nunit/nunit/tree/main/src/NUnitFramework/framework/Compatibility. The method implementation can simply delegate to the built-in non-generic form of the method. The extension method and entire extension class should be wrapped in an `
#if NETFRAMEWORK` conditional inclusion macro to ensure we only compile our polyfill on .NET Framework builds. An example of this can be seen here: https://github.com/nunit/nunit/blob/7cf73431b9c105198b47112d4a74e43b1b4a58bd/src/NUnitFramework/framework/Api/NUnitTestAssemblyRunner.cs#L18-L20

shaig-mahmudov:
Added a generic `Enum.Parse<T>` polyfill in the `NUnit.Compatibility` namespace, wrapped in a `#if NETFRAMEWORK `conditional. This allows the framework to use the generic API on `.NET Framework` builds where it is not natively available. Closes #5100

### DIFF
--- a/src/NUnitFramework/framework/Compatibility/EnumExtensions.cs
+++ b/src/NUnitFramework/framework/Compatibility/EnumExtensions.cs
@@ -24,7 +24,7 @@ namespace System
             }
             /// <summary>
             /// Converts the string representation of the name or numeric value of one or more
-            /// enumerated constants to an equivalent enumerated object. A parameter specifies.
+            /// enumerated constants to an equivalent enumerated object. A parameter specifies
             /// whether the operation is case-insensitive.
             /// </summary>
             /// <typeparam name="T">The enum type to which to convert.</typeparam>


OsirisTerje:
If you run `build` and `build --target test` locally, you will get the same errors you see in the build log here locally. Makes it easier and faster to fix them.

src/NUnitFramework/framework/Compatibility/Enum.cs
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+
+using System;
+
+namespace NUnit.Compatibility
```
stevenaw: This might be a "miss" on my part in the original spec. There's ambiguity errors reported by the compiler. You can likely help make them go away by changing this namespace to `System` so that it matches the namespace of the built-in `Enum` type.

src/NUnitFramework/framework/Compatibility/Enum.cs
```diff
@@ -0,0 +1,41 @@
...
+    internal static class Enum
+    {
+        /// <summary>
+        /// Converts the string representation of the name or numeric value of one or more 
+        /// enumerated constants to an equivalent enumerated object.
+        /// </summary>
+        /// <typeparam name="T">The enum type to which to convert.</typeparam>
+        /// <param name="value">The string representation of the enumeration name or underlying value.</param>
+        /// <returns>An object of type T whose value is represented by value.</returns>
+        public static T Parse<T>(string value) where T : struct
```
stevenaw: If you convert this to use the C#14 extension member syntax then you can extend static methods onto `Enum` itself, similar to how we've added some of the classic asserts back onto the `Assert` class here: https://github.com/nunit/nunit/blob/main/src/NUnitFramework/nunit.framework.legacy/ClassicAssertExtensions.cs#L13-L19

Doing this will then allow you the update elsewhere in the code where we call `Enum.Parse(typeof(EnumType), string)` to instead call the generic `Enum.Parse<EnumType>(string)` method seamlessly

src/NUnitFramework/framework/Compatibility/Enum.cs
```diff
@@ -0,0 +1,41 @@
...
+using System;
+
+namespace NUnit.Compatibility
+{
+#if NETFRAMEWORK
+    /// <summary>
+    /// Provides a polyfill for the generic Enum.Parse{T} method which is available in 
+    /// .NET Core and newer .NET versions but missing in .NET Framework.
+    /// </summary>
+    internal static class Enum
```
stevenaw: I'd recommend renaming this to `EnumExtensions` so that changing the namespace to `System` doesn't cause conflicts

shaig-mahmudov: Hi @stevenaw, I’ve applied the requested changes:
• Namespace: moved from NUnit.Compatibility to System to match the built-in Enum type and avoid ambiguity.
• Extension syntax: updated the implementation to use extension members so Enum.Parse<T>() works seamlessly.
• Class name: renamed to EnumExtensions to avoid potential conflicts.
• Formatting: adjusted generic constraints (SA1127) to match the style guide.
• Local build: verified with `dotnet build` and `dotnet build src/NUnitFramework/framework/nunit.framework.csproj` across the relevant target frameworks.

Thanks for the guidance — happy to make further adjustments if needed.

shaig-mahmudov:
@dotnet-policy-service agree

src/NUnitFramework/framework/Compatibility/Enum.cs
OsirisTerje: @shaig-mahmudov Look at the code @stevenaw  linked to, or just see image below:
image

This is the new syntax, and then you get rid of the `this Enum _` too.

stevenaw: Thanks for your changes @shaig-mahmudov 

Terje is right about the new syntax as it will allow these methods to _appear_ as if they are static methods on the built-in `Enum` class itself despite that they aren't "built-in" on .NET Framework. The goal behind a polyfill like this is to allow code to call a method on all runtimes we support even if it's only built into .NET itself on newer ones.

In other words, we'd want to also able to update these references within this PR:
- https://github.com/nunit/nunit/blob/7cf73431b9c105198b47112d4a74e43b1b4a58bd/src/NUnitFramework/framework/Internal/RuntimeFramework.cs#L387
- https://github.com/nunit/nunit/blob/7cf73431b9c105198b47112d4a74e43b1b4a58bd/src/NUnitFramework/framework/Api/FrameworkController.cs#L115

To instead call
- `Enum.Parse<RuntimeType>(s, true)` 
- `Enum.Parse<InternalTraceLevel>((string)traceLevelValue, true)` 

As appropriate _without_ needing any `#if NETFRAMEWORK` checks around where we call the method.

shaig-mahmudov:
Hi @stevenaw, @TerjeP, I've updated the Enum.Parse polyfill as suggested:
* Switched to the extension(Enum) syntax in EnumExtensions.cs.
* Removed the this Enum _ parameter and updated the XML documentation.
* Updated the calls in RuntimeFramework.cs and FrameworkController.cs to use the cleaner Enum.Parse<T>(...) syntax.
* Ready for another look. Thanks!"

stevenaw:
Thanks @shaig-mahmudov 
The extension method itself looks as expected. Can you please check and see if you forgot to commit any other changes on your end? We'd like to update the calls in `RuntimeFramework.cs` and `FrameworkController.cs` where we use the old `Enum.Parse()` to instead call your new version so that it can be used :)

shaig-mahmudov:
@stevenaw Thanks for the feedback. I’ve updated the enum parsing in `FrameworkController.cs` and `RuntimeFramework.cs` to use the generic `Enum.Parse<T>(..., true)` syntax.

src/NUnitFramework/framework/Compatibility/EnumExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// <summary>
+    /// Provides a polyfill for the generic Enum.Parse{T} method which is available in.
+    /// .NET Core and newer .NET versions but missing in .NET Framework.
+    /// </summary>
+    internal static partial class EnumExtensions
+    {
+        extension(Enum)
+        {
+            /// <summary>
+            /// Converts the string representation of the name or numeric value of one or more.
```
stevenaw:
```suggestion
            /// Converts the string representation of the name or numeric value of one or more
```

src/NUnitFramework/framework/Compatibility/EnumExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+            /// <typeparam name="T">The enum type to which to convert.</typeparam>
+            /// <param name="value">The string representation of the enumeration name or underlying value.</param>
+            /// <returns>An object of type T whose value is represented by value.</returns>
+            public static T Parse<T>(string value)
+                where T : struct
+            {
+                return (T)System.Enum.Parse(typeof(T), value);
+            }
+            /// <summary>
+            /// Converts the string representation of the name or numeric value of one or more.
```
stevenaw:
```suggestion
            /// Converts the string representation of the name or numeric value of one or more
```

src/NUnitFramework/framework/Compatibility/EnumExtensions.cs
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+#if NETFRAMEWORK
+namespace System
+{
+    /// <summary>
+    /// Provides a polyfill for the generic Enum.Parse{T} method which is available in.
```
stevenaw:
```suggestion
    /// Provides a polyfill for the generic Enum.Parse{T} method which is available in
```

src/NUnitFramework/framework/Compatibility/EnumExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+            /// <param name="value">The string representation of the enumeration name or underlying value.</param>
+            /// <returns>An object of type T whose value is represented by value.</returns>
+            public static T Parse<T>(string value)
+                where T : struct
+            {
+                return (T)System.Enum.Parse(typeof(T), value);
+            }
+            /// <summary>
+            /// Converts the string representation of the name or numeric value of one or more
+            /// enumerated constants to an equivalent enumerated object. A parameter specifies.
```
stevenaw:
```suggestion
            /// enumerated constants to an equivalent enumerated object. A parameter specifies
```

stevenaw:
Thanks @shaig-mahmudov ! Your changes look good, I appreciate your quick responses to each feedback.
I've pushed a few minor edits just to remove some extra periods in the xmldocs. A final build is running now and I'll merge upon its successful completion. 

Thanks for your contribution!

stevenaw (approved):

shaig-mahmudov:
@stevenaw Thanks for the review, edits, and guidance! I appreciate it and I’m happy to contribute to the project.

PR #5105, opened 2026-01-09
Fix for #5096, by making a copy of the workitem list.

### LINKED ISSUE #5096: Unclear error if MTP times out in a setup fixture
Take the following unit test app - the setup takes 10 secs to complete, but MTP has been called with `--timeout 5s` to force unit to timeout during setup.  When this happens, unit fails with a `Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')` error.

csproj:
```xml
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <EnableNUnitRunner>true</EnableNUnitRunner>
    <TestingPlatformCommandLineArguments>$(TestingPlatformCommandLineArguments) --timeout 5s</TestingPlatformCommandLineArguments>
  </PropertyGroup>
```

Tests:
```cs
public class Tests
{
    [SetUp]
    public void Setup()
    {
        Thread.Sleep(10000);
    }

    [Test]
    public void Test1()
    {
        Assert.Pass();
    }
}
```
Output:
```
Exit code: 134
  Standard output: Microsoft.Testing.Platform v2.0.2+88c9fc53a3 (UTC 11/11/2025) [osx-arm64 - net10.0]
  NUnit Adapter 6.0.0.0: Test execution started
  Running all tests in /Users/alex/Developer/mtp/artifacts/bin/nunitdemo/debug/nunitdemo.dll
     NUnit3TestExecutor discovered 1 of 1 NUnit test cases using Current Discovery mode, Non-Explicit run
  
  Canceling the test session...
  
  NUnit Adapter 6.0.0.0: Test execution complete
  
  Error output: Unhandled exception. System.AggregateException: One or more errors occurred. (An exception occurred in the driver while stopping the run.)
   ---> NUnit.Engine.NUnitEngineException: An exception occurred in the driver while stopping the run.
   ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
   ---> System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
     at System.Collections.Generic.List`1.get_Item(Int32 index)
     at NUnit.Framework.Internal.Execution.ParallelWorkItemDispatcher.CancelRun(Boolean force)
     at NUnit.Framework.Api.NUnitTestAssemblyRunner.StopRun(Boolean force)
     at NUnit.Framework.Api.FrameworkController.StopRun(Boolean force)
     at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args)
     at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr)
     at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr)
     at System.Reflection.MethodBaseInvoker.InvokeWithOneArg(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
     at NUnit.Engine.Drivers.NUnitNetCore31Driver.ExecuteMethod(MethodInfo method, Object[] args)
     at NUnit.Engine.Drivers.NUnitNetCore31Driver.ExecuteMethod(String methodName, Object[] args)
     at NUnit.Engine.Drivers.NUnitNetCore31Driver.StopRun(Boolean force)
     at NUnit.Engine.Runners.DirectTestRunner.StopRun(Boolean force)
     at NUnit.Engine.Runners.DirectTestRunner.StopRun(Boolean force)
     at NUnit.Engine.Runners.DirectTestRunner.ForcedStop()
     at NUnit.Engine.Runners.MasterTestRunner.StopRun(Boolean force)
     at NUnit.VisualStudio.TestAdapter.NUnitEngine.NUnitEngineAdapter.StopRun() in C:\repos\nunit\nunit3-vs-adapter\src\NUnitTestAdapter\NUnitEngine\NUnitEngineAdapter.cs:line 126
     at NUnit.VisualStudio.TestAdapter.NUnit3TestExecutor.StopRun() in C:\repos\nunit\nunit3-vs-adapter\src\NUnitTestAdapter\NUnit3TestExecutor.cs:line 431
     at NUnit.VisualStudio.TestAdapter.NUnit3TestExecutor.Microsoft.VisualStudio.TestPlatform.ObjectModel.Adapter.ITestExecutor.Cancel() in C:\repos\nunit\nunit3-vs-adapter\src\NUnitTestAdapter\NUnit3TestExecutor.cs:line 254
     at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
  --- End of stack trace from previous location ---
     at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
     at System.Threading.CancellationTokenSource.ExecuteCallbackHandlers(Boolean throwOnFirstException)
     at System.Threading.CancellationTokenSource.ExecuteCallbackHandlers(Boolean throwOnFirstException)
     at System.Threading.TimerQueueTimer.Fire(Boolean isThreadPool)
     at System.Threading.ThreadPoolWorkQueue.Dispatch()
     at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
     at System.Threading.Thread.StartCallback()
  
The following exception occurred when running the test module with RunCommand '/Users/alex/Developer/mtp/artifacts/bin/nunitdemo/debug/nunitdemo' and RunArguments '  --timeout 5s':

System.InvalidOperationException: A test session start event was received without a corresponding test session end.
   at Microsoft.DotNet.Cli.Commands.Test.TestApplication.RunAsync()
   at Microsoft.DotNet.Cli.Commands.Test.TestApplication.RunAsync()
   at Microsoft.DotNet.Cli.Commands.Test.TestApplicationActionQueue.Read(BuildOptions buildOptions, TestOptions testOptions, TerminalTestReporter output, Action`1 onHelpRequested)
```

### LINKED ISSUE #5106: Test execution never ends when Ctrl+C key hit in MTP mode
When using NUnit tests with MTP mode the execution of tests is not terminated when CTRL+C key is pressed in the console. It is visible that CtrlCIssue.UnitTests.Test event is raised, but test engine is running forever from that point. If no CTRL+C is pressed, test session ends in 5 seconds as there is only one test with 5s sleep code.

I tried the same case with MSTest tests, to identify if the issue is in MTP code or NUnit (adapter). When MSTest used instead of NUnit, behavior is correct. After CTRL+C the test session is finished and properly reported. When NUnit is used, I have to kill the process externally (via Task Manager or PowerShell).

How to simulate:
1. Compile the example below
2. Execute ```CtrlCIssue.exe --no-ansi --output Detailed``` command (using no-ANSI console reporter for better logging in the example, MTP logger has no impact)
3. Right after execution when ```NUnit3TestExecutor discovered 1 of 1 NUnit test cases using Current Discovery mode, Non-Explicit run``` message is displayed, hit the CTRL+C key.
4. After few second warning exception is displayed ```The operation was canceled.```, but execution never stops. The message ```Canceling the test session...``` comes from MTP when ```Console.CancelKeyPress``` event is raised.

**Screenshot from the execution**

Image

**CtrlCIssueMSTest.csproj**
```
<Project Sdk="Microsoft.NET.Sdk">
	<PropertyGroup>
		<TargetFramework>net8</TargetFramework>
		<IsPackable>false</IsPackable>
		<IsTestProject>true</IsTestProject>
		<EnableNUnitRunner>true</EnableNUnitRunner>
		<TestingPlatformDotnetTestSupport>true</TestingPlatformDotnetTestSupport>
		<OutputType>Exe</OutputType>
	</PropertyGroup>
	<ItemGroup>
		<PackageReference Include="coverlet.collector" Version="6.0.0" />
		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="18.0.1" />
		<PackageReference Include="Microsoft.Testing.Extensions.Telemetry" Version="2.0.2" />
		<PackageReference Include="Microsoft.Testing.Extensions.TrxReport" Version="2.0.2" />
		<PackageReference Include="Microsoft.Testing.Extensions.VSTestBridge" Version="2.0.2" />
		<PackageReference Include="Microsoft.Testing.Platform" Version="2.0.2" />
		<PackageReference Include="Microsoft.Testing.Platform.MSBuild" Version="2.0.2" />
		<PackageReference Include="NUnit" Version="3.12.0" />
		<PackageReference Include="NUnit.Analyzers" Version="3.8.0" />
		<PackageReference Include="NUnit3TestAdapter" Version="5.1.0" />
	</ItemGroup>
</Project>
```

**UnitTests.cs**
```
using NUnit.Framework;
using System.Threading;

namespace CtrlCIssue
{
    [TestFixture]
    public class UnitTests
    {
        [Test]
        public void Test()
        {
            Thread.Sleep(5000);
        }
    }
}
```

OsirisTerje:
Fixes #5096 
Fixes #5106 

The current code may fail is the list is changed while looping through it.  Some other thread may 1) change the list, since it will not use the same lock, 2) the lock also locks on the list instance, not the ref. , and 3) WorkItemCancelled can modify the list since that code will not honor the lock either.

Whatever is the underlying reason, working on a copy is safer.  

This is the crash before the fix:

image

After the fix, it doesn't crash here but where expected, 

image

We can probably make this nicer too, but that will be in the adapter.

### DIFF
--- a/src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintTests.cs
+++ b/src/NUnitFramework/tests/Constraints/CollectionEquivalentConstraintTests.cs
@@ -437,7 +437,7 @@ public class CollectionEquivalentConstraintTests
         watch.Stop();
         if (watch.ElapsedMilliseconds > LargeCollectionWarnTime)
             Assert.Warn($"{TestContext.CurrentContext.Test.MethodName} took {watch.ElapsedMilliseconds} ms.");
-        if (watch.ElapsedMilliseconds > LargeCollectionFailTime*4)
+        if (watch.ElapsedMilliseconds > LargeCollectionFailTime * 4)
             Assert.Fail($"{TestContext.CurrentContext.Test.MethodName} took {watch.ElapsedMilliseconds} ms.");
     }
 


stevenaw:
Thanks @OsirisTerje good find. 
Approach looks good to me too.

There appears to be one failing test though I haven't checked out why

> StopRun_WhenTestIsRunning_StopsTest (Parallel dispatcher, forced stop) in TestAssemblyRunnerTests

OsirisTerje:
@manfred-brands @stevenaw 
Yes, the failing test is testing the paralleldispatcher, so that is not good, but, it only fails for net framework (462).  This particular test, the StopRun, doesn't run for the .net 6/8/9/10, as it relies on the Thread Abort mechanism.

I have tried to find out what it does, and is bewildered.

For the test, he test itself is killed, but the intermediate suites are not.  The topmost suite (assembly suite) is killed, but not the namespace and class suites. 

The StopRun test loads a test suite like this, with the results as shown:
image

When I stop inside the ParallelWorkItemDispatcher, I do find 5 workitems, but they are not corresponding to the ones I find in the TestStarted events. 

image
The 3 in green there are the assembly, and the two parts of the namespace.

The 2 in red are the other nested test classes in the SlowTests, BBB and CCC.  The AAA class is not present here at all.

**This is where I get confused!**

So, if any of you can shed some light on this, I would greatly appreciate that :D 

Now, the ParallelWorkItemDispatcher clearly works for .Net 6/8/9/10, and the test itself IS killed, so does it matter that the namespace suites are not?  They are not actually running, if I have got this right.


PS: Enabled the test for all .net's, and it fails for those too. Just that one test, exactly same manner.

manfred-brands:
@OsirisTerje The issue seems to to be the ORDER the Suites are cancelled.
Besides to copying and locking your change changes the order from 'last to first' into 'first to last'

When the assembly suite is marked as completed, it disposes the EventPump. This means that the listener no longer gets the events for the completion of the suite and the test won't see those events.

OsirisTerje:
Thanks a ton @manfred-brands !   Really appreciated! 
That small ( --index)  was what I overlooked, and then my head was stuck!  
And also show the importance of being more than one :-)

PR #5113, opened 2026-01-26
Add typed variants of the TestCaseData Fluent Instance Modifiers

### LINKED ISSUE #4717: Provide generic type definitions for `TestCaseData` to make it type safe.
### The Problem
Currently, [`TestCaseData`](https://docs.nunit.org/articles/nunit/writing-tests/TestCaseData.html) uses type-erasure to allow users to pass whatever they want as arguments. 

When dealing with more complex parameters such as collections or functions, this can cause extra work for the developer, and results in repeated type annotations and other cruft. Additionally, it creates an opportunity for type errors at runtime if the developer accidentally passes an incorrectly typed value as a parameter.

For example,
```csharp

[TestFixture]
public class MyTests
{
    [TestCaseSource(typeof(MyDataClass), nameof(MyDataClass.TestCases))]
    public int TestDeferedThingy(int initial, Func<int, int> update) {
        return update(initial);
    }
}

public class MyDataClass
{
    public static IEnumerable<TestCaseData> TestCases {
        get {
            yield return new TestCaseData(1, initial => initial).Returns(1); //<< Compile Error, can't determine type of the callback.
            yield return new TestCaseData(1, 1).Returns(1); //<< Runtime error, can't convert 1 to Func<int, int>
            yield return new TestCaseData(1, int (int i) => i).Returns("1"); //<< Runtime error, can't convert "1" to int
        }
    }
}
```

### A Possible Solution

.NET already provides an example API for a similar problem with the [`Func<...TArgs, TResult>`](https://learn.microsoft.com/en-us/dotnet/api/system.func-1) family of delegates. By providing a similar family of classes `TestCaseData<...TArgs, TResult>`, the above issues could be resolved, as well as providing a nice analogue with the standard library.

For example,
```csharp

[TestFixture]
public class MyTests
{
    [TestCaseSource(typeof(MyDataClass), nameof(MyDataClass.TestCases))]
    public int TestDeferedThingy(int initial, Func<int, int> update) {
        return Thingy.UpdateIfNotZero(initial, update);
    }
}

public class MyDataClass
{
    public static IEnumerable<TestCaseData<int, Func<int, int>, int>> TestCases {
        get {
            yield return new(1, initial => initial).Returns(1); //<< Ok, compiler deduces the type of the callback
            yield return new(1, 1).Returns(1); //<< Informative compile error
            yield return new(1, int (int i) => i).Returns("1"); //<< Informative compile error
        }
    }
}
```

logiclrd:
The `TestCaseData` class has a handful of generic variants that allow the type of the data to be verified at compile time. `TestCaseData` also has a handful of fluent instance modifier methods that permit things such as:

```
  yield return new TestCaseData(many parameters here).SetName("Situation1"); // or what have you
```

But, these don't play well together because the fluent instance modifier methods are defined in the base class and have the base `TestCaseData` class as their return type. This means that you can do this:

```
  static IEnumerable<TestCaseData<int, int>> TestCases()
  {
    yield return new TestCaseData<int, int>(2, 2);
    yield return new TestCaseData<int, int>(-2, -2);
  }
```

...but you can't do this:

```
  static IEnumerable<TestCaseData<int, int>> TestCases()
  {
    yield return new TestCaseData<int, int>(2, 2).SetName("Northwest");
    yield return new TestCaseData<int, int>(-2, -2).SetName("Southeast");
  }
```

If you want to use `SetName` fluently, then you have to erase the types from the return value of the source method, which then opens up the door to making mistakes that are only caught at runtime:

```
  static IEnumerable<TestCaseData> TestCases()
  {
    yield return new TestCaseData<int, int>(2, 2).SetName("Northwest");
    yield return new TestCaseData<int, int>(-2, -2).SetName("Southeast");
    ...
    (other stuff here)
    ...
    yield return new TestCaseData(3.5, 4.5).SetName("oops");
  }
```

This PR creates `new` versions of the fluent instance modifier methods that simply pass their arguments to the base implementation (so there's no duplication) but which return `this` in the context of the typed subclasses.

If there's any compiled code out there referring to the `SetName` on the base classes, it should still run just fine and directly call the base implementation. But newly-compiled code will find the `new` methods hiding the underlying ones, so that `SetName(...)` on a `TestCaseData<T1, T2>` returns that same `TestCaseData<T1, T2>` without erasing type information.

This PR solves one of the outstanding issues in #4717 but does not close it.

#5113 and #5114 in conjunction close #4717.

### DIFF
--- a/src/NUnitFramework/framework/TestCaseData.cs
+++ b/src/NUnitFramework/framework/TestCaseData.cs
@@ -245,7 +245,7 @@ namespace NUnit.Framework
         public new TSelf Returns(object? result)
         {
             base.Returns(result);
-            return (TSelf)this;
+            return GetSelf();
         }
 
         /// <summary>
@@ -260,7 +260,7 @@ namespace NUnit.Framework
         public new TSelf SetName(string? name)
         {
             base.SetName(name);
-            return (TSelf)this;
+            return GetSelf();
         }
 
         /// <summary>
@@ -276,7 +276,7 @@ namespace NUnit.Framework
         public new TSelf SetArgDisplayNames(params string[]? displayNames)
         {
             base.SetArgDisplayNames(displayNames);
-            return (TSelf)this;
+            return GetSelf();
         }
 
         /// <summary>
@@ -293,7 +293,7 @@ namespace NUnit.Framework
         public new TSelf SetArgDisplayNames(params object?[]? displayNames)
         {
             base.SetArgDisplayNames(displayNames);
-            return (TSelf)this;
+            return GetSelf();
         }
 
         /// <summary>
@@ -305,7 +305,7 @@ namespace NUnit.Framework
         public new TSelf SetDescription(string description)
         {
             base.SetDescription(description);
-            return (TSelf)this;
+            return GetSelf();
         }
 
         /// <summary>
@@ -316,7 +316,7 @@ namespace NUnit.Framework
         public new TSelf SetCategory(string category)
         {
             base.SetCategory(category);
-            return (TSelf)this;
+            return GetSelf();
         }
 
         /// <summary>
@@ -328,7 +328,7 @@ namespace NUnit.Framework
         public new TSelf SetProperty(string propName, string propValue)
         {
             base.SetProperty(propName, propValue);
-            return (TSelf)this;
+            return GetSelf();
         }
 
         /// <summary>
@@ -340,7 +340,7 @@ namespace NUnit.Framework
         public new TSelf SetProperty(string propName, int propValue)
         {
             base.SetProperty(propName, propValue);
-            return (TSelf)this;
+            return GetSelf();
         }
 
         /// <summary>
@@ -352,7 +352,7 @@ namespace NUnit.Framework
         public new TSelf SetProperty(string propName, double propValue)
         {
             base.SetProperty(propName, propValue);
-            return (TSelf)this;
+            return GetSelf();
         }
 
         /// <summary>
@@ -361,7 +361,7 @@ namespace NUnit.Framework
         public new TSelf Explicit()
         {
             base.Explicit();
-            return (TSelf)this;
+            return GetSelf();
         }
 
         /// <summary>
@@ -370,7 +370,7 @@ namespace NUnit.Framework
         public new TSelf Explicit(string reason)
         {
             base.Explicit(reason);
-            return (TSelf)this;
+            return GetSelf();
         }
 
         #endregion


logiclrd:
> I would however like some update to test cases to verify their usage.

It wasn't immediately obvious to me _how_ to make a meaningful test, since the methods are replacing existing methods and take exactly the same arguments and have exactly the same side-effects. However, it occurs to me that a test that assigns the return values of the functions to typed variables will fail (to compile) without the change and succeed with it, so I'll make tests that do that.

OsirisTerje:
There is a lot of extra code to go in with this approach.

If you define a virtual Self() method in the TestCaseData class, you can override that in each derived class.  Then you don't need to add anything more in the derived classes at all.

Example:

```csharp
public class TestCaseData : TestCaseParameters
{
    public TestCaseData SetName(string? name)
    {
        TestName = name;
        return Self(); // Virtual method call
    }
    
    protected virtual TestCaseData Self() => this;
}
```

And in the derived classes you just override this like:

```csharp
public class TestCaseData<T> : TestCaseData
{
    protected override TestCaseData<T> Self() => this; // Must override in each class
}
```

You then don't need to repeat code in the derived classes, and the change in the TestCaseData is non-breaking.

logiclrd:
That doesn't solve the problem, unfortunately. The issue is with the return type of the function. The redefinition of the function in each of the subclasses has a return value _of that type_, and that's the key here. When a consumer calls `.SetName` on a `TestCaseData<int, string>`, the return value _must be of type_ `TestCaseData<int, string>`. If that constraint weren't there, the new methods wouldn't be needed at all, because the base implementation would satisfy calls to `.SetName`.

Is the current change breaking? What breaks with it?

manfred-brands:
@OsirisTerje We could reduce the code to a single extension method, but the returns `T` if it wasn't that the compiler picks the base class member method in preference of the extension method and we cannot delete that one as it breaks binary compatibility, even though source compatibility is maintained as the extension method also works for the base class.

OsirisTerje:
Got it, so that means you can use something like a CRTP pattern, add an intermediate generic class that takes your derived class as a template parameter and return that.  That class would then do the overrides generically, and return the correct type. 

That would eliminate the duplicated overrides in the derived classes. 

Something like:

```csharp
public abstract class TestCaseData<TSelf> : TestCaseData
    where TSelf : TestCaseData<TSelf>
{
    public override TSelf SetName(string name)
    {
        base.SetName(name);
        return (TSelf)this;
    }

    // Put all fluent methods here once, returning TSelf.
}

public sealed class TestCaseData<T1, T2> : TestCaseData<TestCaseData<T1, T2>>
{
    public TestCaseData(T1 arg1, T2 arg2)
    {
        // existing ctor work
    }
}
```

If we can avoid duplicating all the code for every derivative class, it would be a lot of code saved.

logiclrd:
Whoops. Mix-and-match with Git Gui. Changed branches in the repo on disk, didn't refresh the GUI, and it pushed the wrong ref!

logiclrd:
Uh okay that's extremely weird. I just explicitly refreshed the GUI and it pushed the same reference. Sigh.

logiclrd:
Superseded by #5138.

PR #5114, opened 2026-01-26
Add TestCaseDataWithReturn types

### LINKED ISSUE #4717: Provide generic type definitions for `TestCaseData` to make it type safe.

logiclrd:
This PR adds a suite of types `TestCaseDataWithReturn<T, ..., TReturn>` that are parallel to `TestCaseData<T, ...>` but provide a typed expected return value.

Automated tests are added exercising the new classes.

Possible usage:

```
  static IEnumerable<TestCaseDataWithReturn<int, int>> DoubleTestCases()
  {
    yield return new TestCaseDataWithReturn<int, int>(2).Returns(4);
    yield return new TestCaseDataWithReturn<int, int>(3, expectedReturnValue: 6);
  }
```

This PR solves one of the outstanding issues in https://github.com/nunit/nunit/issues/4717 but does not close it.

#5113 and #5114 in conjunction close #4717.

### DIFF
--- a/src/NUnitFramework/framework/TestCaseDataWithReturn.cs
+++ b/src/NUnitFramework/framework/TestCaseDataWithReturn.cs
@@ -1,10 +1,14 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
+using System;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+
 namespace NUnit.Framework
 {
     /// <summary>
     /// Implementations of fluent instance modifier methods for TestCaseData&lt;&gt; generic types.
     /// </summary>
-    public abstract class TestCaseDataWithReturnGenericBase<TSelf, TReturn> : TestCaseData
+    public abstract class TestCaseDataWithReturnGenericBase<TSelf, TReturn> : TestCaseParameters
         where TSelf : TestCaseDataWithReturnGenericBase<TSelf, TReturn>
     {
         /// <summary>
@@ -30,7 +34,7 @@ namespace NUnit.Framework
         /// <returns>A modified TestCaseData</returns>
         public TSelf Returns(TReturn result)
         {
-            base.Returns(result);
+            ExpectedResult = result;
             return GetSelf();
         }
 
@@ -43,9 +47,9 @@ namespace NUnit.Framework
         /// <see cref="SetArgDisplayNames(string[])"/> allows you to specify the display names for parameters directly without
         /// needing to use tokens like {m}.
         /// </remarks>
-        public new TSelf SetName(string? name)
+        public TSelf SetName(string? name)
         {
-            base.SetName(name);
+            TestName = name;
             return GetSelf();
         }
 
@@ -59,9 +63,9 @@ namespace NUnit.Framework
         ///     .SetArgDisplayNames("argDisplayName");
         /// </code>
         /// </example>
-        public new TSelf SetArgDisplayNames(params string[]? displayNames)
+        public TSelf SetArgDisplayNames(params string[]? displayNames)
         {
-            base.SetArgDisplayNames(displayNames);
+            ArgDisplayNames = displayNames;
             return GetSelf();
         }
 
@@ -76,9 +80,11 @@ namespace NUnit.Framework
         ///     .SetArgDisplayNames(testData.Name);
         /// </code>
         /// </example>
-        public new TSelf SetArgDisplayNames(params object?[]? displayNames)
+        public TSelf SetArgDisplayNames(params object?[]? displayNames)
         {
-            base.SetArgDisplayNames(displayNames);
+            ArgDisplayNames = displayNames is null
+                ? null
+                : Array.ConvertAll(displayNames, Constraints.MsgUtils.FormatValue);
             return GetSelf();
         }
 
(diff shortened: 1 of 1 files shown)


src/NUnitFramework/framework/TestCaseDataWithReturn.cs
```diff
@@ -0,0 +1,856 @@
...
+        }
+
+        #region Fluent Instance Modifiers
+
+        /// <summary>
+        /// Sets the expected result for the test
+        /// </summary>
+        /// <param name="result">The expected result</param>
+        /// <returns>A modified TestCaseDataWithReturn</returns>
+        public TestCaseDataWithReturn<T, TReturn> Returns(object? result)
```
manfred-brands: Should the type of _result_ not be `TReturn` otherwise there is no type checking.

logiclrd: That's weird. I could have sworn I'd written it that way. Huh. I must have somehow conflated adding constructors with a `TReturn`-typed parameter. That's embarrassing. Thanks for catching it :-)

OsirisTerje:
@logiclrd @manfred-brands 

Is this ready for merging now ?

logiclrd:
> @logiclrd @manfred-brands
> 
> Is this ready for merging now ?

I'll address anything that comes up from any further review, but as of right now I'm not aware of anything outstanding on my end. 🙂

OsirisTerje:
It seems to be the same pattern as in #5113 .   Can you consider if this can be reduced too?

logiclrd:
This can be even better-centralized: after #5113 is merged, this branch can be rebased onto it, and then `TestCaseDataWithReturnGenericBase<>` can be worked into the inheritance tree to share a single common fluent instance modifier implementation.

In fact, I wonder if the same implementation could even do the non-generic `TestCaseData` class, by making it a subclass of the common base...

logiclrd:
Hmm, I've been proofing out that last thought, and it really makes the code very neat and tidy, but I think there's a sticking point that can only be implemented with a breaking change.

With the fully-integrated design, either `TestCaseData` can't have a `Returns` method, or `TestCaseData` can't be the root of the tree.

The problem is that if `Returns(object?)` is inherited, it can't be un-inherited, so if `TestCaseData` is the root of all the configurable specializations `TestCaseParameters`, then its non-generic `Returns(object?)` will be inherited by everything, including the classes that have a typed `Returns(TReturn)` method. This means that if you try to call `Returns` with an argument that isn't a `TReturn`, it'll just silently fall back on the `Returns(object?)` overload.

So, solution 1 is to remove the `Returns` method from `TestCaseData`. But, this breaks anybody who has already written code that uses `TestCaseData` with `Returns`.

Solution 2 is to make `TestCaseData` just a special case of `TestCaseDataWithReturn<T, TReturn>`. Effectively, `TestCaseData`'s entire implementation can be replaced by inheriting from `TestCaseDataWithReturnBase<TestCaseData, object?>`. This solves the problem of typed `Returns` calls, because nobody is inheriting the `object?` overload in `TestCaseData` -- `TestCaseData` is now a leaf. But, this breaks anybody who assumes that `TestCaseData<T>` is a subclass of `TestCaseData`. And everybody does, because test case sources are `IEnumerable<TestCaseData>` which then yield instances of `TestCaseData<T>` or what have you.

Seeing how the code comes together, I personally think this is still a worthwhile change to consider, but it can only be a part of a major version number bump because there's no way to do it without a breaking change to the API. 🙁 

Unless I'm missing a clever solution. 🙂 

I'll throw the `TestCaseData`-as-leaf version into a PR of its own.

logiclrd:
Superseded by #5138.

PR #5121, opened 2026-02-06
Write `Console.Out` to the system output when no test context is available

### LINKED ISSUE #5120: `Console.Write` where `ExecutionContext` flow is suppressed is never written
NUnit tracks the current test using an `AsyncLocal` in `TestExecutionContext`, but async state propagation does not work when `System.Threading.ExecutionContext` flowing is suppressed, causing the test context to be unavailable. In this scenario, using `Console.Write{,Line}(...)` gets rerouted through `TextCapture` into an ad-hoc `TestContext` (`AdhocContext`) of which the output is never written anywhere. `TextCapture`'s aim is to write such out-of-band writes into its `_defaultWriter`, but the implicit availability of `AdhocContext` since #2279 has made those fallback branches dead code.

This problem surfaced when using [ASP.NET's `TestServer`](https://github.com/dotnet/aspnetcore/blob/v10.0.0/src/Hosting/TestHost/src/TestServer.cs), which by-default handles requests [using `ThreadPool.UnsafeQueueUserWorkItem`](https://github.com/dotnet/aspnetcore/blob/7387de91234d3ef751fa50b3d1bfede4130213ff/src/Hosting/TestHost/src/HttpContextBuilder.cs#L155-L165), which doesn't propagate the current `ExecutionContext` (unlike `ThreadPool.QueueUserWorkItem`). The test context can be propagated into the ASP.NET request processing by [enabling `TestServerOptions.PreserveExecutionContext`](https://github.com/dotnet/aspnetcore/blob/7387de91234d3ef751fa50b3d1bfede4130213ff/src/Hosting/TestHost/src/TestServerOptions.cs#L18-L21), but it took me a while to find this option when debugging to figure out where logging got lost.

I'll be opening a PR shortly with a fix.

JoostK:
The implicit availability of `AdhocContext` since #2279 has made the fallback branches dead code, causing the text to be written into the ad-hoc test result which is never written anywhere. This commit restores the original intent of `TextCapture` by detecting this situation and redirecting the write into the default writer.

Fixes #5120

### DIFF
--- a/src/NUnitFramework/tests/TextOutputTests.cs
+++ b/src/NUnitFramework/tests/TextOutputTests.cs
@@ -16,6 +16,7 @@ namespace NUnit.Framework.Tests
     {
         private const string SOME_TEXT = "Should go to the result";
         private const string ERROR_TEXT = "Written directly to console";
+        private const string SOME_MORE_TEXT = "Should be written to stdout";
         private static readonly string NL = Environment.NewLine;
 
         private static string CapturedOutput => TestExecutionContext.CurrentContext.CurrentResult.Output;
@@ -44,7 +45,8 @@ namespace NUnit.Framework.Tests
             {
                 using (ExecutionContext.SuppressFlow())
                 {
-                    Task.Run(() => Console.Write("Should be written to stdout")).Wait();
+                    Task.Run(() => Console.Write(SOME_MORE_TEXT))
+                        .Wait();
 
                     Console.Write(SOME_TEXT);
                 }
@@ -57,7 +59,7 @@ namespace NUnit.Framework.Tests
             Assert.Multiple(() =>
             {
                 Assert.That(TextOutputTests.CapturedOutput, Is.EqualTo(SOME_TEXT));
-                Assert.That(stdout.ToString(), Is.EqualTo("Should be written to stdout"));
+                Assert.That(stdout.ToString(), Is.EqualTo(SOME_MORE_TEXT));
             });
         }
 
@@ -71,7 +73,8 @@ namespace NUnit.Framework.Tests
             {
                 using (ExecutionContext.SuppressFlow())
                 {
-                    Task.Run(() => Console.WriteLine("Should be written to stdout")).Wait();
+                    Task.Run(() => Console.WriteLine(SOME_MORE_TEXT))
+                        .Wait();
 
                     Console.WriteLine(SOME_TEXT);
                 }
@@ -84,7 +87,7 @@ namespace NUnit.Framework.Tests
             Assert.Multiple(() =>
             {
                 Assert.That(TextOutputTests.CapturedOutput, Is.EqualTo(SOME_TEXT + NL));
-                Assert.That(stdout.ToString(), Is.EqualTo("Should be written to stdout" + NL));
+                Assert.That(stdout.ToString(), Is.EqualTo(SOME_MORE_TEXT + NL));
             });
         }
 


src/NUnitFramework/tests/TextOutputTests.cs
```diff
@@ -30,6 +32,48 @@ public void ConsoleWriteLine_WritesToResult()
...
+                {
+                    var context = TestExecutionContext.CurrentContext;
+                    Assert.That(context, Is.InstanceOf<TestExecutionContext.AdhocContext>());
+
+                    var initialOut = context.CurrentResult.Output;
+                    Console.Write("Should not be captured in the ad-hoc context");
+                    Assert.That(context.CurrentResult.Output, Is.EqualTo(initialOut));
+                }).Wait();
+
+                Console.Write(SOME_TEXT);
```
JoostK: This isn't strictly necessary, but I figured it emphasises that writes outside of spawned tasks do get captured in the test output.

src/NUnitFramework/tests/TextOutputTests.cs
```diff
@@ -30,6 +32,48 @@ public void ConsoleWriteLine_WritesToResult()
...
+            using (ExecutionContext.SuppressFlow())
+            {
+                Task.Run(() =>
+                {
+                    var context = TestExecutionContext.CurrentContext;
+                    Assert.That(context, Is.InstanceOf<TestExecutionContext.AdhocContext>());
+
+                    var initialOut = context.CurrentResult.Output;
+                    Console.Write("Should not be captured in the ad-hoc context");
+                    Assert.That(context.CurrentResult.Output, Is.EqualTo(initialOut));
```
JoostK: Alternatively this could be `Assert.That(context.CurrentResult.Output, Is.Empty);`, but that would undesirably also pass in case `context.CurrentResult` is a different instance each time.

src/NUnitFramework/tests/TextOutputTests.cs
```diff
@@ -30,6 +32,48 @@ public void ConsoleWriteLine_WritesToResult()
...
 
+        [Test]
+        public void ConsoleWrite_WritesToResult_AdhocContext()
+        {
+            using (ExecutionContext.SuppressFlow())
+            {
+                Task.Run(() =>
+                {
+                    var context = TestExecutionContext.CurrentContext;
+                    Assert.That(context, Is.InstanceOf<TestExecutionContext.AdhocContext>());
```
JoostK: Alternatively, this could be `Assert.That(context, Is.Not.SameAs(originalContext));` (with `originalContext` captured outside the task) to avoid a symbolic reference to the implementation detail of `AdhocContext`.

src/NUnitFramework/framework/Internal/Execution/TextCapture.cs
```diff
@@ -38,7 +38,7 @@ public override void Write(char value)
         {
             var context = TestExecutionContext.CurrentContext;
 
-            if (context is not null && context.CurrentResult is not null)
+            if (context is not TestExecutionContext.AdhocContext)
```
JoostK: This isn't null safe anymore as the null check is redundant. `nullable` checking agrees, so I removed it.

JoostK:
@dotnet-policy-service agree

JoostK:
**Top:** without the fix, the `Console.Write()` text is not written anywhere (the tests pass because I disabled the assertions to ensure the test output is otherwise the same as with the fix).
**Bottom:** With fix. The line ends up being written after the test finishes, I suspect because the console writer may not be flushed on write.

Screenshot 2026-02-06 at 22 35 32

src/NUnitFramework/tests/TextOutputTests.cs
```diff
@@ -30,6 +32,48 @@ public void ConsoleWriteLine_WritesToResult()
...
+
+                    var initialOut = context.CurrentResult.Output;
+                    Console.Write("Should not be captured in the ad-hoc context");
+                    Assert.That(context.CurrentResult.Output, Is.EqualTo(initialOut));
+                }).Wait();
+
```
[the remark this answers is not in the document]
JoostK: I don't like this suggestion; there is no result to fetch so `Wait()` seems more appropriate than `GetAwaiter().GetResult()`, and `await` is somewhat error prone due to the `AsyncFlowScope` returned from `ExecutionContext.SuppressFlow()` that must be disposed from the same thread that started the scope.

src/NUnitFramework/framework/Internal/Execution/TextCapture.cs
manfred-brands: In optimal case yes, but due to code setting properties in all kinds of places this cannot be relied upon.
`TestExecutionContext` disables `CS8618` _Non-nullable field must contain a non-null value when exiting constructor_.  The class itself does not set `CurrentResult` and there are tests against `null` in the code base.

When converting nunit to nullable, declaring this as possible null would affect to much code.
I might have to put this back on my todo list to ensure the value is set in the constructor.

manfred-brands: The same test on _context_ is done in all 3 overloads.
I added a commit to move this into a single function and convert the 3 methods to one liners

src/NUnitFramework/tests/TextOutputTests.cs
```diff
@@ -30,6 +34,60 @@ public void ConsoleWriteLine_WritesToResult()
...
+        public void ConsoleWrite_WritesToResult_AdhocContext()
+        {
+            var savedOut = Console.Out;
+            var stdout = new StringWriter();
+            Console.SetOut(new TextCapture(stdout));
+            try
+            {
+                using (ExecutionContext.SuppressFlow())
+                {
+                    Task.Run(() => Console.Write("Should be written to stdout")).Wait();
```
manfred-brands: The constant `"Should be written to stdout"` is used 4 times and should be moved to a constant.

I thought that the co-pilot comment could be addressed by making the method `async`:
```csharp
await Task.Run(() => Console.WriteLine(SOME_MORE_TEXT))
          .ConfigureAwait(continueOnCapturedContext: true);
```

But this returns on the same `SyncrhonizationContext`, not necessarily the same `Thread`.
It worked for the `Write` test but not for `WriteLine` ...

manfred-brands (approved):
I made two small changes, otherwise the code is fine.

src/NUnitFramework/framework/Internal/Execution/TextCapture.cs
JoostK: Thanks, this is much better. My aim was to keep the diff small

JoostK:
I suppose this one is ready to be merged.

PR #5123, opened 2026-02-12
Disable TimeoutCommand for anything but .NET Framework

### LINKED ISSUE #4119: Apartment(ApartmentState.STA) doesn't works together with Timeout
Apartment(ApartmentState.STA) doesn't works together with Timeout

I'm currently updating projects from .NET Framework (NUnit 3.10.1) to .NET 6 (NUnit 3.13.3) and found this problem. Real tests do some WPF relates stuff, so it requires STA apartment. The Timeout is used to prevent the test from running indefinitely when something goes wrong.

An example of minimalistic tests:

```cs
[TestFixture]
public class ApartmentTests
{
   [Test, Apartment(ApartmentState.STA), Timeout(1000)]
   public void SomeTestWithTimeout()
   {
      Console.WriteLine("SomeTestWithTimeout {0} {1}", Thread.CurrentThread.ManagedThreadId, Thread.CurrentThread.GetApartmentState());
      Assert.Pass();
   }
   
   [Test, Apartment(ApartmentState.STA)]
   public void SomeTestWithoutTimeout()
   {
      Console.WriteLine("SomeTestWithoutTimeout {0} {1}", Thread.CurrentThread.ManagedThreadId, Thread.CurrentThread.GetApartmentState());
      Assert.Pass();
   }
}
```

nunit3-console output:
```nunit
NUnit Console 3.15.0 (Release)
Copyright (c) 2022 Charlie Poole, Rob Prouse
30 мая 2022 г. 21:21:19

Runtime Environment
   OS Version: Microsoft Windows 10.0.18362
  Runtime: .NET 6.0.4

Test Files
    .\TestProject1.dll

SomeTestWithoutTimeout 6 STA
SomeTestWithTimeout 8 MTA
```

manfred-brands:
Fixes #4119 

TimeoutAttribute has been Obsolete since version 4.0.0 for anything but .NET Framework

### DIFF
--- a/src/NUnitFramework/framework/Internal/Commands/CancelAfterCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/CancelAfterCommand.cs
@@ -17,7 +17,7 @@ namespace NUnit.Framework.Internal.Commands
         private readonly IDebugger _debugger;
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="TimeoutCommand"/> class.
+        /// Initializes a new instance of the <see cref="CancelAfterCommand"/> class.
         /// </summary>
         /// <param name="innerCommand">The inner command</param>
         /// <param name="timeout">Timeout value</param>
--- a/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
+++ b/src/NUnitFramework/framework/Internal/Commands/TimeoutCommand.cs
@@ -5,7 +5,6 @@ using System;
 using System.Threading;
 #else
 using System;
-using System.Threading.Tasks;
 #endif
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal.Abstractions;
@@ -93,29 +92,9 @@ namespace NUnit.Framework.Internal.Commands
         /// <returns>A TestResult</returns>
         public override TestResult Execute(TestExecutionContext context)
         {
-            try
-            {
-                using (new TestExecutionContext.IsolatedContext())
-                {
-                    var testExecution = Task.Run(() => innerCommand.Execute(TestExecutionContext.CurrentContext));
-                    var timedOut = Task.WaitAny(new Task[] { testExecution }, _timeout) == -1;
-
-                    if (timedOut && !_debugger.IsAttached)
-                    {
-                        context.CurrentResult.SetResult(
-                            ResultState.Failure,
-                            $"Test exceeded Timeout value of {_timeout}ms");
-                    }
-                    else
-                    {
-                        context.CurrentResult = testExecution.GetAwaiter().GetResult();
-                    }
-                }
-            }
-            catch (Exception exception)
-            {
-                context.CurrentResult.RecordException(exception, FailureSite.Test);
-            }
+            context.CurrentResult.SetResult(
+                ResultState.Error,
+                $"TargetFramework doesn't support timeout on tests.");
 
             return context.CurrentResult;
         }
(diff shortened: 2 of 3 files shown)


OsirisTerje (approved):

chrischu:
I do not agree with this change. We used the TimeoutAttribute to easily detect tests that run too long and it was working actually (even though we're on .NET Core 10).

CancelAfterAttribute is not a viable alternative for us since it would require us to rewrite ALL our tests to check the passed in CancellationToken.

manfred-brands:
> it was working actually

Not quite true.
The tests were never stopped and continued to run in the background. No teardown code would run.

chrischu:
Mhh so there's no actually viable alternative. We could build our own ITestAction but also that will not be able to invoke the TearDown after the timeout, right (because there's no way to interrupt the actual test execution)?

chrischu:
Mhh there is apparently a way to achieve what my team wants with this code:
```csharp
using System;
using System.Threading;
using System.Threading.Tasks;
using NUnit.Framework;
using NUnit.Framework.Internal.Commands;

[AttributeUsage(AttributeTargets.Assembly)]
public class AssemblyTimeoutAttribute : NUnitAttribute, IWrapTestMethod
{
    private readonly int _timeoutMilliseconds;

    public AssemblyTimeoutAttribute(int timeoutMilliseconds)
    {
        _timeoutMilliseconds = timeoutMilliseconds;
    }

    public TestCommand Wrap(TestCommand command)
    {
        return new TimeoutTestCommand(command, _timeoutMilliseconds);
    }

    private class TimeoutTestCommand : DelegatingTestCommand
    {
        private readonly int _timeoutMilliseconds;

        public TimeoutTestCommand(TestCommand innerCommand, int timeoutMilliseconds)
            : base(innerCommand)
        {
            _timeoutMilliseconds = timeoutMilliseconds;
        }

        public override TestResult Execute(TestExecutionContext context)
        {
            Exception caughtException = null;

            var task = Task.Run(() =>
            {
                try
                {
                    context.CurrentResult = innerCommand.Execute(context);
                }
                catch (Exception ex)
                {
                    caughtException = ex;
                }
            });

            bool completedInTime = task.Wait(_timeoutMilliseconds);

            if (!completedInTime)
            {
                // Timeout occurred
                context.CurrentResult.SetResult(ResultState.Failure,
                    $"Test exceeded timeout of {_timeoutMilliseconds} ms");
            }
            else if (caughtException != null)
            {
                // Propagate original exception
                throw caughtException;
            }

            return context.CurrentResult;
        }
    }
}
```

That should work correctly, right?

manfred-brands:
This has the same problems as the pre-4.5 implementation.

If _completedInTime_ is false, your actual test is still running.

Any Exception thrown afterwards is lost.
Depending on the wrapping order, the Teardown might be executed whilst the test is still running or not at all.

chrischu:
Is there any other way to achieve a per-test timeout that does not require adding CancellationTokens everywhere?

manfred-brands:
What is the nature of your tests that they could hang?
Is it because they do network calls or are they iterating calculations that don't finish

chrischu:
It's happening with web tests with Selenium WebDriver. We're unsure what the actual problem is yet (as we just recently added the [Timeout] attributes to actually discover *which* tests sometimes hang).

drauch:
If the tests still run in the background it does not matter much as the process is going to exit and thereby all still running tests too? Or does it hang the process? Than it would make sense to call `Environment.Exit()` in this attribute?

chrischu:
In the end we decided to go with the `--blame-hang-timeout` parameter of `dotnet test`. We just didn't consider it because we misunderstood it to be a timeout for ALL tests (instead of per tests). Now that we realized this mistake `--blame-hang-timeout` is clearly the right solution.

PR #5125, opened 2026-02-15
Addition of a "NoTests" attribute to allow configuration of how empty parameterized tests are handled

### LINKED ISSUE #3663: When TestCaseSource produces empty cases collection, test should be Passed/Failed/Inconclusive per creator expectation
## Description

I have large set of tests which are run against different versions o Visual Studio and I am migrating them to NUnit3. My goal is to test against single VS while I run my tests, I enforce it by setting environment variable with combination of TestCaseSourceAttribute and method which produces my test cases. Of course single test may be run for different VS versions, and some tests should not be run at all against some VS versions.

The test methods are parameterless, as they get VS version being tested from static class, so signature of example test method looks as follows:

```
// Run test on VS 2015, 2017 and 2019
[TestCaseSource(typeof(VsVersionTestCaseFactory), "VSVersionDataSource", new[] { VSVersion.SinceVS2015 })]
public void Test() { ... } // Test takes VSVersion from provider class instead of using parameters
```

Where `VsVersionTestCaseFactory` is my static class and `VSVersionDataSource` is method which produces test cases:

```
public static IEnumerable<TestCaseData> VSVersionDataSource(VSVersion vsVersion)
		{
			if (IsAllowed(vsVersion)) // checks if vsVersion is the same as version passed in EnvVar
			{
				yield return new TestCaseData();
			}
		}

```

My expectation is that if I will get no test case from `VSVersionDataSource` then my test method will not be called at all, but it is not so simple.

I have checked what is happening uder the hood and found this [piece of code](https://github.com/nunit/nunit/blob/830e0df681eb38490742355900436b4a988a0b52/src/NUnitFramework/framework/Attributes/TestCaseSourceAttribute.cs#L144-L153):

```
            // Part of TestCaseSourceAttribute.cs:

            // If count > 0, error messages will be shown for each case
            // but if it's 0, we need to add an extra "test" to show the message.
            if (count == 0 && method.GetParameters().Length == 0)
            {
                var parms = new TestCaseParameters();
                parms.RunState = RunState.NotRunnable;
                parms.Properties.Set(PropertyNames.SkipReason, "TestCaseSourceAttribute may not be used on a method without parameters");

                yield return _builder.BuildTestMethod(method, suite, parms);
            }
```

Basically my case is specifically coded in `TestCaseSourceAttribute` to create artificial `NotRunnable` test case which causes my test to fail. I understand, that in most cases this may be most expected. I have already read other issues, e.g. #3573 and #1933 and I agree that they are reasonable.

## Expectation

If there are conflicting expectations on this issue, I believe each developer should be able to pick expected result of such test, either by providing additional argument to `TestCaseSourceAttribute`, e.g.:

`[TestCaseSource(typeof(VsVersionTestCaseFactory), "VSVersionDataSource", new[] { VSVersion.SinceVS2015 }, StateWhenNotRun = RunState.Ignored)]`

or by applying additional attribute:

```
[FailIfNotRun] // force test to fail if there are no test cases produced
[TestCaseSource(typeof(VsVersionTestCaseFactory), "VSVersionDataSource", new[] { VSVersion.SinceVS2015 })]
```

of course without applying `FailIfNotRunAttribute` test should be silently ignored (should not be failed).

stevenaw:
Fixes #3663 

Allows for the configuration of a "default test result" when a parameterized test method has no child tests. Theories are specifically excluded as discussed in the original issue. The original discussion also mentions that a `Values` producing no child tests should be an error flagged by an analyzer and so I opted not to special case that here for simplicity.

### DIFF
--- a/src/NUnitFramework/tests/Attributes/NoTestsAttributeTests.cs
+++ b/src/NUnitFramework/tests/Attributes/NoTestsAttributeTests.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Charlie Poole, Rob Prouse and Contributors. MIT License - see LICENSE.txt
 
 using System.Linq;
+using System.Reflection;
 using NUnit.Framework.Interfaces;
 using NUnit.Framework.Internal;
 using NUnit.Framework.Tests.TestUtilities;
@@ -17,12 +18,12 @@ namespace NUnit.Framework.Tests.Attributes
     {
         public static class AssemblyLevelSupport
         {
-            private static System.Reflection.Assembly _dynamicTestAssembly;
+            private static Assembly _dynamicTestAssembly;
 
             [OneTimeSetUp]
             public static void OneTimeSetUp()
             {
-                var compiler = new TestCompiler(typeof(NoTestsAttributeFixture).Assembly);
+                var compiler = new TestCompiler(typeof(NoTestsAttributeFixture));
                 _dynamicTestAssembly = compiler.GenerateInMemoryAssembly(
                     NoTestsAttributeFixture.AssemblyLevelSupportCode);
             }
--- a/src/NUnitFramework/tests/Attributes/ParallelizableAttributeTests.cs
+++ b/src/NUnitFramework/tests/Attributes/ParallelizableAttributeTests.cs
@@ -59,15 +59,16 @@ namespace NUnit.Framework.Tests.Attributes
             Assert.That(context.ParallelScope, Is.EqualTo(scope & ParallelScope.ContextMask));
         }
 
-        private static readonly ParallelScope[] Scopes = [
-            ParallelScope.None,
-            ParallelScope.Self,
-            ParallelScope.Fixtures,
-            ParallelScope.Children,
-            ParallelScope.All,
-            ParallelScope.Self | ParallelScope.Children,
-            ParallelScope.Self | ParallelScope.Fixtures
-        ];
+        private static readonly ParallelScope[] Scopes =
+            [
+                ParallelScope.None,
+                ParallelScope.Self,
+                ParallelScope.Fixtures,
+                ParallelScope.Children,
+                ParallelScope.All,
+                ParallelScope.Self | ParallelScope.Children,
+                ParallelScope.Self | ParallelScope.Fixtures
+            ];
 
         [Test]
         public void MayNotCombineParallelScopeSelfAndParallelScopeNone()
--- a/src/NUnitFramework/tests/Syntax/InvalidCodeTests.cs
+++ b/src/NUnitFramework/tests/Syntax/InvalidCodeTests.cs
@@ -15,7 +15,7 @@ namespace NUnit.Framework.Tests.Syntax
             _compiler = new TestCompiler();
         }
 
-        private static readonly string Template1 =
+        private const string Template1 =
 @"using System;
 using NUnit.Framework;
 using NUnit.Framework.Constraints;


src/NUnitFramework/framework/Attributes/NoTestsAttribute.cs
```diff
@@ -0,0 +1,48 @@
...
+using System;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+
+namespace NUnit.Framework.Attributes
+{
+    /// <summary>
+    /// Indicates the default status of a parameterized test method or test fixture containing no executable child tests.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = false, Inherited = false)]
```
manfred-brands: Should we allow putting this on the `Assembly` level?

src/NUnitFramework/framework/Attributes/NoTestsAttribute.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// Applies the default status of a test with no executable children.
+        /// </summary>
+        /// <param name="test">The test to modify</param>
+        override public void ApplyToTest(Test test)
+        {
+            base.ApplyToTest(test);
+
+            if (test.HasChildren)
```
manfred-brands: If we DO allow this on the assembly level, should this then become recursive or at least have further nesting?

src/NUnitFramework/testdata/NoTestsAttributeFixture.cs
```diff
@@ -0,0 +1,143 @@
...
+                }
+
+                [TestCaseSource(typeof(NoTestsAttributeFixture), nameof(EmptyData))]
+                public static void MethodDoesntSpecify(int actual, int expected)
+                {
+                    Assert.That(actual, Is.EqualTo(expected));
+                }
+            }
+
+            [TestFixture, NoTests(Framework.Interfaces.TestStatus.Passed)]
```
manfred-brands: SA1134: StyleCop.CSharp.ReadabilityRules: Attributes should not share line

Please put attributes on their own line.
You are currently mixing styles.

src/NUnitFramework/framework/Attributes/NoTestsAttribute.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                foreach (var child in test.Tests)
+                {
+                    if (!child.Properties.ContainsKey(PropertyNames.NoTests))
+                    {
+                        child.Properties.Add(PropertyNames.NoTests, _defaultStatus);
```
manfred-brands: I wonder if we could do this without copying values.
Allocating memory and copying for each test is expensive.

src/NUnitFramework/framework/Internal/Execution/CompositeWorkItem.cs
```diff
@@ -96,9 +96,20 @@ protected override void PerformWork()
...
+                            else
                             {
-                                Result.SetResult(ResultState.Failure, "No test cases were provided");
+                                if (Test.TestType == "Theory")
+                                {
+                                    Result.SetResult(ResultState.Failure, "No test cases were provided");
+                                }
+                                else
+                                {
+                                    var property = Test.Properties.Get(PropertyNames.NoTests);
```
manfred-brands: Instead of the copying, could we do a recursive search.
First look in the test, then the fixture, then the suite, then the assembly?
It is an error case so should happen infrequent and doesn't require us to copy values for every test, including every test resulting from parameters.

stevenaw: I was on the fence about this one myself. That was my initial implementation and one that I only reverted in my last commit. I had thought that perhaps this attribute might be seldom-used and that it might be worth the trade-off to copy the values instead of search up the tree for every test in every case.

Though you are right about the error case and perhaps it would still be better to search instead of copy. I should be able to tackle this by just amending my most recent commit.

src/NUnitFramework/framework/Attributes/NoTestsAttribute.cs
stevenaw: This should be pretty possible to do and to test for

stevenaw:
Thanks @manfred-brands 
I've pushed some initial changes, but I wouldn't say it's ready for a review just yet. At a minimum I would want to give it a review myself first, which I won't have time for just yet.

Also, on this suggestion of yours

> I'm also uncertain about the NoTest name and wonder if NoParameters is better?

`NoParameters` is certainly more descriptive to the current scenario. One reason I could think of for `NoTests` is that it's a bit more generic and would describe how to set a single (at this point in time) behaviour of how to handle when "no tests" are present regardless of why. It might (??) be more future-proof.

@OsirisTerje what are your thoughts? I had been following your naming idea with what I've done so far. I'm just guessing at your original rationale.

OsirisTerje:
I suggested `NoTests` (note the `s`)  because that is the observable effect, the Test suite which owns the tests that come from the TestCaseSource, will have `NoTests` if the TestCaseSource is empty for any reason.

To me, a TestCaseSource produce TestCases.   

`NoParameters` sounds more technical to me, although it sorts of say that we have no incoming parameters that could create tests, but I am struggling a bit with that.  

NoTests is not perfect either, but a longer name would be too much.  It is an attribute that says how the TestSuite should behave when there are no tests coming from the TestCaseSource.  

One could say `NoTestCaseBehaviour` or `NoTestCases` too.  The latter could be better than `NoTests`.

stevenaw:
I think the code is now ready for another review, though we might want to wait to merge until we decide if we are going to start on a v5 branch soon

OsirisTerje:
But is this breaking ?

stevenaw:
There should be no breaking changes here @OsirisTerje . I mostly suggested we wait to decide about focusing on v5 changes in case we decided _not_ to pursue a 4.6 as that could change where we merge it to.

I don't think there's any "breaking" reason to hold off on merging this now that we seem to have decided we'll be pursuing a 4.6 release off `main`. I _do_ think we should hold off on merging this until we can get 4.5.1 out in order to keep those changes simple and manageable.

OsirisTerje:
The tests fails on command line `build --target test`, but works in Test Explorer....  😕

stevenaw:
Thanks I can take another look. They had all worked through CLI for me earlier as well as in GH Actions

stevenaw:
@OsirisTerje I've rebased and re-pushed. The tests pass for me locally as well as in GH Actions. What error or failures are you seeing?

OsirisTerje:
The build here works, so I assume it is fine.

OsirisTerje:
@stevenaw Can you fix the conflicts ?  Or should I?

stevenaw:
@OsirisTerje I'm happy to :)

src/NUnitFramework/tests/Attributes/NoTestsAttributeTests.cs
```diff
@@ -0,0 +1,279 @@
...
+using ValueSourceTestData = NUnit.TestData.NoTestsAttributeFixture.ValueSource;
+
+namespace NUnit.Framework.Tests.Attributes
+{
+    [TestFixture]
+    internal static class NoTestsAttributeTests
+    {
+        public static class AssemblyLevelSupport
+        {
+            private static readonly TestCompiler Compiler = new(typeof(NoTestsAttributeFixture).Assembly);
```
stevenaw: I have to update this to use `OneTimeSetUp` to follow the conventions where we use this elsewhere. This will be ready for another look after that

src/NUnitFramework/tests/Attributes/NoTestsAttributeTests.cs
```diff
@@ -0,0 +1,279 @@
...
+    internal static class NoTestsAttributeTests
+    {
+        public static class AssemblyLevelSupport
+        {
+            private static System.Reflection.Assembly _dynamicTestAssembly;
+
+            [OneTimeSetUp]
+            public static void OneTimeSetUp()
+            {
+                var compiler = new TestCompiler(typeof(NoTestsAttributeFixture).Assembly);
```
manfred-brands: Nitpick, you can drop the `.Assembly` suffix as the `TestCompiler` class has an overload for `Type`.

src/NUnitFramework/tests/Attributes/ParallelizableAttributeTests.cs
```diff
@@ -59,16 +59,15 @@ public void ApplyScopeToContext(ParallelScope scope)
             Assert.That(context.ParallelScope, Is.EqualTo(scope & ParallelScope.ContextMask));
         }
 
-        private static readonly ParallelScope[] Scopes = new ParallelScope[]
-        {
+        private static readonly ParallelScope[] Scopes = [
```
manfred-brands: Nitpick. I prefer the `[` to be on its own line to match what we do for '{'.

src/NUnitFramework/tests/Syntax/InvalidCodeTests.cs
```diff
@@ -9,17 +7,12 @@ namespace NUnit.Framework.Tests.Syntax
...
-            typeof(Assert),
-        ];
-
         private TestCompiler _compiler;
 
         [OneTimeSetUp]
         public void OneTimeSetUp()
         {
-            _compiler = new TestCompiler(ReferencedTypes);
+            _compiler = new();
```
manfred-brands: Please don't drop the type when it is now part of the declaration.
Yes the compiler knows the type, but someone looking at the code has to look up the declaration first.

```suggestion
            _compiler = new TestCompiler();
```

src/NUnitFramework/tests/Syntax/InvalidCodeTests.cs
```diff
@@ -50,7 +43,7 @@ public void CodeShouldNotCompile(string fragment)
                 Assert.Fail("Code fragment \"" + fragment + "\" should not compile but it did");
         }
 
-        private static readonly string Template2 =
+        private const string Template2 =
```
manfred-brands, line 46: To be consistent, you should also change the declaration of `Template1` to `const`.

src/NUnitFramework/tests/Attributes/LifeCycleAttributeTests.cs
```diff
@@ -14,19 +12,12 @@ namespace NUnit.Framework.Tests.Attributes
...
-            typeof(DirectoryAssert),
-        ];
-
         private TestCompiler _compiler;
 
         [OneTimeSetUp]
         public void OneTimeSetUp()
         {
-            _compiler = new TestCompiler(ReferencedTypes);
+            _compiler = new(typeof(AssemblyLevelFixtureLifeCycleTest).Assembly);
```
manfred-brands, line 20: Please don't drop the type when it is now part of the declaration.
Yes the compiler knows the type, but someone looking at the code has to look up the declaration first.

```suggestion
            _compiler = new TestCompiler();
```

The reference to the AssemblyLevelFixtureLifeCycleTest isn't needed.
Yes, the _code string_ are defined in that class but only contain standard test code.

manfred-brands (approved):
@stevenaw Approved with a few nitpicks.

stevenaw:
Thanks @manfred-brands 
All your suggestions look good. It's late here now so I'll commit the easy ones through GH and then push a follow-up sometime over the weekend when I'm able.

src/NUnitFramework/tests/Attributes/LifeCycleAttributeTests.cs
stevenaw: This one is needed. The test fixture inherits from a class in the `testdata` assembly

PR #5127, opened 2026-02-17
Fixes #5126 TestResult.*Counts

### LINKED ISSUE #5126: `ITestResult.TotalCount` does not seem to behave as specified
For test process statistics/reporting I need to retrieve the number of executed tests. Using `ITestResult.TotalCount` does not return the result expected, due to the fact that same tests are `[Explicit]` and not executed during a "normal" test run.

`ITestResult.TotalCount` specifies:

> Gets the number of test cases executed when running the test and all its children.

However, in my opinion, an `[Explicit]` test cannot be be considered "executed" during a "normal" test run. The [Repro Sample](https://github.com/user-attachments/files/25347798/Issue5126.zip) demonstrates this:

Actual:
```
=> Issue5126.SetUpFixture
TotalCount = 2
```
Expected:
```
=> Issue5126.SetUpFixture
TotalCount = 1
```
Alternative approach:
Keep `TotalCount` as is but additionally provide `ExecutedCount`.

[Issue5126.zip](https://github.com/user-attachments/files/25347798/Issue5126.zip)

maettu-this:
- Some minor cleanup
- Documentation of `*Counts` refined
- `InitiatedCount` and `CompletedCount` added

Feel free to merge/cherry-pick/reject as feasible.

Fixes #5126

### DIFF
--- a/src/NUnitFramework/framework/Interfaces/ITestResult.cs
+++ b/src/NUnitFramework/framework/Interfaces/ITestResult.cs
@@ -103,10 +103,18 @@ namespace NUnit.Framework.Interfaces
         }
 
         /// <summary>
-        /// Gets the number of asserts executed
+        /// Gets the number of test cases that passed
         /// when running the test and all its children.
         /// </summary>
-        int AssertCount
+        /// <remarks>
+        /// Count reflects test cases that return with
+        /// <see cref="Assert.Pass()"/>,
+        /// <see cref="Assert.Pass(string)"/>,
+        /// <see cref="Assert.Charlie()"/>,
+        /// as well as test cases that return without an
+        /// assertion.
+        /// </remarks>
+        int PassCount
         {
             get;
         }
@@ -136,23 +144,6 @@ namespace NUnit.Framework.Interfaces
         /// <see cref="Assert.Warn(string)"/>.
         /// </remarks>
         int WarningCount
-        {
-            get;
-        }
-
-        /// <summary>
-        /// Gets the number of test cases that passed
-        /// when running the test and all its children.
-        /// </summary>
-        /// <remarks>
-        /// Count reflects test cases that return with
-        /// <see cref="Assert.Pass()"/>,
-        /// <see cref="Assert.Pass(string)"/>,
-        /// <see cref="Assert.Charlie()"/>,
-        /// as well as test cases that return without an
-        /// assertion.
-        /// </remarks>
-        int PassCount
         {
             get;
         }
@@ -194,6 +185,15 @@ namespace NUnit.Framework.Interfaces
         /// <see cref="Assert.Inconclusive()"/>.
         /// </remarks>
         int InconclusiveCount
+        {
+            get;
+        }
+
+        /// <summary>
+        /// Gets the number of asserts executed
+        /// when running the test and all its children.
+        /// </summary>
+        int AssertCount
         {
             get;
         }
--- a/src/NUnitFramework/framework/Internal/Results/TestResult.cs
+++ b/src/NUnitFramework/framework/Internal/Results/TestResult.cs
@@ -346,6 +346,20 @@ namespace NUnit.Framework.Internal
         public abstract int TotalCount { get; }
 
         /// <summary>
+        /// Gets the number of test cases that passed
+        /// when running the test and all its children.
+        /// </summary>
+        /// <remarks>
+        /// Count reflects test cases that return with
+        /// <see cref="Assert.Pass()"/>,
+        /// <see cref="Assert.Pass(string)"/>,
+        /// <see cref="Assert.Charlie()"/>,
+        /// as well as test cases that return without an
+        /// assertion.
+        /// </remarks>
+        public abstract int PassCount { get; }
+
+        /// <summary>
         /// Gets the number of test cases that failed
         /// when running the test and all its children.
         /// </summary>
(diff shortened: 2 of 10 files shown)


maettu-this:
Note that the order of the Fail/Warn/Pass properties differs among `ITestResult/TestResult/TestSuiteResult` and `TestResult*Tests`, for the latter the order is Pass/Fail/Warn. Either let me know whether I shall clean this along with this PR, or fix it when merging.

OsirisTerje:
@maettu-this Thanks for this!   Looks good.  

Just do the cleaning in this PR.

Also, from my comments, would be nice if you also elaborated on this in the documentation, ref my links there.

@nunit/framework-team  Let me merge this PR, I am in the middle of a release. :-)

maettu-this:
> Just do the cleaning in this PR.

Which way around does the team prefer the order?

- Total/Fail/Warn/Pass? (From bad to good. This is the current order in [TestResult](https://github.com/nunit/nunit/blob/main/src/NUnitFramework/framework/Internal/Results/TestResult.cs#L288-L310))
- Total/Pass/Fail/Warn? (The natural order? This is the current order in e.g. [TestResult.xsd](https://github.com/nunit/nunit/blob/main/src/NUnitFramework/framework/Schemas/TestResult.xsd#L27-L30) which I have have just now realized needs to be extended as well)
- Total/Pass/Warn/Fail? (From good to bad. This is my personal favorite.)

OsirisTerje:
> Which way around does the team prefer the order?

I am not so concerned about the order, so any way is fine by me.  This is just source code, so. 

Also be aware that some tools order fields and properties alphabetically.  Use of such tools on this code would change "the order".

If order matters somehow somewhere it would be in the generated XML, there it might help for those few that read it, to see them in a meaningful way.  Your suggested order is fine by me there.

maettu-this:
Changes to XSD/XML now pass CI.

src/NUnitFramework/framework/Interfaces/ITestResult.cs
```diff
@@ -77,19 +77,44 @@ string Message
...
         {
             get;
         }
 
         /// <summary>
-        /// Gets the number of asserts executed
+        /// Gets the number of test cases that got initiated
         /// when running the test and all its children.
         /// </summary>
-        int AssertCount
```
OsirisTerje, line 92: Seems you have renamed this ?  
Existing methods in the interface have to be there, so that we avoid breaking changes.  

It can just redirect to the new one, with a comment it is legacy.

src/NUnitFramework/framework/Interfaces/ITestResult.cs
```diff
@@ -107,16 +139,24 @@ int FailCount
...
         {
             get;
         }
 
         /// <summary>
-        /// Gets the number of test cases that passed
+        /// Gets the number of test cases that have completed
         /// when running the test and all its children.
         /// </summary>
-        int PassCount
```
OsirisTerje, line 119: Same as the comment above, it has to stay.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
```diff
@@ -286,39 +288,120 @@ public int AssertCount
         }
 
         /// <summary>
-        /// Gets the number of test cases executed
+        /// Gets the number of test cases that got initiated
         /// when running the test and all its children.
         /// </summary>
-        public abstract int TotalCount { get; }
```
OsirisTerje, line 292: has this been removed?  Note:  No breaking changes, right?

maettu-this:
Looking at the different orders in detail, I came to the conclusion to make the least intrusive cleanup by following the order in the `Assert` class implementation: Pass - Fail - Warn - Ignore/Skip - Inconclusive.

Cleanup passes CI as well => @OsirisTerje the PR is ready to be reviewed.

maettu-this:
@OsirisTerje please see the individual https://github.com/nunit/nunit/pull/5127/commits:

1. Review all the way through the second last commit, no breaking changes, just the two new added counts.
2. Review the last commit, no breaking changes either, just the reordering where orders mismatched.

src/NUnitFramework/framework/Interfaces/ITestResult.cs
maettu-this: The last reorderig commit https://github.com/nunit/nunit/pull/5127/changes/4fc7e8a27c0d7d6993c1336ed75b237227e3164a shows how `PassCount` and `AssertCount` got "moved" to the matching location.

src/NUnitFramework/framework/Interfaces/ITestResult.cs
maettu-this: The last reorderig commit https://github.com/nunit/nunit/pull/5127/changes/4fc7e8a27c0d7d6993c1336ed75b237227e3164a shows how `PassCount` and `AssertCount` got "moved" to the matching location.

src/NUnitFramework/framework/Internal/Results/TestResult.cs
maettu-this: The last reorderig commit https://github.com/nunit/nunit/pull/5127/changes/4fc7e8a27c0d7d6993c1336ed75b237227e3164a shows how the counts got "moved" to the matching location.

PR #5130, opened 2026-02-19
Fix for issue5129

### LINKED ISSUE #5129: Passing `Array` as test parameter fails with 4.5.0
With release 4.5.0, passing `Array` as test parameter fails with 
`System.IndexOutOfRangeException : Index was outside the bounds of the array.`

The following example works fine with 4.4.0, but breaks with 4.5.0:
```C#
public sealed class Class1
{
    private static IEnumerable<Array> Testcases
    {
        get
        {
            yield return new[] { 1, 2, 3 };
        }
    }

    [TestCaseSource(nameof(Testcases))]
    public void Foo(Array values)
    {
        Assert.That(values.Length, Is.EqualTo(3));
    }
}
```

OsirisTerje:
Fixes #5129

### DIFF
--- a/src/NUnitFramework/tests/Internal/Extensions/ArgumentExtensionsTest.cs
+++ b/src/NUnitFramework/tests/Internal/Extensions/ArgumentExtensionsTest.cs
@@ -73,7 +73,7 @@ namespace NUnit.Framework.Internal.Extensions
         }
 
         [Test]
-        public void ShouldnpackArrayAsArguments_MatchingType_ParamsArray_ReturnsFalse()
+        public void ShouldUnpackArrayAsArguments_MatchingType_ParamsArray_ReturnsFalse()
         {
             var parameters = new MethodWrapper(GetType(), nameof(MethodWithIntArrayAndParamsIntArray)).GetParameters();
             Assert.That(parameters.ShouldUnpackArrayAsArguments(new int[] { 1, 2, 3 }), Is.False);


stevenaw:
Thanks @OsirisTerje 
This looks ok at a glance but I'll try and take a closer look this evening

OsirisTerje:
I've simplified the code in the last two commits, moving the complex logic into extension methods, and added iunit tests for those.  The code that was changed should now be more intention revealing.

src/NUnitFramework/framework/Internal/Extensions/ArgumentsExtensions.cs
```diff
@@ -9,57 +9,138 @@ namespace NUnit.Framework.Internal.Extensions
...
+            /// Determines whether <paramref name="array"/> should be unpacked into individual
+            /// test arguments, or passed as a single argument to the test method.
+            /// </summary>
+            /// <param name="array">The array yielded by the <see cref="TestCaseSourceAttribute"/> source,
+            /// which may either be a container holding one argument per element (e.g. <c>new object[] { a, b, c }</c>)
+            /// or the actual argument value itself (e.g. an <c>int[]</c> to be passed to an <see cref="System.Array"/> parameter).</param>
+            public bool ShouldUnpackArrayAsArguments(Array array)
+            {
+                // A params parameter always absorbs the elements individually
+                if (parameters.LastParameterIsParamsArray())
```
stevenaw: question: If I'm reading this correctly, would this be handled correctly? I think I would expect `expectedArgs` to contain an array of `[1, 2, 3]` and that `extra` would have nothing

```
private static IEnumerable<Array> Testcases
    {
        get
        {
            yield return new[] { 1, 2, 3 };
        }
    }

public void MethodWhichReceivesArrayAndHasParams(Array expectedArgs, params int[] extra)
```

src/NUnitFramework/framework/Internal/Extensions/ArgumentsExtensions.cs
```diff
@@ -9,57 +9,138 @@ namespace NUnit.Framework.Internal.Extensions
...
+
+                Type paramType = parameters[0].ParameterType;
+                Type arrayType = array.GetType();
+
+                // Exact type match: the array IS the argument, not a container
+                if (paramType == arrayType)
+                    return false;
+
+                // Multiple parameters: each element maps to one parameter
+                if (argsNeeded > 1)
```
stevenaw, line 128: question: I'm a little unclear about this one. I notice a comment in the previous implementation:

> // If array has the same number of elements as parameters
                                    // and it does not fit exactly into single existing parameter
                                    // we believe that this array contains arguments

Would this short-circuit that check it's checked below this?

OsirisTerje:
I have added the repro code, with a few extra tests, here:  https://github.com/nunit/nunit.issues/tree/main/Issue5129

src/NUnitFramework/framework/Internal/Extensions/ArgumentsExtensions.cs
OsirisTerje: Yeah, in this case it looks wrong.  (Funny how things become clearer when it gets a name :-)) .  
I didn't change this logic though -- , but this is meant to handle signatures like:

`public void MethodWhichReceivesIntAndHasParams(ìnt value, params int[] extra)`.

The signature with Array and params doesn't make sense, as you say, nothing will get into the `extra`.

Also, `public void MethodWhichReceivesIntAndHasArray(ìnt value, Array extra)`  fails.

I added two tests to the repro for this case, the int case worked as it should, got one value, and the rest to the extra, but the Array + params crashed.  It should have been caught in a check earlier imho.  Or, we could handle it as a special case - if the earlier parameters contain a collection we simply ignore the last collection.

I believe we have to set up some "rules" on how this can be used, so that we don't have to cover all possible constructions here.

src/NUnitFramework/framework/Internal/Extensions/ArgumentsExtensions.cs
OsirisTerje: Not sure about what you mean here - believe I need some more context.

I read it as when I have an array (which is a prerequisite for this, and covers all collections), and I have more than one parameter we assume that is a set of individual parameters.  That has to match exactly to the number of elements in the array.  (If it doesn't, we get internal crashes on that too, so also something we should give a better error message on.)  
I've debugged through the code, and it seems to me it works.

But I understand you see something here, so please elaborate :-)

PS:  We don't really have a good spec of what TestCaseSource can do.  The documentation we have show a few of the ways to use it, but everything regarding collections are absent.  Perhaps we should do something about that.

src/NUnitFramework/framework/Internal/Extensions/ArgumentsExtensions.cs
stevenaw: Thanks @OsirisTerje It's good to hear that the logic was just moved and not changed.

Indeed, these have been hard to fully and properly test in the past. I can't speak to the original rules, though the general guideline we've taken in the past has been to keep it simple and to just try to ensure that we maintain consistent behaviour as if the args from the sources were passed to the test method directly via a regular invocation. 

Though that of course is anything but simple :)

Ex: 
```csharp
private static IEnumerable<Array> Testcases
    {
        get
        {
            yield return new[] { 1, 2, 3 };
        }
    }

    [TestCaseSource(nameof(Testcases))]
    public void Foo(Array values)
    {
        Assert.That(values.Length, Is.EqualTo(3));
    }

    [TestCaseSource(nameof(Testcases))]
    public void Foo2(Array values, params int[] extra)
    {
        Assert.That(values.Length, Is.EqualTo(3));
    }

  // Should be marshaled the same as:
  Foo(Testcases.First())
  Foo2(Testcases.First())
```

I haven't tried yet to see how those actually behave

src/NUnitFramework/framework/Internal/Extensions/ArgumentsExtensions.cs
stevenaw: I think I had a hard time mapping the logic in your new helper to the original code, at least when I was looking at it last night. I was concerned that a case like this, where `argsNeeded` is greater than 1 but not equal to the number of arguments could still cause the array to be unpacked even though the number of arguments in the test method differs from the number of elements in the array.

```
private static IEnumerable<Array> Testcases
{
    get
    {
        yield return new[] { 1, 2, 3 };
    }
}

[TestCaseSource(nameof(Testcases))]
public void Foo(int a, int b)
{
}
```

I admit the same disclaimer as my other comment. I haven't tried running the code yet to see how these behave.

stevenaw: > PS: We don't really have a good spec of what TestCaseSource can do. The documentation we have show a few of the ways to use it, but everything regarding collections are absent. Perhaps we should do something about that.

This is a very good point. I have an issue for myself to add examples for optional, params, etc, but even after that I suspect there would still be gaps

src/NUnitFramework/framework/Internal/Extensions/ArgumentsExtensions.cs
OsirisTerje: The test `public void Foo2(Array values, params int[] extra)` will crash.   So you mean we should fix this, even if the code is meaningless ?

src/NUnitFramework/framework/Internal/Extensions/ArgumentsExtensions.cs
OsirisTerje: Your example here will also crash.  In these cases there has to be an exact match in number of elements and number of parameters.  And the crashes are inside our code, so we 1) have to fix this - but then we have to be clear how we should handle overflow and underflow  2) we catch these earlier and provide  good error message for it

Did this actually work earlier ?

OsirisTerje: I made a new repro, moved back to version 3.13, and noted:

For  public void Foo(int a, int b)
1. Too few parameters or too many parameters:  Fails with message:     Not enough arguments provided, provide at least 3 arguments.   

For  public void Foo(int value, params int[] extra)

2. Fails in 3.13 with     Not enough arguments provided, provide at least 2 arguments.   

This one works directly in 5129 fix.

About 1:   We should fix this to be equal to 3.13.   I might have missed something in the latest refactoring here.  I'll check this tomorrow, if we agree we should be equal to 3.13.

4.4.0 behaves just as 3.13

The image below show the results for all 3 versions we're talking about now:

image

Interesting to see that 4.5.0. and 5129 fix overflow/underflow is not even properly identified.

So with the fix, we only need to fix the overflow/underflow, and then at least make it back to 3.13-4.40 level.   

One could argue that an overflow, we could ignore the overflow arguments,  and for underflow we could assume the parameter overflow values have their default value.     But for now, I think fixing the error messages is the way we go.  

The overflow/underflow is not really errors either, they are failures in the test setup, so it should perhaps be Inconclusive ?

stevenaw: Thanks for catching that!
And FWIW I would agree that any regressions from previous behaviour should be fixed. Bugs related to new features within 4.5 (such as anything specific to method signatures with params or optional args) should also be fixed but separately as I would suggest they are lower priority. They would also likely have workarounds in that a Test/Value/Fixture source could just be updated to pass those args explicitly where relevant.

OsirisTerje:
@manfred-brands @stevenaw 
Anything missing here now, or good to go merge?

src/NUnitFramework/framework/Internal/Extensions/ArgumentsExtensions.cs
manfred-brands: The code isn't meaningless if the test data would be something with a nested array:
`new object[] { new int[] { 1, 2 }, 3, 4 }`

manfred-brands:
@OsirisTerje I haven't had time to look at this in detail.
Your changes with the naming make it clearer, but it is hard to see what the actual fix was.

I will make some time tomorrow to look at this in detail.

manfred-brands:
@OsirisTerje I added a test which has an array inside the array:
```csharp
        private static readonly object[] NestedArray =
        [
            new object[] { new int[] { 2, 3, 4 }, 2 + 3 + 4, 2 * 2 + 3 * 3 + 4 * 4 }
        ];
```
And the following two test signatures:
```csharp
 public void TestWithArrayAndIndividualParameters(int[] values, int sum, int sumSquared)
public void TestWithArrayAndParamsArray(int[] values, params int[] sums)
```

Both pass!

However if I have a test with too many arguments:
```csharp
public void TestWithArrayAndTooManyIndividualParameters(int[] values, int sum, int sumSquared, int sumCubed)
```

The behaviour is correct, just the error message could be improved:
It fails with `Method requires 4 arguments but only 1 were supplied` 
Either the array wasn't unpacked or the error counts the pre-unpacking number of parameters.

However, if there are not enough arguments if fails with an `IndexOutOfRangeException`.
```csharp
 public void TestWithArrayAndNotEnoughIndividualParameters(int[] values, int sum)
```

This is caused by a loop iterating over 2 arrays (_Arguments_ and _parameters_) but testing one `.Length`.
It could be cause by the same error as above as this test is inside an `argsProvided == argsNeeded` where both are `2`, even though `Arguments.Length` is `3`

Ok, I found that issue.
Still looking for the first one.
Debugging is a crime as the code is used for all other tests somewhere.

I want to move these two tests to the testdata project so I can run them here and assert the error message.

manfred-brands:
One other test that fails:

```csharp
[TestCase(new int[] { 7, 8 }, new int[] { 9, 10 })]
[TestCase(new int[] { 11, 12 })]
public void TestWithArrayOfArrays(params int[][] arrays)
```

The one with two elements passes, the one with one doesn't.

When calling it directly in C# they both pass:
```csharp
TestWithArrayOfArrays(new int[] { 7, 8 }, new int[] { 9, 10 });
TestWithArrayOfArrays(new int[] { 11, 12 });
```

So some more work to be done.

I'll see if I can figure this out later today.

OsirisTerje:
Thanks a ton @manfred-brands !

We should get rid of exceptions like IndexOutOfRangeException.

manfred-brands:
I made some changes:

* Added a few more case and have the argument count issue solved. It was dealt with in two places and one wasn't needed.
* Optimize one Unpack/Repack scenario. Not sure how often it would be used.
* Don't unpack an array if it matches the only argument, even if it is a params array.
* We cannot keep an array argument if it has the wrong type. Fixes issue mentioned [above](https://github.com/nunit/nunit/pull/5130#issuecomment-3976265165)

But I initially broke one specific oddity (i.e. classic feature):

```csharp
private static readonly object[] IgnoreWhiteSpaceData =
{
    new object[] { new SimpleObjectCollection("x", "y", "z", " z ") },
    new object[] { new[] { "a", "b", "c", " c " } }
};

[TestCaseSource(nameof(IgnoreWhiteSpaceData))]
public void HonorsIgnoreWhiteSpace(IEnumerable actual)
```

The test accept an IEnumerable and `object[]` is one, so it doesn't unpack the array.
However the test actually expects the first element in that array, matching the declaration below:

```csharp
private static readonly object[] IgnoreWhiteSpaceData =
{
    new SimpleObjectCollection("x", "y", "z", " z "),
    new[] { "a", "b", "c", " c " }
};
```

This is an old test and I have restored the behaviour to unwrap an object[] array with 1 element.

```csharp
private static readonly object[] OneElementWrapped =
[
    new object[] { 42 }
];
private static readonly object[] TwoElementsWrapped =
[
    new object[] { 40, 2 }
];

[TestCaseSource(nameof(OneElementWrapped))]
[TestCaseSource(nameof(TwoElementsWrapped))]
public void TestObjectArray(Array array)
{
    Assert.That(array, Is.Not.Null);
}
```

The first case failed with `Object of type 'System.Int32' cannot be converted to type 'System.Array'.` because it unwrapped the single integer, the second case succeeds as the 2 element array is not unwrapped.

I have confirmed that this is the same with NUnit 3.

Not sure how relevant it is as newer code would declare the arrays as `new int[]` which works for both cases.

stevenaw:
Thanks for the thorough analysis and descriptive summary @manfred-brands . This area of the codebase is challenging and it's good we're keeping all these behaviors while cleaning things up and adding further coverage and documentary discussion here.

It's been difficult to find a sufficient amount of focused time for me to do a thorough and accurate follow-up review on this but from your description it sounds like you've covered my earlier concerns. 

@OsirisTerje Please don't wait on my review on this if you happen to get a chance to look first and are happy with @manfred-brands 's changes.

OsirisTerje:
I'll think we just merge this now.  If we get more flak on this, we can rework a bit more.

PR #5134, opened 2026-02-22
Add generic SameAs method and more test coverage of edge cases

### LINKED ISSUE #2933: Is.SameAs with a reference type constraint
### Problem

If you use `Is.SameAs(3)` or with any value type in general (enums, structs, primitive types), you get a guaranteed failure because a new object gets allocated on the heap when you pass the value type to the `object` parameter of `SameAs`. (Known as 'boxing.') On the flip side, `Is.Not.SameAs(valueTypedExpression)` is guaranteed to always succeed.

If in theory a .NET runtime decides to begin caching common boxed value types like `(object)true` or low integers (the JVM does this), then instead of a guaranteed failure, you end up simply testing a side effect of the runtime you're on.

### Guardrail

As a guardrail, what if we added an overload to require that the parameter be a reference type? That makes sure the argument expression is typed as a class, delegate, interface, or dynamic, all of which have an innate concept of reference equality:

```diff
namespace NUnit.Framework.Constraints
{
    public abstract class Is
    {
        public static SameAsConstraint SameAs(object expected);

+       public static SameAsConstraint SameAs<T>(T expected) where T : class;
    }
}
```

After this change, inherently broken `Is.SameAs` usages would become compile-time errors because the generic overload will always be considered a better match even if the constraint doesn't match:

`Is.SameAs(3)`

> The type 'int' must be a reference type in order to use it as parameter `T`

The workaround would be `Is.SameAs((object)3)` or `IsSameAs<object>(3)`, and both of those should make it clear that what you're doing is inherently meaningless because nothing can be the same as a newly-boxed, used-once value except as a runtime side-effect.

We'd want to keep the old overload to avoid a binary break. (That is, if my helper library references NUnit 3.10 and you use both my helper library and NUnit 3.11, it'd cause a runtime IL failure if my helper library references the non-generic `SameAs(object)` method.)

We could consider this for `Assert.AreSame`, too.

stevenaw:
Fixes #2933 

This PR also adds some additional test coverage over some edge cases such as:
- Interned strings

### DIFF
--- a/src/NUnitFramework/framework/Constraints/SameAsConstraint.cs
+++ b/src/NUnitFramework/framework/Constraints/SameAsConstraint.cs
@@ -15,13 +15,24 @@ namespace NUnit.Framework.Constraints
     /// SameAsConstraint tests whether an object is identical to
     /// the object passed to its constructor
     /// </summary>
-    public class SameAsConstraint(object? expected) : Constraint(expected)
+    public class SameAsConstraint : Constraint
     {
+        private readonly object? _expected;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SameAsConstraint"/> class.
+        /// </summary>
+        /// <param name="expected">The expected object.</param>
+        public SameAsConstraint(object? expected) : base(expected)
+        {
+            _expected = expected;
+        }
+
         /// <summary>
         /// The Description of what this constraint tests, for
         /// use in messages and in the ConstraintResult.
         /// </summary>
-        public override string Description => "same as " + MsgUtils.FormatValue(expected);
+        public override string Description => "same as " + MsgUtils.FormatValue(_expected);
 
         /// <summary>
         /// Test whether the constraint is satisfied by a given value
@@ -30,7 +41,7 @@ namespace NUnit.Framework.Constraints
         /// <returns>True for success, false for failure</returns>
         public override ConstraintResult ApplyTo<TActual>(TActual actual)
         {
-            bool hasSucceeded = ReferenceEquals(expected, actual);
+            bool hasSucceeded = ReferenceEquals(_expected, actual);
 
             return new ConstraintResult(this, actual, hasSucceeded);
         }
--- a/src/NUnitFramework/tests/Assertions/SameFixture.cs
+++ b/src/NUnitFramework/tests/Assertions/SameFixture.cs
@@ -29,8 +29,8 @@ namespace NUnit.Framework.Tests.Assertions
         [Test]
         public void SameUsingInternedConstStrings()
         {
-            string s1 = "S1";
-            string s2 = "S1";
+            string s1 = string.Intern(new string('A', 2));
+            string s2 = string.Intern(s1);
             Assert.That(s1, Is.SameAs(s2));
         }
 


stevenaw:
This is now ready for discussion and review. We might want to wait to merge until we decide if we are going to start on a v5 branch soon

OsirisTerje:
@stevenaw Do you mean this is a breaking change?

Btw,  I need to sit down and figure out for myself what the SameAs really do.  Never used it.....

stevenaw:
@OsirisTerje This constraint is pretty simple, it just checks for reference equality. 

This PR is non-breaking in it's current form, however there would be ways we could take this further if we had a place to put breaking changes:
- Remove the non generic overload which takes `object?`
- Flatten the constraint hierarchy here so that the generic constraint inherits right from `Constraint`

If we _were_ ready to have a v5 branch but we still wanted to maintain a 4.6 release too then it might make sense to have this PR in the current form go towards 4.6 and for those next steps to be added on top as part of porting to a v5 branch

stevenaw:
I approached implementing this a bit different than usual in that I've added some of these tests (for now) to document current behavior whether intentional or not.

FWIW I'm inclined to agree that boxed/nullable ints would likely constitute a misuse of the API even though they could have reference equality in the cases outlined in the test. A misuse in that the underlying type doesn't support reference equality, and also a misuse in that _any_ other combination of values would always be reference unequal.

To that end I am reconsidering that the `class?` constraint may make sense after all. We would need to leave the `object?` overload in place of course for binary compatibility and so the nullable/boxed cases would technically continue to pass until we could remove that in a future v5. I can remove the related tests now if we prefer to consider those boxed/nullable scenarios "unsupported".

manfred-brands:
> I can remove the related tests now if we prefer to consider those boxed/nullable scenarios "unsupported".

Yes as it might confuse people that look at out code.
Especially as you have to suppress NUnit2040 for the `Nullable<T>` struct.

Different boxed variables, even if initialized to the same value are never the same, only the literal same variable is.

```csharp
        [Test]
        public void Test()
        {
            object one = 1;
            object een = 1;

            Assert.That(one, Is.EqualTo(een));
            Assert.That(one, Is.Not.SameAs(een));
        }
```

OsirisTerje:
@stevenaw 

> If we were ready to have a v5 branch but we still wanted to maintain a 4.6 release too then it might make sense to have this PR in the current form go towards 4.6 and for those next steps to be added on top as part of porting to a v5 branch

I have created a V5 branch, we can just start there, but should try to really keep non-breaking stuff in the V4 (main).   We should also use the Breaking label where appropriate.

Also agree we should split this, non-breaking into 4.6, and continue with it in v5.

src/NUnitFramework/framework/Constraints/SameAsConstraint.cs
```diff
@@ -6,24 +6,22 @@ namespace NUnit.Framework.Constraints
...
-        {
-            _expected = expected;
-        }
+    }
 
+    /// <summary>
+    /// SameAsConstraint tests whether an object is identical to
+    /// the object passed to its constructor
+    /// </summary>
+    public class SameAsConstraint(object? expected) : Constraint(expected)
```
manfred-brands: I must have missed this, but I really don't like the primary constructors.
It puts parameters available as mutable fields everywhere.
In reviews it is now not obvious if _expected_ is a field, a local or a parameter.
It also can be changed accidentally as it isn't marked `readonly`.
Potential maintenance headache.

We outlawed the use of these in our company.
I'm not in a position to dictate style here as it is a cooperation, 
but I do suggest to not allow it until it deficiencies are resolved.

Record types are fine because they create read-only properties by default.
So we could consider:
`public record SameAsConstraint(object? Expected) : Constraint(Expected)`

src/NUnitFramework/tests/Assertions/SameFixture.cs
```diff
@@ -17,15 +17,23 @@ public void Same()
...
             var ex = Assert.Throws<AssertionException>(() => Assert.That(ex2, Is.SameAs(ex1)));
             Assert.That(ex?.Message, Does.Contain(expectedMessage));
         }
 
+        [Test]
+        public void SameUsingInternedConstStrings()
+        {
+            string s1 = "S1";
+            string s2 = "S1";
+            Assert.That(s1, Is.SameAs(s2));
```
manfred-brands, line 34: This is test is relying on string interning (different constant string with the same value become the same reference).
It only works inside the same assembly and not across assemblies.
This is not something NUnit controls and hence should not test.

If you want to proof `SameAs` works on `string`, use:
`string s2 = s1;` without the quotes.

manfred-brands (approved):
I'll approve it as my comment have nothing to do with functionality.
Please consider my remarks especially about primary constructors.

src/NUnitFramework/framework/Constraints/SameAsConstraint.cs
stevenaw: Thanks, I had missed the mutability issue there. I 100% agree that we should want this immutable-first by design.

I _think_ records here may be challenging without a larger rewrite as I think records can only inherit from other records. I may be wrong as I'm not at a computer to test it right now. I'll push another set of changes tomorrow with whatever gives us the best immutability guarantees

OsirisTerje: I like the primary constructors 😃    Find them making less clutter.  

We can store that parameter in a read-only field if we like, of assign it to the get only private property, and use that in the code.  Using a property is something I like, because that we get rid of the underscores in field names, and scoping is controlled.

It doesn't prevent us from using the parameter directly however, but would not surprise me if that is something we will see fixed in an upcoming C# version.

src/NUnitFramework/tests/Assertions/SameFixture.cs
stevenaw: FWIW, I look at the purpose of this test a bit differently. NUnit doesn't control interning but consumers _may_ use interning. I felt it would be good to show that NUnit `SameAs` will work for both scenarios. I've updated the test to use the CLR-wide intern pool rather than relying on assembly consts. It should use an intern-ing pattern that works across all runtimes.

src/NUnitFramework/framework/Constraints/SameAsConstraint.cs
stevenaw: I've pushed a commit which I hope balances both of these concerns. The existing non-generic `SameAs` has been converted away from a primary constructor. It had previously been immutable and the PR changed that despite that the `expected` value never changed over the lifetime of the instance. That felt like a behavioural regression to me which I fixed by removing the primary constructor.

The new `SameAs<>` generic constraint I kept using a primary constructor for now as it served to just pass the value through to the base class and this approach felt simpler and produced less code.

OsirisTerje:
Since #2933 is an enhancement, I'll merge this into the v46 branch.

src/NUnitFramework/framework/Constraints/SameAsConstraint.cs
OsirisTerje: I don't think we should generally move away from primary constructors. I have been converting several over just as I have been walking through code.  If we're afraid of them being misused  due to them being mutable, we could add an analyzer rule to prevent that. It could be default off, so that others are not bothered by it, but we could turn it on, and it would then fail our builds if we don't honor it.

manfred-brands: Like described in: https://www.meziantou.net/making-primary-constructor-parameters-read-only.htm

